PR #97, opened 2022-04-16
prototype hybrid sql builder

tobymao:
`print(
    from_("x.y")
    .select("a", "b")
    .select("SUM(c) as c")
    .group_by("1, b")
    .sql(pretty=True)
)`

`print(
    select("a", "b")
    .from_("(Select 1) as z")
    .sql()
)`

i was so excited i wanted to try this out... let me know your thoughts, and happy to work on this branch with you if you like this idea @barakalon

### DIFF
--- a/sqlglot/builder.py
+++ b/sqlglot/builder.py
@@ -1,0 +1,74 @@
+from functools import wraps
+
+import sqlglot.expressions as exp
+
+
+def builder(func):
+    @wraps(func)
+    def builder_method(self, *args, **kwargs):
+        query = Query()
+        query.node = self.node.copy()
+
+        rv = func(query, *args, **kwargs)
+        if rv is not None:
+            return rv
+        return query
+
+    return builder_method
+
+
+class Query:
+    def __init__(self):
+        self.node = exp.Select()
+
+    def to_node(self, obj, spec=None, prefix=None):
+        if isinstance(obj, exp.Expression):
+            return obj
+        if isinstance(obj, Query):
+            return sql.node
+
+        return sqlglot.parse_one(
+            f"{prefix} {obj}" if prefix else str(obj),
+            spec=spec,
+        )
+
+    @builder
+    def select(self, *selects):
+        self.node.args["expressions"].extend(self.to_node(select) for select in selects)
+
+    @builder
+    def from_(self, obj):
+        self.node.args["from"] = self._parse_from(obj)
+
+    @builder
+    def group_by(self, expression):
+        self.node.args["group"] = self.to_node(
+            expression, spec=lambda parser: parser._parse_group, prefix="GROUP BY"
+        )
+
+    def _parse_from(self, obj):
+        return self.to_node(obj, lambda parser: parser._parse_from, prefix="FROM")
+
+    def sql(self, **kwargs):
+        return self.node.sql(**kwargs)
+
+
+class from_(Query):
+    def __init__(self, obj):
+        self.node = exp.Select(
+            **{
+                "from": self._parse_from(obj),
+                "expressions": [],
+                "joins": [],
+            }
+        )
+
+
+class select(Query):
+    def __init__(self, *selects):
+        self.node = exp.Select(
+            **{
+                "expressions": [self.to_node(select) for select in selects],
+                "joins": [],
+            }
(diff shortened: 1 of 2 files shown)


tobymao:
i’m torn on whether it should be a new class or inside the expressions directly

sqlglot/builder.py
```diff
@@ -0,0 +1,74 @@
+from functools import wraps
+
+import sqlglot.expressions as exp
+
+
+def builder(func):
+    @wraps(func)
+    def builder_method(self, *args, **kwargs):
+        query = Query()
+        query.node = self.node.copy()
```
barakalon, line 10: If we simply copy, do the new nodes still maintain a reference to the old parents?

sqlglot/builder.py
```diff
@@ -0,0 +1,74 @@
...
+        self.node.args["expressions"].extend(self.to_node(select) for select in selects)
+
+    @builder
+    def from_(self, obj):
+        self.node.args["from"] = self._parse_from(obj)
+
+    @builder
+    def group_by(self, expression):
+        self.node.args["group"] = self.to_node(
+            expression, spec=lambda parser: parser._parse_group, prefix="GROUP BY"
```
barakalon, line 46: Calling `_parse_*` is a bit hacky. I think some linters should complain about this?

Maybe we can do something similar to this from within the parser class: https://github.com/tobymao/sqlglot/blob/92735b4a754674c61fad3624a634ffc2ab9a2c42/sqlglot/generator.py#L200

The `Expression.key` attribute can be made a class attribute:


```python
class classproperty:
    def __init__(self, f):
        self.f = f

    def __get__(self, obj, owner):
        return self.f(owner)

class Expression:
    ...

    @classproperty
    def key(cls):
        return cls.__name__.lower()
```

Then you can do something like:

```python
class Parser:
    ...
    
    def parse_into(code, expression_type):
        get_attr(f"_parse_{expression_type.key}")(code)

Parser().parse_into("x.y", exp.Table)
```

sqlglot/builder.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+
+    def _parse_from(self, obj):
+        return self.to_node(obj, lambda parser: parser._parse_from, prefix="FROM")
+
+    def sql(self, **kwargs):
+        return self.node.sql(**kwargs)
+
+
+class from_(Query):
```
barakalon, line 56: I think these should just be functions.

sqlglot/builder.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self):
+        self.node = exp.Select()
+
+    def to_node(self, obj, spec=None, prefix=None):
+        if isinstance(obj, exp.Expression):
+            return obj
+        if isinstance(obj, Query):
+            return sql.node
+
+        return sqlglot.parse_one(
```
barakalon, line 30: Dialect?

barakalon (commented):
Certain builder methods need to yield other types of expressions. Or rather, the final query won't always be a `Select` expression. For example, we might need to go from a `Select` expression to a `CTE`.  

Because of that, I figured it'd be easier to write these directly into the Expressions. Otherwise we need wrapper classes for each.

sqlglot/builder.py
barakalon: Or just make public methods for each type haha

barakalon:
Also re: "new class or inside the expressions directly", we need to be able to start modifying an existing expression. Which means if we had a Query class we also need a way to initialize one from an existing expression, which might get a little messy.

sqlglot/builder.py
tobymao: which do you prefer? public or magic method? i can do that refactor in a separate branch

tobymao:
> Also re: "new class or inside the expressions directly", we need to be able to start modifying an existing expression. Which means if we had a Query class we also need a way to initialize one from an existing expression, which might get a little messy.

ok we can try to put this into an expression directly

sqlglot/builder.py
barakalon: Given that there isn't guaranteed to be a method for each expression type, probably safer to have a slew of public methods.

PR #100, opened 2022-04-21
Logical expression builders

barakalon:
Here's a stab at builders for logical expressions.

It supports both ways we discussed:

```python
from sqlglot import and_, or_, condition

and_("a=1", "b=1", or_("c=1", "d=1")

condition("c=1").or_("d=1").and_("b=1", "a=1")
```

@nolamesa @tobymao

### DIFF
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -1,6 +1,6 @@
 from sqlglot.dialects import Dialect
 from sqlglot.errors import ErrorLevel, UnsupportedError, ParseError, TokenError
-from sqlglot.expressions import Expression, Select, And, Or, Not, and_, or_, condition
+from sqlglot.expressions import Expression, Select, and_, or_, not_, condition
 from sqlglot.generator import Generator
 from sqlglot.tokens import Tokenizer, TokenType
 from sqlglot.parser import Parser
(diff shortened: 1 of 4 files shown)


nolamesa (approved):
that was fast 😅

sqlglot/__init__.py
```diff
@@ -1,6 +1,6 @@
 from sqlglot.dialects import Dialect
 from sqlglot.errors import ErrorLevel, UnsupportedError, ParseError, TokenError
-from sqlglot.expressions import Expression, Select
+from sqlglot.expressions import Expression, Select, And, Or, Not, and_, or_, condition
```
tobymao, line 3: why do you need to export And Or

sqlglot/expressions.py
```diff
@@ -1132,7 +1185,7 @@ class Add(Binary):
     pass
 
 
-class And(Binary):
+class And(Binary, Condition):
```
tobymao, line 1188: should
In and Like also be conditions?

barakalon: Oof, yes. I did another pass through to try and tag everything that should be a condition.

sqlglot/__init__.py
barakalon: Good catch. I originally had `and_`, `or_`, etc in this file, but moved 'em down so I can use them within the expression module.

PR #106, opened 2022-04-28
Added a helper method to alias an expression

nolamesa:


sqlglot/expressions.py
```diff
@@ -2012,3 +2014,37 @@ def not_(expression, dialect=None, **opts):
...
+        alias (str or Identifier): the alias name to use. If the name has
+            special charachters it is quoted.
+        dialect (str): the dialect used to parse the input expression.
+        **opts: other options to use to parse the input expressions.
+
+    Returns:
+        Alias: the aliased expression
+    """
+    exp = _maybe_parse(expression, dialect=dialect, parser_opts=opts)
+    if isinstance(alias, Identifier):
```
barakalon, line 2041: `SAFE_IDENTIFIER_RE` feels like parser logic.

I think you can replace these few lines with:

```python
identifier = _maybe_parse(alias, parse_into=Identifier, dialect=dialect, parser_opts=opts)
```

We might also want to add some kind of check in `_maybe_parse`:
```
if isinstance(code_or_expression, Expression):
    if parse_into and not isinstance(code_or_expression, parse_into):
        raise ValueError(...)
    return code_or_expression
```
And add `Identifier` to the map here: https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L261

barakalon: Er, I think you can alias things that aren't identifiers?

e.g.
```SUM(foo) AS bar``

What are we trying to accomplish here? I'm assuming:

```python
select(alias("foo", "bar")).from("baz")
```
?

sqlglot/expressions.py
```diff
@@ -2012,3 +2014,37 @@ def not_(expression, dialect=None, **opts):
...
+        identifier = alias
+    elif isinstance(alias, str):
+        quoted = not re.match(SAFE_IDENTIFIER_RE, alias)
+        identifier = Identifier(this=alias, quoted=quoted)
```
tobymao, line 2045: are you tied to doing safe quotes? pypika doesn't do this, it always quotes

sqlglot/expressions.py
tobymao: you can't parse_into an identifier for example

parse_one("x-1") you want this to be an identifier "x-1", but sqlglot doesn't know you intended it to be a string

sqlglot/expressions.py
```diff
@@ -2012,3 +2014,37 @@ def not_(expression, dialect=None, **opts):
...
+    else:
+        raise ValueError(
```
tobymao, line 2047: yea maybe we don't need this, since it can be Sum(x) as y?

sqlglot/__init__.py
```diff
@@ -1,6 +1,6 @@
 from sqlglot.dialects import Dialect
 from sqlglot.errors import ErrorLevel, UnsupportedError, ParseError, TokenError
-from sqlglot.expressions import Expression, Select, and_, or_, not_, condition
+from sqlglot.expressions import Expression, Select, and_, or_, not_, condition, alias
```
tobymao, line 3: alias_ as alias

sqlglot/expressions.py
nolamesa: it felt a little cleaner than adding quotes everywhere, either way works for me

sqlglot/expressions.py
```diff
@@ -2012,3 +2014,37 @@ def not_(expression, dialect=None, **opts):
         **opts,
     )
     return Not(this=_wrap_operator(this))
+
+
+SAFE_IDENTIFIER_RE = re.compile(r"^[a-zA-Z][\w]*$")
+
+
+def alias(expression, alias, dialect=None, **opts):
```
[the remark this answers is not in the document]
nolamesa: hahah sure that can work fine

sqlglot/expressions.py
nolamesa: that would not make a difference, alias should still be an identifer or a string

tobymao: ah you're right sorry

tobymao (approved):

PR #110, opened 2022-04-29
Ignore empty where clauses when building SQL

nolamesa:


nolamesa:
This allows us to build where statements conditionally.

```python 
where = ""
if foo:
   where = "foo"
...
select('x').from_('tbl').where(where).where(conditions.get('bar'))
```

sqlglot/expressions.py
```diff
@@ -913,6 +913,9 @@ def where(
         Returns:
             Select: the modified expression.
         """
+        expressions = [exp for exp in expressions if exp is not None and exp != ""]
```
barakalon, line 916: I think this should work the same:

```python
expressions = [exp for exp in expressions if exp]
```

It's a little less explicit, but I don't think there are other arguments we're expecting that could evaluate to False?

barakalon: Also, do we want the same thing for `having`? Maybe we should move this into `_apply_conjuction_builder`?

nolamesa: yeah you are right but see my test, I am guessing `where(False)` is valid

tobymao: in his test he allows where(False) which is different, although i wonder if this is correct

tobymao: why would you ever do where False, i think this is tricky, and probably not necessary

nolamesa: it is correct, tried it in postgres

nolamesa: you are right I can move into the conjunction builder, having should do the same

barakalon (approved):

PR #112, opened 2022-04-30
[Sqlglot] Add Support for UDF Registering

robert8138:
Add support for UDF registering. This is useful for sql query that involve UDAFs (e.g. aggregate source in Minerva).


sqlglot/dialects.py
```diff
@@ -695,6 +695,8 @@ def _create_sql(self, e):
 
         if kind.upper() == "TABLE" and temporary is True:
             return f"CREATE TEMPORARY VIEW {self.sql(e, 'this')} AS {self.sql(e, 'expression')}"
+        if kind.upper() == "FUNCTION" and temporary is True:
```
tobymao, line 698: why is this necessary? the reason why the previos line exists is because table != view but function should work?

sqlglot/parser.py
```diff
@@ -456,23 +456,31 @@ def _parse_create(self):
         replace = self._match(TokenType.OR) and self._match(TokenType.REPLACE)
         temporary = self._match(TokenType.TEMPORARY)
 
-        create_token = self._match_set((TokenType.TABLE, TokenType.VIEW)) and self._prev
+        create_token = (
+            self._match_set((TokenType.TABLE, TokenType.VIEW, TokenType.FUNCTION))
```
tobymao, line 460: can you make this a constant?

sqlglot/parser.py
```diff
@@ -456,23 +456,31 @@ def _parse_create(self):
...
         exists = self._parse_exists(not_=True)
-        this = self._parse_table(schema=True)
+        this = None
         expression = None
         properties = None
 
-        if create_token.token_type == TokenType.TABLE:
+        if create_token.token_type == TokenType.FUNCTION:
+            this = self._parse_id_var()
+            if self._match(TokenType.ALIAS):
```
tobymao, line 474: is the AS required?

sqlglot/tokens.py
```diff
@@ -112,6 +112,7 @@ class TokenType(AutoName):
     FILTER = auto()
     FOLLOWING = auto()
     FULL = auto()
+    FUNCTION = auto()
```
tobymao, line 115: is it valid for function to be an alias? if so, add it to ID_VAR_TOKENS

select 1 AS function

select 1 function

select 1 from table function

sqlglot/dialects.py
robert8138: You are right

sqlglot/parser.py
robert8138: Addressed

sqlglot/parser.py
robert8138: Yes, see https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-ddl-create-function.html

sqlglot/tokens.py
robert8138: I don't think so, `FUNCTION` should be reserved keyword in Spark

robert8138: ahh, good point. Added, now it can parse these statements correctly.

```python
In [6]: sqlglot.parse_one("SELECT 1 FROM my_table FUNCTION")
Out[6]: 
(SELECT expressions: 
  (LITERAL this: 1, is_string: False), from: 
  (FROM expressions: 
    (ALIAS this: 
      (TABLE this: 
        (IDENTIFIER this: my_table, quoted: False)), alias: 
      (TABLEALIAS this: 
        (IDENTIFIER this: FUNCTION, quoted: False)))))

In [7]: sqlglot.parse_one("SELECT 1 AS FUNCTION")
Out[7]: 
(SELECT expressions: 
  (ALIAS this: 
    (LITERAL this: 1, is_string: False), alias: 
    (IDENTIFIER this: FUNCTION, quoted: False)))

In [8]: sqlglot.parse_one("SELECT 1 FUNCTION")
Out[8]: 
(SELECT expressions: 
  (ALIAS this: 
    (LITERAL this: 1, is_string: False), alias: 
    (IDENTIFIER this: FUNCTION, quoted: False)))
```

tobymao (approved):

sqlglot/tokens.py
lbthuir: could you please give a demo showing how to use it? 
I am having problems  with parsing udf functions.  Thanks a lot~

PR #115, opened 2022-05-05
Add dateint support

eakmanrq:
Prior to this change sqlglot assumes that dates are either in date string (DS) format ('YYYY-MM-DD') or a timestamp (TS) that start with the DS format. This adds functions that allows users to say that the date could also be in dateint (DI) format (YYYYMMDD). 

This will generate new SQL in cases where the code previously assumed the date was either TS or DS format since it must now also check if it is DI.

### DIFF
--- a/sqlglot/dialects.py
+++ b/sqlglot/dialects.py
@@ -213,29 +213,40 @@ class DuckDB(Dialect):
     DATE_FORMAT = "'%Y-%m-%d'"
     TIME_FORMAT = "'%Y-%m-%d %H:%M:%S'"
 
-    TS_OR_DS_OR_DI_TO_YEAR = "SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 1, 4)"
-    TS_OR_DS_OR_DI_TO_MONTH = "SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 5, 2)"
-    TS_OR_DS_OR_DI_TO_DAY = "SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 7, 2)"
-    TS_OR_DS_OR_DI_TO_DATE_EXPRESSION = (
-        f"CAST({TS_OR_DS_OR_DI_TO_YEAR} || '-' || "
-        f"{TS_OR_DS_OR_DI_TO_MONTH} || '-' || "
-        f"{TS_OR_DS_OR_DI_TO_DAY} AS DATE)"
+    MIXED_TYPE_TO_YEAR = "SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 1, 4)"
+    MIXED_TYPE_TO_MONTH = "SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 5, 2)"
+    MIXED_TYPE_TO_DAY = "SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 7, 2)"
+    MIXED_TYPE_TO_DATE_EXPRESSION = (
+        f"CAST({MIXED_TYPE_TO_YEAR} || '-' || "
+        f"{MIXED_TYPE_TO_MONTH} || '-' || "
+        f"{MIXED_TYPE_TO_DAY} AS DATE)"
     )
-    TS_OR_DS_OR_DI_TO_DATE_STR_EXPRESSION = (
-        f"STRFTIME({TS_OR_DS_OR_DI_TO_DATE_EXPRESSION}, {DATE_FORMAT})"
+    MIXED_TYPE_TO_DATE_STR_EXPRESSION = (
+        f"STRFTIME({MIXED_TYPE_TO_DATE_EXPRESSION}, {DATE_FORMAT})"
     )
-    TS_OR_DS_OR_DI_TO_DI_EXPRESSION = (
+    MIXED_TYPE_TO_DI_EXPRESSION = (
         "CAST(SUBSTR(REPLACE(CAST({this} AS VARCHAR), '-', ''), 1, 8) AS INT)"
     )
 
-    def _unix_to_time(self, expression):
-        return f"TO_TIMESTAMP(CAST({self.sql(expression, 'this')} AS BIGINT))"
+    @classmethod
+    def _mixed_type_add_to_di(cls, generator, expression):
+        this = generator.sql(expression, "this")
+        e = generator.sql(expression, "expression")
+        unit = expression.text("unit") or "DAY"
 
-    def _ts_or_ds_add(self, expression):
-        this = self.sql(expression, "this")
-        e = self.sql(expression, "expression")
+        date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(this=this)
+
+        return f"CAST(strftime({date} + INTERVAL {e} {unit}, {DuckDB.DATEINT_FORMAT}) AS INT)"
+
+    @classmethod
+    def _mixed_type_add_to_ds(cls, generator, expression):
+        this = generator.sql(expression, "this")
+        e = generator.sql(expression, "expression")
         unit = expression.text("unit") or "DAY"
         return f"STRFTIME(CAST({this} AS DATE) + INTERVAL {e} {unit}, {DuckDB.DATE_FORMAT})"
+
+    def _unix_to_time(self, expression):
+        return f"TO_TIMESTAMP(CAST({self.sql(expression, 'this')} AS BIGINT))"
 
     def _date_add(self, expression):
         this = self.sql(expression, "this")
@@ -243,28 +254,19 @@ class DuckDB(Dialect):
         unit = expression.text("unit") or "DAY"
         return f"{this} + INTERVAL {e} {unit}"
 
-    def _ts_or_ds_or_di_to_date_str_sql(self, expression):
+    def _mixed_type_to_date_str_sql(self, expression):
         this = self.sql(expression, "this")
-        return DuckDB.TS_OR_DS_OR_DI_TO_DATE_STR_EXPRESSION.format(this=this)
+        return DuckDB.MIXED_TYPE_TO_DATE_STR_EXPRESSION.format(this=this)
 
-    def _ts_or_ds_or_di_to_di_sql(self, expression):
+    def _mixed_type_to_di_sql(self, expression):
         this = self.sql(expression, "this")
-        return DuckDB.TS_OR_DS_OR_DI_TO_DI_EXPRESSION.format(this=this)
+        return DuckDB.MIXED_TYPE_TO_DI_EXPRESSION.format(this=this)
 
-    def _ts_or_ds_or_di_to_date_sql(self, expression):
+    def _mixed_type_to_date_sql(self, expression):
         this = self.sql(expression, "this")
-        return DuckDB.TS_OR_DS_OR_DI_TO_DATE_EXPRESSION.format(this=this)
+        return DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(this=this)
 
-    def _di_add_sql(self, expression):
-        this = self.sql(expression, "this")
-        e = self.sql(expression, "expression")
-        unit = expression.text("unit") or "DAY"
-
-        date = DuckDB.TS_OR_DS_OR_DI_TO_DATE_EXPRESSION.format(this=this)
-
-        return f"CAST(strftime({date} + INTERVAL {e} {unit}, {DuckDB.DATEINT_FORMAT}) AS INT)"
-
-    def _ts_or_ds_or_di_date_diff_sql(self, expression):
+    def _mixed_type_date_diff_sql(self, expression):
         third_arg = self.sql(expression, "second_date")
         if third_arg:
             first_date_value = expression.text("first_date")
@@ -275,14 +277,21 @@ class DuckDB(Dialect):
             second_date_value = expression.text("first_date")
             unit = "DAY"
 
-        first_date = DuckDB.TS_OR_DS_OR_DI_TO_DATE_EXPRESSION.format(
+        first_date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(
             this=first_date_value
         )
-        second_date = DuckDB.TS_OR_DS_OR_DI_TO_DATE_EXPRESSION.format(
+        second_date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(
             this=second_date_value
         )
 
         return f"DATE_DIFF('{unit}', {first_date}, {second_date})"
+
+    def _mixed_type_add_sql(self, expression):
+        calling_key = expression.key.upper()
+        output_format = expression.text("output_format").upper() or "YYYY-MM-DD"
+        if output_format == "YYYYMMDD" or calling_key == "DIADD":
+            return DuckDB._mixed_type_add_to_di(self, expression)
+        return DuckDB._mixed_type_add_to_ds(self, expression)
 
     transforms = {
         exp.ApproxDistinct: _approx_count_distinct_sql,
@@ -290,8 +299,13 @@ class DuckDB(Dialect):
         exp.DateAdd: _date_add,
         exp.DateDiff: lambda self, e: f"{self.sql(e, 'this')} - {self.sql(e, 'expression')}",
         exp.DateStrToDate: lambda self, e: f"CAST({self.sql(e, 'this')} AS DATE)",
-        exp.DiAdd: _di_add_sql,
+        exp.DiAdd: _mixed_type_add_sql,
         exp.Explode: lambda self, e: f"UNNEST({self.sql(e, 'this')})",
+        exp.MixedTypeAdd: _mixed_type_add_sql,
+        exp.MixedTypeDateDiff: _mixed_type_date_diff_sql,
+        exp.MixedTypeToDate: _mixed_type_to_date_sql,
+        exp.MixedTypeToDateStr: _mixed_type_to_date_str_sql,
+        exp.MixedTypeToDi: _mixed_type_to_di_sql,
         exp.Quantile: lambda self, e: f"QUANTILE({self.sql(e, 'this')}, {self.sql(e, 'quantile')})",
         exp.RegexpLike: lambda self, e: f"REGEXP_MATCHES({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
         exp.RegexpSplit: lambda self, e: f"STR_SPLIT_REGEX({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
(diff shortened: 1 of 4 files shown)


sqlglot/dialects.py
```diff
@@ -221,8 +221,12 @@ class DuckDB(Dialect):
         f"{TS_OR_DS_OR_DI_TO_MONTH} || '-' || "
         f"{TS_OR_DS_OR_DI_TO_DAY} as date)"
     )
-    TS_OR_DS_OR_DI_TO_DATE_STR_EXPRESSION = f"strftime({TS_OR_DS_OR_DI_TO_DATE_EXPRESSION}, {DATE_FORMAT})"
-    TS_OR_DS_OR_DI_TO_DI_EXPRESSION = "CAST(SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 1, 8) as int)"
+    TS_OR_DS_OR_DI_TO_DATE_STR_EXPRESSION = (
+        f"strftime({TS_OR_DS_OR_DI_TO_DATE_EXPRESSION}, {DATE_FORMAT})"
+    )
+    TS_OR_DS_OR_DI_TO_DI_EXPRESSION = (
+        "CAST(SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 1, 8) as int)"
```
tobymao, line 228: AS INT

sqlglot/dialects.py
```diff
@@ -665,9 +683,15 @@ class Presto(Dialect):
     index_offset = 1
     TIME_FORMAT = "'%Y-%m-%d %H:%i:%S'"
 
-    TS_OR_DS_OR_DI_TO_DATE_EXPRESSION = "DATE_PARSE(SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 1, 8), '%Y%m%d')"
-    TS_OR_DS_OR_DI_TO_DATE_STR_EXPRESSION = f"DATE_FORMAT({TS_OR_DS_OR_DI_TO_DATE_EXPRESSION}, '%Y-%m-%d')"
-    TS_OR_DS_OR_DI_TO_DI_EXPRESSION = f"CAST(DATE_FORMAT({TS_OR_DS_OR_DI_TO_DATE_EXPRESSION}, '%Y%m%d') as int)"
+    TS_OR_DS_OR_DI_TO_DATE_EXPRESSION = (
+        "DATE_PARSE(SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 1, 8), '%Y%m%d')"
```
tobymao, line 687: AS VARCHAR

sqlglot/dialects.py
```diff
@@ -209,9 +209,25 @@ def _format_time(exp_class, dialect, default=None):
...
     DATE_FORMAT = "'%Y-%m-%d'"
     TIME_FORMAT = "'%Y-%m-%d %H:%M:%S'"
 
+    TS_OR_DS_OR_DI_TO_YEAR = "SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 1, 4)"
+    TS_OR_DS_OR_DI_TO_MONTH = "SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 5, 2)"
+    TS_OR_DS_OR_DI_TO_DAY = "SUBSTR(REPLACE(CAST({this} as varchar), '-', ''), 7, 2)"
+    TS_OR_DS_OR_DI_TO_DATE_EXPRESSION = (
+        f"CAST({TS_OR_DS_OR_DI_TO_YEAR} || '-' || "
+        f"{TS_OR_DS_OR_DI_TO_MONTH} || '-' || "
+        f"{TS_OR_DS_OR_DI_TO_DAY} as date)"
```
tobymao, line 222: sorry nit, but AS DATE

sqlglot/dialects.py
```diff
@@ -411,6 +484,62 @@ def _time_to_unix(self, expression):
...
+            raise NotImplementedError(f"Unit not implemented for Spark Di Add: {unit}")
+
+        return f"CAST({add_expression} as int)"
+
+    def _ts_or_ds_or_di_date_diff_sql(self, expression):
+        third_arg = self.sql(expression, "second_date")
+        if third_arg:
+            first_date_value = self.sql(expression, "first_date")
+            second_date_value = third_arg
+            unit = self.sql(expression, "unit")
```
tobymao, line 522: this will be easier if you do expression.text("unit") then you don't hae to check the quotes in the string

sqlglot/dialects.py
```diff
@@ -411,6 +484,62 @@ def _time_to_unix(self, expression):
...
+            this=second_date_value
+        )
+
+        if unit == "'DAY'":
+            return f"DATEDIFF({first_date}, {second_date})"
+        if unit == "'MONTH'":
+            return f"MONTHS_BETWEEN({first_date}, {second_date})"
+        if unit == "'YEAR'":
+            return f"ROUND(MONTHS_BETWEEN({first_date}, {second_date}) / 12, 2)"
+        raise NotImplementedError(f"Unit not implemented for Spark Date Diff: {unit}")
```
tobymao, line 541: there's somethnig for this self.unsupported_error

eakmanrq: Hmm self is a generator and does not have a `unsupported_error`. Error message:
```
    raise self.unsupported_error(f"Unit not implemented for Spark Di Add: {unit}")
AttributeError: 'Generator' object has no attribute 'unsupported_error'
```

tobymao:
```def _approx_count_distinct_sql(self, expression):
    if expression.args.get("accuracy"):
        self.unsupported("APPROX_COUNT_DISTINCT does not support accuracy")
    return f"APPROX_COUNT_DISTINCT({self.sql(expression, 'this')})"
```

tobymao:
```    def unsupported(self, message):
        self.unsupported_messages.append(message)
```

my bad, it's called unsupported

sqlglot/expressions.py
```diff
@@ -1589,6 +1597,31 @@ class Min(AggFunc):
     pass
 
 
+class MixedTypeAdd(Func):
+    arg_types = {
+        "this": True,
+        "expression": True,
+        "unit": False,
+        "output_format": False,
```
tobymao, line 1605: call this format for consistency?

sqlglot/expressions.py
```diff
@@ -1589,6 +1597,31 @@ class Min(AggFunc):
...
+    }
+
+
+class MixedTypeDateDiff(Func):
+    arg_types = {"unit": True, "first_date": True, "second_date": False}
```
tobymao, line 1610: is date_diff first_date - second_date?

sqlglot/dialects.py
```diff
@@ -209,31 +209,101 @@ def _format_time(exp_class, dialect, default=None):
...
-        unit = self.sql(expression, "unit").strip("'") or "DAY"
+    @classmethod
+    def _mixed_type_add_to_di(cls, generator, expression):
+        this = generator.sql(expression, "this")
+        e = generator.sql(expression, "expression")
+        unit = expression.text("unit") or "DAY"
+
+        date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(this=this)
+
+        return f"CAST(strftime({date} + INTERVAL {e} {unit}, {DuckDB.DATEINT_FORMAT}) AS INT)"
```
tobymao, line 239: STRFTIME

sqlglot/dialects.py
```diff
@@ -209,31 +209,101 @@ def _format_time(exp_class, dialect, default=None):
...
+        second_date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(
+            this=second_date_value
+        )
+
+        return f"DATE_DIFF('{unit}', {first_date}, {second_date})"
+
+    def _mixed_type_add_sql(self, expression):
+        calling_key = expression.key.upper()
+        output_format = expression.text("output_format").upper() or "YYYY-MM-DD"
+        if output_format == "YYYYMMDD" or calling_key == "DIADD":
```
tobymao, line 290: i don't really use key, can you switch to isinstance(expression, exp.DiAdd)

sqlglot/dialects.py
```diff
@@ -322,9 +392,66 @@ class Hive(Dialect):
...
+            add_expression = f"ADD_MONTHS({date}, {e} * 12)"
+        else:
+            generator.unsupported(f"Unit not implemented for Spark Di Add: {unit}")
+
+        return add_expression
+
+    def _mixed_type_add_sql(self, expression):
+        calling_key = expression.key.upper()
+        output_format = expression.text("output_format").upper() or "YYYY-MM-DD"
+        if output_format == "YYYYMMDD" or calling_key == "DIADD":
```
tobymao, line 451: same

sqlglot/expressions.py
eakmanrq: I actually explicit added the `output_` prefix since I found it confusing when it was just `format`. The reason is that with just format it is not clear if you are providing the format for the provided date or the format of the output. After considering that if you still want `format` then I am happy to change.

tests/test_dialects.py
```diff
@@ -833,19 +1038,22 @@ def test_hive(self):
...
             "DATE_SUB('2020-01-01', 1)",
-            "DATE_FORMAT(DATE_ADD('DAY', 1 * -1, DATE_PARSE(SUBSTR('2020-01-01', 1, 10), '%Y-%m-%d')), '%Y-%m-%d')",
+            "DATE_ADD('DAY', 1 * -1, '2020-01-01')",
             read="hive",
             write="presto",
         )
         self.validate(
             "DATE_ADD('2020-01-01', 1)",
-            "DATE_FORMAT(DATE_ADD('DAY', 1, DATE_PARSE(SUBSTR('2020-01-01', 1, 10), '%Y-%m-%d')), '%Y-%m-%d')",
+            "DATE_ADD('DAY', 1, '2020-01-01')",
```
tobymao, line 1047: is this test correct? have you tested this works? in presto don't you need an actual date type not a varchar

sqlglot/expressions.py
tobymao: ok that's fine

sqlglot/expressions.py
eakmanrq: It looks like this is inconsistent across vendors. Duckdb does `second_date` - `first_date` and then hive/presto do `first_date` - `second_date`. I will update this to explicitly state if it is start or end date and then have them all behave in the same way.

sqlglot/dialects.py
```diff
@@ -461,16 +630,16 @@ def _unix_to_time(self, expression):
 Hive.functions = {
     "APPROX_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
     "COLLECT_LIST": exp.ArrayAgg.from_arg_list,
-    "DATE_ADD": lambda args: exp.TsOrDsAdd(
```
tobymao, line 464: this is fishy

sqlglot/expressions.py
tobymao: great

tests/test_dialects.py
eakmanrq: You are right this doesn't work. I will review this. I originally ran all the conversions manually but that was before the mixed type refactor so I need to do it again.

sqlglot/dialects.py
eakmanrq: `DATE_ADD` is really confusing. For example Hive accepts a date str and returns a date string. Presto expects a date or timestamp and returns a type that matches the input. Duckdb technically doesn't have a DATE_ADD but you can use interval +/- and that behaves like presto. Therefore I start switching the implementations to not change the default behavior of `DATE_ADD` but instead have it use the `DATE_ADD` for that vendor directly. In my mind if a user wants a cross-vendor functionality they should use `MixedTypeAdd` and define an output format in order to be flexible on the input and explicit on expected output.

tobymao: i don't think that approach is safe. users write hive sql and expect it to work in presto. thus the intention of DATE_ADD is given a datestr add it. so that's why in the original code, when hive has date_add it really means TSorDSADD. 

i don't believe we should change this behavior as it's not an optimal experience. the goal is for users to write once and run anywhere. it is indeed tricky because hive sql does a lot of implicit stuff.

tobymao: If a user wants to do cross-vendor functionality, they can writing sqlglot syntax MixedTypeAdd, but it should not be required to get transpilation

tobymao: there is no date object in hive and so this works doesn’t it? perhaps the old code was bugged and in this case hive should convert to string first?

eakmanrq: Hive does have a date type: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=82706456#LanguageManualTypes-date

So I think the ideal way to handle this would be to convert presto's `DATE_ADD` to a `DATE + INTERVAL` like Duckdb does. Hive's date_add would convert to `MixedTypeAdd` in presto with an output format of `YYYY-MM-DD`. I need to look at this a bit more to see how feasible this is though.

sqlglot/dialects.py
```diff
@@ -549,6 +718,42 @@ class Presto(Dialect):
...
+        this = generator.sql(expression, "this")
+        e = generator.sql(expression, "expression")
+        unit = expression.text("unit") or "DAY"
+
+        date = Presto.MIXED_TYPE_TO_DATE_EXPRESSION.format(this=this)
+
+        return f"DATE_FORMAT(DATE_ADD('{unit}', {e}, {date}), '%Y-%m-%d')"
+
+    def _mixed_type_add_sql(self, expression):
+        output_format = expression.text("output_format").upper() or "YYYY-MM-DD"
```
tobymao, line 752: are you sure presto is YYYY-MM-DD? shouldn't it be %Y-%m?

sqlglot/dialects.py
```diff
@@ -209,31 +209,98 @@ def _format_time(exp_class, dialect, default=None):
...
+            end_date_value = self.sql(expression, "start_date")
+            unit = "DAY"
+
+        first_date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(this=start_date_value)
+        second_date = DuckDB.MIXED_TYPE_TO_DATE_EXPRESSION.format(this=end_date_value)
+
+        return f"DATE_DIFF('{unit}', {first_date}, {second_date})"
+
+    def _mixed_type_add_sql(self, expression):
+        output_format = expression.text("output_format").upper() or "YYYY-MM-DD"
```
tobymao, line 286: is this correct ? YYYY?

tests/test_dialects.py
```diff
@@ -193,7 +204,11 @@ def test_duckdb(self):
         )
         self.validate(
             "TS_OR_DS_TO_DATE(x)",
-            "CAST(x AS DATE)",
```
tobymao, line 196: sorry for being so picky, but it looks like doing mixed date is pretty inefficient compared to just do ts_or_ds_to_date, i wonder if it makes sense for ts_or_ds_to_date to maintain the old behavior and to have the addition of mixed_date_type with this mixed type behavior.

big difference between cast(x as date) vs this crazy 3x casting string concat.

tests/test_dialects.py
```diff
@@ -991,6 +1217,77 @@ def test_hive(self):
...
+        self.validate(
+            "MIXED_TYPE_TO_DATE_STR(x)",
+            "DATE_FORMAT(TO_DATE(SUBSTR(REPLACE(CAST(x as string), '-', ''), 1, 8), 'yyyyMMdd'), 'yyyy-MM-dd')",
+            write="hive",
+            identity=False,
+        )
+
+        self.validate(
+            "MIXED_TYPE_TO_DATE(x)",
+            "TO_DATE(SUBSTR(REPLACE(CAST(x as string), '-', ''), 1, 8), 'yyyyMMdd')",
```
tobymao, line 1229: is this correct? shouldn't mixed_type_to_date(x) in hive just be TO_DATE(x)

tests/test_dialects.py
```diff
@@ -991,6 +1217,77 @@ def test_hive(self):
...
+            write="hive",
+            identity=False,
+        )
+
+        self.validate(
+            "MIXED_TYPE_TO_DI(x)",
```
tobymao, line 1235: should it be CAST(TO_DATE(x) AS INT)

tobymao:
sorry for being picky, but this is a pretty big change, after we finally iron it out, i'm going to test it on some live data to make sure nothing is borked

tests/test_dialects.py
```diff
@@ -693,6 +815,89 @@ def test_presto(self):
...
+            (
+                "CAST(DATE_FORMAT(DATE_ADD('DAY', 1, DATE_PARSE(SUBSTR(REPLACE(CAST(20220101 "
+                "AS VARCHAR), '-', ''), 1, 8), '%Y%m%d')), '%Y%m%d') AS INT)"
+            ),
+            write="presto",
+            identity=False,
+        )
+
+        self.validate(
+            "MIXED_TYPE_ADD(20220101, 1, 'DAY', 'YYYY-MM-DD')",
```
tobymao, line 882: this test doesn't seem right to me, the generic format time string should be python style so 

YYYY -> %Y

so i think you assumed the hive is the generic format, but sqlglot uses python time str as format

sqlglot/dialects.py
```diff
@@ -209,31 +209,98 @@ def _format_time(exp_class, dialect, default=None):
...
+        if output_format == "YYYYMMDD" or isinstance(expression, exp.DiAdd):
```
tobymao, line 287: hmm, yyyymmdd can be a string "20000101", it doesn't necessitate date int right??

sqlglot/dialects.py
```diff
@@ -322,9 +389,65 @@ class Hive(Dialect):
...
+        elif unit == "YEAR":
+            add_expression = f"ADD_MONTHS({date}, {e} * 12)"
+        else:
+            generator.unsupported(f"Unit not implemented for Spark Di Add: {unit}")
+
+        return add_expression
+
+    def _mixed_type_add_sql(self, expression):
+        output_format = expression.text("output_format").upper() or "YYYY-MM-DD"
+        if output_format == "YYYYMMDD" or isinstance(expression, exp.DiAdd):
```
tobymao, line 447: yea YYYYMMDD is not really a substitude for diadd

sqlglot/dialects.py
```diff
@@ -461,25 +628,27 @@ def _unix_to_time(self, expression):
...
     ),
-    "DATE_SUB": lambda args: exp.TsOrDsAdd(
+    "DATE_SUB": lambda args: exp.MixedTypeAdd(
         this=list_get(args, 0),
         expression=exp.Mul(
             this=list_get(args, 1),
             expression=exp.Literal.number(-1),
         ),
         unit=exp.Literal.string("DAY"),
+        output_format=exp.Literal.string("YYYY-MM-DD"),
```
tobymao, line 648: i wonder if you need to be calling format_time in many places in this PR

format_time allows us to take in an input format string like hive "YYYY-MM" and convert it to presto "%Y-%M"

sqlglot/dialects.py
```diff
@@ -461,25 +628,27 @@ def _unix_to_time(self, expression):
 Hive.functions = {
     "APPROX_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
     "COLLECT_LIST": exp.ArrayAgg.from_arg_list,
-    "DATE_ADD": lambda args: exp.TsOrDsAdd(
+    "DATE_ADD": lambda args: exp.MixedTypeAdd(
         this=list_get(args, 0),
         expression=list_get(args, 1),
         unit=exp.Literal.string("DAY"),
+        output_format=exp.Literal.string("YYYY-MM-DD"),
```
tobymao, line 635: so in this case, it should probably be Hive.format_time("YYYY-MM-DD") or something like that so that the output_format is in the generic form

tests/test_dialects.py
eakmanrq: Hive does not support int as a type for the `TO_DATE` function. So we need to first get it into a string in the format that we are providing.

tests/test_dialects.py
eakmanrq: You can't cast a date to an int. So you first need to cast it to a string in dateint format and then to an int.

PR #116, opened 2022-05-05
Properly catch parse error when no statement follows CTE

barakalon:
Parsing was raising an `AttributeError: 'NoneType' object has no attribute 'arg_types'` error is _parse_statement returned `None`.

This raises a proper ParseError instead.

@tobymao

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -719,6 +719,10 @@ class Parser:
 
         this = self._parse_statement()
 
+        if this is None:
+            self.raise_error("Failed to parse any statement following CTE")
+            return
+
         if "with" not in this.arg_types:
             self.raise_error(f"{this.key} does not support CTE")
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -721,7 +721,7 @@ class Parser:
 
         if this is None:
             self.raise_error("Failed to parse any statement following CTE")
-            return
+            return None
 
         if "with" not in this.arg_types:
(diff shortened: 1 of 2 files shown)


tests/test_parser.py
```diff
@@ -74,13 +74,16 @@ def test_expression(self):
...
-    def test_function_arguments_validation(self):
+    def test_parse_errors(self):
         with self.assertRaises(ParseError):
             parse_one("IF(a > 0, a, b, c)")
 
         with self.assertRaises(ParseError):
             parse_one("IF(a > 0)")
 
+        with self.assertRaises(ParseError):
+            parse_one("WITH cte AS (SELECT * FROM x)")
```
tobymao, line 85: just double checking, we can still do a partial parse?

tobymao (approved):

sqlglot/parser.py
```diff
@@ -719,6 +719,10 @@ def _parse_with(self):
 
         this = self._parse_statement()
 
+        if this is None:
+            self.raise_error("Failed to parse any statement following CTE")
+            return
```
tobymao, line 724: ah it should be return this

tobymao: equivalent but lint based

tests/test_parser.py
barakalon: Well, there is not CTE expression type anymore, so we'd have to arbitrarily pick an expression type to assume

Issue #117, opened 2022-05-06
Parsing error in CTAS statement with parenthesis and with clause


youngsol:
Following sql works in presto but it produces parse error
```
   CREATE TABLE  schema1.table1 AS
   (
        WITH base AS (
            SELECT *
                    ,row_number() over(partition by col1 order by col2 asc) AS test
            FROM test.table2
        )
        SELECT * FROM base
    )
```
ERROR:
```
sqlglot.errors.ParseError: Expecting ). Line 4, Col: 9.
  
   CREATE TABLE  schema1.table1 AS
   (
        WITH base AS (
            SELECT *
                    ,row_number() over(partition by col1 order by co
```

if you remove parenthesis, it works.
```
   CREATE TABLE  schema1.table1 AS
   --(
        WITH base AS (
            SELECT *
                    ,row_number() over(partition by col1 order by col2 asc) AS test
            FROM test.table2
        )
        SELECT * FROM base
    --)
```

barakalon:
Possible fix: https://github.com/tobymao/sqlglot/pull/118

PR #118, opened 2022-05-06
CTAS with paranthesized expression

barakalon:
Addresses https://github.com/tobymao/sqlglot/issues/117

@tobymao - take a critical eye at this. The WITH/SELECT recursion is getting kind of messy, and I don't have a good mental catalog of what's valid in various dialects.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -480,7 +480,14 @@ class Parser:
                 this if isinstance(this, exp.Schema) else None
             )
             if self._match(TokenType.ALIAS):
+                parenthesized = self._match(TokenType.L_PAREN)
                 expression = self._parse_with()
+                if parenthesized:
+                    self._match(TokenType.R_PAREN)
+                    expression = self.expression(
+                        exp.Subquery,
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -480,7 +480,14 @@ def _parse_create(self):
                 this if isinstance(this, exp.Schema) else None
             )
             if self._match(TokenType.ALIAS):
+                parenthesized = self._match(TokenType.L_PAREN)
```
tobymao, line 483: i don't think this is correct, it doesn't handle recursive parens

tobymao: i have a fix

tobymao:
@barakalon https://github.com/tobymao/sqlglot/commit/9e4b66cdcb553e1ecaca3f20a269298d8b1c7b33

the issue was that on paren, we look for a select instead of a with, with is a superset of select, so it's safe

barakalon:
Sweet. I had tried that out, but the extra recursion felt scary to me. Glad you took care of it :)

PR #119, opened 2022-05-06
add support for partial with

tobymao:
figured it out @barakalon


barakalon (approved):
Ah, I forgot about the `With` expression. Nice!

PR #126, opened 2022-05-16
Support more cases for join builder:

nolamesa:
* join('table')
* join('select x from table')
* join('x on a = b')
* join('join x on a = b')
* join (Join|Select|Table|Subquery)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2,14 +2,11 @@ from copy import deepcopy
 from collections import deque
 from enum import auto
 import re
-import typing
 import inspect
 import sys
 
 from sqlglot.errors import ParseError
 from sqlglot.helper import AutoName, camel_to_snake_case, ensure_list
-
-T = typing.TypeVar("T")
 
 
 class Expression:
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -279,18 +279,21 @@ def parse_into(self, expression_type, raw_tokens, sql=None):
...
-        if not expressions:
-            raise ValueError(f"Failed to parse into {expression_type}")
-
-        return expressions
+        error = None
+        for expression_type in ensure_list(expression_types):
+            if expression_type not in methods:
+                raise TypeError(f"No parser registered for {expression_type}")
+
+            method = methods[expression_type]
```
tobymao, line 291: you can just put this inline on 293

tobymao (approved):

sqlglot/expressions.py
```diff
@@ -6,6 +6,7 @@
 import inspect
 import sys
 
+from sqlglot.errors import ParseError
 from sqlglot.helper import AutoName, camel_to_snake_case, ensure_list
 
 T = typing.TypeVar("T")
```
tobymao, line 12: delete this?

tobymao: and the typing

barakalon (approved):

PR #129, opened 2022-05-20
first pass of optimizer

tobymao:
initial optimizer rule, to rewrite sql into canonical form

useful for parsing arbitrary queries and physical plans

### DIFF
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -62,15 +62,6 @@ def apply_index_offset(expressions, offset):
     return expressions
 
 
-def parent(expression, kind):
-    expression = expression.parent
-    while not isinstance(expression, kind):
-        expression = expression.parent
-        if not expression:
-            raise RuntimeError(f"Could not find parent {kind}")
-    return expression
-
-
 CAMEL_CASE_PATTERN = re.compile("(?<!^)(?=[A-Z])")
 
 
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -62,6 +62,15 @@ def apply_index_offset(expressions, offset):
     return expressions
 
 
+def parent(expression, kind):
+    expression = expression.parent
+    while not isinstance(expression, kind):
+        expression = expression.parent
+        if not expression:
+            raise RuntimeError(f"Could not find parent {kind}")
+    return expression
+
+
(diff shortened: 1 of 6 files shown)


sqlglot/helper.py
```diff
@@ -62,6 +62,15 @@ def apply_index_offset(expressions, offset):
     return expressions
 
 
+def parent(expression, kind):
```
nolamesa, line 65: This is not currently used anywhere right? do you need it? If so perhaps rename to find_ancestor. Also minor but I would change the local variable below from expression to ancestor and return None when not found

sqlglot/optimizer.py
```diff
@@ -0,0 +1,27 @@
...
+        return node
+
+    return expression.transform(qualify)
+
+
+def qualify_columns(expression, _schema):
+    return expression
+
+
+def optimize(expression, schema=None, db=None, catalog=None):
```
nolamesa, line 21: I am not entirely clear how qualifying tables is an optimization

nolamesa (approved):

sqlglot/optimizer.py
barakalon: I think this is a _pre_ optimization step.

sqlglot/optimizer.py
```diff
@@ -0,0 +1,27 @@
...
+            if not node.args.get("catalog"):
+                node.set("catalog", catalog)
+            return node
+        return node
+
+    return expression.transform(qualify)
+
+
+def qualify_columns(expression, _schema):
+    return expression
```
barakalon, line 18: What were you imaging here?
I was thinking DFS in postorder, disambiguating columns as we go up the tree?

barakalon (approved):
Sweet!

I think this canonical SQL will be helpful for the Minerva SQL proxy. But that proxy doesn't have to actually generate/optimize a logical plan.

For a full optimizer, I think we'll need to convert an expression to a logical plan?

sqlglot/optimizer.py
tobymao: right, there are a series of steps needed to take every query into canonical form, this is the first rule

nolamesa (approved):

PR #130, opened 2022-05-20
feat: Expression.replace

barakalon:
This allows expressions within an AST to be surgically replaced.

Kind of an ugly operation.

@tobymao


barakalon:
Maybe this should be a separate function?

```python
def replace(node, new_node):
    ...
```

Kind of weird that the method doesn't really operate on the instance itself...

tobymao (approved):

PR #132, opened 2022-05-20
Added a named selects property on selects

nolamesa:


tobymao (approved):

sqlglot/expressions.py
```diff
@@ -1133,6 +1133,10 @@ def ctas(self, table, properties=None, dialect=None, parser_opts=None, copy=True
             ),
         )
 
+    @property
+    def named_selects(self):
```
barakalon, line 1137: Is it strange for this to be on the base Expression instance? Should this be on Select?

barakalon (approved):
Small comment about where to put the method.

Otherwise, sweet! I need this immediately haha

sqlglot/expressions.py
tobymao: oh, i didn't catch that. yea that doesn't make sense

tobymao: what about subqueryable?

eakmanrq (approved):

sqlglot/expressions.py
nolamesa: it's on select class

barakalon: doh

barakalon (approved):

tobymao (approved):

PR #133, opened 2022-05-21
feat(optimizer): qualify columns

barakalon:
This is an initial swing at the column qualifier.

It's relatively complex, so I wanted to get some early feedback.

If we agree on the direction, just have to:
1. Add some more tests
2. Cleanup a bit

@tobymao @nolamesa


tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -0,0 +1,124 @@
...
+-- Derived tables
+--------------------------------------
+SELECT y.a AS a FROM (SELECT x.a AS a FROM x) AS y(a);
+SELECT y.a AS a FROM (SELECT x.a AS a FROM x) AS y(a);
+
+SELECT y.a FROM (SELECT a AS a FROM x) AS y(a);
+SELECT y.a AS a FROM (SELECT x.a AS a FROM x) AS y(a);
+
+SELECT y.a FROM (SELECT x.a AS a FROM x) AS y;
+SELECT y.a AS a FROM (SELECT x.a AS a FROM x) AS y(a);
```
tobymao, line 32: why is it as y(a) ? this doesn’t seem to be standard syntax is it? dan we just leave it AS y

barakalon: I _think_ the alias column list is standard... 
Here it is in MySQL: https://dev.mysql.com/doc/refman/8.0/en/join.html
Here it is in Presto: https://prestodb.io/docs/current/sql/select.html

It can potentially help with some later steps. Take this for example:

```sql
SELECT 
    *
FROM (
    SELECT a, b FROM x 
    UNION 
    SELECT a, b FROM y) AS z(a, b)
```

The alias column list is sort of like a cache of the output columns...
```python
subquery = expression.find(Subquery)
name = subquery.args["alias"].text("this")
outputs = subquery.args["alias"].args["columns"]

outputs_to_keep = outputs.intersection(columns_by_selectable[name])
projection_pushdown(subquery, outputs_to_keep)
```

Now, this is probably the equivalent of the "named_selections" method @nolamesa added. But something feels nice and clean about making the AST more explicit... it's nice to have a predictable structure. 🤷  I could go either way.

sqlglot/optimizer/optimize.py
```diff
@@ -0,0 +1,11 @@
+from sqlglot.optimizer.qualify_tables import qualify_tables
+from sqlglot.optimizer.qualify_columns import qualify_columns
+
+
+def optimize(expression, schema=None, db=None, catalog=None):
```
tobymao, line 5: can we put this in __init__ so that we can do from sqlglot.optimizer import optimize

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+        schema (dict): Mapping of table names to all available columns
+    Returns:
+        sqlglot.Expression: qualified expression
+    """
+    expression = expression.copy()
+    _qualify_union(expression, schema, Sequence(), {})
+    return expression
+
+
+class Sequence:
```
tobymao, line 26: maybe this should go in helpers

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def __init__(self):
+        self.i = -1
+
+    def next(self):
+        self.i += 1
+        return self.i
+
+
+def _qualify_union(expression, schema, sequence, parent_selectables):
```
tobymao, line 48: the name is a bit misleading cuz it's union or select

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+
+    # Collect all the other relevant nodes in this context
+    # Only traverse down to the next SELECT statements - we'll recurse into those later
+    for node, parent, _ in expression.walk(stop_types=exp.Select):
+        if node is expression:
+            continue  # Skip this node itself - we only care about children
+        if isinstance(node, exp.CTE):
+            ctes.append(node)
+        elif isinstance(node, (exp.Select, exp.Union)):
+            if isinstance(parent, exp.Subquery):
```
tobymao, line 111: how come these are not stored as subqueries? curious about why these are called derived_tables

tests/fixtures/optimizer/qualify_columns.sql
tobymao: i see what you're saying, but the issue with alias columns is i feel it's pretty rare and just duplicate of the aliases / names providied in SELECTS, so named_selections provides exactly this.

since we already force alias everything explicitly, i'd say we should get rid of the table column aliases. if the user provides them, we should "push those down" so that they don't exist in the canonical form.

tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -0,0 +1,124 @@
...
+
+SELECT a FROM (SELECT a AS a FROM x) y;
+SELECT y.a AS a FROM (SELECT x.a AS a FROM x) AS y(a);
+
+SELECT a FROM (SELECT a AS a FROM x);
+SELECT "_alias0".a AS a FROM (SELECT x.a AS a FROM x) AS "_alias0"(a);
```
tobymao, line 38: can we call it "_q_0" or "_query_0" ?

tobymao: for select x + 1, will you make the alias "_col_0" ?

tobymao:
i was thinking that in order to do this we needed to do a a dag ordered approach for example

SELECT 
FROM (
    SELECT 
    FROM (
           SELECT
            FROM c
     )  b
) a

We'd have to do c -> b -> a

So this can be achived by doing something like

expression.find_all(exp.Select).reverse()

it looks like you've managed to do this by doing the standard walk, is this why you needed to implement stop types? are you able to avoid doing duplicate work in your recursion because you pass context and merge it in dictionaries? how do you ensure that the context is correct in the queries where it may not have outter context? for example in the from c, it doesn't have context of b.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+        )
+
+        table_alias = subquery.args.get("alias")
+        if not table_alias:
+            table_alias = exp.TableAlias()
+            subquery.set("alias", table_alias)
+
+        alias = table_alias.args.get("this")
+        if not alias:
+            alias = exp.to_identifier(f"_alias{sequence.next()}")
```
tobymao, line 159: is there a way to use the sqlbuilder for this?

like wasn't there an alias method?

exp.alias_(subquery, "mystringthatgetsconvertedtoidentifier")

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+            table_alias.set("this", alias)
+
+        alias_columns = table_alias.args.get("columns")
+        if not alias_columns:
+            alias_columns = [exp.to_identifier(o) for o in subquery_outputs]
+            table_alias.set("columns", alias_columns)
+
+        selectable_table = alias.text("this")
+
+        selectable_columns = [c.text("this") for c in alias_columns]
```
tobymao, line 169: should it be calling alias_or_name?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+        if not table_alias:
+            table_alias = exp.TableAlias()
+            subquery.set("alias", table_alias)
+
+        alias = table_alias.args.get("this")
+        if not alias:
+            alias = exp.to_identifier(f"_alias{sequence.next()}")
+            table_alias.set("this", alias)
+
+        alias_columns = table_alias.args.get("columns")
```
tobymao, line 162: ah this is alias_columns, yea sorry, confused. i don't think these add any values as a concept

sqlglot/optimizer/qualify_columns.py
tobymao: my suggestion is wrong because they are alias_columns, not actual columns

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+
+def _qualify_tables(tables, schema, parent_selectables):
+    selectables = {}
+    for table in tables:
+        table_name = table.text("this")
+
+        if isinstance(table.parent, exp.Alias):
+            selectable_table = table.parent.alias
+        else:
+            selectable_table = table_name
```
tobymao, line 183: so you add alias to select x AS x even if x == x, to be consistent, we should probably do FROM x AS x as well -- or if we want a cleaner canonical form, we strip out aliases if alias == source. thougths?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+        new_columns = []
+
+        if isinstance(star, exp.Column):
+            tables = [star.text("table")]
+        else:
+            tables = list(selectables)
+
+        for table in tables:
+            columns = selectables.get(table, [])
+            for column in sorted(columns):
```
tobymao, line 206: is the sort here necessary?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+                new_columns.append(
```
tobymao, line 207: is this cleaner than

parse_one(f'"{table}"."{column}"')

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+                    exp.Column(
+                        this=exp.to_identifier(column), table=exp.to_identifier(table)
+                    )
+                )
+        star.replace(*new_columns)
```
tobymao, line 212: i've been looking for something like this, you're able to replace a single child node with multiple?

sort of unrelated to this PR, but can you add this functionality to def transform? so if i return an array it works? 

if i'm understanding correctly, then this is great!

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+def _merge(*selectables):
+    """
+    Merge two "selectable" dictionaries.
+
+    Raise an error if there are any duplicate selectable names.
+    """
+    result = {}
+    for s in selectables:
+        for selectable_table, selectable_columns in s.items():
+            if selectable_table in selectables:
```
tobymao, line 267: is this correct? selectables in a tuple right, so this will never be true, did you mean in result instead?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,321 @@
...
+    """
+    Find the unique columns in a list of columns.
+
+    Example:
+        >>> sorted(_unique_columns(["a", "b", "b", "c"]))
+        ['a', 'c']
+
+    This is necessary because duplicate column names are ambiguous.
+    """
+    unique = set()
```
tobymao, line 311: counts = {}
for column in columns:
   counts[column] = counts.get(column, 0) + 1
return [column for column, count in  counts.items() if count == 1]

tobymao:
my first pass is complete, this is amazing work @barakalon !

tests/fixtures/optimizer/qualify_columns.sql
tobymao: what i  mean is if a user writes (select a AS b) x (c), we should optimize this to be (SELECT a AS c) i think this is safe? if for some reason this is not safe, then ya maybe we need to do this table column alias

tobymao:
```sql
select 1 AS a
where a > 1
```

for example, this is not legal, meaning the aliases don't really have scope inside and mostly serve for projection

and so 
```sql
(select 1 AS a
... reference a
) AS table(c)
```

transformed to

```sql
(select 1 AS c
... reference a
) AS table

shouldn't happen, but again, i've never seen a sql query using alias projection EXCEPT in cases of explosive queries

Comment (barakalon) [reply tests/fixtures/optimizer/qualify_columns.sql:32]: Ok, I'm on board. It'll be more context we have to push down into subqueries... but I think that's fine.

File: sqlglot/optimizer/qualify_columns.py (hunk 3 above)
Comment (barakalon) [reply sqlglot/optimizer/qualify_columns.py:26]: Actually, Python has a built in for this: https://docs.python.org/3/library/itertools.html#itertools.count
I'll just use that

File: sqlglot/optimizer/qualify_columns.py (hunk 4 above)
Comment (barakalon) [reply sqlglot/optimizer/qualify_columns.py:48]: I'll change to _qualify_statement, to match the parser

File: sqlglot/optimizer/qualify_columns.py (hunk 5 above)
Comment (barakalon) [reply sqlglot/optimizer/qualify_columns.py:111]: I haven't quite figured out the naming here. I was borrowing from MySQL documentation.

```sql
SELECT * FROM x WHERE a IN (SELECT ...) <-- subquery
SELECT * FROM (SELECT ...)  <-- derived table... a special case of subqueries
```

Derived table add "selectables" to the current context.
The inverse happens with the other type of subquery... context get's pushed down, which support correlated subqueries.

I agree, the naming is super confusing here. I'll try to figure something out (and add some comments regardless).

tests/fixtures/optimizer/qualify_columns.sql
barakalon: Yeah.

I'm not sure if there's a standard way to generate projection aliases? Some engines seem to have some rule for stringifying the selection... but an incrementing string felt easier.

barakalon:
> it looks like you've managed to do this by doing the standard walk, is this why you needed to implement stop types? 

I'm not sure a simple DFS in "postorder" will work, since the order of recursion matters. For example:

```sql
SELECT * 
FROM (SELECT ...) as x
JOIN (SELECT ...) as y
ON x.a = y.a
```

We need to be sure to recurse into the subqueries _first_, so we can build the "selectables" for the current context before resolving columns. 

So we _are_ doing DFS in postorder... it's just in two step. First, we recurse down to Select statements, then we collect everything in the current context, stopping at Select statements, _then_ we recurse.

I think this might be a common need for these optimizer rules.

> are you able to avoid doing duplicate work in your recursion because you pass context and merge it in dictionaries? 

Yeah, I don't think there's any duplicate work here. We still visit the tree nodes the same amount of times. And we pass context down in the "selectables" dict.

> how do you ensure that the context is correct in the queries where it may not have outter context? for example in the from c, it doesn't have context of b.

That's why we track the different types of subqueries.

CTES get outer context, _plus_ the context can be chained. I think the same will be true for laterals, when we get there.
"Derived tables" don't get this outer context.
"Subqueries" do.

sqlglot/optimizer/qualify_columns.py
barakalon: The naming around aliases get confusing.

There's exp.Alias, exp.Subquery, _and_ exp.TableAlias. In this case, we're dealing with exp.Subquery with a exp.TableAlias child.

sqlglot/optimizer/qualify_columns.py
barakalon: I think there's a balance between _pretty_ and _consistent_.

In my mind, canonical SQL should prioritize consistency over prettiness. So I thought it was reasonable to coerce all the selections to the Alias type.

sqlglot/optimizer/qualify_columns.py
barakalon: This helps with testing, to ensure the expanded columns come in a deterministic order. I can remove it or add a comment?

sqlglot/optimizer/qualify_columns.py
barakalon: They feel similar to me.
I guess I don't always trust `parse` to give me what I want... so the more explicit instantiation feels nice. But I don't have a strong opinion here.

sqlglot/optimizer/qualify_columns.py
barakalon: Yeah, it does exactly that. I think this already works for `transform` in the way I wrote this. I'll add a test to prove it.

sqlglot/optimizer/qualify_columns.py
barakalon: Oo wow nice catch. 
I haven't added the "invalid" tests yet. Hopefully those would've caught this.

sqlglot/optimizer/qualify_columns.py
barakalon: Er, parse_one(f'"{table}"."{column}"') turns:

```sql
SELECT x.a AS a, x.b AS b FROM x
```

into 

```sql
SELECT "x"."a" AS a, "x"."b" AS b FROM x
```

If we go that route, maybe we should be quoting _everything_?

tobymao: canonical form should quote everything, but i sort of have this with sql(identify=true)

sqlglot/optimizer/rules/qualify_columns.py
```diff
@@ -0,0 +1,391 @@
+import dataclasses
```
tobymao, line 1: sorry to nit pick here a bit, but do we need the rules folder? i'd say we can keep rules if we plan to have other stuff in optimizer -- not sure if we will? maybe logical planning? if not, i'd just leave it on the optimizer level

sqlglot/optimizer/rules/qualify_columns.py
```diff
@@ -0,0 +1,391 @@
...
+    Returns:
+        list: output column names
+    """
+    if isinstance(expression, exp.Select):
+        return _qualify_select(expression, context)
+    if isinstance(expression, exp.Union):
+        left = _qualify_select(expression.this, context)
+        right = _qualify_statement(expression.args.get("expression"), context)
+        if set(left) != set(right):
+            raise OptimizeError("UNION columns not equal")
```
tobymao, line 88: in sqlglot, we have different error levels so that users can ignore errors, or just warn, should we do that as well here?

tests/fixtures/identity.sql
```diff
@@ -1,6 +1,5 @@
 SUM(1)
 SUM(CASE WHEN x > 1 THEN 1 ELSE 0 END) / y
-
```
tobymao, line 3: why was this test removed?

tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -0,0 +1,127 @@
...
+SELECT "x"."a" AS "a" FROM x;
+
+SELECT "x"."a" AS "a" FROM x;
+SELECT "x"."a" AS "a" FROM x;
+
+SELECT a AS b FROM x;
+SELECT "x"."a" AS "b" FROM x;
+
+SELECT 1, 2 FROM x;
+SELECT 1 AS "_col_0", 2 AS "_col_1" FROM x;
```
tobymao, line 20: sorry i forgot if i asked this but shouldn't this alsobe FROM x AS x

tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -0,0 +1,127 @@
...
+SELECT "y"."a" AS "a" FROM (SELECT "x"."a" AS "c", "x"."b" AS "b" FROM x) AS y;
+
+SELECT a FROM (SELECT a FROM x) y;
+SELECT "y"."a" AS "a" FROM (SELECT "x"."a" AS "a" FROM x) AS y;
+
+SELECT a FROM (SELECT a AS a FROM x);
+SELECT "_q_0"."a" AS "a" FROM (SELECT "x"."a" AS "a" FROM x) AS "_q_0";
+
+SELECT a FROM (SELECT a FROM (SELECT a FROM x));
+SELECT "_q_0"."a" AS "a" FROM (SELECT "_q_1"."a" AS "a" FROM (SELECT "x"."a" AS "a" FROM x) AS "_q_1") AS "_q_0";
```
tobymao, line 41: i notice your context is a global context and not scoped, since there wasn't a reason to increment _q_1 here, but that's fine, just an observation

tests/test_transpile.py
```diff
@@ -249,33 +250,24 @@ def test_index_offset(self, mock_logger):
         mock_logger.warning.assert_any_call("Applying array index offset (%s)", -1)
 
     def test_identity(self):
-        with open(
-            os.path.join(self.fixtures_dir, "identity.sql"), encoding="utf-8"
-        ) as f:
-            for sql in f:
+        self.assertEqual(transpile("")[0], "")
```
tobymao, line 253: ah i see you replaced my empty test

tobymao (approved):
again, amazing work @barakalon

tests/fixtures/optimizer/qualify_columns.sql
tobymao: some alias as x + 1 AS "x + 1" but this isn't unique or good

sqlglot/optimizer/qualify_columns.py
tobymao: sounds good

sqlglot/optimizer/qualify_columns.py
tobymao: looks like you removed this, but shouldn't it be determinsitic without the sort?

tests/fixtures/optimizer/qualify_columns.sql
barakalon: Yeah, we agree we should do that.
But I think that should be done in `qualify_tables`, so perhaps in a separate PR.

tests/fixtures/optimizer/qualify_columns.sql
barakalon: Ah, right, the counter is global. That's a good point. I cant reset the counter.

tests/test_transpile.py
barakalon: Yeah - I made the fixture loader remove empty lines.
Felt quick and easy to just add this test in.

sqlglot/optimizer/rules/qualify_columns.py
barakalon: I did this because the optimizer functions have the same name as the file.

`optimizer/qualify_columns.py`
```python
def qualify_columns(): ...
```

`optimizer/__init__.py`
```python
from optimizer import qualify_columns

def optimize():
    qualify_columns()
```

`some_other_module.py`
```python
from optimizer import qualify_columns  # <- Is this the module or the function?
```

Other options:
1. change the function names
2. change the file names
3. don't put code in `__init__.py`
4. add aliases on the imports in `__init__.py`

sqlglot/optimizer/rules/qualify_columns.py
barakalon: Ah, interesting.

That might be tricky to implement here. We'd probably want to move these rules into classes so it's easier to store global context (i.e. a list of error messages). We also might have to choose _how_ to ignore errors on a case by case basis... since doing _nothing_ might lead to other errors later.

I get the value of consistent interfaces though... 
And it might make sense to have better error reporting here. An exception from some nested expression might be tricky to debug.

tests/fixtures/optimizer/qualify_columns.sql
tobymao: sounsd good

sqlglot/optimizer/rules/qualify_columns.py
tobymao: true, let's keep it then. the issue being if a column is ambiguous, we won't be able to do some of the later optimizations. we can keep it

sqlglot/optimizer/rules/qualify_columns.py
tobymao: what's your preference? i like doing from optimizer import optimize

barakalon: A `rules` submodule is my preference 😛 

I understand your approach though - ruthless simplicity.

I'll move `optimize` to its own file, but import it in `__init__`.
That should keep things flat and get "from optimizer import optimize"

sqlglot/errors.py
```diff
@@ -19,3 +19,7 @@ class ParseError(ValueError):
 
 class TokenError(ValueError):
     pass
+
+
+class OptimizeError(ValueError):
```
nolamesa, line 24: clearly not affected by this PR but we should add a top level error for better handling outside of sqlglot

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,385 @@
...
+        ambiguous = set(unambiguous_columns).intersection(unique)
+        for column in ambiguous:
+            unambiguous_columns.pop(column)
+        for column in unique.difference(ambiguous):
+            unambiguous_columns[column] = table
+
+    return unambiguous_columns
+
+
+def _unique_columns(columns):
```
nolamesa, line 372: perhaps rename to find_unique_columns

sqlglot/expressions.py
```diff
@@ -141,23 +141,30 @@ def find_ancestor(self, *expression_types):
                 return None
         return ancestor
 
-    def walk(self, bfs=True):
+    def walk(self, bfs=True, stop_types=None):
```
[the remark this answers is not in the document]
barakalon: I like "stop_after" - I'll switch to that.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -0,0 +1,385 @@
...
+        list: output column names
+    """
+    if isinstance(expression, exp.Select):
+        return _qualify_select(expression, context)
+    if isinstance(expression, exp.Union):
+        left = _qualify_select(expression.this, context)
+        right = _qualify_statement(expression.args.get("expression"), context)
+        if set(left) != set(right):
+            raise OptimizeError(f"UNION columns not equal: {left} vs. {right}")
+        return left
```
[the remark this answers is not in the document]
barakalon: This are used during recursion, so outer queries can know the selectable columns of derived tables and CTEs.
I'll add a comment.

PR #135, opened 2022-05-23
feat(optimizer): qualify_tables to add aliases

barakalon:
This adds aliases to all tables.

This also quotes everything.

@tobymao

### DIFF
--- a/sqlglot/optimizer/helper.py
+++ b/sqlglot/optimizer/helper.py
@@ -1,8 +1,0 @@
-from sqlglot import expressions as exp
-
-
-def quote(expression, arg):
-    """If `arg` of `expression` is an Identifier, ensure it's quoted"""
-    node = expression.args.get(arg)
-    if node and isinstance(node, exp.Identifier):
-        node.set("quoted", True)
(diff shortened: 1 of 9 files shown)


sqlglot/optimizer/helper.py
```diff
@@ -0,0 +1,8 @@
+from sqlglot import expressions as exp
+
+
+def quote(expression, arg):
```
tobymao, line 4: can we just have this as a property of expression.?

expression.identify

tests/test_optimizer.py
```diff
@@ -13,9 +13,11 @@ def load_tests(loader, tests, ignore):  # pylint: disable=unused-argument
     """
     This finds and runs all the doctests in the expressions module
     """
-    from sqlglot.optimizer import qualify_columns as mod
+    from sqlglot.optimizer import qualify_columns as mod1
+    from sqlglot.optimizer import qualify_tables as mod2
```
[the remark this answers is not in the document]
barakalon: We already import the functions "qualify_columns" and "qualify_tables"

sqlglot/optimizer/helper.py
barakalon: Sure

tobymao (approved):

PR #136, opened 2022-05-23
fix: qualify_columns refactors

barakalon:
This addresses @nolamesa 's comments on this PR: https://github.com/tobymao/sqlglot/pull/133, which was merged before Nikos got to it.

### DIFF
--- a/sqlglot/errors.py
+++ b/sqlglot/errors.py
@@ -9,17 +9,21 @@ class ErrorLevel(AutoName):
     RAISE = auto()
 
 
-class UnsupportedError(ValueError):
+class SqlglotError(ValueError):
     pass
 
 
-class ParseError(ValueError):
+class UnsupportedError(SqlglotError):
     pass
 
 
-class TokenError(ValueError):
+class ParseError(SqlglotError):
     pass
 
 
-class OptimizeError(ValueError):
+class TokenError(SqlglotError):
+    pass
+
+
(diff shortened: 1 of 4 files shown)


sqlglot/errors.py
```diff
@@ -9,17 +9,21 @@ class ErrorLevel(AutoName):
     RAISE = auto()
 
 
-class UnsupportedError(ValueError):
+class SqlglotError(ValueError):
```
barakalon, line 12: I'm not sure if this should subclass ValueError... but removing that subclass would technically be a breaking change. So I was scared to do otherwise.

sqlglot/expressions.py
```diff
@@ -1190,6 +1190,10 @@ def ctas(self, table, properties=None, dialect=None, parser_opts=None, copy=True
     def named_selects(self):
         return [e.alias_or_name for e in self.args["expressions"] if e.alias_or_name]
 
+    @property
+    def selections(self):
```
tobymao, line 1194: why not selects?

nolamesa (approved):

tobymao (approved):

sqlglot/expressions.py
barakalon: Ah sure

barakalon: FWIW: All these properties have gotten a little confusing... some return strings, some return Expressions.

PR #137, opened 2022-05-24
feat(optimizer): remove unused columns

barakalon:
This adds another optimizer step to remove unused columns.

I think this is also called "projection pushdown"?

I also fixes some bugs for qualify_columns I found while building this.

@tobymao @nolamesa


barakalon:
Er, might have to think harder about certain join types...

sqlglot/optimizer/remove_unused_columns.py
```diff
@@ -0,0 +1,162 @@
+from collections import defaultdict
```
tobymao, line 1: let's call this projection_pushdown

sqlglot/optimizer/remove_unused_columns.py
```diff
@@ -0,0 +1,162 @@
...
+from copy import copy
+
+from sqlglot import expressions as exp
+from sqlglot.errors import OptimizeError
+from sqlglot.optimizer.helper import select_scope
+
+# Sentinel value that means an outer query selecting ALL columns
+SELECT_ALL = "*"
```
tobymao, line 9: let's put this in constants.py

sqlglot/optimizer/remove_unused_columns.py
```diff
@@ -0,0 +1,162 @@
...
+        dict: Mapping of selectable names to columns.
+        This is used during recursion, so the outer query can:
+            1. pullup any selected columns
+            2. pushdown selected columns into CTEs
+    """
+    if isinstance(expression, exp.Select):
+        return _pushdown_select(expression, parent_selections)
+    if isinstance(expression, exp.Union):
+        if expression.args.get("distinct"):
+            # We can't remove selections on UNION ALL
```
tobymao, line 52: is this comment reversed? if it's distinct it means it's not a union all

sqlglot/optimizer/remove_unused_columns.py
```diff
@@ -0,0 +1,162 @@
...
+        # Use the Expression identity for the key since Expressions are hashed by value.
+        columns[id(column)] = column
+
+    # Collect all the selections
+    if not expression.args.get("distinct"):
+        columns = _remove_unused_columns(expression, columns, parent_selections)
+
+    for subquery in scope.subqueries:
+        # Subqueries (as opposed to "derived_tables") aren't "selectable".
+        # So the none of the columns in the current scope to reference these.
```
tobymao, line 98: so none of the

sqlglot/optimizer/qualify_columns.py
```diff
@@ -264,7 +265,7 @@ def _qualify_columns(columns, selectables):
             column.set("table", exp.to_identifier(column_table))
 
 
-def _qualify_subqueries(subqueries, context):
+def _qualify_subqueries(subqueries, selectables, context):
     """
     Derived tables are subqueries that are NOT selectable in a surrounding context.
```
tobymao, line 270: this comment is opposite? Derived tables are subqueries that are selectable


cuz you say ```
    In this case, the subquery cannot be selected from (as opposed to derived tables).```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -42,12 +43,16 @@ class Context:
         aliased_columns: List of column names.
             This should be set for derived tables or CTEs where the parent context
             defines column names for an alias, e.g. "SELECT * FROM (SELECT * FROM x) AS y(a, b)"
+        correlated_selectables: Mapping of name to selectable columns.
```
tobymao, line 46: i saw you moved out class Selectable, do we need to move out Context as well for other optimization functions?

sqlglot/optimizer/remove_unused_columns.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+            raise OptimizeError(msg)
+
+        if parent_selections == SELECT_ALL or selection.alias in parent_selections:
+            new_selections.append(selection)
+        else:
+            # Pop the column out of the set of all columns.
+            # Later, we'll use this set of columns to pushdown the selected columns from inner queries.
+            for column in selection.find_all(exp.Column):
+                columns.pop(id(column))
```
tobymao, line 146: why do id? why not just put column into the set as is?

sqlglot/optimizer/remove_unused_columns.py
barakalon: 👍 
I had it that at first, then changed it to match https://github.com/duckdb/duckdb/blob/master/src/optimizer/remove_unused_columns.cpp
I'll switch it back

sqlglot/optimizer/remove_unused_columns.py
barakalon: I'm a little worried that it doesn't have much meaning outside this module

sqlglot/optimizer/remove_unused_columns.py
barakalon: Ah, indeed. nice catch

sqlglot/optimizer/qualify_columns.py
barakalon: Indeed, once again, nice catch

sqlglot/optimizer/remove_unused_columns.py
tobymao: hm ok, up to you

tobymao (approved):

sqlglot/optimizer/qualify_columns.py
barakalon: I moved out "SelectScope".... not super crazy about the name.

In this case, "Context" is specific to this rule.

I feel like there's potentially a bigger, better abstraction here. I haven't quite nabbed it yet, though

sqlglot/optimizer/remove_unused_columns.py
barakalon:
```python
hash(exp.Column(this="a")) == hash(exp.Column(this="a"))
```
We need to be able to remove the exact column instance from this set. So I'm using `id` instead.

sqlglot/optimizer/qualify_columns.py
nolamesa: For SelectScope naming: you can perhaps rename to SelectParts, or consider moving it inside select class as property methods e.g. tables, columns etc

nolamesa (commented):
As always great work!

sqlglot/optimizer/helper.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    ctes: list = dataclasses.field(default_factory=list)
+    derived_tables: list = dataclasses.field(default_factory=list)
+    subqueries: list = dataclasses.field(default_factory=list)
+    tables: list = dataclasses.field(default_factory=list)
+    columns: list = dataclasses.field(default_factory=list)
+
+
+def select_scope(select):
```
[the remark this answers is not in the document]
barakalon: 🤷  I debated it. I can switch to that

sqlglot/optimizer/helper.py
```diff
@@ -0,0 +1,59 @@
...
+    Returns:
+        SelectScope: SelectScope instance
+    """
+    scope = SelectScope()
+
+    for node, parent, _ in select.walk(stop_after=(exp.Select, exp.Union)):
+        if node is select:
+            continue  # Skip this node itself - we only care about children
+        if isinstance(node, (exp.Select, exp.Union)):
+            if isinstance(parent, exp.CTE):
```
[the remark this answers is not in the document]
barakalon: Ah, interesting. I think that'd double register certain nodes:

```
(SELECT from: 
  (SUBQUERY this:  <- registered as derived_table
    (SELECT ...  <- registered again, as a subquery
```
Let me see if I can get a failing test.

I was also considering having `walk` allow for a stop flag to be sent back to the generator. That might make it easier to stop at arbitrary nodes... but the syntax for that kind of generator is a little clunky.

sqlglot/optimizer/helper.py
```diff
@@ -0,0 +1,59 @@
...
+    won't traverse into the inner query.
+
+    Args:
+         select (exp.Select): expression to search
+    Returns:
+        SelectScope: SelectScope instance
+    """
+    scope = SelectScope()
+
+    for node, parent, _ in select.walk(stop_after=(exp.Select, exp.Union)):
```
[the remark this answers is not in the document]
barakalon: It's "top level"... but the top level isn't inherent to the Expression tree. Take this tree for example:

```
(SELECT expressions:
  (ALIAS: this:
    (COLUMN: ...)) from: 
  (SELECT ...
```

The COLUMN is two levels deep, but we want it here. We're sort of _flattening_ SELECT expressions.

I agree, the name doesn't feel right here. Nor does this abstraction feel great, in general.

sqlglot/optimizer/projection_pushdown.py
```diff
@@ -0,0 +1,162 @@
+from collections import defaultdict
+from copy import copy
+
+from sqlglot import expressions as exp
+from sqlglot.errors import OptimizeError
+from sqlglot.optimizer.helper import select_scope
+
+# Sentinel value that means an outer query selecting ALL columns
+SELECT_ALL = "*"
```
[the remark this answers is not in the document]
barakalon: Ah yes, "*" is confusing.

I think `sentinel = object()` is pretty common. I'll do something like that.

sqlglot/optimizer/projection_pushdown.py
```diff
@@ -0,0 +1,162 @@
...
+        if parent_selections == SELECT_ALL or selection.alias in parent_selections:
+            new_selections.append(selection)
+        else:
+            # Pop the column out of the set of all columns.
+            # Later, we'll use this set of columns to pushdown the selected columns from inner queries.
+            for column in selection.find_all(exp.Column):
+                columns.pop(id(column))
+
+    # If there are no remaining selections, just select a single constant
+    if not new_selections:
```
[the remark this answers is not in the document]
nolamesa: yeah removing joins is tough, safe goes for inner joins which act as a filter effectively

sqlglot/optimizer/helper.py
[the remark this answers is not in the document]
nolamesa: hmm how about this alternative: make it so you call bfs with depth = 1 and then if you have aliases you get back the aliases and pull out the columns with find all

sqlglot/optimizer/optimizer.py
```diff
@@ -1,13 +1,15 @@
...
+from sqlglot.optimizer.projection_pushdown import projection_pushdown
 
 
 def optimize(expression, schema=None, db=None, catalog=None):
     """
     Rewrite a sqlglot AST into an optimized form.
     """
     expression = qualify_tables(expression, db=db, catalog=catalog)
-    expression = qualify_columns(expression, schema)
+    expression = qualify_columns(expression, schema or {})
```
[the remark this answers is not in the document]
barakalon: Do you mean `select *` or `select * from x`?
I don't think the former is a valid statement at all, which sqlglot isn't too strict about yet.
That latter would complain about an unknown table.

barakalon:
SelectParts is a way better name. I'll switch to that.

I'd like to keep it out of Select for now... I'm worried about that class getting too large and confusing. And I don't think this concept is ready for prime time yet.

sqlglot/optimizer/projection_pushdown.py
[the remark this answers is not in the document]
tobymao: it's not a valid select statement to do a select with nothing

PR #139, opened 2022-05-24
Add Dateint Support

eakmanrq:
* Adds support for basic dateint support (Date -> Dateint and Dateint -> Date)
* Adds missing Datestring -> Date
* Make sure `DATE` functions (ex: `DATE_DIFF`) work on date objects
* Have `TS_OR_DS_TO_DATE` behave consistently across Presto and Hive by having Presto cast to date at the end (otherwise it would return a timestamp)


tests/test_dialects.py
```diff
@@ -476,7 +513,7 @@ def test_presto(self):
         )
         self.validate(
             "TS_OR_DS_TO_DATE(x)",
-            "DATE_PARSE(SUBSTR(x, 1, 10), '%Y-%m-%d')",
+            "CAST(DATE_PARSE(SUBSTR(CAST(x AS VARCHAR), 1, 10), '%Y-%m-%d') AS DATE)",
```
tobymao, line 516: doesn't date_parse return a date, why does it needed to be casted here?

eakmanrq: Oh yes thanks for calling that out. So without that cast technically Presto is returning a timestamp object. So that final cast gets it into an actual date.

eakmanrq: I would consider this a bug from before since `TS_OR_DS_TO_DATE` for hive actually returns a date so the two engines had different behavior for the same sqlglot function. Also I find this more intuitive since the name is clearly wanting a date.

tobymao (approved):

tobymao:
@eakmanrq looks good

PR #141, opened 2022-05-25
feat(optimizer): traverse scope

barakalon:
This introduces the `traverse_scope` function.

I've also refactored all the existing optimizers to use this function (including qualify_tables, which had a bug).

@nolamesa @tobymao


sqlglot/optimizer/scope.py
```diff
@@ -0,0 +1,274 @@
...
+def _traverse_derived_tables(derived_tables, scope):
+    for derived_table in derived_tables:
+        for child_scope in _traverse_scope(
+            scope.branch(
+                derived_table.this, outer_column_list=derived_table.alias_column_names
+            )
+        ):
+            yield child_scope
+            # Tables without aliases will be set as ""
+            scope.selectables[derived_table.alias] = child_scope
```
tobymao, line 252: ok to have duplicates here clashing?

sqlglot/optimizer/scope.py
```diff
@@ -0,0 +1,274 @@
...
+        if isinstance(table.parent, exp.Alias):
+            selectable_table = table.parent.alias
+        else:
+            selectable_table = table_name
+
+        if table_name in scope.selectables:
+            # This is a reference to a parent selectable (e.g. a CTE), not an actual table.
+            scope.selectables[selectable_table] = scope.selectables[table_name]
+        elif selectable_table in scope.selectables:
+            raise OptimizeError(f"Duplicate table name: {selectable_table}")
```
tobymao, line 218: is this actually an error?

sqlglot/optimizer/scope.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        Column outputs of this scope.
+
+        For example, for the following expression:
+            SELECT 1 as a, 2 as b FROM x
+
+        The outputs are {"a", "b"}
+
+        Returns:
+            set[str]: outputs
```
tobymao, line 122: named_selects is not a set, it's a list, and should probably stay taht way

sqlglot/optimizer/projection_pushdown.py
```diff
@@ -26,137 +24,54 @@ def projection_pushdown(expression):
...
-        selections = _pushdown_select(expression.this, parent_selections)
-        selections = _merge_selections(
-            selections,
-            _pushdown_select(expression.args.get("expression"), parent_selections),
-        )
+    # We build the scope tree (which is traversed in DFS postorder), then iterate
+    # over the result in reverse order. This should ensure that the set of selected
+    # columns for a particular scope are completely build by the time we get to it.
+    for scope in reversed(list(traverse_scope(expression))):
+        parent_selections = referenced_columns.get(id(scope), {SELECT_ALL})
```
tobymao, line 35: why do you need to get id(scope), shouldn't it default to that? scope is not an expression so it doesn't overide hash right

sqlglot/optimizer/scope.py
barakalon: Hm, I'm not sure what standard SQL behavior is.

Seems like sqlite allows duplicate selectable names... it just throws an error if you reference an ambiguous column.

That means we'd have to track all selectables, including duplicates. Hm

sqlglot/optimizer/scope.py
barakalon: Again, I'm not too sure about standards here. Do you have a good handle on what we should do?

sqlglot/optimizer/projection_pushdown.py
barakalon: Good catch. Updating...

tobymao (approved):

PR #142, opened 2022-05-26
fix(optimizer): add catalog and db to schema

barakalon:
Yikes.

@tobymao

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -218,14 +218,10 @@ def _get_selectable_columns(name, selectables, schema):
 
     # If referencing a table, return the columns from the schema
     if isinstance(selectable, exp.Table):
-        return _get_column_names(schema, selectable)
+        try:
+            return schema.column_names(selectable)
+        except Exception as e:
+            raise OptimizeError(str(e)) from e
 
     # Otherwise, if referencing another scope, return that scope's outputs
     return selectable.outputs
-
-
-def _get_column_names(schema, table):
-    try:
-        return schema.column_names(table)
-    except Exception as e:
-        raise OptimizeError(str(e)) from e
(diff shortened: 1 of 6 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -202,10 +202,21 @@ def _get_selectable_columns(name, selectables, schema):
 
     # If referencing a table, return the columns from the schema
     if isinstance(selectable, exp.Table):
+        catalog_name = selectable.text("catalog")
```
tobymao, line 205: can we make it so catalog and db are optional? so gracefully handle the cases where the schema object has different levels of nesting? i presume in many cases, we'd only pass in tables, and in most cases only db/table in schema. i don't think we'll ever use catalog

barakalon:
@tobymao An elegant interface that supports a simple dictionary _and_ catalog/database nested tables was surprisingly tricky.

I don't like introspecting the provided dictionary too deeply - that can get quite nasty.

So this just assumes that a simple dictionary is only a mapping of tables.

If a user wants databases/catalogs, we'll need other implementations of the abstract `Schema` class. 

I need databases in Minerva, but I also want to lazy load table metadata, so implementing the Schema class is actually better.

tests/test_docs.py
```diff
@@ -2,6 +2,7 @@
 import inspect
 
 import sqlglot
+import sqlglot.optimizer
```
tobymao, line 6: why is this needed? we’re the test broken?

tobymao (approved):
i think this is reasonable, can you implement the catalog and db versions with nested dicts? should be trivial right?

tests/test_docs.py
barakalon: Yeah, it wasn't running locally. Might be a setup quirk on my side... but this felt harmless.

barakalon:
Alright, this now supports:
1. {table: {col: type}}
2. {db: {table: {col: type}}}
3. {catalog: {db: {table: {col: type}}}}

PR #143, opened 2022-05-26
add simplify optimizer

tobymao:


sqlglot/optimizer/simplify.py
```diff
@@ -0,0 +1,100 @@
...
+        right = node.right
+
+        if NULL in (left, right):
+            node.replace(NULL)
+        elif left == right:
+            node.replace(TRUE)
+        elif (
+            isinstance(left, exp.Literal)
+            and isinstance(right, exp.Literal)
+            and left.is_string
```
nolamesa, line 40: so this does not replace `1 = 2` with false? why is that?

tests/fixtures/optimizer/simplify.sql
```diff
@@ -0,0 +1,89 @@
...
+FALSE;
+
+((NOT FALSE) AND (x = x)) AND (TRUE OR 1 <> 3);
+TRUE;
+
+((NOT FALSE) AND (x = x)) AND (FALSE OR 1 <> 2);
+TRUE AND (1 <> 2);
+
+(('a' = 'a') AND TRUE and NOT FALSE);
+TRUE;
```
nolamesa, line 89: can we also add some non happy path tests where replacements dont happen. Also perhaps have a full select statement as an example where this is used

.pylintrc
```diff
@@ -13,6 +13,7 @@ disable=
     no-self-use,
     too-few-public-methods,
     too-many-arguments,
+    too-many-boolean-expressions,
```
nolamesa, line 16: how ironic..

sqlglot/optimizer/simplify.py
```diff
@@ -0,0 +1,100 @@
...
+
+    while True:
+        start = hash(expression)
+        expression = simplify_equality(expression)
+        expression = simplify_not(expression)
+        expression = simplify_conjunctions(expression)
+        expression = simplify_parens(expression)
+
+        if start == hash(expression):
+            break
```
nolamesa, line 23: I wonder if there could be an infinite loop scenario where one rule fights another. Cant think of one but just in case you can think of something

nolamesa (approved):
will you be adding this in the list of optimize rules that run?

barakalon (approved):
nice

sqlglot/optimizer/simplify.py
tobymao: because number comparison is harder, i don't want to do it yet, we can add it later, for example

1.0 = 1

tests/fixtures/optimizer/simplify.sql
tobymao: i have that

((NOT FALSE) AND (x = x)) AND (FALSE OR 1 <> 2);
TRUE AND (1 <> 2);

tobymao:
```1.0 = 1;
1.0 = 1;
```

tobymao:
```FALSE OR x = y;
x = y;
```

sqlglot/optimizer/simplify.py
tobymao: yea, i thought of this, i couldn't think of a case where it happens yet because everything is a reduction

PR #144, opened 2022-05-26
initial pass at decorrelating sub queries

tobymao:


tests/test_optimizer.py
```diff
@@ -17,14 +20,8 @@ def load_tests(loader, tests, ignore):  # pylint: disable=unused-argument
...
     """
-    from sqlglot.optimizer import qualify_columns as module1
-    from sqlglot.optimizer import qualify_tables as module2
-    from sqlglot.optimizer import quote_identities as module3
-    from sqlglot.optimizer import projection_pushdown as module4
-    from sqlglot.optimizer import scope as module5
-
-    for mod in (module1, module2, module3, module4, module5):
-        tests.addTests(doctest.DocTestSuite(mod))
+    for _, module in inspect.getmembers(sqlglot.optimizer, inspect.ismodule):
```
barakalon, line 23: nice

tests/fixtures/optimizer/rewrite_subqueries.sql
```diff
@@ -0,0 +1,14 @@
+SELECT * FROM x AS x WHERE (SELECT y.a FROM y AS y WHERE x.a = y.a) = 1;
+SELECT * FROM x AS x JOIN (SELECT y.a FROM y AS y WHERE TRUE GROUP BY y.a) AS "_d_0" ON _d_0.a = x.a AND ("_d_0".a) = 1 WHERE TRUE;
```
barakalon, line 2: Should we be quoting `_d_0` in the condition too?

sqlglot/optimizer/rewrite_subqueries.py
```diff
@@ -0,0 +1,86 @@
...
+                eq.replace(exp.TRUE)
+                select.replace(
+                    exp.Column(
+                        this=exp.to_identifier(value.alias_or_name),
+                        table=exp.to_identifier(alias),
+                    )
+                )
+
+                condition = select.find_ancestor(
+                    exp.EQ, exp.NEQ, exp.LT, exp.LTE, exp.GT, exp.GTE
```
barakalon, line 75: What about operators like IN?

tests/fixtures/optimizer/rewrite_subqueries.sql
tobymao: i thought about that but doesn’t it happen in a later optimization step which quotes everything?

sqlglot/optimizer/rewrite_subqueries.py
tobymao: i’ll add all of these to a constant somewhere

sqlglot/optimizer/rewrite_subqueries.py
```diff
@@ -0,0 +1,86 @@
...
+                alias = f"_d_{next(sequence)}"
+                eq.replace(exp.TRUE)
+                select.replace(
+                    exp.Column(
+                        this=exp.to_identifier(value.alias_or_name),
+                        table=exp.to_identifier(alias),
+                    )
+                )
+
+                condition = select.find_ancestor(
```
barakalon, line 74: I think it's possible for correlated subqueries in selects:

```sql
SELECT
    (SELECT MAX(c) FROM y WHERE y.b = x.b) AS max
FROM x
```

barakalon (approved):
very cool

tests/fixtures/optimizer/rewrite_subqueries.sql
barakalon: Ah yeah sure,

sqlglot/optimizer/rewrite_subqueries.py
tobymao: is that optimizable one the same way? i’m sure there are many forms were missing, i’ll probably handle them later if at all

barakalon: I think so?

```sql
SELECT 
    _d_0.max
FROM 
    x
JOIN (
    SELECT 
        b, 
        MAX(c)
    FROM 
        y
    GROUP BY b
) AS _d_0
ON _d_0.b = x.b
```

PR #145, opened 2022-05-26
fix(optimizer): correlated subqueries are hard

barakalon:
I had some bugs with correlated subqueries.

I think this is a cleaner way of handling them.

@tobymao This removes "correlated_selectables". I know you need it in your other branch. However, this introduces "is_correlated_subquery", which I think will be better for you.

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -1,7 +1,16 @@
 from copy import copy
+from enum import Enum, auto
 
 from sqlglot import expressions as exp
 from sqlglot.errors import OptimizeError
+
+
+class ScopeType(Enum):
+    ROOT = auto()
+    SUBQUERY = auto()
+    DERIVED_TABLE = auto()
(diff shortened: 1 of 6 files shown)


sqlglot/optimizer/scope.py
```diff
@@ -140,6 +157,62 @@ def selects(self):
...
+    def correlations(self):
+        """
+        Columns referenced by correlated subqueries.
+
+        Returns:
+            list[tuple[exp.Column, Scope]]: List of Column instances from
+                child scopes that reference this scope, along with the child
+                scope instance itself.
+        """
+        result = []
```
tobymao, line 170: should this be a dictionary?

tobymao (approved):

sqlglot/optimizer/scope.py
barakalon: There might be instances multiple of the same logical column. And expressions hash on value, not identity.

PR #146, opened 2022-05-27
Make pretty a module config

eakmanrq:
This allows users to set pretty to True for an entire project if they prefer the default behavior to be pretty.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -68,7 +68,7 @@ class Generator:
         type_mapping=None,
         time_mapping=None,
         time_trie=None,
-        pretty=False,
+        pretty=None,
         identifier=None,
         identify=False,
         quote=None,
@@ -79,11 +79,13 @@ class Generator:
         unsupported_level=ErrorLevel.WARN,
     ):
         # pylint: disable=too-many-arguments
+        import sqlglot
+
         self.transforms = {**self.TRANSFORMS, **(transforms or {})}
         self.type_mapping = type_mapping or {}
         self.time_mapping = time_mapping or {}
         self.time_trie = time_trie
-        self.pretty = pretty
+        self.pretty = pretty if pretty is not None else sqlglot.pretty
         self.configured_pretty = pretty
         self.identifier = identifier or '"'
         self.identify = identify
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlglot/generator.py
```diff
@@ -79,11 +79,13 @@ def __init__(
         unsupported_level=ErrorLevel.WARN,
     ):
         # pylint: disable=too-many-arguments
+        import sqlglot
```
barakalon, line 82: I'm assuming we need to do this because it's a circular import?

eakmanrq: Yep that is correct

tests/test_parser.py
```diff
@@ -120,3 +120,13 @@ def test_annotations(self):
         assert expression.args["expressions"][3].text("this") == "c#annotation3"
         assert expression.args["expressions"][4].text("this") == "annotation4"
         assert expression.args["expressions"][5].text("this") == ""
+
+    def test_pretty_config_override(self):
+        import sqlglot
+
+        self.assertEqual(parse_one("SELECT col FROM x").sql(), "SELECT col FROM x")
+        sqlglot.pretty = True
```
barakalon, line 128: nit: you can use patch:

```python
from unittest.mock import patch

...

with patch("sqlglot.pretty", True):
    self.assertEqual(...)
```

That should prevent all subsequent tests blowing up if this fails.

eakmanrq: That is a great improvement. Thanks!

sqlglot/generator.py
barakalon: Opinions on moving it to another module? Like... `sqlglot.settings`?

I know sqlglot already has a bunch of circular import workarounds like this... but 🤷 . Seems like something we should try to avoid.

eakmanrq: Yeah I actually initially had a version of this where I created a `config` file so it was `sqlglot.config.pretty`. This approach is really simple when using it from Minerva since you just do `sqlglot.pretty = True`. I'm open either way.

barakalon: `sqlglot.config.pretty = True` doesn't feel too different to me. But I don't have a strong opinion here.

tobymao: let’s just keep it as is, very simple

nolamesa (approved):

tobymao:
@eakmanrq feel free to merge

PR #147, opened 2022-05-27
expand multi table selects

tobymao:
still need to clean up things like the find_ancestor(exp.EQ), but first pass is complete

### DIFF
--- a/sqlglot/optimizer/rewrite_subqueries.py
+++ b/sqlglot/optimizer/rewrite_subqueries.py
@@ -7,6 +7,7 @@ import sqlglot.expressions as exp
 def rewrite_subqueries(expression):
     expression = expression.copy()
     expression = decorrelate_subqueries(expression)
+    expression = expand_multi_table_selects(expression)
     return expression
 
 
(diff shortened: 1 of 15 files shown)


sqlglot/optimizer/rewrite_subqueries.py
```diff
@@ -7,6 +7,7 @@
 def rewrite_subqueries(expression):
     expression = expression.copy()
     expression = decorrelate_subqueries(expression)
+    expression = expand_multi_table_selects(expression)
```
barakalon, line 10: Are these subqueries? Maybe we should have two different steps:

"decorrelate_subqueries"
"expand_multi_table_selects"

Instead of both in "rewrite_subqueries"?

sqlglot/optimizer/rewrite_subqueries.py
```diff
@@ -84,3 +85,27 @@ def decorrelate_subqueries(expression):
...
+                    column.find_ancestor(exp.EQ)
+                    for column in (where.find_all(exp.Column) if where else [])
+                    if column.text("table") == alias
+                ]
+                if condition
+            ]
+
+            for condition in conditions:
+                condition.replace(exp.TRUE)
+            parent.join(query, on=conditions, copy=False)
```
barakalon, line 109: I'm not that familiar with comma joins, and I know this is starting small, but will we need something like this?:

```python
if conditions:
    parent.join(query, on=conditions, copy=False)
else:
    parent.join(query, join_type="cross", copy=False)
```

barakalon (approved):
Wow, fast

sqlglot/optimizer/rewrite_subqueries.py
tobymao: a join with conditions is a cross join

PR #152, opened 2022-05-30
simple predicate pushdown

tobymao:


tests/fixtures/optimizer/predicate_pushdown.sql
```diff
@@ -0,0 +1,11 @@
+SELECT x.a AS a FROM (SELECT a FROM x AS x) AS x JOIN y WHERE x.a = 1 AND x.b = 1 AND y.a = 1;
+SELECT x.a AS a FROM (SELECT a FROM x AS x WHERE a = 1 AND x.b = 1) AS x JOIN y ON y.a = 1 WHERE TRUE AND TRUE AND TRUE;
```
barakalon, line 2: Not relevant to this PR, but do we need to run `simplify` again? And maybe pop off the entire WHERE clause if its just true?

sqlglot/optimizer/predicate_pushdown.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Args:
+        expression (sqlglot.Expression): expression to optimize
+    Returns:
+        sqlglot.Expression: optimized expression
+    """
+    for scope in reversed(traverse_scope(expression)):
+        where = scope.expression.args.get("where")
+
+        if not where or where.find(exp.Or):
```
barakalon, line 24: I imagine we're just starting with the basic case, but what about..

```sql
SELECT * 
FROM (...) AS x 
WHERE 
  x.a = 1 OR x.b = 2
```

And others.

What is the actual logical rule?

sqlglot/optimizer/predicate_pushdown.py
```diff
@@ -0,0 +1,71 @@
...
+
+                def replace_alias(column):
+                    # pylint: disable=cell-var-from-loop
+                    if isinstance(column, exp.Column) and column.name in aliases:
+                        return aliases[column.name]
+                    return column
+
+                predicate = predicate.transform(replace_alias)
+            else:
+                continue
```
barakalon, line 63: How do we get here? If selectable is None?

barakalon (approved):

tests/fixtures/optimizer/predicate_pushdown.sql
barakalon: Also, it's strange the predicate became `a = 1`, unqualified. Is that right?

sqlglot/optimizer/optimizer.py
```diff
@@ -17,6 +18,7 @@ def optimize(expression, schema=None, db=None, catalog=None):
     expression = projection_pushdown(expression)
     expression = decorrelate_subqueries(expression)
     expression = expand_multi_table_selects(expression)
+    expression = predicate_pushdown(expression)
```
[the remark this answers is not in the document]
tobymao: def we should do number 1, not sure about number 2 yet

sqlglot/optimizer/predicate_pushdown.py
tobymao: right, this is more complex and requires converting to CNF (conjunctive normal form), i'm actually working on this right now but i have to figure out how to apply the distributive law

tests/fixtures/optimizer/predicate_pushdown.sql
tobymao: correct, this is only testing predicate_pushdown, if you run it through full optimizer, all these things will happen

these tests isolate to one optimizer function

barakalon: `qualify_columns` happens before `predicate_pushdown` right now.

barakalon: (..so I don't think we'd qualify `a` with the full optimizer yet)

PR #154, opened 2022-06-01
conjunctive normal form

tobymao:


sqlglot/optimizer/conjunctive_normal_form.py
```diff
@@ -0,0 +1,90 @@
...
+                    l,
+                    lambda c: exp.and_(
+                        exp.paren(exp.or_(c, r.left)),
+                        exp.paren(exp.or_(c, r.right)),
+                    ),
+                )
+            else:
+                l = exp.and_(exp.or_(l, r.left), exp.or_(l, r.right))
+            return l
+        if isinstance(l, exp.And):
```
barakalon, line 66: Sorry to be the DRY police - is this `l` block the same as the `r`?

barakalon (approved):
Very cool

sqlglot/optimizer/conjunctive_normal_form.py
tobymao: but i am!

barakalon: hahaha then a function it is!

PR #155, opened 2022-06-01
add absorb and eliminate

tobymao:
eliminate is not done yet

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -1,12 +1,8 @@
 import unittest
 
 import sqlglot
-import sqlglot.optimizer
-from sqlglot.optimizer import optimize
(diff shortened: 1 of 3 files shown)


barakalon (approved):
Neat, looks good so far 👍

tests/test_optimizer.py
```diff
@@ -32,57 +27,34 @@ def setUp(self):
             },
         }
 
+    def check_file(self, file, func=None, pretty=False, **kwargs):
+        module = getattr(sqlglot.optimizer, file)
+        func = getattr(module, func or file)
```
barakalon, line 32: So much magic haha
Can we just pass the function in?
It'd help us on dorky IDEs that have formal "refactor" features... not the hardcore folks who do find-and-replace

barakalon (approved):

PR #157, opened 2022-06-02
compare and prune chains of ands and ors to avoid sorting mismatches

tobymao:


sqlglot/optimizer/simplify.py
```diff
@@ -98,17 +97,46 @@ def simplify_conjunctions(expression):
...
+        return _compare_and_prune(expression, exp.Or, TRUE, exp.or_)
+    return expression
+
+
+def _compare_and_prune(expression, kind, compliment, result_func):
+    queue = [expression]
+    args = set()
+
+    while queue:
+        expression = queue.pop()
```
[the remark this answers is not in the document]
tobymao: i know, but i don't really care

sqlglot/optimizer/simplify.py
```diff
@@ -98,17 +97,46 @@ def simplify_conjunctions(expression):
...
+
+        for value in expression.args.values():
+            if isinstance(value, kind):
+                queue.append(value)
+            else:
+                args.add(value)
+
+    args = sorted(args, key=lambda a: a.sql())
```
nolamesa, line 131: why do we sort here? should we try to preserve order?

tobymao: no, we want a AND b == b AND a

PR #159, opened 2022-06-03
really basic stupid subquery elimination

tobymao:


nolamesa (approved):

barakalon (approved):

barakalon:
This is interesting. Since we're not optimizing query plans, CTEs are our way of eliminating subqueries. 

I wonder what implications this has for people trying to convert an optimized sqlglot expression into a query plan?

tobymao:
> This is interesting. Since we're not optimizing query plans, CTEs are our way of eliminating subqueries.
> 
> I wonder what implications this has for people trying to convert an optimized sqlglot expression into a query plan?

It makes it easier because they can just follow the scope traversal to create a DAG.

An expensive subquery would be generated in the cte which they could use down stream as opposed having to do it twice.

PR #160, opened 2022-06-05
add actual duckdb tcp-h tests, fix some bugs after running it

tobymao:


.pylintrc
```diff
@@ -14,7 +14,9 @@ disable=
     too-many-arguments,
     too-many-branches,
     too-many-boolean-expressions,
+    too-many-locals,
```
barakalon, line 17: Similar to our conversations about type annotations/mypy... how often does pylint catch actual issues?

sqlglot/optimizer/expand_multi_table_selects.py
```diff
@@ -1,26 +1,63 @@
...
+            )
             from_.args["expressions"].remove(query)
 
+        if tail:
+            reorder_joins(from_)
+
     return expression
+
+
+def reorder_joins(from_):
```
barakalon, line 45: I think a small comment on what this does would be helpful, since "reorder" doesn't really explain the resulting order.

tests/test_optimizer.py
```diff
@@ -118,87 +201,31 @@ def test_eliminate_subqueries(self):
...
+        )
+
+    def test_tpch_duckdb(self):
+        conn = duckdb.connect()
 
-        self.check_file("tpc-h", optimizer.optimize, schema=schema, pretty=True)
+        def replace_table(sql):
+            expression = sqlglot.parse_one(sql)
+            for table in self.tpch_schema:
+                expression = expression.with_(
```
barakalon, line 214: Why not just load all the tables into duckdb up front?
Something feels strange about mutating the queries we're about to test.

barakalon (approved):
this is awesome

.pylintrc
tobymao: the main benefit is catching unused variables, other than that simple formatting stuff

tests/test_optimizer.py
tobymao: good idea

PR #161, opened 2022-06-06
Transpile levenshtein

th368:
I'm hoping you're still open to accepting random PRs.

This one adds support for the `levenshtein` function, which would be a great help for us within `splink`.

For example:
`sqlglot.transpile("levenshtein(a, b)", read=None, write="presto")` -> `LEVENSHTEIN_DISTANCE(a, b)`
`sqlglot.transpile("levenshtein(a, b)", read=None, write="sqlite")` -> `EDITDIST3(a, b)`

=============================================================================
For completeness, here are my notes and links for available **levenshtein** functions within your supported SQL dialects:

* Duckdb - Levenshtein
* Hive - Levenshtein
* MySQL - Doesn’t appear to be implemented
* Postgres - Levenshtein
* Presto - Levenshtein_distance
* Spark - Levenshtein
* SQLite - under the alias editdist3

Duckdb:
https://duckdb.org/docs/sql/functions/char
Hive:
https://sparkbyexamples.com/apache-hive/hive-built-in-string-functions-with-examples/
Postgres:
Through the following extension - https://towardsdatascience.com/fuzzy-matching-with-levenshtein-and-postgresql-ed66cad01c03
SQLite:
See chapter 8
https://sqlite.org/spellfix1.html

### DIFF
--- a/sqlglot/dialects.py
+++ b/sqlglot/dialects.py
@@ -182,14 +182,6 @@ def _unnest_to_explode_sql(self, expression):
     return self.join_sql(expression)
 
 
-def _levenshtein_sql(self, expression, levenshtein_function):
-
-    expr = expression.args.get("expression")
-    this = self.sql(expression, "this")
-    expr = ", ".join([self.sql(e) for e in expr])
-    return f"{levenshtein_function}({this}, {expr})"
-
-
 def _struct_extract_sql(self, expression):
     this = self.sql(expression, "this")
     struct_key = self.sql(expression, "expression").replace(self.quote, self.identifier)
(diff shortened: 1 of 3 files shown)


sqlglot/dialects.py
```diff
@@ -182,6 +182,14 @@ def _unnest_to_explode_sql(self, expression):
     return self.join_sql(expression)
 
 
+def _levenshtein_sql(self, expression, levenshtein_function):
```
tobymao, line 185: you don’t need this lambda

sqlglot/expressions.py
```diff
@@ -1697,6 +1697,11 @@ class Length(Func):
     pass
 
 
+class Levenshtein(Func):
+    arg_types = {"this": True, "expression": False}
+    is_var_len_args = True
```
tobymao, line 1702: why do you have var len args set? is it not always a and b?

sqlglot/dialects.py
th368: Removed and replaced with some basic lambda funcs

sqlglot/expressions.py
th368: rm'd, thanks

PR #163, opened 2022-06-06
feat: Scope.selectable_references()

barakalon:
references... selectables... referenced_selectables... selectable_references... 
The names might be getting confusing.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -637,7 +637,7 @@ class Subqueryable:
         instance = _maybe_copy(self, copy)
         return Subquery(
             this=instance,
-            alias=to_identifier(alias),
+            alias=TableAlias(this=to_identifier(alias)),
         )
 
     @property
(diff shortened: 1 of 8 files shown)


barakalon:
Some refactoring:

- selectables -> sources
- referenced_selectables -> selected_sources
- columns, tables, derived_tables, subqueries, ctes -> *_nodes (plus this is searched for at access time... might be slightly less efficient, but it should be hard to screw up)
- references -> columns
- correlations -> columns_from_subqueries
- external_references -> columns_referencing_outer_sources
- selectable_references -> source_columns
-

barakalon:
- columns_referencing_outer_sources -> external_columns
- Turns out I can just remove the column_nodes property - its only used by the columns property, so I just consolidated them.
- I removed all the _nodes suffixes.

sqlglot/expressions.py
```diff
@@ -637,7 +637,7 @@ def subquery(self, alias=None, copy=True):
         instance = _maybe_copy(self, copy)
         return Subquery(
             this=instance,
-            alias=to_identifier(alias),
+            alias=TableAlias(this=to_identifier(alias)),
```
tobymao, line 640: can you do alias_(alias, table=True)

tobymao (approved):

sqlglot/expressions.py
barakalon: Mm I don't think so - we need to return a Subquery instance.

PR #165, opened 2022-06-07
[WIP] cnf can be too expensive so we need a backup

tobymao:
compute normalization distance as a heuristic for how expensive it will
be. thus, not all where statements will be in CNF, we need another
strategy.

additionally move all predicate mutations to pushdown predicates and out
of decorrelation and expansion.


sqlglot/expressions.py
```diff
@@ -580,6 +578,46 @@ def is_int(self):
...
 
+    @property
+    def kind(self):
+        return self.text("kind").upper()
+
+    def on(self, *expressions, append=True, dialect=None, parser_opts=None, copy=True):
+        """
+        Append to or set the ON expressions.
+
+        Example:
```
barakalon, line 589: Example copy and pasted from somewhere - needs to be updated

sqlglot/expressions.py
```diff
@@ -580,6 +578,46 @@ def is_int(self):
...
+            *expressions,
+            instance=self,
+            arg="on",
+            append=append,
+            dialect=dialect,
+            parser_opts=parser_opts,
+            copy=copy,
+        )
+
+        if join.kind == "CROSS":
```
barakalon, line 616: What's this about? Should this be `if join.kind == "CROSS" and not expressions`?

sqlglot/optimizer/normalize.py
```diff
@@ -1,9 +1,9 @@
 from sqlglot.helper import while_changing
-from sqlglot.optimizer.simplify import simplify
+from sqlglot.optimizer.simplify import simplify, compare_and_prune, flatten
 import sqlglot.expressions as exp
 
 
-def normalize(expression, dnf=False):
+def normalize(expression, dnf=False, max_distance=128):
```
barakalon, line 6: `max_distance` deserves a spot in the docstring

sqlglot/optimizer/normalize.py
```diff
@@ -20,11 +20,65 @@ def normalize(expression, dnf=False):
...
+        int: difference
+    """
+    return sum(_predicate_lengths(expression, dnf)) - (
+        len(list(expression.find_all(exp.Connector))) + 1
+    )
+
+
+def _predicate_lengths(expression, dnf):
+    """
+    Returns a list of predicates lengths when expanded to normalized form.
```
barakalon, line 63: nit: "predicate lengths"

sqlglot/expressions.py
tobymao: if the current join is a cross join, and you call this method it is no longer a cross join

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -18,58 +20,116 @@ def pushdown_predicates(expression):
...
 
-            if len(sources) != 1:
-                continue
+        if cnf_like:
+            for predicate in predicates:
+                for node in nodes_for_predicate(predicate, scope).values():
+                    if isinstance(node, exp.Join):
+                        predicate.replace(exp.TRUE)
+                        node.on(predicate, copy=False)
+                        break
```
barakalon, line 47: Why break?

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -18,58 +20,116 @@ def pushdown_predicates(expression):
...
+                    if isinstance(node, exp.Select):
+                        predicate.replace(exp.TRUE)
+                        node.where(replace_aliases(node, predicate), copy=False)
+        else:
+            pushdown = set()
```
barakalon, line 52: This block is very hard to follow. 

I know complexity is complex... but maybe we can throw some comments here and there?

sqlglot/optimizer/scope.py
```diff
@@ -154,18 +154,22 @@ def selected_sources(self):
...
-            referenced_names.append(derived_table.alias)
+            referenced_names.append((derived_table.alias, derived_table.this))
 
         result = {}
 
-        for name in referenced_names:
+        for name, node in referenced_names:
             if name in self.sources:
-                result[name] = self.sources[name]
+                result[name] = (node, self.sources[name])
```
barakalon, line 172: We need to update this methods docstring

tests/fixtures/optimizer/pushdown_predicates.sql
```diff
@@ -1,17 +1,20 @@
...
-
-SELECT x.a FROM (SELECT * FROM x) AS x JOIN y WHERE x.a = 1 AND x.b = 1 OR y.a = 1;
-SELECT x.a FROM (SELECT * FROM x) AS x JOIN y WHERE x.a = 1 AND x.b = 1 OR y.a = 1;
-
-SELECT x.a FROM (SELECT x.a AS a, x * 1 AS c FROM x) AS x WHERE x.c = 1;
-SELECT x.a FROM (SELECT x.a AS a, x * 1 AS c FROM x WHERE x * 1 = 1) AS x WHERE TRUE;
-
-SELECT x.a FROM (SELECT x.a AS a, x * 1 AS c FROM x) AS x WHERE x.c = 1 or x.c = 2;
-SELECT x.a FROM (SELECT x.a AS a, x * 1 AS c FROM x WHERE x * 1 = 1 OR x * 1 = 2) AS x WHERE TRUE;
+--SELECT x.a AS a FROM (SELECT x.a FROM x AS x) AS x JOIN y WHERE x.a = 1 AND x.b = 1 AND y.a = 1;
```
barakalon, line 1: Commented on purpose?

tests/fixtures/optimizer/tpc-h/tpc-h.sql
```diff
@@ -978,39 +977,49 @@ LIMIT 100;
 --------------------------------------
 -- TPC-H 19
 --------------------------------------
---select
```
barakalon, line 981: Commented on purpose?

barakalon (approved):
Oomph, this is nuts.

sqlglot/expressions.py
barakalon: Ah yeah, doy.

sqlglot/optimizer/pushdown_predicates.py
tobymao: we only want to push it down to one join

barakalon: I know some folks are anti-comments...
But I think little comments like this are helpful, especially with gnarly code like this.

tests/fixtures/optimizer/pushdown_predicates.sql
tobymao: nope

PR #166, opened 2022-06-08
feat: column builder

barakalon:
I see the a lot in Minerva: `"{table}.{col}"`. That's not safe if either values have dots.


tobymao (approved):

barakalon:
Actually, I think this is fine:
`f'"{table}"."{col}"'`

nolamesa:
what changed in your reasoning?

PR #172, opened 2022-06-14
Toby/planner

tobymao:
i can run tpc-h 1 and the resultsa re the same as duckdb!


sqlglot/expressions.py
```diff
@@ -2285,6 +2285,18 @@ def subquery(expression, alias=None, dialect=None, **opts):
     return Select().from_(expression, dialect=dialect, **opts)
 
 
+def column(col, table=None):
```
nolamesa, line 2288: flagging to @barakalon who wanted to add something similar

sqlglot/helper.py
```diff
@@ -97,3 +97,14 @@ def visit(node, visited):
         visit(node, set())
 
     return result
+
+
+class reverse_key:
```
nolamesa, line 102: is this used anywhere?

nolamesa (approved):
did not look very closely but looks reasonable. left a couple minor comments

sqlglot/helper.py
tobymao: it will be used for order by desc

barakalon (commented):
Very cool!

Three high level comments:
1. Some of the code gets hard to follow. I think this needs a round of cleanup and comments.
2. The executor feels more like an _example_ of interpreting plans. I'm worried about shipping it as part of the package. Maybe we should move it into a root `examples` directory (or something like that)? 
3. The planner seems generally useful, moreso than the executor. I wonder how we can create more comprehensive tests?

PR #173, opened 2022-06-15
add expansion of ambiguous tables to make physical execution easier

tobymao:


sqlglot/optimizer/expand_ambiguous_tables.py
```diff
@@ -0,0 +1,24 @@
+import sqlglot.expressions as exp
+from sqlglot import select, subquery
+from sqlglot.errors import OptimizeError
+from sqlglot.optimizer.scope import traverse_scope
+
+
+def expand_ambiguous_tables(expression):
+    for scope in traverse_scope(expression):
+        if len(scope.sources) == 1:
```
barakalon, line 9: selected_sources?

sqlglot/optimizer/expand_ambiguous_tables.py
```diff
@@ -0,0 +1,24 @@
+import sqlglot.expressions as exp
+from sqlglot import select, subquery
+from sqlglot.errors import OptimizeError
+from sqlglot.optimizer.scope import traverse_scope
+
+
+def expand_ambiguous_tables(expression):
```
barakalon, line 7: Why is the table "ambiguous"?

barakalon: Hm... yeah, its a hard name.
How about.. isolate_table_selects

PR #174, opened 2022-06-17
Toby/joins

tobymao:


sqlglot/executor/context.py
```diff
@@ -0,0 +1,59 @@
...
+    Execution context for sql expressions.
+
+    Context is used to hold relevant data tables which can then be queried on with eval.
+
+    References to columns can either be scalar or vectors. When set_row is used, column references
+    evaluate to scalars while set_range evaluates to vectors. This allows convenient and efficient
+    evaluation of aggregation functions.
+    """
+
+    def __init__(self, tables, env=None):
```
barakalon, line 15: Can you add some descriptions of these parameters to the docstring?

sqlglot/executor/table.py
```diff
@@ -0,0 +1,98 @@
+class Table:
+    def __init__(self, *columns, _columnar=True, **table):
```
barakalon, line 2: lol this is a crazy signature.

sqlglot/executor/table.py
```diff
@@ -0,0 +1,98 @@
...
+        self.range = range(0)
+
+    def __len__(self):
+        return len(self.range)
+
+    def __getitem__(self, column):
+        return (self.table[column][i] for i in self.range)
+
+
+class ColumnarReader:
```
barakalon, line 77: These two interfaces are a bit confusing.

Seems like callers are expected to set `row`. But `ColumnarReader.row` is an index and `RowReader.row` is... something else?

Maybe some docstrings would help.

barakalon (approved):
Sweet, this feels much better.

Some of the more fundamental components (e.g. context, the readers, etc.) are hard to grok by themselves. I.e. to understand the code, you have to read the executor. Docstrings?

sqlglot/executor/table.py
tobymao: need to think about this and maybe delete these classes, we'll see

PR #177, opened 2022-06-21
Add support for 'SELECT TOP <n>' as an alternative syntax for LIMIT.

ptallada:
Hi,

I think I've got it ;)

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -107,6 +107,7 @@ class Parser:
         TokenType.SHOW,
         TokenType.TABLE_SAMPLE,
         TokenType.TEMPORARY,
+        TokenType.TOP,
         TokenType.TRUNCATE,
         TokenType.TRUE,
         TokenType.UNBOUNDED,
@@ -795,13 +796,22 @@ class Parser:
             )
 
         if isinstance(this, exp.Subquery) or self._match(TokenType.SELECT):
+            hint = self._parse_hint()
+            all_ = self._match(TokenType.ALL)
+            distinct = self._match(TokenType.DISTINCT)
+            if all_ and distinct:
+                self.raise_error("Cannot specify both ALL and DISTINCT after SELECT")
+
+            limit = self._parse_limit(top=True)
+            expressions = self._parse_csv(
+                lambda: self._parse_annotation(self._parse_expression())
+            )
+
             this = self.expression(
                 exp.Select,
-                hint=self._parse_hint(),
-                distinct=self._match(TokenType.DISTINCT),
-                expressions=self._parse_csv(
-                    lambda: self._parse_annotation(self._parse_expression())
-                ),
+                hint=hint,
+                distinct=distinct,
+                expressions=expressions,
                 **{
                     "from": this or self._parse_from(),
                     "laterals": self._parse_laterals(),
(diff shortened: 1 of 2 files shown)


ptallada:
I don't understand the lint error :/

sqlglot/parser.py
```diff
@@ -795,13 +795,19 @@ def _parse_select(self):
             )
 
         if isinstance(this, exp.Subquery) or self._match(TokenType.SELECT):
+            hint=self._parse_hint()
+            _=self._match(TokenType.ALL)
```
barakalon, line 799: Should we be matching ALL _or_ DISTINCT?

barakalon (commented):
Nice!

One primary concern: This parses TOP, but it can't generate TOP. 

Also: add tests. If you add one [here](https://github.com/tobymao/sqlglot/blob/main/tests/fixtures/identity.sql), I think you'll see the above problem.

Happy to talk through ideas on how we can handle that!

barakalon:
That's not a lint error - it's a format error. Run `./format_code.sh`.

tobymao:
it may not be necessary  to generate top

barakalon:
Not generating TOP feels like a bandaid for parsing SQLServer syntax. I think a true _transpiler_ would be able to go to _and_ from different dialects.

barakalon:
Yeah, offset makes this even messier

sqlglot/parser.py
ptallada: Yes, it is one or the other, I'll try to fix that

ptallada:
> Nice!
> 
> One primary concern: This parses TOP, but it can't generate TOP.
Yes, although for me it is not a requirement...

> Also: add tests. If you add one [here](https://github.com/tobymao/sqlglot/blob/main/tests/fixtures/identity.sql), I think you'll see the above problem.
On it :)

> Happy to talk through ideas on how we can handle that!
If it is not very difficult, I can invest a pair of hours on it :)

ptallada:
> https://stackoverflow.com/questions/603724/how-to-implement-limit-with-sql-server
> 
> offset fetch

WTF, I was not aware of this mess. My requirement is to transpile ADQL (not MSSQL), which seems to be based on the former (and easier) MSSQL syntax.

sqlglot/parser.py
```diff
@@ -811,7 +820,7 @@ def _parse_select(self):
                     "having": self._parse_having(),
                     "qualify": self._parse_qualify(),
                     "order": self._parse_order(),
-                    "limit": self._parse_limit(),
+                    "limit": limit if limit else self._parse_limit(),
```
tobymao, line 823: limit or self._parse_limit()

sqlglot/parser.py
```diff
@@ -1039,6 +1048,11 @@ def _parse_limit(self):
             return None
         return self.expression(exp.Limit, this=self._parse_number())
 
+    def _parse_top(self):
```
tobymao, line 1051: let's just delete this and add an arg to _parse_limit which is 

def _parse_limit(self, top=False):
self._match(TokenType.TOP if top else TokenType.LIMIT)

sqlglot/tokens.py
```diff
@@ -373,6 +374,7 @@ class Tokenizer:
         "TEMPORARY": TokenType.TEMPORARY,
         "THEN": TokenType.THEN,
         "TIME": TokenType.TIME,
+        "TOP": TokenType.TOP,
```
tobymao, line 377: and TokenType.TOP to the list of id_vars in parser.py

tobymao:
> Ok, I am a bit lost on how to implement TOP generation, any hints?

you probably need to add argument to Generator to support TOP generation, as i said before, i'm fine if we don't add it

tobymao (approved):

PR #184, opened 2022-06-28
fix: Some optimizer speedups

barakalon:
🏎️ 

@tobymao

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -90,7 +90,7 @@ class Expression:
             value: value to set the arg to.
         """
         self.args[arg] = value
-        self._set_parent(arg=value)
+        self._set_parent(**{arg: value})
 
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -90,7 +90,7 @@ def set(self, arg, value):
             value: value to set the arg to.
         """
         self.args[arg] = value
-        self._set_parent(arg=value)
+        self._set_parent(**{arg: value})
```
tobymao, line 93: what’s up with this one?

barakalon: This is fixing a bug.

self._set_parent(arg=value) is different than self._set_parent(**{arg: value})

tobymao: oh wow it is

tobymao (approved):

Issue #203, opened 2022-07-08
Support for DateType INT64 (Bigquery)


bobvanratingen:
Would be nice when INT64 is supported 

```
SELECT
  CAST(field AS INT64) AS alias,
FROM
  `table`
```

yingw787:
I can see if I can implement this!

PR #204, opened 2022-07-09
feat(gh-203): Add support for BigQuery.

yingw787:
# Overview

- Mappings between INT64 and INT, BIGINT, SMALLINT, TINYINT with unit tests
- Mappings between FLOAT64 and FLOAT, DOUBLE with unit tests
- Mappings between NUMERIC and DECIMAL with unit tests
- Mappings between STRING and TEXT with unit tests

Haven't added support for other types of casting (e.g. CHAR/VARCHAR/BYTES/TIMESTAMPTZ) due to additional overhead.

Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types

### DIFF
--- a/tests/test_dialects.py
+++ b/tests/test_dialects.py
@@ -364,6 +364,13 @@ class TestDialects(unittest.TestCase):
             write="bigquery",
         )
 
+        self.validate(
+            "SELECT CAST(`a`.`b` AS TEXT) FROM foo",
+            "SELECT CAST(`a`.`b` AS STRING) FROM foo",
+            read="bigquery",
+            write="bigquery",
+        )
+
     def test_postgres(self):
         self.validate(
             "SELECT CAST(`a`.`b` AS DOUBLE) FROM foo",
--- a/tests/test_dialects.py
+++ b/tests/test_dialects.py
@@ -307,6 +307,14 @@ class TestDialects(unittest.TestCase):
             write="starrocks",
         )
 
+    def test_bigquery(self):
+        self.validate(
+            "SELECT CAST(`a`.`b` AS INT) FROM foo",
+            "SELECT CAST(`a`.`b` AS INT) FROM foo",
+            read="bigquery",
+            write="bigquery"
+        )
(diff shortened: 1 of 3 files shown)


tests/test_dialects.py
```diff
@@ -307,6 +307,70 @@ def test_starrocks(self):
             write="starrocks",
         )
 
+    def test_bigquery(self):
+        self.validate(
+            "SELECT CAST(`a`.`b` AS INT) FROM foo",
+            "SELECT CAST(`a`.`b` AS INT64) FROM foo",
+            read="bigquery",
```
tobymao, line 314: get rid of all reads

tests/test_dialects.py
```diff
@@ -307,6 +307,70 @@ def test_starrocks(self):
...
+            write="bigquery",
+        )
+
+        self.validate(
```
tobymao, line 318: you didn’t test any inverses can you read bigquery int64 and output generic?

tests/test_dialects.py
yingw787: done

tests/test_dialects.py
yingw787: That's a good point, I can add that, had some questions:

- What's generic `sqlglot` SQL? Just leave `write=` unspecified?
- When I add `read='bigquery'` using ```"SELECT CAST(`a`.`b` AS INT64) FROM foo",```, I get this error:

```sql
sqlglot.errors.ParseError: Expected TYPE after CAST. Line 1, Col: 24.
  SELECT CAST(`a`.`b` AS INT64) FROM foo
```

I'm not sure how to modify my Dialect class to support the inverse operation, all of the `type_mapping` class attributes have the generic `exp.DataType.Type.*` on the left-hand side and the string on the right-hand side. Do you have any suggestions for me?

tobymao: yes, just leave write unspecified, it should automatically support the inverse

PR #211, opened 2022-07-11
make it easier to subclass parsers

tobymao:


kelsin (approved):

sqlglot/dialects/dialect.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+
+class _Dialect(type):
+    classes = {}
+
+    @classmethod
+    def import_dialect(cls, key):
+        if key not in cls.classes:
+            importlib.import_module(f"sqlglot.dialects.{key}")
```
izeigerman, line 19: I don't like the idea of hardcoding the sqlglot.dialects. here? Can't we just use the relative path instead?
```
importlib.import_module(f".{key}")

File: README.md [hunk 2]
```diff
@@ -216,84 +216,24 @@ SELECT
 FROM users
 ```
 
-### Customization
```
izeigerman, line 219: Instead of deleting this section should we modify it to show how one can create a custom dialect?

sqlglot/dialects/tableau.py
```diff
@@ -0,0 +1,28 @@
+from sqlglot import exp
+from sqlglot.dialects.dialect import Dialect
+from sqlglot.generator import Generator
+
+
+def _if_sql(self, expression):
```
izeigerman, line 6: Ditto about private functions

izeigerman (approved):
Left a few comments. LGTM once addressed.

nolamesa (approved):

sqlglot/dialects/dialect.py
```diff
@@ -0,0 +1,250 @@
...
+            this=list_get(args, 0),
+            format=Dialect[dialect].format_time(
+                list_get(args, 1) or (Dialect[dialect].time_format if default else None)
+            ),
+        )
+
+    return _format_time
+
+
+def no_sort_array(self, expression):
```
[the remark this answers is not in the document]
tobymao: yes

sqlglot/dialects/hive.py
```diff
@@ -0,0 +1,255 @@
...
+    unnest_to_explode_sql,
+    struct_extract_sql,
+    no_trycast_sql,
+)
+from sqlglot.generator import Generator
+from sqlglot.helper import csv, list_get
+from sqlglot.parser import Parser
+
+
+def _parse_map(args):
```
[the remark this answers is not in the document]
tobymao: me too but, that's a syntax error because i've defined these transforms in the class scope

sqlglot/dialects/duckdb.py
```diff
@@ -0,0 +1,91 @@
...
+    approx_count_distinct_sql,
+    rename_func,
+    no_tablesample_sql,
+)
+from sqlglot.generator import Generator
+from sqlglot.helper import list_get
+from sqlglot.parser import Parser
+
+
+def _unix_to_time(self, expression):
```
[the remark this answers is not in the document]
tobymao: same, can't move it

sqlglot/dialects/dialect.py
tobymao: no, you can't do that, i'd have to include the package

izeigerman: Can we use `sqllib.dialect.__name__` then?

nolamesa: my suggestion was to always import at init and drop the lazy load

PR #215, opened 2022-07-13
DateTimePart + StarExcept (optimizer)

bobvanratingen:
I don't know if this datetime component could be implemented more simply.  As long as the datetime part is not a column.


tobymao:
closed in favor of 4f4e6a8

PR #223, opened 2022-07-14
Introduce ability to calculate changes between two expression trees (diff)

izeigerman:
Previous:

> The implementation relies on a variance of the Myers algorithm with additional heuristics. To apply the Myers algorithm both input expression trees are first unfolded into their list representation using the topological order. 
> 
> To make the output cleaner we apply the following heuristics when checking the equality of individual nodes:
> 1. If both nodes are a `Column`, `Literal`, `Table` or `TableAlias` we do full equality check.
> 2. For `Anonymous` and `DataType` nodes we only check the "this" attribute for equality.
> 3. For all other nodes we only check whether they have a matching type. 
> 
> Thus if the number of arguments changed for a particular node the algorithm doesn't interpret this as a change to this particular node. Instead we expect this change to be captured in children nodes.
> 
> Additionally to make the algorithm's output more stable and predictable we sort the arguments for node types for which the order of arguments doesn't impact semantics. Such node types include `Select` (eg. the list of projections and group by), `In`, `With`, `And` and `Or`. This detail does positively impact the algorithm's behavior, but I'm open to feedback if you think that it makes the algorithm less "correct".

The implementation of the Change Distiller algorithm described by Beat Fluri and Martin Pinzger in their paper https://ieeexplore.ieee.org/document/4339230, which in turn is based on the algorithm by Chawathe et al. described in http://ilpubs.stanford.edu:8090/115/1/1995-46.pdf.

### DIFF
--- a/sqlglot/diff.py
+++ b/sqlglot/diff.py
@@ -142,8 +142,19 @@ class ChangeDistiller:
         leaves_matching_set = self._compute_leaf_matching_set()
         matching_set = leaves_matching_set.copy()
 
-        for source_node_id in self._unmatched_source_nodes.copy():
-            for target_node_id in self._unmatched_target_nodes:
+        ordered_unmatched_source_nodes = {
+            id(n[0]): None
+            for n in self._source.bfs()
+            if id(n[0]) in self._unmatched_source_nodes
+        }
+        ordered_unmatched_target_nodes = {
+            id(n[0]): None
+            for n in self._target.bfs()
+            if id(n[0]) in self._unmatched_target_nodes
+        }
+
+        for source_node_id in ordered_unmatched_source_nodes.keys():
+            for target_node_id in ordered_unmatched_target_nodes.keys():
                 source_node = self._source_index[source_node_id]
                 target_node = self._target_index[target_node_id]
                 if _is_same_label(source_node, target_node):
@@ -174,6 +185,7 @@ class ChangeDistiller:
                         matching_set.add((source_node_id, target_node_id))
                         self._unmatched_source_nodes.remove(source_node_id)
                         self._unmatched_target_nodes.remove(target_node_id)
+                        ordered_unmatched_target_nodes.pop(target_node_id, None)
                         break
 
         return matching_set
--- a/sqlglot/diff.py
+++ b/sqlglot/diff.py
@@ -1,10 +1,15 @@
 from collections import defaultdict
 from dataclasses import dataclass
 from sqlglot.expressions import (
+    Boolean,
+    DataType,
     Expression,
     Identifier,
     Join,
     Literal,
+    Placeholder,
+    Null,
+    Star,
 )
 from sqlglot.helper import ensure_list
 
@@ -76,7 +81,15 @@ def diff(source, target):
     return ChangeDistiller().diff(source, target)
 
 
-LEAF_EXPRESSION_TYPES = (Identifier, Literal)
+LEAF_EXPRESSION_TYPES = (
+    Boolean,
+    DataType,
+    Identifier,
+    Literal,
+    Null,
+    Placeholder,
+    Star,
+)
 
 
 class ChangeDistiller:
--- a/sqlglot/diff.py
+++ b/sqlglot/diff.py
@@ -8,15 +8,19 @@ from sqlglot.expressions import (
     Expression,
     Group,
     Identifier,
+    Join,
     In,
     Literal,
-    Select,
     Table,
     TableAlias,
     With,
 )
 from sqlglot.errors import DiffError
 from sqlglot.helper import ensure_list
+
+
+EXCLUDED_EXPRESSION_TYPES = {Identifier}
+SORT_ARGS_EXPRESSION_TYPES = {Connector, Group, In, With}
 
 
 @dataclass
@@ -35,13 +39,21 @@ class Keep:
     target: Expression
 
 
-def diff(source, target):
+def diff(
+    source,
+    target,
+    excluded_expression_types=None,
+    sort_args_expression_types=None,
+):
     """
     Returns changes between the source and the target expressions.
 
     The underlying logic relies on a variance of the Myers diff algorithm with a custom heuristics for checking
     equality between nodes. Provided the Myers algorithm operates on lists, the source and the target expressions
     are first unfolded into their list representations using the topological order.
+
+    NOTE: this functionality is largely based on heuristics. It does a pretty good job at estimating which AST nodes have
+    been impacted but neither accuracy nor optimality of this estimate are guaranteed. Use it at your own risk.
 
     Examples:
         >>> diff(parse_one("a + b"), parse_one("a + c"))
@@ -60,13 +72,27 @@ def diff(source, target):
 
     Args:
         source (sqlglot.Expression): the source expression.
-        target (sqlglot.Expression): the target expression against which the diff should be calculated.
+        target (sqlglot.Exexcluded_expression_typespression): the target expression against which the diff should be calculated.
+        excluded_expression_types (optional): the set of AST node types which should be excluded from the comparison.
+        sort_args_expression_types (optional): the set of AST node types for which arguments should be sorted. This parameter
+            can be used to tweak the accuracy of the algorithm. For example if a calling code expects the two ASTs to be mostly
+            the same and the position of a majority of nodes is expected to remain unchanged, it may make sense to omit sorting
+            for some nodes.
     Returns:
         the list of Insert, Remove and Keep objects for each node in the source and the target expression trees.
         This list represents a sequence of steps needed to transform the source expression tree into the target one.
     """
-    source_expression_list = _expression_to_list(source)
-    target_expression_list = _expression_to_list(target)
+    if excluded_expression_types is None:
+        excluded_expression_types = EXCLUDED_EXPRESSION_TYPES
+    if sort_args_expression_types is None:
+        sort_args_expression_types = SORT_ARGS_EXPRESSION_TYPES
+
+    source_expression_list = _expression_to_list(
+        source, excluded_expression_types, sort_args_expression_types
+    )
+    target_expression_list = _expression_to_list(
+        target, excluded_expression_types, sort_args_expression_types
+    )
     return _myers_diff(source_expression_list, target_expression_list)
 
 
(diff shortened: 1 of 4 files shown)


izeigerman:
I'd also appreciate if someone finds time to play with this and try it on different queries.

sqlglot/diff.py
```diff
@@ -0,0 +1,162 @@
...
+    TableAlias,
+    With,
+)
+from sqlglot.errors import DiffError
+from sqlglot.helper import ensure_list
+
+
+@dataclass
+class Insert:
+    expr: Expression
```
tobymao, line 23: minor, but can we just call it expression: Expression 

and for keep source / target, you have the types there anyways, so no need to say expr_src

tobymao: yes, just to source and target, it's clear it's an expression

sqlglot/diff.py
```diff
@@ -0,0 +1,162 @@
...
+        _normalize_args_order(arg)
+
+    if isinstance(expr, (Group, In, Select, With)):
+        for k, a in expr.args.items():
+            if isinstance(a, list) and all(
+                isinstance(a_item, Expression) for a_item in a
+            ):
+                expr.args[k] = sorted(a, key=Expression.sql)
+    elif isinstance(expr, Connector):
+        args = [expr.this, expr.args.get("expression")]
```
tobymao, line 146: should you use flatten() for this?

look at simplify.py uniq_sort

sqlglot/diff.py
```diff
@@ -0,0 +1,162 @@
...
+        args = sorted(args, key=Expression.sql)
+        expr.args["this"], expr.args["expression"] = args
+
+    return expr
+
+
+def _expression_only_args(expr):
```
tobymao, line 153: i wonder if unnest_operands will help you, it gives you the added benefit of getting rid of parenthesis

sqlglot/diff.py
```diff
@@ -0,0 +1,162 @@
...
+    return type(expr_src) is type(expr_tgt)
+
+
+EXCLUDED_EXPR_TYPES = (Identifier,)
+
+
+def _expr_to_list(expr):
+    return [
+        n[0]
+        for n in _normalize_args_order(expr.copy()).walk()
```
tobymao, line 130: do we care about the order? i find DFS usually gives better results, you can call dfs explicitly instead of walk

sqlglot/diff.py
izeigerman: I was thinking about `flatten` but I don't think it's applicable here. I need to preserve the original tree structure, while flatten breaks it for me. So if I use `flatten` I'd also need a way to recover the original tree structure afterwards. 

At the same time perhaps recovering can be straightforward since it's just folding all items back into the `And` / `Or`.

sqlglot/diff.py
```diff
@@ -0,0 +1,162 @@
...
+    if isinstance(expr, (Group, In, Select, With)):
+        for k, a in expr.args.items():
+            if isinstance(a, list) and all(
+                isinstance(a_item, Expression) for a_item in a
+            ):
+                expr.args[k] = sorted(a, key=Expression.sql)
+    elif isinstance(expr, Connector):
+        args = [expr.this, expr.args.get("expression")]
+        args = sorted(args, key=Expression.sql)
+        expr.args["this"], expr.args["expression"] = args
```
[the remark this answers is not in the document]
izeigerman: Will do. Though keep in mind that in this case we just change the order and the reference to a parent shouldn't be impacted.

sqlglot/diff.py
tobymao: i wouldn't change the orders of and and ors, but any chained ands or chained ors can be mutated, you can get the expression back by reparsing it, or using exp.condition()

sqlglot/diff.py
izeigerman: Will try DFS and see how it impacts the behavior.

sqlglot/diff.py
izeigerman: https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py#L243

As far as I can tell `unnest_operands` doesn't handle arguments that are themselves lists (eg. `expressions`, etc.). Or am I missing something?

izeigerman: I actually expect it to fail if the list argument is encountered.

izeigerman: yep, and it does!
```
>>> parse_one("a IN (1, 2)").unnest_operands()
  File "/Users/izeigerman/workspace/sqlglot/sqlglot/expressions.py", line 243, in unnest_operands
    return tuple(arg.unnest() for arg in self.args.values() if arg)
  File "/Users/izeigerman/workspace/sqlglot/sqlglot/expressions.py", line 243, in <genexpr>
    return tuple(arg.unnest() for arg in self.args.values() if arg)
AttributeError: 'list' object has no attribute 'unnest'
```

tobymao: gotcha

sqlglot/diff.py
izeigerman: Fixed this

sqlglot/diff.py
izeigerman: Added this

sqlglot/diff.py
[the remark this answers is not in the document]
izeigerman: Done that

sqlglot/diff.py
izeigerman: I tried DFS but it makes the order in which the changes appear in the resulting least really confusing and counterintuitive. I'd rather keep BFS.

sqlglot/diff.py
```diff
@@ -0,0 +1,162 @@
...
+            if x >= len(expr_list_src) and y >= len(expr_list_tgt):
+                return history
+
+            front[k] = x, history
+
+    raise DiffError("Unexpected state")
+
+
+def _is_expr_unchanged(expr_src, expr_tgt):
+    equals_exprs = [Literal, Column, Table, TableAlias]
```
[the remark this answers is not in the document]
izeigerman: Done that

tests/test_diff.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+
+        self._validate_delta_only(
+            diff(
+                parse_one("SELECT a FROM table_one"),
+                parse_one("SELECT a FROM table_two"),
+            ),
+            [
+                Remove(
+                    Table(
```
tobymao, line 39: i have helper methods now for this to make the tests cleaner

exp.table("table_one")

tobymao (approved):

tests/test_diff.py
izeigerman: aaah, nice 👍 will use that

sqlglot/diff.py
```diff
@@ -35,14 +39,22 @@ class Keep:
     target: Expression
 
 
-def diff(source, target):
+def diff(
+    source,
+    target,
+    excluded_expression_types=None,
```
tobymao, line 45: sorry if i'm a minimalist but what about just exclude_types and sort_types

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
+from collections import defaultdict
+from dataclasses import dataclass
+from sqlglot.expressions import (
```
tobymao, line 3: as this gets bigger, i prefer to do from sqlglot import exp or import from sqlglot import expressions as exp

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+    def __init__(self, f=0.6, t=0.6):
+        self.f = f
+        self.t = t
+
+    def diff(self, source, target):
+        self._source = source
+        self._target = target
+        self._source_index = {id(n[0]): n[0] for n in source.bfs()}
+        self._target_index = {id(n[0]): n[0] for n in target.bfs()}
+        self._unmatched_source_nodes = set(self._source_index.keys())
```
tobymao, line 111: set(self._source_index)

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+            for n in self._source.bfs()
+            if id(n[0]) in self._unmatched_source_nodes
+        }
+        ordered_unmatched_target_nodes = {
+            id(n[0]): None
+            for n in self._target.bfs()
+            if id(n[0]) in self._unmatched_target_nodes
+        }
+
+        for source_node_id in ordered_unmatched_source_nodes.keys():
```
tobymao, line 169: keys is unnecessary in iteration

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+                        self._unmatched_target_nodes.remove(target_node_id)
+                        ordered_unmatched_target_nodes.pop(target_node_id, None)
+                        break
+
+        return matching_set
+
+    def _compute_leaf_matching_set(self):
+        leaf_matchings = []
+        for source_leaf in _get_leaves(self._source):
+            for target_leaf in _get_leaves(self._target):
```
tobymao, line 209: this looks expensive

maybe cache the source leaves?

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+                if _is_same_label(source_leaf, target_leaf):
```
tobymao, line 210: what is same_label, should it just be called same_type?

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+                    similarity_score = _dice_coefficient(
+                        source_leaf.sql(), target_leaf.sql()
```
tobymao, line 212: i can see why this would be slow

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+
+    return type(source) is type(target)
+
+
+def _dice_coefficient(string_a, string_b, n=2):
+    n_grams_a = _n_grams(string_a.lower(), n)
+    n_grams_b = _n_grams(string_b.lower(), n)
+
+    total_grams = len(n_grams_a) + len(n_grams_b)
+    if total_grams == 0:
```
tobymao, line 250: if not total_grams

sqlglot/diff.py
```diff
@@ -0,0 +1,300 @@
...
+        for a in expression.args.values():
+            args.extend(ensure_list(a))
+    return [a for a in args if isinstance(a, Expression)]
+
+
+def _all_same_type(tpe, *args):
+    return all(isinstance(a, tpe) for a in args)
+
+
+def _lcs(seq_a, seq_b, equal):
```
tobymao, line 282: comment on what lcs stands for?

tobymao, line 282: curious why equal is a lambda, seems like ti may not need to be? you only call _lcs in one place right?

sqlglot/diff.py
izeigerman: Yes, great catch. That's one reason this thing is so slow. There are others unfortunately.

sqlglot/diff.py
izeigerman: Bingo. This is the most expensive part.

sqlglot/diff.py
izeigerman: Fixed

sqlglot/diff.py
izeigerman: Fixed

sqlglot/diff.py
izeigerman: Fixed

sqlglot/diff.py
izeigerman: Fixed this

sqlglot/diff.py
izeigerman: Renamed

sqlglot/diff.py
izeigerman: Introduced caching for bigrams which along with leaf nodes caching led to 5x runtime improvement. There's still room to improve further but I'm hoping to do this in subsequent iterations.

sqlglot/diff.py
izeigerman: Fixed

sqlglot/diff.py
izeigerman: Added a comment

izeigerman: So the LCS (Least Common Subsequence) is a pretty well-known and generic algorithm and I didn't want to inject our business logic directly into it. I felt like having an abstraction over what's considered to be `equal` is more readable and a better segregation of concerns. 

The alternative was passing the reference to the `matching_set` directly here - https://github.com/tobymao/sqlglot/blob/a4e1a54525ac05ade3dbb73253646cca92ba3c76/sqlglot/diff.py#L137.

Let me know what you think.

tobymao: ok that's fine

PR #237, opened 2022-07-17
Fix CAST as DATETIME

bobvanratingen:


tobymao:
can you add a test case in identity.sql?

tobymao:
i'll handle it

PR #244, opened 2022-07-18
_check_union_outputs -> only same number of columns

bobvanratingen:


tobymao:
this is not correct?

tobymao:
ah ok, i wonder if we should just get rid of the check

bobvanratingen:
that's also fine

PR #260, opened 2022-07-21
Unknown column (not in schema)

bobvanratingen:
When you select a column but it's not in the schema. 
Before this the optimizer said "Ambiguous column" and now it says "Unknown column"

Not sure if you want to add this.

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -125,6 +125,7 @@ def _expand_using(scope, schema):
 def _qualify_columns(scope, schema):
     """Disambiguate columns, ensuring each column specifies a source"""
     unambiguous_columns = None  # lazily loaded
+    scope_all_schema_columns = None  # lazily loaded
 
     for column in scope.columns:
         column_table = column.table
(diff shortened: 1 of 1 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -277,6 +283,27 @@ def _get_unambiguous_columns(source_columns):
...
+        source_columns (dict): Mapping of names to source columns
+    Returns:
+        list: all columns from schema
+    """
+    if not source_columns:
+        return []
+
+    columns = [
+        value
+        for nested_list_value in list(source_columns.values())
```
tobymao, line 300: the list is unnecessary

tobymao: can just do

return [
   column
   for columns in source_columns.values()
   for column in columns
]

bobvanratingen: Ow yes. Indeed.

tobymao: i'll fix it

PR #261, opened 2022-07-21
never force qualify group or order

### LINKED ISSUE #258: Ambiguous column error with Window function
I get an ambiguous column error for: **user_id**. 

```
query = """
SELECT
    table_1.user_id,
    table_1.method_id,
FROM
    table_1
LEFT JOIN table_2 ON table_2.user_id = table_1.user_id
QUALIFY
    ROW_NUMBER() OVER(
        PARTITION BY user_id
        ORDER BY
        method_id DESC
    ) = 1
"""

expression = parse_one(query, read="bigquery")

optimize(
    expression,
    schema={
        "table_1": {
            "id": "INT",
            "name": "STRING",
            "type": "STRING",
            "user_id": "INT",
            "method_id": "INT",
            "dummy_id": "INT",
            "start_time": "TIMESTAMP",
            "date_snapshot": "TIMESTAMP",
        },
        "table_2": {
            "id": "INT",
            "name": "STRING",
            "user_id": "INT",
            "dummy_id": "INT",
            "field_from_table2": "STRING",
            "date_snapshot": "TIMESTAMP",
        }
    },
)
```

tobymao:
fixes #258

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -189,21 +189,11 @@ class Scope:
                 for column in scope.external_columns
             ]
 
-            # Expression.named_selects also includes unaliased columns.
-            # In this case, we want to be sure to only include selects that are aliased.
-            aliased_outputs = {
-                e.alias for e in self.expression.expressions if isinstance(e, exp.Alias)
-            }
-
             self._columns = [
                 c
                 for c in columns + external_columns
-                if c.table
-                or not (
-                    # In some dialects, these clauses can reference output columns
-                    c.find_ancestor(exp.Group, exp.Order, exp.Qualify)
-                    and c.name in aliased_outputs
-                )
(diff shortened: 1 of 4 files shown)


barakalon (commented):
Hm, I'm not sure this is correct.

Here's one example:

From https://docs.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16
> The column names referenced in the ORDER BY clause must correspond to either a column or column alias in the select list **or to a column defined in a table specified in the FROM clause without any ambiguities**.

So
```sql
SELECT
  a,
FROM x
ORDER BY b
```

Should probably be:
```sql
SELECT
  x.a AS a,
FROM x AS x
ORDER BY x.b
```

I think there are other cases as well (e.g. more complex expressions in these clauses).

tobymao:
yes, but the user inputs the first query, if it's wrong, we shouldn't "fix it"

PR #265, opened 2022-07-23
Add support for Spark's filter function

eakmanrq:
`ARRAY_FILTER` is called `FILTER` in spark. So I did a rename in the generator for Spark so it refers to the function correctly.

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -121,9 +121,7 @@ class Presto(Dialect):
 
     class Tokenizer(Tokenizer):
         KEYWORDS = {
-            k: v
-            for k, v in Tokenizer.KEYWORDS.items()
-            if v != TokenType.FILTER
+            k: v for k, v in Tokenizer.KEYWORDS.items() if v != TokenType.FILTER
         }
 
     class Parser(Parser):
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -13,6 +13,7 @@ from sqlglot.dialects.mysql import MySQL
 from sqlglot.generator import Generator
 from sqlglot.helper import csv, list_get
 from sqlglot.parser import Parser
+from sqlglot.tokens import Tokenizer, TokenType
 
 
 def _approx_distinct_sql(self, expression):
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/spark.py
```diff
@@ -63,5 +63,6 @@ class Generator(Hive.Generator):
             exp.Create: _create_sql,
             exp.Map: _map_sql,
             exp.Reduce: rename_func("AGGREGATE"),
+            exp.ArrayFilter: rename_func("FILTER"),
```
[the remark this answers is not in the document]
tobymao: meaning spark read filter output array filter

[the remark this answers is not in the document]
eakmanrq: Yeah good point about covering the opposite and I have added support for that.

I did some research and it looks like only Spark and Presto support array filtering. My latest change adds converting to and from those dialects.

sqlglot/dialects/presto.py
```diff
@@ -118,6 +119,11 @@ class Presto(Dialect):
     time_format = "'%Y-%m-%d %H:%i:%S'"
     time_mapping = MySQL.time_mapping
 
+    class Tokenizer(Tokenizer):
+        KEYWORDS = {
+            k: v for k, v in Tokenizer.KEYWORDS.items() if v != TokenType.FILTER
```
tobymao, line 124: i don’t think this is right, you just need to add filter to the list of functions in parser

tobymao: filter is used i agg expressions like sum(x) filter…

eakmanrq: I didn't think that was supported by Spark but I just confirmed it does actually work. TIL. Well the bad news is that it looks like sqlglot evaluates the keywords before the functions so when I don't have that included it tries to parse FILTER within the context of an aggregate instead of a function.

eakmanrq: I can't get this syntax to work on our Spark release:
```
SELECT
  SUM(play_secs) FILTER (where account_id > 1)
FROM
  VALUES(1, 2) tab(account_id, play_secs)
```
mismatched input '(' expecting <EOF>(line 2, pos 24)

== SQL ==
SELECT
  SUM(play_secs) FILTER (where account_id > 1)
------------------------^^^
FROM
  VALUES(1, 2) tab(account_id, play_secs)
```

eakmanrq: If I remove the that keyword override then it doesn't work since it looks like sqlglot treats it as a filter like in the context above instead of the custom function I am defining.

eakmanrq: I guess regardless if it works for our spark it does seem to be supported: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-groupby.html#description

So we need to figure this out regardless.

tobymao: i’ll figure it out tonight

tobymao:
closed in favor of #266

PR #266, opened 2022-07-24
Test

tobymao:


tobymao:
@eakmanrq you just needed to add FILTER as a FUNCTION

eakmanrq:
Ah I see now. The `_sql_names` feature really helps with situations like this.

PR #280, opened 2022-07-30
correlated unnest fixes #269

### LINKED ISSUE #269: Error with GENERATE_DATE_ARRAY + UNNEST
I get the error "Unknown colum" when I use this query with the optimizer.
```
SELECT
  *
FROM
  table_1
CROSS JOIN
UNNEST(GENERATE_DATE_ARRAY(start_time, date_snapshot, INTERVAL 1 MONTH)) AS time_value

```

tobymao:


sqlglot/optimizer/qualify_columns.py
```diff
@@ -154,14 +154,16 @@ def _qualify_columns(scope, schema):
 
             column_table = unambiguous_columns.get(column_name)
 
-            if not scope.is_subquery:
+            if not scope.is_subquery and scope.scope_type != ScopeType.UNNEST:
```
barakalon, line 157: `is_subquery` just checks the scope type. For consistency, maybe we should have a `is_unnest` property?

barakalon (approved):
Wow, this was a lot slicker than I was expecting. Nice

sqlglot/optimizer/qualify_columns.py
tobymao: done

PR #282, opened 2022-08-02
[WIP] add sqlite keywords

giancarlopro:
hi, i am working with the sql below and needed to parse `AUTOINCREMENT` and `UNIQUE` so i added it direct to code.

```sql
CREATE TABLE props (_id INTEGER PRIMARY KEY AUTOINCREMENT, key TEXT UNIQUE, value TEXT);
```

i'm not sure where to put tests for this feature or whether i should've open an issue before, but i'm willing to, just tell me what to do.

also i need to parse trigger and index creation, should i implement it in this same pr?

```sql
CREATE INDEX message_type_index ON message(message_type);
CREATE TRIGGER quick_replies_bd_for_quick_reply_usage_trigger BEFORE DELETE ON quick_replies BEGIN DELETE FROM quick_reply_usage WHERE quick_reply_id=old._id; END;
```


tobymao:
let's do the index and trigger in another pr.

you should add a test for this in identity.sql and dialects.py#sqlite

your code is missing the generator for unique and autoincrement for sqlite

tobymao:
you need to add TokenType.UNIQUE to ID_VAR_TOKENS since you can probably do like

select x unique as an alias

giancarlopro:
- [x] add generators for unique and autoincrement
- [x] add tests on dialects.py#sqlite
- [x] add TokenType.UNIQUE to ID_VAR_TOKENS
- [x] add tests in identity.sql

i've changed `AUTOINCREMENT` to come after `PRIMARY KEY` on generator, sqlite seem to only work in this order, on mysql both work. And added a `TOKEN_MAPPING` to use `AUTOINCREMENT` when sqlite and `AUTO_INCREMENT` otherwise.

sqlglot/generator.py
```diff
@@ -242,14 +244,17 @@ def columndef_sql(self, expression):
         default = self.sql(expression, "default")
         default = f" DEFAULT {default}" if default else ""
         auto_increment = (
-            " AUTO_INCREMENT" if expression.args.get("auto_increment") else ""
+            " " + self.TOKEN_MAPPING.get(TokenType.AUTO_INCREMENT, "AUTO_INCREMENT")
```
tobymao, line 247: i'm going to refactor this slightly by abstracting it away in sql

tobymao (approved):
fantastic pr. thanks for adding this

giancarlopro:
thank you for your time maintining this, really useful

Issue #304, opened 2022-08-08
Doesn't Differentiate Using Aliases in Group By 


vldt-aem:
The test that I used was:

```python
sql = """SELECT
    table_schema
    , CASE WHEN table_schema = 'my_table' THEN 1 ELSE 0 END as test
    , COUNT(*)

FROM information_schema.columns
GROUP BY
    table_schema
    , test
LIMIT 10 """
print(sqlglot.transpile(sql, read='postgres', write='oracle', identify=True, pretty=True)[0])
```

And what it returned was: 
```
SELECT
  "table_schema",
  CASE
    WHEN "table_schema" = 'my_table'
    THEN 1
    ELSE 0
  END AS "test",
  COUNT(*)
FROM "information_schema"."columns"
GROUP BY
  "table_schema",
  "test"
LIMIT 10
```

Referencing aliases in the group by is not supported in Oracle SQL, so this returns an error

tobymao:
what is the correct behavior?

tobymao:
i guess you're asking to rewrite the case statement in the group by? i think for now it's best to do this yourself using the transform api. 

you can walk the expression tree, find expressions within group by that point to aliases and rewrite them.

barakalon:
What if we add the transformation logic to the generator of Oracle's dialect?

tobymao:
@barakalon sounds good to me, should i reopen this issue?

vldt-aem:
@tobymao @barakalon Awesome, thank you!

PR #305, opened 2022-08-08
unalias group by references for Oracle

barakalon:
fixes https://github.com/tobymao/sqlglot/issues/304


tests/test_dialects.py
```diff
@@ -1749,6 +1749,11 @@ def test_oracle(self):
             "SELECT x FROM y LIMIT 10",
             read="oracle",
         )
+        self.validate(
```
barakalon, line 1752: There's probably a more generic way to test this, since its not necessarily specific to Oracle

tobymao:
@barakalon i think actually this is not oracle specific, i think most engines can't use aliases in the group by

tobymao (approved):

PR #306, opened 2022-08-08
[WIP] Spark only struct support

eakmanrq:
Change to support struct data types for Spark only. BigQuery uses a similar syntax to spark but with a space instead of colon to separate the name and type. Presto is similar to BigQuery syntax but uses the term `ROW` instead of struct. Support for those dialects still needs to be implemented. I'm open to adding it but I need to at least stop here to unblock some work I am currently doing.

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -2,6 +2,7 @@ from sqlglot import exp
 from sqlglot.dialects.dialect import rename_func
 from sqlglot.dialects.hive import Hive, HiveMap
 from sqlglot.helper import list_get
+from sqlglot.tokens import TokenType
 
 
 def _create_sql(self, e):
@@ -20,6 +21,14 @@ def _map_sql(self, expression):
 
 
 class Spark(Hive):
+    class Tokenizer(Hive.Tokenizer):
+        KEYWORDS = {
+            k: v
+            for k, v in
+            Hive.Tokenizer.KEYWORDS.items()
+            if v not in (TokenType.RSHIFT, TokenType.LSHIFT)
+        }
+
     class Parser(Hive.Parser):
         FUNCTIONS = {
             **Hive.Parser.FUNCTIONS,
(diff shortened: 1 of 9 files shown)


sqlglot/dialects/spark.py
```diff
@@ -20,6 +21,14 @@ def _map_sql(self, expression):
 
 
 class Spark(Hive):
+    class Tokenizer(Hive.Tokenizer):
+        KEYWORDS = {
+            k: v
+            for k, v in
+            Hive.Tokenizer.KEYWORDS.items()
+            if v not in (TokenType.RSHIFT, TokenType.LSHIFT)
```
tobymao, line 29: is this necessary?

sqlglot/dialects/spark.py
```diff
@@ -40,6 +49,14 @@ class Parser(Hive.Parser):
                 ),
                 length=list_get(args, 1),
             ),
+            "SHIFTRIGHT": lambda args: exp.BitwiseRightShift(
```
tobymao, line 52: i think you can just do rename_func?

tobymao: or from_arg_list?

sqlglot/parser.py
```diff
@@ -1300,9 +1302,14 @@ def _parse_type(self):
 
     def _parse_types(self):
         index = self._index
+        name = None
 
         if not self._match_set(self.TYPE_TOKENS):
-            return None
+            if self._prev is not None and self._next is not None and self._next.token_type == TokenType.COLON:
```
tobymao, line 1308: this doesn't look quite right, what are you trying to do?

sqlglot/tokens.py
```diff
@@ -188,6 +188,7 @@ class TokenType(AutoName):
     SHOW = auto()
     SOME = auto()
     STORED = auto()
+    STRUCT = auto()
```
tobymao, line 191: if you add this new token type you need to add it to id_var_tokens so that people can use struct as an alias name

select 1 struct

eakmanrq:
Closed in favor of https://github.com/tobymao/sqlglot/pull/310

PR #310, opened 2022-08-09
add more robust struct support

tobymao:


sqlglot/dialects/presto.py
```diff
@@ -118,6 +119,12 @@ class Presto(Dialect):
     time_format = "'%Y-%m-%d %H:%i:%S'"
     time_mapping = MySQL.time_mapping
 
+    class Tokenizer(Tokenizer):
+        KEYWORDS = {
+            **Tokenizer.KEYWORDS,
+            "ROW": TokenType.STRUCT,
```
eakmanrq, line 125: We had talked about how this would probably not be a good idea. It seems like it works out in practice since the only time I ever see row is with `CURRENT ROW` and `CURRENT ROW` get parsed into the `CURRENT_ROW` token. So I think this should end up working out.

sqlglot/parser.py
```diff
@@ -1255,7 +1248,27 @@ def _parse_escape(self, this):
...
+        this = self._parse_term()
+
+        while True:
+            if self._match_set(self.BITWISE):
+                this = self.expression(
+                    self.BITWISE[self._prev.token_type],
+                    this=this,
+                    expression=self._parse_term(),
+                )
+            elif self._match_pair(TokenType.LT, TokenType.LT):
```
eakmanrq, line 1260: This change means that I can have sqlglot parse `SELECT 4 << 1` as Spark dialect and if I write it back out as Spark it will convert to `SELECT SHIFTLEFT(4, 1)`. That output is right but that input is invalid Spark. Before I was seeing that as a problem but it seems like sqlglot as a library is ok with this type of invalid input as long as the output is right?

sqlglot/parser.py
```diff
@@ -1352,6 +1372,14 @@ def _parse_types(self):
             nested=nested,
         )
 
+    def _parse_struct_kwargs(self):
+        this = self._parse_id_var()
+        self._match(TokenType.COLON)
+        data_type = self._parse_types()
+        if not data_type:
+            return None
+        return self.expression(exp.StructKwarg, this=this, expression=data_type)
```
eakmanrq, line 1381: Yeah this is much cleaner. I hadn't thought of the option of taking the Struct type and making it into an expression.

eakmanrq (approved):

sqlglot/parser.py
tobymao: correct, i parse a superset of sql

sqlglot/dialects/presto.py
tobymao: right

PR #317, opened 2022-08-10
Allow parsing properties in any order

eakmanrq:
Prior to this change there was an assumed order which properties on a table could be define. Now they can be in any order. In addition there are now flags that tell the generator if table options are supported and if "COMMENT" is considered an option. This allows for example to read in a comment as option from MySQL but then write it as property for Spark. 

I initially started this work with combining options and properties. As I worked through that I began to realize that BigQuery and Snowflake DDL support is fairly basic right now. I think it would be good to take some time to think through how to update the parser/generator to work with all supported dialects since it seems like there are a lot of inconsistencies across dialects for DDL formats. My bet is that creating a fairly basic base parser/generator to start but likely most dialects will need to implement their own parser/generator. Once that is done then these new generator flags can be removed since each dialect would have their own DDL generation logic.


sqlglot/dialects/hive.py
```diff
@@ -55,38 +55,30 @@ def _properties_sql(self, expression):
...
-    using = ""
-    stored_as = ""
-    partitioned_by = ""
-    location = ""
-
-    for p in properties:
+    values = []
+    for p in properties[:]:
         name = p.name.upper()
+        value = None
```
tobymao, line 61: sorry nit, but i prefer doing else: value = None instead of predefining it

sqlglot/generator.py
```diff
@@ -259,6 +263,23 @@ def columndef_sql(self, expression):
         return f"{column} {kind}{not_null}{default}{collate}{comment}{unique}{primary}{auto_increment}"
 
     def create_sql(self, expression):
+        comment = self.sql(expression, "comment")
+        if comment:
+            if self.TABLE_COMMENT_AS_OPTION:
+                comment = f"{c.COMMENT}={comment}"
+            else:
+                properties = expression.args["properties"] or exp.Properties(
```
tobymao, line 271: get("properties")  since it may not exist?

sqlglot/generator.py
```diff
@@ -259,6 +263,23 @@ def columndef_sql(self, expression):
...
+                    expressions=[]
+                )
+                properties.args["expressions"].append(
```
tobymao, line 274: this is dangerous because you are mutating properties expressions, the generator shouldn't mutate the parent tree

sqlglot/parser.py
```diff
@@ -606,21 +606,20 @@ def _parse_property(self, schema):
...
         properties = []
 
-        if self._match(TokenType.WITH):
-            self._match_l_paren()
-            properties.extend(self._parse_csv(lambda: self._parse_property(schema)))
-            self._match_r_paren()
-        else:
-            if self._match(TokenType.USING):
+        while True:
+            if self._match(TokenType.WITH):
```
tobymao, line 610: should this block be outside of the while True block ?

sqlglot/dialects/mysql.py
```diff
@@ -24,6 +25,9 @@ class MySQL(Dialect):
         "%S": "%S",
     }
 
+    class Tokenizer(Tokenizer):
+        QUOTES = ["'", '"']
```
tobymao, line 29: why did you add this? isn't " an identifier?

eakmanrq: It looks like for strings both are supported but for date and datetime literals you need single quotes:
```
Single quotes should be used for string values like in the VALUES() list. Double quotes are supported by MySQL for string values as well, but single quotes are more widely accepted by other RDBMS, so it is a good habit to use single quotes instead of double.

MySQL also expects DATE and DATETIME literal values to be single-quoted as strings like '2001-01-01 00:00:00'. Consult [the Date and Time Literals](https://dev.mysql.com/doc/refman/5.7/en/date-and-time-literals.html) documentation for more details, in particular alternatives to using the hyphen - as a segment delimiter in date strings.
```
https://stackoverflow.com/questions/11321491/when-to-use-single-quotes-double-quotes-and-backticks-in-mysql

PR #318, opened 2022-08-10
change references generator

giancarlopro:
i've changed references generator because it was raising syntax error both on postgres and sqlite

```
sqlite> create table a (id INTEGER PRIMARY KEY);
sqlite> create table b (id integer primary key, a_id integer, foreign key (a_id) references a(id));
sqlite> create table c (id integer primary key, a_id integer, foreign key (a_id) references (a) (id));
Parse error: near "(": syntax error
   key, a_id integer, foreign key (a_id) references (a) (id));
                                      error here ---^
```

If theres a dialect which uses `REFERENCES (table) (id)`, how can i add that?

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -749,7 +749,7 @@ class Generator:
     def reference_sql(self, expression):
         this = self.sql(expression, "this")
         expressions = self.expressions(expression, flat=True)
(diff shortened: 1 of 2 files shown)


tobymao (approved):
was probably a bug

PR #322, opened 2022-08-11
Implement `TO_CHAR` format string translation

cpcloud:
This PR implements the necessary bits to translate [`TO_CHAR` format strings](https://www.postgresql.org/docs/current/functions-formatting.html)
into Python strftime format strings, enabling transpilation across the various
sqlglot dialects.


tobymao:
thanks!

PR #329, opened 2022-08-13
Flexible DDL Expression Support

eakmanrq:
Prior to this change the DDL for tables were expected to follow a specific order in order to get transpiled correctly. In addition there was a distinction between "properties" and "options" that seemed to apply to MySQL only. 

This change updates the parser to treat all properties the same (no longer finds properties and then options). In addition it makes property specific expressions instead of a generic Property expression. It also preserved the order in which the properties were defined. Dialects can decide which properties should be at the root level, under the `WITH` clause, or under a `TBLPROPERTIES` clause. They can then override the sql generation behavior for each property if needed. The base implementation is configured to generate MySQL correctly since that is how the tests are defined. 

Better DDL support is still needed for Snowflake and BigQuery and I've decided to make that as out of scope for this PR.


sqlglot/expressions.py
```diff
@@ -1358,19 +1393,26 @@ def ctas(self, table, properties=None, dialect=None, parser_opts=None, copy=True
             dialect=dialect,
             parser_opts=parser_opts,
         )
+        properties_expression = None
+        if properties:
+            properties_str = " ".join(
+                [
+                    f"{k} = {v}" if not isinstance(v, str) else f"{k} = '{v}'"
```
tobymao, line 1400: nit: reverse this so that the if statement is simpler if isinstance(str)...

sqlglot/generator.py
```diff
@@ -38,8 +38,18 @@ class Generator:
     """
 
     TRANSFORMS = {
+        exp.AnonymousProperty: lambda self, e: self.property_sql(e),
+        exp.AutoIncrementProperty: lambda self, e: f"AUTO_INCREMENT={e.text('value')}",
```
tobymao, line 42: should we do self.sql(e, 'value') instead of all the text references, and then the quotes vs no quote is automatic

sqlglot/generator.py
```diff
@@ -38,8 +38,18 @@ class Generator:
...
+        exp.CharacterSetProperty: lambda self, e: f"{'DEFAULT ' if e.args['default'] else ''}CHARACTER SET={e.text('value')}",
+        exp.CollateProperty: lambda self, e: f"COLLATE={e.text('value')}",
         exp.DateAdd: lambda self, e: f"DATE_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')}, {self.sql(e, 'unit')})",
         exp.DateDiff: lambda self, e: f"DATE_DIFF({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
+        exp.EngineProperty: lambda self, e: f"ENGINE={e.text('value')}",
+        exp.FileFormatProperty: lambda self, e: f"format = '{e.text('value')}'",
```
tobymao, line 48: looks like your spacing is not consistent? also should it be caps FORMAT

sqlglot/generator.py
```diff
@@ -394,7 +397,31 @@ def partition_sql(self, expression):
         return f"PARTITION({keys})"
 
     def properties_sql(self, expression):
-        return self.properties("WITH", expression)
+        root_expression = expression.copy()
+        with_expression = expression.copy()
+        table_properties_expression = expression.copy()
```
tobymao, line 402: i'm not a fan of how this was done.

we should only have root properties and with properties, tblproperties should only be overridden in hive, hive should overwrite with properties or something to be tblproperties

tobymao: also, let's not copy expression n times and remove, can we just create a new root with the appropriate children?

sqlglot/generator.py
```diff
@@ -947,17 +974,17 @@ def format_time(self, expression):
             self.sql(expression, "format"), self.time_mapping, self.time_trie
         )
 
-    def expressions(self, expression, key=None, flat=False, indent=True):
+    def expressions(self, expression, key=None, flat=False, indent=True, item_sep=", "):
```
tobymao, line 977: maybe just call it sep=", "

sqlglot/parser.py
```diff
@@ -572,93 +539,145 @@ def parse_option(option, token, option_lambda):
...
-
-        return self.expression(
-            exp.Property,
-            this=exp.Literal.string(key),
-            value=value,
-        )
-
-    def _parse_properties(self, schema):
-        properties = []
+        if self._match(TokenType.USING):
```
tobymao, line 545: can we move this all to a dictionary with a lambda

TokenType.USING: lambda ... 

that way every dialect can override how it reads property, also then it will be constant time to match

sqlglot/parser.py
```diff
@@ -572,93 +539,145 @@ def parse_option(option, token, option_lambda):
...
+            return self.expression(
+                exp.TableFormatProperty,
+                this=exp.Literal.string(c.TABLE_FORMAT),
```
tobymao, line 548: these constants were useful because we were using them on read too, but now that we have first class expressions, they are unnecessary, can you delete all the constants and just hardcode the default string here?

sqlglot/parser.py
```diff
@@ -572,93 +539,145 @@ def parse_option(option, token, option_lambda):
...
+                this=exp.Literal.string(c.PARTITIONED_BY),
+                value=self._parse_schema(),
+            )
+        if self._match(TokenType.STORED):
+            self._match(TokenType.ALIAS)
+            self._match(TokenType.EQ)
+            return self.expression(
+                exp.FileFormatProperty,
+                this=exp.Literal.string(c.FILE_FORMAT),
+                value=exp.Literal.string(self._parse_var().text("this")),
```
tobymao, line 563: instead of text("this"), you can just call .name

sqlglot/parser.py
```diff
@@ -572,93 +539,145 @@ def parse_option(option, token, option_lambda):
...
+                value=self._parse_var() or self._parse_string(),
+            )
+        if self._match(TokenType.AUTO_INCREMENT):
+            self._match(TokenType.EQ)
+            return self.expression(
+                exp.AutoIncrementProperty,
+                this=exp.Literal.string(c.AUTO_INCREMENT),
+                value=self._parse_var() or self._parse_number(),
+            )
+        if self._match(TokenType.CHARACTER_SET) or self._match_sequence(
```
tobymao, line 592: this is the only one that won't map in the dictionary, but that's ok, just have that one after the main dictionary matcher

sqlglot/parser.py
```diff
@@ -572,93 +539,145 @@ def parse_option(option, token, option_lambda):
...
+                and potential_default_token.token_type == TokenType.DEFAULT,
+            )
+        if self._match(TokenType.COLLATE):
+            self._match(TokenType.EQ)
+            return self.expression(
+                exp.CollateProperty,
+                this=exp.Literal.string(c.COLLATE),
+                value=self._parse_var() or self._parse_string(),
+            )
+        if self._match(TokenType.SCHEMA_COMMENT):
```
tobymao, line 611: should all of these be match_sequence too? or are there dialects without mandatory eq?

sqlglot/parser.py
```diff
@@ -572,93 +539,145 @@ def parse_option(option, token, option_lambda):
...
+            self._match(TokenType.EQ)
+            return self.expression(
+                exp.SchemaCommentProperty,
+                this=exp.Literal.string(c.COMMENT),
+                value=self._parse_string(),
+            )
+        if self._match_sequence(
+            TokenType.VAR, TokenType.EQ, advance=False
+        ) or self._match_sequence(TokenType.VAR, TokenType.EQ, advance=False):
```
tobymao, line 620: what's this or statement both var eq? typo?

tobymao: how come you're not advancing it? so that you can absorb the key? maybe match_sequence can return a tuple?

sqlglot/parser.py
```diff
@@ -1962,18 +1981,17 @@ def _match_set(self, types):
...
-            return True
-
-        return None
+    def _match_sequence(self, *sequence, advance=True):
+        for i, sequence_token_type in enumerate(sequence):
+            actual_token = self._get_token_at_index(self._index + i)
+            if not actual_token or sequence_token_type != actual_token.token_type:
+                return False
+        if advance:
+            self._advance(len(sequence))
```
tobymao, line 1990: you may actually be able to just do self._advance(i)

sqlglot/parser.py
```diff
@@ -443,12 +444,15 @@ def _find_token(self, token, sql):
 
         return index
 
+    def _get_token_at_index(self, index):
```
tobymao, line 447: nit can you call this def _get_token(self, index)

sqlglot/tokens.py
```diff
@@ -383,6 +384,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "FUNCTION": TokenType.FUNCTION,
         "FOLLOWING": TokenType.FOLLOWING,
         "FOREIGN KEY": TokenType.FOREIGN_KEY,
+        "FORMAT": TokenType.FORMAT,
```
tobymao, line 387: whenever you add a token you need to add it to the ID_VAR_CONSTANTS or else you won't be able to use it as an alias 

select x format

sqlglot/expressions.py
```diff
@@ -759,6 +754,46 @@ class Property(Expression):
     arg_types = {"this": True, "value": True}
 
 
+class TableFormatProperty(Property):
```
tobymao, line 757: do you like having all these with the property name on it? seems a bit unnecessary since the subclass is already property

so just TableFormat and PartitionedBy etc

sqlglot/generator.py
eakmanrq: Good call that helps a lot

sqlglot/generator.py
eakmanrq: This has to do with the identity tests. They expect root expressions to have the space and lowercase while WITH expressions to have no space and uppercase. Do we want to change the identity tests to reflect a more consistent approach?

tobymao: you can fix the tests yes

sqlglot/parser.py
eakmanrq: Yeah typo thanks for catching.

Yeah I want to absorb the key. Personally this pattern seems fine... with later Python releases we get the walrus operator which then combined with returning the tuple could be a nice solution.

sqlglot/parser.py
eakmanrq: `self._advance(i+1)` technically works but IDE isn't happy with it. I think the problem is that in the case where sequence is empty then the i never gets initialized. That really shouldn't happen in practice but going to keep the current way since it seems comparable and removes a warning.

sqlglot/parser.py
eakmanrq: I purposely did it this way to make the parsing more flexible. For example MySQL has `COMMENT=<comment>` while Spark has `COMMENT <comment>`. So I think this approach is more robust since we don't really care from a parsing perspective if the equal is there or not.

sqlglot/parser.py
eakmanrq: Yeah nice improvement.

sqlglot/parser.py
tobymao: what if you set i = 0?

sqlglot/parser.py
tobymao: ok

sqlglot/parser.py
tobymao: sounds good

sqlglot/tokens.py
eakmanrq: Thanks!

sqlglot/expressions.py
eakmanrq: Yeah I actually do like it. When someone is seeing these expressions referenced in code they don't see the class they inherited from. So it is nice making it explicit IMO. 

Also there is `AnonymousProperty` which really needs the property suffix to get it's purposes IMO. That also helps reinforce the point above.

sqlglot/parser.py
eakmanrq: Yeah but now we are adding another line of code... seems all messier then this approach.

sqlglot/expressions.py
tobymao: sounsd good

sqlglot/parser.py
tobymao: as long as there's no performance implication

tobymao: i'm actually a bit afraid that _match_sequence is gonna significantly slow down the parser. you're now dynamically getting grabbing tokens instead of using the cache that was previously used 

self._curr and self._next

was this change necessary? do you even match more than just the pair?

eakmanrq: This goes back to the typo that you mentioned. Originally I was matching on 3 things but then realized later I only needed 2. So yeah I can remove this now since it is not currently needed. Once someone wants to match on >2 items then we will need it.

tobymao: cool yea let’s revert it for now. in general if you need to match things like that you can also use the tokenizer like with time zone

sqlglot/generator.py
eakmanrq: So I do something like this now:
```
root_expression = exp.Properties(expressions=root_properties)
with_expression = exp.Properties(expressions=with_properties)
```

The risk I see though is that since we aren't cloning here we are losing any addition information that could be attached to properties that somewhere in later calls would assume exists. There is not an example of this today but wanted to point out that this could create some non-intuitive behavior. Another way to do this is to clone and then replace `expressions` with a list of properties. From your comment I think what I did is what you are wanting but I want to point out that other options too.

sqlglot/generator.py
```diff
@@ -394,7 +402,27 @@ def partition_sql(self, expression):
...
     def properties_sql(self, expression):
-        return self.properties("WITH", expression)
+        root_properties = []
+        with_properties = []
+
+        for p in expression.expressions:
+            p_class = p.__class__
+            if p_class in self.ROOT_PROPERTIES:
+                root_properties.append(p)
+            if p_class in self.WITH_PROPERTIES:
```
tobymao, line 412: elif?

sqlglot/generator.py
tobymao: yea this is what i was looking for, not exactly sure in understanding the case you’re talking about though

tobymao (approved):

sqlglot/dialects/presto.py
```diff
@@ -151,6 +151,15 @@ class Parser(Parser):
     class Generator(Generator):
         STRUCT_DELIMITER = ("(", ")")
 
+        NATIVE_PROPERTIES = []
```
tobymao, line 154: delete this?

tobymao:
can you delete constants.py?

sqlglot/dialects/presto.py
eakmanrq: Good catch thanks!

eakmanrq:
The file existed before this PR so I just removed the things that I added to it. It should be a small follow up PR if you want it completely removed but I see it as out of scope for this PR.

PR #335, opened 2022-08-16
Fix spark binary type

looi:
ARRAY[BYTE] isn't even valid spark sql (it should be ARRAY<BYTE>),
and it's not the same type as BINARY.

https://spark.apache.org/docs/latest/sql-ref-datatypes.html


tobymao (approved):

tobymao:
thanks

PR #336, opened 2022-08-16
Always return subquery

barakalon:
Always return subqueries, never paranthesized subqueryables.


sqlglot/optimizer/scope.py
```diff
@@ -86,19 +86,32 @@ def _collect(self):
         self._derived_tables = []
         self._raw_columns = []
 
-        for node, *_ in _walk_next_scope(self.expression):
+        # We'll use this variable to pass state into the dfs generator.
+        # Whenever we set it to True, we exclude a subtree from traversal.
+        prune = False
```
barakalon, line 91: This is a wacky way to control the DFS traversal... but I kinda dig it.

tobymao: does this reval the lambda????

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -39,13 +39,13 @@ WITH "cte" AS (
     SELECT
       "x"."a" AS "a"
     FROM "x" AS "x"
-  ) AS "_q_0"
```
tobymao, line 42: why did this get changed? is this correct?

tobymao (approved):

tests/fixtures/optimizer/optimizer.sql
barakalon: These statements were getting treated as derived tables, so they get qualified.

But I think they're actually subqueries, so I think this is correct.

sqlglot/optimizer/scope.py
barakalon: Lambdas can reference variables in the scope they're defined. The values don't get copied. 💥 

So every time the lambda function gets called, it returns the current value of `prune`. And the lambda gets called on every node.

PR #337, opened 2022-08-17
ErrorLevel.IMMEDIATE

barakalon:
This is a big change, so let's think about it.

This raises a "super error" if error level is RAISE. That's still the default, so users will start getting larger error messages.

Other options:
1. Change RAISE to only raise the first error and log the rest
    - I'm not sure why a user would want this behavior
2. Make IMMEDIATE the default
3. Remove IMMEDIATE and make RAISE raise immediately

### DIFF
--- a/sqlglot/errors.py
+++ b/sqlglot/errors.py
@@ -4,9 +4,10 @@ from sqlglot.helper import AutoName
 
 
 class ErrorLevel(AutoName):
-    IGNORE = auto()
-    WARN = auto()
-    RAISE = auto()
+    IGNORE = auto()  # Ignore any parser errors
+    WARN = auto()  # Log any parser errors with ERROR level
+    RAISE = auto()  # Collect all parser errors and raise a single exception
+    IMMEDIATE = auto()  # Immediately raise an exception on the first parser error
 
 
 class SqlglotError(Exception):
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -476,11 +476,11 @@ class Parser:
         return expressions
 
     def check_errors(self):
-        for error in self.errors:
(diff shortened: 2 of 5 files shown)


tobymao:
What do you mean by this is a big change? As it is currently, it looks like it doesn't break anything. Are you suggesting that we should consider doing the bigger change which is #2 or #3?

barakalon:
Not a breaking change, just a big change, in the sense that error messaging is a big part of usability.

One potential issue here: error messages are now unbounded:

```
>>> sqlglot.parse_one("((((((")
...
ParseError: Expecting ). Line 1, Col: 6.
  ((((((

Required keyword: 'this' missing for <class 'sqlglot.expressions.Paren'>. Line 1, Col: 6.
  ((((((

Expecting ). Line 1, Col: 6.
  ((((((

Expecting ). Line 1, Col: 6.
  ((((((

Expecting ). Line 1, Col: 6.
  ((((((

Expecting ). Line 1, Col: 6.
  ((((((

Expecting ). Line 1, Col: 6.
  ((((((

```

So maybe, if we go this direction, we cap the number of errors:
```
ParseError: Expecting ). Line 1, Col: 6.
  ((((((

Required keyword: 'this' missing for <class 'sqlglot.expressions.Paren'>. Line 1, Col: 6.
  ((((((

Expecting ). Line 1, Col: 6.
  ((((((

... and 7 more
```

tobymao:
yea the capped errors seems reasonable, what do you think?

barakalon:
Sure.

And I guess we'll start getting issue reports with longer error messages, so we'll get a feel for how useful this is.

tobymao (approved):

PR #339, opened 2022-08-17
attempt to cleanup greedy subquery parser

tobymao:
almost got it, just failing one test


sqlglot/parser.py
```diff
@@ -576,12 +576,12 @@ def _parse_statement(self):
...
-        # There has to be a more elegant way to parse this, but as a (hopefully) temporary
-        # hack, we're setting this instance attribute to tell the subquery parser to
-        # greedily consume these clauses.
-        self._greedy_subqueries = True
         expression = self._parse_expression()
-        self._greedy_subqueries = False
+
+        if isinstance(expression, (exp.Subquery, exp.Union)):
+            for k, v in self._parse_orders_and_limits().items():
+                expression.set(k, v)
```
barakalon, line 583: Union doesn't have these args

tobymao: now it does

barakalon (approved):

PR #346, opened 2022-08-19
Generate UNION DISTINCT correctly for BigQuery syntax

sid-kap:
According to https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#set_operators, for the UNION command in BigQuery, "You must specify ALL or DISTINCT". 

This updates the BigQuery dialect generator to output `UNION DISTINCT` rather than `UNION` when distinct is True.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -92,11 +92,17 @@ class BigQuery(Dialect):
             exp.DataType.Type.TEXT: "STRING",
         }
 
-        def union_sql(self, expression):
-            return self.prepend_ctes(
-                expression,
-                self.set_operation(
-                    expression,
-                    f"UNION{' DISTINCT' if expression.args.get('distinct') else ' ALL'}",
-                ),
-            )
+        def union_op(self, expression):
+            return f"UNION{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
+
+        def except_op(self, expression):
+            if not expression.args.get("distinct", False):
+                self.unsupported("EXCEPT without DISTINCT is not supported in BigQuery")
+            return super().except_op(expression)
+
+        def intersect_op(self, expression):
+            if not expression.args.get("distinct", False):
+                self.unsupported(
+                    "INTERSECT without DISTINCT is not supported in BigQuery"
+                )
+            return super().intersect_op(expression)
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -91,3 +91,11 @@ class BigQuery(Dialect):
             exp.DataType.Type.BOOLEAN: "BOOL",
             exp.DataType.Type.TEXT: "STRING",
         }
+
+        def union_sql(self, expression):
+            return self.prepend_ctes(
+                expression,
+                self.set_operation(
+                    expression, f"UNION{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
+                ),
+            )
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -96,6 +96,7 @@ class BigQuery(Dialect):
             return self.prepend_ctes(
                 expression,
                 self.set_operation(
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -91,3 +91,12 @@ class Generator(Generator):
             exp.DataType.Type.BOOLEAN: "BOOL",
             exp.DataType.Type.TEXT: "STRING",
         }
+
+        def union_sql(self, expression):
```
barakalon, line 95: I wonder if we should be overriding `set_operation` instead of `union_sql`. I don't know bigquery that well, but from the link you shared, looks like we might want the same for EXCEPT and INTERSECT.

Also, if we override `set_operation` there will be less duplication of Generator code.

barakalon (commented):
Can you add test(s) here: https://github.com/tobymao/sqlglot/blob/main/tests/test_dialects.py#L345

sqlglot/dialects/bigquery.py
barakalon: Ah, EXCEPT and INTERSECT already add the DISTINCT token. Hm.

I'll let @tobymao comment on this one. I wonder if we want to add some other method/attribute on the base Generator so we don't have to override the entire union_sql method.

sid-kap: Thanks for taking a look so quickly!

It looks like BigQuery doesn't support EXCEPT or INTERSECT without DISTINCT, so maybe we should throw an error if you try to render one that doesn't have `args['distinct'] == True`?

barakalon: How about three new methods, `intersect_op`, `except_op`, `union_op`.

The defaults would look like:

```python
def union_op(self, expression):
    return f"UNION{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
```

Just a smaller surface area to override.

> so maybe we should throw an error 

Check out `Generator.unsupported`

sqlglot/generator.py
```diff
@@ -677,11 +677,12 @@ def qualify_sql(self, expression):
...
         return self.prepend_ctes(
             expression,
-            self.set_operation(
-                expression, f"UNION{'' if expression.args.get('distinct') else ' ALL'}"
-            ),
+            self.set_operation(expression, self.union_op(expression)),
         )
 
+    def union_op(self, expression):
+        return f"UNION{'' if expression.args.get('distinct') else ' ALL'}"
```
tobymao, line 684: should we just explicitly make union distinct? then we don't need the bigquery edge case?

sqlglot/dialects/bigquery.py
```diff
@@ -91,3 +91,18 @@ class Generator(Generator):
             exp.DataType.Type.BOOLEAN: "BOOL",
             exp.DataType.Type.TEXT: "STRING",
         }
+
+        def union_op(self, expression):
+            return f"UNION{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
+
+        def except_op(self, expression):
```
tobymao, line 98: what is the behavior in other engines? are all engines except/intersect distinct implicit?

sid-kap: Looks like [Snowflake](https://docs.snowflake.com/en/sql-reference/operators-query.html) does duplicate elimination for INTERSECT and EXCEPT

sqlglot/generator.py
sid-kap: I don't think this is safe in general, e.g. [Snowflake](https://docs.snowflake.com/en/sql-reference/operators-query.html) doesn't accept the DISTINCT syntax. This is the grammar for Snowflake:
```
{ INTERSECT | { MINUS | EXCEPT } | UNION [ ALL ] }
```

sid-kap: [Postgres](https://www.postgresql.org/docs/current/queries-union.html) is the same:
```
query1 UNION [ALL] query2
query1 INTERSECT [ALL] query2
query1 EXCEPT [ALL] query2
```

tobymao: great, thanks for the research

tobymao (approved):

PR #347, opened 2022-08-20
Allow expressions (rather than just identifiers) in SELECT * REPLACE

sid-kap:
The [BigQuery documentation](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_replace) says:
> A `SELECT * REPLACE` statement specifies one or more `expression AS identifier` clauses.

I've confirmed that BigQuery lets me run things like
```sql
SELECT * REPLACE (a + 1 AS a) FROM sid.test
```
and it works as expected.

Currently this code causes a parse error.


tobymao:
great, can you write one test for this in identity?

PR #348, opened 2022-08-20
Add NULLS FIRST/NULLS LAST + Dialect Default Null Handling Support

eakmanrq:
The initial change I made was adding support for parsing `NULLS FIRST` or `NULLS LAST` when defining an order by. As I added this though I realized all these different dialects had different default ways of sorting nulls and when we transpile we would drop these implicit sorting unless we encoded them at parse time. Therefore I looked up the docs on the support dialects and defined their default null sorting functionality. 

I didn't add Starrocks support because it doesn't appear that it even supports `NULLS FIRST` or `NULLS LAST` based on the docs. Also I couldn't find in the docs where they define their default sort order. So I just left the implementation with default settings for now.


sqlglot/enums.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
```
eakmanrq, line 1: @tobymao Where would you like me to put this? I couldn't decide so I figured I would just do this and update with your preference.

sqlglot/dialects/mysql.py
```diff
@@ -27,12 +44,37 @@ class MySQL(Dialect):
     class Parser(Parser):
         STRICT_CAST = False
 
+        def _parse_ordered(self):
+            is_null = self._match(TokenType.ISNULL)
+            if is_null:
+                self._match(TokenType.L_PAREN)
```
tobymao, line 50: do you always expect l_paren ? 

self._match_l_paren()
and self._match_r_paren()

sqlglot/dialects/mysql.py
```diff
@@ -27,12 +44,37 @@ class MySQL(Dialect):
...
+                this = self._parse_conjunction()
```
tobymao, line 51: why do you do this manually?

should you just parse a conjunction , allow ISNULL to be a function and use that?

sqlglot/dialects/mysql.py
```diff
@@ -27,12 +44,37 @@ class MySQL(Dialect):
...
+                is_desc = self._match(TokenType.DESC)
+                self._match(TokenType.ASC)
+            else:
+                this = self._parse_conjunction()
+                is_asc = self._match(TokenType.ASC)
+                is_desc = self._match(TokenType.DESC)
+                nulls_first = False
+                if is_asc or (is_asc is None and is_desc is None):
+                    nulls_first = True
+            desc = is_desc or False
```
tobymao, line 66: this is not necessary

sqlglot/dialects/postgres.py
```diff
@@ -71,6 +72,7 @@ class Tokenizer(Tokenizer):
         }
 
     class Parser(Parser):
+        NULL_ORDERING = NullOrdering.NULLS_ARE_LARGE
```
tobymao, line 75: for parameters that are done both parser and generator, you can make it a lower case class instance,

null_ordering = "large"

sqlglot/parser.py
```diff
@@ -216,6 +219,7 @@ class Parser:
         TokenType.EXTRACT,
         TokenType.FILTER,
         TokenType.FIRST,
+        TokenType.ISNULL,
```
tobymao, line 222: make this an id_var?

tobymao: also a function?

sqlglot/parser.py
```diff
@@ -1219,7 +1225,23 @@ def _parse_sort(self, token_type, exp_class):
     def _parse_ordered(self):
         this = self._parse_conjunction()
         self._match(TokenType.ASC)
-        return self.expression(exp.Ordered, this=this, desc=self._match(TokenType.DESC))
+        is_desc = self._match(TokenType.DESC)
+        is_nulls_first = self._match(TokenType.NULLS_FIRST)
+        is_nulls_last = self._match(TokenType.NULLS_LAST)
+        desc = is_desc or False
```
tobymao, line 1231: let's remove or False's, i'll probably make a PR that removes keys that are None

sqlglot/generator.py
```diff
@@ -53,6 +53,8 @@ class Generator:
         exp.TsOrDsAdd: lambda self, e: f"TS_OR_DS_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')}, {self.sql(e, 'unit')})",
     }
 
+    NULL_ORDERING_SUPPORTED = True
```
tobymao, line 56: update the doc string with the enums in both this and the parser

tobymao: oh this is no longer an enum?

sqlglot/dialects/starrocks.py
```diff
@@ -4,6 +4,8 @@
 
 class StarRocks(MySQL):
     class Generator(MySQL.Generator):
+        NULL_ORDERING_SUPPORTED = False
```
tobymao, line 7: this is not necessary as it's inherited from mysql

sqlglot/generator.py
eakmanrq: So this allows a dialect to set their generator to if they support null ordering or not. This is so MySQL and Starrocks can say they don't have null ordering.

tobymao: nvm, this is something else,

can you add null_ordering to the doc strings

sqlglot/parser.py
```diff
@@ -173,10 +173,13 @@ class Parser:
         TokenType.FORMAT,
         TokenType.FUNCTION,
         TokenType.IF,
+        TokenType.ISNULL,
         TokenType.INTERVAL,
         TokenType.LAZY,
         TokenType.LOCATION,
         TokenType.NEXT,
+        TokenType.NULLS_FIRST,
```
tobymao, line 181: these are not needed in id_var_tokens because they have spaces, only the ones that could become aliases need to go in here

sqlglot/parser.py
tobymao: we don't need to make desc required right?

eakmanrq: I see it as an improvement if it is. If there is no benefit from having 3 states vs. 2 then having 2 seems better.

tobymao (approved):

PR #349, opened 2022-08-20
Add support for 'WHERE x IN UNNEST(...)'

sid-kap:
The BigQuery [`IN` operator](https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#in_operators) supports `UNNEST` arguments:
```
search_value [NOT] IN value_set

value_set:
  {
    (expression[, ...])
    | (subquery)
    | UNNEST(array_expression)
  }
```

This adds support for parsing queries with `WHERE x IN UNNEST(...)` clauses.


sid-kap:
It seems like MySQL and Postgres don't support this. Should I set up the parser to only allow this if parsing BigQuery SQL?

tobymao:
in general parsers are more lenient. 

sqlglot should parse any sql, but never generate bad sql

tobymao (approved):

sid-kap:
Hmm. In that case, we should probably make generator.py fail when rendering this in other dialects (or use a workaround like `SELECT UNNEST(...)`)

tobymao:
can you convert IN unnest() into something that works in all dialects?

if not, you can override other dialects or log unsupported_error

tobymao:
i will say though that this is not a requirement to merge this PR in as in unnest is probably pretty rare. but if you want to solve it go for it

sid-kap:
It looks like `SELECT UNNEST` works in postgres! I'll update the PR in the evening

sqlglot/generator.py
```diff
@@ -798,13 +798,18 @@ def if_sql(self, expression):
...
+        if query:
+            in_sql = self.wrap(query)
+        elif unnest:
+            in_sql = self.in_unnest_op(unnest)
+        else:
+            in_sql = f"({self.expressions(expression, flat=True)})"
         return f"{self.sql(expression, 'this')} IN {in_sql}"
 
+    def in_unnest_op(self, unnest):
+        return f"(SELECT {self.unnest_sql(unnest)})"
```
tobymao, line 811: self.sql(unnest)

sqlglot/dialects/bigquery.py
```diff
@@ -92,6 +92,9 @@ class Generator(Generator):
             exp.DataType.Type.TEXT: "STRING",
         }
 
+        def in_unnest_op(self, unnest):
+            return self.unnest_sql(unnest)
```
tobymao, line 96: self.sql(unnest)

sid-kap:
@tobymao Thanks for reviewing/merging these so quickly!
Do you know when the next release is scheduled for?

tobymao:
i need to plan that out, the current main branch has many breaking changes so i may bump to v5

tobymao:
will probably release v5 tomorrow

sid-kap:
Great, thanks so much!

PR #351, opened 2022-08-20
spark to_date takes in a format which makes things more complicated

### LINKED ISSUE #345: `to_date` with formatting is not supported by `spark` dialects
Spark supports [to_date(e: Column, fmt: String):](https://sparkbyexamples.com/spark/spark-sql-date-and-time-functions/#to_date) function, but trying to parse it produces
> sqlglot.errors.ParseError: The number of provided arguments (2) is greater than the maximum number of supported arguments (1)

Steps to reproduce:

```python
>>> import sqlglot
>>> sqlglot.transpile(sqlglot.transpile("""SELECT TO_DATE('2022-1-1', 'yyyy-dd-mm')""", read='mysql', write='spark')[0], read='spark')
```

tobymao:
fixes #345

### DIFF
--- a/tests/test_dialects.py
+++ b/tests/test_dialects.py
@@ -1390,6 +1390,18 @@ class TestDialects(unittest.TestCase):
             identity=False,
         )
         self.validate(
+            "STR_TO_DATE(x, 'yyyy')",
+            "CAST(DATE_FORMAT(x, 'yyyy') AS DATE)",
+            write="hive",
+            identity=False,
+        )
+        self.validate(
+            "STR_TO_DATE(x, 'yyyy-MM-dd')",
+            "CAST(x AS DATE)",
+            write="hive",
+            identity=False,
+        )
+        self.validate(
             "DATE_FORMAT('2020-01-01', 'yyyy-MM-dd HH:mm:ss')",
             "DATE_FORMAT('2020-01-01', '%Y-%m-%d %H:%i:%S')",
             read="hive",
--- a/tests/test_dialects.py
+++ b/tests/test_dialects.py
@@ -801,13 +801,13 @@ class TestDialects(unittest.TestCase):
         )
         self.validate(
             "DATE_PARSE(x, '%Y-%m-%d %H:%i:%s')",
-            "FROM_UNIXTIME(UNIX_TIMESTAMP(x))",
+            "CAST(x AS TIMESTAMP)",
             read="presto",
             write="hive",
         )
         self.validate(
             "DATE_PARSE(x, '%Y-%m-%d')",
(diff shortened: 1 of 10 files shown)


tests/test_dialects.py
```diff
@@ -801,13 +801,13 @@ def test_presto(self):
...
             "DATE_PARSE(x, '%Y-%m-%d %H:%i:%s')",
-            "FROM_UNIXTIME(UNIX_TIMESTAMP(x))",
+            "DATE_FORMAT(x, 'yyyy-MM-dd HH:mm:ss')",
             read="presto",
             write="hive",
         )
         self.validate(
             "DATE_PARSE(x, '%Y-%m-%d')",
-            "FROM_UNIXTIME(UNIX_TIMESTAMP(x, 'yyyy-MM-dd'))",
+            "DATE_FORMAT(x, 'yyyy-MM-dd HH:mm:ss')",
```
barakalon, line 810: Is this right?

barakalon (approved):

tests/test_dialects.py
eakmanrq: This isn't right. `DATE_PARSE` takes a string and returns a timestamp while `DATE_FORMAT` takes a timestamp and returns a string.

tests/test_dialects.py
```diff
@@ -1521,7 +1527,7 @@ def test_hive(self):
         )
         self.validate(
             "TO_DATE(x)",
-            "TS_OR_DS_TO_DATE_STR(x)",
+            "TS_OR_DS_TO_DATE(x)",
```
eakmanrq, line 1530: Looks like this was a bug before. Nice improvement. 👍

tests/test_dialects.py
```diff
@@ -1925,6 +1931,25 @@ def test_spark(self):
...
 
+        self.validate(
+            "TO_DATE(x, 'yyyy-MM-dd')",
+            "CAST(DATE_PARSE(SUBSTR(CAST(x AS VARCHAR), 1, 10), '%Y-%m-%d') AS DATE)",
+            read="spark",
+            write="presto",
+        )
+        self.validate(
+            "TO_DATE(x, 'yyyy')",
+            "DATE_PARSE(x, '%Y')",
```
eakmanrq, line 1942: These are not the same. `TO_DATE` returns a date while `DATE_PARSE` returns a timestamp.

tests/test_dialects.py
```diff
@@ -1925,6 +1931,25 @@ def test_spark(self):
...
+            read="spark",
+            write="presto",
+        )
+        self.validate(
+            "TO_DATE(x, 'yyyy')",
+            "STRPTIME(x, '%Y')",
```
eakmanrq, line 1948: Same top is date and bottom is timestamp

eakmanrq (approved):

PR #353, opened 2022-08-22
Better unix timestamp with scale argument for Snowflake and Spark

saligrama:
In #307, I added support for Snowflake's TO_TIMESTAMP function, which includes the ability to parse timestamps with scale (seconds, milliseconds, or microseconds), where the parsing was done by using integer division to store the timestamp with adjusted scale.

This pull request modifies that to store the scale as a separate parameter. This means that generators for other dialects can choose the correct function based on scale (e.g., Spark has `TIMESTAMP_SECONDS`, `TIMESTAMP_MILLIS`, and `TIMESTAMP_MICROS`); doing it this way avoids lossiness in timestamps and also does not break when placeholders in prepared statements are passed as the timestamp itself.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -54,13 +54,13 @@ def _snowflake_to_timestamp(args):
 
 
 def _unix_to_time(self, expression):
-    scale = expression.args.get("scale")
+    scale = expression.text("scale")
     timestamp = self.sql(expression, "this")
-    if not scale or scale == exp.UnixToTime.SECONDS:
+    if scale in ["", exp.UnixToTime.SECONDS.name]:
         return f"TO_TIMESTAMP({timestamp})"
-    if scale == exp.UnixToTime.MILLIS:
+    if scale == exp.UnixToTime.MILLIS.name:
         return f"TO_TIMESTAMP({timestamp}, 3)"
-    if scale == exp.UnixToTime.MICROS:
+    if scale == exp.UnixToTime.MICROS.name:
         return f"TO_TIMESTAMP({timestamp}, 9)"
 
     raise ValueError("Improper scale for timestamp")
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -22,16 +22,16 @@ def _snowflake_to_timestamp(args):
             return format_time_lambda(exp.StrToTime, "snowflake")(args)
 
         # case: <numeric_expr> [ , <scale> ]
-        if second_arg.this not in ["0", "3", "9"]:
+        if second_arg.name not in ["0", "3", "9"]:
             raise ValueError(
                 f"Scale for snowflake numeric timestamp is {second_arg}, but should be 0, 3, or 9"
             )
 
-        if second_arg.this == "0":
+        if second_arg.name == "0":
             timescale = exp.UnixToTime.SECONDS
-        elif second_arg.this == "3":
+        elif second_arg.name == "3":
             timescale = exp.UnixToTime.MILLIS
-        elif second_arg.this == "3":
+        elif second_arg.name == "3":
             timescale = exp.UnixToTime.MICROS
 
         return exp.UnixToTime(this=first_arg, scale=timescale)
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -54,13 +54,13 @@ def _snowflake_to_timestamp(args):
 
 
 def _unix_to_time(self, expression):
-    scale = expression.text("scale")
+    scale = expression.args.get("scale")
     timestamp = self.sql(expression, "this")
-    if scale in ["", exp.UnixToTime.SECONDS.name]:
+    if scale in [None, exp.UnixToTime.SECONDS]:
         return f"TO_TIMESTAMP({timestamp})"
-    if scale == exp.UnixToTime.MILLIS.name:
+    if scale == exp.UnixToTime.MILLIS:
         return f"TO_TIMESTAMP({timestamp}, 3)"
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/spark.py
```diff
@@ -27,6 +27,21 @@ def _str_to_date(self, expression):
     return f"TO_DATE({this}, {time_format})"
 
 
+def _unix_to_time(self, expression):
+    scale = expression.args.get("scale")
+    timestamp = self.sql(expression, "this")
+    if not scale:
+        return f"FROM_UNIXTIME({timestamp})"
+    if scale == 0:
```
tobymao, line 35: instead of assuming 0,3,9, what do you think if snowflake converts it into "seconds", "millis", "micros"

when parsing it in

tobymao: @saligrama i mean the actual string representation "seconds" so that we don't have implicit assumptions of what 0,3, and 9 mean

saligrama: It might be better to keep it as 0,3,9 for dialects that don't have a way to distinguish between unix epoch timestamp scales; that way we can do something like `new_timestamp = timestamp / (10 ** int(scale))` rather than a bunch of if statements.

saligrama: Definitely open to using that convention though - as far as my purposes go, just need to handle the Snowflake/Spark cases

saligrama: Ah ok, let me make that change

tobymao: also do exp.Literal.string("seconds")

tobymao: and if you had chosen to do numbers, exp.Literal.number(0)

sqlglot/dialects/snowflake.py
```diff
@@ -27,7 +27,14 @@ def _snowflake_to_timestamp(args):
                 f"Scale for snowflake numeric timestamp is {second_arg}, but should be 0, 3, or 9"
             )
 
-        return exp.UnixToTime(this=first_arg, scale=second_arg)
+        if second_arg.this == "0":
```
tobymao, line 30: second_arg.name

sqlglot/dialects/spark.py
```diff
@@ -32,11 +32,11 @@ def _unix_to_time(self, expression):
     timestamp = self.sql(expression, "this")
     if not scale:
         return f"FROM_UNIXTIME({timestamp})"
-    if scale.name == "0":
+    if scale == exp.UnixToTime.SECONDS:
```
tobymao, line 35: on line 31 change it to epression.text("scale")

sqlglot/dialects/snowflake.py
saligrama: fixed

sqlglot/dialects/spark.py
saligrama: fixed

sqlglot/dialects/spark.py
saligrama: done

tobymao (approved):

tests/test_dialects.py
```diff
@@ -2086,7 +2086,7 @@ def test_snowflake(self):
         )
         self.validate(
             "SELECT TO_TIMESTAMP(1659981729000, 3)",
-            "SELECT TO_TIMESTAMP(1659981729)",
+            "SELECT TO_TIMESTAMP(1659981729000, 3)",
```
tobymao, line 2089: this test is wrong, it should be "micros"

sqlglot/dialects/snowflake.py
```diff
@@ -48,6 +53,19 @@ def _snowflake_to_timestamp(args):
     return exp.UnixToTime.from_arg_list(args)
 
 
+def _unix_to_time(self, expression):
+    scale = expression.args.get("scale")
+    timestamp = self.sql(expression, "this")
+    if not scale or scale == exp.UnixToTime.SECONDS:
```
tobymao, line 59: this is wrong

sqlglot/dialects/snowflake.py
```diff
@@ -48,6 +53,19 @@ def _snowflake_to_timestamp(args):
     return exp.UnixToTime.from_arg_list(args)
 
 
+def _unix_to_time(self, expression):
+    scale = expression.args.get("scale")
```
tobymao, line 57: change this to expresison.text("scale")

tests/test_dialects.py
saligrama: I don't see what's wrong here? We're reading and writing back as snowflake, which wants the numerical argument and not a string like "micros"

sqlglot/dialects/snowflake.py
saligrama: fixed

sqlglot/dialects/snowflake.py
saligrama: fixed

tests/test_dialects.py
tobymao: ah ok, can you write a test where you actually test micros then? i guess identity=True by default

sqlglot/dialects/snowflake.py
```diff
@@ -48,6 +53,19 @@ def _snowflake_to_timestamp(args):
     return exp.UnixToTime.from_arg_list(args)
 
 
+def _unix_to_time(self, expression):
+    scale = expression.text("scale")
```
tobymao, line 57: really sorry about this. my recommendation was bad, i'm doing too many things at once.

i think how you had it was fine because scale is an expression, can you revert it back to expression.args.get("scale")

saligrama: fixed

tobymao:
thanks for the quick iteration

PR #356, opened 2022-08-23
Add support for SQLite square bracket identifier enclosures

### LINKED ISSUE #302: sqlglot doesn't support SQLITE square brackets around keywords
See https://www.sqlite.org/lang_keywords.html

The square bracket notation is used in place of other means of escaping keywords - eg, instead of double quotes. 

```
sqlglot.errors.ParseError: Required keyword: 'this' missing for <class 'sqlglot.expressions.Table'>. Line 2, Col: 14.

CREATE TABLE [Track]
(
    [TrackId] INTEGER  NOT NULL,
    [Name] NVARCHAR(200)  NOT NULL,
    [AlbumId] INTEGER,
```

Column 14 is the square bracket.

Another example:
`SELECT [track].[name] FROM [track];`
Runs fine in SQLite, fails to parse:
```
sqlglot.errors.ParseError: Required keyword: 'expression' missing for <class 'sqlglot.expressions.Dot'>. Line 1, Col: 16.
  SELECT [track].[name] FROM [track];
```

crericha:
Also adds support in SQLite for ` identifier enclosures

Fixes #302

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -53,14 +53,14 @@ class _Dialect(type):
         klass.generator_class = getattr(klass, "Generator", Generator)
 
         klass.tokenizer = klass.tokenizer_class(
-            identifier=klass.identifier,
+            identifiers=klass.identifiers,
             escape=klass.escape,
         )
         return klass
 
 
 class Dialect(metaclass=_Dialect):
-    identifier = None
+    identifiers = None
     escape = "'"
     index_offset = 0
     unnest_column_only = False
@@ -142,7 +142,7 @@ class Dialect(metaclass=_Dialect):
         return self.generator_class(
             **{
                 "quote": self.tokenizer_class.QUOTES[0],
-                "identifier": self.identifier,
+                "identifier": list(self.tokenizer.identifiers.items())[0],
                 "escape": self.escape,
                 "index_offset": self.index_offset,
                 "time_mapping": self.inverse_time_mapping,
@@ -223,7 +223,7 @@ def no_trycast_sql(self, expression):
 
 def struct_extract_sql(self, expression):
     this = self.sql(expression, "this")
-    struct_key = self.sql(expression, "expression").replace(self.quote, self.identifier)
+    struct_key = self.sql(exp.Identifier(this=expression.expression, quoted=True))
     return f"{this}.{struct_key}"
 
 
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -52,15 +52,26 @@ class _Dialect(type):
         klass.parser_class = getattr(klass, "Parser", Parser)
         klass.generator_class = getattr(klass, "Generator", Generator)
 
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -142,7 +142,7 @@ def generator(self, **opts):
         return self.generator_class(
             **{
                 "quote": self.tokenizer_class.QUOTES[0],
-                "identifier": self.identifier,
+                "identifier": list(self.tokenizer.identifiers.items())[0],
```
tobymao, line 145: this is kind of a nit, but i think this might hurt performance a bit, we can probably do this in the line 58 section so that we don't have to do a list of items every time generator is called

sqlglot/generator.py
```diff
@@ -130,7 +130,7 @@ def __init__(
         self.time_trie = time_trie
         self.pretty = pretty if pretty is not None else sqlglot.pretty
         self.configured_pretty = self.pretty
-        self.identifier = identifier or '"'
+        self.identifier = identifier or ('"', '"')
```
tobymao, line 133: what do you think if we change it to

self.identifier_start, self.identifier_end = identifier or ('"', '"') that's probably more clear and then in 401 we can just use those

sqlglot/tokens.py
```diff
@@ -566,17 +566,22 @@ class Tokenizer(metaclass=_Tokenizer):
...
-        identifier=None,
+        identifiers=None,
         escape=None,
     ):
         """
         Tokenizer consumes a sql string and produces an array of :class:`~sqlglot.tokens.Token`
 
         Args
-            identifier (str): the identifier character
+            identifier (array): identifier characters and/or tuples when different start and end identifiers
```
tobymao, line 576: list[str|tuple]

sqlglot/tokens.py
```diff
@@ -566,17 +566,22 @@ class Tokenizer(metaclass=_Tokenizer):
...
             escape (str): the escape code character
         """
-        self.identifier = identifier or '"'
+        self.identifiers = dict(
```
tobymao, line 579: sorry again i'm being a performance nut, but this looks expensive, maybe let's make the api identifier only take in an array of tuples, and then have this tuple stuff happen in the class loader

crericha: It's a tradeoff of uglier identifiers definitions in the sql dialects vs having the extra conditional in the dict comprehension. Whichever you prefer is fine by me.

sqlglot/dialects/dialect.py
```diff
@@ -52,15 +52,26 @@ def __new__(cls, clsname, bases, attrs):
...
+            identifiers=identifiers,
             escape=klass.escape,
         )
+
         return klass
 
 
 class Dialect(metaclass=_Dialect):
-    identifier = None
+    identifiers = None
```
tobymao, line 72: identifiers can be ['"'] and then you don't need the or on line 59

tobymao (approved):

PR #359, opened 2022-08-24
mysql date functions

barakalon:
Adding more MySQL support for some date functions.

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -5,9 +5,61 @@ from sqlglot.dialects.dialect import (
     no_paren_current_date_sql,
     no_tablesample_sql,
     no_trycast_sql,
+    date_add_interval,
 )
 from sqlglot.generator import Generator
+from sqlglot.helper import list_get, invert
 from sqlglot.parser import Parser
+from sqlglot.time import format_time
+
+
+def _date_trunc_sql(self, expression):
+    unit = expression.args.get("unit").name
+
+    this = self.sql(expression.this)
+
+    if unit.lower() == "day":
+        return f"DATE({this})"
+
+    if unit.lower() == "week":
+        concat = f"CONCAT(YEAR({this}), ' ', WEEK({this}, 1), ' 1')"
+        date_format = "%Y %u %w"
+    elif unit.lower() == "month":
+        concat = f"CONCAT(YEAR({this}), ' ', MONTH({this}), ' 1')"
+        date_format = "%Y %c %e"
+    elif unit.lower() == "quarter":
+        concat = f"CONCAT(YEAR({this}), ' ', QUARTER({this}) * 3 - 2, ' 1')"
+        date_format = "%Y %c %e"
+    elif unit.lower() == "year":
+        concat = f"CONCAT(YEAR({this}), ' 1 1')"
+        date_format = "%Y %c %e"
+    else:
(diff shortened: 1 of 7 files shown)


tests/test_dialects.py
```diff
@@ -2359,3 +2360,68 @@ def test_clickhouse(self):
             read="spark",
             write="clickhouse",
         )
+
```
barakalon, line 2363: I'm hoping a structure like this (or something similar... ideas welcomed) can help organize these dialect tests.

Still a ton of tests to move here. 

The basic idea is that the bulk of the test suite should be based on parsing/generating to/from the _abstract syntax_. Not to some random other dialect.

We have a separate test method for generate and parsing since not all operations are reversible.

I hope this clarifies the task of adding complete support for a dialect.

sqlglot/dialects/mysql.py
```diff
@@ -5,9 +5,61 @@
...
+    date_add_interval,
 )
 from sqlglot.generator import Generator
+from sqlglot.helper import list_get, invert
 from sqlglot.parser import Parser
+from sqlglot.time import format_time
+
+
+def _date_trunc_sql(self, expression):
+    unit = expression.args.get("unit").name
```
tobymao, line 17: expression.text("unit")

sqlglot/dialects/mysql.py
```diff
@@ -5,9 +5,61 @@
...
+
+
+def _str_to_date(args):
+    date_format = format_time(list_get(args, 1).name, MySQL.time_mapping)
+    return exp.StrToDate(this=list_get(args, 0), format=exp.Literal.string(date_format))
+
+
+def _str_to_date_sql(self, expression):
+    date_format = format_time(
+        expression.args.get("format").name, invert(MySQL.time_mapping)
```
tobymao, line 50: MySQL.inverse_time_mapping already exists

tobymao: you can also pass in the time_trie 

should you be calling self.format_time though? format_time from time is the lower level method

sqlglot/helper.py
```diff
@@ -125,3 +125,7 @@ def csv_reader(table):
         yield csv_.reader(file, delimiter=delimiter)
     finally:
         file.close()
+
+
+def invert(dictionary):
```
tobymao, line 130: this may not be necessary if you use the premade inverse_time stuff or self.format_time directly

tests/test_dialects.py
tobymao: but then how do you test the reading of dialects? 

i think the structure is ok, but we need to be able to define the read

presto -> spark is the main thing i want to test, as it's an actual use case
presto -> sqlglot is ok, but it's not really end to end

tobymao: what if we created 1 file per dialect, and each dialect always reads that dialect, and then we keep this structure?

sqlglot/dialects/mysql.py
barakalon: whoops

sqlglot/dialects/mysql.py
barakalon: Ah, ok, I hadn't dug deep enough into format_time.

I think all the FUNCTION callables need a `self` parameter. Yikes.

tobymao (approved):

PR #370, opened 2022-08-26
breakout oracle tests

barakalon:


tests/dialects/test_dialect.py
```diff
@@ -1910,6 +1909,13 @@ def test_operators(self):
...
                 "presto": "SELECT a AS b FROM x GROUP BY 1",
                 "hive": "SELECT a AS b FROM x GROUP BY 1",
+                "oracle": "SELECT a AS b FROM x GROUP BY 1",
                 "spark": "SELECT a AS b FROM x GROUP BY 1",
             },
         )
+        self.validate_all(
+            "SELECT x FROM y LIMIT 10",
+            read={
+                "oracle": "SELECT TOP 10 x FROM y",
```
mpf82, line 1919: Oracle has no "SELECT TOP" syntax, instead, starting with Oracle 12, the syntax is ``FETCH FIRST number ROWS ONLY``

barakalon: Huh, interesting. I was just restructuring tests. Seems like this was already an issue.

I don't have much experience with Oracle. @mpf82 - maybe you have the bandwidth to add better support for Oracle?

tobymao: sqlglot allows invalid input, it’s only the output that must be correct. so it’s not neccisarily a bug that oracle allows for top

PR #385, opened 2022-09-01
Don't remove unused selection if referenced in GROUP BY

barakalon:


tests/fixtures/optimizer/tpc-h/tpc-h.sql
```diff
@@ -1043,6 +1043,7 @@ SELECT
   COUNT(*) AS "custdist"
 FROM (
   SELECT
+    "customer"."c_custkey" AS "c_custkey",
```
barakalon, line 1046: This isn't really ideal. But right now we support queries like this:

```sql
SELECT 
  x.b, 
  x.a 
FROM x 
LEFT JOIN y 
GROUP BY b
```

"b" is ambiguous, but some dialects might assume "b" is the output column "x.b". At least we have tests that make that assumption...

sqlglot/optimizer/qualify_columns.py
```diff
@@ -122,6 +123,30 @@ def _expand_using(scope, schema):
...
 
+def _expand_group_numbers(scope):
+    selects = scope.expression.args.get("expressions")
+    group = scope.expression.args.get("group")
+    if not group:
+        return
+
+    new_expressions = []
+    for expression in group.args.get("expressions"):
+        if isinstance(expression, exp.Literal) and expression.is_int:
```
tobymao, line 134: do you mind refactoring so is_int and is_string is on base object so we don't always have to check if the expression is a literal?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -122,6 +123,30 @@ def _expand_using(scope, schema):
...
+            try:
+                select = selects[int(expression.this) - 1]
+            except IndexError:
+                # pylint: disable=raise-missing-from
+                raise OptimizeError(f"Unknown output column: {expression.name}")
+            if isinstance(select, exp.Alias):
+                select = select.this
+            new_expressions.append(select.copy())
```
tobymao, line 142: what's this copy for?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -122,6 +123,30 @@ def _expand_using(scope, schema):
                 scope.replace(column, replacement)
 
 
+def _expand_group_numbers(scope):
```
tobymao, line 126: do we want to expand group numbers?

sqlglot/optimizer/qualify_columns.py
barakalon: we need to copy the expression, or else we'll have one object with 2 parents (well, 1 parent, but I think you get the point)

barakalon:
Alright, this should be ready for another look

sqlglot/optimizer/qualify_columns.py
```diff
@@ -122,40 +125,87 @@ def _expand_using(scope, schema):
...
+            select = selects.get(node.name)
+            if select:
+                scope.clear_cache()
+                if isinstance(select, exp.Alias):
+                    select = select.this
+                return select.copy()
+
+        return node
+
+    group = group.transform(transform)
```
tobymao, line 154: transform is pretty expensive, can we do a copy=False here?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -122,40 +125,87 @@ def _expand_using(scope, schema):
...
+    ):
+        node.replace(new_node)
+
+
+def _expand_positional_references(scope, expressions):
+    new_nodes = []
+    for node in expressions:
+        if node.is_int:
+            try:
+                select = scope.selects[int(node.this) - 1]
```
tobymao, line 177: node.name since it's for sure the text representation

tobymao (approved):
can you do a quick perf check on qualify columns to see that perf hasn't completely tanked?

sqlglot/optimizer/qualify_columns.py
tobymao, line 154: should we do a copy=False here for perf?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -122,40 +125,87 @@ def _expand_using(scope, schema):
...
+
 
+def _expand_order_by(scope):
+    order = scope.expression.args.get("order")
+    if not order:
+        return
+
+    expressions = [o.this for o in order.expressions]
+    for node, new_node in zip(
+        expressions, _expand_positional_references(scope, expressions)
```
tobymao, line 167: expressions can just be an inline generator (o.this ...)

tobymao: or just do zip(order.expressions) and then 169 ordered.this.replace

barakalon: I'm not sure I follow completely, let me know if the replacement is OK.

barakalon:
I'm not seeing a noticeable performance hit.

Here's what I'm using to benchmark:

```python
from sqlglot import parse_one
import timeit


from tests.helpers import TPCH_SCHEMA, load_sql_fixture_pairs


def optimize_tpc_h():
    for _, sql, _ in load_sql_fixture_pairs(f"optimizer/tpc-h/tpc-h.sql"):
        optimize(
            expression=parse_one(sql),
            schema=TPCH_SCHEMA
        )


print(timeit.timeit(optimize_tpc_h, number=100))
```

sqlglot/optimizer/qualify_columns.py
tobymao: i didn't realize you use it twice, what you had was fine, sorry for the swirl

PR #388, opened 2022-09-02
Support for nchar, varchar2, nvarchar. and nvarchar2

mpf82:
Support for NCHAR, VARCHAR2, NVARCHAR. and NVARCHAR2

Added support for NCHAR, VARCHAR2, NVARCHAR. and NVARCHAR2

By default VARCHAR2 will be tokenized as VARCHAR and NVARCHAR2 will be tokenized as NVARCHAR

By default NCHAR maps to CHAR and NVARCHAR maps to VARCHAR

Also made "POWER" the default over "POW", as it's supported by more SQL languages.


mpf82:
By the way, ``test_optimizer`` fails on my system, but it seems to be unrelated to my changes. The other tests seem all to be ok.

tobymao (approved):

tobymao:
ah in order to pass the optimizer tests, you need to install requirements.txt because it uses date-utils

pip install -r requirements.txt

PR #392, opened 2022-09-03
WIP: support MySQL 0xFF-style hex literals

rolandwalker:
First try.

These changes to `tokens.py` were sufficient to make tests pass.

But looking at `.walk()` I see the 0xFF-style literals are parsed as identifiers.  We want literals.  Looks like the next step would be `sqlglot/expressions.py`?  However the logic there seems centered around the idea that anything not string is a number.

Binary literals are mostly like strings.  But they must not be quoted.

What is better: a binary token type, distinct from both string and number?  Or a string type which is marked as unquoted (somehow)?

Also there's nothing here about making the construct specific to a dialect.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -1,4 +1,5 @@
 from enum import auto
+from string import hexdigits
 
 from sqlglot.helper import AutoName
 from sqlglot.trie import in_trie, new_trie
@@ -627,6 +628,8 @@ class Tokenizer(metaclass=_Tokenizer):
                 if white_space == TokenType.BREAK:
                     self._col = 1
                     self._line += 1
+            elif self._char == "0" and self._peek == "x":
+                self._scan_hex_literal()
             elif self._char.isdigit():
                 self._scan_number()
             elif identifier_end:
@@ -776,6 +779,27 @@ class Tokenizer(metaclass=_Tokenizer):
                 return self._advance(-len(literal))
             else:
                 return self._add(TokenType.NUMBER)
+
+    def _scan_hex_literal(self):
+        prefix = "0x"
+        text = prefix
+        self._advance()
+
+        while True:
+            if not self._peek:
+                if text == prefix:
+                    return self._add(TokenType.IDENTIFIER, text)
+                return self._add(TokenType.BINARY, text)
+
+            self._advance()
+
+            if self._char in hexdigits:
+                text = text + self._char.upper()
+            elif text == prefix:
+                self._advance(-1)
+                return self._add(TokenType.IDENTIFIER, text)
+            else:
+                return self._add(TokenType.BINARY, text)
 
     def _scan_string(self, quote):
         if quote not in self.QUOTES:
--- a/tests/fixtures/identity.sql
+++ b/tests/fixtures/identity.sql
@@ -7,6 +7,7 @@ SUM(CASE WHEN x > 1 THEN 1 ELSE 0 END) / y
 1E-2
 1.1E10
 1.12e-10
+0x0A
 -11.023E7 * 3
 (1 * 2) / (3 - 5)
 ((TRUE))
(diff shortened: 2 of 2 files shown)


tobymao:
https://dev.mysql.com/doc/refman/8.0/en/hexadecimal-literals.html 

looks like there’s also x’’

rolandwalker:
In the `x''` form the `x` can be either case, but in the `0x` form the `x` must be lowercase.

sqlglot/tokens.py
```diff
@@ -777,6 +779,27 @@ def _scan_number(self):
...
 
+    def _scan_hex_literal(self):
+        prefix = "0x"
+        text = prefix
+        self._advance()
+
+        while True:
+            if not self._peek:
+                if text == prefix:
+                    return self._add(TokenType.IDENTIFIER, text)
```
tobymao, line 790: i don't think this is correct,

if you de select 0x, it should actually be select 0 AS x

sqlglot/tokens.py
```diff
@@ -627,6 +627,8 @@ def tokenize(self, sql):
                 if white_space == TokenType.BREAK:
                     self._col = 1
                     self._line += 1
+            elif self._char == "0" and self._peek == "x":
```
tobymao, line 630: i think this needs to move into the scan_number branch to handle the select 0x -> 0 as x case

sqlglot/tokens.py
rolandwalker: Based on:
```
mysql> select @@version;
+-----------+
| @@version |
|-----------|
| 8.0.22-13 |
+-----------+
mysql> select 0x;
(1054, "Unknown column '0x' in 'field list'")
mysql> select `0x`;
(1054, "Unknown column '0x' in 'field list'")
mysql> select 0x from table_that_exists;
(1054, "Unknown column '0x' in 'field list'")
mysql> select `0x` from table_that_exists;
(1054, "Unknown column '0x' in 'field list'")
```

sqlglot/tokens.py
```diff
@@ -777,6 +779,27 @@ def _scan_number(self):
...
+
+        while True:
+            if not self._peek:
+                if text == prefix:
+                    return self._add(TokenType.IDENTIFIER, text)
+                return self._add(TokenType.BINARY, text)
+
+            self._advance()
+
+            if self._char in "0123456789abcdefABCDEF":
```
tobymao, line 795: import string

self._char in string.hexdigits

sqlglot/tokens.py
rolandwalker: But if another database tokenizes it differently … that's tricky.

sqlglot/tokens.py
```diff
@@ -777,6 +779,27 @@ def _scan_number(self):
             else:
                 return self._add(TokenType.NUMBER)
 
+    def _scan_hex_literal(self):
```
tobymao, line 782: you may be able to move this whole method into _scan_number

also, can you test out mysql's behavior on how it handles something like 0xbaz ? if it errors out, then we don't need the complicated logic

sqlglot/tokens.py
rolandwalker: postgres: 0 as x
duckdb: 0 as x
sqlite3: Parse error: unrecognized token: "0x"
bigquery: Request contains an invalid argument

tobymao: postgres handles select 0x as select 0 as x which is more correct to me as you can't have ro shouldn't have identifiers prefixed with a 0. so i think it's ok to stick with postgres's interpretation on this one

tobymao: in general 0x should probably be considered user error, but i think i'd prefer to do 0 as x

rolandwalker: Right.  It's such a trivial edge case, and always going to be a mistake.  Although — the Postgres interpretation is influenced by the context that Postgres does not support this form of literal.

sqlglot/tokens.py
rolandwalker: `"Unknown column '0xbaz' in 'field list'`

tobymao:
closed in favor of https://github.com/tobymao/sqlglot/pull/397

PR #395, opened 2022-09-05
Added basic support for CREATE [UNIQUE]  INDEX

mpf82:
Added basic support for CREATE [UNIQUE]  INDEX

References:
* https://www.sqlite.org/lang_createindex.html
* https://www.postgresql.org/docs/current/sql-createindex.html

@tobymao I was thinking of using ``table = self._parse_table(schema=True)`` instead of ``table = self.expression(exp.Table, this=self._parse_id_var())`` in the funtion ``_parse_index``, but then the table parsing would have consomed the columns as "simple identifiers", whereas now the columns are parsed as a tuple of COLUMN - which I personally find much nicer.

Any advice?


sqlglot/parser.py
```diff
@@ -383,7 +383,7 @@ class Parser:
         "offset": lambda self: self._parse_offset(),
     }
 
-    CREATABLES = {TokenType.TABLE, TokenType.VIEW, TokenType.FUNCTION}
+    CREATABLES = {TokenType.TABLE, TokenType.VIEW, TokenType.FUNCTION, TokenType.INDEX}
```
tobymao, line 386: you also need to add index to id_var so that you can do select 1 index -> select 1 as index

sqlglot/parser.py
```diff
@@ -1096,6 +1099,16 @@ def _parse_join(self):
 
         return self.expression(exp.Join, **kwargs)
 
+    def _parse_index(self):
+        index = self._parse_id_var()
+        self._match(TokenType.ON)
+        self._match(TokenType.TABLE)  # hive
+        # table = self._parse_table(schema=True)
```
tobymao, line 1106: if you jus do parse_table(schema=True), then you don't even need columns because now the definition resides in the table schema.

although it's a bit more complex for the index use case, it makes sqlglot more consistent, is it ok if we stick with parse_table(schema=True) ?

mpf82: Well, "ok" is relative, I'd say. For me it would be okay, but if we use ``self._parse_table(schema=True)``, then suddenly this is "valid":

> CREATE UNIQUE INDEX IF NOT EXISTS my_idx ON TABLE tbl (a INT, b)

because it's a valid column definition, but it's definitely wrong SQL.

It has also an influence on the pretty printing.

Compare 

    CREATE UNIQUE INDEX IF NOT EXISTS my_idx ON tbl (a, b)

to

    CREATE UNIQUE INDEX IF NOT EXISTS my_idx ON tbl (
      a,
      b
    )

though that I don't mind, we're using a custom pretty printer anyway.

sqlglot/parser.py
mpf82: fixed in #395

PR #397, opened 2022-09-05
Hex

tobymao:


rolandwalker:
Interesting solution!

alippai:
Should this support:
```sql
select 0x10 & 48 = 16
```
? (evaluates to `1` in MySQL)

tobymao:
isn't this supported already?

alippai:
I thought it still raises an exception on 6.0. I made a mistake with the environments

PR #399, opened 2022-09-06
Refactor quotes to generalize support for non-matching quote prefix and postfixes

crericha:


sqlglot/dialects/bigquery.py
```diff
@@ -34,12 +34,25 @@ def func(self, expression):
 
 
 class BigQuery(Dialect):
-    identifiers = ["`"]
-    escape = "\\"
     unnest_column_only = True
 
     class Tokenizer(Tokenizer):
-        QUOTES = ["'", '"', '"""']
+        QUOTES = [
```
tobymao, line 40: this can be dryed up by doing 
[
for [...]
for ['r'...]
[

sqlglot/generator.py
```diff
@@ -620,8 +630,10 @@ def literal_sql(self, expression):
         if expression.is_string:
             if self._replace_backslash:
                 text = text.replace("\\", "\\\\")
-            text = text.replace(self.quote, self._escaped_quote)
-            return f"{self.quote}{text}{self.quote}"
+            text = text.replace(self.quote_start, self._escaped_quote_start)
```
tobymao, line 633: it sucks that you have to check escape_quote != end every time, maybe instead of self._escaped_quote_start and end, we store a lambda and just call that 

if ...
  self._escape_func = lambda...
else
  ...

sqlglot/tokens.py
```diff
@@ -270,9 +270,17 @@ class _Tokenizer(type):
     def __new__(cls, clsname, bases, attrs):
         klass = super().__new__(cls, clsname, bases, attrs)
 
-        klass.QUOTES = [
-            prefix + quote for prefix in ["", "r"] for quote in klass.QUOTES
-        ]
+        klass.QUOTES = dict(
+            (quote[0], quote[1]) if isinstance(quote, tuple) else (quote, quote)
```
tobymao, line 274: let's switch all of these to be if isinstance(quote, str) inverse so that tuples, and lists could work

sqlglot/tokens.py
```diff
@@ -578,20 +588,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "_peek",
     )
 
-    def __init__(
-        self,
-        identifiers=None,
-        escape=None,
-    ):
-        """
```
tobymao, line 586: let's keep the arg string?

sqlglot/generator.py
crericha: I'm wondering if escape_quote_start doesn't matter here actually and can collapse down. From a parsing perspective, it just can't be the escape_quote_end, right?

tobymao: oh yea i wonder if this is a bug?

crericha: Updated to match my comment.

sqlglot/tokens.py
crericha: Fixed

sqlglot/tokens.py
crericha: Fixed

sqlglot/dialects/bigquery.py
crericha: Fixed

sqlglot/dialects/bigquery.py
```diff
@@ -34,12 +34,16 @@ def func(self, expression):
...
+            quote if len(prefix) == 0 else (prefix + quote, quote)
```
tobymao, line 41: if prefix else

in python "" is falsey

sqlglot/generator.py
```diff
@@ -90,6 +92,8 @@ class Generator:
         "time_trie",
         "pretty",
         "configured_pretty",
+        "quote_start",
+        "quote_end",
```
tobymao, line 96: delete quote from args and the doc string?

tobymao: and slots

tobymao (approved):

sqlglot/dialects/bigquery.py
crericha: Fixed

sqlglot/generator.py
crericha: Fixed

Issue #387, opened 2022-09-02
Postgres: Different UNIQUE notations


mpf82:
Postgres supports various ways to declare a column as unique.

sqlglot currently supports: 
```sql
CREATE TABLE products (
    product_no integer UNIQUE,
    name text,
    price numeric
);
```

It would be nice to also support:
```sql
CREATE TABLE example (
    a integer,
    b integer,
    c integer,
    UNIQUE (a, c)
);
```

and with a named constraint:
```sql
CREATE TABLE products (
    product_no integer CONSTRAINT must_be_different UNIQUE,
    name text,
    price numeric
);
```

Reference: https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS

tobymao:
low priority, happy to help you accomplish this in a PR, or may take a week

tobymao:
fixed by #405

PR #405, opened 2022-09-07
handle different ways of expressing constraints in postgres

ginter:
addresses https://github.com/tobymao/sqlglot/issues/387

pg spec: https://www.postgresql.org/docs/current/sql-createtable.html


sqlglot/parser.py
```diff
@@ -1738,39 +1754,54 @@ def parse_option(option, option_lambda):
...
+                and "NOT NULL",
             )
             parse_option(
                 "comment",
                 lambda: self._match(TokenType.SCHEMA_COMMENT) and self._parse_string(),
             )
             parse_option(
                 "primary",
-                lambda: self._match(TokenType.PRIMARY_KEY),
+                lambda: self._match(TokenType.PRIMARY_KEY) and "PRIMARY KEY",
```
ginter, line 1767: cheating a little bit with the return value to avoid having to create additional exp classes and handlers in the generator

tobymao: what's going on here?

sqlglot/parser.py
```diff
@@ -1715,16 +1720,27 @@ def _parse_column_def(self, this):
             "unique": None,
             "parsed": True,
         }
+        constraint = None
 
         def parse_option(option, option_lambda):
+            nonlocal constraint
```
tobymao, line 1726: why are you using a non local?

tobymao: ah because constraint is defined in 1723

tobymao: can you just use the options dictionary for this?

sqlglot/generator.py
```diff
@@ -306,7 +306,11 @@ def column_sql(self, expression):
     def columndef_sql(self, expression):
         column = self.sql(expression, "this")
         kind = self.sql(expression, "kind")
-        not_null = " NOT NULL" if expression.args.get("not_null") else ""
+        not_null = (
```
tobymao, line 309: should we instead refactor ColumnDef to take constraints / options ?

ginter: I was trying to work with what was there but agree that a refactor is probably in order. see if this feels better

sqlglot/parser.py
ginter: no bc most entries in the dict can themselves be a named constraint

sqlglot/generator.py
```diff
@@ -306,21 +306,43 @@ def column_sql(self, expression):
...
         kind = self.sql(expression, "kind")
-        not_null = " NOT NULL" if expression.args.get("not_null") else ""
-        default = self.sql(expression, "default")
-        default = f" DEFAULT {default}" if default else ""
-        auto_increment = (
-            " " + self.token_sql(TokenType.AUTO_INCREMENT)
-            if expression.args.get("auto_increment")
-            else ""
+
+        if len(expression.args.get("constraints")) == 0:
```
tobymao, line 310:
```suggestion
        constraints = expression.args.get("constraints")
        
        if not constraints:
            return ...
```

sqlglot/generator.py
```diff
@@ -306,21 +306,43 @@ def column_sql(self, expression):
...
+            return f"{column} {kind}"
+
+        constraints = " ".join(
```
tobymao, line 313:
```suggestion
        constraints = self.expressions(expression, key="constraints", sep= " ")
        
        if not constraints:
           return ...
         
```

sqlglot/generator.py
tobymao: scratch that, next one is cleaner

sqlglot/generator.py
```diff
@@ -306,21 +306,43 @@ def column_sql(self, expression):
...
-        collate = f" COLLATE {collate}" if collate else ""
-        comment = self.sql(expression, "comment")
-        comment = f" COMMENT {comment}" if comment else ""
-        primary = " PRIMARY KEY" if expression.args.get("primary") else ""
-        unique = " UNIQUE" if expression.args.get("unique") else ""
-        return f"{column} {kind}{not_null}{default}{collate}{comment}{unique}{primary}{auto_increment}"
+        return f"{column} {kind} {constraints}"
+
+    def columnconstraint_sql(self, expression):
+        this = expression.args.get("this")
```
tobymao, line 319:
```suggestion
        this = self.sql(expression, "this")
```

tobymao:
this is much better

sqlglot/expressions.py
```diff
@@ -558,16 +558,42 @@ class ColumnDef(Expression):
...
+class ColumnConstraint(Expression):
+    arg_types = {"this": False, "kind": True}
+
+
+class AutoIncrementColumnConstraint(Expression):
+    pass
+
+
+class CollateColumnConstraint(Expression):
+    arg_types = {"this": True}
```
tobymao, line 574: this is the default, you can just call pass

sqlglot/expressions.py
```diff
@@ -558,16 +558,42 @@ class ColumnDef(Expression):
...
+
+
+class CommentColumnConstraint(Expression):
+    arg_types = {"value": True}
```
tobymao, line 578:
```suggestion
    pass
```

sqlglot/expressions.py
```diff
@@ -558,16 +558,42 @@ class ColumnDef(Expression):
...
+
+
+class DefaultColumnConstraint(Expression):
+    arg_types = {"expression": True}
```
tobymao, line 582:
```suggestion
    pass
```

sqlglot/expressions.py
```diff
@@ -593,6 +619,10 @@ class ForeignKey(Expression):
     }
 
 
+class Unique(Expression):
+    arg_types = {"columns": True}
```
tobymao, line 623:
```suggestion
    arg_types = {"expressions": True}
```

Issue #389, opened 2022-09-02
CHECK constraint support missing


mpf82:
Similar to the [UNIQUE constraint](https://github.com/tobymao/sqlglot/issues/387), there seem to be several ways to define a CHECK constraint - either as a column constraint or as a table constraint.

The following statement is valid Postgres and contains a column check constraint, a named column check constraint, a table check constraint and a named table check constraint:

```sql
CREATE TABLE products (
    product_no INTEGER,
    name TEXT,
    price NUMERIC CHECK (price > 0),
    discounted_price NUMERIC CONSTRAINT positive_discount CHECK (discounted_price > 0), 
    CHECK (product_no > 1),
    CONSTRAINT valid_discount CHECK (price > discounted_price)
);
```

References:
* https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS
* https://sqlite.org/syntax/table-constraint.html
* https://sqlite.org/syntax/column-constraint.html

tobymao:
low priority, happy to help you accomplish this in a PR, or may take a week

PR #408, opened 2022-09-08
add support for table and column check constraints

ginter:
add support for table and column check constraints. fixes https://github.com/tobymao/sqlglot/issues/389

references:
https://www.postgresql.org/docs/current/sql-createtable.html
https://www.sqlite.org/syntax/column-constraint.html
https://dev.mysql.com/doc/refman/8.0/en/create-table.html


tobymao:
i noticed you didn't parse parenthesis, should do that since it seems to be required on check() ?

PR #409, opened 2022-09-08
added BLOB and CLOB

mpf82:
added BLOB and CLOB

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -68,7 +68,9 @@ class TokenType(AutoName):
     VARCHAR = auto()
     NVARCHAR = auto()
     TEXT = auto()
+    CLOB = auto()
     BINARY = auto()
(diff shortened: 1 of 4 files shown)


sqlglot/tokens.py
```diff
@@ -68,7 +68,9 @@ class TokenType(AutoName):
     VARCHAR = auto()
     NVARCHAR = auto()
     TEXT = auto()
+    CLOB = auto()
```
tobymao, line 71: these are unnecessary as you don't use these tokens
```suggestion
```

sqlglot/tokens.py
```diff
@@ -68,7 +68,9 @@ class TokenType(AutoName):
...
     BINARY = auto()
+    BLOB = auto()
```
tobymao, line 73:
```suggestion
```

PR #415, opened 2022-09-09
feat: Add sort_by and cluster_by to expression builder

robert8138:
We already have SORT & CLUSTER expression and SORT_BY & CLUSTER_BY tokens, they are currently used in DDLs but they can also be used in expression builder. This PR adds that support.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1155,6 +1155,76 @@ class Select(Subqueryable, Expression):
             **opts,
         )
 
+    def sort_by(
+        self, *expressions, append=True, dialect=None, parser_opts=None, copy=True
+    ):
+        """
+        Set the CLUSTER BY expression.
+
+        Example:
+            >>> Select().from_("tbl").select("x").sort_by("x DESC").sql()
+            'SELECT x FROM tbl SORT BY x DESC'
+
+        Args:
+            *expressions (str or Expression): the SQL code strings to parse.
+                If a `Group` instance is passed, this is used as-is.
+                If another `Expression` instance is passed, it will be wrapped in a `SORT`.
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -1155,6 +1155,76 @@ def order_by(self, *expressions, append=True, dialect=None, copy=True, **opts):
             **opts,
         )
 
+    def sort_by(
+        self, *expressions, append=True, dialect=None, parser_opts=None, copy=True
```
tobymao, line 1159: change parse_opts into **opts kwargs, look at the other examples

sqlglot/expressions.py
```diff
@@ -1155,6 +1155,76 @@ def order_by(self, *expressions, append=True, dialect=None, copy=True, **opts):
...
+    ):
+        """
+        Set the CLUSTER BY expression.
```
tobymao, line 1162: sort by

sqlglot/expressions.py
```diff
@@ -1155,6 +1155,76 @@ def order_by(self, *expressions, append=True, dialect=None, copy=True, **opts):
...
+            append=append,
+            copy=copy,
+            prefix="SORT BY",
+            into=Sort,
+            dialect=dialect,
+            parser_opts=parser_opts,
+        )
+
+    def cluster_by(
+        self, *expressions, append=True, dialect=None, parser_opts=None, copy=True
```
tobymao, line 1194: opts kwargs, don't forget to change the docstring as well

robert8138:
Thanks for the quick review. Addressed comments.

tobymao (approved):

Issue #421, opened 2022-09-12
Parser does not understand 'from' keyword in Postgres substring function


g-u-t-h-m-a-n:
Example (valid in Postgres 14.4):

```
select case when substring('abcdefg' from 1 for 2) in ('ab') then 1 else 0 end;
```

Result:
```
ParseError: Required keyword: 'start' missing for <class 'sqlglot.expressions.Substring'>. Line 1, Col: 38.
  select case when substring('abcdefg' from 1 for 2) in ('ab') then 1 else 0 end;
```

tobymao:
thanks for the report

PR #427, opened 2022-09-13
Support alternative syntax for postgres substring

GeorgeSittas:
Initial draft aiming to solve [#421](https://github.com/tobymao/sqlglot/issues/421).

- Is there a way to avoid changing the `arg_types` of `Substring` globally, so that the changes will only apply to a specific dialect? In this commit, I changed it globally so now `substring('foo')` is not regarded as a parse error. This might be true for postgres but not for other dialects. Is this a problem?

- According to the [docs](https://www.postgresql.org/docs/9.1/functions-string.html) (Table 9-6) [1], postgres allows for regex matching inside the substring function. This is currently supported, but the `Substring` attributes have misleading names. Ideally, we'd want `pattern` and `escape`, instead of `start` and `length`, if we have string literals in the `from`, `for` parts [2]. I can try clearing this up a little if needed.

- Finally, I made it so that postgres now generates sql code for the substring function in the from..for.. syntax, which is also compatible with the csv syntax.

[1] These features are supported in postgres 14.5 as well.
[2] Postgres 14.5 also supports the `SUBSTRING(text SIMILAR pattern ESCAPE escape` syntax.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -115,6 +115,7 @@ class Parser:
         TokenType.FALSE,
         TokenType.FIRST,
         TokenType.FOLLOWING,
+        TokenType.FOR,
         TokenType.FORMAT,
         TokenType.FUNCTION,
         TokenType.IF,
@@ -170,6 +171,7 @@ class Parser:
         TokenType.PRIMARY_KEY,
         TokenType.REPLACE,
         TokenType.ROW,
+        TokenType.SUBSTRING,
         TokenType.UNNEST,
         TokenType.VAR,
         TokenType.LEFT,
@@ -365,6 +367,7 @@ class Parser:
     FUNCTION_PARSERS = {
         TokenType.CONVERT: lambda self, _: self._parse_convert(),
         TokenType.EXTRACT: lambda self, _: self._parse_extract(),
+        TokenType.SUBSTRING: lambda self, _: self._parse_substring(),
         **{
             token_type: lambda self, token_type: self._parse_cast(
                 self.STRICT_CAST and token_type == TokenType.CAST
@@ -1916,6 +1919,28 @@ class Parser:
         else:
             to = None
         return self.expression(exp.Cast, this=this, to=to)
+
+    def _parse_substring(self):
+        this = self._parse_term()
+
+        # Postgres supports the form: substring(string [from int] [for int])
+        # https://www.postgresql.org/docs/9.1/functions-string.html @ Table 9-6
+        matching_delimiter = {
+            TokenType.FROM: TokenType.FOR,
+            TokenType.COMMA: TokenType.COMMA,
+        }
+
+        start = None
+        length = None
+
+        if self._match_set((TokenType.FROM, TokenType.COMMA)):
+            delim = self._prev
+            start = self._parse_term()
+
+            if self._match(matching_delimiter[delim.token_type]):
+                length = self._parse_term()
+
+        return self.expression(exp.Substring, this=this, start=start, length=length)
 
     def _parse_window(self, this, alias=False):
         if self._match(TokenType.FILTER):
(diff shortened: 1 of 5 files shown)


sqlglot/parser.py
```diff
@@ -1917,6 +1920,28 @@ def _parse_convert(self):
             to = None
         return self.expression(exp.Cast, this=this, to=to)
 
+    def _parse_substring(self):
+        this = self._parse_term()
+
+        # Postgres supports the form: substring(string [from int] [for int])
+        # https://www.postgresql.org/docs/9.1/functions-string.html @ Table 9-6
+        matching_delimiter = {
```
tobymao, line 1928: let's not do it like this, is it possible to instead check if there's a from, and if so just parse the for, or parse the csv?

sqlglot/tokens.py
```diff
@@ -416,6 +418,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "FULL": TokenType.FULL,
         "FUNCTION": TokenType.FUNCTION,
         "FOLLOWING": TokenType.FOLLOWING,
+        "FOR": TokenType.FOR,
```
tobymao, line 421: you don't need to add this because you already add FOR in the postgres parser

sqlglot/parser.py
```diff
@@ -1917,6 +1920,28 @@ def _parse_convert(self):
...
+        length = None
+
+        if self._match_set((TokenType.FROM, TokenType.COMMA)):
+            delim = self._prev
+            start = self._parse_term()
+
+            if self._match(matching_delimiter[delim.token_type]):
+                length = self._parse_term()
+
+        return self.expression(exp.Substring, this=this, start=start, length=length)
```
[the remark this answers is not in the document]
tobymao: maybe if parse substring fails you can retreat?

sqlglot/parser.py
```diff
@@ -1917,6 +1920,29 @@ def _parse_convert(self):
...
+
+            if self._match(TokenType.FOR):
+                args.append(self._parse_term())
+        elif self._match(TokenType.COMMA):
+            if self._match(TokenType.R_PAREN):
+                self.raise_error("Unexpected ) token")
+
+            args.extend(self._parse_csv(self._parse_term))
+
+        function = self.FUNCTIONS.get(exp.Substring.sql_names()[0])
```
tobymao, line 1940: can’t you just get this method directly?

Substring.from_args_list

sqlglot/parser.py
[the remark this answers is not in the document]
GeorgeSittas: Ah, I see. I'm not sure if that's simple to do with the current implementation, because once we match against the function's parser we lose the option of executing the else branch.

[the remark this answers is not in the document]
tobymao: that’s ok

sqlglot/parser.py
```diff
@@ -1917,6 +1920,28 @@ def _parse_convert(self):
...
+        # Postgres supports the form: substring(string [from int] [for int])
+        # https://www.postgresql.org/docs/9.1/functions-string.html @ Table 9-6
+
+        if self._match(TokenType.FROM):
+            args.append(self._parse_term())
+
+            if self._match(TokenType.FOR):
+                args.append(self._parse_term())
+        elif self._match(TokenType.COMMA):
+            if self._match(TokenType.R_PAREN):
```
tobymao, line 1935: this is unnecessary let’s remove this check

GeorgeSittas: If removed, it recognizes `substring('bla',)` as valid sql. Is this ok?

sqlglot/parser.py
```diff
@@ -1917,6 +1920,28 @@ def _parse_convert(self):
...
+                self.raise_error("Unexpected ) token")
+
+            args.extend(self._parse_csv(self._parse_term))
```
tobymao, line 1938: should parse csv come first and then you don’t need the explicit parse term 

so parse csv, then check the from

sqlglot/parser.py
tobymao: yes

tobymao (approved):

PR #432, opened 2022-09-14
Add trim function

### LINKED ISSUE #425: [Postgres 14.4] Parser does not recognize 'both' keyword in trim() function
Example (valid sql):
```
select trim(both from ' xxx ');
```
Result:
```
ParseError: Expecting ). Line 1, Col: 18.
  select trim(both from ' xxx ');

Expected table name. Line 1, Col: 25.
  select trim(both from ' xxx ');

Required keyword: 'this' missing for <class 'sqlglot.expressions.Table'>. Line 1, Col: 25.
  select trim(both from ' xxx ');

... and 1 more
```

GeorgeSittas:
Implemented the trim function based (mostly) on [this](https://www.w3resource.com/sql/character-functions/trim.php).

- As I understand, Hive supports the syntax `[L|R]TRIM(target)`. If that's true, this implementation doesn't cover Hive.
- Do we generally want to provide links in comments (eg. for sources), like I did in `_parse_trim`?

fixes #425

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -102,6 +102,7 @@ class Parser:
         TokenType.VAR,
         TokenType.ALTER,
         TokenType.BEGIN,
+        TokenType.BOTH,
         TokenType.BUCKET,
         TokenType.CACHE,
         TokenType.COLLATE,
@@ -124,6 +125,7 @@ class Parser:
         TokenType.ISNULL,
         TokenType.INTERVAL,
         TokenType.LAZY,
+        TokenType.LEADING,
         TokenType.LOCATION,
         TokenType.NEXT,
         TokenType.ONLY,
@@ -143,6 +145,7 @@ class Parser:
         TokenType.TABLE_FORMAT,
         TokenType.TEMPORARY,
         TokenType.TOP,
+        TokenType.TRAILING,
         TokenType.TRUNCATE,
         TokenType.TRUE,
         TokenType.UNBOUNDED,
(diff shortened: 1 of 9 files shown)


GeorgeSittas:
> what do you mean hive isn’t supported?

I mean that we can still parse hive but transpilation _to_ hive for trim is not fully correct. The standard `TRIM(target)` will be transpiled correctly, but `TRIM(TRAILING FROM target)` transpiles to itself, which is invalid hive. We expect `RTRIM(target)`.

> we need to make sure hive works well because it’s a very important target

Yeah, makes sense. I'll try to fix the issue in another commit.

mpf82:
LTRIM and RTRIM are widely supported, whereas the ``TRIM( [{LEADING | TRAILING | BOTH}] ...)`` notation seems to be additional contructs supported by a minority of dialects (mysql and postgres, it seems?)

So may I suggest to add full support for TRIM, RTRIM and LTRIM and transpile TRIM with  [{LEADING | TRAILING | BOTH}] to RTRIM/LTRIM respectively? 

e.g. ``TRIM(TRAILING FROM target)`` will transpile to ``RTRIM(target)`` by default in all dialects, even postgres and mysql.

GeorgeSittas:
I did as Mike suggested, with one small difference: when reading postgres / mysql, it generates the dialect-specific trim sql only if it contains the "remove charset" or "collation" parts (which are valid in said dialects).

sqlglot/parser.py
```diff
@@ -1937,6 +1942,34 @@ def _parse_substring(self):
...
         return this
 
+    def _parse_trim(self):
+        # https://www.w3resource.com/sql/character-functions/trim.php
+        # https://docs.oracle.com/javadb/10.8.3.0/ref/rreftrimfunc.html
+
+        position = None
+        collation = None
+
+        if self._match_set((TokenType.LEADING, TokenType.TRAILING, TokenType.BOTH)):
```
tobymao, line 1952: make this a constant

tobymao (approved):

PR #433, opened 2022-09-14
Optimizer: Merge derived tables

barakalon:
New optimizer rule: merge derived tables

References:
https://dev.mysql.com/doc/refman/8.0/en/derived-table-optimization.html
https://mariadb.com/kb/en/derived-table-merge-optimization/

@tobymao this fights with the "isolate_table_selects" rule a bit. Happy to chat more.

cc @ginter

### DIFF
--- a/sqlglot/optimizer/merge_derived_tables.py
+++ b/sqlglot/optimizer/merge_derived_tables.py
@@ -36,7 +36,7 @@ def merge_derived_tables(expression):
 
                 _rename_inner_sources(outer_scope, inner_scope, alias)
                 _merge_from(outer_scope, inner_scope, subquery)
-                _merge_joins(outer_scope, inner_scope)
+                _merge_joins(outer_scope, inner_scope, from_or_join)
                 _merge_expressions(outer_scope, inner_scope, alias)
                 _merge_where(outer_scope, inner_scope, from_or_join)
                 _merge_order(outer_scope, inner_scope)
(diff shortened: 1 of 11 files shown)


sqlglot/optimizer/merge_derived_tables.py
```diff
@@ -0,0 +1,246 @@
...
+                _merge_from(outer_scope, inner_scope, subquery)
+                _merge_joins(outer_scope, inner_scope, from_or_join)
+                _merge_expressions(outer_scope, inner_scope, alias)
+                _merge_where(outer_scope, inner_scope, from_or_join)
+                _merge_order(outer_scope, inner_scope)
+    return expression
+
+
+# If a derived table has these Select args, it can't be merged
+UNMERGABLE_ARGS = set(exp.Select.arg_types.keys()) - {
```
tobymao, line 47: keys is unnecessary, can just do set(arg_types)

sqlglot/optimizer/optimizer.py
```diff
@@ -25,19 +41,18 @@ def optimize(expression, schema=None, db=None, catalog=None):
...
-    expression = unnest_subqueries(expression)
-    expression = expand_multi_table_selects(expression)
-    expression = pushdown_predicates(expression)
-    expression = optimize_joins(expression)
-    expression = eliminate_subqueries(expression)
-    expression = quote_identities(expression)
+    for rule in rules:
+        rule_kwargs = {
+            param: possible_kwargs[param]
+            for param in rule.__code__.co_varnames
```
tobymao, line 54: what the heck is this

tests/fixtures/optimizer/tpc-h/tpc-h.sql
```diff
@@ -120,68 +120,34 @@ SELECT
...
-    "part"."p_partkey" AS "p_partkey",
-    "part"."p_mfgr" AS "p_mfgr",
-    "part"."p_type" AS "p_type",
-    "part"."p_size" AS "p_size"
-  FROM "part" AS "part"
-  WHERE
-    "part"."p_size" = 15
-    AND "part"."p_type" LIKE '%BRASS'
-) AS "part"
+FROM "part" AS "part"
```
tobymao, line 123: none of these sql queries should change, we should isolate all table selects in tpc-h

sqlglot/optimizer/optimizer.py
```diff
@@ -10,8 +11,23 @@
...
+    isolate_table_selects,
+    qualify_columns,
+    pushdown_projections,
+    normalize,
+    unnest_subqueries,
+    expand_multi_table_selects,
+    pushdown_predicates,
+    optimize_joins,
+    eliminate_subqueries,
+    merge_derived_tables,
```
tobymao, line 25: perhaps merging should happen before isolate table selects, so that merging remove reduntant

select * from (select *), but then isolate still works

sqlglot/optimizer/merge_derived_tables.py
barakalon: `.keys()` is in my muscle memory. I think it makes the intent more explicit.
I can remove though.

tests/fixtures/optimizer/tpc-h/tpc-h.sql
barakalon: Why?

sqlglot/optimizer/optimizer.py
barakalon: merge_derived_tables depends on qualify_columns.

Also, in my use case, we don't want isolated table selects. I actually only left it in here so `eliminate_subqueries` can do its thing.

side note: should we be making the rule dependencies explicit? We might be able to automatically order them, which would make it easier to pick and choose rules.

barakalon:
I think the question here is "who does the optimizer optimize for?"

We've talked before about "canonical SQL", but we're starting to see some problems. I.e., I think some rules are really meant to help the executor do its planning - they're not really more semantically optimized.

sqlglot/optimizer/merge_derived_tables.py
tobymao: it hurts performance, i've seen it happen in minerva 1.0

tests/fixtures/optimizer/tpc-h/tpc-h.sql
tobymao: i guess the question is what should be default, it looks like you're changing the default optimized sql to remove isolated selects.

sqlglot/optimizer/optimizer.py
tobymao: maybe later on the rules dependencies, but no need to overcomplicate for now.

i think this comes down to what the default should be

tobymao:
if we're going to make the merge derived tables standard, then we don't even need to run isolated_table_selects. there should be two sets of rules

tobymao (approved):

sqlglot/optimizer/optimizer.py
barakalon: I added a little comment.
Unresolve if you're not happy with this magic.

PR #434, opened 2022-09-14
Support concatenation in like expressions

GeorgeSittas:
This aims to fix the following issue (mentioned in [this](https://github.com/tobymao/sqlglot/discussions/431) discussions post):

Example (valid sql):

```
SELECT * FROM foo WHERE some.column LIKE 'foo' || another.column || 'bar'
```

Result:

```
Invalid expression / Unexpected token. Line 1, Col: 48.
   SELECT * FROM foo WHERE some.column LIKE 'foo' || another.column || 'bar'
                                                  ~~
```


tobymao (approved):

yuchaofan13:
Amazing, thank you very much! I saw in the commit that this was solved by changing the parse_escape to parse_bitwise.    Would you mind briefly explaining how that works?

PR #437, opened 2022-09-15
Support lateral with subqueries

### LINKED ISSUE #424: [Postgres 14.4] Parser does not support subquery after lateral keyword
Example (valid sql):
```
select art from table1 inner join lateral (select art from table2) table2 on table2.art=table2.art
```
Result:
```
ParseError: Expected table name. Line 1, Col: 35.
  select art from table1 inner join lateral (select art from table2) table2 on table2.art=table2.art;

Required keyword: 'this' missing for <class 'sqlglot.expressions.Table'>. Line 1, Col: 35.
  select art from table1 inner join lateral (select art from table2) table2 on table2.art=table2.art;

Expected VIEW after LATERAL. Line 1, Col: 43.
  select art from table1 inner join lateral (select art from table2) table2 on table2.art=table2.art;

... and 2 more
```

GeorgeSittas:
Fixes #424

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1109,12 +1109,7 @@ class Parser:
         if not self._match(TokenType.JOIN):
             return None
 
-        this = (
-            self._parse_lateral()
-            if self._curr and self._curr.token_type == TokenType.LATERAL
-            else self._parse_table()
-        )
-        kwargs = {"this": this}
+        kwargs = {"this": self._parse_lateral() or self._parse_table()}
 
         if side:
             kwargs["side"] = side.text
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1070,8 +1070,12 @@ class Parser:
         if not self._match(TokenType.LATERAL):
             return None
 
-        if not self._match(TokenType.VIEW):
-            self.raise_error("Expected VIEW after LATERAL")
+        subquery = self._parse_select(table=True)
+
+        if subquery:
+            return self.expression(exp.Lateral, this=subquery)
+        elif not self._match(TokenType.VIEW):
+            self.raise_error("Expected subquery or VIEW after LATERAL")
 
         outer = self._match(TokenType.OUTER)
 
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -1070,8 +1070,12 @@ def _parse_lateral(self):
         if not self._match(TokenType.LATERAL):
             return None
 
-        if not self._match(TokenType.VIEW):
-            self.raise_error("Expected VIEW after LATERAL")
+        subquery = self._parse_select(table=True)
```
tobymao, line 1073: should it be parse_with ?

sqlglot/parser.py
```diff
@@ -1105,7 +1109,12 @@ def _parse_join(self):
         if not self._match(TokenType.JOIN):
             return None
 
-        kwargs = {"this": self._parse_table()}
+        this = (
+            self._parse_lateral()
+            if self._curr and self._curr.token_type == TokenType.LATERAL
+            else self._parse_table()
+        )
+        kwargs = {"this": this}
```
tobymao, line 1117: this could just be {"this": self._parse_lateral() or self._parse_table()} right?

GeorgeSittas: Yeah good point

sqlglot/parser.py
GeorgeSittas: What do you mean?

tobymao (approved):

ypogorelova:
Hi, I was also working on this issue. I noticed that there are a couple more cases for using LATERAL in postgres:
the first one:
```
SELECT m.name
FROM manufacturers m LEFT JOIN LATERAL get_product_names(m.id) pname ON true
WHERE pname IS NULL;
```
fails with an error:
`sqlglot.errors.ParseError: Expected subquery or VIEW after LATERAL. Line 2, Col: 40.`

and the second one:
```
SELECT p1.id, p2.id, v1, v2
FROM polygons p1, polygons p2,
     LATERAL vertices(p1.poly) v1,
     LATERAL vertices(p2.poly) v2
WHERE (v1 <-> v2) < 10 AND p1.id != p2.id;
```
fails with an error:
`sqlglot.errors.ParseError: Expected table name. Line 3, Col: 6.`

are those issues supposed to be fixed in another PR?

GeorgeSittas:
> are those issues supposed to be fixed in another PR?

[Fixed](https://github.com/tobymao/sqlglot/commit/185adcb957636ddd1d55eaf999539abe1aa7fc4b).

ypogorelova:
> 

Thank you

PR #439, opened 2022-09-15
fix oracle table alias

mpf82:
oracle does not allow "AS" between table and alias

https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_10002.htm#i2126863


barakalon (approved):

barakalon:
[v6.0.7](https://pypi.org/project/sqlglot/)

PR #440, opened 2022-09-16
Null treatment

prabhur:
The SQL spec defines an optional `null_treatment` clause in some of the window functions. 
Some dialects like Oracle and Snowflake choose to implement this (https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/FIRST_VALUE.html#GUID-D454EC3F-370C-4C64-9B11-33FCB10D95EC)

Additionally, this clause can be present inside the brackets like
`SELECT FIRST_VALUE(TABLE.COLUMN IGNORE NULLS) ...` 
or outside
`SELECT FIRST_VALUE(TABLE.COLUMN) IGNORE NULLS ...` 

The former seems to be already handled by sqlglot, but the latter not. This PR should address the latter.


sqlglot/dialects/sqlite.py
```diff
@@ -29,6 +29,8 @@ class Parser(Parser):
         }
 
     class Generator(Generator):
+        NULL_TREATMENT_IMPLEMENTED = False
```
tobymao, line 32: what is this? this doesn't look used

prabhur: My bad. I should have removed it before submitting. It is a slip up from the previous revision.

tobymao (approved):

PR #444, opened 2022-09-16
Add support for parenthesized joined tables

### LINKED ISSUE #423: [Postgres 14.4] Parser does not understand inline join in from clause
Example (valid sql):

```
select * from (table1 left join table2 on 1 = 1);
```
Result:
```
ParseError: Expecting ). Line 1, Col: 23.
  select * from (table1 left join table2 on 1 = 1);

Invalid expression / Unexpected token. Line 1, Col: 48.
  select * from (table1 left join table2 on 1 = 1);
```

GeorgeSittas:
This PR aims to add support for the "joined tables" or "join construct" form that uses parentheses. Example:

```
SELECT * FROM (table1 LEFT JOIN table2 ON 1 = 1)
```

Sources:
- https://www.postgresql.org/docs/current/queries-table-expressions.html
- https://dev.mysql.com/doc/refman/8.0/en/join.html

fixes #423

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -885,10 +885,6 @@ class Reference(Expression):
     arg_types = {"this": True, "expressions": True}
 
 
-class Table(Expression):
-    arg_types = {"this": True, "db": False, "catalog": False}
-
-
 class Tuple(Expression):
     arg_types = {"expressions": False}
 
@@ -996,6 +992,10 @@ QUERY_MODIFIERS = {
     "limit": False,
     "offset": False,
 }
+
+
+class Table(Expression):
+    arg_types = {"this": True, "db": False, "catalog": False, **QUERY_MODIFIERS}
 
 
 class Union(Subqueryable, Expression):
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -998,6 +994,10 @@ def with_(
 }
 
 
+class Table(Expression):
+    arg_types = {"this": True, "db": False, "catalog": False, **QUERY_MODIFIERS}
```
barakalon, line 998: Should this only be "joins"?

GeorgeSittas: That's a good point, I think it's safe to handle only joins here but let me double check.

GeorgeSittas: If I'm not mistaken, by following MySQL's syntax (2nd link, starting at the table_references rule) you can derive:

```
SELECT * FROM (LATERAL (SELECT * FROM tbl) as t)
```

On the other hand, I tried two different syntax checkers and this doesn't seem to be valid MySQL. Am I missing something?

barakalon: Yeah that doesn't look like a valid query.

I just meant should we only be allowing the "joins" arg, not _all_ the QUERY_MODIFIERS

GeorgeSittas: Yeah that seems correct. I think it's sufficient to use just `_parse_joins` instead of all the query modifier parsers too.

 I'll make a PR to fix this.

PR #447, opened 2022-09-16
Use joins instead of all query modifiers in table

GeorgeSittas:
See [discussion](https://github.com/tobymao/sqlglot/pull/444#discussion_r973199672).

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -541,9 +541,7 @@ class Generator:
             if part
         )
 
-        joins = csv(
-            *[self.sql(sql) for sql in expression.args.get("joins", [])], sep=""
-        )
+        joins = self.expressions(expression, key="joins", sep="")
         return f"{table}{joins}"
 
     def tablesample_sql(self, expression):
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -1032,7 +1032,9 @@ def _parse_subquery(self, this):
         return self.expression(exp.Subquery, this=this, alias=self._parse_table_alias())
 
     def _parse_query_modifiers(self, this):
-        if not isinstance(this, (exp.Subquery, exp.Subqueryable, exp.Table)):
+        if not isinstance(this, (exp.Subquery, exp.Subqueryable)):
+            if isinstance(this, exp.Table):
```
tobymao, line 1036: put this outside the nesting

sqlglot/generator.py
```diff
@@ -541,7 +541,10 @@ def table_sql(self, expression):
             if part
         )
 
-        return f"{table}{self.query_modifiers(expression)}"
+        joins = csv(
```
tobymao, line 544: can you do self.expressions arg joins

tobymao (approved):

PR #450, opened 2022-09-17
Support functions after lateral & add distance op for postgres

GeorgeSittas:
The following form is allowed by postgres but is not currently supported. This PR aims to fix that.

```
SELECT p1.id, p2.id, v1, v2
FROM polygons p1, polygons p2,
     LATERAL vertices(p1.poly) v1,
     LATERAL vertices(p2.poly) v2
WHERE (v1 <-> v2) < 10 AND p1.id != p2.id;
```

Source (7.2.1.5 LATERAL Subqueries): https://www.postgresql.org/docs/current/queries-table-expressions.html

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -32,6 +32,22 @@ def _date_add_sql(kind):
     return func
 
 
+def _distance_sql(self, expression):
+    return self.binary(expression, "<->")
+
+
+def _lateral_sql(self, expression):
+    this = self.sql(expression, "this")
+    if isinstance(expression.this, exp.Subquery):
+        return f"LATERAL{self.sep()}{this}"
+    alias = expression.args["alias"]
+    table = alias.name
+    table = f" {table}" if table else table
+    columns = self.expressions(alias, key="columns", flat=True)
+    columns = f" AS {columns}" if columns else ""
+    return f"LATERAL{self.sep()}{this}{table}{columns}"
+
+
 def _substring_sql(self, expression):
     this = self.sql(expression, "this")
     start = self.sql(expression, "start")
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -32,6 +32,22 @@ def func(self, expression):
     return func
 
 
+def _distance_sql(self, expression):
```
tobymao, line 35: i don't think these need to be postgres specific because <-> doesn't clash

sqlglot/dialects/postgres.py
```diff
@@ -32,6 +32,22 @@ def func(self, expression):
...
+    return self.binary(expression, "<->")
+
+
+def _lateral_sql(self, expression):
+    this = self.sql(expression, "this")
+    if isinstance(expression.this, exp.Subquery):
+        return f"LATERAL{self.sep()}{this}"
+    alias = expression.args["alias"]
+    table = alias.name
```
tobymao, line 44: can you do expression.text("alias") ?

GeorgeSittas: Doing this means that alias would become a string and then `alias.name` etc wouldn't work, right?

tobymao: yea but why do you need alias.name j
if you have it from the text?

GeorgeSittas: Besides that there's also the self.expressions call which uses the alias expression object too.

tobymao: yea just do that

GeorgeSittas: Also, is expression.text("alias") correct? I think this always results in an empty string here.

tobymao: not sure. do what works i guess

PR #451, opened 2022-09-19
Add support for bit strings and hex strings

crericha:
Work in progress. Submitting as a PR for feedback. Comments/questions will be inline.


sqlglot/generator.py
```diff
@@ -142,6 +150,10 @@ def __init__(
         self.quote_end = quote_end or "'"
         self.identifier_start = identifier_start or '"'
         self.identifier_end = identifier_end or '"'
+        self.bit_string_start = bit_string_start
+        self.bit_string_end = bit_string_end
+        self.hex_string_start = hex_string_start
+        self.hex_string_end = hex_string_end
```
crericha, line 156: The delimiter information is getting long. I wonder if it is cleaner to have default behavior here and overridden behavior as a lambda in dialect. For example, most languages (at least in their documentation) don't support bit and hex strings, so the default behavior would be to turn it into decimal; however, for the dialects that do support it, the dialect class will add a lambda that will be called instead, which maintains the hex base and adds the delimiters.

sqlglot/tokens.py
```diff
@@ -626,13 +638,15 @@ def tokenize(self, sql):
...
+            identifier_end = self._IDENTIFIERS.get(self._char)
 
             if white_space:
                 if white_space == TokenType.BREAK:
                     self._col = 1
                     self._line += 1
-            elif self._char == "0" and self._peek == "x":
+            elif self._char == "0" and self._peek.upper() == "B":
+                self._scan_bits()
+            elif self._char == "0" and self._peek.upper() == "X":
```
crericha, line 649: Dialects that support 0x and 0X nomenclature need to be processed here, so the leading digit doesn't confuse things. Later, formats like X'FF' are processed via the trie. This is a bit annoying because there are two hex processing methods now. Do you think that's okay?

tests/dialects/test_mysql.py
```diff
@@ -41,12 +41,6 @@ def test_binary_literal(self):
                 "mysql": "SELECT `0xz`",
             },
         )
-        self.validate_all(
-            "SELECT 0XCC",
-            write={
-                "mysql": "SELECT 0 AS XCC",
-            },
-        )
```
crericha, line 49: In the previous code "SELECT 0xGM" would treat 0xGM as a literal and not like "SELECT 0 as xGM". Should scan_hex be smarter like scan_number to handle these cases?

sqlglot/tokens.py
crericha: Should this logic be checking that 0x or 0X is supported by the dialect before trying to process as a hex number? There are a lot of hardcoded symbols in the scanning logic and I wasn't sure if I should follow that norm or try to make the code smarter (but more complex).

sqlglot/generator.py
```diff
@@ -384,7 +396,16 @@ def tablealias_sql(self, expression):
         return f"{alias}{columns}"
 
     def bitstring_sql(self, expression):
-        return f"b'{self.sql(expression, 'this')}'"
+        if not self.bit_string_start:
+            return f"{self.sql(expression, 'this')}"
```
crericha, line 400: Converting to decimal seems like the right thing to do for dialects that don't support a special format. Do you agree?

sqlglot/generator.py
tobymao: sure, we can do that

sqlglot/generator.py
tobymao: why is it a decimal?

tests/dialects/test_mysql.py
tobymao: why did this test change?

sqlglot/tokens.py
tobymao: can we avoid doing self._char == 0 and self._peek.upper twice? could we maybe put it inside the self._char.isdigit() block?

so like

elif self.char.isdigit():
   if self._char == "0":
      peek = self._peek.upper()
...

sqlglot/tokens.py
```diff
@@ -833,6 +863,42 @@ def _scan_string(self, quote):
...
         return True
 
+    def _scan_bit_string(self, bit_string):
+        bs_end = self._BIT_STRINGS.get(bit_string)
+        if bs_end is None:
+            return False
+
+        self._advance(len(bit_string))
+        text = self._extract_string(bs_end)
+        if bs_end != " ":
```
tobymao, line 873: when does bs_end == " " ?

alippai:
Can we separate bit strings from binary values (number) and hex strings from hex numbers resp? There are dialects (eg. Vertica) where formatting the query results in type change now (from hex number to bit string)

sqlglot/tokens.py
```diff
@@ -833,6 +863,42 @@ def _scan_string(self, quote):
...
+
+        try:
+            self._add(TokenType.BIT_STRING, f"{int(text, 2)}")
+        except ValueError:
+            raise RuntimeError(
+                f"Bit string contains invalid characters from {self._line}:{self._start}"
+            )
+        return True
+
+    def _scan_hex_string(self, hex_string):
```
tobymao, line 884: this looks the same as bit string, should we turn this into one method with args?

sqlglot/tokens.py
```diff
@@ -851,3 +917,24 @@ def _scan_var(self):
             else:
                 break
         self._add(self.KEYWORDS.get(self._text.upper(), TokenType.VAR))
+
+    def _extract_string(self, delimiter):
```
tobymao, line 921: this looks similar to _scan_string, again any opp to dry this up?

sqlglot/generator.py
crericha: Because the dialect doesn't support hexademcial notation

sqlglot/tokens.py
crericha: No longer needed. Will remove.

crericha:
> Can we separate bit strings from binary values (number) and hex strings from hex numbers resp? There are dialects (eg. Vertica) where formatting the query results in type change now (from hex number to bit string)

To verify my understanding, are you saying 0xFF (hex number) and X'FF' (hex string) should be handled differently?

sqlglot/tokens.py
crericha: Yes, there is probably some opportunity to use a shared helper method.

sqlglot/generator.py
tobymao: no i understand that but why decimal? i don't understand how bit strings convert to decimals?

alippai:
> > Can we separate bit strings from binary values (number) and hex strings from hex numbers resp? There are dialects (eg. Vertica) where formatting the query results in type change now (from hex number to bit string)
> 
> To verify my understanding, are you saying 0xFF (hex number) and X'FF' (hex string) should be handled differently?

Yes, exactly. String printed as string, number as number, but not doing an opinionated conversion on parse level (as far as I understand the code, later in the generator I won't know if it was a string or number)

tests/dialects/test_mysql.py
crericha: Discussed in slack. Now that 0XCG converts to a literal, the test is redundant now with the preceding test.

sqlglot/tokens.py
crericha: Fixed

sqlglot/dialects/dialect.py
```diff
@@ -53,12 +53,20 @@ def __new__(cls, clsname, bases, attrs):
...
-            0
-        ]
+        klass.quote_start, klass.quote_end = list(
+            klass.tokenizer_class._QUOTES.items()
+        )[0]
         klass.identifier_start, klass.identifier_end = list(
-            klass.tokenizer_class.IDENTIFIERS.items()
+            klass.tokenizer_class._IDENTIFIERS.items()
         )[0]
+        if len(klass.tokenizer_class._BIT_STRINGS) > 0:
```
tobymao, line 62: you can just do if bit_strings nince len(x) > 0 is truthy

crericha: Fixed

sqlglot/tokens.py
crericha: Combined

sqlglot/tokens.py
crericha: refactored

sqlglot/tokens.py
crericha: refactored

sqlglot/generator.py
crericha: Discussed in slack. Decimal only for the dialects that don't support hexadecimal or bit notation.

crericha:
> > > Can we separate bit strings from binary values (number) and hex strings from hex numbers resp? There are dialects (eg. Vertica) where formatting the query results in type change now (from hex number to bit string)
> > 
> > 
> > To verify my understanding, are you saying 0xFF (hex number) and X'FF' (hex string) should be handled differently?
> 
> Yes, exactly. String printed as string, number as number, but not doing an opinionated conversion on parse level (as far as I understand the code, later in the generator I won't know if it was a string or number)
> 
> Edit: I know some dialects won't differentiate between the two and using only one type of output is actually useful

I spoke with Toby and the summary is that since Vertica isn't a supported dialect (at least currently), I'm not going add complexity to distinguish between string and value. If it is needed in the future, it will be pretty straight-forward to add since the tokenizer can already distinguish.

sqlglot/generator.py
```diff
@@ -389,7 +389,10 @@ def tablealias_sql(self, expression):
         return f"{alias}{columns}"
 
     def bitstring_sql(self, expression):
-        return f"b'{self.sql(expression, 'this')}'"
+        return f"{self.sql(expression, 'this')}"
```
tobymao, line 392: so the default just returns a normal string?

sqlglot/generator.py
```diff
@@ -389,7 +389,10 @@ def tablealias_sql(self, expression):
...
+
+    def hexstring_sql(self, expression):
+        return f"{self.sql(expression, 'this')}"
```
tobymao, line 395: return self.sql(expression, "this")

sqlglot/tokens.py
```diff
@@ -295,14 +288,29 @@ def __new__(cls, clsname, bases, attrs):
             key.upper()
             for key, value in {
                 **klass.KEYWORDS,
-                **{comment: TokenType.COMMENT for comment in klass.COMMENTS},
-                **{quote: TokenType.QUOTE for quote in klass.QUOTES},
+                **{comment: TokenType.COMMENT for comment in klass._COMMENTS},
+                **{quote: TokenType.QUOTE for quote in klass._QUOTES},
+                **{
```
tobymao, line 293: are the bit string and hex string ever needed in ambiguous key words?

tobymao: i see they are now that you leverage them in keywords

sqlglot/generator.py
crericha: Stored as a decimal string, so yeah.

sqlglot/generator.py
crericha: Fixed

tobymao (approved):

PR #452, opened 2022-09-19
Add support for `SELECT` leading commas in SQL generation

akravetz:
Some folks have a preference for leading commas, ex:
```sql
SELECT
    foo
  , bar
  , baz
```
instead of the current behavior, which is trailing commas:

```sql
SELECT 
  foo,
  bar,
  baz
```

this creates an option for leading commas

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1137,10 +1137,10 @@ class Generator:
 
         sql = [self.sql(e) for e in expressions]
         # the only time leading_comma changes the output is if pretty print is enabled
-        # and there is more than 1 expression
-        if self._leading_comma and self.pretty and len(sql) > 1:
-            remaining_expr = "\n".join([f"{sep}{s}" for s in sql[1:]])
-            expressions = f"  {sql[0]}\n{remaining_expr}"
+        if self._leading_comma and self.pretty:
+            expressions = "\n".join(
+                f"{sep}{s}" if i > 0 else s for i, s in enumerate(sql)
+            )
         else:
             expressions = self.sep(sep).join(sql)
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1139,7 +1139,8 @@ class Generator:
         # the only time leading_comma changes the output is if pretty print is enabled
         if self._leading_comma and self.pretty:
             expressions = "\n".join(
-                f"{sep}{s}" if i > 0 else s for i, s in enumerate(sql)
+                f"{sep}{s}" if i > 0 else f"{' ' * self.pad}{s}"
+                for i, s in enumerate(sql)
             )
         else:
             expressions = self.sep(sep).join(sql)
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -41,6 +41,8 @@ class Generator:
         max_unsupported (int): Maximum number of unsupported messages to include in a raised UnsupportedError.
             This is only relevant if unsupported_level is ErrorLevel.RAISE.
             Default: 3
+        leading_comma (bool): if the the comma is leading or trailing in select statements
+            Default: False
     """
 
     TRANSFORMS = {
@@ -131,6 +133,7 @@ class Generator:
         unsupported_level=ErrorLevel.WARN,
         null_ordering=None,
         max_unsupported=3,
+        leading_comma=False,
     ):
         import sqlglot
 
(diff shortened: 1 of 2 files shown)


sqlglot/generator.py
```diff
@@ -108,6 +110,7 @@ class Generator:
         "_indent",
         "_replace_backslash",
         "_escaped_quote_end",
+        "leading_comma",
```
tobymao, line 113: make it private? _leading_comma

sqlglot/generator.py
```diff
@@ -1130,7 +1135,13 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
         if flat:
             return sep.join(self.sql(e) for e in expressions)
 
-        expressions = self.sep(sep).join(self.sql(e) for e in expressions)
+        sql = [self.sql(e) for e in expressions]
+        if len(sql) > 1 and self.leading_comma:
+            remaining_expr = "\n".join([f"{sep}{s}" for s in sql[1:]])
+            expressions = f"  {sql[0]}\n{remaining_expr}"
```
tobymao, line 1141: this doesn't seem to handle pretty vs non pretty? it always has a line break

sqlglot/generator.py
```diff
@@ -1130,7 +1135,13 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
         if flat:
             return sep.join(self.sql(e) for e in expressions)
 
-        expressions = self.sep(sep).join(self.sql(e) for e in expressions)
+        sql = [self.sql(e) for e in expressions]
+        if len(sql) > 1 and self.leading_comma:
```
tobymao, line 1139: self._leading_comma should go first so we don't always do the length check

sqlglot/generator.py
akravetz: ah, good call, I'l add that in

sqlglot/generator.py
akravetz: done

tobymao:
can you add a test case? in test_transpile.py

sqlglot/generator.py
```diff
@@ -1130,7 +1135,15 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
         if flat:
             return sep.join(self.sql(e) for e in expressions)
 
-        expressions = self.sep(sep).join(self.sql(e) for e in expressions)
+        sql = [self.sql(e) for e in expressions]
+        # the only time leading_comma changes the output is if pretty print is enabled
+        # and there is more than 1 expression
+        if self._leading_comma and self.pretty and len(sql) > 1:
+            remaining_expr = "\n".join([f"{sep}{s}" for s in sql[1:]])
+            expressions = f"  {sql[0]}\n{remaining_expr}"
```
tobymao, line 1143: you shouldn't hardcode 2 spaces here

sqlglot/generator.py
```diff
@@ -1130,7 +1135,15 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
         if flat:
             return sep.join(self.sql(e) for e in expressions)
 
-        expressions = self.sep(sep).join(self.sql(e) for e in expressions)
+        sql = [self.sql(e) for e in expressions]
+        # the only time leading_comma changes the output is if pretty print is enabled
+        # and there is more than 1 expression
+        if self._leading_comma and self.pretty and len(sql) > 1:
+            remaining_expr = "\n".join([f"{sep}{s}" for s in sql[1:]])
```
tobymao, line 1142: the square pracket [ are unnecessary

sqlglot/generator.py
```diff
@@ -1130,7 +1135,15 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
         if flat:
             return sep.join(self.sql(e) for e in expressions)
 
-        expressions = self.sep(sep).join(self.sql(e) for e in expressions)
+        sql = [self.sql(e) for e in expressions]
+        # the only time leading_comma changes the output is if pretty print is enabled
+        # and there is more than 1 expression
+        if self._leading_comma and self.pretty and len(sql) > 1:
```
tobymao, line 1141: is this len really necessary?

can you just do

"\n".join(f"{sep}s} if i > 0 else s for i, s in enumerate(sql))

akravetz: The `len()` is a O(1) operation, so I kept it in for the sake of clarity.  I personally find complex list comprehensions like the above hard read/reason about.  That being said, happy to make the change if you'd prefer that/it matches the aesthetic of the code base

tobymao (approved):

tobymao:
thanks for the pr and quick updates

PR #456, opened 2022-09-20
Add redshift dialect (WIP)

GeorgeSittas:
This is an attempt to add support for the redshift dialect. Feedback is welcome. Plan to add more features soon.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -6,30 +6,9 @@ from sqlglot.tokens import TokenType
 class Redshift(Postgres):
     time_format = "'YYYY-MM-DD HH:MI:SS'"
     time_mapping = {
-        "YY": "%y",
-        "yy": "%y",
-        "YYYY": "%Y",
-        "yyyy": "%Y",
-        "MM": "%m",
-        "mm": "%m",
+        **Postgres.time_mapping,
         "MON": "%b",
-        "mon": "%b",
-        "DD": "%d",
-        "dd": "%d",
-        "HH12": "%I",
-        "hh12": "%I",
-        "HH24": "%H",
-        "hh24": "%H",
         "HH": "%H",
-        "hh": "%H",
-        "MI": "%M",
-        "mi": "%M",
-        "SS": "%S",
-        "ss": "%S",
-        "OF": "%z",
-        "of": "%z",
-        "AM": "%p",
-        "PM": "%p",
     }
 
     class Tokenizer(Postgres.Tokenizer):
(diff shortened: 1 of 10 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -0,0 +1,58 @@
...
+            "TIMETZ": TokenType.TIMESTAMPTZ,
+            "VARBYTE": TokenType.VARBYTE,
+            "SIMILAR TO": TokenType.SIMILAR_TO,
+        }
+
+    class Generator(Postgres.Generator):
+        TYPE_MAPPING = {
+            **Postgres.Generator.TYPE_MAPPING,
+            exp.DataType.Type.BINARY: "VARBYTE",
+            exp.DataType.Type.GEOMETRY: "GEOMETRY",
```
tobymao, line 54: do geometry and hll sketch, or super needs these mappings? doesn't seem to do anything

sqlglot/dialects/redshift.py
```diff
@@ -0,0 +1,58 @@
...
+            "TIME": TokenType.TIMESTAMP,
+            "TIMETZ": TokenType.TIMESTAMPTZ,
+            "VARBYTE": TokenType.VARBYTE,
+            "SIMILAR TO": TokenType.SIMILAR_TO,
+        }
+
+    class Generator(Postgres.Generator):
+        TYPE_MAPPING = {
+            **Postgres.Generator.TYPE_MAPPING,
+            exp.DataType.Type.BINARY: "VARBYTE",
```
tobymao, line 53: why do you need a varbyte type? why not just in keywords varbyte -> Binary?

sqlglot/parser.py
```diff
@@ -1547,7 +1554,10 @@ def _parse_types(self):
                 self.raise_error("Expecting >")
 
         if type_token in self.TIMESTAMPS:
-            tz = self._match(TokenType.WITH_TIME_ZONE)
+            tz = (
+                self._match(TokenType.WITH_TIME_ZONE)
+                or type_token == TokenType.TIMESTAMPTZ
```
tobymao, line 1559: was this an old bug?

GeorgeSittas: I think so yeah, timestamptz would be parsed as timestamp.

sqlglot/dialects/redshift.py
GeorgeSittas: Yeah I'll fix this. Redshift supports varbyte, varbinary and binary varying.

sqlglot/dialects/redshift.py
GeorgeSittas: I think there are some use cases. For example:
- https://hevodata.com/learn/redshift-super-data-type/  (`CREATE TABLE region_nations_noshred (rdata SUPER);`)
- https://www.youtube.com/watch?v=f4EpjeuOchE (`CREATE TABLE ... (... active_users HLLSKETCH, ...)`
- https://docs.aws.amazon.com/redshift/latest/dg/GeometryType-function.html (not explicitly used).

But I also didn't find many other references, so I could leave them out for the time being.

tobymao (approved):

PR #468, opened 2022-09-21
avoid mutating parent when rendering if expression

ginter:


sqlglot/generator.py
```diff
@@ -903,7 +903,9 @@ def unique_sql(self, expression):
         return f"UNIQUE ({columns})"
 
     def if_sql(self, expression):
-        return self.case_sql(exp.Case(ifs=[expression], default=expression.args.get("false")))
+        case = exp.Case(ifs=[], default=expression.args.get("false"))
```
tobymao, line 906: this is tricky, i see you’re doing jt because the init mutate

tobymao: i don’t know if we should do it like this. may be cleaner to do a copy? i need to think

barakalon (commented):
This is a really interesting bug.

I think we do this in other places too.

To recap from offline discussion: calling .sql() on an expression can implicitly change the parent-child relationships in the expression, since we set parents in the Expression constructor.

Is there a more elegant solution?

I.e...

```python
class Expression:
    def __init__(self, _set_parent=True, **args):
        ...
```

sqlglot/generator.py
ginter: copy is easier and also works just figured you didn't want the overhead

tobymao:
so a copy in generator.

barakalon:
Ah sure, that's probably better.

barakalon (approved):

sqlglot/generator.py
tobymao: i don't like overhead but, alas

PR #469, opened 2022-09-21
Add Transact-SQL dialect (WIP)

GeorgeSittas:
This is an attempt to add support for the transact sql dialect. Feedback is welcome. Plan to add more features soon.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -26,8 +26,8 @@ class Dialects(str, Enum):
     SQLITE = "sqlite"
     STARROCKS = "starrocks"
     TABLEAU = "tableau"
-    TRANSACTSQL = "transactsql"
     TRINO = "trino"
+    TSQL = "tsql"
 
 
 class _Dialect(type):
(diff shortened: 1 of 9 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -26,6 +26,7 @@ class Dialects(str, Enum):
     SQLITE = "sqlite"
     STARROCKS = "starrocks"
     TABLEAU = "tableau"
+    TRANSACTSQL = "transactsql"
```
tobymao, line 29: maybe let's just call it TSQL ?

sqlglot/tokens.py
```diff
@@ -123,6 +130,7 @@ class TokenType(AutoName):
     CURRENT_TIME = auto()
     CURRENT_TIMESTAMP = auto()
     DIV = auto()
+    DECLARE = auto()
```
tobymao, line 133: why did you add this it's not used?

GeorgeSittas: Yeah I'll remove it, I was trying out the declare syntax and forgot it. We can add it again later if needed.

sqlglot/dialects/dialect.py
GeorgeSittas: sure

tobymao (approved):

PR #471, opened 2022-09-22
Update Spark dialect to use USING instead of STORED AS in the CREATE statement

izeigerman:


tests/dialects/test_spark.py
```diff
@@ -44,15 +44,15 @@ def test_ddl(self):
             write={
                 "presto": "CREATE TABLE test WITH (FORMAT = 'PARQUET') AS SELECT 1",
                 "hive": "CREATE TABLE test STORED AS PARQUET AS SELECT 1",
-                "spark": "CREATE TABLE test STORED AS PARQUET AS SELECT 1",
+                "spark": "CREATE TABLE test USING PARQUET AS SELECT 1",
             },
         )
         self.validate_all(
             "CREATE TABLE test USING ICEBERG STORED AS PARQUET AS SELECT 1",
```
izeigerman, line 51: @tobymao , I think the syntax of this statement is wrong. Should we even have this test case?

izeigerman: More specifically based on Spark documentation and my empirical observation one can't use both `USING` and `STORED AS` as part of the same statement. The former is the part of the new Data Source syntax - https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-datasource.html while the latter is the part of the legacy Hive syntax https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-hiveformat.html

izeigerman: @robert8138 just a heads up that this comment is related to line 51, not 47. Line 47 looks good indeed. I'm going to go ahead and drop an invalid test case.

PR #475, opened 2022-09-23
Fix rename_func so it handles var len arg functions

### LINKED ISSUE #474: rename_func doesn't support variable length functions
`rename_func` throws an error when used on functions with variable length arguments.

Example:

```Python
class Test(Dialect):
    class Generator(Generator):
        TRANSFORMS = {
            **Generator.TRANSFORMS,
            exp.Array: rename_func("LIST_VALUE"),
        }
```

Error:

```sh
✗ python -m sqlglot --read test "SELECT ARRAY(1,2,3)"
  File "/usr/local/Cellar/python@3.8/3.8.13/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/Cellar/python@3.8/3.8.13/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "sqlglot/sqlglot/__main__.py", line 54, in <module>
    sqls = sqlglot.transpile(
  File "sqlglot/sqlglot/__init__.py", line 93, in transpile
    return [
  File "sqlglot/sqlglot/__init__.py", line 94, in <listcomp>
    Dialect.get_or_raise(write)().generate(expression, **opts)
  File "sqlglot/sqlglot/dialects/dialect.py", line 138, in generate
    return self.generator(**opts).generate(expression)
  File "sqlglot/sqlglot/generator.py", line 177, in generate
    sql = self.sql(expression).strip()
  File "sqlglot/sqlglot/generator.py", line 259, in sql
    return getattr(self, exp_handler_name)(expression)
  File "sqlglot/sqlglot/generator.py", line 726, in select_sql
    expressions = self.expressions(expression)
  File "sqlglot/sqlglot/generator.py", line 1117, in expressions
    expressions = self.sep(sep).join(sql)
  File "sqlglot/sqlglot/generator.py", line 1111, in <genexpr>
    sql = (self.sql(e) for e in expressions)
  File "sqlglot/sqlglot/generator.py", line 250, in sql
    return transform(self, expression)
  File "sqlglot/sqlglot/dialects/dialect.py", line 175, in <lambda>
    return lambda self, expression: f"{name}({csv(*[self.sql(e) for e in expression.args.values()])})"
  File "sqlglot/sqlglot/dialects/dialect.py", line 175, in <listcomp>
    return lambda self, expression: f"{name}({csv(*[self.sql(e) for e in expression.args.values()])})"
  File "sqlglot/sqlglot/generator.py", line 255, in sql
ValueError: Expected an Expression. Received <class 'list'>: [(LITERAL this: 1, is_string: False), (LITERAL this: 2, is_string: False), (LITERAL this: 3, is_string: False)]
```

The DuckDB dialect uses a workaround (see https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/duckdb.py#L111), but it should just work.

GeorgeSittas:
This can probably be written in a better way. I tried making the intent clearer by doing `if expression.is_var_len_args`, but apparently it doesn't work because the BitwiseLeft/RightShift tests fail:

```
ERROR: test_bits (tests.dialects.test_hive.TestHive) [x >> 1 -> spark]
...
    if expression.is_var_len_args
AttributeError: 'BitwiseRightShift' object has no attribute 'is_var_len_args'
```

Fixes #474

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -174,7 +174,7 @@ def rename_func(name):
     def _rename(self, expression):
         args = (
             self.expressions(expression, flat=True)
-            if expression.args.get("expressions")
+            if isinstance(expression, exp.Func) and expression.is_var_len_args
             else csv(*[self.sql(e) for e in expression.args.values()])
         )
         return f"{name}({args})"
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -171,7 +171,15 @@ class Dialect(metaclass=_Dialect):
 
 
 def rename_func(name):
-    return lambda self, expression: f"{name}({csv(*[self.sql(e) for e in expression.args.values()])})"
+    def _rename(self, expression):
+        args = (
+            self.expressions(expression, flat=True)
+            if expression.args.get("expressions")
+            else csv(*[self.sql(e) for e in expression.args.values()])
+        )
+        return f"{name}({args})"
+
+    return _rename
 
 
 def approx_count_distinct_sql(self, expression):
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -108,7 +108,7 @@ class DuckDB(Dialect):
         TRANSFORMS = {
             **Generator.TRANSFORMS,
             exp.ApproxDistinct: approx_count_distinct_sql,
-            exp.Array: lambda self, e: f"LIST_VALUE({self.expressions(e, flat=True)})",
+            exp.Array: rename_func("LIST_VALUE"),
             exp.ArraySize: rename_func("ARRAY_LENGTH"),
             exp.ArraySort: _array_sort_sql,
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -171,7 +171,15 @@ def generator(self, **opts):
 
 
 def rename_func(name):
-    return lambda self, expression: f"{name}({csv(*[self.sql(e) for e in expression.args.values()])})"
+    def _rename(self, expression):
+        args = (
+            self.expressions(expression, flat=True)
+            if expression.args.get("expressions")
```
tobymao, line 177: you can check that it’s a func before you call isvarlen arg

tobymao: this may be ok but may be better to do if expressions in args

PR #476, opened 2022-09-23
Add support for removing AST nodes using transform

GeorgeSittas:
Inspired by [this](https://github.com/tobymao/sqlglot/discussions/462) discussion.

Demo:

```python

expression_tree = parse_one("SELECT a, b FROM x")

def fun(node):
    if isinstance(node, exp.Column) and node.name == "b":
        return None
    return node

transformed_tree = expression_tree.transform(fun)
transformed_tree.sql()
```

Which outputs:

```python
SELECT a FROM x
```

Allowing node deletion could certainly lead to problems, eg. removing the only condition in a where clause, although the same thing could be said about other transformations too, so this seems like a useful feature to have.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -7,7 +7,7 @@ from copy import deepcopy
 from enum import auto
 
 from sqlglot.errors import ParseError
-from sqlglot.helper import AutoName, camel_to_snake_case, ensure_list
+from sqlglot.helper import AutoName, camel_to_snake_case, ensure_list, list_get
 
 
 class _Expression(type):
@@ -344,7 +344,7 @@ class Expression(metaclass=_Expression):
 
         return indent + left + right
 
-    def transform(self, fun, *args, copy=True, **kwargs):
+    def transform(self, fun, *args, copy=True, allow_removals=False, **kwargs):
         """
         Recursively visits all tree nodes (excluding already transformed ones)
         and applies the given transformation function to each node.
@@ -362,14 +362,18 @@ class Expression(metaclass=_Expression):
         new_node = fun(node, *args, **kwargs)
 
         if new_node is None:
-            raise ValueError("A transformed node cannot be None")
+            if not allow_removals:
+                raise ValueError("A transformed node cannot be None")
+            return new_node
         if not isinstance(new_node, Expression):
             return new_node
         if new_node is not node:
             new_node.parent = node.parent
             return new_node
 
-        replace_children(new_node, lambda child: child.transform(fun, *args, copy=False, **kwargs))
+        replace_children(
+            new_node, lambda child: child.transform(fun, *args, copy=False, allow_removals=allow_removals, **kwargs)
+        )
         return new_node
 
     def replace(self, expression):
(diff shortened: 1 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -3009,7 +3013,7 @@ def replace_children(expression, fun):
             else:
                 new_child_nodes.append(cn)
 
-        expression.args[k] = new_child_nodes if is_list_arg else new_child_nodes[0]
+        expression.args[k] = new_child_nodes if is_list_arg else list_get(new_child_nodes, 0)
```
GeorgeSittas, line 3016: We can't leave this as `new_child_nodes[0]`: if the transformation returns `None`, we'll get an index error because the innermost loop won't be executed and `new_child_nodes` will remain an empty list.

sqlglot/expressions.py
```diff
@@ -344,7 +344,7 @@ def to_s(self, hide_missing=True, level=0):
 
         return indent + left + right
 
-    def transform(self, fun, *args, copy=True, **kwargs):
+    def transform(self, fun, *args, copy=True, allow_removals=False, **kwargs):
         """
```
tobymao, line 348: i think we can get rid of the allow removals arg

@barakalon what so you think

GeorgeSittas: Shouldn't we keep this info local to `transform`? Since it's not something we would use inside our callback.

tobymao: what do you mean? i'm saying we could just always allow_removals?

GeorgeSittas: Oh ok, I thought you meant to just do `if kwargs.get("allow_removals") ...` instead.

sqlglot/expressions.py
```diff
@@ -344,7 +344,7 @@ def to_s(self, hide_missing=True, level=0):
 
         return indent + left + right
 
-    def transform(self, fun, *args, copy=True, **kwargs):
+    def transform(self, fun, *args, copy=True, allow_removals=False, **kwargs):
```
barakalon, line 347: We should add a note about returning None in the docstring

sqlglot/expressions.py
barakalon: Yeah I think we should always allow removals.

tests/test_expressions.py
```diff
@@ -247,6 +247,16 @@ def fun(node):
 
         self.assertEqual(expression.transform(fun).sql(), "SELECT a, b FROM x")
 
+    def test_transform_node_removal(self):
```
barakalon, line 250: Other tests to consider adding:
1. Remove a node that isn't a list arg
2. Remove all nodes from a list arg
3. Remove the root node

barakalon (approved):
Sweet!

Does this also support `expression.replace(None)`? Or maybe a `expression.pop()` method?

GeorgeSittas:
> Sweet!
> 
> Does this also support `expression.replace(None)`? Or maybe a `expression.pop()` method?

Yes, it does support `expression.replace(None)`, even without these changes. Observe:

```python

expression_tree = parse_one("SELECT a, b FROM x")

for col in expression_tree.find_all(exp.Column):
    if col.name == "b":
        col.replace(None)

expression_tree.sql()
```

Which outputs:

```python
SELECT a FROM x
```

tobymao (approved):

PR #477, opened 2022-09-23
Dialect reset tokenizer

liangchenm:
support concurrent parse


tobymao:
i've chosen to fix this another way. let me know if 085c13045 fixes your problems

liangchenm:
Too many changes in the latest version.  In previous versions, it was solved by resetting the tokenizer

PR #480, opened 2022-09-23
Add parsing logic for Position function

alieh-rymasheuski:


sqlglot/parser.py
```diff
@@ -1896,6 +1897,12 @@ def _parse_convert(self):
             to = None
         return self.expression(exp.Cast, this=this, to=to)
 
+    def _parse_position(self):
+        this = self._parse_term()
+        if self._match(TokenType.IN):
+            string = self._parse_term()
```
GeorgeSittas, line 1903: You could do parse_bitwise here to capture more complex expressions as arguments

alieh-rymasheuski: Done.

sqlglot/expressions.py
```diff
@@ -2353,6 +2353,10 @@ class Posexplode(Func):
     pass
 
 
+class Position(Func):
```
tobymao, line 2356: is this the same as StrPosition? can we reuse that?

alieh-rymasheuski: Yes and no. It is in different syntax. From the [docs](https://www.postgresql.org/docs/current/functions-string.html):
```
 strpos ( string text, substring text ) → integer

Returns first starting index of the specified substring within string, or zero if it's not present. (Same as position(substring in string), but note the reversed argument order.)

strpos('high', 'ig') → 2
```
I wonder as well if we can somehow tie them together. @tobymao, do you suggest updating the parser to produce `StrPosition` object after parsing `POSITION(foo in bar)` expression?

tobymao: yes that’s what i’m suggesting

alieh-rymasheuski: @tobymao, I think it's done.

alieh-rymasheuski: Should we turn the tables and instead make `StrPosition` generate calls to standard functions, as opposed to vendor-specific functions, by default?

tobymao (approved):

PR #484, opened 2022-09-23
Small fixes

GeorgeSittas:


sqlglot/helper.py
```diff
@@ -100,7 +100,7 @@ def csv_reader(table):
     Returns a csv reader given the expression READ_CSV(name, ['delimiter', '|', ...])
 
     Args:
-        expression (Expression): An anonymous function READ_CSV
+        table (exp.Table): An anonymous function READ_CSV disguised as a table expression
```
GeorgeSittas, line 103: I think this might make the intent more clear. What do you think?

tobymao: it's not correct since READ_CSV is not a table, it's an anonymous function

GeorgeSittas: Discussed in slack. It's a table that contains an anonymous function:
```
>>> sqlglot.parse_one("select * from read_csv('filename')")
(SELECT expressions:
  (STAR ), from:
  (FROM expressions:
    (TABLE this:
      (ANONYMOUS this: read_csv, expressions:
        (LITERAL this: filename, is_string: True)))))
```

PR #486, opened 2022-09-24
Add values to unnest support for bigquery

eakmanrq:
BigQuery doesn't support `VALUES` for statically creating a table with values but it does support the same functionality using `UNNEST` on a struct. One tricky part is that it requires that the `UNNEST` not be a subquery like what `VALUES` commonly is. Therefore on BigQuery I check if the target of a subquery is a VALUES expression and if not then I parse it like normal otherwise I convert the subquery to an `UNNEST`.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -34,13 +34,13 @@ def _date_add_sql(data_type, kind):
 
 
 def _subquery_to_unnest_if_values(self, expression):
-    if not isinstance(expression.args.get("this"), exp.Values):
+    if not isinstance(expression.this, exp.Values):
         return self.subquery_sql(expression)
-    rows = [list(tuple_exp.find_all(exp.Literal)) for tuple_exp in expression.args["this"].find_all(exp.Tuple)]
+    rows = [list(tuple_exp.find_all(exp.Literal)) for tuple_exp in expression.this.find_all(exp.Tuple)]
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -33,6 +33,21 @@ def func(self, expression):
     return func
 
 
+def _subquery_to_unnest_if_values(self, expression):
+    if not isinstance(expression.args.get("this"), exp.Values):
```
tobymao, line 37: expression.this

sqlglot/dialects/bigquery.py
```diff
@@ -33,6 +33,21 @@ def func(self, expression):
...
+        return self.subquery_sql(expression)
+    rows = [list(tuple_exp.find_all(exp.Literal)) for tuple_exp in expression.args["this"].find_all(exp.Tuple)]
```
tobymao, line 39: for tuple_exp in expression.this

sqlglot/dialects/bigquery.py
```diff
@@ -33,6 +33,21 @@ def func(self, expression):
...
+    structs = []
+    for row in rows:
+        aliases = [
+            exp.Alias(this=value, alias=column_name)
```
tobymao, line 43: exp.alias_(value, column_name)

tobymao (approved):

PR #488, opened 2022-09-26
Improve unalias group by transform

eakmanrq:
Two examples cases that are not handled correctly with existing unalias:
### Example 1 (Valid but unnecessary change)
```
SELECT cola AS cola, sum(colb) FROM x GROUP BY cola
```
This currently returns
```
SELECT cola AS cola, sum(colb) FROM x GROUP BY 1
```
This is valid but it is unnecessary since the existing group by expression is valid.
### Example 2 (Invalid result)
```
SELECT TO_DATE(cola) AS cola, sum(colb) FROM x GROUP BY TO_DATE(cola)
```
This currently returns 
```
SELECT TO_DATE(cola) AS cola, sum(colb) FROM x GROUP BY TO_DATE(1)
```
This is invalid. 

This change covers both cases to ensure that we only replace with a number when it is actually needed.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -201,6 +201,21 @@ class Expression(metaclass=_Expression):
             ancestor = ancestor.parent
         return ancestor
 
+    def find_ancestor_with_parent(self, *expression_types):
+        """
+        Returns a nearest parent with a parent matching expression_types
+
+        Args:
+            expression_types (type): the expression type(s) to match
+
+        Returns:
+            the parent node
+        """
+        ancestor = self
+        while ancestor and not isinstance(ancestor.parent, expression_types):
+            ancestor = ancestor.parent
+        return ancestor
+
     @property
     def parent_select(self):
         """
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -211,7 +211,7 @@ class Expression(metaclass=_Expression):
         Returns:
             the parent node
         """
-        ancestor = self
+        ancestor = self.parent
         while ancestor and not isinstance(ancestor.parent, expression_types):
             ancestor = ancestor.parent
         return ancestor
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -201,6 +201,21 @@ def find_ancestor(self, *expression_types):
             ancestor = ancestor.parent
         return ancestor
 
+    def find_ancestor_with_parent(self, *expression_types):
```
tobymao, line 204: maybe this hould just be a parameter on find_ancestor?

eakmanrq: I prefer the separate function but open to combining the two if there a strong vote for that.

sqlglot/transforms.py
```diff
@@ -12,14 +12,17 @@ def unalias_group(expression):
...
+            if isinstance(e, exp.Alias)
         }
 
         expression = expression.copy()
 
         for col in expression.find_all(exp.Column):
-            alias_index = aliased_selects.get(col.name)
-            if not col.table and alias_index:
+            alias_index, col_expression = aliased_selects.get(col.name, (None, None))
+            group_by_expression = col.find_ancestor_with_parent(exp.Group)
```
barakalon, line 24: Would it be simpler to just iterate over the immediate children of the Group expression?

i.e.
```python
for col in expression.children:
    if isinstance(col, exp.Column) and ...:
        ...
```

Then we don't need to traverse down and back up the tree.

barakalon (commented):
🤦 oof, that was a silly error of mine. Thanks for fixing this!

sqlglot/transforms.py
eakmanrq: Nice suggestion! I made that change in the latest commit.

tobymao (approved):

PR #495, opened 2022-09-27
Rewrite derived tables as CTEs

barakalon:
Optimizer step to rewrite subqueries as CTEs.

This shouldn't be part of our standard optimizer steps.

cc @jeffwong

### DIFF
--- a/sqlglot/executor/__init__.py
+++ b/sqlglot/executor/__init__.py
@@ -4,7 +4,7 @@ import time
 from sqlglot import parse_one
 from sqlglot.executor.python import PythonExecutor
 from sqlglot.optimizer import RULES, optimize
-from sqlglot.optimizer.merge_derived_tables import merge_derived_tables
+from sqlglot.optimizer.merge_subqueries import merge_subqueries
 from sqlglot.planner import Plan
 
 logger = logging.getLogger("sqlglot")
@@ -12,7 +12,7 @@ logger = logging.getLogger("sqlglot")
 OPTIMIZER_RULES = list(RULES)
 
 # The executor needs isolated table selects
-OPTIMIZER_RULES.remove(merge_derived_tables)
+OPTIMIZER_RULES.remove(merge_subqueries)
 
 
 def execute(sql, schema, read=None):
--- a/sqlglot/executor/__init__.py
+++ b/sqlglot/executor/__init__.py
@@ -3,16 +3,10 @@ import time
 
 from sqlglot import parse_one
 from sqlglot.executor.python import PythonExecutor
-from sqlglot.optimizer import RULES, optimize
-from sqlglot.optimizer.merge_subqueries import merge_subqueries
+from sqlglot.optimizer import optimize
 from sqlglot.planner import Plan
 
 logger = logging.getLogger("sqlglot")
-
-OPTIMIZER_RULES = list(RULES)
-
-# The executor needs isolated table selects
-OPTIMIZER_RULES.remove(merge_subqueries)
 
 
 def execute(sql, schema, read=None):
@@ -34,7 +28,7 @@ def execute(sql, schema, read=None):
     """
     expression = parse_one(sql, read=read)
     now = time.time()
-    expression = optimize(expression, schema, rules=OPTIMIZER_RULES)
+    expression = optimize(expression, schema, leave_tables_isolated=True)
     logger.debug("Optimization finished: %f", time.time() - now)
     logger.debug("Optimized SQL: %s", expression.sql(pretty=True))
     plan = Plan(expression)
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -89,11 +89,7 @@ def _eliminate(scope, existing_ctes, taken):
     if scope.is_union:
         return _eliminate_union(scope, existing_ctes, taken)
 
-    if (scope.is_derived_table and not isinstance(scope.expression, (exp.Unnest, exp.Lateral))) or (
-        scope.is_subquery
-        and not scope.is_correlated_subquery
-        and not scope.expression.find_ancestor(exp.Having, exp.Join)
-    ):
+    if scope.is_derived_table and not isinstance(scope.expression, (exp.Unnest, exp.Lateral)):
         return _eliminate_derived_table(scope, existing_ctes, taken)
 
 
@@ -140,9 +136,7 @@ def _eliminate_derived_table(scope, existing_ctes, taken):
 
     taken[name] = scope
 
-    table = exp.table_(name)
-    if not scope.is_subquery:
-        table = exp.alias_(table, alias=alias)
+    table = exp.alias_(exp.table_(name), alias=alias)
     parent.replace(table)
 
     if not duplicate_cte_alias:
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -52,8 +52,7 @@ def eliminate_subqueries(expression):
     with_ = root.expression.args.get("with")
     if with_:
         for cte in with_.expressions:
-            existing_ctes[cte.this]: cte.alias
-
+            existing_ctes[cte.this] = cte.alias
     new_ctes = []
 
     # We're adding more CTEs, but we want to maintain the DAG order.
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -16,6 +16,11 @@ def eliminate_subqueries(expression):
         >>> eliminate_subqueries(expression).sql()
         'WITH y AS (SELECT * FROM x) SELECT a FROM y AS y'
 
+    This also deduplicates common subqueries:
+        >>> expression = sqlglot.parse_one("SELECT a FROM (SELECT * FROM x) AS y JOIN (SELECT * FROM x) AS z")
+        >>> eliminate_subqueries(expression).sql()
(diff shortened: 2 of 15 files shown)


sqlglot/optimizer/optimizer.py
```diff
@@ -9,6 +9,9 @@
 from sqlglot.optimizer.qualify_columns import qualify_columns
 from sqlglot.optimizer.qualify_tables import qualify_tables
 from sqlglot.optimizer.quote_identities import quote_identities
+from sqlglot.optimizer.rewrite_derived_tables_as_ctes import (  # noqa
```
tobymao, line 12: what's thi lint for

sqlglot/optimizer/rewrite_derived_tables_as_ctes.py
```diff
@@ -0,0 +1,40 @@
...
+        >>> expression = sqlglot.parse_one("SELECT a FROM (SELECT * FROM x) AS y")
+        >>> rewrite_derived_tables_as_ctes(expression).sql()
+        'WITH y AS (SELECT * FROM x) SELECT a FROM y'
+
+    Args:
+        expression (sqlglot.Expression): expression to optimize
+    Returns:
+        sqlglot.Expression: optimized expression
+    """
+    expression = expression.copy()
```
tobymao, line 21: optimizer methods shouldn't have copies, the implicit standard is for the caller to call copy

sqlglot/optimizer/optimizer.py
barakalon: Since it's an unused import.

I'll just remove this import.

sqlglot/optimizer/rewrite_derived_tables_as_ctes.py
barakalon: good call

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,48 +1,149 @@
...
 
 def eliminate_subqueries(expression):
     """
-    Rewrite duplicate subqueries from sqlglot AST.
+    Rewrite subqueries as CTES, deduplicating if possible.
 
     Example:
         >>> import sqlglot
-        >>> expression = sqlglot.parse_one("SELECT 1 AS x, 2 AS y UNION ALL SELECT 1 AS x, 2 AS y")
+        >>> expression = sqlglot.parse_one("SELECT a FROM (SELECT * FROM x) AS y")
```
tobymao, line 15: the example is not as comprehensive, it doesn't show the deduplication behavior, can you modify the example to show deduplication as well?

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,48 +1,149 @@
...
-        queries[query] = queries.get(query, []) + [query]
+    # Map of Expression->alias
+    # Existing CTES in the root expression. We'll use this for deduplication.
+    existing_ctes = {}
 
-    sequence = itertools.count()
+    with_ = root.expression.args.get("with")
+    if with_:
+        for cte in with_.expressions:
+            existing_ctes[cte.this]: cte.alias
```
tobymao, line 55: typo even if it works

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,48 +1,149 @@
...
+        expression.set("with", exp.With(expressions=new_ctes))
 
     return expression
+
+
+def _eliminate(scope, existing_ctes, taken):
+    if scope.is_union:
+        return _eliminate_union(scope, existing_ctes, taken)
+
+    elif (scope.is_derived_table and not isinstance(scope.expression, (exp.Unnest, exp.Lateral))) or (
```
tobymao, line 88: if is unnecesasry

tobymao: elif, it should just be if i think

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,48 +1,153 @@
...
+
+
+def _eliminate(scope, existing_ctes, taken):
+    if scope.is_union:
+        return _eliminate_union(scope, existing_ctes, taken)
+
+    if (scope.is_derived_table and not isinstance(scope.expression, (exp.Unnest, exp.Lateral))) or (
+        scope.is_subquery
+        and not scope.is_correlated_subquery
+        and not scope.expression.find_ancestor(exp.Having, exp.Join)
```
tobymao, line 95: why do you check if the ancestor is a join?

tests/fixtures/optimizer/eliminate_subqueries.sql
```diff
@@ -1,42 +1,67 @@
...
+SELECT 1 AS x, 2 AS y UNION ALL SELECT 1 AS x, 2 AS y;
+WITH cte AS (SELECT 1 AS x, 2 AS y) SELECT cte.x AS x, cte.y AS y FROM cte AS cte UNION ALL SELECT cte.x AS x, cte.y AS y FROM cte AS cte;
+
+-- Union of selects with derived tables
+(SELECT a FROM (SELECT b FROM x)) UNION (SELECT a FROM (SELECT b FROM y));
+WITH cte AS (SELECT b FROM x), cte_2 AS (SELECT a FROM cte AS cte), cte_3 AS (SELECT b FROM y), cte_4 AS (SELECT a FROM cte_3 AS cte_3) (SELECT cte_2.a AS a FROM cte_2 AS cte_2) UNION (SELECT cte_4.a AS a FROM cte_4 AS cte_4);
+
+-- Subquery
+SELECT a FROM x WHERE b = (SELECT y.c FROM y);
+WITH cte AS (SELECT y.c FROM y) SELECT a FROM x WHERE b = cte;
```
tobymao, line 43: is this legal? i'm not sure we want to do this, there's logic in rewriting subqueries like this into joins

sqlglot/executor/__init__.py
```diff
@@ -4,15 +4,15 @@
...
+from sqlglot.optimizer.merge_subqueries import merge_subqueries
 from sqlglot.planner import Plan
 
 logger = logging.getLogger("sqlglot")
 
 OPTIMIZER_RULES = list(RULES)
 
 # The executor needs isolated table selects
-OPTIMIZER_RULES.remove(merge_derived_tables)
+OPTIMIZER_RULES.remove(merge_subqueries)
```
tobymao, line 15: the executor i think may want some of merge subqueries, so it wants to remove some duplicate stuff but not de isolate subqueries? do you think that's feasible?

sqlglot/optimizer/eliminate_subqueries.py
barakalon: duckdb doesn't like it:
```sql
WITH b AS (
  SELECT 1
)
SELECT
  *
FROM x
GROUP BY a
HAVING a > b
# Binder Error: column b must appear in the GROUP BY clause or be used in an aggregate function
```

```sql
WITH b AS (
  SELECT 1
)
SELECT
  *
FROM x
JOIN y
  ON y.a = b
# RuntimeError: Binder Error: Referenced column "b" not found in FROM clause!
# Candidate bindings: "y.a", "x.a"
# LINE 9:   ON y.a = b
#                    ^
```

The tpc-h tests caught this.

This bug existed before in eliminate_subqueries - we just weren't catching it because there were never any common subqueries in an ON or HAVING clause.

tests/fixtures/optimizer/eliminate_subqueries.sql
barakalon: No, it's not legal.

I don't know why I'm doing this. I think I thought the old code did?

I'll remove it.

sqlglot/executor/__init__.py
barakalon: eliminate_subqueries removes duplicates.

barakalon: Maybe we should run "isolate_table_selects" last, during the executor?

tobymao: yea that makes sense

tobymao: why do you need isolate_table_selects again? i removed it and it actually made the optimized sql look better, (predicate pushdown was further)

tobymao: i guess it hurt some of the deduplication

tobymao: ah hah! your merge_where hurts predicate pushdown, it's merging to the where when it should be merging to the join ON

tobymao: you have a bug here

```
    if isinstance(from_or_join, exp.Join) and from_or_join.side:
        # Merge predicates from an outer join to the ON clause
        from_or_join.on(where.this, copy=False)
        from_or_join.set("on", simplify(from_or_join.args.get("on")))
```

it should be not from_or_join.side since you cannot combine predicates in a left join, only inner joins

tobymao: maybe i'm wrong, maybe you can push up both wheres with parents of left join and join.

you definitely cannot push a left join into there where, because that changes left join into a filter which is wrong.

the fix may just be to do

```
if isinstance(from_or_join, exp.Join):
  ...
```

tobymao:
ok i think i figured it out

1. fix the predicate push up with joins (remove the side check)
2. for the merge_derived_tables/merge_subqueries, take in a kwarg to only do cte merging and to merging of subqueries. this way the executor and other systems can take benefit but then customize this particular step

how does that sound?

tobymao:
i think that's fine, and you made the affordance to do so

barakalon:
indeed I did haha
Ok one sec

tobymao (approved):

PR #496, opened 2022-09-27
Fix lambda variables with optimize

eakmanrq:
Prior to this change lambda variables were considered columns and therefore optimize would fail since it would look for columns that didn't exist. Now those variables are excluded but any columns used within a lambda expression are still included.

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -200,9 +200,9 @@ class Scope:
             self._columns = [
                 c
                 for c in columns + external_columns
-                if (not c.find_ancestor(exp.Qualify, exp.Order, exp.Hint)
+                if (not c.find_ancestor(exp.Qualify, exp.Order, exp.Hint, exp.Lambda)
                     or c.table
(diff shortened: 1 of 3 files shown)


barakalon (commented):
I'm a little worried about all these calls to `find_ancestor`. Might be hurting performance.

Is this easier if we just don't traverse into Hint and Lambda expressions? I.e., when collecting? https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L82

tobymao:
@barakalon for this specific one, we've discussed offline that for lambdas, @eakmanrq will post process the expression and turn columns referencing the variables into identifiers

i didn't catch that he added a double call to find_ancestor for the hints, hopefully we can remove that

eakmanrq:
Deprecated by this PR: https://github.com/tobymao/sqlglot/pull/497

eakmanrq:
@barakalon I don't think we can just skip parsing lambdas since they can contain table references in them and we need to be aware of those.

barakalon:
Got it, make sense

PR #497, opened 2022-09-27
Have parser convert lambda variables to identifiers

eakmanrq:
Currently the lambda parser treats the lambda variables as columns when they really are identifiers. This issues becomes more clear when using the optimizer since it would look for the definitions of those columns in sources and the wouldn't exist.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1709,13 +1709,6 @@ class Parser:
         return self._parse_window(this)
 
     def _parse_lambda(self):
-        def replace_lambda_columns_with_identifiers(node, lambda_variables):
-            if isinstance(node, exp.Column):
-                matching_lambda_var = [x for x in lambda_variables if node.name == x.name]
-                if matching_lambda_var:
-                    return matching_lambda_var[0]
-            return node
-
         index = self._index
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -1709,6 +1709,13 @@ def _parse_function(self):
         return self._parse_window(this)
 
     def _parse_lambda(self):
+        def replace_lambda_columns_with_identifiers(node, lambda_variables):
```
tobymao, line 1712: can you make this a first class function on the last line, after _replace_columns_with_dots

change lambda variables to be a set 

self._parse_conjunction().transform(self._replace_lambda, {node.name for node in expressions}

sqlglot/parser.py
```diff
@@ -1709,6 +1709,13 @@ def _parse_function(self):
...
+            if isinstance(node, exp.Column):
+                matching_lambda_var = [x for x in lambda_variables if node.name == x.name]
+                if matching_lambda_var:
```
tobymao, line 1715: if node.name in variables:
   return node.this

tobymao (approved):

PR #502, opened 2022-09-28
Use DATETIME2 in TSQL as default timestamp type

alieh-rymasheuski:
It is the flavor of datetime which "aligns with the SQL Standard".

Docs: https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=azure-sqldw-latest&preserve-view=true

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -32,3 +32,5 @@ class TestTSQL(Validator):
         self.validate_identity("CAST(x AS IMAGE)")
         self.validate_identity("CAST(x AS SQL_VARIANT)")
         self.validate_identity("CAST(x AS BIT)")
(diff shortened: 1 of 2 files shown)


tests/dialects/test_tsql.py
```diff
@@ -32,3 +32,5 @@ def test_types(self):
         self.validate_identity("CAST(x AS IMAGE)")
         self.validate_identity("CAST(x AS SQL_VARIANT)")
         self.validate_identity("CAST(x AS BIT)")
+        self.validate("CAST(x AS DATETIME)", "CAST(x AS DATETIME2)", write="tsql")
```
tobymao, line 35: don't use self.validate, use validate_all

tobymao: i've fixed this for you

PR #503, opened 2022-09-28
Implement type inference for expressions

GeorgeSittas:
This is a draft aiming to introduce type inference for expressions. This PR is for discussion, so any feedback is welcome.

We could use type information in various ways:

- Do some analytical type checking to let the user know if their SQL input is semantically valid w.r.t types.
- Implement further useful optimizer passes that require this information.

The user can provide custom `ANNOTATORS`, `COERCES_TO` dictionaries arguments to `annotate_expression_types`, in order to override the expression annotation & type coercion logic, respectively. This may be useful, as dialects may differ regarding type semantics.

Testing probably needs to be done in a more practical manner, but I haven't thought about this yet.


sqlglot/expressions.py
```diff
@@ -31,7 +31,7 @@ class Expression(metaclass=_Expression):
 
     key = None
     arg_types = {"this": True}
-    __slots__ = ("args", "parent", "arg_key")
+    __slots__ = ("args", "parent", "arg_key", "type")
```
tobymao, line 34: you need to initialize it as self.type = None on line 40

sqlglot/optimizer/annotate_expression_types.py
```diff
@@ -0,0 +1,207 @@
...
+    Args:
+        types (list[exp.DataType.Type]): types, in descending order w.r.t their relative precedence
+    """
+
+    PRECEDED_BY[types[0]] = set()
+    for t1, t2 in zip(types[:-1], types[1:]):
+        PRECEDED_BY[t2] = {t1}.union(PRECEDED_BY[t1])
+
+
+def _create_precedence_mapping():
```
tobymao, line 176: you should not be modifying constants like this.  in functions.

PRECEDED_BY = ... in the global scope,

and then if someone passes in a new one, create a new instance

sqlglot/optimizer/annotate_expression_types.py
```diff
@@ -0,0 +1,207 @@
...
+    exp.Literal: lambda expr, schema, annotators: _annotate_literal(expr, schema, annotators),
+    exp.Boolean: lambda expr, schema, annotators: _annotate_boolean(expr, schema, annotators),
+}
+
+# This maps a type to a set of related types with greater precedence,
+# eg. exp.DataType.Type.FLOAT is mapped to {exp.DataType.Type.DOUBLE}
+PRECEDED_BY = {}
+
+
+def annotate_expression_types(expression, schema=None, annotators=ANNOTATORS):
```
tobymao, line 29: we can just call everything annotate_types

sqlglot/optimizer/annotate_expression_types.py
```diff
@@ -0,0 +1,207 @@
...
+    if not expression:
+        return None
+
+    # Note: maybe there are some expressions that can be allowed to *not* have a type.
+    # In this case, instead of raising we could return None instead. Other alternatives:
+    #
+    # - Always return None if there's no available annotator method (too flexible maybe?).
+    # - Provide a default annotator returning None for expressions we don't care to annotate.
+
+    annotator = annotators.get(type(expression))
```
tobymao, line 70: can you do like expression.__class__

sqlglot/optimizer/annotate_expression_types.py
```diff
@@ -0,0 +1,207 @@
...
+
+# Note: CTEs can augment the schema. In the following example, we need to lookup the
+# schema to find the type of x and thus we should update our schema, after we annotate
+# the CTE, at least temporarily.
+#
+# WITH foo AS (SELECT * FROM bar) SELECT x + 3 FROM foo
+
+
+def _annotate(expression, schema, annotators):
+    if not expression:
```
tobymao, line 61: if not isinstance(expression, exp.Expression)

GeorgeSittas:
Good to go

tobymao (approved):

PR #511, opened 2022-09-30
Classify Values as a Derived Table

eakmanrq:
Some naunces about values that I figured out while working on this:

### Spark
Values with parenthesis around them is valid but the tab alias assigned cannot be used in the select expression if you have the parentheses around them. It does work correctly though if there is no parentheses around values so we remove them by default now.

### Snowflake
You cannot put double quotes around the column names in `tab(<col>)` alias and they cannot have quotes in the select expression. There will be a follow up PR to ensure snowflake doesn't do that when using optimize to add the quotes.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -77,6 +77,7 @@ class Dialect(metaclass=_Dialect):
     alias_post_tablesample = False
     normalize_functions = "upper"
     null_ordering = "nulls_are_small"
+    wrapped_derived_values = True
 
     date_format = "'%Y-%m-%d'"
     dateint_format = "'%Y%m%d'"
@@ -169,6 +170,7 @@ class Dialect(metaclass=_Dialect):
                 "alias_post_tablesample": self.alias_post_tablesample,
                 "normalize_functions": self.normalize_functions,
                 "null_ordering": self.null_ordering,
+                "wrapped_derived_values": self.wrapped_derived_values,
                 **opts,
             }
         )
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -77,7 +77,7 @@ class Dialect(metaclass=_Dialect):
     alias_post_tablesample = False
     normalize_functions = "upper"
     null_ordering = "nulls_are_small"
-    wrapped_derived_values = True
+    wrap_derived_values = True
 
     date_format = "'%Y-%m-%d'"
     dateint_format = "'%Y%m%d'"
@@ -170,7 +170,7 @@ class Dialect(metaclass=_Dialect):
                 "alias_post_tablesample": self.alias_post_tablesample,
                 "normalize_functions": self.normalize_functions,
                 "null_ordering": self.null_ordering,
-                "wrapped_derived_values": self.wrapped_derived_values,
+                "wrap_derived_values": self.wrap_derived_values,
                 **opts,
             }
         )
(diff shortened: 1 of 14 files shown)


sqlglot/optimizer/scope.py
```diff
@@ -399,7 +399,7 @@ def _traverse_scope(scope):
         yield from _traverse_select(scope)
     elif isinstance(scope.expression, exp.Union):
         yield from _traverse_union(scope)
-    elif isinstance(scope.expression, (exp.Lateral, exp.Unnest)):
+    elif isinstance(scope.expression, (exp.Lateral, exp.Unnest, exp.Values)):
```
tobymao, line 402: (exp.Unnest, exp.Lateral, exp.Values) is used in many places

can we make the inheritance tree something

class DerivedTable:

class LiteralDerivedTable(DerivedTable): ( or some better name )

and then just check for instance of that in like the 4 places where this is used

sqlglot/parser.py
```diff
@@ -1186,6 +1191,24 @@ def _parse_unnest(self):
             alias=alias,
         )
 
+    def _parse_derived_table_values(self):
+        is_derived = self._curr is not None and self._curr.token_type == TokenType.L_PAREN
```
tobymao, line 1195:
```suggestion
        is_derived = self._match_pair(TokenType.L_PAREN, TokenType.Values)
        
        if not is_derived or self._match(TokenType.VALUES):
            return None
```

sqlglot/parser.py
```diff
@@ -1186,6 +1191,24 @@ def _parse_unnest(self):
...
+        if not self._match(TokenType.VALUES) and not self._match_pair(TokenType.L_PAREN, TokenType.VALUES):
+            return None
+
+        expressions = self._parse_csv(self._parse_paren_enclosed_csv, self._parse_column)
```
tobymao, line 1199: i don't think any of these new methods are needed, just do

self._parse_value()

eakmanrq: Damn thanks

sqlglot/optimizer/scope.py
eakmanrq: I came up with `SelectlessDerivedTable`. Still not great but does better communicate the aspect that makes these unique which is that they don't have a `SELECT` expression in their definition.

sqlglot/dialects/dialect.py
```diff
@@ -77,6 +77,7 @@ class Dialect(metaclass=_Dialect):
     alias_post_tablesample = False
     normalize_functions = "upper"
     null_ordering = "nulls_are_small"
+    wrapped_derived_values = True
```
tobymao, line 80: change the tense, to wrap_derived_values

sqlglot/optimizer/qualify_columns.py
```diff
@@ -5,7 +5,7 @@
 from sqlglot.optimizer.schema import ensure_schema
 from sqlglot.optimizer.scope import traverse_scope
 
-SKIP_QUALIFY = (exp.Unnest, exp.Lateral)
+SKIP_QUALIFY = exp.SelectlessDerivedTable
```
tobymao, line 8: you can get rid of this constant now

sqlglot/optimizer/scope.py
```diff
@@ -314,8 +314,8 @@ def is_root(self):
 
     @property
     def is_unnest(self):
```
tobymao, line 316: change the name of this too?

sqlglot/parser.py
```diff
@@ -2206,12 +2229,12 @@ def _parse_replace(self):
         self._match_r_paren()
         return columns
 
-    def _parse_csv(self, parse):
-        parse_result = parse()
+    def _parse_csv(self, parse, *args, **kwargs):
```
tobymao, line 2232: this change is no longer necessary

sqlglot/optimizer/scope.py
eakmanrq: Ended up changing to UDTF (User Defined Table Function).

tobymao (approved):

PR #515, opened 2022-09-30
optimizer debug mode

barakalon:
It can be hard to debug the optimizer.

Maybe this will help.

@GeorgeSittas


tobymao:
i don't think this is necessary, you can just put print statements in the before and after for each optimizer when debugging

barakalon:
Writing to a file is kind of helpful

GeorgeSittas:
I agree that a debugger could make things a little easier. Printing is kind of cumbersome when you have multiple steps in the optimizer and trying to pinpoint which one causes a specific transformation.

PR #527, opened 2022-10-03
Scope.walk, Scope.find, and Scope.find_all

barakalon:
We already have logic for traversing within a scope. Let's expose it as a public method.

Solves this: https://github.com/tobymao/sqlglot/discussions/522#discussioncomment-3783662

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -109,35 +109,7 @@ class Scope:
             self._collect()
 
     def walk(self, bfs=True):
-        """
-        Returns a generator object which visits all nodes in this scope.
-
-        This does NOT traverse into subscopes.
-
-        Args:
-            bfs (bool): if set to True the BFS traversal order will be applied,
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/scope.py
```diff
@@ -86,37 +86,93 @@ def _collect(self):
...
+        for node, parent, _ in self.walk(bfs=False):
             if node is self.expression:
                 continue
-            if isinstance(node, exp.Column) and not isinstance(node.this, exp.Star):
+            elif isinstance(node, exp.Column) and not isinstance(node.this, exp.Star):
                 self._raw_columns.append(node)
             elif isinstance(node, exp.Table):
                 self._tables.append(node)
             elif isinstance(node, exp.UDTF):
                 self._derived_tables.append(node)
```
GeorgeSittas, line 97: We can merge this case with the one in line 100 by doing `isinstance(node, (exp.UDTF, exp.Subquery))`.

tobymao (approved):

PR #528, opened 2022-10-03
Add support for column type annotation

GeorgeSittas:
Some notes / thoughts I had while developing these changes:

1. I'm not sure if `TYPE_MAPPING` is a good name, provided we already have this in Generator. Maybe add a `SCHEMA_` prefix?

2. I assume that the `qualify_columns` step is a dependency to this one, because it makes our job easier (for instance, we don't need to think about column ambiguities).

3. We need to display types in a more user-friendly way because doing `<expr>.type` is cumbersome.

4. We're currently annotating sub-expressions while simultaneously traversing scopes, which means that we might visit some of them more than once. I've added a check to avoid the annotation process if it's already done, but there may still be unnecessary overhead to this. Let me know if there's a better way to cache annotated expressions (such as SELECT), so that we don't even have to visit the children expressions again. One idea is to add some sentinel type for SELECT etc instead of having `type = None`, so that we can prune tree branches early.


sqlglot/optimizer/annotate_types.py
```diff
@@ -18,11 +22,13 @@ def annotate_types(expression, schema=None, annotators=None, coerces_to=None):
         schema (dict|sqlglot.optimizer.Schema): Database schema.
         annotators (dict): Maps expression type to corresponding annotation function.
         coerces_to (dict): Maps expression type to set of types that it can be coerced into.
+        type_mapping (dict): Maps type (str) to an expression type (DataType). An example:
```
tobymao, line 25: i don't think we want type_mapping at all here, it should be part of the schema class and it should leverage sqlglot's parser

GeorgeSittas: So, being in schema means that we'd had to `ensure_schema` at the top, correct? Should we add a `get` method to the MappingSchema as well?

tobymao: we should only be dealing with schema objects here yes

sqlglot/optimizer/annotate_types.py
```diff
@@ -97,22 +104,54 @@ class TypeAnnotator:
...
+        scopes = traverse_scope(expression) if expression.__class__ in self.TRAVERSABLES else []
+
+        for scope in scopes:
+            # First annotate the current expression's columns references
+            for col in scope.columns:
+                source = scope.sources[col.table]
+                if isinstance(source, exp.Table):
+                    col.type = self._convert_schema_type(source, col)
+                else:
+                    matching_column = next(s for s in source.selects if col.name == s.alias)
```
barakalon, line 125: s.alias_or_name is safer

sqlglot/optimizer/annotate_types.py
```diff
@@ -97,22 +104,54 @@ class TypeAnnotator:
...
         self.annotators = annotators or self.ANNOTATORS
         self.coerces_to = coerces_to or self.COERCES_TO
+        self.type_mapping = type_mapping or TYPE_MAPPING
 
     def annotate(self, expression):
+        scopes = traverse_scope(expression) if expression.__class__ in self.TRAVERSABLES else []
+
+        for scope in scopes:
+            # First annotate the current expression's columns references
+            for col in scope.columns:
```
barakalon, line 120: I don't think `_annotate_args` stops at subscopes - Will we be passing over the subscopes multiple times?

GeorgeSittas: Yes, we need to somehow mark expressions as "processed", otherwise they'll be visited multiple times. Note that this mainly applies to expressions like SELECT, WITH, WHERE etc that don't have a type. After we annotate an expression we don't explore its subtree again, because there's that `if expression.type: return` check in `_annotate`.

GeorgeSittas: I can try the "sentinel type" approach so that select.type doesn't yield None if we've visited that select expression in the past and we can prune anyway.

barakalon: Ah, gotcha

tobymao: how about None, DataType.UNKNOWN

barakalon: Here's how scope itself prunes the tree: https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L522

barakalon: But I think it's even easier in your case, as you can just pass in a prune callback that checks for the `type` annotation

GeorgeSittas: I think the `set_type` flag can solve the original issue discussed here. Let me know if I missed something.

sqlglot/optimizer/annotate_types.py
```diff
@@ -97,22 +103,47 @@ class TypeAnnotator:
...
 
+    TRAVERSABLES = (exp.Select, exp.Union, exp.UDTF, exp.Subquery)
+
     def __init__(self, schema=None, annotators=None, coerces_to=None):
         self.schema = schema
         self.annotators = annotators or self.ANNOTATORS
         self.coerces_to = coerces_to or self.COERCES_TO
 
     def annotate(self, expression):
+        scopes = traverse_scope(expression) if isinstance(expression, self.TRAVERSABLES) else []
```
tobymao, line 114: why bother with this? why not just do

if isinstance(expression, self.Traversables):
   for scope in traverse_scope(...)

sqlglot/optimizer/schema.py
```diff
@@ -60,6 +80,20 @@ def column_names(self, table):
...
+        Convert a type represented as a string to the corresponding exp.DataType.Type object.
+
+        Args:
+            type_as_string (str): The type we want to convert.
+        Returns:
+            sqlglot.expressions.DataType.Type: The resulting expression type.
+        """
+
+        return self._type_mapping.get(type_as_string) or self._type_mapping.setdefault(
+            type_as_string, exp.maybe_parse(type_as_string, into=exp.DataType).this
```
tobymao, line 94: this could fail and you'll get a weird error with None.this

sqlglot/optimizer/schema.py
```diff
@@ -60,6 +80,20 @@ def column_names(self, table):
...
+        """
+        Convert a type represented as a string to the corresponding exp.DataType.Type object.
+
+        Args:
+            type_as_string (str): The type we want to convert.
+        Returns:
+            sqlglot.expressions.DataType.Type: The resulting expression type.
+        """
+
+        return self._type_mapping.get(type_as_string) or self._type_mapping.setdefault(
```
tobymao, line 93: let's delete type_mapping and instead just leverage dialect?

sqlglot/optimizer/annotate_types.py
```diff
@@ -97,22 +103,46 @@ class TypeAnnotator:
...
     def annotate(self, expression):
+        if isinstance(expression, self.TRAVERSABLES):
+            for scope in traverse_scope(expression):
+                # First annotate the current expression's columns references
+                for col in scope.columns:
+                    source = scope.sources[col.table]
+                    if isinstance(source, exp.Table):
+                        col.type = self.schema.get_column_type(source, col)
+                    else:
+                        matching_column = next(s for s in source.selects if col.name == s.alias_or_name)
```
tobymao, line 122: get rid of the double for loop. 

selects = {s.alias_or_name: s for s in source.selects}

tobymao (approved):

PR #529, opened 2022-10-03
add support for invisible columns

ginter:
add support for invisible columns ([ref](https://dev.mysql.com/doc/refman/8.0/en/invisible-columns.html)). This will let us have more control over the columns that get expanded for select stars in the optimizer.

cc @barakalon

### DIFF
--- a/sqlglot/optimizer/schema.py
+++ b/sqlglot/optimizer/schema.py
@@ -9,12 +9,13 @@ class Schema(abc.ABC):
     """Abstract base class for database schemas"""
 
     @abc.abstractmethod
-    def column_names(self, table):
+    def column_names(self, table, only_visible):
         """
         Get the column names for a table.
 
         Args:
             table (sqlglot.expressions.Table): Table expression instance
+            only_visible (bool): Whether to include invisible columns
         Returns:
             list[str]: list of column names
         """
@@ -49,7 +50,7 @@ class MappingSchema(Schema):
 
         self.forbidden_args = {"catalog", "db", "this"} - set(self.supported_table_args)
 
-    def column_names(self, table):
+    def column_names(self, table, only_visible=False):
         if not isinstance(table.this, exp.Identifier):
             return fs_get(table)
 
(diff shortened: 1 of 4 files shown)


sqlglot/optimizer/schema.py
```diff
@@ -9,12 +9,13 @@ class Schema(abc.ABC):
     """Abstract base class for database schemas"""
 
     @abc.abstractmethod
-    def column_names(self, table):
+    def column_names(self, table, only_visible):
```
barakalon, line 12: set default to False: `only_visible=False`

tests/fixtures/optimizer/qualify_columns__with_invisible.sql
```diff
@@ -0,0 +1,68 @@
+--------------------------------------
```
barakalon, line 1: A lot of these seems copy&pasted from the normal test case.
Do we need all of them?
Duplicate tests cases can be a pain.

barakalon (approved):

tests/fixtures/optimizer/qualify_columns__with_invisible.sql
ginter: I removed a lot of them. the ones I left are mainly dealing with select stars although there are a few testing direct selection of invisible columns.

some are duplicated that's true but it's still different since the schema in this test includes invisible columns. I'll look again tho to see if any don't seem worth it

tests/test_optimizer.py
```diff
@@ -88,6 +88,27 @@ def qualify_columns(expression, **kwargs):
 
         self.check_file("qualify_columns", qualify_columns, schema=self.schema)
 
+    def test_qualify_columns__with_invisible(self):
+        class SchemaWithInvisibleColumns(Schema):
+            def __init__(self, schema, visible):
```
barakalon, line 93: Let's add an optional `visible` parameter to `MappingSchema` and have it do exactly what you do here.

Then we don't need this test class and the `MappingSchema` class becomes a more full implementation of the interface.

PR #535, opened 2022-10-05
Add support for function annotation

GeorgeSittas:


sqlglot/optimizer/annotate_types.py
```diff
@@ -104,6 +108,78 @@ class TypeAnnotator:
         },
     }
 
+    FUNCTION_TYPES = {
```
barakalon, line 111: What about functions like `COERCE`, in which the return type depends on the input type?

tobymao: it should be an optional lambda

barakalon: Ah, we're expecting "custom annotator functions" for this.

tobymao: why do you need FUNCTION_TYPES dict at all? why not just use annotators?

GeorgeSittas: That's what I'm working on, pushing in a bit.

tobymao:
anything with null is in general null

GeorgeSittas:
Outstanding items for this PR:

- [ ] Find a way to handle Nullable expressions (eg. columns, and/or expressions). Should probably use the `Nullable` type.

sqlglot/optimizer/annotate_types.py
```diff
@@ -141,36 +211,45 @@ def _maybe_annotate(self, expression):
...
 
     def _annotate_binary(self, expression):
         self._annotate_args(expression)
 
-        if isinstance(expression, (exp.Condition, exp.Predicate)):
+        left_type = expression.left.type
+        right_type = expression.right.type
+
+        if isinstance(expression, (exp.And, exp.Or)):
+            if (left_type, right_type) == (exp.DataType.Type.NULL, exp.DataType.Type.NULL):
```
tobymao, line 243: just do left_type == null and

sqlglot/optimizer/annotate_types.py
```diff
@@ -141,36 +211,45 @@ def _maybe_annotate(self, expression):
...
+                expression.type = exp.DataType.Type.NULL
+            elif exp.DataType.Type.NULL in (left_type, right_type):
+                expression.type = exp.DataType(this=exp.DataType.Type.NULLABLE, expressions=[exp.DataType.Type.BOOLEAN])
```
tobymao, line 246: DataType.build("NULLABLE", expressions=DataType.build("BOOLEAN")

sqlglot/tokens.py
```diff
@@ -579,6 +579,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "INT8": TokenType.BIGINT,
         "DECIMAL": TokenType.DECIMAL,
         "MAP": TokenType.MAP,
+        "NULLABLE": TokenType.NULLABLE,
```
tobymao, line 582: you can delete the override in clickhouse with this change

PR #536, opened 2022-10-05
cannot push down into multi use ctes

tobymao:


barakalon (approved):

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -154,3 +154,44 @@ SELECT
...
+), o AS (
+  SELECT a, b FROM m WHERE m.a = 2
+) SELECT
+    n.a,
+    n.b,
+    o.b
+FROM n
+FULL OUTER JOIN o ON n.a = o.a
+CROSS JOIN n AS n2
+WHERE o.b > 0 AND n.a = n2.a;
```
eakmanrq, line 171: This is a strange example since having the `n.a` = `o.a` in the where clause makes the query return nothing since n is null.

tobymao: didn't you give me this example? i just added some stuff to it

PR #538, opened 2022-10-06
expand order by last to ensure positional references are replaced with qualified columns

ginter:


tobymao (approved):

tobymao:
@barakalon should i revert this?

barakalon:
No. Let's chat about it first. I'm not sure what the desired behavior is.

barakalon:
Could be a good case for the new "execute" test mode.

barakalon:
duckdb is happy with it: https://github.com/tobymao/sqlglot/pull/542

PR #539, opened 2022-10-06
Fix simplify so that NULL AND FALSE -> FALSE, add NOT NULL -> NULL

GeorgeSittas:


GeorgeSittas:
By the way, `NOT NULL` can be simplified to `NULL`. Should we add this transformation too?

tobymao (approved):

PR #556, opened 2022-10-07
Snowflake table literal

prabhur:
Snowflake Table literal (https://docs.snowflake.com/en/sql-reference/literals-table.html).

Relevant commit: `71b8b8c7c1f6c37a99f972781143e08d48c32441`

Snowflake documentation is bit confusing. There are two concepts. 
1) Table Literal `(select ... FROM TABLE($MYVAR$) ... )`
2) Table Functions 
      -- Having the literal "TABLE " `(select ... FROM TABLE(RESULT_SCAN(LAST_QUERY_ID()))` )
      -- Using a "TABLE" function like `FLATTEN` directly (`select ... FROM mytable A, lateral flatten(...) f1` )

These changes only address 1). Intend to setup the code with 1) and then add 2).


tobymao:
this seems way too complicated. doesn’t it mostly work already?

prabhur:
> this seems way too complicated. doesn’t it mostly work already? 

I thought so too initially. The following test doesn't pass, for example. 

```
    def test_table_literal(self):
        self.validate_all(
            r"""SELECT * FROM TABLE('MYTABLE')""",
            write={
                "snowflake":r"""SELECT * FROM TABLE('MYTABLE')"""
            }
        )
```

I feel the changes are not very complicated. 
They attempt to address this syntax https://docs.snowflake.com/en/sql-reference/literals-table.html#syntax

tobymao:
it’s just because table was a reserved keyword

prabhur:
True. Additionally, in snowflake atleast, there are two ways in which it can be used. I have referenced the relevant documentation in the PR draft. 

What do you think? Happy to be corrected if there are simpler ways. I have not yet submitted the PR. Just wanted to get your thoughts before submitting. Additionally, the changes are scoped only to snowflake without touching the core modules.

tobymao:
make a fresh new pr with just the test cases and i’ll fix tonight

tobymao:
https://github.com/tobymao/sqlglot/commit/d88737dae29bd76064e6b99bd355d0b0283a4415

prabhur:
That was quick :-) Didn't look deep enough into the PLACEHOLDER expression!

Thanks!

PR #559, opened 2022-10-07
Fix alias builder for window expressions

### LINKED ISSUE #558: window frame generates incorrect sql when aliased
Parsing an empty window frame and then generating the SQL from it produces a reference to a non-existent named window frame.

```
In [1]: import sqlglot as sg

In [2]: sg.alias(sg.parse_one("LAG(x) over ()"), "a").sql()
Out[2]: 'LAG(x) OVER a'
```

When the window frame has a `PARTITION BY` or `ORDER BY` or both, the result is also incorrect but for a different reason

Only `PARTITION BY`:

```
In [3]: sg.alias(sg.parse_one("LAG(x) over (partition by y)"), "a").sql()
Out[3]: 'LAG(x) OVER (aPARTITION BY y)'
```

Only `ORDER BY`:

```
In [4]: sg.alias(sg.parse_one("LAG(x) over (order by y)"), "a").sql()
Out[4]: 'LAG(x) OVER (aORDER BY y)'
```

Both `PARTITION BY` and `ORDER BY`:

```
In [5]: sg.alias(sg.parse_one("LAG(x) over (partition by y order by z)"), "a").sql()
Out[5]: 'LAG(x) OVER (aPARTITION BY y ORDER BY z)'
```

The desired result is:

```
In [5]: sg.alias(sg.parse_one("LAG(x) over (partition by y order by z)"), "a").sql()
Out[5]: 'LAG(x) OVER (PARTITION BY y ORDER BY z) AS a'
```

GeorgeSittas:
Not sure if this is the correct approach for this, but the idea was to follow the parser's output:

```python
>>> sqlglot.parse_one("RANK() OVER (y PARTITION BY a, b ORDER BY x DESC) AS foo")
(ALIAS this: # <~~~~ For the AS <alias_name> form, we get an ALIAS node
  (WINDOW this:
    (ANONYMOUS this: RANK), partition_by:
    (COLUMN this:
      (IDENTIFIER this: a, quoted: False)),
    (COLUMN this:
      (IDENTIFIER this: b, quoted: False)), order:
    (ORDER expressions:
      (ORDERED this:
        (COLUMN this:
          (IDENTIFIER this: x, quoted: False)), desc: True, nulls_first: False)), alias:
    (IDENTIFIER this: y, quoted: False)), alias: # <~~~~ For a named_window alias, we get the IDENTIFIER node
  (IDENTIFIER this: foo, quoted: False))
```

I think that we only use the alias inside the parentheses (here: y) for the BigQuery dialect, so the tradeoff of this change is:
- We can't set this alias using the `alias_` builder function.
- We add the ALIAS node by default for all window expressions when using the builder, which seems more "standard".

Happy to follow another direction if this isn't the right approach. See [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls) for the `named_window` syntax.

Fixes #558

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3150,7 +3150,7 @@ def alias_(expression, alias, table=False, dialect=None, quoted=None, **opts):
     alias = to_identifier(alias, quoted=quoted)
     alias = TableAlias(this=alias) if table else alias
 
-    if "alias" in exp.arg_types:
+    if "alias" in exp.arg_types and not isinstance(exp, Window):
         exp = exp.copy()
         exp.set("alias", alias)
         return exp
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -153,6 +153,10 @@ class TestBigQuery(Validator):
         )
 
         self.validate_identity(
+            "SELECT item, purchases, LAST_VALUE(item) OVER (item_window ROWS BETWEEN 2 PRECEDING AND 2 FOLLOWING) AS most_popular FROM Produce WINDOW item_window AS (ORDER BY purchases)"
+        )
+
+        self.validate_identity(
             "SELECT LAST_VALUE(a IGNORE NULLS) OVER y FROM x WINDOW y AS (PARTITION BY CATEGORY)",
         )
 
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -1,6 +1,7 @@
 import unittest
 
 from sqlglot import (
+    alias,
     and_,
     condition,
     except_,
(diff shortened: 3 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -3150,7 +3150,7 @@ def alias_(expression, alias, table=False, dialect=None, quoted=None, **opts):
     alias = to_identifier(alias, quoted=quoted)
     alias = TableAlias(this=alias) if table else alias
 
-    if "alias" in exp.arg_types:
+    if "alias" in exp.arg_types and not isinstance(exp, Window):
```
barakalon, line 3153: What if we switch to an allowlist instead of a denylist?

GeorgeSittas: Deny list seemed like a better choice for this one because the builder for all other expression types (eg. lateral, values, etc) that support the "alias" argument behaves as expected here.

barakalon: K. Then let's add a comment here about why the `alias` arg of Window shouldn't be used.

barakalon (approved):

tobymao (approved):

sqlglot/expressions.py
GeorgeSittas: I'll make a note to add that as an extra in my next PR!

gauravdawar-e6:
@georgesittas I am new to SQLGLOT and I am trying to understand it. I am planning to commit to this repo too. I have one simple and fundamental questions. 

I was trying to check the syntax tree for RANK, DENSE_RANK, ROW_NUMBER basically window function. While analysing it, I found that it is actually mapped to Anonymous functions. Why we are mapping these RANK, DENSE_RANK, ROW_NUMBER as Anonymous function

PR #560, opened 2022-10-07
Add table hint and support it in optimizer

eakmanrq:
Prior to this change hints that contain tables would actually parse them into a column instead of a table. The initial change here is to have them properly parse into tables. The next change is in the optimizer where we tell the `merge_subqueries` rule that it can merge hints and then provided a function to do that merge. 

Hints that are not "JoinHints" are left untouched in this PR since they parse into columns and are handled correctly by default.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -766,7 +766,7 @@ class Hint(Expression):
     arg_types = {"expressions": True}
 
 
-class TableHint(Expression):
+class JoinHint(Expression):
     arg_types = {"this": True, "expressions": True}
 
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -766,6 +766,10 @@ class Hint(Expression):
     arg_types = {"expressions": True}
 
 
+class TableHint(Expression):
+    arg_types = {"this": True, "expressions": True}
+
+
 class Identifier(Expression):
     arg_types = {"this": True, "quoted": False}
 
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -188,8 +188,8 @@ def _merge_from(outer_scope, inner_scope, node_to_replace, alias):
         tables = table_hint.find_all(exp.Table)
         for table in tables:
             if table.alias_or_name == node_to_replace.alias_or_name:
-                new_value = new_subquery.this if isinstance(new_subquery, exp.Alias) else new_subquery
-                table.replace(new_value)
+                new_table = new_subquery.this if isinstance(new_subquery, exp.Alias) else new_subquery
+                table.set("this", exp.to_identifier(new_table.alias_or_name))
     outer_scope.remove_source(alias)
     outer_scope.add_source(new_subquery.alias_or_name, inner_scope.sources[new_subquery.alias_or_name])
 
@@ -299,7 +299,8 @@ def _merge_hints(outer_scope, inner_scope):
         return
     outer_scope_hint = outer_scope.expression.args.get("hint")
     if outer_scope_hint:
-        outer_scope_hint.args["expressions"].extend(inner_scope_hint.expressions)
+        for hint_expression in inner_scope_hint.expressions:
+            outer_scope_hint.append("expressions", hint_expression)
     else:
         outer_scope.expression.set("hint", inner_scope_hint)
 
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -184,8 +184,8 @@ def _merge_from(outer_scope, inner_scope, node_to_replace, alias):
     """
     new_subquery = inner_scope.expression.args.get("from").expressions[0]
     node_to_replace.replace(new_subquery)
-    for table_hint in outer_scope.table_hints:
-        tables = table_hint.find_all(exp.Table)
+    for join_hint in outer_scope.join_hints:
+        tables = join_hint.find_all(exp.Table)
         for table in tables:
             if table.alias_or_name == node_to_replace.alias_or_name:
                 new_table = new_subquery.this if isinstance(new_subquery, exp.Alias) else new_subquery
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -44,6 +44,7 @@ UNMERGABLE_ARGS = set(exp.Select.arg_types) - {
     "joins",
     "where",
     "order",
+    "hint",
 }
 
 
@@ -82,6 +83,7 @@ def merge_ctes(expression, leave_tables_isolated=False):
             _merge_joins(outer_scope, inner_scope, from_or_join)
             _merge_where(outer_scope, inner_scope, from_or_join)
             _merge_order(outer_scope, inner_scope)
+            _merge_hints(outer_scope, inner_scope)
             _pop_cte(inner_scope)
     return expression
 
@@ -101,6 +103,7 @@ def merge_derived_tables(expression, leave_tables_isolated=False):
                 _merge_joins(outer_scope, inner_scope, from_or_join)
                 _merge_where(outer_scope, inner_scope, from_or_join)
                 _merge_order(outer_scope, inner_scope)
+                _merge_hints(outer_scope, inner_scope)
     return expression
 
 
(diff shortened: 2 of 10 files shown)


tobymao (approved):

sqlglot/parser.py
```diff
@@ -431,6 +431,14 @@ class Parser:
         "TRIM": lambda self: self._parse_trim(),
         "CAST": lambda self: self._parse_cast(self.STRICT_CAST),
         "TRY_CAST": lambda self: self._parse_cast(False),
+        "BROADCAST": lambda self: self._parse_table_hint("BROADCAST"),
```
barakalon, line 434: These are all spark specific, yes? Can we move them to that dialect?

tests/test_optimizer.py
```diff
@@ -97,7 +97,7 @@ def check_file(self, file, func, pretty=False, execute=False, **kwargs):
     def test_optimize(self):
         schema = {
             "x": {"a": "INT", "b": "INT"},
-            "y": {"a": "INT", "b": "INT"},
+            "y": {"b": "INT", "c": "INT"},
```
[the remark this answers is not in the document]
barakalon: ah nice, I didn't see this

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -215,3 +215,25 @@ JOIN "n" AS "n2"
   ON "n"."a" = "n2"."a"
 WHERE
   "o"."b" > 0;
+
+# title: Broadcast hint
+# dialect: spark
+WITH m AS (SELECT x.a, x.b FROM x), n AS (SELECT y.b, y.c FROM y), joined as (SELECT /*+ BROADCAST(n) */ m.a, n.c FROM m JOIN n ON m.b = n.b) SELECT joined.a, joined.c FROM joined;
```
barakalon, line 221: Can you use "pretty" SQL? These complex, single line queries are hard to read.

sqlglot/expressions.py
```diff
@@ -766,6 +766,10 @@ class Hint(Expression):
     arg_types = {"expressions": True}
 
 
+class TableHint(Expression):
```
barakalon, line 769: Spark calls this "join hints".

And I imagine other dialects have totally different types of hints.

I'm worried we're making the core expression too Spark specific. Can we parse these expressions into `Hint`, but have the Spark dialect handle parsing the expressions as `Table`s?

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -181,6 +184,12 @@ def _merge_from(outer_scope, inner_scope, node_to_replace, alias):
     """
     new_subquery = inner_scope.expression.args.get("from").expressions[0]
     node_to_replace.replace(new_subquery)
+    for table_hint in outer_scope.table_hints:
+        tables = table_hint.find_all(exp.Table)
+        for table in tables:
+            if table.alias_or_name == node_to_replace.alias_or_name:
+                new_value = new_subquery.this if isinstance(new_subquery, exp.Alias) else new_subquery
+                table.replace(new_value)
```
barakalon, line 192: Make sure to copy:
```python
table.replace(new_value.copy())
```

Or, just set the name
```python
table.set("this", exp.to_identifier(new_name)
```

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -181,6 +184,12 @@ def _merge_from(outer_scope, inner_scope, node_to_replace, alias):
     """
     new_subquery = inner_scope.expression.args.get("from").expressions[0]
     node_to_replace.replace(new_subquery)
+    for table_hint in outer_scope.table_hints:
+        tables = table_hint.find_all(exp.Table)
+        for table in tables:
+            if table.alias_or_name == node_to_replace.alias_or_name:
+                new_value = new_subquery.this if isinstance(new_subquery, exp.Alias) else new_subquery
```
barakalon, line 191: We only want the _alias_ of the new subquery, right? Otherwise we might end up shoving an entire derived table in here. Might want to test that.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -284,6 +293,17 @@ def _merge_order(outer_scope, inner_scope):
     outer_scope.expression.set("order", inner_scope.expression.args.get("order"))
 
 
+def _merge_hints(outer_scope, inner_scope):
+    inner_scope_hint = inner_scope.expression.args.get("hint")
+    if not inner_scope_hint:
+        return
+    outer_scope_hint = outer_scope.expression.args.get("hint")
+    if outer_scope_hint:
+        outer_scope_hint.args["expressions"].extend(inner_scope_hint.expressions)
```
barakalon, line 302: Patching expressions directly won't set `parent` on the child nodes.

```
outer_scope_hint.set("expressions", outer_scope_hint.expressions + inner_scope_hint.expressions)
```

sqlglot/optimizer/merge_subqueries.py
tobymao: let's go with the set

sqlglot/optimizer/merge_subqueries.py
tobymao: ah i didn't catch this, you should use

outer_scope.append("expressions", ...)

sqlglot/expressions.py
eakmanrq: Good to rename to `JoinHint` if we continue with this approach. 👍 

So `Hint` doesn't actually represent a single hint... it represents a collection of hints. Those individual hints prior to this PR were Anonymous functions and then this PR allows us to differentiate the anonymous function ones (which are handled fine since they contain columns) and the join hint ones which behave differently since they contain tables. 

In my mind if we know at parsing time that the hint being expressed is one of these two types then we should put it in an expression that represents that. I did though stop short of implementing though that completely since the anonymous hints aren't causing any issues currently.

barakalon: Ah, got it. 

Is it possible to have a hint with tables and columns? The Spark ones don't, but maybe some other dialect does? Or maybe Spark will _eventually_?

sqlglot/optimizer/merge_subqueries.py
barakalon: ah woah, I didn't know we added `append`. Nice!

eakmanrq: We would need an `outer_scope.extend("expressions", ...)` right? I can add that if you want or I could also go with the set approach.

barakalon: Or write a for loop 😛

tobymao: just do a for loop append?

sqlglot/expressions.py
eakmanrq: Maybe but just some googling suggests we don't have another example to compare against. Generally my preference here is to model the solution for the only given example we have and then iterate as new patterns emerge.

sqlglot/optimizer/merge_subqueries.py
eakmanrq: lol fine

sqlglot/expressions.py
tobymao: agreed

eakmanrq: `coalesce` is a PartitionHint which this PR leaves as an Anonymous function since currently they are handled fine in the codebase as one.

sqlglot/optimizer/merge_subqueries.py
eakmanrq: There isn't an example currently in the tests (and that is all the tests) where `new_subquery` comes back as anything but a table or an alias. Can you describe more the case where you think it might actually return a subquery or am I misunderstanding what you are saying?

barakalon: Try this:
```sql
SELECT 
  x.a, 
  x.c 
FROM (
  SELECT 
    x.a, 
    x.c 
  FROM (
    SELECT 
      x.a, 
      COUNT(1) AS c 
    FROM x 
    GROUP BY x.a
  ) AS x
) AS x
```

eakmanrq: Thanks that did it and I confirmed the test I added covers this case.

sqlglot/expressions.py
eakmanrq: I'm going to stick with the current approach but renamed everything to `JoinHint`. To me this is a good first step and we can generalize as we get more examples (if we ever do).

barakalon (approved):

PR #563, opened 2022-10-09
Improve window support in optimizer

eakmanrq:
Prior to this change the optimizer would push/merge window functions as if they were regular column expressions and therefore you would see them end up in where clauses. The goal of my change was to still allow merge of window expressions between select expressions but disallow them in the where clause.


sqlglot/optimizer/merge_subqueries.py
```diff
@@ -119,6 +119,19 @@ def _mergeable(outer_scope, inner_select, leave_tables_isolated, from_or_join):
     Returns:
         bool: True if can be merged
     """
+
+    def _get_window_expressions_in_where_clause():
```
eakmanrq, line 123: I know there must be a cleaner way to do this...

tobymao: what exactly are you trying to do here?

eakmanrq: Get all the window expressions in the inner select scope. Then check if any of those columns are referenced by the outer scope but only in the outer scope's where clause (I guess I should also check join clause too). If so then block merge otherwise allow the merge.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -119,6 +119,19 @@ def _mergeable(outer_scope, inner_select, leave_tables_isolated, from_or_join):
...
+        window_expressions = list(inner_select.find_all(exp.Window))
```
tobymao, line 124: you don't need to call list here

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -119,6 +119,19 @@ def _mergeable(outer_scope, inner_select, leave_tables_isolated, from_or_join):
...
+        window_alias_names = [window.parent.alias_or_name for window in window_expressions]
```
tobymao, line 125: you probably want to make this a set? {window.parent.alias_or_name for w in window}

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -119,6 +119,19 @@ def _mergeable(outer_scope, inner_select, leave_tables_isolated, from_or_join):
...
+        inner_select_name = inner_select.parent.alias_or_name
+        where_clauses_columns = [column for column in outer_scope.columns if column.find_ancestor(exp.Where)]
```
tobymao, line 127: you're checking every column is in a where? why not find the where first and then find the columns?

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -152,9 +152,11 @@ def nodes_for_predicate(predicate, sources, scope_ref_count):
                 return {}
             nodes[table] = node
         elif isinstance(node, exp.Select) and len(tables) == 1:
+            # We can't push down window expressions
+            has_window_expression = bool([select for select in node.selects if select.find(exp.Window)])
```
tobymao, line 156: any(select.find(exp.Window) for select in node.selects)

sqlglot/optimizer/merge_subqueries.py
eakmanrq: Actually I'm wondering if I should instead check if the column is used anywhere besides a select expression. That would cover other cases like a window in a group by or join.

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -276,3 +276,54 @@ SELECT /*+ COALESCE(3),
 FROM `x` AS `x`
 JOIN `y` AS `y`
   ON `x`.`b` = `y`.`b`;
+
```
barakalon, line 279: Do we need these tests in both fixture files?

Redundant tests can be a pain to fix.

sqlglot/optimizer/merge_subqueries.py
barakalon: Yeah I think it's allowed in some clauses (selections... maybe order by?) but not others.

eakmanrq: Yeah I tested a few cases and explicitly excluded them in my latest commit (also updated the tests).

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -38,7 +38,7 @@ def merge_subqueries(expression, leave_tables_isolated=False):
 
 
 # If a derived table has these Select args, it can't be merged
-UNMERGABLE_ARGS = set(exp.Select.arg_types) - {
+unmergeable_ARGS = set(exp.Select.arg_types) - {
```
tobymao, line 41: why'd you chaneg this?

tobymao (approved):

sqlglot/optimizer/merge_subqueries.py
eakmanrq: Looks like a bad find/replace. Will fix.

PR #568, opened 2022-10-10
PySpark DataFrame API Support

eakmanrq:
This contains a drop-in replacement for the PySpark DataFrame API (see README for details).


dev-requirements.txt
```diff
@@ -0,0 +1,2 @@
+-r requirements.txt
```
tobymao, line 1: requirements.txt is dev-requirements, add the rest into here and delete the old file

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        DateTimeLiteral,
+        DecimalLiteral,
+        Literals,
+        Primitives,
+    )
+    from sqlglot.dataframe.sql.window import WindowSpec
+
+
+class Column:
+    def __init__(self, expression: t.Union[t.Any, exp.Expression]):
```
tobymao, line 25: should you use maybe_parse?

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        return isinstance(self.expression, exp.Alias)
+
+    @property
+    def is_column(self):
+        return isinstance(self.expression, exp.Column)
+
+    @property
+    def column_expression(self) -> exp.Column:
+        if self.is_alias:
+            return self.expression.args["this"]
```
tobymao, line 195: expression.this

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        return self.expression
+
+    @property
+    def alias_or_name(self):
```
tobymao, line 199: missing return type

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        if isinstance(self.expression, exp.Null):
+            return "NULL"
+        if isinstance(self.expression.args.get("this"), exp.Star):
```
tobymao, line 202: replace all places where you do args["this"] to expression.this

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+    def set_table_name(self, table_name: str, copy=False) -> Column:
+        expression = self.expression.copy() if copy else self.expression
+        expression.set("table", exp.to_identifier(table_name))
+        return Column(expression)
+
+    def sql(self, **kwargs) -> Column:
+        return self.expression.sql(dialect="spark", **kwargs)
+
+    def alias(self, name: str) -> Column:
+        new_expression = exp.Alias(alias=exp.to_identifier(name, quoted=True), this=self.column_expression)
```
tobymao, line 228: can you use exp.alias_ instead?

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        new_column = self.copy()
+        new_column.expression.args["ifs"].extend(column_with_if.expression.args["ifs"])
+        return new_column
+
+    def otherwise(self, value: t.Any) -> Column:
+        from sqlglot.dataframe.sql.functions import lit
+
+        true_value = value if isinstance(value, Column) else lit(value)
+        new_column = self.copy()
+        new_column.expression.args["default"] = true_value.column_expression
```
tobymao, line 266: never set args directly, use set

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        lower_bound_exp = self._lit(lowerBound) if not isinstance(lowerBound, Column) else lowerBound
+        upper_bound_exp = self._lit(upperBound) if not isinstance(upperBound, Column) else upperBound
+        return Column(
+            exp.Between(this=self.column_expression, low=lower_bound_exp.expression, high=upper_bound_exp.expression)
+        )
+
+    def over(self, window: WindowSpec) -> Column:
+        window_args = window.expression.args
+        window_args["this"] = self.column_expression
+        return Column(exp.Window(**window_args))
```
tobymao, line 330: can you add all places where you instantiate an expression to have kwargs, like this=

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+    def sparkSession(self):
+        return self.spark
+
+    @property
+    def write(self):
+        return DataFrameWriter(self)
+
+    @property
+    def latest_cte_name(self) -> str:
+        if len(self.expression.ctes) == 0:
```
tobymao, line 68: if not self.expression.ctes

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+    @property
+    def na(self) -> DataFrameNaFunctions:
+        return DataFrameNaFunctions(self)
+
+    def _replace_cte_names_with_hashes(self, expression: exp.Select):
+        expression = expression.copy()
+        ctes = expression.ctes
+        replacement_mapping = {}
+        for cte in ctes:
+            old_name_id = cte.args["alias"].this
```
tobymao, line 99: can you just do cte.alias?

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+            new_hashed_id = exp.to_identifier(
+                self._create_hash_from_expression(cte.this), quoted=old_name_id.args["quoted"]
+            )
+            replacement_mapping[old_name_id] = new_hashed_id
```
tobymao, line 103: there's a helper method exp.alias_(table=True) you can use

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+    def _convert_leaf_to_cte(self, sequence_id: t.Optional[str] = None) -> DataFrame:
+        df = self._resolve_pending_hints()
+        sequence_id = sequence_id or df.sequence_id
+        expression = df.expression.copy()
+        cte_expression, cte_name = df._create_cte_from_expression(expression=expression, sequence_id=sequence_id)
+        new_expression = exp.Select()
+        new_expression = df._add_ctes_to_expression(new_expression, expression.ctes + [cte_expression])
+        sel_columns = df._get_outer_select_columns(cte_expression)
+        star_columns = [col for col in sel_columns if isinstance(col.expression.this, exp.Star)]
+        if len(star_columns) > 0:
```
tobymao, line 149: if star_columns

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+                                df.pending_hints.remove(hint)
+                                break
+                hint_expression.args.get("expressions").append(hint)
+        if hint_expression.expressions:
+            expression.set("hint", hint_expression)
+        return df
+
+    def _hint(self, hint_name: str, args: t.List[Column]) -> DataFrame:
+        hint_name = hint_name.upper()
+        if hint_name in {
```
tobymao, line 188: make this a constant?

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+            if optimize:
+                select_expression = optimize_func(select_expression)
+            select_expression = df._replace_cte_names_with_hashes(select_expression)
+
+            # temporary hack to remove extra CTE that was replaced with table until removing unused CTEs is added
+            # to optimizer
+            filtered_ctes = [
+                cte
+                for cte in select_expression.ctes
+                if cte.alias_or_name not in {x.alias_or_name for x in replacement_mapping.values()}
```
tobymao, line 283: replaceme_mapping double for loop?

sqlglot/dataframe/README.md
```diff
@@ -0,0 +1,224 @@
...
+spark = SparkSession()
+
+df = (
+    spark
+    .table('employee')
+    .groupBy(F.col("age"))
+    .agg(F.countDistinct(F.col("employee_id")).alias("num_employees")) 
+)
+
+print(df.sql())  # Spark will be the dialect used by default
```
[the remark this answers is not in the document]
eakmanrq: It is pretty by default: https://github.com/tobymao/sqlglot/blob/5f2b954a9e940676cdb8512a707a95d8b40c23a3/sqlglot/dataframe/sql/dataframe.py#L318

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+            ambiguous_cols = [col for col in cols if not col.column_expression.table]
+            if ambiguous_cols:
+                join_table_identifiers = [x.this for x in get_tables_from_expression_with_join(self.expression)]
+                cte_names_in_join = [x.this for x in join_table_identifiers]
+                for ambiguous_col in ambiguous_cols:
+                    ctes_with_column = [
+                        cte
+                        for cte in self.expression.ctes
+                        if cte.alias_or_name in cte_names_in_join
+                        and ambiguous_col.alias_or_name in cte.args["this"].named_selects
```
tobymao, line 338: cte.this

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,741 @@
...
+
+        old_values = None
+        subset = ensure_list(subset)
+        new_df = self.copy()
+        all_columns = self._get_outer_select_columns(new_df.expression)
+        all_column_mapping = {column.alias_or_name: column for column in all_columns}
+
+        columns = self._ensure_and_normalize_cols(subset) if subset else all_columns
+        if isinstance(to_replace, dict):
+            old_values = list(to_replace.keys())
```
tobymao, line 617: list(to_replace) don',t need keys

sqlglot/dataframe/sql/functions.py
```diff
@@ -0,0 +1,1251 @@
...
+    from sqlglot.dataframe.sql.dataframe import DataFrame
+
+
+def col(column_name: t.Union[ColumnOrName, t.Any]) -> Column:
+    return Column(column_name)
+
+
+def lit(value: t.Optional[t.Any] = None) -> Column:
+    if isinstance(value, str):
+        return Column(glotexp.Literal(this=str(value), is_string=True))
```
tobymao, line 21: exp.Literal.string(...)

sqlglot/dataframe/sql/functions.py
```diff
@@ -0,0 +1,1251 @@
...
+
+
+def first(col: ColumnOrName, ignorenulls: bool = None) -> Column:
+    if ignorenulls is not None:
+        return Column.invoke_anonymous_function(col, "FIRST", ignorenulls)
+    return Column.invoke_anonymous_function(col, "FIRST")
+
+
+def grouping_id(*cols: ColumnOrName) -> Column:
+    if len(cols) == 0:
```
tobymao, line 373: not cols

sqlglot/dataframe/sql/session.py
```diff
@@ -0,0 +1,145 @@
...
+        if schema is not None and (
+            not isinstance(schema, (StructType, str, list))
+            or (isinstance(schema, list) and not isinstance(schema[0], str))
+        ):
+            raise NotImplementedError("Only schema of either list or string of list supported")
+
+        if schema is not None:
+            column_mapping = get_column_mapping_from_schema_input(schema)
+        elif isinstance(data[0], dict):
+            column_mapping = {col_name.strip(): None for col_name in data[0].keys()}
```
tobymao, line 61: you don't need keys

sqlglot/expressions.py
```diff
@@ -3131,6 +3134,14 @@ def to_identifier(alias, quoted=None):
     return identifier
 
 
+def to_table(sql_path, pivots=None):
```
barakalon, line 3137: I think this is the first time the concept of a "sql_path" has entered this file.

Should this function live here or somewhere else?

sqlglot/helper.py
```diff
@@ -24,7 +27,17 @@ def list_get(arr, index):
 def ensure_list(value):
     if value is None:
         return []
-    return value if isinstance(value, (list, tuple, set)) else [value]
+    return [*value] if isinstance(value, (list, tuple, set)) else [value]
```
barakalon, line 30: I think this introduces loads of copy operations. throughout sqlglot. 

Based on how we use `ensure_list`, I think that's _safe_, but I wonder how it effects performance?

sqlglot/dataframe/sql/dataframe.py
eakmanrq: I need the id value and it looks like `cte.alias` returns a string

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -37,12 +41,16 @@ def pushdown_projections(expression):
...
-            left, right = scope.union_scopes
-            referenced_columns[left] = parent_selections
-            referenced_columns[right] = parent_selections
+            left_union, right_union = scope.union_scopes
+            referenced_columns[left_union] = parent_selections
+            referenced_columns[right_union] = parent_selections
 
-        if isinstance(scope.expression, exp.Select):
-            _remove_unused_selections(scope, parent_selections)
+        if isinstance(scope.expression, exp.Select) and scope != right_union:
```
barakalon, line 48: `right_union` can be undefined

sqlglot/dataframe/sql/window.py
```diff
@@ -0,0 +1,116 @@
...
+    def sql(self, **kwargs) -> str:
+        return self.expression.sql(dialect="spark", **kwargs)
+
+    def partitionBy(self, *cols: t.Union[ColumnOrName, t.List[ColumnOrName]]) -> WindowSpec:
+        from sqlglot.dataframe.sql.column import Column
+
+        cols = flatten(cols) if isinstance(cols[0], (list, set, tuple)) else cols
+        expressions = [Column.ensure_col(x).expression for x in cols]
+        window_spec = self.copy()
+        if window_spec.expression.args.get("partition_by") is None:
```
tobymao, line 59: probabyl can just do not get partition_by

sqlglot/dataframe/sql/window.py
```diff
@@ -0,0 +1,116 @@
...
+            window_spec.expression.args["partition_by"] = expressions
```
tobymao, line 60: don't ever set stuff manually, do set or append, here with line 62

sqlglot/dataframe/sql/window.py
```diff
@@ -0,0 +1,116 @@
...
+        return window_spec
+
+    def orderBy(self, *cols: t.Union[ColumnOrName, t.List[ColumnOrName]]) -> WindowSpec:
+        from sqlglot.dataframe.sql.column import Column
+
+        cols = flatten(cols) if isinstance(cols[0], (list, set, tuple)) else cols
+        expressions = [Column.ensure_col(x).expression for x in cols]
+        window_spec = self.copy()
+        if window_spec.expression.args.get("order") is None:
+            window_spec.expression.args["order"] = exp.Order(expressions=[])
```
tobymao, line 72: same here

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -37,12 +41,16 @@ def pushdown_projections(expression):
...
+            removed_indexes = _remove_unused_selections(scope, parent_selections)
+            # The left union is used for column names to select and if we remove columns from the left
+            # we need to also remove those same columns in the right that were at the same position
+            if scope == left_union:
```
barakalon, line 52: When can this happen?

sqlglot/dialects/spark.py
```diff
@@ -18,12 +14,6 @@ def _create_sql(self, e):
     return create_with_partitions_sql(self, e)
 
 
-def _map_sql(self, expression):
```
tobymao, line 21: why did you remove this?

sqlglot/expressions.py
```diff
@@ -1286,6 +1287,8 @@ def group_by(self, *expressions, append=True, dialect=None, copy=True, **opts):
         Returns:
             Select: the modified expression.
         """
+        if not expressions:
```
tobymao, line 1290: how come?

sqlglot/expressions.py
tobymao, line 3137: docstring

sqlglot/schema.py
```diff
@@ -1,13 +1,24 @@
...
+from sqlglot import expressions as exp
 from sqlglot.errors import OptimizeError
-from sqlglot.helper import csv_reader
+from sqlglot.helper import csv_reader, ensure_table
 
 
 class Schema(abc.ABC):
     """Abstract base class for database schemas"""
 
+    @abc.abstractmethod
```
barakalon, line 12: Adding an abstractmethod makes this a backwards incompatible change.

I also think of `Schema` as being owned by the optimizer - i.e., it defines the interface that the optimizer depends on.

I think we want a second interface - `MutableSchema`.

sqlglot/helper.py
tobymao, line 30: what's this?

sqlglot/dataframe/sql/dataframe.py
eakmanrq: Yeah this is part of the hacky section that will be replaced once we improve the optimizer to remove unused CTEs.

sqlglot/schema.py
```diff
@@ -121,6 +171,24 @@ def ensure_schema(schema):
     return MappingSchema(schema)
 
 
+def ensure_column_mapping(mapping):
+    from sqlglot.dataframe.sql import types as df_types
```
barakalon, line 175: This dependency is a smell IMO.

Should the dataframe API have its own schema?

barakalon (commented):
wowow. This is crazy.

sqlglot/optimizer/pushdown_projections.py
barakalon: I don't think scope defines an `__eq__` method?

tests/dialects/test_duckdb.py
```diff
@@ -92,7 +92,7 @@ def test_duckdb(self):
                 "duckdb": "STR_SPLIT(x, 'a')",
                 "presto": "SPLIT(x, 'a')",
                 "hive": "SPLIT(x, CONCAT('\\\\Q', 'a'))",
-                "spark": "SPLIT(x, CONCAT('\\\\Q', 'a'))",
+                "spark": "SPLIT(x, 'a')",
```
tobymao, line 95: you tested this right?

tests/dialects/test_presto.py
```diff
@@ -40,7 +40,7 @@ def test_cast(self):
                 "duckdb": "CAST(MAP(LIST_VALUE(1), LIST_VALUE(1)) AS MAP<INT, INT>)",
                 "presto": "CAST(MAP(ARRAY[1], ARRAY[1]) AS MAP(INTEGER, INTEGER))",
                 "hive": "CAST(MAP(1, 1) AS MAP<INT, INT>)",
-                "spark": "CAST(MAP_FROM_ARRAYS(ARRAY(1), ARRAY(1)) AS MAP<INT, INT>)",
+                "spark": "CAST(MAP(1, 1) AS MAP<INT, INT>)",
```
tobymao, line 43: why did you change this?

sqlglot/dataframe/sql/_typing.pyi
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
eakmanrq: @tobymao and I are discussing now what to do about the types. The reason for including types was that the PySpark DataFrame API has type support and therefore provides really good autocomplete and suggestions to users. I didn't want to have this replacement provide a worse experience for them. I'm open though to switching to docstring types if that is the long term vision for type support in sqlglot.

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+    @classmethod
+    def ensure_col(cls, value: t.Union[ColumnOrName, int, float]):
+        return cls(value)
+
+    @classmethod
+    def ensure_cols(cls, args: t.List[t.Union[int, str, Column]]) -> t.List[Column]:
+        return [cls.ensure_col(x) if not isinstance(x, Column) else x for x in args]
+
+    @classmethod
+    def _lit(cls, value: t.Optional[t.Any] = None) -> Column:
```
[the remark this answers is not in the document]
eakmanrq: `lit` is from the PySpark DataFrame API and it stands for `Literal`.

sqlglot/__init__.py
```diff
@@ -21,12 +21,15 @@
...
 from sqlglot.generator import Generator
 from sqlglot.parser import Parser
+from sqlglot.schema import MappingSchema
 from sqlglot.tokens import Tokenizer, TokenType
 
 __version__ = "6.3.1"
 
 pretty = False
 
+schema = MappingSchema()
```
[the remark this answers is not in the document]
tobymao: i think so, we can discuss later

sqlglot/dialects/spark.py
eakmanrq: So when I started added in the map functions from the PySpark DataFrame API I noticed that we weren't actually using the map function from Spark but rather creating maps using `MAP_FROM_ARRAYS` even if there weren't arrays. Since Spark had direct support for creating maps like one would expect I removed this workaround.

tobymao: map in spark is Map(a, b, c, d)
in order to do arrays you need to do MAP_FROM_ARRAYS, are you sure this is correct to remove?

sqlglot/dataframe/sql/column.py
izeigerman, line 25: [Nitpick] Seems like we expect very specific types in the constructor. Should we enumerate them in the Union instead of just using Any?

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+
+    @operation(Operation.FROM)
+    def dropna(
+        self,
+        how: str = "any",
+        thresh: t.Optional[int] = None,
+        subset: t.Optional[t.Union[str, t.Tuple[str, ...], t.List[str]]] = None,
+    ) -> DataFrame:
+        if self.expression.ctes[-1].find(exp.Star) is not None:
+            raise RuntimeError("Cannot use `dropna` when a * expression is used")
```
izeigerman, line 524: [Nitpick] we should consider using more specialized error classes

sqlglot/schema.py
```diff
@@ -149,6 +217,35 @@ def _nested_get(d, *path):
...
+        return
+    if len(keys) == 1:
+        d[keys[0]] = value
+        return
+    subd = d
+    for key in keys[:-1]:
+        if key not in subd:
+            subd = subd.setdefault(key, {})
+        else:
+            subd = subd[key]
```
izeigerman, line 244: what if the target key is not a dict?

sqlglot/schema.py
```diff
@@ -49,37 +60,49 @@ class MappingSchema(Schema):
...
-        elif depth == 3:  # {db: {table: {col: type}}}
-            self.supported_table_args = ("db", "this")
-        elif depth == 4:  # {catalog: {db: {table: {col: type}}}}
-            self.supported_table_args = ("catalog", "db", "this")
-        else:
-            raise OptimizeError(f"Invalid schema shape. Depth: {depth}")
-
-        self.forbidden_args = {"catalog", "db", "this"} - set(self.supported_table_args)
+        self.supported_table_args = []
+        self.forbidden_args = []
```
izeigerman, line 69: Shall we call this `forbidden_table_args` to clarify and to be consistent with `supported_table_args`?

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+    ) -> Column:
+        column = cls.ensure_col(column) if column is not None else None
+        new_expression = (
+            callable_expression(this=column.column_expression, **kwargs)
+            if column is not None
+            else callable_expression(**kwargs)
+        )
+        return Column(new_expression)
+
+    def binary_op(self, clazz: t.Callable, other: Column, **kwargs) -> Column:
```
tobymao, line 175: replace all clazz with klass

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+        return Column(new_expression)
+
+    @classmethod
+    def invoke_expression_over_column(
+        cls, column: t.Union[ColumnOrName, None], callable_expression: t.Callable, **kwargs
+    ) -> Column:
+        column = cls.ensure_col(column) if column is not None else None
+        new_expression = (
+            callable_expression(this=column.column_expression, **kwargs)
+            if column is not None
```
tobymao, line 170: for all of these i prefer the inverse which is

if column is None else ... you have a couple of these calls

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+            datetime_literal = exp.Literal(this=value.strftime("%Y-%m-%d %H:%M:%S"), is_string=True)
+            return cls(
+                exp.Column(
+                    this=exp.StrToTime(
+                        this=datetime_literal, format=exp.Literal(this="YYYY-MM-DD HH:MM:SS", is_string=True)
+                    )
+                )
+            )
+        if isinstance(value, datetime.date):
+            date_literal = exp.Literal(this=value.strftime("%Y-%m-%d"), is_string=True)
```
tobymao, line 149: exp.Literal.string

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+            return cls(
+                exp.Column(this=exp.StrToDate(this=date_literal, format=exp.Literal(this="YYYY-MM-DD", is_string=True)))
```
tobymao, line 151: same

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @property
+    def column_expression(self) -> exp.Column:
+        if self.is_alias:
+            return self.expression.this
+        return self.expression
+
+    @property
+    def alias_or_name(self) -> str:
+        if isinstance(self.expression, exp.Null):
```
tobymao, line 200: can you just do this in base sqlglot?

sqlglot/dataframe/sql/const.py
```diff
@@ -0,0 +1,10 @@
+JOIN_HINTS = {
```
tobymao, line 1: can we call it constants.py

tobymao: was a separate file necessary? could have just put it an the top where this was referenced

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+import functools
+import typing as t
+import zlib
+from copy import copy
+
+import sqlglot
+from sqlglot import expressions as exp
+from sqlglot.dataframe.sql import functions as F
+from sqlglot.dataframe.sql.column import Column
+from sqlglot.dataframe.sql import const
```
tobymao, line 12: constants import as c

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+class DataFrame:
+    def __init__(
+        self,
+        spark: SparkSession,
+        expression: exp.Select,
+        branch_id: str = None,
+        sequence_id: str = None,
+        last_op: t.Optional[Operation] = Operation.INIT,
+        pending_hints: t.List[exp.Expression] = None,
+        output_expression_container: OutputExpressionContainer = exp.Select(),
```
tobymao, line 38: isn't it dangerous to do this because python evaluates Select() once, it should probably be None and then you do container = container or exp.Select()

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+        return expression
+
+    @classmethod
+    def _get_outer_select_columns(cls, item: t.Union[exp.Expression, DataFrame]) -> t.List[Column]:
+        expression = item.expression if isinstance(item, DataFrame) else item
+        return [Column(x) for x in dict.fromkeys(expression.find(exp.Select).args.get("expressions", []))]
+
+    @classmethod
+    def _create_hash_from_expression(cls, expression: exp.Select):
+        value = expression.sql(dialect="spark").encode("utf-8")
```
tobymao, line 234: does this need to be consistent? we also have _normalize_args

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+                expression.set("expression", select_without_ctes)
+                if select_expression.ctes:
+                    expression.set("with", exp.With(expressions=select_expression.ctes))
+            elif expression_type == exp.Select:
+                expression = select_expression
+            else:
+                raise ValueError(f"Invalid expression type: {expression_type}")
+            output_expressions.append(expression)
+
+        return [expression.sql(**{"dialect": dialect, "pretty": True, **kwargs}) for expression in output_expressions]
```
tobymao, line 307: pretty default

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+
+    def copy(self, **kwargs) -> DataFrame:
+        kwargs = {**{k: copy(v) for k, v in vars(self).copy().items()}, **kwargs}
```
tobymao, line 310: i'm not a big fan of this pattern, i prfer things inline

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+        return self._set_operation(exp.Intersect, other, False)
+
+    @operation(Operation.FROM)
+    def exceptAll(self, other: DataFrame) -> DataFrame:
+        return self._set_operation(exp.Except, other, False)
+
+    @operation(Operation.SELECT)
+    def distinct(self) -> DataFrame:
+        expression = self.expression.copy()
+        expression.set("distinct", exp.Distinct())
```
tobymao, line 500: make this a sqlglot base method

sqlglot/dataframe/sql/readwriter.py
```diff
@@ -0,0 +1,76 @@
...
+
+class DataFrameWriter:
+    def __init__(self, df: DataFrame, spark: SparkSession = None, mode: str = None, by_name: bool = False):
+        self._df = df
+        self._spark = spark or df.spark
+        self._mode = mode
+        self._by_name = by_name
+
+    def copy(self, **kwargs) -> DataFrameWriter:
+        kwargs = {**{k: copy(v) for k, v in vars(self).copy().items()}, **kwargs}
```
tobymao, line 33: second time i've seen this, let's make a helper

sqlglot/helper.py
tobymao: agree, we shoudln't do this as it will hurt perf

sqlglot/schema.py
tobymao: does this need to be an abc?

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+    from sqlglot.dataframe.sql.window import WindowSpec
+
+
+class Column:
+    def __init__(self, expression: t.Union[t.Any, exp.Expression]):
+        if isinstance(expression, str):
+            expression = sqlglot.parse_one(expression, read="spark")
+        elif isinstance(expression, Column):
+            expression = expression.expression
+        self.expression = expression
```
izeigerman, line 30: Do we need to ensure the validity of the passed expression in any way? I assume not every expression can be `Column`.

sqlglot/dataframe/sql/session.py
```diff
@@ -0,0 +1,145 @@
...
+        self,
+        data: t.Iterable[t.Union[t.Dict[str, t.Any], t.Iterable[t.Any]]],
+        schema: t.Optional[SchemaInput] = None,
+        samplingRatio: t.Optional[float] = None,
+        verifySchema: bool = False,
+    ) -> DataFrame:
+        from sqlglot.dataframe.sql.dataframe import DataFrame
+
+        if samplingRatio is not None or verifySchema:
+            raise NotImplementedError("Sampling Ration and Verify Schema are not supported")
```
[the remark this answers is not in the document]
izeigerman: Typo in my Type lol

sqlglot/helper.py
eakmanrq: Without this change though we aren't really ensuring it is a list which is confusing.

sqlglot/optimizer/pushdown_projections.py
eakmanrq: I think line 32 ensures that it will be defined unless I am missing something.

sqlglot/helper.py
tobymao: it’s just a bad name. it’s ensure enumerable, the better way is to do list(x) where you actually need a list

sqlglot/optimizer/pushdown_projections.py
eakmanrq: Good callout. It was falling back to an object instance comparison so I changed this to `is` instead of `==`.

sqlglot/dataframe/sql/column.py
eakmanrq: Yes this was a bad shortcut that I forgot to fix. Will address this as I integrate mypy and clean up the existing types.

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,730 @@
...
+            self.copy()
+            .withColumn("row_num", F.row_number().over(window))
+            .where(F.col("row_num") == F.lit(1))
+            .drop("row_num")
+        )
+
+    @operation(Operation.FROM)
+    def dropna(
+        self,
+        how: str = "any",
```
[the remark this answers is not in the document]
eakmanrq: This line covers the all case: `minimum_num_nulls = 1 if how == "any" else len(null_check_columns)`

sqlglot/dataframe/sql/dataframe.py
eakmanrq: The goal of this is to produce a consistent hash that is determined by the logic inside the CTE. The reason why we want it consistent is because that way if someone uses the tool to generate SQL to check it into VCS then they SQL will not change if their logic doesn't change.

sqlglot/dataframe/sql/dataframe.py
eakmanrq: Is the issue that it is pretty by default? I think users of the DataFrame API will want pretty by default. I'm not expecting many use cases where individuals will want the minimal SQL returns for performance reasons.

sqlglot/dataframe/sql/dataframe.py
eakmanrq: I'm not sure what you mean by this.

tobymao: make a method in expression.py that does this

eakmanrq: Ah I see there already is one. Will switch to that.

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,320 @@
...
+        if self.expression is None or isinstance(self.expression, (int, float, bool, dict, Iterable, datetime.date)):
```
tobymao, line 31: would it be better to do like

if not isinstance(expression, exp.Expression):

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,320 @@
...
+    @property
+    def is_alias(self):
+        return isinstance(self.expression, exp.Alias)
+
+    @property
+    def is_column(self):
+        return isinstance(self.expression, exp.Column)
+
+    @property
+    def column_expression(self) -> exp.Column:
```
tobymao, line 187: this is a duplicate of expression.unalias()

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def copy(self) -> Column:
+        return Column(self.expression.copy())
+
+    def set_table_name(self, table_name: str, copy=False) -> Column:
+        expression = self.expression.copy() if copy else self.expression
+        expression.set("table", exp.to_identifier(table_name))
+        return Column(expression)
+
+    def sql(self, **kwargs) -> Column:
```
tobymao, line 214: allow to pass in dialect too?

sql(**{"dialect": "spark", **kwargs})

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def isNull(self) -> Column:
+        new_expression = exp.Is(this=self.column_expression, expression=exp.Null())
+        return Column(new_expression)
+
+    def isNotNull(self) -> Column:
+        new_expression = exp.Not(this=exp.Is(this=self.column_expression, expression=exp.Null()))
+        return Column(new_expression)
+
+    def cast(self, dataType: t.Union[str, "DataType"]):
```
tobymao, line 267: use the real DataType?

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,320 @@
...
+        return self.invoke_expression_over_column(self, exp.ILike, expression=self._lit(other).expression)
+
+    def substr(self, startPos: t.Union[int, Column], length: t.Union[int, Column]) -> Column:
+        startPos = self._lit(startPos) if not isinstance(startPos, Column) else startPos
+        length = self._lit(length) if not isinstance(length, Column) else length
+        return Column.invoke_expression_over_column(
+            self, exp.Substring, start=startPos.expression, length=length.expression
+        )
+
+    def isin(self, *cols: t.Union["Primitives", t.Iterable["Primitives"]]):
```
tobymao, line 301: types in this file need to be updated

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+
+class DataFrame:
+    def __init__(
+        self,
+        spark: SparkSession,
+        expression: exp.Select,
+        branch_id: str = None,
+        sequence_id: str = None,
+        last_op: t.Optional[Operation] = Operation.INIT,
+        pending_hints: t.List[exp.Expression] = None,
```
tobymao, line 48: needs to be t.Optional

tests/test_executor.py
```diff
@@ -7,11 +8,13 @@
 from sqlglot import exp, parse_one
 from sqlglot.executor import execute
 from sqlglot.executor.python import Python
-from tests.helpers import FIXTURES_DIR, TPCH_SCHEMA, load_sql_fixture_pairs
+from tests.helpers import FIXTURES_DIR, TPCH_SCHEMA, load_sql_fixture_pairs, string_to_bool
 
 DIR = FIXTURES_DIR + "/optimizer/tpc-h/"
 
 
+@unittest.skipIf(string_to_bool(os.environ.get("SKIP_INTEGRATION", '0').lower()),
```
tobymao, line 16: can you make this var reusable?

sqlglot/dataframe/sql/column.py
eakmanrq: Yeah good call updated

tests/dataframe/integration/dataframe_validator.py
```diff
@@ -0,0 +1,151 @@
...
+import warnings
+
+import sqlglot
+from tests.helpers import string_to_bool
+
+if t.TYPE_CHECKING:
+    from pyspark.sql import DataFrame as SparkDataFrame
+
+
+@unittest.skipIf(string_to_bool(os.environ.get("SKIP_INTEGRATION", '0').lower()),
```
tobymao, line 13: yea lets just import SKIP_INTEGRATION instead of this whole thing

sqlglot/dataframe/sql/column.py
eakmanrq: awesome great addition

sqlglot/dataframe/sql/dataframe.py
tobymao: yep, consistency

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+            for hint in df.pending_join_hints:
+                for sequence_id_expression in hint.expressions:
+                    sequence_id_or_name = sequence_id_expression.alias_or_name
+                    sequence_ids_to_match = [sequence_id_or_name]
+                    if sequence_id_or_name in df.spark.name_to_sequence_id_mapping:
+                        sequence_ids_to_match = df.spark.name_to_sequence_id_mapping[sequence_id_or_name]
+                    matching_ctes = [
+                        cte for cte in reversed(expression.ctes) if cte.args["sequence_id"] in sequence_ids_to_match
+                    ]
+                    if matching_ctes:
```
tobymao, line 184: the if statement is unnecessary

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+                        for matching_cte in matching_ctes:
+                            if matching_cte.alias_or_name in [join_table.alias_or_name for join_table in join_tables]:
```
tobymao, line 186: take this for loop outside and make it a set?

join_aliases = {join_table.alias_or_name for . in .}

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+            hint_expression = exp.JoinHint(
+                this=hint_name, expressions=[exp.to_table(parameter.alias_or_name) for parameter in args]
+            )
+        else:
+            hint_expression = exp.Anonymous(this=hint_name, expressions=[parameter.expression for parameter in args])
+        new_df = self.copy()
+        new_df.pending_hints.append(hint_expression)
+        return new_df
+
+    def _set_operation(self, clazz: t.Callable, other: DataFrame, distinct: bool):
```
tobymao, line 207: change all clazz to klass

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+        df = self._convert_leaf_to_cte()
+        df.expression.ctes[-1].set("cache_storage_level", storage_level)
+        return df
+
+    @classmethod
+    def _add_ctes_to_expression(cls, expression: exp.Expression, ctes: t.List[exp.CTE]) -> exp.Expression:
+        expression = expression.copy()
+        with_expression = expression.args.get("with")
+        if with_expression:
+            existing_ctes = with_expression.args["expressions"]
```
tobymao, line 228: with_expression.expressions

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+            existsing_cte_names = [x.alias_or_name for x in existing_ctes]
```
tobymao, line 229: make this a set {name for .. .}

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+                    existing_ctes.append(cte)
+        else:
+            existing_ctes = ctes
+        expression.set("with", exp.With(expressions=existing_ctes))
+        return expression
+
+    @classmethod
+    def _get_outer_select_columns(cls, item: t.Union[exp.Expression, DataFrame]) -> t.List[Column]:
+        expression = item.expression if isinstance(item, DataFrame) else item
+        return [Column(x) for x in dict.fromkeys(expression.find(exp.Select).args.get("expressions", []))]
```
tobymao, line 241: are you trying to deduplicate this?

tobymao: find(exp.Select).expressions, automatically gives you the empty list default

sqlglot/dataframe/sql/dataframe.py
tobymao: so let's get rid of the pretty default because everywhere else pretty is not default

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -0,0 +1,737 @@
...
+        from sqlglot.dataframe.sql.functions import lit
+
+        values = None
+        columns = None
+        new_df = self.copy()
+        all_columns = self._get_outer_select_columns(new_df.expression)
+        all_column_mapping = {column.alias_or_name: column for column in all_columns}
+        if isinstance(value, dict):
+            values = value.values()
+            columns = self._ensure_and_normalize_cols(list(value.keys()))
```
tobymao, line 580: list(value) don't need keys

sqlglot/dataframe/sql/session.py
```diff
@@ -0,0 +1,145 @@
...
+from sqlglot.dataframe.sql.readwriter import DataFrameReader
+from sqlglot.dataframe.sql.types import StructType
+from sqlglot.dataframe.sql.util import get_column_mapping_from_schema_input
+
+if t.TYPE_CHECKING:
+    from sqlglot.dataframe.sql._typing import SchemaInput
+
+
+class SparkSession:
+    known_ids: t.ClassVar[t.Set[str]] = set()
```
tobymao, line 20: mutable class vars? this is intended?

sqlglot/dialects/spark.py
```diff
@@ -98,20 +88,22 @@ class Generator(Hive.Generator):
         }
 
         TRANSFORMS = {
-            **{k: v for k, v in Hive.Generator.TRANSFORMS.items() if k not in {exp.ArraySort}},
+            **{k: v for k, v in Hive.Generator.TRANSFORMS.items() if k not in {exp.ArraySort, exp.Split, exp.ILike}},
```
tobymao: remove the split?

sqlglot/schema.py
```diff
@@ -49,37 +60,49 @@ class MappingSchema(Schema):
...
+
+    def _get_table_args_from_table(self, table):
+        if table.args.get("catalog") is not None:
+            return "catalog", "db", "this"
+        if table.args.get("db") is not None:
+            return "db", "this"
+        return ("this",)
+
+    def copy(self, **kwargs):
+        kwargs = {**{"schema": copy(self.schema)}, **kwargs}
```
tobymao, line 82: why not just do self.schema.copy()

sqlglot/dataframe/sql/dataframe.py
eakmanrq: Yeah I think this can actually be removed. Thanks for calling it out.

sqlglot/schema.py
tobymao: why is this double nested?

MappingSchema(**{"schema": self.schema.copy(), **kwargs})

tests/dialects/test_duckdb.py
tobymao: confirm should revert

sqlglot/dialects/spark.py
eakmanrq: Hmm I did I think your page hadn't refreshed.

sqlglot/dataframe/sql/session.py
eakmanrq: Yes and I believe this will be fine in practice. I think technically though this could be a singleton if we wanted.

sqlglot/expressions.py
eakmanrq: So the way that the PySpark DataFrame API thinks about this is that an empty group by prior to an aggregate is valid and represents grouping by nothing. This change allows the sqlglot builder to behave in the same way. If you would prefer for the builder to not behave that way then I can remove this and then do a check in the DataFrame API version.

sqlglot/dataframe/sql/column.py
```diff
@@ -0,0 +1,330 @@
...
+    def ensure_col(cls, value: t.Union[ColumnOrName, int, float]):
+        return cls(value)
+
+    @classmethod
+    def ensure_cols(cls, args: t.List[t.Union[int, str, Column]]) -> t.List[Column]:
+        return [cls.ensure_col(x) if not isinstance(x, Column) else x for x in args]
+
+    @classmethod
+    def _lit(cls, value: t.Optional[t.Any] = None) -> Column:
+        if value is None:
```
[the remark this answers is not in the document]
eakmanrq: Great suggestion. I wish I had known about that converter earlier. The only difference was how we handled dictionaries so I left that part separate.

sqlglot/dataframe/README.md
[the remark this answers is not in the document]
eakmanrq: This feels more to me like a high level usage of sqlglot and therefore pretty by default instead of low level calls where we assume not pretty.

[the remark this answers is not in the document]
tobymao: all sqglot should be non pretty by default, users can set pretty=True if they want the other behavior

sqlglot/helper.py
eakmanrq: Ok I reverted this change

sqlglot/expressions.py
eakmanrq: I think it makes sense here alongside the other function to help construct expressions (like `to_alias` and `to_identity`)

sqlglot/helper.py
eakmanrq: reverted

tests/dialects/test_presto.py
eakmanrq: reverted

sqlglot/schema.py
eakmanrq: @barakalon Yeah I think the second `MutableSchema` interface makes sense. Working on that change now.

sqlglot/schema.py
eakmanrq: Yeah going that direction now: https://github.com/tobymao/sqlglot/pull/568#discussion_r992944368

sqlglot/expressions.py
```diff
@@ -3321,6 +3346,12 @@ def convert(value):
             keys=[convert(k) for k in value.keys()],
             values=[convert(v) for v in value.values()],
         )
+    if isinstance(value, datetime.datetime):
+        datetime_literal = Literal.string(value.strftime("%Y-%m-%d %H:%M:%S"))
```
tobymao, line 3350: i don't believe these are correct. you're hard-coding hive date formats here. it might be more generic to do

Cast(this=literal, to=DataType.build("timestamp"))

Also should you handle timezones? in that case maybe you want utc time or using iso format

this holds the same for date

i also have a method called TimeStringToTime and DateStringToDate, the issue with those functions is that i haven't implemented them for every dialect yet, although we could make those default to CAST(...)

this may be the way to go

tobymao:
you added a bad file test_dialects.py

sqlglot/schema.py
eakmanrq: We know that it will be given the structure of the schema mapping we are using in this file. A good reason why this function exists here and not as a general utility function because of assumptions like that.

sqlglot/dataframe/sql/functions.py
```diff
@@ -0,0 +1,1251 @@
...
+    return Column(column_name)
+
+
+def lit(value: t.Optional[t.Any] = None) -> Column:
+    if isinstance(value, str):
+        return Column(glotexp.Literal.string(str(value)))
+    return Column(value)
+
+
+def greatest(*cols: ColumnOrName) -> Column:
```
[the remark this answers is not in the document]
eakmanrq: I don't want to hold up this PR for that feature so I created an issue: https://github.com/tobymao/sqlglot/issues/597

sqlglot/dataframe/sql/column.py
eakmanrq: Many can. I think I'm actually good with this approach for now and see if we get any feedback about issues. My guess is that this won't be an issue.

eakmanrq:
Addressed all the feedback and added mypy. I did add some ignores but not many and one of the main issues was around dealing with inputs that may or may not be nested and therefore I may or may not flatten.

tobymao (approved):

sqlglot/schema.py
```diff
@@ -0,0 +1,297 @@
...
+            table (sqlglot.expressions.Table): The source table.
+            column (sqlglot.expressions.Column): The target column.
+        Returns:
+            sqlglot.expressions.DataType.Type: The resulting column type.
+        """
+
+
+class MutableSchema(Schema):
+    """
+    Based on MappingSchema but mutable so you can add tables over time.
```
barakalon, line 37: Strange that the subclass is "based on the MappingSchema", a superclass.

I think this class should be a lot simpler. It should be more of an interface.

tobymao: i would have expected the default Mapping Schema to come first

eakmanrq: I think that docstring was just not updated after the refactor. 

Mapping schema did come first before the last request for changes. The last request asked for this hierarchy of Schema -> MutableSchema(Schema) -> MappingSchema(MutableSchema)

barakalon: Right now, all `MappingSchema` does is _unimplement_ add_table. That's silly.

I think it's fine for the the MappingSchema to implement add_table. I just didn't think add_table should be part of the main Schema interface.

We only need this intermediate interface, MutableSchema, if the DataFrame API wants to support custom Schema types.

sqlglot/schema.py
```diff
@@ -0,0 +1,300 @@
...
+from sqlglot.helper import csv_reader
+
+
+class Schema(abc.ABC):
+    """Abstract base class for database schemas"""
+
+    @abc.abstractmethod
+    def add_table(self, table, column_mapping=None):
+        """
+        Registers the table to be a known schema to be accessed later. Some implementing classes may
```
barakalon, line 14: Can you rephrase this description? It's not very clear.

barakalon (approved):

barakalon (approved):

Issue #564, opened 2022-10-09
Optimizer: Remove Unused CTE Statements


eakmanrq:
Currently if a CTE exists that is not used at all to produce the final result the optimizer leaves it. We should be removing these so that the optimized SQL only includes CTEs that are actually used.

barakalon:
I was hoping to do something similar for LEFT JOINs

tobymao:
left join removal needs to be an option because left joins can change cardinality despite not being used

tobymao:
can probably reuse the ref count code i have

barakalon:
I think you can remove the left join if you know there are unique values for the right side of the join condition. That happens if:
1. The inner key is joined on a column with a unique constraint (e.g. a primary key). Probably not worth adding this logic right now since we focus on OLAP use cases.
2. The inner query is a SELECT DISTINCT.
3. The inner query has a GROUP BY clause, and all the grouped columns are in the join condition. This might actually be common, especially since we create these types of inner queries in unnest_subqueries

tobymao:
yes, can definitely do 2 or 3

PR #572, opened 2022-10-11
Optimizer: Eliminate joins

barakalon:
Adds optimizer rule to remove unused LEFT JOINs

More context: https://github.com/tobymao/sqlglot/issues/564

### DIFF
--- a/sqlglot/optimizer/eliminate_joins.py
+++ b/sqlglot/optimizer/eliminate_joins.py
@@ -82,7 +82,7 @@ def _unique_outputs(scope):
                 unique_outputs.add(select.alias_or_name)
 
         # All the grouped expressions must be in the output
-        if len(grouped_expressions - grouped_outputs) == 0:
+        if not grouped_expressions.difference(grouped_outputs):
             return unique_outputs
         else:
             return set()
--- a/sqlglot/optimizer/eliminate_joins.py
+++ b/sqlglot/optimizer/eliminate_joins.py
@@ -5,13 +5,32 @@ from sqlglot.optimizer.simplify import simplify
 
 
 def eliminate_joins(expression):
+    """
+    Remove unused joins from an expression.
+
+    Currently, this only remove LEFT JOINs when we know that the join condition
+    doesn't produce duplicate rows.
+
+    Example:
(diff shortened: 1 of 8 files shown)


sqlglot/optimizer/eliminate_joins.py
```diff
@@ -0,0 +1,140 @@
...
+                unique_outputs.add(select.alias_or_name)
+
+        # All the grouped expressions must be in the output
+        if len(grouped_expressions - grouped_outputs) == 0:
+            return unique_outputs
+
+    return set()
+
+
+def join_condition(join):
```
barakalon, line 98: Extracted from the planner.

Might be a better module to put this...

tobymao:
what about ctes? should we rename the method?

barakalon:
I think that should be a separate rule.
I was going to do it after.

sqlglot/optimizer/eliminate_joins.py
tobymao: how did you find this

sqlglot/optimizer/eliminate_joins.py
```diff
@@ -0,0 +1,138 @@
...
+
+        unique_outputs = set()
+        for select in scope.selects:
+            output = select.unalias()
+            if output in grouped_expressions:
+                grouped_outputs.add(output)
+                unique_outputs.add(select.alias_or_name)
+
+        # All the grouped expressions must be in the output
+        if len(grouped_expressions - grouped_outputs) == 0:
```
tobymao, line 85: groped_expressions >= grouped_outptus

tobymao (approved):

sqlglot/optimizer/eliminate_joins.py
barakalon: haha I figured you must've implemented some similar logic with the executor, so I went looking for it. And bam, exactly what I needed.

sqlglot/optimizer/eliminate_joins.py
barakalon: This is a set operation. I'll use explicit set methods so the intent is clearer.

PR #578, opened 2022-10-11
Optimizer: Eliminate cross joins

barakalon:
We can also remove cross joins if we know the inner query only has one output row.

### DIFF
--- a/sqlglot/optimizer/eliminate_joins.py
+++ b/sqlglot/optimizer/eliminate_joins.py
@@ -103,7 +103,11 @@ def _unique_outputs(scope):
 
 
 def _has_single_output_row(scope):
-    return all(isinstance(e.unalias(), exp.AggFunc) for e in scope.selects) or _is_limit_1(scope)
+    return (
+        all(isinstance(e.unalias(), exp.AggFunc) for e in scope.selects)
+        or _is_limit_1(scope)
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/eliminate_joins.py
```diff
@@ -87,12 +96,21 @@ def _unique_outputs(scope):
...
 
-    if all(isinstance(e.unalias(), exp.AggFunc) for e in scope.selects):
+    if _has_single_output_row(scope):
         return set(scope.expression.named_selects)
 
     return set()
 
 
+def _has_single_output_row(scope):
+    return all(isinstance(e.unalias(), exp.AggFunc) for e in scope.selects) or _is_limit_1(scope)
```
tobymao, line 106: what about 

Select 1

or 

values ()

tobymao (approved):

sqlglot/optimizer/eliminate_joins.py
barakalon: Oh, woah - I wonder what happens with `VALUES` right now... I don't think the optimizer handles that at all.

barakalon: I'll add logic for a SELECT with no FROM, but I'll skip on VALUES for now

PR #581, opened 2022-10-12
Fix duckdb array type

eakmanrq:
https://duckdb.org/docs/sql/data_types/overview

DuckDB expects that array are declared using `INTERIOR[]` instead of `ARRAY<INTERIOR>`.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -71,8 +71,7 @@ def _struct_pack_sql(self, expression):
 def _datatype_sql(self, expression):
     type_value = expression.this
     if type_value == exp.DataType.Type.ARRAY:
-        interior = self.expressions(expression, flat=True)
-        return f"{interior}[]"
+        return f"{self.expressions(expression, flat=True)}[]"
     else:
         return self.datatype_sql(expression)
 
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -68,6 +68,15 @@ def _struct_pack_sql(self, expression):
     return f"STRUCT_PACK({', '.join(args)})"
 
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -68,6 +68,15 @@ def _struct_pack_sql(self, expression):
     return f"STRUCT_PACK({', '.join(args)})"
 
 
+def _datatype_sql(self, expression):
+    type_value = expression.this
+    if type_value == exp.DataType.Type.ARRAY:
+        interior = self.expressions(expression, flat=True)
```
tobymao, line 74: sorry to nit but i'm not a fan of this style can we just do

if expression.this == exp.DataType.Type.ARRAY:
    return f"{self.expressions(expression, flat=True)}[]"

sqlglot/parser.py
```diff
@@ -1676,6 +1676,11 @@ def _parse_types(self):
         is_struct = type_token == TokenType.STRUCT
         expressions = None
 
+        if not nested and self._match_pair(TokenType.L_BRACKET, TokenType.R_BRACKET):
```
tobymao, line 1679: what is this?

tobymao: is this if you do int[] which means int array?

eakmanrq: Yeah `int[]` means an int array.

sqlglot/parser.py
```diff
@@ -1676,6 +1676,11 @@ def _parse_types(self):
...
+            return exp.DataType(
+                this=exp.DataType.Type.ARRAY, expressions=[exp.DataType.build(type_token.value.upper())], nested=True
```
tobymao, line 1681: you don't need to call upper here

PR #588, opened 2022-10-13
Remove table properties from duckdb

eakmanrq:
I don't see anywhere in the DuckDB documentation that it says that you can define any table properties. It also doesn't support partitioning. So this PR removes all table properties from DuckDB generated DDL statements.

- [ ] Have columns defined in partition expressions be included in the column list for DuckDB


tobymao (approved):

eakmanrq:
We've decided to leave the partition columns issue unaddressed in this PR.

PR #591, opened 2022-10-13
Make selects and named_selects required of subqueryable objects

eakmanrq:


sqlglot/expressions.py
```diff
@@ -1105,6 +1105,14 @@ def ctes(self):
             return []
         return with_.expressions
 
+    @property
+    def selects(self):
+        raise NotImplementedError("Subqueryable objects must implement `selects`")
```
barakalon, line 1110: Which expressions don't implement this?
Should we just implement them?

barakalon (approved):

sqlglot/expressions.py
eakmanrq: They all do now (I just merged in a change that implemented it in union). I put this in place now to make it clear that future expressions should implement.

PR #592, opened 2022-10-13
Add E'\\xDEADBEEF' (postgres) and N'Some Text' (mysql)

GeorgeSittas:
Opening this PR for feedback. Aims to support syntax mentioned in https://github.com/tobymao/sqlglot/pull/586. If this is on the right track, I can try implementing the `N'literal'` syntax similarly.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -56,6 +56,7 @@ class TokenType(AutoName):
     VAR = auto()
     BIT_STRING = auto()
     HEX_STRING = auto()
+    BYTEA_ESCAPE = auto()
 
     # types
     BOOLEAN = auto()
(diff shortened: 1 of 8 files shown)


sqlglot/tokens.py
```diff
@@ -924,8 +933,8 @@ def _scan_numeric_string(self, string_start):
 
         try:
             self._add(token_type, f"{int(text, base)}")
-        except ValueError:
```
tobymao, line 927: don't change this

tobymao: do something different is base is None and change the method name

GeorgeSittas: changing name in just a sec

GeorgeSittas: done

tobymao (approved):

PR #601, opened 2022-10-17
Add to_column and add helper for splitting to num words

eakmanrq:
I removed the doc string tests for to_table and to_column since they didn't really work before due to a recursion issue.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3241,12 +3241,12 @@ def to_table(sql_path: str, **kwargs) -> Table:
 
 def to_column(sql_path: str, **kwargs) -> Column:
     """
-    Create a column from a `[schema].[table]` sql path. Schema is optional.
+    Create a column from a `[table].[column]` sql path. Schema is optional.
 
     If a column is passed in then that column is returned.
 
     Args:
-        sql_path: `[schema].[table]` string
+        sql_path: `[table].[column]` string
     Returns:
         Table: A column expression
     """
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -3237,13 +3234,28 @@ def to_table(sql_path, **kwargs):
...
 
 
+def to_column(sql_path: str, **kwargs) -> Column:
+    """
+    Create a column from a `[schema].[table]` sql path. Schema is optional.
+
+    If a column is passed in then that column is returned.
+
+    Args:
+        sql_path: `[schema].[table]` string
```
tobymao, line 3249: it should be [table].[column]

sqlglot/helper.py
```diff
@@ -170,3 +171,27 @@ def find_new_name(taken, base):
...
+        [None, 'db', 'table']
+        >>> split_num_words("db.table", ".", 3, fill_from_start=False)
+        ['db', 'table', None]
+        >>> split_num_words("db.table", ".", 1)
+        ['db', 'table']
+    """
+    words = value.split(sep)
+    if fill_from_start:
+        return [x if x is not None else x for x in [None] * (min_num_words - len(words)) + words]
+    return [x if x is not None else x for x in words + [None] * (min_num_words - len(words))]
```
tobymao, line 197: can you reverse all the is not to just x is none

sqlglot/expressions.py
tobymao: this is similar to def column but i guess it takes a path? i wonder if def column and def table could be merged with to_column and to_table so we don't need both

sqlglot/helper.py
eakmanrq: actually that isn't even needed anymore so I will simplify

tobymao (approved):

PR #602, opened 2022-10-18
Add support for array_concat

eakmanrq:
Also simplified the `rename_func`.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -2,7 +2,7 @@ from enum import Enum
 
 from sqlglot import exp
 from sqlglot.generator import Generator
-from sqlglot.helper import list_get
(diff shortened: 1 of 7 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -181,11 +181,7 @@ def generator(self, **opts):
 
 def rename_func(name):
     def _rename(self, expression):
-        args = (
-            expression.expressions
-            if isinstance(expression, exp.Func) and expression.is_var_len_args
-            else expression.args.values()
-        )
+        args = list(flatten(expression.args.values()))
```
tobymao, line 184: i don't think you need the list here right

eakmanrq: right thanks

PR #607, opened 2022-10-18
Transpile snowflake epoch_ time parts

GeorgeSittas:
This is an attempt to tackle https://github.com/tobymao/sqlglot/discussions/604.

References:
- https://prestodb.io/docs/current/functions/datetime.html
- https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts
- https://docs.snowflake.com/en/sql-reference/functions/date_part.html

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -69,6 +69,20 @@ def _unix_to_time(self, expression):
     raise ValueError("Improper scale for timestamp")
 
 
+# https://docs.snowflake.com/en/sql-reference/functions/date_part.html
+# https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts
+def _parse_date_part(self):
+    this = self._parse_var() or self._parse_type()
+    self._match(TokenType.COMMA)
+    expression = self._parse_bitwise()
+
+    if this.name.upper().startswith("EPOCH"):
+        ts = self.expression(exp.Cast, this=expression, to=exp.DataType.build("TIMESTAMP"))
+        return self.expression(exp.TimeToUnix, this=ts)
+
+    return self.expression(exp.Extract, this=this, expression=expression)
+
+
 class Snowflake(Dialect):
     null_ordering = "nulls_are_large"
     time_format = "'yyyy-mm-dd hh24:mi:ss'"
@@ -115,7 +129,7 @@ class Snowflake(Dialect):
 
         FUNCTION_PARSERS = {
             **Parser.FUNCTION_PARSERS,
-            "DATE_PART": lambda self: self._parse_extract(),
+            "DATE_PART": _parse_date_part,
         }
 
         FUNC_TOKENS = {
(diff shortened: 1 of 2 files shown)


tests/dialects/test_snowflake.py
```diff
@@ -240,11 +240,18 @@ def test_timestamps(self):
             },
         )
         self.validate_all(
-            "SELECT DATE_PART(month FROM a::DATETIME)",
+            "SELECT DATE_PART(month, a::DATETIME)",
```
GeorgeSittas, line 243: This was changed to reflect DATE_PART's syntax as shown in the [docs](https://docs.snowflake.com/en/sql-reference/functions/date_part.html#syntax).

sqlglot/dialects/snowflake.py
```diff
@@ -69,6 +69,20 @@ def _unix_to_time(self, expression):
...
 
 
+# https://docs.snowflake.com/en/sql-reference/functions/date_part.html
+# https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts
+def _parse_date_part(self):
+    this = self._parse_var() or self._parse_type()
+    self._match(TokenType.COMMA)
+    expression = self._parse_bitwise()
+
+    if this.name.upper().startswith("EPOCH"):
```
tobymao, line 79: why do you do startswith? shouldn't you check epoch_seconds and milliseconds?

GeorgeSittas: I started by checking those (+micro, nano), but I was unsure how to maintain the precision level because we're transforming the expression by using casts.

tobymao: well that just depends on how other formats store time, should probably be fine

sqlglot/dialects/snowflake.py
```diff
@@ -76,9 +76,24 @@ def _parse_date_part(self):
...
+    name = this.name.upper()
+    if name.startswith("EPOCH"):
+        scale = None
+
+        if name.startswith("EPOCH_MILLISECOND"):
+            scale = 10 ** 3
+        elif name.startswith("EPOCH_MICROSECOND"):
+            scale = 10 ** 6
+        elif name.startswith("EPOCH_NANOSECOND"):
+            scale = 10 ** 9
```
tobymao, line 88:
```suggestion
        else:
            scale = None
```

sqlglot/dialects/snowflake.py
```diff
@@ -76,9 +76,24 @@ def _parse_date_part(self):
...
+
```
tobymao, line 89: else:
    scale = None

sqlglot/dialects/snowflake.py
```diff
@@ -76,9 +76,24 @@ def _parse_date_part(self):
     self._match(TokenType.COMMA)
     expression = self._parse_bitwise()
 
-    if this.name.upper().startswith("EPOCH"):
+    name = this.name.upper()
+    if name.startswith("EPOCH"):
+        scale = None
+
+        if name.startswith("EPOCH_MILLISECOND"):
```
tobymao, line 81: you don't need starts with here

GeorgeSittas: I did this to allow both epoch_millisecond and epoch_milliseconds

sqlglot/dialects/snowflake.py
```diff
@@ -76,9 +76,24 @@ def _parse_date_part(self):
     self._match(TokenType.COMMA)
     expression = self._parse_bitwise()
 
-    if this.name.upper().startswith("EPOCH"):
+    name = this.name.upper()
+    if name.startswith("EPOCH"):
```
tobymao, line 80: just do 

if name == ...
elif name == 
else ...

you don't need starts with at all i think

tobymao (approved):

PR #608, opened 2022-10-19
determine order by references when qualifying columns

ginter:
Attempt to qualify order by references that appear to be column references and not aliases.

This helps in Minerva SQL as we need columns to be qualified to aid in rewriting queries. There is also value in doing this optimization to remove some inconsistency with how different engines handle this case of an ambiguous alias/column reference in an order by clause.

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -211,6 +211,23 @@ def _qualify_columns(scope, resolver):
             if column_table:
                 column.set("table", exp.to_identifier(column_table))
 
+    # Determine whether each reference in the order by clause is to a column or an alias.
+    order = scope.find(exp.Order)
+    if order is not None:
+        for column in order.find_all(exp.Column):
+            column_table = column.table
+            column_name = column.name
+
+            if column_table or isinstance(column.parent, exp.Ordered) or column.name not in resolver.all_columns:
+                continue
+
+            column_table = resolver.get_table(column_name)
+
+            if column_table is None:
+                raise OptimizeError(f"Ambiguous column: {column_name}")
+
+            column.set("table", exp.to_identifier(column_table))
+
 
 def _expand_stars(scope, resolver):
     """Expand stars to lists of column selections"""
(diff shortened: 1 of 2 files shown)


ginter:
@barakalon this attempts to address the bug we're running into with SR. not sure if this is the approach you had in mind or not, just lmk.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -211,6 +211,23 @@ def _qualify_columns(scope, resolver):
             if column_table:
                 column.set("table", exp.to_identifier(column_table))
 
+    # Determine whether each reference in the order by clause is to a column or an alias.
+    order = scope.find(exp.Order)
```
tobymao, line 215: maybe better to do like

for ordered in scope.find_all(exp.Ordered)

tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -72,6 +72,9 @@ SELECT x.a AS a, x.b AS b FROM x AS x ORDER BY a;
 SELECT a FROM x ORDER BY b;
 SELECT x.a AS a FROM x AS x ORDER BY x.b;
 
+SELECT SUM(a) AS a FROM x ORDER BY SUM(a);
```
tobymao, line 75: i assume you tested this case in the various engines and this is the behavior?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -211,6 +211,23 @@ def _qualify_columns(scope, resolver):
...
+    if order is not None:
+        for column in order.find_all(exp.Column):
```
barakalon, line 217: Should this logic be here?: https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L235

barakalon: That is, Scope.columns has logic to exclude columns in the Order clause. Sounds like that's wrong.

This change fixes qualify_columns, but it doesn't fix everything else: Scope.columns still won't return these columns.

ginter: I considered putting it there but we'll need access to the schema to do so. Also I assumed you intentionally left those out from scope#columns bc the reference may be to an alias, in which case qualifying it would be incorrect

tests/fixtures/optimizer/qualify_columns.sql
ginter: this mimics the mysql behavior for whether it resolves to an alias or a column. Starrocks always resolves to an alias if there is one which leads to unexpected behavior in situations like this - the ordering was random when I tested this scenario.

barakalon (approved):
Talked offline.

PR #614, opened 2022-10-19
Refactor table alias semantics & fix parenthesized join construct

### LINKED ISSUE #612: [Postgres[ Parser does not understand table aliases inside join construct
Related to: https://github.com/tobymao/sqlglot/issues/423

The same error occurs as in the issue above when using table aliases inside the parentheses.

So this works (fixed in previous issue):
```
select * from (table1 left join table2 on 1 = 1);
```

But this doesn't:
```
select * from (table1 t1 left join table2 t2 on 1 = 1);
```

```
ParseError: Expecting ). Line 1, Col: 26.
  select * from (table1 t1 left join table2 t2 on 1 = 1)

Invalid expression / Unexpected token. Line 1, Col: 54.
  select * from (table1 t1 left join table2 t2 on 1 = 1)
```

Using sqlglot 8.0.2.

GeorgeSittas:
WIP (experimenting with changing table alias semantics...)
Fixes #612


tobymao:
fyi @barakalon this is going to change how we represent alias's on tables. before we did 

alias ( table), but now alias will be a property in table

barakalon:
I'm surprised the optimizer didn't explode with this. Nice.

@ginter I think this will simplify some of our code.

tobymao:
@barakalon it does explode, @GeorgeSittas is fixing it now

GeorgeSittas:
It does seem to simplify parts of the code indeed

sqlglot/parser.py
```diff
@@ -1120,6 +1120,9 @@ def _parse_subquery(self, this):
         )
 
     def _parse_query_modifiers(self, this):
+        if isinstance(this, exp.Alias):
```
tobymao, line 1123: i don't think this is needed?

tests/dialects/test_dialect.py
```diff
@@ -716,6 +716,13 @@ def test_joined_tables(self):
                 "mysql": "SELECT * FROM (tbl1 JOIN LATERAL (SELECT * FROM bla) AS tbl)",
             },
         )
+        self.validate_all(
+            "SELECT * FROM (table1 t1 LEFT JOIN table2 t2 ON 1 = 1)",
```
tobymao, line 720: i think you only need an identity sql test for this one

barakalon (approved):
Awesome!

This will be another breaking change 🙃

PR #615, opened 2022-10-20
change default error level to immediate

tobymao:
errorlevel.RAISE kind of sucks, i think going to immediate is better, and we're bumping version 9 tomorrow anyways, thoughts?


barakalon (approved):
Agreed. Let’s do it

PR #619, opened 2022-10-20
taerts/tsql datename and datepart

TimothyAerts:
Datename and Datepart added

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -32,6 +32,11 @@ class TSQL(Dialect):
         }
 
     class Parser(Parser):
+        FUNCTIONS = {
+            **Parser.FUNCTIONS,
+            "CHARINDEX": exp.StrPosition.from_arg_list,
+        }
+
         def _parse_convert(self):
             to = self._parse_types()
             self._match(TokenType.COMMA)
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -90,6 +90,7 @@ class TSQL(Dialect):
             this = self._parse_field()
             return self.expression(exp.Cast, this=this, to=to)
 
+
     class Generator(Generator):
         TYPE_MAPPING = {
             **Generator.TYPE_MAPPING,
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -3,9 +3,42 @@
...
 from sqlglot.tokens import Tokenizer, TokenType
+from sqlglot.helper import list_get
+from sqlglot.time import format_time
 
 
 class TSQL(Dialect):
+    date_name_mapping = {"weekday": "%A", "dw": "%A", "w": "%A", "month": "%b", "mm": "%b", "m": "%b"}
+    time_mapping = {
+        "yyyy": "%Y",
+        "yy": "%Y",
```
tobymao, line 14: is this correct?

sqlglot/dialects/tsql.py
```diff
@@ -35,6 +68,20 @@ class Parser(Parser):
         FUNCTIONS = {
             **Parser.FUNCTIONS,
             "CHARINDEX": exp.StrPosition.from_arg_list,
+            "DATENAME": lambda args: exp.TimeToStr(
+                this=list_get(args, 1),
+                format=exp.Literal(
+                    this=format_time(list_get(args, 0).sql(), {**TSQL.time_mapping, **TSQL.date_name_mapping}),
```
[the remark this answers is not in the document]
TimothyAerts: Hi,

First attempt was with the format_time_lambda method which I think is the preferred solution?

Stepped away from that because the format_time expectes a string. 
Because the parameters in TSQL for these functions are not wrapped in quotes the tokenizer produces a Column: Identifier for the parameter. 
So I went with the option to forcefully take the value as a string into the format_time. I'll try to see if I can come up with something better. Do you have any suggestions or tips?

sqlglot/dialects/tsql.py
TimothyAerts: TSQL has different conventions in its functions:
DateName:
image
DatePart:
image

So I figured to create a default time_mapping which is then overwritten depending on which date function is used.
Please let me know if this is not desirable.

PR #620, opened 2022-10-21
taerts/tsql datename datepart

TimothyAerts:
Hi,

Apologies for the mess I made in the previous PR I am new to this way of working with forks and got my remote's mixed up.
Here another attempt.
About the double time mapping:
TSQL has different conventions in its functions:
DateName:
image
DatePart:
image

So I figured to create a default time_mapping which is then overwritten depending on which date function is used.
Please let me know if this is not desirable.

Again apologies for the previously messed up PR

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -111,7 +111,7 @@ class Spark(Hive):
             exp.StructKwarg: lambda self, e: f"{self.sql(e, 'this')}: {self.sql(e, 'expression')}",
             exp.TimestampTrunc: lambda self, e: f"DATE_TRUNC({self.sql(e, 'unit')}, {self.sql(e, 'this')})",
             exp.VariancePop: rename_func("VAR_POP"),
-            exp.PartsToDate: rename_func("MAKE_DATE"),
+            exp.DateFromParts: rename_func("MAKE_DATE"),
         }
 
         WRAP_DERIVED_VALUES = False
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -111,6 +111,7 @@ class Spark(Hive):
             exp.StructKwarg: lambda self, e: f"{self.sql(e, 'this')}: {self.sql(e, 'expression')}",
             exp.TimestampTrunc: lambda self, e: f"DATE_TRUNC({self.sql(e, 'unit')}, {self.sql(e, 'this')})",
             exp.VariancePop: rename_func("VAR_POP"),
+            exp.PartsToDate: lambda self, e: f"MAKE_DATE({self.sql(e,'this')},{self.sql(e,'month')},{self.sql(e, 'day')})",
         }
 
         WRAP_DERIVED_VALUES = False
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -111,7 +111,7 @@ class Spark(Hive):
             exp.StructKwarg: lambda self, e: f"{self.sql(e, 'this')}: {self.sql(e, 'expression')}",
             exp.TimestampTrunc: lambda self, e: f"DATE_TRUNC({self.sql(e, 'unit')}, {self.sql(e, 'this')})",
             exp.VariancePop: rename_func("VAR_POP"),
-            exp.PartsToDate: lambda self, e: f"MAKE_DATE({self.sql(e,'this')},{self.sql(e,'month')},{self.sql(e, 'day')})",
+            exp.PartsToDate: rename_func("MAKE_DATE"),
         }
 
         WRAP_DERIVED_VALUES = False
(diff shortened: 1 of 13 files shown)


sqlglot/expressions.py
```diff
@@ -2411,6 +2411,10 @@ class TimeTrunc(Func, TimeUnit):
     arg_types = {"this": True, "unit": True, "zone": False}
 
 
+class PartsToDate(Func):
+    arg_types = {"this": True, "month": True, "day": True}
```
GeorgeSittas, line 2415: this -> year

sqlglot/dialects/tsql.py
```diff
@@ -1,14 +1,68 @@
 from sqlglot import exp
 from sqlglot.dialects.dialect import Dialect
 from sqlglot.generator import Generator
+from sqlglot.helper import list_get
 from sqlglot.parser import Parser
+from sqlglot.time import format_time
 from sqlglot.tokens import Tokenizer, TokenType
 
 
+def tsql_format_time_lambda(exp_class, mapping=None, default=None):
```
GeorgeSittas, line 10: Why can't we use format time lambda here?

TimothyAerts: TSQL has inversted arguments and the arguments are not tokanized correctly.
The first argument is tokenized as column value, passing that to the regular format time lambda causes unexpected behaviour where the argument is not time_mapped.

GeorgeSittas: Ok, I'll check it more thoroughly later then and let you know if I can come up with an alternative.

sqlglot/dialects/tsql.py
```diff
@@ -4,20 +4,65 @@
 from sqlglot.generator import Generator
 from sqlglot.helper import list_get
 from sqlglot.parser import Parser
+from sqlglot.time import format_time
 from sqlglot.tokens import Tokenizer, TokenType
 
 
+def tsql_format_time_lambda(exp_class, mapping=None, default=None):
+    corrected_mapping = {
```
tobymao, line 12: make this a constant?

TimothyAerts: Constant inside the method or constant in the class and passed along to the method?

tobymao: make it a constant at the tsql file level, and just reference it

TimothyAerts: put it at file level and gave it a more generic name. function has a switch to use the overwrite for full expressions.

sqlglot/dialects/tsql.py
```diff
@@ -4,20 +4,63 @@
...
+FULL_EXP_TIME_MAPPING = {"weekday": "%A", "dw": "%A", "w": "%A", "month": "%B", "mm": "%B", "m": "%B"}
+
+
+def tsql_format_time_lambda(exp_class, full_exp_mapping=None, default=None):
+    def _format_time(args):
+        return exp_class(
+            this=list_get(args, 1),
+            format=exp.Literal.string(
+                format_time(
+                    list_get(args, 0).text("this") or (TSQL.time_format if default is True else default),
```
tobymao, line 19:
```suggestion
                    list_get(args, 0).name or (TSQL.time_format if default is True else default),
```

sqlglot/expressions.py
```diff
@@ -2411,6 +2411,10 @@ class TimeTrunc(Func, TimeUnit):
     arg_types = {"this": True, "unit": True, "zone": False}
 
 
+class PartsToDate(Func):
```
tobymao, line 2414:
```suggestion
class DateFromParts(Func):
```

tobymao, line 2414:
```suggestion
class DateFromParts(Func):
    _sql_names = ["DATEFROMPARTS"]
```

sqlglot/dialects/tsql.py
```diff
@@ -104,6 +142,10 @@ class Parser(Parser):
         FUNCTIONS = {
             **Parser.FUNCTIONS,
             "CHARINDEX": exp.StrPosition.from_arg_list,
+            "DATEFROMPARTS": exp.PartsToDate.from_arg_list,
```
tobymao, line 145:
```suggestion
```

sqlglot/dialects/spark.py
```diff
@@ -111,6 +111,7 @@ class Generator(Hive.Generator):
             exp.StructKwarg: lambda self, e: f"{self.sql(e, 'this')}: {self.sql(e, 'expression')}",
             exp.TimestampTrunc: lambda self, e: f"DATE_TRUNC({self.sql(e, 'unit')}, {self.sql(e, 'this')})",
             exp.VariancePop: rename_func("VAR_POP"),
+            exp.PartsToDate: rename_func("MAKE_DATE"),
```
tobymao, line 114:
```suggestion
            exp.DateFromParts: rename_func("MAKE_DATE"),
```

Issue #597, opened 2022-10-15
DataFrame API Automated Testing for New Expressions


eakmanrq:
We want a way to make sure that as new expressions get added to sqlglot that the DataFrame API is also updated to use that expression instead of Anonymous functions. If we had a script that would be part of the test suit that would compare known expressions with currently anonymous functions then it could fail when there is a function that needs to be added to the DataFrame API.

tobymao:
we can probably parse the function and detect if you call invoke anonomyous on it, either that or you'll have to manually decorate each function with a custom decorator

PR #621, opened 2022-10-21
Adding support for several TSQL functions to Spark

GijsVoogd:
Adding support for several TSQL functions to Spark

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -189,6 +189,8 @@ class Hive(Dialect):
             "TO_DATE": format_time_lambda(exp.TsOrDsToDate, "hive"),
             "UNIX_TIMESTAMP": format_time_lambda(exp.StrToUnix, "hive", True),
             "YEAR": lambda args: exp.Year(this=exp.TsOrDsToDate.from_arg_list(args)),
+            "REPEAT": exp.Repeat.from_arg_list,
+            
         }
 
     class Generator(Generator):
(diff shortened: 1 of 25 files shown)


sqlglot/dialects/hive.py
```diff
@@ -245,6 +246,7 @@ class Generator(Generator):
             exp.TsOrDsAdd: lambda self, e: f"DATE_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
             exp.TsOrDsToDate: _to_date_sql,
             exp.TryCast: no_trycast_sql,
+            exp.Repeat: lambda self, e: f"REPEAT({self.format_args(e.this, e.args.get('times'))})",
```
tobymao, line 249: i don't think this is necessary

sqlglot/expressions.py
```diff
@@ -2488,7 +2488,7 @@ class Least(Func):
 
 
 class Length(Func):
-    pass
+    arg_types = {"this": True}
```
GeorgeSittas, line 2491: This is redundant -- (Func < Condition <) Expression already defines arg_types like this.

GeorgeSittas:
@GijsVoogd any plans to finish this?

GijsVoogd:
> @GijsVoogd any plans to finish this?

Yes, will work on it today

sqlglot/dialects/hive.py
GijsVoogd: removed it, thanks

sqlglot/expressions.py
GijsVoogd: removed it, thank you

GeorgeSittas:
The reason this test failed is documented [here](https://github.com/tobymao/sqlglot/issues/597). You can change sqlglot.dataframe.sql.functions:809 as follows to fix it:

```python
def repeat(col: ColumnOrName, n: int) -> Column:
    return Column.invoke_expression_over_column(col, glotexp.Repeat, times=lit(n))
```

GeorgeSittas (approved):
Looks good!

GijsVoogd:
> Looks good!

Thanks for the help George!

PR #625, opened 2022-10-22
Cleanup of DataFrame API Functions

eakmanrq:
Thanks to the test @tobymao added we can now identify when we are using an anonymous function when there is actually a SQLGlot function we should be using. This PR addresses all of those cases. 

In addition I also cheated a bit with the initial implementation of the functions with lambdas but assuming the variable names of the lambdas and then providing overrides to those that weren't compatible with the PySpark API. This removes all of that and also cleans up the lambda implementation a lot. 

Finally this also removes some excessive function calls.

### DIFF
--- a/tests/test_time.py
+++ b/tests/test_time.py
@@ -5,7 +5,7 @@ from sqlglot.time import format_time
 
 class TestTime(unittest.TestCase):
     def test_format_time(self):
-        self.assertEqual(format_time("", {}), "")
+        self.assertEqual(format_time("", {}), None)
         self.assertEqual(format_time(" ", {}), " ")
         mapping = {"a": "b", "aa": "c"}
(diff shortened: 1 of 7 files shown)


tests/test_time.py
```diff
@@ -5,7 +5,7 @@
 
 class TestTime(unittest.TestCase):
     def test_format_time(self):
-        self.assertEqual(format_time("", {}), "")
+        self.assertEqual(format_time("", {}), None)
```
eakmanrq, line 8: To me these seems like more proper behavior since we would want to treat an empty string like no value exists at all. This works well with other functions in the codebase where we check if arguments are None and if so we remove them from the output.

tobymao (approved):

PR #629, opened 2022-10-24
extend dateadd tsql/hive

TimothyAerts:
Noticed the dateadd addition and I was working on this part as well.
Some extensions to the already implemented, similar problems to other date functions in tsql arise here.

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -18,6 +18,24 @@ from sqlglot.helper import list_get
 from sqlglot.parser import Parser, parse_var_map
 from sqlglot.tokens import Tokenizer
 
+# (FuncType, Multiplier)
+DATE_DELTA_INTERVAL = {
+    "YEAR": ("ADD_MONTHS", 12),
+    "MONTH": ("ADD_MONTHS", 1),
+    "QUARTER": ("ADD_MONTHS", 4),
+    "WEEK": ("DATE_ADD", 7),
+    "DAY": ("DATE_ADD", 1),
+}
+
+
+def _add_date(self, expression):
+    unit = expression.text("unit").upper()
+    func, multiplier = DATE_DELTA_INTERVAL.get(unit, ("DATE_ADD", 1))
+    modified_increment = (
+        int(expression.text("expression")) * multiplier if expression.expression.is_number else expression.expression
+    )
+    return f"{func}({self.sql(expression, 'this')}, {modified_increment})"
+
 
 def _array_sort(self, expression):
     if expression.expression:
@@ -211,7 +229,7 @@ class Hive(Dialect):
             exp.ArraySize: rename_func("SIZE"),
             exp.ArraySort: _array_sort,
             exp.With: no_recursive_cte_sql,
-            exp.DateAdd: lambda self, e: f"DATE_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
+            exp.DateAdd: _add_date,
             exp.DateDiff: lambda self, e: f"DATEDIFF({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
             exp.DateStrToDate: rename_func("TO_DATE"),
             exp.DateToDi: lambda self, e: f"CAST(DATE_FORMAT({self.sql(e, 'this')}, {Hive.dateint_format}) AS INT)",
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -22,7 +22,7 @@ from sqlglot.tokens import Tokenizer
 DATE_DELTA_INTERVAL = {
     "YEAR": ("ADD_MONTHS", 12),
     "MONTH": ("ADD_MONTHS", 1),
-    "QUARTER": ("ADD_MONTHS", 4),
+    "QUARTER": ("ADD_MONTHS", 3),
     "WEEK": ("DATE_ADD", 7),
     "DAY": ("DATE_ADD", 1),
 }
@@ -35,6 +35,16 @@ def _add_date(self, expression):
         int(expression.text("expression")) * multiplier if expression.expression.is_number else expression.expression
     )
     return f"{func}({self.sql(expression, 'this')}, {modified_increment})"
+
+
+def _date_diff(self, expression):
+    unit = expression.text("unit").upper()
+    sql_func = "MONTHS_BETWEEN" if unit == "MONTH" else "DATEDIFF"
+    _, multiplier = DATE_DELTA_INTERVAL.get(unit, ("", 1))
+    if multiplier > 1:
+        return f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')}) / {multiplier}"
+    else:
+        return f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')})"
 
 
 def _array_sort(self, expression):
@@ -230,7 +240,7 @@ class Hive(Dialect):
             exp.ArraySort: _array_sort,
             exp.With: no_recursive_cte_sql,
             exp.DateAdd: _add_date,
-            exp.DateDiff: lambda self, e: f"DATEDIFF({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
+            exp.DateDiff: _date_diff,
             exp.DateStrToDate: rename_func("TO_DATE"),
             exp.DateToDi: lambda self, e: f"CAST(DATE_FORMAT({self.sql(e, 'this')}, {Hive.dateint_format}) AS INT)",
             exp.DiToDate: lambda self, e: f"TO_DATE(CAST({self.sql(e, 'this')} AS STRING), {Hive.dateint_format})",
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -37,14 +37,13 @@ def _add_date(self, expression):
     return f"{func}({self.sql(expression, 'this')}, {modified_increment})"
 
 
-def _date_diff(self, expression):
+def _date_diff_sql(self, expression):
     unit = expression.text("unit").upper()
-    sql_func = "MONTHS_BETWEEN" if unit == "MONTH" or unit == "QUARTER" else "DATEDIFF"
+    sql_func = "MONTHS_BETWEEN" if unit in ["YEAR", "QUARTER", "MONTH"] else "DATEDIFF"
     _, multiplier = DATE_DELTA_INTERVAL.get(unit, ("", 1))
-    if multiplier > 1:
-        return f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')}) / {multiplier}"
-    else:
-        return f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')})"
+    multiplier_sql = f" / {multiplier}" if multiplier > 1 else ""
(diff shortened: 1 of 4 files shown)


TimothyAerts:
Not perfect, since nothing with lower granularity than days is addressed.
Figured it would be an ok start like this.

tests/dialects/test_tsql.py
```diff
@@ -243,3 +243,28 @@ def test_convert_date_format(self):
...
+        self.validate_all("SELECT DATEADD(qq, 1, '2017/08/25')", write={"spark": "SELECT ADD_MONTHS('2017/08/25', 3)"})
+        self.validate_all("SELECT DATEADD(wk, 1, '2017/08/25')", write={"spark": "SELECT DATE_ADD('2017/08/25', 7)"})
+
+    def test_date_diff(self):
+        self.validate_identity("SELECT DATEDIFF(year, '2021/01/01', '2020/01/01')")
+        self.validate_all(
+            "SELECT DATEDIFF(year, '2021/01/01', '2020/01/01')",
+            write={
+                "tsql": "SELECT DATEDIFF(year, '2021/01/01', '2020/01/01')",
+                "spark": "SELECT DATEDIFF('2020/01/01', '2021/01/01') / 12",
```
GeorgeSittas, line 261: Are you sure this is correct? Why do we divide by 12? Shouldn't the end date come first? Check [this](https://spark.apache.org/docs/2.3.0/api/sql/index.html#datediff).

sqlglot/dialects/hive.py
```diff
@@ -18,6 +18,34 @@
...
+
+def _date_diff(self, expression):
+    unit = expression.text("unit").upper()
+    sql_func = "MONTHS_BETWEEN" if unit == "MONTH" or unit == "QUARTER" else "DATEDIFF"
+    _, multiplier = DATE_DELTA_INTERVAL.get(unit, ("", 1))
+    if multiplier > 1:
+        return f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')}) / {multiplier}"
+    else:
+        return f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')})"
+
```
GeorgeSittas, line 48: Let's rename this to _date_diff_sql. Also, maybe it can be simplified to:
```python
def _date_diff_sql(self, expression):
    unit = expression.text("unit").upper()
    sql_func = "MONTHS_BETWEEN" if unit == "MONTH" or unit == "QUARTER" else "DATEDIFF"
    _, multiplier = DATE_DELTA_INTERVAL.get(unit, ("", 1))
    multiplier_sql = f" / {multiplier}" if multiplier > 1 else ""
    diff_sql = f"{sql_func}({self.sql(expression, 'this')}, {self.sql(expression, 'expression')})"
    return f"{diff_sql}{multiplier_sql}"
```

sqlglot/dialects/tsql.py
```diff
@@ -25,6 +42,14 @@ def _format_time(args):
     return _format_time
 
 
+def parse_date_delta(args, kind="ADD"):
+    unit = DATE_DELTA_INTERVAL.get(list_get(args, 0).text("this").lower(), "day")
+    if kind == "SUB":
+        return exp.DateDiff(this=list_get(args, 2), expression=list_get(args, 1), unit=unit)
+    else:
+        return exp.DateAdd(this=list_get(args, 2), expression=list_get(args, 1), unit=unit)
```
GeorgeSittas, line 50: Are we sure that calling a method on list_get(...) won't cause an exception? Also, maybe we can simplify this to:
```python
def parse_date_delta(args, kind="ADD"):
    unit = DATE_DELTA_INTERVAL.get(list_get(args, 0).name.lower(), "day")
    exp_class = exp.DateDiff if kind == "SUB" else exp.DateAdd
    return exp_class(this=list_get(args, 2), expression=list_get(args, 1), unit=unit)
```

tests/dialects/test_dialect.py
```diff
@@ -437,13 +437,13 @@ def test_time(self):
...
-            "DATE_ADD(x, y, 'day')",
+            "DATE_ADD(x, y, day)",
             read={
-                "tsql": "DATEADD('day', y, x)",
+                "tsql": "DATEADD(day, y, x)",
             },
             write={
                 "postgres": UnsupportedError,
-                "tsql": "DATEADD('day', y, x)",
+                "tsql": "DATEADD(day, y, x)",
```
GeorgeSittas, line 446: Why were these tests changed?

GeorgeSittas (dismissed):

tests/dialects/test_dialect.py
TimothyAerts: They seemed like TSQL tests where the parameters are not quoted. 
SELECT DATEDIFF(year, '2017/08/25', '2011/08/25') AS DateDiff;

GeorgeSittas:
I didn't mean to submit the review as a request, I just left some comments we can think about.

tests/dialects/test_tsql.py
TimothyAerts: SELECT DATEDIFF(year, '2011/08/25', '2017/08/25') AS DateDiff; => 6
SELECT DATEDIFF('2011/08/25', '2017/08/25') => -1292

Where I am wrong and where the 12 confusion comes from is that this should convert into months_between

tests/dialects/test_dialect.py
GeorgeSittas: Let's move it to test_tsql then (or even better, let's try to check tsql in test_time's "DATE_ADD(x, 1, 'day')" case).

sqlglot/dialects/tsql.py
TimothyAerts: DATEDIFF(interval, date1, date2)
DATEADD (datepart , number , date )  

Both always should have 3 parameters (as far as I can tell from the docs) thus having it crash on an exception if list_get fails might not be the worst?

sqlglot/dialects/hive.py
```diff
@@ -18,6 +18,33 @@
...
+DATE_DELTA_INTERVAL = {
+    "YEAR": ("ADD_MONTHS", 12),
+    "MONTH": ("ADD_MONTHS", 1),
+    "QUARTER": ("ADD_MONTHS", 3),
+    "WEEK": ("DATE_ADD", 7),
+    "DAY": ("DATE_ADD", 1),
+}
+
+
+def _add_date(self, expression):
```
GeorgeSittas, line 31: Let's add a _sql suffix to this one too.

sqlglot/dialects/tsql.py
```diff
@@ -143,9 +166,8 @@ class Parser(Parser):
             **Parser.FUNCTIONS,
             "CHARINDEX": exp.StrPosition.from_arg_list,
             "ISNULL": exp.Coalesce.from_arg_list,
-            "DATEADD": lambda args: exp.DateAdd(
-                this=list_get(args, 2), expression=list_get(args, 1), unit=list_get(args, 0)
-            ),
+            "DATEADD": lambda args: parse_date_delta(args, kind="ADD"),
+            "DATEDIFF": lambda args: parse_date_delta(args, kind="SUB"),
```
GeorgeSittas, line 170: We can further simplify by just passing in the expression class directly here, instead of comparing strings.

GeorgeSittas: Also, there are some similar examples in bigquery, mysql (_date_add) you can check out.

sqlglot/dialects/tsql.py
GeorgeSittas: I think it'd be better to notify the user that we're missing a third argument instead of throwing an AttributeError.

tests/dialects/test_dialect.py
TimothyAerts: Gave it a go, let me know if you think this suffices

sqlglot/dialects/tsql.py
TimothyAerts: Something like this? or should the error be thrown from a different location?
I see very little dialects implementing errors.
```
def parse_date_delta(exp_class):
    def inner_func(args):
        if not list_get(args, 2):
          raise IndexError(f"missing parameters for class: {exp_class}. 3 parameters are expected")
        unit = DATE_DELTA_INTERVAL.get(list_get(args, 0).name.lower(), "day")
        return exp_class(this=list_get(args, 2), expression=list_get(args, 1), unit=unit)

    return inner_func
```

sqlglot/dialects/hive.py
```diff
@@ -18,6 +18,33 @@
...
+    func, multiplier = DATE_DELTA_INTERVAL.get(unit, ("DATE_ADD", 1))
+    modified_increment = (
+        int(expression.text("expression")) * multiplier if expression.expression.is_number else expression.expression
+    )
+    return f"{func}({self.sql(expression, 'this')}, {modified_increment})"
+
+
+def _date_diff_sql(self, expression):
+    unit = expression.text("unit").upper()
+    sql_func = "MONTHS_BETWEEN" if unit in ["YEAR", "QUARTER", "MONTH"] else "DATEDIFF"
```
GeorgeSittas, line 42: List can be converted to a set, and factored out as a constant maybe?

sqlglot/dialects/tsql.py
```diff
@@ -205,6 +234,7 @@ class Generator(Generator):
 
         TRANSFORMS = {
             **Generator.TRANSFORMS,
-            exp.DateAdd: lambda self, e: f"DATEADD({self.format_args(e.args.get('unit'), e.expression, e.this)})",
+            exp.DateAdd: lambda self, e: generate_date_delta(self, e),
+            exp.DateDiff: lambda self, e: f"DATEDIFF({self.format_args(e.args.get('unit'), e.expression, e.this)})",
```
GeorgeSittas, line 238: We should be able to reuse generate_date_delta for DateDiff too, right? Also, maybe rename generate_date_delta --> _date_delta_sql for consistency.
```python
exp.DateAdd: generate_date_delta,
```

sqlglot/dialects/tsql.py
GeorgeSittas: I think we can raise here. What do you say @tobymao?

sqlglot/dialects/hive.py
```diff
@@ -18,6 +18,33 @@
...
+}
+
+
+def _add_date_sql(self, expression):
+    unit = expression.text("unit").upper()
+    func, multiplier = DATE_DELTA_INTERVAL.get(unit, ("DATE_ADD", 1))
+    modified_increment = (
+        int(expression.text("expression")) * multiplier if expression.expression.is_number else expression.expression
+    )
+    return f"{func}({self.sql(expression, 'this')}, {modified_increment})"
```
tobymao, line 37: all funcs should use self.format_args(expression.this, modified_increment)

sqlglot/dialects/tsql.py
TimothyAerts: Yes you're right

sqlglot/dialects/hive.py
TimothyAerts: `   python 
return f"{func}({self.format_args(e.text('unit'), e.expression, e.this)})"
`
Is that ok with the e.text?

tobymao: it might be but you can just do e.args.get("unit")

TimothyAerts: It takes the quotes that way I thought, which would be incorrect for TSQL.

TimothyAerts: ran quick test:
image

tobymao (approved):

PR #634, opened 2022-10-24
Add (Azure) Databricks dialect

TimothyAerts:
I'm currently trying to go towards databricks and while dealing with datediff and date_add noticed that databricks is not limited to just the spark sql language.
It has implemented some functions in the TSQL style. DATEDIFF(unit, int, datetime) and DATEDIFF(datetime, datetime) both exist in databricks. As well as DATE_ADD(datetime, int) and DATEADD(unit, int, datetime). 

The spark transpiler should not deal with this since the TSQL like functions do not exist in spark (as far as I am aware of).
Spark Date functions : https://spark.apache.org/docs/2.3.0/api/sql/search.html?q=date
Databricks TSQL DATEDIFF: https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/functions/datediff3
Datebricks spark DATEDIFF: https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/functions/datediff

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -1,28 +1,9 @@
 from sqlglot import exp
+from sqlglot.dialects.dialect import generate_tsql_date_delta, parse_date_delta
 from sqlglot.dialects.spark import Spark
-from sqlglot.helper import list_get
-
-
-def generate_date_delta(self, e):
-    func = "DATEADD" if isinstance(e, exp.DateAdd) else "DATEDIFF"
-    return f"{func}({self.format_args(e.text('unit'), e.expression, e.this)})"
-
-
-def parse_date_delta(exp_class):
-    def inner_func(args):
-        spark_func = len(args) == 2
-        this = list_get(args, 0) if spark_func else list_get(args, 2)
-        expression = list_get(args, 1) if spark_func else list_get(args, 1)
-        unit = exp.Literal.string("DAY") if spark_func else list_get(args, 0)
-        return exp_class(this=this, expression=expression, unit=unit)
-
-    return inner_func
 
 
 class Databricks(Spark):
-    class Tokenizer(Spark.Tokenizer):
-        pass
-
     class Parser(Spark.Parser):
         FUNCTIONS = {
             **Spark.Parser.FUNCTIONS,
@@ -34,6 +15,6 @@ class Databricks(Spark):
     class Generator(Spark.Generator):
         TRANSFORMS = {
             **Spark.Generator.TRANSFORMS,
-            exp.DateAdd: lambda self, e: generate_date_delta(self, e),
-            exp.DateDiff: lambda self, e: generate_date_delta(self, e),
+            exp.DateAdd: lambda self, e: generate_tsql_date_delta(self, e),
+            exp.DateDiff: lambda self, e: generate_tsql_date_delta(self, e),
         }
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -1,0 +1,17 @@
+from sqlglot import exp
+from sqlglot.dialects.spark import Spark
+
+
+class Databricks(Spark):
+    class Tokenizer(Spark.Tokenizer):
+        pass
+
+    class Parser(Spark.Parser):
+        FUNCTIONS = {**Spark.Parser.FUNCTIONS}
+
+    class Generator(Spark.Generator):
+        TRANSFORMS = {
+            **Spark.Generator.TRANSFORMS,
+            exp.DateAdd: lambda self, e: f"DATE_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
+            exp.DateDiff: lambda self, e: f"DATE_DIFF({self.sql(e, 'this')}, {self.sql(e, 'expression')})",
+        }
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -1,5 +1,6 @@
 from sqlglot import exp
 from sqlglot.dialects.spark import Spark
+from sqlglot.helper import list_get
 
 
 class Databricks(Spark):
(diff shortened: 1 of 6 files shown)


tobymao:
this is not pure databricks though, i assume this is the azure special databricks? does this also apply to the non azure one?

sqlglot/dialects/databricks.py
```diff
@@ -0,0 +1,39 @@
+from sqlglot import exp
+from sqlglot.dialects.spark import Spark
+from sqlglot.helper import list_get
+
+
+def generate_date_delta(self, e):
```
tobymao, line 6: can you dry this up with the tsql stuff? put it into dialect.py or a shared module

sqlglot/dialects/databricks.py
```diff
@@ -0,0 +1,39 @@
...
+    func = "DATEADD" if isinstance(e, exp.DateAdd) else "DATEDIFF"
+    return f"{func}({self.format_args(e.text('unit'), e.expression, e.this)})"
+
+
+def parse_date_delta(exp_class):
```
tobymao, line 11: dry this up too

sqlglot/dialects/databricks.py
```diff
@@ -0,0 +1,39 @@
...
+        this = list_get(args, 0) if spark_func else list_get(args, 2)
+        expression = list_get(args, 1) if spark_func else list_get(args, 1)
+        unit = exp.Literal.string("DAY") if spark_func else list_get(args, 0)
+        return exp_class(this=this, expression=expression, unit=unit)
+
+    return inner_func
+
+
+class Databricks(Spark):
+    class Tokenizer(Spark.Tokenizer):
```
tobymao, line 23: this isn't needed?

sqlglot/dialects/databricks.py
TimothyAerts: It is really only applicable to TSQL, Databricks. Would putting it in Dialect not be setting false expectations?

sqlglot/dialects/databricks.py
TimothyAerts: Moved to Dialect, would be able to replace the lambda in dialects such as Presto, but it's lacking tests so didn't want to go for it in this pr.

TimothyAerts:
On the Azure vs generic part, I was unsure on whether this was Azure specific. I added it in the title since this is the only environment where I actually test the functions. However, looking at the generic databricks documentation we see both datadiffs aswell:
https://docs.databricks.com/spark/latest/spark-sql/language-manual/functions/datediff3.html

On top of that I think these implementations are tied to a runtime version, which I imagine are equal across platforms.

sqlglot/dialects/dialect.py
```diff
@@ -331,3 +332,20 @@ def create_with_partitions_sql(self, expression):
             expression.set("this", schema)
 
     return self.create_sql(expression)
+
+
+def generate_tsql_date_delta(self, e):
```
GeorgeSittas, line 337: Rename this so that we have the _sql suffix like other functions in this file? Also I don't think the "tsql" part is necessary, since this is used by others dialects too and may create confusion later on.

sqlglot/dialects/dialect.py
```diff
@@ -331,3 +332,20 @@ def create_with_partitions_sql(self, expression):
...
+
+
+def parse_date_delta(exp_class, unit_mapping=None):
+    def inner_func(args):
+        tsql_func = len(args) == 3
+        this = list_get(args, 2) if tsql_func else list_get(args, 0)
+        expression = list_get(args, 1) if tsql_func else list_get(args, 1)
+        unit = list_get(args, 0) if tsql_func else exp.Literal.string("DAY")
+        unit = unit_mapping.get(unit.name.lower(), unit) if unit_mapping else unit
+        return exp_class(this=this, expression=expression, unit=unit)
```
GeorgeSittas, line 349: I'm a little skeptical about this.. I think we don't want this to handle TSQL as a separate case, since no other function in the `dialects.py` file does something similar for other dialects. Maybe we should somehow make the logic / names more generic instead of the current `tsql_func` stuff?

TimothyAerts: I see your point.
However, this parse would work as a default for some other dialects for example presto could use this function instead of the lambda it has.  All other dialects use the default parser.

GeorgeSittas: If most dialects use the default version of this, can't we simplify this by calling it inside tsql and rearranging the arguments of the produced expression or something?

I'm not sure if that's the best approach, we can wait for Toby's response on this.

sqlglot/dialects/dialect.py
TimothyAerts: I am looking for a way to indicate that this is the 3 parameter version without having to fully type generate_date_delta_three_parameters. Any suggestions?
In my mind the TSQL was to indicate the tsql 3 parameter style, but you are right that it is confusing.

sqlglot/dialects/dialect.py
TimothyAerts: Thats how it was, but Databricks needs to do it the exact same way (and presto aswell).
So it is not only about tsql anymore.
Databricks is special since it has the option for the 2 parameter and 3 parameter function.

sqlglot/dialects/dialect.py
GeorgeSittas: How does `date_delta_with_unit_sql` look?

TimothyAerts: I like it, should have thought of that, thank you. as for the placement, I agree this will not be used by many dialects since the other 3 param dialects have different structure as far as I could tell. So only databricks and tsql would use this.

sqlglot/dialects/databricks.py
tobymao: you can import from tsql

sqlglot/dialects/dialect.py
tobymao: yea the function shouldn't use the word tsql

tobymao (approved):

PR #636, opened 2022-10-24
Add using for join builder, update some comments

### LINKED ISSUE #633: Feature: Add support for USING with join
I see it in arg types, but I'm not sure how to leverage it.

```
 sg.select(
    PARTITION.quote(),
    PATIENT_ID,
    OR_CASE_ID
).from_('merged_df').join('vte_diagnosis_df', using='patient_id, encounter_id')
```

GeorgeSittas:
Fixes #633

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1633,7 +1633,7 @@ class Select(Subqueryable):
             >>> Select().select("*").from_("tbl").join("tbl2", on="tbl1.y = tbl2.y").sql()
             'SELECT * FROM tbl JOIN tbl2 ON tbl1.y = tbl2.y'
 
-            >>> Select().select("1").from_("a").join("b", using="x, y, z").sql()
+            >>> Select().select("1").from_("a").join("b", using=["x", "y", "z"]).sql()
             'SELECT 1 FROM a JOIN b USING (x, y, z)'
 
             Use `join_type` to change the type of join:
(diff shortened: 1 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -1641,6 +1683,11 @@ def join(
             on = and_(*ensure_list(on), dialect=dialect, **opts)
             join.set("on", on)
 
+        if using:
+            if isinstance(using, str):
+                using = [to_identifier(name) for name in using.split(", ")]
```
barakalon, line 1688: Why are we accepting a CSV instead of a list of strings?

sqlglot/expressions.py
```diff
@@ -937,6 +937,42 @@ def on(self, *expressions, append=True, dialect=None, copy=True, **opts):
 
         return join
 
+    def using(self, *expressions, append=True, dialect=None, copy=True, **opts):
```
barakalon, line 940: This isn't tested.

GeorgeSittas: Thanks for pointing out, I forgot to add more tests after the docstring.

sqlglot/expressions.py
GeorgeSittas: Right, I'll fix this too. Didn't notice that's what we were doing for `on` due to docstring type signature confusion. Thanks!

tobymao: you can call ensure_list

sqlglot/expressions.py
```diff
@@ -1641,6 +1683,11 @@ def join(
             on = and_(*ensure_list(on), dialect=dialect, **opts)
             join.set("on", on)
 
+        if using:
+            if isinstance(using, str):
```
tobymao, line 1687: what if it's a list of strings?, ensure list should be first

GeorgeSittas: Can we also accept a list of expressions? Is that a case we should consider?

barakalon: I guess someone could pass an Identifier or a list of Identifiers, so yeah?

tobymao (approved):

PR #641, opened 2022-10-26
TSQL support for lateral joins

10bas10:
Added support for the CROSS APPLY and OUTER APPLY from T-SQL. In order to achieve this I had to add the "view" attribute to make the distinction between LATERAL VIEW and LATERAL JOINS.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -166,6 +166,9 @@ class TSQL(Dialect):
             "SQL_VARIANT": TokenType.VARIANT,
             "NVARCHAR(MAX)": TokenType.TEXT,
             "VARCHAR(MAX)": TokenType.TEXT,
+            "APPLY": TokenType.LATERAL,
+            "OUTER": TokenType.OUTER,
+            "LEFT OUTER": TokenType.LEFT,
         }
 
     class Parser(Parser):
@@ -225,6 +228,49 @@ class TSQL(Dialect):
 
             # Entails a simple cast without any format requirement
             return self.expression(exp.Cast if strict else exp.TryCast, this=this, to=to)
+
+        def _parse_lateral(self):
+            if self._match_pair(TokenType.CROSS, TokenType.LATERAL):
+                side = None
+                kind = "INNER"
+                outer = False
+
+            elif self._match_pair(TokenType.OUTER, TokenType.LATERAL):
+                side = "LEFT"
+                kind = "OUTER"
+                outer = True
+            else:
+                return None
+
+            subquery = self._parse_select(table=True)
+
+            if subquery:
+                return self.expression(exp.Join, this=self.expression(exp.Lateral, this=subquery, view=False), side=side, kind=kind)
+
+            # Functions
+            func = self._parse_function()
+            table_alias = self._parse_id_var(any_token=False)
+
+            if self._match(TokenType.L_PAREN):
+                columns = self._parse_csv(self._parse_id_var)
+                self._match(TokenType.R_PAREN)
+
+            return self.expression(
+                exp.Join,
+                this = self.expression(
+                    exp.Lateral,
+                    this=func,
+                    outer=outer,
+                    view=False,
+                    alias=self.expression(
+                        exp.TableAlias,
+                        this=table_alias,
+                        columns=columns,
+                    ),
+                ),
+                side=side,
+                kind=kind,
+            )
 
     class Generator(Generator):
         TYPE_MAPPING = {
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -222,6 +222,7 @@ class TSQL(Dialect):
             return self.expression(exp.Cast if strict else exp.TryCast, this=this, to=to)
 
         def _parse_lateral(self):
+            # Determine the type of join
             if self._match_pair(TokenType.CROSS, TokenType.LATERAL):
                 side = None
                 kind = "INNER"
@@ -234,15 +235,16 @@ class TSQL(Dialect):
             else:
                 return None
 
+            # Determine whether a subquery is applicable
             subquery = self._parse_select(table=True)
-
             if subquery:
                 return self.expression(exp.Join, this=self.expression(exp.Lateral, this=subquery, view=False), side=side, kind=kind)
 
-            # Functions
+            # If no subquery applicable, a function applies
             func = self._parse_function()
             table_alias = self._parse_id_var(any_token=False)
 
+            # Check for any explicit column specifications
             if self._match(TokenType.L_PAREN):
                 columns = self._parse_csv(self._parse_id_var)
                 self._match(TokenType.R_PAREN)
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -237,7 +237,7 @@ class TSQL(Dialect):
             subquery = self._parse_select(table=True)
             if subquery:
                 return self.expression(
-                    exp.Join, this=self.expression(exp.Lateral, this=subquery, view=False), side=side, kind=kind
+                    exp.Join, this=self.expression(exp.Lateral, this=subquery), side=side, kind=kind
                 )
 
             # If no subquery applicable, a function applies
@@ -256,7 +256,6 @@ class TSQL(Dialect):
                     exp.Lateral,
                     this=func,
                     outer=outer,
-                    view=False,
                     alias=self.expression(
                         exp.TableAlias,
                         this=table_alias,
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -247,6 +247,7 @@ class TSQL(Dialect):
             table_alias = self._parse_id_var(any_token=False)
 
             # Check for any explicit column specifications
+            columns = None
             if self._match(TokenType.L_PAREN):
                 columns = self._parse_csv(self._parse_id_var)
                 self._match(TokenType.R_PAREN)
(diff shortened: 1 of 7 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -158,6 +158,9 @@ class Tokenizer(Tokenizer):
             "SQL_VARIANT": TokenType.VARIANT,
             "NVARCHAR(MAX)": TokenType.TEXT,
             "VARCHAR(MAX)": TokenType.TEXT,
+            "APPLY": TokenType.LATERAL,
+            "OUTER": TokenType.OUTER,
+            "LEFT OUTER": TokenType.LEFT,
```
tobymao, line 163: outer and left outer are already tokens, please remove them

sqlglot/parser.py
```diff
@@ -1173,14 +1173,15 @@ def _parse_lateral(self):
         subquery = self._parse_select(table=True)
 
         if subquery:
-            return self.expression(exp.Lateral, this=subquery)
+            return self.expression(exp.Lateral, this=subquery, view=False)
 
-        self._match(TokenType.VIEW)
+        view = True if self._match(TokenType.VIEW) else False
```
tobymao, line 1178:
```suggestion
        view = self._match(TokenType.VIEW) 
```

sqlglot/dialects/tsql.py
```diff
@@ -218,6 +221,54 @@ def _parse_convert(self, strict):
             # Entails a simple cast without any format requirement
             return self.expression(exp.Cast if strict else exp.TryCast, this=this, to=to)
 
+        def _parse_lateral(self):
```
tobymao, line 224: are you sure this is necessary? can we just add a simplified version of this to the generic parser?

sqlglot/dialects/tsql.py
10bas10: Ah yes, removed the dialect specific tokens

sqlglot/parser.py
10bas10: Made the property optional

sqlglot/expressions.py
```diff
@@ -975,7 +975,7 @@ def using(self, *expressions, append=True, dialect=None, copy=True, **opts):
 
 
 class Lateral(UDTF):
-    arg_types = {"this": True, "outer": False, "alias": False}
+    arg_types = {"this": True, "view": True, "outer": False, "alias": False}
```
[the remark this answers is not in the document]
10bas10: Made the property optional

sqlglot/dialects/tsql.py
tobymao: in general, we try to avoid overriding core methods and putting logic into the generic parser.

the sqlglot parser is supposed to be a superset of as much sql as possible.

10bas10: Would love your input on this. Problems i encountered were:
- CROSS / OUTER are the tokens to identify that its a join. This however should take place in combination of the token of APPLY, as you also have CROSS JOIN. 
- Also the parsing of column specifications in T-SQL is enclosed by parenthesis instead of separated by an alias

If you have any suggestions for me into another direction, im curious to find out.

tobymao: right now you're converting apply to lateral at tokenize time, just keep it as an apply and then you can see when that happens to differentiate.

additionally, we already handle column spec (table alias) parsing i believe

CROSS APPLY (SELECT v, y FROM t) t(v, y)

so for example t(v, y)

if apply, then instead of t AS x, then look for t with table alias

10bas10: I will have a look whether i can integrate it in the generic parser.

sqlglot/generator.py
```diff
@@ -716,15 +716,21 @@ def lambda_sql(self, expression):
...
         table = alias.name
         table = f" {table}" if table else table
         columns = self.expressions(alias, key="columns", flat=True)
         columns = f" AS {columns}" if columns else ""
-        return f"{op_sql}{self.sep()}{this}{table}{columns}"
+
+        if expression.args.get("view"):
+            op_sql = self.seg(f"LATERAL VIEW{' OUTER' if expression.args.get('outer') else ''}")
+            return f"{op_sql}{self.sep()}{this}{table}{columns}"
+        else:
```
tobymao, line 732: no else here

sqlglot/parser.py
```diff
@@ -1167,28 +1167,61 @@ def _parse_from(self):
         return self.expression(exp.From, expressions=self._parse_csv(self._parse_table))
 
     def _parse_lateral(self):
-        if not self._match(TokenType.LATERAL):
-            return None
+        index = self._index
+        if self._match_set([TokenType.OUTER, TokenType.CROSS]):
```
tobymao, line 1171: pass in a real set or tuple {x, y}

sqlglot/parser.py
```diff
@@ -1167,28 +1167,61 @@ def _parse_from(self):
...
-                columns=(self._parse_csv(self._parse_id_var) if self._match(TokenType.ALIAS) else None),
+                this=table_alias,
+                columns=columns,
             ),
         )
 
+        if lat_type == TokenType.LATERAL:
+            return expression
+        if lat_type == TokenType.APPLY:
+            return self.expression(
```
tobymao, line 1218: missing if block? maybe lateral is default?

sqlglot/parser.py
```diff
@@ -1167,28 +1167,61 @@ def _parse_from(self):
...
+        view = self._match(TokenType.VIEW)
         outer = self._match(TokenType.OUTER)
 
-        return self.expression(
+        if not this:
+            this = self._parse_function()
+
+        table_alias = self._parse_id_var(any_token=False)
+        columns = None
+        if self._match_set([TokenType.ALIAS, TokenType.L_PAREN]):
```
tobymao, line 1197: i think this could probably get cleaned up

sqlglot/tokens.py
```diff
@@ -442,6 +443,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "CONSTRAINT": TokenType.CONSTRAINT,
         "CREATE": TokenType.CREATE,
         "CROSS": TokenType.CROSS,
+        "APPLY": TokenType.APPLY,
```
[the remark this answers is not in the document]
10bas10: Moved it to the T-SQL dialect

sqlglot/generator.py
10bas10: The else is now default

sqlglot/parser.py
10bas10: Changed

sqlglot/parser.py
```diff
@@ -1167,28 +1167,61 @@ def _parse_from(self):
...
-            return None
+        index = self._index
+        if self._match_set([TokenType.OUTER, TokenType.CROSS]):
+            cross = self._prev.token_type == TokenType.CROSS
+            kind = "inner" if cross else "outer"
+            side = None if cross else "left"
+            outer = not cross
 
-        subquery = self._parse_select(table=True)
+            if not self._match(TokenType.APPLY):
```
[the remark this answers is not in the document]
10bas10: This approach is required due to the fact that i need to acquire the token whether its an CROSS or an OUTER. Thats why i did not implement a match_pair.

sqlglot/parser.py
10bas10: Changed it so lateral is default.

sqlglot/parser.py
10bas10: What do you mean by this? Would you rewrite this in a different manner?

sqlglot/dialects/tsql.py
```diff
@@ -158,6 +158,7 @@ class Tokenizer(Tokenizer):
             "SQL_VARIANT": TokenType.VARIANT,
             "NVARCHAR(MAX)": TokenType.TEXT,
             "VARCHAR(MAX)": TokenType.TEXT,
+            "APPLY": TokenType.APPLY,
```
tobymao, line 161: really sorry about the swirl, but i think we can add this to the generic tokenizer but we neeed to in the parser add 

APPLY to ID_VAR_TOKENS and perhaps TABLE_ALIAS_TOKENS like NATURAL

sqlglot/parser.py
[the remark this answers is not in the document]
tobymao: then why not do

cross_apply = self._match_pair(cross, apply)
outer_apply = ...

sqlglot/parser.py
tobymao: yes, but i don't have time to dig in deeply perhaps @GeorgeSittas can take a look 

maybe tonight

GeorgeSittas: Yep, wrapping up some stuff and will take a closer look to see if I can help.

sqlglot/parser.py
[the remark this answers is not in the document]
10bas10: Changed it to a match_pair and in that way no retreat is also needed.

sqlglot/parser.py
10bas10: Rewrote the part, let me know what you think.

sqlglot/dialects/tsql.py
10bas10: No problem, moved it back to the generic tokenizer.

GeorgeSittas (approved):

sqlglot/parser.py
GeorgeSittas: Looks good.

tobymao:
great, except i'm going to remove the "inner" join as default is just join

tobymao:
thanks again for the great pr @10bas10

tobymao:
@10bas10 https://github.com/tobymao/sqlglot/commit/bd3f07d4f1b1f4659553472849fcfeb2a55aabac

tobymao:
i believe this also fixes a bug because in the previous code, the OUTER was always getting set to False since you always did outer = self._match(TokenType.OUTER)

PR #642, opened 2022-10-26
Annotate SQLGlot with python types (WIP)

GeorgeSittas:
Opening this PR to start a discussion about adding type annotations throughout SQLGlot. I don't have a lot of experience with typing in python, so there may be some errors on the way. Any feedback is greatly appreciated!

As an extra, I'll try to polish/add documentation and maybe refactor some bits of code that can be improved.

### DIFF
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -1,5 +1,7 @@
 """## Python SQL parser, transpiler and optimizer."""
 
+from __future__ import annotations
+
 import typing as t
 
 from sqlglot import expressions as exp
@@ -52,7 +54,7 @@ def parse(sql: str, read: t.Optional[str] = None, **opts) -> t.List[t.Optional[E
 
 
 def parse_one(
-    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression] = None, **opts
+    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression | str] = None, **opts
 ) -> t.Optional[Expression]:
     """
     Parses the given SQL string and returns a syntax tree for the first parsed SQL statement.
@@ -83,7 +85,7 @@ def transpile(
     write: t.Optional[str] = None,
     identity: bool = True,
     error_level: t.Optional[ErrorLevel] = None,
-    **opts
+    **opts,
 ) -> t.List[str]:
     """
     Parses the given SQL string in accordance with the source dialect and returns a list of SQL strings transformed
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -37,7 +37,7 @@ pretty = False
 schema = MappingSchema()
 
 
-def parse(sql: str, read: t.Optional[str] = None, **opts) -> t.List[t.Optional[Expression]]:
+def parse(sql: str, read: t.Optional[str | Dialect] = None, **opts) -> t.List[t.Optional[Expression]]:
     """
     Parses the given SQL string into a collection of syntax trees, one per parsed SQL statement.
 
@@ -54,7 +54,7 @@ def parse(sql: str, read: t.Optional[str] = None, **opts) -> t.List[t.Optional[E
 
 
 def parse_one(
-    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression | str] = None, **opts
+    sql: str, read: t.Optional[str | Dialect] = None, into: t.Optional[Expression | str] = None, **opts
 ) -> t.Optional[Expression]:
     """
     Parses the given SQL string and returns a syntax tree for the first parsed SQL statement.
@@ -81,8 +81,8 @@ def parse_one(
 
 def transpile(
     sql: str,
-    read: t.Optional[str] = None,
-    write: t.Optional[str] = None,
+    read: t.Optional[str | Dialect] = None,
+    write: t.Optional[str | Dialect] = None,
     identity: bool = True,
     error_level: t.Optional[ErrorLevel] = None,
     **opts,
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -1,5 +1,7 @@
 """## Python SQL parser, transpiler and optimizer."""
 
+import typing as t
+
 from sqlglot import expressions as exp
 from sqlglot.dialects import Dialect, Dialects
 from sqlglot.diff import diff
@@ -33,38 +35,36 @@ pretty = False
 schema = MappingSchema()
 
 
-def parse(sql, read=None, **opts):
+def parse(sql: str, read: t.Optional[str] = None, **opts) -> t.List[t.Optional[Expression]]:
     """
-    Parses the given SQL string into a collection of syntax trees, one per
-    parsed SQL statement.
+    Parses the given SQL string into a collection of syntax trees, one per parsed SQL statement.
 
     Args:
-        sql (str): the SQL code string to parse.
-        read (str): the SQL dialect to apply during parsing
-            (eg. "spark", "hive", "presto", "mysql").
+        sql: the SQL code string to parse.
+        read: the SQL dialect to apply during parsing (eg. "spark", "hive", "presto", "mysql").
         **opts: other options.
 
     Returns:
-        typing.List[Expression]: the list of parsed syntax trees.
+        The resulting syntax tree collection.
     """
     dialect = Dialect.get_or_raise(read)()
     return dialect.parse(sql, **opts)
 
 
-def parse_one(sql, read=None, into=None, **opts):
+def parse_one(
+    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression] = None, **opts
+) -> t.Optional[Expression]:
     """
-    Parses the given SQL string and returns a syntax tree for the first
-    parsed SQL statement.
+    Parses the given SQL string and returns a syntax tree for the first parsed SQL statement.
 
     Args:
-        sql (str): the SQL code string to parse.
-        read (str): the SQL dialect to apply during parsing
-            (eg. "spark", "hive", "presto", "mysql").
-        into (Expression): the SQLGlot Expression to parse into
+        sql: the SQL code string to parse.
+        read: the SQL dialect to apply during parsing (eg. "spark", "hive", "presto", "mysql").
+        into: the SQLGlot Expression to parse into.
         **opts: other options.
 
     Returns:
-        Expression: the syntax tree for the first parsed statement.
+        The syntax tree for the first parsed statement.
     """
 
     dialect = Dialect.get_or_raise(read)()
@@ -77,25 +77,29 @@ def parse_one(sql, read=None, into=None, **opts):
     return result[0] if result else None
 
 
-def transpile(sql, read=None, write=None, identity=True, error_level=None, **opts):
+def transpile(
+    sql: str,
+    read: t.Optional[str] = None,
+    write: t.Optional[str] = None,
+    identity: bool = True,
+    error_level: t.Optional[ErrorLevel] = None,
+    **opts
+) -> t.List[str]:
     """
-    Parses the given SQL string using the source dialect and returns a list of SQL strings
-    transformed to conform to the target dialect. Each string in the returned list represents
-    a single transformed SQL statement.
+    Parses the given SQL string in accordance with the source dialect and returns a list of SQL strings transformed
+    to conform to the target dialect. Each string in the returned list represents a single transformed SQL statement.
 
     Args:
-        sql (str): the SQL code string to transpile.
-        read (str): the source dialect used to parse the input string
-            (eg. "spark", "hive", "presto", "mysql").
-        write (str): the target dialect into which the input should be transformed
-            (eg. "spark", "hive", "presto", "mysql").
-        identity (bool): if set to True and if the target dialect is not specified
-            the source dialect will be used as both: the source and the target dialect.
-        error_level (ErrorLevel): the desired error level of the parser.
+        sql: the SQL code string to transpile.
+        read: the source dialect used to parse the input string (eg. "spark", "hive", "presto", "mysql").
+        write: the target dialect into which the input should be transformed (eg. "spark", "hive", "presto", "mysql").
+        identity: if set to `True` and if the target dialect is not specified the source dialect will be used as both:
+            the source and the target dialect.
+        error_level: the desired error level of the parser.
         **opts: other options.
 
     Returns:
-        typing.List[str]: the list of transpiled SQL statements / expressions.
+        The list of transpiled SQL statements.
     """
     write = write or read if identity else write
     return [
--- a/sqlglot/errors.py
+++ b/sqlglot/errors.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import typing as t
 from enum import auto
 
@@ -31,7 +33,7 @@ class OptimizeError(SqlglotError):
     pass
 
 
-def concat_errors(errors: t.Union[ErrorLevel, str], maximum: int) -> str:
+def concat_errors(errors: t.Sequence[t.Any], maximum: int) -> str:
     msg = [str(e) for e in errors[:maximum]]
     remaining = len(errors) - maximum
     if remaining > 0:
--- a/sqlglot/errors.py
+++ b/sqlglot/errors.py
@@ -1,3 +1,4 @@
+import typing as t
 from enum import auto
 
 from sqlglot.helper import AutoName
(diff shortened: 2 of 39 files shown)


sqlglot/__init__.py
```diff
@@ -33,38 +35,36 @@
 schema = MappingSchema()
 
 
-def parse(sql, read=None, **opts):
+def parse(sql: str, read: t.Optional[str] = None, **opts) -> t.List[t.Optional[Expression]]:
```
barakalon, line 38: `Optional[Expression]` means `Expression | None`. Can elements of the list be None?

sqlglot/__init__.py
```diff
@@ -33,38 +35,36 @@
...
+        The resulting syntax tree collection.
     """
     dialect = Dialect.get_or_raise(read)()
     return dialect.parse(sql, **opts)
 
 
-def parse_one(sql, read=None, into=None, **opts):
+def parse_one(
+    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression] = None, **opts
+) -> t.Optional[Expression]:
```
barakalon, line 56: Same question here - can this return None?

sqlglot/__init__.py
```diff
@@ -33,38 +35,36 @@
...
-        typing.List[Expression]: the list of parsed syntax trees.
+        The resulting syntax tree collection.
     """
     dialect = Dialect.get_or_raise(read)()
     return dialect.parse(sql, **opts)
 
 
-def parse_one(sql, read=None, into=None, **opts):
+def parse_one(
+    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression] = None, **opts
```
tobymao, line 55: into can also be a string, so you need a union

t.Optional[Expression | str] 

and add the future annotations at the top to support this

tobymao:
you need to turn on mypy for these files so it actually checks it

sqlglot/errors.py
```diff
@@ -30,7 +31,7 @@ class OptimizeError(SqlglotError):
     pass
 
 
-def concat_errors(errors, maximum):
+def concat_errors(errors: t.Union[ErrorLevel, str], maximum: int) -> str:
```
barakalon, line 34: I don't think `Union[ErrorLevel, str]` is right. I think it should be `Sequence[Any]`

sqlglot/helper.py
```diff
@@ -7,40 +7,57 @@
 from copy import copy
 from enum import Enum
 
+from sqlglot.expressions import Expression, Table
```
tobymao, line 10: if t.TYPE_CHECKING:

sqlglot/helper.py
```diff
@@ -7,40 +7,57 @@
...
+def list_get(arr: t.List[t.Any], index: int) -> t.Optional[t.Any]:
+    """Returns the value in `arr` at position `index`, or `None` if `index` is out of bounds."""
     try:
         return arr[index]
     except IndexError:
         return None
 
 
-def ensure_list(value):
+def ensure_list(value: t.Any) -> t.Union[t.List[t.Any], t.Tuple[t.Any], t.Set[t.Any]]:
```
tobymao, line 29: let's | instead of union

sqlglot/helper.py
```diff
@@ -7,40 +7,57 @@
...
 
 
 class AutoName(Enum):
-    def _generate_next_value_(name, _start, _count, _last_values):
+    def _generate_next_value_(name: str, _start: int, _count: int, _last_values: int) -> str:
         return name
 
 
-def list_get(arr, index):
+def list_get(arr: t.List[t.Any], index: int) -> t.Optional[t.Any]:
```
barakalon, line 21: Use a TypeVar:

```python
T = t.TypeVar("T")

def list_get(arr: t.Sequence[T], index: int) -> t.Optional[T]:
    ...
```

sqlglot/helper.py
barakalon, line 29: The proper thing to use here is [overload](https://docs.python.org/3/library/typing.html#typing.overload).

sqlglot/helper.py
```diff
@@ -53,7 +70,18 @@ def subclasses(module_name, classes, exclude=()):
     ]
 
 
-def apply_index_offset(expressions, offset):
+def apply_index_offset(expressions: t.List[Expression], offset: int) -> t.List[Expression]:
```
barakalon, line 73: This should also use a TypeVar

sqlglot/helper.py
eakmanrq, line 29: The fact that this doesn't actually ensure a list type becomes even more confusing when we start adding types. Since we don't like the cost of casting to list can we rename to `ensure_iterable`? Also should we establish a term in the codebase for referencing an iterable that is not a str or byte and then use that in the name? Maybe `ensure_collection`?

sqlglot/helper.py
```diff
@@ -64,14 +92,26 @@ def apply_index_offset(expressions, offset):
...
 
 
-def camel_to_snake_case(name):
+def camel_to_snake_case(name: str) -> str:
+    """Converts `name` from camelCase to snake_case and returns the result."""
     return CAMEL_CASE_PATTERN.sub("_", name).upper()
 
 
-def while_changing(expression, func):
+def while_changing(expression: t.Optional[Expression], func: t.Callable) -> t.Optional[Expression]:
```
barakalon, line 104: I think this is what you want:

```python
E = t.TypeVar("E", bound=Expression)

def while_changing(expression: t.Optional[Expression], func: t.Callable[[Expression], E]) -> E:
```

The TypeVar might be overkill. But in any case, I think we should annotate the Callable.

sqlglot/helper.py
```diff
@@ -80,10 +120,19 @@ def while_changing(expression, func):
     return expression
 
 
-def tsort(dag):
+def tsort(dag: t.Dict[t.Any, t.List[t.Any]]) -> t.List[t.Any]:
```
barakalon, line 123: Use a TypeVar

sqlglot/helper.py
tobymao: ensure_collection is fine

GeorgeSittas: Agreed -- I had some trouble deciding on a name for this function. I wasn't sure what was the intent we wanted to capture because there are generally two uses throughout the codebase:

1. Convert something into a list, because we expect it to be a list in other parts of the code.
2. Convert something into a list only so that we can iterate over it.

If we do `ensure_iterable`, it might obscure the intent for case 1.

sqlglot/__init__.py
GeorgeSittas: Yes, for instance:
```python
>>> import sqlglot
>>> sqlglot.parse("")
[None]
```

sqlglot/helper.py
eakmanrq: This might be a pain but my vote would be to have `ensure_list` and `ensure_collection` and then functions would use the one appropriate for their use case. For example in the DataFrame implementation I really do want a list so I would use `ensure_list` (and take the performance consequences) while other places we might actually be fine with a collection.

GeorgeSittas:
> you need to turn on mypy for these files so it actually checks it

I think we're already using mypy, no? This is in `run_checks.sh` currently:
```bash
python -m mypy sqlglot tests
```

tobymao:
> > you need to turn on mypy for these files so it actually checks it
> 
> I think we're already using mypy, no? This is in `run_checks.sh` currently:
> 
> ```shell
> python -m mypy sqlglot tests
> ```

check setup.cfg, i think it ignores all erros in sqlglot except datafram.*

sqlglot/helper.py
GeorgeSittas: Got it, I'll try that.

sqlglot/errors.py
GeorgeSittas: Oops, ErrorLevel is definitely wrong there. Although does `Sequence` capture exceptions? For example, in the parser we call `concat_errors` with `self.errors`, which can have a type of `List[ParseError]`.

barakalon: `Sequence` is an abstract type. I think it works for both List and Tuple? I might have that wrong...

But in either case, the element type is Any. We simply call `str` on the elements, which I think works for all objects.

GeorgeSittas: Yeah you're right, I got confused. Will fix this too.

GeorgeSittas:
I tried to address all raised issues:

- @barakalon I was skeptical regarding `ensure_list` and `ensure_collection` types.. Is the `@overload` used correctly?
- AFAIK python doesn't compute the `for` loop's iterable >1 times, so I think it should be ok to simplify the related parts.

I activated mypy for the entire sqlglot module, so as expected it yells a lot right now. I'll try to fix them up one by one by annotating the rest of the code soon.

sqlglot/helper.py
```diff
@@ -187,16 +286,38 @@ def split_num_words(value: str, sep: str, min_num_words: int, fill_from_start: b
...
+        >>> is_collection("test")
+        False
+
+    Args:
+        value: the value to check if it is a collection.
+
+    Returns:
+        A `bool` value indicating if it is a collection.
+    """
+    return issubclass(type(value), Collection) and not isinstance(value, (str, bytes))
```
barakalon, line 315: Can this not be `isinstance(value, Collection)`?

barakalon (commented):
Looks right to me

GeorgeSittas:
A question I had while annotating `schema.py` was whether it's correct to use `type: ignore`, for example in `column_names`. The reason I did that is because there were some errors when I did `table.this` etc., due to the `table` parameter having a type of `exp.Table | str`, even though we know it's been converted to `Table`. Does this make sense?

sqlglot/__init__.py
```diff
@@ -33,38 +37,36 @@
 schema = MappingSchema()
 
 
-def parse(sql, read=None, **opts):
+def parse(sql: str, read: t.Optional[str] = None, **opts) -> t.List[t.Optional[Expression]]:
```
tobymao, line 40: read can also be a dialect enum i think so str | Dialect

sqlglot/__init__.py
```diff
@@ -33,38 +37,36 @@
...
-        typing.List[Expression]: the list of parsed syntax trees.
+        The resulting syntax tree collection.
     """
     dialect = Dialect.get_or_raise(read)()
     return dialect.parse(sql, **opts)
 
 
-def parse_one(sql, read=None, into=None, **opts):
+def parse_one(
+    sql: str, read: t.Optional[str] = None, into: t.Optional[Expression | str] = None, **opts
```
tobymao, line 57: same here

sqlglot/__init__.py
```diff
@@ -77,25 +79,29 @@ def parse_one(sql, read=None, into=None, **opts):
     return result[0] if result else None
 
 
-def transpile(sql, read=None, write=None, identity=True, error_level=None, **opts):
+def transpile(
+    sql: str,
+    read: t.Optional[str] = None,
```
tobymao, line 84: same

sqlglot/helper.py
```diff
@@ -1,48 +1,112 @@
...
 class AutoName(Enum):
-    def _generate_next_value_(name, _start, _count, _last_values):
+    """This is used for creating enum classes where `auto()` is the string form of the corresponding value's name."""
+
+    def _generate_next_value_(name, _start, _count, _last_values):  # type: ignore
         return name
 
 
-def list_get(arr, index):
+def sequence_get(seq: t.Sequence[T], index: int) -> t.Optional[T]:
```
tobymao, line 30: maybe just seq_get

sqlglot/helper.py
```diff
@@ -1,48 +1,112 @@
...
         return []
-    return value if isinstance(value, (list, tuple, set)) else [value]
+    elif isinstance(value, (list, tuple)):
+        return list(value)
 
+    return [value]
+
+
+@t.overload
+def ensure_collection(value: t.Collection[T]) -> t.Collection[T]:
```
tobymao, line 57: this is how you do this?

sqlglot/helper.py
```diff
@@ -1,48 +1,112 @@
...
+    Args:
+        value: the value of interest.
 
-def csv(*args, sep=", "):
+    Returns:
+        The value if it's a collection, or else the value wrapped in a list.
+    """
+    if value is None:
+        return []
+    return value if is_collection(value) else [value]
```
tobymao, line 78: is_collection doesn't need to be another function call, you don't use it anywhere

sqlglot/helper.py
```diff
@@ -1,48 +1,112 @@
...
+def ensure_collection(value: t.Collection[T]) -> t.Collection[T]:
+    ...
+
+
+@t.overload
+def ensure_collection(value: T) -> t.Collection[T]:
+    ...
+
+
+def ensure_collection(value):
```
tobymao, line 66: can't this just be 

```
def ensure_collection(value: T | t.Collection[T]) -> t.Collection[T] 
```

or does the type checker get confused?

sqlglot/helper.py
```diff
@@ -53,7 +117,18 @@ def subclasses(module_name, classes, exclude=()):
     ]
 
 
-def apply_index_offset(expressions, offset):
+def apply_index_offset(expressions: t.List[E], offset: int) -> t.List[E]:
```
tobymao, line 120: why E? why not just use Expression you have it in t.TYpeChecking now

sqlglot/helper.py
```diff
@@ -119,14 +213,14 @@ def open_file(file_name):
 
 
 @contextmanager
-def csv_reader(table):
+def csv_reader(table: Table) -> t.Any:
```
tobymao, line 216: returns a csv reader then?

sqlglot/schema.py
```diff
@@ -60,57 +71,62 @@ class MappingSchema(Schema):
...
 
         Args:
-            table (sqlglot.expressions.Table|str): Table expression instance or string representing the table
-            column_mapping (dict|str|sqlglot.dataframe.sql.types.StructType|list): A column mapping that describes the structure of the table
+            table: the `Table` expression instance or string representing the table.
+            column_mapping: a column mapping that describes the structure of the table.
         """
-        table = exp.to_table(table)
-        self._validate_table(table)
+        table = exp.to_table(table)  # type: ignore
```
tobymao, line 104: why do you need ignore here

sqlglot/helper.py
GeorgeSittas: I believe so yes; if I'm not mistaken that's what Barak suggested a few days ago too.

sqlglot/helper.py
GeorgeSittas: > or does the type checker get confused?

That seems to be the case:
```python
sqlglot/helper.py:88: error: List item 0 has incompatible type "Union[T, str, bytes]"; expected "T"
```

sqlglot/helper.py
barakalon: I recommended this.
```python
l = [exp.Table(...), exp.Table(...)  # this is list[exp.Table]
apply_index_offset(l)  # using TypeVar, the return type is exp.Table. Without it, it's exp.Expression
```

I know, I know. You love type annotations Toby.

sqlglot/helper.py
barakalon: If value is `T | t.Collection[T]`, `T` will swallow up any type.
```python
ensure_collection([])  # the return type would be t.Collection[list]
```

sqlglot/helper.py
tobymao: ah i see

sqlglot/helper.py
GeorgeSittas: Right, whereas using overloads we expect mypy to explore all variants in the same order as they're declared, correct?

barakalon: I think so. Might want to verify that.

GeorgeSittas: Check, see here: https://mypy.readthedocs.io/en/stable/more_types.html#type-checking-calls-to-overloads

TL;DR
```
When you call an overloaded function, mypy will infer the correct return type by picking the best matching variant, after taking into consideration both the argument types and arity. However, a call is never type checked against the implementation. This is why mypy will report calls like mouse_event(5, 25, 3) as being invalid even though it matches the implementation signature.

If there are multiple equally good matching variants, mypy will select the variant that was defined first. [...]
```

sqlglot/helper.py
GeorgeSittas: I did some investigation about the csv module, and I think it doesn't expose its internal reader/writer classes. That's why I used `t.Any`. For example, I think the problem is made clear in [this thread](https://stackoverflow.com/questions/51264355/how-to-type-annotate-object-returned-by-csv-writer).

sqlglot/schema.py
GeorgeSittas: Discussed in slack.

tobymao (approved):

PR #646, opened 2022-10-27
Allow CREATE ... TRANSIENT ... TABLE for Snowflake

ttzhou:
Allow valid parsing of `CREATE ... TRANSIENT ... TABLE` statement in Snowflake.

Proposed behavior for transpiling to other dialects is to simply omit it. Did this with class attribute on the Snowflake dialect `Generator` + a janky-ish check in `generator.create_sql`. If more preferred way, happy to change.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -188,7 +188,7 @@ class Snowflake(Dialect):
         }
 
     class Generator(Generator):
-        can_create_transient_table = True
+        CREATE_TRANSIENT = True
 
         TRANSFORMS = {
             **Generator.TRANSFORMS,
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -188,6 +188,8 @@ class Snowflake(Dialect):
         }
 
     class Generator(Generator):
+        can_create_transient_table = True
+
         TRANSFORMS = {
             **Generator.TRANSFORMS,
             exp.ArrayConcat: rename_func("ARRAY_CAT"),
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -65,6 +65,9 @@ class Generator:
         exp.VolatilityProperty: lambda self, e: self.sql(e.name),
     }
 
+    # whether 'CREATE ... TRANSIENT ... TABLE' is allowed
+    # can override in dialects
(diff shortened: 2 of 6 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -188,6 +188,8 @@ class Tokenizer(Tokenizer):
         }
 
     class Generator(Generator):
+        can_create_transient_table = True
```
tobymao, line 191: get rid of this constant

sqlglot/tokens.py
```diff
@@ -257,6 +257,7 @@ class TokenType(AutoName):
     TABLE_FORMAT = auto()
     TABLE_SAMPLE = auto()
     TEMPORARY = auto()
+    TRANSIENT = auto()
```
tobymao, line 260: ad the transient keyword to the ID_VARS property in parser

sqlglot/generator.py
```diff
@@ -368,15 +368,18 @@ def create_sql(self, expression):
         expression_sql = self.sql(expression, "expression")
         expression_sql = f"AS{self.sep()}{expression_sql}" if expression_sql else ""
         temporary = " TEMPORARY" if expression.args.get("temporary") else ""
+        transient = (
```
tobymao, line 371: add CREATE_TRANSIENT to the base generator constants

then this is just

if self.CREATE_TRANSIENT and expression.args.get("transient")

GeorgeSittas (approved):

GeorgeSittas:
Looks good! Maybe consider adding some more tests?

ttzhou:
RE: tests - does it make sense to throw an error if TRANSIENT is used for dialects that don't support it (i.e. `Generator.CREATE_TRANSIENT = false`).

Otherwise, not sure what other tests we might add here.

GeorgeSittas:
No need to be restrictive here; the contract is that sqlglot will parse correct expressions, so it's not a problem if we allow transient everywhere.

GeorgeSittas:
@ttzhou the test doesn't look right; I think it should be:
```python
self.validate_all(
  "create transient table a (id int)",
  write={
    "postgres" : "CREATE TABLE a (id INT)",
    "snowflake" : "CREATE TRANSIENT TABLE a (id INT)"
  }
)
```
Check out how `read` and `write` are interpreted in the test suite to make sure.

Issue #647, opened 2022-10-27
Issue with convert function in Tsql - missing parenthesis


arnovkh:
Hello,

Getting an issue when trying to parse following function

`sqlStatement = """
select CONVERT(varchar(10),b.field,120) a from dbo.test b
"""
 parsed_sql = parse_one(sqlStatement,read='tsql').sql(dialect = 'tsql')
    print(parsed_sql)`

error received is sqlglot.errors.ParseError: Expecting ). Line 2, Col: 29.

removing table alias in front of field works (see below) but it cannot detect tables using find_all(exp.Table)

`sqlStatement = """
select CONVERT(varchar(10), field,120) a from dbo.test b
"""
 parsed_sql = parse_one(sqlStatement,read='tsql')
    print(parsed_sql)`

PR #649, opened 2022-10-27
T-SQL Fix for column references in a convert

10bas10:


GeorgeSittas:
Is this an attempt to solve #647 ? If so, then I apologize; I didn't see this PR at the time I was working on mine.

My approach was to do `parse_expression` instead of `parse_column`, because that's more general, although it's probably ok to do `parse_column`. In any case, maybe add some more tests and we can use this as the fix.

sqlglot/dialects/tsql.py
```diff
@@ -186,7 +186,7 @@ class Parser(Parser):
         def _parse_convert(self, strict):
             to = self._parse_types()
             self._match(TokenType.COMMA)
-            this = self._parse_field()
+            this = self._parse_column()
```
GeorgeSittas, line 189: Let's change the base parser to do the same thing.

GeorgeSittas (approved):

sqlglot/dialects/tsql.py
10bas10: Good thinking

tobymao:
i don’t think this is correct. you want to do parse_type

tobymao:
i misunderstood what the issue was looks good

tobymao:
since it's a function though, the default is to usually do parse_conjunction

PR #650, opened 2022-10-27
Fix convert issue with qualified column fields

### LINKED ISSUE #647: Issue with convert function in Tsql - missing parenthesis

GeorgeSittas:
Fixes #647

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -186,7 +186,7 @@ class TSQL(Dialect):
         def _parse_convert(self, strict):
             to = self._parse_types()
             self._match(TokenType.COMMA)
-            this = self._parse_field()
+            this = self._parse_expression()
 
             # Retrieve length of datatype and override to default if not specified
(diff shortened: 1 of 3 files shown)


GeorgeSittas:
Let's accept https://github.com/tobymao/sqlglot/pull/649 instead of this.

PR #654, opened 2022-10-28
TSQL format

TimothyAerts:
Attempt at TSQL format, format is quite versatile in that it can format any input type.
I am not fond of the regex but it does the trick in sort of safely switching between number formatting and date formatting.
The switch is needed since most target dialects will not have a versatile format like TSQL.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -65,11 +65,7 @@ def generate_date_delta_with_unit_sql(self, e):
 
 
 def generate_format_sql(self, e):
-    fmt = (
-        e.args["format"]
-        if isinstance(e, exp.NumberToStr)
-        else exp.Literal.string(self.format_time(e.args.get("format").name))
-    )
+    fmt = e.args["format"] if isinstance(e, exp.NumberToStr) else exp.Literal.string(self.format_time(e.text("format")))
     return f"FORMAT({self.format_args(e.this, fmt)})".replace("Y", "y")
 
 
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -52,7 +52,7 @@ def tsql_format_time_lambda(exp_class, full_format_mapping=None, default=None):
 
 def parse_format(args):
     fmt = list_get(args, 1)
-    number_fmt = not re.search(DATE_FMT_RE, fmt.this) or fmt.this in TRANSPILE_SAFE_NUMBER_FMT
+    number_fmt = not re.search(DATE_FMT_RE, fmt.this) or fmt.name in TRANSPILE_SAFE_NUMBER_FMT
     if number_fmt:
         return exp.NumberToStr(this=list_get(args, 0), format=fmt)
     return exp.TimeToStr(this=list_get(args, 0), format=exp.Literal.string(format_time(fmt.name, TSQL.time_mapping)))
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -28,7 +28,8 @@ DATE_DELTA_INTERVAL = {
     "d": "day",
 }
 
-DATE_FMT_RE = "([dD]{1,2})|([mM]{1,2})|([yY]{1,4})|([hH]{1,2})|([sS]{1,2})"
+
+DATE_FMT_RE = re.compile("([dD]{1,2})|([mM]{1,2})|([yY]{1,4})|([hH]{1,2})|([sS]{1,2})")
 # Current known single letter options:
 #   "d","D", "c", "C", "f","F", "g", "G", "m","o", "r", "s" ,"u","U", "T","t","Y"
 # N = Numeric, C=Currency
@@ -52,7 +53,7 @@ def tsql_format_time_lambda(exp_class, full_format_mapping=None, default=None):
 
 def parse_format(args):
     fmt = list_get(args, 1)
-    number_fmt = not re.search(DATE_FMT_RE, fmt.this) or fmt.name in TRANSPILE_SAFE_NUMBER_FMT
+    number_fmt = not DATE_FMT_RE.search(fmt.this) or fmt.name in TRANSPILE_SAFE_NUMBER_FMT
     if number_fmt:
         return exp.NumberToStr(this=list_get(args, 0), format=fmt)
     return exp.TimeToStr(this=list_get(args, 0), format=exp.Literal.string(format_time(fmt.name, TSQL.time_mapping)))
@@ -67,9 +68,7 @@ def generate_format_sql(self, e):
     fmt = (
         e.args["format"]
         if isinstance(e, exp.NumberToStr)
-        else exp.Literal.string(
-            format_time(e.args.get("format").name, {v: k.lower() for k, v in TSQL.time_mapping.items()})
-        )
+        else exp.Literal.string(format_time(e.args.get("format").name, TSQL.inversed_lower_case_time_mapping))
     )
     return f"FORMAT({self.format_args(e.this, fmt)})"
 
@@ -162,6 +161,7 @@ class TSQL(Dialect):
         "120": "%Y-%m-%d %H:%M:%S",
         "121": "%Y-%m-%d %H:%M:%S.%f",
     }
+    inversed_lower_case_time_mapping = {v: k.lower() for k, v in time_mapping.items()}
 
     class Tokenizer(Tokenizer):
         IDENTIFIERS = ['"', ("[", "]")]
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -32,7 +32,7 @@ DATE_FMT_RE = "([dD]{1,2})|([mM]{1,2})|([yY]{1,4})|([hH]{1,2})|([sS]{1,2})"
 # Current known single letter options:
 #   "d","D", "c", "C", "f","F", "g", "G", "m","o", "r", "s" ,"u","U", "T","t","Y"
 # N = Numeric, C=Currency
-TRANSPILE_SAFE_NUMBER_FMT = ("N", "C")
+TRANSPILE_SAFE_NUMBER_FMT = {"N", "C"}
 
 
 def tsql_format_time_lambda(exp_class, full_format_mapping=None, default=None):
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -42,11 +50,34 @@ def _format_time(args):
...
+        e.args.get("format")
+        if isinstance(e, exp.NumberToStr)
+        else exp.Literal.string(
+            format_time(e.args.get("format").name, {v: k.lower() for k, v in TSQL.time_mapping.items()})
+        )
+    )
+    print(self.sql(e, "format"))
+    return f"FORMAT({self.format_args(e.this, fmt)})"
+
+
```
GeorgeSittas, line 80: You can simplify `e.args.get("format")` to `e.args["format"]` (also: you forgot the print).

sqlglot/dialects/tsql.py
```diff
@@ -26,6 +28,12 @@
     "d": "day",
 }
 
+DATE_FMT_RE = "([dD]{1,2})|([mM]{1,2})|([yY]{1,4})|([hH]{1,2})|([sS]{1,2})"
+# Current known single letter options:
+#   "d","D", "c", "C", "f","F", "g", "G", "m","o", "r", "s" ,"u","U", "T","t","Y"
+# N = Numeric, C=Currency
+TRANSPILE_SAFE_NUMBER_FMT = ("N", "C")
```
GeorgeSittas, line 35: Make this a set? We might want to add more choices in the future.

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +50,34 @@ def _format_time(args):
     return _format_time
 
 
+def parse_format(args):
+    fmt = list_get(args, 1)
+    number_fmt = not re.search(DATE_FMT_RE, fmt.this) or fmt.this in TRANSPILE_SAFE_NUMBER_FMT
+    if number_fmt:
+        return exp.NumberToStr(this=list_get(args, 0), format=fmt)
+    else:
```
GeorgeSittas, line 58: We can get rid of the else clause here.

GeorgeSittas (commented):
At a first glance it looks good to me. I think MySQL and Databricks dialects can make use of this too.

sqlglot/dialects/tsql.py
```diff
@@ -26,6 +28,12 @@
     "d": "day",
 }
 
+DATE_FMT_RE = "([dD]{1,2})|([mM]{1,2})|([yY]{1,4})|([hH]{1,2})|([sS]{1,2})"
```
tobymao, line 31: can you compile this ahead of time? re.compile

what exactly does this do? it finds a d or D with length 2?

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +50,30 @@ def _format_time(args):
     return _format_time
 
 
+def parse_format(args):
+    fmt = list_get(args, 1)
+    number_fmt = not re.search(DATE_FMT_RE, fmt.this) or fmt.this in TRANSPILE_SAFE_NUMBER_FMT
```
tobymao, line 55: fmt.name instead of fmt.this

tobymao: do the fmt.name in transpilesafe because it's cheaper

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +50,30 @@ def _format_time(args):
...
+        return exp.NumberToStr(this=list_get(args, 0), format=fmt)
+    return exp.TimeToStr(this=list_get(args, 0), format=exp.Literal.string(format_time(fmt.name, TSQL.time_mapping)))
+
+
 def generate_date_delta_with_unit_sql(self, e):
     func = "DATEADD" if isinstance(e, exp.DateAdd) else "DATEDIFF"
     return f"{func}({self.format_args(e.text('unit'), e.expression, e.this)})"
 
 
+def generate_format_sql(self, e):
```
tobymao, line 66: fmt = e.text("format")

sqlglot/dialects/tsql.py
TimothyAerts: I can yes, its not the holy grail. It basically tries to find a group with  a d, m, y, h or s in both capital and lowercase format.
So it does hit on sentences/words with any of those characters.
However stuff like ###,####.### does not hit the regex. In most cases this type of stuff would be used to format numbers but there are no guarantees. Only guarantee is possible with knowing the 1 parameter type.

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +50,30 @@ def _format_time(args):
...
+    fmt = (
+        e.args["format"]
+        if isinstance(e, exp.NumberToStr)
+        else exp.Literal.string(
+            format_time(e.args.get("format").name, {v: k.lower() for k, v in TSQL.time_mapping.items()})
```
tobymao, line 71: you shouldn't be doing the dictionary comprehension every time in here, create the inrverted one as a constant

sqlglot/dialects/tsql.py
TimothyAerts: so to add, it is an educated guess.

sqlglot/dialects/tsql.py
TimothyAerts: If I take it as text I need to recast as Literal in order to get format_args to produce the text with quotes right?

sqlglot/dialects/tsql.py
TimothyAerts: good point

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +51,28 @@ def _format_time(args):
     return _format_time
 
 
+def parse_format(args):
+    fmt = list_get(args, 1)
+    number_fmt = not DATE_FMT_RE.search(fmt.this) or fmt.name in TRANSPILE_SAFE_NUMBER_FMT
```
tobymao, line 56:
```suggestion
    number_fmt = fmt.name in TRANSPILE_SAFE_NUMBER_FMT or not DATE_FMT_RE.search(fmt.this) 
```

sqlglot/dialects/tsql.py
tobymao: ok, then only do it for the format_time one

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +51,28 @@ def _format_time(args):
...
 def generate_date_delta_with_unit_sql(self, e):
     func = "DATEADD" if isinstance(e, exp.DateAdd) else "DATEDIFF"
     return f"{func}({self.format_args(e.text('unit'), e.expression, e.this)})"
 
 
+def generate_format_sql(self, e):
+    fmt = (
+        e.args["format"]
+        if isinstance(e, exp.NumberToStr)
+        else exp.Literal.string(format_time(e.args.get("format").name, TSQL.inversed_lower_case_time_mapping))
```
tobymao, line 71: why would you want a lower case mapping? what's the point of that since time_mapping already has lower case stuff? don't you only want an inverse mapping? additionally, sqlglot automatically produces an inverse_time_mapping, so you don't need to remake it

TimothyAerts: tsl doesn't allow YYYY for example in FORMAT

TimothyAerts: Technically it will allow it but for example FORMAT('01-01-1990', dd.mm.YYYY) => 01.01.YYYY which is not the intended output. Better solution might be to force somehow the format_time to return the lowercase by default but I dont think other dialects have any benefit of that.

sqlglot/dialects/tsql.py
TimothyAerts: ah sorry I misunderstood, indeed way better

sqlglot/dialects/tsql.py
tobymao: just do the lower case afterwards intsead of in the mapping

TimothyAerts: Yea you are right it doesn't change anything the way it is here. But I should have checked for YYYY in the dict comprehension. MM and mm are different formaters, so pushing all to lower is wrong.

sqlglot/dialects/tsql.py
```diff
@@ -42,11 +51,24 @@ def _format_time(args):
...
+
+
 def generate_date_delta_with_unit_sql(self, e):
     func = "DATEADD" if isinstance(e, exp.DateAdd) else "DATEDIFF"
     return f"{func}({self.format_args(e.text('unit'), e.expression, e.this)})"
 
 
+def generate_format_sql(self, e):
+    fmt = e.args["format"] if isinstance(e, exp.NumberToStr) else exp.Literal.string(self.format_time(e.text("format")))
+    return f"FORMAT({self.format_args(e.this, fmt)})".replace("Y", "y")
```
tobymao, line 69: why not just do all lower?

tobymao, line 69:
```suggestion
    return f"FORMAT({self.format_args(e.this, fmt)})".lower()
```

TimothyAerts: mm != MM

TimothyAerts: SELECT FORMAT (GETDATE(), 'dd MM yyy') =>28 10 2022
SELECT FORMAT (GETDATE(), 'dd mm yyy') => 28 03 2022 (minutes)

tobymao: this should just be handled in time_mapping then, time_mapping should return Y as y, maybe switch the ordering of time_mapping?

TimothyAerts: same goes for the 1 letter formats, u, U , g, G etc.

tobymao: https://join.slack.com/t/tobiko-data/shared_invite/zt-1gw56z9k5-XnobJdCffsu0bWXV0PNwiA 

you can join my slack if you wanna chat, maybe easier

TimothyAerts: Hi Toby,

Please review the solution I went with relying on order (feels wierd I always pushed people to not rely on it but python has evolved :)) and took the single letters as seperate time mapping.

tobymao (approved):

PR #655, opened 2022-10-28
TSQL parsing + symbol for concatenation or arithmetic operation

10bas10:
Hi All,

Im facing quite a challenge for parsing the + symbol in T-SQL. I would love to get some help on my approach.

The problem is that there are two kind of contexts possible when using the + symbol:
- For arithmetic operations such as column1 + 234
- For string concatenations such as column1 + '234'

The problem is to identify it is either a + symbol for concatenation or not. In case of a string concatenation I managed to translate this to DPipe and is handled ok. However you would need to scan for string values in order to determine the context of string concatenation. This will never be a 100% solution as you would need the datatypes of the columns involved. However im trying a best effort approach here.

My approach is to determine the related set of literals and determine whether its a string concatenation based on that. After I would like to replace all of the exp.Add with a exp.DPipe. 

Do you guys have any good ideas on this problem?

Thanks!

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -249,10 +249,9 @@ class TSQL(Dialect):
                     expressions = [term.expression]
                     this = term.this
 
-                    while isinstance(this, (exp.Add, exp.Paren)):
-                        if isinstance(this, exp.Add):
-                            expressions.insert(0, this.expression)
-                            this = this.this
+                    while isinstance(this, exp.Add):
+                        expressions.insert(0, this.expression)
+                        this = this.this
                     else:
                         expressions.insert(0, this)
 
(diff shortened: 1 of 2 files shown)


tobymao:
this cannot be done this way. this can only be done with type inference

tobymao:
for the simple case, it’s better to do this as a transform as opposed to at parse time

sqlglot/dialects/tsql.py
```diff
@@ -219,6 +220,50 @@ def _parse_convert(self, strict):
             # Entails a simple cast without any format requirement
             return self.expression(exp.Cast if strict else exp.TryCast, this=this, to=to)
 
+        def _parse_term(self):
```
tobymao, line 223: take a look at sqlglot/transforms and how that is used. something like this would go there

10bas10: Ill have a look thanks!

PR #663, opened 2022-11-02
Replace annotations with SQL comments

GeorgeSittas:
- [x] Fix `PARTITION` so that it doesn't store tuples for its arguments.
- [x] Handle multi-line comments.
- [x] Always output comments regardless of `pretty` mode or not (or discard them if corresponding flag is set).
- [x] Investigate whether all dialects support `--` and `/**/`.
- [x] Try to make the following work: 
  - `select 1 /* comment */ + 1`.
  - `select * from x --comment`.
- [x] Change the tokenizer's logic so that we can have leading comments in the same line (e.g. /* comment */ foo).
- [x] Make sure we don't transpile comments into hints.
- [x] Try to simplify parser logic so that we don't call `_attach_comment` in many places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,15 @@ LEFT JOIN `baz`
   ON `f`.`a` = `baz`.`a`
 ```
 
-Finally, SQLGlot tries to preserve comments when formatting SQL code using the `pretty=True` option:
+Comments are preserved in a best-effort basis when traspiling SQL code:
 
 ```python
 sql = """
-SELECT            -- comment1
+/* multi
+   line
+   comment
+*/
+SELECT
   CAST(x AS INT), -- comment2
   y               -- comment3
 FROM
@@ -135,7 +139,11 @@ print(sqlglot.transpile(sql, pretty=True)[0])
 ```
 
 ```sql
-SELECT -- comment1
+/* multi
+   line
+   comment
+*/
+SELECT
   CAST(x AS INT), -- comment2
   y -- comment3
 FROM bar
--- a/README.md
+++ b/README.md
@@ -129,13 +129,15 @@ sql = """
    comment
 */
 SELECT
-  CAST(x AS INT), -- comment2
-  y               -- comment3
+  tbl.cola /* comment 1 */ + tbl.colb /* comment 2 */,
+  CAST(x AS INT), # comment 3
+  y               -- comment 4
 FROM
-  bar
+  bar /* comment 5 */,
+  tbl #          comment 6
 """
 
-print(sqlglot.transpile(sql, pretty=True)[0])
+print(sqlglot.transpile(sql, read='mysql', pretty=True)[0])
 ```
 
 ```sql
@@ -144,9 +146,10 @@ print(sqlglot.transpile(sql, pretty=True)[0])
    comment
 */
 SELECT
-  CAST(x AS INT), -- comment2
-  y -- comment3
-FROM bar
+  tbl.cola /* comment 1 */ + tbl.colb /* comment 2 */,
+  CAST(x AS INT) /* comment 3 */,
+  y /* comment 4 */
+FROM bar /* comment 5 */, tbl /* comment 6 */
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -147,9 +147,9 @@ print(sqlglot.transpile(sql, read='mysql', pretty=True)[0])
 */
 SELECT
   tbl.cola /* comment 1 */ + tbl.colb /* comment 2 */,
-  CAST(x AS INT) /* comment 3 */,
-  y /* comment 4 */
-FROM bar /* comment 5 */, tbl /* comment 6 */
+  CAST(x AS INT), -- comment 3
+  y -- comment 4
+FROM bar /* comment 5 */, tbl /*          comment 6*/
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,6 @@ Contributions are very welcome in SQLGlot; read the [contribution guide](https:/
    * [Unsupported Errors](#unsupported-errors)
    * [Build and Modify SQL](#build-and-modify-sql)
    * [SQL Optimizer](#sql-optimizer)
-   * [SQL Annotations](#sql-annotations)
    * [AST Introspection](#ast-introspection)
    * [AST Diff](#ast-diff)
    * [Custom Dialects](#custom-dialects)
@@ -120,6 +119,28 @@ JOIN `bar` AS `b`
 LEFT JOIN `baz`
   ON `f`.`a` = `baz`.`a`
 ```
+
+Finally, SQLGlot tries to preserve comments when formatting SQL code using the `pretty=True` option:
+
+```python
+sql = """
+SELECT            -- comment1
+  CAST(x AS INT), -- comment2
+  y               -- comment3
+FROM
+  bar
+"""
+
+print(sqlglot.transpile(sql, pretty=True)[0])
+```
+
+```sql
+SELECT -- comment1
+  CAST(x AS INT), -- comment2
+  y -- comment3
+FROM bar
+```
+
 
 ### Metadata
 
@@ -247,17 +268,6 @@ SELECT
 FROM "x" AS "x"
 WHERE
   "x"."Z" = CAST('2021-02-01' AS DATE)
-```
-
-### SQL Annotations
-
-SQLGlot supports annotations in the sql expression. This is an experimental feature that is not part of any of the SQL standards but it can be useful when needing to annotate what a selected field is supposed to be. Below is an example:
-
-```sql
-SELECT
-  user # primary_key,
-  country
-FROM users
 ```
 
 ### AST Introspection
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -365,6 +365,7 @@ class Expression(metaclass=_Expression):
             )
             for k, vs in self.args.items()
         }
+        args["comment"] = self.comment
         args = {k: v for k, v in args.items() if v or not hide_missing}
 
         right = ", ".join(f"{k}: {v}" for k, v in args.items())
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -135,6 +135,7 @@ class Expression(metaclass=_Expression):
     def __deepcopy__(self, memo):
         copy = self.__class__(**deepcopy(self.args))
         copy.comment = self.comment
+        copy.type = self.type
         return copy
 
     def copy(self):
@@ -366,6 +367,7 @@ class Expression(metaclass=_Expression):
             for k, vs in self.args.items()
         }
         args["comment"] = self.comment
+        args["type"] = self.type
         args = {k: v for k, v in args.items() if v or not hide_missing}
 
         right = ", ".join(f"{k}: {v}" for k, v in args.items())
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -85,9 +85,18 @@ class Expression(metaclass=_Expression):
             return field.this
         return ""
 
-    def find_comment(self, key):
+    def find_comment(self, key: str) -> str:
+        """
+        Finds the comment that is attached to specified child node.
+
+        Args:
+            key: the key of the target child node (e.g. "this", "expression", etc).
+
+        Returns:
+            The comment attached to the child node, or the empty string, if it doesn't exist.
+        """
         field = self.args.get(key)
-        return field.comment if field is not None else ""
+        return field.comment if isinstance(field, Expression) else ""
 
     @property
     def is_string(self):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -133,7 +133,9 @@ class Expression(metaclass=_Expression):
         return self.alias or self.name
 
     def __deepcopy__(self, memo):
-        return self.__class__(**deepcopy(self.args))
+        copy = self.__class__(**deepcopy(self.args))
+        copy.comment = self.comment
+        return copy
 
     def copy(self):
         new = deepcopy(self)
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -85,7 +85,7 @@ class Expression(metaclass=_Expression):
             return field.this
         return ""
 
-    def comment_(self, key):
+    def find_comment(self, key):
         field = self.args.get(key)
         return field.comment if field is not None else ""
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -87,7 +87,7 @@ class Expression(metaclass=_Expression):
 
     def find_comment(self, key: str) -> str:
         """
-        Finds the comment that is attached to specified child node.
+        Finds the comment that is attached to a specified child node.
 
         Args:
             key: the key of the target child node (e.g. "this", "expression", etc).
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -85,6 +85,10 @@ class Expression(metaclass=_Expression):
             return field.this
         return ""
 
+    def comment_(self, key):
+        field = self.args.get(key)
+        return field.comment if field is not None else ""
+
     @property
     def is_string(self):
         return isinstance(self, Literal) and self.args["is_string"]
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -40,13 +40,14 @@ class Expression(metaclass=_Expression):
 
     key = None
     arg_types = {"this": True}
-    __slots__ = ("args", "parent", "arg_key", "type")
+    __slots__ = ("args", "parent", "arg_key", "type", "comment")
 
     def __init__(self, **args):
         self.args = args
         self.parent = None
         self.arg_key = None
         self.type = None
+        self.comment = None
 
         for arg_key, value in self.args.items():
             self._set_parent(arg_key, value)
@@ -577,17 +578,6 @@ class Unionable(Expression):
 
 class UDTF(DerivedTable, Unionable):
     pass
-
-
-class Annotation(Expression):
-    arg_types = {
-        "this": True,
-        "expression": True,
-    }
-
-    @property
-    def alias(self):
-        return self.expression.alias_or_name
 
 
 class Cache(Expression):
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -276,7 +276,7 @@ class Generator:
             for i, line in enumerate(lines)
         )
 
-    def sql(self, expression, key=None):
+    def sql(self, expression, key=None, comment=False, **kwargs):
         if not expression:
             return ""
 
@@ -284,6 +284,13 @@ class Generator:
             return expression
 
         if key:
+            if comment:
+                return self.maybe_comment(
+                    self.sql(expression.args.get(key), comment=comment, **kwargs),
+                    expression.find_comment(key),
+                    **kwargs,
+                )
+
             return self.sql(expression.args.get(key))
 
         transform = self.TRANSFORMS.get(expression.__class__)
@@ -619,12 +626,12 @@ class Generator:
             for part in [
                 self.sql(expression, "catalog"),
                 self.sql(expression, "db"),
-                self.maybe_comment(self.sql(expression, "this"), expression.find_comment("this")),
+                self.sql(expression, "this", comment=True),
             ]
             if part
         )
 
-        alias = self.maybe_comment(self.sql(expression, "alias"), expression.find_comment("alias"))
+        alias = self.sql(expression, "alias", comment=True)
         alias = f"{sep}{alias}" if alias else ""
         laterals = self.expressions(expression, key="laterals", sep="")
         joins = self.expressions(expression, key="joins", sep="")
@@ -905,7 +912,7 @@ class Generator:
 
         return self.query_modifiers(
             expression,
-            self.wrap(expression),
+            self.maybe_comment(self.wrap(expression), expression.comment),
             self.expressions(expression, key="pivots", sep=" "),
             f" AS {alias}" if alias else "",
         )
@@ -1262,10 +1269,8 @@ class Generator:
         return f"SHOW {self.sql(expression, 'this')}"
 
     def binary(self, expression, op):
-        left = self.maybe_comment(self.sql(expression, "this"), expression.find_comment("this"))
-        right = self.maybe_comment(
-            self.sql(expression, "expression"), expression.find_comment("expression")
-        )
+        left = self.sql(expression, "this", comment=True)
+        right = self.sql(expression, "expression", comment=True)
         return f"{left} {op} {right}"
 
     def function_fallback_sql(self, expression):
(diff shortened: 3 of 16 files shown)


sqlglot/parser.py
```diff
@@ -660,6 +660,12 @@ def _advance(self, times=1):
     def _retreat(self, index):
         self._advance(index - self._index)
 
+    def _attach_comment(self, node, comment):
+        try:
+            node.comment = comment
+        finally:
+            return node
+
```
GeorgeSittas, line 668: This method is sometimes called from `parse_csv` with a tuple argument for `node`, so `node.column` fails. I thought that the try / finally approach is a good choice to handle any such scenario where we don't have the `comment` attribute and is better than `hasattr`, probably.

sqlglot/parser.py
```diff
@@ -660,6 +660,12 @@ def _advance(self, times=1):
     def _retreat(self, index):
         self._advance(index - self._index)
 
+    def _attach_comment(self, node, comment):
+        try:
```
tobymao, line 664: why try catch?

sqlglot/parser.py
```diff
@@ -2482,23 +2485,25 @@ def _parse_replace(self):
...
 
-    def _parse_csv(self, parse):
-        parse_result = parse()
+    def _parse_csv(self, parse_method):
+        parse_result = parse_method()
         items = [parse_result] if parse_result is not None else []
 
         while self._match(TokenType.COMMA):
-            parse_result = parse()
+            self._attach_comment(parse_result, self._prev.comment)
```
tobymao, line 2493: it doesn't matter if it's newlines?

sqlglot/tokens.py
```diff
@@ -392,13 +392,13 @@ class Tokenizer(metaclass=_Tokenizer):
...
         "~": TokenType.TILDA,
         "?": TokenType.PLACEHOLDER,
-        "#": TokenType.ANNOTATION,
         "@": TokenType.PARAMETER,
         # used for breaking a var like x'y' but nothing else
         # the token type doesn't matter
         "'": TokenType.QUOTE,
         "`": TokenType.IDENTIFIER,
         '"': TokenType.IDENTIFIER,
+        "#": TokenType.HASH,
```
GeorgeSittas, line 401: We needed this because otherwise `#` would be recognized as a VAR, due to not being included in `KEYWORD_TRIE`.

tests/test_parser.py
```diff
@@ -127,26 +127,28 @@ def test_parameter(self):
...
         expression = parse_one(
             """
-            SELECT
-                a #annotation1,
-                b as B #annotation2:testing ,
-                "test#annotation",c#annotation3, d #annotation4,
-                e #,
-                f # space
+            --comment1
+            SELECT /* this won't be used */
```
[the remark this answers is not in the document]
GeorgeSittas: Because in the tokens module if we have 2 comments attached in _the same_ token, the trailing one is discarded. The reason is that it didn't really made sense to have two different comments attached to the same token. Also, we'd have to carry both leading and trailing comments for every token otherwise.

sqlglot/parser.py
```diff
@@ -660,6 +660,12 @@ def _advance(self, times=1):
     def _retreat(self, index):
         self._advance(index - self._index)
 
+    def _attach_comment(self, node, comment):
+        try:
+            node.comment = comment
```
tobymao, line 665: always strip the comment

comment.strip()

tests/test_parser.py
```diff
@@ -127,26 +127,28 @@ def test_parameter(self):
...
+                a, --comment2
```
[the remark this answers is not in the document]
GeorgeSittas: Yeah, multi-line comments are not generated correctly at the moment, but I think it's just a matter of changing the default comment to be /* .. */ instead of -- for this to be fixed.

tests/test_parser.py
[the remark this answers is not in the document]
tobymao: but this comment should be attached to the select because it's right after it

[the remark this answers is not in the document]
tobymao: what if you just did

select -- coment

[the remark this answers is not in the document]
GeorgeSittas: If select didn't have a leading comment, that'd bind to it and it would work as expected.

README.md
```diff
@@ -121,6 +120,28 @@ LEFT JOIN `baz`
   ON `f`.`a` = `baz`.`a`
 ```
 
+Finally, SQLGlot tries to preserve comments when formatting SQL code using the `pretty=True` option:
```
tobymao, line 123: don't need Finally

sqlglot/generator.py
```diff
@@ -218,6 +218,15 @@ def sep(self, sep=" "):
     def seg(self, sql, sep=" "):
         return f"{self.sep(sep)}{sql}"
 
+    def maybe_comment(self, sql, comment):
+        if not self._comments or not comment:
+            return sql
+
+        if not self.pretty:
+            return f"{sql}{f' /* {comment} */'}"
```
tobymao, line 226: i think the comment start and comment end should be passed in because different dialects could override it

basically find a comment with start and end, and one without there are two different types

default can be -- and /*

tobymao: or if you do some research and find everybody supports -- and /* then we don't need to add it

sqlglot/generator.py
```diff
@@ -499,8 +503,8 @@ def identifier_sql(self, expression):
     def partition_sql(self, expression):
         keys = csv(
             *[
-                f"{k.args['this']}='{v.args['this']}'" if v else k.args["this"]
-                for k, v in expression.args.get("this")
+                f"{prop.this.this}={prop.args['value']}" if prop.args["value"] else prop.this.this
```
tobymao, line 506: is this right?

why not 

prop.text("this") or prop.name

and prop.text("value")

tests/test_transpile.py
```diff
@@ -64,23 +64,40 @@ def test_space(self):
...
+        self.validate(
+            """
+            /*
+              multi
+              line
+              comment
+            */
+            SELECT * FROM foo
+            """,
+            """/* multi
```
tobymao, line 94: maybe we shouldn't do a strip and just preserve everything

sqlglot/generator.py
GeorgeSittas: The dialects we currently have in sqlglot seem to support this syntax so we can omit the extra arguments for now I think.

sqlglot/generator.py
```diff
@@ -50,9 +50,6 @@ class Generator:
             The default is on the smaller end because the length only represents a segment and not the true
             line length.
             Default: 80
-        annotations: Whether or not to show annotations in the SQL when `pretty` is True.
```
tobymao, line 53: we should have an option to strip comments

tobymao: comments: True by default, you can pass in comments: False to strip comments

GeorgeSittas:
I think we should always default to using multi-line comments (/* ... */) if we can [1], because otherwise it would be tricky to decide whether single line comments (--) would produce valid SQL code. For example, if we used single line comments with the pretty=True option, we'd get the following (invalid) SQL as output:

```python
>>> print(sqlglot.transpile("select 1 /*c*/ + 2 /*c*/ + 3, 2", pretty=True)[0])
SELECT
  1 --c + 2 --c + 3,
  2
```

[1]: I say "when we can", because some dialects (e.g. Snowflake, if I'm not mistaken) may not support multi-line comments and so it's trivially safe to use single-line comments for them.

sqlglot/generator.py
GeorgeSittas: Alternatively, I was thinking:

- Always strip if we're not in pretty mode, or if the comment doesn't contain any newlines. The reason is that:
  - For the former, we want to use /**/ comments and it'd look nice if we ensured that we'll always have 1 leading / trailing space.

  - For the latter, I think preserving leading / trailing space would be unnecessary because the comment appears in a single line and so it's not like we're discarding away important information (?).

- Otherwise, strip comment conditionally, like you suggested.

Does that make sense?

tobymao: i don't mean strip(), i mean remove the comments

tobymao: so False means remove comments entirely

GeorgeSittas: Ah ok I'll get it back in then.

tobymao:
it's not safe to use single line comments in snowflake because you can write in a dialect that uses multiline and then converted to snowflake would break,  another way to do it would be to always make sure there's a new line after a single comment

sqlglot/expressions.py
```diff
@@ -84,6 +85,10 @@ def text(self, key):
             return field.this
         return ""
 
+    def find_comment(self, key):
```
tobymao, line 88: doc string

sqlglot/expressions.py
```diff
@@ -84,6 +85,10 @@ def text(self, key):
...
+        field = self.args.get(key)
+        return field.comment if field is not None else ""
```
tobymao, line 90: if isinstance(field, Expression)

sqlglot/generator.py
```diff
@@ -218,6 +219,15 @@ def sep(self, sep=" "):
     def seg(self, sql, sep=" "):
         return f"{self.sep(sep)}{sql}"
 
+    def maybe_comment(self, sql, comment):
+        if not comment or not self._comments:
+            return sql
+
+        if not self.pretty or "\n" not in comment:
+            return f"{sql} /* {comment.strip()} */"
```
tobymao, line 227: i think we never want to strip also your newline check isn't cross platform, you may need to do a newline check 

https://stackoverflow.com/questions/1331815/regular-expression-to-match-cross-platform-newline-characters

sqlglot/generator.py
```diff
@@ -218,6 +219,15 @@ def sep(self, sep=" "):
...
+
+        return f"/*{comment}*/\n{sql}"
```
tobymao, line 229: wait this always puts sql behind?

sqlglot/generator.py
tobymao, line 227: looks like we don't do -- ? i thought we'd have the ability to do -- if pretty and no line breaks

sqlglot/parser.py
```diff
@@ -660,6 +660,11 @@ def _advance(self, times=1):
     def _retreat(self, index):
         self._advance(index - self._index)
 
+    def _attach_comment(self, node, comment):
+        if node and node.comment is None and comment is not None:
```
tobymao, line 664: you don't need to check comment is not None

sqlglot/generator.py
GeorgeSittas: The reason I call strip here is that it's awkward to have `/* text*/`, whenever the input sql has a `--` comment. I guess it does make sense to never strip comments, if for example you want to do something like the following in SQL:

```sql
-- Args:
--     foo: ...
```
Although, at the time of writing this I was thinking what I wrote in [this](https://github.com/tobymao/sqlglot/pull/663#discussion_r1012459775) reply. I can do it so that we never strip if you think that's the correct way.

sqlglot/parser.py
```diff
@@ -2482,23 +2488,28 @@ def _parse_replace(self):
...
 
         while self._match(TokenType.COMMA):
-            parse_result = parse()
+            self._attach_comment(parse_result, self._prev.comment)
+            parse_result = parse_method()
             if parse_result is not None:
                 items.append(parse_result)
 
+        # This makes sure we don't attach a comment twice to the last term in a binary expression
+        if not isinstance(parse_result, exp.Binary):
```
tobymao, line 2502: why would this happen?

sqlglot/generator.py
GeorgeSittas: Correct -- though that's only for pretty mode and for when we don't have any newlines involved. This is how we're able to preserve the following form, instead of cluttering the output with a multiline comment that's awkwardly formatted:
```sql
/* multi
    line
    comment */
SELECT * FROM bla```

File: sqlglot/generator.py (hunk 15 above)
Comment (GeorgeSittas) [reply sqlglot/generator.py:227]: Yes -- I've documented why using `--` is incorrect with the current formatting rules in the generator [here](https://github.com/tobymao/sqlglot/pull/663#issuecomment-1301583400).

File: sqlglot/parser.py (hunk 18 above)
Comment (GeorgeSittas) [reply sqlglot/parser.py:2502]:
```sql
SELECT a /*x*/ + b /*x*/
```
This would otherwise wrongly produce:
```sql
SELECT a /* x */ + b /* x */ /* x */
```
That's because we're adding the `/* x */` comment on the Binary node which is printed in the generator's `expressions` method, and we also print it for the primary `b` value. Let me know if you have any ideas on how we can make this cleaner.

sqlglot/generator.py
GeorgeSittas: Discussed offline, maybe it's possible we can fix this. I'll take a look.

sqlglot/generator.py
```diff
@@ -218,6 +222,18 @@ def sep(self, sep=" "):
     def seg(self, sql, sep=" "):
         return f"{self.sep(sep)}{sql}"
 
+    def maybe_comment(self, sql, comment, single_line=False):
+        if not comment or not self._comments:
+            return sql
+
+        if not self.pretty:
+            return f"{sql} /*{comment}*/"
```
tobymao, line 230: this is not safe, you need to do /* {comment} */ 
the space is important because you could do

select /* +hi */ 

but if you turn it into /*+ hi */, it becomes a spark hint

same in mysql with ! instead of +

sqlglot/generator.py
```diff
@@ -218,6 +222,18 @@ def sep(self, sep=" "):
     def seg(self, sql, sep=" "):
         return f"{self.sep(sep)}{sql}"
 
+    def maybe_comment(self, sql, comment, single_line=False):
+        if not comment or not self._comments:
+            return sql
+
+        if not self.pretty:
```
tobymao, line 229: if not self.pretty or NEWLINE_RE.search(comment)

sqlglot/generator.py
tobymao: i wonder how we can do this without adding an extra space?

maybe strip out one space? haha

sqlglot/generator.py
```diff
@@ -1267,21 +1282,35 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
...
+        # The only time leading_comma changes the output is if pretty print is enabled
         if self._leading_comma and self.pretty:
             pad = " " * self.pad
             expressions = "\n".join(
-                f"{sep}{s}" if i > 0 else f"{pad}{s}" for i, s in enumerate(sql)
+                f"{sep if i > 0 else pad}{self.maybe_comment(sql, com, single_line=True)}"
+                for i, (sql, com) in enumerate(sql_with_comment)
+            )
+        elif self.pretty:
+            sep = sep.strip()
```
tobymao, line 1298: why do you call sep.strip() what's going on here?

sqlglot/generator.py
```diff
@@ -1267,21 +1282,35 @@ def expressions(self, expression, key=None, flat=False, indent=True, sep=", "):
             return ""
 
         if flat:
-            return sep.join(self.sql(e) for e in expressions)
+            return sep.join(self.maybe_comment(self.sql(e), e.comment) for e in expressions)
 
-        sql = (self.sql(e) for e in expressions)
-        # the only time leading_comma changes the output is if pretty print is enabled
+        sql_with_comment = [(self.sql(e), e.comment) for e in expressions]
+        num_sqls = len(sql_with_comment)
```
tobymao, line 1288: i think this section might be cleaner if we do. right now it seems a bit much

 for i, (sql, comment) in enumerate(sql_with_comment):
    if self.pretty:
         if self._leading_comma:
         else:
    else:
        ....

sqlglot/parser.py
```diff
@@ -1852,11 +1856,13 @@ def _parse_column(self):
                 this = self.expression(exp.Dot, this=this, expression=field)
             this = self._parse_bracket(this)
 
-        return this
+        return self._attach_comment(this, self._prev.comment)
 
     def _parse_primary(self):
         if self._match_set(self.PRIMARY_PARSERS):
-            return self.PRIMARY_PARSERS[self._prev.token_type](self, self._prev)
+            return self._attach_comment(
```
tobymao, line 1863: i'm not sure if this is the right place to put it, you can just put it in each primary parser?

sqlglot/parser.py
tobymao: sorry i don't understand why that would happen

in select a / * x */ + b /* y*/

the comments would be directly attached to a and b, there should be no issues

sqlglot/parser.py
```diff
@@ -1933,7 +1939,7 @@ def _parse_function(self, functions=None):
             else:
                 this = self.expression(exp.Anonymous, this=this, expressions=args)
         self._match_r_paren()
-        return self._parse_window(this)
+        return self._parse_window(self._attach_comment(this, self._prev.comment))
```
tobymao, line 1942: instead of wrapping _attach_comment everywhere, what if you just added it to

def expression ? so anytime we make an expression we also parse for comments

sqlglot/generator.py
GeorgeSittas: We don't strip the comments anymore at all though, so if the user has `/*+ hi */` then the type hint is printed, whereas for `/* +hi */`, the comment is printed. Leading and trailing whitespace is preserved. Does that make sense?

GeorgeSittas: But for any case I'm leaving these as examples of the current behavior for reference:
```python
>>> sqlglot.transpile("SELECT /*+ BROADCAST(y) */ x.b FROM x JOIN y ON x.b = y.b", read='spark')[0]
'SELECT /*+ BROADCAST(y) */ x.b FROM x JOIN y ON x.b = y.b'
>>>
>>> sqlglot.transpile("SELECT /* +BROADCAST(y) */ x.b FROM x JOIN y ON x.b = y.b", read='spark')[0]
'SELECT /* +BROADCAST(y) */ x.b FROM x JOIN y ON x.b = y.b'
>>>
>>> sqlglot.transpile("SELECT /*+BROADCAST(y)*/ x.b FROM x JOIN y ON x.b = y.b", read='spark')[0]
'SELECT /*+ BROADCAST(y) */ x.b FROM x JOIN y ON x.b = y.b'
```

sqlglot/generator.py
GeorgeSittas: That would happen before too in pretty mode, due to `sep()` -- it's so we don't have a trailing space after each comma if we're about to drop a newline right after it.

sqlglot/parser.py
GeorgeSittas: Right, sorry, maybe I miscommunicated that. I'll try to explain it a little better:

- We want to preserve comments for SELECT's expressions, which is currently handled in `Generator.expressions`.

- If we want to be able to preserve comments for the children of a Binary expression, we also have to add comment generation logic in the corresponding methods of the Generator (here: `Generator.binary`). This is what I did to have `SELECT a /*x*/ + b /*y*/` transpile into itself.

- However, this last change is conflicting with the logic for attaching comments in SELECT's expressions. To see this, let's suppose that we didn't have the instance check for Binary in the `_parse_csv` method. Then, `/*y*/` would be attached to both the `Binary` node _and_ the `Identifier` node for `b`, which means that we would generate it twice:

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT a /*x*/ + b /*y*/")[0] # without the isinstance check
'SELECT a /*x*/ + b /*y*/ /*y*/'
```

Does that make sense? Do you see an other way around this issue?

sqlglot/generator.py
GeorgeSittas: Agreed it seems to much. I'll try that, thanks.

sqlglot/parser.py
GeorgeSittas: What were your concerns? I think it's less code-duplication if we have it here instead of each primary parser separately.

sqlglot/parser.py
GeorgeSittas: That sounds interesting, although I'm concerned it might not take care of lower level calls that don't delegate to `parse_expression` in their body. I'll try it.

sqlglot/generator.py
tobymao: as long as --+ doesn't turn into a hint then this is ok

sqlglot/parser.py
```diff
@@ -1872,11 +1872,13 @@ def _parse_column(self):
                 this = self.expression(exp.Dot, this=this, expression=field)
             this = self._parse_bracket(this)
 
-        return this
+        return self._attach_comment(this, self._prev.comment)
```
tobymao, line 1875: why is this necessary?

sqlglot/parser.py
```diff
@@ -2417,12 +2419,15 @@ def _parse_alias(self, this, explicit=False):
...
             self._match_r_paren()
+            aliases.comment = self._prev.comment
             return aliases
 
         alias = self._parse_id_var(any_token)
 
         if alias:
-            return self.expression(exp.Alias, this=this, alias=alias)
+            alias = self.expression(exp.Alias, this=this, alias=alias)
+            alias.comment = self._prev.comment
```
tobymao, line 2429: why is this necessary? shouldn't the id_var have the comment already?

sqlglot/parser.py
```diff
@@ -2417,12 +2419,15 @@ def _parse_alias(self, this, explicit=False):
                 expressions=self._parse_csv(lambda: self._parse_id_var(any_token)),
             )
             self._match_r_paren()
+            aliases.comment = self._prev.comment
```
tobymao, line 2422: this is unnecessary if you add comment parsing on parens

sqlglot/parser.py
```diff
@@ -1149,11 +1159,9 @@ def _parse_table_alias(self):
         if not alias and not columns:
             return None
 
-        return self.expression(
-            exp.TableAlias,
-            this=alias,
-            columns=columns,
-        )
+        alias = self.expression(exp.TableAlias, this=alias, columns=columns)
+        alias.comment = self._prev.comment
```
tobymao, line 1163: why is this needed

sqlglot/expressions.py
```diff
@@ -119,7 +133,9 @@ def alias_or_name(self):
         return self.alias or self.name
 
     def __deepcopy__(self, memo):
-        return self.__class__(**deepcopy(self.args))
+        copy = self.__class__(**deepcopy(self.args))
+        copy.comment = self.comment
+        return copy
```
GeorgeSittas, line 138: Maybe we also want to do `copy.type = self.type` here?

sqlglot/expressions.py
```diff
@@ -349,6 +365,7 @@ def to_s(self, hide_missing=True, level=0):
             )
             for k, vs in self.args.items()
         }
+        args["comment"] = self.comment
         args = {k: v for k, v in args.items() if v or not hide_missing}
```
GeorgeSittas, line 369: Same as above: `args["type"] = self.type`?

GeorgeSittas (commented):
Thanks for the help, the changes look good.

PR #664, opened 2022-11-02
Correction in column type retrieval in case of qualified table names

10bas10:
When using qualified table names like database.table, the retrieval from the schema dictionary failed. 

I changed the way the lookup takes place by taking into account the depth of the schema.

### DIFF
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -163,22 +163,30 @@ class MappingSchema(Schema):
         visible = _nested_get(self.visible, *zip(self.supported_table_args, args))  # type: ignore
         return [col for col in columns if col in visible]  # type: ignore
 
-    def get_column_type(self, table, column):
+    def get_column_type(self, table: exp.Table | str, column: exp.Column) -> exp.DataType.Type:
         try:
-            table_args = [
-                table.text(p)
-                for p in self.supported_table_args or self._get_table_args_from_table(table)
-            ]
-            schema_type = (
-                _nested_get(
-                    self.schema, *zip(self.supported_table_args, table_args), raise_on_missing=False
+            if isinstance(table, exp.Table):
+                table_args = [
+                    table.text(p)
+                    for p in self.supported_table_args or self._get_table_args_from_table(table)
+                ]
+                table_schema = _nested_get(
+                    self.schema,
+                    *zip(self.supported_table_args, table_args),
+                    raise_on_missing=False,
                 )
-                .get(column.name)
-                .upper()
-            )
+            else:
+                table_schema = self.schema.get(table)
+
+            schema_type = table_schema.get(
+                column if isinstance(column, str) else column.name
+            ).upper()
+
             return self._convert_type(schema_type)
         except:
-            raise OptimizeError(f"Failed to get type for column {column.sql()}")
+            raise OptimizeError(
+                f"Failed to get type for column {column if isinstance(column, str) else column.sql()}"
+            )
 
     def _convert_type(self, schema_type: str) -> exp.DataType.Type:
         """
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -163,7 +163,9 @@ class MappingSchema(Schema):
         visible = _nested_get(self.visible, *zip(self.supported_table_args, args))  # type: ignore
         return [col for col in columns if col in visible]  # type: ignore
 
-    def get_column_type(self, table: exp.Table | str, column: exp.Column) -> exp.DataType.Type:
+    def get_column_type(
+        self, table: exp.Table | str, column: exp.Column | str
+    ) -> exp.DataType.Type:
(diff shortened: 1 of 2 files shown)


sqlglot/schema.py
```diff
@@ -165,10 +165,25 @@ def column_names(self, table: exp.Table | str, only_visible: bool = False) -> t.
 
     def get_column_type(self, table: exp.Table | str, column: exp.Column) -> exp.DataType.Type:
```
GeorgeSittas, line 166: `column` can be either an `exp.Column` or a `str`, so we should add the latter here.

sqlglot/schema.py
```diff
@@ -165,10 +165,25 @@ def column_names(self, table: exp.Table | str, only_visible: bool = False) -> t.
...
+                ]
+                table_schema = _nested_get(
+                    self.schema,
+                    *zip(self.supported_table_args, table_args),
+                    raise_on_missing=False,
+                )
+            else:
+                schema_type = self.schema.get(table, {})
+
+            schema_type = table_schema.get(column if isinstance(column, str) else column.name).upper()  # type: ignore
```
GeorgeSittas, line 181: There's an execution path (`table` is not an `exp.Table`) where `table_schema` is undefined here.

GeorgeSittas:
Just a couple of comments from me; it looks good.

There are some places where the schema module feels awkward to me.. For example, can't we simplify this somehow?
```python
x in self.supported_table_args or self._get_table_args_from_table(table)
```

It's a repeated pattern and at first glance it's not immediately obvious why that's necessary. Maybe I'll take a look and try to refactor stuff when I finish with the other PR.

sqlglot/schema.py
```diff
@@ -165,10 +165,25 @@ def column_names(self, table: exp.Table | str, only_visible: bool = False) -> t.
...
+                    table.text(p)
+                    for p in self.supported_table_args or self._get_table_args_from_table(table)
+                ]
+                table_schema = _nested_get(
+                    self.schema,
+                    *zip(self.supported_table_args, table_args),
+                    raise_on_missing=False,
+                )
+            else:
+                schema_type = self.schema.get(table, {})
```
barakalon, line 179: I don't think this line does anything, since we immediately set schema_type again? Unless I'm misreading something?

barakalon (approved):
Can we add tests for this?

sqlglot/schema.py
10bas10: Yes thank you, i made a typo in variable assignment here.

sqlglot/schema.py
10bas10: Indeed, added the str

GeorgeSittas (approved):

sqlglot/schema.py
```diff
@@ -163,12 +163,29 @@ def column_names(self, table: exp.Table | str, only_visible: bool = False) -> t.
...
         try:
-            schema_type = self.schema.get(table.name, {}).get(column.name).upper()  # type: ignore
+            if isinstance(table, exp.Table):
+                table_args = [
+                    table.text(p)
+                    for p in self.supported_table_args or self._get_table_args_from_table(table)
+                ]
+                table_schema = _nested_get(
+                    self.schema,
+                    *zip(self.supported_table_args, table_args),
```
tobymao, line 177: line 173 assumes table args can be empty, but then you use it here

perhaps better to do 

supported_table_args = self.supported_table_args or self._get

then you can use this throughout

sqlglot/schema.py
10bas10: Indeed, that was a bug for the str approach. Changed it now.

sqlglot/schema.py
10bas10: Done

sqlglot/schema.py
```diff
@@ -163,12 +163,29 @@ def column_names(self, table: exp.Table | str, only_visible: bool = False) -> t.
...
+                    raise_on_missing=False,
+                )
+            else:
+                table_schema = self.schema.get(table, {})
+
```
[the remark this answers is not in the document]
10bas10: Done

tests/test_schema.py
```diff
@@ -3,6 +3,7 @@
 from sqlglot import table
```
tobymao, line 3:
```suggestion
from sqlglot import exp, table
```

tests/test_schema.py
```diff
@@ -1,8 +1,9 @@
 import unittest
 
-from sqlglot import table
+from sqlglot import exp, table
 from sqlglot.dataframe.sql import types as df_types
 from sqlglot.schema import MappingSchema, ensure_schema
+from sqlglot import exp
```
tobymao, line 6:
```suggestion
```

tobymao (approved):

PR #665, opened 2022-11-02
MySQL SHOW commands

barakalon:
https://dev.mysql.com/doc/refman/8.0/en/show.html

This also adds `USE` to the base parser/generator.


sqlglot/dialects/mysql.py
```diff
@@ -12,6 +12,316 @@
 )
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
+from sqlglot.trie import in_trie, new_trie
+
```
barakalon, line 16: I didn't want to pollute `sqlglot.expressions` with all these expression types, especially since they're not really transpile-able.

barakalon: I guess some of them are transpile-able... hm

tests/test_parser.py
```diff
@@ -49,6 +49,12 @@ def test_command(self):
         self.assertEqual(expressions[1].sql(), "ADD JAR s3://a")
         self.assertEqual(expressions[2].sql(), "SELECT 1")
 
+    def test_use(self):
```
tobymao, line 52: this just needs to be an identity test

barakalon:
@tobymao this is ready for another review

tobymao:
ok, will take a look later thanks for taking another stab at this

sqlglot/parser.py
```diff
@@ -2512,6 +2528,26 @@ def _parse_wrapped_id_vars(self):
...
 
+    def _parse_use(self):
+        return self.expression(exp.Use, this=self._parse_id_var())
+
+    def _parse_show(self):
+        this = []
+        trie = self._show_trie
+        while True:
+            # The current token might be multiple words
+            key = self._curr.text.split(" ")
```
tobymao, line 2539: do we need upper() here?

tobymao (approved):

PR #667, opened 2022-11-03
support alternate implementation of DateTrunc function

ginter:
The current sqlglot implementation of DATE_TRUNC assumes the timestamp is the first argument, but many dialects implement a version where the arguments are reversed as well as having the "unit" argument being just a string.

Since only bigquery uses the existing signature, we will use the new signature format as the default and implement a custom parser for bigquery where the arguments are reversed.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2419,8 +2419,8 @@ class DateDiff(Func, TimeUnit):
     arg_types = {"this": True, "expression": True, "unit": False}
 
 
-class DateTrunc(Func, TimeUnit):
-    arg_types = {"this": True, "unit": True, "zone": False}
+class DateTrunc(Func):
+    arg_types = {"this": True, "expression": True, "zone": False}
 
 
 class DatetimeAdd(Func, TimeUnit):
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -23,6 +23,13 @@ def _date_add(expression_class):
     return func
 
 
+def _date_trunc(args):
+    unit = seq_get(args, 1)
+    if isinstance(unit, exp.Column):
+        unit = exp.Var(this=unit.name)
+    return exp.DateTrunc(this=seq_get(args, 0), expression=unit)
+
+
 def _date_add_sql(data_type, kind):
     def func(self, expression):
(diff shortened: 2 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -2823,7 +2828,9 @@ def _norm_arg(arg):
     return arg.lower() if isinstance(arg, str) else arg
 
 
-ALL_FUNCTIONS = subclasses(__name__, Func, (AggFunc, Anonymous, Func))
+CONFLICTING_FUNCTIONS = (DateTrunc2,)
```
tobymao, line 2831: what is this?

sqlglot/dialects/starrocks.py
```diff
@@ -3,9 +3,16 @@
 from sqlglot import exp
 from sqlglot.dialects.dialect import arrow_json_extract_sql, rename_func
 from sqlglot.dialects.mysql import MySQL
+from sqlglot.expressions import DateTrunc2
```
tobymao, line 6: just do exp.DateTrunc2?

tobymao:
@ginter i don't understand, i don't think DATE_TRUNC has any special handling. why don't you just use the same class?

sqlglot/dialects/starrocks.py
tobymao: i'm not sure i understand why you need datetrunc2 in the first place

barakalon:
There should be one DateTrunc function. It's args are a superset of all dialects, but its representation is specific to sqlglot's abstract syntax.

Dialects that swap the arguments should handle this in their Parsers/Generators.

Otherwise transpilation becomes a nightmare.

_(he says after putting up a PR like [this](https://github.com/tobymao/sqlglot/pull/665))_

ginter:
The problem is that DateTrunc inherits from TimeUnit which does this:
```
class TimeUnit(Expression):
          if isinstance(unit, Column):
            args["unit"] = Var(this=unit.name)
```

this doesn't affect the generator much but it prevents the optimizer from properly categorizing the columns for eg DATE_TRUNC('week', ds)

ginter:
spoke with Toby offline, will just inline the alternate form in the dialects it's needed

tobymao (approved):

PR #669, opened 2022-11-04
implement null safe equal <=> operator

ginter:
NULL-safe equal. This operator performs an equality comparison like the [=]noperator, but returns 1 rather than NULL if both operands are NULL, and 0 rather than NULL if one operand is NULL.

The [<=>] operator is equivalent to the standard SQL IS NOT DISTINCT FROM operator.
https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to


sqlglot/tokens.py
```diff
@@ -423,6 +424,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "<=": TokenType.LTE,
         "<>": TokenType.NEQ,
         "!=": TokenType.NEQ,
+        "<=>": TokenType.NULLSAFE_EQ,
```
tobymao, line 427: also add "IS NOT DISTINCT FROM" ?

tests/test_transpile.py
```diff
@@ -350,3 +350,7 @@ def unsupported(level):
         with self.assertRaises(UnsupportedError) as ctx:
             unsupported(ErrorLevel.IMMEDIATE)
         self.assertEqual(str(ctx.exception).count(error), 1)
+
+    def test_nullsafe_eq(self):
```
tobymao, line 354: let's not put this here, let's put it in dialects and then test both ways

sqlglot/tokens.py
tobymao: there's also is distinct from what is that?

sqlglot/optimizer/simplify.py
```diff
@@ -257,6 +257,12 @@ def _simplify_binary(expression, a, b):
                 return TRUE if not_ else FALSE
             if a == NULL:
                 return FALSE if not_ else TRUE
+    elif isinstance(expression, exp.NullSafeEQ):
+        if a == b:
```
tobymao, line 261: is this correct? how do you know if column a == column b? you don't know the values

sqlglot/parser.py
```diff
@@ -1632,12 +1633,16 @@ def _parse_range(self):
...
-        this = self.expression(
-            exp.Is,
-            this=this,
-            expression=self._parse_null() or self._parse_boolean(),
-        )
-        return self.expression(exp.Not, this=this) if negate else this
+        if self._match(TokenType.DISTINCT_FROM):
+            klass = exp.NullSafeEQ if negate else exp.NullSafeNEQ
+            return self.expression(klass, this=this, expression=self._parse_expression())
+        else:
```
tobymao, line 1639: you can remove the else now

sqlglot/tokens.py
```diff
@@ -145,6 +146,7 @@ class TokenType(AutoName):
     DESCRIBE = auto()
     DETERMINISTIC = auto()
     DISTINCT = auto()
+    DISTINCT_FROM = auto()
```
ginter, line 149: unsure if it's preferable to piggy back off of the `IS` or expand both operators fully here

tobymao: this is good

sqlglot/optimizer/simplify.py
ginter: I was following the same logic as the EQ case below it except NULL <=> NULL is TRUE so I put it before the null check
```
    if isinstance(expression, exp.EQ) and a == b:
        return TRUE
```

tobymao: brain fart, this looks correct, because a == b means that the column is the same

tobymao (approved):

PR #673, opened 2022-11-04
SET commands

barakalon:
MySQL SET commands.

This also adds MySQL's session parameters: @@

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -166,6 +166,7 @@ class MySQL(Dialect):
             "_UTF32": TokenType.INTRODUCER,
             "_UTF8MB3": TokenType.INTRODUCER,
             "_UTF8MB4": TokenType.INTRODUCER,
+            "@@": TokenType.SESSION_PARAMETER,
         }
 
     class Parser(parser.Parser):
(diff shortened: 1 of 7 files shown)


tobymao (approved):

mpf82:
``SET`` used to be parsed as a "command", correct? and as such, you could basically write anything following it.

This PR breaks parsing valid SET statements, e.g.

Postgres: 
```sql
SET application_name TO "app_admin"
SET SESSION application_name TO "app_admin"
SET LOCAL application_name = "app_admin"
``` 
Syntax: 
> SET [ SESSION | LOCAL ] configuration_parameter { TO | = } { value | 'value' | DEFAULT }
> SET [ SESSION | LOCAL ] TIME ZONE { value | 'value' | LOCAL | DEFAULT }

Reference: https://www.postgresql.org/docs/current/sql-set.html

---

Setting multiple values at once in TSQL:

``SET QUOTED_IDENTIFIER, ANSI_NULLS ON``

Reference: https://learn.microsoft.com/en-us/sql/t-sql/statements/set-statements-transact-sql?view=sql-server-ver16#considerations-when-you-use-the-set-statements

tobymao:
can you implement #2 for now?

barakalon:
I'll be out all day today and did this in a rush:
https://github.com/tobymao/sqlglot/pull/676

Feel free to take over/merge.

tobymao:
ok i'll take over thanks!

PR #674, opened 2022-11-07
Type inheritance for functions based on args

10bas10:
Made a new function to support type inheritance for specific functions that should only inherit from specific attributes. The function allows one of the following parameters:

- A string and that will refer to an entry within the args dictionary, the array of expressions or its single expression will be included in the inheritance of datatypes;
- An expression;
- An list of expressions.

From the list of expressions applicable, the datatype will be inherited.

### DIFF
--- a/sqlglot/optimizer/annotate_types.py
+++ b/sqlglot/optimizer/annotate_types.py
@@ -96,10 +96,10 @@ class TypeAnnotator:
         exp.DiToDate: lambda self, expr: self._annotate_with_type(expr, exp.DataType.Type.DATE),
         exp.Exp: lambda self, expr: self._annotate_with_type(expr, exp.DataType.Type.DOUBLE),
         exp.Floor: lambda self, expr: self._annotate_with_type(expr, exp.DataType.Type.INT),
-        exp.Case: lambda self, expr: self._annotate_custom(expr, "default", "ifs"),
-        exp.If: lambda self, expr: self._annotate_custom(expr, "true", "false"),
-        exp.Coalesce: lambda self, expr: self._annotate_custom(expr, "this", "expressions"),
-        exp.IfNull: lambda self, expr: self._annotate_custom(expr, "this", "expression"),
+        exp.Case: lambda self, expr: self._annotate_by_args(expr, "default", "ifs"),
+        exp.If: lambda self, expr: self._annotate_by_args(expr, "true", "false"),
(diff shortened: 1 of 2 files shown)


GeorgeSittas (commented):
Just a few comments. We should add some tests that cover other argument types too; for instance we should check the case where we have a list of expressions or a single expression, instead of just strings.

This seems like a reasonable addition, thanks for the PR!

sqlglot/optimizer/annotate_types.py
```diff
@@ -329,3 +339,28 @@ def _annotate_literal(self, expression):
...
+                if isinstance(arg_expr, list):
+                    expressions += arg_expr
+                else:
+                    expressions.append(arg_expr)
+            else:
+                expressions.append(arg)
+        last_datatype = None
+
+        for expr in expressions:
+            if hasattr(expr, "type"):
```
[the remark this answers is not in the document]
10bas10: None values are currently also coming in, but ill filter those at the earlier stage.

10bas10:
> 

I could also clean that part out as I dont see a usecase yet for it. What do you think?

GeorgeSittas:
> I could also clean that part out as I dont see a usecase yet for it. What do you think?

Yes, that could work too. My initial reaction was that strings would probably suffice for this function.

sqlglot/optimizer/annotate_types.py
```diff
@@ -329,3 +339,20 @@ def _annotate_literal(self, expression):
...
+
+    def _annotate_by_args(self, expression, *args):
+        self._annotate_args(expression)
+        expressions = []
+        for arg in args:
+            arg_expr = expression.args.get(arg)
+            if isinstance(arg_expr, list):
+                expressions += (expr for expr in arg_expr if expr)
+            elif arg_expr:
+                expressions.append(arg_expr)
```
GeorgeSittas, line 351: What about the following? In any case, I think we should prefer `extend` over `+=` here, because that's more consistent with sqlglot's codebase.

```suggestion
        for arg in args:
            arg_expr = expression.args.get(arg)
            expressions.extend(expr for expr in ensure_list(arg_expr) if expr)
```

GeorgeSittas (approved):

sqlglot/optimizer/annotate_types.py
10bas10: Seems cleaner indeed

PR #676, opened 2022-11-07
Make SET a Command by default

barakalon:
Kinda janky.

See https://github.com/tobymao/sqlglot/pull/673 for context.

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -169,9 +169,7 @@ class MySQL(Dialect):
             "@@": TokenType.SESSION_PARAMETER,
         }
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -169,6 +169,10 @@ class Tokenizer(tokens.Tokenizer):
             "@@": TokenType.SESSION_PARAMETER,
         }
 
+        COMMANDS = {*tokens.Tokenizer.COMMANDS}
```
tobymao, line 172: commands = COMMANDS - {SET, SHOW} ?

tobymao (approved):

PR #677, opened 2022-11-08
Add support for snowflake's flatten function

### LINKED ISSUE #675: Difficulty parsing snowflake query with `split()` nested in `flatten()`
Hello!

I've been unable to get sqlglot to parse the following query and I'm not sure if it's user error or a bug? Would love to get this resolved:

```python

import sqlglot

query = f"""
select
  dag_report.acct_id,
  dag_report.report_date,
  dag_report.report_uuid,
  dag_report.airflow_name,
  dag_report.dag_id,
  f.value::varchar as operator
from cs.telescope.dag_report,
table(flatten(input=>split(operators, ','))) f
"""

def get_query_columns(query):
    column_names = []
    for expression in sqlglot.parse_one(query, read="snowflake").find(exp.Select).args["expressions"]:
        if isinstance(expression, exp.Alias):
            column_names.append(expression.text("alias"))
        elif isinstance(expression, exp.Column):
            column_names.append(expression.text("this"))

    return column_names

if __name__ == '__main__':
    column_names = get_query_columns(query)
    print(column_names)

```

Error:

```text
sqlglot.errors.ParseError: Required keyword: 'this' missing for <class 'sqlglot.expressions.GT'>. Line 10, Col: 43.
```

Any help would be greatly appreciated. 🍻

GeorgeSittas:
fixes #675

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -71,18 +71,6 @@ def _unix_to_time_sql(self, expression):
     raise ValueError("Improper scale for timestamp")
 
 
-def _flatten_sql(self, expression):
-    args = [
-        f"{'INPUT' if key == 'this' else key.upper()} => {self.sql(arg)}"
-        for key, arg in expression.args.items()
-    ]
-
-    if self.pretty and self.text_width(args) > self._max_text_width:
-        nl = "\n"
-        return f"FLATTEN({self.indent({nl} + f',{nl}'.join(args) + {nl}, skip_first=True, skip_last=True)})"
-    return f'FLATTEN({", ".join(args)})'
-
-
 # https://docs.snowflake.com/en/sql-reference/functions/date_part.html
 # https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts
 def _parse_date_part(self):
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -71,6 +71,18 @@ def _unix_to_time(self, expression):
...
+def _flatten_sql(self, expression):
+    args = [
+        f"{'INPUT' if key == 'this' else key.upper()} => {self.sql(arg)}"
+        for key, arg in expression.args.items()
+    ]
+
+    if self.pretty and self.text_width(args) > self._max_text_width:
+        nl = "\n"
+        return f"FLATTEN({self.indent({nl} + f',{nl}'.join(args) + {nl}, skip_first=True, skip_last=True)})"
+    return f'FLATTEN({", ".join(args)})'
```
GeorgeSittas, line 83: I was trying to use `format_args` for this, because it's essentially the same logic, but with different sql formatting. Maybe we can dry this up by adding a `sql_lambda` argument to the `format_args` which takes care of the formatting, so we can reuse it for situations like this. For example, perhaps we could do:

```python
    def format_args(self, *args, sql_lambda=None):
        sql_lambda = sql_lambda or self.sql
        args = tuple(sql_lambda(arg) for arg in args if arg is not None)
        if self.pretty and self.text_width(args) > self._max_text_width:
            return self.indent("\n" + f",\n".join(args) + "\n", skip_first=True, skip_last=True)
        return ", ".join(args)
```
So then `_flatten_sql` would become:

```python
def _flatten_sql(self, expression):
    sql_lambda = lambda key_arg: f"{key_arg[0]} => {key_arg[1]}"
    return self.format_args(*[(key, arg) for key, arg in expression.args.items()], sql_lambda=sql_lambda)
```

tobymao:
i don't understand this change, why not just add => as arrow in snowflake and be done with it?

GeorgeSittas:
Right, it was more complex than needed. I wasn't aware that `->` is used to produce a lambda expression at the time.

I hope this time I got it right, let me know if I should fix anything.

PR #683, opened 2022-11-10
execute SQL on python objects

barakalon:
My brain is fried right now, so I should do another pass on cleaning this up tomorrow.

But I wanted to put up my progress so far.

### DIFF
--- a/sqlglot/executor/__init__.py
+++ b/sqlglot/executor/__init__.py
@@ -50,7 +50,7 @@ def _ensure_tables(tables):
         if isinstance(table, Table):
             result[name] = table
         else:
-            columns = tuple(table[0].keys()) if table else ()
+            columns = tuple(table[0]) if table else ()
             rows = [tuple(row[c] for c in columns) for row in table]
             result[name] = Table(columns=columns, rows=rows)
     return result
--- a/sqlglot/executor/__init__.py
+++ b/sqlglot/executor/__init__.py
@@ -3,13 +3,14 @@ import time
 
 from sqlglot import parse_one
 from sqlglot.executor.python import PythonExecutor
+from sqlglot.executor.table import Table
 from sqlglot.optimizer import optimize
 from sqlglot.planner import Plan
 
 logger = logging.getLogger("sqlglot")
 
 
-def execute(sql, schema, read=None):
+def execute(sql, schema, read=None, tables=None):
     """
     Run a sql query against data.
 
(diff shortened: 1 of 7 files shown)


sqlglot/executor/python.py
```diff
@@ -164,24 +181,38 @@ def join(self, step, context):
         condition = self.generate(step.condition)
         projections = self.generate_tuple(step.projections)
 
-        if not condition or not projections:
+        if not condition and not projections:
             return source_context
+        elif not projections:
```
barakalon, line 186: This all feels pretty messy to me.

Without projections, we need to keep the nested tables in the resulting context.

sqlglot/executor/__init__.py
```diff
@@ -23,9 +24,11 @@ def execute(sql, schema, read=None):
                 3. {catalog: {db: {table: {col: type}}}}
         read (str): the SQL dialect to apply during parsing
             (eg. "spark", "hive", "presto", "mysql").
+        tables (dict[str, list[dict]|sqlglot.executor.Table]): additional tables to register
     Returns:
         sqlglot.executor.Table: Simple columnar data structure.
     """
+    tables = _ensure_tables(tables)
```
tobymao, line 31: given tables, you should be able to infer the schema

sqlglot/executor/__init__.py
```diff
@@ -34,6 +37,20 @@ def execute(sql, schema, read=None):
...
+
+def _ensure_tables(tables):
+    result = {}
+    if not tables:
+        return result
+    for name, table in tables.items():
+        if isinstance(table, Table):
+            result[name] = table
+        else:
+            columns = tuple(table[0].keys()) if table else ()
```
tobymao, line 53: you don't need keys

sqlglot/executor/python.py
tobymao: i don't get this, you may have to explain it to me

sqlglot/executor/__init__.py
barakalon: Unless the tables are empty

sqlglot/executor/__init__.py
barakalon: 🤦 how many times have you told me this? lol

sqlglot/executor/python.py
barakalon: let's chat today

barakalon (commented):
@tobymao Alright I got this working

PR #684, opened 2022-11-10
Add transaction/commit expressions

GeorgeSittas:
I'm not very proud of the parser logic -- I'll try to see if I can clean it up tomorrow.

- [x] Fix bigquery `BEGIN` transpilation. It's currently transpiled (wrongly) to BEGIN TRANSACTION because we don't have logic for parsing it as a "block start" statement yet.


GeorgeSittas:
I'm leaving this sticky note I wrote up while investigating transaction syntax for reference. Might need to verify some of those, but it should be good. For the time being we can leave Oracle and T-SQL dialects out.

image

sqlglot/parser.py
```diff
@@ -2565,6 +2567,24 @@ def _parse_select_or_expression(self):
     def _parse_use(self):
         return self.expression(exp.Use, this=self._parse_id_var())
 
+    def _parse_transaction(self):
+        this = None
+        if (
+            self._match_text("DEFERRED")
+            or self._match_text("IMMEDIATE")
+            or self._match_text("EXCLUSIVE")
```
GeorgeSittas, line 2575: I don't like this much. Maybe a better approach would be to change `_match_text` so that it can check against a set of strings (kwarg) for a _single_ match.

sqlglot/parser.py
```diff
@@ -377,6 +377,8 @@ class Parser(metaclass=_Parser):
         TokenType.CACHE: lambda self: self._parse_cache(),
         TokenType.UNCACHE: lambda self: self._parse_uncache(),
         TokenType.USE: lambda self: self._parse_use(),
+        TokenType.START: lambda self: self._parse_transaction(),
```
tobymao, line 380: there shouldn't be a token start, i think dialects should just map start to begin

sqlglot/parser.py
tobymao: i guess add a new one that does an in clause

sqlglot/parser.py
```diff
@@ -2565,6 +2567,24 @@ def _parse_select_or_expression(self):
...
+        ):
+            this = self._prev.text
+
+        self._match_text("TRANSACTION") or self._match_text("WORK")
+
+        modes = []
+        while self._match(TokenType.VAR) or self._match(TokenType.COMMA):
+            modes.append(self._prev.text)
+
+        modes = " ".join(modes).split(" , ") if modes else None
```
tobymao, line 2585: i don't think we should convert it into a string can we map the different modes and make them first class?

sqlglot/tokens.py
```diff
@@ -259,6 +259,7 @@ class TokenType(AutoName):
     SOME = auto()
     SORT_BY = auto()
     STABLE = auto()
+    START = auto()
```
tobymao, line 262: we don't need this

sqlglot/parser.py
GeorgeSittas: I'm not sure I understood what you have in mind. So maybe create a "Mode" expression node and store a list of these with `this` set to the corresponding string?

What about `Transaction.this`? Do we want to also store that as a string literal node, or is it fine that way?

sqlglot/tokens.py
```diff
@@ -244,7 +244,6 @@ class TokenType(AutoName):
     RETURNS = auto()
     RIGHT = auto()
     RLIKE = auto()
-    ROLLBACK = auto()
```
tobymao, line 247: why did you get rid of rollback? roll back is separate from commit

sqlglot/dialects/bigquery.py
```diff
@@ -151,6 +151,11 @@ class Parser(parser.Parser):
             TokenType.TABLE,
         }
 
+        def _parse_transaction(self):
+            if self._match_text_seq("TRANSACTION"):
+                return self.expression(exp.Transaction)
+            return self.expression(exp.Command, this=self._prev.text)
```
tobymao, line 157: what is this one for?

GeorgeSittas: In bigquery BEGIN is valid and different from BEGIN TRANSACTION, as it signifies the start of a block statement.

tobymao: ok put a comment here for that and add a test

GeorgeSittas: Done, there's also one test in the parser from before.

sqlglot/parser.py
GeorgeSittas: discussed in slack

GeorgeSittas:
I think this is good to go. Let me know if there's anything I missed.

sqlglot/parser.py
```diff
@@ -2565,6 +2570,37 @@ def _parse_select_or_expression(self):
...
+            mode = []
+            while self._match(TokenType.VAR):
+                mode.append(self._prev.text)
+
+            if mode:
+                modes.append(" ".join(mode))
+            if not self._match(TokenType.COMMA):
+                break
+
+        return self.expression(exp.Transaction, this=this, modes=modes or None)
```
tobymao, line 2591: i don't think you need to pass None here

PR #686, opened 2022-11-10
optionally check whether value is a function when parsing data type

ginter:
the current implementation is overly permissive for what it parses as a data type. this change tries to further disambiguate type literals of the form `TIMESTAMP(3) '2022-01-01'` | `DATE(1) '2022-01-01'` from functions `TIMESTAMP('2022-01-01')` | `DATE('2022-01-01')` | `BOOLEAN(1)` etc.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1820,8 +1820,15 @@ class Parser(metaclass=_Parser):
                     this=exp.DataType.Type.TIMESTAMPLTZ,
                     expressions=expressions,
                 )
-            else:
-                self._match(TokenType.WITHOUT_TIME_ZONE)
+            elif self._match(TokenType.WITHOUT_TIME_ZONE):
+                value = exp.DataType(
+                    this=exp.DataType.Type.TIMESTAMP,
+                    expressions=expressions,
+                )
+
+            maybe_func = maybe_func and value is None
+
+            if value is None:
                 value = exp.DataType(
                     this=exp.DataType.Type.TIMESTAMP,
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -1804,27 +1806,39 @@ def _parse_types(self):
...
+                    expressions=expressions,
+                )
 
-            return exp.DataType(
-                this=exp.DataType.Type.TIMESTAMP,
-                expressions=expressions,
-            )
+        if maybe_func and check_func:
+            index2 = self._index
+            peek = self._parse_column()
```
tobymao, line 1832: any way we can do an actual peek here instead of a full column parse?  

can you just check the token type of self._curr token? what is the case where we don’t have a column

what is the case where you know

tests/dialects/test_dialect.py
```diff
@@ -590,6 +590,22 @@ def test_time(self):
                 "spark": "DATE_ADD(CAST('2020-01-01' AS DATE), 1)",
             },
         )
+        self.validate_all(
+            "TIMESTAMP '2022-01-01'",
```
tobymao, line 594: can we add a test of that case you talked about like time(1) ‘2022’

sqlglot/parser.py
ginter: after looking, every case I've found of a type literal that optionally takes a modifier - so it's ambiguous whether it's a function - has a string for the expression

tobymao (approved):

PR #688, opened 2022-11-10
Created an generic approach to concat symbol

10bas10:
Because of the + symbol in T-SQL with different kind of operation methods I have implemented an approach to replace the ADD class with a DPipe class in case of the context of string concatenation.

The two cases applicable in T-SQL:
- 'test' + 'string' = 'teststring'
- 1 + 2 = 3

As @tobymao already said earlier its not really something that fits in the parser part. Also this requires to use the optimizer to assign datatypes to columns from schema. By implementing it in the transforms.py file, we are able to use this at several target dialects.

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -81,3 +81,15 @@ def delegate(attr: str) -> t.Callable:
 
 
 UNALIAS_GROUP = {exp.Group: preprocess([unalias_group], delegate("group_sql"))}
+ADD_TO_DPIPE = {
+    exp.Add: lambda self, e: exp.DPipe(this=e.this, expression=e.expression)
+    if e.type
+    in [
+        exp.DataType.Type.VARCHAR,
+        exp.DataType.Type.NVARCHAR,
+        exp.DataType.Type.CHAR,
+        exp.DataType.Type.NCHAR,
+        exp.DataType.Type.TEXT,
+    ]
+    else e
+}
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -82,7 +82,7 @@ def delegate(attr: str) -> t.Callable:
 
 UNALIAS_GROUP = {exp.Group: preprocess([unalias_group], delegate("group_sql"))}
 ADD_TO_DPIPE = {
-    exp.Add: lambda self, e: exp.DPipe(this=e.this, expression=e.expression)
+    exp.Add: lambda self, e: self.dpipe_sql(e)
     if e.type
     in [
         exp.DataType.Type.VARCHAR,
@@ -91,5 +91,5 @@ ADD_TO_DPIPE = {
         exp.DataType.Type.NCHAR,
         exp.DataType.Type.TEXT,
     ]
-    else e
+    else self.add_sql(e)
 }
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -80,16 +80,24 @@ def delegate(attr: str) -> t.Callable:
     return _transform
 
 
-UNALIAS_GROUP = {exp.Group: preprocess([unalias_group], delegate("group_sql"))}
-ADD_TO_DPIPE = {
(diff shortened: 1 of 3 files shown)


10bas10:
Testcase coming up

sqlglot/transforms.py
```diff
@@ -81,3 +81,15 @@ def _transform(self, *args, **kwargs):
...
+ADD_TO_DPIPE = {
+    exp.Add: lambda self, e: self.dpipe_sql(e)
+    if e.type
+    in [
+        exp.DataType.Type.VARCHAR,
+        exp.DataType.Type.NVARCHAR,
+        exp.DataType.Type.CHAR,
+        exp.DataType.Type.NCHAR,
+        exp.DataType.Type.TEXT,
+    ]
```
GeorgeSittas, line 93: Maybe we could export some helper functions to help us identify an expression's general type class (e.g if it's a string, a numeric, date etc), assuming we've already inferred its type. I think that might be useful for other type inference applications we might have in the future.

Otherwise, If that's not something we should prioritize right now, you could convert that list into a set.

10bas10: Indeed, would be nice to have such a supertype available. For now i converted it into a set. Also i rewrote the approach a bit to also be able to use the function within test_transforms.py.

sqlglot/transforms.py
```diff
@@ -80,4 +80,24 @@ def _transform(self, *args, **kwargs):
...
+
+def transform_add(self, e):
+    expr = add_or_dpipe(e)
+    delegate_func = f"{expr.key}_sql"
+    to_sql = delegate(delegate_func)
+    return to_sql(self, e)
+
+
 UNALIAS_GROUP = {exp.Group: preprocess([unalias_group], delegate("group_sql"))}
+ADD_TO_DPIPE = {exp.Add: transform_add}
```
tobymao, line 103: no one really uses types yet so i don't think this should be a mandatory transform,  let's delete ADD_TO_DIPE and transform_add

and then in your own code, you can call query.transform(add_or_pipe)

sqlglot/transforms.py
```diff
@@ -80,4 +80,24 @@ def _transform(self, *args, **kwargs):
     return _transform
 
 
+def add_or_dpipe(expression):
```
tobymao, line 83:
```suggestion
def add_char_to_dpipe(expression):
```

sqlglot/transforms.py
GeorgeSittas: > For now i converted it into a set.

I think you wrote it as a tuple instead. Was this a typo?

sqlglot/transforms.py
tobymao: another idea is to put this in the annotator as a post annotation transform

barakalon:
I think we can have a `canonicalize` optimizer step that can do this type of transformation.

sqlglot/transforms.py
```diff
@@ -80,4 +80,24 @@ def _transform(self, *args, **kwargs):
...
+    if isinstance(expression, exp.Add) and expression.type in (
+        exp.DataType.Type.VARCHAR,
+        exp.DataType.Type.NVARCHAR,
+        exp.DataType.Type.CHAR,
+        exp.DataType.Type.NCHAR,
+        exp.DataType.Type.TEXT,
+    ):
+        expression = exp.DPipe(this=expression.this, expression=expression.expression)
```
barakalon, line 91: Which is more standard, `||` or `CONCAT`?

tobymao: probably concat

tobymao:
@10bas10 this is quite difficult, i'm doing this now. i'll close this for now and implement this for you

tobymao:
https://github.com/tobymao/sqlglot/pull/698

tobymao:
https://github.com/tobymao/sqlglot/commit/169ba000031f3655e49e256adbcbd42bd4b87ca5

PR #697, opened 2022-11-13
Fix subquery after cte issue

### LINKED ISSUE #696: Parentheses in WITH clause cause an parse error
Hi, 
I am trying to use the library to parse some of my BIGQUERY queries and I found a parse error in the some of them.
The following is a **valid** query:

`
WITH sub_query AS (
    SELECT  a FROM table
)
(SELECT a FROM sub_query)
`

However it fails with the following error:
**sqlglot.errors.ParseError: subquery does not support CTE. Line 6, Col: 12.**

It seems like the parentheses around the last query causes this issue

Any thoughts how hard would be to solve it?

GeorgeSittas:
Fixes #696


tobymao:
what about nested parens

GeorgeSittas:
> what about nested parens

Hmm that's a good point, it currently does not handle them. I'll take a look, maybe we should allow CTEs before subquries in the parser.

tobymao (approved):

PR #706, opened 2022-11-15
Executor: support static queries

barakalon:


sqlglot/executor/python.py
```diff
@@ -126,6 +128,9 @@ def scan(self, step, context):
 
         return self.context({step.name: sink})
 
+    def static(self):
+        yield RowReader(()), self.context({})
```
tobymao, line 132: wow so easy?

sqlglot/executor/python.py
```diff
@@ -89,13 +89,15 @@ def table(self, expressions):
     def scan(self, step, context):
         source = step.source
 
-        if isinstance(source, exp.Expression):
+        if source and isinstance(source, exp.Expression):
```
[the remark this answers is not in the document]
barakalon: nope

sqlglot/executor/python.py
barakalon: Haha yeah. This PR felt great.

tobymao (approved):

PR #707, opened 2022-11-15
Add Apache Drill Dialect to SqlGlot

cgivre:
This PR adds a dialect for Apache Drill to SQLGlot. 

Remaining work:

- [ ] Add support for Drill `table()` function
- [ ] Add unit tests for `table()` function.

### DIFF
--- a/sqlglot/dialects/drill.py
+++ b/sqlglot/dialects/drill.py
@@ -6,7 +6,7 @@ from sqlglot.dialects.dialect import (
     create_with_partitions_sql,
     format_time_lambda,
     rename_func, str_position_sql,
-    no_trycast_sql
+    no_trycast_sql, no_pivot_sql
 )
 from sqlglot.dialects.postgres import _lateral_sql
 from sqlglot.tokens import TokenType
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/drill.py
```diff
@@ -0,0 +1,174 @@
...
+        "''T''": "T",
+    }
+
+    class Tokenizer(tokens.Tokenizer):
+        QUOTES = ["'"]
+        IDENTIFIERS = ["`"]
+        ESCAPES = ["\\"]
+        ENCODE = "utf-8"
+
+        KEYWORDS = {**tokens.Tokenizer.KEYWORDS, "VARBINARY": TokenType.BINARY}
```
tobymao, line 115: why did you add varbinary? that's already a type

tobymao:
this is a pretty good start, why don't we get this merged in and then fix the xlsx table afterwards?

cgivre:
> this is a pretty good start, why don't we get this merged in and then fix the xlsx table afterwards?

@tobymao I'd defer to you on that.   Drill's SQL dialect follows ANSI 2003 pretty closely, so I don't anticipate there's much more do to.  I'm happy to proceed if that's what you think is best.  Going to sleep now, but tomorrow, I'll remove the unit test for the `table()` function.

tobymao:
ok yea, let's just get this merged. it's nice and clean. the table function is going to be much harder

sqlglot/dialects/drill.py
cgivre: Removed.  Drill doesn't have a `BINARY` data type, just a `VARBINARY`, so that's what I was trying to do here, but I think that is covered elsewhere.

cgivre: Gotcha.... and it's already there. 👍 

https://github.com/tobymao/sqlglot/blob/f1f4fa2dfd9e20277e8496623c339ef9ee34b687/sqlglot/dialects/drill.py#L129-L132

PR #708, opened 2022-11-15
Add support for aggregates without the GROUP BY clause

### LINKED ISSUE #703: execute(): Exception when aggregating without GROUP BY
sqlglot 10.0.2, Python 3.8.4

`>>> execute('select sum(x) from t', tables={'t': [{'x': 1}, {'x': 2}]})`

Result:

`sqlglot.errors.ExecuteError: Step 'Scan: t (140100782263984)' failed: '_col_0'`

If I add column 'y' above, with the same value for every row, and group by that column, the execute returns as expected.

GeorgeSittas:
Fixes #703

### DIFF
--- a/sqlglot/executor/context.py
+++ b/sqlglot/executor/context.py
@@ -1,4 +1,11 @@
+from __future__ import annotations
+
+import typing as t
+
 from sqlglot.executor.env import ENV
+
+if t.TYPE_CHECKING:
+    from sqlglot.executor.table import Table, TableIter
 
 
 class Context:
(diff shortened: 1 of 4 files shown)


sqlglot/executor/context.py
```diff
@@ -31,18 +38,18 @@ def eval_tuple(self, codes):
         return tuple(self.eval(code) for code in codes)
 
     @property
-    def table(self):
+    def table(self) -> Table:
         if self._table is None:
-            self._table = list(self.tables.values())[0]
+            self._table = list(self.tables.values())[0]  # type: ignore
```
tobymao, line 43: don't do type ignores

sqlglot/planner.py
```diff
@@ -150,15 +150,15 @@ def from_expression(
...
+        if group or aggregations:
             aggregate = Aggregate()
             aggregate.source = step.name
             aggregate.name = step.name
             aggregate.operands = tuple(
                 alias(operand, alias_) for operand, alias_ in operands.items()
             )
             aggregate.aggregations = aggregations
-            aggregate.group = group.expressions
+            aggregate.group = group.expressions if group else []
```
tobymao, line 161: don't need this one

PR #709, opened 2022-11-16
Executor/MySQL fixes

barakalon:
An assortment of fixes/additions from getting the executor to worth the mysql-mimic (https://github.com/kelsin/mysql-mimic/pull/18)

### DIFF
--- a/sqlglot/executor/env.py
+++ b/sqlglot/executor/env.py
@@ -1,6 +1,8 @@
 import datetime
+import inspect
 import re
 import statistics
+from functools import wraps
 
 from sqlglot.helper import PYTHON_VERSION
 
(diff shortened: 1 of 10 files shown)


sqlglot/executor/env.py
```diff
@@ -16,6 +16,41 @@ def __lt__(self, other):
...
 
+def filter_nulls(func):
+    def _func(values):
+        return func(v for v in values if v is not None)
+
+    return _func
+
+
+def str_position(substr, this, position=None):
+    if substr is None or this is None:
```
tobymao, line 27: is there a way we can have the helper do this as well? maybe like a decorator to return None if any arg is None?

PR #717, opened 2022-11-16
Executor: fix sort on non-projections

barakalon:


sqlglot/executor/python.py
```diff
@@ -286,23 +286,29 @@ def aggregate(self, step, context):
...
+        sort_ctx.sort(self.generate_tuple(step.key))
 
         if not math.isinf(step.limit):
-            context.table.rows = context.table.rows[0 : step.limit]
+            sort_ctx.table.rows = sort_ctx.table.rows[0 : step.limit]
 
-        return self.context({step.name: context.table})
+        output = Table(
+            projection_columns,
+            rows=[r[len(context.columns) : len(all_columns)] for r in sort_ctx.table.rows],
```
barakalon, line 309: This isn't the most efficient thing to do... but I couldn't immediately see a better solution that didn't involve more fundamental changes.

sqlglot/executor/python.py
```diff
@@ -286,23 +286,29 @@ def aggregate(self, step, context):
     def sort(self, step, context):
         projections = self.generate_tuple(step.projections)
 
-        sink = self.table(step.projections)
+        projection_columns = [p.alias_or_name for p in step.projections]
+        all_columns = list(context.columns) + projection_columns
+        sink = self.table(all_columns)
```
tobymao, line 291: what if they columns have the same name? do you need to use column_range?

sqlglot/executor/python.py
tobymao: what does this do and what is the more fundamental change ?

sqlglot/executor/python.py
barakalon: Projections get the `None` table in `sort_ctx`. Everything else is nested by the table name. (You wrote that logic haha)

And the optimizer handles qualifying columns for us. So it just works :)



I added a test to demonstrate what happens.

```sql
SELECT 
  100 - ORD(a) AS a, 
  a AS i 
FROM x 
ORDER BY a
```

returns:
```python
[(1, "c"), (2, "b"), (3, "a")]
```

sqlglot/executor/python.py
barakalon: This copies `rows`, keeping only the projections columns.

A more fundamental change would let us do this without copying. I don't know exactly how to do that yet. `column_range` is a similar idea, but it doesn't work as-is.

PR #718, opened 2022-11-16
Small comment formatting fixup

GeorgeSittas:
This doesn't address a serious issue; it's more of a "formatting one". It aims to address examples similar to the following:

```python
>>> sql = """
... select
...   * /* multi
... line */
... """
>>> print(sqlglot.transpile(sql, pretty=True)[0])
SELECT
  */* multi
  line */
```
With this change, it's instead formatted as:
```python
SELECT
  * /* multi
  line */
```

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -241,7 +241,7 @@ class Generator:
         if not NEWLINE_RE.search(comment):
             return f"{sql} --{comment.rstrip()}" if single_line else f"{sql} /*{comment}*/"
 
-        return f" /*{comment}*/" if not sql else f"/*{comment}*/\n{sql}"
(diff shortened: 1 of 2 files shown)


sqlglot/generator.py
```diff
@@ -241,7 +241,7 @@ def maybe_comment(self, sql, expression, single_line=False):
         if not NEWLINE_RE.search(comment):
             return f"{sql} --{comment.rstrip()}" if single_line else f"{sql} /*{comment}*/"
 
-        return f"/*{comment}*/\n{sql}"
+        return f" /*{comment}*/" if not sql else f"/*{comment}*/\n{sql}"
```
tobymao, line 244:
```suggestion
        return  f"/*{comment}*/\n{sql}" if sql else f" /*{comment}*/"
```

tobymao (approved):

PR #725, opened 2022-11-17
Added ability to parse Redshift DISTKEY, SORTKEY, DISTSTYLE

yuchaofan13:
Example syntax: CREATE TABLE SOUP DISTKEY(soup1) SORTKEY (soup2) DISTSTYLE AUTO

For now I've created individual properties for each and added these to ROOT_PROPERTIES in the base generator. However, distkey and sortkey both use parentheses- not sure where I should be putting their properties instead. Also not sure whether I should have something like 'allow_distkey = True' in the Redshift dialect subclass.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1182,6 +1182,9 @@ class Generator:
         zone = self.sql(expression, "this")
         return f"CURRENT_DATE({zone})" if zone else "CURRENT_DATE"
 
+    def collate_sql(self, expression):
+        return self.binary(expression, "COLLATE")
+
     def command_sql(self, expression):
         return f"{self.sql(expression, 'this').upper()} {expression.text('expression').strip()}"
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -94,6 +94,9 @@ class Generator:
     ROOT_PROPERTIES = {
         exp.ReturnsProperty,
         exp.LanguageProperty,
+        exp.DistStyleProperty,
+        exp.DistKeyProperty,
+        exp.SortKeyProperty,
     }
 
     WITH_PROPERTIES = {
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -158,6 +158,8 @@ class Parser(metaclass=_Parser):
         TokenType.DELETE,
         TokenType.DESCRIBE,
         TokenType.DETERMINISTIC,
+        TokenType.DISTKEY,
+        TokenType.DISTSTYLE,
         TokenType.EXECUTE,
         TokenType.ENGINE,
         TokenType.ESCAPE,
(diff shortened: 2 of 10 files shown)


sqlglot/generator.py
```diff
@@ -94,6 +94,9 @@ class Generator:
     ROOT_PROPERTIES = {
         exp.ReturnsProperty,
         exp.LanguageProperty,
+        exp.DistStyleProperty,
+        exp.DistKeyProperty,
+        exp.SortKeyProperty,
```
GeorgeSittas, line 99: Should we override `ROOT_PROPERTIES` for Redshift and add them there?

sqlglot/parser.py
```diff
@@ -452,6 +456,9 @@ class Parser(metaclass=_Parser):
         TokenType.PARTITIONED_BY: lambda self: self._parse_partitioned_by(),
         TokenType.SCHEMA_COMMENT: lambda self: self._parse_schema_comment(),
         TokenType.STORED: lambda self: self._parse_stored(),
+        TokenType.DISTKEY: lambda self: self._parse_distkey(),
+        TokenType.DISTSTYLE: lambda self: self._parse_diststyle(),
+        TokenType.SORTKEY: lambda self: self._parse_sortkey(),
```
GeorgeSittas, line 461: Same here: should we add these to Redshift only?

sqlglot/parser.py
```diff
@@ -846,6 +857,37 @@ def _parse_stored(self):
...
+        )
+
+    def _parse_diststyle(self):
+        this = exp.Literal.string("DISTSTYLE")
+        value = exp.Literal.string(self._parse_var().name)
+        return self.expression(
+            exp.DistStyleProperty,
+            this=this,
+            value=value,
+        )
```
GeorgeSittas, line 889: Same

GeorgeSittas (commented):
I left a few comments.

By the way, I think the distkey/sortkey properties you can have a compound key. This is documented [here](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_examples.html), so we might want to change how we parse / represent these two?

sqlglot/generator.py
tobymao: yea let's add these to redshift

sqlglot/parser.py
tobymao: probably ok to be here

GeorgeSittas:
@yuchaofan13 regarding what I said earlier:

> By the way, I think the distkey/sortkey properties can have a compound key. This is documented [here](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_examples.html), so we might want to change how we parse / represent these two?

It's ok if we push your changes for now. I'll try to refactor the way properties are handled and I can look into that too.

yuchaofan13:
@GeorgeSittas nice, thank you! My understanding of the codebase is still quite lacking and I wasn't sure how to modify the properties so will be curious to see the refactoring! (this is my first ever pull request hahaha) I hadn't actually heard of the compound key before but I have some other features in mind too.

GeorgeSittas:
> so will be curious to see the refactoring

@yuchaofan13 I've opened a PR about this: https://github.com/tobymao/sqlglot/pull/733.

PR #726, opened 2022-11-17
Add support for HEX / UNHEX and ENCODE / DECODE to Presto and Spark

izeigerman:


sqlglot/dialects/presto.py
```diff
@@ -106,6 +117,11 @@ def _ts_or_ds_add_sql(self, expression):
     return f"DATE_ADD({unit}, {e}, DATE_PARSE(SUBSTR({this}, 1, 10), {Presto.date_format}))"
 
 
+def _ensure_utf8(charset):
+    if charset.text("this").lower() != "utf-8":
```
tobymao, line 121:
```suggestion
    if charset.name.lower() != "utf-8":
```

sqlglot/dialects/presto.py
```diff
@@ -106,6 +117,11 @@ def _ts_or_ds_add_sql(self, expression):
...
+        raise UnsupportedError(f"Unsupported charset {charset.sql()}")
```
tobymao, line 122:
```suggestion
        raise UnsupportedError(f"Unsupported charset {charset}")
```

sqlglot/expressions.py
```diff
@@ -2623,6 +2631,10 @@ class GroupConcat(Func):
     arg_types = {"this": True, "separator": False}
 
 
+class Hex(Func):
+    arg_types = {"this": True}
```
GeorgeSittas, line 2635: We can `pass` here, because `arg_types` is already defined like this in `Expression` (same for `Unhex`).

GeorgeSittas (approved):

sqlglot/expressions.py
izeigerman: good point,thanks

PR #727, opened 2022-11-17
Executor set operations

barakalon:


sqlglot/planner.py
```diff
@@ -346,3 +354,45 @@ def _to_s(self, indent: str) -> t.List[str]:
             lines.append(f"{indent}  - {expression.sql()}")
 
         return lines
+
+
+class SetOperation(Step):
```
barakalon, line 359: I'm not sure which is more elegant - 
One `SetOperation` class or three classes: `Union`, `Except`, and `Intersect`.

tobymao (approved):
incredible

PR #733, opened 2022-11-19
Refactor properties

GeorgeSittas:
Opening this for some early feedback. The end goal is to simplify the way properties are handled, so that we don't need to store each prop name in `this`. The parser can also be simplified because all properties are more or less parsed according to the same logic: `PROP_NAME [=] PROP_VALUE`.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1081,7 +1081,7 @@ class DistKeyProperty(Property):
 
 
 class SortKeyProperty(Property):
-    arg_types = {"this": True}
+    arg_types = {"this": True, "compound": False}
 
 
 class DistStyleProperty(Property):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1132,21 +1132,21 @@ class Properties(Expression):
     arg_types = {"expressions": True}
 
     NAME_TO_PROPERTY = {
-        "TABLE_FORMAT": TableFormatProperty,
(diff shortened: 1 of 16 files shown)


GeorgeSittas:
Leaving this [sticky note](https://user-images.githubusercontent.com/46752250/202861029-80b57a15-78d1-46ba-b30b-1fc8af081721.png) for reference.

sqlglot/expressions.py
```diff
@@ -1135,27 +1131,36 @@ class VolatilityProperty(Property):
...
-        "CHARACTER_SET": CharacterSetProperty,
-        "COLLATE": CollateProperty,
-        "COMMENT": SchemaCommentProperty,
-        "ENGINE": EngineProperty,
-        "FORMAT": FileFormatProperty,
-        "LOCATION": LocationProperty,
-        "PARTITIONED_BY": PartitionedByProperty,
+    NAME_TO_PROPERTY = {
         "TABLE_FORMAT": TableFormatProperty,
+        "PARTITIONED_BY": PartitionedByProperty,
```
tobymao, line 1136: sort this alphabetically

sqlglot/tokens.py
```diff
@@ -130,6 +130,7 @@ class TokenType(AutoName):
     COMMAND = auto()
     COMMENT = auto()
     COMMIT = auto()
+    COMPOUND = auto()
```
tobymao, line 133: add this to id_var

PR #736, opened 2022-11-21
Support Bigquery Unary Plus Op

ShayYaari:


sqlglot/generator.py
```diff
@@ -1119,6 +1119,9 @@ def paren_sql(self, expression):
     def neg_sql(self, expression):
         return f"-{self.sql(expression, 'this')}"
 
+    def pos_sql(self, expression):
+        return f"+{self.sql(expression, 'this')}"
```
GeorgeSittas, line 1123: I don't think `+` is needed here.

GeorgeSittas (approved):

sqlglot/generator.py
ShayYaari: I can remove it if needed.
Just thought about implementing the same behavior as `neg_sql`

GeorgeSittas: Maybe it's fine. Let's wait for Toby's input on this.

PR #739, opened 2022-11-21
Fix TSQL CONCAT single argument transpilation issue

GeorgeSittas:
Fixes https://github.com/tobymao/sqlglot/discussions/738

Maybe we could also consider moving this change over to the base generator?


tobymao:
@GeorgeSittas move it to base

PR #740, opened 2022-11-21
Fix unary negation generator bug

GeorgeSittas:


sqlglot/generator.py
```diff
@@ -1117,7 +1117,10 @@ def paren_sql(self, expression):
         return self.prepend_ctes(expression, sql)
 
     def neg_sql(self, expression):
-        return f"-{self.sql(expression, 'this')}"
+        # This makes sure we don't convert "- - 5" to "--5", which is a comment
+        this_sql = self.sql(expression, "this")
+        sep = " " if seq_get(this_sql, 0) == "-" else ""
```
tobymao, line 1122: why do you need seq_get? when would this_sql ever be empty?

GeorgeSittas: I was thinking defensively here, but yeah I don't think it will ever be empty. Fixing it in a bit.

PR #741, opened 2022-11-21
Optimizer: simplify unary positive sign

GeorgeSittas:


sqlglot/optimizer/simplify.py
```diff
@@ -239,6 +239,10 @@ def simplify_literals(expression):
             if value[0] == "-":
                 return exp.Literal.number(value[1:])
             return exp.Literal.number(f"-{value}")
+    elif isinstance(expression, exp.Pos):
+        this = expression.this
+        if this.is_number:
```
tobymao, line 244: what about the case 

+ - 1 ?

tobymao: ```+-1```

GeorgeSittas: Added some more tests for these.

barakalon (approved):

tests/fixtures/optimizer/simplify.sql
```diff
@@ -274,6 +274,15 @@ TRUE;
 -(-1);
 1;
 
+- -+1;
+1;
+
++-1;
```
tobymao, line 280: is it correct? it doesn’t negate it? is + a noop?

sqlglot/generator.py
```diff
@@ -1128,7 +1128,7 @@ def neg_sql(self, expression):
         return f"-{sep}{this_sql}"
 
     def pos_sql(self, expression):
-        return f"+{self.sql(expression, 'this')}"
+        return f"{self.sql(expression, 'this')}"
```
tobymao, line 1131: you don't need the f

tobymao (approved):

PR #743, opened 2022-11-21
revert _parse_cache changes

barakalon:
I'm pretty sure this is meant to handle Spark's [CACHE TABLE](https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-aux-cache-cache-table.html) statement.

The "=" is optional. This is causing some breakages in our CI, so I'm simply reverting the refactor for now.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1027,7 +1027,12 @@ class Parser(metaclass=_Parser):
         options = []
 
         if self._match(TokenType.OPTIONS):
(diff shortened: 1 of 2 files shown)


GeorgeSittas:
Oops, my bad for not investigating further. Sorry for the inconvenience!

barakalon:
Oh, don't sweat it. Hard to refactor when there aren't proper tests 👍

PR #746, opened 2022-11-22
Add structured ParseError properties

andrewdoss-bit:
Proposing attaching structured properties to `ParseError` for convenient programmatic access and formatting for arbitrary frontends.  

First proposed `sqlglot` PR, so please let me know of any style issues or conventions I may I have overlooked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,28 @@ sqlglot.errors.ParseError: Expecting ). Line 1, Col: 13.
               ~~~~
 ```
 
+Structured syntax error details are also accessible for progammatic use:
+
+```python
+import sqlglot
+try:
+  sqlglot.transpile("SELECT foo( FROM bar")
+except sqlglot.errors.ParseError as e:
+  print(e.error_props)
+```
+
+Output:
+```python
+[{
+  'description': 'Expecting )',
+  'line': 1,
+  'col': 13,
+  'start_context': 'SELECT foo( ',
+  'highlight': 'FROM',
+  'end_context': ' bar'
+}]
+```
+
 ### Unsupported Errors
 
 Presto `APPROX_DISTINCT` supports the accuracy argument which is not supported in Hive:
--- a/README.md
+++ b/README.md
@@ -190,14 +190,14 @@ sqlglot.errors.ParseError: Expecting ). Line 1, Col: 13.
               ~~~~
 ```
 
-Structured syntax error details are also accessible for progammatic use:
+Structured syntax errors are accessible for programmatic use:
 
 ```python
 import sqlglot
 try:
-  sqlglot.transpile("SELECT foo( FROM bar")
+    sqlglot.transpile("SELECT foo( FROM bar")
 except sqlglot.errors.ParseError as e:
-  print(e.error_props)
+    print(e.errors)
 ```
 
 Output:
--- a/sqlglot/errors.py
+++ b/sqlglot/errors.py
@@ -25,26 +25,35 @@ class ParseError(SqlglotError):
     def __init__(
         self,
         message: str,
+        errors: t.Optional[t.List[t.Dict[str, t.Any]]] = None,
+    ):
+        super().__init__(message)
+        self.errors = errors or []
+
+    @classmethod
+    def new(
+        cls,
+        message: str,
         description: t.Optional[str] = None,
         line: t.Optional[int] = None,
         col: t.Optional[int] = None,
         start_context: t.Optional[str] = None,
         highlight: t.Optional[str] = None,
         end_context: t.Optional[str] = None,
-        error_props: t.Optional[t.List[t.Dict[str, t.Any]]] = None,
-    ):
-        super().__init__(message)
-        self.error_props = error_props or []
-        props = {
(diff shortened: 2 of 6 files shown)


sqlglot/errors.py
```diff
@@ -47,3 +69,7 @@ def concat_errors(errors: t.Sequence[t.Any], maximum: int) -> str:
     if remaining > 0:
         msg.append(f"... and {remaining} more")
     return "\n\n".join(msg)
+
+
+def concat_error_props(errors: t.Sequence[ParseError], maximum: int) -> t.List[t.Dict[str, t.Any]]:
+    return [props for e in errors[:maximum] for props in e.error_props]
```
barakalon, line 75: We truncate the string so stacktraces don't explode.

But I don't think we have to truncate the list here, as its intended for programmatic use.

sqlglot/errors.py
```diff
@@ -22,7 +22,29 @@ class UnsupportedError(SqlglotError):
...
+    def __init__(
+        self,
+        message: str,
+        description: t.Optional[str] = None,
+        line: t.Optional[int] = None,
+        col: t.Optional[int] = None,
+        start_context: t.Optional[str] = None,
+        highlight: t.Optional[str] = None,
+        end_context: t.Optional[str] = None,
+        error_props: t.Optional[t.List[t.Dict[str, t.Any]]] = None,
```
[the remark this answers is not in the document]
andrewdoss-bit: I also like `errors` better 👍.

sqlglot/errors.py
andrewdoss-bit: Thanks for explaining, I was wondering about that. Fixed.

sqlglot/parser.py
```diff
@@ -625,7 +625,7 @@ def parse_into(self, expression_types, raw_tokens, sql=None):
                 return self._parse(parser, raw_tokens, sql)
             except ParseError as e:
                 error = e
-        raise ParseError(f"Failed to parse into {expression_types}") from error
+        raise ParseError.new(f"Failed to parse into {expression_types}") from error
```
barakalon, line 628: Ah, interesting.

This instance of `ParseError` will always have an empty `errors` list. It's really an aggregation of multiple sub errors.

I guess it won't immediately effect whatever you're doing, because you're probably not using `parse_into`.

But perhaps we need a different error type? Or maybe there's some clever way to concatenate all the sub errors together?

e.g.
```python
errors = []
for ...:
    try:
        ...
    except ParseError as e:
        errors.append(e)
raise ParseError(
    f"Failed to parse into {expression_types}", 
    errors=merge_errors(errors)
) from errors[-1]
```

That doesn't distinguish which errors are from which expression_types... but at least it keeps a more consistent ParseError interface.

sqlglot/errors.py
```diff
@@ -47,3 +78,7 @@ def concat_errors(errors: t.Sequence[t.Any], maximum: int) -> str:
     if remaining > 0:
         msg.append(f"... and {remaining} more")
     return "\n\n".join(msg)
+
+
+def concat_struct_errors(errors: t.Sequence[ParseError]) -> t.List[t.Dict[str, t.Any]]:
```
barakalon, line 83: `struct` is an overloaded word.

How about we call this `merge_errors` and the above `concat_messages`?

README.md
```diff
@@ -190,6 +190,28 @@ sqlglot.errors.ParseError: Expecting ). Line 1, Col: 13.
...
 ```
 
+Structured syntax error details are also accessible for progammatic use:
+
+```python
+import sqlglot
+try:
+  sqlglot.transpile("SELECT foo( FROM bar")
+except sqlglot.errors.ParseError as e:
+  print(e.error_props)
```
barakalon, line 200: error_props -> errors

barakalon (commented):
Yep, this looks right to me.

A couple more small comments...

Thanks for being flexible 😄

README.md
andrewdoss-bit: Fixed in next commit.

sqlglot/errors.py
andrewdoss-bit: Thanks, agree those are better names. Fixed in next commit.

sqlglot/parser.py
andrewdoss-bit: ah, yes, good point. Correct, I am not currently using `parse_into`, so I didn't think about capturing contextual errors here. I agree that being consistent in providing structures errors is better for the interface. 

I think your proposal in this comment is a good improvement, although I agree that there may be a way to simplify the concatenation further.

I've pushed and tested this approach, with the addition of an optional `into_expression` key per your comment about not distinguishing `expression_types`.

I think it could also be reasonable to make the `into_expression` key conditional, rather than often `None` for other types of `ParseErrors` but have no strong opinion either way.

andrewdoss-bit:
No problem — appreciate the consideration of my request/proposal! I also appreciate this touches an interface, so I expect to need feedback/iteration.

barakalon (approved):
Sweet! LGTM

Issue #712, opened 2022-11-16
sqlglot seems to remove adjacent comments


SudarshanVS:
The following line,

```
Select * from Table /*comment 1*/
/*comment 2*/
```

becomes,
```
Select * from Table /*comment 2*/
```

after we run the parser,

```
expression_tree = parse_one(query)
```

This is an unexpected behaviour.

GeorgeSittas:
Is that correct? I tried to reproduce this and got:

```python
>>> sql = """
... Select * from Table /*comment 1*/
... /*comment 2*/
... """
>>> expr = sqlglot.parse_one(sql)
>>> expr
(SELECT expressions:
  (STAR ), from:
  (FROM expressions:
    (TABLE this:
      (IDENTIFIER this: Table, quoted: False), comment: comment 1)))
>>> expr.sql()
'SELECT * FROM Table /* comment 1 */'
>>> print(expr.sql(pretty=True))
SELECT
  *
FROM Table /* comment 1 */
```

SQLGlot tries to preserve comments in a best-effort basis, because It's not a trivial problem to do a completely lossless transpilation. Basically, comments need to somehow be attached to tokens and then to AST nodes according to a set of heuristics, so in this case the second comment gets discarded during the tokenizing phase because there's no nearby token to attach it to (that is, without complicating the logic further, like attaching several comments to a token etc).

SudarshanVS:
Hi George,

Thank you for your explanation. :)

I am indeed getting the behaviour as mentioned in my original post.

For the below code
```
q = "Select * from Table /*comment 1*/ /*comment 2*/"
print(sqlglot.parse_one(q).sql(pretty=True))
```

I get the following result,
```
SELECT
  *
FROM Table /* comment 2 */
```

I am using sqlglot `v10.0.0`

It is very strange that you are getting a completely different but similar result!!

GeorgeSittas:
Hello SudarshanVS,

The logic is different if you place the comment on the same line as the previous token, as opposed to placing it on a new line. That's why we get different outputs. However, you have a point, since the tokenizer is expected to not overwrite the previous token's comment if any.

mpf82:
Just my 2 cents: what about token.comment as a list, so you could attach multiple comments in order? The default could still be ``None``, so you don't create a bunch of unused empty lists by default.

GeorgeSittas:
@mpf82 I'll check if / how this can be implemented a little later, thanks for the suggestion!

[...]

PR #750, opened 2022-11-22
Refactor comments so they're stored in a list

GeorgeSittas:
The motivation behind this PR can be found in [this](https://github.com/tobymao/sqlglot/issues/712) conversation. It's still a WIP, since I haven't tweaked the generator to allow for multiple comment printing, but it'll come soon.


sqlglot/expressions.py
```diff
@@ -43,14 +43,14 @@ class Expression(metaclass=_Expression):
...
-    __slots__ = ("args", "parent", "arg_key", "type", "comment")
+    __slots__ = ("args", "parent", "arg_key", "type", "comments")
 
     def __init__(self, **args):
         self.args = args
         self.parent = None
         self.arg_key = None
         self.type = None
-        self.comment = None
+        self.comments = []
```
tobymao, line 53: i hope this doesn't impact performance

sqlglot/expressions.py
```diff
@@ -88,18 +88,18 @@ def text(self, key):
...
         Args:
             key: the key of the target child node (e.g. "this", "expression", etc).
 
         Returns:
-            The comment attached to the child node, or the empty string, if it doesn't exist.
+            The comments attached to the child node, or the empty list, if it doesn't exist.
         """
         field = self.args.get(key)
-        return field.comment if isinstance(field, Expression) else ""
+        return field.comments if isinstance(field, Expression) else []
```
tobymao, line 102: well since line 53 is always [], then you don't need it otherwise make line 53 self.comments default to None which may be better performance wise

sqlglot/expressions.py
GeorgeSittas: I'll revisit this and try to get rid of the empty list in favor of None. For now I just wanted to change it so that it works like before with the new representation.

sqlglot/expressions.py
GeorgeSittas: I think the point of this method was to avoid doing `None.comments`, which happened when we allow partial expressions. Should have probably documented this better.

tobymao: ah i misread it nvm

GeorgeSittas: `find_comments` is not actually used anywhere, so I think we can get rid of it for now.

GeorgeSittas:
@tobymao I experimented with the comment formatting a bit and removed the `--`-style comments. The simplifications / consistency we get out of this seems nice, what do you say? I can also revert back to the previous logic and try to think about how multiple comments will be formatted.

sqlglot/generator.py
```diff
@@ -226,25 +226,24 @@ def sep(self, sep=" "):
...
-            return f"/*{comment}*/{self.sep()}{sql}"
+    def maybe_comment(self, sql, expression):
+        comments = expression.comments if self._comments else None
 
-        if not self.pretty:
-            return f"{sql} /*{comment}*/"
+        if not comments:
+            return sql
 
-        if not NEWLINE_RE.search(comment):
```
tobymao, line 244: did you delete the regex?

GeorgeSittas: done

PR #757, opened 2022-11-23
Distinct on fixes

GeorgeSittas:
This PR aims to fix how we handle aliases after DISTINCT ON clauses.

I tried to fix the `eliminate_distinct_on` transformation since the parser changes caused it to break, but it's still a bit clunky probably.

Reference for the DISTINCT ON syntax: https://www.postgresql.org/docs/current/sql-select.html


barakalon:
What does `SELECT DISTINCT ON (x) AS x` mean? I don't see it mentioned in the docs.

tobymao:
good point… a user submitted that and i thought it was valid syntax but i don’t see it in the docs. should try in postgres

tobymao:
i can confirm there's no such thing as on (x) as x... phew

GeorgeSittas:
Lol, I even had the grammar rule in front of me but I misread it and thought we could apply an alias to the distinct on.. thanks for catching this!

PR #758, opened 2022-11-24
improve comment parsing

mpf82:
Now comments after "FROM" or "=" will no longer be discarded, though the order is currently messed up.

```sql
/*a*/SELECT/*b*/*/*c*/FROM/*d*/tbl/*e*/WHERE/*f*/1/*g*/=/*h*/2/*i*/--j
```

would previously translate to

``/* a */ /* b */ SELECT * /* c */ FROM tbl /* e */ WHERE 1 /* g */ = 2 /* i */ /* j */  ``

missing comments ``d, f, and h``

with this PR the above SQL parses and translates to

```
(SELECT expressions:
  (STAR comments: ['c']), from:
  (FROM expressions:
    (TABLE this:
      (IDENTIFIER this: tbl, quoted: False), comments: ['e']), comments: ['d']), where:
  (WHERE this:
    (EQ this:
      (LITERAL this: 1, is_string: False, comments: ['g']), expression:
      (LITERAL this: 2, is_string: False, comments: ['i', 'j']), comments: ['h'])), comments: ['a', 'b'])
/* a */ /* b */ SELECT * /* c */ FROM tbl /* e */ /* d */ WHERE 1 /* g */ = 2 /* i */ /* j */ /* h */    
```

Now only missing comment ``f``, but notice that comments ``d and h`` are added after the expression, not after the token.

That's because currenlty, the generator creates the expression and then appends (or prepends in case of SELECT) the comments in ``self.maybe_comment``.

Maybe someone has a better idea to do this, that's why I'm creating this PR as a draft.

cc: @GeorgeSittas

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -103,7 +103,7 @@ class Generator:
         exp.TableFormatProperty,
     }
 
-    WITH_SEPARATED_COMMENTS = (exp.Select,)
+    WITH_SEPARATED_COMMENTS = (exp.Select, exp.From, exp.Where, exp.Binary)
 
     __slots__ = (
         "time_mapping",
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import copy
 import logging
 import typing as t
 
@@ -1227,8 +1226,9 @@ class Parser(metaclass=_Parser):
     def _parse_from(self):
         if not self._match(TokenType.FROM):
             return None
-        this = self.expression(exp.From, comments=self._prev_comments, expressions=self._parse_csv(self._parse_table))
-        return this
+        return self.expression(
+            exp.From, comments=self._prev_comments, expressions=self._parse_csv(self._parse_table)
+        )
 
     def _parse_lateral(self):
         outer_apply = self._match_pair(TokenType.OUTER, TokenType.APPLY)
@@ -1491,7 +1491,9 @@ class Parser(metaclass=_Parser):
     def _parse_where(self, skip_where_token=False):
         if not skip_where_token and not self._match(TokenType.WHERE):
             return None
-        return self.expression(exp.Where, comments=self._prev_comments, this=self._parse_conjunction())
+        return self.expression(
+            exp.Where, comments=self._prev_comments, this=self._parse_conjunction()
+        )
 
     def _parse_group(self, skip_group_by_token=False):
         if not skip_group_by_token and not self._match(TokenType.GROUP_BY):
@@ -2500,7 +2502,10 @@ class Parser(metaclass=_Parser):
 
         while self._match_set(expressions):
             this = self.expression(
-                expressions[self._prev.token_type], this=this, comments=self._prev_comments, expression=parse_method()
+                expressions[self._prev.token_type],
+                this=this,
+                comments=self._prev_comments,
+                expression=parse_method(),
             )
 
         return this
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1223,8 +1223,10 @@ class Parser(metaclass=_Parser):
     def _parse_from(self):
         if not self._match(TokenType.FROM):
             return None
-
-        return self.expression(exp.From, expressions=self._parse_csv(self._parse_table))
+        comments = self._prev_comments
+        this = self.expression(exp.From, expressions=self._parse_csv(self._parse_table))
+        this.comments = comments
+        return this
 
     def _parse_lateral(self):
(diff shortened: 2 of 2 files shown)


GeorgeSittas:
I think the changes are LGTM, can't see another way to extract the comments from "special" keyword tokens such as `FROM` and `WHERE`. As for the generator issue, what if we prepended the comments of `FROM` like we do for select? That way we can have the comments appear closer to the keyword they were originally attached to.

mpf82:
tryin to use ``expression`` to do this, instead of a helper function.

A helper function would probably have the drawback that ``self`` is already advanced and we cannot access the comments for the token.

that's why it's also important to set ``comments=`` before calling any functions that advance ``self``, otherwise this would fail, too. (well, not fail, but then ``self`` would point to the wrong token).

sqlglot/generator.py
```diff
@@ -103,7 +103,7 @@ class Generator:
         exp.TableFormatProperty,
     }
 
-    WITH_SEPARATED_COMMENTS = (exp.Select,)
+    WITH_SEPARATED_COMMENTS = (exp.Select, exp.From, exp.Where, exp.Binary)
```
tobymao, line 106: why did you add binary here

sqlglot/parser.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
```
tobymao, line 3: what are you using copy for?

sqlglot/parser.py
```diff
@@ -693,10 +694,13 @@ def raise_error(self, message, token=None):
         self.errors.append(error)
 
     def expression(self, exp_class, **kwargs):
+        comments = kwargs.pop("comments", None)
```
tobymao, line 697: just add comment as a first class kwarg default nine

tobymao: none

sqlglot/parser.py
```diff
@@ -1223,8 +1227,8 @@ def _parse_hint(self):
     def _parse_from(self):
         if not self._match(TokenType.FROM):
             return None
-
-        return self.expression(exp.From, expressions=self._parse_csv(self._parse_table))
+        this = self.expression(exp.From, comments=self._prev_comments, expressions=self._parse_csv(self._parse_table))
+        return this
```
tobymao, line 1231: get rid of this and make it a one liner

sqlglot/parser.py
mpf82: that's a leftover from playing around, removed

sqlglot/generator.py
mpf82: given ``1=/*c*/2`` and ``1+/*c*/2``, this will result in 

``/* c */ 1 = 2`` and ``/* c */ 1 + 2``

otherwise the comment will be after the expression

sqlglot/parser.py
mpf82: I cannot get this to work. You meant to changing

``
class Expression(metaclass=_Expression):
    arg_types = {"this": True, "comment": False}
``

correct? 

But that means changing ``arg_types `` for every class that bring it's own ``arg_types``, which is impractical.

Unless I misunderstood what you're suggesting ...

sqlglot/parser.py
mpf82: done

sqlglot/generator.py
tobymao: don’t you want to just keep the formatting?

tobymao: or is the current behavior not 1 + /* x */ 2

sqlglot/parser.py
tobymao: no just to this method add

def expression(self, comment=None, **kwargs

sqlglot/generator.py
mpf82: As I said in my first post, by default the comment would be added after the expression. 

Since ``1 = 2`` is the (equality) expression, the comment would be added afterwards. Of course, keeping the original position of the comment would be preferred, but that's probably to be solved at a later point. As GeorgeSittas sugested, it might be preferrable for some expressions, to have the comment moved to the front, if the alternative is to have it placed at the end.

tobymao: so why is /* x */ 1 + 2 better than 1 + 2 /* x */, should we just leave it the default?

mpf82: Consider this expression:  ``2 * /*c1*/ (3+ /*c2*/ 4)``

Default (comment at the end): ``2 * (3 + 4 /* c2 */) /* c1 */``

Comment at the start: ``/* c1 */ 2 * (/* c2 */ 3 + 4)``

Thus I would argue that the comment at the start is better in this case.

PR #759, opened 2022-11-24
Added ability to parse CREATE TABLE A LIKE B and CREATE TABLE A (LIKE B INCLUDING...)

yuchaofan13:
CREATE TABLE A LIKE B is allowed in MySQL (https://dev.mysql.com/doc/refman/5.7/en/create-table-like.html) whilst CREATE TABLE A (LIKE B) is allowed in PostgreSQL (https://www.postgresql.org/docs/current/sql-createtable.html).

In Postgres you can further specify the type of deep copy you want to make, e.g. whether to include/exclude constraints, compression etc. (EXCLUDE being the default option). The syntax is then like 'CREATE TABLE A (LIKE B INCLUDING CONSTRAINTS INCLUDING COMPRESSION INCLUDING STATISTICS EXCLUDING COMMENTS)'.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1075,8 +1075,10 @@ class SortKeyProperty(Property):
 class DistStyleProperty(Property):
     arg_types = {"this": True}
 
+
 class IncludingProperty(Property):
     arg_types = {"this": True}
+
 
 class LocationProperty(Property):
     arg_types = {"this": True}
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1080,6 +1080,10 @@ class IncludingProperty(Property):
     arg_types = {"this": True}
 
 
+class ExcludingProperty(Property):
+    arg_types = {"this": True}
+
+
 class LocationProperty(Property):
     arg_types = {"this": True}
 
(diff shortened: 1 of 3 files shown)


tobymao:
can you add a test?

yuchaofan13:
> can you add a test?

I know that running parse_one() on the above expressions now gives the desired output. However, I tried adding tests to test_postgres and test_mysql and they failed.

I'm unfamiliar with how validate_identity/validate_all work- how exactly does SQLGlot determine the dialect? Is it to do with the generator/transforms (which I'm also unfamiliar with, and haven't made changes to in this PR yet)?

I've seen examples of adding dialect-specific properties such as TRANSIENT and DISTSTYLE but here I've modified the way 'create table' is parsed and I'm not sure how to go about things.

Some guidance would greatly appreciated!

tobymao:
what test did you add and how did they fail?

yuchaofan13:
> what test did you add and how did they fail?

In test_mysql.py: self.validate_identity("CREATE TABLE A LIKE B")
Which led to AssertionError: 'CREATE TABLE A LIKE B' != 'CREATE TABLE A AS A LIKE B'

And in test_postgres.py: self.validate_identity(
            "CREATE TABLE A (LIKE B INCLUDING STATISTICS INCLUDING CONSTRAINTS INCLUDING DEFAULTS EXCLUDING COMMENTS)"
        )
Which led to AssertionError: 'CREATE TABLE A (LIKE B INCLUDING STATISTICS INCLUDING CO[44 chars]NTS)' != 'CREATE TABLE A AS A LIKE B'

tobymao:
seems like your code doesn't work then

tobymao:
closed in favor of #760

PR #760, opened 2022-11-25
Like

tobymao:
@yuchaofan13

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -453,6 +453,7 @@ class MySQL(Dialect):
             exp.CharacterSetProperty,
             exp.CollateProperty,
             exp.SchemaCommentProperty,
+            exp.LikeProperty,
         }
 
         WITH_PROPERTIES: t.Set[t.Type[exp.Property]] = set()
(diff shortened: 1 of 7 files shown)


yuchaofan13:
Thanks for fixing my code- I thought modifying _parse_create() was the only option and didn't realise TokenType.LIKE could have different parsing functions across RANGE_PARSERS, PROPERTY_PARSERS and CONSTRAINT_PARSERS.

Do you have any idea as to why my original code wasn't working?

tobymao:
yea, there were many problems. most of the code you added wasn’t even being called

yuchaofan13:
> yea, there were many problems. most of the code you added wasn’t even being called

yeah parse_one gave the desired output but I guess things weren't working properly under the hood. Do you have any advice/SOP for making code changes like this? I'm gradually developing a better understanding of the overall codebase but still relying a lot on copying similar code or messily modifying existing code.

tobymao:
you need to really understand what is getting called and what's happening. parsers are in general quite complex because they are very recursive, tracing through the stack and seeing how things work internally will give you a better understanding of what's going on.

yuchaofan13:
> you need to really understand what is getting called and what's happening. parsers are in general quite complex because they are very recursive, tracing through the stack and seeing how things work internally will give you a better understanding of what's going on.

Gotcha, thanks for the advice. Which key modules would you recommend first trying to understand?

tobymao:
the parser and generator

PR #768, opened 2022-11-25
Add support for the INTO clause in SELECT

### LINKED ISSUE #765: SELECT INTO not supported
Parsing
```sql
select * into newevent from event;
```
from https://docs.aws.amazon.com/redshift/latest/dg/r_SELECT_INTO.html#r_SELECT_INTO-examples
fails with
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 10.
```

Edit: Exists also in Postgres: https://www.postgresql.org/docs/9.1/sql-selectinto.html

GeorgeSittas:
Fixes #765


sqlglot/tokens.py
```diff
@@ -274,6 +274,7 @@ class TokenType(AutoName):
     UNBOUNDED = auto()
     UNCACHE = auto()
     UNION = auto()
+    UNLOGGED = auto()
```
tobymao, line 277: as this to id var

sqlglot/parser.py
```diff
@@ -1131,9 +1131,15 @@ def _parse_select(self, nested=False, table=False):
                 limit=limit,
             )
             this.comments = comments
+
+            into = self._parse_into()
```
tobymao, line 1135: how does this affect using into as an alias

GeorgeSittas: Good point, I'll look into it in a sec.

GeorgeSittas:
```python
>>> sqlglot.parse_one("SELECT x AS INTO FROM bar").sql()
'SELECT x AS INTO FROM bar'
```

PR #769, opened 2022-11-27
"decode" support in Snowflake. This function is used to decode text i…

prabhur:
"decode" support in Snowflake. This function is used to decode text in postgres. However, Oracle and Snowflake implement this as a conditional.

Notes: 
1) These changes are specific to snowflake dialect.
2) It was tempting to parse the decode function into it's consituents - expression, search-replace pairs and an optional default. However, just parsed everything in a list to keep things simple.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -81,11 +81,11 @@ def _parse_date_part(self):
     name = this.name.upper()
     if name.startswith("EPOCH"):
         if name.startswith("EPOCH_MILLISECOND"):
-            scale = 10**3
+            scale = 10 ** 3
         elif name.startswith("EPOCH_MICROSECOND"):
-            scale = 10**6
+            scale = 10 ** 6
         elif name.startswith("EPOCH_NANOSECOND"):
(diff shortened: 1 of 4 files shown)


tobymao:
i don’t really understand this change, it seems unnecessary, you basically reimplemented var args

tobymao:
closed in favor of https://github.com/tobymao/sqlglot/commit/283516e4ec5ab9bf9fbbf7a15228ecc5b032fc34

prabhur:
Hmm. Didn’t notice the Matches expression. Additionally, I wouldn’t have thought about using rename_func as well. This example explains it well.

tobymao:
there wasn’t matches, i created it

prabhur:
Thanks and Sorry. I see it now. Wasn't paying attention to it when browsing via my mobile after waking up.

A follow-up question for my knowledge. Looking at https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/snowflake.py#L138
and `FUNCTION_PARSERS` is it accurate that a parsing routine is required when we intend to split the expression into useful pieces that need some kind of processing?

PR #770, opened 2022-11-27
Added ability to parse END (COMMIT equivalent in Postgres) and the CHAIN option

yuchaofan13:
https://www.postgresql.org/docs/current/sql-end.html

Added a test that works (I now understand how testing works with the generator!) but maybe you want to modify the way it's parsed/shows up in the AST.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2114,7 +2114,11 @@ class Transaction(Command):
 
 
 class Commit(Command):
-    arg_types = {}  # type: ignore
+    arg_types = {"chain": False}
+
+
+class End(Command):
+    arg_types = {"chain": False}
 
 
 class Rollback(Command):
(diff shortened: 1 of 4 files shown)


tobymao:
the approach is not quite correct, instead you should map end to commit in postgres at the tokenizer phase

yuchaofan13:
> the approach is not quite correct, instead you should map end to commit in postgres at the tokenizer phase
> 
> then you can remove most of the code you’ve added

Ah, that makes a lot more sense! I knew there must be something like this but wasn't sure exactly where it was. 

I added `"END": TokenType.COMMIT` to `KEYWORDS` in postgres.py- however, when I ran test_postgres.py it broke the functionality for parsing "CASE WHEN....END". Am I adding to the wrong place?

tobymao:
ah i guess we can't do it that way then

GeorgeSittas:
Opened a separate PR and tried to clean things up a bit. Take a look: https://github.com/tobymao/sqlglot/pull/773

PR #774, opened 2022-11-28
Add support for STRING_AGG <-> GROUP_CONCAT transpilation

GeorgeSittas:
Fixes https://github.com/tobymao/sqlglot/discussions/754

This seems to do the trick, although I'll take a look an try to simplify/refactor things in a bit.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -17,6 +17,7 @@ FULL_FORMAT_TIME_MAPPING = {
     "mm": "%B",
     "m": "%B",
 }
+
 DATE_DELTA_INTERVAL = {
     "year": "year",
     "yyyy": "year",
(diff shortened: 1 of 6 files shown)


GeorgeSittas:
@tobymao
Ok, I'd say this is good to go. Not sure if there's any part that can be greatly simplified. Seems like STRING_AGG is a T-SQL function so I think it makes sense to not move these implementations over to the base parser / generator. Let me know if I've missed anything.

sqlglot/dialects/tsql.py
```diff
@@ -87,6 +89,29 @@ def generate_format_sql(self, e):
...
+
+    order = ""
+    if isinstance(e.this, exp.Order):
+        if e.this.this:
+            this = e.this.this
+            e.this.this.pop()
+        order = f" WITHIN GROUP ({self.sql(e.this)[1:]})"
+
+    separator = e.args.get("separator")
+    separator = separator.name if separator else exp.Literal.string(",")
```
tobymao, line 111: why are using Literal.string? why not just ","

GeorgeSittas: Because the formatting wasn't quite right. If we use a raw string it outputs `STRING_AGG(x, ,)` for example.

GeorgeSittas:
Sources for sqlite/postgres additions:

- https://www.postgresql.org/docs/9.0/functions-aggregate.html
- https://www.sqlite.org/lang_aggfunc.html#group_concat

tobymao (approved):

PR #778, opened 2022-11-29
Fix postgres string_agg parsing & array generation

GeorgeSittas:
The `DISTINCT` keyword is supported for postgres after all (it wasn't clear from the docs, but works just fine in postgres15).

As an extra, I fixed postgres array generation because the `ARRAY(...)` form isn't supported, I think.
Source on this: https://www.postgresql.org/docs/current/arrays.html

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -76,16 +76,16 @@ def _trim_sql(self, expression):
 
 def _string_agg_sql(self, expression):
     expression = expression.copy()
+    separator = expression.args.get("separator") or exp.Literal.string(",")
 
     order = ""
     this = expression.this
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -78,7 +78,7 @@ def _string_agg_sql(self, expression):
     expression = expression.copy()
 
     order = ""
-    this = expression.find(exp.Distinct) or expression.this
+    this = expression.this
     if isinstance(expression.this, exp.Order):
```
tobymao, line 82: replace expression.this with this?

GeorgeSittas: What about now?

tobymao (approved):

PR #780, opened 2022-11-29
Improve Snowflake array and map support

eakmanrq:
Snowflake does not support types associated with their MAP (OBJECT) and ARRAY types. Also Snowflake is limited in options for constructing maps and therefore you must use a function (`OBJECT_CONSTRUCT`) to create the MAP.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -152,6 +152,7 @@ class Snowflake(Dialect):
             "ARRAY_CONSTRUCT": exp.Array.from_arg_list,
             "RLIKE": exp.RegexpLike.from_arg_list,
             "DECODE": exp.Matches.from_arg_list,
+            "OBJECT_CONSTRUCT": parser.parse_var_map,
         }
 
         FUNCTION_PARSERS = {
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -6,6 +6,7 @@ from sqlglot.dialects.dialect import (
     format_time_lambda,
     inline_array_sql,
     rename_func,
+    var_map_sql,
 )
 from sqlglot.expressions import Literal
 from sqlglot.helper import seq_get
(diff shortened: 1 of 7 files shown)


tests/dialects/test_hive.py
```diff
@@ -459,6 +459,7 @@ def test_hive(self):
                 "hive": "MAP(a, b, c, d)",
                 "presto": "MAP(ARRAY[a, c], ARRAY[b, d])",
                 "spark": "MAP(a, b, c, d)",
+                "snowflake": "OBJECT_CONSTRUCT(a, b, c, d)",
```
eakmanrq, line 462: @tobymao any advice on how to be implement this? Basically I need the parser to understand that `OBJECT_CONSTRUCT` function creates a MAP value.

sqlglot/dialects/snowflake.py
```diff
@@ -198,7 +207,10 @@ class Generator(generator.Generator):
             **generator.Generator.TRANSFORMS,
             exp.Array: inline_array_sql,
             exp.ArrayConcat: rename_func("ARRAY_CAT"),
+            exp.DataType: _datatype_sql,
             exp.If: rename_func("IFF"),
+            exp.Map: lambda self, e: var_map_sql(self, e, "OBJECT_CONSTRUCT"),
```
tobymao, line 212: i don't think this is correct, map is not the same as a varmap

sqlglot/dialects/snowflake.py
```diff
@@ -100,6 +101,14 @@ def _parse_date_part(self):
     return self.expression(exp.Extract, this=this, expression=expression)
 
 
+def _datatype_sql(self, expression):
+    if expression.this == exp.DataType.Type.ARRAY:
+        return "ARRAY"
```
tobymao, line 106: so these are needed because snowflake doesn't have these as nested data structures but using variants?

tests/dialects/test_hive.py
GeorgeSittas: What about:
```python
FUNCTIONS = {
    **parser.Parser.FUNCTIONS,
    "ARRAYAGG": exp.ArrayAgg.from_arg_list,
    "IFF": exp.If.from_arg_list,
    "TO_TIMESTAMP": _snowflake_to_timestamp,
    "ARRAY_CONSTRUCT": exp.Array.from_arg_list,
    "RLIKE": exp.RegexpLike.from_arg_list,
    "DECODE": exp.Matches.from_arg_list,
    "OBJECT_CONSTRUCT": parser.parse_var_map,
}
```

sqlglot/dialects/snowflake.py
eakmanrq: Yep that is right

tests/dialects/test_hive.py
eakmanrq: Awesome this is perfect! Thanks!

PR #788, opened 2022-12-03
added support for redshift ENCODE, fix parsing for DISTSTYLE ALL

yuchaofan13:
https://docs.aws.amazon.com/redshift/latest/dg/c_Compression_encodings.html

Originally I thought I could just map ENCODE to TokenType.CONSTRAINT in redshift.py but that wouldn't be correct and wouldn't generate properly.

I also tried creating tokentype.ENCODE but that messes up existing parsing for ENCODE for other dialects, which is very different.

So instead I opted for creating tokentype.COMPRESSION and mapping ENCODE to COMPRESSION in redshift.py. 

https://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html
I also made some very minor changes to how distkey, sortkey and diststyle are parsed. In particular, one of the default options for diststyle is 'all' which _parse_var can't handle, so I changed it to _parse_id_var. I decided to do the same for distkey and sortkey because something fairly innocuous like 'distkey(location)' leads to an error.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -745,7 +745,7 @@ class DefaultColumnConstraint(ColumnConstraintKind):
 
 
 class EncodeColumnConstraint(ColumnConstraintKind):
-    arg_types = {"this": True}
+    pass
 
 
 class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -740,12 +740,12 @@ class CommentColumnConstraint(ColumnConstraintKind):
     pass
 
 
-class CompressionColumnConstraint(ColumnConstraintKind):
-    arg_types = {"this": True}
-
-
 class DefaultColumnConstraint(ColumnConstraintKind):
     pass
+
+
+class EncodeColumnConstraint(ColumnConstraintKind):
+    arg_types = {"this": True}
 
 
 class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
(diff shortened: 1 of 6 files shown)


GeorgeSittas:
I think that `ENCODE` would be referred to as a "property" in SQLGlot (instead of "constraint"), given that it comes after the schema. So, I'd suggest:

- Adding an `ENCODE` token and mapping `"ENCODE"` to it for the Redshift dialect.
- Renaming "compression" to "encode", e.g. `EncodeProperty`.

I didn't really understand why having an `ENCODE` token messes up the parsing logic for the other dialects.

yuchaofan13:
@GeorgeSittas My bad, I should have clarified that ENCODE is defined on a columnar basis with Redshift so it's part of the schema like this:
`create table product(
product_id int encode raw,
product_name char(20) encode bytedict);`

Your suggestion for mapping ENCODE to Redshift was completely on point though. When I tried it originally, I was adding both ENCODE = auto() and "ENCODE": TokenType.ENCODE to tokens.py. Unfamiliar with the enum library so I didn't know that the former creates the TokenType Class, whilst the latter maps it in the parser- hence why the parser was failing to parse something like "encode(x, 'utf-8')"!

Have made these changes!

GeorgeSittas (approved):

sqlglot/expressions.py
```diff
@@ -744,6 +744,10 @@ class DefaultColumnConstraint(ColumnConstraintKind):
     pass
 
 
+class EncodeColumnConstraint(ColumnConstraintKind):
+    arg_types = {"this": True}
```
GeorgeSittas, line 748:
```suggestion
class EncodeColumnConstraint(ColumnConstraintKind):
    pass
```

GeorgeSittas:
Great, thanks for the PR!

PR #790, opened 2022-12-03
Add support for redshift DECODE

### LINKED ISSUE #789: Redshift DECODE
Current parsing for Redshift DECODE leads to a parsing error: https://docs.aws.amazon.com/redshift/latest/dg/r_DECODE_expression.html
e.g. SELECT DECODE(COL1, 'replace_this', 'with_this', 'replace_that', 'with_that') 

Redshift DECODE is essentially REPLACE but allows unlimited arguments. 

However, mapping DECODE to REPLACE in the tokenizer didn't work and I realised this was the wrong approach. I then tried adding DECODE to FUNC_TOKENS in redshift.py but that still didn't work. Not sure how to override DECODE in expressions.

GeorgeSittas:
Fixes #789

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from sqlglot import exp, transforms
+from sqlglot.dialects.dialect import rename_func
 from sqlglot.dialects.postgres import Postgres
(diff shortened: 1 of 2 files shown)


yuchaofan13:
Ahh, so that's how you do it. Nice

PR #801, opened 2022-12-05
Enable table aliasing for UPDATE

yuchaofan13:
https://www.postgresql.org/docs/current/sql-update.html

This seems like a Postgres exclusive feature that allows the syntax 'UPDATE TABLE1 T1 SET T1.SOME_COL = SOME_VALUE'.
Schema = True was preventing the alias from ever getting parsed. I don't think the schema is ever an argument for UPDATE in any dialects (correct me if I'm wrong) 

Just one other thing: what does the line `table = (not schema and self._parse_function()) or self._parse_id_var(False)` mean? In particular, what actually does id_var represent?

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1188,12 +1188,12 @@ class Parser(metaclass=_Parser):
 
     def _parse_table_alias(self, is_update=False):
         any_token = self._match(TokenType.ALIAS)
-        if is_update:
-            alias = self._parse_id_var(
-                any_token=any_token, tokens=self.TABLE_ALIAS_TOKENS - {TokenType.SET}
-            )
-        else:
-            alias = self._parse_id_var(any_token=any_token, tokens=self.TABLE_ALIAS_TOKENS)
+        alias = self._parse_id_var(
+            any_token=any_token,
+            tokens=self.TABLE_ALIAS_TOKENS - {TokenType.SET}
+            if is_update
+            else self.TABLE_ALIAS_TOKENS,
+        )
         columns = None
 
         if self._match(TokenType.L_PAREN):
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1035,7 +1035,7 @@ class Parser(metaclass=_Parser):
         return self.expression(
             exp.Update,
             **{
-                "this": self._parse_table(),
+                "this": self._parse_table(is_update=True),
                 "expressions": self._match(TokenType.SET) and self._parse_csv(self._parse_equality),
                 "from": self._parse_from(),
                 "where": self._parse_where(),
@@ -1186,9 +1186,9 @@ class Parser(metaclass=_Parser):
             alias=alias,
         )
 
-    def _parse_table_alias(self):
+    def _parse_table_alias(self, is_update=False):
         any_token = self._match(TokenType.ALIAS)
-        if self._tokens[self._index - 2].token_type == TokenType.UPDATE:
+        if is_update:
             alias = self._parse_id_var(
                 any_token=any_token, tokens=self.TABLE_ALIAS_TOKENS - {TokenType.SET}
             )
@@ -1345,7 +1345,7 @@ class Parser(metaclass=_Parser):
             columns=self._parse_expression(),
         )
 
-    def _parse_table(self, schema=False):
+    def _parse_table(self, schema=False, is_update=False):
         lateral = self._parse_lateral()
 
         if lateral:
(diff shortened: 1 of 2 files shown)


yuchaofan13:
okay that broke more things than expected

sqlglot/parser.py
```diff
@@ -203,7 +203,6 @@ class Parser(metaclass=_Parser):
         TokenType.SCHEMA_COMMENT,
         TokenType.SEED,
         TokenType.SEMI,
-        TokenType.SET,
```
tobymao, line 206: you can't remove this,

this remove the ability to do select 1 set

GeorgeSittas:
> what does the line table = (not schema and self._parse_function()) or self._parse_id_var(False) mean? 

In some places where we expect to parse a table, the table might be a function call (e.g. `READ_CSV`). This is conflicting with the syntax for schema, where you also need to parse something of the form `<table identifier> ( ... )`. So in the left operand of the `or` it says "parse me a function if we're not at the state of parsing a schema".

> In particular, what actually does id_var represent?

`ID_VAR_TOKENS` are tokens that can be used as identifiers, so `parse_id_var`, so for example you can encounter these where you expect an alias or a table name etc.

I might've missed something though. Toby can probably explain these better.

yuchaofan13:
@GeorgeSittas Thanks for the explanation, makes a lot more sense now!

Let me try modifying _parse_table_alias

yuchaofan13:
My current code change is failing because I forgot to add 'if self._index >1', but the approach still isn't general enough if the table is nested (e.g. db.table_name) such that the UPDATE statement is not two tokens away. 

I think adding an argument 'is_update' to _parse_table then passing this through to _parse_table_alias would work, but is there a more elegant way?

tests/dialects/test_postgres.py
```diff
@@ -94,6 +94,7 @@ def test_postgres(self):
         self.validate_identity("COMMENT ON TABLE mytable IS 'this'")
         self.validate_identity("SELECT e'\\xDEADBEEF'")
         self.validate_identity("SELECT CAST(e'\\176' AS BYTEA)")
+        self.validate_identity("UPDATE MYTABLE AS T1 SET T1.COL = 13")
```
GeorgeSittas, line 97: This doesn't test the alias syntax you're trying to fix (`update table t1 ....`). You should instead use `validate_all`.

GeorgeSittas:
> is there a more elegant way?

I think you can just do:

1) Remove `schema=True` inside update's parser (done).
2) Add `TokenType.SET` in the base parser's `TABLE_ALIAS_TOKENS` set.

@tobymao that would mean we can't have `SET` as a table alias, but I think it's ok to make this compromise. Unless, of course, there's a simple way to allow this syntax.

tobymao:
yes, it's an alias

yuchaofan13:
Yeah, I'll add the is_update argument and change the test later today.

sqlglot/parser.py
```diff
@@ -1186,9 +1186,14 @@ def _parse_cte(self):
             alias=alias,
         )
 
-    def _parse_table_alias(self):
+    def _parse_table_alias(self, is_update=False):
```
tobymao, line 1189: instead, you should pass in the table alias tokens

sqlglot/parser.py
```diff
@@ -1035,7 +1035,7 @@ def _parse_update(self):
         return self.expression(
             exp.Update,
             **{
-                "this": self._parse_table(schema=True),
+                "this": self._parse_table(is_update=True),
```
tobymao, line 1038: make a constant that is TABLE_ALIAS_TOKENS - {SET} and pass it in in here

sqlglot/parser.py
yuchaofan13: that would be more general, makes sense 👍

tobymao:
@yuchaofan13 feel free to join my slack https://join.slack.com/t/tobiko-data/shared_invite/zt-1l4uaam1u-hJgiSrNk0_~6fqZGewTHGA

sqlglot/parser.py
```diff
@@ -1035,7 +1035,7 @@ def _parse_update(self):
         return self.expression(
             exp.Update,
             **{
-                "this": self._parse_table(schema=True),
+                "this": self._parse_table(alias_tokens=self.TABLE_ALIAS_TOKENS - {TokenType.SET}),
```
tobymao, line 1038: make this a constant under TABLE_ALIAS_TOKENS where it is defined

yuchaofan13: ohhhh ok, guess it makes sense to have constants in the same place instead of doing it on the fly

GeorgeSittas (approved):

PR #812, opened 2022-12-09
Qualified lateral functions

10bas10:
Added a class variable at the parser class to make a distinction between dialects on parsing table or column aliases. This is due to the fact that at postgresql in case of a lateral join on a function, no table alias is expected.

Also changed the order of query modifiers in order to have columns available which can be used at laterals. This also required a fix in a test case in test_parser.py

Also added support for unicode literal strings at T-SQL along with a type translation for DATETIME to TIMESTAMP in Hive.

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -244,7 +244,7 @@ class Postgres(Dialect):
 
     class Parser(parser.Parser):
         STRICT_CAST = False
-        TREAT_LAT_FUNC_AS_LAT_VIEW = True
+        LATERAL_FUNCTION_AS_VIEW = True
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -244,6 +244,7 @@ class Postgres(Dialect):
 
     class Parser(parser.Parser):
         STRICT_CAST = False
+        TREAT_LAT_FUNC_AS_LAT_VIEW = True
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -224,6 +224,12 @@ class TSQL(Dialect):
     class Tokenizer(tokens.Tokenizer):
         IDENTIFIERS = ['"', ("[", "]")]
 
+        QUOTES = [
+            (prefix + quote, quote) if prefix else quote
+            for quote in ["'", '"']
+            for prefix in ["", "n", "N"]
+        ]
+
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
             "BIT": TokenType.BOOLEAN,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1287,7 +1287,9 @@ class Parser(metaclass=_Parser):
             return None
 
         if not this:
-            this = self._parse_function()
+            this = self._parse_function() or self._parse_id_var(False)
+            while self._match(TokenType.DOT):
+                this = exp.Dot(this=this, expression=self._parse_function() or self._parse_id_var())
 
         table_alias = self._parse_id_var(any_token=False)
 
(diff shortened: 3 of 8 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -224,6 +224,12 @@ class TSQL(Dialect):
     class Tokenizer(tokens.Tokenizer):
         IDENTIFIERS = ['"', ("[", "]")]
 
+        QUOTES = [
+            (prefix + quote, quote) if prefix else quote
+            for quote in ["'", '"']
+            for prefix in ["", "n", "N"]
+        ]
```
GeorgeSittas, line 231: I think we have something similar for MySQL, but we handle it as an "introducer" instead. If it's the same concept, then we should probably change either dialect so that we maintain consistency.

sqlglot/parser.py
```diff
@@ -1287,23 +1288,30 @@ def _parse_lateral(self):
...
+                columns = self._parse_csv(self._parse_id_var)
+                self._match_r_paren()
 
         expression = self.expression(
             exp.Lateral,
             this=this,
             view=view,
             outer=outer,
-            alias=self.expression(exp.TableAlias, this=table_alias, columns=columns),
+            alias=exp.TableAlias(this=table_alias, columns=columns),
```
GeorgeSittas, line 1314: Why was the `self.expression` call removed here? We try to use this as much as possible because it includes some logic for comment attaching etc.

GeorgeSittas (commented):
Just a few comments. I haven't had the chance to investigate on the related semantics, though. For example, are the two versions of the query in `test_parser` equivalent?

sqlglot/parser.py
```diff
@@ -1287,23 +1288,30 @@ def _parse_lateral(self):
             return None
 
         if not this:
-            this = self._parse_function()
-
-        table_alias = self._parse_id_var(any_token=False)
+            this = self._parse_function() or self._parse_id_var(False)
```
tobymao, line 1291: why are you hardcoding False here?

sqlglot/dialects/tsql.py
10bas10: I implemented this the same like in bigquery.

sqlglot/parser.py
```diff
@@ -1287,23 +1288,30 @@ def _parse_lateral(self):
...
-            columns = self._parse_csv(self._parse_id_var)
-            self._match_r_paren()
+        table_alias = None
+        if view or self.TREAT_LAT_FUNC_AS_LAT_VIEW:
+            table_alias = self._parse_id_var(any_token=False)
+            if self._match(TokenType.ALIAS):
+                columns = self._parse_csv(self._parse_id_var)
+        else:
+            self._match(TokenType.ALIAS)
+            table_alias = self._parse_id_var(any_token=False)
```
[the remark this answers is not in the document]
10bas10: Made it more consistent

sqlglot/parser.py
10bas10: Reverted that change

sqlglot/dialects/postgres.py
```diff
@@ -244,6 +244,7 @@ class Tokenizer(tokens.Tokenizer):
 
     class Parser(parser.Parser):
         STRICT_CAST = False
+        TREAT_LAT_FUNC_AS_LAT_VIEW = True
```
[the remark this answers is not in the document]
10bas10: Renamed

sqlglot/parser.py
10bas10: Made it more consistent within its function

10bas10:
> Just a few comments. I haven't had the chance to investigate on the related semantics, though. For example, are the two versions of the query in `test_parser` equivalent?

Not equivalent no. The old situation wont parse, a lateral is using data from a CROSS JOIN that was in the wrong order.

sqlglot/dialects/tsql.py
GeorgeSittas: I'm not sure if this is correct after all, because you can't preserve the `N` to later transpile it correctly.

tobymao: unless you're ok losing the N

PR #815, opened 2022-12-14
Improve values builder to support names and types

eakmanrq:
Prior to this change the values builder would only support values with an alias. Now you can provide types and column names. This allows you to create a values expression that will match a desired table definition which includes the column names and types. 

Edit: Removed types support since it would make the expressions too long. Use a subquery if you want to type.

### DIFF
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -22,7 +22,6 @@ from sqlglot.expressions import (
     or_,
     select,
     subquery,
-    DataType,
 )
 from sqlglot.expressions import table_ as table
 from sqlglot.expressions import to_column, to_table, union
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -22,6 +22,7 @@ from sqlglot.expressions import (
     or_,
     select,
     subquery,
+    DataType,
 )
 from sqlglot.expressions import table_ as table
 from sqlglot.expressions import to_column, to_table, union
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3674,9 +3674,9 @@ def table_(table, db=None, catalog=None, quoted=None, alias=None) -> Table:
 
 
 def values(
-    values: t.Sequence[t.Tuple[str | Expression, ...]],
+    values: t.Iterable[t.Tuple[str | Expression, ...]],
     alias: t.Optional[str] = None,
-    columns_and_types: t.Optional[t.List[str | t.Tuple[str, str | DataType]]] = None,
+    columns_and_types: t.Optional[t.Iterable[str | t.Tuple[str, str | DataType]]] = None,
 ) -> Values:
     """Build VALUES statement.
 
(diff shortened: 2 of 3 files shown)


sqlglot/__init__.py
```diff
@@ -22,6 +22,7 @@
     or_,
     select,
     subquery,
+    DataType,
```
tobymao, line 25: hm not sure about this one, we don’t expose any others. thoughts?

sqlglot/expressions.py
```diff
@@ -3673,25 +3673,57 @@ def table_(table, db=None, catalog=None, quoted=None, alias=None) -> Table:
     )
 
 
-def values(values, alias=None) -> Values:
+def values(
+    values: t.Sequence[t.Tuple[str | Expression, ...]],
+    alias: t.Optional[str] = None,
+    columns_and_types: t.Optional[t.List[str | t.Tuple[str, str | DataType]]] = None,
```
tobymao, line 3679: make this a dict and call it schema?

sqlglot/__init__.py
eakmanrq: I don't feel strongly either way. To me it seemed like DataType would be a common expressions users would want and therefore was as common/important as these other expressions. It is not a big deal to switch though.

sqlglot/expressions.py
```diff
@@ -3673,25 +3673,57 @@ def table_(table, db=None, catalog=None, quoted=None, alias=None) -> Table:
...
+                column_names.append(to_identifier(column[0]))
+                if isinstance(column[1], str):
+                    column_types.append(DataType.build(column[1]))
+                else:
+                    column_types.append(column[1])
+            else:
+                column_names.append(to_identifier(column))
+    if column_types:
+        casted_rows = []
+        for row in values:
```
tobymao, line 3712: hmm… you’re just doing a cast on everything? im not sure we should do that, it’s more efficient to just nest this in a select and cast it, so we may not need this api?

sqlglot/__init__.py
tobymao: yea let’s get rid of it

sqlglot/expressions.py
tobymao: no because the sql string is way bigger right? let’s chat tomorrow about this

sqlglot/expressions.py
tobymao: maybe make it an iterable then

sqlglot/expressions.py
eakmanrq: Right the size of the expression is a fair concern. I think we should still make it an option for users although we might in other contexts decide to not use it and cast the select.

sqlglot/expressions.py
eakmanrq: did you mean collection (like what we use in `ensure_collection`)?

tobymao: no i meant the type

tobymao (approved):

PR #820, opened 2022-12-15
Refactor SQLGlot workflow, CI/CD, improve pdoc template

GeorgeSittas:
This is an attempt to improve the workflow in SQLGlot.

- Replaced the `run_checks.sh` script with a `Makefile`.
- Added pre-commit hooks for lint checks.
- Moved dev. dependencies in `setup.py`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: install install-dev install-pre-commit test style docs docs-serve
+.PHONY: install install-dev install-pre-commit test style check docs docs-serve
 
 install:
 	pip install -e .
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ test:
 style:
 	pre-commit run --all-files
 
+check: style test
+
 docs:
 	pdoc/cli.py -o pdoc/docs
 
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,15 @@
-.PHONY: install install-dev tests style docs docs-serve
+.PHONY: install install-dev install-pre-commit test style docs docs-serve
 
 install:
 	pip install sqlglot
 
 install-dev:
 	pip install -e ".[dev]"
+
+install-pre-commit:
 	pre-commit install
 
-tests:
+test:
 	python -m unittest
 
 style:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: install install-dev install-pre-commit test style docs docs-serve
 
 install:
-	pip install sqlglot
+	pip install -e .
 
 install-dev:
 	pip install -e ".[dev]"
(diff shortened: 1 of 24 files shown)


Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: install install-dev tests style docs docs-serve
+
+install:
+	pip install sqlglot
+
+install-dev:
+	pip install -e ".[dev]"
+	pre-commit install
```
tobymao, line 8: we should not install pre-commit

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+tests:
```
tobymao, line 10: make test

Makefile
GeorgeSittas: Should we include a separate rule for installing it, or just ignore its installation completely?

eakmanrq: To expand on this: `pre-commit install` actually installs the git hooks for pre-commit which Toby wants to keep optional: https://pre-commit.com/#3-install-the-git-hook-scripts

So just having pre-commit pip-install allows the style command to work for everyone but then having a separate `install-pre-commit:` lets someone opt-in to having the commit hooks run automatically. 
```
install-pre-commit:
	pre-commit install
```

GeorgeSittas: Done.

Makefile
GeorgeSittas: Done.

Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY: install install-dev install-pre-commit test style docs docs-serve
+
+install:
+	pip install sqlglot
```
barakalon, line 4: I don't think this is right.

Do we need this?

Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+install-pre-commit:
+	pre-commit install
+
+test:
+	python -m unittest
+
+style:
+	pre-commit run --all-files
+
```
barakalon, line 17: There should be one command that does both `make test` and `make style`

barakalon (approved):

Makefile
GeorgeSittas: I think I messed this one up, probably meant to do something like `pip install -e .`

Makefile
GeorgeSittas: What about `make checks`?

GeorgeSittas: Done.

README.md
```diff
@@ -40,31 +40,30 @@ pip3 install sqlglot
...
-pdoc sqlglot --docformat google
+make docs-serve
 ```
 
 ## Run Tests and Lint
 
 ```
-# set `SKIP_INTEGRATION=1` to skip integration tests
-./run_checks.sh
+make style
```
barakalon, line 63: `make check`

tobymao (approved):

PR #824, opened 2022-12-20
up to tpch-20, find join key for DNF conditions

tobymao:


barakalon (approved):
Awesome - getting close!

PR #826, opened 2022-12-21
New: add support for the ALTER TABLE command

GeorgeSittas:
This is a WIP; the current goal is to add support for the ADD / DROP / ALTER COLUMN syntax.

I think I'll try to reuse `exp.ColumnDef` for `ADD COLUMN`, `exp.Drop` for `DROP COLUMN` and maybe create a new expression type for `ALTER COLUMN` -- I'm not sure if there's an existing class that's appropriate for the last one.

Sources:

- DuckDB: https://duckdb.org/docs/sql/statements/alter_table.html
- Redshift: https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html
- BigQuery: https://cloud.google.com/bigquery/docs/managing-table-schemas
- Snowflake: https://docs.snowflake.com/en/sql-reference/sql/alter-table.html
- SparkSQL: https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-alter-table.html
- PostgreSQL: https://www.postgresql.org/docs/current/sql-altertable.html


sqlglot/generator.py
```diff
@@ -1245,6 +1245,19 @@ def rollback_sql(self, expression: exp.Rollback) -> str:
         savepoint = f" TO {savepoint}" if savepoint else ""
         return f"ROLLBACK{savepoint}"
 
+    def altertable_sql(self, expression: exp.AlterTable) -> str:
+        action = expression.args["action"]
+
+        if isinstance(action, exp.ColumnDef):
+            action = f"ADD COLUMN {self.sql(action)}"
```
tobymao, line 1252: what about alter column?

GeorgeSittas: I'm going to add it soon

GeorgeSittas:
Right now we should be able to parse DuckDB's syntax forms for ADD / DROP / ALTER COLUMN. Remaining tasks are:

- Make sure we can parse these statements for the other mentioned dialects too.
- Write tests that don't include some optional keywords.
- Refactor the way we handle `ALTER *` commands in the tokenizer.
- Iterate to find opportunities and simplify the code if possible.

sqlglot/expressions.py
```diff
@@ -731,6 +731,17 @@ class ColumnDef(Expression):
     }
 
 
+class AlterColumn(Expression):
+    arg_types = {
+        "this": True,
+        "dtype": False,
```
tobymao, line 737: why dtype and not type?

sqlglot/parser.py
```diff
@@ -551,13 +553,14 @@ class Parser(metaclass=_Parser):
 
     MODIFIABLES = (exp.Subquery, exp.Subqueryable, exp.Table)
 
-    CREATABLES = {
-        TokenType.TABLE,
-        TokenType.VIEW,
+    CREATEABLE = {
```
tobymao, line 556: why was the spelling here changed? i don't think createable is a word, https://www.thefreedictionary.com/creatable

sqlglot/tokens.py
```diff
@@ -111,6 +111,7 @@ class TokenType(AutoName):
 
     # keywords
     ALIAS = auto()
+    ALTER_TABLE = auto()
```
tobymao, line 114: i thought you were going to delete this?

sqlglot/parser.py
GeorgeSittas: You're right, my bad. I should've checked -- I thought it's the same as "manageable" and changed it on a whim.
Fixed.

sqlglot/expressions.py
GeorgeSittas: Because we use `dtype` in other places in the codebase too for DataType objects, whereas I didn't see a mention of "type" besides the Expression slot.

sqlglot/tokens.py
GeorgeSittas: Done

sqlglot/dialects/spark.py
```diff
@@ -87,6 +88,21 @@ class Parser(Hive.Parser):
             "SHUFFLE_REPLICATE_NL": lambda self: self._parse_join_hint("SHUFFLE_REPLICATE_NL"),
         }
 
+        CREATABLES = {
+            *Hive.Parser.CREATABLES,  # type: ignore
+            TokenType.COLUMNS,
```
tobymao, line 93: why did you add columns? i don't see how this is used?

GeorgeSittas: good point, I missed this -- this is indeed unneeded.

tobymao (approved):

PR #858, opened 2022-12-28
Add a window clause builder to Select.

voberoi:
A few comments about this PR:

* The argument to Select used everywhere else in the code was `windows`, so I updated `QUERY_MODIFIERS` to reflect that.
* I had `window(...)` follow the same pattern as other builders of lists of expressions.
* I needed a parse function that would return one named `Window` expression.
* So I split out `_parse_named_window(...)` for that purpose -- and it's the registered parser for Window expressions. Which might be confusing -- given that window expressions can take a lot of forms.

In the future it might make sense to have `NamedWindow` as a separate expression, or `WindowClause` or something of the sort. Not sure if you want to do that now, or if this is sufficient.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1910,9 +1910,7 @@ class Select(Subqueryable):
             **opts,
         )
 
-    def window(
-        self, *expressions, append=True, dialect=None, copy=True, **opts
-    ) -> Select:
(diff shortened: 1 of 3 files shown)


tobymao (approved):
great

tobymao:
make check

to run linters and stuff

voberoi:
Done, thanks

PR #860, opened 2022-12-29
recursive cte scope fixes #856

### LINKED ISSUE #856: Support for recursive CTEs self-reference in their scope.
[Recursive CTEs](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-RECURSIVE) do not contain themselves as a CTE source when inspecting their scope.

The following CTE does not contain `t` in the `SELECT * FROM t` scope when inspecting it using `sqlglot.optimizer.scope.traverse_scope`:

```sql
WITH RECURSIVE t AS (
    SELECT * FROM t
)
SELECT 1
```

Also:

> [The RECURSIVE keyword is not usually needed after WITH in systems other than PostgreSQL](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#Common_table_expression).

tobymao:


barakalon (commented):
We should add some tests in optimizer.sql

tobymao:
@barakalon i've added a test. i went down a rabbit hole and found a bunch of bugs. this was hard

PR #862, opened 2022-12-29
Use STRING_AGG instead of GROUP_CONCAT in BigQuery.

voberoi:
For posterity: I added a docstring to the `bigquery` dialect module stating that it is meant to support Standard SQL.

`GROUP_CONCAT` is only supported in BigQuery's Legacy SQL.


sqlglot/dialects/bigquery.py
```diff
@@ -168,6 +170,7 @@ class Generator(generator.Generator):
             exp.DatetimeAdd: _date_add_sql("DATETIME", "ADD"),
             exp.DatetimeSub: _date_add_sql("DATETIME", "SUB"),
             exp.DateDiff: lambda self, e: f"DATE_DIFF({self.sql(e, 'this')}, {self.sql(e, 'expression')}, {self.sql(e.args.get('unit', 'DAY'))})",
+            exp.GroupConcat: rename_func("STRING_AGG"),
```
tobymao, line 173: can you also add the inverse so that string_agg can be read in as a group_concat? line 139

tobymao (approved):

PR #868, opened 2022-12-31
[WIP] Unquote columns in Snowflake values

eakmanrq:
For whatever reason, something that I researched more deeply months ago when I wrote this, Snowflake is picky about the columns being quotes in the `VALUES` expression. This breaks the DataFrame API when someone creates a DataFrame from values which can be a common use cases when running tests (less common in production). 

The fix currently in the PR involves two things:
1. Adds an argument to `quote_identities` that allows you to tell it to now quote the columns in `VALUES` expression. 
2. When the DataFrame API generates sql for the snowflake dialect is makes sure to enable not quoting those columns so that way the output is valid.

The problem though is that this approach only fixes the problem for the DataFrame API and also only fixes it when running optimize. Ideally all Snowflake generated `VALUES` SQL would be valid regardless of the context in which it was generated. Looking for feedback on the best way to do this within sqlglot.

### DIFF
--- a/sqlglot/dataframe/sql/dataframe.py
+++ b/sqlglot/dataframe/sql/dataframe.py
@@ -303,7 +303,8 @@ class DataFrame:
         for expression_type, select_expression in select_expressions:
             select_expression = select_expression.transform(replace_id_value, replacement_mapping)
             if optimize:
-                select_expression = optimize_func(select_expression)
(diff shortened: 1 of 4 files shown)


tobymao:
can’t you just do this in the snowflake generator for values?

PR #869, opened 2022-12-31
Have snowflake generate values columns without quotes

eakmanrq:
Snowflake has a bug (I created a ticket) where it can't execute table aliases for VALUES expressions that have quotes around the column names. Since the usage of quotes has to be consistent that also means you can't quote these columns when you reference them throughout your query.

This PR changes the Snowflake generator so that way it properly unquotes these columns.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -115,7 +115,7 @@ def _values_sql(self, expression: exp.Values) -> str:
     if alias:
         for column in alias.args.get("columns", []):
             column.set("quoted", False)
-    return self.values_sql(expression)
+    return self.no_identify(lambda: self.values_sql(expression))
 
 
 def _select_sql(self, expression: exp.Select) -> str:
@@ -136,7 +136,7 @@ def _select_sql(self, expression: exp.Select) -> str:
     for identifier in all_identifiers:
         if identifier in values_identifiers:
             identifier.set("quoted", False)
-    return self.select_sql(expression)
+    return self.no_identify(lambda: self.select_sql(expression))
 
 
 class Snowflake(Dialect):
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -110,17 +110,27 @@ def _datatype_sql(self, expression):
 
 
 def _values_sql(self, expression: exp.Values) -> str:
-    """Due to a bug in Snowflake we want to make sure that all columns in a VALUES table alias are unquoted."""
+    """Due to a bug in Snowflake we want to make sure that all columns in a VALUES table alias are unquoted.
+
+    We also want to make sure that after we find matches where we need to unquote a column that we prevent users
+    from adding quotes to the column by using the `identify` argument when generating the SQL.
+    """
     alias = expression.args.get("alias")
+    unquoted_a_column = False
     if alias:
         for column in alias.args.get("columns", []):
             column.set("quoted", False)
-    return self.no_identify(lambda: self.values_sql(expression))
+            unquoted_a_column = True
+    if unquoted_a_column:
+        return self.no_identify(lambda: self.values_sql(expression))
+    return self.values_sql(expression)
 
 
 def _select_sql(self, expression: exp.Select) -> str:
     """Due to a bug in Snowflake we want to make sure that all columns in a VALUES table alias are unquoted and also
-    that all columns in a SELECT are unquoted.
+    that all columns in a SELECT are unquoted. We also want to make sure that after we find matches where we need
+    to unquote a column that we prevent users from adding quotes to the column by using the `identify` argument when
+    generating the SQL.
 
     Note: We make an assumption that any columns referenced in a VALUES expression should be unquoted throughout the
     expression. This might not be true in a case where the same column name can be sourced from another table that can
@@ -132,11 +142,17 @@ def _select_sql(self, expression: exp.Select) -> str:
             [v.args.get("alias", exp.Alias()).args.get("columns", []) for v in values_expressions]
         )
     )
+    if not values_identifiers:
+        return self.select_sql(expression)
     all_identifiers = expression.find_all(exp.Identifier)
+    unquoted_a_column = False
     for identifier in all_identifiers:
         if identifier in values_identifiers:
             identifier.set("quoted", False)
-    return self.no_identify(lambda: self.select_sql(expression))
+            unquoted_a_column = True
+    if unquoted_a_column:
+        return self.no_identify(lambda: self.select_sql(expression))
+    return self.select_sql(expression)
 
 
 class Snowflake(Dialect):
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -9,7 +9,7 @@ from sqlglot.dialects.dialect import (
     var_map_sql,
 )
 from sqlglot.expressions import Literal
-from sqlglot.helper import seq_get
+from sqlglot.helper import flatten, seq_get
 from sqlglot.tokens import TokenType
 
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -109,6 +109,52 @@ def _datatype_sql(self, expression):
...
+    generating the SQL.
+
+    Note: We make an assumption that any columns referenced in a VALUES expression should be unquoted throughout the
+    expression. This might not be true in a case where the same column name can be sourced from another table that can
+    properly quote but should be true in most cases.
+    """
+    values_expressions = expression.find_all(exp.Values)
+    values_identifiers = set(
+        flatten(
+            [v.args.get("alias", exp.Alias()).args.get("columns", []) for v in values_expressions]
```
GeorgeSittas, line 142:
```suggestion
            v.args.get("alias", exp.Alias()).args.get("columns", []) for v in values_expressions
```

GeorgeSittas (commented):
What about quoted columns in other places in the query? For example:

```sql
SELECT `c0`, `c1`
FROM (VALUES (1, 2), (3, 4)) AS `t0`(`c0`, `c1`)
WHERE `c0` = 1  -- Do we need to unquote this too?
```

Looks good otherwise.

sqlglot/dialects/snowflake.py
```diff
@@ -216,11 +262,13 @@ class Generator(generator.Generator):
             exp.Parameter: lambda self, e: f"${self.sql(e, 'this')}",
             exp.PartitionedByProperty: lambda self, e: f"PARTITION BY {self.sql(e, 'this')}",
             exp.Matches: rename_func("DECODE"),
+            exp.Select: _select_sql,
```
tobymao, line 265: move these private methods into the Generator and just override the methods

def select_sql(...)

sqlglot/dialects/snowflake.py
```diff
@@ -109,6 +109,52 @@ def _datatype_sql(self, expression):
...
+    """Due to a bug in Snowflake we want to make sure that all columns in a VALUES table alias are unquoted.
+
+    We also want to make sure that after we find matches where we need to unquote a column that we prevent users
+    from adding quotes to the column by using the `identify` argument when generating the SQL.
+    """
+    alias = expression.args.get("alias")
+    unquoted_a_column = False
+    if alias:
+        for column in alias.args.get("columns", []):
+            column.set("quoted", False)
```
tobymao, line 122: delete this line, it's not necessary and we shouldn't be mutating things in the generator

sqlglot/dialects/snowflake.py
```diff
@@ -109,6 +109,52 @@ def _datatype_sql(self, expression):
...
+def _values_sql(self, expression: exp.Values) -> str:
+    """Due to a bug in Snowflake we want to make sure that all columns in a VALUES table alias are unquoted.
+
+    We also want to make sure that after we find matches where we need to unquote a column that we prevent users
+    from adding quotes to the column by using the `identify` argument when generating the SQL.
+    """
+    alias = expression.args.get("alias")
+    unquoted_a_column = False
+    if alias:
+        for column in alias.args.get("columns", []):
```
tobymao, line 121: unquoted = alias and alias.args.get("columns")

sqlglot/dialects/snowflake.py
```diff
@@ -109,6 +109,52 @@ def _datatype_sql(self, expression):
...
+            [v.args.get("alias", exp.Alias()).args.get("columns", []) for v in values_expressions]
+        )
+    )
+    if not values_identifiers:
+        return self.select_sql(expression)
+    all_identifiers = expression.find_all(exp.Identifier)
+    unquoted_a_column = False
+    for identifier in all_identifiers:
+        if identifier in values_identifiers:
+            identifier.set("quoted", False)
```
tobymao, line 151: same stuff here

sqlglot/dialects/snowflake.py
```diff
@@ -109,6 +109,52 @@ def _datatype_sql(self, expression):
...
+            unquoted_a_column = True
+    if unquoted_a_column:
+        return self.no_identify(lambda: self.select_sql(expression))
```
tobymao, line 154: don't you only want to do no_identify on expressions using the identifier?

sqlglot/dialects/snowflake.py
eakmanrq: I don't understand. If I remove this line then we don't unquote anything.

tobymao: then the code isn’t correct. test it with identify=True

eakmanrq: Works as expected:
```
>>> sqlglot.parse_one("SELECT `c0`, `c1` FROM (VALUES (1, 2), (3, 4)) AS `t0`(`c0`, `c1`)", read="spark").sql(dialect="snowflake", identify=True)
'SELECT c0, c1 FROM (VALUES (1, 2), (3, 4)) AS "t0"(c0, c1)'
```

eakmanrq:
> What about quoted columns in other places in the query? For example:
> 
> ```sql
> SELECT `c0`, `c1`
> FROM (VALUES (1, 2), (3, 4)) AS `t0`(`c0`, `c1`)
> WHERE `c0` = 1  -- Do we need to unquote this too?
> ```
> 
> Looks good otherwise.

Yeah this will also unquote that column in the where which it needs to. Snowflake requires that once a column is quotes that it is always quotes when it is referenced throughout the query.

sqlglot/dialects/snowflake.py
tobymao: why would you need to set quoted then?

tobymao: i see. it’s on only inverse true. need a new flag on no identify

eakmanrq: This line I think is where we are misaligned: https://github.com/tobymao/sqlglot/blob/e69fa6ce02e136193a404dd56bb0d0e86870601f/sqlglot/generator.py#L549

It is saying that if either `identify=True` or the column is defined as `quoted` then we add the quotes. So if we just set `identify` as `False` then we still have `quoted` set to true and therefore the quote are set. These lines where I set `quoted` to `False` mean that we ensure that both `quoted` is `False` and `identify` is `False` and therefore we don't add the quotes on those columns we are targeting.

tobymao: yea but it’s a mutation. we should add a force no identify

eakmanrq: Well I still want to identify references that aren't the column. I just want to modify the columns that are part of the values alias.

tobymao: right l, so only no identify should be used for those nodes. or you need to do this in a non mutating manner, maybe through transform

eakmanrq: So no identify on the nodes won't work as we discussed above. I don't see how we would do a forcing solution without mutating them.

Can explain what you mean by using a transform? An example you can reference?

GeorgeSittas: What if we copied the original expression instead? I think we already do something similar for `STRING_AGG` to avoid mutations in the generator.

tobymao: right now you’re calling set directly on the nodes. this is a mutation .

instead of you call transform on the select node, you will make a copy and not mutate the original. your original code used a transform i think.

GeorgeSittas: > your original code used a transform i think.

Do you mean `STRING_AGG`, or Ryan's? In any case, just to clarify: I meant something like:

```python
# Do this if a mutation is needed
...
```

tobymao: i mean ryan.

i’m suggesting

expression.transform(…)

tobymao (approved):

PR #874, opened 2023-01-02
Transpile DATEDIFF to postgres

GeorgeSittas:
This is a WIP -- thought I'd start a PR early to get some feedback and iterate faster. The units "week", "month", "quarter" and "year" are currently missing from the implementation, but I'll add them soon. I'm trying to figure out how to reduce code duplication as much as possible for them.

Regarding my current approach, I'm unsure whether hardcoding the mult / div operators as strings is the way to go. An alternative could be to return a tuple that contains the factor and a flag that determines whether to multiply or to divide, although it's not necessary cleaner.

cc: @tobymao

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -15,6 +15,15 @@ from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
 from sqlglot.transforms import delegate, preprocess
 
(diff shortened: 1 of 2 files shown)


tobymao:
seems ok

GeorgeSittas:
Ok cool, I'll get the other units in too then.

tobymao (approved):

PR #880, opened 2023-01-03
Add support for parsing stdin in CLI

mgradowski:
Hey! The CLI is missing a way to read files or output from other tools, which is essential for my use case (statements with hundreds of LOC). My contribution is a pretty low effort way to add it.

I decided to use `-` as a sentinel value [1] to indicate that the CLI should parse the stdin instead of argv for two reasons:
1. It's an established convention in many of GNU coreutil tools [2].
2. I don't think there's a dialect for which `-` would parse successfully.

[1] https://en.wikipedia.org/wiki/Sentinel_value
[2] https://www.gnu.org/software/coreutils/manual/coreutils.html#Common-options

### DIFF
--- a/sqlglot/__main__.py
+++ b/sqlglot/__main__.py
@@ -1,9 +1,15 @@
 import argparse
+import sys
 
 import sqlglot
 
 parser = argparse.ArgumentParser(description="Transpile SQL")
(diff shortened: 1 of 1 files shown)


tobymao:
thanks

PR #881, opened 2023-01-04
Add wider support for timestr and datestr

eakmanrq:
A common approach for getting a "time string" or "date string" into timestamp/date is to just cast them to the desired type. Since these are standardized formats (iso formats) then engines generally supporting that cast without having to use custom function.

This PR adds supports to more engines to support this pattern and also standardizes the way we do this across all dialects into specific functions.


sqlglot/dialects/presto.py
```diff
@@ -224,8 +221,8 @@ class Generator(generator.Generator):
             exp.StructExtract: struct_extract_sql,
             exp.TableFormatProperty: lambda self, e: f"TABLE_FORMAT='{e.name.upper()}'",
             exp.FileFormatProperty: lambda self, e: f"FORMAT='{e.name.upper()}'",
-            exp.TimeStrToDate: _timestr_to_time_sql,
-            exp.TimeStrToTime: _timestr_to_time_sql,
+            exp.TimeStrToDate: timestrtotime_sql,
```
eakmanrq, line 224: This looks strange to me since we don't actually cast it into a date but this matches existing functionality.

tobymao (approved):

PR #884, opened 2023-01-04
Add support for clickhouse's parametric function syntax

### LINKED ISSUE #882: ClickHouse quantile/quantileIf functions fail to parse
ClickHouse has a somewhat unique way of spelling these:

```
9bf006fcf914 :) select quantile(0.5)(a) from (select 1 a union all select 2);

SELECT quantile(0.5)(a)
FROM
(
    SELECT 1 AS a
    UNION ALL
    SELECT 2
)

Query id: fe37056a-efe0-45bf-9d34-95de2916ecef

┌─quantile(0.5)(a)─┐
│              1.5 │
└──────────────────┘

1 row in set. Elapsed: 0.002 sec.

9bf006fcf914 :) select quantileIf(0.5)(a, true) from (select 1 a union all select 2);

SELECT quantileIf(0.5)(a, true)
FROM
(
    SELECT 1 AS a
    UNION ALL
    SELECT 2
)

Query id: d3b2e971-c9af-4a6c-8ba3-93caacc246bc

┌─quantileIf(0.5)(a, true)─┐
│                      1.5 │
└──────────────────────────┘

1 row in set. Elapsed: 0.002 sec.

9bf006fcf914 :) select quantileIf(0.5)(a, false) from (select 1 a union all select 2);

SELECT quantileIf(0.5)(a, false)
FROM
(
    SELECT 1 AS a
    UNION ALL
    SELECT 2
)

Query id: 27307fa4-2b25-4382-8e6c-ef3d4eb1b306

┌─quantileIf(0.5)(a, false)─┐
│                       nan │
└───────────────────────────┘

1 row in set. Elapsed: 0.002 sec.
```

When parsed with `sqlglot` I get a parse error:

```
In [1]: import sqlglot as sg

In [2]: sg.__version__
Out[2]: '10.3.2'

In [3]: sg.parse_one("select quantile(0.5)(a) from (select 1 a union all select 2)", read="clickhouse")
...
ParseError: Required keyword: 'quantile' missing for <class 'sqlglot.expressions.Quantile'>. Line 1, Col: 20.
  select quantile(0.5)(a) from (select 1 a union all select 2)
```

GeorgeSittas:
Phew... this was more tricky than expected.

Fixes #882

### DIFF
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -356,6 +356,15 @@ def flatten(values: t.Iterable[t.Iterable[t.Any] | t.Any]) -> t.Generator[t.Any,
             yield value
 
 
+def count_params(function: t.Callable) -> int:
+    """
+    Returns the number of formal parameters expected by a function, without counting "self"
+    and "cls", in case of instance and class methods, respectively.
+    """
+    count = function.__code__.co_argcount
+    return count - 1 if inspect.ismethod(function) else count
+
+
 def dict_depth(d: t.Dict) -> int:
     """
     Get the nesting depth of a dictionary.
@@ -374,6 +383,7 @@ def dict_depth(d: t.Dict) -> int:
 
     Args:
         d (dict): dictionary
+
     Returns:
         int: depth
     """
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2049,7 +2049,23 @@ class Parser(metaclass=_Parser):
             args = self._parse_csv(self._parse_lambda)
 
             if function:
-                this = function(args)
+                params = None
+
+                # Clickhouse supports function calls like foo(x, y)(z), so for these we need to also parse the
+                # second parameter list (i.e. "(z)") and the corresponding function will receive both arg lists.
+                if function.__code__.co_name == "<lambda>" and function.__code__.co_argcount == 2:
+                    index = self._index
+                    self._match_r_paren()
+
+                    if self._match(TokenType.L_PAREN):
+                        params = self._parse_csv(self._parse_lambda)
+                    else:
+                        self._retreat(index)
+
+                    this = function(args, params)
+                else:
+                    this = function(args)
+
                 self.validate_expression(this, args)
             else:
                 this = self.expression(exp.Anonymous, this=this, expressions=args)
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -5,7 +5,13 @@ import typing as t
 
 from sqlglot import exp
 from sqlglot.errors import ErrorLevel, ParseError, concat_messages, merge_errors
-from sqlglot.helper import apply_index_offset, ensure_collection, ensure_list, seq_get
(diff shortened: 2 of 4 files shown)


sqlglot/parser.py
```diff
@@ -2049,7 +2049,23 @@ def _parse_function(self, functions=None):
             args = self._parse_csv(self._parse_lambda)
 
             if function:
-                this = function(args)
+                params = None
+
+                # Clickhouse supports function calls like foo(x, y)(z), so for these we need to also parse the
+                # second parameter list (i.e. "(z)") and the corresponding function will receive both arg lists.
+                if function.__code__.co_name == "<lambda>" and function.__code__.co_argcount == 2:
```
GeorgeSittas, line 2056: The `<lambda>` check was added because without it we would also match functions like `format_time_lambda` which expects two arguments (but the second one is a helper we hard code into the calls, if I'm not mistaken).

tobymao: this doesn't seem correct to me, format_time_lambda should return a lambda that only has one argument

sqlglot/parser.py
```diff
@@ -2049,7 +2049,23 @@ def _parse_function(self, functions=None):
...
+                    index = self._index
+                    self._match_r_paren()
```
tobymao, line 2058: i don't like doing the retreat here, can we do this another way?

sqlglot/parser.py
GeorgeSittas: fixed.

sqlglot/parser.py
GeorgeSittas: fixed.

sqlglot/helper.py
```diff
@@ -356,6 +356,15 @@ def flatten(values: t.Iterable[t.Iterable[t.Any] | t.Any]) -> t.Generator[t.Any,
             yield value
 
 
+def count_params(function: t.Callable) -> int:
```
tobymao, line 359: how expensive is this? if it's expensive can we do this at load time instead of run time?

sqlglot/parser.py
```diff
@@ -2049,7 +2055,18 @@ def _parse_function(self, functions=None):
             args = self._parse_csv(self._parse_lambda)
 
             if function:
-                this = function(args)
+                params = None
```
tobymao, line 2058: move this into an else statement in 2065

sqlglot/helper.py
GeorgeSittas: According to [the source](https://github.com/python/cpython/blob/ace008c531dd685a30c1dd68f9b5ba35f20171cf/Lib/inspect.py#L82-L90), seems like it's doing an instance check under the hood. Maybe it's ok?

tobymao: benchmark it,

import time

now  = time.time()
...

tobymao (approved):

cpcloud:
Thanks for doing this!

PR #890, opened 2023-01-05
Added support for multiple T-SQL features

10bas10:
- Added support for EOMONTH in T-SQL
- Made sure that if its a temporal type no precision is beeing specified in the datatypes at Hive and depending dialects
- Changed class of GETDATE() to CurrentTimestamp as it is also helding time
- Added support for SYSDATETIME
- Fixed the parsing of a CAST on a subquery

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -80,13 +80,13 @@ def _parse_eomonth(args):
     month_lag = seq_get(args, 1)
     unit = DATE_DELTA_INTERVAL.get("month")
 
-    if month_lag:
+    if month_lag is None:
+        return exp.LastDateOfMonth(this=date)
+    else:
+        # Remove month lag argument in parser as its compared with the number of arguments of the resulting class
         args.remove(month_lag)
 
-    if month_lag is not None:
         return exp.LastDateOfMonth(this=exp.DateAdd(this=date, expression=month_lag, unit=unit))
-    else:
-        return exp.LastDateOfMonth(this=date)
 
 
 def generate_date_delta_with_unit_sql(self, e):
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -75,6 +75,20 @@ def _parse_format(args):
     )
 
 
+def _parse_eomonth(args):
+    date = seq_get(args, 0)
+    month_lag = seq_get(args, 1)
+    unit = DATE_DELTA_INTERVAL.get("month")
+
+    if month_lag:
+        args.remove(month_lag)
+
+    if month_lag is not None:
+        return exp.LastDateOfMonth(this=exp.DateAdd(this=date, expression=month_lag, unit=unit))
+    else:
+        return exp.LastDateOfMonth(this=date)
+
+
 def generate_date_delta_with_unit_sql(self, e):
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -75,6 +75,20 @@ def _parse_format(args):
     )
 
 
+def _parse_eomonth(args):
+    date = seq_get(args, 0)
+    month_lag = seq_get(args, 1)
+    unit = DATE_DELTA_INTERVAL.get("month")
+
+    if month_lag:
+        args.remove(month_lag)
```
GeorgeSittas, line 84: Why was this needed?

sqlglot/dialects/tsql.py
```diff
@@ -75,6 +75,20 @@ def _parse_format(args):
...
+
+    if month_lag is not None:
```
tobymao, line 86: switch order, if month is None:

sqlglot/parser.py
```diff
@@ -2001,7 +2001,9 @@ def _parse_primary(self):
             self._match_r_paren()
 
             if isinstance(this, exp.Subqueryable):
-                this = self._parse_set_operations(self._parse_subquery(this))
+                this = self._parse_set_operations(
+                    self._parse_subquery(this=this, parse_alias=False)
```
tobymao, line 2005: can you write a test case for this? what's going on here

GeorgeSittas: I was also confused, but I think there's a test in this PR that covers this:

```python
SELECT CAST((SELECT x FROM y) AS VARCHAR) AS test
```

Either way, it seems non-trivial, so maybe a comment could be used here?

tobymao: i got it now

sqlglot/dialects/tsql.py
10bas10: This is because in validate_expression a check is performed to compare the number of arguments. As EOMONTH has two, while the LAST_DAY has one, it eould gibe an error.

https://github.com/tobymao/sqlglot/blob/9015f85b79c24ed43b0e667f9b46d20884050849/sqlglot/parser.py#L758

GeorgeSittas: I see, I think a comment would help here for future reference.

10bas10: Will do

sqlglot/dialects/tsql.py
```diff
@@ -75,6 +75,20 @@ def _parse_format(args):
...
+    month_lag = seq_get(args, 1)
+    unit = DATE_DELTA_INTERVAL.get("month")
+
+    if month_lag is None:
+        return exp.LastDateOfMonth(this=date)
+    else:
+        # Remove month lag argument in parser as its compared with the number of arguments of the resulting class
+        args.remove(month_lag)
+
+        return exp.LastDateOfMonth(this=exp.DateAdd(this=date, expression=month_lag, unit=unit))
```
GeorgeSittas, line 89:
```suggestion

    # Remove month lag argument in parser as its compared with the number of arguments of the resulting class
    args.remove(month_lag)

    return exp.LastDateOfMonth(this=exp.DateAdd(this=date, expression=month_lag, unit=unit))
```

GeorgeSittas (approved):

PR #891, opened 2023-01-05
Schema normalization: lower() column names if not quoted

GeorgeSittas:
This PR introduces a way to normalize schemas, by converting their column's names to lowercase, unless they're quoted. The problem it aims to fix is the following:

In the current main branch, the optimizer fails for the query below due to the inconsistency between schema column names and those produced by the `lower_identities` rule in the optimizer.

```python
import sqlglot

print(
    optimize(
        sqlglot.parse_one("""
            SELECT A OR (B OR (C AND D))
            FROM x
            WHERE Z = date '2021-01-01' + INTERVAL '1' month OR 1 = 0
        """),
        schema={"x": {"A": "INT", "B": "INT", "C": "INT", "D": "INT", "Z": "STRING"}}
    ).sql(pretty=True)
)
```

### DIFF
--- a/sqlglot/optimizer/optimizer.py
+++ b/sqlglot/optimizer/optimizer.py
@@ -15,7 +15,7 @@ from sqlglot.optimizer.pushdown_projections import pushdown_projections
 from sqlglot.optimizer.qualify_columns import qualify_columns
 from sqlglot.optimizer.qualify_tables import qualify_tables
 from sqlglot.optimizer.unnest_subqueries import unnest_subqueries
-from sqlglot.schema import AbstractMappingSchema, ensure_schema
+from sqlglot.schema import ensure_schema
 
 RULES = (
     lower_identities,
@@ -58,10 +58,6 @@ def optimize(expression, schema=None, db=None, catalog=None, rules=RULES, **kwar
         sqlglot.Expression: optimized expression
     """
     schema = ensure_schema(schema or sqlglot.schema)
-
-    if isinstance(schema, AbstractMappingSchema):
-        schema.normalize()
-
     possible_kwargs = {"db": db, "catalog": catalog, "schema": schema, **kwargs}
     expression = expression.copy()
     for rule in rules:
--- a/sqlglot/optimizer/optimizer.py
+++ b/sqlglot/optimizer/optimizer.py
@@ -15,6 +15,7 @@ from sqlglot.optimizer.pushdown_projections import pushdown_projections
 from sqlglot.optimizer.qualify_columns import qualify_columns
 from sqlglot.optimizer.qualify_tables import qualify_tables
 from sqlglot.optimizer.unnest_subqueries import unnest_subqueries
+from sqlglot.schema import AbstractMappingSchema, ensure_schema
 
 RULES = (
     lower_identities,
@@ -56,7 +57,12 @@ def optimize(expression, schema=None, db=None, catalog=None, rules=RULES, **kwar
     Returns:
         sqlglot.Expression: optimized expression
     """
-    possible_kwargs = {"db": db, "catalog": catalog, "schema": schema or sqlglot.schema, **kwargs}
+    schema = ensure_schema(schema or sqlglot.schema)
+
+    if isinstance(schema, AbstractMappingSchema):
+        schema.normalize()
+
+    possible_kwargs = {"db": db, "catalog": catalog, "schema": schema, **kwargs}
     expression = expression.copy()
     for rule in rules:
 
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -193,7 +193,7 @@ class MappingSchema(AbstractMappingSchema[t.Dict[str, str]], Schema):
                 column = sqlglot.parse_one(column_name, read=self.dialect, into=exp.Identifier)  # type: ignore
                 assert isinstance(column, exp.Identifier)
 
-                normalized_name = column.sql()
+                normalized_name = column.sql(dialect=self.dialect)
                 if not column.quoted:
                     normalized_name = normalized_name.lower()
 
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -198,26 +198,13 @@ class MappingSchema(AbstractMappingSchema[t.Dict[str, str]], Schema):
             columns = _nested_get(schema, *zip(keys, keys))
             assert columns is not None
 
-            normalized_keys = []
-            for key in keys:
-                key_id = sqlglot.parse_one(key, read=self.dialect, into=exp.Identifier)  # type: ignore
-                assert isinstance(key_id, exp.Identifier)
-
-                normalized_name = key_id.sql(dialect=self.dialect)
-                if not key_id.quoted:
-                    normalized_name = normalized_name.lower()
-
-                normalized_keys.append(normalized_name)
-
+            normalized_keys = [self._normalize_name(key) for key in keys]
             for column_name, column_type in columns.items():
-                column_id = sqlglot.parse_one(column_name, read=self.dialect, into=exp.Identifier)  # type: ignore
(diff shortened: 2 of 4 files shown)


sqlglot/optimizer/optimizer.py
```diff
@@ -56,7 +57,12 @@ def optimize(expression, schema=None, db=None, catalog=None, rules=RULES, **kwar
     Returns:
         sqlglot.Expression: optimized expression
     """
-    possible_kwargs = {"db": db, "catalog": catalog, "schema": schema or sqlglot.schema, **kwargs}
+    schema = ensure_schema(schema or sqlglot.schema)
+
+    if isinstance(schema, AbstractMappingSchema):
```
tobymao, line 62: why do you check the type here?

sqlglot/schema.py
```diff
@@ -96,6 +96,27 @@ def supported_table_args(self) -> t.Tuple[str, ...]:
 
         return self._supported_table_args
 
+    def normalize(self) -> None:
```
tobymao, line 99: make this private and call it in the iniatilizer have it take an argument of mapping and return a mapping

sqlglot/optimizer/optimizer.py
GeorgeSittas: Because if the user supplies a custom `Schema` that doesn't support the `normalize` method, it'll crash. Might not be the best way to do it though, maybe we could check if the schema object has a `normalize` attribute instead? Not sure.

tobymao: normalize should not have to be called, it should be a part of the init

sqlglot/schema.py
GeorgeSittas: That's how I initially approached it, but I thought that maybe it's too opinionated to implicitly change the user-supplied schema @ init time? Also, there were some problems with the executor, because `Tables` inherits from `AbstractMappingSchema` and so it would implicitly use that normalization logic.

Are we sure we want to have this be the default behavior in `__init__`?

tobymao: i don't understand why you wouldn't want it

GeorgeSittas: Because there might be another use case (besides optimization) where the user doesn't want their schema changed under the hood, for whatever reason. As I see it, the problem right now is local to the optimizer, but that's only because we change the identifiers to lowercase anyway. No problem with moving it back to `__init__`, though.

barakalon: > Tables inherits from AbstractMappingSchema and so it would implicitly use that normalization logic.

I think we need similar logic for `Tables`, right? We just might not want to transform the column names, since that might be slow...

We can also just move this to `MappingSchema` for now, which would avoid any conflict with `Tables`.

tests/test_schema.py
```diff
@@ -184,3 +184,12 @@ def test_schema_get_column_type(self):
...
         self.assertEqual(schema.get_column_type("foo", "bar").this, exp.DataType.Type.INT)
+
+    def test_schema_normalization(self):
+        schema = ensure_schema(
+            {"x": {"y": {"z": {"a": "INT", '"B"': "VARCHAR"}, "w": {"C": "INT"}}}}
+        )
+        expected_mapping = {"x": {"y": {"z": {"a": "INT", '"B"': "VARCHAR"}, "w": {"c": "INT"}}}}
+
+        schema.normalize()
+        self.assertEqual(schema.mapping, expected_mapping)
```
barakalon, line 195: While `Schema.mapping` isn't explicitly private, this still feels like we're checking some private attribute of the schema that might change.

Can we test this using more explicitly public APIs? e.g. "column_names"?

tests/test_schema.py
```diff
@@ -184,3 +184,12 @@ def test_schema_get_column_type(self):
 
         schema = MappingSchema({"foo": {"bar": parse_one("INT", into=exp.DataType)}})
         self.assertEqual(schema.get_column_type("foo", "bar").this, exp.DataType.Type.INT)
+
+    def test_schema_normalization(self):
+        schema = ensure_schema(
+            {"x": {"y": {"z": {"a": "INT", '"B"': "VARCHAR"}, "w": {"C": "INT"}}}}
+        )
+        expected_mapping = {"x": {"y": {"z": {"a": "INT", '"B"': "VARCHAR"}, "w": {"c": "INT"}}}}
```
barakalon, line 192: Will we have issues preserving the quotes of `"B"`? This is something we should be testing better...

sqlglot/schema.py
```diff
@@ -96,6 +96,27 @@ def supported_table_args(self) -> t.Tuple[str, ...]:
...
+        schema = self.mapping
+        flattened_schema = flatten_schema(schema, depth=dict_depth(schema) - 1)
+
+        normalized_mapping: t.Dict = {}
+        for keys in flattened_schema:
+            columns = _nested_get(schema, *zip(keys, keys))
+            assert columns is not None
+
+            for column_name, column_type in columns.items():
+                column = exp.to_identifier(column_name)
```
barakalon, line 109: I don't think this is right.

I think we want `parse_one(column_name, into=exp.Identifier)`

sqlglot/schema.py
```diff
@@ -96,6 +96,27 @@ def supported_table_args(self) -> t.Tuple[str, ...]:
...
+    def normalize(self) -> None:
+        schema = self.mapping
+        flattened_schema = flatten_schema(schema, depth=dict_depth(schema) - 1)
+
+        normalized_mapping: t.Dict = {}
+        for keys in flattened_schema:
+            columns = _nested_get(schema, *zip(keys, keys))
+            assert columns is not None
+
+            for column_name, column_type in columns.items():
```
barakalon, line 108: Don't we also need to normalize catalog, db, and table names?

sqlglot/schema.py
GeorgeSittas: Got it, I'll try to temporarily move it to `MappingSchema` and take another pass when the other issues are resolved and see if it can be improved.

tests/test_schema.py
GeorgeSittas: Good point. Building on this, should we also prefix `mapping` with an underscore to make it feel more "private"? I see that it's only used in the executor's `__init__.py` file at the moment, but maybe we could move that logic inside the `Tables` class.

tests/test_schema.py
GeorgeSittas: Again, good point. Will update this, thanks.

sqlglot/schema.py
GeorgeSittas: Why do we need to parse this? Did I miss something?

tests/test_schema.py
barakalon: Possibly. Maybe let's worry about that later?

sqlglot/schema.py
barakalon: `to_identifier` assumes the string is the identifier and adds quotes if necessary.

```python
exp.to_identifier('"S"').sql()
# '""S""'

parse_one('"S"', into=exp.Identifier).sql()
# '"S"'
```

tests/test_schema.py
GeorgeSittas: I updated the test method -- happy to add more cases if you think I missed something.

GeorgeSittas:
This is ready for another review -- I think I addressed all raised issues.

sqlglot/schema.py
```diff
@@ -180,6 +181,33 @@ def copy(self, **kwargs) -> MappingSchema:
             }
         )
 
+    def normalize(self, schema: t.Dict) -> t.Dict:
```
tobymao, line 184: you can make this private

GeorgeSittas: Done

tobymao (approved):

PR #896, opened 2023-01-06
Allow WITH [NO] DATA in CREATE TABLE AS

treysp:
Allow `WITH [NO] DATA` option in `CREATE TABLE AS`.

- Add `data` arg to Create expression
- Use text matching in `_parse_create`
- Update generator
- Add identity tests


tobymao (approved):
thanks

PR #897, opened 2023-01-06
dump/load

barakalon:
Dump/load an AST to/from JSON

Context: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1673005789088219

### DIFF
--- a/sqlglot/serde.py
+++ b/sqlglot/serde.py
@@ -1,0 +1,59 @@
+from sqlglot import expressions as exp
+
+
+def dump(node):
+    """
+    Recursively dump an AST into a JSON-serializable dict.
+    """
+    if isinstance(node, list):
+        return [dump(i) for i in node]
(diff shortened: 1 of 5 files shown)


sqlglot/serde.py
```diff
@@ -0,0 +1,59 @@
+from sqlglot import expressions as exp
+
+
+def dump(node):
```
tobymao, line 4: add types

GeorgeSittas (approved):
Looks good, what about a test for comments too, for the sake of completeness?

barakalon:
The identity tests include comments

GeorgeSittas:
Right, my bad. I thought we only had them under tests for pretty. All good.

sqlglot/serde.py
barakalon: Kind of a weird thing to annotate... but I gave it a shot

tobymao (approved):

PR #903, opened 2023-01-09
Add python type hints to parser.py

GeorgeSittas:
I made an attempt to add some type hints to the parser module. The "match" methods haven't been typed yet, because they cause a lot of errors. I may take another pass and see if I can type them too.

I've probably made some mistakes since there were a lot of changes involved, so please let me know if you find anything and I'll try to fix it.

On a side note: I'm not sure if the ROI is worth it for adding python types, in the end. It might be good for some other modules, but for the parser it didn't feel natural; returning the most general type "Expression" for most methods doesn't provide much information. Thoughts on this? (Besides that, it's definitely a frustrating process :-))

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -675,8 +675,8 @@ class Parser(metaclass=_Parser):
     ) -> t.List[t.Optional[exp.Expression]]:
         """
         Parses a list of tokens into a given Expression type. If a collection of Expression
-        types is given instead, this method will try to parse the token list into one each
-        of them, stopping at the first one for which the parsing succeeds.
+        types is given instead, this method will try to parse the token list into each one
+        of them, stopping at the first for which the parsing succeeds.
 
         Args:
             expression_types: the expression type(s) to try and parse the token list into.
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -806,7 +806,7 @@ class Parser(metaclass=_Parser):
 
         Args:
             expression: the expression to validate.
-            args: an optional list of items that was used to instantiate the expression, if it's is Func.
+            args: an optional list of items that was used to instantiate the expression, if it's a Func.
         """
         if self.error_level == ErrorLevel.IGNORE:
             return
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -60,7 +60,7 @@ class Parser(metaclass=_Parser):
             Default: "nulls_are_small"
     """
 
-    FUNCTIONS: dict[str, t.Callable] = {
+    FUNCTIONS: t.Dict[str, t.Callable] = {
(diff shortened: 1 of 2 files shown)


tobymao:
@barakalon can you take a look and make a call?

sqlglot/parser.py
```diff
@@ -681,7 +701,9 @@ def parse_into(self, expression_types, raw_tokens, sql=None):
             errors=merge_errors(errors),
         ) from errors[-1]
 
-    def _parse(self, parse_method, raw_tokens, sql=None):
+    def _parse(
+        self, parse_method: t.Callable, raw_tokens: t.List[Token], sql: t.Optional[str] = None
```
barakalon, line 705: Should parse_method be `t.Callable[[Parser], Expression]`?

barakalon (commented):
Looks good to me.

As for whether or not typing is worth it... 🤷‍♂️ 

_steam of consciousness..._

I think there are two groups to consider: **developers** of sqlglot and **users** of sqlglot.

For **users**, beyond annotating the main public methods, I think the main opportunity for type annotations would be on Expression args, as writing code that touches an AST is a bit fragile right now.

But given how Expression.arg_types works, that would be a _huge_ change. And it'd sacrifice parsing flexibility for type safety, which might not be worth it right now.

Annotating all these internal parse methods affects **developers** of sqlglot. 

Sqlglot has such a comprehensive test suite that I doubt this helps prevent many bugs. 

Does it help new folks contribute to the parser?

GeorgeSittas:
Thanks for the feedback, Barak.

I agree with what you're saying. For me, the advantage of having type information is that it _sometimes_ makes the API a bit easier to understand and reason about (e.g. "what kind of arguments does this method expect? why?"). That is, I benefit by viewing type information as _documentation_.

Besides that, having to code in the correct types and/or iterate on an API to make it adapt to new changes sometimes adds an overhead to the whole development process, that would otherwise be avoided without having types altogether.

I'm interested to see how other people think about this topic too, though.

A related article I stumbled upon recently: https://lukeplant.me.uk/blog/posts/python-type-hints-parsy-case-study/

sqlglot/parser.py
```diff
@@ -735,11 +765,26 @@ def raise_error(self, message, token=None):
...
         )
+
         if self.error_level == ErrorLevel.IMMEDIATE:
             raise error
+
         self.errors.append(error)
 
-    def expression(self, exp_class, comments=None, **kwargs):
+    def expression(
+        self, exp_class: exp._Expression, comments: t.Optional[t.List[str]] = None, **kwargs
```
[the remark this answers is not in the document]
GeorgeSittas: Done.

sqlglot/parser.py

barakalon (approved):

GeorgeSittas:
I'm checking the PR and after the cleanups I'll merge.

PR #905, opened 2023-01-10
feat: Make kind arg optional in ColumnDef expression

robert8138:
In Hive/Spark, users can define a view with additional comment clause for each column, see [here](https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-view.html#syntax). These columns actually do not need to specify the data type, so this PR makes the `kind` argument optional in ColumnDef expression. 

### Before the PR
Parsing of this type of create view statement would fail

```python
sqllgot.parse_one("CREATE TABLE z (a COMMENT 'blah')")

ParseError: Expecting ). Line 1, Col: 19.
  CREATE TABLE z (a COMMENT 'blah')
```

### After the PR

```python
In [1]: import sqlglot

In [2]: sql = "CREATE TABLE z (a COMMENT 'blah')"

In [3]: sqlglot.parse_one(sql)
Out[3]: 
(CREATE this: 
  (SCHEMA this: 
    (TABLE this: 
      (IDENTIFIER this: z, quoted: False)), expressions: 
    (COLUMNDEF this: 
      (IDENTIFIER this: a, quoted: False), constraints: 
      (COLUMNCONSTRAINT kind: 
        (COMMENTCOLUMNCONSTRAINT this: 
          (LITERAL this: blah, is_string: True))))), kind: TABLE, transient: False, external: False)
```

Feedback is welcome


sqlglot/generator.py
```diff
@@ -363,8 +363,12 @@ def columndef_sql(self, expression: exp.ColumnDef) -> str:
         constraints = self.expressions(expression, key="constraints", sep=" ", flat=True)
         exists = "IF NOT EXISTS " if expression.args.get("exists") else ""
 
+        if not kind and not constraints:
```
tobymao, line 366: can you make this just return f"{exists}{column}{kind}{constraint}"

and then above add the appropriate spaces, that way we can return all these if branches

robert8138: Updated

GeorgeSittas (approved):

tobymao (approved):

PR #911, opened 2023-01-11
Add support for clickhouse GLOBAL IN syntax

### LINKED ISSUE #909: Support "GLOBAL IN" in sql query while using transpile function
I'm using CLICKHOUSE dialect

```python
from sqlglot import transpile, Dialects
sql = """
SELECT field 
FROM table 
WHERE
    (SELECT count(*) FROM q) > 0
    AND scheme.field GLOBAL IN
    (SELECT DISTINCT field FROM q)
"""

sql = transpile(sql, read=Dialects.CLICKHOUSE, identify=True, pretty=True)[0]
```

Here is the ERROR that I get:
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 5, Col: 22.
  SELECT field 
FROM table 
where
    (select count(*) from q) > 0
    and scheme.field g̲l̲o̲b̲a̲l̲ in
    (select distinct field from q))
```

GeorgeSittas:
fixes #909

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2449,6 +2449,7 @@ class In(Predicate):
         "query": False,
         "unnest": False,
         "field": False,
+        "is_global": False,
     }
 
 
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp, generator, parser, tokens
 from sqlglot.dialects.dialect import Dialect, inline_array_sql, var_map_sql
 from sqlglot.parser import parse_var_map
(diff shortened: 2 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -2449,6 +2449,7 @@ class In(Predicate):
         "query": False,
         "unnest": False,
         "field": False,
+        "is_global": False,
```
tobymao, line 2452: just call global

GeorgeSittas: I didn't call it global on purpose, because it's awkward when setting it as a kwarg due to it being a reserved keyword.

tobymao: i think there are other cases like that such as “as”

ok we’ll up to you

tobymao (approved):

PR #913, opened 2023-01-11
Allow index creation in CREATE TABLE

treysp:
Teradata requires an index when a table is created (unless specifically overridden). It will create one automatically for a `CREATE TABLE [AS]` call, but people usually specify them explicitly. This PR allows those index specifications to be parsed.

I started by creating a new Teradata dialect, but since these changes are in the create parser/generator the dialect approach would have to contain independent copies of those. It seems like you're generally trying to avoid that, so I modified the core create parser/generator in this PR. Happy to change it to a dialect if that's your preference.

This PR implements the following changes:
- Allow `AND [NO] STATISTICS` option in `CREATE TABLE AS`
- Allow specification of index(es) in `CREATE TABLE` and `CREATE TABLE AS`
    - Support `UNIQUE` and `PRIMARY` index options
    - Support named or unnamed indexes
    - Support indexes with or without column specifications
    - Support arbitrarily many indexes

One implementation detail of note is adding a `parse_subquery_alias` argument to `_parse_select` that is passed on to `_parse_subquery` inside it. This argument prevents automatically parsing the token following a subquery as an alias.


sqlglot/generator.py
```diff
@@ -434,6 +434,26 @@ def create_sql(self, expression: exp.Create) -> str:
...
+        no_primary_index = " NO PRIMARY INDEX" if expression.args.get("no_primary_index") else ""
+
+        indexes = expression.args.get("indexes")
+        index_sql = ""
+        if indexes is not None:
+            indexes_sql = []
+            for index in indexes:
+                ind_unique = " UNIQUE" if index.args.get("unique") else ""
+                ind_primary = " PRIMARY" if index.args.get("primary") else ""
+                ind_name = f' {index.args.get("this")}' if index.args.get("this") else ""
```
tobymao, line 453: ind_name = f" {index.name}" if index.name else ""

sqlglot/parser.py
```diff
@@ -1511,6 +1535,21 @@ def _parse_index(self) -> exp.Expression:
             columns=self._parse_expression(),
         )
 
+    def _parse_create_table_index(self):
+        unique = self._match_text_seq("UNIQUE")
```
tobymao, line 1539: no need to do text because unique and primary are tokens

sqlglot/parser.py
```diff
@@ -1511,6 +1535,21 @@ def _parse_index(self) -> exp.Expression:
...
+        primary = self._match_text_seq("PRIMARY")
+        self._match(TokenType.INDEX)
+        index = self._parse_id_var()
+        return self.expression(
+            exp.Index,
+            this=index,
+            columns=self.expression(
```
tobymao, line 1546: i'm not sure we want tuples here, it should probably just be an array 

columns=self._parse_wrapped_csv(self._parse_column)

sqlglot/parser.py
```diff
@@ -945,14 +948,32 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
             this = self._parse_table(schema=True)
             properties = self._parse_properties()
             if self._match(TokenType.ALIAS):
-                expression = self._parse_select(nested=True)
+                expression = self._parse_select(nested=True, parse_subquery_alias=False)
```
tobymao, line 951: why did you need to add this?

sqlglot/parser.py
```diff
@@ -945,14 +948,32 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
...
+                if self._match_text_seq("AND", "STATISTICS"):
+                    statistics = True
+                elif self._match_text_seq("AND", "NO", "STATISTICS"):
+                    statistics = False
+
+                no_primary_index = self._match_text_seq("NO", "PRIMARY", "INDEX")
+
+                indexes = []
+                while (
+                    self._match_text_seq("UNIQUE", advance=False)
```
tobymao, line 968: use token

sqlglot/parser.py
treysp: I'm not seeing a primary token (just primary key) - should I create one or use text?

sqlglot/parser.py
treysp: Got it. Is this how I should handle it in the generator?

`ind_columns = f' ({", ".join([self.sql(col) for col in index.args.get("columns")])})'`

GeorgeSittas: > Got it. Is this how I should handle it in the generator?

You can use the `expressions` directly for this, I think. It defaults to `","` as a separator and you can provide `"columns"` as a key.

sqlglot/parser.py
treysp: This `_parse_select` calls `_parse_subquery` internally. If `parse_alias=False` isn't passed to `_parse_subquery` it assumes the token following the subquery is an alias. In this `CREATE TABLE` context, we know for sure that that token is not an alias.

For example, consider a command like `CREATE TABLE AS (SELECT a,b) INDEX (a)`. The token in the "INDEX" position will never be an alias for a subquery.

With default `_parse_subquery`:
image

With `parse_alias=False` passed down to `_parse_subquery`:
image

The latter is the output we're looking for.

Is there a better way to accomplish this?

tobymao: i didn't understand what case this would be used for, make sure this is a test case

sqlglot/parser.py
```diff
@@ -1512,6 +1533,22 @@ def _parse_index(self) -> exp.Expression:
             columns=self._parse_expression(),
         )
 
+    def _parse_create_table_index(self):
+        unique = self._prev.token_type == TokenType.UNIQUE
+        primary = self._prev.text.upper() == "PRIMARY" or self._match_text_seq("PRIMARY")
```
tobymao, line 1538: why do you check _prev.text == primary and match_text primary?

sqlglot/parser.py
```diff
@@ -946,14 +949,29 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
...
 
+                if self._match_text_seq("AND", "STATISTICS"):
+                    statistics = True
+                elif self._match_text_seq("AND", "NO", "STATISTICS"):
+                    statistics = False
+
+                no_primary_index = self._match_text_seq("NO", "PRIMARY", "INDEX")
+
+                indexes = []
+                while (
```
tobymao, line 968: maybe this should be

while (self._curr and self._curr.text.upper() in ("UNIQUE", "PRIMARY", "INDEX) 

or make UNIQUE, PRIMARY, INDEX in a constant,

then you won't need that code in the create_table_index section

that way you don't advance

tobymao: do you need this condition at all? could you do

while True:
  self._parse_create_table_index()

   if not index:
      break

sqlglot/parser.py
treysp: The text could be either "PRIMARY INDEX" or "UNIQUE PRIMARY INDEX", and the first word has already been consumed when we hit this line.

If "primary" is the first word in the text it will be _prev and "index" will be _curr. If primary is the second word in the text, "unique" will be _prev and "primary" will be _curr.

tobymao: then you can do

self._match_text_seq("PRIMARY", "INDEX") 
self._match_text_seq("UNIQUE", "PRIMARY", "INDEX")

sqlglot/parser.py
treysp: Is this an appropriate test case in test_parser.py `test_select`?

```
self.assertIsNone( 
    parse_one("create table a as (select b from c) index").find(exp.TableAlias)
    )
```

sqlglot/parser.py
treysp: Sorry, I worded that comment poorly - there are actually 6 distinct text sequences

sqlglot/parser.py
treysp: Yes, that's a much better solution!

sqlglot/parser.py
tobymao: ok

GeorgeSittas (approved):
LGTM

tobymao (approved):

PR #923, opened 2023-01-12
Fix: hive/spark identifiers can now begin with a digit

### LINKED ISSUE #916: [spark dialect] Parser doesn't work when column name starts with digits and used with alias
reproduction:

```python
from sqlglot import parse_one
sql = "select A.1a as b from test_a A"
parse_one(sql, read='spark')
```

error:
```python
  File "/opt/homebrew/lib/python3.8/site-packages/sqlglot/__init__.py", line 84, in parse_one
  File "/opt/homebrew/lib/python3.8/site-packages/sqlglot/dialects/dialect.py", line 150, in parse
    return self.parser(**opts).parse(self.tokenizer.tokenize(sql), sql)
  File "/opt/homebrew/lib/python3.8/site-packages/sqlglot/parser.py", line 666, in parse
    return self._parse(
  File "/opt/homebrew/lib/python3.8/site-packages/sqlglot/parser.py", line 730, in _parse
  File "/opt/homebrew/lib/python3.8/site-packages/sqlglot/parser.py", line 773, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 13.
  SELECT A.1a as b from test_a A
```

BTW when column name starts with digits but without alias, the parser works, but incorrectly,  the digits are recognized as table name 
```python
from sqlglot import parse_one
sql="""SELECT A.1a from test_a A"""
parse_one(sql, read='spark')
>>> (SELECT expressions: 
  (ALIAS this: 
    (COLUMN this: 
      (IDENTIFIER this: 1, quoted: False), table: 
      (IDENTIFIER this: A, quoted: False)), alias: 
    (IDENTIFIER this: a, quoted: False)), from: 
  (FROM expressions: 
    (TABLE this: 
      (IDENTIFIER this: test_a, quoted: False), alias: 
      (TABLEALIAS this: 
        (IDENTIFIER this: A, quoted: False)))))
```

GeorgeSittas:
Fixes #916

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
             "black",
             "duckdb",
             "isort",
-            "mypy",
+            "mypy>=0.990",
             "pandas",
             "pyspark",
             "python-dateutil",
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -725,6 +725,8 @@ class Tokenizer(metaclass=_Tokenizer):
     COMMENTS = ["--", ("/*", "*/")]
     KEYWORD_TRIE = None  # autofilled
 
+    IDENTIFIER_CAN_START_WITH_DIGIT = False
+
     __slots__ = (
         "sql",
         "size",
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -28,7 +28,7 @@
             "black",
             "duckdb",
             "isort",
-            "mypy",
+            "mypy>=0.990",
```
GeorgeSittas, line 31: This was changed because we have the following in `serde.py`

```python
Node = t.Union[t.List["Node"], exp.DataType.Type, exp.Expression, JSON]
```

Which seems to be supported after version `0.990` of mypy ([source](https://stackoverflow.com/a/53639221)).

tobymao (approved):

sqlglot/tokens.py
```diff
@@ -943,16 +945,23 @@ def _scan_number(self) -> None:
...
                 token_type = self.KEYWORDS.get(self.NUMERIC_LITERALS.get(literal))  # type: ignore
+
                 if token_type:
+                    self._add(TokenType.NUMBER, number_text)
                     self._add(TokenType.DCOLON, "::")
                     return self._add(token_type, literal)  # type: ignore
+                elif self.IDENTIFIER_CAN_START_WITH_DIGIT:
+                    return self._add(TokenType.VAR)
+
+                self._add(TokenType.NUMBER, number_text)
```
GeorgeSittas, line 964: This is so that we can correctly parse `1a` for example in duckdb as `1 AS a`.

PR #924, opened 2023-01-12
BigQuery Fixes

eakmanrq:
Main changes:
1. BigQuery does not allow precision on parameterized types when doing casts. Therefore added a transformer to remove these when casts are detected.
2. When doing an `UNNEST(...` (similar to `VALUES` but for BigQuery) BigQuery auto detects types from the first row of values. If one of those values are a null then it will potentially select the wrong type. Therefore updated values to accept column names and types and if types are provided then the first row is casted to the desired type to make sure type inference works correctly.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3785,7 +3785,7 @@ def values(
         types = list(columns.values())
         expressions[0].set(
             "expressions",
-            tuple(Cast(this=x, to=types[i]) for i, x in enumerate(expressions[0].expressions)),
+            [Cast(this=x, to=types[i]) for i, x in enumerate(expressions[0].expressions)],
         )
     return Values(
         expressions=expressions,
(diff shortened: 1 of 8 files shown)


sqlglot/expressions.py
```diff
@@ -3777,8 +3780,15 @@ def values(
         if columns
         else TableAlias(this=to_identifier(alias) if alias else None)
     )
+    expressions = [convert(tup) for tup in values]
+    if columns and isinstance(columns, dict):
+        types = list(columns.values())
+        expressions[0].set(
+            "expressions",
+            tuple(Cast(this=x, to=types[i]) for i, x in enumerate(expressions[0].expressions)),
```
tobymao, line 3788:
```suggestion
            [Cast(this=x, to=types[i]) for i, x in enumerate(expressions[0].expressions)],
```

sqlglot/transforms.py
```diff
@@ -82,6 +82,30 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def remove_precision_parameterized_types(expression: exp.Expression) -> exp.Expression:
+    """
+    Some dialects only allow the precision for parameterized types to be defined in the DDL and not in other expressions.
+    This transforms removes the precision from parameterized types in expressions.
+    """
+    expression = expression.copy()
```
tobymao, line 90: the copy is not needed, transform by default does copy

tobymao (approved):

PR #934, opened 2023-01-13
Update oracle.py

gfmartins:
I'm in doubt if it should be equal to postgresql... or mysql implementations
https://docs.oracle.com/cd/B12037_01/server.101/b10759/functions184.htm

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -389,7 +389,7 @@ def trim_sql(self, expression):
     remove_chars = self.sql(expression, "expression")
     collation = self.sql(expression, "collation")
 
-    # Use TRIM/LTRIM/RTRIM syntax if the expression isn't postgres-specific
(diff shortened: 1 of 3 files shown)


barakalon (commented):
Can you move this to [dialect.py](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/dialect.py) and have both this and [postgres.py](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/postgres.py) use it?

Also, a [test](https://github.com/tobymao/sqlglot/blob/main/tests/dialects/test_oracle.py) would be nice :)

GeorgeSittas:
Hey @gfmartins, any plans to finish this? Let me know if you need any help.

gfmartins:
I tried. (:

GeorgeSittas (approved):
Looks good. Can you also add some tests for Oracle? Check out `test_postgres.py` starting from line 178.

PR #940, opened 2023-01-16
Improve datatype build

eakmanrq:


setup.cfg
```diff
@@ -1,6 +1,7 @@
 [mypy]
 disallow_untyped_calls = False
 no_implicit_optional = True
+check-untyped-defs = True
```
eakmanrq, line 4: I upgraded to latest mypy and needed to add this

tobymao (approved):

sqlglot/expressions.py
```diff
@@ -2202,11 +2202,16 @@ class Type(AutoName):
...
-    def build(cls, dtype, **kwargs) -> DataType:
-        return DataType(
-            this=dtype if isinstance(dtype, DataType.Type) else DataType.Type[dtype.upper()],
-            **kwargs,
-        )
+    def build(cls, dtype: str | DataType.Type, **kwargs) -> DataType:
+        from sqlglot import parse_one
+
+        if isinstance(dtype, str):
+            data_type_exp = t.cast(DataType, parse_one(dtype, into=DataType))
```
GeorgeSittas, line 2209: What if `parse_one` fails here?

GeorgeSittas (approved):
LGTM.

sqlglot/expressions.py
eakmanrq: `cast` is a no-op call at runtime so it doesn't matter what `parse_one` actually does. It is a way to force type checking to recognize a value as a given type.

GeorgeSittas: Yep, I was referring to line 2214 because `data_type_exp.args` may crash if the parsing fails.

eakmanrq: Yeah I see. Addressed here: https://github.com/tobymao/sqlglot/pull/940/commits/8ee8308bcd4d578f09c7ff24cf80408e309db813

PR #941, opened 2023-01-16
Handle all types in builder

eakmanrq:
There are some types that don't get parsed to a data type that aren't handled correctly in the current implementation. An example of this would be the `UNKNOWN` type. Therefore this change has us first check if we have a direct match of the type name against the enums and if so we use that. If not we then parse in order to handle the more complex examples.


sqlglot/expressions.py
```diff
@@ -2206,7 +2206,11 @@ def build(cls, dtype: str | DataType.Type, **kwargs) -> DataType:
         from sqlglot import parse_one
 
         if isinstance(dtype, str):
-            data_type_exp = parse_one(dtype, into=DataType)
+            data_type_exp: t.Optional[Expression]
+            if dtype.upper() in cls.Type.__members__:
+                data_type_exp = DataType(this=DataType.Type[dtype.upper()])
+            else:
+                data_type_exp = parse_one(dtype, into=DataType)
```
tobymao, line 2213: the build method needs to take in a dialect for the parse

eakmanrq: Addressed: https://github.com/tobymao/sqlglot/pull/941/commits/ae089d1eaefe86f3f8414bff4967a3578a2571aa

tobymao (approved):

PR #942, opened 2023-01-16
Have core parser methods return Expression instead of Optional[Expression]

GeorgeSittas:


sqlglot/__init__.py
```diff
@@ -39,9 +39,7 @@
 schema = MappingSchema()
 
 
-def parse(
-    sql: str, read: t.Optional[str | Dialect] = None, **opts
-) -> t.List[t.Optional[Expression]]:
+def parse(sql: str, read: t.Optional[str | Dialect] = None, **opts) -> t.List[Expression]:
```
tobymao, line 42: why did you change parse? i thought you only were going to do parse_one

sqlglot/__init__.py
```diff
@@ -83,7 +83,10 @@ def parse_one(
     else:
         result = dialect.parse(sql, **opts)
 
-    return result[0] if result else None
+    if not result:
+        raise ParseError(f"Unable to parse {sql}")
```
tobymao, line 87: "No expression was parsed from '{sql}'"

tobymao (approved):

PR #958, opened 2023-01-17
Fix: accept some keywords as ID_VARS for BigQuery

### LINKED ISSUE #954: Using reserved words as identifiers
In Bigquery, various reserved words and builtin function names can be used as unquoted identifiers.

E.g. `WITH view AS (SELECT 1 AS x) SELECT * FROM view` is valid in BQ but raises a ParseError "Expected CTE to have alias".  Renaming the CTE to something else stops the error.

Another example is `WITH q AS (SELECT 'A' as a, 51 as b, 23 as c UNION ALL SELECT 'B', 77, 0) SELECT * FROM q UNPIVOT(values FOR quarter IN (b, c))` which fails with ParseError "Expecting FOR" because the unpivot of `values`; another identifier parses ok.

Another example is `SELECT CAST(time*1000 AS INT64) FROM (SELECT 0.1 AS time)` where `time` is used as a column name; sqlglot raises ParseError "Expected AS after CAST".  Naming the column something else stops the error.

GeorgeSittas:
This is a WIP because I haven't solved the issue related to the `time` identifier.

Fixes #954


sqlglot/dialects/bigquery.py
```diff
@@ -165,6 +165,14 @@ class Parser(parser.Parser):
             TokenType.TABLE,
         }
 
+        ID_VAR_TOKENS = {
+            *parser.Parser.ID_VAR_TOKENS,  # type: ignore
+            TokenType.VALUES,
+            TokenType.VIEW,
```
tobymao, line 171: why not put this in the base?

GeorgeSittas: `VALUES` can't be put in the base parser because in some places we have e.g. `VALUES (1, 2, 3)`, so it tries to parse it as aliases and fails. We can put `VIEW` in there, though. Do you think we should make this change?

tobymao: can other system uses view as an alia’s?

GeorgeSittas: Seems like postgres and duckdb allow it. Making the change in a sec.

GeorgeSittas:
One item to go for this PR:

```sql
SELECT CAST(time*1000 AS INT64) FROM (SELECT 0.1 AS time)
```

PR #971, opened 2023-01-19
Fixes #970

### LINKED ISSUE #970: SQL Server Procedures and Variables
Seems sqlglot cannot parse such stuff:

```sql
CREATE procedure [TRANSF].[SP_Merge_Sales_Real]

   @Loadid INT
  --,@YearMonth int=0
  ,@NumberOfRows INT = NULL OUT

AS
BEGIN
        SET XACT_ABORT ON;

        DECLARE     @DWH_DateCreated    DATETIME                = CONVERT(DATETIME, getdate(), 104);
        DECLARE     @DWH_DateModified   DATETIME                = CONVERT(DATETIME, getdate(), 104);
        DECLARE     @DWH_IdUserCreated  int                             = SUSER_ID (SYSTEM_USER);
        DECLARE         @DWH_IdUserModified int                         = SUSER_ID (SYSTEM_USER)

                         DECLARE @SalesAmountBefore float
                         SELECT @SalesAmountBefore=SUM(SalesAmount) FROM TRANSF.[Pre_Merge_Sales_Real] S

        --ALTER INDEX ALL on [TRANSF].[SP_Merge_Sales_Real] rebuild WITH (DATA_COMPRESSION = ROW)

--      DBCC SHRINKFILE(CORE_log, 1)

END
```

aersam:


GeorgeSittas:
This requires only `"DECLARE": TokenType.COMMAND"` in TSQL's `KEYWORDS` mapping, no need to make a new token.

tests/dialects/test_tsql.py
```diff
@@ -13,6 +13,30 @@ def test_tsql(self):
...
+        CREATE procedure [TRANSF].[SP_DO_STUFF]
+
+   @Loadid INT 
+  --,@YearMonth int=0
+  ,@NumberOfRows INT = NULL OUT
+AS 
+BEGIN 
+    PRINT 'hi'
+END
+        """)
```
GeorgeSittas, line 39: You should use `validate_all` or set the pretty parameter here; SQLGlot doesn't format by default, so your SQL will be transformed in a single line. Look at some other tests for reference and let me know if anything's unclear.

GeorgeSittas:
@aersam let me know if you need any help on this. The syntax for CREATE PROCEDURE might not be very straightforward to capture unless you're familiar with the codebase.

aersam:
> @aersam let me know if you need any help on this. The syntax for CREATE PROCEDURE might not be very straightforward to capture unless you're familiar with the codebase.

It seems to be quite difficult given the parser splits by semicolon. Can yoh have a look at my approach? I basically think there is need for kind of a sub-parser. However it does not work yet...

GeorgeSittas:
> It seems to be quite difficult given the parser splits by semicolon. Can yoh have a look at my approach? I basically think there is need for kind of a sub-parser. However it does not work yet...

Your approach feels unusually complicated. I'll try to simplify things and ping you if/when I make a PR.

aersam:
> > It seems to be quite difficult given the parser splits by semicolon. Can yoh have a look at my approach? I basically think there is need for kind of a sub-parser. However it does not work yet...
> 
> Your approach feels unusually complicated. I'll try to simplify things and ping you if/when I make a PR.

Great! I prefer simple code 🙂

PR #975, opened 2023-01-19
Add support for alternative CTE syntax (clickhouse)

### LINKED ISSUE #973: ClickHouse WITH statement
According to ClickHouse [docs](https://clickhouse.com/docs/en/sql-reference/statements/select/with/), `WITH` statement looks like `WITH <expression> AS <identifier>`. `sqlglot` expects that identifier precedes expression. So if I'll try to parse SQL like:
```sql
WITH '2019-08-01 15:23:00' as ts_upper_bound
SELECT *
FROM hits
WHERE
    EventDate = toDate(ts_upper_bound) AND
    EventTime <= ts_upper_bound
```
I have exception:
```python
sqlglot.errors.ParseError: Expected CTE to have alias. Line 2, Col: 6.
```
One more problem is SQL generation. If I try to generate something like:
```sql
WITH trim('   A   ') AS val SELECT val
```
I expect to write something like:
```python
select("val").with_("val", as_="trim('   a   ')").sql(dialect=Dialects.CLICKHOUSE)
```
but receive this SQL, which is invalid in ClickHouse:
```sql
WITH val AS (TRIM('   a   ')) SELECT val
```
If I switch arguments of `with_` method, I'll almost achieve expected:
```python
select("val").with_(parse_one("trim('   a   ')"), as_="val", dialect=Dialects.CLICKHOUSE).sql(dialect=Dialects.CLICKHOUSE)
```
But resulting SQL is also invalid due to parentheses around `val`:
```sql
WITH TRIM('   a   ') AS (val) SELECT val
```
I guess ClickHouse dialect should have specified behaviour for `expressions.CTE` parsing/generation.

GeorgeSittas:
Fixes #973 

Reference: https://clickhouse.com/docs/en/sql-reference/statements/select/with/

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1357,10 +1357,10 @@ class Parser(metaclass=_Parser):
             or not alias.this
             or not self._match_pair(TokenType.ALIAS, TokenType.L_PAREN, advance=False)
         ):
-            statement: t.Optional[exp.Expression]
-
             if self._match(TokenType.ALIAS):
-                statement = self.expression(exp.Alias, this=alias, alias=self._parse_id_var())
+                statement: t.Optional[exp.Expression] = self.expression(
+                    exp.Alias, this=alias, alias=self._parse_id_var()
+                )
             else:
                 statement = self._parse_statement()
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1350,24 +1350,8 @@ class Parser(metaclass=_Parser):
 
     def _parse_cte(self) -> exp.Expression:
         alias = self._parse_table_alias()
-
-        # Alternative Clickhouse syntax: https://clickhouse.com/docs/en/sql-reference/statements/select/with/
-        if (
-            not alias
-            or not alias.this
-            or not self._match_pair(TokenType.ALIAS, TokenType.L_PAREN, advance=False)
-        ):
-            if self._match(TokenType.ALIAS):
-                statement: t.Optional[exp.Expression] = self.expression(
-                    exp.Alias, this=alias, alias=self._parse_id_var()
-                )
-            else:
-                statement = self._parse_statement()
-
-            if not isinstance(statement, exp.Alias):
-                self.raise_error("Expected CTE to have alias")
-
-            return self.expression(exp.CTE, this=statement, alias=statement and statement.alias)
+        if not alias or not alias.this:
+            self.raise_error("Expected CTE to have alias")
 
         if not self._match(TokenType.ALIAS):
             self.raise_error("Expected AS in CTE")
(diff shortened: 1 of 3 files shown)


GeorgeSittas (commented):
The fix feels a bit complicated. Let me know if I missed an easier way to implement this.

sqlglot/parser.py
```diff
@@ -1350,8 +1350,24 @@ def _parse_with(self, skip_with_token: bool = False) -> t.Optional[exp.Expressio
...
     def _parse_cte(self) -> exp.Expression:
         alias = self._parse_table_alias()
-        if not alias or not alias.this:
-            self.raise_error("Expected CTE to have alias")
+
+        # Alternative Clickhouse syntax: https://clickhouse.com/docs/en/sql-reference/statements/select/with/
+        if (
+            not alias
+            or not alias.this
+            or not self._match_pair(TokenType.ALIAS, TokenType.L_PAREN, advance=False)
```
GeorgeSittas, line 1358: The `_match_pair` check here is so that we can parse `WITH (SELECT 1) AS foo`.

sqlglot/parser.py
```diff
@@ -1350,8 +1350,24 @@ def _parse_with(self, skip_with_token: bool = False) -> t.Optional[exp.Expressio
...
+        ):
+            statement: t.Optional[exp.Expression]
```
tobymao, line 1360: you don't need this initializer, put it inline 1363, statement: t.Optioanl = self.expression

sqlglot/parser.py
```diff
@@ -1350,8 +1350,24 @@ def _parse_with(self, skip_with_token: bool = False) -> t.Optional[exp.Expressio
...
+
+            if self._match(TokenType.ALIAS):
+                statement = self.expression(exp.Alias, this=alias, alias=self._parse_id_var())
+            else:
+                statement = self._parse_statement()
+
+            if not isinstance(statement, exp.Alias):
+                self.raise_error("Expected CTE to have alias")
+
+            return self.expression(exp.CTE, this=statement, alias=statement and statement.alias)
```
tobymao, line 1370: the statement is no longer an alias,  i think probably for CTE, we should add a new field to represent this as an expression right? this is kind of a hack

sqlglot/parser.py
```diff
@@ -1370,9 +1386,13 @@ def _parse_table_alias(
             any_token=any_token, tokens=alias_tokens or self.TABLE_ALIAS_TOKENS
         )
 
+        index = self._index
         if self._match(TokenType.L_PAREN):
             columns = self._parse_csv(lambda: self._parse_column_def(self._parse_id_var()))
-            self._match_r_paren()
+            self._match(TokenType.R_PAREN)
+
+            if not columns:
```
tobymao, line 1394: what case is this solving?

sqlglot/parser.py
tobymao: not sure i get it

sqlglot/parser.py
GeorgeSittas: Well the syntax is always something like `<foo> AS <bar>`, so the idea was that:

- For anything we're already supporting, the behavior won't change.
- For clickhouse-specific CTEs, we'll parse the `<foo> AS <bar>` as an `exp.Alias` node and rely on that.

sqlglot/parser.py
GeorgeSittas: Sorry, my bad. I got confused with the examples. What I meant to say is that it solves the following form:

```sql
WITH sum(bytes) AS foo SELECT foo FROM system.parts
```

Without this pair match check, `sum(bytes)` will be parsed as a table alias right above this branch and the parsing would fail later.

PR #977, opened 2023-01-19
Add Teradata dialect

treysp:
This pull request:

- Adds a Teradata dialect
- Parses Teradata's `TRANSLATE` function for casting to a new character set as an `exp.Cast`
- Parses Teradata's `UPDATE FROM` syntax (`FROM` before `SET`, where other dialects are `SET` before `FROM`)


tobymao (approved):
great pr, thanks

PR #978, opened 2023-01-19
Redshift Improvements

eakmanrq:
Key changes:
1. Redshift doesn't support values outside of an `INSERT INTO` expression. Therefore we convert this to a union and have a warning about the poor performance with unioning a lot of values
2. This adds support for `ALTER TABLE ... RENAME TO...`. Some engines support a more direct `RENAME TABLE` syntax but I didn't tackle that here
3. Redshift puts the `LIKE` expression in paranthesis and the rest at root. I handle this by having `WITH` properties drop list (`WITH` isn't supported by Redshift).
4. Redshift converts `TEXT` to `VARCHAR(255)` which is strange since the max size is much larger. So we override this to make `TEXT` be `VARCHAR(max)`. That is generally what is meant by other engines when they say `TEXT`/`STRING` without precision.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -75,13 +75,16 @@ class Redshift(Postgres):
             """
             if not isinstance(expression.unnest().parent, exp.From):
                 return super().values_sql(expression)
-            rows = [tuple_exp.expressions for tuple_exp in expression.find_all(exp.Tuple)]
+            rows = [tuple_exp.expressions for tuple_exp in expression.expressions]
             selects = []
             for i, row in enumerate(rows):
-                columns = [
-                    exp.alias_(value, column_name) if i == 0 else value
-                    for value, column_name in zip(row, expression.args["alias"].args["columns"])
-                ]
+                if i == 0:
+                    columns = [
+                        exp.alias_(value, column_name)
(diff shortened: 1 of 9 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -60,3 +64,55 @@ class Generator(Postgres.Generator):
...
+            VARCHAR of max length which is `VARCHAR(max)` in Redshift. Therefore if we get a `TEXT` data type
+            without precision we convert it to `VARCHAR(max)` and if it does have precision then we just convert
+            `TEXT` to `VARCHAR`.
+            """
+            if expression.this == exp.DataType.Type.TEXT:
+                expression = expression.copy()
+                expression.set("this", exp.DataType.Type.VARCHAR)
+                precision = expression.args.get("expressions")
+                if not precision:
+                    expression.set("expressions", [exp.to_identifier("max")])
```
tobymao, line 117: expression.append("expressions", exp.to_identifier("max"))

tobymao: why is this even needed? why not just do a TYPE_MAPPING?

TYPE.FLOAT: "VARCHAR(MAX)"

eakmanrq: Decided to not do the `TYPE_MAPPING` change

tobymao (approved):

PR #986, opened 2023-01-20
Add no schema binding

eakmanrq:
There is a way to define views in Redshift that have `WITH NO SCHEMA BINDING` and that "unbinds" the view to underlying table(s) so that allows you to alter/drop the underlying table(s).

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -480,10 +480,13 @@ class Generator:
                 materialized,
             )
         )
+        no_schema_binding = (
+            " WITH NO SCHEMA BINDING" if expression.args.get("no_schema_binding") else ""
+        )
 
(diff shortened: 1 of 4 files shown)


sqlglot/generator.py
```diff
@@ -480,10 +480,13 @@ def create_sql(self, expression: exp.Create) -> str:
                 materialized,
             )
         )
+        no_schema_binding = (
+            " WITH NO SCHEMA BINDING" if expression.args.get("no_schema_binding") else ""
+        )
 
         post_expression_modifiers = "".join((data, statistics, no_primary_index))
```
tobymao, line 487: does it go after indices? you could put it in post_expression_modifiers

tobymao (approved):

GeorgeSittas (approved):

PR #990, opened 2023-01-20
Fix variable/identifier tokenization for TSQL to include @

GeorgeSittas:
This PR changes the "var" tokenizing logic for TSQL, so that the `@` is considered part of the token, instead of tokenizing it as a "parameter". This felt more natural & robust, even though we could already parse names prefixed by @ before.

Also: I don't think that adding the `PARAMETER` token to the `prefix_tokens` parameter of `_parse_id_var` would solve the problem, because there might be places in the codebase where we call `_parse_primary` first and then `_parse_id_var`, meaning that in these cases we'd parse the input as a parameter instead of an identifier.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -263,7 +263,7 @@ class TSQL(Dialect):
             "XML": TokenType.XML,
         }
 
-        # TSQL allows @ to appear as a variable/identifier prefix
+        # TSQL allows @, # to appear as a variable/identifier prefix
         SINGLE_TOKENS = tokens.Tokenizer.SINGLE_TOKENS.copy()
         SINGLE_TOKENS.pop("@")
         SINGLE_TOKENS.pop("#")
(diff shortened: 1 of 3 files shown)


GeorgeSittas:
Resolving conflicts in a sec

sqlglot/dialects/tsql.py
```diff
@@ -263,6 +263,10 @@ class Tokenizer(tokens.Tokenizer):
             "XML": TokenType.XML,
         }
 
+        # TSQL allows @ to appear as a variable/identifier prefix
```
tobymao, line 266: well if you do that, doesn't that mean you should delete table_prefix_tokens?

GeorgeSittas: Good catch, I can `pop` the `HASH` token too and get rid of it completely.

GeorgeSittas: Done

tobymao (approved):

PR #995, opened 2023-01-20
Add support for TSQL 'FOR SYSTEM_TIME' clause

### LINKED ISSUE #993: Parsing of t-sql FOR SYSTEM_TIME
T-SQL FOR SYSTEM_TIME Syntax is not supported
https://learn.microsoft.com/en-us/sql/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table?view=sql-server-ver16

Spark knows something similar: https://issues.apache.org/jira/browse/SPARK-37219

GeorgeSittas:
Fixes #993 

Reference: https://learn.microsoft.com/en-us/sql/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table?view=sql-server-ver16

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -304,10 +304,8 @@ class TSQL(Dialect):
             if not self._match_text_seq("FOR", "SYSTEM_TIME"):
                 return None
 
-            system_time: t.Optional[exp.Expression] = None
-
             if self._match(TokenType.ASOF):
-                system_time = self.expression(
+                system_time: t.Optional[exp.Expression] = self.expression(
                     exp.SystemTime, this=self._parse_bitwise(), kind="AS OF"
                 )
             elif self._match_set((TokenType.FROM, TokenType.BETWEEN)):
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -326,6 +326,7 @@ class TSQL(Dialect):
             elif self._match(TokenType.ALL):
                 system_time = self.expression(exp.SystemTime, kind="ALL")
             else:
+            	system_time = None
                 self.raise_error("Unable to parse FOR SYSTEM_TIME clause")
 
             return system_time
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -304,7 +304,7 @@ class TSQL(Dialect):
                 return None
 
             if self._match_text_seq("AS", "OF"):
-                system_time: t.Optional[exp.Expression] = self.expression(
+                system_time = self.expression(
                     exp.SystemTime, this=self._parse_bitwise(), kind="AS OF"
                 )
             elif self._match_set((TokenType.FROM, TokenType.BETWEEN)):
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -298,23 +300,67 @@ class Parser(parser.Parser):
             DataType.Type.NCHAR,
         }
 
-        def _parse_convert(self, strict):
+        def _parse_system_time(self) -> t.Optional[exp.Expression]:
+            if not self._match_text_seq("FOR", "SYSTEM_TIME"):
+                return None
+
+            system_time: t.Optional[exp.Expression] = None
```
tobymao, line 307: you don't need to do this declaration, just put it on line 310 inline

sqlglot/dialects/tsql.py
```diff
@@ -356,3 +402,47 @@ class Generator(generator.Generator):
...
+            end = self.sql(expression, "expression")
+            if kind == "FROM":
+                return f"FOR SYSTEM_TIME FROM {start} TO {end}"
+            if kind == "BETWEEN":
+                return f"FOR SYSTEM_TIME BETWEEN {start} AND {end}"
+
+            return f"FOR SYSTEM_TIME CONTAINED IN ({start}, {end})"
+
+        def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
+            table = ".".join(
```
tobymao, line 422: exp.table_name()

GeorgeSittas: This was left over, I'm cleaning it up now

sqlglot/dialects/tsql.py
```diff
@@ -242,6 +243,7 @@ class Tokenizer(tokens.Tokenizer):
 
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
+            "AS OF": TokenType.ASOF,
```
tobymao, line 246: asof is a common keyword, i think we can put this in the base?

sqlglot/parser.py
```diff
@@ -1558,6 +1558,27 @@ def _parse_create_table_index(self) -> t.Optional[exp.Expression]:
             amp=amp,
         )
 
+    def _parse_table_parts(self, schema: bool = False) -> exp.Expression:
```
GeorgeSittas, line 1561: I factored this out so that I could override it in TSQL and parse the `FOR SYSTEM_TIME` before the alias.

sqlglot/dialects/tsql.py
GeorgeSittas: Should we just do "AS OF" or "ASOF" too? (the latter is used in clickhouse).

tobymao: only as of, asof is not safe

tobymao: hm maybe not since you could do "select 1 as of"

tobymao: yea forget it

GeorgeSittas: Lol right, missed this too. Reverting the last commit then

GeorgeSittas: it's probably unsafe to do this in tsql too then, I'll just `match_text_seq`

tobymao (approved):

PR #1003, opened 2023-01-23
Add support for the ADD CONSTRAINT clause in ALTER TABLE

### LINKED ISSUE #997: Support for Check Constraint Transpiling
Both, T-SQL and Spark/Databricks support check constraints.
I had hope to be able to transpile something like this:

```t-sql
ALTER TABLE [schema].[tablename]
    ADD CONSTRAINT [CHK_Name] CHECK ([IdDwh] IS NOT NULL AND [IdDwh]<>(0));
```

Seems not to be supported?

GeorgeSittas:
Fixes #997 

Reference: https://docs.databricks.com/sql/language-manual/sql-ref-syntax-ddl-alter-table-add-constraint.html


sqlglot/parser.py
```diff
@@ -3028,8 +3089,14 @@ def _parse_alter(self) -> t.Optional[exp.Expression]:
...
-        if self._match_text_seq("ADD", advance=False):
-            actions = self._parse_csv(self._parse_add_column)
+
+        index = self._index
+        if self._match_text_seq("ADD"):
+            if self._match_set(self.ADD_CONSTRAINT_TOKENS):
+                actions = self._parse_csv(self._parse_add_constraint)
+            else:
+                self._retreat(index)
+                actions = self._parse_csv(self._parse_add_column)
```
GeorgeSittas, line 3099: This retreat is useful because there are some dialects where you can have a csv list of `ADD COLUMN ....`, so `_parse_add_column` first matches that pair of tokens to enable the use of `_parse_csv`.

Issue #999, opened 2023-01-23
Optimizer | Nonstandard behavior ORDER BY


gfmartins:
"Nonstandard" sqlglot behavior in ORDER BY:

```
print(
    optimize(
        sqlglot.parse_one("""
                             select
                                sum(A) as sumA,
                                CAST(number as INT)
                            from x
                            order by
                                number
        """ ),
        schema={"x": {"A": "INT",  "number": "FLOAT"}}
    ).sql(pretty=True)
)
```

**optimizes to:**
```
SELECT
  SUM("x"."a") AS "sumA",
  CAST("x"."number" AS INT) AS "_col_1"
FROM "x" AS "x"
ORDER BY
  "number"   <-- Expecting it to be "x"."number"
```

=====================================

```
print(
    optimize(
        sqlglot.parse_one("""
                            select
                                sum(A) as sumA,
                                CAST(date as DATETIME)
                            from x
                            order by
                                date
        """),
        schema={"x": {"A": "INT",  "date": "DATETIME"}}
    ).sql(pretty=True)
)
```

**optimizes to:**
```
SELECT
  SUM("x"."a") AS "sumA",
  "x"."date" AS "_col_1"
FROM "x" AS "x"
ORDER BY
  "date"  <---- Expecting it to be "x"."date" OR "_col_1"
```

tobymao:
this is by design. different engines do different things and so sqlglot chooses this form as it seems to be the most common / safest

barakalon:
This seems like a bug to me. It should only leave it unqualified if there is an select alias with that name.

PR #1005, opened 2023-01-23
Stricter named_selects

barakalon:
fixes https://github.com/tobymao/sqlglot/issues/999

`Expression.alias_or_name` is a bit too lenient in many cases:

```python
parse_one("1 + 2").alias_or_name
# '1'
```

So this introduces yet another concept - "output_name". 
I'm not crazy about adding a new method. But this seems safer than mucking with `Expression.alias_or_name`. Happy to hear opinions here.

@tobymao @GeorgeSittas

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2175,15 +2175,9 @@ class Placeholder(Expression):
 class Null(Condition):
     arg_types: t.Dict[str, t.Any] = {}
 
-    @property
-    def output_name(self):
-        return self.name
-
 
 class Boolean(Condition):
-    @property
-    def output_name(self):
-        return self.name
+    pass
 
 
 class DataType(Expression):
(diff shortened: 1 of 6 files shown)


barakalon:
I'm also not sure if the output_name of `CAST(a AS INT)` should be `a`, as it is now.

tobymao:
do we really need to have output_name?

barakalon:
Either that or:
1. We change the default behavior of `Expression.name`
2. We move this logic into the optimizer... but that would introduce even more conflicting concepts.

barakalon:
Or maybe we should need to fix Expression.name for a few expression types, e.g. Binary?

GeorgeSittas:
I think the issue still remains even with these changes, I can repro them in this branch locally. Am I missing something?

barakalon:
@GeorgeSittas notice that CAST now has a "number" alias, so it's correct.

This PR now gives us control here. If we prefer, we can remove "CAST"'s output name.

tobymao (approved):

GeorgeSittas:
> @GeorgeSittas notice that CAST now has a "number" alias, so it's correct.

Totally missed this, thanks for pointing out.

PR #1008, opened 2023-01-24
Fix TSQL UDF BEGIN parsing

GeorgeSittas:
Addresses some leftover issues in #970

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -877,10 +877,19 @@ class Parser(metaclass=_Parser):
             return self.STATEMENT_PARSERS[self._prev.token_type](self)
 
         if self._match_set(Tokenizer.COMMANDS):
+            command_kind = self._prev.text
+            command_body = self._curr.text if self._curr else ""
+
+            while self._curr:
+                self._advance()
+                if not self._curr:
+                    break
+                command_body += " " + self._curr.text
+
             return self.expression(
                 exp.Command,
-                this=self._prev.text,
-                expression=self._parse_string(),
+                this=command_kind,
+                expression=None if not command_body else exp.Literal.string(command_body),
             )
 
         expression = self._parse_expression()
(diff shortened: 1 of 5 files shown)


sqlglot/parser.py
```diff
@@ -877,10 +877,19 @@ def _parse_statement(self) -> t.Optional[exp.Expression]:
             return self.STATEMENT_PARSERS[self._prev.token_type](self)
 
         if self._match_set(Tokenizer.COMMANDS):
+            command_kind = self._prev.text
+            command_body = self._curr.text if self._curr else ""
```
tobymao, line 881: why is this part necessary?

GeorgeSittas: In the tokenizer, line 847, we have this condition for tokenizing whatever comes after a command as a string:
```python
        if token_type in self.COMMANDS and (
            len(self.tokens) == 1 or self.tokens[-2].token_type == TokenType.SEMICOLON
        ):
```
This works if we have for example a single command to parse, like `SET foo bla bar`, because it will tokenize it as `<COMMAND, text: SET> <STRING, text: foo bla bar>`. The problem with the UDF was that the above condition would never be true because the command appears in the middle of a statement, so we'd tokenize the above like `<COMMAND, text: SET> <VAR, text: foo> <VAR, text: bla> <VAR, text: bar>`, which is harder to parse. The code I added makes sure we consume all remaining tokens and use their texts to construct a string, like it's done for the simple case in the tokenizer.

GeorgeSittas: The while loop I added should be a no-op for the general case, because `self._curr` will be `None`.

sqlglot/tokens.py
```diff
@@ -844,8 +844,11 @@ def _add(self, token_type: TokenType, text: t.Optional[str] = None) -> None:
         )
         self._comments = []
 
+        # If we have either a semicolon or a begin token before the command's token, we'll parse
+        # whatever follows the command's token as a string
         if token_type in self.COMMANDS and (
-            len(self.tokens) == 1 or self.tokens[-2].token_type == TokenType.SEMICOLON
+            len(self.tokens) == 1
+            or self.tokens[-2].token_type in (TokenType.SEMICOLON, TokenType.BEGIN)
```
tobymao, line 851: make this a constant and a set

tobymao (approved):

sqlglot/tokens.py
GeorgeSittas: done

PR #1015, opened 2023-01-26
Cleaned up some docs

GeorgeSittas:
I removed the `expressions.md` document in favor of improving / adding some docstrings in `expressions.py`.

I think that the existing documentation (docstrings) on the helper functions is not too bad, but it may not be easy enough to find looking at the docs. Any suggestion on how we can improve this?

Also, it feels like there are several commenting styles in the codebase, so it can feel a bit inconsistent. Should we establish a style and stick with it from now on?


tobymao (approved):

sqlglot/dataframe/README.md
```diff
@@ -7,24 +7,24 @@ Currently many of the common operations are covered and more functionality will
...
-      * Ex: `StructType([StructField('cola', StringType()), StructField('colb', IntegerType())])`
-    * A string of names and types similar to what is supported in `createDataFrame`
-      * Ex: `cola: STRING, colb: INT`
-    * [Not Recommended] A list of string column names without type
-      * Ex: ['cola', 'colb']
-      * The lack of types may limit functionality in future releases
-  * See [Registering Custom Schema](#registering-custom-schema-class) for information on how to skip this step if the information is stored externally
-* Add `.sql(pretty=True)` to your final DataFrame command to return a list of sql statements to run that command
+    * Dictionary where the keys are column names and values are string of the Spark SQL type name.
+      * Ex: `{'cola': 'string', 'colb': 'int'}`.
```
barakalon, line 15: Things that aren't sentences don't periods

sqlglot/__init__.py
```diff
@@ -35,8 +36,10 @@
 __version__ = "10.5.6"
 
 pretty = False
+"""Whether to format generated SQL by default."""
```
barakalon, line 39: is this a pdocs thing? Docstrings under a variable?

barakalon (approved):

sqlglot/dataframe/README.md
GeorgeSittas: Wasn't aware, thanks for clarifying.

sqlglot/__init__.py
GeorgeSittas: Yep, it makes the variable appear in the API documentation:

Screenshot 2023-01-26 at 6 53 16 PM

PR #1017, opened 2023-01-26
Add support for Snowflake EXCLUDE & RENAME keywords

griffatrasgo:
This PR adds support for `SELECT * EXCLUDE` and `SELECT * RENAME` [Snowflake syntax](https://docs.snowflake.com/en/sql-reference/sql/select.html#syntax).

Example SQL that breaks without this change:
```

breaking_sql = """
SELECT DISTINCT
  TBL1.* EXCLUDE DATE_FIELD, ANOTHER_FIELD RENAME TBL1.BAD_FIELD_NAME AS GOOD_FIELD_NAME, C AS D
  TBL2.* EXCLUDE NOT_ANOTHER_FIELD RENAME A AS B
FROM DATABASE.SCHEMA.TABLE TBL1
JOIN DB.SC.TBL AS TBL2 ON TBL1.ID = TBL2.ID
WHERE INT_FIELD = 55
LIMIT 10
"""
for select in parse_one(breaking_sql, read="snowflake").find_all(exp.Select):
    for projection in select.expressions:
        print(projection.alias_or_name)
```

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1551,6 +1551,8 @@ class Select(Subqueryable):
         "expressions": False,
         "hint": False,
         "distinct": False,
+        "excludes": False,
+        "renames": False,
         "into": False,
         "from": False,
         **QUERY_MODIFIERS,
@@ -2036,6 +2038,51 @@ class Select(Subqueryable):
         """
         instance = _maybe_copy(self, copy)
         instance.set("distinct", Distinct() if distinct else None)
+        return instance
+
+    def exclude(self, columns=list, copy=True) -> Select:
+        """
+        Set the EXCLUDE expression.
+
+        Example:
+            >>> select("*").exclude(["x", "y"]).from_("tbl").sql()
+            'SELECT * EXCLUDE x, y FROM tbl'
+
+        Args:
+            columns (list[str]): columns to exclude from the SELECT * statement
+            copy (bool): if `False`, modify this expression instance in-place.
+
+        Returns:
+            Select: the modified expression.
+        """
+        if not isinstance(columns, list):
+            raise ValueError(f"expecting columns to be a list, not {type(columns)}")
+        instance = _maybe_copy(self, copy)
+        instance.set("exclude", Exclude(expressions=columns))
+        return instance
+
+    def rename(self, columns=dict, copy=True) -> Select:
+        """
+        Set the RENAME expression.
+
+        Example:
+            >>> select("*").rename({"x", "y"}).from_("tbl").sql()
+            'SELECT * EXCLUDE x, y FROM tbl'
+
+        Args:
+            columns (dict[str, str]): columns to alias in the SELECT * statement
+            copy (bool): if `False`, modify this expression instance in-place.
+
+        Returns:
+            Select: the modified expression.
+        """
+        if not isinstance(columns, dict):
+            raise ValueError(f"expecting columns to be a dict, not {type(columns)}")
+        instance = _maybe_copy(self, copy)
+        instance.set(
+            "rename",
+            Rename(expressions=[f"{k} AS {v}" for k, v in columns.items()]),
+        )
         return instance
 
     def ctas(self, table, properties=None, dialect=None, copy=True, **opts) -> Create:
@@ -2812,6 +2859,14 @@ class LastDateOfMonth(Func):
 
 class Extract(Func):
     arg_types = {"this": True, "expression": True}
+
+
+class Exclude(Expression):
+    arg_types = {"expressions": False}
+
+
+class Rename(Expression):
+    arg_types = {"expressions": False}
 
 
 class TimestampAdd(Func, TimeUnit):
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -589,6 +589,12 @@ class Generator:
     def except_op(self, expression: exp.Except) -> str:
         return f"EXCEPT{'' if expression.args.get('distinct') else ' ALL'}"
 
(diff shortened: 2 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
         instance.set("distinct", Distinct() if distinct else None)
         return instance
 
+    def exclude(self, columns=list, copy=True) -> Select:
```
tobymao, line 2043:
```suggestion
    def exclude(self, columns: list, copy: bool=True) -> Select:
```

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+        Returns:
+            Select: the modified expression.
+        """
+        if not isinstance(columns, list):
+            raise ValueError(f"expecting columns to be a list, not {type(columns)}")
+        instance = _maybe_copy(self, copy)
+        instance.set("exclude", Exclude(expressions=columns))
+        return instance
+
+    def rename(self, columns=dict, copy=True) -> Select:
```
tobymao, line 2064:
```suggestion
    def rename(self, columns=t.Dict[str, str], copy=True) -> Select:
```

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+    def exclude(self, columns=list, copy=True) -> Select:
+        """
+        Set the EXCLUDE expression.
+
+        Example:
+            >>> select("*").exclude(["x", "y"]).from_("tbl").sql()
+            'SELECT * EXCLUDE x, y FROM tbl'
+
+        Args:
+            columns (list[str]): columns to exclude from the SELECT * statement
```
tobymao, line 2052:
```suggestion
            columns: columns to exclude from the SELECT * statement
```

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+            copy (bool): if `False`, modify this expression instance in-place.
```
tobymao, line 2053:
```suggestion
            copy: if `False`, modify this expression instance in-place.
```

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+
+        Returns:
+            Select: the modified expression.
+        """
+        if not isinstance(columns, list):
```
tobymao, line 2058: delete this section

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+            raise ValueError(f"expecting columns to be a list, not {type(columns)}")
+        instance = _maybe_copy(self, copy)
+        instance.set("exclude", Exclude(expressions=columns))
```
tobymao, line 2061:
```suggestion
        instance.set("exclude", Exclude(expressions=[to_identifier(column) for column in columns]))
```

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+        return instance
+
+    def rename(self, columns=dict, copy=True) -> Select:
+        """
+        Set the RENAME expression.
+
+        Example:
+            >>> select("*").rename({"x", "y"}).from_("tbl").sql()
```
tobymao, line 2069: example is wrong

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+            'SELECT * EXCLUDE x, y FROM tbl'
+
+        Args:
+            columns (dict[str, str]): columns to alias in the SELECT * statement
```
tobymao, line 2073: remove types

sqlglot/expressions.py
```diff
@@ -2038,6 +2040,51 @@ def distinct(self, distinct=True, copy=True) -> Select:
...
+            copy (bool): if `False`, modify this expression instance in-place.
+
+        Returns:
+            Select: the modified expression.
+        """
+        if not isinstance(columns, dict):
```
tobymao, line 2079: delete this

sqlglot/expressions.py
```diff
@@ -2814,6 +2861,14 @@ class Extract(Func):
     arg_types = {"this": True, "expression": True}
 
 
+class Exclude(Expression):
+    arg_types = {"expressions": False}
```
tobymao, line 2865: should be True for both

sqlglot/generator.py
```diff
@@ -589,6 +589,12 @@ def except_sql(self, expression: exp.Except) -> str:
     def except_op(self, expression: exp.Except) -> str:
         return f"EXCEPT{'' if expression.args.get('distinct') else ' ALL'}"
 
+    def exclude_sql(self, expression: exp.Exclude) -> str:
+        return f" EXCLUDE {', '.join(expression.args.get('expressions'))}"
```
tobymao, line 593: why do you have a space in front? not sure this is correct

tobymao, line 593: you don't need a get since we should always have expressions for this and rename

tobymao, line 593: you should do self.expressions(expression) instead of manual join

griffatrasgo:
Can be achieved much simpler with Keyword mapping. Will open a separate PR

PR #1018, opened 2023-01-26
Implement func builder, refactor expression validation, fix bug

GeorgeSittas:
- Bug fixed: due to inheritance, `_sql_names` wasn't updated properly (e.g. `TryCast.sql_names` resulted in `["CAST"]`).
- Created a `validate` method in `Expression`, as it felt more natural. This led to a refactor of `validate_expression`.
- Created a helper function called `func`, which can be used for programmatically building function expressions.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4289,9 +4289,8 @@ def func(name: str, *args, dialect: t.Optional[Dialect | str] = None, **kwargs) 
     if from_args_list:
         function = from_args_list(args) if args else from_args_list.__self__(**kwargs)  # type: ignore
     else:
-        function = (
-            Anonymous(this=name, expressions=args) if args else Anonymous(this=name, **kwargs)
-        )
+        kwargs = kwargs or {"expressions": args}
+        function = Anonymous(this=name, **kwargs)
 
     for error_message in function.error_messages(args):
         raise ValueError(error_message)
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -535,6 +535,35 @@ class Expression(metaclass=_Expression):
         assert isinstance(self, type_)
         return self
 
+    def validate(self, args: t.Optional[t.List] = None) -> t.Generator[str, None, None]:
+        """
+        Checks if this expression is valid, making sure that all its mandatory arguments are set.
+
+        Args:
+            args: an optional list of items that was used to instantiate the expression, if it's a Func.
+
+        Yields:
+            Error messages for all possible errors that were found.
+        """
+        for k in self.args:
+            if k not in self.arg_types:
+                yield f"Unexpected keyword: '{k}' for {self.__class__}"
+        for k, mandatory in self.arg_types.items():
+            v = self.args.get(k)
+            if mandatory and (v is None or (isinstance(v, list) and not v)):
+                yield f"Required keyword: '{k}' missing for {self.__class__}"
+
+        if (
+            args
+            and isinstance(self, Func)
+            and len(args) > len(self.arg_types)
+            and not self.is_var_len_args
+        ):
+            yield (
+                f"The number of provided arguments ({len(args)}) is greater than "
+                f"the maximum number of supported arguments ({len(self.arg_types)})"
+            )
+
     def dump(self):
         """
         Dump this Expression to a JSON-serializable dict.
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -535,6 +535,35 @@ def assert_is(self, type_):
         assert isinstance(self, type_)
         return self
 
+    def validate(self, args: t.Optional[t.List] = None) -> t.Generator[str, None, None]:
```
tobymao, line 538: why is args optional?

tobymao: don't yield, just return a list of error messages?

GeorgeSittas: Because we might've instantiated a function with kwargs, or the args used @ instantiation time might not be known at all @ validation time.

sqlglot/expressions.py
```diff
@@ -4220,6 +4246,51 @@ def _replace_placeholders(node, args, **kwargs):
...
+def func(name: str, *args, **kwargs) -> Func:
+    """
+    Returns a Func expression.
+
+    Examples:
+        >>> abs = func("abs", this=Literal.number(5))
+        >>> abs.sql()
+        'ABS(5)'
+        >>>
+        >>> cast = func("cast", this=Literal.number(5), to=DataType.build("DOUBLE"))
```
tobymao, line 4258: you should be able to just pass this=5,

sqlglot/expressions.py
```diff
@@ -4220,6 +4246,51 @@ def _replace_placeholders(node, args, **kwargs):
...
+    Returns:
+        An instance of the function of interest, or an anonymous function, if `name` doesn't
+        correspond to an existing `sqlglot.expressions.Func` class.
+    """
+    if args and kwargs:
+        raise ValueError("Can't use both args and kwargs to instantiate a function.")
+
+    from sqlglot.parser import Parser
+
+    from_args_list = Parser.FUNCTIONS.get(camel_to_snake_case(name))
```
tobymao, line 4279: why do you need to call camel_to_snake_case?

sqlglot/expressions.py
```diff
@@ -4220,6 +4246,51 @@ def _replace_placeholders(node, args, **kwargs):
...
+
```
tobymao, line 4280: do a pass on args / kwargs to run convert(...)

sqlglot/expressions.py
barakalon, line 4279: I feel like we should be using specific dialects?

GeorgeSittas: Because of line 2711:

```python
cls._sql_names = [camel_to_snake_case(cls.__name__)]
```

If we're relying on `Parser.FUNCTIONS` to find a matching `exp.Func` class, I think we need this as a pre-processing step for `name`.

sqlglot/expressions.py
tobymao: the name validate is not good if you're going to keep it a generator, it should be like error_messages

sqlglot/expressions.py
tobymao: i don't get it? the user is passing in the snake case already, we don't need to convert that

tobymao: right, func needs to take in a dialect argument so you can produce the correct parser

GeorgeSittas: Ah yeah, I see what you mean. Yeah, it doesn't really make sense. I'm reverting this, thanks.

GeorgeSittas: Got it, good point.

sqlglot/expressions.py
GeorgeSittas: Changed

sqlglot/expressions.py
GeorgeSittas: Fixed

GeorgeSittas: Done

sqlglot/expressions.py
GeorgeSittas: Fixed

sqlglot/expressions.py
GeorgeSittas: Done

tobymao (approved):

PR #1020, opened 2023-01-27
Add required import to example

DavidPratten:


tobymao:
this is not required in the context of the example, so i'm going to close this for now

PR #1021, opened 2023-01-27
Map Snowflake EXCLUDE and RENAME keywords

griffatrasgo:
This PR maps Snowflake's proprietary EXCLUDE and RENAME keywords to standard EXCEPT and REPLACE keywords.

It supports parsing these keywords only when Snowflake is the specified dialect. It does not support using them in sql construction.

example code to test:
```

sql = """
SELECT DISTINCT
  TBL1.* EXCLUDE DATE_FIELD, ANOTHER_FIELD RENAME TBL1.BAD_NAME AS GOOD_NAME, C AS D
FROM DATABASE.SCHEMA.TABLE TBL1
JOIN DB.SC.TBL AS TBL2 ON TBL1.ID = TBL2.ID
WHERE INT_FIELD = 55
LIMIT 10
"""

parse_one(sql, read="snowflake")
parse_one(sql, read="snowflake").sql("snowflake")
transpile(sql, read="snowflake", write="snowflake")
```

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp, generator, parser, tokens
 from sqlglot.dialects.dialect import (
     Dialect,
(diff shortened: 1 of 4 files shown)


tobymao:
can you add snowflake tests in tests/dialects/snowflake.py?

sqlglot/dialects/snowflake.py
```diff
@@ -183,6 +185,26 @@ class Parser(parser.Parser):
             TokenType.PARTITION_BY: lambda self: self._parse_partitioned_by(),
         }
 
+        def _parse_except(self) -> t.Optional[t.List[t.Optional[exp.Expression]]]:
+            """
+            Override except parser to support Snowflake syntax without parens
+            """
+            if not self._match(TokenType.EXCEPT):
+                return None
+            if self._match(TokenType.L_PAREN, advance=False):
```
tobymao, line 194: sqlglot tries to make the default parser as leniant as possible, can we move this to the base parser?

tobymao: additionally, you're going to need to have a customer generator for snowflake to properly output EXCLUDE instead of EXCEPT

GeorgeSittas (approved):
Looks good

PR #1028, opened 2023-01-27
Add initial draft for the 'Dialects' section in docs

GeorgeSittas:


tobymao (approved):

sqlglot/dialects/__init__.py
```diff
@@ -1,3 +1,64 @@
+"""
+## Dialects
+
+One of the core abstractions in SQLGlot is the concept of a "dialect". The `Dialect` class essentially implements a
```
barakalon, line 4: These first few sentences feel a bit redundant and confusing.

Try introducing dialects as their own concept, separate from sqlglot.

"""
While there is a SQL standard, most SQL engines use a variation of the standard....

SQLGlot's `Dialect` classes implement dialects.

The base `Dialect` class aims to be as generic as possible....
"""

GeorgeSittas: Thanks for the feedback, I'll try to clean it up and make a follow up PR based on these suggestions.

I should probably wait a little before merging :)

GeorgeSittas: @barakalon does this look any better?
Screenshot 2023-01-27 at 8 10 32 PM

GeorgeSittas: Maybe the repetition of the word "dialect" feels a bit awkward?

barakalon: That looks better, yeah.

It's still a bit wordy.

> The most valuable of all talents is that of never using two words when one will do.
> \- Thomas Jefferson

It's usually a good idea to give writing a dedicated read through for _removing redundant information_ and _simplifying language_.

> While there is a SQL standard, no database vendor fully implements it. Most SQL engines support a variation of that standard, which has led to a plethora of similar, yet different SQL dialects, each of them existing for its own special purposes. Under these circumstances, it's usually very difficult to write portable SQL code, as the syntax may very from dialect to dialect. One common example is the usage of date/time functions, which can be especially hard to deal with.

Duplicates of "Most SQL engines support a variation of that standard":
- no database vendor fully implements it
- which has led to a plethora of similar, yet different SQL dialects
- each of them existing for its own special purposes
- as the syntax may very from dialect to dialect

Duplicates of "it's usually very difficult to write portable SQL code":
- which can be especially hard to deal with.
 
Overly complex language:
- Under these circumstances, it's usually very difficult to -> This makes it difficult to 
- One common example is the usage of -> One common example is

Here's a simplified paragraph, which I think has about the same amount of information:

> While there is a SQL standard, most SQL engines support a variation of that standard. This makes it difficult to write portable SQL code. One common example is date/time functions.

GeorgeSittas: Great feedback, much appreciated! I'll update the corresponding PR.

PR #1031, opened 2023-01-27
make optimizer more leniant

tobymao:


sqlglot/optimizer/isolate_table_selects.py
```diff
@@ -1,15 +1,18 @@
 from sqlglot import alias, exp
 from sqlglot.errors import OptimizeError
 from sqlglot.optimizer.scope import traverse_scope
+from sqlglot.schema import ensure_schema
 
 
-def isolate_table_selects(expression):
+def isolate_table_selects(expression, schema):
```
barakalon, line 7: schema=None

barakalon (approved):
Haha I love that the code is simpler.

PR #1032, opened 2023-01-27
Refactor dialect docs

GeorgeSittas:
I made some revisions, based on the feedback in https://github.com/tobymao/sqlglot/pull/1028#discussion_r1089212407.

### DIFF
--- a/sqlglot/dialects/__init__.py
+++ b/sqlglot/dialects/__init__.py
@@ -1,17 +1,18 @@
 """
 ## Dialects
 
-One of the core abstractions in SQLGlot is the concept of a "dialect". The `Dialect` class essentially implements a
-"SQLGlot dialect", which aims to be as generic and ANSI-compliant as possible. It relies on the base `Tokenizer`,
-`Parser` and `Generator` classes to achieve this goal, so these need to be very lenient when it comes to consuming
-SQL code.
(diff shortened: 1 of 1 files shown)


sqlglot/dialects/__init__.py
```diff
@@ -1,17 +1,18 @@
...
-"SQLGlot dialect", which aims to be as generic and ANSI-compliant as possible. It relies on the base `Tokenizer`,
-`Parser` and `Generator` classes to achieve this goal, so these need to be very lenient when it comes to consuming
-SQL code.
-
-However, there are cases where the syntax of different SQL dialects varies wildly, even for common tasks. One such
-example is the date/time functions, which can be hard to deal with. For this reason, it's sometimes necessary to
-override the base dialect in order to specialize its behavior. This can be easily done in SQLGlot: supporting new
-dialects is as simple as subclassing from `Dialect` and overriding its various components (e.g. the `Parser` class),
-in order to implement the target behavior.
+While there is a SQL standard, no database vendor fully implements it. Most SQL engines support a variation of
```
tobymao, line 4: i don't like the first sentence it's more so that there's a SQL standard, but it's not really standard in that no one strictly adheres to it

GeorgeSittas:
Made a few adjustments according to suggestions from @barakalon.

barakalon (approved):

PR #1033, opened 2023-01-27
Teradata CREATE TABLE options

treysp:
This PR adds Teradata "table options" to `CREATE TABLE`, using the existing `properties` argument for `exp.Create`.

Currently, for tables `_parse_create` allows the syntax `CREATE [flags] TABLE [table name] [table schema] [properties] [flags] [indexes]`.

Beyond those, Teradata has comma-separated "options." These differ from the existing `properties` by their location in the `CREATE TABLE` syntax. Options go between the table name and table schema. For example, "NO LOG" in `CREATE TABLE a, NO LOG (b INT)`. 

Teradata only has "options" between the table name and schema, and does not have `properties` between the table schema and flags. Therefore, we can modify the existing `properties` code for implementation.


tobymao:
does teradata ALSO have properties? could you just store these as properties and have teradata render / read them from the front?

sqlglot/parser.py
```diff
@@ -553,6 +553,15 @@ class Parser(metaclass=_Parser):
         TokenType.PROPERTIES: lambda self: self._parse_wrapped_csv(self._parse_property),
     }
 
+    OPTION_PARSERS = {
+        TokenType.BLOCKCOMPRESSION: lambda self: self._parse_blockcompression(),
```
tobymao, line 557: we shouldn't need token types for these, we should make it so that OPTION_PARSERS takes in text, otherwise you have to always add token types

so change to 
{
"BLOCKCOMPRESSION": ...
} 
and delete all the token types you've added

we could possibly refactor property parsers to do the same thing but don't necessarily have to do that

treysp:
Teradata does NOT also have properties, and using the existing properties code is much cleaner than my initial approach!

sqlglot/generator.py
```diff
@@ -473,7 +507,7 @@ def create_sql(self, expression: exp.Create) -> str:
 
         indexes = expression.args.get("indexes")
         index_sql = ""
-        if indexes is not None:
+        if indexes is not None and len(indexes) > 0:
```
tobymao, line 510: how about if indexes

sqlglot/generator.py
```diff
@@ -440,8 +459,20 @@ def uniquecolumnconstraint_sql(self, _) -> str:
         return "UNIQUE"
 
     def create_sql(self, expression: exp.Create) -> str:
-        this = self.sql(expression, "this")
         kind = self.sql(expression, "kind").upper()
+        has_before_properties = expression.args.get("properties")
+        has_before_properties = (
+            has_before_properties.args.get("before") if has_before_properties else None
```
tobymao, line 465: why do you need a "before" property on exp.Properties?, isn't it sufficient to have BEFORE_PROPERTIES constant?

sqlglot/parser.py
```diff
@@ -558,6 +558,38 @@ class Parser(metaclass=_Parser):
         TokenType.PROPERTIES: lambda self: self._parse_wrapped_csv(self._parse_property),
     }
 
+    PROPERTY_PARSERS_TEXT = {
```
tobymao, line 561: can you just add this to property_parsers? i don't like how there's another dict

sqlglot/parser.py
```diff
@@ -1010,6 +1058,16 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
             begin=begin,
         )
 
+    def _parse_property_before(self) -> t.Optional[exp.Expression]:
+        self._match_text_seq("NO")
```
tobymao, line 1062: what's the point of this? doesn't this just ignore no, dual, and default?

treysp: no/dual/default can be the first word of multiple options, so they have to be consumed before we can figure out which option parser to call.

These do that consumption, then the option parsers determine no/dual/default by looking at the previous token (like "BEFORE" on line 567).

One alternative approach would be to do the no/dual/default routing here like:
```
if self._match_text_seq("NO"):
    if self.PROPERTY_PARSERS_TEXT.get(self._curr.text.upper()):
        return self.PROPERTY_PARSERS_TEXT[self._curr.text.upper()](self, no=True)
```

GeorgeSittas:
Adding this for future reference: https://docs.teradata.com/r/Teradata-VantageTM-SQL-Data-Definition-Language-Syntax-and-Examples/September-2020/Table-Statements/CREATE-TABLE-and-CREATE-TABLE-...-AS/CREATE-TABLE-Syntax

sqlglot/parser.py
```diff
@@ -1039,6 +1097,147 @@ def _parse_property_assignment(self, exp_class: t.Type[exp.Expression]) -> exp.E
...
+    def _parse_withjournaltable(self) -> exp.Expression:
+        self._match_text_seq("WITH", "JOURNAL", "TABLE")
+        self._match(TokenType.EQ)
+        return self.expression(exp.WithJournalTableProperty, this=self._parse_table_parts())
+
+    def _parse_log(self, no=False) -> exp.Expression:
+        self._match_text_seq("LOG")
+        return self.expression(exp.LogProperty, no=no)
+
+    def _parse_journal(self, no=False, dual=False, before=False) -> exp.Expression:
```
GeorgeSittas, line 1135: Seems like the `before` parameter is not used

sqlglot/parser.py
```diff
@@ -1039,6 +1097,147 @@ def _parse_property_assignment(self, exp_class: t.Type[exp.Expression]) -> exp.E
...
+
+    def _parse_checksum(self) -> exp.Expression:
+        self._match_text_seq("CHECKSUM")
+        self._match(TokenType.EQ)
+
+        on = None
+        if self._match(TokenType.ON):
+            on = True
+        elif self._match_text_seq("OFF"):
+            on = False
```
GeorgeSittas, line 1154: If we had `match_text_seq` return `None` instead of `False`, this could be more clean.

```suggestion
        on = not self._match_text_seq("OFF") or self._match(TokenType.ON)
```

GeorgeSittas (commented):
Left a few comments, but looks good in general.

I think I'd tackle this problem by adding a new arg in `exp.Create` instead of trying to reuse `properties` and introduce changes elsewhere in the code, because I think that the current approach might be more complicated than needed.

sqlglot/generator.py
```diff
@@ -686,6 +733,82 @@ def likeproperty_sql(self, expression: exp.LikeProperty) -> str:
...
+        if default is not None or min is not None:
+            if default:
+                property = "DEFAULT"
+            elif min:
+                property = "MINIMUM"
+            else:
+                property = "MAXIMUM"
+            return f"{property} DATABLOCKSIZE"
+        else:
+            units = expression.args.get("units"):
```
GeorgeSittas, line 774:
```suggestion
            units = expression.args.get("units")
```

GeorgeSittas:
I'll merge and try to fix some errors I introduced + see if I can refactor it.

Thanks for the PR! 👍

sqlglot/generator.py
```diff
@@ -65,6 +65,11 @@ class Generator:
         exp.ReturnsProperty: lambda self, e: self.naked_property(e),
         exp.ExecuteAsProperty: lambda self, e: self.naked_property(e),
         exp.VolatilityProperty: lambda self, e: e.name,
+        exp.FallbackProperty: lambda self, e: f"{'NO ' if e.args.get('no') else ''}FALLBACK{' PROTECTION' if e.args.get('protection') else ''}",
+        exp.WithJournalTableProperty: lambda self, e: f"WITH JOURNAL TABLE={self.sql(e, 'this')}",
+        exp.LogProperty: lambda self, e: f"{'NO ' if e.args.get('no') else ''}LOG",
+        exp.JournalProperty: lambda self, e: f"{'NO ' if e.args.get('no') else ''}{'DUAL ' if e.args.get('dual') else ''}{'BEFORE ' if e.args.get('before') else ''}JOURNAL",
```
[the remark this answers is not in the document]
treysp: Submitting another small PR to address this and your other comment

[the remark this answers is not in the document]
GeorgeSittas: Thanks! I'll probably add the other changes discussed in this thread tomorrow.

sqlglot/parser.py
treysp: Fixed in #1068

sqlglot/parser.py
treysp: The parsers identify no/dual/default by looking at _prev, so we need those to have already been consumed.

sqlglot/generator.py
treysp: I think that should be sufficient - I'll confirm and submit another PR

PR #1040, opened 2023-01-29
support match recognize except for pattern regex fixes #1038

### LINKED ISSUE #1038: AST + parsing support for `MATCH_RECOGNIZE`
Snowflake implements `MATCH_RECOGNIZE`, which sqlglot doesn't seem to currently parse. I believe Oracle also implements `MATCH_RECOGNIZE`.

https://docs.snowflake.com/en/sql-reference/constructs/match_recognize.html

I'd be happy to provide a PR for this, but won't be able to get to it for a few weeks.

tobymao:


tobymao:
@barakalon i'm not sure how to parse a regex for example

select 
match_recognize(
pattern (^+$s) 
)

i'm thinking of two ways... in snowflake/oracle tokenizers, hardcode looking for regex in pattern (i'll just serialize the strings)

another more radical approach is by leveraging the parser's debug sql string (we use parser.sql in order show more context in debug / error messages) the parser has access to the original SQL string which i could extract the text as is.

the latter idea is interesting and could have other uses

barakalon:
Interesting.

"hardcode looking for regex in pattern" feels purer to me. I.e. it seems reasonable that certain token types would kickoff sub-tokenizers. 

"leveraging the parser's debug sql string" feels hacky but practical if the first option is too invasive.

(Take that all with a grain of salt - I haven't spent much time with the tokenizer)

tobymao:
@barakalon i ended up going with the sql string approach. the issue is the tokenizer has not context so i don't know to kick off something

PR #1050, opened 2023-01-30
Add support for dudkdb brace syntax, transpile struct_pack to it

### LINKED ISSUE #1045: DuckDB struct syntax fails to parse
DuckDB has [a Python-like struct construction syntax](https://duckdb.org/docs/sql/data_types/struct.html#creating-structs) that sqlglot cannot parse right now:

```
In [7]: import sqlglot as sg

In [8]: sg.__version__
Out[8]: '10.5.6'

In [9]: sg.parse_one("SELECT {'a': 1} AS x", read="duckdb")
...
ParseError: Invalid expression / Unexpected token. Line 1, Col: 8.
  SELECT {'a': 1} AS x
```

Here it is in the DuckDB CLI:

```
❯ duckdb <<< "SELECT {'a': 1} AS x"
┌───────────────────┐
│         x         │
│ struct(a integer) │
├───────────────────┤
│ {'a': 1}          │
└───────────────────┘
```

GeorgeSittas:
Fixes #1045


sqlglot/dialects/duckdb.py
```diff
@@ -66,12 +66,11 @@ def _sort_array_reverse(args):
     return exp.SortArray(this=seq_get(args, 0), asc=exp.false())
 
 
-def _struct_pack_sql(self, expression):
+def _struct_sql(self, expression):
     args = [
-        self.binary(e, ":=") if isinstance(e, exp.EQ) else self.sql(e)
-        for e in expression.expressions
+        f"'{e.name or e.this.name}': {self.sql(e, 'expression')}" for e in expression.expressions
```
GeorgeSittas, line 71: The `e.name or e.this.name` part takes care of both the `struct_pack` call, which contains EQ expression, and the brace syntax, which uses Slice expressions.

sqlglot/tokens.py
```diff
@@ -451,9 +451,11 @@ class Tokenizer(metaclass=_Tokenizer):
         },
         **{
             f"{prefix}{key}": TokenType.BLOCK_END
-            for key in ("}}", "%}", "#}")
+            for key in ("%}", "#}")
             for prefix in ("", "+", "-")
         },
+        "+}}": TokenType.BLOCK_END,
+        "-}}": TokenType.BLOCK_END,
```
GeorgeSittas, line 458: Discussed in slack w/ Toby -- Matching a `}}` block end after this PR will require a pair match on `R_BRACE`.

sqlglot/dialects/duckdb.py
tobymao: may be safer to do like

self.sql(exp.Literal.string(e.name)) to have escapes

tobymao (approved):

PR #1061, opened 2023-01-31
Fix Hive WITH CUBE/ROLLUP syntax

### LINKED ISSUE #1055: Hive Dialect for "WITH ROLLUP" error
when parsing hive sql, there is a Parse Error

image
image
AND **the error** is 
image

This SQL case is just copied from the official documentation https://cwiki.apache.org/confluence/display/Hive/Enhanced+Aggregation%2C+Cube%2C+Grouping+and+Rollup

GeorgeSittas:
Fixes #1055

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp, generator, parser, tokens, transforms
 from sqlglot.dialects.dialect import (
     Dialect,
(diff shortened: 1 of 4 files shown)


GeorgeSittas:
An alternative for this one is to add a flag in the base generator to switch between the previous supported syntax and Hive's `WITH CUBE/ROLLUP`, so that we duplicate less code over to `hive.py`. Happy to go either way.

GeorgeSittas:
> does hive support the standard way?

The CUBE/ROLLUP docs didn't have an example, so I would guess not.

tobymao:
can we do a flag on parser and gen?

GeorgeSittas:
Yeah I'll fix it

GeorgeSittas:
@tobymao should we rather add the "WITH CUBE/ROLLUP" logic in the base Parser, Generator without using a flag approach? That way we can capture both syntax forms without having to exclude either with a flag.

tobymao:
yes that's what i was suggesting

GeorgeSittas:
Ok I made the changes. It feels like it can be simplified more, let me know if I missed a cleaner way to implement this.

tobymao (approved):

PR #1064, opened 2023-01-31
Add support for ALTER TABLE DROP PARTITION

### LINKED ISSUE #1059: Hive Dialect for 'drop partition' error
sqlglot cannot parse 'Alter table drop partition'
image
image

GeorgeSittas:
Fixes #1059


sqlglot/generator.py
```diff
@@ -629,13 +634,7 @@ def national_sql(self, expression: exp.National) -> str:
...
 
     def partition_sql(self, expression: exp.Partition) -> str:
-        keys = csv(
-            *[
-                f"{prop.name}={self.sql(prop, 'value')}" if prop.args.get("value") else prop.name
-                for prop in expression.this
-            ]
-        )
-        return f"PARTITION({keys})"
+        return f"PARTITION({self.expressions(expression, 'this')})"
```
tobymao, line 637: in generally expressions should be called with "expressions", not with "this"

sqlglot/generator.py
```diff
@@ -1419,14 +1418,19 @@ def altertable_sql(self, expression: exp.AlterTable) -> str:
             actions = self.expressions(expression, "actions")
         elif isinstance(actions[0], exp.Delete):
             actions = self.expressions(expression, "actions", flat=True)
-        elif isinstance(actions[0], self.WITH_SINGLE_ALTER_TABLE_ACTION):
+        elif actions[0].__class__ in self.WITH_SINGLE_ALTER_TABLE_ACTION:
```
tobymao, line 1421: why was this needed

GeorgeSittas: Not necessary to use, it's only helpful for reporting "unsupported" errors I think. We can probably put all cases under the `else` branch instead. Should I do that?

tobymao (approved):

Issue #1066, opened 2023-02-01
GLOB operator is not supported for duckdb and sqlite


cldellow:
Hi, sqlglot is amazing! Thank you for this library.

I noticed that this fails:

```python
import sqlglot

print(sqlglot.transpile("SELECT 'foo' GLOB '?'", read="duckdb", write="sqlite")[0])
```

I expected it to print `SELECT 'foo' GLOB '?'`, but it gave an exception:

```
Traceback (most recent call last):
  File "main.py", line 3, in <module>
    print(sqlglot.transpile("SELECT 'foo' GLOB '?'", read="duckdb", write="sqlite")[0])
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/sqlglot/__init__.py", line 124, in transpile
    for expression in parse(sql, read, error_level=error_level)
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/sqlglot/__init__.py", line 60, in parse
    return dialect.parse(sql, **opts)
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/sqlglot/dialects/dialect.py", line 151, in parse
    return self.parser(**opts).parse(self.tokenizer.tokenize(sql), sql)
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/sqlglot/parser.py", line 684, in parse
    return self._parse(
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/sqlglot/parser.py", line 749, in _parse
    self.raise_error("Invalid expression / Unexpected token")
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/sqlglot/parser.py", line 792, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 21.
  SELECT 'foo' GLOB '?'
```

Searching the code base, I think this operator isn't supported for any dialect yet. I'm not sure which flavours of SQL support it - duckdb ([docs](https://duckdb.org/docs/sql/functions/patternmatching.html#glob)) and sqlite ([docs](https://www.sqlite.org/lang_expr.html#glob)) do.

I imagine this might be an easy fix, but I don't know where to start

GeorgeSittas:
Hey!

First of all, since this operator is supported by at least two dialects, e.g. duckdb and sqlite, I'd say it's worth to add in the base Parser, Generator classes. We generally try to make the base classes handle as much SQL as possible.

[...]

PR #1067, opened 2023-02-01
add GLOB for sqlite, duckdb

cldellow:
fixes tobymao/sqlglot#1066

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -85,6 +85,7 @@ class DuckDB(Dialect):
             **tokens.Tokenizer.KEYWORDS,
             ":=": TokenType.EQ,
             "CHARACTER VARYING": TokenType.VARCHAR,
+            "GLOB": TokenType.GLOB,
         }
 
     class Parser(parser.Parser):
--- a/sqlglot/dialects/sqlite.py
+++ b/sqlglot/dialects/sqlite.py
@@ -50,6 +50,7 @@ class SQLite(Dialect):
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
             "AUTOINCREMENT": TokenType.AUTO_INCREMENT,
+            "GLOB": TokenType.GLOB,
         }
 
     class Parser(parser.Parser):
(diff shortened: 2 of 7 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -85,6 +85,7 @@ class Tokenizer(tokens.Tokenizer):
             **tokens.Tokenizer.KEYWORDS,
             ":=": TokenType.EQ,
             "CHARACTER VARYING": TokenType.VARCHAR,
+            "GLOB": TokenType.GLOB,
```
GeorgeSittas, line 88: This is redundant because you've already added it in the base tokenizer's mapping

sqlglot/dialects/sqlite.py
```diff
@@ -50,6 +50,7 @@ class Tokenizer(tokens.Tokenizer):
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
             "AUTOINCREMENT": TokenType.AUTO_INCREMENT,
+            "GLOB": TokenType.GLOB,
```
GeorgeSittas, line 53: Same

tests/dialects/test_sqlite.py
```diff
@@ -50,6 +50,13 @@ def test_ddl(self):
         )
 
     def test_sqlite(self):
+        self.validate_all(
```
GeorgeSittas, line 53: You can write tests for glob directly in fixtures/identity, because it's supported by the base parser, generator and glob expressions generally transpile back to themselves.

GeorgeSittas (approved):
Left a few comments but looks good otherwise. Thanks!

Also, it'd be great to add some more tests.

tests/dialects/test_sqlite.py
tobymao: jut put this test in identity.sql

cldellow:
Thanks for that feedback, I've pushed an update.

I'm seeing lint failures from black in files that this PR doesn't touch: https://github.com/tobymao/sqlglot/actions/runs/4065549105/jobs/7000406980 - I'm not sure what's going on there.

PR #1069, opened 2023-02-01
Only match on property texts, remove some prop tokens

GeorgeSittas:
Tried to clean up `PROPERTY_PARSERS` once again, so that we now only use strings instead of mixing them up with tokens.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -175,13 +175,9 @@ class Parser(metaclass=_Parser):
         TokenType.DEFAULT,
         TokenType.DELETE,
         TokenType.DESCRIBE,
-        TokenType.DETERMINISTIC,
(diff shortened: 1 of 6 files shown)


sqlglot/parser.py
```diff
@@ -1123,6 +1102,20 @@ def _parse_fallback(self, no=False) -> exp.Expression:
             exp.FallbackProperty, no=no, protection=self._match_text_seq("PROTECTION")
         )
 
+    def _parse_with_property(
```
GeorgeSittas, line 1105: This was introduced because previously we were matching both `"WITH"` and `TokenType.WITH`, so we somehow need to disambiguate.

tobymao (approved):

PR #1075, opened 2023-02-02
add: transpile ifnull from bigquery to coalesce

rogerzs:
### Add new translation

this pull request adds the IFNULL function conversion to Coalesce Function.

IFNULL Function hasn't a direct correspondent, but it does the same thing as the coalesce and NVL

**OBS**: is my first PR in repo, please tell me if i did something wrong :)

### DIFF
--- a/tests/test_transpile.py
+++ b/tests/test_transpile.py
@@ -256,6 +256,14 @@ FROM bar /* comment 5 */, tbl /*          comment 6 */""",
         )
         self.validate("SELECT IF(a > 1, 1) FROM foo", "SELECT CASE WHEN a > 1 THEN 1 END FROM foo")
 
+    def test_if_null(self):
+        self.validate(
+            "SELECT IFNULL(1, NULL) FROM foo",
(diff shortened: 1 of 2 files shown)


tests/test_transpile.py
```diff
@@ -256,6 +256,14 @@ def test_if(self):
         )
         self.validate("SELECT IF(a > 1, 1) FROM foo", "SELECT CASE WHEN a > 1 THEN 1 END FROM foo")
 
+    def test_if_null(self):
```
tobymao, line 259: i think we should transpile all IFNULL calls to COALESCE in every dialect, so let's validate that.

delete this test and write a new one in test_dialects where you can use the validate_all command

remove the IFNULL mapping that you added to bigquery, and also remove the IFNULL in Tableau.

add this mapping to the base .py FUNCTIONS

GeorgeSittas:
I'm merging this and will push the remaining changes that Toby suggested.

Thanks @rogerzs!

PR #1078, opened 2023-02-03
Improve typing if using into

eakmanrq:
The goal for this was to improve typing so that if someone provides a type for `into` then we know that type is returned so now mypy will recognize the type as the return type.

[Reference discussion on the complexities of doing this](https://github.com/python/mypy/issues/3737)


tobymao (approved):
sigh

PR #1080, opened 2023-02-03
Refactor CREATE properties generation

treysp:
- Combine BEFORE_PROPERTIES, WITH_PROPERTIES, and ROOT_PROPERTIES into a single property location dictionary PROPERTIES_LOCATION

- Collect all dialect-specific properties into PROPERTIES_LOCATION

- Only retain dialect-specific property locations if they override default in PROPERTIES_LOCATION

- Refactor properties routing logic in create_sql

- Remove unneeded exp.Properties "before" argument


treysp:
I reordered the spark tests because presto's lack of support for LocationProperty was causing the hive and spark tests to fail. I was unable to determine why state wasn't resetting between each dialect in validate_all.

I can't reproduce the problem outside the test suite - transpiling presto then hive/spark works fine in shell or .py file executed from cli.

GeorgeSittas:
> I reordered the spark tests because presto's lack of support for LocationProperty was causing the hive and spark tests to fail. I was unable to determine why state wasn't resetting between each dialect in validate_all.

I can reproduce this locally too.. Sounds like a bug to me, there's probably a dictionary being mutated somewhere.

treysp:
I don't think you're missing anything, and I agree that it's likely a bug. Wasn't sure how you wanted to handle, so I just changed the test order as an interim step.

Should I pull back the PR until I've tracked it down and resolved?

GeorgeSittas:
No need to pull back the PR, just take a look and see if you can track it down. I'll examine your PR more closely soon and see if I can find something.

sqlglot/generator.py
```diff
@@ -455,19 +460,22 @@ def uniquecolumnconstraint_sql(self, _) -> str:
...
-        has_before_properties = expression.args.get("properties")
-        has_before_properties = (
-            has_before_properties.args.get("before") if has_before_properties else None
-        )
-        if kind == "TABLE" and has_before_properties:
+        properties = expression.args.get("properties")
+        properties_locs = self.locate_properties(properties) if properties else {}
+        if properties and properties_locs["root_with"]:
+            # subset expression properties to just root and with properties
+            expression.set("properties", exp.Properties(expressions=properties_locs["root_with"]))
```
treysp, line 467: This overwrites the expressions properties, removing the unsupported properties.

If the input isn't re-parsed between tests, then LocationProperty is being removed during the Presto test and not being replaced afterward.

treysp: I did it this way so I could leave all the code that handles the WITH and ROOT properties unmodified.

Should I take another approach that doesn't modify the original expression?

GeorgeSittas: Yes it's not correct to mutate the original expression in generator methods. If you must change the expression, then you need to copy it first.

sqlglot/generator.py
```diff
@@ -99,34 +99,39 @@ class Generator:
...
-        exp.LikeProperty,
-    }
-
-    WITH_PROPERTIES = {
-        exp.Property,
-        exp.FileFormatProperty,
-        exp.PartitionedByProperty,
-        exp.TableFormatProperty,
+    PROPERTIES_LOCATION = {
+        exp.AfterJournalProperty: "pre_schema",
```
tobymao, line 103: can you make this an enum?

what are the different types?

sqlglot/generator.py
```diff
@@ -695,25 +704,55 @@ def properties(
...
             expressions = self.expressions(properties, sep=sep, indent=False)
-            expressions = expressions if before else self.wrap(expressions)
+            expressions = self.wrap(expressions) if wrapped else expressions
             return f"{prefix}{' ' if prefix and prefix != ' ' else ''}{expressions}{suffix}"
         return ""
 
     def with_properties(self, properties: exp.Properties) -> str:
         return self.properties(properties, prefix=self.seg("WITH"))
 
+    def locate_properties(self, properties: exp.Properties) -> dict:
```
tobymao, line 718: t.Dict[str, exp.Property] ?

sqlglot/generator.py
```diff
@@ -99,34 +99,39 @@ class Generator:
...
-        exp.SortKeyProperty,
-        exp.LikeProperty,
-    }
-
-    WITH_PROPERTIES = {
-        exp.Property,
-        exp.FileFormatProperty,
-        exp.PartitionedByProperty,
-        exp.TableFormatProperty,
+    PROPERTIES_LOCATION = {
```
GeorgeSittas, line 102: If we want to keep this, then maybe it can be simplified by doing something like:

```python
PROPERTIES_LOCATION = {
    "pre_schema": {
        exp.AfterJournalPropery,
        ...
    },
    "post_schema_root": { ... }
}
```

sqlglot/generator.py
```diff
@@ -695,25 +704,55 @@ def properties(
...
-            expressions = expressions if before else self.wrap(expressions)
+            expressions = self.wrap(expressions) if wrapped else expressions
             return f"{prefix}{' ' if prefix and prefix != ' ' else ''}{expressions}{suffix}"
         return ""
 
     def with_properties(self, properties: exp.Properties) -> str:
         return self.properties(properties, prefix=self.seg("WITH"))
 
+    def locate_properties(self, properties: exp.Properties) -> dict:
+        properties_locs = {}
```
tobymao, line 719: can you do this in one pass in stead of n passes?

for property in properties.expressions:
   if ...
     locations[...] = ..
  elif...

tests/dialects/test_spark.py
```diff
@@ -39,9 +39,9 @@ def test_ddl(self):
             "CREATE TABLE x USING ICEBERG PARTITIONED BY (MONTHS(y)) LOCATION 's3://z'",
             write={
                 "duckdb": "CREATE TABLE x",
-                "presto": "CREATE TABLE x WITH (TABLE_FORMAT='ICEBERG', PARTITIONED_BY=ARRAY['MONTHS'])",
                 "hive": "CREATE TABLE x USING ICEBERG PARTITIONED BY (MONTHS(y)) LOCATION 's3://z'",
                 "spark": "CREATE TABLE x USING ICEBERG PARTITIONED BY (MONTHS(y)) LOCATION 's3://z'",
+                "presto": "CREATE TABLE x WITH (TABLE_FORMAT='ICEBERG', PARTITIONED_BY=ARRAY['MONTHS'])",
```
tobymao, line 44: move this back

sqlglot/generator.py
GeorgeSittas: Or, as Toby mentioned, we can also use enum values as keys.

tobymao: that’s better

tobymao: let’s go with george’s solution

treysp: That's how I originally implemented it, seems like it could go either way.

This is simpler in the processing code, a bit more complicated for dialects to override (although that may not be a real concern).

tobymao: i see, it's harder to override because then you have to get the inner.

ok, let's go with enums, the dialects should be easier to override

treysp: Updated to use enums

sqlglot/generator.py
```diff
@@ -695,25 +723,46 @@ def properties(
...
             return f"{prefix}{' ' if prefix and prefix != ' ' else ''}{expressions}{suffix}"
         return ""
 
     def with_properties(self, properties: exp.Properties) -> str:
         return self.properties(properties, prefix=self.seg("WITH"))
 
+    def locate_properties(
+        self, properties: exp.Properties
+    ) -> t.Dict[exp.Properties.Location, list[exp.Property]]:
+        properties_locs: t.Dict[exp.Properties.Location, list[exp.Property]]
```
tobymao, line 740:
```suggestion
        properties_locs: t.Dict[exp.Properties.Location, list[exp.Property]]  = {key: [] for key in exp.Properties.Location}
```

sqlglot/generator.py
```diff
@@ -695,25 +723,46 @@ def properties(
...
         return ""
 
     def with_properties(self, properties: exp.Properties) -> str:
         return self.properties(properties, prefix=self.seg("WITH"))
 
+    def locate_properties(
+        self, properties: exp.Properties
+    ) -> t.Dict[exp.Properties.Location, list[exp.Property]]:
+        properties_locs: t.Dict[exp.Properties.Location, list[exp.Property]]  = {key: [] for key in exp.Properties.Location}
+        properties_locs = {key: [] for key in exp.Properties.Location}
```
tobymao, line 741:
```suggestion
```

tobymao (approved):

tobymao:
nice work, thanks for the great addition and quick iteration

Issue #1077, opened 2023-02-02
generate_series() is not transpiled to sequence() when going from postgres to trino sql.


andrewhong5297:
Something like this in postgres

`generate_series('2019-01-01'::timestamp, now(), '1day') as day`

should be the following in trino

`sequence(cast('2019-01-01' as timestamp), cast(now() as timestamp), '1' day) as day`

PR #1081, opened 2023-02-03
Add generate_series postgres function, fix parsing for now()

GeorgeSittas:
Fixes (partially) #1077 

Outstanding items:

- [ ] Fix generation for presto/trino: check type of 1st arg & cast 2nd to match it if necessary. Valid types for these are `bigint`, `date` & `timestamp(p)`. The 3rd (step) arg can be interval day to second/month for the last two.
- [ ] Fix parsing for postgres: if the 3rd arg is an interval of the form `[INTERVAL] <string>` (e.g. "1 day"), convert it into a proper `Interval` expression so that it can be transpiled more easily.

Can either tackle these items for this PR, or open a separate one later.
cc: @tobymao


tests/dialects/test_bigquery.py
```diff
@@ -170,7 +170,7 @@ def test_bigquery(self):
                 "bigquery": "CURRENT_TIMESTAMP()",
                 "duckdb": "CURRENT_TIMESTAMP()",
                 "postgres": "CURRENT_TIMESTAMP",
-                "presto": "CURRENT_TIMESTAMP()",
+                "presto": "CURRENT_TIMESTAMP",
```
GeorgeSittas, line 173: Checked both docs & ran it locally -- `current_timestamp` is supposed to be a no-parentheses function.

PR #1086, opened 2023-02-03
Fix quoted id/var parsing

### LINKED ISSUE #1084: SQLite to Postgres single-quote alias conversions
The following query is valid SQLite–

```
SELECT 1 AS 'This is valid';
```

The transpiler from SQLite to Postgres strips the quotes and produces a Postgres syntax error:

```
sqlglot.transpile("SELECT 1 AS 'This is valid in sqlite'", read='sqlite', write='postgres')
> ['SELECT 1 AS This is valid in sqlite']
```

This is unfortunately not valid Postgres.

```
postgres=# SELECT 1 AS This is valid in sqlite;
ERROR:  syntax error at or near "is"
LINE 1: SELECT 1 AS This is valid in sqlite;
```

Correct output would be `SELECT 1 AS "This is valid in sqlite"`

GeorgeSittas:
Fixes #1084

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3156,7 +3156,9 @@ class Parser(metaclass=_Parser):
                 prefix += self._prev.text
 
         if (any_token and self._advance_any()) or self._match_set(tokens or self.ID_VAR_TOKENS):
-            return exp.Identifier(this=prefix + self._prev.text, quoted=False)
+            quoted = self._prev.token_type == TokenType.STRING
+            return exp.Identifier(this=prefix + self._prev.text, quoted=quoted)
+
         return None
 
     def _parse_string(self) -> t.Optional[exp.Expression]:
--- a/tests/fixtures/identity.sql
+++ b/tests/fixtures/identity.sql
@@ -161,6 +161,7 @@ SELECT 1 FROM test
 SELECT * FROM a, b, (SELECT 1) AS c
 SELECT a FROM test
 SELECT 1 AS filter
+SELECT 1 AS 'quoted alias'
 SELECT SUM(x) AS filter
 SELECT 1 AS range FROM test
 SELECT 1 AS count FROM test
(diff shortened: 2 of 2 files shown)


sqlglot/parser.py
```diff
@@ -3156,7 +3156,9 @@ def _parse_id_var(
                 prefix += self._prev.text
 
         if (any_token and self._advance_any()) or self._match_set(tokens or self.ID_VAR_TOKENS):
-            return exp.Identifier(this=prefix + self._prev.text, quoted=False)
+            quoted = self._prev.token_type == TokenType.STRING
+            return exp.Identifier(this=prefix + self._prev.text, quoted=quoted)
```
tobymao, line 3160: maybe it should instead be

exp.identifier(...)

GeorgeSittas: I think we'll also need to do `token_type == TokenType.STRING` for `exp.to_identifier`. If we just do

```
return exp.to_identifier(prefix + self._prev.text)
```

, there are several tests that break:

```
AssertionError: 'X((y AS z))."1"' != 'X((y AS z)).1'
- X((y AS z))."1"
?             - -
+ X((y AS z)).1
```

```
AssertionError: 'CREATE PROCEDURE foo @a INTEGER, @b INTEGER AS SELECT @a = SUM[17 chars] bar' != 'CREATE PROCEDURE foo "@a" INTEGER, "@b" INTEGER AS SELECT "@a"[23 chars] bar'
- CREATE PROCEDURE foo @a INTEGER, @b INTEGER AS SELECT @a = SUM(bla) FROM baz AS bar
+ CREATE PROCEDURE foo "@a" INTEGER, "@b" INTEGER AS SELECT "@a" = SUM(bla) FROM baz AS bar
?                      +  +          +  +                   +  +
```

The tests that fail contain identifiers with special characters like @, $ and numbers in the front, because these are not captured by `SAFE_IDENTIFIER_RE` and so are force-quoted.

tobymao: ok forget it then

tobymao (approved):

PR #1087, opened 2023-02-03
Fix transpilation of generate_series for trino

### LINKED ISSUE #1077: generate_series() is not transpiled to sequence() when going from postgres to trino sql.

GeorgeSittas:
Completed the remaining tasks for https://github.com/tobymao/sqlglot/pull/1081

Fixes #1077 

Note: for the time being, I only handled casting to `timestamp` (if needed). This can be easily refined in the future if we want to include other cases too.

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -29,7 +29,7 @@ DATE_DIFF_FACTOR = {
 }
 
 # This pattern is used to capture valid postgres string intervals like '  15  days '
-INTERVAL_STRING_RE = re.compile("([0-9]+)([a-zA-Z]+)")
+INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
 
 
 def _date_add_sql(kind):
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import re
+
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -26,6 +28,9 @@
     "DAY": " / 86400",
 }
 
+# This pattern is used to capture valid postgres string intervals like '  15  days '
+INTERVAL_STRING_RE = re.compile("([0-9]+)([a-zA-Z]+)")
```
tobymao, line 32: it seems like this could be a generic method, maybe make this a function so it's not only used in generate_series?

also maybe it should be in dialects.py

GeorgeSittas: You mean just extracting a number and a var from a string the matches the re? Or creating an Interval expression too?

tobymao: we probably always want an interval right?

GeorgeSittas: I guess

GeorgeSittas:
Maybe change `to_interval` to receive a string Literal too?

tobymao:
sure

tobymao (approved):

tobymao (approved):

Issue #1091, opened 2023-02-04
Unable to parse mysql8 create view 


Karl-Wu:
I am working on column level lineage parser with sqlglot. Things were going well but trying to parse the following statement.

CREATE ALGORITHM=UNDEFINED DEFINER=`dbt`@`%` SQL SECURITY DEFINER VIEW `stg_payments` AS with `source` as (select `raw_payments`.`id` AS `id`,`raw_payments`.`order_id` AS `order_id`,`raw_payments`.`payment_method` AS `payment_method`,`raw_payments`.`amount` AS `amount` from `raw_payments`), `renamed` as (select `source`.`id` AS `payment_id`,`source`.`order_id` AS `order_id`,`source`.`payment_method` AS `payment_method`,(`source`.`amount` / 100) AS `amount` from `source`) select `renamed`.`payment_id` AS `payment_id`,`renamed`.`order_id` AS `order_id`,`renamed`.`payment_method` AS `payment_method`,`renamed`.`amount` AS `amount` from `renamed`

Above sql was generated by dbt and it is critical in our case.
https://dev.mysql.com/doc/refman/8.0/en/create-view.html

tobymao:
i think we should leverage the properties that @treysp created here with a pre kind

treysp:
Agreed - I'll add these as properties

treysp:
will do!

tobymao:
feel free to join our slack if you wanna chat at any time https://join.slack.com/t/tobiko-data/shared_invite/zt-1ma66d79v-a4dbf4DUpLAQJ8ptQrJygg

GeorgeSittas:
Fixed in https://github.com/tobymao/sqlglot/commit/12bd655654d3264616b8e5e3c92e2ff23d6a070c

PR #1101, opened 2023-02-06
Add POST_CREATE properties

treysp:
Adds mysql properties from #1091 in new POST_CREATE location


treysp:
This parses and generates the options correctly, but transpile is not respecting case sensitivity in a couple of places.

This test I was going to add to identity doesn't pass:
`CREATE ALGORITHM=UNDEFINED DEFINER=dbt@% SQL SECURITY DEFINER VIEW a AS (SELECT a FROM b)`

It transpiles as:
`CREATE ALGORITHM=UNDEFINED DEFINER=DBT@% SQL SECURITY DEFINER VIEW A AS (SELECT A FROM B)`

I'll come back and work on this later today, but figured I'd get it in a pr in case you want to finish it off before I get back to it.

sqlglot/expressions.py
```diff
@@ -1285,6 +1285,18 @@ class Property(Expression):
     arg_types = {"this": True, "value": True}
 
 
+class AlgorithmProperty(Expression):
+    arg_types = {"this": True}
+
+
+class DefinerProperty(Expression):
+    arg_types = {"this": True}
```
GeorgeSittas, line 1293: You can just do `pass` here.

sqlglot/expressions.py
```diff
@@ -1285,6 +1285,18 @@ class Property(Expression):
...
+
+
+class SqlSecurityProperty(Expression):
+    arg_types = {"definer": True}
```
GeorgeSittas, line 1297: Maybe we can just use `this` for this and then do `pass` too.

GeorgeSittas (commented):
Looks good, let's add some tests.

GeorgeSittas:
Seems like I missed your comment about tests etc. Thanks for the PR, I'll try to address remaining comments and merge it.

treysp:
Gah, I totally missed that it was being parsed as a Command due to the _parse_create routing. Sorry for the messy/incomplete PR...

GeorgeSittas:
No worries

sqlglot/expressions.py
```diff
@@ -1428,10 +1442,12 @@ class Properties(Expression):
     arg_types = {"expressions": True}
 
     NAME_TO_PROPERTY = {
```
tobymao, line 1444: i'm not a big fan of the NAME_TO_PROPERTY dictionary. why isn't this just a property on on the property?

sqlglot/generator.py
```diff
@@ -552,6 +556,15 @@ def create_sql(self, expression: exp.Create) -> str:
                 )
             index_sql = "".join(indexes_sql)
 
+        postcreate_props_sql = ""
+        if properties and properties_locs[exp.Properties.Location.POST_CREATE]:
```
tobymao, line 560: can you refactor all the places where we do if properties and properties[xxx] and make properties default to a dict

properties = expression.args.get("Propertyers") or {}

then you can just use get

sqlglot/parser.py
```diff
@@ -1129,6 +1138,26 @@ def _parse_with_property(
...
+        if not user or not host:
+            return None
+
+        return exp.DefinerProperty(this=f"{user}@{host}")
+
+    def _parse_sqlsecurity(self) -> exp.Expression:
+        if self._match_text_seq("DEFINER"):
+            definer = True
+        elif self._match_text_seq("INVOKER"):
+            definer = False
```
tobymao, line 1158: isn't defined undefined here? if it doesn't match these two branches?

sqlglot/expressions.py
GeorgeSittas: Agreed, it doesn't feel very elegant. On the other hand, we used to store the property name in `this`, which wasn't very elegant either.

sqlglot/parser.py
GeorgeSittas: fixed

GeorgeSittas:
I'm gonna merge this so that I can iterate faster and address the remaining two comments from Toby.

PR #1117, opened 2023-02-08
Add more type hints, make some small fixes

GeorgeSittas:


sqlglot/dialects/dialect.py
```diff
@@ -11,6 +11,9 @@
 from sqlglot.tokens import Tokenizer
 from sqlglot.trie import new_trie
 
+if t.TYPE_CHECKING:
```
tobymao, line 14: you don’t need to hide in type checking unless there’s a cyclic dependency

sqlglot/dialects/bigquery.py
```diff
@@ -14,8 +16,11 @@
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
```
tobymao, line 18: here too. everywhere

GeorgeSittas: Thanks, missed that one

tobymao (approved):

PR #1136, opened 2023-02-09
expand_laterals

barakalon:
Support for "lateral column alias references" (e.g. https://aws.amazon.com/about-aws/whats-new/2018/08/amazon-redshift-announces-support-for-lateral-column-alias-reference/). These are valid in snowflake, redshift, duckdb, and possibly more.

```sql
SELECT
  x.a + 1 AS i,
  i + 1 AS j,
  j + 1 AS k
FROM x;

-- to

SELECT
  x.a + 1 AS i,
  x.a + 1 + 1 AS j,
  x.a + 1 + 1 + 1 AS k
FROM x;
```

Not crazy about splitting `validate_qualify_columns` out. But this seemed like a practical way to not keep complicating `qualify_columns` logic.

### DIFF
--- a/sqlglot/optimizer/expand_laterals.py
+++ b/sqlglot/optimizer/expand_laterals.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+import typing as t
+
+from sqlglot import exp
+
+
+def expand_laterals(expression: exp.Expression) -> exp.Expression:
+    """
+    Expand lateral column alias references.
+
+    This assumes `qualify_columns` as already run.
+
+    Example:
+        >>> import sqlglot
+        >>> sql = "SELECT x.a + 1 AS b, b + 1 AS c FROM x"
+        >>> expression = sqlglot.parse_one(sql)
+        >>> expand_laterals(expression).sql()
+        'SELECT x.a + 1 AS b, x.a + 1 + 1 AS c FROM x'
+
+    Args:
+        expression: expression to optimize
+    Returns:
+        optimized expression
+    """
+    for select in expression.find_all(exp.Select):
+        alias_to_expression: t.Dict[str, exp.Expression] = {}
+        for projection in select.expressions:
+            for column in projection.find_all(exp.Column):
+                if not column.table and column.name in alias_to_expression:
+                    column.replace(alias_to_expression[column.name].copy())
+                if isinstance(projection, exp.Alias):
+                    alias_to_expression[projection.alias] = projection.this
+    return expression
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -47,7 +47,7 @@ def validate_qualify_columns(expression):
     for scope in traverse_scope(expression):
         if isinstance(scope.expression, exp.Select):
             unqualified_columns.extend(scope.unqualified_columns)
-            if scope.external_columns and not scope.is_udtf and not scope.is_correlated_subquery:
+            if scope.external_columns and not scope.is_correlated_subquery:
                 raise OptimizeError(f"Unknown table: {scope.external_columns[0].table}")
 
     if unqualified_columns:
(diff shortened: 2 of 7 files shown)


sqlglot/optimizer/expand_laterals.py
```diff
@@ -0,0 +1,30 @@
+from sqlglot import expressions as exp
```
tobymao, line 1:
```suggestion
from sqlglot import exp
```

sqlglot/optimizer/expand_laterals.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Example:
+        >>> import sqlglot
+        >>> sql = "SELECT x.a + 1 AS b, b + 1 AS c FROM x"
+        >>> expression = sqlglot.parse_one(sql)
+        >>> expand_laterals(expression).sql()
+        'SELECT x.a + 1 AS b, x.a + 1 + 1 AS c FROM x'
+
+    Args:
+        expression (sqlglot.Expression): expression to optimize
```
tobymao, line 18: remove the type hints

GeorgeSittas (approved):

tests/fixtures/optimizer/expand_laterals.sql
```diff
@@ -0,0 +1,40 @@
...
+  FROM x
+);
+SELECT
+  *
+FROM (
+  SELECT
+    x.a + 1 AS i,
+    x.a + 1 + 1 AS j
+  FROM x
+);
```
tobymao, line 40: what about the case where it's ambiguous, we need to default to the table and not the alias, is that tested

select x.b AS a,
a
FROM x

a should be the table

GeorgeSittas:
Maybe add a test case that displays how the optimizer handles ambiguous cases like the following?

```python
>>> import sqlglot
>>>
>>> sql = "SELECT x.a + 1 AS b, b + 1 AS c FROM x"
>>> schema = {"x": {"a": "INT", "b": "INT"}}
>>>
>>> optimize(sqlglot.parse_one(sql), schema=schema).sql()
'SELECT "x"."a" + 1 AS "b", "x"."b" + 1 AS "c" FROM "x" AS "x"'
>>>
>>> optimize(sqlglot.parse_one(sql), schema={"x": {"a": "INT"}}).sql()
'SELECT "x"."a" + 1 AS "b", "x"."a" + 2 AS "c" FROM "x" AS "x"'
```

tests/fixtures/optimizer/expand_laterals.sql
GeorgeSittas: seems like it's the table column, I also tried that now

GeorgeSittas: it's because `expand_laterals` runs after `qualify_columns` and the replacement condition in the former involves checking if there's a table in the column expression.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -37,11 +37,24 @@ def qualify_columns(expression, schema):
...
     return expression
 
 
+def validate_qualify_columns(expression):
+    """Raise an `OptimizeError` if any columns aren't qualified"""
+    unqualified_columns = []
+    for scope in traverse_scope(expression):
+        if isinstance(scope.expression, exp.Select):
+            unqualified_columns.extend(scope.unqualified_columns)
+            if scope.external_columns and not scope.is_udtf and not scope.is_correlated_subquery:
```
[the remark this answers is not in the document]
tobymao: do we still need udtf check here now that i've added udtf support?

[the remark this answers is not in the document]
barakalon: Probably not

tests/fixtures/optimizer/expand_laterals.sql
barakalon: That's right, this implicitly depends on qualify columns.

I could make this independent, but that'll require building scope. I'm not sure what the scope performance overhead is, but in general I try to avoid constructing it unnecessarily.. 

Happy to switch.

tobymao: no i was just asking for a test

PR #1142, opened 2023-02-09
Add support for ${..} parameter style

### LINKED ISSUE #1138: Databricks: Widget references not parsing with braces, and no WIDGET keyword support
Databricks allows the use of widgets, which are variables that can be passed between notebook jobs. They are referenced using a $ with or without braces. The following is an example of usage from the Databricks documentation:
[https://docs.databricks.com/notebooks/widgets.html#language-sql](url)

image

Trying to parse this statement in sqlglot gives a ParseError:

image

To be clear, without braces the $ works as intended:

image

Furthermore, the WIDGET keyword is not listed. I have gotten around this by extending the Spark tokenizer class with specific widget commands. I can continue to do this, but support for the WIDGET keyword would help.

image

GeorgeSittas:
fixes #1138


sqlglot/dialects/oracle.py
```diff
@@ -113,3 +113,6 @@ class Tokenizer(tokens.Tokenizer):
             "VARCHAR2": TokenType.VARCHAR,
             "NVARCHAR2": TokenType.NVARCHAR,
         }
+
+        SINGLE_TOKENS = tokens.Tokenizer.SINGLE_TOKENS.copy()
+        SINGLE_TOKENS.pop("$")
```
GeorgeSittas, line 118: Moving the dollar sign to the base tokenizer caused a problem in the `V$SESSION` oracle test, because it parsed it as `V AS $SESSION`.

sqlglot/expressions.py
```diff
@@ -2399,7 +2399,7 @@ def output_name(self):
 
 
 class Parameter(Expression):
-    pass
+    arg_types = {"this": True, "start": True, "end": False}
```
tobymao, line 2402: paramater should not have start and end

sqlglot/parser.py
```diff
@@ -3306,6 +3304,16 @@ def _parse_star(self) -> t.Optional[exp.Expression]:
...
 
+    def _parse_parameter(self) -> exp.Expression:
+        start = self._prev.text
+        if self._match(TokenType.L_BRACE):
+            start += "{"
+
+        this = self._parse_var() or self._parse_primary()
+        end = "}" if self._match(TokenType.R_BRACE) else None
+
+        return self.expression(exp.Parameter, this=this, start=start, end=end)
```
tobymao, line 3315: instead of start, and end, just do wrapped

sqlglot/dialects/dialect.py
```diff
@@ -429,3 +429,8 @@ def trim_sql(self: Generator, expression: exp.Trim) -> str:
     from_part = "FROM " if trim_type or remove_chars else ""
     collation = f" COLLATE {collation}" if collation else ""
     return f"TRIM({trim_type}{remove_chars}{from_part}{target}{collation})"
+
+
+def parameter_sql(self, expression: exp.Parameter) -> str:
```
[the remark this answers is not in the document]
GeorgeSittas: yea good idea, fixing it now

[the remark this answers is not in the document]
GeorgeSittas: done

tobymao (approved):

PR #1154, opened 2023-02-12
Add support for insert or ignore

EdAyers:
Using the [sqlite definition](https://www.sqlite.org/syntax/insert-stmt.html). 

Still need to do unit tests, I'm not sure what to do if the sql dialect doesn't support `INSERT OR IGNORE`. Eg Postgres I think instead uses `ON CONFLICT DO NOTHING` after values. Maybe it's better just to not have this.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1070,25 +1070,17 @@ class Index(Expression):
         "amp": False,  # teradata
     }
 
+
 class Insert(Expression):
     arg_types = {
         "with": False,
         "this": True,
-        "or_cond": False,
+        "or_": False,
         "expression": False,
         "overwrite": False,
         "exists": False,
         "partition": False,
     }
-
-    @property
-    def or_cond(self):
-        oc = self.args.get("or_cond")
-        if oc is not None:
-            return oc.text
-        else:
-            return None
-
 
 
 # https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1070,16 +1070,25 @@ class Index(Expression):
         "amp": False,  # teradata
     }
 
-
 class Insert(Expression):
     arg_types = {
         "with": False,
         "this": True,
+        "or_cond": False,
         "expression": False,
         "overwrite": False,
         "exists": False,
         "partition": False,
     }
+
+    @property
+    def or_cond(self):
+        oc = self.args.get("or_cond")
+        if oc is not None:
+            return oc.text
+        else:
+            return None
+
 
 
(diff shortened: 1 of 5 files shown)


tobymao:
run `make check` to do styling and tests

sqlglot/expressions.py
```diff
@@ -1070,17 +1070,26 @@ class Index(Expression):
         "amp": False,  # teradata
     }
 
-
 class Insert(Expression):
     arg_types = {
         "with": False,
         "this": True,
+        "or_cond": False,
```
tobymao, line 1077: just rename this to "or"

sqlglot/expressions.py
```diff
@@ -1070,17 +1070,26 @@ class Index(Expression):
...
         "expression": False,
         "overwrite": False,
         "exists": False,
         "partition": False,
     }
 
+    @property
+    def or_cond(self):
```
tobymao, line 1085: delete this method

sqlglot/generator.py
```diff
@@ -932,20 +932,22 @@ def lockingproperty_sql(self, expression: exp.LockingProperty) -> str:
 
     def insert_sql(self, expression: exp.Insert) -> str:
         overwrite = expression.args.get("overwrite")
+        or_cond = "" if expression.or_cond is None else f"OR {expression.or_cond} "
```
tobymao, line 935: self.sql(expression, "or")

sqlglot/tokens.py
```diff
@@ -117,6 +117,7 @@ class TokenType(AutoName):
     OBJECT = auto()
 
     # keywords
+    ABORT = auto()
```
tobymao, line 120: remove all of these keywords that you added

sqlglot/parser.py
```diff
@@ -1423,6 +1432,11 @@ def _parse_describe(self) -> exp.Expression:
 
         return self.expression(exp.Describe, this=this, kind=kind)
 
+    def _parse_or_cond(self):
+        if self._match_text_seq("OR"):
+          if self._match_set(self.INSERT_OR_KINDS):
```
tobymao, line 1437: self._match_texts(...)

sqlglot/parser.py
```diff
@@ -1423,6 +1432,11 @@ def _parse_describe(self) -> exp.Expression:
 
         return self.expression(exp.Describe, this=this, kind=kind)
 
+    def _parse_or_cond(self):
```
tobymao, line 1435: add type -> TokenType

GeorgeSittas:
@EdAyers any plans to finish this? Let me know if you need any help

sqlglot/parser.py
EdAyers: I can't figure out what I'm supposed to return here

GeorgeSittas:
Closed in favor of #1184

PR #1158, opened 2023-02-13
Fix translating DATE_TRUNC function for spark dialect

alesk:
In the main branch, parsing `DATE_TRUNC('day', some_ts)` erroneously translates to `TRUNC(some_ts, 'day')`. Spark sql has a DATE_TRUNC functions since version 2.3.0. Rendering to spark dialect should preserve the `DATE_TRUNC` function.

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp, parser
 from sqlglot.dialects.dialect import create_with_partitions_sql, rename_func, trim_sql
 from sqlglot.dialects.hive import Hive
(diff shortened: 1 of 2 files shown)


barakalon (commented):
What's the difference between `exp.DateTrunc` and `exp.TimestampTrunc`?
Should this be using `exp.DateTrunc`?

alesk:
> What's the difference between `exp.DateTrunc` and `exp.TimestampTrunc`? Should this be using `exp.DateTrunc`?

There is a [TRUNC(ts, unit)](https://spark.apache.org/docs/2.3.0/api/sql/index.html#trunc) in Spark, and this function should be translated to `exp.DateTrunc` expression. It operates on the arguments of type `date` and returns `date`. The `DATE_TRUNC` operates on timestamps and should use the internal `exp.TimestampTrunc` representation.

sqlglot/dialects/spark.py
```diff
@@ -71,6 +81,7 @@ class Parser(Hive.Parser):
                 ),
                 length=seq_get(args, 1),
             ),
+            "DATE_TRUNC": _timestamp_trunc,
```
tobymao, line 84: why not just do rename_func ?

sqlglot/dialects/spark.py
```diff
@@ -44,6 +46,14 @@ def _unix_to_time(self, expression):
     raise ValueError("Improper scale for timestamp")
 
 
+def _timestamp_trunc(args: t.Sequence) -> exp.Expression:
+    unit = seq_get(args, 0)
+    if isinstance(unit, exp.Column):
```
tobymao, line 51: why would this happen? there's no test case for this, can you just use rename_func?

tobymao:
will reimplemnt this

PR #1159, opened 2023-02-13
initial xml table support

tobymao:


sqlglot/parser.py
```diff
@@ -582,12 +582,37 @@ class Parser(metaclass=_Parser):
     }
 
     CONSTRAINT_PARSERS = {
-        TokenType.CHECK: lambda self: self.expression(
-            exp.Check, this=self._parse_wrapped(self._parse_conjunction)
+        "AUTOINCREMENT": lambda self: self._parse_auto_increment(),
```
tobymao, line 585: all of these can be changed to : self._parse_auto_increment

sqlglot/parser.py
```diff
@@ -3010,6 +3015,13 @@ def _parse_foreign_key(self) -> exp.Expression:
         )
 
     def _parse_primary_key(self) -> exp.Expression:
+        desc = None
```
tobymao, line 3018: desc = (self.Match(ASC) or ...) and token_type == TokenType.desc

sqlglot/parser.py
```diff
@@ -582,12 +582,37 @@ class Parser(metaclass=_Parser):
     }
 
     CONSTRAINT_PARSERS = {
-        TokenType.CHECK: lambda self: self.expression(
```
tobymao, line 585: can you delete all t he tokens? that are no longer used

sqlglot/parser.py
GeorgeSittas: We need the lambdas because we can't reference `self` there, right?

sqlglot/parser.py
GeorgeSittas: Rewrote it with `self._match_set` instead.

sqlglot/parser.py
tobymao: ah right

sqlglot/parser.py
GeorgeSittas: got rid of some of them

GeorgeSittas:
Ok should be good

PR #1175, opened 2023-02-15
Fix group by parsing to correctly handle rollup and cube

### LINKED ISSUE #1166: More About Presto Group by
In #1079 , we have solved the order problem about four types. 

However, in Presto, there can be multiple keywords like `group by rollup(field_a), rollup(field_b), rollup(field_c)`, which is equivalent to `rollup(field_a, field_b, field_c)`. So it is still a problem.

```python
sqlglot.parse_one(
"""
select a
from t
group by a, rollup(b), rollup(c), rollup(d)
"""
)
```

GeorgeSittas:
Fixes #1166 

- [ ] Add test cases
- [ ] Fix SQL generation.


sqlglot/generator.py
```diff
@@ -1068,21 +1068,26 @@ def group_sql(self, expression: exp.Group) -> str:
             f"{self.seg('GROUPING SETS')} {self.wrap(grouping_sets)}" if grouping_sets else ""
         )
 
-        cube = expression.args.get("cube")
-        if cube is True:
-            cube = self.seg("WITH CUBE")
+        cube = ensure_list(expression.args.get("cube"))
```
tobymao, line 1071: cube = expression.args.get("cube")

if cube is True:
   ...
elif isinstance(cube, list):
else
...
    

i prefer this i think

GeorgeSittas: The tradeoff that led to this change was that the parser is more simple in this way, I think.

tobymao (approved):

sqlglot/generator.py
GeorgeSittas:
```python
        while True:
            expressions = self._parse_csv(self._parse_conjunction)
            if expressions:
                elements["expressions"].extend(expressions)

            grouping_sets = self._parse_grouping_sets()
            if grouping_sets:
                elements["grouping_sets"].extend(grouping_sets)

            rollup = None
            cube = None

            with_ = self._match(TokenType.WITH)
            if self._match(TokenType.ROLLUP):
                rollup = with_ or self._parse_wrapped_csv(self._parse_column)
                elements["rollup"].extend(ensure_list(rollup))

            if self._match(TokenType.CUBE):
                cube = with_ or self._parse_wrapped_csv(self._parse_column)
                elements["cube"].extend(ensure_list(cube))

            if not (expressions or grouping_sets or rollup or cube):
                break
```

Treating everything as a list here is useful, because otherwise the defaultdict doesn't help much

PR #1177, opened 2023-02-15
Fix UDF kwarg parsing

### LINKED ISSUE #1176: Parse error: character varying
I am trying to parse a create function sql. The error message is like this.

```
ParseError: Expecting ). Line 1, Col: 81.
  CREATE OR REPLACE FUNCTION function_name (input_a character varying DEFAULT NULL::character varying)
```
The first "character" is underlined.

It seems **character varying** cannot be parse correctly, which should be equivalent to VARCHAR.

> The notations varchar(n) and char(n) are aliases for character varying(n) and character(n), respectively. If specified, the length must be greater than zero and cannot exceed 10485760. character without length specifier is equivalent to character(1). If character varying is used without length specifier, the type accepts strings of any size.

Reference: https://www.postgresql.org/docs/current/datatype-character.html

GeorgeSittas:
Fixes #1176

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2760,7 +2760,7 @@ class Parser(metaclass=_Parser):
         if not self._match(TokenType.L_PAREN):
             return this
 
-        expressions = self._parse_csv(self._parse_udf_kwarg)
+        expressions = self._parse_csv(lambda: self._parse_column_def(self._parse_id_var()))
         self._match_r_paren()
         return self.expression(
             exp.UserDefinedFunction, this=this, expressions=expressions, wrapped=True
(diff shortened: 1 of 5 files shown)


tobymao (approved):

tobymao:
nice

sqlglot/parser.py
```diff
@@ -2760,7 +2760,7 @@ def _parse_user_defined_function(
         if not self._match(TokenType.L_PAREN):
             return this
 
-        expressions = self._parse_csv(self._parse_udf_kwarg)
+        expressions = self._parse_csv(lambda: self._parse_column_def(self._parse_id_var()))
```
tobymao, line 2763: maybe. make this a function? so it’s not done in tsql too

GeorgeSittas: sure

PR #1180, opened 2023-02-15
Teradata - user-defined types require "SYSUDTLIB." prefix in schema

treysp:
Teradata requires prefixing user-defined types with "SYSUDTLIB." in a schema definition.

For example, UDT "mytype" would be specified as `CREATE TABLE a (b SYSUDTLIB.mytype(1))`.


sqlglot/expressions.py
```diff
@@ -2487,6 +2487,7 @@ class DataType(Expression):
         "expressions": False,
         "nested": False,
         "values": False,
+        "syslib": False,
```
tobymao, line 2490: rename to prefix perhaps?

sqlglot/generator.py
```diff
@@ -653,6 +653,9 @@ def hexstring_sql(self, expression: exp.HexString) -> str:
     def datatype_sql(self, expression: exp.DataType) -> str:
         type_value = expression.this
         type_sql = self.TYPE_MAPPING.get(type_value, type_value.value)
+        syslib_sql = expression.args.get("syslib")
```
tobymao, line 656: prefix

treysp:
I couldn't find any other SQL dialects that required this kind of prefix, so I hard-coded "SYSUDTLIB.".

My initial implementation in [this commit](https://github.com/tobymao/sqlglot/pull/1180/commits/c015927d05f6355df7bd39865081fa33c3911aea) parsed the prefix, but that seemed overly complicated if Teradata is the only dialect that needs this.

tobymao:
i don't think it needs to be that complicated, for now, i'd change it to prefix, and then hardcode looking for sysudtlib. if we find any other, then we can make that section more complex

sqlglot/generator.py
```diff
@@ -653,6 +653,9 @@ def hexstring_sql(self, expression: exp.HexString) -> str:
     def datatype_sql(self, expression: exp.DataType) -> str:
         type_value = expression.this
         type_sql = self.TYPE_MAPPING.get(type_value, type_value.value)
+        prefix_sql = expression.args.get("prefix")
+        prefix_sql = f"{prefix_sql}." if prefix_sql else ""
+        type_sql = f"{prefix_sql}{type_sql}"
```
GeorgeSittas, line 658: I think it might be better to put this logic only in teradata. Like:
```python
    type_sql = super().datatype_sql(expression)
    prefix_sql = expression.args.get("prefix")
    return f"{prefix_sql}.{type_sql}" if prefix_sql else ""
```

GeorgeSittas: Also, if it's always `SYSUDTLIB` you could just use a boolean arg instead?

GeorgeSittas: Also, maybe we can just use a boolean arg for this, since SYSUDTLIB is the only supported prefix right now.

tobymao (approved):

Issue #1168, opened 2023-02-14
Optimize Error in qualify_columns in presto unnest 


ruichengnew:
```python
optimize(
    sqlglot.parse_one(
        """
select table.a, t.b
from table_t
cross join unnest(split(b, ',')) AS view_t(b)
""",
        read="presto",
    ),
    {},
).sql()
```

Looking at the example above, the same named column `b` in unnest split function should come from `table_t`, but not `view_t`.

However, sqlglot gave the output below, which is not right

```python
SELECT "table_t"."a" AS "a", "view_t"."b" AS "b" FROM "table_t" AS "table_t" CROSS JOIN UNNEST(SPLIT("view_t"."b", \',\')) AS "view_t"("b")
```

By the way, in 'hive' dialect, sqlglot works well with `lateral view`.

PR #1181, opened 2023-02-15
UDTF scope refactor

barakalon:
fixes https://github.com/tobymao/sqlglot/issues/1168

This adds `lateral_sources`. I'm not crazy about yet another concept in scope.

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -584,8 +584,7 @@ def _traverse_tables(scope):
     for join in scope.expression.args.get("joins") or []:
         expressions.append(join.this)
 
-    for lateral in scope.expression.args.get("laterals") or []:
-        expressions.append(lateral)
+    expressions.extend(scope.expression.args.get("laterals") or [])
 
     for expression in expressions:
         if isinstance(expression, exp.Table):
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -57,7 +57,7 @@ class Scope:
     ):
         self.expression = expression
         self.sources = sources or {}
-        self.lateral_sources = {**lateral_sources} if lateral_sources else {}
+        self.lateral_sources = lateral_sources.copy() if lateral_sources else {}
         self.sources.update(self.lateral_sources)
         self.outer_column_list = outer_column_list or []
         self.parent = parent
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -252,7 +252,7 @@ class Scope:
 
             external_columns = [
                 column
-                for scope in self.subquery_scopes + self.udtf_scopes
+                for scope in itertools.chain(self.subquery_scopes, self.udtf_scopes)
                 for column in scope.external_columns
             ]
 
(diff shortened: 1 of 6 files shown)


tests/fixtures/optimizer/optimizer.sql
```diff
@@ -1,15 +1,10 @@
...
-WITH "z_2" AS (
-  SELECT
-    "z"."a" AS "a"
-  FROM "z" AS "z"
-)
 SELECT
   "z"."a" AS "a",
   "q"."m" AS "m"
-FROM "z_2" AS "z"
+FROM "z" AS "z"
```
tobymao, line 7: why was this fixed? this is great

sqlglot/optimizer/scope.py
```diff
@@ -511,82 +535,99 @@ def _traverse_union(scope):
...
+    if from_:
+        expressions.extend(from_.expressions)
 
-        if table_name in scope.sources:
-            # This is a reference to a parent source (e.g. a CTE), not an actual table.
-            scope.sources[source_name] = scope.sources[table_name]
+    for join in scope.expression.args.get("joins") or []:
+        expressions.append(join.this)
+
+    for lateral in scope.expression.args.get("laterals") or []:
```
GeorgeSittas, line 587: same as toby's comment above

sqlglot/optimizer/scope.py
```diff
@@ -511,82 +535,99 @@ def _traverse_union(scope):
...
+        if isinstance(expression, exp.Table):
+            table_name = expression.name
+            source_name = expression.alias_or_name
+
+            if table_name in scope.sources:
+                # This is a reference to a parent source (e.g. a CTE), not an actual table.
+                sources[source_name] = scope.sources[table_name]
+            else:
+                sources[source_name] = expression
+            continue
```
GeorgeSittas, line 600: we can also do if-elif-else here to remove this `continue`

sqlglot/optimizer/scope.py
```diff
@@ -511,82 +535,99 @@ def _traverse_union(scope):
...
+    # Traverse FROMs, JOINs, and LATERALs in the order they are defined
+    expressions = []
+    from_ = scope.expression.args.get("from")
+    if from_:
+        expressions.extend(from_.expressions)
 
-        if table_name in scope.sources:
-            # This is a reference to a parent source (e.g. a CTE), not an actual table.
-            scope.sources[source_name] = scope.sources[table_name]
+    for join in scope.expression.args.get("joins") or []:
```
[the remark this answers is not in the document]
barakalon: This appends `join.this`, and I think `or` is safer, in case "joins" is None.

We could do
```python
expressions.extend(j.this for j in scope.expression.args.get("joins") or [])
```

But I think that's harder to read.

sqlglot/optimizer/scope.py
```diff
@@ -47,22 +53,28 @@ def __init__(
         outer_column_list=None,
         parent=None,
         scope_type=ScopeType.ROOT,
+        lateral_sources=None,
     ):
         self.expression = expression
         self.sources = sources or {}
+        self.lateral_sources = {**lateral_sources} if lateral_sources else {}
```
tobymao, line 60: lateral_sources.copy()

sqlglot/optimizer/scope.py
barakalon: It'd have to be
```python
if ...:
    ...
else:
    if ...:
        ...
    else:
        ...
    for child_scope in _traverse_scope(...):
        ...
```

GeorgeSittas: Oh sorry, I missed the following `for` block. Makes sense.

tobymao (approved):

PR #1183, opened 2023-02-15
Fix parsing of 'ANY' clause

### LINKED ISSUE #1178: parser fails to parse anonymous function in an ILIKE ANY clause
Hi there, 

We are building an internal udf which currently fails the sqlglot parser.

```
sqlglot.__version__
>> '11.1.1'
```

Given the following query (where `xxx` is a udf and expected to be parsed as an anonymous function)
```
from sqlglot import parse_one

query = """
SELECT *
FROM X
WHERE name ILIKE ANY xxx('a', 'b')
"""

parse_one(query)
```

Results in the following exception:
```
ParseError: Invalid expression / Unexpected token. Line 4, Col: 22.
  
SELECT *
FROM X
WHERE name ILIKE ANY xxx('a', 'b')
```

GeorgeSittas:
Fix #1178


tobymao (approved):

shaiv-syg:
Thanks!!!

PR #1184, opened 2023-02-15
Add support for INSERT OR ... sqlite syntax

GeorgeSittas:
Attempt to clean up https://github.com/tobymao/sqlglot/pull/1154

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1094,7 +1094,7 @@ class Insert(Expression):
         "overwrite": False,
         "exists": False,
         "partition": False,
-        "or": False,
+        "alternative": False,
     }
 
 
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -1094,6 +1094,7 @@ class Insert(Expression):
         "overwrite": False,
         "exists": False,
         "partition": False,
+        "or_": False,
```
tobymao, line 1097: we can just call it "or"

tobymao (approved):

sqlglot/expressions.py
GeorgeSittas: I don't think we can, it clashes with python's `or` when we try to instantiate an `exp.Insert` object using kwargs.

Issue #1182, opened 2023-02-15
`REGEXP_EXTRACT` bigquery function does not extract matching group in duckdb


nsfinkelstein:
Instead, the function extracts a pattern that matches the full regex.

To reproduce:

```py
import duckdb
import sqlglot
import pandas as pd

df = pd.DataFrame({'col': ['10', '11', '12', '13', '14', '15']})
original = "select regexp_extract(col, '1(\d)') as extracted from df"
transpiled = sqlglot.transpile(original, read='bigquery', write='duckdb')[0]
duckdb.connect().execute(transpiled).df()
#   extracted
# 0        10
# 1        11
# 2        12
# 3        13
# 4        14
# 5        15
```

The result should instead be:

```py
#   extracted
# 0        0
# 1        1
# 2        2
# 3        3
# 4        4
# 5        5
```

tobymao:
@nsfinkelstein is the behavior of bigquery that it returns the first match group? seems like duckdb needs to add the index 1

nsfinkelstein:
Thanks for the quick response. Yes, the default behavior is that bigquery returns the first match group (at least when there is only one match group, I haven't verified for when there is more than one). This is also reflected in the second example in their docs (https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_extract).

tobymao:
how would you translate bigquery regexp_extract into duckdb? i need help on this

tobymao:
yes, but that's not correct, you should only do this if the regex has a match group.

tobymao:
i don't think this is possible to do correctly. bigquery returns an error if there is more than one capture group, so it doesn't support match groups as duckdb does.

i'm going to close this for now as not planned. this is something that you can build a custom transform for. basically, you take your bigquery regexp, and then adding an arg of 1 to it before generating duckdb.

nsfinkelstein:
I would think it is possible to do correctly in the direction of transpiling bigquery to duckdb, because as you say duckdb allows more general matches. I suspect this is by far the more common direction of transpilation, as people use duckdb to unit test production systems (which is my use case as well).

It's not possible in the other direction only if the regex in duckdb SQL has more than one match group. In this case I'd expect sqlglot to raise an error.

If you point me in the right direction, I can put in a pull request to address both these points.

tobymao:
@nsfinkelstein 

1. Create a new class
```python
RegexpExtract(Func):
[...]

PR #1187, opened 2023-02-16
account for differences in regexp_extract translation

nsfinkelstein:
This pull request address issue https://github.com/tobymao/sqlglot/issues/1182, according to the approach described in https://github.com/tobymao/sqlglot/issues/1182#issuecomment-1431861858.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
 import typing as t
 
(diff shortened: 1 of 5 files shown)


nsfinkelstein:
@tobymao I'm sorry to say that I cannot identify the issue from the logs. Any pointers as to why the checks are failing would be appreciated.

tobymao:
@nsfinkelstein you need to run the linters

make check

tobymao:
i'll finish this for you, nice work. thanks for the pr

nsfinkelstein:
Thank you

PR #1189, opened 2023-02-16
Remove mapping from {{ to BLOCK_START

GeorgeSittas:


tobymao (approved):

tests/test_tokens.py
```diff
@@ -38,7 +38,8 @@ def test_jinja(self):
             tokens,
             [
                 (TokenType.SELECT, "SELECT"),
-                (TokenType.BLOCK_START, "{{"),
+                (TokenType.L_BRACE, "{"),
+                (TokenType.L_BRACE, "{"),
```
tobymao, line 42: why two?

tobymao (approved):

PR #1196, opened 2023-02-17
Add COMPRESS and INLINE LENGTH column constraints

treysp:


tests/fixtures/identity.sql
```diff
@@ -546,8 +546,8 @@ CREATE TABLE customer (pk BIGINT NOT NULL GENERATED ALWAYS AS IDENTITY (START WI
 CREATE TABLE customer (pk BIGINT NOT NULL GENERATED ALWAYS AS IDENTITY (START WITH 10))
 CREATE TABLE customer (pk BIGINT NOT NULL GENERATED ALWAYS AS IDENTITY (CYCLE))
 CREATE TABLE foo (baz_id INT REFERENCES baz(id) DEFERRABLE)
-CREATE TABLE foo (baz CHAR(4) CHARACTER SET LATIN UPPERCASE NOT CASESPECIFIC)
-CREATE TABLE foo (baz DATE FORMAT 'YYYY/MM/DD' TITLE 'title')
+CREATE TABLE foo (baz CHAR(4) CHARACTER SET LATIN UPPERCASE NOT CASESPECIFIC COMPRESS 'a')
+CREATE TABLE foo (baz DATE FORMAT 'YYYY/MM/DD' TITLE 'title' INLINE LENGTH 1 COMPRESS ('a', 'b'))
```
tobymao, line 550: can you add a test case like

create table foo (baz int, inline text)

treysp: Definitely. I realize now that the parser is loose and will allow "INLINE [anything]". Since the constraint is "INLINE LENGTH [number]", should the parser look for "LENGTH" and return None if it's not found?

Or should the parser stay loose and use your exact test case to make sure anything can come through after INLINE?

treysp: It looks like the constraint is preventing correct parsing of an 'inline' column name, so I'll work on this some more and resubmit.

tobymao: this was fixed recently in main, do you have that?

tobymao: https://github.com/tobymao/sqlglot/commit/9453c5e381ef45e8c24e9ac9728c9b5c7e583112

treysp: ah - i pulled it down but didn't have it merged. my bad.

sqlglot/parser.py
```diff
@@ -2882,6 +2884,14 @@ def _parse_auto_increment(self) -> exp.Expression:
 
         return exp.AutoIncrementColumnConstraint()
 
+    def _parse_compress(self) -> exp.Expression:
+        if self._match(TokenType.L_PAREN, advance=False):
+            return self.expression(
+                exp.CompressColumnConstraint, this=self._parse_wrapped_csv(self._parse_bitwise)
+            )
+        else:
+            return self.expression(exp.CompressColumnConstraint, this=self._parse_bitwise())
```
GeorgeSittas, line 2893:
```suggestion

        return self.expression(exp.CompressColumnConstraint, this=self._parse_bitwise())
```

GeorgeSittas (approved):

PR #1198, opened 2023-02-18
Change name of lineage root Table node to refer to column name

junpeng-jp:
Suggested fix for #1199. Changed the name of the root Table node to `c.sql()` instead of `c.table`

```python
 if table not in tables:
     tables[table] = Node(name=c.sql(), source=source, expression=source)
     node.downstream.append(tables[table])
```
image

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -113,7 +113,7 @@ def lineage(
                 )
             else:
                 if table not in tables:
-                    tables[table] = Node(name=str(c.this), source=source, expression=source)
+                    tables[table] = Node(name=str(c), source=source, expression=source)
                 node.downstream.append(tables[table])
 
         return node
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -113,7 +113,7 @@ def lineage(
                 )
             else:
                 if table not in tables:
-                    tables[table] = Node(name=table, source=source, expression=source)
+                    tables[table] = Node(name=str(c.this), source=source, expression=source)
                 node.downstream.append(tables[table])
 
         return node
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -113,7 +113,7 @@ def lineage(
                 )
             else:
                 if table not in tables:
-                    tables[table] = Node(name=str(c), source=source, expression=source)
+                    tables[table] = Node(name=c.sql(), source=source, expression=source)
                 node.downstream.append(tables[table])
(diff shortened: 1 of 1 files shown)


sqlglot/lineage.py
```diff
@@ -113,7 +113,7 @@ def to_node(
                 )
             else:
                 if table not in tables:
-                    tables[table] = Node(name=table, source=source, expression=source)
+                    tables[table] = Node(name=str(c.this), source=source, expression=source)
```
GeorgeSittas, line 116:
```suggestion
                    tables[table] = Node(name=c.name, source=source, expression=source)
```

junpeng-jp: @GeorgeSittas Just made another change to refer to the fully qualified column name so that the Node's name makes sense in context of its other metadata like `node.downstream` & node.expression` which is referencing the Table node

image

junpeng-jp: Here's a sample of the node's dataclass attributes
```python
Node(name='table.col1', expression=(TABLE this: 
  (IDENTIFIER this: table, quoted: False), db: 
  (IDENTIFIER this: db, quoted: False), catalog: 
  (IDENTIFIER this: catalog, quoted: False), alias: 
  (TABLEALIAS this: 
    (IDENTIFIER this: table, quoted: False))), source=(TABLE this: 
  (IDENTIFIER this: table, quoted: False), db: 
  (IDENTIFIER this: db, quoted: False), catalog: 
  (IDENTIFIER this: catalog, quoted: False), alias: 
  (TABLEALIAS this: 
    (IDENTIFIER this: table, quoted: False))), downstream=[])
```

GeorgeSittas: Yea, it makes sense to use the fully qualified name, otherwise we might lose context.

sqlglot/lineage.py
```diff
@@ -113,7 +113,7 @@ def to_node(
                 )
             else:
                 if table not in tables:
-                    tables[table] = Node(name=table, source=source, expression=source)
+                    tables[table] = Node(name=str(c), source=source, expression=source)
```
GeorgeSittas, line 116: `str(c)` is not very idiomatic in sqlglot, I'd rather use `c.sql()` here.

junpeng-jp: got it, will switch to `c.sql()`

GeorgeSittas (approved):

tobymao (approved):

PR #1200, opened 2023-02-19
Make type annotation more robust

GeorgeSittas:


tobymao:
add a test case, it should no longer crash

tobymao (approved):

GeorgeSittas:
> add a test case, it should no longer crash

In main, SQLGlot crashes only for the four rules specified in the below example. If we use the complete optimizer pipeline it doesn't crash, so I wasn't sure if we want to test just this subset of rules.

```python
>>> import sqlglot
>>> expr = sqlglot.parse_one("SELECT * FROM (SELECT * FROM bla.foo) WHERE baz <= 0")
>>>
>>> from sqlglot.optimizer.annotate_types import annotate_types
>>> from sqlglot.optimizer.expand_laterals import expand_laterals
>>> from sqlglot.optimizer.qualify_columns import qualify_columns
>>> from sqlglot.optimizer.qualify_tables import qualify_tables
>>>
>>>
>>> rules = (
...     qualify_tables,
...     qualify_columns,
...     expand_laterals,
...     annotate_types,
... )
>>>
>>> optimize(expr, rules=rules).sql()
  File "/Users/georgesittas/Code/tobiko/sqlglot/sqlglot/optimizer/optimizer.py", line 90, in optimize
    expression = rule(expression, **rule_kwargs)
  File "/Users/georgesittas/Code/tobiko/sqlglot/sqlglot/optimizer/annotate_types.py", line 31, in annotate_types
    return TypeAnnotator(schema, annotators, coerces_to).annotate(expression)
  File "/Users/georgesittas/Code/tobiko/sqlglot/sqlglot/optimizer/annotate_types.py", line 290, in annotate
    col.type = selects[col.table][col.name].type
KeyError: 'baz'
>>>
>>> optimize(expr).sql()
'SELECT * FROM "bla"."foo" AS "foo" WHERE "foo"."baz" <= 0'
```

tobymao:
due to pushdown? if so we should add pushdown as well to sqlmesh

GeorgeSittas:
> due to pushdown?

Probably, yeah.

PR #1202, opened 2023-02-20
Add is_star helper for exp.Select

GeorgeSittas:


sqlglot/expressions.py
```diff
@@ -2424,6 +2424,10 @@ def lock(self, update: bool = True, copy: bool = True) -> Select:
     def named_selects(self) -> t.List[str]:
         return [e.output_name for e in self.expressions if e.alias_or_name]
 
+    @property
```
tobymao, line 2427: can you add this in subquery able? so unions have it too

GeorgeSittas: Sure

GeorgeSittas: I added it to `Union` and `Subquery` instead, because the former also contains `expression` (besides `this`) and the latter doesn't inherit from `Subqueryable`.

tobymao (approved):

PR #1206, opened 2023-02-21
Extend the diff implementation with ability to provide a list of matched node pairs

izeigerman:


sqlglot/diff.py
```diff
@@ -58,7 +58,11 @@ class Keep:
     Edit = t.Union[Insert, Remove, Move, Update, Keep]
 
 
-def diff(source: exp.Expression, target: exp.Expression) -> t.List[Edit]:
+def diff(
+    source: exp.Expression,
+    target: exp.Expression,
+    pre_matchings: t.List[t.Tuple[exp.Expression, exp.Expression]] | None = None,
```
tobymao, line 64: i'd just call it matches

sqlglot/diff.py
```diff
@@ -109,16 +115,42 @@ def __init__(self, f: float = 0.6, t: float = 0.6) -> None:
...
-    def diff(self, source: exp.Expression, target: exp.Expression) -> t.List[Edit]:
+    def diff(
+        self,
+        source: exp.Expression,
+        target: exp.Expression,
+        pre_matchings: t.List[t.Tuple[exp.Expression, exp.Expression]] | None = None,
+    ) -> t.List[Edit]:
+        pre_matchings = pre_matchings or []
+
+        pre_matched_nodes = {n: n for pair in pre_matchings for n in pair}
```
tobymao, line 126: is this problematic if n is the same sql as another? the hash of an expression is the sql so for example if you had

select 1 -> select 2

select 1 -> select 3, it'd override the dict

tobymao: if you're using a dict here, why not just have the input be a dict?

izeigerman: Please note that this dict is not a dict of `select 1 -> select 2`. This is a dict of `select 1 -> select 1`, i.e. each element (both source and target) is mapped to itself. We simply flatten the list here. This is needed in order to compose a mapping of node ids - from input node ids to ids of corresponding nodes used in input trees.

> why not just have the input be a dict?

doesn't really buy us much implementation-wise but incurs additional runtime overhead on a caller side to compose this dict, which is not cheap for expressions.

sqlglot/diff.py
izeigerman: Renamed

sqlglot/diff.py
GeorgeSittas: Regardless of the mapping semantics though, Toby's point still holds, I think. If `select 1` appears two times in the match list, then the first occurrence will be overwritten and you'll lose the reference to an expression. I assume this is important, since you use `id` on the value you get from the dict lookup later on.

izeigerman: Dict doesn't help here, because similarly the value shouldn't show up more than once, not just the key. If we want to mitigate this we need to have an explicit validation instead, which I can add. But again dict doesn't help with that.

sqlglot/diff.py
```diff
@@ -109,16 +115,41 @@ def __init__(self, f: float = 0.6, t: float = 0.6) -> None:
         self.t = t
         self._sql_generator = Dialect().generator()
 
-    def diff(self, source: exp.Expression, target: exp.Expression) -> t.List[Edit]:
+    def diff(
+        self,
+        source: exp.Expression,
+        target: exp.Expression,
+        matchings: t.List[t.Tuple[exp.Expression, exp.Expression]] | None = None,
```
tobymao, line 122: matches

GeorgeSittas: I think matching is a valid term in this context (e.g. https://en.wikipedia.org/wiki/Matching_(graph_theory)).

tobymao: sounds werid to me

izeigerman: Yeah, I prefer matching over matches. "Matches" sound like something you use to start a fire.

GeorgeSittas (approved):

PR #1208, opened 2023-02-21
feature: add _meta slot & meta property for more flexible annotations

junpeng-jp:
Summary of changes
 - added `_meta` slot to `Expression`, defaults for `None`
 - added `self.metadata` property which returns `self._meta`
 - added `set_metadata(meta_key, meta_value)` method which:
   1. instantiates `self._meta = {}` if needed
   2. sets `self._meta[meta_key] = meta_value`

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -82,7 +82,7 @@ class Expression(metaclass=_Expression):
 
     key = "expression"
     arg_types = {"this": True}
-    __slots__ = ("args", "parent", "arg_key", "comments", "_type", "meta")
+    __slots__ = ("args", "parent", "arg_key", "comments", "_type", "_meta")
 
     def __init__(self, **args: t.Any):
         self.args: t.Dict[str, t.Any] = args
@@ -90,7 +90,7 @@ class Expression(metaclass=_Expression):
         self.arg_key: t.Optional[str] = None
         self.comments: t.Optional[t.List[str]] = None
         self._type: t.Optional[DataType] = None
-        self.meta: t.Optional[t.Dict[str, t.Any]] = {}
+        self._meta: t.Optional[t.Dict[str, t.Any]] = None
 
         for arg_key, value in self.args.items():
             self._set_parent(arg_key, value)
@@ -220,8 +220,11 @@ class Expression(metaclass=_Expression):
             dtype = DataType.build(dtype)
         self._type = dtype  # type: ignore
 
-    def set_metadata(self, meta_key, meta_value):
-        self.meta[meta_key] = meta_value
+    @property
+    def meta(self):
+        if self._meta is None:
+            self._meta = {}
+        return self._meta
 
     def __deepcopy__(self, memo):
         copy = self.__class__(**deepcopy(self.args))
@@ -232,7 +235,7 @@ class Expression(metaclass=_Expression):
             copy.type = self.type.copy()
 
         if self.meta is not None:
-            copy.meta = deepcopy(self.meta)
+            copy._meta = deepcopy(self.meta)
 
         return copy
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -231,7 +231,7 @@ class Expression(metaclass=_Expression):
         if self.comments is not None:
             copy.comments = deepcopy(self.comments)
 
-        if self.type is not None:
+        if self._type is not None:
             copy._type = self._type.copy()
 
         if self._meta is not None:
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -232,7 +232,7 @@ class Expression(metaclass=_Expression):
             copy.comments = deepcopy(self.comments)
 
         if self.type is not None:
-            copy.type = self.type.copy()
+            copy._type = self._type.copy()
 
         if self._meta is not None:
             copy._meta = deepcopy(self.meta)
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -227,7 +227,10 @@ class Expression(metaclass=_Expression):
         copy = self.__class__(**deepcopy(self.args))
         copy.comments = self.comments
         copy.type = self.type
-        copy._meta = self.metadata
+
+        if self.meta is not None:
+            copy.meta = deepcopy(self.meta)
+
         return copy
 
     def copy(self):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -225,8 +225,11 @@ class Expression(metaclass=_Expression):
 
     def __deepcopy__(self, memo):
         copy = self.__class__(**deepcopy(self.args))
-        copy.comments = self.comments
-        copy.type = self.type
+        if self.comments is not None:
+            copy.comments = deepcopy(self.comments)
+
+        if self.type is not None:
+            copy.type = self.type.copy()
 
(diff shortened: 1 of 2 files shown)


tobymao:
what is your actual use case for this?

sqlglot/expressions.py
```diff
@@ -219,10 +220,21 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
             dtype = DataType.build(dtype)
         self._type = dtype  # type: ignore
 
+    @property
+    def metadata(self) -> t.Optional[t.Dict[str, t.Any]]:
```
tobymao, line 224: rename to just def meta -> t.Dict[str, t.Any]

sqlglot/expressions.py
```diff
@@ -219,10 +220,21 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
...
+        return self._meta
```
tobymao, line 225: self._meta or {}

junpeng-jp:
> what is your actual use case for this?

We were discussing on providing more flexible metadata annotations that just the one provided by `anontate_types`. Discussion at this [slack thread](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1676899261518179)

sqlglot/expressions.py
tobymao: if you're going to keep it as an optional, there's no need for the instance var to be private

not exactly sure how you're planning to use it though, so i think having it default to always a dict is convenient

sqlglot/expressions.py
```diff
@@ -219,10 +220,21 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
...
+
+    def set_metadata(self, meta_key, meta_value):
```
tobymao, line 227: i don't really like this method, let's remove it, maybe you can just do this in the getter

@property
def meta(self) -> t.Dict[str, t.Any]:
   if self._meta is None:
       self._meta = {}
  return self._meta

then it's always available and you don't need this at all

tobymao:
> > what is your actual use case for this?
> 
> We were discussing on providing more flexible metadata annotations than just the one provided by `anontate_types`. Discussion at this [slack thread](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1676899261518179)

are you going to submit an actual use of this to sqlglot?

sqlglot/expressions.py
```diff
@@ -219,10 +220,21 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
...
+        if not self._meta:
+            self._meta = {}
+
+        self._meta[meta_key] = meta_value
+
     def __deepcopy__(self, memo):
         copy = self.__class__(**deepcopy(self.args))
         copy.comments = self.comments
         copy.type = self.type
+        copy._meta = self.metadata
```
tobymao, line 237: this doesn't perform a deep copy of metadata, i don't think this is correct

tobymao: now that i look at it comments is wrong too

if self._meta is not None:
    copy.meta = deecopy(self.meta) 

do something similar for copy.comments

GeorgeSittas: Yea that slipped, `type` is wrong too I think. We need to do `.copy()` if it's not `None`.

sqlglot/expressions.py
junpeng-jp: There was some discussion to lazily initialise this dictionary as it could be an empty dictionary majority of the time. It would only be populated when adding metadata

tobymao: it's similar in that it's only going to get set if you ask for metadata

junpeng-jp:
> > > what is your actual use case for this?
> > 
> > 
> > We were discussing on providing more flexible metadata annotations than just the one provided by `anontate_types`. Discussion at this [slack thread](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1676899261518179)
> 
> are you going to submit an actual use of this to sqlglot?

I'm marking this as draft as a potential idea. What is your recommendation on this use-case? I can maybe experiment on this first on my own as I would also need to modify things like the Schema & writing a more flexible annotation function

tobymao:
> > > > what is your actual use case for this?
> > > 
> > > 
> > > We were discussing on providing more flexible metadata annotations than just the one provided by `anontate_types`. Discussion at this [slack thread](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1676899261518179)
> > 
> > 
> > are you going to submit an actual use of this to sqlglot?
> 
> I'm marking this as draft as a potential idea. What is your recommendation on this use-case? I can maybe experiment on this first on my own as I would also need to modify things like the Schema & writing a more flexible annotation function

it's hard for me to gauge it without a concrete use case

tobymao:
it's innocuous enough though that we can get it in. so let's fix up the PR and merge it

sqlglot/expressions.py
```diff
@@ -82,15 +82,15 @@ class Expression(metaclass=_Expression):
...
+    __slots__ = ("args", "parent", "arg_key", "comments", "_type", "meta")
 
     def __init__(self, **args: t.Any):
         self.args: t.Dict[str, t.Any] = args
         self.parent: t.Optional[Expression] = None
         self.arg_key: t.Optional[str] = None
         self.comments: t.Optional[t.List[str]] = None
         self._type: t.Optional[DataType] = None
-        self._meta: t.Optional[t.Dict[str, t.Any]] = None
+        self.meta: t.Optional[t.Dict[str, t.Any]] = {}
```
GeorgeSittas, line 93: This should be `None` by default, and only set to a dictionary lazily if the user tries to access it. You need a `@property` that first checks if the internal field is `None`, in order to initialize it properly and return the dict.

GeorgeSittas (approved):

sqlglot/expressions.py
```diff
@@ -219,10 +220,23 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
...
         copy = self.__class__(**deepcopy(self.args))
-        copy.comments = self.comments
-        copy.type = self.type
+        if self.comments is not None:
+            copy.comments = deepcopy(self.comments)
+
+        if self.type is not None:
+            copy.type = self.type.copy()
+
+        if self.meta is not None:
```
tobymao, line 237:
```suggestion
        if self._meta is not None:
```

sqlglot/expressions.py
```diff
@@ -219,10 +220,23 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
...
+        return self._meta
+
     def __deepcopy__(self, memo):
         copy = self.__class__(**deepcopy(self.args))
-        copy.comments = self.comments
-        copy.type = self.type
+        if self.comments is not None:
+            copy.comments = deepcopy(self.comments)
+
+        if self.type is not None:
```
tobymao, line 234:
```suggestion
        if self._type is not None:
```

sqlglot/expressions.py
```diff
@@ -219,10 +220,23 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
...
+            copy.type = self.type.copy()
```
tobymao, line 235:
```suggestion
            copy._type = self._type.copy()
```

tobymao (approved):

PR #1211, opened 2023-02-21
Remove Teradata CREATE args from Snowflake create_sql

treysp:
#1205

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -329,7 +329,7 @@ class Snowflake(Dialect):
             return f"AUTOINCREMENT{start}{increment}"
 
         def create_sql(self, expression: exp.Create) -> str:
-            unsupported = ["set", "multiset", "global_temporary"]
+            unsupported = ["set", "multiset", "global_temporary", "volatile"]
             for prop in unsupported:
                 expression.set(prop, False)
             create_sql = super().create_sql(expression)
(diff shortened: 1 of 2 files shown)


treysp:
I've been meaning to pull most of those args into a new "properties" location so we can use the properties machinery and dialect-specific unsupported approach

sqlglot/dialects/snowflake.py
```diff
@@ -327,3 +327,10 @@ def generatedasidentitycolumnconstraint_sql(
             increment = expression.args.get("increment")
             increment = f" INCREMENT {increment}" if increment else ""
             return f"AUTOINCREMENT{start}{increment}"
+
+        def create_sql(self, expression: exp.Create) -> str:
```
tobymao, line 331: delete this and make a new propety location unsupported, then snowflake just needs to override the properties_location

sqlglot/dialects/snowflake.py
```diff
@@ -327,3 +327,10 @@ def generatedasidentitycolumnconstraint_sql(
...
+            unsupported = ["set", "multiset", "global_temporary", "volatile"]
+            for prop in unsupported:
+                expression.set(prop, False)
```
GeorgeSittas, line 334: We shouldn't mutate any expression in the generator. Whenever something like this is needed, you first need to copy it.

sqlglot/dialects/snowflake.py
```diff
@@ -327,3 +327,10 @@ def generatedasidentitycolumnconstraint_sql(
...
+            create_sql = super().create_sql(expression)
+            return create_sql
```
GeorgeSittas, line 336:
```suggestion
            return super().create_sql(expression)
```

PR #1213, opened 2023-02-22
Postgres INET type

joocer:
Basic implementation of Postgres INET type (see https://www.postgresql.org/docs/current/datatype-net-types.html#DATATYPE-INET)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -237,6 +237,7 @@ class Postgres(Dialect):
             "TEMP": TokenType.TEMPORARY,
             "UUID": TokenType.UUID,
             "CSTRING": TokenType.PSEUDO_TYPE,
+            "INET": TokenType.INET,
         }
 
         SINGLE_TOKENS = {
@@ -277,6 +278,7 @@ class Postgres(Dialect):
             exp.DataType.Type.BINARY: "BYTEA",
             exp.DataType.Type.VARBINARY: "BYTEA",
             exp.DataType.Type.DATETIME: "TIMESTAMP",
+            exp.DataType.Type.INET: "INET",
         }
 
         TRANSFORMS = {
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -237,7 +237,6 @@ class Postgres(Dialect):
             "TEMP": TokenType.TEMPORARY,
             "UUID": TokenType.UUID,
             "CSTRING": TokenType.PSEUDO_TYPE,
-            "INET": TokenType.INET,
         }
 
         SINGLE_TOKENS = {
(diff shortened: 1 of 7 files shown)


joocer:
It doesn't appear to be converting prefix syntax to CAST syntax:

`SELECT * FROM IPS WHERE IP = INET '127.0.0.1'` to `SELECT * FROM IPS WHERE IP = CAST('127.0.0.1')`

tests/dialects/test_postgres.py
```diff
@@ -370,6 +370,18 @@ def test_postgres(self):
...
+        self.validate_all(
+            "CAST('127.0.0.1/32' AS INET)",
+            read={
+                "postgres": "CAST('127.0.0.1/32' AS INET)",
+            },
+        )
+        self.validate_all(
+            "INET '127.0.0.1/32'",
+            read={
+                "postgres": "CAST('127.0.0.1/32' AS INET)",
```
tobymao, line 382: you have it inverted, this should be wrote

tobymao: you can combine these two tests into one

joocer: I see, the "read" is the input and we're testing the inputs to match the expected output, but we've declared the output first. It makes sense when you know that's what it's doing, but I can also see why I got it backwards. Thanks for the hint.

sqlglot/dialects/postgres.py
```diff
@@ -237,6 +237,7 @@ class Tokenizer(tokens.Tokenizer):
             "TEMP": TokenType.TEMPORARY,
             "UUID": TokenType.UUID,
             "CSTRING": TokenType.PSEUDO_TYPE,
+            "INET": TokenType.INET,
```
GeorgeSittas, line 240: This is unnecessary if you're going to also have it in the base tokenizer.

sqlglot/dialects/postgres.py
```diff
@@ -277,6 +278,7 @@ class Generator(generator.Generator):
             exp.DataType.Type.BINARY: "BYTEA",
             exp.DataType.Type.VARBINARY: "BYTEA",
             exp.DataType.Type.DATETIME: "TIMESTAMP",
+            exp.DataType.Type.INET: "INET",
```
GeorgeSittas, line 281: Ditto for the base generator.

joocer: Thanks George, I've taken the Postgres specific code out which included moving the unit test, I'm seeing something I wasn't expecting in the tests. Something I didn't see before updating the Postgres dialect code.

for the input `INET '127.0.0.1'` it translates to `CAST('127.0.0.1' AS INET)` as expected, but for the input `CAST('127.0.0.1' AS INET)` it translates to `TRY_CAST('127.0.0.1' AS INET)`. Is this expected behaviour from an implementation perspective? 

This appears to mean if I ran the output of a transpile through the engine again, I'd get a different response back.

GeorgeSittas: Yes, it's expected: depending on the source dialect, cast is parsed into either exp.Cast or exp.TryCast. There is a flag in the parser that determines whether the cast is to be parsed as strict or not.

GeorgeSittas: Also, forgot to mention: some dialects (e.g. postgres) override the SQL generation logic for TryCast so that it actually transpiles back to `CAST` for them (even though the AST node is TryCast).

GeorgeSittas (approved):

PR #1219, opened 2023-02-23
Refactor POST_CREATE and POST_EXPRESSION properties

treysp:
Convert Create args to properties in POST_CREATE location:
- set 
- multiset 
- global_temporary 
- volatile 
- temporary 
- transient 
- external 
- materialized 

Convert Create args to properties in POST_EXPRESSION location:
- data
- statistics
- no_primary_index

Add preserve_rows property in POST_EXPRESSION location for ON COMMIT PRESERVE ROWS #1205

Make SetProperty unsupported in Snowflake (Teradata "CREATE MULTISET TABLE a" -> Snowflake "CREATE TABLE a") #1205


treysp:
The Teradata syntax "CREATE VOLATILE TABLE" isn't supported in Snowflake, and the "obvious" was to handle is to set the volatility property to unsupported in Snowflake. However, the volatility property is also mapped to keyword IMMUTABLE (which snowflake does support).

You can see in https://github.com/tobymao/sqlglot/blob/b1434186b9097a3bdd9aa24c8f0aa0e6ba53c41c/sqlglot/parser.py#L536 that the VolatilityProperty is capturing all of deterministic, immutable, stable, and volatile.

Not sure the best way to handle this. We can pull them into separate properties, but that will complicate transpiling one of those words to another.

treysp:
I also see that the CREATE_TRANSIENT flag was essentially playing the role of UNSUPPORTED properties for the Transient property in POST_CREATE location. 

I don't see this being used in any other dialects, so I'll remove it when responding to review comments.

https://github.com/tobymao/sqlglot/blob/b1434186b9097a3bdd9aa24c8f0aa0e6ba53c41c/sqlglot/generator.py#L90
https://github.com/tobymao/sqlglot/blob/b1434186b9097a3bdd9aa24c8f0aa0e6ba53c41c/sqlglot/dialects/snowflake.py#L218

sqlglot/dialects/spark.py
```diff
@@ -8,7 +8,13 @@
...
-    temporary = e.args.get("temporary")
+    properties = e.args.get("properties")
+    if properties:
+        temporary = any(
+            [isinstance(prop, exp.TemporaryProperty) for prop in properties.expressions]
+        )
+    else:
+        temporary = None
 
     if kind.upper() == "TABLE" and temporary is True:
```
tobymao, line 19: if kind.upper() == "TABLE" and any(isinstance(prop, Temp) for prop in e.args.get("properties", []))

sqlglot/parser.py
```diff
@@ -1104,21 +1110,13 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
...
-            temporary=temporary,
-            transient=transient,
-            external=external,
             replace=replace,
-            unique=unique,
-            materialized=materialized,
             data=data,
             statistics=statistics,
             no_primary_index=no_primary_index,
+            preserve_rows=preserve_rows,
```
tobymao, line 1119: why isn't preserve rows also in properties?

GeorgeSittas:
> The Teradata syntax "CREATE VOLATILE TABLE" isn't supported in Snowflake, and the "obvious" way to handle is to set the volatility property to unsupported in Snowflake. However, the volatility property is also mapped to keyword IMMUTABLE (which snowflake does support).
> 
> You can see in
> 
> https://github.com/tobymao/sqlglot/blob/b1434186b9097a3bdd9aa24c8f0aa0e6ba53c41c/sqlglot/parser.py#L536
> that the VolatilityProperty is capturing all of deterministic, immutable, stable, and volatile.
> 
> Not sure the best way to handle this. We can pull them into separate properties, but that will complicate transpiling one of those words to another.

Maybe you can also check if the string captured in `exp.VolatilityProperty` is legal in Snowflake. So, for instance, you'd only allow `IMMUTABLE` if that's the only valid one.

sqlglot/parser.py
treysp: Update moves all 4 of those into a new properties location

sqlglot/dialects/spark.py
treysp: What's the correct way to iterate over properties.expressions in this syntax?

`any(isinstance(prop, Temp) for prop in e.args.get("properties", []))`

GeorgeSittas: Alternatively, you could also do something like:
```python
properties = e.args.get("properties")
any(isinstance(prop, Temp) for prop in (properties.expressions if properties else []))
```

tobymao: if this is common make a method on the instance

treysp:
> > The Teradata syntax "CREATE VOLATILE TABLE" isn't supported in Snowflake, and the "obvious" way to handle is to set the volatility property to unsupported in Snowflake. However, the volatility property is also mapped to keyword IMMUTABLE (which snowflake does support).
> > You can see in
> > https://github.com/tobymao/sqlglot/blob/b1434186b9097a3bdd9aa24c8f0aa0e6ba53c41c/sqlglot/parser.py#L536
> > 
> > that the VolatilityProperty is capturing all of deterministic, immutable, stable, and volatile.
> > Not sure the best way to handle this. We can pull them into separate properties, but that will complicate transpiling one of those words to another.
> 
> Maybe you can also check if the string captured in `exp.VolatilityProperty` is legal in Snowflake. So, for instance, you'd only allow `IMMUTABLE` if that's the only valid one.

That makes sense, but I don't think it would work if multiple volatility words are used for different things in the same dialect and/or word usages aren't aligned across dialects. (So in one dialect VOLATILE means X and IMMUTABLE means Y, but in another dialect VOLATILE means Y and IMMUTABLE means X.)

I think resolving this will probably require reviewing volatility-related words across the dialects. I'm happy to do that, but it may take me a bit to get to it.

sqlglot/parser.py
```diff
@@ -1398,6 +1389,29 @@ def _parse_partitioned_by(self) -> exp.Expression:
...
+
+    def _parse_noprimaryindex(self) -> exp.Expression:
+        self._match_text_seq("PRIMARY", "INDEX")
+        return exp.NoPrimaryIndexProperty()
+
+    def _parse_oncommit(self) -> exp.Expression:
+        self._match_text_seq("COMMIT", "PRESERVE", "ROWS")
+        return exp.OnCommitProperty()
+
+    # def _parse_unique_property(self) -> exp.Expression:
```
tobymao, line 1410: delete this

PR #1222, opened 2023-02-23
Expression.root

barakalon:


barakalon:
Not sure if this is worth it... but curious what yall think

tobymao (approved):
what do you need this for?

barakalon:
Crawl up to the top of a tree and add a CTE

PR #1230, opened 2023-02-27
parse create database

tekumara:
basic parsing of CREATE DATABASE in the same way as CREATE SCHEMA


sqlglot/parser.py
```diff
@@ -1032,6 +1034,7 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
         elif create_token.token_type == TokenType.INDEX:
             this = self._parse_index()
         elif create_token.token_type in (
```
GeorgeSittas, line 1036: Maybe we should factor this out into a class variable.

GeorgeSittas (approved):

sqlglot/parser.py
tobymao: 52b32261

PR #1233, opened 2023-02-28
Teradata partition by RANGE_N function

treysp:
- Add Teradata function RANGE_N
- Update create_sql to allow POST_INDEX properties even if no indexes are specified 
    - In Teradata PARTITION BY goes after a primary index if one is specified, before non-primary indexes if no primary index is specified, and is allowed without any indexes at all


sqlglot/dialects/teradata.py
```diff
@@ -136,3 +150,16 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             type_sql = super().datatype_sql(expression)
             prefix_sql = expression.args.get("prefix")
             return f"SYSUDTLIB.{type_sql}" if prefix_sql else type_sql
+
+        def rangen_sql(self, expression: exp.RangeN) -> str:
+            this = self.sql(expression, "this")
+            expressions_sql = ", ".join(
+                [self.sql(e) for e in expression.args.get("expressions", [])]
+            )
```
GeorgeSittas, line 158:
```suggestion
            expressions_sql = self.expressions(expression)
```

sqlglot/dialects/teradata.py
```diff
@@ -136,3 +150,16 @@ def datatype_sql(self, expression: exp.DataType) -> str:
...
+            each_sql = (
+                f" EACH {self.sql(expression.args.get('each'))}"
+                if expression.args.get("each")
+                else ""
+            )
```
GeorgeSittas, line 163:
```suggestion
            each_sql = self.sql(expression, "each")
            each_sql = f" EACH {each_sql}" if each_sql else ""
```

sqlglot/generator.py
```diff
@@ -564,21 +571,18 @@ def create_sql(self, expression: exp.Create) -> str:
...
-                indexes_sql.append(
-                    f"{ind_unique}{ind_primary}{ind_amp} INDEX{ind_name}{ind_columns}"
-                )
+                if len(indexes_sql) == 0:
+                    if index.args.get("primary"):
+                        indexes_sql.append(f"{ind_sql}{postindex_props_sql}")
+                    else:
+                        indexes_sql.append(f"{postindex_props_sql}{ind_sql}")
+                else:
+                    indexes_sql.append(ind_sql)
```
GeorgeSittas, line 582: What about checking `if indexes_sql` first to avoid `len` altogether? Seems fine either way, although ideally we'd like to simplify this a bit, if possible.

GeorgeSittas (commented):
Left a few comments/suggestions, but looks good otherwise.

sqlglot/tokens.py
```diff
@@ -252,6 +252,7 @@ class TokenType(AutoName):
     QUALIFY = auto()
     QUOTE = auto()
     RANGE = auto()
+    RANGE_N = auto()
```
tobymao, line 255: why did you need to add this? i don't think it's needed

sqlglot/parser.py
```diff
@@ -1384,11 +1384,14 @@ def _parse_partition_by(self) -> t.List[t.Optional[exp.Expression]]:
 
     def _parse_partitioned_by(self) -> exp.Expression:
         self._match(TokenType.EQ)
-        return self.expression(
-            exp.PartitionedByProperty,
-            this=self._parse_schema() or self._parse_bracket(self._parse_field()),
+        this = (
+            self._parse_schema()
+            or self._parse_function()
```
tobymao, line 1389: are you sure this needs to be changed? _parse_field parses function already

sqlglot/tokens.py
treysp: you're right

sqlglot/dialects/teradata.py
```diff
@@ -105,6 +106,19 @@ def _parse_update(self) -> exp.Expression:
...
             )
 
+        def _parse_rangen(self):
+            this = self._parse_id_var()
+            self._match(TokenType.BETWEEN)
+
+            expressions = []
+            while self._next and not self._match_text_seq("EACH"):
+                self._match(TokenType.COMMA)
+                expressions.append(self._parse_conjunction() or self._parse_expression())
```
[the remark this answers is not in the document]
GeorgeSittas: > EACH allows date intervals like "INTERVAL '1' MONTH", but since there's no alias allowed _parse_conjunction should work. I'll update that and the expressions.

Maybe even `_parse_bitwise` is just fine, although I don't remember the exact call sequence. I always look it up if needed.

> I assume I should choose the parser as far down the list as possible?

That's correct. In that way, you reduce function call overhead and possible errors that might result from "overshooting" with regard to what you're able to parse at a given step. If you just want an id var, it'd be a waste to call `_parse_expression`, because that'd lead to many unnecessary function calls that will probably lead to `_parse_id_var` anyway.

GeorgeSittas (approved):
Much better 👍

sqlglot/generator.py
```diff
@@ -564,21 +571,20 @@ def create_sql(self, expression: exp.Create) -> str:
...
+                ind_sql = f"{ind_unique}{ind_primary}{ind_amp} INDEX{ind_name}{ind_columns}"
+
+                if indexes_sql:
+                    indexes_sql.append(ind_sql)
+                else:
+                    indexes_sql.append(
+                        f"{ind_sql}{postindex_props_sql}"
+                        if index.args.get("primary")
+                        else f"{postindex_props_sql}{ind_sql}"
                     )
```
GeorgeSittas, line 583: An idea to further simplify this:
```python
ind_sql = f"{ind_unique}{ind_primary}{ind_amp} INDEX{ind_name}{ind_columns}"
ind_sql = f"{ind_sql}{postindex_props_sql}" if index.args.get("primary") else f"{postindex_props_sql}{ind_sql}"
indexes_sql.append(ind_sql)
```

GeorgeSittas: Scratch that, I misread a part of the code. This won't work.

treysp: Wouldn't this add in postindex_props_sql every time through the loop instead of just the first?

PR #1236, opened 2023-03-01
refactor: return more precise types from find methods

tekumara:
find methods now return the type of the expression being sought (or a union for multiple types), rather then `Any`.

this means the caller doesn't need to cast the result.

the increased precision revealed some mypy type errors, which I've fixed

### DIFF
--- a/sqlglot/dataframe/sql/dataframe.py
+++ b/sqlglot/dataframe/sql/dataframe.py
@@ -260,7 +260,9 @@ class DataFrame:
     @classmethod
     def _get_outer_select_columns(cls, item: t.Union[exp.Expression, DataFrame]) -> t.List[Column]:
         expression = item.expression if isinstance(item, DataFrame) else item
-        return [Column(x) for x in expression.find(exp.Select).expressions]
+        select = expression.find(exp.Select)
+        assert select, "No select expression found"
+        return [Column(x) for x in select.expressions]
 
     @classmethod
     def _create_hash_from_expression(cls, expression: exp.Select):
--- a/sqlglot/dataframe/sql/dataframe.py
+++ b/sqlglot/dataframe/sql/dataframe.py
@@ -260,9 +260,7 @@ class DataFrame:
     @classmethod
     def _get_outer_select_columns(cls, item: t.Union[exp.Expression, DataFrame]) -> t.List[Column]:
         expression = item.expression if isinstance(item, DataFrame) else item
-        select = expression.find(exp.Select)
-        assert select, "No select expression found"
-        return [Column(x) for x in select.expressions]
+        return [Column(x) for x in (expression.find(exp.Select) or exp.Select()).expressions]
 
     @classmethod
     def _create_hash_from_expression(cls, expression: exp.Select):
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -361,7 +361,7 @@ def create_with_partitions_sql(self: Generator, expression: exp.Create) -> str:
         expression = expression.copy()
         prop = expression.find(exp.PartitionedByProperty)
         this = prop and prop.this
-        if prop and not isinstance(this, exp.Schema):
+        if prop and this and not isinstance(this, exp.Schema):
```
tekumara, line 364: handle `this` being `None`

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -260,7 +260,9 @@ def _add_ctes_to_expression(cls, expression: exp.Select, ctes: t.List[exp.CTE])
     @classmethod
     def _get_outer_select_columns(cls, item: t.Union[exp.Expression, DataFrame]) -> t.List[Column]:
         expression = item.expression if isinstance(item, DataFrame) else item
-        return [Column(x) for x in expression.find(exp.Select).expressions]
+        select = expression.find(exp.Select)
+        assert select, "No select expression found"
```
tekumara, line 264: handle `select` being `None`

sqlglot/dialects/dialect.py
tobymao: this one is not necessary, line 363 indicates that this is not None 

this = prop and prop.this, so this must not be None

sqlglot/expressions.py
```diff
@@ -306,7 +308,7 @@ def find(self, *expression_types, bfs=True):
         """
         return next(self.find_all(*expression_types, bfs=bfs), None)
 
-    def find_all(self, *expression_types, bfs=True):
+    def find_all(self, *expression_types: t.Type[E], bfs=True) -> t.Generator[E, None, None]:
```
tobymao, line 311: t.Iterable[E]

sqlglot/dataframe/sql/dataframe.py
tobymao: you can just do

return [Column(x) for x in (expression.find(exp.Select) or exp.Select().expressions)]

sqlglot/expressions.py
tekumara: 👍 i can make the return type `t.Iterator[E]` ... `t.Iterable[E]` will fail to type check because an iterator doesn't support `next`, which is used above on line 309

sqlglot/dataframe/sql/dataframe.py
tekumara: this will fail to type check because `expression.find(exp.Select)` is not iterable

tobymao: sorry, i over complicated it just do

for x in expression.find(exp.Select) or []

sqlglot/dialects/dialect.py
tekumara: `prop.this` can return `None` though

tobymao: this code is a bit confusing in general

maybe you should remove line 363

and do

if prop and not isinstance(prop, exp.Schema)

tekumara: 👍 good point, i've removed line 363, and updated the line below that used `this` ... I still need to check `prop.this` is not `None` before its used below... have a look at the latest commit and see what you think

tobymao (approved):

PR #1244, opened 2023-03-03
Improve parsing of DIV0 and ZEROIFNULL for snowflake

### LINKED ISSUE #1243: Missing snowflake functions
Transpiling the follow query from `Snowflake` to `SQLite` returns the same query even though `div0` and `nullifzero` both are unavailable in `SQLite`.

Input
```
sql_query = """
select
    div0(foo,bar) as foobar,
    zeroifnull(baz) as baz
"""
sqlglot.transpile(sql_query, read="snowflake", write="sqlite")
```
Output
`['SELECT DIV0(foo, bar) AS foobar, ZEROIFNULL(baz) AS baz']
`

GeorgeSittas:
Fixes #1243

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -106,6 +106,20 @@ def _parse_date_part(self):
     return self.expression(exp.Extract, this=this, expression=expression)
 
 
+# https://docs.snowflake.com/en/sql-reference/functions/div0
+def _div0_to_if(args):
+    cond = exp.EQ(this=seq_get(args, 1), expression=exp.Literal.number(0))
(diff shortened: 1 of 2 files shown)


GeorgeSittas:
I think that we also need to change how we parse `a / b` for Snowflake, because it defaults to float division. A possible solution might be to multiply the divident or the divisor by `1.0` so that it coerces the expression type implicitly in other dialects.

For example, `1/2` yields `0.5` in Snowflake but `0` in SQLite, whereas `1.0/2` yields `0.5` for both dialects.

tobymao:
interesting, postgres also does 1 / 2 = 0

tobymao (approved):

PR #1245, opened 2023-03-03
Fix division operator transpilation across dialects

GeorgeSittas:


sqlglot/generator.py
```diff
@@ -1743,17 +1746,28 @@ def respectnulls_sql(self, expression: exp.RespectNulls) -> str:
...
+
+        if not self.INTEGER_DIVISION:
+            return self.sql(exp.Cast(this=div, to=exp.DataType.build("INT")))
+
+        return div
+
+    def floatdiv_sql(self, expression: exp.FloatDiv) -> str:
+        if self.INTEGER_DIVISION:
+            # Multiply divident by 1.0 so that the expression's type can be coerced to float
+            this = exp.Mul(this=expression.this, expression=exp.Literal.number(1.0))
```
tobymao, line 1766: let's do a cast instead

GeorgeSittas: donezo

sqlglot/generator.py
```diff
@@ -1762,11 +1762,8 @@ def div_sql(self, expression: exp.Div) -> str:
 
     def floatdiv_sql(self, expression: exp.FloatDiv) -> str:
         if self.INTEGER_DIVISION:
-            # Multiply divident by 1.0 so that the expression's type can be coerced to float
-            this = exp.Mul(this=expression.this, expression=exp.Literal.number(1.0))
-            return self.div_sql(
-                exp.Div(this=exp.Paren(this=this), expression=expression.expression)
-            )
+            this = exp.Cast(this=expression.this, to=exp.DataType.build("DOUBLE"))
+            return self.div_sql(exp.Div(this=this, expression=expression.expression))
```
tobymao, line 1766: this=exp.cast(expression.this, "double")

GeorgeSittas: Ah I forgot we added that, thanks.

GeorgeSittas: Discussed in slack: keeping it as is because for the cast to int right above this we can't use `exp.cast`, due to infinite recursion problems.

sqlglot/parser.py
```diff
@@ -326,7 +332,6 @@ class Parser(metaclass=_Parser):
     FACTOR = {
         TokenType.DIV: exp.IntDiv,
         TokenType.LR_ARROW: exp.Distance,
-        TokenType.SLASH: exp.Div,
```
tobymao, line 329: can we leave this for completeness and only override it if it's floatdiv?

GeorgeSittas: sure

tobymao (approved):

Issue #1225, opened 2023-02-24
SET commands not decomposed


joocer:
Parsing a `SET` command such as 

~~~sql
SET @q = 4;
~~~

results in this:
_(note, the equals expression is just presented as a string literal)_

~~~
(COMMAND this: SET, expression: 
  (LITERAL this: @q = 4, is_string: True))
~~~

I had expected the statement to be decomposed into something like this:
_(note, the identifier and literal are different nodes in the tree)_
~~~
(COMMAND this: SET, expression: 
  ((IDENTIFIER this: @q, quoted: False)
  (LITERAL this: 4, is_string: False)))
~~~

tobymao:
many command in SQL haven't been high priority and so are parsed in the catchall "command" node.

basic parsing of set should be easy with just doing parse_conjunction(), but i'm not sure how SET is used in all dialects.

joocer:
Thanks @tobymao,

Looking at another non-SELECT example
~~~sql
SHOW FULL COLUMNS FROM $astronauts"
~~~

I get a similar, COMMAND node with the first keyword and then the rest of the statement as a single literal. I'm guessing from the name only that `parse_conjuction()` is extracting predicate expressions (which `@q = 4` looks like) but won't parse `SHOW` statements?

tobymao:
show is an entirely different command with it's own syntax

tobymao:
1. remove TokenType.SET from Tokenizer.COMMANDS
2. add it to Parser.STATEMENT_PARSERS
3. parser func create parse_set
4. create a SetStatement node in expression.py
5. return a self.expression(exp.SetStatement, this=self._parse_conjunction())
6. write unit tests

please do some research on various major dialects to see what is possible

spark, snowflake, bigquery, mysql, postgres

have an understanding of what needs to be supported in the SET command, this will dictate what step 5 actually does.

joocer:
Initial reconnaissance of documentation:

[spark](https://spark.apache.org/docs/latest/sql-ref-syntax-aux-conf-mgmt-set.html)
[snowflake](https://docs.snowflake.com/en/sql-reference/sql/set)
[bigquery](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#set)
[MySQL](https://dev.mysql.com/doc/refman/8.0/en/set-statement.html)
[postgres](https://www.postgresql.org/docs/current/sql-set.html)
[trino](https://trino.io/docs/current/sql.html)

Initial Observations:
[...]

PR #1250, opened 2023-03-04
#1225 - SET commands not decomposed

joocer:
whilst I write a set of test cases, can you review and approach please?


sqlglot/expressions.py
```diff
@@ -818,7 +818,10 @@ class SetItem(Expression):
         "expressions": False,
         "kind": False,
         "collate": False,  # MySQL SET NAMES statement
-        "global": False,
+        "global_": False,
```
tobymao, line 821: just call it global since it's a dictionary, when you set it you can use a dict, this is how it's done for "with"

sqlglot/generator.py
```diff
@@ -1365,6 +1365,19 @@ def qualify_sql(self, expression: exp.Qualify) -> str:
         this = self.indent(self.sql(expression, "this"))
         return f"{self.seg('QUALIFY')}{self.sep()}{this}"
 
+    def setitem_sql(self, expression):
+        print(dir(expression.this.left))
```
tobymao, line 1369: don't forget to remove these

sqlglot/tokens.py
```diff
@@ -701,6 +704,10 @@ class Tokenizer(metaclass=_Tokenizer):
         "PREPARE": TokenType.COMMAND,
         "TRUNCATE": TokenType.COMMAND,
         "VACUUM": TokenType.COMMAND,
+        "GLOBAL": TokenType.GLOBAL,
```
tobymao, line 707: delete all of this, it's not needed, just use match_text_seq

sqlglot/tokens.py
```diff
@@ -303,6 +303,9 @@ class TokenType(AutoName):
     WITHOUT_TIME_ZONE = auto()
     UNIQUE = auto()
 
+    SESSION = auto()
```
tobymao, line 306: delete this

sqlglot/parser.py
```diff
@@ -3829,8 +3826,42 @@ def _parse_merge(self) -> exp.Expression:
             expressions=whens,
         )
 
-    def _parse_set(self) -> exp.Expression:
-        return self.expression(exp.Set, expressions=self._parse_csv(self._parse_set_item))
+    def _parse_set(self):
+        start = self._prev
+
+        local = self._match(TokenType.LOCAL) is not None
```
tobymao, line 3832: local = not self._match_text_seq("LOCAL")

etc

sqlglot/parser.py
```diff
@@ -3829,8 +3826,42 @@ def _parse_merge(self) -> exp.Expression:
...
+        global_ = self._match(TokenType.GLOBAL) is not None
+        session = self._match(TokenType.SESSION) is not None
+
```
tobymao, line 3835: expression = self._parse_equality() , let's handle the TO later, but this is not the way, i'll have to fiddle with it

sqlglot/parser.py
```diff
@@ -3829,8 +3826,42 @@ def _parse_merge(self) -> exp.Expression:
...
+            expression=right,
+        )
+
+        return self.expression(
+            exp.SetItem,
+            this=this,
+            local=local,
+            global_=global_,
+            session=session,
+            assignment_operator=assignment_operator
```
tobymao, line 3863: delete assignment_operator, we can always just do equality

sqlglot/expressions.py
joocer: I've done a text search for `with=` and `with =` to find how the reserved word has been used as a variable name and can't find it.

It's not a clash with a reserved word here, but it is in parser.py (about line 3861) where the keys in the dict defined here are expanded to be named arguments to the expression constructor.

sqlglot/generator.py
joocer: thanks

sqlglot/expressions.py
tobymao: you can’t use kwargs, you need to use a dict

joocer: yeah that would do it

joocer:
@tobymao, I'm getting an error that suggests that the base SET handler is running even though there is a specific MySQL handler written. I had assumed that the dialect's handler would supersede the base handler.

Am I interpreting the error correctly and have I missed doing something to allow/enable inheritance?

tobymao:
go ahead and pay your latest and we can help debug

joocer:
I've pushed the unit tests I've written - this is the logs of GHA running the tests:
https://github.com/tobymao/sqlglot/actions/runs/4394691961/jobs/7695885692

This is the error I believe is caused by the MySQL `SET` handler code not running:
> AssertionError: 'SET @var_name = expr' != '@var_name = expr'

PR #1255, opened 2023-03-06
Add support for DuckDB/Snowflake SAMPLE clauses

### LINKED ISSUE #1247: DuckDB sample syntax fails to parse
All of [the examples](https://duckdb.org/docs/sql/samples.html) fail to parse:

```
In [10]: import sqlglot as sg

In [11]: sql = """SELECT * FROM tbl USING SAMPLE 5
    ...: SELECT * FROM tbl USING SAMPLE 10%
    ...: SELECT * FROM tbl USING SAMPLE 10 PERCENT (bernoulli)
    ...: SELECT * FROM tbl USING SAMPLE reservoir(50 ROWS) REPEATABLE (100)
    ...: SELECT * FROM tbl USING SAMPLE 10% (system, 377)
    ...: SELECT * FROM tbl TABLESAMPLE RESERVOIR(20%), tbl2 WHERE tbl.i=tbl2.i
    ...: SELECT * FROM tbl, tbl2 WHERE tbl.i=tbl2.i USING SAMPLE RESERVOIR(20%)""".splitlines()

In [12]: nstmts = len(sql)

In [13]: nfailed = 0

In [14]: for line in sql:
    ...:     try:
    ...:         sg.parse_one(line, read="duckdb")
    ...:     except sg.ParseError as e:
    ...:         print(e)
    ...:         nfailed += 1
Invalid expression / Unexpected token. Line 1, Col: 19.
  SELECT * FROM tbl USING SAMPLE 5
Invalid expression / Unexpected token. Line 1, Col: 19.
  SELECT * FROM tbl USING SAMPLE 10%
Invalid expression / Unexpected token. Line 1, Col: 19.
  SELECT * FROM tbl USING SAMPLE 10 PERCENT (bernoulli)
Invalid expression / Unexpected token. Line 1, Col: 19.
  SELECT * FROM tbl USING SAMPLE reservoir(50 ROWS) REPEATABLE (100)
Invalid expression / Unexpected token. Line 1, Col: 19.
  SELECT * FROM tbl USING SAMPLE 10% (system, 377)
Expecting ). Line 1, Col: 43.
  SELECT * FROM tbl TABLESAMPLE RESERVOIR(20%), tbl2 WHERE tbl.i=tbl2.i
Invalid expression / Unexpected token. Line 1, Col: 44.
  SELECT * FROM tbl, tbl2 WHERE tbl.i=tbl2.i USING SAMPLE RESERVOIR(20%)

In [15]: nfailed == nstmts
Out[15]: True
```

GeorgeSittas:
Fixes #1247 

References:
- https://docs.snowflake.com/en/sql-reference/constructs/sample#examples
- https://duckdb.org/docs/sql/samples.html#examples


sqlglot/tokens.py
```diff
@@ -290,6 +290,7 @@ class TokenType(AutoName):
     UPDATE = auto()
     USE = auto()
     USING = auto()
+    USING_SAMPLE = auto()
```
tobymao, line 293: i don't think you need this

tests/dialects/test_hive.py
```diff
@@ -626,25 +626,25 @@ def test_hive(self):
             },
         )
         self.validate_all(
-            "SELECT * FROM x TABLESAMPLE(1) AS foo",
+            "SELECT * FROM x TABLESAMPLE (1) AS foo",
```
tobymao, line 629: why add a space? it's a func isn't it

sqlglot/parser.py
```diff
@@ -3503,8 +3512,14 @@ def _parse_identifier(self) -> t.Optional[exp.Expression]:
...
 
-    def _parse_var(self, any_token: bool = False) -> t.Optional[exp.Expression]:
-        if (any_token and self._advance_any()) or self._match(TokenType.VAR):
+    def _parse_var(
+        self, any_token: bool = False, tokens: t.Optional[t.Collection[TokenType]] = None
+    ) -> t.Optional[exp.Expression]:
+        if (
+            (any_token and self._advance_any())
+            or self._match(TokenType.VAR)
+            or self._match_set(tokens or {})
```
tobymao, line 3521: self._match_set(tokens) if tokens else False

tests/dialects/test_hive.py
GeorgeSittas: I saw some examples, e.g. in Snowflake, where they used spaces and thought it'd be ok:
```sql
SELECT * FROM testtable TABLESAMPLE (100);  -- Snowflake example
```
Another reason is that before we'd get a space when `method != ""`, which was inconsistent to the no-space function-style formatting in case no method was passed. Examples from our tests before this PR:

```python
SELECT a FROM test TABLESAMPLE BERNOULLI (50) -- Space between BERNOULLI and (50)
SELECT a FROM test TABLESAMPLE(100) -- No space here, inconsistent with above
```

This also works well with DuckDB's syntax for `USING SAMPLE`.

sqlglot/tokens.py
GeorgeSittas: Two reasons for adding this:

- Easily exclude `USING` from the table alias set
- I wanted to attach the using sample expression in the query modifiers list, because it's supposed to be applied to the resulting set instead of a single table. This is why I added the `as_modifier` flag and checked if we can match USING SAMPLE.

sqlglot/parser.py
```diff
@@ -2164,11 +2166,11 @@ def _parse_derived_table_values(self) -> t.Optional[exp.Expression]:
 
         return self.expression(exp.Values, expressions=expressions, alias=self._parse_table_alias())
 
-    def _parse_table_sample(self) -> t.Optional[exp.Expression]:
+    def _parse_table_sample(self, as_modifier: bool = False) -> t.Optional[exp.Expression]:
         if not self._match(TokenType.TABLE_SAMPLE):
-            return None
+            if not (as_modifier and self._match(TokenType.USING_SAMPLE)):
```
tobymao, line 2171: move this up one line?into the same table sample

sqlglot/tokens.py
tobymao: using is already not allowed in table alias, so you can just exclude that

i don't understand the as_modifier flag

GeorgeSittas: > using is already not allowed in table alias, so you can just exclude that

Ah, I missed this because it's not in `ID_VAR_TOKENS` either and so thought we weren't excluding it.

> i don't understand the as_modifier flag

If we just mapped `USING SAMPLE` to `TokenType.TABLE_SAMPLE`, then the former would be attached to the table, instead of being a query modifier. From an AST perspective this is not correct, which is why I introduced the `as_modifier` flag. Does it make sense?

tobymao: well you shouldn't map using sample, but you can just to match_text right?

GeorgeSittas:
I think most issues have been addressed, let me know if I missed something.

sqlglot/parser.py
```diff
@@ -2177,27 +2179,33 @@ def _parse_table_sample(self) -> t.Optional[exp.Expression]:
         size = None
         seed = None
 
-        self._match_l_paren()
+        kind = "TABLESAMPLE" if self._prev.token_type == TokenType.TABLE_SAMPLE else "USING SAMPLE"
```
GeorgeSittas, line 2182: Maybe we should be using a boolean flag here instead..

tobymao: not sure

tobymao (approved):

Issue #1259, opened 2023-03-07
Number of improvements can be made to Snowflake parsing


cillianod-wf:
I've been playing around with the transpiling of Snowflake to SQLite and have noticed a number of features which are missing. I've split them up into different categories:

**Datetime functions**
- [x] [datediff()](https://docs.snowflake.com/en/sql-reference/functions/datediff) https://github.com/tobymao/sqlglot/commit/313f22486949ad806e2ce72941c4c2514e5ff6ce
- [x] [current_date()](https://docs.snowflake.com/en/sql-reference/functions/current_date) https://github.com/tobymao/sqlglot/commit/3871647e60a33212b565245cbd73e8963e377d71
- [x] `cast(datetime as date)`, when this gets transpiled it remains as is, the behaviour between Snowflake and SQLite is different however, it returns as follows - Snowflake=`yyyy-mm-dd`, SQLite=`yyyy` 

**Transformations**
- [x] [least()](https://docs.snowflake.com/en/sql-reference/functions/least) https://github.com/tobymao/sqlglot/commit/1e5ce4629bd49f43c45133fe1f755a49460f4f5a
- [x] [square()](https://docs.snowflake.com/en/sql-reference/functions/square) https://github.com/tobymao/sqlglot/commit/9f5c943fab01ddd50c4e0f07e25ca3a244efef65
- [x] [to_char()/to_varchar()](https://docs.snowflake.com/en/sql-reference/functions/to_char) https://github.com/tobymao/sqlglot/commit/8905eaa9658a73dccb088f65e5585d95bd1abcae (transpiling "format" for dialects that don't support this function not planned)
- [x] [nullifzero()](https://docs.snowflake.com/en/sql-reference/functions/nullifzero) https://github.com/tobymao/sqlglot/commit/7480d8abd7ec23a9552e5066fbcaaf2ebbcb1082

**Aggregations**
- [x] [boolor_agg()](https://docs.snowflake.com/en/sql-reference/functions/boolor_agg) https://github.com/tobymao/sqlglot/commit/662feeed1c6b68558c92b0d2bb9df34fa70413ac
- [x] [count_if()](https://docs.snowflake.com/en/sql-reference/functions/count_if) https://github.com/tobymao/sqlglot/commit/978741651a84b1aa5fb2615c1d232afe6856f305

**SQL Statements**
[...]

PR #1260, opened 2023-03-07
Improve Snowflake parsing of nullizero

cillianod-wf:
Part of [Issue 1259](https://github.com/tobymao/sqlglot/issues/1259)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -120,6 +120,12 @@ def _zeroifnull_to_if(args):
     return exp.If(this=cond, true=exp.Literal.number(0), false=seq_get(args, 0))
 
 
+# https://docs.snowflake.com/en/sql-reference/functions/zeroifnull
+def _nullifzero_to_if(args):
+    cond = exp.EQ(this=seq_get(args, 0), expression=exp.Literal.number(0))
+    return exp.If(this=cond, true=exp.Null(), false=seq_get(args, 0))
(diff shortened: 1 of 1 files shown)


GeorgeSittas (approved):
LGTM

Can you please add some tests for both `NULLIFZERO` and `ZEROIFNULL` functions? Seems like I forgot to test the latter in my recent PR..

tests/dialects/test_snowflake.py
```diff
@@ -62,6 +62,28 @@ def test_snowflake(self):
...
         )
+        self.validate_all(
+            "ZEROIFNULL(foo)",
+            write={
+                "snowflake": "IFF(foo IS NULL, 0, foo)",
+                "sqlite": "CASE WHEN foo IS NULL THEN 0 ELSE foo END",
+                "presto": "IF(foo IS NULL, 0, foo)",
+                "spark": "IF(foo IS NULL, 0, foo)",
+                "hive": "IF(foo IS NULL, 0, foo)",
+                "duckdb": "CASE WHEN foo IS NULL THEN 0 ELSE foo END",
```
GeorgeSittas, line 73: You should either explicitly create an `IS` expression in `_nullifzero_to_if` or use `=` here instead.

GeorgeSittas:
I'll just fix the zeroifnull issue and then it's good to go. Seems like I messed up with the NULL comparison.

Thanks for the PR!

PR #1261, opened 2023-03-07
Fix: transpile LEAST, MIN correctly (SQLite, Snowflake)

GeorgeSittas:
- https://docs.snowflake.com/en/sql-reference/functions/least
- https://sqlite.org/lang_corefunc.html#min_scalar


sqlglot/dialects/sqlite.py
```diff
@@ -94,9 +90,17 @@ class Generator(generator.Generator):
...
         }
 
+        def fetch_sql(self, expression):
+            return self.limit_sql(exp.Limit(expression=expression.args.get("count")))
+
+        def least_sql(self, expression):
+            if len(expression.expressions) > 1:
+                return rename_func("MIN")(self, expression)
+
+            return self.expressions(expression)
```
GeorgeSittas, line 102: `MIN` with one argument in sqlite is interpreted as an aggregate function instead of a scalar one, which is why I took separate cases here.

sqlglot/dialects/snowflake.py
```diff
@@ -324,6 +324,10 @@ def intersect_op(self, expression):
                 self.unsupported("INTERSECT with All is not supported in Snowflake")
             return super().intersect_op(expression)
 
+        def min_sql(self, expression):
+            name = "LEAST" if expression.expressions else "MIN"
+            return rename_func(name)(self, expression)
```
GeorgeSittas, line 329: I think other dialects can leverage this too, e.g. postgres seems to use the same semantics as snowflake.

tobymao (approved):

tobymao:
minor version bump due to args change for least

sqlglot/expressions.py
```diff
@@ -3505,11 +3505,13 @@ class Matches(Func):
...
 class Max(AggFunc):
-    arg_types = {"this": True, "expression": False}
+    arg_types = {"this": True, "expressions": False}
+    is_var_len_args = True
 
 
 class Min(AggFunc):
-    arg_types = {"this": True, "expression": False}
+    arg_types = {"this": True, "expressions": False}
+    is_var_len_args = True
```
GeorgeSittas, line 3514: This is probably a breaking change, let me know if I missed something by changing the `expression` arg. The `is_var_len_args` change was needed to support SQLite's variable length `MIN` and `MAX` variants.

PR #1265, opened 2023-03-08
sqlite datediff

tobymao:
can you check this out and take this to the finish line?


GeorgeSittas:
Fixed tests and did a quick search. Seems legit, but I'm not 100% if we preserve the semantics exactly.

tobymao:
it’s not, it’s an estimation

PR #1266, opened 2023-03-08
fix exp.func bug

barakalon:
[This PR](https://github.com/tobymao/sqlglot/pull/1261) introduced a little bug in `exp.func`:

```python
exp.func("MIN", 1).sql() == 'MIN(1, )'
```


tobymao:
there should never be tuples in args

tobymao:
closed in favor of #7469e92d

PR #1270, opened 2023-03-08
Improve support for DuckDB types

### LINKED ISSUE #1267: DuckDB unsigned integer types fail to parse
[Unsigned integer types](https://duckdb.org/docs/sql/data_types/overview.html#general-purpose-data-types) fail to parse:

```
In [13]: sg.parse_one("select cast(1 as UINTEGER)", read="duckdb")
...
ParseError: Expected TYPE after CAST. Line 1, Col: 18.
  select cast(1 as UINTEGER)
```

It would also be good to add support for all the type aliases if possible.

GeorgeSittas:
Fixes #1267

- Moved some DuckDB tests to `test_cast`.
- Added support for missing type aliases.
- Added support for `BIT` as a standalone type, because it's semantics are different from `BOOLEAN` for DuckDB.
- All `INT` aliases are mapped to `INT` instead of the "default" type name shown in the docs, which is `INTEGER`.
- All `BLOB` aliases are mapped to `BLOB`.
- Created new types representing the unsigned variants of integer types.

Reference: https://duckdb.org/docs/sql/data_types/overview.html#general-purpose-data-types


cpcloud:
Thanks for the quick turnaround!

PR #1271, opened 2023-03-08
Fix parsing of <@, @> and && for postgres

### LINKED ISSUE #1268: Postgres array comparison operators not supported
Following Postgres array operations [1] raise ParseError:
```
SELECT ARRAY[1, 2, 3] @> ARRAY[1, 2] AS left_contains;
SELECT ARRAY[1, 2, 3] && ARRAY[1, 2] AS has_overlap;
```

And this one is parsed but incorrectly (`<@` is parsed as `<`):
```
SELECT ARRAY[1, 2, 3] <@ ARRAY[1, 2] AS right_contains;
```

[1] https://www.postgresql.org/docs/current/functions-array.html

GeorgeSittas:
Fixes #1268

- https://www.postgresql.org/docs/current/functions-array.html


sqlglot/tokens.py
```diff
@@ -127,6 +128,7 @@ class TokenType(AutoName):
     ANY = auto()
     APPLY = auto()
     ARRAY = auto()
+    ARRAY_OVERLAPS = auto()
```
GeorgeSittas, line 131: Added this instead of using `OVERLAPS` because there's already an `OVERLAPS` binary operator in postgres.

sqlglot/tokens.py
```diff
@@ -50,6 +50,7 @@ class TokenType(AutoName):
     PARAMETER = auto()
     SESSION_PARAMETER = auto()
     NATIONAL = auto()
+    CONTAINS = auto()
```
tobymao, line 53: should this be array contains?

GeorgeSittas: Sure

sqlglot/tokens.py
tobymao: maybe it should be damp? double amp

sqlglot/tokens.py
tobymao: maybe atlt?

GeorgeSittas: What is atlt?

tobymao: less than at

so ltat, what do we do for the other json ops?

GeorgeSittas: > why is there only one token instead of two?

It made sense to me because it's roughly the same relation, just with the operands reversed.

tobymao: isn’t that the same as < and > ?

GeorgeSittas: Done (discussed in slack)

tobymao (approved):
you can make DAMP a global token

PR #1279, opened 2023-03-09
Map TIMESTAMPTZ to 'TIMESTAMP' for Spark

### LINKED ISSUE #1272: type `TIMESTAMPTZ` doesn't exist in sparksql
sqlglot transpiles a trino type like `timestamp(6) with time zone` as the hive type `TIMESTAMPTZ` for spark, which isn't a spark type.  I'd be happy to submit a PR if there's advice about how to address this.

```
>>> sqlglot.transpile('select cast(created as timestamp(6) with time zone) created from some_table', read="trino", write="spark")
['SELECT CAST(created AS TIMESTAMPTZ) AS created FROM some_table']
```

GeorgeSittas:
Fixes #1272

We can probably change hive too here, but I wasn't 100% sure about its semantics.

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -263,6 +263,7 @@ class Hive(Dialect):
             exp.DataType.Type.TEXT: "STRING",
             exp.DataType.Type.DATETIME: "TIMESTAMP",
             exp.DataType.Type.VARBINARY: "BINARY",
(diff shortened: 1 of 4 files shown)


tobymao (commented):
yea please change hive

GeorgeSittas:
Made it so that `TIMESTAMPTZ` is transpiled to `TIMESTAMP` for Hive.

PR #1280, opened 2023-03-09
Added support for expression building for DELETE

mpf82:
This PR aims to add support for the following methods:

Delete.from_() and Delete.where()

Examples:

``Delete(this="tbl").where("x = 1")``

> "DELETE FROM tbl WHERE x = 1"

``Delete().from_("tbl").where("x = 1")``

> "DELETE FROM tbl WHERE x = 1"

---

If this PR is fine, I'll look into doing the same for INSERT and UPDATE

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1035,7 +1035,7 @@ class Delete(Expression):
 
     def delete(
         self,
-        table: str | Table,
+        table: str | Expression,
         dialect: DialectType = None,
         copy: bool = True,
         **opts,
@@ -1048,7 +1048,7 @@ class Delete(Expression):
             'DELETE FROM tbl'
 
         Args:
-            table (str | Table): the table from which to delete.
+            table (str | Expression): the table from which to delete.
             dialect (str): the dialect used to parse the input expression.
             copy (bool): if `False`, modify this expression instance in-place.
             opts (kwargs): other options to use to parse the input expressions.
@@ -1107,7 +1107,7 @@ class Delete(Expression):
 
     def returning(
         self,
-        expression: str | Returning,
+        expression: str | Expression,
         dialect: DialectType = None,
         copy: bool = True,
         **opts,
@@ -4245,7 +4245,7 @@ def from_(*expressions, dialect=None, **opts) -> Select:
 def update(
     table: str | Table,
     properties: dict,
-    where: str | Condition = None,
+    where: str | Expression = None,
     from_: str | Expression = None,
     dialect: DialectType = None,
     **opts,
@@ -4260,7 +4260,7 @@ def update(
     Args:
         *properties (Dict[str, Any]): dictionary of properties to set which are
             auto converted to sql objects eg None -> NULL
-        where (str|Condition): sql conditional parsed into a WHERE statement
+        where (str|Expression): sql conditional parsed into a WHERE statement
         from_ (str|Table): sql statement parsed into a FROM statement
         dialect (str): the dialect used to parse the input expressions.
         **opts: other options to use to parse the input expressions.
@@ -4293,8 +4293,8 @@ def update(
 
 def delete(
     table: str | Expression,
-    where: str | Condition = None,
-    returning: str | Returning = None,
+    where: str | Expression = None,
+    returning: str | Expression = None,
     dialect: DialectType = None,
     **opts,
 ) -> Delete:
@@ -4306,7 +4306,8 @@ def delete(
         'DELETE FROM my_table WHERE id > 1'
 
     Args:
-        where (str|Condition): sql conditional parsed into a WHERE statement
+        where (str|Expression): sql conditional parsed into a WHERE statement
+        returning (str|Expression): sql conditional parsed into a RETURNING statement
         dialect (str): the dialect used to parse the input expressions.
         **opts: other options to use to parse the input expressions.
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1093,6 +1093,7 @@ class Delete(Expression):
             **opts,
         )
 
+
 class Drop(Expression):
     arg_types = {
         "this": False,
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1050,10 +1050,10 @@ class Delete(Expression):
             'DELETE FROM tbl'
 
         Args:
-            table (str | Expression): the table from which to delete.
-            dialect (str): the dialect used to parse the input expression.
-            copy (bool): if `False`, modify this expression instance in-place.
-            opts (kwargs): other options to use to parse the input expressions.
+            table: the table from which to delete.
+            dialect: the dialect used to parse the input expression.
+            copy: if `False`, modify this expression instance in-place.
+            opts: other options to use to parse the input expressions.
 
         Returns:
             Delete: the modified expression.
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -1033,6 +1033,66 @@ class Constraint(Expression):
 class Delete(Expression):
     arg_types = {"with": False, "this": False, "using": False, "where": False, "returning": False}
 
+    def from_(self, expression, dialect=None, copy=True, **opts) -> Delete:
```
tobymao, line 1036: add types. also need to add a helper method for delete like there is for select

tobymao: oh there already is one?

sqlglot/expressions.py
```diff
@@ -1033,6 +1033,66 @@ class Constraint(Expression):
...
+        """
+        Set the FROM part of the DELETE expression.
+
+        Example:
+            >>> Delete().from_("tbl").sql()
```
tobymao, line 1041: examples should use the helper delete, not the class

tests/test_build.py
```diff
@@ -510,6 +510,20 @@ def test_build(self):
                 .qualify("row_number() OVER (PARTITION BY a ORDER BY b) = 1"),
                 "SELECT * FROM table QUALIFY ROW_NUMBER() OVER (PARTITION BY a ORDER BY b) = 1",
             ),
+            (lambda: exp.delete("tbl", "x = 1").from_("table"), "DELETE FROM table WHERE x = 1"),
+            (
+                lambda: exp.delete("tbl", "x = 1").from_("table").where("y = 2"),
+                "DELETE FROM table WHERE x = 1 AND y = 2",
+            ),
+            (lambda: exp.Delete(this="tbl"), "DELETE FROM tbl"),
```
tobymao, line 518: this is not correct usage, this is a string and needs to be a table. this is why you should only use the helper and not the class

tests/test_build.py
```diff
@@ -510,6 +510,20 @@ def test_build(self):
...
+            (lambda: exp.Delete(this="tbl").where("x = 1"), "DELETE FROM tbl WHERE x = 1"),
+            (lambda: exp.Delete().from_("tbl").where("x = 1"), "DELETE FROM tbl WHERE x = 1"),
+            (
+                lambda: exp.Delete(this="tbl").from_("tbl2").where("x = 1"),
+                "DELETE FROM tbl2 WHERE x = 1",
+            ),
+            (lambda: exp.Delete().from_(exp.From(this="tbl")), "DELETE FROM tbl"),
+            (lambda: exp.Delete().from_(exp.Table(this="tbl")), "DELETE FROM tbl"),
```
tobymao, line 526: same with these tests, they don’t keep the correctness of the ast

sqlglot/expressions.py
mpf82: I've copied those method signatures from Select, but I can add the types there, too.

Select helper method is Select.select, while the delete helper method is on the module level.

Should the delete helpher method be move from the module to Delete.delete as well? 

Currently, the delete helpher methods forces you to add a where part, I'll probably make that optional.

GeorgeSittas: If we can easily add types for these methods, that'd be ideal. Otherwise, we can defer for later, given that our methods were already untyped to begin with here.

> Select helper method is Select.select, while the delete helper method is on the module level.

We have both `exp.Select.select` and `exp.select`, which are two different callables. I think the original suggestion was to replicate this API for the `delete` functionality, i.e. have both `exp.Delete.delete` and `exp.delete`.

sqlglot/expressions.py
```diff
@@ -1033,6 +1035,113 @@ class Constraint(Expression):
...
+
+        Example:
+            >>> delete("tbl").sql()
+            'DELETE FROM tbl'
+
+        Args:
+            table (str | Expression): the table from which to delete.
+            dialect (str): the dialect used to parse the input expression.
+            copy (bool): if `False`, modify this expression instance in-place.
+            opts (kwargs): other options to use to parse the input expressions.
```
GeorgeSittas, line 1056: The parentheses can be omitted now that we have type hints.

sqlglot/expressions.py
```diff
@@ -1033,6 +1035,113 @@ class Constraint(Expression):
...
+
+        Args:
+            *expressions (str | Expression): the SQL code strings to parse.
+                If an `Expression` instance is passed, it will be used as-is.
+                Multiple expressions are combined with an AND operator.
+            append (bool): if `True`, AND the new expressions to any existing expression.
+                Otherwise, this resets the expression.
+            dialect (str): the dialect used to parse the input expressions.
+            copy (bool): if `False`, modify this expression instance in-place.
+            opts (kwargs): other options to use to parse the input expressions.
```
GeorgeSittas, line 1094: Same comment for parentheses.

sqlglot/expressions.py
```diff
@@ -4146,38 +4262,44 @@ def update(table, properties, where=None, from_=None, dialect=None, **opts) -> U
     Args:
         *properties (Dict[str, Any]): dictionary of properties to set which are
             auto converted to sql objects eg None -> NULL
-        where (str): sql conditional parsed into a WHERE statement
-        from_ (str): sql statement parsed into a FROM statement
+        where (str|Expression): sql conditional parsed into a WHERE statement
+        from_ (str|Table): sql statement parsed into a FROM statement
```
GeorgeSittas, line 4266: Same

sqlglot/expressions.py
```diff
@@ -4186,19 +4308,20 @@ def delete(table, where=None, dialect=None, **opts) -> Delete:
         'DELETE FROM my_table WHERE id > 1'
 
     Args:
-        where (str|Condition): sql conditional parsed into a WHERE statement
+        where (str|Expression): sql conditional parsed into a WHERE statement
+        returning (str|Expression): sql conditional parsed into a RETURNING statement
```
GeorgeSittas, line 4312: Same

sqlglot/expressions.py
```diff
@@ -4186,19 +4308,20 @@ def delete(table, where=None, dialect=None, **opts) -> Delete:
...
     Returns:
         Delete: the syntax tree for the DELETE statement.
     """
-    return Delete(
-        this=maybe_parse(table, into=Table, dialect=dialect, **opts),
-        where=Where(this=where)
-        if isinstance(where, Condition)
-        else maybe_parse(where, into=Where, dialect=dialect, prefix="WHERE", **opts),
-    )
+    delete_expr = Delete().delete(table, dialect=dialect, **opts)
```
GeorgeSittas, line 4319: I think we can do `copy=False` here because the instance created by `Delete()` becomes garbage.

sqlglot/expressions.py
```diff
@@ -4186,19 +4308,20 @@ def delete(table, where=None, dialect=None, **opts) -> Delete:
...
+    if where:
+        delete_expr = delete_expr.where(where, dialect=dialect, **opts)
```
GeorgeSittas, line 4321: Same

sqlglot/expressions.py
```diff
@@ -4186,19 +4308,20 @@ def delete(table, where=None, dialect=None, **opts) -> Delete:
...
+    if returning:
+        delete_expr = delete_expr.returning(returning, dialect=dialect, **opts)
```
GeorgeSittas, line 4323: Same

sqlglot/expressions.py
```diff
@@ -626,6 +626,8 @@ def load(cls, obj):
     t.Type[Expression],
     t.Collection[t.Union[str, t.Type[Expression]]],
 ]
+ExpressionType = t.Union[str, Expression]
+ExpressionTypeNone = t.Union[str, Expression, None]
```
GeorgeSittas, line 630: I'd personally prefer to use `t.Optional` instead of creating another type just for `None`, but let's see what others think about this too.

GeorgeSittas (commented):
Left some comments but I think the code looks good.

sqlglot/expressions.py
mpf82: Agreed. I'll leave it as-is for now.

tobymao: the name is strange, maybe it should be StrOrExp

tobymao: i don't think it's necessary to make an alias for the optional one, just do t.Optional[StrOrExp]

tobymao:
#1296

PR #1284, opened 2023-03-09
Add support for parsing COUNT_IF and transpiling it to SUM(CASE ..)

GeorgeSittas:


sqlglot/dialects/dialect.py
```diff
@@ -412,6 +412,16 @@ def min_or_least(self: Generator, expression: exp.Min) -> str:
...
 
 
+def count_if_to_sum(self: Generator, expression: exp.CountIf) -> str:
+    cond = expression.this
+
+    if isinstance(expression.this, exp.Distinct):
+        cond = expression.this.expressions[0]
+        self.unsupported("DISTINCT is not supported when converting COUNT_IF to SUM")
+
+    return self.sql(exp.func("sum", exp.func("if", cond, 1, 0)))
```
tobymao, line 422: return self.func("sum", self.func("if", cond, 1, 0))

GeorgeSittas: I don't think we can do this here because we need a string, right?

sqlglot/parser.py
```diff
@@ -3219,6 +3220,30 @@ def _parse_if(self) -> t.Optional[exp.Expression]:
...
+        if all_ and distinct:
+            self.raise_error("Cannot specify both ALL and DISTINCT in COUNT_IF")
+
+        cond = self._parse_conjunction()
+
+        index = self._index
+        if self._match_pair(TokenType.R_PAREN, TokenType.FILTER):
+            if self._match(TokenType.L_PAREN):
+                self._match(TokenType.WHERE)
+                cond = exp.and_(cond, self._parse_conjunction())
```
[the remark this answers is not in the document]
tobymao: why is this necessary? don't we already support filter and distinct?

sqlglot/dialects/dialect.py
tobymao: doesn’t self.func return a string?

GeorgeSittas: Nope, returns an `exp.Func`

tobymao: https://github.com/tobymao/sqlglot/blob/main/sqlglot/generator.py#L1859

are you sure?

GeorgeSittas: ignore me, making the fix

sqlglot/parser.py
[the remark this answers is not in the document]
GeorgeSittas: Ok I completely removed this function. Nice!

sqlglot/dialects/dialect.py
GeorgeSittas: I didn't change the internal `exp.func` because `if` needs to be transpiled correctly across dialects (e.g. sqlite supports `IIF` instead of `IF`).

tobymao (approved):

PR #1288, opened 2023-03-10
Implement eliminate_qualify transformation

GeorgeSittas:
Reference: https://docs.snowflake.com/en/sql-reference/constructs/qualify

Tested the Snowflake example:

```sql
SELECT i, p, o FROM qt QUALIFY ROW_NUMBER() OVER (PARTITION BY p ORDER BY o) = 1;
```

Gives same results as what we get by transpiling it in SQLite:

```sql
SELECT i, p, o FROM (SELECT i, p, o, ROW_NUMBER() OVER (PARTITION BY p ORDER BY o NULLS LAST) AS _w_0 FROM qt) WHERE _w_0 = 1;
```

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -60,8 +60,7 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
         and expression.args["distinct"].args.get("on")
         and isinstance(expression.args["distinct"].args["on"], exp.Tuple)
     ):
-        distinct_cols = expression.args["distinct"].args["on"].expressions
-        expression.args["distinct"].pop()
+        distinct_cols = expression.args["distinct"].pop().args["on"].expressions
         outer_selects = expression.selects
         row_number = find_new_name(expression.named_selects, "_row_number")
         window = exp.Window(
@@ -70,8 +69,7 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
         )
         order = expression.args.get("order")
         if order:
-            window.set("order", order.copy())
-            order.pop()
+            window.set("order", order.pop().copy())
         window = exp.alias_(window, row_number)
         expression.select(window, copy=False)
         return exp.select(*outer_selects).from_(expression.subquery()).where(f'"{row_number}" = 1')
@@ -91,9 +89,15 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
     otherwise we won't be able to refer to it in the outer query's WHERE clause.
     """
     if isinstance(expression, exp.Select) and expression.args.get("qualify"):
-        outer_selects = exp.select(*[s.alias_or_name or s for s in expression.selects])
-        qualify_filters = expression.args["qualify"].this
-        expression.args["qualify"].pop()
+        taken = set(expression.named_selects)
+        for select in expression.selects:
+            if not select.alias_or_name:
+                alias = find_new_name(taken, "_c")
+                select.replace(exp.alias_(select.copy(), alias))
+                taken.add(alias)
+
+        outer_selects = exp.select(*[select.alias_or_name for select in expression.selects])
+        qualify_filters = expression.args["qualify"].pop().this
 
         for expr in qualify_filters.find_all((exp.Window, exp.Column)):
             if isinstance(expr, exp.Window):
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
+import itertools
 import typing as t
 
+from sqlglot import expressions as exp
 from sqlglot.helper import find_new_name
 
 if t.TYPE_CHECKING:
     from sqlglot.generator import Generator
-
-from sqlglot import expressions as exp
 
 
 def unalias_group(expression: exp.Expression) -> exp.Expression:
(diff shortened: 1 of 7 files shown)


sqlglot/transforms.py
```diff
@@ -79,6 +79,35 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
...
+
+    if isinstance(expression, exp.Select) and expression.args.get("qualify"):
+        outer_selects = exp.select(*[s.alias_or_name or s for s in expression.selects])
+        qualify_filters = expression.args["qualify"].this
+        expression.args["qualify"].pop()
+        sequence = itertools.count()
+
+        for window in qualify_filters.find_all(exp.Window):
+            window_alias = f"_w_{next(sequence)}"
+            expression.select(exp.alias_(window.copy(), window_alias), copy=False)
```
tobymao, line 103: does transform auto copy expression? or are you mutating it here

sqlglot/transforms.py
```diff
@@ -79,6 +79,35 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
...
+            window.replace(exp.column(window_alias))
+
+        return outer_selects.from_(expression.subquery()).where(qualify_filters)
```
tobymao, line 106: you're missing an alias

tests/dialects/test_snowflake.py
```diff
@@ -17,6 +17,14 @@ def test_snowflake(self):
         )
         self.validate_identity("COMMENT IF EXISTS ON TABLE foo IS 'bar'")
 
+        self.validate_all(
+            "SELECT i, p, o FROM qt QUALIFY ROW_NUMBER() OVER (PARTITION BY p ORDER BY o) = 1",
+            write={
+                "": "SELECT i, p, o FROM qt QUALIFY ROW_NUMBER() OVER (PARTITION BY p ORDER BY o NULLS LAST) = 1",
+                "snowflake": "SELECT i, p, o FROM qt QUALIFY ROW_NUMBER() OVER (PARTITION BY p ORDER BY o) = 1",
+                "sqlite": "SELECT i, p, o FROM (SELECT i, p, o, ROW_NUMBER() OVER (PARTITION BY p ORDER BY o NULLS LAST) AS _w_0 FROM qt) WHERE _w_0 = 1",
```
tobymao, line 25: you should have a subquery alias and then qualify everything

sqlglot/transforms.py
GeorgeSittas: `preprocess` copies the expression before applying any transformations.

tests/dialects/test_snowflake.py
GeorgeSittas: Is the current approach problematic, or is it just bad practice to leave it unaliased? I didn't qualify on purpose because I was thinking the correspondence between subquery projections and top-level query projections is 1-1, so there shouldn't be any ambiguities, right?

tobymao: bad practice and postgres doesn’t allow it

GeorgeSittas: Got it, I'll make the change then, thanks for clarifying.

GeorgeSittas: I found a bug where unselected columns participating in the QUALIFY clause would not be pushed down, so external references were invalid. I tried to fix it in my latest commit, but I haven't yet qualified every column reference. If we need to have this, should I just transform every column in the outer selects & outer where clause?

sqlglot/transforms.py
```diff
@@ -79,6 +78,36 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
...
+    The idea behind this transformation can be seen in Snowflake's documentation for QUALIFY:
+    https://docs.snowflake.com/en/sql-reference/constructs/qualify
+
+    Some dialects don't support window functions in the WHERE clause, so we need to include them as
+    projections in the subquery, in order to refer to them in the outer filter using aliases. Also,
+    if a column is referenced in the QUALIFY clause but is not selected, we need to include it too,
+    otherwise we won't be able to refer to it in the outer query's WHERE clause.
+    """
+    if isinstance(expression, exp.Select) and expression.args.get("qualify"):
+        outer_selects = exp.select(*[s.alias_or_name or s for s in expression.selects])
```
tobymao, line 94: this is not right, you should put a test case for this but you need to create an alias if it doesn't have one

imagine you have something like select a + 1, you're going to then re-select a + 1 in the outer select but a won't exist.

sqlglot/transforms.py
```diff
@@ -79,6 +78,36 @@ def eliminate_distinct_on(expression: exp.Expression) -> exp.Expression:
...
+        qualify_filters = expression.args["qualify"].this
```
tobymao, line 95: qualify = expression.args["qualify"].pop()
qualify_filters = qualify.this

sqlglot/transforms.py
GeorgeSittas: Good catch, thanks. Let me know if I missed something in my implementation.

tests/dialects/test_snowflake.py
tobymao: not sure what you mean

tobymao (approved):

tests/dialects/test_snowflake.py
GeorgeSittas: The bug was that before `p` and `o` wouldn't be selected in the inner query [here](https://github.com/tobymao/sqlglot/pull/1288/files#diff-0b04b14c1b4c9192c8ba11f96a80aa0b79b1a2979835a01767f7c778fdf4eaecR86-R87).

Do we need to qualify all columns with subquery alias in the outer select expression?

tobymao: no, because you can always just use the optimizer

PR #1291, opened 2023-03-13
Rename day of week function for duckdb

cillianod-wf:
Currently `dayofweek()` transpiles to `day_of_week()` in duckdb but this is incorrect, it should be `dayofweek()` as stated in the [docs](https://duckdb.org/docs/sql/functions/datepart.html#part-functions).

### DIFF
--- a/tests/dialects/test_presto.py
+++ b/tests/dialects/test_presto.py
@@ -189,34 +189,38 @@ class TestPresto(Validator):
         )
 
         self.validate_all(
-            "DAY_OF_WEEK(timestamp '2012-08-08 01:00')",
+            "DAY_OF_WEEK(timestamp '2012-08-08 01:00:00')",
             write={
-                "spark": "DAYOFWEEK(CAST('2012-08-08 01:00' AS TIMESTAMP))",
-                "presto": "DAY_OF_WEEK(CAST('2012-08-08 01:00' AS TIMESTAMP))",
+                "spark": "DAYOFWEEK(CAST('2012-08-08 01:00:00' AS TIMESTAMP))",
+                "presto": "DAY_OF_WEEK(CAST('2012-08-08 01:00:00' AS TIMESTAMP))",
+                "duckdb": "DAYOFWEEK(CAST('2012-08-08 01:00:00' AS TIMESTAMP))",
             },
         )
 
         self.validate_all(
-            "DAY_OF_MONTH(timestamp '2012-08-08 01:00')",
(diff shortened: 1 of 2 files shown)


cillianod-wf:
Will also add changes to `day_of_month` etc.

tests/dialects/test_presto.py
```diff
@@ -193,6 +193,7 @@ def test_time(self):
             write={
                 "spark": "DAYOFWEEK(CAST('2012-08-08 01:00' AS TIMESTAMP))",
                 "presto": "DAY_OF_WEEK(CAST('2012-08-08 01:00' AS TIMESTAMP))",
+                "duckdb": "DAYOFWEEK(CAST('2012-08-08 01:00' AS TIMESTAMP))"
```
GeorgeSittas, line 196: Looks like this cast is invalid in duckdb:

```python
>>> duckdb.connect().execute("SELECT DAYOFWEEK(CAST('2012-08-08 01:00' AS TIMESTAMP))").fetchdf()
duckdb.ConversionException: Conversion Error: timestamp field value out of range: "2012-08-08 01:00", expected format is (YYYY-MM-DD HH:MM:SS[.US][±HH:MM| ZONE])
```

Let's change it to a valid one: `DAYOFWEEK(CAST('2012-08-08 01:00:00' AS TIMESTAMP))`.

GeorgeSittas (approved):

tests/dialects/test_presto.py
cillianod-wf: Nice one, thanks for the heads up! Will make that change too

cillianod-wf: Is this an entirely separate issue in that `timestamp '2012-08-08 01:00'` transpiles to `CAST('2012-08-08 01:00' AS TIMESTAMP)` in `duckdb` when it should be `CAST('2012-08-08 01:00:00' AS TIMESTAMP)`? Maybe I can add the test elsewhere to give something like `DAY_OF_WEEK('2012-08-08')` -> `DAYOFWEEK('2012-08-08')`?

GeorgeSittas: I meant that we can just add `:00` for this test. We don't need to transpile this case correctly for DuckDB.

cillianod-wf: Cool, I've made changes now and I think it aligns with what you've said above but let me know if not!

GeorgeSittas: LGTM, thanks for the PR!

PR #1292, opened 2023-03-13
Move SET parsing/generating logic from MySQL to base

### LINKED ISSUE #1225: SET commands not decomposed

GeorgeSittas:
- Moved a subset of the logic we use for parsing / generating SET statements to the base classes, so that we can parse statements like `SET variable = value` with finer granularity.
- When the base `_parse_set` method fails, the parser will fallback to parsing the input as a command.

cc: @joocer @tobymao @barakalon 

Fixes #1225

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3879,7 +3879,7 @@ class Parser(metaclass=_Parser):
 
         left = self._parse_primary() or self._parse_id_var()
 
-        if not self._match(TokenType.EQ):
+        if not self._match_texts(("=", "TO")):
             self._retreat(index)
             return None
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -460,6 +460,7 @@ class Parser(metaclass=_Parser):
         TokenType.LOAD_DATA: lambda self: self._parse_load_data(),
         TokenType.MERGE: lambda self: self._parse_merge(),
         TokenType.ROLLBACK: lambda self: self._parse_commit_or_rollback(),
+        TokenType.SET: lambda self: self._parse_set(),
         TokenType.UNCACHE: lambda self: self._parse_uncache(),
         TokenType.UPDATE: lambda self: self._parse_update(),
         TokenType.USE: lambda self: self.expression(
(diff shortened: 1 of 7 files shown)


sqlglot/parser.py
```diff
@@ -3864,7 +3911,18 @@ def _parse_merge(self) -> exp.Expression:
         )
 
     def _parse_set(self) -> exp.Expression:
-        return self.expression(exp.Set, expressions=self._parse_csv(self._parse_set_item))
+        index = self._index
+        try:
+            return self.expression(exp.Set, expressions=self._parse_csv(self._parse_set_item))
+        except ParseError:
```
barakalon, line 3917: This could lead to bugs that are hard to debug, as we silence the error.

GeorgeSittas: Agreed, I don't feel very good about this too. Maybe we should get rid of `_parse_set_item` and `_default_parse_set_Item` and have `_parse_set_item_assignment` return `None` on failure, so we can gracefully handle it. How does that sound?

tobymao: you should definitely not use try catch . if there are any tokens left after a set, it’s failed

GeorgeSittas: > if there are any tokens left after a set, it’s failed

Not necessarily, since spark [supports](https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-aux-conf-mgmt-set.html#examples) `SET;`. Agreed with the comments regarding try-catch though, I will try to remove that.

tobymao: if you parse set; there are no tokens left, just check for self._curr and there are none so you don’t raise

GeorgeSittas: Ignore me, I misread your comment and got confused, you're right.

GeorgeSittas: Yep, I realized what you mean now -- thanks for the suggestion

GeorgeSittas: Got rid of try / catch.

sqlglot/parser.py
```diff
@@ -3863,8 +3862,71 @@ def _parse_merge(self) -> exp.Expression:
...
+        self, kind: t.Optional[str] = None
+    ) -> t.Optional[exp.Expression]:
+        index = self._index
+
+        if kind in {"GLOBAL", "SESSION"} and self._match_text_seq("TRANSACTION"):
+            return self._parse_set_transaction(global_=kind == "GLOBAL")
+
+        left = self._parse_primary() or self._parse_id_var()
+
+        if not self._match(TokenType.EQ):
```
tobymao, line 3882: why don't you parse TO?

PR #1293, opened 2023-03-13
Allow parsing strings as table references

GeorgeSittas:
Apparently, DuckDB allows this:

```
>>> import duckdb
>>> duckdb.connect().execute("WITH 'x.y' AS (SELECT 1) SELECT * FROM 'x.y'").fetchdf()
   1
0  1
>>> duckdb.connect().execute("WITH 'x' AS (SELECT 1) SELECT * FROM 'x'").fetchdf()
   1
0  1
>>> duckdb.connect().execute("WITH 'x' AS (SELECT 1) SELECT * FROM x").fetchdf()
   1
0  1
```

This seems non-standard, but it felt innocuous enough to support. The places where I added `self._parse_string()` would lead to an exception without it anyway, in case we weren't able to parse a table.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1759,7 +1759,7 @@ class Parser(metaclass=_Parser):
         return self.expression(exp.With, expressions=expressions, recursive=recursive)
 
     def _parse_cte(self) -> exp.Expression:
-        alias = self._parse_table_alias()
+        alias = self._parse_table_alias() or self._parse_string()
         if not alias or not alias.this:
             self.raise_error("Expected CTE to have alias")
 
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -1759,7 +1759,7 @@ def _parse_with(self, skip_with_token: bool = False) -> t.Optional[exp.Expressio
         return self.expression(exp.With, expressions=expressions, recursive=recursive)
 
     def _parse_cte(self) -> exp.Expression:
-        alias = self._parse_table_alias()
+        alias = self._parse_table_alias() or self._parse_string()
```
tobymao, line 1762: why not add it to parse table alias?

sqlglot/parser.py
```diff
@@ -2046,7 +2046,12 @@ def _parse_create_table_index(self) -> t.Optional[exp.Expression]:
     def _parse_table_parts(self, schema: bool = False) -> exp.Expression:
         catalog = None
         db = None
-        table = (not schema and self._parse_function()) or self._parse_id_var(any_token=False)
+
+        table = (
+            (not schema and self._parse_function())
+            or self._parse_id_var(any_token=False)
+            or self._parse_string()
```
tobymao, line 2053: don’t return a string literal, convert the string into an identifier

tobymao: quoted

tobymao (approved):

PR #1299, opened 2023-03-15
Handle structs gracefully in qualify columns

GeorgeSittas:
This is a WIP, I'm trying to fix the optimizer so that it doesn't break when we have structs.

EDIT: I'm going to try and move this change to `qualify_columns` instead of `pushdown_projections`.

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -37,6 +37,7 @@ def qualify_columns(expression, schema):
             _qualify_outputs(scope)
         _expand_group_by(scope, resolver)
         _expand_order_by(scope)
+
     return expression
 
 
@@ -213,6 +214,23 @@ def _qualify_columns(scope, resolver):
             # column_table can be a '' because bigquery unnest has no table alias
             if column_table:
                 column.set("table", column_table)
+        elif column_table not in scope.sources:
+            # structs are used as tables (e.g. "struct"."field"), so we need to qualify them separately
+            struct_fields = list(val for val in reversed(column.args.values()) if val is not None)
+
+            root_struct = struct_fields[0]
+            struct_table = resolver.get_table(root_struct.name)
+
+            if struct_table:
+                while column.parent and isinstance(column.parent, exp.Dot):
+                    column = column.parent
+                    struct_fields.append(column.expression)
+
+                new_column = exp.column(root_struct, table=struct_table)
+                for field in struct_fields[1:]:
+                    new_column = exp.Dot(this=new_column, expression=exp.to_identifier(field))
+
+                column.replace(new_column)
 
     columns_missing_from_scope = []
     # Determine whether each reference in the order by clause is to a column or an alias.
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -216,7 +216,7 @@ def _qualify_columns(scope, resolver):
                 column.set("table", column_table)
         elif column_table not in scope.sources:
             # structs are used like tables (e.g. "struct"."field"), so they need to be qualified
-            # separately by parsing them as <table>.<column>.dot.dot.dot ...
+            # separately and represented as <table>.<column>.dot.dot.dot ...
 
             struct_fields = [val for val in reversed(list(column.args.values())) if val is not None]
             struct_root = struct_fields[0]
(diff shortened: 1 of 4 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -213,6 +214,31 @@ def _qualify_columns(scope, resolver):
             # column_table can be a '' because bigquery unnest has no table alias
             if column_table:
                 column.set("table", column_table)
+        elif column_table not in scope.sources:
+            # structs are used like tables (e.g. "struct"."field"), so they need to be qualified
+            # separately by parsing them as <table>.<column>.dot.dot.dot ...
+
+            struct_fields = [val for val in reversed(list(column.args.values())) if val is not None]
```
tobymao, line 221: do you need to call list() ?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -213,6 +214,31 @@ def _qualify_columns(scope, resolver):
...
+                struct_root = struct_fields[0]
+            else:
+                struct_table = resolver.get_table(struct_root.name)
+
+            if struct_table:
+                while column.parent and isinstance(column.parent, exp.Dot):
+                    column = column.parent
+                    struct_fields.append(column.expression)
+
+                new_column = exp.column(struct_root, table=struct_table)
```
tobymao, line 237: what if there's a db or catalog?

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: Python 3.7 doesn't support `reversed` for `dict_values`, so I opted for this workaround for the time being.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -213,6 +214,31 @@ def _qualify_columns(scope, resolver):
...
+                for field in struct_fields[1:]:
+                    new_column = exp.Dot(this=new_column, expression=exp.to_identifier(field))
```
tobymao, line 239: why do you need to call to identifier on field, isn't it already an identifier?

GeorgeSittas: Yea I'll change this, it was left over from a previous iteration

sqlglot/optimizer/qualify_columns.py
```diff
@@ -213,6 +214,31 @@ def _qualify_columns(scope, resolver):
...
+            struct_fields = [val for val in reversed(list(column.args.values())) if val is not None]
+            struct_root = struct_fields[0]
+
+            if struct_root.name in scope.sources:
+                # struct is already qualified, but we still need to change the AST representation
+                struct_table = struct_root
+                struct_fields = struct_fields[1:]
+                struct_root = struct_fields[0]
+            else:
+                struct_table = resolver.get_table(struct_root.name)
```
tobymao, line 230: why would this case happen?

GeorgeSittas: This is executed if the selection isn't already qualified, e.g. `"struct"."f1"` instead of `"cte"."struct"."f1"`.

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: I think the table here refers to the alias of the source `[catalog.][db.]table`. Can you provide an example & expected behavior so that I can check locally?

GeorgeSittas: Ok maybe there is a problem here:

```python
>>>
>>> optimize("SELECT c.d.t.struct.f1.f2 AS f2 FROM c.d.t").sql()
'SELECT "c"."d"."t"."struct"."f1"."f2" AS "f2" FROM "c"."d"."t" AS "t"'
```

I guess we only want `"t"` here before `"struct"`, because of the alias?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -373,10 +399,14 @@ def get_table(self, column_name: str) -> t.Optional[exp.Identifier]:
...
             while node and node.alias != table_name:
                 node = node.parent
+
         node_alias = node.args.get("alias")
         if node_alias:
             return node_alias.this
-        return exp.to_identifier(table_name)
+
+        return exp.to_identifier(
+            table_name, quoted=node.this.quoted if isinstance(node, exp.Table) else False
```
tobymao, line 408: else None, we don't want it to be always false that changes behavior right

GeorgeSittas: Thanks for catching this

PR #1304, opened 2023-03-16
Drop ORDER BY clause when transpiling ARRAY_AGG to hive

### LINKED ISSUE #1303: ARRAY_AGG() conversion between Trino and Spark
ARRAY_AGG() in Trino supports  `array_agg(x ORDER BY y DESC)` which currently is translated to Spark as `COLLECT_LIST(x ORDER BY y DESC)` but Spark doesn't support that syntax at all -- just `COLLECT_LIST(x)`.  Could we fix this mapping to syntactically parse (and drop the ordering) or is there a way to support the ordering?

GeorgeSittas:
Fixes #1303

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -339,8 +339,10 @@ class Hive(Dialect):
         }
 
         def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
-            arg = expression.this if isinstance(expression.this, exp.Order) else expression
-            return f"COLLECT_LIST({self.sql(arg, 'this')})"
+            return self.func(
+                "COLLECT_LIST",
+                expression.this.this if isinstance(expression.this, exp.Order) else expression.this,
+            )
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/hive.py
```diff
@@ -335,13 +338,17 @@ class Generator(generator.Generator):
             exp.TableFormatProperty: exp.Properties.Location.POST_SCHEMA,
         }
 
-        def with_properties(self, properties):
+        def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
+            arg = expression.this if isinstance(expression.this, exp.Order) else expression
+            return f"COLLECT_LIST({self.sql(arg, 'this')})"
```
tobymao, line 343: self.func("COLLECT_LIST", expression.this.this if isinstance(expression.this, exp.Order) else expression.this)

tobymao (approved):

PR #1306, opened 2023-03-17
canonicalize.ensure_bool_predicates

barakalon:
query:
```sql
SELECT
  a
FROM (
  SELECT 1 AS a
) AS x
WHERE a
```

Postgres:
```
error: argument of WHERE must be type boolean, not type integer
```

MySQL:
```
a
1
```

### DIFF
--- a/sqlglot/optimizer/canonicalize.py
+++ b/sqlglot/optimizer/canonicalize.py
@@ -84,4 +84,4 @@ def _replace_cast(node: exp.Expression, to: str) -> None:
 
 def _replace_int_predicate(expression: exp.Expression) -> None:
     if expression.type and expression.type.this in exp.DataType.INTEGER_TYPES:
-        expression.replace(exp.GT(this=expression.copy(), expression=exp.Literal.number(0)))
+        expression.replace(exp.NEQ(this=expression.copy(), expression=exp.Literal.number(0)))
--- a/sqlglot/optimizer/canonicalize.py
+++ b/sqlglot/optimizer/canonicalize.py
@@ -17,6 +17,7 @@ def canonicalize(expression: exp.Expression) -> exp.Expression:
     expression = add_text_to_concat(expression)
     expression = coerce_type(expression)
     expression = remove_redundant_casts(expression)
+    expression = ensure_bool_predicates(expression)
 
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/canonicalize.py
```diff
@@ -52,6 +53,17 @@ def remove_redundant_casts(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def ensure_bool_predicates(expression: exp.Expression) -> exp.Expression:
+    if isinstance(expression, exp.Connector):
+        _replace_int_predicate(expression.left)
+        _replace_int_predicate(expression.right)
+
+    elif isinstance(expression, (exp.Where, exp.Having)):
```
GeorgeSittas, line 61: Should we consider `exp.Qualify` as well?

GeorgeSittas (approved):

sqlglot/optimizer/canonicalize.py
```diff
@@ -68,3 +80,8 @@ def _replace_cast(node: exp.Expression, to: str) -> None:
     cast = exp.Cast(this=node.copy(), to=data_type)
     cast.type = data_type
     node.replace(cast)
+
+
+def _replace_int_predicate(expression: exp.Expression) -> None:
+    if expression.type and expression.type.this in exp.DataType.INTEGER_TYPES:
+        expression.replace(exp.GT(this=expression.copy(), expression=exp.Literal.number(0)))
```
tobymao, line 87: are negative types falsey?

sqlglot/optimizer/canonicalize.py
barakalon: I don't know much about QUALIFY. Do we annotate all the window functions properly?

sqlglot/optimizer/canonicalize.py
barakalon: Switched to NEQ

sqlglot/optimizer/canonicalize.py
GeorgeSittas: Good point, I'm not sure yet. We can tackle this in a future PR then.

Issue #1310, opened 2023-03-17
Qualified column names within parametric functions (ClickHouse)


wlhjason:
(Sorry, another ClickHouse one!)

The following ClickHouse SQL expressions lead to a parse error of the form `Expecting )` at `.`:
* `SELECT histogram(5)(t.x)`
* `SELECT groupUniqArray(2)(t.y)`
* `SELECT exponentialTimeDecayedAvg(60)(t."nested.value_col", t.time_col)`

(There is an incomplete list of parametric functions in ClickHouse here: https://clickhouse.com/docs/en/sql-reference/aggregate-functions/parametric-functions)

tobymao:
@wlhjason what's the function signature of exponentialTimeDecayedAvg i can't find documentation on it

wlhjason:
@tobymao It's kind of buried in the [implementation](https://clickhouse.com/codebrowser/ClickHouse/src/Processors/Transforms/WindowTransform.cpp.html)...

```
exponentialTimeDecayedAvg(decay_length: Float64)(value: Number, time: DateTime | Number)
```

There's also `exponentialTimeDecayedSum` and `exponentialTimeDecayedMax` with the same signature, plus `exponentialTimeDecayedCount` which omits the `value` input.

tobymao:
i'll add what you have here, and then you can follow my PR and add the rest

wlhjason:
Sure thing - thanks!

tobymao:
check out this pr, it will show you how to add and test  the rest. ping me on slack if you have any questions

wlhjason:
Amazing, will do - thanks @tobymao!

PR #1313, opened 2023-03-19
Support parametric anonymous functions

wlhjason:
Add general support for parsing parametric function calls of the form `foo(param1, param2)(arg1, arg2)`.

Addresses #1310 without the need to implement one expression type per parametric function.


wlhjason:
@tobymao After taking a look at 1e0c17f20aecae9959dd3cacc879363a225e73a5 and realising that ClickHouse has quite a few parametric functions ([`quantiles`](https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/quantiles) alone has many variants like `quantilesExact` and `quantilesExactIf`), I thought I'd try a different approach. See what you think?

sqlglot/parser.py
```diff
@@ -2849,21 +2843,23 @@ def _parse_function(
...
-                # Clickhouse supports function calls like foo(x, y)(z), so for these we need to also parse the
-                # second parameter list (i.e. "(z)") and the corresponding function will receive both arg lists.
-                if count_params(function) == 2:
-                    params = None
-                    if self._match_pair(TokenType.R_PAREN, TokenType.L_PAREN):
-                        params = self._parse_csv(self._parse_lambda)
-
-                    this = function(args, params)
+            # ClickHouse supports parametric function calls like `foo(param1, param2)(arg1, arg2)`
+            if self._match_pair(TokenType.R_PAREN, TokenType.L_PAREN):
```
GeorgeSittas, line 2847: I'm not sure if this is correct, because e.g. in Spark you can have

```sql
select posexplode(array(5,6,7)) (p,c)
```

You don't want to parse this as a parametric function, `(p, c)` is an alias expression without the `AS` keyword.

GeorgeSittas (commented):
It'd be nice if we could limit this only for Clickhouse, provided that the `func(...)(...)` syntax doesn't clash with any other "alias" syntax, like the one I mentioned below.

tobymao:
i don’t know about this. if you do this then clickhouse won’t be able to transpile to anything. additionally only clickhouse has parameterized functions. i think we should probably just add them as the previous pr

sqlglot/parser.py
wlhjason: Of course...

wlhjason:
Thanks for taking a look @tobymao @GeorgeSittas - I see now that the current implementation is no good, so I'll go ahead and close this.

To expand on the original problem: ClickHouse has >10 parametric functions (including ~10 variants of `quantile`), each of which can have one _or more_ ~10 [combinators](https://clickhouse.com/blog/aggregate-functions-combinators-in-clickhouse-for-arrays-maps-and-states) applied (as in `SELECT quantileExactIfOrNull(0.5)(number, number > 1) FROM numbers(5)`), so the number of valid parametric function names is huge.

Do you still think I should just go ahead and add some of these by hand? I'd also be happy to try to look for a more general solution.

tobymao:
ah interesting. so there are combinators. no i wouldn’t do this by hand. i didn’t know clickhouse has variants of these things. id make one class that takes in the parameters like sum, avg, etc so that it’s not so tedious. let me know if you need more details

wlhjason:
@tobymao Cool, I'll take a stab later - thanks!

PR #1319, opened 2023-03-20
fix missing concat type annotating

hollandjake:
## Purpose
There was existing logic for type annotating the other concat methods, however the standard `CONCAT` was missed out

## Description of changes
- Implemented string type for `CONCAT`
- Added unit test to cover it

### DIFF
--- a/sqlglot/optimizer/annotate_types.py
+++ b/sqlglot/optimizer/annotate_types.py
@@ -108,6 +108,7 @@ class TypeAnnotator:
         exp.If: lambda self, expr: self._annotate_by_args(expr, "true", "false"),
         exp.Coalesce: lambda self, expr: self._annotate_by_args(expr, "this", "expressions"),
         exp.IfNull: lambda self, expr: self._annotate_by_args(expr, "this", "expression"),
+        exp.Concat: lambda self, expr: self._annotate_with_type(expr, exp.DataType.Type.VARCHAR),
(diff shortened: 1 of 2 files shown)


GeorgeSittas:
Thanks for the PR!

Just one comment: I'm not sure if defaulting to `VARCHAR` is correct, because e.g. `CONCAT(NULL)` produces`NULL`, which in SQLGlot terms would have type `NULL`.

tobymao:
@GeorgeSittas sorry i didn't catch that can you fix it?

GeorgeSittas:
Yeah, just wanted to note it here in case I missed something.

tobymao:
@GeorgeSittas if this is true, isn't Avg wrong? 

```      
  exp.Avg: lambda self, expr: self._annotate_with_type(expr, exp.DataType.Type.DOUBLE),
```

tobymao:
basically, every annotate_with_type is wrong isn't it?

hollandjake:
interesting! i've never come across using NULL in a concat. Obviously if it was a field param which just happened to be NULL then we wouldn't be able to handle that. but if its a NULL constant that would be doable

GeorgeSittas:
> interesting! i've never come across using NULL in a concat. Obviously if it was a field param which just happened to be NULL then we wouldn't be able to handle that. but if its a NULL constant that would be doable

Right, I think we'd discussed about this issue in the past and decided to leave the annotations as is for now, because we don't support information about nullability etc in the optimizer. You can ignore my comment and if needed I'll revisit this later.

PR #1322, opened 2023-03-21
added sqlite glob function

mpf82:
added sqlite glob function

reference: https://www.sqlite.org/lang_corefunc.html#glob

### DIFF
--- a/sqlglot/dialects/sqlite.py
+++ b/sqlglot/dialects/sqlite.py
@@ -37,6 +37,11 @@ class SQLite(Dialect):
             "EDITDIST3": exp.Levenshtein.from_arg_list,
         }
 
+        FUNC_TOKENS = {
+            *parser.Parser.FUNC_TOKENS,
+            TokenType.GLOB,
+        }
+
     class Generator(generator.Generator):
         TYPE_MAPPING = {
             **generator.Generator.TYPE_MAPPING,  # type: ignore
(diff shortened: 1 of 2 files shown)


mpf82:
I just notived that duckdb also has a glob function: https://duckdb.org/docs/sql/functions/patternmatching.html#glob-function-to-find-filenames

should I add ``TokenType.GLOB`` to FUNC_TOKENS in parser.py instead?

GeorgeSittas (approved):

mpf82:
> This doesn't seem correct because we're not parsing `GLOB(...)` into an `exp.Glob` expression. I think `exp.Func` also needs to inherit from `exp.Func`.

Sorry, what seems not correct?

I was checking how it's done for the ``LIKE`` function:

``SELECT LIKE('%b%', 'abc')``
> (SELECT expressions:
  (ANONYMOUS this: LIKE, expressions:
    (LITERAL this: %b%, is_string: True),
    (LITERAL this: abc, is_string: True)))

and with this PR:

``SELECT GLOB('*y*', 'xyz')``
> (SELECT expressions:
  (ANONYMOUS this: GLOB, expressions:
    (LITERAL this: *y*, is_string: True),
    (LITERAL this: xyz, is_string: True)))

GeorgeSittas:
Since we have the `exp.Glob` expression, I think we should use it for representing `GLOB(x, y)`. The same applies to `LIKE`; I wasn't aware this is how we currently represent the `LIKE(x, y)` function in an AST level.

Note that `x LIKE y` is implemented as `LIKE(y, x)` according to the [documentation](https://www.sqlite.org/lang_expr.html). Same goes for `GLOB`.

PR #1324, opened 2023-03-21
added PRAGMA statement

mpf82:
SQLITE: https://www.sqlite.org/pragma.html

DuckDB: https://duckdb.org/docs/sql/pragmas.html

PL/SQL: https://juliandontcheff.wordpress.com/2023/01/03/the-106-oracle-pl-sql-pragmas/

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1191,6 +1191,17 @@ class Generator:
         )
         return f"SET{expressions}"
 
+    def pragma_sql(self, expression: exp.Pragma) -> str:
+        this = self.sql(expression, "this")
+        schema = self.sql(expression, "schema")
+        schema = f"{schema}." if schema else ""
+        value = self.sql(expression, "value")
+        func = expression.args.get("func")
+        if func:
+            return f"PRAGMA {schema}{this}({value})"
+        value = f" = {value}" if value else ""
+        return f"PRAGMA {schema}{this}{value}"
+
     def lock_sql(self, expression: exp.Lock) -> str:
         if self.LOCKING_READS_SUPPORTED:
             lock_type = "UPDATE" if expression.args["update"] else "SHARE"
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1196,8 +1196,7 @@ class Generator:
         schema = self.sql(expression, "schema")
         schema = f"{schema}." if schema else ""
         value = self.sql(expression, "value")
-        func = expression.args.get("func")
-        if func:
+        if expression.args.get("func"):
             return f"PRAGMA {schema}{this}({value})"
         value = f" = {value}" if value else ""
         return f"PRAGMA {schema}{this}{value}"
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -461,6 +461,7 @@ class Parser(metaclass=_Parser):
         TokenType.INSERT: lambda self: self._parse_insert(),
         TokenType.LOAD_DATA: lambda self: self._parse_load_data(),
         TokenType.MERGE: lambda self: self._parse_merge(),
+        TokenType.PRAGMA: lambda self: self._parse_pragma(),
         TokenType.ROLLBACK: lambda self: self._parse_commit_or_rollback(),
         TokenType.SET: lambda self: self._parse_set(),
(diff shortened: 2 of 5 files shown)


sqlglot/tokens.py
```diff
@@ -252,6 +252,7 @@ class TokenType(AutoName):
     PERCENT = auto()
     PIVOT = auto()
     PLACEHOLDER = auto()
+    PRAGMA = auto()
```
GeorgeSittas, line 255: Let's include this in `ID_VAR_TOKENS` as well.

sqlglot/parser.py
```diff
@@ -3880,6 +3881,25 @@ def _parse_merge(self) -> exp.Expression:
             expressions=whens,
         )
 
+    def _parse_pragma(self) -> t.Optional[exp.Expression]:
+        this = self._parse_primary() or self._parse_id_var()
+        schema = None
+        value = None
+        func = None
+
+        while self._match(TokenType.DOT):
```
GeorgeSittas, line 3890: I think this can be converted to an if statement instead.

sqlglot/parser.py
```diff
@@ -3880,6 +3881,25 @@ def _parse_merge(self) -> exp.Expression:
...
+            schema = this
+            this = self._parse_primary() or self._parse_id_var()
+
+        if self._match(TokenType.EQ):
+            self._match(TokenType.EQ)
```
GeorgeSittas, line 3895: We've already consumed the EQ token here, so this seems redundant unless there's a syntax form I missed that expects `==` as a token.

sqlglot/parser.py
```diff
@@ -3880,6 +3881,25 @@ def _parse_merge(self) -> exp.Expression:
...
+        this = self._parse_primary() or self._parse_id_var()
+        schema = None
+        value = None
+        func = None
+
+        while self._match(TokenType.DOT):
+            schema = this
+            this = self._parse_primary() or self._parse_id_var()
+
+        if self._match(TokenType.EQ):
```
GeorgeSittas, line 3894: I think we can dry this out to

```python
func = self._match(TokenType.L_PAREN)
self._match(TokenType.EQ)
value = self._parse_var_or_string() or self._parse_number() or self._parse_id_var()
self._match(TokenType.R_Paren)
```

sqlglot/generator.py
```diff
@@ -1191,6 +1191,17 @@ def set_sql(self, expression: exp.Set) -> str:
         )
         return f"SET{expressions}"
 
+    def pragma_sql(self, expression: exp.Pragma) -> str:
+        this = self.sql(expression, "this")
+        schema = self.sql(expression, "schema")
+        schema = f"{schema}." if schema else ""
+        value = self.sql(expression, "value")
+        func = expression.args.get("func")
+        if func:
```
GeorgeSittas, line 1200:
```suggestion
        if expression.args.get("func"):
```

GeorgeSittas (commented):
Looks good, just a few comments

sqlglot/parser.py
mpf82: yeah, that's a leftover from a previous attempt where I checked the token w/o advancing

sqlglot/parser.py
mpf82: Changed it the way you suggested, but personally, I think the readability suffers.

GeorgeSittas (approved):

sqlglot/parser.py
GeorgeSittas: Thanks for making the changes, happy to revert it later on if Toby agrees about the readability issue. It's a common pattern to dry out things like this in the codebase, so I'm probably biased towards this style more.

sqlglot/parser.py
```diff
@@ -3880,6 +3882,22 @@ def _parse_merge(self) -> exp.Expression:
             expressions=whens,
         )
 
+    def _parse_pragma(self) -> t.Optional[exp.Expression]:
+        this = self._parse_primary() or self._parse_id_var()
+        schema = None
+        value = None
```
tobymao, line 3888:
```suggestion
```

tobymao:
@mpf82 i've greatly simplified pragma with e8cff1cb

PR #1329, opened 2023-03-22
Parse GLOB and LIKE functions into exp.Glob, exp.Like

GeorgeSittas:


mpf82:
Awesome, however sqlite supports ``like(X,Y,Z)`` which should translate to ``Y LIKE X ESCAPE Z``

Currently the ``z`` parameter is ignored.

Expected: ``r"SELECT LIKE('%y%', 'xyz', '\')"`` -> ``r"SELECT 'xyz' LIKE '%y%' ESCAPE '\'"``

GeorgeSittas:
Cool, should be easy to add that. I'll look into it soon :-)

GeorgeSittas:
@mpf82 https://github.com/tobymao/sqlglot/pull/1336

PR #1332, opened 2023-03-23
simplify nested connectors

barakalon:


sqlglot/optimizer/simplify.py
```diff
@@ -37,10 +37,10 @@ def _simplify(expression, root=True):
         exp.replace_children(node, lambda e: _simplify(e, False))
         node = simplify_not(node)
         node = flatten(node)
-        node = simplify_connectors(node)
+        node = simplify_connectors(node, root)
```
tobymao, line 40: seems like root is missing in a couple of the other functions, let's add root for every function that relies on same_parent

barakalon: Done. But it bled into normalize.

tobymao:
#1333

PR #1335, opened 2023-03-23
Optimizer: fix star expansion when USING clause is present

GeorgeSittas:
Before:

```python
>>>
>>> optimize("SELECT * FROM x JOIN y USING(id)", schema={"x": {"id": "int"}, "y": {"id": "int"}}).sql()
'SELECT "x"."id" AS "id", "y"."id" AS "id" FROM "x" AS "x" JOIN "y" AS "y" ON "x"."id" = "y"."id"'
```

This transformation doesn't preserve the semantics of the `USING` clause, because it expands `id` into two columns, both named `"id"`, instead of merging them into a single one. For example, in Postgres SQL:

```sql
-- Tables x and y both have a single column named "id", with values 1, 2 and 3.
georgesittas=# SELECT * FROM x JOIN y USING(id);
 id
----
  1
  2
  3
(3 rows)
```

After:

```python
>>>
>>> optimize("SELECT * FROM x JOIN y USING(id)", schema={"x": {"id": "int"}, "y": {"id": "int"}}).sql()
'SELECT COALESCE("x"."id", "y"."id") AS "id" FROM "x" AS "x" JOIN "y" AS "y" ON "x"."id" = "y"."id"'
```

Notice that this matches the behavior of Postgres and is aligned with the current `_expand_using` [behavior](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/qualify_columns.py#L129):

```sql
georgesittas=# SELECT COALESCE(x.id, y.id) AS id FROM x AS x JOIN y AS y ON x.id = y.id'
 id
----
  1
  2
  3
(3 rows)
```

Some more examples I tried (including left and right joins) can be found [here](https://user-images.githubusercontent.com/46752250/227366023-44d24fb9-45ad-46a3-8bcd-5911509d5850.png).

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -30,10 +30,10 @@ def qualify_columns(expression, schema):
         resolver = Resolver(scope, schema)
         _pop_table_column_aliases(scope.ctes)
         _pop_table_column_aliases(scope.derived_tables)
-        _expand_using(scope, resolver)
+        using_column_tables = _expand_using(scope, resolver)
         _qualify_columns(scope, resolver)
         if not isinstance(scope.expression, exp.UDTF):
-            _expand_stars(scope, resolver)
+            _expand_stars(scope, resolver, using_column_tables)
             _qualify_outputs(scope)
         _expand_group_by(scope, resolver)
         _expand_order_by(scope)
@@ -133,6 +133,8 @@ def _expand_using(scope, resolver):
                     replacement = exp.alias_(replacement, alias=column.name)
 
                 scope.replace(column, replacement)
+
+    return column_tables
 
 
 def _expand_group_by(scope, resolver):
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -258,12 +260,18 @@ def _qualify_columns(scope, resolver):
...
     """Expand stars to lists of column selections"""
 
     new_selections = []
     except_columns = {}
     replace_columns = {}
+    coalesced_columns = set()
+
+    # Create a mapping of automatically joined column names to sets of
+    # the corresponding tables, to make the lookup more efficient
+    ordered_column_tables = using_column_tables
```
barakalon, line 273: Perhaps we should use some kind of ordered set when we originally construct this object? I.e. a dict with `None` values.

barakalon (approved):

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: We could, yeah. I think I like this idea as it'll simplify the code here.

GeorgeSittas: On the other hand, maybe the current approach is a bit cleaner.. Doing `tables[table] = None` in `_expand_using` might look a bit confusing.

GeorgeSittas: Done

sqlglot/optimizer/qualify_columns.py
```diff
@@ -258,12 +261,13 @@ def _qualify_columns(scope, resolver):
...
 
 
-def _expand_stars(scope, resolver):
+def _expand_stars(scope, resolver, using_column_tables):
     """Expand stars to lists of column selections"""
 
     new_selections = []
     except_columns = {}
     replace_columns = {}
+    coalesced_columns = set()
```
tobymao, line 270: do you need this? can you just pop using_column_tables?

tobymao: guess not

PR #1340, opened 2023-03-24
Parse tilde as REGEXP_MATCHES for DuckDB

### LINKED ISSUE #1338: Tilde operator fails to parse with the duckdb dialect
Fails for duckdb, but works for postgres:

```
In [1]: import sqlglot as sg

In [2]: sg.__version__
Out[2]: '11.4.1'

In [3]: sg.parse_one("select x ~ y", read="postgres")
Out[3]:
(SELECT expressions:
  (REGEXPLIKE this:
    (COLUMN this:
      (IDENTIFIER this: x, quoted: False)), expression:
    (COLUMN this:
      (IDENTIFIER this: y, quoted: False))))

In [4]: sg.parse_one("select x ~ y", read="duckdb")
...
ParseError: Invalid expression / Unexpected token. Line 1, Col: 10.
  select x ~ y
```

`~` is a valid duckdb operator:

```
❯ duckdb
v0.7.1
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D select 'x' ~ 'y' as m;
┌─────────┐
│    m    │
│ boolean │
├─────────┤
│ false   │
└─────────┘
```

GeorgeSittas:
Fixes #1338

This isn't complete, because `~5` is a valid expression, corresponding to the 2's complement of 5. I assume that it's not used as commonly, so maybe it's ok to leave it as is. Otherwise, happy to revisit this and try to fix parsing so that we can recognize it too. Notice that we also had this problem with postgres prior to this PR.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -80,6 +80,7 @@ class DuckDB(Dialect):
     class Tokenizer(tokens.Tokenizer):
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
+            "~": TokenType.RLIKE,
             ":=": TokenType.EQ,
             "ATTACH": TokenType.COMMAND,
             "BINARY": TokenType.VARBINARY,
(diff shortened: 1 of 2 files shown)


tobymao (approved):

cpcloud:
Thank you!

Issue #1349, opened 2023-03-27
JSON_EXTRACT_SCALAR is not a valid function in MySQL


jerryleooo:
```
import sqlglot
s = '''select * from requests where get_json_object(stream_data, '$.data.results') is not null'''
sqlglot.transpile(s, read='hive', write='mysql')
```

the result is `["SELECT * FROM requests WHERE NOT JSON_EXTRACT_SCALAR(stream_data, '$.data.results') IS NULL"]`, while JSON_EXTRACT_SCALAR is not a valid function in MySQL.

GeorgeSittas:
What is the equivalent way of expressing this in MySQL (+bonus for docs)? JSON handling is non-standard, so it's not sufficiently covered yet across many dialects.

jerryleooo:
Hi @GeorgeSittas , understood, and I guess it's very hard to determine if the extracted value is scalar or not without schema information.

and if I am correct, the equivalent function is `JSON_EXTRACT `,
MySQL 5.7: https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-extract
MySQL 8.0: https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-extract

I tried making a fix here: https://github.com/tobymao/sqlglot/pull/1350

GeorgeSittas:
Cool, thanks for the PR @jerryleooo!

PR #1350, opened 2023-03-27
fix MySQL json_extract_scalar

jerryleooo:
Fix https://github.com/tobymao/sqlglot/issues/1349


sqlglot/dialects/mysql.py
```diff
@@ -404,6 +404,8 @@ class Generator(generator.Generator):
             exp.DayOfYear: rename_func("DAYOFYEAR"),
             exp.WeekOfYear: rename_func("WEEKOFYEAR"),
             exp.GroupConcat: lambda self, e: f"""GROUP_CONCAT({self.sql(e, "this")} SEPARATOR {self.sql(e, "separator") or "','"})""",
+            exp.JSONExtract: rename_func("JSON_EXTRACT"),
```
GeorgeSittas, line 407: This isn't needed.

GeorgeSittas (approved):
Looks good, let's just remove the redundant `exp.JSONExtract` dict entry.

sqlglot/dialects/mysql.py
jerryleooo: correct, removed

tobymao (approved):

PR #1362, opened 2023-03-29
Fix parsing for the logarithm function

### LINKED ISSUE #1360: Log10 vs LN
sqlglot does not transpile SQL using log() correctly. 

sqlglot.transpile("select exp(log(2))", read="mysql", write="duckdb") incorrectly outputs `select exp(log(2))` when it should be `select exp(ln(2))`

```
select exp(log(2)) 
```

Returns different results on different platforms. 

In MySQL and SQL Server, [log](https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html) is the natural log, so the result of the statement is 2

In PostGres and DuckDB, for instance, [log](https://www.postgresql.org/docs/9.1/functions-math.html) is log base 10, so the result is 1.35.

GeorgeSittas:
Fixes #1360

References:
- https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#log
- https://clickhouse.com/docs/en/sql-reference/functions/math-functions
- https://docs.databricks.com/sql/language-manual/functions/log.html
- https://duckdb.org/docs/sql/functions/numeric.html
- https://cwiki.apache.org/confluence/display/hive/languagemanual+udf
- https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_log
- https://docs.oracle.com/cd/B13789_01/server.101/b10759/functions070.htm
- https://www.postgresql.org/docs/9.1/functions-math.html
- https://prestodb.io/docs/current/functions/math.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_LOG.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_LOG.html
- https://docs.snowflake.com/en/sql-reference/functions/log
- https://spark.apache.org/docs/2.3.0/api/sql/index.html
- https://www.sqlite.org/lang_mathfunc.html#log
- https://docs.teradata.com/r/Teradata-Database-SQL-Functions-Operators-Expressions-and-Predicates/March-2017/Arithmetic-Trigonometric-Hyperbolic-Operators/Functions/LOG-Function
- https://learn.microsoft.com/en-us/sql/t-sql/functions/log-transact-sql?view=sql-server-ver16
- https://drill.apache.org/docs/math-and-trig/

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -213,6 +213,9 @@ class BigQuery(Dialect):
             ),
         }
 
+        LOGARITHM_BASE_FIRST = False
+        LOGARITHM_DEFAULTS_TO_LN = True
+
     class Generator(generator.Generator):
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,  # type: ignore
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -213,8 +213,8 @@ class BigQuery(Dialect):
             ),
         }
 
-        LOGARITHM_BASE_FIRST = False
-        LOGARITHM_DEFAULTS_TO_LN = True
+        LOG_BASE_FIRST = False
+        LOG_DEFAULTS_TO_LN = True
 
     class Generator(generator.Generator):
         TRANSFORMS = {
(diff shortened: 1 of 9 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -213,6 +213,9 @@ class Parser(parser.Parser):
             ),
         }
 
+        LOGARITHM_BASE_FIRST = False
+        LOGARITHM_DEFAULTS_TO_LN = True
```
tobymao, line 217: LOG_BASE_FIRST
LOG_DEFAULTS_TO_LN

tobymao (approved):

PR #1374, opened 2023-04-01
Parse DECODE function into CASE when it has >2 args

GeorgeSittas:
- https://docs.snowflake.com/en/sql-reference/functions/decode
- https://docs.aws.amazon.com/redshift/latest/dg/r_DECODE_expression.html
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/DECODE.html
- https://spark.apache.org/docs/latest/api/sql/index.html#decode

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -8,6 +8,7 @@ from sqlglot.errors import ErrorLevel, UnsupportedError, concat_messages
 from sqlglot.helper import apply_index_offset, csv, seq_get, should_identify
 from sqlglot.time import format_time
 from sqlglot.tokens import TokenType
+from sqlglot.transforms import DECODE_TO_CASE
 
 logger = logging.getLogger("sqlglot")
 
@@ -55,6 +56,7 @@ class Generator:
     """
 
     TRANSFORMS = {
+        **DECODE_TO_CASE,  # type: ignore
         exp.DateAdd: lambda self, e: self.func(
             "DATE_ADD", e.this, e.expression, exp.Literal.string(e.text("unit"))
         ),
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -8,7 +8,6 @@ from sqlglot.errors import ErrorLevel, UnsupportedError, concat_messages
 from sqlglot.helper import apply_index_offset, csv, seq_get, should_identify
 from sqlglot.time import format_time
 from sqlglot.tokens import TokenType
-from sqlglot.transforms import DECODE_TO_CASE
 
 logger = logging.getLogger("sqlglot")
 
@@ -56,7 +55,6 @@ class Generator:
     """
 
     TRANSFORMS = {
-        **DECODE_TO_CASE,  # type: ignore
         exp.DateAdd: lambda self, e: self.func(
             "DATE_ADD", e.this, e.expression, exp.Literal.string(e.text("unit"))
         ),
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -674,6 +674,7 @@ class Parser(metaclass=_Parser):
     FUNCTION_PARSERS: t.Dict[str, t.Callable] = {
         "CAST": lambda self: self._parse_cast(self.STRICT_CAST),
         "CONVERT": lambda self: self._parse_convert(self.STRICT_CAST),
+        "DECODE": lambda self: self._parse_decode(),
         "EXTRACT": lambda self: self._parse_extract(),
         "JSON_OBJECT": lambda self: self._parse_json_object(),
         "LOG": lambda self: self._parse_logarithm(),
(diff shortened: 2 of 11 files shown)


sqlglot/transforms.py
```diff
@@ -133,6 +133,48 @@ def remove_precision_parameterized_types(expression: exp.Expression) -> exp.Expr
...
+    needs special treatment in CASE expressions, since we need to explicitly check for them using
+    the `IS NULL` predicate, instead of relying on pattern matching.
+
+    - https://stackoverflow.com/a/3209948/17518317
+    - https://docs.snowflake.com/en/sql-reference/functions/decode
+    """
+    if isinstance(expression, exp.Matches):
+        from sqlglot.optimizer.simplify import simplify
+
+        select = expression.this
```
GeorgeSittas, line 148: The name `select` here was picked to be consistent with Snowflake's documentation:

Screenshot 2023-04-01 at 10 20 08 PM

sqlglot/generator.py
```diff
@@ -55,6 +56,7 @@ class Generator:
     """
 
     TRANSFORMS = {
+        **DECODE_TO_CASE,  # type: ignore
```
tobymao, line 59: this doesn't seem like the correct approach. if you want all dialects to convert decode, then you should just do it at parse time.

sqlglot/transforms.py
```diff
@@ -133,6 +133,48 @@ def remove_precision_parameterized_types(expression: exp.Expression) -> exp.Expr
...
+        expressions = expression.expressions
+
+        ifs = []
+        for search, result in zip(expressions[::2], expressions[1::2]):
+            search = simplify(search)
```
tobymao, line 153: why do you need simpliify

sqlglot/transforms.py
```diff
@@ -133,6 +133,48 @@ def remove_precision_parameterized_types(expression: exp.Expression) -> exp.Expr
...
+
+            if isinstance(search, exp.Literal):
+                ifs.append(exp.If(this=exp.EQ(this=select, expression=search), true=result))
+            elif isinstance(search, exp.Null):
+                ifs.append(exp.If(this=exp.Is(this=select, expression=exp.Null()), true=result))
+            else:
+                cond = exp.Or(
```
tobymao, line 160: exp.or_ and and

sqlglot/generator.py
GeorgeSittas: It's currently not converted for all dialects, i.e. `DECODE(...)` is transpiled to itself for Snowflake, Redshift, Oracle. Do you think we should just do this at parse time and convert it regardless of the dialect?

sqlglot/transforms.py
GeorgeSittas: It's not needed per se. The reason I used it is so that we don't produce the `IS NULL` checks for expressions consisting of literals that can be reduced to a single literal value, like arithmetic expressions. No problem with removing this completely -- maybe the overhead is not worth it.

sqlglot/transforms.py
GeorgeSittas: Thanks, will fix.

sqlglot/generator.py
tobymao: how is it not converted? is it because they override match to decode?

sqlglot/transforms.py
tobymao: let’s remove it for now

sqlglot/generator.py
GeorgeSittas: yep

tobymao: this doesn’t follow the sqlglot pattern. the generic dialect should also handle match or decode. 

languages that don’t support decode should use this transform.

tobymao: this way if you do snowflake -> generic -> snowflake it’s still uses decode.

otherwise you lose it

sqlglot/parser.py
```diff
@@ -3345,6 +3346,51 @@ def _parse_convert(self, strict: bool) -> t.Optional[exp.Expression]:
...
+        needs special treatment, since we need to explicitly check for it with `IS NULL`,
+        instead of relying on pattern matching.
+        """
+        args = self._parse_csv(self._parse_conjunction)
+
+        if len(args) < 3:
+            return self.expression(exp.Decode, this=seq_get(args, 0), charset=seq_get(args, 1))
+
+        expression, *expressions = args
+        if not expression:
```
tobymao, line 3366: why would expression ever be None?

tobymao (approved):

sqlglot/parser.py
GeorgeSittas: Mypy was yelling that anything in args could be none due to parse csv, so the following copies would fail with attribute error.

PR #1387, opened 2023-04-06
added date functions in executor

laurentiupiciu:


tobymao:
i'm going to alter these to match sqlglot's syntax, for example getdate should really be currentdate

PR #1394, opened 2023-04-07
Fix: attempt to transpile MySQL DATE_FORMAT to Snowflake

### LINKED ISSUE #1393: Translate dialect - MySQL to Snowflake
Hello,

I'm trying to translate MySQL dialect into Snowflake dialect.
I did some tests, but the translation doesn't seem to work.

> It can be used to format SQL or translate between [19 different dialects](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/__init__.py) like [DuckDB](https://duckdb.org/), [Presto](https://prestodb.io/), [Spark](https://spark.apache.org/), [Snowflake](https://www.snowflake.com/en/), and [BigQuery](https://cloud.google.com/bigquery/).

**Actual behaviour**
```python
sqlglot.transpile("select DATE_FORMAT(created_date, '%Y-%m')", read="mysql", write="snowflake")[0]
# "SELECT DATE_FORMAT(created_date, '%Y-%m')"
```

**Expected behaviour**
```python
sqlglot.transpile("select DATE_FORMAT(created_date, '%Y-%m')", read="mysql", write="snowflake")[0]
# "SELECT to_char(created_date, 'yyyy-mm')"
```

`sqlglot==11.5.2`

Thanks!

GeorgeSittas:
Fixes #1393

<details><summary>Outdated description, see my next comment.</summary>
- Improved type hints.
- Mapped `"DATE_FORMAT"` to `exp.DateToDateStr` for MySQL, which is also generated back to `"DATE_FORMAT"`.
- Transpiled `exp.DateToDateStr` to `TO_CHAR` for Snowflake.

The last one is tricky.. For the time being I simply used a "hack" to detect whether `'%'` appears in the format string, and if it does I use Snowflake's inverse time mapping to try and get a valid format (see commented doc link for valid formats). This is currently a best-effort approach, as:

1. not all MySQL formats are included in Snowflake's time mapping and
2. there are some duplicate values in this dict, so it's not a 1-1 mapping

Some basic formats are working fine, but others still don't work. Happy to try a different approach, hopefully there's a simpler way to tackle this.
</details>

cc: @4sushi

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -20,7 +20,6 @@ from sqlglot.dialects.dialect import (
 from sqlglot.expressions import Literal
 from sqlglot.helper import flatten, seq_get
 from sqlglot.parser import binary_range_parser
-from sqlglot.time import format_time
 from sqlglot.tokens import TokenType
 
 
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -111,27 +116,45 @@ def _parse_date_part(self):
     return self.expression(exp.Extract, this=this, expression=expression)
 
 
+def _datetodatestr_sql(self: generator.Generator, expression: exp.DateToDateStr) -> str:
+    date_format = expression.args.get("format")
+    if not date_format:
+        return f"TO_CHAR({self.sql(expression, 'this')})"
```
tobymao, line 122: self.func("TO_CHAR", expression.this)

GeorgeSittas:
I studied `format_time_lambda` a bit and I realized my implementation could be greatly simplified. I almost replicated what we do for Drill, with the addition that I added the cast to `TIMESTAMP` for Snowflake (we need this otherwise `TO_CHAR` breaks if the first argument is not a date/time/timestamp value, i.e. a simple string).

Also added and fixed several tests. I've manually tested these changes across all affected dialects, they looked good (hopefully didn't missed something).

tobymao (approved):

PR #1396, opened 2023-04-07
Add alias to column lineage nodes

vchan:
Track the source of a node in an alias attribute.

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -71,7 +71,11 @@ def lineage(
     optimized = optimize(expression, schema=schema, rules=rules)
     scope = build_scope(optimized)
     tables: t.Dict[str, Node] = {}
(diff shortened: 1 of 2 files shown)


sqlglot/lineage.py
```diff
@@ -70,6 +71,7 @@ def lineage(
     optimized = optimize(expression, schema=schema, rules=rules)
     scope = build_scope(optimized)
     tables: t.Dict[str, Node] = {}
+    aliases = {dt.alias: dt.comments[0][len("source: ") :] for dt in scope.derived_tables}
```
GeorgeSittas, line 74:
```suggestion
    aliases = {dt.alias: dt.comments[0].split()[1] for dt in scope.derived_tables}
```

tobymao (approved):

GeorgeSittas (approved):

PR #1408, opened 2023-04-10
Fix: ensure name normalization for schema methods

GeorgeSittas:
This PR aims to make the normalization of names in `MappingSchema` more robust. We're already normalizing names, as introduced in https://github.com/tobymao/sqlglot/pull/891, but that's only limited to the schema constructor. All of the exposed methods (`add_table`, `column_names`, `get_column_type`) don't normalize their input arguments.

cc: @tobymao @barakalon

### DIFF
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -179,6 +179,62 @@ class MappingSchema(AbstractMappingSchema[t.Dict[str, str]], Schema):
             }
         )
 
+    def add_table(
+        self, table: exp.Table | str, column_mapping: t.Optional[ColumnMapping] = None
+    ) -> None:
+        """
+        Register or update a table. Updates are only performed if a new column mapping is provided.
+
+        Args:
+            table: the `Table` expression instance or string representing the table.
+            column_mapping: a column mapping that describes the structure of the table.
+        """
+        normalized_table = self._normalize_table(self._ensure_table(table))
+        normalized_column_mapping = {
+            self._normalize_name(key): value
+            for key, value in ensure_column_mapping(column_mapping).items()
+        }
+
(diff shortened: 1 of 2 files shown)


sqlglot/schema.py
```diff
@@ -206,84 +262,40 @@ def _normalize(self, schema: t.Dict) -> t.Dict:
...
 
     def _normalize_name(self, name: str) -> str:
         try:
-            identifier: t.Optional[exp.Expression] = sqlglot.parse_one(
-                name, read=self.dialect, into=exp.Identifier
+            identifier = t.cast(
+                exp.Identifier, sqlglot.parse_one(name, read=self.dialect, into=exp.Identifier)
             )
         except:
-            identifier = exp.to_identifier(name)
```
tobymao, line 239: why was this changed? what case triggered the need to change this

GeorgeSittas: There wasn't a case that triggered this change, but it felt reasonable to return `name` here if we are unable to parse it into an identifier. I guess there are two cases to consider here. If `name` can't be parsed and it is converted `to_identifier`, then the resulting `exp.Identifier` is:

- quoted, in which case we'd return `name` (this behaviour is preserved in this PR), or
- not quoted, so we'd return `name.lower()`.

How we deal with the latter case is the only difference here. The question is: why should we convert to an identifier and call `.lower` to begin with, if the name couldn't be parsed as one? Seemed more logical to just return `name`, but I might be wrong. Happy to revert this change if it looks risky.

FWIW, in my original PR Barak had suggested using `parse_one` here too (https://github.com/tobymao/sqlglot/pull/891#discussion_r1062872629), but I can see how `to_identifier` might make this more conservative.

PR #1413, opened 2023-04-12
Fix: improve parsing and transpilation of INTERVAL expressions

### LINKED ISSUE #1410: Support for Presto/Trino's interval syntax
As far as I can tell, most SQL dialects use `interval '1 days'` to specify intervals, where the unit is inside the quotes.

This is not valid in [Presto](https://prestodb.io/docs/current/functions/datetime.html#functions-datetime--page-root) and [Trino](https://trino.io/docs/current/functions/datetime.html#date-and-time-operators), where the value is the only thing inside quotes, and the unit comes after the quotes, i.e. `interval '1' day`, or `interval '3' year`. (Also there's no week interval unit in these dialects.) It looks like `interval '1' day` is also valid syntax in Postgres, so maybe this has to do with translating the interval to `interval '1' day` if the dialect doesn't support `interval '1 days'`.

Is it possible to add this rewrite? I'd be happy to contribute.

I don't really know, but I'd assume there needs to be a Presto override for `Generator.interval_sql()`.

And that the unit needs to be extracted from the value literal?

```terminal
>>> sqlglot.parse_one("select interval '1' day")
(SELECT expressions:
  (INTERVAL this:
    (LITERAL this: 1, is_string: True), unit:
    (IDENTIFIER this: day, quoted: False)))
>>> sqlglot.parse_one("select interval '1 days'")
(SELECT expressions:
  (INTERVAL this:
    (LITERAL this: 1 days, is_string: True)))
```

GeorgeSittas:
Fixes #1410

Note: I've manually tested the changes in the test suite to verify they work.

cc: @vegarsti


vegarsti:
Amazing!

sqlglot/dialects/mysql.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from sqlglot import exp, generator, parser, tokens
+from sqlglot.dialects.bigquery import date_add
```
GeorgeSittas, line 4: At first I thought about drying this up by moving it in `dialect.py`, but we already have a `parse_date_delta` function which seems to do something similar, so I wasn't sure what name to use and left it like this.

tobymao: you should dry it up

GeorgeSittas: Any suggestion for the name? This is used for `DateSub` as well, so ideally I'd use something like `parse_date_delta`, but as I said that's already reserved. The easy way out is to move it to `dialect.py` as `date_add` and move on.

GeorgeSittas: Done

vegarsti: Hadn't heard "dry it up" before, nice

GeorgeSittas: We use it frequently around here :-) See also: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

vegarsti: Yeah, got the DRY reference, just hadn't heard it verbed before 😄

tobymao (approved):

vegarsti:
Is there any set release cadence/triggers? I'd love to start using this instead of the hacky fix we've done in our code 😀

vegarsti:
🙏🏻 Thanks!

GeorgeSittas:
@vegarsti v11.5.4 is deployed

vegarsti:
Excellent, thank you!

vegarsti:
Trino doesn't have the `week` unit -- is it complicated to change week intervals to 7 * day intervals?

GeorgeSittas:
Indeed, it'd be better to have these as separate issues! Feel free to create them and I'll take a look soon.

PR #1414, opened 2023-04-12
Feat(optimizer): Improve qualify columns for self joins without schema

GeorgeSittas:
Before:

```python
>>> optimize("select x from t join t").sql()
# sqlglot.errors.OptimizeError: Unknown table: '' for column 'x'
```

After:

```python
>>> optimize("select x from t join t").sql()
'SELECT "t_2"."x" AS "x" FROM "t" AS "t" JOIN "t" AS "t_2"'
```

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -405,8 +405,9 @@ class Resolver:
 
         Args:
             column_name: The column name to find the table for.
+
         Returns:
-            The table name if it can be found/inferred.
+            The table name, if it can be found/inferred.
         """
         if self._unambiguous_columns is None:
             self._unambiguous_columns = self._get_unambiguous_columns(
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -416,13 +417,25 @@ def get_table(self, column_name: str) -> t.Optional[exp.Identifier]:
...
-                source
-                for source, columns in self._get_all_source_columns().items()
-                if not columns or "*" in columns
-            )
-            if len(sources_without_schema) == 1:
-                table_name = sources_without_schema[0]
+            source_name_to_alias = {}
+
+            # Set of pairs (source_name, is_table)
+            distinct_sources_without_schema = set()
```
GeorgeSittas, line 423: I used a pair here in order to distinguish between `Scope` and `Table` sources, because otherwise we'd get some erroneous transformations in some cases. Consider the example:

```python
SELECT o FROM e AS t1 JOIN (SELECT * FROM t2 AS t2) AS e;
```

Here, the original table name of `t1` is `e`, which is also used as the alias for a `Scope` source, hence we must make sure not to collapse these two different sources into a single one.

barakalon:
Is this right?

`x` is ambiguous in `select x from t join t`. Do some engines support this?

GeorgeSittas:
I'm taking a closer look, did this a bit hastily so you might be right. Marking this as a draft until I validate these changes.

GeorgeSittas:
@barakalon you're right, definitely my bad for not cross-checking this change earlier. I got why it's wrong now, so closing this PR. Thanks!

PR #1416, opened 2023-04-13
Feat(optimizer): Make _expand_using more lenient in qualify columns

GeorgeSittas:
An attempt to make the qualify columns more lenient.

Before:
```python
>>> optimize("SELECT y FROM foo JOIN bar USING(x)", schema={"foo": {"x": "int"}}).sql()
# sqlglot.errors.OptimizeError: Cannot automatically join: x
```

After:
```python
>>> optimize("SELECT y FROM foo JOIN bar USING(x)", schema={"foo": {"x": "int"}}).sql()
'SELECT "bar"."y" AS "y" FROM "foo" AS "foo" JOIN "bar" AS "bar" ON "foo"."x" = "bar"."x"'
```

### DIFF
--- a/tests/fixtures/optimizer/qualify_columns.sql
+++ b/tests/fixtures/optimizer/qualify_columns.sql
@@ -317,6 +317,9 @@ SELECT COALESCE(y.b, z.b) AS b, COALESCE(y.c, z.c) AS c FROM y AS y JOIN z AS z 
 SELECT * FROM y JOIN z USING(b, c) WHERE b = 2 AND c = 3;
 SELECT COALESCE(y.b, z.b) AS b, COALESCE(y.c, z.c) AS c FROM y AS y JOIN z AS z ON y.b = z.b AND y.c = z.c WHERE COALESCE(y.b, z.b) = 2 AND COALESCE(y.c, z.c) = 3;
 
+-- We can safely convert `b` to `x.b` in the following two queries, because the original queries
+-- would be invalid if `b` also existed in `t`'s schema (which we don't know), due to ambiguity.
+
 # execute: false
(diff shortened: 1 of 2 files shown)


tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -317,6 +317,14 @@ SELECT COALESCE(y.b, z.b) AS b, COALESCE(y.c, z.c) AS c FROM y AS y JOIN z AS z
 SELECT * FROM y JOIN z USING(b, c) WHERE b = 2 AND c = 3;
 SELECT COALESCE(y.b, z.b) AS b, COALESCE(y.c, z.c) AS c FROM y AS y JOIN z AS z ON y.b = z.b AND y.c = z.c WHERE COALESCE(y.b, z.b) = 2 AND COALESCE(y.c, z.c) = 3;
 
+# execute: false
+SELECT b FROM x JOIN t USING(a);
```
tobymao, line 321: since it's using, shouldn't the coalesce logic kick in?

GeorgeSittas: FWIW if we knew that `b` belongs to both `x` and `t`, the current behaviour is to raise an optimizer error:

```python
>>> optimize("select x from t1 join t2 using(id)", schema={"t1": {"x": "int", "id": "int"}, "t2": {"x": "int", "id": "int"}}).sql()
# sqlglot.errors.OptimizeError: Unknown table: '' for column 'x'
```

GeorgeSittas: Discussed in Slack.

PR #1423, opened 2023-04-14
Fix: improve parsing of MATCH_RECOGNIZE clause

### LINKED ISSUE #1422: bug in match recognize implementation
I'm using sqlglot to parse some oracle SQL and noticed a parseError is raised with some sample code from Oracle docs.

source: https://oracle-base.com/articles/12c/pattern-matching-in-oracle-database-12cr1
```
SELECT *
FROM   sales_history MATCH_RECOGNIZE (
         PARTITION BY product
         ORDER BY tstamp
         MEASURES  STRT.tstamp AS start_tstamp,
                   LAST(UP.tstamp) AS peak_tstamp,
                   LAST(DOWN.tstamp) AS end_tstamp,
                   MATCH_NUMBER() AS mno
         ONE ROW PER MATCH
         AFTER MATCH SKIP TO LAST DOWN
         PATTERN (STRT UP+ FLAT* DOWN+)
         DEFINE
           UP AS UP.units_sold > PREV(UP.units_sold),
           FLAT AS FLAT.units_sold = PREV(FLAT.units_sold),
           DOWN AS DOWN.units_sold < PREV(DOWN.units_sold)
       ) MR
```

on parsing this statement i get 
```
Exception has occurred: ParseError
Expecting ). Line 5, Col: 47.
          PARTITION BY product
         ORDER BY tstamp
         MEASURES  STRT.tstamp AS start_tstamp[4m,[0m
                   LAST(UP.tstamp) AS peak_tstamp,
                   LAST(DOWN.tstamp) AS end_tsta
  File "C:\gitRepos\parse\main.py", line 12, in <module>
    stmnt = parser.parse(sql=sql)
sqlglot.errors.ParseError: Expecting ). Line 5, Col: 47.
          PARTITION BY product
         ORDER BY tstamp
         MEASURES  STRT.tstamp AS start_tstamp[4m,[0m
                   LAST(UP.tstamp) AS peak_tstamp,
                   LAST(DOWN.tstamp) AS end_tsta
```

after some digging in the codebase i figured there is an implementation error in both parsing MEASURES and DEFINE,
where they both dont take into account several measures or definitions.

the fix for measures seems simple enough:
```
def _parse_match_recognize(self) -> t.Optional[exp.Expression]:
    if not self._match(TokenType.MATCH_RECOGNIZE):
        return None
    self._match_l_paren()

    partition = self._parse_partition_by()
    order = self._parse_order()
    measures = (
        #self._parse_alias(self._parse_conjunction())
        self._parse_csv(self._parse_expression)
        if self._match_text_seq("MEASURES")
        else None
    )...
```

measures parses successfully (while assuming UP, DOWN and STRT are tables which might be a misnomer? maybe a new symbol token should be introduced?)
but now define has kind of the same bug:
```
  File "C:\gitRepos\parse\main.py", line 12, in <module>
    stmnt = parser.parse(sql=sql)
sqlglot.errors.ParseError: Expecting ). Line 13, Col: 20.
   MATCH SKIP TO LAST DOWN
         PATTERN (STRT UP+ FLAT* DOWN+)
         DEFINE
           UP AS UP[4m.[0munits_sold > PREV(UP.units_sold),
           FLAT AS FLAT.units_sold = PREV(FLAT.units_sold),
```

but for the DEFINE clause, self._parse_expression wont cut it because the symbol(alias) comes before the expression,
so i guess a new function should be defined.

GeorgeSittas:
Fixes #1422

References:
- https://oracle-base.com/articles/12c/pattern-matching-in-oracle-database-12cr1
- https://docs.oracle.com/cd/E29542_01/apirefs.1111/e12048/pattern_recog.htm#CQLLR2208
- https://docs.snowflake.com/en/sql-reference/constructs/match_recognize

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1476,6 +1476,7 @@ class MatchRecognize(Expression):
         "after": False,
         "pattern": False,
         "define": False,
+        "alias": False,
     }
 
 
@@ -3167,9 +3168,9 @@ class Neg(Unary):
     pass
 
 
-# Special Functions
 class Alias(Expression):
-    arg_types = {"this": True, "alias": False}
+    # "reverse" controls the relative order of "this" and "alias" for SQL generation
+    arg_types = {"this": True, "alias": False, "reverse": False}
 
     @property
     def output_name(self):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3169,8 +3169,7 @@ class Neg(Unary):
 
 
 class Alias(Expression):
-    # "reverse" controls the relative order of "this" and "alias" for SQL generation
-    arg_types = {"this": True, "alias": False, "reverse": False}
+    arg_types = {"this": True, "alias": False}
 
     @property
     def output_name(self):
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1329,16 +1329,16 @@ class Generator:
     def matchrecognize_sql(self, expression: exp.MatchRecognize) -> str:
         partition = self.partition_by_sql(expression)
         order = self.sql(expression, "order")
-        measures = self.sql(expression, "measures")
-        measures = self.seg(f"MEASURES {measures}") if measures else ""
+        measures = self.expressions(expression, "measures")
+        measures = self.seg(f"MEASURES{self.seg(measures)}") if measures else ""
         rows = self.sql(expression, "rows")
         rows = self.seg(rows) if rows else ""
         after = self.sql(expression, "after")
         after = self.seg(after) if after else ""
         pattern = self.sql(expression, "pattern")
         pattern = self.seg(f"PATTERN ({pattern})") if pattern else ""
-        define = self.sql(expression, "define")
-        define = self.seg(f"DEFINE {define}") if define else ""
+        define = self.expressions(expression, "define")
+        define = self.seg(f"DEFINE{self.seg(define)}") if define else ""
         body = "".join(
             (
                 partition,
@@ -1350,7 +1350,9 @@ class Generator:
                 define,
             )
         )
-        return f"{self.seg('MATCH_RECOGNIZE')} {self.wrap(body)}"
+        alias = self.sql(expression, "alias")
+        alias = f" {alias}" if alias else ""
+        return f"{self.seg('MATCH_RECOGNIZE')} {self.wrap(body)}{alias}"
 
     def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
         limit = expression.args.get("limit")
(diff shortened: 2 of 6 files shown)


sqlglot/expressions.py
```diff
@@ -3167,9 +3168,9 @@ class Neg(Unary):
     pass
 
 
-# Special Functions
 class Alias(Expression):
-    arg_types = {"this": True, "alias": False}
+    # "reverse" controls the relative order of "this" and "alias" for SQL generation
+    arg_types = {"this": True, "alias": False, "reverse": False}
```
GeorgeSittas, line 3173: I added "reverse" here, because the order is reversed in the DEFINE clause, see test case:

```sql
...
UP AS UP.units_sold > PREV(UP.units_sold),  -- alias goes first...
...
```

There are at least three different ways to make this work:

1) The way I did it in this PR.
2) Swap the values of "this" and "alias". Note that this will create a wrong / misleading AST node and is a hack.
3) Use the `meta` property instead, like `meta["generate_in_reverse"] = True`. This will require creating a helper method for parsing aliases in reverse, because we need to set the meta after the expression is created and so can't be done in the lambda (see code in parser).

This feels like a niche feature, so I'm not sure which approach is the best.

GeorgeSittas:
Simplified the implementation by moving the alias reversing logic inside of `matchrecognize_sql`. I also refactored `Generator.expressions`, so that it can also work with lists of strings instead of a single expression. This has the benefit of allowing us to reuse the prettifying logic for lists of sqls we construct manually, like in this example.

sqlglot/parser.py
```diff
@@ -1972,9 +1971,13 @@ def _parse_match_recognize(self) -> t.Optional[exp.Expression]:
         else:
             pattern = None
 
-        define = (
-            self._parse_alias(self._parse_conjunction()) if self._match_text_seq("DEFINE") else None
+        parse_alias_reverse = lambda: self.expression(
```
tobymao, line 1974: can you put this inline?

sqlglot/generator.py
```diff
@@ -1329,16 +1327,20 @@ def ordered_sql(self, expression: exp.Ordered) -> str:
...
         after = self.seg(after) if after else ""
         pattern = self.sql(expression, "pattern")
         pattern = self.seg(f"PATTERN ({pattern})") if pattern else ""
-        define = self.sql(expression, "define")
-        define = self.seg(f"DEFINE {define}") if define else ""
+        definition_sqls = [
+            f"{self.sql(definition, 'alias')} AS {self.sql(definition, 'this')}"
+            for definition in expression.args.get("define", [])
+        ]
+        definitions = t.cast(str, self.expressions(sqls=definition_sqls))
```
tobymao, line 1342: why do you need to cast this?

GeorgeSittas: Forgot to change, wanted to reuse `definitions` but had some mypy problems. Thanks!

GeorgeSittas: Fixed

sqlglot/parser.py

tobymao (approved):

PR #1424, opened 2023-04-14
Fix(oracle): parsing bug in _parse_xml_table

GeorgeSittas:
Addresses https://github.com/tobymao/sqlglot/commit/bd96e9613fdcd536e50980130074d2cc9e1a004e#commitcomment-108967287.

cc: @mpf82 let me know if converting `getxmltype` to uppercase (context: `dbms_xmlgen.getxmltype(..)`) is ok. Same goes for the new AST representation, which now uses a `Dot` node instead of a `Table`:

```        
(DOT this:
  (VAR this: dbms_xmlgen), expression:
  (ANONYMOUS this: getxmltype, expressions:
    (LITERAL this: SELECT table_name, column_name, data_default FROM user_tab_columns, is_string: True))), by_ref: False))))
```

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -7,11 +7,6 @@ from sqlglot.dialects.dialect import Dialect, no_ilike_sql, rename_func, trim_sq
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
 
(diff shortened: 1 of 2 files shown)


tobymao (approved):

GeorgeSittas:
Seems like it's case insensitive, so I'm going to go ahead and merge this PR. Happy to revert/fix if it causes an issue.

Screenshot 2023-04-14 at 9 18 53 PM

PR #1428, opened 2023-04-14
duckdb fixes

ginter:
- duckdb [doesn't support column comments](https://github.com/duckdb/duckdb/discussions/6290)
- ts_or_ds_add wasn't recursively generating the this arg before interpolating

### DIFF
--- a/tests/dialects/test_duckdb.py
+++ b/tests/dialects/test_duckdb.py
@@ -149,6 +149,12 @@ class TestDuckDB(Validator):
             },
         )
         self.validate_all(
+            "CREATE TABLE IF NOT EXISTS table (cola INT COMMENT 'cola', colb STRING) USING ICEBERG PARTITIONED BY (colb)",
+            write={
+                "duckdb": "CREATE TABLE IF NOT EXISTS table (cola INT, colb TEXT)",
+            },
+        )
+        self.validate_all(
             "LIST_VALUE(0, 1, 2)",
             read={
                 "spark": "ARRAY(0, 1, 2)",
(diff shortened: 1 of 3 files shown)


tests/dialects/test_duckdb.py
```diff
@@ -339,6 +345,12 @@ def test_duckdb(self):
                 "bigquery": "ARRAY_CONCAT([1, 2], [3, 4])",
             },
         )
+        self.validate_all(
+            "SELECT CAST(CAST(x AS DATE) AS DATE) + INTERVAL 1 DAY",
```
ginter, line 349: the double cast is awkward but there's another test that expects the explicit cast. seems more hacky to try to inspect the rendered sql for 'this' to see if we need to cast again. open to other suggestions...

barakalon: checking if `isinstance(this, exp.Cast) and this.is_type(exp.DataType.Type.DATE)` doesn't seem _too_ bad to me.. but I can go either way. 

This is easier with type annotations, but alas, that only happens during optimization.

barakalon (approved):

tests/dialects/test_duckdb.py
ginter: it's not that straightforward tho bc we don't know if it's a Cast until we render it at which point we'd have to reparse

barakalon: (but that might be overkill)

tobymao: this is ok, this optimaziton should go in simplify

PR #1429, opened 2023-04-14
Fix: Expand as we traverse lineage

vchan:
This PR
- Recursively expands expressions and extract source aliases as we traverse lineage
- Removes cache of source tables because we were losing columns when a query references multiple columns from the same source table

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5064,7 +5064,7 @@ def expand(
             if source:
                 subquery = source.subquery(node.alias or name)
                 subquery.comments = [f"source: {name}"]
-                return subquery.transform(_expand, copy=copy)
+                return subquery.transform(_expand, copy=False)
         return node
 
     return expression.transform(_expand, copy=copy)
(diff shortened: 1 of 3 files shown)


sqlglot/lineage.py
```diff
@@ -110,18 +108,17 @@ def to_node(
         for c in set(select.find_all(exp.Column)):
             table = c.table
             source = scope.sources[table]
+            if source.expression and sources:
+                expression = exp.expand(source.expression, parsed_sources)
```
tobymao, line 112: why is this needed? why don't you do this in the pre?

vchan: Okay, I'll update expand.

sqlglot/expressions.py
```diff
@@ -5059,7 +5064,7 @@ def _expand(node: Expression):
             if source:
                 subquery = source.subquery(node.alias or name)
                 subquery.comments = [f"source: {name}"]
-                return subquery
+                return subquery.transform(_expand, copy=copy)
```
tobymao, line 5067: i think copy can always be False here because teh parent respects the copy

PR #1432, opened 2023-04-15
Fix(hive, spark): improve transpilation of TABLESAMPLE clause

### LINKED ISSUE #1425: TABLESAMPLE BERNOULLI doesn't translate from TrinoSQL to SparkSQL
`TABLESAMPLE BERNOULLI (50)` in TrinoSQL isn't translated cleanly to SparkSQL, as SparkSQL doesn't support `TABLESAMPLE BERNOULLI (50)`.  `TABLESAMPLE (50)` would be a fine translation -- it will use a random sampling, but the result will be similar.

GeorgeSittas:
Fixes #1425

Note: changed some tests because they were incorrect, i.e. `SELECT * FROM x TABLESAMPLE (10)` is incorrect for hive, which also doesn't support different sampling methods like `BERNOULLI` etc.

References:
- https://cwiki.apache.org/confluence/display/hive/languagemanual+sampling
- https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sampling.html
- https://prestodb.io/docs/current/sql/select.html#tablesample

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -177,6 +177,7 @@ class Hive(Dialect):
         IDENTIFIERS = ["`"]
         STRING_ESCAPES = ["\\"]
         ENCODE = "utf-8"
+        IDENTIFIER_CAN_START_WITH_DIGIT = True
 
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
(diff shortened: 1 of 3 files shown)


GeorgeSittas:
cc: @joshmarcus can you take a look and verify that these changes look ok?

tobymao (approved):

PR #1433, opened 2023-04-15
Fix: remove default kind from _parse_drop to make it more lenient

### LINKED ISSUE #1430: Oracle: ALTER TABLE x DROP PRIMARY KEY parse error
```
sqlglot.errors.ParseError: Expected table name but got <Token token_type: TokenType.PRIMARY_KEY, text: PRIMARY KEY, line: 3, col: 6, end: 37, comments: []>. Line 3, Col: 6.
  
alter table bicikli
drop primary key;
     ^^^^^^^^^^^
```

GeorgeSittas:
Fixes #1430


tests/test_transpile.py
```diff
@@ -322,7 +322,6 @@ def test_alter(self):
             "ALTER TABLE integers ADD k INTEGER",
             "ALTER TABLE integers ADD COLUMN k INT",
         )
-        self.validate("ALTER TABLE integers DROP k", "ALTER TABLE integers DROP COLUMN k")
```
GeorgeSittas, line 325: This test was removed because `DROP k` now transpiles back to itself, since it's parsed as a command.

tobymao (approved):

PR #1437, opened 2023-04-17
added new param

yiphei:


tobymao:
hello and welcome. can you explain what this is?

tobymao:
I appreciate the pr, but with no explanation I'm going to close this for now, happy to continue discussing. My guess is you're looking to be able to parse even without valid sql. In that case, just pass in error_level=ErrorLevel.IGNORE.

PR #1442, opened 2023-04-18
Feat(presto): convert week to 7 days in INTERVAL expressions

### LINKED ISSUE #1439: Trino/Presto doesn't have `week` interval unit
Trino/Presto doesn't have the `week` interval unit, but this is (not surprisingly!) not caught by SQLGlot. To support this, `interval '1' week` would need to be rewritten to `7 * interval '1' day` or something.

So ideally this would work
```python
assert "SELECT 7 * INTERVAL '1' day" == sqlglot.transpile("SELECT INTERVAL '1' week" read="postgres", write="trino")
```

GeorgeSittas:
Fixes #1439

cc: @vegarsti let me know if this solution works for you. Expressions like `(5 * INTERVAL '7' day)` seem to work for Trino.

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp, generator, parser, tokens, transforms
 from sqlglot.dialects.dialect import (
     Dialect,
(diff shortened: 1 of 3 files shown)


tobymao (approved):

vegarsti:
Amazing! Do you guys never rest? 😄

PR #1445, opened 2023-04-18
Feat(bigquery,presto): improve transpilation of INTERVAL (plural to singular)

### LINKED ISSUE #1440: Trino/Presto doesn't have plural interval units
It's not valid to say `interval '2' days` in Trino, it has to be `interval '1' day`. Postgres and others support both `day` and `days`, however, but this is not (yet) handled by SQLGlot.

So ideally this would work
```python
assert "SELECT INTERVAL '2' day" == sqlglot.transpile("SELECT INTERVAL '2' days", read="postgres", write="trino")
```

GeorgeSittas:
Fixes #1440

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -206,6 +206,10 @@ class BigQuery(Dialect):
         LOG_DEFAULTS_TO_LN = True
 
     class Generator(generator.Generator):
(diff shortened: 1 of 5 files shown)


tobymao (approved):

vegarsti:
Thanks a lot! 🎉

PR #1453, opened 2023-04-19
Teradata CAST DATE FORMAT to DATE_FORMAT()

treysp:
Transpile Teradata's `CAST('1/1999' as DATE FORMAT 'yyyy-mm')` to `DATE_FORMAT('1/1999', 'yyyy-mm')` for spark, databricks, and mysql

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -29,12 +29,6 @@ class Databricks(Spark):
 
         PARAMETER_TOKEN = "$"
 
-        def cast_sql(self, expression: exp.Cast) -> str:
-            if expression.to.is_type(exp.DataType.Type.DATE) and expression.to.args.get("format"):
-                return self.func("DATE_FORMAT", expression.this, expression.to.args.get("format"))
-
-            return super(Spark.Generator, self).cast_sql(expression)
-
     class Tokenizer(Spark.Tokenizer):
         SINGLE_TOKENS = {
             **Spark.Tokenizer.SINGLE_TOKENS,
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -29,6 +29,16 @@ class Databricks(Spark):
 
         PARAMETER_TOKEN = "$"
 
+        def cast_sql(self, expression: exp.Cast) -> str:
+            if (
+                isinstance(expression, exp.Cast)
+                and expression.to.is_type(exp.DataType.Type.DATE)
+                and expression.to.args.get("format")
+            ):
+                return self.func("DATE_FORMAT", expression.this, expression.to.args.get("format"))
+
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/databricks.py
```diff
@@ -29,6 +29,16 @@ class Generator(Spark.Generator):
 
         PARAMETER_TOKEN = "$"
 
+        def cast_sql(self, expression: exp.Cast) -> str:
```
tobymao, line 32: do only databricks and teradata have this syntax? should it be used more generally?

sqlglot/parser.py
```diff
@@ -3361,6 +3361,9 @@ def _parse_cast(self, strict: bool) -> exp.Expression:
         elif to.this == exp.DataType.Type.CHAR:
             if self._match(TokenType.CHARACTER_SET):
                 to = self.expression(exp.CharacterSet, this=self._parse_var_or_string())
+        elif to.this == exp.DataType.Type.DATE:
+            if self._match(TokenType.FORMAT):
+                to.set("format", self._parse_conjunction())
```
tobymao, line 3366: you need to add format to type args, you shouldn't just set things like this

sqlglot/dialects/teradata.py
```diff
@@ -161,3 +161,13 @@ def rangen_sql(self, expression: exp.RangeN) -> str:
             each_sql = f" EACH {each_sql}" if each_sql else ""
 
             return f"RANGE_N({this} BETWEEN {expressions_sql}{each_sql})"
+
+        def cast_sql(self, expression: exp.Cast) -> str:
```
tobymao, line 165: delete all of this it's not needed, you only need to modify the base def datatype_sql

tobymao: datatype_sql already has "nested" so add format there

sqlglot/dialects/databricks.py
tobymao: you don't need to check the type of expression, it will be cast

tobymao: the function? perhaps make it a class property or function that to import

there are examples of functions in dialects.py

sqlglot/parser.py
tobymao: yes that's correct

sqlglot/dialects/teradata.py
tobymao: the idea is for the generic dialect to be very powerful. how else would you represent it?

sqlglot/dialects/databricks.py
tobymao: can't you make as hared function?

tobymao:
i decided to rewrite this pr

https://github.com/tobymao/sqlglot/commit/35518cfd8b61d530aa654389d13d51d77537ebf3

PR #1454, opened 2023-04-20
Add Spark's timestamp function to Parser.FUNCTIONS

vegarsti:
From https://spark.apache.org/docs/latest/api/sql/index.html#timestamp:

> Casts the value `expr` to the target data type `timestamp`.

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -94,6 +94,10 @@ class Spark(Hive):
                 unit=exp.var(seq_get(args, 0)),
             ),
             "TRUNC": lambda args: exp.DateTrunc(unit=seq_get(args, 1), this=seq_get(args, 0)),
+            "TIMESTAMP": lambda args: exp.Cast(
+                this=seq_get(args, 0),
+                to="timestamp",
+            ),
         }
 
         FUNCTION_PARSERS = {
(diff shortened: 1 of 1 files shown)


sqlglot/dialects/spark.py
```diff
@@ -94,6 +94,10 @@ class Parser(Hive.Parser):
                 unit=exp.var(seq_get(args, 0)),
             ),
             "TRUNC": lambda args: exp.DateTrunc(unit=seq_get(args, 1), this=seq_get(args, 0)),
+            "TIMESTAMP": lambda args: exp.Cast(
+                this=seq_get(args, 0),
+                to="timestamp",
+            ),
```
GeorgeSittas, line 100:
```suggestion
            "TIMESTAMP": lambda args: exp.cast(seq_get(args, 0), "timestamp"),
```

GeorgeSittas (commented):
We can use `exp.cast` directly for this, see below. Let's also add a test.

sqlglot/dialects/spark.py
GeorgeSittas: Setting `to="timestamp"` is not correct because we need it to be a data type expression.

vegarsti:
> We can use `exp.cast` directly for this, see below. Let's also add a test.

Oh nice, thank you! Will add a test as well!

GeorgeSittas:
Getting this PR in, will also add a test. Thanks!

vegarsti:
Thank you!

PR #1456, opened 2023-04-20
Fixes for several transpilation issues for T-SQL

10bas10:
Fixed a couple of things:
- Added support for SYSTEM_USER function in T-SQL, supported with and without parenthesises
- Added support for JOIN HINTS coming from T-SQL
- Fixed the order of LATERAL JOINS by registering LATERAL JOINS under the "joins" arg instead of the "laterals" arg
- Fixed the issue where two LATERAL JOINS after eachother resulted in a parsing error by retreating if there was no match
- Added some testcases to check on the issues above


sqlglot/parser.py
```diff
@@ -108,6 +108,8 @@ class Parser(metaclass=_Parser):
         TokenType.CURRENT_USER: exp.CurrentUser,
     }
 
+    JOIN_HINTS = set()  # type: set[TokenType]
```
GeorgeSittas, line 111: Note: this should be `t.Set[str]` if we don't use tokens for the join hints.

```suggestion
    JOIN_HINTS: t.Set[TokenType] = set()
```

sqlglot/dialects/tsql.py
```diff
@@ -296,6 +297,14 @@ class Parser(parser.Parser):
...
             "SUSER_SNAME": exp.CurrentUser.from_arg_list,
+            "SYSTEM_USER": exp.CurrentUser.from_arg_list,
+        }
+
+        JOIN_HINTS = {
+            TokenType.LOOP,
+            TokenType.HASH,
+            TokenType.MERGE,
+            TokenType.REMOTE,
         }
```
GeorgeSittas, line 308: Maybe we should use a set of strings here instead? Nowadays we try to not create new tokens for things like `LOOP` and just use `self._match_texts`.

tests/dialects/test_tsql.py
```diff
@@ -631,3 +637,29 @@ def test_current_user(self):
...
+        self.validate_all(
+            "SYSTEM_USER()",
+            write={"spark": "CURRENT_USER()"},
+        )
+        self.validate_all(
+            "SYSTEM_USER",
+            write={"spark": "CURRENT_USER()"},
+        )
+
+    def test_join_hint(self):
```
GeorgeSittas, line 649: We should also test the generation of hints for T-SQL (i.e. they should be preserved).

GeorgeSittas (commented):
Left a few comments. Thanks for the PR @10bas10!

sqlglot/dialects/tsql.py
10bas10: Thanks, this is a bit more handy in this case indeed.

sqlglot/parser.py
10bas10: Will change that indeed.

tests/dialects/test_tsql.py
10bas10: I did not see the value in generating the hints, due to the fact you are usually transpiling towards a different dialect and those hints arent usable in those cases. Thats why I just fixed the parsing part. Is that ok with you?

GeorgeSittas: We should generate them because someone might want to use SQLGlot to format T-SQL, and so we can't break their queries by discarding those hints.

10bas10: Ah ok, will take a look at that one.

GeorgeSittas: Thanks, I think it should just involve handling the "hints" arg in `join_sql`.

sqlglot/parser.py
```diff
@@ -306,6 +308,7 @@ class Parser(metaclass=_Parser):
         TokenType.ROW,
         TokenType.UNNEST,
         TokenType.VAR,
+        TokenType.HASH,
```
GeorgeSittas, line 311: Why is this needed here?

10bas10: Ah, forgot to remove that when reverting the tokens.

tests/dialects/test_tsql.py
10bas10: Still, not in the generic class as they are only supported in T-SQL. You would only want to output when the read dialect is the same as the write dialect. Or do you have a good idea on this?

GeorgeSittas: One idea is to create a new class variable called something like `WITH_JOIN_HINT` that will control whether or not the `hint` arg will be used, and set its default value to be `False` (T-SQL will set that to be `True` -- see other recently added variables for reference).

sqlglot/parser.py
```diff
@@ -1867,7 +1870,10 @@ def _parse_query_modifiers(self, this: t.Optional[exp.Expression]) -> None:
             join = self._parse_join()
             comma = None if table else self._match(TokenType.COMMA)
             if lateral:
-                this.append("laterals", lateral)
+                if isinstance(lateral, exp.Join):
```
tobymao, line 1873: why was this changed, along with the optimizer?

10bas10: This was changed to keep the correct order of joins. When adding it to the laterals, the lateral joins are executed after the joins.

10bas10: As for the optimizer. The laterals coming from lateral joins werent traversed correctly.

tobymao: why is a lateral a join?

10bas10: Well when using a lateral join it is a join in my opinion. Lateral view however is not a join. 

Im using it in spark:
https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-join.html

GeorgeSittas (approved):
LGTM, only a bit skeptical about the lateral / join changes.

@10bas10 can you provide a couple of examples that demonstrate the need to make these  changes in the parser / scope modules?

sqlglot/generator.py
```diff
@@ -208,6 +208,8 @@ class Generator:
         exp.WithJournalTableProperty: exp.Properties.Location.POST_NAME,
     }
 
+    WITH_JOIN_HINT = False
```
tobymao, line 211: the default should be true, and others should say false, that way the generic generator is more inclusive

sqlglot/optimizer/scope.py
```diff
@@ -587,7 +587,11 @@ def _traverse_tables(scope):
     for join in scope.expression.args.get("joins") or []:
         expressions.append(join.this)
 
-    expressions.extend(scope.expression.args.get("laterals") or [])
+    for lateral in scope.expression.args.get("laterals") or []:
+        if isinstance(lateral, exp.Join):
```
tobymao, line 591: i'm not sure i still understand this.

if you have

select ... lateral join y

you add the join into expressions but not the lateral? 

but why?

sqlglot/parser.py
tobymao: so if this is a lateral join, shouldn't this be a lateral? how is a lateral join a join? if that's true maybe we should fix the join parser to parse laterals, instead of fixing it here

sqlglot/generator.py
GeorgeSittas: I think I originally suggested setting the default to `False` because only T-SQL has this concept AFAIK (basically to avoid repeating setting this flag to `False` for the other dialects).

sqlglot/parser.py
10bas10: Well as i see it, its a join on a lateral subquery or table valued function. For example im able to do LEFT JOIN LATERAL or a JOIN LATERAL in spark. So i would see it as a nested reference in those cases.

sqlglot/generator.py
10bas10: We have kind of an issue i think with these database engine specifics, because like with the already implemented table hints I see that the generated sql in case of transpiling T-SQL to Spark SQL gives the following input/output:

Input:
```
SELECT
  a.x,
  b.y
FROM a
INNER JOIN b WITH (SNAPSHOT)
  ON a.id = b.id
```

While `WITH (SNAPSHOT)` wont work in Spark SQL and is something T-SQL specific. Shouldnt we have something available which checks whether the read dialect is the same as the write dialect and generate these dialect specific syntax?

sqlglot/optimizer/scope.py
10bas10: As the lateral is nested within the join. As i mentioned earlier i would see this as a join on the lateral subquery or table valued function. Therefore the lateral is part of the "this" arg of the join, while the join is listed at the "joins".

sqlglot/generator.py
tobymao: spark should just not have the hint

10bas10: Indeed, thats the problem with this approach. Do you have an idea on how to make sure the hint isnt outputted for other dialects?

sqlglot/parser.py
```diff
@@ -2058,11 +2062,29 @@ def _parse_join_side_and_kind(
         )
 
     def _parse_join(self, skip_join_token: bool = False) -> t.Optional[exp.Expression]:
+        index = self._index
         natural, side, kind = self._parse_join_side_and_kind()
 
-        if not skip_join_token and not self._match(TokenType.JOIN):
+        hint = None
+        if self._match_texts(self.JOIN_HINTS):
```
tobymao, line 2069:
```suggestion
        hint = self._prev.text if self._match_texts(self.JOIN_HINTS) else None
```

sqlglot/generator.py
10bas10: Yeah, but what if another dialect also has table hints but in a whole different manner. In that situation you need to be able to identify whether the hint belongs to a specific dialect or we would need to map them somehow right?

sqlglot/parser.py
```diff
@@ -2058,11 +2062,29 @@ def _parse_join_side_and_kind(
...
+            hint = self._prev.text
+
+        join = self._match(TokenType.JOIN)
+        if not skip_join_token and not join:
+            self._retreat(index)
+            kind = None
+            natural = None
+            side = None
+
+        outer_apply = self._match_pair(TokenType.OUTER, TokenType.APPLY, False)
```
tobymao, line 2079: what's the point of this? you don't advance here so how does this work?

sqlglot/generator.py
10bas10: Ok, ill leave this for now as T-SQL is the first dialect with a join hint.

tobymao: i'll change it to true after you merge it in

10bas10: I dont understand. That would result in join hints from TSQL in for example Spark SQL?

tobymao: spark needs to say false then

10bas10: Then it would work for Spark, but not for postgresql eg.

tobymao: right, so we need to add false to every dialect that doesn't support it

10bas10: Ah right, im on track now :). Ill do so in a few.

10bas10: Ok i pushed the altered default value in the generator and created an override for the other dialects.

sqlglot/dialects/starrocks.py
```diff
@@ -21,6 +21,8 @@ class Parser(MySQL.Parser):  # type: ignore
         }
 
     class Generator(MySQL.Generator):  # type: ignore
+        WITH_JOIN_HINT = False
+
```
GeorgeSittas, line 25:
```suggestion
```

sqlglot/dialects/trino.py
```diff
@@ -6,6 +6,8 @@
 
 class Trino(Presto):
     class Generator(Presto.Generator):
+        WITH_JOIN_HINT = False
+
```
GeorgeSittas, line 10:
```suggestion
```

tobymao (approved):

PR #1460, opened 2023-04-21
Convert VOLATILE from CREATE arg to property

treysp:
Motivating problem: Teradata `CREATE VOLATILE TABLE` should transpile to `CREATE TABLE` for all other dialects (from #1447).

This PR converts VOLATILE from a CREATE arg to a property and uses the unsupported property machinery to suppress it for other dialects.


sqlglot/generator.py
```diff
@@ -200,10 +201,11 @@ class Generator:
         exp.SetProperty: exp.Properties.Location.POST_CREATE,
         exp.SortKeyProperty: exp.Properties.Location.POST_SCHEMA,
         exp.SqlSecurityProperty: exp.Properties.Location.POST_CREATE,
+        exp.StabilityProperty: exp.Properties.Location.POST_SCHEMA,
         exp.TableFormatProperty: exp.Properties.Location.POST_WITH,
         exp.TemporaryProperty: exp.Properties.Location.POST_CREATE,
         exp.TransientProperty: exp.Properties.Location.POST_CREATE,
-        exp.VolatilityProperty: exp.Properties.Location.POST_SCHEMA,
+        exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
```
tobymao, line 208: don't make general code unsupported

sqlglot/parser.py
```diff
@@ -1194,6 +1192,13 @@ def _parse_property_before(self) -> t.Optional[exp.Expression]:
         return None
 
     def _parse_property(self) -> t.Optional[exp.Expression]:
+        if self._prev.token_type in (
```
tobymao, line 1195: why is this needed?, why don't you treat it like a normal property

GeorgeSittas: Should we factor these tokens out into a class constant set?

tobymao: why don't you just use self._index ?

tobymao: can you describe to me exactly with sql examples why you need this?

tobymao: perfect

tobymao: can you make the change that you proposed and i can review that

tobymao (approved):

PR #1461, opened 2023-04-21
Fix: parse interval units as vars instead of identifiers

GeorgeSittas:


sqlglot/parser.py
```diff
@@ -2579,9 +2579,7 @@ def _parse_interval(self) -> t.Optional[exp.Expression]:
             return None
 
         this = self._parse_primary()
-        unit = self._parse_function() or self._parse_id_var(
-            any_token=False, tokens=self.INTERVAL_VARS
-        )
+        unit = self._parse_function() or self._parse_var()
```
tobymao, line 2582: will this also parse strings?

tobymao (approved):

sqlglot/parser.py
GeorgeSittas: What example do you have in mind? This handles things like `interval '1 day'` because the string's already parsed and stored in `this`.

PR #1464, opened 2023-04-21
[fix] edge cases for lineage node generation

Sharadh:
This PR fixes a couple of issues with `lineage.py:to_node`. Specifically, it handles two edge cases:
* If the lineage maps to a CTE with a `SELECT *`, the node is correctly generated
* If the lineage maps to an external source _not specified_ in the `sources` map, the node gracefully falls back to a placeholder value

While I was here, I also added some inline comments to help grok the code.

### Details
If the last node came from a CTE with a `SELECT *`, the function failed with
```
pytest -vv tests/test_lineage.py -k test_lineage_source_with_star
...
>       select = next(select for select in scope.selects if select.alias_or_name == column_name)
E       StopIteration

sqlglot/lineage.py:97: StopIteration
```
This is now handled correctly.

If the lineage involves a source that was not specified in the `sources` map, the function failed with
```
pytest -vv tests/test_lineage.py -k test_lineage_external_col
...
        for c in set(select.find_all(exp.Column)):
            table = c.table
>           source = scope.sources[table]
E           KeyError: ''

sqlglot/lineage.py:113: KeyError
```
This is now gracefully handled by falling back to the placeholder expression `?` as the source.

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -101,7 +101,7 @@ def lineage(
             if maybe_select.alias_or_name == column_name:
                 # If we get an exact match, stop searching.
                 select = maybe_select
-                break           
+                break
             if isinstance(maybe_select, exp.Star):
                 # If we get a `*`, make a note but keep moving. This means we will prefer
                 # an exact match if both a `*` and an exact match exist:
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -127,15 +127,21 @@ def lineage(
         # Find all columns that went into creating this one to list their lineage nodes.
         for c in set(select.find_all(exp.Column)):
             table = c.table
-            source = scope.sources[table]
+            source = scope.sources.get(table)
 
-            # If the table itself came from a scope, recurse into that one using the unaliased column name.
-            # Else, we've reached the end of the line - just create a node.
             if isinstance(source, Scope):
+                # If the table itself came from a scope, recurse into that one using the unaliased column name.
                 to_node(
                     c.name, scope=source, scope_name=table, upstream=node, alias=aliases.get(table)
                 )
+            elif source:
+                # Else, if we found a source that's not a scope, we've reached the end of the line.
+                node.downstream.append(Node(name=c.sql(), source=source, expression=source))
             else:
+                # If a source is not found, we can't go any further - this column's lineage is
+                # unknown. This can happen if the definition of a source used in a query is
+                # not passed into the `sources` map.
+                source = exp.Placeholder()
                 node.downstream.append(Node(name=c.sql(), source=source, expression=source))
 
         return node
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -94,6 +94,8 @@ def lineage(
                 )
             return node
 
+        # Find the specific select clause that is the source of the column we want.
+        # This can either be a specific, named select or a generic `*` clause.
         select = None
         for maybe_select in scope.selects:
             if maybe_select.alias_or_name == column_name:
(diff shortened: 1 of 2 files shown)


sqlglot/lineage.py
```diff
@@ -94,29 +94,51 @@ def to_node(
...
-        select = next(select for select in scope.selects if select.alias_or_name == column_name)
+        # Find the specific select clause that is the source of the column we want.
+        # This can either be a specific, named select or a generic `*` clause.
+        select = None
+        for maybe_select in scope.selects:
+            if maybe_select.alias_or_name == column_name:
+                # If we get an exact match, stop searching.
+                select = maybe_select
+                break
+            elif isinstance(maybe_select, exp.Star):
```
tobymao, line 105: the elif is unncessary, just else is fine

Sharadh: Hrmm, how so? Only a `*` or an exact match should be chosen - non matches should be ignored, no?

tobymao: what about this?

select = next(select for select in scope.selects if select.alias_or_name == column_name, exp.Star() if scope.expression.is_star else raise)

tobymao: @Sharadh since you do a break, you don't need an elif, just an if, but i think i have a more elegant solution

Sharadh: Oh yeah, you meant:
> the elif is unncessary, just ~else~ if is fine

That makes sense!

> ```
> select = next(select for select in scope.selects if select.alias_or_name == column_name, exp.Star() if scope.expression.is_star else raise)
> ```
This wouldn't work if the `exp.Star()` appears first in the selects list; we'd want to check if a more specific select matches if both a `*` and a more specific select are found.

tobymao: the star trigger only happens when we can't find the explicit match right?

tobymao:
great pr btw, thanks for doing this!

sqlglot/lineage.py
Sharadh: So, consider this (the second element of the generator):
```
exp.Star() if scope.expression.is_star
```
That's only triggered if the expression is `SELECT * FROM foo`. But what about something like `SELECT *, x FROM foo`?
1. If we're searching for "y", we want the Star condition to be triggered
2. If we're searching for "x", we _don't_ want the Star condition to be triggered (we want the more specific x)

I think both of these cases ^ will not work under the proposed change?

Sharadh: Ah, I wonder if the nuance I'm missing is: the optimizer would convert:
```
SELECT *, x FROM foo
```
to
```
SELECT y FROM foo
```
or
```
SELECT x FROM foo
```
for 1/ and 2/ respectively?

Sharadh:
@tobymao my pleasure - thanks for the project, enjoying it so far!

sqlglot/lineage.py
tobymao: doesn’t my solution provide that ? i’ll try it later

tobymao:
```python
        select = next(
            (select for select in scope.selects if select.alias_or_name == column_name),
            exp.Star() if scope.expression.is_star else None,
        )

        if not select:
            raise ValueError(f"Could not find {column_name} in {scope.expression}")
```
i just did this and the tests pass, so i'll commit it

tobymao:
cc233943

sqlglot/lineage.py
Sharadh: Ack - I just realized `is_star` is overridden for `Select`: https://github.com/tobymao/sqlglot/blob/8dc4d802502045c635af1019eb94d8197be49b5c/sqlglot/expressions.py#L2676-L2678

This is indeed more elegant - thanks for the refactor!

Issue #1467, opened 2023-04-22
comment without end makes parser raise IndexError


klyusba:
`select 1 /* ` will fail
It will be more convenient to highlight start of comment as syntax error, like PostgreSQL do:
` Unterminated block comment started at position 9 in SQL select 1 /*. Expected */ sequence`

PR #1468, opened 2023-04-22
Fix: skip the comment start when tokenizing a delimited comment

### LINKED ISSUE #1466: support division in a comment
`select 1 /*/2 */` will raise error, while it is valid query.
As far as I understend issue is that combination /*/ is treated like start and end of comment, while it is not

GeorgeSittas:
Fixes #1466

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -953,8 +953,10 @@ class Tokenizer(metaclass=_Tokenizer):
         comment_end = self._COMMENTS[comment_start]  # type: ignore
 
         if comment_end:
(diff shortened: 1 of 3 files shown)


tobymao (approved):

klyusba:
I wonder if one should skip comment_start_len in any case

klyusba:
Thanks for fix!

GeorgeSittas:
> I wonder if one should skip comment_start_len in any case

I suppose one issue that could arise is like the one you mentioned in #1467. I'll take a look soon, thanks for the reports!

PR #1471, opened 2023-04-24
Fix bug in optimizer.simplify

mbosc:
Similarly to what done in the other cases, EQ <-> LT/LTE/GT/GTE/NEQ simplification should not be triggered for ORs.

To reproduce:
``` 
sqlglot.optimizer.optimize(f'SELECT * FROM MetaDataTable T WHERE T."location" <> \'Rome\' AND (T."year" = 2018 OR T."year" <> 2018)')
# returns SELECT ... WHERE FALSE ```

### CODE REVIEW

**Comment by tobymao** [comment]
can you add a test case in simplify.sql?

**Comment by tobymao** [comment]
you should probably just move the or_ outside of the clause to simplify all of them at once

**Comment by mbosc** [comment]
done 👌🏻

**Review by tobymao — APPROVED** [review]

**Comment by tobymao** [comment]
thanks! looks like your test has failed

File: tests/fixtures/optimizer/simplify.sql [hunk 1]
```diff
@@ -572,3 +572,6 @@ x > 3;
 
 'a' < 'b';
 TRUE;
+
+x = 2018 OR x <> 2018;
+x = 2018 OR x <> 2018;
```
tobymao, line 577:
```suggestion
x <> 2018 OR x = 2018;
```

mbosc:
thx!

Issue #1455, opened 2023-04-20
Add support for Spark functions


vegarsti:
Spark has [a bunch of built-in functions](https://spark.apache.org/docs/latest/api/sql/index.html), such as `timestamp()` which is a shorthand for casting something to a timestamp. It'd be nice to have these in Spark's Parser, right now this is treated as an anonymous function so it won't get transpiled.

I'd be happy to add these, and I've taken the liberty to add the `timestamp` one in this PR: https://github.com/tobymao/sqlglot/pull/1454

GeorgeSittas:
Are there any functions that you currently need? I'm not sure if it's high priority to go through the whole list and add all missing expressions because there's a lot of them.

cc: @tobymao

tobymao:
closing as not planned, but we’ll happily accept your prs, thanks!

PR #1477, opened 2023-04-25
Add Spark's date function to Parser.FUNCTIONS

vegarsti:
Similar to https://github.com/tobymao/sqlglot/pull/1454, and encouraged by #1455.

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -89,7 +89,7 @@ class Spark(Hive):
             "WEEKOFYEAR": lambda args: exp.WeekOfYear(
                 this=exp.TsOrDsToDate(this=seq_get(args, 0)),
             ),
-            "DATE": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("date")),
+            "DATE": lambda args: exp.cast(seq_get(args, 0), "date"),
             "DATE_TRUNC": lambda args: exp.TimestampTrunc(
                 this=seq_get(args, 1),
                 unit=exp.var(seq_get(args, 0)),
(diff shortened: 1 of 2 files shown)


GeorgeSittas (approved):
Thanks for the PR! You can fix the style using `make style`, but as a rule of thumb just do `make check` before committing.

LGTM -- Are these `timestamp`, `date` functions only available in Spark SQL, or in Hive as well? We usually add functionality to Hive because Spark inherits it, but I'm not sure if we should in this case.

vegarsti:
Ah, thanks, will do that going forward!

Great question  -- will check!

vegarsti:
Doesn't look like these are in Hive, no: https://cwiki.apache.org/confluence/plugins/servlet/mobile?contentId=27362046#LanguageManualUDF-DateFunctions

I don't have a hive db i can test on, though, so don't know for sure

GeorgeSittas:
Ok no worries, we can be conservative and leave it as is then.

vegarsti:
Btw just so you know, not at the computer for a few hours, so  can't fix the syntax yet.

sqlglot/dialects/spark.py
```diff
@@ -89,6 +89,9 @@ class Parser(Hive.Parser):
             "WEEKOFYEAR": lambda args: exp.WeekOfYear(
                 this=exp.TsOrDsToDate(this=seq_get(args, 0)),
             ),
+            "DATE": lambda args: exp.Cast(
+                this=seq_get(args, 0), to=exp.DataType.build("date")
+            ),
```
GeorgeSittas, line 94:
```suggestion
            "DATE": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("date")),
```

GeorgeSittas:
> Btw just so you know, not at the computer for a few hours, so can't fix the style yet.

No problem, went ahead and fixed it for you.

vegarsti:
Thanks!!

PR #1481, opened 2023-04-25
Convert tokens to str

yiphei:


tobymao:
@yiphei are you doing this by accident?

PR #1486, opened 2023-04-26
Feat: infer PIVOT column names at parse time

GeorgeSittas:
This is the first step towards fixing the optimization of PIVOT operators. Since pivots create new (derived?) tables, we need to know the names of the corresponding columns in order to qualify them, or expand stars from them, for example. This PR introduces a way to infer these column names, because different engines have different naming schemes.

I think there are at least two ways to "optimize" pivots:

1. [Convert](https://stackoverflow.com/questions/5846007/sql-query-to-pivot-a-column-using-case-when) them into `CASE` expressions as a pre-canonicalization step.
2. Convert a query that selects from a pivoted table into an aliased subquery and use that as a source to select only the columns of interest. This will enable us to qualify the corresponding columns, since we'll have a source to refer to.

Both of these transformations require us to know all columns in the pivoted table for the general case. For example:

```sql
-- Assume that tbl contains columns col1, col2 and col3.
SELECT * FROM (SELECT * FROM tbl) PIVOT (MAX(col1) FOR col2 IN ('pval1', 'pval2'))
```
Here, the resulting table will contain `pval1`, `pval2` _and_ `col3`. The last one is included implicitly, because we `SELECT *` from `tbl` and so if we don't know its schema we can't optimize it correctly. If we _do_ know its schema, maybe we can convert this query into:

```sql
SELECT _piv.col3, _piv.pval1, _piv.pval2 FROM (
  SELECT * FROM (SELECT * FROM tbl) PIVOT (MAX(col1) FOR col2 IN ('pval1', 'pval2'))
) AS _piv(col3, pval1, pval2)
```

(*) There is a small issue because e.g. in BigQuery we can't define column names in table aliases like this, but we can probably find a workaround for that.

See also: https://github.com/tobymao/sqlglot/discussions/1449.


sqlglot/parser.py
```diff
@@ -747,6 +747,9 @@ class Parser(metaclass=_Parser):
 
     CONVERT_TYPE_FIRST = False
 
+    QUOTED_PIVOT_COLUMNS: t.Optional[bool] = None
```
tobymao, line 750: why is this optional?

sqlglot/dialects/spark.py
```diff
@@ -111,16 +113,22 @@ class Parser(Hive.Parser):
...
                 exp.Drop,
                 this=self._parse_schema(),
                 kind="COLUMNS",
             )
 
+        def _pivot_column_names(self, pivot_columns: t.List[exp.Expression]) -> t.List[str]:
+            # Spark doesn't add a suffix to the pivot columns when there's a single aggregation
+            if len(pivot_columns) == 1:
+                return [""]
+            return [agg.alias or agg.sql(dialect="spark").lower() for agg in pivot_columns]
```
tobymao, line 130: why is it lower?

sqlglot/parser.py
GeorgeSittas: It's because we use its value as an argument to `exp.to_identifier` and it has different semantics for `False` and `None`. We must quote if the identifier is not safe in dialects that create names such as `pivotcolumn_avg(col)`.

sqlglot/dialects/spark.py
GeorgeSittas: That's the name we get back when running spark. Observe:

```sql
spark-sql> WITH person(id, name, age, class, address) AS (SELECT * FROM VALUES (100, 'John', 30, 1, 'Street 1'), (200, 'Mary', NULL, 1, 'Street 2'), (300, 'Mike', 80, 3, 'Street 3'), (400, 'Dan', 50, 4, 'Street 4')) SELECT * FROM person PIVOT (SUM(age), AVG(class) AS c FOR name IN ('John', 'Mike' AS mike));
id	address	John_sum(age)	John_c	mike_sum(age)	mike_c
200	Street 2	NULL	NULL	NULL	NULL
100	Street 1	30	1.0	NULL	NULL
300	Street 3	NULL	NULL	80	3.0
400	Street 4	NULL	NULL	NULL	NULL
Time taken: 0.653 seconds, Fetched 4 row(s)
```

tobymao: what if it's quoted?

GeorgeSittas: Good point, it doesn't work well with quoted names that might contain uppercase letters. The issue is that function names are converted to lowercase in spark and so we need to mimic that somehow, but also respect the quoting / casing of quoted identifiers used as arguments to these aggregations.

So if we had `... PIVOT (SUM(`AgE`), .....)`,, we'd need to convert this to `John_sum(AgE)` (quoted).

GeorgeSittas: Do you think I can just do `.sql(..., normalize_func="lower")` for this?

tobymao: normalize_func is different than the identifiers right?

GeorgeSittas: It simply controls whether we're going to output ``` SUM(`FoO`) ``` as is or convert `SUM` to lowercase.

GeorgeSittas: Changed it

GeorgeSittas: Ah damn there's still an issue because it parses it in the following identifier (will be double-quoted when doing .sql)
```
(IDENTIFIER this: John_sum(`FoO`), quoted: True)
```

GeorgeSittas: Fyi spark does:
```sql
spark-sql> WITH person(id, name, `AgE`, class, address) AS (SELECT * FROM VALUES (100, 'John', 30, 1, 'Street 1'), (200, 'Mary', NULL, 1, 'Street 2'), (300, 'Mike', 80, 3, 'Street 3'), (400, 'Dan', 50, 4, 'Street 4')) SELECT * FROM person PIVOT (SUM(`AgE`), AVG(class) AS c FOR name IN ('John', 'Mike' AS mike));
id	address	John_sum(AgE)	John_c	mike_sum(AgE)	mike_c
200	Street 2	NULL	NULL	NULL	NULL
100	Street 1	30	1.0	NULL	NULL
300	Street 3	NULL	NULL	80	3.0
400	Street 4	NULL	NULL	NULL	NULL
Time taken: 1.145 seconds, Fetched 4 row(s)
```

sqlglot/dialects/spark.py
```diff
@@ -111,16 +113,43 @@ class Parser(Hive.Parser):
...
+                else:
+                    """
+                    This case corresponds to aggregations without aliases being used as suffixes
+                    (e.g. col_avg(foo)). We need to unquote identifiers because they're going to
+                    be quoted in the base parser's `_parse_pivot` method, due to `to_identifier`.
+                    Otherwise, we'd end up with `col_avg(`foo`)` (notice the double quotes).
+
+                    Moreover, function names are lowercased in order to mimic Spark's naming scheme.
+                    """
+                    agg_all_unquoted = agg.copy().transform(
```
tobymao, line 144: why are you doing a copy? transform already copies

GeorgeSittas: Missed it, thanks!

tobymao (approved):

PR #1501, opened 2023-04-28
Feat(presto): transpile explode/posexplode into (cross join) unnest

### LINKED ISSUE #1495: Transform exp.Explode and exp.Posexplode into exp.Unnest
In some languages (at least Spark SQL), `select explode(array_col) as a` will unravel the `array_col` such that each element is on one line. If this is selected along with other columns, this is semantically the same as a cross join.

Some languages (not sure which, but at least Presto and Trino) have support for this as `select a from unnest(array_col) as unnested_alias.b`. If there's a select from a table, it will be a cross join. 

Examples of what we want

```
select explode(array_col)
-> select col from unnest(array_col) as some_alias(col)

select explode(array_col) as alias
-> select alias from unnest(array_col) as some_alias(alias)

select posexplode(array_col)
-> select pos, col from unnest(array_col) with ordinality as some_alias(col, pos)

select a, posexplode(array_col) from table
-> select a, pos, col cross join unnest(array_col) with ordinality as some_alias(col, pos)
```

There is support for this going in the opposite direction, cross join unnest to lateral view explode. Used in presto -> hive:
https://github.com/tobymao/sqlglot/blob/e40cb6312f8492d51fed5df454efca39bd517e39/sqlglot/transforms.py#L140-L161 
(Frankly, I'm not sure of the difference in just using explode and using lateral view explode.)

With help from @GeorgeSittas I've [added support for this](https://github.com/duneanalytics/harmonizer/blob/e022b92fe0bd75ceacea25d6aa2124a66f3e6e85/dune/harmonizer/dialects/dunesql.py#L8-L68) in [duneanalytics/harmonizer](https://github.com/duneanalytics/harmonizer) by adding it as a preprocessing step. Test cases are [here](https://github.com/duneanalytics/harmonizer/blob/e022b92fe0bd75ceacea25d6aa2124a66f3e6e85/tests/test_dialect.py#L42-L116).

Would love to harden this and get it in upstream, so I'm planning to create a PR for it.

GeorgeSittas:
Fixes #1495

cc: @vegarsti I included the test cases you've added to the harmonizer repo, adapted appropriately.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4713,6 +4713,8 @@ def alias_(
 
     if table:
         table_alias = TableAlias(this=alias)
+
+        exp = exp.copy()
         exp.set("alias", table_alias)
 
         if not isinstance(table, bool):
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -197,11 +197,11 @@ def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
                 taken_source_names.add(unnest_source_alias)
 
                 if not explode_alias:
-                    explode_alias = find_new_name(taken_select_names, "_col")
+                    explode_alias = find_new_name(taken_select_names, "col")
                     taken_select_names.add(explode_alias)
 
                     if is_posexplode:
-                        pos_alias = find_new_name(taken_select_names, "_pos")
+                        pos_alias = find_new_name(taken_select_names, "pos")
                         taken_select_names.add(pos_alias)
 
                 if is_posexplode:
@@ -212,8 +212,7 @@ def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
                     column_names = [explode_alias]
                     to_replace.replace(exp.column(explode_alias))
 
-                # This mutates `unnest` by setting its table alias arg
-                exp.alias_(unnest, unnest_source_alias, table=column_names)
+                unnest = exp.alias_(unnest, unnest_source_alias, table=column_names)
 
                 if not expression.args.get("from"):
                     expression.from_(unnest, copy=False)
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -197,21 +197,22 @@ def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
                 taken_source_names.add(unnest_source_alias)
 
                 if not explode_alias:
-                    explode_alias = find_new_name(taken_select_names, "_c")
+                    explode_alias = find_new_name(taken_select_names, "_col")
                     taken_select_names.add(explode_alias)
 
                     if is_posexplode:
-                        pos_alias = find_new_name(taken_select_names, "_c")
+                        pos_alias = find_new_name(taken_select_names, "_pos")
                         taken_select_names.add(pos_alias)
 
                 if is_posexplode:
-                    column_names = [pos_alias, explode_alias]
+                    column_names = [explode_alias, pos_alias]
                     to_replace.pop()
-                    expression.select(exp.column(explode_alias), exp.column(pos_alias), copy=False)
+                    expression.select(pos_alias, explode_alias, copy=False)
                 else:
                     column_names = [explode_alias]
                     to_replace.replace(exp.column(explode_alias))
 
+                # This mutates `unnest` by setting its table alias arg
                 exp.alias_(unnest, unnest_source_alias, table=column_names)
 
                 if not expression.args.get("from"):
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -138,7 +138,7 @@ def remove_precision_parameterized_types(expression: exp.Expression) -> exp.Expr
 
 
 def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
-    """Convert cross join unnest into lateral view explode. Used in presto -> hive"""
+    """Convert cross join unnest into lateral view explode (used in presto -> hive)."""
     if isinstance(expression, exp.Select):
         for join in expression.args.get("joins") or []:
             unnest = join.this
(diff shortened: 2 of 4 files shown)


vegarsti:
Ah you're fast, I was hoping to get some open source points here 😄

sqlglot/transforms.py
```diff
@@ -161,6 +161,67 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+
+                unnest_source_alias = find_new_name(taken_source_names, "_u")
+                taken_source_names.add(unnest_source_alias)
+
+                if not explode_alias:
+                    explode_alias = find_new_name(taken_select_names, "_c")
+                    taken_select_names.add(explode_alias)
+
+                    if is_posexplode:
+                        pos_alias = find_new_name(taken_select_names, "_c")
```
vegarsti, line 204: Should this be more descriptive? It's a position column, so it being the same name as the unnested column (both are "_c") seems a bit weird to me. Could it be "_col" and "_pos", or "_c" and "_i" or something?

vegarsti (approved):
Really nice! Thanks for adding this. Honestly a bit bummed to not get this PR myself, but guess I was too slow. The code looks better than what I did in harmonize though!

sqlglot/transforms.py
vegarsti: Actually, since explode and posexplode default to pos and col in Spark, imo it would be better if we defaulted to that in the translated queries. (It would be better for our users) Otherwise their queries will break if they depend on the names of the columns (as many do). What do you think?

sqlglot/transforms.py
```diff
@@ -161,6 +161,67 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+                        taken_select_names.add(pos_alias)
+
+                if is_posexplode:
+                    column_names = [pos_alias, explode_alias]
+                    to_replace.pop()
+                    expression.select(exp.column(explode_alias), exp.column(pos_alias), copy=False)
```
tobymao, line 210: you don't need exp.column

sqlglot/transforms.py
```diff
@@ -161,6 +161,67 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+                else:
+                    column_names = [explode_alias]
+                    to_replace.replace(exp.column(explode_alias))
+
+                exp.alias_(unnest, unnest_source_alias, table=column_names)
```
tobymao, line 215: you aren't using this anywhere?

GeorgeSittas: This is mutating `unnest` because we're providing the `table` argument.

sqlglot/transforms.py
GeorgeSittas: > Should this be more descriptive? It's a position column, so it being the same name as the unnested column (both are "_c") seems a bit weird to me. Could it be "_col" and "_pos", or "_c" and "_i" or something?

This seems reasonable to me and it's not a big deal to apply this change. Thanks for the suggestion.

> Actually, since explode and posexplode default to pos and col in Spark, imo it would be better if we defaulted to that in the translated queries. (It would be better for our users) Otherwise their queries will break if they depend on the names of the columns (as many do). What do you think?

I'm not sure about this one. I was skeptical too, but Spark's approach is not ideal either because we may end up with duplicate column names and hence ambiguity if we just hardcode "col" and "pos". For example:

```
spark-sql> with t(pos) as (select 1) SELECT pos, posexplode(sequence(2, 3)) from t;
pos	pos	col
1	0	2
1	1	3
Time taken: 0.529 seconds, Fetched 2 row(s)
```

If we preserve spark's naming scheme, it will result in an invalid query for presto / trino:

```
trino> WITH t(pos) AS (SELECT 1) SELECT pos, pos, _c_2 FROM t CROSS JOIN UNNEST(SEQUENCE(2, 3)) WITH ORDINALITY AS _u(_c_2, pos);
Query 20230428_200142_00026_tqcwb failed: line 1:34: Column 'pos' is ambiguous
WITH t(pos) AS (SELECT 1) SELECT pos, pos, _c_2 FROM t CROSS JOIN UNNEST(SEQUENCE(2, 3)) WITH ORDINALITY AS _u(_c_2, pos)
```

Here, I just renamed `_c` to `pos` to demonstrate the point, but you get the idea. Maybe a better alternative would be to try and use the `pos`, `col` names, unless there's an ambiguity, in which case we'd provide a warning of some sort and continue with the auto-generated names. Then, we'd get a valid query but downstream references to `pos` might break.

sqlglot/transforms.py
GeorgeSittas: Left a comment to make this clearer

sqlglot/transforms.py
```diff
@@ -161,6 +161,68 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+                if is_posexplode:
+                    column_names = [explode_alias, pos_alias]
+                    to_replace.pop()
+                    expression.select(pos_alias, explode_alias, copy=False)
+                else:
+                    column_names = [explode_alias]
+                    to_replace.replace(exp.column(explode_alias))
+
+                # This mutates `unnest` by setting its table alias arg
+                exp.alias_(unnest, unnest_source_alias, table=column_names)
```
tobymao, line 216: this behavior is not good, and we should fix that, alias_ should not mutate it. can you fix alias_ so it doesn't mutate and also don't do that here?

tobymao: you can mutate directly here

sqlglot/transforms.py
vegarsti: Ah yes, I meant to default to col and pos but to add the postfix if those are taken.

vegarsti: Posexplode doesn't allow for an alis since it results in two columns, so in that case the user is at the mercy of those names

vegarsti: Sorry, it's actually possible by providing a tuple of names! Like we do here for unnest with ordinality, I suppose! So
`select posexplode(col) as (c, p)` works.

https://stackoverflow.com/a/54310860/3600147

sqlglot/transforms.py
```diff
@@ -161,6 +161,68 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+
+                # This ensures that we won't use [POS]EXPLODE's argument as a new selection
+                if isinstance(explode_arg, exp.Column):
+                    taken_select_names.add(explode_arg.output_name)
+
+                unnest_source_alias = find_new_name(taken_source_names, "_u")
+                taken_source_names.add(unnest_source_alias)
+
+                if not explode_alias:
+                    explode_alias = find_new_name(taken_select_names, "_col")
```
vegarsti, line 200:
```suggestion
                    explode_alias = find_new_name(taken_select_names, "col")
```

sqlglot/transforms.py
```diff
@@ -161,6 +161,68 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+                    taken_select_names.add(explode_alias)
+
+                    if is_posexplode:
+                        pos_alias = find_new_name(taken_select_names, "_pos")
```
vegarsti, line 204:
```suggestion
                        pos_alias = find_new_name(taken_select_names, "pos")
```

sqlglot/transforms.py
GeorgeSittas: Makes sense yeah I can make the change. I think we'll need to bump version after this because it's gonna be a breaking change.

sqlglot/transforms.py
GeorgeSittas: > Ah yes, I meant to default to col and pos but to add the postfix if those are taken. Added suggested commits for it below. This would be very helpful to us, so hoping you choose that over the ones with underscore!

Ah, great! Sure thing, will make the fix.

GeorgeSittas: Done.

vegarsti: Thank you!

sqlglot/expressions.py
```diff
@@ -4713,6 +4713,8 @@ def alias_(
 
     if table:
         table_alias = TableAlias(this=alias)
+
+        exp = exp.copy()
```
tobymao, line 4717: only copy if expression is an expression, if it's a string, you don't need this

tobymao (approved):

Issue #1500, opened 2023-04-28
Multiple Joins w/ Selects Still Fail


kgdiem:
Follow up to https://tobiko-data.slack.com/archives/C0448SFS3PF/p1682527687012379 / 
https://github.com/tobymao/sqlglot/blob/main/sqlglot/dataframe/README.md?plain=1#L228

I made a 3 examples here using SQLGlot 11.6.0: https://github.com/kgdiem/sqlglot-join-example/blob/main/mismatched_join_columns.py 

Even in the event of selecting every column on the joined tables, a join key always references the wrong table in the unoptimized select

Still have to look at how we're making a dataframe that generates an empty select statement in the unoptimized SQL ([ref](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1682628523070129?thread_ts=1682527687.012379&cid=C0448SFS3PF))

eakmanrq:
Thanks @kgdiem for your repros. Small typo in one of them: https://github.com/kgdiem/sqlglot-join-example/blob/main/mismatched_join_columns.py#L55

Should be `f.lit('X')`

This PR fixes the issues identified: https://github.com/tobymao/sqlglot/pull/1509

PR #1509, opened 2023-04-29
Fix Dataframe Triple Join issue

eakmanrq:
This also fixes an issue where the join columns were assumed to be expression in left -> right order. 

The DataFrame API will now return multiple columns with the same name if a select is not provided to make it clear which one you want. This matches Spark behavior (as proven by the tests). 

Resolves: https://github.com/tobymao/sqlglot/issues/1500

### DIFF
--- a/sqlglot/dataframe/sql/dataframe.py
+++ b/sqlglot/dataframe/sql/dataframe.py
@@ -16,7 +16,7 @@ from sqlglot.dataframe.sql.readwriter import DataFrameWriter
 from sqlglot.dataframe.sql.transforms import replace_id_value
 from sqlglot.dataframe.sql.util import get_tables_from_expression_with_join
 from sqlglot.dataframe.sql.window import Window
-from sqlglot.helper import ensure_list, object_to_dict
+from sqlglot.helper import ensure_list, object_to_dict, seq_get
 from sqlglot.optimizer import optimize as optimize_func
 
 if t.TYPE_CHECKING:
@@ -146,9 +146,9 @@ class DataFrame:
     def _ensure_list_of_columns(self, cols):
         return Column.ensure_cols(ensure_list(cols))
 
-    def _ensure_and_normalize_cols(self, cols):
+    def _ensure_and_normalize_cols(self, cols, expression: t.Optional[exp.Select] = None):
         cols = self._ensure_list_of_columns(cols)
-        normalize(self.spark, self.expression, cols)
+        normalize(self.spark, expression or self.expression, cols)
         return cols
 
     def _ensure_and_normalize_col(self, col):
@@ -365,6 +365,10 @@ class DataFrame:
                     x.this for x in get_tables_from_expression_with_join(self.expression)
                 ]
                 cte_names_in_join = [x.this for x in join_table_identifiers]
+                # If we a columns resolves to multiple CTE expressions then we want to use each CTE left-to-right
+                # and therefore we allow multiple columns with the same name in the result. This matches the behavior
+                # of Spark.
+                resolved_column_position: t.Dict[str, int] = {}
                 for ambiguous_col in ambiguous_cols:
                     ctes_with_column = [
                         cte
@@ -372,13 +376,16 @@ class DataFrame:
                         if cte.alias_or_name in cte_names_in_join
                         and ambiguous_col.alias_or_name in cte.this.named_selects
                     ]
-                    # If the select column does not specify a table and there is a join
-                    # then we assume they are referring to the left table
-                    if len(ctes_with_column) > 1:
-                        table_identifier = self.expression.args["from"].args["expressions"][0].this
+                    if ambiguous_col in resolved_column_position:
+                        cte = seq_get(ctes_with_column, resolved_column_position[ambiguous_col] + 1)
+                        if cte:
+                            resolved_column_position[ambiguous_col] += 1
+                        else:
+                            cte = ctes_with_column[resolved_column_position[ambiguous_col]]
                     else:
-                        table_identifier = ctes_with_column[0].args["alias"].this
-                    ambiguous_col.expression.set("table", table_identifier)
+                        cte = ctes_with_column[0]
+                        resolved_column_position[ambiguous_col] = 0
+                    ambiguous_col.expression.set("table", cte.alias_or_name)
         return self.copy(
             expression=self.expression.select(*[x.expression for x in cols], **kwargs), **kwargs
         )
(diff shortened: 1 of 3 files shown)


eakmanrq:
This fixed all my repros but did not fix the ones provided. So still looking at this.

eakmanrq:
There was just a typo in the provided test case. All those tests pass too.

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -365,20 +365,27 @@ def select(self, *cols, **kwargs) -> DataFrame:
                     x.this for x in get_tables_from_expression_with_join(self.expression)
                 ]
                 cte_names_in_join = [x.this for x in join_table_identifiers]
+                # If we a columns resolves to multiple CTE expressions then we want to use each CTE left-to-right
```
tobymao, line 368: if we have

tobymao: or if a

tobymao: why do you want to use ctes left to right?

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -365,20 +365,27 @@ def select(self, *cols, **kwargs) -> DataFrame:
...
+                # and therefore we allow multiple columns with the same name in the result. This matches the behavior
+                # of Spark.
+                resolved_column_position: t.Dict[str, int] = {}
                 for ambiguous_col in ambiguous_cols:
                     ctes_with_column = [
                         cte
                         for cte in self.expression.ctes
                         if cte.alias_or_name in cte_names_in_join
```
tobymao, line 376: you should do this outside of the loop, and probably make a set

for cte in self.expression.ctes if cte.alias_or_name in cte_names in join

tobymao: in general this code isn't really correct, it should really be using scopes but i guess we can leave that for another day

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -365,20 +365,27 @@ def select(self, *cols, **kwargs) -> DataFrame:
...
                         and ambiguous_col.alias_or_name in cte.this.named_selects
                     ]
-                    # If the select column does not specify a table and there is a join
-                    # then we assume they are referring to the left table
-                    if len(ctes_with_column) > 1:
-                        table_identifier = self.expression.args["from"].args["expressions"][0].this
+                    if ambiguous_col in resolved_column_position:
+                        cte = seq_get(ctes_with_column, resolved_column_position[ambiguous_col] + 1)
```
tobymao, line 380: i don't understand this code, can you explain it?

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -420,59 +427,84 @@ def join(
...
         other_df = other_df._convert_leaf_to_cte()
-        pre_join_self_latest_cte_name = self.latest_cte_name
-        columns = self._ensure_and_normalize_cols(on)
-        join_type = how.replace("_", " ")
-        if isinstance(columns[0].expression, exp.Column):
-            join_columns = [
-                Column(x).set_table_name(pre_join_self_latest_cte_name) for x in columns
+        join_columns = self._ensure_list_of_columns(on)
+        # We will determine actual join expression later so we provide a dummy one for now.
+        join_expression = self.expression.copy().join(
```
tobymao, line 432: why are you doing a copy? join already does a copy

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -420,59 +427,84 @@ def join(
...
+            other_df.latest_cte_name, on=exp.Column(), join_type=how.replace("_", " ")
```
tobymao, line 433: why are doing a null column? that's invalid

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -420,59 +427,84 @@ def join(
...
+            """
+            Unique characteristics of join on column names only:
+            * The column names are put at the front of the select list
+            * The column names are deduplicated across the entire select list and only the column names (other dups are allowed)
+            """
+            tables = get_tables_from_expression_with_join(join_expression)
+            potential_ctes = [
+                cte
+                for cte in join_expression.ctes
+                if cte.alias_or_name in [x.alias_or_name for x in tables]
```
tobymao, line 450: move this gen out of the tight loop

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -420,59 +427,84 @@ def join(
...
+                    if join_column.alias_or_name in cte.this.named_selects:
+                        left_column = join_column.copy().set_table_name(cte.alias_or_name)
+                        right_column = join_column.copy().set_table_name(other_df.latest_cte_name)
+                        join_column_pairs.append((left_column, right_column))
+                        num_matching_ctes += 1
+                if num_matching_ctes > 1:
+                    raise ValueError(
+                        f"Column {join_column.alias_or_name} is ambiguous. Please specify the table name."
+                    )
+                elif num_matching_ctes == 0:
```
tobymao, line 468: why not just else?

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -420,59 +427,84 @@ def join(
...
-                other_df.latest_cte_name, on=join_clause.expression, join_type=join_type
-            )
-        )
-        new_df.expression = new_df._add_ctes_to_expression(
-            new_df.expression, other_df.expression.ctes
-        )
+            select_column_names = [
+                column_name
+                for column_name in select_column_names
+                if column_name not in [x.alias_or_name for x in join_columns]
```
tobymao, line 487: tight loop

sqlglot/dataframe/sql/dataframe.py
eakmanrq: To match Spark behavior

sqlglot/dataframe/sql/dataframe.py
eakmanrq: I explain that in the comment above. This on expression can be anything since we calculate it on later steps

sqlglot/dataframe/sql/dataframe.py
eakmanrq: If it was else then that would include the case where `num_matching_ctes` is 1 and therefore it would raise an error. This way it only captures if it is 0.

sqlglot/dataframe/sql/dataframe.py
eakmanrq: What do you mean by incorrect? This code was written prior to scopes so that is why it isn't currently used but I believe this functions correctly. There are likely a few places that would benefit from scopes now that they exist.

sqlglot/dataframe/sql/dataframe.py
eakmanrq: Simplified a bit and left a comment

sqlglot/dataframe/sql/dataframe.py
tobymao: why don't you just leave it null?

eakmanrq: Ok that works.

PR #1522, opened 2023-05-02
Feat(oracle): support KEEP (.. [FIRST|LAST] ..) window function syntax

### LINKED ISSUE #1520: KEEP in an Oracle SQL Statement is retrieving a ParseError
If you try to parse the statement below it returns an Invalid expression / unexpected token error
`
sql_query = """ SELECT  MIN(column_name) KEEP (DENSE_RANK FIRST ORDER BY column_name DESC)  FROM table_name;
"""
parsed = parse_one(sql_query , read='oracle')
`

**Retrieved Error:**
image

In general, the KEEP clause is used in combination with the DENSE_RANK, RANK, or ROW_NUMBER functions to identify the rows that satisfy the criteria for the aggregate function, and then the KEEP clause applies the aggregate function only to those rows. More info here: [Oracle Docs](https://docs.oracle.com/database/121/SQLRF/functions003.htm#SQLRF20035)

GeorgeSittas:
Fixes #1520


tests/fixtures/identity.sql
```diff
@@ -818,3 +818,4 @@ JSON_OBJECT('x': NULL, 'y': 1 ABSENT ON NULL WITH UNIQUE KEYS)
 JSON_OBJECT('x': 1 RETURNING VARCHAR(100))
 JSON_OBJECT('x': 1 RETURNING VARBINARY FORMAT JSON ENCODING UTF8)
 SELECT if.x
+SELECT MIN(column_name) KEEP (DENSE_RANK FIRST ORDER BY column_name DESC) FROM table_name
```
tobymao, line 821: add a keep test?

tobymao, line 821: select y keep

does this break?

GeorgeSittas: Doesn't break, but I added a test for `sum(y) keep` which did break before. I isolated `KEEP` for oracle only.

sqlglot/dialects/oracle.py
```diff
@@ -166,6 +168,7 @@ class Tokenizer(tokens.Tokenizer):
             **tokens.Tokenizer.KEYWORDS,
             "(+)": TokenType.JOIN_MARKER,
             "COLUMNS": TokenType.COLUMN,
+            "KEEP": TokenType.KEEP,
```
tobymao, line 171: you only need one or the other, since you specify window_before_paren_token, then everyone can parse keep as a token and then you can put this into id_var, or you can get rid of line 64,

GeorgeSittas: I think I made the change you suggested, but let me know if I missed something.

PR #1524, opened 2023-05-03
Feat: Use alternative transform for dialects that do not support distinct on

crericha:


sqlglot/dialects/sqlite.py
```diff
@@ -65,6 +65,7 @@ class Generator(generator.Generator):
 
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,  # type: ignore
+            **transforms.ELIMINATE_DISTINCT_ON,  # type: ignore
```
tobymao, line 68: this is not correct, you've munged eliminate_qualify since they're both keyed on select, you need to do custom here

exp.Select: preprocess([eliminate_distinct_on, eliminate_qualify])

can you delete all transform constants in transform.py and replace them with this more explicit version?

sqlglot/dialects/hive.py
```diff
@@ -280,6 +280,7 @@ class Generator(generator.Generator):
 
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,  # type: ignore
+            **transforms.ELIMINATE_DISTINCT_ON,  # type: ignore
```
tobymao, line 283: broken here too

PR #1525, opened 2023-05-03
Fix: use preprocess instead of expanding transform dicts BREAKING

GeorgeSittas:
Cleaned up https://github.com/tobymao/sqlglot/pull/1524.

@tobymao I think we need to stop using the `**transforms.SOME_TRANSFORM` pattern, because it can easily lead to bugs in the future. For example, prior to Chris' PR we had the following entry in the `TRANSFORMS` map of bigquery:

```python
exp.Select: transforms.preprocess([_unqualify_unnest])
```

This would cause the `**transforms.ELIMINATE_DISTINCT_ON` to be overridden and it's tricky to spot right away.

cc: @crericha


tests/dialects/test_redshift.py
```diff
@@ -101,7 +101,22 @@ def test_redshift(self):
         self.validate_all(
             "SELECT DISTINCT ON (a) a, b FROM x ORDER BY c DESC",
```
GeorgeSittas, line 102: Note: this is parsed with Redshift as the input dialect, which displays the `NULL`s first with the `DESC` order.

Screenshot 2023-05-03 at 4 19 03 PM

tobymao:
can you delete the constants in transform.py?

PR #1526, opened 2023-05-03
Fix: make some SQL builders pure

GeorgeSittas:


sqlglot/expressions.py
```diff
@@ -3538,14 +3540,16 @@ class Case(Func):
     arg_types = {"this": False, "ifs": True, "default": False}
 
     def when(self, condition: ExpOrStr, then: ExpOrStr, copy: bool = True, **opts) -> Case:
-        this = self.copy() if copy else self
-        this.append("ifs", If(this=maybe_parse(condition, **opts), true=maybe_parse(then, **opts)))
-        return this
+        instance = _maybe_copy(self, copy)
+        instance.append(
+            "ifs", If(this=maybe_parse(condition, **opts), true=maybe_parse(then, **opts))
```
GeorgeSittas, line 3545: Here we don't copy `condition` and `then` because `copy` is not included in `opts`. Should we default to copying these expressions, i.e. pass in `**{**opts, copy: True}` to `maybe_parse`?

sqlglot/expressions.py
```diff
@@ -809,10 +811,10 @@ def __ror__(self, other: ExpOrStr) -> Or:
         return self._binop(Or, other, reverse=True)
 
     def __neg__(self) -> Neg:
-        return Neg(this=_wrap(self, Binary))
+        return Neg(this=_wrap(self.copy(), Binary))
```
GeorgeSittas, line 814: Can also make it so that `_wrap` copies `expression` by default whenever it creates a `Paren` that wraps it.

sqlglot/expressions.py
```diff
@@ -714,16 +714,18 @@ def _binop(self, klass: t.Type[E], other: ExpOrStr, reverse=False) -> E:
...
+            this=self.copy(), expressions=[convert(e, copy=True) for e in ensure_list(other)]
+        )
 
     def isin(self, *expressions: ExpOrStr, query: t.Optional[ExpOrStr] = None, **opts) -> In:
         return In(
-            this=self,
-            expressions=[convert(e) for e in expressions],
+            this=self.copy(),
+            expressions=[convert(e, copy=True) for e in expressions],
             query=maybe_parse(query, **opts) if query else None,
```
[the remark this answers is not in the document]
tobymao: add copy flag to the method definition and pass it in to everything

sqlglot/expressions.py
tobymao: why don't you just pass copy in?

tobymao:
can you double check and_, not_, or_ and, and paren?

GeorgeSittas:
Yeah will take a look, thanks for reviewing

sqlglot/expressions.py
GeorgeSittas: Done.

sqlglot/expressions.py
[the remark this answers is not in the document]
GeorgeSittas: Done.

GeorgeSittas:
> can you double check and_, not_, or_ and, and paren?

They don't copy by default:

```python
>>> import sqlglot
>>> c = sqlglot.exp.Condition(this="x")
>>> a = sqlglot.exp.and_(c, "true")
>>> id(a.this)
4368906528
>>> id(c)
4368906528
```

tobymao:
@GeorgeSittas yes, and i'd just double check where it's used and see if we need to put false

GeorgeSittas:
Got it, will look into this soon

PR #1528, opened 2023-05-03
Support regex function in Starrocks

acreux:
Support regex in Starrocks

Function is documented at https://docs.starrocks.io/en-us/2.4/sql-reference/sql-functions/like_predicate-functions/regexp

MySQL uses `REGEXP_LIKE`
https://dev.mysql.com/doc/refman/8.0/en/regexp.html


tobymao:
is this different from mysql?

GeorgeSittas (approved):
I checked, `REGEXP` isn't valid MySQL so LGTM.

acreux:
indeed, MySQL uses `REGEXP_LIKE`

PR #1494, opened 2023-04-27
Add Spark's int cast function

vegarsti:
I meant to add this to #1485 yesterday, but ran into some noob git issue where I didn't manage to push to the remote before it was merged https://github.com/tobymao/sqlglot/pull/1485#issuecomment-1523444653

PR #1530, opened 2023-05-03
Fix(Spark): Add DOUBLE, FLOAT cast functions

vegarsti:
Sorry, thought I'd added all these! Same as https://github.com/tobymao/sqlglot/pull/1494 etc.

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -100,6 +100,8 @@ class Spark(Hive):
                 this=seq_get(args, 1),
                 unit=exp.var(seq_get(args, 0)),
             ),
+            "FLOAT": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("float")),
+            "DOUBLE": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("double")),
             "STRING": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("string")),
             "TRUNC": lambda args: exp.DateTrunc(unit=seq_get(args, 1), this=seq_get(args, 0)),
             "TIMESTAMP": lambda args: exp.Cast(
--- a/tests/dialects/test_spark.py
+++ b/tests/dialects/test_spark.py
@@ -225,34 +225,15 @@ TBLPROPERTIES (
                 "trino": "SELECT DATE_DIFF('MONTH', CAST(SUBSTR(CAST('2020-01-01' AS VARCHAR), 1, 10) AS DATE), CAST(SUBSTR(CAST('2020-03-05' AS VARCHAR), 1, 10) AS DATE))",
             },
         )
-        self.validate_all(
-            "BOOLEAN(x)",
-            write={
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/spark.py
```diff
@@ -100,6 +100,8 @@ class Parser(Hive.Parser):
                 this=seq_get(args, 1),
                 unit=exp.var(seq_get(args, 0)),
             ),
+            "FLOAT": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("float")),
+            "DOUBLE": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("double")),
             "STRING": lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build("string")),
```
GeorgeSittas, line 105: Should we dry these out? I'm thinking something like:

```python
...
def _parse_as_cast(to_type: str) -> exp.Expression:
    return lambda args: exp.Cast(this=seq_get(args, 0), to=exp.DataType.build(to_type))

...
            "FLOAT": _parse_as_cast("float"),
...
```

tobymao:
@vegarsti this needs to be redone / rebased

sqlglot/dialects/spark.py
vegarsti: Good idea! will just lift that suggestion, thanks!

tests/dialects/test_spark.py
```diff
@@ -232,6 +232,20 @@ def test_spark(self):
                 "spark": "CAST(x AS BOOLEAN)",
             },
         )
+        self.validate_all(
+            "FLOAT(x)",
+            write={
+                "": "CAST(x AS FLOAT)",
+                "spark": "CAST(x AS FLOAT)",
+            },
+        )
```
vegarsti, line 241: @GeorgeSittas  Should I add a dry up the tests too or is this fine for now?

tobymao: yea let's dry up the tests

vegarsti: Cool! Will do in the morning

vegarsti: Does this work? cfa50ae7fbbb3e21ee256de40eda45474f5862c8

https://github.com/tobymao/sqlglot/blob/cfa50ae7fbbb3e21ee256de40eda45474f5862c8/tests/dialects/test_spark.py#L229-L243

Opted for keeping the STRING case out of the loop since I felt it made the loop less clear, but could also do this or something 🤷🏻‍♂️ 

```python
        for data_type in ("BOOLEAN", "DATE", "DOUBLE", "FLOAT", "INT", "STRING", "TIMESTAMP"):
            self.validate_all(
                f"{data_type}(x)",
                write={
                    "": f"CAST(x AS {data_type if data_type != "STRING" else "TEXT"})",
                    "spark": f"CAST(x AS {data_type})",
                },
            )
```

GeorgeSittas (approved):
LGTM, thanks!

PR #1536, opened 2023-05-03
Add BYTES type to BigQuery dialect

relud:
fixes https://github.com/tobymao/sqlglot/issues/1533


sqlglot/dialects/bigquery.py
```diff
@@ -139,6 +139,7 @@ class Tokenizer(tokens.Tokenizer):
             "GEOGRAPHY": TokenType.GEOGRAPHY,
             "FLOAT64": TokenType.DOUBLE,
             "INT64": TokenType.BIGINT,
+            "BYTES": TokenType.BINARY,
```
GeorgeSittas, line 142: You should also add the entry `TokenType.BINARY: "BYTES"` to `TYPE_MAPPING`.

GeorgeSittas (approved):
Thanks for the PR!

PR #1548, opened 2023-05-04
Fix(oracle): set post_tablesample_alias=True to fix alias parsing

### LINKED ISSUE #1541: Alias to SAMPLE(CLAUSE) throw ParseError for Oracle dialect
If you try to parse the statement below it returns an Invalid expression / unexpected token error
`sql_query = """  SELECT * FROM table_name tn SAMPLE(25) s"""  
parsed = parse_one(sql_query , read='oracle')`

**Retrieve error**
image

Even if there are not many use-case (reasons) for giving an alias name to SAMPLE, it could be helpful in some situation and furthermore it does not raise an error if you try to compile via Oracle

GeorgeSittas:
Fixes #1541

### DIFF
--- a/tests/dialects/test_snowflake.py
+++ b/tests/dialects/test_snowflake.py
@@ -426,29 +426,18 @@ class TestSnowflake(Validator):
     def test_sample(self):
         self.validate_identity("SELECT * FROM testtable TABLESAMPLE BERNOULLI (20.3)")
         self.validate_identity("SELECT * FROM testtable TABLESAMPLE (100)")
+        self.validate_identity("SELECT * FROM testtable TABLESAMPLE SYSTEM (3) SEED (82)")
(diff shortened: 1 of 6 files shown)


tests/dialects/test_snowflake.py
```diff
@@ -426,29 +426,18 @@ def test_null_treatment(self):
     def test_sample(self):
```
GeorgeSittas, line 426: We used to convert `SAMPLE` to `TABLESAMPLE`, [which are both synonymous in Snowflake](https://docs.snowflake.com/en/sql-reference/constructs/sample):

> SAMPLE and TABLESAMPLE are synonymous and can be used interchangeably.

With the change in the parser related to `kind` (see `_parse_table_sample`), we now actually use whatever the token's text was, so I adapted these tests accordingly.

PR #1552, opened 2023-05-04
Feat: refactor hex/bit literals so that their text is preserved

GeorgeSittas:
xref https://github.com/duneanalytics/harmonizer/issues/41

cc: @vegarsti

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -77,7 +77,7 @@ class _Dialect(type):
             bs_start, bs_end = list(klass.tokenizer_class._BIT_STRINGS.items())[0]
             klass.generator_class.TRANSFORMS[
                 exp.BitString
-            ] = lambda self, e: f"{bs_start}{int(self.sql(e, 'this')):b}{bs_end}"
+            ] = lambda self, e: f"{bs_start}{self.sql(e, 'this')}{bs_end}"
         if (
             klass.tokenizer_class._HEX_STRINGS
             and exp.HexString not in klass.generator_class.TRANSFORMS
@@ -85,7 +85,7 @@ class _Dialect(type):
             hs_start, hs_end = list(klass.tokenizer_class._HEX_STRINGS.items())[0]
             klass.generator_class.TRANSFORMS[
                 exp.HexString
-            ] = lambda self, e: f"{hs_start}{int(self.sql(e, 'this')):X}{hs_end}"
+            ] = lambda self, e: f"{hs_start}{self.sql(e, 'this')}{hs_end}"
         if (
             klass.tokenizer_class._BYTE_STRINGS
             and exp.ByteString not in klass.generator_class.TRANSFORMS
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -1033,7 +1033,9 @@ class Tokenizer(metaclass=_Tokenizer):
         self._advance()
         value = self._extract_value()
         try:
-            self._add(TokenType.BIT_STRING, f"{int(value, 2)}")
+            # If `value` can't be converted to a binary, fallback to tokenizing it as an identifier
+            int(value, 2)
+            self._add(TokenType.BIT_STRING, value[2:])  # Drop the 0b
         except ValueError:
             self._add(TokenType.IDENTIFIER)
 
@@ -1041,7 +1043,9 @@ class Tokenizer(metaclass=_Tokenizer):
         self._advance()
         value = self._extract_value()
         try:
-            self._add(TokenType.HEX_STRING, f"{int(value, 16)}")
+            # If `value` can't be converted to a hex, fallback to tokenizing it as an identifier
+            int(value, 16)
+            self._add(TokenType.HEX_STRING, value[2:])  # Drop the 0x
         except ValueError:
             self._add(TokenType.IDENTIFIER)
 
@@ -1066,7 +1070,7 @@ class Tokenizer(metaclass=_Tokenizer):
         self._add(TokenType.NATIONAL if quote[0].upper() == "N" else TokenType.STRING, text)
         return True
 
-    # X'1234, b'0110', E'\\\\\' etc.
+    # X'1234', b'0110', E'\\\\\' etc.
     def _scan_formatted_string(self, string_start: str) -> bool:
         if string_start in self._HEX_STRINGS:
             delimiters = self._HEX_STRINGS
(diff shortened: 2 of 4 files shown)


sqlglot/tokens.py
```diff
@@ -1033,15 +1033,19 @@ def _scan_bits(self) -> None:
         self._advance()
         value = self._extract_value()
         try:
-            self._add(TokenType.BIT_STRING, f"{int(value, 2)}")
+            # If `value` can't be converted to a binary, fallback to tokenizing it as an identifier
+            int(value, 2)
+            self._add(TokenType.BIT_STRING, value[2:])  # Drop the 0b
```
tobymao, line 1038: how do you know the length of 0b?

GeorgeSittas: The code path that leads to this function has already checked for the `0b` prefix (same for hex).

tests/dialects/test_mysql.py
```diff
@@ -161,6 +161,17 @@ def test_hexadecimal_literal(self):
...
         )
+        self.validate_all(
+            "SELECT 0x0000CC",
+            write={
+                "mysql": "SELECT x'0000CC'",
+                "sqlite": "SELECT x'0000CC'",
+                "spark": "SELECT X'0000CC'",
+                "trino": "SELECT X'0000CC'",
+                "bigquery": "SELECT 0x0000CC",
+                "oracle": "SELECT 204",
```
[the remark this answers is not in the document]
GeorgeSittas: Good point, will do.

sqlglot/tokens.py
tobymao: k sounds good

sqlglot/tokens.py
```diff
@@ -1033,15 +1033,19 @@ def _scan_bits(self) -> None:
...
             self._add(TokenType.IDENTIFIER)
 
     def _scan_hex(self) -> None:
         self._advance()
         value = self._extract_value()
         try:
-            self._add(TokenType.HEX_STRING, f"{int(value, 16)}")
+            # If `value` can't be converted to a hex, fallback to tokenizing it as an identifier
+            int(value, 16)
+            self._add(TokenType.HEX_STRING, value[2:])  # Drop the 0x
```
vegarsti, line 1048: So this is the crucial bit, right?

GeorgeSittas: There were a couple other parts that needed to be changed as well. Minor changes though, yeah.

crericha, line 1048: Is removing the first two characters valid for all cases? Aren't we getting lucky that the two current formats start with `0x` and `X'`?

GeorgeSittas: Yep, discussed in https://github.com/tobymao/sqlglot/pull/1552#discussion_r1185453439

GeorgeSittas: Btw `X'` is handled elsewhere, see `_scan_formatted_string`.

tests/dialects/test_mysql.py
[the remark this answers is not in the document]
GeorgeSittas: Will finish this one tomorrow, have to do some cleanups & investigate what other dialects too in order to improve our test coverage on this.

sqlglot/tokens.py
vegarsti: but in theory a dialect could have an arbitrary prefix for byte/hex, and if it's not exactly two characters, this would be wrong - right?

GeorgeSittas: The current implementation takes care of `X'..'` and `[0x|0X]...` literals and these changes are robust with respect to this. We can adapt the scanner accordingly in the future if there are any issues, but the current approach should be as robust as it already was (at least it will be with the changes I plan to make soon).

vegarsti: Okay, sounds good! Thanks!

sqlglot/dialects/dialect.py
```diff
@@ -70,30 +70,38 @@ def __new__(cls, clsname, bases, attrs):
...
-            ] = lambda self, e: f"{hs_start}{int(self.sql(e, 'this')):X}{hs_end}"
-        if (
-            klass.tokenizer_class._BYTE_STRINGS
-            and exp.ByteString not in klass.generator_class.TRANSFORMS
-        ):
-            be_start, be_end = list(klass.tokenizer_class._BYTE_STRINGS.items())[0]
-            klass.generator_class.TRANSFORMS[
-                exp.ByteString
-            ] = lambda self, e: f"{be_start}{self.sql(e, 'this')}{be_end}"
+        tokenizer_dict = klass.tokenizer_class.__dict__
```
tobymao, line 73: why are you doing this? can you just do hasattr?

GeorgeSittas: Nope, tried it already, `hasattr` returns the attribute if it's defined in the superclass and so it doesn't work in this case.

sqlglot/dialects/dialect.py
```diff
@@ -70,30 +70,38 @@ def __new__(cls, clsname, bases, attrs):
...
+        tokenizer_transforms = klass.generator_class.TRANSFORMS
+
+        if exp.BitString not in tokenizer_transforms and "bitstring_sql" not in tokenizer_dict:
```
tobymao, line 76: i think you can make this dry

GeorgeSittas: we can dry this up for bitstrings and hexstrings, but bytestrings have a small difference: they don't convert any value, but rather just do `self.sql(e, "this")`

sqlglot/generator.py
```diff
@@ -716,13 +740,22 @@ def tablealias_sql(self, expression: exp.TableAlias) -> str:
         return f"{alias}{columns}"
 
     def bitstring_sql(self, expression: exp.BitString) -> str:
-        return self.sql(expression, "this")
+        this = self.sql(expression, "this")
+        if self.bit_start or self.bit_end:
```
tobymao, line 744: why are you checking of self.bit_end? can bit_end ever be empty? can bit start be empty?

GeorgeSittas: > can bit_end ever be empty

That's correct, for example in Clickhouse we have `BIT_STRINGS = [("0b", "")]`, so `bit_end` would be empty.

> can bit start be empty?

Not with the current formats allowed, so I can simplify these checks to `if self.bit_start`.

GeorgeSittas: My initial reasoning was that by checking `bit_start` it'd be more robust (e.g. for future cases where we might get literals that only have an ending delimiter), but on second thought we can probably postpone it until there's an actual issue.

tobymao (approved):

vegarsti:
Eagerly looking forward to this being released 👀

Issue #1555, opened 2023-05-05
Postgres' inline power operator has incorrect operator precedence


vegarsti:
In Postgres, `x / y ^ z` is parsed as `(x / y) ^ z` when it should be `x / (y ^ z)`:

```
>>> import sqlglot
>>> import sqlglot
>>> sqlglot.__version__
'12.1.0'
>>> sqlglot.parse_one("select x / y ^ z", read="postgres")
(SELECT expressions:
  (POW this:
    (DIV this:
      (COLUMN this:
        (IDENTIFIER this: x, quoted: False)), expression:
      (COLUMN this:
        (IDENTIFIER this: y, quoted: False))), expression:
    (COLUMN this:
      (IDENTIFIER this: z, quoted: False))))
```

Docs link: https://www.postgresql.org/docs/15/functions-math.html

Test case in https://github.com/tobymao/sqlglot/pull/1556/commits/4eb737d79434b4a6c6d8510fb8035a0738ca6d37:
https://github.com/tobymao/sqlglot/blob/4eb737d79434b4a6c6d8510fb8035a0738ca6d37/tests/dialects/test_postgres.py#L177-L183

vegarsti:
I think this is the culprit: https://github.com/tobymao/sqlglot/blob/6124d0c2e413357f5fe947ae6d24ef341c0f8d55/sqlglot/dialects/postgres.py#L276-L279

It should have function precedence (higher than div, mul) instead of factor precedence (same as div, mul)

tobymao:
i'm looking at this

vegarsti:
Thanks!

PR #1556, opened 2023-05-05
Fix(Postgres): Correct operator precedence for ^ inline power operator

vegarsti:
To fix https://github.com/tobymao/sqlglot/issues/1555


vegarsti:
Closing because this shouldn't be open as a draft here. sorry!

PR #1565, opened 2023-05-08
Fix(presto, spark): remove WITHIN GROUP when transpiling percentile_[cont|disc]

GeorgeSittas:
Example usage:

```python
>>> sqlglot.transpile("WITH sales(amount) AS (SELECT * FROM (VALUES (100), (200), (300), (400), (500), (600), (700), (800), (900), (1000)) sales(amount)) SELECT PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY amount) AS median FROM sales", "postgres", "trino")[0]
'WITH sales(amount) AS (SELECT * FROM (VALUES (100), (200), (300), (400), (500), (600), (700), (800), (900), (1000)) AS sales(amount)) SELECT APPROX_PERCENTILE(amount, 0.5) AS median FROM sales'
```

In Postgres:
```python
georgesittas=# WITH sales(amount) AS (SELECT * FROM (VALUES (100), (200), (300), (400), (500), (600), (700), (800), (900), (1000)) sales(amount)) SELECT PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY amount) AS median FROM sales;
 median
--------
    550
(1 row)
```

In Trino:
```python
trino> WITH sales(amount) AS (SELECT * FROM (VALUES (100), (200), (300), (400), (500), (600), (700), (800), (900), (1000)) AS sales(amount)) SELECT APPROX_PERCENTILE(amount, 0.5) AS median FROM sales;
 median 
--------
    600 
(1 row)
```

From what I can see, there's no exact percentile function in Presto / Trino, hence the different column value. Not sure if we can do something better than what's proposed in this PR.

cc: @vegarsti

References:
- https://trino.io/docs/current/functions/aggregate.html
- https://www.postgresql.org/docs/15/functions-aggregate.html#FUNCTIONS-ORDEREDSET-TABLE
- https://docs.databricks.com/sql/language-manual/functions/percentile_approx.html
- https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -336,6 +336,9 @@ class Presto(Dialect):
             exp.UnixToTime: rename_func("FROM_UNIXTIME"),
             exp.UnixToTimeStr: lambda self, e: f"CAST(FROM_UNIXTIME({self.sql(e, 'this')}) AS VARCHAR)",
             exp.VariancePop: rename_func("VAR_POP"),
+            exp.WithinGroup: transforms.preprocess(
+                [transforms.remove_within_group_for_percentiles]
+            ),
         }
 
         def interval_sql(self, expression: exp.Interval) -> str:
@@ -348,13 +351,3 @@ class Presto(Dialect):
             modes = expression.args.get("modes")
             modes = f" {', '.join(modes)}" if modes else ""
             return f"START TRANSACTION{modes}"
-
-        def withingroup_sql(self, expression: exp.WithinGroup) -> str:
-            if isinstance(expression.this, (exp.PercentileCont, exp.PercentileDisc)) and isinstance(
-                expression.expression, exp.Order
-            ):
-                # Unwrap an Order > Ordered > obj expression into just obj (e.g. a Column)
-                input_value = expression.expression.expressions[0].this
-                return self.func("APPROX_PERCENTILE", input_value, expression.this.this)
-
-            return super().withingroup_sql(expression)
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -348,3 +348,13 @@ class Presto(Dialect):
             modes = expression.args.get("modes")
             modes = f" {', '.join(modes)}" if modes else ""
             return f"START TRANSACTION{modes}"
+
+        def withingroup_sql(self, expression: exp.WithinGroup) -> str:
+            if isinstance(expression.this, (exp.PercentileCont, exp.PercentileDisc)) and isinstance(
+                expression.expression, exp.Order
+            ):
+                # Unwrap an Order > Ordered > obj expression into just obj (e.g. a Column)
+                input_value = expression.expression.expressions[0].this
+                return self.func("APPROX_PERCENTILE", input_value, expression.this.this)
+
+            return super().withingroup_sql(expression)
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/presto.py
```diff
@@ -348,3 +348,13 @@ def transaction_sql(self, expression: exp.Transaction) -> str:
             modes = expression.args.get("modes")
             modes = f" {', '.join(modes)}" if modes else ""
             return f"START TRANSACTION{modes}"
+
+        def withingroup_sql(self, expression: exp.WithinGroup) -> str:
```
tobymao, line 352: should this instead be a transform? i think spark would use it as well

tobymao: looks like duckdb already does this transformation https://duckdb.org/docs/sql/aggregates.html

tobymao: expression.find(exp.Ordered)

GeorgeSittas: > should this instead be a transform? i think spark would use it as well

Sure

> looks like duckdb already does this transformation

Awesome, good to know.

> expression.find(exp.Ordered)

Wdym?

tobymao: ya

GeorgeSittas: Done

sqlglot/dialects/spark2.py
```diff
@@ -177,31 +177,34 @@ class Generator(Hive.Generator):
         TRANSFORMS = {
```
GeorgeSittas, line 177: Added `exp.WithinGroup` transform, the rest were just sorted in alphabetical order.

sqlglot/transforms.py
```diff
@@ -229,12 +229,26 @@ def remove_target_from_merge(expression: exp.Expression) -> exp.Expression:
...
 
 
+def remove_within_group_for_percentiles(expression: exp.Expression) -> exp.Expression:
+    if (
+        isinstance(expression, exp.WithinGroup)
+        and isinstance(expression.this, (exp.PercentileCont, exp.PercentileDisc))
+        and isinstance(expression.expression, exp.Order)
+    ):
+        quantile = expression.this.this
+        input_value = t.cast(exp.Ordered, expression.find(exp.Ordered)).this
```
GeorgeSittas, line 239: The cast was added to suppress a MyPy error. Since `expression.expression` is an `exp.Order`, it'll always contain an `exp.Ordered` expression for its `expressions` arg, so this should be safe to use.

sqlglot/transforms.py
```diff
@@ -244,9 +258,28 @@ def preprocess(
...
+            # has the same type as the final expression and there's no _sql method available for it,
+            # because then it'd re-enter _to_sql.
+            if expression_type is type(expression):
+                raise ValueError(
+                    f"Expression type {expression.__class__.__name__} requires a _sql method in order to be transformed."
+                )
+
+            return transforms_handler(self, expression)
+
+        raise ValueError(f"Unsupported expression type {expression.__class__.__name__}.")
```
GeorgeSittas, line 283: The motivation behind these changes was that we don't have a `_sql` method for `exp.ApproxQuantile`, so the transformation I added would fail at `getattr`. I think it makes `_to_sql` a bit more robust.

GeorgeSittas (commented):
Left a few comments for clarity. The changes in `test_postgres` were mostly style.

tobymao (approved):

PR #1568, opened 2023-05-08
feat(Clickhouse): Support large data types

matthax:
This PR aims to fix a few issues I've identifier with certain data types in ClickHouse.

The supported data types can be found [here](https://clickhouse.com/docs/en/sql-reference/data-types).

Some of these just required an addition to the `TYPE_MAPPING` but others were not implemented in `tokens.py` so I added implementations for them.

Additionally, I _think_ a few were missing from `Parser.TYPE_TOKENS` so I've added them there as well, though I am not certain this is correct. Please feel free to request changes or make suggestions as needed.

Lastly, I added a new test `test_signed_and_unsigned_types` to cover these data types in the `Clickhouse` dialect. Let me know if you'd like to see additional test cases, and where they should go.


sqlglot/dialects/clickhouse.py
```diff
@@ -33,11 +33,20 @@ class Tokenizer(tokens.Tokenizer):
...
+            "UINT32": TokenType.UINT,
             "INT64": TokenType.BIGINT,
-            "INT8": TokenType.TINYINT,
+            "UINT64": TokenType.UBIGINT,
+            "INT128": TokenType.INT128,
+            "UINT128": TokenType.UINT128,
+            "INT256": TokenType.INT256,
+            "UINT256": TokenType.UINT256,
             "TUPLE": TokenType.STRUCT,
+            # "MAP": TokenType.STRUCT,
```
GeorgeSittas, line 49: Why is this here? I think we already handle this correctly:
```python
>>> sqlglot.transpile("a::map", "clickhouse")
['CAST(a AS Map)']
```

GeorgeSittas (approved):
Looks good, not 100% sure if something's missing. @tobymao can you double check too?

Thanks for the PR!

sqlglot/dialects/clickhouse.py
tobymao, line 49:
```suggestion
```

tobymao (approved):
thanks for the great pr!

sqlglot/dialects/clickhouse.py
matthax: You do, I meant to remove that thank you!

PR #1569, opened 2023-05-09
fix: subquery selects

barakalon:
We weren't propagating `selects` from subqueries

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -846,11 +846,6 @@ class DerivedTable(Expression):
 
     @property
     def selects(self):
-        # For EXPLODE
-        alias = self.args.get("alias")
-        if alias:
-            return alias.columns
-
         if self.this and isinstance(self.this, Subqueryable):
             return self.this.selects
 
@@ -924,7 +919,12 @@ class Unionable(Expression):
 
 
 class UDTF(DerivedTable, Unionable):
-    pass
+    @property
+    def selects(self):
+        alias = self.args.get("alias")
+        if alias:
+            return alias.columns
+        return []
 
 
 class Cache(Expression):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -846,10 +846,7 @@ class DerivedTable(Expression):
 
     @property
     def selects(self):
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -846,10 +846,14 @@ def alias_column_names(self):
 
     @property
     def selects(self):
+        # For EXPLODE
```
GeorgeSittas, line 849: What does this comment mean?

GeorgeSittas (approved):

sqlglot/expressions.py
barakalon: From what I can tell, the code block below was meant to support EXPLODE expressions

barakalon: Let me see if I can clean this up..

sqlglot/expressions.py
```diff
@@ -846,10 +846,9 @@ def alias_column_names(self):
 
     @property
     def selects(self):
-        alias = self.args.get("alias")
+        if self.this and isinstance(self.this, Subqueryable):
+            return self.this.selects
 
-        if alias:
-            return alias.columns
         return []
```
GeorgeSittas, line 852: Maybe we can also do:
```python
return self.this.selects if isinstance(self.this, Subqueryable) else []
```

sqlglot/expressions.py
```diff
@@ -920,7 +919,12 @@ def except_(self, expression, distinct=True, dialect=None, **opts):
 
 
 class UDTF(DerivedTable, Unionable):
-    pass
+    @property
+    def selects(self):
+        alias = self.args.get("alias")
+        if alias:
+            return alias.columns
+        return []
```
GeorgeSittas, line 927: One liner here as well?

sqlglot/expressions.py
```diff
@@ -846,10 +846,9 @@ def alias_column_names(self):
 
     @property
     def selects(self):
-        alias = self.args.get("alias")
+        if self.this and isinstance(self.this, Subqueryable):
```
tobymao, line 849: don’t need the self.this

barakalon: doh

PR #1571, opened 2023-05-09
Support for parameterized queries in the ClickHouse Dialect

matthax:
ClickHouse [supports parameterization](https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters) in the form of `{placeholder : type}`, this PR adds support for that using the existing `Placeholder` type.

Notes:
- This PR makes it possible to generate a query like `SELECT * FROM :identifier` which _is_ possible in ClickHouse but not in most other database engines. I couldn't figure out a way to cleanly disable this. Any insight / suggestions would be great
- There's no way to currently test something like `INSERT INTO table_map VALUES ({'key1':1, 'key2':10}), ({'key1':2,'key2':20}), ({'key1':3,'key2':30})` because this statement isn't parsed correctly. I can address that issue in another PR.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -237,7 +237,7 @@ class ClickHouse(Dialect):
             return super().cte_sql(expression)
 
         def placeholder_sql(self, expression: exp.Placeholder) -> str:
-            placeholder_type = expression.args.get("type")
-            if not placeholder_type:
+            type_sql = self.sql(expression, "type")
+            if not type_sql:
                 self.unsupported("Cannot create a placeholder without a type.")
-            return f"{{{expression.name}: {self.sql(placeholder_type)}}}"
+            return f"{{{expression.name}: {type_sql}}}"
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -84,6 +84,39 @@ class ClickHouse(Dialect):
 
         LOG_DEFAULTS_TO_LN = True
 
+        def _parse_placeholder(self) -> t.Optional[exp.Expression]:
+            """
+            Parse a placeholder expression like SELECT {abc: UInt32} or FROM {table: Identifier}
+            https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters
+            """
+            index = self._index
+
+            if self._match(TokenType.L_BRACE):
+                # We might be getting a variable from a SELECT clause, or potentially an identifier from a FROM clause
+                if not self._match_set({TokenType.VAR, TokenType.TABLE}):
+                    return None
+
+                this = self._prev.text
+                # Ensure we've got a colon delimiting the placeholder name and placeholder type
+                if not self._match(TokenType.COLON):
+                    return None
+                data_type: t.Optional[t.Union[str, exp.Expression]] = self._parse_types()
+                # If we couldn't get the type, we could be looking at the "Identifier" literal
+                if not data_type:
+                    identifier_seq = "Identifier"
+                    if not self._match_text_seq(identifier_seq.upper()):
+                        return None
+                    data_type = identifier_seq
+
+                placeholder = self.expression(exp.Placeholder, this=this, type=data_type)
+
+                # Ensure we've matched our closing brace and have a valid placeholder expression
+                if self._match(TokenType.R_BRACE) and placeholder:
+                    return placeholder
+                # At this point, it wasn't a valid placeholder. Go back to the saved start index and try to parse again
+                self._retreat(index)
+            return None
+
         def _parse_in(
             self, this: t.Optional[exp.Expression], is_global: bool = False
         ) -> exp.Expression:
@@ -94,6 +127,10 @@ class ClickHouse(Dialect):
         def _parse_table(
             self, schema: bool = False, alias_tokens: t.Optional[t.Collection[TokenType]] = None
         ) -> t.Optional[exp.Expression]:
+            placeholder = self._parse_placeholder()
+            if placeholder:
+                return placeholder
+
             this = super()._parse_table(schema=schema, alias_tokens=alias_tokens)
 
             if self._match(TokenType.FINAL):
@@ -198,3 +235,9 @@ class ClickHouse(Dialect):
                 return self.sql(expression, "this")
 
             return super().cte_sql(expression)
+
+        def placeholder_sql(self, expression: exp.Placeholder) -> str:
+            placeholder_type = expression.args.get("type")
+            if not placeholder_type:
+                self.unsupported("Cannot create a placeholder without a type.")
+            return f"{{{expression.name}: {self.sql(placeholder_type)}}}"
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -84,6 +84,39 @@ class Parser(parser.Parser):
...
+        def _parse_placeholder(self) -> t.Optional[exp.Expression]:
+            """
+            Parse a placeholder expression like SELECT {abc: UInt32} or FROM {table: Identifier}
+            https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters
+            """
+            index = self._index
+
+            if self._match(TokenType.L_BRACE):
+                # We might be getting a variable from a SELECT clause, or potentially an identifier from a FROM clause
+                if not self._match_set({TokenType.VAR, TokenType.TABLE}):
```
tobymao, line 96: i don't think this is right, you're consuming the lbrace and then not doing anything,

just do 

this = self._parse_id_var()

sqlglot/dialects/clickhouse.py
```diff
@@ -84,6 +84,39 @@ class Parser(parser.Parser):
...
+                    return None
+
+                this = self._prev.text
+                # Ensure we've got a colon delimiting the placeholder name and placeholder type
+                if not self._match(TokenType.COLON):
```
tobymao, line 101: sqlglot is very permissive so we can make colon optional, get rid of the if and just do self._match(TokenType.COLON)

sqlglot/dialects/clickhouse.py
```diff
@@ -84,6 +84,39 @@ class Parser(parser.Parser):
...
+                    if not self._match_text_seq(identifier_seq.upper()):
+                        return None
+                    data_type = identifier_seq
+
+                placeholder = self.expression(exp.Placeholder, this=this, type=data_type)
+
+                # Ensure we've matched our closing brace and have a valid placeholder expression
+                if self._match(TokenType.R_BRACE) and placeholder:
+                    return placeholder
+                # At this point, it wasn't a valid placeholder. Go back to the saved start index and try to parse again
```
tobymao, line 116: do you have specific syntax that hits this retreat?

sqlglot/dialects/clickhouse.py
```diff
@@ -94,6 +127,10 @@ def _parse_in(
         def _parse_table(
             self, schema: bool = False, alias_tokens: t.Optional[t.Collection[TokenType]] = None
         ) -> t.Optional[exp.Expression]:
+            placeholder = self._parse_placeholder()
```
tobymao, line 130: i don't know if this is the right place to put this. perhaps the base parser needs to be adjusted to handle place holders

sqlglot/dialects/clickhouse.py
```diff
@@ -84,6 +84,39 @@ class Parser(parser.Parser):
...
+                # Ensure we've got a colon delimiting the placeholder name and placeholder type
+                if not self._match(TokenType.COLON):
+                    return None
+                data_type: t.Optional[t.Union[str, exp.Expression]] = self._parse_types()
+                # If we couldn't get the type, we could be looking at the "Identifier" literal
+                if not data_type:
+                    identifier_seq = "Identifier"
+                    if not self._match_text_seq(identifier_seq.upper()):
+                        return None
+                    data_type = identifier_seq
```
GeorgeSittas, line 109: I think this can be reduce to just `data_type = self._parse_type()`.

sqlglot/dialects/clickhouse.py
```diff
@@ -84,6 +84,39 @@ class Parser(parser.Parser):
...
         LOG_DEFAULTS_TO_LN = True
 
+        def _parse_placeholder(self) -> t.Optional[exp.Expression]:
+            """
+            Parse a placeholder expression like SELECT {abc: UInt32} or FROM {table: Identifier}
+            https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters
+            """
+            index = self._index
+
+            if self._match(TokenType.L_BRACE):
```
GeorgeSittas, line 94: 1. Can / should we just return `None` if this doesn't match?
2. Can we remove the retreat by returning early if the first two tokens are not `{` and a `VAR` or `TABLE`? Not sure if there are other tokens, but maybe we can get rid of the retreat below.

sqlglot/dialects/clickhouse.py
```diff
@@ -198,3 +235,9 @@ def cte_sql(self, expression: exp.CTE) -> str:
                 return self.sql(expression, "this")
 
             return super().cte_sql(expression)
+
+        def placeholder_sql(self, expression: exp.Placeholder) -> str:
+            placeholder_type = expression.args.get("type")
+            if not placeholder_type:
+                self.unsupported("Cannot create a placeholder without a type.")
+            return f"{{{expression.name}: {self.sql(placeholder_type)}}}"
```
GeorgeSittas, line 243:
```suggestion
        def placeholder_sql(self, expression: exp.Placeholder) -> str:
            type_sql = self.sql(expression, "type")
            if not type_sql:
                self.unsupported("Cannot create a placeholder without a type.")
            return f"{{{expression.name}: {type_sql}}}"
```

sqlglot/dialects/clickhouse.py
matthax: The reason I didn't go with `_parse_type()` here is because that has `check_funcs=True` which converts the `Map` type to a function instead of the `Map` type

e.g. 
```- SELECT {abc: UInt32}, {b: String}, {c: DateTime}, {d: map(String, [UInt8])}, {e: Tuple(UInt8, String)}
?                                                       ^           ^     ^
+ SELECT {abc: UInt32}, {b: String}, {c: DateTime}, {d: Map(String, Array(UInt8))}, {e: Tuple(UInt8, String)}
?                                                       ^           ^^^^^^     ^
```

matthax: Perhaps I should just update the dialect around `Map` as a type vs. `map` as a function

sqlglot/dialects/clickhouse.py
tobymao: yea i don't think so

GeorgeSittas:
@matthax any plans to finish this?

GeorgeSittas:
Sounds good, let us know if you need any help!

pkit:
@matthax 
> There's no way to currently test something like INSERT INTO table_map VALUES ({'key1':1, 'key2':10}), ({'key1':2,'key2':20}), ({'key1':3,'key2':30}) because this statement isn't parsed correctly. I can address that issue in another PR.

Clickhouse itself avoids parsing it till actual execution. So, don't bother.
See comments [here](https://github.com/ClickHouse/ClickHouse/blob/c306deb1aee7e4e7693ff38e9fc99ef0497c853a/programs/format/Format.cpp#L165)

GeorgeSittas:
@matthax https://github.com/tobymao/sqlglot/commit/50025eac213b2c4406fecaa2cb46b6108d9f75c8

matthax:
> @matthax went ahead and completed this in https://github.com/tobymao/sqlglot/commit/50025eac213b2c4406fecaa2cb46b6108d9f75c8

That's awesome, appreciate you taking this up!

PR #1574, opened 2023-05-09
Update the typing of other in Condition's magic methods

vchan:


GeorgeSittas (commented):
Let's make sure to also fix `other` in `_binop`: https://github.com/tobymao/sqlglot/pull/1574/files#diff-7857fedd1d1451b1b9a5b8efaa1cc292c02e7ee4f0d04d7e2f9d5bfb9565802cR715

sqlglot/expressions.py
```diff
@@ -712,7 +712,7 @@ def not_(self, copy=True):
         """
         return not_(self, copy=copy)
 
-    def _binop(self, klass: t.Type[E], other: ExpOrStr, reverse=False) -> E:
+    def _binop(self, klass: t.Type[E], other: t.Any, reverse=False) -> E:
```
GeorgeSittas, line 715:
```suggestion
    def _binop(self, klass: t.Type[E], other: t.Any, reverse: bool = False) -> E:
```

GeorgeSittas (approved):

PR #1582, opened 2023-05-11
Fix rename table expression for DuckDB

vchan:
Specifying the target table name with a schema/db causes a syntax error in DuckDB.


sqlglot/dialects/duckdb.py
```diff
@@ -240,3 +240,12 @@ def tablesample_sql(
...
             return super().tablesample_sql(expression, seed_prefix="REPEATABLE", sep=sep)
+
+        def renametable_sql(self, expression: exp.RenameTable) -> str:
+            """DuckDB only supports renaming a table in the same schema"""
+            expression = expression.copy()
+            target_table = expression.this
+            for arg in target_table.args:
+                if arg != "this":
+                    target_table.set(arg, None)
+            return super().renametable_sql(expression)
```
tobymao, line 251: super().renametable(expression.transform(lambda n: exp.table(n.this) if isinstance(n, exp.Table) else n))

PR #1592, opened 2023-05-11
Fix(oracle): change the data type parser to allow type specifiers

### LINKED ISSUE #1587: Oracle: VARCHAR(xxx CHAR) results in ParseError
In Oracle, you can specify the size of a VARCHAR type as VARCHAR(1234) but you can also specify whether the 1234 are BYTE or CHAR. These last keywords give an error in sqlglot.

```
import sqlglot
sqlglot.parse_one("SELECT CAST(NULL AS VARCHAR2(2328 CHAR)) AS COL1", read="oracle")
```

**Official Documentation**
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlqr/Data-Types.html#GUID-219C338B-FE60-422A-B196-2F0A01CAD9A4

GeorgeSittas:
Fixes #1587 

If we end up merging this, it will be a breaking change because the AST for data types like `VARCHAR(MAX)` will change.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2985,7 +2985,7 @@ class Boolean(Condition):
     pass
 
 
-class DataTypeSizeSpecifier(Expression):
+class DataTypeSize(Expression):
     arg_types = {"this": True, "expression": False}
 
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2985,6 +2985,10 @@ class Boolean(Condition):
     pass
 
 
+class DataTypeSizeSpecifier(Expression):
+    arg_types = {"this": True, "expression": False}
+
+
 class DataType(Expression):
     arg_types = {
         "this": True,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2811,6 +2811,12 @@ class Parser(metaclass=_Parser):
 
         return this
 
+    def _parse_type_size_specifier(self) -> t.Optional[exp.Expression]:
+        this = self._parse_conjunction()
+        if not this:
+            return None
+        return self.expression(exp.DataTypeSizeSpecifier, this=this, expression=self._parse_type())
+
     def _parse_types(self, check_func: bool = False) -> t.Optional[exp.Expression]:
         index = self._index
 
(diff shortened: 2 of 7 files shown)


sqlglot/parser.py
```diff
@@ -2789,6 +2789,12 @@ def _parse_type(self) -> t.Optional[exp.Expression]:
 
         return this
 
+    def _parse_type_size_specifier(self) -> t.Optional[exp.Expression]:
+        this = self._parse_conjunction()
+        if not this:
+            return None
+        return self.expression(exp.DataTypeSizeSpecifier, this=this, expression=self._parse_type())
+
```
GeorgeSittas, line 2797: I think we can make this a bit safer (i.e. backwards compatible) if we return `this`, in case `_parse_type` returns `None`.

tobymao: why was this a conjunction before?

tobymao: Does that mean that INT(x / 10000) is now a data type specifier? that should be fixed

sqlglot/expressions.py
```diff
@@ -2976,6 +2976,10 @@ class Boolean(Condition):
     pass
 
 
+class DataTypeSizeSpecifier(Expression):
```
tobymao, line 2979: let's just call this DataTypeSize

sqlglot/parser.py
GeorgeSittas: Correction: how we parse this depends on the context. Since `INT` is used as a function in this example, it will get parsed as an anonymous one (or as a `Cast` for spark -- see Vegard's recent additions). For example:

```
>>> sqlglot.parse_one("int(x / 100)")
(ANONYMOUS this: int, expressions:
  (DIV this:
    (COLUMN this:
      (IDENTIFIER this: x, quoted: False)), expression:
    (LITERAL this: 100, is_string: False)))
>>> sqlglot.parse_one("cast(x as int(x / 100))")  -- This doesn't make sense
(CAST this:
  (COLUMN this:
    (IDENTIFIER this: x, quoted: False)), to:
  (DATATYPE this: Type.INT, expressions:
    (DATATYPESIZE this:
      (DIV this:
        (COLUMN this:
          (IDENTIFIER this: x, quoted: False)), expression:
        (LITERAL this: 100, is_string: False))), nested: False, prefix: False))
>>> 
```

To conclude, we don't ever reach `_parse_type_size_specifier` in `sqlglot.parse_one("select int(x/100)")`.

GeorgeSittas: yea probably, I'm taking a look

GeorgeSittas: Seems like we can get rid of it.. all tests seem to pass in main with:
```
-                expressions = self._parse_csv(self._parse_conjunction)
+                expressions = self._parse_csv(self._parse_type)
```

GeorgeSittas: I think it's good now, let me know if I can make any other improvements.

tobymao (approved):

PR #1596, opened 2023-05-11
Fix rename table expression for BigQuery

vchan:
```sql
ALTER TABLE mydataset.mytable
RENAME TO mydataset.mynewtable;
```
causes the following error in BigQuery:
`name "mydataset.mynewtable" cannot contain dot.`

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -333,3 +333,11 @@ class BigQuery(Dialect):
 
         def with_properties(self, properties: exp.Properties) -> str:
             return self.properties(properties, prefix=self.seg("OPTIONS"))
+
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -333,3 +333,11 @@ def intersect_op(self, expression: exp.Intersect) -> str:
 
         def with_properties(self, properties: exp.Properties) -> str:
             return self.properties(properties, prefix=self.seg("OPTIONS"))
+
```
tobymao, line 336: isn't this copy and pasted? make this a flag on the generator base class and have this done in the base generator

GeorgeSittas (approved):
LGTM

PR #1598, opened 2023-05-11
Fix(snowflake): Handle form of CONVERT_TIMEZONE with a source TZ

pmsanford:
Snowflake's [CONVERT_TIMEZONE](https://docs.snowflake.com/en/sql-reference/functions/convert_timezone) has two forms:

```sql
CONVERT_TIMEZONE( <target_tz> , <source_timestamp> )
-- and
CONVERT_TIMEZONE( <source_tz> , <target_tz> , <source_timestamp_ntz> )
```

The second is required if the timestamp doesn't have timezone information. Currently, parsing
```sql
CONVERT_TIMEZONE('UTC', 'America/Los_Angeles', tzcolumn)
```
produces
```sql
CONVERT_TIMEZONE('UTC', 'America/Los_Angeles')
```
when re-rendered, which is invalid. This PR still treats the target-only case as `exp.AtTimeZone`, but parses the source-and-target case as `exp.Anonymous` so that rendering back to Snowflake will be correct.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -17,7 +17,7 @@ from sqlglot.dialects.dialect import (
     ts_or_ds_to_date_sql,
     var_map_sql,
 )
-from sqlglot.expressions import Literal
+from sqlglot.expressions import Literal, maybe_parse
 from sqlglot.helper import flatten, seq_get
 from sqlglot.parser import binary_range_parser
 from sqlglot.tokens import TokenType
@@ -143,6 +143,14 @@ def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
     return self.datatype_sql(expression)
 
 
+def _parse_convert_timezone(args: t.Sequence) -> exp.Expression:
+    if len(args) == 3:
+        converted: t.List[exp.Expression] = [maybe_parse(arg, dialect=Snowflake) for arg in args]
+        return exp.Anonymous(this="CONVERT_TIMEZONE", expressions=converted)
+    else:
+        return exp.AtTimeZone(this=seq_get(args, 1), zone=seq_get(args, 0))
+
+
 class Snowflake(Dialect):
     null_ordering = "nulls_are_large"
     time_format = "'yyyy-mm-dd hh24:mi:ss'"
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -143,6 +143,14 @@ def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
     return self.datatype_sql(expression)
 
 
+def _parse_convert_timezone(args: t.Sequence) -> exp.Expression:
+    if len(args) == 3:
+        converted: t.List[exp.Expression] = [maybe_parse(arg, dialect=Snowflake) for arg in args]
```
tobymao, line 148: why are you calling maybe_parse here? everything should already be parsed

GeorgeSittas, line 148: Is there a reason you used `maybe_parse` here? The `args` parameter should aready be the parsed args.

sqlglot/dialects/snowflake.py
```diff
@@ -143,6 +143,14 @@ def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
...
+        return exp.Anonymous(this="CONVERT_TIMEZONE", expressions=converted)
+    else:
+        return exp.AtTimeZone(this=seq_get(args, 1), zone=seq_get(args, 0))
```
GeorgeSittas, line 151:
```suggestion
    return exp.AtTimeZone(this=seq_get(args, 1), zone=seq_get(args, 0))
```

sqlglot/dialects/snowflake.py
```diff
@@ -143,6 +143,14 @@ def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
     return self.datatype_sql(expression)
 
 
+def _parse_convert_timezone(args: t.Sequence) -> exp.Expression:
+    if len(args) == 3:
+        converted: t.List[exp.Expression] = [maybe_parse(arg, dialect=Snowflake) for arg in args]
+        return exp.Anonymous(this="CONVERT_TIMEZONE", expressions=converted)
```
tobymao, line 149: exp.func("CONVERT_TIMEZONE", expression=...)

sqlglot/dialects/snowflake.py
```diff
@@ -143,6 +143,14 @@ def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
...
+    else:
```
tobymao, line 150: remove the else statement it's not needed

GeorgeSittas (commented):
Thanks for the PR! Looks good, left a comment.

sqlglot/dialects/snowflake.py
pmsanford: I initially tried `exp.func`, but since `CONVERT_TIMEZONE` is in the `parser.FUNCTIONS` dict, it ends up calling this function again recursively on line 5414 of expressions.py and causing a recursion loop. (that's also why I was calling maybe_parse - I was looking at the body of `exp.func`, but it makes sense that it's not necessary here now)

GeorgeSittas (approved):

tobymao (approved):

PR #1599, opened 2023-05-12
Fix: expand alias refs was buggy and did the samething expand lateral…

tobymao:
…s did


tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -401,3 +401,33 @@ SELECT ROW_NUMBER() OVER (PARTITION BY x.a ORDER BY x.b) AS row_num FROM x AS x
...
+  j + 1 AS k
+FROM x;
+SELECT x.a + 1 AS i, x.a + 1 + 1 AS j, x.a + 1 + 1 + 1 AS k FROM x AS x;
+
+# title: noop - reference comes before alias
+SELECT
+  b + 1 AS j,
+  x.a + 1 AS i
+FROM x;
+SELECT x.b + 1 AS j, x.a + 1 AS i FROM x AS x;
```
GeorgeSittas, line 422: I think this test is supposed to have `i + 1 AS j` as the first projection.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -158,45 +160,43 @@ def _expand_using(scope, resolver):
...
-        elif isinstance(node, exp.Expression) and not isinstance(node, exp.Subqueryable):
-            exp.replace_children(node, transform, source_first)
+        for column in node.find_all(exp.Column):
+            if id(column) not in columns:
+                continue
 
-        return node
+            table = resolver.get_table(column.name) if resolve_agg and not column.table else None
+            if table and column.find_ancestor(exp.AggFunc):
+                column.set("table", table)
```
GeorgeSittas, line 185: Why are we checking for aggregates here? Is this part covered in a test?

GeorgeSittas: Should we copy `table` to avoid sharing the same `Identifier` object here?

tests/fixtures/optimizer/qualify_columns.sql
GeorgeSittas: cc: @barakalon

barakalon: lol yeah. I think my original test was wrong.

PR #1601, opened 2023-05-12
Fix(Presto): Add correct alias when transpiling SEQUENCE with alias to UNNEST(SEQUENCE(...))

### LINKED ISSUE #1600: Presto: Transpiling SEQUENCE with alias to UNNEST(SEQUENCE(...)) has incorrect alias
#1527 implemented a fix for transpiling SEQUENCE to Presto, so we now wrap SEQUENCE in an UNNEST call if used as a source. But there's an edge case left, where if we original SEQUENCE has an alias, that alias is not translated properly to the Presto code.

Code snippet
```
>>> import sqlglot
>>> sqlglot.__version__
'12.3.0'
>>> sqlglot.transpile("SELECT * FROM t CROSS JOIN GENERATE_SERIES(2, 4) s", read="postgres", write="presto")[0]
'SELECT * FROM t CROSS JOIN UNNEST(SEQUENCE(2, 4)) AS s'
```

The unnest requires a tuple (?) alias, so we need `s` to be `_u(s)` (like in #1495). Thus the transpiled SQL here should be 
```
SELECT * FROM t CROSS JOIN UNNEST(SEQUENCE(2, 4)) AS _u(s)
```

vegarsti:
Fixes #1600

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -149,6 +149,8 @@ def _sequence_sql(self: generator.Generator, expression: exp.GenerateSeries) -> 
 
     sql = self.func("SEQUENCE", start, end, step)
     if isinstance(expression.parent, exp.Table):
+        if expression.parent.alias:
+            expression.parent.args["alias"] = f"_u({expression.parent.alias})"
         sql = f"UNNEST({sql})"
 
     return sql
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -127,38 +127,6 @@ def _ts_or_ds_add_sql(self: generator.Generator, expression: exp.TsOrDsAdd) -> s
     )
 
 
-def _sequence_sql(self: generator.Generator, expression: exp.GenerateSeries) -> str:
-    start = expression.args["start"]
-    end = expression.args["end"]
-    step = expression.args.get("step")
-
-    target_type = None
-
-    if isinstance(start, exp.Cast):
-        target_type = start.to
-    elif isinstance(end, exp.Cast):
-        target_type = end.to
-
-    if target_type and target_type.this == exp.DataType.Type.TIMESTAMP:
-        to = target_type.copy()
-
-        if target_type is start.to:
-            end = exp.Cast(this=end, to=to)
-        else:
-            start = exp.Cast(this=start, to=to)
-
-    sql = self.func("SEQUENCE", start, end, step)
-    if isinstance(expression.parent, exp.Table):
-        if expression.parent.alias:
-            source_name = "_u"
-            expression.parent.set(
-                "alias", exp.TableAlias(this=source_name, columns=[expression.parent.alias])
-            )
-        sql = f"UNNEST({sql})"
-
-    return sql
-
-
 def _ensure_utf8(charset: exp.Literal) -> None:
     if charset.name.lower() != "utf-8":
         raise UnsupportedError(f"Unsupported charset {charset}")
@@ -299,7 +267,6 @@ class Presto(Dialect):
             exp.Decode: _decode_sql,
             exp.DiToDate: lambda self, e: f"CAST(DATE_PARSE(CAST({self.sql(e, 'this')} AS VARCHAR), {Presto.dateint_format}) AS DATE)",
             exp.Encode: _encode_sql,
-            exp.GenerateSeries: _sequence_sql,
             exp.Group: transforms.preprocess([transforms.unalias_group]),
             exp.Hex: rename_func("TO_HEX"),
             exp.If: if_sql,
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/presto.py
```diff
@@ -149,6 +149,8 @@ def _sequence_sql(self: generator.Generator, expression: exp.GenerateSeries) ->
 
     sql = self.func("SEQUENCE", start, end, step)
     if isinstance(expression.parent, exp.Table):
+        if expression.parent.alias:
+            expression.parent.args["alias"] = f"_u({expression.parent.alias})"
         sql = f"UNNEST({sql})"
```
vegarsti, line 154: This is a bit rough!

1. I'm not sure if mutating the parent expression like this is OK
2. We probably need to ensure that the table alias name (`_u`) isn't taken

tests/dialects/test_postgres.py
```diff
@@ -233,6 +233,14 @@ def test_postgres(self):
                 "tsql": "SELECT * FROM t CROSS JOIN GENERATE_SERIES(2, 4)",
             },
         )
+        self.validate_all(
+            "SELECT * FROM t CROSS JOIN GENERATE_SERIES(2, 4) AS s",
+            write={
+                "postgres": "SELECT * FROM t CROSS JOIN GENERATE_SERIES(2, 4) AS s",
+                "presto": "SELECT * FROM t CROSS JOIN UNNEST(SEQUENCE(2, 4)) AS _u(s)",
+                "trino": "SELECT * FROM t CROSS JOIN UNNEST(SEQUENCE(2, 4)) AS _u(s)",
+            },
```
vegarsti, line 242: I didn't add a `tsql` test case here, because I'm not familiar with it, so any pointers would be helpful!

sqlglot/dialects/presto.py
```diff
@@ -149,6 +149,8 @@ def _sequence_sql(self: generator.Generator, expression: exp.GenerateSeries) ->
 
     sql = self.func("SEQUENCE", start, end, step)
     if isinstance(expression.parent, exp.Table):
+        if expression.parent.alias:
+            expression.parent.args["alias"] = f"_u({expression.parent.alias})"
```
vegarsti, line 153: If we want to ensure `_u` isn't taken, we could do this? Like in https://github.com/tobymao/sqlglot/pull/1501/files.

```suggestion
            from sqlglot.optimizer.scope import build_scope
            taken_source_names = set(build_scope(expression.parent_select).selected_sources)
            source_name = find_new_name(taken_source_names, "_u")
            expression.parent.args["alias"] = f"{source_name}({expression.parent.alias})"
```

vegarsti, line 153: This is a bit rough!

1. I'm not sure if mutating the parent expression like this is OK
2. We probably need to ensure that the table alias name (`_u`) isn't taken? If so, we could do this? Like in https://github.com/tobymao/sqlglot/pull/1501/files.

```suggestion
        if expression.parent.alias:
            from sqlglot.optimizer.scope import build_scope
            taken_source_names = set(build_scope(expression.parent_select).selected_sources)
            source_name = find_new_name(taken_source_names, "_u")
            expression.parent.args["alias"] = f"{source_name}({expression.parent.alias})"
```

GeorgeSittas: > I'm not sure if mutating the parent expression like this is OK

It's not -- generator methods are not supposed to mutate without first copying. Additionally, the way you tried to mutate the `alias` arg is wrong for two more reasons:

1) You need to store a `TableAlias` with the respective `Identifier` under it that wraps around the target name, instead of a raw string (see example). There are some helper methods for similar things, in `expressions.py`.
2) Mutations in an `Expression` must always be done through the `.set(...)` method, otherwise the resulting AST will be incorrect (i.e. there won't be a link between children / parent expressions).

> We probably need to ensure that the table alias name (_u) isn't taken?

That is correct, for the result to be fully correct you'd have to avoid name clashes. At this point maybe I'd rather create a transformation for this instead of doing it in a `_sql` method (i.e. create a new function in `transforms.py` that does any needed changes in the AST).

I'd suggest to get these changes in with your PR and then I'll look if I can refactor some stuff, e.g. finding all used source names more easily etc.

```python
>>> sqlglot.parse_one("select * from foo a")
(SELECT expressions:
  (STAR ), from:
  (FROM expressions:
    (TABLE this:
      (IDENTIFIER this: foo, quoted: False), alias:
      (TABLEALIAS this:
        (IDENTIFIER this: a, quoted: False)))))
```

tests/dialects/test_postgres.py
GeorgeSittas: For `tsql` we should just transpile the original query back to itself:

Screenshot 2023-05-12 at 4 34 24 PM

GeorgeSittas (commented):
Thanks for catching this, left a couple of comments.

tests/dialects/test_postgres.py
vegarsti: What's this site? seems handy!

sqlglot/dialects/presto.py
vegarsti: Perfect, thanks so much for the detailed response. Will try to get this in soon.

tests/dialects/test_postgres.py
GeorgeSittas: It's https://onecompiler.com/sqlserver

sqlglot/dialects/presto.py
```diff
@@ -149,6 +149,11 @@ def _sequence_sql(self: generator.Generator, expression: exp.GenerateSeries) ->
 
     sql = self.func("SEQUENCE", start, end, step)
     if isinstance(expression.parent, exp.Table):
+        if expression.parent.alias:
```
tobymao, line 152: i don't understand how this code works. first of all you're mutating expression.parent but you shouldn't be.

but second of all sql is already generated at this point, this line of code isn't really doing what you're intending to do

vegarsti: Yeah, has a chat with George, working on a proper solution!

vegarsti: Probably the reason it works is it generates the sql for the leaf node first  and then for the parent, an alias?

vegarsti:
@tobymao I pushed what I have so far, but I haven't been bale to find out how to replace the alias correctly.

tobymao:
https://github.com/tobymao/sqlglot/commit/2f7473b65edaa75f05400129ffb1e9dedffecbbc

vegarsti:
Ah, smart to do it on Table

PR #1603, opened 2023-05-12
Feat(clickhouse): parse ternary operator

### LINKED ISSUE #1602: Clickhouse: Ternary if parsing
Minimal code to reproduce:
```python
import sqlglot

sqlglot.parse_one(
    "flagcolumn ? created : now()",
    read="clickhouse"
)
```

While trying to parse expression with clickhouse ternary if, get the following error:
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 28.
  flagcolumn ? created : now()
```

Documentation:
https://clickhouse.com/docs/en/sql-reference/functions/conditional-functions

GeorgeSittas:
Fixes #1602

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -552,12 +552,12 @@ class Parser(metaclass=_Parser):
     RANGE_PARSERS = {
         TokenType.BETWEEN: lambda self, this: self._parse_between(this),
         TokenType.GLOB: binary_range_parser(exp.Glob),
-        TokenType.OVERLAPS: binary_range_parser(exp.Overlaps),
+        TokenType.ILIKE: binary_range_parser(exp.ILike),
         TokenType.IN: lambda self, this: self._parse_in(this),
+        TokenType.IRLIKE: binary_range_parser(exp.RegexpILike),
         TokenType.IS: lambda self, this: self._parse_is(this),
         TokenType.LIKE: binary_range_parser(exp.Like),
-        TokenType.ILIKE: binary_range_parser(exp.ILike),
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -2985,8 +2999,18 @@ def _parse_column(self) -> t.Optional[exp.Expression]:
                 field = self._parse_types()
                 if not field:
                     self.raise_error("Expected type")
-            elif op:
+            elif op and self._curr:
```
GeorgeSittas, line 3002: Without `self._curr` here we might raise an index error in the `self._advance` call below (happened earlier).

GeorgeSittas:
With these additions our parser is actually more lenient than Clickhouse's. For example, we can handle this:

```python
>>> import sqlglot
>>> sqlglot.parse_one("x ? y ? 1 : 2 : 3")
(IF this:
  (COLUMN this:
    (IDENTIFIER this: x, quoted: False)), true:
  (IF this:
    (COLUMN this:
      (IDENTIFIER this: y, quoted: False)), true:
    (LITERAL this: 1, is_string: False), false:
    (LITERAL this: 2, is_string: False)), false:
  (LITERAL this: 3, is_string: False))
```

Seems like Clickhouse fails to parse this:
Screenshot 2023-05-12 at 6 03 24 PM

tobymao:
maybe this should be clickhouse only so you don't have to worry about conflicts.

also, maybe you should treat it like the exponent use case for postgres. it can have another precedence

PR #1611, opened 2023-05-13
Fix: remove unconditional expression copy

tobymao:
is this what you meant? can you go through the different usages and see what makes the most sense to do here? copy false or true as default?


GeorgeSittas:
Yeah this is what I meant. It'd be a bit safer to do `copy=True` by default here.

PR #1614, opened 2023-05-13
Fix(clickhouse): join type/kind ordering issues

pkit:
CH can parse both `LEFT ANY JOIN` and `ANY LEFT JOIN` successfully 
But the canonical order (by the docs) is `LEFT ANY JOIN` 
This PR parses "inverted" order and produces a canonical one

Notes:
- CH uses `ASOF`, `SEMI`, `ANTI`, `ANY` as reserved words. Can be tested by doing something like: 
`FROM t semi SEMI LEFT JOIN` - fails
`FROM t "semi" SEMI LEFT JOIN` - passes
That's why I have removed them from `TABLE_ALIAS_TOKENS`


pkit:
@tobymao wasn't hard, but probably I'm missing something?

tobymao:
nice

PR #1615, opened 2023-05-13
Fix(clickhouse): `USING`, `[PRIMARY|FOREIGN] KEY` allows unwrapped col list

pkit:
I.e. both `USING a, b, c` and `USING (a, b, c)` are allowed 
Although the wrapped variant is a canonical one (see docs)
https://clickhouse.com/docs/en/sql-reference/statements/select/join#examples

Unwrapped columns and expression lists are allowed in other places too. Will update later.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -195,6 +195,11 @@ class ClickHouse(Dialect):
                 return self.expression(exp.Quantile, this=params[0], quantile=this)
             return self.expression(exp.Quantile, this=this, quantile=exp.Literal.number(0.5))
 
+        def _parse_wrapped_id_vars(
+            self, optional: bool = False
+        ) -> t.List[t.Optional[exp.Expression]]:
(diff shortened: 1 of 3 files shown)


tobymao:
everywhere with wrapped id vars in clickhouse allows unwrapped? your change seems a bit dangerous

should you target it just for the using query?

pkit:
> everywhere with wrapped id vars in clickhouse allows unwrapped? your change seems a bit dangerous

Yes, as far as I can see in the ClickHouse code.
But I will re-check it.
It has not a lot of places with id lists: `primary|foreign key` and `using`
Will update

pkit:
It looks good.
Right now in sqlglot `_parse_wrapped_id_vars()` is used in the follwoing places:
- `primary|foreign key` - ok
- `join using` - ok
- `references` - (only if `(` is present)
- `unique` - (only if `(` is present)

tobymao:
thanks for checking i’ll review and merge tonight. at a wedding today.

PR #1617, opened 2023-05-14
Feat(optimizer): optimize pivots

GeorgeSittas:
Fixes https://github.com/tobymao/sqlglot/discussions/1449

This PR introduces optimizer logic for handling the PIVOT operator, i.e. explode & qualify the columns of the table it produces where necessary. It's a first draft towards the general solution, which is more difficult because we may have multiple JOINs and / or PIVOTs, UNPIVOTs chained together in non-trivial ways.

We could also try to transform PIVOT operators using `exp.Case` expressions. This seems a bit less straightforward to me if we want to get it right in all cases. For example:

1. What do we `GROUP BY` if there are additional columns besides what's referenced in the `exp.Pivot` expression? 
2. Can we always map the aggregations trivially into projections? How would we handle stuff like `COUNT(*)`?
3. How does this transformation behave when there are JOINs and / or other (UN)PIVOT operators applied?

The major advantage of this approach, though, would be that we'd get back a canonical query without `PIVOT`s.

References:
- https://docs.aws.amazon.com/redshift/latest/dg/r_FROM_clause-pivot-unpivot-examples.html
- https://stackoverflow.com/questions/5846007/sql-query-to-pivot-a-column-using-case-when
- https://www.phind.com/search?cache=5b176c80-4468-4d71-8b33-f3bd9bb37578

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -275,7 +275,7 @@ def no_tablesample_sql(self: Generator, expression: exp.TableSample) -> str:
 
 def no_pivot_sql(self: Generator, expression: exp.Pivot) -> str:
     self.unsupported("PIVOT unsupported")
-    return self.sql(expression)
+    return ""
 
 
 def no_trycast_sql(self: Generator, expression: exp.TryCast) -> str:
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -133,6 +133,10 @@ def _eliminate_union(scope, existing_ctes, taken):
 
 
 def _eliminate_derived_table(scope, existing_ctes, taken):
+    # This ensures we don't drop the "pivot" arg from a pivoted subquery
+    if scope.parent.pivots:
+        return None
+
     parent = scope.expression.parent
     name, cte = _new_cte(scope, existing_ctes, taken)
 
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -167,6 +167,7 @@ def _mergeable(outer_scope, inner_scope, leave_tables_isolated, from_or_join):
         and isinstance(inner_select, exp.Select)
         and not any(inner_select.args.get(arg) for arg in UNMERGABLE_ARGS)
         and inner_select.args.get("from")
+        and not outer_scope.pivots
         and not any(e.find(exp.AggFunc, exp.Select) for e in inner_select.expressions)
         and not (leave_tables_isolated and len(outer_scope.selected_sources) > 1)
         and not (
--- a/sqlglot/optimizer/pushdown_projections.py
+++ b/sqlglot/optimizer/pushdown_projections.py
@@ -39,8 +39,9 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
     for scope in reversed(traverse_scope(expression)):
         parent_selections = referenced_columns.get(scope, {SELECT_ALL})
 
-        if scope.expression.args.get("distinct"):
-            # We can't remove columns SELECT DISTINCT nor UNION DISTINCT
+        if scope.expression.args.get("distinct") or scope.parent and scope.parent.pivots:
+            # We can't remove columns SELECT DISTINCT nor UNION DISTINCT. The same holds if
+            # we select from a pivoted source in the parent scope.
             parent_selections = {SELECT_ALL}
 
         if isinstance(scope.expression, exp.Union):
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -6,7 +6,6 @@ import typing as t
 from sqlglot import alias, exp
 from sqlglot.errors import OptimizeError
 from sqlglot.helper import seq_get
-from sqlglot.optimizer.expand_laterals import expand_laterals as _expand_laterals
 from sqlglot.optimizer.scope import Scope, traverse_scope, walk_in_scope
 from sqlglot.schema import Schema, ensure_schema
 
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -302,7 +302,6 @@ def _expand_stars(scope, resolver, using_column_tables):
 
     has_pivoted_source = pivot and not pivot.args.get("unpivot")
     if has_pivoted_source:
-        # We're using a dictionary here in order to preserve order
         pivot_columns = set(col.output_name for col in pivot.find_all(exp.Column))
 
         pivot_output_columns = [col.output_name for col in pivot.args.get("columns", [])]
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -253,7 +253,7 @@ def _qualify_columns(scope, resolver):
             if scope.pivots and not column.find_ancestor(exp.Pivot):
                 # If the column is under the Pivot expression, we need to qualify it
                 # using the name of the pivoted source instead of the pivot's alias
-                column.set("table", scope.pivots[0].args["alias"])
+                column.set("table", exp.to_identifier(scope.pivots[0].alias))
                 continue
 
             column_table = resolver.get_table(column_name)
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -5,6 +5,8 @@ import typing as t
 
 from sqlglot import alias, exp
 from sqlglot.errors import OptimizeError
+from sqlglot.helper import seq_get
+from sqlglot.optimizer.expand_laterals import expand_laterals as _expand_laterals
 from sqlglot.optimizer.scope import Scope, traverse_scope, walk_in_scope
 from sqlglot.schema import Schema, ensure_schema
 
@@ -65,7 +67,7 @@ def validate_qualify_columns(expression):
     for scope in traverse_scope(expression):
         if isinstance(scope.expression, exp.Select):
             unqualified_columns.extend(scope.unqualified_columns)
-            if scope.external_columns and not scope.is_correlated_subquery:
+            if scope.external_columns and not scope.is_correlated_subquery and not scope.pivots:
                 column = scope.external_columns[0]
                 raise OptimizeError(
                     f"""Column '{column}' could not be resolved{f" for table: '{column.table}'" if column.table else ''}"""
@@ -249,6 +251,12 @@ def _qualify_columns(scope, resolver):
                 raise OptimizeError(f"Unknown column: {column_name}")
 
         if not column_table:
+            if scope.pivots and not column.find_ancestor(exp.Pivot):
+                # If the column is under the Pivot expression, we need to qualify it
+                # using the name of the pivoted source instead of the pivot's alias
+                column.set("table", scope.pivots[0].args["alias"])
+                continue
+
             column_table = resolver.get_table(column_name)
 
             # column_table can be a '' because bigquery unnest has no table alias
@@ -272,6 +280,13 @@ def _qualify_columns(scope, resolver):
             if column_table:
                 column.replace(exp.Dot.build([exp.column(root, table=column_table), *parts]))
 
+    for pivot in scope.pivots:
+        for column in pivot.find_all(exp.Column):
+            if not column.table and column.name in resolver.all_columns:
+                column_table = resolver.get_table(column.name)
+                if column_table:
+                    column.set("table", column_table)
+
 
 def _expand_stars(scope, resolver, using_column_tables):
     """Expand stars to lists of column selections"""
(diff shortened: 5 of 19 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -275,7 +275,7 @@ def no_tablesample_sql(self: Generator, expression: exp.TableSample) -> str:
 
 def no_pivot_sql(self: Generator, expression: exp.Pivot) -> str:
     self.unsupported("PIVOT unsupported")
-    return self.sql(expression)
+    return ""
```
GeorgeSittas, line 278: This fixes an infinite loop when `self.unsupported` doesn't raise.

Btw, it seems like DuckDB [does support a PIVOT statement](https://duckdb.org/docs/sql/statements/pivot.html), although its syntax seems a bit different (although they say they also support the "standard" PIVOT syntax?). I tried some examples locally but couldn't get them to work, so not sure what's up with that. In any case, we can tackle this in a separate PR.

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -133,6 +133,10 @@ def _eliminate_union(scope, existing_ctes, taken):
 
 
 def _eliminate_derived_table(scope, existing_ctes, taken):
+    # This ensures we don't drop the "pivot" arg from a pivoted subquery
+    if scope.parent.pivots:
+        return None
```
GeorgeSittas, line 138: An example for why this was needed:

```sql
SELECT * FROM (SELECT x, y FROM tbl) PIVOT (SUM(x) FOR y IN ('a', 'b'))
```

Without the condition I added, the subquery would get eliminated and we'd lose the `pivot` arg because its attached in the `exp.Subquery` node, instead of its `this` arg (which is what the CTE is created of later on).

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -167,6 +167,7 @@ def _outer_select_joins_on_inner_select_join():
         and isinstance(inner_select, exp.Select)
         and not any(inner_select.args.get(arg) for arg in UNMERGABLE_ARGS)
         and inner_select.args.get("from")
+        and not outer_scope.pivots
```
GeorgeSittas, line 170: We cant merge subqueries that are pivoted, because again we'll lose the `pivot` arg. For example:

```python
>>> print(optimize("SELECT * FROM (SELECT a, b, c FROM sc.tb) PIVOT (SUM(c) FOR b IN ('x','y','z'))").sql(pretty=True))
SELECT
  "_q_1"."a" AS "a",
  "_q_1"."x" AS "x",
  "_q_1"."y" AS "y",
  "_q_1"."z" AS "z"
FROM "sc"."tb" AS "tb"  -- subquery was merged and we lost the pivot
```

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -39,8 +39,9 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
     for scope in reversed(traverse_scope(expression)):
         parent_selections = referenced_columns.get(scope, {SELECT_ALL})
 
-        if scope.expression.args.get("distinct"):
-            # We can't remove columns SELECT DISTINCT nor UNION DISTINCT
+        if scope.expression.args.get("distinct") or scope.parent and scope.parent.pivots:
+            # We can't remove columns SELECT DISTINCT nor UNION DISTINCT. The same holds if
+            # we select from a pivoted source in the parent scope.
```
GeorgeSittas, line 44: We can't remove columns from a subquery query when its pivoted, because the pivot produces new columns and so the subquery's selections might seem to be unused when they're actually used in the pivot itself.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -65,7 +66,7 @@ def validate_qualify_columns(expression):
     for scope in traverse_scope(expression):
         if isinstance(scope.expression, exp.Select):
             unqualified_columns.extend(scope.unqualified_columns)
-            if scope.external_columns and not scope.is_correlated_subquery:
+            if scope.external_columns and not scope.is_correlated_subquery and not scope.pivots:
```
GeorgeSittas, line 69: This check was needed because the columns qualified with the pivot's alias are treated as external by the scope and so we'd hit the "column could not be resolved for table ..." error.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -249,6 +250,12 @@ def _qualify_columns(scope, resolver):
                 raise OptimizeError(f"Unknown column: {column_name}")
 
         if not column_table:
+            if scope.pivots and not column.find_ancestor(exp.Pivot):
+                # If the column is under the Pivot expression, we need to qualify it
+                # using the name of the pivoted source instead of the pivot's alias
+                column.set("table", scope.pivots[0].args["alias"])
+                continue
```
GeorgeSittas, line 257: Take for example:

```sql
SELECT x, y FROM u PIVOT (SUM(f) FOR h IN ('x', 'y'));
```

Here, we want to qualify `f` and `h` using `u` instead of the pivot's (generated) alias `_q_0`. The call to `find_ancestor` solves this problem, because otherwise we'd get `PIVOT (SUM(_q_0.f) FOR _q_0.h IN ('x', 'y'))`.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -352,6 +375,22 @@ def _add_replace_columns(expression, tables, replace_columns):
         replace_columns[id(table)] = columns
 
 
+def _add_pivot_columns(pivot, source_columns, columns):
+    pivot_output_columns = [col.output_name for col in pivot.args.get("columns", [])]
+    if not pivot_output_columns:
+        pivot_output_columns = [col.alias_or_name for col in pivot.expressions]
```
GeorgeSittas, line 381: The parser should always produce this `columns` arg, but this check was added as a sanity check / fallback in case we get an `exp.Pivot` expression that was programmatically produced and doesn't have it set.

GeorgeSittas:
Left some comments on the code for clarity, let me know if something's not clear. Interested to hear alternatives, would love to simplify this somehow.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -249,6 +250,12 @@ def _qualify_columns(scope, resolver):
                 raise OptimizeError(f"Unknown column: {column_name}")
 
         if not column_table:
+            if scope.pivots and not column.find_ancestor(exp.Pivot):
+                # If the column is under the Pivot expression, we need to qualify it
+                # using the name of the pivoted source instead of the pivot's alias
+                column.set("table", exp.to_identifier(scope.pivots[0].alias))
+                continue
```
GeorgeSittas, line 257: Take for example:

```sql
SELECT x, y FROM u PIVOT (SUM(f) FOR h IN ('x', 'y'));
```

Here, we want to qualify `f` and `h` using `u` instead of the pivot's (generated) alias `_q_0`. The call to `find_ancestor` solves this problem, because otherwise we'd get the (invalid) form:

```sql
PIVOT (SUM(_q_0.f) FOR _q_0.h IN ('x', 'y'))`
```

sqlglot/optimizer/scope.py
```diff
@@ -603,8 +615,13 @@ def _traverse_tables(scope):
...
             if table_name in scope.sources:
-                # This is a reference to a parent source (e.g. a CTE), not an actual table.
-                sources[source_name] = scope.sources[table_name]
+                # This is a reference to a parent source (e.g. a CTE), not an actual table, unless
+                # it is pivoted, because then we get back a new table and hence a new source.
+                pivots = expression.args.get("pivots")
+                if pivots:
+                    sources[pivots[0].alias] = expression
+                else:
+                    sources[source_name] = scope.sources[table_name]
```
GeorgeSittas, line 624: The example that motivated this change was:

```sql
WITH cte AS (...) SELECT * FROM cte PIVOT ...
```

Here, when we're qualifying tables we're skipping the pivoted `cte` table because it's treated as a `Scope` source in `for name, source in scope.sources.items(): ...`, so the pivot is never given an alias.

sqlglot/optimizer/qualify_tables.py
```diff
@@ -60,12 +64,16 @@ def qualify_tables(expression, db=None, catalog=None, schema=None):
                     source = source.replace(
                         alias(
                             source,
-                            name if name else next_name(),
+                            name or source.name or next_name(),
```
GeorgeSittas, line 67: See example shown in https://github.com/tobymao/sqlglot/pull/1617/files#r1193208207. If the pivot doesn't have an alias, we'll store `''` as a name in the `sources` dict, and so `source.name` here ensures we'll use `tbl` as the alias of the table source.

sqlglot/dialects/dialect.py
tobymao: shouldn't it be super() ?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -352,6 +375,22 @@ def _add_replace_columns(expression, tables, replace_columns):
...
+    pivot_output_columns = [col.output_name for col in pivot.args.get("columns", [])]
+    if not pivot_output_columns:
+        pivot_output_columns = [col.alias_or_name for col in pivot.expressions]
+
+    pivot_columns = set(column.output_name for column in pivot.find_all(exp.Column))
+    implicit_columns = list(set(source_columns) - pivot_columns)
+
+    columns.extend(
+        [
+            exp.alias_(exp.column(name, table=pivot.alias), name)
```
tobymao, line 388: copy=False here

sqlglot/optimizer/qualify_columns.py
```diff
@@ -352,6 +375,22 @@ def _add_replace_columns(expression, tables, replace_columns):
         replace_columns[id(table)] = columns
 
 
+def _add_pivot_columns(pivot, source_columns, columns):
+    pivot_output_columns = [col.output_name for col in pivot.args.get("columns", [])]
```
tobymao, line 379: this looks expensive, should we move these calls outside the nested loop

GeorgeSittas: Ok done

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: Done

sqlglot/dialects/dialect.py
tobymao: i'm sure many other dialects don't support pivots so this is fine

GeorgeSittas: This can't be done without running the optimizer and with the current approach we don't transform them to CASEs

sqlglot/optimizer/qualify_columns.py
```diff
@@ -297,9 +324,20 @@ def _expand_stars(scope, resolver, using_column_tables):
...
             if table not in scope.sources:
                 raise OptimizeError(f"Unknown table: {table}")
+
             columns = resolver.get_source_columns(table, only_visible=True)
 
             if columns and "*" not in columns:
+                if has_pivoted_source:
+                    implicit_columns = list(set(columns) - pivot_columns)
+                    new_selections.extend(
+                        [
```
tobymao, line 334: can you make this a generator instead ? ( instead of [

sqlglot/optimizer/scope.py
```diff
@@ -372,6 +373,17 @@ def join_hints(self):
             return []
         return self._join_hints
 
+    @property
+    def pivots(self):
+        if not self._pivots:
+            self._pivots = [
+                pivot
+                for node in self.tables + self.derived_tables
+                for pivot in node.args.get("pivots", [])
```
tobymao, line 382: pivots or [] instead of get("pivots", [])

sqlglot/optimizer/qualify_columns.py
```diff
@@ -297,9 +324,20 @@ def _expand_stars(scope, resolver, using_column_tables):
         for table in tables:
             if table not in scope.sources:
                 raise OptimizeError(f"Unknown table: {table}")
+
             columns = resolver.get_source_columns(table, only_visible=True)
 
             if columns and "*" not in columns:
+                if has_pivoted_source:
+                    implicit_columns = list(set(columns) - pivot_columns)
```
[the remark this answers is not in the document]
GeorgeSittas: From what I remember it depends on the dialect. I can check again soon and let you know.

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: Sure

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
GeorgeSittas: > doesn't set() change the determinsm?

I wouldn't expect it to, but I guess python doesn't provide guarantees on that? Honestly, I'd be surprised if an implementation made this non-deterministic, but using a dict seems like a good choice too for this. Will update.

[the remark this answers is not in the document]
tobymao: set() is 100% not deterministic in python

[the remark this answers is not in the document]
GeorgeSittas: I'm not sure what you mean by "not deterministic". If the argument to `set` is fixed, why would it be not deterministic?

tobymao (approved):

GeorgeSittas:
@tobymao made a few more changes, let me know if you take a look.

- Moved the `unqualify_pivot_columns` transform to spark since it's the only dialect using it.
- Created an `_unalias_pivot` transform just for spark that removes table aliases from pivots.
- Improved handling of the `alias` argument in `subquery`.
- Added more spark tests & new optimizer test that demonstrates the above transformations.

PR #1624, opened 2023-05-15
Feat(bigquery): allow first part of table names to contain dashes

GeorgeSittas:
Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#table_names

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -160,8 +160,6 @@ class BigQuery(Dialect):
         LOG_BASE_FIRST = False
         LOG_DEFAULTS_TO_LN = True
 
-        TABLE_NAME_ALLOWS_DASHES = True
-
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,  # type: ignore
             "DATE_TRUNC": lambda args: exp.DateTrunc(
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -2245,6 +2247,16 @@ def _parse_table_part() -> t.Optional[exp.Expression]:
             )
 
         table = _parse_table_part()
+
+        # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#table_names
+        if self.TABLE_NAME_ALLOWS_DASHES and isinstance(table, exp.Identifier):
```
tobymao, line 2252: why is this done here, shouldn't this be in parse_table_parts?

GeorgeSittas: Dashes are only allowed to the leftmost identifier according to the documentation, so I did this here to 1) capture that and 2) reduce unneeded overhead when parsing the rest of the parts.

GeorgeSittas: If this seems like unnecessary visual noise to have in the base parser, I can factor it out and have it only in BigQuery.

tobymao: let's just put this in bigquery because it's so unstandard

GeorgeSittas: done

sqlglot/dialects/bigquery.py
```diff
@@ -224,6 +224,24 @@ class Parser(parser.Parser):
             "OPTIONS": lambda self: exp.Properties(expressions=self._parse_with_property()),
         }
 
+        def _parse_table_part(self, schema: bool = False) -> t.Optional[exp.Expression]:
+            this = (
```
tobymao, line 228: can you make this a first class function and call it in both places so it's not copy and pasted?

GeorgeSittas: sure

sqlglot/dialects/bigquery.py
```diff
@@ -224,6 +224,24 @@ class Parser(parser.Parser):
...
+                (not schema and self._parse_function())
+                or self._parse_id_var(any_token=False)
+                or self._parse_string_as_identifier()
+            )
+
+            # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#table_names
+            if isinstance(this, exp.Identifier):
+                table_name = this.name
```
tobymao, line 236: what if it's a-b-c?

GeorgeSittas: It's handled in the following loop:

```python
>>> import sqlglot
>>> sqlglot.transpile("select * from a-b-c", "bigquery")
['SELECT * FROM a-b-c']
```

sqlglot/dialects/bigquery.py
GeorgeSittas: done

tests/dialects/test_bigquery.py
```diff
@@ -6,6 +6,10 @@ class TestBigQuery(Validator):
     dialect = "bigquery"
 
     def test_bigquery(self):
+        self.validate_identity("SELECT * FROM a-b-c")
```
myschkyna, line 9: Can we also add an example with multiple hyphens in combination with tokens of type TokenType.DOT?

```sql
SELECT * FROM a-b-c.mydataset.mytable
```

tobymao (approved):

tests/dialects/test_bigquery.py
GeorgeSittas: done

PR #1625, opened 2023-05-15
added support for query parameters parsing for bigquery

cheunhong:
queries such as 
```sql
SELECT count(DISTINCT `bank`.`country`) AS `country_unique_count` 
FROM `public`.`bank` 
WHERE `bank`.`created_at` BETWEEN @`created_at_1` AND @`created_at_2`
```
were unable to be parsed by bigquery parser


sqlglot/parser.py
```diff
@@ -4013,7 +4013,7 @@ def _parse_star(self) -> t.Optional[exp.Expression]:
 
     def _parse_parameter(self) -> exp.Expression:
         wrapped = self._match(TokenType.L_BRACE)
-        this = self._parse_var() or self._parse_primary()
+        this = self._parse_var() or self._parse_primary() or self._parse_id_var()
```
tobymao, line 4016:
```suggestion
        this = self._parse_var() or self._parse_primary() or self._parse_identifier()
```

sqlglot/dialects/bigquery.py
```diff
@@ -333,3 +333,6 @@ def intersect_op(self, expression: exp.Intersect) -> str:
 
         def with_properties(self, properties: exp.Properties) -> str:
             return self.properties(properties, prefix=self.seg("OPTIONS"))
+
+        def parameter_sql(self, expression: exp.Parameter) -> str:
+            return f"{self.PARAMETER_TOKEN}`{expression.this.this}`"
```
tobymao, line 338: can you show the documentation that this is required?

tobymao:
add some test cases please

sqlglot/dialects/bigquery.py
GeorgeSittas: The backticks are not necessary, and I'm not sure if this is always correct. For example, in https://cloud.google.com/bigquery/docs/parameterized-queries there's a query where you can omit them:

```
bq query \
   --use_legacy_sql=false \
   --parameter=corpus::romeoandjuliet \
   --parameter=min_word_count:INT64:250 \
   'SELECT
     word,
     word_count
   FROM
     `bigquery-public-data.samples.shakespeare`
   WHERE
     corpus = @corpus
   AND
     word_count >= @min_word_count
   ORDER BY
     word_count DESC;'
```

tobymao:
bc0b021d

PR #1630, opened 2023-05-15
Feat: create builders for the INSERT statement

GeorgeSittas:
Resolves https://github.com/tobymao/sqlglot/discussions/1621

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1479,6 +1479,88 @@ class Insert(Expression):
         "alternative": False,
     }
 
+    def insert(
+        self, expression: ExpOrStr, copy: bool = True, dialect: DialectType = None, **opts
+    ) -> Insert:
+        """
+        Set the INSERT's expression.
+
+        Example:
+            >>> Insert(this=Table(this=to_identifier("tbl"))).insert("SELECT * FROM tbl2").sql()
+            'INSERT INTO tbl SELECT * FROM tbl2'
+
+        Args:
+            expression: the SQL code string to parse.
+                If an `Expression` instance is passed, it will be used as-is.
+            copy: if `False`, modify this expression instance in-place.
+            dialect: the dialect used to parse the input expression.
+            other options to use to parse the input expressions.
+
+        Returns:
+            The modified expression.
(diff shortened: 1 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -1479,6 +1479,123 @@ class Insert(Expression):
         "alternative": False,
     }
 
+    def insert(
```
tobymao, line 1482: this method is a bit strange, i'm not sure we need it, why would you ever use this and not just exp.insert?

sqlglot/expressions.py
```diff
@@ -1479,6 +1479,123 @@ class Insert(Expression):
...
+        instance = _apply_builder(
+            expression, self, arg="expression", dialect=dialect, copy=copy, **opts
+        )
+
+        if overwrite:
+            instance.set("overwrite", True)
+
+        return instance
+
+    def into(
```
tobymao, line 1517: this method is also strange, isn't insert into always together?

sqlglot/expressions.py
```diff
@@ -4742,6 +4862,41 @@ def delete(
...
+        columns: optionally the table's column names.
+        overwrite: whether to INSERT OVERWRITE or not.
+        dialect: the dialect used to parse the input expressions.
+        copy: whether or not to copy the expression.
+        **opts: other options to use to parse the input expressions.
+
+    Returns:
+        Insert: the syntax tree for the INSERT statement.
+    """
+    return (
```
tobymao, line 4893: just implement all thaat stuff here and remove the Insert().insert.into, i'm not really sure i buy into that api

GeorgeSittas: Yeah.. I mimicked the other builders without thinking much about it, but what you're saying makes sense.

sqlglot/expressions.py
```diff
@@ -1479,6 +1479,42 @@ class Insert(Expression):
         "alternative": False,
     }
 
+    def with_(
```
GeorgeSittas, line 1482: Didn't remove this one to reduce the arg count in the `insert` method below.

PR #1635, opened 2023-05-16
Fix(clickhouse): map ApproxDistinct to uniq, AnyValue to any

### LINKED ISSUE #1633: APPROX_COUNT_DISTINCT and ANY_VALUE
There are at least 2 (unrelated) functions that exist in BigQuery and have equivalents in ClickHouse: `ANY_VALUE()` is `any()` (note: exclusively lower case, `ANY` won't work) and `APPROX_COUNT_DISTINCT` is `uniq()` (note: same)

**Official Documentation**
- https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#any_value
- https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#approx_count_distinct
- https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/any
- https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/uniq#agg_function-uniq

Sqlglot completely ignores this:
```python
>>> sqlglot.transpile(
...     "SELECT APPROX_COUNT_DISTINCT(x) FROM (SELECT ANY_VALUE(y) x FROM (SELECT 1 y))",
...     read="bigquery",
...     write="clickhouse"
... )[0]
'SELECT APPROX_COUNT_DISTINCT(x) FROM (SELECT ANY_VALUE(y) AS x FROM (SELECT 1 AS y))'
```
(I've put the 2 in the same query, but those are 2 close but unrelated issues)

I'm pretty sure there are a ton of other things that are done differently, and not sure if the issue is in the parsing from bigquery dialect or dumping into clickhouse dialect (though I imagine it's the 2nd one).

What would be the best course of action ?

GeorgeSittas:
Fixes #1633

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -77,7 +77,6 @@ class Parser(metaclass=_Parser):
 
     FUNCTIONS: t.Dict[str, t.Callable] = {
         **{name: f.from_arg_list for f in exp.ALL_FUNCTIONS for name in f.sql_names()},
-        "APPROX_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
         "DATE_TO_DATE_STR": lambda args: exp.Cast(
             this=seq_get(args, 0),
             to=exp.DataType(this=exp.DataType.Type.TEXT),
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3085,14 +3085,18 @@ class Parser(metaclass=_Parser):
         if not self._curr:
             return None
 
+        this = self._curr.text
         token_type = self._curr.token_type
 
+        upper = this.upper()
+        next_is_paren = self._next and self._next.token_type == TokenType.L_PAREN
+
         if self._match_set(self.NO_PAREN_FUNCTION_PARSERS, advance=False):
-            if self._next and self._next.token_type != TokenType.L_PAREN:
+            if not next_is_paren and upper not in self.FUNCTION_PARSERS:
                 self._advance()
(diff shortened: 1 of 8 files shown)


ewjoachim:
I wonder if the transpilation the other way around works (clickhouse > bigquery) given that we modified the parser of BQ and the Generator of CH and not the other way around.

GeorgeSittas:
No, it doesn't currently work backwards. I'm working on a fix, but have to make some adjustments in the parser to handle Clickhouse's partial case sensitivity in function names... *Sigh*.

sqlglot/parser.py
```diff
@@ -77,6 +77,7 @@ class Parser(metaclass=_Parser):
 
     FUNCTIONS: t.Dict[str, t.Callable] = {
         **{name: f.from_arg_list for f in exp.ALL_FUNCTIONS for name in f.sql_names()},
+        "APPROX_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
```
tobymao, line 80: you can just put this in the sql names in expression

GeorgeSittas: Ah right

GeorgeSittas: Should I also add `IFNULL` as an alternative name for `Coalesce` while I'm at it?

ewjoachim:
> ~to handle Clickhouse's partial case sensitivity in function names... _Sigh_~.

I feel you. I discovered this today and sighed as well. Not all SQL dialects are created equal :)

tobymao (approved):

PR #1638, opened 2023-05-16
Refactor(CI/CD): make it so that deployment only requires tag push

GeorgeSittas:


sqlglot/__init__.py
```diff
@@ -50,7 +50,11 @@
     T = t.TypeVar("T", bound=Expression)
 
 
-__version__ = "13.0.1"
+try:
+    from sqlglot._version import __version__, __version_tuple__  # type: ignore
+except ImportError:
+    pass
```
mpf82, line 56: ``sqlglot._version`` is currently missing.

Why fail silently?

GeorgeSittas: What do you mean?

mpf82: print("sqlglot.version", sqlglot.__version__)

> AttributeError: module 'sqlglot' has no attribute '__version__'

mpf82: I have checked out the github project and installed it as development packge

> pip install -e .

mpf82: But there is no ``_version.py`` in https://github.com/tobymao/sqlglot/tree/main/sqlglot

How can ``import sqlglot._version`` work for you?

GeorgeSittas: If you `pip install -e .` after cloning it, the `_version.py` file is generated.

GeorgeSittas: Right, what you're saying makes sense, although I'm not entirely sure if the SQLGlot `__version__` attribute is supposed to be part of the public API. I'd rather wait on Toby's / Iaroslav's feedback on this because I mimicked what we do in SQLMesh and so not sure if I'm missing any needed context on this.

GeorgeSittas: @mpf82 after chatting with toby we concluded that we'll just print a warning in case the import fails. My suggestion above would also require a `pip install sqlglot` to get the latest version for `__version__`, so it's not really any different compared to `pip install -e .`

PR #1646, opened 2023-05-18
Fix(clickhouse): use _parse_expression in _parse_if to allow ternaries

### LINKED ISSUE #1645: Clickhouse: Ternary if parsing (2)
Minimal code to reproduce:
```python
import sqlglot

exp = "if(bar(col), sign > 0 ? foo() : 0, 1)"
sqlglot.parse(exp, read='clickhouse')
```
While trying to parse expression with clickhouse ternary if, get the following error:
```
sqlglot.errors.ParseError: Expecting ). Line 1, Col: 24.
  if(bar(col), sign > 0 ? foo() : 0, 1)
```

Observed on using ternary if inside if functiion.

Documentation:
https://clickhouse.com/docs/en/sql-reference/functions/conditional-functions

GeorgeSittas:
Fixes #1645

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3532,7 +3532,7 @@ class Parser(metaclass=_Parser):
 
     def _parse_if(self) -> t.Optional[exp.Expression]:
         if self._match(TokenType.L_PAREN):
-            args = self._parse_csv(self._parse_conjunction)
+            args = self._parse_csv(self._parse_expression)
             this = exp.If.from_arg_list(args)
             self.validate_expression(this, args)
             self._match_r_paren()
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -121,6 +121,12 @@ class TestClickhouse(Validator):
     def test_ternary(self):
         self.validate_all("x ? 1 : 2", write={"clickhouse": "CASE WHEN x THEN 1 ELSE 2 END"})
         self.validate_all(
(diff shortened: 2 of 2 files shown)


sqlglot/parser.py
```diff
@@ -3532,7 +3532,7 @@ def _parse_case(self) -> t.Optional[exp.Expression]:
 
     def _parse_if(self) -> t.Optional[exp.Expression]:
         if self._match(TokenType.L_PAREN):
-            args = self._parse_csv(self._parse_conjunction)
+            args = self._parse_csv(self._parse_expression)
```
tobymao, line 3535: are ternary farther up the tree than conjunction? could conjunction call tern?

GeorgeSittas: That's how I had originally implemented it, but moved it one level above it because the logic cut across several methods and so was too invasive.

tobymao: but what is correct? what has more precedence

GeorgeSittas: Conjunction binds more tightly, i.e. `a and b ? foo : bar` is to be parsed as `(a and b) ? foo : bar`, so the current approach is reasonable with respect to that.

The only current issue is that in order to handle ternaries whenever we'd have to use `_parse_expression` instead of conjunction, unless we mix ternary and conjunction logic into `_parse_conjunction`, depending on whether `?` is matched. I'm not yet sure how straightforward this change will be, but could be an easy fix if you think it's worth it.

tobymao (approved):

PR #1651, opened 2023-05-18
Fix(postgres, redshift): use single string interval logic

### LINKED ISSUE #1649: Postgres interval not formatted correctly
Postgres interval [1] value is not formatted (or parsed?) correctly.

```py
query = """
SELECT GENERATE_SERIES('2023-01-01', '2024-01-01', INTERVAL '1 week') AS week
"""
parsed = sqlglot.parse_one(query, read="postgres")
print(parsed.sql(dialect="postgres"))
```

Output:
```sql
SELECT GENERATE_SERIES('2023-01-01', '2024-01-01', INTERVAL '1' week) AS week
```
Fails with `ERROR:  syntax error at or near "week"`

Expected output:
```sql
SELECT GENERATE_SERIES('2023-01-01', '2024-01-01', INTERVAL '1 week') AS week
```

[1] https://www.postgresql.org/docs/current/functions-datetime.html

GeorgeSittas:
Fixes #1649

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -304,7 +304,7 @@ class Postgres(Dialect):
             return self.expression(exp.Extract, this=part, expression=value)
 
     class Generator(generator.Generator):
-        INTERVAL_ALLOWS_PLURAL_FORM = False
+        SINGLE_STRING_INTERVAL = True
         LOCKING_READS_SUPPORTED = True
         JOIN_HINTS = False
         TABLE_HINTS = False
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1852,7 +1852,7 @@ class Generator:
 
         if self.SINGLE_STRING_INTERVAL:
             this = expression.this.name if expression.this else ""
-            return f"INTERVAL '{this}{unit}'"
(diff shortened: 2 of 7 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -304,7 +304,7 @@ def _parse_date_part(self) -> exp.Expression:
             return self.expression(exp.Extract, this=part, expression=value)
 
     class Generator(generator.Generator):
-        INTERVAL_ALLOWS_PLURAL_FORM = False
```
GeorgeSittas, line 307: I tested against both engines (postgres, redshift), this was invalid as plural forms are allowed.

sqlglot/generator.py
```diff
@@ -1852,7 +1852,7 @@ def interval_sql(self, expression: exp.Interval) -> str:
 
         if self.SINGLE_STRING_INTERVAL:
             this = expression.this.name if expression.this else ""
-            return f"INTERVAL '{this}{unit}'"
+            return f"INTERVAL '{this}{unit}'" if this else f"INTERVAL{unit}"
```
GeorgeSittas, line 1855: Needed because we can have `value::interval day`, which would be transpiled to `cast(value as interval ' day')`.

PR #1654, opened 2023-05-18
Fix(redshift): handle VALUES clause more robustly

GeorgeSittas:
cc: @eakmanrq

### DIFF
--- a/tests/dialects/test_redshift.py
+++ b/tests/dialects/test_redshift.py
@@ -182,6 +182,16 @@ class TestRedshift(Validator):
 
     def test_values(self):
         self.validate_all(
+            "SELECT * FROM (VALUES (1, 2)) AS t",
+            write={"redshift": "SELECT * FROM (SELECT 1, 2) AS t"},
+        )
+        self.validate_all(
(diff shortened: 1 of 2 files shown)


tests/dialects/test_redshift.py
```diff
@@ -181,6 +181,16 @@ def test_identity(self):
         )
 
     def test_values(self):
+        self.validate_all(
+            "SELECT * FROM (VALUES (1, 2)) AS t",
```
GeorgeSittas, line 185: This would previously cause an error if we were to transpile it to redshift:

```python
    for value, column_name in zip(row, expression.args["alias"].args["columns"])
TypeError: 'NoneType' object is not iterable
```

tests/dialects/test_redshift.py
```diff
@@ -181,6 +181,16 @@ def test_identity(self):
...
+            write={"redshift": "SELECT * FROM (SELECT 1, 2) AS t"},
+        )
+        self.validate_all(
+            "SELECT * FROM (VALUES (1)) AS t1(id) CROSS JOIN (VALUES (1)) AS t2(id)",
```
GeorgeSittas, line 189: Probably a rare use case of `VALUES`, but thought I'd include it anyway. We weren't removing the second `VALUES` clause because only `exp.From` was checked as a possible parent.

PR #1656, opened 2023-05-18
Generate DATE_PART instead of EXTRACT for Snowflake

pmsanford:
This PR replaces generation of [`EXTRACT`](https://docs.snowflake.com/en/sql-reference/functions/date_part) with generation of [`DATE_PART`](https://docs.snowflake.com/en/sql-reference/functions/date_part) in the Snowflake dialect.

The motivation for this change is that, for queries using `DATE_PART` where the first argument is a string, the generated SQL is rejected by Snowflake. Taking an example from the unit tests, if I attempt to run the following on Snowflake:
```sql
SELECT EXTRACT('month' FROM a)
```
I get the following error:
```
SQL compilation error: syntax error line 1 at position 23 unexpected 'FROM'. syntax error line 1 at position 29 unexpected ')'.
```
It succeeds compilation if I remove the quotes around `'month'`.

`EXTRACT(... FROM ...)` only supports unquoted arguments. `DATE_PART(..., ...)` supports both quoted and unquoted arguments - so `DATE_PART('month', a)` and `DATE_PART(month, a)` are both valid. Unfortunately, the quoted argument forms don't appear to be documented on the Snowflake site, but I have run into the quoted forms in the wild.

My initial approach was on the parsing side, but this would only fix Snowflake -> Snowflake generation. Making this change on the generation side will generate valid SQL regardless of the source dialect.

I also considered attempting to convert the first argument to the unquoted variant on generation - that seems viable as well, but more complicated, since `DATE_PART` accepts both quoted and unquoted arguments. If that seems like a better solution I'm happy to do a rev.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -119,6 +119,12 @@ def _parse_date_part(self: parser.Parser) -> t.Optional[exp.Expression]:
     return self.expression(exp.Extract, this=this, expression=expression)
 
 
+def _snowflake_generate_extract(self: Snowflake.Generator, expression: exp.Extract) -> str:
+    this = self.sql(expression, "this")
+    expression_sql = self.sql(expression, "expression")
+    return f"DATE_PART({this}, {expression_sql})"
+
+
 # https://docs.snowflake.com/en/sql-reference/functions/div0
 def _div0_to_if(args: t.Sequence) -> exp.Expression:
     cond = exp.EQ(this=seq_get(args, 1), expression=exp.Literal.number(0))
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -309,6 +315,7 @@ class Generator(generator.Generator):
             exp.DateStrToDate: datestrtodate_sql,
             exp.DataType: _datatype_sql,
             exp.DayOfWeek: rename_func("DAYOFWEEK"),
+            exp.Extract: _snowflake_generate_extract,
```
tobymao, line 318: can you just do exp.Extract: rename_func("DATE_PART")

pmsanford: Ah nice, even cleaner

tobymao:
tahnks!

PR #1665, opened 2023-05-19
Fix(optimizer): expand refs in QUALIFY into corresponding projections

### LINKED ISSUE #1661: Pushdown projections seems to handle QUALIFY clauses incorrectly
```
sql = """
with cte as (
  select
    a,
    row_number() over (
        partition by id, timestamp
        order by a desc
    ) as index

  from source
  qualify index
)

select a from cte
"""

schema = {"source":{"a": "STRING", "id": "STRING", "timestamp": "TIMESTAMP"}}
```

If you run, 

```
from sqlglot.optimizer.qualify_columns import qualify_columns, validate_qualify_columns
from sqlglot.optimizer.isolate_table_selects import isolate_table_selects
from sqlglot.optimizer.pushdown_predicates import pushdown_predicates
from sqlglot.optimizer.pushdown_projections import pushdown_projections
from sqlglot.optimizer.lower_identities import lower_identities
optimize(sql, dialect = "bigquery", schema = schema, rules = (lower_identities,
                                        qualify_tables,
                                        isolate_table_selects,
                                        qualify_columns,
                                        # pushdown_projections,
                                        )).sql(dialect="bigquery")
```

With the above code, SQLglot correctly returns:

```
WITH cte AS (
    SELECT
        source.a AS a,
        ROW_NUMBER() OVER (
            PARTITION BY source.id,
            source.timestamp
            ORDER BY
                source.a DESC
        ) AS index
    FROM
        source AS source QUALIFY index
)
SELECT
    cte.a AS a
FROM
    cte
```

But if you also run with pushdown_projections, the same optimization returns

```
WITH cte AS (
    SELECT
        source.a AS a
    FROM
        source AS source QUALIFY index
)
SELECT
    cte.a AS a
FROM
    cte
```

This is now invalid SQL, since the QUALIFY statement is included, but it's definition in the SELECT clause is not included.

I ran the above code for Snowflake, Postgres, and Redshift and see the same thing, so it appears to be a cross-dialect issue.

GeorgeSittas:
Fixes #1661

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -194,6 +194,7 @@ def _expand_alias_refs(scope: Scope, resolver: Resolver) -> None:
     replace_columns(expression.args.get("where"))
     replace_columns(expression.args.get("group"))
     replace_columns(expression.args.get("having"), resolve_agg=True)
+    replace_columns(expression.args.get("qualify"), resolve_agg=True)
     replace_columns(expression.args.get("order"), expand=False, resolve_agg=True)
     scope.clear_cache()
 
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlglot/optimizer/qualify_columns.py
```diff
@@ -194,6 +194,7 @@ def replace_columns(
     replace_columns(expression.args.get("where"))
     replace_columns(expression.args.get("group"))
     replace_columns(expression.args.get("having"), resolve_agg=True)
+    replace_columns(expression.args.get("qualify"), resolve_agg=True)
```
GeorgeSittas, line 197: Checking whether we can get rid of `resolve_agg` here..

PR #1666, opened 2023-05-20
Feat!(schema, snowflake): normalize snowflake identifiers to uppercase

GeorgeSittas:
This PR aims to improve our handling of Snowflake identifiers [1] in the schema and optimizer modules, as we've had several issues in the past due to quote / case mismatching. I'll highlight the main changes:

- [BREAKING] Normalize unquoted Snowflake identifiers to uppercase in `MappingSchema`.
- [BREAKING] Renamed `lower_identities` to `normalize_identifiers`, added a `dialect` argument which is used to decide whether to normalize to uppercase or lowercase characters.
- [BREAKING] Added a `dialect` argument in `canonicalize` which is used in `should_identify`.
- Added ~a `dialects_match` helper~ implementations of `__eq__` and `__hash__` for both `_Dialect` and `Dialect` to enable easy equivalence checking for `DialectType` objects. A dialect's hash value is now its name's hash value.
- Added a `dialect` argument in `should_identify` to control whether to use `isupper` or `islower`.
- Added a `dialect` argument in several methods of `MappingSchema`.

Note that the names generated by the optimizer, such as `_q_0`, are not converted to uppercase because they're created after the `normalize_identifiers` pass. This is also the reason why they remain unquoted after `optimize` for Snowflake, since they're lowercase and hence are deemed "unsafe to quote" by `should_identify`.

I'm putting this up to get some early feedback -- still need to do some testing on this.

[1] https://docs.snowflake.com/en/sql-reference/identifiers-syntax


sqlglot/dialects/dialect.py
```diff
@@ -110,6 +110,10 @@ class Dialect(metaclass=_Dialect):
     parser_class = None
     generator_class = None
 
+    @property
+    def __name__(self) -> str:
```
tobymao, line 114: why do you need this?

sqlglot/helper.py
```diff
@@ -421,18 +422,34 @@ def first(it: t.Iterable[T]) -> T:
     return next(i for i in it)
 
 
-def should_identify(text: str, identify: str | bool) -> bool:
+def dialects_match(this: DialectType, other: DialectType) -> bool:
+    from sqlglot.dialects import Dialect
+
```
tobymao, line 427: can't you just do this == other if they're both dialects?

sqlglot/schema.py
```diff
@@ -209,38 +237,55 @@ def add_table(
...
-    def column_names(self, table: exp.Table | str, only_visible: bool = False) -> t.List[str]:
-        table_ = self._normalize_table(self._ensure_table(table))
-        schema = self.find(table_)
+    def column_names(
+        self,
+        table: exp.Table | str,
+        only_visible: bool = False,
+        dialect: DialectType = None,
+    ) -> t.List[str]:
+        dialect = dialect or self.dialect
```
tobymao, line 249: you shouldn't have to do this in every place, normalize_table and ensure_table should handle it right

sqlglot/helper.py
tobymao: delete this entire method, overide def __eq__ on the dialect class and have it match the string

tobymao: then you just just do everywhere

dialect == "snowflake" and you're done

tests/fixtures/optimizer/normalize_identifiers.sql
```diff
@@ -1,21 +1,37 @@
...
+SELECT A FROM X;
+
 SELECT "A" FROM "X";
 SELECT "A" FROM "X";
 
 SELECT a AS A FROM x;
 SELECT a AS A FROM x;
 
+# dialect: snowflake
+SELECT A AS a FROM X;
```
tobymao, line 15: why isn't the alias caps?

sqlglot/helper.py
GeorgeSittas: `DialectType` matches strings, `Dialect`s and Type[Dialect]s too. The last one can return the dialect's name through the `__name__` attribute, which is why I also defined `__name__` in `Dialect`. So basically the reasoning was that we convert any strings to `Dialect`s and we can compare them by comparing their `__name__`s.

tests/fixtures/optimizer/normalize_identifiers.sql
GeorgeSittas: For the same reason we don't lowercase them in the test case right above this

sqlglot/schema.py
GeorgeSittas: good point thanks

sqlglot/dialects/dialect.py
GeorgeSittas: explained below

sqlglot/helper.py
tobymao: but you don’t need the match method can’t you just do dialect == str

GeorgeSittas: You mean just move the logic inside Dialect's eq?

tobymao: why are you getting an actual dialect class? but yes moving logic into eq

GeorgeSittas: Sure, can try to move this into eq.

> Why are you getting an actual dialect class?

Because that's allowed in DialectType's definition right now. I guess the point is to be able to do something like:

```
from sqlglot.dialects import Snowflake

parse_one(..., read=Snowflake)
```

i.e. not have to instantiate the dialect class necessarily to use methods that expect a dialect argument.

tobymao: can you override equality for classses?

tobymao: so Snoeflake == “snowflake”

tobymao: maybe on the meta class?

tests/fixtures/optimizer/normalize_identifiers.sql
tobymao: it seems like it should though no?

tobymao: isn’t it an identifier?

GeorgeSittas: I think @barakalon wrote this pass, he might be able to provide more context. There is special handling for aliases:

```
# We need to leave the output aliases unchanged, so the selects need special handling
```

tobymao: ok fine

sqlglot/helper.py
GeorgeSittas: updated

sqlglot/dialects/dialect.py
```diff
@@ -42,6 +42,20 @@ class Dialects(str, Enum):
 class _Dialect(type):
     classes: t.Dict[str, t.Type[Dialect]] = {}
 
+    def __eq__(cls, other: DialectType) -> bool:  # type: ignore
```
tobymao, line 45: why is it type ignore? other should be t.Any not DialecTType

sqlglot/dialects/dialect.py
```diff
@@ -42,6 +42,20 @@ class Dialects(str, Enum):
...
+        if cls is other:
+            return True
+        if isinstance(other, str):
+            return cls is cls.get(other.lower())
```
tobymao, line 49: i don't think you need to call lower, no one should be calling SNOWFLAKE

sqlglot/dialects/dialect.py
```diff
@@ -42,6 +42,20 @@ class Dialects(str, Enum):
...
+        if isinstance(other, Dialect):
+            return cls is type(other)
+
+        return False
+
+    def __hash__(cls) -> int:
+        # This will ensure that equal _Dialects have the same hashes
+        return super().__hash__()
```
tobymao, line 57: can you explain again why this was necessary

sqlglot/dialects/dialect.py
```diff
@@ -110,6 +124,9 @@ class Dialect(metaclass=_Dialect):
     parser_class = None
     generator_class = None
 
+    def __eq__(self, other: DialectType) -> bool:  # type: ignore
```
tobymao, line 127: t.Any

sqlglot/dialects/dialect.py
```diff
@@ -110,6 +124,9 @@ class Dialect(metaclass=_Dialect):
...
+        return type(self) == other
```
tobymao, line 128: why do you need this one? shouldn't the metaclass wokr?

sqlglot/schema.py
```diff
@@ -262,63 +301,80 @@ def _normalize(self, schema: t.Dict) -> t.Dict:
...
 
-        return identifier.name if identifier.quoted else identifier.name.lower()
+        name = identifier.name
+
+        if identifier.quoted:
+            return name
+
+        # Only Snowflake resolves unquoted identifiers as uppercase.
+        # https://docs.snowflake.com/en/sql-reference/identifiers-syntax
+        return name.upper() if dialect == "snowflake" else name.lower()
```
tobymao, line 339: maybe we should make this a dialect flag, there could be others in the future, and i don't like how snowflake is hardcoded in 2 places now

sqlglot/dialects/dialect.py
GeorgeSittas: Without overriding it I was getting the following error:

```
TypeError: unhashable type: '_Dialect'
```

<details close>
<summary>Full stack trace</summary>

```python
ERROR: tests.test_transpile (unittest.loader._FailedTest)
ImportError: Failed to import test module: tests.test_transpile
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "sqlglot/tests/test_transpile.py", line 5, in <module>
    from sqlglot import parse_one, transpile
  File "sqlglot/sqlglot/__init__.py", line 13, in <module>
    from sqlglot.dialects.dialect import Dialect as Dialect, Dialects as Dialects
  File "sqlglot/sqlglot/dialects/__init__.py", line 59, in <module>
    from sqlglot.dialects.bigquery import BigQuery
  File "sqlglot/sqlglot/dialects/bigquery.py", line 9, in <module>
    from sqlglot.dialects.dialect import (
  File "sqlglot/sqlglot/dialects/dialect.py", line 228, in <module>
    DialectType = t.Union[str, Dialect, t.Type[Dialect], None]
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/typing.py", line 312, in inner
    return func(*args, **kwds)
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/typing.py", line 403, in __getitem__
    return self._getitem(self, parameters)
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/typing.py", line 516, in Union
    parameters = _remove_dups_flatten(parameters)
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/typing.py", line 281, in _remove_dups_flatten
    return tuple(_deduplicate(params))
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/typing.py", line 255, in _deduplicate
    all_params = set(params)
TypeError: unhashable type: '_Dialect'
```

Seems like the error is caused because of the `DialectType = Union[... t.Type[Dialect] ...]` definition; Python internally stores these types in a set which is probably a hash under the hood, and so when it tries to store `t.Type[Dialect]` it fails with the above error.

</details>

See https://docs.python.org/3/reference/datamodel.html#object.__hash__

sqlglot/dialects/dialect.py
GeorgeSittas: Nope, without it:

```python
AssertionError: <sqlglot.dialects.snowflake.Snowflake object at 0x125740700> != 'snowflake'
```

sqlglot/schema.py
GeorgeSittas: Agreed, will try to refactor

sqlglot/dialects/dialect.py
GeorgeSittas: Updated so that we do `__hash__ = type.__hash__` as per the documentation.

sqlglot/dialects/dialect.py
GeorgeSittas: fixed

sqlglot/dialects/dialect.py
GeorgeSittas: fixed

sqlglot/dialects/dialect.py
GeorgeSittas: fixed

sqlglot/dialects/dialect.py
GeorgeSittas: Updated yet again so that we actually hash the dialect's _name_ now, instead of using `type.__hash__`. This was needed to enable checks like `dialect in {"snowflake", "bigquery"}`.

sqlglot/schema.py
GeorgeSittas: Took a stab at this, let me know what you think. Happy to clean up if needed.

PR #1671, opened 2023-05-22
Fix(parser): add asc, desc to id var tokens

### LINKED ISSUE #1669: Parsing fails when selecting from a column called `desc` on Snowflake
Hi, I found a parsing issue when we read from a column called `desc` in Snowflake.

```
import sqlglot

my_sql  = """
with my_cte as (
    select 'a' as desc
)

select 
    desc as description 
from my_cte
"""

sqlglot.transpile(my_sql, read="snowflake")
```

raises:
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 7, Col: 8.
```

The code runs correctly in Snowflake though.

GeorgeSittas:
Fixes #1669


tests/fixtures/identity.sql
```diff
@@ -822,3 +822,5 @@ SELECT NEXT VALUE FOR db.schema.sequence_name OVER (ORDER BY foo), col
 SELECT PERCENTILE_CONT(x, 0.5) OVER ()
 SELECT PERCENTILE_CONT(x, 0.5 RESPECT NULLS) OVER ()
 SELECT PERCENTILE_CONT(x, 0.5 IGNORE NULLS) OVER ()
+WITH my_cte AS (SELECT 'a' AS desc) SELECT desc AS description FROM my_cte
```
tobymao, line 825: does this break ordered?

select x from y order by x desc

GeorgeSittas:
```python
>>> import sqlglot
>>> sqlglot.parse_one("select x from y order by x desc")
(SELECT expressions:
  (COLUMN this:
    (IDENTIFIER this: x, quoted: False)), from:
  (FROM this:
    (TABLE this:
      (IDENTIFIER this: y, quoted: False))), order:
  (ORDER expressions:
    (ORDERED this:
      (COLUMN this:
        (IDENTIFIER this: x, quoted: False)), desc: True, nulls_first: False)))
```

GeorgeSittas: Seems like `_parse_ordered` and `_parse_primary_key` are the only places where we reference `TokenType.DESC/ASC`. Let me take a closer look and I'll reply with an ACK if the execution flow looks ok.

PR #1687, opened 2023-05-24
Feat: implement transform to add column names to recursive CTEs

### LINKED ISSUE #1683: Trino recursive CTE requires column list
The following is what SQLGlot outputs when generating a recursive CTE.

```
>>> query = """WITH RECURSIVE t AS (
...     SELECT 1 as n
...     UNION ALL
...     SELECT n + 1 as n FROM t WHERE n < 4
... )
... SELECT sum(n) FROM t
... """
>>> import sqlglot
>>> sqlglot.__version__
'13.3.1'
>>> print(sqlglot.transpile(query, read='postgres', write='trino')[0])
WITH RECURSIVE t AS (SELECT 1 AS n UNION ALL SELECT n + 1 AS n FROM t WHERE n < 4) SELECT SUM(n) FROM t
```

However, this fails to execute in Trino (and probably Presto?), because it requires the table alias here to specify the column list, so in this case it should have been `WITH RECURSIVE t(n) AS ...`, which runs. If the column list length doesn't match, then it will also fail.

https://trino.io/docs/current/sql/select.html#with-recursive-clause

GeorgeSittas:
Fixes #1683

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -259,7 +259,10 @@ def add_recursive_cte_column_names(expression: exp.Expression) -> exp.Expression
                 if isinstance(query, exp.Union):
                     query = query.this
 
-                cte.args["alias"].set("columns", [s.alias or next_name() for s in query.selects])
+                cte.args["alias"].set(
+                    "columns",
+                    [exp.to_identifier(s.alias_or_name or next_name()) for s in query.selects],
+                )
 
     return expression
 
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -247,6 +247,19 @@ def remove_within_group_for_percentiles(expression: exp.Expression) -> exp.Expre
     return expression
 
 
+def add_recursive_cte_column_names(expression: exp.Expression) -> exp.Expression:
+    if isinstance(expression, exp.With) and expression.recursive:
+        for cte in expression.expressions:
+            if not cte.args["alias"].columns:
+                query = cte.this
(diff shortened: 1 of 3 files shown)


vegarsti (approved):

sqlglot/transforms.py
```diff
@@ -247,6 +247,19 @@ def remove_within_group_for_percentiles(expression: exp.Expression) -> exp.Expre
...
 
+def add_recursive_cte_column_names(expression: exp.Expression) -> exp.Expression:
+    if isinstance(expression, exp.With) and expression.recursive:
+        for cte in expression.expressions:
+            if not cte.args["alias"].columns:
+                query = cte.this
+                if isinstance(query, exp.Union):
+                    query = query.this
+
+                cte.args["alias"].set("columns", [s.alias for s in query.selects if s.alias])
```
tobymao, line 258: what happens if there is no alias? should we make one up?

GeorgeSittas: I think we can do that yeah. Will get on it in a bit.

vegarsti: FWIW the only accessible names from outside the CTE are the names in the column list, e.g.

Screenshot 2023-05-24 at 18 37 16

vegarsti: Right, yeah! The above was only for Trino. But you can do `select *` there too

GeorgeSittas: Done

PR #1689, opened 2023-05-25
Chore!: improve python type hints

GeorgeSittas:
I did a sweep in the codebase to clean up the type hints. All files under `sqlglot/dialects` should be now fully typed.

This is still a WIP, gonna continue iterating on it tomorrow to do any remaining fixes.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -347,7 +347,7 @@ def var_map_sql(
 
 def format_time_lambda(
     exp_class: t.Type[E], dialect: str, default: t.Optional[bool | str] = None
-) -> t.Callable[[t.Sequence], E]:
+) -> t.Callable[[t.List], E]:
     """Helper used for time expressions.
 
     Args:
@@ -359,7 +359,7 @@ def format_time_lambda(
         A callable that can be used to return the appropriately formatted time expression.
     """
 
-    def _format_time(args: t.Sequence):
+    def _format_time(args: t.List):
         return exp_class(
             this=seq_get(args, 0),
             format=Dialect[dialect].format_time(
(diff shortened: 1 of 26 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -347,7 +347,7 @@ def var_map_sql(
 
 def format_time_lambda(
     exp_class: t.Type[E], dialect: str, default: t.Optional[bool | str] = None
-) -> t.Callable[[t.Sequence], E]:
+) -> t.Callable[[t.List], E]:
```
tobymao, line 350: why are you changing this one?

sqlglot/dialects/dialect.py
```diff
@@ -396,8 +396,8 @@ def create_with_partitions_sql(self: Generator, expression: exp.Create) -> str:
 
 def parse_date_delta(
     exp_class: t.Type[E], unit_mapping: t.Optional[t.Dict[str, str]] = None
-) -> t.Callable[[t.Sequence], E]:
-    def inner_func(args: t.Sequence) -> E:
+) -> t.Callable[[t.List], E]:
```
tobymao, line 399: same

sqlglot/dialects/dialect.py
GeorgeSittas: Mypy was yelling because e.g. in postgres we have:

```python
args[2] = exp.to_interval(step.this)
```

Then, I checked and in all the places where we have `some_func(args: t.Sequence)`, we have `some_func` in the `FUNCTIONS` map of the parser. This is only used in `_parse_function` and we're always passing in a _list_ of args that's returned from `_parse_csv`, so `t.Sequence` was more general than needed in this case and decided to refactor them all in one go for consistency.

sqlglot/dialects/dialect.py
GeorgeSittas: See above

sqlglot/dialects/tsql.py
```diff
@@ -16,6 +16,8 @@
 from sqlglot.time import format_time
 from sqlglot.tokens import TokenType
 
+E = t.TypeVar("E", bound="exp.Expression")
```
tobymao, line 19: we have this in a couple places, can we move this to like typing or something

GeorgeSittas: Yeah I can do that

GeorgeSittas: Done

PR #1693, opened 2023-05-25
Feat!(snowflake): normalize identifiers as uppercase

GeorgeSittas:
cc: @tobymao


GeorgeSittas:
FYI an optimizer test that was producing invalid SQL (for reference, from the "revert" PR):

```sql
-- input query
with cte1("id", foo) as (select 1, 2) select "id" from cte1;

-- optimized results

-- before breaking snowflake commit
WITH "cte1" AS (SELECT 1 AS "id") SELECT "cte1"."id" AS "id" FROM "cte1";

-- after breaking snowflake commit (invalid query)
WITH "CTE1" AS (SELECT 1 AS id) SELECT "CTE1"."id" AS "id" FROM "CTE1";
```

tobymao:
the issue with this is in qualify_outputs, the quote needs to be maintained

        if aliased_column:
            selection.set("alias", exp.to_identifier(aliased_column))

GeorgeSittas:
Closing this in favor of #1697

PR #1695, opened 2023-05-25
Fix(clickhouse): allow aliases in tuple function arguments

### LINKED ISSUE #1690: Valid ClickHouse expression fails to parse
**Before you file an issue**
- Make sure you specify the "read" dialect eg. parse_one(sql, read="spark"): **See the code snippet below**
- Check if the issue still exists on main: **Tried against `main`, still exists**

Successful clickhouse fiddle query:
https://fiddle.clickhouse.com/604acdad-bbe4-4226-97ae-5e825aa79db2

Failing sqlglot snippet:

```
In [3]: import sqlglot as sg

In [4]: sg.__version__
Out[4]: '14.0.0'

In [5]: sg.parse_one("SELECT CAST(tuple(1 as `a`, 2 as `b`, 3.0 as `c`).2 AS Nullable(
   ...: String))", read="clickhouse")
```

**Official Documentation**

I'm not sure what "official SQL documentation" means 😅.

GeorgeSittas:
Fixes #1690

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -678,6 +678,8 @@ class Parser(metaclass=_Parser):
         ),
     }
 
+    FUNCTIONS_WITH_ALIASED_ARGS = {"STRUCT"}
+
     FUNCTION_PARSERS: t.Dict[str, t.Callable] = {
         "CAST": lambda self: self._parse_cast(self.STRICT_CAST),
         "CONVERT": lambda self: self._parse_convert(self.STRICT_CAST),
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -3119,7 +3120,9 @@ def _parse_function(
                 functions = self.FUNCTIONS
 
             function = functions.get(upper)
-            args = self._parse_csv(self._parse_lambda)
+            args = self._parse_csv(
```
tobymao, line 3123: move the check out of the csv

PR #1697, opened 2023-05-26
Feat!(snowflake): normalize identifiers as uppercase

tobymao:


GeorgeSittas (commented):
Very nice! Left a few comments.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -416,6 +416,16 @@ def _qualify_outputs(scope):
     scope.expression.set("expressions", new_selections)
 
 
+def _normalize_identifiers(expression: exp.Expression, dialect: DialectType) -> exp.Expression:
+    if isinstance(expression, exp.Identifier):
```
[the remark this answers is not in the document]
tobymao: nope, we don't preserve quotes anymore

sqlglot/optimizer/normalize_identifiers.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns:
+        The transformed expression.
+    """
+    # We need to leave the output aliases unchanged, so the selects need special handling
+    _normalize_selects(expression, dialect=dialect)
+
+    # These clauses can reference output aliases and also need special handling
+    _normalize_order(expression, dialect=dialect)
+    _normalize_having(expression, dialect=dialect)
```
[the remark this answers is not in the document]
tobymao: i got rid of this

sqlglot/optimizer/normalize_identifiers.py
```diff
@@ -0,0 +1,37 @@
+from sqlglot import exp
+from sqlglot.dialects.dialect import RESOLVES_IDENTIFIERS_AS_UPPERCASE, DialectType
+
+
+def normalize_identifiers(
+    expression: exp.Expression, dialect: DialectType = None
+) -> exp.Expression:
+    """
+    Normalize all unquoted identifiers to either lower or upper case, depending on
```
tobymao, line 9: cc @barakalon i've basically deleted this whole file

sqlglot/optimizer/qualify_columns.py
```diff
@@ -416,6 +414,21 @@ def _qualify_outputs(scope):
...
+def quote_identifiers(
+    expression: exp.Expression, dialect: DialectType, identify: str | bool
+) -> exp.Expression:
+    """Makes sure all identifiers that need to be quoted are quoted."""
+    if isinstance(expression, exp.Identifier):
+        name = expression.this
+        expression.set(
+            "quoted",
+            not exp.SAFE_IDENTIFIER_RE.match(name)
+            or case_sensitive(name, dialect=dialect)
```
GeorgeSittas, line 426: We can't quote if an identifier looks case-sensitive right? It might break semantics.

```suggestion
            or not case_sensitive(name, dialect=dialect)
```

tobymao: no, because of normalize_identifiers, we know that everything that has upper case stuff needs to be quoted. this code is correct.

GeorgeSittas (approved):

sqlglot/optimizer/qualify_columns.py
GeorgeSittas:
```python
case_sensitive(name, dialect=dialect)
or should_identify(name, identify, dialect=dialect),
```
is equivalent to
```python
case_sensitive(name, dialect=dialect)
or not case_sensitive(name, dialect=dialect),
```
when identify == "safe" . Doesn’t seem to be consistent, right? Am I missing something? Also, what about:

```python
>>> qualify(sqlglot.parse_one('select 1 as "a"'), quote_identifiers=True).sql()
'SELECT 1 AS a'
>>> qualify(sqlglot.parse_one('select 1 as a'), quote_identifiers=True).sql()
'SELECT 1 AS a'
```

Don’t we want the following to return a quoted `a` here, since it’s safe to do so?

tobymao: well it’s not the same when identify is not safe

for your example; you need to pass in identify =safe

GeorgeSittas: Got it, thanks for pointing it out, I wrongly assumed that `identify` was initialized to `"safe"` in `qualify`. All good.

sqlglot/helper.py
```diff
@@ -419,18 +420,29 @@ def first(it: t.Iterable[T]) -> T:
     return next(i for i in it)
 
 
-def should_identify(text: str, identify: str | bool) -> bool:
+def case_sensitive(text: str, dialect: DialectType) -> bool:
+    """Checks if text contains any case sensitive charecters depending on dialect."""
```
GeorgeSittas, line 424: Nit: typo
```suggestion
    """Checks if text contains any case sensitive characters depending on dialect."""
```

GeorgeSittas (approved):

PR #1701, opened 2023-05-29
Fix: set quote_identifiers in qualify, add normalize flag in schema

GeorgeSittas:
I guess I could've split this PR into two separate ones, as it introduces both a fix and a feature, but anyway.

1) Activated the `quote_identifiers` flag in the `qualify` rule by default, because this case is currently failing:

```python
>>> optimize("select * from tbl", dialect="snowflake", schema={"tbl": {'"a"': "int"}}).sql()
  File "sqlglot/sqlglot/optimizer/optimizer.py", line 90, in optimize
    expression = rule(expression, **rule_kwargs)
  File "sqlglot/sqlglot/optimizer/annotate_types.py", line 31, in annotate_types
    return TypeAnnotator(schema, annotators, coerces_to).annotate(expression)
  File "sqlglot/sqlglot/optimizer/annotate_types.py", line 296, in annotate
    col.type = self.schema.get_column_type(source, col)
  File "sqlglot/sqlglot/schema.py", line 289, in get_column_type
    raise SchemaError(f"Unknown column type '{column_type}'")
sqlglot.errors.SchemaError: Unknown column type 'None'
```

2) Added a new flag in `MappingSchema` to control whether or not the normalization logic will kick off. This might be useful when e.g. we want to add (unquoted) names that are already normalized, and a 2nd normalization pass would mess them up.

3) Added, improved some type hints and comments.


sqlglot/dataframe/sql/dataframe.py
```diff
@@ -299,7 +299,9 @@ def sql(self, dialect="spark", optimize=True, **kwargs) -> t.List[str]:
         for expression_type, select_expression in select_expressions:
             select_expression = select_expression.transform(replace_id_value, replacement_mapping)
             if optimize:
-                select_expression = optimize_func(select_expression, identify="always")
+                select_expression = t.cast(
+                    exp.Select, optimize_func(select_expression, identify="always")
```
tobymao, line 303: identify=True

tobymao:
did you add a test for this?

GeorgeSittas:
> did you add a test for this?

Yeah, I used an existing one you'd wrote: https://github.com/tobymao/sqlglot/pull/1701/files#diff-739b55fbe772c7434981d6a9f1f6d0dfa499468cb5251b26c588589fb66108cdR713-R716

tobymao:
what's the case for feature 2? when did you encounter a need for it?

why did annotate types fail?

GeorgeSittas:
> why did annotate types fail?

```python
>>> optimize("select * from tbl", dialect="snowflake", schema={"tbl": {'"a"': "int"}}).sql()
> sqlglot/optimizer/annotate_types.py(297)annotate()
-> col.type = self.schema.get_column_type(source, col)
(Pdb) col
(COLUMN this:
  (IDENTIFIER this: a, quoted: False), table:
  (IDENTIFIER this: TBL, quoted: False))
```

Notice how the column `a` is unquoted here (due to `quoted_identifiers=False` in `qualify` and the fact that `annotate_types` runs before `canonicalize`). Passing this in `get_column_type` will result in a failed lookup, because `a` is normalized into `A` and hence will not match what we have in the schema anymore:

```python
-> column_type = table_schema.get(normalized_column_name)
(Pdb) normalized_column_name
'A'
(Pdb) self.mapping
{'TBL': {'a': 'int'}}
```

sqlglot/dataframe/sql/dataframe.py
GeorgeSittas: removed the flag altogether since `identify=True` is the default for both `qualify` and `canonicalize`.

tobymao:
this is problematic. sqlmesh doesn’t quote things which means annotate types would fail. could we fix annotate types to work without quotes?

GeorgeSittas:
So the issue is not actually related directly to `annotate_types` -- it's related to the schema's `get_column_type` method, because the following `raise` is reached:

```python
            column_type = table_schema.get(normalized_column_name)

            if isinstance(column_type, exp.DataType):
                return column_type
            elif isinstance(column_type, str):
                return self._to_data_type(column_type.upper(), dialect=dialect)

            raise SchemaError(f"Unknown column type '{column_type}'")
```

> sqlmesh doesn’t quote things which means annotate types would fail. could we fix annotate types to work without quotes?

Why can't we add quotes over there as well?

> maybe the problem is the double normalize?

You mean in both `normalize_identifiers` and the schema? If that's what you mean then yeah, that's the problem here, but I'm not sure yet if it's correct to remove the normalization logic for `get_column_type`.

tobymao:
now that qualify normalizes everything. schema shouldn’t need to renormalize

GeorgeSittas:
Ok sounds interesting, I'll explore this idea a bit more tomorrow and see if we can get rid of the schema normalization logic altogether.

tobymao:
or else just move identify before annotate types.

GeorgeSittas:
> now that qualify normalizes everything. schema shouldn’t need to renormalize

So I'm not sure about this one: a user may supply whatever names they want for the schema, and they may not necessarily match the normalized names `qualify` produces, so the lookups would fail. We need to keep the schema names and the ones produces by the optimizer in sync.

sqlglot/optimizer/qualify.py
```diff
@@ -59,19 +68,17 @@ def qualify(
...
-    )
+    if qualify_columns:
+        expression = qualify_columns_func(
+            expression, schema, expand_alias_refs=expand_alias_refs, infer_schema=infer_schema
+        )
 
     if quote_identifiers:
-        expression = expression.transform(
-            qualify_columns.quote_identifiers, dialect, identify, copy=False
+        expression = quote_identifiers_func(
```
tobymao, line 77: switch this back, to qualify_columns.quote_identefiers

tests/fixtures/optimizer/canonicalize.sql
```diff
@@ -10,8 +10,8 @@ SELECT CAST(1 AS VARCHAR) AS "a" FROM "w" AS "w";
 SELECT CAST(1 + 3.2 AS DOUBLE) AS a FROM w AS w;
 SELECT 1 + 3.2 AS "a" FROM "w" AS "w";
 
-SELECT CAST("2022-01-01" AS DATE) + INTERVAL '1' day;
-SELECT CAST("2022-01-01" AS DATE) + INTERVAL '1' day AS "_col_0";
```
tobymao, line 14: was this a bug?

GeorgeSittas: Yes, `"2022-01-01"` is an identifier and it can't be resolved in this context.

PR #1724, opened 2023-06-05
clickhouse: CREATE DICTIONARY support

pkit:
added all the vast Dictionary spec from [here](https://clickhouse.com/docs/en/sql-reference/dictionaries)

- bonus: fixed the unwrapped `PRIMARY KEY ident1, ident2, ident3` syntax
- bonus: fixed a typo

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -36,6 +36,7 @@ class ClickHouse(Dialect):
             "ASOF": TokenType.ASOF,
             "ATTACH": TokenType.COMMAND,
             "DATETIME64": TokenType.DATETIME64,
+            "DICTIONARY": TokenType.DICTIONARY,
             "FINAL": TokenType.FINAL,
             "FLOAT32": TokenType.FLOAT,
             "FLOAT64": TokenType.DOUBLE,
@@ -245,6 +246,13 @@ class ClickHouse(Dialect):
             self, optional: bool = False
         ) -> t.List[t.Optional[exp.Expression]]:
             return super()._parse_wrapped_id_vars(optional=True)
+
+        def _parse_primary_key(
+            self, wrapped_optional: bool = False, in_props: bool = False
+        ) -> exp.Expression:
+            if in_props:
+                return super()._parse_primary_key(wrapped_optional=True, in_props=True)
+            return super()._parse_primary_key(wrapped_optional=wrapped_optional, in_props=in_props)
 
     class Generator(generator.Generator):
         STRUCT_DELIMITER = ("(", ")")
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -188,6 +188,8 @@ class Generator:
         exp.Cluster: exp.Properties.Location.POST_SCHEMA,
         exp.DataBlocksizeProperty: exp.Properties.Location.POST_NAME,
         exp.DefinerProperty: exp.Properties.Location.POST_CREATE,
+        exp.DictRange: exp.Properties.Location.POST_SCHEMA,
+        exp.DictProperty: exp.Properties.Location.POST_SCHEMA,
         exp.DistKeyProperty: exp.Properties.Location.POST_SCHEMA,
         exp.DistStyleProperty: exp.Properties.Location.POST_SCHEMA,
         exp.EngineProperty: exp.Properties.Location.POST_SCHEMA,
@@ -1082,7 +1084,7 @@ class Generator:
 
     def lockingproperty_sql(self, expression: exp.LockingProperty) -> str:
         kind = expression.args.get("kind")
-        this: str = f" {this}" if expression.this else ""
+        this: str = f" {expression.this}" if expression.this else ""
         for_or_in = expression.args.get("for_or_in")
         lock_type = expression.args.get("lock_type")
         override = " OVERRIDE" if expression.args.get("override") else ""
@@ -2321,6 +2323,33 @@ class Generator:
             self.unsupported("Format argument unsupported for TO_CHAR/TO_VARCHAR function")
 
         return self.sql(exp.cast(expression.this, "text"))
+
+    def dictproperty_sql(self, expression: exp.DictProperty) -> str:
+        this = self.sql(expression, "this")
+        kind = self.sql(expression, "kind")
+        settings = []
+        for s in expression.args["settings"]:
+            key = self.sql(s, "this")
+            value = self.sql(s, "value")
+            settings.append(f"{key} {value}")
+        if self.pretty and settings:
+            settings_sql = self.indent(
+                self.sep().join([""] + settings + [""]), skip_first=True, skip_last=True
+            )
+        else:
+            settings_sql = self.sep().join(settings)
+        return f"{this}({kind}({settings_sql}))"
+
+    def dictrange_sql(self, expression: exp.DictRange) -> str:
+        this = self.sql(expression, "this")
+        max = expression.args.get("max")
+        min = expression.args.get("min")
+        min_value = self.sql(min, "this")
+        if min == max:
+            return f"{this}({min_value})"
+        else:
+            max_value = self.sql(max, "this")
+            return f"{this}(MIN {min_value} MAX {max_value})"
 
 
 def cached_generator(
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -205,6 +205,7 @@ class Parser(metaclass=_Parser):
         TokenType.SCHEMA,
         TokenType.TABLE,
         TokenType.VIEW,
+        TokenType.DICTIONARY,
     }
 
     CREATABLES = {
@@ -592,6 +593,8 @@ class Parser(metaclass=_Parser):
         ),
         "JOURNAL": lambda self, **kwargs: self._parse_journal(**kwargs),
         "LANGUAGE": lambda self: self._parse_property_assignment(exp.LanguageProperty),
+        "LAYOUT": lambda self: self._parse_dict_property(this="LAYOUT"),
+        "LIFETIME": lambda self: self._parse_dict_range(this="LIFETIME"),
         "LIKE": lambda self: self._parse_create_like(),
         "LOCATION": lambda self: self._parse_property_assignment(exp.LocationProperty),
         "LOCK": lambda self: self._parse_locking(),
@@ -606,7 +609,8 @@ class Parser(metaclass=_Parser):
         "PARTITION BY": lambda self: self._parse_partitioned_by(),
         "PARTITIONED BY": lambda self: self._parse_partitioned_by(),
         "PARTITIONED_BY": lambda self: self._parse_partitioned_by(),
-        "PRIMARY KEY": lambda self: self._parse_primary_key(),
+        "PRIMARY KEY": lambda self: self._parse_primary_key(in_props=True),
+        "RANGE": lambda self: self._parse_dict_range(this="RANGE"),
         "RETURNS": lambda self: self._parse_returns(),
         "ROW": lambda self: self._parse_row(),
         "ROW_FORMAT": lambda self: self._parse_property_assignment(exp.RowFormatProperty),
@@ -615,6 +619,7 @@ class Parser(metaclass=_Parser):
             exp.SettingsProperty, expressions=self._parse_csv(self._parse_set_item)
         ),
         "SORTKEY": lambda self: self._parse_sortkey(),
+        "SOURCE": lambda self: self._parse_dict_property(this="SOURCE"),
         "STABLE": lambda self: self.expression(
             exp.StabilityProperty, this=exp.Literal.string("STABLE")
         ),
@@ -3495,16 +3500,18 @@ class Parser(metaclass=_Parser):
             exp.ForeignKey, expressions=expressions, reference=reference, **options  # type: ignore
         )
 
-    def _parse_primary_key(self) -> exp.Expression:
+    def _parse_primary_key(
+        self, wrapped_optional: bool = False, in_props: bool = False
+    ) -> exp.Expression:
         desc = (
             self._match_set((TokenType.ASC, TokenType.DESC))
             and self._prev.token_type == TokenType.DESC
         )
 
-        if not self._match(TokenType.L_PAREN, advance=False):
+        if not in_props and not self._match(TokenType.L_PAREN, advance=False):
             return self.expression(exp.PrimaryKeyColumnConstraint, desc=desc)
 
-        expressions = self._parse_wrapped_csv(self._parse_field)
+        expressions = self._parse_wrapped_csv(self._parse_field, optional=wrapped_optional)
         options = self._parse_key_constraint_options()
         return self.expression(exp.PrimaryKey, expressions=expressions, options=options)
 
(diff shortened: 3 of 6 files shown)


sqlglot/generator.py
```diff
@@ -2322,6 +2324,33 @@ def tochar_sql(self, expression: exp.ToChar) -> str:
 
         return self.sql(exp.cast(expression.this, "text"))
 
+    def dictproperty_sql(self, expression: exp.DictProperty) -> str:
+        this = self.sql(expression, "this")
+        kind = self.sql(expression, "kind")
+        settings = []
+        for s in expression.args["settings"]:
```
tobymao, line 2331: settings is not required in your spec, so this could be not set, do expression.args("settings") or []:

tobymao: but givven how you're using it, you should really do

self.expressions(expression, key="settings") and make sure the key value is generated correctly

sqlglot/generator.py
```diff
@@ -2322,6 +2324,33 @@ def tochar_sql(self, expression: exp.ToChar) -> str:
...
+        if self.pretty and settings:
+            settings_sql = self.indent(
+                self.sep().join([""] + settings + [""]), skip_first=True, skip_last=True
+            )
+        else:
+            settings_sql = self.sep().join(settings)
+        return f"{this}({kind}({settings_sql}))"
+
+    def dictrange_sql(self, expression: exp.DictRange) -> str:
+        this = self.sql(expression, "this")
```
tobymao, line 2344: what are you trying to do here?

sqlglot/parser.py
```diff
@@ -205,6 +205,7 @@ class Parser(metaclass=_Parser):
         TokenType.SCHEMA,
         TokenType.TABLE,
         TokenType.VIEW,
+        TokenType.DICTIONARY,
```
tobymao, line 208: you need to add this to id_var_tokens

sqlglot/parser.py
```diff
@@ -4484,6 +4491,47 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+        settings = []
+
+        self._match(TokenType.L_PAREN)
+        kind = self._parse_id_var()
+        if not kind:
+            self.raise_error(f"Unexpected token: {self._curr.text}")
+        assert kind
+        wrapped = self._match(TokenType.L_PAREN)
+
+        if wrapped:
```
tobymao, line 4504: if self._match(TokenType.L_PAREN)

sqlglot/generator.py
```diff
@@ -1082,7 +1084,7 @@ def isolatedloadingproperty_sql(self, expression: exp.IsolatedLoadingProperty) -
 
     def lockingproperty_sql(self, expression: exp.LockingProperty) -> str:
         kind = expression.args.get("kind")
-        this: str = f" {this}" if expression.this else ""
+        this: str = f" {expression.this}" if expression.this else ""
```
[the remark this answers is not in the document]
pkit: That's your code, just fixed the typo. :)

sqlglot/parser.py
```diff
@@ -4484,6 +4491,47 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+
+        return self.expression(
+            exp.DictProperty,
+            this=this,
+            kind=kind.this,
+            settings=settings,
+        )
+
+    def _parse_dict_range(self, this: str) -> exp.DictRange:
+        self._match(TokenType.L_PAREN)
```
tobymao, line 4524: these parenthesis are all optional?

sqlglot/parser.py
```diff
@@ -4484,6 +4491,47 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+        has_min = self._match_text_seq("MIN")
+        min = self._parse_var() or self._parse_primary()
+        if has_min:
+            self._match_text_seq("MAX")
+            max = self._parse_var() or self._parse_primary()
+        else:
+            max = min
+        self._match(TokenType.R_PAREN)
```
tobymao, line 4532: you should only match r paren if you matched an l paren

sqlglot/parser.py
pkit: Ah, yeah, missed this one after refactoring!

sqlglot/parser.py
pkit: Yup

pkit: Ah, sorry, these ones are not really

sqlglot/generator.py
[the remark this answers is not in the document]
GeorgeSittas: Yep, thanks for catching it, this must've slipped in a PR. Just pointing it out to make sure we fix it correctly.

sqlglot/parser.py
```diff
@@ -592,6 +593,8 @@ class Parser(metaclass=_Parser):
         ),
         "JOURNAL": lambda self, **kwargs: self._parse_journal(**kwargs),
         "LANGUAGE": lambda self: self._parse_property_assignment(exp.LanguageProperty),
+        "LAYOUT": lambda self: self._parse_dict_property(this="LAYOUT"),
```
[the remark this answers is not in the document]
tobymao: it's ok for now since there's a bunch of teradata and crap here to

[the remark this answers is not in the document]
tobymao: when there's a conflict with can move it  out

sqlglot/parser.py
tobymao: then switch self._match_l_paren()

sqlglot/parser.py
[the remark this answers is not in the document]
pkit: Dunno, all the earlier ones are here.
Should I?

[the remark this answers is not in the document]
tobymao: leave it it's fine

sqlglot/parser.py
```diff
@@ -606,7 +609,8 @@ class Parser(metaclass=_Parser):
         "PARTITION BY": lambda self: self._parse_partitioned_by(),
         "PARTITIONED BY": lambda self: self._parse_partitioned_by(),
         "PARTITIONED_BY": lambda self: self._parse_partitioned_by(),
-        "PRIMARY KEY": lambda self: self._parse_primary_key(),
+        "PRIMARY KEY": lambda self: self._parse_primary_key(in_props=True),
```
tobymao, line 612: is this override safe? or is this clickhouse specific

sqlglot/parser.py
```diff
@@ -4484,6 +4491,47 @@ def _parse_as_command(self, start: Token) -> exp.Command:
         size = len(start.text)
         return exp.Command(this=text[:size], expression=text[size:])
 
+    def _parse_dict_property(self, this: str) -> exp.DictProperty:
+        settings = []
+
+        self._match(TokenType.L_PAREN)
```
tobymao, line 4497: is this optional?

sqlglot/parser.py
```diff
@@ -4484,6 +4491,47 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+        kind = self._parse_id_var()
+        if not kind:
+            self.raise_error(f"Unexpected token: {self._curr.text}")
```
tobymao, line 4500: self._curr could be none and you'll error out here

sqlglot/generator.py
pkit: I'm not sure how to use it with `self.expressions()` as these are `key value` and not `key=value`

sqlglot/parser.py
pkit: Is it `None` on EOF?

sqlglot/parser.py
pkit: AFAIK `PRIMARY KEY` in table properties is CH-specific

sqlglot/parser.py
tobymao: it's None, which means eof

sqlglot/generator.py
pkit: @tobymao unfortunately only the dictionary `kind` properties are `key value` in CH
So these are pretty specific. I can probably subclass `exp.Property` for that, but is it worth it?

sqlglot/dialects/clickhouse.py
```diff
@@ -246,6 +247,13 @@ def _parse_wrapped_id_vars(
         ) -> t.List[t.Optional[exp.Expression]]:
             return super()._parse_wrapped_id_vars(optional=True)
 
+        def _parse_primary_key(
+            self, wrapped_optional: bool = False, in_props: bool = False
+        ) -> exp.Expression:
+            if in_props:
```
tobymao, line 253: return super().__(wrapped_optional=wrapped_optional or in_props, in_props=in_props)

sqlglot/parser.py
pkit: Added

sqlglot/parser.py
pkit: Done

sqlglot/parser.py
pkit: Done

sqlglot/parser.py
pkit: Fixed

sqlglot/parser.py
tobymao: what about teradata?

sqlglot/generator.py
tobymao: yes let's do that

pkit: Done!

sqlglot/parser.py
pkit: Dunno, do you have tests for it? Or docs? Will check it out.

tobymao: https://www.mysqltutorial.org/mysql-primary-key/

what about mysql?

sqlglot/generator.py
```diff
@@ -2322,6 +2324,27 @@ def tochar_sql(self, expression: exp.ToChar) -> str:
...
+    def dictproperty_sql(self, expression: exp.DictProperty) -> str:
+        this = self.sql(expression, "this")
+        kind = self.sql(expression, "kind")
+        settings_sql = self.expressions(expression, key="settings", sep=" ")
+        args = f"({self.sep('')}{settings_sql}{self.seg(')', sep='')}" if settings_sql else "()"
+        return f"{this}({kind}{args})"
+
+    def dictrange_sql(self, expression: exp.DictRange) -> str:
+        this = self.sql(expression, "this")
+        max = expression.args.get("max")
```
tobymao, line 2336: why get min and max? can you jsut do

min = self.sql(expression, "min")
max = self.sql(expression, "max")

sqlglot/parser.py
```diff
@@ -4484,6 +4492,46 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
         return exp.Command(this=text[:size], expression=text[size:])
 
+    def _parse_dict_property(self, this: str) -> exp.DictProperty:
+        settings = []
+
+        self._match_l_paren()
+        kind = self._parse_id_var()
+        if not kind:
+            self.raise_error(f"Unexpected token")
+        assert kind
```
tobymao, line 4502: let's remove this assert and make 4519 kind.this more generous, this allows for partials

sqlglot/parser.py
```diff
@@ -4484,6 +4492,46 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+                    break
+                settings.append(self.expression(exp.DictSubProperty, this=key, value=value))
+            self._match(TokenType.R_PAREN)
+
+        self._match_r_paren()
+
+        return self.expression(
+            exp.DictProperty,
+            this=this,
+            kind=kind.this,
```
tobymao, line 4519: kind.this if kind else None

sqlglot/parser.py
```diff
@@ -4484,6 +4492,46 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+
+    def _parse_dict_range(self, this: str) -> exp.DictRange:
+        self._match_l_paren()
+        has_min = self._match_text_seq("MIN")
+        min = self._parse_var() or self._parse_primary()
+        if has_min:
+            self._match_text_seq("MAX")
+            max = self._parse_var() or self._parse_primary()
+        else:
+            max = min
```
tobymao, line 4531: why do you need to set max to min here? why not just leave it None?

sqlglot/parser.py
pkit: Not in props, I only changed `PRIMARY KEY` parsing in the "after schema" part.
Mysql is still "in the schema".

sqlglot/generator.py
pkit: I'm not sure it's good to compare the string representation here.

sqlglot/parser.py
pkit: I'm not sure about it as these are for dictionaries, and AFAIK dicts in CH do not work at all when the "kind" is unset.
See:
```
$ echo "create dictionary d (is UInt64, data String) SOURCE()" | clickhouse-format 
Code: 62. DB::Exception: Syntax error (query): failed at position 53 (')'): )
. Expected one of: function with key-value arguments, identifier. (SYNTAX_ERROR), Stack trace (when copying this message, always include the lines below):

0. ./build_docker/./src/Common/Exception.cpp:92: DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x000000000dcaeaf5 in /usr/bin/clickhouse
1. DB::Exception::createDeprecated(String const&, int, bool) @ 0x0000000008ace3ed in /usr/bin/clickhouse
2. ./build_docker/./src/Parsers/parseQuery.cpp:362: DB::parseQueryAndMovePosition(DB::IParser&, char const*&, char const*, String const&, bool, unsigned long, unsigned long) @ 0x0000000014c64300 in /usr/bin/clickhouse
3. ./build_docker/./programs/format/Format.cpp:163: mainEntryClickHouseFormat(int, char**) @ 0x000000000de998ce in /usr/bin/clickhouse
4. ./build_docker/./programs/main.cpp:0: main @ 0x000000000874616f in /usr/bin/clickhouse
5. ? @ 0x00007fc82b329d90 in ?
6. __libc_start_main @ 0x00007fc82b329e40 in ?
7. _start @ 0x0000000007e8d3ae in /usr/bin/clickhouse
 (version 23.5.1.1345 (official build))
```

sqlglot/parser.py
pkit: Not correct from the logical standpoint. As CH uses `MIN` and `MAX` as a range `[min, max]` and if there is only one number it's `[number, number]` range internally.
For example if you programmatically change `(MIN 10 MAX 20)` with `MAX = 10` it should produce `(10)`

sqlglot/parser.py
pkit: Note: `clickhouse-format` is not a validator, it only checks the syntax. So I would like `sqlglot` to have exactly the same behavior.

sqlglot/parser.py
pkit: Another idea is to use `(MIN n MAX n)` as a canonical representation of `(n)`

sqlglot/parser.py
pkit: Teradata doesn't have `PRIMARY KEY` in props too, only in schema.

sqlglot/generator.py
tobymao: why not?

sqlglot/parser.py
tobymao: it doesn't matter, as i said this is for partials. and sqlglot is not a validator

we want to be able to parse with error mode ignore

create dictionary d(...) SOURCE()

sqlglot/parser.py
tobymao: yea i'm fine with (min n max n) for n

sqlglot/generator.py
pkit: You have created `def __eq__()` for a reason, if everybody would compare only strings, what's the point? :)

pkit: But I can change to string here, as these are pretty simple, just literals.

sqlglot/parser.py
pkit: ah, ok

sqlglot/generator.py
tobymao: right, it's just the literal and should simplify the code

sqlglot/parser.py
tobymao: should we remove the str comparison and do

max = max or min

pkit: Ooof, I was wrong, checked the whole code of how CH builds dictionaries and it's different from what I remembered.
Lifetime has `min_sec = 0` and `max_sec = 0` by default. So, `(n) == (MIN 0 MAX n)`!
Will canonicalize like that!
See [here](https://github.com/ClickHouse/ClickHouse/blob/9f80900d6f587383780d2a40f8173093dce68a5a/src/Parsers/ParserDictionary.cpp#L36)

sqlglot/generator.py
```diff
@@ -1411,7 +1413,7 @@ def lock_sql(self, expression: exp.Lock) -> str:
         return f"{lock_type}{expressions}{wait or ''}"
 
     def literal_sql(self, expression: exp.Literal) -> str:
-        text = expression.this or ""
+        text = expression.this if expression.this is not None else ""
```
pkit, line 1416: This one was converting literal `0` to `""`

tobymao: it shouldn't because this should always be a string, "0", and not the literal 0

sqlglot/parser.py
```diff
@@ -4484,6 +4492,44 @@ def _parse_as_command(self, start: Token) -> exp.Command:
...
+    def _parse_dict_range(self, this: str) -> exp.DictRange:
+        self._match_l_paren()
+        has_min = self._match_text_seq("MIN")
+        if has_min:
+            min = self._parse_var() or self._parse_primary()
+            self._match_text_seq("MAX")
+            max = self._parse_var() or self._parse_primary()
+        else:
+            max = self._parse_var() or self._parse_primary()
+            min = self.expression(exp.Literal, this=0, is_string=False)
```
tobymao, line 4529: min = exp.Literal.number(0)

tobymao (approved):
thanks for the quick iteration, looks good except for the 2 small comments

sqlglot/generator.py
tobymao, line 1416:
```suggestion
        text = expression.this or ""
```

PR #1726, opened 2023-06-06
Feat(presto): transpile 'epoch' to '1970-01-01 00:00:00' in time-like casts

### LINKED ISSUE #1725: cannot correctly convert syntax "timestamp 'epoch'" from redshift to presto/trino
```
sql = """

select CAST(
    DATE_TRUNC( 
    'day', 
    timestamp 'epoch' + creation_time * interval '1 second' 
    ) AS DATE
) AS dt

"""
sqlglot.transpile(sql, read="redshift", write="trino")[0]
```

the output is:
`"SELECT TRY_CAST(DATE_TRUNC('day', CAST('epoch' AS TIMESTAMP) + creation_time * INTERVAL '1' second) AS DATE) AS dt"`

trino would throw the error: `Value cannot be cast to timestamp: epoch`

GeorgeSittas:
Fixes #1725

Haven't investigated much, so I'm not sure if this transformation is 100% correct; for the example in #1725 it works fine.

However, one difference is that Spark produces the following when casting `'epoch'` to a timezone:

```
spark-sql (default)> select cast('epoch' as timestamp);
1970-01-01 02:00:00
```

Whereas this transformation yields the following when transpiling the above from Spark to Trino:

```
trino> SELECT TRY_CAST('1970-01-01 00:00:00' AS TIMESTAMP);
          _col0          
-------------------------
 1970-01-01 00:00:00.000 
(1 row)
```

Not sure why it produces `02:00:00` in the former case.. Maybe it's interpreting the `'epoch'` using a different timezone?

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -268,6 +268,17 @@ def add_recursive_cte_column_names(expression: exp.Expression) -> exp.Expression
     return expression
 
 
+def epoch_cast_to_ts(expression: exp.Expression) -> exp.Expression:
+    if (
+        isinstance(expression, (exp.Cast, exp.TryCast))
+        and expression.name.lower() == "epoch"
+        and expression.to.this in exp.DataType.TEMPORAL_TYPES
+    ):
+        expression.this.replace(exp.Literal.string("1970-01-01 00:00:00"))
+
+    return expression
+
+
 def preprocess(
     transforms: t.List[t.Callable[[exp.Expression], exp.Expression]],
 ) -> t.Callable[[Generator, exp.Expression], str]:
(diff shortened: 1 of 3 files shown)


sqlglot/transforms.py
```diff
@@ -268,6 +268,17 @@ def add_recursive_cte_column_names(expression: exp.Expression) -> exp.Expression
     return expression
 
 
+def epoch_cast_to_ts(expression: exp.Expression) -> exp.Expression:
+    if (
+        isinstance(expression, (exp.Cast, exp.TryCast))
+        and expression.name.lower() == "epoch"
+        and expression.to.this in exp.DataType.TEMPORAL_TYPES
```
GeorgeSittas, line 275: I added this check to ensure correctness in the general case, although maybe we can get rid of it? Not sure what else one would want to cast `'epoch'` into.

PR #1729, opened 2023-06-06
Fix: conditionally quote identifiers that start with a digit

### LINKED ISSUE #1727: Trino aliases cannot start with an integer if unquoted
This is a valid Spark query, but it's not in Trino, I get an error "identifiers must not start with a digit; surround the identifier with double quotes"

```
>>> import sqlglot
>>> sqlglot.__version__
'15.0.0'
>>> sqlglot.transpile("SELECT 1 AS 1x", read='spark', write='trino')[0]
'SELECT 1 AS 1x'
```

The desired output here is then

```
>>> sqlglot.transpile("SELECT 1 AS 1x", read='spark', write='trino')[0]
'SELECT 1 AS "1x"'
```

GeorgeSittas:
Fixes #1727

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -149,6 +149,7 @@ def _to_date_sql(self: generator.Generator, expression: exp.TsOrDsToDate) -> str
 
 class Hive(Dialect):
     alias_post_tablesample = True
+    identifiers_can_start_with_digit = True
 
     time_mapping = {
         "y": "%Y",
@@ -190,7 +191,6 @@ class Hive(Dialect):
         IDENTIFIERS = ["`"]
         STRING_ESCAPES = ["\\"]
         ENCODE = "utf-8"
-        IDENTIFIER_CAN_START_WITH_DIGIT = True
 
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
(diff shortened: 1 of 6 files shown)


sqlglot/parser.py
```diff
@@ -3060,11 +3060,7 @@ def _parse_column_ops(self, this: exp.Expression) -> exp.Expression:
                     else exp.Literal.string(value)
                 )
             else:
-                field = (
-                    self._parse_star()
-                    or self._parse_function(anonymous=True)
-                    or self._parse_id_var()
-                )
+                field = self._parse_field(anonymous_func=True)
```
GeorgeSittas, line 3063: This was changed because in `<something>.1` we used to parse `1` as an identifier, so with this PR it was force-quoted and several tests broke (that I guess wanted `1` to be treated as an index into some nested type instead).

sqlglot/dialects/hive.py
```diff
@@ -283,6 +283,7 @@ class Generator(generator.Generator):
         JOIN_HINTS = False
         TABLE_HINTS = False
         INDEX_ON = "ON TABLE"
+        IDENTIFIER_CAN_START_WITH_DIGIT = True
```
tobymao, line 286: let's move this out to the dialect? because it's in both tokenizer and generator, and then it can get passed in to both

GeorgeSittas: yeah, good point

GeorgeSittas: done

vegarsti:
Thanks!

PR #1730, opened 2023-06-06
Fix!(spark): cast to string instead of (var)char in non DDL expressions

GeorgeSittas:
xref: https://github.com/duneanalytics/harmonizer/pull/70

cc: @vegarsti

### DIFF
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -195,7 +195,7 @@ class Spark2(Hive):
             """
             this = super()._parse_types(check_func=check_func, schema=schema)
 
-            if isinstance(this, exp.DataType) and this.is_type(
+            if not schema and isinstance(this, exp.DataType) and this.is_type(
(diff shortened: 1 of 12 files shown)


GeorgeSittas:
FYI I cross checked with Spark v2.4.8, and the behavior seems to be the same:

```
spark-sql> SELECT CAST(123456 AS VARCHAR(3));
[....]
123456
Time taken: 2.634 seconds, Fetched 1 row(s)
```

I don't have access to a Hive cluster to verify whether this behavior is the same for Hive. If someone can check that, we can also move the changes to Hive.

sqlglot/dialects/spark2.py
```diff
@@ -172,6 +172,39 @@ class Parser(Hive.Parser):
             "SHUFFLE_REPLICATE_NL": lambda self: self._parse_join_hint("SHUFFLE_REPLICATE_NL"),
         }
 
+        def _parse_types(
```
tobymao, line 175: this should probably go in hive override

PR #1734, opened 2023-06-06
Fix: ensure pivot can be used as a table name

GeorgeSittas:


sqlglot/parser.py
```diff
@@ -1912,8 +1912,16 @@ def _parse_select(
 
             this = self._parse_query_modifiers(this)
         elif (table or nested) and self._match(TokenType.L_PAREN):
-            this = self._parse_table() if table else self._parse_select(nested=True)
-            this = self._parse_set_operations(self._parse_query_modifiers(this))
+            if self._match(TokenType.PIVOT):
```
tobymao, line 1915: what about select * from (pivot)

GeorgeSittas: this is a syntax error in both postgres and bigquery

GeorgeSittas:
```
georgesittas=# with pivot as (select 1) select * from (pivot);
ERROR:  syntax error at or near ")"
LINE 1: with pivot as (select 1) select * from (pivot);
                                                     ^
```
Screenshot 2023-06-07 at 12 59 39 AM

PR #1738, opened 2023-06-07
Fix!(presto): ensure ||, CONCAT args are strings

### LINKED ISSUE #1735: Can not convert '||' from redshift to presto/trino when the arguments are not string typed
```
sql = """
select 'a' || 1 || 1.1
"""
converted_sql = sqlglot.transpile(sql, read="redshift", write="trino")[0]
print(converted_sql)
```
```
SELECT 'a' || 1 || 1.1

```

Trino would throw error `Unexpected parameters (varchar(1), integer) for function concat. Expected: concat(char(x), char(y)), concat(array(E), E) E, concat(E, array(E)) E, concat(array(E)) E, concat(varchar), concat(varbinary)'` when executing the output SQL, as Trino doesn't covert the type implicitly like Redshift does.

GeorgeSittas:
Fixes #1735

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -8,8 +8,6 @@ from sqlglot.dialects.dialect import (
     inline_array_sql,
     no_pivot_sql,
     rename_func,
-    safeconcat_sql,
-    safedpipe_sql,
     var_map_sql,
 )
 from sqlglot.errors import ParseError
@@ -25,6 +23,7 @@ def _lower_func(sql: str) -> str:
 class ClickHouse(Dialect):
     NORMALIZE_FUNCTIONS: bool | str = False
     NULL_ORDERING = "nulls_are_last"
+    STRICT_STRING_CONCAT = True
 
     class Tokenizer(tokens.Tokenizer):
         COMMENTS = ["--", "#", "#!", ("/*", "*/")]
@@ -107,7 +106,6 @@ class ClickHouse(Dialect):
         }
 
         LOG_DEFAULTS_TO_LN = True
-        STRICT_STRING_CONCAT = True
 
         QUERY_MODIFIER_PARSERS = {
             **parser.Parser.QUERY_MODIFIER_PARSERS,
@@ -308,8 +306,6 @@ class ClickHouse(Dialect):
             exp.Quantile: lambda self, e: self.func("quantile", e.args.get("quantile"))
             + f"({self.sql(e, 'this')})",
             exp.RegexpLike: lambda self, e: f"match({self.format_args(e.this, e.expression)})",
-            exp.SafeConcat: safeconcat_sql,
-            exp.SafeDPipe: safedpipe_sql,
             exp.StrPosition: lambda self, e: f"position({self.format_args(e.this, e.args.get('substr'), e.args.get('position'))})",
             exp.VarMap: lambda self, e: _lower_func(var_map_sql(self, e)),
         }
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -8,6 +8,8 @@ from sqlglot.dialects.dialect import (
     inline_array_sql,
     no_pivot_sql,
     rename_func,
+    safeconcat_sql,
+    safedpipe_sql,
     var_map_sql,
 )
 from sqlglot.errors import ParseError
@@ -105,6 +107,7 @@ class ClickHouse(Dialect):
         }
 
         LOG_DEFAULTS_TO_LN = True
+        STRICT_STRING_CONCAT = True
 
         QUERY_MODIFIER_PARSERS = {
             **parser.Parser.QUERY_MODIFIER_PARSERS,
@@ -305,6 +308,8 @@ class ClickHouse(Dialect):
             exp.Quantile: lambda self, e: self.func("quantile", e.args.get("quantile"))
             + f"({self.sql(e, 'this')})",
             exp.RegexpLike: lambda self, e: f"match({self.format_args(e.this, e.expression)})",
+            exp.SafeConcat: safeconcat_sql,
+            exp.SafeDPipe: safedpipe_sql,
             exp.StrPosition: lambda self, e: f"position({self.format_args(e.this, e.args.get('substr'), e.args.get('position'))})",
             exp.VarMap: lambda self, e: _lower_func(var_map_sql(self, e)),
         }
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -334,6 +334,16 @@ class ClickHouse(Dialect):
             "NAMED COLLECTION",
         }
 
+        def safeconcat_sql(self, expression: exp.SafeConcat) -> str:
+            # Clickhouse errors out if we try to cast a NULL value to TEXT
+            return self.func(
+                "CONCAT",
+                *[
+                    exp.func("if", e.is_(exp.null()), e, exp.cast(e, "text"))
+                    for e in expression.expressions
+                ],
+            )
+
         def cte_sql(self, expression: exp.CTE) -> str:
             if isinstance(expression.this, exp.Alias):
                 return self.sql(expression, "this")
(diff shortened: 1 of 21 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -277,6 +277,14 @@ def inline_array_sql(self: Generator, expression: exp.Array) -> str:
     return f"[{self.expressions(expression)}]"
 
 
+def safeconcat_sql(self: Generator, expression: exp.SafeConcat) -> str:
+    return self.func("CONCAT", *[exp.cast(e, "text") for e in expression.expressions])
+
+
+def safedpipe_sql(self: Generator, expression: exp.SafeDPipe) -> str:
+    return self.func("CONCAT", *[exp.cast(e, "text") for e in expression.flatten()])
```
GeorgeSittas, line 285: `CONCAT` is equivalent to `||` for presto, trino, clickhouse which are the dialects of interest that need this, so I transpiled the latter to the former because it was easier compared to flattening the `||`.

sqlglot/executor/env.py
```diff
@@ -151,6 +151,7 @@ def interval(this, unit):
     "CAST": cast,
     "COALESCE": lambda *args: next((a for a in args if a is not None), None),
     "CONCAT": null_if_any(lambda *args: "".join(args)),
+    "SAFECONCAT": null_if_any(lambda *args: "".join(args)),
```
tobymao, line 154: this needs to convert all args to strings

GeorgeSittas: ah missed it, thanks

sqlglot/dialects/clickhouse.py
```diff
@@ -8,6 +8,8 @@
     inline_array_sql,
     no_pivot_sql,
     rename_func,
+    safeconcat_sql,
```
tobymao, line 11: this always goes hand in hand, should we do a dialect flag instead?

sqlglot/generator.py
```diff
@@ -1770,6 +1770,11 @@ def concat_sql(self, expression: exp.Concat) -> str:
             return self.sql(expression.expressions[0])
         return self.function_fallback_sql(expression)
 
+    def safeconcat_sql(self, expression: exp.SafeConcat) -> str:
+        if len(expression.expressions) == 1:
```
tobymao: why would len ever be 1 ? what’s the point of this check

sqlglot/dialects/clickhouse.py
tobymao: refactor strict cast as well if it’s rasy

sqlglot/generator.py
GeorgeSittas: I changed this to delegate to `concat_sql`; the point of this check is to reduce something like `CONCAT('a')` to `'a'` (the former is valid MySQL, for example).

tobymao: why bother? that's a job for the optimizer

sqlglot/parser.py
```diff
@@ -2852,11 +2857,11 @@ def _parse_bitwise(self) -> t.Optional[exp.Expression]:
 
         while True:
             if self._match_set(self.BITWISE):
-                this = self.expression(
-                    self.BITWISE[self._prev.token_type],
-                    this=this,
-                    expression=self._parse_term(),
-                )
+                expr_type = self.BITWISE[self._prev.token_type]
```
tobymao, line 2860: this change isn't necessary, you can just get change the bitwise mapping right?

GeorgeSittas: Ah right, good call thanks!

sqlglot/generator.py
GeorgeSittas: it helps transpile `concat('a')` e.g. to presto/trino correctly, because it requires >1 arguments. There might be other dialects that have these semantics too, so it doesn't hurt to keep this imo.

```
trino> select concat('a');
Query 20230607_113101_00000_4ki82 failed: There must be two or more concatenation arguments
```

sqlglot/dialects/clickhouse.py
GeorgeSittas: Refactored this into a dialect flag called `strict_string_concat`

sqlglot/optimizer/canonicalize.py
```diff
@@ -26,7 +26,7 @@ def canonicalize(expression: exp.Expression) -> exp.Expression:
 
 def add_text_to_concat(node: exp.Expression) -> exp.Expression:
     if isinstance(node, exp.Add) and node.type and node.type.this in exp.DataType.TEXT_TYPES:
-        node = exp.Concat(this=node.this, expression=node.expression)
+        node = exp.Concat(expressions=[node.left, node.right])
```
GeorgeSittas, line 29: This looks like a bug, `CONCAT` didn't have `this` and `expression` as args, I wonder how it used to work.

sqlglot/dialects/dialect.py
```diff
@@ -107,6 +107,7 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
         klass.tokenizer_class.identifiers_can_start_with_digit = (
             klass.identifiers_can_start_with_digit
         )
+        klass.parser_class.strict_string_concat = klass.strict_string_concat
```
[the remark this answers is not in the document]
tobymao: yea let's refactor it later

sqlglot/optimizer/canonicalize.py
GeorgeSittas: I guess we didn't error out before because `concat_sql` relies on `function_fallback_sql` under the hood and so it'd pick up args not defined in `arg_types`:

```python
    def function_fallback_sql(self, expression: exp.Func) -> str:
        args = []
        for arg_value in expression.args.values():
            if isinstance(arg_value, list):
                for value in arg_value:
                    args.append(value)
            else:
                args.append(arg_value)

        return self.func(expression.sql_name(), *args)
```

tests/dataframe/unit/test_functions.py
```diff
@@ -1278,7 +1278,7 @@ def test_concat(self):
         col = SF.concat(SF.col("cola"), SF.col("colb"))
         self.assertEqual("CONCAT(cola, colb)", col.sql())
         col_single = SF.concat("cola")
-        self.assertEqual("cola", col_single.sql())
+        self.assertEqual("CONCAT(cola)", col_single.sql())
```
GeorgeSittas, line 1281: This isn't reduced to `cola` because `SF` functions instantiate the expressions manually instead of parsing them, so since the generator now doesn't override `concat_sql` to do the `len(..) == 1` check, it outputs it as is. Not sure if this is ok or not.

GeorgeSittas:
Waiting on https://github.com/tobymao/sqlglot/pull/1750 to get merged so I can simplify some parts of this PR.

sqlglot/generator.py
```diff
@@ -2074,6 +2074,11 @@ def intdiv_sql(self, expression: exp.IntDiv) -> str:
     def dpipe_sql(self, expression: exp.DPipe) -> str:
         return self.binary(expression, "||")
 
+    def safedpipe_sql(self, expression: exp.SafeDPipe) -> str:
+        if self.STRICT_STRING_CONCAT:
+            return self.func("CONCAT", *[exp.cast(e, "text") for e in expression.flatten()])
```
tobymao, line 2079: change to gen (

tobymao (approved):

sqlglot/generator.py
GeorgeSittas: is there a difference in the semantics? I thought it'd make no difference since we expand it to match the `*args`

tobymao: one creates a list, the other is an iterator. so you don’t incur the overhead of creating a list

GeorgeSittas: I see, makes sense, thanks for clarifying

GeorgeSittas: Done. FYI I also fixed https://github.com/tobymao/sqlglot/pull/1738/commits/14f2386ca447e5409c31d1d3c8e9ed001aff007b#diff-362498362ae539e216cd83ebfe510977107648742f1a1bee7b2db7b5b7d27974R1747 similarly.

PR #1748, opened 2023-06-08
Refactor(parser): reduce copy-paste in CREATE properties handling

pkit:


sqlglot/parser.py
```diff
@@ -1157,13 +1157,16 @@ def _parse_create(self) -> t.Optional[exp.Expression]:
         begin = None
         clone = None
 
+        def extend_props(temp_props):
```
GeorgeSittas, line 1160: Let's add some type hints here.

GeorgeSittas (approved):
Thanks, you beat me to it. I was actually preparing a refactor that also simplified this, using something like:

```python
        def _update_properties(
            props: t.Optional[exp.Properties], before: t.Optional[bool] = None
        ) -> t.Optional[exp.Properties]:
            new_props = self._parse_properties(before=before)
            if props and new_props:
                props.expressions.extend(new_props.expressions)

            return props or new_props
```

Meaning that the pattern in my code was `properties = _update_properties(properties)`.

sqlglot/parser.py
pkit: Added

GeorgeSittas: Thanks

pkit:
I did it with assignment first, but I was concerned about speed a little.
As it would needlessly reassign. So I've invented this one.

sqlglot/parser.py
GeorgeSittas: FYI the type signature in the snippet I pasted is not consistent with the types in current main branch because I've fixed several type hints. You can use this to fix the error:

```python
def _parse_properties(self, before: t.Optional[bool] = None) -> t.Optional[exp.Properties]:
```

PR #1749, opened 2023-06-08
Feat!(clickhouse): add CREATE ON CLUSTER support

### LINKED ISSUE #1747: clickhouse: ON CLUSTER is not supported
see https://clickhouse.com/docs/en/sql-reference/distributed-ddl

I can implement this one, but I'm not sure about design.
On one hand it's a CH specific feature that has no usage otherwise.
On the other hand it's really similar to Teradata "properties before schema" feature, which is not sub-classed.
Any ideas?
Just add it to the main parser?

pkit:
Fixes #1747

- moved Teradata tests to their specific test suite
- split POST_NAME props into dialects


sqlglot/parser.py
```diff
@@ -1181,7 +1181,7 @@ def extend_props(temp_props: t.Optional[exp.Expression]) -> None:
             table_parts = self._parse_table_parts(schema=True)
 
             # exp.Properties.Location.POST_NAME
-            if self._match(TokenType.COMMA):
+            if self._match(TokenType.COMMA) or self._match(TokenType.ON, advance=False):
```
tobymao, line 1184: don't add this, just remove the if

sqlglot/dialects/clickhouse.py
```diff
@@ -328,3 +352,27 @@ def parameterizedagg_sql(self, expression: exp.Anonymous) -> str:
 
         def placeholder_sql(self, expression: exp.Placeholder) -> str:
             return f"{{{expression.name}: {self.sql(expression, 'kind')}}}"
+
+        def oncluster_sql(self, expression: exp.OnCluster) -> str:
```
tobymao, line 356: you need to ad this to the main generator or else we will error out because we can't handle this expression type

sqlglot/dialects/clickhouse.py
```diff
@@ -252,6 +252,17 @@ def _parse_primary_key(
                 wrapped_optional=wrapped_optional or in_props, in_props=in_props
             )
 
+        def _parse_on_property(self) -> t.Optional[exp.Property]:
```
tobymao, line 255: add this to core parser

sqlglot/dialects/clickhouse.py
```diff
@@ -252,6 +252,17 @@ def _parse_primary_key(
...
+            if self._match_text_seq("CLUSTER"):
+                return self._parse_on_cluster()
+            return None
+
+        def _parse_on_cluster(self) -> t.Optional[exp.Property]:
```
tobymao, line 260: just move this inline for now to on property

sqlglot/dialects/clickhouse.py
```diff
@@ -252,6 +252,17 @@ def _parse_primary_key(
                 wrapped_optional=wrapped_optional or in_props, in_props=in_props
             )
 
+        def _parse_on_property(self) -> t.Optional[exp.Property]:
+            if self._match_text_seq("CLUSTER"):
+                return self._parse_on_cluster()
```
tobymao, line 257: if this the on property parser doens't return anything, we should probably retreat

sqlglot/dialects/clickhouse.py
```diff
@@ -328,3 +352,27 @@ def parameterizedagg_sql(self, expression: exp.Anonymous) -> str:
...
+                    exp.Properties(expressions=locations[exp.Properties.Location.POST_NAME]),
+                    wrapped=False,
+                )
+                schema_exp = self.expressions(expression.this)
+                if self.pretty:
+                    this_schema = f"{self.sep()}({self.seg(schema_exp)}{self.sep()})"
+                else:
+                    this_schema = f"({schema_exp})"
+                return f"{this_name} {this_properties} {this_schema}"
+            else:
```
tobymao, line 377: else is not needed

sqlglot/dialects/clickhouse.py
```diff
@@ -328,3 +352,27 @@ def parameterizedagg_sql(self, expression: exp.Anonymous) -> str:
...
+            return f"ON CLUSTER {self.sql(expression, 'this')}"
+
+        def createable_sql(
+            self,
+            expression: exp.Create,
+            locations: dict[exp.Properties.Location, list[exp.Property]],
+        ) -> str:
+            kind = self.sql(expression, "kind").upper()
+            if kind in self.ON_CLUSTER_TARGETS and locations.get(exp.Properties.Location.POST_NAME):
+                this_name = self.sql(expression.this, "this")
```
tobymao, line 366: was this really necessary? seems like a lot of copy and paste? can you just do this with flags in the generator?

sqlglot/parser.py
pkit: I think it will always raise in the property parser then:
```
        if self._match_texts(self.PROPERTY_PARSERS):
            parser = self.PROPERTY_PARSERS[self._prev.text.upper()]
            try:
                return parser(self, **{k: v for k, v in kwargs.items() if v})
            except TypeError:
                self.raise_error(f"Cannot parse property '{self._prev.text}'")
```

sqlglot/dialects/clickhouse.py
pkit: When assembled programmatically?
Interesting. But isn't that's what `write=` is for?

sqlglot/dialects/clickhouse.py
pkit: Just `return super()` ? Seemed less clear to me, but ok

sqlglot/dialects/clickhouse.py
pkit: Handling `pretty=True` should better be uniform, I think.

sqlglot/parser.py
tobymao: just do 

self._match(TokenType.COMMA)

i removed this in my local and it worked fine

sqlglot/dialects/clickhouse.py
tobymao: if you do write="" it will fail

sqlglot/dialects/clickhouse.py
tobymao: it's a pylint standard to do

if ...:
   return 
return ...

sqlglot/dialects/clickhouse.py
tobymao: yea but can't you make this method more dry? teradata and clickhouse are copy pasted

sqlglot/dialects/clickhouse.py
pkit: It is in core parser too.

sqlglot/dialects/clickhouse.py
pkit: yeah, that's how it was initially. :)

sqlglot/dialects/clickhouse.py
pkit: It's more DRY now, and also improved the pretty printer in Teradata

sqlglot/generator.py
```diff
@@ -1582,9 +1578,12 @@ def select_sql(self, expression: exp.Select) -> str:
     def schema_sql(self, expression: exp.Schema) -> str:
         this = self.sql(expression, "this")
         this = f"{this} " if this else ""
-        sql = f"({self.sep('')}{self.expressions(expression)}{self.seg(')', sep='')}"
+        sql = self.columns_sql(expression)
         return f"{this}{sql}"
 
+    def columns_sql(self, expression: exp.Schema) -> str:
```
tobymao, line 1584:
```suggestion
    def schema_columns_sql(self, expression: exp.Schema) -> str:
```

tobymao (approved):

PR #1750, opened 2023-06-09
Refactor!: misc. improvements in formatting, type hints, dialect class variables

GeorgeSittas:
- Dialect properties like `time_mapping` have been converted to uppercase for consistency, i.e. `TIME_MAPPING`.
- Dialect properties removed from constructors; they're now set using metaprogramming in the `_Dialect` class.
- Generator `normalize_functions` arg has changed: now `False` means "don't normalize", `True` is equiv. to `"upper"`.
- Removed the `IfNull` expression in favor of adding `IFNULL` as a `_sql_name` in `Coalesce`.
- Renamed `IS_BOOL` generator flag to `IS_BOOL_ALLOWED` for clarity.
- Tried to replace some Expression constructors with the corresponding helpers in some places.
- Did some manual formatting by adding whitespace to reduce visual noise, unwrapped some multi-line expressions.
- Improved type hints in the Parser.
- Cleaned up some comments.
- Made `validate_expression` return its argument, to improve composability.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -154,15 +154,15 @@ class Dialect(metaclass=_Dialect):
     TIME_MAPPING: t.Dict[str, str] = {}
 
     # autofilled
-    tokenizer_class = None
-    parser_class = None
-    generator_class = None
+    tokenizer_class = Tokenizer
+    parser_class = Parser
+    generator_class = Generator
 
     # A trie of the time_mapping keys
     TIME_TRIE: t.Optional[t.Dict] = None
 
-    INVERSE_TIME_MAPPING: t.Optional[t.Dict[str, str]] = None
-    INVERSE_TIME_TRIE: t.Optional[t.Dict] = None
+    INVERSE_TIME_MAPPING: t.Dict[str, str] = {}
+    INVERSE_TIME_TRIE: t.Dict = {}
 
     def __eq__(self, other: t.Any) -> bool:
         return type(self) == other
@@ -220,14 +220,14 @@ class Dialect(metaclass=_Dialect):
     @property
     def tokenizer(self) -> Tokenizer:
         if not hasattr(self, "_tokenizer"):
-            self._tokenizer = t.cast(t.Type[Tokenizer], self.tokenizer_class)()
+            self._tokenizer = self.tokenizer_class()
         return self._tokenizer
 
     def parser(self, **opts) -> Parser:
-        return t.cast(t.Type[Parser], self.parser_class)(**opts)
+        return self.parser_class(**opts)
 
     def generator(self, **opts) -> Generator:
-        return t.cast(t.Type[Generator], self.generator_class)(**opts)
+        return self.generator_class(**opts)
 
 
 DialectType = t.Union[str, Dialect, t.Type[Dialect], None]
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -153,7 +153,7 @@ class Dialect(metaclass=_Dialect):
     # and the value the target time format
     TIME_MAPPING: t.Dict[str, str] = {}
 
-    # autofilled
+    # Autofilled
     tokenizer_class = Tokenizer
     parser_class = Parser
     generator_class = Generator
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -159,7 +159,7 @@ class Dialect(metaclass=_Dialect):
     generator_class = Generator
 
     # A trie of the time_mapping keys
-    TIME_TRIE: t.Optional[t.Dict] = None
+    TIME_TRIE: t.Dict = {}
 
     INVERSE_TIME_MAPPING: t.Dict[str, str] = {}
     INVERSE_TIME_TRIE: t.Dict = {}
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -126,16 +126,31 @@ class _Dialect(type):
 
 
 class Dialect(metaclass=_Dialect):
+    # Determines the base index offset for arrays
     INDEX_OFFSET = 0
+
+    # If true unnest table aliases are considered only as column aliases
     UNNEST_COLUMN_ONLY = False
+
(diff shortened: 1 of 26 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -99,44 +99,58 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
...
+        klass.RAW_START, klass.RAW_END = get_start_end(TokenType.RAW_STRING)
 
-        klass.tokenizer_class.identifiers_can_start_with_digit = (
-            klass.identifiers_can_start_with_digit
-        )
+        dialect_properties = {
+            **{
+                k: v
+                for k, v in vars(klass).items()
+                if not callable(v) and not isinstance(v, classmethod) and not k.startswith("__")
```
tobymao, line 111: aren't callable(v) and classmethod the same?

GeorgeSittas: That's what I expected too, but for some reason that didn't seem to be the case.

GeorgeSittas: Observe:

```python
>>> class A:
...     @classmethod
...     def foo(cls):
...             pass
...     a = 5
...
>>>
>>> {k: v for k, v in vars(A).items() if not callable(v) and not k.startswith("__")}
{'foo': <classmethod(<function A.foo at 0x1011bf130>)>, 'a': 5}
```

sqlglot/generator.py
```diff
@@ -235,146 +229,114 @@ class Generator:
...
+    IDENTIFIER_START = '"'
+    IDENTIFIER_END = '"'
+    STRING_ESCAPE = "'"
+    IDENTIFIER_ESCAPE = '"'
+    INDEX_OFFSET = 0
+    UNNEST_COLUMN_ONLY = False
+    ALIAS_POST_TABLESAMPLE = False
+    IDENTIFIERS_CAN_START_WITH_DIGIT = False
+    NORMALIZE_FUNCTIONS: bool | str = "upper"
+    NULL_ORDERING = "nulls_are_small"
```
tobymao, line 255: add the comments back for this

sqlglot/generator.py
```diff
@@ -235,146 +229,114 @@ class Generator:
...
+    QUOTE_END = "'"
+    IDENTIFIER_START = '"'
+    IDENTIFIER_END = '"'
+    STRING_ESCAPE = "'"
+    IDENTIFIER_ESCAPE = '"'
+    INDEX_OFFSET = 0
+    UNNEST_COLUMN_ONLY = False
+    ALIAS_POST_TABLESAMPLE = False
+    IDENTIFIERS_CAN_START_WITH_DIGIT = False
+    NORMALIZE_FUNCTIONS: bool | str = "upper"
```
tobymao, line 254: why did you change the type to bool? leave it as string, then you don't need to check is True later down

sqlglot/generator.py
GeorgeSittas: I can move them to `Dialect`, since that's where all dialect props are defined. What do you say?

sqlglot/generator.py
GeorgeSittas: Because we want to support switching off normalization. This was previously done through `None`, but since this property is both set in the dialect and can be passed in as an argument, `None` can mean "not set" in the context of the constructor, in which case we use the dialect's default setting. This is why I used `False` instead of `None`, hence the `bool` type. FWIW I think `False` conveys the intent more clearly too.

sqlglot/generator.py
tobymao: ok

tobymao (approved):

sqlglot/generator.py
GeorgeSittas: Done

sqlglot/dialects/dialect.py
```diff
@@ -220,14 +220,14 @@ def tokenize(self, sql: str) -> t.List[Token]:
...
+            self._tokenizer = self.tokenizer_class()
         return self._tokenizer
 
     def parser(self, **opts) -> Parser:
-        return t.cast(t.Type[Parser], self.parser_class)(**opts)
+        return self.parser_class(**opts)
 
     def generator(self, **opts) -> Generator:
-        return t.cast(t.Type[Generator], self.generator_class)(**opts)
+        return self.generator_class(**opts)
```
GeorgeSittas: Should we get rid of `parser` and `generator` and just rename the internal fields (*_class) to these instead? Perhaps do the same for the `tokenizer`?

cc: @tobymao

GeorgeSittas:
Left one final question, but otherwise this PR is ready to go so feel free to merge.

sqlglot/dialects/dialect.py
GeorgeSittas: FYI: I've made it so that the internal `*_class` fields of `Dialect` are initialized to the corresponding classes instead of `None`.

tobymao: not sure what you mea

GeorgeSittas: I mean get rid of the `parser` method and rename `parser_class` to `parser` (and similarly got the `generator` method).

tobymao: sure

GeorgeSittas: Nvm I think it doesn't hurt to leave it as is. Tried to make the change and it felt kinda asymmetric because we don't do the same for the tokenizer (since we cache it), so it looks weird. I'll just go ahead and merge without this.

GeorgeSittas:
FYI I reverted the type hint change in the following commit because mypy was yelling
https://github.com/tobymao/sqlglot/commit/7000a6f137aabb5d2d2417179501905f37810768#diff-63eb8dd82d3561bc414e3e13cf59516bd6584c5766a10072157a8a5aee4ad85fL3542-R3542

Note that I changed the signature of `_parse_bracket` in this PR since the previous one didn't seem correct. I think the reason mypy is yelling now is because the main body of the method is also typed and so type checking kicks off. Before we used to have `@t.overload`s but the actual definition didn't specify types, so my guess is that the method wasn't checked.

PR #1752, opened 2023-06-09
Feat(clickhouse): support CREATE VIEW TO syntax

pkit:
see: https://clickhouse.com/docs/en/sql-reference/statements/create/view#materialized-view


sqlglot/tokens.py
```diff
@@ -284,6 +284,7 @@ class TokenType(AutoName):
     STRUCT = auto()
     TABLE_SAMPLE = auto()
     TEMPORARY = auto()
+    TO = auto()
```
GeorgeSittas, line 287: This isn't used / needed, let's get rid of it and should be good.

tests/dialects/test_clickhouse.py
```diff
@@ -45,6 +45,9 @@ def test_clickhouse(self):
         self.validate_identity(
             "CREATE MATERIALIZED VIEW test_view ON CLUSTER cl1 (id UInt8) ENGINE=AggregatingMergeTree() ORDER BY tuple() AS SELECT * FROM test_data"
         )
+        self.validate_identity(
+            "CREATE MATERIALIZED VIEW test_view ON CLUSTER cl1 (id UInt8) TO table1 AS SELECT * FROM test_data"
+        )
```
GeorgeSittas, line 50: From the docs it seems like `db.table` is also a valid form (see: `[TO[db.]name]`), should we add a test for it too?

pkit: Good idea!

tobymao (approved):

PR #1761, opened 2023-06-13
Fix(Postgres): Support UNNEST

### LINKED ISSUE #1760: Support Postgres UNNEST
Postgres' [`unnest`](https://www.postgresql.org/docs/15/functions-array.html) does the same as e.g. Spark's [`explode`](https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.functions.explode.html). `exp.Unnest` exists and is tokenized, but for Postgres it's parsed as an anonymous function:

```
>>> sqlglot.__version__
'16.0.0'
>>> sqlglot.parse_one("select unnest(c) from t", read='postgres')
(SELECT expressions:
  (ANONYMOUS this: unnest, expressions:
    (COLUMN this:
      (IDENTIFIER this: c, quoted: False))), from:
  (FROM this:
    (TABLE this:
      (IDENTIFIER this: t, quoted: False))))
>>> sqlglot.parse_one("select unnest(array[1])", read='postgres')
(SELECT expressions:
  (ANONYMOUS this: unnest, expressions:
    (ARRAY expressions:
      (LITERAL this: 1, is_string: False))))
```

On transpile to Presto (which is the behaviour I want), its behaviour should be the same as for `exp.Explode`.

vegarsti:
Fixes #1760.

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -269,6 +269,7 @@ class Postgres(Dialect):
             "NOW": exp.CurrentTimestamp.from_arg_list,
             "TO_CHAR": format_time_lambda(exp.TimeToStr, "postgres"),
             "TO_TIMESTAMP": _to_timestamp,
+            "UNNEST": lambda args: exp.Unnest(expressions=[seq_get(args, 0)]),
         }
 
         FUNCTION_PARSERS = {
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -269,7 +269,7 @@ class Postgres(Dialect):
             "NOW": exp.CurrentTimestamp.from_arg_list,
             "TO_CHAR": format_time_lambda(exp.TimeToStr, "postgres"),
             "TO_TIMESTAMP": _to_timestamp,
-            "UNNEST": lambda args: exp.Unnest(expressions=[seq_get(args, 0)]),
+            "UNNEST": exp.Explode.from_arg_list,
         }
 
         FUNCTION_PARSERS = {
@@ -330,6 +330,7 @@ class Postgres(Dialect):
             **generator.Generator.TRANSFORMS,
             exp.BitwiseXor: lambda self, e: self.binary(e, "#"),
             exp.ColumnDef: transforms.preprocess([_auto_increment_to_serial, _serial_to_generated]),
+            exp.Explode: rename_func("UNNEST"),
             exp.JSONExtract: arrow_json_extract_sql,
             exp.JSONExtractScalar: arrow_json_extract_scalar_sql,
             exp.JSONBExtract: lambda self, e: self.binary(e, "#>"),
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -154,8 +154,8 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
-    """Convert explode/posexplode into unnest (used in hive -> presto)."""
+def to_unnest_cross_join(expression: exp.Expression) -> exp.Expression:
+    """Convert unnest/explode/posexplode into unnest cross join (Presto requires it to be a cross join)."""
     if isinstance(expression, exp.Select):
         from sqlglot.optimizer.scope import build_scope
 
@@ -179,10 +179,12 @@ def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
                 explode_alias = select.aliases[1].name
                 select = select.this
 
-            if isinstance(select, (exp.Explode, exp.Posexplode)):
+            if isinstance(select, (exp.Unnest, exp.Explode, exp.Posexplode)):
                 is_posexplode = isinstance(select, exp.Posexplode)
 
-                explode_arg = select.this
+                explode_arg = (
+                    select.expressions[0] if isinstance(select, exp.Unnest) else select.this
+                )
                 unnest = exp.Unnest(expressions=[explode_arg.copy()], ordinality=is_posexplode)
 
                 # This ensures that we won't use [POS]EXPLODE's argument as a new selection
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -154,8 +154,8 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def to_unnest_cross_join(expression: exp.Expression) -> exp.Expression:
-    """Convert unnest/explode/posexplode into unnest cross join (Presto requires it to be a cross join)."""
+def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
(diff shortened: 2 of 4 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -269,6 +269,7 @@ class Parser(parser.Parser):
             "NOW": exp.CurrentTimestamp.from_arg_list,
             "TO_CHAR": format_time_lambda(exp.TimeToStr, "postgres"),
             "TO_TIMESTAMP": _to_timestamp,
+            "UNNEST": lambda args: exp.Unnest(expressions=[seq_get(args, 0)]),
```
vegarsti, line 272: I don't know where you draw the line on when to extract this into a function, but I reckon this is fine to inline?

sqlglot/transforms.py
```diff
@@ -154,8 +154,8 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
-    """Convert explode/posexplode into unnest (used in hive -> presto)."""
+def to_unnest_cross_join(expression: exp.Expression) -> exp.Expression:
+    """Convert unnest/explode/posexplode into unnest cross join (Presto requires it to be a cross join)."""
```
vegarsti, line 158: I changed the name of this since it's no longer explode to unnest, but rather adding the cross join. Advice on naming much appreciated.

sqlglot/transforms.py
```diff
@@ -179,10 +179,12 @@ def explode_to_unnest(expression: exp.Expression) -> exp.Expression:
...
                 select = select.this
 
-            if isinstance(select, (exp.Explode, exp.Posexplode)):
+            if isinstance(select, (exp.Unnest, exp.Explode, exp.Posexplode)):
                 is_posexplode = isinstance(select, exp.Posexplode)
 
-                explode_arg = select.this
+                explode_arg = (
+                    select.expressions[0] if isinstance(select, exp.Unnest) else select.this
+                )
```
vegarsti, line 187: Given that these changes were all that was required for supporting unnest here, I opted for putting it in this function. But I'd understand if you prefer putting it in a separate function

tests/dialects/test_postgres.py
```diff
@@ -85,6 +85,20 @@ def test_ddl(self):
                 read="postgres",
             )
 
+    def test_unnest(self):
+        self.validate_all(
+            "SELECT UNNEST(c) FROM t",
+            write={
+                "presto": "SELECT col FROM t CROSS JOIN UNNEST(c) AS _u(col)",
```
vegarsti, line 92: Should maybe add a hive translation here, too? I think it should become explode in Hive? i.e. `SELECT EXPLODE(c) FROM t`

vegarsti: On second look, that might be a different PR

vegarsti:
> Can't we just map `UNNEST` to `exp.Explode` for postgres and make sure to rename it @ generation time? I think this would remove the need to add all this custom logic to also handle `exp.Unnest`

Oh, interesting, will try that!

vegarsti:
`exp.Explode` takes a `this` while `exp.Unnest` takes an `expressions`, though, does that matter? How does the mapping/renaming work?

GeorgeSittas:
> How does the mapping/renaming work?

You have to map `"UNNEST"` to `exp.Explode.from_arg_list` in `Postgres.Parser.FUNCTIONS` and then add an entry in `Postgres.Generator.TRANSFORMS` to map `exp.Explode` to `rename_func("UNNEST")`.

vegarsti:
@GeorgeSittas That was neat! Thanks for the explanation.

GeorgeSittas (approved):
Looks way cleaner :)

GeorgeSittas:
Ah I just noticed something in [postgres's docs](https://www.postgresql.org/docs/15/functions-array.html): `UNNEST` can actually receive >1 arguments. With this approach we'll crash if we get the following query (from the doc page):

```sql
SELECT * FROM UNNEST(ARRAY[1,2], ARRAY['foo','bar','baz']) AS x(a,b)
```

So this is actually breaking, because before we'd parse just parse this into an `exp.Anonymous` without problems. Maybe we can simply allow >1 args in `exp.Explode`, in order to also keep the solution here simple?

cc: @tobymao

tobymao:
sure, please fix

vegarsti:
> Ah I just noticed something in [postgres's docs](https://www.postgresql.org/docs/15/functions-array.html): `UNNEST` can actually receive >1 arguments. With this approach we'll crash if we get the following query (from the doc page):
> 
> ```sql
> SELECT * FROM UNNEST(ARRAY[1,2], ARRAY['foo','bar','baz']) AS x(a,b)
> ```
> 
> So this is actually breaking, because before we'd just parse this into an `exp.Anonymous` without problems. Maybe we can simply allow >1 args in `exp.Explode`, in order to also keep the solution here simple?
> 
> cc: @tobymao

Ah, good catch! What does it entail to let it take more than one argument? I won't have time to do that tonight, so feel free to do it if you have time. Otherwise happy to try tomorrow!

GeorgeSittas:
No worries, I'll play around with it locally and let you know when I have something.

GeorgeSittas:
Ha, this is actually already taken care of -- I think there's no problem with this approach!

- Postgres docs mention that the multi-argument form is only allowed in a query's FROM clause.
- In the definition of `_parse_table` we try to parse an `UNNEST` expression before hitting the function parser:

```python
    def _parse_table(
        self, schema: bool = False, alias_tokens: t.Optional[t.Collection[TokenType]] = None
        lateral = self._parse_lateral()
        if lateral:
            return lateral

        unnest = self._parse_unnest()
        if unnest:
            return unnest
        ...
```

This means that with this PR the parsing of multi-argument `UNNEST` expressions doesn't break for Postgres:

```python
>>> import sqlglot
>>> sqlglot.parse_one("SELECT * FROM UNNEST(ARRAY[1,2]) AS x(a,b)", "postgres")
(SELECT expressions:
  (STAR ), from:
  (FROM this:
    (UNNEST expressions:
      (ARRAY expressions:
        (LITERAL this: 1, is_string: False),
        (LITERAL this: 2, is_string: False)), alias:
      (TABLEALIAS this:
        (IDENTIFIER this: x, quoted: False), columns:
        (IDENTIFIER this: a, quoted: False),
        (IDENTIFIER this: b, quoted: False)))))
>>> sqlglot.parse_one("SELECT UNNEST(c) FROM t", "postgres")
(SELECT expressions:
  (EXPLODE this:
    (COLUMN this:
      (IDENTIFIER this: c, quoted: False))), from:
  (FROM this:
    (TABLE this:
      (IDENTIFIER this: t, quoted: False))))
```

There's only a (minor) issue, where `UNNEST`s in the `FROM` clause are represented as `exp.Unnest` nodes, whereas single-argument `UNNEST`s anywhere else are represented as `exp.Explode` nodes. If we don't mind having this disparity, then this PR is good to go.

PR #1780, opened 2023-06-15
Fix!(duckdb): wrap negations, transpile week, quarter to multiplication

### LINKED ISSUE #1779: Transforming a presto literal timestamp query to duckdb
**Before you file an issue**
This issue report contains 2 problems with target dialect of duckdb
The read dialect is "presto"
The target dialect is "duckdb"
The query was generated by Looker 

Confirmed

```
import sqlglot

sql = "select ((DATE_ADD('week', -5, DATE_TRUNC('DAY', DATE_ADD('day', (0 - MOD((DAY_OF_WEEK(CAST(CAST(DATE_TRUNC('DAY', NOW()) AS DATE) AS TIMESTAMP)) % 7) - 1 + 7, 7)), CAST(CAST(DATE_TRUNC('DAY', NOW()) AS DATE) AS TIMESTAMP)))))) as t1"
glot_tree = sqlglot.parse_one(sql, "presto");
print(glot_tree.sql("duckdb"))

# Produces the below output which is non functional on DuckDB
# SELECT ((DATE_TRUNC('DAY', CAST(CAST(DATE_TRUNC('DAY', CURRENT_TIMESTAMP) AS DATE) AS TIMESTAMP) + INTERVAL (0 - MOD((DAYOFWEEK(CAST(CAST(DATE_TRUNC('DAY', CURRENT_TIMESTAMP) AS DATE) AS TIMESTAMP)) % 7) - 1 + 7, 7)) day) + INTERVAL -5 week)) AS t1

# The below is the correct output, DUCKDB does not like negative intervals nor does it support week intervals
# Bad Query
# SELECT interval -1 day
# Good Alternatives
# SELECT interval (-1) day
# SELECT to_days(-1)

# DuckDB does not support week intervals
# Bad Query
# SELECT interval -1 week
# Good Query
# SELECT interval (-1 * 7 )  day
# SELECT to_days(-1 * 7)


```

**Official Documentation**

https://duckdb.org/docs/sql/functions/interval.html

Finally i realized that there is an easier fix by wrapping the interval part in simple quote but it the macro call approach might be more flexible in terms of compos-ability of one what might do with SQL 
select '-1 week'
select '5 day'

GeorgeSittas:
Ran all tests in the corresponding engines, they look good.

Note: the tweak I made in the Generator class was to enable the overriding of some (previously) tuple attributes.

References:

- https://duckdb.org/docs/sql/functions/interval.html
- https://duckdb.org/docs/sql/functions/datepart#usable-in-date-part-specifiers-only

Fixes #1779

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -12,7 +12,17 @@ from sqlglot.tokens import TokenType
 logger = logging.getLogger("sqlglot")
 
 
-class Generator:
+class _Generator(type):
+    def __new__(cls, clsname, bases, attrs):
+        klass = super().__new__(cls, clsname, bases, attrs)
+
+        klass._WITH_SEPARATED_COMMENTS = tuple(klass.WITH_SEPARATED_COMMENTS)
+        klass._UNWRAPPED_INTERVAL_VALUES = tuple(klass.UNWRAPPED_INTERVAL_VALUES)
+
+        return klass
+
+
+class Generator(metaclass=_Generator):
     """
     Generator converts a given syntax tree to the corresponding SQL string.
 
@@ -234,9 +244,11 @@ class Generator:
     # Keywords that can't be used as unquoted identifier names
     RESERVED_KEYWORDS: t.Set[str] = set()
 
-    WITH_SEPARATED_COMMENTS = (exp.Select, exp.From, exp.Where, exp.With)
+    # Expressions whose comments are separated from them for better formatting
+    WITH_SEPARATED_COMMENTS = {exp.Select, exp.From, exp.Where, exp.With}
 
-    UNWRAPPED_INTERVAL_VALUES = (exp.Column, exp.Literal, exp.Neg, exp.Paren)
+    # Expressions that can remain unwrapped when appearing in the context of an INTERVAL
+    UNWRAPPED_INTERVAL_VALUES = {exp.Column, exp.Literal, exp.Neg, exp.Paren}
 
     SENTINEL_LINE_BREAK = "__SQLGLOT__LB__"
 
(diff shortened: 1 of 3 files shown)


sqlglot/generator.py
```diff
@@ -12,7 +12,17 @@
 logger = logging.getLogger("sqlglot")
 
 
-class Generator:
+class _Generator(type):
+    def __new__(cls, clsname, bases, attrs):
+        klass = super().__new__(cls, clsname, bases, attrs)
+
+        klass._WITH_SEPARATED_COMMENTS = tuple(klass.WITH_SEPARATED_COMMENTS)
```
tobymao, line 19: what's the point of this?

GeorgeSittas: Mostly trying to respect type constraints, as overriding a tuple attribute to create a tuple of different type (or cardinality, in this case) in a subclass is not a valid form and mypy yells. Happy to revert the change and throw some `# type: ignore`s in there if you think it's unnecessary overhead / complexity.

tobymao: why do you need tuples? just keep it a set?

GeorgeSittas: We need tuples because these expression types are used in instance checks

tobymao: just make the class WITH_SEPARATE_COMMENTS: t.Tuple[t.Type[exp.Expression], ...]

GeorgeSittas: Yep, tried that unfortunately it also didn't solve the problem:

```
sqlglot/dialects/duckdb.py:238: error: Incompatible types in assignment (expression has type "Tuple[Type[Expression], ...]", base class "Generator" defined the type as "Tuple[Type[Column], Type[Literal], Type[Neg], Type[Paren]]")  [assignment]

PR #1784, opened 2023-06-15
Refactor!: move normalization logic in Dialect, update case-sensitivity info

GeorgeSittas:


sqlglot/dataframe/sql/readwriter.py
```diff
@@ -26,7 +26,7 @@ def table(self, tableName: str) -> DataFrame:
             .from_(tableName)
             .select(
                 *(
-                    column if should_identify(column, "safe") else f'"{column}"'
+                    column if can_identify(column, "safe") else f'"{column}"'
```
GeorgeSittas, line 29: I renamed `should_identify` to `can_identify` exactly because this section was awkward to read. I think it's easy to read this and go "why are we _identifying_ if we should _not_ identify, according to the condition?".

sqlglot/helper.py
```diff
@@ -432,24 +432,28 @@ def first(it: t.Iterable[T]) -> T:
 
 def case_sensitive(text: str, dialect: DialectType) -> bool:
     """Checks if text contains any case sensitive characters depending on dialect."""
-    from sqlglot.dialects.dialect import RESOLVES_IDENTIFIERS_AS_UPPERCASE
+    from sqlglot.dialects.dialect import Dialect
 
-    unsafe = str.islower if dialect in RESOLVES_IDENTIFIERS_AS_UPPERCASE else str.isupper
+    dialect = Dialect.get_or_raise(dialect)
+    if dialect.RESOLVES_IDENTIFIERS_AS_UPPERCASE is None:
+        return False
```
GeorgeSittas, line 439: Note: maybe we can refactor this section somehow because we're doing a `Dialect` lookup every time, so the overhead may be non-trivial when we're calling `case_sensitive` in `_quote` which is passed to `transform` (inside of `quote_identifiers`). Similarly, `should_identify` calls `case_sensitive` as well.

GeorgeSittas: Good point -- also moved `_quote` in `Dialect` for the same reason.

sqlglot/dialects/dialect.py
```diff
@@ -139,6 +132,10 @@ class Dialect(metaclass=_Dialect):
     # Determines whether or not the table alias comes after tablesample
     ALIAS_POST_TABLESAMPLE = False
 
+    # Determines whether or not unquoted identifiers are resolved as uppercase
+    # When set to None, it means that the dialect treats all identifiers as case-insensitive
+    RESOLVES_IDENTIFIERS_AS_UPPERCASE: t.Optional[bool] = False
```
tobymao, line 137: hive/spark is also case insensitive, can you double check that with presto?

sqlglot/helper.py
```diff
@@ -430,29 +430,25 @@ def first(it: t.Iterable[T]) -> T:
...
-    """Checks if text contains any case sensitive characters depending on dialect."""
-    from sqlglot.dialects.dialect import RESOLVES_IDENTIFIERS_AS_UPPERCASE
-
-    unsafe = str.islower if dialect in RESOLVES_IDENTIFIERS_AS_UPPERCASE else str.isupper
-    return any(unsafe(char) for char in text)
-
-
-def should_identify(text: str, identify: str | bool, dialect: DialectType = None) -> bool:
-    """Checks if text should be identified given an identify option.
+def can_identify(text: str, identify: str | bool, dialect: DialectType = None) -> bool:
```
tobymao, line 433: why doesn't this move to dialect too?

tobymao (approved):

sqlglot/dialects/bigquery.py
```diff
@@ -126,6 +130,17 @@ class BigQuery(Dialect):
         "TZH": "%z",
     }
 
+    @classmethod
+    def normalize_identifier(cls, expression: E) -> E:
+        # In BigQuery CTEs are not case-sensitive, but table names are. The following check is
+        # essentially a heuristic to detect tables based on whether or not they're qualified.
+        if isinstance(expression, exp.Identifier) and not (
+            isinstance(expression.parent, exp.Table) and len(expression.parent.parts) > 1
```
GeorgeSittas, line 138: Maybe we can just check if `.db` is truthy here instead of calling `.parts`?

PR #1790, opened 2023-06-16
Fix: use IF expression instead of COALESCE in _parse_concat

GeorgeSittas:
Addresses #1789

cc: @mpf82

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -344,7 +344,9 @@ class ClickHouse(Dialect):
             return self.func(
                 "CONCAT",
                 *[
-                    exp.func("if", e.is_(exp.null()), e, exp.cast(e, "text"))
+                    exp.func(
+                        "if", exp.Is(this=e, expression=exp.null()), e.copy(), exp.cast(e, "text")
+                    )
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -344,7 +344,9 @@ def safeconcat_sql(self, expression: exp.SafeConcat) -> str:
             return self.func(
                 "CONCAT",
                 *[
-                    exp.func("if", e.is_(exp.null()), e, exp.cast(e, "text"))
+                    exp.func(
+                        "if", exp.Is(this=e, expression=exp.null()), e.copy(), exp.cast(e, "text")
+                    )
```
GeorgeSittas, line 349: This wasn't picked up by mypy because the `expressions` property is untyped, but doing `e.is_(..)` in this context is not correct because the method is only defined for `Condition`s and these arguments could be anything.

@tobymao

GeorgeSittas:
Seems like Postgres implements the same semantics for all conditional expressions, so using `CASE` here doesn't fix the problem... 🤦

Closing this for now; I'll try think of a different workaround.

Reference: https://www.postgresql.org/docs/current/functions-conditional.html

PR #1792, opened 2023-06-16
Fix(parser): cast coalesce arg to text in the context of a CONCAT call

GeorgeSittas:
Addresses #1789

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -344,7 +344,7 @@ class ClickHouse(Dialect):
             return self.func(
                 "CONCAT",
                 *[
-                    exp.func("if", e.is_(exp.null()), e, exp.cast(e, "text"))
+                    exp.func("if", exp.Is(this=e, expression=exp.null()), e, exp.cast(e, "text"))
                     for e in expression.expressions
                 ],
             )
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -344,8 +344,8 @@ class ClickHouse(Dialect):
             return self.func(
                 "CONCAT",
                 *[
-                    exp.func("if", exp.Is(this=e, expression=exp.null()), e, exp.cast(e, "text"))
-                    for e in expression.expressions
+                    exp.func("if", e.is_(exp.null()), e, exp.cast(e, "text"))
+                    for e in t.cast(t.List[exp.Condition], expression.expressions)
                 ],
             )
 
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -344,7 +344,7 @@ def safeconcat_sql(self, expression: exp.SafeConcat) -> str:
             return self.func(
                 "CONCAT",
                 *[
-                    exp.func("if", e.is_(exp.null()), e, exp.cast(e, "text"))
+                    exp.func("if", exp.Is(this=e, expression=exp.null()), e, exp.cast(e, "text"))
```
GeorgeSittas, line 347: This wasn't picked up by mypy because the expressions property is untyped, but doing e.is_(..) in this context is not correct because the method is only defined for Conditions and these arguments could be anything.

tobymao: you can't safe contact an identifier, e should always be a column

tests/dialects/test_postgres.py
```diff
@@ -577,13 +577,19 @@ def test_bool_or(self):
...
         self.validate_all(
             "CONCAT(a, b)",
             write={
-                "": "CONCAT(COALESCE(a, ''), COALESCE(b, ''))",
-                "duckdb": "CONCAT(COALESCE(a, ''), COALESCE(b, ''))",
-                "postgres": "CONCAT(COALESCE(a, ''), COALESCE(b, ''))",
-                "presto": "CONCAT(CAST(COALESCE(a, '') AS VARCHAR), CAST(COALESCE(b, '') AS VARCHAR))",
+                "": "CONCAT(COALESCE(CAST(a AS TEXT), ''), COALESCE(CAST(b AS TEXT), ''))",
+                "duckdb": "CONCAT(COALESCE(CAST(a AS TEXT), ''), COALESCE(CAST(b AS TEXT), ''))",
+                "postgres": "CONCAT(COALESCE(CAST(a AS TEXT), ''), COALESCE(CAST(b AS TEXT), ''))",
```
tobymao, line 591: so this is one of the downsides of this approach, postgres concat(a, b) would ideally stay as an identity, this is similar to tsordstodate, if we had created a SAFE_CONCAT, then we would get so much noise

sqlglot/dialects/clickhouse.py
GeorgeSittas: Ok I reverted this change and just made `Placeholder`, `Parameter` and `ParameterSession` be `Condition`s.

tests/dialects/test_postgres.py
GeorgeSittas: Discussed in Slack -- we already have a `SafeConcat` expression. The issue is that we have two axes of variability:

- The first one is “concat’s args must all be strings” --> controlled by the STRICT_STRING_CONCAT property
- The second one is “null values in a concat are interpreted as empty strings” --> controlled by the CONCAT_NULL_OUTPUTS_STRING property

PR #1797, opened 2023-06-17
Refactor!: simplify mypy type hints for parse_one

### LINKED ISSUE #1796: not-sure-if-bug: strange mypy behavior in the latest sqlglot
```python

a = parse_one("SELECT * FROM t").find(exp.Table)
print(a)
```

```bash
$ python -m mypy test.py
test.py:3: error: <nothing> has no attribute "find"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

Could be a mypy bug or the usual mypy idiocy, but just FYI.

GeorgeSittas:
Fixes #1796

The idea is that there's no point for all the different overloads, since the only thing we're interested in is matching `into`'s type with the return type.

### DIFF
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -84,58 +84,17 @@ def parse(sql: str, read: DialectType = None, **opts) -> t.List[t.Optional[Expre
 
 
 @t.overload
-def parse_one(
-    sql: str,
-    read: None = None,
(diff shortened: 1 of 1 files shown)


eakmanrq (approved):
Ah nice it makes it a lot easier when you don't have to cover all the combinations. Thanks for cleaning this up!

PR #1803, opened 2023-06-19
Feat(duckdb): support TO_JSON

r1b:
Only added `write=duckdb`. I think we would need something like an `exp.JSONParse` to support `read=duckdb`, since `TO_JSON` returns a JSON type instead of TEXT.

Ref: [DuckDB - JSON Creation Functions](https://duckdb.org/docs/extensions/json#json-creation-functions)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -85,6 +85,21 @@ def _regexp_extract_sql(self: generator.Generator, expression: exp.RegexpExtract
     )
 
 
+def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> str:
+    this = expression.this
+    if not this.type:
+        from sqlglot.optimizer.annotate_types import annotate_types
+
+        annotate_types(this)
+
+    if this.type.is_type("json"):
+        sql = self.sql(this)
+    else:
+        sql = self.func("TO_JSON", this, expression.args.get("options"))
+
+    return f"CAST({sql} AS TEXT)"
+
+
 class DuckDB(Dialect):
     NULL_ORDERING = "nulls_are_last"
 
@@ -195,6 +210,7 @@ class DuckDB(Dialect):
             exp.IntDiv: lambda self, e: self.binary(e, "//"),
             exp.JSONExtract: arrow_json_extract_sql,
             exp.JSONExtractScalar: arrow_json_extract_scalar_sql,
+            exp.JSONFormat: _json_format_sql,
             exp.JSONBExtract: arrow_json_extract_sql,
             exp.JSONBExtractScalar: arrow_json_extract_scalar_sql,
             exp.LogicalOr: rename_func("BOOL_OR"),
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -491,6 +491,7 @@ class TestBigQuery(Validator):
             read={"bigquery": "TO_JSON_STRING(x)"},
             write={
                 "bigquery": "TO_JSON_STRING(x)",
+                "duckdb": "CAST(TO_JSON(x) AS TEXT)",
                 "presto": "JSON_FORMAT(x)",
                 "spark": "TO_JSON(x)",
             },
--- a/tests/dialects/test_presto.py
+++ b/tests/dialects/test_presto.py
@@ -665,6 +665,7 @@ class TestPresto(Validator):
             },
             write={
                 "bigquery": "TO_JSON_STRING(x)",
+                "duckdb": "CAST(TO_JSON(x) AS TEXT)",
                 "presto": "JSON_FORMAT(x)",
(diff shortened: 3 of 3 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -85,6 +85,21 @@ def _regexp_extract_sql(self: generator.Generator, expression: exp.RegexpExtract
     )
 
 
+def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> str:
+    this = expression.this
+    if not this.type:
+        from sqlglot.optimizer.annotate_types import annotate_types
+
+        annotate_types(this)
```
GeorgeSittas, line 93: What case triggered the need to annotate types here? Can't we always just cast to `TEXT`?

GeorgeSittas:
Btw thanks for the PRs @r1b!

sqlglot/dialects/duckdb.py
tobymao: this should only be done in the annotate types modules, let’s delete this

GeorgeSittas: I assume this whole section was copy-pasted from Hive?

https://github.com/tobymao/sqlglot/blob/4da37aa5caec407f66405b39f37d5dad057e66c2/sqlglot/dialects/hive.py#L87-L96

GeorgeSittas: IIRC the Hive approach used `annotate_types` because there was an issue with transpiling Trino's `JSON_FORMAT`, where we'd get back the invalid `TO_JSON(TO_JSON(..))` SQL in Spark (see https://github.com/tobymao/sqlglot/commit/aef9cfa9c for reference).

FWIW, it looks like we're still not transpiling Trino correctly to Spark, cc: @tobymao. For example:

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT JSON_FORMAT(JSON '[1, 2, 3]')", "trino", "spark")
["SELECT TO_JSON('[1, 2, 3]')"]
```

The resulting SQL here is not valid Spark:

```
spark-sql (default)> SELECT TO_JSON('[1, 2, 3]');
[DATATYPE_MISMATCH.INVALID_JSON_SCHEMA] Cannot resolve "to_json([1, 2, 3])" due to data type mismatch: Input schema "STRING" must be a struct, an array or a map.; line 1 pos 7;
```

I'm not sure if we can always transpile this function correctly without having a JSON parser, but for cases where we have nested (*) data structures represented as JSON strings, maybe we can alter the logic of `_json_format_sql` in Hive to output something like the following:

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT JSON_FORMAT(JSON '[1, 2, 3]')", "trino", "spark")
["SELECT TO_JSON(FROM_JSON('[1, 2, 3]', SCHEMA_OF_JSON('[1, 2, 3]')))"]
```

This is now valid Spark:

```
spark-sql (default)> SELECT TO_JSON(FROM_JSON('[1, 2, 3]', SCHEMA_OF_JSON('[1, 2, 3]')));
to_json(from_json([1, 2, 3]))
[1,2,3]
Time taken: 0.261 seconds, Fetched 1 row(s)
```

Similarly, but for a JSON string that encodes a map structure:

```
spark-sql (default)> SELECT TO_JSON(FROM_JSON('{"a": 1, "b": "c"}', SCHEMA_OF_JSON('{"a": 1, "b": "c"}')));
to_json(from_json({"a": 1, "b": "c"}))
{"a":1,"b":"c"}
Time taken: 4.434 seconds, Fetched 1 row(s)
```

Note that the outer `TO_JSON` call was added to cast it into a string again, which is what Trino's `JSON_FORMAT` [function](https://github.com/tobymao/sqlglot/assets/46752250/a250bf73-3c62-456b-8e2f-81062c4a231a) does.

```python
>>> from pyspark.sql import SQLContext
>>> from pyspark.context import SparkContext
>>> SQLContext(SparkContext()).sql("""SELECT TO_JSON(FROM_JSON('{"a": 1, "b": "c"}', SCHEMA_OF_JSON('{"a": 1, "b": "c"}')))""")
DataFrame[to_json(from_json({"a": 1, "b": "c"})): string]
```

I can try to implement this in another PR.

(*) I'm saying "nested" because e.g. for `'"a"'` this conversion fails:

```
spark-sql (default)> SELECT TO_JSON(FROM_JSON('"a"', SCHEMA_OF_JSON('"a"')));
[DATATYPE_MISMATCH.INVALID_JSON_SCHEMA] Cannot resolve "from_json("a")" due to data type mismatch: Input schema "STRING" must be a struct, an array or a map.; line 1 pos 15;
```

tobymao (approved):

sqlglot/dialects/duckdb.py
tobymao: sounds good

PR #1806, opened 2023-06-19
fix(snowflake): add support for TOP <n> keyword

### LINKED ISSUE #1805: Snowflake: TOP <n> is not supported
Copying recent issue #1798, Snowflake supports the same Teradata syntax for [Top N](https://docs.snowflake.com/en/sql-reference/constructs/top_n).

The below fails in sqlglot
```
import sqlglot

sqlglot.transpile("select top 10 column1 from table1.column1", identity=True, read="snowflake")

ParseError: Invalid expression / Unexpected token. Line 1, Col: 13.
  select top 10 column1 from table1.column1
```

Solution is the same applied in #1799 (add TOP in supported keywords list in snowflake dialect). 

Thank you for this library!

ftom:
Fixes #1805.


GeorgeSittas (approved):
Thanks :)

PR #1810, opened 2023-06-20
Feat: Add Postgres user-defined type support

eakmanrq:
Example from information schema: 
```
SELECT column_name, data_type FROM information_schema.columns WHERE table_name = 'table' AND table_schema = 'schema';

 id                                | bigint
 custom_col                        | USER-DEFINED
 ds                                | text
```


tobymao:
need to add to parser?

tobymao:
nvm

eakmanrq:
I added it to the tokens. The way I implemented it is more generic and not postgres specific. Basically assuming for now that all engines with custom types will represent them in this way but this can obviously be changed over time.

PR #1817, opened 2023-06-23
Feat(databricks): add support for REPLACE WHERE in INSERT statement

GeorgeSittas:
Reference: https://docs.databricks.com/sql/language-manual/sql-ref-syntax-dml-insert-into.html


eakmanrq (approved):
LGTM. Thanks for doing this!

PR #1826, opened 2023-06-26
Fix: dont expand bq pseudocolumns in optimizer star expansion

z3z1ma:
Bigquery has pseudocolumns for time ingestion partitioned tables. These pseudocolumns are `_PARTITIONTIME` and `PARTITIONDATE`. Both of these are reserved identifiers, furthermore neither are returned in * expansion by bigquery. So we should respect that during the optimizer's star expansion. To not do so would break a valid query (by adding reserved idents to the projection).


z3z1ma:
Should be good now @tobymao 

`has_pivoted_source` was not a sufficient type guard for mypy FYI, so the guards are more explicit in 2 places. Cost should be practically nothing.

PR #1828, opened 2023-06-26
Feat(mysql): add support for SIGNED [INTEGER] and UNSIGNED [INTEGER] types

brosoul:
Feat(mysql): add support for SIGNED [INTEGER] and UNSIGNED [INTEGER] types
Reference: https://dev.mysql.com/doc/refman/5.7/en/cast-functions.html#function_cast


tobymao:
please fix the build

brosoul:
> please fix the build

Done

tests/dialects/test_mysql.py
```diff
@@ -49,6 +49,10 @@ def test_ddl(self):
     def test_identity(self):
         self.validate_identity("CAST(x AS ENUM('a', 'b'))")
         self.validate_identity("CAST(x AS SET('a', 'b'))")
+        self.validate_identity("CAST(x AS SIGNED)", "CAST(x AS BIGINT)")
+        self.validate_identity("CAST(x AS SIGNED INTEGER)", "CAST(x AS BIGINT)")
+        self.validate_identity("CAST(x AS UNSIGNED)", "CAST(x AS UBIGINT)")
+        self.validate_identity("CAST(x AS UNSIGNED INTEGER)", "CAST(x AS UBIGINT)")
```
GeorgeSittas, line 55: I think MySQL disallows casting to these types. For example, in my local client (which is running MySQL v8):

```
mysql> SELECT CAST(1 AS BIGINT);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'BIGINT)' at line 1
```

We might need to map these types differently. I can take a closer look soon.

Docs: https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html

brosoul: You are right, CAST X AS BIGINT is not supported in MySQL, but in its document declaration
>Signed [INTEGER], products a signed BIGINT value

Should I add the SIGNED and UNSIGNED types to the TokenType, or is there another better method? 🤔
BTW, the role of TYPE_MAPPING confuses me.
@GeorgeSittas

GeorgeSittas: > Should I add the SIGNED and UNSIGNED types to the TokenType, or is there another better method? 🤔

That's one idea, yeah.

Another approach would be to override `datatype_sql` in the Postgres dialect and generate `[UN]SIGNED` instead of `BIGINT`, if the target type of the cast is `DataType.Type.BIGINT`. This way we'd avoid adding redundant tokens.

cc: @tobymao any preference here?

>  BTW, the role of TYPE_MAPPING confuses me.

This is used to to map `DataType.Type` enums to the corresponding type strings for the dialect of interest.

For example, the canonical representation of an integer type in SQLGlot is the `DataType.Type.INT` enum, but we need to map this into `"INTEGER"` for Presto, because it calls its integer type `INTEGER` instead of `INT`.

brosoul: Is there a similar implementation in sqlglot 😂?  
I am puzzled about how the `datatype_sql` of Postgres dialect affects the type of cast in MySQL

GeorgeSittas: My bad, this was a typo. I meant *MySQL dialect. 😅

GeorgeSittas: > Is there a similar implementation in sqlglot 😂?

There are some dialects that override `datatype_sql` (e.g. redshift), so you could check these out.

But don't worry about it, this might be a bit more involved for a first PR, so I'll take a look soon and try to resolve this issue.

To reiterate, the problem is that one can't interchangeably use `BIGINT` and `SIGNED` in MySQL:

```
mysql> create table foo (id bigint);
Query OK, 0 rows affected (0.03 sec)
mysql> create table foo (id signed);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'signed)' at line 1

mysql> select cast(1 as bigint);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'bigint)' at line 1
mysql> select cast(1 as signed);
+-------------------+
| cast(1 as signed) |
+-------------------+
|                 1 |
+-------------------+
1 row in set (0.00 sec)
```

tobymao: wouldn't you override cast, not data_type_sql?

GeorgeSittas: You're right, again my bad. Can't think clearly today :)

But yeah, the idea is that `cast_sql` should check the `to` arg, and if it's `BIGINT` it should use `SIGNED` instead (similarly for `UBIGINT`, it should use `UNSIGNED`).

PR #1832, opened 2023-06-26
Fix(mysql): convert (U)BIGINT to (UN)SIGNED in CAST expressions

GeorgeSittas:
cc: @brosoul

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -445,6 +445,17 @@ class MySQL(Dialect):
 
         LIMIT_FETCH = "LIMIT"
 
+        def cast_sql(self, expression: exp.Cast) -> str:
+            """(U)BIGINT is not allowed in a CAST expression, so we use (UN)SIGNED instead."""
+            if expression.to.this == exp.DataType.Type.BIGINT:
+                to = "SIGNED"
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -445,6 +445,17 @@ class Generator(generator.Generator):
 
         LIMIT_FETCH = "LIMIT"
 
+        def cast_sql(self, expression: exp.Cast) -> str:
+            """(U)BIGINT is not allowed in a CAST expression, so we use (UN)SIGNED instead."""
+            if expression.to.this == exp.DataType.Type.BIGINT:
+                to = "SIGNED"
+            elif expression.to.this == exp.DataType.Type.UBIGINT:
+                to = "UNSIGNED"
```
GeorgeSittas, line 453: Didn't use `is_type` on purpose here, because it fails when `to` is an `exp.CharacterSet`. This is the case here:

```python
        self.validate_all(
            "CONVERT(x USING latin1)",
            write={
                "mysql": "CAST(x AS CHAR CHARACTER SET latin1)",
            },
        )
```

brosoul:
Thanks, the PR helped me better understand the `Generator`🫡

PR #1837, opened 2023-06-27
Fix: interval precedence parsing bug

### LINKED ISSUE #1834: MySQL parser not support operator in DATE_ADD params after INTERVAL
DATE_ADD, DATE_SUB ... 

examle:

`
query = "select DATE_ADD('2023-06-23 12:00:00', INTERVAL 2 * 2  MONTH) from x"

exp = sqlglot.parse_one(query, read=sqlglot.dialects.mysql.MySQL)
`

error:
`
sqlglot.errors.ParseError: Expecting ). Line 2, Col: 68.
  
select DATE_ADD('2023-06-23 12:00:00', INTERVAL 2 * 2  MONTH) from x
`

GeorgeSittas:
Fixes #1834

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2863,7 +2863,11 @@ class Parser(metaclass=_Parser):
         if not self._match(TokenType.INTERVAL):
             return None
 
-        this = self._parse_primary() or self._parse_term()
+        if self._match(TokenType.STRING, advance=False):
+            this = self._parse_primary()
+        else:
+            this = self._parse_term()
+
         unit = self._parse_function() or self._parse_var()
 
         # Most dialects support, e.g., the form INTERVAL '5' day, thus we try to parse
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -2863,7 +2863,11 @@ def _parse_interval(self) -> t.Optional[exp.Interval]:
         if not self._match(TokenType.INTERVAL):
             return None
 
-        this = self._parse_primary() or self._parse_term()
```
tobymao, line 2866: why not just this = self._parse_term() ?

GeorgeSittas: It was breaking at least the following test case:

```python
>>> import sqlglot
>>> sqlglot.transpile("INTERVAL '1 days' * 5")
["INTERVAL ('1 days' * 5)"]
```

The parentheses result in a parse error e.g. in postgres

PR #1839, opened 2023-06-27
Fix(bigquery): handle reserved keywords

serkef:
All reserved keywords are now supported, according to BQ documentation: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords

Solves: https://github.com/TobikoData/sqlmesh/issues/1042


serkef:
not sure if those should in lower case or capitalized

GeorgeSittas (approved):

PR #1841, opened 2023-06-27
Feat!(snowflake): add support for identifier literals

### LINKED ISSUE #1840: Parsing Snowflake Table Name Identifiers Throws Errors
**Issue**
- Dialect: snowflake

When parsing create or drop table statements which build a table name literal using SQL variables, the parser throws a parsing error. This should be valid snowflake code.

```
sqlglot.errors.ParseError: Expecting ). Line 2, Col: 45.
  SET TABLE_NAME = 'table';
CREATE TABLE IDENTIFIER($TABLE_NAME) (COLUMN1 VARCHAR, COLUMN2 VARCHAR);
DROP TABLE IDENTIFIER($TABLE_NAME);
```

It seems to believe the `IDENTIFIER()` function is actually the table name and the following function parameters as the column declarations.

```
SET TABLE_NAME = 'table';

CREATE TABLE IDENTIFIER($TABLE_NAME) (COLUMN1 VARCHAR, COLUMN2 VARCHAR);

DROP TABLE IDENTIFIER($TABLE_NAME);
```

**Official Documentation**
https://docs.snowflake.com/en/sql-reference/identifier-literal

GeorgeSittas:
Fixes #1840

Reference: https://docs.snowflake.com/en/sql-reference/identifier-literal

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -272,8 +272,11 @@ class Snowflake(Dialect):
             any_token: bool = True,
             tokens: t.Optional[t.Collection[TokenType]] = None,
         ) -> t.Optional[exp.Expression]:
-            if self._match_text_seq("IDENTIFIER", "(") or self._match_text_seq("TABLE", "("):
-                identifier = super()._parse_id_var(any_token=any_token, tokens=tokens)
+            if self._match_text_seq("IDENTIFIER", "("):
+                identifier = (
+                    super()._parse_id_var(any_token=any_token, tokens=tokens)
+                    or self._parse_string()
+                )
                 self._match_r_paren()
                 return self.expression(exp.ToIdentifier, this=identifier)
 
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -278,7 +278,7 @@ class Snowflake(Dialect):
                     or self._parse_string()
                 )
                 self._match_r_paren()
-                return self.expression(exp.ToIdentifier, this=identifier)
+                return self.expression(exp.Anonymous, this="IDENTIFIER", expressions=[identifier])
 
             return super()._parse_id_var(any_token=any_token, tokens=tokens)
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -267,6 +267,21 @@ class Parser(parser.Parser):
...
+            any_token: bool = True,
+            tokens: t.Optional[t.Collection[TokenType]] = None,
+        ) -> t.Optional[exp.Expression]:
+            if self._match_text_seq("IDENTIFIER", "("):
+                identifier = (
+                    super()._parse_id_var(any_token=any_token, tokens=tokens)
+                    or self._parse_string()
+                )
+                self._match_r_paren()
+                return self.expression(exp.ToIdentifier, this=identifier)
```
GeorgeSittas, line 281: This is reached when parsing DDL table names, since `schema` is set to `True` in that case and so we don't hit the `_parse_function` call here:

https://github.com/tobymao/sqlglot/blob/19295cc5d0080883af183771512f8e8a4050eecd/sqlglot/parser.py#L2311

In all other cases, `IDENTIFIER(...)` is parsed into `exp.ToIdentifier` in `_parse_function`.

tests/dialects/test_snowflake.py
```diff
@@ -648,40 +652,40 @@ def test_table_literal(self):
         # All examples from https://docs.snowflake.com/en/sql-reference/literals-table.html
         self.validate_all(
             r"""SELECT * FROM TABLE('MYTABLE')""",
-            write={"snowflake": r"""SELECT * FROM TABLE('MYTABLE')"""},
+            write={"snowflake": r"""SELECT * FROM IDENTIFIER('MYTABLE')"""},
```
GeorgeSittas, line 655: These two forms are equivalent -- the tests were changed in because `TABLE(...)` is now parsed into an `exp.ToIdentifier` expression and its first `_sql_name` is `IDENTIFIER`, instead of table.

PR #1847, opened 2023-06-28
Feat(bigquery): pushdown CTE column names

GeorgeSittas:


sqlglot/dialects/bigquery.py
```diff
@@ -104,6 +107,33 @@ def _unqualify_unnest(expression: exp.Expression) -> exp.Expression:
...
+        column_names = expression.alias_column_names
+
+        if cte_query.is_star:
+            logger.warning(
+                "Can't push down CTE column names for star queries. Run the query through"
+                " the optimizer or use 'qualify' to expand the star projections first."
+            )
+            return expression
+
+        expression.args["alias"].set("columns", None)
```
tobymao, line 123: expression.args["alias"].pop("columns")

GeorgeSittas: Will this work with `columns` being a list?

tobymao (approved):

sqlglot/dialects/bigquery.py
tobymao: aren't we just getting rid of columns?

GeorgeSittas: Maybe we can update the API of `pop` so that it receives a `key` arg to set to `None`.

GeorgeSittas: > aren't we just getting rid of columns?

Yep, it's just that this API is not currently supported. No big issue though, I can just add an optional arg in `pop`.

PR #1848, opened 2023-06-28
Feat(bigquery): add support for casting to string w/ format

GeorgeSittas:
Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_functions#cast_as_string

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -335,9 +335,6 @@ class BigQuery(Dialect):
             **generator.Generator.TRANSFORMS,
             exp.ApproxDistinct: rename_func("APPROX_COUNT_DISTINCT"),
             exp.ArraySize: rename_func("ARRAY_LENGTH"),
-            exp.AtTimeZone: lambda self, e: self.func(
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -517,6 +513,16 @@ class Generator(generator.Generator):
             "within",
         }
 
+        def attimezone_sql(self, expression: exp.AtTimeZone) -> str:
+            if not isinstance(expression.parent, exp.Cast):
+                return self.func(
+                    "TIMESTAMP", self.func("DATETIME", expression.this, expression.args.get("zone"))
+                )
+            return super().attimezone_sql(expression)
```
GeorgeSittas, line 521: This was changed because we'd get the following inside of a cast expression:

```sql
TIMESTAMP(DATETIME('YYYY-MM-DD HH24:MI:SS TZH:TZM', 'Asia/Kolkata'))
```

This is invalid syntax in the context of a `cast(.. as string [format <fmt> [at time zone <tz>]])`.

PR #1849, opened 2023-06-28
Feature!: Recursively transform children

tehClayton:
Perhaps I'm misunderstanding the purpose of the recursive transform method, but it seems like if we return the newly updated node immediately, we won't end up traversing the child nodes at all.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -536,7 +536,6 @@ class Expression(metaclass=_Expression):
             return new_node
         if new_node is not node:
             new_node.parent = node.parent
-            return new_node
 
         replace_children(new_node, lambda child: child.transform(fun, *args, copy=False, **kwargs))
         return new_node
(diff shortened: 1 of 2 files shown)


tobymao:
@barakalon i believe you first implemented transform and this behavior of not recursively transforming new nodes. 

do you remember the reason for this? in cases where i needed to do that, i started using replace_children directly instead to work around this.

GeorgeSittas:
There's a test case for this:

```python
    def test_transform_no_infinite_recursion(self):
        expression = parse_one("a")

        def fun(node):
            if isinstance(node, exp.Column) and node.name == "a":
                return parse_one("FUN(a)")
            return node

        self.assertEqual(expression.transform(fun).sql(), "FUN(a)")
```

tehClayton:
> @barakalon i believe you first implemented transform and this behavior of not recursively transforming new nodes.
> 
> do you remember the reason for this? in cases where i needed to do that, i started using replace_children directly instead to work around this.

This will work for me, happy to close this. Thanks!!

barakalon:
no me - @izeigerman : https://github.com/tobymao/sqlglot/commit/ff5d7f1b9e17201966043cf81e8106fbbec023d9

PR #1865, opened 2023-06-30
Feat(makefile): add rule to skip integration tests

GeorgeSittas:


Makefile
```diff
@@ -12,10 +12,15 @@ install-pre-commit:
 test:
 	python -m unittest
 
-style:
+test-unit:
```
tobymao, line 15:
```suggestion
unit:
```

Makefile
```diff
@@ -12,10 +12,15 @@ install-pre-commit:
...
+	SKIP_INTEGRATION=1 python -m unittest
+
+lint:
```
tobymao, line 18: revert it back to style

Makefile
```diff
@@ -12,11 +12,16 @@ install-pre-commit:
 test:
 	python -m unittest
 
+test-unit:
```
tobymao, line 15: can you change this one to unit

Makefile
```diff
@@ -12,11 +12,16 @@ install-pre-commit:
...
+	SKIP_INTEGRATION=1 python -m unittest
+
 style:
 	pre-commit run --all-files
 
 check: style test
 
+unit: style test-unit
```
tobymao, line 23: we don't need this one

GeorgeSittas: gotcha getting rid of it in a sec

Makefile
GeorgeSittas: Do we want linter checks too, or just keep it simple?

PR #1871, opened 2023-06-30
Fix(mysql): improve parsing of INSERT .. SELECT statement

### LINKED ISSUE #1862: mysql parser not support  some INSERT operations
1. mysql parser not support  `INSERT ... ON DUPLICATE KEY UPDATE Statement`
example sql followed
> INSERT INTO t1 (a,b,c) VALUES (1,2,3),(4,5,6) ON DUPLICATE KEY UPDATE c=VALUES(a)+VALUES(b);

ref: https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html

2. mysql parse not support `INSERT IGNORE INTO xxx `

ref : https://dev.mysql.com/doc/refman/5.7/en/insert-select.html

GeorgeSittas:
Fixes #1862

References:
- https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html
- https://dev.mysql.com/doc/refman/5.7/en/insert-select.html
- https://www.mysqltutorial.org/mysql-insert-ignore/

### DIFF
--- a/tests/dialects/test_mysql.py
+++ b/tests/dialects/test_mysql.py
@@ -10,7 +10,19 @@ class TestMySQL(Validator):
         self.validate_identity("UPDATE items SET items.price = 0 WHERE items.id >= 5 LIMIT 10")
         self.validate_identity("DELETE FROM t WHERE a <= 10 LIMIT 10")
         self.validate_identity(
-            "INSERT INTO x VALUES (1, 'a', 2.0) ON DUPLICATE KEY UPDATE SET x.id = 1"
+            "INSERT IGNORE INTO subscribers (email) VALUES ('john.doe@gmail.com'), ('jane.smith@ibm.com')"
+        )
+        self.validate_identity(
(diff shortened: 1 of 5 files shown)


tests/dialects/test_mysql.py
```diff
@@ -10,7 +10,19 @@ def test_ddl(self):
         self.validate_identity("UPDATE items SET items.price = 0 WHERE items.id >= 5 LIMIT 10")
         self.validate_identity("DELETE FROM t WHERE a <= 10 LIMIT 10")
         self.validate_identity(
-            "INSERT INTO x VALUES (1, 'a', 2.0) ON DUPLICATE KEY UPDATE SET x.id = 1"
```
GeorgeSittas, line 13: This test was added in https://github.com/tobymao/sqlglot/commit/5df76cee6 and is invalid in MySQL (tested locally with MySQL version 8). This also checks out with the examples in their documentation page for `INSERT ... SELECT`.

PR #1880, opened 2023-07-03
Fix(mysql): add support more kind for MEMBER OF content

brosoul:
The content type of 'Member of' is not just a string

image


GeorgeSittas (approved):
Let's add another test for the `MEMBER OF(custinfo->'$.zipcode')` form; it should be transpiled to `JSON_EXTRACT`.

PR #1881, opened 2023-07-03
Feat!(mysql): add support for multi-table DELETE syntax

### LINKED ISSUE #1877: Mysql parser not support Multi-Table Deletes
mysql parser not support Multi-Table Deletes
example sql followed
>  DELETE t1, t2 FROM t1 INNER JOIN t2 INNER JOIN t3 WHERE t1.id=t2.id AND t2.id=t3.id;

refs: https://dev.mysql.com/doc/refman/8.0/en/delete.html

GeorgeSittas:
Fixes #1877

Reference: https://dev.mysql.com/doc/refman/8.0/en/delete.html

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1818,12 +1818,17 @@ class Parser(metaclass=_Parser):
         return self._parse_as_command(self._prev)
 
     def _parse_delete(self) -> exp.Delete:
-        self._match(TokenType.FROM)
+        # This handles MySQL's "Multiple-Table Syntax"
+        # https://dev.mysql.com/doc/refman/8.0/en/delete.html
+        tables = None
+        if not self._match(TokenType.FROM, advance=False):
+            tables = self._parse_csv(self._parse_table) or None
 
         return self.expression(
             exp.Delete,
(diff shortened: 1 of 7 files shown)


sqlglot/parser.py
```diff
@@ -1835,7 +1840,7 @@ def _parse_update(self) -> exp.Update:
             **{  # type: ignore
                 "this": self._parse_table(alias_tokens=self.UPDATE_ALIAS_TOKENS),
                 "expressions": self._match(TokenType.SET) and self._parse_csv(self._parse_equality),
-                "from": self._parse_from(modifiers=True),
+                "from": self._parse_from(joins=True),
```
GeorgeSittas, line 1843: This is the only method where the `modifiers` argument is set to `True` in `_parse_from`. I don't think we need to actually parse _all_ modifiers here, but just the JOINs, hence the change.

FYI the only other "modifiers" in `Table` (which is parsed inside `_parse_from`) are `pivots` and `laterals`, but like I said, I doubt these would be useful in either `UPDATE` or `DELETE`. Can fix this in the future easily if it's a problem.

PR #1890, opened 2023-07-05
Feat!(spark): better coverage for decimals, strings, datetime, bin types

dmoore247:
Update test coverage, update tsql to spark mappings for more 1-to-1 mappings on INT, TINYINT, BIGINT etc.

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -50,7 +50,7 @@ class Spark(Spark2):
             exp.DataType.Type.SMALLMONEY: "DECIMAL(6, 4)",
             exp.DataType.Type.VARBINARY: "BINARY",
             exp.DataType.Type.BIT: "BOOLEAN",
-            exp.DataType.Type.UNIQUEIDENTIFIER: "BINARY",
+            exp.DataType.Type.UNIQUEIDENTIFIER: "STRING",
         }
         TRANSFORMS = Spark2.Generator.TRANSFORMS.copy()
         TRANSFORMS.pop(exp.DateDiff)
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -43,9 +43,6 @@ class Spark(Spark2):
     class Generator(Spark2.Generator):
         TYPE_MAPPING = {
             **Spark2.Generator.TYPE_MAPPING,
-            exp.DataType.Type.TINYINT: "TINYINT",
-            exp.DataType.Type.SMALLINT: "SMALLINT",
-            exp.DataType.Type.BIGINT: "BIGINT",
             exp.DataType.Type.MONEY: "DECIMAL(15, 4)",
             exp.DataType.Type.SMALLMONEY: "DECIMAL(6, 4)",
             exp.DataType.Type.UNIQUEIDENTIFIER: "STRING",
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -48,8 +48,6 @@ class Spark(Spark2):
             exp.DataType.Type.BIGINT: "BIGINT",
             exp.DataType.Type.MONEY: "DECIMAL(15, 4)",
             exp.DataType.Type.SMALLMONEY: "DECIMAL(6, 4)",
-            exp.DataType.Type.VARBINARY: "BINARY",
-            exp.DataType.Type.BIT: "BOOLEAN",
             exp.DataType.Type.UNIQUEIDENTIFIER: "STRING",
         }
         TRANSFORMS = Spark2.Generator.TRANSFORMS.copy()
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -41,6 +41,17 @@ class Spark(Spark2):
         }
 
     class Generator(Spark2.Generator):
+        TYPE_MAPPING = {
+            **Spark2.Generator.TYPE_MAPPING,
+            exp.DataType.Type.TINYINT: "TINYINT",
+            exp.DataType.Type.SMALLINT: "SMALLINT",
+            exp.DataType.Type.BIGINT: "BIGINT",
+            exp.DataType.Type.MONEY: "DECIMAL(15, 4)",
+            exp.DataType.Type.SMALLMONEY: "DECIMAL(6, 4)",
+            exp.DataType.Type.VARBINARY: "BINARY",
+            exp.DataType.Type.BIT: "BOOLEAN",
+            exp.DataType.Type.UNIQUEIDENTIFIER: "BINARY",
+        }
         TRANSFORMS = Spark2.Generator.TRANSFORMS.copy()
         TRANSFORMS.pop(exp.DateDiff)
         TRANSFORMS.pop(exp.Group)
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -262,7 +262,7 @@ class TestBigQuery(Validator):
                 "duckdb": "CAST(a AS BIGINT)",
                 "presto": "CAST(a AS BIGINT)",
                 "hive": "CAST(a AS BIGINT)",
-                "spark": "CAST(a AS LONG)",
+                "spark": "CAST(a AS BIGINT)",
             },
         )
         self.validate_all(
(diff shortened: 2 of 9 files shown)


tests/dialects/test_bigquery.py
```diff
@@ -262,7 +262,7 @@ def test_bigquery(self):
                 "duckdb": "CAST(a AS BIGINT)",
                 "presto": "CAST(a AS BIGINT)",
                 "hive": "CAST(a AS BIGINT)",
-                "spark": "CAST(a AS LONG)",
```
tobymao, line 265: this is great can you also add the spark2 test

sqlglot/dialects/spark.py
```diff
@@ -41,6 +41,17 @@ class Parser(Spark2.Parser):
...
     class Generator(Spark2.Generator):
+        TYPE_MAPPING = {
+            **Spark2.Generator.TYPE_MAPPING,
+            exp.DataType.Type.TINYINT: "TINYINT",
+            exp.DataType.Type.SMALLINT: "SMALLINT",
+            exp.DataType.Type.BIGINT: "BIGINT",
+            exp.DataType.Type.MONEY: "DECIMAL(15, 4)",
+            exp.DataType.Type.SMALLMONEY: "DECIMAL(6, 4)",
+            exp.DataType.Type.VARBINARY: "BINARY",
+            exp.DataType.Type.BIT: "BOOLEAN",
```
tobymao, line 52: is this and uniqueidentifier correct?

sqlglot/dialects/spark.py
```diff
@@ -41,6 +41,17 @@ class Parser(Spark2.Parser):
         }
 
     class Generator(Spark2.Generator):
+        TYPE_MAPPING = {
+            **Spark2.Generator.TYPE_MAPPING,
+            exp.DataType.Type.TINYINT: "TINYINT",
+            exp.DataType.Type.SMALLINT: "SMALLINT",
+            exp.DataType.Type.BIGINT: "BIGINT",
+            exp.DataType.Type.MONEY: "DECIMAL(15, 4)",
+            exp.DataType.Type.SMALLMONEY: "DECIMAL(6, 4)",
```
tobymao, line 50: how did you figure out this mapping?

GeorgeSittas: Not sure if you're asking about the `MONEY` types Toby, but I'm leaving this for reference, regarding the `INT` types:

```
Time taken: 4.181 seconds, Fetched 1 row(s)
spark-sql (default)> select cast(1 as byte);
CAST(1 AS TINYINT)
1
Time taken: 0.234 seconds, Fetched 1 row(s)
spark-sql (default)> select cast(1 as short);
CAST(1 AS SMALLINT)
1
Time taken: 0.229 seconds, Fetched 1 row(s)
spark-sql (default)> select cast(1 as long);
CAST(1 AS BIGINT)
1
Time taken: 0.207 seconds, Fetched 1 row(s)
```

Seems like Spark 3 uses `TINYINT`, `SMALLINT` and `BIGINT` instead of `BYTE`, `SHORT` and `LONG`, respectively, which are the current types as defined in the Spark2 dialect.

sqlglot/dialects/spark.py
```diff
@@ -41,6 +41,17 @@ class Parser(Spark2.Parser):
...
+            exp.DataType.Type.VARBINARY: "BINARY",
```
GeorgeSittas, line 51: This is redundant, `VARBINARY` is already mapped to `"BINARY"` in the Hive dialect, which is a superclass.

sqlglot/dialects/spark.py
GeorgeSittas: The `BIT` entry is redundant for the same reason as `VARBINARY`. The `UNIQUEIDENTIFIER` entry doesn't seem correct per T-SQL's docs: https://learn.microsoft.com/en-us/sql/t-sql/data-types/uniqueidentifier-transact-sql?view=sql-server-ver16?

GeorgeSittas (commented):
Thanks, left a couple of comments.

Maybe this should be marked as a breaking change since it'll change SQL generation for Spark.

dmoore247:
Thanks @tobymao  and @GeorgeSittas 
Happy to incorporate feedback. The mappings are based on research our team has done.
As you may know, there is an inheritance along this lineage: Hive -> Spark2 -> Spark -> Databricks
Let me look into exactly when `TINYINT`, `SMALLINT` and `BIGINT` were introduced, they could have been introduced mid spark2...

Another note, I may want to update the `REAL` as that reportedly is the same as `FLOAT(24)`

sqlglot/dialects/spark.py
dmoore247: Double checking on our recommendations for `UNIQUEIDENTIFIER`

tests/dialects/test_bigquery.py
dmoore247: In looking through the rest of the `test_bigquery.py` all the tests are based on `spark` rather than spark2. Any updates for spark2 should probably be applied in a separate pull request. Would you agree?

sqlglot/dialects/spark.py
dmoore247: @GeorgeSittas I see now you're saying VARBINARY and BIT are redundant with the HIVE definition from which SPARK inherits. They are removed now.

GeorgeSittas:
Btw @dmoore247 seems like `TINYINT`, `SMALLINT` and `BIGINT` were also available in Spark >= v2.4.8:

```
(env) ➜  spark-sql
...
23/07/06 16:18:51 INFO SparkContext: Running Spark version 2.4.8
...
spark-sql> select cast(1 as tinyint);
1
Time taken: 3.107 seconds, Fetched 1 row(s)
spark-sql> select cast(1 as smallint);
1
Time taken: 0.213 seconds, Fetched 1 row(s)
spark-sql> select cast(1 as bigint);
1
Time taken: 0.189 seconds, Fetched 1 row(s)
```

So, we can move these `TYPE_MAPPING` changes to the `Spark2` dialect.

tests/dialects/test_bigquery.py
GeorgeSittas: If you move the changes to the Spark2 dialect, then maybe testing "spark" suffices, since it inherits from Spark2

sqlglot/dialects/spark.py
GeorgeSittas: It looks like people [consider](https://stackoverflow.com/questions/582797/should-you-choose-the-money-or-decimalx-y-datatypes-in-sql-server) `DECIMAL(19, 4)` and `DECIMAL(10, 4)` for `MONEY` and `SMALLMONEY` accordingly.

dmoore247:
Reference for MONEY types:
"The money and smallmoney data types are accurate to a ten-thousandth of the monetary units that they represent."
https://learn.microsoft.com/en-us/sql/t-sql/data-types/money-and-smallmoney-transact-sql?view=sql-server-ver16

dmoore247:
@GeorgeSittas Ironically, moving the TYPE_MAPPING changes for INT, TINYINT, SMALLINT to spark2 essentially means removing them from spark2 reverting back to HIVE mappings for them.

GeorgeSittas:
> Reference for MONEY types: "The money and smallmoney data types are accurate to a ten-thousandth of the monetary units that they represent." https://learn.microsoft.com/en-us/sql/t-sql/data-types/money-and-smallmoney-transact-sql?view=sql-server-ver16

Gotcha, so it seems like `DECIMAL(15, 4)` and `DECIMAL(6, 4)` will do the trick for T-SQL -> Spark. What about other dialects, though? For example Postgres [seems](https://www.postgresql.org/docs/current/datatype-money.html) to support -92233720368547758.08 to +92233720368547758.07.

Maybe a broader `DECIMAL` type will be more robust?

dmoore247:
I found a related issue, FLOAT(n) in tsql maps to either FLOAT or DOUBLE or some decimal/numeric in Spark.
Currently, the mapping produces a FLOAT(n) in spark which isn't supported by spark.
Should we address here or later in another pull request?

GeorgeSittas:
> I found a related issue, FLOAT(n) in tsql maps to either FLOAT or DOUBLE or some decimal/numeric in Spark. Currently, the mapping produces a FLOAT(n) in spark which isn't supported by spark. Should we address here or later in another pull request?

Let's address this in another PR.

I'm going to merge this one and do the final minor cleanups. Do you wanna make a PR for the issue with `FLOAT(n)`? If spark doesn't allow a type length, maybe we can just drop it.

GeorgeSittas (approved):
LGTM - one final question about the precision of `DECIMAL` in other dialects and should be good to go.

Thanks @dmoore247!

dmoore247:
@GeorgeSittas I could tackle the FLOAT(n) in another PR. Any directional advice on where to apply the change?

GeorgeSittas:
>  If I understand your comment, you're asking if these MONEY mappings in TSQL to other dialects should be generalized to the other dialects?

So if we want to transpile `CAST(... AS MONEY)` from T-SQL to Spark, then using `DECIMAL(15, 4)` will suffice. But since `DataType.Type.MONEY` also represents the `MONEY` type for e.g. the Postgres dialect, we need to make sure that transpiling the above cast from Postgres to Spark doesn't mess up the intended precision.

> I could tackle the FLOAT(n) in another PR. Any directional advice on where to apply the change?

If I understood correctly, the issue is that Spark (v2 or v3) doesn't support a precision for the `FLOAT` type? If so, then we could drop the precision in the Spark2 (maybe Hive?) dialect by updating the `Spark2.Generator.cast_sql` override.

dmoore247:
My override only applies to Spark, right? Perhaps we can have a quick slack chat or call?

GeorgeSittas:
> My override only applies to Spark, right? Perhaps we can have a quick slack chat or call?

That's correct. Even if the AST for `FLOAT(5)` contains the precision information (i.e. a `DataTypeSize` node), we should ignore it at SQL generation time.

PR #1894, opened 2023-07-05
Fix!(parser, optimizer): improve parsing, optimizing of nested tables, joins

### LINKED ISSUE #1879: Using optimizer with nested joins
When using a nested join, where table b and c need to be joined before making the join to table a is not working without using parenthesis, as it will result in a parsing error. This has been filed as separate bug. However when using parenthesis at this nested join, the query is actually wrongly parsed. As the parenthesis is currently parsed as a subquery, while it is not. 

When using the optimizer on this AST, an error occurs `OptimizeError: Column 'c.id' could not be resolved for table: 'c'`, this is due to the fact that table alias c does not exist anymore if you parse the parenthesis as a subquery. 

```
import sqlglot
from sqlglot import optimizer
expr = sqlglot.parse_one("""
SELECT
 a.id
,b.test
,c.test2
FROM a
LEFT JOIN (
    b
    INNER JOIN c
        ON c.id = b.id
)
    ON b.id = a.id
""", read="tsql")

schema={
    "testcat": {
        "testdb": {
            "a": {
                "id": "INT",
            },
            "b": {
                "id": "INT",
                "test": "STRING",
            },
            "c": {
                "id": "INT",
                "test2": "STRING",
            },
        }
    }
}

expr = optimizer.optimize(
    expression=expr,
    schema=schema,
)
```

GeorgeSittas:
Fixes #1879


sqlglot/optimizer/qualify_tables.py
```diff
@@ -66,13 +78,20 @@ def qualify_tables(
...
                         source.set("catalog", exp.to_identifier(catalog))
 
+                # Unnest parenthesized table
+                if isinstance(source.parent, exp.Paren):
+                    source.parent.replace(source)
+
+                # Unnest joins attached in tables by appending them to closest query
+                for join in source.args.get("joins") or []:
+                    scope.expression.append("joins", join)
+                else:
```
tobymao, line 88: you don't need an else here, it doesn't change anything

GeorgeSittas: Thanks, got rid of it.

sqlglot/expressions.py
```diff
@@ -274,12 +274,16 @@ def append(self, arg_key: str, value: t.Any) -> None:
...
+        Sets arg_key to value.
 
         Args:
-            arg_key (str): name of the expression arg.
+            arg_key: name of the expression arg.
             value: value to set the arg to.
         """
+        if value is None:
+            self.args.pop(arg_key, None)
+            return
```
GeorgeSittas, line 285: I noticed the following unintuitive (imo) behaviour when setting an arg to `None`, which motivated this change:

```python
>>> import sqlglot
>>> expr = sqlglot.parse_one("select * from foo join bar")
>>> expr.set("joins", None)
>>> type(expr.args.get("joins", []))
<class 'NoneType'>
```

The issue here is that setting a dict entry to `None` does not have any additional meaning like "deleting the expression". It simply results in a dict entry being set to `None`, so in this case `"joins" -> None` and as shown above this doesn't play well with the semantics of the `get` method on dicts.

Do we ever want to maintain a `None` arg in an expression without deleting it? If an expression is optional, we'll use the `.get(...)` call anyway and it'll return `None`.

FYI: this led to some test failures because we use the `for join in expr.args.get("joins", [])` pattern in the codebase.

GeorgeSittas: Perhaps we could also add an optional `arg_key` parameter to the `pop` method as an alternative? It currently works only on expressions, so we can't do `expr.args["joins"].pop()`, but we could instead have `expr.pop("joins")`

tobymao: that's why you shouldn't do get("", []) and always do get("") or []

sqlglot/parser.py
```diff
@@ -1969,9 +1969,32 @@ def _parse_select(
...
+                if not alias:
+                    return exp.paren(this, copy=False)
+
+                joins = this.args.get("joins")
+                this.set("joins", None)
+
+                this = this.replace(exp.select("*").from_(this.copy()))
+                this.set("joins", joins)
+
+                self._retreat(index)
```
tobymao, line 1994: why are you retreating?

GeorgeSittas: The idea was to let the alias be parsed in `_parse_subquery` later, but seeing it again I think we can avoid retreating and attach it manually to whatever `_parse_subquery` returns.

GeorgeSittas: ^ Implemented this one

sqlglot/parser.py
```diff
@@ -1969,10 +1969,35 @@ def _parse_select(
...
+            # a star query that selects from it.
+            if isinstance(this, (exp.Table, exp.Paren)):
+                # Removes redundant parentheses, e.g. in "SELECT * FROM ((tbl1 JOIN tbl2)) AS t"
+                this = t.cast(exp.Expression, this.unnest())
+
+                alias = self._parse_table_alias()
+                if not alias:
+                    return exp.paren(this, copy=False)
+
+                joins = this.args.get("joins")
```
tobymao, line 1988: joins = this.args.pop("joins", None), don't need the set right?

sqlglot/parser.py
```diff
@@ -1969,10 +1969,35 @@ def _parse_select(
...
+                this.set("joins", None)
+
+                this = this.replace(exp.select("*").from_(this.copy()))
```
tobymao, line 1991: from_ copy = False ? why do you need to call this.copy

tests/fixtures/optimizer/qualify_tables.sql
```diff
@@ -19,25 +19,46 @@ SELECT (SELECT y.c FROM c.db.y AS y) FROM c.db.x AS x;
...
+# title: redundant parentheses (1)
+SELECT * FROM (tbl AS tbl);
+SELECT * FROM c.db.tbl AS tbl;
+
+# title: redundant parentheses (2)
+SELECT * FROM ((((tbl AS tbl))));
+SELECT * FROM c.db.tbl AS tbl;
+
+# title: redundant parentheses around join construct
+SELECT * FROM a LEFT JOIN (b INNER JOIN c ON c.id = b.id) ON b.id = a.id;
```
tobymao, line 35: do we also handle

select from ( (b join c) join d)

so join d is not on a table, but on a paren / subquery?

sqlglot/parser.py
GeorgeSittas: You need the set too because in `SELECT * FROM (t1 JOIN t2) t` the join is attached in `t1` and so we need to move that over to the produced `Select` expression that replaces `t1`

sqlglot/parser.py
GeorgeSittas: > why do you need to call this.copy

Because otherwise we get an infinite loop.

Without the copy you first construct an AST where `this`'s parent becomes the `FROM` and then you try to replace `this` with the produced `Select` expression, which contains a reference to it.

tests/fixtures/optimizer/qualify_tables.sql
GeorgeSittas: Nope, parser error, good catch! Working on a fix

tobymao: can you check if it's valids ql?

GeorgeSittas: Already did, it runs on postgres:

```
georgesittas=# with b as (select 1), c as (select 1), d as (select 1) select * from ( (b cross join c) cross join d);
 ?column? | ?column? | ?column?
----------+----------+----------
        1 |        1 |        1
(1 row)
```

GeorgeSittas: Tried to resolve this in https://github.com/tobymao/sqlglot/pull/1894/commits/a481a6c62de037f7000470b81156bf90b3bdff7f

sqlglot/parser.py
```diff
@@ -1969,10 +1969,31 @@ def _parse_select(
...
+            # names, and so we want to parse it as a Subquery to represent the inner scope appropriately.
+            # Additionally, we want the node under the Subquery to be an actual query, so we will replace
+            # the table reference with a star query that selects from it.
+            if isinstance(this, exp.Table):
+                alias = self._parse_table_alias()
+                if not alias:
+                    this.set("wrapped", True)
+                    return this
+
+                this.set("wrapped", None)
```
tobymao, line 1985: why is this needed?

tobymao (approved):

PR #1895, opened 2023-07-05
Fix(duckdb): improve transpilation of BigQuery DATE function

GeorgeSittas:
References:
- https://duckdb.org/docs/sql/functions/date.html
- https://duckdb.org/docs/sql/functions/dateformat.html
- https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -41,20 +41,15 @@ def _date_delta_sql(self: generator.Generator, expression: exp.DateAdd | exp.Dat
 
 # BigQuery -> DuckDB conversion for the DATE function
 def _date_sql(self: generator.Generator, expression: exp.Date) -> str:
-    expressions = expression.expressions
+    result = f"CAST({self.sql(expression, 'this')} AS DATE)"
+    zone = self.sql(expression, "zone")
 
-    timelike = self.sql(expressions[0])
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -39,6 +39,26 @@ def _date_delta_sql(self: generator.Generator, expression: exp.DateAdd | exp.Dat
...
+
+    if len(expressions) == 2:
+        zone = self.sql(expressions[1])
+        date_str = self.func("STRFTIME", result, "'%d/%m/%Y'")
+        date_str = f"{date_str} || ' ' || {zone}"
+
+        # This will create a TIMESTAMP with time zone information
+        result = self.func("STRPTIME", date_str, "'%d/%m/%Y %Z'")
+    elif len(expressions) > 2:
+        self.unsupported("Unsupported number of arguments for DATE")
```
GeorgeSittas, line 57: I wonder if we should handle the case where the length is equal to 3 here. Since BigQuery parses such a call to `DATE` as an `exp.DateFromParts` node, it shouldn't ever happen, but if a user manually creates an `exp.Date` node with 3 arguments instead of using `exp.DateFromParts`, they'll hit this unsupported error.

tobymao (approved):

PR #1896, opened 2023-07-06
feat: Add FLOAT(n) mappings to Hive

dmoore247:
Hive/Spark2/Spark don't support FLOAT(n) data type.
This feat maps FLOAT(n) to float if n <= 32 and double otherwise

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -414,6 +414,16 @@ class Hive(Dialect):
             return self.properties(properties, prefix=self.seg("TBLPROPERTIES"))
 
         def datatype_sql(self, expression: exp.DataType) -> str:
+            if expression.this == (exp.DataType.Type.FLOAT):
+                # Not dealing with n>64
+                dt = expression.find(exp.DataTypeSize)
+                if dt:
+                    size = int(dt.name)
+                    if size <= 32:
+                        expression = exp.DataType.build("float")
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/hive.py
```diff
@@ -414,6 +414,16 @@ def with_properties(self, properties: exp.Properties) -> str:
...
+            if expression.this == (exp.DataType.Type.FLOAT):
+                # Not dealing with n>64
+                dt = expression.find(exp.DataTypeSize)
+                if dt:
+                    size = int(dt.name)
+                    if size <= 32:
+                        expression = exp.DataType.build("float")
+                    else:
+                        expression = exp.DataType.build("double")
+
```
GeorgeSittas, line 426: I think I'd write this like:

```python
...
elif expression.is_type("float"):
    size_expression = expression.find(exp.DataTypeSize)
    if size_expression:
        size = int(size_expression.name)
        expression = (
            exp.DataType.build("float") if size <= 32 else exp.DataType.build("double")
        )
```

GeorgeSittas: The comment seems kinda unnecessary here too.

GeorgeSittas (commented):
Looks good, left a couple of comments.

The tests can probably be condensed a bit without losing coverage, but seems ok either way.

PR #1900, opened 2023-07-08
Fix!(optimizer): preserve parenthesized joins, tables

### LINKED ISSUE #1898: Optimizer transforms nested joins incorrectly
Hey @GeorgeSittas, thanks a lot for the fix on the nested joins. That will work on my part, as i only need to be able to parse and use the scope class. However i did a check on the optimizer as well and the SQL seems to be transformed into SQL that is functionally different from the original.

As you can see in the output the query is beeing transformed into regular joins and the table b and c arent nested anymore.

The output:
```
SELECT
  "a"."id" AS "id",
  "b"."test" AS "test",
  "c"."test2" AS "test2"
FROM "a" AS "a"
LEFT JOIN "b" AS "b"
  ON "b"."id" = "a"."id"
JOIN "c" AS "c"
  ON "c"."id" = "b"."id"
```

The code i used:
```
import sqlglot
from sqlglot import optimizer
expr = sqlglot.parse_one("""
SELECT
 a.id
,b.test
,c.test2
FROM a
LEFT JOIN (
    b
    INNER JOIN c
        ON c.id = b.id
)
    ON b.id = a.id
""", read="tsql")

schema={
    "testcat": {
        "testdb": {
            "a": {
                "id": "INT",
            },
            "b": {
                "id": "INT",
                "test": "STRING",
            },
            "c": {
                "id": "INT",
                "test2": "STRING",
            },
        }
    }
}

expr = optimizer.optimize(
    expression=expr,
    schema=schema,
)

postgres = sqlglot.Dialect.get_or_raise("postgres")()
print(postgres.generate(expr, pretty=True))
```

GeorgeSittas:
Fixes #1898


sqlglot/optimizer/scope.py
```diff
@@ -353,10 +357,27 @@ def selects(self):
...
+
+        For example, for the following expression:
+            SELECT 1 as a, 2 as b FROM x
+
+        The outputs are the "a" and " b" expressions.
+
+        Returns:
+            list[exp.Expression]: expressions
+        """
+        if isinstance(self.expression, exp.Table):
```
tobymao, line 377: just add named_selects to table directly

tobymao:
just to confirm qualify columns works with this approach?

sqlglot/optimizer/qualify_tables.py
```diff
@@ -66,19 +53,9 @@ def qualify_tables(
                     if not source.args.get("catalog"):
                         source.set("catalog", exp.to_identifier(catalog))
```
tobymao, line 54: so it's not necessary anymore to do the select * on subqueries?

sqlglot/optimizer/scope.py
```diff
@@ -353,10 +357,27 @@ def selects(self):
         Returns:
             list[exp.Expression]: expressions
         """
-        if isinstance(self.expression, exp.Union):
```
tobymao, line 356: why'd you get rid of the union?

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -701,3 +701,76 @@ SELECT
     "x"."a" * "x"."b" - "x"."b"
   ) AS "f"
 FROM "x" AS "x";
+
+# title: wrapped table without alias
+# execute: false
+SELECT * FROM (tbl);
```
tobymao, line 707: can you add tests that don't just do select *?

tests/test_parser.py
```diff
@@ -582,56 +582,26 @@ def test_parse_floats(self):
...
-        expr = parse_one("select * from ((a cross join ((b cross join c) cross join d)))")
-        self.assertEqual(expr.sql(), "SELECT * FROM (a CROSS JOIN (b CROSS JOIN c CROSS JOIN d))")
-
-        expr = parse_one(
-            "select * from ((a cross join ((b cross join c) cross join (d cross join e))))"
-        )
-        self.assertEqual(
-            expr.sql(), "SELECT * FROM (a CROSS JOIN (b CROSS JOIN c CROSS JOIN (d CROSS JOIN e)))"
-        )
+        subquery = parse_one("select * from (table)").args["from"].this
```
tobymao, line 585: these tests don't seem useful, you could probably just delete them if they are in identity.sql

sqlglot/optimizer/scope.py
GeorgeSittas: Because `Union.selects` (and `Union.named_selects`) are already implemented like this:

https://github.com/tobymao/sqlglot/blob/3b215adc413772bc1af46a67e2410b38ad8872a2/sqlglot/expressions.py#L2401-L2402

tests/fixtures/optimizer/optimizer.sql
GeorgeSittas: Yep still have some testing to do for sure

tests/test_parser.py
GeorgeSittas: Sure I can get rid of them

sqlglot/optimizer/qualify_tables.py
GeorgeSittas: Looks like it, but I need to check more examples to be sure + run the full optimizer on them to see what's up.

GeorgeSittas:
> just to confirm qualify columns works with this approach?

There are probably more bugs I need to resolve, I'll try to get additional tests in for e.g. qualify_columns.

sqlglot/optimizer/scope.py
GeorgeSittas: should we get rid of both `scope.selects` and `scope.named_selects` then for consistency? we can just do `scope.expression.[named_]selects` in places where we need it

sqlglot/optimizer/qualify_tables.py
GeorgeSittas: I'll add it back again because it takes care of some issues in qualify columns I discovered.

sqlglot/optimizer/scope.py
tobymao: yea

tobymao: we can get rid

GeorgeSittas: got rid of 'em both

tests/fixtures/optimizer/optimizer.sql
GeorgeSittas: Added some more tests but want to add even more, not sure if I covered all interesting cases.

sqlglot/optimizer/isolate_table_selects.py
```diff
@@ -12,7 +12,12 @@ def isolate_table_selects(expression, schema=None):
             continue
 
         for _, source in scope.selected_sources.values():
-            if not isinstance(source, exp.Table) or not schema.column_names(source):
+            if (
+                not isinstance(source, exp.Table)
+                or not schema.column_names(source)
+                or isinstance(source.parent, exp.Subquery)
+                or isinstance(source.parent.parent, exp.Table)
```
GeorgeSittas, line 19: The first added instance check ensures we don't expand `t` in `(t)` into a star query, whereas the second one ensures we similarly don't expand `t2` in `(t1 join t2)`.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -97,8 +97,11 @@ def merge_derived_tables(expression, leave_tables_isolated=False):
         for subquery in outer_scope.derived_tables:
             from_or_join = subquery.find_ancestor(exp.From, exp.Join)
             alias = subquery.alias_or_name
-            inner_scope = outer_scope.sources[alias]
-            if _mergeable(outer_scope, inner_scope, leave_tables_isolated, from_or_join):
+            inner_scope = outer_scope.sources.get(alias)
+
+            if inner_scope and _mergeable(
+                outer_scope, inner_scope, leave_tables_isolated, from_or_join
+            ):
```
GeorgeSittas, line 104: The invariant "every source has an alias" does not hold anymore, since you can have queries like:

```sql
SELECT * FROM (t1 JOIN t2)
```

There's a single scope here, since there's no alias attached to the nested source `(t1 JOIN t2)`.

sqlglot/optimizer/qualify_tables.py
```diff
@@ -66,19 +65,9 @@ def qualify_tables(
...
-                source.set("wrapped", None)
-
                 if not source.alias:
-                    source = source.replace(
-                        alias(
-                            source, name or source.name or next_alias_name(), copy=True, table=True
-                        )
-                    )
+                    # Mutates source by attaching an alias to it
+                    alias(source, name or source.name or next_alias_name(), copy=False, table=True)
```
GeorgeSittas, line 70: This change was important because we may have a nested join like `(tbl1 JOIN tbl2)`, where `tbl1` is an ancestor of `tbl2` in the AST. Since we want to treat both tables as sources, we will most likely get a situation where `tbl1` is replaced first and so when the loop visits `tbl2` its reference will be dangling and the alias addition will not be reflected in the resulting AST.

For example, this is what would happen due to this `replace` call:

```
SELECT * FROM (t1 CROSS JOIN t2) --> SELECT * FROM (t1 AS t1 CROSS JOIN t2)
```

sqlglot/optimizer/scope.py
```diff
@@ -122,7 +122,11 @@ def _collect(self):
                 self._udtfs.append(node)
             elif isinstance(node, exp.CTE):
                 self._ctes.append(node)
-            elif isinstance(node, exp.Subquery) and isinstance(parent, (exp.From, exp.Join)):
+            elif (
+                isinstance(node, exp.Subquery)
+                and isinstance(parent, (exp.From, exp.Join))
+                and _is_subquery_scope(node)
+            ):
```
GeorgeSittas, line 129: Since nested joins such as `(t1 JOIN t2)` don't introduce a new scope, `walk` will visit them too, so we want to make sure they won't be treated as derived tables.

sqlglot/optimizer/scope.py
```diff
@@ -274,6 +278,7 @@ def columns(self):
                     not ancestor
                     or column.table
                     or isinstance(ancestor, exp.Select)
+                    or (isinstance(ancestor, exp.Table) and not isinstance(ancestor.this, exp.Func))
```
GeorgeSittas, line 281: The check for `Func` here was necessary because the following test was failing without it:

```sql
-- `hive_partition` is qualified without the check
SELECT x FROM READ_PARQUET('path.parquet', hive_partition=1);
```

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -701,3 +701,108 @@ SELECT
...
+  "z"."c" AS "c",
+  "x"."a" AS "a",
+  "x"."b" AS "b"
+FROM "x" AS "x"
+LEFT JOIN (
+  "y" AS "y"
+    JOIN "z" AS "z"
+      ON "y"."c" = "z"."c"
+)
+  ON "x"."b" = "y"."b";
```
GeorgeSittas, line 808: This query is not executed because there's a discrepancy between the schema used for optimization and the one used for execution.

sqlglot/optimizer/merge_subqueries.py
GeorgeSittas: Update on this: the changes here were not needed after all because I "accidentally" solved the problem by not treating `(t1 join t2)` as a derived table.

sqlglot/optimizer/isolate_table_selects.py
tobymao: but isn't t2 a table? why would it be triggered? same with the second case, the second case 

t1 join t2, it's a table right?

sqlglot/optimizer/qualify_tables.py
```diff
@@ -15,26 +15,18 @@ def qualify_tables(
...
+    (t1 JOIN t2) AS t will be expanded into (SELECT * FROM t1 AS t1, t2 AS t2) AS t.
 
     Examples:
         >>> import sqlglot
         >>> expression = sqlglot.parse_one("SELECT 1 FROM tbl")
         >>> qualify_tables(expression, db="db").sql()
         'SELECT 1 FROM db.tbl AS tbl'
         >>>
-        >>> expression = sqlglot.parse_one("SELECT * FROM (tbl)")
+        >>> expression = sqlglot.parse_one("SELECT 1 FROM (t1 JOIN t2) AS t")
```
tobymao, line 27: why did you change this to 1 from *?

GeorgeSittas: no important reason, just to be consistent with the previous example in the docstring

tobymao (approved):

PR #1904, opened 2023-07-10
Fix(postgres): correctly parse jsonb[]

### LINKED ISSUE #1903: Postgres: arrays of JSONB raise parsing errors
**Before you file an issue**
[x] Make sure you specify the "read" dialect eg. parse_one(sql, read="spark")
[x] Check if the issue still exists on main 

```
import sqlglot
sqlglot.parse("CREATE TABLE test ( elems jsonb[] )", read="postgres")
```

raises:
```
ParseError: Failed to parse 'JSONB' into <class 'sqlglot.expressions.DataType'>
```

**Official Documentation**

https://www.postgresql.org/docs/current/arrays.html

vmax:
Fixes #1903

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3033,6 +3033,9 @@ class Parser(metaclass=_Parser):
             exp.DataTypeSize, this=this, expression=self._parse_var(any_token=True)
         )
 
+    def _build_datatype(self, type_token, expressions) -> exp.DataType:
+        return exp.DataType.build(type_token.value, expressions=expressions)
+
     def _parse_types(
         self, check_func: bool = False, schema: bool = False
     ) -> t.Optional[exp.Expression]:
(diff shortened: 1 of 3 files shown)


tobymao:
thanks for the PR, but this is not the right fix, we should remove use of DataType.build in the parser

sqlglot/parser.py
```diff
@@ -3033,6 +3033,9 @@ def _parse_type_size(self) -> t.Optional[exp.DataTypeSize]:
             exp.DataTypeSize, this=this, expression=self._parse_var(any_token=True)
         )
 
+    def _build_datatype(self, type_token, expressions) -> exp.DataType:
```
tobymao, line 3036: delete this

vmax:
@tobymao what would be the correct way to make created `exp.DataType` dialect-aware? for example:

```
sqlglot.parse_one('JSONB', into=sqlglot.expressions.DataType)
```

raises `ParseError: Failed to parse 'JSONB' into <class 'sqlglot.expressions.DataType'>`

whereas

```
sqlglot.parse_one('JSONB', into=sqlglot.expressions.DataType, read='postgres')
```

correctly parses:
```
(DATATYPE this: Type.JSONB, nested: False, prefix: False)
```

GeorgeSittas:
What do you mean "to make created exp.DataType dialect-aware"? How / why would you want it to be dialect-aware?

In general, SQLGlot expressions (AST nodes) are deliberately not dialect-aware because that simplifies transpilation. So you can think of it as trying to parse each query into a canonical AST and then map the AST back to the correct concepts based on the output dialect.

tobymao:
superseded by 6ccb5956

PR #1907, opened 2023-07-10
Feat(tsql): improve support for transaction statements

dmoore247:
Enrich syntax parsing for BEGIN, COMMIT and ROLLBACK Transactions.


sqlglot/dialects/tsql.py
```diff
@@ -281,8 +282,16 @@ class Tokenizer(tokens.Tokenizer):
         QUOTES = ["'", '"']
         HEX_STRINGS = [("0x", ""), ("0X", "")]
 
+        COMMAND_PREFIX_TOKENS = {TokenType.SEMICOLON}
+
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
+            "BEGIN": TokenType.COMMAND,
+            "BEGIN TRANSACTION": TokenType.BEGIN,
+            "COMMIT": TokenType.COMMIT,
```
tobymao, line 291: these are unnecessary, why did you add them?

sqlglot/dialects/tsql.py
```diff
@@ -358,13 +370,75 @@ class Parser(parser.Parser):
         STATEMENT_PARSERS = {
             **parser.Parser.STATEMENT_PARSERS,
             TokenType.END: lambda self: self._parse_command(),
+            TokenType.ALIAS: lambda self: self._parse_alias(),
```
tobymao, line 373: why did you add this

sqlglot/dialects/tsql.py
```diff
@@ -358,13 +370,75 @@ class Parser(parser.Parser):
...
+
+            Returns:
+                exp.Commit | exp.Rollback: _description_
+            """
+            rollback = False
+            if self._prev.token_type == TokenType.ROLLBACK:
+                rollback = True
+
+            transaction = None
+            if self._match_texts({"TRAN", "TRANSACTION"}):
```
tobymao, line 401: transaction = self._match_texts and self._prev.text

sqlglot/dialects/tsql.py
```diff
@@ -358,13 +370,75 @@ class Parser(parser.Parser):
...
+                transaction = self._prev.text
+            txn_name = self._parse_id_var()
+
+            durability = None
+            if self._match_text_seq("WITH", "(", "DELAYED_DURABILITY", "="):
```
tobymao, line 406: don't do this, use real matching

self.match(TokenType.With):
self .match_l_paren() etc

sqlglot/expressions.py
```diff
@@ -3960,7 +3966,7 @@ def else_(self, condition: ExpOrStr, copy: bool = True, **opts) -> Case:
 
 
 class Cast(Func):
-    arg_types = {"this": True, "to": True, "format": False}
+    arg_types = {"this": True, "to": False, "format": False}
```
tobymao, line 3969: why did you change this, this is probably not correct

sqlglot/parser.py
```diff
@@ -1205,7 +1205,8 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
             extend_props(self._parse_properties())
 
             self._match(TokenType.ALIAS)
-            begin = self._match(TokenType.BEGIN)
+            begin = self._match_text_seq("BEGIN")
+            #            begin = self._match(TokenType.BEGIN)
```
tobymao, line 1209: why did you change this

sqlglot/dialects/tsql.py
GeorgeSittas: Why do we need `_prev.text` in the first place here? If I understand this correctly, we can always just output `TRANSACTION` instead of the abbreviated form, so we can also use a flag to detect this:

```python
transaction = self._match_texts(...)
```

sqlglot/dialects/tsql.py
```diff
@@ -539,3 +615,35 @@ def returning_sql(self, expression: exp.Returning) -> str:
...
+        def _durability_sql(self, expression) -> str:
+            durability = expression.args.get("durability")
+            durability_sql = ""
+            if durability is not None:
+                if durability:
+                    durability = "ON"
+                else:
+                    durability = "OFF"
+                durability_sql = f" WITH (DELAYED_DURABILITY = {durability})"
+            return durability_sql
```
GeorgeSittas, line 635:
```suggestion
            if durability is not None:
                return f" WITH (DELAYED_DURABILITY = {'ON' if durability else 'OFF'})"
            return ""
```

GeorgeSittas: Let's just get rid of this whole method and move the suggested logic in `commit_sql` instead, since that's where `durability` is used.

sqlglot/dialects/tsql.py
```diff
@@ -539,3 +615,35 @@ def returning_sql(self, expression: exp.Returning) -> str:
...
+            this = self.sql(expression, "this")
+            this = f" {this}" if this else ""
+            transaction = expression.args.get("transaction")
+            transaction = f" {transaction}" if transaction else ""
+            return f"COMMIT{transaction}{this}{self._durability_sql(expression)}"
+
+        def rollback_sql(self, expression: exp.Rollback) -> str:
+            this = self.sql(expression, "this")
+            this = f" {this}" if this else ""
+            transaction = expression.args.get("transaction")
```
GeorgeSittas, line 647:
```suggestion
            transaction = self.sql(expression, "transaction")
```

sqlglot/dialects/tsql.py
```diff
@@ -539,3 +615,35 @@ def returning_sql(self, expression: exp.Returning) -> str:
             into = self.sql(expression, "into")
             into = self.seg(f"INTO {into}") if into else ""
             return f"{self.seg('OUTPUT')} {self.expressions(expression, flat=True)}{into}"
+
+        def transaction_sql(self, expression: exp.Transaction) -> str:
+            this = self.sql(expression, "this")
+            this = f" {this}" if this else ""
+            mark = expression.args.get("mark")
+            mark = f" WITH MARK {mark}" if mark else ""
```
GeorgeSittas, line 623: This is needed otherwise you implicitly use the default dialect for `mark` when interpolating it.

```suggestion
            mark = self.sql(expression, "mark")
            mark = f" WITH MARK {mark}" if mark else ""
```

sqlglot/expressions.py
```diff
@@ -3472,15 +3472,21 @@ class Command(Expression):
...
 
 
 class Commit(Expression):
-    arg_types = {"chain": False}
+    arg_types = {
+        "chain": False,
+        "this": False,
+        "transaction": False,
+        "durability": False,
+        "modes": False,
```
GeorgeSittas, line 3484: Doesn't seem like `modes` is used anywhere. Let's delete it if it isn't needed.

tests/dialects/test_tsql.py
```diff
@@ -389,7 +389,62 @@ def test_types_bin(self):
             },
         )
 
+    def test_ddl(self):
+        sql = """CREATE TABLE #mytemp (a INTEGER,  b CHAR(2), c TIME(4), d FLOAT(24));"""
+        self.validate_all(
+            " ".join(sql.split()),
```
GeorgeSittas, line 395: This doesn't seem correct. You can do:

```python
self.validate_identity(
    "CREATE TABLE #mytemp (a INTEGER, b CHAR(2), c TIME(4), d FLOAT(24))"
)
```

tests/dialects/test_tsql.py
```diff
@@ -389,7 +389,62 @@ def test_types_bin(self):
...
+        # self.validate_identity("BEGIN TRAN")
+        self.validate_identity("BEGIN TRANSACTION transaction_name")
+        self.validate_identity("BEGIN TRANSACTION @tran_name_variable")
+        self.validate_identity("BEGIN TRANSACTION transaction_name WITH MARK 'description'")
+
+    def test_commit(self):
+        # COMMIT [ { TRAN | TRANSACTION }  [ transaction_name | @tran_name_variable ] ] [ WITH ( DELAYED_DURABILITY = { OFF | ON } ) ] [ ; ]
+
+        self.validate_identity("COMMIT")
+        self.validate_all("COMMIT TRAN")
```
GeorgeSittas, line 417: This doesn't seem correct as well. Assuming we make the change I suggested in the parser, this should become:

```python
self.validate_all("COMMIT TRAN", write={"tsql": "COMMIT TRANSACTION"})
```

tests/dialects/test_tsql.py
```diff
@@ -389,7 +389,62 @@ def test_types_bin(self):
...
+        )
+
+    def test_rollback(self):
+        # Applies to SQL Server and Azure SQL Database
+        # ROLLBACK { TRAN | TRANSACTION }
+        #     [ transaction_name | @tran_name_variable
+        #     | savepoint_name | @savepoint_variable ]
+        # [ ; ]
+        self.validate_identity("ROLLBACK")
+        self.validate_all("ROLLBACK TRAN")
```
GeorgeSittas, line 436: Ditto

tests/dialects/test_tsql.py
```diff
@@ -389,7 +389,62 @@ def test_types_bin(self):
...
+        self.validate_identity("ROLLBACK TRANSACTION transaction_name")
+        self.validate_identity("ROLLBACK TRANSACTION @tran_name_variable")
+
     def test_udf(self):
+        self.validate_identity("BEGIN")
+        self.validate_identity("END")
+        self.validate_identity("SET XACT_ABORT ON")
+        self.validate_identity(
+            "DECLARE @DWH_DateCreated DATETIME = CONVERT(DATETIME, getdate(), 104)"
+        )
```
GeorgeSittas, line 447: These don't look like they belong in `test_udf`. Let's move them in `test_tsql`.

tests/dialects/test_tsql.py
```diff
@@ -446,6 +501,7 @@ def test_udf(self):
             pretty=True,
         )
 
+    def test_fullproc(self):
```
GeorgeSittas, line 504: Why was this moved into a different testing method?

sqlglot/expressions.py
dmoore247: modes is used by Presto

sqlglot/dialects/tsql.py
dmoore247: "BEGIN" and "BEGIN TRANSACTION" mappings are required to disambiguate the flow control BEGIN..END and the transactional uses of BEGIN.

dmoore247: COMMIT and ROLLBACK can be removed.

sqlglot/dialects/tsql.py
dmoore247: I put this in to try to work on CREATE PROCEDURE refinement. I'll remove for now.

sqlglot/dialects/tsql.py
```diff
@@ -358,13 +370,75 @@ class Parser(parser.Parser):
...
+                ROLLBACK { TRAN | TRANSACTION }
+                    [ transaction_name | @tran_name_variable
+                    | savepoint_name | @savepoint_variable ]
+                    [ ; ]
+
+            Returns:
+                exp.Commit | exp.Rollback: _description_
+            """
+            rollback = False
+            if self._prev.token_type == TokenType.ROLLBACK:
```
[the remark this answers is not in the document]
dmoore247: Looks good, accept

sqlglot/dialects/tsql.py
dmoore247: I could use a little help on this one, it falls apart on me as soon as I add self.match_l_paren().

sqlglot/expressions.py
dmoore247: I had a CONVERT test case out of place and fell into changing this. I'll revert.

sqlglot/dialects/tsql.py
GeorgeSittas: You could also write this as:

```python
if self._match_pair(TokenType.WITH, TokenType.L_PAREN):
    self._match_text_seq("DELAYED_DURABILITY")
    self._match(TokenType.EQ)
    ...
```

sqlglot/dialects/tsql.py
dmoore247: Done.

sqlglot/dialects/tsql.py
dmoore247: Done.

sqlglot/dialects/tsql.py
dmoore247: That works.

sqlglot/dialects/tsql.py
dmoore247: Ok, replaced the two instances of `expression.args.get("transaction")`

sqlglot/dialects/tsql.py
dmoore247: Done

sqlglot/expressions.py
dmoore247: presto.py: 348
```python
def transaction_sql(self, expression: exp.Transaction) -> str:
            modes = expression.args.get("modes")
            modes = f" {', '.join(modes)}" if modes else ""
            return f"START TRANSACTION{modes}"
```

tests/dialects/test_tsql.py
dmoore247: I had another test for spark/databricks, but I removed that from this PR. I'll clean this for now.

dmoore247: Also TIME(4) gets mapped to TIMESTAMP(4)

dmoore247: I removed the weird join.split() which was used for the other match.

GeorgeSittas:
@dmoore247 let us know when you push the changes so we can re-review. Thanks for the quick iterations here.

tests/dialects/test_tsql.py
dmoore247: Done.

tests/dialects/test_tsql.py
dmoore247: Done.

sqlglot/expressions.py
dmoore247: You're right about the Commit(), fixing.

tests/dialects/test_tsql.py
GeorgeSittas: I think I'd just move `BEGIN`, `END`, `SET XACT_ABORT ON` to `test_tsql`. I'm not sure if the `DECLARE` one is udf-related, but if it isn't I'd move that there too. We can keep `test_fullproc` as is, maybe rename to a non-abbreviated form like `test_procedure_definition`?

sqlglot/dialects/tsql.py
```diff
@@ -283,6 +284,8 @@ class Tokenizer(tokens.Tokenizer):
 
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
+            "BEGIN": TokenType.COMMAND,
+            "SET": TokenType.SET,
```
GeorgeSittas, line 288: This and `"WITH": TokenType.WITH` are redundant, these mappings are already in the base tokenizer.

sqlglot/dialects/tsql.py
```diff
@@ -365,6 +371,63 @@ class Parser(parser.Parser):
...
+                    [ WITH ( DELAYED_DURABILITY = { OFF | ON } ) ]
+                    [ ; ]
+
+                ROLLBACK { TRAN | TRANSACTION }
+                    [ transaction_name | @tran_name_variable
+                    | savepoint_name | @savepoint_variable ]
+                    [ ; ]
+
+            Returns:
+                exp.Commit | exp.Rollback: _description_
```
GeorgeSittas, line 387: Let's get rid of this part, doesn't add much information

sqlglot/dialects/tsql.py
```diff
@@ -365,6 +371,63 @@ class Parser(parser.Parser):
...
+        def _parse_transaction(self) -> exp.Transaction:
+            """Syntax:
+                BEGIN { TRAN | TRANSACTION }
+                [ { transaction_name | @tran_name_variable }
+                [ WITH MARK [ 'description' ] ]
+                ]
+                [ ; ]
+
+            Returns:
+                exp.Transaction: _description_
```
GeorgeSittas, line 420: Ditto

sqlglot/dialects/tsql.py
```diff
@@ -365,6 +371,63 @@ class Parser(parser.Parser):
...
+            """
+            mark = None
+
+            txn_name = self._parse_id_var()
+
+            if self._match_text_seq("WITH", "MARK"):
+                mark = self._parse_string()
+
+            return self.expression(exp.Transaction, this=txn_name, mark=mark)
```
GeorgeSittas, line 429: Nit: can also write this a bit more compactly

```python
transaction = self.expression(exp.Transaction, this=self._parse_id_var())
if self._match_text_seq("WITH", "MARK"):
    transaction.set("mark", self._parse_string())

return self.expression(exp.Transaction
```

GeorgeSittas (commented):
Ok I think there are a few comments left and this should be good to go.

sqlglot/parser.py
dmoore247: TokenType.BEGIN refers to Begin transaction where in `CREATE PROCEDURE` here in the create method we want BEGIN / END as in control flow ?

tests/dialects/test_tsql.py
dmoore247: Moved.

sqlglot/dialects/tsql.py
dmoore247: Would your return be:
```python
return transaction
```

GeorgeSittas: That's correct, made a typo. Edited.

sqlglot/dialects/tsql.py
dmoore247: Ok

sqlglot/dialects/tsql.py
dmoore247: Ok

sqlglot/dialects/tsql.py
dmoore247: Removed these tokens.

sqlglot/dialects/tsql.py
dmoore247: Removed.

sqlglot/dialects/tsql.py
dmoore247: Removed

sqlglot/dialects/tsql.py
```diff
@@ -539,3 +597,31 @@ def returning_sql(self, expression: exp.Returning) -> str:
...
+            this = self.sql(expression, "this")
+            if this in ["TRAN", "TRANSACTION"]:
+                mark = self.sql(expression, "mark")
+                mark = f" WITH MARK {mark}" if mark else ""
+                return f"BEGIN TRANSACTION{mark}"
+            else:
+                this = f" {this}" if this else ""
+                mark = self.sql(expression, "mark")
+                mark = f" WITH MARK {mark}" if mark else ""
+                return f"BEGIN TRANSACTION{this}{mark}"
```
GeorgeSittas, line 609:
```suggestion
        def transaction_sql(self, expression: exp.Transaction) -> str:
            this = self.sql(expression, "this")
            this = f" {this}" if this else ""
            mark = self.sql(expression, "mark")
            mark = f" WITH MARK {mark}" if mark else ""
            return f"BEGIN TRANSACTION{this}{mark}
```

sqlglot/parser.py
```diff
@@ -1205,7 +1205,7 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
             extend_props(self._parse_properties())
 
             self._match(TokenType.ALIAS)
-            begin = self._match(TokenType.BEGIN)
+            begin = self._match_text_seq("BEGIN")
```
GeorgeSittas, line 1208: This needs to be reverted

sqlglot/dialects/tsql.py
dmoore247: Done, pushed.

sqlglot/dialects/tsql.py
```diff
@@ -365,6 +368,57 @@ class Parser(parser.Parser):
...
+            [ ; ]
+            """
+            if self._match_texts(("TRAN", "TRANSACTION")):
+                # we have a transaction and not a BEGIN
+                transaction = self.expression(exp.Transaction, this=self._parse_id_var())
+                if self._match_text_seq("WITH", "MARK"):
+                    transaction.set("mark", self._parse_string())
+                return transaction
+
+            return self.expression(exp.Command, this="BEGIN")
```
GeorgeSittas, line 420: This should instead be:

```suggestion
            return self._parse_as_command(self._prev)
```

The `exp.Command` class is generally used as a fallback and just stores a raw string from the "command" token to the end of the chunk being parsed. This allows the parser to be lenient instead of failing, even in cases where we don't actually parse the syntax of a dialect-specific command.

Implementing this suggestion covers what you've already tested + more, so it's more robust.

sqlglot/parser.py
dmoore247: reverted.

sqlglot/dialects/tsql.py
dmoore247: I found in some cases, `this` was `TRAN` or `TRANSACTION` and ended up with `BEGIN TRANSACTION TRAN`

GeorgeSittas: What's an example where this happens?

sqlglot/dialects/tsql.py
dmoore247: Done.

dmoore247:
@GeorgeSittas Looks like I resolved all the issues.

PR #1916, opened 2023-07-13
Fix!(optimizer): improve handling of wrapped subqueries

GeorgeSittas:


sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -144,8 +144,9 @@ def _eliminate_derived_table(scope, existing_ctes, taken):
     name, cte = _new_cte(scope, existing_ctes, taken)
 
     table = exp.alias_(exp.table_(name), alias=parent.alias or name)
-    parent.replace(table)
+    table.set("joins", parent.args.get("joins"))
 
+    parent.replace(table)
     return cte
```
GeorgeSittas, line 150: This was changed in order to correctly handle the following query:

```sql
SELECT * FROM ((SELECT c FROM t1) JOIN t2)
```

Notice that we're factoring out `SELECT c FROM t1` into a CTE and then calling `replace(table)` on it. This has the unintended side effect of getting rid of the `joins` that were initially attached on the subquery.

The problem is shown below, where the joins haven't been reattached to the table:

```python
>>> optimize("SELECT * FROM ((SELECT c FROM t1) JOIN t2)").sql()
'WITH "_q_0" AS (SELECT "t1"."c" AS "c" FROM "t1" AS "t1") SELECT * FROM ("_q_0" AS "_q_0")'
```

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -176,6 +176,7 @@ def _outer_select_joins_on_inner_select_join():
 
     return (
         isinstance(outer_scope.expression, exp.Select)
+        and not outer_scope.expression.is_star
```
GeorgeSittas, line 179: This was motivated by the same example as in `eliminate_subqueries`, where the schema is unknown:

```python
>>> optimize("SELECT * FROM ((SELECT c FROM t1) JOIN t2)").sql()
'SELECT * FROM ("t1" AS "t1" CROSS JOIN "t2" AS "t2")'
```

Notice how the `c` projection was dropped in the optimized form.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -242,6 +243,7 @@ def _merge_from(outer_scope, inner_scope, node_to_replace, alias):
         alias (str)
     """
     new_subquery = inner_scope.expression.args["from"].this
+    new_subquery.set("joins", node_to_replace.args.get("joins"))
```
GeorgeSittas, line 246: This was added because we can have `<CTEs> SELECT ... FROM (tbl1 JOIN tbl2)`. If we just merged a CTE into `tbl1`, then we'd lose the `JOIN tbl2` part because it's attached on the Table.

A concrete example (without this addition):

```python
>>> schema = {"x": {"a": "int", "b": "int"}, "y": {"b": "int", "c": "int"}}
>>> optimize("SELECT b FROM ((SELECT a FROM x) INNER JOIN y ON a = b)", schema=schema).sql()
'SELECT "y"."b" AS "b" FROM ("x" AS "x")'
```

The expression just before merging CTEs is:

```sql
WITH _q_0 AS (SELECT x.a AS a FROM x AS x), y_2 AS (SELECT y.b AS b FROM y AS y) SELECT y.b AS b FROM (_q_0 AS _q_0 JOIN y_2 AS y ON _q_0.a = y.b)
```

We basically replace `_q_0` with the first CTEs table name and we lose the join, as explained above.

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -61,6 +61,9 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
             if remove_unused_selections:
                 _remove_unused_selections(scope, parent_selections, schema)
 
+            if scope.expression.is_star:
+                continue
+
```
GeorgeSittas, line 66: We don't want to hit the following section where we update `referenced_columns` if there's a star projection in the outer scope, because then we might treat the inner scope's projections as "unused", even though this isn't safe.

For example:

```python
>>> from sqlglot.optimizer.pushdown_projections import pushdown_projections
>>> import sqlglot
>>> pushdown_projections(sqlglot.parse_one("SELECT * FROM ((SELECT c FROM t1) JOIN t2)")).sql()
'SELECT * FROM ((SELECT 1 AS _ FROM t1), t2)'
```

GeorgeSittas: I also tried to include a check like `scope.parent.expression.is_star` in the above condition:

https://github.com/tobymao/sqlglot/blob/726306e9a77a700555b31fbe136e6438f711178f/sqlglot/optimizer/pushdown_projections.py#L42-L45

But this broke e.g. the following test:

```sql
SELECT a FROM (SELECT * FROM aa UNION ALL SELECT * FROM bb UNION ALL SELECT * from cc);
SELECT _q_0.a AS a FROM (SELECT aa.a AS a FROM aa AS aa UNION ALL SELECT bb.a AS a FROM bb AS bb UNION ALL SELECT cc.a AS a FROM cc AS cc) AS _q_0;
```

Because we weren't pushing down the projection `a` into the inner scope, due to the fact that `UNION ALL` scope is the parent of `SELECT * FROM aa` and it's a star.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -410,7 +411,9 @@ def _expand_stars(
             else:
                 return
 
-    scope.expression.set("expressions", new_selections)
+    # Ensures we don't overwrite the initial selections with an empty list
+    if new_selections:
+        scope.expression.set("expressions", new_selections)
```
GeorgeSittas, line 416: This isn't currently necessary for the tests to pass, but it was when I wrote it because I ended up with a query like `SELECT FROM ...` due to `new_selections` being empty.

We can remove it or keep it, but imo it's worth having this as a sanity check since it's cheap.

tests/fixtures/optimizer/pushdown_projections.sql
```diff
@@ -16,9 +16,6 @@ SELECT x.a AS a FROM x AS x JOIN (SELECT y.b AS b FROM y AS y) AS z ON x.b = z.b
 SELECT x1.a FROM (SELECT * FROM x) AS x1, (SELECT * FROM x) AS x2;
 SELECT x1.a AS a FROM (SELECT x.a AS a FROM x AS x) AS x1, (SELECT 1 AS _ FROM x AS x) AS x2;
 
-SELECT x1.a FROM (SELECT * FROM x) AS x1, (SELECT * FROM x) AS x2;
-SELECT x1.a AS a FROM (SELECT x.a AS a FROM x AS x) AS x1, (SELECT 1 AS _ FROM x AS x) AS x2;
-
```
GeorgeSittas, line 21: Duplicate

tests/fixtures/optimizer/pushdown_projections.sql
```diff
@@ -77,9 +74,6 @@ SELECT _q_0.a AS a FROM (SELECT aa.a AS a FROM aa AS aa UNION ALL SELECT bb.a AS
 SELECT a FROM (SELECT a FROM aa UNION ALL SELECT * FROM bb UNION ALL SELECT * from cc);
 SELECT _q_0.a AS a FROM (SELECT aa.a AS a FROM aa AS aa UNION ALL SELECT bb.a AS a FROM bb AS bb UNION ALL SELECT cc.a AS a FROM cc AS cc) AS _q_0;
 
-SELECT a FROM (SELECT * FROM aa UNION ALL SELECT * FROM bb UNION ALL SELECT * from cc);
-SELECT _q_0.a AS a FROM (SELECT aa.a AS a FROM aa AS aa UNION ALL SELECT bb.a AS a FROM bb AS bb UNION ALL SELECT cc.a AS a FROM cc AS cc) AS _q_0;
-
```
GeorgeSittas, line 82: Duplicate

sqlglot/optimizer/qualify_columns.py
tobymao: i've seen this before too

tobymao: it'd be good to add a test for it if possible

tests/fixtures/optimizer/qualify_tables.sql
```diff
@@ -101,3 +101,11 @@ SELECT * FROM (SELECT * FROM c.db.tbl1 AS tbl1 JOIN c.db.tbl2 AS tbl2 ON id1 = i
...
 SELECT * FROM (tbl1 AS tbl1 JOIN (tbl2 AS tbl2 JOIN tbl3 AS tbl3 ON id2 = id3) AS _q_0 ON id1 = id3) AS _q_1;
 SELECT * FROM (SELECT * FROM c.db.tbl1 AS tbl1 JOIN (SELECT * FROM c.db.tbl2 AS tbl2 JOIN c.db.tbl3 AS tbl3 ON id2 = id3) AS _q_0 ON id1 = id3) AS _q_1;
+
+# title: wrapped subquery without alias
+SELECT * FROM ((SELECT * FROM t));
+SELECT * FROM ((SELECT * FROM c.db.t AS t) AS _q_0);
+
+# title: wrapped subquery without alias joined with a table
+SELECT * FROM ((SELECT * FROM t1) INNER JOIN t2 ON a = b);
+SELECT * FROM ((SELECT * FROM c.db.t1 AS t1) AS _q_0 INNER JOIN c.db.t2 AS t2 ON a = b);
```
tobymao, line 111: can you add some tests to qualify_columns? it's easier to reason about those tests than optimizer

tobymao (approved):

sqlglot/optimizer/scope.py
```diff
@@ -616,7 +616,7 @@ def _is_derived_table(expression: exp.Subquery) -> bool:
     as it doesn't introduce a new scope. If an alias is present, it shadows all names
     under the Subquery, so that's one exception to this rule.
     """
-    return bool(expression.alias or not isinstance(expression.unnest(), exp.Table))
+    return bool(expression.alias or isinstance(expression.this, exp.Subqueryable))
```
tobymao, line 619: is this and the other isinstance check on 127 and 742 the bulk of the change? it seems like everything else is just handling the side effect of that?

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: yep I'll try to come up with a test case that covers it

tests/fixtures/optimizer/qualify_tables.sql
GeorgeSittas: yes

sqlglot/optimizer/scope.py
GeorgeSittas: more or less -- all changes in the scope module were key and the rest were just side effects, yes

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: added a test, see https://github.com/tobymao/sqlglot/pull/1916/commits/22803c03dc503f4c7ea7bdd9322f59ce9eea25e7

tests/fixtures/optimizer/qualify_tables.sql
GeorgeSittas: added several tests for qualify columns too, mostly taken from optimizer.sql

PR #1918, opened 2023-07-13
feat(spark) add support for tsql's SELECT ... INTO #temptable to spark

dmoore247:
TSQL has `SELECT ... INSERT INTO #temptable FROM sourcetable` which translates 
well into `CREATE OR REPLACE VIEW AS SELECT ... FROM sourcetable`


dmoore247:
We might want to test this with more complex FROM clauses with joins and multiple tables...

We might also want to map #temptable name in tsql to `#temptable` in spark to help users recognize it as a view and to minimize the name space conflict with existing tables. (another PR perhaps?)

sqlglot/dialects/spark2.py
```diff
@@ -199,6 +221,14 @@ class Generator(Hive.Generator):
             exp.Reduce: rename_func("AGGREGATE"),
             exp.StrToDate: _str_to_date,
             exp.StrToTime: lambda self, e: f"TO_TIMESTAMP({self.sql(e, 'this')}, {self.format_time(e)})",
+            exp.Select: transforms.preprocess(
+                [
```
tobymao, line 225: make the base hive transform an array so that you can call them here so you're not copy and pasting

tobymao:
@dmoore247 thanks for the PR, i really appreciate you taking the time to do it. but i'm not super inclined to accept it. i think this particular use case is a bit specific. i'm leaning towards rejecting this PR.

you can implement this as a custom transform in your custom code. i'm going to sleep on this and let @GeorgeSittas chime in.

GeorgeSittas:
I agree with Toby on this particular transformation being a bit too specific. On the other hand, there may be an action item here because e.g. t-sql, postgres and redshift all support the SELECT-INTO syntax, and so converting it into a CTAS might help with transpiling it to other dialects.

See the following section from postgres' docs:

> [CREATE TABLE AS](https://www.postgresql.org/docs/9.1/sql-createtableas.html) is functionally similar to SELECT INTO. CREATE TABLE AS is the recommended syntax, since this form of SELECT INTO is not available in ECPG or PL/pgSQL, because they interpret the INTO clause differently. Furthermore, CREATE TABLE AS offers a superset of the functionality provided by SELECT INTO.

> Prior to PostgreSQL 8.1, the table created by SELECT INTO included OIDs by default. In PostgreSQL 8.1, this is not the case — to include OIDs in the new table, the [default_with_oids](https://www.postgresql.org/docs/9.1/runtime-config-compatible.html#GUC-DEFAULT-WITH-OIDS) configuration variable must be enabled. Alternatively, CREATE TABLE AS can be used with the WITH OIDS clause.

Based on this, it may make sense to implement this transformation in the future. In any case, I agree that dealing with the `#` prefix should probably be moved into custom application code.

- https://www.postgresql.org/docs/9.1/sql-selectinto.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_SELECT_INTO.html#r_SELECT_INTO-examples
- https://learn.microsoft.com/en-us/sql/t-sql/queries/select-into-clause-transact-sql?view=sql-server-ver16

tobymao:
for now i think you should just do this in your application

PR #1929, opened 2023-07-18
Fix(spark,duckdb): transpile TO_TIMESTAMP, MONTHS_BETWEEN correctly

### LINKED ISSUE #1928: Spark functions months_bewteen & to_timestamp not transpiling to DuckDB
I'm attempting to transpile from Spark to DuckDB. Sqlglot doesn't seem to be recognising the spark functions "months_between" and "to_timestamp". 

See example below (apologies for the image, I can't paste the code directly), neither of those functions gets transpiled:

image

"months_bewteen" should become "date_diff", and
"to_timestamp" should become "strptime" in duckdb, see below:

image

**Official Documentation**
[spark: months_bewteen](https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.functions.months_between.html)
[spark: to_timestamp](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_timestamp.html)
[duckdb: time functions](https://duckdb.org/docs/sql/functions/timestamp.html)

GeorgeSittas:
Fixes #1928

References:
- https://spark.apache.org/docs/latest/api/sql/index.html#datediff
- https://duckdb.org/docs/sql/functions/timestamp.html
- https://cwiki.apache.org/confluence/display/hive/languagemanual+udf

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -40,7 +40,7 @@ def _date_add_sql(
 
 
 def _derived_table_values_to_unnest(self: generator.Generator, expression: exp.Values) -> str:
-    if not expression.find_ancestor(exp.From, exp.Select):
+    if not expression.find_ancestor(exp.From, exp.Join):
         return self.values_sql(expression)
 
     alias = expression.args.get("alias")
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -40,7 +40,7 @@ def _date_add_sql(
 
 
 def _derived_table_values_to_unnest(self: generator.Generator, expression: exp.Values) -> str:
-    if not expression.find_ancestor(exp.From):
+    if not expression.find_ancestor(exp.From, exp.Select):
         return self.values_sql(expression)
 
     alias = expression.args.get("alias")
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -40,7 +40,7 @@ def _date_add_sql(
 
 
 def _derived_table_values_to_unnest(self: generator.Generator, expression: exp.Values) -> str:
-    if not isinstance(expression.unnest().parent, exp.From):
+    if not isinstance(expression.parent, exp.From):
         return self.values_sql(expression)
 
     alias = expression.args.get("alias")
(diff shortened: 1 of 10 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -40,7 +40,7 @@ def func(self, expression):
 
 
 def _derived_table_values_to_unnest(self: generator.Generator, expression: exp.Values) -> str:
-    if not isinstance(expression.unnest().parent, exp.From):
+    if not isinstance(expression.parent, exp.From):
```
GeorgeSittas, line 43: This is a no-op because we know `expression` is an instance of `Values` here, so went ahead and fixed it while I was at it.

tobymao: just because expression is a values, doesn't mean it's parent is a from? i'm not sure i understand

sqlglot/dialects/spark2.py
```diff
@@ -132,6 +133,9 @@ class Parser(Hive.Parser):
             ),
             "STRING": _parse_as_cast("string"),
             "TIMESTAMP": _parse_as_cast("timestamp"),
+            "TO_TIMESTAMP": lambda args: _parse_as_cast("timestamp")(args)
+            if len(args) == 1
+            else format_time_lambda(exp.StrToTime, "spark")(args),
```
tobymao, line 138: is this always true? can't you call to_timestamp on a date object?

sqlglot/dialects/bigquery.py
GeorgeSittas: I mean that `expression.unnest()` seems like a no-op here: `expression` is of type `exp.Values` which doesn't override `unnest`.

sqlglot/dialects/spark2.py
GeorgeSittas: Based on their docs, you can't:

Screenshot 2023-07-18 at 9 23 41 PM

Based on spark-sql's behaviour, you can:

```python
spark-sql (default)> select to_timestamp(cast('2020-01-01' as date));
2020-01-01 00:00:00
Time taken: 4.96 seconds, Fetched 1 row(s)
```

I just found out myself too, assumed the docs would be accurate :)

What do you think we should do about this?

sqlglot/dialects/bigquery.py
GeorgeSittas: Makes sense, will update

PR #1937, opened 2023-07-20
Feat(clickhouse): add support for the logical xor function

### LINKED ISSUE #1936: `xor` function fails to parse with the clickhouse dialect

```
In [12]: import sqlglot as sg

In [13]: sg.__version__
Out[13]: '17.6.0'

In [14]: sg.parse_one("select xor(true, false)", read="clickhouse")
...
ParseError: Invalid expression / Unexpected token. Line 1, Col: 10.
  select xor(true, false)
```

**Official Documentation**

https://clickhouse.com/docs/en/sql-reference/functions/logical-functions#xor

GeorgeSittas:
Fixes #1936

References:
- https://clickhouse.com/docs/en/sql-reference/functions/logical-functions#xor
- https://dev.mysql.com/doc/refman/5.7/en/logical-operators.html#operator_xor

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -58,12 +58,18 @@ class ClickHouse(Dialect):
         }
 
     class Parser(parser.Parser):
+        FUNC_TOKENS = {
+            *parser.Parser.FUNC_TOKENS,
+            TokenType.XOR,
+        }
+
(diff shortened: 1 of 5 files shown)


GeorgeSittas:
Seems like duckdb also supports `xor(a, b)`, will add support for that shortly

tobymao (approved):

PR #1941, opened 2023-07-21
Fixes #1940 for at least create/insert on limited query set

### LINKED ISSUE #1940: WITH expressions are incorrectly attached to INSERT/CREATE?
**Before you file an issue**
Checked against latest main

This SQL with either INSERT or CREATE results in a tree hierarchy that does not qualify end to end, because the WITH statement is attached to the create or insert instead of the embedded SELECT.

```WITH cte AS (
        SELECT 
                d.a, 
                b + 1 AS c, 
                e.f 
        FROM 
                db1.d 
        INNER JOIN db2.e ON db1.d.a = e.a 
        WHERE db1.d.a = 1),
second_cte AS (
SELECT * FROM cte
)

INSERT x
SELECT 
        *
FROM
        second_cte```

The following code 
```
        qualified = qualify(
                expression=node,
                schema=schema,
                db="db1",
                catalog="s",
                dialect="bigquery",
        )
        print(repr(qualified))
```

Shows a hierarchy that doesn't fully qualify the SELECT statements end to end.
Code that fixes this is supplied in a draft PR, but because of my lack of knowledge of sqlglot, it is very likely not the correct way of solving this. So mostly looking for feedback/confirmation, in the hope that the correct solution for it is chosen. There are likely other statements that suffer from the same issue.

gtoonstra:
See https://github.com/tobymao/sqlglot/issues/1940.

This is very likely the incorrect way of solving it, but it serves to request feedback / demonstration that it will make the statements end-to-end qualifyable.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -992,7 +992,6 @@ class Uncache(Expression):
 
 class Create(Expression):
     arg_types = {
-        "with": False,
         "this": True,
         "kind": True,
         "expression": False,
@@ -1527,7 +1526,6 @@ class Index(Expression):
 
 class Insert(Expression):
     arg_types = {
-        "with": False,
         "this": True,
         "expression": False,
         "conflict": False,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1933,6 +1933,12 @@ class Parser(metaclass=_Parser):
 
             if "with" in this.arg_types:
                 this.set("with", cte)
+            elif this.key == "create":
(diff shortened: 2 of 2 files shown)


gtoonstra:
The failing CI already shows it wouldn't work for all kinds of expressions, but hopefully this PR will lead to a better solution.

GeorgeSittas:
This doesn't really solve the problem, see discussion in #1940. We'll probably need more / different changes to fully address it.

I'll go ahead and close this PR for now, but we can continue discussing about this in the corresponding issue.

PR #1949, opened 2023-07-22
Feat(optimizer): add support for resolving CTEs in CREATE statements

gtoonstra:
Pull request to resolve CTEs when CREATE statement is at the root of the AST. 

I've dived a lot more into the scopes and resolvers and came up with this implementation so far.

Looking forward to feedback.


sqlglot/expressions.py
```diff
@@ -1006,6 +1006,31 @@ class Create(Expression):
...
+
+    @property
+    def is_star(self) -> bool:
+        if isinstance(self.expression, Select):
+            return self.expression.is_star
+        return False
+
+    @property
+    def selects(self) -> t.List[Expression]:
+        if isinstance(self.expression, Select):
```
tobymao, line 1030: this seems too specific, should it instead be like checking subqueryable, derived_table or has the attribute selects?

tobymao: all the isinstance checks i think should be more general

sqlglot/expressions.py
```diff
@@ -1006,6 +1006,31 @@ class Create(Expression):
         "clone": False,
     }
 
+    @property
+    def ctes(self):
```
tobymao, line 1010: should we move these helper methods to a shared base class and have this applied to Insert and Update or other DDLs? 

then in the optimizer you can check for the base class instead of just Create

tests/test_optimizer.py
```diff
@@ -806,3 +806,13 @@ def test_no_pseudocolumn_expansion(self):
             ),
             parse_one('SELECT "a"."a" AS "a", "a"."b" AS "b" FROM "a" AS "a"'),
         )
+
+    def test_qualify_columns_create_with_cte(self):
```
tobymao, line 810: don't add this test here, just put it in qualify_columns.sql tests

tobymao:
thanks so much for the PR!

sqlglot/optimizer/scope.py
```diff
@@ -509,7 +509,9 @@ def traverse_scope(expression: exp.Expression) -> t.List[Scope]:
     Returns:
         list[Scope]: scope instances
     """
-    if not isinstance(expression, exp.Unionable):
+    if isinstance(expression, exp.Create):
+        return list(_traverse_scope(Scope(expression)))
+    elif not isinstance(expression, exp.Unionable):
```
GeorgeSittas, line 514: Maybe we should make this a bit more generic by creating a DDLExpression class or something (Toby mentioned something like this in another comment) that `Create` and `Insert` inherit from, so that we can deal with both of them at the same time.

tobymao:
@gtoonstra are you planning to finish up this pr?

tests/fixtures/optimizer/qualify_columns_create.sql
```diff
@@ -0,0 +1,29 @@
+# title: Qualify columns for a create with CTE
+# execute: false
+WITH cte AS (SELECT a FROM db.y) CREATE TABLE s AS SELECT * FROM cte;
+WITH cte AS (SELECT a AS a FROM db.y AS y) CREATE TABLE s AS SELECT cte.a AS a FROM cte;
```
GeorgeSittas, line 4: Why doesn't `cte` have an alias in the final `SELECT` here? Seems a bit off. Ditto for subsequent tests.

GeorgeSittas (commented):
LGTM, one question & agreed with Toby about renaming `TableChange` to `DDL`. I believe `INSERT` and `UPDATE` should just work with a few changes (i.e. do instance checks on the common superclass instead of just `CREATE`).

Btw, if we're to create a new test file for qualify columns, I'd rather use a bit more generic name like `qualify_columns_ddl.sql` or something to anticipate for the `INSERT`, `UPDATE` tests as well.

Nice 👍

sqlglot/optimizer/scope.py
```diff
@@ -741,6 +745,11 @@ def _traverse_udtfs(scope):
 
     scope.sources.update(sources)
```
GeorgeSittas, line 747:
```suggestion


```

GeorgeSittas:
I'll complete this, but most of the work should be done. Thanks for the PR @gtoonstra!

PR #1957, opened 2023-07-25
feat(tsql) Better temporary table handling

dmoore247:
TSQL (Synapse, SQL Server) uses `#`, `##` and `@` to denote local temporary table, global temporary table and table variable. TSQL does not have the `TEMPORARY` keyword. Temporary tables occur very frequently in TSQL code instead of CTEs as a way to modularize stored procedure code and code within powerbi reports.


GeorgeSittas:
Closing in favor of #1958

PR #1958, opened 2023-07-25
Feat(tsql): improve transpilation of temp table DDLs

GeorgeSittas:
@dmoore247 this covers most of the tests you wrote, but I haven't done something about table vars yet. Could continue in a future PR or feel free to take a stab yourself too.

Also considered creating a couple of new expressions instead of adding new args in `Identifier`, but the resulting code with this approach looked a bit cleaner I think. Don't feel strongly about it, so happy to pivot if needed.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -307,10 +307,6 @@ class TSQL(Dialect):
             "SYSTEM_USER": TokenType.CURRENT_USER,
         }
 
-        # TSQL allows @, # to appear as a variable/identifier prefix
-        SINGLE_TOKENS = tokens.Tokenizer.SINGLE_TOKENS.copy()
-        SINGLE_TOKENS.pop("#")
-
     class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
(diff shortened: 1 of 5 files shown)


dmoore247:
Local and Global temp table support is huge. TSQL and Snowflake report temp table variables but I haven't encountered them much. Doing these later is good.

sqlglot/dialects/tsql.py
```diff
@@ -518,6 +514,36 @@ def _parse_user_defined_function(
             expressions = self._parse_csv(self._parse_function_parameter)
             return self.expression(exp.UserDefinedFunction, this=this, expressions=expressions)
 
+        def _parse_id_var(
+            self,
+            any_token: bool = True,
+            tokens: t.Optional[t.Collection[TokenType]] = None,
+        ) -> t.Optional[exp.Expression]:
```
tobymao, line 521: is_temporary = self._match(TokenType.Hash)
is_global = is_temporary and  self._match(TokenType.Hash)

sqlglot/generator.py
```diff
@@ -2426,6 +2427,9 @@ def expressions(
         result_sqls = []
         for i, e in enumerate(expressions):
             sql = self.sql(e, comment=False)
+            if not sql:
```
tobymao, line 2430: why was this never hit before?

GeorgeSittas: Probably didn't have a use case where an expression was generating an empty string.

tobymao: what hits it now?

GeorgeSittas: The test case I tried was actually mis-formatted due to this line: https://github.com/tobymao/sqlglot/pull/1958/files#diff-362498362ae539e216cd83ebfe510977107648742f1a1bee7b2db7b5b7d27974L992: `TemporaryProperty` generates `""` for T-SQL and so even though `expressions` is simply `[""]` in this section, we still output a redundant space separator.

I mentioned this in a [comment](https://github.com/tobymao/sqlglot/pull/1958#discussion_r1274042732) above, and you can easily see that if we had more properties in the DDL we'd hit the same issue due to the `sep.join(...)` in the `expressions` method.

tobymao (approved):

PR #1960, opened 2023-07-26
Fix(optimizer): traverse UNNEST scope

GeorgeSittas:


sqlglot/optimizer/scope.py
```diff
@@ -711,6 +711,31 @@ def _traverse_subqueries(scope):
         scope.subquery_scopes.append(top)
 
 
+def _traverse_udtfs(scope):
+    sources = {}
+
+    if isinstance(scope.expression, exp.Unnest):
```
tobymao, line 717: why are you checking specifically for unnest?

sqlglot/optimizer/scope.py
```diff
@@ -711,6 +711,31 @@ def _traverse_subqueries(scope):
...
+        unnested_source = scope.expression.expressions[0]
```
tobymao, line 718: why only the first value?

sqlglot/optimizer/scope.py
GeorgeSittas: I'll work on handling `exp.Lateral` too. I think these two are the only meaningful `UDTF`s to deal with here, as the other one is `Values`.

sqlglot/optimizer/scope.py
GeorgeSittas: Will update to traverse all expressions

sqlglot/parser.py
```diff
@@ -2242,7 +2242,12 @@ def _parse_lateral(self) -> t.Optional[exp.Lateral]:
             return None
 
         if not this:
-            this = self._parse_function() or self._parse_id_var(any_token=False)
+            this = (
+                self._parse_unnest()
```
GeorgeSittas, line 2246: Without this we used to parse the `UNNEST` in `... JOIN LATERAL UNNEST(...)` into `exp.Anonymous`.

PR #1961, opened 2023-07-27
Feat(mysql): improve support for DDL index column constraints

### LINKED ISSUE #1959: getting error for foreign key constrain for mysql
code =  from sqlglot import parse_one,dialects
sql = """
CREATE TABLE IF NOT EXISTS `industry_info` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `industry_id` bigint(20) NOT NULL,
  `industry_column_1` varchar(1000) ,
  `industry_column_2` varchar(1000) ,
  `industry_column_3` varchar(1000) ,
  PRIMARY KEY (`id`),
  UNIQUE KEY `unique_industry_id` (`industry_id`),
  KEY `fk_industry_info_id` (`industry_id`)
) ;
     """
s = parse_one(sql, dialect = dialects.Dialects.MYSQL)
print(s)

error        -        raise error
sqlglot.errors.ParseError: Expecting ). Line 10, Col: 27.
  n_3` varchar(1000) ,
  PRIMARY KEY (`id`),
  UNIQUE KEY `unique_industry_id` (`industry_id`),
  KEY `fk_industry_info_id` (`industry_id`)
) ;
**mainly the error is because of this foreign key constraint line   KEY `fk_industry_info_id` (`industry_id`)**

GeorgeSittas:
Fixes #1959 -- still a WIP. An alternative I considered for _index_options_ was to create a separate class for each option. That way we'd avoid the if checks in the generator, but we'd have a handful more expression types to deal with, so wasn't sure and went ahead with this approach instead because it felt simple. Will refine & add more tests tomorrow morning.

Reference: https://dev.mysql.com/doc/refman/8.0/en/create-table.html

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -303,6 +303,21 @@ class MySQL(Dialect):
             "NAMES": lambda self: self._parse_set_item_names(),
         }
 
+        CONSTRAINT_PARSERS = {
+            **parser.Parser.CONSTRAINT_PARSERS,
+            "FULLTEXT": lambda self: self._parse_index_constraint(kind="FULLTEXT"),
+            "INDEX": lambda self: self._parse_index_constraint(),
+            "KEY": lambda self: self._parse_index_constraint(),
+            "SPATIAL": lambda self: self._parse_index_constraint(kind="SPATIAL"),
+        }
+
+        SCHEMA_UNNAMED_CONSTRAINTS = {
+            *parser.Parser.SCHEMA_UNNAMED_CONSTRAINTS,
+            "FULLTEXT" "INDEX",
+            "KEY",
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -327,6 +343,57 @@ class Parser(parser.Parser):
...
+            if kind:
+                self._match_texts({"INDEX", "KEY"})
+
+            this = self._parse_id_var()
+            type_ = self._match(TokenType.USING) and self._advance_any() and self._prev.text
+            schema = self._parse_schema()
+
+            options = []
+            while True:
+                opt = None
```
tobymao, line 358: i prefer to just do the eles: None

tobymao (approved):

PR #1962, opened 2023-07-27
feat(tsql) datediff(...,INTEGER,...) support

dmoore247:
Widely used convention in TSQL involves passing 0 as the start_date to datediff. As in this code to find the first day of the month corresponding to `@mydate`. Other integer start_dates work as "days since `1900-01-01`. In fact, it's most accurate to use `'1900-01-01 00:00:00.000'` instead of 0.

Example from [Stack Overflow](https://stackoverflow.com/a/1520803/7379644)
```sql
SELECT DATEADD(month, DATEDIFF(month, 0, @mydate), 0) AS StartOfMonth
```

Other sql dialects, such as duckdb, expect start_date to be of type TIMESTAMP or type DATE. Spark will accept varchar form of a date / timestamp and implicitly perform the date/timestamp conversion.

Implementation of this feature will increase real-world conversion rate of TSQL to other dialects.

This pull request starts with the unittests. The implementation will follow.


tobymao:
in the future please make the pr once it’s ready to be reviewed, no need to add unit tests first

sqlglot/dialects/tsql.py
```diff
@@ -132,6 +133,28 @@ def generate_date_delta_with_unit_sql(
...
 
 
+def _parse_date_delta(
+    exp_class: t.Type[E], unit_mapping: t.Optional[t.Dict[str, str]] = None
+) -> t.Callable[[t.List], E]:
+    def inner_func(args: t.List) -> E:
+        unit_based = len(args) == 3
+        this = args[2] if unit_based else seq_get(args, 0)
+        unit = args[0] if unit_based else exp.Literal.string("DAY")
+        unit = exp.var(unit_mapping.get(unit.name.lower(), unit.name)) if unit_mapping else unit
```
GeorgeSittas, line 143: Can we dry out this copy-pasted section?

sqlglot/dialects/tsql.py
```diff
@@ -132,6 +133,28 @@ def generate_date_delta_with_unit_sql(
...
+
+        start_date = str(seq_get(args, 1))
+        if start_date.isnumeric():
+            # perform date addition
+            days = int(start_date)
+            epoc = datetime.datetime.strptime("1900-01-01", "%Y-%M-%d")
```
GeorgeSittas, line 149: [Nit] This is a bit misleading, since "epoch" generally refers to 1970-01-01. I'm not sure what terminology T-SQL uses, but should we call it "start" or something else?

tests/dialects/test_tsql.py
```diff
@@ -818,8 +818,23 @@ def test_add_date(self):
...
     def test_date_diff(self):
-        self.validate_identity("SELECT DATEDIFF(year, '2020/01/01', '2021/01/01')")
-
+        self.validate_identity("SELECT DATEDIFF(year, '2020-01-01', '2021/01/01')")
+        self.validate_all(
+            "SELECT DATEDIFF(quarter, 0, '2021-01-01')",
+            write={
+                "tsql": "SELECT DATEDIFF(quarter, '1900-01-01', '2021-01-01')",
+                "spark": "SELECT DATEDIFF(quarter, '1900-01-01', '2021-01-01')",
+                # "duckdb": "SELECT DATEDIFF('quarter', TIMESTAMP'1900-01-01', TIMESTAMP'2021-01-01')",
```
GeorgeSittas, line 827: You can use `DateStrToDate` or perhaps the broader `TimeStrToTime` to represent T-SQL's `DATEDIFF` arguments at the AST level. That way you'll automatically cover these DuckDB test cases.

sqlglot/dialects/tsql.py
tobymao: change it to datetime.date(1900, 1, 1)

GeorgeSittas:
Closing this in favor of #1967

PR #1967, opened 2023-07-27
Feat(tsql): improve support for the DATEDIFF function

GeorgeSittas:


sqlglot/dialects/dialect.py
```diff
@@ -555,11 +555,11 @@ def right_to_substring_sql(self: Generator, expression: exp.Left) -> str:
 
 
 def timestrtotime_sql(self: Generator, expression: exp.TimeStrToTime) -> str:
-    return f"CAST({self.sql(expression, 'this')} AS TIMESTAMP)"
+    return self.sql(exp.cast(expression.this, "timestamp"))
 
 
 def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
-    return f"CAST({self.sql(expression, 'this')} AS DATE)"
+    return self.sql(exp.cast(expression.this, "date"))
```
GeorgeSittas, line 562: These changes were made so that `TIMESTAMP` and `DATE` are generated according to the target dialect, instead of being hard-coded. This was important for T-SQL which generates `DATETIME2` for `exp.DataType.Type.TIMESTAMP`.

sqlglot/dialects/tsql.py
```diff
@@ -166,6 +170,29 @@ def _string_agg_sql(self: generator.Generator, expression: exp.GroupConcat) -> s
...
+def _parse_date_delta(
+    exp_class: t.Type[E], unit_mapping: t.Optional[t.Dict[str, str]] = None
+) -> t.Callable[[t.List], E]:
+    def inner_func(args: t.List) -> E:
+        unit = seq_get(args, 0)
+        if unit and unit_mapping:
+            unit = exp.var(unit_mapping.get(unit.name.lower(), unit.name))
+
+        start_date = seq_get(args, 1)
+        if isinstance(start_date, exp.Literal) and start_date.this.isnumeric():
```
tobymao, line 182: why not start_date.is_number or start_date.is_int

also can you look at is_int and is correct? it looks off

GeorgeSittas: Good point, I like that better thanks

dmoore247: Scarily, 1.5 is a valid `DATETIME` value. Not a valid `DATETIME2` value.

GeorgeSittas: Added a test for 1.5 as well, tested using sql fiddle and transpilation looks fine

PR #1971, opened 2023-07-27
Fix(mysql): simplify LIMIT, OFFSET when their expression is complex

GeorgeSittas:
MySQL prohibits complex expressions in the `LIMIT` and `OFFSET` clauses, so we need to simplify them.

For example, the following query is invalid in MySQL:

```sql
SELECT 1 LIMIT 0 + 1
```

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -18,6 +18,7 @@ from sqlglot.dialects.dialect import (
     no_trycast_sql,
     parse_date_delta_with_interval,
     rename_func,
+    simplify_expression,
     strposition_to_locate_sql,
 )
 from sqlglot.helper import seq_get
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -553,6 +553,26 @@ class Generator(generator.Generator):
             exp.DataType.Type.VARCHAR: "CHAR",
         }
 
+        def limit_sql(self, expression: exp.Limit, top: bool = False) -> str:
+            if not isinstance(expression.expression, exp.Literal):
```
tobymao, line 557: maybe we can make this a helper and it can also be used in postgres?

GeorgeSittas: postgres doesn't suffer from this problem, but maybe I could just simplify in the base generator instead of overriding so that all dialects can leverage it?

tobymao: postgres also copies an expression and calls simplify

GeorgeSittas: Ah that's what you meant, I thought you were talking about simplifying the limit expression. Sure, yea.

PR #1972, opened 2023-07-28
Refactor(optimizer): improve handling of DDL optimization

GeorgeSittas:
cc @gtoonstra

Maybe the name `DDL` isn't the best for this, given that `INSERT` is actually in the DML category..

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -990,7 +990,28 @@ class Uncache(Expression):
     arg_types = {"this": True, "exists": False}
 
 
-class Create(Expression):
+class DDL(Expression):
+    @property
+    def ctes(self):
+        with_ = self.args.get("with")
+        if not with_:
+            return []
+        return with_.expressions
+
+    @property
+    def named_selects(self) -> t.List[str]:
+        if isinstance(self.expression, Subqueryable):
+            return self.expression.named_selects
+        return []
+
+    @property
+    def selects(self) -> t.List[Expression]:
+        if isinstance(self.expression, Subqueryable):
+            return self.expression.selects
+        return []
+
+
+class Create(DDL):
     arg_types = {
         "with": False,
         "this": True,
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -990,7 +990,7 @@ class Uncache(Expression):
     arg_types = {"this": True, "exists": False}
 
 
-class TableChange(Expression):
+class DDL(Expression):
```
GeorgeSittas, line 993: I didn't have `UPDATE` inherit from this because it doesn't have a `Subqueryable` expression like `CREATE` and `INSERT`, so it'd need more effort to qualify / optimize its `SET .. FROM ..` syntax.

sqlglot/optimizer/scope.py
```diff
@@ -746,10 +747,16 @@ def _traverse_udtfs(scope):
...
 
-def _traverse_create(scope):
+def _traverse_ddl(scope):
     yield from _traverse_ctes(scope)
-    yield from _traverse_subqueries(scope)
-    yield from _traverse_tables(Scope(scope.expression.expression))
+
+    query_scope = scope.branch(
+        scope.expression.expression, scope_type=ScopeType.DERIVED_TABLE, chain_sources=scope.sources
+    )
```
GeorgeSittas, line 755: Picked `DERIVED_TABLE` here because `ROOT` is already taken for the DDL statement itself. Might not be the cleanest approach but I think it's fine for now?

GeorgeSittas:
Regarding the comment I left here https://github.com/tobymao/sqlglot/pull/1949#discussion_r1274356815, it seems like this is a known "issue" based on:

https://github.com/tobymao/sqlglot/blob/9d67283b3185baa43f2591a75c01493455948d40/tests/fixtures/optimizer/qualify_tables.sql#L17-L20

tobymao (approved):

PR #1973, opened 2023-07-28
Fix(tsql): improve UDF parsing

### LINKED ISSUE #1966: Parsing tsql stored proc with declare variable
**Error description**
I get the following sqlglot.errors.ParseError: Invalid expression / Unexpected token Line 4, Col: 28. which is the datatype varchar. 

I have look through the library and testet with and without the variable and it might be related to the way that the chunks is made when parsing. 

```sql

sql = """
CREATE PROC [dbo].[transform_proc] AS 

DECLARE @CurrentDate VARCHAR(20);
SET @CurrentDate = CONVERT(VARCHAR(20), GETDATE(), 120);

CREATE TABLE [target_schema].[target_table]
WITH (DISTRIBUTION = REPLICATE, HEAP)
AS 

SELECT 
	@CurrentDate AS DWCreatedDate
FROM source_schema.sourcetable;
"""

for table in parse_one(sql, dialect='tsql').find_all(exp.Table):
    print(table.db + '.' + table.name)
```

**Official Documentation**
https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-develop-ctas

GeorgeSittas:
Fixes #1966


sqlglot/expressions.py
```diff
@@ -1931,6 +1931,10 @@ class EngineProperty(Property):
     arg_types = {"this": True}
 
 
+class HeapProperty(Property):
+    arg_types = {}
```
GeorgeSittas, line 1935: @bayees can you make sure that `HEAP` is a valid table option for T-SQL? I didn't find it in the documentation: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16

tobymao (approved):

sqlglot/expressions.py
GeorgeSittas: Nvm I just checked the link you provided in the issue: https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-develop-ctas

PR #1976, opened 2023-07-28
Fix(parser): parse placeholder as fallback for boolean, null, star

### LINKED ISSUE #1975: MySQL tokenizer errors on "IS ?"
Hello,
I am trying out sqlglot and ran into an error when testing various parameterized mysql queries.

The following works just fine
```
query = "SELECT a FROM b WHERE c IS TRUE;"
parse_one(query, dialect='mysql')

query = "SELECT a FROM b WHERE c = ?;"
parse_one(query, dialect='mysql')
```

But when a placeholder is used with the `IS` operator it errors.
```
query = "SELECT a FROM b WHERE c IS ?;"
parse_one(query, dialect='mysql')
```
The error
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 26.
  SELECT a FROM b WHERE c IS ?;
```

I tried searching around the closed issues to see if this was previously discussed, as well as in slack, but I wasn't able to find anything.

Any ideas what's going on? Is this intentional?

EDIT: grammar

GeorgeSittas:
Fixes #1975


sqlglot/parser.py
```diff
@@ -4299,19 +4299,19 @@ def _parse_var_or_string(self) -> t.Optional[exp.Expression]:
...
             return self.PRIMARY_PARSERS[TokenType.TRUE](self, self._prev)
         if self._match(TokenType.FALSE):
             return self.PRIMARY_PARSERS[TokenType.FALSE](self, self._prev)
-        return None
+        return self._parse_placeholder()
 
     def _parse_star(self) -> t.Optional[exp.Expression]:
         if self._match(TokenType.STAR):
             return self.PRIMARY_PARSERS[TokenType.STAR](self, self._prev)
-        return None
```
tobymao, line 4314: is this one necessari

GeorgeSittas: not really just added it for consistency, happy to remove it

PR #1983, opened 2023-07-31
Fix!(clickhouse): preserve original function names instead of uppercasing

### LINKED ISSUE #1982: invalid transpilation of clickhouse `isNaN` function

```
In [9]: import sqlglot as sg

In [10]: sg.__version__
Out[10]: '17.8.6'

In [11]: sg.parse_one("select isNaN(1.0)", read="clickhouse").sql(dialect="clickhouse")
Out[11]: 'SELECT IS_NAN(1.0)'
```

Expected output is:

```
Out[11]: 'SELECT isNaN(1.0)'
```

**Official Documentation**

https://clickhouse.com/docs/en/sql-reference/functions/other-functions#isnanx

Is there a way I can disable function normalization for any and all function calls for clickhouse?

These incorrect translation bugs keep popping up and I'd like for function calls to pass through unchanged if possible.

GeorgeSittas:
Fixes #1982

Some case-sensitive clickhouse functions have a corresponding `Expression` class and so when we parse them, their original name is discarded and we use whatever we have in `_sql_names` at generation time, which is not valid.

Changing all `Func`s to store their name in `this` like we do for `Anonymous` is cumbersome, because we have many `Func`s that reserve `this` to represent an argument. Instead, in this approach I used the `Expression._meta` attribute to preserve the original name, because it felt like the least intrusive way to fix this issue.

cc: @cpcloud

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -225,7 +225,6 @@ class ClickHouse(Dialect):
             anonymous: bool = False,
             optional_parens: bool = True,
         ) -> t.Optional[exp.Expression]:
-            name = self._curr and self._curr.text
             func = super()._parse_function(
                 functions=functions, anonymous=anonymous, optional_parens=optional_parens
             )
@@ -240,9 +239,6 @@ class ClickHouse(Dialect):
                         expressions=func.expressions,
                         params=params,
                     )
-
-            if func and name:
-                func.meta["name"] = name
 
             return func
 
@@ -356,9 +352,6 @@ class ClickHouse(Dialect):
             "FUNCTION",
             "NAMED COLLECTION",
         }
-
-        def function_name(self, expression: exp.Func) -> str:
-            return (expression._meta and expression.meta.get("name")) or expression.sql_name()
 
         def safeconcat_sql(self, expression: exp.SafeConcat) -> str:
             # Clickhouse errors out if we try to cast a NULL value to TEXT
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -225,6 +225,7 @@ def _parse_function(
             anonymous: bool = False,
             optional_parens: bool = True,
         ) -> t.Optional[exp.Expression]:
+            name = self._curr and self._curr.text
```
tobymao, line 228: can we move all of this to main parsers / generators with a flag?

sqlglot/generator.py
```diff
@@ -2379,7 +2379,11 @@ def function_fallback_sql(self, expression: exp.Func) -> str:
             elif arg_value is not None:
                 args.append(arg_value)
 
-        return self.func(expression.sql_name(), *args)
+        name = expression.sql_name()
```
tobymao, line 2382: this should be called outside the block

so if self.normalize_functions... else

GeorgeSittas: added an `else` block to set it there

sqlglot/generator.py
```diff
@@ -2379,7 +2379,12 @@ def function_fallback_sql(self, expression: exp.Func) -> str:
             elif arg_value is not None:
                 args.append(arg_value)
 
-        return self.func(expression.sql_name(), *args)
+        if not self.normalize_functions:
```
tobymao, line 2382: invert it

tobymao (approved):

PR #1988, opened 2023-08-01
Fix(optimizer): wrap scalar subquery replacement in a MAX call

### LINKED ISSUE #1987: Nested Query Bug - Column '_u_0._col_0' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.
<h3>Code snippet:</h3>

```python
import sqlglot
from sqlglot.optimizer import optimize, RULES as optimize_rules
DCT_RULES = {x.__name__ : x for x in optimize_rules}
DCT_RULES.pop('quote_identifiers')
optimize_rules = tuple(list(DCT_RULES.values()))

if __name__ == '__main__':

  sqlglot_metadata = {
    "catalogo_proveedores": {
      "ID_PROVEEDOR": "INT",
      "PROVEEDOR": "STRING"
    },
    "compra_real": {
      "ANNUAL": "STRING",
      "TRIMESTRE": "STRING",
      "PIEZAS": "FLOAT",
      "PRECIO": "FLOAT",
      "IMPORTE": "FLOAT",
      "CLAVE_CBYCM": "INT",
      "ID_PROVEEDOR": "INT"
    },
  }
  # This query is working in this form
  sql_query = """SELECT PROVEEDOR, SUM(IMPORTE) / (SELECT SUM(IMPORTE) FROM compra_real WHERE CLAVE_CBYCM = 2133 AND ANNUAL = '2022') * 100 AS Percentage_Imports FROM compra_real INNER JOIN catalogo_proveedores ON compra_real.ID_PROVEEDOR = catalogo_proveedores.ID_PROVEEDOR WHERE CLAVE_CBYCM = 2133 AND ANNUAL = '2022' GROUP BY PROVEEDOR"""

  expr = optimize(
    expression=sqlglot.parse_one(sql_query, read="tsql"),
    schema=sqlglot_metadata,
    rules=optimize_rules,
  )

  print(expr.sql(pretty=True, dialect="tsql"))
```

<h3>Optimised query</h3>

```
WITH _u_0 AS (
  SELECT
    SUM(compra_real.importe) AS _col_0
  FROM compra_real AS compra_real
  WHERE
    compra_real.annual = '2022' AND compra_real.clave_cbycm = 2133
)
SELECT
  catalogo_proveedores.proveedor AS proveedor,
  SUM(compra_real.importe) / _u_0._col_0 * 100 AS percentage_imports
FROM compra_real AS compra_real
CROSS JOIN _u_0 AS _u_0
JOIN catalogo_proveedores AS catalogo_proveedores
  ON compra_real.id_proveedor = catalogo_proveedores.id_proveedor
WHERE
  compra_real.annual = '2022' AND compra_real.clave_cbycm = 2133
GROUP BY
  catalogo_proveedores.proveedor
```

<h3>The error:</h3>

```
Column '_u_0._col_0' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.
```

<h3>Expected output:</h3>
If the GROUP BY clause contains also _u_0._col_0, then the query will be executed successfully.

GeorgeSittas:
Fixes #1987

Tested this in SQL fiddle, seems to be good.

### DIFF
--- a/sqlglot/optimizer/unnest_subqueries.py
+++ b/sqlglot/optimizer/unnest_subqueries.py
@@ -53,15 +53,11 @@ def unnest(select, parent_select, next_alias_name):
         clause = predicate.find_ancestor(exp.Having, exp.Where, exp.Join)
         clause_parent_select = clause.parent_select if clause else None
 
-        if (
-            (isinstance(clause, exp.Having) and clause_parent_select is parent_select)
-            or not clause
-            or (
-                clause_parent_select is not parent_select
-                and (
-                    parent_select.args.get("group")
-                    or any(projection.find(exp.AggFunc) for projection in parent_select.selects)
-                )
+        if (isinstance(clause, exp.Having) and clause_parent_select is parent_select) or (
+            (not clause or clause_parent_select is not parent_select)
+            and (
+                parent_select.args.get("group")
+                or any(projection.find(exp.AggFunc) for projection in parent_select.selects)
             )
         ):
             column = exp.Max(this=column)
--- a/sqlglot/optimizer/unnest_subqueries.py
+++ b/sqlglot/optimizer/unnest_subqueries.py
@@ -48,13 +48,21 @@ def unnest(select, parent_select, next_alias_name):
 
     # This subquery returns a scalar and can just be converted to a cross join
     if not isinstance(predicate, (exp.In, exp.Any)):
-        having = predicate.find_ancestor(exp.Having)
         column = exp.column(select.selects[0].alias_or_name, alias)
 
+        clause = predicate.find_ancestor(exp.Having, exp.Where, exp.Join)
+        clause_parent_select = clause.parent_select if clause else None
+
         if (
-            (having and having.parent_select is parent_select)
-            or parent_select.args.get("group")
-            or any(projection.find(exp.AggFunc) for projection in parent_select.selects)
+            (isinstance(clause, exp.Having) and clause_parent_select is parent_select)
+            or not clause
+            or (
+                clause_parent_select is not parent_select
+                and (
+                    parent_select.args.get("group")
+                    or any(projection.find(exp.AggFunc) for projection in parent_select.selects)
+                )
+            )
         ):
             column = exp.Max(this=column)
 
--- a/sqlglot/optimizer/unnest_subqueries.py
+++ b/sqlglot/optimizer/unnest_subqueries.py
@@ -46,20 +46,20 @@ def unnest(select, parent_select, next_alias_name):
     if not predicate or parent_select is not predicate.parent_select:
         return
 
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/unnest_subqueries.py
```diff
@@ -46,20 +46,20 @@ def unnest(select, parent_select, next_alias_name):
...
+    # This subquery returns a scalar and can just be converted to a cross join
     if not isinstance(predicate, (exp.In, exp.Any)):
         having = predicate.find_ancestor(exp.Having)
         column = exp.column(select.selects[0].alias_or_name, alias)
-        if having and having.parent_select is parent_select:
+
+        if (
+            (having and having.parent_select is parent_select)
+            or any(projection.find(exp.AggFunc) for projection in parent_select.selects)
+            or parent_select.args.get("group")
```
tobymao, line 57: put the group by first

GeorgeSittas: Good point, thanks

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
```diff
@@ -6589,9 +6609,9 @@ WITH "cs_or_ws_sales" AS (
     ON "customer_address"."ca_county" = "store"."s_county"
     AND "customer_address"."ca_state" = "store"."s_state"
   JOIN "_u_0" AS "_u_0"
-    ON "date_dim"."d_month_seq" >= "_u_0"."_col_0"
+    ON "date_dim"."d_month_seq" >= MAX("_u_0"."_col_0")
```
tobymao, line 6612: i don't think this one is correct

GeorgeSittas: Made a few changes to address this, let me know when you take another look

sqlglot/optimizer/unnest_subqueries.py
```diff
@@ -46,20 +46,28 @@ def unnest(select, parent_select, next_alias_name):
...
+            (isinstance(clause, exp.Having) and clause_parent_select is parent_select)
+            or not clause
+            or (
+                clause_parent_select is not parent_select
+                and (
+                    parent_select.args.get("group")
+                    or any(projection.find(exp.AggFunc) for projection in parent_select.selects)
+                )
+            )
+        ):
```
GeorgeSittas, line 66: This became a bit more complex, but on the plus-side we'll avoid computing `parent_select` twice for `clause`, or repeating `column = exp.Max(this=column)`. Can try to break it up into an `else` branch if it's too messy.

GeorgeSittas: Ah I noticed this is incorrect actually, let me fix that really quick..

PR #1991, opened 2023-08-03
Feat: add terse coalesce operator

z3z1ma:
One of the most common operations we perform in queries other than casting is coalescing. Offering a terse syntax available to all dialects just as you do with the `::` casting is hugely beneficial. This feature is also immediately attractive within a SQLMesh project and is just generally a pretty neat showcase of the power of sqlglot.

`SELECT x ?? y FROM z` -> `SELECT COALESCE(x, y) FROM z`
`"SELECT a, b ?? 'No Data' FROM z"` -> `SELECT a, COALESCE(b, 'No Data') FROM z`

cc: @tobymao

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -369,7 +369,6 @@ class Parser(metaclass=_Parser):
         TokenType.CARET: exp.BitwiseXor,
         TokenType.PIPE: exp.BitwiseOr,
         TokenType.DPIPE: exp.DPipe,
-        TokenType.DQMARK: exp.Coalesce,
     }
 
     TERM = {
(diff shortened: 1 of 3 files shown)


z3z1ma:
While you could, probably detect `??` chained one after another and put them all into a single `COALESCE(.., .., ..)` function, its significantly simpler this way, probably more resilient, and just as correct as far as I can tell. And the code footprint to enable this is beautifully simple. 🤓

GeorgeSittas (approved):
LGTM, good idea Alex 👍

By the way, implementing the `??` operator in this way means that expressions like

```
a + b ?? c
```

will be parsed as `a + COALESCE(b, c)`. Perhaps this is something obvious, just wanted to point it out in case people don't feel this should be the operator's precedence. Of course, one can always wrap `a + b` in parentheses to produce `COALESCE((a + b), c)`.

tobymao:
@z3z1ma i'm not sure this is right, i would have thought this was closer to a bitwise operation like concat x || y. can we move it out of column ops?

sqlglot/parser.py
```diff
@@ -3222,6 +3228,10 @@ def _parse_column_ops(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.
                 field = self._parse_types()
                 if not field:
                     self.raise_error("Expected type")
+            elif op_token == TokenType.DQMARK:
```
tobymao, line 3231: i don't think this is necessary

sqlglot/parser.py
```diff
@@ -369,6 +369,7 @@ class Parser(metaclass=_Parser):
         TokenType.CARET: exp.BitwiseXor,
         TokenType.PIPE: exp.BitwiseOr,
         TokenType.DPIPE: exp.DPipe,
+        TokenType.DQMARK: exp.Coalesce,
```
tobymao, line 372: remove this

sqlglot/parser.py
```diff
@@ -3020,9 +3021,14 @@ def _parse_bitwise(self) -> t.Optional[exp.Expression]:
 
         while True:
             if self._match_set(self.BITWISE):
-                this = self.expression(
-                    self.BITWISE[self._prev.token_type], this=this, expression=self._parse_term()
-                )
+                if self._prev.token_type == TokenType.DQMARK:
```
tobymao, line 3024: can you just do a self._match() instead of this?

PR #1993, opened 2023-08-04
Fix(parser, duckdb): decode/encode in duckdb don't take charset

charsmith:
Duckdb doesn't take a param for charset on [decode/encode](https://duckdb.org/docs/sql/functions/blob.html).

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -5,7 +5,7 @@ from enum import Enum
 
 from sqlglot import exp
 from sqlglot._typing import E
-from sqlglot.errors import ParseError, UnsupportedError
+from sqlglot.errors import ParseError
 from sqlglot.generator import Generator
 from sqlglot.helper import flatten, seq_get
 from sqlglot.parser import Parser
@@ -577,7 +577,7 @@ def encode_decode_sql(self: Generator, expression: exp.Expression, name: str) ->
         charset = expression.args["charset"]
 
         if charset.name.lower() != "utf-8":
-            raise UnsupportedError(f"Unsupported charset {charset}")
+            self.unsupported(f"Unsupported charset {charset}")
 
         expression = expression.copy()
         del expression.args["charset"]
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -5,7 +5,7 @@ from enum import Enum
 
 from sqlglot import exp
 from sqlglot._typing import E
-from sqlglot.errors import ParseError
+from sqlglot.errors import ParseError, UnsupportedError
 from sqlglot.generator import Generator
 from sqlglot.helper import flatten, seq_get
 from sqlglot.parser import Parser
@@ -569,6 +569,20 @@ def timestrtotime_sql(self: Generator, expression: exp.TimeStrToTime) -> str:
 
 def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
     return self.sql(exp.cast(expression.this, "date"))
+
+
+# Used for Presto and Duckdb which use functions that don't support charset, and assume utf-8
+def encode_decode_sql(self: Generator, expression: exp.Expression, name: str) -> str:
+    if "charset" in expression.args:
+        charset = expression.args["charset"]
+
+        if charset.name.lower() != "utf-8":
+            raise UnsupportedError(f"Unsupported charset {charset}")
+
+        expression = expression.copy()
+        del expression.args["charset"]
+
+    return rename_func(name)(self, expression)
 
 
 def min_or_least(self: Generator, expression: exp.Min) -> str:
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -11,6 +11,7 @@ from sqlglot.dialects.dialect import (
     binary_from_function,
     date_trunc_to_time,
     datestrtodate_sql,
+    encode_decode_sql,
     format_time_lambda,
     no_comment_column_constraint_sql,
     no_properties_sql,
@@ -25,9 +26,11 @@ from sqlglot.dialects.dialect import (
     timestrtotime_sql,
     ts_or_ds_to_date_sql,
 )
-from sqlglot.errors import UnsupportedError
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
 
 
 def _ts_or_ds_add_sql(self: generator.Generator, expression: exp.TsOrDsAdd) -> str:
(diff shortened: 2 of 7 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -96,6 +97,21 @@ def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> s
...
+
+def _decode_sql(self: generator.Generator, expression: exp.Decode) -> str:
+    _ensure_utf8(expression.args["charset"])
+    return self.func("DECODE", expression.this, expression.args.get("replace"))
+
+
+def _encode_sql(self: generator.Generator, expression: exp.Encode) -> str:
+    _ensure_utf8(expression.args["charset"])
+    return f"ENCODE({self.sql(expression, 'this')})"
+
```
GeorgeSittas, line 113: This section is (almost) copy-pasted from presto.py -- can we move these in dialect.py and reuse them in both presto and duckdb instead of duplicating it here?

sqlglot/dialects/duckdb.py
```diff
@@ -96,6 +97,21 @@ def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> s
...
 
 
+def _ensure_utf8(charset: exp.Literal) -> None:
+    if charset.name.lower() != "utf-8":
+        raise UnsupportedError(f"Unsupported charset {charset}")
+
+
+def _decode_sql(self: generator.Generator, expression: exp.Decode) -> str:
+    _ensure_utf8(expression.args["charset"])
+    return self.func("DECODE", expression.this, expression.args.get("replace"))
```
GeorgeSittas, line 107: What is `replace` here? Seems like [duckdb only supports](https://duckdb.org/docs/sql/functions/blob.html) a `blob` argument in `DECODE`?

tests/dialects/test_duckdb.py
```diff
@@ -583,6 +583,48 @@ def test_bool_or(self):
             write={"duckdb": "SELECT a, BOOL_OR(b) FROM table GROUP BY a"},
         )
 
+    def test_encode_decode(self):
```
GeorgeSittas, line 586: Let's include `"duckdb": ...` in these tests instead of just spark and presto.

GeorgeSittas (commented):
Thanks for the PR, left a few of comments.

I'm not sure if we should alter the base parser for this. I would personally just add a couple of entries in DuckDB's `Parser.FUNCTIONS` (see how it's done in presto.py), or update the presto dialect so that it routes to `_parse_decode/encode` in the base parser and reuse that in duckdb.

sqlglot/parser.py
```diff
@@ -3947,6 +3953,16 @@ def _parse_decode(self) -> t.Optional[exp.Decode | exp.Case]:
...
 
+    def _parse_encode(self) -> exp.Encode:
+        args = self._parse_csv(self._parse_conjunction)
+
+        if len(args) < 2:
+            return self.expression(
+                exp.Encode, this=seq_get(args, 0), charset=exp.Literal.string("utf-8")
+            )
+
+        return self.expression(exp.Encode, this=seq_get(args, 0), charset=seq_get(args, 1))
```
GeorgeSittas, line 3964: Ditto about duplication: this is almost the same as `_parse_decode`. If we decide to keep these, we should dry them up and have them receive the expression type as a parameter in order to reuse the logic for both `Encode` and `Decode`.

sqlglot/dialects/duckdb.py
```diff
@@ -96,6 +97,21 @@ def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> s
     return f"CAST({sql} AS TEXT)"
 
 
+def _ensure_utf8(charset: exp.Literal) -> None:
+    if charset.name.lower() != "utf-8":
+        raise UnsupportedError(f"Unsupported charset {charset}")
```
tobymao, line 102: this is not quite correct, you should call self.unspported on the generator

sqlglot/dialects/duckdb.py
```diff
@@ -96,6 +97,21 @@ def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> s
...
+
+
+def _decode_sql(self: generator.Generator, expression: exp.Decode) -> str:
+    _ensure_utf8(expression.args["charset"])
```
tobymao, line 106: charset may not exist since you made it false, so use get

sqlglot/dialects/duckdb.py
charsmith: Yes, this was code I copied from the presto dialect a couple of weeks ago to make this work, and I didn't notice that I left replace. Will fix.

sqlglot/dialects/dialect.py
```diff
@@ -571,6 +571,20 @@ def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
...
+# Used for Presto and Duckdb which use functions that don't support charset, and assume utf-8
+def encode_decode_sql(self: Generator, expression: exp.Expression, name: str) -> str:
+    if "charset" in expression.args:
+        charset = expression.args["charset"]
+
+        if charset.name.lower() != "utf-8":
+            self.unsupported(f"Unsupported charset {charset}")
+
+        expression = expression.copy()
+        del expression.args["charset"]
```
charsmith, line 583: I don't love this, happy to change this if there is a better way.

tests/dialects/test_duckdb.py
charsmith: Added.

sqlglot/dialects/duckdb.py
charsmith: Done

sqlglot/dialects/duckdb.py
charsmith: Now that I have refactored the parser, I think that charset is always available.

sqlglot/parser.py
charsmith: Refactored this a bit.  Hope this is closer to what you had in mind.

sqlglot/dialects/dialect.py
```diff
@@ -571,6 +571,20 @@ def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
     return self.sql(exp.cast(expression.this, "date"))
 
 
+# Used for Presto and Duckdb which use functions that don't support charset, and assume utf-8
+def encode_decode_sql(self: Generator, expression: exp.Expression, name: str) -> str:
+    if "charset" in expression.args:
+        charset = expression.args["charset"]
```
GeorgeSittas, line 577:
```suggestion
    charset = expression.args.get("charset")
    if charset:
```

sqlglot/dialects/duckdb.py
```diff
@@ -167,6 +171,12 @@ class Parser(parser.Parser):
             "XOR": binary_from_function(exp.BitwiseXor),
         }
 
+        FUNCTION_PARSERS = {
+            **parser.Parser.FUNCTION_PARSERS,
+            "ENCODE": lambda self: self._parse_encode_decode(exp.Encode),
+            "DECODE": lambda self: self._parse_encode_decode(exp.Decode),
+        }
```
GeorgeSittas, line 178: We can just do these inline I think, like:

```python
            "DECODE": lambda self: self.expression(
                exp.Decode, this=self._parse_conjunction(), charset=exp.Literal.string("utf-8")
            ),
            "ENCODE": lambda self: self.expression(
                exp.Encode, this=self._parse_conjunction(), charset=exp.Literal.string("utf-8")
            ),
```

sqlglot/dialects/dialect.py
```diff
@@ -571,6 +571,20 @@ def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
     return self.sql(exp.cast(expression.this, "date"))
 
 
+# Used for Presto and Duckdb which use functions that don't support charset, and assume utf-8
+def encode_decode_sql(self: Generator, expression: exp.Expression, name: str) -> str:
```
GeorgeSittas, line 575: I think you also need a boolean flag here to control whether or not `replace` will be generated.

sqlglot/dialects/dialect.py
GeorgeSittas: I think you can just do `return self.func(name, expression.this, expression.args.get("replace") if replace else None)` (see other comment about `replace` flag).

GeorgeSittas:
Looks good, just a few final comments.

GeorgeSittas:
@charsmith I'll wrap this up, should be just a couple of changes. Thanks!

PR #1995, opened 2023-08-04
fix(parser, redshift): parse dateadd alias date_add

gafeol:
Redshift also supports using `DATE_ADD` as an alias of `DATEADD`. Right now, they both functions are parsed differently:

```
sql = "SELECT DATE_ADD('day', 1, date('2023-01-01'))"
sqlglot.transpile(sql, read='redshift', write='redshift')
> ["SELECT DATEADD(day, 1, 'day')"]

sql = "SELECT DATEADD('day', 1, date('2023-01-01'))"
sqlglot.transpile(sql, read='redshift', write='redshift')
> ["SELECT DATEADD(day, 1, DATE('2023-01-01'))
```

Added the same parser used on DATEADD to fix this difference.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -37,6 +37,11 @@ class Redshift(Postgres):
                 expression=seq_get(args, 1),
                 unit=seq_get(args, 0),
             ),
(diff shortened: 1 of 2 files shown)


tobymao:
thanks for the pr! can you make it a function so we can DRY it up?

PR #2002, opened 2023-08-07
Fix(optimizer): TableAlias column names must shadow source columns

GeorgeSittas:
This is still a WIP, am currently trying to fix some bugs in `pushdown_projections`.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -192,6 +192,13 @@ class Expression(metaclass=_Expression):
         return self.text("alias")
 
     @property
+    def alias_column_names(self) -> t.List[str]:
+        table_alias = self.args.get("alias")
+        if not table_alias:
+            return []
+        return [c.name for c in table_alias.args.get("columns") or []]
+
+    @property
     def name(self) -> str:
         return self.text("this")
 
@@ -883,13 +890,6 @@ class Predicate(Condition):
 
 
 class DerivedTable(Expression):
-    @property
-    def alias_column_names(self) -> t.List[str]:
-        table_alias = self.args.get("alias")
-        if not table_alias:
-            return []
-        return [c.name for c in table_alias.args.get("columns") or []]
-
     @property
     def selects(self) -> t.List[Expression]:
         return self.this.selects if isinstance(self.this, Subqueryable) else []
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -7,7 +7,7 @@ from sqlglot import alias, exp
 from sqlglot._typing import E
 from sqlglot.dialects.dialect import Dialect, DialectType
 from sqlglot.errors import OptimizeError
-from sqlglot.helper import seq_get
+from sqlglot.helper import ensure_list, seq_get
 from sqlglot.optimizer.scope import Scope, traverse_scope, walk_in_scope
 from sqlglot.schema import Schema, ensure_schema
 
@@ -58,6 +58,7 @@ def qualify_columns(
         if not isinstance(scope.expression, exp.UDTF):
             _expand_stars(scope, resolver, using_column_tables, pseudocolumns)
             _qualify_outputs(scope)
+
         _expand_group_by(scope)
         _expand_order_by(scope, resolver)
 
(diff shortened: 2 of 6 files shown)


sqlglot/expressions.py
```diff
@@ -191,6 +191,13 @@ def alias(self) -> str:
             return self.args["alias"].name
         return self.text("alias")
 
+    @property
+    def alias_column_names(self) -> t.List[str]:
+        table_alias = self.args.get("alias")
+        if not table_alias:
+            return []
+        return [c.name for c in table_alias.args.get("columns") or []]
+
```
GeorgeSittas, line 200: Moved here because it wasn't available to `Table`, although it can also apply to it. Since we have some other helpers like `alias` in `Expression`, this change felt ok.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -528,61 +539,82 @@ def get_table(self, column_name: str) -> t.Optional[exp.Identifier]:
...
-        return source.expression.named_selects
+        return [alias or name for (name, alias) in itertools.zip_longest(columns, column_aliases)]
 
-    def _get_all_source_columns(self):
+    def _get_all_source_columns(self) -> t.Dict[str, t.List[str]]:
         if self._source_columns is None:
             self._source_columns = {
-                k: self.get_source_columns(k)
-                for k in itertools.chain(self.scope.selected_sources, self.scope.lateral_sources)
+                name: self.get_source_columns(name, source=seq_get(ensure_list(source), 0))
```
tobymao, line 585: why are you doing ensure_list seq_get?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -465,6 +472,10 @@ def _qualify_outputs(scope: Scope):
     scope.expression.set("expressions", new_selections)
 
 
+def _get_source_node(scope: Scope, source_name: str) -> t.Optional[exp.Expression]:
+    return seq_get(scope.selected_sources.get(source_name) or [], 0)
```
tobymao, line 476: why was this added?

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: `selected_sources` is a dict from `str` to a `tuple`, from which we want the first element. On the other hand, `lateral_sources` maps a `str` directly to a `Table | Scope` object.

The motivation is to always get the source code, regardless of whether we're examining a `selected_source` or a `lateral_source`.

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: Reduce repetition - this piece of logic is used in 4 places where we want to map a source name back to its node.

GeorgeSittas: I'll try to see if I can refactor and get rid of this altogether, I don't like this addition either.

GeorgeSittas:
I came up with a workaround for the projection pushdown issues I had encountered.

```sql
WITH cte(x, y) AS (SELECT 1, 2) SELECT cte.* FROM cte AS cte(a);  -- before qualifying
WITH cte AS (SELECT 1 AS x, 2 AS y) SELECT cte.a AS a, cte.y AS y FROM cte AS cte(a);  -- after qualifying
```

The issue was that in the above test case, `cte.x` was showing up as an unused projection and was getting eliminated in the projection pushdown rule, so `cte` would end up having just a single projection because of that, hence yielding an invalid query.

The workaround I implemented is that now I count how many column aliases there are per selected source in the parent column and then ensure that the inner scope has at least as many projections preserved in the "elimination" phase. So, for this specific example, we count that the `cte` selected source in the outer scope has one column alias and hence we need to at least preserve `x`, because there's a one-to-one correspondence between the alias list and the projections.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -528,61 +529,78 @@ def get_table(self, column_name: str) -> t.Optional[exp.Identifier]:
...
+            columns = self.schema.column_names(original_source, only_visible)
+        elif isinstance(original_source, Scope) and isinstance(
+            original_source.expression, exp.Values
+        ):
+            columns = original_source.expression.alias_column_names
+        else:
+            # Otherwise, if referencing another scope, return that scope's named selects
+            columns = original_source.expression.named_selects
+
+        source = seq_get(self.scope.selected_sources.get(name) or [], 0)
```
tobymao, line 558: let's just do

source = self.scope.selected_sources[name] if name in ... else None

tobymao (approved):

sqlglot/optimizer/qualify_columns.py
GeorgeSittas: Refactored it, a little bit differently but imo clear too. Happy to change later if you prefer this form.

PR #2006, opened 2023-08-08
Feat: add apache doris dialect

liujiwen-up:
Implemented dialect support for Apache Doris


sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+        CAST_MAPPING = {}
+
+        TYPE_MAPPING = {
+            **MySQL.Generator.TYPE_MAPPING,
+            exp.DataType.Type.TEXT: "STRING",
+            exp.DataType.Type.TIMESTAMP: "DATETIME",
+            exp.DataType.Type.TIMESTAMPTZ: "DATETIME",
+        }
+
+        TRANSFORMS = {
```
GeorgeSittas, line 74: Nit: can we try to sort these alphabetically so they're easier to read / search?

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+            **MySQL.Generator.TRANSFORMS,
+            exp.ApproxDistinct: approx_count_distinct_sql,
+            exp.JSONExtractScalar: arrow_json_extract_sql,
+            exp.JSONExtract: arrow_json_extract_sql,
+            exp.DateDiff: rename_func("DATEDIFF"),
```
GeorgeSittas, line 79: This should be removed, the correct transform already exists in MySQL's dict:

https://github.com/tobymao/sqlglot/blob/c73790d0249cf0f21f855ed4921c9a8727e2d5ff/sqlglot/dialects/mysql.py#L503

According to Doris' SQL manual you can only have two arguments in `DATEDIFF`, so your override will generate incorrect SQL when you transpile from dialects that can also accept a `unit` argument.

It seems like Starrocks overrides similarly, seems like a bug based on https://docs.starrocks.io/en-us/2.5/sql-reference/sql-functions/date-time-functions/datediff

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+            exp.RegexpLike: rename_func("REGEXP"),
```
GeorgeSittas, line 80: Need to add an entry in `Parser.FUNCTIONS` for this:

```python
"REGEXP": exp.RegexpLike.from_arg_list
```

Ditto about Starrocks suffering from the same issue.

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+            exp.Coalesce: rename_func("NVL"),
+            exp.CurrentTimestamp: lambda self, e: "NOW()",
```
GeorgeSittas, line 82:
```suggestion
            exp.CurrentTimestamp: lambda *_: "NOW()",
```

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+            exp.TimeToStr: lambda self, e: f"DATE_FORMAT({self.sql(e, 'this')}, {self.format_time(e)})",
```
GeorgeSittas, line 83: This should be removed, already exists in MySQL.

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+            exp.ToChar: lambda self, e: f"DATE_FORMAT({self.sql(e, 'this')}, {self.format_time(e)})",
+            exp.StrToUnix: lambda self, e: f"UNIX_TIMESTAMP({self.sql(e, 'this')}, {self.format_time(e)})",
+            exp.TimestampTrunc: lambda self, e: self.func(
+                "DATE_TRUNC", exp.Literal.string(e.text("unit")), e.this
+            ),
```
GeorgeSittas, line 88: This is also used like that in spark and starrocks. Perhaps factor it out into a helper?

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+            exp.ArrayAgg: rename_func("COLLECT_LIST"),
+            exp.SetAgg: rename_func("COLLECT_SET"),
+            exp.TsOrDsAdd: lambda self, e: f"DATE_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')})",  # Only for day level
+            exp.TsOrDsToDate: _to_date_sql,
+            exp.Map: rename_func("ARRAY_MAP"),
+            exp.RegexpSplit: rename_func("SPLIT_BY_STRING"),
+            exp.Split: rename_func("SPLIT_BY_STRING"),
+            exp.Quantile: rename_func("PERCENTILE"),
+            exp.ApproxQuantile: rename_func("PERCENTILE_APPROX"),
+            exp.TimeStrToUnix: rename_func("UNIX_TIMESTAMP"),
```
GeorgeSittas, line 103: This should be removed, already exists in MySQL.

GeorgeSittas (commented):
Thanks for refactoring the previous PR.

I think we should revert all changes in the HTML files, because:
- They're not necessary since they're auto-generated
- There might be conflicts when the github action tries to generate them later.

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+def _time_format(
+    self: generator.Generator, expression: exp.UnixToStr | exp.StrToUnix
+) -> t.Optional[str]:
+    time_format = self.format_time(expression)
+    if time_format == Doris.TIME_FORMAT:
+        return None
+    return time_format
+
+
+class Doris(MySQL):
```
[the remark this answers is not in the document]
liujiwen-up: Because doris is compatible with most of the syntax of mysql, but with the development of the doris community, some newly added syntax is different from the usage of mysql. For the convenience of maintenance later, the doris dialect is added.

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+    DATE_FORMAT = "'yyyy-MM-dd'"
+    DATEINT_FORMAT = "'yyyyMMdd'"
+    TIME_FORMAT = "'yyyy-MM-dd HH:mm:ss'"
+
+    TIME_MAPPING = {
```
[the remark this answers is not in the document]
liujiwen-up: I will delete it for the time being, thank you

sqlglot/dialects/doris.py
liujiwen-up: OK, thanks

sqlglot/dialects/doris.py
GeorgeSittas: Ok so there's actually a detail I missed: Starrocks supports two variants, called `DATEDIFF` and `DATE_DIFF`. The main difference between is that only the latter supports the `unit` argument, so this needs more careful handling.

sqlglot/dialects/doris.py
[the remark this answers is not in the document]
liujiwen-up: Yes, mysql and doris have some similar mappings, but we found that some mappings are not supported when going from hive to doris, so we added some on the previous basis to adapt to the syntax conversion of hive to doris

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+        "%y": "%%Y",
+        "%": "%%",
+    }
+
+    class Parser(MySQL.Parser):
+        FUNCTIONS = {
+            **MySQL.Parser.FUNCTIONS,
+            "DATE_TRUNC": lambda args: exp.TimestampTrunc(
+                this=seq_get(args, 1), unit=seq_get(args, 0)
+            ),
```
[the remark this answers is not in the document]
liujiwen-up: Let's dry it out into a helper in dialect.py.
Sorry, I don't quite understand the meaning of this sentence, do you mean to implement this method in dialect.py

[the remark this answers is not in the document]
tobymao: yes dry means don’t repeat yourself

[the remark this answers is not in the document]
tobymao: please clean up any instance of copy and paste

sqlglot/dialects/doris.py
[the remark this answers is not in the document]
tobymao: how similar is it to starrocks? we support that as well

sqlglot/dialects/doris.py
liujiwen-up: I re-implemented this part, the usage form of doris is a little different

sqlglot/dialects/doris.py
liujiwen-up: ok,thanks

sqlglot/dialects/doris.py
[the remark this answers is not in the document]
liujiwen-up: The degree of compatibility with starrocks is similar to that of mysql. starrcoks came from fork doris and went out alone, but with the development and improvement of the community later, it will become more and more different from starrocks, and I plan to spend more time to perfect it Convert different data sources into doris.thanks

sqlglot/dialects/doris.py
liujiwen-up: ok,thanks

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,111 @@
...
+    rename_func,
+)
+from sqlglot.dialects.mysql import MySQL
+from sqlglot.helper import seq_get
+
+
+def _to_date_sql(self: generator.Generator, expression: exp.TsOrDsToDate) -> str:
+    this = self.sql(expression, "this")
+    self.format_time(expression)
+    return f"TO_DATE({this})"
```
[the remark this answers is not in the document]
liujiwen-up: Because the to_date() function in doris is the syntax https://doris.apache.org/zh-CN/docs/dev/sql-manual/sql-functions/date-time-functions/to_date, without parameters.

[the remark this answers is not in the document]
tobymao: this is not a valid link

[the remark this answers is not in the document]
liujiwen-up: https://doris.apache.org/zh-CN/docs/1.2/sql-manual/sql-functions/date-time-functions/to_date/

liujiwen-up:
Edited based on all comments, hopefully incorporated, thanks

GeorgeSittas:
Thanks @liujiwen-up, will review again soon

docs/sqlglot/dialects/dialect.html
```diff
@@ -1396,6 +1396,8 @@ <h1 class="modulename">
     
 
                             </div>
+
+
                             <div class="inherited">
```
GeorgeSittas, line 1401:
```suggestion
                            <div class="inherited">
```

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,105 @@
...
+            exp.JSONExtract: arrow_json_extract_sql,
+            exp.RegexpLike: rename_func("REGEXP"),
+            exp.RegexpSplit: rename_func("SPLIT_BY_STRING"),
+            exp.SetAgg: rename_func("COLLECT_SET"),
+            exp.StrToUnix: lambda self, e: f"UNIX_TIMESTAMP({self.sql(e, 'this')}, {self.format_time(e)})",
+            exp.Split: rename_func("SPLIT_BY_STRING"),
+            exp.TimeStrToDate: rename_func("TO_DATE"),
+            exp.ToChar: lambda self, e: f"DATE_FORMAT({self.sql(e, 'this')}, {self.format_time(e)})",
+            exp.TsOrDsAdd: lambda self, e: f"DATE_ADD({self.sql(e, 'this')}, {self.sql(e, 'expression')})",  # Only for day level
+            exp.TsOrDsToDate: _to_date_sql,
```
GeorgeSittas, line 94:
```suggestion
            exp.TsOrDsToDate: lambda self, e: self.func("TO_DATE", e.this),
```

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,105 @@
...
+    rename_func,
+)
+from sqlglot.dialects.mysql import MySQL
+from sqlglot.helper import seq_get
+
+
+def _to_date_sql(self: MySQL.Generator, expression: exp.TsOrDsToDate) -> str:
+    this = self.sql(expression, "this")
+    self.format_time(expression)
+    return f"TO_DATE({this})"
```
GeorgeSittas, line 18: This is unnecessary, let's get rid of it.

sqlglot/dialects/doris.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _time_format(
+    self: generator.Generator, expression: exp.UnixToStr | exp.StrToUnix
+) -> t.Optional[str]:
+    time_format = self.format_time(expression)
+    if time_format == Doris.TIME_FORMAT:
+        return None
+    return time_format
```
GeorgeSittas, line 27: We should factor this out into a helper in `dialect.py`, we do the exact same thing in Hive.

PR #2012, opened 2023-08-08
Feat(hive): improve transpilation of Bigquery's TIMESTAMP_ADD

GeorgeSittas:


tobymao (approved):

sqlglot/dialects/hive.py
```diff
@@ -426,6 +426,7 @@ class Generator(generator.Generator):
             exp.StrToTime: _str_to_time_sql,
             exp.StrToUnix: _str_to_unix_sql,
             exp.StructExtract: struct_extract_sql,
+            exp.TimestampAdd: lambda self, e: f"{self.sql(e, 'this')} + {self.sql(exp.Interval(this=e.expression.copy(), unit=e.text('unit') or 'DAY'))}",
```
tobymao, line 429: does hive support this? why not use  date add

GeorgeSittas: perhaps you're right, I'll investigate and try to simplify

GeorgeSittas: So for spark it seems like `dateadd` will not work because it only uses "days" as its granularity: https://spark.apache.org/docs/latest/api/sql/index.html#dateadd. Tried this and got the following failure:

```
spark-sql (default)> SELECT DATE_ADD(CAST('2008-12-25 15:30:00+00' AS TIMESTAMP), INTERVAL 10 MINUTE);
[DATATYPE_MISMATCH.UNEXPECTED_INPUT_TYPE] Cannot resolve "date_add(CAST(2008-12-25 15:30:00+00 AS TIMESTAMP), INTERVAL '10' MINUTE)" due to data type mismatch: Parameter 2 requires the ("INT" or "SMALLINT" or "TINYINT") type, however "INTERVAL '10' MINUTE" has the type "INTERVAL MINUTE".; line 1 pos 7;
'Project [unresolvedalias(date_add(cast(cast(2008-12-25 15:30:00+00 as timestamp) as date), INTERVAL '10' MINUTE), None)]
+- OneRowRelation
```

I'm not sure if adding a datetime with an interval in Hive is valid, but based on their docs `date_add` doesn't seem to help with intervals too. I can also move this change to just spark2 instead of hive.

Screenshot 2023-08-09 at 3 23 07 AM

tobymao: https://docs.databricks.com/en/sql/language-manual/functions/dateadd.html

GeorgeSittas: Ah, right, spark supports units too:

```
spark-sql (default)> SELECT dateadd(MINUTE, 10, TIMESTAMP'2022-03-31 00:00:00');
timestampadd(MINUTE, 10, TIMESTAMP '2022-03-31 00:00:00')
2022-03-31 00:10:00
Time taken: 0.272 seconds, Fetched 1 row(s)
```

So this change probably makes sense.

tobymao: i thought you just tried it and said it didn’t work

GeorgeSittas: I called it with different arguments, it doesn't work with an interval value. This `dateadd(unit, value, datetime)` variant is a newer one (not in their docs) and I forgot about it.

tobymao: i think timestamp add probabkt should default to date add

GeorgeSittas: Updated

PR #2019, opened 2023-08-09
feat: simplify COALESCE

barakalon:
This is a tricky one.

I don't think this is necessarily simpler _semantically_. But it does remove COALESCE functions, which is kind of a normalization.

And it _does_ optimize queries for StarRocks. Not sure about other engines.

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -486,13 +486,17 @@ def simplify_coalesce(expression):
     if isinstance(rightmost_arg, CONSTANTS) and isinstance(other, CONSTANTS):
         rightmost_arg.pop()
 
+        # Remove the COALESCE function. This is an optimization, skipping a simplify iteration,
+        # since we already remove COALESCE at the top of this function.
+        coalesce = coalesce if coalesce.expressions else coalesce.this
+
         return exp.or_(
             exp.and_(
-                coalesce.copy().is_(exp.null()).not_(),
-                expression.copy(),
+                coalesce.is_(exp.null()).not_(),
+                expression,
             ),
             exp.and_(
-                coalesce.copy().is_(exp.null()),
+                coalesce.is_(exp.null()),
                 type(expression)(this=rightmost_arg.copy(), expression=other.copy()),
             ),
         )
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -469,15 +469,16 @@ def simplify_coalesce(expression):
     else:
         return expression
 
-    # 'COALESCE(x, 1) = 2' ->
-    #     '(
-    #         NOT COALESCE(x) IS NULL
-    #         AND COALESCE(x) = 2
-    #      )
-    #      OR (
-    #          COALESCE(x) IS NULL
-    #          AND 1 = 2
-    #      )
+    # COALESCE(x, 1) = 2
+    # becomes
+    # (
+    #     NOT COALESCE(x) IS NULL
+    #     AND COALESCE(x) = 2
+    # )
+    # OR (
+    #     COALESCE(x) IS NULL
+    #     AND 1 = 2
+    # )
     # This may seem more complex, but subsequent simplify steps will further reduce this.
     #
     # This transformation is valid for non-constants,
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -66,6 +66,7 @@ def simplify(expression):
         node.parent = expression.parent
         node = simplify_literals(node, root)
         node = simplify_parens(node)
+        node = simplify_coalesce(node)
         if root:
             expression.replace(node)
         return node
@@ -184,6 +185,7 @@ COMPARISONS = (
     *GT_GTE,
     exp.EQ,
     exp.NEQ,
+    exp.Is,
 )
 
 INVERSE_COMPARISONS = {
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -430,6 +432,74 @@ def simplify_parens(expression):
     return expression
 
 
+CONSTANTS = (
+    exp.Literal,
+    exp.Boolean,
+    exp.Null,
+    exp.Date,
```
tobymao, line 439: how is date a literal?

Date(x) ?

barakalon: oh whoops

sqlglot/optimizer/simplify.py
```diff
@@ -430,6 +432,74 @@ def simplify_parens(expression):
...
+    else:
+        return expression
+
+    # COALESCE(x, y, z)
+    #                ^ grab that guy
+    coalesce_args = coalesce.expressions
+    if coalesce_args:
+        rightmost_arg = coalesce_args[-1]
+    else:
+        return expression
```
GeorgeSittas, line 470: Isn't it guaranteed that `coalesce.expressions` will always contain ≥1 items here?

barakalon: Because of the first line in the function?

I'm cheating and using one function to both simplify COALESCE and comparisons containing a COALESCE. So this is a different expression.

GeorgeSittas: I'm not sure I follow, isn't `coalesce` set to the `Coalesce` expression in the `if-elif-else` branch right above?

barakalon: `COALESCE(x) = 1`

barakalon: I guess it depends on whether we transform in post-order or pre-order

GeorgeSittas: Aah the piece I missed was that `Coalesce` is one of those expressions where you have both `this` and `expressions`. Makes sense.

sqlglot/optimizer/simplify.py
```diff
@@ -430,6 +432,74 @@ def simplify_parens(expression):
...
+        coalesce = expression.right
+        other = expression.left
+    else:
+        return expression
+
+    # COALESCE(x, y, z)
+    #                ^ grab that guy
+    coalesce_args = coalesce.expressions
+    if coalesce_args:
+        rightmost_arg = coalesce_args[-1]
```
tobymao, line 468: don't you want to iterate and fight the constant? not necessarily the last one,

coalesce(x, 1, 2) 

yes this is bad sql but you wanna ignore the 2 anyway

sqlglot/optimizer/simplify.py
```diff
@@ -430,6 +432,74 @@ def simplify_parens(expression):
...
+    else:
+        return expression
+
+    # COALESCE(x, 1) = 2
+    # becomes
+    # (
+    #     NOT COALESCE(x) IS NULL
```
tobymao, line 475: isn't it easy to pop off the coalesce? and just do 

not x is null

if it's a one liner it may be worth it

sqlglot/optimizer/simplify.py
```diff
@@ -430,6 +432,74 @@ def simplify_parens(expression):
...
+    # This may seem more complex, but subsequent simplify steps will further reduce this.
+    #
+    # This transformation is valid for non-constants,
+    # but I think it really only does anything if they are both constants.
+    if isinstance(rightmost_arg, CONSTANTS) and isinstance(other, CONSTANTS):
+        rightmost_arg.pop()
+
+        return exp.or_(
+            exp.and_(
+                coalesce.copy().is_(exp.null()).not_(),
```
tobymao, line 491: you don't need to call copy, is_ builder already calls copy

sqlglot/optimizer/simplify.py
```diff
@@ -430,6 +432,74 @@ def simplify_parens(expression):
...
+                expression.copy(),
```
tobymao, line 492: same with this one, and_ is copy by default, unless you want to do and_ copy=False

tests/fixtures/optimizer/simplify.sql
```diff
@@ -602,3 +602,33 @@ TRUE;
...
 x <> 2018 OR x = 2018;
+
+--------------------------------------
+-- Coalesce
+--------------------------------------
+COALESCE(x);
+x;
+
+COALESCE(x, 1) = 2;
+x = 2 AND NOT x IS NULL;
```
tobymao, line 613: why do we need the null check?

sqlglot/optimizer/simplify.py
barakalon: This already simplifies that correctly, because we don't pop off the COALESCE.

```sql
COALESCE(x, 1, 2) = 2
```
Becomes
```sql
(
  NOT COALESCE(x, 1) IS NULL
  AND COALESCE(x, 1) = 2
)
OR (
  COALESCE(x, 1) IS NULL
  AND 2 = 2
)
```
Becomes
```sql
(
  NOT COALESCE(x, 1) IS NULL
  AND COALESCE(x, 1) = 2
)
OR (
  COALESCE(x, 1) IS NULL
)
```
Becomes
```sql
(
  NOT (
    (
      NOT COALESCE(x) IS NULL
      AND COALESCE(x) IS NULL
    ) OR (
      COALESCE(x) IS NULL
      AND 1 IS NULL
    )
  )
  AND (
    (
      NOT COALESCE(x) IS NULL
      AND COALESCE(x) = 2
    ) OR (
      COALESCE(x) IS NULL
      AND 1 = 2
    )
  )
)
OR (
  (
    NOT COALESCE(x) IS NULL
    AND COALESCE(x) IS NULL
  ) OR (
    COALESCE(x) IS NULL
    AND 1 IS NULL
  )
)
```
Becomes
```sql
NOT x IS NULL AND x = 2
```

Very roundabout haha
But kinda elegant...

tests/fixtures/optimizer/simplify.sql
GeorgeSittas: I think the `null` will bubble up without this check and you may get different results if you only keep `x = 2`:

```
georgesittas=# with cte(x) as (select 1 union all select null) select coalesce(x, 1) = 2 from cte;
 ?column?
----------
 f
 f
(2 rows)

georgesittas=# with cte(x) as (select 1 union all select null) select x = 2 from cte;
 ?column?
----------
 f

(2 rows)

georgesittas=# with cte(x) as (select 1 union all select null) select x = 2 and not x is null from cte;
 ?column?
----------
 f
 f
(2 rows)
```

sqlglot/optimizer/simplify.py
barakalon: Ah, you're right.
The worst bugs I've dealt with in sqlglot is when I forget to copy and the AST gets cycles. So I trigger happy on the copies now.

tests/fixtures/optimizer/simplify.sql
barakalon:
```
COALESCE(NULL, 1) = 2 == FALSE
NULL = 2 == NULL
NULL = 2 AND NOT NULL IS NULL = FALSE
```

sqlglot/optimizer/simplify.py
barakalon: Sure, I can do add that as a little optimzation

sqlglot/optimizer/simplify.py
barakalon: Its not so bad to find the first constant arg. So I'll just do that.

PR #2035, opened 2023-08-11
Feat(presto,oracle): add support for INTERVAL span types

### LINKED ISSUE #2027: parsing trino `interval day to second` and `interval year to month` into a type doesn't work
```
In [9]: import sqlglot as sg

In [10]: sg.__version__
Out[10]: '17.9.1'

In [11]: sg.parse_one("interval day to second", into=sg.exp.DataType, read="trino")
...
ParseError: Failed to parse 'interval day to second' into <class 'sqlglot.expressions.DataType'>

In [12]: sg.parse_one("interval year to month", into=sg.exp.DataType, read="trino")
...
ParseError: Failed to parse 'interval year to month' into <class 'sqlglot.expressions.DataType'>
```

Not related to a problem

Ideally these can be parsed into the `INTERVAL` type

Not sure what the alternatives would be

**Additional context**

N/A

GeorgeSittas:
Fixes #2027

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3859,6 +3859,18 @@ class TimeUnit(Expression):
         super().__init__(**args)
 
 
+# https://www.oracletutorial.com/oracle-basics/oracle-interval/
+# https://trino.io/docs/current/language/types.html#interval-year-to-month
+class IntervalYearToMonthSpan(Expression):
+    arg_types = {}
+
+
+# https://www.oracletutorial.com/oracle-basics/oracle-interval/
+# https://trino.io/docs/current/language/types.html#interval-day-to-second
+class IntervalDayToSecondSpan(Expression):
+    arg_types = {}
+
+
 class Interval(TimeUnit):
     arg_types = {"this": False, "unit": False}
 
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -3859,6 +3859,18 @@ def __init__(self, **args):
...
+# https://trino.io/docs/current/language/types.html#interval-year-to-month
+class IntervalYearToMonthSpan(Expression):
+    arg_types = {}
+
+
+# https://www.oracletutorial.com/oracle-basics/oracle-interval/
+# https://trino.io/docs/current/language/types.html#interval-day-to-second
+class IntervalDayToSecondSpan(Expression):
+    arg_types = {}
+
```
GeorgeSittas, line 3872: FYI the approach of creating expressions here will be useful if / when we decide to support Oracle's full syntax of these interval types, since they can also be parameterized with additional values:

```
INTERVAL '120-3' YEAR(3) TO MONTH
INTERVAL '11 10:09:08.555' DAY TO SECOND(3)
```

sqlglot/parser.py
```diff
@@ -3184,10 +3184,17 @@ def _parse_types(
             elif self._match_text_seq("WITHOUT", "TIME", "ZONE"):
                 maybe_func = False
         elif type_token == TokenType.INTERVAL:
-            unit = self._parse_var()
+            span: t.Optional[t.List[exp.Expression]] = None
```
tobymao, line 3187: switch this to an else and put the type on the first one

PR #2036, opened 2023-08-11
Feat(clickhouse): add support for FixedString(N) type

### LINKED ISSUE #2031: Support for parsing ClickHouse's `FixedString(N)` type

Not related to a problem.

I'd like sqlglot to support parsing the ClickHouse [`FixedString(N)`](https://clickhouse.com/docs/en/sql-reference/data-types/fixedstring) type.

I haven't considered any alternatives

**Additional context**

N/A

GeorgeSittas:
Fixes #2031


sqlglot/dialects/clickhouse.py
```diff
@@ -38,6 +38,7 @@ class Tokenizer(tokens.Tokenizer):
             "DATETIME64": TokenType.DATETIME64,
             "DICTIONARY": TokenType.DICTIONARY,
             "FINAL": TokenType.FINAL,
+            "FIXEDSTRING": TokenType.FIXEDSTRING,
```
tobymao, line 41: did you add this to the TYPE_TOKEN in parser.py?

sqlglot/dialects/clickhouse.py
```diff
@@ -66,6 +67,11 @@ class Parser(parser.Parser):
             "XOR": lambda args: exp.Xor(expressions=args),
         }
 
+        TYPE_TOKENS = {
+            *parser.Parser.TYPE_TOKENS,
+            TokenType.FIXEDSTRING,
```
tobymao, line 72: this should be in the base parser, not clickhouse

GeorgeSittas: This is a very clickhouse-specific feature - I deliberately didn't include it in the base parser because it's not really relevant to anyone else. Is there an other reason I missed, besides "extending" the sqlglot dialect?

tobymao: TokenTypes are global, so there's no reason not to put it in base, i've already made it clickhouse specific by puting the string -> token mapping only in clickhouse

GeorgeSittas: Got it, makes sense I guess because we'll also avoid creating a copy of `TYPE_TOKENS` in this way.

Making in the fix in a bit.

GeorgeSittas: Done

PR #2041, opened 2023-08-11
Fix: parse time[(p)] with time zone correctly

### LINKED ISSUE #2028: `time(precision) with time zone` is parsed incorrectly into `TIMESTAMPTZ`

```
In [7]: import sqlglot as sg

In [8]: sg.__version__
Out[8]: '17.9.1'

In [9]: sg.parse_one("time(7) with time zone", into=sg.exp.DataType, read="trino")
Out[9]:
(DATATYPE this: Type.TIMESTAMPTZ, expressions:
  (DATATYPESIZE this:
    (LITERAL this: 7, is_string: False)))
```

**Official Documentation**

The specific combination of `time (P) with time zone` [isn't documented](https://trino.io/docs/current/language/types.html#time-with-time-zone), but using `trino-cli` it's supported:

```
trino:default> create table t (x time (3) with time zone);
CREATE TABLE
trino:default> drop table t; create table t (x time with time zone);
CREATE TABLE
```

sqlglot also incorrectly returns `TIMESTAMPTZ` for `time with time zone`

GeorgeSittas:
Fixes #2028


sqlglot/dialects/duckdb.py
```diff
@@ -89,6 +89,11 @@ def _struct_sql(self: generator.Generator, expression: exp.Struct) -> str:
 def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
     if expression.is_type("array"):
         return f"{self.expressions(expression, flat=True)}[]"
+
+    # Type TIMESTAMP WITH TIME ZONE does not support any modifiers
+    if expression.is_type("timestamptz", "timetz"):
+        return "TIMESTAMPTZ"
+
```
GeorgeSittas, line 96: This changed was based on an error I saw while testing:

```python
>>> import duckdb
>>> duckdb.connect().execute("select cast('2020-01-01 00:00:00' as timestamptz(6))").fetchdf()
duckdb.ParserException: Parser Error: Type TIMESTAMP WITH TIME ZONE does not support any modifiers!
```

Note that DuckDB has neither a `timetz` type, nor a `time with time zone` one, so the `timetz` check here is a temporary solution to transpiling such types from e.g. presto to it. I think in the future we might want to implement a different workaround for this to allow single "time" values with timezone information.

sqlglot/dialects/duckdb.py
```diff
@@ -89,6 +89,11 @@ def _struct_sql(self: generator.Generator, expression: exp.Struct) -> str:
 def _datatype_sql(self: generator.Generator, expression: exp.DataType) -> str:
     if expression.is_type("array"):
         return f"{self.expressions(expression, flat=True)}[]"
+
+    # Type TIMESTAMP WITH TIME ZONE does not support any modifiers
+    if expression.is_type("timestamptz", "timetz"):
+        return "TIMESTAMPTZ"
```
cpcloud, line 95: For give my ignorance, but should this be returning `TIMESTAMPTZ` even though `expression` can be either a timestamptz or timetz?

GeorgeSittas: DuckDB doesn't support `timetz` as a type - see https://github.com/tobymao/sqlglot/pull/2041#discussion_r1291602767 for more context

cpcloud: Are you on an older version?

It works for me:

```
D select cast(now() as time with time zone);
┌────────────────────────────────────┐
│ CAST(now() AS TIME WITH TIME ZONE) │
│        time with time zone         │
├────────────────────────────────────┤
│ 17:14:16.675+00                    │
└────────────────────────────────────┘
```

tobymao (approved):

sqlglot/dialects/duckdb.py
cpcloud: Works in Python too:

```
❯ python -c 'import duckdb; print(duckdb.execute("select cast(now() as time with time zone)").fetchall())'
[(datetime.time(17, 15, 38, 764000),)]
…/ibis on  add-trino-source is 📦 v6.1.0 via 🐍 v3.10.12 via ❄️  impure (shell)
```

cpcloud: `timetz` looks valid too:

```
❯ python -c 'import duckdb; print(duckdb.execute("select cast(now() as timetz)").fetchall())'
[(datetime.time(17, 17, 40, 982000),)]
```

GeorgeSittas: Oh, whoops, you're actually right. I must've assumed that they didn't support this type because it's not mentioned in their [documentation](https://duckdb.org/docs/sql/data_types/overview.html) (only `timestamp with time zone` is), but I should've definitely double-checked.

Will update, thanks for catching this!

cpcloud:
Thanks for knocking out all these PRs folks.

It's really going to help us as start our port to replace sqlalchemy with sqlglot + DBAPIs 🚀

sqlglot/dialects/duckdb.py
GeorgeSittas: Alrighty, fixed!

PR #2056, opened 2023-08-14
Fix(parser): ensure identifiers aren't treated as NO_PAREN_FUNCTIONS

GeorgeSittas:
There was a regression due to https://github.com/tobymao/sqlglot/commit/0b4c6c859339d012f8302ff7fcb21195587ac428:

```python
>>> import sqlglot
>>> sqlglot.parse_one('SELECT "case"')
sqlglot.errors.ParseError: Expected END after CASE. Line 1, Col: 13.
  SELECT "case"
          ~~~~
```

This is because we use the text of `"CASE"` to do a dict lookup in `NO_PAREN_FUNCTION_PARSERS` and so instead of parsing an identifier we hit the above error in `_parse_case`.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -728,6 +728,11 @@ class Parser(metaclass=_Parser):
         "NEXT": lambda self: self._parse_next_value_for(),
     }
 
+    INVALID_FUNC_NAME_TOKENS = {
+        TokenType.IDENTIFIER,
+        TokenType.STRING,
+    }
+
     FUNCTIONS_WITH_ALIASED_ARGS = {"STRUCT"}
 
     FUNCTION_PARSERS = {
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlglot/parser.py
```diff
@@ -3356,7 +3356,7 @@ def _parse_function(
         upper = this.upper()
 
         parser = self.NO_PAREN_FUNCTION_PARSERS.get(upper)
-        if optional_parens and parser:
+        if optional_parens and parser and token_type != TokenType.IDENTIFIER:
```
tobymao, line 3359: what if it's a string?

GeorgeSittas: The simple cases work because `_parse_primary` is invoked before `_parse_function` in `_parse_field`:

```python
➜  sqlglot git:(jo/function_parser_fixup) python
Python 3.10.9 (main, Dec 15 2022, 10:44:50) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import sqlglot
>>> sqlglot
>>> sqlglot.parse_one("select 'case'")
(SELECT expressions:
  (LITERAL this: case, is_string: True))
```

But I think you're raising a valid point - I'll come up with a way to make this more robust.

PR #2064, opened 2023-08-15
fix: pushdown predicate to HAVING

barakalon:
This was a surprising bug


sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -81,7 +81,11 @@ def pushdown_cnf(predicates, scope, scope_ref_count):
                 break
             if isinstance(node, exp.Select):
                 predicate.replace(exp.true())
-                node.where(replace_aliases(node, predicate), copy=False)
+                inner_predicate = replace_aliases(node, predicate)
```
tobymao, line 84: what about dnf?

barakalon: whoops

PR #2067, opened 2023-08-16
(doris) fix trunc oracle to doris convert

liujiwen-up:
- There is no trunc function in doris. It is replaced by date_trunc and date. When there is no parameter, the date function is used.
https://doris.apache.org/docs/dev/sql-manual/sql-functions/date-time-functions/date-trunc/
https://doris.apache.org/docs/dev/sql-manual/sql-reference/Data-Types/DATE
- Through the previous trunc analysis to Anonymous, a parsing correspondence was added to the parser, and a sysdate correspondence was added to the tokens.
- And added tests in test_doris.
- At the same time, I have a suggestion to put the DATE_DELTA_INTERVAL dictionary in tsql in the parent class for everyone to use together.


sqlglot/dialects/doris.py
```diff
@@ -9,6 +9,21 @@
     time_format,
 )
 from sqlglot.dialects.mysql import MySQL
+from sqlglot.dialects.tsql import DATE_DELTA_INTERVAL
+
+
+def handle_date_trunc(self, expression: exp.DateTrunc) -> str:
+    expr = self.sql(expression, "this")
+    unit = expression.text("unit").lower()
+    if unit == "none":
```
tobymao, line 18: why would unit be the string "none"?

sqlglot/parser.py
```diff
@@ -92,6 +92,9 @@ class Parser(metaclass=_Parser):
             length=exp.Literal.number(10),
         ),
         "VAR_MAP": parse_var_map,
+        "TRUNC": lambda args: exp.DateTrunc(
```
tobymao, line 95: delete this, and add a mapping in doris, 

"TRUNC": exp.DateTrunc.from_args_list

sqlglot/tokens.py
```diff
@@ -739,6 +739,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "TRUNCATE": TokenType.COMMAND,
         "VACUUM": TokenType.COMMAND,
         "USER-DEFINED": TokenType.USERDEFINED,
+        "SYSDATE": TokenType.CURRENT_TIMESTAMP,
```
tobymao, line 742: delete this, and add it to doris

tests/dialects/test_doris.py
```diff
@@ -19,3 +19,19 @@ def test_regex(self):
...
+
+    def test_trunc(self):
+        self.validate_all(
+            "SELECT TRUNC(sysdate)",
+            write={
+                "doris": "SELECT DATE(NOW())",
+            },
+        )
+
+    def test_trunc(self):
```
tobymao, line 31: can you please review your PRs before you submit them? this is an example of a low effort pr, you have a duplicate test_trunc here, 

delete both test_truncs and create

def test_doris:

and put all of these tests in there

sqlglot/dialects/doris.py
```diff
@@ -9,6 +9,21 @@
...
+        return f"DATE({expr})"
+    else:
+        mapped_unit = (
```
tobymao, line 21: this is not very good code, can you do this properly?

tobymao: i'm not sure what this code is supposed to do, and i don't even think it's tested

tobymao:
you need to make your commits adhere to the standard

Fix(doris): ...

tobymao:
@liujiwen-up if your future PRs are not of higher quality, we won't accept anymore. 

please put more effort into them. if you're unsure of what that means join our slack and we can explain.

sqlglot/tokens.py
liujiwen-up: I want to convert from other dialects to doris, only add the mapping in doris, it will not work for other dialects

sqlglot/parser.py
liujiwen-up: I want to convert from other dialects to doris, only add the mapping in doris, it will not work for other dialects

sqlglot/dialects/doris.py
liujiwen-up: When turnc has only one parameter, unit is none

sqlglot/tokens.py
tobymao: SYSDATE is not valid in every dialect, only certain dialects. if you want to map from CURRENT_TIMESTAMP to SYSDATE in doris, this is the the right approach.

sqlglot/dialects/doris.py
tobymao: that is not true, unit can never be none. this code doesn't seem to be correct or you are probably misunderstanding how it's supposed to work

sqlglot/parser.py
tobymao: not every dialect supports trunc

sqlglot/tokens.py
liujiwen-up: OK, the oracle dialect supports it, then I will write it in oracle

sqlglot/parser.py
liujiwen-up: OK, the oracle dialect supports it, then I will write it in oracle

sqlglot/dialects/doris.py
liujiwen-up: image

sqlglot/dialects/doris.py
liujiwen-up: sorry, here is my problem

tests/dialects/test_doris.py
liujiwen-up: sorry, here is my problem

sqlglot/dialects/doris.py
tobymao: it's because the code you wrote in parser.py line 95 is wrong

liujiwen-up: how should i get in touch with you

tobymao: tobikodata.com/slack

tobymao:
closing this pr for now

PR #2077, opened 2023-08-16
Fix!(optimizer): remove redundant parens during subquery elimination

### LINKED ISSUE #2075: Optimizer is incorrect when superfluous parens wrap subqueries
```
from sqlglot import parse_one
from sqlglot.dialects import Snowflake

sql = """
SELECT 
  ("SUBQUERY_0"."KEY") AS "SUBQUERY_1_COL_0"
FROM 
  (
    SELECT 
      * 
    FROM 
      ((( -- <-------------- ***note multiple parens****
          select 
            * 
          from 
            (
              select 
                event_name as key, 
                insert_ts 
              from 
                (
                  select 
                    insert_ts, 
                    event_name 
                  from 
                    sales
                  where 
                    insert_ts > '2023-08-07 21:03:35.590 -0700'
                )
        )
      ))) AS "SF_CONNECTOR_QUERY_ALIAS"
  ) AS "SUBQUERY_0"

"""

print("========= with parens ==========")
optimized = optimize(parse_one(sql, dialect=Snowflake), dialect=Snowflake)
print(optimized.sql(pretty=True))

print("========= without parens ==========")
without_double_parens = sql.replace("((", "").replace("))","")
optimized = optimize(parse_one(without_double_parens, dialect=Snowflake), dialect=Snowflake)
print(optimized.sql(pretty=True))
```
returns

```
========= with parens ==========
SELECT
  "SF_CONNECTOR_QUERY_ALIAS"."KEY" AS "SUBQUERY_1_COL_0"
FROM (
  (
    "cte" AS "cte"
  )
) AS "SF_CONNECTOR_QUERY_ALIAS"
========= without parens ==========
SELECT
  "SALES"."EVENT_NAME" AS "SUBQUERY_1_COL_0"
FROM "SALES" AS "SALES"
WHERE
  "SALES"."INSERT_TS" > '2023-08-07 21:03:35.590 -0700'
```

I've run into a case where the optimizer is sensitive to the number of [superfluous] parentheses wrapping a subquery, and further returns incorrect logic in the "with parens" case. Note entirely missing `where` clause in the `with parens` case.

The parsed ASTs seem to be the same, so I'm assuming this is in the optimizer somewhere.

GeorgeSittas:
Fixes #2075

### DIFF
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -143,7 +143,7 @@ def _eliminate_derived_table(scope, existing_ctes, taken):
         return None
 
     # Get rid of redundant exp.Subquery expressions, i.e. those that are just used as wrappers
-    to_replace = scope.expression.parent.reversed_unnest()
+    to_replace = scope.expression.parent.unwrap()
     name, cte = _new_cte(scope, existing_ctes, taken)
     table = exp.alias_(exp.table_(name), alias=to_replace.alias or name)
     table.set("joins", to_replace.args.get("joins"))
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -143,7 +143,7 @@ def _eliminate_derived_table(scope, existing_ctes, taken):
         return None
 
     # Get rid of redundant exp.Subquery expressions, i.e. those that are just used as wrappers
-    to_replace = scope.expression.parent.unnest(reverse=True)
+    to_replace = scope.expression.parent.reversed_unnest()
     name, cte = _new_cte(scope, existing_ctes, taken)
     table = exp.alias_(exp.table_(name), alias=to_replace.alias or name)
     table.set("joins", to_replace.args.get("joins"))
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -142,13 +142,21 @@ def _eliminate_derived_table(scope, existing_ctes, taken):
     if scope.parent.pivots or isinstance(scope.parent.expression, exp.Lateral):
         return None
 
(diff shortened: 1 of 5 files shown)


sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -142,13 +142,21 @@ def _eliminate_derived_table(scope, existing_ctes, taken):
...
 
-    parent = scope.expression.parent
+    # Get rid of redundant exp.Subquery expressions, i.e. those that are used as parentheses
+    to_replace = scope.expression.parent
+    while (
+        isinstance(to_replace, exp.Subquery)
+        and to_replace.same_parent
+        and not any(v is not None and k != "this" for k, v in to_replace.args.items())
+    ):
+        to_replace = to_replace.parent
```
GeorgeSittas, line 152: The idea here is that we don't want remove subqueries that have additional args, e.g. laterals, alias, joins, ...

Perhaps we can also aggregate these "special" args into a set, so that we can avoid checking _all_ query modifier args in `exp.Subquery` and just focus on them?

tobymao: can this just be implement in def unnest?

GeorgeSittas: Where do you mean?

tobymao: there's a method unnest that is supposed to do this behavior

GeorgeSittas: I'll take a look, perhaps we can do this "unnesting" at an earlier stage

GeorgeSittas: @tobymao I took a stab at this, can you take another look? Happy to pivot if this is not what you had in mind.

smomen:
much appreciated @GeorgeSittas :)

PR #2081, opened 2023-08-17
Fix!(clickhouse): prefer `String` datatype instead of `TEXT` when generating SQL

kszucs:
The native string type for clickhouse is `String` which has the following aliases: 
- `LONGTEXT`
- `MEDIUMTEXT` 
- `TINYTEXT` 
- `TEXT` 
- `LONGBLOB` 
- `MEDIUMBLOB`
- `TINYBLOB` 
- `BLOB`
- `VARCHAR`
- `CHAR`

Relevant documentation section: https://clickhouse.com/docs/en/sql-reference/data-types/string


GeorgeSittas (approved):
LGTM - should we add a test to demonstrate the conversion of `VARCHAR(255)` -> `String`?

PR #2086, opened 2023-08-17
Fix!: Have Spark put CTE at front of insert

eakmanrq:
This syntax is what I have used in the past (with Spark 2.4) and confirmed that this works with Spark 3.2 and latest Databricks release. Newer releases of Databricks (and likely Spark) do support having the CTE after the INSERT but this format is more universally supported.

### DIFF
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -114,6 +114,13 @@ def _unqualify_pivot_columns(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def _insert_sql(self: Hive.Generator, expression: exp.Insert) -> str:
+    expression = expression.copy()
+    with_ = expression.expression.args.pop("with", None)
+    expression.set("with", with_)
+    return self.insert_sql(expression)
+
+
 class Spark2(Hive):
     class Parser(Hive.Parser):
         FUNCTIONS = {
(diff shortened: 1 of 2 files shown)


GeorgeSittas (commented):
I'd first check whether there's a CTE in `Insert`'s expression before copying the whole thing, but otherwise LGTM 👍 

This should probably also be marked as breaking.

sqlglot/dialects/spark2.py
```diff
@@ -114,6 +114,15 @@ def _unqualify_pivot_columns(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def _insert_sql(self: Hive.Generator, expression: exp.Insert) -> str:
+    if not expression.expression.args.get("with"):
+        return self.insert_sql(expression)
+    expression = expression.copy()
+    with_ = expression.expression.args.pop("with")
+    expression.set("with", with_)
+    return self.insert_sql(expression)
```
GeorgeSittas, line 123: Can clean this up a bit by inverting the condition:

```suggestion
    if expression.expression.args.get("with"):
        expression = expression.copy()
        expression.set("with", expression.expression.args.pop("with"))

    return self.insert_sql(expression)
```

GeorgeSittas (approved):

sqlglot/dialects/spark2.py
eakmanrq: Nice!

GeorgeSittas (approved):

PR #2096, opened 2023-08-18
Feat!: add support for casting to user defined types

### LINKED ISSUE #2091: add support for table cast to parse the tables correctly.
Hello,

Parser is unable to parse the query when there is a table cast function used in the query.

**Code for your reference -**

```

query ="""
SELECT count(1)
INTO V_Temp
FROM TABLE(CAST(somelist as data_list))
WHERE col like '%contact'
"""

expressions = sqlglot.parse_one(query, read='oracle')

for i in x:
print(listTables)
```

Above code fails as it cannot process table casting and throws the error as 'Expected TYPE after CAST'
Request you to add support for table cast in query parsing.

GeorgeSittas:
Fixes #2091

References:
- https://docs.oracle.com/cd/E11882_01/appdev.112/e25519/create_type.htm
- https://www.postgresql.org/docs/current/sql-createtype.html (Redshift doesn't support it)
- https://github.com/prestodb/presto/issues/16324 (Trino doesn't support it)
- https://docs.teradata.com/r/76g1CuvvQlYBjb2WPIuk3g/j8bj5N~MCmNGnYNs3IoA~g
- https://learn.microsoft.com/en-us/sql/t-sql/statements/create-type-transact-sql?view=sql-server-ver16

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -220,6 +220,7 @@ class Hive(Dialect):
     class Parser(parser.Parser):
         LOG_DEFAULTS_TO_LN = True
         STRICT_CAST = False
+        SUPPORTS_USER_DEFINED_TYPES = False
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -89,7 +89,7 @@ def _date_diff_sql(self: generator.Generator, expression: exp.DateDiff) -> str:
 
 def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> str:
     this = expression.this
-    if isinstance(this, exp.Cast) and this.is_type("json", only_kind=True) and this.this.is_string:
+    if isinstance(this, exp.Cast) and this.is_type("json") and this.this.is_string:
         # Since FROM_JSON requires a nested type, we always wrap the json string with
         # an array to ensure that "naked" strings like "'a'" will be handled correctly
         wrapped_json = exp.Literal.string(f"[{this.this.name}]")
@@ -344,8 +344,7 @@ class Hive(Dialect):
             if this and not schema:
                 return this.transform(
                     lambda node: node.replace(exp.DataType.build("text"))
-                    if isinstance(node, exp.DataType)
-                    and node.is_type("char", "varchar", only_kind=True)
+                    if isinstance(node, exp.DataType) and node.is_type("char", "varchar")
                     else node,
                     copy=False,
                 )
@@ -490,7 +489,7 @@ class Hive(Dialect):
                 expression = exp.DataType.build("text")
             elif expression.this in exp.DataType.TEMPORAL_TYPES:
                 expression = exp.DataType.build(expression.this)
-            elif expression.is_type("float", only_kind=True):
+            elif expression.is_type("float"):
                 size_expression = expression.find(exp.DataTypeParam)
                 if size_expression:
                     size = int(size_expression.name)
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -89,7 +89,7 @@ def _date_diff_sql(self: generator.Generator, expression: exp.DateDiff) -> str:
 
 def _json_format_sql(self: generator.Generator, expression: exp.JSONFormat) -> str:
     this = expression.this
-    if isinstance(this, exp.Cast) and this.is_type("json") and this.this.is_string:
+    if isinstance(this, exp.Cast) and this.is_type("json", only_kind=True) and this.this.is_string:
         # Since FROM_JSON requires a nested type, we always wrap the json string with
         # an array to ensure that "naked" strings like "'a'" will be handled correctly
         wrapped_json = exp.Literal.string(f"[{this.this.name}]")
@@ -344,7 +344,8 @@ class Hive(Dialect):
             if this and not schema:
                 return this.transform(
                     lambda node: node.replace(exp.DataType.build("text"))
-                    if isinstance(node, exp.DataType) and node.is_type("char", "varchar")
+                    if isinstance(node, exp.DataType)
+                    and node.is_type("char", "varchar", only_kind=True)
                     else node,
                     copy=False,
                 )
@@ -489,7 +490,7 @@ class Hive(Dialect):
                 expression = exp.DataType.build("text")
             elif expression.this in exp.DataType.TEMPORAL_TYPES:
                 expression = exp.DataType.build(expression.this)
-            elif expression.is_type("float"):
+            elif expression.is_type("float", only_kind=True):
                 size_expression = expression.find(exp.DataTypeParam)
                 if size_expression:
                     size = int(size_expression.name)
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4114,7 +4114,7 @@ class Cast(Func):
     def is_type(self, *dtypes: str | DataType | DataType.Type) -> bool:
         to = self.to
         if isinstance(to, Identifier):
-            return any(to.name == dtype for dtype in dtypes if isinstance(dtype, str))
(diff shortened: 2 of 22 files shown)


sqlglot/expressions.py
```diff
@@ -4115,15 +4115,19 @@ def is_type(self, *dtypes: str | DataType | DataType.Type) -> bool:
...
+
+
 class CastToStrType(Func):
-    arg_types = {"this": True, "expression": True}
+    arg_types = {"this": True, "to": True}
 
 
-class Collate(Binary):
-    pass
+class CastToUserDefinedType(Func):
```
tobymao, line 4126: why did you introduce a new cast? just use the regular one

GeorgeSittas: Because `Cast` also exposes `is_type`, which expects the `to` arg to be something that can be built into a `DataType`, so if a user called it with a UDT they would get unexpected `ParseError`s.

One workaround would be to enhance `Cast.is_type` so that it wraps the `to.is_type(..)` call in a try-except block and falls back to identifier comparison.

tobymao: make a new data type that is a custom one

tobymao: so make it the custom data types first class

sqlglot/dialects/hive.py
```diff
@@ -488,7 +490,7 @@ def datatype_sql(self, expression: exp.DataType) -> str:
                 expression = exp.DataType.build("text")
             elif expression.this in exp.DataType.TEMPORAL_TYPES:
                 expression = exp.DataType.build(expression.this)
-            elif expression.is_type("float"):
+            elif expression.is_type("float", only_kind=True):
```
tobymao, line 493: why are you adding only_kind everywhere? float should basically always equal float right?

array == array

array[int] != array[str]

sqlglot/expressions.py
```diff
@@ -3532,16 +3536,39 @@ class Type(AutoName):
...
                 data_type_exp: t.Optional[Expression] = DataType(this=DataType.Type[upper])
             else:
-                data_type_exp = parse_one(dtype, read=dialect, into=DataType)
+                try:
+                    data_type_exp = parse_one(dtype, read=dialect, into=DataType)
+                except ParseError as ex:
+                    if udt:
+                        return DataType(this=DataType.Type.USERDEFINED, expression=dtype, **kwargs)
+
+                    raise ValueError(f"Unparsable data type value: {dtype}. {ex}")
```
tobymao, line 3571: just do raise i think

tobymao: actually it's not necessary, it happens on line 3574

sqlglot/expressions.py
```diff
@@ -3532,16 +3536,39 @@ class Type(AutoName):
...
 
         if isinstance(dtype, str):
             upper = dtype.upper()
             if upper in DataType.META_TYPES:
                 data_type_exp: t.Optional[Expression] = DataType(this=DataType.Type[upper])
             else:
-                data_type_exp = parse_one(dtype, read=dialect, into=DataType)
+                try:
+                    data_type_exp = parse_one(dtype, read=dialect, into=DataType)
+                except ParseError as ex:
```
tobymao, line 3567: possible to do without the try catch? it doesn't just return null?

sqlglot/expressions.py
```diff
@@ -3554,8 +3581,22 @@ def build(
...
         return DataType(**{**data_type_exp.args, **kwargs})
 
-    def is_type(self, *dtypes: str | DataType | DataType.Type) -> bool:
-        return any(self.this == DataType.build(dtype).this for dtype in dtypes)
+    def is_type(self, *dtypes: str | DataType | DataType.Type, only_kind: bool = False) -> bool:
+        """
+        Checks whether this DataType matches one of the provided data types.
+
+        Args:
+            dtypes: the data types to compare this DataType to.
```
tobymao, line 3589: i feel like by default it should be something like

array == array
array[int] = array
array[int] != array[double]

float = float
float = float(10.0)
float(1.0) != float(10.0)

does that make sense?

sqlglot/expressions.py
GeorgeSittas: I tried without the try-catch first but it doesn't always return `None`, due to this check in `_parse`:

https://github.com/tobymao/sqlglot/blob/6fa1581ae0bdf7e0073a7b9a355e72241e5e6959/sqlglot/parser.py#L967-L968

Tbh, I'm not sure if the `is None` check below is needed - if the parsing fails, it seems like it'll always raise instead of hitting that block.

sqlglot/expressions.py
GeorgeSittas: we can also do `raise` yeah, mostly wanted to follow the same style as in 3574

sqlglot/expressions.py
GeorgeSittas: yeah I was a bit skeptical regarding the type comparisons.. i'm not sure if allowing `type = type(precision)` is correct, e.g. `varchar` is not necessarily the same as `varchar(10)`

sqlglot/dialects/hive.py
GeorgeSittas: > float should basically always equal float right?

my guess to this is basically the same as in https://github.com/tobymao/sqlglot/pull/2096#discussion_r1299437320: the reason i implemented it like this was to ensure we don't assume any kind of equality between types that don't share the same precision

sqlglot/expressions.py
tobymao: if you're asking for a varchar, you don't really care if it's a varchar or a varchar10,

same if you're asking if something is an array, you don't care if it's an array<str> or array<int>

GeorgeSittas: updated, but might need a little more refactoring - let me know what you think

sqlglot/expressions.py
```diff
@@ -3528,23 +3532,54 @@ class Type(AutoName):
...
             else:
-                data_type_exp = parse_one(dtype, read=dialect, into=DataType)
+                try:
+                    data_type_exp = parse_one(dtype, read=dialect, into=DataType)
+                except ParseError as e:
+                    if udt:
+                        return DataType(this=DataType.Type.USERDEFINED, expression=dtype, **kwargs)
 
-            if data_type_exp is None:
-                raise ValueError(f"Unparsable data type value: {dtype}")
```
GeorgeSittas, line 3547: I couldn't trace a path in the code where `data_type_exp` could be `None` here - if parsing fails, I believe we'll always get an exception here.

sqlglot/expressions.py
```diff
@@ -3555,7 +3589,32 @@ def build(
...
+        arrays or structs will be compared using "structural equivalence" semantics, so e.g.
+        array<int> != array<float>.
+
+        Args:
+            dtypes: the data types to compare this DataType to.
+
+        Returns:
+            True, if and only if there is a type in `dtypes` which is equal to this DataType.
+        """
+        if self.expressions and self.this in DataType.NESTED_TYPES:
```
tobymao, line 3603: why the self.this in datatype.nested_types?

doesn't that mean

float(3.0) == float(12.0)

tests/test_expressions.py
```diff
@@ -896,3 +896,26 @@ def test_unnest(self):
...
+
+        ast = parse_one("CAST(x AS ARRAY<INT>)")
+        assert ast.is_type("ARRAY")
+        assert ast.is_type("ARRAY<INT>")
+        assert not ast.is_type("ARRAY<FLOAT>")
+
+        ast = parse_one("CAST(x AS STRUCT<a INT, b FLOAT>)")
+        assert ast.is_type("STRUCT")
+        assert ast.is_type("STRUCT<a INT, b FLOAT>")
+        assert not ast.is_type("STRUCT<a VARCHAR, b INT>")
```
tobymao, line 914: add a test case

varchar(5) != varchar(10)

sqlglot/expressions.py
GeorgeSittas: I see, I misinterpreted the above suggestion - will update

PR #2105, opened 2023-08-23
Feat: Support dialect across all Dataframe Operations

eakmanrq:
Slack thread that prompted this change: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1692679901587439?thread_ts=1692679773.389409&cid=C0448SFS3PF

Prior to this change all input was assumed to be Spark and then you could transpile that final query to whatever dialect you wanted. One challenge with this approach is when the user wants to do something that isn't natively supported by Spark. This example was that the user wanted case-sensitive columns but all column input into the Dataframe was assumed to be Spark and therefore case-insensitive. 

This change makes it so that the SparkSession can be configured for a specific dialect and therefore that dialect would be used for parsing all user input. As shown by the tests, this then support case-sensitive input if the target engine is case-sensitive. 

The old behavior of allowing a dialect to be defined with the `sql` method is still supported for backwards compatibility. A warning is shown if their dialect doesn't match the configured dialect and shows them how to fix this going forward. The plan is to eventually drop `dialect` from `sql` in a future major release.

### DIFF
--- a/sqlglot/dataframe/sql/column.py
+++ b/sqlglot/dataframe/sql/column.py
@@ -5,7 +5,6 @@ import typing as t
 import sqlglot
 from sqlglot import expressions as exp
 from sqlglot.dataframe.sql.types import DataType
-from sqlglot.dialects import Spark
 from sqlglot.helper import flatten, is_iterable
 
 if t.TYPE_CHECKING:
@@ -15,19 +14,20 @@ if t.TYPE_CHECKING:
 
 class Column:
     def __init__(self, expression: t.Optional[t.Union[ColumnOrLiteral, exp.Expression]]):
+        from sqlglot.dataframe.sql.session import SparkSession
+
         if isinstance(expression, Column):
             expression = expression.expression  # type: ignore
         elif expression is None or not isinstance(expression, (str, exp.Expression)):
             expression = self._lit(expression).expression  # type: ignore
-
-        expression = sqlglot.maybe_parse(expression, dialect="spark")
+        elif not isinstance(expression, exp.Column):
+            expression = sqlglot.maybe_parse(expression, dialect=SparkSession().dialect).transform(
+                SparkSession().dialect.normalize_identifier, copy=False
+            )
         if expression is None:
             raise ValueError(f"Could not parse {expression}")
 
-        if isinstance(expression, exp.Column):
-            expression.transform(Spark.normalize_identifier, copy=False)
-
-        self.expression: exp.Expression = expression
+        self.expression: exp.Expression = expression  # type: ignore
 
     def __repr__(self):
         return repr(self.expression)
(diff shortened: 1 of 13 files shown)


sqlglot/dataframe/sql/column.py
```diff
@@ -15,19 +14,22 @@
 
 class Column:
     def __init__(self, expression: t.Optional[t.Union[ColumnOrLiteral, exp.Expression]]):
-        if isinstance(expression, Column):
+        from sqlglot.dataframe.sql.session import SparkSession
+
+        if isinstance(expression, exp.Column):
+            expression = expression
```
tobymao, line 20: this doesn't do anything

sqlglot/dataframe/sql/column.py
```diff
@@ -15,19 +14,22 @@
...
+        elif isinstance(expression, Column):
             expression = expression.expression  # type: ignore
```
tobymao, line 22: type ignore here is probbaly not correct and can be done better

sqlglot/dataframe/sql/column.py
```diff
@@ -15,19 +14,22 @@
...
         elif expression is None or not isinstance(expression, (str, exp.Expression)):
             expression = self._lit(expression).expression  # type: ignore
-
-        expression = sqlglot.maybe_parse(expression, dialect="spark")
+        else:
+            expression = sqlglot.maybe_parse(expression, dialect=SparkSession().dialect).transform(
+                SparkSession().dialect.normalize_identifier
```
tobymao, line 27: copy=False because this is a fresh object

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -291,15 +296,24 @@ def _get_select_expressions(
         select_expressions.append(expression_select_pair)  # type: ignore
         return select_expressions
 
-    def sql(self, dialect="spark", optimize=True, **kwargs) -> t.List[str]:
+    def sql(self, dialect: t.Optional[str] = None, optimize: bool = True, **kwargs) -> t.List[str]:
```
tobymao, line 299: it should be DialectType

sqlglot/dataframe/sql/session.py
```diff
@@ -149,3 +153,38 @@ def _join_hint_names(self) -> t.Set[str]:
...
+
+        def config(
+            self,
+            key: t.Optional[str] = None,
+            value: t.Optional[t.Any] = None,
+            *,
+            map: t.Optional[t.Dict[str, t.Any]] = None,
+            **kwargs: t.Any,
+        ) -> SparkSession.Builder:
+            if key and key == self.SQLFRAME_DIALECT_KEY:
```
tobymao, line 177: you don't need to check if key, just check key == dialect key

sqlglot/dataframe/sql/session.py
```diff
@@ -149,3 +153,38 @@ def _join_hint_names(self) -> t.Set[str]:
...
+                self.dialect = value
+            if map and self.SQLFRAME_DIALECT_KEY in map:
```
tobymao, line 179: elif?

PR #2116, opened 2023-08-27
Feat/tsql Add transpiling for `SET KEY VALUE`

dmoore247:
Add transpiling for `SET KEY VALUE`.
These statements now run on Spark / Databricks.

** Documentation **
[Set Statements](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-statements-transact-sql?view=sql-server-ver16)
[Set Variable](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-local-variable-transact-sql?view=sql-server-ver16)

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -200,6 +200,17 @@ def _parse_date_delta(
     return inner_func
 
 
+def setitem_sql(self, expression: exp.SetItem) -> str:
+    """tsql does not use '=' in SET command but does use '=' with Variables"""
+
+    left = self.sql(expression.this.left)
+    right = self.sql(expression.this.right)
+    if isinstance(expression.this.left, exp.Identifier):
+        return f"{left} {right}"
+    else:
+        return f"{left} = {right}"
+
+
 class TSQL(Dialect):
     RESOLVES_IDENTIFIERS_AS_UPPERCASE = None
     NULL_ORDERING = "nulls_are_small"
@@ -397,6 +408,22 @@ class TSQL(Dialect):
         LOG_DEFAULTS_TO_LN = True
 
         CONCAT_NULL_OUTPUTS_STRING = True
+
+        def _parse_set(self, unset: bool = False, tag: bool = False) -> exp.Set | exp.Command:
+            """
+            tsql uses
+                SET @myvar = <expression>
+            and
+                SET KEY VALUE
+            """
+            left = self._parse_primary() or self._parse_id_var()
+            self._match_texts(("="))
+            right = self._parse_statement() or self._parse_id_var()
+            this = self.expression(exp.EQ, this=left, expression=right)
+
+            return self.expression(
+                exp.Set, expressions=[self.expression(exp.SetItem, this=this)], unset=unset, tag=tag
+            )
 
         def _parse_projections(self) -> t.List[exp.Expression]:
             """
@@ -643,6 +670,8 @@ class TSQL(Dialect):
             exp.Min: min_or_least,
             exp.NumberToStr: _format_sql,
             exp.Select: transforms.preprocess([transforms.eliminate_distinct_on]),
+            # exp.Set: set_sql,
+            exp.SetItem: setitem_sql,
             exp.SHA: lambda self, e: self.func("HASHBYTES", exp.Literal.string("SHA1"), e.this),
             exp.SHA2: lambda self, e: self.func(
                 "HASHBYTES",
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -670,7 +670,6 @@ class TSQL(Dialect):
             exp.Min: min_or_least,
             exp.NumberToStr: _format_sql,
             exp.Select: transforms.preprocess([transforms.eliminate_distinct_on]),
-            # exp.Set: set_sql,
             exp.SetItem: setitem_sql,
             exp.SHA: lambda self, e: self.func("HASHBYTES", exp.Literal.string("SHA1"), e.this),
             exp.SHA2: lambda self, e: self.func(
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -58,6 +58,13 @@ TRANSPILE_SAFE_NUMBER_FMT = {"N", "C"}
 DEFAULT_START_DATE = datetime.date(1900, 1, 1)
 
 
+def setitem_sql(self, expression: exp.SetItem) -> str:
+    left = self.sql(expression.this.left)   # IDENTIFIER, PARAMETER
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -398,6 +409,22 @@ class Parser(parser.Parser):
 
         CONCAT_NULL_OUTPUTS_STRING = True
 
+        def _parse_set(self, unset: bool = False, tag: bool = False) -> exp.Set | exp.Command:
```
tobymao, line 412: this is not correct at all, i think you want to to make set_item_assignment just have optional eq

sqlglot/dialects/tsql.py
```diff
@@ -200,6 +200,17 @@ def inner_func(args: t.List) -> E:
     return inner_func
 
 
+def setitem_sql(self, expression: exp.SetItem) -> str:
```
tobymao, line 203: just put this in the generator

sqlglot/dialects/tsql.py
```diff
@@ -200,6 +200,17 @@ def inner_func(args: t.List) -> E:
...
+    """tsql does not use '=' in SET command but does use '=' with Variables"""
```
tobymao, line 204: what does this comment mean?

sqlglot/dialects/tsql.py
```diff
@@ -200,6 +200,17 @@ def inner_func(args: t.List) -> E:
...
+
+    left = self.sql(expression.this.left)
+    right = self.sql(expression.this.right)
+    if isinstance(expression.this.left, exp.Identifier):
+        return f"{left} {right}"
+    else:
```
tobymao, line 210: else return is unneded

sqlglot/dialects/tsql.py
dmoore247: If I understand you correctly, I tried that early on and again after the comment. I was not able to get it to work by manipulating the logic:
```python
        left = self._parse_primary() or self._parse_id_var()

        if not self._match_texts(("=", "TO")):
             self._retreat(index)
             return None

        right = self._parse_statement() or self._parse_id_var()
        this = self.expression(exp.EQ, this=left, expression=right)

        return self.expression(exp.SetItem, this=this, kind=kind)
    ```

File: sqlglot/dialects/tsql.py (hunk 2 above)
Comment (dmoore247) [reply sqlglot/dialects/tsql.py]: It's a little crowded but got it down to a one liner

File: sqlglot/dialects/tsql.py (hunk 1 above)
Comment (dmoore247) [reply sqlglot/dialects/tsql.py:401]: One desirable outcome, the parsed statement is classed as a SET expression rather than the Command expression

File: sqlglot/dialects/tsql.py (hunk 3 above)
Comment (dmoore247) [reply sqlglot/dialects/tsql.py]: Documenting differences in SET syntax within TSQL. Comment is now gone as code incorporated into Generator.

File: sqlglot/dialects/tsql.py [hunk 5]
```diff
@@ -643,6 +659,7 @@ class Generator(generator.Generator):
             exp.Min: min_or_least,
             exp.NumberToStr: _format_sql,
             exp.Select: transforms.preprocess([transforms.eliminate_distinct_on]),
+            exp.SetItem: lambda self, e: f"{self.sql(e.this.left)}{' ' if isinstance(e.this.left, exp.Identifier) else ' = '}{self.sql(e.this.right)}",
```
GeorgeSittas, line 662: @dmoore247 can you explain the logic behind this condition? why do you check for an identifier?

sqlglot/dialects/tsql.py
GeorgeSittas: I think something along these lines might do the trick:

```diff
+++ b/sqlglot/dialects/tsql.py
@@ -643,6 +643,7 @@ class TSQL(Dialect):
             exp.Min: min_or_least,
             exp.NumberToStr: _format_sql,
             exp.Select: transforms.preprocess([transforms.eliminate_distinct_on]),
+            exp.SetItem: lambda self, e: f"{self.sql(e.this.left)}{' ' if isinstance(e.this.left, exp.Identifier) else ' = '}{self.sql(e.this.right)}",

+++ b/sqlglot/parser.py
@@ -4808,10 +4808,11 @@ class Parser(metaclass=_Parser):

         left = self._parse_primary() or self._parse_id_var()

-        if not self._match_texts(("=", "TO")):
+        if not left:
             self._retreat(index)
             return None

+        self._match_texts(("=", "TO"))
```

sqlglot/dialects/tsql.py
```diff
@@ -398,6 +398,22 @@ class Parser(parser.Parser):
...
 
+        def _parse_set(self, unset: bool = False, tag: bool = False) -> exp.Set | exp.Command:
+            """
+            tsql uses
+                SET @myvar = <expression>
+            and
+                SET KEY VALUE
+            """
+            left = self._parse_primary() or self._parse_id_var()
+            self._match_texts(("="))
```
GeorgeSittas, line 409: FYI this could be just `self._match(TokenType.EQ)`

sqlglot/dialects/tsql.py
dmoore247: Hi @GeorgeSittas thanks very much for your helpful input. Let me take a look.

sqlglot/dialects/tsql.py
dmoore247: `SET @var =` and `SET KEY VALUE` need to be treated differently. I happened to noticed one was parsed as an IDENTIFIER and the other was not. Any advice?

dmoore247:
@GeorgeSittas This is good input, I should be able to wrap this up soon. Thanks.

sqlglot/dialects/tsql.py
dmoore247: Found example where this logic isn't working... `SET @user OFF`
https://github.com/tobymao/sqlglot/blame/edd3804410d43904da12fcc65ffe3bb16774cf0f/tests/fixtures/identity.sql#L187

Back to drawing board.

liujiwen-up (approved):

dmoore247:
Please try **not** to look at the last commit.

I'm confounded with `SET @user OFF` as it is mapping to `SET @user = OFF` instead of `SET user = OFF`.
The parser takes anything with a leading `@` and maps it to a `exp.Parameter`. Hmmm, If `@user` were transformed to an identifier, would that help or is that more messed up?

GeorgeSittas:
Hey @dmoore247, thanks for iterating on this!

I see that there are some complications, so I'd say don't worry about it for the time being if it's not easy to solve; I'll close this PR and open a new one in 1-2 days when I'll have more bandwidth to work on this and let you know.

dmoore247:
Thanks @GeorgeSittas , I'll be ooo for a week anyways.

PR #2117, opened 2023-08-27
Suppress generation of 'NULL' constraint during create table

dmoore247:
Hive, Spark, Databricks support the `NOT NULL` constraint but do not support the `NULL` constraint.

This PR suppresses 'NULL' column constraint generation in the Hive dialect and by extension Spark2, Spark(3), Databricks.

** NOT NULL constraint documentation **
https://issues.apache.org/jira/browse/HIVE-16575

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -109,6 +109,10 @@ def _array_sort_sql(self: generator.Generator, expression: exp.ArraySort) -> str
     return f"SORT_ARRAY({self.sql(expression, 'this')})"
 
 
+def notnullcolumnconstraint_sql(self, expression: exp.NotNullColumnConstraint) -> str:
+    return f"{'' if expression.args.get('allow_null') else 'NOT NULL'}"
+
+
 def _property_sql(self: generator.Generator, expression: exp.Property) -> str:
     return f"'{expression.name}'={self.sql(expression, 'value')}"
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/hive.py
```diff
@@ -109,6 +109,10 @@ def _array_sort_sql(self: generator.Generator, expression: exp.ArraySort) -> str
     return f"SORT_ARRAY({self.sql(expression, 'this')})"
 
 
+def notnullcolumnconstraint_sql(self, expression: exp.NotNullColumnConstraint) -> str:
```
tobymao, line 112: move this directly into the generator, remove the lambda that pass in

sqlglot/dialects/hive.py
```diff
@@ -109,6 +109,10 @@ def _array_sort_sql(self: generator.Generator, expression: exp.ArraySort) -> str
...
+    return f"{'' if expression.args.get('allow_null') else 'NOT NULL'}"
```
tobymao, line 113: get rid of the f string it’s not necessary

tests/dialects/test_tsql.py
```diff
@@ -5,6 +5,15 @@
 class TestTSQL(Validator):
     dialect = "tsql"
 
+    def test_create_table(self):
+        self.validate_identity("CREATE TABLE x (A INTEGER NOT NULL, B INTEGER NULL)")
```
tobymao, line 9: remove the identity test and just put it in the validate all test

delete this function and add it to the tsql tests

sqlglot/dialects/hive.py
dmoore247: The f string prevents line wrap
image

tobymao: doesn't matter, please remove it

sqlglot/dialects/hive.py
dmoore247: Done

sqlglot/dialects/hive.py
dmoore247: Done

dmoore247:
:) Thanks!

PR #2120, opened 2023-08-28
fix(presto): transpile DATE_SUB

barakalon:
https://prestodb.io/docs/current/functions/datetime.html#interval-functions

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -297,7 +297,7 @@ class Presto(Dialect):
             exp.DateSub: lambda self, e: self.func(
                 "DATE_ADD",
                 exp.Literal.string(e.text("unit") or "day"),
-                exp.Mul(this=e.expression, expression=exp.Literal.number(-1)),
+                e.expression * -1,
                 e.this,
             ),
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/presto.py
```diff
@@ -294,6 +294,12 @@ class Generator(generator.Generator):
             ),
             exp.DateStrToDate: lambda self, e: f"CAST(DATE_PARSE({self.sql(e, 'this')}, {Presto.DATE_FORMAT}) AS DATE)",
             exp.DateToDi: lambda self, e: f"CAST(DATE_FORMAT({self.sql(e, 'this')}, {Presto.DATEINT_FORMAT}) AS INT)",
+            exp.DateSub: lambda self, e: self.func(
+                "DATE_ADD",
+                exp.Literal.string(e.text("unit") or "day"),
+                exp.Mul(this=e.expression, expression=exp.Literal.number(-1)),
```
tobymao, line 300: this causes a mutation on e.expression.parent let's do instead

e.expression * -1, which forces a copy and is cleaner

tobymao (approved):

PR #2134, opened 2023-08-31
Fix(parser): support order in update statement

brosoul:
support order in UPDATE statement
refs: https://dev.mysql.com/doc/refman/8.0/en/update.html

PR #2135, opened 2023-08-31
Feat: add support for ORDER BY clause in UPDATE statement

GeorgeSittas:
Reference: https://dev.mysql.com/doc/refman/8.0/en/update.html


GeorgeSittas:
Gonna go ahead and close this in favor of #2134.

PR #2136, opened 2023-08-31
Fix(bigquery): parse JSON_OBJECT properly for key-value pairs

middagj:
This PR fixes the different JSON_OBJECT function in BigQuery. See https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_object

I only fixed it for Signature 1 as it is relatively easy. For Signature 2 I think there needs be some changes to incorporate this syntax in the parser.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -385,25 +385,6 @@ class BigQuery(Dialect):
 
             return table
 
-        def _parse_json_object(self) -> exp.JSONObject:
-            expressions: t.List[exp.Expression] = []
-            while True:
-                key = self._parse_field()
-                if key is None:
-                    if not expressions:
-                        break
-                    self.raise_error("Expecting key argument")
-                if not self._match(TokenType.COMMA):
-                    self.raise_error("Expecting a matching value argument")
-                value = self._parse_expression()
-                if value is None:
-                    self.raise_error("Expecting value argument")
-                expressions.append(self.expression(exp.JSONKeyValue, this=key, expression=value))
-                if not self._match(TokenType.COMMA):
-                    break
-
-            return self.expression(exp.JSONObject, expressions=expressions)
-
     class Generator(generator.Generator):
         EXPLICIT_UNION = True
         INTERVAL_ALLOWS_PLURAL_FORM = False
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -385,6 +385,25 @@ class BigQuery(Dialect):
 
             return table
 
+        def _parse_json_object(self) -> exp.JSONObject:
+            expressions: t.List[exp.Expression] = []
+            while True:
+                key = self._parse_field()
+                if key is None:
+                    if not expressions:
+                        break
+                    self.raise_error("Expecting key argument")
+                if not self._match(TokenType.COMMA):
+                    self.raise_error("Expecting a matching value argument")
+                value = self._parse_expression()
+                if value is None:
+                    self.raise_error("Expecting value argument")
+                expressions.append(self.expression(exp.JSONKeyValue, this=key, expression=value))
+                if not self._match(TokenType.COMMA):
+                    break
+
+            return self.expression(exp.JSONObject, expressions=expressions)
+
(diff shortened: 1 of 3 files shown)


middagj:
I was doing some investigation and it seems that MySQL has similar syntax: https://dev.mysql.com/doc/refman/8.0/en/json-creation-functions.html#function_json-object

sqlglot/dialects/bigquery.py
```diff
@@ -385,6 +385,25 @@ def _parse_table_parts(self, schema: bool = False) -> exp.Table:
...
+                    self.raise_error("Expecting a matching value argument")
+                value = self._parse_expression()
+                if value is None:
+                    self.raise_error("Expecting value argument")
+                expressions.append(self.expression(exp.JSONKeyValue, this=key, expression=value))
+                if not self._match(TokenType.COMMA):
+                    break
+
+            return self.expression(exp.JSONObject, expressions=expressions)
+
```
GeorgeSittas, line 406: This whole method seems unnecessary, it probably suffices to do:

```diff
+++ b/sqlglot/parser.py
@@ -4139,9 +4139,9 @@ class Parser(metaclass=_Parser):
     def _parse_json_key_value(self) -> t.Optional[exp.JSONKeyValue]:
         self._match_text_seq("KEY")
         key = self._parse_field()
-        self._match(TokenType.COLON)
+        self._match_set((TokenType.COLON, TokenType.COMMA))
         self._match_text_seq("VALUE")
-        value = self._parse_field()
+        value = self._parse_column()
```

GeorgeSittas:
Thanks for the PR @middagj, I made a suggestion to clean it up a bit but looks good otherwise.

Regarding the 2nd signature, I don't think it's too hard to support. Perhaps we need to override `_parse_json_object` for BigQuery and do a post-processing (i.e. let the superclass' variant run first) to see if there's just a single JSON kv pair whose key is an array, in order to manually fix the AST

sqlglot/dialects/bigquery.py
GeorgeSittas: What problem does this cause in this case?

middagj: This parse all arguments with comma separated and apply the `self._parse_json_key_value` method, there won't be a key and a value inside the `_parse_json_key_value`, but only one of them.

GeorgeSittas: I'm not sure if this is a real issue, iirc the variants that `_parse_json_object` is supposed to cover are `JSON_OBJECT(k:v [, ...])` and `JSON_OBJECT(KEY k VALUE v [, ...])`. To this point, I also didn't get any test failures locally after applying the suggested changes.

middagj: Ah yes, you are right. I thought _parse_csv would split and passes the items, but it just splits and calls the next parse method. Your patch then looks indeed much better. Let me use that.

middagj:
> Thanks for the PR @middagj, I made a suggestion to clean it up a bit but looks good otherwise.
> 
> Regarding the 2nd signature, I don't think it's too hard to support. Perhaps we need to override `_parse_json_object` for BigQuery and do a post-processing (i.e. let the superclass' variant run first) to see if there's just a single JSON kv pair whose key is an array, in order to manually fix the AST. Note that if we decide to do this, then in my suggestion above we probably also need to change:
> 
> ```python
>          key = self._parse_field()
> ```
> 
> to:
> 
> ```python
>          key = self._parse_column()
> ```

If you don't mind I leave that to a next PR. This already helps me parsing some queries which now fail. I did apply the second part as at least in BigQuery you can put other things than just a field into the value.

Do I need to squash myself or will you do it in GitHub?

GeorgeSittas:
Thanks for the quick iterations 👍 

> If you don't mind I leave that to a next PR. This already helps me parsing some queries which now fail. I did apply the second part as at least in BigQuery you can put other things than just a field into the value.

Sounds good with me, if you want I can take this on as well; just let me know if you plan to work on it so we don't overlap.

> Do I need to squash myself or will you do it in GitHub?

Nah, we usually do it @ merge time, so don't worry about it.

GeorgeSittas (approved):

PR #2138, opened 2023-08-31
Fix(tsql): remove COLUMN keyword from ALTER TABLE

### LINKED ISSUE #2137: tsql ALTER TABLE doesn't use keyword COLUMN when adding a column
```
>>> from sqlglot import transpile
>>> transpile("ALTER TABLE a ADD COLUMN b INT", write="tsql")
['ALTER TABLE a ADD COLUMN b INTEGER']
```

Expected output:
`'ALTER TABLE a ADD b INTEGER'`

treysp:
Fixes #2137

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -581,6 +581,7 @@ class TSQL(Dialect):
         QUERY_HINTS = False
         RETURNING_END = False
         NVL2_SUPPORTED = False
+        ALTER_TABLE_ADD_COLUMN_KEYWORD = False
 
         TYPE_MAPPING = {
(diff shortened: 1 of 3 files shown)


tobymao:
what about for schema?, you only changed columndef

GeorgeSittas (approved):

GeorgeSittas:
> what about for schema?, you only changed columndef

I think that the schema case corresponds to Spark which has the `ADD COLUMNS` syntax instead of a singular `COLUMN`. The relevant context can be found in the initial commit that introduced `ALTER TABLE ...`: https://github.com/tobymao/sqlglot/commit/f57bdd4ad

PR #2139, opened 2023-08-31
Fix(mysql): generate JSON_OBJECT properly

middagj:
Similar to #2136, MySQL also uses a comma to separate key and value.

https://dev.mysql.com/doc/refman/8.0/en/json-creation-functions.html#function_json-object

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -593,6 +593,9 @@ class MySQL(Dialect):
         def jsonarraycontains_sql(self, expression: exp.JSONArrayContains) -> str:
             return f"{self.sql(expression, 'this')} MEMBER OF({self.sql(expression, 'expression')})"
 
+        def jsonkeyvalue_sql(self, expression: exp.JSONKeyValue) -> str:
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -593,6 +593,9 @@ def xor_sql(self, expression: exp.Xor) -> str:
         def jsonarraycontains_sql(self, expression: exp.JSONArrayContains) -> str:
             return f"{self.sql(expression, 'this')} MEMBER OF({self.sql(expression, 'expression')})"
 
+        def jsonkeyvalue_sql(self, expression: exp.JSONKeyValue) -> str:
+            return f"{self.sql(expression, 'this')}, {self.sql(expression, 'expression')}"
+
```
GeorgeSittas, line 598: I'd move this to dialect.py and then add an entry in the `TRANSFORMS` dict of both BQ and MySQL that map `exp.JSONKeyValue` to it (there are several examples of this pattern in the codebase).

middagj: Sure, let me try that.

GeorgeSittas (approved):

PR #2140, opened 2023-08-31
Fix(tsql): support adding multiple columns with ALTER TABLE

treysp:
Most dialects repeat `ADD COLUMN` for every column being added: `ALTER TABLE a ADD COLUMN b INT, ADD COLUMN c INT`

tsql only puts `ADD` at the beginning of the column list: `ALTER TABLE a ADD b INT, c INT`

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2271,11 +2271,14 @@ class Generator:
         actions = expression.args["actions"]
 
         if isinstance(actions[0], exp.ColumnDef):
-            actions = self.expressions(
-                expression,
-                key="actions",
-                prefix="ADD COLUMN " if self.ALTER_TABLE_ADD_COLUMN_KEYWORD else "ADD ",
-            )
+            if self.ALTER_TABLE_ADD_COLUMN_KEYWORD:
+                actions = self.expressions(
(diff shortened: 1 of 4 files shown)


sqlglot/generator.py
```diff
@@ -2271,11 +2271,14 @@ def altertable_sql(self, expression: exp.AlterTable) -> str:
...
-                prefix="ADD COLUMN " if self.ALTER_TABLE_ADD_COLUMN_KEYWORD else "ADD ",
-            )
+            if self.ALTER_TABLE_ADD_COLUMN_KEYWORD:
+                actions = self.expressions(
+                    expression,
+                    key="actions",
+                    prefix="ADD COLUMN ",
+                )
+            else:
+                actions = f"ADD {self.expressions(expression,key='actions')}"
```
tobymao, line 2281: you're missing a space? did you run make style?

tobymao: why isn't this self.expressions(prefix...) like the previous?

treysp: I ran `pre-commit run` on the staged files before making the commit. It reformatted the tests, didn't do anything here. Fixed now.

PR #2142, opened 2023-09-01
Feat: add support for transpiling some datetime functions from bq to databricks

fireis:
This PR aims to improve compatibility on datetime functions when transpiling from BigQuery to Databricks SQL. 

I've tested the results on both engines and noticed the difference that the datetime results on BQ are in  ISO 8601 (with the T between date and time), while the proposed changes are not. This behavior should be acceptable for my current flows, but if you think there is a better way to do this (or any other part), I'll be happy to adjust!

BigQuery:
image

Databricks:
image

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from sqlglot import exp, transforms
-from sqlglot.dialects.dialect import parse_date_delta
+from sqlglot.dialects.dialect import parse_date_delta, timestamptrunc_sql
 from sqlglot.dialects.spark import Spark
 from sqlglot.dialects.tsql import generate_date_delta_with_unit_sql
 from sqlglot.tokens import TokenType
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/databricks.py
```diff
@@ -28,6 +28,19 @@ class Generator(Spark.Generator):
             **Spark.Generator.TRANSFORMS,
             exp.DateAdd: generate_date_delta_with_unit_sql,
             exp.DateDiff: generate_date_delta_with_unit_sql,
+            exp.DatetimeAdd: lambda self, e: self.func(
+                "TIMESTAMPADD", e.text("unit"), e.text("expression"), e.this
+            ),
+            exp.DatetimeSub: lambda self, e: self.func(
+                "TIMESTAMPADD",
+                e.text("unit"),
+                exp.Mul(this=e.text("expression"), expression=exp.Literal.number(-1)),
```
tobymao, line 37:
```suggestion
                exp.Mul(this=e.expression.copy(), expression=exp.Literal.number(-1)),
```

sqlglot/dialects/databricks.py
```diff
@@ -28,6 +28,19 @@ class Generator(Spark.Generator):
...
+                e.this,
+            ),
+            exp.DatetimeDiff: lambda self, e: self.func(
+                "TIMESTAMPDIFF", e.text("unit"), exp.Literal.string(e.text("expression")), e.this
```
tobymao, line 41:
```suggestion
                "TIMESTAMPDIFF", e.text("unit"), e.expression, e.this
```

tobymao (approved):

sqlglot/dialects/databricks.py
```diff
@@ -28,6 +28,19 @@ class Generator(Spark.Generator):
             **Spark.Generator.TRANSFORMS,
             exp.DateAdd: generate_date_delta_with_unit_sql,
             exp.DateDiff: generate_date_delta_with_unit_sql,
+            exp.DatetimeAdd: lambda self, e: self.func(
+                "TIMESTAMPADD", e.text("unit"), e.text("expression"), e.this
```
tobymao, line 32:
```suggestion
                "TIMESTAMPADD", e.text("unit"), e.expression, e.this
```

PR #2143, opened 2023-09-01
Fix(bigquery): allow non-identifiers in subsequent table parts

middagj:
I came across table names that were prefixed with a number which sqlglot could not parse. I added an index to easily distinguish between the dashes (which can only happen for the first part) and the unquoted non-identifiers which can only happen for subsequent parts. The extra code path for dashes is only needed for unquoted identifiers, so I added that extra check.


sqlglot/dialects/bigquery.py
```diff
@@ -357,17 +357,29 @@ class Parser(parser.Parser):
...
+        def _parse_table_part(
+            self, index: int = 0, schema: bool = False
+        ) -> t.Optional[exp.Expression]:
+            this = super()._parse_table_part(index=index, schema=schema)
 
             # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#table_names
-            if isinstance(this, exp.Identifier):
+            # table name that is an unquoted identifier can additionally include single dashes
+            # only the first identifier in the table path can have dashes
+            if index == 0 and isinstance(this, exp.Identifier) and not this.quoted:
```
tobymao, line 368: i don't know if we need index checks

sqlglot/dialects/bigquery.py
```diff
@@ -357,17 +357,29 @@ class Parser(parser.Parser):
...
 
-                this = exp.Identifier(this=table_name, quoted=this.args.get("quoted"))
+                this = exp.Identifier(this=table_name, quoted=False)
+
+            # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#path_expressions
+            # table identifiers work similar as path expressions where subsequent parts can include non-identifiers
+            if index > 0 and not this and self._advance_any():
+                table_name = self._prev.text
+                if self._match(TokenType.VAR):
+                    table_name += self._prev.text
```
GeorgeSittas, line 381: I'm not sure if this is correct; what if a subsequent part is an identifier of the form `<digit><name>` where `<name>` doesn't correspond to the `VAR` token?

GeorgeSittas:
@middagj is the purpose of this PR to improve the parsing of "path expressions" (as per BQ's terminology)? I'm not sure I fully understand these changes. Can we just activate `IDENTIFIERS_CAN_START_WITH_DIGIT` for BQ if the problem is only related to that, or are there issues with that?

tobymao:
it looks like only table names can start with a number

middagj:
> @middagj is the purpose of this PR to improve the parsing of "path expressions" (as per BQ's terminology)? I'm not sure I fully understand these changes.
> 
> Can we activate `IDENTIFIERS_CAN_START_WITH_DIGIT` for BQ if the problem is only related to that, or are there issues with that? This might not be fully correct too, just trying to think of a way to reduce the complexity here.

Like Toby says, only table names have this, and only when used as subsequent part. Also, it also works for table names that are complete numbers or even with table names that are same as reserved identifier:

```sql
SELECT * FROM foo.bar.25
SELECT * FROM foo.bar.declare
```

sqlglot/dialects/bigquery.py
middagj: I guess we can go without, but it is more correct and actually helps to isolate the logic. The first and second are orthogonal.

sqlglot/dialects/bigquery.py
middagj: Perhaps I don't understand the tokenising and parsing then fully, but I thought that in that case the `<name>` corresponds to the `VAR` token. Can you give an example?

tobymao:
i don't know if i'm a big fan of how this change was done, may tinker with it later today

sqlglot/dialects/bigquery.py
GeorgeSittas: If you have `foo.25values` for example, the second part will be tokenized into a number and a `VALUES` token, so it won't be caught by the `VAR` check here.

middagj: Ah yes, that is the annoying thing when tokenizing in general. So actually we should use the ID_VAR_TOKENS set?

tobymao:
i think i'm going to go with IDENTIFIERS_CAN_START_WITH_DIGIT, i don't think there's an issue except that we're more lenient than bigquery

middagj:
> i think i'm going to go with IDENTIFIERS_CAN_START_WITH_DIGIT, i don't think there's an issue except that we're more lenient than bigquery

But you cannot parse this with only that option:
```sql
SELECT * FROM foo.bar.25
```

tobymao:
nope, i'm working on it

tobymao:
you code didn't work properly with the test case

select * from a.25b c

tobymao:
d4d7a5c5

middagj:
Never considered the space case. Thanks for fixing it!

Issue #848, opened 2022-12-26
Postgres parse error on pseudo-type cast


alexandermalyga:
The following valid SQL raises a parse error:
```
sql = """
SELECT 'abc'::CSTRING
"""

sqlglot.parse_one(sql, read="postgres")
```

This also applies to other Postgres-specific special types like `regclass`

https://www.postgresql.org/docs/15/datatype-pseudo.html

tobymao:
i'm only going to do one of these, can you make a PR for the rest that you need?

alexandermalyga:
Sure no problem, I could do that using your PR as reference. 
Also, sorry for the spam on these Postgres issues, I will try to submit more PRs with them, but the codebase is a little unfamiliar/intimidating.

tobymao:
no problem.

https://github.com/tobymao/sqlglot/commit/58737ea251d9c874ef19f7179d31040f8edf29ff#diff-7fe5bd723cf2d072429f107bc81b4809e43e254b7bc2ef0116698721f0fd3661R230

add the rest to this list of key word mappings and then add tests like i did with CSTRING

PR #2150, opened 2023-09-04
Add postgres Object Identifier Types

Nitrino:
Related to #848

This PR adds support for Postgres Object Identifier Types. 
https://www.postgresql.org/docs/8.1/datatype-oid.html

This fixes the exceptions when parsing schemas like this:
```sql
ALTER TABLE 
  public.active_rule_parameters
ALTER COLUMN
  id
SET 
  DEFAULT nextval('public.active_rule_parameters_id_seq' :: regclass);
```


Nitrino:
CI seems to be failing for reasons beyond my control.

GeorgeSittas (approved):

PR #2151, opened 2023-09-04
Fix(hive): parse <number> <date_part> as an interval instead of an alias

### LINKED ISSUE #2123: (hive) hive parsing error, date addition and subtraction are parsed into aliases
`sqlglot.parse_one("""cast('1998-01-06' as date) + 30 days""", "hive")
TRY_CAST('1998-01-06' AS DATE) + 30 AS days

The syntax tree is as follows:
Parse days into an alias

[(ALIAS this: 
  (ADD this: 
    (TRYCAST this: 
      (LITERAL this: 1998-01-06, is_string: True), to: 
      (DATATYPE this: Type.DATE, nested: False, prefix: False)), expression: 
    (LITERAL this: 30, is_string: False)), alias: 
  (IDENTIFIER this: days, quoted: False))]
`

GeorgeSittas:
Fixes #2123

Tested in https://demo.gethue.com/hue/editor/?type=hiv (enter `demo` for both the username and the password):

Screenshot 2023-09-04 at 7 15 59 PM

The date parts were chosen based on Hive's [docs](https://cwiki.apache.org/confluence/display/hive/languagemanual+types#LanguageManualTypes-Intervals) for the interval type:

image

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -47,7 +47,29 @@ TIME_DIFF_FACTOR = {
     "HOUR": " / 3600",
 }
 
+INTERVAL_VARS = {
+    "SECOND",
+    "SECONDS",
+    "MINUTE",
+    "MINUTES",
+    "DAY",
+    "DAYS",
+    "MONTH",
+    "MONTHS",
+    "YEAR",
+    "YEARS",
+}
+
+
 DIFF_MONTH_SWITCH = ("YEAR", "QUARTER", "MONTH")
+
+
+def _parse_number(self: Hive.Parser, token: TokenType) -> t.Optional[exp.Expression]:
+    number = parser.Parser.PRIMARY_PARSERS[TokenType.NUMBER](self, token)
+    if self._match(TokenType.VAR, advance=False) and self._curr.text.upper() in INTERVAL_VARS:
+        return exp.Interval(this=number, unit=self._parse_var())
+
+    return number
 
 
 def _add_date_sql(self: generator.Generator, expression: exp.DateAdd | exp.DateSub) -> str:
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/hive.py
```diff
@@ -47,9 +47,31 @@
...
+    "SECONDS",
+    "MINUTE",
+    "MINUTES",
+    "DAY",
+    "DAYS",
+    "MONTH",
+    "MONTHS",
+    "YEAR",
+    "YEARS",
+}
```
GeorgeSittas, line 61: Wasn't sure but maybe we should move this in utils?

tobymao: well this will be unique for every dialect cc, @barakalon was taking a look at this

sqlglot/dialects/hive.py
```diff
@@ -47,9 +47,31 @@
...
+
+
 DIFF_MONTH_SWITCH = ("YEAR", "QUARTER", "MONTH")
 
 
+def _parse_number(self: Hive.Parser, token: TokenType) -> t.Optional[exp.Expression]:
+    number = parser.Parser.PRIMARY_PARSERS[TokenType.NUMBER](self, token)
```
tobymao, line 68: why not call super?

GeorgeSittas: because it's not in the context of a class definition and so python won't accept it

sqlglot/dialects/hive.py
GeorgeSittas: Makes sense, will resolve this for now then

sqlglot/dialects/hive.py
tobymao: super(...).x

PR #2156, opened 2023-09-05
Feat!(snowflake): improve parsing for SHOW statement

### LINKED ISSUE #2148: Show doesn't yield any metadata
```

tables = parse_one('SHOW PRIMARY KEYS IN "TEST"."PUBLIC"."customers"', read="snowflake", dialect="snowflake").find_all(exp.Table)
```

Expected output:

List of tables, in this case it should be `customers`

Instead receiving an empty result

Official documentation:
https://docs.snowflake.com/en/sql-reference/sql/show-primary-keys

GeorgeSittas:
Fixes #2148

cc @tobymao had already started working on this; it doesn't parse all `SHOW` variants, but it solves the original issue and falls back to parsing `SHOW ...` as a `Command` in all other cases. It also serves as a foundation for completing the parsing of `SHOW` statements if we ever need to extend it.

Reference: https://docs.snowflake.com/en/sql-reference/sql/show-primary-keys

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -338,12 +338,6 @@ class Snowflake(Dialect):
 
             return self.expression(exp.Show, this=this, scope=scope, scope_kind=scope_kind)
 
-        def _parse_show(self) -> t.Optional[exp.Expression]:
-            parser = self._find_parser(self.SHOW_PARSERS, self.SHOW_TRIE)
-            if parser:
-                return parser(self)
-            return self._parse_as_command(self._prev)
-
     class Tokenizer(tokens.Tokenizer):
         STRING_ESCAPES = ["\\", "'"]
         HEX_STRINGS = [("x'", "'"), ("X'", "'")]
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -190,6 +190,13 @@ def _parse_regexp_replace(args: t.List) -> exp.RegexpReplace:
     return regexp_replace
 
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -190,6 +190,13 @@ def _parse_regexp_replace(args: t.List) -> exp.RegexpReplace:
     return regexp_replace
 
 
+def _show_parser(*args: t.Any, **kwargs: t.Any) -> t.Callable[[Snowflake.Parser], exp.Show]:
+    def _parse(self: Snowflake.Parser) -> exp.Show:
+        return self._parse_show_snowflake(*args, **kwargs)
+
+    return _parse
+
+
```
GeorgeSittas, line 199: This is basically what we do for MySQL; since it's just these two dialects atm, I thought I'd avoid introducing more abstractions to dry this out. Perhaps we can move these duplicated chunks of logic in the base parser if / when we want to add `SHOW` parsing for a third dialect?

sqlglot/dialects/snowflake.py
```diff
@@ -289,6 +296,16 @@ class Parser(parser.Parser):
...
         }
 
+        STATEMENT_PARSERS = {
+            **parser.Parser.STATEMENT_PARSERS,
+            TokenType.SHOW: lambda self: self._parse_show(),
+        }
+
+        SHOW_PARSERS = {
+            "PRIMARY KEYS": _show_parser("PRIMARY KEYS"),
+            "TERSE PRIMARY KEYS": _show_parser("PRIMARY KEYS"),
```
GeorgeSittas, line 306: The `TERSE` doesn't do anything here:

Screenshot 2023-09-05 at 12 46 33 PM

PR #2164, opened 2023-09-06
feat(mysql): add support unsigned integer

Nitrino:
MySQL supports unsigned integer data type.
https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html

This PR adds support for basic syntax like this one:
```sql
CREATE TABLE t (id INT UNSIGNED)
```

**Question:**
MySQL has the ability to set the "size" to integer, this is used with the ZEROFILL attribute.
> When used in conjunction with the optional (nonstandard) ZEROFILL attribute, the default padding of spaces is replaced with zeros. For example, for a column declared as [INT(4) ZEROFILL](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html), a value of 5 is retrieved as 0005.

This adds problems to parsing, as such a string can occur and the parsing fails:
```sql
CREATE TABLE t (id INT(10) UNSIGNED)
```
I haven't found a way to elegantly add this, can you suggest how to implement it or help with it?

### DIFF
--- a/tests/dialects/test_mysql.py
+++ b/tests/dialects/test_mysql.py
@@ -86,7 +86,7 @@ class TestMySQL(Validator):
         self.validate_all(
             "CREATE TABLE t (id INT UNSIGNED)",
             write={
-                "mysql": "CREATE TABLE t (id UNSIGNEDINT)",
+                "mysql": "CREATE TABLE t (id UINT)",
             },
         )
 
--- a/tests/dialects/test_mysql.py
+++ b/tests/dialects/test_mysql.py
@@ -83,6 +83,12 @@ class TestMySQL(Validator):
                 "mysql": "CREATE TABLE IF NOT EXISTS industry_info (a BIGINT(20) NOT NULL AUTO_INCREMENT, b BIGINT(20) NOT NULL, c VARCHAR(1000), PRIMARY KEY (a), UNIQUE d (b), INDEX e (b))",
             },
         )
+        self.validate_all(
+            "CREATE TABLE t (id INT UNSIGNED)",
+            write={
+                "mysql": "CREATE TABLE t (id UNSIGNEDINT)",
+            },
+        )
 
     def test_identity(self):
         self.validate_identity(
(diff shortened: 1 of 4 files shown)


tests/dialects/test_mysql.py
```diff
@@ -83,6 +83,12 @@ def test_ddl(self):
                 "mysql": "CREATE TABLE IF NOT EXISTS industry_info (a BIGINT(20) NOT NULL AUTO_INCREMENT, b BIGINT(20) NOT NULL, c VARCHAR(1000), PRIMARY KEY (a), UNIQUE d (b), INDEX e (b))",
             },
         )
+        self.validate_all(
+            "CREATE TABLE t (id INT UNSIGNED)",
+            write={
+                "mysql": "CREATE TABLE t (id UNSIGNEDINT)",
```
GeorgeSittas, line 89: This is incorrect right?

GeorgeSittas:
The approach chosen in this PR doesn't feel correct; we already have tokens / expressions to represent unsigned integers (namely, `UINT`). Not sure what the nuances were that necessitated adding `UNSIGNEDINT`, but I can take a look at MySQL's syntax soon and see if I help with its implementation.

Nitrino:
@GeorgeSittas I just missed the presence of UINT, I'll replace the new type with an existing one.
But the problem with parameterized types will still remain.

tests/dialects/test_mysql.py
GeorgeSittas: Does MySQL recognize UINT as a valid type?

GeorgeSittas: This test means "assert that `CREATE TABLE t (id INT UNSIGNED)` in MySQL is transpiled to `CREATE TABLE t (id UINT)` in MySQL", so yeah it's incorrect.

GeorgeSittas:
@Nitrino can you open a ticket for this? I'll have a look later today.

GeorgeSittas:
Closing this for now.

PR #2165, opened 2023-09-06
Feat(spark): add support for raw strings

### LINKED ISSUE #2162: Spark raw String Support, comonly used with regexes
This fails with sqlglot:

```python
import sqlglot

sql = """select regexp_replace('100-200', r'([^0-9])', '')"""
sqlglot.parse_one(sql, read="databricks")
```

**Official Documentation**
https://spark.apache.org/docs/latest/sql-ref-literals.html

GeorgeSittas:
Fixes #2162

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -35,6 +35,13 @@ def _parse_datediff(args: t.List) -> exp.Expression:
 
 
 class Spark(Spark2):
+    class Tokenizer(Spark2.Tokenizer):
+        RAW_STRINGS = [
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/spark.py
```diff
@@ -35,6 +35,13 @@ def _parse_datediff(args: t.List) -> exp.Expression:
 
 
 class Spark(Spark2):
+    class Tokenizer(Spark2.Tokenizer):
```
GeorgeSittas, line 38: Tested Spark v2.4.8 and it doesn't support raw strings

Issue #2161, opened 2023-09-06
Teradata to Databricks transpilation of "Sample" keyword


Ashay4u:
When working on Teradata to Databricks conversion I ran into another issue where "Sample" clause in Teradata is not getting transpiled correctly. Details below

**Sample code**

```
sql="""
	SELECT * FROM 
	Employee 
	Sample 0.33, .25, .1;
"""


try:
    new_query = sqlglot.transpile(sql, read='teradata', write='databricks', pretty=True)[0]
    print(new_query)
except sqlglot.errors.ParseError as error:
    print(traceback.format_exc())
    print(error.errors)
```

**Expected output:**

```
SELECT * FROM
Employee
TABLESAMPLE (33 PERCENT)
  ;
```

**Current SqlGlot Output**

```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 4, Col: 12.

        SELECT * FROM
        Employee
        Sample 0.33, .25, .1;


[{'description': 'Invalid expression / Unexpected token', 'line': 4, 'col': 12, 'start_context': '\n\tSELECT * FROM \n\tEmployee \n\tSample ', 'highlight': '0.33', 'end_context': ', .25, .1;\n', 'into_expression': None}]
```

**Official Documentation**
Source: https://www.teradatapoint.com/teradata/teradata-sample-function.htm
Target: https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-sampling.html

GeorgeSittas:
Transpiling `SAMPLE ...` is currently out of scope because it's not trivial, but I created a PR that at least adds support for the Teradata form.

PR #2169, opened 2023-09-06
Feat(teradata): add support for the SAMPLE clause

GeorgeSittas:
Motivated by #2161

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -45,6 +45,7 @@ class Teradata(Dialect):
             "MOD": TokenType.MOD,
             "NE": TokenType.NEQ,
             "NOT=": TokenType.NEQ,
+            "SAMPLE": TokenType.TABLE_SAMPLE,
             "SEL": TokenType.SELECT,
             "ST_GEOMETRY": TokenType.GEOMETRY,
             "TOP": TokenType.TOP,
@@ -55,6 +56,8 @@ class Teradata(Dialect):
         SINGLE_TOKENS.pop("%")
 
     class Parser(parser.Parser):
+        TABLESAMPLE_CSV = True
+
         CHARSET_TRANSLATORS = {
             "GRAPHIC_TO_KANJISJIS",
             "GRAPHIC_TO_LATIN",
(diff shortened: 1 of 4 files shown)


Ashay4u:
Pardon my limited understanding here, but where is tablesample_sql from teradata.py called from? I did not reach that function in my copy of the code

GeorgeSittas:
@Ashay4u it's called from the base generator's `sql` method when calling `transpile` on a query that contains the `SAMPLE` clause. What do you mean you didn't reach that function?

GeorgeSittas:
Can you share your example so I can try to reproduce this? What I've tried so far reaches that code.

Ashay4u:
Code below
```
sql="""
SELECT * FROM 
	Employee 
	Sample 0.33, .25, .1;
"""

try:
    new_query = sqlglot.transpile(sql, read='teradata', write='databricks', pretty=True)[0]
    print(new_query)
except sqlglot.errors.ParseError as error:
    print(traceback.format_exc())
    print(error.errors)
```

My code does not stop at tablesample_sql in generator.py too during transpile. Am i missing a change somewhere

GeorgeSittas:
Yeah, it doesn't reach that method because you're transpiling to _databricks_. See related comment: https://github.com/tobymao/sqlglot/issues/2161#issuecomment-1708826633.

Ashay4u:
Thanks for clarifying. Where should I make changes to the code to try and fix this? I will try to get this to some acceptable fix

Ashay4u:
For my use case, its an acceptable loss. Is there an easy fix for it?

GeorgeSittas:
The fix is not trivial, at least if you're not experienced with SQLGlot's codebase. One idea is to override `_parse_table_sample` in Teradata and:

- Call `super()._parse_table_sample`
- If that returns a `TableSample`, you can check the first arg of `expressions` and depending on what kind of number it is you can also set another arg like `percent`, `rows` etc.

Ashay4u:
Thanks. Will give it a try. How do I familiarize myself with the codebase details. Is there documentation I can refer or just debug through the code to understand? Again, thanks a lot for being so prompt and helpful

GeorgeSittas:
Sure thing, glad you want to contribute. You can also join our [slack](https://tobikodata.com/slack)btw  if you want to chat.

> How do I familiarize myself with the codebase details. Is there documentation I can refer or just debug through the code to understand?

Yeah I'd suggest debugging through the code and looking into past issues & their fixes to understand the work flow. I think the above outline should cover your use case, so try to work around that.

PR #2171, opened 2023-09-06
Fix!(tsql): tsql doesn't support boolean literals or the IS operator

treysp:
This PR addresses these issues:
- tsql [doesn't support boolean literals](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/comparison-operators-transact-sql?view=sql-server-ver16&redirectedfrom=MSDN#boolean-data-type)
  - In the context of assignment, the standard convention is to use the BIT data type and `1`/`0` for `TRUE`/`FALSE`
  - In the context of evaluation, the standard convention is to use `1=1`/`1=0` for `TRUE`/`FALSE`
- tsql only allows the `IS` operator for `IS NULL` and `IS NOT NULL` (does not support `IS [NOT] TRUE`)

The PR adds the following functionality:
- When an `exp.Boolean` is contained in any of the following expression types it is assumed to be a literal and converted to `1`/`0`:
  - exp.EQ (`a = TRUE`), exp.NEQ (`a != FALSE`)
  - exp.Is (`a IS TRUE`)
  - exp.In (`a IN (TRUE)`)
  - exp.Select (`SELECT TRUE`)
  - exp.Alias (`SELECT TRUE as a`)
- In an expression type other than those above, an `exp.Boolean` is converted to `1=1`/`1=0`
- When an `exp.Boolean` is the expression for `exp.Is`, the output uses `=` instead of `IS` (`a IS TRUE` --> `a = 1`)


sqlglot/dialects/tsql.py
```diff
@@ -630,7 +631,18 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
-        LIMIT_FETCH = "FETCH"
+        def boolean_sql(self, expression: exp.Boolean) -> str:
+            bit_types = {exp.EQ, exp.NEQ, exp.Is, exp.In, exp.Select, exp.Alias}
```
GeorgeSittas, line 635: Let's factor this out of T-SQL's class altogether so it's not redefined in each call, i.e. at the top-level of `tsql.py`. Let's also name it `BIT_TYPES` to be consistent with constant naming.

GeorgeSittas (approved):
LGTM 👍

sqlglot/dialects/tsql.py
```diff
@@ -630,7 +633,16 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
-        LIMIT_FETCH = "FETCH"
+        def boolean_sql(self, expression: exp.Boolean) -> str:
+            if type(expression.parent) in BIT_TYPES:
+                return "1" if expression.this else "0"
+
+            return "(1 = 1)" if expression.this else "(1 = 0)"
```
tobymao, line 640: @eakmanrq i think this should fix the issue you were seeing in tsql with "where false" right?

tobymao:
this is a breaking change

PR #2172, opened 2023-09-06
Feat(mysql): improve support for unsigned int types

GeorgeSittas:
Fixes #2166

Reference: https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html

cc: @Nitrino

PR #2173, opened 2023-09-06
Fix!: MySQL Timestamp Data Types

eakmanrq:
Correctly maps `TIMESTAMP` -> `DATETIME` and uses `TIMESTAMP` function when creating the timestamp types.

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -563,9 +563,16 @@ class MySQL(Dialect):
             exp.DataType.Type.UTINYINT: "TINYINT",
         }
 
+        TIMESTAMP_TYPE_MAPPING = {
+            exp.DataType.Type.TIMESTAMP: "DATETIME",
+            exp.DataType.Type.TIMESTAMPTZ: "TIMESTAMP",
+            exp.DataType.Type.TIMESTAMPLTZ: "TIMESTAMP",
+        }
+
         TYPE_MAPPING = {
             **generator.Generator.TYPE_MAPPING,
             **UNSIGNED_TYPE_MAPPING,
+            **TIMESTAMP_TYPE_MAPPING,
         }
 
         TYPE_MAPPING.pop(exp.DataType.Type.MEDIUMTEXT)
@@ -618,6 +625,13 @@ class MySQL(Dialect):
             return f"{self.sql(expression, 'this')} MEMBER OF({self.sql(expression, 'expression')})"
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
+            if expression.to.this in {
+                exp.DataType.Type.TIMESTAMP,
+                exp.DataType.Type.TIMESTAMPTZ,
+                exp.DataType.Type.TIMESTAMPLTZ,
+            }:
+                return super().func("TIMESTAMP", expression.this)
+
             to = self.CAST_MAPPING.get(expression.to.this)
 
             if to:
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -631,7 +631,7 @@ class MySQL(Dialect):
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
             if expression.to.this in self.TIMESTAMP_FUNC_TYPES:
-                return super().func("TIMESTAMP", expression.this)
+                return self.func("TIMESTAMP", expression.this)
 
             to = self.CAST_MAPPING.get(expression.to.this)
 
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -599,6 +599,11 @@ class MySQL(Dialect):
             exp.DataType.Type.VARCHAR: "CHAR",
         }
 
+        TIMESTAMP_FUNC_TYPES = {
+            exp.DataType.Type.TIMESTAMPTZ,
+            exp.DataType.Type.TIMESTAMPLTZ,
+        }
+
         def datatype_sql(self, expression: exp.DataType) -> str:
             # https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html
             result = super().datatype_sql(expression)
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -555,11 +555,22 @@ class Generator(generator.Generator):
...
+                k: v
+                for k, v in generator.Generator.TYPE_MAPPING.items()
+                if k
+                not in {
+                    exp.DataType.Type.MEDIUMTEXT,
+                    exp.DataType.Type.LONGTEXT,
+                    exp.DataType.Type.MEDIUMBLOB,
+                    exp.DataType.Type.LONGBLOB,
+                }
+            },
```
GeorgeSittas, line 569: This is going to run for every key in `TYPE_MAPPING`, right? Not sure why this change was necessary.

sqlglot/dialects/mysql.py
```diff
@@ -555,11 +555,22 @@ class Generator(generator.Generator):
...
+            exp.DataType.Type.TIMESTAMP: "DATETIME",
+            exp.DataType.Type.TIMESTAMPTZ: "TIMESTAMP",
+            exp.DataType.Type.TIMESTAMPLTZ: "TIMESTAMP",
```
GeorgeSittas, line 572: I think we can do something similar to what I did [here](https://github.com/tobymao/sqlglot/pull/2172/files#diff-14ef9d8b99f48dcf8f6a4e3bbab7bf55a9d85d2b83d684c3c822f0f94fcd1d07R558) to factor this types out.

sqlglot/dialects/mysql.py
```diff
@@ -599,6 +610,13 @@ def jsonarraycontains_sql(self, expression: exp.JSONArrayContains) -> str:
             return f"{self.sql(expression, 'this')} MEMBER OF({self.sql(expression, 'expression')})"
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
+            if expression.to.this in {
+                exp.DataType.Type.TIMESTAMP,
+                exp.DataType.Type.TIMESTAMPTZ,
+                exp.DataType.Type.TIMESTAMPLTZ,
```
GeorgeSittas, line 616: See comment in type mapping w.r.t to drying this section up.

tobymao: use the Temporal_types constant

sqlglot/dialects/mysql.py
eakmanrq: Ah I see you just changed that. I didn't see the code change when I copy/pasted this over. Will fix.

sqlglot/dialects/mysql.py
tobymao: just do 
```suggestion
TYPE_MAPPING = TYPE_MAPPING.copy()

for from, to in (
  (MEDIUMTEXT, None),
  (TIMESTAMP, "DATETIME")
 : 
 if to:
   TYPE_MAPPING[from] = to
   els:
   TYPE_MAPPING.pop(from)

```

eakmanrq: Replaced with approach used in previous PR

sqlglot/dialects/mysql.py
eakmanrq: Temporal includes date

GeorgeSittas: That covers all temporal types though right, e.g. `TIME`, `DATETIME`, .. This might be valid, but I'm not 100% sure about MySQL's specifics here; if we only care about `TIMESTAMP[TZ|LTZ]` though, we can just do

```python
if expression.to.this in self.TIMESTAMP_TYPE_MAPPING:
```

tobymao: yea, just use that constant than, otherwise what's the point of it?

eakmanrq: Excluded DATETIME and then also added Doris support which doesn't seem to have a `TIMESTAMP` func so this changed a bit.

sqlglot/dialects/mysql.py
```diff
@@ -563,9 +563,16 @@ class Generator(generator.Generator):
             exp.DataType.Type.UTINYINT: "TINYINT",
         }
 
+        TIMESTAMP_TYPE_MAPPING = {
+            exp.DataType.Type.TIMESTAMP: "DATETIME",
+            exp.DataType.Type.TIMESTAMPTZ: "TIMESTAMP",
+            exp.DataType.Type.TIMESTAMPLTZ: "TIMESTAMP",
```
GeorgeSittas, line 569: If this isn't used anymore then we should probably move these inline in `TYPE_MAPPING`

GeorgeSittas (approved):
A couple last comments but otherwise looks good 👍

sqlglot/dialects/mysql.py
```diff
@@ -618,6 +630,9 @@ def jsonarraycontains_sql(self, expression: exp.JSONArrayContains) -> str:
             return f"{self.sql(expression, 'this')} MEMBER OF({self.sql(expression, 'expression')})"
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
+            if expression.to.this in self.TIMESTAMP_FUNC_TYPES:
+                return super().func("TIMESTAMP", expression.this)
```
GeorgeSittas, line 634: No need to do `super` here, `self.func` also works (nit)

sqlglot/dialects/mysql.py
```diff
@@ -592,6 +599,11 @@ class Generator(generator.Generator):
             exp.DataType.Type.VARCHAR: "CHAR",
         }
 
+        TIMESTAMP_FUNC_TYPES = {
```
tobymao, line 602: move this into DataType

TIMESTAMP_TYPES

you're also missing TIMESTAMP

tobymao: nevermind

sqlglot/dialects/mysql.py
tobymao:
```suggestion
                return self.func("TIMESTAMP", expression.this)
```

PR #2174, opened 2023-09-06
Fix: None identifier on merge with insert

eakmanrq:


GeorgeSittas (approved):
Perhaps an alternative form for the lambda (albeit a bit less explicit) could be:

```python
            lambda identifier: identifier and Postgres.normalize_identifier(identifier).name
```

Don't feel strongly about it, though.

PR #2182, opened 2023-09-08
feat: add TINYTEXT and TINYBLOB types

Nitrino:
MySQL and ClickHouse supports `TINYTEXT` and `TINYBLOB` types. 
This PR adds these types to the library.

**Docs:**
https://dev.mysql.com/doc/refman/8.0/en/blob.html
https://clickhouse.com/docs/en/sql-reference/data-types/string


Nitrino:
@GeorgeSittas I added the tests to ClickHouse and found that additional mapping was still needed in the parser.
But I haven't found mapping for other BLOB and TEXT varieties. I think there's either a bug or I'm misunderstanding something.

GeorgeSittas:
> But I haven't found mapping for other BLOB and TEXT varieties. I think there's either a bug or I'm misunderstanding something.

What kind of mapping are you referring to? Is there something that doesn't work as expected, or you're just confused about how the code functions?

GeorgeSittas:
Ah, I see. Yeah it seems like we should either move the MySQL `LONGBLOB` etc mappings in the base Tokenizer, or add them to ClickHouse too:

```python
"LONGBLOB": TokenType.LONGBLOB,
```

Nitrino:
Yes, I can move them within the current PR or I think it's better to do it separately so as not to mix as it's outside the scope of this PR

GeorgeSittas:
Sure, let's do it in a separate PR. I can take it on if you want, should be a few simple changes.

PR #2183, opened 2023-09-08
Fix(clickhouse): add missing type mappings for string types

GeorgeSittas:
Related discussion: https://github.com/tobymao/sqlglot/pull/2182


sqlglot/dialects/clickhouse.py
```diff
@@ -51,9 +51,16 @@ class Tokenizer(tokens.Tokenizer):
             "INT32": TokenType.INT,
             "INT64": TokenType.BIGINT,
             "INT8": TokenType.TINYINT,
+            "LONGBLOB": TokenType.LONGBLOB,
+            "LONGTEXT": TokenType.LONGTEXT,
```
tobymao, line 55: they map 1-1 so let's move them to base?

GeorgeSittas: sounds good, i'll also change a couple of mappings (tinytext, tinyblob) in base to be consistent

PR #2186, opened 2023-09-11
fix(snowflake): add output_name for semi-structured access

teraamp:
At least in Snowflake accessing a semi-structured field, i.e. SELECT table.column:field gives an output name of "field", rather than _col_0 or something. This adds this behavior.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3961,6 +3961,10 @@ class Between(Predicate):
 class Bracket(Condition):
     arg_types = {"this": True, "expressions": True}
 
+    @property
+    def output_name(self):
+        return self.expressions[0].this
+
 
 class SafeBracket(Bracket):
     """Represents array lookup where OOB index yields NULL instead of causing a failure."""
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -128,7 +128,7 @@ def _mergeable(outer_scope, inner_scope, leave_tables_isolated, from_or_join):
     def _is_a_window_expression_in_unmergable_operation():
         window_expressions = inner_select.find_all(exp.Window)
         window_alias_names = {window.parent.alias_or_name for window in window_expressions}
-        inner_select_name = inner_select.parent.alias_or_name
+        inner_select_name = from_or_join.alias_or_name
         unmergable_window_columns = [
             column
             for column in outer_scope.columns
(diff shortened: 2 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -3961,6 +3961,11 @@ class Between(Predicate):
 class Bracket(Condition):
     arg_types = {"this": True, "expressions": True}
 
+    @property
+    def output_name(self):
+        if self.expressions:
```
GeorgeSittas, line 3966: Why was this check necessary? The `expressions` arg is mandatory for `Bracket`.

GeorgeSittas:
It seems like the formatting checks failed; can you please update this PR after first running `make style`?

sqlglot/expressions.py
teraamp: Fair enough, removed.

teraamp:
> It seems like the formatting checks failed; can you please update this PR after first running `make style`?

Fixed.

sqlglot/expressions.py
```diff
@@ -3961,6 +3961,10 @@ class Between(Predicate):
 class Bracket(Condition):
     arg_types = {"this": True, "expressions": True}
 
+    @property
+    def output_name(self):
+        return self.expressions[0].this
+
```
GeorgeSittas, line 3967: On second thought, I'm not sure if this correct.. What if you have a `Bracket` like `a.b[0]`? Or what if you have multiple expressions under the `Bracket`? Just saw that the return type hint was omitted as well; we require `output_name` to return a `str` but `expressions[0].this` isn't necessarily a string value (if we were to go with this approach, perhaps something like `self.expressions[0].name` would be more appropriate).

@tobymao what do you think?

tobymao: you can check if this is a literal or a var or something?

GeorgeSittas: Yeah, makes sense. An alternative solution could be:

- Check if there's just a single one or more expressions
- If there's only one, output its `name` only if it's a var / string / identifier

GeorgeSittas: If we agree on this approach I can take this PR to the finish line.

tobymao: just use the output name of the argument

GeorgeSittas: https://github.com/tobymao/sqlglot/commit/a81dd14a6de1a50438eae64c2dd20e4841c29572

Issue #2187, opened 2023-09-11
Unable to parse the oracle query when query has JSON components


sashindeitidata:
Hello Toby & Team,
Warm greetings!

**Thank you for fixing all the issues we reported in the past. We really appreciate your efforts! 
We have been using your parser for our work and it has been a great help!**

Could you please help us by providing support for following -

Parser is throwing an error when there are following JSON components are present in the query, like-

1. JSON_OBJECT
2. JSON_ARRAYAGG
3. JSON_ARRAY
4. JSON_TABLE
5. JSON_VALUE

**Oracle Documentation -** 
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_ARRAYAGG.html#GUID-6D56077D-78DE-4CC0-9498-225DDC42E054

**Code for your reference -**

```
import sqlglot.expressions as exp

query1 ="""
SELECT
CASE WHEN DBMS_LOB.getlength(info) < 32000
THEN DBMS_LOB.SUBSTR(info)
END info_txt
, info info_clob
FROM schemaname.tablename ar
INNER JOIN JSON_TABLE(:emps,'$[*]' COLUMNS empno NUMBER PATH '$') jt
ON ar.empno = jt.empno
"""

query2 ="""
SELECT DISTINCT res.party_id,
CASE WHEN JSON_VALUE(info,'$.grp_id') IS NOT NULL THEN 'CF'
          WHEN JSON_VALUE(info,'$.txn_id') IS NOT NULL THEN 'PF'
          WHEN JSON_VALUE(info,'$.agnt_id') IS NOT NULL THEN 'SF'
END run_typ
FROM schemaname.tablename res
CROSS JOIN JSON_TABLE([res.info](http://res.info/), '$.results[*]' COLUMNS
     curr_role VARCHAR2(20) PATH '$.curr_role'
     , proj_hiatus_dt VARCHAR2(10) PATH '$.proj_cooloff'
     , clock_reset VARCHAR2(10) PATH '$.clock_reset'
) jt
WHERE proj_hiatus_dt < TO_CHAR(CURRENT_DATE,'YYYY-MM-DD')
OR clock_reset < TO_CHAR(CURRENT_DATE,'YYYY-MM-DD')
"""

query3 ="""
SELECT JSON_ARRAYAGG(JSON_OBJECT(
        KEY 'RNK' VALUE RNK,
        KEY 'RATING_CODE' VALUE RATING_CODE,
[...]

PR #2190, opened 2023-09-11
Feat!(oracle): refactor parsing of 'FORMAT JSON', add support for JSON_ARRAY

GeorgeSittas:
Fixes item 3 of #2187

References:
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_ARRAY.html
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_OBJECT.html

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -69,6 +69,10 @@ class Oracle(Dialect):
 
         FUNCTION_PARSERS: t.Dict[str, t.Callable] = {
             **parser.Parser.FUNCTION_PARSERS,
+            "JSON_ARRAY": lambda self: self._parse_json_array(
+                exp.JSONArray,
+                expressions=self._parse_csv(lambda: self._parse_format_json(self._parse_bitwise())),
+            ),
             "XMLTABLE": _parse_xml_table,
         }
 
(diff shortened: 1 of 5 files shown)


tests/dialects/test_oracle.py
```diff
@@ -6,6 +6,7 @@ class TestOracle(Validator):
     dialect = "oracle"
 
     def test_oracle(self):
+        self.validate_identity("SELECT JSON_OBJECT(k1: v1 FORMAT JSON, k2: v2 FORMAT JSON)")
```
GeorgeSittas, line 9: This is the use case (same syntax applies to `JSON_ARRAY`) that triggered the need to refactor the parsing of `FORMAT JSON` - see Oracle's docs for reference.

tobymao (approved):

sqlglot/dialects/oracle.py
```diff
@@ -69,6 +69,10 @@ class Parser(parser.Parser):
 
         FUNCTION_PARSERS: t.Dict[str, t.Callable] = {
             **parser.Parser.FUNCTION_PARSERS,
+            "JSON_ARRAY": lambda self: self._parse_json_array(
+                exp.JSONArray,
+                expressions=self._parse_csv(lambda: self._parse_format_json(self._parse_bitwise())),
+            ),
```
GeorgeSittas, line 75: FYI moved this in Oracle because MySQL also has a `JSON_ARRAY` function with different syntax, and so it would be parsed incorrectly if this was in base. Perhaps we can pop the entry off of MySQL's function parsers instead?

PR #2191, opened 2023-09-11
Feat(oracle): add support for JSON_TABLE

### LINKED ISSUE #2187: Unable to parse the oracle query when query has JSON components

GeorgeSittas:
Fixes #2187

Note that the `COLUMNS` clause requires parentheses according to Oracle's spec, so right now the form `COLUMNS ...` without parentheses fails, cc @sashindeitidata are you sure your queries are valid Oracle SQL w.r.t. this? If they are, then this should be an easy fix (i.e. not requiring parens).

The `JSON_TABLE` hasn't been fully implemented, and also `JSON_VALUE` is still parsed into an `Anonymous`. I believe the issue should be addressed with these PRs. We'll be happy to accept contributions related to improving support for these functions. I think adding full support is out of scope for the time being.

Reference: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4525,6 +4525,23 @@ class JSONArrayAgg(Func):
     }
 
 
+# https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html
+# Note: parsing of JSON column definitions is currently incomplete.
+class JSONColumnDef(Expression):
+    arg_types = {"this": True, "kind": False, "path": False}
+
+
+# # https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html
+class JSONTable(Func):
+    arg_types = {
+        "this": True,
+        "expressions": True,
+        "path": False,
+        "error_handling": False,
+        "empty_handling": False,
+    }
+
+
 class OpenJSONColumnDef(Expression):
     arg_types = {"this": True, "kind": True, "path": False, "as_json": False}
 
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -81,6 +81,7 @@ class Oracle(Dialect):
                 this=self._parse_format_json(self._parse_bitwise()),
                 order=self._parse_order(),
             ),
+            "JSON_TABLE": lambda self: self._parse_json_table(),
             "XMLTABLE": _parse_xml_table,
         }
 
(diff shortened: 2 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -4525,6 +4525,23 @@ class JSONArrayAgg(Func):
     }
 
 
+# https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html
+# Note: parsing of JSON column definitions is currently incomplete.
+class JSONColumnDef(Expression):
+    arg_types = {"this": True, "kind": False, "path": False}
```
GeorgeSittas, line 4531: I didn't reuse `ColumnDef` on purpose here, because there are several JSON Column Defn. -specific args that we might need to add in the future, if we decide to improve support for this, so I didn't want to pollute the former by adding these args to it.

sashindeitidata:
> Fixes #2187
> 
> Note that the `COLUMNS` clause requires parentheses according to Oracle's spec, so right now the form `COLUMNS ...` without parentheses fails, cc @sashindeitidata are you sure your queries are valid Oracle SQL w.r.t. this? If they are, then this should be an easy fix (i.e. not requiring parens).
> 
> The `JSON_TABLE` hasn't been fully implemented, and also `JSON_VALUE` is still parsed into an `Anonymous`. I believe the issue should be addressed with these PRs. We'll be happy to accept contributions related to improving support for these functions. I think adding full support is out of scope for the time being.
> 
> Reference: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html

Hi @GeorgeSittas 

Shared queries are valid and works fine when we tested. `COLUMNS ...` without parentheses works fine too.
It would be great if you could add support for the same as well.

Thanks, Sachin

GeorgeSittas:
> > Fixes #2187
> > Note that the `COLUMNS` clause requires parentheses according to Oracle's spec, so right now the form `COLUMNS ...` without parentheses fails, cc @sashindeitidata are you sure your queries are valid Oracle SQL w.r.t. this? If they are, then this should be an easy fix (i.e. not requiring parens).
> > The `JSON_TABLE` hasn't been fully implemented, and also `JSON_VALUE` is still parsed into an `Anonymous`. I believe the issue should be addressed with these PRs. We'll be happy to accept contributions related to improving support for these functions. I think adding full support is out of scope for the time being.
> > Reference: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/JSON_TABLE.html
> 
> Hi @GeorgeSittas
> 
> Shared queries are valid and works fine when we tested. `COLUMNS ...` without parentheses works fine too. It would be great if you could add support for the same as well.
> 
> Thanks, Sachin

Fixed in https://github.com/tobymao/sqlglot/commit/b9f5edee02aed346ebaea767274cc08e3960419b

PR #2194, opened 2023-09-12
Feat: Add redshift concat_ws support

eakmanrq:
Redshift doesn't natively support `CONCAT_WS` so this adds support using double pipe. It also adds an explicit parser for `CONCAT_WS` in order to support the `CONCAT_NULL_OUTPUTS_STRING` flag.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import typing as t
 from enum import Enum
+from functools import reduce
 
 from sqlglot import exp
 from sqlglot._typing import E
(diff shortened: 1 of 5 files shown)


sqlglot/parser.py
```diff
@@ -4083,6 +4084,17 @@ def _parse_concat(self) -> t.Optional[exp.Expression]:
             exp.Concat if self.STRICT_STRING_CONCAT else exp.SafeConcat, expressions=args
         )
 
+    def _parse_concat_ws(self) -> t.Optional[exp.Expression]:
+        delim, *args = self._parse_csv(self._parse_conjunction)
```
tobymao, line 4088: in general the parser is pretty lenient, so in ignore mode, this shouldn't raise an error because there are no args, for example, concat_ws() should not raise a python error here

tobymao: this is pretty much copy and pasted it from concat maybe let's make parse_concat take an argument

eakmanrq: The function is pretty simple so it feels a bit overkill to creating branchinig behavior in `_parse_concat`. I did though put the section where we COALESCE the values into a dedicated method that they both call since that logic was directly repeated. 

In terms of the first comment, this doesn't work for me: `parse_one("CONCAT()")`. Is that expected?

tobymao: you need to set the error_level to ignore

sqlglot/parser.py
```diff
@@ -5140,3 +5147,10 @@ def _replace_lambda(
                     else:
                         column.replace(dot_or_id)
         return node
+
+    def _ensure_string_if_null(self, *args: exp.Expression) -> t.List[exp.Expression]:
```
tobymao, line 5151: this should take in an iterable of expressions so you don't have to splat

sqlglot/dialects/dialect.py
```diff
@@ -656,11 +657,18 @@ def _ts_or_ds_to_date_sql(self: Generator, expression: exp.TsOrDsToDate) -> str:
 
 def concat_to_dpipe_sql(self: Generator, expression: exp.Concat | exp.SafeConcat) -> str:
     expression = expression.copy()
-    this, *rest_args = expression.expressions
-    for arg in rest_args:
-        this = exp.DPipe(this=this, expression=arg)
+    return self.sql(reduce(lambda x, y: exp.DPipe(this=x, expression=y), expression.expressions))
```
GeorgeSittas, line 660: Nice one!

PR #2196, opened 2023-09-12
Refactor: fix mutation bug in Column.to_dot, simplify Dot.build

GeorgeSittas:


sqlglot/expressions.py
```diff
@@ -3795,13 +3796,7 @@ def build(self, expressions: t.Sequence[Expression]) -> Dot:
...
             raise ValueError(f"Dot requires >= 2 expressions.")
 
-        a, b, *expressions = expressions
-        dot = Dot(this=a, expression=b)
-
-        for expression in expressions:
-            dot = Dot(this=dot, expression=expression)
-
-        return dot
+        return t.cast(Dot, reduce(lambda x, y: Dot(this=x, expression=y), expressions))
```
GeorgeSittas, line 3799: @eakmanrq :)

tobymao: i don't like this syntax

PR #2199, opened 2023-09-12
Fix(tokenizer): treat quote as escape only if its followed by itself

### LINKED ISSUE #2198: failed to parse mysql string literals
sqlglot treats both "'" and '\\' as string escapes in mysql dialect, and it failed to parse such sqls:

`select "'abc'";` 

```
import sqlglot
sqlglot.transpile('select "\'abc\'";', read='mysql')
```

```
  File "/Users/robert/workspace/sqlglot/.venv/lib/python3.9/site-packages/sqlglot/__init__.py", line 125, in parse_one
  File "/Users/robert/workspace/sqlglot/.venv/lib/python3.9/site-packages/sqlglot/dialects/dialect.py", line 288, in parse
  File "/Users/robert/workspace/sqlglot/.venv/lib/python3.9/site-packages/sqlglot/dialects/dialect.py", line 302, in tokenize
    return self.tokenizer.tokenize(sql)
  File "/Users/robert/workspace/sqlglot/.venv/lib/python3.9/site-packages/sqlglot/tokens.py", line 827, in tokenize
    raise TokenError(f"Error tokenizing '{context}'") from e
sqlglot.errors.TokenError: Error tokenizing 'select "'abc'"'
```

while in mysql:

```
mysql> select "'abc'";
+-------+
| 'abc' |
+-------+
| 'abc' |
+-------+
1 row in set (0.00 sec)
```

**Official Documentation**
according to official doc, single quote seems much more a special syntax other than general string escape char:

- A ' inside a string quoted with ' may be written as ''.
- A " inside a string quoted with " may be written as "".

https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

GeorgeSittas:
Fixes #2198

### DIFF
--- a/tests/dialects/test_mysql.py
+++ b/tests/dialects/test_mysql.py
@@ -246,7 +246,7 @@ class TestMySQL(Validator):
         self.validate_identity("""'"abc"'""")
         self.validate_identity(
             r"'\'a'",
-            r"'''a'",
+            "'''a'",
         )
         self.validate_identity(
             '''"'abc'"''',
--- a/tests/dialects/test_mysql.py
+++ b/tests/dialects/test_mysql.py
@@ -243,6 +243,15 @@ class TestMySQL(Validator):
         )
 
     def test_escape(self):
+        self.validate_identity(r"""'"abc"'""")
+        self.validate_identity(
+            r"'\'a'",
+            r"'''a'",
+        )
+        self.validate_identity(
+            r'''"'abc'"''',
+            r"'''abc'''",
+        )
         self.validate_all(
             r"'a \' b '' '",
             write={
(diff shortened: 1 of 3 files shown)


tests/dialects/test_mysql.py
```diff
@@ -243,6 +243,15 @@ def test_canonical_functions(self):
         )
 
     def test_escape(self):
+        self.validate_identity(r"""'"abc"'""")
```
tobymao, line 246: you don't need the r

GeorgeSittas: good point thanks

tests/dialects/test_mysql.py
```diff
@@ -243,6 +243,15 @@ def test_canonical_functions(self):
         )
 
     def test_escape(self):
+        self.validate_identity("""'"abc"'""")
+        self.validate_identity(
+            r"'\'a'",
+            r"'''a'",
+        )
+        self.validate_identity(
+            r'''"'abc'"''',
```
tobymao, line 252: same here

PR #2221, opened 2023-09-14
fix(mysql): transpile ISNULL to IS NULL

barakalon:
IS NULL seems more standard than the ISNULL function.

I've wrapped it in parenthesis because I'm scared of operator precedence issues.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -760,3 +760,7 @@ def is_parse_json(expression: exp.Expression) -> bool:
     return isinstance(expression, exp.ParseJSON) or (
         isinstance(expression, exp.Cast) and expression.is_type("json")
     )
+
+
+def isnull_to_is_null(args: t.List) -> exp.Expression:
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -760,3 +760,7 @@ def is_parse_json(expression: exp.Expression) -> bool:
     return isinstance(expression, exp.ParseJSON) or (
         isinstance(expression, exp.Cast) and expression.is_type("json")
     )
+
+
+def isnull_to_is_null(args: t.List) -> exp.Expression:
+    return exp.Paren(this=exp.Paren(this=args[0]).is_(exp.null()))
```
tobymao, line 766: seq_get(args, 0)

tobymao: exp.paren()

GeorgeSittas, line 766: I think the inner parentheses can be omitted? Isn't `IS` non-associative?

barakalon: ah sure

tobymao (approved):

PR #2223, opened 2023-09-14
fix(mysql): TIMESTAMP -> CAST

barakalon:
cc @ginter

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -95,16 +95,6 @@ def _date_add_sql(kind: str) -> t.Callable[[MySQL.Generator, exp.DateAdd | exp.D
     return func
 
 
-def _timestamp_to_cast(args: t.List) -> exp.Expression:
-    if len(args) == 1:
-        return exp.cast(
-            args[0],
-            to=exp.DataType.Type.DATETIME,
-        )
-    # This adds a time expression to a datetime. There's probably a more standard way to transpile this.
-    return exp.func("TIMESTAMP", *args)
-
-
 class MySQL(Dialect):
     # https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -243,6 +243,10 @@ class Parser(parser.Parser):
                 format=exp.Literal.string("%B"),
             ),
             "STR_TO_DATE": _str_to_date,
+            "TIMESTAMP": lambda args: exp.cast(
+                args[0],
```
barakalon, line 247: note: didn't use `seq_get` because it returns an optional type, and cast doesn't accept an optional

tobymao (approved):
why is this needed?

sqlglot/dialects/mysql.py
tobymao: you could avoid the cast then

exp.Cast(this=seq_get(), ...)

barakalon:
trying to transpile `TIMESTAMP(x)` to presto, which doesnt have such a function

sqlglot/dialects/mysql.py
barakalon: I changed it up a bit after reading mysql docs

tobymao:
> trying to transpile `TIMESTAMP(x)` to presto, which doesnt have such a function

wouldn't it make more sense to have presto generate the cast?

tobymao (changes requested):
let's do this on generation

barakalon:
switched to generation. I'm a little scared because TIMESTAMP is also a type

tobymao:
bigquery also has timestamp func

barakalon:
Hm, yeah.
https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#timestamp

Seems like current implementation will still work, although we're overloading the `expression` arg for mysql and bigquery.

tobymao (approved):

PR #2224, opened 2023-09-15
Feat/gen table

dmoore247:
Remove properties generated by TSQL that HIve/Spark/Databricks don't support.
- clustered Primary keys
- clustered storage 
etc.


sqlglot/dialects/hive.py
```diff
@@ -466,6 +477,12 @@ class Generator(generator.Generator):
             exp.NumberToStr: rename_func("FORMAT_NUMBER"),
             exp.LastDateOfMonth: rename_func("LAST_DAY"),
             exp.National: lambda self, e: self.national_sql(e, prefix=""),
+            exp.Constraint: constraint_sql,
+            exp.ClusteredColumnConstraint: lambda self, e: f"({self.expressions(e, 'this', indent=False)})",
```
tobymao, line 481: what are you trying to do here? this looks wrong

sqlglot/dialects/hive.py
```diff
@@ -51,6 +51,17 @@
 DIFF_MONTH_SWITCH = ("YEAR", "QUARTER", "MONTH")
 
 
+def constraint_sql(self, expression: exp.Constraint) -> str:
+    this = self.sql(expression, "this")
+
+    # remove unsupported properties
+    expression.expressions.remove(expression.find(exp.Properties))
```
tobymao, line 58: i don't understand what this is doing, it doesn't look correct

tests/dialects/test_tsql.py
```diff
@@ -18,16 +18,28 @@ def test_tsql(self):
...
-            ) ON [PRIMARY]
+                [zip_cd] [varchar](5) NULL NOT FOR REPLICATION,
+                [zip_cd_mkey] [varchar](5) NOT NULL,
+                CONSTRAINT [pk_mytable] PRIMARY KEY CLUSTERED ([zip_cd_mkey] ASC)
+                WITH (PAD_INDEX = ON, STATISTICS_NORECOMPUTE = OFF) ON [INDEX]
+            ) ON [SECONDARY]
             """,
-            'CREATE TABLE x ("zip_cd" VARCHAR(5) NULL NOT FOR REPLICATION CONSTRAINT "pk_mytable" PRIMARY KEY CLUSTERED ("zip_cd_mkey") WITH (PAD_INDEX=ON, STATISTICS_NORECOMPUTE=OFF) ON "PRIMARY") ON "PRIMARY"',
+            write={
+                #'tsql':'CREATE TABLE x ("zip_cd" VARCHAR(5) NULL NOT FOR REPLICATION CONSTRAINT "pk_mytable" PRIMARY KEY CLUSTERED ("zip_cd_mkey") WITH (PAD_INDEX=ON, STATISTICS_NORECOMPUTE=OFF) ON "PRIMARY") ON "PRIMARY"',
```
tobymao, line 31: uncomment this

sqlglot/dialects/hive.py
dmoore247: Hive / Spark support some constraints but not all constraints other dialects that might emit.

This code intends to filter out the non-supported constraints, so this code does a find and removes them.

tobymao: what constraints are being removed though? just all properties?, i think you can just do

for prop in expression.find_all(exp.Properties):
    prop.pop()

dmoore247:
We're waiting for your response to question above...

Is remove bad? It worked better than iterating over a collection and popping from the same collection which appeared to create a race condition.
Is the concern that properties are removed or how the properties are being removed?

tobymao:
what question?

sqlglot/dialects/hive.py
dmoore247: Is `remove` bad? It worked better than iterating over a collection and popping from the same collection which appeared to create a race condition.
Is the concern that properties are removed or how the properties are being removed?

dmoore247 (commented):
Review complete

sqlglot/dialects/hive.py
tobymao: this is not resolved, you didn't answer the question about why this is being removed.

tobymao: it seems too broad, and it should use pop, and not remove

sqlglot/dialects/hive.py
```diff
@@ -51,6 +51,17 @@
 DIFF_MONTH_SWITCH = ("YEAR", "QUARTER", "MONTH")
 
 
+def constraint_sql(self, expression: exp.Constraint) -> str:
+    this = self.sql(expression, "this")
```
tobymao, line 55: you need to call copy up here, move this function into the generator directly and clean it up

sqlglot/dialects/hive.py
```diff
@@ -466,6 +477,12 @@ class Generator(generator.Generator):
             exp.NumberToStr: rename_func("FORMAT_NUMBER"),
             exp.LastDateOfMonth: rename_func("LAST_DAY"),
             exp.National: lambda self, e: self.national_sql(e, prefix=""),
+            exp.Constraint: constraint_sql,
+            exp.ClusteredColumnConstraint: lambda self, e: f"({self.expressions(e, 'this', indent=False)})",
+            exp.NonClusteredColumnConstraint: lambda self, e: f"({self.expressions(e, 'this', indent=False)})",
+            exp.NotForReplicationColumnConstraint: lambda self, e: "",
```
tobymao, line 483: i think you could just change the properties location intsead of doing this

tobymao: change it to UNSUPPORTED location property

dmoore247: I'm really not understanding this feedback.

Do you mean the "property's location" or the "location property"?
'location' property as in table location as in the storage path, and how does that relate to a clustered constraint?

tobymao: look up Properties.Location and how it's used, switch it to UNSUPPORTED

dmoore247: The pattern appears to be to return "" for unsupported options.

sqlglot/dialects/hive.py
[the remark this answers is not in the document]
dmoore247: Let me research

sqlglot/dialects/hive.py
dmoore247: Copy the expression object? and 
in-line the code with the Generator list for more concise code?

tobymao: yea there’s no need to pass this in

sqlglot/dialects/hive.py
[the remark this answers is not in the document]
dmoore247: Essentially removing the 'CLUSTERED' keyword from the generation.

sqlglot/dialects/hive.py
dmoore247: The `this` is the name of the constraint. Removing this removes the constraint name.

```python
def constraint_sql(self, expression: exp.Constraint) -> str:

    # remove unsupported properties
    expression.expressions.remove(expression.find(exp.Properties))
    expression.expressions.remove(expression.find(exp.OnProperty))

    expressions = self.expressions(expression, sep=" ", flat=True)
    return f"CONSTRAINT {this} {expressions}"
    
```

dmoore247:
Toby, is there anything else you need?

tests/dialects/test_tsql.py
```diff
@@ -18,16 +18,29 @@ def test_tsql(self):
             'CREATE TABLE x (CONSTRAINT "pk_mytable" UNIQUE NONCLUSTERED (a DESC)) ON b (c)'
         )
 
-        self.validate_identity(
+        self.maxDiff = None
```
tobymao, line 21: delete this maxDiff

dmoore247: Done

PR #2226, opened 2023-09-15
Feat: add support for scoped user-defined types

### LINKED ISSUE #2217: add support for table cast to parse the tables correctly when casting is done with user defined table types
Hello Team,
Warm greetings!

Parser is unable to parse query when TABLE(CAST(:VAR AS PAKAGE.UERDEFINEDTABLETYPE)) is used.
TABLE(CAST(...)) retuning wrong table name. But when we used Schema or pakage name o access the userdefined datatype, then sqlglot fails.

**Oracle Documentation -**
https://asktom.oracle.com/pls/apex/f?p=100:11:0::::P11_QUESTION_ID:6600331069731
https://docs.oracle.com/javadb/10.8.3.0/ref/rrefsqlj33562.html

**Code for your reference -**

```

# previously raised issue, but it returns 'TABLE(CAST(somelist ' as output
query1 ="""
SELECT col1, col2
FROM table1, TABLE(CAST(somelist as data_list))
WHERE col like '%contact'
"""

# user defined table datatype can be accessed with pakage.<type>
# below query generate error as it expects ')' and not the dot(.)
query2="""
SELECT col1, col2
FROM table1, TABLE(CAST(somelist as pakage.data_list))
WHERE col like '%contact'
"""
expressions = sqlglot.parse_one(query2, read='oracle')

for i in x:
  print(listTables)
```

Note that above queries are valid oracle queries.

query1 returns incorrect data and query2 fails.
Request you to add support for the same.

Thanks ton in advance!

Thanks,
Sachin

GeorgeSittas:
Fixes #2217

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3279,7 +3279,12 @@ class Parser(metaclass=_Parser):
                 if tokens[0].token_type in self.TYPE_TOKENS:
                     self._prev = tokens[0]
                 elif self.SUPPORTS_USER_DEFINED_TYPES:
-                    return exp.DataType.build(identifier.name, udt=True)
+                    type_name = identifier.name
+
+                    if self._match(TokenType.DOT):
+                        type_name = f"{type_name}.{self._advance_any() and self._prev.text}"
+
+                    return exp.DataType.build(type_name, udt=True)
                 else:
                     return None
             else:
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3281,7 +3281,7 @@ class Parser(metaclass=_Parser):
                 elif self.SUPPORTS_USER_DEFINED_TYPES:
                     type_name = identifier.name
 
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -3279,7 +3279,12 @@ def _parse_types(
                 if tokens[0].token_type in self.TYPE_TOKENS:
                     self._prev = tokens[0]
                 elif self.SUPPORTS_USER_DEFINED_TYPES:
-                    return exp.DataType.build(identifier.name, udt=True)
+                    type_name = identifier.name
+
+                    if self._match(TokenType.DOT):
```
tobymao, line 3284: while self._match(TokenType.DOT) maybe

GeorgeSittas: good point, fixing

tobymao (approved):

PR #2230, opened 2023-09-15
Fix: unescape escape sequences

### LINKED ISSUE #2225: bigquery whitespace characters need to be escaped in string and bytes literals
Looks like newlines are handled but the remaining characters are not escaped correctly.

```
In [22]: import sqlglot as sg

In [23]: sg.__version__
Out[23]: '18.0.0'

In [24]: sql = sg.parse_one("SELECT '\a\b\f\n\r\t\v'", read="bigquery").sql("bigquery")

In [25]: !bq shell <<< "$sql"
Welcome to BigQuery! (Type help for more information.)
ibis-gbq> Error in query string: Error processing job 'ibis-gbq:bqjob_r1fef3289cccddd3_0000018a987c16e2_1': Syntax error: Unclosed string literal at [1:8]
ibis-gbq> Goodbye.
```

**Official Documentation**

https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#escape_sequences

GeorgeSittas:
Fixes #2225

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -12,7 +12,7 @@ from sqlglot.tokens import Tokenizer, TokenType
 
 logger = logging.getLogger("sqlglot")
 
-UNESCAPED_SEQUENCES_TABLE = str.maketrans(
+UNESCAPED_SEQUENCE_TABLE = str.maketrans(
     {
         "\a": "\\a",
         "\b": "\\b",
@@ -1607,7 +1607,7 @@ class Generator:
     def escape_str(self, text: str) -> str:
         text = text.replace(self.QUOTE_END, self._escaped_quote_end)
         if self.UNESCAPE_SEQUENCES:
-            text = text.translate(UNESCAPED_SEQUENCES_TABLE)
+            text = text.translate(UNESCAPED_SEQUENCE_TABLE)
         elif self.pretty:
             text = text.replace("\n", self.SENTINEL_LINE_BREAK)
         return text
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -12,18 +12,6 @@ from sqlglot.tokens import Tokenizer, TokenType
 
 logger = logging.getLogger("sqlglot")
 
-UNESCAPED_SEQUENCE_TABLE = str.maketrans(
-    {
-        "\a": "\\a",
-        "\b": "\\b",
-        "\f": "\\f",
-        "\n": "\\n",
-        "\r": "\\r",
-        "\t": "\\t",
-        "\v": "\\v",
-    }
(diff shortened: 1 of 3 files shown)


sqlglot/generator.py
```diff
@@ -12,6 +12,18 @@
 
 logger = logging.getLogger("sqlglot")
 
+UNESCAPED_SEQUENCE_TABLE = str.maketrans(
```
tobymao, line 15: this should be per dialect i think, everyone is different and configurable

sqlglot/generator.py
```diff
@@ -320,7 +332,7 @@ class Generator:
     STRICT_STRING_CONCAT = False
     NORMALIZE_FUNCTIONS: bool | str = "upper"
     NULL_ORDERING = "nulls_are_small"
-    ESCAPE_LINE_BREAK = False
+    UNESCAPE_SEQUENCES = False
```
tobymao, line 335: could probably remove this in favor of whether or not this mapping exists?

sqlglot/generator.py
```diff
@@ -1594,8 +1606,8 @@ def literal_sql(self, expression: exp.Literal) -> str:
 
     def escape_str(self, text: str) -> str:
         text = text.replace(self.QUOTE_END, self._escaped_quote_end)
-        if self.ESCAPE_LINE_BREAK:
-            text = text.replace("\n", "\\n")
+        if self.UNESCAPE_SEQUENCES:
+            text = text.translate(UNESCAPED_SEQUENCE_TABLE)
         elif self.pretty:
             text = text.replace("\n", self.SENTINEL_LINE_BREAK)
```
tobymao, line 1612: is this section correct?

sqlglot/generator.py
GeorgeSittas: sounds good with me, shouldn't be much of a change

sqlglot/generator.py
GeorgeSittas: sure

sqlglot/generator.py
GeorgeSittas: The `SENTINEL_LINE_BREAK` one?

tobymao: yea when does this hit, just double check it

sqlglot/generator.py
GeorgeSittas: made these changes

sqlglot/generator.py
GeorgeSittas: seems necessary based on https://github.com/tobymao/sqlglot/commit/fd4dbacb3

PR #2231, opened 2023-09-15
Fix(tsql): include catalog and db in create if not exists

treysp:


sqlglot/dialects/tsql.py
```diff
@@ -667,7 +667,19 @@ def create_sql(self, expression: exp.Create) -> str:
                 if kind == "SCHEMA":
                     sql = f"""IF NOT EXISTS (SELECT * FROM information_schema.schemata WHERE schema_name = {identifier}) EXEC('{sql}')"""
                 elif kind == "TABLE":
-                    sql = f"""IF NOT EXISTS (SELECT * FROM information_schema.tables WHERE table_name = {identifier}) EXEC('{sql}')"""
+                    assert table
```
GeorgeSittas, line 670: Shouldn't we also apply this fix to the other cases for `kind`?

sqlglot/dialects/tsql.py
```diff
@@ -667,7 +667,19 @@ def create_sql(self, expression: exp.Create) -> str:
...
+                    where = f"table_name = {self.sql(exp.Literal.string(table.name))}"
```
GeorgeSittas, line 671: I don't think the `exp.Literal.string` adds anything here, you can simply do `f"table_name = {table.name}"`.

sqlglot/dialects/tsql.py
treysp: it's selecting from `sys.indexes` using an object id, and that table doesn't include schema/catalog names

treysp: sorry, that was for index -- i'll add it for schema

treysp: and schema can only come in unqualified, so identifier will always work

GeorgeSittas: i see, makes sense

GeorgeSittas (approved):

sqlglot/dialects/tsql.py
tobymao, line 671: where = exp.column("table_name").eq(table.name)

sqlglot/dialects/tsql.py
```diff
@@ -667,7 +667,19 @@ def create_sql(self, expression: exp.Create) -> str:
...
+                    where = (
```
tobymao, line 672: where = table_name.and_(...)

PR #2233, opened 2023-09-15
fix: don't unnest subqueries when the parent select doesn't have a fr…

ginter:
…om since the subquery can't be joined

I think there's probably still a way to optimize by lifting the subquery but doesn't seem worth it for these types of queries


tobymao (approved):

tobymao:
run make style

PR #2237, opened 2023-09-16
Fix(tsql): generate SELECT INTO from CTAS

treysp:
- Removed temp table tests that use invalid CTAS syntax

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -661,8 +661,13 @@ class TSQL(Dialect):
             exists = expression.args.pop("exists", None)
             sql = super().create_sql(expression)
 
+            table = expression.find(exp.Table)
+
+            if kind == "TABLE" and expression.expression:
+                assert table
+                sql = f"SELECT * INTO {exp.table_name(table)} FROM ({self.sql(expression.expression)}) AS temp"
+
             if exists:
-                table = expression.find(exp.Table)
                 identifier = self.sql(exp.Literal.string(exp.table_name(table) if table else ""))
                 if kind == "SCHEMA":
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -661,8 +661,13 @@ def create_sql(self, expression: exp.Create) -> str:
             exists = expression.args.pop("exists", None)
             sql = super().create_sql(expression)
 
+            table = expression.find(exp.Table)
+
+            if kind == "TABLE" and expression.expression:
+                assert table
+                sql = f"SELECT * INTO {exp.table_name(table)} FROM ({self.sql(expression.expression)}) AS temp"
```
tobymao, line 668: why are you using table_name? what if it needs quotes? why not self.sql(table) ?

GeorgeSittas (approved):
Agree with Toby's comment; looks good besides that

sqlglot/dialects/tsql.py
treysp: When testing, MS-SQL errored if quotes were used for the table name no matter what type of escaping I used. After further digging, it appears that there is an MS-SQL setting to control use of quotes table names that had been set to off in my testing instance (not sure how, it is on by default).

Will fix.

sqlglot/dialects/tsql.py
```diff
@@ -661,8 +661,13 @@ def create_sql(self, expression: exp.Create) -> str:
             exists = expression.args.pop("exists", None)
             sql = super().create_sql(expression)
 
+            table = expression.find(exp.Table)
+
+            if kind == "TABLE" and expression.expression:
+                assert table
```
GeorgeSittas, line 667: Is this `assert` still needed?

PR #2241, opened 2023-09-16
Fix(snowflake): use CASE instead of FILTER(WHERE cond) for conditional aggregation

### LINKED ISSUE #2239: aggregate filter compiles to invalid syntax for some dialects

```
In [18]: import sqlglot as sg

In [19]: sg.parse_one("select avg(x) filter (where true) from t", read="duckdb").sql("snowflake")
Out[19]: 'SELECT AVG(x) FILTER(WHERE TRUE) FROM t'
```

**Official Documentation**

Not sure where is documented in the Snowflake, but it doesn't work when I run queries in the UI

image

I suspect a translation into the dialect's `if`/`iif`/`iff` function would acceptable when a backend doesn't have `filter` syntax.

GeorgeSittas:
Fixes #2239

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -191,6 +191,9 @@ class Generator:
     # UNNEST WITH ORDINALITY (presto) instead of UNNEST WITH OFFSET (bigquery)
     UNNEST_WITH_ORDINALITY = True
 
+    # Whether or not FILTER (WHERE cond) can be used for conditional aggregation
+    AGGREGATE_FILTER_SUPPORTED = True
+
     TYPE_MAPPING = {
         exp.DataType.Type.NCHAR: "CHAR",
         exp.DataType.Type.NVARCHAR: "VARCHAR",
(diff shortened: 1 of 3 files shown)


sqlglot/generator.py
```diff
@@ -959,9 +962,16 @@ def fetch_sql(self, expression: exp.Fetch) -> str:
...
-        return f"{this} FILTER({where})"
+        if self.AGGREGATE_FILTER_SUPPORTED:
+            this = self.sql(expression, "this")
+            where = self.sql(expression, "expression").strip()
+            return f"{this} FILTER({where})"
+
+        agg = expression.this.copy()
+        agg_arg = agg.this
+        cond = expression.expression.this
+        agg_arg.replace(exp.Case().when(cond, agg_arg).else_(exp.null()))
```
tobymao, line 973: let’s do an jf starement

GeorgeSittas: done

sqlglot/generator.py
```diff
@@ -959,9 +962,16 @@ def fetch_sql(self, expression: exp.Fetch) -> str:
...
-        return f"{this} FILTER({where})"
+        if self.AGGREGATE_FILTER_SUPPORTED:
+            this = self.sql(expression, "this")
+            where = self.sql(expression, "expression").strip()
+            return f"{this} FILTER({where})"
+
+        agg = expression.this.copy()
+        agg_arg = agg.this
+        cond = expression.expression.this
+        agg_arg.replace(exp.If(this=cond.copy(), true=agg_arg.copy(), false=exp.null()))
```
tobymao, line 973: the false is unnecessary  it’s implied null

GeorgeSittas: fixed this in https://github.com/tobymao/sqlglot/pull/2243/commits/d5ad990a93d438a5cf748eb754e6e4daaefd326e

PR #2242, opened 2023-09-16
Feat: eliminate semi/anti joins transformation

### LINKED ISSUE #2240: transform semi and anti join to exists and not exists for backends that don't have semi/anti join syntax

Not really related to a problem, except that I can't take a semi join query from a backend that supports them to one that doesn't.

I'd like sqlglot to take this query:

```sql
SELECT
  *
FROM t
LEFT SEMI JOIN s
  ON t.c = s.c
```

and turn it into

```sql
SELECT
  t.*
FROM t
WHERE EXISTS (
  SELECT 1
  FROM s
  WHERE t.c = s.c
)
```

and similarly for `LEFT ANTI JOIN`

In the context of its use (ibis), I'm going to implement this transformation in ibis for now.

**Additional context**

None that I can think of.

GeorgeSittas:
Fixes #2240


sqlglot/dialects/clickhouse.py
```diff
@@ -118,7 +118,7 @@ class Parser(parser.Parser):
             TokenType.ARRAY,
         }
 
-        TABLE_ALIAS_TOKENS = {*parser.Parser.TABLE_ALIAS_TOKENS} - {
+        TABLE_ALIAS_TOKENS = parser.Parser.TABLE_ALIAS_TOKENS.copy() - {
```
tobymao, line 121: doesn’t the - copy?

sqlglot/dialects/duckdb.py
```diff
@@ -191,6 +191,11 @@ class Parser(parser.Parser):
             ),
         }
 
+        TABLE_ALIAS_TOKENS = parser.Parser.TABLE_ALIAS_TOKENS.copy() - {
```
tobymao, line 194: same

sqlglot/dialects/clickhouse.py
GeorgeSittas: nice catch, thanks

PR #2243, opened 2023-09-17
Fix(snowflake): rename GenerateSeries, include offset in unnest_sql

GeorgeSittas:
- Made a few corrections on the `unnest_sql` override, so that Snowflake can [generate the array index](https://docs.snowflake.com/en/sql-reference/functions/flatten#output) when the `offset` arg is set.
- Made sure `IFF` [always has three arguments](https://docs.snowflake.com/en/sql-reference/functions/iff).

The `UNNEST([1, 2, 3])` duckdb -> snowflake case needs a bit more work but I couldn't figure out what was wrong with a quick eyeball, cc @tobymao can you take a look at the Snowflake query below (result of transpilation) and see if I missed something?

Screenshot 2023-09-17 at 3 35 39 AM

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -187,7 +187,8 @@ def explode_to_unnest(index_offset: int = 0) -> t.Callable[[exp.Expression], exp
                 table=[series_alias],
             )
 
-            for select in expression.selects:
+            # we use list here because expression.selects is mutated inside the loop
+            for select in list(expression.selects):
                 to_replace = select
                 pos_alias = ""
                 explode_alias = ""
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -187,7 +187,7 @@ def explode_to_unnest(index_offset: int = 0) -> t.Callable[[exp.Expression], exp
                 table=[series_alias],
             )
 
-            for select in list(expression.selects):
+            for select in expression.selects:
                 to_replace = select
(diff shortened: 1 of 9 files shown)


sqlglot/transforms.py
```diff
@@ -187,7 +187,7 @@ def new_name(names: t.Set[str], name: str) -> str:
                 table=[series_alias],
             )
 
-            for select in list(expression.selects):
+            for select in expression.selects:
```
tobymao, line 190: i added this on purpose due to an infinite loop, is this not needed anymore?

tobymao: expression.expressions is getting mutated  on line 230

GeorgeSittas: Damn thanks, I changed it on a whim and didn't notice the mutation happening. Reverted.

GeorgeSittas:
Ok I think I figured out what the issue is: [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/array_generate_range#syntax)'s `ARRAY_GENERATE_RANGE` has an exclusive end, whereas [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_array), [Postgres](https://www.postgresql.org/docs/current/functions-srf.html) and [Presto](https://prestodb.io/docs/current/functions/array.html) have an inclusive end.

PR #2254, opened 2023-09-18
Fix(dataframe): ensure Column.alias preserves quotes

GeorgeSittas:
The issue here was that 1) `name` was parsed according to Spark, instead of using the session's dialect 2) if quotes were present in `name`, the whole alias name would be treated as an "unsafe identifier" in `to_identifier`, which resulted in adding more quotes and thus generating incorrect SQL.

### DIFF
--- a/sqlglot/dataframe/sql/column.py
+++ b/sqlglot/dataframe/sql/column.py
@@ -214,9 +214,12 @@ class Column:
     def alias(self, name: str) -> Column:
         from sqlglot.dataframe.sql.session import SparkSession
 
-        alias: exp.Expression = sqlglot.maybe_parse(name, dialect=SparkSession().dialect)
+        dialect = SparkSession().dialect
+        alias: exp.Expression = sqlglot.maybe_parse(name, dialect=dialect)
         new_expression = exp.alias_(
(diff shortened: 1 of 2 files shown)


tobymao (approved):

sqlglot/dataframe/sql/column.py
```diff
@@ -212,7 +212,12 @@ def sql(self, **kwargs) -> str:
         return self.expression.sql(**{"dialect": SparkSession().dialect, **kwargs})
 
     def alias(self, name: str) -> Column:
-        new_expression = exp.alias_(self.column_expression, name)
+        from sqlglot.dataframe.sql.session import SparkSession
+
+        alias: exp.Expression = sqlglot.maybe_parse(name, dialect=SparkSession().dialect)
+        new_expression = exp.alias_(
+            self.column_expression, alias.this if isinstance(alias, exp.Column) else name
```
GeorgeSittas, line 219: I check for `exp.Column` here because there were some tests where `name` was `avg(...)`, generated in https://github.com/tobymao/sqlglot/blob/main/sqlglot/dataframe/sql/group.py#L24

PR #2256, opened 2023-09-18
Fix!: preserve ascending order of sorting when present

### LINKED ISSUE #2253: sort order is not rendered for duckdb

DuckDB has a setting `default_order`, which can be set to `'DESC'`, and results in unspecified sort order in `ORDER BY` queries being ordered according to `default_order`.

sqlglot seems to assume (reasonably so, I think) that `ORDER BY x ASC` is equivalent to `ORDER BY x`, which is an invalid assumption for duckdb given the presence of `default_order`:

```
In [4]: import sqlglot as sg

In [5]: sg.__version__
Out[5]: '18.5.1'

In [6]: sg.parse_one("select i from range(5) _ (i) order by i asc", read="duckdb").sql("duckdb")
Out[6]: 'SELECT i FROM RANGE(5) AS _(i) ORDER BY i'

In [7]: !duckdb
D set default_order = 'DESC';
D select i from range(5) _ (i) order by i;
┌───────┐
│   i   │
│ int64 │
├───────┤
│     4 │
│     3 │
│     2 │
│     1 │
│     0 │
└───────┘
```

**Official Documentation**

- https://duckdb.org/docs/sql/query_syntax/orderby.html

GeorgeSittas:
Fixes #2253

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -46,8 +46,6 @@ class Generator:
             The default is on the smaller end because the length only represents a segment and not the true
             line length.
             Default: 80
-        default_order: The default order to use in the ORDER BY clause if ASC / DESC are omitted.
-            Default: ""
         comments: Whether or not to preserve comments in the output SQL code.
             Default: True
     """
@@ -362,7 +360,6 @@ class Generator:
         "max_unsupported",
         "leading_comma",
         "max_text_width",
-        "default_order",
         "comments",
         "unsupported_messages",
         "_escaped_quote_end",
@@ -382,7 +379,6 @@ class Generator:
         max_unsupported: int = 3,
         leading_comma: bool = False,
         max_text_width: int = 80,
-        default_order: str = "",
         comments: bool = True,
     ):
         import sqlglot
@@ -396,7 +392,6 @@ class Generator:
         self.max_unsupported = max_unsupported
         self.leading_comma = leading_comma
         self.max_text_width = max_text_width
-        self.default_order = f" {default_order}" if default_order else ""
         self.comments = comments
 
         # This is both a Dialect property and a Generator argument, so we prioritize the latter
@@ -1670,7 +1665,7 @@ class Generator:
         nulls_are_small = self.NULL_ORDERING == "nulls_are_small"
         nulls_are_last = self.NULL_ORDERING == "nulls_are_last"
 
-        sort_order = " DESC" if desc else self.default_order
+        sort_order = " DESC" if desc else (" ASC" if desc is False else "")
         nulls_sort_change = ""
         if nulls_first and (
             (asc and nulls_are_large) or (desc and nulls_are_small) or nulls_are_last
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -46,6 +46,8 @@ class Generator:
             The default is on the smaller end because the length only represents a segment and not the true
             line length.
             Default: 80
+        default_order: The default order to use in the ORDER BY clause if ASC / DESC are omitted.
+            Default: ""
         comments: Whether or not to preserve comments in the output SQL code.
             Default: True
     """
@@ -360,6 +362,7 @@ class Generator:
         "max_unsupported",
         "leading_comma",
         "max_text_width",
+        "default_order",
         "comments",
         "unsupported_messages",
         "_escaped_quote_end",
@@ -379,6 +382,7 @@ class Generator:
         max_unsupported: int = 3,
         leading_comma: bool = False,
         max_text_width: int = 80,
+        default_order: str = "",
         comments: bool = True,
     ):
         import sqlglot
@@ -392,6 +396,7 @@ class Generator:
         self.max_unsupported = max_unsupported
         self.leading_comma = leading_comma
         self.max_text_width = max_text_width
+        self.default_order = f" {default_order}" if default_order else ""
         self.comments = comments
 
         # This is both a Dialect property and a Generator argument, so we prioritize the latter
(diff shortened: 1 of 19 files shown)


sqlglot/generator.py
```diff
@@ -1665,7 +1670,7 @@ def ordered_sql(self, expression: exp.Ordered) -> str:
         nulls_are_small = self.NULL_ORDERING == "nulls_are_small"
         nulls_are_last = self.NULL_ORDERING == "nulls_are_last"
 
-        sort_order = " DESC" if desc else ""
+        sort_order = " DESC" if desc else f"{self.default_order}"
```
tobymao, line 1673: self.default_order, don't need the quotes

sqlglot/generator.py
```diff
@@ -46,6 +46,8 @@ class Generator:
             The default is on the smaller end because the length only represents a segment and not the true
             line length.
             Default: 80
+        default_order: The default order to use in the ORDER BY clause if ASC / DESC are omitted.
```
tobymao, line 49: i don't think this should be passed in a string, just a boolean

GeorgeSittas: sounds good with me, will update it so it expects an optional boolean and `None` will stand for `""`.

GeorgeSittas: yea I was thinking something like `sort_asc` or something

cpcloud: Actually, I don't think I under what `True` or `False` would mean here.

Does `True` mean "always render the order" and `False` mean "render only if `DESC`?

GeorgeSittas: If the arg was named `default_order_is_asc`, then `True` means `ASC` is printed in the "default" `ORDER BY`, `False` means `DESC` is printed and `None` means it's left as is, i.e `ORDER BY a`.

I'm actually taking a closer look again to see if I can just preserve `ASC` / `DESC` as you suggested in the issue. I tried to do this initially but we have a lot of test failures and I need to verify it's a safe change to make.

GeorgeSittas: https://github.com/tobymao/sqlglot/pull/2256/commits/198099d80a671684680bbd89039a0c5bfefe998b

sqlglot/parser.py
```diff
@@ -2944,20 +2944,20 @@ def _parse_sort(self, exp_class: t.Type[E], token: TokenType) -> t.Optional[E]:
 
     def _parse_ordered(self) -> exp.Ordered:
         this = self._parse_conjunction()
-        self._match(TokenType.ASC)
 
-        is_desc = self._match(TokenType.DESC)
+        asc = self._match(TokenType.ASC)
+        desc = self._match(TokenType.DESC) or (asc and False)
```
tobymao, line 2949: what and False?

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
```diff
@@ -4529,7 +4529,7 @@ SELECT
   "item"."i_category" AS "i_category",
   "item"."i_class" AS "i_class",
   GROUPING("item"."i_category") + GROUPING("item"."i_class") AS "lochierarchy",
-  RANK() OVER (PARTITION BY GROUPING("item"."i_category") + GROUPING("item"."i_class"), CASE WHEN GROUPING("item"."i_class") = 0 THEN "item"."i_category" END ORDER BY SUM("store_sales"."ss_net_profit") / SUM("store_sales"."ss_ext_sales_price")) AS "rank_within_parent"
+  RANK() OVER (PARTITION BY GROUPING("item"."i_category") + GROUPING("item"."i_class"), CASE WHEN GROUPING("item"."i_class") = 0 THEN "item"."i_category" END ORDER BY SUM("store_sales"."ss_net_profit") / SUM("store_sales"."ss_ext_sales_price") ASC) AS "rank_within_parent"
```
tobymao, line 4532: maybe the the optimizer canonicalize should get rid of asc

sqlglot/parser.py
GeorgeSittas: It could be rewritten as `None if asc is None else False`

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
GeorgeSittas: Why should we get rid of it?

sqlglot/parser.py
tobymao: ok, because you want it to be none? that's fine

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
tobymao: because order by x == order by x asc

GeorgeSittas: should we do this in a separate PR?

sqlglot/parser.py
GeorgeSittas: yes that was the intent

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
tobymao: no, just do it here, it affects these tests

GeorgeSittas: done

cpcloud:
Just to verify, the ASC Will be removed in an optimization pass?

Won't that reintroduce the bug if I send in a query with ASC, optimize it, and then send that optimized SQL to duckdb with a different default sort order?

cpcloud:
Is there an issue with keeping ASC around? Are there databases for which that is a meaningful optimization?

tobymao:
in general order by x == order by x asc, duckdb setting is an anomoly. we canonicalize to make sql equal. if you want to make it an option to turn that off you can, or remove canonicalization

cpcloud:
Got it!

Wouldn't SQL generated with ASC still be canonical? Are there dialects where ASC isn't supported?

Happy to discuss, just not sure I'm following the purpose of removing the ASC. Is the idea to keep the generated code smaller?

cpcloud:
If it's just a weird DuckDB setting then I guess I can live with it. Hopefully people aren't altering that setting too much 😬😬

PR #2261, opened 2023-09-19
Fix(hive): get rid of any ASC in a ClusteredColumnConstraint

GeorgeSittas:
cc @tobymao @dmoore247

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -148,18 +148,6 @@ def _to_date_sql(self: Hive.Generator, expression: exp.TsOrDsToDate) -> str:
     return f"TO_DATE({this})"
 
 
-def clusteredcolumnconstraint_sql(
-    self: Hive.Generator,
-    expression: exp.ClusteredColumnConstraint | exp.NonClusteredColumnConstraint,
-) -> str:
-    expression = expression.copy()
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/hive.py
```diff
@@ -148,6 +148,18 @@ def _to_date_sql(self: Hive.Generator, expression: exp.TsOrDsToDate) -> str:
     return f"TO_DATE({this})"
 
 
+def clusteredcolumnconstraint_sql(
+    self: Hive.Generator,
+    expression: exp.ClusteredColumnConstraint | exp.NonClusteredColumnConstraint,
+) -> str:
+    expression = expression.copy()
```
GeorgeSittas, line 155: I just put this `copy` here to avoid complicating the code, i.e. checking if there's at least one `desc` arg set to `False` etc... Shouldn't be too costly, I think.

tobymao: it needs to be done on all constraints, probably not clustered/nonclustered, those are tsql specific

GeorgeSittas: What other constraint has an `Ordered` child?

tobymao: i don't know

PR #2263, opened 2023-09-19
Fix(tsql): correctly escape single quotes in EXEC

treysp:


sqlglot/dialects/tsql.py
```diff
@@ -671,19 +671,20 @@ def create_sql(self, expression: exp.Create) -> str:
 
             if exists:
                 identifier = self.sql(exp.Literal.string(exp.table_name(table) if table else ""))
+                sql = exp.Literal.string(sql).sql()
```
tobymao, line 674: sql = self.sql(exp.Literal.string(...))

you need the sql generation to be dialect specific

GeorgeSittas (approved):

PR #2271, opened 2023-09-20
feat(optimizer): simplify date_trunc

barakalon:
Simplify DATE_TRUNC functions.

This is particularly useful because of date partitioned tables.

```sql
SELECT
  DATE_TRUNC('week', CAST(ts AS DATE)) AS week,
  SUM(measure) AS metric
FROM fact
WHERE DATE_TRUNC('week', CAST(ts AS DATE)) >= CAST('2023-01-01' AS DATE)
GROUP BY 1
```

The WHERE predicate is kind of silly, but it happens. Especially by programmatically generated queries.

Some engines don't have this optimization, in which case partition pruning will fail.

### DIFF
--- a/sqlglot/_typing.py
+++ b/sqlglot/_typing.py
@@ -4,7 +4,9 @@ import typing as t
 
 import sqlglot
 
-# A little hack for backwards compatibility with Python 3.7, which doesn't have Protocols.
+# A little hack for backwards compatibility with Python 3.7.
+# For example, we might want a TypeVar for objects that support comparison e.g. SupportsRichComparisonT from typeshed.
+# But Python 3.7 doesn't support Protocols, so we'd also need typing_extensions, which we don't want as a dependency.
 A = t.TypeVar("A", bound=t.Any)
 
 E = t.TypeVar("E", bound="sqlglot.exp.Expression")
--- a/sqlglot/_typing.py
+++ b/sqlglot/_typing.py
@@ -4,21 +4,8 @@ import typing as t
 
 import sqlglot
 
+# A little hack for backwards compatibility with Python 3.7, which doesn't have Protocols.
+A = t.TypeVar("A", bound=t.Any)
 
-class Comparable(t.Protocol):
-    def __lt__(self, __other: t.Any) -> bool:
-        ...
-
-    def __le__(self, __other: t.Any) -> bool:
-        ...
-
-    def __gt__(self, __other: t.Any) -> bool:
-        ...
-
-    def __ge__(self, __other: t.Any) -> bool:
-        ...
-
-
-C = t.TypeVar("C", bound=Comparable)
 E = t.TypeVar("E", bound="sqlglot.exp.Expression")
 T = t.TypeVar("T")
--- a/sqlglot/_typing.py
+++ b/sqlglot/_typing.py
@@ -4,5 +4,21 @@ import typing as t
 
 import sqlglot
 
+
+class Comparable(t.Protocol):
+    def __lt__(self, __other: t.Any) -> bool:
+        ...
+
+    def __le__(self, __other: t.Any) -> bool:
+        ...
+
+    def __gt__(self, __other: t.Any) -> bool:
+        ...
+
+    def __ge__(self, __other: t.Any) -> bool:
+        ...
+
+
+C = t.TypeVar("C", bound=Comparable)
 E = t.TypeVar("E", bound="sqlglot.exp.Expression")
 T = t.TypeVar("T")
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4329,8 +4329,8 @@ class DateTrunc(Func):
     arg_types = {"unit": True, "this": True, "zone": False}
 
     @property
-    def unit(self) -> t.Optional[Expression]:
-        return self.args.get("unit")
+    def unit(self) -> Expression:
+        return self.args["unit"]
 
 
 class DatetimeAdd(Func, TimeUnit):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4328,6 +4328,10 @@ class DateDiff(Func, TimeUnit):
 class DateTrunc(Func):
     arg_types = {"unit": True, "this": True, "zone": False}
 
+    @property
+    def unit(self) -> t.Optional[Expression]:
+        return self.args.get("unit")
+
 
 class DatetimeAdd(Func, TimeUnit):
     arg_types = {"this": True, "expression": True, "unit": False}
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4396,6 +4396,10 @@ class TimestampDiff(Func, TimeUnit):
 class TimestampTrunc(Func, TimeUnit):
     arg_types = {"this": True, "unit": True, "zone": False}
 
+    @property
+    def unit(self) -> Expression:
+        return self.args["unit"]
+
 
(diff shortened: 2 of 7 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -530,6 +555,132 @@ def simplify_concat(expression):
     return new_args[0] if len(new_args) == 1 else concat_type(expressions=new_args)
 
 
+def simplify_datetrunc(expression):
+    """Remove redundant DATE_TRUNC functions"""
+    if isinstance(expression, exp.DateTrunc) and expression.unit.name.lower() == "day":
+        return expression.this
```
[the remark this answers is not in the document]
barakalon: Ah, I thought we parsed into TimestampTrunc for those.

Maybe I should add support for that function too...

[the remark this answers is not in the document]
GeorgeSittas: Hm you could be right I didn't double-check this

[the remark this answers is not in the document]
GeorgeSittas: It seems that both Snowflake and Redshift are indeed parsed into TimestampTrunc, so since there's a path to fix future issues (i.e. parse `DATE_TRUNC` into `TimestampTrunc`), this is probably ok?

sqlglot/_typing.py
```diff
@@ -4,5 +4,8 @@
 
 import sqlglot
 
+# A little hack for backwards compatibility with Python 3.7, which doesn't have Protocols.
+A = t.TypeVar("A", bound=t.Any)
```
GeorgeSittas, line 8: Why is `bound=t.Any` needed here?

sqlglot/expressions.py
```diff
@@ -4328,6 +4328,10 @@ class DateDiff(Func, TimeUnit):
 class DateTrunc(Func):
     arg_types = {"unit": True, "this": True, "zone": False}
 
+    @property
+    def unit(self) -> t.Optional[Expression]:
+        return self.args.get("unit")
```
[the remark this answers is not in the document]
barakalon: Well, args isn't typed appropriately. So I think you'd need to assert or cast?

sqlglot/_typing.py
barakalon: Added a comment

sqlglot/optimizer/simplify.py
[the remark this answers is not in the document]
barakalon: I think so. I can also check the type of `this` and make sure its a date.

sqlglot/expressions.py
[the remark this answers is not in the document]
GeorgeSittas: That's a core assumption throughout the codebase though, right?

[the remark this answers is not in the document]
tobymao: why do you need to assert? should be fine

sqlglot/helper.py
```diff
@@ -435,3 +436,26 @@ def dict_depth(d: t.Dict) -> int:
 def first(it: t.Iterable[T]) -> T:
     """Returns the first element from an iterable (useful for sets)."""
     return next(i for i in it)
+
+
+def merge_ranges(ranges: t.List[t.Tuple[A, A]]) -> t.List[t.Tuple[A, A]]:
+    # type: ignore
```
tobymao, line 442: why do you need a type ignore?

sqlglot/helper.py
```diff
@@ -435,3 +436,26 @@ def dict_depth(d: t.Dict) -> int:
...
+    if not ranges:
+        return []
+
+    # First, sort the ranges by starts
+    sorted_ranges = sorted(ranges, key=lambda x: x[0])  # type: ignore
```
tobymao, line 447: can't you just sort? by default it should sort on the first of the tuple

sqlglot/optimizer/simplify.py
```diff
@@ -530,6 +552,134 @@ def simplify_concat(expression):
...
+        # This will always be False, except for NULL values.
+        return None
+
+    return floor, floor + interval(unit)
+
+
+def _datetrunc_eq_expression(left: exp.Expression, drange: DateRange) -> exp.Expression:
+    """Get the logical expression for a date range"""
+    return exp.and_(
+        exp.GTE(this=left.copy(), expression=date_literal(drange[0])),
```
tobymao, line 586: left <= ...

sqlglot/optimizer/simplify.py
```diff
@@ -530,6 +552,134 @@ def simplify_concat(expression):
...
+
+def _datetrunc_neq(
+    left: exp.Expression, date: datetime.date, unit: str
+) -> t.Optional[exp.Expression]:
+    drange = _datetrunc_range(date, unit)
+    if not drange:
+        return None
+
+    return exp.and_(
+        exp.LT(this=left.copy(), expression=date_literal(drange[0])),
```
tobymao, line 610: same

sqlglot/optimizer/simplify.py
```diff
@@ -630,6 +770,56 @@ def date_literal(date):
     )
 
 
+def interval(unit: str, n: int = 1):
+    from dateutil.relativedelta import relativedelta
+
+    if unit == "year":
+        return relativedelta(years=1 * n)
+    elif unit == "quarter":
```
tobymao, line 778: don't need the elif

sqlglot/optimizer/simplify.py
```diff
@@ -630,6 +770,56 @@ def date_literal(date):
...
+    elif unit == "day":
+        return relativedelta(days=1 * n)
+    else:
+        raise UnsupportedUnit(f"Unsupported unit: {unit}")
+
+
+def date_floor(d: datetime.date, unit: str) -> datetime.date:
+    if unit == "year":
+        return datetime.date(year=d.year, month=1, day=1)
+    elif unit == "quarter":
```
tobymao, line 793: same

sqlglot/helper.py
barakalon: whoops. forgot to remove.

sqlglot/helper.py
barakalon: chatgpt may have wrote this function...

let me clean it up.

sqlglot/optimizer/simplify.py
barakalon: We only define the dunder comparison methods on `Condition`. This can lead to attribute error if, for some reason, `left` is not a `Condition`.
Perhaps we should define the dunder methods on Expression?

sqlglot/optimizer/simplify.py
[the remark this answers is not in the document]
barakalon: I actually think we don't need this function. The other one should handle it fine. Will remove.

sqlglot/helper.py
tobymao: seriously? lol please don't introduce chatgpt stuff into this pristine hand crafted code base

sqlglot/optimizer/simplify.py
tobymao: yea let's move that?

sqlglot/helper.py
barakalon: haha just that function.
I thought a thousand interviewees must've written a merge ranges function in their life.

sqlglot/optimizer/simplify.py
barakalon: done

tobymao (approved):

PR #2275, opened 2023-09-21
Feat!: improve transpilation of T-SQL's SET assignment command

GeorgeSittas:
Completes https://github.com/tobymao/sqlglot/pull/2116.

cc: @dmoore247

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -51,7 +51,11 @@ class Databricks(Spark):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
-        PARAMETER_TOKEN = "$"
+        def parameter_sql(self, expression: exp.Parameter) -> str:
+            # This corresponds to "widget parameters" in Databricks, i.e. ${x}
+            # https://docs.databricks.com/en/notebooks/widgets.html#language-sql
+            this = self.sql(expression, "this")
+            return f"${{{this}}}" if expression.args.get("wrapped") else this
 
     class Tokenizer(Spark.Tokenizer):
         HEX_STRINGS = []
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -51,12 +51,6 @@ class Databricks(Spark):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
-        def parameter_sql(self, expression: exp.Parameter) -> str:
-            # This corresponds to "widget parameters" in Databricks, i.e. ${x}
-            # https://docs.databricks.com/en/notebooks/widgets.html#language-sql
-            this = self.sql(expression, "this")
-            return f"${{{this}}}" if expression.args.get("wrapped") else this
-
     class Tokenizer(Spark.Tokenizer):
         HEX_STRINGS = []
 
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/databricks.py
```diff
@@ -51,7 +51,11 @@ class Generator(Spark.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
-        PARAMETER_TOKEN = "$"
+        def parameter_sql(self, expression: exp.Parameter) -> str:
+            # This corresponds to "widget parameters" in Databricks, i.e. ${x}
+            # https://docs.databricks.com/en/notebooks/widgets.html#language-sql
+            this = self.sql(expression, "this")
+            return f"${{{this}}}" if expression.args.get("wrapped") else this
```
GeorgeSittas, line 58: Only Databricks hit the "wrapped" section of this method and it was added in https://github.com/tobymao/sqlglot/pull/1142/files. I didn't generate "widget params" on purpose, because I wasn't sure if this is a widely used feature & if it'd thus make sense to generate `${x}` in place of `@x`.

@dmoore247 thoughts?

dmoore247:
Correct usage in spark sql (3.4.1):
```python
spark.sql("SET myvar=myliteral")
spark.sql("SELECT '${myvar}'")
```

GeorgeSittas:
@dmoore247 I don't think you're using my branch, or I missed something:

```python
➜  sqlglot git:(jo/tsql_set_fixes) python
Python 3.10.13 (main, Aug 24 2023, 22:36:46) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import sqlglot
>>> sqlglot.transpile("set key value", "tsql", "databricks")
['SET key = value']
```

This is not matching your "transpiled sql" output. Can you double check?

GeorgeSittas:
Fixed the `exp.Parameter` generation for Spark, and hence Databricks. I think all cases should be covered now.

dmoore247:
I think the support for SET in spark2 is a bit more limited than the most recent releases in Spark3 and Databricks.

dmoore247:
This irony... `5                    SELECT @KEY ==> SELECT ${KEY}                     <SELECT '${KEY}'> 	?False`

`SELECT ${KEY}` is correct if KEY was set to an expression like `(SELECT 13)`

```python
from pyspark.sql import SparkSession
spark = SparkSession.builder.master("local").getOrCreate()

print( spark.sql("SET myvar=myliteral").collect() )
print( spark.sql("SELECT '${myvar}'").collect() )

print( spark.sql("SET myvar=(SELECT 13)").collect() )
print( spark.sql("SELECT ${myvar}").collect() )


## output
# [Row(key='myvar', value='myliteral')]
# [Row(myliteral='myliteral')]
# [Row(key='myvar', value='(SELECT 13)')]
# [Row(scalarsubquery()=13)]
```

dmoore247:
I think it's more common to have `SET @myvar = (SELECT ....)` so let's merge what you have. Let me sort it out with the engineers w.r.t. literals.

GeorgeSittas:
Ok, waiting for Toby to take a quick look and it's good to go from me. Feel free to base future fixes off of this PR.

sqlglot/dialects/hive.py
```diff
@@ -376,6 +376,7 @@ class Generator(generator.Generator):
         INDEX_ON = "ON TABLE"
         EXTRACT_ALLOWS_QUOTES = False
         NVL2_SUPPORTED = False
+        SUPPORTS_PARAMETERS = False
```
tobymao, line 379: hive does support params

https://cwiki.apache.org/confluence/display/Hive/LanguageManual+VariableSubstitution

GeorgeSittas: I see, will update

PR #2277, opened 2023-09-21
feat(postgres): add support OID postgres type

Nitrino:
Postgres supports data type OID and it is possible to explicitly create a column with this type. 
This PR adds support for OID data type

**Docs**
[Object Identifier Types](https://www.postgresql.org/docs/16/datatype-oid.html)
The list of all data types can be viewed with the command `\dT *`

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3534,6 +3534,7 @@ class DataType(Expression):
         NUMRANGE = auto()
         NVARCHAR = auto()
         OBJECT = auto()
+        OID = auto()
         ROWVERSION = auto()
         SERIAL = auto()
         SET = auto()
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -200,6 +200,7 @@ class Parser(metaclass=_Parser):
         TokenType.VARIANT,
         TokenType.OBJECT,
         TokenType.OBJECT_IDENTIFIER,
+        TokenType.OID,
         TokenType.INET,
         TokenType.IPADDRESS,
         TokenType.IPPREFIX,
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -154,6 +154,7 @@ class TokenType(AutoName):
     IMAGE = auto()
     VARIANT = auto()
     OBJECT = auto()
+    OID = auto()
     INET = auto()
     IPADDRESS = auto()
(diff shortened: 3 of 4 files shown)


GeorgeSittas:
This is already supported, I think you forgot to specify the `read` dialect to be Postgres:

```python
>>> import sqlglot
>>> sqlglot.parse_one("CREATE TABLE public.propertydata (propertyvalue oid)", "postgres")
(CREATE this:
  (SCHEMA this:
    (TABLE this:
      (IDENTIFIER this: propertydata, quoted: False), db:
      (IDENTIFIER this: public, quoted: False)), expressions:
    (COLUMNDEF this:
      (IDENTIFIER this: propertyvalue, quoted: False), kind:
      (OBJECTIDENTIFIER this: oid))), kind: TABLE, exists: False)
>>> print(sqlglot.transpile("CREATE TABLE public.propertydata (propertyvalue oid)", "postgres")[0])
CREATE TABLE public.propertydata (propertyvalue OID)
```

Nitrino:
@GeorgeSittas There's a difference, `OID` is detected as an `OBJECTIDENTIFIER`, not a `DATATYPE`. 
For example, here is how the parsing looks like if the type was `integer`:
```python
>>> sqlglot.parse_one("CREATE TABLE public.propertydata (propertyvalue integer)", "postgres")
(CREATE this:
  (SCHEMA this:
    (TABLE this:
      (IDENTIFIER this: propertydata, quoted: False), db:
      (IDENTIFIER this: public, quoted: False)), expressions:
    (COLUMNDEF this:
      (IDENTIFIER this: propertyvalue, quoted: False), kind:
      (DATATYPE this: Type.INT, nested: False, prefix: False))), kind: TABLE, exists: False)
```

Nitrino:
@GeorgeSittas What do you think of my comment above?

Here is a simple example that causes bugs in my code

```python
expr_int = sqlglot.parse_one("CREATE TABLE public.propertydata (propertyvalue integer)", "postgres")
expr_oid = sqlglot.parse_one("CREATE TABLE public.propertydata (propertyvalue oid)", "postgres")
print(str(expr_int.find(sql_exp.DataType)))
print(str(expr_oid.find(sql_exp.DataType)))

# INT
# None
```

In this case `OID` should act as `DATATYPE`, but is detected as `OBJECTIDENTIFIER`.

PR #2279, opened 2023-09-21
Fix!(postgres): parse create index method (e.g. btree) correctly

### LINKED ISSUE #2278: Support postgres Operator Classes and Operator Families
An index definition can specify an operator class for each column of an index. 
Right now, the library does not support such cases

```python
    sql = """
        CREATE INDEX index_routes_on_path_text_pattern_ops ON public.routes USING btree (path varchar_pattern_ops);
        CREATE INDEX index_ci_pipelines_on_project_idandrefandiddesc ON public.ci_pipelines USING btree (project_id, ref, id DESC);
        CREATE INDEX index_issues_on_title_trigram ON public.issues USING gin (title public.gin_trgm_ops);
    """
    parsed_sql_exprs = sqlglot.parse(sql, read="postgres")
```

**Official Documentation**
https://www.postgresql.org/docs/current/indexes-opclass.html

GeorgeSittas:
Fixes #2278

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2464,7 +2464,7 @@ class Parser(metaclass=_Parser):
             index = self._parse_id_var()
             table = None
 
-        using = self._parse_field() if self._match(TokenType.USING) else None
+        using = self._parse_var(any_token=True) if self._match(TokenType.USING) else None
 
         if self._match(TokenType.L_PAREN, advance=False):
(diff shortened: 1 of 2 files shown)


GeorgeSittas:
Waiting on a followup question I left on the issue, might need to do some more work for this to be complete but this covers the 2nd statement.

PR #2280, opened 2023-09-21
Fix: ensure Expression is not an iterable to avoid inf. loops

GeorgeSittas:
Before:

```python
>>> import sqlglot
>>> for x in sqlglot.parse_one("select 1"):
...     pass
...
<infinite loop - hangs>
```

After:
```python
  File "sqlglot/expressions.py", line 764, in __iter__
    raise TypeError(f"'{self.__class__.__name__}' object is not iterable")
TypeError: 'Select' object is not iterable
```

### DIFF
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -380,7 +380,9 @@ def is_iterable(value: t.Any) -> bool:
     Returns:
         A `bool` value indicating if it is an iterable.
     """
-    return hasattr(value, "__iter__") and not isinstance(value, (str, bytes))
+    from sqlglot import Expression
+
+    return hasattr(value, "__iter__") and not isinstance(value, (str, bytes, Expression))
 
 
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlglot/helper.py
```diff
@@ -380,7 +380,9 @@ def is_iterable(value: t.Any) -> bool:
     Returns:
         A `bool` value indicating if it is an iterable.
     """
-    return hasattr(value, "__iter__") and not isinstance(value, (str, bytes))
+    from sqlglot import Expression
+
+    return hasattr(value, "__iter__") and not isinstance(value, (str, bytes, Expression))
```
GeorgeSittas, line 386: Not sure how expensive this import can be; it was the simplest solution, but perhaps a try-except block where we call `iter(value)` is faster / better?

tobymao: nah it's fine

GeorgeSittas: cool

PR #2281, opened 2023-09-21
feat(optimizer): simplify_equality

barakalon:
We have expressions like this:

```sql
DATE_ADD(
  DATE_ADD(
    DATE_TRUNC(
      'week', 
      DATE_SUB(
        x, 
        1, 
        DAY
      )
    ), 
    1, 
    DAY
  ), 
  1, 
  YEAR
) < CAST('2021-01-08' AS DATE);
```

This optimization, plus the previous operations, gives:
```sql
x < CAST('2020-01-07' AS DATE);
```


tests/fixtures/optimizer/simplify.sql
```diff
@@ -757,3 +757,57 @@ x < CAST('2022-01-01' AS DATE) AND x >= CAST('2021-01-01' AS DATE);
...
+x + 1 <= 3;
+x <= 2;
+
+x + 1 <> 3;
+x <> 2;
+
+1 + x + 1 = 3 + 1;
+x = 2;
+
+x - INTERVAL 1 DAY = CAST('2021-01-01' AS DATE);
```
tobymao, line 794: what would happen if it's 

x - interval 1 day = cast(y as date)

barakalon: Nothing. The right side must be a date literal.

tobymao (approved):

sqlglot/optimizer/simplify.py
```diff
@@ -368,6 +369,87 @@ def absorb_and_eliminate(expression, root=True):
...
+
+        if not a_predicate(a) and b_predicate(b):
+            pass
+        elif not a_predicate(b) and b_predicate(a):
+            a, b = b, a
+        else:
+            return expression
+
+        return expression.__class__(
+            this=a.copy(), expression=INVERSE_OPS[l.__class__](this=r.copy(), expression=b.copy())
```
tobymao, line 448: are these copies needed?

barakalon: ah shit sorry i hit merge.
I'm not sure.
In general, I copy when creating a new expression using child nodes from an existing tree. When do you?

tobymao: in the optimizer, i only copy when i have to, because it's mutation heavy and expensive

barakalon: https://github.com/tobymao/sqlglot/pull/2282

GeorgeSittas: It seems like we're dropping the `expression` arg here anyway, so the old AST is replaced by this new one.. so it doesn't seem like the copies are needed? They'd be needed if you kept around `expression`, because then its children would have their `parent` changed and the AST would be corrupted.

PR #2283, opened 2023-09-21
Fix: improve performance of VALUES -> UNION ALL transpilation

GeorgeSittas:
Before this fix, it wasn't possible to transpile a 10k sized `VALUES` clause into Redshift without increasing the recursion limit, due to the deep nesting of the generated `exp.Union` expressions.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1407,9 +1407,6 @@ class Generator:
             return f"{values} AS {alias}" if alias else values
 
         # Converts `VALUES...` expression into a series of select unions.
-        # Note: If you have a lot of unions then this will result in a large number of recursive statements to
-        # evaluate the expression. You may need to increase `sys.setrecursionlimit` to run and it can also be
-        # very slow.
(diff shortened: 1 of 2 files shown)


sqlglot/generator.py
```diff
@@ -1425,14 +1422,9 @@ def values_sql(self, expression: exp.Values) -> str:
...
 
-        subquery_expression: exp.Select | exp.Union = selects[0]
-        if len(selects) > 1:
-            for select in selects[1:]:
-                subquery_expression = exp.union(
-                    subquery_expression, select, distinct=False, copy=False
-                )
-
-        return self.subquery_sql(subquery_expression.subquery(expression.alias, copy=False))
+        alias = f" AS {expression.alias}" if expression.alias else ""
```
tobymao, line 1425: is this only done for non pretty?

GeorgeSittas: Nope - that's a good point though. I can make it so that this is only done for non-pretty.

PR #2291, opened 2023-09-22
fix(bigquery): reverse log argument order

cpcloud:
Reverse logarithm argument order according to https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#log.


cpcloud:
Hm, lemme rethink this for a sec, changing `LOG_BASE_FIRST` to `True` doesn't make sense according to the documentation.

cpcloud:
Actually, nevermind. This is correct, I'm just "holding it wrong" 🙂

cpcloud:
I'll open an issue though, because I'm not sure whether the behavior I'm seeing is correct.

PR #2293, opened 2023-09-22
fix(bigquery): preserve log argument order when parsing and generation dialects match

### LINKED ISSUE #2292: should bigquery generation reverse the order of `log` arguments?
Wondering if this is the correct behavior:

```
In [3]: sg.parse_one("LOG(x, y)").sql("bigquery")
Out[3]: 'LOG(x, y)'

In [4]: sg.parse_one("LOG(x, y)", read="bigquery").sql("bigquery")
Out[4]: 'LOG(y, x)'
```

How can sqlglot know my intent here? Is it that sqlglot's behavior is to assume that the meaning of any `log` call is always base-first?

cpcloud:
Fixes #2292

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -178,6 +178,7 @@ class BigQuery(Dialect):
     UNNEST_COLUMN_ONLY = True
     SUPPORTS_USER_DEFINED_TYPES = False
     SUPPORTS_SEMI_ANTI_JOIN = False
+    LOG_BASE_FIRST = False
 
     # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#case_sensitivity
     RESOLVES_IDENTIFIERS_AS_UPPERCASE = None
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -419,6 +419,7 @@ class Generator(generator.Generator):
         RENAME_TABLE_WITH_DB = False
         NVL2_SUPPORTED = False
         UNNEST_WITH_ORDINALITY = False
+        LOG_BASE_FIRST = False
```
GeorgeSittas, line 422: Since this flag is used in both the `Parser` and the `Generator`, can we make it a `Dialect` constant instead? Check for example how `ALIAS_POST_TABLESAMPLE` is handled.

GeorgeSittas (commented):
LGTM, just one minor suggestion. Thanks!

sqlglot/dialects/bigquery.py
cpcloud: Lemme give it shot.

cpcloud: Done!

GeorgeSittas (approved):

sqlglot/dialects/bigquery.py
GeorgeSittas: Added for consistency: https://github.com/tobymao/sqlglot/commit/79fc4d1a990cd4b86d55e5a1ce8c37ce5556765c

PR #2294, opened 2023-09-22
test: allow running the test suite without having pyspark installed

cpcloud:
Would love to be able to run `pytest` full stop without having to install `pyspark`. This PR skips test files that require pyspark for test collection.


cpcloud:
Ah, for some reason I that y'all used pytest, nevermind!

GeorgeSittas:
> Ah, for some reason I that y'all used pytest, nevermind!

Yep, was just about to comment about `unittest`! You can skip integration tests that use `pyspark` with `make unit`.

PR #2295, opened 2023-09-22
Fix(parser): make kwarg parsing more robust

### LINKED ISSUE #2288: Parser fails when using "interval" as name for a struct field
This is a snippet that reproduces the problem:

```python
import sqlglot

working_sql = """
CREATE TABLE my_table (
  col STRUCT<`interval`: DOUBLE COMMENT 'aaa'>
)
"""

broken_sql = """
CREATE TABLE my_table (
  col STRUCT<interval: DOUBLE COMMENT 'aaa'>
)
"""

sqlglot.parse_one(working_sql, read="databricks")
print("If 'interval' is surrounded by backticks it works")

print("But if it is not then it breaks:")
sqlglot.parse_one(broken_sql, read="databricks")
```

As far as I can see (please correct me if I'm wrong), **even without backticks the parser should work**. Relevant links:
- https://docs.databricks.com/en/sql/language-manual/data-types/struct-type.html
- https://docs.databricks.com/en/sql/language-manual/sql-ref-identifiers.html
- https://docs.databricks.com/en/sql/language-manual/sql-ref-ansi-compliance.html#sql-keywords -> even if `spark.sql.ansi.enabled` is true, `interval` is not a reserved keyword

Thanks!

GeorgeSittas:
Fixes #2288

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3426,7 +3426,11 @@ class Parser(metaclass=_Parser):
         return this
 
     def _parse_struct_types(self) -> t.Optional[exp.Expression]:
-        this = self._parse_type() or self._parse_id_var()
+        if self._next and self._next.token_type == TokenType.COLON:
+            this = self._parse_id_var()
+        else:
+            this = self._parse_type() or self._parse_id_var()
+
         self._match(TokenType.COLON)
         return self._parse_column_def(this)
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3249,8 +3249,8 @@ class Parser(metaclass=_Parser):
             return self.UNARY_PARSERS[self._prev.token_type](self)
         return self._parse_at_time_zone(self._parse_type())
 
-    def _parse_type(self) -> t.Optional[exp.Expression]:
-        interval = self._parse_interval()
+    def _parse_type(self, parse_interval: bool = True) -> t.Optional[exp.Expression]:
+        interval = parse_interval and self._parse_interval()
         if interval:
             return interval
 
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -3426,7 +3426,11 @@ def _parse_types(
         return this
 
     def _parse_struct_types(self) -> t.Optional[exp.Expression]:
-        this = self._parse_type() or self._parse_id_var()
+        if self._next and self._next.token_type == TokenType.COLON:
```
tobymao, line 3429: not sure i understand this change, struct<interval double> is also legal, the semi colon is optional

GeorgeSittas: The issue here was that `interval: int` was parsed as an `Interval` with a `Placeholder` child. So the idea is: if we have <something> <colon> ... then this is probably a "column definition" (in parser terms), and so we want to parse <something> as an id var to avoid this scenario. `STRUCT(interval)` works fine if that's what you mean.

tobymao: what about struct(interval double)

GeorgeSittas:
```python
>>> sqlglot.parse_one("struct(interval double)")
(STRUCT expressions:
  (INTERVAL this:
    (COLUMN this:
      (IDENTIFIER this: double, quoted: False))))
```

I see what you mean, didn't remember that `STRUCT(<name> <type>)` was valid

GeorgeSittas: What if we dodge `_parse_interval` in `_parse_type` with a flag then?

```python
def _parse_type(self, parse_interval: bool = True) -> ...
```

tobymao: sure

tobymao (approved):

PR #2296, opened 2023-09-22
Fix!(databricks): allow dotted identifiers as keys in TBLPROPERTIES

### LINKED ISSUE #2289: Parser fails if table property key is not quoted
This is a snippet that reproduces the problem:

```python
import sqlglot

working_sql = """
CREATE TABLE my_table ()
TBLPROPERTIES (
  'delta.dataSkippingNumIndexedCols' = 15
);
"""

broken_sql = """
CREATE TABLE my_table ()
TBLPROPERTIES (
  delta.dataSkippingNumIndexedCols = 15
);
"""

sqlglot.parse_one(working_sql, read="databricks")
print("If the table property key is a string literal it works")

print("But if it is not then it breaks:")
sqlglot.parse_one(broken_sql, read="databricks")
```

According to [Databricks' documentation](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-tblproperties.html#syntax), table property keys can be string literals or dot-separated identifiers. 

Thanks!

GeorgeSittas:
Fixes #2289


sqlglot/parser.py
```diff
@@ -1422,6 +1422,18 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
                 this=key,
                 value=self._parse_column() or self._parse_var(any_token=True),
             )
+        else:
```
tobymao, line 1425: why don't you just match for non strings? why this whole branch?

GeorgeSittas: I'll do another pass and try to refactor this

GeorgeSittas: Please take another look, tried to simplify things a bit.

sqlglot/expressions.py
```diff
@@ -1185,7 +1185,7 @@ def to_dot(self) -> Dot:
                 parts.append(parent.expression)
             parent = parent.parent
 
-        return Dot.build(deepcopy(parts))
+        return Dot.build(deepcopy(parts)) if len(parts) > 1 else parts[0]
```
GeorgeSittas, line 1188: This would previously break in `Dot.build` in cases where `parts` only had one item.

sqlglot/parser.py
```diff
@@ -1410,20 +1410,18 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
...
             return self.expression(exp.SqlSecurityProperty, definer=self._match_text_seq("DEFINER"))
 
-        assignment = self._match_pair(
-            TokenType.VAR, TokenType.EQ, advance=False
-        ) or self._match_pair(TokenType.STRING, TokenType.EQ, advance=False)
+        index = self._index
+        key = self._parse_column()
 
-        if assignment:
-            key = self._parse_var_or_string()
```
GeorgeSittas, line 1418: This PR is now breaking on an AST level because `key`s have become `Identifier`s (current) vs `Var`s (before). Besides that I think it should be fine?

tobymao (approved):

PR #2298, opened 2023-09-22
Fix(redshift): generate correct SQL VALUES clause alias

GeorgeSittas:
We're currently using the raw alias name when converting `VALUES` to `UNION ALL`, so we drop any quotes:

```python
>>> import sqlglot
>>> sqlglot.transpile('select * from (values (1)) as `t`(a)', 'spark', 'redshift')
['SELECT * FROM (SELECT 1 AS a) AS t']
```


GeorgeSittas:
Still a WIP, working on another fixup

tobymao (approved):

PR #2299, opened 2023-09-22
feat(optimizer): replace date funcs

barakalon:


tests/fixtures/optimizer/canonicalize.sql
```diff
@@ -28,3 +28,15 @@ SELECT "x"."a" AS "a" FROM "x" AS "x" GROUP BY "x"."a" HAVING SUM("x"."b") <> 0
...
 SELECT a FROM x WHERE 1;
 SELECT "x"."a" AS "a" FROM "x" AS "x" WHERE 1 <> 0;
+
+--------------------------------------
+-- Replace date functions
+--------------------------------------
+DATE('2023-01-01');
+CAST('2023-01-01' AS DATE);
+
+TIMESTAMP('2023-01-01');
```
tobymao, line 38: is this right, shouldn't you cast this to timestamp?

barakalon: Whoops, yes

PR #2300, opened 2023-09-22
Fix/identity tsql->databricks

dmoore247:
transpile from tsql to databricks was not generating the `ALWAYS` which is required by Databricks. 

>>> import sqlglot
>>> sqlglot.transpile('''CREATE TABLE dbo.Products (ProductID INT GENERATED AS IDENTITY (START WITH 1 INCREMENT BY 1) NOT NULL)''',read='tsql',write='databricks')
['CREATE TABLE dbo.Products (ProductID INT GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1) NOT NULL)']


sqlglot/dialects/databricks.py
```diff
@@ -51,6 +51,15 @@ class Generator(Spark.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
+        PARAMETER_TOKEN = "$"
+
+        def generatedasidentitycolumnconstraint_sql(
+            self, expression: exp.GeneratedAsIdentityColumnConstraint
+        ) -> str:
+            expression = expression.copy()
+            expression.set("this", " ")  # trigger ALWAYS in super class
```
tobymao, line 60: it should be True

sqlglot/dialects/databricks.py
```diff
@@ -51,6 +51,15 @@ class Generator(Spark.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
+        PARAMETER_TOKEN = "$"
```
tobymao, line 54: why did you add this back in? i think @GeorgeSittas removed it

dmoore247: I'll double check.

sqlglot/dialects/databricks.py
dmoore247: Meaning set "this" to `True` ?

tobymao: yes

sqlglot/dialects/databricks.py
dmoore247: George did remove it, my merge divereged so my mistake. Fixing...

dmoore247:
Thanks!Sent via my iPhoneOn Sep 22, 2023, at 11:55 PM, Toby Mao ***@***.***> wrote:﻿
Merged #2300 into main.

Issue #2301, opened 2023-09-23
bug: Databricks COMPUTED column is parsed as a `GENERATEDASIDENTITYCOLUMNCONSTRAINT`


dmoore247:
Databricks COMPUTED column is parsed as a `GENERATEDASIDENTITYCOLUMNCONSTRAINT` causing computed column to be rendered as IDENTITY column

**Fully reproducible code snippet**
```python
>>> import sqlglot

>>> sqlglot.parse_one('CREATE TABLE foo (x INT GENERATED ALWAYS AS (YEAR(y)))',read='databricks')
(CREATE this: 
  (SCHEMA this: 
    (TABLE this: 
      (IDENTIFIER this: foo, quoted: False)), expressions: 
    (COLUMNDEF this: 
      (IDENTIFIER this: x, quoted: False), kind: 
      (DATATYPE this: Type.INT, nested: False, prefix: False), constraints: 
      (COLUMNCONSTRAINT kind: 
        (GENERATEDASIDENTITYCOLUMNCONSTRAINT this: True, expression: 
          (YEAR this: 
            (TSORDSTODATE this: 
              (COLUMN this: 
                (IDENTIFIER this: y, quoted: False)))))))), kind: TABLE, exists: False)


>>> sqlglot.transpile('CREATE TABLE foo (x INT GENERATED ALWAYS AS (YEAR(y)))', read='databricks',write='tsql')
['CREATE TABLE foo (x INTEGER IDENTITY(1, 1))']
```

**Official Documentation**
https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html

```
{ { [CREATE OR] REPLACE TABLE | CREATE [EXTERNAL] TABLE [ IF NOT EXISTS ] }
  table_name
  [ table_specification ]
  [ USING data_source ]
  [ table_clauses ]
  [ AS query ] }

table_specification
  ( { column_identifier column_type [ NOT NULL ]
      [ GENERATED ALWAYS AS ( expr ) |
        GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( [ START WITH start ] [ INCREMENT BY step ] ) ] |
        DEFAULT default_expression ]
      [ COMMENT column_comment ]
      [ column_constraint ] } [, ...]
    [ , table_constraint ] [...] )

table_clauses
  { OPTIONS clause |
    PARTITIONED BY clause |
    clustered_by_clause |
[...]

PR #2303, opened 2023-09-23
Fix!(databricks): parse GENERATED ALWAYS AS into ComputedColumnConstraint

dmoore247:
fixes https://github.com/tobymao/sqlglot/issues/2301 the parsing of `GENERATED ALWAYS AS` is now into a ComputedColumnConstraint

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -72,22 +72,6 @@ class Databricks(Spark):
             this = self.sql(expression, "this")
             return f"GENERATED ALWAYS AS {this}"
 
-        def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
-            c = expression.find(exp.GeneratedAsIdentityColumnConstraint)
-            kind = self.sql(expression, "kind")
-            if c and isinstance(expression.this, exp.Identifier) and kind == "INT":
-                # only BIGINT GENERATED ALWAYS AS IDENTITY constraints are supported, upcast to BIGINT
-                expression = expression.copy()
-                expression.set("kind", "BIGINT")
-            return super().columndef_sql(expression, sep)
-
-        def generatedasidentitycolumnconstraint_sql(
-            self, expression: exp.GeneratedAsIdentityColumnConstraint
-        ) -> str:
-            expression = expression.copy()
-            expression.set("this", True)  # trigger ALWAYS in super class
-            return super().generatedasidentitycolumnconstraint_sql(expression)
-
     class Tokenizer(Spark.Tokenizer):
         HEX_STRINGS = []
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/databricks.py
```diff
@@ -23,6 +25,21 @@ class Parser(Spark.Parser):
...
+            self._match_text_seq("ALWAYS")
+            self._match(TokenType.ALIAS)
+            this = self._parse_expression()
+
+            return self.expression(
+                exp.ComputedColumnConstraint,
+                this=this,
+            )
+
+        CONSTRAINT_PARSERS = {
```
tobymao, line 38: move constraint parsers above the function

sqlglot/dialects/databricks.py
```diff
@@ -23,6 +25,21 @@ class Parser(Spark.Parser):
             TokenType.COLON: exp.JSONExtract,
         }
 
+        def _parse_generated(self, kind: t.Optional[str] = None) -> exp.ComputedColumnConstraint:
```
tobymao, line 28: should this be in hive or spark class instead or is this databricks only?

sqlglot/dialects/databricks.py
```diff
@@ -51,6 +68,10 @@ class Generator(Spark.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
+        def computedcolumnconstraint_sql(self, expression: exp.ComputedColumnConstraint) -> str:
```
tobymao, line 71: can you link the documentation for computed column constraint? why is generated always as a computed column constraint?

sqlglot/dialects/databricks.py
GeorgeSittas: Why define a new method for this? We could also reuse `_parse_generated_as_identity` and then construct the appropriate `ComputedColumnConstraint` instance from it to avoid duplication.

sqlglot/dialects/databricks.py
GeorgeSittas: Ditto - if we override `_parse_generated_as_identity` we can remove this altogether.

sqlglot/dialects/databricks.py
GeorgeSittas: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16 I think the clause of interest is `computed_column_definition` in the spec.

sqlglot/dialects/databricks.py
GeorgeSittas: It should be Spark and Databricks (not Spark2), see https://github.com/apache/spark/pull/38823

GeorgeSittas:
Closed in favor of #2306.

dmoore247:
Thanks!Sent via my iPhoneOn Sep 23, 2023, at 11:47 AM, Jo ***@***.***> wrote:﻿
Closed #2303.

PR #2306, opened 2023-09-23
Fix!(spark): parse computed column constraints appropriately

### LINKED ISSUE #2301: bug: Databricks COMPUTED column is parsed as a `GENERATEDASIDENTITYCOLUMNCONSTRAINT`

GeorgeSittas:
Fixes #2301

References:
- https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html
- https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/relational-databases/tables/specify-computed-columns-in-a-table?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16
- https://github.com/apache/spark/pull/38823

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -207,7 +207,7 @@ class Generator:
     # The keyword used to clone a table in a DDL statement
     CLONE_KEYWORD = "CLONE"
 
-    # Whether or not to incldue the type of a computed column in the CREATE DDL
+    # Whether or not to include the type of a computed column in the CREATE DDL
     COMPUTED_COLUMN_WITH_TYPE = True
 
     TYPE_MAPPING = {
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -207,6 +207,9 @@ class Generator:
     # The keyword used to clone a table in a DDL statement
     CLONE_KEYWORD = "CLONE"
 
+    # Whether or not to incldue the type of a computed column in the CREATE DDL
+    COMPUTED_COLUMN_WITH_TYPE = True
+
     TYPE_MAPPING = {
         exp.DataType.Type.NCHAR: "CHAR",
         exp.DataType.Type.NVARCHAR: "VARCHAR",
@@ -651,6 +654,9 @@ class Generator:
         constraints = f" {constraints}" if constraints else ""
         position = self.sql(expression, "position")
         position = f" {position}" if position else ""
+
+        if expression.find(exp.ComputedColumnConstraint) and not self.COMPUTED_COLUMN_WITH_TYPE:
+            kind = ""
 
         return f"{exists}{column}{kind}{constraints}{position}"
 
--- a/tests/dialects/test_databricks.py
+++ b/tests/dialects/test_databricks.py
@@ -32,6 +32,7 @@ class TestDatabricks(Validator):
             "CREATE TABLE foo (x INT GENERATED ALWAYS AS (YEAR(y)))",
             write={
                 "databricks": "CREATE TABLE foo (x INT GENERATED ALWAYS AS (YEAR(TO_DATE(y))))",
+                "tsql": "CREATE TABLE foo (x INTEGER AS YEAR(CAST(y AS DATE)))",
             },
         )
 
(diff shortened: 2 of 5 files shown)


tests/dialects/test_databricks.py
```diff
@@ -32,6 +32,7 @@ def test_databricks(self):
             "CREATE TABLE foo (x INT GENERATED ALWAYS AS (YEAR(y)))",
             write={
                 "databricks": "CREATE TABLE foo (x INT GENERATED ALWAYS AS (YEAR(TO_DATE(y))))",
+                "tsql": "CREATE TABLE foo (x INTEGER AS YEAR(CAST(y AS DATE)))",
```
GeorgeSittas, line 35: Double-checking if this is correct in T-SQL..

tobymao (approved):

tests/dialects/test_databricks.py
GeorgeSittas: It looks like we need to drop `INTEGER` here, otherwise it's a syntax error, cc @dmoore247

GeorgeSittas: Just verified that the following statement is valid T-SQL.

```sql
CREATE TABLE foo (x DATETIME, y AS YEAR(CAST(x AS DATE)))
```

sqlglot/generator.py
```diff
@@ -207,6 +207,9 @@ class Generator:
     # The keyword used to clone a table in a DDL statement
     CLONE_KEYWORD = "CLONE"
 
+    # Whether or not to incldue the type of a computed column in the CREATE DDL
```
GeorgeSittas, line 210:
```suggestion
    # Whether or not to include the type of a computed column in the CREATE DDL
```

PR #2309, opened 2023-09-23
Fix!(databricks): upgrade IDENTITY column type to BIGINT if INT 

dmoore247:
Only BIGINT column type is supported on Databricks as IDENTITY column.
It's harmless to upgrade INT to BIGINT for identity columns and is what a human transpiler would do.
a INT column type will generate a parse exception in spark/databricks.

**Official Documentation**
https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html#parameters

GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( [ START WITH start ] [ INCREMENT BY step ] ) ]

Applies to: check marked yes Databricks SQL check marked yes Databricks Runtime 10.3 and above

Defines an identity column. When you write to the table, and do not provide values for the identity column, it will be automatically assigned a unique and statistically increasing (or decreasing if step is negative) value. This clause is only supported for Delta Lake tables. This clause can only be used for columns with **BIGINT** data type.

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -52,12 +52,12 @@ class Databricks(Spark):
         }
 
         def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
-            c = expression.find(exp.GeneratedAsIdentityColumnConstraint)
+            constraint = expression.find(exp.GeneratedAsIdentityColumnConstraint)
             kind = self.sql(expression, "kind")
-            if c and isinstance(expression.this, exp.Identifier) and kind == "INT":
+            if constraint and isinstance(expression.this, exp.Identifier) and kind == "INT":
                 # only BIGINT generated identity constraints are supported
                 expression = expression.copy()
-                expression.set("kind", "BIGINT")
+                expression.set("kind", exp.DataType.build("bigint"))
             return super().columndef_sql(expression, sep)
 
         def generatedasidentitycolumnconstraint_sql(
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -51,6 +51,15 @@ class Databricks(Spark):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
+        def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
+            c = expression.find(exp.GeneratedAsIdentityColumnConstraint)
+            kind = self.sql(expression, "kind")
+            if c and isinstance(expression.this, exp.Identifier) and kind == "INT":
+                # only BIGINT generated identity constraints are supported
+                expression = expression.copy()
+                expression.set("kind", "BIGINT")
+            return super().columndef_sql(expression, sep)
+
         def generatedasidentitycolumnconstraint_sql(
             self, expression: exp.GeneratedAsIdentityColumnConstraint
         ) -> str:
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/databricks.py
```diff
@@ -51,6 +51,15 @@ class Generator(Spark.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
+        def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
+            c = expression.find(exp.GeneratedAsIdentityColumnConstraint)
+            kind = self.sql(expression, "kind")
+            if c and isinstance(expression.this, exp.Identifier) and kind == "INT":
+                # only BIGINT generated identity constraints are supported
+                expression = expression.copy()
+                expression.set("kind", "BIGINT")
```
GeorgeSittas, line 60: I'd rename `c` to `constraint` or something. Also, what if it's a different kind of integer, besides `INT`?

Finally, a couple of general comments:
- Compare types using the `is_type` helper, e.g. `kind.is_type("int")`.
- Imitate the AST produced by `parse_one`, so `kind` here should be set to `exp.DataType.build("bigint")`.

GeorgeSittas:
Let's try to stick to the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format for the next PRs. Do we need to update the Spark dialect instead of Databricks for this?

sqlglot/dialects/databricks.py
dmoore247: @GeorgeSittas 
If type is already BIGINT, nothing to do, if type is date or string or guid, then that's maybe way off semantically to coerce. If type is a SHORT or BYTE, which probably are probably not used. If INT64, INT128 those could be coerced to BIGINT... Thoughts?

dmoore247: `kind` is a `str` and it does not have a method `.is_type()`
When searching, `this` and `expression` are used with `.is_type() helper`

dmoore247:
Spark + Delta Lake doesn't have IDENTITY column type at this time so this should just go into the Databricks dialect for now.

sqlglot/dialects/databricks.py
```diff
@@ -51,6 +51,15 @@ class Generator(Spark.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
+        def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
+            constraint = expression.find(exp.GeneratedAsIdentityColumnConstraint)
+            kind = self.sql(expression, "kind")
+            if constraint and isinstance(expression.this, exp.Identifier) and kind == "INT":
```
GeorgeSittas, line 57: What's the purpose of the instance check here? Also, the question remains: what if the column definition has a different integer type? For example, if you were transpiling a `SMALLINT` column from T-SQL to Databricks, then you wouldn't convert that into a `BIGINT` because you'd miss this branch. I think the correct approach is to get `kind = expression.args.get("kind")` and if it's a `DataType` then check if `kind.this` is in `DataType.INTEGER_TYPES`.

Another question here is what happens when you have integer types w/ length associated, e.g. `BIGINT(<num>)`. Right now we drop `<num>`, but should we perhaps keep it?

dmoore247: - Spark doesn't support `INT(<num>)`, `BIGINT(<num>)` so I don't think it helps to preserve `<num>`, dropping <num> is appropriate.
- Thanks for the pointer to `DataTypeINTEGER_TYPES`

GeorgeSittas (approved):

dmoore247:
Thanks all!

PR #2311, opened 2023-09-24
fix(clickhouse): fix incorrect array param generation for clickhouse quantiles

cpcloud:


sqlglot/dialects/clickhouse.py
```diff
@@ -21,6 +21,16 @@ def _lower_func(sql: str) -> str:
     return sql[:index].lower() + sql[index:]
 
 
+def _quantile(self, e):
```
GeorgeSittas, line 24: Nit: I'd add the `_sql` suffix for consistency:

```suggestion
def _quantile_sql(self, e):
```

sqlglot/dialects/clickhouse.py
```diff
@@ -21,6 +21,16 @@ def _lower_func(sql: str) -> str:
...
+    quantile = e.args.get("quantile")
```
GeorgeSittas, line 25: We can omit the `get` call here since `quantile` is a required arg:

```suggestion
    quantile = e.args["quantile"]
```

GeorgeSittas (commented):
Looks look, a couple of small suggestions

GeorgeSittas (approved):

PR #2314, opened 2023-09-25
Fix: generate comments before prepending CTEs

### LINKED ISSUE #2313: Comment order appears out of sync when CTEs are included (multiple dialects)

I tested the below on multiple dialects (e.g. bigquery, databricks, snowflake), and all have the same issue. 

```
from sqlglot import exp, maybe_parse

sql = """
with cte as (
    select
        a,
        b
    from tbl 
)
/* test comment */
select /* test comment 2 */
   * 
from cte
"""

print(maybe_parse(sql, dialect = "snowflake").sql(dialect = "snowflake", comments = True))
```

**Actual output**
```
/* test comment */ /* test comment 2 */ WITH cte AS (SELECT a, b FROM tbl) SELECT * FROM cte
```

**Expected output**
ideally
```
 WITH cte AS (  SELECT a, b FROM tbl)  /* test comment */ SELECT /* test comment 2 */ * FROM cte
```
or at least 
```
 WITH cte AS (  SELECT a, b FROM tbl)  /* test comment */ /* test comment 2 */  SELECT * FROM cte
```

GeorgeSittas:
Fixes #2313

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -592,10 +592,13 @@ class Generator:
         else:
             raise ValueError(f"Expected an Expression. Received {type(expression)}: {expression}")
 
-        sql = self.maybe_comment(sql, expression) if self.comments and comment else sql
+        # We've already generated the comments in prepend_ctes for expressions that have CTEs
+        if self.comments and comment and "with" not in expression.arg_types:
+            sql = self.maybe_comment(sql, expression)
 
         if self._cache is not None:
             self._cache[expression_id] = sql
+
         return sql
 
     def uncache_sql(self, expression: exp.Uncache) -> str:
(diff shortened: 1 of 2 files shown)


sqlglot/generator.py
```diff
@@ -592,10 +592,13 @@ def sql(
         else:
             raise ValueError(f"Expected an Expression. Received {type(expression)}: {expression}")
 
-        sql = self.maybe_comment(sql, expression) if self.comments and comment else sql
+        # We've already generated the comments in prepend_ctes for expressions that have CTEs
+        if self.comments and comment and "with" not in expression.arg_types:
```
tobymao, line 596: this doesn't seem to be the most robust way to tell if something has prepend_ctes called on it

tobymao: what if you store a dict of commented things instead

GeorgeSittas: Sure I think that works too, this just seemed like a simpler approach but I see why it may not be very robust. I'll do a refactor and implement your suggestion instead.

PR #2319, opened 2023-09-25
fix(optimizer): a couple simplify_date_trunc enhancements

barakalon:
Fixes a little bug when casting a non-literal to a date.
Also handles nested casts.

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -731,6 +731,9 @@ def simplify_datetrunc_predicate(expression: exp.Expression) -> exp.Expression:
         unit = l.unit.name.lower()
         date = extract_date(r)
 
+        if not date:
+            return expression
+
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -811,18 +822,49 @@ def eval_boolean(expression, a, b):
     return None
 
 
-def extract_date(cast):
-    # The "fromisoformat" conversion could fail if the cast is used on an identifier,
-    # so in that case we can't extract the date.
+def cast_as_date(value: t.Any) -> t.Optional[datetime.date]:
+    if isinstance(value, datetime.datetime):
+        return value.date()
+    elif isinstance(value, datetime.date):
```
tobymao, line 828: don't need ielif same with line 839

tobymao (approved):

PR #2322, opened 2023-09-26
Fix!(spark): create temporary table needs provider

dmoore247:
Spark's create temporary table needs provider (`using parquet`)

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -96,3 +96,17 @@ class Spark(Spark2):
                 return self.func("DATEDIFF", unit, start, end)
 
             return self.func("DATEDIFF", end, start)
+
+        def create_sql(self, expression: exp.Create) -> str:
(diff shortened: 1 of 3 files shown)


tobymao:
is there documentation for this?

GeorgeSittas:
I could verify that Spark ≥v3.4.0 doesn't like a `CREATE TEMPORARY TABLE` without a provider:

```
spark-sql (default)> CREATE TEMPORARY TABLE mytemp (a INT, b CHAR(2), c TIMESTAMP, d FLOAT);

CREATE TEMPORARY TABLE without a provider is not allowed.(line 1, pos 0)

== SQL ==
CREATE TEMPORARY TABLE mytemp (a INT, b CHAR(2), c TIMESTAMP, d FLOAT)
^^^
```

dmoore247:
@tobymao Can't locate official documentation on this. This requirement was discovered by looking at the error messages from OS Pyspark as documented by @GeorgeSittas

PR #2324, opened 2023-09-26
Fix!: raise ParseError on duplicate query modifiers

herbert-allium:
Currently, when sqlglot parses SQL queries with duplicate query modifiers, it silently discards everything except the last one.

For example:
```sql
SELECT * FROM foo WHERE bar1 WHERE bar2 WHERE bar3
```
is parsed as if it were `SELECT * FROM foo WHERE bar3`.

This PR throws an error on such cases instead.

### Impact

I'm not sure how big the impact of this change will be, but I'm not surprised if it is quite big. I have added some tests and all the current tests pass but I am cautious that this might silently break other things.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2223,6 +2223,8 @@ class Parser(metaclass=_Parser):
                     key, expression = parser(self)
 
                     if expression:
+                        if this.args.get(key):
+                            raise ParseError(f"{key} is already set")
                         this.set(key, expression)
                         if key == "limit":
                             offset = expression.args.pop("offset", None)
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -225,6 +225,12 @@ class TestParser(unittest.TestCase):
         with self.assertRaises(ParseError):
             parse_one("SELECT foo( FROM bar")
 
+        with self.assertRaises(ParseError):
+            parse_one("SELECT * FROM foo WHERE true WHERE false")
+
+        with self.assertRaises(ParseError):
+            parse_one("SELECT * FROM foo LIMIT 10 LIMIT 100")
+
         self.assertEqual(
             parse_one(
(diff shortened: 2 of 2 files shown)


sqlglot/parser.py
```diff
@@ -2223,6 +2223,8 @@ def _parse_query_modifiers(
                     key, expression = parser(self)
 
                     if expression:
+                        if this.args.get(key):
```
herbert-allium, line 2226: I am also not sure if `this.args` is supposed to be an internal variable to the `Expression` class or not, but I am not able to find any public methods for it

herbert-allium, line 2226: I am also not sure if `this.args` is supposed to be an internal variable to the `Expression` class or not, but I am not able to find any public methods for it

GeorgeSittas:
SQLGlot doesn't aim to be a validator, it's the user's responsibility to provide valid SQL, so I'm not sure what the added value is here. What triggered the need for this change?

herbert-allium:
The syntax tree returned is not accurate though, I think one would expect that either there are two where clauses in the syntax tree (if we supply two where clauses) or it raises a syntax error.

Anyway, this is the first time I am aware that validation is not SQLGlot's goal (also just noticed it in the README), so apologies for that! Feel free to close this PR if you think this is not aligned with the direction of the project :)

GeorgeSittas:
> The syntax tree returned is not accurate though, I think one would expect that either there are two where clauses in the syntax tree (if we supply two where clauses) or it raises a syntax error.

Yep, there are no guarantees about validation in SQLGlot, so the reasoning here is generally garbage in -> garbage out.

I'll go ahead and close this one, but appreciate you taking the time to contribute to the project. If you wanna chat, you can also reach us in [slack](https://tobikodata.com/slack), e.g. for developing future PRs.

PR #2326, opened 2023-09-26
Fix!(hive): remove UNIQUE constraints as unsupported

dmoore247:
Hive/Spark/Databricks do not support unique column constraints.


sqlglot/dialects/hive.py
```diff
@@ -480,6 +480,13 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        def create_sql(self, expression: exp.Create) -> str:
+            # remove Unique Column Constraints
+            for constraint in expression.find_all(exp.UniqueColumnConstraint):
+                if constraint.parent:
+                    constraint.parent.pop()
+            return super().create_sql(expression)
```
GeorgeSittas, line 488: This looks iffy, we already have an entry for `exp.Create` in `TRANSFORMS`.. I'd rather remove that entry and have this method return `create_with_partitions_sql(...)` instead.

GeorgeSittas (commented):
I tried some DDLs in Spark to check this, and it seems like `UNIQUE` is rejected. Please, for future PRs / issues link any necessary docs / sources because this fact-checking is time-consuming.

Thanks for the PR in any case.

sqlglot/dialects/hive.py
GeorgeSittas: Spark2 also overrides `create_sql` to return `create_with_partitions_sql`, should we instead fall back to Hive's `create_sql` method?

dmoore247: Ok, makes sense to return `create_with_partitions_sql`

dmoore247:
@GeorgeSittas it's a challenge to find docs on features that are not supported. :)

Would including errors and error messages be useful documentation for PRs?
Would opening issues here to document be helpful, or just include with the PR?

GeorgeSittas:
> @GeorgeSittas it's a challenge to find docs on features that are not supported. :)
> 
> Would including errors and error messages be useful documentation for PRs? Would opening issues here to document be helpful, or just include with the PR?

Yep, any testing / errors encountered in the corresponding environments definitely help if there are no available references to check that a feature is unsupported. I get where you're coming from, it's just that we're going to double-check these things anyway, so any additional info is super helpful.

sqlglot/dialects/hive.py
dmoore247: My first attempt to return `create_with_partitions_sql(...)` results in an infinite loop. I'll look at this more closely later in the day, though at first glance looks like `generator:create_with_partitions_sql(...)` calls `create_sql`

sqlglot/dialects/hive.py
```diff
@@ -51,6 +51,13 @@
 DIFF_MONTH_SWITCH = ("YEAR", "QUARTER", "MONTH")
 
 
+def _create_sql(self, expression: exp.Create) -> str:
+    # remove Unique Column Constraints
+    for constraint in expression.find_all(exp.UniqueColumnConstraint):
+        if constraint.parent:
+            constraint.parent.pop()
+    return create_with_partitions_sql(self, expression)
```
GeorgeSittas, line 59: Can we just move this method inside Hive's `Generator` class? Also, can we merge its logic with Spark2's `_create_sql`? If that's possible then Spark will pick that up through inheritance.

sqlglot/dialects/hive.py
```diff
@@ -429,7 +436,8 @@ class Generator(generator.Generator):
             if e.args.get("allow_null")
             else "NOT NULL",
             exp.VarMap: var_map_sql,
-            exp.Create: create_with_partitions_sql,
+            #exp.Create: create_with_partitions_sql,
+            exp.Create: _create_sql,
```
GeorgeSittas, line 440: This won't be needed if `_create_sql` becomes a `Generator` method.

GeorgeSittas (commented):
Looks good, couple of comments. Need to run `make style` but looks ok I think.

sqlglot/dialects/hive.py
dmoore247: @GeorgeSittas I wanted to ask you this same question...
[Hive supports CREATE TEMPORARY TABLE AS SELECT](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/using-hiveql/content/hive_create_a_hive_temporary_table.html) ... so moving the spark2 temporary table logic to hive would undo that.

I would be fine with moving the UNIQUE logic to spark2 and not worrying about that for Hive at this time. Thoughts?

GeorgeSittas: Ok no problem leaving Spark2 `_create_sql` as is then.

> I would be fine with moving the UNIQUE logic to spark2 and not worrying about that for Hive at this time. Thoughts?

Nah let's just keep it in Hive. Can we instead return `super(Hive.Generator, self).create_sql(expression)` for Spark2 in the fallback case where we don't find a temporary property? That way we'll reuse Hive's logic, but we need to first move `_create_sql` in Hive's generator and remove the `_` prefix to associate it with `exp.Create`.

sqlglot/dialects/hive.py
dmoore247: To override `create_sql` in Hive Generator class, the execution flow will go 
```
hive.py:create_sql() -> return create_with_partitions_sql ->
dialect.py:create_with_partitions_sql -> self.create_sql() ->
hive.py:create_sql() -> ...
```

GeorgeSittas: Hm I see, makes sense. Can probably leave it as is then

sqlglot/dialects/hive.py
```diff
@@ -51,6 +51,14 @@
 DIFF_MONTH_SWITCH = ("YEAR", "QUARTER", "MONTH")
 
 
+def _create_sql(self, expression: exp.Create) -> str:
+    # remove Unique Column Constraints
+    for constraint in expression.find_all(exp.UniqueColumnConstraint):
+        if constraint.parent:
+            constraint.parent.pop()
```
GeorgeSittas, line 58: We need to copy `expression` before popping here because this is a mutation

dmoore247:
@GeorgeSittas 
This creates an infinite loop
```python
hive.py:
        def create_sql(self, expression: exp.Create) -> str:
            # remove Unique Column Constraints
            for constraint in expression.find_all(exp.UniqueColumnConstraint):
                if constraint.parent:
                    constraint.parent.pop()
            return create_with_partitions_sql(self, expression)
```

GeorgeSittas:
> @GeorgeSittas This creates an infinite loop
> 
> ```python
> hive.py:
>         def create_sql(self, expression: exp.Create) -> str:
>             # remove Unique Column Constraints
>             expression = expression.copy()
>             for constraint in expression.find_all(exp.UniqueColumnConstraint):
>                 if constraint.parent:
>                     constraint.parent.pop()
>             return create_with_partitions_sql(self, expression)
> ```

Yeah because `create_with_partitions_sql` ends up returning `self.create_sql(expression)` like you mentioned in an earlier followup. Maybe the way you have it now is ok. What's causing these CI failures?

dmoore247:
I'm working on the CI features... my earlier PR 'USING PARQUET` is messing with this PR. I'll need a little bit of time to fix.

dmoore247:
We had overrides on _create_sql in hive,spark2,spark and databricks. This last change consolidated all of them into a hive.py:_create_sql. There is one Generator method to add temp table storage provider that was unavoidable ([Hive does not need a storage provider ](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/using-hiveql/content/hive_create_a_hive_temporary_table.html))

sqlglot/dialects/hive.py
```diff
@@ -480,6 +506,10 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        def temporary_storage_provider(self, expression):
+            # Hive has no temporary storage provider (there are hive settings though)
+            return expression
```
GeorgeSittas, line 511: Let's add type annotations here

sqlglot/dialects/spark2.py
```diff
@@ -236,6 +218,13 @@ class Generator(Hive.Generator):
         WRAP_DERIVED_VALUES = False
         CREATE_FUNCTION_RETURN_AS = False
 
+        def temporary_storage_provider(self, expression):
```
GeorgeSittas, line 221: Ditto

GeorgeSittas (approved):

GeorgeSittas:
Thanks for the quick iterations on this, @dmoore247!

sqlglot/dialects/spark2.py
dmoore247: of course, my bad.

sqlglot/dialects/hive.py
```diff
@@ -480,6 +506,10 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        def temporary_storage_provider(self, expression: exp.Create):
```
GeorgeSittas, line 509:
```suggestion
        def temporary_storage_provider(self, expression: exp.Create) -> exp.Create:
```

sqlglot/dialects/spark2.py
```diff
@@ -236,6 +218,13 @@ class Generator(Hive.Generator):
         WRAP_DERIVED_VALUES = False
         CREATE_FUNCTION_RETURN_AS = False
 
+        def temporary_storage_provider(self, expression: exp.Create):
```
GeorgeSittas, line 221:
```suggestion
        def temporary_storage_provider(self, expression: exp.Create) -> exp.Create:
```

dmoore247: Thanks!

PR #2329, opened 2023-09-26
fix(mysql): transpile CHAR

barakalon:
Not sure the best way to approach this one.

Sometimes its only `CHR`: 
https://www.postgresql.org/docs/9.1/functions-string.html

Sometimes its either `CHR` or `CHAR`:
https://docs.databricks.com/en/sql/language-manual/functions/chr.html

Sometimes its only `CHAR`:
https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_char

And in the case of MySQL, it supports a few extra arguments. I guess we could make this a separate expression type... but that doesn't feel quite right, since in the single argument form (the most standard), they're the same.


sqlglot/expressions.py
```diff
@@ -4293,6 +4293,12 @@ class Coalesce(Func):
     _sql_names = ["COALESCE", "IFNULL", "NVL"]
 
 
+class Chr(Func):
```
tobymao, line 4296: default should be Char

tobymao: nvm i guess if postgres and databricks like chr, i'm ok with it

tobymao (approved):

sqlglot/dialects/mysql.py
```diff
@@ -717,3 +730,9 @@ def _oldstyle_limit_sql(self, expression: exp.Show) -> str:
                 limit_offset = f"{offset}, {limit}" if offset else limit
                 return f" LIMIT {limit_offset}"
             return ""
+
+        def chr_sql(self, expression: exp.Chr) -> str:
+            this = self.expressions(sqls=[expression.this] + expression.expressions)
```
GeorgeSittas, line 735: Isn't sqls` supposed to be a list of strings?

tobymao: i think the type hint is wrong in expressions, it can be exp or str

Issue #2323, opened 2023-09-26
ParseError raised when clickhouse SQL include sample clause


yagggi:
When parsing clickhouse SQL with `sample` clause, there's an error raised.

**Fully reproducible code snippet**

```
import sqlglot

sql = """
SELECT sum(PageViews * _sample_factor)
FROM visits
SAMPLE 10000000
"""

print(sqlglot.parse_one(sql, read="clickhouse"))
```
```
ParseError: Invalid expression / Unexpected token. Line 3, Col: 15.
  SELECT sum(PageViews * _sample_factor)
FROM visits
SAMPLE 10000000
```

**Official Documentation**
[clickhouse sample clause](https://clickhouse.com/docs/en/sql-reference/statements/select/sample)

yagggi:
Hi @GeorgeSittas , thanks for your quick reaction, but I noticed that there is a weird behavior, a redundant parenthesis is required otherwise `ParseError` will be raised. Please see below code snippet for reference:

```
In [64]: sqlglot.__version__
Out[64]: '18.8.0'

In [68]: sql = """
    ...: SELECT * FROM (SELECT a from b SAMPLE 0.01)
    ...: """

In [69]: print(sqlglot.parse_one(sql, read="clickhouse"))
...
ParseError: Expecting ). Line 2, Col: 43.

In [70]: sql = """
    ...: SELECT * FROM (SELECT a from b SAMPLE 0.01)) /* <-- here is the required redundant ')' */
    ...: """

In [71]: print(sqlglot.parse_one(sql, read="clickhouse"))
SELECT * FROM (SELECT a FROM b SAMPLE (0.01))
```

GeorgeSittas:
@yagggi thanks, seems like an oversight. Fixed in the linked PR

yagggi:
Have tested the latest version and `SAMPLE` works fine, thanks.

PR #2332, opened 2023-09-27
Fix(clickhouse): don't generate parentheses, match R_PAREN conditionally

GeorgeSittas:
Addresses https://github.com/tobymao/sqlglot/issues/2323#issuecomment-1736991605 and fixes another generation bug.

References:
- https://clickhouse.com/docs/en/sql-reference/statements/select/sample
- https://play.clickhouse.com/play?user=play#U0VMRUNUIHN1bShmb28gKiBiYXIpIEZST00gYmxhIFNBTVBMRSAoMTAwMDAwMDAp

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2734,14 +2734,14 @@ class Parser(metaclass=_Parser):
         )
         method = self._parse_var(tokens=(TokenType.ROW,))
 
-        self._match(TokenType.L_PAREN)
+        matched_l_paren = self._match(TokenType.L_PAREN)
 
         if self.TABLESAMPLE_CSV:
             num = None
             expressions = self._parse_csv(self._parse_primary)
         else:
             expressions = None
-            num = self._parse_primary()
+            num = self._parse_term()
 
         if self._match_text_seq("BUCKET"):
             bucket_numerator = self._parse_number()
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -2734,14 +2734,18 @@ def _parse_table_sample(self, as_modifier: bool = False) -> t.Optional[exp.Table
...
         if self.TABLESAMPLE_CSV:
             num = None
             expressions = self._parse_csv(self._parse_primary)
         else:
             expressions = None
-            num = self._parse_primary()
+            num = (
+                self._parse_factor()
+                if self._match(TokenType.NUMBER, advance=False)
+                else self._parse_primary()
```
GeorgeSittas, line 2747: This turned out a bit awkward.. the motivation was to support stuff like `SAMPLE 1 / 10` (shown as an example in ClickHouse's docs). The thing is that `_parse_term` wouldn't suffice because DuckDB also allows `SAMPLE 10%`, so the parser would try to parse this into a `Mod` and fail due to not finding a RHS. We also need to ensure we're calling `_parse_factor` against a number because otherwise it might consume `BUCKET`.

tobymao (approved):

PR #2334, opened 2023-09-27
fix(mysql,optimizer): TO_DAYS transpilation and more date casting

barakalon:


tests/dialects/test_mysql.py
```diff
@@ -506,6 +506,17 @@ def test_mysql_time(self):
         self.validate_identity("FROM_UNIXTIME(a, b)")
         self.validate_identity("FROM_UNIXTIME(a, b, c)")
         self.validate_identity("TIME_STR_TO_UNIX(x)", "UNIX_TIMESTAMP(x)")
+        self.validate_all(
+            "SELECT TO_DAYS('2007-10-07')",
+            write={
+                "mysql": "SELECT TO_DAYS('2007-10-07')",
+                "presto": "SELECT DATE_DIFF('DAY', CAST('0000-01-01' AS DATE), '2007-10-07')",
```
GeorgeSittas, line 513: Is `'2007-10-07'` a valid value for the third argument? Trino seems to expect a `DATE` here:

```
trino> SELECT DATE_DIFF('DAY', CAST('0000-01-01' AS DATE), '2007-10-07');
Query 20230927_183328_00002_4y9yd failed: line 1:8: Unexpected parameters (varchar(3), date, varchar(10)) for function date_diff. Expected: date_diff(varchar(x), date, date), date_diff(varchar(x), timestamp(p), timestamp(p)), date_diff(varchar(x), timestamp(p) with time zone, timestamp(p) with time zone), date_diff(varchar(x), time(p), time(p)), date_diff(varchar(x), time(p) with time zone, time(p) with time zone)
SELECT DATE_DIFF('DAY', CAST('0000-01-01' AS DATE), '2007-10-07')

trino> SELECT DATE_DIFF('DAY', CAST('0000-01-01' AS DATE), CAST('2007-10-07' AS DATE));
 _col0  
--------
 733321 
(1 row)

Query 20230927_183408_00003_4y9yd, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
0.54 [0 rows, 0B] [0 rows/s, 0B/s]
```

GeorgeSittas: Perhaps MySQL needs to parse `TO_DAYS`'s argument into a `TsOrDsToDate`?

barakalon: Yes, that's right.
I think this test just isn't right.

I added the type casting here:
https://github.com/tobymao/sqlglot/pull/2334/files#diff-d19f6f1f7cc671a042d531f94b85a280e7b62a4b43b1acc1d2b57270d03fbdafR357

I'm not sure of a better place to do it.

GeorgeSittas: That cast doesn't apply to `expression.this` though. If you pass in a string, the transform will use that in `DateDiff`. What I meant is to add an entry in `MySQL.Parser.FUNCTIONS` so that we wrap the argument of `TO_DAYS` with a `TsOrDsToDate`. Then the generation should just work for Presto.

barakalon: Ah, sorry. Wrong link.
https://github.com/tobymao/sqlglot/pull/2334/files#diff-28f5a8e638b2f114cfd7411b4589283d786380e7f1cc040fd990c8290887b2dcR62

We apply the cast to expression.this.

If we cast during parsing, then we must do this for _all_ date functions. Since, for example, MySQL almost always allows strings for date arguments, and Presto almost always doesnt.

That could lead to a lot of redundant casting.

GeorgeSittas: >  I think this test just isn't right.

It's just that the mysql -> presto transpilation wasn't working properly due to the cast being omitted; with the suggested change above I think you'll be able to add that test back again but with the cast added

GeorgeSittas: > If we cast during parsing, then we must do this for all date functions. Since, for example, MySQL almost always allows strings for date arguments, and Presto almost always doesnt.

Isn't the point to do that everywhere so that the transpilation can work without the optimizer though? For example we parse several date expressions in redshift like that.

barakalon: I need to essentially the same thing for everything here: https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html

Many of these are super common date functions. What's different is the _types_ that are allowed.

If we push type casting to the parser, then all users will likely end up with lots of redundant casts. For example:
`SELECT YEAR(CAST('2021-01-01' AS DATE))` -> `SELECT YEAR(CAST(CAST('2021-01-01' AS DATE) AS DATE))`

And this will apply to _all_ dialects, not just mysql.

We can move the casting to the generator, in which case this redundancy will only be the case for certain dialects (e.g. presto).

Or, we can keep it in the optimizer, in which case we can be smart about when to add the cast, but transpiling won't automatically work in some cases.

The optimizer feels best to me, but maybe that's because I'm already using it.

GeorgeSittas: I get where you're coming from. Maybe another idea is to both handle this at parse time and try to reduce the redundant casts at generation time by applying the `remove_redundant_casts` rule after running the type annotations. Although, that does add some more overhead to the generator.

@tobymao what do you think?

sqlglot/dialects/presto.py
```diff
@@ -354,6 +354,7 @@ class Generator(generator.Generator):
             exp.TimeStrToUnix: lambda self, e: f"TO_UNIXTIME(DATE_PARSE({self.sql(e, 'this')}, {Presto.TIME_FORMAT}))",
             exp.TimeToStr: lambda self, e: f"DATE_FORMAT({self.sql(e, 'this')}, {self.format_time(e)})",
             exp.TimeToUnix: rename_func("TO_UNIXTIME"),
+            exp.ToDays: transforms.preprocess([transforms.todays_to_datediff]),
```
tobymao, line 357: in general, let's avoid transforms except for big heavy machinary like SELECT transforms.

for this one, let's just add a dialect generator function

sqlglot/transforms.py
```diff
@@ -377,6 +377,16 @@ def eliminate_semi_and_anti_joins(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def todays_to_datediff(expression: exp.Expression) -> exp.Expression:
```
tobymao, line 380: move this to dialects

sqlglot/transforms.py
```diff
@@ -377,6 +377,16 @@ def eliminate_semi_and_anti_joins(expression: exp.Expression) -> exp.Expression:
...
+    if isinstance(expression, exp.ToDays):
+        return exp.DateDiff(
+            this=expression.this,
+            expression=exp.cast(exp.Literal.string("0000-01-01"), to=exp.DataType.Type.DATE),
```
tobymao, line 384: this doesn't seem right, should it be epoch?

barakalon: Nope. lol

barakalon: > Given a date date, returns a day number (the number of days since year 0). Returns NULL if date is NULL.
> 
> mysql> SELECT TO_DAYS(950501);
>        -> 728779
> mysql> SELECT TO_DAYS('2007-10-07');
>         -> 733321

sqlglot/dialects/presto.py
barakalon: roger

sqlglot/transforms.py
tobymao: if this is mysql specific, you could just do this the transform at parse time, so to_days turns into a date diff, and then you won't have to transpile it in every other dialect

tests/dialects/test_mysql.py
tobymao: i wouldn't worry about redundant casts for now, but for this specific case, i think to_days is mysql specific enough that it may make sense to do this in the generator

barakalon: moving this to tsordstodate during parsing...

sqlglot/dialects/dialect.py
```diff
@@ -797,3 +797,13 @@ def move_insert_cte_sql(self: Generator, expression: exp.Insert) -> str:
         expression = expression.copy()
         expression.set("with", expression.expression.args["with"].pop())
     return self.insert_sql(expression)
+
+
+def to_days_to_datediff_sql(self: Generator, expression: exp.ToDays) -> str:
+    return self.sql(
+        exp.DateDiff(
+            this=expression.this,
```
GeorgeSittas, line 805: There's a mutation going on here due to the `expression.this` having its `parent` changed.

tobymao: the approach is changing

barakalon:
Ok, I switched it to TsOrDsToDate.

This isn't perfect. There are still lots of functions lefts. For example, `EXTRACT`, which needs something like `TsOrDsToDatetime`.

sqlglot/optimizer/simplify.py
```diff
@@ -854,15 +846,27 @@ def cast_value(value: t.Any, to: exp.DataType) -> t.Optional[t.Union[datetime.da
     return None
 
 
-def extract_date(cast: exp.Cast) -> t.Optional[t.Union[datetime.date, datetime.date]]:
+def extract_date(cast: exp.Expression) -> t.Optional[t.Union[datetime.date, datetime.date]]:
+    to: exp.DataType
```
GeorgeSittas, line 850: Is this needed? We can also move it inline in one of the following assignments

sqlglot/optimizer/simplify.py
```diff
@@ -854,15 +846,27 @@ def cast_value(value: t.Any, to: exp.DataType) -> t.Optional[t.Union[datetime.da
...
+    if isinstance(cast, exp.Cast):
+        to = cast.to
+    elif isinstance(cast, exp.TsOrDsToDate):
+        to = exp.DataType.build(exp.DataType.Type.DATE)
```
GeorgeSittas, line 854: Can also do `build("date)"`

sqlglot/optimizer/simplify.py
```diff
@@ -878,7 +882,9 @@ def extract_interval(expression):
 def date_literal(date):
     return exp.cast(
         exp.Literal.string(date),
-        "DATETIME" if isinstance(date, datetime.datetime) else "DATE",
+        exp.DataType.Type.DATETIME
+        if isinstance(date, datetime.datetime)
+        else exp.DataType.Type.DATE,
```
GeorgeSittas, line 887: Ditto about `exp.DataType.Type.DATETIME` -> `"datetime"` I guess

sqlglot/optimizer/simplify.py
barakalon: Why? that seems slower, given it requires parsing. And in this case, we know what type we want.

GeorgeSittas: Fair enough, I just feel part of that API was to reduce the visual noise of writing the fqn of enums, but don't feel strongly about it

sqlglot/optimizer/simplify.py
barakalon: whoops, you're right

tests/fixtures/optimizer/simplify.sql
```diff
@@ -740,6 +743,9 @@ x >= CAST('2022-01-01' AS DATE);
 DATE_TRUNC('year', x) > CAST('2021-01-02' AS DATE);
 x >= CAST('2022-01-01' AS DATE);
 
+DATE_TRUNC('year', x) > TS_OR_DS_TO_DATE(TS_OR_DS_TO_DATE('2021-01-02'));
+x >= CAST('2022-01-01' AS DATE);
```
GeorgeSittas, line 747: Nice one!

sqlglot/dialects/mysql.py
```diff
@@ -86,15 +86,24 @@ def _trim_sql(self: MySQL.Generator, expression: exp.Trim) -> str:
...
         this = self.sql(expression, "this")
         unit = expression.text("unit").upper() or "DAY"
         return f"DATE_{kind}({this}, {self.sql(exp.Interval(this=expression.expression.copy(), unit=unit))})"
 
     return func
 
 
+def _ts_or_ds_to_date_sql(self: MySQL.Generator, expression: exp.TsOrDsToDate) -> str:
+    time_format = expression.args.get("format")
+    if time_format:
```
tobymao, line 102: need to copy expression here due to a mutation

tests/dialects/test_mysql.py
```diff
@@ -506,6 +506,33 @@ def test_mysql_time(self):
...
+                "presto": "SELECT (DATE_DIFF('DAY', CAST(CAST('0000-01-01' AS TIMESTAMP) AS DATE), CAST(CAST(x AS TIMESTAMP) AS DATE)) + 1)",
+            },
+        )
+        self.validate_all(
+            "SELECT DATEDIFF(x, y)",
+            write={"mysql": "SELECT DATEDIFF(x, y)", "presto": "SELECT DATE_DIFF('day', y, x)"},
+        )
+        self.validate_all(
+            "DAYOFYEAR(x)",
+            write={"mysql": "DAYOFYEAR(DATE(x))"},
```
tobymao, line 522: if you wanted to make this identity, you could remove the date right?

GeorgeSittas: This is a side effect of representing `DAYOFYEAR(x)` using a `TsOrDsToDate` wrapper for `x` right?

tobymao: yea but in the gen of all these functions, we can check for tsords

barakalon: ok, I gave that a shot

tobymao (approved):

Issue #2325, opened 2023-09-26
escaped `\0` characters in clickhouse are interpreted as null bytes


cpcloud:
**Fully reproducible code snippet**

```
In [17]: import sqlglot as sg

In [18]: expr = sg.parse_one("select '\\0'").sql("clickhouse")

In [19]: expr
Out[19]: "SELECT '\\0'"

In [20]: print(expr)
SELECT '\0'

In [21]: !clickhouse local <<< "$expr"
\0

In [22]: !clickhouse local <<< "SELECT ASCII('\0')"
0

In [23]: !clickhouse local <<< "SELECT ASCII('\1')"
92

In [24]: !clickhouse local <<< "SELECT ASCII('\x01')"
1
```

I think because `\\0` in Python means "the character `\` followed by the character `0`" and not "the null byte", sqlglot should backslash-escape `\0` strings.

Note that it seems to be specific to `\0` and not other backslashed characters:

```
In [29]: !clickhouse local <<< "SELECT '\0'"
\0

In [30]: !clickhouse local <<< "SELECT '\1'"
\\1
```

**Official Documentation**

https://clickhouse.com/docs/en/sql-reference/syntax#string

GeorgeSittas:
Should it be `expr = sg.parse_one("select '\0'").sql("clickhouse")` instead of including a double escape?

cpcloud:
`\0` is a null byte, so I don't think so:

```
>>> '\0'
'\x00'
>>> ord('\0')
0
```

GeorgeSittas:
What's the intent here though? What are you trying to transpile (read, write, etc)?

cpcloud:
I'm trying to generate code for regular expression replacement with ClickHouse and I want the meaning of escaped strings to match that of Python:

Escaped:

```
albatross :) select REGEXP_REPLACE('ab', '.', '\\0\\0'); -- works because of the escape

SELECT REGEXP_REPLACE('ab', '.', '\\0\\0')

Query id: 8c549122-cca0-4dd4-a923-0b8ad85ab7a8

┌─replaceRegexpAll('ab', '.', '\\0\\0')─┐
│ aabb                                  │
└───────────────────────────────────────┘

1 row in set. Elapsed: 0.001 sec.
```

Unescaped:

```
albatross :) select REGEXP_REPLACE('ab', '.', '\0\0'); -- fails because \0 means null byte

SELECT REGEXP_REPLACE('ab', '.', '\0\0')

[...]

PR #2336, opened 2023-09-27
Fix!: revert escape sequence changes introduced in #2230

GeorgeSittas:
This PR reverts https://github.com/tobymao/sqlglot/commit/66aadfc60b9bffe914b89043b293a3e7357d5dde#, because my understanding of the intended "escaping" semantics at the time was incorrect.

Phillip's original BQ query in #2225 was `SELECT '\a\b\f\n\r\t\v'`, i.e. this is what one would type verbatim in BQ's editor to run it against the engine. However, the python string `"SELECT '\a\b\f\n\r\t\v'"` that was fed into `transpile` is not the "correct" representation of the above query, because it's not what one would get if they were to read that query from a file, i.e. the source of truth for what the SQL means, in python:

```
➜ cat test.sql
SELECT '\a\b\f\n\r\t\v'
➜ python
Python 3.10.13 (main, Aug 24 2023, 22:36:46) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> open('test.sql', 'r').read()
"SELECT '\\a\\b\\f\\n\\r\\t\\v'\n"
```

Feeding that last python string to `transpile`, we get back the original BQ query, so the roundtrip of 1) storing the query in a file 2) reading that query from the file into its python string representation and 3) generating SQL back for that same dialect yields the exact same query, which is the intended behavior of SQLGlot:

```
>>> import sqlglot
>>> sqlglot.transpile("SELECT '\\a\\b\\f\\n\\r\\t\\v'\n", "bigquery")
["SELECT '\\a\\b\\f\\n\\r\\t\\v'"]
>>> print(sqlglot.transpile("SELECT '\\a\\b\\f\\n\\r\\t\\v'\n", "bigquery")[0])
SELECT '\a\b\f\n\r\t\v'
```

See the discussion in https://github.com/tobymao/sqlglot/issues/2325 for more context. 

cc: @cpcloud

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -333,8 +333,6 @@ class Generator:
         exp.Paren,
     )
 
-    UNESCAPED_SEQUENCE_TABLE = None  # type: ignore
-
     SENTINEL_LINE_BREAK = "__SQLGLOT__LB__"
 
     # Autofilled
@@ -347,6 +345,7 @@ class Generator:
     STRICT_STRING_CONCAT = False
     NORMALIZE_FUNCTIONS: bool | str = "upper"
     NULL_ORDERING = "nulls_are_small"
+    ESCAPE_LINE_BREAK = False
 
     can_identify: t.Callable[[str, str | bool], bool]
 
@@ -1659,8 +1658,8 @@ class Generator:
 
     def escape_str(self, text: str) -> str:
         text = text.replace(self.QUOTE_END, self._escaped_quote_end)
-        if self.UNESCAPED_SEQUENCE_TABLE:
-            text = text.translate(self.UNESCAPED_SEQUENCE_TABLE)
+        if self.ESCAPE_LINE_BREAK:
+            text = text.replace("\n", "\\n")
         elif self.pretty:
             text = text.replace("\n", self.SENTINEL_LINE_BREAK)
         return text
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -417,6 +417,7 @@ class BigQuery(Dialect):
         TABLE_HINTS = False
         LIMIT_FETCH = "LIMIT"
         RENAME_TABLE_WITH_DB = False
+        ESCAPE_LINE_BREAK = True
         NVL2_SUPPORTED = False
         UNNEST_WITH_ORDINALITY = False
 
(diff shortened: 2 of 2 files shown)


sqlglot/generator.py
```diff
@@ -347,6 +345,7 @@ class Generator:
     STRICT_STRING_CONCAT = False
     NORMALIZE_FUNCTIONS: bool | str = "upper"
     NULL_ORDERING = "nulls_are_small"
+    ESCAPE_LINE_BREAK = False
```
cpcloud, line 348: Interesting. Didn't see this before 😄

cpcloud (approved):

sqlglot/generator.py
GeorgeSittas: Yep - Toby added this a while ago because BQ doesn't allow actual newlines in single-quoted strings, i.e.

```sql
SELECT '
'
```

So if we ever encounter such a string, e.g. in Postgres, and want to transpile it to BQ we make sure to escape it:

```python
>>> import sqlglot
>>> sqlglot.transpile("""'
... '""", "postgres", "bigquery")
["'\\n'"]
```

Postgres:
```
georgesittas=# select '
georgesittas'# ';
 ?column?
----------
         +

(1 row)
```

BQ:
Screenshot 2023-09-28 at 1 48 02 AM

cpcloud: This feels like it has the reproducibility problem if I then read the bigquery string in and pass `read="postgres"`, no?

tobymao (approved):

sqlglot/generator.py
tobymao: if you read it from a file, a true line break is \n, and what you see in the bq ui would actually be \\n

GeorgeSittas: What do you mean?

cpcloud: Read in a bigquery select query with bigquery newline from a file, parse it as postgres. Is that the same result as passing in an ASCII 10 character, as you specified in your example?

GeorgeSittas: > Read in a bigquery select query with bigquery newline from a file, parse it as postgres.

I'm not sure if that premise makes sense? Why would you parse a BigQuery SQL file using Postgres? That's like feeding possibly invalid SQL into `parse_one` and expecting back a correct AST, right?

cpcloud: Sorry, what I mean is: parse it as bigquery, transpile it to postgres. Are those equivalent when executed?

GeorgeSittas: Hmm I see what you mean, yea in that case the transpilation breaks because we generate `SELECt '\\n'` - not equivalent. I'll think about this one tomorrow, might be that the transpilation towards postgres is broken because it doesn't treat the backslash as an escape.

PR #2358, opened 2023-10-02
Fix(hive): don't generate BYTE when transpiling Oracle's VARCHAR(5 BYTE)

### LINKED ISSUE #2356: Handle BYTE statement in Oracle DDL to Spark
using sqlglot to transpile some Oracle DDL to Spark that I have from a customer.  The modified code changed the `VARCHAR2(5 BYTE)` to `VACHAR(5 BYTE)` but the DDL still fails in Spark with the `(5 BYTE)`.   Once removing the word BYTE the DDL executed fine.

GeorgeSittas:
Fixes #2356

This shouldn't be breaking, I remember adding support for the `expression` argument only to handle Oracle's `BYTE`, `CHAR` type specifiers, see https://github.com/tobymao/sqlglot/commit/3ee16c5e8

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -562,18 +562,13 @@ class Hive(Dialect):
                 expression = exp.DataType.build("text")
             elif expression.this in exp.DataType.TEMPORAL_TYPES:
                 expression = exp.DataType.build(expression.this)
-
-            size_expression = expression.find(exp.DataTypeParam)
-            if size_expression:
-                if expression.is_type("float"):
+            elif expression.is_type("float"):
+                size_expression = expression.find(exp.DataTypeParam)
+                if size_expression:
                     size = int(size_expression.name)
                     expression = (
                         exp.DataType.build("float") if size <= 32 else exp.DataType.build("double")
                     )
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/hive.py
```diff
@@ -562,13 +562,18 @@ def datatype_sql(self, expression: exp.DataType) -> str:
...
+            if size_expression:
+                if expression.is_type("float"):
                     size = int(size_expression.name)
                     expression = (
                         exp.DataType.build("float") if size <= 32 else exp.DataType.build("double")
                     )
+                elif size_expression.expression:
+                    # This removes keywords like `BYTE` (Oracle) in `CAST(x AS VARCHAR(5 BYTE))`
+                    expression = expression.copy()
+                    t.cast(exp.DataTypeParam, expression.find(exp.DataTypeParam)).expression.pop()
```
GeorgeSittas, line 576: On second thought, maybe we should control this using a generator flag instead so that we only generate these specifiers for Oracle?

tobymao: sure

GeorgeSittas: Done

PR #2359, opened 2023-10-02
Feat(snowflake): add support for staged file file_format clause

GeorgeSittas:
References:
- https://docs.snowflake.com/en/user-guide/querying-stage
- https://docs.snowflake.com/en/sql-reference/constructs/from#syntax

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -344,7 +344,16 @@ class Snowflake(Dialect):
                 table = self._parse_string()
 
             if table:
-                return self.expression(exp.Table, this=table)
+                file_format = None
+                pattern = None
+
+                if self._match_text_seq("(", "FILE_FORMAT", "=>"):
+                    file_format = self._parse_string() or super()._parse_table_parts()
+                    if self._match_text_seq(",", "PATTERN", "=>"):
+                        pattern = self._parse_string()
+                    self._match_r_paren()
+
+                return self.expression(exp.Table, this=table, format=file_format, pattern=pattern)
 
             return super()._parse_table_parts(schema=schema)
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -344,7 +344,16 @@ def _parse_table_parts(self, schema: bool = False) -> exp.Table:
                 table = self._parse_string()
 
             if table:
-                return self.expression(exp.Table, this=table)
+                file_format = None
+                pattern = None
+
+                if self._match_text_seq("(", "FILE_FORMAT", "=>"):
```
tobymao, line 350: does this syntax exist outside of this file format stuff? is it a generic kwarg for snowfake?

GeorgeSittas: Good call, I'm not sure but it seems like we have another couple of examples that look similar. It seems like this syntax is used only in a few clauses, though.

tobymao: can you parse a wrapped csv of kwargs?

GeorgeSittas: We need lookahead to ensure there's a `=>` here because you may get a table name followed by a table alias with named columns, as in

```sql
SELECT * FROM @foo t(bla)  -- valid snowflake
```

The `t(bla)` part will be consumed by `self._parse_wrapped_csv(self._parse_lambda)` as a function call instead of a table alias, so we'd need to be a little more smart than that here. Do you see any other simple way to reuse it?

I'm personally fine with this approach for now, if it's a common snowflake pattern for kwargs (i.e. we get another similar issue) we can refactor this into a helper?

tobymao (approved):

PR #2360, opened 2023-10-02
fix(mysql): DATE_ADD for datetimes

barakalon:
The recent logic I added for transpiling DATE_ADD in MySQL was wrong.

MySQL's DATE_ADD semantics are different than Hive's: https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-add

I'm not sure of a good way to handle this at the parser/generator level without a bunch of new TsOrDs* functions. So I added it to the optimizer for now.


tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
```diff
@@ -9775,7 +9775,7 @@ JOIN "date_dim" AS "d1"
   ON "catalog_sales"."cs_sold_date_sk" = "d1"."d_date_sk"
   AND "d1"."d_week_seq" = "d2"."d_week_seq"
   AND "d1"."d_year" = 2002
-  AND "d3"."d_date" > CONCAT("d1"."d_date", INTERVAL '5' day)
+  AND "d3"."d_date" > "d1"."d_date" + INTERVAL '5' day
```
barakalon, line 9778: Seems like this change caught a bug.

sqlglot/expressions.py
```diff
@@ -234,6 +234,9 @@ def type(self) -> t.Optional[DataType]:
 
     @type.setter
     def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
+        self.set_type(dtype)
+
+    def set_type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
```
GeorgeSittas, line 239: Why was this needed?

sqlglot/expressions.py
```diff
@@ -4053,6 +4056,16 @@ def unit(self) -> t.Optional[Var]:
...
 
 
+class IntervalOp(TimeUnit):
+    arg_types = {"unit": True, "expression": True}
+
+    def interval(self):
+        return Interval(
+            this=self.expression.copy(),
+            unit=self.unit.copy(),
+        )
```
GeorgeSittas, line 4066: Should we make this a property and call it `to_interval` or something?

sqlglot/optimizer/annotate_types.py
```diff
@@ -421,3 +484,14 @@ def _annotate_by_args(
             )
 
         return expression
+
+    def _annotate_dateadd(self, expression: exp.IntervalOp) -> exp.IntervalOp:
+        self._annotate_args(expression)
+
+        if expression.this.type.is_type(*exp.DataType.TEXT_TYPES):
```
GeorgeSittas, line 491: Should we instead phrase this as `expression.this.type in exp.DataType.TEXT_TYPES` to avoid unpacking?

tests/test_optimizer.py
```diff
@@ -546,6 +546,52 @@ def test_binary_annotation(self):
...
+                "SELECT CAST('2023-01-01' AS DATETIME) + INTERVAL '1' HOUR",
+            ),
+            (
+                "SELECT '2023-01-01 00:00:01' + INTERVAL '1' HOUR",
+                exp.DataType.Type.DATETIME,
+                "SELECT CAST('2023-01-01 00:00:01' AS DATETIME) + INTERVAL '1' HOUR",
+            ),
+            ("SELECT 'nonsense' + INTERVAL '1' DAY", exp.DataType.Type.UNKNOWN),
+            ("SELECT x.a + INTERVAL '1' DAY FROM x AS x", exp.DataType.Type.DATE),
+            ("SELECT x.a + INTERVAL '1' HOUR FROM x AS x", exp.DataType.Type.DATE),
```
GeorgeSittas, line 574: Why is this a date? This doesn't seem to match MySQL's behavior, for example:

```
mysql> select cast('2020-01-01' as date) + interval '1' hour;
+------------------------------------------------+
| cast('2020-01-01' as date) + interval '1' hour |
+------------------------------------------------+
| 2020-01-01 01:00:00                            |
+------------------------------------------------+
1 row in set (0.00 sec)
```

GeorgeSittas (commented):
Nice work @barakalon, these date/time optimizations look really interesting. I wonder if they are valid in all supported dialects though? Date/time stuff can vary wildly...

Left a few comments to consider.

sqlglot/expressions.py
barakalon: mypy doesn't like
```python
expression.set = DataType.Type.Foo
```
Seems to be using the type of the getter, not the setter?

sqlglot/expressions.py
barakalon: I'll call it to_interval

sqlglot/optimizer/annotate_types.py
barakalon: ah sure

tests/test_optimizer.py
barakalon: Ah, you're right. This should be a datetime. good catch

barakalon:
Yeah, I also am unsure of all the different nuances of different dialects.
I guess it should be easy to make configurable though, given the implementation with `BINARY_COERCIONS`?

sqlglot/expressions.py
tobymao: just do type ignore then

sqlglot/expressions.py
barakalon: Hm, `to_interval` isn't quite right, because it doesn't operate on the entire expression. Just the interval part.
And making it a property doesn't feel right either, since this creates a new instance.

GeorgeSittas: I see what you mean, feel free to ignore this then.

GeorgeSittas (approved):

PR #2361, opened 2023-10-02
Feat: switch identifier normalization off using comments

GeorgeSittas:


tobymao:
not sure if this change is correct, i think it should just be done in the optimizer, then you won't have to check the parent right?

GeorgeSittas:
Not sure what you mean, why wouldn't we need to check the parent? The reasoning for doing that here is that a comment next to an identifier may be attached to a higher-level AST node like a `Column` or a `Table`, so we need to make sure to check if that's the case.

tobymao:
we normalize identifiers through optimizer/normalize_identifiers. we call transform, simply check the meta to stop traversal into the node. so when the column is checked, we stopped traversing, it will simplify this change

GeorgeSittas:
> we normalize identifiers through optimizer/normalize_identifiers. we call transform, simply check the meta to stop traversal into the node. so when the column is checked, we stopped traversing, it will simplify this change

Done

tobymao (approved):

PR #2367, opened 2023-10-03
Fix: unescape escape sequences on read, re-escape them on generation

### LINKED ISSUE #2325: escaped `\0` characters in clickhouse are interpreted as null bytes

GeorgeSittas:
Fixes #2325 & addresses https://github.com/tobymao/sqlglot/pull/2336#discussion_r1339336075

cc @cpcloud


sqlglot/generator.py
```diff
@@ -1670,8 +1670,8 @@ def literal_sql(self, expression: exp.Literal) -> str:
 
     def escape_str(self, text: str) -> str:
         text = text.replace(self.QUOTE_END, self._escaped_quote_end)
-        if self.ESCAPE_LINE_BREAK:
-            text = text.replace("\n", "\\n")
+        if self.INVERSE_ESCAPE_SEQUENCES:
+            text = "".join(self.INVERSE_ESCAPE_SEQUENCES.get(ch, ch) for ch in text)
```
GeorgeSittas, line 1674: FYI I didn't use `str.translate` because it's not suited for a 1 -> many translation (i.e. `\n` to `\\n`).

sqlglot/tokens.py
```diff
@@ -1203,6 +1204,13 @@ def _extract_string(self, delimiter: str, escapes=None) -> str:
                 if self._end:
                     raise TokenError(f"Missing {delimiter} from {self._line}:{self._start}")
 
+                if self.ESCAPE_SEQUENCES and self._peek and self._char in self.STRING_ESCAPES:
+                    escaped_sequence = self.ESCAPE_SEQUENCES.get(self._char + self._peek)
+                    if escaped_sequence:
+                        self._advance(2)
```
tobymao, line 1210: should this be len(escaped_sequence)

tobymao (approved):

sqlglot/tokens.py
GeorgeSittas: It should be incorrect here because otherwise it'd be`len('\n') == 1`; the `2` comes from the `self._char + self._peek` part, i.e. skip these two characters that were condensed into one.

This logic doesn't take into account escape sequences of different lengths, so we'd need to do another refactor to support them if any dialect has them

tobymao: ok

tests/dialects/test_clickhouse.py
```diff
@@ -85,6 +85,16 @@ def test_clickhouse(self):
             "CREATE MATERIALIZED VIEW test_view (id UInt8) TO db.table1 AS SELECT * FROM test_data"
         )
 
+        self.validate_all(
+            "SELECT '\\0'",
+            read={
+                "": "SELECT '\0'",
```
GeorgeSittas, line 91:
```suggestion
                "mysql": "SELECT '\0'",
```

PR #2368, opened 2023-10-03
Fix(optimizer): handle edge case in DATE_TRUNC simplification

GeorgeSittas:


sqlglot/optimizer/simplify.py
```diff
@@ -728,7 +728,7 @@ def simplify_datetrunc_predicate(expression: exp.Expression) -> exp.Expression:
         l = expression.this
         rs = expression.expressions
 
-        if all(_is_datetrunc_predicate(l, r) for r in rs):
+        if rs and all(_is_datetrunc_predicate(l, r) for r in rs):
```
GeorgeSittas, line 731: For the added test case `rs` is `[]` and so the `all` predicate is trivially true. This was leading to the following line and hence causing an `AttributeError`, as demonstrated below:

```python
>>> import sqlglot
>>> simplify(sqlglot.parse_one("concat(a, b) in (select * from foo where cond)"))
  File "sqlglot/optimizer/simplify.py", line 89, in simplify
    expression = while_changing(expression, _simplify)
  File "sqlglot/helper.py", line 216, in while_changing
    expression = func(expression)
  File "sqlglot/optimizer/simplify.py", line 82, in _simplify
    node = simplify_datetrunc_predicate(node)
  File "sqlglot/optimizer/simplify.py", line 100, in wrapped
    return func(expression, *args, **kwargs)
  File "sqlglot/optimizer/simplify.py", line 732, in simplify_datetrunc_predicate
    unit = l.unit.name.lower()
AttributeError: 'SafeConcat' object has no attribute 'unit'

tobymao (approved):

PR #2371, opened 2023-10-04
fix(postgres): translate variance_pop to var_pop and variance to var_samp

cpcloud:
Fixes a bug in transalation of population variance and also translates sample variance into `var_samp` which is compatible with datafusion postgres.


GeorgeSittas (approved):
LGTM

PR #2373, opened 2023-10-04
Feat(redshift): add support for Redshift's super array index iteration

### LINKED ISSUE #2372: The Redshift dialect does not support iterations on an array with: x AS y AT z

- dialect: redshift

Code:
```python
import sqlglot
sql_code = """
SELECT c_name,
       orders.o_orderkey AS orderkey,
       index AS orderkey_index
FROM customer_orders_lineitem c, c.c_orders AS orders AT index 
ORDER BY orderkey_index;
"""
expressions = sqlglot.parse(sql_code, read="redshift")
```

Output:
```console
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 5, Col: 56.
  y AS orderkey,
       index AS orderkey_index
FROM customer_orders_lineitem c, c.c_orders AS orders AT index 
ORDER BY orderkey_index;
```

Same for query:
```sql
SELECT attr as attribute_name, json_typeof(val) as value_type 
FROM customer_orders_lineitem c, UNPIVOT c.c_orders[0] AS val AT attr 
WHERE c_custkey = 9451;
```

**Official Documentation**

https://docs.aws.amazon.com/redshift/latest/dg/query-super.html#unnest

GeorgeSittas:
Fixes #2372

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2465,6 +2465,7 @@ class Table(Expression):
         "version": False,
         "format": False,
         "pattern": False,
+        "index": False,
     }
 
     @property
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1350,13 +1350,17 @@ class Generator:
         pivots = f" {pivots}" if pivots else ""
         joins = self.expressions(expression, key="joins", sep="", skip_first=True)
         laterals = self.expressions(expression, key="laterals", sep="")
+
         file_format = self.sql(expression, "format")
         if file_format:
             pattern = self.sql(expression, "pattern")
             pattern = f", PATTERN => {pattern}" if pattern else ""
             file_format = f" (FILE_FORMAT => {file_format}{pattern})"
 
-        return f"{table}{version}{file_format}{alias}{hints}{pivots}{joins}{laterals}"
+        index = self.sql(expression, "index")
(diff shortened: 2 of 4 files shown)


GeorgeSittas:
I'm still taking a look at the `UNPIVOT` test case in the issue, it's an exception to Redshift's `UNPIVOT` syntax and needs a bit more work to fix..

GeorgeSittas:
> I'm still taking a look at the `UNPIVOT` test case in the issue, it's an exception to Redshift's `UNPIVOT` syntax and needs a bit more work to fix..

Ok took a stab at this, should be ready for a review now.

tobymao (approved):

sqlglot/dialects/redshift.py
```diff
@@ -31,6 +31,7 @@ class Redshift(Postgres):
     RESOLVES_IDENTIFIERS_AS_UPPERCASE = None
 
     SUPPORTS_USER_DEFINED_TYPES = False
+    INDEX_OFFSET = 0
```
GeorgeSittas, line 34: FYI I tested this in Redshift directly and it can also be seen in their [docs](https://docs.aws.amazon.com/redshift/latest/dg/query-super.html) too:

Screenshot 2023-10-04 at 6 58 37 PM

PR #2380, opened 2023-10-05
Feat(bigquery): improve support for CREATE MODEL DDL statement

GeorgeSittas:
References:
- https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-create#create_model_syntax
- https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-create-remote-model#create_model_syntax


sqlglot/parser.py
```diff
@@ -236,6 +236,7 @@ class Parser(metaclass=_Parser):
         TokenType.SCHEMA,
         TokenType.TABLE,
         TokenType.VIEW,
+        TokenType.MODEL,
```
tobymao, line 239: does model need to be added to id_var?

GeorgeSittas: it's included indirectly because every `CREATABLE` is a valid identifier

tobymao (approved):

PR #2381, opened 2023-10-06
Feat(schema): add method to check if column exists

GeorgeSittas:
Thought something like this might be an improvement to the API

### DIFF
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -88,6 +88,27 @@ class Schema(abc.ABC):
             The resulting column type.
         """
 
+    def has_column(
+        self,
+        table: exp.Table | str,
+        column: exp.Column,
+        dialect: DialectType = None,
+        normalize: t.Optional[bool] = None,
+    ) -> bool:
+        """
+        Returns whether or not `column` appears in `table`'s schema.
+
+        Args:
+            table: the source table.
+            column: the target column.
+            dialect: the SQL dialect that will be used to parse `table` if it's a string.
+            normalize: whether to normalize identifiers according to the dialect of interest.
+
+        Returns:
+            True if the column appears in the schema, False otherwise.
+        """
+        return column.name in self.column_names(table, dialect=dialect, normalize=normalize)
+
     @property
     @abc.abstractmethod
     def supported_table_args(self) -> t.Tuple[str, ...]:
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -308,6 +308,22 @@ class MappingSchema(AbstractMappingSchema[t.Dict[str, str]], Schema):
 
         return exp.DataType.build("unknown")
 
+    def has_column(
+        self,
+        table: exp.Table | str,
+        column: exp.Column,
+        dialect: DialectType = None,
+        normalize: t.Optional[bool] = None,
+    ) -> bool:
+        normalized_table = self._normalize_table(table, dialect=dialect, normalize=normalize)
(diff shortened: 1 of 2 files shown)


barakalon (commented):
I'm assuming this is intended to be used somewhere in sqlmesh?

If this is intended to be used anywhere within the optimizer, it needs to be on the abstract `Schema` class.

sqlglot/schema.py
```diff
@@ -308,6 +308,22 @@ def get_column_type(
 
         return exp.DataType.build("unknown")
 
+    def has_column(
```
tobymao, line 311: why don't you define __contains__ instead?

tobymao: you'd have to pass in an ast node or parse it

GeorgeSittas: chatted in Slack - this doesn't seem as flexible because the schema kicks off normalization automatically and so the burden of normalizing a table / column would be transferred to the user

barakalon:
I guess we can't just do `"y" in s.column_names("x")` because of normalization?

And still, if you think this will be a useful utility, we probably want it on the Schema interface. Unless you enforce a MappingSchema is used in sqlmesh?

GeorgeSittas:
> I guess we can't just do "y" in s.column_names("x") because of normalization?

I think so, because the schema may or may not contain normalized names (not necessarily related to internal schema `normalize` attribute), which is why I started off with this interface instead of `__contains__`, i.e. to allow for extra flexibility.

> And still, if you think this will be a useful utility, we probably want it on the Schema interface. Unless you enforce a MappingSchema is used in sqlmesh?

Happy to add this to the `Schema` interface, it seems like a general-purpose schema utility anyway. The only reason I didn't do it from the start was to avoid breaking custom schema implementations in client code, since no one's currently using this interface anyway.

barakalon:
Maybe we should ditch ABC in favor of regular ol' `raise NotImplementedError`. Or perhaps we can just add a simple, default implementation that doesn't normalize and uses `column_names`.

sqlglot/schema.py
```diff
@@ -308,6 +308,22 @@ def get_column_type(
...
+        self,
+        table: exp.Table | str,
```
tobymao, line 313: what if there's a catalog and db?

GeorgeSittas: they will be included in the AST node for the table - either the supplied one or the one that will be parsed from the `str` value

tobymao (approved):

tobymao:
what about @barakalon comment on schema vs mapping schema?

GeorgeSittas:
> what about @barakalon comment on schema vs mapping schema?

Just added a simple implementation in the base class here https://github.com/tobymao/sqlglot/pull/2381/commits/c92017cb7d73bcb979ad5d5c05898496e73e9032, @barakalon is this what you had in mind?

sqlglot/schema.py
```diff
@@ -71,7 +71,7 @@ def column_names(
     def get_column_type(
         self,
         table: exp.Table | str,
-        column: exp.Column,
+        column: exp.Column | str,
```
GeorgeSittas, line 74: The previous type hint was more restrictive than needed, we already supported strings in the implementation, e.g. as shown in this test: https://github.com/tobymao/sqlglot/blob/main/tests/test_schema.py#L167

barakalon (approved):
perfect :+1:

PR #2383, opened 2023-10-07
Feat(optimizer): simplify CONCAT_WS

GeorgeSittas:
This PR:

- Fixes a typo in the optimizer (~which is a breaking change, I guess~): compliments -> complements 😄 
- Improves `simplify_concat` by extending it to also handle `CONCAT_WS`

Regarding the second point, a core assumption here is that the separator always comes first in the argument list of `CONCAT_WS`. This is the case in several dialects I checked (*), and is also reinforced by the tests we have. In any case, I think the optimization is valid and if there's ever an exception to this then it should be probably handled at parse time in order to facilitate transpilation too.

(*): Postgres, MySQL, DuckDB, Spark, Presto, Trino, Snowflake, T-SQL

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -617,7 +617,6 @@ def simplify_concat(expression):
         return expression
 
     if isinstance(expression, exp.ConcatWs):
-        # Skip the separator value
         sep_expr, *expressions = expression.expressions
         sep = sep_expr.name
         concat_type = exp.ConcatWs
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -619,10 +619,8 @@ def simplify_concat(expression):
     if isinstance(expression, exp.ConcatWs):
         # Skip the separator value
         sep_expr, *expressions = expression.expressions
-
-        # CONCAT_WS is equivalent to CONCAT when the separator is the empty string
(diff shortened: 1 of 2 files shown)


GeorgeSittas:
> In any case, I think the optimization is valid and if there's ever an exception to this then it should be probably handled at parse time in order to facilitate transpilation too.

On the other hand... maybe we could just get rid of `CONCAT_WS` at parse time altogether, e.g. by parsing

```sql
CONCAT_WS(sep, a, b, ...)
```

as

```sql
CONCAT(a, sep, b, sep, ...)
```

The advantages of this would be to 1) get rid of this optimization since we won't have to deal with `ConcatWs` at all, 2) get rid of `exp.ConcatWs` and 3) seamlessly transpile `CONCAT_WS` to dialects that don't support it (e.g. BigQuery, Redshift, SQLite to name a few). What do people think?

sqlglot/optimizer/simplify.py
```diff
@@ -287,19 +287,19 @@ def _simplify_comparison(expression, left, right, or_=False):
     return None
 
 
-def remove_compliments(expression, root=True):
+def remove_complements(expression, root=True):
```
tobymao, line 290: these are not public apis so aren’t breaking

tobymao (approved):

PR #2385, opened 2023-10-08
feat(databricks): transpile UPDATE FROM to MERGE INTO

dmoore247:
`UPDATE .. FROM` sql semantics is used to make a corelated update to a target table, where new values in the target are a function of values in a second table. TSQL

TSQL, PostgreSQL, Redshift, Snowflake do support `UPDATE .. FROM` sql syntax.
Databricks, DuckDB, Oracle don't support `UPDATE .. FROM` sql syntax but do support `MERGE INTO`.

This PR builds a `transforms.py:update_from` (`exp.Update`) transformer, and tests for and implements for the Databricks SQL dialect.

[Slack discussion:](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1695848302882859)

**Official Documentation References**
Additional References:
[Databricks MERGE INTO](https://docs.databricks.com/en/sql/language-manual/delta-merge-into.html)
[Databricks Update:](https://docs.databricks.com/en/sql/language-manual/delta-update.html)
[Oracle Update:](https://docs.oracle.com/javadb/10.8.3.0/ref/rrefsqlj26498.html)
[Oracle Merge:](https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/MERGE.html)
[Redshift UPDATE:](https://docs.aws.amazon.com/redshift/latest/dg/r_UPDATE.html)
[DuckDB UPDATE:](https://duckdb.org/docs/sql/statements/update.html)

TSQL syntax checking example:
image

Databricks syntax checking example:
image

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -9,6 +9,43 @@ if t.TYPE_CHECKING:
     from sqlglot.generator import Generator
 
 
+def update_from(expression: exp.Expression) -> exp.Expression:
+    """Transform UPDATE FROM to MERGE INTO"""
+    from_expression = expression.find(exp.From)
+    if isinstance(expression, exp.Update) and from_expression:
+        # perform MERGE INTO
+        """
+        class Update(Expression):
+            arg_types = {
+                    "with": False,
+                    "this": False,
+                    "expressions": True,
+                    "from": False,
+                    "where": False,
+                    "returning": False,
+                    "order": False,
+                    "limit": False,
+                }
+        class Merge(Expression):
+            arg_types = {"this": True, "using": True, "on": True, "expressions": True}
+        """
+        join = from_expression.find(exp.Join)
+        if join:
+            on = join.args.get("on")
+            table = join.find(exp.Table)
+            this = expression.this
+            expression.find(exp.Comment)
+
+            # {"matched": True, "source": False, "condition": False, "then": True}
+            # WHEN MATCHED THEN UPDATE SET state = z.state
+            then = exp.Update(expressions=expression)
+            when = exp.When(matched=True, then=then)
+
+            expression = exp.Merge(this=this, using=table, on=on, expressions=[when])
+
+    return expression
+
+
 def unalias_group(expression: exp.Expression) -> exp.Expression:
     """
     Replace references to select aliases in GROUP BY clauses.
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -13,35 +13,19 @@ def update_from(expression: exp.Expression) -> exp.Expression:
     """Transform UPDATE FROM to MERGE INTO"""
     from_expression = expression.find(exp.From)
     if isinstance(expression, exp.Update) and from_expression:
-        # perform MERGE INTO
-        """
-        class Update(Expression):
-            arg_types = {
-                    "with": False,
-                    "this": False,
-                    "expressions": True,
-                    "from": False,
-                    "where": False,
-                    "returning": False,
-                    "order": False,
-                    "limit": False,
-                }
-        class Merge(Expression):
-            arg_types = {"this": True, "using": True, "on": True, "expressions": True}
-        """
         join = from_expression.find(exp.Join)
         if join:
             on = join.args.get("on")
             table = join.find(exp.Table)
             this = expression.this
-            expression.find(exp.Comment)
-
-            # {"matched": True, "source": False, "condition": False, "then": True}
-            # WHEN MATCHED THEN UPDATE SET state = z.state
-            then = exp.Update(expressions=expression)
+            then = exp.Update(expressions=expression.expressions)
             when = exp.When(matched=True, then=then)
-
             expression = exp.Merge(this=this, using=table, on=on, expressions=[when])
+        else:
+            table = from_expression.find(exp.Table)
+            expression = exp.Update(
+                this=table, expressions=expression.expressions, where=expression.args["where"]
+            )
 
     return expression
 
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -49,6 +49,7 @@ class Databricks(Spark):
                 ]
             ),
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
+            exp.Update: transforms.preprocess([transforms.update_from]),
         }
 
         def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
--- a/tests/dialects/test_databricks.py
+++ b/tests/dialects/test_databricks.py
@@ -190,3 +190,12 @@ class TestDatabricks(Validator):
                 "databricks": "WITH x AS (SELECT 1) SELECT * FROM x",
             },
         )
+
+    def test_update_from(self):
+        self.validate_all(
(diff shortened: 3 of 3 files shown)


sqlglot/transforms.py
```diff
@@ -9,6 +9,27 @@
     from sqlglot.generator import Generator
 
 
+def update_from(expression: exp.Expression) -> exp.Expression:
```
GeorgeSittas, line 12: We should rename this to reflect the intent, i.e. something like `update_from_to_merge_into`.

sqlglot/transforms.py
```diff
@@ -9,6 +9,27 @@
...
+    """Transform UPDATE FROM to MERGE INTO"""
+    from_expression = expression.find(exp.From)
+    if isinstance(expression, exp.Update) and from_expression:
+        join = from_expression.find(exp.Join)
+        if join:
+            on = join.args.get("on")
+            table = join.find(exp.Table)
```
GeorgeSittas, line 19: Why do we need `find` here?

```suggestion
            table = join.this
```

sqlglot/transforms.py
```diff
@@ -9,6 +9,27 @@
...
+            this = expression.this
+            then = exp.Update(expressions=expression.expressions)
+            when = exp.When(matched=True, then=then)
+            expression = exp.Merge(this=this, using=table, on=on, expressions=[when])
+        else:
+            table = from_expression.find(exp.Table)
```
GeorgeSittas, line 25: Same here, should we instead do `from_expression.this`?

sqlglot/transforms.py
```diff
@@ -9,6 +9,27 @@
...
+            expression = exp.Update(
+                this=table, expressions=expression.expressions, where=expression.args["where"]
```
GeorgeSittas, line 27: >  where=expression.args["where"]

What if there's no `where` arg in the `Update`? Not sure why that is the case, but it's not a required arg so if it's missing this will raise a key error.

sqlglot/transforms.py
```diff
@@ -9,6 +9,27 @@
     from sqlglot.generator import Generator
 
 
+def update_from(expression: exp.Expression) -> exp.Expression:
+    """Transform UPDATE FROM to MERGE INTO"""
+    from_expression = expression.find(exp.From)
+    if isinstance(expression, exp.Update) and from_expression:
+        join = from_expression.find(exp.Join)
+        if join:
```
GeorgeSittas, line 17: This looks incomplete, can't we have more than one joins? What happens in that case?

GeorgeSittas (commented):
Thanks for the PR Douglas, left some comments.

We should also add this transformation for Oracle if it doesn't support the UPDATE-FROM syntax.

sqlglot/transforms.py
dmoore247: Let me check this out.

sqlglot/transforms.py
dmoore247: Let me check this out

sqlglot/transforms.py
dmoore247: Is there a better technique to get or return None if missing? 
There are lots of patterns in the code base, I get lost sometimes.

sqlglot/transforms.py
dmoore247: If any join then follow that path. 
It is probably worth building more complex test cases to explore these cases.

dmoore247:
@GeorgeSittas Thanks for the review!
[Hive now has `MERGE` ](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/using-hiveql/content/hive_merge_data_in_hive_tables.html)
[OSS Spark (3)] [Has `MERGE` if the storage format is Delta](https://stackoverflow.com/questions/46613907/does-apache-spark-sql-support-merge-clause)

I don't have Hive/Oracle environments to test with, I can add those, but only go off of the specs if that makes sense to you.

sqlglot/transforms.py
GeorgeSittas: I think `where=expression.args.get("where")` would suffice here.

sqlglot/transforms.py
GeorgeSittas: > If any join then follow that path.

You would ignore the rest of the joins though, since you only deal with the first one.

GeorgeSittas:
> @GeorgeSittas Thanks for the review! [Hive now has `MERGE` ](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/using-hiveql/content/hive_merge_data_in_hive_tables.html) [OSS Spark (3)] [Has `MERGE` if the storage format is Delta](https://stackoverflow.com/questions/46613907/does-apache-spark-sql-support-merge-clause)
> 
> I don't have Hive/Oracle environments to test with, I can add those, but only go off of the specs if that makes sense to you.

Either way Hive and Spark don't support the `UPDATE..FROM` syntax, correct?

cc @eakmanrq tagging since IIRC you've some experience with these statements.

dmoore247:
@GeorgeSittas Correct,

Hive does not have [UPDATE](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=82903069#LanguageManualDML-Update) FROM
Spark Delta / Spark Iceberg does not have UPDATE FROM

sqlglot/transforms.py
dmoore247: Found real-world example and tested. The generated join is far from complete. Will have to research.

dmoore247: The strategy is to convert the multiple joins into a sub-select...
```sql
-- tsql
UPDATE ru_new SET NPI = i.identifier
      FROM t_merge ru_new
      JOIN mytbl1 ui
        ON ru_new.user_id = ui.user_id
      JOIN mytbl2 i
        ON ui.identifier_id = i.id
      JOIN mytble3 it
        ON i.identifier_type = it.identifier_type
     WHERE it.short_description = 'NPI'
        and i.identifier like replicate('[0-9]', 10)

-- spark delta sql/databricks sql
MERGE INTO t_merge AS ru_new
USING (
    SELECT
        ui.user_id,
        i.identifier AS npi
    FROM mytbl1 ui
    JOIN mytbl2 i
        ON ui.identifier_id = i.id
    JOIN mytble3 it
        ON i.identifier_type = it.identifier_type
    WHERE it.short_description = 'NPI'
        AND i.identifier rLIKE '^[0-9]{10}$'
) AS src
ON ru_new.user_id = src.user_id
WHEN MATCHED THEN
    UPDATE SET ru_new.NPI = src.npi;
```

dmoore247: I'll have to do some more work, thanks for the catch :)

dmoore247: I've done the work for the test above. I haven't pushed it because it's ugly, lol.

GeorgeSittas: Can you share a link so that I can take a quick look and see if it's in the right direction?

dmoore247: @GeorgeSittas If I push, it goes directly into this PR.

dmoore247: @GeorgeSittas Here you go: https://gist.github.com/dmoore247/2d2feff1dd79c63d49fc00eb9892e1de

GeorgeSittas: Yeah that looks overly complex, can you briefly explain what you tried to implement?

GeorgeSittas:
@dmoore247 closing this PR for now as it's become inactive, let's continue chatting about how to complete it in Slack.

PR #2386, opened 2023-10-08
Feat(optimizer): propagate constants

GeorgeSittas:
Motivated by SQLite: https://www.sqlite.org/optoverview.html.

Perhaps we can treat casts as "constant expressions" and do the same for them, e.g.

```sql
x = CAST(5 AS BIGINT) AND y = x -- => x = CAST(5 AS BIGINT) AND y = CAST(5 AS BIGINT)
```

Screenshot 2023-10-08 at 8 45 43 PM

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -487,7 +487,7 @@ class Expression(metaclass=_Expression):
         """
         for node, _, _ in self.dfs(prune=lambda n, p, *_: p and not type(n) is self.__class__):
             if not type(node) is self.__class__:
-                yield node.unnest() if unnest else node
+                yield node.unnest() if unnest and not isinstance(node, Subquery) else node
 
     def __str__(self) -> str:
         return self.sql()
@@ -6388,27 +6388,6 @@ def func(name: str, *args, dialect: DialectType = None, **kwargs) -> Func:
         raise ValueError(error_message)
 
     return function
-
-
-def normalized(expression: Expression, dnf: bool = False) -> bool:
-    """
-    Checks whether a given expression is in a normal form of interest.
-
-    Example:
-        >>> normalized(maybe_parse("(a AND b) OR c OR (d AND e)"), dnf=True)
-        True
-        >>> normalized(maybe_parse("(a OR b) AND c"))
-        True
-        >>> normalized(maybe_parse("a AND (b OR c)"), dnf=True)
-        False
-
-    Args:
-        expression: The target expression.
-        dnf: Whether or not to check if the expression is in Disjunctive Normal Form (DNF).
-            Default: False, i.e. we check if it's in Conjunctive Normal Form (CNF).
-    """
-    ancestor, root = (And, Or) if dnf else (Or, And)
-    return not any(connector.find_ancestor(ancestor) for connector in expression.find_all(root))
 
 
 def true() -> Boolean:
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6390,6 +6390,27 @@ def func(name: str, *args, dialect: DialectType = None, **kwargs) -> Func:
     return function
 
 
+def normalized(expression: Expression, dnf: bool = False) -> bool:
+    """
+    Checks whether a given expression is in a normal form of interest.
+
+    Example:
+        >>> normalized(maybe_parse("(a AND b) OR c OR (d AND e)"), dnf=True)
+        True
+        >>> normalized(maybe_parse("(a OR b) AND c"))
+        True
+        >>> normalized(maybe_parse("a AND (b OR c)"), dnf=True)
+        False
+
+    Args:
+        expression: The target expression.
+        dnf: Whether or not to check if the expression is in Disjunctive Normal Form (DNF).
+            Default: False, i.e. we check if it's in Conjunctive Normal Form (CNF).
+    """
+    ancestor, root = (And, Or) if dnf else (Or, And)
+    return not any(connector.find_ancestor(ancestor) for connector in expression.find_all(root))
+
+
 def true() -> Boolean:
     """
     Returns a true Boolean expression.
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -372,7 +372,7 @@ def absorb_and_eliminate(expression, root=True):
 
 def propagate_constants(expression, root=True):
     """
-    Propagate constants for conjunctions normalized into DNF:
+    Propagate constants for conjunctions in DNF:
 
     SELECT * FROM t WHERE a = b AND b = 5 becomes
     SELECT * FROM t WHERE a = 5 AND b = 5
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -392,7 +392,7 @@ def propagate_constants(expression, root=True):
             l, r = eq.left, eq.right
 
             # TODO: create a helper that can be used to detect nested literal expressions such
-            # as CAST('2012-01-01' AS DATE), since we usually want to treat those as literals too
+            # as CAST(123456 AS BIGINT), since we usually want to treat those as literals too
             if isinstance(l, exp.Column) and isinstance(r, exp.Literal):
                 pass
             elif isinstance(r, exp.Column) and isinstance(l, exp.Literal):
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -401,16 +401,14 @@ def propagate_constants(expression, root=True):
 
         if constant_mapping:
             for column in expression.find_all(exp.Column):
-                id_and_constant = constant_mapping.get(column)
+                parent = column.parent
+                column_id, constant = constant_mapping.get(column) or (None, None)
                 if (
-                    id_and_constant
-                    and id(column) != id_and_constant[0]
-                    and not (
-                        isinstance(column.parent, exp.Is)
-                        and type(column.parent.expression) is exp.Null
-                    )
+                    column_id is not None
+                    and id(column) != column_id
+                    and not (isinstance(parent, exp.Is) and isinstance(parent.expression, exp.Null))
                 ):
-                    column.replace(id_and_constant[1].copy())
+                    column.replace(constant.copy())
 
     return expression
 
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -402,7 +402,9 @@ def propagate_constants(expression, root=True):
         if constant_mapping:
             for column in expression.find_all(exp.Column):
                 id_and_constant = constant_mapping.get(column)
-                if id_and_constant and id(column) != id_and_constant[0]:
+                if id_and_constant and id(column) != id_and_constant[0] and not (
+                    isinstance(column.parent, exp.Is) and type(column.parent.expression) is exp.Null
+                ):
                     column.replace(id_and_constant[1].copy())
 
     return expression
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -8,6 +8,7 @@ from decimal import Decimal
 from sqlglot import exp
 from sqlglot.generator import cached_generator
 from sqlglot.helper import first, merge_ranges, while_changing
+from sqlglot.optimizer.scope import find_all_in_scope
 
 # Final means that an expression should not be simplified
 FINAL = "final"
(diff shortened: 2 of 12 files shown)


tests/fixtures/optimizer/simplify.sql
```diff
@@ -625,22 +625,22 @@ t0.x = t1.x AND t0.y < t1.y AND t0.y <= t1.y;
...
 --------------------------------------
--- Coalesce
+-- COALESCE
 --------------------------------------
 COALESCE(x);
 x;
 
 COALESCE(x, 1) = 2;
-x = 2 AND NOT x IS NULL;
+x = 2;
```
GeorgeSittas, line 634: I just realized this is incorrect because if `x` is `NULL` it'll result in `NULL` whereas before it'd result in `FALSE`. I'll try to fix it soon.

GeorgeSittas: Tried to address this in https://github.com/tobymao/sqlglot/pull/2386/commits/b5d0c7f5e0fcaf6fece342c7cdbd201563184fb1, need to think about it a bit more though probably..

sqlglot/optimizer/simplify.py
```diff
@@ -369,6 +370,51 @@ def absorb_and_eliminate(expression, root=True):
...
+    if (
+        isinstance(expression, exp.And)
+        and (root or not expression.same_parent)
+        and normalized(expression, dnf=True)
+    ):
+        constant_mapping: t.Dict[exp.Column, t.Tuple[int, exp.Literal]] = {}
+        for eq in expression.find_all(exp.EQ):
+            l, r = eq.left, eq.right
+
+            if isinstance(l, exp.Column) and isinstance(r, exp.Literal):
```
barakalon, line 393: We should have some kind of helper for literals, since we often treat things like `CAST('2012-01-01' AS DATE)` as a literal. Maybe a problem for another day.

sqlglot/optimizer/simplify.py
```diff
@@ -369,6 +370,51 @@ def absorb_and_eliminate(expression, root=True):
...
+def propagate_constants(expression, root=True):
+    """
+    Propagate constants for conjunctions in DNF:
+
+    SELECT * FROM t WHERE a = b AND b = 5 becomes
+    SELECT * FROM t WHERE a = 5 AND b = 5
+
+    Reference: https://www.sqlite.org/optoverview.html
+    """
+    from sqlglot.optimizer.normalize import normalized
```
barakalon, line 382: Kind of a smell that we need a circular dependency here.
Maybe `normalized` should be in its own module, separate from `normalize`.

sqlglot/optimizer/simplify.py
```diff
@@ -369,6 +370,51 @@ def absorb_and_eliminate(expression, root=True):
...
+                pass
+            elif isinstance(r, exp.Column) and isinstance(l, exp.Literal):
+                l, r = r, l
+            else:
+                continue
+
+            constant_mapping[l] = (id(l), r)
+
+        if constant_mapping:
+            for column in expression.find_all(exp.Column):
```
barakalon, line 403: Be careful with `expression.find_all` in predicates - this will traverse into subqueries. I think you want [find_all_in_scope](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L777)

sqlglot/optimizer/simplify.py
```diff
@@ -369,6 +370,49 @@ def absorb_and_eliminate(expression, root=True):
...
+                parent = column.parent
+                column_id, constant = constant_mapping.get(column) or (None, None)
+                if (
+                    column_id is not None
+                    and id(column) != column_id
+                    and not (isinstance(parent, exp.Is) and isinstance(parent.expression, exp.Null))
+                ):
+                    column.replace(constant.copy())
```
barakalon, line 411: Is replacement right? That takes it a step further than the sqlite rule, which says it just adds the extra predicate.

Do any planners take advantage of column equality in join conditions?

sqlglot/optimizer/simplify.py
GeorgeSittas: Makes sense, was thinking about this too

sqlglot/optimizer/simplify.py
GeorgeSittas: Good catch, I'll make the change

sqlglot/optimizer/simplify.py
GeorgeSittas: Doesn't SQLite also replace the variable? Based on the example I thought it does

sqlglot/optimizer/simplify.py
GeorgeSittas: Fixed - added a test case for this too.

sqlglot/optimizer/simplify.py
GeorgeSittas: Yeah.. I tried to move `normalized` into `helper` at first but then I needed to do a local import because of another circular dependency with `from sqlglot import exp` so I just left it as shown here..

sqlglot/optimizer/simplify.py
barakalon: The example doesn’t show replacement. I also wonder if there is different behavior on join conditions.

GeorgeSittas: Yeah, will dive deeper. Also a good catch, thanks!

sqlglot/optimizer/simplify.py
GeorgeSittas: Moved `normalized` in `expressions.py` since it can also be viewed as a general-purpose "expression helper".

sqlglot/expressions.py
```diff
@@ -487,7 +487,7 @@ def flatten(self, unnest=True):
         """
         for node, _, _ in self.dfs(prune=lambda n, p, *_: p and not type(n) is self.__class__):
             if not type(node) is self.__class__:
-                yield node.unnest() if unnest else node
+                yield node.unnest() if unnest and not isinstance(node, Subquery) else node
```
GeorgeSittas, line 490: This seems like a bug: if one were to call `flatten` on e.g. a `Connector` with a `Subquery` child, the child would also be unnested.

```python
>>> list(parse_one("a and (select true from t where id = 1)").flatten())[1].sql()
'SELECT TRUE FROM t WHERE id = 1'
```

sqlglot/optimizer/simplify.py
GeorgeSittas: I found another subtle bug with `find_all` vs `find_all_in_scope`, see https://github.com/tobymao/sqlglot/pull/2386/files#diff-18e3fc3b5cad74391bd8f9f25228b86f02a446bbd9dbecbb3d9083e8c565c2fdR87. The corresponding test case is https://github.com/tobymao/sqlglot/pull/2386/files#diff-93a74220acfd05f3226eaa0df3befb5b86cd4c8bfed680556dbfd29ce23e75cfR904-R905

sqlglot/optimizer/simplify.py
GeorgeSittas: Ended up moving it back to `normalize.py` because of the bug I mentioned below - we needed `find_all_in_scope` which was also causing circular dependency issues... 🤦 

I ended up doing `import sqlglot` to refer to `normalized` using the full module path .

sqlglot/optimizer/simplify.py
```diff
@@ -369,6 +372,50 @@ def absorb_and_eliminate(expression, root=True):
...
+            # TODO: create a helper that can be used to detect nested literal expressions such
+            # as CAST(123456 AS BIGINT), since we usually want to treat those as literals too
+            if isinstance(l, exp.Column) and isinstance(r, exp.Literal):
+                pass
+            elif isinstance(r, exp.Column) and isinstance(l, exp.Literal):
+                l, r = r, l
+            else:
+                continue
+
+            constant_mapping[l] = (id(l), r)
```
tobymao, line 403: why not put id(l) in the hash? that's most expensive part

sqlglot/optimizer/simplify.py
GeorgeSittas: It seems that there are some gnarly cases that are tricky to take care of - mostly because of coercions. Interestingly, SQLite doesn't implement this transformation for `ON` clauses of `LEFT`, `RIGHT` joins:

```c
      ...
      /* Do not propagate constants on any ON clause if there is a
      ** RIGHT JOIN anywhere in the query */
      x.mExcludeOn = EP_InnerON | EP_OuterON;
    }else{
      /* Do not propagate constants through the ON clause of a LEFT JOIN */
      x.mExcludeOn = EP_OuterON;
      ...
```

Some documentation on SQLite's implementation can be found [here](https://github.com/tobymao/sqlglot/assets/46752250/e97b5a63-f0cb-46f2-a47d-3f69bfe58376) (screenshot of their [source code](https://sqlite.org/src/doc/trunk/README.md)).

On the other hand, MariaDB [mentions](https://mariadb.com/kb/en/equality-propagation-optimization/) this transformation in the context of `WHERE` clauses, but towards the end of the page they also have this (note the remark about `ON` expressions):

Screenshot 2023-10-09 at 5 46 56 AM

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
```diff
@@ -2029,18 +2029,7 @@ JOIN "date_dim" AS "date_dim"
...
-  AND "customer_demographics"."cd_education_status" = 'Secondary'
-  AND "customer_demographics"."cd_marital_status" = 'D'
-  AND "customer_demographics"."cd_marital_status" = 'M'
-  AND "customer_demographics"."cd_marital_status" = 'U'
-  AND "household_demographics"."hd_dep_count" = 1
-  AND "household_demographics"."hd_dep_count" = 3
-  AND "store_sales"."ss_hdemo_sk" = "household_demographics"."hd_demo_sk"
-  AND "store_sales"."ss_sales_price" <= 100.00
-  AND "store_sales"."ss_sales_price" >= 150.00
+  ON FALSE
```
tobymao, line 2032: this looks wrong, what is this?

sqlglot/optimizer/simplify.py
GeorgeSittas: Not fully sure what you mean, can you elaborate? The intent was to use the hash to find all the matching columns (i.e. different `Expression` objects) in the conjunction by doing a lookup, and `id(l)` needs to be part of the stored value because we need to avoid replacing the actual column participating in the equality with the constant in that equality (i.e. don't replace `x` with `10` in `x = 10`).

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
GeorgeSittas: Well I'm not sure if I'm missing some weird `NULL` shenanigans, but if you look at the `ON` clause it has the following conditions (prior to my changes):

```
  AND "customer_demographics"."cd_education_status" = 'Advanced Degree'
  AND "customer_demographics"."cd_education_status" = 'Primary'
```

These are contradicting, hence the `FALSE` value replacing the whole thing if I understand this correctly. Perhaps there was a bug prior to this PR?

sqlglot/optimizer/simplify.py
GeorgeSittas: Regarding

> The example doesn’t show replacement

It's now clear that SQLite does indeed replace the variable with the value:

Screenshot 2023-10-09 at 6 03 43 AM

tests/fixtures/optimizer/tpc-ds/tpc-ds.sql
GeorgeSittas: To be clear I'm still exploring `LEFT` and `RIGHT` joins in this PR, SQLite seems to avoid any transformations on the `ON` clause for them (see above discussion with Barak), so there's probably a catch.

tobymao: there was a previous bug and this exposes it, it should be fixed

but it's education_status = advanced degree or education_status = primary

GeorgeSittas: Fixed in https://github.com/tobymao/sqlglot/pull/2386/commits/d40fc163f593ff47a883428f9769a6e045776734

GeorgeSittas:
We discussed about this PR in Slack and decided to keep the constant propagation rule opt-in for now, as it's relatively unstable and needs some more investigation / polishing.

tobymao (approved):

PR #2390, opened 2023-10-09
Fix: correctly handle agg subexpressions with no selections

ginter:


sqlglot/optimizer/pushdown_projections.py
```diff
@@ -112,6 +115,9 @@ def _remove_unused_selections(scope, parent_selections, schema, alias_count):
                 star = True
             removed = True
 
+        if not is_agg and selection.find(exp.AggFunc):
```
barakalon, line 118: Will this also be the case in:
```
SELECT
  a
FROM x
GROUP BY a
```
?

ginter: no only if there's an agg func. added an additional test

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -9,7 +9,9 @@
 SELECT_ALL = object()
 
 # Selection to use if selection list is empty
-DEFAULT_SELECTION = lambda: alias("1", "_")
+DEFAULT_SELECTION = lambda is_agg: alias(
+    exp.Count(this=exp.Literal.number(1)) if is_agg else "1", "_"
```
tobymao, line 13: exp.Max may be better for intent?

tobymao (approved):

barakalon (approved):

PR #2392, opened 2023-10-10
Fix!(schema): don't preemptively uppercase data types

GeorgeSittas:


sqlglot/schema.py
```diff
@@ -326,7 +326,7 @@ def get_column_type(
             if isinstance(column_type, exp.DataType):
                 return column_type
             elif isinstance(column_type, str):
-                return self._to_data_type(column_type.upper(), dialect=dialect)
+                return self._to_data_type(column_type, dialect=dialect)
```
GeorgeSittas, line 329: We were previously forcing types to be uppercase here as a normalization step in order to look them up in the schema's `_type_mapping_cache`, but this is not correct in the case of nested types because it could also lead to changing the original identifiers used e.g. in structs.

PR #2394, opened 2023-10-10
Fix(schema): don't specialize type variable in MappingSchema

GeorgeSittas:
In SQLMesh we use the `MappingSchema` to store `DataType`s and it works just fine. Not sure why we restricted the type hint to just `str` for the types to begin with.

### DIFF
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -192,7 +192,7 @@ class AbstractMappingSchema(t.Generic[T]):
         )
 
 
-class MappingSchema(AbstractMappingSchema[t.Dict[str, str]], Schema):
+class MappingSchema(AbstractMappingSchema[t.Dict[str, str | exp.DataType]], Schema):
     """
     Schema based on a nested mapping.
 
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -192,7 +192,7 @@ class AbstractMappingSchema(t.Generic[T]):
         )
 
 
-class MappingSchema(AbstractMappingSchema[t.Dict[str, str | exp.DataType]], Schema):
+class MappingSchema(AbstractMappingSchema[t.Dict[str, t.Union[str, exp.DataType]]], Schema):
     """
     Schema based on a nested mapping.
 
--- a/sqlglot/schema.py
+++ b/sqlglot/schema.py
@@ -192,7 +192,7 @@ class AbstractMappingSchema(t.Generic[T]):
         )
 
 
-class MappingSchema(AbstractMappingSchema[t.Dict[str, t.Union[str, exp.DataType]]], Schema):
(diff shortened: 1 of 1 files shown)


GeorgeSittas:
Hmm I guess this is breaking because mypy may yell if a current user of `MappingSchema` calls a string method on a type, respecting the previous type hints, since the string method won't exist on the `DataType`.

tobymao (approved):

sqlglot/schema.py
```diff
@@ -192,7 +192,7 @@ def nested_get(
         )
 
 
-class MappingSchema(AbstractMappingSchema[t.Dict[str, str]], Schema):
+class MappingSchema(AbstractMappingSchema, Schema):
```
GeorgeSittas, line 195: cc @barakalon can you remind me why this was `t.Dict[str, str]` to begin with? It seems like it was first introdued in https://github.com/tobymao/sqlglot/commit/f50b2918b#diff-9e702532a0bf04ce1423d3d2160b3131a25d68ea2a07ddc8b1e4462d2c8da2f7R134 but I don't remember the context.

We clearly allow both `str` and `DataType` (e.g. we have code that considers these two cases for the columns' types, see `get_column_type`), so an alternative I considered was updating this type to be `t.Dict[str, str | DataType]`, but this would then be breaking and the user would have to explicitly cast the result if they wanted to use `AbstractMappingSchema.find`, which is annoying.

barakalon: Probably no good reason. Schema pre-exists a lot of maturation of DataType.

removing the type parameter implicitly makes this AbstractMappingSchema[Any]. Is that what we want?

GeorgeSittas: I'm not fully sure; my reasoning was that being more strict by specifying the type as a union of `str` and `DataType` would be more annoying for users because they'd have to account for both cases. I'm also not sure if allowing `str` column types instead of converting them to `DataType`s is helpful?

barakalon: This is the only place that type parameter is used: https://github.com/tobymao/sqlglot/blob/main/sqlglot/schema.py#L165

and it seems like that’s meant to be a protected method. So maybe just remove the generic altogether? Might not be worth it.

GeorgeSittas: Happy to remove it altogether too, I don't see much value in it if we don't specialize the type anyway.

> it seems like that’s meant to be a protected method

Not sure about this one though, e.g. we're using `find` in SQLMesh as a way to get the columns-to-types mapping of a given table.

barakalon: And do you coerce from string|DataType to DataType?

GeorgeSittas: IIRC we store the column types as `DataType` values initially so it just returns them later on

PR #2400, opened 2023-10-12
Feat(duckdb): add parsing support for timestamp types with unit suffixes

cpcloud:
Add support for parsing duckdb timestamp types with unit suffixes


sqlglot/dialects/duckdb.py
```diff
@@ -321,6 +325,10 @@ class Generator(generator.Generator):
             exp.DataType.Type.UINT: "UINTEGER",
             exp.DataType.Type.VARBINARY: "BLOB",
             exp.DataType.Type.VARCHAR: "TEXT",
+            exp.DataType.Type.TIMESTAMP_S: "TIMESTAMP_S",
+            exp.DataType.Type.TIMESTAMP_MS: "TIMESTAMP_MS",
+            exp.DataType.Type.TIMESTAMP_US: "TIMESTAMP_US",
```
cpcloud, line 330: We could potentially leave this out as parse target and just turn `TIMESTAMP_US` into `TIMESTAMP`, but I wasn't sure exactly how to do that.

GeorgeSittas (approved):

sqlglot/dialects/duckdb.py
cpcloud: Done!

cpcloud:
Ah right, thanks!

cpcloud:
Thanks for the quick review!

PR #2403, opened 2023-10-13
 Feat!(optimizer): simplify conditionals (IF, CASE)

GeorgeSittas:
Assumptions made in this PR:

1. `CASE x WHEN y THEN z ELSE w END` is simply syntactic sugar for `CASE WHEN x = y THEN z ELSE w END`
2. `CASE WHEN FALSE THEN x END` evaluates to `NULL` (tested with postgres, mysql, duckdb, sqlite, spark, bq, snowflake, redshift, trino)


tobymao:
how does it handle the equality of nulls?

GeorgeSittas:
> how does it handle the equality of nulls?

In general `CASE x WHEN NULL THEN 1 ELSE 2 END` results in `2` from what I've seen, and some engines like BQ suggest using the form `CASE WHEN <cond> ...` instead to deal with `NULL`s. Do you have a specific example in mind?

tobymao:
case null when null ...

GeorgeSittas:
> case null when null ...

```python
>>> from sqlglot.optimizer.canonicalize import canonicalize
>>>
>>> canonicalize(parse_one("case null when null then 1 else 2 end")).sql()
'CASE WHEN NULL = NULL THEN 1 ELSE 2 END'
>>>
>>> simplify(canonicalize(parse_one("case null when null then 1 else 2 end"))).sql()
'CASE WHEN NULL THEN 1 ELSE 2 END'
```

Postgres:
```
georgesittas=# select case null when null then 1 else 2 end;
 case
------
    2
(1 row)

georgesittas=# select CASE WHEN NULL = NULL THEN 1 ELSE 2 END;
 case
------
    2
(1 row)
```

tobymao:
some engines don't allow null = null

sqlglot/optimizer/simplify.py
```diff
@@ -695,6 +696,29 @@ def simplify_concat(expression):
...
 
 
+def simplify_conditionals(expression):
+    """Simplifies expressions like IF, CASE if their condition is statically known."""
+    if isinstance(expression, exp.Case):
+        this = expression.this
+        for case in expression.args["ifs"]:
+            cond = this.eq(case.this) if this else case.this
+
+            if always_true(cond):
```
tobymao, line 706: why do you need to do this? shouldn't the simplifier already handle this?

sqlglot/optimizer/canonicalize.py
```diff
@@ -83,6 +84,16 @@ def remove_ascending_order(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def expand_case_conditions(expression: exp.Expression) -> exp.Expression:
```
tobymao, line 87: do we need this? can this just go in simplify?

sqlglot/optimizer/simplify.py
GeorgeSittas: not really:

```python
>>> optimize("if(true, 1, 2)").sql()
'CASE WHEN TRUE THEN 1 ELSE 2 END'
```

tests/fixtures/optimizer/simplify.sql
```diff
@@ -911,13 +911,58 @@ t1.a = 39 AND t2.b = t1.a AND t3.c = t2.b;
...
+x;
+
+IF(FALSE, x, y);
+y;
+
+IF(FALSE, x);
+NULL;
+
+IF(NULL, x, y);
+CASE WHEN NULL THEN x ELSE y END;
```
tobymao, line 942: this seems suboptimal, this should always be y

tests/fixtures/optimizer/simplify.sql
```diff
@@ -911,13 +911,58 @@ t1.a = 39 AND t2.b = t1.a AND t3.c = t2.b;
...
+CASE WHEN TRUE THEN x ELSE y END;
+x;
+
+CASE WHEN FALSE THEN x ELSE y END;
+y;
+
+CASE WHEN FALSE THEN x WHEN FALSE THEN y WHEN TRUE THEN z END;
+z;
+
+-- Simplified to w if the canonicalize rule runs first
```
tobymao, line 956: canonicalization never runs before simplify, i think we should get rid of the canonicalizer

sqlglot/optimizer/canonicalize.py
GeorgeSittas: could be moved yeah, will do

sqlglot/optimizer/simplify.py
tobymao: you need to canocalizie first right?

GeorgeSittas:
> some engines don't allow null = null

FWIW this yields null in postgres, sqlite, mysql, duckdb, spark, trino, snowflake, redshift, clickhouse, tsql (couldn't test teradata, oracle, hive, drill, doris, etc).

It doesn't work in BigQuery, though: "Operands of = cannot be literal NULL", but perhaps this needs to be fixed in the generator anyway so as to generate `IS` instead of `=` if either operand is statically known to be `NULL`.

sqlglot/optimizer/simplify.py
GeorgeSittas: not sure what you mean - this is running every optimization, canonicalization included

tests/fixtures/optimizer/simplify.sql
GeorgeSittas: Most of the dialects I mentioned which support `null` as a condition value treat it as falsey, so maybe this could be done. I'm curious though about why we have `always_true` and we don't have `always_false` which takes nulls into account?

tests/fixtures/optimizer/simplify.sql
GeorgeSittas: Yeah not sure if there's much value to keeping them separated besides making the intent of these transformations more clear (as in: one "canonicalizes", the other actually "simplifies"). Another thing could be that canonicalizations require type inference but that could be easily checked before calling them to avoid that overhead.

tests/fixtures/optimizer/simplify.sql
tobymao: always_false should take nulls into account

sqlglot/optimizer/canonicalize.py
tobymao: yea, just do this in simplify

GeorgeSittas: Done in https://github.com/tobymao/sqlglot/pull/2403/commits/9a6481904aa1123a0d3c5ad71b0b2e47c883ae4b

GeorgeSittas: I also fixed a bug here: https://github.com/tobymao/sqlglot/pull/2403/commits/9a6481904aa1123a0d3c5ad71b0b2e47c883ae4b#diff-28f5a8e638b2f114cfd7411b4589283d786380e7f1cc040fd990c8290887b2dcR73

tests/fixtures/optimizer/simplify.sql
GeorgeSittas: Done https://github.com/tobymao/sqlglot/pull/2403/commits/a5ad243e0dba7434243bac0d8488be0fb801ed39

tests/fixtures/optimizer/simplify.sql
GeorgeSittas: Let's do this in a followup PR, cc @barakalon

barakalon: I guess canonicalize and simplify have become a dumping grounds for rules that operate on simple expressions, while most of the other rules tend to deal with SELECT.

I think merging them makes sense.

tobymao: sorry, i didn't mean get rid of canonicalizer, i meant move this rule from canocialize to simplify

tobymao (approved):

tests/fixtures/optimizer/simplify.sql
GeorgeSittas: Ah 😁

Is there a value to keeping the canonicalize rule around though? I was also thinking what Barak said.

tobymao: because you need types

GeorgeSittas: we can have a section in the simplifier where we check if there are types, and if so then apply the functions in canonicalize.py that need them (there are a couple that don't even need types)

tests/fixtures/optimizer/simplify.sql
```diff
@@ -911,13 +911,73 @@ t1.a = 39 AND t2.b = t1.a AND t3.c = t2.b;
...
+CASE WHEN FALSE THEN x ELSE y END;
+y;
+
+CASE WHEN FALSE THEN x WHEN FALSE THEN y WHEN TRUE THEN z END;
+z;
+
+CASE NULL WHEN NULL THEN x ELSE y END;
+y;
+
+CASE 4 WHEN 1 THEN x WHEN 2 THEN y WHEN 3 THEN z ELSE w END;
```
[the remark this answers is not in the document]
GeorgeSittas: Added

PR #2404, opened 2023-10-14
Fix: make `RegexpILike` a subclass of `Binary` as well as `Func`

samuelcolvin:
Maybe I'm missing something, but this looks like a simple omission to me - `RegexpLike` is a subclass of `Binary`, so I guess `RegexpILike` should be too.

Thanks so much for sqlglot by the way, it's awesome. Let me know if I can sponsor you.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4877,7 +4877,7 @@ class RegexpLike(Binary, Func):
     arg_types = {"this": True, "expression": True, "flag": False}
 
 
-class RegexpILike(Func):
+class RegexpILike(Binary, Func):
     arg_types = {"this": True, "expression": True, "flag": False}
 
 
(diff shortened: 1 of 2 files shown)


GeorgeSittas (approved):
> Maybe I'm missing something, but this looks like a simple omission to me [...]

You're right, this seems like an oversight - LGTM, thanks! We could also add a couple of tests, for example to assert that the following two are `Binary` when parsed (Postgres):

```
'thomas' ~ '.*thomas.*'
'thomas' ~* '.*thomas.*'
```

samuelcolvin:
> We could also add a couple of tests

Done.

tobymao:
@samuelcolvin thanks for the pr! happy to chat about sponsorship or partnership, wanna join our slack? tobikodata.com/slack or i can dm you on twitter

PR #2405, opened 2023-10-14
Fix!(bigquery): reduce eq and neq to null if either operand is null

GeorgeSittas:
Context: https://github.com/tobymao/sqlglot/pull/2403#issuecomment-1760681507

Screenshot 2023-10-14 at 5 06 03 PM

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -632,15 +632,6 @@ class BigQuery(Dialect):
             "within",
         }
 
-        def eq_sql(self, expression: exp.EQ) -> str:
-            # Operands of = cannot be NULL in BigQuery
-            if isinstance(expression.left, exp.Null):
-                return self.sql(expression.right.is_(expression.left))
-            if isinstance(expression.right, exp.Null):
-                return self.sql(expression.left.is_(expression.right))
-
-            return self.binary(expression, "=")
-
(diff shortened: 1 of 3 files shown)


tobymao:
don’t we do this for another dialect?

GeorgeSittas:
> don’t we do this for another dialect?

Not sure, couldn't find `eq` overridden elsewhere.

tobymao:
do we want to do this for all dialects? can you checkout the meaning of = null in others? it's usually incorrect, we could do this at parse time as well

GeorgeSittas:
> do we want to do this for all dialects? can you checkout the meaning of = null in others? it's usually incorrect, we could do this at parse time as well

I have documented some of the dialects we support that were easy to test:

> > FWIW select (null = null) is null yields null in postgres, sqlite, mysql, duckdb, spark, trino, snowflake, redshift, clickhouse, tsql (couldn't test teradata, oracle, hive, drill, doris, etc).

So yeah perhaps we can replace `=` with the `IS` operator at parse time, but that'd be a breaking change because it changes the semantics. Not sure if I'm missing any weird edge case.

tobymao:
ok if it’s only bq we can keep it

GeorgeSittas:
Cool I think this is good to go from me then.

tobymao (approved):

PR #2407, opened 2023-10-14
Fix(clickhouse): map RegexpILike to match

### LINKED ISSUE #2406: `REGEXP_I_LIKE` method doesn't exist in ClickHouse

sqlglot converts regexp ilike to `REGEXP_I_LIKE` in ClickHouse which doesn't exist.

```py
import sqlglot

sql = "'thomas' ~* '.*thomas.*'"
print(sqlglot.parse_one(sql, read='postgres').sql(dialect='clickhouse'))
#> REGEXP_I_LIKE('thomas', '.*thomas.*')
```

The the best option of an equivalent I've found is something like `f"match('{haystack}', '{pattern}?i')"` to use the [`match`](https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#match) function.

<details>
<summary>Confirmation the function doesn't exist</summary>

There's no such `REGEXP_I_LIKE` method in Clickhouse, so no documentation to link to, but here's the output if I run that query against CH:

```
SELECT REGEXP_I_LIKE('thomas', '.*thomas.*')

Query id: f38f80b0-59a7-4146-b33c-e345078480ee


0 rows in set. Elapsed: 0.447 sec. 

Received exception from server (version 23.8.1):
Code: 46. DB::Exception: Received from localhost:9000. DB::Exception: Unknown function REGEXP_I_LIKE. Maybe you meant: ['REGEXP_REPLACE']: While processing REGEXP_I_LIKE('thomas', '.*thomas.*'). (UNKNOWN_FUNCTION)
```

</details>

GeorgeSittas:
Fixes #2406

Couldn't find another variant for case-insensitive matching. There's `ILIKE` but I don't think it works with regex patterns?

Reference: https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -388,7 +388,6 @@ class ClickHouse(Dialect):
             exp.Pivot: no_pivot_sql,
             exp.Quantile: _quantile_sql,
             exp.RegexpLike: lambda self, e: f"match({self.format_args(e.this, e.expression)})",
-            exp.RegexpILike: lambda self, e: f"match({self.format_args(e.this, e.expression)})",
             exp.StartsWith: rename_func("startsWith"),
             exp.StrPosition: lambda self, e: f"position({self.format_args(e.this, e.args.get('substr'), e.args.get('position'))})",
             exp.VarMap: lambda self, e: _lower_func(var_map_sql(self, e)),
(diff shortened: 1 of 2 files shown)


tobymao:
don't you need to modify the regex to add case sensitivy?

GeorgeSittas:
Ah I missed that, thanks. I can fix it a bit later unless you wanna take this to the finish line.

tobymao:
https://github.com/google/re2/wiki/Syntax maybe regex on the regx to see if there's a section (?flag)

samuelcolvin:
Awesome, thanks for catching that.

PR #2409, opened 2023-10-15
add some more types

samuelcolvin:
I had quite a lot of `cast`s in my code, this should remove most of them.

It's a shame we can't mark `expression.this` as a `Expression` which is always seems to be for me, but in you code you have some places where you're assuming it's a `str`, so I assume it's not always an `Expression`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,8 +25,8 @@ repos:
         files: ^(sqlglot/|tests/|setup.py)
       - id: mypy
         name: mypy
-        entry: mypy
+        entry: mypy sqlglot tests
         language: system
         types: [ python ]
         files: ^(sqlglot/|tests/)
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -25,8 +25,8 @@ repos:
         files: ^(sqlglot/|tests/|setup.py)
       - id: mypy
         name: mypy
-        entry: mypy
+        entry: mypy sqlglot tests
```
samuelcolvin, line 28: you were previously running `mypy` on every file that changed, but typing changes can't be isolated per file, hence mypy had to typecheck the entire code base of each file changed.

This should be much faster.

tobymao:
can you please fix the build?

tobymao:
i'll take this to the finish line

PR #2416, opened 2023-10-17
Fix: Trino View Column and Partition Column

eakmanrq:


tests/dialects/test_databricks.py
```diff
@@ -6,8 +6,8 @@ class TestDatabricks(Validator):
 
     def test_databricks(self):
         self.validate_identity("CREATE TABLE t (c STRUCT<interval: DOUBLE COMMENT 'aaa'>)")
-        self.validate_identity("CREATE TABLE my_table () TBLPROPERTIES (a.b=15)")
-        self.validate_identity("CREATE TABLE my_table () TBLPROPERTIES ('a.b'=15)")
+        self.validate_identity("CREATE TABLE my_table TBLPROPERTIES (a.b=15)")
+        self.validate_identity("CREATE TABLE my_table TBLPROPERTIES ('a.b'=15)")
```
eakmanrq, line 10: I tested this on Databricks and it actually did let me a create an empty table but I wonder what the use case of this is. Is it worth adding logic to make sure Databricks can make an empty table?

eakmanrq: Oh and after I created the empty table I got an error trying to read from it saying it had no columns

eakmanrq: I could change `schema_columns_sql` to this if we want to preserve this behavior:
```
  def schema_columns_sql(self, expression: exp.Schema) -> str:
      if expression.expressions or expression.parent.args.get("kind") == 'TABLE':
          return f"({self.sep('')}{self.expressions(expression)}{self.seg(')', sep='')}"
      return ""
```

sqlglot/dialects/presto.py
```diff
@@ -168,6 +169,19 @@ def _first_last_sql(self: Presto.Generator, expression: exp.First | exp.Last) ->
     return rename_func("ARBITRARY")(self, expression)
 
 
+def _create_sql(self: Presto.Generator, expression: exp.Create) -> str:
```
tobymao, line 172: let's override create_sql instead of passing it in, passing it in is more for small funcs

tests/dialects/test_databricks.py
tobymao: is there a difference between create table () and no paren?

https://github.com/tobymao/sqlglot/issues/2289

eakmanrq: Both are valid

eakmanrq: Reading from both gives me the same error so I believe so. I'm just testing on Delta on Databricks btw not sure if there are nuances across formats.

PR #2417, opened 2023-10-17
Feat(optimizer): qualify lateral flatten for snowflake

### LINKED ISSUE #2413: Support lineage through a LATERAL FLATTEN from clause in Snowflake
I have a Snowflake SQL query like `SELECT VALUE FROM TABLE_A, LATERAL FLATTEN(TABLE_A.VARIANT_COLUMN)`. I would like to be able to use the `lineage` function to get the lineage tree of the column "VALUE" (in case TABLE_A is actually a subquery, cte, etc).

lineage should understand that VALUE's lineage is through TABLE_A.VARIANT_COLUMN (or through a subquery or CTE in place of TABLE_A). The `FLATTEN` function, when the Snowflake dialect is selected, should be treated as a select expression that adds the columns documented as [its output](https://docs.snowflake.com/en/sql-reference/functions/flatten#output).

The lineage tree of value should look something like VALUE -> FLATTEN(TABLE_A>VARIANT_COLUMN) -> TABLE_A.VARIANT_COLUMN -> TABLE_A. 

My plan was to have the Snowflake dialect parser produce a TableAlias expression when it parses a FLATTEN function populated with with the columns produces by the expression. Right now a TableAlias is produced, but it has no columns, so lineage fails raising a ValueError.

I'm happy to work on this feature if it's something you want to support in sqlglot but wanted to open this issue to make sure my suggestion of an implementation strategy make sense and is something you'd accept.

edpaget:
Adds columns created by the FLATTEN function in snowflake to the TableAlias created when using it in a FROM ..., LATERAL FLATTEN call following the solution suggested by @GeorgeSittas. To get the full lineage I had to make each column created by the function an alias of the flatten function so we could get the lineage back to the source column flatten was reading from. It seems a little gross, so also wanted to check if there should be a exp.Flatten AST node that encapsulates this behaviour instead. 

Closes #2413

### DIFF
--- a/sqlglot/dataframe/sql/functions.py
+++ b/sqlglot/dataframe/sql/functions.py
@@ -1113,7 +1113,7 @@ def reverse(col: ColumnOrName) -> Column:
 
 
 def flatten(col: ColumnOrName) -> Column:
-    return Column.invoke_anonymous_function(col, "FLATTEN")
+    return Column.invoke_expression_over_column(None, expression.Flatten, expressions=[col])
 
 
 def map_keys(col: ColumnOrName) -> Column:
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -206,7 +206,6 @@ class Snowflake(Dialect):
     TIME_FORMAT = "'YYYY-MM-DD HH24:MI:SS'"
     SUPPORTS_USER_DEFINED_TYPES = False
     SUPPORTS_SEMI_ANTI_JOIN = False
-    FLATTEN_COLUMNS = ["SEQ", "KEY", "PATH", "INDEX", "VALUE", "THIS"]
 
     TIME_MAPPING = {
         "YYYY": "%Y",
@@ -326,6 +325,7 @@ class Snowflake(Dialect):
             TokenType.MOD,
             TokenType.SLASH,
         }
+        FLATTEN_COLUMNS = ["SEQ", "KEY", "PATH", "INDEX", "VALUE", "THIS"]
 
         def _parse_lateral(self) -> t.Optional[exp.Lateral]:
             lateral = super()._parse_lateral()
@@ -334,7 +334,7 @@ class Snowflake(Dialect):
 
             if isinstance(lateral.this, exp.Flatten):
                 table_alias = lateral.args.get("alias")
-                columns = [exp.to_identifier(col) for col in Snowflake.FLATTEN_COLUMNS]
+                columns = [exp.to_identifier(col) for col in self.FLATTEN_COLUMNS]
                 if table_alias and not table_alias.args.get("columns"):
                     table_alias.set("columns", columns)
                 elif not table_alias:
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -206,6 +206,7 @@ class Snowflake(Dialect):
     TIME_FORMAT = "'YYYY-MM-DD HH24:MI:SS'"
     SUPPORTS_USER_DEFINED_TYPES = False
     SUPPORTS_SEMI_ANTI_JOIN = False
+    FLATTEN_COLUMNS = ["SEQ", "KEY", "PATH", "INDEX", "VALUE", "THIS"]
 
     TIME_MAPPING = {
         "YYYY": "%Y",
@@ -325,6 +326,21 @@ class Snowflake(Dialect):
             TokenType.MOD,
             TokenType.SLASH,
         }
+
+        def _parse_lateral(self) -> t.Optional[exp.Lateral]:
+            lateral = super()._parse_lateral()
+            if not lateral:
+                return lateral
+
+            if isinstance(lateral.this, exp.Flatten):
+                table_alias = lateral.args.get("alias")
+                columns = [exp.to_identifier(col) for col in Snowflake.FLATTEN_COLUMNS]
+                if table_alias and not table_alias.args.get("columns"):
+                    table_alias.set("columns", columns)
+                elif not table_alias:
+                    lateral.set("alias", self.expression(exp.TableAlias, columns=columns))
+
+            return lateral
 
         def _parse_table_parts(self, schema: bool = False) -> exp.Table:
             # https://docs.snowflake.com/en/user-guide/querying-stage
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -262,6 +262,7 @@ class Snowflake(Dialect):
             ),
             "DATEDIFF": _parse_datediff,
             "DIV0": _div0_to_if,
+            "FLATTEN": exp.Explode.from_arg_list,
             "IFF": exp.If.from_arg_list,
             "LISTAGG": exp.GroupConcat.from_arg_list,
             "NULLIFZERO": _nullifzero_to_if,
@@ -332,13 +333,13 @@ class Snowflake(Dialect):
             if not lateral:
                 return lateral
 
-            if isinstance(lateral.this, exp.Flatten):
+            if isinstance(lateral.this, exp.Explode):
                 table_alias = lateral.args.get("alias")
                 columns = [exp.to_identifier(col) for col in self.FLATTEN_COLUMNS]
                 if table_alias and not table_alias.args.get("columns"):
                     table_alias.set("columns", columns)
                 elif not table_alias:
-                    lateral.set("alias", self.expression(exp.TableAlias, columns=columns))
+                    exp.alias_(lateral, "_flattened", table=columns, copy=False)
 
             return lateral
 
@@ -468,6 +469,7 @@ class Snowflake(Dialect):
             exp.DateStrToDate: datestrtodate_sql,
             exp.DataType: _datatype_sql,
             exp.DayOfWeek: rename_func("DAYOFWEEK"),
+            exp.Explode: rename_func("FLATTEN"),
             exp.Extract: rename_func("DATE_PART"),
             exp.GenerateSeries: lambda self, e: self.func(
                 "ARRAY_GENERATE_RANGE", e.args["start"], e.args["end"] + 1, e.args.get("step")
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4168,7 +4168,6 @@ class Abs(Func):
     pass
 
 
-# https://docs.snowflake.com/en/sql-reference/functions/flatten
 class Flatten(Func):
     arg_types = {"expressions": True}
     is_var_len_args = True
@@ -4546,8 +4545,10 @@ class Exp(Func):
     pass
 
 
+# https://docs.snowflake.com/en/sql-reference/functions/flatten
 class Explode(Func):
-    pass
+    arg_types = {"this": True, "expressions": False}
+    is_var_len_args = True
 
 
 class ExplodeOuter(Explode):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4168,6 +4168,12 @@ class Abs(Func):
     pass
 
 
+# https://docs.snowflake.com/en/sql-reference/functions/flatten
+class Flatten(Func):
+    arg_types = {"expressions": True}
+    is_var_len_args = True
+
+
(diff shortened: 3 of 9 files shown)


tests/fixtures/optimizer/qualify_tables.sql
```diff
@@ -109,3 +109,8 @@ SELECT * FROM ((SELECT * FROM c.db.t AS t) AS _q_0);
 # title: wrapped subquery without alias joined with a table
 SELECT * FROM ((SELECT * FROM t1) INNER JOIN t2 ON a = b);
 SELECT * FROM ((SELECT * FROM c.db.t1 AS t1) AS _q_0 INNER JOIN c.db.t2 AS t2 ON a = b);
+
+# title: using lateral flatten
+# dialect: snowflake
+SELECT flattened.value FROM z, LATERAL FLATTEN(input => result, outer => true) as flattened;
+SELECT flattened.value FROM c.db.z AS z, LATERAL FLATTEN(input => result, outer => TRUE) AS flattened(FLATTEN(input => result, outer => TRUE) AS SEQ, FLATTEN(input => result, outer => TRUE) AS KEY, FLATTEN(input => result, outer => TRUE) AS PATH, FLATTEN(input => result, outer => TRUE) AS INDEX, FLATTEN(input => result, outer => TRUE) AS VALUE, FLATTEN(input => result, outer => TRUE) AS THIS);
```
GeorgeSittas, line 116: @edpaget can you explain this test a little? it looks way more complex than expected.

What I had in mind is `... flatten(...) as t(seq, key, ....)`, why do we need to repeat all these `FLATTEN` calls?

sqlglot/optimizer/qualify_tables.py
```diff
@@ -96,5 +98,25 @@ def qualify_tables(
...
+
+
+def _qualify_snowflake_lateral(lateral: exp.Lateral) -> None:
+    if isinstance(lateral.this, exp.Anonymous) and lateral.this.this and lateral.this.this.upper() == "FLATTEN":
+        table_alias = lateral.args.get("alias")
+        if table_alias:
+            table_alias.set(
+                "columns",
+                [
+                    exp.Alias(this=lateral.this, alias=exp.to_identifier(col))
```
GeorgeSittas, line 118: The `lateral.this` part is probably repeating all the `FLATTEN`s, it doesn't seem right.

GeorgeSittas (commented):
Hey @edpaget, I left a few comments.

I chatted with Toby about this and maybe it'd be better if we moved this to the parser after all. Let me see if I can clean this up a bit on top of your PR.

EDIT: please make sure to also run `make style` to clean up any formatting-related errors and allow CI to succeed.

tests/fixtures/optimizer/qualify_tables.sql
edpaget: To answer both this question and the [one here](https://github.com/tobymao/sqlglot/pull/2417#discussion_r1362489989). This allowed the lineage function to find the source for these columns.

edpaget: I agree its bad! Happy to solve this in the parser.

GeorgeSittas: What is `lineage` doing when you create a table alias like the one I mentioned above? Why doesn't it work?

sqlglot/optimizer/qualify_tables.py
```diff
@@ -96,5 +98,25 @@ def qualify_tables(
...
+
 
     return expression
+
+
+FLATTEN_COLUMNS = ["SEQ", "KEY", "PATH", "INDEX", "VALUE", "THIS"]
+
+
+def _qualify_snowflake_lateral(lateral: exp.Lateral) -> None:
+    if isinstance(lateral.this, exp.Anonymous) and lateral.this.this and lateral.this.this.upper() == "FLATTEN":
```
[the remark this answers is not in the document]
edpaget: For sure. I didn't know if there was some kind of policy about not implementing exp. classes when it was only supporting one database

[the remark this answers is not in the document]
tobymao: we can always create new expressions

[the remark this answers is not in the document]
GeorgeSittas: I think this should do the trick

```diff
+++ b/sqlglot/dataframe/sql/functions.py
@@ -1113,7 +1113,7 @@ def reverse(col: ColumnOrName) -> Column:


 def flatten(col: ColumnOrName) -> Column:
-    return Column.invoke_anonymous_function(col, "FLATTEN")
+    return Column.invoke_expression_over_column(None, expression.Flatten, expressions=[col])

--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4168,6 +4168,11 @@ class Abs(Func):
     pass


+class Flatten(Func):
+    arg_types = {"expressions": True}
+    is_var_len_args = True
+
+
```

GeorgeSittas:
@edpaget let me know if you want me to take a stab at this or prefer to keep at it so I can prioritize accordingly. There may be just a couple of things needed to get this to work, still need to understand why `flatten(...) as t(...)` fails to enable `lineage` to work as expected.

You can also join our [Slack](https://tobikodata.com/slack) if you want, it'll help us iterate faster on this.

tests/fixtures/optimizer/qualify_tables.sql
edpaget: It stops when it finds the column in the lateral flatten scope and doesn't have a way to find the source lateral flatten is reading from.

edpaget:
I'll join the slack. I reserved time in my team's sprint to work this and would enjoy understanding more sqlglot internals, so can keep working on it if it's not too much trouble for you.

GeorgeSittas:
> I'll join the slack. I reserved time in my team's sprint to work this and would enjoy understanding more sqlglot internals, so can keep working on it if it's not too much trouble for you.

Sounds good, thanks!

sqlglot/dialects/snowflake.py
```diff
@@ -206,6 +206,7 @@ class Snowflake(Dialect):
     TIME_FORMAT = "'YYYY-MM-DD HH24:MI:SS'"
     SUPPORTS_USER_DEFINED_TYPES = False
     SUPPORTS_SEMI_ANTI_JOIN = False
+    FLATTEN_COLUMNS = ["SEQ", "KEY", "PATH", "INDEX", "VALUE", "THIS"]
```
GeorgeSittas, line 209: I'd move this either in the top-level or inside Snowflake's parser class.

sqlglot/dataframe/sql/functions.py
```diff
@@ -1113,7 +1113,7 @@ def reverse(col: ColumnOrName) -> Column:
 
 
 def flatten(col: ColumnOrName) -> Column:
-    return Column.invoke_anonymous_function(col, "FLATTEN")
+    return Column.invoke_expression_over_column(None, expression.Flatten, expressions=[col])
```
tobymao, line 1116: is this semantically the same for spark?? if not it’s may not be right

sqlglot/dialects/snowflake.py
```diff
@@ -326,6 +327,21 @@ class Parser(parser.Parser):
...
+            if not lateral:
+                return lateral
+
+            if isinstance(lateral.this, exp.Flatten):
+                table_alias = lateral.args.get("alias")
+                columns = [exp.to_identifier(col) for col in Snowflake.FLATTEN_COLUMNS]
+                if table_alias and not table_alias.args.get("columns"):
+                    table_alias.set("columns", columns)
+                elif not table_alias:
+                    lateral.set("alias", self.expression(exp.TableAlias, columns=columns))
```
GeorgeSittas, line 341: Don't we need to give an alias to the lateral source here? For example `_flattened` or something. We could also use the `exp.alias_` helper to set the alias / columns here.

sqlglot/dataframe/sql/functions.py
GeorgeSittas: I think the only difference is that now we have a dedicated `Flatten` expression as opposed to an `Anonymous` one. For reference, Spark's `FLATTEN` is documented in https://spark.apache.org/docs/latest/api/sql/index.html#flatten. We also have a couple of dataframe tests for this (see test_flatten).

edpaget: Was using [this diff](https://github.com/tobymao/sqlglot/pull/2417#discussion_r1362516004), but flatten is a list flatten function not the same as the snowflake function.

GeorgeSittas: Yeah they definitely don't have the same semantics and syntax, I think though Toby was asking whether or not this is breaking in terms of SQLGlot AST representation?

sqlglot/dialects/snowflake.py
edpaget: Didn't know if we could have a None alias. It looks like self._parse_table_alias _could_ have an empty alias since [it only tests](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L2229) for `not alias and not columns`. I can exp.alias_ here.

sqlglot/dataframe/sql/functions.py
tobymao: flatten should be list flatten, let’s make another node that is snowflakes variant

sqlglot/dialects/snowflake.py
GeorgeSittas: It can have a `None` alias, but that'd look like the following query:

```sql
SELECT 1 FROM t AS (c)
```

Another nuance here is that if there are multiple `FLATTEN` calls without aliases we might need to create unique names for them by using something like `sqlglot.helper.find_new_name`, but we can address this later to if needed I guess.

sqlglot/dataframe/sql/functions.py
edpaget: so have a exp.SnowflakeFlatten?

tobymao: call it something more generic but yes, dunno what’s a good phrase but don’t use the word snowflake, describe its func

GeorgeSittas: snowflake's flatten is basically an `explode`, right? perhaps we can reuse that node instead? cc @tobymao

reference: https://docs.snowflake.com/en/sql-reference/functions/flatten

> Flattens (explodes) compound values into multiple rows.

sqlglot/dialects/snowflake.py
edpaget: It seems like multiple lateral flatten without aliases wouldn't work anyway (hopefully 🤞)

sqlglot/dataframe/sql/functions.py
tobymao: yea, then just map snowflake flatten to explode, that could help with transpilation

sqlglot/dialects/snowflake.py
GeorgeSittas: Ah I see, that's good. Snowflake seems to create the same name for flattens without aliases (i.e. `flatten`) and chokes due to duplicate aliases. Good catch 👍

sqlglot/dataframe/sql/functions.py
edpaget: Mapped it to explode

tobymao (approved):

sqlglot/expressions.py
```diff
@@ -4168,6 +4168,11 @@ class Abs(Func):
     pass
 
 
+class Flatten(Func):
+    arg_types = {"expressions": True}
+    is_var_len_args = True
```
GeorgeSittas, line 4173: This is not used right?

GeorgeSittas (approved):
Just one comment, LGTM otherwise. Thanks for the quick iteration @edpaget!

sqlglot/expressions.py
tobymao: delete it and merge

GeorgeSittas: I see, it was added to represent Spark's `Flatten` judging from the dataframe test. I'll merge and delete it since we don't use it elsewhere.

tobymao: i thought you meant just the is_var_len_args

PR #2424, opened 2023-10-18
Feat(lineage): lineage from UDTFs that use columns

edpaget:
This is the second part to solve extracting a full lineage from a `LATERAL FLATTEN(...` statement in Snowflake #2413, and adds support for getting all columns used in a UDTF to then build downstream nodes for those columns. 

This extends the set of columns generated by doing a `select.find_all(exp.Column)` with any columns found by searching `source.find_all(exp.Column)`, when the source is an instance of `exp.UDTF`. I'm not 100% that's the best way to implement it -- will select.find_all(...) ever return other columns when reading from a UDTF? -- if not this should probably be an `if isinstance(source, exp.UDTF)...else...` instead.

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -143,7 +143,13 @@ def lineage(
             upstream.downstream.append(node)
 
         # Find all columns that went into creating this one to list their lineage nodes.
-        for c in set(select.find_all(exp.Column)):
+        source_columns = set(select.find_all(exp.Column))
+
+        # If the source is a UDTF find columns used in the UTDF to generate the table
+        if isinstance(source, exp.UDTF):
+            source_columns |= set(source.find_all(exp.Column))
+
+        for c in set(source_columns):
             table = c.table
             source = scope.sources.get(table)
 
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -149,7 +149,7 @@ def lineage(
         if isinstance(source, exp.UDTF):
             source_columns |= set(source.find_all(exp.Column))
 
-        for c in set(source_columns):
+        for c in source_columns:
             table = c.table
             source = scope.sources.get(table)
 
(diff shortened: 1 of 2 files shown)


sqlglot/lineage.py
```diff
@@ -143,7 +143,13 @@ def to_node(
...
 
         # Find all columns that went into creating this one to list their lineage nodes.
-        for c in set(select.find_all(exp.Column)):
+        source_columns = set(select.find_all(exp.Column))
+
+        # If the source is a UDTF find columns used in the UTDF to generate the table
+        if isinstance(source, exp.UDTF):
+            source_columns |= set(source.find_all(exp.Column))
+
+        for c in set(source_columns):
```
GeorgeSittas, line 152: This `set` constructor seems redundant

edpaget: oops

GeorgeSittas (approved):
Looks good at first glance, might need to think a bit more about how these changes interact with the other UDTFs (`Unnest`, `Values`).

edpaget:
Happy to add some tests around those.

edpaget:
I can add more tests. Was just pulled into some meetings before I could implement.

edpaget:
I tried writing a test for `select id, x from table_test, lateral unnest(table_test.xs) x` which should at least be valid postgres sql, but this is being qualified to `SELECT id AS id, x AS x FROM table_test AS table_test, LATERAL UNNEST(table_test.xs) AS x AS _q_0`, which feels unexpected to me and means I haven't been able to write a test that actually gets to the `unnest` function.

edpaget:
imo, I think this is good to ship as is (at least for my use case) and we could circle back and look at how unnest is working.

GeorgeSittas:
I think there's a bug in the `qualify_tables` rule because it adds a redundant alias to the udtf when it shouldn't - I'll take a look afterwards.

> imo, I think this is good to ship as is (at least for my use case) and we could circle back and look at how unnest is working.

Sounds good.

PR #2439, opened 2023-10-22
Fix(clickhouse): treat CURRENT_DATE as a function

### LINKED ISSUE #2438: Clickhouse `CURRENT_DATE` is not rendered as a function
```py
import sqlglot

print('version:', sqlglot.__version__)

print(sqlglot.transpile('select current_date', read='postgres', write='clickhouse'))
```

Outputs:

```
version: 18.16.1
['SELECT CURRENT_DATE']
```

This is invalid in Clickhouse, see [the docs](https://clickhouse.com/docs/en/sql-reference/functions/date-time-functions#today), `CURRENT_DATE` (alias of `today` and `curdate`) is a function, thus the output in this case should be `SELECT current_date()`.

Case doesn't seem to matter in this this case, although the docs show lowercase.

GeorgeSittas:
Fixes #2438

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -375,6 +375,7 @@ class ClickHouse(Dialect):
             exp.ApproxDistinct: rename_func("uniq"),
             exp.Array: inline_array_sql,
             exp.CastToStrType: rename_func("CAST"),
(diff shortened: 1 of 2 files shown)


samuelcolvin (approved):
LGTM.

PR #2441, opened 2023-10-23
Feat(optimizer): infer bracket type

GeorgeSittas:


tobymao (approved):

sqlglot/optimizer/annotate_types.py
```diff
@@ -525,3 +526,20 @@ def _annotate_dateadd(self, expression: exp.IntervalOp) -> exp.IntervalOp:
...
         return expression
+
+    def _annotate_bracket(self, expression: exp.Bracket) -> exp.Bracket:
+        self._annotate_args(expression)
+
+        bracket_arg = expression.expressions[0]
+
+        if isinstance(bracket_arg, exp.Slice):
+            self._set_type(expression, expression.this.type)
+        elif expression.this.type.is_type(exp.DataType.Type.ARRAY):
```
tobymao, line 537: one more test case is the map type, map[1] or map[x] or map['s'] etc

GeorgeSittas: looking into this now

GeorgeSittas: Ok added support for inference on map brackets as well

tobymao (approved):

PR #2443, opened 2023-10-23
Fix(redshift): don't rewrite JSON_PARSE to PARSE_JSON

### LINKED ISSUE #2442: Erroneous handling of redshift's JSON_PARSE
**sqlglot version: 18.16.1**

```python
import sqlglot
sql = "SELECT JSON_PARSE('[10001,10002,\"abc\"]');"
parsed = sqlglot.parse_one(sql,dialect="redshift")
parsed.sql(dialect="redshift")
#'SELECT PARSE_JSON(\'[10001,10002,"abc"]\')'
```
The generated sql triggers an error when executed in redshift

>Failed to execute query: ERROR: function parse_json("unknown") does not exist
  Hint: No function matches the given name and argument types. You may need to add explicit type casts.

**Official Documentation**
https://docs.aws.amazon.com/redshift/latest/dg/JSON_PARSE.html

purcell:
Fixes #2442, following the example of how the Trino dialect overrides the naming for this function.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -175,6 +175,7 @@ class Redshift(Postgres):
             exp.GeneratedAsIdentityColumnConstraint: generatedasidentitycolumnconstraint_sql,
             exp.JSONExtract: _json_sql,
             exp.JSONExtractScalar: _json_sql,
+            exp.ParseJSON: rename_func("JSON_PARSE"),
             exp.SafeConcat: concat_to_dpipe_sql,
             exp.Select: transforms.preprocess(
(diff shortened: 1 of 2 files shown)


GeorgeSittas (approved):
Thanks for the fix!

purcell:
And thanks for the fast merge 🙏

purcell:
Just so I know for my own planning purposes, do you think there might be a release this week which includes this fix?

GeorgeSittas:
I just release v18.17.0, should be live in a couple of minutes

purcell:
Amazing, thank you! 🙏

Issue #2426, opened 2023-10-19
Trino to spark-sql date_diff gives syntax issue.


as-sher:
When transpiling date_diff in trino to spark-sql

Trino
SELECT date_diff('month', TIMESTAMP '2020-02-29 00:00:00', TIMESTAMP '2020-03-02 00:00:00');

Spark-sql 
SELECT DATEDIFF(month, CAST('2020-02-29 00:00:00' AS TIMESTAMP), CAST('2020-03-02 00:00:00' AS TIMESTAMP))

But spark does not support 3 params in datediff 
https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.sql.functions.datediff.html
Please can you help here?

GeorgeSittas:
It does work with 3 parameters, it's just not documented well:

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT date_diff('month', TIMESTAMP '2020-02-29 00:00:00', TIMESTAMP '2020-03-02 00:00:00');", "trino", "spark")
["SELECT DATEDIFF(month, CAST('2020-02-29 00:00:00' AS TIMESTAMP), CAST('2020-03-02 00:00:00' AS TIMESTAMP))"]
```

```
spark-sql (default)> SELECT DATEDIFF(month, CAST('2020-02-29 00:00:00' AS TIMESTAMP), CAST('2020-03-02 00:00:00' AS TIMESTAMP));
0
Time taken: 4.423 seconds, Fetched 1 row(s)
```

GeorgeSittas:
FWIW if you're using Spark v2 then there's an option to transpile towards that to avoid this issue if needed:

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT date_diff('month', TIMESTAMP '2020-02-29 00:00:00', TIMESTAMP '2020-03-02 00:00:00');", "trino", "spark2")
["SELECT MONTHS_BETWEEN(CAST('2020-03-02 00:00:00' AS TIMESTAMP), CAST('2020-02-29 00:00:00' AS TIMESTAMP))"]
```

as-sher:
Thanks a ton

as-sher:
@GeorgeSittas  the query
`["SELECT MONTHS_BETWEEN(CAST('2020-03-02 00:00:00' AS TIMESTAMP), CAST('2020-02-29 00:00:00' AS TIMESTAMP))"]`
Gives a decimal number which is not same as redshift.

[...]

PR #2445, opened 2023-10-23
Fix!(hive): cast months_between to int when generated from datediff

GeorgeSittas:
Resolves https://github.com/tobymao/sqlglot/issues/2426#issuecomment-1774417332

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -1058,18 +1058,18 @@ WHERE
             },
         )
         self.validate_all(
-            "SELECT DATEDIFF(year, '2020/01/01', '2021/01/01')",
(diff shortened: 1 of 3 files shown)


tests/dialects/test_tsql.py
```diff
@@ -1058,18 +1058,18 @@ def test_date_diff(self):
             },
         )
         self.validate_all(
-            "SELECT DATEDIFF(year, '2020/01/01', '2021/01/01')",
+            "SELECT DATEDIFF(year, '2020-01-01', '2021-01-01')",
```
GeorgeSittas, line 1061: I changed this test because Spark doesn't recognize the `yy/mm/dd` format and so it returns `NULL`, which means the transpilation is not correct.

PR #2452, opened 2023-10-24
Fix(lineage): handle unions with SELECT STAR

edpaget:
Building lineage when unions have a `SELECT *` statement fails with a StopIteration error because there is no fallback case if the exact column is not found in the Union's select expressions. This adds a fallback case to handle these selects when building the lineage from the union.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2450,7 +2450,7 @@ QUERY_MODIFIERS = {
     "windows": False,
     "distribute": False,
     "sort": False,
-    "cluster": False,
+    "clustass": False,
     "order": False,
     "limit": False,
     "offset": False,
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2450,7 +2450,7 @@ QUERY_MODIFIERS = {
     "windows": False,
     "distribute": False,
     "sort": False,
-    "clustass": False,
+    "cluster": False,
     "order": False,
     "limit": False,
     "offset": False,
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -93,36 +93,15 @@ def lineage(
 
         # Find the specific select clause that is the source of the column we want.
         # This can either be a specific, named select or a generic `*` clause.
-        select = (
-            scope.expression.selects[column]
-            if isinstance(column, int)
-            else next(
-                (select for select in scope.expression.selects if select.alias_or_name == column),
-                exp.Star() if scope.expression.is_star else None,
-            )
-        )
-
-        if not select:
-            raise ValueError(f"Could not find {column} in {scope.expression}")
+        select = _must_column_from_select(scope, column)
 
         if isinstance(scope.expression, exp.Union):
             upstream = upstream or Node(name="UNION", source=scope.expression, expression=select)
 
-            index = (
-                column
-                if isinstance(column, int)
-                else next(
-                    (
-                        i
-                        for i, select in enumerate(scope.expression.selects)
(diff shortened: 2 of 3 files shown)


GeorgeSittas (approved):
LGTM, should we assert in the test that the downstream nodes are table_a and table_b?

edpaget:
I'll update the tests to assert on the tables as well. I realized after my change that the index = assignment in the Union if block was actually just duplicating the same thing as the select statement, so I ended up removing it entirely, which seem to be fine to me.

sqlglot/lineage.py
```diff
@@ -102,24 +103,13 @@ def to_node(
             )
         )
 
-        if not select:
-            raise ValueError(f"Could not find {column} in {scope.expression}")
+        raise ValueError(f"Could not find {column} in {scope.expression}")
```
GeorgeSittas, line 106: this doesn't look correct

sqlglot/expressions.py
```diff
@@ -2450,7 +2450,7 @@ def with_(
     "windows": False,
     "distribute": False,
     "sort": False,
-    "cluster": False,
+    "clustass": False,
```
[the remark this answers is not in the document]
edpaget: oops

edpaget:
The last thing I'm trying to get to is that the 'source' for the select isn't the table and is instead the entire select * from table_a statement, which seems not quite right?

sqlglot/lineage.py
```diff
@@ -108,18 +109,8 @@ def to_node(
...
-                else next(
-                    i
-                    for i, select in enumerate(scope.expression.selects)
-                    if select.alias_or_name == column
-                )
-            )
-
             for s in scope.union_scopes:
-                to_node(index, scope=s, upstream=upstream)
+                to_node(select.name, scope=s, upstream=upstream)
```
GeorgeSittas, line 113: These two don't seem equivalent - we have code that also handles the case where `column` is `int` above. What is the context behind this change?

edpaget: After my change the expressions that were setting index and select where the same except that index was returning an int or star and select a column or star expression, I did not see a reason to keep the duplicate code.

GeorgeSittas:
> The last thing I'm trying to get to is that the 'source' for the select isn't the table and is instead the entire select * from table_a statement, which seems not quite right?

yeah what you're saying looks reasonable, perhaps we need to descend into the select * and pick any source tables. This is not specific to the union case you're dealing with, it seems like it's more general:

```python
>>> l = lineage("c", "with t1 as (select * from t2) select c from t1")
>>> l.downstream
[Node(name='t1.c', expression=(STAR ), source=(SELECT expressions:
  (STAR ), from:
  (FROM this:
    (TABLE this:
      (IDENTIFIER this: t2, quoted: False), alias:
      (TABLEALIAS this:
        (IDENTIFIER this: t2, quoted: False))))), downstream=[], alias='')]
```

edpaget:
This part of the test is the last thing I'm not sure about. Should the * in the select expression have a source that is the tabel the * is selecting from? It seems like that already does not work. I can follow up with another PR to fix that.

edpaget:
Oops, posted that at the same time.

GeorgeSittas:
> This part of the test is the last thing I'm not sure about. Should the * in the select expression have a source that is the tabel the * is selecting from? It seems like that already does not work. I can follow up with another PR to fix that.

doing it in another pr sounds good too

edpaget:
I'll remove the commented out assert and then this should be good.

GeorgeSittas (approved):

PR #2456, opened 2023-10-24
feat(lineage): terminal table node for select star

edpaget:
adds final nodes to lineage trees when a `select *` expression is found named "*" with the source set to each table in the select * scope.

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -159,6 +159,11 @@ def lineage(
         if upstream:
             upstream.downstream.append(node)
 
+        # if the select is a star add all scope sources as downstreams
+        if select.is_star:
+            for source in scope.sources.values():
+                node.downstream.append(Node(name=select.sql(), source=source, expression=source))
+
         # Find all columns that went into creating this one to list their lineage nodes.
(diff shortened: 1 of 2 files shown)


sqlglot/lineage.py
```diff
@@ -159,6 +159,11 @@ def to_node(
         if upstream:
             upstream.downstream.append(node)
 
+        # if the select is a star add all scope sources as downstreams
+        if select.is_star:
+            for source in scope.sources.values():
+                node.downstream.append(Node(name=select.sql(), source=source, expression=source))
```
GeorgeSittas, line 165: @tobymao do we care if the `name` is set to dialect-unaware SQL in `Node`s?

GeorgeSittas (approved):
Looks good - thanks for the PRs @edpaget!

sqlglot/lineage.py
tobymao: it’s just going be select * right?

GeorgeSittas: Yeah, though you can also have star projections like ``` `t`.* ```

PR #2457, opened 2023-10-24
Fix: add table alias in unnest with columns only

### LINKED ISSUE #2455: BigQuery to DuckDB transpiring creates incorrect query
```
import sqlglot
from sqlglot.errors import ErrorLevel
from sqlglot.dialects.dialect import Dialects

query = """
WITH
  Sequences AS (
    SELECT 1 AS id, [0, 1, 1, 2, 3, 5] AS some_numbers
    UNION ALL
    SELECT 2 AS id, [2, 4, 8, 16, 32] AS some_numbers
    UNION ALL
    SELECT 3 AS id, [5, 10] AS some_numbers
  )
SELECT id AS matching_rows
FROM Sequences
WHERE EXISTS(SELECT * FROM UNNEST(some_numbers) AS x WHERE x > 5);
"""
sqlglot.transpile(query, read=Dialects.BIGQUERY, write=Dialects.DUCKDB, error_level=ErrorLevel.RAISE)
```

This gives:
```
'WITH Sequences AS (SELECT 1 AS id, [0, 1, 1, 2, 3, 5] AS some_numbers UNION ALL SELECT 2 AS id, [2, 4, 8, 16, 32] AS some_numbers UNION ALL SELECT 3 AS id, [5, 10] AS some_numbers) SELECT id AS matching_rows FROM Sequences WHERE EXISTS(SELECT * FROM UNNEST(some_numbers) AS (x) WHERE x > 5)'
```

but the `AS (x)` was supposed to be `AS tbl(x)`

GeorgeSittas:
Fixes #2455

This is probably not the most robust way to solve the issue, but not sure. What do you think @tobymao?

```
D WITH Sequences AS (
>     SELECT 1 AS id, [0, 1, 1, 2, 3, 5] AS some_numbers UNION ALL
>     SELECT 2 AS id, [2, 4, 8, 16, 32] AS some_numbers UNION ALL
>     SELECT 3 AS id, [5, 10] AS some_numbers
> )
> SELECT
>   id AS matching_rows
> FROM Sequences
> WHERE EXISTS(
>   SELECT * FROM UNNEST(some_numbers) AS _t(x) WHERE x > 5
> );
┌───────────────┐
│ matching_rows │
│     int32     │
├───────────────┤
│             2 │
│             3 │
└───────────────┘
```

Screenshot 2023-10-24 at 8 38 06 PM

### DIFF
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -143,6 +143,13 @@ class TestBigQuery(Validator):
         self.validate_all("x <> ''''''", write={"bigquery": "x <> ''"})
         self.validate_all("CAST(x AS DATETIME)", read={"": "x::timestamp"})
         self.validate_all(
+            "SELECT * FROM t WHERE EXISTS(SELECT * FROM unnest(nums) AS x WHERE x > 1)",
+            write={
+                "bigquery": "SELECT * FROM t WHERE EXISTS(SELECT * FROM UNNEST(nums) AS x WHERE x > 1)",
+                "duckdb": "SELECT * FROM t WHERE EXISTS(SELECT * FROM UNNEST(nums) AS _t(x) WHERE x > 1)",
+            },
+        )
+        self.validate_all(
             "NULL",
             read={
                 "duckdb": "NULL = a",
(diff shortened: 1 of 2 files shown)


tests/dialects/test_bigquery.py
```diff
@@ -476,9 +483,8 @@ def test_bigquery(self):
             },
             write={
                 "bigquery": "SELECT * FROM UNNEST(['7', '14']) AS x",
-                "presto": "SELECT * FROM UNNEST(ARRAY['7', '14']) AS (x)",
-                "hive": "SELECT * FROM UNNEST(ARRAY('7', '14')) AS (x)",
-                "spark": "SELECT * FROM UNNEST(ARRAY('7', '14')) AS (x)",
+                "presto": "SELECT * FROM UNNEST(ARRAY['7', '14']) AS _t(x)",
+                "spark": "SELECT * FROM UNNEST(ARRAY('7', '14')) AS _t(x)",
```
GeorgeSittas, line 487: Two notes here:

- The presto test was previously invalid, i.e. `AS (x)` is invalid syntax:
```
trino> SELECT * FROM UNNEST(ARRAY['7', '14']) AS (x);
Query 20231024_194935_00002_hyjxp failed: line 1:43: mismatched input '('. Expecting: <identifier>
```
- The hive/spark tests were also invalid, for two reasons: the `(x)` is invalid syntax and there's no `UNNEST` function; we need to use `EXPLODE` but this needs some work and we'd better keep it out of this PR's scope.

PR #2461, opened 2023-10-25
Feat(mysql): convert full outer join to union with left/right outer joins

### LINKED ISSUE #2458: Feat(mysql): transpile FULL OUTER JOIN to LEFT/UNION/RIGHT
MySQL does not support `FULL OUTER JOIN`. Whenever sqlglot is transpiling from another dialect, I'd like it to generate the equivalent operation of `LEFT JOIN` `UNION` `RIGHT JOIN`.

GeorgeSittas:
Fixes #2458

This is an incomplete fix since the generic solution is more tricky.

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -393,7 +393,7 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             full_outer_join.set("side", "left")
             expression_copy.args["joins"][index].set("side", "right")
 
-            return exp.union(expression, expression_copy)
+            return exp.union(expression, expression_copy, copy=False)
 
     return expression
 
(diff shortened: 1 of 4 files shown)


sqlglot/transforms.py
```diff
@@ -374,6 +374,30 @@ def eliminate_semi_and_anti_joins(expression: exp.Expression) -> exp.Expression:
...
+            if join.side == "FULL" and join.kind == "OUTER"
+        ]
+
+        if len(full_outer_joins) == 1:
+            expression_copy = expression.copy()
+            index, full_outer_join = full_outer_joins[0]
+            full_outer_join.set("side", "left")
+            expression_copy.args["joins"][index].set("side", "right")
+
+            return exp.union(expression, expression_copy)
```
tobymao, line 396: copy=False here?

tobymao (approved):

sqlglot/transforms.py
GeorgeSittas: `exp.union` doesn't copy by default it seems, should I change this?

tobymao: yea

GeorgeSittas: done

PR #2465, opened 2023-10-26
 Feat(postgres): support WITH ORDINALITY in table functions 

GeorgeSittas:
Fixes https://github.com/TobikoData/sqlmesh/issues/1619

Reference: https://www.postgresql.org/docs/9.5/sql-select.html

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2486,6 +2486,7 @@ class Table(Expression):
         "format": False,
         "pattern": False,
         "index": False,
+        "offset": False,
     }
 
     @property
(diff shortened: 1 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -2649,11 +2650,7 @@ class Update(Expression):
 
 
 class Values(UDTF):
-    arg_types = {
-        "expressions": True,
-        "ordinality": False,
```
GeorgeSittas, line 2654: This didn't seem to be used anywhere in the codebase and no tests failed after getting rid of it.

PR #2466, opened 2023-10-26
Feat!(hive): add support for ${a:b}-style vars

### LINKED ISSUE #2464: Parsing of HiveConf variables do not succeed

Hive & Spark SQL parsing does not seem to support hive's variable syntax.

https://sparkbyexamples.com/apache-hive/how-to-set-variables-in-hive-scripts/

Ideally, `parse` would be able to recognize the variables that were set and interpolate them into the SQL and parse them correctly.

Here's a script that is currently failing and should succeed:
```
from sqlglot import parse, Dialects
sql = """
set src_table=stage.foo;
set src_path=s3://var/stage/foo;

set dst_table=dw.foo;
set dst_path=s3://var/dw/foo;

create external table ${hiveconf:src_table}
(
	id                  	bigint
)
partitioned by
(
    ds STRING
)
stored as parquet
location '${hiveconf:src_path}';
"""

parsed_sql = parse(sql, read=Dialects.SPARK)
print(repr(parsed_sql))

parsed_sql = parse(sql, read=Dialects.HIVE)
print(repr(parsed_sql))
```

Two alternatives are:
1. Interpolate these variables prior to sending them to `parse`.  Given that these are valid Hive / Spark SQL commands, it feels like sqlglot should be able to handle them.
2. Not use these set statements.  Honestly, this would be better, but they are littered all over the codebase, so removing them is a somewhat significant project.

**Additional context**

This seems related to [this PR](https://github.com/tobymao/sqlglot/pull/1292), in that the `set` commands are now getting parsed correctly.  Unfortunately, the usage of the variables still does not seem to be.

GeorgeSittas:
Fixes #2464

Reference: https://sparkbyexamples.com/apache-hive/how-to-set-variables-in-hive-scripts/#hivevar-variables


sqlglot/parser.py
```diff
@@ -4793,10 +4793,15 @@ def _parse_star(self) -> t.Optional[exp.Expression]:
         return self._parse_placeholder()
 
     def _parse_parameter(self) -> exp.Parameter:
-        wrapped = self._match(TokenType.L_BRACE)
+        self._match(TokenType.L_BRACE)
+
         this = self._parse_var() or self._parse_identifier() or self._parse_primary()
```
tobymao, line 4798: can this be parse_id_var or self._parse_primary()

GeorgeSittas: let me take a look

tobymao: or parse_field() handles it

PR #2467, opened 2023-10-27
Feat!(hive,tsql): move CTEs to top-level, fix T-SQL recursive CTEs

GeorgeSittas:
This PR:

- Moves all CTEs to the top-level expression for Hive, Spark2 and T-SQL, since they don't allow nested CTEs
- Fixes a transpilation issue towards T-SQL related to recursive CTEs (the `RECURSIVE` keyword needs to be omitted)
- Adds support for parsing & generating CTEs in `MERGE` statements

The motivation behind touching `generate` instead of adding new entries in `TRANSFORMS` was mainly to avoid duplicating the same code for ~10 expressions that support CTEs x2 dialects (Hive, T-SQL).

One downside of the current approach is that Hive and T-SQL suffer from an additional perf. hit due to the preemptive `copy` call that happens in `move_ctes_to_top_level` when the input expression is 1) at the root of the AST and 2) supports CTEs.


sqlglot/transforms.py
```diff
@@ -398,6 +398,41 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
...
+    are invalid in those dialects. This transformation ensures all CTEs are moved
+    to the top level so that the final SQL code is valid from a syntax standpoint.
+
+    TODO: handle name clashes whilst moving CTEs (it can get quite tricky & costly).
+    """
+    # This transformation is only applied to the root expression, if it supports CTEs
+    if expression.parent or "with" not in expression.arg_types:
+        return expression
+
+    expression = expression.copy()
```
tobymao, line 417: transforms shouldn't manually copy, this should be done by the caller

tobymao: if you want to be most efficient, you could only copy if you find With Statements

sqlglot/transforms.py
```diff
@@ -398,6 +398,41 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
...
+    top_level_with = expression.args.get("with")
+    for node in expression.find_all(exp.With):
+        if node.parent is expression:
+            continue
+
+        inner_with = node.pop()
+        if not top_level_with:
+            top_level_with = inner_with
+            expression.set("with", top_level_with)
```
tobymao, line 426: what happens if the with alias collides?

sqlglot/transforms.py
GeorgeSittas: Agree, this transformation was initially defined in dialect.py and so the copy was left over since it wasnt supposed to be used in combination with preprocess.

sqlglot/transforms.py
GeorgeSittas: There's no name clash detection in this PR, so it'd shadow any name or if there were duplicate ctes it may crash depending on the engine I guess. This though was one of the core assumptions that allowed this to be simpler because checking for name clashes is more costly since cte names are global. To my understanding though having nested ctes with clashing names should be a rare case and we can always reiterate later?

sqlglot/transforms.py
GeorgeSittas: Addressed both comments

PR #2468, opened 2023-10-27
feat(postgres): support `PARTITION BY`

judahrand:
Currently the Postgres dialect doesn't handle this Postgres extension syntax. It is the same as BigQuery and Drill etc.

https://www.postgresql.org/docs/current/sql-createtable.html


GeorgeSittas (approved):

GeorgeSittas:
Thanks for the PR!

PR #2473, opened 2023-10-27
fix(mysql): str_to_date for datetime

barakalon:
Transpiling `STR_TO_DATE(x, '%Y-%m-%dT%T')` to `CAST(DATE_PARSE(x, '%Y-%m-%dT%T') AS DATE)` is wrong.

From [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_str-to-date):
> STR_TO_DATE() returns a [DATETIME](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) value if the format string contains both date and time parts, or a [DATE](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) or [TIME](https://dev.mysql.com/doc/refman/8.0/en/time.html) value if the string contains only date or time parts

I don't love doing this with a regex at parse time... open to other ideas.

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import re
 import typing as t
 
 from sqlglot import exp, generator, parser, tokens, transforms
@@ -63,7 +62,18 @@ def _date_trunc_sql(self: MySQL.Generator, expression: exp.DateTrunc) -> str:
 
 # All specifiers for time parts (as opposed to date parts)
 # https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format
-TIME_SPECIFIER_PATTERN = re.compile(r"(?<!%)%[fHhIiklprSsT]")
+TIME_SPECIFIERS = {"f", "H", "h", "I", "i", "k", "l", "p", "r", "S", "s", "T"}
+
+
+def _has_time_specifier(date_format: str) -> bool:
+    i = 0
+    while i < len(date_format):
+        if date_format[i] == "%":
+            i += 1
+            if i < len(date_format) and date_format[i] in TIME_SPECIFIERS:
+                return True
+        i += 1
+    return False
 
 
 def _str_to_date(args: t.List) -> exp.StrToDate | exp.StrToTime:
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -60,9 +61,20 @@ def _date_trunc_sql(self: MySQL.Generator, expression: exp.DateTrunc) -> str:
     return f"STR_TO_DATE({concat}, '{date_format}')"
 
 
-def _str_to_date(args: t.List) -> exp.StrToDate:
-    date_format = MySQL.format_time(seq_get(args, 1))
-    return exp.StrToDate(this=seq_get(args, 0), format=date_format)
+# All specifiers for time parts (as opposed to date parts)
+# https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format
+TIME_SPECIFIER_PATTERN = re.compile(r"(?<!%)%[fHhIiklprSsT]")
```
tobymao, line 66: why do you need negative look back?

barakalon: I removed the regex entirely. But it was a failed attempt at handling an escaped `%%`

tobymao (approved):

sqlglot/dialects/mysql.py
```diff
@@ -60,9 +60,31 @@ def _date_trunc_sql(self: MySQL.Generator, expression: exp.DateTrunc) -> str:
...
-    date_format = MySQL.format_time(seq_get(args, 1))
-    return exp.StrToDate(this=seq_get(args, 0), format=date_format)
+# All specifiers for time parts (as opposed to date parts)
+# https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format
+TIME_SPECIFIERS = {"f", "H", "h", "I", "i", "k", "l", "p", "r", "S", "s", "T"}
+
+
+def _has_time_specifier(date_format: str) -> bool:
+    i = 0
+    while i < len(date_format):
```
GeorgeSittas, line 70: I think we can factor out `len(date_format)` as a loop invariant?

GeorgeSittas (approved):
Just one comment, looks good otherwise

sqlglot/dialects/mysql.py
barakalon: ah, sure

Issue #2069, opened 2023-08-16
Snowflake parsing issue


darshan-rathi-df:
**Input Sql**

```
WITH first_activity AS (
  SELECT USER_ID, MIN(TO_TIMESTAMP(EVENT_DATE)) AS first_activity_date
  FROM DF.USER_EVENTS
  GROUP BY USER_ID
)
SELECT COUNT(*) AS active_users_count
FROM first_activity
WHERE DATEDIFF(day, first_activity_date, CURRENT_TIMESTAMP()) > 90
```

I am using sqlglot transpile function in the following manner
```sqlglot.transpile(query, write="snowflake", pretty=True)[0]```

**Output Query**

```
WITH first_activity AS (
  SELECT
    USER_ID,
    MIN(TO_TIMESTAMP(EVENT_DATE)) AS first_activity_date
  FROM DF.USER_EVENTS
  GROUP BY
    USER_ID
)
SELECT
  COUNT(*) AS active_users_count
FROM first_activity
WHERE
  DATEDIFF(, first_activity_date, day) > 90
```

The output query is failing as there is a leading comma in the DATEDIFF function. Moreover I'm not sure why it changed the function signature and parameters altogether.

GeorgeSittas:
Thanks, will look into this soon

GeorgeSittas:
@darshan-rathi-df what's the input dialect? You should specify `read="snowflake"` in `transpile` if it's Snowflake, for example:

```python
>>> sql = """
... WITH first_activity AS (
...   SELECT USER_ID, MIN(TO_TIMESTAMP(EVENT_DATE)) AS first_activity_date
...   FROM DF.USER_EVENTS
...   GROUP BY USER_ID
... )
... SELECT COUNT(*) AS active_users_count
... FROM first_activity
... WHERE DATEDIFF(day, first_activity_date, CURRENT_TIMESTAMP()) > 90
... """
>>>
>>> import sqlglot
>>>
>>> print(sqlglot.transpile(sql, "snowflake", pretty=True)[0])
WITH first_activity AS (
  SELECT
    USER_ID,
    MIN(TO_TIMESTAMP(EVENT_DATE, 'yyyy-mm-DD hh24:mi:ss')) AS first_activity_date
  FROM DF.USER_EVENTS
  GROUP BY
    USER_ID
)
SELECT
  COUNT(*) AS active_users_count
FROM first_activity
WHERE
[...]

PR #2475, opened 2023-10-27
Fix(snowflake): don't add time format in TO_TIMESTAMP if not supplied

GeorgeSittas:
Addresses https://github.com/tobymao/sqlglot/issues/2069#issuecomment-1783424741

For more context, this was first added in: https://github.com/tobymao/sqlglot/commit/8532f465b

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -32,7 +32,7 @@ def _check_int(s: str) -> bool:
 
 
 # from https://docs.snowflake.com/en/sql-reference/functions/to_timestamp.html
-def _parse_to_timestamp(args: t.List) -> t.Union[exp.StrToTime, exp.UnixToTime, exp.Cast]:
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -60,8 +60,8 @@ def _parse_to_timestamp(args: t.List) -> t.Union[exp.StrToTime, exp.UnixToTime]:
     # reduce it using `simplify_literals` first and then check if it's a Literal.
     first_arg = seq_get(args, 0)
     if not isinstance(simplify_literals(first_arg, root=True), Literal):
-        # case: <variant_expr>
-        return format_time_lambda(exp.StrToTime, "snowflake", default=True)(args)
+        # case: <variant_expr> or other expressions such as columns
+        return exp.cast(first_arg, "timestamp")
```
tobymao, line 64: should this be timestrtotime?

tobymao (approved):

sqlglot/dialects/snowflake.py
GeorgeSittas: ah, sure makes sense - changed

PR #2477, opened 2023-10-28
Refactor!: generator now always copies, making transforms much simpler

tobymao:


tobymao:
this is a v19 change

sqlglot/optimizer/simplify.py
```diff
@@ -1070,3 +1067,69 @@ def _flat_simplify(expression, simplifier, root=True):
                 lambda a, b: expression.__class__(this=a, expression=b), operands
             )
     return expression
+
+
+def gen(expression: t.Any) -> str:
+    """Simple pseudo sql generator for quickly generating sortable and uniq strings.
+
+    Sorting and deduping sql is a necessary step for optimization. Calling the actual
+    generator is expensive so we have a bare minimum sql generator here.
```
GeorgeSittas, line 1076: why is calling the generator expensive? can't we set `copy=False` or something? this seems do add quite some complexity

GeorgeSittas:
Haven't checked yet thoroughly but I'd be a bit cautious with the copies in `transform.py` because the assumption before was that each expression would be copied by default so maybe some copies were intentional? Just thinking out loud

sqlglot/optimizer/simplify.py
tobymao: generator is super expensive, so i needed to add a cache

GeorgeSittas: we can't bring it back?

tobymao: it’s not necessary anymore and removes complexity

sqlglot/generator.py
```diff
@@ -2962,15 +2948,6 @@ def _simplify_unless_literal(self, expression: E) -> E:
         if not isinstance(expression, exp.Literal):
             from sqlglot.optimizer.simplify import simplify
 
-            expression = simplify(expression.copy())
+            expression = simplify(expression)
 
         return expression
-
-
-def cached_generator(
```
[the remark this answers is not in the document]
tobymao: correct, needed the cache because sql gen was so expensive

PR #2484, opened 2023-10-30
Feat(tsql): add PERIOD FOR SYSTEM_TIME constraint, SYSTEM_VERSIONING property

Rik-de-Kort:
Add GENERATED ALWAYS AS ROW START/END column constraint, PERIOD FOR SYSTEM_TIME table constraint, and SYSTEM_VERSIONING property, following the [T-SQL spec for CREATE TABLE](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql#full-syntax).

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -403,11 +403,6 @@ class TSQL(Dialect):
             TokenType.END: lambda self: self._parse_command(),
         }
 
-        CONSTRAINT_PARSERS = {
-            **parser.Parser.CONSTRAINT_PARSERS,
-            "PERIOD": lambda self: self._parse_period_for_system_time(),
-        }
-
         LOG_DEFAULTS_TO_LN = True
 
         CONCAT_NULL_OUTPUTS_STRING = True
@@ -584,16 +579,6 @@ class TSQL(Dialect):
                 self._retreat(index)
 
             return super()._parse_if()
-
-        def _parse_period_for_system_time(self) -> exp.PeriodForSystemTimeConstraint:
-            self._match(TokenType.TIMESTAMP_SNAPSHOT)
-            self._match_l_paren()
-
-            start = self._parse_id_var()
-            self._match(TokenType.COMMA)
-            end = self._parse_id_var()
-            self._match_r_paren()
-            return exp.PeriodForSystemTimeConstraint(start=start, end=end)
 
         def _parse_unique(self) -> exp.UniqueColumnConstraint:
             return self.expression(
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -580,6 +580,16 @@ class TSQL(Dialect):
 
             return super()._parse_if()
 
+        def _parse_period_for_system_time(self) -> exp.PeriodForSystemTimeConstraint:
+            self._match(TokenType.TIMESTAMP_SNAPSHOT)
+            self._match_l_paren()
+
+            start = self._parse_id_var()
+            self._match(TokenType.COMMA)
+            end = self._parse_id_var()
+            self._match_r_paren()
+            return exp.PeriodForSystemTimeConstraint(start=start, end=end)
+
         def _parse_unique(self) -> exp.UniqueColumnConstraint:
             return self.expression(
                 exp.UniqueColumnConstraint,
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -403,6 +403,11 @@ class TSQL(Dialect):
             TokenType.END: lambda self: self._parse_command(),
         }
 
+        CONSTRAINT_PARSERS = {
+            **parser.Parser.CONSTRAINT_PARSERS,
+            "PERIOD": lambda self: self._parse_period_for_system_time(),
+        }
+
         LOG_DEFAULTS_TO_LN = True
 
         CONCAT_NULL_OUTPUTS_STRING = True
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1356,10 +1356,7 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
 
 
 class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
-    arg_types = {
-        "start": False,
-        "hidden": False,
-    }
+    arg_types = {"start": False, "hidden": False}
 
 
 # https://dev.mysql.com/doc/refman/8.0/en/create-table.html
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2028,6 +2028,10 @@ class HeapProperty(Property):
     arg_types = {}
 
 
+class HistoryTableProperty(Property):
+    arg_types = {"this": True}
+
+
 class ToTableProperty(Property):
     arg_types = {"this": True}
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1296,8 +1296,9 @@ class ColumnConstraintKind(Expression):
 class AutoIncrementColumnConstraint(ColumnConstraintKind):
     pass
 
+
 class PeriodForSystemTimeConstraint(ColumnConstraintKind):
-    ar_types = {"this": False, "start": True, "end": True}
+    arg_types = {"this": True, "expression": True}
 
 
 class CaseSpecificColumnConstraint(ColumnConstraintKind):
@@ -1355,9 +1356,7 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
 
 
 class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
-    # this:
     arg_types = {
-        "this": False,
         "start": False,
         "hidden": False,
     }
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1296,6 +1296,9 @@ class ColumnConstraintKind(Expression):
 class AutoIncrementColumnConstraint(ColumnConstraintKind):
     pass
 
+class PeriodForSystemTimeConstraint(ColumnConstraintKind):
+    ar_types = {"this": False, "start": True, "end": True}
+
 
 class CaseSpecificColumnConstraint(ColumnConstraintKind):
     arg_types = {"not_": True}
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1356,7 +1356,7 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
 
 
 class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
-    arg_types = {"start": False, "hidden": False}
+    arg_types = {"start": True, "hidden": False}
 
 
 # https://dev.mysql.com/doc/refman/8.0/en/create-table.html
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2254,6 +2254,11 @@ class WithJournalTableProperty(Property):
     arg_types = {"this": True}
 
 
+class WithSystemVersioningProperty(Property):
+    # this -> history table name, expression -> data consistency check
+    arg_types = {"this": True, "expression": False}
+
+
 class Properties(Expression):
     arg_types = {"expressions": True}
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1351,6 +1351,15 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
     }
 
 
+class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
+    # this:
+    arg_types = {
+        "this": False,
+        "start": False,
+        "hidden": False,
+    }
+
+
 # https://dev.mysql.com/doc/refman/8.0/en/create-table.html
 class IndexColumnConstraint(ColumnConstraintKind):
     arg_types = {
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2028,10 +2028,6 @@ class HeapProperty(Property):
     arg_types = {}
 
 
-class HistoryTableProperty(Property):
-    arg_types = {"this": True}
-
-
 class ToTableProperty(Property):
     arg_types = {"this": True}
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -750,7 +750,7 @@ class Generator:
         return f"GENERATED{this} AS {expr}{sequence_opts}"
 
     def generatedasrowcolumnconstraint_sql(
-        self, expression: exp.GeneratedAsIdentityColumnConstraint
+        self, expression: exp.GeneratedAsRowColumnConstraint
     ) -> str:
         start = "START" if expression.args["start"] else "END"
         hidden = " HIDDEN" if expression.args["hidden"] else ""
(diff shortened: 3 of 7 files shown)


sqlglot/tokens.py
```diff
@@ -284,6 +284,7 @@ class TokenType(AutoName):
     PARTITION = auto()
     PARTITION_BY = auto()
     PERCENT = auto()
+    PERIOD = auto()
```
GeorgeSittas, line 287: I don't think this is needed, let's get rid of it.

sqlglot/dialects/tsql.py
```diff
@@ -403,6 +403,11 @@ class Parser(parser.Parser):
             TokenType.END: lambda self: self._parse_command(),
         }
 
+        CONSTRAINT_PARSERS = {
+            **parser.Parser.CONSTRAINT_PARSERS,
+            "PERIOD": lambda self: self._parse_period_for_system_time(),
+        }
```
GeorgeSittas, line 409: Since this doesn't clash with other constraints, we can move it to the base parser. The philosophy is that whenever we can, we add logic to the base classes so that the dialect they define (or "SQLGlot" dialect) is broader.

sqlglot/dialects/tsql.py
```diff
@@ -580,6 +585,16 @@ def _parse_if(self) -> t.Optional[exp.Expression]:
...
             return super()._parse_if()
 
+        def _parse_period_for_system_time(self) -> exp.PeriodForSystemTimeConstraint:
+            self._match(TokenType.TIMESTAMP_SNAPSHOT)
+            self._match_l_paren()
+
+            start = self._parse_id_var()
+            self._match(TokenType.COMMA)
+            end = self._parse_id_var()
+            self._match_r_paren()
```
GeorgeSittas, line 595: Ditto about moving this whole method in the base class. We can also replace these lines with:

```python
self._match(TokenType.TIMESTAMP_SNAPSHOT)
id_vars = self._parse_wrapped_id_vars
return exp.PeriodForSystemTimeConstraint(start=seq_get(id_vars, 0), end=seq_get(id_vars, 1))
```

sqlglot/expressions.py
```diff
@@ -1296,6 +1296,9 @@ class ColumnConstraintKind(Expression):
 class AutoIncrementColumnConstraint(ColumnConstraintKind):
     pass
 
+class PeriodForSystemTimeConstraint(ColumnConstraintKind):
+    ar_types = {"this": False, "start": True, "end": True}
```
GeorgeSittas, line 1300: Typo on the `ar_types` - it needs to be `arg_types`.

I understand that we need two children expressions for this: one for `start` and one for `end`. If this is the case, let's do `arg_types = {"this": True, "expression": True}` and update the the parser method accordingly.

sqlglot/parser.py
```diff
@@ -3875,6 +3875,14 @@ def _parse_generated_as_identity(
             this = self.expression(exp.GeneratedAsIdentityColumnConstraint, this=True)
 
         self._match(TokenType.ALIAS)
+
+        if self._match_text_seq("ROW"):
+            start = self._match_text_seq("START")
+            if not start:
+                self._match_text_seq("END")
```
GeorgeSittas, line 3882:
```suggestion
                self._match(TokenType.END)
```

sqlglot/parser.py
```diff
@@ -3875,6 +3875,14 @@ def _parse_generated_as_identity(
...
+            hidden = self._match_text_seq("HIDDEN")
+            return self.expression(exp.GeneratedAsRowColumnConstraint, this=True, start=start, hidden=hidden)
```
GeorgeSittas, line 3884:
```suggestion
            return self.expression(exp.GeneratedAsRowColumnConstraint, start=start, hidden=hidden)
```

sqlglot/expressions.py
```diff
@@ -1351,6 +1354,15 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
     }
 
 
+class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
+    # this:
+    arg_types = {
+        "this": False,
```
GeorgeSittas, line 1360: We don't need `this` here, can just keep `start` and `hidden`

GeorgeSittas (commented):
Thanks for the PR @Rik-de-Kort -- left some comments. We'll probably need a few iterations to get this to the finish line but it shouldn't be too hard.

Let's also include documentation for the added clauses, e.g. in the PR description, so it's easy to verify the validity of these changes. Also, for future PRs make sure to check out the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) guideline so that the commit titles are formatted accordingly.

sqlglot/tokens.py
tobymao, line 287: i don't think this is necessary

sqlglot/parser.py
```diff
@@ -4047,6 +4060,14 @@ def _parse_foreign_key(self) -> exp.ForeignKey:
     def _parse_primary_key_part(self) -> t.Optional[exp.Expression]:
         return self._parse_field()
 
+    def _parse_period_for_system_time(self) -> exp.PeriodForSystemTimeConstraint:
+        self._match(TokenType.TIMESTAMP_SNAPSHOT)
+
+        id_vars = self._parse_wrapped_id_vars()
+        return exp.PeriodForSystemTimeConstraint(
+            this=seq_get(id_vars, 0), expression=seq_get(id_vars, 1)
+        )
```
GeorgeSittas, line 4069:
```suggestion
        return self.expression(
            exp.PeriodForSystemTimeConstraint,
            this=seq_get(id_vars, 0),
            expression=seq_get(id_vars, 1),
        )
```

sqlglot/expressions.py
```diff
@@ -1351,6 +1355,13 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
     }
 
 
+class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
+    arg_types = {
+        "start": False,
+        "hidden": False,
+    }
```
GeorgeSittas, line 1362: Nit:
```suggestion
    arg_types = {"start": False, "hidden": False}
```

GeorgeSittas (commented):
Re: open questions

> Should I be returning a t.Optional[exp.Expression] from _parse_period_for_system_time?

No I think it's ok as it is now.

> Should the generated as parsers be defined on the dialect parser (i.e. in tsql.py) or the generic parser?

Already answered in a comment. For posterity:

Since this doesn't clash with other constraints, we can move it to the base parser. The philosophy is that whenever we can, we add logic to the base classes so that the dialect they define (or "SQLGlot" dialect) is broader.

Rik-de-Kort:
Thanks for the prompt and helpful reviews! I think I got the table option bit to parse correctly now as well, but I'm unsure what kind of expression I should be returning. Should `WithSystemVersioningProperty` have a custom property `data_consistency_check` that is True or False, or should there be a custom Expression for `DATA_CONSISTENCY_CHECK` with `this=True` for `ON`? And if the latter, what subclass of Expression do I use for that?

GeorgeSittas:
> Thanks for the prompt and helpful reviews! I think I got the table option bit to parse correctly now as well, but I'm unsure what kind of expression I should be returning. Should `WithSystemVersioningProperty` have a custom property `data_consistency_check` that is True or False, or should there be a custom Expression for `DATA_CONSISTENCY_CHECK` with `this=True` for `ON`? And if the latter, what subclass of Expression do I use for that?

It seems like the `ON (...)` part is just a csv of property assignments like `foo = bla`? If that's the case then we don't have to create a separate expression to represent it - can probably just call `self._parse_wrapped(self._parse_properties)` and store the `Properties` object in an argument of the `WithSystemVersioningProperty` expression. Does that make sense?

sqlglot/parser.py
```diff
@@ -1523,6 +1525,19 @@ def _parse_volatile_property(self) -> exp.VolatileProperty | exp.StabilityProper
 
         return self.expression(exp.StabilityProperty, this=exp.Literal.string("VOLATILE"))
 
+    def _parse_system_versioning_property(self) -> exp.WithSystemVersioningProperty:
+        self._match_pair(TokenType.EQ, TokenType.ON)
+        self._match_l_paren()
+        self._match_texts(["HISTORY_TABLE"])
+        self._match(TokenType.EQ)
+        this = exp.WithSystemVersioningProperty(this=(self._parse_table_parts()))
```
tobymao, line 1533: get rid of the extra parenthesis

sqlglot/parser.py
```diff
@@ -1523,6 +1525,19 @@ def _parse_volatile_property(self) -> exp.VolatileProperty | exp.StabilityProper
 
         return self.expression(exp.StabilityProperty, this=exp.Literal.string("VOLATILE"))
 
+    def _parse_system_versioning_property(self) -> exp.WithSystemVersioningProperty:
+        self._match_pair(TokenType.EQ, TokenType.ON)
+        self._match_l_paren()
+        self._match_texts(["HISTORY_TABLE"])
```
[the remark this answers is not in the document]
tobymao: you should switch this to self._match_text_seq("HISTORY_TABLE")

sqlglot/parser.py
```diff
@@ -1523,6 +1525,19 @@ def _parse_volatile_property(self) -> exp.VolatileProperty | exp.StabilityProper
...
+        self._match(TokenType.EQ)
+        this = exp.WithSystemVersioningProperty(this=(self._parse_table_parts()))
+        if self._match(TokenType.COMMA):
+            self._match_texts(["DATA_CONSISTENCY_CHECK"])
```
tobymao, line 1535: same

sqlglot/parser.py
```diff
@@ -1523,6 +1525,19 @@ def _parse_volatile_property(self) -> exp.VolatileProperty | exp.StabilityProper
...
+            self._match(TokenType.EQ)
+            this.set("expression", self._match_texts(["ON", "OFF"]))
```
tobymao, line 1537: change this to a tuple ("ON", "OFF")

sqlglot/parser.py
GeorgeSittas, line 1533: instantiation of expressions should generally happen through `self.expression`

Rik-de-Kort:
> @Rik-de-Kort can you also share some docs on the syntax spec etc for these added clauses?

https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16#full-syntax

Rik-de-Kort:
> It seems like the `ON (...)` part is just a csv of property assignments like `foo = bla`? If that's the case then we don't have to create a separate expression to represent it - can probably just call `self._parse_wrapped(self._parse_properties)` and store the `Properties` object in an argument of the `WithSystemVersioningProperty` expression. Does that make sense?

I had a go at this for the last hour or so, but the parser gets stuck on the fact that there is a table in the `HISTORY_TABLE` field.  I tried hacking in a parsing a Table if we find an Identifier into `_parse_field`, but that turns out terrible (see previous commit https://github.com/tobymao/sqlglot/pull/2484/commits/fab5272d024d981d4650c0812b5668cb7e235c2f) and wreaks all kinds of havok in the test suite. Since the spec is quite explicit in the things that can go here, I think I would like to stick with the explicit "hand-rolled" parser.

Finally, my `mypy` pre-commit hook seems broken. Cannot find implementation for "sqlglot._version" even though I have properly pip installed everything.

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1
sqlglot/__init__.py:56: error: Cannot find implementation or library stub for module named "sqlglot._version"  [import-not-found]
sqlglot/__init__.py:56: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 137 source files)
```

GeorgeSittas:
Will review again shortly, thanks!

sqlglot/generator.py
```diff
@@ -748,6 +749,18 @@ def generatedasidentitycolumnconstraint_sql(
 
         return f"GENERATED{this} AS {expr}{sequence_opts}"
 
+    def generatedasrowcolumnconstraint_sql(
+        self, expression: exp.GeneratedAsIdentityColumnConstraint
```
GeorgeSittas, line 753:
```suggestion
        self, expression: exp.GeneratedAsRowColumnConstraint
```

sqlglot/expressions.py
```diff
@@ -1351,6 +1355,10 @@ class GeneratedAsIdentityColumnConstraint(ColumnConstraintKind):
     }
 
 
+class GeneratedAsRowColumnConstraint(ColumnConstraintKind):
+    arg_types = {"start": False, "hidden": False}
```
GeorgeSittas, line 1359: It seems to me that `start` is mandatory so this needs to become `True`

```suggestion
    arg_types = {"start": True, "hidden": False}
```

sqlglot/generator.py
```diff
@@ -748,6 +749,18 @@ def generatedasidentitycolumnconstraint_sql(
...
+    ) -> str:
+        start = "START" if expression.args["start"] else "END"
+        hidden = " HIDDEN" if expression.args["hidden"] else ""
```
GeorgeSittas, line 756:
```suggestion
        hidden = " HIDDEN" if expression.args.get("hidden") else ""
```

sqlglot/generator.py
```diff
@@ -748,6 +749,18 @@ def generatedasidentitycolumnconstraint_sql(
...
+        return f"GENERATED ALWAYS AS ROW {start}{hidden}"
+
+    def periodforsystemtimeconstraint_sql(
+        self, expression: exp.PeriodForSystemTimeConstraint
+    ) -> str:
+        return f"PERIOD FOR SYSTEM_TIME ({expression.this}, {expression.expression})"
```
GeorgeSittas, line 762:
```suggestion
        return f"PERIOD FOR SYSTEM_TIME ({self.sql(expression, 'this')}, {self.sql(expression, 'expression')})"
```

sqlglot/generator.py
```diff
@@ -1255,6 +1268,20 @@ def withdataproperty_sql(self, expression: exp.WithDataProperty) -> str:
             statistics_sql = f" AND {'NO ' if not statistics else ''}STATISTICS"
         return f"{data_sql}{statistics_sql}"
 
+    def withsystemversioningproperty_sql(self, expression: exp.WithSystemVersioningProperty) -> str:
+        sql = f"WITH(SYSTEM_VERSIONING=ON"
```
GeorgeSittas, line 1272:
```suggestion
        sql = "WITH(SYSTEM_VERSIONING=ON"
```

sqlglot/generator.py
```diff
@@ -1255,6 +1268,20 @@ def withdataproperty_sql(self, expression: exp.WithDataProperty) -> str:
...
+        if expression.this is None:
+            return f"{sql})"
```
GeorgeSittas, line 1274: This branch is inconsistent with the fact that `this` is mandatory in the definition of `WithSystemVersioningProperty`. You need to either have `this` be optional (i.e. map to `False`) or rewrite this section because it seems like dead code right now.

GeorgeSittas: It seems like it's not tested too - let's make sure we test all of the new execution paths

GeorgeSittas:
@Rik-de-Kort this seems to be almost done so I'll take it to the finish line - will ping you once I make the update so you can inspect the final changes.

Thanks for the contribution!

Rik-de-Kort:
@GeorgeSittas My pleasure, and thanks for the final push! I think we have some timezone differences so wasn't able to respond until now.

PR #2493, opened 2023-10-31
feat(optimizer): more date function coercion

barakalon:
More date function stuff because MySQL is so lax about date coercion and presto is so strict.

I'm not super happy about mutations in `annotate_types`... but it's convenient place to do it, since we already do the required type checks there.


sqlglot/optimizer/annotate_types.py
```diff
@@ -511,22 +521,31 @@ def _annotate_by_args(
...
 
         if expression.this.type.this in exp.DataType.TEXT_TYPES:
-            datatype = _coerce_literal_and_interval(expression.this, expression.interval())
-        elif (
-            expression.this.type.is_type(exp.DataType.Type.DATE)
-            and expression.text("unit").lower() not in DATE_UNITS
-        ):
-            datatype = exp.DataType.Type.DATETIME
+            datatype = _coerce_date_literal(expression.this, expression.unit)
+        elif expression.this.type.is_type(*exp.DataType.TEMPORAL_TYPES):
```
GeorgeSittas, line 531:
```suggestion
        elif expression.this.type.this in exp.DataType.TEMPORAL_TYPES:
```

sqlglot/optimizer/annotate_types.py
```diff
@@ -511,22 +521,31 @@ def _annotate_by_args(
...
+            datatype = exp.DataType.Type.UNKNOWN
 
         self._set_type(expression, datatype)
         return expression
 
+    def _annotate_datediff(self, expression: exp.DateDiff) -> exp.DateDiff:
+        self._annotate_args(expression)
+
+        for e in (expression.this, expression.expression):
+            if not e.type.is_type(*exp.DataType.TEMPORAL_TYPES):
```
GeorgeSittas, line 543:
```suggestion
            if not e.type.this in exp.DataType.TEMPORAL_TYPES:
```

tobymao:
is it a hassle to do this in canonicalize?

GeorgeSittas (commented):
FWIW mutations in `annotate_types` means that the semantics may be changed for SQLMesh queries as well since we use it there, so it might be best to move them to `canonicalize.py` or something.

barakalon:
ok, moved to canonicalize

tobymao (approved):

PR #2495, opened 2023-10-31
Fix(teradata): support `**` for exponent

hsheth2:
See https://docs.teradata.com/r/SQL-Functions-Operators-Expressions-and-Predicates/June-2017/Arithmetic-Trigonometric-Hyperbolic-Operators/Functions

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -36,7 +36,7 @@ class Teradata(Dialect):
         # https://docs.teradata.com/r/SQL-Functions-Operators-Expressions-and-Predicates/June-2017/Arithmetic-Trigonometric-Hyperbolic-Operators/Functions
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/teradata.py
```diff
@@ -116,6 +118,16 @@ class Parser(parser.Parser):
             "TRANSLATE": lambda self: self._parse_translate(self.STRICT_CAST),
         }
 
+        EXPONENT = {
```
tobymao, line 121: now that we have to dialects that have this, can you refactor this out of postgres, and move it to the base parser?

sqlglot/tokens.py
```diff
@@ -56,6 +56,7 @@ class TokenType(AutoName):
     SESSION_PARAMETER = auto()
     DAMP = auto()
     XOR = auto()
+    DOUBLE_STAR = auto()
```
tobymao, line 59: can you call this DSTAR as that's more consistent with the patterwe have

PR #2509, opened 2023-11-03
Fix: unify named and unnamed constraint parsing

Rik-de-Kort:
`CREATE TABLE "test" ("name" CHAR (7) NOT NULL, "internal_id" VARCHAR(10) NOT NULL, UNIQUE NONCLUSTERED ("internal_id" ASC))` fails to parse whereas `CREATE TABLE "test" ("name" CHAR (7) NOT NULL, "internal_id" VARCHAR(10) NOT NULL, Constraint CTest UNIQUE NONCLUSTERED ("internal_id" ASC))` does not due to the former not parsing multiple constraints like the latter.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1421,6 +1421,10 @@ class Constraint(Expression):
     arg_types = {"this": True, "expressions": True}
 
 
+class UnnamedConstraint(Expression):
+    arg_types = {"this": False, "expressions": True}
+
+
 class Delete(Expression):
     arg_types = {
         "with": False,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4024,19 +4024,23 @@ class Parser(metaclass=_Parser):
         return this
 
     def _parse_constraint(self) -> t.Optional[exp.Expression]:
-        if not self._match(TokenType.CONSTRAINT):
-            return self._parse_unnamed_constraint(constraints=self.SCHEMA_UNNAMED_CONSTRAINTS)
-
-        this = self._parse_id_var()
+        index = self._index
+        this = self._parse_id_var() if self._match(TokenType.CONSTRAINT) else None
         expressions = []
-
         while True:
             constraint = self._parse_unnamed_constraint() or self._parse_function()
             if not constraint:
                 break
             expressions.append(constraint)
 
-        return self.expression(exp.Constraint, this=this, expressions=expressions)
+        if not expressions:
+            self._retreat(index)
+            return None
+        return self.expression(
+            exp.UnnamedConstraint if this is None else exp.Constraint,
+            this=this,
+            expressions=expressions,
+        )
 
     def _parse_unnamed_constraint(
         self, constraints: t.Optional[t.Collection[str]] = None
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -1453,3 +1453,11 @@ FROM OPENJSON(@json) WITH (
                 "spark": "SET count = (SELECT COUNT(1) FROM x)",
             },
         )
+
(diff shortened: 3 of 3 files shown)


Rik-de-Kort:
Some stuff that is still missing:
- [ ] Parsing and generating `inline` constraint properly
- [ ] Figure out what is going on with MySQL `autoincrement` getting dropped on the floor
- [ ] Figure out why "MONTHS" is getting dropped in PySpark

GeorgeSittas:
Not sure why we need to parse multiple unnamed constraints to solve this. Is this the only use-case where this happens?

I think we can represent the `UNIQUE NONCLUSTERED` constraint by nesting a nonclustered one inside of the unique. For example, the following changes in tsql.py allow parsing & generation of the constraints:

```diff
+++ b/sqlglot/dialects/tsql.py
@@ -580,12 +580,12 @@ class TSQL(Dialect):
             return super()._parse_if()

         def _parse_unique(self) -> exp.UniqueColumnConstraint:
-            return self.expression(
-                exp.UniqueColumnConstraint,
-                this=None
-                if self._curr and self._curr.text.upper() in {"CLUSTERED", "NONCLUSTERED"}
-                else self._parse_schema(self._parse_id_var(any_token=False)),
-            )
+            if self._match_texts({"CLUSTERED", "NONCLUSTERED"}):
+                this = self.CONSTRAINT_PARSERS[self._prev.text.upper()](self)
+            else:
+                this = self._parse_schema(self._parse_id_var(any_token=False))
+
+            return self.expression(exp.UniqueColumnConstraint, this=this)
```

sqlglot/expressions.py
```diff
@@ -1421,6 +1421,10 @@ class Constraint(Expression):
     arg_types = {"this": True, "expressions": True}
 
 
+class UnnamedConstraints(Expression):
+    arg_types = {"this": False, "expressions": True}
```
GeorgeSittas, line 1425: FYI `this` is unused here. @Rik-de-Kort it seems like there might be some confusion around expression arguments, let me know if it's unclear and we can chat on Slack but in general `this` is not special so you don't need to add it everywhere.

sqlglot/parser.py
```diff
@@ -3993,9 +3990,11 @@ def _parse_generated_as_identity(
 
         return this
 
-    def _parse_inline(self) -> exp.InlineLengthColumnConstraint:
+    def _parse_inline(self) -> t.Optional[exp.InlineLengthColumnConstraint]:
         self._match_text_seq("LENGTH")
-        return self.expression(exp.InlineLengthColumnConstraint, this=self._parse_bitwise())
+        if self._match(TokenType.NUMBER):
+            return self.expression(exp.InlineLengthColumnConstraint, this=self._prev.text)
```
GeorgeSittas, line 3996: What prompted the changes to this method?

GeorgeSittas (commented):
Left a couple of additional comments but in general the approach looks more complex than might be needed.

cc @tobymao

sqlglot/parser.py
Rik-de-Kort: A test in identity.sql was failing ( `CREATE TABLE foo (baz DATE FORMAT 'YYYY/MM/DD' TITLE 'title' INLINE LENGTH 1 COMPRESS ('a', 'b')`) which, upon reading the [spec](https://www.ibm.com/docs/en/db2-for-zos/13?topic=statements-create-table), seemed like this was the thing to fix.

Rik-de-Kort:
Overall I think our discussion here and on slack clearly points to this PR taking the wrong approach. I'll revert the changes and introduce more direct parsing of the UNIQUE NONCLUSTERED constraint.

Rik-de-Kort:
replaced by #2518

PR #2513, opened 2023-11-04
fix!: transpile division

barakalon:
Ok, here's a swing.

Check out the test matrix in `test_dialect` - I think that explains the behavior.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -17,6 +17,15 @@ from sqlglot.trie import new_trie
 B = t.TypeVar("B", bound=exp.Binary)
 
 
+def _cast_div_operands_sql(self: Generator, expression: exp.Div) -> str:
+    l, r = expression.left, expression.right
+
+    if not l.is_type(*exp.DataType.FLOAT_TYPES) and not r.is_type(*exp.DataType.FLOAT_TYPES):
+        l.replace(exp.cast(l.copy(), to=exp.DataType.Type.DOUBLE))
+
+    return self.binary(expression, "/")
+
+
 class Dialects(str, Enum):
     DIALECT = ""
 
@@ -135,6 +144,11 @@ class _Dialect(type):
 
         klass.generator_class.can_identify = klass.can_identify
 
+        if klass.TYPED_DIVISION:
+            klass.parser_class.FACTOR = {**klass.parser_class.FACTOR, TokenType.SLASH: exp.TypedDiv}
+            klass.generator_class.TRANSFORMS[exp.Div] = _cast_div_operands_sql
+            klass.generator_class.TRANSFORMS[exp.TypedDiv] = lambda self, e: self.binary(e, "/")
+
         return klass
 
 
@@ -176,6 +190,9 @@ class Dialect(metaclass=_Dialect):
     # Indicates the default null ordering method to use if not explicitly set
     # Options are: "nulls_are_small", "nulls_are_large", "nulls_are_last"
     NULL_ORDERING = "nulls_are_small"
+
+    # Whether the behavior of a / b depends on the types of a and b
+    TYPED_DIVISION = False
 
     DATE_FORMAT = "'%Y-%m-%d'"
     DATEINT_FORMAT = "'%Y%m%d'"
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -238,6 +238,9 @@ class Expression(metaclass=_Expression):
             dtype = DataType.build(dtype)
         self._type = dtype  # type: ignore
 
+    def is_type(self, *dtypes) -> bool:
+        return self.type is not None and self.type.is_type(*dtypes)
+
     @property
     def meta(self) -> t.Dict[str, t.Any]:
         if self._meta is None:
@@ -3881,6 +3884,10 @@ class BitwiseXor(Binary):
 
 
 class Div(Binary):
+    pass
+
+
+class TypedDiv(Binary):
     pass
 
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2591,6 +2591,15 @@ class Generator:
     def div_sql(self, expression: exp.Div) -> str:
         return self.binary(expression, "/")
 
+    def typeddiv_sql(self, expression: exp.TypedDiv) -> str:
+        l, r = expression.left, expression.right
+
+        if l.is_type(*exp.DataType.INTEGER_TYPES) and r.is_type(*exp.DataType.INTEGER_TYPES):
+            l.replace(exp.cast(l.copy(), to=exp.DataType.Type.BIGINT))
+            r.replace(exp.cast(r.copy(), to=exp.DataType.Type.BIGINT))
+
+        return self.binary(expression, "/")
+
     def overlaps_sql(self, expression: exp.Overlaps) -> str:
         return self.binary(expression, "OVERLAPS")
 
--- a/sqlglot/optimizer/annotate_types.py
+++ b/sqlglot/optimizer/annotate_types.py
@@ -165,6 +165,7 @@ class TypeAnnotator(metaclass=_TypeAnnotator):
         exp.DataType.Type.DOUBLE: {
             exp.ApproxQuantile,
             exp.Avg,
+            exp.Div,
             exp.Exp,
             exp.Ln,
             exp.Log,
@@ -272,6 +273,7 @@ class TypeAnnotator(metaclass=_TypeAnnotator):
         exp.Slice: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.UNKNOWN),
         exp.Sum: lambda self, e: self._annotate_by_args(e, "this", "expressions", promote=True),
         exp.TryCast: lambda self, e: self._annotate_with_type(e, e.args["to"]),
+        exp.TypedDiv: lambda self, e: self._annotate_typeddiv(e),
         exp.VarMap: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.MAP),
     }
 
@@ -530,4 +532,19 @@ class TypeAnnotator(metaclass=_TypeAnnotator):
         else:
             self._set_type(expression, exp.DataType.Type.UNKNOWN)
 
+        return expression
+
+    def _annotate_typeddiv(self, expression: exp.TypedDiv) -> exp.TypedDiv:
+        self._annotate_args(expression)
+
+        l, r = expression.left, expression.right
+        l_is_int = l.type and l.type.this in exp.DataType.INTEGER_TYPES
+        r_is_int = r.type and r.type.this in exp.DataType.INTEGER_TYPES
+
+        if l_is_int and r_is_int:
+            datatype = exp.DataType.Type.BIGINT
+        else:
+            datatype = exp.DataType.Type.DOUBLE
+
+        self._set_type(expression, datatype)
         return expression
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -1123,6 +1123,7 @@ GEN_MAP = {
     exp.Sub: lambda e: _binary(e, "-"),
     exp.Subquery: lambda e: f"({gen(e.args.values())})",
     exp.Table: lambda e: gen(e.args.values()),
+    exp.TypedDiv: lambda e: _binary(e, "/"),
     exp.Var: lambda e: e.name,
 }
 
--- a/tests/dialects/test_dialect.py
+++ b/tests/dialects/test_dialect.py
@@ -7,6 +7,7 @@ from sqlglot import (
     ParseError,
     TokenError,
     UnsupportedError,
+    exp,
     parse_one,
 )
 from sqlglot.dialects import Hive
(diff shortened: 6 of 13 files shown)


barakalon:
The main risk:

Transpiling `x / y` from dialect A to B, where:
- A is correctly marked as TYPED_DIVISION = True
- B is incorrectly left as TYPED_DIVISION = False
- x and y both are annotated with INT

OR

Transpiling `x / y` from dialect A to B, where:
- A is correctly left as TYPED_DIVISION = False
- B is incorrectly marked as TYPED_DIVISION = True

Although the second possibility seems less likely

barakalon:
Options:
1. Be diligent about marking all the dialects correctly
2. Remove the default behavior of casting while generating SQL for TypedDiv, and instead only turn that on for select dialects. Its a little clunkier (requires two flags), but then we only make active mistakes, not passive mistakes.

tobymao:
i'd much prefer to go with option 1 and nip this in the bud.

it's not quite as risky as IDENTIY A -> A producing wrong results

sqlglot/dialects/dialect.py
```diff
@@ -135,6 +144,11 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
 
         klass.generator_class.can_identify = klass.can_identify
 
+        if klass.TYPED_DIVISION:
+            klass.parser_class.FACTOR = {**klass.parser_class.FACTOR, TokenType.SLASH: exp.TypedDiv}
+            klass.generator_class.TRANSFORMS[exp.Div] = _cast_div_operands_sql
```
tobymao, line 149: let's not do this and just override the base generator directly, reference the constant TYPED_DIVISION there

sqlglot/dialects/dialect.py
```diff
@@ -135,6 +144,11 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
 
         klass.generator_class.can_identify = klass.can_identify
 
+        if klass.TYPED_DIVISION:
+            klass.parser_class.FACTOR = {**klass.parser_class.FACTOR, TokenType.SLASH: exp.TypedDiv}
```
tobymao, line 148: i'm not sure how easy it is to do this for the parser because it's a constant

sqlglot/dialects/dialect.py
```diff
@@ -177,6 +191,9 @@ class Dialect(metaclass=_Dialect):
     # Options are: "nulls_are_small", "nulls_are_large", "nulls_are_last"
     NULL_ORDERING = "nulls_are_small"
 
+    # Whether the behavior of a / b depends on the types of a and b
```
tobymao, line 194: can you add that by False means a /b is always a double

sqlglot/expressions.py
```diff
@@ -238,6 +238,9 @@ def type(self, dtype: t.Optional[DataType | DataType.Type | str]) -> None:
             dtype = DataType.build(dtype)
         self._type = dtype  # type: ignore
 
+    def is_type(self, *dtypes) -> bool:
+        return self.type is not None and self.type.is_type(*dtypes)
```
tobymao, line 242: self.type and self.type.is_type(*dtypes)

sqlglot/generator.py
```diff
@@ -2591,6 +2591,15 @@ def safedpipe_sql(self, expression: exp.SafeDPipe) -> str:
     def div_sql(self, expression: exp.Div) -> str:
         return self.binary(expression, "/")
 
+    def typeddiv_sql(self, expression: exp.TypedDiv) -> str:
+        l, r = expression.left, expression.right
+
+        if l.is_type(*exp.DataType.INTEGER_TYPES) and r.is_type(*exp.DataType.INTEGER_TYPES):
```
tobymao, line 2597: right, so this cast should only happen if TYPE_DIVISION = False, as opposed to overriding the constants in dialect.py

sqlglot/optimizer/simplify.py
```diff
@@ -1123,6 +1123,7 @@ def gen(expression: t.Any) -> str:
     exp.Sub: lambda e: _binary(e, "-"),
     exp.Subquery: lambda e: f"({gen(e.args.values())})",
     exp.Table: lambda e: gen(e.args.values()),
+    exp.TypedDiv: lambda e: _binary(e, "/"),
```
tobymao, line 1126: surprised you found this

tests/dialects/test_dialect.py
```diff
@@ -1442,6 +1443,48 @@ def test_operators(self):
...
+                "oracle": "a / b",
+                "postgres": "CAST(a AS DOUBLE PRECISION) / b",
+                "presto": "CAST(a AS DOUBLE) / b",
+                "snowflake": "a / b",
+                "sqlite": "CAST(a AS REAL) / b",
+            },
+        )
+
+    def test_div(self):
+        typed_div = exp.TypedDiv(this=exp.column("a"), expression=exp.column("b"))
```
tobymao, line 1462: i guess we should add a new builder to support column // column !

tests/dialects/test_dialect.py
```diff
@@ -1442,6 +1443,48 @@ def test_operators(self):
...
+
+        for expression, types, dialect, expected in [
+            (typed_div, (None, None), typed_div_dialect, "a / b"),
+            (typed_div, (None, None), div_dialect, "a / b"),
+            (div, (None, None), typed_div_dialect, "CAST(a AS DOUBLE) / b"),
+            (div, (None, None), div_dialect, "a / b"),
+            (typed_div, (INT, INT), typed_div_dialect, "a / b"),
+            (typed_div, (INT, INT), div_dialect, "CAST(a AS BIGINT) / CAST(b AS BIGINT)"),
+            (div, (INT, INT), typed_div_dialect, "CAST(a AS DOUBLE) / b"),
+            (div, (INT, INT), div_dialect, "a / b"),
```
tobymao, line 1477: what about (INT, FLOAT) tests

tests/dialects/test_duckdb.py
```diff
@@ -536,7 +536,7 @@ def test_time(self):
             write={
                 "bigquery": "UNIX_TO_TIME(x / 1000)",
                 "duckdb": "TO_TIMESTAMP(x / 1000)",
-                "presto": "FROM_UNIXTIME(x / 1000)",
+                "presto": "FROM_UNIXTIME(CAST(x AS DOUBLE) / 1000)",
```
tobymao, line 539: interesting, was this a bug

tobymao:
```
    A is correctly marked as TYPED_DIVISION = True
    B is incorrectly left as TYPED_DIVISION = False
    x and y both are annotated with INT
```

so if A is marked as typed_division, then it will produce 3 / 2 == TypedDiv,

If B is left as TypedDivision False, doesn't that simply mean when you transpile it, will become 

CAST(3 / 2 AS BIGINT), this is not incorrect, just unnecessary, because you've preserved the semantics of A?

However, if you parse B, 3 / 2 -> Div when it should have been TypedDiv, then when you transpile to A, it will produce cast(3 as double) / 2. B should have produced 1, but it got transpiled to something that produces 1.5.

sqlglot/dialects/dialect.py
barakalon: The Generator doesn't have a reference to the Dialect.
In @GeorgeSittas 's PR we added the flag to both the Generator and the Parser. Happed to switch to that, too.

sqlglot/expressions.py
barakalon: mypy doesn't like that because the return type is `t.Optional[bool]`. Quite annoying. Do you know a way around that?

sqlglot/optimizer/simplify.py
barakalon: find in project > "exp.Div"

tests/dialects/test_dialect.py
barakalon: I think that'd be IntDiv, not TypedDiv?

tests/dialects/test_dialect.py
barakalon: I can add those too, although I'm not sure it will add much more coverage.
But I think it's easier to just add a test than work it out in my head lol.

sqlglot/dialects/dialect.py
tobymao: i believe theres magic. if you define a constant in generator, it will auto propagate the dialect constants there, so yea let’s add a constant to generator

sqlglot/expressions.py
tobymao: ah…

tests/dialects/test_duckdb.py
barakalon: No, because `FROM_UNIXTIME` seems to cast to integer.

barakalon:
Ah, yes. OK, that's a third bad scenario.

tests/dialects/test_dialect.py
tobymao: i guess that’s ok, python and sql are not exactly the same

barakalon: I think that'd be counter intuitive. 
Also, seems like `__floordiv__` already does IntDiv lol

tobymao: ah ok

GeorgeSittas:
Dialects where `1 / 2` yields `0.5`:

- BigQuery
- ClickHouse
- DataBricks
- DuckDB
- Hive
- MySQL
- [Oracle](https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/arithmetic-functions.html#operator_divide)
- Snowflake
- Spark
- Starrocks (checked by Barak)

Dialects where `1 / 2` yields `0`:
- [Drill](https://stackoverflow.com/questions/41040212/apache-drill-sql-division-fails-when-numerator-denominator)
- Postgres
- Presto
- Redshift
- SQLite
- [Teradata](https://docs.teradata.com/r/SQL-Functions-Operators-Expressions-and-Predicates/June-2017/Arithmetic-Trigonometric-Hyperbolic-Operators/Functions/Binary-Arithmetic-Result-Data-Types/Integer-Division-and-Truncation)
- Trino
- T-SQL

sqlglot/generator.py
```diff
@@ -2591,6 +2591,15 @@ def safedpipe_sql(self, expression: exp.SafeDPipe) -> str:
...
+            l.replace(exp.cast(l.copy(), to=exp.DataType.Type.BIGINT))
+            r.replace(exp.cast(r.copy(), to=exp.DataType.Type.BIGINT))
```
GeorgeSittas, line 2599: If I understand this correctly, e.g. `x / y` is parsed into a `TypedDiv` when reading Postgres and so when transpiling it back we just cast the x and y to ints? What about generating towards MySQL which doesn't truncate then? It doesn't seem like the casts reflect that MySQL should also truncate?

sqlglot/dialects/dialect.py
```diff
@@ -17,6 +17,15 @@
 B = t.TypeVar("B", bound=exp.Binary)
 
 
+def _cast_div_operands_sql(self: Generator, expression: exp.Div) -> str:
+    l, r = expression.left, expression.right
+
+    if not l.is_type(*exp.DataType.FLOAT_TYPES) and not r.is_type(*exp.DataType.FLOAT_TYPES):
+        l.replace(exp.cast(l.copy(), to=exp.DataType.Type.DOUBLE))
```
GeorgeSittas, line 24: Should this happen conditionally, only when writing towards dialects that support float div semantics?

sqlglot/dialects/dialect.py
GeorgeSittas: yeah I think that's why we also have this segment right below it:

```python
        if not klass.STRICT_STRING_CONCAT and klass.DPIPE_IS_STRING_CONCAT:
            klass.parser_class.BITWISE[TokenType.DPIPE] = exp.SafeDPipe
```

sqlglot/dialects/dialect.py
GeorgeSittas: that's right there's some magic here which is implemented here:

```python
        # Pass required dialect properties to the tokenizer, parser and generator classes
        for subclass in (klass.tokenizer_class, klass.parser_class, klass.generator_class):
            for name, value in dialect_properties.items():
                if hasattr(subclass, name):
                    setattr(subclass, name, value)
```

GeorgeSittas: this basically allows you to sync generator/parser/tokenizer constants with those defined in the dialect class if the names match

sqlglot/optimizer/annotate_types.py
```diff
@@ -531,3 +533,18 @@ def _annotate_bracket(self, expression: exp.Bracket) -> exp.Bracket:
...
+        self._annotate_args(expression)
+
+        l, r = expression.left, expression.right
+        l_is_int = l.type and l.type.this in exp.DataType.INTEGER_TYPES
+        r_is_int = r.type and r.type.this in exp.DataType.INTEGER_TYPES
+
+        if l_is_int and r_is_int:
+            datatype = exp.DataType.Type.BIGINT
+        else:
+            datatype = exp.DataType.Type.DOUBLE
```
GeorgeSittas, line 547: i'm not sure if this is always correct - e.g. what if either operand is a `DECIMAL` with specific precision etc?

sqlglot/dialects/dialect.py
barakalon: So we add the TYPED_DIVISION attribute to the generator, but not the parser?
And then mutate the parser here in `dialect`, but conditionally change the generation logic in the _sql methods?

sqlglot/generator.py
barakalon: ah, you are correct. this should be CAST(l / r AS BIGINT)

sqlglot/optimizer/annotate_types.py
barakalon: I think the coercion rules for division are tricky.... hm

sqlglot/dialects/dialect.py
barakalon: Not sure I follow. But this _is_ being used conditionally

sqlglot/dialects/dialect.py
tobymao: yea, i think that's how we currently do things

sqlglot/dialects/dialect.py
GeorgeSittas: Ah we set it at the metaclass and I think when I made the comment I hadn't noticed yet.

So, coming back to (what I think was) Toby's comment, should we do this @ the method level and not at the metaclass, like:

```python
if self.TYPED_DIVISION and (above logic):
    l.replace(exp.cast(l.copy(), to=exp.DataType.Type.DOUBLE))
```

sqlglot/optimizer/annotate_types.py
barakalon: I'll switch this to _maybe_coerce, so at least it maintain existing behavior of Div

barakalon:
Super helpful @GeorgeSittas !

Alright, I think this is ready for another review.

tobymao (approved):

sqlglot/generator.py
```diff
@@ -2589,6 +2594,28 @@ def safedpipe_sql(self, expression: exp.SafeDPipe) -> str:
...
+                l.replace(exp.cast(l.copy(), to=exp.DataType.Type.DOUBLE))
+
+        return self.binary(expression, "/")
+
+    def typeddiv_sql(self, expression: exp.TypedDiv) -> str:
+        if self.TYPED_DIVISION:
+            return self.binary(expression, "/")
+
+        l, r = expression.left, expression.right
+        if l.is_type(*exp.DataType.INTEGER_TYPES) and r.is_type(*exp.DataType.INTEGER_TYPES):
```
GeorgeSittas, line 2612: Nit: maybe we could negate the condition here and just do a single `return self.binary(expression, "/")`?

GeorgeSittas (approved):
Nice! Left one last minor nit suggestion, but LGTM 🤞.

sqlglot/generator.py
barakalon: Yeah, I thought about that..
That's less code, but I think it's slightly harder to read.
Negates conditions are generally confusing. Plus, this keeps the branch symmetrical with the div_sql

GeorgeSittas: Sounds good, marking this as resolved then 👍

PR #2516, opened 2023-11-06
fix!: typed div and safe div semantics

barakalon:
Most dialects throw an error for `1 / 0`. Some return `null`.

This safely transpiles in one direction - from dialects that return null to those that raise an error.


sqlglot/expressions.py
```diff
@@ -6542,6 +6542,27 @@ def func(name: str, *args, dialect: DialectType = None, **kwargs) -> Func:
     return function
 
 
+def case(
```
barakalon, line 6545: Not actually used here - I added it when transpiling NULLIF to a CASE statement. But then I couldn't find any dialects that don't support NULLIf.

GeorgeSittas:
Dialects that raise on division by 0:
- Bigquery
- DataBricks
- [Oracle](https://stackoverflow.com/questions/5027808/how-to-avoid-divide-by-zero-error-in-an-sql-query) (?)
- Postgres
- Presto
- Redshift
- Snowflake
- [Teradata](https://stackoverflow.com/questions/39623177/teradata-returning-division-by-zero-error-how-to-adress)
- Trino
- T-SQL

Dialects that yield `NULL` on division by 0:
- ClickHouse
- DuckDB
- Hive
- MySQL
- Spark
- Spark2
- SQLite

sqlglot/dialects/spark2.py
```diff
@@ -101,6 +101,8 @@ def _unqualify_pivot_columns(expression: exp.Expression) -> exp.Expression:
 
 
 class Spark2(Hive):
+    SAFE_DIVISION = False
```
GeorgeSittas, line 104: Why is this `False`? I think it needs to be `False` only for databricks, see https://github.com/tobymao/sqlglot/pull/2516#issuecomment-1796481170

sqlglot/dialects/dialect.py
```diff
@@ -185,6 +182,10 @@ class Dialect(metaclass=_Dialect):
     # True means a / b is integer division if both a and b are integers.
     TYPED_DIVISION = False
 
+    # False means 1 / 0 throws an error.
+    # True means 1 / 0 returns null.
```
tobymao, line 186: what about where it returns nan?

tobymao: is there a difference between 1 / 0 and 1.0 / 0?

sqlglot/expressions.py
tobymao: it's fine

tobymao (approved):

sqlglot/dialects/dialect.py
barakalon: 1 / 0:

- ClickHouse: ?
- DuckDB: nan
- Hive: null
- MySQL: null
- Spark: ?
- Spark2: ?
- SQLite: null

@GeorgeSittas - can you fill in the blanks for clickhouse and spark?

GeorgeSittas: DuckDB, ClickHouse and Spark all return `NULL` for both `1 / 0` and `1.0 / 0`

barakalon: You sure about that for duckdb? I might be on an old version

barakalon: Ah, I'm calling the `.df()` method, that's why

barakalon: ok, looks like nothing return nan.

barakalon:
@GeorgeSittas I'll wait for your review before merging

GeorgeSittas:
Thanks - doing a pass at it now

sqlglot/optimizer/annotate_types.py
```diff
@@ -270,10 +271,10 @@ class TypeAnnotator(metaclass=_TypeAnnotator):
         exp.Max: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
         exp.Min: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
         exp.Null: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.NULL),
+        exp.Nullif: lambda self, e: self._annotate_by_args(e, "this"),
```
GeorgeSittas, line 274: This isn't correct for every dialect, e.g. postgres seems to have some coercion rules going on:

```
georgesittas=# select pg_typeof(nullif(1, 1.0));
 pg_typeof
-----------
 numeric
(1 row)

georgesittas=# select pg_typeof(nullif(1.0, 1));
 pg_typeof
-----------
 numeric
(1 row)

georgesittas=# select pg_typeof(nullif(1, 1));
 pg_typeof
-----------
 integer
(1 row)
```

GeorgeSittas: Perhaps it's safer to do a `_maybe_coerce` on the two args of `Nullif` or something

barakalon: Cool, seems like that might be what postgres is doing

GeorgeSittas (approved):
LGTM - just that one last comment re: the type inference of `exp.Nullif`.

PR #2522, opened 2023-11-07
Fix(mysql): don't duplicate CTEs in FULL OUTER JOIN transform

treysp:


GeorgeSittas (approved):

GeorgeSittas:
For some reason the semver message wasn't carried over to the merged commit. I think if the commit title is changed in GitHub then @ merge time it needs to be updated otherwise for some reason it uses the commit message (which doesn't contain semver info here).

Just pointing out for future merges.

PR #2524, opened 2023-11-07
Feat(optimizer): annotate type of ABS

GeorgeSittas:


tobymao:
what about ceiling and floor

GeorgeSittas:
> what about ceiling and floor

They're already annotated as `INT` in `TYPE_TO_EXPRESSIONS`

tobymao (approved):

PR #2529, opened 2023-11-08
Fix(snowflake): avoid advancing beyond array limit when parsing staged files

### LINKED ISSUE #2528: List index out of range for Snowflake SQL transpilation
Ran the issue using the latest `main` build `sqlglot-19.0.4.dev18`

Read: `snowflake`
Write `spark` 

When running the following query on Snowflake's publically available documentation, we get an `IndexError: list index out of range`. It's unclear which part of the query is unable to be tokenized. 

Code snippet 

https://docs.snowflake.com/en/sql-reference/sql/create-external-table

```
query_1666 = 'SELECT metadata$filename FROM @s1/;'

transpiled_sqls = transpile(query_1666, read='snowflake', write='spark', pretty=True, error_level=None)
```

georgesittas:
Fixes #2528

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -349,7 +349,7 @@ class Snowflake(Dialect):
             table: t.Optional[exp.Expression] = None
             if self._match_text_seq("@"):
                 table_name = "@"
-                while True:
+                while self._curr:
                     self._advance()
                     table_name += self._prev.text
                     if not self._match_set(self.STAGED_FILE_SINGLE_TOKENS, advance=False):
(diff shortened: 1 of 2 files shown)


georgesittas:
This doesn't fix the transpilation issue when targeting Spark because it doesn't support unquoted identifiers that contain `$`:

```
spark-sql (default)> select 1 as a$b;

[PARSE_SYNTAX_ERROR] Syntax error at or near '$'.(line 1, pos 13)

== SQL ==
select 1 as a$b
-------------^^^
```

This is a pain because we need to either change the tokenizer representation of `a$b` or add an expensive check for `Var` to see if there's `$` in its name in order to preserve that info for transpilation.

cc @tobymao

tobymao:
ok for now

PR #2534, opened 2023-11-09
Fix: transpile Snowflake structs correctly

### LINKED ISSUE #2533: Conversion from Snowflake Object to Spark StructType does not produce syntactically correct result
Ran the issue using main `sqlglot-19.0.4.dev18`

Read: `snowflake`
Write: `spark`

We are trying to convert Snowflake SQL code that creates an `OBJECT` using `{foo: bar}` syntax. In Spark, we would like to create the corresponding data type, a `StructType`. The transpiled Spark code is not syntactically correct and throws an error `AnalysisException: [INVALID_EXTRACT_BASE_FIELD_TYPE] Can't extract a value from "Manitoba". Need a complex type [STRUCT, ARRAY, MAP] but got "STRING".` 

This is presumably because the transpiled code is trying to make use the the `foo[bar]` method in SparkSQL to extract struct fields. However, this syntax doesn't work when _creating_ a struct. 

Code snippet: 

```
from pyspark.sql.session import SparkSession
from sqlglot import transpile

spark = SparkSession.builder.getOrCreate()
query = "SELECT { 'Manitoba': 'Winnipeg', 'foo': 'bar' } AS province_capital;"

transpiled_query = transpile(query, read='snowflake', write='spark', pretty=True, error_level=None)[0]

spark.sql(transpiled_query)
```

Produces the following string **(INCORRECT RESULT)**

```
SELECT
  STRUCT('Manitoba'['Winnipeg'], 'foo'['bar']) AS province_capital
```

Properly converted code in Spark SQL: 

```
SELECT STRUCT('Winnipeg' AS Manitoba, 'bar' AS foo) AS province_capital
```

Images with results 

Snowflake documentation for creating semi-structured data types using `{foo: bar}` syntax. 
https://docs.snowflake.com/en/sql-reference/data-types-semistructured

Spark SQL `struct` function 
https://spark.apache.org/docs/3.5.0/api/sql/#struct

georgesittas:
Fixes #2533

### DIFF
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -224,6 +224,19 @@ class Spark2(Hive):
         WRAP_DERIVED_VALUES = False
         CREATE_FUNCTION_RETURN_AS = False
 
+        def struct_sql(self, expression: exp.Struct) -> str:
+            args = []
+            for arg in expression.expressions:
+                if isinstance(arg, self.KEY_VALUE_DEFINITONS):
+                    if isinstance(arg, exp.Bracket):
+                        args.append(exp.alias_(arg.this, arg.expressions[0].name))
+                    else:
+                        args.append(exp.alias_(arg.expression, arg.this.name))
+                else:
+                    args.append(arg)
+
+            return self.func("STRUCT", *args)
+
         def temporary_storage_provider(self, expression: exp.Create) -> exp.Create:
             # spark2, spark, Databricks require a storage provider for temporary tables
             provider = exp.FileFormatProperty(this=exp.Literal.string("parquet"))
--- a/tests/dialects/test_duckdb.py
+++ b/tests/dialects/test_duckdb.py
@@ -356,14 +356,14 @@ class TestDuckDB(Validator):
             "STRUCT_PACK(x := 1, y := '2')",
             write={
                 "duckdb": "{'x': 1, 'y': '2'}",
-                "spark": "STRUCT(x = 1, y = '2')",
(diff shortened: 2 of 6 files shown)


tests/dialects/test_duckdb.py
```diff
@@ -356,14 +356,14 @@ def test_duckdb(self):
             "STRUCT_PACK(x := 1, y := '2')",
             write={
                 "duckdb": "{'x': 1, 'y': '2'}",
-                "spark": "STRUCT(x = 1, y = '2')",
+                "spark": "STRUCT(1 AS x, '2' AS y)",
```
georgesittas, line 359: Both this and the other test below were invalid, e.g. `STRUCT(x = 1, y = '2')` is incorrect Spark in this case:

```
spark-sql (default)> select STRUCT(x = 1, y = '2');
[UNRESOLVED_COLUMN.WITHOUT_SUGGESTION] A column or function parameter with name `x` cannot be resolved. ; line 1 pos 14;
'Project [unresolvedalias(struct(col1, ('x = 1), col2, ('y = 2)), None)]
+- OneRowRelation
```

sqlglot/dialects/spark2.py
```diff
@@ -224,6 +224,19 @@ class Generator(Hive.Generator):
         WRAP_DERIVED_VALUES = False
         CREATE_FUNCTION_RETURN_AS = False
 
+        def struct_sql(self, expression: exp.Struct) -> str:
```
georgesittas, line 227: added to spark2 because hive doesn't seem to support this struct syntax

tobymao (approved):

PR #2537, opened 2023-11-09
feat(optimizer): simplify DATE_ADD on literals

barakalon:
Another option is to canonicalize `DATE_ADD(x, 1, day)` to `x + INTERVAL 1 DAY`, but intervals vary quite a bit between dialects, and our support for them is pretty immature, so that feels risky.


sqlglot/optimizer/simplify.py
```diff
@@ -557,9 +562,9 @@ def _simplify_binary(expression, a, b):
     elif _is_date_literal(a) and isinstance(b, exp.Interval):
         a, b = extract_date(a), extract_interval(b)
         if a and b:
-            if isinstance(expression, exp.Add):
+            if isinstance(expression, (exp.Add, exp.DateAdd, exp.DatetimeAdd)):
```
tobymao, line 565: can you add a test? dateadd -> date, datetimeadd -> datetime ?

sqlglot/optimizer/simplify.py
```diff
@@ -507,6 +509,9 @@ def simplify_literals(expression, root=True):
                 return exp.Literal.number(value[1:])
             return exp.Literal.number(f"-{value}")
 
+    if isinstance(expression, DATE_OPS):
```
georgesittas, line 512: We can also get rid of `DATE_OPS` to avoid introducing more constants and instead do `type(expression) in INVERSE_DATE_OPS`

barakalon: Sure

Interesting test:

```python
class A:
    ...

class B:
    ...

class C:
    ...

class D:
    ...

SET = {A, B, C, D}
TYPES = (A, B, C, D)

b = B()

import timeit

setup = "from __main__ import SET, TYPES, b"
print(timeit.timeit("type(b) in TYPES", setup=setup, number=100000))
print(timeit.timeit("isinstance(b, TYPES)", setup=setup, number=100000))
```

Python 3.9:
```
type in   : 0.01205877
isinstance: 0.010168221999999998
```

Python 3.11:
```
type in   : 0.003846660954877734
isinstance: 0.004812692059203982
```

sqlglot/optimizer/simplify.py
barakalon: I _think_ I added the tests you want. Let me know if otherwise.

sqlglot/optimizer/simplify.py
georgesittas: Interesting indeed

PR #2543, opened 2023-11-09
feat: more robust boolean conversions for tsql

tobymao:


tobymao:
i didn't do this generically because presto has a boolean type and tsql doesn't

sqlglot/expressions.py
```diff
@@ -3689,6 +3689,7 @@ class Type(AutoName):
         Type.BIGINT,
         Type.INT128,
         Type.INT256,
+        Type.BIT,
```
barakalon, line 3692: Same as TINYINT?

sqlglot/generator.py
```diff
@@ -233,6 +233,9 @@ class Generator:
     # Whether or not nested CTEs (e.g. defined inside of subqueries) are allowed
     SUPPORTS_NESTED_CTES = True
 
+    # Whether or not conditions require booleans WHERE x = 0 vs WHERE x
+    ENSURE_BOOLS = False
```
barakalon, line 237: Is this also "0/1 can be used as true/false"?

sqlglot/optimizer/canonicalize.py
```diff
@@ -70,17 +70,23 @@ def remove_redundant_casts(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+# this is designed for presto, there is a similar transform for tsql
+# this is different in that it only operates on int types, this is because
+# presto has a boolean type whereas tsql doesn't (people use bits)
+# with y as (select true as x) select x = 0 -- illegal presto query
```
barakalon, line 76: select x = 0 from y

barakalon (approved):

sqlglot/optimizer/canonicalize.py
```diff
@@ -70,17 +70,23 @@ def remove_redundant_casts(expression: exp.Expression) -> exp.Expression:
...
         _replace_int_predicate(expression.right)
-
-    elif isinstance(expression, (exp.Where, exp.Having)) or (
+    elif isinstance(expression, exp.Not):
+        _replace_int_predicate(expression.this)
         # We can't replace num in CASE x WHEN num ..., because it's not the full predicate
-        isinstance(expression, exp.If)
-        and not (isinstance(expression.parent, exp.Case) and expression.parent.this)
+    elif isinstance(expression, exp.If) and not (
+        isinstance(expression.parent, exp.Case) and expression.parent.this
```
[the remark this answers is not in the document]
tobymao: it's not exactly the same, can the one in simplify be changed?

sqlglot/transforms.py
```diff
@@ -431,6 +431,32 @@ def move_ctes_to_top_level(expression: exp.Expression) -> exp.Expression:
...
+
+    for node, *_ in expression.walk():
+        if isinstance(node, exp.Connector):
+            _ensure_bool(node.left)
+            _ensure_bool(node.right)
+        elif isinstance(node, exp.Not):
+            _ensure_bool(node.this)
+        # We can't replace num in CASE x WHEN num ..., because it's not the full predicate
+        elif isinstance(node, exp.If) and not (
+            isinstance(node.parent, exp.Case) and node.parent.this
```
georgesittas, line 451: It's repeated here as well

sqlglot/optimizer/canonicalize.py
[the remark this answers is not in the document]
georgesittas: i'll take a look

georgesittas:
@tobymao should we factor out the logic that traverses conditions? It seems like there's a common pattern between `ensure_bool_predicates` and the `ensure_bools` in that they both traverse the same kind of subtrees. Perhaps we could have a template that does this which receives a callback to do the related transformations?

sqlglot/optimizer/canonicalize.py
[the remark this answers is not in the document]
georgesittas: ok i see what you mean, i guess let's leave the simplify condition as-is, but we can still dry this up so that it can be reused in the other section i commented in this pr

sqlglot/expressions.py
tobymao: no, tiny int is 0 255

sqlglot/generator.py
tobymao: sorry what does this mean?

sqlglot/transforms.py
tobymao: this is not the same as what's in simplify

georgesittas: it's the same as the ensure bool predicates one though right? https://github.com/tobymao/sqlglot/pull/2543#discussion_r1388446273

georgesittas (approved):
LGTM - should we implement `transforms.ensure_bools` using `replace_children` like in canonicalize.py?

tobymao:
not sure that's necessary

sqlglot/generator.py
barakalon: nvm

PR #2544, opened 2023-11-09
Fix!: disallow nested CTEs for Spark and Databricks

georgesittas:
Even though nested CTEs are allowed in Spark v3.4.0, they are not allowed in earlier minor versions such as v3.1.* and v3.2.* so we disallow them for the v3 version completely in order to ensure compatibility with these versions.


eakmanrq (approved):

tobymao (approved):
this is breaking

sqlglot/expressions.py
```diff
@@ -1009,7 +1009,6 @@ def selects(self) -> t.List[Expression]:
 
 class Cache(Expression):
     arg_types = {
-        "with": False,
```
georgesittas, line 1012: This was added a very long time ago and it doesn't seem to serve any purpose. I ran the following statement in most of the dialects we support and got back a syntax error:

```sql
WITH a AS (SELECT * FROM b) CACHE TABLE a
```

Spark does support using CTEs inside of a `Cache` expression, like so:

```sql
CACHE LAZY TABLE x AS WITH a AS (SELECT 1) SELECT a.* FROM a
```

but this produces a different AST, where the CTE is attached to the `Select` node and node the `Cache` one. 

The reason I needed to get rid of this arg was because it was making CTEs like the one in the above valid Spark statement bubble up to the `Cache` itself and so it would later be wrongly generated as:

```sql
WITH ... CACHE ...
```

PR #2545, opened 2023-11-09
Fix!(snowflake): don't quote DUAL table because it has special meaning

georgesittas:
Snowflake treats `DUAL` in a special way when it's used in a `FROM` clause, and quoting it breaks queries.

For example:

```
CREATE TABLE "DUAL" (a DOUBLE);
SELECT * FROM "DUAL";
```

Screenshot 2023-11-09 at 11 13 50 PM

```
SELECT * FROM DUAL;
```

Screenshot 2023-11-09 at 11 14 22 PM

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -248,7 +248,7 @@ class Snowflake(Dialect):
         ):
             return t.cast(E, expression)
 
-        return Dialect.quote_identifier(expression, identify=identify)
+        return super().quote_identifier(expression, identify=identify)
 
     class Parser(parser.Parser):
         IDENTIFY_PIVOT_STRINGS = True
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -241,7 +241,7 @@ class Snowflake(Dialect):
     def quote_identifier(cls, expression: E, identify: bool = True) -> E:
         if isinstance(expression, exp.Identifier) and not (
             # This disables quoting DUAL in SELECT ... FROM DUAL, because Snowflake treats an
-            # unquoted DUAL keyword in a special way and does not map it to an existing table
+            # unquoted DUAL keyword in a special way and does not map it to a user-defined table
             expression.name.lower() == "dual"
             and isinstance(expression.parent, exp.Table)
         ):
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -236,6 +237,19 @@ class Snowflake(Dialect):
...
+        # This disables quoting DUAL in SELECT ... FROM DUAL, because Snowflake treats an
+        # unquoted DUAL keyword in a special way and does not map it to a user-defined table
+        if (
+            isinstance(expression, exp.Identifier)
+            and isinstance(expression.parent, exp.Table)
+            and expression.name.lower() == "dual"
+        ):
+            return t.cast(E, expression)
+
+        return Dialect.quote_identifier(expression, identify=identify)
```
georgesittas, line 251: This is problematic because it's changing the receiver class and hence we don't get the `quote_identifier` behavior for Snowflake.. Trying to patch it now.

sqlglot/dialects/snowflake.py
```diff
@@ -236,6 +237,19 @@ class Snowflake(Dialect):
...
     }
 
+    @classmethod
+    def quote_identifier(cls, expression: E, identify: bool = True) -> E:
+        # This disables quoting DUAL in SELECT ... FROM DUAL, because Snowflake treats an
+        # unquoted DUAL keyword in a special way and does not map it to a user-defined table
+        if (
+            isinstance(expression, exp.Identifier)
+            and isinstance(expression.parent, exp.Table)
+            and expression.name.lower() == "dual"
```
tobymao, line 247: what if it's quoted

georgesittas: It will remain quoted

tobymao: why do you call lower though

georgesittas: because you can have both upper and lower case versions of the unquoted version and so you gotta normalize to compare

georgesittas: if you match anything which is `dual` when lowercased then you don't wanna touch it, if it's quoted then fine but if not you can't proceed calling `Dialect.quote_identifier` on it because you risk messing up the query

PR #2549, opened 2023-11-10
Fix!(tsql): ensure derived table projections are named / have aliases

georgesittas:
This aims to fix a bug where T-SQL projections are left unnamed for derived tables, which MSSQL doesn't like:

```
1> with t as (select 1) select * from t
2> go
...
No column name was specified for column 1 of 't'.
1> with t as (select 1 as c) select * from t
2> go
c
-----------
          1

(1 rows affected)
1> select * from (select 1) subq
2> go
...
No column name was specified for column 1 of 'subq'.
1> select * from (select 1 as c) subq
2> go
c
-----------
          1

(1 rows affected)
```

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -8,7 +8,7 @@ from sqlglot._typing import E
 from sqlglot.dialects.dialect import Dialect, DialectType
 from sqlglot.errors import OptimizeError
 from sqlglot.helper import seq_get
-from sqlglot.optimizer.scope import Scope, traverse_scope, walk_in_scope
+from sqlglot.optimizer.scope import Scope, build_scope, traverse_scope, walk_in_scope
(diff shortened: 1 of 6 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -450,10 +450,16 @@ def _add_replace_columns(
...
-def _qualify_outputs(scope: Scope) -> None:
+def qualify_outputs(scope_or_expression: Scope | exp.Expression) -> None:
     """Ensure all output columns are aliased"""
-    new_selections = []
+    if isinstance(scope_or_expression, exp.Expression):
+        scope = build_scope(scope_or_expression)
+        if not isinstance(scope, Scope):
+            return
+    else:
+        scope = scope_or_expression
```
georgesittas, line 460: Figured I'd make this accessible as a public helper of the `qualify_columns` module because it might be helpful elsewhere as well.

PR #2551, opened 2023-11-10
Fix(tsql): preserve column projection quotes for newly added Alias nodes

georgesittas:


sqlglot/dialects/tsql.py
```diff
@@ -213,7 +213,25 @@ def qualify_derived_table_outputs(expression: exp.Expression) -> exp.Expression:
...
+            if isinstance(c, exp.Column) and not c.alias
+        )
+
+        qualify_outputs(subqueryable)
+
+        # Preserve the quoting information of columns for newly added Alias nodes
+        subqueryable_selects = subqueryable.selects
+        for select_index in unaliased_column_projection_indexes:
+            select = subqueryable_selects[select_index]
+            if isinstance(select, exp.Alias):
```
tobymao, line 232: shouldn't this always be an alias because we qualify outputs?

georgesittas: yes it's a sanity check, I can probably get rid of it

sqlglot/dialects/tsql.py
```diff
@@ -213,7 +213,25 @@ def qualify_derived_table_outputs(expression: exp.Expression) -> exp.Expression:
...
+                column_identifier = select.this.this
+                select.args["alias"].set("quoted", column_identifier.args.get("quoted"))
```
tobymao, line 234: maybe let's check that this is an idintifier and called identifier.quoted

sqlglot/dialects/tsql.py
georgesittas: removed

sqlglot/dialects/tsql.py
georgesittas: added

sqlglot/dialects/tsql.py
```diff
@@ -213,7 +213,25 @@ def qualify_derived_table_outputs(expression: exp.Expression) -> exp.Expression:
...
+            if isinstance(c, exp.Column) and not c.alias
+        )
+
+        qualify_outputs(subqueryable)
+
+        # Preserve the quoting information of columns for newly added Alias nodes
+        subqueryable_selects = subqueryable.selects
+        for select_index in unaliased_column_indexes:
+            alias = subqueryable_selects[select_index]
+            column_identifier = alias.this.this
```
tobymao, line 232: this should be alias.this, and check that alias.this is a column

georgesittas: yea makes sense, changed

georgesittas: > and check that alias.this is a column

I assume you meant "is an identifier" instead?

PR #2554, opened 2023-11-11
Feat!: transpile NULLS FIRST/LAST to dialects that dont support it

georgesittas:
Reference: https://dba-presents.com/databases/sql-server/36-order-by-and-nulls-last-in-sql-server


tobymao (approved):

tests/dialects/test_dialect.py
```diff
@@ -1966,3 +1959,50 @@ def test_nested_ctes(self):
                 "tsql": "WITH t1(x) AS (SELECT 1), t2(y) AS (SELECT 2) SELECT * FROM (SELECT y AS y FROM t2) AS subq",
             },
         )
+
+    def test_unsupported_null_ordering(self):
```
georgesittas, line 1963: FYI I also sanity-checked this by running all queries under this test in the corresponding engines to ensure the results are the same, using data defined in a CTE:

```sql
WITH t(c) AS (SELECT 1 UNION ALL SELECT 2 UNION ALL SELECT NULL UNION ALL SELECT 0) SELECT * FROM t ORDER BY c;
```

PR #2560, opened 2023-11-13
Fix(snowflake): only generate TRY_CAST if cast value is of text type

georgesittas:
From Snowflake's [docs](https://docs.snowflake.com/en/sql-reference/functions/try_cast):

image

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -562,6 +562,14 @@ class Snowflake(Dialect):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        def trycast_sql(self, expression: exp.TryCast) -> str:
+            value = expression.this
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -562,6 +562,14 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        def trycast_sql(self, expression: exp.TryCast) -> str:
+            value = expression.this
+            if value.is_string or value.is_type(*exp.DataType.TEXT_TYPES):
```
tobymao, line 567: what happens if the user does

parse_one("""
select TRY_CAST(x AS int)
""", read="snowflake")

x is a string, but the type isn't known to sqlglot, it should maintain a try_cast shouldn't it

georgesittas: that's a good point thanks, will fix

tobymao (approved):

PR #2562, opened 2023-11-14
Feat!(snowflake): deactivate collation in REGEXP_LIKE

georgesittas:
From Snowflake's docs: "Arguments with collation specifications are currently not supported."

For example, both of the following queries fail in Snowflake:

```
-- Function REGEXP_LIKE does not support collation.
SELECT REGEXP_LIKE(COLLATE('foo', 'en-ci'), 'pattern');
SELECT REGEXP_LIKE('foo', COLLATE('pattern', 'en-ci'));
```

This PR ensures that there's no collation in either argument to avoid such errors.

References:
- https://docs.snowflake.com/en/sql-reference/functions/regexp_like#collation-details
- https://docs.snowflake.com/en/sql-reference/collation#specification-examples

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -167,14 +167,21 @@ def _datatype_sql(self: Snowflake.Generator, expression: exp.DataType) -> str:
     return self.datatype_sql(expression)
 
 
-def _regexpilike_sql(self: Snowflake.Generator, expression: exp.RegexpILike) -> str:
-    flag = expression.text("flag")
+def _regexplike_sql(self: Snowflake.Generator, expression: exp.RegexpILike | exp.RegexpLike) -> str:
+    flag = expression.args.get("flag")
 
-    if "i" not in flag:
-        flag += "i"
+    if isinstance(expression, exp.RegexpILike):
+        case_insensitive_flag = exp.Literal.string("i")
+        flag = self.func("CONCAT", case_insensitive_flag, flag) if flag else case_insensitive_flag
 
+    # Arguments with collation specifications are currently not supported, so we
+    # use an empty string as collation, which indicates to use no collation. See:
+    # https://docs.snowflake.com/en/sql-reference/functions/regexp_like
     return self.func(
-        "REGEXP_LIKE", expression.this, expression.expression, exp.Literal.string(flag)
+        "REGEXP_LIKE",
+        self.func("COLLATE", expression.this, "''"),
+        self.func("COLLATE", expression.expression, "''"),
+        flag,
     )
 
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -167,14 +167,21 @@ def _datatype_sql(self: Snowflake.Generator, expression: exp.DataType) -> str:
     return self.datatype_sql(expression)
 
 
-def _regexpilike_sql(self: Snowflake.Generator, expression: exp.RegexpILike) -> str:
-    flag = expression.text("flag")
```
georgesittas, line 171: We can't use `text("flag")` anymore because `REGEXP_LIKE` in Snowflake expects a third argument (`parameters`) which doesn't have to be a string literal and so this call would return an empty string if, say, we got a column instead.

sqlglot/dialects/snowflake.py
```diff
@@ -167,14 +167,21 @@ def _datatype_sql(self: Snowflake.Generator, expression: exp.DataType) -> str:
...
+def _regexplike_sql(self: Snowflake.Generator, expression: exp.RegexpILike | exp.RegexpLike) -> str:
+    flag = expression.args.get("flag")
 
-    if "i" not in flag:
-        flag += "i"
+    if isinstance(expression, exp.RegexpILike):
+        case_insensitive_flag = exp.Literal.string("i")
+        flag = self.func("CONCAT", case_insensitive_flag, flag) if flag else case_insensitive_flag
```
georgesittas, line 175: I didn't add any tests for the `CONCAT` code path because I don't know of a dialect that actually supports `flags` as a third argument in `RegexpILike`. I can probably just add one anyway for the SQLGlot dialect if necessary.

sqlglot/dialects/snowflake.py
```diff
@@ -167,14 +167,21 @@ def _datatype_sql(self: Snowflake.Generator, expression: exp.DataType) -> str:
...
 
+    # Arguments with collation specifications are currently not supported, so we
+    # use an empty string as collation, which indicates to use no collation. See:
+    # https://docs.snowflake.com/en/sql-reference/functions/regexp_like
     return self.func(
-        "REGEXP_LIKE", expression.this, expression.expression, exp.Literal.string(flag)
+        "REGEXP_LIKE",
+        self.func("COLLATE", expression.this, "''"),
```
tobymao, line 182: why are you adding collate here?

tobymao: shouldn't you just remove collate and warn if you see it?

georgesittas: Discussed in Slack - both having explicit collations here and @ DDL time are user errors and we don't need to do something about them.

PR #2564, opened 2023-11-14
Feat(executor): add support for array_unique_agg

wezham:
https://docs.snowflake.com/en/sql-reference/functions/array_unique_agg 

> The function ignores NULL values in column. If column contains only NULL values or the table containing column is empty, the function returns an empty ARRAY.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -258,6 +258,7 @@ class Snowflake(Dialect):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "ARRAYAGG": exp.ArrayAgg.from_arg_list,
+            "ARRAY_UNIQUE_AGG": exp.ArrayUniqueAgg.from_arg_list,
             "ARRAY_CONSTRUCT": exp.Array.from_arg_list,
             "ARRAY_GENERATE_RANGE": lambda args: exp.GenerateSeries(
                 # ARRAY_GENERATE_RANGE has an exlusive end; we normalize it to be inclusive
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -258,7 +258,6 @@ class Snowflake(Dialect):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "ARRAYAGG": exp.ArrayAgg.from_arg_list,
-            "ARRAY_UNIQUE_AGG": exp.ArrayUniqueAgg.from_arg_list,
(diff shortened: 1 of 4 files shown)


tests/test_executor.py
```diff
@@ -768,6 +768,31 @@ def test_group_by(self):
                 self.assertEqual(result.columns, columns)
                 self.assertEqual(result.rows, expected)
 
+    def test_array_unique_agg(self):
```
wezham, line 771: I didn't really like the idea of writing a test for one function, although it didn't seem like it should be grouped with the other functions as the test data didn't seem appropriate for unique testing

sqlglot/dialects/snowflake.py
```diff
@@ -258,6 +258,7 @@ class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "ARRAYAGG": exp.ArrayAgg.from_arg_list,
+            "ARRAY_UNIQUE_AGG": exp.ArrayUniqueAgg.from_arg_list,
```
georgesittas, line 261: This is not necessary, I believe `ARRAY_UNIQUE_AGG` will be parsed as `ArrayUniqueAgg`

tests/test_executor.py
georgesittas: What about adding it to `test_group_by`?

georgesittas (approved):

sqlglot/dialects/snowflake.py
georgesittas: Let's add an identity test for Snowflake, i.e. `ARRAY_UNIQUE_AGG(x)` should map to itself

georgesittas, line 261:
```suggestion
```

georgesittas:
Thanks for the PR @wezham! I'll apply the last changes and take it to the finish line.

georgesittas:
https://github.com/tobymao/sqlglot/commit/cf5beb8f14df742c672efa27ca1a79e4897d9bbd

georgesittas:
@wezham check out https://github.com/tobymao/sqlglot/pull/2566 as well, Toby pointed out that there's already a `SetAgg` expression that represents the `ARRAY_UNIQUE_AGG` concept so I went ahead and got rid of the expression class you added to use that and enable transpiling from and to Snowflake.

sqlglot/dialects/snowflake.py
wezham: I was a little unclear why this mapping exists in the first place. When do you add methods to this mapping?

PR #2566, opened 2023-11-14
Refactor!: get rid of SetAgg to use ArrayUniqueAgg for consistency

georgesittas:


sqlglot/dialects/snowflake.py
```diff
@@ -519,6 +520,7 @@ class Generator(generator.Generator):
                     transforms.eliminate_semi_and_anti_joins,
                 ]
             ),
+            exp.ArrayUniqueAgg: rename_func("ARRAY_UNIQUE_AGG"),
```
tobymao: why does this need to be renamed?

georgesittas: leftover from the previous commit

wezham:
Thanks for following up on this. I think I still need to familiarise myself with the code base a bit more to understand these changes

PR #2568, opened 2023-11-15
Fix(snowflake): Snowflake only supports literals in LIMIT

treysp:


tobymao:
add a test, look for other tests of this type are sefined

PR #2569, opened 2023-11-15
Feat(executor): add support for null replacement value in ARRAY_JOIN

georgesittas:
cc @wezham

### DIFF
--- a/sqlglot/executor/env.py
+++ b/sqlglot/executor/env.py
@@ -6,7 +6,7 @@ from functools import wraps
 
 from sqlglot import exp
 from sqlglot.generator import Generator
-from sqlglot.helper import PYTHON_VERSION, filter_none
+from sqlglot.helper import PYTHON_VERSION
 
 
 class reverse_key:
@@ -23,7 +23,7 @@ class reverse_key:
 def filter_nulls(func, empty_null=True):
     @wraps(func)
     def _func(values):
-        filtered = tuple(filter_none(values))
+        filtered = tuple(v for v in values if v is not None)
         if not filtered and empty_null:
             return None
         return func(filtered)
@@ -141,7 +141,7 @@ def interval(this, unit):
 
 @null_if_any("this", "expression")
 def arrayjoin(this, expression, null=None):
-    return expression.join(filter_none((x if x is not None else null for x in this)))
+    return expression.join(x for x in (x if x is not None else null for x in this) if x is not None)
 
 
 ENV = {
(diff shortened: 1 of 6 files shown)


sqlglot/executor/env.py
```diff
@@ -139,6 +139,11 @@ def interval(this, unit):
     return datetime.timedelta(**{unit: float(this)})
 
 
+@null_if_any("this", "expression")
+def arrayjoin(this, expression, null=None):
+    return expression.join(filter(None, (x or null for x in this)))
```
tobymao, line 144: you need to check x is none because x could be falsey

georgesittas: ha good one yeah I missed it, thanks!

georgesittas: Addressed + added `filter_none` helper to get rid of `None` values in iterables. Cleaned up several sections.

sqlglot/helper.py
```diff
@@ -466,6 +466,11 @@ def merge_ranges(ranges: t.List[t.Tuple[A, A]]) -> t.List[t.Tuple[A, A]]:
     return merged
 
 
+def filter_none(iterable: t.Iterable) -> t.Iterator:
```
tobymao, line 469: does this hurt performance?

georgesittas: I can check, should I just run bench.py w/ before & after?

georgesittas: I'd expect it to be a minimal hit (if any) due to the function call overhead but we can test it.

tobymao: function call overhead is a good amount

tobymao: i'm a bit wary of this helper now and think we may not need it

georgesittas: FWIW, ran the benchmark with Python 3.10.13 and got these (main vs my branch, respectively):

```
➜  benchmarks git:(main) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01261 (1.0) | 0.00986 (0.782) |
|           short |   0.00067 (1.0) | 0.00056 (0.841) |
|            long |   0.01010 (1.0) | 0.00721 (0.713) |
|           crazy |   0.03203 (1.0) | 0.02216 (0.691) |
➜  benchmarks git:(jo/array_join_improvement) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01225 (1.0) | 0.00792 (0.646) |
|           short |   0.00066 (1.0) | 0.00056 (0.846) |
|            long |   0.01212 (1.0) | 0.00729 (0.601) |
|           crazy |   0.03215 (1.0) | 0.02237 (0.695) |
```

The test suite run times (`make style && make test`):

main: Ran 991 tests in 97.591s
my branch: Ran 991 tests in 98.233s

barakalon:
```python
from timeit import timeit

data = [1, 2, 3, None]
data_big = data * 100

def filter_none(iterable):
    return (x for x in iterable if x is not None)

setup = "from __main__ import data, data_big, filter_none"

print("comprehension:", timeit("[a for a in data if a is not None]", setup=setup))
print("function:", timeit("[a for a in filter_none(data)]", setup=setup))
print("comprehension (big):", timeit("[a for a in data_big if a is not None]", setup=setup))
print("function (big):", timeit("[a for a in filter_none(data_big)]", setup=setup))
```

Output:
```
comprehension: 0.2941456299740821
function: 0.613222278887406
comprehension (big): 8.075551941990852
function (big): 16.2689924351871
```

Python 3.11.5

georgesittas: Thanks @barakalon! I'll go ahead and revert this, then.

PR #2570, opened 2023-11-17
feat(sql): propagate constants to projections

barakalon:
Tableau sends our application queries like this:

```sql
SELECT
  SUM(a) AS a,
  MIN(b) AS b
FROM x
WHERE b = '123'
```

This is a reasonable query for Tableau to generate, but these are problematic for our application, which has a complex query rewriter.

This rule rewrites this as:
```sql
SELECT
  SUM(a) AS a,
  MIN('123') AS b
FROM x
WHERE b = '123'
```

Which is slightly simpler.

I chose to add this to constant propagation, since its kinda similar. Happy to move it somewhere else.

@georgesittas I think this simplifies the constant propagation code a little bit. Let me know what you think.

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -5,7 +5,6 @@ import typing as t
 from collections import deque
 from decimal import Decimal
 
-import sqlglot
 from sqlglot import exp
 from sqlglot.helper import first, is_iterable, merge_ranges, while_changing
 from sqlglot.optimizer.scope import find_all_in_scope, walk_in_scope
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -374,47 +373,88 @@ def absorb_and_eliminate(expression, root=True):
...
 
     SELECT * FROM t WHERE a = b AND b = 5 becomes
     SELECT * FROM t WHERE a = 5 AND b = 5
 
     Reference: https://www.sqlite.org/optoverview.html
+
+    This also propagates constants into projections:
+
+    SELECT a FROM x WHERE a = 1 becomes
+    SELECT 1 FROM x WHERE a = 1
```
georgesittas, line 431: An example where this transformation doesn't fully preserve the semantics:

```
mysql> create temporary table foo with t(c) as (select '1') select c from t where c = 1;
Query OK, 1 row affected (0.01 sec)
Records: 1  Duplicates: 0  Warnings: 0

mysql> describe foo;
+-------+------------+------+-----+---------+-------+
| Field | Type       | Null | Key | Default | Extra |
+-------+------------+------+-----+---------+-------+
| c     | varchar(1) | NO   |     |         | NULL  |
+-------+------------+------+-----+---------+-------+
1 row in set (0.00 sec)

mysql> select * from foo;
+---+
| c |
+---+
| 1 |
+---+
1 row in set (0.01 sec)
```

And if you transformed it..

```
mysql> create temporary table bar with t(c) as (select '1') select 1 from t where c = 1;
Query OK, 1 row affected (0.00 sec)
Records: 1  Duplicates: 0  Warnings: 0

mysql> select * from bar;
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.01 sec)

mysql> describe bar;
+-------+------+------+-----+---------+-------+
| Field | Type | Null | Key | Default | Extra |
+-------+------+------+-----+---------+-------+
| 1     | int  | NO   |     | 0       | NULL  |
+-------+------+------+-----+---------+-------+
1 row in set (0.00 sec)
```

georgesittas: Not sure if this is important for your use-case (or in general lol) though, @barakalon

sqlglot/optimizer/simplify.py
```diff
@@ -374,47 +373,88 @@ def absorb_and_eliminate(expression, root=True):
...
     """
+    if isinstance(expression, exp.Select):
+        where = expression.args.get("where")
+        if not where:
+            return expression
 
-    if (
-        isinstance(expression, exp.And)
-        and (root or not expression.same_parent)
-        and sqlglot.optimizer.normalize.normalized(expression, dnf=True)
```
georgesittas, line 390: Why did we remove the DNF check for `expression`?

barakalon: I don't think you actually wanted DNF? Shouldn't it be CNF?
Here, we check the top level ANDs for equalities and don't traverse into any ORs when replacing. I think its effectively the same thing.

sqlglot/optimizer/simplify.py
```diff
@@ -374,47 +373,88 @@ def absorb_and_eliminate(expression, root=True):
...
-    ):
-        constant_mapping = {}
-        for expr, *_ in walk_in_scope(expression, prune=lambda node, *_: isinstance(node, exp.If)):
```
georgesittas, line 393: Curious why the `If` instance check was dropped here. Is this a side effect of how the constant mapping is constructed now?

sqlglot/optimizer/simplify.py
barakalon: Ah, good find!

Two issues:
1. The alias. Early optimizations should fix this, since they'll add the alias. But we can check for that here too and ensure its the same.
2. The type. Oof, this one is tricky. Perhaps we move this to canonicalize and add a CAST? And perhaps we don't replace if there's any type mismatch, too avoid coercion issues?

sqlglot/optimizer/simplify.py
georgesittas: > I don't think you actually wanted DNF? Shouldn't it be CNF?

I think DNF was right here - you want to aggregate all conjunction-only terms and run the constant propagation rule inside them. So for example if you have `(A Λ B Λ C) V (D Λ E Λ F) V ...`, which is in DNF, you can propagate constants in each of the sub-terms.

sqlglot/optimizer/simplify.py
barakalon: Right, we only check top level equality operators. We don't recurse.

sqlglot/optimizer/simplify.py
georgesittas: >  Here, we check the top level ANDs for equalities and don't traverse into any ORs when replacing. I think its effectively the same thing.

I see. I'd need to think about this a bit more to understand the implications, e.g. in inputs where the con/disjunctions are not explicitly grouped (i.e. A AND B OR C, stuff like that). At the time I think the DNF check felt more safe, which is why I did it that way.

barakalon: Ah, that's a good point. I forget the AST doesn't actually have correct precedence

sqlglot/optimizer/simplify.py
georgesittas: 1. I guess it's a tradeoff.. Checking about aliases here means we don't tight couply with earlier rules in the optimizer pipeline, but introduces complexity. (OTOH I don't remember if we can currently safely run this rule without earlier steps, just thinking out loud)
2. Yeah this is definitely an _oof_ moment. Checking for a type mismatch probably makes more sense if we want to avoid this edge case. Not sure I understood the `CAST` suggestion.

sqlglot/optimizer/simplify.py
barakalon: Er, or does it? I'll have to think through this...

sqlglot/optimizer/simplify.py
barakalon: 1. I think simplify is one of the few rules that can safely be run without prior steps.
2. I mean CAST to ensure the exact same type. E.G. `varchar(1)` vs `varchar(2)`

sqlglot/optimizer/simplify.py
georgesittas: The AST _does_ encode the correct precedence in general, but different dialects may prioritize the parsing of `AND` and `OR` differently, i.e. assign different precedence levels to them when not grouped (though I don't have an example for this right now, but can check)

sqlglot/optimizer/simplify.py
georgesittas: Ah yeah, makes sense.

barakalon:
Given all the nuances here and the questionable generic value of this optimization, I’m tempted to close this and just give it a go in my application first, where the value is higher and risk is lower

PR #2578, opened 2023-11-19
Fix!(optimizer): annotate text + numeric using the numeric type

georgesittas:
BigQuery, ClickHouse, DuckDB, Trino, Presto: int + str fails for both 1 + ‘1’ and 1 + ‘a’
Postgres, Redshift, Snowflake, T-SQL: int + str fails only for 1 + ‘a’, yields 2 for 1 + ‘1’
Hive (, Spark2?): 1 + ‘1’ yields 2, 1 + ‘a’ yields null
Spark (, Databricks?): 1 + '1' yields 2.0, 1 + 'a' yields null
MySQL, SQLite: 1 + ‘1’ yields 2, 1 + ‘a’ yields 1

Note that Spark (at least ≥v3.4.0, which I'm testing with) infers the type of `1 + '1'` to be `double` (see below). I haven't taken this into account in this PR since it seems to be the exception to the rule, but happy to discuss more about it.

```python
>>> spark.sql("select 1 + '1'")
DataFrame[(1 + 1): double]
```

### DIFF
--- a/tests/fixtures/optimizer/canonicalize.sql
+++ b/tests/fixtures/optimizer/canonicalize.sql
@@ -10,6 +10,9 @@ SELECT CAST(1 AS VARCHAR) AS "a" FROM "w" AS "w";
 SELECT CAST(1 + 3.2 AS DOUBLE) AS a FROM w AS w;
 SELECT 1 + 3.2 AS "a" FROM "w" AS "w";
 
+SELECT '1' + 1 AS "col";
+SELECT '1' + 1 AS "col";
+
 SELECT CAST('2022-01-01' AS DATE) + INTERVAL '1' day;
 SELECT CAST('2022-01-01' AS DATE) + INTERVAL '1' day AS "_col_0";
 
--- a/tests/fixtures/optimizer/canonicalize.sql
+++ b/tests/fixtures/optimizer/canonicalize.sql
@@ -13,6 +13,9 @@ SELECT 1 + 3.2 AS "a" FROM "w" AS "w";
 SELECT '1' + 1 AS "col";
 SELECT '1' + 1 AS "col";
 
+SELECT '1' + '1' AS "col";
+SELECT CONCAT('1', '1') AS "col";
+
 SELECT CAST('2022-01-01' AS DATE) + INTERVAL '1' day;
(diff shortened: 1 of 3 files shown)


tests/fixtures/optimizer/canonicalize.sql
```diff
@@ -10,6 +10,9 @@ SELECT CAST(1 AS VARCHAR) AS "a" FROM "w" AS "w";
 SELECT CAST(1 + 3.2 AS DOUBLE) AS a FROM w AS w;
 SELECT 1 + 3.2 AS "a" FROM "w" AS "w";
 
+SELECT '1' + 1 AS "col";
+SELECT '1' + 1 AS "col";
```
georgesittas, line 14: This was previously transformed into `SELECT CONCAT('1', 1) AS "col"` which doesn't match the semantics implemented by the aforementioned dialects.

barakalon (approved):
I think the optimize still converts int + str to CONCAT? Perhaps we should make that optimization only happen on str + str. We have code in our application - which assumes mysql semantics - to work around this.

tests/fixtures/optimizer/canonicalize.sql
tobymao: does '1' + '1' still equal concat?

georgesittas: yes

georgesittas:
> I think the optimize still converts int + str to CONCAT? Perhaps we should make that optimization only happen on str + str. We have code in our application - which assumes mysql semantics - to work around this.

Don't think so, observe:

```python
>>> optimize("SELECT 1 + '1'").sql()
'SELECT 1 + \'1\' AS "_col_0"'
```

> We have code in our application - which assumes mysql semantics - to work around this.

Does this mean this PR will enable you to get rid of that logic? I think it's good to go from me but let me know if there are outstanding items to chat about.

georgesittas:
So basically the optimizer (canonicalize.py) converts an `Add` to `Concat` if the type of the _former_ is a string. Which is why this PR solves that at the type inference level instead of the canonicalization rule.

tobymao (approved):

Issue #1745, opened 2023-06-08
Type error when converting datediff from redshift to trino 


pangyifish:
```
sql = "select datediff(week,'2009-01-01','2009-12-31')"
converted_sql = sqlglot.transpile(sql, read="redshift", write="trino")[0]
print(converted_sql)
SELECT DATE_DIFF('week', '2009-01-01', '2009-12-31')
```

Trino error: `Unexpected parameters (varchar(4), varchar(10), varchar(10)) for function date_diff. Expected: date_diff(varchar(x), date, date), date_diff(varchar(x), timestamp(p), timestamp(p)), date_diff(varchar(x), timestamp(p) with time zone, timestamp(p) with time zone), date_diff(varchar(x), time(p), time(p)), date_diff(varchar(x), time(p) with time zone, time(p) with time zone)'
`

Changing the SQL to `SELECT DATE_DIFF('week', DATE'2009-01-01', DATE'2009-12-31')` works in Trino

https://trino.io/docs/current/functions/datetime.html

tobymao:
feel free to start making prs

pangyifish:
> feel free to start making prs

Sure, thanks!


+DATEADD function has the same issue

tobymao:
we’ll address both , use these at examples for future issues

tobymao:
also would be great if you joined or slack so we can help out with prs

tobikodata.com/slack

pangyifish:
> 

Sure, joined already!

PR #2583, opened 2023-11-21
Add failing test for dateadd on redshift

purcell:
On Redshift, DATEADD can be used to modify timestamps, yielding timestamps. However, when sqlglot normalises such cases, the timestamps are forcibly cast to dates, which is a very different result.

This PR includes failing test cases, but I'm not sure how best to construct a fix. Failure messages follow:

```
FAIL: test_redshift (tests.dialects.test_redshift.TestRedshift.test_redshift) [SELECT DATEADD('hour',0,CAST('2020-02-02 01:03:05.124' AS TIMESTAMP)) -> redshift]
  File "/Users/steve/Projects/External/sqlglot/tests/dialects/test_dialect.py", line 60, in validate_all
    self.assertEqual(
AssertionError: "SELE[17 chars]0, CAST(CAST('2020-02-02 01:03:05.124' AS TIMESTAMP) AS DATE))" != "SELE[17 chars]0, CAST('2020-02-02 01:03:05.124' AS TIMESTAMP))"
- SELECT DATEADD(hour, 0, CAST(CAST('2020-02-02 01:03:05.124' AS TIMESTAMP) AS DATE))
?                         -----                                            ---------
+ SELECT DATEADD(hour, 0, CAST('2020-02-02 01:03:05.124' AS TIMESTAMP))


FAIL: test_redshift (tests.dialects.test_redshift.TestRedshift.test_redshift) [SELECT DATEDIFF('second', '2020-02-02 01:03:05.124','2020-02-02 00:00:00.000') -> redshift]
  File "/Users/steve/Projects/External/sqlglot/tests/dialects/test_dialect.py", line 60, in validate_all
    self.assertEqual(
AssertionError: "SELE[49 chars]' AS DATE), CAST('2020-02-02 00:00:00.000' AS DATE))" != "SELE[49 chars]' AS TIMESTAMP), CAST('2020-02-02 00:00:00.000' AS TIMESTAMP))"
- SELECT DATEDIFF(second, CAST('2020-02-02 01:03:05.124' AS DATE), CAST('2020-02-02 00:00:00.000' AS DATE))
?                                                           ^ ^^                                     ^ ^^
+ SELECT DATEDIFF(second, CAST('2020-02-02 01:03:05.124' AS TIMESTAMP), CAST('2020-02-02 00:00:00.000' AS TIMESTAMP))
?                                                           ^^^^^^ ^^                                     ^^^^^^ ^^
```

### DIFF
--- a/tests/dialects/test_redshift.py
+++ b/tests/dialects/test_redshift.py
@@ -77,6 +77,18 @@ class TestRedshift(Validator):
             },
         )
         self.validate_all(
+            "SELECT DATEADD('hour',0,CAST('2020-02-02 01:03:05.124' AS TIMESTAMP))",
+            write={
+                "redshift": "SELECT DATEADD(hour, 0, CAST('2020-02-02 01:03:05.124' AS TIMESTAMP))",
+            },
+        )
+        self.validate_all(
+            "SELECT DATEDIFF('second', '2020-02-02 01:03:05.124','2020-02-02 00:00:00.000')",
+            write={
+                "redshift": "SELECT DATEDIFF(second, CAST('2020-02-02 01:03:05.124' AS TIMESTAMP), CAST('2020-02-02 00:00:00.000' AS TIMESTAMP))",
+            },
+        )
+        self.validate_all(
             "SELECT STRTOL('abc', 16)",
             read={
                 "trino": "SELECT FROM_BASE('abc', 16)",


purcell:
The same issue is present with DATEDIFF, in fact.

purcell:
I added a further test for datediff to illustrate.

georgesittas:
Hi @purcell, thanks for reporting this - I'll take a look shortly.

georgesittas:
It looks like we have two options here, with different tradeoffs:

- Leave `DATEADD` and `DATEDIFF` arguments as-is, thus fully preserving Redshift semantics but breaking transpilation towards dialects such as Presto which doesn't support executing these functions with e.g. `VARCHAR` values (see https://github.com/tobymao/sqlglot/issues/1745 and https://github.com/tobymao/sqlglot/commit/bb1f1a035 for more context).
- Add casts to `TIMESTAMP` instead of `DATE`. This should fix truncation issues when generating SQL back for Redshift where the arguments are `TIMESTAMP`s already. It should also not break the above transpilation towards Presto I assume, but need to double check this. It will, however, alter the semantics when the arguments are `TIME` or `TIMETZ`, because we will be casting them to `TIMESTAMP` instead.

Any thoughts / preferences here @tobymao?

purcell:
I'd probably avoid adding casts altogether in cases where functions can accept different types. But I don't have a good mental map of the scheme sqlglot uses.

purcell:
I don't think you can just cast to a different type instead:
```
DATEDIFF( datepart, {date|time|timetz|timestamp}, {date|time|timetz|timestamp} )
```

georgesittas:
> I'd probably avoid adding casts altogether in cases where functions can accept different types. But I don't have a good mental map of the scheme sqlglot uses.

That would be ideal, but like I said it basically comes down to deciding what tradeoffs need to be made. In this particular case, if we don't add casts then transpiling `DATEADD` and `DATEDIFF` into e.g. Presto will break when the arguments are strings.

georgesittas:
> I don't think you can just cast to a different type instead:
> 
> ```
> DATEDIFF( datepart, {date|time|timetz|timestamp}, {date|time|timetz|timestamp} )
> ```

I need to investigate this a bit more, but if `TIMESTAMP` can serve as a "superset" type in this context, perhaps that's "good enough".

In any case, I think I'll wait for Toby chime in on this one.

georgesittas:
Ok I think we've found an alternative option which takes care of the issues mentioned above. I'll look into implementing it soon.

purcell:
At least for `DATEADD`, you couldn't just cast to timestamp, because that would implicitly change the return type too: https://docs.aws.amazon.com/redshift/latest/dg/r_DATEADD_function.html

I suspect that for both functions one might hit all sorts of cases with implicit time/date/timezone conversions depending on the session, connection and DB, so that it would be very hard to reason about whether casting to a single type would be safe.

purcell:
Oh cool, just saw your comment. Will be interested to see what you come up with! :)

georgesittas:
@purcell closing this PR in favor of https://github.com/tobymao/sqlglot/pull/2588

purcell:
Nice, makes sense — that looks like a good approach!

georgesittas:
@purcell https://github.com/tobymao/sqlglot/commit/eb15a5fbcebf3d7920b74576a8343f1607b143b5 this should fix the Redshift issues.

purcell:
Nice! That was a big change. 💪 Thank you!

PR #2588, opened 2023-11-22
Fix!(redshift): leave DATEADD and DATEDIFF as-is when generating them

georgesittas:
The return type of these functions [depends](https://docs.aws.amazon.com/redshift/latest/dg/r_DATEADD_function.html) on their arguments, so it's unsafe to preemptively cast the arguments to DATE. This parses DATEADD into TsOrDsAdd and DATEDIFF into TsOrDsDiff so that we can leave them unaffected when generating Redshift and still allow them to be transpiled in other dialects.

- [x] Implement / improve SQL generation for `TsOrDsAdd` and `TsOrDsDiff` for at least the dialects that already supported `TsOrDsAdd`
- [x] Test queries against engines to ensure semantics are not broken


sqlglot/dialects/bigquery.py
```diff
@@ -521,7 +537,8 @@ class Generator(generator.Generator):
             exp.TimestampSub: date_add_interval_sql("TIMESTAMP", "SUB"),
             exp.TimeStrToTime: timestrtotime_sql,
             exp.Trim: lambda self, e: self.func(f"TRIM", e.this, e.expression),
-            exp.TsOrDsAdd: date_add_interval_sql("DATE", "ADD"),
```
tobymao, line 524: was this incorrect?

georgesittas: Yes, `date_add_interval_sql` doesn't add a cast and BQ doesn't like ts/ds in `DATE_ADD`

sqlglot/dialects/duckdb.py
```diff
@@ -322,6 +322,12 @@ class Generator(generator.Generator):
             exp.TimeToUnix: rename_func("EPOCH"),
             exp.TsOrDiToDi: lambda self, e: f"CAST(SUBSTR(REPLACE(CAST({self.sql(e, 'this')} AS TEXT), '-', ''), 1, 8) AS INT)",
             exp.TsOrDsAdd: _ts_or_ds_add_sql,
+            exp.TsOrDsDiff: lambda self, e: self.func(
+                "DATE_DIFF",
+                f"'{e.args.get('unit') or 'day'}'",
+                exp.cast(e.expression, "DATE"),
```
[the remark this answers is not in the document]
georgesittas: Ah I think I missed this one because I changed the BQ to use `TIMESTAMP` afterwards. Will probably change this as well, ty.

sqlglot/dialects/bigquery.py
georgesittas: So basically Hive `DATE_ADD` and `DATE_SUB` -> BigQuery would have this problem, as `TsOrDsAdd` is only used in `Hive` when parsing these functions.

sqlglot/dialects/presto.py
```diff
@@ -95,18 +95,21 @@ def _ts_or_ds_to_date_sql(self: Presto.Generator, expression: exp.TsOrDsToDate)
...
+    name: str,
+) -> t.Callable[[Presto.Generator, exp.TsOrDsAdd | exp.TsOrDsDiff], str]:
+    def _delta_sql(self: Presto.Generator, expression: exp.TsOrDsAdd | exp.TsOrDsDiff) -> str:
+        this = expression.this
+        expr = expression.expression
 
-    if not isinstance(this, exp.CurrentDate):
-        this = exp.cast(exp.cast(expression.this, "TIMESTAMP", copy=True), "DATE")
+        if not isinstance(this, exp.CurrentDate):
+            this = exp.cast(exp.cast(this, "TIMESTAMP", copy=True), "DATE")
```
tobymao, line 106: is copy still necessary?

sqlglot/dialects/bigquery.py
georgesittas: This is an example where the transpilation produced invalid BigQuery:

```python
>>> sqlglot.transpile("DATE_ADD('2020-01-01 01:01:01', 1)", "hive", "bigquery")
["DATE_ADD('2020-01-01 01:01:01', INTERVAL 1 DAY)"]
```

Hive:

Screenshot 2023-11-22 at 4 33 40 PM

BQ:

Screenshot 2023-11-22 at 4 33 57 PM

In this PR we get:

```python
>>> import sqlglot
>>> sqlglot.transpile("DATE_ADD('2020-01-01 01:01:01', 1)", "hive", "bigquery")
["DATE_ADD(CAST('2020-01-01 01:01:01' AS DATETIME), INTERVAL 1 DAY)"]
```

So this now runs in BQ, though it produces a `TIMESTAMP`:

Screenshot 2023-11-22 at 4 34 20 PM

sqlglot/generator.py
```diff
@@ -61,6 +61,9 @@ class Generator:
         exp.TsOrDsAdd: lambda self, e: self.func(
             "TS_OR_DS_ADD", e.this, e.expression, exp.Literal.string(e.text("unit"))
         ),
+        exp.TsOrDsDiff: lambda self, e: self.func(
+            "TS_OR_DS_DIFF", e.this, e.expression, exp.Literal.string(e.text("unit"))
```
tobymao, line 65: why does this and dsadd need an override?

georgesittas: We probably don't need them. The `TsOrDsAdd` one was added a long time ago (😄): https://github.com/tobymao/sqlglot/commit/c8063bfdb#diff-362498362ae539e216cd83ebfe510977107648742f1a1bee7b2db7b5b7d27974

georgesittas: PR no. 39

georgesittas: Got rid of 'em

tests/dialects/test_hive.py
```diff
@@ -328,7 +328,6 @@ def test_time(self):
                 "presto": "DATE_ADD('DAY', 1, CAST(CAST('2020-01-01' AS TIMESTAMP) AS DATE))",
                 "hive": "DATE_ADD('2020-01-01', 1)",
                 "spark": "DATE_ADD('2020-01-01', 1)",
-                "": "TS_OR_DS_ADD('2020-01-01', 1, 'DAY')",
```
tobymao, line 331: leave this test and see what it transpiles to?

tobymao (approved):

tests/dialects/test_hive.py
georgesittas: Sure, I think the only difference is that the `DAY` unit is not generated as a string but as a var.

sqlglot/dialects/bigquery.py
```diff
@@ -184,6 +185,16 @@ def _array_contains_sql(self: BigQuery.Generator, expression: exp.ArrayContains)
     )
 
 
+def _ts_or_ds_add_sql(self: BigQuery.Generator, expression: exp.TsOrDsAdd) -> str:
+    return date_add_interval_sql("DATE", "ADD")(self, ts_or_ds_add_cast(expression))
+
+
+def _ts_or_ds_diff_sql(self: BigQuery.Generator, expression: exp.TsOrDsDiff) -> str:
+    expression.this.replace(exp.cast(expression.this, "TIMESTAMP", copy=True))
+    expression.expression.replace(exp.cast(expression.this, "TIMESTAMP", copy=True))
```
georgesittas, line 194: The reasoning here is that `DATEDIFF` [returns](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_diff) a `BIGINT` in both BigQuery and Redshift, so promoting the args `TIMESTAMP` shouldn't really affect the result, unless the value to-be-cast is a _time_ string because then the cast will fail. But then again, we can't really handle all cases without annotating types.

sqlglot/dialects/postgres.py
```diff
@@ -60,16 +66,16 @@ def _date_diff_sql(self: Postgres.Generator, expression: exp.DateDiff) -> str:
...
+    start = f"CAST({self.sql(expression, 'expression')} AS TIMESTAMP)"
 
     if factor is not None:
         return f"CAST(EXTRACT(epoch FROM {end} - {start}){factor} AS BIGINT)"
 
     age = f"AGE({end}, {start})"
 
     if unit == "WEEK":
-        unit = f"EXTRACT(year FROM {age}) * 48 + EXTRACT(month FROM {age}) * 4 + EXTRACT(day FROM {age}) / 7"
+        unit = f"EXTRACT(days FROM ({end} - {start})) / 7"
```
georgesittas, line 78: This must've been an oversight on my end when I implemented this logic... The fix was motivated by an [SO thread](https://stackoverflow.com/a/24075015), after I found out that the results were incorrect when transpiling a given test SQL query from Redshift to Postgres

PR #2591, opened 2023-11-23
Fix(tsql): generate DATEPART when the format is quarter

### LINKED ISSUE #2586: Inaccurate sql generation for DATEPART in T-SQL
In TSQL queries while reading DATEPART function seems to be converted to FORMAT function, which would work fine but 'quarter' does not seem to be part of [format types](https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings). 
Code snippet:
`quarter_query = """
SELECT
    YEAR(sale_date) AS Year,
    DATEPART(QUARTER, sale_date) AS Quarter,
    SUM(sale_price) AS QuarterlySales
FROM
    Sales_AI
GROUP BY
    YEAR(sale_date),
    DATEPART(QUARTER, sale_date)
ORDER BY
    Year, Quarter
"""
print(parse_one(quarter_query, 'tsql').sql(dialect='tsql'))
`
Output: `"SELECT YEAR(sale_date) AS Year, FORMAT(CAST(sale_date AS DATETIME2), 'quarter') AS Quarter, SUM(sale_price) AS QuarterlySales FROM Sales_AI GROUP BY YEAR(sale_date), FORMAT(CAST(sale_date AS DATETIME2), 'quarter') ORDER BY Year, Quarter"`
Worked fine when time period was month. 
Would also like to know if there are any resources for creating a custom dialect to let datepart stay as datepart after parsing.

georgesittas:
Fixes #2586

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -153,6 +153,11 @@ def _format_sql(self: TSQL.Generator, expression: exp.NumberToStr | exp.TimeToSt
             )
         )
     )
+
+    # There is no format for "quarter"
+    if fmt.name.lower() == "quarter":
+        return self.func("DATEPART", "QUARTER", expression.this)
+
     return self.func("FORMAT", expression.this, fmt, expression.args.get("culture"))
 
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -245,9 +250,6 @@ class TSQL(Dialect):
 
     TIME_MAPPING = {
         "year": "%Y",
-        "qq": "%q",
-        "q": "%q",
-        "quarter": "%q",
```
georgesittas, line 250: There's no `%q` format in Python's `strftime`

tobymao (approved):

PR #2602, opened 2023-11-28
refactor!: avoid unnecessary copying in normalization

tobymao:


sqlglot/schema.py
```diff
@@ -451,16 +447,17 @@ def normalize_name(
...
-) -> str:
+) -> exp.Identifier:
     if isinstance(identifier, str):
         identifier = exp.parse_identifier(identifier, dialect=dialect)
 
-    if not normalize:
-        return identifier.name
-
     # This can be useful for normalize_identifier
     identifier.meta["is_table"] = is_table
```
georgesittas, line 455: I think this needs to be moved after the following branch?

tobymao: why?

georgesittas: Because IIRC it's only needed for normalizations, right? If we won't normalize then why should we mutate here?

georgesittas (approved):
Seems good, I understand the "unnecessary" copying refers to the previous `_normalize_table` implementation, right? So now we only copy if we need to normalize?

tobymao:
right, also converting identifiers to string and reparsing them

PR #2614, opened 2023-12-01
Fix: generate UnixToTime correctly (spark, bq, presto, snowflake, duckdb)

georgesittas:
cc @j1ah0ng

References:
- https://duckdb.org/docs/sql/functions/timestamp.html
- https://docs.snowflake.com/en/sql-reference/functions/to_timestamp
- https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#timestamp_seconds
- https://prestodb.io/docs/current/functions/datetime.html
- https://spark.apache.org/docs/latest/api/sql/index.html


sqlglot/dialects/presto.py
```diff
@@ -169,6 +169,22 @@ def _first_last_sql(self: Presto.Generator, expression: exp.First | exp.Last) ->
...
+def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str:
+    scale = expression.args.get("scale")
+    timestamp = self.sql(expression, "this")
+    if scale in {None, exp.UnixToTime.SECONDS}:
+        return rename_func("FROM_UNIXTIME")(self, expression)
+    if scale == exp.UnixToTime.MILLIS:
+        return f"FROM_UNIXTIME(CAST({timestamp} AS DOUBLE) / 1000)"
+    if scale == exp.UnixToTime.MICROS:
+        return f"FROM_UNIXTIME(CAST({timestamp} AS DOUBLE) / 1000000)"
+    if scale == exp.UnixToTime.NANOS:
```
tobymao, line 181: many nanos will lose precision if you use a double, you probably want to cast to a bigint

tobymao (approved):

PR #2613, opened 2023-11-30
Fix: preserve alias quotes in explode_to_unnest

georgesittas:
Addresses https://github.com/tobymao/sqlglot/issues/2607#issuecomment-1834410405

cc @cpcloud

PR #2615, opened 2023-12-01
Fix!: qualify columns added in explode to unnest transformation

georgesittas:
Addresses https://github.com/tobymao/sqlglot/pull/2613#issuecomment-1834716818, TL;DR:

```python
>>> sqlglot.transpile('WITH t(x, "value") AS (SELECT [1,2,3], 1) SELECT UNNEST(t.x) as "value" FROM t', 'duckdb', 'snowflake')
['WITH t(x, "value") AS (SELECT [1, 2, 3], 1) SELECT IFF(pos = pos_2, "value", NULL) AS "value" FROM t, TABLE(FLATTEN(INPUT => ARRAY_GENERATE_RANGE(0, (GREATEST(ARRAY_SIZE(t.x)) - 1) + 1))) AS _u(seq, key, path, index, pos, this) CROSS JOIN TABLE(FLATTEN(INPUT => t.x)) AS _u_2(seq, key, path, pos_2, "value", this) WHERE pos = pos_2 OR (pos > (ARRAY_SIZE(t.x) - 1) AND pos_2 = (ARRAY_SIZE(t.x) - 1))']
```

In the transpiled query, `"value"` is ambiguous.

@tobymao can you double-check that I qualified these correctly? I ran several queries against their corresponding engines and the results checked out but there are two many identifiers and I'm not 100% sure I got this right.

cc @cpcloud


tobymao (approved):

cpcloud (approved):
LGTM!

cpcloud:
I tried this PR out with the ibis refactor, and now all the generated SQL executes (still working out some issues though)

PR #2618, opened 2023-12-01
Fix(tsql, teradata): Distinct goes before top

treysp:


georgesittas (approved):

sqlglot/generator.py
```diff
@@ -1985,10 +1985,11 @@ def select_sql(self, expression: exp.Select) -> str:
                     )
                 kind = ""
 
+        top_distinct = f"{distinct}{hint}{top}" if self.LIMIT_IS_TOP else f"{top}{hint}{distinct}"
```
georgesittas, line 1988: I'd add a comment here and explain that `LIMIT_IS_TOP` is used as a proxy for the `DISTINCT` being put before `TOP` because only Teradata and T-SQL set this flag and they both require that to be true.

georgesittas (approved):

PR #2619, opened 2023-12-01
Fix(duckdb): arrays are 1-indexed

### LINKED ISSUE #2617: DuckDB bracket operator is 1-indexed
**Before you file an issue**
- Make sure you specify the "read" dialect eg. parse_one(sql, read="spark") Bigquery, for example
- Check if the issue still exists on main yes

DuckDB arrays are 1-indexed. When providing a literal to the subscript/bracket operator the expected output for a transpilation of
```sql
SELECT some_arr[0] AS first FROM blah
```
should render to
```sql
SELECT some_arr[1] AS first FROM blah
```
in DuckDB. However, this is not the case:
```python
>>> transpile("SELECT some_arr[0] AS first FROM blah", read="bigquery", write="duckdb")
['SELECT some_arr[0] AS first FROM blah']
```

I attempted to patch this by incrementing in the case that `exp.Bracket.expressions` is of length 1 and type `Literal`, but this does not seem to be directionally correct (do you still increment in the case where the subscript is a column?). I think there are some semantics concerns here which I'm not well-versed enough to address.

**Official Documentation**

[DuckDB array documentation](https://duckdb.org/docs/sql/data_types/array.html)

georgesittas:
Fixes #2617

### DIFF
--- a/tests/dialects/test_duckdb.py
+++ b/tests/dialects/test_duckdb.py
@@ -163,6 +163,11 @@ class TestDuckDB(Validator):
             read={
                 "bigquery": "SELECT some_arr[0] AS first FROM blah",
             },
+            write={
+                "bigquery": "SELECT some_arr[0] AS first FROM blah",
+                "duckdb": "SELECT some_arr[1] AS first FROM blah",
+                "presto": "SELECT some_arr[1] AS first FROM blah",
+            },
         )
         self.validate_all(
             "SELECT * FROM produce PIVOT(SUM(sales) FOR quarter IN ('Q1', 'Q2'))",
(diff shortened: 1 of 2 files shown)


tests/dialects/test_duckdb.py
```diff
@@ -1,11 +1,14 @@
...
+
 from sqlglot import ErrorLevel, UnsupportedError, exp, parse_one, transpile
 from tests.dialects.test_dialect import Validator
 
 
 class TestDuckDB(Validator):
     dialect = "duckdb"
 
-    def test_duckdb(self):
+    @patch("sqlglot.helper.logger")
```
tobymao, line 10: why are you patching the logger? isolate the logger override to a with statement

tobymao: looks like you did this for presto, which is not good, that should be cleaned up too

tobymao: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs

PR #2622, opened 2023-12-04
Fix!: parse functions with positional args in exp.func

### LINKED ISSUE #2621: BigQuery log argument order is incorrect

```
In [8]: import sqlglot as sg

In [9]: sg.__version__
Out[9]: '19.7.1.dev13'

In [10]: print(sg.func("log", sg.to_identifier('x'), 2).sql(dialect="bigquery"))
LOG(2, x)
```

**Official Documentation**

https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#log

### LINKED ISSUE #2631: sql generation is broken

```
In [4]: import sqlglot as sg

In [5]: sg.func("concat", sg.exp.convert("a")).sql("duckdb")
...
TypeError: 'Literal' object is not iterable
```

Bisecting `main` against 019e0e5ba4b5df1ef3b34510c2fa07f8623af364, which doesn't have this issue it looks like 6e71c348eab63125e412382381acbcbd79efac43 is the first bad commit.

georgesittas:
We previously used to construct functions with positional arguments by passing them directly to the corresponding `from_arg_list` method.

This was problematic because for some functions the argument order varies between dialects (e.g. `LOG`) and since we weren't actually parsing the function but instead constructing it directly, the produced AST was wrong.

Fixes #2621
Fixes #2631

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5275,7 +5275,7 @@ def _norm_arg(arg):
 
 
 ALL_FUNCTIONS = subclasses(__name__, Func, (AggFunc, Anonymous, Func))
-NAME_TO_FUNC = {name: func for func in ALL_FUNCTIONS for name in func.sql_names()}
+FUNCTION_BY_NAME = {name: func for func in ALL_FUNCTIONS for name in func.sql_names()}
 
 
 # Helpers
@@ -6630,7 +6630,7 @@ def func(name: str, *args, copy: bool = True, dialect: DialectType = None, **kwa
         elif constructor.__name__ == "from_arg_list":
             function = constructor.__self__(**kwargs)  # type: ignore
         else:
-            constructor = NAME_TO_FUNC.get(name.upper())
+            constructor = FUNCTION_BY_NAME.get(name.upper())
             if constructor:
                 function = constructor(**kwargs)
             else:
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -109,7 +109,7 @@ class Parser(metaclass=_Parser):
     """
 
     FUNCTIONS: t.Dict[str, t.Callable] = {
-        **{name: func.from_arg_list for name, func in exp.NAME_TO_FUNC.items()},
+        **{name: func.from_arg_list for name, func in exp.FUNCTION_BY_NAME.items()},
         "CONCAT": parse_concat,
         "CONCAT_WS": parse_concat_ws,
         "DATE_TO_DATE_STR": lambda args: exp.Cast(


sqlglot/expressions.py
```diff
@@ -6621,14 +6622,20 @@ def func(name: str, *args, dialect: DialectType = None, **kwargs) -> Func:
...
+    converted: t.List[Expression] = [maybe_parse(arg, dialect=dialect, copy=copy) for arg in args]
+    kwargs = {key: maybe_parse(value, dialect=dialect, copy=copy) for key, value in kwargs.items()}
 
     parser = Dialect.get_or_raise(dialect)().parser()
     from_args_list = parser.FUNCTIONS.get(name.upper())
 
     if from_args_list:
-        function = from_args_list(converted) if converted else from_args_list.__self__(**kwargs)  # type: ignore
+        if converted:
+            function: Func = maybe_parse(
```
tobymao, line 6633: i don't think this is the right approach, instead, get the FUNCTIONS from parser and call that so you don't have to call sql and reparse

sqlglot/generator.py
```diff
@@ -2721,10 +2721,13 @@ def trycast_sql(self, expression: exp.TryCast) -> str:
         return self.cast_sql(expression, safe_prefix="TRY_")
 
     def log_sql(self, expression: exp.Log) -> str:
```
tobymao, line 2723: why is this changing?

georgesittas: Because it relies on the order of `expression.args` which is incorrect. For example if you do something like `log = exp.Log(expression=..., this=...)` then the arg order won't match the order in arg types and we'll get unexpected behavior. In the context of this PR, the second test I added initially failed because of this.

sqlglot/expressions.py
georgesittas: Hmm, isn't that what we already do though? It doesn't seem sufficient, I believe you have to hit the parser for e.g. functions that exist in `FUNCTION_PARSERS`, otherwise we won't use that "special parsing" logic and the AST could end up being corrupted (like in the case of `LOG`).

tobymao: i don't think you can safely call function_parsers because they don't necessarily expect commas, i think you can only use FUNCTIONS whichwould handle arg order correctly

georgesittas: Yeah I didn't mean to call function parsers directly, I meant let the parser figure out what parsing method to use by calling `maybe_parse`. The issue with `FUNCTIONS` is that they don't have special logic for "weird" functions such as `LOG` etc to figure out the correct order. See for example the defn. of `LOG`: https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L4566

georgesittas: Essentially `FUNCTIONS` doesn't have all sufficient information to figure out how to route the `exp.func` arguments to `args`.

georgesittas: I'm not sure I understand the suggested path to solving the issue at hand. Are you thinking about moving `_parse_logarithm` to `FUNCTIONS` instead of `FUNCTION_PARSERS`?

georgesittas:
@tobymao did a refactor based on what we discussed, please take another look. I'll see if I need to leave any inline comments to explain non-obvious changes..

sqlglot/dialects/duckdb.py
```diff
@@ -215,15 +221,8 @@ class Parser(parser.Parser):
...
-            **parser.Parser.FUNCTION_PARSERS,
-            "DECODE": lambda self: self.expression(
-                exp.Decode, this=self._parse_conjunction(), charset=exp.Literal.string("utf-8")
-            ),
-            "ENCODE": lambda self: self.expression(
-                exp.Encode, this=self._parse_conjunction(), charset=exp.Literal.string("utf-8")
-            ),
-        }
+        FUNCTION_PARSERS = parser.Parser.FUNCTION_PARSERS.copy()
+        FUNCTION_PARSERS.pop("DECODE", None)
```
georgesittas, line 225: We have `"DECODE"` in the base parser's `FUNCTION_PARSERS` which would take precedence over `FUNCTIONS`. This wasn't previously an issue because DuckDB overrode `"DECODE"` so we'd hit the correct logic.

sqlglot/expressions.py
```diff
@@ -6613,14 +6614,25 @@ def func(name: str, *args, dialect: DialectType = None, **kwargs) -> Func:
...
-    from_args_list = parser.FUNCTIONS.get(name.upper())
+    converted: t.List[Expression] = [maybe_parse(arg, dialect=dialect, copy=copy) for arg in args]
+    kwargs = {key: maybe_parse(value, dialect=dialect, copy=copy) for key, value in kwargs.items()}
 
-    if from_args_list:
-        function = from_args_list(converted) if converted else from_args_list.__self__(**kwargs)  # type: ignore
+    constructor = dialect.parser_class.FUNCTIONS.get(name.upper())
+    if constructor:
+        if converted:
+            if "dialect" in constructor.__code__.co_varnames:
```
georgesittas, line 6625: Initially I had moved the `__code__.co_varnames` into `helper.py` as `parameter_names` but I wasn't sure if people would like that because of possible perf. implications.

sqlglot/expressions.py
```diff
@@ -6613,14 +6614,25 @@ def func(name: str, *args, dialect: DialectType = None, **kwargs) -> Func:
...
+                function = constructor(converted, dialect=dialect)
+            else:
+                function = constructor(converted)
+        elif constructor.__name__ == "from_arg_list":
+            function = constructor.__self__(**kwargs)  # type: ignore
```
georgesittas, line 6630: We had a bug here: if someone passed in kwargs for the arguments and the callback fetched from `FUNCTIONS` is not the `from_arg_list` method, then we would fail on the `__self__` attribute access. See:

```python
>>> exp.func("glob", this="a", expression="b")
  File "sqlglot/expressions.py", line 6623, in func
    function = from_args_list(converted) if converted else from_args_list.__self__(**kwargs)  # type: ignore
AttributeError: 'function' object has no attribute '__self__'. Did you mean: '__call__'?
```

(`"GLOB"` is [mapped](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L81) to `lambda args: …` instead of a `from_arg_list method`)

tobymao: why can't we just load the glob with kwargs?

georgesittas: Because currently you don't have easy access to the expression corresponding to glob (i.e. `Glob`) through `Parser.FUNCTIONS`.

If there was way to easily map the input function name to the corresponding expression type, perhaps that would solve the issue because we could do that in the `else` branch as a fallback. Maybe make a new constant map for function name -> `Func` subclass to do this?

tobymao: yes, just look for the expression?

georgesittas: Yeah `LOG` is a valid example. I'll check out the constant map idea, like we define `ALL_FUNCTIONS` in expression.

georgesittas: Take a look at https://github.com/tobymao/sqlglot/pull/2622/commits/0945894c7df0477c707806574871d2cdd6609bc8

sqlglot/parser.py
```diff
@@ -3790,9 +3826,15 @@ def _parse_function_call(
             args = self._parse_csv(lambda: self._parse_lambda(alias=alias))
 
             if function and not anonymous:
-                func = self.validate_expression(function(args), args)
+                if "dialect" in function.__code__.co_varnames:
```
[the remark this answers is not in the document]
georgesittas: Chatted in Slack, leaving these for future reference:

Main branch:
```python
➜  benchmarks git:(main) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01295 (1.0) | 0.00790 (0.609) |
|           short |   0.00072 (1.0) | 0.00056 (0.771) |
|            long |   0.00994 (1.0) | 0.00675 (0.679) |
|           crazy |   0.03235 (1.0) | 0.02202 (0.680) |
➜  benchmarks git:(main) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01207 (1.0) | 0.01311 (1.085) |
|           short |   0.00071 (1.0) | 0.00056 (0.783) |
|            long |   0.01045 (1.0) | 0.00687 (0.657) |
|           crazy |   0.03237 (1.0) | 0.02214 (0.684) |
➜  benchmarks git:(main) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01264 (1.0) | 0.00928 (0.734) |
|           short |   0.00071 (1.0) | 0.00055 (0.786) |
|            long |   0.01090 (1.0) | 0.00716 (0.657) |
|           crazy |   0.03242 (1.0) | 0.02205 (0.680) |
➜  benchmarks git:(main) python
>>> from timeit import timeit
>>>
>>> small = " + ".join("AVG(x)" for _ in range(4))
>>> large = " + ".join("AVG(x)" for _ in range(400))
>>>
>>> timeit(lambda: parse_one(small), number=10)
0.005644832999678329
>>> timeit(lambda: parse_one(small), number=100)
0.030172165999829303
>>> timeit(lambda: parse_one(small), number=1000)
0.19596233400079655
>>> timeit(lambda: parse_one(large), number=10)
0.18065029100398533
>>> timeit(lambda: parse_one(large), number=100)
1.6704134159954265
>>> timeit(lambda: parse_one(large), number=1000)
16.47456775000319
```

This branch:
```python
➜  benchmarks git:(main) gco jo/fix_func_creation
Switched to branch 'jo/fix_func_creation'
Your branch is up to date with 'origin/jo/fix_func_creation'.
➜  benchmarks git:(jo/fix_func_creation) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01109 (1.0) | 0.00953 (0.859) |
|           short |   0.00072 (1.0) | 0.00056 (0.786) |
|            long |   0.01009 (1.0) | 0.00682 (0.675) |
|           crazy |   0.03215 (1.0) | 0.02226 (0.692) |
➜  benchmarks git:(jo/fix_func_creation) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01574 (1.0) | 0.01143 (0.726) |
|           short |   0.00072 (1.0) | 0.00056 (0.772) |
|            long |   0.01057 (1.0) | 0.00683 (0.646) |
|           crazy |   0.03278 (1.0) | 0.02203 (0.672) |
➜  benchmarks git:(jo/fix_func_creation) python bench.py
|           Query |         sqlglot |         sqltree |
| --------------- | --------------- | --------------- |
|            tpch |   0.01491 (1.0) | 0.01106 (0.741) |
|           short |   0.00159 (1.0) | 0.00056 (0.350) |
|            long |   0.01025 (1.0) | 0.00681 (0.663) |
|           crazy |   0.03208 (1.0) | 0.02282 (0.711) |
➜  benchmarks git:(jo/fix_func_creation) python
>>> from timeit import timeit
>>>
>>> small = " + ".join("AVG(x)" for _ in range(4))
>>> large = " + ".join("AVG(x)" for _ in range(400))
>>>
>>> timeit(lambda: parse_one(small), number=10)
0.0020670000012614764
>>> timeit(lambda: parse_one(small), number=100)
0.018444124994857702
>>> timeit(lambda: parse_one(small), number=1000)
0.1749117090002983
>>> timeit(lambda: parse_one(large), number=10)
0.16667683300329372
>>> timeit(lambda: parse_one(large), number=100)
1.7125379589997465
>>> timeit(lambda: parse_one(large), number=1000)
16.948921541996242
```

tobymao (approved):

PR #2627, opened 2023-12-05
Feat!: introduce Dialect settings, make MySQL case-sensitive by default

georgesittas:
This introduces a way for SQLGlot users to specify `Dialect`-level settings, by including them in the dialect string: `"dialect, key1 = value1, ..."`.

Now, `Dialect.get_or_raise` will return a `Dialect` instance instead of a type, to ensure that the user-supplied settings override the defaults.

The motivation behind this is that we needed to provide a way to override the normalization strategy of MySQL, because we realized that it depends on the underlying OS and hence a single default behavior cannot capture all possible strategies employed by different systems.

This feature creates a way to add more dialect-specific settings in the future, such as the dialect's version etc.

Finally, MySQL's default normalization strategy has been changed so that all identifiers are treated as case-sensitive, regardless of being quoted. This matches MySQL's behavior on Linux systems, so we decided to use that as a default. Users that require a different strategy can simply specify:

    dialect = "mysql, normalization_strategy = lowercase"

This will match MySQL's behavior in MacOS and Windows systems, for example.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -504,7 +504,6 @@ class Tokenizer(metaclass=_Tokenizer):
     QUOTES: t.List[t.Tuple[str, str] | str] = ["'"]
     STRING_ESCAPES = ["'"]
     VAR_SINGLE_TOKENS: t.Set[str] = set()
-    ESCAPE_SEQUENCES: t.Dict[str, str] = {}
 
     # Autofilled
     _COMMENTS: t.Dict[str, str] = {}


sqlglot/dialects/dialect.py
```diff
@@ -46,6 +49,31 @@ class Dialects(str, Enum):
...
+class NormalizationStrategy(str, Enum):
+    """Specifies the strategy according to which identifiers should be normalized."""
+
+    LOWERCASE = "lowercase"  # Unquoted identifiers are lowercased
+    UPPERCASE = "uppercase"  # Unquoted identifiers are uppercased
+    CASE_SENSITIVE = "case_sensitive"  # Always case-sensitive, regardless of quotes
+    CASE_INSENSITIVE = "case_insensitive"  # Always case-insensitive, regardless of quotes
+
+    @property
+    def is_lowercase(self):
```
tobymao, line 61: i don't like this pattern of properties, just use the enums

sqlglot/dialects/dialect.py
```diff
@@ -223,26 +253,73 @@ class Dialect(metaclass=_Dialect):
...
-        return hash(type(self))
+    @t.overload
+    @classmethod
+    def get_or_raise(cls, dialect: DialectType, instance: Literal[True]) -> Dialect:
+        ...
 
+    @t.overload
     @classmethod
-    def get_or_raise(cls, dialect: DialectType) -> t.Type[Dialect]:
+    def get_or_raise(cls, dialect: DialectType, instance: Literal[False]) -> t.Type[Dialect]:
```
tobymao, line 268: why would ever want a non instance? let's just make it simpler and not have instance flag

sqlglot/dialects/dialect.py
```diff
@@ -223,26 +253,73 @@ class Dialect(metaclass=_Dialect):
...
-            return dialect.__class__
+            dialect = cls
+        elif isinstance(dialect, _Dialect):
+            dialect = dialect
+        elif isinstance(dialect, str):
+            dialect_arg = dialect
+            dialect, *kv_pairs = dialect.split(",")
+
+            try:
+                kwargs = {kv[0].strip(): kv[1].strip() for kv in (kv.split("=") for kv in kv_pairs)}
```
tobymao, line 305: for k, v in (

sqlglot/generator.py
```diff
@@ -395,6 +395,7 @@ class Generator:
     NULL_ORDERING = "nulls_are_small"
 
     can_identify: t.Callable[[str, str | bool], bool]
```
tobymao, line 397: maybe let's just pass the dialect would that simplify the code of having to set this on dialect init?

sqlglot/dialects/dialect.py
georgesittas: I don't have a strong opinion on this, I just started off with the `instance` flag to make sure we've discussed about the performance tradeoffs of always instantiating vs returning a type (previous behavior)

sqlglot/optimizer/qualify_columns.py
```diff
@@ -40,7 +40,7 @@ def qualify_columns(
     """
     schema = ensure_schema(schema)
     infer_schema = schema.empty if infer_schema is None else infer_schema
-    pseudocolumns = Dialect.get_or_raise(schema.dialect).PSEUDOCOLUMNS
+    pseudocolumns = Dialect.get_or_raise(schema.dialect, instance=False).PSEUDOCOLUMNS
```
tobymao, line 43: schema.dialect should already be an instance so we don't have to reinstantiate it

sqlglot/dialects/dialect.py
georgesittas: As in `<my_enum>.value == ...`? It seems a little too verbose :P

sqlglot/dialects/dialect.py
tobymao: yea but in most places we instantiate already

sqlglot/__init__.py
```diff
@@ -83,7 +83,7 @@ def parse(
     Returns:
         The resulting syntax tree collection.
     """
-    dialect = Dialect.get_or_raise(read or dialect)()
+    dialect = Dialect.get_or_raise(read or dialect)
     return dialect.parse(sql, **opts)
```
tobymao, line 87: Dialect.get_or_raise(read or dialect).parse(...)

sqlglot/__init__.py
```diff
@@ -159,6 +159,6 @@ def transpile(
     """
     write = (read if write is None else write) if identity else write
     return [
-        Dialect.get_or_raise(write)().generate(expression, copy=False, **opts) if expression else ""
+        Dialect.get_or_raise(write).generate(expression, copy=False, **opts) if expression else ""
```
tobymao, line 162: move the dialect outside of the loop

sqlglot/dialects/dialect.py
tobymao: no, you can just check strategy is Strategy.LOWERCASE etc

sqlglot/dialects/dialect.py
georgesittas: Good catch, thanks

sqlglot/generator.py
georgesittas: I'll have a look and see if it simplifies it

sqlglot/dialects/dialect.py
georgesittas: Ok

sqlglot/dataframe/sql/dataframe.py
```diff
@@ -297,27 +297,38 @@ def _get_select_expressions(
...
         return select_expressions
 
-    def sql(
-        self, dialect: t.Optional[DialectType] = None, optimize: bool = True, **kwargs
-    ) -> t.List[str]:
+    def sql(self, dialect: DialectType = None, optimize: bool = True, **kwargs) -> t.List[str]:
         from sqlglot.dataframe.sql.session import SparkSession
 
-        if dialect and Dialect.get_or_raise(dialect)() != SparkSession().dialect:
+        dialect_arg = dialect
```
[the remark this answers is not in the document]
georgesittas: I liked this approach more because it's a local change as opposed to changing all other places where we reference `dialect` and the intent is made explicit (i.e. we only use it for error reporting). It's also just a couple of places, here and the `get_or_raise` method. Fine either way, just wanted to share context.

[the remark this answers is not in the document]
tobymao: ok fine

georgesittas:
@tobymao addressed PR comments in https://github.com/tobymao/sqlglot/pull/2627/commits/6926365ff21dab6b61a27dc83d61d83678603b76

sqlglot/dataframe/sql/dataframe.py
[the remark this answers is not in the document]
georgesittas: Addressed this as well after all, see:
- https://github.com/tobymao/sqlglot/pull/2627/commits/6926365ff21dab6b61a27dc83d61d83678603b76#diff-fbb39ad01c051fb9448d4194537b53f8ff1861201921bfef2ef541757365a197R262
- https://github.com/tobymao/sqlglot/pull/2627/commits/54f87d84c18669ad1a5b222516c8b0a29f780591

sqlglot/dialects/dialect.py
```diff
@@ -46,6 +46,15 @@ class Dialects(str, Enum):
     Doris = "doris"
 
 
+class NormalizationStrategy(str, Enum):
+    """Specifies the strategy according to which identifiers should be normalized."""
+
+    LOWERCASE = "lowercase"  # Unquoted identifiers are lowercased
```
tobymao, line 52: why not use autoname like tokentype

georgesittas: sure, should we do that for `Dialects` as well or is there a reason we've hardcoded the enum values there?

tobymao: do they map one to one? if so then yea

sqlglot/generator.py
georgesittas: Refactored

sqlglot/dialects/dialect.py
georgesittas: refactored the normalization strategy but not the dialects because `DIALECT` maps to `""` so it's not 1-1

sqlglot/dialects/dialect.py
```diff
@@ -256,36 +283,71 @@ def format_time(
...
+        else:
+            self.normalization_strategy = NormalizationStrategy(normalization_strategy.upper())
+
+    def __eq__(self, other: t.Any) -> bool:
+        # Does not currently take dialect state into account
+        return type(self) == other
+
+    def __hash__(self) -> int:
+        # Does not currently take dialect state into account
+        return hash(type(self))
```
georgesittas, line 300: @tobymao can you take a look at these two methods? What do you think their behavior should be?

I tried to change `__eq__` to take `self.normalization_strategy` into account but then if we wanted to change `__hash__` to match that, `Snowflake() in {"snowflake"}` would yield `False` (i.e. we have tests that failed)

georgesittas:
This is ready for another review, I double-checked the constants and it seems ok.

sqlglot/dialects/dialect.py
georgesittas: Chatted in Slack: this behavior is fine for now, i.e. we won't take dialect state into account when comparing.

PR #2641, opened 2023-12-08
Tokenizer implementation in Rust

izeigerman:
This update introduce the native implementation of the Tokenizer written in Rust. All parsing features available in the Python's version have been implemented but haven't been tested using the complete test suite yet.

Here are the remaining items that are out of the scope of this PR and will be addressed in future PRs:

1. Codegen the Rust's `TokenType` enum from the Python's version to ensure that new tokens are added in one place only.
2. A configuration flag which enables the native implementation in all parsers.
3. Configure tests to run with the native implementation enabled.
4. Update CI/CD to build and release the native package for all supported platforms.
5. Add some unit tests to the rust codebase to cover some internals (eg. trie implementation).
6. Refactor the Rust code to  make it more idiomatic and to clean up python-related workarounds.


sqlglot/tokens.py
```diff
@@ -1246,3 +1246,45 @@ def _extract_string(self, delimiter: str, escapes=None) -> str:
...
         return text
+
+    @classmethod
+    def tokenize_native(cls, sql: str) -> t.List[Token]:
+        from sqlglotrs import (  # type: ignore
+            Tokenizer as RsTokenizer,
+            TokenizerSettings as RsTokenizerSettings,
+        )
+
+        settings = RsTokenizerSettings(
```
tobymao, line 1257: doesn't this mean this will happen every time we call tokenize_native? can we do this on module load instead?

sqlglot/tokens.py
```diff
@@ -1246,3 +1246,45 @@ def _extract_string(self, delimiter: str, escapes=None) -> str:
...
+
+        return [
+            Token(
+                token_type=TokenType[str(token.token_type)],
+                text=token.text,
+                line=token.line,
+                col=token.column,
+                start=token.start,
+                end=token.end,
+                comments=list(token.comments),
```
tobymao, line 1287: what is the type here? why do you have to call list on it

sqlglotrs/src/settings.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub var_single_tokens: HashSet<char>,
+    pub commands: HashSet<TokenType>,
+    pub command_prefix_tokens: HashSet<TokenType>,
+    pub identifiers_can_start_with_digit: bool,
+}
+
+impl TokenizerSettings {
+    pub fn default() -> TokenizerSettings {
+        TokenizerSettings {
+            white_space: HashMap::from([
```
tobymao, line 31: why do we need to copy this over? can python hand it over? i don't want to have to manually update rust everytime i add a token to python

sqlglotrs/src/token.rs
```diff
@@ -0,0 +1,428 @@
+use pyo3::prelude::*;
+use strum_macros::EnumString;
+
+#[allow(non_camel_case_types, dead_code)]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, EnumString)]
+#[pyclass]
+pub enum TokenType {
```
tobymao, line 7: is this auto generated?

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+    settings: TokenizerSettings,
+    keyword_trie: Trie,
+}
+
+#[pymethods]
+impl Tokenizer {
+    #[new]
+    #[pyo3(signature = (settings=None))]
+    pub fn new(settings: Option<TokenizerSettings>) -> Tokenizer {
+        let settings = settings.unwrap_or(TokenizerSettings::default());
```
tobymao, line 18: ah i see, can we just get rid of the default here and make settings mandatory, then we don't need the duplication?

izeigerman: I wanted to keep it just for initial stages, so it's easier to reproduce and debug issues directly in rust without involving the python side.

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+            peek_char: '\0',
+            previous_token_line: None,
+            keyword_trie,
+            settings,
+        }
+    }
+
+    fn tokenize(&mut self) -> Vec<Token> {
+        self.scan(None);
+        std::mem::replace(&mut self.tokens, Vec::new())
```
tobymao, line 89: what is this

sqlglotrs/src/token.rs
izeigerman: Not yet. Going to add this in follow-up PRs. It won't be checked into the repo at that point.

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+        }
+        if !self.tokens.is_empty() && !self.comments.is_empty() {
+            self.tokens
+                .last_mut()
+                .unwrap()
+                .append_comments(&mut self.comments);
+        }
+    }
+
+    fn advance(&mut self, i: usize, alnum: bool, backtrack: bool) {
```
tobymao, line 129: why do you need backtrack? can't you just use negative i?

also, alnum is unnecessary, we can delete all of that

sqlglotrs/src/tokenizer.rs
izeigerman: Instead of copying the vector we're moving it here, but since we can leave the variable uninitialized we replace it with the empty vector.

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+
+        self.is_end = self.current >= self.size;
+        self.current_char = self.char_at(self.current - 1);
+        self.peek_char = if self.is_end {
+            '\0'
+        } else {
+            self.char_at(self.current)
+        };
+
+        if alnum && self.current_char.is_alphanumeric() {
```
tobymao, line 158: this whole section is unnecessary

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+            self.current_char = self.char_at(self.current - 1);
+        }
+    }
+
+    fn peek(&self, i: usize) -> char {
+        let index = self.current + i;
+        if index < self.size {
+            self.char_at(index)
+        } else {
+            '\0'
```
tobymao, line 178: what happens if you have \0 as a char? isn't that confusing? why isn't this just an optional char?

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    fn chars(&self, size: usize) -> String {
+        let start = self.current - 1;
+        let end = start + size;
+        if end <= self.size {
+            self.sql[start..end].iter().collect()
+        } else {
+            String::from("")
```
tobymao, line 188: is this a constant? or is it being allocated every time?

sqlglotrs/src/tokenizer.rs
izeigerman: Rust type system is very rigid. You can't really subtract or add integers of different types. Eg.`column`, `line`, etc are all unsigned types. Passing a negative value requires a signed type. But you can't mix signed and unsigned types in arithmetic.

sqlglotrs/src/tokenizer.rs
izeigerman: Because using Option becomes cumbersome very quickly. You basically need to unwrap it explicitly everywhere you use it.

sqlglotrs/src/tokenizer.rs
tobymao: ok, as long as we delete it in the future

sqlglotrs/src/tokenizer.rs
izeigerman: It's allocated on heap every time. The type system doesn't permit using a constant value here, since String type is mutable in rust.

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+
+    fn scan_bits(&mut self) {
+        self.scan_radix_string(2, TokenType::BIT_STRING);
+    }
+
+    fn scan_hex(&mut self) {
+        self.scan_radix_string(16, TokenType::HEX_STRING);
+    }
+
+    fn scan_radix_string(&mut self, radix: u32, radix_token_type: TokenType) {
```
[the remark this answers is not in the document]
izeigerman: I'm open to suggestions.

sqlglotrs/src/tokenizer.rs
izeigerman: that's the plan

sqlglotrs/src/tokenizer.rs
tobymao: just make them signed 64 bit integers?

sqlglot/tokens.py
izeigerman: I've just checked and it seems like it should be python list already. Will fix this.

sqlglotrs/src/tokenizer.rs
tobymao: it's unnecessarily complexity for python performance

sqlglotrs/src/tokenizer.rs
tobymao: can you ever have \0 in a user submitted char?

sqlglotrs/src/tokenizer.rs
izeigerman: usize is a more fitting type and this is the type returned by `.len()` method for vectors and strings. Changing all these attributes to signed just to support this one case doesn't feel right

izeigerman: Actually I've just found a way to use arithmetic which includes signed and unsigned integers: https://stackoverflow.com/questions/53453628/how-do-i-add-a-signed-integer-to-an-unsigned-integer-in-rust

sqlglotrs/src/token.rs
```diff
@@ -0,0 +1,428 @@
...
+            text,
+            line,
+            column,
+            start,
+            end,
+            comments,
+        }
+    }
+
+    pub fn number(num: u32) -> Token {
```
barakalon, line 389: I don't think these are used in the python Token class? Might be worth leaving out here.

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+            self.advance(1, false, false);
+
+            if self.current_char == '\0' {
+                break;
+            }
+
+            if !self.settings.white_space.contains_key(&self.current_char) {
+                if self.current_char.is_digit(10) {
+                    self.scan_number();
+                } else {
```
barakalon, line 104: I think you can do something like this?
```rust
if {
    self.scan_number();
} else if let Some(identifier_end) = self.settings.identifiers.get(&self.current_char) {
    self.scan_identifier(&identifier_end.to_string());
} else {
    self.scan_keyword();
}
```

sqlglotrs/src/tokenizer.rs
```diff
@@ -0,0 +1,642 @@
...
+        if !self.tokens.is_empty() && !self.comments.is_empty() {
+            self.tokens
+                .last_mut()
+                .unwrap()
+                .append_comments(&mut self.comments);
+        }
+    }
+
+    fn advance(&mut self, i: usize, alnum: bool, backtrack: bool) {
+        match self.settings.white_space.get(&self.current_char) {
```
barakalon, line 130: I feel like the `if let ... else` syntax is better for these little checks.
Seems more idiomatic too: https://rust-lang.github.io/rfcs/0160-if-let.html

sqlglotrs/src/tokenizer.rs
barakalon: +1 for '\0'
I landed on the same conclusion separately.

barakalon (approved):
beautiful

setup.py
```diff
@@ -31,6 +31,7 @@
             "pdoc",
             "pre-commit",
             "types-python-dateutil",
+            "maturin>=1.4,<2.0",
```
georgesittas, line 34: Perhaps we can add a new entry in `extras_require` and move this there, e.g. `dev_rs`?

sqlglot/tokens.py
```diff
@@ -1246,3 +1246,45 @@ def _extract_string(self, delimiter: str, escapes=None) -> str:
...
+
+        settings = RsTokenizerSettings(
+            white_space={k: v.name for k, v in cls.WHITE_SPACE.items()},
+            single_tokens={k: v.name for k, v in cls.SINGLE_TOKENS.items()},
+            keywords={k: v.name for k, v in cls.KEYWORDS.items()},
+            numeric_literals=cls.NUMERIC_LITERALS,
+            identifiers=cls._IDENTIFIERS,
+            identifier_escapes=cls._IDENTIFIER_ESCAPES,
+            string_escapes=cls._STRING_ESCAPES,
+            escape_sequences=cls.ESCAPE_SEQUENCES,  # type: ignore
```
georgesittas, line 1265: Mentioned it in Slack, but leaving this here as well for reference: `cls.ESCAPE_SEQUENCES` and `cls.IDENTIFIERS_CAN_START_WITH_DIGIT` are invalid references after https://github.com/tobymao/sqlglot/commit/be89da3747fa95d98e0d5a28d19d98f5822a8979 because the tokenizer now accesses them through the instance `dialect` attribute, i.e. `self.dialect.ESCAPE_SEQUENCES`.

sqlglotrs/src/token.rs
izeigerman: Yeah, good point.

sqlglotrs/src/tokenizer.rs
izeigerman: TIL, thanks

.github/workflows/python-package.yml
```diff
@@ -17,6 +17,10 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Create and activate virtual environment
+      run: |
+        python -m venv .venv
+        source ./.venv/bin/activate
```
obazna, line 23: Just FYI. In GitHub Ations every `run` step is executed in a separate shell. You need to activate your virtual environment for every `run` step. I.e. run `. .venv/bin/activate` first.

sqlglotrs/src/tokenizer.rs
izeigerman: very nice!

setup.py
izeigerman: Why?

izeigerman: If we plan to run the entire test suite against, why shouldn't it be a part of `dev`?

.github/workflows/python-package.yml
obazna: Without activating the virtual environment in step `Install dependencies` your dependencies will be installed globally but not inside `.venv` dir as you expected. 
Nevertheless `maturin develop` will automatically detect the virtual environment and install `sqlglotrs` package inside it, not globally as other packages. Is that what you expected? Probably not. 

BTW, `maturin develop` is needlessly restrictive by requiring precreated venv. [Quite interesting discussion. ](https://github.com/PyO3/maturin/issues/284)

izeigerman: > BTW, maturin develop is needlessly restrictive by requiring precreated venv.

💯 , quite annoying.

> Nevertheless maturin develop will automatically detect the virtual environment and install sqlglotrs package inside it, not globally as other packages.

Ah, this explains my observations. The implicit nature of this behavior is quite annoying as well.

Thanks for catching this! Will fix this accordingly.

setup.py
georgesittas: ah I see, makes sense

sqlglotrs/src/tokenizer.rs
izeigerman: I figured we should be able to test just fine without these. Removed default settings

tobymao (approved):

PR #2643, opened 2023-12-09
Automatically generate the Rust implementation of the TokenType enum from the Python one

izeigerman:


georgesittas (approved):

gen_rs_token_type.py
```diff
@@ -0,0 +1,21 @@
...
+
+#[allow(non_camel_case_types, dead_code)]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, EnumString)]
+#[pyclass]
+pub enum TokenType {{
+{token_types_str}
+}}
+"""
+
+with open(sys.argv[1], "w") as f:
```
barakalon, line 20:
```python
if __name__ == "__main__":
    ...
```
Just easy to do and safer.

.gitignore
```diff
@@ -142,4 +142,7 @@ spark_warehouse
 sqlglot/_version.py
 
 # Emacs files
-*~
\ No newline at end of file
+*~
+
+# Auto-generated files
+sqlglotrs/src/token_type.rs
```
barakalon, line 148: IMO we should just commit the generated file and add a unit test that verifies the entries are the same across python and rust.

barakalon (approved):

.gitignore
izeigerman: What’s the benefit of this approach?

izeigerman: Feels like the existing approach is a more fool proof, doesn't require a manual intervention on the developer side and/or a know-how on how to update the source file for the test to pass.

barakalon: That's true.
I think the main benefit is the browse-ability of the repo. It can be annoying trying to trace through some unfamiliar codebase when it has generated code.

But I don't know if that's more valuable than developer experience. So nvm.

tobymao (approved):

PR #2644, opened 2023-12-10
Enable tests for the Rust tokenizer

izeigerman:
This PR adds a flag which controls whether to use the native tokenizer or the Python one. It also enables the continuous testing of the native implementation using the existing test suite.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import typing as t
 from enum import auto
 
@@ -9,6 +10,14 @@ from sqlglot.trie import TrieResult, in_trie, new_trie
 
 if t.TYPE_CHECKING:
     from sqlglot.dialects.dialect import DialectType
+
+
+try:
+    import sqlglotrs  # noqa
+
+    USE_NATIVE_TOKENIZER = os.environ.get("SQLGLOT_NATIVE_TOKENIZER", "1") == "1"
+except ImportError:
+    USE_NATIVE_TOKENIZER = False
 
 
 class TokenType(AutoName):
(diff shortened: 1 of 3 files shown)


sqlglot/tokens.py
```diff
@@ -11,6 +12,14 @@
     from sqlglot.dialects.dialect import DialectType
 
 
+try:
+    import sqlglotrs  # noqa
```
georgesittas, line 16: Nit: should we do `# type: ignore` instead here, similar to https://github.com/tobymao/sqlglot/blob/main/sqlglot/tokens.py#L1260? I don't see `noqa` being used anywhere else in the codebase (does it serve a different purpose?), so this comment is mostly for improving consistency.

sqlglot/tokens.py
```diff
@@ -11,6 +12,14 @@
...
+
+    USE_NATIVE_TOKENIZER = os.environ.get("SQLGLOT_NATIVE_TOKENIZER", "1") == "1"
```
georgesittas, line 18: Does this mean we'll use the rust tokenizer by default?

izeigerman: Yes, assuming a user installs the sqlglotrs package which would be an explicit action.

sqlglot/tokens.py
izeigerman: Please note that I didn’t do this because of mypy, but because autoflake would delete an unused import. 

# type applies to mypy only afaik, but will try your suggestion anyway

georgesittas: Ahh that makes sense, thanks for clarifying. Feel free to ignore then.

georgesittas (approved):

tobymao (approved):

sqlglot/tokens.py
izeigerman: Actually `type: ignore` did seem to work. So I was wrong about my assumption. Thanks!

georgesittas: Cool!

PR #2647, opened 2023-12-11
fix(optimizer): coerce DATE_ADD expression to int

barakalon:
Tableau generates queries for MySQL like this:
```sql
DATE_ADD(date_col, INTERVAL TRUNCATE(CAST(string_col AS DECIMAL), 0) DAY)
```


tests/fixtures/optimizer/canonicalize.sql
```diff
@@ -82,6 +82,9 @@ DATE_ADD(CAST("x" AS DATE), 1, 'YEAR');
 DATE_ADD('2023-01-01', 1, 'YEAR');
 DATE_ADD(CAST('2023-01-01' AS DATE), 1, 'YEAR');
 
+DATE_ADD('2023-01-01', 1.0, 'YEAR');
```
tobymao, line 85: can the unit be like hour where floats might make sense?

barakalon:
Hm, might be the wrong approach because of string intervals

tests/fixtures/optimizer/canonicalize.sql
barakalon: Might just do this for presto, which only allows bigint

PR #2648, opened 2023-12-11
fix(presto): coerce DATEADD expression to BIGINT

barakalon:
Take 2.

I'm scared to put this in the optimizer, since we don't handle intervals very well, and there seems to be quite a bit of differences between dialects. I don't love putting it in the generator, but it seems safe:

```
SELECT DATE_ADD('DAY', 1.0, CAST('2023-01-01' AS DATE))
...
trino error: line 1:8: Unexpected parameters (varchar(3), double, date) for function date_add. Expected: date_add(varchar(x), bigint, date), date_add(varchar(x), bigint, time(p) with time zone), date_add(varchar(x), bigint, time(p)), date_add(varchar(x), bigint, timestamp(p) with time zone), date_add(varchar(x), bigint, timestamp(p))
```

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -187,6 +187,10 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
 
 
 def _to_int(expression: exp.Expression) -> exp.Expression:
+    if not expression.type:
+        from sqlglot.optimizer.annotate_types import annotate_types
+
+        annotate_types(expression)
     if expression.type and expression.type.this not in exp.DataType.INTEGER_TYPES:
         return exp.cast(expression, to=exp.DataType.Type.BIGINT)
     return expression
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -188,7 +188,7 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
 
 def _to_int(expression: exp.Expression) -> exp.Expression:
     if expression.type and expression.type.this not in exp.DataType.INTEGER_TYPES:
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/presto.py
```diff
@@ -186,6 +186,12 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
     return ""
 
 
+def _to_int(expression: exp.Expression) -> exp.Expression:
+    if expression.type and expression.type.this not in exp.DataType.INTEGER_TYPES:
```
barakalon, line 190: not sure if we should do this if no type is set.

georgesittas: perhaps you can check if a type is set and if not then call annotate types on the `expression` subtree? see https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/snowflake.py#L595

sqlglot/dialects/presto.py
```diff
@@ -186,6 +186,12 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
...
+        return exp.cast(expression.copy(), to=exp.DataType.Type.BIGINT)
```
georgesittas, line 191: do you need the cast here?

tobymao: you shouldn't copy, here, it's a double copy

barakalon: This is the whole point of the change, right?

barakalon: ah, copy, not cast. whoopsie

sqlglot/dialects/presto.py
barakalon: ah, cool, sure

tobymao (approved):

Issue #2653, opened 2023-12-12
UNION parsing issue with Postgres dialect


J000Z:
It may not limited to Postgres dialect. But I will refer to Postgres document on Union https://www.postgresql.org/docs/current/queries-union.html

Expectations:
- respect left-to-right association when parsing multi unions
- `intersect` should have highest association priority
- suffix `limit` by default applies to the union instead of the last select

Examples:
- `select * from A union select * from B union select * from C` should be parsed to 
```
exp.Union
    - left: exp.Union
        - left: select * from A
        - right: select * from B
    - right: select * from C
```

- `select * from A union select * from B intersect select * from C` should be parsed to 
```
exp.Union
    - left: select * from A
    - right: exp.Intersect
        - left: select * from B
        - right: select * from C
```

- `select * from A union select * from B limit 10` should be parsed to
```
exp.Union
    - left: select * from A
    - right: select * from B
    - limit: 10
```

georgesittas:
Fwiw the second example does work as expected:

```python
>>> import sqlglot
>>> sqlglot.parse_one("select 1 union all select 2 intersect select 1")
(UNION this:
  (SELECT expressions:
    (LITERAL this: 1, is_string: False)), distinct: False, by_name: False, expression:
  (INTERSECT this:
    (SELECT expressions:
      (LITERAL this: 2, is_string: False)), distinct: True, by_name: False, expression:
    (SELECT expressions:
      (LITERAL this: 1, is_string: False))))
```

But yeah, if you reverse the order of the operations then the precedence is not captured correctly:

```python
>>> sqlglot.parse_one("select 1 intersect select 2 union all select 1")
(INTERSECT this:
  (SELECT expressions:
    (LITERAL this: 1, is_string: False)), distinct: True, by_name: False, expression:
  (UNION this:
    (SELECT expressions:
      (LITERAL this: 2, is_string: False)), distinct: False, by_name: False, expression:
    (SELECT expressions:
      (LITERAL this: 1, is_string: False))))
```

tobymao:
all have been addressed except for precedence of intersect, which is not consistente across dialects, i'm going to mark that as out of scope for now, you should really use parenthesis to determine precedence.

PR #2657, opened 2023-12-12
Fix!(parser): attach ORDER, LIMIT, FETCH to set ops instead of rhs query

georgesittas:
Fixes the last item in #2653


georgesittas:
FYI the reason I included `order` and `fetch` is because Postgres mentions the following in their [docs](https://www.postgresql.org/docs/current/queries-union.html):

> You can also surround an individual query with parentheses. **This is important if the query needs to use any of the clauses discussed in following sections, such as LIMIT**. Without parentheses, you'll get a syntax error, or else the clause will be understood as applying to the output of the set operation rather than one of its inputs.

The "following sections" seem to only contain these three args, as far as "query modifiers" go:

Screenshot 2023-12-12 at 4 45 15 PM

georgesittas:
It seems that most dialects follow Postgres' semantics, i.e. they attach `LIMIT` to the union instead of its rhs. One exception is ClickHouse, which seems to attach it to the rhs in both of the following examples:

```sql
with t1 as (select 1 as x union all select 2), t2 as (select 1 as x) select x from t1 union all select x from t2 limit 1;
with t1 as (select 1 as x union all select 2), t2 as (select 1 as x) select x from t1 union all (select x from t2 limit 1);
```

Another exception is T-SQL, because it uses `TOP` instead of `LIMIT`, so you can't have ambiguity as to where it should be attached. We previously arbitrarily attached these modifiers to the rhs and so transpiling towards T-SQL would basically limit it using `TOP`, but now it's dropped. This means that queries that contain `LIMIT` and set ops will have their rightmost `LIMIT` removed when targeting T-SQL.

If needed, I can add a transformation in T-SQL's generation which converts `X UNION Y LIMIT Z` into `SELECT TOP Z * FROM (X UNION Y)`. I'm currently working on adding a flag to handle the ClickHouse case as well (i.e. keep previous behavior for it).

sqlglot/parser.py
```diff
@@ -3259,21 +3263,34 @@ def _parse_set_operations(self, this: t.Optional[exp.Expression]) -> t.Optional[
...
+        comments = self._prev.comments
+        distinct = self._match(TokenType.DISTINCT) or not self._match(TokenType.ALL)
+        by_name = self._match_text_seq("BY", "NAME")
+        expression = self._parse_set_operations(
+            self._parse_select(nested=True, parse_set_operation=False)
+        )
+
+        set_operation_modifiers: t.Dict[str, t.Any] = {}
+        if self.MODIFIERS_ATTACHED_TO_UNION and isinstance(expression, exp.Expression):
+            for arg in self.UNION_MODIFIERS:
```
tobymao, line 3281: can we not do a for loop?

georgesittas: will be a bit verbose / duplicative because we have to check if the args exist etc

georgesittas: so the for loop is just to compact that a bit

tobymao (approved):

PR #2661, opened 2023-12-12
Fix(optimizer): handle case where top-level query is a Subquery in scope

### LINKED ISSUE #2654: optimize doesn't work with parenthesis
```
e = parse_one("""
    with x as (
        select a from t0
    )
    (
        SELECT * FROM x
        UNION ALL
        SELECT * FROM x
        LIMIT 10
    )
    limit 10
""", dialect="postgres")

print(optimize(e).sql(pretty=True))
```
got
```
WITH "x" AS (
  SELECT
    "a"
  FROM "t0"
)
(
  WITH "cte_2" AS (
    SELECT
      *
    FROM "x" AS "x"
    LIMIT 10
  )
  SELECT
    * AS "*"
  FROM "x" AS "x"
  UNION ALL
  SELECT
    "cte_2"."*" AS "*"
  FROM "cte_2" AS "cte_2"
)
LIMIT 10
```
where the columns are not qualified.

georgesittas:
Fixes #2654

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -515,7 +515,10 @@ def _traverse_scope(scope):
     elif isinstance(scope.expression, exp.Union):
         yield from _traverse_union(scope)
     elif isinstance(scope.expression, exp.Subquery):
-        yield from _traverse_subqueries(scope)
+        if scope.is_root:
+            yield from _traverse_select(scope)
+        else:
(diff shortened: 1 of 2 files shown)


georgesittas:
I actually need to rebase this off of #2657 because the optimizer test will change due to the changes in the parser..

barakalon (approved):

tobymao (approved):

PR #2663, opened 2023-12-13
fix(optimizer): eliminate eliminate_unions

barakalon:
whew. This surfaced a few nasty bugs.

### DIFF
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -174,6 +174,22 @@ def _mergeable(outer_scope, inner_scope, leave_tables_isolated, from_or_join):
             for col in inner_projections[selection].find_all(exp.Column)
         )
 
+    def _is_recursive():
+        # Recursive CTEs look like this:
+        #     WITH RECURSIVE cte AS (
+        #       SELECT * FROM x  <-- inner scope
+        #       UNION ALL
+        #       SELECT * FROM cte  <-- outer scope
+        #     )
+        cte = inner_scope.expression.parent
+        node = outer_scope.expression.parent
+
+        while node:
+            if node is cte:
+                return True
+            node = node.parent
+        return False
+
     return (
         isinstance(outer_scope.expression, exp.Select)
         and not outer_scope.expression.is_star
@@ -197,6 +213,7 @@ def _mergeable(outer_scope, inner_scope, leave_tables_isolated, from_or_join):
         )
         and not _outer_select_joins_on_inner_select_join()
         and not _is_a_window_expression_in_unmergable_operation()
+        and not _is_recursive()
     )
 
 
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -290,11 +290,6 @@ class TestExecutor(unittest.TestCase):
                 [(1,), (2,), (3,)],
             ),
             (
-                "SELECT 1 AS a UNION SELECT 2 AS a UNION SELECT 3 AS a",
-                ["a"],
-                [(1,), (2,), (3,)],
-            ),
-            (
                 "SELECT 1 / 2 AS a",
                 ["a"],
                 [
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -95,44 +95,11 @@ def eliminate_subqueries(expression):
 
 
 def _eliminate(scope, existing_ctes, taken):
-    if scope.is_union:
-        return _eliminate_union(scope, existing_ctes, taken)
-
     if scope.is_derived_table:
         return _eliminate_derived_table(scope, existing_ctes, taken)
 
     if scope.is_cte:
         return _eliminate_cte(scope, existing_ctes, taken)
-
-
-def _eliminate_union(scope, existing_ctes, taken):
-    duplicate_cte_alias = existing_ctes.get(scope.expression)
-
-    alias = duplicate_cte_alias or find_new_name(taken=taken, base="cte")
-
-    taken[alias] = scope
-
-    # Try to maintain the selections
-    expressions = scope.expression.selects
-    selects = [
-        exp.alias_(exp.column(e.alias_or_name, table=alias), alias=e.alias_or_name, copy=False)
-        for e in expressions
-        if e.alias_or_name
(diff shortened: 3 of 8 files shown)


sqlglot/optimizer/merge_subqueries.py
```diff
@@ -174,6 +174,26 @@ def _outer_select_joins_on_inner_select_join():
...
+    def _is_recursive():
+        # Recursive CTEs look like this:
+        #     WITH RECURSIVE cte AS (
+        #       SELECT * FROM ...
+        #       UNION ALL
+        #       SELECT *
+        #       FROM cte
+        #     )
+        # In this case, the inner scope is SELECT * FROM ..., NOT the entire CTE.
+        # So we traverse up to the inner scope's parent.
```
georgesittas, line 186: @barakalon what do you mean here? not sure I get it

tests/test_executor.py
```diff
@@ -289,11 +289,6 @@ def test_set_operations(self):
                 ["a"],
                 [(1,), (2,), (3,)],
             ),
-            (
-                "SELECT 1 AS a UNION SELECT 2 AS a UNION SELECT 3 AS a",
-                ["a"],
-                [(1,), (2,), (3,)],
-            ),
```
georgesittas, line 296: Why was this removed?

georgesittas (commented):
Awesome that you eliminated this 😄 🚀 

One question so that I can understand the recursive CTE change a bit better. Also I haven't dug deep into the TPC-DS test changes yet.

georgesittas:
@barakalon it seems like this PR fixes another bug I found earlier..

Main:

```python
>>> execute("with t1 as (select 1 as x union all select 2 as x), t2 as (select 1 as x) select x from t1 union all select x from t2 limit 1")
  File "sqlglot/executor/python.py", line 39, in execute
    contexts[node] = self.scan(node, context)
  File "sqlglot/executor/python.py", line 105, in scan
    context, table_iter = self.scan_table(step)
  File "sqlglot/executor/python.py", line 133, in scan_table
    context = self.context({step.source.alias_or_name: table})
  File "sqlglot/executor/python.py", line 81, in context
    return Context(tables, env=self.env)
  File "sqlglot/executor/context.py", line 30, in __init__
    self.range_readers = {name: table.range_reader for name, table in self.tables.items()}
  File "sqlglot/executor/context.py", line 30, in <dictcomp>
    self.range_readers = {name: table.range_reader for name, table in self.tables.items()}
AttributeError: 'NoneType' object has no attribute 'range_reader'

The above exception was the direct cause of the following exception:

  File "sqlglot/executor/__init__.py", line 88, in execute
    result = PythonExecutor(tables=tables_).execute(plan)
  File "sqlglot/executor/python.py", line 61, in execute
    raise ExecuteError(f"Step '{node.id}' failed: {e}") from e
sqlglot.errors.ExecuteError: Step 'Scan: t2 (4380431776)' failed: 'NoneType' object has no attribute 'range_reader'
```

This branch:

```python
>>> execute("with t1 as (select 1 as x union all select 2 as x), t2 as (select 1 as x) select x from t1 union all select x from t2 limit 1")
x
1
2
1
```

Let's add this test as well? The reason it was previously failing was due to the optimizer producing incorrect SQL:

```python
>>> print(optimize("with t1 as (select 1 as x union all select 2 as x), t2 as (select 1 as x) select x from t1 union all select x from t2 limit 1").sql(pretty=True))
WITH "cte" AS (
  SELECT
    2 AS "x"
), "t1" AS (
  SELECT
    "t2"."x" AS "x"
  FROM "t2" AS "t2"
  UNION ALL
  SELECT
    "cte"."x" AS "x"
  FROM "cte" AS "cte"
), "t2" AS (
  SELECT
    1 AS "x"
)
SELECT
  "t1"."x" AS "x"
FROM "t1"
UNION ALL
SELECT
  "t2"."x" AS "x"
FROM "t2"
LIMIT 1
```

georgesittas:
> EDIT: actually, execute's result in this branch seems iffy as well, the LIMIT 1 should apply to the UNION and so I'd expect a single row to be returned..

https://github.com/tobymao/sqlglot/pull/2665

sqlglot/optimizer/merge_subqueries.py
barakalon: will update comment

tests/test_executor.py
barakalon: Its a duplicate of the test directly above

georgesittas: oh missed that, nice catch 👍

barakalon:
Righto - rebased

tobymao (approved):

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -174,6 +174,22 @@ def _outer_select_joins_on_inner_select_join():
...
+        #       SELECT * FROM x  <-- inner scope
+        #       UNION ALL
+        #       SELECT * FROM cte  <-- outer scope
+        #     )
+        cte = inner_scope.expression.parent
+        node = outer_scope.expression.parent
+
+        while node:
+            if node is cte:
+                return True
```
georgesittas, line 189: Shouldn't we check that the `recursive` flag is set as well?

barakalon: `recursive` is set on `With`, not `CTE`. We _can_, but I don't know if its necessary

georgesittas: I guess I was a little confused about this loop/check being used as a proxy for detecting recursion because of the `_is_recursive` name. Is it sufficient? You can also have a CTE which is non-recursive and contains a `UNION` inside of it.

barakalon: IIUC, a recursive WITH statement can include non-recursive ctes _and_ recursive ctes, so you might still want to merge.
If the outer scope is a child of the inner scope, you know you're recursive. That should be true for all scopes.

georgesittas: I see, makes sense

georgesittas (approved):
LGTM

@barakalon do we cover the recursive cte condition in a test?

barakalon:
Yep, theres an existing test

PR #2668, opened 2023-12-13
Fix(snowflake): refactor location paths

georgesittas:


sqlglot/dialects/snowflake.py
```diff
@@ -372,29 +372,25 @@ def _parse_lateral(self) -> t.Optional[exp.Lateral]:
...
-                    if not self._match_set(self.STAGED_FILE_SINGLE_TOKENS, advance=False):
-                        break
-                    while self._match_set(self.STAGED_FILE_SINGLE_TOKENS):
-                        table_name += self._prev.text
-
-                table = exp.var(table_name)
+            if self._match_text_seq("@", advance=False):
+                table = self._parse_location_path()
             elif self._match(TokenType.STRING, advance=False):
                 table = self._parse_string()
```
tobymao, line 378: else table = None

sqlglot/dialects/snowflake.py
```diff
@@ -372,29 +372,25 @@ def _parse_lateral(self) -> t.Optional[exp.Lateral]:
         def _parse_table_parts(self, schema: bool = False) -> exp.Table:
             # https://docs.snowflake.com/en/user-guide/querying-stage
             table: t.Optional[exp.Expression] = None
```
tobymao, line 374: delete this one

sqlglot/dialects/snowflake.py
```diff
@@ -372,29 +372,25 @@ def _parse_lateral(self) -> t.Optional[exp.Lateral]:
...
                 table = self._parse_string()
 
             if table:
                 file_format = None
                 pattern = None
 
-                if self._match_text_seq("(", "FILE_FORMAT", "=>"):
-                    file_format = self._parse_string() or super()._parse_table_parts()
-                    if self._match_text_seq(",", "PATTERN", "=>"):
+                self._match(TokenType.L_PAREN)
```
tobymao, line 384: is this optional?

sqlglot/dialects/snowflake.py
```diff
@@ -438,17 +434,17 @@ def _parse_alter_table_swap(self) -> exp.SwapTable:
 
         def _parse_location(self) -> exp.LocationProperty:
             self._match(TokenType.EQ)
+            return self.expression(exp.LocationProperty, this=self._parse_location_path())
 
-            parts = [self._parse_var(any_token=True)]
+        def _parse_location_path(self) -> exp.Var:
+            parts = [self._advance_any(ignore_reserved=True)]
```
tobymao, line 440: why do we need this first one?

sqlglot/dialects/snowflake.py
georgesittas: That's right, based on the tests (which were motivated by examples on Snowflake's docs IIRC).

sqlglot/dialects/snowflake.py
georgesittas: because the first token in the path token sequence can be "disconnected" from its previous tokens and so we won't enter the loop and crash in `exp.var("")`.

tobymao (approved):

PR #2669, opened 2023-12-13
Go back to converting the Rust token into the Python one

izeigerman:


sqlglot/tokens.py
```diff
@@ -973,7 +967,19 @@ def tokenize_native(self, sql: str) -> t.List[Token]:
         if not self._NATIVE_TOKENIZER:
             raise SqlglotError("Native tokenizer is not available")
 
+        all_token_types = list(TokenType)
```
tobymao, line 970: this happens every call, isn't that a waste? make this a constant?

tobymao (approved):

PR #2672, opened 2023-12-14
Fix: create .venv when publishing so that maturin doesn't crash

georgesittas:


georgesittas:
Closing in favor of https://github.com/tobymao/sqlglot/pull/2673

PR #2673, opened 2023-12-14
Release pipeline for sqlglotrs

izeigerman:


.github/workflows/python-publish.yml
```diff
@@ -5,7 +5,95 @@ on:
     tags:
       - "v*"
 
+permissions:
+  contents: read
+
+env:
+  python_interpreters: 3.7 3.8 3.9 3.10 3.11
```
tobymao, line 12: can we start testing 3.12?

tobymao (approved):

georgesittas:
Seems like this clashes with https://github.com/tobymao/sqlglot/pull/2672/files? I can close the other one and do everything here in one go

georgesittas (approved):

.github/workflows/python-publish.yml
izeigerman: Let's do this in a separate PR and enable it for the core package as well.

PR #2684, opened 2023-12-15
Feat(presto): add support for TO_CHAR

georgesittas:


sqlglot/dialects/presto.py
```diff
@@ -196,6 +196,17 @@ def _to_int(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def _parse_to_char(args: t.List) -> exp.TimeToStr:
```
tobymao, line 199: sorry, why are we doing teradata here?

georgesittas: Because this is under the "teradata functions" section in both presto's and trino's [docs](https://prestodb.io/docs/current/functions/teradata.html?highlight=to_char#to_char) and the time formats are different from presto's `TIME_MAPPING`, i.e. the format is what Teradata uses.

PR #2686, opened 2023-12-15
fix!: sort comparisons

tobymao:


sqlglot/optimizer/simplify.py
```diff
@@ -131,17 +132,32 @@ def rewrite_between(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+COMPLEMENT_COMPARISONS: t.Dict[t.Type[exp.Expression], t.Type[exp.Expression]] = {
```
georgesittas, line 135: Why is the type hint needed here?

sqlglot/optimizer/simplify.py
```diff
@@ -881,6 +885,23 @@ def simplify_datetrunc_predicate(expression: exp.Expression) -> exp.Expression:
...
+        l, r = expression.this, expression.expression
+        l_column = isinstance(l, exp.Column)
+        r_column = isinstance(r, exp.Column)
+        l_literal = isinstance(l, (exp.Literal, exp.Boolean))
+        r_literal = isinstance(r, (exp.Literal, exp.Boolean))
+
+        if (l_column and not r_column) or (r_literal and not l_literal):
+            return expression
+        if (r_column and not l_column) or (l_literal and not r_literal) or (gen(l) > gen(r)):
+            return INVERSE_COMPARISONS.get(expression.__class__, expression.__class__)(
```
georgesittas, line 899: This looks like a typo? Did you mean `COMPLEMENT_COMPARISONS` or am I misreading this?

tobymao: it's not a typo, you don't want to invert = when sorting 

a = b == b = a

georgesittas: Ah I didn't see that you do `get(expression.__class__, expression.__class__)`. Makes sense, thanks for clarifying.

Perhaps we could also clean up some of these `__class__` attributes by doing `type(expression)`?

PR #2692, opened 2023-12-16
fix(optimizer): CTE aliases overridden in FROM

barakalon:
We need to maintain references to CTE scopes. 

```sql
WITH q AS (
  SELECT
    a
  FROM y
)
SELECT
  a
FROM x AS q  -- q is overridden in this scope
WHERE
  a IN (
    SELECT
      a
    FROM q  -- this is a new subquery scope, but q should still be the CTE
  );
```

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -102,11 +102,11 @@ class Scope:
         """Branch from the current scope to a new, inner scope"""
         return Scope(
             expression=expression.unnest(),
-            sources=sources.copy() if sources else {},
+            sources=sources.copy() if sources else None,
             parent=self,
             scope_type=scope_type,
             cte_sources={**self.cte_sources, **(cte_sources or {})},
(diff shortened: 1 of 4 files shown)


sqlglot/optimizer/scope.py
```diff
@@ -92,13 +96,17 @@ def clear_cache(self):
...
 
-    def branch(self, expression, scope_type, chain_sources=None, **kwargs):
+    def branch(
+        self, expression, scope_type, sources=None, cte_sources=None, lateral_sources=None, **kwargs
+    ):
         """Branch from the current scope to a new, inner scope"""
         return Scope(
             expression=expression.unnest(),
-            sources={**self.cte_sources, **(chain_sources or {})},
+            sources=sources.copy() if sources else {},
```
tobymao, line 105: just pass None here because you already instantiate {} on empty dict, same with lateral_sources

tobymao (approved):

PR #2693, opened 2023-12-16
Set a reference to the TokenType enum in the rust Token object

izeigerman:


sqlglotrs/src/lib.rs
```diff
@@ -13,8 +13,10 @@ pub use self::tokenizer::Tokenizer;
 #[derive(Debug)]
 #[pyclass]
 pub struct Token {
-    #[pyo3(get)]
+    #[pyo3(get, name = "token_type_index")]
```
tobymao, line 16: so this is token type index?

sqlglotrs/src/lib.rs
```diff
@@ -41,6 +43,7 @@ impl Token {
     ) -> Token {
         Python::with_gil(|py| Token {
             token_type,
+            token_type_py: PyNone::get(py).into(),
```
tobymao, line 46: just instantiating a python none value?

tobymao (approved):

sqlglotrs/src/lib.rs
izeigerman: Exactly, yes

sqlglotrs/src/lib.rs
izeigerman: it is, yes. But the rust code base doesn't need to worry about this detail. Do you suggest renaming the attribute?

tobymao: it’s ok

PR #2696, opened 2023-12-18
Fix(dialects):snowflake to_array implementation

wezham:


sqlglot/dialects/snowflake.py
```diff
@@ -162,6 +162,11 @@ def _nullifzero_to_if(args: t.List) -> exp.If:
     cond = exp.EQ(this=seq_get(args, 0), expression=exp.Literal.number(0))
     return exp.If(this=cond, true=exp.Null(), false=seq_get(args, 0))
 
+def _toarray_no_modify_on_array_or_struct(args: t.List) -> exp.If:
+    if isinstance(seq_get(args, 0), exp.Array) or isinstance(seq_get(args, 0), exp.Struct):
```
wezham, line 166: I am not sure if variant is represented as struct nor of the naming conventions

georgesittas:
Hey @wezham, thanks for taking a stab at this but I don't think this approach is quite correct. For example, `VARIANT`s are not handled by the `if isinstance(seq_get(args, 0), exp.Array)` check, but besides that I think this should be handled using type annotations which is a more general solution to the problem. I'll take a stab at this and ping you in the new PR.

wezham:
> Hey @wezham, thanks for taking a stab at this but I don't think this approach is quite correct. For example, `VARIANT`s are not handled by the `if isinstance(seq_get(args, 0), exp.Array)` check, but besides that I think this should be handled using type annotations which is a more general solution to the problem. I'll try to fix it this way and drop a ping when the PR is ready.

No worries @georgesittas I am aware it doesn't cover the variant case I was a bit confused how to implement that. See https://tobiko-data.slack.com/archives/C0448SFS3PF/p1702868766647939

PR #2700, opened 2023-12-18
Fix(snowflake): implement correct parsing semantics for TO_ARRAY

### LINKED ISSUE #2695: Snowflake to_array implemented incorrectly
The docs for snowflake implementation of [to_array](https://docs.snowflake.com/en/sql-reference/functions/to_array) state 

> If the input is an ARRAY, or VARIANT containing an array value, the result is unchanged.
> For NULL or a JSON null input, returns NULL.
> For any other value, the result is a single-element array containing this value.

As we can see below, the input as an array, the result is an array of arrays which is not unchanged. 
```
>>> parse_one("select to_array(['test']) from test;", dialect="snowflake").sql(pretty=True)
"SELECT\n  ARRAY(ARRAY('test'))\nFROM test"
```

```
sqlglot==20.2.0
```

georgesittas:
Fixes #2695

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -33,6 +33,19 @@ def _check_int(s: str) -> bool:
     return s.isdigit()
 
 
+def _parse_to_array(args: t.List) -> exp.Expression:
+    arg = seq_get(args, 0)
+    if isinstance(arg, exp.Expression):
+        from sqlglot.optimizer.annotate_types import annotate_types
+
+        # https://docs.snowflake.com/en/sql-reference/functions/to_array
+        arg = annotate_types(arg)
+        if arg.is_type(exp.DataType.Type.ARRAY):
+            return arg
+
+    return exp.Array.from_arg_list(args)
+
+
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -33,6 +33,21 @@ def _check_int(s: str) -> bool:
...
+    arg = seq_get(args, 0)
+    if isinstance(arg, exp.Expression):
+        from sqlglot.optimizer.annotate_types import annotate_types
+
+        # https://docs.snowflake.com/en/sql-reference/functions/to_array
+        arg = annotate_types(arg)
+        if arg.is_type(exp.DataType.Type.ARRAY):
+            return arg
+        if arg.is_type(exp.DataType.Type.VARIANT):
+            return exp.Anonymous(this="TO_ARRAY", expressions=[arg])
```
georgesittas, line 46: @wezham addressed the `VARIANT` case more conservatively here because I realized it's more complex. The difficulty is that we need to check the type of the _wrapped_ value of the variant, so I think we'd only be able to do that if we checked a few known cases like casts to variants etc where we can access the wrapped value. Didn't seem worth adding that complexity..

PR #2705, opened 2023-12-18
Fix(presto): remove equals sign from CREATE TABLE comment

treysp:
https://prestodb.io/docs/current/sql/create-table.html
https://trino.io/docs/current/sql/create-table.html

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -387,6 +387,7 @@ class Presto(Dialect):
             exp.Right: right_to_substring_sql,
             exp.SafeDivide: no_safe_divide_sql,
             exp.Schema: _schema_sql,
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/presto.py
```diff
@@ -387,6 +387,7 @@ class Generator(generator.Generator):
             exp.Right: right_to_substring_sql,
             exp.SafeDivide: no_safe_divide_sql,
             exp.Schema: _schema_sql,
+            exp.SchemaCommentProperty: lambda self, e: f"COMMENT {e.this}",
```
georgesittas, line 390:
```suggestion
            exp.SchemaCommentProperty: lambda self, e: self.naked_property(e),
```

georgesittas (approved):

PR #2706, opened 2023-12-19
chore: new docs - AST primer

barakalon:
Thought this might be helpful for a lot of folks.

Happy to move this somewhere else, too (I'm not really sure how to add arbitrary markdown to pdocs). Also happy to keep this out of this repo, if you don't think this belongs here.

### DIFF
--- a/posts/ast_primer.md
+++ b/posts/ast_primer.md
@@ -32,7 +32,7 @@ repr(ast)
 #         (IDENTIFIER this: x, quoted: False)))))
 ```
 
-This is a textual representation of the internal data structure. Here's a breakdown of some its components:
+This is a textual representation of the internal data structure. Here's a breakdown of some of its components:
 ```
 Expression type    child key
   |               /
@@ -83,10 +83,10 @@ class Column(Expression):
 
 `Column` subclasses `Expression`.
 
-`arg_types` is a class attribute that specifies the possible children. The `args` keys of an Expression instance correspond to the `arg_types` keys of its class. The values of the `arg_types` dict are `True` if the key is required, meaning it is guaranteed to be present in `args`.
+`arg_types` is a class attribute that specifies the possible children. The `args` keys of an Expression instance correspond to the `arg_types` keys of its class. The values of the `arg_types` dict are `True` if the key is required.
 
 There are some common `arg_types` keys:
-- "this": This is typically used for the primary child. In `Column`, "this" is this the identifier for the name of the column.
+- "this": This is typically used for the primary child. In `Column`, "this" is the identifier for the column's name.
 - "expression": This is typically used for the secondary child
 - "expressions": This is typically used for a primary list of children
 
@@ -94,7 +94,7 @@ There aren't strict rules for when these keys are used, but they help with some 
 - `Expression.this`: shorthand for `self.args.get("this")`
 - `Expression.expression`: similarly, shorthand for the expression arg
 - `Expression.expressions`: similarly, shorthand for the expressions list arg
-- `Expression.name`: text  name for whatever `this` is
+- `Expression.name`: text name for whatever `this` is
 
 `arg_types` don't specify the possible Expression types of children. This can be a challenge when you are writing code to traverse a particular AST and you don't know what to expect. A common trick is to parse an example query and print out the `repr`.
 
@@ -318,7 +318,7 @@ ast = (
 ```
 
 > [!WARNING]
-> High-level builder methods will attempt to parse string arguments into Expressions. This can be very convenient, but make sure to keep in mind dialect of the string. If its written in a specific dialect, you need to set the `dialect` argument.
+> High-level builder methods will attempt to parse string arguments into Expressions. This can be very convenient, but make sure to keep in mind the dialect of the string. If its written in a specific dialect, you need to set the `dialect` argument.
 > 
 > You can avoid parsing by passing Expressions as arguments, e.g. `.where(exp.column("b") < 4)` instead of `.where("b < 4")`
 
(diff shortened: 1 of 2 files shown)


posts/ast_primer.md
```diff
@@ -0,0 +1,405 @@
...
+- `Expression.this`: shorthand for `self.args.get("this")`
+- `Expression.expression`: similarly, shorthand for the expression arg
+- `Expression.expressions`: similarly, shorthand for the expressions list arg
+- `Expression.name`: text  name for whatever `this` is
+
+`arg_types` don't specify the possible Expression types of children. This can be a challenge when you are writing code to traverse a particular AST and you don't know what to expect. A common trick is to parse an example query and print out the `repr`.
+
+You can traverse an AST using just args, but there are some higher-order functions for programmatic traversal.
+
+> [!NOTE]
```
georgesittas, line 103: TIL, awesome trick!

georgesittas (approved):
Love this, awesome work Barak! 👏

georgesittas:
> Happy to move this somewhere else, too (I'm not really sure how to add arbitrary markdown to pdocs). Also happy to keep this out of this repo, if you don't think this belongs here.

My first reaction thinking this was that it may not be very intuitive to search for a tutorial like this one under `posts`. What about moving it under `sqlglot` instead? I think that way it'd show up as the default markdown document when you opened that directory in github too?

posts/ast_primer.md
```diff
@@ -0,0 +1,405 @@
...
+#         (COLUMN this: 
+#           (IDENTIFIER this: a, quoted: False)), from: 
+#         (FROM this: 
+#           (TABLE this: 
+#             (IDENTIFIER this: x, quoted: False)))), alias: 
+#       (TABLEALIAS this: 
+#         (IDENTIFIER this: x, quoted: False)))))
+```
+
+This is a textual representation of the internal data structure. Here's a breakdown of some its components:
```
georgesittas, line 35:
```suggestion
This is a textual representation of the internal data structure. Here's a breakdown of some of its components:
```

posts/ast_primer.md
```diff
@@ -0,0 +1,405 @@
...
+    arg_types = {
+      "this": True, 
+      "table": False,
+      ...
+    }
+```
+
+`Column` subclasses `Expression`.
+
+`arg_types` is a class attribute that specifies the possible children. The `args` keys of an Expression instance correspond to the `arg_types` keys of its class. The values of the `arg_types` dict are `True` if the key is required, meaning it is guaranteed to be present in `args`.
```
georgesittas, line 86: > meaning it is guaranteed to be present in `args`.

There is a gotcha / small detail here that you can bypass that guarantee and still create a "partial" AST if you want, for example using the `IGNORE` error flag. Do you think there's value in hinting that here or in a footnote?

posts/ast_primer.md
```diff
@@ -0,0 +1,405 @@
...
+
+There are some common `arg_types` keys:
+- "this": This is typically used for the primary child. In `Column`, "this" is this the identifier for the name of the column.
+- "expression": This is typically used for the secondary child
+- "expressions": This is typically used for a primary list of children
```
georgesittas, line 91: I like how you summed it up using "primary", "secondary" etc, super clean description. I was always puzzled when I'd try to put into words what `this` represented lol

posts/ast_primer.md
```diff
@@ -0,0 +1,405 @@
...
+
+There aren't strict rules for when these keys are used, but they help with some of the convenience methods available on all Expression types:
+- `Expression.this`: shorthand for `self.args.get("this")`
+- `Expression.expression`: similarly, shorthand for the expression arg
+- `Expression.expressions`: similarly, shorthand for the expressions list arg
+- `Expression.name`: text  name for whatever `this` is
```
georgesittas, line 97:
```suggestion
- `Expression.name`: text name for whatever `this` is
```

posts/ast_primer.md
```diff
@@ -0,0 +1,405 @@
...
+    exp
+    .select("a", "b")
+    .from_("x")
+    .where("b < 4")
+    .limit(10)
+)
+```
+
+> [!WARNING]
+> High-level builder methods will attempt to parse string arguments into Expressions. This can be very convenient, but make sure to keep in mind dialect of the string. If its written in a specific dialect, you need to set the `dialect` argument.
```
georgesittas, line 321:
```suggestion
> High-level builder methods will attempt to parse string arguments into Expressions. This can be very convenient, but make sure to keep in mind the dialect of the string. If its written in a specific dialect, you need to set the `dialect` argument.
```

georgesittas (commented):
Would also make sense to reference this in the README so it's easily accessible.

tobymao:
if you put it as a doc string in mkdwon it should show up. 

pretty awesome post thanks for doing it!

posts/ast_primer.md
barakalon: ah, good to know. I'll use softer language.

posts/ast_primer.md
barakalon: haha nice

barakalon:
If its a docstring, then you can only really read it in pdocs, since Github won't render it.

I think we want to be using [portray](https://timothycrosley.github.io/portray/), not pdocs directly.

> NOTE: For most projects, the best way to use pdocs is using [portray](https://timothycrosley.github.io/portray/).

That would give us more control over non-autogenerated docs.

barakalon:
Ah, I see in sqlmesh you use both mkdocs and pdocs.

For now, I propose we keep this in "posts", add a link to it somewhere in the README, and link people to it when necessary in Slack

posts/ast_primer.md
barakalon: this is only github flavored markdown

tobymao:
can you add a link to it in this pr in the readme?

what you can do in a doc string is include the post, then it will show up in pdocs

tobymao:
look at sqlglot/__init__.py

"""
.. include:: ../README.md

----
"""

tobymao (approved):

PR #2710, opened 2023-12-20
feat(optimizer): simplify date comparisons

barakalon:
date literals are people too

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -105,6 +105,7 @@ class ClickHouse(Dialect):
             ),
             "MAP": parse_var_map,
             "MATCH": exp.RegexpLike.from_arg_list,
+            "RANDCANONICAL": exp.Rand.from_arg_list,
             "UNIQ": exp.ApproxDistinct.from_arg_list,
             "XOR": lambda args: exp.Xor(expressions=args),
         }
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -397,6 +397,7 @@ class ClickHouse(Dialect):
             exp.Pivot: no_pivot_sql,
             exp.Quantile: _quantile_sql,
             exp.RegexpLike: lambda self, e: f"match({self.format_args(e.this, e.expression)})",
+            exp.Rand: rename_func("randCanonical"),
             exp.StartsWith: rename_func("startsWith"),
             exp.StrPosition: lambda self, e: f"position({self.format_args(e.this, e.args.get('substr'), e.args.get('position'))})",
             exp.VarMap: lambda self, e: _lower_func(var_map_sql(self, e)),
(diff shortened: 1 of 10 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -276,7 +276,9 @@ def _simplify_comparison(expression, left, right, or_=False):
         rargs = {rl, rr}
 
         matching = largs & rargs
-        columns = {m for m in matching if isinstance(m, exp.Column)}
+        columns = {
+            m for m in matching if not isinstance(m, exp.Literal) and not _is_date_literal(m)
```
barakalon, line 280: I think this logic gets shaky on something like:

```sql
RANDOM() > 1 AND RANDOM() > 0
```

But even in that case, I think many engines will do common subexpression elimination and only calculate RANDOM once?

georgesittas: Not sure this is a reasonable assumption, I think each `RANDOM()` call should yield different results. E.g.
```
georgesittas=# select random(), random();
       random        |      random
---------------------+-------------------
 0.44325693039061487 | 0.671197612400882
(1 row)
```

tobymao: why don't you just do is not _is_constant() and not random?

barakalon: Ah thanks for checking!

Well, you could have `1 + RANDOM()`, so we'd need to do something like `not _is_constant(m) and not m.find(exp.Random)`. And I'm worried that there might be other nondeterministic expressions we're over looking.

So maybe I'll start with something narrower for now...

georgesittas (approved):

tobymao (approved):
make sure to transpile rand to random when necessary

barakalon:
There are a few dialects with non-standard random signatures.
e.g. 
[presto](https://prestodb.io/docs/current/functions/math.html#id0) the first parameter is the max value
Teradata seems to have RAND(min, max)
Some engines return a random integer, while most return [0, 1)

I think the only issue is dialects that accept multiple args, like Teradata, where parsing will now fail. The rest won't be correctly transpiled... but we'll be no worse off than before.

Also, [snowflake](https://docs.snowflake.com/en/sql-reference/functions-data-generation) and [clickhouse](https://clickhouse.com/docs/en/sql-reference/functions/random-functions) have smorgasbords of nondeterministic functions. Not sure if we want to account for all those.

sqlglot/dialects/clickhouse.py
```diff
@@ -397,6 +397,7 @@ class Generator(generator.Generator):
             exp.Pivot: no_pivot_sql,
             exp.Quantile: _quantile_sql,
             exp.RegexpLike: lambda self, e: f"match({self.format_args(e.this, e.expression)})",
+            exp.Rand: rename_func("randCanonical"),
```
tobymao, line 400: don't we need to parse randCanonical -> rand?

barakalon: ah, yes

PR #2714, opened 2023-12-22
refactor!: normalize units to all be upper case

tobymao:


georgesittas:
@tobymao did you test this across engines? I'm not sure if uppercase / lowercase versions are equivalent for all engines.

tobymao:
> @tobymao did you test this across engines? I'm not sure if uppercase / lowercase versions are equivalent for all engines.

no i didn’t, could you take a quick look if you have time?

georgesittas (approved):
Checked the changed tests and they seemed good.

> i checked clickhouse and it was fine. and the case is mixed all over anyways. so need this as a baseline anyways for proper transpilation

Agreed yeah, canonicalizing the case makes sense 👍

PR #2719, opened 2023-12-22
allow JSON_EXTRACT_SCALAR in BQ to pass a single argument

haritamar:
BigQuery allows `JSON_EXTRACT_SCALAR` to get only a single argument, where the expression defaults to '$', as can be seen [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract_scalar).

So I added a change that will automatically add this default.
In my change after converting back to SQL the '$' parameter will be explicitly added - I thought it makes sense but please let me know if not.


tobymao:
thanks for the PR!

PR #2723, opened 2023-12-27
feat!: Change Expression.__repr__ format

barakalon:
As [discussed on slack](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1703632977455049):

The current format can be really hard to read.
```python
from sqlglot import parse_one

ast = parse_one("SELECT a, b FROM (SELECT a, b FROM x) AS x")

print(repr(ast))
```

```
(SELECT expressions: **[1]** **[3]**
  (COLUMN this: 
    (IDENTIFIER this: a, quoted: False)),
  (COLUMN this:
    (IDENTIFIER this: b, quoted: False)), from: **[2]**
  (FROM this:
    (SUBQUERY this:
      (SELECT expressions:
        (COLUMN this:
          (IDENTIFIER this: a, quoted: False)),
        (COLUMN this:
          (IDENTIFIER this: b, quoted: False)), from:
        (FROM this: 
          (TABLE this: 
            (IDENTIFIER this: x, quoted: False)))), alias: 
      (TABLEALIAS this: 
        (IDENTIFIER this: x, quoted: False)))))
```

Some problems of this format:
[1] No delineator for lists make it unclear when an arg has a list value and unclear where the list begins and ends.
[2] Arg keys at the end of lines make it hard to tell which expression the arg applies to.
[3] All-caps expression types don't match class names exactly

Inspired by Python's [ast](https://docs.python.org/3/library/ast.html) module, I propose we switch to this:

```
Select(
  expressions=[
    Column(
    Column(
      this=Identifier(this=b, quoted=False))],
  from=From(
    this=Subquery(
      this=Select(
        expressions=[
          Column(
          Column(
            this=Identifier(this=b, quoted=False))],
        from=Table(
      alias=TableAlias(
        this=Identifier(this=a, quoted=False))))))
```
This is similarly compact and, IMO, solves problems 1-3.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -497,7 +497,7 @@ class Expression(metaclass=_Expression):
         return self.sql()
 
     def __repr__(self) -> str:
-        return self._to_s()
+        return _to_s(self)
 
     def sql(self, dialect: DialectType = None, **opts) -> str:
         """
@@ -513,30 +513,6 @@ class Expression(metaclass=_Expression):
         from sqlglot.dialects import Dialect
 
         return Dialect.get_or_raise(dialect).generate(self, **opts)
-
-    def _to_s(self, hide_missing: bool = True, level: int = 0) -> str:
-        indent = "" if not level else "\n"
-        indent += "".join(["  "] * level)
-        left = f"({self.key.upper()} "
-
-        args: t.Dict[str, t.Any] = {
-            k: ", ".join(
-                v._to_s(hide_missing=hide_missing, level=level + 1)
-                if hasattr(v, "_to_s")
-                else str(v)
-                for v in ensure_list(vs)
-                if v is not None
-            )
-            for k, vs in self.args.items()
-        }
-        args["comments"] = self.comments
-        args["type"] = self.type
-        args = {k: v for k, v in args.items() if v or not hide_missing}
-
-        right = ", ".join(f"{k}: {v}" for k, v in args.items())
-        right += ")"
-
-        return indent + left + right
 
     def transform(self, fun, *args, copy=True, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -5439,6 +5416,34 @@ def maybe_copy(instance, copy=True):
...
 
 
+def _to_s(node: t.Any, hide_missing: bool = True, level: int = 0) -> str:
+    """Generate a textual representation of an Expression tree"""
+    if isinstance(node, Expression):
+        args = {k: v for k, v in node.args.items() if v is not None or not hide_missing}
+        if node.comments or not hide_missing:
+            args["comments"] = node.comments
+        if node.type or not hide_missing:
+            args["type"] = node.type
```
tobymao, line 5426: are there any args with called type? there could be one in the future, maybe underscore these or something

sqlglot/expressions.py
```diff
@@ -5439,6 +5416,34 @@ def maybe_copy(instance, copy=True):
...
+
+        # Inline Expressions that don't have any child Expressions.
+        # This helps make expressions like Identifier more compact.
+        if any(isinstance(v, (Expression, list, tuple)) for v in args.values()):
```
tobymao, line 5430: i don't think a child can be a tuple, only a list. this method actually seems useful on base expression, it could be called is_leaf or something... wonder if this could be used elsewhere

georgesittas (approved):

sqlglot/expressions.py
barakalon: Ah, I was just copying the logic of `ensure_list`. But I can change it to just `list`.

sqlglot/expressions.py
barakalon: I added an underscore. Kinda makes it hard to read... but I think that's OK for now since type annotation is a little more advanced.

barakalon:
Added some logic for handling multiline comments/strings

sqlglot/expressions.py
georgesittas: what about renaming `type` to something else? e.g. `dtype` comes to mind, it's used by pandas / numpy as well

barakalon: Actually, `dtype` _is_ an arg, but `type` isn't haha

We can also just append the field to a list of strings, so worst case we get duplicate "type" fields in the representation.

georgesittas: Ah I see.. the underscore seems ok for now too, makes sense to have some kind of notation for non-arg attributes

PR #2725, opened 2023-12-27
Chore: Update installation steps in README to include sqlglotrs

izeigerman:


tobymao (commented):
shouldn’t this be an option?

izeigerman:
@tobymao don't we want it to be a default one though?

tobymao:
> @tobymao don't we want it to be a default one though?

ok let’s make it default but say if you want pure python to exclude it

izeigerman:
@tobymao updated, let me know

tobymao (approved):

PR #2726, opened 2023-12-27
Add option to pushdown CTE aliases earlier in qualify

dangoldin:
Based on Slack conversation with @barakalon we wanted to support the following valid Snowflake query:

```
WITH y (c) AS (
  SELECT 
    SUM(a)
  FROM (
    SELECT 1 a
  ) AS x
  HAVING c > 0
)
SELECT * FROM y
```

where the `HAVING c > 0` is referencing the CTE alias `c`


tests/test_optimizer.py
```diff
@@ -285,6 +285,21 @@ def test_qualify_columns__with_invisible(self):
         schema = MappingSchema(self.schema, {"x": {"a"}, "y": {"b"}, "z": {"b"}})
         self.check_file("qualify_columns__with_invisible", qualify_columns, schema=schema)
 
+    def test_pushdown_cte_alias_columns(self):
+        # Both of these should have SUM(a) aliased to "c"
+        queries = [
```
barakalon, line 290: Let's just create another fixture file. We can add more tests for unions, subqueries, etc.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -495,6 +495,38 @@ def quote_identifiers(expression: E, dialect: DialectType = None, identify: bool
...
+    Args:
+        expression: Expression to pushdown.
+
+    Returns:
+        The expression with the CTE aliases pushed down into the projection.
+    """
+    for cte in expression.find_all(exp.CTE):
+        if cte.alias_column_names:
+            new_expressions = []
+            for _alias, projection in zip(cte.alias_column_names, cte.this.expressions):
```
barakalon, line 519: Might need special handling for UNIONs:

```sql
WITH x (c) AS (
    SELECT
      SUM(a)
    FROM x
    GROUP BY b
    HAVING c > 0

    UNION ALL

    SELECT
      SUM(a)
    FROM y
    GROUP BY b
    HAVING c > 0
) ...
```

I'm actually curious what Snowflake does here...

Also, I think it's possible to have subqueries in a CTE. Weird stuff like:
```sql
WITH x AS (
  (
    SELECT 1 a
  )
  LIMIT 1
)
SELECT * FROM x
```

And again, I'm curious what snowflake does in these cases.

barakalon (commented):
Nice PR!

Here's some examples of using the dialect attributes to customize optimizer behavior:
https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/simplify.py#L1095
https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/qualify_columns.py#L43

sqlglot/optimizer/qualify_columns.py
dangoldin: I tried the first case (without the GROUP BY b) after your initial comment and Snowflake throws an error so I didn't code for it. I tested with the following and it worked as expected:

```
WITH y (c) AS (
  SELECT sum(a) from (SELECT SUM(a) AS a
                      FROM (SELECT 1 a) AS x
                      UNION ALL
                      SELECT sum(d) as a
                      FROM (SELECT 1 d) as y) t
                                      HAVING c > 0
)
SELECT * FROM y;
```

The second case with the subqueries is valid but also not sure how the pushdown should work in this case. I tried turning into the WITH x(c) ... and adding the HAVING clause but that's not valid in Snowflake:

```
-- INVALID
WITH x(c) AS (
  (
    SELECT 1 a
  )
  HAVING c > 0
)
SELECT * FROM x;

-- VALID and handles the pushdown
WITH x(c) AS (
    SELECT sum(1) a
  HAVING c > 0
  LIMIT 1
)
SELECT * FROM x;"""

-- INVALID
WITH x(c) AS (
    (SELECT sum(1) a)
  HAVING c > 0
  LIMIT 1
)
SELECT * FROM x;
```

tests/test_optimizer.py
dangoldin: Just added it.

dangoldin:
> Nice PR!
> 
> Here's some examples of using the dialect attributes to customize optimizer behavior: https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/simplify.py#L1095 https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/qualify_columns.py#L43

Just took a stab at it - let me know if this is what you intended here but I made it optional and default to what is set in the Dialect if not explicitly passed in.

sqlglot/optimizer/qualify_columns.py
barakalon: Got it.
Let's add some tests for some of these in `tests/fixtures/optimizer/pushdown_cte_alias_columns.sql`, even if its a noop, to make sure it doesn't throw an error. I'm worried because this code makes an implicit assumption that its dealing with an `exp.Select` instance.

sqlglot/optimizer/qualify.py
```diff
@@ -63,9 +67,18 @@ def qualify(
     expression = normalize_identifiers(expression, dialect=dialect)
     expression = qualify_tables(expression, db=db, catalog=catalog, schema=schema)
 
+    pushdown_cte_alias_columns = (
```
barakalon, line 70: Move this to just above the `if` statement its meant for.

sqlglot/dialects/dialect.py
```diff
@@ -220,6 +220,11 @@ class Dialect(metaclass=_Dialect):
     For example, such columns may be excluded from `SELECT *` queries.
     """
 
+    PUSHDOWN_CTE_ALIAS_COLUMNS = False
+    """
+    Whether or not to qualify outputs before columns.
```
barakalon, line 225: I don't think this description is precise enough.
"outputs" is vague, as is "columns".

Maybe we should call this "PREFER_CTE_ALIAS_COLUMN", and the description should describe how the CTE alias columns will override any projection aliases in the subquery.

barakalon (commented):
A couple more minor comments :+1:

dangoldin:
Addressed latest round of feedback - let me know if the docs are good. I may have over done it a bit.

sqlglot/optimizer/qualify.py
```diff
@@ -28,6 +29,7 @@ def qualify(
     validate_qualify_columns: bool = True,
     quote_identifiers: bool = True,
     identify: bool = True,
+    pushdown_cte_alias_columns: t.Optional[bool] = None,
```
tobymao, line 32: this option is unnecessary as it comes from the dialect

sqlglot/dialects/dialect.py
```diff
@@ -220,6 +220,24 @@ class Dialect(metaclass=_Dialect):
     For example, such columns may be excluded from `SELECT *` queries.
     """
 
+    PREFER_CTE_ALIAS_COLUMN = False
```
tobymao, line 223: is this untrue in all dialects and only true in snowflake?

barakalon (approved):
LGTM

I yield to Toby.

sqlglot/optimizer/qualify.py
dangoldin: Makes sense. I removed it - had it initially before I added it to the dialect.

sqlglot/dialects/dialect.py
dangoldin: I've only found it in Snowflake so far and tried Postgres, DuckDB, and BigQuery. @barakalon tried it in MySQL + SQLlite as well I think.

barakalon: Yeah, I tried MySQL, SQLite, Postgres, Trino, and Hive

PR #2734, opened 2023-12-29
feat: clickhouse: add aggregate parsing

pkit:
add ability to parse aggregate functions for CH dialect add special expressions for:
- aggregate function
- parametrised aggregate (alredy exists)
- aggregate combinator
- parametrised aggregate combinator


sqlglot/dialects/clickhouse.py
```diff
@@ -261,6 +387,14 @@ def _parse_join(
                 join.set("global", join.args.pop("method", None))
             return join
 
+        def _get_combinator_parts(self, func_name: str) -> t.List[str]:
```
tobymao, line 390: this is extremely inefficient. can you instead leverage trie.py? build the trie as a class constant and use it here to do prefix matching. you can reverse the string and feed it into the trie

sqlglot/dialects/clickhouse.py
```diff
@@ -489,6 +654,15 @@ def parameterizedagg_sql(self, expression: exp.Anonymous) -> str:
             params = self.expressions(expression, key="params", flat=True)
             return self.func(expression.name, *expression.expressions) + f"({params})"
 
+        def anonymousaggfunc_sql(self, expression: exp.Anonymous) -> str:
+            return self.func(expression.name, *expression.expressions)
+
+        def combinedaggfunc_sql(self, expression: exp.Anonymous) -> str:
+            return self.anonymousaggfunc_sql(expression)
+
+        def combinedparameterizedagg_sql(self, expression: exp.Anonymous) -> str:
```
tobymao, line 663: are the types here right? shouldn't it be like CombinedParameterizedAgg ?

sqlglot/dialects/clickhouse.py
pkit: Yeah, I can use trie, didn't know that it's there.

pkit: Maybe it's better to put all agg func names as prefixes in a trie then?
Or both.

sqlglot/dialects/clickhouse.py
pkit: It works, but yeah, expression probably needs to be `exp.CombinedParameterizedAgg`, lol

sqlglot/dialects/clickhouse.py
tobymao: sure, do whatever is cleanest / most efficient

tobymao: i guess to avoid reverse of string, you can put the inverse in the trie

sqlglot/dialects/clickhouse.py
```diff
@@ -272,14 +406,45 @@ def _parse_function(
...
                 if params:
+                    if parts:
+                        return self.expression(
+                            exp.CombinedParameterizedAgg,
+                            this=func.this,
+                            expressions=func.expressions,
+                            params=params,
+                            parts=parts,
+                        )
+                    else:
```
georgesittas, line 427: `else` can be omitted here

sqlglot/dialects/clickhouse.py
pkit: I've tried a quite a lot of approaches, and it looks like trie is not faster...
I.e. it's faster for shorter functions, like:
```python
_agg_func_trie = new_trie(AGG_FUNCTIONS)
print(timeit(lambda: in_trie(_agg_func_trie, "quantile"), number=10000))
```
```
0.003980431938543916
```
```python
timeit(lambda: _get_combinator_parts("quantile"), number=10000)
```
```
0.007004864048212767
```
But for longer ones it's exactly the same or slower.
And unfortunately one search in a trie doesn't cut it, as I still need to know where the suffix ends.
Trie that counts depth is immediately slower...

pkit: Just adding all strings to a huge dict is 10x faster. lol
```
0.0008398909121751785
```

tobymao: perfect use a dict

PR #2736, opened 2023-12-30
Fix: alter table rename should not qualify with db in postgres

z3z1ma:
image

The correct SQL to emit is as follows:

ALTER TABLE <[name](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-PARMS-NAME)> RENAME TO <[new_name](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-PARMS-NEW-NAME)>`

Note the subtle difference in terminology. Its not explicitly called out as a gotcha which would be nice in the postgres docs.


georgesittas (approved):
Thanks @z3z1ma 👍

PR #2744, opened 2024-01-01
Fix!: improve transpilation of JSON value extraction

### LINKED ISSUE #2738: : is parsed as Bracket (Snowflake)
`:` in the snowflake dialect is parsed to `Bracket`:
```python
>>> import sqlglot; sqlglot.parse_one("""select parse_json('{"fruit":"banana"}'):fruit""", read="snowflake")
Select(
  expressions=[
    Bracket(
      this=ParseJSON(
        this=Literal(this={"fruit":"banana"}, is_string=True)),
      expressions=[
        Literal(this=fruit, is_string=True)])])
```

But `:` is [GET_PATH ](https://docs.snowflake.com/en/sql-reference/functions/get_path).

NB: In duckdb this would be equivalent to `JSONExtract`.

georgesittas:
Fixes #2738

References:
- https://docs.snowflake.com/en/sql-reference/functions/get_path#examples
- https://docs.snowflake.com/en/user-guide/querying-semistructured
- https://prestodb.io/docs/current/functions/json.html
- https://learn.microsoft.com/en-us/sql/t-sql/functions/json-value-transact-sql?view=sql-server-ver16
- https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract


sqlglot/expressions.py
```diff
@@ -4118,13 +4118,6 @@ class Bracket(Condition):
     # https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#array_subscript_operator
     arg_types = {"this": True, "expressions": True, "offset": False, "safe": False}
 
-    @property
-    def output_name(self) -> str:
```
tobymao, line 4122: why'd you get rid of this?

sqlglot/parser.py
```diff
@@ -4284,11 +4284,16 @@ def _parse_bracket(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Exp
                 self.expression(exp.Slice, expression=self._parse_conjunction())
             ]
         else:
+            # Temporarily disable the colon range parser (if any) so that we will end up
```
tobymao, line 4287: don't do this

sqlglot/dialects/duckdb.py
```diff
@@ -341,6 +342,7 @@ class Generator(generator.Generator):
             exp.DiToDate: lambda self, e: f"CAST(STRPTIME(CAST({self.sql(e, 'this')} AS TEXT), {DuckDB.DATEINT_FORMAT}) AS DATE)",
             exp.Encode: lambda self, e: encode_decode_sql(self, e, "ENCODE", replace=False),
             exp.Explode: rename_func("UNNEST"),
+            exp.GetPath: lambda self, e: f"{self.sql(e, 'this')} -> {self.sql(e, 'expression')}",
```
tobymao, line 345: add this to some of the big dialects

sqlglot/expressions.py
georgesittas: it was originally [introduced](https://github.com/tobymao/sqlglot/pull/2186) so that we'd produce `field` as the alias for `SELECT table.column:field` when qualifying Snowflake queries. Thought I'd get rid of it since the representation changed but it's not necessary - can bring it back.

sqlglot/parser.py
georgesittas: Yeah it's not very good... let me see if I can refactor it

sqlglot/dialects/duckdb.py
georgesittas: Checking some examples, it may be as simple as prepending a dollar in some cases. For instance:

Snowflake:
```sql
WITH vartab AS (select parse_json('[{"attr": [{"name": "banana"}]}]') AS v)
SELECT GET_PATH(v, '[0].attr[0].name') FROM vartab;
```

BigQuery:
```sql
WITH vartab AS (select parse_json('[{"attr": [{"name": "banana"}]}]') AS v)
SELECT JSON_EXTRACT(v, '$[0].attr[0].name') FROM vartab;
```

The thing is that we need to check whether the Snowflake path begins with a `[` or not, because if it doesn't then we need to prepend `$.` instead, I think.

georgesittas: Check out https://github.com/tobymao/sqlglot/pull/2744/commits/3f0497ecd5776ed84dc19938646f0791fc24f4d2

sqlglot/dialects/dialect.py
```diff
@@ -971,3 +971,29 @@ def _delta_sql(self: Generator, expression: DATE_ADD_OR_DIFF) -> str:
         )
 
     return _delta_sql
+
+
+def prepend_dollar_to_path(expression: exp.GetPath) -> exp.GetPath:
```
tobymao, line 976: maybe let's call this path_to_jsonpath

sqlglot/expressions.py
tobymao: let's bring it back

tobymao: we still have x['y'] right

sqlglot/parser.py
tobymao: any ideas on this?

georgesittas: Taking a look now, it's a bit gnarly

sqlglot/dialects/dialect.py
georgesittas: renamed

sqlglot/expressions.py
georgesittas: Ah, makes sense. Brought it back.

sqlglot/parser.py
georgesittas: @tobymao take a look at https://github.com/tobymao/sqlglot/pull/2744/commits/3910a94299027b11f018b84d9df59e6a2c8976c6

tobymao (approved):

PR #2750, opened 2024-01-02
Refactor!: set sample clause keyword(s) as class constant to enable transpilation

### LINKED ISSUE #2747: Convert SAMPLE (snowflake) to SAMPLE USING (duckdb)
```python
>>> sqlglot.parse_one("""
SELECT * FROM example SAMPLE (3) SEED (82);
            """, read="snowflake").sql(dialect="duckdb")
'SELECT * FROM example SAMPLE (3) REPEATABLE (82)'
```
however [duckdb requires USING SAMPLE](https://duckdb.org/docs/sql/query_syntax/sample), eg:
```
SELECT * FROM example USING SAMPLE (3) REPEATABLE (82)
```

I think this is because the parse tree is different, eg:

Snowflake

```python
>>> sqlglot.parse_one("""
SELECT * FROM example SAMPLE (3) SEED (82);
            """, read="snowflake").find(exp.Select).args["from"].this
(TABLESAMPLE size: 
  (LITERAL this: 3, is_string: False), seed: 
  (LITERAL this: 82, is_string: False), kind: SAMPLE, this: 
  (TABLE this: 
    (IDENTIFIER this: example, quoted: False)))
````                
           
Duckdb
```python
>>> sqlglot.parse_one("""
SELECT * FROM example USING SAMPLE (3) REPEATABLE (82);
            """, read="duckdb").find(exp.Select).args["sample"]
(TABLESAMPLE size: 
  (LITERAL this: 3, is_string: False), seed: 
  (LITERAL this: 82, is_string: False), kind: USING SAMPLE)
```              

Should these be the same parse trees? If not no worries, I'll add a transform instead.

georgesittas:
Fixes #2747

- DuckDB uses `USING SAMPLE` to sample the whole resulting relation in a query, and `TABLESAMPLE` for sampling just a simple intermediate result. See https://duckdb.org/docs/sql/samples.html
- Snowflake treats `SAMPLE` and `TABLESAMPLE` as synonyms, see https://docs.snowflake.com/en/sql-reference/constructs/sample

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -484,6 +484,7 @@ class ClickHouse(Dialect):
         STRUCT_DELIMITER = ("(", ")")
         NVL2_SUPPORTED = False
         TABLESAMPLE_REQUIRES_PARENS = False
+        SAMPLE_CLAUSE = "SAMPLE"
 
         STRING_TYPE_MAPPING = {
             exp.DataType.Type.CHAR: "String",
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -307,6 +307,7 @@ class DuckDB(Dialect):
         RENAME_TABLE_WITH_DB = False
         NVL2_SUPPORTED = False
         SEMI_ANTI_JOIN_WITH_SIDE = False
+        SAMPLE_CLAUSE = "USING SAMPLE"
 
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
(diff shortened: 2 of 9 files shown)


tekumara:
Thanks for this!

When transpiling from snowflake -> duckdb with this fix I now get:
```python
sqlglot.parse_one("""
SELECT * FROM example SAMPLE (3) SEED (82);
            """, read="snowflake").sql(dialect="duckdb")
'SELECT * FROM example TABLESAMPLE (3) REPEATABLE (82)'
```

BTW - should the transpiler also translate the `3` in snowflake to `3%` in duckdb to accomodate the different semantics (ie: [snowflake](https://docs.snowflake.com/en/sql-reference/constructs/sample#:~:text=which%20each%20row%20has%20a%2010%25%20probability) vs [duckdb](https://duckdb.org/docs/sql/samples.html#:~:text=select%20a%20sample%20of%205%20rows)) or should I do that myself with a transform?

georgesittas:
Hey @tekumara, is the only issue now that Snowflake has "percentage" semantics in `SAMPLE (3)` whereas DucKDB has "count" semantics in `TABLESAMPLE (3)`?

Can you make a new issue about this? I'll take a look tomorrow.

PR #2752, opened 2024-01-03
Fix(bigquery): enable transpilation of single-argument TIME func

### LINKED ISSUE #2749: Support TIME (BigQuery) to TIME (duckdb)
Thx for this great library! The following feature is missing:
BigQuery TIME is not proper translated to duckdb TIME 
`sqlglot.transpile("select TIME('2008-12-25 15:30:00')", read="bigquery", write="duckdb")`

returns: `SELECT TIME('2008-12-25 15:30:00')`

Expected: `SELECT TIME '2008-12-25 15:30:00'`

georgesittas:
Fixes #2749

Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/time_functions#time

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -215,6 +215,12 @@ def _unix_to_time_sql(self: BigQuery.Generator, expression: exp.UnixToTime) -> s
     return ""
 
 
+def _parse_time(args: t.List) -> exp.Func:
+    if len(args) == 1:
+        return exp.TsOrDsToTime(this=args[0])
+    return exp.Anonymous(this="TIME", expressions=args)
+
+
 class BigQuery(Dialect):
     WEEK_OFFSET = -1
     UNNEST_COLUMN_ONLY = True
@@ -358,6 +364,7 @@ class BigQuery(Dialect):
                 this=seq_get(args, 0),
                 expression=seq_get(args, 1) or exp.Literal.string(","),
             ),
+            "TIME": _parse_time,
             "TIME_ADD": parse_date_delta_with_interval(exp.TimeAdd),
             "TIME_SUB": parse_date_delta_with_interval(exp.TimeSub),
             "TIMESTAMP_ADD": parse_date_delta_with_interval(exp.TimestampAdd),
@@ -594,6 +601,7 @@ class BigQuery(Dialect):
             exp.TsOrDsAdd: _ts_or_ds_add_sql,
             exp.TsOrDsDiff: _ts_or_ds_diff_sql,
             exp.TsOrDsToDate: ts_or_ds_to_date_sql("bigquery"),
+            exp.TsOrDsToTime: rename_func("TIME"),
             exp.Unhex: rename_func("FROM_HEX"),
             exp.UnixToTime: _unix_to_time_sql,
             exp.Values: _derived_table_values_to_unnest,
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -26,7 +26,6 @@ from sqlglot.dialects.dialect import (
     rename_func,
     timestrtotime_sql,
     ts_or_ds_add_cast,
-    ts_or_ds_to_date_sql,
 )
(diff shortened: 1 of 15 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -215,6 +215,12 @@ def _unix_to_time_sql(self: BigQuery.Generator, expression: exp.UnixToTime) -> s
     return ""
 
 
+def _parse_time(args: t.List) -> exp.Func:
+    if len(args) == 1:
+        return exp.TsOrDsToTime(this=args[0])
+    return exp.Anonymous(this="TIME", expressions=args)
```
georgesittas, line 221: Decided to use this as a fallback so that the scope of the PR is smaller and the patch is more conservative. It's not straightforward to transpile all variants so I thought we can start with the single argument one and iterate from there if ever needed..

tobymao, line 221: why is this anonymous? why don't we have a first class expression for this?

tests/dialects/test_bigquery.py
```diff
@@ -190,6 +193,13 @@ def test_bigquery(self):
         self.validate_all('x <> """"""', write={"bigquery": "x <> ''"})
         self.validate_all("x <> ''''''", write={"bigquery": "x <> ''"})
         self.validate_all("CAST(x AS DATETIME)", read={"": "x::timestamp"})
+        self.validate_all(
+            "SELECT TIME('2008-12-25 15:30:00')",
+            write={
+                "bigquery": "SELECT TIME('2008-12-25 15:30:00')",
+                "duckdb": "SELECT CAST('2008-12-25 15:30:00' AS TIME)",
+            },
+        )
```
georgesittas, line 202: Probably gonna test a few more dialects first before this is fully ready

sqlglot/generator.py
```diff
@@ -3112,6 +3112,9 @@ def toarray_sql(self, expression: exp.ToArray) -> str:
         cond_for_null = arg.is_(exp.null())
         return self.sql(exp.func("IF", cond_for_null, exp.null(), exp.Array(expressions=[arg])))
 
+    def tsordstotime_sql(self, expression: exp.TsOrDsToTime) -> str:
```
tobymao, line 3115: what's this? this kills round trip

sqlglot/dialects/bigquery.py
georgesittas: Decided to use this as a fallback so that the scope of the PR is smaller and the patch is more conservative. It's not straightforward to transpile all variants so I thought we can start with the single argument one and iterate from there if ever needed..

tobymao: ok

georgesittas: For example we can do the 3 argument variant in a followup PR, since it'll enable us to transpile it to T-SQL's [TIMEFROMPARTS](https://learn.microsoft.com/en-us/sql/t-sql/functions/timefromparts-transact-sql?view=sql-server-ver16#syntax) call, but I'd probably create a different expression for that like `TimeFromParts` since we already have `DateFromParts`

tobymao: ok

sqlglot/generator.py
georgesittas: how? this is done to avoid overriding everywhere with a cast needlessly, if you need an identity roundtrip then you override the transform entry like I did for BQ right?

tobymao: that’s what we do for tsordstodate, if you’re going to do that you should refactor that as well

georgesittas: I see, let me take another look and I'll refactor if needed then

tobymao: tsordstodate could be cleaned up if we had this just need to make sure we don’t have extraneous casts

georgesittas: take a look here @tobymao https://github.com/tobymao/sqlglot/pull/2752/commits/bcf2b912fdff69a1b86366c2190ab231ff07e8eb

PR #2753, opened 2024-01-03
Feat: guess the correct dialect in case we get an unknown one

georgesittas:
I've typed `duckbd` instead of `duckdb` so many times and it's finally occurred to me that the UX can be improved a bit :D

Reference: https://docs.python.org/3/library/difflib.html#difflib.get_close_matches

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -307,7 +307,7 @@ class Dialect(metaclass=_Dialect):
             if not result:
                 from difflib import get_close_matches
 
-                similar = seq_get(get_close_matches(dialect_name, cls.classes, n=1), 0)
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -305,7 +305,13 @@ def get_or_raise(cls, dialect: DialectType) -> Dialect:
 
             result = cls.get(dialect_name.strip())
             if not result:
-                raise ValueError(f"Unknown dialect '{dialect_name}'.")
+                from difflib import get_close_matches
+
+                similar = seq_get(get_close_matches(dialect_name, cls.classes, n=1), 0) or ""
+                if similar:
+                    similar = f" Did you mean {similar}?"
```
georgesittas, line 312: A detail here is that we won't ever suggest the SQLGlot dialect `""` but that felt ok

tobymao (approved):

PR #2761, opened 2024-01-03
Fix!: improve table sampling transpilation

### LINKED ISSUE #2757: SAMPLE defaults when transpiling from Snowflake <-> duckdb
As [mentioned here](https://github.com/tobymao/sqlglot/pull/2750#issuecomment-1874857353) the semantics of default SAMPLE parts differ between Snowflake and duckdb, eg:
1. size without a modifier means [percent in snowflake](https://docs.snowflake.com/en/sql-reference/constructs/sample#:~:text=which%20each%20row%20has%20a%2010%25%20probability) and [rows in duckdb](https://duckdb.org/docs/sql/samples.html#:~:text=select%20a%20sample%20of%205%20rows)
1. without a sampling method [Snowflake defaults to BERNOULLI](  https://docs.snowflake.com/en/sql-reference/constructs/sample#:~:text=If%20no%20method%20is%20specified%2C%20the%20default%20is%20BERNOULLI) whereas [duckdb defaults](https://duckdb.org/docs/sql/samples.html#bernoulli:~:text=By%20default%2C%20DuckDB%20uses%20reservoir%20sampling%20when%20an%20exact%20number%20of%20rows%20is%20sampled%2C%20and%20system%20sampling%20when%20a%20percentage%20is%20specified.%20The%20sampling%20methods%20are%20described%20in%20detail%20below.)  to uses reservoir sampling for rows and system sampling when a percentage is specified

I'm not sure if the appropriate transform should be made when transpiling.

georgesittas:
Fixes #2757

Need to take a look at the sampling methods to make this complete for #2757. @tekumara if it turns out more involved than expected we may do this at a later time.

References:
- https://docs.snowflake.com/en/sql-reference/constructs/sample
- https://duckdb.org/docs/sql/samples.html#examples
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6
- https://clickhouse.com/docs/en/sql-reference/statements/select/sample
- https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-sampling.html
- https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sampling.html
- https://sqlrambling.net/2018/01/24/tablesample-basic-examples/


tobymao (approved):

georgesittas:
Will try to add a few more tests and wrap this up

sqlglot/dialects/postgres.py
```diff
@@ -465,7 +466,6 @@ class Generator(generator.Generator):
             exp.TimestampTrunc: timestamptrunc_sql,
             exp.TimeStrToTime: timestrtotime_sql,
             exp.TimeToStr: lambda self, e: f"TO_CHAR({self.sql(e, 'this')}, {self.format_time(e)})",
-            exp.TableSample: no_tablesample_sql,
```
georgesittas, line 468: Postgres does support `TABLESAMPLE` for materialized tables / views:

```
georgesittas=# select * from t tablesample system (90);
 c
---
 1
(1 row)
georgesittas=# select * from t tablesample system (90) repeatable (55);
 c
---
 1
(1 row)
```

PR #2763, opened 2024-01-03
Feat: add STR_TO_TIME parsing and generation for teradata and sqlite

indrajeetadityaroy9:
The PR introduces the functionality of converting strings to date-time or timestamp objects with custom formats in Teradata and SQLite. 

https://dbmstutorials.com/random_teradata/teradata-string-to-timestamp.html
https://www.sqlitetutorial.net/sqlite-date-functions/sqlite-strftime-function/

### DIFF
--- a/sqlglot/dialects/sqlite.py
+++ b/sqlglot/dialects/sqlite.py
@@ -27,6 +27,16 @@ def _date_add_sql(self: SQLite.Generator, expression: exp.DateAdd) -> str:
     modifier = f"'{modifier} {unit.name}'" if unit else f"'{modifier}'"
     return self.func("DATE", expression.this, modifier)
 
+def str_to_date_sql(self: SQLite.Generator, expression: exp.StrToDate) -> str:
+    date_string = self.sql(expression, "this")
+    format_specifier = self.format_time(expression)
+    return f"STRFTIME({format_specifier}, {date_string})"
+
+
+def str_to_time_sql(self: SQLite.Generator, expression: exp.StrToTime) -> str:
+    time_string = self.sql(expression, "this")
+    format_specifier = self.format_time(expression)
+    return f"STRFTIME({format_specifier}, {time_string})"
 
 def _transform_create(expression: exp.Expression) -> exp.Expression:
     """Move primary key to a column and enforce auto_increment on primary keys."""
@@ -138,6 +148,8 @@ class SQLite(Dialect):
             exp.TableSample: no_tablesample_sql,
             exp.TimeStrToTime: lambda self, e: self.sql(e, "this"),
             exp.TryCast: no_trycast_sql,
+            exp.StrToDate: str_to_date_sql,
+            exp.StrToTime: str_to_time_sql
         }
 
         PROPERTIES_LOCATION = {
--- a/sqlglot/dialects/sqlite.py
+++ b/sqlglot/dialects/sqlite.py
@@ -27,6 +27,7 @@ def _date_add_sql(self: SQLite.Generator, expression: exp.DateAdd) -> str:
     modifier = f"'{modifier} {unit.name}'" if unit else f"'{modifier}'"
     return self.func("DATE", expression.this, modifier)
 
+
 def str_to_date_sql(self: SQLite.Generator, expression: exp.StrToDate) -> str:
     date_string = self.sql(expression, "this")
     format_specifier = self.format_time(expression)
@@ -37,6 +38,7 @@ def str_to_time_sql(self: SQLite.Generator, expression: exp.StrToTime) -> str:
     time_string = self.sql(expression, "this")
     format_specifier = self.format_time(expression)
     return f"STRFTIME({format_specifier}, {time_string})"
+
 
 def _transform_create(expression: exp.Expression) -> exp.Expression:
     """Move primary key to a column and enforce auto_increment on primary keys."""
@@ -149,7 +151,7 @@ class SQLite(Dialect):
             exp.TimeStrToTime: lambda self, e: self.sql(e, "this"),
             exp.TryCast: no_trycast_sql,
             exp.StrToDate: str_to_date_sql,
-            exp.StrToTime: str_to_time_sql
+            exp.StrToTime: str_to_time_sql,
         }
 
         PROPERTIES_LOCATION = {
--- a/sqlglot/dialects/sqlite.py
+++ b/sqlglot/dialects/sqlite.py
@@ -28,16 +28,10 @@ def _date_add_sql(self: SQLite.Generator, expression: exp.DateAdd) -> str:
     return self.func("DATE", expression.this, modifier)
 
 
-def str_to_date_sql(self: SQLite.Generator, expression: exp.StrToDate) -> str:
-    date_string = self.sql(expression, "this")
+def str_to_datetime_sql(self: SQLite.Generator, expression) -> str:
+    datetime_string = self.sql(expression, "this")
     format_specifier = self.format_time(expression)
-    return f"STRFTIME({format_specifier}, {date_string})"
-
-
-def str_to_time_sql(self: SQLite.Generator, expression: exp.StrToTime) -> str:
-    time_string = self.sql(expression, "this")
-    format_specifier = self.format_time(expression)
-    return f"STRFTIME({format_specifier}, {time_string})"
+    return f"STRFTIME({format_specifier}, {datetime_string})"
 
 
 def _transform_create(expression: exp.Expression) -> exp.Expression:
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/teradata.py
```diff
@@ -7,6 +7,18 @@
 from sqlglot.tokens import TokenType
 
 
+def str_to_date_sql(self: Teradata.Generator, expression: exp.StrToDate) -> str:
+    date_string = self.sql(expression, "this")
+    format_specifier = self.format_time(expression)
+    return f"TO_DATE({date_string}, {format_specifier})"
```
georgesittas, line 13: Looks like this is not used anywhere, let's get rid of it

tests/dialects/test_sqlite.py
```diff
@@ -140,6 +140,19 @@ def test_sqlite(self):
             read={"snowflake": "LEAST(x, y, z)"},
             write={"snowflake": "LEAST(x, y, z)"},
         )
+        self.validate_all(
+            "SELECT STR_TO_DATE('2022-01-01', '%Y-%m-%d')",
```
georgesittas, line 144: Take a look at how `validate_all` is [defined](https://github.com/tobymao/sqlglot/blob/3bd811d67e7001046812f2a8590bd07e23b81c88/tests/dialects/test_dialect.py#L27C14-L27C14): the first argument gets parsed according to the dialect corresponding to the surrounding class (here: SQLite). This means that it's generally SQL which is valid in that dialect, i.e. `STRFTIME` instead of `STR_TO_DATE`.

sqlglot/dialects/sqlite.py
```diff
@@ -27,6 +27,16 @@ def _date_add_sql(self: SQLite.Generator, expression: exp.DateAdd) -> str:
...
+def str_to_date_sql(self: SQLite.Generator, expression: exp.StrToDate) -> str:
+    date_string = self.sql(expression, "this")
+    format_specifier = self.format_time(expression)
+    return f"STRFTIME({format_specifier}, {date_string})"
+
+
+def str_to_time_sql(self: SQLite.Generator, expression: exp.StrToTime) -> str:
+    time_string = self.sql(expression, "this")
+    format_specifier = self.format_time(expression)
+    return f"STRFTIME({format_specifier}, {time_string})"
```
georgesittas, line 40: These look identical, can we consolidate them into a single function and then reuse it for both `StrToDate` and `StrToTime` to avoid code duplication?

tests/dialects/test_sqlite.py
```diff
@@ -140,6 +140,19 @@ def test_sqlite(self):
...
+                        "mysql": "SELECT STR_TO_DATE('2022-01-01', '%Y-%m-%d')",
+                        "sqlite": "SELECT STRFTIME('%Y-%m-%d', '2022-01-01')"
+                    }
+        )
+        self.validate_all(
+            "SELECT CAST('2022-01-01' AS DATE FORMAT '%Y-%m-%d')",
+            write = {
+                "sqlite": "SELECT STRFTIME('%Y-%m-%d', '2022-01-01')"
+            }
+        )
```
georgesittas: This test is incorrect, perhaps because of a misunderstanding surrounding `validate_all`. The way you've written this is like expecting the two queries to produce the same results because you're both reading and writing SQLite, but if you run them against the engine you'll notice they produce different results.

tests/dialects/test_teradata.py
```diff
@@ -205,3 +205,41 @@ def test_cast(self):
...
+            write={
+                "teradata": "TO_TIMESTAMP('2024-01-01 13:14:15', 'YYYY-MM-DD HH24:MI:SS')",
+                "oracle": "TO_TIMESTAMP('2024-01-01 13:14:15', 'YYYY-MM-DD HH24:MI:SS')",
+                "postgres": "TO_TIMESTAMP('2024-01-01 13:14:15', 'YYYY-MM-DD HH24:MI:SS')",
+            },
+        )
+        self.validate_all(
+            "CAST('2024-01-01' AS DATE FORMAT 'YYYY-MM-DD')",
+            write={"teradata": "CAST('2024-01-01' AS DATE FORMAT 'YYYY-MM-DD')"},
+        )
```
georgesittas, line 229: Nit: we generally try to use `validate_identity(input[, expected])` for tests like this where we're only interested to check the parsing / generation roundtrip for the given dialect, because it's more compact. So this would look like:

```python
        self.validate_identity("CAST('2024-01-01' AS DATE FORMAT 'YYYY-MM-DD')")
```

tests/dialects/test_sqlite.py
```diff
@@ -140,6 +140,19 @@ def test_sqlite(self):
             read={"snowflake": "LEAST(x, y, z)"},
             write={"snowflake": "LEAST(x, y, z)"},
         )
+        self.validate_all(
+            "SELECT STR_TO_DATE('2022-01-01', '%Y-%m-%d')",
+            write = {
+                        "mysql": "SELECT STR_TO_DATE('2022-01-01', '%Y-%m-%d')",
+                        "sqlite": "SELECT STRFTIME('%Y-%m-%d', '2022-01-01')"
+                    }
+        )
```
georgesittas: Run `make style` for commits that are part of a PR so that the CI doesn't crash due to styling / linter issues

sqlglot/dialects/teradata.py
```diff
@@ -202,6 +208,7 @@ class Generator(generator.Generator):
             exp.StrToDate: lambda self, e: f"CAST({self.sql(e, 'this')} AS DATE FORMAT {self.format_time(e)})",
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
             exp.Use: lambda self, e: f"DATABASE {self.sql(e, 'this')}",
+            exp.StrToTime: lambda self, e: str_to_time_sql(self, e),
```
georgesittas, line 211:
```suggestion
            exp.StrToTime: str_to_time_sql,
```

tests/dialects/test_sqlite.py
```diff
@@ -140,6 +140,16 @@ def test_sqlite(self):
...
+        self.validate_all(
+            "SELECT STR_TO_DATE('2022-01-01', '%Y-%m-%d')",
+            write={
+                "mysql": "SELECT STR_TO_DATE('2022-01-01', '%Y-%m-%d')",
+                "sqlite": "SELECT STRFTIME('%Y-%m-%d', '2022-01-01')",
+            },
+        )
+        self.validate_identity("STRFTIME('%Y-%m-%d %H:%M:%f', '2024-01-01 13:14:15.123')")
+        self.validate_identity("STRFTIME('%Y-%m-%d %I:%M:%S %p', '2024-01-01 01:14:15 PM')")
+        self.validate_identity("STRFTIME('%d-%b-%Y', '01-Jan-2024')")
```
georgesittas, line 152: Let's try to increase the coverage here by testing how `STRFTIME` is transpiled to some other important dialects like postgres, mysql, duckdb

georgesittas: Also, I'm curious about the last two tests here because SQLite seems to return `NULL` for them or something?

```sql
sqlite> SELECT STRFTIME('%Y-%m-%d %I:%M:%S %p', '2024-01-01 01:14:15 PM');

sqlite> SELECT STRFTIME('%d-%b-%Y', '01-Jan-2024');

```

Are these formats correct? Why are we not getting back a proper value?

tests/dialects/test_teradata.py
```diff
@@ -205,3 +205,28 @@ def test_cast(self):
...
+                "postgres": "TO_TIMESTAMP('2024-01-01 13:14:15', 'YYYY-MM-DD HH24:MI:SS')",
+            },
+        )
+        self.validate_identity("CAST('2024-01-01' AS DATE FORMAT 'YYYY-MM-DD')")
+        self.validate_identity("CAST('2024/01/01' AS DATE FORMAT 'YYYY/MM/DD')")
+        self.validate_identity(
+            "TO_TIMESTAMP('2024-01-01 13:14:15.123', 'YYYY-MM-DDBHH:MI:SS.S(3)')"
+        )
+        self.validate_identity("TO_TIMESTAMP('2024-01-01 01:14:15 PM', 'YYYY-MM-DDbHH:MI:SSBT')")
+        self.validate_identity("CAST('01-Jan-2024' AS DATE FORMAT 'DD-MMM-YYYY')")
```
georgesittas, line 232: Let's also try to increase the coverage here - the identity `CAST(... AS DATE FORMAT ...)` tests don't seem to be testing any new functionality here, we already have a few tests for this here: https://github.com/tobymao/sqlglot/blob/main/tests/dialects/test_teradata.py#L183

georgesittas (commented):
Left a few more comments.

PR #2765, opened 2024-01-03
Fix(bigquery): fix parsing of COUNTIF

giovannipcarvalho:


tobymao:
thanks, can you add a test for bigquery -> bigquery?

tests/dialects/test_bigquery.py
```diff
@@ -251,6 +251,7 @@ def test_bigquery(self):
                 "duckdb": "SELECT STRFTIME(CAST('2023-12-25' AS DATE), '%Y%m%d')",
             },
         )
+        self.validate_all("SELECT COUNTIF(x)", write={"duckdb": "SELECT COUNT_IF(x)"})
```
tobymao, line 254: add a bigquery test here "bigquery": ... and make sure that passes, i assume it won't because you haven't added a rename_func for bigquery

sqlglot/dialects/bigquery.py
```diff
@@ -328,6 +328,7 @@ class Parser(parser.Parser):
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
+            "COUNTIF": exp.CountIf.from_arg_list,
```
georgesittas, line 331: Can you instead do this so that other dialects that support the `COUNTIF` variant can benefit?
```python
class CountIf(AggFunc):
    _sql_names = ["COUNT_IF", "COUNTIF"]
```

tests/dialects/test_bigquery.py
georgesittas, line 254: Let's also test the BQ -> BQ and DuckDB -> BQ cases

georgesittas: Ah @tobymao beat me to it, lol

giovannipcarvalho: just did.  the test still passes

sqlglot/dialects/bigquery.py
giovannipcarvalho: you mean change the original definition of `CountIf` in expressions.py?

georgesittas: That's right, change it to the above and get rid of the `FUNCTIONS` entry.

tests/dialects/test_bigquery.py
giovannipcarvalho: oh, I hadn't thought about the duckdb -> bigquery case.

it actually fails, I am not sure why though. I think it now thinks duckdb should not have an underscore.

```
AssertionError: 'SELECT COUNT_IF(x)' != 'SELECT COUNTIF(x)'
- SELECT COUNT_IF(x)
?             -
+ SELECT COUNTIF(x)
```

tobymao: thanks for the quick iteration, you're missing a rename func in the generator

giovannipcarvalho: oh, the rename_func I believe

georgesittas (approved):

PR #2771, opened 2024-01-05
Feat(optimizer): add support for the UNPIVOT operator

georgesittas:
Still need to test a few more engines but this should be a good start (will iterate on this PR).

References:
- https://docs.snowflake.com/en/sql-reference/constructs/unpivot
- https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unpivot_operator
- https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-unpivot.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_FROM_clause-pivot-unpivot-examples.html#r_FROM_clause-unpivot-examples
- https://learn.microsoft.com/en-us/sql/t-sql/queries/from-using-pivot-and-unpivot?view=sql-server-ver16

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4118,6 +4118,15 @@ class Alias(Expression):
         return self.alias
 
 
+# BigQuery requires the UNPIVOT column list aliases to be either strings or ints, but
+# Spark requires identifiers. This enables us to transpile between the two easily.
+#
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unpivot_operator
+# https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-unpivot.html
+class PivotAlias(Alias):
+    pass
+
+
 class Aliases(Expression):
     arg_types = {"this": True, "expressions": True}
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3532,8 +3532,8 @@ class Pivot(Expression):
     }
 
     @property
-    def is_unpivot(self) -> t.Optional[bool]:
-        return self.args.get("unpivot")
+    def unpivot(self) -> bool:
+        return bool(self.args.get("unpivot"))
 
 
 class Window(Condition):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4119,10 +4119,7 @@ class Alias(Expression):
 
 
 # BigQuery requires the UNPIVOT column list aliases to be either strings or ints, but
-# Spark requires identifiers. This enables us to transpile between the two easily.
-#
-# https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unpivot_operator
-# https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-unpivot.html
+# other dialects require identifiers. This enables us to transpile between them easily.
 class PivotAlias(Alias):
     pass
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3531,6 +3531,10 @@ class Pivot(Expression):
         "include_nulls": False,
     }
 
+    @property
+    def is_unpivot(self) -> t.Optional[bool]:
+        return self.args.get("unpivot")
+
 
 class Window(Condition):
     arg_types = {
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -382,7 +382,6 @@ def _expand_stars(
     replace_columns: t.Dict[int, t.Dict[str, str]] = {}
     coalesced_columns = set()
 
-    pivot_columns = None
     pivot_output_columns = None
     pivot_exclude_columns = None
 
@@ -397,7 +396,7 @@ def _expand_stars(
                     c.output_name for e in field.expressions for c in e.find_all(exp.Column)
                 }
         else:
-            pivot_columns = set(c.output_name for c in pivot.find_all(exp.Column))
+            pivot_exclude_columns = set(c.output_name for c in pivot.find_all(exp.Column))
 
             pivot_output_columns = [c.output_name for c in pivot.args.get("columns", [])]
             if not pivot_output_columns:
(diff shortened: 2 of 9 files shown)


sqlglot/expressions.py
```diff
@@ -3522,6 +3522,10 @@ class Pivot(Expression):
         "include_nulls": False,
     }
 
+    @property
+    def is_unpivot(self) -> t.Optional[bool]:
```
tobymao, line 3526: just call it unpivot and make it a bool not optional

sqlglot/optimizer/qualify_columns.py
```diff
@@ -73,21 +76,41 @@ def qualify_columns(
...
+
             if scope.external_columns and not scope.is_correlated_subquery and not scope.pivots:
                 column = scope.external_columns[0]
-                raise OptimizeError(
-                    f"""Column '{column}' could not be resolved{f" for table: '{column.table}'" if column.table else ''}"""
-                )
+                for_table = f" for table: '{column.table}'" if column.table else ""
+                raise OptimizeError(f"Column '{column}' could not be resolved{for_table}")
+
+            if unqualified_columns and scope.pivots and scope.pivots[0].is_unpivot:
```
tobymao, line 89: what about like any(pivot.unpivot for pivot in scope.pivots)

tobymao: or a for loop on the pivots because you use it on 93

sqlglot/optimizer/qualify_columns.py
```diff
@@ -358,18 +382,26 @@ def _expand_stars(
...
-    # TODO: handle optimization of multiple PIVOTs (and possibly UNPIVOTs) in the future
     pivot_columns = None
     pivot_output_columns = None
-    pivot = t.cast(t.Optional[exp.Pivot], seq_get(scope.pivots, 0))
+    pivot_exclude_columns = None
 
-    has_pivoted_source = pivot and not pivot.args.get("unpivot")
-    if pivot and has_pivoted_source:
-        pivot_columns = set(col.output_name for col in pivot.find_all(exp.Column))
+    pivot = t.cast(t.Optional[exp.Pivot], seq_get(scope.pivots, 0))
```
tobymao, line 389: can you have multiple pivots? why not just do a for loop?

georgesittas: You can but even if we do the for loop it won't work correctly; the semantics are tricky to get right because e.g. each pivot feeds the columns it produces into the next, one so it becomes messy when there are more than one...

I'd just leave it for now and look into this when there's more demand, chained pivots should be rare anyways.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -358,18 +382,26 @@ def _expand_stars(
...
+        if pivot.is_unpivot:
+            pivot_output_columns = [c.output_name for c in _unpivot_columns(pivot)]
+
+            field = pivot.args.get("field")
+            if isinstance(field, exp.In):
+                pivot_exclude_columns = {
+                    c.output_name for e in field.expressions for c in e.find_all(exp.Column)
+                }
+        else:
+            pivot_columns = set(c.output_name for c in pivot.find_all(exp.Column))
```
tobymao, line 400: why is this a set and 393 a list?

sqlglot/optimizer/qualify_columns.py
georgesittas: Let's do the 1 to many transition progressively, don't think this will provide much value here because we still won't handle the many (un)pivot case ungracefully. At least this way I think it will work correctly for one and if there are >1 we'll raise due to ambiguous columns

tobymao: ok

tobymao (approved):

sqlglot/optimizer/qualify_columns.py
georgesittas: this is just speeding up lookups, the lists are used to preserve the order

tobymao: so you do need both variants? they can't both be sets?

georgesittas: yep, there's no performance gain in using a set for `pivot_output_columns` whereas iirc using a list has the nice side-effect of _generally_ preserving the order of the columns as returned by the engines when there's a star

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -644,6 +643,61 @@ FROM (
...
+SELECT
+  `_q_0`.`product` AS `product`,
+  `_q_0`.`quarter` AS `quarter`,
+  `_q_0`.`sales` AS `sales`
+FROM `produce` AS `produce` UNPIVOT(`sales` FOR `quarter` IN (`produce`.`q1`, `produce`.`q2`, `produce`.`q3`, `produce`.`q4`)) AS `_q_0`;
+
+# title: unpivoted table source with multiple value columns
+# execute: false
+# dialect: bigquery
+SELECT * FROM produce AS produce(product, q1, q2, q3, q4) UNPIVOT((first_half_sales, second_half_sales) FOR semesters IN ((Q1, Q2) AS 'semester_1', (Q3, Q4) AS 'semester_2'));
```
georgesittas, line 693: FYI Spark also supports this "multi-column unpivot" operator, but uses actual identifiers for the aliases. From their doc [page](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-unpivot.html):

```sql
-- multiple value columns can be unpivoted per row
SELECT * FROM sales_quarterly
    UNPIVOT EXCLUDE NULLS (
        (first_quarter, second_quarter)
        FOR half_of_the_year IN (
            (q1, q2) AS H1,
            (q3, q4) AS H2
        )
    );
```

If we want to enable the BQ -> Spark transpilation, i.e. convert strings/numbers (former) to identifiers (latter), I see two options:

1. Implement a transformer that converts all aliases to identifiers and apply it to `Pivot` for Spark
2. Define a new expression called `PivotAlias` so we'll just control _its_ generation logic per dialect

I'm personally leaning towards the 2nd one because it's simpler and will enable the opposite transpilation as well, since we'll be able to convert identifier aliases to strings when going from Spark to BQ. If there are no objections I'll go ahead and implement this.

tobymao: ok

georgesittas: Fixed in https://github.com/tobymao/sqlglot/pull/2771/commits/cb6c5f9cd0ae14805704dea4f58ddaa2be33afd2

I found that this is not related to the "multi-column unpivot" form I mentioned after all because aliases can appear in "single-column unpivot" forms too. This means that this fix enables the transpilation of BQ `UNPIVOT`s with aliases to and from other dialects too, such as duckdb and redshift.

georgesittas:
Think this is good to go, feel free to take another look

tobymao:
looks good

georgesittas:
Cool thanks for checking - will do another pass later and merge shortly.

PR #2776, opened 2024-01-05
Fix(optimizer): replace star with outer column list

georgesittas:


sqlglot/optimizer/scope.py
```diff
@@ -79,6 +79,10 @@ def __init__(
         self.cte_scopes = []
         self.union_scopes = []
         self.udtf_scopes = []
+
+        if scope_type == ScopeType.UNION and parent and not outer_column_list:
```
barakalon, line 83: Any way to do this during scope traversal? This conditional, plus crawling back up the tree, feels hacky

georgesittas: I see what you mean, let me see if I can refactor it

georgesittas: Check out https://github.com/tobymao/sqlglot/pull/2776/commits/5d5b08bc95f73291eeb7ab4966bf6f27d8226d3b

barakalon (approved):
Nice 👍

PR #2778, opened 2024-01-06
Feat(snowflake): add support for SHOW COLUMNS 

andrew-sha:
Reference: https://docs.snowflake.com/en/sql-reference/sql/show-columns

### DIFF
--- a/tests/dialects/test_snowflake.py
+++ b/tests/dialects/test_snowflake.py
@@ -1420,10 +1420,9 @@ MATCH_RECOGNIZE (
 
     def test_show(self):
         # Parsed as Command
-        self.validate_identity("SHOW COLUMNS IN TABLE dt_test")
         self.validate_identity("SHOW TABLES LIKE 'line%' IN tpch.public")
 
-        ast = parse_one("SHOW TABLES HISTORY IN tpch.public")
+        ast = parse_one("SHOW TABLES HISTORY IN tpch.public", read="snowflake")
         self.assertIsInstance(ast, exp.Command)
 
         # Parsed as Show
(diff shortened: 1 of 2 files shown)


tests/dialects/test_snowflake.py
```diff
@@ -1443,11 +1442,29 @@ def test_show(self):
...
+        self.validate_identity("SHOW COLUMNS LIKE '_foo%' IN TABLE dt_test")
+        self.validate_identity("SHOW COLUMNS IN VIEW")
+        self.validate_identity("SHOW COLUMNS LIKE '_foo%' IN VIEW dt_test")
+
+        ast = parse_one("SHOW COLUMNS LIKE '_testing%' IN dt_test", read="snowflake")
+        self.assertIsInstance(ast, exp.Show)
+        table = ast.find(exp.Table)
+        literal = ast.find(exp.Literal)
+
+        self.assertIsNotNone(table)
```
tobymao, line 1462: this is not necessary,you already test that this is true with 1463, same with 1465, so let's remove the assertNone tests

tests/dialects/test_snowflake.py
```diff
@@ -1443,11 +1442,29 @@ def test_show(self):
...
         self.assertEqual(table.sql(dialect="snowflake"), '"TEST"."PUBLIC"."customers"')
 
+        self.validate_identity("SHOW COLUMNS")
+        self.validate_identity("SHOW COLUMNS IN TABLE dt_test")
+        self.validate_identity("SHOW COLUMNS LIKE '_foo%' IN TABLE dt_test")
+        self.validate_identity("SHOW COLUMNS IN VIEW")
+        self.validate_identity("SHOW COLUMNS LIKE '_foo%' IN VIEW dt_test")
+
+        ast = parse_one("SHOW COLUMNS LIKE '_testing%' IN dt_test", read="snowflake")
+        self.assertIsInstance(ast, exp.Show)
```
tobymao, line 1458: assert instance tests are also not really necessary because if they were commands, the ast.finds would fail, so we can get rid of those

tobymao:
@andrew-sha fantastic PR, great work! small comments

PR #2783, opened 2024-01-07
Feat: extend submodule annotate_types to handle STRUCT

fool1280:
Extend submodule annotate_types to handle STRUCT


tobymao:
can you fix the build and write some unit tests?

sqlglot/optimizer/annotate_types.py
```diff
@@ -509,6 +515,23 @@ def _annotate_by_args(
...
+        if struct:
+            self._set_type(
+                expression,
+                exp.DataType(
+                    this=exp.DataType.Type.STRUCT,
+                    expressions=[
+                        exp.ColumnDef(
+                            this=exp.to_identifier(expr.alias),
+                            kind=expr.type,
+                            nested=expr.args.get("nested"),
```
georgesittas, line 527: This doesn't look right, `ColumnDef` has no arg named `nested`. I think we just need to get rid of this.

tests/test_optimizer.py
```diff
@@ -548,6 +548,39 @@ def test_scope_warning(self, logger):
             level="warning",
         )
 
+    def test_struct_type_annotation(self):
+        expression = annotate_types(parse_one("select struct(1 as col)"))
+        self.assertEqual(expression.expressions[0].type.this, exp.DataType.Type.STRUCT)
+        self.assertEqual(expression.expressions[0].expressions[0].type.this, exp.DataType.Type.INT)
```
georgesittas, line 554: Not sure about this test, shouldn't we be checking whether the type is `STRUCT<INT>` here? Which would translate to matching the AST structure you've created with the `ColumnDef`s etc..

georgesittas: Ditto about the other tests under this method

georgesittas (commented):
A few final comments @fool1280, looks good otherwise - good job 👍.

Nit: I'd clean up the test a bit, instead of repeating the `expression.expressions[0]` you can store that in a variable or something prior to calling `assertEqual`.

sqlglot/optimizer/annotate_types.py
fool1280: done

tests/test_optimizer.py
fool1280: done :D

fool1280:
I just refactored the test, and also delete the `nested` arg in `ColumnDef`, thank you!

georgesittas (approved):
Awesome, LGTM!

tobymao (approved):

georgesittas:
This may be a little more nuanced than originally expected, because the struct's `expressions` may not always be `Alias`es. For example, here you have `PropertyEQ`:

```python
>>> import sqlglot
>>> sqlglot.parse_one("struct_pack(a := 1)", "duckdb")
Struct(
  expressions=[
    PropertyEQ(
      this=Column(
      expression=Literal(this=1, is_string=False))])
```

You also may have nothing, e.g. this is valid Spark: `SELECT STRUCT(1, 'foo', NULL)`. I think it makes sense to start with this PR though though and we can iterate later.

PR #2795, opened 2024-01-08
Fix(clickhouse): allow transpilation of countIf, fix 2 arg variant parsing

### LINKED ISSUE #2792: two-argument version of clickhouse's `countIf` function incorrectly fails to parse
This used to work:

```
In [3]: import sqlglot as sg

In [4]: sg.__version__
Out[4]: '20.7.1'

In [5]: sg.parse_one("select countIf(x, y)", read="clickhouse")
...
ParseError: The number of provided arguments (2) is greater than the maximum number of supported arguments (1). Line 1, Col: 20.
  select countIf(x, y)
```

ClickHouse CLI:

```
localhost :) select countIf(1, false), countIf(1, true);

SELECT
    countIf(1, false),
    countIf(1, true)

Query id: cd5dea26-fae2-4bdc-a18b-edbf5b910fda

┌─countIf(1, false)─┬─countIf(1, true)─┐
│                 0 │                1 │
└───────────────────┴──────────────────┘
```

Unfortunately the clickhouse docs on `countIf`'s signature are nonexistent, with only a single mention of the function in the `If` combinators section here https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators#-if

georgesittas:
Fixes #2792

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -23,14 +23,23 @@ def _lower_func(sql: str) -> str:
     return sql[:index].lower() + sql[index:]
 
 
-def _quantile_sql(self, e):
+def _quantile_sql(self: ClickHouse.Generator, e: exp.Quantile) -> str:
     quantile = e.args["quantile"]
     args = f"({self.sql(e, 'this')})"
+
     if isinstance(quantile, exp.Array):
         func = self.func("quantiles", *quantile)
     else:
         func = self.func("quantile", quantile)
+
     return func + args
+
+
+def _parse_count_if(args: t.List) -> exp.CountIf | exp.CombinedAggFunc:
+    if len(args) == 1:
+        return exp.CountIf(this=seq_get(args, 0))
+
+    return exp.CombinedAggFunc(this="countIf", expressions=args, parts=("count", "If"))
 
 
 class ClickHouse(Dialect):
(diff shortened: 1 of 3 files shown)


georgesittas:
Could also just `pop` the `COUNTIF` entry from `FUNCTIONS` here and let all `countIf` variants be parsed as a `CombinedAggFunc`. The reason I overrode the entry was to enable the transpilation of the single-arg variant towards other dialects.

tobymao:
can we add a test that checks every single one of these combinator functions to make sure they accept multiple args?

georgesittas:
Sure, how about a for loop over [`AGG_FUNC_MAPPING`](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/clickhouse.py#L240) that ensures we roundtrip correctly if we call the function with, say, 4 arguments (`func(x, y, z, w)`)?

tobymao:
sure, would it have caught this one?

georgesittas:
> sure, would it have caught this one?

No, because the way Clickhouse's parser works is that it calls the superclass' `_parse_function` and then only constructs the Clickhouse-specific combined / parameterized expressions off of it if the result is an `Anonymous`. In this case `COUNTIF` was being parsed as `exp.CountIf` (which was a side-effect of https://github.com/tobymao/sqlglot/commit/2a94f2ba0d93ee1454a2d19cc8577e211fd5cbe0), so it would not hit that logic.

Tbh thinking about this again I'm not sure if it's gonna be worth the overhead of testing this because most functions in the mapping I mentioned are very Clickhouse-specific and they should already be parsed correctly (🤞) - `countIf` seems like an exception.

tobymao:
ok looks good then

PR #2800, opened 2024-01-09
Fix(snowflake): apply range parser after colon, if any

### LINKED ISSUE #2798: ParseError when using LIKE/ILIKE on an element in an object in Snowflake
I'm getting `ParseError: Invalid expression / Unexpected token` when using `LIKE` or `ILIKE` on an element within an object in Snowflake.

Example:
```
import sqlglot
sqlglot.parse(""" select parse_json('{"x": "hello"}'):x like 'hello' """, read="snowflake")
sqlglot.parse(""" select data:x like 'hello' from some_table """, read="snowflake")
```

Both of these cause the parsing error, but both are valid Snowflake statements.

georgesittas:
Fixes #2798

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -328,6 +328,10 @@ def _parse_colon_get_path(
         if not self._match(TokenType.COLON):
             break
 
+    if self._match_set(self.RANGE_PARSERS):
+        expression = self.RANGE_PARSERS[self._prev.token_type](self, this)
+        this = expression if isinstance(expression, exp.Expression) else this
+
     return this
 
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -328,6 +328,10 @@ def _parse_colon_get_path(
         if not self._match(TokenType.COLON):
             break
 
+    if self._match_set(self.RANGE_PARSERS):
+        expression = self.RANGE_PARSERS[self._prev.token_type](self, this)
+        this = expression if isinstance(expression, exp.Expression) else this
```
tobymao, line 333: this = expression or this ?

tobymao (approved):

PR #2801, opened 2024-01-10
Feat: add support for Heredocs in Databricks Python UDFs

viethungle-vt1401:
This PR adds support for Heredocs in the `AS dollar_quoted_string` clause for Python user-defined functions in [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-sql-function.html#create-python-functions).


tests/dialects/test_databricks.py
```diff
@@ -28,6 +28,13 @@ def test_databricks(self):
             "SELECT * FROM sales UNPIVOT EXCLUDE NULLS (sales FOR quarter IN (q1 AS `Jan-Mar`))"
         )
 
+        self.validate_all(
+            "CREATE FUNCTION greet(name STRING) RETURNS STRING LANGUAGE PYTHON AS $$def greet(name):\n\t return f'Hello {name}!'$$",
+            write={
+                "snowflake": "CREATE FUNCTION greet(name TEXT) RETURNS TEXT LANGUAGE PYTHON AS 'def greet(name):\n\t return f\\'Hello {name}!\\''"
+            },
+        )
+
```
georgesittas, line 37: This test is no good, we're not testing the roundtrip from Databricks to Databricks. Have you double-checked that this is valid Snowflake?

sqlglot/parser.py
```diff
@@ -5475,6 +5476,37 @@ def _parse_comprehension(
...
+        # Note: we've already matched a $ token before reaching here
+        self._match(TokenType.PARAMETER)
+        tag_start = self._prev
+
+        while not self._match(TokenType.PARAMETER):
+            self._advance()
+            if not self._curr:
+                self.raise_error("No closing $ found.")
+
+        tag_text = self._find_sql(tag_start, self._prev)
```
georgesittas, line 5489: Why is there a while loop here? I think this section can be simplified, don't think we need `_find_sql` here, you can just get the token text directly from the tag's token, if there's a tag.

sqlglot/parser.py
```diff
@@ -1362,9 +1362,10 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
             # exp.Properties.Location.POST_SCHEMA ("schema" here is the UDF's type signature)
             extend_props(self._parse_properties())
 
-            self._match(TokenType.ALIAS)
-
-            if self._match(TokenType.COMMAND):
+            if self._match(TokenType.ALIAS) and self._match(TokenType.PARAMETER, advance=False):
```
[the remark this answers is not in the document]
georgesittas: Also, it's hard to understand why we're matching a `PARAMETER` token when reading this section. Is this syntax only valid for Databricks? Maybe we should be matching a dollar text instead using `self._match_text_seq("$")` or something?

sqlglot/parser.py
```diff
@@ -5475,6 +5476,37 @@ def _parse_comprehension(
...
+
+        tag_start = None
+        while True:
+            if self._match(TokenType.PARAMETER, advance=False):
+                if tag_start and self._find_sql(tag_start, self._curr) == tag_text:
+                    this = self._find_sql(heredoc_start, heredoc_end)
+                    self._advance()
+
+                    return self.expression(exp.RawString, this=this)
+                else:
```
georgesittas, line 5501: This `else` here is unnecessary

sqlglot/parser.py
```diff
@@ -5475,6 +5476,37 @@ def _parse_comprehension(
...
+        tag_text = self._find_sql(tag_start, self._prev)
+        heredoc_start = self._curr
+        heredoc_end = self._prev
+
+        tag_start = None
+        while True:
+            if self._match(TokenType.PARAMETER, advance=False):
+                if tag_start and self._find_sql(tag_start, self._curr) == tag_text:
+                    this = self._find_sql(heredoc_start, heredoc_end)
+                    self._advance()
```
georgesittas, line 5498: Can this be simplified? I'd expect this section to just contain some lookahead checks. For example: "if there's a $ followed by a token that matches the tag text (if any), _and_ there's another dollar, we're at the heredoc's end and only then we should capture the raw text with `_find_sql`".

georgesittas (commented):
Left some comments to consider. I did a quick first pass so let me know if I missed something in your logic.

sqlglot/parser.py
```diff
@@ -1362,9 +1362,12 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
             # exp.Properties.Location.POST_SCHEMA ("schema" here is the UDF's type signature)
             extend_props(self._parse_properties())
 
-            self._match(TokenType.ALIAS)
+            alias_ = self._match(TokenType.ALIAS)
```
tobymao, line 1365: you can just put this inline in the if statement since you don't use alias_ anywhere

sqlglot/parser.py
```diff
@@ -5475,6 +5478,31 @@ def _parse_comprehension(
             condition=condition,
         )
 
+    def _parse_heredoc(self) -> exp.RawString:
+        tag_texts = ["$", "$"]
```
tobymao, line 5482: what about something like

tags = []

if self._match_text_seq("$"):
  tags.append("$")
else: 
   return None

tags.append(self._advance_any().text)

if tags[-1] != "$":
   if self._match_text("$"):
      tags.append("$")
   else:
      self.raise_error(...)

sqlglot/parser.py
```diff
@@ -5475,6 +5478,31 @@ def _parse_comprehension(
...
+
+        if not self._match_text_seq("$"):
+            tag_texts.insert(1, self._curr.text)
+            self._advance()
+
+        self._match_text_seq("$")
+
+        heredoc_start = self._curr
+
+        while True:
```
tobymao, line 5494: while self._curr

sqlglot/parser.py
```diff
@@ -5550,10 +5578,17 @@ def _match_texts(self, texts, advance=True):
             return True
         return False
 
-    def _match_text_seq(self, *texts, advance=True):
+    def _match_text_seq(self, *texts, advance=True, upper_case=True):
```
tobymao, line 5581: i don't think we need this upper_case flag, it's a bit edge casey, should be fine to just not have this check

tests/dialects/test_databricks.py
tobymao: you should also test the here docs version

sqlglot/parser.py
georgesittas: I actually suggested we don't put that in the if haha.. the reasoning was to make this a bit more clear, i.e. that we'll always consume the alias regardless of the taken branch

georgesittas: But it's fine

sqlglot/generator.py
```diff
@@ -77,6 +77,7 @@ class Generator:
         exp.ExecuteAsProperty: lambda self, e: self.naked_property(e),
         exp.ExternalProperty: lambda self, e: "EXTERNAL",
         exp.HeapProperty: lambda self, e: "HEAP",
+        exp.Heredoc: lambda self, e: f"${e.args['tag'] or ''}$\n{e.this}\n${e.args['tag'] or ''}$",
```
tobymao, line 80: make this e.args("tag") and move this to a first class function

def heredoc_sql ...

sqlglot/parser.py
```diff
@@ -5475,6 +5476,53 @@ def _parse_comprehension(
...
+                tag_text = self._curr.text
+                tags.append(tag_text)
+                self._advance()
+            else:
+                tag_text = None
+                tags.append("$")
+        else:
+            self.raise_error("No closing $ found")
+
+        if tag_text:
```
tobymao, line 5498: tag_text can be undefined here because the raise_error on line 5496 doesn't necessarily break

sqlglot/parser.py
```diff
@@ -5475,6 +5476,53 @@ def _parse_comprehension(
...
+            if self._is_connected() and self._match_text_seq("$"):
+                tags.append("$")
+            else:
+                self.raise_error("No closing $ found")
+
+        heredoc_start = self._curr
+
+        while self._curr:
+            index = self._index
```
tobymao, line 5507: shouldn't this index be outside of the for loop? perhaps it should be at line 5480?

sqlglot/parser.py
```diff
@@ -5475,6 +5476,53 @@ def _parse_comprehension(
...
+            heredoc_end = self._prev
+            tags_matched = True
+
+            for tag in tags:
```
tobymao, line 5511: why don't you do match_text_seq here?

viethungle-vt1401: `match_text_seq` would convert everything to upper case, meanwhile we want the beginning and closing tags to be the same, case sensitive (e.g. `$foo$` and `$FOO$` shouldn't match)

sqlglot/parser.py
viethungle-vt1401: since I'm not using `_match_text_seq`, the `index` keeps track of `self._curr` is before we try to match subsequent tokens. If the matching fails at any point, we return to `index` before calling `self._advance()` to advance `self._curr` (`index` here serves a similar purpose to `index` in `_match_text_seq`)

sqlglot/parser.py
tobymao: as i said, i don't think this edge case matters much, so let's just use match_text_seq

sqlglot/generator.py
```diff
@@ -77,6 +77,7 @@ class Generator:
         exp.ExecuteAsProperty: lambda self, e: self.naked_property(e),
         exp.ExternalProperty: lambda self, e: "EXTERNAL",
         exp.HeapProperty: lambda self, e: "HEAP",
+        exp.Heredoc: lambda self, e: self.heredoc_sql(e),
```
tobymao, line 80: delete this, it's unnecessary

sqlglot/parser.py
```diff
@@ -5475,6 +5484,44 @@ def _parse_comprehension(
...
+        else:
+            return None
+
+        if self._curr and self._is_connected():
+            if not self._match_text_seq("$"):
+                tag_text = self._curr.text
+                tags.append(tag_text.upper())
+                self._advance()
+            else:
+                tags.append("$")
```
georgesittas, line 5502: Perhaps you can replace this whole if-else block with `tags.append(self._advance() and self._prev.text.upper())`?

sqlglot/parser.py
```diff
@@ -5475,6 +5484,44 @@ def _parse_comprehension(
...
+    def _parse_heredoc(self) -> t.Optional[exp.Heredoc]:
+        tags = []
+        tag_text = None
+
+        if self._match_text_seq("$"):
+            tags.append("$")
+        else:
+            return None
+
+        if self._curr and self._is_connected():
```
georgesittas, line 5496: I think `self._curr` isn't needed here? Just `_is_connected()` seems to be enough.

sqlglot/parser.py
```diff
@@ -5475,6 +5484,44 @@ def _parse_comprehension(
...
+            if not self._match_text_seq("$"):
+                tag_text = self._curr.text
+                tags.append(tag_text.upper())
+                self._advance()
+            else:
+                tags.append("$")
+        else:
+            self.raise_error("No closing $ found")
+
+        if tag_text:
```
georgesittas, line 5506: And so instead of checking `tag_text` here, the proxy would be if the last tag appended is _not_ a single dollar?

sqlglot/parser.py
```diff
@@ -5475,6 +5484,44 @@ def _parse_comprehension(
...
+            if self._is_connected() and self._match_text_seq("$"):
+                tags.append("$")
```
georgesittas, line 5508: Do we need the `_is_connected` check here?

sqlglot/parser.py
georgesittas: @tobymao wdyt?

sqlglot/parser.py
```diff
@@ -1362,9 +1362,18 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
-
-            if self._match(TokenType.COMMAND):
+            heredoc_matched = False
+            if self._match(TokenType.ALIAS):
+                if self._match_text_seq("$", advance=False):
+                    expression = self._parse_heredoc()
+                    heredoc_matched = True
+                elif self._match(TokenType.HEREDOC_STRING):
+                    expression = self.expression(exp.Heredoc, this=self._prev.text)
+                    heredoc_matched = True
```
georgesittas, line 1372: should we consolidate these two branches by handling it all in `_parse_heredoc` and setting `heredoc_matched` to true if we get back a non-`None` result? just trying to see if we can simplify it a bit

sqlglot/parser.py
tobymao: sure

sqlglot/parser.py
viethungle-vt1401: This is to catch invalid tags like $foo $

sqlglot/parser.py
georgesittas: ah may need to actually `self._advance()` before appending `self._prev.text.upper()` @viethungle-vt1401, sorry for this (it seems like it's related to the mypy issue)

viethungle-vt1401: I tried specifying `return None`for `_advance()`, but that still didn't satisfy the linter, so going with your approach!

sqlglot/parser.py
georgesittas: I wouldn't bother covering these edge cases, doesn't seem worth it. SQLGlot in general doesn't function as a SQL validator so we tend to avoid introducing complexity when possible, especially for these obscure cases.

sqlglot/parser.py
```diff
@@ -1362,9 +1362,13 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
             # exp.Properties.Location.POST_SCHEMA ("schema" here is the UDF's type signature)
             extend_props(self._parse_properties())
 
-            self._match(TokenType.ALIAS)
+            expression = None
+            if self._match(TokenType.ALIAS):
+                expression = self._parse_heredoc()
```
georgesittas, line 1367:
```suggestion
            expression = self._match(TokenType.ALIAS) and self._parse_heredoc()
```

sqlglot/parser.py
georgesittas: Whatever actually, let's leave it like it is now, should be ok

georgesittas (commented):
Ok one minor final comment here.

@viethungle-vt1401 let's test the `raise_error` code path, as well as the

```python
            if expression:
```

one. If we can't yet generate these props correctly (according to the Slack discussion), I'd say either fix this here or just drop this change in the parser altogether.

pkit:
Why not just?
```diff
diff --git a/sqlglot/dialects/databricks.py b/sqlglot/dialects/databricks.py
index 8e55b6a9..4258b563 100644
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -81,3 +81,5 @@ class Databricks(Spark):
 
     class Tokenizer(Spark.Tokenizer):
         HEX_STRINGS = []
+        HEREDOC_STRINGS = ["$"]
+        VAR_SINGLE_TOKENS = {"$"}
```
```python
print(repr(sqlglot.parse_one("""
  CREATE FUNCTION add_one(x INT) RETURNS INT LANGUAGE PYTHON AS $FOO$def add_one(x):\n  return x+1$FOO$
""", read="databricks")))

Create(
  this=UserDefinedFunction(
    this=Identifier(this=add_one, quoted=False),
    expressions=[
      ColumnDef(
        kind=DataType(this=Type.INT, nested=False, prefix=False))],
    wrapped=True),
  kind=FUNCTION,
  expression=RawString(this=def add_one(x):
        return x+1),
  exists=False,
  properties=Properties(
    expressions=[
      ReturnsProperty(
        this=DataType(this=Type.INT, nested=False, prefix=False)),
      LanguageProperty(
        this=Identifier(this=PYTHON, quoted=False))]),
  end=False)

```

pkit:
Ah, I get it, it's because round-trip is impossible and tag is lost here: https://github.com/tobymao/sqlglot/blob/70b280f24df0f4a0d4cc8262d72ce46412f76be3/sqlglotrs/src/tokenizer.rs#L403C21-L403C24
Lol
Probably need to return the tag too...

viethungle-vt1401:
> Ok one minor final comment here.
> 
> @viethungle-vt1401 let's test the `raise_error` code path, as well as the
> 
> ```python
>             if expression:
>                 extend_props(self._parse_properties())
> ```
> 
> one. If we can't yet generate these props correctly (according to the Slack discussion), I'd say either fix this here or just drop this change in the parser altogether.

If that's the case, I would drop the change in the parser and leave it for future updates. What's your verdict on this?

georgesittas:
> > Ok one minor final comment here.
> > @viethungle-vt1401 let's test the `raise_error` code path, as well as the
> > ```python
> >             if expression:
> >                 extend_props(self._parse_properties())
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > one. If we can't yet generate these props correctly (according to the Slack discussion), I'd say either fix this here or just drop this change in the parser altogether.
> 
> If that's the case, I would drop the change in the parser and leave it for future updates. What's your verdict on this?

I'd just drop the logic in the parser for now and care about it in the future.

georgesittas (approved):
Looks good, nice work 👍

PR #2810, opened 2024-01-10
fix(oracle): translate variance_pop to var_pop and variance to var_samp

gforsyth:
Fixes a bug in the Oracle dialect.

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -51,12 +51,9 @@ def to_char(args: t.List) -> exp.TimeToStr | exp.ToChar:
 
 
 def _variance_sql(self: Oracle.Generator, expression: exp.Variance) -> str:
-    this = self.sql(expression, "this")
-    return_zero = self.sql(expression, "return_zero")
-
-    func = "VARIANCE" if return_zero else "VAR_SAMP"
-
-    return f"{func}({this})"
+    if expression.args.get("return_zero"):
+        return rename_func("VARIANCE")(self, expression)
+    return rename_func("VAR_SAMP")(self, expression)
 
 
 class Oracle(Dialect):
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -50,6 +50,15 @@ def to_char(args: t.List) -> exp.TimeToStr | exp.ToChar:
     return exp.ToChar.from_arg_list(args)
 
 
+def _variance_sql(self: Oracle.Generator, expression: exp.Variance) -> str:
+    this = self.sql(expression, "this")
+    return_zero = self.sql(expression, "return_zero")
+
+    func = "VARIANCE" if return_zero else "VAR_SAMP"
+
+    return f"{func}({this})"
+
+
 class Oracle(Dialect):
     ALIAS_POST_TABLESAMPLE = True
     LOCKING_READS_SUPPORTED = True
@@ -196,6 +205,8 @@ class Oracle(Dialect):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
             exp.Trim: trim_sql,
             exp.UnixToTime: lambda self, e: f"TO_DATE('1970-01-01','YYYY-MM-DD') + ({self.sql(e, 'this')} / 86400)",
+            exp.VariancePop: rename_func("VAR_POP"),
+            exp.Variance: _variance_sql,
         }
 
         PROPERTIES_LOCATION = {
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5373,8 +5373,13 @@ class Upper(Func):
     _sql_names = ["UPPER", "UCASE"]
 
 
+# https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/VARIANCE.html
+# oracle variance returns 0 if called on 1-element input set
+# var_samp returns NULL if called 1-element input set
 class Variance(AggFunc):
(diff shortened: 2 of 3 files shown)


tests/dialects/test_oracle.py
```diff
@@ -313,3 +313,11 @@ def test_connect_by(self):
...
         for query in (f"{body}{start}{connect}", f"{body}{connect}{start}"):
             self.validate_identity(query, pretty, pretty=True)
+
+    def test_variance(self):
+        self.validate_all("VAR_SAMP(x)", write={"oracle": "VAR_SAMP(x)"})
+        self.validate_all("VAR_POP(x)", write={"oracle": "VAR_POP(x)"})
+        self.validate_all("VARIANCE(x)", write={"oracle": "VAR_SAMP(x)"})
+        self.validate_all(
+            "VAR_POP(x)", read={"": "VARIANCE_POP(x)"}, write={"oracle": "VAR_POP(x)"}
+        )
```
georgesittas, line 323: Let's use `self.validate_identity` for the first three tests since we're only doing a roundtrip, or even better keep the `validate_all` and augment the test to check how this is transpiled from and to Oracle (e.g. check Snowflake, DuckDB, ...).

Also, a nit / small note: I know that this isn't very consistent in the codebase, but I try to format these `validate_all` tests as:

```python
self.validate_all(
    ...,
    read={
        ...,
    },
    write={
        ...,
    },
)
```

This reads a bit better imo.

sqlglot/dialects/oracle.py
```diff
@@ -196,6 +196,8 @@ class Generator(generator.Generator):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
             exp.Trim: trim_sql,
             exp.UnixToTime: lambda self, e: f"TO_DATE('1970-01-01','YYYY-MM-DD') + ({self.sql(e, 'this')} / 86400)",
+            exp.VariancePop: rename_func("VAR_POP"),
+            exp.Variance: rename_func("VAR_SAMP"),
```
georgesittas, line 200: Even though `VARIANCE` is usually an alias to `VAR_SAMP`, it seems that they're not completely interchangeable in [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/VAR_SAMP.html#GUID-314D5831-0E26-4ABF-9F46-35F78F97DA52):

> This function is similar to VARIANCE, except that given an input set of one element, VARIANCE returns 0 and VAR_SAMP returns null.

Perhaps we should capture this nuance with an optional arg in `Variance`?

gforsyth: Woof.  Ok, so an optional arg to allow a user to explicitly request translating to `VARIANCE` instead of `VAR_SAMP`?

sqlglot/expressions.py
```diff
@@ -5373,9 +5373,14 @@ class Upper(Func):
...
 
 
+# https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/VARIANCE.html
+# oracle variance returns 0 if called on 1-element input set
+# var_samp returns NULL if called 1-element input set
 class Variance(AggFunc):
     _sql_names = ["VARIANCE", "VARIANCE_SAMP", "VAR_SAMP"]
 
+    arg_types = {"this": True, "return_zero": False}
+
```
gforsyth, line 5383: Is the pattern here to define a `variance` rule in all the other dialects to ignore this argument?

tests/dialects/test_oracle.py
```diff
@@ -313,3 +313,13 @@ def test_connect_by(self):
 
         for query in (f"{body}{start}{connect}", f"{body}{connect}{start}"):
             self.validate_identity(query, pretty, pretty=True)
+
+    def test_variance(self):
+        self.validate_identity("VAR_SAMP(x)", "VAR_SAMP(x)")
+        self.validate_identity("VAR_POP(x)", "VAR_POP(x)")
```
georgesittas, line 319:
```suggestion
        self.validate_identity("VAR_SAMP(x)")
        self.validate_identity("VAR_POP(x)")
```

sqlglot/dialects/oracle.py
```diff
@@ -50,6 +50,15 @@ def to_char(args: t.List) -> exp.TimeToStr | exp.ToChar:
     return exp.ToChar.from_arg_list(args)
 
 
+def _variance_sql(self: Oracle.Generator, expression: exp.Variance) -> str:
+    this = self.sql(expression, "this")
+    return_zero = self.sql(expression, "return_zero")
+
+    func = "VARIANCE" if return_zero else "VAR_SAMP"
+
+    return f"{func}({this})"
```
georgesittas, line 59:
```suggestion
    if expression.args.get("return_zero"):
        return rename_func("VARIANCE")(self, expression)
    return rename_func("VAR_SAMP")(self, expression)
```

georgesittas (commented):
@gforsyth we need to also add an entry in `FUNCTIONS` so that when `VARIANCE` is parsed in Oracle it sets the `Variance` expression's `return_zero` to true. That way you'll preserve the roundtrip, i.e. `VARIANCE` -> `VARIANCE` in Oracle.

sqlglot/expressions.py
georgesittas: You would need to match Oracle's semantics so that the transpilation is correct when you're going from Oracle -> X. So, produce something like `IFNULL(VAR_SAMP(x), 0)`. I think we need to override `variance_sql` in the base class and control its behavior with class constants to do this properly.

gforsyth:
Thanks for the pointers @georgesittas -- I'm going to close this out until I have more time to work on it.

PR #2812, opened 2024-01-11
Handle struct with no alias

fool1280:
Handle struct with no alias by mixing `ColumnDef` and `DataType`


georgesittas:
I think we may need to fix how we parse the `struct<int>` type, right now it seems like we're parsing the `int` into an identifier:

```python
>>> import sqlglot
>>> sqlglot.parse_one("cast(x as struct<int>)")
Cast(
  this=Column(
  to=DataType(
    this=Type.STRUCT,
    expressions=[
      Identifier(this=int, quoted=False)],
    nested=True,
    prefix=False))
```

sqlglot/optimizer/annotate_types.py
```diff
@@ -516,21 +516,20 @@ def _annotate_by_args(
             )
 
         if struct:
+            expressions = [
+                expr.type
+                if expr.args.get("alias") is None
```
georgesittas, line 521:
```suggestion
                if not expr.alias
```

georgesittas (approved):
Looks good, let's add a few tests for this, e.g. annotate the following:

```sql
-- I guess it should produce STRUCT<col: INT, VARCHAR>?
SELECT STRUCT(1 as col, 'foo')
```

georgesittas:
Please also rebase to grab https://github.com/tobymao/sqlglot/commit/d678182765eaca1c258b9d408d12a245ebf86832.

georgesittas (approved):
LGTM, nice work 👍

georgesittas:
@fool1280 https://github.com/tobymao/sqlglot/commit/9b4b8d8783d350b9f922f3cf64f519be6f742050

PR #2820, opened 2024-01-12
Feat: transpile SELECT .. INTO to dialects that do not support it

giorgosnikolaou:
TODO: Integrate test cases to the existing framework

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -258,6 +258,9 @@ class Generator:
     # INSERT OVERWRITE TABLE x override
     INSERT_OVERWRITE = " OVERWRITE TABLE"
 
+    # Dialect supports "SELECT INTO"
+    SUPPORT_SELECT_INTO = False
+
     TYPE_MAPPING = {
         exp.DataType.Type.NCHAR: "CHAR",
         exp.DataType.Type.NVARCHAR: "VARCHAR",
@@ -2019,6 +2022,10 @@ class Generator:
         return [locks, self.sql(expression, "sample")]
 
     def select_sql(self, expression: exp.Select) -> str:
+        into = expression.args.get("into")
+        if self.SUPPORT_SELECT_INTO == False and into:
+            expression.args["into"].pop()
+
         hint = self.sql(expression, "hint")
         distinct = self.sql(expression, "distinct")
         distinct = f" {distinct}" if distinct else ""
@@ -2063,7 +2070,19 @@ class Generator:
             self.sql(expression, "into", comment=False),
             self.sql(expression, "from", comment=False),
         )
-        return self.prepend_ctes(expression, sql)
+
+        sql = self.prepend_ctes(expression, sql)
+
+        if self.SUPPORT_SELECT_INTO == False and into:
+            # Ensures that SELECT INTO TEMP UNLOGGED -> CREATE TEMPORARY TABLE
+            kind = (
+                " TEMPORARY"
+                if into.args.get("temporary")
+                else (" UNLOGGED" if into.args.get("unlogged") else "")
+            )
+            sql = f'CREATE{kind} TABLE {into.args["this"]} AS {sql}'
+
+        return sql
 
     def schema_sql(self, expression: exp.Schema) -> str:
         this = self.sql(expression, "this")
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -258,8 +258,11 @@ class Generator:
     # INSERT OVERWRITE TABLE x override
     INSERT_OVERWRITE = " OVERWRITE TABLE"
 
-    # Dialect supports "SELECT INTO"
-    SUPPORT_SELECT_INTO = False
+    # Whether or not the SELECT .. INTO syntax is used instead of CTAS
+    SUPPORTS_SELECT_INTO = False
+
+    # Whether UNLOGGED tables are allowed
+    SUPPORTS_UNLOGGED_TABLES = False
 
     TYPE_MAPPING = {
         exp.DataType.Type.NCHAR: "CHAR",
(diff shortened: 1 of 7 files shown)


sqlglot/generator.py
```diff
@@ -258,6 +258,9 @@ class Generator:
     # INSERT OVERWRITE TABLE x override
     INSERT_OVERWRITE = " OVERWRITE TABLE"
 
+    # Dialect supports "SELECT INTO"
+    SUPPORT_SELECT_INTO = False
```
georgesittas, line 262: Let's rename to `SUPPORTS_SELECT_INTO` which is more consistent with our naming schemes.

I'd also edit the comment a bit to have it say something like "Whether or not the SELECT .. INTO syntax is used instead of CTAS"

sqlglot/generator.py
```diff
@@ -2019,6 +2022,10 @@ def after_limit_modifiers(self, expression: exp.Expression) -> t.List[str]:
         return [locks, self.sql(expression, "sample")]
 
     def select_sql(self, expression: exp.Select) -> str:
+        into = expression.args.get("into")
+        if self.SUPPORT_SELECT_INTO == False and into:
```
georgesittas, line 2026:
```suggestion
        if not self.SUPPORT_SELECT_INTO and into:
```

sqlglot/generator.py
```diff
@@ -2019,6 +2022,10 @@ def after_limit_modifiers(self, expression: exp.Expression) -> t.List[str]:
...
+            expression.args["into"].pop()
```
georgesittas, line 2027:
```suggestion
            into.pop()
```

sqlglot/generator.py
```diff
@@ -2063,7 +2070,19 @@ def select_sql(self, expression: exp.Select) -> str:
             self.sql(expression, "into", comment=False),
             self.sql(expression, "from", comment=False),
         )
-        return self.prepend_ctes(expression, sql)
+
+        sql = self.prepend_ctes(expression, sql)
+
+        if self.SUPPORT_SELECT_INTO == False and into:
```
georgesittas, line 2076:
```suggestion
        if not self.SUPPORT_SELECT_INTO and into:
```

sqlglot/generator.py
```diff
@@ -2063,7 +2070,19 @@ def select_sql(self, expression: exp.Select) -> str:
...
+            # Ensures that SELECT INTO TEMP UNLOGGED -> CREATE TEMPORARY TABLE
+            kind = (
+                " TEMPORARY"
+                if into.args.get("temporary")
+                else (" UNLOGGED" if into.args.get("unlogged") else "")
+            )
+            sql = f'CREATE{kind} TABLE {into.args["this"]} AS {sql}'
```
georgesittas, line 2083: - `.this` is a shorthand for `.args["this"]`
- you shouldn't interpolate expressions because they're implicitly generated using the default dialect, you need to make sure the generation goes through the `sql` method which carries on the dialect information

```suggestion
            sql = f'CREATE{kind} TABLE {self.sql(into.this)} AS {sql}'
```

sqlglot/generator.py
```diff
@@ -2063,7 +2070,19 @@ def select_sql(self, expression: exp.Select) -> str:
...
+
+        sql = self.prepend_ctes(expression, sql)
+
+        if self.SUPPORT_SELECT_INTO == False and into:
+            # Ensures that SELECT INTO TEMP UNLOGGED -> CREATE TEMPORARY TABLE
+            kind = (
+                " TEMPORARY"
+                if into.args.get("temporary")
+                else (" UNLOGGED" if into.args.get("unlogged") else "")
+            )
```
georgesittas, line 2082: Let's rename this to `table_kind` since `kind` is used above in a different context. Also, I'm not sure if `UNLOGGED` is correct here - have you checked whether dialects support the `CREATE UNLOGGED ... TABLE ...` syntax?

My guess is you need to just do `table_kind = " TEMPORARY" if into.args.get("temporary") else ""` here.

georgesittas (commented):
Ok did a first pass @giorgosnikolaou

sqlglot/generator.py
georgesittas: Ok nevermind, I just verified that postgres supports the `CREATE UNLOGGED TABLE ... AS ...` syntax, so we can keep this logic.

georgesittas: The ideal case actually would be to control whether or not to generate `UNLOGGED` with another flag, because that way we would avoid generating invalid CTAS statements for dialects that don't support unlogged tables.

So I'd rewrite this into something that looks like this:

```python
if into.args.get("temporary"):
    kind = " TEMPORARY"
elif self.SUPPORTS_UNLOGGED_TABLES and into.args.get("unlogged"):
    kind = " UNLOGGED"
else:
    kind = ""
```

tests/dialects/test_postgres.py
```diff
@@ -99,6 +105,12 @@ def test_ddl(self):
                 "postgres": "CREATE TABLE products (product_no INT, name TEXT, price DECIMAL, UNIQUE (product_no, name))"
             },
         )
+        self.validate_all(
+            "WITH t(c) AS (SELECT 1) SELECT * INTO TEMP UNLOGGED foo FROM (SELECT c AS c FROM t) AS temp",
+            write={
+                "postgres": "WITH t(c) AS (SELECT 1) SELECT * INTO TEMPORARY foo FROM (SELECT c AS c FROM t) AS temp",
+            },
+        )
```
georgesittas, line 113: This test isn't quite correct, you can't have both `TEMP` and `UNLOGGED` in the same statement.

```suggestion
```

georgesittas (approved):
Looks good, will merge and do a small cleanup.

Nice work @giorgosnikolaou!

PR #2823, opened 2024-01-13
Fix!(parser): parse struct unnamed types into DataType instead of Identifier

georgesittas:


tobymao (approved):

pkit:
@georgesittas doesn't work for ClickHouse
```
print(repr(sqlglot.parse_one("CREATE TABLE t (tt Tuple(String, Int32))", read="clickhouse")))

Create(
  this=Schema(
    this=Table(
      this=Identifier(this=t, quoted=False)),
    expressions=[
      ColumnDef(
        this=Identifier(this=tt, quoted=False),
        kind=DataType(
          this=Type.STRUCT,
          expressions=[
            Identifier(this=String, quoted=False),
            Identifier(this=Int32, quoted=False)],
          nested=True,
          prefix=False))]),
  kind=TABLE,
  exists=False)
```

PR #2829, opened 2024-01-13
Fix(clickhouse): add ipv4/6 data type parser

pkit:
See https://clickhouse.com/docs/en/sql-reference/data-types/ipv4 and https://clickhouse.com/docs/en/sql-reference/data-types/ipv6
It's pretty similar to Presto `IPADDRESS` type, but sematically different as Presto type is IPV6 with IPV4 convertion.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3662,6 +3662,8 @@ class DataType(Expression):
         INTERVAL = auto()
         IPADDRESS = auto()
         IPPREFIX = auto()
+        IPV4 = auto()
+        IPV6 = auto()
         JSON = auto()
         JSONB = auto()
         LONGBLOB = auto()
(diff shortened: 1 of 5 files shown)


georgesittas (approved):
lgtm, thanks

sqlglot/expressions.py
```diff
@@ -3662,6 +3662,8 @@ class Type(AutoName):
         INTERVAL = auto()
         IPADDRESS = auto()
         IPPREFIX = auto()
+        IPV4 = auto()
```
tobymao, line 3665: is ipv4 the same as ip address?

pkit: Nope.
> It's pretty similar to Presto IPADDRESS type, but sematically different as Presto type is IPV6 with IPV4 convertion.

tobymao:
thanks!

pkit:
If I understand Presto docs correctly then ClickHouse `IPv6` can be cast to `IPADDRESS` but `IPADDRESS` cannot be really cast to `IPv6` when it represents ipv4 address.

PR #2832, opened 2024-01-14
Feat(clickhouse): AggregateFunction data type

pkit:
these can be used in column definitions:
- AggregateFunction
- SimpleAggregateFunction

See https://clickhouse.com/docs/en/sql-reference/data-types/aggregatefunction and https://clickhouse.com/docs/en/sql-reference/data-types/simpleaggregatefunction

These are pretty peculiar, 'cos the first argument is either a "function identifier" or a "parameterized function identifier".
`sqlglot` has no notion of both, so I've just opted out for usual identifiers/functions, works in my case.
From ClickHouse docs it's not clear where else it can be used except column definitions, probably nowhere.
For expressions they use special functions like `initializeAggregation()` or `finalizeAggregation()`

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -148,6 +148,11 @@ class Parser(metaclass=_Parser):
         TokenType.ENUM16,
     }
 
+    AGGREGATE_TYPE_TOKENS = {
+        TokenType.AGGREGATEFUNCTION,
+        TokenType.SIMPLEAGGREGATEFUNCTION,
+    }
+
     TYPE_TOKENS = {
         TokenType.BIT,
         TokenType.BOOLEAN,
@@ -241,6 +246,7 @@ class Parser(metaclass=_Parser):
         TokenType.NULL,
         *ENUM_TYPE_TOKENS,
         *NESTED_TYPE_TOKENS,
+        *AGGREGATE_TYPE_TOKENS,
     }
 
     SIGNED_TO_UNSIGNED_TYPE_TOKEN = {
@@ -3627,6 +3633,7 @@ class Parser(metaclass=_Parser):
 
         nested = type_token in self.NESTED_TYPE_TOKENS
         is_struct = type_token in self.STRUCT_TYPE_TOKENS
+        is_aggregate = type_token in self.AGGREGATE_TYPE_TOKENS
         expressions = None
         maybe_func = False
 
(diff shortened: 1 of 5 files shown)


tobymao:
i'm not sure we need the option, you can simplify this by not having the allow_functions flag and just always parse em

pkit:
Not really, it fails in other places, where functions are not really allowed. I've tried.

sqlglot/parser.py
```diff
@@ -3612,6 +3622,8 @@ def _parse_types(
                         type_name = f"{type_name}.{self._advance_any() and self._prev.text}"
 
                     return exp.DataType.build(type_name, udt=True)
+                elif allow_functions:
+                    return identifier
```
tobymao, line 3626: do you have a test case that hits this line?

pkit: yep, all the cases with `AggregateFunction(func, ...)` `func` is a bare function name identifier

pkit: Maybe it needs to be restricted to a very specific set of parent data types though...

tobymao: doesn't SimpleAggregateFunction(sum, Float64) get parsed by line 3602, wouldn't it turn SimpleAggregateFunction into a function?

tobymao: ah i see what you're saying, yea, we'd only need to parse functions when it's simpleaggfunc or the other

pkit: `SimpleAggregateFunction` is `nested=True` data type.
We are now in un-nesting part at that point.

pkit: Yeah. I think flag is ugly.
So you're suggesting to add a new data type family? And then parse it separately?
Putting only `SimpleAggregateFunction` and `AggregateFunction` there?

pkit: something like:
```python
if self._match_set(self.AGGREGATE_TYPE_TOKENS):
    # do the bare identifier + function stuff

```

pkit: Ok, I think I've got a better idea. Will propose tomorrow.

tobymao: exactly

georgesittas: Hey @pkit, any updates on this PR?

pkit: Soon. It was my birthday yesterday. Lol.

tobymao: happy birthday!

georgesittas: Ahah gotcha - happy birthday! 😄

pkit: Thanks!
pushed a new one

georgesittas (approved):

Issue #2831, opened 2024-01-14
Transpiling Oracle > duckdb TO_TIMESTAMP function and time format issue


Rhiyo:
Hello, I'm trying to transpile between Oracle and Duck DB, an example of my oracle code is:

`SELECT * FROM wh.table WHERE ts > TO_TIMESTAMP('2024-12-12 12:12:12.000000`,'YYY-MM-DD HH24:MI:SS.FF6')`

The TO_TIMESTAMP but keeps the same original parameters. DuckDB TO_TIMESTAMP function expects a double, the replacement function would be `strptime` however the time format would also needs to be changed to `%Y-%m-%d %H:%M:%S.%f`

Right now I am manually replacing both, if there is some other SQLGLOT functionality that does this and I am missing it, I apologize in advance.

pkit:
Looks like a bug to me:
```python
import sqlglot
sql = """
SELECT * FROM wh.table WHERE ts > TO_TIMESTAMP('2024-12-12 12:12:12.000000','YYY-MM-DD HH24:MI:SS.FF6')
"""
expr = sqlglot.parse_one(sql, read="oracle")
print(repr(expr))
```
```
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=table, quoted=False),
      db=Identifier(this=wh, quoted=False))),
  where=Where(
    this=GT(
      this=Column(
        this=Identifier(this=ts, quoted=False)),
      expression=Anonymous(
        this=TO_TIMESTAMP,
        expressions=[
          Literal(this=2024-12-12 12:12:12.000000, is_string=True),
          Literal(this=YYY-MM-DD HH24:MI:SS, is_string=True)]))))

```
It means that for "oracle" dialect doesn't recognize `TO_TIMESTAMP` as `exp.StrToTime` and thus it's transpiled "as-is".

Rhiyo:
@pkit I believe it may be the same for TO_DATE

pkit:
Lol.
```
SELECT TO_DATE('2024-12-12', 'YYYY-MM-DD') ==> SELECT CAST(STRPTIME('2024-12-12', '%Y-%m-%d') AS DATE)
```
Is it ok?^

pkit:
PR updated.

Rhiyo:
I believe that is correct :)

PR #2833, opened 2024-01-14
Fix(oracle): TO_TIMESTAMP not parsed as StrToTime

pkit:
Fixes: #2831

Supporting only `.FF6` is probably too strict. Open to suggestions (python uses 6 digits with padding for `.%f` so...).


tobymao (approved):
@pkit thanks so much for your contributions, really appreciate it

PR #2840, opened 2024-01-17
Feat: add dameng dialect

liugddx:
https://www.dameng.com/

### DIFF
--- a/sqlglot/dialects/dameng.py
+++ b/sqlglot/dialects/dameng.py
@@ -10,15 +10,12 @@ class Dameng(Oracle):
         }
 
     class Generator(Oracle.Generator):
-        CAST_MAPPING = {}
 
         TYPE_MAPPING = {
             **Oracle.Generator.TYPE_MAPPING
         }
 
         LAST_DAY_SUPPORTS_DATE_PART = False
-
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/dameng.py
```diff
@@ -5,18 +5,11 @@
 
 class Dameng(Oracle):
     class Parser(Oracle.Parser):
-        FUNCTIONS = {
-            **Oracle.Parser.FUNCTIONS
-        }
+        FUNCTIONS = {**Oracle.Parser.FUNCTIONS}
```
tobymao, line 8: none of these are necessary, please delete

liugddx: > none of these are necessary, please delete

Done. Thank you for your review.

georgesittas:
@liugddx thanks for the PRs, but it seems like dameng and gbase8a don't have any new behavior and so we're closing them for now as you can just use Oracle and MySQL respectively. If you planned on adding new logic to them you can let us know (e.g. in [Slack](https://tobikodata.com/slack)) and we can chat about it before you proceed with a new PR to make sure we're in sync.

PR #2843, opened 2024-01-17
Fix(clickhouse,doris): fix the transformation of ArraySum

echo-hhj:
https://clickhouse.com/docs/en/sql-reference/functions/array-functions#arraysum
https://doris.apache.org/docs/dev/sql-manual/sql-functions/array-functions/array-sum/

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -104,6 +104,7 @@ class ClickHouse(Dialect):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "ANY": exp.AnyValue.from_arg_list,
+            "ARRAYSUM": exp.ArraySum.from_arg_list,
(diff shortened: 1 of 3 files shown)


georgesittas:
Please add a test to ensure ClickHouse -> ClickHouse works as expected, i.e. that the generator preserves `arraySum`.

georgesittas (approved):
Tested this locally, looks good. Will merge and add the test, but keep in mind to test the roundtrip as well for your next PRs :)

georgesittas:
Nvm I needed to add a generator entry to ensure the -> clickhouse direction works when coming from other dialects. The roundtrip worked because we don't normalize function names in clickhouse so arraySum would be preserved.

See https://github.com/tobymao/sqlglot/commit/7949a4f295fc0a9f0becca9b6460f8517ec733f1

echo-hhj:
ok, I get it.

PR #2845, opened 2024-01-17
feat(snowflake): SHOW SCHEMAS/OBJECTS

### LINKED ISSUE #2784: Support for SHOW SCHEMAS and SHOW OBJECTS (Snowflake)
- [SHOW SCHEMAS docs](https://docs.snowflake.com/en/sql-reference/sql/show-schemas)
- [SHOW OBJECTS docs](https://docs.snowflake.com/en/sql-reference/sql/show-objects)
 
NB: Although not mentioned in the docs, SHOW OBJECTS supports a similar syntax to SHOW SCHEMAS, eg: this is valid
```
show terse objects in db1.schema1 starts with 'a' limit 1 from 'b';
```

tekumara:
resolves #2784

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -503,7 +503,7 @@ class Snowflake(Dialect):
         SHOW_PARSERS = {
             "SCHEMAS": _show_parser("SCHEMAS"),
             "TERSE SCHEMAS": _show_parser("SCHEMAS"),
-            "OBJECTS": _show_parser("SCHEMAS"),
+            "OBJECTS": _show_parser("OBJECTS"),
             "TERSE OBJECTS": _show_parser("OBJECTS"),
             "PRIMARY KEYS": _show_parser("PRIMARY KEYS"),
             "TERSE PRIMARY KEYS": _show_parser("PRIMARY KEYS"),
@@ -624,11 +624,11 @@ class Snowflake(Dialect):
                 if self._match_text_seq("ACCOUNT"):
                     scope_kind = "ACCOUNT"
                 elif self._match_set(self.DB_CREATABLES):
-                    scope_kind = self._prev.text
+                    scope_kind = self._prev.text.upper()
                     if self._curr:
                         scope = self._parse_table()
                 elif self._curr:
-                    scope_kind = "TABLE"
+                    scope_kind = "SCHEMA" if this == "OBJECTS" else "TABLE"
                     scope = self._parse_table()
 
             starts_with = self._parse_string() if self._match(TokenType.STARTS_WITH) else None
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -501,6 +501,10 @@ class Snowflake(Dialect):
         }
 
         SHOW_PARSERS = {
+            "SCHEMAS": _show_parser("SCHEMAS"),
+            "TERSE SCHEMAS": _show_parser("SCHEMAS"),
+            "OBJECTS": _show_parser("SCHEMAS"),
+            "TERSE OBJECTS": _show_parser("OBJECTS"),
             "PRIMARY KEYS": _show_parser("PRIMARY KEYS"),
             "TERSE PRIMARY KEYS": _show_parser("PRIMARY KEYS"),
             "COLUMNS": _show_parser("COLUMNS"),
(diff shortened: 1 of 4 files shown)


georgesittas:
Hey @tekumara, thanks a lot for the PR. We'll review soon 👍

sqlglot/expressions.py
```diff
@@ -1130,9 +1130,12 @@ class SetItem(Expression):
 class Show(Expression):
     arg_types = {
         "this": True,
+        "terse": False,
         "target": False,
         "offset": False,
+        "starts_with": False,
         "limit": False,
+        "from_": False,
```
georgesittas, line 1138: Let's just do `"from"` instead and then you can instantiate the expression using the `**{"from": ..., ....}` syntax

sqlglot/tokens.py
```diff
@@ -341,8 +341,10 @@ class TokenType(AutoName):
     SOME = auto()
     SORT_BY = auto()
     START_WITH = auto()
+    STARTS_WITH = auto()
     STRUCT = auto()
     TABLE_SAMPLE = auto()
+    TERSE = auto()
```
georgesittas, line 347: Get rid of these two tokens, I don't think we need them. Use `_match_text_seq` instead in the parser.

sqlglot/dialects/snowflake.py
```diff
@@ -610,21 +614,39 @@ def _parse_show_snowflake(self, this: str) -> exp.Show:
...
                 elif self._curr:
-                    scope_kind = "TABLE"
+                    scope_kind = "SCHEMA" if this == "OBJECTS" else "TABLE"
                     scope = self._parse_table()
 
+            starts_with = self._parse_string() if self._match(TokenType.STARTS_WITH) else None
+
+            limit = self._parse_limit()
+
+            from_ = self._parse_string() if self._match(TokenType.FROM) else None
```
georgesittas, line 638: Can also move these inline in the `self.expression` return call.

georgesittas (commented):
Looks good, a few comments but should be good to go otherwise

sqlglot/tokens.py
tekumara: I've removed `TERSE` but I'm not sure how to remove the `STARTS WITH` token ... if i do the `STARTS` ends up as `VAR` and is consumed as the database name, 

eg:
```
show terse schemas in database db1 starts with 'a' limit 10 from 'b'` I get:
```

without a `STARTS WITH` token is parsed as:
```
[
    <Token token_type: TokenType.SHOW, text: show, line: 1, col: 4, start: 0, end: 3, comments: []>, 
    <Token token_type: TokenType.VAR, text: terse, line: 1, col: 10, start: 5, end: 9, comments: []>,
    <Token token_type: TokenType.VAR, text: schemas, line: 1, col: 18, start: 11, end: 17, comments: []>,
    <Token token_type: TokenType.IN, text: in, line: 1, col: 21, start: 19, end: 20, comments: []>,
    <Token token_type: TokenType.DATABASE, text: database, line: 1, col: 30, start: 22, end: 29, comments: []>,
    <Token token_type: TokenType.VAR, text: db1, line: 1, col: 34, start: 31, end: 33, comments: []>,
    <Token token_type: TokenType.VAR, text: starts, line: 1, col: 41, start: 35, end: 40, comments: []>,
    <Token token_type: TokenType.WITH, text: with, line: 1, col: 46, start: 42, end: 45, comments: []>,
    <Token token_type: TokenType.STRING, text: a, line: 1, col: 50, start: 47, end: 49, comments: []>,
    <Token token_type: TokenType.LIMIT, text: limit, line: 1, col: 56, start: 51, end: 55, comments: []>,
    <Token token_type: TokenType.NUMBER, text: 10, line: 1, col: 59, start: 57, end: 58, comments: []>,
    <Token token_type: TokenType.FROM, text: from, line: 1, col: 64, start: 60, end: 63, comments: []>,
    <Token token_type: TokenType.STRING, text: b, line: 1, col: 68, start: 65, end: 67, comments: []>
]
```

georgesittas: You can solve this by using `_parse_table_parts` instead of `_parse_table`; the latter has additional logic for alias parsing which isn't needed for this statement. So, something like this (in [these](https://github.com/tobymao/sqlglot/pull/2845/files#diff-c3dbaf62f01fc7a9baf342a000c2351ebea03abdede048f92d234c44e11c7bfeR627-R632) lines):

```python
                    if self._curr:
                        scope = self._parse_table_parts()
                elif self._curr:
                    scope_kind = "SCHEMA" if this == "OBJECTS" else "TABLE"
                    scope = self._parse_table_parts()
```

And so you'd have this in the return statement:

```python
+                    "starts_with": self._match_text_seq("STARTS", "WITH") and self._parse_string(),
```

georgesittas:
@tekumara went ahead and took this to the finish line, see https://github.com/tobymao/sqlglot/commit/567c8232f91579ea590391aee43832acb932dfe5.

Thanks again!

PR #2848, opened 2024-01-18
fix(snowflake): Support more scale values for Snowflake's to_timestamp function

dangoldin:
The docs have 0, 3, and 9 as examples but it works for all values between 0 and 9, inclusive. This removes the check and instead tries to choose the right scale and defaults to None if it fails.

image

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -40,18 +40,23 @@ def _parse_to_timestamp(args: t.List) -> t.Union[exp.StrToTime, exp.UnixToTime, 
             # case: <string_expr> [ , <format> ]
             return format_time_lambda(exp.StrToTime, "snowflake")(args)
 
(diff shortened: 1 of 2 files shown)


tobymao:
@dangoldin in the future, please write some tests, thanks for the pr!

tobymao:
this has been reverted because it doesn't work as expected, i suspect a different approach is needed

dangoldin:
Got it - What part didn't work?

tobymao:
https://github.com/tobymao/sqlglot/commit/1be93e45d8347e5fa8a4e39dad625c6dd66ea461

tobymao:
@dangoldin the generation of sql, you didn't test it

dangoldin:
Ah my bad. I ran the unit tests in the README and thought they'd flag it but see now that they were only testing the existing time scales so the new ones broke things. To your point should have written tests for this.

tobymao:
no worries, next time!

PR #2852, opened 2024-01-18
feat: remove target alias in Trino merge

eakmanrq:
Some connectors for Trino support merge (like Iceberg and Delta) and, similar to Postgres, they require that there is not an alias defined for target. 

This is not added for Presto since Presto does not support MERGE: https://github.com/prestodb/presto/issues/20578

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -4,7 +4,7 @@ import typing as t
 from enum import Enum, auto
 from functools import reduce
 
-from sqlglot import exp
+from sqlglot import exp, transforms
 from sqlglot._typing import E
 from sqlglot.errors import ParseError
 from sqlglot.generator import Generator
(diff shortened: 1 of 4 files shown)


georgesittas (approved):

sqlglot/dialects/dialect.py
```diff
@@ -993,3 +993,33 @@ def no_last_day_sql(self: Generator, expression: exp.LastDay) -> str:
     minus_one_day = exp.func("date_sub", plus_one_month, 1, "day")
 
     return self.sql(exp.cast(minus_one_day, "date"))
+
+
+def remove_target_from_merge(self: Generator, expression: exp.Merge) -> str:
```
tobymao, line 998: this should go in transforms.py

tobymao: ah i see

tobymao: why do we need transforms.preprocess at all then? just call transform

georgesittas: Ah, sure. Perhaps we can also get rid of the closure and just do this in a single nesting level, deferring to `Generator.merge_sql` (which will also be reached if we call `self.sql(merge_expression)` at the end of this method

georgesittas: @eakmanrq I can try out this refactor in a followup PR if you want

eakmanrq: sounds good I will just merge this as is

PR #2862, opened 2024-01-19
Fix: qualify alter table table refs in optimizer qualify

z3z1ma:
As per discussion. We should qualify these table references if passed that particular expression type.

### DIFF
--- a/sqlglot/optimizer/qualify_tables.py
+++ b/sqlglot/optimizer/qualify_tables.py
@@ -46,17 +46,16 @@ def qualify_tables(
     db = exp.parse_identifier(db, dialect=dialect) if db else None
     catalog = exp.parse_identifier(catalog, dialect=dialect) if catalog else None
 
-    def _qualify(tree: E) -> E:
-        for reference in tree.find_all(exp.Table):
-            if isinstance(reference.this, exp.Identifier):
-                if not reference.args.get("db"):
-                    reference.set("db", db)
-                if not reference.args.get("catalog") and reference.args.get("db"):
-                    reference.set("catalog", catalog)
-        return tree
+    def _qualify(*tables: exp.Table) -> None:
+        for table in tables:
+            if isinstance(table.this, exp.Identifier):
+                if not table.args.get("db"):
+                    table.set("db", db)
+                if not table.args.get("catalog") and table.args.get("db"):
+                    table.set("catalog", catalog)
 
     if isinstance(expression, exp.AlterTable):
-        _qualify(expression)
+        _qualify(*expression.find_all(exp.Table))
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/qualify_tables.py
```diff
@@ -46,6 +46,15 @@ def qualify_tables(
     db = exp.parse_identifier(db, dialect=dialect) if db else None
     catalog = exp.parse_identifier(catalog, dialect=dialect) if catalog else None
 
+    if isinstance(expression, exp.AlterTable):
+        for reference in expression.find_all(exp.Table):
+            if isinstance(reference.this, exp.Identifier):
+                if not reference.args.get("db"):
+                    reference.set("db", db)
+                if not reference.args.get("catalog") and reference.args.get("db"):
+                    reference.set("catalog", catalog)
```
georgesittas, line 55: Should we factor this section out? it's the same as in L78-82.

georgesittas (approved):
LGTM, just one comment.

sqlglot/optimizer/qualify_tables.py
tobymao: this is not the right approach

tobymao: i'm going to factor this out and use walk

tobymao: @z3z1ma can you change it to do this instead

```python

    for table in expression.find_all(exp.Table):
        if isinstance(table.this, exp.Identifier):
            if not table.args.get("db"):
                table.set("db", db)
            if not table.args.get("catalog") and table.args.get("db"):
                table.set("catalog", catalog)
```

delete the other sections

tobymao: move your test into qualify_tables.sql

sqlglot/optimizer/qualify_tables.py
```diff
@@ -46,6 +46,18 @@ def qualify_tables(
...
+        if isinstance(table.this, exp.Identifier):
+            if not table.args.get("db"):
+                table.set("db", db)
+            if not table.args.get("catalog") and table.args.get("db"):
+                table.set("catalog", catalog)
+
+    if not isinstance(expression, exp.Subqueryable):
+        for table in expression.find_all(exp.Table):
+            _qualify(table)
+        return t.cast(E, expression)
```
tobymao, line 59: we don't want to do an early return here

tobymao, line 59:
```suggestion
```

tobymao (approved):

PR #2865, opened 2024-01-20
Feat!(optimizer): create aliases for CTE Table references

georgesittas:


sqlglot/optimizer/qualify_tables.py
```diff
@@ -114,5 +114,10 @@ def _qualify(table: exp.Table) -> None:
                 if isinstance(udtf, exp.Values) and not table_alias.columns:
                     for i, e in enumerate(udtf.expressions[0].expressions):
                         table_alias.append("columns", exp.to_identifier(f"_col_{i}"))
+            else:
+                for table in scope.cte_table_references[name]:
```
tobymao, line 118: why do you need cte table refs, why not find all unqualified tabkes?

georgesittas: Can do that too, this way we simply avoid the traversal.

georgesittas: This approach seemed like almost 0 cost and doesn't affect queries that don't have CTE Table refs. The other way around we'd have to traverse _all_ queries and check for unaliased tables which can be a significant overhead.

tobymao: ok we can keep it but let’s not use default dict

tobymao: since others will access it. default dict is ok for internal

tobymao: is there no way to get access to tables here?

georgesittas: it's tricky to access the tables in `qualify_tables` because we loop over the sources and they're created in such a way that we store scopes for CTEs, so there's no reference to the actual table at hand.

> since others will access it. default dict is ok for internal

Makes sense, will update

georgesittas: https://github.com/tobymao/sqlglot/pull/2865/commits/61e40cab9167a8944960a59209b5f98d8aaa0969

sqlglot/optimizer/qualify_tables.py
```diff
@@ -114,5 +114,14 @@ def _qualify(table: exp.Table) -> None:
                 if isinstance(udtf, exp.Values) and not table_alias.columns:
                     for i, e in enumerate(udtf.expressions[0].expressions):
                         table_alias.append("columns", exp.to_identifier(f"_col_{i}"))
+            else:
+                for node, parent, _ in scope.walk():
+                    if (
+                        isinstance(node, exp.Table)
+                        and not node.alias
+                        and isinstance(parent, (exp.From, exp.Join))
```
tobymao, line 122: when does the parent matter? could you just do find_all?

georgesittas: It matters because you can have `Table`s in `CREATE TABLE <table>` statements, as well as `INSERT INTO` statements and also hints. I initially did the find_all version but had several test failures because of this. For example:

```diff
- WITH cte AS (SELECT _q_0.a AS a FROM (SELECT x.a AS a FROM x AS x) AS _q_0) CREATE TABLE s AS SELECT cte.a AS a FROM cte AS cte
+ WITH cte AS (SELECT _q_0.a AS a FROM (SELECT x.a AS a FROM x AS x) AS _q_0) CREATE TABLE s AS s AS SELECT cte.a AS a FROM cte AS cte
?                                                                                              +++++
```

```diff
- WITH cte AS (SELECT y.b AS b FROM y AS y) INSERT INTO s SELECT cte.b AS b FROM cte AS cte
+ WITH cte AS (SELECT y.b AS b FROM y AS y) INSERT INTO s AS s SELECT cte.b AS b FROM cte AS cte
?                                                        +++++
```

```diff
- SELECT /*+ BROADCAST(x), MERGE(x, y) */ x.a AS a, y.c AS c FROM x AS x JOIN y AS y ON x.b = y.b
+ SELECT /*+ BROADCAST(x AS m), MERGE(x AS m, y AS n) */ x.a AS a, y.c AS c FROM x AS x JOIN y AS y ON x.b = y.b
?                       +++++          +++++   +++++
```

PR #2866, opened 2024-01-20
feat: add support for rename column

gableh:
Adding support for ALTER TABLE RENAME COLUMN e.g https://trino.io/docs/current/sql/alter-table.html


sqlglot/expressions.py
```diff
@@ -6578,6 +6582,28 @@ def rename_table(old_name: str | Table, new_name: str | Table) -> AlterTable:
...
+        new_column: The new name of the column
+
+    Returns:
+        Alter table expression
+    """
+    table = to_table(table_name)
+    return AlterTable(
+        this=table,
+        actions=[
+            RenameColumn(this=old_column, to=new_column, exists=exists),
```
georgesittas, line 6602: This isn't quite right, you need to convert possible strings into `Column`s (same as what you did on L6598):
```suggestion
            RenameColumn(this=to_column(old_column), to=to_column(new_column), exists=exists),
```

sqlglot/parser.py
```diff
@@ -5320,7 +5320,13 @@ def _parse_alter_table_drop(self) -> t.List[exp.Expression]:
...
         return self._parse_csv(self._parse_drop_column)
 
-    def _parse_alter_table_rename(self) -> exp.RenameTable:
+    def _parse_alter_table_rename(self) -> exp.RenameTable | exp.RenameColumn:
+        if self._match_text_seq("COLUMN"):
+            exists = self._parse_exists()
+            old_column = self._parse_column()
+            self._match_text_seq("TO")
+            new_column = self._parse_column()
+            return self.expression(exp.RenameColumn, this=old_column, to=new_column, exists=exists)
```
georgesittas, line 5329: Nit:
```suggestion
            return self.expression(exp.RenameColumn, this=old_column, to=new_column, exists=exists)

```

tests/test_expressions.py
```diff
@@ -943,6 +943,16 @@ def test_rename_table(self):
             "ALTER TABLE t1 RENAME TO t2",
         )
 
+    def test_rename_column(self):
```
georgesittas, line 946: Let's add another test that doesn't have the `exists` argument here

tests/test_parser.py
```diff
@@ -564,6 +564,16 @@ def test_rename_table(self):
...
+    def test_rename_column(self):
+        self.assertEqual(
+            parse_one("ALTER TABLE foo RENAME COLUMN c1 TO c2").sql(),
+            "ALTER TABLE foo RENAME COLUMN c1 TO c2",
+        )
+        self.assertEqual(
+            parse_one("ALTER TABLE foo RENAME COLUMN IF EXISTS c1 TO c2").sql(),
+            "ALTER TABLE foo RENAME COLUMN IF EXISTS c1 TO c2",
+        )
+
```
georgesittas, line 576: Let's move these to `identity.sql`

tests/test_expressions.py
georgesittas: Also, let's move these to `test_build.py`

sqlglot/parser.py
```diff
@@ -5320,7 +5320,13 @@ def _parse_alter_table_drop(self) -> t.List[exp.Expression]:
         self._retreat(index)
         return self._parse_csv(self._parse_drop_column)
 
-    def _parse_alter_table_rename(self) -> exp.RenameTable:
+    def _parse_alter_table_rename(self) -> exp.RenameTable | exp.RenameColumn:
+        if self._match_text_seq("COLUMN"):
```
georgesittas, line 5324:
```suggestion
        if self._match(TokenType.COLUMN):
```

georgesittas (commented):
Thanks for the PR @gableh!

This is possibly breaking, so we need to ensure that all dialects either follow the same syntax spec, or have a safe fallback for parsing different specs into `exp.Command`, i.e. what we're already doing.

Can you investigate what the other dialects do for `ALTER TABLE ... RENAME COLUMN`?

sqlglot/expressions.py
gableh: updated to use old_column_name and new_column_name in args, similar to table_name

gableh:
Regarding `what the other dialects do for ALTER TABLE ... RENAME COLUMN?`

Clickhouse, Presto, Trino and MySQL have the same syntax we are looking for here

PostgreSQL, Redshift, Snowflake, Spark (with v2 tables), sqlite, tableau and teradata support without the IF EXISTS syntax  

Bigquery has a superset -> able to support multiple column renaming at once via ` RENAME COLUMN [IF EXISTS] column_to_column[, ...]`

DuckDB has a superset ->  able to use ALTER TABLE RENAME COLUMN but the COLUMN literal isnt required e.g ALTER TABLE integers RENAME i TO j

Databricks supports without the IF EXISTS syntax but has an additional specification of `field_identifiers` which im not sure what it means

t-sql, doris, starrocks, oracle, hive and drill do not support this syntax, having different methods of renaming a column. 


Given that the support is split roughly 50-50, would it still make sense to have it in the generic parser.py or should it be shifted out into the dialect specific parsers?

sqlglot/expressions.py
```diff
@@ -6578,6 +6582,33 @@ def rename_table(old_name: str | Table, new_name: str | Table) -> AlterTable:
...
+    old_column_name: str | Column,
+    new_column_name: str | Column,
+    exists: t.Optional[bool] = None,
+) -> AlterTable:
+    """Build ALTER TABLE... RENAME COLUMN... expression
+
+    Args:
+        table_name: Name of the table
+        old_column: The old name of the column
+        new_column: The new name of the column
```
georgesittas, line 6596:
```suggestion
        new_column: The new name of the column
        exists: Whether or not to add the `IF EXISTS` clause
```

georgesittas:
> Regarding `what the other dialects do for ALTER TABLE ... RENAME COLUMN?`
> 
> Clickhouse, Presto, Trino and MySQL have the same syntax we are looking for here
> 
> PostgreSQL, Redshift, Snowflake, Spark (with v2 tables), sqlite, tableau and teradata support without the IF EXISTS syntax
> 
> Bigquery has a superset -> able to support multiple column renaming at once via ` RENAME COLUMN [IF EXISTS] column_to_column[, ...]`
> 
> DuckDB has a superset -> able to use ALTER TABLE RENAME COLUMN but the COLUMN literal isnt required e.g ALTER TABLE integers RENAME i TO j
> 
> Databricks supports without the IF EXISTS syntax but has an additional specification of `field_identifiers` which im not sure what it means
> 
> t-sql, doris, starrocks, oracle, hive and drill do not support this syntax, having different methods of renaming a column.
> 
> Given that the support is split roughly 50-50, would it still make sense to have it in the generic parser.py or should it be shifted out into the dialect specific parsers?

Yes, it makes sense to have this in the base parser and have it consume as much of this syntax as possible. For this PR, let's just make sure that the syntax not currently covered doesn't result in a `ParserError` and we always parse unsupported variants into `exp.Command`. Let's get some tests in for this and then it should be good to go.

sqlglot/parser.py
```diff
@@ -5339,7 +5350,11 @@ def _parse_alter(self) -> exp.AlterTable | exp.Command:
 
         parser = self.ALTER_PARSERS.get(self._prev.text.upper()) if self._prev else None
         if parser:
-            actions = ensure_list(parser(self))
+            parse_result = parser(self)
+            if parse_result is None:
+                return self._parse_as_command(start)
+
+            actions = ensure_list(parse_result)
```
georgesittas, line 5357: This can be reverted, if `parse_result` is `None` then `ensure_list` will return an empty list for `actions` and the `self.expression` call below will fail since `actions` is a required argument.

```suggestion
            actions = ensure_list(parse_result)
```

georgesittas (approved):
LGTM - made one final suggestion.

Thanks @gableh!

sqlglot/parser.py
```diff
@@ -5339,7 +5350,7 @@ def _parse_alter(self) -> exp.AlterTable | exp.Command:
 
         parser = self.ALTER_PARSERS.get(self._prev.text.upper()) if self._prev else None
         if parser:
-            actions = ensure_list(parser(self))
+            actions = ensure_list(parse_result)
```
georgesittas, line 5353:
```suggestion
            actions = ensure_list(parser(self))
```

sqlglot/parser.py
gableh: I think this would be considered a breaking change though since the output of this would then always be an exp.AlterTable instead of exp.Command. Would it be better to change it to 
```
            if not self._curr and actions is not None:
```
then?

georgesittas: Nvm, I was wrong apparently. Your change makes sense after all because failing at `expression` is no good.

georgesittas: > I think this would be considered a breaking change though since the output of this would then always be an exp.AlterTable instead of exp.Command. Would it be better to change it to
> 
> ```
>             if not self._curr and actions is not None:
> ```
> 
> then?

Yeah, you can do `and actions` because `actions` will be an empty list. Can you make this change as well?

georgesittas (approved):

Issue #2839, opened 2024-01-16
Incorrect parse between redshift JSON_EXTRACT_PATH_TEXT and databricks JSON path expression


FlaviovLeal:
Hello,

So I tried the following code and got no errors. But the result SQL would not work on Databricks. 

```Python
import sqlglot

sql = """select JSON_EXTRACT_PATH_TEXT('{ "farm": {"barn": { "color": "red", "feed stocked": true }}}', 'farm', 'barn', 'color');"""
print(sqlglot.transpile(sql, read='redshift', write='databricks')[0])
# The printed results are as follow:
#  SELECT JSON_EXTRACT_PATH_TEXT('{ "farm": {"barn": { "color": "red", "feed stocked": true }}}', 'farm', 'barn', 'color')
```

There are two possible solutions that would work on Databricks

```SQL
select '{ "farm": {"barn": { "color": "red", "feed stocked": true }}}':['farm']['barn']['color'];
```
And 
```SQL
select '{ "farm": {"barn": { "color": "red", "feed stocked": true }}}':farm.barn.color;
```

I recommend the first solution, as the following would work
```SQL
select '{ "farm": {"barn": { "color": "red", "feed stocked": true }}}':['farm']['barn']['feed stocked'];
```
but the following would not 
```SQL
select '{ "farm": {"barn": { "color": "red", "feed stocked": true }}}':farm.barn.feed stocked;
```

I tried looking into the code but I was unable to find a solution. On the databricks side, it looks like it would work ([Transformer](https://github.com/tobymao/sqlglot/blob/a6d396b79cfa199a66b04af9ed62bcd7cd619096/sqlglot/dialects/databricks.py#L52C17-L52C28), [Parser](https://github.com/tobymao/sqlglot/blob/a6d396b79cfa199a66b04af9ed62bcd7cd619096/sqlglot/dialects/databricks.py#L29C13-L29C22)
), as long as no more than one keyword was used and there were no spaces on the keyword ('color' would work, but 'feed stocked' would not). But I'm unsure if it should use [JSONExtract](https://github.com/tobymao/sqlglot/blob/a6d396b79cfa199a66b04af9ed62bcd7cd619096/sqlglot/expressions.py#L4907) as its a child from the binary class and it can take multiple arguments. 

 
Lastly, despite saying it's redshift. The same problem should happen on postgres, as this function exists there as well.

georgesittas:
[...]

Issue #1544, opened 2023-05-04
json access syntax between TrinoSQL and SparkSQL


joshmarcus:
When translating a json object access from TrinoSQL to SparkSQL, sometimes the definition of the object access needs to be transformed.  For example,
This function in Trino SQL (read="trino")
```
json_extract_scalar(column, '$["property"]')
```
is translated to
```
get_json_object(dimensions, '$["property"]')
```

but SparkSQL only uses dot notation, so the second argument needs to be:
```
get_json_object(dimensions, '$.property')
```
**Fully reproducible code snippet**
```
>>> query = """select json_extract_scalar(column, '$["property"]') from table"""
>>> sqlglot.transpile(query, read="trino", write="spark")
['SELECT GET_JSON_OBJECT(column, \'$["property"]\') FROM table']
```
Expected result: 
```
['SELECT GET_JSON_OBJECT(column, \'$.property\') FROM table']
```
**Official Documentation**
https://trino.io/docs/current/functions/json.html#json-path-modes (trino supports both syntaxes)
https://docs.databricks.com/sql/language-manual/functions/get_json_object.html

tobymao:
this is out of scope. it'd require having a json path parser. maybe one day... if someone is willing to implement a json path parser in pure python, i'll accept it.

for now i'd suggest that you write your queries in dot notation if you want them to be compatible, or you can do a custom transform

def transform(node):
   if isinstance(node, exp.JSONExtractScalar):
       ....

georgesittas:
FYI @joshmarcus this is solved in the latest main branch:

```python
>>> import sqlglot
>>> print(sqlglot.transpile("""json_extract_scalar(column, '$["property"]')""", "trino", "spark")[0])
GET_JSON_OBJECT(column, '$.property')
```

PR #2867, opened 2024-01-20
feat: jsonpath parsing

tobymao:
starting point to make https://github.com/tobymao/sqlglot/issues/2839 and https://github.com/tobymao/sqlglot/issues/1544


georgesittas (approved):
Nice!

sqlglot/jsonpath.py
```diff
@@ -0,0 +1,210 @@
...
+        if _match(TokenType.STAR):
+            return _node("wildcard")
+        if _match(TokenType.PLACEHOLDER):
+            start = i
+
+            while True:
+                if _match(TokenType.L_BRACKET):
+                    _parse_bracket()  # nested call which we can throw away
+                if _curr() in (TokenType.R_BRACKET, None):
+                    break
```
georgesittas, line 99: If I understand correctly, this allows us to successfully parse `[?(...)`. Do we wanna add another branch here that raises if `_curr()` returns `None` for better error handling?

georgesittas: Nvm I just saw https://github.com/tobymao/sqlglot/pull/2867/files#diff-c779839b81bb6254e97ad8a6e17e9efbb09031d113a4942040fdd2cc3dcd9275R145, I think it's handled there.

sqlglot/jsonpath.py
```diff
@@ -0,0 +1,210 @@
...
+        return node
+
+    nodes = []
+
+    while _curr():
+        if _match(TokenType.DOLLAR):
+            nodes.append(_node("root"))
+        elif _match(TokenType.DOT):
+            recursive = _prev().text == ".."
+            value = _match(TokenType.VAR) or _match(TokenType.STAR)
```
georgesittas, line 156: What if you have `$.select`? You won't capture the value because it's a `SELECT` token, not a `VAR`, right?

sqlglot/jsonpath.py
```diff
@@ -0,0 +1,210 @@
...
+            nodes.append(
+                _node("recursive" if recursive else "child", value=value.text if value else None)
+            )
+        elif _match(TokenType.L_BRACKET):
+            nodes.append(_parse_bracket())
+        elif _match(TokenType.VAR):
+            nodes.append(_node("key", _prev().text))
```
georgesittas, line 163: What case does this cover? When can you come across a `VAR` outside of the `DOT` branch? Also, ditto about other tokens like `SELECT` that aren't `VAR` but could still be valid keys. They won't be recognized as `VAR`s.

tests/test_jsonpath.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+
+    def test_cts_file(self):
+        with open(os.path.join(FIXTURES_DIR, "jsonpath", "cts.json")) as file:
+            tests = json.load(file)["tests"]
+
+        # sqlglot json path generator rewrites to a normal form
+        overrides = {
+            """$['a',1]""": """$["a",1]""",
+            """$['a',1]""": """$["a",1]""",
```
georgesittas, line 40: are these duplicated on purpose? there are some more duplicates below.

sqlglot/jsonpath.py
```diff
@@ -0,0 +1,210 @@
...
+                if isinstance(literal, str):
+                    node = _node("key", indexes[0])
+                elif isinstance(literal, dict) and literal["kind"] == "filter":
+                    node = _node("selector", indexes[0])
+                else:
+                    node = _node("subscript", indexes[0])
+            else:
+                node = _node("union", indexes)
+        else:
+            raise ParseError(_error("Cannot have empty segment"))
```
georgesittas, line 143: FYI based on https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html, it seems like there's another valid syntax form that I _think_ this section doesn't handle correctly: `[(expression)]`.

sqlglot/jsonpath.py
tobymao: all keywords have been replaced, so there is no select token

sqlglot/jsonpath.py
tobymao: everything is a var basically, again, select and other special sql tokens don''t exist

sqlglot/jsonpath.py
georgesittas: dang, totally missed that - nice!

PR #2874, opened 2024-01-22
Chore(parser): warn when parsing (>1 tokens) SQL into exp.Command

georgesittas:
With this, we're going to start emitting warnings when we generate `exp.Command` expressions due to unsupported syntax. This should help our users understand the corresponding ASTs when produced and take action if needed.


sqlglot/parser.py
```diff
@@ -1216,7 +1216,24 @@ def _retreat(self, index: int) -> None:
...
             self._advance(index - self._index)
 
+    def _warn_unsupported(self) -> None:
+        if len(self._tokens) <= 1:
+            return
+
+        # We use _find_sql because self.sql may comprise multiple chunks, and we're only
+        # interested in emitting a warning for the one being currently processed.
+        sql = self._find_sql(self._tokens[0], self._tokens[-1])
+        if sql:
```
tobymao, line 1226: when would there not be any sql?

georgesittas: if `self.sql` is not ever supplied, which today seems like a possibility? E.g. when you supply a token list directly.

tobymao: we can just get rid of the if sql; check

georgesittas: done https://github.com/tobymao/sqlglot/pull/2874/commits/abac3f5b943a3201960ca528566c1809bb4809da

tobymao (approved):

Issue #2872, opened 2024-01-22
Failed to parse SET statement in Postgres CREATE FUNCTION


edongashi:
**Fully reproducible code snippet**

```py
from sqlglot import parse

parse(
    """
CREATE FUNCTION add(integer, integer) RETURNS integer
    SET search_path TO 'public'
    AS 'select $1 + $2;'
    LANGUAGE SQL
    IMMUTABLE;
""",
    dialect="postgres",
)
```

Raises

```
Traceback (most recent call last):
  File "<...>/create-function-repro.py", line 3, in <module>
    parse(
  File "<...>/python3.9/site-packages/sqlglot/__init__.py", line 86, in parse
    return Dialect.get_or_raise(read or dialect).parse(sql, **opts)
  File "<...>/python3.9/site-packages/sqlglot/dialects/dialect.py", line 442, in parse
    return self.parser(**opts).parse(self.tokenize(sql), sql)
  File "<...>/python3.9/site-packages/sqlglot/parser.py", line 1026, in parse
    return self._parse(
  File "<...>/python3.9/site-packages/sqlglot/parser.py", line 1095, in _parse
    self.raise_error("Invalid expression / Unexpected token")
  File "<...>/python3.9/site-packages/sqlglot/parser.py", line 1136, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 3, Col: 28.
  
CREATE FUNCTION add(integer, integer) RETURNS integer
    SET search_path TO 'public'
    AS 'select $1 + $2;'
    LANGUAGE SQL
    IMMUTABLE;
```

**Official Documentation**

- https://www.postgresql.org/docs/current/sql-createfunction.html

Notice the allowed statement `SET configuration_parameter { TO value | = value | FROM CURRENT }` in `CREATE FUNCTION`.

Unrelated, but `STRICT/RETURNS NULL ON NULL INPUT` also fails, thus I omitted it from the example above.

PR #2875, opened 2024-01-22
Feat: make the CREATE parser more lenient

georgesittas:
This basically allows the create parser to produce a `Command` when it hasn't consumed the whole token stream, meaning that we'll avoid raising `ParseError`s when e.g. unknown table properties are encountered. Instead, we'll simply accept whatever the user supplies, store it as a string and when https://github.com/tobymao/sqlglot/pull/2874 is in, emit a warning to let them know.

The motivation is to avoid having to add first-class support for DDL clauses that may be too engine-specific and not transpilable. Of course, if we ever need to transpile or switch off specific props or analyze the AST, we'll still have to implement parsing and generation logic for the corresponding properties, but hopefully we should not need that often.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1449,6 +1449,9 @@ class Parser(metaclass=_Parser):
                     exp.Clone, this=self._parse_table(schema=True), shallow=shallow, copy=copy
                 )
 
+        if self._curr:
+            return self._parse_as_command(start)
+
         return self.expression(
             exp.Create,
             comments=comments,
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -1449,6 +1449,9 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
                     exp.Clone, this=self._parse_table(schema=True), shallow=shallow, copy=copy
                 )
 
+        if self._curr:
```
tobymao, line 1452: test case?

tobymao (approved):

sqlglot/parser.py
georgesittas: sure, I think there are a couple more props missing based on https://github.com/tobymao/sqlglot/issues/2872, let's see if this picks them up:

> Unrelated, but STRICT/RETURNS NULL ON NULL INPUT also fails, thus I omitted it from the example above.

georgesittas: https://github.com/tobymao/sqlglot/pull/2875/commits/4b441d2348014c41b575175350a6c061712fbc21

PR #2883, opened 2024-01-25
Refactor!: improve transpilation of JSON paths across dialects

### LINKED ISSUE #2839: Incorrect parse between redshift JSON_EXTRACT_PATH_TEXT and databricks JSON path expression

georgesittas:
Fixes #2839, still a WIP; opening for some feedback

References:
- https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract
- https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html
- http://prestodb.io/docs/0.284/functions/json.html
- https://www.postgresql.org/docs/9.3/functions-json.html
- https://docs.aws.amazon.com/redshift/latest/dg/JSON_EXTRACT_PATH_TEXT.html
- https://docs.snowflake.com/en/sql-reference/functions/json_extract_path_text
- https://learn.microsoft.com/en-us/sql/relational-databases/json/validate-query-and-change-json-data-with-built-in-functions-sql-server?view=sql-server-ver16#JSONCompare
- https://www.sqlite.org/json1.html#jex


georgesittas:
One improvement we could do is to consolidate multiple `->` or `->>` applications into a single `JSONExtract` node by merging the respective JSON paths, but I'm thinking to leave it for a future PR. Right now we parse these chained operations into multiple `JSONExtract` nodes:

```python
>>> import sqlglot
>>> sqlglot.parse_one("x -> 'a' -> 'b' -> 'z'", "postgres")
JSONExtract(
  this=JSONExtract(
    this=JSONExtract(
      this=Column(
      expression=JSONPath(
        this=[
          {'kind': 'root'},
          {'kind': 'key', 'value': 'a'}])),
    expression=JSONPath(
      this=[
        {'kind': 'root'},
        {'kind': 'key', 'value': 'b'}])),
  expression=JSONPath(
    this=[
      {'kind': 'root'},
      {'kind': 'key', 'value': 'z'}]))
>>> sqlglot.parse_one("x -> 'a' -> 'b' -> 'z'", "postgres").sql("bigquery")
"JSON_EXTRACT(JSON_EXTRACT(JSON_EXTRACT(x, '$.a'), '$.b'), '$.z')"
>>> sqlglot.parse_one("x -> 'a' -> 'b' -> 'z'", "postgres").sql("duckdb")
"x -> '$.a' -> '$.b' -> '$.z'"
```

sqlglot/dialects/dialect.py
```diff
@@ -498,13 +499,14 @@ def _if_sql(self: Generator, expression: exp.If) -> str:
     return _if_sql
 
 
-def arrow_json_extract_sql(self: Generator, expression: exp.JSONExtract | exp.JSONBExtract) -> str:
-    return self.binary(expression, "->")
+def arrow_json_extract_sql(self: Generator, expression: exp.JSONExtract) -> str:
+    this = expression.this
+    if self.JSON_TYPE_REQUIRED_FOR_EXTRACTION and isinstance(this, exp.Literal) and this.is_string:
+        this.replace(exp.cast(this.copy(), "json"))
```
tobymao, line 505: i think you should make cast default to copy, so you don't need to do this

sqlglot/dialects/dialect.py
```diff
@@ -498,13 +499,14 @@ def _if_sql(self: Generator, expression: exp.If) -> str:
...
 
+    if type(expression) in (exp.JSONExtract, exp.JSONBExtract):
```
tobymao, line 507: maaybe this should be isinstance, and jsonextracscalar shouldn't be a subclass?

georgesittas: yeah I was thinking about this, not sure why these are subclasses of `JSONExtract`

tobymao: shouldn't this only be true if size of json path is one?

sqlglot/dialects/dialect.py
georgesittas: sure

sqlglot/dialects/dialect.py
tobymao: how do we handle multiple paths?

georgesittas: `JSONExtract` by definition has one path, this is supposed to be a helper to generate arrows. Multiple paths would be parsed into nested `JSONExtract` expressions and each of those would be generated separately.

sqlglot/dialects/dialect.py
```diff
@@ -1019,3 +1021,45 @@ def merge_without_target_sql(self: Generator, expression: exp.Merge) -> str:
...
+                    segments.append(_node("subscript", int(text)))
+                else:
+                    segments.append(_node("child", text))
+            elif supports_null_if_invalid:
+                null_if_invalid = arg
+
+        this = seq_get(args, 0)
+        jsonpath = exp.JSONPath(this=segments)
+
+        # This is done to avoid failing in the expression validator due to the arg count
```
tobymao, line 1049: del args[2:]

sqlglot/dialects/dialect.py
tobymao: what i mean is a json path with nested calls,

json_path(x, '$.a.b.c')

sqlglot/dialects/dialect.py
```diff
@@ -1019,3 +1021,45 @@ def merge_without_target_sql(self: Generator, expression: exp.Merge) -> str:
...
+
+        return expr_type(this=this, expression=jsonpath, null_if_invalid=null_if_invalid)
+
+    return _parse_json_extract_path
+
+
+def json_path_segments(self: Generator, expression: exp.JSONPath) -> t.List[str]:
+    segments = []
+    for segment in expression.this:
+        path = generate_json_path([segment], mapping=self.JSON_PATH_MAPPING)
```
tobymao, line 1061: is this right? wouldn't this produce "." when we just want the key name?

tobymao: ah i see you use json_path_mapping

sqlglot/serde.py
```diff
@@ -45,7 +45,7 @@ def load(obj: JSON) -> Node:
     """
     if isinstance(obj, list):
         return [load(i) for i in obj]
-    if isinstance(obj, dict):
+    if isinstance(obj, dict) and "class" in obj:
```
tobymao, line 48: is this the first time we store a dict? may want to think about this

tests/dialects/test_dialect.py
```diff
@@ -1097,32 +1097,51 @@ def test_order_by(self):
...
+                "bigquery": "JSON_EXTRACT(x, '$.y')",
+                "duckdb": "x -> 'y'",
+                "doris": "x -> '$.y'",
+                "mysql": "JSON_EXTRACT(x, '$.y')",
                 "postgres": "x->'y'",
-                "presto": "JSON_EXTRACT(x, 'y')",
-                "starrocks": "x -> 'y'",
-                "doris": "x -> 'y'",
+                "presto": "JSON_EXTRACT(x, '$.y')",
+                "sqlite": "x -> '$.y'",
```
tobymao, line 1108: is this correct?

tests/dialects/test_dialect.py
```diff
@@ -1097,32 +1097,51 @@ def test_order_by(self):
...
+                "presto": "JSON_EXTRACT(x, '$.y')",
+                "spark": "GET_JSON_OBJECT(x, '$.y')",
+                "sqlite": "x -> '$.y'",
+                "starrocks": "x -> '$.y'",
+                "tsql": "ISNULL(JSON_QUERY(x, '$.y'), JSON_VALUE(x, '$.y'))",
             },
         )
         self.validate_all(
-            "JSON_EXTRACT_SCALAR(x, 'y')",
+            "JSON_EXTRACT_SCALAR(x, '$.y')",
```
tobymao, line 1126: can you add a test with nested paths

'$.y[0].z'

sqlglot/serde.py
georgesittas: that's correct, but cc @barakalon to double-check since iirc he made the initial implementation of serde

tests/dialects/test_dialect.py
georgesittas: which one?

tests/dialects/test_dialect.py
georgesittas: sure

tests/dialects/test_dialect.py
tobymao: sqlite, does it really take a true json path? or is it like postgres and just takes in 'y'

georgesittas:
FYI the only remaining dialect I need to refactor is Snowflake, but their path syntax is a bit weirder than the rest of the dialects that don't support the JSON path syntax as-is. DuckDB doesn't support the full spec of JSON path too and they seem to have augmented the path syntax with their own, for example the `#` here is not in the official spec:

```sql
SELECT json_extract('{"duck": [1, 2, 3]}', '$.duck[#-1]');
```

Refer to the linked docs for more info. You can also have JSON Pointer syntax for DuckDB paths which will be left as-is.

tests/dialects/test_dialect.py
georgesittas: they do, most of the stuff I wrote, I tested in the corresponding engines and double-checked with docs

georgesittas: check out https://www.sqlite.org/json1.html#jex

sqlglot/dialects/dialect.py
georgesittas: i see what you mean, but just to make sure: your concern is what happens when we generate sql for dialects that don't support the json path syntax and we need to split these paths into multiple extractions?

sqlglot/serde.py
georgesittas: while we're at this topic, please note the changes to `_norm_arg` and `hashable_args` too

sqlglot/expressions.py
```diff
@@ -5466,6 +5473,13 @@ class NextValueFor(Func):
...
 
 def _norm_arg(arg):
+    arg_type = type(arg)
+
+    if arg_type is list:
+        return tuple(_norm_arg(a) for a in arg)
+    if arg_type is dict:
+        return frozenset((k, _norm_arg(v)) for k, v in arg.items())
+
     return arg.lower() if type(arg) is str else arg
```
tobymao, line 5483: use arg_type here?

sqlglot/dialects/dialect.py
tobymao: well for postgres, it would be json_extract_path(x, 'a', 'b', 'c')

sqlglot/dialects/dialect.py
```diff
@@ -1019,3 +1024,45 @@ def merge_without_target_sql(self: Generator, expression: exp.Merge) -> str:
...
+    supports_null_if_invalid: bool = False,
+) -> t.Callable[[t.List], E]:
+    def _parse_json_extract_path(args: t.List) -> E:
+        null_if_invalid = None
+
+        segments: t.List[exp.JSONPathPart] = [exp.JSONPathRoot()]
+        for arg in args[1:]:
+            if isinstance(arg, exp.Literal):
+                text = arg.name
+                if text.isnumeric():
```
tobymao, line 1040: let's factor out the is_int from expression.py is_number and use that here

sqlglot/dialects/dialect.py
```diff
@@ -1019,3 +1024,45 @@ def merge_without_target_sql(self: Generator, expression: exp.Merge) -> str:
...
+                    segments.append(exp.JSONPathSubscript(this=int(text)))
+                else:
+                    segments.append(exp.JSONPathChild(this=text))
+            elif supports_null_if_invalid:
+                null_if_invalid = arg
```
tobymao, line 1045: should this be True? why are you using arg

sqlglot/dialects/postgres.py
```diff
@@ -375,9 +389,16 @@ class Generator(generator.Generator):
...
         SUPPORTS_SELECT_INTO = True
-        # https://www.postgresql.org/docs/current/sql-createtable.html
+        JSON_TYPE_REQUIRED_FOR_EXTRACTION = True
         SUPPORTS_UNLOGGED_TABLES = True
 
+        JSON_PATH_MAPPING = {
+            exp.JSONPathChild: lambda n: n.name,
+            exp.JSONPathKey: lambda n: n.name,
+            exp.JSONPathRoot: lambda _: "",
+            exp.JSONPathSubscript: lambda n: generator.generate_json_path([n.this]),
```
tobymao, line 399: what happens when you get a foreign node? do you handle warnings?

sqlglot/dialects/sqlite.py
```diff
@@ -62,6 +62,21 @@ def _transform_create(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def _parse_json_extract(args: t.List) -> exp.JSONExtract:
+    this, *paths = args
+    path = parser.parse_json_path(seq_get(paths, 0))
+    if isinstance(path, exp.JSONPath):
+        for p in paths[1:]:
+            parsed = parser.parse_json_path(p)
```
tobymao, line 70: what's up with this sub parser?

georgesittas: Sqlite supports multiple json paths in `JSON_EXTRACT`, i.e. `JSON_EXTRACT(x, '$.x', '$.y', ...)`. This parses each one of them and then merges them into a single AST node. The `paths[1:]` slice is taken so that we skip the root node in all subsequent paths.

sqlglot/dialects/postgres.py
georgesittas: yep, check out this section: https://github.com/tobymao/sqlglot/pull/2883/files#diff-c779839b81bb6254e97ad8a6e17e9efbb09031d113a4942040fdd2cc3dcd9275R223, as well as https://github.com/tobymao/sqlglot/pull/2883/files#diff-63eb8dd82d3561bc414e3e13cf59516bd6584c5766a10072157a8a5aee4ad85fR71-R72

sqlglot/dialects/sqlite.py
tobymao: interesting

sqlglot/dialects/dialect.py
georgesittas: The idea is that `arg` should be the last element here [in redshift](https://docs.aws.amazon.com/redshift/latest/dg/JSON_EXTRACT_PATH_TEXT.html), and it should be the `null_if_invalid` arg since it seems like the `path_elems` need to be strings for it to accept the function call.

TL;DR to detect the flag when we don't find a `Literal`, and that should be the last arg for Redshift.

sqlglot/dialects/dialect.py
georgesittas: ah maybe we can just do `arg.is_int` here, will refactor it

sqlglot/dialects/sqlite.py
georgesittas: I'll double-check this section because it may be incorrect, noticed a similar weirdness with [MySQL](https://github.com/tobymao/sqlglot/pull/2883/files#diff-14ef9d8b99f48dcf8f6a4e3bbab7bf55a9d85d2b83d684c3c822f0f94fcd1d07R152-R155) and refactored what I initially implemented there too.

sqlglot/dialects/postgres.py
tobymao: shouldn't you call raise_error so that the error handling gets respected?

sqlglot/dialects/dialect.py
tobymao: well in some engines it can be an int, others it's '1'

sqlglot/dialects/postgres.py
georgesittas: For cases like duckdb that may be too aggressive by default, because you can have a json path but you can also have a json pointer path and so you don't really wanna hard fail there. Maybe the warning is a bit of a safer choice to start with, what do you think?

tobymao: there's nothing worse than producing the incorrect result. that's the ultimate sin in data. so if they want to ignore warnings and produce bad sql, they should opt in with errror_level ignore

sqlglot/dialects/dialect.py
georgesittas: Can you elaborate on your initial suggestion here a bit? `is_number` is true for literals with the `is_string` arg set to false and `is_int` relies on that logic. Here we inspect the actual text because of what you said in the last reply; are you suggesting we move this as a new property under `Expression` or something?

sqlglot/dialects/sqlite.py
georgesittas: I got rid of this after all because SQLite and MySQL semantics match when there are >1 paths, see https://github.com/tobymao/sqlglot/pull/2883/commits/9d5f9f5a188fb904834e5a12a8cbe09990f60827. Thanks for calling this out!

sqlglot/dialects/dialect.py
tobymao: i'm suggesting you factor out the try catch int check, and call that here on text instead of isnumeric

tests/dialects/test_dialect.py
georgesittas: check out https://github.com/tobymao/sqlglot/pull/2883/commits/91f073add8cf12bfb0f544dc090ecbbd83a3ded8

sqlglot/generator.py
```diff
@@ -2367,6 +2411,15 @@ def matchagainst_sql(self, expression: exp.MatchAgainst) -> str:
     def jsonkeyvalue_sql(self, expression: exp.JSONKeyValue) -> str:
         return f"{self.sql(expression, 'this')}{self.JSON_KEY_VALUE_PAIR_SEP} {self.sql(expression, 'expression')}"
 
+    def jsonpath_sql(self, expression: exp.JSONPath) -> str:
+        path = generate_json_path(
+            expression.expressions,
+            mapping=self._JSON_PATH_MAPPING,
+            unsupported_callback=self.unsupported,
+        )
+        path = path[1:] if path.startswith(".") else path
```
tobymao, line 2420: when would path start with .? you could just do lstrip(".")

tobymao (approved):

sqlglot/dialects/dialect.py
georgesittas: something like this then? https://github.com/tobymao/sqlglot/pull/2883/commits/a52a54ab3dc1fc1bc164050e36561bb1657d1621

tobymao: yea, exactly but put it in helpers

sqlglot/generator.py
georgesittas: For example when you parse `SELECT JSON('{"fruit":"banana"}') -> 'fruit'` in duckdb and you wanna generate that in snowflake, you'd get a prefix `.` without a dollar sign (since the root is not generated)

PR #2886, opened 2024-01-25
fix: type imports

tobymao:


sqlglot/dialects/dialect.py
```diff
@@ -14,11 +13,14 @@
...
 
-B = t.TypeVar("B", bound=exp.Binary)
-
 DATE_ADD_OR_DIFF = t.Union[exp.DateAdd, exp.TsOrDsAdd, exp.DateDiff, exp.TsOrDsDiff]
 DATE_ADD_OR_SUB = t.Union[exp.DateAdd, exp.TsOrDsAdd, exp.DateSub]
 
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
+
+    B = t.TypeVar("B", bound=exp.Binary)
```
georgesittas, line 22: may be worth putting this in _typing.py too, we define it in annotate_types.py as well

georgesittas (approved):

fattybenji:
Hi, seems like this PR breaks the import for the lib. If I just install version 2.10 and import sqlglot I get `ModuleNotFoundError: No module named 'typing_extensions'`. Maybe typing_extensions should be in `install_requires` ?

Should I create an issue ?

PR #2904, opened 2024-02-01
Feat(redshift): parse GETDATE

erickpeirson:
This adds `GETDATE` to the parser for Redshift.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -71,6 +71,7 @@ class Redshift(Postgres):
             "DATE_ADD": _parse_date_delta(exp.TsOrDsAdd),
             "DATEDIFF": _parse_date_delta(exp.TsOrDsDiff),
             "DATE_DIFF": _parse_date_delta(exp.TsOrDsDiff),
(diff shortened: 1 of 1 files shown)


georgesittas (approved):
Looks good but let's test this as well. Thanks :)

erickpeirson:
> Looks good but let's test this as well. Thanks :)

@georgesittas Adding some tests now. Quick question: I noticed that `exp.CurrentTimestamp` renders as `SYSDATE` for Redshift. There is a subtle difference between `GETDATE` and `SYSDATE` in that the former returns the timestamp at the start of the transaction (if there is one), whereas `SYSDATE` does not. Many other dialects have isomorphic functions. 

I don't want to get too ambitious, but I'm wondering whether we should better distinguish between these with distinct expression types? E.g. `CurrentTimestamp` versus `CurrentSystemTimestamp`?

erickpeirson:
I'll take https://github.com/tobymao/sqlglot/pull/2909 as my answer :-)

georgesittas:
Hah!

Issue #2907, opened 2024-02-02
bug: `unnest_sql` hits errors when the alias is the 'Identifier' object and it has no attribute 'columns'


chelsea-lin:
**Fully reproducible code snippet**
This bug is minor and can be fixed by checking the `alias` object has the `columns` attribute before calling it. When the `alias` is a `sqlglot.expressions.Identifier`, the exception would be triggered. Here is the call stacks:

<img width="1171" alt="Screenshot 2024-02-01 at 4 17 12 PM" src="https://github.com/tobymao/sqlglot/assets/124939984/23fd4037-dff6-402c-a457-6ff0a72b75db">

tobymao:
i don't believe alias can be an identifier by natural means (using the parser). so i'm going to close this for now. if you can show me a test case where unnest alias is an identifier, we should fix that instead.

chelsea-lin:
Hi @tobymao, thank you to look at the issue. 
I'm using `sqlglot` within [ibis](https://github.com/ibis-project/ibis) to generate BigQuery SQL with a `UNNEST` node. Below are the mininum codes to repro the issue within `sqlglot` only. I think I might be creating an `UNNEST` object incorrectly. Any insights would be greatly appreciated!

```
In [23]: unnest=sg.parse_one("UNNEST(array_types.x)")
    ...: alias = sg.to_identifier("t0")
    ...: 

In [24]: sg.expressions.Unnest(expressions=[unnest], offset=None).as_(alias).sql()
Out[24]: 'UNNEST(UNNEST(array_types.x)) AS t0'

In [25]: sg.expressions.Unnest(expressions=[unnest], offset=None).as_(alias).sql(dialect="bigquery")
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[25], line 1
----> 1 sg.expressions.Unnest(expressions=[unnest], offset=None).as_(alias).sql(dialect="bigquery")

File ~/src/forkibis/venv/lib/python3.11/site-packages/sqlglot/expressions.py:528, in Expression.sql(self, dialect, **opts)
    516 """
    517 Returns SQL string representation of this tree.
    518 
   (...)
    524     The SQL string.
    525 """
    526 from sqlglot.dialects import Dialect
--> 528 return Dialect.get_or_raise(dialect).generate(self, **opts)

[...]

PR #2908, opened 2024-02-02
fix: check 'alias' has a attribute 'columns' in the unnest_sql function

chelsea-lin:
Fixing #2907


tobymao:
can you add a test case where this is true? this fix doesn't seem correct to me since alias should always be an Alias expr.

tobymao:
happy to reopen this with a more clear picture of the use case you're imagining. but reasoning was given in #2907

PR #2909, opened 2024-02-02
Fix!(redshift): properly distinguish SYSDATE and GETDATE()

georgesittas:
After merging https://github.com/tobymao/sqlglot/commit/3fa92cac285cbb2bd9d8b5724dadb77be7e12731, I realized that there's a subtle difference between `SYSDATE` and `GETDATE()`. From Redshift's docs:

> SYSDATE returns the start date and time for the current transaction, not for the start of the current statement.

> GETDATE returns the current date and time in the current session time zone (UTC by default). It returns the start date or time of the current statement, even when it is within a transaction block.

References:
- https://docs.aws.amazon.com/redshift/latest/dg/r_SYSDATE.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_GETDATE.html


sqlglot/generator.py
```diff
@@ -2999,7 +2999,9 @@ def func(
         return f"{self.normalize_func(name)}{prefix}{self.format_args(*args)}{suffix}"
 
     def format_args(self, *args: t.Optional[str | exp.Expression]) -> str:
-        arg_sqls = tuple(self.sql(arg) for arg in args if arg is not None)
+        arg_sqls = tuple(
+            self.sql(arg) for arg in args if arg is not None and not isinstance(arg, bool)
+        )
```
georgesittas, line 3004: This seemed something that could be helpful in general: we usually fail in `sql` when a `bool` is encountered which makes it annoying to add boolean flags in `Func`s, since we would then need to override stuff / handle the bools explicitly. This should allow funcs to have bool args that function as switches, like `transaction` here. Open to other ideas as well.

erickpeirson:
@georgesittas It looks like we may need this distinction in Oracle, too. Probably elsewhere

georgesittas:
> @georgesittas It looks like we may need this distinction in Oracle, too. Probably elsewhere

Sure, can you check what other dialects do this so we can fix them too?

sqlglot/generator.py
tobymao: doesn't this indicate that something is wrong? shouldn't we have two different expressions?

tobymao: because in this case, sqlglot dialect will immediate lose context of the bool arg

georgesittas: The idea was that both `SYSDATE` and `GETDATE()` encode the same functionality, and whether or not they capture the timestamp of the transaction vs the statement is just a property of the expression, hence why I didn't make a new node - it seemed redundant. So in this context, this is one example of needing a bool flag.

I feel like I've had the issue with the bools in the past, i.e. the generator would raise because there was a flag in some arg of an expression, but I don't have other concrete examples atm. Happy to switch into defining a new expression for this if it makes more sense.

georgesittas: chatted in slack - will update to make the behavior more explicit

sqlglot/generator.py
```diff
@@ -2676,6 +2676,9 @@ def currentdate_sql(self, expression: exp.CurrentDate) -> str:
         zone = self.sql(expression, "this")
         return f"CURRENT_DATE({zone})" if zone else "CURRENT_DATE"
 
+    def currenttimestamp_sql(self, expression: exp.CurrentTimestamp) -> str:
+        return f"CURRENT_TIMESTAMP({self.sql(expression, 'this')})"
```
tobymao, line 2680: self.func("CURRENT_TIMESTAMP", expression.this) ?

PR #2911, opened 2024-02-02
Fix(duckdb): fix JSON pointer path parsing, reduce warning noise

georgesittas:
This addresses two issues:

1. Right now, we parse DuckDB's JSON pointer paths incorrectly, and as a result we also clutter them at generation time:

```python
>>> import sqlglot
>>> print(sqlglot.transpile("""SELECT json_extract('{"duck": [1, 2, 3]}', '/duck/0');""", "duckdb")[0])
SELECT '{"duck": [1, 2, 3]}' -> '$["/duck/0"]'
```

2. DuckDB supports a variant of the [JSONPath](https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html) syntax spec, where one can access elements from the back of an array using the following syntax:

```sql
SELECT json_extract('{"duck": [1, 2, 3]}', '$.duck[#-1]');
-- 3
```

We currently emit a warning when encountering paths like this which can be noisy, since DuckDB users will probably expect this to _parse normally_.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -444,7 +444,7 @@ class Dialect(metaclass=_Dialect):
 
         return expression
 
-    def parse_json_path(self, path: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
+    def to_json_path(self, path: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
         if isinstance(path, exp.Literal):
             path_text = path.name
             if path.is_number:
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import typing as t
 from enum import Enum, auto
 from functools import reduce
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -441,6 +444,19 @@ def quote_identifier(self, expression: E, identify: bool = True) -> E:
 
         return expression
 
+    def parse_json_path(self, path: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
```
tobymao, line 447: why is this done in the dialect and not just in the parser?

sqlglot/parser.py
```diff
@@ -558,12 +546,12 @@ class Parser(metaclass=_Parser):
         TokenType.ARROW: lambda self, this, path: self.expression(
             exp.JSONExtract,
             this=this,
-            expression=parse_json_path(path),
+            expression=self.dialect.parse_json_path(path),
```
tobymao, line 549: why don't we just call self.parse_json_path?

sqlglot/dialects/dialect.py
georgesittas: Because this is currently used in `sqlglot.parser.parse_extract_json_with_path`, which is basically a helper to dry up `FUNCTIONS` entries. That lambda doesn't have access to the parser instance (*), so we'd have to move ~7 such entries in `FUNCTION_PARSERS` instead; iirc we want to avoid that.

Maybe the `parse` prefix in this method is a misnomer, because we are not really "parsing" anything, we're just constructing a new expression out of a given one. At least that's how I think about it, so it didn't seem that bad to have it in the dialect.

Thoughts?

sqlglot/parser.py
georgesittas: Let's discuss this in https://github.com/tobymao/sqlglot/pull/2911#discussion_r1476873600 since they're related

sqlglot/dialects/dialect.py
tobymao: maybe to_json_path

tobymao (approved):

PR #2915, opened 2024-02-05
Add parse support for `BPCHAR`

erickpeirson:
This adds parsing support for `BPCHAR`, which is used in [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_Character_types.html) and [Postgres](https://www.postgresql.org/docs/current/datatype-character.html).


tests/test_parser.py
```diff
@@ -99,6 +99,11 @@ def test_structs(self):
     def test_float(self):
         self.assertEqual(parse_one(".2"), parse_one("0.2"))
 
+    def test_bpchar(self):
+        cast = parse_one("cast(foo as bpchar)")
```
tobymao, line 103: can you delete this test and just put it into identity.sql as a fixture?

erickpeirson: Sure thing; done

PR #2920, opened 2024-02-06
Refactor: move JSON path generation logic in Generator

georgesittas:
Addresses items (1) and (2) mentioned in the "to do" section of https://github.com/tobymao/sqlglot/pull/2914.

Transpiling `*` is a little more nuanced and requires more investigation, because despite some dialects not supporting this syntax, I think it may still be possible to simulate it by simply omitting it. For example, in DuckDB you can do:

```
D SELECT '["a", "b", "c"]' -> '$[*]';
┌───────────────────────────────┐
│ ('["a", "b", "c"]' -> '$[*]') │
│            json[]             │
├───────────────────────────────┤
│ ["a", "b", "c"]               │
└───────────────────────────────┘
```

But in BigQuery this syntax is not supported, since they don't recognize the `*` inside the brackets ("Unsupported operator in JSONPath: *"). However, if we completely omit the brackets, it seems like the result matches:

```
> SELECT JSON_EXTRACT('["a", "b", "c"]', '$');
f0_
["a","b","c"]
```

The exact same thing holds e.g. for Trino. Not sure if we wanna solve this now, though.


sqlglot/generator.py
```diff
@@ -2426,13 +2405,29 @@ def jsonkeyvalue_sql(self, expression: exp.JSONKeyValue) -> str:
...
+
+        if isinstance(expression, int):
+            return str(expression)
+
+        if self.JSON_PATH_SINGLE_QUOTE_ESCAPE:
+            escaped = expression.replace("'", "\\'")
+            escaped = f"'{escaped}'"
+        else:
+            escaped = expression.replace('"', '\\"')
+            escaped = f'"{escaped}"'
```
georgesittas, line 2428: Some context surrounding this section... as mentioned in the previous PR, BQ and Spark escape "unsafe" keys using single quotes, i.e. `"['key with spaces']"`.

I thought of two ways to take this into account in the generator:

- escape the single quotes in line 2425, i.e. `escaped = f"\\'{escaped}\\'"`, in order to leave line 2435 untouched, but that felt wrong because it'd be assuming that `QUOTE_START` and `QUOTE_END` are single quotes.
- change the string quotes being used for the json path string literal, which is what I implemented here; I liked this approach more because it makes the two branches here symmetrical and it feels easier to understand.

Not sure if this different escaping behavior in BQ / Spark can be implemented in a cleaner fashion.. FWIW, I'm kinda ok with how this turned out here, @tobymao any thoughts?

tobymao: why can't we always make quote start the same? that's what we do for regular sql isn't it? so always do single quote to normalize it

sqlglot/jsonpath.py
```diff
@@ -142,111 +143,55 @@ def _parse_bracket() -> exp.JSONPathPart:
...
+    return exp.JSONPath(expressions=expressions)
 
 
-def _generate_json_path_key(node: exp.JSONPathKey, **kwargs) -> str:
-    this = node.this
-    if this == "*" or exp.SAFE_IDENTIFIER_RE.match(this):
-        return f".{this}"
+def generate(path: exp.JSONPath, dialect: DialectType = None) -> str:
+    """Generates a string from a JSONPath expression, optionally targetting a given SQL dialect."""
+    return path.sql(dialect=dialect)[1:-1]  # The slice is used to remove the string delimiters
```
tobymao, line 176: this isn't used anywhere right? this makes a copy of the expression

sqlglot/generator.py
tobymao: not really a big fan of this implementation, i'd rather just have one quoting style if possible

sqlglot/jsonpath.py
georgesittas: May need to refactor this one yeah, it's only used in the test suite for json paths. Previously we'd generate an unquoted path and then later stringify it in the generator, but with this approach the quotes are added first and the tests fail because they weren't expected in cts.json

sqlglot/generator.py
georgesittas: > why can't we always make quote start the same? that's what we do for regular sql isn't it? so always do single quote to normalize it

We can do that if we escape the single quotes I think, see first bullet. How does that look? So for BQ and Spark you'd have `'$[\'key with spaces \']'`.

tobymao: so sqlglot always generates quotse with single quote ', can we default to double quotes for keys?

georgesittas: what's an example of that? just in case this is what you're suggesting, note that this is invalid BQ and will return `NULL` in Spark:

```
'$["key with spaces"]'
```

Moreover, if we did `"$['key with spaces']"` which is basically reverting the quotes, it wouldn't be the same in dialects that use double quotes for identifiers.

tobymao: ok, so if '$["key with spaces"]' is illegal, then just use single quotes and  back slashes as you suggested

georgesittas: done https://github.com/tobymao/sqlglot/pull/2920/commits/9c81272791ae57d48262123e051c4e99dd5de01a

sqlglot/jsonpath.py
georgesittas: Removed in https://github.com/tobymao/sqlglot/pull/2920/commits/d29aeb70b2497a9562e36492cc7a29f049214cb7

georgesittas: Removed in https://github.com/tobymao/sqlglot/pull/2920/commits/d29aeb70b2497a9562e36492cc7a29f049214cb7

tobymao (approved):

PR #2921, opened 2024-02-06
fix: have table exclude this if schema target

eakmanrq:
Prior to this change table objects which were the target of `CREATE SCHEMA` or `DROP SCHEMA` would set this as the db name and then db as the catalog name (if provided). The rendered SQL is fine but if we call something like `qualify_table` on the expression and try to set a catalog if one does not exist then it will add the catalog when it is not needed since it will see the catalog filed as None when really catalog is set but it is just set in the `db` field.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1352,7 +1352,9 @@ class Parser(metaclass=_Parser):
             exp.Drop,
             comments=start.comments,
             exists=exists or self._parse_exists(),
-            this=self._parse_table(schema=True),
+            this=self._parse_table(
+                schema=True, is_db_reference=self._prev.token_type == TokenType.SCHEMA
+            ),
             kind=kind,
             temporary=temporary,
             materialized=materialized,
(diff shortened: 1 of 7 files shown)


sqlglot/parser.py
```diff
@@ -1352,7 +1352,7 @@ def _parse_drop(self, exists: bool = False) -> exp.Drop | exp.Command:
             exp.Drop,
             comments=start.comments,
             exists=exists or self._parse_exists(),
-            this=self._parse_table(schema=True),
+            this=self._parse_table(schema=True, is_db_reference=kind == "SCHEMA"),
```
tobymao, line 1355: what happens if schema is not capitalized?

tests/dialects/test_bigquery.py
```diff
@@ -1272,3 +1273,19 @@ def test_json_object(self):
 
         with self.assertRaises(ParseError):
             transpile("SELECT JSON_OBJECT('a', 1, 'b') AS json_data", read="bigquery")
+
+    def test_parse_drop_schema(self):
```
tobymao, line 1277: let's move all of these tests into test_parsers. they are very repetitive.

you can do a for look

for dialect in snowflake, bigquery, ...
 

additionally, don't really need to breakdown and test the ast like, just check the ast equality

ast == exp.Drop(...)

sqlglot/parser.py
tobymao: why not just check the token type like you do in create?

georgesittas (commented):
Looks good, agree with Toby's comments

sqlglot/parser.py
eakmanrq: Updated: https://github.com/tobymao/sqlglot/blob/76999173554117381a15884698fcbd4126e65125/sqlglot/parser.py#L1355-L1357

PR #2923, opened 2024-02-06
Feat: improve transpilation of CREATE TABLE LIKE statement

### LINKED ISSUE #2922: Transpiling a CREATE TABLE ... LIKE ... statement to DuckDB does not use the existing DuckDB syntax
```python
import sqlglot

sqlglot.transpile("CREATE TABLE t2 LIKE t1", read="snowflake", write="duckdb")
```

**Official Documentation**
https://duckdb.org/docs/sql/statements/create_table.html

While DuckDB does not support the LIKE property, but the same can be expressed with:
```SQL
CREATE TABLE t2 AS FROM t1 LIMIT 0;
```

georgesittas:
Fixes #2922

We can probably transfer this logic to the base generator and control it with a flag, so other dialects besides duckdb can leverage it. Can open a separate PR for it later or do it here, just need a bit more time to investigate.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1293,7 +1293,7 @@ class Generator(metaclass=_Generator):
         if expression.expressions:
             self.unsupported("Transpilation of LIKE property options is unsupported")
 
-        select = exp.select("*").from_(self.sql(expression, "this")).limit(0)
+        select = exp.select("*").from_(expression.this).limit(0)
         return f"AS {self.sql(select)}"
 
     def fallbackproperty_sql(self, expression: exp.FallbackProperty) -> str:
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -303,6 +303,12 @@ class Generator(metaclass=_Generator):
     # Whether or not UNLOGGED tables can be created
     SUPPORTS_UNLOGGED_TABLES = False
 
+    # Whether or not the CREATE TABLE LIKE statement is supported
+    SUPPORTS_CREATE_TABLE_LIKE = True
+
+    # Whether or not the LikeProperty needs to be specified inside of the schema clause
+    LIKE_PROPERTY_INSIDE_SCHEMA = False
+
     # Whether or not the JSON extraction operators expect a value of type JSON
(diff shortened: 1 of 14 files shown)


tobymao (approved):
a flag would be nice

georgesittas:
yep doing that work now

georgesittas:
Improved the transpilation as discussed above.

References:
- https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_like
- https://clickhouse.com/docs/en/sql-reference/statements/create/table
- https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-like.html
- https://doris.apache.org/docs/dev/sql-manual/sql-reference/Data-Definition-Statements/Create/CREATE-TABLE-LIKE/
- https://dev.mysql.com/doc/refman/8.0/en/create-table-like.html
- https://docs.oracle.com/cd/E17952_01/mysql-8.0-en/create-table-like.html
- https://www.postgresql.org/docs/current/sql-createtable.html
- http://prestodb.io/docs/0.285/sql/create-table.html
- https://trino.io/docs/current/sql/create-table.html
- https://stackoverflow.com/questions/616104/what-is-the-equivalent-of-create-table-like-in-sql-server
- https://www.sqlite.org/lang_createtable.html

sqlglot/generator.py
```diff
@@ -1274,9 +1280,21 @@ def property_sql(self, expression: exp.Property) -> str:
...
+            like = f"LIKE {self.sql(expression, 'this')}{options}"
+            if self.LIKE_PROPERTY_INSIDE_SCHEMA and not isinstance(expression.parent, exp.Schema):
+                like = f"({like})"
+
+            return like
+
+        if expression.expressions:
+            self.unsupported("Transpilation of LIKE property options is unsupported")
+
+        select = exp.select("*").from_(self.sql(expression, "this")).limit(0)
```
tobymao, line 1296: why do you call sql on the from_? can you just pass it in?

tobymao (approved):

sqlglot/generator.py
georgesittas: ah missed it, thanks

Issue #2051, opened 2023-08-14
Support for Clickhouse JSON Extract* Functions


BTheunissen:
**Is your feature request related to a problem? Please describe.**
Clickhouse supports a series of JSON string member extraction functions as documented [here](https://clickhouse.com/docs/en/sql-reference/functions/json-functions#jsonextractstringjson-indices_or_keys).

It would be helpful for these to be implemented in SQLGlot in a manner which will allow transpilation of Clickhouse queries into other SQL languages that also support JSON manipulation functions, in my current use-case these functions are used quite often for parsing columns containing unstructured data.

**Describe the solution you'd like**
Support `JSONExtractString` function in Clickhouse dialect (and other appropriate JSON functions supported by Clickhouse that make sense to support in SQLGlot).

GeorgeSittas:
Hey, thanks for the report.

I'll take a look at these JSON functions soon, but I don't think it's currently in scope to add support for all of them. I can add support for `JSONExtractString` and if you want you can use that as an example to implement the rest.

BTheunissen:
I'd be happy to! Of the functions I believe `JSONExtractString` is probably the lowest hanging fruit that is used the most broadly.

GeorgeSittas:
@BTheunissen I'm a bit skeptical about mapping `JSONExtractString` into an expression type: this function is used a bit differently in ClickHouse, compared to the equivalent tools in other dialects, so it doesn't seem obvious that doing this will help with the transpilation much.

For instance, the Postgres `->>` [operator](https://www.postgresql.org/docs/9.5/functions-json.html) is used on `JSON`-typed values, whereas the above function expects a string argument. Also, other dialects have different formatting conventions for extracting JSON fields, e.g. we have a test like `stream_data ->> '$.data.results'` for MySQL.

Perhaps it would help if you shared a bit more context on what you want to transpile it to, but I'm not sure if the ROI of adding this will be worth it.

BTheunissen:
I think I was misinterpreting the expression type usage in the codebase as it currently stands. I was hoping that as JSON functions are generally supported across most SQL dialects, we should be able to find a common form of "get an element from this JSON document", and then identify those functions as being interchangeable between SQL dialects, e.g. `JSONExtractString` in a Clickhouse query could be mapped into `JSON_EXTRACT_STRING` in DuckDB, etc.

If mapping functions such as these between dialects is currently not in scope within the project then that is totally understandable.

GeorgeSittas:
[...]

PR #2925, opened 2024-02-07
Feat(clickhouse): add support for JSONExtractString, clean up some helpers

georgesittas:
Several changes here, not necessarily related to each other:

- Improve parsing / transpilation of clickhouse's `JSONExtractString`, fixes https://github.com/tobymao/sqlglot/issues/2051. Note that clickhouse uses 1-based indexing, so I had to slightly change the logic in some parts to take this into account. Perhaps and interesting observation here is that the dialect offset is not necessarily tied to the JSON path offset.
- Refactored some helpers that concern JSON path parsing and generation and are used by clickhouse, postgres, redshift.
- Fixed some issues related to columns being used in json paths (supported by clickhouse, duckdb, postgres). This involved a change that may need some attention in the parser (I split `_parse_column` into two methods, can elaborate if needed).
- Finally, fixed an issue with multi-arg `JSON_EXTRACT` calls. It turns out that I had a bug in my implementation and so previously we'd support up to 3 arguments because of the `null_if_invalid` flag, but 4 would fail. I extended both `JSONExtract` and `JSONExtractScalar` to accept arbitrarily many args to amend this.

References:
- https://clickhouse.com/docs/en/sql-reference/functions/json-functions
- https://docs.aws.amazon.com/redshift/latest/dg/JSON_EXTRACT_PATH_TEXT.html

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -21,6 +21,9 @@ DATE_ADD_OR_SUB = t.Union[exp.DateAdd, exp.TsOrDsAdd, exp.DateSub]
 if t.TYPE_CHECKING:
     from sqlglot._typing import B, E
 
+    F = t.TypeVar("F", bound="exp.Func")
+    JSON_EXTRACT_TYPE = t.Union[exp.JSONExtract, exp.JSONExtractScalar]
+
 logger = logging.getLogger("sqlglot")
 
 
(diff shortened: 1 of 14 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -21,6 +21,9 @@
 if t.TYPE_CHECKING:
     from sqlglot._typing import B, E
 
+    F = t.TypeVar("F", bound="exp.Func")
```
tobymao, line 24: let's put this in typing

tobymao (approved):

PR #2926, opened 2024-02-07
Feat(clickhouse): add support for LIMIT BY clause

georgesittas:
Addresses https://tobiko-data.slack.com/archives/C0448SFS3PF/p1707220020280219.

Based on https://clickhouse.com/docs/en/sql-reference/statements/select/limit-by, it seems like we can also have

```sql
LIMIT n OFFSET offset_value BY expressions
```

This PR doesn't handle this version properly, at least not yet. I wasn't sure how we wanna do this, provided that `OFFSET` is treated as a separate modifier. Should we just add an `expressions` arg to `exp.Offset` and do something similar in its parser / generator?

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1796,7 +1796,7 @@ class Lambda(Expression):
 
 
 class Limit(Expression):
-    arg_types = {"this": False, "expression": True, "offset": False}
+    arg_types = {"this": False, "expression": True, "offset": False, "expressions": False}
 
 
 class Literal(Condition):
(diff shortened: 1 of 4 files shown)


tobymao (approved):

georgesittas:
> Should we just add an expressions arg to exp.Offset and do something similar in its parser / generator

Ended up implementing this, also fixed the `LIMIT offset, n BY expressions` case which was generated erroneously. See https://github.com/tobymao/sqlglot/pull/2926/commits/b0a7b67de18bd1a51f9de05fef86fde70c76919a

PR #2934, opened 2024-02-08
Fix: remove find method from Schema

georgesittas:
Chatted with Barak and realized this method imposes an additional constraint on Schema implementors which is probably unnecessary, so I decided to revert the change and only keep it under `AbstractMappingSchema`.

Context behind the addition of this method: https://github.com/tobymao/sqlglot/pull/2869


barakalon (approved):
thanks!

PR #2935, opened 2024-02-08
Fix(postgres): fallback to parameter parser if heredoc is untokenizable

### LINKED ISSUE #2931: Parameter `$1` for `postgres` is not supported
In version `v18.8.0`, heredoc strings support was added, and `"$": TokenType.PARAMETER` was replaced with `"$": TokenType.HEREDOC_STRING`. Since then, sql query with a parameter like `$1` cannot be parsed.

Here is the original PR [Feat!: add support for heredoc strings (Postgres, ClickHouse) #2328](https://github.com/tobymao/sqlglot/pull/2328)

```
import sqlglot

schema = {"t": {"x": "text"}}
sql = "select x from t where $1::text = 'ok'"
expression = sqlglot.parse_one(sql, dialect="postgres")
print(expression.sql(dialect="postgres"))
qualified_expr = qualify(expression, schema=schema, dialect="postgres")
annotated_expr = annotate_types(qualified_expr, schema=schema)
print(annotated_expr.selects[0].type)
```
Exception was raised. Expected output is `TEXT`.
```
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/tokens.py", line 836, in tokenize
    self._scan()
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/tokens.py", line 859, in _scan
    self._scan_keywords()
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/tokens.py", line 994, in _scan_keywords
    if self._scan_string(word):
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/tokens.py", line 1136, in _scan_string
    tag = "" if self._char == end else self._extract_string(end)
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/tokens.py", line 1203, in _extract_string
    raise TokenError(f"Missing {delimiter} from {self._line}:{self._start}")
sqlglot.errors.TokenError: Missing $ from 1:22

The above exception was the direct cause of the following exception:

  File "/Users/eric/tmp/test.py", line 20, in <module>
    expression = sqlglot.parse_one(sql, dialect="postgres")
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/__init__.py", line 125, in parse_one
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/dialects/dialect.py", line 304, in parse
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/dialects/dialect.py", line 318, in tokenize
    return self.tokenizer.tokenize(sql)
  File "/Users/eric/venv/lib/python3.10/site-packages/sqlglot/tokens.py", line 841, in tokenize
    raise TokenError(f"Error tokenizing '{context}'") from e
sqlglot.errors.TokenError: Error tokenizing 'select x from t where $1::text = 'ok'
```

georgesittas:
Fixes #2931

Postgres requires heredoc tags to follow the same lexical rules as the unquoted identifiers, so a digit after the `HEREDOC_STRING` token (`$`) can unambiguously be assumed to be a parameter number.

Reference: https://www.postgresql.org/docs/current/sql-syntax-lexical.html

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -518,9 +518,7 @@ class _Tokenizer(type):
                 semicolon=_TOKEN_TYPE_TO_INDEX[TokenType.SEMICOLON],
                 string=_TOKEN_TYPE_TO_INDEX[TokenType.STRING],
                 var=_TOKEN_TYPE_TO_INDEX[TokenType.VAR],
-                heredoc_string_alternative=_TOKEN_TYPE_TO_INDEX[
-                    klass.HEREDOC_STRING_ALTERNATIVE or TokenType.VAR
-                ],
+                heredoc_string_alternative=_TOKEN_TYPE_TO_INDEX[klass.HEREDOC_STRING_ALTERNATIVE],
             )
             klass._RS_TOKENIZER = RsTokenizer(settings, token_types)
         else:
@@ -581,7 +579,7 @@ class Tokenizer(metaclass=_Tokenizer):
     HEREDOC_TAG_IS_IDENTIFIER = False
 
     # Token that we'll generate as a fallback if the heredoc prefix doesn't correspond to a heredoc
-    HEREDOC_STRING_ALTERNATIVE: t.Optional[TokenType] = None
+    HEREDOC_STRING_ALTERNATIVE = TokenType.VAR
 
     # Autofilled
     _COMMENTS: t.Dict[str, str] = {}
@@ -1264,7 +1262,7 @@ class Tokenizer(metaclass=_Tokenizer):
                     and not self._peek.isidentifier()
                     and not self._peek == end
                 ):
-                    if self.HEREDOC_STRING_ALTERNATIVE:
+                    if self.HEREDOC_STRING_ALTERNATIVE != token_type.VAR:
                         self._add(self.HEREDOC_STRING_ALTERNATIVE)
                     else:
                         self._scan_var()
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -504,6 +504,7 @@ class _Tokenizer(type):
                 command_prefix_tokens={
                     _TOKEN_TYPE_TO_INDEX[v] for v in klass.COMMAND_PREFIX_TOKENS
                 },
+                heredoc_tag_is_identifier=klass.HEREDOC_TAG_IS_IDENTIFIER,
             )
             token_types = RsTokenTypeSettings(
                 bit_string=_TOKEN_TYPE_TO_INDEX[TokenType.BIT_STRING],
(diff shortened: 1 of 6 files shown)


sqlglot/tokens.py
```diff
@@ -517,6 +518,9 @@ def _quotes_to_format(
                 semicolon=_TOKEN_TYPE_TO_INDEX[TokenType.SEMICOLON],
                 string=_TOKEN_TYPE_TO_INDEX[TokenType.STRING],
                 var=_TOKEN_TYPE_TO_INDEX[TokenType.VAR],
+                heredoc_string_alternative=_TOKEN_TYPE_TO_INDEX[
+                    klass.HEREDOC_STRING_ALTERNATIVE or TokenType.VAR
```
izeigerman, line 522:
```
or TokenType.VAR
```
I don't see the same logic being applied to the Python tokenizer. Is this expected?

sqlglotrs/src/tokenizer.rs
```diff
@@ -399,6 +399,19 @@ impl<'a> TokenizerState<'a> {
             } else if *token_type == self.token_types.bit_string {
                 (Some(2), *token_type, end.clone())
             } else if *token_type == self.token_types.heredoc_string {
+                if self.settings.heredoc_tag_is_identifier
+                    && !(self.peek_char.is_alphabetic() || self.peek_char == '_')
```
izeigerman, line 403:
```
|| self.peek_char == '_'
```
this is not in the Python version either.

georgesittas: That's right, the Python version uses `isidentifier()` which aims to replicate this logic. I searched tokenizer.rs and it seems like the other `isidentifier()` call is mirrored like what I did here, inside `scan_number()`. Should we factor this out in a helper with a similar name? Is there no rust equivalent for alpha + `_` checking?

sqlglot/tokens.py
izeigerman: Mirroring would be easier to maintain for sure. Less cognitive load

sqlglotrs/src/tokenizer.rs
izeigerman: > Should we factor this out in a helper with a similar name? Is there no rust equivalent for alpha + _ checking?

ah, makes sense. No, I didn't find one

sqlglot/tokens.py
tobymao: agree rust should try to mimic python implementation whenever possible

georgesittas:
Addressed feedback:

- Python side now mimics Rust side w.r.t. the `HEREDOC_STRING_ALTERNATIVE` logic
- Factored out the `is_alphabetic() ||  ... == '_'` logic so it resembles the Python side more.

PR #2937, opened 2024-02-08
fix(spark): CREATE TABLE ... PARTITIONED BY fixes

barakalon:
Spark 3 supports two CREATE TABLE formats:
https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-hiveformat.html
https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-datasource.html

Some of the spark dialect assumes the former, some assumes the latter, and you can easily end up with an invalid statement.

All this DDL transpilation feels very messy to me... I hated this change.

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -466,6 +466,90 @@ def unqualify_columns(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def remove_unique_constraints(expression: exp.Expression) -> exp.Expression:
+    assert isinstance(expression, exp.Create)
+    for constraint in expression.find_all(exp.UniqueColumnConstraint):
+        if constraint.parent:
+            constraint.parent.pop()
+
+    return expression
+
+
+def ctas_with_tmp_tables_to_create_tmp_view(
+    expression: exp.Expression,
+    tmp_storage_provider: t.Callable[[exp.Expression], exp.Expression] = lambda e: e,
+) -> exp.Expression:
+    assert isinstance(expression, exp.Create)
+    properties = expression.args.get("properties")
+    temporary = any(
+        isinstance(prop, exp.TemporaryProperty)
+        for prop in (properties.expressions if properties else [])
+    )
+
+    # CTAS with temp tables map to CREATE TEMPORARY VIEW
+    kind = expression.args["kind"]
+    if kind.upper() == "TABLE" and temporary:
+        if expression.expression:
+            return exp.Create(
+                kind="TEMPORARY VIEW",
+                this=expression.this,
+                expression=expression.expression,
+            )
+        else:
+            # CREATE TEMPORARY TABLE may require storage provider
+            return tmp_storage_provider(expression)
+
+    return expression
+
+
+def move_schema_columns_to_partitioned_by(expression: exp.Expression) -> exp.Expression:
+    """
+    In Hive, the PARTITIONED BY property acts as an extension of a table's schema. When the
+    PARTITIONED BY value is an array of column names, they are transformed into a schema.
+    The corresponding columns are removed from the create statement.
+    """
+    assert isinstance(expression, exp.Create)
+    has_schema = isinstance(expression.this, exp.Schema)
+    is_partitionable = expression.args.get("kind").upper() in {"TABLE", "VIEW"}
+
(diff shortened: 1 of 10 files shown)


tests/dialects/test_hive.py
```diff
@@ -152,7 +152,7 @@ def test_ddl(self):
                 "duckdb": "CREATE TABLE x (w TEXT)",  # Partition columns should exist in table
                 "presto": "CREATE TABLE x (w VARCHAR, y INTEGER, z INTEGER) WITH (PARTITIONED_BY=ARRAY['y', 'z'])",
                 "hive": "CREATE TABLE x (w STRING) PARTITIONED BY (y INT, z INT)",
-                "spark": "CREATE TABLE x (w STRING) PARTITIONED BY (y INT, z INT)",
+                "spark": "CREATE TABLE x (w STRING, y INT, z INT) PARTITIONED BY (y, z)",
```
tobymao, line 155: what about for spark 2? i feel like this was done on purpose?

tests/dialects/test_spark.py
```diff
@@ -125,13 +126,6 @@ def test_ddl(self):
                 "spark": "ALTER TABLE StudentInfo DROP COLUMNS (LastName, DOB)",
             },
         )
-        self.validate_all(
```
tobymao, line 128: why is this test removed?

sqlglot/transforms.py
```diff
@@ -466,6 +466,91 @@ def unqualify_columns(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def remove_unique_constraints(expression: exp.Expression) -> exp.Expression:
+    assert isinstance(expression, exp.Create)
```
georgesittas, line 470: should we instead switch this to `if isinstance(...):` and then return expression by default as in other transformations?

tests/dialects/test_hive.py
barakalon: I don't know wtf is going on here. Documentation is horrible, of course.
spark 2 seems to use hive format.
spark 3 seems to introduce a new format, in addition to the old format.
`CREATE TABLE ... USING ICEBERG...` seems to be the new format. And the code was already forcing that for spark 2 and 3. So I bet there are many cases that were never working...

We only just caught this at airbnb because the update the converts all vars to uppercase. We were setting the kind to lowercase "table", so this check was never true: https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/dialect.py#L664

tests/dialects/test_spark.py
barakalon: its a duplicate

sqlglot/transforms.py
barakalon: Probably better to fail. If this isn't exp.Create, something has gone wrong.
I'd set it as the type annotation.. but the typing is hard.

sqlglot/transforms.py
```diff
@@ -466,6 +466,91 @@ def unqualify_columns(expression: exp.Expression) -> exp.Expression:
...
+    # CTAS with temp tables map to CREATE TEMPORARY VIEW
+    kind = expression.args["kind"]
+    if kind.upper() == "TABLE" and temporary:
+        if expression.expression:
+            return exp.Create(
+                kind="TEMPORARY VIEW",
+                this=expression.this,
+                expression=expression.expression,
+            )
+        else:
```
tobymao, line 498: the else isn't needed here

sqlglot/transforms.py
```diff
@@ -466,6 +466,91 @@ def unqualify_columns(expression: exp.Expression) -> exp.Expression:
...
+    assert isinstance(expression, exp.Create)
+    properties = expression.args.get("properties")
+    temporary = any(
+        isinstance(prop, exp.TemporaryProperty)
+        for prop in (properties.expressions if properties else [])
+    )
+
+    # CTAS with temp tables map to CREATE TEMPORARY VIEW
+    kind = expression.args["kind"]
+    if kind.upper() == "TABLE" and temporary:
```
tobymao, line 491: can you make kind a property on create that does this auto upper?

tobymao (approved):

PR #2938, opened 2024-02-08
fix: ignore nulls with order by closes #2932

### LINKED ISSUE #2932: Order of `IGNORE NULLS` in BigQuery generated SQL is incorrect
`IGNORE NULLS` needs to occur before `LIMIT`, otherwise it's an error.

```
In [7]: import sqlglot as sg

In [8]: sg.__version__
Out[8]: '21.0.1'

In [9]: sg.parse_one('select array_agg(x ignore nulls limit 1)', read='bigquery').sql('bigquery')
Out[9]: 'SELECT array_agg(x LIMIT 1 IGNORE NULLS)'
```

**Official Documentation**

https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate-function-calls

tobymao:


tests/dialects/test_bigquery.py
```diff
@@ -62,14 +67,14 @@ def test_bigquery(self):
         self.validate_all(
             "SELECT ARRAY_AGG(DISTINCT x IGNORE NULLS ORDER BY a, b DESC LIMIT 10) AS x",
             write={
-                "duckdb": "SELECT ARRAY_AGG(DISTINCT x ORDER BY a NULLS FIRST, b DESC LIMIT 10 IGNORE NULLS) AS x",
+                "duckdb": "SELECT ARRAY_AGG(DISTINCT x IGNORE NULLS ORDER BY a NULLS FIRST, b DESC LIMIT 10) AS x",
```
georgesittas, line 70: I'd also include BQ in these tests, also the `"SELECT PERCENTILE_CONT(x, 0.5 RESPECT NULLS) OVER ()"` above with the two args

tests/dialects/test_bigquery.py
```diff
@@ -62,14 +67,14 @@ def test_bigquery(self):
...
                 "spark": "SELECT COLLECT_LIST(DISTINCT x ORDER BY a, b DESC LIMIT 10) IGNORE NULLS AS x",
             },
         )
         self.validate_all(
             "SELECT ARRAY_AGG(DISTINCT x IGNORE NULLS ORDER BY a, b DESC LIMIT 1, 10) AS x",
             write={
-                "duckdb": "SELECT ARRAY_AGG(DISTINCT x ORDER BY a NULLS FIRST, b DESC LIMIT 1, 10 IGNORE NULLS) AS x",
+                "duckdb": "SELECT ARRAY_AGG(DISTINCT x IGNORE NULLS ORDER BY a NULLS FIRST, b DESC LIMIT 1, 10) AS x",
```
georgesittas, line 77: This looks incorrect:
```
D SELECT ARRAY_AGG(DISTINCT x IGNORE NULLS ORDER BY a NULLS FIRST, b DESC LIMIT 1, 10) AS x;
Error: Parser Error: syntax error at or near "ORDER"
LINE 1: ...LECT ARRAY_AGG(DISTINCT x IGNORE NULLS ORDER BY a NULLS FIRST, b DESC LIMIT 1,...
                                                  ^
```

tobymao: where should ignore nulls go?

georgesittas: idk - it doesn't appear in the syntax spec? https://duckdb.org/docs/sql/aggregates.html

sqlglot/generator.py
```diff
@@ -2850,12 +2850,18 @@ def respectnulls_sql(self, expression: exp.RespectNulls) -> str:
...
 
     def _embed_ignore_nulls(self, expression: exp.IgnoreNulls | exp.RespectNulls, text: str) -> str:
-        if self.IGNORE_NULLS_IN_FUNC:
-            this = expression.find(exp.AggFunc)
-            if this:
-                sql = self.sql(this)
-                sql = sql[:-1] + f" {text})"
-                return sql
+        if self.IGNORE_NULLS_IN_FUNC and not expression.args.get("inline"):
+            for klass in (exp.Order, exp.Limit):
```
georgesittas, line 2854: Interesting, so order will be closer to the arg than limit and that's why you look for it first in this loop with `find`?

georgesittas: can we do this without introducing the `inline` argument? i think `meta` can work too right? we can avoid introducing the arg only for this

tests/dialects/test_bigquery.py
tobymao: ah you were testing duckdb, ya it doesn't work for duckb, https://duckdb.org/docs/sql/window_functions.html

sqlglot/generator.py
tobymao: yea, also realzide's ther's a having min by i'll do another pr for that

georgesittas: I was looking at the having min / max earlier, i'm not sure if it's applicable to all agg funcs in BQ. I'd double check if it's only an `ANY_VALUE` thing because it may not be necessary. I'm saying this because we also don't support that syntax in general, just for `ANY_VALUE`, and that's also the only func they have in their examples where that clause is used.

PR #2939, opened 2024-02-09
chore: cleanup having max

tobymao:


sqlglot/generator.py
```diff
@@ -2860,16 +2860,27 @@ def ignorenulls_sql(self, expression: exp.IgnoreNulls) -> str:
...
         if self.IGNORE_NULLS_IN_FUNC and not expression.meta.get("inline"):
-            for klass in (exp.Order, exp.Limit):
-                mod = expression.find(klass)
-
-                if mod:
-                    this = expression.__class__(this=mod.this.copy())
-                    this.meta["inline"] = True
-                    mod.this.replace(this)
-                    return self.sql(expression.this)
+            mods = sorted(
```
georgesittas, line 2871: I think I'd add a comment like this here because it's non-trivial; ou have to know the parser control flow to understand the sort order
```suggestion
            # The first modifier here will be the one closest to the AggFunc's arg
            mods = sorted(
```

georgesittas: why not do a for loop like previously?

tobymao: it’s more expensive because you have to search n times

georgesittas: ok, do you wanna go with the comment or leave it as-is? i think that's the only comment, looks good otherwise

georgesittas (approved):

PR #2944, opened 2024-02-09
fix: explode to unnest for bigquery

chelsea-lin:


tobymao:
thanks for  the pr, but i don't really want to maintain two different versions of this function without a good reason

chelsea-lin:
It sounds good to me if we can avoid code duplication! Thanks for your fixing!

PR #2946, opened 2024-02-09
Fix(bigquery): fix annotation of timestamp(x)

georgesittas:
Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#timestamp


sqlglot/dialects/bigquery.py
```diff
@@ -168,12 +168,18 @@ def _pushdown_cte_column_names(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def _parse_timestamp(args: t.List) -> exp.StrToTime:
+def _parse_parse_timestamp(args: t.List) -> exp.StrToTime:
```
tobymao, line 171: why is it called parse parse

georgesittas: The function we're constructing is called `PARSE_TIMESTAMP`, so that's where the suffix comes from. The `_parse` prefix is to be consistent with other helpers in the file (and in general).

Imho "parse" is the wrong term to use for `FUNCTIONS` entries - a better fit would be something like "construct" which reflects what we're doing. What do you think?

tobymao: sure you can change it to build

sqlglot/expressions.py
```diff
@@ -4738,7 +4738,12 @@ class Extract(Func):
 
 
 class Timestamp(Func):
-    arg_types = {"this": False, "expression": False}
+    arg_types = {"this": False, "expression": False, "with_tz": False}
+
+    @property
+    def return_type(self) -> DataType:
```
tobymao, line 4744: shouldn't you just override type here?

sqlglot/dialects/bigquery.py
georgesittas: let's do in a followup PR, I'll make sure to clean up other similar function names

tobymao: sure

tobymao: we can just leave it as is

sqlglot/expressions.py
georgesittas: hmm not sure, that kinda leaves room for future bugs if we ever decide for some reason to set the type of Timestamp to something else

georgesittas: maybe we can do though `return self._type or ...`

tobymao: should we just have this logic in annotate_types, and then call annotator in the section that you use it?

georgesittas: so instead of `expr.return_type` do `annotate_types(expr)` first and then `expr.type` in the other places we need this info?

tobymao: yea, that way we only have typing logic in one place, the annotator

georgesittas: that could work, only issue being that we'd annotate the whole subtree of the function call even though we only care about the outermost node

georgesittas: ok let's do it that way, it seems like that path shouldn't be hit too frequently anyway so 🤷

PR #2953, opened 2024-02-12
feat(clickhouse): add basic support for system statement

GaliFFun:
First of all, thanks a lot for your project. It really helps me during at my job. But I still have a few problems with the ClickHouse support, one problem being the lack of [SYSTEM statement](https://clickhouse.com/docs/en/sql-reference/statements/system) support. We use this statement to stop merges on temporary tables in ETL processes, so ETL code parsing fails without that one. If you think the support for SYSTEM statement should be more than just having command (which is honestly perfectly fine for me, as it could be easily parsed manually, if needed at all), please let me know and I'll modify the PR

### DIFF
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -369,6 +369,7 @@ class TestClickhouse(Validator):
             "SELECT STARTS_WITH('a', 'b'), STARTSWITH('a', 'b')",
             write={"clickhouse": "SELECT startsWith('a', 'b'), startsWith('a', 'b')"},
         )
+        self.validate_identity("SYSTEM STOP MERGES foo.bar")
 
     def test_cte(self):
         self.validate_identity("WITH 'x' AS foo SELECT foo")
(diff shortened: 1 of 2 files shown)


tests/dialects/test_clickhouse.py
```diff
@@ -369,6 +369,7 @@ def test_clickhouse(self):
             "SELECT STARTS_WITH('a', 'b'), STARTSWITH('a', 'b')",
             write={"clickhouse": "SELECT startsWith('a', 'b'), startsWith('a', 'b')"},
         )
+        self.validate_identity("SYSTEM STOP MERGES foo.bar")
```
georgesittas, line 372: This silences the relevant warning when running the test suite.
```suggestion
        self.validate_identity("SYSTEM STOP MERGES foo.bar", check_command_warning=True)
```

georgesittas (approved):
Parsing the statement as a `Command` seems fine for now - thanks for the kind words and of course for the PR!

tests/dialects/test_clickhouse.py
GaliFFun: Will keep that in mind for my future contributions, thanks!

PR #2955, opened 2024-02-12
fix(optimizer): qualify_columns optimizations for wide tables

barakalon:
Small optimization for tables with lots of columns.


georgesittas (approved):
Good catch

PR #2962, opened 2024-02-13
Fix(parser): enable parsing of values into Identifier for some dialects

### LINKED ISSUE #2957: Snowflake keyword in dot notation
**Before you file an issue**

```sql
-- Snowflake
SELECT  col1,
        value:values::string AS nested_value
FROM my_table
```

Gives an error
```
sqlglot.errors.ParseError: Required keyword: 'expression' missing for <class 'sqlglot.expressions.JSONExtract'>. Line 2, Col: 13.
```

In particular, it seems to be tripped up by the "values" name of the field. Changing the nested field name to be something else, or using `value['values']::string` parses correctly.

I'm also pretty surprised that Snowflake allows "values" to show up unquoted in the column expression. If you want to test this behavior yourself, I used this script

```sql
CREATE TEMP TABLE my_table (
    col1 INT,
    col2 VARIANT
);

INSERT INTO my_table (col1, col2)
SELECT *
FROM (
    SELECT 1 as col1, PARSE_JSON('{"values": "dummy_value1"}') as col2
    UNION ALL
    SELECT 2 as col1, PARSE_JSON('{"values": "dummy_value2"}')
    UNION ALL
    SELECT 3 as col1, PARSE_JSON('{"values": "dummy_value3"}')
) AS dummy_data;


SELECT  col1,
        col2:values::string AS nested_value
FROM my_table
```

**Official Documentation**

georgesittas:
Fixes #2957

I investigated a bit and this is what I found for the dialects currently supported by SQLGlot w.r.t. the `VALUES` clause syntax:

- Doesn't have a `VALUES` clause: BQ, Redshift

- '(' needs to follow `VALUES`: ClickHouse, DuckDB, Postgres, Snowflake, SQLite, T-SQL

- 'ROW' needs to follow `VALUES`: MySQL, probably Doris and Starrocks too due to lineage (haven't tested)

- "Anything" can follow `VALUES` (side effect of allowing stuff like `VALUES 1, 2, 3`): DataBricks, Hive, Oracle, Presto, Spark, Teradata, Trino

This PR allows `values` to be parsed into an `Identifier` for the first three categories. For the last one, fixing the parser is a bit more tricky (due to ambiguity) and so we decided to tackle it in a future PR. FWIW Presto and Trino don't seem to allow an unquoted `values` to be used as an identifier:

```
trino> select 1 as values;
Query 20240213_162108_00003_sw76y failed: line 1:13: mismatched input 'values'. Expecting: <identifier>
select 1 as values
```

However, Spark for example allows it:

```
spark-sql (default)> with t as (select 1 as values) select values from t;
1
```

References:
- https://duckdb.org/docs/sql/query_syntax/values.html
- https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-values.html
- https://dev.mysql.com/doc/refman/8.0/en/values.html
- https://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj11277.html
- https://www.postgresql.org/docs/current/sql-values.html
- https://prestodb.io/docs/current/sql/values.html
- https://docs.snowflake.com/en/sql-reference/constructs/values
- https://www.sqlite.org/lang_select.html
- https://teradata.github.io/presto/docs/0.167-t/sql/values.html
- https://learn.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver16

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -47,6 +47,11 @@ class Redshift(Postgres):
     }
 
     class Parser(Postgres.Parser):
+        ID_VAR_TOKENS = {
+            *Postgres.Parser.ID_VAR_TOKENS,
+            TokenType.VALUES,
+        }
+
         FUNCTIONS = {
             **Postgres.Parser.FUNCTIONS,
             "ADD_MONTHS": lambda args: exp.TsOrDsAdd(
--- a/tests/fixtures/identity.sql
+++ b/tests/fixtures/identity.sql
@@ -847,3 +847,5 @@ WITH use(use) AS (SELECT 1) SELECT use FROM use
 SELECT recursive FROM t
 SELECT (ROW_NUMBER() OVER (PARTITION BY user ORDER BY date ASC) - ROW_NUMBER() OVER (PARTITION BY user, segment ORDER BY date ASC)) AS group_id FROM example_table
 CAST(foo AS BPCHAR)
+values
+SELECT values
--- a/tests/fixtures/identity.sql
+++ b/tests/fixtures/identity.sql
@@ -849,3 +849,4 @@ SELECT (ROW_NUMBER() OVER (PARTITION BY user ORDER BY date ASC) - ROW_NUMBER() O
 CAST(foo AS BPCHAR)
 values
 SELECT values
+SELECT values AS values FROM t WHERE values + 1 > 3
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -266,6 +266,7 @@ class Hive(Dialect):
     class Parser(parser.Parser):
         LOG_DEFAULTS_TO_LN = True
         STRICT_CAST = False
+        VALUES_FOLLOWED_BY_PAREN = False
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -445,6 +445,7 @@ class MySQL(Dialect):
 
         LOG_DEFAULTS_TO_LN = True
         STRING_ALIASES = True
+        VALUES_FOLLOWED_BY_PAREN = False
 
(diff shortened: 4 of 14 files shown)


georgesittas:
There's a new DuckDB [version](https://github.com/duckdb/duckdb/releases/tag/v0.10.0), hence the test failures. Looking into it now..

tests/fixtures/identity.sql
```diff
@@ -847,3 +847,5 @@ WITH use(use) AS (SELECT 1) SELECT use FROM use
 SELECT recursive FROM t
 SELECT (ROW_NUMBER() OVER (PARTITION BY user ORDER BY date ASC) - ROW_NUMBER() OVER (PARTITION BY user, segment ORDER BY date ASC)) AS group_id FROM example_table
 CAST(foo AS BPCHAR)
+values
+SELECT values
```
tobymao, line 851: can you also add select values as values

sqlglot/dialects/redshift.py
```diff
@@ -47,6 +47,11 @@ class Redshift(Postgres):
     }
 
     class Parser(Postgres.Parser):
+        ID_VAR_TOKENS = {
+            *Postgres.Parser.ID_VAR_TOKENS,
+            TokenType.VALUES,
```
tobymao, line 52: does bigquery also define this?

georgesittas: Yep, fyi I didn't move this logic into the metaclass because it's just these two that don't support `VALUES` and it felt overkill to dry it up like that. Maybe a third one would justify it.

tests/fixtures/identity.sql
georgesittas: yes, done

sqlglot/dialects/redshift.py
tobymao: nah it's fine

tobymao (approved):

tobymao (approved):

PR #2965, opened 2024-02-13
Fix: don't treat VALUES as a keyword in BigQuery, Redshift

### LINKED ISSUE #2963: Parse error for BigQuery when `values` column referenced in `STRUCT()` expression
In BigQuery `VALUES` isn't a reserved keyword and can be used as a column name.  However, when sqlglot parses a query where a `values` column is referenced in a `STRUCT()` expression it raises an error.

It looks like the `values` column reference is being misinterpreted as a keyword in that case.  It appears this was also true in previous versions of sqlglot, but it simply resulted in an incorrect AST and didn't fail with an error (we noticed this issue when upgrading sqlglot from 20.4.0 to 21.1.0).  I'm guessing the change in behavior is due to #2650.

Ideally sqlglot would not treat the `values` column reference as a keyword in such cases.

```python
sql = '''WITH test AS (SELECT 'abc' AS values) SELECT STRUCT(values AS v) FROM test'''
sqlglot.parse(sql, read='bigquery')
```
Results in the following error:
```
sqlglot.errors.ParseError: Required keyword: 'this' missing for <class 'sqlglot.expressions.Alias'>. Line 1, Col: 64.
```

**Official Documentation**
- https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords
- https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#constructing_a_struct

georgesittas:
Fixes #2963

Many of the lines changed are just reordered tests - did that to group `validate_identity` and `validate_all` together but can revert if it's too noisy. Will comment inline to point out the changed / added ones.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -330,7 +330,6 @@ class BigQuery(Dialect):
             "RECORD": TokenType.STRUCT,
             "TIMESTAMP": TokenType.TIMESTAMPTZ,
         }
-
         KEYWORDS.pop("DIV", None)
         KEYWORDS.pop("VALUES", None)
 
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -330,7 +330,9 @@ class BigQuery(Dialect):
             "RECORD": TokenType.STRUCT,
             "TIMESTAMP": TokenType.TIMESTAMPTZ,
         }
-        KEYWORDS.pop("DIV")
+
+        KEYWORDS.pop("DIV", None)
+        KEYWORDS.pop("VALUES", None)
 
     class Parser(parser.Parser):
         PREFIXED_PIVOT_COLUMNS = True
@@ -407,11 +409,6 @@ class BigQuery(Dialect):
         NESTED_TYPE_TOKENS = {
             *parser.Parser.NESTED_TYPE_TOKENS,
             TokenType.TABLE,
-        }
-
-        ID_VAR_TOKENS = {
-            *parser.Parser.ID_VAR_TOKENS,
-            TokenType.VALUES,
         }
 
         PROPERTY_PARSERS = {
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -330,8 +330,8 @@ class BigQuery(Dialect):
             "RECORD": TokenType.STRUCT,
             "TIMESTAMP": TokenType.TIMESTAMPTZ,
         }
-        KEYWORDS.pop("DIV", None)
-        KEYWORDS.pop("VALUES", None)
+        KEYWORDS.pop("DIV")
+        KEYWORDS.pop("VALUES")
 
     class Parser(parser.Parser):
         PREFIXED_PIVOT_COLUMNS = True
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -158,7 +158,7 @@ class Redshift(Postgres):
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
         }
-        KEYWORDS.pop("VALUES", None)
+        KEYWORDS.pop("VALUES")
 
         # Redshift allows # to appear as a table identifier prefix
         SINGLE_TOKENS = Postgres.Tokenizer.SINGLE_TOKENS.copy()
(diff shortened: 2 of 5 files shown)


tests/dialects/test_bigquery.py
```diff
@@ -1013,45 +1017,20 @@ def test_bigquery(self):
             },
         )
         self.validate_all(
-            "SELECT cola, colb FROM (VALUES (1, 'test')) AS tab(cola, colb)",
-            write={
-                "spark": "SELECT cola, colb FROM VALUES (1, 'test') AS tab(cola, colb)",
+            "SELECT cola, colb FROM UNNEST([STRUCT(1 AS cola, 'test' AS colb)])",
+            read={
```
georgesittas, line 1021: These tests were changed, previous ones were incorrect in both that the input queries were invalid in BQ and the expected queries were again invalid BQ.

tests/dialects/test_bigquery.py
```diff
@@ -187,6 +100,16 @@ def test_bigquery(self):
             )
             assert "'END FOR'" in cm.output[0]
 
+        self.validate_identity("CREATE SCHEMA x DEFAULT COLLATE 'en'")
+        self.validate_identity("CREATE TABLE x (y INT64) DEFAULT COLLATE 'en'")
+        self.validate_identity("PARSE_JSON('{}', wide_number_mode => 'exact')")
+        self.validate_identity("FOO(values)")
+        self.validate_identity("STRUCT(values AS value)")
```
georgesittas, line 107: The problem raised by the issue would also occur with any function, because function args are parsed using `_parse_lambda` which ends up calling `_parse_select_or_expression` and previously we were hitting the branch where the `VALUES` clause parsing happens. Hence the recently [changed](https://github.com/tobymao/sqlglot/commit/1c67f030cd9df530e26c620079b2298b1db97d50) code wasn't reached.

sqlglot/dialects/bigquery.py
```diff
@@ -330,7 +330,8 @@ class Tokenizer(tokens.Tokenizer):
             "RECORD": TokenType.STRUCT,
             "TIMESTAMP": TokenType.TIMESTAMPTZ,
         }
-        KEYWORDS.pop("DIV")
+        KEYWORDS.pop("DIV", None)
```
tobymao, line 333: why did you have to add None here?

sqlglot/parser.py
```diff
@@ -2974,7 +2970,7 @@ def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
 
     def _parse_derived_table_values(self) -> t.Optional[exp.Values]:
         is_derived = self._match_pair(TokenType.L_PAREN, TokenType.VALUES)
-        if not is_derived and not self._match(TokenType.VALUES):
+        if not is_derived and not self._match_text_seq("VALUES"):
```
tobymao, line 2973: why text and not a token?

sqlglot/dialects/bigquery.py
georgesittas: I started by doing that for `"VALUES"` with the reasoning that it would protect by throwing a key error in the future if there's no keyword to pop for whatever reason, and then I updated that to apply the same logic. Should I revert?

tobymao: yea, i don't understand why that's necessary,  get rid of the None, it's unnecessary

sqlglot/dialects/redshift.py
```diff
@@ -163,6 +158,7 @@ class Tokenizer(Postgres.Tokenizer):
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
         }
+        KEYWORDS.pop("VALUES", None)
```
tobymao, line 161: here too

sqlglot/parser.py
tobymao: hm, if bq and redshift have values statements, shouldn't they have the token?

georgesittas: The motivation for getting rid of the token was that we don't wanna hit the `if match(VALUES)` branch in the `SELECT` parser, because that path is reached very often and we don't want that for BQ / Redshift. If we don't pop them off then we'll probably need more shenanigans with flags etc, like `if match(VALUES) and SUPPORTS_VALUES_CLAUSE: ...`. The insert parser would need to be updated either way I think.

tobymao: ok sounds good

tobymao (approved):

PR #2974, opened 2024-02-14
Fix(postgres): ensure json extraction can roundtrip unaltered

### LINKED ISSUE #2971: json_extract_path is added for JSONB columns
This [PR](https://github.com/tobymao/sqlglot/pull/2883) introduces a regression with postgres

Before this PR I could do:
``python
from sqlglot import parse_one
query = """ 
SELECT
   (j->'i')::int,
  (j->>'i')::int,
   (j->'f')::float,
   (j->>'f')::float
FROM
  (SELECT '{"i":123,"f":12.34}'::jsonb) t(j)
"""
parse_one(query).sql(dialect="postgres")
 ```
That would give me a query that i could execute
```sql
SELECT 
  CAST((j -> 'i') AS INT),
  CAST((j ->> 'i') AS INT),
  CAST((j -> 'f') AS REAL),
  CAST((j ->> 'f') AS REAL)
FROM (SELECT CAST('{"i":123,"f":12.34}' AS jsonb)) AS t(j)
```

With version 21.1.0, the same code returns
```sql
SELECT
  CAST((JSON_EXTRACT_PATH(j, 'i')) AS INT),
  CAST((JSON_EXTRACT_PATH_TEXT(j, 'i')) AS INT),
  CAST((JSON_EXTRACT_PATH(j, 'f')) AS REAL),
  CAST((JSON_EXTRACT_PATH_TEXT(j, 'f')) AS REAL)
FROM (SELECT CAST('{"i":123,"f":12.34}' AS jsonb)) AS t(j)
```

When I execute it I have this error : `ERROR:  function json_extract_path(jsonb, unknown) does not exist` because `json_extract_path` expects a JSON.
If I manually change the query to cast `j` to json like this

```sql
SELECT
  CAST((JSON_EXTRACT_PATH(j, 'i')) AS INT),
  CAST((JSON_EXTRACT_PATH_TEXT(j, 'i')) AS INT),
  CAST((JSON_EXTRACT_PATH(j, 'f')) AS REAL),
  CAST((JSON_EXTRACT_PATH_TEXT(j, 'f')) AS REAL)
FROM (SELECT CAST('{"i":123,"f":12.34}' AS json)) AS t(j)
```

I have this error : `ERROR:  cannot cast type json to integer` because you can not cast json to integer you need to either a `JSONB` or a `TEXT`.

georgesittas:
Fixes #2971

Based on Postgres' [docs](https://www.postgresql.org/docs/9.3/functions-json.html), the various arrow json operators expect their document operand to be a JSON[B] value.

> [Table 9-40](https://www.postgresql.org/docs/9.3/functions-json.html#FUNCTIONS-JSON-OP-TABLE) shows the operators that are available for use with JSON (see [Section 8.14](https://www.postgresql.org/docs/9.3/datatype-json.html)) data.

Additionally, their function counterparts also work with raw text JSON documents:

```sql
json_extract_path('{"f2":{"f3":1},"f4":{"f5":99,"f6":"foo"}}', 'f4')
```

This means that we can't _always_ produce just _one_ of these two syntax forms: if we produce an arrow then we'll break queries where JSON documents are strings (like above), whereas if we produce a function we'll break queries where the document's type is incompatible (like in the linked issue, where we need a `JSONB*` alternative).

This PR is an attempt to preserve this information at the AST level so we can satisfy the type constraints at generation time by producing a suitable operator / function.

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -487,8 +487,10 @@ class Hive(Dialect):
             exp.If: if_sql(),
             exp.ILike: no_ilike_sql,
             exp.IsNan: rename_func("ISNAN"),
-            exp.JSONExtract: rename_func("GET_JSON_OBJECT"),
-            exp.JSONExtractScalar: rename_func("GET_JSON_OBJECT"),
+            exp.JSONExtract: lambda self, e: self.func("GET_JSON_OBJECT", e.this, e.expression),
+            exp.JSONExtractScalar: lambda self, e: self.func(
+                "GET_JSON_OBJECT", e.this, e.expression
+            ),
             exp.JSONFormat: _json_format_sql,
             exp.Left: left_to_substring_sql,
             exp.Map: var_map_sql,
--- a/sqlglot/dataframe/sql/functions.py
+++ b/sqlglot/dataframe/sql/functions.py
@@ -1030,7 +1030,7 @@ def posexplode_outer(col: ColumnOrName) -> Column:
 
 
 def get_json_object(col: ColumnOrName, path: str) -> Column:
-    return Column.invoke_expression_over_column(col, expression.JSONExtract, path=lit(path))
+    return Column.invoke_expression_over_column(col, expression.JSONExtract, expression=lit(path))
 
 
 def json_tuple(col: ColumnOrName, *fields: str) -> Column:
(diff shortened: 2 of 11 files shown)


sqlglot/dialects/hive.py
```diff
@@ -487,8 +487,10 @@ class Generator(generator.Generator):
             exp.If: if_sql(),
             exp.ILike: no_ilike_sql,
             exp.IsNan: rename_func("ISNAN"),
-            exp.JSONExtract: rename_func("GET_JSON_OBJECT"),
-            exp.JSONExtractScalar: rename_func("GET_JSON_OBJECT"),
+            exp.JSONExtract: lambda self, e: self.func("GET_JSON_OBJECT", e.this, e.expression),
+            exp.JSONExtractScalar: lambda self, e: self.func(
+                "GET_JSON_OBJECT", e.this, e.expression
```
georgesittas, line 492: Btw @tobymao changes like this is why I kinda wanted to filter out bools

tobymao: maybe we can try it later

tobymao (approved):

PR #2977, opened 2024-02-15
Refactor: clean up tech debt in dialect implementations

georgesittas:
I've been meaning to do this sweep for quite some time. The changes at a high level are:

- Several misc. refactors, like the `_derived_table_values_to_unnest` one
- Use `build` instead of `parse` when defining helper callbacks that are used as values in `FUNCTIONS` entries - the latter term [wasn't quite correct](https://github.com/tobymao/sqlglot/pull/2946#discussion_r1484646526), as there's no _parsing_ involved but rather only constructing expressions
- Use `self.func` in most places over manually constructing strings like `f"FOO({...})`. This ensures we'll use the pretty-printing logic for these functions when needed. There are some exceptions that I didn't change because they were more complex, e.g. they had additional casts and nested function calls and I wanted to avoid the parsing overhead
- Rename a few things here and there to better capture the intent of the corresponding functions / variables / methods
- Try to enforce some level of consistency w.r.t. naming schemes, formatting, code structure etc

These changes are somewhat opinionated, happy to chat more and revert those that others don't agree with. The motivation was to gradually try and make the codebase a bit cleaner so that newcomers can navigate it more easily and avoid replicating smelly patterns. A nice side effect was that in the process I found some things that could be cleaned up.

Need to do another pass on this to make sure I didn't miss something & double-check that the changes are good.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -148,59 +148,65 @@ class _Dialect(type):
 
 class Dialect(metaclass=_Dialect):
     INDEX_OFFSET = 0
-    """Determines the base index offset for arrays."""
+    """The base index offset for arrays."""
 
     WEEK_OFFSET = 0
-    """Determines the day of week of DATE_TRUNC(week). Defaults to 0 (Monday). -1 would be Sunday."""
+    """First day of the week in DATE_TRUNC(week). Defaults to 0 (Monday). -1 would be Sunday."""
 
     UNNEST_COLUMN_ONLY = False
-    """Determines whether or not `UNNEST` table aliases are treated as column aliases."""
+    """Whether `UNNEST` table aliases are treated as column aliases."""
 
     ALIAS_POST_TABLESAMPLE = False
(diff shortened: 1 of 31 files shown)


tobymao (approved):

sqlglot/dialects/dialect.py
```diff
@@ -184,17 +184,17 @@ class Dialect(metaclass=_Dialect):
     """Determines how function names are going to be normalized."""
 
     LOG_BASE_FIRST = True
-    """Determines whether the base comes first in the `LOG` function."""
+    """Determines whether or not the base comes first in the `LOG` function."""
```
barakalon, line 187: PyCharm's default inspector complains about "whether or not", saying the "or not" is redundant haha

barakalon (approved):

sqlglot/dialects/dialect.py
georgesittas: Lol, kinda makes sense.. I'll get rid of it.

PR #2982, opened 2024-02-16
feat: Add basic support for CREATE USER/ROLE in snowflake.

DanCardin:
https://docs.snowflake.com/en/sql-reference/sql/create-user

The idea here is to add somewhat rudimentary support for `CREATE USER/ROLE` statements syntactically, so it can be more strictly validated elsewhere. The majority of the options for the statement (beyond name) are definitely snowflake specific, so it seemed like sqlglot the project would have no incentive to want to support full validation of all the options and whatnot.

For that reason, I put all the code inside the snowflake dialect. Although I'm sure there's a minimum baseline of intercompatibility (i.e. `create user name` with no options) that could be implemented more broadly...

I didn't see any other examples of similar syntax being parsed, so I was sort of guessing at the right constructs to use (I was looking primarily at how table creation properties are defined) and/or the right way to go about testing the output structure.


georgesittas:
Hey, I don't think the approach in this PR is quite correct. Take a look at a similar PR [here](https://github.com/tobymao/sqlglot/pull/2985), the approach should be similar here too.

georgesittas:
Gonna close this PR and see if I can add the `USER`, `ROLE` kinds when I [refactor](https://github.com/tobymao/sqlglot/pull/2985#pullrequestreview-1887042549) the `CREATABLES` dict.

PR #2983, opened 2024-02-17
Feat(tests): update test_executor with tpc-ds 

fool1280:
Add method to test TPC-DS


tests/test_executor.py
```diff
@@ -91,7 +94,37 @@ def to_csv(expression):
                     ),
                 )
             ):
-                with self.subTest(f"tpch-h {i + 1}"):
+                with self.subTest(f"tpc-h {i + 1}"):
+                    sql, _ = self.sqls[i]
```
tobymao, line 98: this can probably be a helper method so that we can reuse both right? perhaps the whole test could be a helper method

fool1280: Done

sqlglot/executor/python.py
```diff
@@ -150,7 +150,8 @@ def scan_csv(self, step):
                             types.append(type(ast.literal_eval(v)))
                         except (ValueError, SyntaxError):
                             types.append(str)
-                context.set_row(tuple(t(v) for t, v in zip(types, row)))
+                # handle '' in csv
+                context.set_row(tuple(t(v) if v else 0 for t, v in zip(types, row)))
```
georgesittas, line 154: What triggered the need for this change?

tests/test_executor.py
```diff
@@ -17,40 +18,60 @@
...
+                FROM READ_CSV('{DIR_TPCH}{table}.csv.gz', delim='|', header=True, columns={columns})
+                """
+            )
+
+        for table, columns in TPCDS_SCHEMA.items():
+            cls.tpcds_conn.execute(
+                f"""
+                CREATE VIEW {table} AS
+                SELECT *
+                FROM READ_CSV('{DIR_TPCDS}{table}.csv.gz', delim=',', header=True, columns={columns})
```
georgesittas, line 50: We can write this a bit more compactly by making a double for loop:

```python
for (schema, dir, delim) in ((TPCH_SCHEMA, DIR_TPCH, '|'), (TPCDS_SCHEMA, DIR_TPCDS, ',')):
    ...
```

Also, why do we need two connections? If we're gonna load them sequentially and then test the two query sets in isolation, then won't one db connection suffice?

tests/test_executor.py
```diff
@@ -17,40 +18,60 @@
...
+        ]
 
     @classmethod
     def tearDownClass(cls):
-        cls.conn.close()
+        cls.tpch_conn.close()
+        cls.tpcds_conn.close()
 
-    def cached_execute(self, sql):
+    def cached_execute(self, sql, type_exec="tpc-h"):
```
georgesittas, line 69: Let's use a flag here instead: `tpch: bool = True`

georgesittas: Or, actually, if we can use just a single connection then this won't be needed anymore, so let's check that first.

tests/test_executor.py
```diff
@@ -66,40 +87,67 @@ def test_py_dialect(self):
...
-        for i, (sql, optimized) in enumerate(self.sqls[:20], start=1):
+        for i, (sql, optimized) in enumerate(self.tpch_sqls[:20], start=1):
             with self.subTest(f"{i}, {sql}"):
                 a = self.cached_execute(sql)
-                b = self.conn.execute(optimized).fetchdf()
+                b = self.tpch_conn.execute(optimized).fetchdf()
                 self.rename_anonymous(b, a)
                 assert_frame_equal(a, b)
 
+    def subtestHelper(self, i, table, type_exec="tpc-h"):
```
georgesittas, line 97: We can use a flag here as well

tests/test_executor.py
```diff
@@ -66,40 +87,67 @@ def test_py_dialect(self):
...
+        with self.subTest(f"{type_exec} {i + 1}"):
+            if type_exec == "tpc-h":
+                sql, _ = self.tpch_sqls[i]
+            else:
+                sql, _ = self.tpcds_sqls[i]
```
georgesittas, line 102: Nit: maybe rewrite this a bit more compactly? Don't feel strongly about it, just gets rid of some extra visual noise.
```suggestion
            sql, _ = self.tpch_sqls[i] if tpch else self.tpcds_sqls[i]
```

tests/test_executor.py
```diff
@@ -66,40 +87,67 @@ def test_py_dialect(self):
...
+                    ),
+                )
+            ):
+                self.subtestHelper(i, table, "tpc-h")
+
+    def test_execute_tpcds(self):
+        def to_csv(expression):
+            if isinstance(expression, exp.Table) and os.path.exists(
+                f"{DIR_TPCDS}{expression.name}.csv.gz"
+            ):
```
georgesittas, line 134: I guess this `exists` check was needed because `expression` can also be a CTE (given that it doesn't correspond to a CSV file)?

tests/test_executor.py
```diff
@@ -66,40 +87,67 @@ def test_py_dialect(self):
...
                     (
                         (parse_one(sql).transform(to_csv).sql(pretty=True), TPCH_SCHEMA)
-                        for sql, _ in self.sqls
+                        for sql, _ in self.tpch_sqls
+                    ),
+                )
+            ):
+                self.subtestHelper(i, table, "tpc-h")
+
+    def test_execute_tpcds(self):
```
georgesittas, line 130: Looks like we can probably [dry](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this up a bit more. You can give it a shot.

sqlglot/executor/python.py
fool1280: I ran into this error `invalid literal for int() with base 10: ''`, as in the tpc-ds dataset there some null values

tobymao: that probably means there's another bug so don't do that

tests/test_executor.py
fool1280: There are two `customer.csv` for tpch and tpcds but with different schema. I'm considering merging them by union by name.

tests/test_executor.py
fool1280: yes

tests/test_executor.py
```diff
@@ -66,18 +90,28 @@ def test_py_dialect(self):
...
-                b = self.conn.execute(transpile(optimized, write="duckdb")[0]).fetchdf()
+                a = self.cached_execute(sql, tpch=True)
+                b = self.tpch_conn.execute(transpile(optimized, write="duckdb")[0]).fetchdf()
                 self.rename_anonymous(b, a)
                 assert_frame_equal(a, b)
 
+    def subtestHelper(self, i, table, tpch=True):
+        with self.subTest(f"{'tpc-h' if tpch else 'tpc-ds'} {i + 1}"):
+            sql, _ = self.tpch_sqls[i] if tpch else self.tpcds_sqls[i]
+            a = self.cached_execute(sql, tpch=True)
```
georgesittas, line 103:
```suggestion
            a = self.cached_execute(sql, tpch=tpch)
```

test.py
georgesittas, line 1: probably forgot to remove this one

sqlglot/executor/python.py
```diff
@@ -142,15 +142,16 @@ def scan_csv(self, step):
...
                 if not types:
                     for v in row:
                         try:
                             types.append(type(ast.literal_eval(v)))
                         except (ValueError, SyntaxError):
                             types.append(str)
-                context.set_row(tuple(t(v) for t, v in zip(types, row)))
+
+                # for empty string, set default value instead of casting
+                context.set_row(tuple(t(v) if v else t() for t, v in zip(types, row)))
```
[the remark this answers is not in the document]
fool1280: I don't think this could work because there are some aggregate queries and they are invalid with `None`:
```
WITH customer_total_return
     AS (SELECT sr_customer_sk     AS ctr_customer_sk,
                sr_store_sk        AS ctr_store_sk,
                Sum(sr_return_amt) AS ctr_total_return
         FROM   store_returns,
                date_dim
         WHERE  sr_returned_date_sk = d_date_sk
                AND d_year = 2001
         GROUP  BY sr_customer_sk,
                   sr_store_sk)
SELECT c_customer_id
FROM   customer_total_return ctr1,
       store,
       customer
WHERE  ctr1.ctr_total_return > (SELECT Avg(ctr_total_return) * 1.2
                                FROM   customer_total_return ctr2
                                WHERE  ctr1.ctr_store_sk = ctr2.ctr_store_sk)
       AND s_store_sk = ctr1.ctr_store_sk
       AND s_state = 'TN'
       AND ctr1.ctr_customer_sk = c_customer_sk
ORDER  BY c_customer_id
LIMIT 100
``` 

Error: `Step 'Aggregate: customer_total_return (4936004752)' failed: '<' not supported between instances of 'NoneType' and 'int'`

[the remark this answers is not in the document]
tobymao: that means that the aggregation code is not correct. they should be able to support None

tests/test_executor.py
```diff
@@ -17,40 +18,59 @@
...
     @classmethod
     def tearDownClass(cls):
-        cls.conn.close()
+        cls.tpch_conn.close()
+        cls.tpcds_conn.close()
 
-    def cached_execute(self, sql):
+    def cached_execute(self, sql, tpch=True):
         if sql not in self.cache:
-            self.cache[sql] = self.conn.execute(transpile(sql, write="duckdb")[0]).fetchdf()
```
tobymao, line 53: conn = self.tpch_conn if tpch else ...

sqlglot/executor/python.py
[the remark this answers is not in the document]
fool1280: Oh ok, let me investigate more

tests/test_executor.py
```diff
@@ -66,18 +86,28 @@ def test_py_dialect(self):
         self.assertEqual(generate(parse_one("x is null")), "scope[None][x] is None")
 
     def test_optimized_tpch(self):
-        for i, (sql, optimized) in enumerate(self.sqls, start=1):
+        for i, (_, sql, optimized) in enumerate(self.tpch_sqls, start=1):
             with self.subTest(f"{i}, {sql}"):
-                a = self.cached_execute(sql)
-                b = self.conn.execute(transpile(optimized, write="duckdb")[0]).fetchdf()
+                a = self.cached_execute(sql, tpch=True)
```
georgesittas, line 91:
```suggestion
                a = self.cached_execute(sql)
```

tests/test_executor.py
fool1280: done

sqlglot/executor/python.py
```diff
@@ -142,15 +142,18 @@ def scan_csv(self, step):
...
             for row in reader:
                 if not types:
                     for v in row:
                         try:
                             types.append(type(ast.literal_eval(v)))
                         except (ValueError, SyntaxError):
                             types.append(str)
-                context.set_row(tuple(t(v) for t, v in zip(types, row)))
+
+                # for NULL values with type str, we convert them to None instead of casting
```
georgesittas, line 153:
```suggestion
                # We can't cast empty values ('') to non-string types, so we convert them to None instead
```

georgesittas (approved):
Looks good for a first pass, we can iterate on this moving forward.

Nice work @fool1280 👍

sqlglot/executor/python.py
```diff
@@ -142,15 +142,18 @@ def scan_csv(self, step):
...
                     for v in row:
                         try:
                             types.append(type(ast.literal_eval(v)))
                         except (ValueError, SyntaxError):
                             types.append(str)
-                context.set_row(tuple(t(v) for t, v in zip(types, row)))
+
+                # We can't cast empty values ('') to non-string types, so we convert them to None instead
+                context.set_row(
+                    tuple(None if (v == "" and t is not str) else t(v) for t, v in zip(types, row))
```
tobymao, line 155:
```suggestion
                    tuple(None if (t is not str and v == "") else t(v) for t, v in zip(types, row))
```

fool1280:
thanks Toby!

PR #2984, opened 2024-02-17
Feat(duckdb): implement generation logic for exp.ArrayAny

georgesittas:
Reference: https://duckdb.org/docs/sql/functions/nested.html

### DIFF
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -66,10 +66,10 @@ class TestExecutor(unittest.TestCase):
         self.assertEqual(generate(parse_one("x is null")), "scope[None][x] is None")
 
     def test_optimized_tpch(self):
-        for i, (sql, optimized) in enumerate(self.sqls[:20], start=1):
+        for i, (sql, optimized) in enumerate(self.sqls, start=1):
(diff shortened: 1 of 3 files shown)


tests/test_executor.py
```diff
@@ -66,10 +66,10 @@ def test_py_dialect(self):
         self.assertEqual(generate(parse_one("x is null")), "scope[None][x] is None")
 
     def test_optimized_tpch(self):
-        for i, (sql, optimized) in enumerate(self.sqls[:20], start=1):
+        for i, (sql, optimized) in enumerate(self.sqls, start=1):
             with self.subTest(f"{i}, {sql}"):
                 a = self.cached_execute(sql)
-                b = self.conn.execute(optimized).fetchdf()
+                b = self.conn.execute(parse_one(optimized).sql(dialect="duckdb")).fetchdf()
```
georgesittas, line 72: @tobymao we may need a bit more work to make this complete, because `conn.execute(...)` is called in other code paths as well and the SQL we pass it isn't necessarily generated for DuckDB to my understanding. What do you think?

tobymao (approved):

PR #2985, opened 2024-02-17
feat(snowflake): create storage integration 

tekumara:
parse as a `Create` expression.

https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration


tests/dialects/test_snowflake.py
```diff
@@ -1548,6 +1548,29 @@ def test_show(self):
...
         self.assertEqual(users_exp.this, "USERS")
 
+    def test_storage_integration(self):
+        ast = parse_one(
+            """CREATE STORAGE INTEGRATION s3_int
+TYPE=EXTERNAL_STAGE
+STORAGE_PROVIDER='S3'
+STORAGE_AWS_ROLE_ARN='arn:aws:iam::001234567890:role/myrole'
+ENABLED=TRUE
+STORAGE_ALLOWED_LOCATIONS=('s3://mybucket1/path1/', 's3://mybucket2/path2/')""",
```
tekumara, line 1558: Some of these properties are mandatory. I wasn't sure if I should be creating a Properties subclass for them (or something else)?

georgesittas (approved):
Thanks @tekumara 👍

@tobymao perhaps we could consider refactoring the creatables so they're strings and we can avoid adding tokens for each new DDL.

tests/dialects/test_snowflake.py
georgesittas: Should be good if this is already parseable into a `Property` due to them all following this key = value syntax. We usually define new properties if there's custom / unrecognized syntax for them, so we can also map them back to that syntax later on.

tobymao:
@georgesittas yea let’s do that after

tests/dialects/test_snowflake.py
```diff
@@ -1548,6 +1548,29 @@ def test_show(self):
...
+            read="snowflake",
+        )
+        assert isinstance(ast, exp.Create)
```
georgesittas, line 1561: Let's also test that generating `ast` results in the original statement. Other than that, should be good to go.

tobymao: i don’t like this ast test, can we just test identity

tekumara: I can add back the [identity test I had here](https://github.com/tobymao/sqlglot/pull/2985/commits/9ebbfe45ab0d231db3e6f342c875e32680e670eb). I removed it because it passes without my changes, and doesn't test that the statement is parsed as an `exp.Create`

georgesittas: We now warn when a statement or expression gets parsed into a `Command`, so as long as there are no warnings when running the test suite, I guess that should be a good-enough signal - at least for these statements where it's easy to see what they're being parsed into.

tekumara:
Thanks @georgesittas for accepting and merging this. I'll wait until after the refactor to add CREATE STAGE and CREATE SEQUENCE.

PR #2989, opened 2024-02-18
feat(snowflake): add basic support for adding UNIQUE column constraints

whummer:
First of all - big kudos for providing this awesome library! 🙌 🚀 

This PR adds basic support for adding `UNIQUE` column constraints in `ALTER TABLE` statements for Snowflake: https://docs.snowflake.com/en/sql-reference/sql/alter-table

Example:
```
ALTER TABLE authors ADD CONSTRAINT c1 UNIQUE (id, email)
```

---
This is my first contribution in this repo, so not sure if the implementation is appropriate, tests are in the right place, etc - very happy for any feedback. Thanks!


tests/test_transpile.py
```diff
@@ -747,7 +747,6 @@ def test_command_identity(self):
             "ALTER SEQUENCE IF EXISTS baz RESTART WITH boo",
             "ALTER SESSION SET STATEMENT_TIMEOUT_IN_SECONDS=3",
             "ALTER TABLE integers DROP PRIMARY KEY",
-            "ALTER TABLE s_ut ADD CONSTRAINT s_ut_uq UNIQUE hajo",
```
whummer, line 750: Removed this test case from here, as this syntax is now supported (and this query no longer gets parsed as an `exp.Command` but as an `exp.AlterTable` expression now). Please confirm if this change is appropriate @georgesittas

georgesittas:
@whummer thanks for the PR - will take a look tomorrow.

tests/test_transpile.py
georgesittas: Yep, this change looks reasonable. Thanks for double-checking it.

tests/dialects/test_snowflake.py
```diff
@@ -519,6 +519,12 @@ def test_snowflake(self):
                 "snowflake": "ALTER TABLE foo ADD COLUMN id INT AUTOINCREMENT START 1 INCREMENT 1",
             },
         )
+        self.validate_all(
+            "ALTER TABLE foo ADD CONSTRAINT c1 UNIQUE (email)",
+            write={
+                "snowflake": "ALTER TABLE foo ADD CONSTRAINT c1 UNIQUE (email)",
+            },
+        )
```
georgesittas, line 527: We usually write tests like this, where the read and write dialects are the same, using `validate_identity`:

```python
self.validate_identity("ALTER TABLE foo ADD CONSTRAINT c1 UNIQUE (email)")
```

Let's make this change and move it appropriately so it's grouped with the rest of the `identity` checks (nit).

tests/dialects/test_snowflake.py
```diff
@@ -1599,3 +1605,18 @@ def test_try_cast(self):
...
+        query = "ALTER TABLE authors ADD CONSTRAINT c1 UNIQUE (email)"
+        alter = parse_one(query, read="snowflake")
+        self.assertIsInstance(alter, exp.AlterTable)
+        self.assertEqual(query, alter.sql())
+
+        # test parsing with multiple unique columns
+        query = "ALTER TABLE authors ADD CONSTRAINT c1 UNIQUE (id, email)"
+        alter = parse_one(query, read="snowflake")
+        self.assertIsInstance(alter, exp.AlterTable)
+        self.assertEqual(query, alter.sql())
```
georgesittas, line 1622: I believe we lately have been avoiding adding tests like this one, right @tobymao? Can probably just have a couple of identity tests.

Other than that, one observation is that it's good practice to provide `dialect="snowflake"` in these `sql()` calls.

georgesittas:
Closing this in favor of #2990, feel free to check it out. Again, thanks for the PR regardless! :-)

whummer:
Awesome, thanks @georgesittas  for the detailed review, and for taking over the polished implementation! 🚀

PR #2999, opened 2024-02-21
Feat: support LEFT JOIN UNNEST in duckdb

r1b:
Sample test cases:

Screenshot 2024-02-20 at 8 49 48 PM
Screenshot 2024-02-20 at 8 48 05 PM

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -540,11 +540,11 @@ class DuckDB(Dialect):
         def join_sql(self, expression: exp.Join) -> str:
             if (
                 expression.side == "LEFT"
-                and expression.args.get("on") is None
+                and not expression.args.get("on")
                 and isinstance(expression.this, exp.Unnest)
             ):
                 # Some dialects support `LEFT JOIN UNNEST(...)` without an explicit ON clause
                 # DuckDB doesn't, but we can just add a dummy ON clause that is always true
-                return super().join_sql(expression.on(exp.Literal.number(1)))
+                return super().join_sql(expression.on(exp.true()))
 
             return super().join_sql(expression)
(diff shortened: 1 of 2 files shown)


tobymao:
i don’t understand the point of this pr. why do you need to add on 1?

tobymao:
if the goal is to get bigquery to translate to duckdb, this pr is not correct, this should be done purely on generation

r1b:
Sorry, should have provided more background here - duckdb parser does not appear to like `LEFT JOIN` without an ON clause:

Screenshot 2024-02-20 at 10 06 17 PM

> if the goal is to get bigquery to translate to duckdb, this pr is not correct, this should be done purely on generation

This is the goal - can you clarify what you mean by "on generation"?

tobymao:
sqlglot had three parts tokenized, parser, generator. you should only modify the generator

r1b:
Thanks - moved this logic to the generator

sqlglot/dialects/duckdb.py
```diff
@@ -536,3 +536,15 @@ def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
             if isinstance(expression.parent, exp.UserDefinedFunction):
                 return self.sql(expression, "this")
             return super().columndef_sql(expression, sep)
+
+        def join_sql(self, expression: exp.Join) -> str:
+            if (
+                expression.side == "LEFT"
+                and expression.args.get("on") is None
```
tobymao, line 543:
```suggestion
                and not expression.args.get("on")
```

sqlglot/dialects/duckdb.py
```diff
@@ -536,3 +536,15 @@ def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
...
+                and isinstance(expression.this, exp.Unnest)
+            ):
+                # Some dialects support `LEFT JOIN UNNEST(...)` without an explicit ON clause
+                # DuckDB doesn't, but we can just add a dummy ON clause that is always true
+                return super().join_sql(expression.on(exp.Literal.number(1)))
```
tobymao, line 548:
```suggestion
                return super().join_sql(expression.on(exp.true()))
```

tobymao:
thanks for the pr!

PR #3002, opened 2024-02-21
Feat: add support for more Snowflake SHOW commands

DanCardin:
`SHOW IMPORTED KEYS` and `SHOW PRIMARY KEYS` appear to be undocumented (as far as i can tell), but definitely work.

The full set being added here are those which are referenced in `snowflake-sqlalchemy`, but dont already exist here.


sqlglot/dialects/snowflake.py
```diff
@@ -603,7 +611,9 @@ def _parse_show_snowflake(self, this: str) -> exp.Show:
                     if self._curr:
                         scope = self._parse_table_parts()
                 elif self._curr:
-                    scope_kind = "SCHEMA" if this in ("OBJECTS", "TABLES") else "TABLE"
+                    scope_kind = (
+                        "SCHEMA" if this in {"OBJECTS", "TABLES", "VIEWS", "SEQUENCES"} else "TABLE"
```
georgesittas, line 615: Let's factor this set out into a class constant, e.g. `SCHEMA_KINDS`

tests/dialects/test_snowflake.py
```diff
@@ -1549,6 +1549,81 @@ def test_show(self):
         self.assertTrue(isinstance(users_exp, exp.Show))
         self.assertEqual(users_exp.this, "USERS")
 
+    def show_views(self):
```
georgesittas, line 1552: This isn't a valid test method name, you need the `test_` prefix.

But, ignoring this point, let's just consolidate all these new methods into `test_show` since it already exists. Also, it looks like a lot of these tests are duplicated and basically the only thing that differs is the keyword following `SHOW`. It would be awesome if you could clean this up a bit with a for loop or something, otherwise I'll do it in another PR.

georgesittas (approved):
LGTM, just a couple of comments. Appreciate the PRs, @DanCardin 👍

tests/dialects/test_snowflake.py
```diff
@@ -1464,90 +1464,69 @@ def test_match_recognize(self):
...
-        self.validate_identity(
-            "show terse tables in db1.schema1 starts with 'a' limit 10 from 'b'",
-            "SHOW TERSE TABLES IN SCHEMA db1.schema1 STARTS WITH 'a' LIMIT 10 FROM 'b'",
-        )
-
-        ast = parse_one('SHOW PRIMARY KEYS IN "TEST"."PUBLIC"."customers"', read="snowflake")
-        table = ast.find(exp.Table)
-
-        self.assertEqual(table.sql(dialect="snowflake"), '"TEST"."PUBLIC"."customers"')
+        show_kinds = [
```
DanCardin, line 1467: So I dont know if this is what you were expecting or not. Most of each show variant has slightly different supported options, defaults and whatnot. (I can undo this commit and let you refactor it yourself, if this isn't what you were going for)

I couldn't see a way to make the tests more compact with a loop, other than by categorizing each differing variant and using that to choose what to do.

Arguably this is a decent amount less easy to read/follow for each individual option, but it's certainly more compact.

---

I guess also relevant, is that the parser doesn't currently distinguish between the different options in a bunch of ways. The availability of `IN TABLE` as an option, `HISTORY` being a supported option, etc.

Those seem like preexisting issues that might one day be worth revising (or not, if we only care about the max syntax), but i figured i'd point it out.

But as such, I just opted to not test those variants for the options that dont appear to support them IRL.

georgesittas: I see what you mean, maybe you're right that splitting them across different methods could be easier to read / follow. It seems that's what we already do in [MySQL's case](https://github.com/tobymao/sqlglot/blob/main/tests/dialects/test_mysql.py).

Apologies for the back and forth, but if possible let's revert to the previous commit like you suggested (keep the constant, though), and I can clean this up by refactoring the existing `test_show` as well.

Appreciate the quick turnaround on this :)

DanCardin: no problem! done!

Issue #2988, opened 2024-02-18
Failed to parse TSQL query with 'OPTION' statement


ShayYaari:
Hi!

I'm trying to use sqlglot to parse the following query:

SELECT a FROM b OPTION (MAXDOP 1)

using Dialect.TSQL.

It fails with the following error:

`
sqlglot.errors.ParseError: Expecting ). Line 2, Col: 32.
`

Any suggestions?

Thanks!

georgesittas:
Hey, this isn't high priority right now so we don't plan to implement it. Feel free to make a PR if you need this - my guess is that the `OPTION` should be supported as a "query modifier", at least only for T-SQL (unless other dialects have something similar). If you need any guidance, feel free to reach out in [Slack](https://tobikodata.com/slack).

georgesittas:
Fixed in https://github.com/tobymao/sqlglot/commit/51f8d5897b18e6f7c0bc66881a3e36c8842ff2ff by @nadav-botanica!

PR #3003, opened 2024-02-21
Feat(tsql): added support for parsing the OPTION clause

nadav-botanica:


georgesittas:
Hey @nadav-botanica, thanks a lot for this PR - I can tell you've put a lot of effort in making it.

Having said that, this doesn't look like the right approach for supporting `OPTION` in T-SQL. I'll need to dig deeper to understand the nuances, but having to add `option` in all these different AST nodes and overriding core methods such as `_parse_select` etc just to anticipate this clause seems like a smell.

As I said in a recent [issue](https://github.com/tobymao/sqlglot/issues/2988#issuecomment-1951229407), my guess is that `OPTION` needs to be treated as a "query modifier" (using sqlglot's lingo), so you have to touch as few places as possible in the code.

nadav-botanica:
The current parsing implemented in sql doesn't support the OPTION clause, as specified here: https://learn.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=sql-server-ver16
This patch aims to add support

georgesittas:
Can you try to refactor this based on the above suggestions? We can keep iterating in this PR if you want. Feel free to reach out in [Slack](https://tobikodata.com/slack) if you need assistance.

nadav-botanica:
Hi @georgesittas - thanks for your rapid response and elaborate answer!
I had a look at `_parse_query_modifiers`, and while it does look to fit for the case of adding support for `OPTION` in case of `SELECT` clauses, I had some problems helping it work for `DELETE` and `UPDATE` statements.
I would greatly appreciate your input!

georgesittas:
> Hi @georgesittas - thanks for your rapid response and elaborate answer! I had a look at `_parse_query_modifiers`, and while it does look to fit for the case of adding support for `OPTION` in case of `SELECT` clauses, I had some problems helping it work for `DELETE` and `UPDATE` statements. I would greatly appreciate your input!

Hmm, perhaps we could keep the `option` arg in `Delete` and `Update` and factor out the `OPTION` modifier parsing logic into a method so that it can be reused for them?

georgesittas:
Looks like it will need some more work to refactor this PR, so I'm going to close it for now. @nadav-botanica we're probably gonna tackle this one soon too, but in case you've already made progress feel free to open another PR (probably reach out in Slack in that case).

PR #3005, opened 2024-02-21
Refactor!: make implicit unnest syntax explicit by using UNNEST calls

### LINKED ISSUE #2996: Bigquery implicit unnest transpile error
**Before you file an issue**

```
>>> import sqlglot
>>> sqlglot.transpile("""WITH Coordinates AS (SELECT [1,2] AS position)
... SELECT results FROM Coordinates, Coordinates.position AS results;
'WITH Coordinates AS (SELECT ARRAY[1, 2] AS position) SELECT results FROM Coordinates, Coordinates.position AS results'
```
Bigquery has implicit unnest, while the presto does not have, so this presto sql is wrong

**Official Documentation**
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#implicit_unnesting

georgesittas:
Fixes #2996

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -70,6 +70,8 @@ class Redshift(Postgres):
             "SYSDATE": lambda self: self.expression(exp.CurrentTimestamp, transaction=True),
         }
 
+        SUPPORTS_IMPLICIT_UNNEST = True
+
         def _parse_table(
             self,
             schema: bool = False,
@@ -123,27 +125,6 @@ class Redshift(Postgres):
                 return self.expression(exp.ApproxDistinct, this=seq_get(func.this.expressions, 0))
             self._retreat(index)
             return None
-
-        def _parse_query_modifiers(
-            self, this: t.Optional[exp.Expression]
-        ) -> t.Optional[exp.Expression]:
-            this = super()._parse_query_modifiers(this)
-
-            if this:
-                refs = set()
-
-                for i, join in enumerate(this.args.get("joins", [])):
-                    refs.add(
-                        (
-                            this.args["from"] if i == 0 else this.args["joins"][i - 1]
-                        ).this.alias.lower()
-                    )
-
-                    table = join.this
-                    if isinstance(table, exp.Table) and not join.args.get("on"):
-                        if table.parts[0].name.lower() in refs:
-                            table.replace(table.to_column())
-            return this
 
     class Tokenizer(Postgres.Tokenizer):
         BIT_STRINGS = []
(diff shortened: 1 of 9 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -225,6 +206,18 @@ class Generator(Postgres.Generator):
...
 
+        def unnest_sql(self, expression: exp.Unnest) -> str:
+            args = expression.expressions
+            num_args = len(args)
+
+            if num_args > 1:
+                self.unsupported(f"Unsupported number of arguments in UNNEST: {num_args}")
+                return ""
+
+            arg = self.sql(args[0])
```
tobymao, line 217: what if args is 0? maybe use seq_get

sqlglot/parser.py
```diff
@@ -2452,6 +2455,25 @@ def _parse_subquery(
             alias=self._parse_table_alias() if parse_alias else None,
         )
 
+    def _implicit_unnests_to_explicit(self, this: E) -> E:
+        refs = {this.args["from"].this.alias_or_name.lower()}
+        for i, join in enumerate(this.args.get("joins") or []):
+            table = join.this
+            if isinstance(table, exp.Table) and not join.args.get("on"):
+                if table.parts[0].name.lower() in refs:
```
tobymao, line 2463: why are you using name.lower, shouldn't we normalize the names with the dialect?

sqlglot/dialects/redshift.py
```diff
@@ -124,27 +126,6 @@ def _parse_approximate_count(self) -> t.Optional[exp.ApproxDistinct]:
             self._retreat(index)
             return None
 
-        def _parse_query_modifiers(
```
tobymao, line 127: why was this one removed?

sqlglot/parser.py
georgesittas: we probably should, though note that this is how it was previously implemented in redshift

sqlglot/dialects/redshift.py
georgesittas: it was moved to the base class and altered slightly - now controlled with a class constant because BQ's logic is similar

sqlglot/parser.py
tobymao: yea but redshift, everything is lower case so that makes sense :)

georgesittas: ah lol, 2 steps ahead

georgesittas: makes sense - will refactor thanks for catching

georgesittas: There's an annoying edge case in BQ if we normalize the identifiers.. we currently use a [heuristic](https://github.com/tobymao/sqlglot/blob/984049648991b8b1f89827ec35b29da603e6fb64/sqlglot/dialects/bigquery.py#L272) to check whether a table ref. should be normalized which is to check if it's qualified. The reason is that CTEs are case-insensitive while tables usually aren't. This breaks the following test:

```sql
SELECT results FROM Coordinates, Coordinates.position AS results;
SELECT results FROM c.db.Coordinates AS Coordinates, UNNEST(Coordinates.position) AS results;
```

because the first `Coordinates` ref. gets normalized, since it doesn't have a `db` set, but the second `Coordinates` in `Coordinates.position` _doesn't_, because it _does_ have a `db` set. So the two don't match when we check the `refs` set and it's not converted into a column.

tobymao (approved):

PR #3010, opened 2024-02-22
Feat: Supporting RANGE <-> GENERATE_SERIES between DuckDB & SQLite

VaggelisD:
Hi,

The following PR aims to address [discussion#2995 ](https://github.com/tobymao/sqlglot/discussions/2995)

Key changes:

- SQLite now explicitly rejects table alias columns. In order to implement the sample query from the discussion which uses this DuckDB feature, SQLite now wraps the output of `exp.GenerateSeries` in a subquery with the proper (i.e first) column reference if a TableAlias is binding to the function call. 

- The AST node `exp.GenerateSeries` now carries one more argument (state) to mark whether the end is inclusive/exclusive, as the same function appears on other dialects and/or in other flavors e.g. DuckDB supports `RANGE` (exclusive end) _and_ `GENERATE_SERIES` (inclusive end). 

Future PRs should hopefully: 
- Aim to support the flavors of `GENERATE_SERIES / RANGE` across all dialects through the `exp.GenerateSeries`
- Transform the range appropriately (wherever possible) e.g in DuckDB the proper transpilation should be `RANGE(1, 5) -> GENERATE_SERIES(1, 4)` to match the data output

Docs
---------
- [DuckDB Range Functions](https://duckdb.org/docs/sql/functions/nested.html#range-functions)
- [SQLite GENERATE_SERIES Function](https://www.sqlite.org/series.html)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -79,6 +79,24 @@ def _build_date_diff(args: t.List) -> exp.Expression:
     return exp.DateDiff(this=seq_get(args, 2), expression=seq_get(args, 1), unit=seq_get(args, 0))
 
 
+def _build_range(args: t.List, is_end_exclusive: bool = False) -> exp.Expression:
+    # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
+    # for the rules that follow
+
+    if len(args) == 1:
+        # DuckDB allows the creation of ranges with only a single argument
+        # being the "stop", thus we must canonicalize the form
+        # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
+        start = exp.Literal.number("0")
+        stop = seq_get(args, 0)
+        args = [start, stop]
+
+    gen_series = exp.GenerateSeries.from_arg_list(args)
+    gen_series.args["is_end_exclusive"] = is_end_exclusive
+
+    return gen_series
+
+
 def _build_make_timestamp(args: t.List) -> exp.Expression:
     if len(args) == 1:
         return exp.UnixToTime(this=seq_get(args, 0), scale=exp.UnixToTime.MICROS)
@@ -267,6 +285,8 @@ class DuckDB(Dialect):
             "TO_TIMESTAMP": exp.UnixToTime.from_arg_list,
             "UNNEST": exp.Explode.from_arg_list,
             "XOR": binary_from_function(exp.BitwiseXor),
+            "GENERATE_SERIES": lambda args: _build_range(args, False),
+            "RANGE": lambda args: _build_range(args, True),
         }
 
         FUNCTION_PARSERS = parser.Parser.FUNCTION_PARSERS.copy()
@@ -548,3 +568,18 @@ class DuckDB(Dialect):
                 return super().join_sql(expression.on(exp.true()))
 
             return super().join_sql(expression)
+
+        def generateseries_sql(self, expression: exp.GenerateSeries) -> str:
+            is_end_exclusive = expression.args.get("is_end_exclusive")
+
+            # DuckDB has the following functions:
+            # - Generate_Series(a, b) -> generates [a, b]
+            # - Range(a, b) -> generates [a, b)
+            if is_end_exclusive:
+                start = expression.args.get("start")
+                end = expression.args.get("end")
+                step = expression.args.get("step")
+
+                return self.func("RANGE", start, end, step)
+
+            return super().generateseries_sql(expression)
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -81,15 +81,9 @@ def _build_date_diff(args: t.List) -> exp.Expression:
 
 def _build_range(args: t.List, is_end_exclusive: bool = False) -> exp.Expression:
     # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
-    # for the rules that follow
-
     if len(args) == 1:
-        # DuckDB allows the creation of ranges with only a single argument
-        # being the "stop", thus we must canonicalize the form
-        # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
-        start = exp.Literal.number("0")
-        stop = seq_get(args, 0)
-        args = [start, stop]
+        # DuckDB uses 0 as a default for the series' start when it's omitted
+        args.insert(0, exp.Literal.number("0"))
 
     gen_series = exp.GenerateSeries.from_arg_list(args)
     gen_series.args["is_end_exclusive"] = is_end_exclusive
@@ -570,16 +564,9 @@ class DuckDB(Dialect):
             return super().join_sql(expression)
 
         def generateseries_sql(self, expression: exp.GenerateSeries) -> str:
-            is_end_exclusive = expression.args.get("is_end_exclusive")
-
-            # DuckDB has the following functions:
-            # - Generate_Series(a, b) -> generates [a, b]
-            # - Range(a, b) -> generates [a, b)
-            if is_end_exclusive:
-                start = expression.args.get("start")
-                end = expression.args.get("end")
-                step = expression.args.get("step")
-
-                return self.func("RANGE", start, end, step)
+            # GENERATE_SERIES(a, b) -> [a, b], RANGE(a, b) -> [a, b)
+            if expression.args.get("is_end_exclusive"):
+                expression.set("is_end_exclusive", None)
+                return rename_func("RANGE")(self, expression)
 
             return super().generateseries_sql(expression)
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -79,16 +79,19 @@ def _build_date_diff(args: t.List) -> exp.Expression:
     return exp.DateDiff(this=seq_get(args, 2), expression=seq_get(args, 1), unit=seq_get(args, 0))
 
 
-def _build_range(args: t.List, is_end_exclusive: bool = False) -> exp.Expression:
-    # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
-    if len(args) == 1:
-        # DuckDB uses 0 as a default for the series' start when it's omitted
-        args.insert(0, exp.Literal.number("0"))
+def _build_generate_series(end_exclusive: bool = False) -> t.Callable[[t.List], exp.GenerateSeries]:
+    def _builder(args: t.List) -> exp.GenerateSeries:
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -79,6 +79,24 @@ def _build_date_diff(args: t.List) -> exp.Expression:
...
 
 
+def _build_range(args: t.List, is_end_exclusive: bool = False) -> exp.Expression:
+    # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
+    # for the rules that follow
+
+    if len(args) == 1:
+        # DuckDB allows the creation of ranges with only a single argument
+        # being the "stop", thus we must canonicalize the form
+        # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
```
georgesittas, line 89:
```suggestion
        # DuckDB uses 0 as a default for the series' start when it's omitted
```

sqlglot/dialects/duckdb.py
```diff
@@ -79,6 +79,24 @@ def _build_date_diff(args: t.List) -> exp.Expression:
...
+        start = exp.Literal.number("0")
+        stop = seq_get(args, 0)
+        args = [start, stop]
```
georgesittas, line 92:
```suggestion
        args.insert(0, exp.Literal.number("0"))
```

sqlglot/dialects/duckdb.py
```diff
@@ -548,3 +568,18 @@ def join_sql(self, expression: exp.Join) -> str:
...
+
+            # DuckDB has the following functions:
+            # - Generate_Series(a, b) -> generates [a, b]
+            # - Range(a, b) -> generates [a, b)
+            if is_end_exclusive:
+                start = expression.args.get("start")
+                end = expression.args.get("end")
+                step = expression.args.get("step")
+
+                return self.func("RANGE", start, end, step)
```
georgesittas, line 583:
```suggestion
            # GENERATE_SERIES(a, b) -> [a, b], RANGE(a, b) -> [a, b)
            if expression.args.get("is_end_exclusive"):
                expression.set("is_end_exclusive", None)
                return rename_func("RANGE")(self, expression)
```

sqlglot/dialects/sqlite.py
```diff
@@ -92,6 +92,7 @@ class Generator(generator.Generator):
         NVL2_SUPPORTED = False
         JSON_PATH_BRACKETED_KEY_SUPPORTED = False
         SUPPORTS_CREATE_TABLE_LIKE = False
+        SUPPORTS_TABLE_ALIAS_COLUMNS = False
```
georgesittas, line 95: Let's test this, i.e. show how transpiling something like `SELECT * FROM t(c)` to SQLite (a) generates `SELECT * FROM t` and (b) logs a warning (look for the `self.assertLogs` tests to see how (b) is implemented).

sqlglot/dialects/sqlite.py
```diff
@@ -173,6 +174,18 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
...
+            parent = expression.parent
+
+            if parent:
+                alias = parent.args.get("alias")
+                sql = super().generateseries_sql(expression)
+                if alias:
+                    first_column_ref = alias.columns[0]
+                    alias.args.get("columns").pop()
+                    return f"(SELECT value AS {self.sql(first_column_ref, 'this')} FROM {sql})"
+            return sql
```
georgesittas, line 187: This will result in a `NameError` if there's no `parent` because `sql` is undefined.

Regardless, we avoid hard-coding SQL like this when possible, because for example it can't be pretty-formatted if someone passes `pretty=True` to the generator.

An alternative way to implement this transformation:

```python
parent = expression.parent
alias = parent and parent.args.get("alias")

if isinstance(alias, exp.TableAlias) and alias.columns:
    column_alias = alias.columns[0]
    alias.set("columns", None)
    sql = self.sql(exp.select(exp.alias_("value", column_alias)).from_(expression).subquery())
else:
    sql = super().generateseries_sql(self, expression)

return sql
```

georgesittas (commented):
Great work - left some comments.

sqlglot/dialects/sqlite.py
VaggelisD: Ack

sqlglot/dialects/duckdb.py
georgesittas: You're making a very valid point, but the `Generator` [copies ASTs by default](https://github.com/tobymao/sqlglot/blob/2a88e40da89fa083bbd8fd0174082fa8e677780a/sqlglot/generator.py#L542), so mutations aren't a concern.

sqlglot/dialects/sqlite.py
VaggelisD: Done

sqlglot/generator.py
```diff
@@ -3454,3 +3454,10 @@ def _ensure_string_if_null(self, values: t.List[exp.Expression]) -> t.List[exp.E
             for value in values
             if value
         ]
+
+    def generateseries_sql(self, expression: exp.GenerateSeries) -> str:
+        start = expression.args.get("start")
+        end = expression.args.get("end")
+        step = expression.args.get("step")
+
+        return self.func("GENERATE_SERIES", start, end, step)
```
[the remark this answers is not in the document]
VaggelisD: Done

tests/dialects/test_duckdb.py
```diff
@@ -626,6 +625,26 @@ def test_duckdb(self):
...
 
+        self.validate_identity("SELECT * FROM RANGE(1, 5, 10)")
+        self.validate_identity("SELECT * FROM GENERATE_SERIES(2, 13, 4)")
+
+        self.validate_all(
+            "WITH t AS (SELECT i, i*i*i*i*i AS i5 FROM RANGE(1,5) t(i)) SELECT * FROM t",
+            write={
+                "sqlite": "WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM (SELECT value AS i FROM GENERATE_SERIES(1, 5)) AS t) SELECT * FROM t",
+            },
+        )
```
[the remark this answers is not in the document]
VaggelisD: Ack

sqlglot/dialects/duckdb.py
```diff
@@ -79,6 +79,24 @@ def _build_date_diff(args: t.List) -> exp.Expression:
     return exp.DateDiff(this=seq_get(args, 2), expression=seq_get(args, 1), unit=seq_get(args, 0))
 
 
+def _build_range(args: t.List, is_end_exclusive: bool = False) -> exp.Expression:
```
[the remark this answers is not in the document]
VaggelisD: Ack

tests/dialects/test_duckdb.py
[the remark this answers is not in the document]
VaggelisD: Two issues:

- The DuckDB roundtrip produces an extra `AS` keyword, i.e `RANGE(1) t(i) -> RANGE(1) AS t(i)` but semantically it's the same
- For the reverse operation (SQLite output statement transpiled back to DuckDB) there is an error as DuckDB does not use "value" as the default column name but the function name. For example, this is viable DuckDB:

`SELECT t.range FROM RANGE(1, 5) AS t`

I haven't come up with a trivial solution to fixing the reverse operation yet, if it's not needed at present I can rethink it and open a new PR to solve it.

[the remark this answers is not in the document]
georgesittas: Sounds good 👍
Feel free to mark this as resolved when done adding the duckdb -> duckdb test.

sqlglot/dialects/sqlite.py
VaggelisD: Done, introduced that test case to `test_sqlite.py` in accordance to similar checks in `test_bigquery.py`

tests/dialects/test_duckdb.py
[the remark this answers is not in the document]
VaggelisD: Done

tests/dialects/test_duckdb.py
```diff
@@ -626,6 +625,31 @@ def test_duckdb(self):
...
+            "WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM RANGE(1, 5) t(i)) SELECT * FROM t",
+            write={
+                "sqlite": "WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM (SELECT value AS i FROM GENERATE_SERIES(1, 5)) AS t) SELECT * FROM t",
+            },
+        )
+
+        self.validate_identity(
+            """WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM RANGE(1, 5) t(i)) SELECT * FROM t""",
+            """WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM RANGE(1, 5) AS t(i)) SELECT * FROM t""",
+        )
```
georgesittas, line 641: No need to make another test here, we can put it in the above `validate_all`, like:

```python
        self.validate_all(
            "WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM RANGE(1, 5) t(i)) SELECT * FROM t",
            write={
                "duckdb": "WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM RANGE(1, 5) AS t(i)) SELECT * FROM t",
                "sqlite": "WITH t AS (SELECT i, i * i * i * i * i AS i5 FROM (SELECT value AS i FROM GENERATE_SERIES(1, 5)) AS t) SELECT * FROM t",
            },
        )
```

VaggelisD: Done

sqlglot/dialects/duckdb.py
```diff
@@ -79,6 +79,21 @@ def _build_date_diff(args: t.List) -> exp.Expression:
...
 
+def _build_generate_series(end_exclusive: bool = False) -> t.Callable[[t.List], exp.GenerateSeries]:
+    def _builder(args: t.List) -> exp.GenerateSeries:
+        # Check https://duckdb.org/docs/sql/functions/nested.html#range-functions
+        if len(args) == 1:
+            # DuckDB uses 0 as a default for the series' start when it's omitted
+            args.insert(0, exp.Literal.number("0"))
+
+        gen_series = exp.GenerateSeries.from_arg_list(args)
+        gen_series.args["is_end_exclusive"] = end_exclusive
```
georgesittas, line 90:
```suggestion
        gen_series.set("is_end_exclusive", end_exclusive)
```

georgesittas (approved):
LGTM

PR #3017, opened 2024-02-23
fix!: refactor structs to always be property eqs

### LINKED ISSUE #3015: wrong transformation from bigquery struct to presto row
**Before you file an issue**

```
>>> sqlglot.transpile("""select struct(true as sign, 1 as value)
'SELECT ROW(TRUE AS sign, 1 AS value)'
```
and
```
trino> SELECT ROW(TRUE AS sign, 1 AS value)
    -> ;
Query 20240129_170954_00005_56ath failed: line 1:17: mismatched input 'AS'. Expecting: '%', ')', '*', '+', ',', '-', '.', '/', 'AND', 'AT', 'OR', 'ORDER', '[', '||', <predicate>
SELECT ROW(TRUE AS sign, 1 AS value)
```

**Official Documentation**

tobymao:
closes #3015, allowing presto to have named structs if types are known

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -631,3 +631,15 @@ class Hive(Dialect):
         def version_sql(self, expression: exp.Version) -> str:
             sql = super().version_sql(expression)
             return sql.replace("FOR ", "", 1)
+
+        def struct_sql(self, expression: exp.Struct) -> str:
+            values = []
+
+            for i, e in enumerate(expression.expressions):
+                if isinstance(e, exp.PropertyEQ):
+                    self.unsupported("Hive does not support named structs.")
+                    values.append(e.expression)
+                else:
+                    values.append(e)
+
+            return self.func("STRUCT", *values)
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -263,14 +263,9 @@ class Spark2(Hive):
         CREATE_FUNCTION_RETURN_AS = False
 
         def struct_sql(self, expression: exp.Struct) -> str:
-            args = []
-            for arg in expression.expressions:
-                if isinstance(arg, self.KEY_VALUE_DEFINITIONS):
-                    args.append(exp.alias_(arg.expression, arg.this.name))
-                else:
-                    args.append(arg)
+            from sqlglot.generator import Generator
 
-            return self.func("STRUCT", *args)
+            return Generator.struct_sql(self, expression)
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
             if is_parse_json(expression.this):
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -875,6 +875,8 @@ class Parser(metaclass=_Parser):
 
     FUNCTIONS_WITH_ALIASED_ARGS = {"STRUCT"}
 
+    KEY_VALUE_DEFINITIONS = (exp.Alias, exp.EQ, exp.PropertyEQ, exp.Slice)
+
     FUNCTION_PARSERS = {
         "CAST": lambda self: self._parse_cast(self.STRICT_CAST),
         "CONVERT": lambda self: self._parse_convert(self.STRICT_CAST),
@@ -4112,6 +4114,9 @@ class Parser(metaclass=_Parser):
             alias = upper in self.FUNCTIONS_WITH_ALIASED_ARGS
             args = self._parse_csv(lambda: self._parse_lambda(alias=alias))
 
+            if alias:
+                args = self._kv_to_prop_eq(args)
+
             if function and not anonymous:
                 if "dialect" in function.__code__.co_varnames:
                     func = function(args, dialect=self.dialect)
@@ -4131,6 +4136,26 @@ class Parser(metaclass=_Parser):
 
         self._match_r_paren(this)
         return self._parse_window(this)
+
+    def _kv_to_prop_eq(self, expressions: t.List[exp.Expression]) -> t.List[exp.Expression]:
+        transformed = []
+
+        for e in expressions:
+            if isinstance(e, self.KEY_VALUE_DEFINITIONS):
+                if isinstance(e, exp.Alias):
+                    e = self.expression(exp.PropertyEQ, this=e.args.get("alias"), expression=e.this)
+
+                if not isinstance(e, exp.PropertyEQ):
+                    e = self.expression(
+                        exp.PropertyEQ, this=exp.to_identifier(e.name), expression=e.expression
+                    )
+
+                if isinstance(e.this, exp.Column):
+                    e.this.replace(e.this.this)
+
+            transformed.append(e)
+
+        return transformed
 
     def _parse_function_parameter(self) -> t.Optional[exp.Expression]:
         return self._parse_column_def(self._parse_id_var())
@@ -4544,7 +4569,7 @@ class Parser(metaclass=_Parser):
 
         # https://duckdb.org/docs/sql/data_types/struct.html#creating-structs
         if bracket_kind == TokenType.L_BRACE:
-            this = self.expression(exp.Struct, expressions=expressions)
+            this = self.expression(exp.Struct, expressions=self._kv_to_prop_eq(expressions))
         elif not this or this.name.upper() == "ARRAY":
             this = self.expression(exp.Array, expressions=expressions)
         else:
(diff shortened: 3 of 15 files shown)


tests/dialects/test_bigquery.py
```diff
@@ -18,6 +18,17 @@ class TestBigQuery(Validator):
...
+            "SELECT STRUCT(1, 2, 3), STRUCT(), STRUCT('abc'), STRUCT(1, t.str_col), STRUCT(1 as a, 'abc' AS b), STRUCT(str_col AS abc)",
+            write={
+                "bigquery": "SELECT STRUCT(1, 2, 3), STRUCT(), STRUCT('abc'), STRUCT(1, t.str_col), STRUCT(1 AS a, 'abc' AS b), STRUCT(str_col AS abc)",
+                "duckdb": "SELECT {'_0': 1, '_1': 2, '_2': 3}, {}, {'_0': 'abc'}, {'_0': 1, '_1': t.str_col}, {'a': 1, 'b': 'abc'}, {'abc': str_col}",
+                "spark": "SELECT STRUCT(1, 2, 3), STRUCT(), STRUCT('abc'), STRUCT(1, t.str_col), STRUCT(1 AS a, 'abc' AS b), STRUCT(str_col AS abc)",
+                "snowflake": "SELECT OBJECT_CONSTRUCT('', 1, '', 2, '', 3), OBJECT_CONSTRUCT(), OBJECT_CONSTRUCT('', abc), OBJECT_CONSTRUCT('', 1, '', str_col), OBJECT_CONSTRUCT('a', 1, 'b', 'abc'), OBJECT_CONSTRUCT('abc', str_col)",
+                # struct is a fallback without type inference
+                "trino": "SELECT ROW(1, 2, 3), ROW(), ROW('abc'), ROW(1, t.str_col), CAST(ROW(1, 'abc') AS ROW(a INTEGER, b VARCHAR)), STRUCT(str_col AS abc)",
+            },
+        )
```
georgesittas, line 31: Note that:

- `STRUCT()` is generally not supported I think in these dialects, only Snowflake allowed `OBJECT_CONSTRUCT()`
- This is invalid Snowflake `OBJECT_CONSTRUCT('', 1, '', 2, '', 3)` due to duplicate keys

Not sure if we wanna do something about this, just observing.

tests/dialects/test_bigquery.py
```diff
@@ -603,11 +614,11 @@ def test_bigquery(self):
             read={"spark": "select posexplode_outer([])"},
         )
         self.validate_all(
-            "SELECT AS STRUCT ARRAY(SELECT AS STRUCT b FROM x) AS y FROM z",
+            "SELECT AS STRUCT ARRAY(SELECT AS STRUCT 1 AS b FROM x) AS y FROM z",
```
georgesittas, line 617: Did this need to change? I think both this and the previous one are valid BQ queries, but I'm curious if the old one can't be transpiled for some reason?

tests/dialects/test_duckdb.py
```diff
@@ -55,8 +55,15 @@ def test_duckdb(self):
...
-            self.validate_all(struct_value, write={"presto": UnsupportedError})
+        self.validate_all(
+            "{'a': 1, 'b': '2'}", write={"presto": "CAST(ROW(1, '2') AS ROW(a INTEGER, b VARCHAR))"}
+        )
+        self.validate_all(
+            "struct_pack(a := 1, b := 2)",
+            write={"presto": "CAST(ROW(1, 2) AS ROW(a INTEGER, b INTEGER))"},
+        )
+
+        self.validate_all("struct_pack(a := 1, b := x)", write={"presto": UnsupportedError})
```
georgesittas, line 66: let's also test the roundtrip to duckdb

tests/dialects/test_hive.py
```diff
@@ -611,12 +611,6 @@ def test_hive(self):
                 "spark": "GET_JSON_OBJECT(x, '$.name')",
             },
         )
-        self.validate_all(
-            "STRUCT(a = b, c = d)",
-            read={
-                "snowflake": "OBJECT_CONSTRUCT(a, b, c, d)",
-            },
-        )
```
georgesittas, line 619: why was this removed?

tests/dialects/test_snowflake.py
```diff
@@ -372,13 +372,13 @@ def test_snowflake(self):
             write={"snowflake": "SELECT * FROM (VALUES (0)) AS foo(bar)"},
         )
         self.validate_all(
-            "OBJECT_CONSTRUCT(a, b, c, d)",
+            "OBJECT_CONSTRUCT('a', b, 'c', d)",
```
georgesittas, line 375: fyi `OBJECT_CONSTRUCT(a, b)` is also valid snowflake:

```
-- yields { "foo": 1 }
WITH t(a) AS (SELECT 'foo')
SELECT OBJECT_CONSTRUCT(a, 1) from t
```

georgesittas:
Nice!

tests/dialects/test_hive.py
tobymao: didn't find any docs on hive struct

tests/dialects/test_bigquery.py
tobymao: the issue is if you do it it's an unnamed struct so i changed it

tests/dialects/test_hive.py
tobymao: added unnamed structs for hive, it doesn't look like it uses key value

sqlglot/parser.py
```diff
@@ -4132,6 +4137,26 @@ def _parse_function_call(
         self._match_r_paren(this)
         return self._parse_window(this)
 
+    def _kv_to_prop_eq(self, expressions: t.List[exp.Expression]) -> t.List[exp.Expression]:
+        transformed = []
+
+        for e in expressions:
+            if isinstance(e, self.KEY_VALUE_DEFINITIONS):
+                if isinstance(e, exp.Alias):
+                    e = self.expression(exp.PropertyEQ, this=e.args.get("alias"), expression=e.this)
```
georgesittas, line 4146:
```suggestion
                    e = self.expression(exp.PropertyEQ, this=e.args["alias"], expression=e.this)
```

sqlglot/parser.py
```diff
@@ -4132,6 +4137,26 @@ def _parse_function_call(
...
+
+                if not isinstance(e, exp.PropertyEQ):
+                    e = self.expression(
+                        exp.PropertyEQ, this=exp.to_identifier(e.name), expression=e.expression
+                    )
+
+                if isinstance(e.this, exp.Column):
+                    e.this.replace(e.this.this)
```
georgesittas, line 4154: Just double-checking, but we replace here so that we only keep the identifier right?

sqlglot/dialects/hive.py
```diff
@@ -631,3 +631,15 @@ def datatype_sql(self, expression: exp.DataType) -> str:
         def version_sql(self, expression: exp.Version) -> str:
             sql = super().version_sql(expression)
             return sql.replace("FOR ", "", 1)
+
+        def struct_sql(self, expression: exp.Struct) -> str:
+            values = []
+
+            for i, e in enumerate(expression.expressions):
+                if isinstance(e, exp.PropertyEQ):
+                    self.unsupported("Hive does not support named structs.")
```
georgesittas, line 640: I know this is gonna be annoying, but FWIW Spark v2 supports `STRUCT(1 AS a)` (so maybe this is also supported by Hive?) and also I found an [SO thread](https://stackoverflow.com/a/32386011) that shows a different function `select named_struct("a", 1)`, supposedly supported by Hive as well AFAICT. Works in both Spark v2 and v3.

sqlglot/dialects/spark2.py
```diff
@@ -293,5 +283,10 @@ def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
                 ),
             )
 
+        def struct_sql(self, expression: exp.Struct) -> str:
+            from sqlglot.generator import Generator
+
+            return Generator.struct_sql(self, expression)
```
georgesittas, line 289: Can also add this in `TRANSFORMS` if we wanna avoid defining this method? It should take precedence.

georgesittas (commented):
I think we may wanna add another test for snowflake to check that
```
OBJECT_CONSTRUCT(a, b, c, d)
```
can be roundtripped.

sqlglot/parser.py
tobymao: yea, we don't want a column, it will get qualified etc, it's an identifier

sqlglot/dialects/hive.py
tobymao: no, that's a plugin

georgesittas (approved):

PR #3018, opened 2024-02-23
Feat: Add reference to lineage node

vchan:


sqlglot/lineage.py
```diff
@@ -204,13 +207,15 @@ def to_node(
             source = scope.sources.get(table)
 
             if isinstance(source, Scope):
+                reference = dict(scope.references).get(table)
```
tobymao, line 210: i assume references is a list of tuples because there could be duplicates, this would essentially clobber the duplicates, but it's just for metadata i guess, maybe just do a for loop here instead of creating the dict?

reference if next(... , None)

sqlglot/lineage.py
```diff
@@ -18,6 +18,7 @@ class Node:
     expression: exp.Expression
     source: exp.Expression
     downstream: t.List[Node] = field(default_factory=list)
+    source_name: str = ""
```
tobymao, line 21: what's the difference between alias and source_name ?

vchan: `source_name` is intended to be the name of the source (the query passed into the `sources` argument of the lineage function) this node came from. `alias` is intended for the CTE name the node came from.

georgesittas:
This seems like a breaking change btw, should probably mark as such.

sqlglot/lineage.py
vchan: I found selected_sources and it seemed like a better option than references.

tobymao (approved):

sqlglot/lineage.py
```diff
@@ -204,13 +213,15 @@ def to_node(
             source = scope.sources.get(table)
 
             if isinstance(source, Scope):
+                selected_node, selected_source = scope.selected_sources.get(table, (None, None))
```
georgesittas, line 216: Maybe I'm missing something, but I think `selected_source` is unused?
```suggestion
                selected_node, _ = scope.selected_sources.get(table, (None, None))
```

georgesittas (approved):

sqlglot/lineage.py
vchan: Ah, you're right.

PR #3019, opened 2024-02-23
feat: add unnest type annotations

tobymao:


sqlglot/optimizer/annotate_types.py
```diff
@@ -602,7 +607,30 @@ def _annotate_div(self, expression: exp.Div) -> exp.Div:
...
         return expression
+
+    def _annotate_unnest(self, expression: exp.Unnest) -> exp.Unnest:
+        self._annotate_args(expression)
+        child = seq_get(expression.expressions, 0)
+
+        if child:
+            self._set_type(expression, seq_get(child.type.expressions, 0))
+        else:
+            self._set_type(expression, None)
```
georgesittas, line 635: what about
```suggestion
        self._set_type(expression, child and seq_get(child.type.expressions, 0))
```

sqlglot/expressions.py
```diff
@@ -1399,6 +1399,10 @@ class ColumnDef(Expression):
     def constraints(self) -> t.List[ColumnConstraint]:
         return self.args.get("constraints") or []
 
+    @property
+    def kind(self) -> t.Optional[DataType]:
+        return self.args.get("kind")
```
georgesittas, line 1404: Let's clean up some places where we do `expr.args.get("kind")`, I believe there are only three (transforms.py, databricks.py, postgres.py)

georgesittas (approved):

PR #3021, opened 2024-02-23
Fix!(bigquery): preserve quoted table paths

### LINKED ISSUE #3014: Unexpected unnest
**Before you file an issue**

```
>>> import sqlglot
>>> sqlglot.transpile("""WITH Coordinates AS (SELECT [1,2] AS position)
... SELECT results FROM Coordinates, `Coordinates.position` AS results;
```
should have result
```
'WITH Coordinates AS (SELECT ARRAY[1, 2] AS position) SELECT results FROM Coordinates, Coordinates.position AS results'
```
actually I got the unnest

```
WITH Coordinates AS (SELECT ARRAY[1, 2] AS position) SELECT results FROM Coordinates, UNNEST(Coordinates.position) AS _t(results)
```
**Official Documentation**

georgesittas:
Fixes #3014

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -780,12 +780,20 @@ class BigQuery(Dialect):
             "within",
         }
 
-        def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
+        def table_parts(self, expression: exp.Table) -> str:
+            # Depending on the context, `x.y` may not resolve to the same data source as `x`.`y`, so
+            # we need to make sure the quotes are preserved in the first case.
+            #
+            # For example, if there is a CTE whose name clashes with a schema name, then the former
+            # will return a table in that schema, whereas the latter will return the CTE's y column:
+            #
+            # - WITH x AS (SELECT [1, 2] AS y) SELECT * FROM x, `x.y`   -> cross join
+            # - WITH x AS (SELECT [1, 2] AS y) SELECT * FROM x, `x`.`y` -> implicit unnest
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -778,6 +780,11 @@ class Generator(generator.Generator):
             "within",
         }
 
+        def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
+            if expression.meta.get("quoted_table"):
```
tobymao, line 784: can we do this one without mutation? since it's so common

expression = exp.Table(this=exp.Identifier(this=p.naem for ...))

georgesittas: What's the advantage of constructing a new `Table` vs mutating here? Remember that we'll also have to carry over the other args in the table.

georgesittas: Whoops, good catch yeah - I intended for this to be an identifier.

> what do you mean you have to carry over args?

I mean that you can't just do `expression = exp.Table(this=...)` - what if `expression` had more [args](https://github.com/tobymao/sqlglot/blob/17e34e79d22e3c8211f1bf42047d4ed3557628b6/sqlglot/expressions.py#L2620)? With this overwrite here you'll drop them.

georgesittas: https://github.com/tobymao/sqlglot/pull/3021/commits/a9ac1c21e690ed6d3c3c0087542bf769677f0306

tobymao (approved):

PR #3025, opened 2024-02-26
Feat(tsql): add support for OPTION clause, select only

nadav-botanica:
This is a new implementation of  the OPTION clause parsing, using query modifiers.
Note that in order to use query modifiers, a lot of the tsql specific logic had to be hoisted up to the generic parser.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -442,6 +442,11 @@ class TSQL(Dialect):
     class Parser(parser.Parser):
         SET_REQUIRES_ASSIGNMENT_DELIMITER = False
 
+        QUERY_MODIFIER_PARSERS = {
+            **parser.Parser.QUERY_MODIFIER_PARSERS,
+            TokenType.OPTION: lambda self: ("option", self._parse_option()),
+        }
+
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "CHARINDEX": lambda args: exp.StrPosition(
@@ -495,103 +500,6 @@ class TSQL(Dialect):
         STATEMENT_PARSERS = {
             **parser.Parser.STATEMENT_PARSERS,
             TokenType.END: lambda self: self._parse_command(),
-        }
-
-        OPTION_PREFIX_TO_PARSE_FUNCTION = {
-            "LABEL": lambda self: self._generic_parse_option(
-                "LABEL",
-                [
-                    lambda: self._match(TokenType.EQ)
-                    and self._return_option_with_value("LABEL", self._parse_string())
-                ],
-            ),
-            "HASH": lambda self: self._generic_parse_option("HASH", ["GROUP", "UNION", "JOIN"]),
-            "ORDER": lambda self: self._generic_parse_option("ORDER", ["GROUP"]),
-            "CONCAT": lambda self: self._generic_parse_option("CONCAT", ["UNION"]),
-            "MERGE": lambda self: self._generic_parse_option("MERGE", ["UNION", "JOIN"]),
-            "LOOP": lambda self: self._generic_parse_option("LOOP", ["JOIN"]),
-            "DISABLE_OPTIMIZED_PLAN_FORCING": lambda self: self._generic_parse_option(
-                "DISABLE_OPTIMIZED_PLAN_FORCING", []
-            ),
-            "EXPAND": lambda self: self._generic_parse_option("EXPAND", ["VIEWS"]),
-            "FAST": lambda self: self._generic_parse_option(
-                "FAST", [lambda: self._return_option_with_value("FAST", self._parse_number())]
-            ),
-            "FORCE": lambda self: self._generic_parse_option(
-                "FORCE", ["ORDER", "EXTERNALPUSHDOWN", "SCALEOUTEXECUTION"]
-            ),
-            "DISABLE": lambda self: self._generic_parse_option(
-                "DISABLE", ["EXTERNALPUSHDOWN", "SCALEOUTEXECUTION"]
-            ),
-            "IGNORE_NONCLUSTERED_COLUMNSTORE_INDEX": lambda self: self._generic_parse_option(
-                "IGNORE_NONCLUSTERED_COLUMNSTORE_INDEX", []
-            ),
-            "KEEP": lambda self: self._generic_parse_option("KEEP", ["PLAN"]),
-            "KEEPFIXED": lambda self: self._generic_parse_option("KEEPFIXED", ["PLAN"]),
-            "MAX_GRANT_PERCENT": lambda self: self._generic_parse_option(
-                "MAX_GRANT_PERCENT",
-                [
-                    lambda: self._match(TokenType.EQ)
-                    and self._generic_parse_option(
-                        "MAX_GRANT_PERCENT",
-                        [
-                            lambda: self._return_option_with_value(
-                                "MAX_GRANT_PERCENT", self._parse_number(), requires_equals=True
-                            )
-                        ],
-                    )
-                ],
-            ),
-            "MIN_GRANT_PERCENT": lambda self: self._generic_parse_option(
-                "MIN_GRANT_PERCENT",
-                [
-                    lambda: self._match(TokenType.EQ)
-                    and self._generic_parse_option(
-                        "MIN_GRANT_PERCENT",
-                        [
-                            lambda: self._return_option_with_value(
-                                "MIN_GRANT_PERCENT", self._parse_number(), requires_equals=True
-                            )
-                        ],
-                    )
-                ],
-            ),
-            "MAXDOP": lambda self: self._generic_parse_option(
-                "MAXDOP", [lambda: self._return_option_with_value("MAXDOP", self._parse_number())]
-            ),
-            "MAXRECURSION": lambda self: self._generic_parse_option(
-                "MAXRECURSION",
-                [lambda: self._return_option_with_value("MAXRECURSION", self._parse_number())],
-            ),
-            "NO_PERFORMANCE_SPOOL": lambda self: self._generic_parse_option(
-                "NO_PERFORMANCE_SPOOL", []
-            ),
-            "OPTIMIZE": lambda self: self._generic_parse_option(
-                "OPTIMIZE",
-                [
-                    # There is still no support for the following syntax:
-                    # OPTIMIZE FOR ( @variable_name { UNKNOWN | = <literal_constant> } [ , ...n ] )
-                    lambda: self._match_text_seq("FOR")
-                    and self._generic_parse_option("OPTIMIZE FOR", ["UNKNOWN"])
-                ],
-            ),
-            "PARAMETERIZATION": lambda self: self._generic_parse_option(
-                "PARAMETERIZATION", ["SIMPLE", "FORCED"]
-            ),
-            "QUERYTRACEON": lambda self: self._generic_parse_option(
-                "QUERYTRACEON",
-                [lambda: self._return_option_with_value("QUERYTRACEON", self._parse_number())],
-            ),
-            "RECOMPILE": lambda self: self._generic_parse_option("RECOMPILE", []),
-            "ROBUST": lambda self: self._generic_parse_option("ROBUST", ["PLAN"]),
-            "USE": lambda self: self._generic_parse_option(
-                "USE",
-                [
-                    lambda: self._match_text_seq("PLAN")
-                    and self._return_option_with_value("USE PLAN", self._parse_string())
-                ],
-            ),
-            # No support yet for TABLE HINT
         }
 
         LOG_DEFAULTS_TO_LN = True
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -502,6 +508,13 @@ class Parser(parser.Parser):
         STRING_ALIASES = True
         NO_PAREN_IF_COMMANDS = False
 
+        def _parse_string(self) -> t.Optional[exp.Expression]:
+            if self._curr is not None and self._curr.token_type == TokenType.NATIONAL_STRING:
+                result = self.PRIMARY_PARSERS[TokenType.NATIONAL_STRING](self, self._curr)
+                self._match(TokenType.NATIONAL_STRING)
+                return result
+            return super()._parse_string()
+
```
georgesittas, line 517: Why do we need this override?

sqlglot/expressions.py
```diff
@@ -2556,9 +2556,20 @@ def with_(
...
 
 
+# https://learn.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=sql-server-ver16
+# https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver16
+class Option(Expression):
+    arg_types = {"option_type": True, "option_value": False, "requires_equals": False}
+
+
+class Options(Expression):
+    arg_types = {"options": False}
```
georgesittas, line 2570: Let's get rid of this and just store a list of `Option` in an `"options"` query modifier arg instead.

sqlglot/expressions.py
```diff
@@ -2556,9 +2556,20 @@ def with_(
...
     "settings": False,
     "format": False,
+    "option": False,
 }
 
 
+# https://learn.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=sql-server-ver16
+# https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver16
+class Option(Expression):
+    arg_types = {"option_type": True, "option_value": False, "requires_equals": False}
```
georgesittas, line 2566: Maybe rename this to something like `QueryOption`? If possible, let's try not to use a generic name like "option" for something that's only relevant to T-SQL.

sqlglot/generator.py
```diff
@@ -2083,9 +2083,22 @@ def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
             self.sql(expression, "order"),
             *offset_limit_modifiers,
             *self.after_limit_modifiers(expression),
+            *self.sql(expression, "option"),
```
georgesittas, line 2086: So if you changed the `"option"` to `"options"` then you could change this line to:

```python
self.expressions(expression, key="options")
```

sqlglot/expressions.py
georgesittas: Also, the `option_` prefix is redundant - we already know this is under `Option`.

Maybe use something like `this` and `expression` for the first two args? The `requires_equals` arg can be omitted if you make a map with known options that need the `=` which you can leverage at generation time.

sqlglot/generator.py
```diff
@@ -2083,9 +2083,22 @@ def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
...
             sep="",
         )
 
+    def option_sql(self, expression: exp.Option):
+        option_type = self.sql(expression, "option_type")
+        option_value = self.sql(expression, "option_value")
+        if option_value:
+            optional_equal_sign = "= " if expression.args.get("requires_equals") else ""
```
georgesittas, line 2094: So here you'd only add the `=` if the option type exists in the aforementioned set of keywords.

sqlglot/generator.py
```diff
@@ -2083,9 +2083,22 @@ def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
...
+            return f"{option_type} {optional_equal_sign}{option_value}"
+        return option_type
+
+    def options_sql(self, expression: exp.Options):
+        options_sql = csv(*[self.sql(opt) for opt in expression.args.get("options", [])])
+        return f" OPTION({options_sql})"
```
georgesittas, line 2100: This should be irrelevant after refactoring, but when you need a csv in the future you can do:

```python
self.expressions(expression, key="options")
```

sqlglot/parser.py
```diff
@@ -923,6 +923,101 @@ class Parser(metaclass=_Parser):
         TokenType.START_WITH: lambda self: ("connect", self._parse_connect()),
     }
 
+    OPTION_PREFIX_TO_PARSE_FUNCTION = {
```
georgesittas, line 926: Let's not pollute the base parser with this - it's very specific to T-SQL, so imo we should move it there.

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
                 break
         return this
 
+    def _generic_parse_option(
+        self, option_name: str, valid_continuations: t.List[t.Union[str, t.Callable]]
+    ) -> t.Optional[exp.Option]:
+        # Handle the special case of options with no continuations.
+        if not valid_continuations:
+            return self._return_option_with_value(
+                self.expression(exp.Identifier, this=option_name), None
```
georgesittas, line 2586: We shouldn't be constructing `Identifier`s here - in reality these are `Var`s. They represent different things.

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
...
+            elif isinstance(valid_continuation, str):
+                if self._match_text_seq(valid_continuation):
+                    return self._return_option_with_value(
+                        self.expression(exp.Identifier, this=option_name), valid_continuation
+                    )
+            else:
+                # Shouldn't happen.
+                self.raise_error(
+                    f"_generic_parse_option only accepts strings and callables as valid continuations, not '{valid_continuation}'"
+                )
```
georgesittas, line 2603: We don't need this

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
...
+
+        for valid_continuation in valid_continuations:
+            if callable(valid_continuation):
+                res = valid_continuation()
+                if res is not None:
+                    return res
+            elif isinstance(valid_continuation, str):
+                if self._match_text_seq(valid_continuation):
+                    return self._return_option_with_value(
+                        self.expression(exp.Identifier, this=option_name), valid_continuation
```
georgesittas, line 2597: Ditto about `Identifier` vs `Var`

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
...
+        return None
+
+    def _return_option_with_value(
+        self,
+        option_type: t.Optional[exp.Expression],
+        option_value: t.Optional[t.Union[str, exp.Expression]],
+        requires_equals=False,
+    ) -> t.Optional[exp.Option]:
+        if isinstance(option_value, str):
+            option_value = self.expression(exp.Identifier, this=option_value)
```
georgesittas, line 2614: Ditto

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
...
+        )
+
+    def _parse_single_option(self) -> t.Optional[exp.Expression]:
+        cur_prefix = self._curr.text.upper()
+        if cur_prefix in self.OPTION_PREFIX_TO_PARSE_FUNCTION:
+            self._match_text_seq(cur_prefix.upper())
+            return self.OPTION_PREFIX_TO_PARSE_FUNCTION[cur_prefix](self)
+        else:
+            self.raise_error("Invalid option type.")
+            return None
```
georgesittas, line 2629:
```suggestion
        self.raise_error("Invalid option type.")
        return None
```

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
...
+
+    def _parse_option(self):
+        self._match(TokenType.OPTION)
+        self._match_l_paren()
+        options = self._parse_csv(self._parse_single_option)
+        options = [opt for opt in options if opt is not None]
+        self._match_r_paren()
```
georgesittas, line 2636:
```suggestion
        options = self._parse_wrapped_csv(self._parse_single_option)
```

sqlglot/parser.py
```diff
@@ -2482,6 +2577,65 @@ def _parse_query_modifiers(
...
+        return self.expression(exp.Options, options=options)
```
georgesittas, line 2637: This should be irrelevant

sqlglot/dialects/tsql.py
nadav-botanica: For example, look [here](https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver16), at the `USE PLAN <XML>` syntax - it accepts regular strings as well as national strings - It might be correct in general to parse national strings as strings - WDYT?

sqlglot/expressions.py
nadav-botanica: Will do

sqlglot/parser.py
nadav-botanica: Done

sqlglot/dialects/tsql.py
georgesittas: National strings should be parsed as [`National`](https://github.com/tobymao/sqlglot/blob/1612e622bd3514d9ca366837f47452969e5267d8/sqlglot/expressions.py#L1891) - they're not necessarily equivalent to strings, right?

sqlglot/expressions.py
nadav-botanica: Done

sqlglot/generator.py
nadav-botanica: Done

sqlglot/parser.py
nadav-botanica: Done

sqlglot/parser.py
nadav-botanica: Why? What if a programmer makes a mistake and put something that isn't a string or a callable?

sqlglot/generator.py
nadav-botanica: Sure - thanks!

sqlglot/parser.py
georgesittas: It should be caught by mypy and most likely by the tests as well.

We're usually not writing defensive code like this one, because it's too noisy. Maybe only for very commonly reached / important code paths that are user-facing.

sqlglot/parser.py
georgesittas: `Var` represents "keywords", whereas `Identifier` represents names of data objects such as columns, tables, etc. This distinction is important because for example identifiers are transformed by the optimizer whereas vars are not.

nadav-botanica: OK - so we might want to construct a `var` in case of keywords, and a Literal otherwise - WDYT?

sqlglot/generator.py
nadav-botanica: Is there any way to wrap the csv result in parentheses when generating the sql, if using expressions?

sqlglot/parser.py
georgesittas: Var sounds good for keywords, yes. What test case would correspond to a `Literal`?

nadav-botanica: For example, `MAXDOP 15`

sqlglot/generator.py
georgesittas: You can either use `self.wrap(self.expressions(...))` or manually wrap it using an f-string, like `f"({...})"`. The former may be better because it formats the wrapped csv when `pretty=True`. You can play with it a bit and see if it works as expected, otherwise use the f-string wrapping as a fallback?

nadav-botanica: This would mean I can't use the `prefix` option for the `OPTION` - I think I would need to resort to using `wrap` + f string, and checking whether the `OPTION` actually exists - therefore I think something like this would work:
```
        option_sql = ""
        if expression.args.get("options", []):
            option_sql = f" OPTION{self.wrap(self.expressions(expression, key='options'))}"
```

WDYT?

sqlglot/generator.py
nadav-botanica: Done

sqlglot/parser.py
nadav-botanica: Done

sqlglot/generator.py
georgesittas: Makes sense - this probably needs to be done before the `return` statement is reached. Here's how I'd write this:

```python
if options:
    options = f" OPTION{self.wrap(options)}"
```

sqlglot/dialects/tsql.py
nadav-botanica: It appears that they are treated the same way (at least in the case of OPTION parsing)  - we can drop this change for the moment, and not support labels/XMLs that are passed to the OPTION clause.
WDYT?

nadav-botanica: We should support, ideally, both of these syntaxes:
`USE PLAN N'<xml_plan>'`
`USE PLAN '<xml_plan>'`
We can check for either of them when parsing the content of the PLAN, but it appears not very generic, instead of just accepting a string IMHO..

georgesittas:
> @georgesittas - It appear that I don't have permission to re-run the tests (and they didn't re-run automatically after commit ammending) - can you please re-run them for me?

ACK.

sqlglot/dialects/tsql.py
```diff
@@ -502,6 +508,110 @@ class Parser(parser.Parser):
...
         NO_PAREN_IF_COMMANDS = False
 
+        OPTION_PREFIX_TO_PARSE_FUNCTION = {
+            "LABEL": lambda self: self._generic_parse_option(
+                "LABEL",
+                [
+                    lambda: self._match(TokenType.EQ)
+                    and self._return_option_with_value("LABEL", self._parse_primary())
+                ],
+            ),
```
georgesittas, line 518: Can we refactor this map definition to simplify it a bit? One idea is to fetch (e.g., `LABEL`) the key using `self._prev.text.upper()` instead of repeating it in the first arg. I implemented something similar in another [PR](https://github.com/tobymao/sqlglot/pull/3027/files#diff-14ef9d8b99f48dcf8f6a4e3bbab7bf55a9d85d2b83d684c3c822f0f94fcd1d07R493).

sqlglot/dialects/tsql.py
```diff
@@ -502,6 +508,110 @@ class Parser(parser.Parser):
...
+            # No support yet for TABLE HINT
+        }
+
+        def _parse_string(self) -> t.Optional[exp.Expression]:
+            if self._curr is not None and self._curr.token_type == TokenType.NATIONAL_STRING:
+                result = self.PRIMARY_PARSERS[TokenType.NATIONAL_STRING](self, self._curr)
+                self._match(TokenType.NATIONAL_STRING)
+                return result
+            return super()._parse_string()
```
georgesittas, line 613: Probably missed this one? If we keep the `_parse_primary()` in `USE` it should not be needed anymore.

sqlglot/dialects/tsql.py
```diff
@@ -699,6 +809,8 @@ class Generator(generator.Generator):
         SUPPORTS_SELECT_INTO = True
         JSON_PATH_BRACKETED_KEY_SUPPORTED = False
 
+        OPTIONS_THAT_REQUIRE_EQUAL = ["MAX_GRANT_PERCENT", "MIN_GRANT_PERCENT", "LABEL"]
```
georgesittas, line 812:
```suggestion
        OPTIONS_THAT_REQUIRE_EQUAL = ("MAX_GRANT_PERCENT", "MIN_GRANT_PERCENT", "LABEL")
```

sqlglot/generator.py
```diff
@@ -321,6 +321,9 @@ class Generator(metaclass=_Generator):
     # Whether any(f(x) for x in array) can be implemented by this dialect
     CAN_IMPLEMENT_ARRAY_ANY = False
 
+    # Look at tsql.py for concrete examples of options that require adding a '=' when displaying them
+    OPTIONS_THAT_REQUIRE_EQUAL: t.List[str] = []
```
georgesittas, line 325: You can omit this if you define it as a file-level constant in tsql.py

sqlglot/parser.py
```diff
@@ -2482,6 +2485,51 @@ def _parse_query_modifiers(
                 break
         return this
 
+    def _generic_parse_option(
```
georgesittas, line 2488: Can we also move these option-specific parser methods to tsql.py?

georgesittas (commented):
A few more comments - appreciate the quick iterations on this @nadav-botanica!

sqlglot/parser.py
```diff
@@ -923,6 +923,9 @@ class Parser(metaclass=_Parser):
         TokenType.START_WITH: lambda self: ("connect", self._parse_connect()),
     }
 
+    # For concrete implementation, look at tsql
+    OPTION_PREFIX_TO_PARSE_FUNCTION: t.Dict[str, t.Callable] = {}
```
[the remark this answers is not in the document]
nadav-botanica: Hey @georgesittas, I think there's something we need to decide here in regards to whether OPTION is a generic SQL feature or not. In practice, I've only seen it in TSQL (Oracle has a similar feature, with specially-crafted comments that define similar behaviour). As a TSQL-only feature, I wanted to implement this in the TSQL parser and moved it up to the generic one as requested.
Can you please share your thoughts on whether this feature belongs in the TSQL or generic parser?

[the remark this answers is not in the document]
georgesittas: Even though we occasionally implement new features in the base classes to make them more "generic", imo supporting `OPTION` in particular only for T-SQL seemed more reasonable because the options are T-SQL-specific. @tobymao may have a different opinion here, so I'll defer to him.

My intention from the start was to try and keep this in T-SQL, apologies if there was a misunderstanding w.r.t.

> I wanted to implement this in the TSQL parser and moved it up to the generic one as requested.

Let's connect in [Slack](https://tobikodata.com/slack) to iterate / chat more easily?

tests/dialects/test_tsql.py
```diff
@@ -1,11 +1,101 @@
...
+            # Missing commas
+            "SELECT * FROM Table1 OPTION(HASH GROUP HASH GROUP)",
+        ]
+        for query in raising_queries:
+            with self.assertRaises(ParseError, msg=f"When running '{query}'"):
+                self.parse_one(query)
+
+        logger = logging.getLogger("sqlglot")
+        with self.assertLogs(logger) as cm:
+            parse_one("ALTER TABLE table2 OPTION(HASH GROUP)")
```
georgesittas, line 94:
```suggestion
            parse_one("ALTER TABLE table2 OPTION(HASH GROUP)", read="tsql")
```

sqlglot/generator.py
```diff
@@ -2083,9 +2087,13 @@ def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
             self.sql(expression, "order"),
             *offset_limit_modifiers,
             *self.after_limit_modifiers(expression),
+            options,
             sep="",
         )
 
+    def queryoption_sql(self, expression: exp.QueryOption):
```
georgesittas, line 2094:
```suggestion
    def queryoption_sql(self, expression: exp.QueryOption) -> str:
```

sqlglot/dialects/tsql.py
```diff
@@ -779,6 +894,14 @@ class Generator(generator.Generator):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        def queryoption_sql(self, expression: exp.QueryOption):
```
georgesittas, line 897:
```suggestion
        def queryoption_sql(self, expression: exp.QueryOption) -> str:
```

tests/dialects/test_tsql.py
```diff
@@ -1,11 +1,101 @@
...
+            # Must be followed by 'PLAN"
+            "SELECT * FROM Table1 OPTION(KEEPFIXED)",
+            # Should expect a numeric literal
+            "SELECT * FROM Table1 OPTION(FAST 'abcd')",
+            # Missing commas
+            "SELECT * FROM Table1 OPTION(HASH GROUP HASH GROUP)",
+        ]
+        for query in raising_queries:
+            with self.assertRaises(ParseError, msg=f"When running '{query}'"):
+                self.parse_one(query, read="tsql")
```
georgesittas, line 90: Whoops, didn't notice this was `self.parse_one` so I went ahead and used the dialect out of habbit haha
```suggestion
                self.parse_one(query)
```

tests/dialects/test_tsql.py
```diff
@@ -1,11 +1,101 @@
...
+            # Missing commas
+            "SELECT * FROM Table1 OPTION(HASH GROUP HASH GROUP)",
+        ]
+        for query in raising_queries:
+            with self.assertRaises(ParseError, msg=f"When running '{query}'"):
+                self.parse_one(query)
+
+        logger = logging.getLogger("sqlglot")
+        with self.assertLogs(logger) as cm:
+            parse_one("ALTER TABLE table2 OPTION(HASH GROUP)", read="tsql")
```
georgesittas, line 94:
```suggestion
            self.parse_one("ALTER TABLE table2 OPTION(HASH GROUP)")
```

Issue #2850, opened 2024-01-18
Support `TRUNCATE TABLE` statments


judahrand:
**Is your feature request related to a problem? Please describe.**
Currently, `TRUNCATE TABLE table` is parsed like this:

```python
>>> sqlglot.parse_one('TRUNCATE TABLE "table"')
(COMMAND this: TRUNCATE, expression: 
  (LITERAL this: TABLE "table", is_string: True))
```

This doesn't seem correct. 

**Describe the solution you'd like**

I'd expect something more like:

```python
>>> sqlglot.parse_one('TRUNCATE TABLE "table"')
(TRUNCATE this: 
  (IDENTIFIER this: table, quoted: True) , kind: TABLE)
```
**Describe alternatives you've considered**

**Additional context**

tobymao:
not planned, feel free to contribute a PR if you need this

tobymao:
we like to keep the issues list clean and as an indication of work that is actively being worked on and will be completed soon. you can filter on the issues list of the 'not planned' tag.

i don't have an objection to the suggested functionality, but it's not high priority.

benknoble:
This used to be supported, if I'm reading the history right. In https://github.com/tobymao/sqlglot/commit/48b8b186aaf9f5380c5b72012c39e6a7370222e2 from https://github.com/tobymao/sqlglot/issues/63, you added a TRUNCATE token. But later in https://github.com/tobymao/sqlglot/commit/1d7c5e74b2cfe48b0f8c9152905714bf090e7c1a you made TRUNCATE a command.

TRUNCATE is part of the SQL standard (introduced in 2008) according to https://en.wikipedia.org/wiki/Truncate_(SQL).

Can we please bring back correct parsing of the truncate command?

georgesittas:
The result was the same for the commit you're referencing:

```python
➜  sqlglot git:(48b8b186) ✗ python
Python 3.10.13 (main, Nov  1 2023, 16:34:31) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlglot
>>> sqlglot.parse_one('TRUNCATE TABLE "table"')
(COMMAND this: TRUNCATE, expression:
  (LITERAL this:  TABLE "table", is_string: True))
>>>
```

benknoble:
[...]

PR #3026, opened 2024-02-26
Feat: Support for TRUNCATE TABLE/DATABASE DDL

VaggelisD:
This PR aims to solve [issue 2885](https://github.com/tobymao/sqlglot/issues/2850) by supporting `TRUNCATE` natively in the AST for the following dialects:

BigQuery, Databricks, DuckDB, Clickhouse, Hive, MySQL, Postgres, Spark, Snowflake, Redshift, T-SQL, Presto.

The common / "superset" representation among all previous dialects is the following:

`TRUNCATE [TABLE | DATABASE] [IF EXISTS] [[ONLY] table_name [*], ]* [-- PROPERTIES -- ]`

Where `[-- PROPERTIES -- ]` can be:

- Postgres: `[ RESTART IDENTITY | CONTINUE IDENTITY ] [ CASCADE | RESTRICT ]`
- Databricks / Hive / Spark (similar spec): `[PARTITION partition_spec]`
- T-SQL: `[ WITH ( PARTITIONS ( { <partition_number_expression> | <range> }  [ , ...n ] ) ) ]`
- Clickhouse: `[ON CLUSTER cluster]`

Design Notes:
- Distinguish between the function `TRUNCATE(a, b)` and `TRUNCATE` DDL by returning an anonymous function if `LPAREN` is the 2nd token in the stream
- Add "ONLY" table prefix keyword in `exp.Table` which is used in [Postgres's table inheritance](https://www.postgresql.org/docs/current/tutorial-inheritance.html)
- Add `exp.PartitionRange` which is used for T-SQL's partition clause. The Databricks / Spark / Hive partition clause is already implemented in `exp.Partition`
- SQLite does not implement `TRUNCATE` DDL, it's implicitly supported as a `DELETE FROM table` DML. 

TRUNCATE DDL Docs
-------------------------

[Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-truncate-table.html), [Hive](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-TruncateTable), [Spark](https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-ddl-truncate-table.html), [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/truncate-table-transact-sql?view=sql-server-ver16), [Clickhouse](https://clickhouse.com/docs/en/sql-reference/statements/truncate), [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/truncate-table), [DuckDB](https://duckdb.org/docs/sql/statements/delete.html), [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#truncate_table_statement), [Presto](https://prestodb.io/docs/current/sql/truncate.html), [MySQL](https://dev.mysql.com/doc/refman/8.0/en/truncate-table.html), [Postgres](https://www.postgresql.org/docs/current/sql-truncate.html), [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_TRUNCATE.html), [SQLite](https://www.techonthenet.com/sqlite/truncate.php#:~:text=SQLite%20does%20not%20have%20an,TRUNCATE%20optimizer%20handles%20the%20rest)

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -138,6 +138,7 @@ class Redshift(Postgres):
             "TOP": TokenType.TOP,
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
+            "ONLY": TokenType.ONLY,
         }
         KEYWORDS.pop("VALUES")
 
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -138,7 +138,6 @@ class Redshift(Postgres):
             "TOP": TokenType.TOP,
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
-            "ONLY": TokenType.ONLY,
         }
         KEYWORDS.pop("VALUES")
 
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -682,9 +682,7 @@ class TSQL(Dialect):
             return self.expression(exp.UniqueColumnConstraint, this=this)
 
         def _parse_partition(self) -> t.Optional[exp.Partition]:
-            if not self._match_pair(TokenType.WITH, TokenType.L_PAREN) or not self._match_text_seq(
-                "PARTITIONS"
-            ):
+            if not self._match_text_seq("WITH", "(", "PARTITIONS"):
                 return None
 
             def parse_range():
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -682,10 +682,9 @@ class TSQL(Dialect):
             return self.expression(exp.UniqueColumnConstraint, this=this)
 
         def _parse_partition(self) -> t.Optional[exp.Partition]:
-            if not self._match_pair(TokenType.WITH, TokenType.L_PAREN):
-                return None
-
-            if not self._match_text_seq("PARTITIONS"):
+            if not self._match_pair(TokenType.WITH, TokenType.L_PAREN) or not self._match_text_seq(
+                "PARTITIONS"
+            ):
                 return None
 
             def parse_range():
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -681,6 +681,29 @@ class TSQL(Dialect):
 
             return self.expression(exp.UniqueColumnConstraint, this=this)
 
+        def _parse_partition(self) -> t.Optional[exp.Partition]:
+            if not self._match_pair(TokenType.WITH, TokenType.L_PAREN):
+                return None
+
+            if not self._match_text_seq("PARTITIONS"):
+                return None
+
+            def parse_range():
+                low = self._parse_bitwise()
+                high = self._parse_bitwise() if self._match_text_seq("TO") else None
+
+                return (
+                    self.expression(exp.PartitionRange, this=low, expression=high) if high else low
+                )
+
+            partition = self.expression(
+                exp.Partition, expressions=self._parse_wrapped_csv(parse_range)
+            )
+
+            self._match_r_paren()
+
+            return partition
+
     class Generator(generator.Generator):
         LIMIT_IS_TOP = True
         QUERY_HINTS = False
@@ -990,3 +1013,6 @@ class TSQL(Dialect):
                 this_sql = self.sql(this)
             expression_sql = self.sql(expression, "expression")
             return self.func(name, this_sql, expression_sql if expression_sql else None)
+
+        def partition_sql(self, expression: exp.Partition) -> str:
+            return f"WITH (PARTITIONS({self.expressions(expression, flat=True)}))"
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1224,6 +1224,19 @@ class Create(DDL):
         return kind and kind.upper()
 
 
+class Truncate(Expression):
+    arg_types = {
+        "expressions": True,
+        "is_database": False,
+        "exists": False,
+        "only": False,
+        "cluster": False,
+        "identity": False,
+        "option": False,
+        "partition": False,
+    }
+
+
 # https://docs.snowflake.com/en/sql-reference/sql/create-clone
 # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_clone_statement
 # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_copy
@@ -1906,6 +1919,10 @@ class LoadData(Expression):
 
 class Partition(Expression):
     arg_types = {"expressions": True}
+
+
+class PartitionRange(Expression):
+    arg_types = {"this": True, "expression": False}
 
 
 class Fetch(Expression):
@@ -2633,6 +2650,7 @@ class Table(Expression):
         "pattern": False,
         "ordinality": False,
         "when": False,
+        "only": False,
     }
 
     @property
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1224,7 +1224,7 @@ class Create(DDL):
         return kind and kind.upper()
 
 
-class Truncate(Expression):
+class TruncateTable(Expression):
     arg_types = {
         "expressions": True,
         "is_database": False,
@@ -1922,7 +1922,7 @@ class Partition(Expression):
 
 
 class PartitionRange(Expression):
-    arg_types = {"this": True, "expression": False}
+    arg_types = {"this": True, "expression": True}
 
 
 class Fetch(Expression):
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1546,6 +1546,7 @@ class Generator(metaclass=_Generator):
 
     def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
         table = self.table_parts(expression)
+        only = "ONLY " if expression.args.get("only") else ""
         version = self.sql(expression, "version")
         version = f" {version}" if version else ""
         alias = self.sql(expression, "alias")
@@ -1572,7 +1573,7 @@ class Generator(metaclass=_Generator):
         if when:
             table = f"{table} {when}"
 
-        return f"{table}{version}{file_format}{alias}{hints}{pivots}{joins}{laterals}{ordinality}"
+        return f"{only}{table}{version}{file_format}{alias}{hints}{pivots}{joins}{laterals}{ordinality}"
 
     def tablesample_sql(
         self,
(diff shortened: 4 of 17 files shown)


sqlglot/parser.py
```diff
@@ -5911,3 +5919,50 @@ def _replace_lambda(
...
+
+        # Clickhouse supports TRUNCATE DATABASE as well
+        is_database = self._match(TokenType.DATABASE)
+
+        self._match(TokenType.TABLE)
+
+        exists = self._parse_exists(not_=False)
+
+        def _parse_truncate_tables() -> t.Optional[exp.Expression]:
+            this = self._parse_table(schema=True, is_db_reference=is_database)
```
georgesittas, line 5936: Curious about the use of `schema=True` here - we shouldn't be expecting a schema, right?
```suggestion
            this = self._parse_table(is_db_reference=is_database)
```

sqlglot/parser.py
```diff
@@ -5911,3 +5919,50 @@ def _replace_lambda(
...
+
+            # Postgres has wildcard (table) suffix operator
```
georgesittas, line 5938:
```suggestion
            # Postgres supports a wildcard (table) suffix operator, which is a no-op in this context
```

sqlglot/parser.py
```diff
@@ -5911,3 +5919,50 @@ def _replace_lambda(
...
+            self._match_text_seq("*")
+
+            return this
+
+        expressions = self._parse_csv(_parse_truncate_tables)
+
+        cluster = self._parse_id_var() if self._match_text_seq("ON", "CLUSTER") else None
```
georgesittas, line 5945: We already have some logic in clickhouse to parse the `ON CLUSTER` modifier, see https://github.com/tobymao/sqlglot/blob/1612e622bd3514d9ca366837f47452969e5267d8/sqlglot/dialects/clickhouse.py#L501

Can we somehow reuse that instead of reimplementing it here? Same goes for the generation logic.

sqlglot/parser.py
```diff
@@ -5911,3 +5919,50 @@ def _replace_lambda(
...
+            identity = "CONTINUE"
+        else:
+            identity = None
+
+        if self._match_text_seq("CASCADE") or self._match_text_seq("RESTRICT"):
+            option = self._prev.text
+        else:
+            option = None
+
+        return self.expression(
```
georgesittas, line 5959: Is it guaranteed that this method parses the complete / all specs? If not, we should ensure that if there's a `self._curr` here we'll use the `Command` fallback parser.

tests/dialects/test_databricks.py
```diff
@@ -38,6 +38,8 @@ def test_databricks(self):
             "CREATE FUNCTION add_one(x INT) RETURNS INT LANGUAGE PYTHON AS $FOO$def add_one(x):\n  return x+1$FOO$"
         )
 
+        self.validate_identity("TRUNCATE TABLE t1 PARTITION(age = 10, name = 'test', address)")
```
georgesittas, line 41: Let's also add a test with the `LIKE` operator shown in Databricks' docs

tests/dialects/test_postgres.py
```diff
@@ -321,6 +321,18 @@ def test_postgres(self):
...
+        self.validate_identity("TRUNCATE TABLE t1 RESTART IDENTITY")
+        self.validate_identity("TRUNCATE TABLE t1 CASCADE")
+        self.validate_identity("TRUNCATE TABLE t1 RESTRICT")
+        self.validate_identity("TRUNCATE TABLE t1 CONTINUE IDENTITY CASCADE")
+        self.validate_identity("TRUNCATE TABLE t1 RESTART IDENTITY RESTRICT")
+
+        self.validate_all(
+            "TRUNCATE TABLE ONLY t1, t2*, ONLY t3, t4, t5*",
+            write={"postgres": "TRUNCATE TABLE ONLY t1, t2, ONLY t3, t4, t5"},
+        )
```
georgesittas, line 334: Nice. Can you add a test for `SELECT * FROM ONLY t`?

```suggestion
        self.validate_identity(
            "TRUNCATE TABLE ONLY t1, t2*, ONLY t3, t4, t5*",
            "TRUNCATE TABLE ONLY t1, t2, ONLY t3, t4, t5",
        )
```

sqlglot/tokens.py
```diff
@@ -754,6 +755,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "TEMPORARY": TokenType.TEMPORARY,
         "THEN": TokenType.THEN,
         "TRUE": TokenType.TRUE,
+        "TRUNCATE": TokenType.TRUNCATE,
```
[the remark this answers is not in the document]
georgesittas: When you fix this and the similar issue related to `ONLY`, make sure to add tests in `identity.sql` to check that we can indeed round-trip both `SELECT truncate` and `SELECT only`.

sqlglot/parser.py
```diff
@@ -2912,6 +2914,9 @@ def _parse_table(
 
         bracket = parse_bracket and self._parse_bracket(None)
         bracket = self.expression(exp.Table, this=bracket) if bracket else None
+
+        only = self._match_text_seq("ONLY")
```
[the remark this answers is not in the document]
VaggelisD: Good catch; The token workaround seems to have restored the CTE regression. Just to clarify, you found out that `ONLY` is reserved in both DuckDB and Redshift, albeit for different use case? Meaning, these dialects do not have table inheritance such as Postgres, but adding it as a token for them too will block such queries, although not in a very user-friendly way:

```
>>> parse_one("select c from only", dialect="duckdb")
...
sqlglot.errors.ParseError: Expected table name but got None. Line 1, Col: 18.
  select c from only
```

sqlglot/parser.py
georgesittas: Ah, we _do_ need `schema=True` here otherwise we'll try to parse a table alias etc. Your implementation is correct, apologies for the misleading comment.

sqlglot/tokens.py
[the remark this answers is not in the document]
VaggelisD: Done

sqlglot/parser.py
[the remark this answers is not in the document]
VaggelisD: Done - Added `ONLY` as a reserved keyword in Postgres, DuckDB, Redshift

sqlglot/dialects/tsql.py
```diff
@@ -681,6 +681,29 @@ def _parse_unique(self) -> exp.UniqueColumnConstraint:
 
             return self.expression(exp.UniqueColumnConstraint, this=this)
 
+        def _parse_partition(self) -> t.Optional[exp.Partition]:
+            if not self._match_pair(TokenType.WITH, TokenType.L_PAREN):
+                return None
+
+            if not self._match_text_seq("PARTITIONS"):
+                return None
```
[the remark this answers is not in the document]
VaggelisD: Done

sqlglot/expressions.py
```diff
@@ -1908,6 +1921,10 @@ class Partition(Expression):
     arg_types = {"expressions": True}
 
 
+class PartitionRange(Expression):
+    arg_types = {"this": True, "expression": False}
```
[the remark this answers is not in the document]
VaggelisD: Done, good catch, `PartitionRange` was supposed to parse an optional range e.g. `low ['TO' high]` but refactored it later to parse bitwise directly if `TO high` was missing

sqlglot/generator.py
```diff
@@ -3472,3 +3473,30 @@ def struct_sql(self, expression: exp.Struct) -> str:
...
+    def partitionrange_sql(self, expression: exp.PartitionRange) -> str:
+        low = self.sql(expression, "this")
+        high = self.sql(expression, "expression")
+
+        return f"{low} TO {high}"
+
+    def truncate_sql(self, expression: exp.Truncate) -> str:
+        target = "DATABASE" if expression.args.get("is_database") else "TABLE"
+        tables = self.expressions(expression)
+        tables = f" {tables}"
```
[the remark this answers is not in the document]
VaggelisD: Done

sqlglot/parser.py
```diff
@@ -5911,3 +5919,50 @@ def _replace_lambda(
                     else:
                         column.replace(dot_or_id)
         return node
+
+    def _parse_truncate(self) -> t.Optional[exp.Truncate] | exp.Expression:
+        # Not to be confused with TRUNCATE(number, decimals) function call
+        if self._match(TokenType.L_PAREN):
+            return self._parse_function_call(anonymous=True)
+
```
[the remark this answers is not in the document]
VaggelisD: Done, thanks for the heads up

sqlglot/parser.py
VaggelisD: Done

sqlglot/parser.py
VaggelisD: Not all dialects were covered in this PR so added a `exp.Command` fallback case

tests/dialects/test_databricks.py
VaggelisD: Done

tests/dialects/test_postgres.py
VaggelisD: Done

sqlglot/dialects/redshift.py
```diff
@@ -138,6 +138,7 @@ class Tokenizer(Postgres.Tokenizer):
             "TOP": TokenType.TOP,
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
+            "ONLY": TokenType.ONLY,
```
georgesittas, line 141: We can get rid of this, because it's already inherited from Postgres
```suggestion
```

sqlglot/parser.py
```diff
@@ -5911,3 +5923,54 @@ def _replace_lambda(
...
+        is_database = self._match(TokenType.DATABASE)
+
+        self._match(TokenType.TABLE)
+
+        exists = self._parse_exists(not_=False)
+
+        def _parse_tables() -> t.Optional[exp.Expression]:
+            return self._parse_table(schema=True, is_db_reference=is_database)
+
+        expressions = self._parse_csv(_parse_tables)
```
georgesittas, line 5945:
```suggestion
        expressions = self._parse_csv(
            lambda: self._parse_table(schema=True, is_db_reference=is_database)
        )
```

sqlglot/parser.py
```diff
@@ -5911,3 +5923,54 @@ def _replace_lambda(
...
+
+        # Clickhouse supports TRUNCATE DATABASE as well
+        is_database = self._match(TokenType.DATABASE)
+
+        self._match(TokenType.TABLE)
+
+        exists = self._parse_exists(not_=False)
+
+        def _parse_tables() -> t.Optional[exp.Expression]:
+            return self._parse_table(schema=True, is_db_reference=is_database)
```
georgesittas, line 5943:
```suggestion
```

sqlglot/parser.py
VaggelisD: Done

sqlglot/parser.py
VaggelisD: Done

sqlglot/dialects/redshift.py
VaggelisD: True! Done

sqlglot/dialects/tsql.py
```diff
@@ -681,6 +681,28 @@ def _parse_unique(self) -> exp.UniqueColumnConstraint:
 
             return self.expression(exp.UniqueColumnConstraint, this=this)
 
+        def _parse_partition(self) -> t.Optional[exp.Partition]:
+            if not self._match_pair(TokenType.WITH, TokenType.L_PAREN) or not self._match_text_seq(
+                "PARTITIONS"
+            ):
```
georgesittas, line 687: Even though token matching should be faster, we should probably use a single `_match_text_seq` call here to avoid consuming `WITH (` in case we don't find a subsequent `PARTITIONS` token.
```suggestion
            if not self._match_text_seq("WITH", "(", "PARTITIONS"):
```

georgesittas (approved):
Awesome 🎉

PR #3027, opened 2024-02-26
Feat(mysql): add LOCK property, allow properties after ALTER TABLE

### LINKED ISSUE #3020: Online DDL in Mysql
Could support for Mysql's online DDL please be added to Sqlglot? That is, DDL operations should support `ALGORITHM` and `LOCK` clauses.

See, e.g. https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl.html

Here's an example of what I tried to do, the last step is what I'd like to see improved.

```
>>> import sys; sys.version_info
sys.version_info(major=3, minor=11, micro=7, releaselevel='final', serial=0)

>>> import sqlglot; sqlglot._version.version_tuple
(21, 1, 2)

>>> sqlglot.parse_one("ALTER TABLE t1 ADD COLUMN x INT;", dialect="mysql")
AlterTable(
  this=Table(
    this=Identifier(this=t1, quoted=False)),
  actions=[
    ColumnDef(
      kind=DataType(this=Type.INT, nested=False))])

>>> sqlglot.parse_one("ALTER TABLE t1 ADD COLUMN x INT, ALGORITHM=INPLACE, LOCK=EXCLUSIVE;", dialect="mysql")
'ALTER TABLE t1 ADD COLUMN x INT, ALGORITHM=INPLACE, LOCK=EXCLUSIVE' contains unsupported syntax. Falling back to parsing as a 'Command'.
Command(this=ALTER, expression=TABLE t1 ADD COLUMN x INT, ALGORITHM=INPLACE, LOCK=EXCLUSIVE)
```

I'll take a look at the code myself but I suspect it will take me some time to understand how Sqlglot's modules are interrelated and the authors can probably do it faster/better.

Thanks!

georgesittas:
Fixes #3020

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -484,12 +484,10 @@ class MySQL(Dialect):
             self._match(TokenType.EQ)
             return exp.Property(this=exp.var(name), value=self._parse_primary_or_var())
 
-        def _parse_table_option_eq(
-            self, option_parts: t.Optional[t.List[str]] = None
-        ) -> exp.Property:
+        def _parse_table_option_eq(self, *option_parts: str) -> exp.Property:
             name = self._prev.text.upper()
-            if option_parts and self._match_text_seq(option_parts):
-                name = f" {' '.join(option_parts)}"
+            if option_parts and self._match_text_seq(*option_parts):
+                name = f"{name} {' '.join(option_parts)}"
 
             self._match(TokenType.EQ)
             return self.expression(
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -391,6 +391,11 @@ class MySQL(Dialect):
             "WARNINGS": _show_parser("WARNINGS"),
         }
 
+        PROPERTY_PARSERS = {
+            **parser.Parser.PROPERTY_PARSERS,
+            "LOCK": lambda self: self._parse_property_assignment(exp.LockProperty),
+        }
+
         SET_PARSERS = {
             **parser.Parser.SET_PARSERS,
             "PERSIST": lambda self: self._parse_set_item_assignment("PERSIST"),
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -408,6 +408,41 @@ class Parser(parser.Parser):
             "SPATIAL": lambda self: self._parse_index_constraint(kind="SPATIAL"),
         }
 
+        TABLE_OPTION_PARSERS = {
```
georgesittas, line 411: I'm wondering if we should make new property classes for these, wdyt @tobymao? I just realized some of them are already supported (e.g. `ALGORITHM` and `CHECKSUM`), so perhaps a `_parse_csv(self._parse_property)` could do the trick here if we created expression types for the unsupported ones.

tobymao: not sure what you mean

tobymao (approved):

sqlglot/dialects/mysql.py
georgesittas: We can already parse some of these table options, e.g.

https://github.com/tobymao/sqlglot/blob/9595240a1c0f0e5ace9f67f31564e5d5edb9a9d2/tests/dialects/test_mysql.py#L73

The corresponding AST for this is:

```python
>>> import sqlglot
>>> sqlglot.parse_one("CREATE TABLE z (a INT) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARACTER SET=utf8 COLLATE=utf8_bin COMMENT='x'", "mysql")
Create(
  this=Schema(
    this=Table(
      this=Identifier(this=z, quoted=False)),
    expressions=[
      ColumnDef(
        this=Identifier(this=a, quoted=False),
        kind=DataType(this=Type.INT, nested=False))]),
  kind=TABLE,
  properties=Properties(
    expressions=[
      EngineProperty(
        this=Identifier(this=InnoDB, quoted=False)),
      AutoIncrementProperty(
        this=Literal(this=1, is_string=False)),
      CharacterSetProperty(
        this=Var(this=utf8),
        default=True),
      CollateProperty(
        this=Identifier(this=utf8_bin, quoted=False)),
      SchemaCommentProperty(
        this=Literal(this=x, is_string=True))]))
```

I'll see if I can refactor my PR to get rid of the table option parsers and reuse `_parse_properties` or something.

georgesittas: https://github.com/tobymao/sqlglot/pull/3027/commits/dba33c04c7d0d03ef3442998cc7159a1df2540c3

PR #3031, opened 2024-02-26
Fix: allow trailing comma in ORDER BY list

georgesittas:


tobymao (approved):

sqlglot/parser.py
```diff
@@ -3350,8 +3350,12 @@ def _parse_sort(self, exp_class: t.Type[E], token: TokenType) -> t.Optional[E]:
             return None
         return self.expression(exp_class, expressions=self._parse_csv(self._parse_ordered))
 
-    def _parse_ordered(self, parse_method: t.Optional[t.Callable] = None) -> exp.Ordered:
+    def _parse_ordered(
```
tobymao, line 3353: what about group and having?

georgesittas: group works fine, having doesn't expect a csv so it'd break
```python
>>> sqlglot.transpile("select * from t group by a,")
['SELECT * FROM t GROUP BY a']
```

PR #3034, opened 2024-02-27
Refactor: simplify parsing of keyword sequences as Vars

georgesittas:


sqlglot/parser.py
```diff
@@ -5692,11 +5677,29 @@ def _parse_set(self, unset: bool = False, tag: bool = False) -> exp.Set | exp.Co
...
+        for keywords in continuations or []:
+            if isinstance(keywords, str):
+                keywords = (keywords,)
+
+            if self._match_text_seq(*keywords):
+                option = f"{option} {' '.join(keywords)}"
+                break
+        else:
+            if continuations:
+                self.raise_error(f"Unknown option {option}")
```
tobymao, line 5700: is this more strict than it needs to be?

tobymao: yea

georgesittas: Hmm maybe it's actually better to keep it, at least for now. I tried the retreat approach but e.g. for T-SQL we try to match a `)` after the option parser is called and that fails so the error message is cryptic - it says we're expecting a `)` but in reality we just failed to parse the option and retreated without consuming the tokens.

tobymao: k

PR #3039, opened 2024-02-27
Fix(optimizer): unnest Subquery in the scope's _traverse_select

georgesittas:
This PR aims to address the following discrepancy.

Case with parentheses wrapping around query:
```python
from sqlglot import maybe_parse
from sqlglot.optimizer import build_scope, qualify

expression = maybe_parse("WITH baz AS (SELECT id FROM external) (SELECT o.value::INT AS value FROM bar AS o LEFT JOIN baz AS m ON o.id = m.id)", dialect="duckdb")
qualified = qualify.qualify(
    expression,
    dialect="duckdb",
    **{"validate_qualify_columns": False, "identify": False},
)
scope = build_scope(qualified)
print(scope.sources)
# {'baz': Scope<SELECT external.id AS id FROM external AS external>}
```

Same top-level query, without parentheses:
```python
from sqlglot import maybe_parse
from sqlglot.optimizer import build_scope, qualify

expression = maybe_parse("WITH baz AS (SELECT id FROM external) SELECT o.value::INT AS value FROM bar AS o LEFT JOIN baz AS m ON o.id = m.id", dialect="duckdb")
qualified = qualify.qualify(
    expression,
    dialect="duckdb",
    **{"validate_qualify_columns": False, "identify": False},
)
scope = build_scope(qualified)
print(scope.sources)
# {'baz': Scope<SELECT external.id AS id FROM external AS external>, 'o': Table(
#   this=Identifier(this=bar, quoted=False),
#   alias=TableAlias(
#     this=Identifier(this=o, quoted=False))), 'm': Scope<SELECT external.id AS id FROM external AS external>}
```

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -621,19 +621,23 @@ def _is_derived_table(expression: exp.Subquery) -> bool:
 def _traverse_tables(scope):
     sources = {}
 
+    scope_expression = scope.expression
+    if isinstance(scope_expression, exp.Subquery):
+        scope_expression = scope_expression.unnest()
+
     # Traverse FROMs, JOINs, and LATERALs in the order they are defined
     expressions = []
-    from_ = scope.expression.args.get("from")
+    from_ = scope_expression.args.get("from")
     if from_:
         expressions.append(from_.this)
 
-    for join in scope.expression.args.get("joins") or []:
+    for join in scope_expression.args.get("joins") or []:
         expressions.append(join.this)
 
-    if isinstance(scope.expression, exp.Table):
-        expressions.append(scope.expression)
+    if isinstance(scope_expression, exp.Table):
+        expressions.append(scope_expression)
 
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/scope.py
```diff
@@ -621,19 +621,23 @@ def _is_derived_table(expression: exp.Subquery) -> bool:
 def _traverse_tables(scope):
     sources = {}
 
+    scope_expression = scope.expression
+    if isinstance(scope_expression, exp.Subquery):
+        scope_expression = scope_expression.unnest()
```
tobymao, line 626: is unnesting always safe? what if we have joins or order bys etc? or will those never happen?

georgesittas: That's a good point, I'll need to investigate a bit more. It seems like `scope.tables` has a similar problem so will have to figure that out too.

georgesittas:
I'll probably have to go over this from the start based on our discussion with Barak, so I'll go ahead and close this PR.

PR #3051, opened 2024-02-28
Fix!(tsql): preserve CONVERT function

### LINKED ISSUE #3023: TSQL convert function incorrect conversion to sql string
I'm using the TSQL dialect and can confirm that it's an issue with the latest version 21.2.1

```python
from sqlglot import select

select_statement = select(
    "CONVERT(varchar, '2023-01-01 23:10:11.123', 121) AS current_datetime", dialect=TSQL
).from_("my_table")
sql = select_statement.sql(dialect=TSQL)
print(sql)
```

Output

```tsql
 SELECT CAST(FORMAT('2023-01-01 23:10:11.123', 'yyyy-MM-dd HH:mm:ss.S') AS VARCHAR(30)) AS current_datetime FROM my_table
```

Which isn't valid TSQL, apparently you can't pass it a varchar in the first position, you have to do something like this apparently with format.
Note .S should be .f repeated to however many digits of precision you need.

```tsql
DECLARE @d DATETIME = '2023-01-01 23:10:11.123'
SELECT CAST(FORMAT(@d, 'yyyy-MM-dd HH:mm:ss.fff') AS VARCHAR(30)) AS current_datetime
```

It would probably be much better that it not try and optimise the original convert statement which can take a varchar.

georgesittas:
Fixes #3023

Reference: https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -1041,12 +1041,6 @@ class Tokenizer(metaclass=_Tokenizer):
     def _text(self) -> str:
         return self.sql[self._start : self._current]
 
-    def peek(self, i: int = 0) -> str:
-        i = self._current + i
-        if i < self.size:
-            return self.sql[i]
-        return ""
-
     def _add(self, token_type: TokenType, text: t.Optional[str] = None) -> None:
         self._prev_token_line = self._line
 
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -609,9 +609,8 @@ class Hive(Dialect):
             return self.properties(properties, prefix=self.seg("TBLPROPERTIES"))
 
         def datatype_sql(self, expression: exp.DataType) -> str:
-            if (
-                expression.this in (exp.DataType.Type.VARCHAR, exp.DataType.Type.NVARCHAR)
-                and not expression.expressions
+            if expression.this in self.PARAMETERIZABLE_TEXT_TYPES and (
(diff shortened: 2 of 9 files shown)


sqlglot/tokens.py
```diff
@@ -1041,12 +1041,6 @@ def _advance(self, i: int = 1, alnum: bool = False) -> None:
     def _text(self) -> str:
         return self.sql[self._start : self._current]
 
-    def peek(self, i: int = 0) -> str:
```
tobymao, line 1044: is this not used?

tobymao (approved):

sqlglot/tokens.py
georgesittas: Yep, after the `decimal` refactor it was dead code, see for example the warning in the gh action deployment [job](https://github.com/tobymao/sqlglot/actions/runs/8088464225/job/22102560924):

```rust
warning: method `peek` is never used
   --> src/tokenizer.rs:181:8
    |
84  | impl<'a> TokenizerState<'a> {
    | --------------------------- method in this implementation
...
181 |     fn peek(&self, i: usize) -> Result<char, TokenizerError> {
    |        ^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: `sqlglotrs` (lib) generated 1 warning
```

PR #3052, opened 2024-02-29
Feat(test): handle NULL value in TPC-DS 

fool1280:


sqlglot/executor/context.py
```diff
@@ -78,7 +78,8 @@ def filter(self, condition) -> None:
     def sort(self, key) -> None:
         def sort_key(row: t.Tuple) -> t.Tuple:
             self.set_row(row)
-            return self.eval_tuple(key)
+            eval_result = self.eval_tuple(key)
+            return tuple((t is None, t) for t in eval_result)
```
tobymao, line 82:
```suggestion
            return tuple((t is None, t) for t in self.eval_tuple(key))
```

sqlglot/executor/context.py
```diff
@@ -78,7 +78,8 @@ def filter(self, condition) -> None:
     def sort(self, key) -> None:
         def sort_key(row: t.Tuple) -> t.Tuple:
             self.set_row(row)
-            return self.eval_tuple(key)
+            eval_result = self.eval_tuple(key)
```
tobymao, line 81: delete this one

PR #3055, opened 2024-02-29
fix: Fix Hive UnixToTime regression, README stale results

VaggelisD:
In the process of validating the README examples, 2 diffs with current SQLGlot results were found:

1) The first example transpiled `EPOCH_MS (DuckDB) -> FROM_UNIXTIME (Hive)` but the respective AST Node `exp.UnixToTime` was not generating properly in Hive

2) The parse error examples in the existing README have mismatches in the `col` position of the missing token (right parenthesis) and the surrounding context does not seem right. In the latest SQLGLot version, the parse error diagnostics have improved but the examples have not been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,12 +202,12 @@ When the parser detects an error in the syntax, it raises a ParseError:
 
 ```python
 import sqlglot
-sqlglot.transpile("SELECT foo FROM(SELECT baz FROM t")
+sqlglot.transpile("SELECT foo FROM (SELECT baz FROM t")
 ```
 
 ```
 sqlglot.errors.ParseError: Expecting ). Line 1, Col: 33.
-  SELECT foo FROM(SELECT baz FROM t
+  SELECT foo FROM (SELECT baz FROM t
                                   ~
 ```
 
@@ -216,7 +216,7 @@ Structured syntax errors are accessible for programmatic use:
 ```python
 import sqlglot
 try:
-    sqlglot.transpile("SELECT foo FROM(SELECT baz FROM t")
+    sqlglot.transpile("SELECT foo FROM (SELECT baz FROM t")
 except sqlglot.errors.ParseError as e:
     print(e.errors)
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -206,9 +206,9 @@ sqlglot.transpile("SELECT foo( FROM bar")
 ```
 
 ```
-sqlglot.errors.ParseError: Expecting ). Line 1, Col: 13.
+sqlglot.errors.ParseError: Expecting ). Line 1, Col: 20.
   select foo( FROM bar
-              ~~~~
+                   ~~~
```
georgesittas, line 211: The reason this column changed is because the parser became more lenient: you can now do `SELECT foo (FROM bar)` and that parses successfully because `FROM table` is valid DuckDB syntax.

This example may be a bit confusing though, it's not trivial to understand why SQLGlot underlines `bar` instead of `TABLE`. What do you say about changing this and the next one ("structured error") to this example instead?

```
>>> sqlglot.parse_one("SELECT foo FROM (SELECT x FROM t")
...
sqlglot.errors.ParseError: Expecting ). Line 1, Col: 32.
  SELECT foo FROM (SELECT x FROM t
                                 ~
```

georgesittas (approved):

README.md
VaggelisD: Hmm, I thought `FROM <table>` was already valid at that point, otherwise I'd expect the initial example to have a different parse error such as `Expected TABLE / SELECT` right after the opening parenthesis. 

The example you're proposing does seem easier to digest as it's more ansi-like /  not tied to DuckDB per se, I'll use it in my next commit

georgesittas: Since `FROM` is a reserved keyword, I believe most other dialects would point to `FROM` as the culprit, e.g.:

```
sqlite> SELECT foo (FROM bar;
Parse error: near "FROM": syntax error
  SELECT foo (FROM bar;
              ^--- error here
```

PR #3057, opened 2024-02-29
Fix: ensure UDF identifier quotes are preserved

georgesittas:


sqlglot/optimizer/simplify.py
```diff
@@ -1219,6 +1219,11 @@ def gen(expression: t.Any) -> str:
 }
 
 
+def _anonymous(e: exp.Anonymous) -> str:
+    name = e.this.upper() if isinstance(e.this, str) else e.sql()
```
tobymao, line 1223: why are you calling e.sql() here? that ruins the purpose of gen

georgesittas: Oversight, will fix - thanks

georgesittas: done

PR #3069, opened 2024-03-01
Fix: drop CLUSTER/DISTRIBUTED/SORT BY modifiers when unsupported

### LINKED ISSUE #3066: Transpile spark => presto should drop "CLUSTER/DISTRIBUTE/SORT BY" clauses as no equivalent in presto
**Before you file an issue**
Transpile spark => presto should drop "CLUSTER/DISTRIBUTE/SORT BY" clauses as no equivalent in presto

Following queries work in spark. However the query after transpile to presto fails in presto because presto syntax does not have CLUSTER/DISTRIBUTE/SORT BY.  Note SORT BY is different from ORDER BY.

```
sqlglot.transpile("select * from abc sort by id_user", "spark", "presto")
sqlglot.transpile("select * from abc cluster by id_user", "spark", "presto")
sqlglot.transpile("select * from abc distribute by id_user", "spark", "presto")
```

**Official Documentation**

Trino documentation which shows NO existence of CLUSTER/DISTRIBUTE/SORT BY
https://trino.io/docs/current/sql/select.html#

SPARK documentation which shows syntax includes CLUSTER/DISTRIBUTE/SORT BY
https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select.html

georgesittas:
Fixes #3066

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -137,6 +137,13 @@ class _Dialect(type):
         if enum not in ("", "bigquery"):
             klass.generator_class.SELECT_KINDS = ()
 
+        if enum not in ("", "hive", "spark", "spark2"):
+            modifier_transforms = klass.generator_class.AFTER_HAVING_MODIFIER_TRANSFORMS.copy()
+            for modifier in ("cluster", "distribute", "sort"):
+                modifier_transforms.pop(modifier, None)
+
+            klass.generator_class.AFTER_HAVING_MODIFIER_TRANSFORMS = modifier_transforms
+
         if not klass.SUPPORTS_SEMI_ANTI_JOIN:
             klass.parser_class.TABLE_ALIAS_TOKENS = klass.parser_class.TABLE_ALIAS_TOKENS | {
                 TokenType.ANTI,
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -350,6 +350,18 @@ class Generator(metaclass=_Generator):
         "YEARS": "YEAR",
     }
 
+    AFTER_HAVING_MODIFIER_TRANSFORMS = {
+        "cluster": lambda self, e: self.sql(e, "cluster"),
+        "distribute": lambda self, e: self.sql(e, "distribute"),
+        "qualify": lambda self, e: self.sql(e, "qualify"),
+        "sort": lambda self, e: self.sql(e, "sort"),
+        "windows": lambda self, e: (
+            self.seg("WINDOW ") + self.expressions(e, key="windows", flat=True)
+            if e.args.get("windows")
+            else ""
+        ),
+    }
+
     TOKEN_MAPPING: t.Dict[TokenType, str] = {}
 
     STRUCT_DELIMITER = ("<", ">")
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -137,6 +137,13 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
         if enum not in ("", "bigquery"):
             klass.generator_class.SELECT_KINDS = ()
 
+        if enum not in ("", "hive", "spark", "spark2"):
```
tobymao, line 140: what about databricks?

sqlglot/dialects/dialect.py
```diff
@@ -137,6 +137,13 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
...
+            modifier_transforms = klass.generator_class.AFTER_HAVING_MODIFIER_TRANSFORMS.copy()
+            for modifier in ("cluster", "distribute", "sort"):
+                modifier_transforms.pop(modifier, None)
```
tobymao, line 143: why do you need the None here? you shouldn't be popping twice right

sqlglot/dialects/dialect.py
```diff
@@ -137,6 +137,13 @@ def get_start_end(token_type: TokenType) -> t.Tuple[t.Optional[str], t.Optional[
         if enum not in ("", "bigquery"):
             klass.generator_class.SELECT_KINDS = ()
 
+        if enum not in ("", "hive", "spark", "spark2"):
+            modifier_transforms = klass.generator_class.AFTER_HAVING_MODIFIER_TRANSFORMS.copy()
```
tobymao, line 141: why do you need to copy?

sqlglot/generator.py
```diff
@@ -2111,17 +2123,7 @@ def offset_limit_modifiers(
         ]
 
     def after_having_modifiers(self, expression: exp.Expression) -> t.List[str]:
```
tobymao, line 2125: i think we don't need this method anymore which is nice

tobymao: we can improve now select, and remove other things like

prewhere, laterals, connect, match etc

sqlglot/dialects/dialect.py
georgesittas: I missed the "various other clauses" part in their [docs](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select.html) and assumed they didn't support these modifiers. Seems like I was wrong so gonna add that as well.

sqlglot/dialects/dialect.py
georgesittas: Because otherwise you mutate the parent class' dict

sqlglot/dialects/dialect.py
georgesittas: The way this is currently implemented, if you load `Hive` first then it will pop these off and then `Spark2` which inherits from it won't have anything to pop, so it raises a `KeyError`.

sqlglot/generator.py
georgesittas: Can take a look. Btw, one detail with these dicts is that we need to make sure we remember to define the kv pairs in the right order. For example, I intentionally put `"sort"` after `"distribute"` here, because that's the correct order.

sqlglot/dialects/dialect.py
tobymao: so then don't put spark, and just put in hive?

sqlglot/generator.py
tobymao: just add a test

sqlglot/dialects/dialect.py
georgesittas: Ah apologies, I got confused - we actually _don't_ wanna pop the modifiers for these dialects. The issue arises for the other dialects, e.g. you pop in postgres and so redshift will raise the key error.

sqlglot/generator.py
georgesittas: Cool we should have a few tests already

tobymao (approved):

PR #3072, opened 2024-03-03
feat(snowflake): parse CREATE SEQUENCE

### LINKED ISSUE #2954: support parsing sequence related statements
When using fakesnow for unit testing we discovered that `CREATE SEQUENCE seq1` is currently not supported by sqlglot. As is accessing sequences.

It would be terrific it sqlglot could also transpile sequence creation, access and deletion. 

We have currently mocked the parts of our code that relied on the sequences being present. Since the sequence is only created implicitly via SQLalchemy, we also considered changing the table definitions during testing.

**Additional context**
at least between duckdb and snowflake there are bigger differences in how to access a sequence vs creating one:
duckdb: `select nextval(seq1)`
snowflake: `select seq1.nextval`

tekumara:
see https://docs.snowflake.com/en/sql-reference/sql/create-sequence

partially resolves #2954

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1214,6 +1214,8 @@ class Create(DDL):
         "begin": False,
         "end": False,
         "clone": False,
+        "start": False,
+        "increment": False,
     }
 
     @property
@@ -3617,6 +3619,14 @@ class Star(Expression):
     @property
     def output_name(self) -> str:
         return self.name
+
+
+class Increment(Expression):
+    arg_types = {"increment": True}
+
+
+class Start(Expression):
+    arg_types = {"this": True}
 
 
 class Parameter(Condition):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3622,7 +3622,7 @@ class Star(Expression):
 
 
 class Increment(Expression):
-    arg_types = {"increment": True}
+    arg_types = {"this": True}
 
 
 class Start(Expression):
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -958,7 +958,13 @@ class Generator(metaclass=_Generator):
         clone = self.sql(expression, "clone")
         clone = f" {clone}" if clone else ""
 
-        expression_sql = f"CREATE{modifiers} {kind}{exists_sql} {this}{properties_sql}{expression_sql}{postexpression_props_sql}{index_sql}{no_schema_binding}{clone}"
+        start = self.sql(expression, "start")
+        start = f" {start}" if start else ""
+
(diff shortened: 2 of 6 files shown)


sqlglot/expressions.py
```diff
@@ -3619,6 +3621,14 @@ def output_name(self) -> str:
         return self.name
 
 
+class Increment(Expression):
+    arg_types = {"this": True}
+
+
+class Start(Expression):
+    arg_types = {"this": True}
+
```
georgesittas, line 3630: We don't need these, right? You can just store the expressions as is in the corresponding arguments.

Btw, you could also do `pass` here.

sqlglot/generator.py
```diff
@@ -1541,6 +1547,14 @@ def historicaldata_sql(self, expression: exp.HistoricalData) -> str:
         expr = self.sql(expression, "expression")
         return f"{this} ({kind} => {expr})"
 
+    def start_sql(self, expression: exp.Start) -> str:
+        this = self.sql(expression, "this")
+        return f"START WITH {this}"
+
+    def increment_sql(self, expression: exp.Increment) -> str:
+        this = self.sql(expression, "this")
+        return f"INCREMENT BY {this}"
```
georgesittas, line 1556: So we'd get rid of these as well and do this in `create_sql`.

georgesittas:
Hey @tekumara, this may be good for now but we need to ensure `CREATE SEQUENCE` doesn't fail to parse in other dialects. E.g. right now we accept the following Oracle statement, but parse it as a `Command`:

```sql
CREATE SEQUENCE customers_seq
 START WITH     1000
 INCREMENT BY   1
 NOCACHE
 NOCYCLE;
```

Please check & test a few more syntax specs that aren't covered by this PR, as well as the remaining clauses in Snowflake's syntax. For example, I believe the parser still doesn't recognize the `=`, the preceding `WITH` and the `ORDER` / `NOORDER` keywords. We should also make sure these are still handled gracefully.

Some references I found that may be useful:
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571
- https://www.postgresql.org/docs/current/sql-createsequence.html
- https://learn.microsoft.com/en-us/sql/t-sql/statements/create-sequence-transact-sql?view=sql-server-ver16
- https://duckdb.org/docs/sql/statements/create_sequence.html

VaggelisD:
Hi @tekumara,

Thanks for your contribution! It's now merged and in a follow up iteration of mine I'll address @georgesittas 's comments in order to test & generalize `CREATE SEQUENCE` across dialects if you want to follow this issue.

PR #3073, opened 2024-03-03
Fix!(sqlite): allow ON CONFLICT constraint after UNIQUE

### LINKED ISSUE #3070: SQLite ParseError: Unique Table Constraint- cannot parse On Conflict Clause
sqlglot Verison is 22.2.0

#### MVE
```python
import sqlglot

inputstring = """CREATE TABLE a (
    b,
    c,
    UNIQUE (b, c) ON CONFLICT IGNORE
);"""

print(sqlglot.parse(inputstring, dialect='sqlite'))
```

#### Raises
```
sqlglot.errors.ParseError: Expecting ). Line 4, Col: 20.
  CREATE TABLE a (
        b,
    c,
    UNIQUE (b, c) ON CONFLICT IGNORE
    );
```
(**ON** is underlined in terminal)

#### Official Docs
[SQLite Create Table Docs](https://www.sqlite.org/lang_createtable.html)

Tested at on the [official fiddle](https://sqlite.org/fiddle/) to make sure SQL was a valid statement using
```sqlite
CREATE TABLE a (
    b,
    c,
    UNIQUE (b, c) ON CONFLICT IGNORE
);

INSERT INTO a(b,c) VALUES (1,1), (2,1), (1,1);

SELECT * FROM a;
```

Output is two rows as expected: (1,1), (2,1)

georgesittas:
Fixes #3070

Reference: https://www.sqlite.org/lang_createtable.html

Screenshot 2024-03-03 at 4 23 26 PM

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1605,7 +1605,7 @@ class TitleColumnConstraint(ColumnConstraintKind):
 
 
 class UniqueColumnConstraint(ColumnConstraintKind):
-    arg_types = {"this": False, "index_type": False}
+    arg_types = {"this": False, "index_type": False, "on_conflict": False}
 
 
 class UppercaseColumnConstraint(ColumnConstraintKind):
@@ -1883,8 +1883,8 @@ class OnConflict(Expression):
     arg_types = {
         "duplicate": False,
         "expressions": False,
-        "nothing": False,
-        "key": False,
+        "action": False,
+        "conflict_keys": False,
         "constraint": False,
     }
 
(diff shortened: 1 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -1883,8 +1883,8 @@ class OnConflict(Expression):
     arg_types = {
         "duplicate": False,
         "expressions": False,
-        "nothing": False,
```
tobymao, line 1886: breaking change?

tobymao (approved):

sqlglot/expressions.py
georgesittas: yeah thanks

PR #3090, opened 2024-03-06
feat: Generalizing CREATE SEQUENCE

VaggelisD:
Fixes #3072 

Generalize `CREATE SEQUENCE` among Oracle, T-SQL, Postgres, DuckDB, Snowflake.

Design Notes:
-----------------
- Added `UnloggedProperty` that exists in Postgres for tables and sequences
- Made parsing of `CREATE SEQUENCE` order agnostic as dialects had differences between them
- Non-repeating options from Oracle's dialect were encapsulated in an `OPTIONS` dictionary to simplify parsing
- Future note: It seems that some overlapping options are also parsed in `exp.GeneratedAsIdentityColumnConstraint`, maybe a refactor between them is favorable?
- 2nd iteration: Added `GlobalProperty`(post-create properly) and `SharingProperty` (Oracle post-name) to accommodate for regressions on a command identity , which enabled it's full roundtrip. 

Docs
--------
- [Postgres](https://www.postgresql.org/docs/current/sql-createsequence.html)
- [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571)
- [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-sequence-transact-sql?view=sql-server-ver16)
- [DuckDB](https://duckdb.org/docs/sql/statements/create_sequence.html)
- [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-sequence)


sqlglot/parser.py
```diff
@@ -1551,10 +1574,55 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+        exists = self._parse_exists(not_=True)
+        this = self._parse_table(schema=True)
+
+        seq = exp.Create(
+            this=this,
+            exists=exists,
+            replace=replace,
+            kind="SEQUENCE",
+            properties=properties,
+        )
```
georgesittas, line 1589: This doesn't look correct, we shouldn't be duplicating logic like that in here. Probably better to keep the branch in `_parse_create` and factor the rest of the logic (options, properties etc) out into a separate method.

sqlglot/parser.py
```diff
@@ -1551,10 +1574,55 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+            elif self._match(TokenType.START_WITH) or self._match_text_seq("START"):
+                seq.set("start", self._parse_term())
+            elif self._match_text_seq("CACHE"):
+                # T-SQL allows empty CACHE which is initialized dynamically
+                seq.set("cache", self._parse_term() or exp.Var(this=""))
+            elif self._match_text_seq("OWNED", "BY"):
+                owned = None if self._match_text_seq("NONE") else self._parse_column()
+                seq.set("owned", owned)
+            elif self._match_text_seq("COMMENT", "="):
+                seq.set("comment", self._parse_string())
```
georgesittas, line 1612: We shouldn't be duplicating logic here too, we already handle this as a DDL property:

```python
>>> import sqlglot
>>> sqlglot.parse_one("CREATE TABLE z (a INT) COMMENT='x'")
Create(
  this=Schema(
    this=Table(
      this=Identifier(this=z, quoted=False)),
    expressions=[
      ColumnDef(
        this=Identifier(this=a, quoted=False),
        kind=DataType(this=Type.INT, nested=False))]),
  kind=TABLE,
  properties=Properties(
    expressions=[
      SchemaCommentProperty(                                 <--- this one
        this=Literal(this=x, is_string=True))]))
```

I'm wondering whether we can treat these "options" as "properties" instead, so that we can just repeat the same pattern as in other DDLs: `extend_props(self._parse_properties())`. Maybe this will allow us to get rid of the extra logic, like you mention here:

> Future note: It seems that some overlapping options are also parsed in exp.GeneratedAsIdentityColumnConstraint, maybe a refactor between them is favorable?

There is indeed some overlap and either now or in the future we should figure out if we can reduce the duplication.

sqlglot/parser.py
```diff
@@ -1551,10 +1574,55 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+                seq.set("data_type", self._parse_types())
+            elif self._match_text_seq("INCREMENT"):
+                self._match_text_seq("BY")
+                seq.set("increment", self._parse_term())
+            elif self._match_text_seq("MINVALUE"):
+                seq.set("minvalue", self._parse_term())
+            elif self._match_text_seq("MAXVALUE"):
+                seq.set("maxvalue", self._parse_term())
+            elif self._match(TokenType.START_WITH) or self._match_text_seq("START"):
+                seq.set("start", self._parse_term())
```
georgesittas, line 1604: Don't you need `_parse_equality` here (and possibly elsewhere) to consume the `EQ`? From Snowflake's docs:

```
[ START [ WITH ] [ = ] <initial_value> ]
```

sqlglot/parser.py
```diff
@@ -1551,10 +1574,55 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+            elif self._match_text_seq("CACHE"):
+                # T-SQL allows empty CACHE which is initialized dynamically
+                seq.set("cache", self._parse_term() or exp.Var(this=""))
```
georgesittas, line 1607: I'm not sure if I'd use an empty `Var` as a proxy for the single `CACHE` keyword here, maybe something like `True` would be better? On the other hand, this section may get refactored if we treat these options as properties.

sqlglot/parser.py
```diff
@@ -1551,10 +1574,55 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+            elif self._match_text_seq("OWNED", "BY"):
+                owned = None if self._match_text_seq("NONE") else self._parse_column()
+                seq.set("owned", owned)
```
georgesittas, line 1610: Don't we need to generate `OWNED BY NONE` here?

sqlglot/parser.py
```diff
@@ -1551,10 +1574,55 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+            elif self._match_text_seq("COMMENT", "="):
+                seq.set("comment", self._parse_string())
+            elif self._match_text_seq("SHARING", "="):
+                seq.set(
+                    "sharing",
+                    self._prev.text if self._match_texts(("METADATA", "DATA", "NONE")) else None,
```
georgesittas, line 1616:
```suggestion
                    self._match_texts(("METADATA", "DATA", "NONE")) and self._prev.text,
```

georgesittas (commented):
Maybe let's try to refactor this by leveraging `_parse_properties()` (see comment).

sqlglot/parser.py
georgesittas: Ok I guess we don't, but maybe document this accordingly?

> OWNED BY NONE, the default, specifies that there is no such association.

sqlglot/parser.py
VaggelisD: Good catch, Snowflake does allow for an optional `EQ`. As a general note which can apply in future PRs too, since other dialects do not seem to specify `EQ` in their syntax and it's optional anyways, would it be better to consume the `EQ` token (just for Snowflake) and `_parse_term()` - as a way of canonicalizing `CREATE SEQUENCE` while also simplifying transpilation - or preserve the full roundtrip by going with `_parse_equality()` instead?

sqlglot/parser.py
VaggelisD: Yes, storing the `NONE` is superfluous in this case, but since you brought I think it's probably simpler to just preserve the token in this case instead of consuming and ignoring in such cases. The code is the exactly same (creating an `exp.Var('NONE')` instead of `None` without the confusion / need to validate it by searching for the docs, and - although maybe extreme - it might protect us in cases were the respective dialect changes it's default initialized value which would turn this simplification into a historical bug.

sqlglot/parser.py
VaggelisD: With `CREATE SEQUENCE` being it's own AST node, does this comment still hold? Bear in mind that we might have to parse some modifiers (e.g. `TEMPORARY`, `OR REPLACE`) before we get to the `SEQUENCE` token which would dispatch `_parse_create_sequence()`, and if we allow the control flow to continue to `_parse_create` we may have to multiplex between `exp.Create` and `exp.CreateSequence` parts. Also, since dialects have some differences between them, I thought that by parsing under a `while` would require that the expression is created beforehand, so I can utilize `exp.set(...)`.

georgesittas: The suggestion is to parse this into `exp.Create` like all other DDLs. Let's sync on Slack for how we can clean this up.

sqlglot/parser.py
```diff
@@ -1432,7 +1432,7 @@ def _parse_exists(self, not_: bool = False) -> t.Optional[bool]:
             and self._match(TokenType.EXISTS)
         )
 
-    def _parse_create(self) -> exp.Create | exp.Command:
+    def _parse_create(self) -> exp.Expression:
```
georgesittas, line 1435: Why did this change? We're still returning either a `Create` or a `Command`.

sqlglot/parser.py
```diff
@@ -1512,6 +1510,12 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
                 schema=True, is_db_reference=create_token.token_type == TokenType.SCHEMA
             )
 
+            sequence_opts = (
+                self._parse_sequence_properties()
+                if create_token.token_type == TokenType.SEQUENCE
+                else None
+            )
+
```
georgesittas, line 1518: This isn't really what I had in mind - the idea was to handle any sequence properties inside of `_parse_property` so that they would be appended to `Create.args["properties"]` instead of in a new arg.

sqlglot/parser.py
```diff
@@ -1574,19 +1578,11 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
-        this = self._parse_table(schema=True)
-
-        seq = exp.Create(
-            this=this,
-            exists=exists,
-            replace=replace,
-            kind="SEQUENCE",
-            properties=properties,
-        )
+    def _parse_sequence_properties(self) -> exp.SequenceProperties:
```
georgesittas, line 1584: I thought we were moving `AS <type>` inside `_parse_create`?

sqlglot/parser.py
```diff
@@ -1595,30 +1591,38 @@ def _parse_create_sequence(self, replace, properties) -> exp.Create | exp.Comman
...
             elif self._match(TokenType.START_WITH) or self._match_text_seq("START"):
+                self._match_text_seq("=")
                 seq.set("start", self._parse_term())
             elif self._match_text_seq("CACHE"):
                 # T-SQL allows empty CACHE which is initialized dynamically
-                seq.set("cache", self._parse_term() or exp.Var(this=""))
+                cache: t.Optional[exp.Expression | bool] = self._parse_term()
+                if cache and not isinstance(cache, exp.Literal):
+                    self._retreat(self._index - 1)
+                    cache = True
```
georgesittas, line 1608: why are we retreating here? can't we just do `_parse_number`?

sqlglot/parser.py
```diff
@@ -1595,30 +1591,38 @@ def _parse_create_sequence(self, replace, properties) -> exp.Create | exp.Comman
...
+                seq.set("cache", cache)
             elif self._match_text_seq("OWNED", "BY"):
-                owned = None if self._match_text_seq("NONE") else self._parse_column()
+                # "OWNED BY NONE" is the default
+                owned = self._prev.text if self._match_text_seq("NONE") else self._parse_column()
```
georgesittas, line 1612: Imo there's no point in carrying around redundant info here, since `OWNED BY NONE` can be omitted. I'd personally just set it to `None`.

sqlglot/parser.py
```diff
@@ -1595,30 +1591,38 @@ def _parse_create_sequence(self, replace, properties) -> exp.Create | exp.Comman
...
                 seq.set("owned", owned)
-            elif self._match_text_seq("COMMENT", "="):
-                seq.set("comment", self._parse_string())
+            elif self._match_text_seq("COMMENT"):
+                seq.set("comment", self._parse_property_assignment(exp.SchemaCommentProperty))
```
georgesittas, line 1615: If we updated the property parser and rearranged things properly we wouldn't have to deal with `COMMENT` and `SHARING` here - they'd be automatically parsed as properties.

sqlglot/parser.py
```diff
@@ -1595,30 +1591,38 @@ def _parse_create_sequence(self, replace, properties) -> exp.Create | exp.Comman
...
                 )
             else:
                 opt = self._parse_var_from_options(self.CREATE_SEQUENCE)
                 if opt:
                     options.append(opt)
+                else:
+                    break
 
         seq.set("options", options if options else None)
         return seq
```
georgesittas, line 1628: This should return `None` if the sequence option is empty, otherwise you create an incorrect AST node (regardless of the generator handling the empty `exp.SequenceProperties` case).

georgesittas (commented):
A few more comments @VaggelisD

mpf82:
Great idea, but is the idea "only" to allow parsing and generating within the same dialect?

Because, if you also want to support transpiling, by looking at the code and the tests, I am not sure if Oracle's ``NOCYCLE`` and ``NOCACHE`` are correctly supported or will be correctly transpiled. 

Side note: ``NOSHARD`` is added twice in ``CREATE_SEQUENCE``

georgesittas:
Hey Mike, thanks for the review / feedback 👋 

>  is the idea "only" to allow parsing and generating within the same dialect?

For now, that's what we're aiming for. The goal is to allow fine-grained parsing of the `CREATE SEQUENCE` statement, so that at minimum it can be roundtripped correctly for a given dialect, while also allowing one to analyze its AST. Some basic variants of this statement should already be transpilable as is, but aiming to cover all possible options is too much work, given the number of dialects supporting `CREATE SEQUENCE`, and the ROI is questionable imo.

Perhaps at some point in the future we could refactor this to control more easily which options / properties are generated per dialect to improve the transpilation.

P.S. just double-checking, but is this related to transpilation, or did you also notice an issue for the Oracle -> Oracle case?

> I am not sure if Oracle's NOCYCLE and NOCACHE are correctly supported

sqlglot/generator.py
```diff
@@ -972,14 +979,44 @@ def create_sql(self, expression: exp.Create) -> str:
...
+                start,
+                increment,
+                minvalue,
+                maxvalue,
+                cache_str,
+                options,
+                owned,
+            )
+        )
+        return sequence_options.lstrip()
```
georgesittas, line 1019:
```suggestion
        return f"{start}{increment}{minvalue}{maxvalue}{cache_str}{options}{owned}".lstrip()
```

sqlglot/parser.py
```diff
@@ -972,6 +976,30 @@ class Parser(metaclass=_Parser):
     )
     CONFLICT_ACTIONS["DO"] = ("NOTHING", "UPDATE")
 
+    CREATE_SEQUENCE: OPTIONS_TYPE = {
+        "SCALE": (("EXTEND"), ("NOEXTEND")),
+        "SHARD": (("EXTEND"), ("NOEXTEND")),
+        "NO": (("CYCLE"), ("CACHE"), ("MAXVALUE"), ("MINVALUE")),
```
georgesittas, line 982:
```suggestion
        "SCALE": ("EXTEND", "NOEXTEND"),
        "SHARD": ("EXTEND", "NOEXTEND"),
        "NO": ("CYCLE", "CACHE", "MAXVALUE", "MINVALUE"),
```

sqlglot/parser.py
```diff
@@ -1497,7 +1524,14 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
 
-            expression = self._parse_ddl_select()
+            expression = (
+                self._parse_type()
+                if create_token.token_type == TokenType.SEQUENCE
+                else self._parse_ddl_select()
+            )
+
+            if create_token.token_type == TokenType.SEQUENCE:
+                extend_props(self._parse_properties())
```
georgesittas, line 1534:
```suggestion
            if create_token.token_type == TokenType.SEQUENCE:
                expression = self._parse_types()
                extend_props(self._parse_properties())
            else:
                expression = self._parse_ddl_select()
```

sqlglot/parser.py
```diff
@@ -1598,7 +1660,8 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
 
         if not self._match(TokenType.EQ):
             self._retreat(index)
-            return None
+            seq_properties = self._parse_sequence_properties()
+            return seq_properties or None
```
georgesittas, line 1664:
```suggestion
            return self._parse_sequence_properties()
```

sqlglot/parser.py
```diff
@@ -1551,10 +1579,44 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+                seq.set("minvalue", self._parse_term())
+            elif self._match_text_seq("MAXVALUE"):
+                seq.set("maxvalue", self._parse_term())
+            elif self._match(TokenType.START_WITH) or self._match_text_seq("START"):
+                self._match_text_seq("=")
+                seq.set("start", self._parse_term())
+            elif self._match_text_seq("CACHE"):
+                # T-SQL allows empty CACHE which is initialized dynamically
+                cache: t.Optional[exp.Expression | bool] = self._parse_number() or True
+                seq.set("cache", cache)
```
georgesittas, line 1605:
```suggestion
                seq.set("cache", self._parse_number() or True)
```

sqlglot/parser.py
```diff
@@ -1551,10 +1579,44 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
...
+            elif self._match_text_seq("OWNED", "BY"):
+                # "OWNED BY NONE" is the default
+                owned = None if self._match_text_seq("NONE") else self._parse_column()
+                seq.set("owned", owned)
```
georgesittas, line 1609:
```suggestion
                seq.set("owned", None if self._match_text_seq("NONE") else self._parse_column())
```

mpf82:
> P.S. just double-checking, but is this related to transpilation, or did you also notice an issue for the Oracle -> Oracle case?

Aftert looking again, I guess I mean just transpiling.

I was kinda confused when looking at 

```python
        cache = expression.args.get("cache")
        if cache is None:
            cache_str = ""
        elif cache is True:
            cache_str = " CACHE"
        else:
            cache_str = f" CACHE {cache}"
```

and not seeing ``NOCACHE`` in the code and also not seeing a test case.
But other options, such as ``NOSCALE`` seem to validate just fine.

 So ``NOCACHE`` and others must be generated from ``options = self.expressions(expression, key="options", flat=True, sep=" ")``, right? 
Then I think I understand how it works.

Which also means that if you construct something like ``"CREATE SEQUENCE seq START WITH 1 NO CACHE NOCACHE CACHE 10"`` it will also also "validate", correct?
Which is fine, since sqlglot's a parser, not a validator.

georgesittas:
That's correct, these options are generated using the `expressions` call you mentioned.

georgesittas (approved):
Few final cleanup suggestions, nice work @VaggelisD.

PR #3092, opened 2024-03-06
fix: dataframe optimize user input

### LINKED ISSUE #3091: spark sql SELECT MAX without column alias throws error
```sql
spark = SparkSession.builder.config("sqlframe.dialect", "spark").getOrCreate()
df = spark.sql("""
    SELECT
        MAX(col)
    FROM (SELECT 1 as col) t
    """)
```
throws `sqlglot.errors.ParseError: No expression was parsed from ''` because its `name` [here](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dataframe/sql/dataframe.py#L173) is an empty string. This seems to be an issue with expressions that inherit from [Func](https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py#L4330) (MIN, MAX, ABS, etc). Changing [Max](https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py#L5216) to inherit from [Condition](https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py#L903) directly fixes the issue. This isn't a problem for some more complex expressions that leverage multiple inheritance like [DateAdd](https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py#L4700).

**Official Documentation**
https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select.html#parameters - aliases are optional in spark sql

eakmanrq:
Prior to this change user inputted queries were not optimized and therefore could lack expected aliases. This is now fixed.

I plan on making a follow up PR to write spark specific rules to optimizer so it will qualify columns based on how spark does it rather than sqlglot's format.

Fixes #3091

### DIFF
--- a/sqlglot/dataframe/sql/session.py
+++ b/sqlglot/dataframe/sql/session.py
@@ -12,6 +12,8 @@ from sqlglot.dataframe.sql.readwriter import DataFrameReader
 from sqlglot.dataframe.sql.types import StructType
 from sqlglot.dataframe.sql.util import get_column_mapping_from_schema_input
 from sqlglot.helper import classproperty
+from sqlglot.optimizer import optimize as optimize_func
+from sqlglot.optimizer.qualify_columns import quote_identifiers
 
 if t.TYPE_CHECKING:
     from sqlglot.dataframe.sql._typing import ColumnLiterals, SchemaInput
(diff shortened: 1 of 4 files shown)


sqlglot/dataframe/sql/session.py
```diff
@@ -12,6 +12,8 @@
 from sqlglot.dataframe.sql.types import StructType
 from sqlglot.dataframe.sql.util import get_column_mapping_from_schema_input
 from sqlglot.helper import classproperty
+from sqlglot.optimizer import optimize as optimize_func
```
georgesittas, line 15: We can now omit the alias here, since there's no `optimize` arg shadowing the import anymore.

sqlglot/dataframe/sql/session.py
```diff
@@ -104,8 +106,15 @@ def createDataFrame(
         sel_expression = exp.Select(**select_kwargs)
         return DataFrame(self, sel_expression)
 
+    def _optimize(
+        self, expression: exp.Expression, dialect: t.Optional[Dialect] = None
+    ) -> exp.Expression:
+        dialect = dialect or self.dialect
+        quote_identifiers(expression, dialect=dialect)
```
georgesittas, line 113: Why did we need to quote here prior to optimizing? Quotes are also added by the optimizer.

georgesittas (approved):

sqlglot/dataframe/sql/session.py
eakmanrq: Some tests fail if I remove it. This is likely covering up a bigger underlying issue that I will look into later.

PR #3094, opened 2024-03-07
Fix!(optimizer): promote typed div to double if operands are non-float

georgesittas:


sqlglot/expressions.py
```diff
@@ -3778,24 +3778,37 @@ class Type(AutoName):
...
+        Type.UINT128,
+        Type.UINT256,
+        Type.UMEDIUMINT,
+        Type.USMALLINT,
+        Type.UTINYINT,
     }
 
     FLOAT_TYPES = {
-        Type.FLOAT,
+        Type.BIGDECIMAL,
```
barakalon, line 3805: I think DECIMAL et al. are definitively _not_ floating point, right?

barakalon: Do we want something like `REAL_TYPES`?

georgesittas: Donezo - https://github.com/tobymao/sqlglot/pull/3094/commits/18f513c413287023fe059e4ec1ba255a46294eaf

barakalon (approved):

PR #3096, opened 2024-03-07
Refactor!(optimizer): only create scopes for the DDL queries

### LINKED ISSUE #3086: Lineage  - not downstreaming to subquery expression
I'm trying to generate column level lineage from given INSERT statement.

I'm exploring column RUN_DATE,  and expect to get expression for last available downstream.

I expect to get same expression of last step in walk for sql_nok as in sql_ok.

Name: sqlglot
Version: 22.2.1

# Python code

```python
from sqlglot.lineage import lineage

sql_nok = """
insert /*+ HINT1*/ 
into	AML_NWU.DLT_TAXDAILY 
( 
	ORG_UNIT_CODE,
	RUN_DATE 
)
select
	nd.ORG_UNIT_CODE,
	nd.RUN_DATE
from
(select /*+ HINT2*/ 
	MAP_TAXDAILY.ORG_UNIT_CODE as ORG_UNIT_CODE,
	to_date('2023-12-19', 'YYYY-MM-DD') as RUN_DATE
from	AML_NWU.MAP_TAXDAILY MAP_TAXDAILY
where (1=1)
and RUN_DATE = to_date('2023-12-19', 'YYYY-MM-DD')
) nd
"""

sql_ok = """
insert /*+ HINT1*/ 
into	AML_NWU.DLT_TAXDAILY 
( 
	ORG_UNIT_CODE,
	RUN_DATE 
)
select
	MAP_TAXDAILY.ORG_UNIT_CODE as ORG_UNIT_CODE,
	to_date('2023-12-19', 'YYYY-MM-DD') as RUN_DATE
from	AML_NWU.MAP_TAXDAILY MAP_TAXDAILY
where (1=1)
and RUN_DATE = to_date('2023-12-19', 'YYYY-MM-DD')

"""

#sql = sql_nok
sql = sql_ok

column = 'RUN_DATE'
lin = lineage(column, sql, dialect='oracle')

for n in lin.walk():
  print(str(n.name) + ' - ' + str(n.expression))
```

georgesittas:
We used to create Scopes for both Query and DDL expressions, but the logic that handled the latter was brittle, incomplete and resulted in another expression type that users of the Scope module had to take into account, thus making the interface more complex to use.

This makes it so that we'll only optimize the _queries_ of statements such as CTAS and INSERT INTO, instead of the whole thing. Same holds for lineage inference - we'll only analyze the DDL queries.

Fixes #3086

### DIFF
--- a/sqlglot/optimizer/qualify.py
+++ b/sqlglot/optimizer/qualify.py
@@ -14,7 +14,6 @@ from sqlglot.optimizer.qualify_columns import (
 )
 from sqlglot.optimizer.qualify_tables import qualify_tables
 from sqlglot.schema import Schema, ensure_schema
-from sqlglot.transforms import extract_ddl_query
 
 
 def qualify(
@@ -67,11 +66,6 @@ def qualify(
         only the child will be qualified and that will be part of the final DDL expression.
     """
     schema = ensure_schema(schema, dialect=dialect)
-
-    original = expression
-    expression = extract_ddl_query(expression)
-    ddl = original is not expression
-
     expression = normalize_identifiers(expression, dialect=dialect)
     expression = qualify_tables(expression, db=db, catalog=catalog, schema=schema)
 
@@ -95,9 +89,5 @@ def qualify(
 
     if validate_qualify_columns:
         validate_qualify_columns_func(expression)
-
-    if ddl:
-        original.set("expression", expression)
-        expression = original
 
     return expression
--- a/sqlglot/optimizer/qualify.py
+++ b/sqlglot/optimizer/qualify.py
@@ -62,8 +62,7 @@ def qualify(
         identify: If True, quote all identifiers, else only necessary ones.
 
     Returns:
-        The qualified expression. If the original expression is a DDL with a `Query` child, then
-        only the child will be qualified and that will be part of the final DDL expression.
+        The qualified expression.
     """
     schema = ensure_schema(schema, dialect=dialect)
     expression = normalize_identifiers(expression, dialect=dialect)
--- a/sqlglot/optimizer/qualify.py
+++ b/sqlglot/optimizer/qualify.py
@@ -70,6 +70,8 @@ def qualify(
 
     original = expression
     expression = extract_ddl_query(expression)
+    ddl = original is not expression
+
     expression = normalize_identifiers(expression, dialect=dialect)
     expression = qualify_tables(expression, db=db, catalog=catalog, schema=schema)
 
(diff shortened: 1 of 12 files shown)


sqlglot/transforms.py
```diff
@@ -576,6 +574,30 @@ def struct_kv_to_alias(expression: exp.Expression) -> exp.Expression:
...
+        >>> ddl.transform(extract_ddl_query).sql()
+        'WITH cte AS (SELECT 1 AS c) SELECT c FROM cte'
+    """
+    query = expression.expression
+    if isinstance(expression, exp.DDL) and isinstance(query, exp.Query):
+        ddl_with = expression.args.get("with")
+        expression = query
+        if ddl_with:
+            for cte in ddl_with.pop().expressions:
+                query.with_(cte.alias, cte.this, copy=False)
```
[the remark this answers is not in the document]
georgesittas: Yeah I kinda skipped this case on purpose but thinking about this again you're right. If the query also has CTEs that reference the DDL ctes then this breaks them. Will fix, thanks for catching.

sqlglot/optimizer/qualify.py
```diff
@@ -89,4 +94,8 @@ def qualify(
     if validate_qualify_columns:
         validate_qualify_columns_func(expression)
 
+    if isinstance(original, exp.DDL) and isinstance(original.expression, exp.Query):
```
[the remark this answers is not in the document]
georgesittas: Done

sqlglot/transforms.py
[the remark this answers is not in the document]
georgesittas: Fixed

sqlglot/optimizer/qualify.py
```diff
@@ -62,9 +63,15 @@ def qualify(
...
 
     Returns:
-        The qualified expression.
+        The qualified expression. If the original expression is a DDL with a `Query` child, then
+        only the child will be qualified and that will be part of the final DDL expression.
     """
     schema = ensure_schema(schema, dialect=dialect)
+
+    original = expression
+    expression = extract_ddl_query(expression)
```
tobymao, line 72: why is this being called in qualify and optimize? then it happens twice

georgesittas: Because qualify is usually called outside of optimize (e.g. in sqlmesh), and same holds for qualify tables. My reasoning was that this call was low cost and is justifiable to repeat a couple or three times if necessary to ensure these basic rules can work with ddls as well.

tobymao: ok i guess it's low cost

tobymao: @georgesittas if that's the case, we don't need it in optimize, because it happens in qualify right

tobymao (approved):

Issue #3093, opened 2024-03-07
TO_NUMBER(number)  is parsed wrongly from oracle to doris


yx-keith:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`
- Check if the issue still exists on main

**Fully reproducible code snippet**
Please include a fully reproducible code snippet or the input sql, dialect, and expected output.
```
import sqlglot
sql="""
SELECT to_number(x) from tmp;
"""
sqlglot.transpile(sql,read="oracle", write="doris")[0]
```
output is below:
`SELECT to_number(x) from tmp;`
oracle has to_number function,but doris don't have it.I think it should be converted to cast(x as double)

**Official Documentation**
Please include links to official SQL documentation related to your issue.

georgesittas:
This is a bit more nuanced than just simply converting `TO_NUMBER` into a cast at parse time, because there are more arguments involved and you need to preserve them. It's not high priority for us so it's out of scope right now, but feel free to make a PR if you need it. I'd suggest looking at how `TO_CHAR` is handled first.

PR #3100, opened 2024-03-08
Feat(doris):add support for to_number(x)

codeDing18:
fix [#3093](https://github.com/tobymao/sqlglot/issues/3093)

### DIFF
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -11,10 +11,6 @@ from sqlglot.dialects.dialect import (
 from sqlglot.dialects.mysql import MySQL
 
 
-def _tonumber_sql(self, expression: exp.ToNumber) -> str:
-    if expression.args.get("format"):
-        self.unsupported("Format argument unsupported for TO_NUMBER function")
-    return self.sql(exp.cast(expression.this, "double"))
 
 
 class Doris(MySQL):
@@ -43,6 +39,7 @@ class Doris(MySQL):
         }
 
         LAST_DAY_SUPPORTS_DATE_PART = False
+        SUPPORTS_TO_NUMBER = False
 
         TIMESTAMP_FUNC_TYPES = set()
 
@@ -72,7 +69,6 @@ class Doris(MySQL):
             exp.TimestampTrunc: lambda self, e: self.func(
                 "DATE_TRUNC", e.this, "'" + e.text("unit") + "'"
             ),
-            exp.ToNumber: _tonumber_sql,
             exp.UnixToStr: lambda self, e: self.func(
                 "FROM_UNIXTIME", e.this, time_format("doris")(self, e)
             ),
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -11,8 +11,6 @@ from sqlglot.dialects.dialect import (
 from sqlglot.dialects.mysql import MySQL
 
 
-
-
 class Doris(MySQL):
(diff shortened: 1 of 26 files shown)


sqlglot/expressions.py
```diff
@@ -4462,6 +4462,10 @@ class ToChar(Func):
     arg_types = {"this": True, "format": False, "nlsparam": False}
 
 
+class ToNumber(Func):
+    arg_types = {"this": True, "format": False, "nlsparam": False}
+
```
georgesittas, line 4467: When defining `Func`s, you need to take into account what other dialects do. This definition implicitly makes it so that `TO_NUMBER` is a function that expects at most three arguments for _every_ dialect. This is not necessarily the case though, for example Snowflake also [supports](https://docs.snowflake.com/en/sql-reference/functions/to_decimal) `to_number(expr, fmt, precision, scale)`, so this PR would break queries that use this function with four arguments. So you need to take this into account and make sure `ToNumber` can represent Snowflake's variant as well, or fall back to parsing it as `Anonymous` if there are >2 arguments in Snowflake.

Additionally right now the base generator converts `ToNumber` into a cast. This will also break valid queries that use `TO_NUMBER` with a format in snowflake, postgres, and various other dialects that support this function. This needs to also be fixed if you're gonna introduce this expression.

With a quick look, these are the dialects that support `TO_NUMBER` and the respective supported arguments:

```
to_number(expr, fmt)
databricks, drill, postgres, redshift, spark

to_number(expr, fmt, nlsparam)
oracle, teradata 

to_number(expr, fmt, precision, scale)
snowflake
```

Please refactor this PR so that it handles these cases & make sure that it's properly tested. At the very minimum we need to ensure that a query containing `TO_NUMBER` in these dialects can be roundtripped (e.g. `parse_one` and `generate` with read=write should not change `TO_NUMBER`).

codeDing18: I have made modifications, please review it

sqlglot/dialects/doris.py
```diff
@@ -11,6 +11,12 @@
 from sqlglot.dialects.mysql import MySQL
 
 
+def _tonumber_sql(self, expression: exp.ToNumber) -> str:
+    if expression.args.get("format"):
+        self.unsupported("Format argument unsupported for TO_NUMBER function")
+    return self.sql(exp.cast(expression.this, "double"))
```
georgesittas, line 17: Let's move this function to the base Generator so that other dialects can also leverage this logic.

We should have a class-level constant in the generator called `SUPPORTS_TO_NUMBER`, which defaults to `True` and is set to `False` for the dialects that don't support it - see my previous comment about which dialects support it. If it's set, we need to use `fallback_function_sql` so that `TO_NUMBER` will be correctly generated, otherwise we should use this logic.

As a final note, let's change the warning to "Unsupported TO_NUMBER function, converting to a CAST as DOUBLE".

georgesittas (changes requested):
@codeDing18 left another comment. Please test this more extensively, it affects several dialects so we need to ensure it doesn't break.

sqlglot/dialects/doris.py
codeDing18: I have modified it and add test for all dialects, please review

codeDing18:
> @codeDing18 left another comment. Please test this more extensively, it affects several dialects so we need to ensure it's not breaking.

@georgesittas  I have modified it and add test for all dialects, please review

codeDing18:
@georgesittas   Thanks for the correction, I will learn from it

PR #3102, opened 2024-03-08
feat(duckdb): add support for exp.ArrayJoin

seruman:
Noticed this while transpiling a Snowflake query in to DuckDB;
```python
query = "SELECT ARRAY_TO_STRING(id, ',') AS ids FROM some_table;"
e = sqlglot.parse_one(query, read="snowflake")
dsql = e.sql(dialect="duckdb", pretty=True)

print(dsql)

# SELECT
#   ARRAY_JOIN(id, ',') AS ids
# FROM some_table
```

### DIFF
--- a/tests/dialects/test_duckdb.py
+++ b/tests/dialects/test_duckdb.py
@@ -666,6 +666,13 @@ class TestDuckDB(Validator):
             """SELECT i FROM GENERATE_SERIES(0, 12) AS _(i) ORDER BY i ASC""",
         )
 
+        self.validate_all(
+            "SELECT ARRAY_TO_STRING(fruits, ',') AS all_fruits FROM t",
+            read={
+                "snowflake": "SELECT ARRAY_TO_STRING(fruits, ',') AS all_fruits FROM t",
+            },
+        )
+
     def test_array_index(self):
         with self.assertLogs(helper_logger) as cm:
             self.validate_all(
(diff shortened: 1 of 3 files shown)


tests/dialects/test_duckdb.py
```diff
@@ -666,6 +666,13 @@ def test_duckdb(self):
             """SELECT i FROM GENERATE_SERIES(0, 12) AS _(i) ORDER BY i ASC""",
         )
 
+        self.validate_all(
```
tobymao, line 669: you can delete this test

tests/dialects/test_snowflake.py
```diff
@@ -678,6 +678,7 @@ def test_snowflake(self):
             write={
                 "spark": "ARRAY_JOIN(x, '')",
                 "snowflake": "ARRAY_TO_STRING(x, '')",
+                "duckdb": "ARRAY_TO_STRING(x, '')",
```
tobymao, line 681: add a test here read={
  "duckdb": "ARRAY_TO_STRING(x, '')"

tests/dialects/test_duckdb.py
seruman: done

tests/dialects/test_snowflake.py
seruman: done

tobymao:
thanks!

PR #3103, opened 2024-03-08
feat: no recursion copy

tobymao:


sqlglot/expressions.py
```diff
@@ -248,17 +248,37 @@ def meta(self) -> t.Dict[str, t.Any]:
...
+
+        while stack:
+            node, copy = stack.pop()
+
+            if node.comments is not None:
+                copy.comments = deepcopy(node.comments)
+            if node._type is not None:
+                copy._type = node._type.copy()
+            if node._meta is not None:
+                copy._meta = deepcopy(node._meta)
```
georgesittas, line 263: should we copy `self._hash` too?

georgesittas (approved):
Looks good, nice

sqlglot/expressions.py
```diff
@@ -248,17 +248,37 @@ def meta(self) -> t.Dict[str, t.Any]:
...
+
+            for k, v in node.args.items():
+                if isinstance(v, Expression):
+                    c = v.__class__()
+                    copy.set(k, c)
+                    stack.append((v, c))
+                elif isinstance(v, list):
+                    for e in v:
```
georgesittas, line 271: FYI the failing test has to do with this branch - the edge case is `[]`. We need to copy that too.

PR #3105, opened 2024-03-08
feat: no recursion dfs

tobymao:


sqlglot/expressions.py
```diff
@@ -349,11 +349,11 @@ def depth(self) -> int:
             return self.parent.depth + 1
         return 0
 
-    def iter_expressions(self) -> t.Iterator[t.Tuple[str, Expression]]:
+    def iter_expressions(self, reverse: bool = False) -> t.Iterator[t.Tuple[str, Expression]]:
         """Yields the key and expression for all arguments, exploding list args."""
-        for k, vs in self.args.items():
+        for k, vs in reversed(self.args.items() if reverse else self.args.items()):
```
georgesittas, line 354: There was probably a typo here, I assume you meant the following?

```python
reversed(self.args.items()) if reverse else self.args.items()
```

tobymao: because dfs is done in reverse order

PR #3106, opened 2024-03-08
feat(lineage): fall back to infering lineage using source tables

rexledesma:
## Summary

When taking a look at lineage against `jaffle_shop`, I was surprised that we couldn't infer the lineage of the [`customer_lifetime_value`](https://github.com/dbt-labs/jaffle_shop/blob/main/models/customers.sql#L57) column to its tables present in the standalone query.

```python
lineage(
    column="customer_lifetime_value",
    sql=parse_one(customers_sql).expression,
).to_html()
```

Screenshot 2024-03-08 at 4 43 22 PM

Instead of displaying `?`, I wonder if we can instead display an inference that this column inherits from its source tables (`stg_payments`, `stg_orders`). We fall back to this inference if the source columns don't give us sufficient table information because of our lack of qualification.

Screenshot 2024-03-08 at 5 35 18 PM

Just need to remove duplicate leaves from the rendered graph and add tests.

## Test Plan
TODO: fix existing tests, add test case that understands column lineage when selecting from tables without full schema information

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -231,27 +231,50 @@ def to_node(
 
     for c in source_columns:
         table = c.table
-        source = scope.sources.get(table)
+        table_source = scope.sources.get(table)
 
-        if isinstance(source, Scope):
+        if isinstance(table_source, Scope):
             selected_node, _ = scope.selected_sources.get(table, (None, None))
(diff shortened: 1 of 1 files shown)


tobymao:
i don’t think i’m a big fan of this approach. imo the schema should just be passed in.

PR #3112, opened 2024-03-09
Fix!: traverse union scopes iteratively

georgesittas:


sqlglot/optimizer/scope.py
```diff
@@ -548,30 +549,39 @@ def _traverse_select(scope):
...
-            scope_type=ScopeType.UNION,
-        )
-    ):
-        yield right
+    while expression_stack:
+        expression = expression_stack.pop()
 
-    scope.union_scopes = [left, right]
+        if isinstance(expression, exp.Union):
+            union_scope = Scope(expression, parent=seq_get(union_scope_stack, -1))
```
tobymao, line 560: how come you're not using branch?

georgesittas: I changed this locally, using branch now. This was probably the reason why the tests failed.

georgesittas:
yep it works

georgesittas:
Hmm, the traversal though is still problematic...

```python
>>> from sqlglot import exp, parse_one
>>> from sqlglot.optimizer.scope import build_scope
>>>
>>> large_union = parse_one(" UNION ALL ".join(["SELECT x FROM t"] * 1000))
>>> len(list(large_union.find_all(exp.Select)))
1000
>>>
>>> scope = build_scope(large_union) # ok
>>> for s in scope.traverse():
...     print(s)
...
    for child_scope in itertools.chain(
RecursionError: maximum recursion depth exceeded while calling a Python object
```

sqlglot/optimizer/scope.py
```diff
@@ -438,7 +438,22 @@ def traverse(self):
...
+                scopes = []
+                while stack:
+                    is_right_tree, scope = stack.pop()
+                    if isinstance(scope.expression, exp.Union):
+                        left, right = scope.union_scopes
+                        stack.extend([(True, right), (False, left)])
+                        scopes.append(scope)
+                    else:
+                        yield from scope.traverse()
+                        if is_right_tree:
```
tobymao, line 452: why do you only do this for the right tree?

sqlglot/optimizer/scope.py
```diff
@@ -522,7 +537,9 @@ def _traverse_scope(scope):
     if isinstance(scope.expression, exp.Select):
         yield from _traverse_select(scope)
     elif isinstance(scope.expression, exp.Union):
+        yield from _traverse_ctes(scope)
```
tobymao, line 540: is this needed if we traverse_ctes inside of traverse_union?

sqlglot/optimizer/scope.py
tobymao: should you just add scope to the stack? so it goes right, scope, left ?

sqlglot/optimizer/scope.py
```diff
@@ -548,30 +565,38 @@ def _traverse_select(scope):
 
 
 def _traverse_union(scope):
-    yield from _traverse_ctes(scope)
+    prev_scope = None
```
tobymao, line 568: how come you need this prev_scope concept?

sqlglot/optimizer/scope.py
georgesittas: We need to do yield the left scope tree, then the right, and then the scope. Scope is already in the stack, and I can't just yield from it directly because then the order will be messed up. This check ensures that both subtrees have been visited already. What did you have in mind?

sqlglot/optimizer/scope.py
georgesittas: Yeah, had to pivot a couple of times to get this write. This handles the ctes of the top-level union scope.

sqlglot/optimizer/scope.py
georgesittas: I ended up doing it like this because if e.g. you have UNION(UNION(a, b), c) then you first visit a, b so you can construct the inner union scope, and then you visit c and can use the UNION(a, b) and c to construct the outer union scope. So basically, prev is used to keep track of the "last constructed" left scope tree. I also tried a couple approaches here but didn't turn out as well, we can chat more about this.

sqlglot/optimizer/scope.py
tobymao: i was thinking you could do

stack.extend((right, scope, left)) and then just do the right thing accordingly, rather than store is_right_scope

georgesittas: If you add the scope to the stack again won't it keep entering that branch where you extend the stack, resulting in an infinite loop? Maybe if we added another "visited" bool in the stack we could avoid re-entering it, but I guess that defeats the purpose of this simplification, right?

sqlglot/optimizer/scope.py
```diff
@@ -438,7 +438,22 @@ def traverse(self):
         for child_scope in itertools.chain(
             self.cte_scopes, self.union_scopes, self.table_scopes, self.subquery_scopes
         ):
-            yield from child_scope.traverse()
+            if child_scope.is_union:
+                stack = [(False, child_scope)]
```
barakalon, line 442: We should use `deque`

sqlglot/optimizer/scope.py
```diff
@@ -438,7 +438,22 @@ def traverse(self):
         for child_scope in itertools.chain(
```
barakalon, line 438: Why dont we just do the whole traversal without recursion?

sqlglot/optimizer/scope.py
tobymao: don’t you only need deque if you’re using a queue? this is just a stack

barakalon: Ah, yes, whoops

tobymao (approved):

PR #3128, opened 2024-03-12
feat: Adding NAME data type in Postgres/Redshift

### LINKED ISSUE #3123: Redshift name type
Reopening https://github.com/tobymao/sqlglot/issues/3107

I filed the bug against the postgres dialect, but was actually using Redshift myself. It seems this bug only appears for Redshift, despite the redshift dialect being based on postgres.

```
from sqlglot import parse_one

parse_one("""
SELECT pgc.relname::CHARACTER varying AS textin
  FROM pg_class pgc
 WHERE pgc.relname <> 'my_table'::name
 """, read="redshift")
```

VaggelisD:
Fixes #3123 

Docs
--------
- https://www.postgresql.org/docs/current/datatype-character.html
- https://dba.stackexchange.com/questions/217533/what-is-the-data-type-name-in-postgresql

### DIFF
--- a/tests/dialects/test_postgres.py
+++ b/tests/dialects/test_postgres.py
@@ -608,7 +608,13 @@ class TestPostgres(Validator):
                 "postgres": "x / y ^ z",
             },
         )
-
+        self.validate_all(
+            "SELECT pgc.relname::CHARACTER VARYING AS textin FROM pg_class pgc WHERE pgc.relname <> 'my_table'::name",
+            write={
+                "postgres": "SELECT CAST(pgc.relname AS VARCHAR) AS textin FROM pg_class AS pgc WHERE pgc.relname <> CAST('my_table' AS NAME)",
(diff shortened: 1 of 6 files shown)


tests/dialects/test_postgres.py
```diff
@@ -608,7 +608,13 @@ def test_postgres(self):
                 "postgres": "x / y ^ z",
             },
         )
-
+        self.validate_all(
+            "SELECT pgc.relname::CHARACTER VARYING AS textin FROM pg_class pgc WHERE pgc.relname <> 'my_table'::name",
```
georgesittas, line 612: We don't really care about the `CHARACTER VARYING` bit in this test, we test this elsewhere. I think it's worth changing that with `VARCHAR` so you can also add the redshift -> postgres direction here using the `read` kwarg.

georgesittas (approved):

PR #3129, opened 2024-03-12
fix(athena): Fix CREATE TABLE properties, STRING data type

### LINKED ISSUE #3126: Athena Iceberg Tables parsing issue
Hi,
I want to parse a SQL Statement that creates an Iceberg table on Athena:

```sql
create table if not exists tmp.mytable (
    name string
)
location 's3://bucket/tmp/mytable/'
tblproperties (
  'table_type'='iceberg',
  'format'='parquet'
);
```
running 
```python
stmts = sqlglot.parse(sql, read=sqlglot.Dialects.ATHENA)
stmts[0].sql()
```
returns:
```sql
CREATE TABLE IF NOT EXISTS tmp.mytable 
    (name TEXT) 
LOCATION 's3://bucket/tmp/mytable/' 
WITH (
    table_type='iceberg', 
    FORMAT='parquet'
)
```

Unfortunately, the syntax in Athena is different for Iceberg Tables and Hive-style tables.

The parsed statement should look like this:

```sql
CREATE TABLE IF NOT EXISTS tmp.mytable 
    (name STRING) 
LOCATION 's3://bucket/tmp/mytable/' 
TBLPROPERTIES (
    'table_type'='iceberg', 
    'FORMAT'='parquet'
)
```

Instead of WITH -> TBLPROPERTIES
The keys in the this block are wrapped in upper quotes and iceberg has slightly different data types. In this case STRING instead of TEXT

https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-supported-data-types.html
https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html

VaggelisD:
Fixes #3126

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from sqlglot import exp
 from sqlglot.dialects.trino import Trino
 from sqlglot.tokens import TokenType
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/athena.py
```diff
@@ -10,3 +11,27 @@ class Parser(Trino.Parser):
...
+
+        TYPE_MAPPING = {
+            **Trino.Generator.TYPE_MAPPING,
+            exp.DataType.Type.TEXT: "STRING",
+        }
+
+        TRANSFORMS = {
+            **Trino.Generator.TRANSFORMS,
+            exp.FileFormatProperty: lambda self, e: f"'FORMAT'={self.sql(e, 'this')}",
+        }
```
georgesittas, line 29: Do we need this? Isn't it covered by the `property_sql` override?

georgesittas (approved):
Looks good - just one comment

PR #3134, opened 2024-03-13
Fix: don't mutate parent nested classes if undefined in a dialect

georgesittas:
There was a bug where if a class didn't override the nested classes (`Tokenizer`, `Parser`, `Generator`), we could end up overwriting attributes of the nested classes of its parent class. For example, if we got rid of PRQL's empty `Generator` override, we would end up with various broken tests because the metaclass code overwrote `Dialect.generator_class`.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -112,9 +112,14 @@ class _Dialect(type):
 
         klass.INVERSE_ESCAPE_SEQUENCES = {v: k for k, v in klass.ESCAPE_SEQUENCES.items()}
 
-        klass.tokenizer_class = getattr(klass, "Tokenizer", Tokenizer)
-        klass.parser_class = getattr(klass, "Parser", Parser)
-        klass.generator_class = getattr(klass, "Generator", Generator)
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -112,9 +112,18 @@ def __new__(cls, clsname, bases, attrs):
...
+        base_parser = (getattr(base, "parser_class", Parser),)
+        base_generator = (getattr(base, "generator_class", Generator),)
+
+        klass.tokenizer_class = klass.__dict__.get(
+            "Tokenizer", type("Tokenizer", base_tokenizer, {})
+        )
+        klass.parser_class = klass.__dict__.get("Parser", type("Parser", base_parser, {}))
+        klass.generator_class = klass.__dict__.get(
+            "Generator", type("Generator", base_generator, {})
+        )
```
georgesittas, line 126: Note the change to `__dict__` over `getattr` here - the motivation was to only check _this_ class' attributes. E.g. we don't want to share the same `Parser` attribute between `Presto` and `Trino` (the latter doesn't override it).

PR #3137, opened 2024-03-13
fix(duckdb): Slice + Array bug

### LINKED ISSUE #3136: valid duckdb query fails to parse

This query cannot be parsed by sqlglot, but it is a valid duckdb query:

```
In [18]: import sqlglot as sg, sqlglot.expressions as sge

In [19]: import duckdb

In [20]: sql = """
    ...: SELECT
    ...:   LIST_APPLY(
    ...:     RANGE(
    ...:       1,
    ...:       GREATEST(
    ...:         LENGTH(LIST_VALUE(CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT))),
    ...:         LENGTH(LIST_VALUE(CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT)))
    ...:       ) + 1
    ...:     ),
    ...:     i -> {'f1': LIST_VALUE(CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT))[i], 'f2': LIST_VALUE(CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT))[i]}
    ...:   ) AS "ArrayZip()"
    ...: """

In [21]: duckdb.execute(sql).df()
Out[21]:
                                          ArrayZip()
0  [{'f1': 1, 'f2': 1}, {'f1': 2, 'f2': 2}, {'f1'...

In [22]: sg.parse_one(sql, read="duckdb")
...
ParseError: Expected }. Line 12, Col: 2.
  AS TINYINT))[i], 'f2': LIST_VALUE(CAST(1 AS TINYINT), CAST(2 AS TINYINT), CAST(3 AS TINYINT))[i]}
  ) AS "ArrayZip()"
```

VaggelisD:
Fixes #3136 

Caught 3 consecutive bugs:

- When matching `R_BRACKET` or `R_BRACE` inside `parse_bracket`, the short-circuit should be flipped, otherwise a single `parse_bracket` call can match 2 closing tokens, e.g. in `{'x': arr[i]}`, the parsing of `arr[i]` will also match the following `}` 

- When transforming `exp.Slice` to `exp.PropertyEq`, the `exp.Slice` name was used instead of `exp.Slice.this` name

- DuckDB requires wrapped arrays if the indexing is in place, i.e `SELECT [1, 2][1]` does not work, but `SELECT ([1, 2])[1]` does. The function `LIST_VALUE()` is parsed and generated as an `exp.Array` so it's also affected by this behavior; thus, the DuckDB `exp.Bracket` generation was enhanced to wrap arrays.

### DIFF
--- a/tests/dialects/test_duckdb.py
+++ b/tests/dialects/test_duckdb.py
@@ -701,6 +701,18 @@ class TestDuckDB(Validator):
             self.validate_identity(
                 "[x.STRING_SPLIT(' ')[1] FOR x IN ['1', '2', 3] IF x.CONTAINS('1')]"
             )
+            self.validate_identity(
+                """SELECT LIST_VALUE(1)[1]""",
+                """SELECT ([1])[1]""",
+            )
+            self.validate_identity(
+                """{'x': LIST_VALUE(1)[1]}""",
+                """{'x': ([1])[1]}""",
+            )
+            self.validate_identity(
+                """SELECT LIST_APPLY(RANGE(1, 4), i -> {'f1': LIST_VALUE(1, 2, 3)[i], 'f2': LIST_VALUE(1, 2, 3)[i]})""",
+                """SELECT LIST_APPLY(RANGE(1, 4), i -> {'f1': ([1, 2, 3])[i], 'f2': ([1, 2, 3])[i]})""",
+            )
 
             self.assertEqual(
                 cm.output,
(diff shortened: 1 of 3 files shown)


tests/dialects/test_duckdb.py
```diff
@@ -711,6 +723,10 @@ def test_array_index(self):
                     "WARNING:sqlglot:Applying array index offset (1)",
                     "WARNING:sqlglot:Applying array index offset (-1)",
                     "WARNING:sqlglot:Applying array index offset (1)",
+                    "WARNING:sqlglot:Applying array index offset (-1)",
+                    "WARNING:sqlglot:Applying array index offset (1)",
+                    "WARNING:sqlglot:Applying array index offset (-1)",
+                    "WARNING:sqlglot:Applying array index offset (1)",
```
georgesittas, line 729: I think this assertion is confusing, the sequence by itself doesn't explain much besides the fact that there are a bunch of warnings logged. I'd rather either 1) get rid of it altogether (so won't need the `as cm` part above), or 2) reduce the scope of the assertion to only a single test so that we observe the dec/increment of the index once.

georgesittas (approved):
LGTM

tests/dialects/test_duckdb.py
VaggelisD: By changing the array indexes to be an identifier instead of an integer literal we can get rid of these 4 warnings and maintain the tests, would that work?

georgesittas: I'd just do (2) instead of augmenting this list every time we add an we test tbh

PR #3139, opened 2024-03-14
Refactor!: desugar LOG2 and LOG10 by converting them into LOG

### LINKED ISSUE #3138: Sqlglot does not correctly translate log2 and log10 functions from Trino to Snowflake
`LOG10` and `LOG2` functions in "trino" dialect are not correctly converted to LOG function in "snowflake" dialect

Way to reproduce:
```
>>> sqlglot.parse_one("select log10(10)", dialect="trino").sql(dialect="snowflake")
'SELECT LOG10(10)'
```

**Official Documentation**
Trino documentation: https://trino.io/docs/current/functions/math.html
Snowflake documentation: https://docs.snowflake.com/en/sql-reference/functions/log

georgesittas:
Fixes #3138


georgesittas:
Hmm let me double check actually if this is valid, maybe not all of the dialects support the generic `LOG` function.

tobymao (approved):
surprised we had this

PR #3141, opened 2024-03-14
feat!!: make transform non recursive and cleanup walk() to only yield…

tobymao:
… a node


sqlglot/expressions.py
```diff
@@ -91,12 +91,13 @@ class Expression(metaclass=_Expression):
...
     key = "expression"
     arg_types = {"this": True}
-    __slots__ = ("args", "parent", "arg_key", "comments", "_type", "_meta", "_hash")
+    __slots__ = ("args", "parent", "arg_key", "index", "comments", "_type", "_meta", "_hash")
 
     def __init__(self, **args: t.Any):
         self.args: t.Dict[str, t.Any] = args
         self.parent: t.Optional[Expression] = None
         self.arg_key: t.Optional[str] = None
+        self.index: t.Optional[int] = None
```
georgesittas, line 100: Let's document this in the class' docstring

sqlglot/expressions.py
```diff
@@ -253,6 +254,8 @@ def __deepcopy__(self, memo):
 
         while stack:
             node, copy = stack.pop()
+            copy.arg_key = node.arg_key
```
georgesittas, line 257: This may be redundant, I believe we set `arg_key` in both `set` and `append` (check `_set_parent`)

sqlglot/expressions.py
```diff
@@ -427,7 +434,7 @@ def root(self) -> Expression:
             expression = expression.parent
         return expression
 
-    def walk(self, bfs=True, prune=None):
+    def walk(self, bfs: bool=True, prune=None) -> t.Iterator[Expression]:
```
georgesittas, line 437:
```suggestion
    def walk(self, bfs: bool = True, prune: t.Optional[t.Callable] = None) -> t.Iterator[Expression]:
```

sqlglot/expressions.py
```diff
@@ -445,46 +452,47 @@ def walk(self, bfs=True, prune=None):
         else:
             yield from self.dfs(prune=prune)
 
-    def dfs(self, parent=None, key=None, prune=None):
+    def dfs(self, prune=None) -> t.Iterator[Expression]:
```
georgesittas, line 455:
```suggestion
    def dfs(self, prune: t.Optional[t.Callable] = None) -> t.Iterator[Expression]:
```

sqlglot/expressions.py
georgesittas: Can make the callable more specific, i.e. it expects an expression and returns a bool

sqlglot/expressions.py
```diff
@@ -445,46 +452,47 @@ def walk(self, bfs=True, prune=None):
...
+            if prune and prune(node):
                 continue
 
-            for k, v in node.iter_expressions(reverse=True):
-                stack.append((v, node, k))
+            for v in node.iter_expressions(reverse=True):
+                stack.append(v)
 
-    def bfs(self, prune=None):
+    def bfs(self, prune=None) -> t.Iterator[Expression]:
```
georgesittas, line 476: ditto about type hint

sqlglot/expressions.py
```diff
@@ -547,7 +555,7 @@ def sql(self, dialect: DialectType = None, **opts) -> str:
 
         return Dialect.get_or_raise(dialect).generate(self, **opts)
 
-    def transform(self, fun, *args, copy=True, **kwargs):
+    def transform(self, fun: t.Callable, *args: t.Any, copy: bool = True, **kwargs) -> Expression:
```
georgesittas, line 558: ditto about more specific callable

tobymao: this can't be done, because it accepts kwargs

georgesittas (commented):
Looks good

sqlglot/expressions.py
```diff
@@ -564,17 +579,20 @@ def transform(self, fun, *args, copy=True, **kwargs):
...
-        replace_children(new_node, lambda child: child.transform(fun, *args, copy=False, **kwargs))
-        return new_node
+            if root:
+                if new_node is not node:
+                    node.replace(new_node)
+            else:
+                root = new_node
+
+        assert root
+        return root.assert_is(Expression)
```
georgesittas, line 595: Curious why these `assert_is` calls were added, is it for mypy?

sqlglot/expressions.py
```diff
@@ -6934,12 +6986,32 @@ def replace_children(expression: Expression, fun: t.Callable, *args, **kwargs) -
...
-        expression.args[k] = new_child_nodes if is_list_arg else seq_get(new_child_nodes, 0)
+        expression.set(k, new_child_nodes if is_list_arg else seq_get(new_child_nodes, 0))
+
+
+def replace_tree(expression: Expression, fun: t.Callable) -> Expression:
+    """
+    Replace an entire tree with the result of function calls on each node.
+
+    This will be traversed in reverse dfs, so leaves first.
+    If new nodes are created as a result of function calls, those will always be evaluated.
```
georgesittas, line 7000: Nice!

I caught myself thinking what "evaluated" means in this context, btw. May be worth rephrasing to clarify.

tests/test_expressions.py
```diff
@@ -520,7 +520,6 @@ def remove_column_b(node):
             return node
 
         self.assertEqual(expression.transform(remove_column_b).sql(), "SELECT a FROM x")
-        self.assertEqual(expression.transform(lambda _: None), None)
```
georgesittas, line 523: Is this not true anymore?

georgesittas (approved):

sqlglot/expressions.py
tobymao: yes

tests/test_expressions.py
tobymao: right because we don't allow transform to return None

PR #3145, opened 2024-03-14
Feat(tsql): transpile LIMIT with OFFSET properly

### LINKED ISSUE #3144: Concerns about Translating SQLite Queries to TSQL
1. **Offset Issue:**
   
   When attempting to convert an SQLite query to TSQL using the provided code snippet:

   ```python
   import sqlglot

   sql_a = "SELECT name, (SELECT COUNT(*) FROM orders WHERE orders.customer_id = customers.id) AS order_count FROM customers LIMIT 5, 10"

   sql_b = sqlglot.transpile(sql_a, read="sqlite", write="tsql")[0]

   print(sql_b)
   ```

   The resulting TSQL output is as follows:

   ```sql
   SELECT TOP 10 name, (SELECT COUNT(*) FROM orders WHERE orders.customer_id = customers.id) AS order_count FROM customers
   ```

   It appears that the translated TSQL does not include information about the offset of 5. Instead, it only selects the top 10 rows, neglecting the fact that it should select 10 rows after an offset of 5. Could you please verify if this behavior is intended?

2. **Completeness of Translation:**

   I had a question about the thoroughness of the translation process from SQLite to TSQL. Does the transpiler cover all possible cases of SQLite queries that are incompatible with TSQL syntax? Are there any known clauses or scenarios where the transpiler fails to convert the syntax accurately?

georgesittas:
Fixes #3144

Reference: https://www.microsoftpressstore.com/articles/article.aspx?p=2314819

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -328,6 +328,21 @@ def _json_extract_sql(
     return self.func("ISNULL", json_query, json_value)
 
 
+def _replace_limit_with_offset_fetch(expression: exp.Expression) -> exp.Expression:
+    limit = expression.args.get("limit")
+    if (
+        isinstance(expression, exp.Select)
+        and isinstance(limit, exp.Limit)
+        and expression.args.get("offset")
+    ):
+        limit.replace(exp.Fetch(direction="NEXT", count=limit.expression))
+
+        if not expression.args.get("order"):
+            expression.order_by(exp.select(exp.null()).subquery(), copy=False)
+
+    return expression
+
+
 class TSQL(Dialect):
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -328,6 +328,21 @@ def _json_extract_sql(
     return self.func("ISNULL", json_query, json_value)
 
 
+def _replace_limit_with_offset_fetch(expression: exp.Expression) -> exp.Expression:
```
tobymao, line 331: doesn't this happen in the generator? why do we need this?

georgesittas: For context, `TOP` can't be used with `OFFSET`, so we need to produce the `FETCH` clause if we need to offset the rowset.

PR #3149, opened 2024-03-15
Feat: add skip limit token kwarg

z3z1ma:
The way `_parse_limit` is implemented, adding a `skip_limit_token` kwarg forces us to assume 1 of 2 branches (limit/fetch). The direction is, of course, to use the limit code path because our goal is ultimately to enable an optional `@LIMIT` macro in SQLMesh. We achieve this via simple short circuit.


tobymao:
run make style

PR #3150, opened 2024-03-15
feat(mysql): Support for multi arg GROUP_CONCAT

### LINKED ISSUE #3142: feat: support for mysql group_concat with multi columns

```python 
import sqlglot
sql = "select a, group_concat(b, ' ', c SEPARATOR ',') from table group by a;"
sqlglot.parse_one(sql, read="mysql")
# or 
sqlglot.transpile(sql, read="mysql", write="postgres", pretty=True)

```

```
sqlglot/parser.py:1255, in Parser.raise_error(self, message, token)
   1243 error = ParseError.new(
   1244     f"{message}. Line {token.line}, Col: {token.col}.\n"
   1245     f"  {start_context}\033[4m{highlight}\033[0m{end_context}",
   (...)
   1251     end_context=end_context,
   1252 )
   1254 if self.error_level == ErrorLevel.IMMEDIATE:
-> 1255     raise error
   1257 self.errors.append(error)

ParseError: Expecting ). Line 1, Col: 38.
```

transpile the sql to postgresql
output:
select a STRING_AGG(b || ' ' || c, ',')  from table group  a;

**Additional context**

VaggelisD:
Fixes #3142 

Change MySQL parsing for `exp.GroupConcat` to reduce/normalize the list of expressions into an `exp.Concat` node, thus making it easily transpilable to other dialects which only accept a single expression.   

Docs
------------
- https://dev.mysql.com/doc/refman/8.3/en/aggregate-functions.html#function_group-concat

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -615,37 +615,25 @@ class MySQL(Dialect):
 
         def _parse_group_concat(self) -> t.Optional[exp.Expression]:
             def concat_exprs(node, exprs):
-                if node and isinstance(node, exp.Distinct):
-                    exprs = node.expressions
-                    concat_exprs = (
-                        [self.expression(exp.Concat, expressions=exprs)]
-                        if len(exprs) > 1
-                        else exprs
-                    )
-                    return self.expression(exp.Distinct, expressions=concat_exprs)
+                if node and isinstance(node, exp.Distinct) and len(node.expressions) > 1:
+                    concat_exprs = [self.expression(exp.Concat, expressions=node.expressions)]
+                    node.set("expressions", concat_exprs)
+                    return node
                 return (
                     exprs[0] if len(exprs) == 1 else self.expression(exp.Concat, expressions=args)
                 )
 
             args = self._parse_csv(self._parse_lambda)
-            if len(args) > 1 and isinstance(args[-1], exp.Order):
-                # Case #1 (Only Order By): GROUP_CONCAT(a, b, c ORDER BY d ...)
-                # Order By has consumed only 'c', undo it and put all concatenated exprs in it's place
-                order = args[-1]
+            order = args[-1] if isinstance(args[-1], exp.Order) else None
+
+            if order:
+                # Order By is the last (or only) expression in the list and has consumed the 'expr' before it,
+                # remove 'expr' from exp.Order and add it back to args
                 args.pop()
                 args.append(order.this)
-                order.set("this", concat_exprs(None, args))
-                this = order
-            elif len(args) == 1 and isinstance(args[0], exp.Order):
-                # Case #2 (Distinct + Order By): GROUP_CONCAT([DISTINCT] a, b, c ORDER BY d ...)
-                order = args[0]
-                if isinstance(order.this, exp.Distinct):
-                    order.set("this", concat_exprs(order.this, order.this.expressions))
-                this = order
-            else:
-                # Case #3 (Only Distinct): GROUP_CONCAT([DISTINCT] a, b, c ...)
-                # Case #4 (Only exprs): GROUP_CONCAT(a, b, c, ...)
-                this = concat_exprs(args[0], args)
+                order.set("this", concat_exprs(order.this, args))
+
+            this = order or concat_exprs(args[0], args)
 
             separator = self._parse_field() if self._match(TokenType.SEPARATOR) else None
 
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -319,11 +319,7 @@ class MySQL(Dialect):
         FUNCTION_PARSERS = {
             **parser.Parser.FUNCTION_PARSERS,
             "CHAR": lambda self: self._parse_chr(),
-            "GROUP_CONCAT": lambda self: self.expression(
-                exp.GroupConcat,
-                this=self._parse_lambda(),
-                separator=self._match(TokenType.SEPARATOR) and self._parse_field(),
-            ),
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,32 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
 
             return self.expression(exp.Chr, **kwargs)
 
+        def _parse_group_concat(self) -> t.Optional[exp.Expression]:
+            def concat_exprs(node, exprs):
+                if node and isinstance(node, exp.Distinct) and len(node.expressions) > 1:
```
georgesittas, line 618:
```suggestion
                if isinstance(node, exp.Distinct) and len(node.expressions) > 1:
```

georgesittas: Let's add type hints to the closure.

sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,32 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
...
+                    concat_exprs = [self.expression(exp.Concat, expressions=node.expressions)]
+                    node.set("expressions", concat_exprs)
+                    return node
+                return (
+                    exprs[0] if len(exprs) == 1 else self.expression(exp.Concat, expressions=args)
+                )
+
+            args = self._parse_csv(self._parse_lambda)
+            order = args[-1] if isinstance(args[-1], exp.Order) else None
```
georgesittas, line 627: Let's instead do:

```python
order = seq_get(args, -1)
if not isinstance(order, exp.Order):
    order = None
```

Since SQLGlot allows partial parsing (`error_level=ErrorLevel.IGNORE`), we don't want this to raise an index error if there are no args.

sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,32 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
...
+
+            if order:
+                # Order By is the last (or only) expression in the list and has consumed the 'expr' before it,
+                # remove 'expr' from exp.Order and add it back to args
+                args.pop()
+                args.append(order.this)
+                order.set("this", concat_exprs(order.this, args))
+
+            this = order or concat_exprs(args[0], args)
```
georgesittas, line 636: Ditto about `seq_get(args, 0)` over the `args[0]` syntax

sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,32 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
...
+                    node.set("expressions", concat_exprs)
+                    return node
+                return (
+                    exprs[0] if len(exprs) == 1 else self.expression(exp.Concat, expressions=args)
+                )
+
+            args = self._parse_csv(self._parse_lambda)
+            order = args[-1] if isinstance(args[-1], exp.Order) else None
+
+            if order:
```
georgesittas, line 629:
```suggestion
            if order and args:
```

sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,32 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
...
+                # Order By is the last (or only) expression in the list and has consumed the 'expr' before it,
+                # remove 'expr' from exp.Order and add it back to args
+                args.pop()
+                args.append(order.this)
```
georgesittas, line 633: Maybe
```suggestion
                args[-1] = order.this
```

georgesittas (commented):
Looks good 👍

sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,37 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
 
             return self.expression(exp.Chr, **kwargs)
 
+        def _parse_group_concat(self) -> t.Optional[exp.Expression]:
+            def concat_exprs(
+                node: t.Optional[exp.Expression], exprs: t.List[exp.Expression]
+            ) -> exp.Expression:
+                if isinstance(node, exp.Distinct) and len(node.expressions) > 1:
+                    concat_exprs = [self.expression(exp.Concat, expressions=node.expressions)]
```
georgesittas, line 621: We need to set `safe=True` in this `Concat` instantiation and the one below, because MySQL allows `CONCAT(1, 2)` but other dialects don't. With `safe` we'll [cast](https://github.com/tobymao/sqlglot/blob/11fe3692090dc0feb940994ba6bfe839b4fdfce1/sqlglot/generator.py#L2469) the args to strings when transpiling.

georgesittas (approved):
Final comment, LGTM otherwise.

sqlglot/dialects/mysql.py
georgesittas: Let's add a test for this with the fix.

sqlglot/dialects/mysql.py
```diff
@@ -617,6 +613,41 @@ def _parse_chr(self) -> t.Optional[exp.Expression]:
...
+                    concat_exprs = [
+                        self.expression(exp.Concat, expressions=node.expressions, safe=True)
+                    ]
+                    node.set("expressions", concat_exprs)
+                    return node
+                return (
+                    exprs[0]
+                    if len(exprs) == 1
+                    else self.expression(exp.Concat, expressions=args, safe=True)
+                )
```
georgesittas, line 630:
```suggestion
                if len(exprs) == 1:
                    return exprs[0]
                return self.expression(exp.Concat, expressions=args, safe=True)
```

PR #3160, opened 2024-03-18
feat(snowflake): Adding support for DATE, TO_DATE, TRY_TO_DATE functions

### LINKED ISSUE #3152: transpile snowflake try_to_date to duckdb

```python
from sqlglot import transpile

transpile("SELECT try_to_date(x, 'MMDDYYYY')", read="snowflake", write="duckdb")
```
output
```
SELECT TRY_TO_DATE(x, 'MMDDYYYY')
```
however, duckdb has no `try_to_date` function, most relevant I think is [strptime](https://duckdb.org/docs/sql/functions/dateformat.html)

**Official Documentation**

snowflake related function [try_to_date](https://docs.snowflake.com/en/sql-reference/functions/try_to_date)
duckdb related function [strptime](https://duckdb.org/docs/sql/functions/dateformat.html)

VaggelisD:
Fixes #3152 

Add support for the mentioned functions by:

- Converting all of them to `exp.TsOrDsToDate` expressions if the format is present _or_ to date casts otherwise
- Adding a new argument `safe` in `exp.TsOrDsToDate` to preserve the roundtrip of `TRY_TO_DATE`, which has a different error handling behavior compared to `TO_DATE / DATE` 

For the transpilation issue #3152, no change was needed in DuckDB as it already transpiles the aforementioned expression to `strptime` (note that it produces a `TIMESTAMP`, thus the cast to `DATE`):

```
>>> import sqlglot
>>> sqlglot.transpile("SELECT try_to_date('01-01-1999', 'MM-DD-YYYY')", read="snowflake", write="duckdb")
SELECT CAST(STRPTIME('01-01-1999', '%m-%d-%Y') AS DATE)
```

Docs
-----------

1. [Snowflake TRY_TO_DATE](https://docs.snowflake.com/en/sql-reference/functions/try_to_date)
2. [Snowflake DATE / TO_DATE](https://docs.snowflake.com/en/sql-reference/functions/to_date)
3. [DuckDB strptime](https://duckdb.org/docs/sql/functions/dateformat.html#strptime-examples)


sqlglot/dialects/snowflake.py
```diff
@@ -46,7 +48,10 @@ def _builder(args: t.List) -> exp.Func:
                     return exp.UnixToTime(this=value, scale=seq_get(args, 1))
                 if not is_float(value.this):
                     return build_formatted_time(exp.StrToTime, "snowflake")(args)
-
+            if kind == exp.DataType.Type.DATE:
+                formatted_exp = build_formatted_time(exp.TsOrDsToDate, "snowflake")(args)
+                formatted_exp.set("safe", safe)
+                return formatted_exp
```
georgesittas, line 54: We need to be careful here, snowflake has `TO_DATE('<integer>')` - we can't produce a `TsOrDsToDate` for it because it represents a different concept. Check out the logic in the branch right above. I think I'd fall back to producing an anonymous function if we have the integer case for now.

sqlglot/dialects/snowflake.py
```diff
@@ -28,7 +28,9 @@
 
 
 # from https://docs.snowflake.com/en/sql-reference/functions/to_timestamp.html
-def _build_timestamp(name: str, kind: exp.DataType.Type) -> t.Callable[[t.List], exp.Func]:
+def _build_timestamp(
+    name: str, kind: exp.DataType.Type, safe: bool = False
+) -> t.Callable[[t.List], exp.Func]:
```
georgesittas, line 33: Let's rename this to `_build_datetime` since it now handles various types (time, timestamp, date).

sqlglot/dialects/snowflake.py
```diff
@@ -997,3 +1005,13 @@ def struct_sql(self, expression: exp.Struct) -> str:
                     values.append(e)
 
             return self.func("OBJECT_CONSTRUCT", *flatten(zip(keys, values)))
+
+        def tsordstodate_sql(self, expression: exp.TsOrDsToDate) -> str:
+            safe = expression.args.get("safe")
+            name = "TRY_TO_DATE" if safe else "TO_DATE"
```
georgesittas, line 1011:
```suggestion
            name = "TRY_TO_DATE" if expression.args.get("safe") else "TO_DATE"
```

sqlglot/dialects/snowflake.py
```diff
@@ -997,3 +1005,13 @@ def struct_sql(self, expression: exp.Struct) -> str:
...
+
+            this = self.sql(expression, "this")
+            format = self.sql(expression, "format")
+            format = f", {self.format_time(expression)}" if format else ""
+
+            return f"{name}({this}{format})"
```
georgesittas, line 1017:
```suggestion
            return self.func(name, expression.this, self.format_time(expression))
```

georgesittas: We can even put this whole thing in `TRANSFORMS` if you like, e.g.

```python
exp.TsOrDsToDate: lambda self,
e: self.func("TRY_TO_DATE" if e.args.get("safe") else "TO_DATE", e.this, self.format_time(e)),
```

sqlglot/dialects/snowflake.py
georgesittas: Let's add a test for this as well.

VaggelisD: I will change it so that the `'<integer>'` case falls back to anonymous function. The correct approach would be to map it to `UnixToTime` using the [appropriate scale](https://docs.snowflake.com/en/sql-reference/functions/to_date#usage-notes) but it's not straightforward and the docs mention a deprecation warning.

georgesittas: Sounds good

georgesittas (approved):

lei0zhou:
the command above works great. 
```
>>> import sqlglot
>>> sqlglot.transpile("SELECT try_to_date('01-01-1999', 'MM-DD-YYYY')", read="snowflake", write="duckdb")
SELECT CAST(STRPTIME('01-01-1999', '%m-%d-%Y') AS DATE)
```

but seems this command still not work, when the parameter is a column_name instead of a string
```
sqlglot.transpile("SELECT try_to_date(x, 'MMDDYYYY')", read="snowflake", write="duckdb")
```

georgesittas:
@VaggelisD this is because we've reused the logic that we have for `TO_TIMESTAMP`. In that function you can basically have both a numeric expression and a string expression as the first argument, so we couldn't safely convert it to a `StrToTime` without having a literal. It seems though that for `TO_DATE` and `TRY_TO_DATE` there's only one variant with two arguments - the `<str>, <fmt>` one - so it should be safe to convert it accordingly. Wanna take a look?

PR #3166, opened 2024-03-19
Fix: fill in missing implementation details for replace(None)

### LINKED ISSUE #3165: running simplify multiple times fails

```
>>> from sqlglot.optimizer.scope import exp
>>> ast=parse_one("SELECT a,c,b from table1 where 1 = 1;")
>>> w = ast.find(exp.Where)
>>> w2 = simplify(w)
>>> w2 = simplify(w)
  File "lib/python3.9/site-packages/sqlglot/optimizer/simplify.py", line 109, in simplify
    expression = while_changing(expression, _simplify)
  File "lib/python3.9/site-packages/sqlglot/helper.py", line 211, in while_changing
    expression = func(expression)
  File "lib/python3.9/site-packages/sqlglot/optimizer/simplify.py", line 106, in _simplify
    expression.replace(node)
  File "lib/python3.9/site-packages/sqlglot/expressions.py", line 624, in replace
    value = parent.args[key]
KeyError: 'where'
>>> 
```

in my code i run simplify multiple times as I reduce some elements of a query. in this case, elements in the where clause.

as you see in the sample code, first time works fine but then the where clause was removed and the code fails. 
This used to work before, so I guess some check that they key exists has been removed.

georgesittas:
Fixes #3165

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -621,7 +621,7 @@ class Expression(metaclass=_Expression):
             return expression
 
         key = self.arg_key
-        value = parent.args[key]
+        value = parent.args.get(key)
 
         if isinstance(value, list):
             index = self.index
@@ -639,7 +639,7 @@ class Expression(metaclass=_Expression):
                 else:
                     value[index] = expression
                     parent._set_parent(key, expression, index=index)
-        else:
+        elif value is not None:
             if expression is None:
                 parent.args.pop(key)
             else:
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -342,6 +342,9 @@ class TestOptimizer(unittest.TestCase):
     def test_simplify(self):
         self.check_file("simplify", simplify, set_dialect=True)
 
+        expression = parse_one("SELECT a, c, b FROM table1 WHERE 1 = 1")
+        self.assertEqual(simplify(simplify(expression.find(exp.Where))).sql(), "WHERE TRUE")
+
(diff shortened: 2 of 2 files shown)


tobymao:
i think the root issue though is that the expression's parent is not None

tobymao:
https://github.com/tobymao/sqlglot/commit/eb0a6c31c92d16abe785087271b14f7611ff24bc

streamnsight:
@tobymao @georgesittas 
Hi guys, while I appreciate the quick turnaround on this, it seems like there is another change in behavior here:
Shouldn't the `remove_where_true` function in simplify remove the `WHERE TRUE` clause altogether? 
If the test passes with a result of WHERE TRUE after 2 passes of simplify, something seems off.

georgesittas:
IIRC this should match the old behavior: `WHERE TRUE` was indeed removed from the original AST. Using your example:

```python
>>> ast=parse_one("SELECT a,c,b from table1 where 1 = 1;")
>>>
>>> w = ast.find(exp.Where)
>>> w2 = simplify(w)
>>> w2 = simplify(w)
>>>
>>> w2.sql()
'WHERE TRUE'
>>> ast
Select(
  expressions=[
    Column(
    Column(
      this=Identifier(this=c, quoted=False)),
    Column(
      this=Identifier(this=b, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=table1, quoted=False))))
>>> ast.sql()
'SELECT a, c, b FROM table1'
```

Did you mean something else?

PR #3167, opened 2024-03-19
fix(snowflake): Allow non-literal expressions too in DATE functions

VaggelisD:
Fixes #3160

Postfix to allow arbitrary expressions in DATE / TO_DATE / TRY_TO_DATE functions

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -33,7 +33,6 @@ def _build_datetime(
 ) -> t.Callable[[t.List], exp.Func]:
     def _builder(args: t.List) -> exp.Func:
         value = seq_get(args, 0)
-        int_value = False
 
         if isinstance(value, exp.Literal):
             int_value = is_int(value.this)
@@ -50,7 +49,7 @@ def _build_datetime(
                 if not is_float(value.this):
                     return build_formatted_time(exp.StrToTime, "snowflake")(args)
 
-        if len(args) == 2 and kind == exp.DataType.Type.DATE and not int_value:
+        if len(args) == 2 and kind == exp.DataType.Type.DATE:
             formatted_exp = build_formatted_time(exp.TsOrDsToDate, "snowflake")(args)
             formatted_exp.set("safe", safe)
             return formatted_exp
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -50,7 +50,7 @@ def _build_datetime(
                 if not is_float(value.this):
                     return build_formatted_time(exp.StrToTime, "snowflake")(args)
 
-        if kind == exp.DataType.Type.DATE and not int_value:
+        if len(args) == 2 and kind == exp.DataType.Type.DATE and not int_value:
             formatted_exp = build_formatted_time(exp.TsOrDsToDate, "snowflake")(args)
             formatted_exp.set("safe", safe)
             return formatted_exp
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -33,6 +33,7 @@ def _build_datetime(
 ) -> t.Callable[[t.List], exp.Func]:
     def _builder(args: t.List) -> exp.Func:
         value = seq_get(args, 0)
+        int_value = False
 
         if isinstance(value, exp.Literal):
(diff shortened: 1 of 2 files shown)


tobymao (approved):

sqlglot/dialects/snowflake.py
```diff
@@ -48,10 +49,12 @@ def _builder(args: t.List) -> exp.Func:
...
                     return build_formatted_time(exp.StrToTime, "snowflake")(args)
-            if kind == exp.DataType.Type.DATE and not int_value:
-                formatted_exp = build_formatted_time(exp.TsOrDsToDate, "snowflake")(args)
-                formatted_exp.set("safe", safe)
-                return formatted_exp
+
+        if kind == exp.DataType.Type.DATE and not int_value:
+            formatted_exp = build_formatted_time(exp.TsOrDsToDate, "snowflake")(args)
+            formatted_exp.set("safe", safe)
+            return formatted_exp
```
georgesittas, line 56: I think we must only hit this branch if there are exactly two arguments, right? Otherwise you risk parsing these incorrectly:

```
TO_DATE( <timestamp_expr> )
TO_DATE( '<integer>' )
TO_DATE( <variant_expr> )
```

georgesittas: Let's add a test to verify these are parsed / roundtripped as expected.

VaggelisD: If there's only a single argument, there are two cases:
- If it's a literal, it enters the initial branch and generates an `exp.Cast` instead of `exp.TsOrDsToDate`
- If it's not a literal (e.g. column name, function call) it generates an `exp.TsOrDsToDate` with the arbitrary expression

I'm not sure if there's an issue with the second case, to name a couple examples which execute fine on both engines:

```
DATE(TO_TIMESTAMP('2017-01-01 12:00:00'))
- Roundtrip: TO_DATE(CAST('2017-01-01 12:00:00' AS TIMESTAMPNTZ))
- DuckDB transpile: CAST(CAST('2017-01-01 12:00:00' AS TIMESTAMP) AS DATE)


WITH t AS (SELECT '2017-01-01' AS x) SELECT TO_DATE(t.x) FROM t
- Roundtrip: WITH t AS (SELECT '2017-01-01' AS x) SELECT TO_DATE(t.x) FROM t
- DuckDB transpile: WITH t AS (SELECT '2017-01-01' AS x) SELECT CAST(t.x AS DATE) FROM t
```

For the `TO_DATE('<integer>')` case, there's already a test case; It's parsed as an `exp.Anonymous` function because I made sure to preverse the `is not int_value` condition

georgesittas: What if you have this then?

```sql
with t as (select '31536000' as c)
select to_date(c) from t
```

Here the semantics of `to_date` is to convert a unix timestamp into a date, but you're producing a `TsOrDsToDate` which will be generated as a cast to date in duckdb and fail when we hit the engine. I guess that may be ok since producing an anonymous would also result in a crash since there's no try_to_date function, but it seems like the AST won't be accurate then.

sqlglot/dialects/snowflake.py
```diff
@@ -48,10 +49,12 @@ def _builder(args: t.List) -> exp.Func:
                     return exp.UnixToTime(this=value, scale=seq_get(args, 1))
                 if not is_float(value.this):
                     return build_formatted_time(exp.StrToTime, "snowflake")(args)
-            if kind == exp.DataType.Type.DATE and not int_value:
-                formatted_exp = build_formatted_time(exp.TsOrDsToDate, "snowflake")(args)
-                formatted_exp.set("safe", safe)
-                return formatted_exp
+
+        if len(args) == 2 and kind == exp.DataType.Type.DATE and not int_value:
```
georgesittas, line 53: Do we still need this?

```suggestion
        if len(args) == 2 and kind == exp.DataType.Type.DATE:
```

georgesittas: If not we can get rid of L36 as well.

georgesittas (approved):

sqlglot/dialects/snowflake.py
VaggelisD: Not really, it's covered implicitely

PR #3179, opened 2024-03-20
fix(clickhouse): Fixing FORMAT being parsed as implicit alias

### LINKED ISSUE #3176: Clickhouse SELECT ... FORMAT raises ParseError

```
import unittest

from sqlglot import parse_one
from sqlglot.dialects import ClickHouse


class TestClickhouseParseOneSelectFormat(unittest.TestCase):
    def test_parse_one_format_tab_separated(self):
        query = 'SELECT 1 FORMAT TabSeparated'
        ast = parse_one(sql=query, dialect=ClickHouse)
        self.assertIsNotNone(ast)

    def test_parse_one_format_native(self):
        query = 'SELECT 1 FORMAT Native'
        ast = parse_one(sql=query, dialect=ClickHouse)
        self.assertIsNotNone(ast)


if __name__ == '__main__':
    unittest.main()

```
**Native ParseError**

```
           sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 22.
             SELECT 1 FORMAT Native
```

**TabSeparated ParseError**

```
           sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 28.
             SELECT 1 FORMAT TabSeparated
```

**Official Documentation**
FORMAT clause official documentation: https://clickhouse.com/docs/en/sql-reference/statements/select/format 

Thanks in advance

VaggelisD:
Fixes #3176 

Although the `FORMAT` query modifier is already supported in SQLGLot, Clickhouse containts the anti-pattern that it's both a query modifier _and_ a valid identifier (in some cases, as shown below), so before this PR it was parsed as an alias for the query of the issue #3176 . 

However, it seems that it's not allowed in an "implicit" alias i.e without `AS` token, probably in an attempt to disambiguate it in the cases above. For example:

- `SELECT 1 FORMAT <format_val>` -> Valid, `FORMAT` is the query modifier with the following var as value
- `SELECT 1 AS FORMAT` -> Valid, `FORMAT` is the alias identifier of expression `1`
- `SELECT 1 FOO` -> Valid, `FOO` is the "implicit" alias identifier of expression `1`
- `SELECT 1 FORMAT` -> Invalid

This pattern might be present in other dialects (or in other Clickhouse keywords as well), and since Clickhouse uses the base Dialect functions `parse_alias`, `parse_id_var` etc the change was added there.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1053,6 +1053,9 @@ class Parser(metaclass=_Parser):
 
     UNNEST_OFFSET_ALIAS_TOKENS = ID_VAR_TOKENS - SET_OPERATIONS
 
+    # Aliases without 'AS' might cause disambiguity
+    EXCLUDED_IMPLICIT_ALIAS_IDS: t.List[str] = []
+
     STRICT_CAST = True
 
     PREFIXED_PIVOT_COLUMNS = False
@@ -5390,6 +5393,9 @@ class Parser(metaclass=_Parser):
         )
 
         if alias:
+            if any_token is None and alias.this.upper() in self.EXCLUDED_IMPLICIT_ALIAS_IDS:
+                self._retreat(self._index - 1)
+                return this
             this = self.expression(exp.Alias, comments=comments, this=this, alias=alias)
             column = this.this
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -405,7 +405,7 @@ class Parser(metaclass=_Parser):
         TokenType.WINDOW,
     }
 
-    ALIAS_TOKENS = ID_VAR_TOKENS.copy()
+    ALIAS_TOKENS = ID_VAR_TOKENS
 
     COMMENT_TABLE_ALIAS_TOKENS = TABLE_ALIAS_TOKENS - {TokenType.IS}
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -405,6 +405,8 @@ class Parser(metaclass=_Parser):
         TokenType.WINDOW,
     }
 
+    ALIAS_TOKENS = ID_VAR_TOKENS.copy()
+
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -5390,6 +5393,9 @@ def _parse_alias(
         )
 
         if alias:
+            if any_token is None and alias.this.upper() in self.EXCLUDED_IMPLICIT_ALIAS_IDS:
```
tobymao, line 5396: if not any_token

tobymao: why don't we just remove FORMAT from id_var why do we need excluded_implicit_alias_ids?

VaggelisD: From my testing, `FORMAT` is a valid identifier in all cases _except_ when it serves as an 'implicit' alias, so in a way Clickhouse allows this middle state. Removing it from `ID_VAR` will solve this issue but it will disallow it's general identifier use

georgesittas (approved):

sqlglot/parser.py
```diff
@@ -405,6 +405,8 @@ class Parser(metaclass=_Parser):
         TokenType.WINDOW,
     }
 
+    ALIAS_TOKENS = ID_VAR_TOKENS.copy()
```
tobymao, line 408: we don't need to copy here right?

VaggelisD: Hm, we don't need to copy because the base dialect will forever be as lenient in alias tokens as is in identifier tokens - making the reference in this case - and because in whatever sub-dialects do overwrite it, the set difference will create a copy anyways?

tobymao: yea

VaggelisD: I think that's fine then without the copy, good catch.

tobymao (approved):

PR #3180, opened 2024-03-20
Feat: transpile Snowflake's ADDTIME

georgesittas:
References:
- https://duckdb.org/docs/sql/functions/time
- https://docs.snowflake.com/en/sql-reference/functions/timeadd
- https://cloud.google.com/bigquery/docs/reference/standard-sql/time_functions#time_add
- https://docs.snowflake.com/en/sql-reference/functions/timestampadd


VaggelisD:
`DATEADD` can also accept timestamps, and there's also `TIMESTAMPADD` which is an alias of `DATEADD`. Would it be inside the scope of this PR to also include it as well using the new build functions and `exp.TimestampAdd`?

georgesittas:
Sure, should be an easy fix. I'll do it in a bit.

tobymao (approved):

PR #3190, opened 2024-03-21
fix: Fix backtracking through try/catch exceptions

### LINKED ISSUE #3175: ClickHouse: fail to parse "WITH <expression> AS <identifier>"
**Description**

`sqlglot` fails to parse `WITH <expression> AS <identifier>`.

**Before you file an issue**

- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")` -- Yes, I have specified the "read" dialect.
- Check if the issue still exists on main -- I have cloned and used the latest main branch. But the issue still exists.

```python
import sqlglot
from sqlglot.dialects.clickhouse import ClickHouse

query = "WITH 1 AS var SELECT var"
sqlglot.maybe_parse(query, dialect=ClickHouse(), error_level=sqlglot.ErrorLevel.RAISE)
```

**Official Documentation**

Link to [WITH Clause syntax and its example](https://clickhouse.com/docs/en/sql-reference/statements/select/with#syntax) available in the ClickHouse documentation website.

VaggelisD:
Fixes #3175 

The issue above highlighted 2 existing bugs, which stem from using `try/catch` statements in order to backtrack from wrong paths. This was the case with Clickhouses CTE, which supports 2 different kinds of syntax and the existing code tries to reuse the base parser's `parse_cte()` or backtrack from it and parse the second "flavor". Note that this pattern is also observed in `parse_schema`.

This approach can show different behaviors according to the user set ErrorLevel e.g. with `ErrorLevel.IGNORE` there's no error raised thus `try/catch` is not activated, so the introduced function `try_parse` attempts to solve this issue in a general way .

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -15,7 +15,6 @@ from sqlglot.dialects.dialect import (
     rename_func,
     var_map_sql,
 )
-from sqlglot.errors import ParseError
 from sqlglot.helper import is_int, seq_get
 from sqlglot.tokens import Token, TokenType
 
@@ -385,21 +384,20 @@ class ClickHouse(Dialect):
             return super()._parse_position(haystack_first=True)
 
         # https://clickhouse.com/docs/en/sql-reference/statements/select/with/
-        def _parse_cte(self) -> exp.CTE:
-            index = self._index
-            try:
-                # WITH <identifier> AS <subquery expression>
-                return super()._parse_cte()
-            except ParseError:
-                # WITH <expression> AS <identifier>
-                self._retreat(index)
+        def _parse_cte(self) -> exp.Expression:
+            # WITH <identifier> AS <subquery expression>
+            cte = super()._try_parse(super()._parse_cte)
 
-                return self.expression(
+            if not cte:
+                # WITH <expression> AS <identifier>
+                cte = self.expression(
                     exp.CTE,
                     this=self._parse_conjunction(),
                     alias=self._parse_table_alias(),
                     scalar=True,
                 )
+
+            return cte
 
         def _parse_join_parts(
             self,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1346,6 +1346,32 @@ class Parser(metaclass=_Parser):
             exp.Command, this=self._prev.text.upper(), expression=self._parse_string()
         )
 
+    def _try_parse(
+        self, parse_method: t.Callable, retreat: bool = False
+    ) -> t.Optional[exp.Expression]:
+        """
+        Attemps to backtrack if a parse function that contains a try/catch internally raises an error. This behavior can
+        be different depending on the uset-set ErrorLevel, so _try_parse aims to solve this by setting & resetting
+        the parser state accordingly
+        """
+        index = self._index
+        errors = self.errors.copy()
+        error_level = self.error_level
+
+        self.error_level = ErrorLevel.IMMEDIATE
+        this = None
+        try:
+            this = parse_method()
+        except ParseError:
+            this = None
+        finally:
+            if not this or retreat:
+                self._retreat(index)
+            self.error_level = error_level
+            self.errors = errors
+
+        return this
+
     def _parse_comment(self, allow_exists: bool = True) -> exp.Expression:
         start = self._prev
         exists = self._parse_exists() if allow_exists else None
@@ -2542,7 +2568,7 @@ class Parser(metaclass=_Parser):
             exp.With, comments=comments, expressions=expressions, recursive=recursive
         )
 
-    def _parse_cte(self) -> exp.CTE:
+    def _parse_cte(self) -> exp.Expression:
         alias = self._parse_table_alias(self.ID_VAR_TOKENS)
         if not alias or not alias.this:
             self.raise_error("Expected CTE to have alias")
@@ -4428,17 +4454,8 @@ class Parser(metaclass=_Parser):
         )
 
     def _parse_schema(self, this: t.Optional[exp.Expression] = None) -> t.Optional[exp.Expression]:
-        index = self._index
-
-        if not self.errors:
-            try:
-                if self._parse_select(nested=True):
-                    return this
-            except ParseError:
-                pass
-            finally:
-                self.errors.clear()
-                self._retreat(index)
+        if self._try_parse(lambda: self._parse_select(nested=True), retreat=True):
+            return this
 
         if not self._match(TokenType.L_PAREN):
             return this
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -385,22 +384,21 @@ def _parse_position(self, haystack_first: bool = False) -> exp.StrPosition:
...
-            index = self._index
-            try:
-                # WITH <identifier> AS <subquery expression>
-                return super()._parse_cte()
-            except ParseError:
-                # WITH <expression> AS <identifier>
-                self._retreat(index)
+        def _parse_cte(self) -> exp.Expression:
+            # WITH <identifier> AS <subquery expression>
+            cte = super()._try_parse(super()._parse_cte)
```
georgesittas, line 389:
```suggestion
            cte = self._try_parse(super()._parse_cte)
```

sqlglot/parser.py
```diff
@@ -1346,6 +1346,32 @@ def _parse_command(self) -> exp.Command:
...
+    def _try_parse(
+        self, parse_method: t.Callable, retreat: bool = False
+    ) -> t.Optional[exp.Expression]:
+        """
+        Attemps to backtrack if a parse function that contains a try/catch internally raises an error. This behavior can
+        be different depending on the uset-set ErrorLevel, so _try_parse aims to solve this by setting & resetting
+        the parser state accordingly
+        """
+        index = self._index
+        errors = self.errors.copy()
```
georgesittas, line 1358: Why are we copying here?

sqlglot/parser.py
```diff
@@ -1346,6 +1346,32 @@ def _parse_command(self) -> exp.Command:
...
+        error_level = self.error_level
+
+        self.error_level = ErrorLevel.IMMEDIATE
+        this = None
```
georgesittas, line 1362:
```suggestion
```

sqlglot/parser.py
```diff
@@ -1346,6 +1346,32 @@ def _parse_command(self) -> exp.Command:
...
+        try:
+            this = parse_method()
+        except ParseError:
+            this = None
+        finally:
+            if not this or retreat:
```
georgesittas, line 1368: Why would we ever _not_ want to retreat? Isn't the whole point here that `parse_method` could consume input that we need to successfully take an alternative parser path?

sqlglot/parser.py
```diff
@@ -1346,6 +1346,32 @@ def _parse_command(self) -> exp.Command:
             exp.Command, this=self._prev.text.upper(), expression=self._parse_string()
         )
 
+    def _try_parse(
+        self, parse_method: t.Callable, retreat: bool = False
+    ) -> t.Optional[exp.Expression]:
```
georgesittas, line 1351: Is it possible to improve the type hint here so that the return type is the same as `parse_method`'s?

tests/dialects/test_clickhouse.py
```diff
@@ -417,6 +418,15 @@ def test_cte(self):
         self.assertIsInstance(query.args["with"].expressions[0].this, exp.Subquery)
         self.assertEqual(query.args["with"].expressions[0].alias, "y")
 
+        query = "WITH 1 AS var SELECT var"
+        for error_level in [ErrorLevel.IGNORE, ErrorLevel.RAISE, ErrorLevel.IMMEDIATE]:
+            self.assertEqual(
+                parse_one(query, dialect=self.dialect, error_level=error_level).sql(
+                    dialect=self.dialect
+                ),
+                query,
```
georgesittas, line 427:
```suggestion
```suggestion
                self.parse_one(query, error_level=error_level).sql(dialect=self.dialect),
                query,
```
```

File: sqlglot/parser.py [hunk 7]
```diff
@@ -2542,7 +2568,7 @@ def _parse_with(self, skip_with_token: bool = False) -> t.Optional[exp.With]:
             exp.With, comments=comments, expressions=expressions, recursive=recursive
         )
 
-    def _parse_cte(self) -> exp.CTE:
+    def _parse_cte(self) -> exp.Expression:
```
georgesittas, line 2571: So, regarding the type hint comment, this should ideally stay as `exp.CTE`.

sqlglot/parser.py
tobymao, line 1362: i don't think this is necessary

tobymao (approved):

sqlglot/parser.py
georgesittas: Ah, nvm, I got it. LGTM 👍

sqlglot/parser.py
VaggelisD: This would be the case that `parse_method ` is on the right path and does not need to backtrack; in the Clickhouse case, `parse_method` is `None` if the CTE is expression-based, but if it's a typical CTE then `parse_method` contains the actual sub-tree, so retreating here wouldn't be correct right?

sqlglot/parser.py
VaggelisD: I tried to preserve it, but by using `try_parse` I was fighting with `make style` regarding this issue, will give it another shot

sqlglot/parser.py
VaggelisD: If the `error_level` isn't `IMMEDIATE`, the `try/catch` statements that are contained in `parse_method` might not fire at all (or too late, as in `RAISE`), right?

sqlglot/parser.py
georgesittas: Consuming input does not necessarily mean it's on the right path. We could consume input and then fail. I think the disconnect is that we should _always_ retreat, but only do it in the `except` branch.

sqlglot/parser.py
georgesittas: We explicitly set the error level here to `IMMEDIATE` though, right?

tests/dialects/test_clickhouse.py
VaggelisD: If the read dialect isn't set as `Clickhouse`, the expression-based CTE isn't parseable so this unit test would fail if I'm not losing something

georgesittas: Note the `self.` in front of `parse_one` - in tests you can do this and the dialect will be set for you automatically.

VaggelisD: Required a small refactor in `Validator::parse_one` to allow for kwarg propagation

VaggelisD:
A few notes regarding the latest commit:
- Changed it so that `parse_schema` can disambiguate between Schema definition and Subquery/CTE by looking ahead 1 token in search of select statement start tokens, instead of relying on `try_parse`
- Reset `try_parse` typehint to `exp.CTE` since it's now the only usecase in the code. If `parse_cte` can also be manually disambiguated, then `try_parse` can be removed altogether. Otherwise, if it's used in the future again for a different usecase it's new type hint can be resolved then e.g. in a union of hints

PR #3191, opened 2024-03-21
fix(postgres): Fix ARROW/DARROW column operators

### LINKED ISSUE #3185: Fails to produce SQL representation of transformed tree when a JSON key has a `-`
Having a dash in the key is valid but sqlglot warns against it and raises a `ValueError`

`2024-03-20 21:44:30 [ WARNING] Invalid JSON path syntax. Unexpected TokenType.DASH at index 1: en-US (dialect.py:485)`

` ValueError: Expected an Expression. Received <class 'bool'>: True`

```py
import sqlglot 

sql = """
    SELECT ("data" ->> 'en-US') AS "acat" FROM "my_table" 
    """
parsed_query = sqlglot.parse_one(sql, read="postgres")

def transformer(node):
    ### Actually doing stuff with the node but for example just returning it ###
    return node

transformed_tree = parsed_query.transform(transformer)

## Fails to parse with '-' in the json key.
modified_sql_query = transformed_tree.sql(dialect="postgres")
```

VaggelisD:
Fixes #3185 

Fix the following issue by overriding Postgres's `COLUMN_OPERATORS` for `TokenType.ARROW` and `TokenType.DARROW` to build `JsonExtractPath` & `JsonExtractPathScalar` expressions instead of calling `to_json_path`

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -357,6 +357,16 @@ class Postgres(Dialect):
 
         JSON_ARROWS_REQUIRE_JSON_TYPE = True
 
+        COLUMN_OPERATORS = {
+            **parser.Parser.COLUMN_OPERATORS,
+            TokenType.ARROW: lambda self, this, path: build_json_extract_path(
+                exp.JSONExtract, arrow_req_json_type=self.JSON_ARROWS_REQUIRE_JSON_TYPE
+            )([this, path]),
+            TokenType.DARROW: lambda self, this, path: build_json_extract_path(
+                exp.JSONExtractScalar, arrow_req_json_type=self.JSON_ARROWS_REQUIRE_JSON_TYPE
+            )([this, path]),
+        }
+
         def _parse_operator(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -357,6 +357,16 @@ class Parser(parser.Parser):
 
         JSON_ARROWS_REQUIRE_JSON_TYPE = True
 
+        COLUMN_OPERATORS = {
+            **parser.Parser.COLUMN_OPERATORS,
+            TokenType.ARROW: lambda self, this, path: build_json_extract_path(
```
tobymao, line 362: this should go in base, not postgres

georgesittas: There are some dialects (e.g. DuckDB) where you can do `x -> <json path>`, so the previous approach was valid for those. Should we instead count those that _do_ support it vs those that don't and refactor accordingly? I.e. if there are more in the latter category, have this in base and move the `to_json_path` to dialect.py or something.

tobymao: ok

tobymao (approved):

PR #3193, opened 2024-03-22
Add more Snowflake aggregate functions

ttzhou:
[Slack message genesis, while it lasts](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1710794131224809)

[Snowflake Aggregate Functions](https://docs.snowflake.com/en/sql-reference/functions-aggregation)

For Snowflake dialect:

* Change MEDIAN -> PERCENTILE_CONT
* Add CORR
* Add COVAR_POP
* Add COVAR_SAMP

Not sure if I am attempting too little or too much for the transpilation.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5745,12 +5745,24 @@ class Upper(Func):
     _sql_names = ["UPPER", "UCASE"]
 
 
+class PearsonCorrelation(Binary, AggFunc):
+    _sql_names = ["CORR"]
+
+
 class Variance(AggFunc):
     _sql_names = ["VARIANCE", "VARIANCE_SAMP", "VAR_SAMP"]
 
 
 class VariancePop(AggFunc):
     _sql_names = ["VARIANCE_POP", "VAR_POP"]
+
+
+class CovarianceSamp(Binary, AggFunc):
+    _sql_names = ["COVAR_SAMP"]
+
+
+class CovariancePop(Binary, AggFunc):
+    _sql_names = ["COVAR_POP"]
 
 
 class Week(Func):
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -5745,6 +5745,10 @@ class Upper(Func):
     _sql_names = ["UPPER", "UCASE"]
 
 
+class PearsonCorrelation(Binary, AggFunc):
```
tobymao, line 5748: can you just call this Corr?

tobymao: the can remove _sql_names

sqlglot/expressions.py
```diff
@@ -5753,6 +5757,14 @@ class VariancePop(AggFunc):
     _sql_names = ["VARIANCE_POP", "VAR_POP"]
 
 
+class CovarianceSamp(Binary, AggFunc):
```
tobymao, line 5760: same here,just rename them to CovarSamp

sqlglot/expressions.py
ttzhou: There are [other forms](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CORR_A.html#GUID-B2DED35A-2ECE-4DF0-BDA4-28F28B7BCA23) of correlations though, figured we'd be more explicit just in case.

georgesittas:
What are we waiting for this? Are you still working on something?

tobymao:
no i don’t think so

ttzhou:
Sorry @georgesittas - I went to bed and hadn't a chance to mark as ready for review :D You folks are too fast.

PR #3194, opened 2024-03-22
feat: redshift starts with support

eakmanrq:
Redshift doesn't support `startswith` so this generates a compatible equivalent of it


tobymao (approved):

georgesittas (approved):
LGTM, I wonder if there could be precedence issues due to the added `||` operator?

PR #3204, opened 2024-03-22
Fix(tokenizer): don't increment array cursor by 2 on CRLF

### LINKED ISSUE #3200: Windows line endings cause IndexOutOfRange during tokenization
I am parsing view and stored procedure definitions from windows-based mssql.

I am getting failures from IndexOutOfRange and strange parsing contexts where the leading character has been removed.

I think I have identified the problem + solution - are you open to a PR (with unit test)?

The issue is this:

In _advance:
https://github.com/tobymao/sqlglot/blob/a18444dbd7ccfc05b189dcb2005c85a1048cc8de/sqlglot/tokens.py#L1008

If it sees \r\n, then it sets i to 2
_current is increased by i
_char always looks at position _current-1

This can lead to it looking past the end of the source text, as _current-1 > size

I believe the fix is something like changing:
https://github.com/tobymao/sqlglot/blob/a18444dbd7ccfc05b189dcb2005c85a1048cc8de/sqlglot/tokens.py#L1022

to

```
        if self._end:
            self._peek = ""
            self._char = self.sql[self.size - 1]
        else:
            self._char = self.sql[self._current - 1]
            self._peek = self.sql[self._current]
```

georgesittas:
Fixes #3200

We used to increment `i` by 2 inside `_advance`, in order to basically skip the `\r\n` sequence and have the `line` attribute be automatically updated to the right value. However, this didn't take into account an edge case where the SQL string can end in `\r\n` and so it resulted in an off-by-one index error, as reported in the linked issue.

Since the original motivation for this change was to fix the `line` attribute, which was getting incremented twice for every `\r\n` sequence, I changed our approach to simply not increment it when we find the `\r` in the sequence and only do it on the subsequent `\n`.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -565,8 +565,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "~": TokenType.TILDA,
         "?": TokenType.PLACEHOLDER,
         "@": TokenType.PARAMETER,
-        # used for breaking a var like x'y' but nothing else
-        # the token type doesn't matter
+        # Used for breaking a var like x'y' but nothing else the token type doesn't matter
         "'": TokenType.QUOTE,
         "`": TokenType.IDENTIFIER,
         '"': TokenType.IDENTIFIER,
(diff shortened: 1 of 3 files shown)


mr-miles:
It's totally immaterial but I noticed the column gets set to 1 but the line number doesnt change til the next iteration ... how about

```
            if not(self._char == "\r" and self._peek == "\n"):
               self._col = 1
               self._line += 1
```

georgesittas:
> It's totally immaterial but I noticed the column gets set to 1 but the line number doesnt change til the next iteration ... how about
> 
> ```
>             if not(self._char == "\r" and self._peek == "\n"):
>                self._col = 1
>                self._line += 1
> ```
> 
> anyhow thanks for the swift attention!

Done! I liked this approach more, I think it simplifies the logic 👍

izeigerman (approved):

PR #3223, opened 2024-03-26
feat: Support for MySQL & Redshift UnixTotime

### LINKED ISSUE #3214: function from_unixtime trans error
source code:
from sqlglot import transpile
print(transpile("select from_unixtime(1711366265)", read="mysql", write="postgres"))
print(transpile("select from_unixtime(1711366265)", read="mysql", write="redshift"))

// output 
['SELECT FROM_UNIXTIME(1711366265)']
['SELECT FROM_UNIXTIME(1711366265)']

but postgres and redshift has no function from_unixtime,  will post error

VaggelisD:
Fixes #3214 

Introduce support for:
- MySQL `exp.UnixToTime` roundtrip according to syntax: `FROM_UNIXTIME(unix_timestamp [,format])`
- Redshift `exp.UnixToTime` generation only, as there's no equivalent function to transpile from/to

Postgres support already exists, utilizing the function `TO_TIMESTAMP(unix_timestamp)`.

Docs
----------
- [MySQL FROM_UNIXTIME](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_from-unixtime)
- [Redshift epoch to time](https://stackoverflow.com/questions/39815425/how-to-convert-epoch-to-datetime-redshift)
- [Postgres TO_TIMESTAMP](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TABLE)


sqlglot/expressions.py
```diff
@@ -5707,7 +5707,14 @@ class UnixToStr(Func):
 # https://prestodb.io/docs/current/functions/datetime.html
 # presto has weird zone/hours/minutes
 class UnixToTime(Func):
-    arg_types = {"this": True, "scale": False, "zone": False, "hours": False, "minutes": False}
+    arg_types = {
+        "this": True,
+        "format": False,
```
georgesittas, line 5712: Let's make `format` the last key in the dictionary to ensure there are no weird argument order shenanigans. For example if someone used `from_arg_list` and expected the second one to be `scale`, this change here would mess up the AST.

tests/dialects/test_mysql.py
```diff
@@ -581,6 +579,25 @@ def test_mysql_time(self):
...
+            },
+        )
+        self.validate_all(
+            "SELECT FROM_UNIXTIME(1711366265, '%Y %D %M')",
+            write={
+                "mysql": "SELECT FROM_UNIXTIME(1711366265, '%Y %D %M')",
+                "postgres": "SELECT TO_TIMESTAMP(1711366265)",
+                "redshift": "SELECT (TIMESTAMP 'epoch' + 1711366265 * INTERVAL '1 SECOND')",
+            },
+        )
```
georgesittas, line 600: Since the MySQL query doesn't really map well to postgres and redshift, can we just make this an identity test to avoid future confusion?

MySQL
```
mysql> SELECT FROM_UNIXTIME(1711366265, '%Y %D %M');
+---------------------------------------+
| FROM_UNIXTIME(1711366265, '%Y %D %M') |
+---------------------------------------+
| 2024 25th March                       |
+---------------------------------------+
1 row in set (0.00 sec)
```

Postgres:
```
georgesittas=# SELECT TO_TIMESTAMP(1711366265);
      to_timestamp
------------------------
 2024-03-25 13:31:05+02
(1 row)
```

Redshift:
```
SELECT (TIMESTAMP 'epoch' + 1711366265 * INTERVAL '1 SECOND')
-- ?column?
-- 2024-03-25 11:31:05
```

georgesittas (approved):
LGTM, just a couple of suggestions.

tests/dialects/test_mysql.py
VaggelisD: Was debating whether it's better to use `validate_all` in order to show that the `format` arg is dropped in transpile or to just use an identity check, will revert to the latter

georgesittas: Yeah, we should avoid creating tests like that because they're unstable and aren't really useful. If we want to put in the work and handle the unsupported arguments properly, we'd need to just do `self.unsupported` and check for an `UnsupportedError` in the validate all call.

PR #3226, opened 2024-03-26
Fix(optimizer): constrain UDTF scope boundary

### LINKED ISSUE #3209: Invalid SQL output from optimize function
Hello guys, I have been using the optimizer function, and I _think_ I have found a strange behaviour which is causing the optimize function to output an invalid query.

Whilst I found this isse on a trickier query, I have tried to create a simple example which illustrates the issue for reproducibility. Here it is:

```sql
WITH cte1
            AS (SELECT col1,
                       col2
                  FROM c1.d1.t1
                ),
     cte2
            AS (
                SELECT col1
                        , ARRAY_AGG(DISTINCT col3) AS col3_arr
                  FROM c1.d1.t2
                  WHERE col4 = 1
                  GROUP BY col1
                )
SELECT IF(cte1.col1 IN UNNEST(cte2.col3_arr), 1, 0) AS col5
FROM cte1
LEFT OUTER JOIN cte2 
USING (col1)
```

When I call optimize on this as follows:
```python
ast = parse_one(sql, dialect='bigquery')
print(optimize(ast,dialect='bigquery').sql(dialect='bigquery',pretty=True))
```

I get this: 

```sql
SELECT
  IF(`t1`.`col1` IN UNNEST(`cte2`.`col3_arr`), 1, 0) AS `col5`
FROM `c1`.`d1`.`t1` AS `t1`
```

However this is not valid sql - the CTE which is being referenced in the optimized query - `cte2` - is no longer defined, and the join relation has also been completely lost.

Is this an issue, or am I doing something wrong here?

Thanks guys!

_Originally posted by @rorynormaness in https://github.com/tobymao/sqlglot/discussions/3208_

georgesittas:
Fixes #3209


tests/fixtures/optimizer/optimizer.sql
```diff
@@ -1388,3 +1388,40 @@ WHERE
 ORDER BY
   COUNT(DISTINCT `cs1`.`cs_order_number`)
 LIMIT 100;
+
+# title: query that uses an UNNEST which doesn't introduce a new scope
```
barakalon, line 1392: Should we be (1) including regressions as-is or (2) simplifying them down to their core issue?

(1) gives us more confidence we're fixing a specific issue
(2) might be easier to debug if it ever fails later

barakalon (approved):

tests/fixtures/optimizer/optimizer.sql
georgesittas: Yeah, agreed that we should be simplifying these, which is why I added a minimal test in `test_optimizer.py` too. This one was more like a "bonus" end-to-end test in my view, but it may be redundant. Do you think we should just get rid of it?

barakalon: I don't have a strong opinion - just opening the discussion.

I like these string-to-string fixture tests better than the scope unit tests.

I also like the idea of a regression test suite. But I guess its _slightly_ higher on the "testing pyramid".

I would vote for a minimal string-to-string fixture test, unless the regression is a very important use case or captures some complexity that's hard to confidently simplify.

georgesittas: That makes sense to me 👍

Let me see if I can simplify this one a bit more. May just move it to the projection pushdown fixture file.

barakalon (approved):

PR #3234, opened 2024-03-27
feat(optimizer): Support for small integer CAST elimination

### LINKED ISSUE #3229: feature: optimizer: statically analyze conditions to check for literals
[Ibis sometimes generates fairly unoptimized SQL because it doesn't distinguish well between literals known at compile time, and values that could only be known at run time](https://github.com/ibis-project/ibis/pull/8786). Many times this could be avoided though. Note in that issue the `... CASE WHEN CAST(1 AS TINYINT) >= 0 ...`. This is trivial to see that this is always true. If we could statically analyze the expression and determine this is always true or always false, then Ibis could avoid a lot of the branching.

I think plugging this into the optimizer system.

In ibis we could keep track of if an expression is a literal ourselves, but then this would require use to carry around a bit of extra metadata around during the compilation process. This

**Additional context**
The simplest case would be be to optimize `CAST(1 AS TINYINT) >= 0` to `TRUE`. An extension could be to actually optimize away parent `CASE` conditions etc that use the condition. But that could be a followup, this first step would be adequate to solve our needs in ibis.

If you think this is a good feature, I would be happy to help with it, but with the caveat that I am unfamiliar with this codebase so it would be difficult for me. If you wanted to be a hero, it would be much appreciated :)

VaggelisD:
Fixes #3229 

This is the first step towards eliminating redundant casts. As an initial prototype, the elimination is done in the following cases:

- The cast follows the format `CAST(<integer literal> AS <integer type>)` 
- The cast is a LHS/RHS (or both) of a binary expression e.g. `CAST(1 AS INT) >= 1`. Standalone casts are not optimized away (even if they're seemingly redundant) as there could be side-effects e.g. conforming to column/schema specifications
- If the two previous rules apply, the  cast is optimized away only if (1) the `<integer literal>` is in the range `[-128, 127]` and it's casted to any signed type or (2) if it's in the range `[0, 255]` and it's casted to an unsigned type. This is done because in both cases, the upcast does not alter the literal's value; In contrast, a downcast can overflow the value, thus the general case is not supported as of now.

Any changes/additions to this optimization step must ensure that the optimizer preserves correctness by validating that the cast elimination is completely safe.


sqlglot/optimizer/simplify.py
```diff
@@ -547,7 +547,29 @@ def simplify_literals(expression, root=True):
...
 
 
+def _simplify_integer_cast(expr: exp.Expression) -> exp.Expression:
+    if isinstance(expr, exp.Cast) and expr.this.is_int:
+        num = int(expr.this.name)
+        # Remove the (up)cast from small (byte-sized) integers in predicates which is side-effect free. Downcasts on any
+        # integer type might cause overflow, thus the cast cannot be eliminated and the behavior is
+        # engine-dependent
+        if (
+            exp.DataType.TINYINT_MIN <= num <= exp.DataType.TINYINT_MAX
```
barakalon, line 557: How hard would it be to support all upcasts in one go?

sqlglot/optimizer/simplify.py
```diff
@@ -547,7 +547,29 @@ def simplify_literals(expression, root=True):
 NULL_OK = (exp.NullSafeEQ, exp.NullSafeNEQ, exp.PropertyEQ)
 
 
+def _simplify_integer_cast(expr: exp.Expression) -> exp.Expression:
+    if isinstance(expr, exp.Cast) and expr.this.is_int:
```
barakalon, line 551: Should we handle recursion?

```
CAST(CAST(1 AS SMALLINT) AS BIGINT)
```

barakalon (commented):
Some overlap with https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/canonicalize.py#L82

But I get why you're doing it here...

sqlglot/expressions.py
```diff
@@ -3973,6 +3981,11 @@ class Type(AutoName):
         Type.TIMETZ,
     }
 
+    TINYINT_MIN = -128
```
tobymao, line 3984: this should go in simplify, it may be dialect specific in the future

sqlglot/optimizer/simplify.py
georgesittas: For context, we discussed doing this in one go, but it may be a bit risky because the type widths aren't necessarily global for integer types. Handling a single byte was a conservative choice that seemed the most safe. The alternative would be to ensure that the widths are the same or control them per dialect so that the optimizer won't silently "fix" broken queries that would overflow if the data were corrupted.

georgesittas:
Let's also add a test for `CAST(x AS TINYINT) = 1`, i.e. there's no literal inside the `CAST`.

sqlglot/optimizer/simplify.py
```diff
@@ -547,7 +548,21 @@ def simplify_literals(expression, root=True):
...
+            (0 <= num <= 255) and expr.to.this in DataType.UNSIGNED_INTEGER_TYPES
+        ):
+            return expr.this
+
+    return expr
+
+
 def _simplify_binary(expression, a, b):
+    a = _simplify_integer_cast(a)
+    b = _simplify_integer_cast(b)
```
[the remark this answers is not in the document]
tobymao: you can do comparisons safely like = and > but not sure about all binary ops

tobymao:
> Some overlap with https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/canonicalize.py#L82
> 
> But I get why you're doing it here...

should this be moved to simplify?

sqlglot/optimizer/simplify.py
barakalon: Makes sense

georgesittas:
> > > Some overlap with https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/canonicalize.py#L82
> > > But I get why you're doing it here...
> > 
> > 
> > should this be moved to simplify?
> > ah maybe not because we don’t have types
> 
> Simplify runs after annotate types. There’s some awkward overlap between simplify and canonicalize, in general.
> 
> If dependency on types is the distinguisher, maybe this should move to canonicalize?

This doesn't seem like it needs type annotations though, right? It simply checks whether we have an integer in the cast, and it only applies the simplification if its value is in the appropriate range.

tobymao:
> > > > Some overlap with https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/canonicalize.py#L82
> > > > But I get why you're doing it here...
> > > 
> > > 
> > > should this be moved to simplify?
> > > ah maybe not because we don’t have types
> > 
> > 
> > Simplify runs after annotate types. There’s some awkward overlap between simplify and canonicalize, in general.
> > If dependency on types is the distinguisher, maybe this should move to canonicalize?
> 
> This doesn't seem like it needs type annotations though, right? It simply checks whether we have an integer in the cast, and it only applies the simplification if its value is in the appropriate range.

that's right, this doesn't need types

barakalon:
> > > > > Some overlap with https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/canonicalize.py#L82
> > > > > But I get why you're doing it here...
> > > > 
> > > > 
> > > > should this be moved to simplify?
> > > > ah maybe not because we don’t have types
> > > 
> > > 
> > > Simplify runs after annotate types. There’s some awkward overlap between simplify and canonicalize, in general.
> > > If dependency on types is the distinguisher, maybe this should move to canonicalize?
> > 
> > 
> > This doesn't seem like it needs type annotations though, right? It simply checks whether we have an integer in the cast, and it only applies the simplification if its value is in the appropriate range.
> 
> that's right, this doesn't need types

Ah, sure

sqlglot/optimizer/simplify.py
```diff
@@ -547,7 +553,34 @@ def simplify_literals(expression, root=True):
 NULL_OK = (exp.NullSafeEQ, exp.NullSafeNEQ, exp.PropertyEQ)
 
 
+def _simplify_integer_cast(expr: exp.Expression) -> exp.Expression:
+    if isinstance(expr, exp.Cast) and isinstance(expr.this, exp.Cast):
+        this = _simplify_integer_cast(expr.this)
```
tobymao, line 558: what happens if it's like

cast(cast(x as int) as bool) ?

VaggelisD: Good question; In an intermediate implementation I had created a closure called `is_integer_cast` that checked if the cast type is of `INTEGER_TYPE`, in order to make sure that the cast above wouldn't even recurse. However, I decided that it's probably superfluous; To illustrate, let's unroll the calls:

- On the first invocation of `simplify_integer_cast`, we see a double cast, so we recursively call it again on the first branch
- On the second invocation, we see `CAST(1 AS INT)` which is a viable optimization in our case, thus we return / bubble up the value `Literal(1)`. 
- Coming back up on the first invocation, we now have `this = Literal(1)` making the expression to `CAST(1 AS BOOL)`. That ends up being rejected as it's incorrect; Note that the recursive calls don't change the arguments in place, thus when the rejection happens on the top call, `return expr` makes sure to return the unoptimized version of the "top" expression as-is in order to repair the comparison. 

Is that enough to cover these cases, or should I add explicit checks to make sure that the cast type is of `INTEGER_TYPES` (which would also prune a function call as well)? Also, I thought I had actually added a test showcasing exactly that e.g. with nested casts to `TEXT`, but I must have removed it, will add a few now.

VaggelisD: Good question, and even though I had tested this locally and included a test case for it, I somehow can't find it so I'll make sure to include a few "faulty" ones on the next iteration. My thinking was that this case is handled as is, so let's unroll the calls to see it:

- On the first invocation of `_simplify_integer_cast`, we see a nested cast, thus we recurse
- On the second invocation of `_simplify_integer_cast`, we now see `CAST(1 as INT)`. Here, the cast is removed because the conditions are met, and thus we return / bubble-up the value `Literal(1)`
- Coming back on the first invocation, we now have `this = Literal(1)` so semantically the top-level expression is now equivalent to `CAST(1 AS BOOLEAN)`. This, of course, does not meet the criteria so it's "rejected" and we return the whole `expr` as is. Note that the recursion calls don't alter the arguments in place, so by preserving the nodes we've "repaired" the whole expression with no cost.

This is not to be confused with the canonicalize rule of `remove_redundant_cast` that on later optimization passes would actually remove `CAST(1 AS INT)`. If the nested casts where e.g. `CAST(CAST(1 AS BIGINT) AS BOOL)`, then we'd observe that both simplify and canonicalize would leave it as is.

tobymao (approved):

georgesittas (approved):

barakalon (approved):

PR #3241, opened 2024-03-28
Fix: properly parse/generate duckdb MAP {..} syntax, annotate MAPs

georgesittas:
Reference: https://duckdb.org/docs/sql/data_types/map.html


sqlglot/generator.py
```diff
@@ -3675,3 +3628,51 @@ def convert_sql(self, expression: exp.Convert) -> str:
             transformed = cast(this=value, to=to, safe=safe)
 
         return self.sql(transformed)
+
+    def _jsonpathkey_sql(self, expression: exp.JSONPathKey) -> str:
```
georgesittas, line 3632: Nothing to see here - just moved these underscore methods to the bottom of the generator file so they're grouped like helper methods. The only change is the `TRANSFORMS` entry.

tobymao (approved):

Issue #3247, opened 2024-03-30
Unable to parse Trino statement with non-escaped limit column


john-bodley:
According to the [documentation](https://trino.io/docs/current/language/reserved.html) `LIMIT` is not a reserved keyword, though it seems like SQLGlot is unable to parse statements referring to a column named `limit` unless it is escaped.

**Before you file an issue**
- [x] Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`
- [x] Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`
- [x] Check if the issue still exists on main

**Fully reproducible code snippet**
Please include a fully reproducible code snippet or the input sql, dialect, and expected output.

```python
>>> parse_one("SELECT limit FROM foo.bar", read="trino")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/__init__.py", line 124, in parse_one
    result = dialect.parse(sql, **opts)
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/dialects/dialect.py", line 490, in parse
    return self.parser(**opts).parse(self.tokenize(sql), sql)
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 1163, in parse
    return self._parse(
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 1229, in _parse
    expressions.append(parse_method(self))
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 1458, in _parse_statement
    expression = self._parse_set_operations(expression) if expression else self._parse_select()
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 2521, in _parse_select
    this = self._parse_query_modifiers(this)
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 2669, in _parse_query_modifiers
    key, expression = parser(self)
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 952, in <lambda>
    TokenType.LIMIT: lambda self: ("limit", self._parse_limit()),
  File "/Users/johnbodley/repos/github.com/sqlglot/sqlglot/parser.py", line 3620, in _parse_limit
    limit_exp = self.expression(
[...]

PR #3255, opened 2024-04-01
feat(snowflake): MATCH_CONDITION in ASOF JOIN

### LINKED ISSUE #3254: `MATCH_CONDITION` not recognised in `ASOF JOIN` for Snowflake
Snowflake recently added [the `ASOF` join type](https://docs.snowflake.com/en/sql-reference/constructs/asof-join) which introduces [the `MATCH_CONDITION` keyword](https://docs.snowflake.com/en/sql-reference/constructs/asof-join#syntax) in the join:

```sql
SELECT ...
FROM <left_table> ASOF JOIN <right_table>
  MATCH_CONDITION (<left_table.timecol> <comparison_operator> <right_table.timecol>)
  [ ON <table.col> = <table.col> [ AND ... ] ]
```

However, this doesn't appear to be supported in SQLGlot yet (details below).

SQLGlot to understand and parse the `MATCH_CONDITION` keyword when used as part of an `ASOF` join with the Snowflake dialect.

I don't think there are any alternatives, other than simply commenting out the `MATCH_CONDITION` line so that SQLGlot ignores it in the parsing 😄 

**Additional context**
I'll add an MWE here with corresponding screenshots -- I'm using `sqlglot==23.3.0` (with `sqlglotrs==0.1.3`).

Here's an example of a SQL query that runs fine in Snowflake:

```sql
WITH DATA(VAL) AS (
    SELECT *
    FROM VALUES (1), (2), (3)
)

SELECT *
FROM DATA AS DATA_L
    ASOF JOIN DATA AS DATA_R
        MATCH_CONDITION (DATA_L.VAL > DATA_R.VAL)
ORDER BY DATA_L.VAL
```

...and a screenshot to show that the result is printed as expected:

image

Here's a corresponding example of a Python module to parse this query:

```python
import sqlglot


SQL = """
WITH DATA(VAL) AS (
    SELECT *
    FROM VALUES (1), (2), (3)
)

SELECT *
FROM DATA AS DATA_L
    ASOF JOIN DATA AS DATA_R
        MATCH_CONDITION (DATA_L.VAL > DATA_R.VAL)
ORDER BY DATA_L.VAL
"""

print(sqlglot.parse_one(SQL, dialect="snowflake"))
```

...and a screenshot of the error I get when I run this:

image

VaggelisD:
Fixes #3254 

Syntax:

```
SELECT ...
FROM <left_table> ASOF JOIN <right_table>
  MATCH_CONDITION (<left_table.timecol> <comparison_operator> <right_table.timecol>)
  [ ON <table.col> = <table.col> [ AND ... ] ]
```

From the docs, the important parsing rules of `MATCH_CONDITION`:
- The parentheses are required
- The comparison operator cannot be of equality (thus, `self._parse_comparison` was used for finer granularity)

Docs
----------
- [Snowflake ASOF](https://docs.snowflake.com/en/sql-reference/constructs/asof-join)

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2907,6 +2907,9 @@ class Parser(metaclass=_Parser):
         if hint:
             kwargs["hint"] = hint
 
+        if self._match(TokenType.MATCH_CONDITION):
+            kwargs["match_condition"] = self._parse_wrapped(self._parse_comparison)
+
         if self._match(TokenType.ON):
             kwargs["on"] = self._parse_conjunction()
         elif self._match(TokenType.USING):
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -335,6 +335,11 @@ class Snowflake(Dialect):
     class Parser(parser.Parser):
         IDENTIFY_PIVOT_STRINGS = True
 
+        ID_VAR_TOKENS = {
+            *parser.Parser.ID_VAR_TOKENS,
+            TokenType.MATCH_CONDITION,
+        }
+
         TABLE_ALIAS_TOKENS = parser.Parser.TABLE_ALIAS_TOKENS | {TokenType.WINDOW}
 
         FUNCTIONS = {
@@ -698,6 +703,7 @@ class Snowflake(Dialect):
             "EXCLUDE": TokenType.EXCEPT,
             "ILIKE ANY": TokenType.ILIKE_ANY,
             "LIKE ANY": TokenType.LIKE_ANY,
+            "MATCH_CONDITION": TokenType.MATCH_CONDITION,
             "MATCH_RECOGNIZE": TokenType.MATCH_RECOGNIZE,
             "MINUS": TokenType.EXCEPT,
             "NCHAR VARYING": TokenType.VARCHAR,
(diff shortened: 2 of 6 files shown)


sqlglot/parser.py
```diff
@@ -2907,6 +2907,9 @@ def _parse_join(
         if hint:
             kwargs["hint"] = hint
 
+        if self._match(TokenType.MATCH_CONDITION):
```
georgesittas, line 2910: Can we just `_match_text_seq` here? We've been avoiding introducing new tokens.

VaggelisD: That was my initial implementation, but turns out that `a ASOF JOIN b MATCH_CONDITION(...)` will result in `MATCH_CONDITION` being parsed as `b`'s TableAlias, so I thought the proper way forward would be to declare a token in order to impose finer control. Are there any easy workarounds to this issue that I'm missing?

georgesittas: I see. I believe that the current implementation will also result in the behavior described in https://github.com/tobymao/sqlglot/issues/3247#issuecomment-2028481995.

Basically, `TABLE_ALIAS_TOKENS` won't contain the newly added token even though it's contained in `ID_VAR_TOKENS`. This is a side-effect and a non-trivial one at that. It may change in the future if we decide to create the former constant in the metaclass.

Maybe we can keep what you have here for now. Perhaps do this as well to make it more future-proof:

```python
TABLE_ALIAS_TOKENS.pop(TokenType.MATCH_CONDITION, None)
```

We can discuss later about how we want to deal with the linked issue.

georgesittas (approved):

PR #3256, opened 2024-04-01
Fix!: properly transpile escape sequences

georgesittas:
This change was motivated by the following issue, as [reported in Slack](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1711787293499419):

```python
>>> import sqlglot
>>> print(sqlglot.parse_one("select split('123|789', '\\\\|')", "spark").sql("spark"))
SELECT SPLIT('123|789', '\\|')
>>> print(sqlglot.parse_one("select split('123|789', '\\\\|')", "spark").sql("presto"))
SELECT REGEXP_SPLIT('123|789', '\\|')
```

The Spark -> Presto conversion here is incorrect because Presto doesn't use `\` as an escape character, and so including a second one changes the output:

```
spark-sql (default)> SELECT SPLIT('123|789', '\\|');
["123","789"]
```

vs

```
presto> SELECT REGEXP_SPLIT('123|789', '\\|');
           _col0           
---------------------------
 [, 1, 2, 3, |, 7, 8, 9, ] 
```

I believe the underlying issue was that we stored the escaped string in the token, and were not "unescaping" it at generation time. This PR amends this by storing the unescaped string at the tokenization stage so we can add escapes as needed at generation time, according to each dialect's rules.

In the process, I found that many of our tests were incorrect, either because the Python representation of the SQL strings wasn't properly escaped, or just simply because the query outputs didn't match (I assume some of them hadn't been tested).

I also implemented `sqlglot.tokenize` which is similar to `parse_one` and `transpile`. This made life a lot easier during testing and I can see it being a useful utility for sqlglot users.


georgesittas:
Cc @cpcloud you may be interested in this, I recall us chatting about a similar issue some time ago.

tests/dialects/test_hive.py
```diff
@@ -736,13 +739,12 @@ def test_hive(self):
...
     def test_escapes(self) -> None:
-        self.validate_identity("'\n'")
+        self.validate_identity("'\n'", "'\\n'")
         self.validate_identity("'\\n'")
-        self.validate_identity("'\\\n'")
+        self.validate_identity("'\\\n'", "'\\\\\\n'")
         self.validate_identity("'\\\\n'")
         self.validate_identity("''")
         self.validate_identity("'\\\\'")
-        self.validate_identity("'\\z'")
```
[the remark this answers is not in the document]
georgesittas: It's probably not an uncommon behavior in general. Python does something similar:

```
>>> print('\z')
\z

**Review by tobymao — APPROVED** [review]

File: sqlglot/dialects/dialect.py [hunk 2]
```diff
@@ -260,8 +260,8 @@ class Dialect(metaclass=_Dialect):
     If empty, the corresponding trie will be constructed off of `TIME_MAPPING`.
     """
 
-    ESCAPE_SEQUENCES: t.Dict[str, str] = {}
-    """Mapping of an unescaped escape sequence to the corresponding character."""
+    UNESCAPED_SEQUENCES: t.Dict[str, str] = {}
+    """Mapping of an unescaped sequence (`\\n`) to its escaped version (`\n`)."""
```
VaggelisD, line 264: Should this be "Mapping of an escaped sequence to it's unescaped version" ?

georgesittas:
FYI: renamed `ESCAPE_SEQUENCES` to `UNESCAPED_SEQUENCES` and `INVERSE_ESCAPE_SEQUENCES` to `ESCAPED_SEQUENCES` for clarity.

sqlglot/dialects/dialect.py
georgesittas: Yeah thanks for catching

PR #3258, opened 2024-04-02
Feat(prql): Add support for TAKE 

fool1280:
[PRQL take transform doc](https://prql-lang.org/book/reference/stdlib/transforms/take.html)

### DIFF
--- a/sqlglot/dialects/prql.py
+++ b/sqlglot/dialects/prql.py
@@ -29,6 +29,7 @@ class PRQL(Dialect):
         TRANSFORM_PARSERS = {
             "DERIVE": lambda self, query: self._parse_selection(query),
             "SELECT": lambda self, query: self._parse_selection(query, append=False),
+            "TAKE": lambda self, query: self._parse_take(query),
         }
 
         def _parse_statement(self) -> t.Optional[exp.Expression]:
@@ -77,6 +78,12 @@ class PRQL(Dialect):
             ]
 
             return query.select(*selects, append=append, copy=False)
+        
+        def _parse_take(self, query: exp.Query) -> exp.Query:
+            if self._match(TokenType.NUMBER): # TAKE <NUMBER>
+                return query.limit(self._prev.text) # LIMIT <NUMBER>
+            else:
+                self.raise_error("Error parsing 'TAKE' keyword")
 
         def _parse_expression(self) -> t.Optional[exp.Expression]:
             if self._next and self._next.token_type == TokenType.ALIAS:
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/prql.py
```diff
@@ -78,6 +79,22 @@ def _parse_selection(self, query: exp.Query, append: bool = True) -> exp.Query:
 
             return query.select(*selects, append=append, copy=False)
 
+        def _parse_take(self, query: exp.Query) -> exp.Query:
+            if self._match(TokenType.NUMBER):
+                first = self._prev.text  # TAKE 4..10 -> self._prev.text = "4."
```
georgesittas, line 84: This doesn't look right. We probably need to tokenize `1..10` differently. One way is to produce `[1, .., 10]`, i.e. create a `DDOT` token that represents `".."` and then add a method that parses `NUMBER DDOT NUMBER` into a `Slice` or something.

sqlglot/dialects/prql.py
```diff
@@ -78,6 +79,22 @@ def _parse_selection(self, query: exp.Query, append: bool = True) -> exp.Query:
...
+            else:
+                self.raise_error("Expecting a number")
+
+            if self._match_pair(TokenType.DOT, TokenType.NUMBER):  # TAKE <start>..<end>
+                first = int(first[:-1])
+                second = int(self._prev.text)
+                if second >= first:  # valid range
+                    return query.limit(second - first + 1).offset(first - 1)
+                else:
+                    return query.limit(0)
```
georgesittas, line 94: Why is this doing `limit(0)`?

fool1280: From the [PRQL Playground](https://prql-lang.org/playground/), the `from x take 1..0` seems to be a valid query and produce `select * from x limit 0`, so I want to handle that here.

sqlglot/dialects/prql.py
```diff
@@ -78,6 +79,11 @@ def _parse_selection(self, query: exp.Query, append: bool = True) -> exp.Query:
 
             return query.select(*selects, append=append, copy=False)
 
+        def _parse_take(self, query: exp.Query) -> exp.Query:
+            if self._match(TokenType.NUMBER):
+                val = self._prev.text  # TAKE <NUMBER>
+            return query.limit(int(val))  # TODO: TAKE for ranges a..b
```
georgesittas, line 85: Let's simplify this to:

```suggestion
            return query.limit(self._parse_number())  # TODO: TAKE for ranges a..b
```

georgesittas (approved):
One last comment and should be good to go

sqlglot/dialects/prql.py
fool1280: thanks George!

georgesittas: Sure thing! I guess we need to update the type hint so it's optional as well.

PR #3261, opened 2024-04-02
fix: Fix STRPOS for Presto & Trino

### LINKED ISSUE #3259: Converstion of snowflake POSITION to Trino STRPOS Is incorrect
When translating Snowflake dialect to Trino with the POSITION function it has a 3rd parameter allowing you to define the start position in the  string from which you can start looking for the expression where as in trino the 3rd argument in STRPOS in not the starting point in which to look in the string but the the N Instance of the character you are looking for.

document reference:
1. Snowflake POSITION - https://docs.snowflake.com/en/sql-reference/functions/position
2. Trino STRPOS - https://trino.io/docs/current/functions/string.html#strpos:~:text=0%20is%20returned.-,strpos,-(string%2C

Reproducible code : 

snowflake:
``` 
SELECT POSITION( 'A','ADFGAC', 3) => Returns position 5
```

Trino:
``` 
SELECT STRPOS( 'A','ADFGAC', 3) => Returns position 0 - looks for the 3 acurance of A
```

VaggelisD:
Fixes #3259 

Trino and Presto share the same strpos syntax: `strpos(string, substring, instance)`; This is different from all other dialects which use the 3rd argument as the starting position in the `string` to search from e.g. `strpos(string, substring, position)`. In order to preserve the start position semantics, the existing DuckDB - like pattern (wrapping the `string` with `SUBSTR` and offsetting the index by that pos) was utilized. This now correctly transpiles the faulty #3259 query while also preserving the Presto <-> Trino roundtrip if `instance` is used:

- Snowflake -> Trino/Presto (now, result in all 3 engines is 5)
```
>>> sqlglot.transpile("POSITION( 'A','ADFGAC', 3)", read="snowflake", write="presto")
["STRPOS(SUBSTR('ADFGAC', 3), 'A') + 3 - 1"]
```

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -619,7 +619,7 @@ def str_position_sql(
 
     if position:
         # Normalize third 'pos' argument into 'SUBSTR(..) + offset' across dialects
-        this = f"SUBSTR({this}, {position})"
+        this = self.func("SUBSTR", this, position)
         position_offset = f" + {position} - 1"
 
     return self.func("STRPOS", this, substr, instance) + position_offset
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -608,13 +608,18 @@ def no_map_from_entries_sql(self: Generator, expression: exp.MapFromEntries) -> 
     return ""
 
 
-def str_position_sql(self: Generator, expression: exp.StrPosition) -> str:
+def str_position_sql(
+    self: Generator, expression: exp.StrPosition, generate_instance: bool = False
+) -> str:
     this = self.sql(expression, "this")
     substr = self.sql(expression, "substr")
     position = self.sql(expression, "position")
+    instance = self.sql(expression, "instance") if generate_instance else None
+    instance = f", {instance}" if instance else ""
+
     if position:
-        return f"STRPOS(SUBSTR({this}, {position}), {substr}) + {position} - 1"
-    return f"STRPOS({this}, {substr})"
+        return f"STRPOS(SUBSTR({this}, {position}), {substr}{instance}) + {position} - 1"
+    return f"STRPOS({this}, {substr}{instance})"
 
 
 def struct_extract_sql(self: Generator, expression: exp.StructExtract) -> str:
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -614,12 +614,15 @@ def str_position_sql(
     this = self.sql(expression, "this")
     substr = self.sql(expression, "substr")
     position = self.sql(expression, "position")
-    instance = self.sql(expression, "instance") if generate_instance else None
-    instance = f", {instance}" if instance else ""
+    instance = expression.args.get("instance") if generate_instance else None
+    position_offset = ""
 
     if position:
-        return f"STRPOS(SUBSTR({this}, {position}), {substr}{instance}) + {position} - 1"
-    return f"STRPOS({this}, {substr}{instance})"
+        # Normalize third 'pos' argument into 'SUBSTR(..) + offset' across dialects
+        this = f"SUBSTR({this}, {position})"
+        position_offset = f" + {position} - 1"
+
+    return self.func("STRPOS", this, substr, instance) + position_offset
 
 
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/presto.py
```diff
@@ -546,3 +545,14 @@ def create_sql(self, expression: exp.Create) -> str:
             if kind == "VIEW" and schema.expressions:
                 expression.this.set("expressions", None)
             return super().create_sql(expression)
+
```
tobymao, line 548: is this duplicated anywhere?

VaggelisD: The `presto.py::strposition_sql` mirrors the `dialect.py::str_position_sql`. I didn't want to include `instance` generation in the base Dialect as this would mess up transpilations from Presto/Trino to other dialects, but we could merge these 2 by using a new (optional, defaults to False) boolean argument to choose when to generate the `instance` arg

tobymao: why don’t you just call the dialect version? why did you reimplement it?

VaggelisD: The Dialect version does not support generation of `instance` as it's currently implemented, so any Presto/Trino roundtrips or transpilations between each other would incorrectly drop it. However, as  the Dialect's `str_to_position_sql` generates exactly the same functions as Presto/Trino would, in my upcoming commit I've merged these 2 functions, hopefully that should be better.

sqlglot/dialects/dialect.py
```diff
@@ -608,13 +608,21 @@ def no_map_from_entries_sql(self: Generator, expression: exp.MapFromEntries) ->
...
     substr = self.sql(expression, "substr")
     position = self.sql(expression, "position")
+    instance = expression.args.get("instance") if generate_instance else None
+    position_offset = ""
+
     if position:
-        return f"STRPOS(SUBSTR({this}, {position}), {substr}) + {position} - 1"
-    return f"STRPOS({this}, {substr})"
+        # Normalize third 'pos' argument into 'SUBSTR(..) + offset' across dialects
+        this = f"SUBSTR({this}, {position})"
```
georgesittas, line 622:
```suggestion
        this = self.func("SUBSTR", this, position)
```

sqlglot/dialects/presto.py
```diff
@@ -401,7 +402,7 @@ class Generator(generator.Generator):
                 ]
             ),
             exp.SortArray: _no_sort_array,
-            exp.StrPosition: rename_func("STRPOS"),
+            exp.StrPosition: lambda self, e: str_position_sql(self, e, True),
```
georgesittas, line 405:
```suggestion
            exp.StrPosition: lambda self, e: str_position_sql(self, e, generate_instance=True),
```

tests/dialects/test_presto.py
```diff
@@ -177,6 +177,17 @@ def test_regex(self):
...
         )
+        self.validate_all(
+            "STRPOS('ABC', 'A', 3)",
+            read={
+                "trino": "STRPOS('ABC', 'A', 3)",
+            },
+            write={
+                "presto": "STRPOS('ABC', 'A', 3)",
+                "trino": "STRPOS('ABC', 'A', 3)",
+                "snowflake": "POSITION('A', 'ABC')",
```
georgesittas, line 188: Should we log an "unsupported" warning here? So if `instance` is set and the `generate_instance` is not, then `self.unsupported("Instance parameter is not supported by STRPOS")` or something.

VaggelisD: In general I'd be a fan of it, but note that not all dialects use `str_to_position_sql`. For example, Snowflake generates `exp.StrPosition` on it's own, so by transpiling Trino/Presto to Snowflake you won't hit the same unsupported warning as you would by transpiling to e.g. Postgres, even though both semantically would have `generate_instance = False`. Also, since having "overloaded" expressions with potentially unnecessary args is a common phenomenon, I thought that silently dropping `instance` is OK in this case. 

In the case that we do want to go forward with an unsupported warning, I've already prepared a commit with the necessary changes

georgesittas: That's ok, if there's a split like that in the generator method overrides we don't have to introduce complexity. This is something we could improve in the future.

PR #3273, opened 2024-04-03
Fix(optimizer): preserve the original type when creating a date literal

georgesittas:
This PR aims to address the following bug:

```python
>>> simplify(parse_one("date_trunc(cast('2023-10-01' as timestamp), quarter)", "bigquery")).sql("bigquery")
"CAST('2023-10-01 00:00:00' AS DATETIME)"
```

The problem is that `TIMESTAMP` is converted into `DATETIME`, but these are incompatible types in BigQuery and so the optimizer may break the original query.

Since `date_literal` is used in several places in `simplify.py`, this issue appears in them as well. Another example where the optimizer breaks a query:

```python
>>>
>>> sql = """
... WITH t AS (SELECT CAST('2020-01-01' AS TIMESTAMP) AS x)
... SELECT DATE_TRUNC(x, year) = CAST('2020-01-01' AS TIMESTAMP) FROM t
... """
>>>
>>> simplify(parse_one(sql, "bigquery")).sql("bigquery")
"WITH t AS (SELECT CAST('2020-01-01' AS TIMESTAMP) AS x) SELECT x < CAST('2021-01-01 00:00:00' AS DATETIME) AND x >= CAST('2020-01-01 00:00:00' AS DATETIME) FROM t"
```

Original:

```sql
-- f0_: True
WITH t AS (SELECT CAST('2020-01-01' AS TIMESTAMP) AS x)
SELECT DATE_TRUNC(x, year) = CAST('2020-01-01' AS TIMESTAMP) FROM t
```

Whereas the "simplified" results in the following error:

> No matching signature for operator < for argument types: TIMESTAMP, DATETIME. Supported signature: ANY < ANY at [3:64]


sqlglot/optimizer/simplify.py
```diff
@@ -1097,15 +1119,25 @@ def extract_interval(expression):
...
 
-def date_literal(date):
-    return exp.cast(
-        exp.Literal.string(date),
-        (
+def extract_type(*expressions):
+    target_type = None
+    for expression in expressions:
+        target_type = expression.to if isinstance(expression, exp.Cast) else expression.type
+        if target_type:
```
barakalon, line 1126: What is target_type is UNKNOWN?

barakalon (commented):
I brought this up in slack, but wanted to play with the idea more:

Looking at [bigquery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type), AFAICT the main difference is that TIMESTAMP has a timezone, DATETIME doesn't.

What if we attach tzinfo the date extracted by `extract_date` if its a TIMESTAMP.

Then, in `date_literal`, we do something like:
```python
if isinstance(date, datetime.datetime) and date.tzinfo:
    # TIMESTAMP
elif isinstance(date, datetime.datetime):
    # DATETIME
else:
    # DATE
```

Then we don't need to pass the target_type around everywhere.

I'm not sure how that would play for other dialects, but its probably no worse than before this PR.

sqlglot/optimizer/simplify.py
georgesittas: It should be ignored in `date_literal` using the `TEMPORAL_TYPES` check.

georgesittas:
> I brought this up in slack, but wanted to play with the idea more:
> 
> Looking at [bigquery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type), AFAICT the main difference is that TIMESTAMP has a timezone, DATETIME doesn't.
> 
> What if we attach tzinfo the date extracted by `extract_date` if its a TIMESTAMP.
> 
> Then, in `date_literal`, we do something like:
> 
> ```python
> if isinstance(date, datetime.datetime) and date.tzinfo:
>     # TIMESTAMP
> elif isinstance(date, datetime.datetime):
>     # DATETIME
> else:
>     # DATE
> ```
> 
> Then we don't need to pass the target_type around everywhere.
> 
> I'm not sure how that would play for other dialects, but its probably no worse than before this PR.

I'm not sure if we can produce a single date/time type here like we do in other normalizations, because different dialects may expect different types. First example that comes to mind is Snowflake with its different [timestamp variants](https://docs.snowflake.com/en/sql-reference/data-types-datetime#timestamp-ltz-timestamp-ntz-timestamp-tz).

tobymao:
what’s an example where the type would change? if we simplify something it’s type shouldn’t change because people are expecting a type in their schema

georgesittas:
> what’s an example where the type would change? if we simplify something it’s type shouldn’t change because people are expecting a type in their schema

Posted a couple of 'em in the description above, did you mean something else?

>  if we simplify something it’s type shouldn’t change because people are expecting a type in their schema

Agreed - these type changes (`TIMESTAMP` -> `DATETIME`) are happening in main branch.

tobymao (approved):

barakalon (approved):

sqlglot/optimizer/simplify.py
```diff
@@ -1097,15 +1119,25 @@ def extract_interval(expression):
...
+            break
+
+    return target_type
+
+
+def date_literal(date, target_type=None):
+    if not target_type or not target_type.is_type(*exp.DataType.TEMPORAL_TYPES):
```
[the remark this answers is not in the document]
georgesittas: e.g. when `target_type` is `UNKNOWN` or `NULL`

PR #3291, opened 2024-04-09
Feat(prql): Add FILTER 

fool1280:
1. Add support for `FILTER`: https://prql-lang.org/book/reference/stdlib/transforms/filter.html
2. Add support for `APPEND`, `REMOVE`, `INTERSECT` (some are experimental in prql): https://prql-lang.org/book/reference/stdlib/transforms/append.html

### DIFF
--- a/sqlglot/dialects/prql.py
+++ b/sqlglot/dialects/prql.py
@@ -83,7 +83,7 @@ class PRQL(Dialect):
             return query.limit(num) if num else None
 
         def _parse_filter(self, query: exp.Select) -> t.Optional[exp.Query]:
-            if self._match(TokenType.L_PAREN): # TODO: FILTER ( expr1 || expr2 && ...)
+            if self._match(TokenType.L_PAREN):  # TODO: FILTER ( expr1 || expr2 && ...)
                 filter = self._parse_conjunction()
                 if not self._match(TokenType.R_PAREN):
                     self.raise_error("Expecting )")
--- a/sqlglot/dialects/prql.py
+++ b/sqlglot/dialects/prql.py
@@ -83,7 +83,7 @@ class PRQL(Dialect):
             return query.limit(num) if num else None
 
         def _parse_filter(self, query: exp.Select) -> t.Optional[exp.Query]:
-            if self._match(TokenType.L_PAREN):
+            if self._match(TokenType.L_PAREN): # TODO: FILTER ( expr1 || expr2 && ...)
                 filter = self._parse_conjunction()
                 if not self._match(TokenType.R_PAREN):
                     self.raise_error("Expecting )")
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/prql.py
```diff
@@ -81,6 +85,36 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
             return query.limit(num) if num else None
 
+        def _parse_filter(self, query: exp.Select) -> t.Optional[exp.Query]:
+            if self._match(TokenType.L_PAREN):  # TODO: FILTER ( expr1 || expr2 && ...)
+                filter = self._parse_conjunction()
+                if not self._match(TokenType.R_PAREN):
+                    self.raise_error("Expecting )")
+            else:
+                filter = self._parse_conjunction()
+            return query.where(filter) if filter else None
```
georgesittas, line 95: Why do we need to separately handle the `(boolean_expression)` case? I think you only need a single `_parse_conjunction` call, the parser will take care of the rest.

So, assuming the above holds, I would just get rid of this whole method and do the following inline:

```python
            "FILTER": lambda self, query: query.where(self._parse_conjunction()),
```

sqlglot/dialects/prql.py
```diff
@@ -81,6 +85,36 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
+        def _parse_set_op(
+            self,
+            query: exp.Query,
+            append: bool = False,
+            intersect: bool = False,
+            remove: bool = False,
+        ) -> t.Optional[exp.Query]:
+            from_ = self._parse_from(skip_from_token=True)
+            if from_:
+                table = exp.select("*").from_(from_, copy=False)
```
georgesittas, line 106: Let's factor out the `select("*").from_(table, copy=False)` logic into a helper and use it to define the set operation mappings inline:

```python
            "APPEND": lambda self, query: query.union(
                _select_all(self._parse_table()), distinct=False, copy=False
            ),
            "REMOVE": lambda self, query: query.except_(
                _select_all(self._parse_table()), distinct=False, copy=False
            ),
            "INTERSECT": lambda self, query: query.intersect(
                _select_all(self._parse_table()), distinct=False, copy=False
            ),
```

Then, we could get rid of this method as well.

sqlglot/dialects/prql.py
```diff
@@ -81,6 +85,36 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
+            return query.where(filter) if filter else None
+
+        def _parse_set_op(
+            self,
+            query: exp.Query,
+            append: bool = False,
+            intersect: bool = False,
+            remove: bool = False,
+        ) -> t.Optional[exp.Query]:
+            from_ = self._parse_from(skip_from_token=True)
```
georgesittas, line 104: Why do we `_parse_from` here? Don't we only need the table part (`_parse_table`)?

georgesittas (approved):

georgesittas:
@tobymao what do you think about moving `select_all` to expressions.py? It seems like we use it in a few places.

PR #3295, opened 2024-04-10
Feat: improve transpilation of datetime functions to Teradata

maureen-daum:
These fixes resolve database/syntax errors encountered when using sql transpiled to the teradata dialect

- Cast doubles to DOUBLE PRECISION to avoid syntax errors
- Use `CURRENT_TIMETAMP` to avoid syntax errors
- Implement `DATE_ADD`/`DATE_SUB` with interval addition/subtraction
- Support week/quarter intervals by translating to multiplications over days
- Support `EXTRACT(QUARTER ...)` by using `TO_CHAR`

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -305,6 +305,14 @@ class Teradata(Dialect):
 
             return super().createable_sql(expression, locations)
 
+        def extract_sql(self, expression: exp.Extract) -> str:
+            this = self.sql(expression, "this")
+            if this.upper() != "QUARTER":
+                return super().extract_sql(expression)
+
+            to_char = exp.func("to_char", expression.expression, exp.Literal.string("Q"))
+            return self.sql(exp.cast(to_char, "int"))
+
         def interval_sql(self, expression: exp.Interval) -> str:
             multiplier = 0
             unit = expression.text("unit")
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -304,3 +304,17 @@ class Teradata(Dialect):
                 return f"{this_name}{this_properties}{self.sep()}{this_schema}"
 
             return super().createable_sql(expression, locations)
+
+        def interval_sql(self, expression: exp.Interval) -> str:
+            multiplier = 0
+            unit = expression.text("unit")
+
+            if unit.startswith("WEEK"):
+                multiplier = 7
+            elif unit.startswith("QUARTER"):
+                multiplier = 90
+
+            if multiplier:
+                return f"({multiplier} * {super().interval_sql(exp.Interval(this=expression.this, unit=exp.var('DAY')))})"
+
+            return super().interval_sql(expression)
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -13,6 +13,30 @@ from sqlglot.dialects.dialect import (
 from sqlglot.tokens import TokenType
 
 
+def _date_add_sql(
+    kind: t.Literal["+", "-"],
+) -> t.Callable[[Teradata.Generator, exp.DateAdd | exp.DateSub], str]:
+    def func(self: Teradata.Generator, expression: exp.DateAdd | exp.DateSub) -> str:
+        this = self.sql(expression, "this")
+        unit = expression.args.get("unit")
+
+        if expression.expression.is_negative:
+            kind_to_op = {"+": "-", "-": "+"}
+            simplify_expr = expression.expression.this
+        else:
+            kind_to_op = {"+": "+", "-": "-"}
+            simplify_expr = expression.expression
+
+        expression = self._simplify_unless_literal(simplify_expr)
+        if not isinstance(expression, exp.Literal):
+            self.unsupported("Cannot add non literal")
+
+        expression.set("is_string", True)
+        return f"{this} {kind_to_op[kind]} {self.sql(exp.Interval(this=expression, unit=unit))}"
+
+    return func
+
+
 class Teradata(Dialect):
     SUPPORTS_SEMI_ANTI_JOIN = False
     TYPED_DIVISION = True
@@ -216,6 +240,8 @@ class Teradata(Dialect):
             exp.ToNumber: to_number_with_nls_param,
             exp.Use: lambda self, e: f"DATABASE {self.sql(e, 'this')}",
             exp.CurrentTimestamp: lambda *_: "CURRENT_TIMESTAMP",
+            exp.DateAdd: _date_add_sql("+"),
+            exp.DateSub: _date_add_sql("-"),
         }
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -189,6 +189,7 @@ class Teradata(Dialect):
         TYPE_MAPPING = {
             **generator.Generator.TYPE_MAPPING,
             exp.DataType.Type.GEOMETRY: "ST_GEOMETRY",
+            exp.DataType.Type.DOUBLE: "DOUBLE PRECISION",
         }
 
         PROPERTIES_LOCATION = {
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -215,6 +215,7 @@ class Teradata(Dialect):
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
             exp.ToNumber: to_number_with_nls_param,
             exp.Use: lambda self, e: f"DATABASE {self.sql(e, 'this')}",
+            exp.CurrentTimestamp: lambda *_: "CURRENT_TIMESTAMP",
         }
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
--- a/tests/dialects/test_teradata.py
+++ b/tests/dialects/test_teradata.py
@@ -276,3 +276,19 @@ class TestTeradata(Validator):
                 "snowflake": "SELECT DATEADD(WEEK, 5, '2023-01-01')",
             },
         )
+        self.validate_all(
+            "CAST(TO_CHAR(x, 'Q') AS INT)",
+            read={
+                "teradata": "CAST(TO_CHAR(x, 'Q') AS INT)",
+                "snowflake": "DATE_PART(QUARTER, x)",
+                "bigquery": "EXTRACT(QUARTER FROM x)",
+            },
+        )
+        self.validate_all(
+            "EXTRACT(MONTH FROM x)",
+            read={
+                "teradata": "EXTRACT(MONTH FROM x)",
+                "snowflake": "DATE_PART(MONTH, x)",
+                "bigquery": "EXTRACT(MONTH FROM x)",
+            },
+        )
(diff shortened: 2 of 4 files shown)


sqlglot/dialects/teradata.py
```diff
@@ -13,6 +13,30 @@
...
+    def func(self: Teradata.Generator, expression: exp.DateAdd | exp.DateSub) -> str:
+        this = self.sql(expression, "this")
+        unit = expression.args.get("unit")
+
+        expression = self._simplify_unless_literal(expression.expression)
+        if not isinstance(expression, exp.Literal):
+            self.unsupported("Cannot add non literal")
+
+        this_sql = self.sql(expression, "this")
+        if this_sql.startswith("-"):
```
tobymao, line 28: this section doesn’t seems a bit suspect. you shouldn’t generate sql just to check if it’s a subtraction, just check the class of the ast node

maureen-daum: Agree it's not the nicest. What do you expect the class of the ast node to be? For a sql like `SELECT DATE_ADD('2023-01-01', -5, DAY)` on line 27 `expression` looks like `Literal(this=-5, is_string=False)`

tobymao: it shouldn't be that, it should be neg(literal)

```
python -m sqlglot "date_add('2020-01-01', -5)" --parse
DateAdd(
  this=Literal(this=2020-01-01, is_string=True),
  expression=Neg(
    this=Literal(this=5, is_string=False)))
```

maureen-daum: ah yup, the `_simplify_unless_literal` was folding the neg into the literal. I updated to check for the neg before that call

sqlglot/dialects/teradata.py
```diff
@@ -13,6 +13,29 @@
...
+            kind_to_op = {"+": "-", "-": "+"}
+            simplify_expr = expression.expression.this
+        else:
+            kind_to_op = {"+": "+", "-": "-"}
+            simplify_expr = expression.expression
+        expression = self._simplify_unless_literal(simplify_expr)
+        if not isinstance(expression, exp.Literal):
+            self.unsupported("Cannot add non literal")
+
+        expression.args["is_string"] = True
```
georgesittas, line 33:
```suggestion
        expression.args.set("is_string", True)
```

sqlglot/dialects/teradata.py
```diff
@@ -13,6 +13,29 @@
...
+        this = self.sql(expression, "this")
+        unit = expression.args.get("unit")
+
+        if isinstance(expression.expression, exp.Neg):
+            kind_to_op = {"+": "-", "-": "+"}
+            simplify_expr = expression.expression.this
+        else:
+            kind_to_op = {"+": "+", "-": "-"}
+            simplify_expr = expression.expression
+        expression = self._simplify_unless_literal(simplify_expr)
```
georgesittas, line 29: Nit:
```suggestion

        expression = self._simplify_unless_literal(simplify_expr)
```

sqlglot/dialects/teradata.py
```diff
@@ -276,3 +303,32 @@ def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> st
                 return f"{this_name}{this_properties}{self.sep()}{this_schema}"
 
             return super().createable_sql(expression, locations)
+
+        def extract_sql(self, expression: exp.Extract) -> str:
+            this = (
+                self.sql(expression, "this") if self.EXTRACT_ALLOWS_QUOTES else expression.this.name
+            )
```
georgesittas, line 310: This condition is always true right?

sqlglot/dialects/teradata.py
```diff
@@ -276,3 +303,32 @@ def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> st
...
+            if this.upper() != "QUARTER":
+                return super().extract_sql(expression)
+            expression_sql = self.sql(expression, "expression")
+            sql = f"TO_CHAR({expression_sql}, 'Q')"
+            return self.sql(
+                exp.Cast(
+                    this=sql,
+                    to=exp.DataType(this=exp.DataType.Type.INT),
+                )
+            )
```
georgesittas, line 320:
```suggestion
            to_char = exp.func("to_char", expression.expression, exp.Literal.string("Q"))
            return self.sql(exp.cast(to_char, "int"))
```

tests/dialects/test_teradata.py
```diff
@@ -210,3 +210,73 @@ def test_cast(self):
                 "teradata": "TRYCAST('-2.5' AS DECIMAL(5, 2))",
             },
         )
+
+    def test_time(self):
+        self.validate_identity("SELECT CURRENT_TIMESTAMP")
+
```
georgesittas, line 216: This is redundant, it's included in the following `validate_all` test.
```suggestion
```

tests/dialects/test_teradata.py
```diff
@@ -210,3 +210,73 @@ def test_cast(self):
...
+                "snowflake": "CURRENT_TIMESTAMP()",
+            },
+            write={
+                "teradata": "CURRENT_TIMESTAMP",
+                "snowflake": "CURRENT_TIMESTAMP()",
+            },
+        )
+
+        self.validate_all(
+            "SELECT DATE_ADD('2023-01-01', 5, YEAR)",
```
georgesittas, line 229: This isn't valid Teradata syntax, right? Let's use valid syntax for the first argument, because tests like these can be confusing. We can use `read` to show how `DATE_ADD` is transpiled _to_ Teradata.

georgesittas: Ditto about all other test cases that do this.

tests/dialects/test_teradata.py
```diff
@@ -210,3 +210,73 @@ def test_cast(self):
...
+            },
+        )
+        self.validate_all(
+            "SELECT INTERVAL '1 quarter'",
+            write={"teradata": "SELECT (90 * INTERVAL '1' DAY)"},
+        )
+        self.validate_all(
+            "SELECT INTERVAL '1 week'",
+            write={"teradata": "SELECT (7 * INTERVAL '1' DAY)"},
+        )
```
georgesittas, line 261: Ditto about using `self.validate_all` with `read` here. In general, though, when testing X -> X we prefer to use `self.validate_identity` because it's more compact.

sqlglot/dialects/teradata.py
```diff
@@ -13,6 +13,29 @@
...
 
 
+def _date_add_sql(
+    kind: t.Literal["+", "-"],
+) -> t.Callable[[Teradata.Generator, exp.DateAdd | exp.DateSub], str]:
+    def func(self: Teradata.Generator, expression: exp.DateAdd | exp.DateSub) -> str:
+        this = self.sql(expression, "this")
+        unit = expression.args.get("unit")
+
+        if isinstance(expression.expression, exp.Neg):
```
tobymao, line 23: what happens if there are multiple negatives? maybe it's better to simplify first, and add a method to expression which is like is_negative or something which handles the case? i can add this

tobymao: https://github.com/tobymao/sqlglot/commit/2ea438b89f76a357390d657fe3f9e01d2a79e7e4

sqlglot/dialects/teradata.py
maureen-daum: indeed it is, I'll simplify

tests/dialects/test_teradata.py
maureen-daum: Makes sense, I'll update these

tests/dialects/test_teradata.py
maureen-daum: Makes sense. I think in this case I'll make the sql the teradata sql, and then read the 1 week expression to make sure that we translate weeks to days

sqlglot/dialects/teradata.py
georgesittas: Yeah that was a typo, it needs to be `expression.set("is_string", True)`

maureen-daum: that works, thanks!

sqlglot/dialects/teradata.py
maureen-daum: this is much nicer to read!

sqlglot/dialects/teradata.py
maureen-daum: thanks for adding that! I updated to check `if expression.expression.is_negative`

sqlglot/dialects/teradata.py
```diff
@@ -276,3 +304,25 @@ def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> st
...
+            this = self.sql(expression, "this")
+            if this.upper() != "QUARTER":
+                return super().extract_sql(expression)
+
+            to_char = exp.func("to_char", expression.expression, exp.Literal.string("Q"))
+            return self.sql(exp.cast(to_char, "int"))
+
+        def interval_sql(self, expression: exp.Interval) -> str:
+            multiplier: t.Optional[int] = None
+            unit = expression.text("unit").lower()
```
tobymao, line 318: i think there may not be a need to call lower here, since it's always upper case, just change the startswith strings to upper case WEEK etc

sqlglot/dialects/teradata.py
```diff
@@ -276,3 +304,25 @@ def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> st
...
+
+            if unit.startswith("week"):
+                multiplier = 7
+            if unit.startswith("quarter"):
```
tobymao, line 322:
```suggestion
            elif unit.startswith("quarter"):
```

sqlglot/dialects/teradata.py
```diff
@@ -276,3 +304,25 @@ def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> st
...
+        def extract_sql(self, expression: exp.Extract) -> str:
+            this = self.sql(expression, "this")
+            if this.upper() != "QUARTER":
+                return super().extract_sql(expression)
+
+            to_char = exp.func("to_char", expression.expression, exp.Literal.string("Q"))
+            return self.sql(exp.cast(to_char, "int"))
+
+        def interval_sql(self, expression: exp.Interval) -> str:
+            multiplier: t.Optional[int] = None
```
tobymao, line 317:
```suggestion
            multiplier = 0
```

tests/dialects/test_teradata.py
```diff
@@ -210,3 +210,72 @@ def test_cast(self):
                 "teradata": "TRYCAST('-2.5' AS DECIMAL(5, 2))",
             },
         )
+
+    def test_time(self):
```
georgesittas, line 214: Let's also test teradata -> teradata to ensure the roundtrip is preserved for those tests where we only have eg. `read={"snowflake": ...}`. You can add `"teradata"` to the `read` dict as well.

georgesittas (approved):
Looks good, one final comment and should be good to go.

Thanks for the PR @maureen-daum 👍

tests/dialects/test_teradata.py
maureen-daum: good idea, I added teradata to the `read` dict for all of these tests 👍

maureen-daum:
> Looks good, one final comment and should be good to go.
> 
> Thanks for the PR @maureen-daum 👍

Thanks for the review! Just checking, is it ok that all of these commits have the `fix` prefix? (I'm not sure if you prefer `fix` or `feat` for these types of changes)

georgesittas:
> > Looks good, one final comment and should be good to go.
> > Thanks for the PR @maureen-daum 👍
> 
> Thanks for the review! Just checking, is it ok that all of these commits have the `fix` prefix? (I'm not sure if you prefer `fix` or `feat` for these types of changes)

Both would be fine, although I think I'd go with "feat" here because we're _improving_ the transpilation to Teradata. Appreciate the quick iteration!

PR #3297, opened 2024-04-10
Feat(prql): Add support for SORT

fool1280:
https://prql-lang.org/book/reference/stdlib/transforms/sort.html

### DIFF
--- a/sqlglot/dialects/prql.py
+++ b/sqlglot/dialects/prql.py
@@ -109,9 +109,11 @@ class PRQL(Dialect):
             if order and not self._match(TokenType.R_BRACE, expression=query):
                 self.raise_error("Expecting }")
             return (
-                query.order_by(order)
+                query.order_by(order, copy=False)
                 if order
-                else query.order_by(self._parse_order(skip_order_token=True))
+                else query.order_by(
+                    self.expression(exp.Order, expressions=self._parse_csv(self._parse_ordered))
+                )
             )
 
         def _parse_expression(self) -> t.Optional[exp.Expression]:
--- a/sqlglot/dialects/prql.py
+++ b/sqlglot/dialects/prql.py
@@ -105,16 +105,11 @@ class PRQL(Dialect):
             return query.limit(num) if num else None
 
         def _parse_order_by(self, query: exp.Select) -> t.Optional[exp.Query]:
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/prql.py
```diff
@@ -103,6 +104,16 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
             return query.limit(num) if num else None
 
+        def _parse_order_by(self, query: exp.Select) -> t.Optional[exp.Query]:
+            order = self._parse_sort(exp.Order, TokenType.L_BRACE)
+            if order and not self._match(TokenType.R_BRACE, expression=query):
+                self.raise_error("Expecting }")
+            return (
+                query.order_by(order)
+                if order
+                else query.order_by(self._parse_order(skip_order_token=True))
```
tobymao, line 114: copy=False everywhere

tobymao: i don't think you need parse_order here, you can just do 

self.expression(exp.Order, expressions=self._parse_csv(self._parse_ordered))

sqlglot/dialects/prql.py
```diff
@@ -103,6 +104,18 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
+            order = self._parse_sort(exp.Order, TokenType.L_BRACE)
+            if order and not self._match(TokenType.R_BRACE, expression=query):
+                self.raise_error("Expecting }")
+            return (
+                query.order_by(order, copy=False)
+                if order
+                else query.order_by(
+                    self.expression(exp.Order, expressions=self._parse_csv(self._parse_ordered))
+                )
+            )
```
georgesittas, line 117: I think this section needs to be:

```python
l_brace = self._match(TokenType.L_BRACE)
expressions = self._parse_csv(self._parse_ordered)
if l_brace and not self._match(TokenType.R_BRACE):
    self.raise_error("Expecting }")
...
```

AFAICT we can only omit the braces for the `sort` operator only when there's a single column, so this should cover both that and the multiple column case.

georgesittas: We'll probably need to override `_parse_ordered` to convert this syntax into nodes with `ASC` / `DESC`:

```
sort {age, -tenure, +salary}
```

georgesittas (approved):

PR #3299, opened 2024-04-11
Feat(prql): Handle DESC with sort

fool1280:


sqlglot/dialects/prql.py
```diff
@@ -104,8 +104,18 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
             num = self._parse_number()  # TODO: TAKE for ranges a..b
             return query.limit(num) if num else None
 
+        def _parse_ordered(
+            self, parse_method: t.Optional[t.Callable] = None
+        ) -> t.Optional[exp.Ordered]:
+            asc = self._match(TokenType.PLUS)
+            desc = self._match(TokenType.DASH) or (asc and False)
+            term = super()._parse_ordered(parse_method)
```
georgesittas, line 112: Nit:
```suggestion
            term = super()._parse_ordered(parse_method=parse_method)
```

sqlglot/dialects/prql.py
```diff
@@ -104,8 +104,18 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
+            if term and desc:
+                return term.desc()
```
georgesittas, line 114: This will create an `Ordered` under an `Ordered`, look into the definition of `desc`: https://github.com/tobymao/sqlglot/blob/b523bb630b28866ebab581d43e99f0b2b821ec12/sqlglot/expressions.py#L908-L909

You need to do `term.set("desc", True)` instead if `term and desc` here.

sqlglot/dialects/prql.py
```diff
@@ -104,8 +104,18 @@ def _parse_take(self, query: exp.Query) -> t.Optional[exp.Query]:
...
+            return term
+
         def _parse_order_by(self, query: exp.Select) -> t.Optional[exp.Query]:
-            l_brace = self._match(TokenType.L_BRACE)
+            l_brace = self._match(TokenType.L_BRACE)  # TODO: support sort with expression
```
georgesittas, line 118: It's not very clear what expressions you're referring to here. We do support expressions, just not s-strings yet. No need to have these TODOs imo since this is still in an experimental stage.

```suggestion
            l_brace = self._match(TokenType.L_BRACE)
```

tests/dialects/test_prql.py
```diff
@@ -41,10 +41,18 @@ def test_prql(self):
             "FROM x SORT age",
             "SELECT * FROM x ORDER BY age",
         )
+        self.validate_identity(
+            "FROM x SORT -age",
```
georgesittas, line 45: I'm not sure if this is valid PRQL syntax, IIRC you need to wrap `-age` in braces?

georgesittas: Btw, it also seems that PRQL's keywords are case-sensitive. So `FROM x` is not valid, but `from x` is. You can use their [playground](https://prql-lang.org/playground/) environment to check these things:

Screenshot 2024-04-11 at 1 39 35 PM

tests/dialects/test_prql.py
```diff
@@ -5,56 +5,56 @@ class TestPRQL(Validator):
     dialect = "prql"
 
     def test_prql(self):
-        self.validate_identity("FROM x", "SELECT * FROM x")
```
tobymao, line 8: why did you change these tests?

fool1280: > Btw, it also seems that PRQL's keywords are case-sensitive. So `FROM x` is not valid, but `from x` is. You can use their [playground](https://prql-lang.org/playground/) environment to check these things:
> 
> Screenshot 2024-04-11 at 1 39 35 PM

PRQL keyword are case sensitive so I just lowercase all of them

georgesittas (approved):

PR #3301, opened 2024-04-11
feat!: allow unions to be limited directly and stop subquerying since…

### LINKED ISSUE #3300: Solving: 'Union' object has no attribute 'order_by'
Filed this as a request as I'm not sure if it's a bug or by design.

Basically I'm trying to add an `order by` clause to a union which is valid, but when using the `sqlglot` primitives it's not straight forward.

The following approaches both give the error `AttributeError: 'Union' object has no attribute 'order_by'`
```py
import sqlglot

select = sqlglot.parse_one("select * from hi union select * from hello")

select.order_by('ts').sql()
sqlglot.parse_one(select.sql()).order_by('ts').sql()
```

But the following works
```py
# ...
sqlglot.parse_one(f'{select.sql()} order by ts').sql()
# prints 'SELECT * FROM hi UNION SELECT * FROM hello ORDER BY ts'
```

Should the `union` interface match the general `select` interface?

See above

tobymao:
… it is not necessary

closes #3300

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -668,6 +668,7 @@ class ClickHouse(Dialect):
         TABLE_HINTS = False
         EXPLICIT_UNION = True
         GROUPINGS_SEP = ""
+        OUTER_UNION_MODIFIERS = False
 
         # there's no list in docs, but it can be found in Clickhouse code
         # see `ClickHouse/src/Parsers/ParserCreate*.cpp`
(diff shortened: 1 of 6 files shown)


georgesittas (approved):
LGTM, just a couple of comments. FYI, all dialects apply the `LIMIT` to the `UNION` for this query:

```sql
SELECT * FROM t1 UNION ALL SELECT * FROM t2 LIMIT 1
```

... Except for ClickHouse, which applies it to the right branch of the union. Perhaps we can implement a transformation rule for set operations in CH so that if there are query modifiers attached directly to eg. a `UNION` we'll wrap the whole thing in a `SELECT *` and add them to that instead? IIRC the same thing applies for `ORDER BY` etc.

PR #3302, opened 2024-04-11
Fix!: cast less aggressively

georgesittas:
This addresses the following issue (note the additional `TIMESTAMP` added):

```python
>>> import sqlglot
>>> sql = "SELECT DATE_FORMAT(CAST(FROM_UTC_TIMESTAMP(foo::TIMESTAMP, 'America/Los_Angeles') AS TIMESTAMP), 'yyyy-MM-dd HH:mm:ss') AS foo FROM t"
>>> expr.sql("databricks")
"SELECT DATE_FORMAT(CAST(CAST(FROM_UTC_TIMESTAMP(CAST(foo AS TIMESTAMP), 'America/Los_Angeles') AS TIMESTAMP) AS TIMESTAMP), 'yyyy-MM-dd HH:mm:ss') AS foo FROM t"
```

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -197,8 +197,10 @@ def _ts_or_ds_add_sql(self: BigQuery.Generator, expression: exp.TsOrDsAdd) -> st
 
 
 def _ts_or_ds_diff_sql(self: BigQuery.Generator, expression: exp.TsOrDsDiff) -> str:
-    expression.this.replace(exp.cast(expression.this, "TIMESTAMP", copy=True))
-    expression.expression.replace(exp.cast(expression.expression, "TIMESTAMP", copy=True))
+    expression.this.replace(exp.cast(expression.this, exp.DataType.Type.TIMESTAMP, copy=True))
+    expression.expression.replace(
+        exp.cast(expression.expression, exp.DataType.Type.TIMESTAMP, copy=True)
+    )
     unit = unit_to_var(expression)
     return self.func("DATE_DIFF", expression.this, expression.expression, unit)
 
@@ -214,7 +216,9 @@ def _unix_to_time_sql(self: BigQuery.Generator, expression: exp.UnixToTime) -> s
     if scale == exp.UnixToTime.MICROS:
         return self.func("TIMESTAMP_MICROS", timestamp)
 
-    unix_seconds = exp.cast(exp.Div(this=timestamp, expression=exp.func("POW", 10, scale)), "int64")
+    unix_seconds = exp.cast(
+        exp.Div(this=timestamp, expression=exp.func("POW", 10, scale)), exp.DataType.Type.BIGINT
+    )
     return self.func("TIMESTAMP_SECONDS", unix_seconds)
 
 
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -197,10 +197,8 @@ def _ts_or_ds_add_sql(self: BigQuery.Generator, expression: exp.TsOrDsAdd) -> st
 
 
 def _ts_or_ds_diff_sql(self: BigQuery.Generator, expression: exp.TsOrDsDiff) -> str:
-    expression.this.replace(exp.cast(expression.this, exp.DataType.Type.TIMESTAMP, copy=True))
-    expression.expression.replace(
-        exp.cast(expression.expression, exp.DataType.Type.TIMESTAMP, copy=True)
-    )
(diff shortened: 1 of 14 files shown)


georgesittas:
Not sure if we want to do this at the `exp.cast` level instead of manually using `cast_unless` everywhere, wdyt @tobymao? I.e. replace `cast` with `cast_unless`.

tobymao:
let's change the behavior of cast to be cast unless, and then allow for flags to 1. change the type, 2. allow users to add cast no matter what, so this is a breaking change to cast

sqlglot/dialects/dialect.py
```diff
@@ -772,11 +772,11 @@ def no_timestamp_sql(self: Generator, expression: exp.Timestamp) -> str:
         from sqlglot.optimizer.annotate_types import annotate_types
 
         target_type = annotate_types(expression).type or exp.DataType.Type.TIMESTAMP
-        return self.sql(exp.cast(expression.this, to=target_type))
+        return self.sql(exp.cast(expression.this, target_type))
     if expression.text("expression").lower() in TIMEZONES:
         return self.sql(
             exp.AtTimeZone(
-                this=exp.cast(expression.this, to=exp.DataType.Type.TIMESTAMP),
+                this=exp.cast(expression.this, "timestamp"),
```
tobymao, line 779: i prefer using the data types in core code so that there's no parsing overhead

sqlglot/expressions.py
```diff
@@ -6871,7 +6871,14 @@ def column(
     return this
 
 
-def cast(expression: ExpOrStr, to: DATA_TYPE, copy: bool = True, **opts) -> Cast:
+def cast(
+    expression: ExpOrStr,
+    to: DATA_TYPE,
+    force_cast: bool = False,
```
tobymao, line 6877: maybe always: bool or force: bool

sqlglot/expressions.py
```diff
@@ -6871,7 +6871,14 @@ def column(
...
+    overwrite_cast: bool = False,
```
tobymao, line 6878: where this ever used? call it overwrite

tobymao:
this is a breaking change

sqlglot/dialects/dialect.py
georgesittas: sounds good

sqlglot/expressions.py
georgesittas: will do always

sqlglot/expressions.py
georgesittas: This isn't used anywhere, but I thought you meant to add this flag in (1), maybe I misinterpreted your suggestion?

> and then allow for flags to 1. change the type, 2. allow users to add cast no matter what

Seemed like a reasonable setting but I'm happy to remove.

tobymao: no because cast_unless had the api 

cast_unless(x, to:int, unless: varchar)

you don't have the unless argument that we used to have

tobymao: so it should be unless: t.Optional[DATA_TYPE]

georgesittas: The `unless` part wasn't used anywhere in the codebase as well, as in, in all cases `unless` matched `to`. Do you think we should still have it?

tobymao: i guess we can delete both

sqlglot/dialects/bigquery.py
```diff
@@ -197,8 +197,10 @@ def _ts_or_ds_add_sql(self: BigQuery.Generator, expression: exp.TsOrDsAdd) -> st
 
 
 def _ts_or_ds_diff_sql(self: BigQuery.Generator, expression: exp.TsOrDsDiff) -> str:
-    expression.this.replace(exp.cast(expression.this, "TIMESTAMP", copy=True))
-    expression.expression.replace(exp.cast(expression.expression, "TIMESTAMP", copy=True))
+    expression.this.replace(exp.cast(expression.this, exp.DataType.Type.TIMESTAMP, copy=True))
```
tobymao, line 200: copy is already true, we don't need it

tobymao (approved):

sqlglot/dialects/dialect.py
georgesittas: Changed all instances of string types to their `DataType.Type` equivalents

sqlglot/expressions.py
georgesittas: got rid of them both

sqlglot/dialects/bigquery.py
georgesittas: good catch 👍

PR #3303, opened 2024-04-11
Feat(teradata): handle transpile of quarter function

maureen-daum:
Add a `QUARTER` expression type. I tried to base this off of the places where `exp.Month` is used but may have missed something

Teradata doesn't have a `QUARTER` function to extract the quarter from a date, so transform it to an `EXTRACT(QUARTER ...)` expression. Teradata also doesn't support `EXTRACT(QUARTER ...)`, but this is already handled in `extract_sql`.

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -36,6 +36,10 @@ def _date_add_sql(
     return func
 
 
+def _quarter_sql(self: Teradata.Generator, expression: exp.Quarter) -> str:
+    return self.sql(exp.Extract(this="QUARTER", expression=expression.this))
+
+
 class Teradata(Dialect):
     SUPPORTS_SEMI_ANTI_JOIN = False
     TYPED_DIVISION = True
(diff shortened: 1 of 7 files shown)


sqlglot/dialects/teradata.py
```diff
@@ -325,3 +325,9 @@ def interval_sql(self, expression: exp.Interval) -> str:
                 return f"({multiplier} * {super().interval_sql(exp.Interval(this=expression.this, unit=exp.var('DAY')))})"
 
             return super().interval_sql(expression)
+
+        def anonymous_sql(self, expression: exp.Anonymous) -> str:
```
tobymao, line 329: you should never overwrite this, instead add Quarter as a first class function in expressions.py

maureen-daum: Gotcha, I'll update to do that

maureen-daum: I added a `exp.Quarter`, and  I tried to base this off of the places where exp.Month is used but may have missed something

tobymao (approved):

sqlglot/dialects/teradata.py
```diff
@@ -36,6 +36,10 @@ def func(self: Teradata.Generator, expression: exp.DateAdd | exp.DateSub) -> str
     return func
 
 
+def _quarter_sql(self: Teradata.Generator, expression: exp.Quarter) -> str:
+    return self.sql(exp.Extract(this="QUARTER", expression=expression.this))
```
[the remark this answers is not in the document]
tobymao: i did this change on top for you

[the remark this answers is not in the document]
maureen-daum: awesome, thanks!

PR #3304, opened 2024-04-11
Fix!: parse right-hand side of <value> IN (<query>) as a Subquery

georgesittas:
Before:

```python
>>> parse_one("x in (select 1)")
In(
  this=Column(
  query=Select(
    expressions=[
      Literal(this=1, is_string=False)]))
```

After:

```python
>>> parse_one("x in (select 1)")
In(
  this=Column(
  query=Subquery(
    this=Select(
      expressions=[
        Literal(this=1, is_string=False)])))
```

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -856,7 +856,7 @@ class Expression(metaclass=_Expression):
         **opts,
     ) -> In:
         subquery = maybe_parse(query, copy=copy, **opts) if query else None
-        if isinstance(subquery, Query) and not isinstance(subquery, Subquery):
+        if subquery and not isinstance(subquery, Subquery):
             subquery = subquery.subquery(copy=False)
 
         return In(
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -855,10 +855,14 @@ def isin(
         copy: bool = True,
         **opts,
     ) -> In:
+        subquery = maybe_parse(query, copy=copy, **opts) if query else None
+        if isinstance(subquery, Query) and not isinstance(subquery, Subquery):
+            subquery = subquery.subquery(copy=False)
```
georgesittas, line 860: Perhaps we can always wrap if it's a `Query`? this implementation intended for `isin(query="(select 1)")` to not produce an AST where the `Select` is wrapped twice.

barakalon:
```python
if subquery and not isinstance(subquery.Subquery):
    ...
```
?

barakalon (approved):

tobymao (approved):

PR #3306, opened 2024-04-12
Feat(bigquery): handle more extract parts

maureen-daum:
BigQuery has more limited `part` options for `Extract` than other dialects. Map from Snowflake and Databricks `parts` to the corresponding BigQuery one. This isn't a complete mapping but seems like a reasonable starting point.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -229,6 +229,44 @@ def _build_time(args: t.List) -> exp.Func:
     return exp.Anonymous(this="TIME", expressions=args)
 
 
+# Ref(values): https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#extract
+# Ref (keys): https://docs.snowflake.com/en/sql-reference/functions-date-time#supported-date-and-time-parts
+# Ref (keys): https://docs.databricks.com/en/sql/language-manual/functions/extract.html#returns
+EXTRACT_PART_MAPPING = {
+    "WEEKDAY": "DAYOFWEEK",
+    "DOW": "DAYOFWEEK",
+    "DW": "DAYOFWEEK",
+    "D": "DAY",
+    "DD": "DAY",
+    "DAYS": "DAY",
+    "DAYOFMONTH": "DAY",
+    "YEARDAY": "DAYOFYEAR",
+    "DY": "DAYOFYEAR",
+    "DOY": "DAYOFYEAR",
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -229,6 +229,44 @@ def _build_time(args: t.List) -> exp.Func:
     return exp.Anonymous(this="TIME", expressions=args)
 
 
+# Ref(values): https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#extract
+# Ref (keys): https://docs.snowflake.com/en/sql-reference/functions-date-time#supported-date-and-time-parts
+# Ref (keys): https://docs.databricks.com/en/sql/language-manual/functions/extract.html#returns
+EXTRACT_PART_MAPPING = {
```
tobymao, line 235: i'd like this to be done in a bit more generic way. right now, only bigquery has the ability to transpile snowflake/spark to itself. instead, for every dialect, we need to convert the part into a sqlglot variant, so for example

dd -> day, days -> day.

this may be good enough, then every dialect will be normalized.

if we want to go further we can define gen overrides, so for example sqlglot day -> DAYS if we wanted to, but that may not be necessary.

so to reiterate, add some base parser / dialect mapping all variants to a sqlglot normalized variant. if necessary (maybe it's not because there's no need), allow dialects to define a mapping from sqlglot variant to it's own variant.

maureen-daum: Makes sense, that's definitely a better solution. It'll take me a bit longer to do this. Is it ok that this means that some transpilations won't be an identity? (e.g., `transpile("EXTRACT(dd FROM x)", read="snowflake", write="snowflake")` -> `EXTRACT(day FROM x)`

tobymao: correct, it's never been a contract to produce identities. for exmaple

select x::int -> select (cast x as int)

georgesittas: Hey @maureen-daum 👋

Just wanted to double-check: is this a WIP on your end?

maureen-daum: Oh yup it is! I'll mark as a draft (or I can close and just open a new pr once it's ready)

georgesittas: @maureen-daum I'll go ahead and close this PR for now, let's iterate on a new one when it's ready.

PR #3307, opened 2024-04-12
Fix: transform eliminate_qualify on generated columns

viplazylmht:
Input:
```python
>>> sql = """
with data as (
  select 1 as id, 'a' as name
  union all
  select 1 AS ID, 'b' as name
)                
select
  id,
  concat(name, id) new_name,
from data
qualify row_number() over (partition by id order by new_name desc) = 1
"""
```

Before: 
```python
>>> from sqlglot.transforms import eliminate_qualify

>>> print(sqlglot.parse_one(sql, read="bigquery").transform(eliminate_qualify).sql(dialect="bigquery", pretty=True))
SELECT
  id,
  new_name
FROM (
  WITH data AS (
    SELECT
      1 AS id,
      'a' AS name
    UNION ALL
    SELECT
      1 AS ID,
      'b' AS name
  )
  SELECT
    id,
    CONCAT(name, id) AS new_name,
    row_number() OVER (PARTITION BY id ORDER BY new_name DESC) AS _w
  FROM data
) AS _t
WHERE
  _w = 1
```
Give the generated SQL before the patch to the bigquery console and perform dry-run, output error is: `Unrecognized name: new_name at [17:49]`

After the patch:
```python
>>> from sqlglot.transforms import eliminate_qualify

>>> print(sqlglot.parse_one(sql, read="bigquery").transform(eliminate_qualify).sql(dialect="bigquery", pretty=True))
SELECT
  id,
  new_name
FROM (
  WITH data AS (
    SELECT
      1 AS id,
      'a' AS name
    UNION ALL
    SELECT
      1 AS ID,
      'b' AS name
  )
  SELECT
    id,
    CONCAT(name, id) AS new_name,
    row_number() OVER (PARTITION BY id ORDER BY CONCAT(name, id) DESC) AS _w,
    name
  FROM data
) AS _t
WHERE
  _w = 1
```
The column `new_name` is replaced by `CONCAT(name, id)`, and the query can run successfully.

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -116,9 +116,11 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
         select_candidates = exp.Window if expression.is_star else (exp.Window, exp.Column)
         for expr in qualify_filters.find_all(select_candidates):
             if isinstance(expr, exp.Window):
-                for column in expr.find_all(exp.Column):
-                    if column.name in select_aliases:
-                        column.replace(select_aliases.get(column.name))
+                if select_aliases:
+                    for column in expr.find_all(exp.Column):
+                        select = select_aliases.get(column.name)
+                        if select:
+                            column.replace(select)
 
                 alias = find_new_name(expression.named_selects, "_w")
                 expression.select(exp.alias_(expr, alias), copy=False)
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -108,7 +108,7 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
         outer_selects = exp.select(*[select.alias_or_name for select in expression.selects])
         qualify_filters = expression.args["qualify"].pop().this
         select_aliases = {
-            select.alias_or_name: select.args.get("this")
+            select.alias: select.this
             for select in expression.selects
             if isinstance(select, exp.Alias)
         }
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -116,7 +116,9 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
                         if select.alias_or_name == column.name:
                             if isinstance(select, exp.Func):
                                 column.replace(select)
-                            elif isinstance(select, exp.Alias) and isinstance(select.args.get("this"), exp.Func):
+                            elif isinstance(select, exp.Alias) and isinstance(
+                                select.args.get("this"), exp.Func
+                            ):
                                 column.replace(select.args.get("this"))
 
                 alias = find_new_name(expression.named_selects, "_w")
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -116,7 +116,7 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
         select_candidates = exp.Window if expression.is_star else (exp.Window, exp.Column)
         for expr in qualify_filters.find_all(select_candidates):
             if isinstance(expr, exp.Window):
-                if select_aliases:
+                if expression_by_alias:
                     for column in expr.find_all(exp.Column):
                         select = select_aliases.get(column.name)
                         if select:
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -94,8 +94,8 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
     projections in the subquery, in order to refer to them in the outer filter using aliases. Also,
     if a column is referenced in the QUALIFY clause but is not selected, we need to include it too,
     otherwise we won't be able to refer to it in the outer query's WHERE clause. In adidion, if a
-    column is referenced in the QUALIFY clause but is build from a functio istead of a column or a
-    column alias, replace the column by the funtion inside.
+    column is referenced in the QUALIFY clause but is newly generated in the select clause (column
+    rename or create from a function), replace the column by the original.
     """
     if isinstance(expression, exp.Select) and expression.args.get("qualify"):
         taken = set(expression.named_selects)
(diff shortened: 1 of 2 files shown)


VaggelisD:
Hi @viplazylmht, good catch and thanks for your contribution. I'll leave a couple comments:

sqlglot/transforms.py
```diff
@@ -93,7 +93,9 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
     Some dialects don't support window functions in the WHERE clause, so we need to include them as
     projections in the subquery, in order to refer to them in the outer filter using aliases. Also,
     if a column is referenced in the QUALIFY clause but is not selected, we need to include it too,
-    otherwise we won't be able to refer to it in the outer query's WHERE clause.
+    otherwise we won't be able to refer to it in the outer query's WHERE clause. In adidion, if a
+    column is referenced in the QUALIFY clause but is build from a functio istead of a column or a
+    column alias, replace the column by the funtion inside.
```
VaggelisD, line 98: Please fix the typos here, there a few

sqlglot/transforms.py
```diff
@@ -109,6 +111,16 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
...
             if isinstance(expr, exp.Window):
+                for column in expr.find_all(exp.Column):
+                    for select in expression.selects:
+                        if select.alias_or_name == column.name:
+                            if isinstance(select, exp.Func):
+                                column.replace(select)
+                            elif isinstance(select, exp.Alias) and isinstance(
+                                select.args.get("this"), exp.Func
+                            ):
+                                column.replace(select.args.get("this"))
```
VaggelisD, line 122: Shouldn't we replace all aliases here with their original expression? A couple examples that will be missed now, if I'm not mistaken:

- `SELECT name as new_name`
- `SELECT name || 'test' as new_name`

These won't necessarily be `exp.Func`

VaggelisD:
Also, please include a test case in your PR, such examples are grouped in `test_transforms.py::test_eliminate_qualify()`

sqlglot/transforms.py
```diff
@@ -109,6 +111,16 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
         select_candidates = exp.Window if expression.is_star else (exp.Window, exp.Column)
         for expr in qualify_filters.find_all(select_candidates):
             if isinstance(expr, exp.Window):
+                for column in expr.find_all(exp.Column):
+                    for select in expression.selects:
```
georgesittas, line 115: We shouldn't be iterating over `selects` like this. Let's instead create a dict of `select.alias_or_name` -> `select` for all `expression.selects` and do a lookup with `column.name` as the key.

sqlglot/transforms.py
viplazylmht: You are right, in your first case, it is a column, I will update then.

The second case is a exp.Func because `||` will be parsed into exp.Concat I think. 

However, now I think the generated column will always covered by an alias, because without alias they can not reuse in other place. Tell me if I right then I will remove the non-alias part.

sqlglot/transforms.py
viplazylmht: thanks for your amazing knowledge!

sqlglot/transforms.py
georgesittas: No the alias is not required, eg. this is a valid BQ query:

```sql
with data as (
  select 1 as id, 'a' as name
  union all
  select 1 AS ID, 'b' as name
)
select
  id,
  name,
from data
qualify row_number() over (partition by id order by name desc) = 1
```

viplazylmht: @georgesittas in your case, `name` is a column from the `data` table, not generated one.

georgesittas: Ah, so I guess we wouldn't encounter a problem for it then after the transformation runs? If that's the case then what you said about doing this replacement only for aliases makes sense.

georgesittas: And if you do this, then you can add an additional filter in the dict I mentioned in my other comment so that only the aliased projections are included.

sqlglot/transforms.py
```diff
@@ -93,7 +93,9 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
     Some dialects don't support window functions in the WHERE clause, so we need to include them as
     projections in the subquery, in order to refer to them in the outer filter using aliases. Also,
     if a column is referenced in the QUALIFY clause but is not selected, we need to include it too,
-    otherwise we won't be able to refer to it in the outer query's WHERE clause.
+    otherwise we won't be able to refer to it in the outer query's WHERE clause. In adidion, if a
+    column is referenced in the QUALIFY clause but is newly generated in the select clause (column
+    rename or create from a function), replace the column by the original.
```
georgesittas, line 98:
```suggestion
    otherwise we won't be able to refer to it in the outer query's WHERE clause. Finally, if a
    newly aliased projection is referenced in the QUALIFY clause, it will be replaced by the
    corresponding expression to avoid creating invalid column references.
```

georgesittas (commented):
Looks good, I made a few final suggestions.

cc @tobymao in case you want to take a look. Planning to merge shortly otherwise.

sqlglot/transforms.py
```diff
@@ -105,10 +107,21 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
 
         outer_selects = exp.select(*[select.alias_or_name for select in expression.selects])
         qualify_filters = expression.args["qualify"].pop().this
+        select_aliases = {
```
georgesittas, line 110:
```suggestion
        expression_by_alias = {
```

sqlglot/transforms.py
```diff
@@ -105,10 +107,21 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
...
+            select.alias: select.this
+            for select in expression.selects
+            if isinstance(select, exp.Alias)
+        }
 
         select_candidates = exp.Window if expression.is_star else (exp.Window, exp.Column)
         for expr in qualify_filters.find_all(select_candidates):
             if isinstance(expr, exp.Window):
+                if select_aliases:
```
georgesittas, line 119:
```suggestion
                if expression_by_alias:
```

sqlglot/transforms.py
```diff
@@ -105,10 +107,21 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
...
+                    for column in expr.find_all(exp.Column):
+                        select = select_aliases.get(column.name)
+                        if select:
+                            column.replace(select)
```
georgesittas, line 123:
```suggestion
                        expr = expression_by_alias.get(column.name)
                        if expr:
                            column.replace(expr)
```

viplazylmht:
@georgesittas Jo after your refactoring, some tests are failed

georgesittas:
I fixed them after merging.

PR #3316, opened 2024-04-12
feat: allow qualify to also annotate on the fly for unnest support

tobymao:


tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -494,8 +494,13 @@ SELECT x AS x, y AS y FROM UNNEST([1, 2]) AS x WITH OFFSET AS y;
...
 # dialect: bigquery
 # execute: false
-select x, a, x.a from unnest([STRUCT(1 AS a)]) as x;
-SELECT x AS x, a AS a, x.a AS a FROM UNNEST([STRUCT(1 AS a)]) AS x;
+select x, a, x.a from unnest([STRUCT(1 AS a)]) as x CROSS JOIN m;
+SELECT x AS x, a AS a, x.a AS a FROM UNNEST([STRUCT(1 AS a)]) AS x CROSS JOIN m AS m;
+
+# dialect: bigquery
+# execute: false
+WITH cte AS (SELECT STRUCT(1 AS a) AS x) select a, x, x.a from cte, UNNEST(x) CROSS JOIN m;
```
georgesittas, line 502: Let's make `x` an array so this is valid BigQuery - we can't `UNNEST` a struct. Additionally, if you project `x.a` you then need to add an alias `AS x` to the `UNNEST`, otherwise you'll get the following error:

>  Cannot access field a on a value with type ARRAY<STRUCT<a INT64>> at ...

georgesittas (approved):
Nice!

PR #3317, opened 2024-04-15
fix(optimizer)!: Preserve DPipe in simplify_concat

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/2439

The `simplify_concat` rule would merge a list of binary `exp.DPipe` concatenations into one `exp.Concat` function call while constant folding any running string literals on the way. That (optimized) expression would then be generated as `CONCAT(a, b, ...)` which might not have the same semantics as `a || b` in every dialect e.g. `tsvector` in Postgres.

This PR aims to preserve `exp.DPipe` by tagging any `exp.Concat` expressions generated by `simplify_concat` with `dpipe_source=True`, which will cause them to be generated back as a list of (now optimized) binary `exp.DPipe` concatenations

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -767,6 +767,8 @@ def simplify_concat(expression):
         args = {
             "safe": expression.args.get("safe"),
             "coalesce": expression.args.get("coalesce"),
+            "dpipe_source": isinstance(expression, exp.DPipe)
+            or expression.args.get("dpipe_source"),
         }
 
     new_args = []
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4866,7 +4866,7 @@ class Chr(Func):
 
 
 class Concat(Func):
-    arg_types = {"expressions": True, "safe": False, "coalesce": False}
+    arg_types = {"expressions": True, "safe": False, "coalesce": False, "dpipe_source": False}
     is_var_len_args = True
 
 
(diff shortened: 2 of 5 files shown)


georgesittas:
Isn't the issue local to `simplify_concat`? Why do we need to introduce yet another arg to `Concat` for this? Imo this should be as simple as checking the type of the input expression and then ensuring we return that type before returning, eg. if we have a `Concat` at the end but the original type was `DPipe`, then do something like this (without the `sql` part):

https://github.com/tobymao/sqlglot/blob/4f1691a221f3d7395774f8c131a656a3ec531534/sqlglot/dialects/dialect.py#L875-L876

VaggelisD:
@georgesittas Check out the latest commit, `reduce` did simplify things a lot, thanks!

sqlglot/optimizer/simplify.py
```diff
@@ -784,6 +785,9 @@ def simplify_concat(expression):
     if concat_type is exp.ConcatWs:
         new_args = [sep_expr] + new_args
 
+    if concat_type is exp.Concat and isinstance(expression, exp.DPipe):
```
georgesittas, line 788: Nit:
```suggestion
    elif concat_type is exp.Concat and isinstance(expression, exp.DPipe):
```

georgesittas (approved):
LGTM

sqlglot/optimizer/simplify.py
```diff
@@ -783,6 +784,8 @@ def simplify_concat(expression):
 
     if concat_type is exp.ConcatWs:
         new_args = [sep_expr] + new_args
+    elif concat_type is exp.Concat and isinstance(expression, exp.DPipe):
```
georgesittas, line 787: We can actually simplify this a bit more
```suggestion
    elif isinstance(expression, exp.DPipe):
```

georgesittas (approved):

PR #3325, opened 2024-04-16
feat(Vertica): Add Vertica Dialect

Rajasekhar-Vuppala:
We are trying to add vertica dialect as we need this for datahub column level lineage. In this PR, We just added a base dialect and added datatypes only. upcoming PR's we will add more functionality.

We created Vertica.py we defined a class  Vertica that inherits from the Dialect class in the sqlglot library. This class helps translate SQL expressions to Vertica-specific syntax.  We imported tokenizer class to handle vertica specific quoting and identifiers , added keywords which extends default keywords by mapping vertica datatypes to their corresponding token type values. 
generator class extend from sqlglot generator class to translate vertica specifc representations and added type_mapping dictionary that maps sqlglot datatypes to vertica datatypes.

Added test cases for all the datatypes that we mentioned above.


tests/dialects/test_vertica.py
```diff
@@ -0,0 +1,9 @@
+from tests.dialects.test_dialect import Validator
+
+
+class VerticaDialectTest(Validator):
+    maxDiff = None  # type:ignore
+    dialect = "vertica"  # type:ignore
+
+    def test_vertica(self):
+        pass
```
georgesittas, line 9: Please add tests, and also adapt this to the style of other test suites, eg.

```python
class TestBigQuery(Validator):
    dialect = "bigquery"
    maxDiff = None
```

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,45 @@
...
+from sqlglot.generator import Generator
+
+
+class Vertica(Dialect):
+    # Ignoring mypy errors for forward declarations
+    class Tokenizer(Tokenizer):  # type: ignore
+        QUOTES = [
+            "'",
+            '"',
+        ]  # Strings can be delimited by either single or double quotes
```
georgesittas, line 13:
```suggestion
        QUOTES = ["'", '"']
```

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,45 @@
...
+        IDENTIFIERS = ["`"]  # Identifiers can be delimited by backticks
```
georgesittas, line 14:
```suggestion
        IDENTIFIERS = ["`"]
```

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # Include Vertica specific keywords
```
georgesittas, line 16:
```suggestion
```

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,45 @@
...
+            "FLOAT64": TokenType.DOUBLE,
+            "VARCHAR": TokenType.STRING,
+            "DATE": TokenType.DATE,
+            "TIMESTAMP": TokenType.TIMESTAMP,
+        }
+
+    # Ignoring mypy errors for forward declarations
+    class Generator(Generator):  # type: ignore
+        TRANSFORMS = {
+            exp.Array: lambda self, e: f"[{self.expressions(e)}]",
```
georgesittas, line 29: We should reuse `inline_array_sql`, see dialect.py

georgesittas (commented):
Please share docs for these changes.

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,40 @@
...
+        QUOTES = ["'", '"']
+        IDENTIFIERS = ["`"]
+
+        KEYWORDS = {
+            **Tokenizer.KEYWORDS,
+            "INT64": TokenType.BIGINT,
+            "FLOAT64": TokenType.DOUBLE,
+            "VARCHAR": TokenType.STRING,
+            "DATE": TokenType.DATE,
+            "TIMESTAMP": TokenType.TIMESTAMP,
```
georgesittas, line 19:
```suggestion
```

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,40 @@
...
+        exp.DataType.Type.SMALLINT: "INT64",
+        exp.DataType.Type.INT: "INT64",
+        exp.DataType.Type.BIGINT: "INT64",
+        exp.DataType.Type.DECIMAL: "NUMERIC",
+        exp.DataType.Type.FLOAT: "FLOAT64",
+        exp.DataType.Type.DOUBLE: "FLOAT64",
+        exp.DataType.Type.BOOLEAN: "BOOL",
+        exp.DataType.Type.TEXT: "STRING",
+        exp.DataType.Type.DATE: "DATE",
+        exp.DataType.Type.TIMESTAMP: "TIMESTAMP",
```
georgesittas, line 39:
```suggestion
```

sqlglot/dialects/vertica.py
Rajasekhar-Vuppala: Could you please provide us why we need to remove Date and timestamp?

sqlglot/dialects/vertica.py
Rajasekhar-Vuppala: Could you please provide us why we need to remove Date and timestamp?

sqlglot/dialects/vertica.py
georgesittas: They are redundant, you already have them in the base class

sqlglot/dialects/vertica.py
georgesittas: They're redundant, see the generator's datatype_sql method

sqlglot/dialects/vertica.py
```diff
@@ -0,0 +1,29 @@
...
+
+    class Generator(generator.Generator):
+        TYPE_MAPPING = {
+            exp.DataType.Type.TINYINT: "INT64",
+            exp.DataType.Type.SMALLINT: "INT64",
+            exp.DataType.Type.INT: "INT64",
+            exp.DataType.Type.BIGINT: "INT64",
+            exp.DataType.Type.DECIMAL: "NUMERIC",
+            exp.DataType.Type.FLOAT: "FLOAT64",
+            exp.DataType.Type.DOUBLE: "FLOAT64",
```
VaggelisD, line 26: Can you link to any documentation showing that all these numeric types are INT64/FLOAT64? 

From what I [found here](https://www.vertica.com/docs/9.3.x/HTML/Content/Authoring/SQLReferenceManual/DataTypes/Numeric/INTEGER.htm), all these types seem supported

tests/dialects/test_vertica.py
```diff
@@ -0,0 +1,88 @@
...
+        self.validate_all(
+            """
+            CREATE TABLE test_table (
+                col_smallint INT64
+            );
+            """,
+            write={
+                "vertica": """CREATE TABLE test_table (col_smallint INT64)""",
+            },
+        )
```
VaggelisD, line 28: If the query/DDL you're testing remains the same after "roundtripping" i.e parsed and generated back in Vertica, you can also use `validate_identity` for simplicity. Also, no need to include the semicolon at the end of the query

tests/dialects/test_vertica.py
```diff
@@ -0,0 +1,88 @@
+from tests.dialects.test_dialect import Validator
+
+
+class TestVertica(Validator):
+    dialect = "vertica"  # type:ignore
+    maxDiff = None  # type:ignore
```
VaggelisD, line 6: Are these `type:ignore` still needed here?

Rajasekhar-Vuppala: mypy error we are getting, we tried with other connectors in sqlglot we got the same error, that's why we added

Rajasekhar-Vuppala:
we are working on this sqlglot library very first time, we are still working on it. we will do changes and will request again for review

tests/dialects/test_vertica.py
VaggelisD: Make sure to format your PR's with sqlglot's `make style`, and for the specific versions of the packages you can run `make install-dev`. Also, the same formatter is ran as part of the ci/cd tests on every commit, so if something breaks you can also see it on this PR.

Let me know if that solution drops the error.

georgesittas:
> we are working on this sqlglot library very first time, we are still working on it. we will do changes and will request again for review

Assuming the scope of this PR is what's been already implemented, I'd say the action items to get this to a point where it can be merged are:

- Link any relevant documentation from Vertica's SQL specification. This is what I meant in my [previous review](https://github.com/tobymao/sqlglot/pull/3325#pullrequestreview-2003608614).
- Ensure the type overrides are correct and point us to the relevant section in the docs. Based on Vaggelis' [comment](https://github.com/tobymao/sqlglot/pull/3325#discussion_r1569005721), it seems like some of them may be incorrect.
- Refactor the tests:
  - Fix the mypy issues so that we can get rid of the `type: ignore`s (they seem unnecessary)
  - Reduce visual noise by using `validate_identity` instead of `validate_all`. The former is used when the same dialect is used for both input and output and we don't care about testing the transpilation towards other dialects
  - If we're only interested in testing that the type mappings are correct, then we don't need to use `CREATE` statements for this. We could simply use `CAST(x AS <type>)`, and perhaps avoid duplication by using a for loop to test the type mappings all at once (make sure to use `with self.subTest(...)` if you do this)

Rajasekhar-Vuppala:
Thanks for the review. We are closing this PR, We will raise new PR with all the changes and actions items you were mentioned.

PR #3327, opened 2024-04-16
feat!: Introduce partition in exp.Table

### LINKED ISSUE #3319: Hive. Query parsing fails when column is named application for insert overwrite partition
**Before you file an issue**
[x] Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`
[x] Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`
[x] Check if the issue still exists on main - issue is on 

```
import sqlglot
q = """
INSERT INTO a.b PARTITION (DAY = '2024-04-14')
(
  application,
  column
)
SELECT
  'app' AS application,
  column
FROM
a.c
"""
asts = sqlglot.parse(q, dialect='hive')

q = """
INSERT INTO a.b PARTITION (DAY = '2024-04-14')
(
  `application`,
  column
)
SELECT
  'app' AS `application`,
  column
FROM
a.c
"""
asts = sqlglot.parse(q, dialect='hive')
```

Exception:
```
File /opt/miniconda3/envs/databricks-repos/lib/python3.10/site-packages/sqlglot/parser.py:2553, in Parser._parse_select(self, nested, table, parse_subquery_alias, parse_set_operation)
   2546     this = (
   2547         self._parse_table()
   2548         if table
   2549         else self._parse_select(nested=True, parse_set_operation=False)
   2550     )
   2551     this = self._parse_query_modifiers(self._parse_set_operations(this))
-> 2553 self._match_r_paren()
   2555 # We return early here so that the UNION isn't attached to the subquery by the
   2556 # following call to _parse_set_operations, but instead becomes the parent node
   2557 return self._parse_subquery(this, parse_alias=parse_subquery_alias)

File /opt/miniconda3/envs/databricks-repos/lib/python3.10/site-packages/sqlglot/parser.py:6154, in Parser._match_r_paren(self, expression)
   6152 def _match_r_paren(self, expression: t.Optional[exp.Expression] = None) -> None:
   6153     if not self._match(TokenType.R_PAREN, expression=expression):
-> 6154         self.raise_error("Expecting )")

File /opt/miniconda3/envs/databricks-repos/lib/python3.10/site-packages/sqlglot/parser.py:1280, in Parser.raise_error(self, message, token)
   1268 error = ParseError.new(
   1269     f"{message}. Line {token.line}, Col: {token.col}.\n"
   1270     f"  {start_context}\033[4m{highlight}\033[0m{end_context}",
   (...)
   1276     end_context=end_context,
   1277 )
   1279 if self.error_level == ErrorLevel.IMMEDIATE:
-> 1280     raise error
   1282 self.errors.append(error)

ParseError: Expecting ). Line 4, Col: 14.

INSERT INTO a.b PARTITION (DAY = '2024-04-14')
```

**Official Documentation**
Nothing specific to this error, general docs: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DML

VaggelisD:
Fixes #3319 

Currently, `_parse_insert()` parses a partition clause _only after_ the table/schema is parsed through `_parse_table()`, which is not flexible enough for the following syntax (supported by atleast Hive, Databricks):

```
INSERT { OVERWRITE | INTO } [ TABLE ] table_name
    [ PARTITION clause ]
    [ ( column_name [, ...] ) | BY NAME ]
    query
```

This PR introduces `partition` arg in `exp.Table` in order to solve this constraint.

Also, as a by-product of this PR we might notice different (but semantically improved) behavior in other dialects. For example, in MySQL one can query specific partitions:

```
SELECT * FROM t1 PARTITION (p0)
```

Before this PR, the partition clause would be parsed as an alias of `t1`:
```
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=t1, quoted=False),
      alias=TableAlias(
        this=Identifier(this=partition, quoted=False),
        columns=[
          Identifier(this=p0, quoted=False)]))))
```

Now, the partition clause binds to the `exp.Table`: 
```
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=t1, quoted=False),
      partition=Partition(
        expressions=[
          Column(
            this=Identifier(this=p0, quoted=False))]))))
```


sqlglot/parser.py
```diff
@@ -3162,6 +3161,9 @@ def _parse_table(
         # Postgres supports a wildcard (table) suffix operator, which is a no-op in this context
         self._match_text_seq("*")
 
+        if self._match(TokenType.PARTITION, advance=False):
+            this.set("partition", self._parse_partition())
+
```
georgesittas, line 3166: I think this messes up the following query (valid in duckdb, postgres):

```sql
with t as (select 1 as c) select partition.c from t partition
```

VaggelisD: Hmm, `TokenType.PARTITION` is already a valid identifier token. Would it suffice to `parse_partition()` only if the token is followed by `L_PAREN`, otherwise retreat?

georgesittas: I don't think so, you can name columns in the alias so it'd be ambiguous:

```
D with t as (select 1 as c) select partition.d from t partition (d);
┌───────┐
│   d   │
│ int32 │
├───────┤
│     1 │
└───────┘
```

I think we need to conditionally do this only for dialects that support this clause.

sqlglot/parser.py
```diff
@@ -1109,6 +1109,9 @@ class Parser(metaclass=_Parser):
     # Whether or not interval spans are supported, INTERVAL 1 YEAR TO MONTHS
     INTERVAL_SPANS = True
 
+    # Whether the dialect allows syntax such as: select * from t1 partition (p0)
```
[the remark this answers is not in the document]
VaggelisD: Sounds good. 

Testing the MySQL codepath for RTT should yield the same query back in both main and in this PR, but the AST is different. Do you mean to test it's partition selection by somehow checking the `partition` arg in `exp.Table`?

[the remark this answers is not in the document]
georgesittas: How is it different? You mean now we parse an actual partition clause instead of the alias, right? If so, then for

> Do you mean to test it's partition selection by somehow checking the partition arg in exp.Table?

A simple validate identity will suffice (since we won't have the `AS ...` now)

PR #3340, opened 2024-04-22
Feat: allow comments to be attached for identifiers used in definitions

georgesittas:
Context: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1713777996852009

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4216,7 +4216,10 @@ class Parser(metaclass=_Parser):
         ):
             this = self._parse_id_var()
 
-        return self.expression(exp.Column, this=this) if isinstance(this, exp.Identifier) else this
+        if isinstance(this, exp.Identifier):
+            this = self.expression(exp.Column, comments=this.pop_comments(), this=this)
+
+        return this
 
     def _parse_column_ops(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
         this = self._parse_bracket(this)
@@ -4319,7 +4322,6 @@ class Parser(metaclass=_Parser):
         any_token: bool = False,
         tokens: t.Optional[t.Collection[TokenType]] = None,
         anonymous_func: bool = False,
-        comment: bool = False,
     ) -> t.Optional[exp.Expression]:
         if anonymous_func:
             field = (
@@ -4330,7 +4332,7 @@ class Parser(metaclass=_Parser):
             field = self._parse_primary() or self._parse_function(
                 anonymous=anonymous_func, any_token=any_token
             )
-        return field or self._parse_id_var(any_token=any_token, tokens=tokens, comment=comment)
+        return field or self._parse_id_var(any_token=any_token, tokens=tokens)
 
     def _parse_function(
         self,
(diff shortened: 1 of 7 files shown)


sqlglot/parser.py
```diff
@@ -5517,16 +5518,18 @@ def _parse_id_var(
...
-
-        if (any_token and self._advance_any()) or self._match_set(tokens or self.ID_VAR_TOKENS):
+        expression = self._parse_identifier()
+        if not expression and (
+            (any_token and self._advance_any()) or self._match_set(tokens or self.ID_VAR_TOKENS)
+        ):
             quoted = self._prev.token_type == TokenType.STRING
-            return exp.Identifier(this=self._prev.text, quoted=quoted)
+            expression = exp.Identifier(this=self._prev.text, quoted=quoted)
+            if comment:
```
tobymao, line 5529: why is this needed? why can't we just always add comments?

tobymao: self.expression(exp.Identifier, ...)

georgesittas: Because the comments associated with identifiers are actually attached in columns, due to the `self.expression(exp.Column, ...)` call, which in turn calls `self._add_comments`.

sqlglot/parser.py
```diff
@@ -5517,16 +5518,18 @@ def _parse_id_var(
...
+                self._add_comments(expression)
```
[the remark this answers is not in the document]
georgesittas: It's handled automatically because `_parse_identifier` constructs an expression using `self.expression`, which in turn takes care of attaching comments:

```python
return self.expression(exp.Identifier, this=self._prev.text, quoted=True)
```

sqlglot/parser.py
georgesittas: So if you prefer moving comments at the identifier level (which is a breaking change), then your suggestion would work fine. Perhaps we could route `Column.comments` to `Column.this.comments` or something.

georgesittas: FWIW the behavior is now inconsistent because delimited identifiers do get comments attached to them, as I explained in https://github.com/tobymao/sqlglot/pull/3340#discussion_r1575032887. So perhaps we should either change that or make sure we also do it for non-delimited ones.

sqlglot/parser.py
```diff
@@ -5503,13 +5506,13 @@ def _parse_alias(
         )
 
         if alias:
+            comments += alias.pop_comments()
```
tobymao, line 5509: why += instead of extends?

tests/test_parser.py
```diff
@@ -357,7 +357,7 @@ def test_comments_insert(self):
         )
 
         self.assertEqual(expression.comments, ["comment1.1", "comment1.2", "comment1.3"])
-        self.assertEqual(expression.this.comments, ["comment2"])
+        self.assertEqual(expression.this.this.comments, ["comment2"])
```
tobymao, line 360: why did this change? don't we push the comments up?

georgesittas: Because this corresponds to a `Table`. May need to push the comments to tables as well.

georgesittas: [done](https://github.com/tobymao/sqlglot/pull/3340/commits/47a57456d665a2ad65f5a80b0575f000abfcec43)

tobymao (approved):

PR #3343, opened 2024-04-23
Feat(prql): add support for AGGREGATE

fool1280:
1. `AGGREGATE`: https://prql-lang.org/book/reference/stdlib/transforms/aggregate.html
2. Add support for `this` keyword: https://prql-lang.org/book/reference/syntax/keywords.html

### DIFF
--- a/sqlglot/dialects/prql.py
+++ b/sqlglot/dialects/prql.py
@@ -60,11 +60,13 @@ class PRQL(Dialect):
             ),
         }
 
-        FUNCTIONS = {
-            "AVERAGE": exp.Avg.from_arg_list,
-            "MIN": exp.Min.from_arg_list,
-            "MAX": exp.Max.from_arg_list,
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/prql.py
```diff
@@ -53,6 +55,18 @@ class Parser(parser.Parser):
                 _select_all(self._parse_table()), distinct=False, copy=False
             ),
             "SORT": lambda self, query: self._parse_order_by(query),
+            "AGGREGATE": lambda self, query: self._parse_selection(
+                query, self._parse_aggregate, append=False
+            ),
+        }
+
+        AGG_FUNCTIONS = {
+            "MIN": lambda arg: exp.Min(this=arg),
```
tobymao, line 64: i'm not sure this is right, we shouldn't have to hardcode all of these, can we programatically do this better? like just parse all the functions without paranethesis and map that to the func?

PR #3347, opened 2024-04-24
feat(clickhouse): Parse window functions in ParameterizedAggFuncs

### LINKED ISSUE #3344: ParseError when handling valid ClickHouse SQL with quantileGK and window functions
### Description
When attempting to parse a valid ClickHouse SQL statement that uses the `quantileGK` function in conjunction with window functions, `sqlglot` raises a `ParseError`. This issue seems related to the parsing of nested function calls or specific handling of window specifications in conjunction with functions like `quantileGK`.

### Steps to Reproduce
1. Use the following SQL query intended for ClickHouse:
   ```sql
   select
       id,
       quantileGK(100, 0.95)(reading) over (partition by id order by toUnixTimestamp(ts) range between 30000 preceding and current row) as quant
   from temp_measurements
   ```
2. Attempt to parse the SQL using `sqlglot.parse_one(offending_sql)`

### Expected Behavior
The SQL should be parsed successfully without errors, as it is a valid query in ClickHouse.

### Actual Behavior
`sqlglot` fails to parse the SQL and throws the following error:
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 4, Col: 36.
```

Official documentation for [quantileGK](https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/quantileGK) function.

VaggelisD:
Fixes #3344

Clickhouse defines it's own `_parse_function()` which reuses (and extends) `parser.py::_parse_function()`. The latter will also parse the following window function provided that the next tokens are matched e.g. `aggfunc(...) OVER ...`.

However, for CH's parameterized aggregation functions, the nested params will be _between_ the function call and the rest of the window spec, which will cause `parser.py::_parse_function()` to miss the window specification as it can't match the appropriate window tokens right after the function call. For this case, CH must retry to parse it _after_ having parsed the params.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -254,6 +254,14 @@ class ClickHouse(Dialect):
             "sparkBar",
             "sumCount",
             "largestTriangleThreeBuckets",
+            "histogram",
+            "sequenceMatch",
+            "sequenceCount",
+            "windowFunnel",
+            "retention",
+            "uniqUpTo",
+            "sequenceNextNode",
+            "exponentialTimeDecayedAvg",
         }
 
         AGG_FUNCTIONS_SUFFIXES = [
@@ -472,16 +480,17 @@ class ClickHouse(Dialect):
 
             func = expr.this if isinstance(expr, exp.Window) else expr
 
-            if isinstance(func, exp.Anonymous):
-                parts = self.AGG_FUNC_MAPPING.get(func.this)
-                is_combined = parts and parts[1]
+            # Aggregate functions can be split in <aggfunc name><suffix>
+            parts = self.AGG_FUNC_MAPPING.get(func.this) if func else None
+
+            if isinstance(func, exp.Anonymous) and parts:
                 params = self._parse_func_params(func)
 
                 kwargs = {
                     "this": func.this,
                     "expressions": func.expressions,
                 }
-                if is_combined:
+                if parts[1]:
                     kwargs["parts"] = parts
                     exp_class = exp.CombinedParameterizedAgg if params else exp.CombinedAggFunc
                 else:
@@ -501,7 +510,7 @@ class ClickHouse(Dialect):
                     # Params have blocked super()._parse_function() from parsing the following window
                     # (if that exists) as they're standing between the function call and the window spec
                     expr = self._parse_window(func)
-                elif parts:
+                else:
                     expr = func
 
             return expr
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -463,48 +463,51 @@ class ClickHouse(Dialect):
             optional_parens: bool = True,
             any_token: bool = False,
         ) -> t.Optional[exp.Expression]:
-            func = super()._parse_function(
+            expr = super()._parse_function(
                 functions=functions,
                 anonymous=anonymous,
                 optional_parens=optional_parens,
                 any_token=any_token,
             )
 
+            func = expr.this if isinstance(expr, exp.Window) else expr
+
             if isinstance(func, exp.Anonymous):
                 parts = self.AGG_FUNC_MAPPING.get(func.this)
+                is_combined = parts and parts[1]
                 params = self._parse_func_params(func)
 
-                if params:
-                    if parts and parts[1]:
-                        func = self.expression(
-                            exp.CombinedParameterizedAgg,
-                            this=func.this,
-                            expressions=func.expressions,
-                            params=params,
-                            parts=parts,
-                        )
-                    func = self.expression(
-                        exp.ParameterizedAgg,
-                        this=func.this,
-                        expressions=func.expressions,
-                        params=params,
-                    )
-                    return self._parse_window(func)
-                if parts:
-                    if parts[1]:
-                        return self.expression(
-                            exp.CombinedAggFunc,
-                            this=func.this,
-                            expressions=func.expressions,
-                            parts=parts,
-                        )
-                    return self.expression(
-                        exp.AnonymousAggFunc,
-                        this=func.this,
-                        expressions=func.expressions,
-                    )
+                if is_combined:
+                    exp_class = exp.CombinedParameterizedAgg if params else exp.CombinedAggFunc
+                else:
+                    exp_class = exp.ParameterizedAgg if params else exp.AnonymousAggFunc
 
-            return func
+                kwargs = {
+                    "exp_class": exp_class,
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -476,20 +476,20 @@ def _parse_function(
...
                         exp.ParameterizedAgg,
                         this=func.this,
                         expressions=func.expressions,
                         params=params,
                     )
-
+                    return self._parse_window(func)
                 if parts:
                     if parts[1]:
                         return self.expression(
```
georgesittas, line 495: What about this and L501? Shouldn't we call `_parse_window` once at the end of this method?

VaggelisD: I was thinking about this but I think this will be a no-op for the case of `parts`, which only serves the purpose of  splitting the aggfunc name from it's sufix. If we don't have `params`, then there is nothing standing between the function call and the window spec. Thus, this can be fully parsed by `parser.py::_parse_function()`. 

Also, this whole branch is only accessed if the `parser.py__parse_function()` call has returned an anonymous function, so if a window was parsed too we will never reach this point. Sidenote: If i'm not missing something, that means that another side-effect was found; CH can't tag `exp.Anonymous` functions as `exp.CombinedParameterizedAgg / exp.ParameterizedAgg` if they're parsed and wrapped in a `exp.Window`, maybe we can check `window.this` and substitute it after the fact.

georgesittas: > I was thinking about this but I think this will be a no-op

Sounds good, I think I see what you mean.

> CH can't tag exp.Anonymous functions as `exp.CombinedAggFunc` / `exp.AnonymousAggFunc` if they're parsed and wrapped in a `exp.Window`

Good catch, can we fix this?

VaggelisD: There started being too much noise so I tried to refactor things and hopefully make some conditions clearer to see.

Is this version preferred or should I merge the sidenote fix with the old code?

sqlglot/dialects/clickhouse.py
```diff
@@ -463,48 +463,51 @@ def _parse_function(
...
                 anonymous=anonymous,
                 optional_parens=optional_parens,
                 any_token=any_token,
             )
 
+            func = expr.this if isinstance(expr, exp.Window) else expr
+
             if isinstance(func, exp.Anonymous):
                 parts = self.AGG_FUNC_MAPPING.get(func.this)
+                is_combined = parts and parts[1]
```
georgesittas, line 477: This seems hacky, we shouldn't be checking `parts[1]` as a proxy for >= 2 items. Let's be explicit by using `len`.

georgesittas: Nvm ignore me, I forgot `parts` is a pair.

sqlglot/dialects/clickhouse.py
```diff
@@ -463,48 +463,51 @@ def _parse_function(
...
+                    exp_class = exp.ParameterizedAgg if params else exp.AnonymousAggFunc
+
+                kwargs = {
+                    "exp_class": exp_class,
+                    "this": func.this,
+                    "expressions": func.expressions,
+                }
+
+                if is_combined:
+                    kwargs["parts"] = parts
```
georgesittas, line 492: Nit: I think I'd rearrange these a bit to avoid the redundant `is_combined` check:

```suggestion
                kwargs = {
                    "this": func.this,
                    "expressions": func.expressions,
                }
                if is_combined:
                    kwargs["parts"] = parts
                    exp_class = exp.CombinedParameterizedAgg if params else exp.CombinedAggFunc
                else:
                    exp_class = exp.ParameterizedAgg if params else exp.AnonymousAggFunc

                kwargs["exp_class"] = exp_class
```

sqlglot/dialects/clickhouse.py
```diff
@@ -463,48 +463,51 @@ def _parse_function(
...
+                    kwargs["params"] = params
+
+                func = self.expression(**kwargs)
+
+                # Params have blocked super()._parse_function() from parsing the following window
+                # (if that exists) as they're standing between the function call and the window spec
+                if params:
+                    func = self._parse_window(func)
+
+                # The window's func was parsed as Anonymous in base parser, fix it's
```
georgesittas, line 503: Let's test this bit

sqlglot/dialects/clickhouse.py
```diff
@@ -463,48 +463,51 @@ def _parse_function(
...
+                # type to be CH style CombinedAnonymousAggFunc / AnonymousAggFunc
+                if isinstance(expr, exp.Window):
+                    expr.set("this", func)
+                elif parts or params:
```
georgesittas, line 507: we can probably connect this branch with L500 because `if params` is mutually exclusive with `isinstance(expr, exp.Window)`, right?

georgesittas (commented):
Ok looks good, a few more comments from me

sqlglot/dialects/clickhouse.py
VaggelisD: That's right, if `params` is set then we shouldn't expect to have already parsed a window beforehand

tests/dialects/test_clickhouse.py
```diff
@@ -826,3 +829,28 @@ def test_ddl(self):
         self.validate_identity(
             "CREATE TABLE t1 (a String EPHEMERAL, b String EPHEMERAL func(), c String MATERIALIZED func(), d String ALIAS func()) ENGINE=TinyLog()"
         )
+
+    def test_agg_functions(self):
+        def extract_agg_func(query):
+            return parse_one(query, read="clickhouse").args.get("expressions")[0].this
```
georgesittas, line 835:
```suggestion
            return parse_one(query, read="clickhouse").selects[0].this
```

georgesittas (approved):
chatted about simplifying the `isinstance(expr, Window)` block in slack, lgtm otherwise

sqlglot/dialects/clickhouse.py
```diff
@@ -463,48 +471,51 @@ def _parse_function(
...
-                        expressions=func.expressions,
-                    )
-
-            return func
+                    kwargs["params"] = params
+
+                func = self.expression(**kwargs)
+
+                if isinstance(expr, exp.Window):
+                    # The window's func was parsed as Anonymous in base parser, fix it's
```
georgesittas, line 508:
```suggestion
                    # The window's func was parsed as Anonymous in base parser, fix its
```

georgesittas (approved):

PR #3348, opened 2024-04-24
fix(tsql): Convert TIMESTAMP to ROWVERSION, transpile both to BINARY

### LINKED ISSUE #3345: SQL Server Rowversion/Timestamp not parsed correctly
In MS SQL Server, there is a funny type called `rowversion` and has an alias `timestamp`. It has nothing in common with timestamp types in other databases. It's basically an `binary(8)` type which stores an incrementing number

https://learn.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver16 

Sqlglot does not seem to recognize this type. It would be nice if sqlglot would transpile it correctly to binary types. 

```python

import sqlglot.expressions as ex
import sqlglot

print(  # transpiled type timestamp as is, should be binary
    sqlglot.parse_one("create table a (b timestamp)", dialect="tsql").sql("databricks")
)

print(  # prints ROWVERSION instead of BINARY
    sqlglot.parse_one("create table a (b rowversion)", dialect="tsql").sql("databricks")
)

print(  # good, this works! databricks timestamp is datetime2 in ms sql
    sqlglot.parse_one("create table a (b timestamp)", dialect="databricks").sql("tsql")
)


```

VaggelisD:
Fixes #3345

In T-SQL:
- Parse `"TIMESTAMP"` into `TokenType.ROWVERSION` as the former is deprecated in favor of the latter
- Explicitly preserve the roundtrip generation of `DataType.ROWVERSION -> "ROWVERSION"`

In all other dialects:
- Convert `DataType.ROWVERSION` to `TokenType.BINARY` by default

[T-SQL Rowversion](https://learn.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver16)

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -464,6 +464,7 @@ class TSQL(Dialect):
             "SMALLMONEY": TokenType.SMALLMONEY,
             "SQL_VARIANT": TokenType.VARIANT,
             "TOP": TokenType.TOP,
+            "TIMESTAMP": TokenType.ROWVERSION,
             "UNIQUEIDENTIFIER": TokenType.UNIQUEIDENTIFIER,
             "UPDATE STATISTICS": TokenType.COMMAND,
             "XML": TokenType.XML,
@@ -755,6 +756,7 @@ class TSQL(Dialect):
             exp.DataType.Type.TIMESTAMP: "DATETIME2",
             exp.DataType.Type.TIMESTAMPTZ: "DATETIMEOFFSET",
             exp.DataType.Type.VARIANT: "SQL_VARIANT",
+            exp.DataType.Type.ROWVERSION: "ROWVERSION",
         }
 
         TYPE_MAPPING.pop(exp.DataType.Type.NCHAR)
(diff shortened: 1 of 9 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -755,6 +756,7 @@ class Generator(generator.Generator):
             exp.DataType.Type.TIMESTAMP: "DATETIME2",
             exp.DataType.Type.TIMESTAMPTZ: "DATETIMEOFFSET",
             exp.DataType.Type.VARIANT: "SQL_VARIANT",
+            exp.DataType.Type.ROWVERSION: "ROWVERSION",
```
tobymao, line 759: shouldn't we just map rowversion -> binary for tsql, and then have binary map to rowversion? we don't need a rowversion data type

VaggelisD: The docs specify the following:

_A nonnullable rowversion column is semantically equivalent to a binary(8) column. A nullable rowversion column is semantically equivalent to a varbinary(8) column._

Since the semantics are dynamic, I thought that allowing the `ROWVERSION` type would be best (with that in mind, might be safer to generate `Type.ROWVERSION -> VARBINARY` to all other dialects ?)

georgesittas (approved):

sqlglot/dialects/tsql.py
georgesittas: > might be safer to generate `Type.ROWVERSION -> VARBINARY` to all other dialects ?

That makes sense to me.

tobymao (approved):

PR #3352, opened 2024-04-25
fix(duckdb)!: Combine aggregate functions with orderby from WITHIN GROUP

### LINKED ISSUE #3350: snowflake -> duckdb ordered listagg transpilation error
DuckDB doesn't use the `WITHIN GROUP` syntax to order a listagg output. Its instead added into the function call like `listagg(col, sep ORDER BY col2)`

```{python}
snowsql = """
    select
        col0,
        listagg(col1, ', ') within group (order by col2)
    from (values
        (1, 'a', 3),
        (2, 'b', 2),
        (3, 'c', 1)
    )
    group by col0
"""
ducksql = sqlglot.transpile(snowsql, read='snowflake', write='duckdb')[0]
print(ducksql)
```
yields
```{sql}
SELECT col0, GROUP_CONCAT(col1, ', ') WITHIN GROUP (ORDER BY col2) FROM (VALUES (1, 'a', 3), (2, 'b', 2), (3, 'c', 1)) GROUP BY col0
```

which throws `ParserException: Parser Error: Unknown ordered aggregate "GROUP_CONCAT".`

I would expect the transpiler to output this:
```{sql}
select
    col0,
    group_concat(col1, ', ' order by col2)
from (values
    (1, 'a', 3),
    (1, 'b', 2),
    (1, 'c', 1)
)
group by col0
```

**Official Documentation**

https://duckdb.org/docs/sql/aggregates
https://docs.snowflake.com/en/sql-reference/functions/listagg

VaggelisD:
Fixes #3350

This PR aims to merge the aggregate function from `exp.WithinGroup` with the order by clause, as per DuckDB syntax. 

Although DuckDB still allows `PERCENTILE_CONT` & `PERCENTILE_DIST` to be defined in `WITHIN GROUP`, this PR will also convert them to their logically equivalent modes; This has the benefit that `rename_unless_within_group` is not needed anymore.

Docs
-------
- [DuckDB Aggregate Functions](https://duckdb.org/docs/sql/aggregates#ordered-set-aggregate-functions)
- [Snowflake LISTAGG + WITHIN GROUP](https://docs.snowflake.com/en/sql-reference/functions/listagg)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -28,7 +28,7 @@ from sqlglot.dialects.dialect import (
     timestrtotime_sql,
     unit_to_var,
 )
-from sqlglot.helper import flatten, seq_get
+from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
 
 
@@ -153,16 +153,6 @@ def _unix_to_time_sql(self: DuckDB.Generator, expression: exp.UnixToTime) -> str
         return self.func("MAKE_TIMESTAMP", timestamp)
 
     return self.func("TO_TIMESTAMP", exp.Div(this=timestamp, expression=exp.func("POW", 10, scale)))
-
-
-def _rename_unless_within_group(
-    a: str, b: str
-) -> t.Callable[[DuckDB.Generator, exp.Expression], str]:
-    return lambda self, expression: (
-        self.func(a, *flatten(expression.args.values()))
-        if isinstance(expression.find_ancestor(exp.Select, exp.WithinGroup), exp.WithinGroup)
-        else self.func(b, *flatten(expression.args.values()))
-    )
 
 
 class DuckDB(Dialect):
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -617,3 +607,19 @@ def bracket_sql(self, expression: exp.Bracket) -> str:
                     bracket = f"({bracket})[1]"
 
             return bracket
+
+        def withingroup_sql(self, expression: exp.WithinGroup) -> str:
+            expression_sql = self.sql(expression, "expression")
+
+            func = expression.this
+            if isinstance(func, (exp.PercentileCont, exp.PercentileDisc)):
```
georgesittas, line 615: Nit:
```suggestion
            if isinstance(func, transform.PERCENTILES):
```

georgesittas: We could also put this in expressions.py

tests/dialects/test_snowflake.py
```diff
@@ -821,6 +821,13 @@ def test_snowflake(self):
                 "snowflake": "CASE WHEN x = a OR (x IS NULL AND a IS NULL) THEN b WHEN x = c OR (x IS NULL AND c IS NULL) THEN d ELSE e END",
             },
         )
+        self.validate_all(
+            "SELECT LISTAGG(col1, ', ') WITHIN GROUP (ORDER BY col2) FROM t",
+            read={"snowflake": "SELECT LISTAGG(col1, ', ') WITHIN GROUP (ORDER BY col2) FROM t"},
+            write={
+                "duckdb": "SELECT GROUP_CONCAT(col1, ', ' ORDER BY col2) FROM t",
```
georgesittas, line 828: Does the duckdb -> snowflake direction work?

VaggelisD: Doesn't seem like it sadly. Should DuckDB parse ordered agg functions as `exp.WithinGroup` now that the generation is supported, just so it can generalize across other dialects like Snowflake, Postgres etc?

georgesittas: We can leave it for another PR to avoid digressing here.

tests/dialects/test_snowflake.py
```diff
@@ -821,6 +821,13 @@ def test_snowflake(self):
...
+            },
```
georgesittas, line 829:
```suggestion
            write={
                "duckdb": "SELECT GROUP_CONCAT(col1, ', ' ORDER BY col2) FROM t",
                "snowflake": "SELECT LISTAGG(col1, ', ') WITHIN GROUP (ORDER BY col2) FROM t",
            },
```

georgesittas (approved):

PR #3355, opened 2024-04-26
fix: export optimize

### LINKED ISSUE #3354: "optimize" is not exported from module "sqlglot.optimizer"
```
"optimize" is not exported from module "sqlglot.optimizer"
  Import from "sqlglot.optimizer.optimizer" instead
```

for:
```python
```

CleanShot 2024-04-26 at 15 01 41@2x

I think it needs to be added to `__init__.py`

tekumara:
resolves  #3354


tobymao:
https://github.com/tobymao/sqlglot/issues/3354#issuecomment-2078691510

PR #3359, opened 2024-04-26
feat: COPY TO/FROM statement

VaggelisD:
Introduce end-to-end support for `COPY TO` / `COPY FROM` statement

Design notes:

- Dialects parse `COPY` parameters parameters using different separators from each other. Introduced a new dialect flag to control that behavior at parse/generation
- Introduce parsing of external location credentials (Snowflake & Redshift). These might appear in other statements e.g. Redshift's `UNLOAD` but I couldn't find existing support
- Copy parameters are parsed in a best effort catch-all manner, as each dialect specifies a long list. There's a fallback to `exp.Command` if current parsing fails to pick up more nuanced options

Docs
------
- [DuckDB](https://duckdb.org/docs/sql/statements/copy.html)
- [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/copy-into-transact-sql?view=azure-sqldw-latest)
- [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html)
- [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#syntax)
- [Postgres](https://www.postgresql.org/docs/current/sql-copy.html)
- [Databricks](https://docs.databricks.com/en/sql/language-manual/delta-copy-into.html)

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -318,6 +318,9 @@ class Dialect(metaclass=_Dialect):
     UNICODE_START: t.Optional[str] = None
     UNICODE_END: t.Optional[str] = None
 
+    # Separator of COPY statement parameters
+    COPY_PARAMS_SEP: t.Optional[TokenType] = None
+
     @classmethod
     def get_or_raise(cls, dialect: DialectType) -> Dialect:
         """
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -319,7 +319,7 @@ class Dialect(metaclass=_Dialect):
     UNICODE_END: t.Optional[str] = None
 
     # Separator of COPY statement parameters
-    COPY_PARAMS_SEP: t.Optional[TokenType] = None
+    COPY_PARAMS_ARE_CSV: bool = True
 
     @classmethod
     def get_or_raise(cls, dialect: DialectType) -> Dialect:
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -319,7 +319,7 @@ class Dialect(metaclass=_Dialect):
     UNICODE_END: t.Optional[str] = None
 
     # Separator of COPY statement parameters
-    COPY_PARAMS_ARE_CSV: bool = True
+    COPY_PARAMS_ARE_CSV = True
 
     @classmethod
     def get_or_raise(cls, dialect: DialectType) -> Dialect:
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1961,13 +1961,10 @@ class CopyParameter(Expression):
 
 class Copy(Expression):
     arg_types = {
-        "into": False,
         "this": True,
         "kind": True,
         "files": True,
         "credentials": False,
-        "with_token": False,
-        "wrapped": False,
         "format": False,
         "params": False,
     }
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1956,7 +1956,7 @@ class Connect(Expression):
 
 
 class CopyParameter(Expression):
-    arg_types = {"this": True, "value": False}
+    arg_types = {"this": True, "expression": False}
 
 
 class Copy(Expression):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1955,6 +1955,34 @@ class Connect(Expression):
     arg_types = {"start": False, "connect": True, "nocycle": False}
 
 
+class CopyParameter(Expression):
+    arg_types = {"this": True, "value": False}
+
+
+class Copy(Expression):
+    arg_types = {
+        "into": False,
+        "this": True,
+        "kind": True,
+        "files": True,
+        "credentials": False,
+        "with_token": False,
+        "wrapped": False,
+        "format": False,
+        "params": False,
+    }
+
+
+class Credentials(Expression):
+    arg_types = {
+        "credentials": False,
+        "encryption": False,
+        "storage": False,
+        "iam_role": False,
+        "region": False,
+    }
+
+
 class Prior(Expression):
     pass
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3757,3 +3757,53 @@ class Generator(metaclass=_Generator):
         if self.pretty:
             return string.replace("\n", self.SENTINEL_LINE_BREAK)
         return string
+
+    def copyparameter_sql(self, expression: exp.CopyParameter) -> str:
+        option = self.sql(expression, "this")
+        value = self.sql(expression, "value")
+        value = f" {value}" if value else ""
+
+        return f"{option}{value}"
+
+    def copy_sql(self, expression: exp.Copy) -> str:
+        this = self.sql(expression, "this")
+        this = f" INTO {this}" if expression.args.get("into") else f" {this}"
+
+        credentials = self.sql(expression, "credentials")
+        credentials = f" {credentials}" if credentials else ""
+
+        kind = " FROM " if expression.args.get("kind") else " TO "
+        files = self.expressions(expression, "files", flat=True)
+
+        sep = ", " if self.dialect.COPY_PARAMS_SEP else " "
+        params = self.expressions(expression, "params", flat=True, sep=sep)
+        if params:
+            params = self.wrap(params) if expression.args.get("wrapped") else params
+            params = f" WITH {params}" if expression.args.get("with_token") else f" {params}"
+
+        return f"COPY{this}{kind}{files}{credentials}{params}"
+
+    def credentials_sql(self, expression: exp.Credentials) -> str:
+        cred_expr = expression.args.get("credentials")
+        if isinstance(cred_expr, exp.Literal):
+            # Redshift case: CREDENTIALS <string>
+            credentials = self.sql(expression, "credentials")
+            credentials = f"CREDENTIALS {credentials}"
+        else:
+            # Snowflake case: CREDENTIALS = (...)
+            credentials = self.expressions(expression, "credentials", flat=True, sep=" ")
+            credentials = f"CREDENTIALS = ({credentials})" if credentials else ""
+
+        storage = self.sql(expression, "storage")
+        storage = f" {storage}" if storage else ""
+
+        encryption = self.expressions(expression, "encryption", flat=True, sep=" ")
+        encryption = f"ENCRYPTION = ({encryption})" if encryption else ""
+
+        iam_role = self.sql(expression, "iam_role")
+        iam_role = f"IAM_ROLE {iam_role}" if iam_role else ""
+
+        region = self.sql(expression, "region")
+        region = f" REGION {region}" if region else ""
+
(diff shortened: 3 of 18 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -318,6 +318,9 @@ class Dialect(metaclass=_Dialect):
     UNICODE_START: t.Optional[str] = None
     UNICODE_END: t.Optional[str] = None
 
+    # Separator of COPY statement parameters
+    COPY_PARAMS_SEP: t.Optional[TokenType] = None
```
georgesittas, line 322: We should make this be `COMMA` by default

sqlglot/tokens.py
```diff
@@ -647,6 +648,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "COMMIT": TokenType.COMMIT,
         "CONNECT BY": TokenType.CONNECT_BY,
         "CONSTRAINT": TokenType.CONSTRAINT,
+        "COPY": TokenType.COPY,
```
georgesittas, line 651: We need to add this to `ID_VAR_TOKENS`, it's a valid identifier in eg. ddb:

```
D with t as (select 1 as copy) select copy from t;
┌───────┐
│ copy  │
│ int32 │
├───────┤
│     1 │
└───────┘
```

sqlglot/generator.py
```diff
@@ -3757,3 +3757,53 @@ def _replace_line_breaks(self, string: str) -> str:
...
+
+    def copy_sql(self, expression: exp.Copy) -> str:
+        this = self.sql(expression, "this")
+        this = f" INTO {this}" if expression.args.get("into") else f" {this}"
+
+        credentials = self.sql(expression, "credentials")
+        credentials = f" {credentials}" if credentials else ""
+
+        kind = " FROM " if expression.args.get("kind") else " TO "
+        files = self.expressions(expression, "files", flat=True)
```
georgesittas, line 3776:
```suggestion
        files = self.expressions(expression, key="files", flat=True)
```

sqlglot/generator.py
```diff
@@ -3757,3 +3757,53 @@ def _replace_line_breaks(self, string: str) -> str:
...
+
+        sep = ", " if self.dialect.COPY_PARAMS_SEP else " "
+        params = self.expressions(expression, "params", flat=True, sep=sep)
```
georgesittas, line 3779:
```suggestion
        params = self.expressions(expression, key="params", flat=True, sep=sep)
```

sqlglot/generator.py
```diff
@@ -3757,3 +3757,53 @@ def _replace_line_breaks(self, string: str) -> str:
...
+        this = self.sql(expression, "this")
+        this = f" INTO {this}" if expression.args.get("into") else f" {this}"
+
+        credentials = self.sql(expression, "credentials")
+        credentials = f" {credentials}" if credentials else ""
+
+        kind = " FROM " if expression.args.get("kind") else " TO "
+        files = self.expressions(expression, "files", flat=True)
+
+        sep = ", " if self.dialect.COPY_PARAMS_SEP else " "
```
georgesittas, line 3778:
```suggestion
        sep = ", " if self.dialect.COPY_PARAMS_SEP == TokenType.COMMA else " "
```

sqlglot/dialects/dialect.py
georgesittas: Actually, if the only separators are comma and space, we should rename this to something like `COPY_PARAMS_ARE_CSV`

sqlglot/parser.py
```diff
@@ -6292,3 +6298,94 @@ def _parse_with_operator(self) -> t.Optional[exp.Expression]:
...
+        this = (
+            self._parse_conjunction()
+            if self._match(TokenType.L_PAREN, advance=False)
+            else self._parse_table(schema=True)
+        )
+
+        if self._match(TokenType.FROM):
+            kind = True
+        elif self._match_text_seq("TO"):
+            kind = False
```
georgesittas, line 6367: We need to set a default value to `kind`, we'll get a `NameError` instead of a `ParseError`

sqlglot/parser.py
```diff
@@ -6292,3 +6298,94 @@ def _parse_with_operator(self) -> t.Optional[exp.Expression]:
...
+        return expr
+
+    def _parse_copy(self):
+        start = self._prev
+
+        into = self._match_text_seq("INTO")
+
+        this = (
+            self._parse_conjunction()
+            if self._match(TokenType.L_PAREN, advance=False)
```
georgesittas, line 6360: It seems like we don't have test coverage for the `INTO ( ...` path?

sqlglot/parser.py
```diff
@@ -6292,3 +6298,94 @@ def _parse_with_operator(self) -> t.Optional[exp.Expression]:
...
+        if self._match(TokenType.FROM):
+            kind = True
+        elif self._match_text_seq("TO"):
+            kind = False
+
+        files = self._parse_csv(self._parse_conjunction)
+        credentials = self._parse_credentials()
+
+        with_token = self._match_text_seq("WITH")
+        wrapped = self._match(TokenType.L_PAREN, advance=False)
```
georgesittas, line 6373: I'm not sure if `wrapped` should be part of the AST. Isn't this something we're better off controlling at generation time? Is there a dialect that supports _both_ the wrapped and the unwrapped form?

sqlglot/generator.py
```diff
@@ -3757,3 +3757,53 @@ def _replace_line_breaks(self, string: str) -> str:
...
+    def credentials_sql(self, expression: exp.Credentials) -> str:
+        cred_expr = expression.args.get("credentials")
+        if isinstance(cred_expr, exp.Literal):
+            # Redshift case: CREDENTIALS <string>
+            credentials = self.sql(expression, "credentials")
+            credentials = f"CREDENTIALS {credentials}"
+        else:
+            # Snowflake case: CREDENTIALS = (...)
+            credentials = self.expressions(expression, "credentials", flat=True, sep=" ")
+            credentials = f"CREDENTIALS = ({credentials})" if credentials else ""
```
georgesittas, line 3795: Should we factor this out into a helper method? Eg. `generated_credentials`, which generates the "snowflake" case by default and then can be overridden in Redshift. Is Redshift the only one with this syntax?

sqlglot/parser.py
VaggelisD: I was hesitant to add more flags for these details, such as `WRAPPED`. There isn't a dialect that supports both at the same time as far as I've seen, but e.g. Snowflake doesn't wrap it's options while others do

sqlglot/generator.py
VaggelisD: Both dialects have unique syntax, so I thought that prefering one over the other might not be semantically correct and that there's not much to gain from that extra fragmentation. However, either approach works for me

sqlglot/parser.py
georgesittas: it's fine to add these flags, we do it for several things already and it's better vs adding complexity to the AST imo

sqlglot/parser.py
```diff
@@ -6292,3 +6298,94 @@ def _parse_with_operator(self) -> t.Optional[exp.Expression]:
...
+            value = None
+            # Some options are defined as functions with the values as params
+            if not isinstance(option, exp.Func):
+                # Different dialects might separate options and values by white space, "=" and "AS"
+                eq = self._match(TokenType.EQ)
+                self._match(TokenType.ALIAS)
+                value = self._parse_unquoted_field()
+                if eq:
+                    option = exp.EQ(this=option, expression=value)
+                    value = None
```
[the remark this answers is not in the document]
georgesittas: Yeah I think flags are generally preferable over AST complexity, because the latter is also user-facing

sqlglot/parser.py
VaggelisD: Checkout `test_duckdb` and `test_postgres`, there's some `COPY <query> TO <file>` tests stressing this path. For other dialects that don't support `COPY TO`, it shouldn't be legal to specify a subquery as a table

georgesittas: Ah got it, I was expecting `INTO` so the keyword search didn't show any hits.

VaggelisD: `INTO` is another token that differs between dialects just like `WITH` i.e some dialects don't have it

sqlglot/dialects/dialect.py
```diff
@@ -318,6 +318,9 @@ class Dialect(metaclass=_Dialect):
     UNICODE_START: t.Optional[str] = None
     UNICODE_END: t.Optional[str] = None
 
+    # Separator of COPY statement parameters
+    COPY_PARAMS_ARE_CSV: bool = True
```
georgesittas, line 322:
```suggestion
    COPY_PARAMS_ARE_CSV = True
```

georgesittas (commented):
Thanks for addressing the comments 👍

georgesittas (approved):
Haven't looked closely into each syntax spec, but lgtm. I'll wait for Toby to take a look as well since this is relatively big.

sqlglot/expressions.py
```diff
@@ -1955,6 +1955,31 @@ class Connect(Expression):
     arg_types = {"start": False, "connect": True, "nocycle": False}
 
 
+class CopyParameter(Expression):
+    arg_types = {"this": True, "value": False}
```
tobymao, line 1959: should value just be "expression"

sqlglot/parser.py
```diff
@@ -6292,3 +6298,84 @@ def _parse_with_operator(self) -> t.Optional[exp.Expression]:
         op = self._parse_var(any_token=True)
 
         return self.expression(exp.WithOperator, this=this, op=op)
+
+    def _parse_copy_parameters(self) -> t.List[exp.CopyParameter]:
+        sep = TokenType.COMMA if self.dialect.COPY_PARAMS_ARE_CSV else None
```
tobymao, line 6303: i don't think you need to specify dialect here, doesn't it get passed down to the class?

tobymao: the usual pattern is to do self.COPY_PARAMS_ARE_CSV

georgesittas: IIRC the reason we put this in dialect is because it's used in both the parser and generator

tobymao: do we do that anywhere else? we share other things in both places right

georgesittas: yep eg. we have `self.dialect.ALIAS_POST_TABLESAMPLE` both in the parser and the generator

georgesittas: these dialect-wide constants are not propagated intentionally because we'd need to duplicate constant definitions in the generator, parser and tokenizer

tobymao (approved):

sqlglot/parser.py
georgesittas: eg. if you make the above `self.ALIAS_POST_TABLESAMPLE` you'll get several failures:

```
    AttributeError: 'Parser' object has no attribute 'ALIAS_POST_TABLESAMPLE'
```

tobymao:
one comment to rename value -> expression for consistency

PR #3367, opened 2024-04-29
feat(mysql): Transpile TimestampTrunc

### LINKED ISSUE #3366: Timestamp trunc method issue from postgres to MySQL
I was trying to convert an SQL query from Postgres to MySQL. Check the following code snippet

```
import sqlglot
sqlglot.transpile("SELECT date_trunc('hour', timestamp '2001-02-16 20:38:40') FROM dual", read="postgres", write="mysql") 
```

This returns

`["SELECT TIMESTAMP_TRUNC(CAST('2001-02-16 20:38:40' AS DATETIME), HOUR) FROM dual"]`

But MySQL doesn't have TIMESTAMP_TRUNC method. Please look into this.

VaggelisD:
Fixes #3366

MySQL cannot generate `exp.TimestampTrunc` using an existing function; However, one can simulate it with a combination of date add & diff ([source](https://stackoverflow.com/a/32955740))

Docs for `DATE_TRUNC`
-----------
- [Postgres](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC)
- [Presto](https://prestodb.io/docs/current/functions/datetime.html#date_trunc) 
- [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/date_trunc)
- [Databricks / Spark](https://docs.databricks.com/en/sql/language-manual/functions/date_trunc.html)

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -869,7 +869,6 @@ class MySQL(Dialect):
             return f"CHAR({this}{using})"
 
         def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
-            this = self.sql(expression, "this")
             unit = expression.args.get("unit")
 
             # Pick an old-enough date to avoid negative timestamp diffs
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -876,7 +876,7 @@ class MySQL(Dialect):
             start_ts = "'0000-01-01 00:00:00'"
 
             # Source: https://stackoverflow.com/a/32955740
-            timestamp_diff = build_date_delta(exp.TimestampDiff)([unit, start_ts, this])
+            timestamp_diff = build_date_delta(exp.TimestampDiff)([unit, start_ts, expression.this])
             interval = exp.Interval(this=timestamp_diff, unit=unit)
             dateadd = build_date_delta_with_interval(exp.DateAdd)([start_ts, interval])
 
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -867,3 +867,29 @@ class MySQL(Dialect):
             charset = expression.args.get("charset")
             using = f" USING {self.sql(charset)}" if charset else ""
             return f"CHAR({this}{using})"
+
+        def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
+            this = self.sql(expression, "this")
+            unit = expression.args.get("unit")
+            unix_timestart = "'1900-01-01 00:00:00'"
+
+            mysql_unit = unit
+            if isinstance(unit, exp.Var):
+                # Match abbrevations, most coming from Snowflake
+                unit = unit.this.upper()
+                if unit.startswith("Y"):
+                    mysql_unit = "YEAR"
+                elif unit == "MM" or unit.startswith("MON"):
+                    mysql_unit = "MONTH"
+                elif unit.startswith("W"):
+                    mysql_unit = "WEEK"
+                elif unit.startswith("D"):
+                    mysql_unit = "DAY"
+                elif unit.startswith("MIN"):
+                    mysql_unit = "MINUTE"
+                elif unit.startswith("SEC"):
+                    mysql_unit = "SECOND"
+                elif unit.startswith("MS"):
+                    mysql_unit = "MILLISECOND"
+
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -867,3 +867,29 @@ def chr_sql(self, expression: exp.Chr) -> str:
...
             return f"CHAR({this}{using})"
+
+        def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
+            this = self.sql(expression, "this")
+            unit = expression.args.get("unit")
+            unix_timestart = "'1900-01-01 00:00:00'"
+
+            mysql_unit = unit
+            if isinstance(unit, exp.Var):
+                # Match abbrevations, most coming from Snowflake
```
georgesittas, line 878: This shouldn't be needed, abbreviated units are [canonicalized](https://github.com/tobymao/sqlglot/blob/e82a30b6563547daea0bb087e1b6b5bf3b0532d3/sqlglot/dialects/snowflake.py#L144) at parse time for Snowflake. If for some reason they weren't in what you've tested, then we should fix that instead.

sqlglot/dialects/mysql.py
```diff
@@ -867,3 +867,29 @@ def chr_sql(self, expression: exp.Chr) -> str:
...
+                elif unit.startswith("D"):
+                    mysql_unit = "DAY"
+                elif unit.startswith("MIN"):
+                    mysql_unit = "MINUTE"
+                elif unit.startswith("SEC"):
+                    mysql_unit = "SECOND"
+                elif unit.startswith("MS"):
+                    mysql_unit = "MILLISECOND"
+
+            return f"DATE_ADD({unix_timestart}, interval TIMESTAMPDIFF({mysql_unit}, {unix_timestart}, {this}) {mysql_unit})"
```
georgesittas, line 895: Let's generate this programmatically by leveraging `exp` builders and `self.func`

sqlglot/dialects/mysql.py
```diff
@@ -867,3 +867,29 @@ def chr_sql(self, expression: exp.Chr) -> str:
             charset = expression.args.get("charset")
             using = f" USING {self.sql(charset)}" if charset else ""
             return f"CHAR({this}{using})"
+
+        def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
+            this = self.sql(expression, "this")
+            unit = expression.args.get("unit")
+            unix_timestart = "'1900-01-01 00:00:00'"
```
georgesittas, line 874:
```suggestion
            unix_timestart = exp.Literal.string('1900-01-01 00:00:00')
```

georgesittas: We should comment that this value was arbitrarily picked as an "old-enough date" (AFAICT), so as to avoid getting back a negative number from the `TIMESTAMPDIFF`.

Let's also rename this to something better while we're at it - `unix_timestart` is a little confusing given that this isn't 1970-01-01. I'd probably do `start_ts` & add this comment to clear any confusion.

georgesittas (approved):

tobymao (approved):

sqlglot/dialects/mysql.py
```diff
@@ -867,3 +867,17 @@ def chr_sql(self, expression: exp.Chr) -> str:
...
+
+        def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
+            this = self.sql(expression, "this")
+            unit = expression.args.get("unit")
+
+            # Pick an old-enough date to avoid negative timestamp diffs
+            start_ts = "'0000-01-01 00:00:00'"
+
+            # Source: https://stackoverflow.com/a/32955740
+            timestamp_diff = build_date_delta(exp.TimestampDiff)([unit, start_ts, this])
```
georgesittas, line 879:
```suggestion
            timestamp_diff = build_date_delta(exp.TimestampDiff)([unit, start_ts, expression.this])
```

sqlglot/dialects/mysql.py
```diff
@@ -867,3 +867,17 @@ def chr_sql(self, expression: exp.Chr) -> str:
             charset = expression.args.get("charset")
             using = f" USING {self.sql(charset)}" if charset else ""
             return f"CHAR({this}{using})"
+
+        def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
+            this = self.sql(expression, "this")
```
georgesittas, line 872:
```suggestion
```

georgesittas (commented):
Small cleanup

PR #3380, opened 2024-04-30
fix(optimizer): Remove XOR from connector simplifications

### LINKED ISSUE #3372: XOR transformed into an OR depending on the alphabetical order of operands
Hello,

I stumbled upon an apparent bug still present in the 23.12.1 release (I was using 20.9.0 when I first saw the bug). When using the `optimize` function on a MySQL constraint, result depends on the alphabetical order of the operands (with one of the results being wrong):

```python
from sqlglot import parse_one

# Returns '`bar` IS NULL XOR `foo` IS NULL'
optimize(parse_one('(bar IS NULL) XOR (foo IS NULL)', read="mysql")).sql("mysql")

# Returns '`bar` IS NULL OR `foo` IS NULL'
optimize(parse_one('(foo IS NULL) XOR (bar IS NULL)', read="mysql")).sql("mysql")
```

In the second example, the XOR is transformed into an OR which in this case does not seem to be equivalent. I would expect both results to be the same (or at least not converting the XOR into OR).

Is there something wrong in the way I used those functions or is this a bug ?

I can give more details if needed !
Thank you

VaggelisD:
Fixes #3372

The `simplify` rule aims to simplify `AND` & `OR` connectors by removing duplicate exprs such as `A OR A -> A` but fails to consider boolean `XOR` (supported by MySQL) which has different semantics i.e for `A = True`, `A XOR A = False`


VaggelisD:
Q: Is the first branch of `_simplify_connectors` superfluous since `uniq_sort` will already deduplicate connectors such as `A OR A`?

sqlglot/expressions.py
```diff
@@ -6563,6 +6563,29 @@ def or_(
     return t.cast(Condition, _combine(expressions, Or, dialect, copy=copy, **opts))
 
 
+def xor_(
```
georgesittas, line 6566: The underscore is used eg. in `or_` because `or` is a reserved keyword

```suggestion
def xor(
```

tests/test_optimizer.py
```diff
@@ -427,6 +427,12 @@ def test_simplify(self):
 """.strip(),
         )
 
+        xor_query = parse_one("A XOR D XOR B XOR E XOR F XOR G XOR E XOR A", read="mysql")
+        simplified_xor = optimizer.simplify.simplify(xor_query)
+        self.assertEqual(
+            simplified_xor.sql(dialect="mysql"), "A XOR A XOR B XOR D XOR E XOR E XOR F XOR G"
```
georgesittas, line 433: Can't we simplify this further? Eg. A XOR A is always false.

georgesittas:
> Q: Is the first branch of `_simplify_connectors` superfluous since `uniq_sort` will already deduplicate connectors such as `A OR A`?

Probably, perhaps @tobymao or @barakalon can share context here. You can try and if tests don't fail it _should_ be ok?

VaggelisD:
@georgesittas  Found a new use for the branch in `_simplify_connectors`, as to further optimize `A XOR A -> False` we can extend it to cover that case.

I also went ahead and split the test_optimizer test in 2, one to show the alphabetic sort of `uniq_sort` in XORs and one to showcase the new optimization

georgesittas:
Yeah I think `A XOR A -> FALSE` makes sense

tests/test_optimizer.py
```diff
@@ -427,6 +427,18 @@ def test_simplify(self):
 """.strip(),
         )
 
+        # uniq_sort for XORs
```
barakalon, line 430: These tests should be in the fixture file

barakalon (approved):

sqlglot/expressions.py
```diff
@@ -6563,6 +6563,29 @@ def or_(
     return t.cast(Condition, _combine(expressions, Or, dialect, copy=copy, **opts))
 
 
+def xor(
+    *expressions: t.Optional[ExpOrStr], dialect: DialectType = None, copy: bool = True, **opts
+) -> Condition:
```
tobymao, line 6568: shouldn't this return xor?

tests/test_optimizer.py
VaggelisD: It seems that MySQL is the only dialect that explicitly supports & parses logical XORs, so I decided for now to keep the parsers as is and isolate these tests in `test_optimizer`. If we want to move these to `simplify.sql`, we'd have to move `exp.Xor` in `parser.py`'s `CONJUCTION_PARSERS` afaict

georgesittas: Can't we just annotate the test using `# dialect: mysql`?

VaggelisD: Oh right, that should work too, moving the tests back to the fixture file

sqlglot/expressions.py
VaggelisD: This follows the conventions of `and_()` and `or_()` which return `Condition` with the combined conjuctions as expression afaict

sqlglot/optimizer/simplify.py
```diff
@@ -365,10 +367,17 @@ def uniq_sort(expression, root=True):
...
-        result_func = exp.and_ if isinstance(expression, exp.And) else exp.or_
         flattened = tuple(expression.flatten())
-        deduped = {gen(e): e for e in flattened}
-        arr = tuple(deduped.items())
+
+        if isinstance(expression, exp.Xor):
+            result_func = exp.xor
+            # Do not deduplicate XOR as A XOR A != A if A == True
+            deduped = None
+            arr = tuple([(gen(e), e) for e in flattened])
```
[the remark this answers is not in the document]
VaggelisD: Hmm, looks like by indexing `arr` below this point we disallow generator expressions (not subscriptable)

sqlglot/expressions.py
georgesittas: I think this is tech debt, I added the type hints a long time ago. Can you check if we can make them `Or`, `And` and `Xor`?

VaggelisD: By adjusting both the type hint and the `return t.cast(<type>,  ...)` to the appropriate type I don't see any complaints from style, unit and test. Should I include that refactor here?

georgesittas (approved):

sqlglot/expressions.py
georgesittas: Yea let's do it

georgesittas: Ah, I just remembered why I used `Condition` instead of `Or`.. It's because `or_(1)` works and yields a single literal. So the previous one was good. Apologies for the back and forth @VaggelisD, let's just revert that last change and this should be good to go 👍.

PR #3382, opened 2024-04-30
fix(duckdb): Preserve DATE_SUB roundtrip

### LINKED ISSUE #3373: Syntactically incorrect result produced when using DuckDB `date_sub` as an input
Hei,

when trying to format already existing duckdb sql-ddl code that contains a `date_sub` function, sqlglot will create a sql-ddl string that contains the `interval` keyword which produces a syntax error when executing it in duckdb. To the best of my knowledge, this keyword is not available in duckdb.

Documentation duckdb time functions: https://duckdb.org/docs/sql/functions/time

### Code that causes an error

```python3
import duckdb
import sqlglot

sql_in= """
CREATE TABLE "testtable" (
  "Number"	NUMERIC(5,2) NOT NULL,
  "Name" STRUCT(title VARCHAR, firstname VARCHAR, lastname VARCHAR) NOT NULL,
  "date_to_check"	DATE NOT NULL CHECK(date_sub('year', "date_to_check", '2020-01-01') < 20),
  PRIMARY KEY("Number")
);
"""
output = sqlglot.transpile(sql_in, read="duckdb", write="duckdb", pretty=True, normalize_functions="upper")

duckdb.sql(output[0])
```

### content of the sqlglot output

```sql
"CREATE TABLE "testtable" (
  "Number" DECIMAL(5, 2) NOT NULL,
  "Name" STRUCT(title TEXT, firstname TEXT, lastname TEXT) NOT NULL,
  "date_to_check" DATE NOT NULL CHECK ('year' - INTERVAL ("date_to_check") 2020-01-01 < 20),
  PRIMARY KEY ("Number")
)"
```

### Error Message

```error
  File "/opt/pycharm-2023.3/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
           ^^^^^^
  File "<input>", line 1, in <module>
duckdb.duckdb.ParserException: Parser Error: syntax error at or near "2020"
```

VaggelisD:
Fixes #3373

Current `DATE_SUB` support in DuckDB was limited to a [prior BigQuery -> DDB transpilation case](https://github.com/tobymao/sqlglot/pull/1705), in which BQ's interval-based `DATE_SUB` would convert to an equivalent DDB interval expression. 

However, DuckDB supports `DATE_SUB` (also `DATESUB`) natively, which this PR aims to support for it's roundtrip.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -31,9 +31,6 @@ from sqlglot.dialects.dialect import (
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
 
-if t.TYPE_CHECKING:
-    from sqlglot._typing import E
-
 
 def _ts_or_ds_add_sql(self: DuckDB.Generator, expression: exp.TsOrDsAdd) -> str:
     this = self.sql(expression, "this")
@@ -44,13 +41,6 @@ def _ts_or_ds_add_sql(self: DuckDB.Generator, expression: exp.TsOrDsAdd) -> str:
 def _date_delta_sql(
     self: DuckDB.Generator, expression: exp.DateAdd | exp.DateSub | exp.TimeAdd
 ) -> str:
-    if isinstance(expression, exp.DateSub):
-        subexpr = expression.expression
-        if not isinstance(subexpr, exp.Literal) or not subexpr.is_int:
-            # Preserve the DATE_SUB roundtrip; If subexpr was an integer, we'd
-            # convert to interval as in below (e.g. BigQuery -> DuckDB transpilation)
-            return self.func("DATE_SUB", f"'{expression.unit}'", subexpr, expression.this)
-
     this = self.sql(expression, "this")
     unit = unit_to_var(expression)
     op = "+" if isinstance(expression, (exp.DateAdd, exp.TimeAdd)) else "-"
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -41,6 +44,13 @@ def _ts_or_ds_add_sql(self: DuckDB.Generator, expression: exp.TsOrDsAdd) -> str:
 def _date_delta_sql(
     self: DuckDB.Generator, expression: exp.DateAdd | exp.DateSub | exp.TimeAdd
 ) -> str:
+    if isinstance(expression, exp.DateSub):
+        subexpr = expression.expression
+        if not isinstance(subexpr, exp.Literal) or not subexpr.is_int:
+            # Preserve the DATE_SUB roundtrip; If subexpr was an integer, we'd
+            # convert to interval as in below (e.g. BigQuery -> DuckDB transpilation)
+            return self.func("DATE_SUB", f"'{expression.unit}'", subexpr, expression.this)
```
georgesittas, line 52: Can't we just always generate the datesub syntax?

sqlglot/dialects/duckdb.py
```diff
@@ -77,8 +87,9 @@ def _build_sort_array_desc(args: t.List) -> exp.Expression:
     return exp.SortArray(this=seq_get(args, 0), asc=exp.false())
 
 
-def _build_date_diff(args: t.List) -> exp.Expression:
-    return exp.DateDiff(this=seq_get(args, 2), expression=seq_get(args, 1), unit=seq_get(args, 0))
+# Parse part-based functions, check https://duckdb.org/docs/sql/functions/date.html#date-functions
+def _build_date_part_exp(exp_class: t.Type[E], args: t.List) -> exp.Expression:
+    return (exp_class)(this=seq_get(args, 2), expression=seq_get(args, 1), unit=seq_get(args, 0))
```
georgesittas, line 92: I'd rename this to `_build_date_delta` (the "delta" part is commonly used in sqlglot for these functions). Also, can we make this return a closure for consistent style? So then you'd eg. do this in `TRANSFORMS`:

```python
"DATEDIFF": _build_date_part_exp(exp.DateDiff)
```

sqlglot/dialects/duckdb.py
georgesittas: Discussed in Slack - there's a concern related to `INTERVAL x <unit>`, ie. the expression is a column.

georgesittas (approved):

VaggelisD:
As discussed, decided to make `DATE_SUB` anonymous to implicitly preserve the roundtrip as-is, since all other dialects except DDB use intervals. Properly transpiling `DATE_SUB` will require type inference to differentiate from date and interval exprs

PR #3383, opened 2024-04-30
fix(postgres): Infer FROM clause in parse_substring()

VaggelisD:
Context: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1714464345865119

Postgres supports the following syntax: `SUBSTRING(<string> [from int] [for int])`, meaning that one can supply either `FROM`, `FOR` or both. 

However, our parser will only parse `for` if it has parsed `from` too, which isn't a Postgres constraint as mentioned.

[Postgres SUBSTRING ](https://www.postgresql.org/docs/9.1/functions-string.html)


VaggelisD:
Already [fixed](https://github.com/tobymao/sqlglot/commit/3e8de7124b735a6ab52971a3e51702c4e7b74be5), Jo's too fast

PR #3405, opened 2024-05-03
fix(trino|presto): adding cast support for "hyperloglog" column type

uncledata:


tobymao:
this looks incomplete, can you add a test case following the existing patterns that we have?

uncledata:
hm, somehow not all moved. One sec

sqlglot/tokens.py
```diff
@@ -199,6 +199,7 @@ class TokenType(AutoName):
     AGGREGATEFUNCTION = auto()
     SIMPLEAGGREGATEFUNCTION = auto()
     TDIGEST = auto()
+    HYPERLOGLOG = auto()
```
georgesittas, line 202: I don't think we need a new token, we already have `HLLSKETCH`.

uncledata: just to see if I'm on the right track:
I should remove HyperLogLog mentions I added to "expressions", "parser" and "tokens" files (since HyperLogLog has similar logic/handling as already defined HLLSKETCH), right? I should make use of HLLSketch in "presto" file?

I'm sorry if questions sounds silly, first time trying to contribute..

uncledata: basically add:
            exp.DataType.Type.HLLSKETCH: "HYPERLOGLOG",

uncledata: I've made changes, LMK if that's what you meant!

sqlglot/dialects/presto.py
```diff
@@ -317,6 +317,7 @@ class Generator(generator.Generator):
             exp.DataType.Type.STRUCT: "ROW",
             exp.DataType.Type.DATETIME: "TIMESTAMP",
             exp.DataType.Type.DATETIME64: "TIMESTAMP",
+            exp.DataType.Type.HLLSKETCH: "HYPERLOGLOG",
```
georgesittas, line 320: We also need to map the hyperloglog keyword to the hllsketch token right?

uncledata: ah, basically add: "HYPERLOGLOG": TokenType.HLLSKETCH 
 in the keywords? map, so we're mapping through this. Right?

georgesittas (approved):

uncledata:
Thanks for the support! hopefully next incoming PR's will be smoother.

PR #3422, opened 2024-05-08
Updated tokens.py

gitvibav:
When providing the query with the top command, the parse_one function previously raised an exception. I solved this by updating the tokens.py file and providing a new token type called top. This also solves the issue of converting the query with the top command to another query language, like PostgreSQL, which does not support the top command, and now, after this change, when using transpile, it adds the LIMIT keyword and provides the proper query.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -764,6 +764,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "TEMP": TokenType.TEMPORARY,
         "TEMPORARY": TokenType.TEMPORARY,
         "THEN": TokenType.THEN,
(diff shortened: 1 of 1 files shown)


tobymao:
@gitvibav i don't understand what issue this is trying to solve, you didn't add any test cases, i'm going to close this PR for now since it's relatively low quality.

unclear description
unformatted code
no tests

gitvibav:
Since I am new to open source contributions, I didn't know the guidelines. I am really sorry for that. I will get back to you with an extended description and test cases.

tobymao:
no worries, if there is indeed an issue, i'll help you get a proper fix in.

PR #3426, opened 2024-05-08
feat(snowflake): Support for APPROX_PERCENTILE

### LINKED ISSUE #3424: APPROX_PERCENTILE is translated from Trino to APPROX_QUANTILE in Snowflake, which does not exist in Snowflake
```
>>> sql = "SELECT APPROX_PERCENTILE(a,1, 0.5, 0.001) from t"
>>> select = sqlglot.parse_one(sql, "trino")
>>> select.sql("snowflake")
'SELECT APPROX_QUANTILE(a, 0.5, 0.001, 1) FROM t'
```
It should be translated to `APPROX_PERCENTILE`

**Official Documentation**
https://docs.snowflake.com/en/sql-reference/functions/approx_percentile

VaggelisD:
Fixes #3424 

Introduce Snowflake's `APPROX_PERCENTILE(<expr>, <percentile>)` function & support transpilation with the (existing) Presto/Trino implementation

Docs
------
- [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/approx_percentile)

### DIFF
--- a/tests/dialects/test_snowflake.py
+++ b/tests/dialects/test_snowflake.py
@@ -832,6 +832,18 @@ WHERE
                 "snowflake": "SELECT LISTAGG(col1, ', ') WITHIN GROUP (ORDER BY col2) FROM t",
             },
         )
+        self.validate_all(
+            "SELECT APPROX_PERCENTILE(a, 0.5) FROM t",
+            read={
+                "trino": "SELECT APPROX_PERCENTILE(a, 1, 0.5, 0.001) FROM t",
+                "presto": "SELECT APPROX_PERCENTILE(a, 1, 0.5, 0.001) FROM t",
+            },
+            write={
+                "trino": "SELECT APPROX_PERCENTILE(a, 0.5) FROM t",
+                "presto": "SELECT APPROX_PERCENTILE(a, 0.5) FROM t",
+                "snowflake": "SELECT APPROX_PERCENTILE(a, 0.5) FROM t",
+            },
+        )
 
     def test_null_treatment(self):
         self.validate_all(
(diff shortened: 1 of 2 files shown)


tests/dialects/test_snowflake.py
```diff
@@ -832,6 +832,18 @@ def test_snowflake(self):
                 "snowflake": "SELECT LISTAGG(col1, ', ') WITHIN GROUP (ORDER BY col2) FROM t",
             },
         )
+        self.validate_all(
+            "SELECT APPROX_PERCENTILE(a, 0.5) FROM t",
+            read={
+                "trino": "SELECT APPROX_PERCENTILE(a, 1, 0.5, 0.001) FROM t",
+                "presto": "SELECT APPROX_PERCENTILE(a, 1, 0.5, 0.001) FROM t",
+            },
```
georgesittas, line 840: Can we use `self.unsupported` to note notify the user that `0.5` and `0.001` couldn't be transpiled? I _think_ these correspond to Presto's:

```
approx_percentile(x, w, percentage, accuracy) → [same as x]
```

case, and Trino simply supports it due to inheritance. @baruchoxman can you verify that this is the correct / intended prototype?

georgesittas: Yes that seems correct, that's what I remember too.

baruchoxman: I wonder if it can be possible to override this behavior, and ignore the weight given that the weight is 1 (and thus can be ignored)?
And same for the accuracy parameter - just ignore it when transpiling to Snowflake. 
I guess we can always create our derived Snowflake dialect which overrides the `approxquantile_sql` funciton behavior, but given that this is the format certain BI tools create in their Trino queries, wondering if can have a default

PR #3428, opened 2024-05-08
feat(clickhouse): add support for partition expression

GaliFFun:
ClickHouse partition expression syntax differs from the one used in base Parser, documentation reference: https://clickhouse.com/docs/en/sql-reference/statements/alter/partition#how-to-set-partition-expression

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -583,22 +583,15 @@ class ClickHouse(Dialect):
             if not self._match(TokenType.PARTITION):
                 return None
 
-            # Support for PARTITION ID '201901' syntax
             if self._match_text_seq("ID"):
-                return self.expression(
-                    exp.Partition,
-                    expressions=[
-                        self.expression(
-                            exp.PartitionId,
-                            this=self._parse_string(),
-                        )
-                    ],
-                )
+                # Corresponds to the PARTITION ID <string_value> syntax
+                expressions: t.List[exp.Expression] = [
+                    self.expression(exp.PartitionId, this=self._parse_string())
+                ]
+            else:
+                expressions = self._parse_expressions()
 
-            return self.expression(
-                exp.Partition,
-                expressions=self._parse_expressions(),
-            )
+            return self.expression(exp.Partition, expressions=expressions)
 
     class Generator(generator.Generator):
         QUERY_HINTS = False
@@ -855,4 +848,4 @@ class ClickHouse(Dialect):
             return f"PARTITION {self.expressions(expression, flat=True)}"
 
         def partitionid_sql(self, expression: exp.PartitionId) -> str:
-            return f"ID {expression.this}"
+            return f"ID {self.sql(expression.this)}"
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -578,6 +578,28 @@ class ClickHouse(Dialect):
                 granularity=granularity,
             )
 
+        def _parse_partition(self) -> t.Optional[exp.Partition]:
+            # https://clickhouse.com/docs/en/sql-reference/statements/alter/partition#how-to-set-partition-expression
+            if not self._match(TokenType.PARTITION):
+                return None
+
+            # Support for PARTITION ID '201901' syntax
+            if self._match_text_seq("ID"):
+                return self.expression(
+                    exp.Partition,
+                    expressions=[
+                        self.expression(
+                            exp.PartitionId,
+                            this=self._parse_string(),
+                        )
+                    ],
+                )
+
+            return self.expression(
+                exp.Partition,
+                expressions=self._parse_expressions(),
+            )
+
     class Generator(generator.Generator):
         QUERY_HINTS = False
         STRUCT_DELIMITER = ("(", ")")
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -578,6 +578,28 @@ def _parse_index_constraint(
                 granularity=granularity,
             )
 
+        def _parse_partition(self) -> t.Optional[exp.Partition]:
+            # https://clickhouse.com/docs/en/sql-reference/statements/alter/partition#how-to-set-partition-expression
+            if not self._match(TokenType.PARTITION):
```
georgesittas, line 583: It looks like both `PART` and `PARTITION` are valid keywords, so we can do the following here:

```python
if not self._match_texts(("PART", "PARTITION")):
    return None
```

sqlglot/dialects/clickhouse.py
```diff
@@ -578,6 +578,28 @@ def _parse_index_constraint(
...
+            if self._match_text_seq("ID"):
+                return self.expression(
+                    exp.Partition,
+                    expressions=[
+                        self.expression(
+                            exp.PartitionId,
+                            this=self._parse_string(),
+                        )
+                    ],
+                )
```
georgesittas, line 596: Let's rewrite this slightly:
```suggestion
            if self._match_text_seq("ID"):
                # Corresponds to the PARTITION ID <string_value> syntax
                expressions = [self.expression(exp.PartitionId, this=self._parse_string())]
            else:
                expressions = self._parse_expressions()

            return self.expression(exp.Partition, expressions=expressions)
```

sqlglot/dialects/clickhouse.py
```diff
@@ -578,6 +578,28 @@ def _parse_index_constraint(
...
+
+            return self.expression(
+                exp.Partition,
+                expressions=self._parse_expressions(),
+            )
```
georgesittas, line 601:
```suggestion
```

sqlglot/dialects/clickhouse.py
```diff
@@ -828,3 +850,9 @@ def indexcolumnconstraint_sql(self, expression: exp.IndexColumnConstraint) -> st
             granularity = f" GRANULARITY {granularity}" if granularity else ""
 
             return f"INDEX{this}{expr}{index_type}{granularity}"
+
+        def partition_sql(self, expression: exp.Partition) -> str:
+            return f"PARTITION {self.expressions(expression, flat=True)}"
+
+        def partitionid_sql(self, expression: exp.PartitionId) -> str:
+            return f"ID {expression.this}"
```
georgesittas, line 858:
```suggestion
            return f"ID {self.sql(expression.this)}"
```

sqlglot/expressions.py
```diff
@@ -2177,6 +2177,11 @@ class PartitionRange(Expression):
     arg_types = {"this": True, "expression": True}
 
 
+# https://clickhouse.com/docs/en/sql-reference/statements/alter/partition#how-to-set-partition-expression
+class PartitionId(Expression):
+    arg_types = {"this": True}
```
georgesittas, line 2182: `arg_types` is already defined like this in the superclass, so we can simplify this a bit:
```suggestion
    pass
```

georgesittas (commented):
Thanks for the PR @GaliFFun! I've left a few comments, should be good to go after we address them.

sqlglot/dialects/clickhouse.py
GaliFFun: Unfortunately, mypy does not like this.

```
Incompatible types in assignment (expression has type "list[Expression]", variable has type "list[PartitionId]")
```

How would you like me to fix this issue? I can declare the variable like this and mypy will stop complaining:

```python
expressions: t.List[exp.Expression]
```

Or I can replace the if clause with if expression. Mypy also stops complaining:

```python
  expressions = (
      # Corresponds to the PARTITION ID <string_value> syntax
      [self.expression(exp.PartitionId, this=self._parse_string())]
      if self._match_text_seq("ID")
      else self._parse_expressions()
  )
  ```

Comment (georgesittas) [reply sqlglot/dialects/clickhouse.py:596]: I updated my comment earlier and included a type hint in the first branch, doesn't that work?

Comment (GaliFFun) [reply sqlglot/dialects/clickhouse.py:596]: Will check that! Did not see the update :(

File: sqlglot/dialects/clickhouse.py (hunk 1 above)
Comment (GaliFFun) [reply sqlglot/dialects/clickhouse.py:583]: I feel like that the `PART` should be separated from `PARTITION`. The current implementation does store the used keyword, so generating SQL will yield `PARTITION` every time. Moreover, not every `ALTER` partition statement supports working with parts. And unfortunately, the doc is poorly written and the `PART` expression supports only string literals:

```sql
alter table tmp.foo drop part tuple(toYYYYMM(toDate('2019-01-25')))
```

```
DB::Exception: Syntax error: failed at position 31 ('tuple'): tuple(toYYYYMM(toDate('2019-01-25'))). Expected string literal. (SYNTAX_ERROR)
```

```sql
alter table tmp.foo drop part id '1'
```

```
DB::Exception: Syntax error: failed at position 31 ('id'): id '1'. Expected string literal. (SYNTAX_ERROR)
```

georgesittas: Oh they have different semantics? I thought they simply allowed it as an abbreviation, is that not the case?

sqlglot/dialects/clickhouse.py
GaliFFun: Yeah, it works now! Updated the code

sqlglot/dialects/clickhouse.py
GaliFFun: Yeah, the part is basically part of data on the disk, and partition is logical subset of parts with the same partition key

georgesittas: Gotcha, thanks for clarifying. Looks good then.

georgesittas (approved):
Appreciate the quick turnaround here!

PR #3437, opened 2024-05-09
Fix: handle empty string in connector comment padding

uncledata:
Problem:
if comment is an empty string pad_comment fails on IndexError on comment[0]. Adding a simple check if comment is not empty

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -648,9 +648,8 @@ class Generator(metaclass=_Generator):
         return f"{self.sep(sep)}{sql}"
 
     def pad_comment(self, comment: str) -> str:
-        if comment:
-            comment = " " + comment if comment[0].strip() else comment
-            comment = comment + " " if comment[-1].strip() else comment
+        comment = " " + comment if comment[0].strip() else comment
+        comment = comment + " " if comment[-1].strip() else comment
         return comment
 
     def maybe_comment(
@@ -2846,7 +2845,8 @@ class Generator(metaclass=_Generator):
                 stack.append(expression.right)
                 if expression.comments:
                     for comment in expression.comments:
-                        op += f" /*{self.pad_comment(comment)}*/"
+                        if comment:
+                            op += f" /*{self.pad_comment(comment)}*/"
                 stack.extend((op, expression.left))
             return op
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -647,8 +647,7 @@ class Generator(metaclass=_Generator):
     def seg(self, sql: str, sep: str = " ") -> str:
         return f"{self.sep(sep)}{sql}"
 
-    @staticmethod
-    def pad_comment(comment: str) -> str:
+    def pad_comment(self, comment: str) -> str:
         if comment:
             comment = " " + comment if comment[0].strip() else comment
             comment = comment + " " if comment[-1].strip() else comment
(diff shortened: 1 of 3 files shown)


tobymao:
please add a test case

tests/test_generator.py
```diff
@@ -43,3 +44,12 @@ def test_identify(self):
...
         assert parse_one("X as 1").sql(identify="safe") == 'X AS "1"'
+
+    def test_pad_comment(self):
+        self.assertEqual(Generator().pad_comment(""), "")
+        self.assertEqual(Generator().pad_comment(" leading"), " leading ")
+        self.assertEqual(Generator().pad_comment("trailing "), " trailing ")
+        self.assertEqual(
+            Generator().pad_comment(" leading and trailing "), " leading and trailing "
+        )
+        self.assertEqual(Generator().pad_comment("no padding"), " no padding ")
```
georgesittas, line 55: Can we instead test this end-to-end by simply adding an example that fails to roundtrip in pretty.sql?

tests/test_generator.py
```diff
@@ -43,3 +44,12 @@ def test_identify(self):
         assert parse_one("X").sql(identify="safe") == "X"
         assert parse_one("x as 1").sql(identify="safe") == '"x" AS "1"'
         assert parse_one("X as 1").sql(identify="safe") == 'X AS "1"'
+
+    def test_pad_comment(self):
+        self.assertEqual(Generator().pad_comment(""), "")
```
georgesittas, line 49: You don't have to instantiate `Generator` multiple times, just do it once

sqlglot/generator.py
```diff
@@ -647,9 +647,11 @@ def sep(self, sep: str = " ") -> str:
     def seg(self, sql: str, sep: str = " ") -> str:
         return f"{self.sep(sep)}{sql}"
 
-    def pad_comment(self, comment: str) -> str:
-        comment = " " + comment if comment[0].strip() else comment
-        comment = comment + " " if comment[-1].strip() else comment
+    @staticmethod
+    def pad_comment(comment: str) -> str:
```
georgesittas, line 651: Nit: I understand this was changed because `self` is not referenced in the body of the method, but note that we don't use `staticmethod`s anywhere else in the code base at the moment, so this creates a small inconsistency.

@tobymao wdyt? Should we refactor other similar methods into `@staticmethod`s?

sqlglot/generator.py
```diff
@@ -648,8 +648,9 @@ def seg(self, sql: str, sep: str = " ") -> str:
         return f"{self.sep(sep)}{sql}"
 
     def pad_comment(self, comment: str) -> str:
-        comment = " " + comment if comment[0].strip() else comment
-        comment = comment + " " if comment[-1].strip() else comment
+        if comment:
```
georgesittas, line 651: This makes the `if comment` check in L673 redundant now, let's get rid of it

uncledata:
hm it seems that if moving the check if comment is empty str/None to the method will still cause some issues for some tests:
1. returning an empty string will cause /**/ in situations like:
```sql
select x, --
from foo""",
```
   2. returning None will make the comment become /*None*/
 

 Maybe it's better to add a check to connector_sql and in maybe_comment? I see it's validated like this everywhere else (
```sql
if comment:
```

uncledata:
TL;DR:

Revert my changes in the first place and add if comment check in connector_sql:

```python
                        if comment:
                            op += f" /*{self.pad_comment(comment)}*/"
```

georgesittas:
Sure let's do that

georgesittas (approved):
LGTM, let's add a test in `test_transpile.py` and I'll get this in

PR #3440, opened 2024-05-09
fix: func to binary MOD generation

### LINKED ISSUE #3431: Incorrect transpiling of MOD function from trino to snowflake
Appears in 3.14
The following expression in Trino
`(MOD((DAY_OF_YEAR(world."orders.date") % 7) - 1 + 7, 7))`
is transpiled in Snowflake as 
`((DAYOFYEAR(orders.date) % 7) - 1 + 7 % 7)`

The transpiled expression returns incorrect result, as values could be negative (-1). Missing a parenthesis before the modulo operator, to make it correct.

Expected result:
`(MOD((DAYOFYEAR(orders.date) % 7) - 1 + 7, 7))`
or
`(((DAYOFYEAR(orders.date) % 7) - 1 + 7 ) % 7)`

```
from sqlglot import Dialects, parse_one
trino_sql = 'SELECT (MOD((DAY_OF_YEAR(orders.date) % 7) - 1 + 7, 7)) FROM orders'
parsed = parse_one(sql, dialect=Dialects.TRINO)
snowflake_sql = parsed.sql(dialect=Dialects.SNOWFLAKE)
```

**Official Documentation**
https://trino.io/docs/current/functions/datetime.html#day_of_year
https://trino.io/docs/current/functions/math.html#mod
https://docs.snowflake.com/en/sql-reference/functions/year
https://docs.snowflake.com/en/sql-reference/functions/mod

VaggelisD:
Fixes #3431

Dialects such as Trino support both `MOD(a, b)` and `a % b` representations of the modulo operation. Until now, the generation from the former to the latter would happen as is, which is incorrect for non-leaf operations; For example, the function call has the following semantics: 

`MOD(8 - 1 + 7, 7) == (8 - 1 + 7) % 7 == 0` 

But is generated to the following flat expression:

`8 - 1 + 7 % 7 == 8 - 1 + (7 % 7) == 7`

To solve this issue, when parsing a `MOD(...)` function call, the LHS is wrapped if it's a series of binary expressions i.e adds, subs etc, otherwise it remains as-is (case of single literals, nested function calls, columns etc).


sqlglot/dialects/bigquery.py
```diff
@@ -609,7 +609,6 @@ class Generator(generator.Generator):
             exp.IntDiv: rename_func("DIV"),
             exp.JSONFormat: rename_func("TO_JSON_STRING"),
             exp.Max: max_or_greatest,
-            exp.Mod: rename_func("MOD"),
```
georgesittas, line 612: I think we can bring this back and just do:

```python
exp.Mod: lambda self, e: self.func("MOD", e.this.unnest(), expression.expression.unnest()),
```

sqlglot/parser.py
```diff
@@ -74,6 +74,17 @@ def _builder(args: t.List, dialect: Dialect) -> E:
     return _builder
 
 
+def build_mod(args: t.List) -> exp.Mod:
+    this = seq_get(args, 0)
+    expression = seq_get(args, 1)
+
+    # Wrap the LHS if it's not a single node, e.g. MOD(a + 1, 7) -> (a + 1) % 7
```
georgesittas, line 81: We need to do this for both operands, e.g.

```
mysql> SELECT MOD(7, 8 - 1 + 7);
+-------------------+
| MOD(7, 8 - 1 + 7) |
+-------------------+
|                 7 |
+-------------------+
1 row in set (0.00 sec)

mysql> SELECT 7 % 8 - 1 + 7;
+---------------+
| 7 % 8 - 1 + 7 |
+---------------+
|            13 |
+---------------+
1 row in set (0.01 sec)
```

VaggelisD: True, no idea why I was fixated on LHS

georgesittas, line 81:
```suggestion
    # Wrap the operands if they are binary noes, e.g. MOD(a + 1, 7) -> (a + 1) % 7
```

georgesittas (approved):

PR #3441, opened 2024-05-09
feat(clickhouse): add support for ALTER TABLE REPLACE PARTITION statement

GaliFFun:
Added support for ClickHouse ALTER TABLE REPLACE PARTITION, which is very useful in ETL processes (which my company analyzes using sqlglot). ON CLUSTER support isn't there yet, willing to do that in the near future.

Doc reference:
https://clickhouse.com/docs/en/sql-reference/statements/alter/partition#replace-partition

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -878,5 +878,6 @@ class ClickHouse(Dialect):
             return f"ID {self.sql(expression.this)}"
 
         def replacepartition_sql(self, expression: exp.ReplacePartition) -> str:
-            return f"REPLACE {self.sql(expression.expression)} FROM {self.sql(expression, 'source')}"
-}"
+            return (
+                f"REPLACE {self.sql(expression.expression)} FROM {self.sql(expression, 'source')}"
+            )
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -878,5 +878,5 @@ class ClickHouse(Dialect):
             return f"ID {self.sql(expression.this)}"
 
         def replacepartition_sql(self, expression: exp.ReplacePartition) -> str:
-            source = expression.args.get("source")
-            return f"REPLACE {self.sql(expression.expression)} FROM {source}"
+            return f"REPLACE {self.sql(expression.expression)} FROM {self.sql(expression, 'source')}"
+}"
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -604,20 +604,8 @@ class ClickHouse(Dialect):
             if not partition or not self._match(TokenType.FROM):
                 return None
 
-            db = None
-            table = self._parse_id_var(any_token=False)
-
-            if self._match(TokenType.DOT):
-                db = table
-                table = self._parse_id_var(any_token=False)
-
-            if not table:
-                return None
-
             return self.expression(
-                exp.ReplacePartition,
-                expression=partition,
-                source=self.expression(exp.Table, this=table, db=db),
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -849,3 +876,7 @@ def partition_sql(self, expression: exp.Partition) -> str:
 
         def partitionid_sql(self, expression: exp.PartitionId) -> str:
             return f"ID {self.sql(expression.this)}"
+
+        def replacepartition_sql(self, expression: exp.ReplacePartition) -> str:
+            source = expression.args.get("source")
+            return f"REPLACE {self.sql(expression.expression)} FROM {source}"
```
georgesittas, line 882: Keep in mind to avoid interpolating expressions, because that silently invokes `.sql()` without a dialect and may produce incorrect SQL.
```suggestion
            return f"REPLACE {self.sql(expression.expression)} FROM {self.sql(expression, 'source')}"
}"
```

sqlglot/dialects/clickhouse.py
```diff
@@ -593,6 +598,28 @@ def _parse_partition(self) -> t.Optional[exp.Partition]:
...
+                table = self._parse_id_var(any_token=False)
+
+            if not table:
+                return None
+
+            return self.expression(
+                exp.ReplacePartition,
+                expression=partition,
+                source=self.expression(exp.Table, this=table, db=db),
+            )
```
georgesittas, line 621:
```suggestion
            return self.expression(
                exp.ReplacePartition, expression=partition, source=self._parse_table_parts()
            )
```

GaliFFun: I wanted to use this one, but ClickHouse allows only identifiers here, and `self._parse_table_parts()` allows to use string literals as well. If you think that would be okay, then I'll fix it if course

sqlglot/dialects/clickhouse.py
GaliFFun: Will keep that in mind, thanks!

sqlglot/dialects/clickhouse.py
georgesittas: Yep it's ok, sqlglot's parser is generally lenient because it doesn't aim to validate syntax

GaliFFun: OK then! Will keep that in mind as well :)

VaggelisD:
By looking at the Clickhouse docs, it seems that there's a long list of partition-related `ALTER`s with repeating specs, which will clutter `expressions.py`. Do we need that level of granularity? Would it make sense to have a catch-all `exp.AlterPartition` expression with an extra `kind` keyword storing the action e.g. `ATTACH / FREEZE / UNFREEZE` etc?

Also, for the `[ON CLUSTER <cluster>]`, I think you can solve it across board in `parser.py::_parse_alter()` by calling `_parse_on_property()` if `ON` is matched right before you visit the respective alter parsers.

GaliFFun:
@VaggelisD this seems like a good idea. The reason I used the current approach is the existence of `exp.DropPartition`. Which of the statements do you think should be included in `exp.AlterPartition`? I can modify the PR to include those and `ON CLUSTER` support as well, but it will take some time

VaggelisD:
The `ON CLUSTER` will not be a big problem, if you want to take a stab at it in this PR feel free and we can step in to help, otherwise we can revisit it in the future. 

As for the specialized vs catch-all expressions, no matter the approach I think it still be good to limit the scope of each PR to one statement along with it's tests (as you've done already), so don't feel pressured to add everything in at once right away. 

Let's wait until others chime in on this issue

georgesittas (approved):

PR #3445, opened 2024-05-09
fix: improve error handling for nested schema levels

tobymao:


sqlglot/schema.py
```diff
@@ -494,16 +497,17 @@ def ensure_column_mapping(mapping: t.Optional[ColumnMapping]) -> t.Dict:
...
 ) -> t.List[t.List[str]]:
     tables = []
     keys = keys or []
+    depth = dict_depth(schema) - 1 if depth is None else depth
 
     for k, v in schema.items():
-        if depth >= 2:
-            tables.extend(flatten_schema(v, depth - 1, keys + [k]))
-        elif depth == 1:
+        if depth == 1 or not isinstance(v, dict):
```
georgesittas, line 507: when can `v` _not_ be a `dict`?

georgesittas (approved):

sqlglot/schema.py
tobymao: when the schema has different levels as show in the test

georgesittas: I see, makes sense

PR #3449, opened 2024-05-10
fix(snowflake): COPY Subquery postfix

### LINKED ISSUE #3434: Parsing issue for Snowflake COPY INTO queries with subqueries
Hi folks, 

looks like we're facing an issue when parsing Snowflake `COPY INTO` queries with `SELECT` subqueries:
```
query = "COPY INTO test (c1) FROM (SELECT $1:c1::string FROM @mystage)"
parse_one(query, read="snowflake")
```

... which yields the following error:
```
sqlglot.errors.ParseError: Expected table name but got <Token token_type: TokenType.L_PAREN, text: (, line: 1, col: 32, start: 31, end: 31, comments: []>. Line 1, Col: 32.
COPY INTO t_d56103e1 (c1) FROM (SELECT $1:c1::string FROM @"mystage")
```

(The query has been validated to run successfully against Snowflake.)

Please let me know if you need any additional information, and whether I can help out with a PR/fix for this. 🙌  Thanks!

VaggelisD:
Fixes #3434

This PR addresses the following issues:
- COPY would only allow files in the `FROM` clause, extend it to allow subqueries as well

- When parsing staged files such as `@foo` we'd also consume the following parentheses coming from (optional) staged file options such as `SELECT * FROM @foo (FILE_FORMAT => ...)`. However, if there are no options specified we would end up consuming _any_ following `R_PAREN` as well e.g. `SELECT * FROM (SELECT * FROM @foo)`

- For transforming semi-structured data Snowflake specifies the subquery columns as `[<alias>.]$<file_col_num>[.<element>]`. However, the tokenizer always appends the `TokenType.DOT` to following a number e.g. `$1.elem -> $1.e lem`. To solve this issue, the tokenizer will only parse an integer (and not a real) if the number is prefixed by a token of type `TokenType.PARAMETER`

- The base parser parses `exp.Parameter` as `'[' <this> ':' <expression> ']'` but only generates `this`. Snowflake parsers `$<num>` as `exp.Parameter` which can optionally be followed by a path & element name such as `$1:a.c`. For this reason, `exp.Parameter` now generates `<this> [':' <expression>]` 

Docs
-----
- [Snowflake COPY](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table)
- [Snowflake COPY Transform Subquery specs](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#transformation-parameters)
- [Snowflake Querying semi-structured data](https://docs.snowflake.com/en/user-guide/querying-semistructured#traversing-semi-structured-data)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -693,11 +693,10 @@ class Snowflake(Dialect):
             parts = [self._advance_any(ignore_reserved=True)]
 
             # We avoid consuming a comma token because external tables like @foo and @bar
-            # can be joined in a query with a comma separator.
-            while (
-                self._is_connected()
-                and not self._match(TokenType.COMMA, advance=False)
-                and not self._match(TokenType.R_PAREN, advance=False)
+            # can be joined in a query with a comma separator, as well as closing paren
+            # in case of subqueries
+            while self._is_connected() and not self._match_set(
+                (TokenType.COMMA, TokenType.R_PAREN), advance=False
             ):
                 parts.append(self._advance_any(ignore_reserved=True))
 
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -602,7 +602,7 @@ class Snowflake(Dialect):
                 pattern = None
 
                 wrapped = self._match(TokenType.L_PAREN)
-                while self._curr and (wrapped and not self._match(TokenType.R_PAREN)):
+                while self._curr and wrapped and not self._match(TokenType.R_PAREN):
                     if self._match_text_seq("FILE_FORMAT", "=>"):
                         file_format = self._parse_string() or super()._parse_table_parts(
                             is_db_reference=is_db_reference
@@ -684,7 +684,7 @@ class Snowflake(Dialect):
         def _parse_file_location(self) -> t.Optional[exp.Expression]:
             # Parse either a subquery or a staged file
             return (
-                self._parse_primary()
+                self._parse_select(table=True)
                 if self._match(TokenType.L_PAREN, advance=False)
                 else self._parse_table_parts()
             )
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -601,8 +601,8 @@ class Snowflake(Dialect):
                 file_format = None
                 pattern = None
 
-                self._match(TokenType.L_PAREN)
-                while self._curr and not self._match(TokenType.R_PAREN):
+                wrapped = self._match(TokenType.L_PAREN)
+                while self._curr and (wrapped and not self._match(TokenType.R_PAREN)):
                     if self._match_text_seq("FILE_FORMAT", "=>"):
                         file_format = self._parse_string() or super()._parse_table_parts(
                             is_db_reference=is_db_reference
@@ -682,14 +682,23 @@ class Snowflake(Dialect):
             return self.expression(exp.LocationProperty, this=self._parse_location_path())
 
         def _parse_file_location(self) -> t.Optional[exp.Expression]:
-            return self._parse_table_parts()
+            # Parse either a subquery or a staged file
+            return (
+                self._parse_primary()
+                if self._match(TokenType.L_PAREN, advance=False)
+                else self._parse_table_parts()
+            )
 
         def _parse_location_path(self) -> exp.Var:
             parts = [self._advance_any(ignore_reserved=True)]
 
             # We avoid consuming a comma token because external tables like @foo and @bar
             # can be joined in a query with a comma separator.
-            while self._is_connected() and not self._match(TokenType.COMMA, advance=False):
+            while (
+                self._is_connected()
+                and not self._match(TokenType.COMMA, advance=False)
+                and not self._match(TokenType.R_PAREN, advance=False)
+            ):
                 parts.append(self._advance_any(ignore_reserved=True))
 
             return exp.var("".join(part.text for part in parts if part))
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2310,7 +2310,10 @@ class Generator(metaclass=_Generator):
 
     def parameter_sql(self, expression: exp.Parameter) -> str:
         this = self.sql(expression, "this")
-        return f"{self.PARAMETER_TOKEN}{this}"
+        expr = self.sql(expression, "expression")
+        expr = f":{expr}" if expr else ""
+
+        return f"{self.PARAMETER_TOKEN}{this}{expr}"
 
     def sessionparameter_sql(self, expression: exp.SessionParameter) -> str:
         this = self.sql(expression, "this")
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2310,10 +2310,7 @@ class Generator(metaclass=_Generator):
 
     def parameter_sql(self, expression: exp.Parameter) -> str:
         this = self.sql(expression, "this")
-        expr = self.sql(expression, "expression")
-        expr = f":{expr}" if expr else ""
-
-        return f"{self.PARAMETER_TOKEN}{this}{expr}"
+        return f"{self.PARAMETER_TOKEN}{this}"
 
     def sessionparameter_sql(self, expression: exp.SessionParameter) -> str:
         this = self.sql(expression, "this")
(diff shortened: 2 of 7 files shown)


sqlglot/tokens.py
```diff
@@ -1187,6 +1187,8 @@ def _scan_number(self) -> None:
             if self._peek.isdigit():
                 self._advance()
             elif self._peek == "." and not decimal:
+                if self.tokens and self.tokens[-1].token_type == TokenType.PARAMETER:
```
tobymao, line 1190: why would self.tokens be empty?

sqlglot/dialects/snowflake.py
```diff
@@ -601,8 +601,8 @@ def _parse_table_parts(
                 file_format = None
                 pattern = None
 
-                self._match(TokenType.L_PAREN)
-                while self._curr and not self._match(TokenType.R_PAREN):
+                wrapped = self._match(TokenType.L_PAREN)
+                while self._curr and (wrapped and not self._match(TokenType.R_PAREN)):
```
tobymao, line 605: i don't think you need the parenthesis

self._curr and wrappend and not self._match right?

sqlglot/dialects/snowflake.py
```diff
@@ -682,14 +682,23 @@ def _parse_location_property(self) -> exp.LocationProperty:
             return self.expression(exp.LocationProperty, this=self._parse_location_path())
 
         def _parse_file_location(self) -> t.Optional[exp.Expression]:
-            return self._parse_table_parts()
+            # Parse either a subquery or a staged file
+            return (
+                self._parse_primary()
```
tobymao, line 687: maybe it should be self._parse_select(table=True)

sqlglot/tokens.py
VaggelisD: This pattern appears in a couple other places in the tokenizer, it's main purpose afaict is to safeguard against statements which are standalone numbers such as the very first tests in `identity.sql`

sqlglot/dialects/snowflake.py
VaggelisD: Right, shouldn't make a difference. I usually wrap such paired bools to put a mental (and future) note when re-reading these chained conditions. Do we want to preserve the most simplified version of branches in general?

tobymao: yea let's simplify

tobymao (approved):

sqlglot/dialects/snowflake.py
```diff
@@ -682,14 +682,23 @@ def _parse_location_property(self) -> exp.LocationProperty:
             return self.expression(exp.LocationProperty, this=self._parse_location_path())
 
         def _parse_file_location(self) -> t.Optional[exp.Expression]:
-            return self._parse_table_parts()
+            # Parse either a subquery or a staged file
+            return (
+                self._parse_select(table=True)
+                if self._match(TokenType.L_PAREN, advance=False)
+                else self._parse_table_parts()
+            )
```
georgesittas, line 690: Can't we just do `self._parse_table()` here?

sqlglot/generator.py
```diff
@@ -2310,7 +2310,10 @@ def star_sql(self, expression: exp.Star) -> str:
 
     def parameter_sql(self, expression: exp.Parameter) -> str:
         this = self.sql(expression, "this")
-        return f"{self.PARAMETER_TOKEN}{this}"
+        expr = self.sql(expression, "expression")
+        expr = f":{expr}" if expr else ""
```
georgesittas, line 2314: Why do we do `:{expr}`? The colon part should be parsed separately right? E.g.

```
>>> import sqlglot
>>> sqlglot.parse_one("a:b", "snowflake")
JSONExtract(
  this=Column(
  expression=JSONPath(
    expressions=[
      JSONPathRoot(),
      JSONPathKey(this=b)]))
```

tests/dialects/test_snowflake.py
```diff
@@ -1869,3 +1870,14 @@ def test_copy(self):
         self.validate_identity(
             """COPY INTO mytable FROM 'azure://myaccount.blob.core.windows.net/mycontainer/data/files' CREDENTIALS = (AZURE_SAS_TOKEN = 'token') ENCRYPTION = (TYPE = 'AZURE_CSE' MASTER_KEY = 'kPx...') FILE_FORMAT = (FORMAT_NAME = my_csv_format)"""
         )
+        self.validate_identity(
+            """COPY INTO mytable (col1, col2) FROM 's3://mybucket/data/files' FILES = ('file1', 'file2') PATTERN = 'pattern' file_format = (FORMAT_NAME = my_csv_format NULL_IF = ('str1', 'str2')) PARSE_HEADER = TRUE""",
```
georgesittas, line 1874: should we just make this `FILE_FORMAT` and get rid of the duplicate line?

sqlglot/dialects/snowflake.py
```diff
@@ -682,14 +682,23 @@ def _parse_location_property(self) -> exp.LocationProperty:
...
         def _parse_location_path(self) -> exp.Var:
             parts = [self._advance_any(ignore_reserved=True)]
 
             # We avoid consuming a comma token because external tables like @foo and @bar
             # can be joined in a query with a comma separator.
-            while self._is_connected() and not self._match(TokenType.COMMA, advance=False):
+            while (
+                self._is_connected()
+                and not self._match(TokenType.COMMA, advance=False)
+                and not self._match(TokenType.R_PAREN, advance=False)
```
georgesittas, line 700: was this added to avoid consuming the closing `R_PAREN` of a subquery in `(SELECT ... FROM @foo)`?

sqlglot/dialects/snowflake.py
VaggelisD: This would indeed combine the two but it will consume the next token as table alias, right?

sqlglot/generator.py
VaggelisD: The parser consumes `TokenType.COLON` and the following expression, but we could override `parse_parameter()` in snowflake.py to parse just `this` ?

tests/dialects/test_snowflake.py
VaggelisD: Oh right, no idea why I originally split it in 2 lines, good catch

sqlglot/dialects/snowflake.py
VaggelisD: Right, this combined with the `wrapped` condition in `parse_table_parts()`

sqlglot/dialects/snowflake.py
georgesittas: Ah I see, yeah makes sense

sqlglot/generator.py
georgesittas: I don't think consuming the colon as part of the `$` parameter is correct. E.g. in this branch we get:

```python
>>> import sqlglot
>>> sqlglot.parse_one("SELECT $1:a.b", "snowflake")
Select(
  expressions=[
    Dot(
      this=Parameter(
        this=Literal(this=1, is_string=False),
        expression=Var(this=a)),
      expression=Identifier(this=b, quoted=False))])
```

Shouldn't this be a `JSONExtract` node instead? The `COLON` token makes sense as a placeholder I think. We need to make sure to not mess up the parsing of `:variable` in Snowflake, because that's [also valid](https://docs.snowflake.com/en/sql-reference/bind-variables).

sqlglot/dialects/snowflake.py
georgesittas, line 700: What about
```suggestion
                and not self._match_set((TokenType.COMMA, TokenType.R_PAREN), advance=False)
```

sqlglot/generator.py
VaggelisD: Right, I guess both representations are correct in Snowflake but the AST is not 100% there. That's because the base parser `_parse_parameter()` seems to be consuming a series of tokens and expressions but generates just `this` even before this PR, and although I wasn't sure why I didn't want to extend the scope even more. That being said:

1. Do we want to make sure that the logic in `_parse_parameter()` is still needed in other dialects as well ? If it's not needed, we can just simplify it to parse & generate `<TokenType.PARAMETER> <this>`

2. If the parameter logic is indeed correct and needed elsewhere, would it be best that we override it in snowflake to only parse `this`? From what I see, that will leave the colon as is which will be parsed properly into a json path

georgesittas: I think only hive needs to consume the `COLON` token, this is the only test that failed after commenting it out:

```sql
SELECT ${hiveconf:some_var}
```

I think I added it a while ago without realizing this would affect Snowflake's syntax. Perhaps we can try overriding it in Hive to fix this.

georgesittas: We can now get rid of this logic in the snowflake generator method

georgesittas (approved):
LGTM, thanks for addressing the comments 👍

PR #3450, opened 2024-05-10
feat: add support for ON property in ALTER and DROP statements

GaliFFun:
Added support for `ON` property in `ALTER` and `DROP` statements, so `ON CLUSTER` could be used with the ClickHouse dialect. A little follow-up PR to #3441. 

Documentation reference:
https://clickhouse.com/docs/en/sql-reference/statements/drop
https://clickhouse.com/docs/en/sql-reference/statements/alter/column

### DIFF
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -439,6 +439,10 @@ class TestClickhouse(Validator):
         )
         self.validate_identity("ALTER TABLE visits REPLACE PARTITION ID '201901' FROM visits_tmp")
 
+        self.validate_identity("ALTER TABLE visits ON CLUSTER test_cluster DROP COLUMN col1")
+        self.validate_identity("DROP TABLE visits ON CLUSTER test_cluster")
+        self.validate_identity("DROP DICTIONARY foo ON CLUSTER test_cluster")
+
     def test_cte(self):
         self.validate_identity("WITH 'x' AS foo SELECT foo")
         self.validate_identity("WITH ['c'] AS field_names SELECT field_names")
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -439,10 +439,6 @@ class TestClickhouse(Validator):
         )
         self.validate_identity("ALTER TABLE visits REPLACE PARTITION ID '201901' FROM visits_tmp")
 
-        self.validate_identity("ALTER TABLE visits ON CLUSTER test_cluster DROP COLUMN col1")
-        self.validate_identity("DROP TABLE visits ON CLUSTER test_cluster")
-        self.validate_identity("DROP DICTIONARY foo ON CLUSTER test_cluster")
-
     def test_cte(self):
         self.validate_identity("WITH 'x' AS foo SELECT foo")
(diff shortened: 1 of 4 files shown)


tests/dialects/test_clickhouse.py
```diff
@@ -439,6 +439,10 @@ def test_clickhouse(self):
         )
         self.validate_identity("ALTER TABLE visits REPLACE PARTITION ID '201901' FROM visits_tmp")
 
+        self.validate_identity("ALTER TABLE visits ON CLUSTER test_cluster DROP COLUMN col1")
+        self.validate_identity("DROP TABLE visits ON CLUSTER test_cluster")
+        self.validate_identity("DROP DICTIONARY foo ON CLUSTER test_cluster")
+
```
VaggelisD, line 445: We could include tests for other clustered createables as well such as `DROP DATABASE`, `DROP FUNCTION` etc, most Clickhouse ones already exist in `parser.py::CREATABLES`. Also, since they all follow the same syntax, I think it will be fine to automate their identity tests with subtests, e.g. check out how we test different combinations of `MATCH_RECOGNIZE` in `test_snowflake.py::test_match_recognize()`.

Also a small nit, try to reduce empty lines in these chained identity tests, it's best to use them when we add  other types of tests  (e.g. `validate_all`) right after these

VaggelisD (commented):
Great work, really appreciate the PRs!

I think it looks good, two minor comments from my initial pass:

tests/dialects/test_clickhouse.py
GaliFFun: Some of the `CREATABLES` are not supported by the `DROP` statement parser, so the parser falls back to parsing those as `Command`. Would it be okay with you if I use the following implementation?

```python
    def test_drop_on_cluster(self):
        for creatable in ClickHouse.Parser.CREATABLES:
            self.validate_identity(
                f"DROP {creatable.name} test ON CLUSTER test_cluster", check_command_warning=True
            )
```

GaliFFun: Oops, setting the check_command_warning to True fails those tests now. Any ideas on how to handle those outputs better?

VaggelisD: I think for now we can test them by listing them such as:

```
def test_drop_on_cluster(self):
  for creatable in ("DATABASE", "TABLE", "FUNCTION" ...):
    with self.subtest(f"Test DROP {creatable} ON CLUSTER):
       self.validate_identity(f"DROP {creatable} test ON CLUSTER test_cluster") 
```

When we expand to more Clickhouse specific creatables, we can revisit these tests and see if it's worth to refactor them by overriding `CREATABLES`.

tobymao: i'm fine either way, we can keep what you had

GaliFFun: Updated the tests using the suggestion from @VaggelisD. I've used the subset of `Parser.CREATABLES` and CREATE targets from ClickHouse [docs](https://clickhouse.com/docs/en/sql-reference/statements/createhttps://clickhouse.com/docs/en/sql-reference/statements/create)

Issue #3451, opened 2024-05-10
Incorrect generating of SHA expression to bigquery dialect


viplazylmht:
SQLglot version: 23.15.1
The issue still exists on main

**Fully reproducible code snippet**

Actually:
```python
>>> from sqlglot import parse_one
>>> parse_one("SHA(col_x)", read="").sql(dialect="bigquery")
'SHA(col_x)'
>>>
```
Expected output:
```python
>>> from sqlglot import parse_one
>>> parse_one("SHA(col_x)", read="").sql(dialect="bigquery")
'SHA1(col_x)'
>>>
```

**Official Documentation**
- https://cloud.google.com/bigquery/docs/reference/standard-sql/hash_functions#sha1

PR #3453, opened 2024-05-10
fix(bigquery): fix SHA1 generator

viplazylmht:
Fixes https://github.com/tobymao/sqlglot/issues/3451


tobymao:
let's run make style and we can get this in thank you so much

PR #3457, opened 2024-05-10
Feat!: improve transpilation of TABLESAMPLE clause

### LINKED ISSUE #3456: Rendering of `TABLESAMPLE` yields invalid `PERCENT` keyword for Postgres dialect
Hi folks, 

big kudos for the quick turnaround on some of the issues I've recently filed, really appreciated! 🙌 🚀 

Looks like we've just hit have another small one - related to rendering of `TABLESAMPLE` percentages for Postgres as target.

Given the following code:
```
query = "SELECT * FROM test AS _tmp TABLESAMPLE (5)"
result = parse_one(query, read="snowflake").sql(dialect="postgres")
print(result)
```

... `sqlglot` generates:
```
SELECT * FROM test AS _tmp TABLESAMPLE (5 PERCENT)
```

... however, this is not valid Postgres syntax, as Postgres:
* (1) requires the `SYSTEM` keyword to be present as default, and 
* (2) does not recognize the `PERCENT` keyword. 

So, the correct syntax in Postgres would be:
```
SELECT * FROM test AS _tmp TABLESAMPLE SYSTEM (5)
```

For point (2), we've been able to use the following workaround in our code base:
```
    Postgres.TABLESAMPLE_SIZE_IS_PERCENT = True
```

We also have a workaround for point (1) in place, so not a blocker, but would be great if it could be fixed for the Postgres dialect as well. 🙌

georgesittas:
Fixes #3456

References:
- https://duckdb.org/docs/sql/samples.html
- https://docs.snowflake.com/en/sql-reference/constructs/sample

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -304,6 +304,17 @@ class DuckDB(Dialect):
             ),
         }
 
+        def _parse_table_sample(self, as_modifier: bool = False) -> t.Optional[exp.TableSample]:
+            # https://duckdb.org/docs/sql/samples.html
+            sample = super()._parse_table_sample(as_modifier=as_modifier)
+            if sample and not sample.args.get("method"):
+                if sample.args.get("size"):
+                    sample.set("method", exp.var("RESERVOIR"))
+                else:
+                    sample.set("method", exp.var("SYSTEM"))
+
+            return sample
+
         def _parse_bracket(
             self, this: t.Optional[exp.Expression] = None
         ) -> t.Optional[exp.Expression]:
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -304,6 +304,17 @@ class Parser(parser.Parser):
...
         }
 
+        def _parse_table_sample(self, as_modifier: bool = False) -> t.Optional[exp.TableSample]:
+            # https://duckdb.org/docs/sql/samples.html
+            sample = super()._parse_table_sample(as_modifier=as_modifier)
+            if sample and not sample.args.get("method"):
+                if sample.args.get("size"):
+                    sample.set("method", exp.var("RESERVOIR"))
+                else:
+                    sample.set("method", exp.var("SYSTEM"))
```
georgesittas, line 314: > By default, DuckDB uses reservoir sampling when an exact number of rows is sampled, and system sampling when a percentage is specified. The sampling methods are described in detail below.

sqlglot/dialects/snowflake.py
```diff
@@ -334,6 +334,7 @@ def quote_identifier(self, expression: E, identify: bool = True) -> E:
 
     class Parser(parser.Parser):
         IDENTIFY_PIVOT_STRINGS = True
+        DEFAULT_SAMPLING_METHOD = "BERNOULLI"
```
georgesittas, line 337: > Sampling method is optional. If no method is specified, the default is BERNOULLI.

tobymao (approved):

PR #3462, opened 2024-05-11
feat: Add T-SQL DECLARE statement parsing

jlucas-fsp:
This PR adds DECLARE statement parsing to the TSQL dialect of SQLGlot. I haven't implemented any sql generation functionality for the new Declare and DeclareItem expression types, so running expression.sql will not work (let me know if this should be implemented before this PR could be merged)

This feature can parse scalar DECLAREs:
"DECLARE @var INT",
"DECLARE @var INT = 1",
"DECLARE @var1 INT, @var2 varchar(2)",
"DECLARE @string varchar(2) = 'AB'",
"DECLARE @var INT = (SELECT num FROM table WHERE condition)"

And, it can parse variable table DECLAREs:
"
DECLARE @myTable TABLE (
    ID INT NOT NULL,
    CONSTRAINT PK_ID PRIMARY KEY CLUSTERED (ID)
)
"

I didn't add any tests for this code to the existing test suite because the existing architecture seems to rely on parsing raw SQL into an AST, then converting the AST back into SQL to compare for accuracy, and this code can't convert the AST back into SQL yet.

Thank you all so much for all the work you've done on this library!


tobymao:
this won’t be accepted without tests. do you not plan to add any?

jlucas-fsp:
@tobymao, Is there any way to test this without implementing sql generation features as well? Or, do you recommend I implement generation features anyway and then proceed with making tests?

tobymao:
i’m not sure what you’re trying to achieve, you need to implement both parsing and generation

jlucas-fsp:
I've implemented SQL generation functionality and added tests for this (all passing on my machine). Those changes don't seem to automatically come to this PR, and I'm not sure what standard procedure would be in this case. I'm happy to open another PR with those new changes, but if you reopen this PR and those changes get automatically added here, that may be preferable.

To answer the implicit question in your previous message: My usecase for sqlglot doesn't involve transpilation or sql generation at all, it only involves the conversion of SQL to an AST. This is the first PR I've made to this repo and I wasn't sure exactly how much functionality around DECLAREs in TSQL I'd need to implement to get the parsing functionality in. Thank you for the reframe, and for future PRs, I now know what's expected.

tobymao:
i’ve reopened the pr, we can take a look at what you’ve done

tobymao:
in order for this or to be accepted, please link all documentation of declare, all forms need to be implemented or fall back to command, otherwise this pr would be a regression

jlucas-fsp:
Here is the documentation for TSQL DECLARE:
https://learn.microsoft.com/en-us/sql/t-sql/language-elements/declare-local-variable-transact-sql?view=sql-server-ver16

I covered all the examples from the documentation in my tests (the only unsupported feature is inline indexes and inline constraints not prefixed with the CONSTRAINT keyword). If the DECLARE statement doesn't perfectly align with the structure I'm expecting, or errors out, the whole statement falls back to being parsed as a command

Are there any other changes I should make?

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +716,53 @@ def parse_range():
...
+                    if self._match(TokenType.PARAMETER, advance=False):
+                        var = self._parse_id_var()
+                        size = None
+                        value = None
+                        # moving past the AS keyword if it's there
+                        self._match_text_seq("AS")
+
+                        if self._match(TokenType.TABLE):
+                            table = self.expression(exp.Table, this=var)
+                            data_type = self._parse_schema(this=table)
```
georgesittas, line 733: This also looks hacky, we should figure out a better way to construct the AST here.

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +716,53 @@ def parse_range():
...
+                        else:
+                            data_type = self._parse_types()  # Parse data type
+                            # Look for an optional size within parentheses
+                            if self._match(TokenType.L_PAREN):
+                                size = self._parse_bitwise()
+                                self._match(TokenType.R_PAREN)
```
georgesittas, line 739: This is redundant, the parser already handles parameterized types. I think `_parse_type()` suffices?

georgesittas: Let's also remove these comments - "Parse data type" doesn't really provide any useful insight into what's going on here.

tests/dialects/test_tsql.py
```diff
@@ -261,7 +260,8 @@ def test_tsql(self):
         self.validate_identity("SELECT * FROM ##foo")
         self.validate_identity("SELECT a = 1", "SELECT 1 AS a")
         self.validate_identity(
-            "DECLARE @TestVariable AS VARCHAR(100)='Save Our Planet'", check_command_warning=True
+            "DECLARE @TestVariable AS VARCHAR(100) = 'Save Our Planet'",
+            # check_command_warning=True
```
georgesittas, line 264: Since we don't parse `DECLARE` as a command anymore, why keep this commented? Just get rid of it.

tests/dialects/test_tsql.py
```diff
@@ -902,8 +902,8 @@ def test_rollback(self):
 
     def test_udf(self):
         self.validate_identity(
-            "DECLARE @DWH_DateCreated DATETIME = CONVERT(DATETIME, getdate(), 104)",
-            check_command_warning=True,
+            "DECLARE @DWH_DateCreated AS DATETIME2 = CONVERT(DATETIME2, GETDATE(), 104)",
+            # check_command_warning=True,
```
georgesittas, line 906: Ditto.

tests/dialects/test_tsql.py
```diff
@@ -975,9 +975,9 @@ def test_fullproc(self):
             BEGIN
                 SET XACT_ABORT ON;
 
-                DECLARE @DWH_DateCreated DATETIME = CONVERT(DATETIME, getdate(), 104);
-                DECLARE @DWH_DateModified DATETIME = CONVERT(DATETIME, getdate(), 104);
-                DECLARE @DWH_IdUserCreated INTEGER = SUSER_ID (SYSTEM_USER);
+                DECLARE @DWH_DateCreated AS DATETIME = CONVERT(DATETIME, getdate(), 104);
+                DECLARE @DWH_DateModified DATETIME2 = CONVERT(DATETIME2, GETDATE(), 104);
+                DECLARE @DWH_IdUserCreated INTEGER = SUSER_ID (CURRENT_USER());
```
georgesittas, line 980: Why does only the first one have `AS` between the variable name and the type?

georgesittas: Also, can `CURRENT_USER()` and `SYSTEM_USER` be used interchangeably?

tests/dialects/test_tsql.py
```diff
@@ -987,18 +987,18 @@ def test_fullproc(self):
...
+            "DECLARE @DWH_IdUserModified AS INTEGER = SUSER_ID(CURRENT_USER())",
+            "DECLARE @SalesAmountBefore AS FLOAT",
             "SELECT @SalesAmountBefore = SUM(SalesAmount) FROM TRANSF.[Pre_Merge_Sales_Real] AS S",
             "END",
         ]
 
-        with self.assertLogs(parser_logger):
-            for expr, expected_sql in zip(parse(sql, read="tsql"), expected_sqls):
-                self.assertEqual(expr.sql(dialect="tsql"), expected_sql)
+        # with self.assertLogs(parser_logger):
```
georgesittas, line 999: Remove this comment

tests/dialects/test_tsql.py
```diff
@@ -987,18 +987,18 @@ def test_fullproc(self):
 
         expected_sqls = [
             "CREATE PROCEDURE [TRANSF].[SP_Merge_Sales_Real] @Loadid INTEGER, @NumberOfRows INTEGER AS BEGIN SET XACT_ABORT ON",
-            "DECLARE @DWH_DateCreated DATETIME = CONVERT(DATETIME, getdate(), 104)",
-            "DECLARE @DWH_DateModified DATETIME = CONVERT(DATETIME, getdate(), 104)",
-            "DECLARE @DWH_IdUserCreated INTEGER = SUSER_ID (SYSTEM_USER)",
-            "DECLARE @DWH_IdUserModified INTEGER = SUSER_ID (SYSTEM_USER)",
-            "DECLARE @SalesAmountBefore float",
+            "DECLARE @DWH_DateCreated AS DATETIME2 = CONVERT(DATETIME2, GETDATE(), 104)",
+            "DECLARE @DWH_DateModified AS DATETIME2 = CONVERT(DATETIME2, GETDATE(), 104)",
```
georgesittas, line 991: Is this `CONVERT` still valid? Let's double check that the `DATETIME` -> `DATETIME2` doesn't silently break it.

tests/dialects/test_tsql.py
```diff
@@ -1012,14 +1012,14 @@ def test_fullproc(self):
...
-            "CREATE PROC [dbo].[transform_proc] AS DECLARE @CurrentDate VARCHAR(20)",
+            "CREATE PROC [dbo].[transform_proc] AS DECLARE @CurrentDate AS VARCHAR(20)",
             "SET @CurrentDate = CONVERT(VARCHAR(20), GETDATE(), 120)",
             "CREATE TABLE [target_schema].[target_table] (a INTEGER) WITH (DISTRIBUTION=REPLICATE, HEAP)",
         ]
 
-        with self.assertLogs(parser_logger):
-            for expr, expected_sql in zip(parse(sql, read="tsql"), expected_sqls):
-                self.assertEqual(expr.sql(dialect="tsql"), expected_sql)
+        # with self.assertLogs(parser_logger):
```
georgesittas, line 1020: Remove this

tests/dialects/test_tsql.py
```diff
@@ -1825,3 +1825,41 @@ def test_qualify_derived_table_outputs(self):
...
+            "DECLARE @X AS UserDefinedTableType",
+        )
+
+        # current unsupported cases
+        # inline constraints without the constrain keyword, and inline indexes are not supported by DECLARE parsing or CREATE TABLE parsing
+        # self.assertRaises(
+        #     ParseError,
+        #     parse_one,
+        #     "DECLARE @MyTableVar TABLE (EmpID INT NOT NULL, PRIMARY KEY CLUSTERED (EmpID), UNIQUE NONCLUSTERED (EmpID), INDEX CustomNonClusteredIndex NONCLUSTERED (EmpID));",
+        # )
```
georgesittas, line 1859: Let's remove these as well, no need to keep them around if they're unsupported.

georgesittas (commented):
Thanks for the PR @jlucas-fsp, I've left some comments.

georgesittas:
Let's add a test for `DECLARE @cursor_variable_name CURSOR` as well.

sqlglot/dialects/tsql.py
jlucas-fsp: I'm not sure how to improve this. Is the `if` check the hacky part, or is it the fact that I'm creating the table expression manually and passing it into `_parse_schema`? The `if` check seems necessary to me and I'm not sure how else to get `_parse_schema` the table information any other way

sqlglot/dialects/tsql.py
```diff
@@ -1067,3 +1123,21 @@ def drop_sql(self, expression: exp.Drop) -> str:
...
+            declare_items_str = ", ".join(declare_items)
+
+            return f"DECLARE {declare_items_str}"
+
+        def declareitem_sql(self, expression: exp.DeclareItem) -> str:
+            variable = self.sql(expression, "this")
+            type = self.sql(expression, "kind")
+            # generating sql from a table looks like "TableName (Schema)",
+            # but since we already have the table name (as variable), replace the "name" with the type "TABLE"
+            type = type.replace(variable, "TABLE")
```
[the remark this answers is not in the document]
jlucas-fsp: I'm not sure how to avoid using this kind of hack here. I'm using the same method to parse the table info in `DECLARE @var AS TABLE (Schema)` as is used to parse `CREATE TABLE var (Schema)`. Therefore, the .`sql()` method naturally creates the same SQL for both statements. This is bad because the ordering of `TABLE` and `ExampleTableName` is reversed between the two statements. Without creating a different parsing method, specific to variable tables that generates a different Expression (which seems redundant and creates its own issues), I don't see how to avoid using this hack. Though, perhaps I'm missing some part of the codebase that's meant to handle this kind of thing in a cleaner way. Maybe I need a transform here?

jlucas-fsp (commented):
Resolved most of the suggestions, but there are some suggestions I'm unsure about how to satisfy

georgesittas:
Thanks for addressing these comments @jlucas-fsp, I'll take another look shortly.

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +715,33 @@ def parse_range():
 
             return partition
 
+        def _parse_declare(self) -> exp.Declare | exp.Command:
+            expressions = self._try_parse(partial(self._parse_csv, self._parse_declareitem))
+            if not expressions:
```
georgesittas, line 720: The parsing may succeed, so we'd get a non-falsey `expressions` here, but there may be leftover tokens that we haven't consumed and so we'd produce an incomplete AST node. This addition ensures we will still fallback into parsing a command for these cases.
```suggestion
            if not expressions or self._curr:
```

sqlglot/expressions.py
```diff
@@ -1445,6 +1445,14 @@ class Pragma(Expression):
     pass
 
 
+class Declare(Expression):
+    arg_types = {"expressions": True}
+
+
+class DeclareItem(Expression):
+    arg_types = {"this": True, "kind": True, "default": False, "size": False}
```
georgesittas, line 1453:
```suggestion
    arg_types = {"this": True, "kind": True, "default": False}
```

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +715,33 @@ def parse_range():
...
+                return self._parse_as_command(self._prev)
+            return self.expression(exp.Declare, expressions=expressions)
+
+        def _parse_declareitem(self) -> t.Optional[exp.DeclareItem]:
+            var = self._parse_id_var(any_token=False, tokens=[TokenType.PARAMETER])
+            if not var:
+                return None
+            size = None
```
georgesittas, line 728:
```suggestion

```

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +715,33 @@ def parse_range():
...
             return partition
 
+        def _parse_declare(self) -> exp.Declare | exp.Command:
+            expressions = self._try_parse(partial(self._parse_csv, self._parse_declareitem))
+            if not expressions:
+                return self._parse_as_command(self._prev)
+            return self.expression(exp.Declare, expressions=expressions)
+
+        def _parse_declareitem(self) -> t.Optional[exp.DeclareItem]:
+            var = self._parse_id_var(any_token=False, tokens=[TokenType.PARAMETER])
```
georgesittas, line 725: I believe we can simply `_parse_id_var()` here right? No need to constrain this, `_parse_identifier()` will parse the `Parameter` because it calls `_parse_placeholder()` as a fallback.
```suggestion
            var = self._parse_id_var()
```

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +715,33 @@ def parse_range():
...
+            if not var:
+                return None
+            size = None
+            value = None
+            # moving past the AS keyword if it's there
+            self._match_text_seq("AS")
+
```
georgesittas, line 732:
```suggestion
            self._match(TokenType.ALIAS)
```

sqlglot/dialects/tsql.py
```diff
@@ -708,6 +715,33 @@ def parse_range():
...
+                table = self.expression(exp.Table, this=var)
+                data_type = self._parse_schema(this=table)
+            else:
+                data_type = self._parse_types()
+                if self._match(TokenType.EQ):
+                    value = self._parse_bitwise()
+
+            return self.expression(
+                exp.DeclareItem, this=var, kind=data_type, default=value, size=size
+            )
```
georgesittas, line 743:
```suggestion
            return self.expression(exp.DeclareItem, this=var, kind=data_type, default=value)
```

georgesittas (commented):
@jlucas-fsp a few final comments. Regarding the hack / cleaning up part, don't worry about it. Once these comments are addressed I'll do another pass to try and clean it up 👍

jlucas-fsp:
> @jlucas-fsp a few final comments. Regarding the hack / cleaning up part, don't worry about it. Once these comments are addressed I'll do another pass to try and clean it up 👍

Thank you so much for your detailed feedback on this PR, I very much appreciate it :)

georgesittas:
@jlucas-fsp check out https://github.com/tobymao/sqlglot/commit/49f7f857634ae85547c805ac53911895407dd7cb

Issue #3460, opened 2024-05-11
Behavior of HEX function is different between dialects


viplazylmht:
## Problem:

HEX function encodes a bytes or string into a hexadecimal encoding, but the result of the function is vary between dialects, due to the case of the result.

For example, 
in bigquery (lowercase):
```sql
select to_hex(MD5('Hello'))
|  Row  |              f0_            |  
| ----- | --------------------------- |
|   1   | 8b1a9953c4611296a827abf8c47 |
 ```
in presto/trino (uppercase):
```python
presto> select to_hex(to_utf8('some string'));
         _col0          
------------------------
 736F6D6520737472696E67 
(1 row)
```

in clickhouse return [uppercase](https://clickhouse.com/docs/en/sql-reference/functions/encoding-functions#hex).
in spark (uppercase), but hex(md5_digest(x)) produce a lowercase result:
```python
>>> spark.sql("select hex('Hello')").collect()
[Row(hex(Hello)='48656C6C6F')]
>>> spark.sql("select md5('Hello')").collect()
[Row(md5(Hello)='8b1a9953c4611296a827abf8c47804d7')]
```
In snowflake, the default behavior is uppercase, but we can pass an additional argument to control the case: https://docs.snowflake.com/en/sql-reference/functions/hex_encode#argumentshttps://docs.snowflake.com/en/sql-reference/functions/hex_encode#arguments

And so on...

=> So, as a transpiler tool, should we ensure the behavior of this function while transpiling?

## The solution if needed:

We can add a new exp.UpperHex (or exp.LowerHex, depends on the case of default exp.Hex).
Also, we add a flag to Dialect class, which mark the behavior of HEX function of the target dialect is uppercase or lowercase, then can safety and can simplify the SQL while parsing or generating the sql.

For example, if read dialect is bigquery (lowercase), write dialect is trino (uppercase), the following step will happen when transpiling the sql `select lower(to_hex(col_x))`: 

1. Parse sql to bigquery: becasue bq are already lowercased, the tree will look like: `exp.SELECT (exp.Hex( col_x ))`
2. Generate trino sql: because trino is uppercased, generator will be generating `SELECT LOWER(TO_HEX(col_x))`

If we transpile `select upper(to_hex(col_x))` from bigquery to trino, the code will work like this:
1. parse sql to `exp.Select( exp.UpperHex( col_x ))`
2. Generate trinno sql: `SELECT TO_HEX(col_x)`

**Describe alternatives you've considered**

[...]

PR #3463, opened 2024-05-12
feat(transpiler): handle different hex behavior for dialects

viplazylmht:
Fixes https://github.com/tobymao/sqlglot/issues/3460

- Default hex function of all dialects, including `SQLGlot` dialect, to uppercase. 
- Set HEX_LOWERCASE to True in the dialect if it produces lowercase output.
- Simplify LOWER and UPPER expression depending on HEX_LOWERCASE configuration.


sqlglot/dataframe/sql/functions.py
```diff
@@ -908,6 +908,10 @@ def hex(col: ColumnOrName) -> Column:
     return Column.invoke_expression_over_column(col, expression.Hex)
 
 
+def upperhex(col: ColumnOrName) -> Column:
+    return Column.invoke_expression_over_column(col, expression.UpperHex)
```
georgesittas, line 912: why was this added?

sqlglot/dialects/bigquery.py
```diff
@@ -245,6 +245,8 @@ class BigQuery(Dialect):
     # https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table
     PSEUDOCOLUMNS = {"_PARTITIONTIME", "_PARTITIONDATE"}
 
+    HEX_LOWERCASE: t.Optional[bool] = True
```
georgesittas, line 248:
```suggestion
    HEX_LOWERCASE: bool = True
```

tests/dataframe/unit/test_functions.py
```diff
@@ -1188,8 +1188,14 @@ def test_bin(self):
 
     def test_hex(self):
         col_str = SF.hex("cola")
-        self.assertEqual("HEX(cola)", col_str.sql())
+        self.assertEqual("LOWER(HEX(cola))", col_str.sql())
```
georgesittas, line 1191: why did this test change?

georgesittas:
Thanks for the PR @viplazylmht, I'll review more carefully tomorrow

tests/dataframe/unit/test_functions.py
viplazylmht: Because `col_str`  is a exp.Hex (lower hex), but when we call sql without dialect, it will use `SQLGlot` dialect as the default. And default of  HEX_LOWERCASE is False, so exp.Hex will produce `LOWER(HEX(x))` instead. 

So, I made another test for upperhex which work exactly like this old test:
https://github.com/tobymao/sqlglot/blob/ac4bc7e4447c818b42da7c8250a7713fcf1afb12/tests/dataframe/unit/test_functions.py#L1195-L1200

sqlglot/dataframe/sql/functions.py
viplazylmht: @georgesittas hmm as mentioned as below, i made `def test_upperhex(self):` to represent the logic of the old `def test_hex(self):`

georgesittas: Functions defined in this file are supposed to match pyspark's dataframe API, so not sure if it makes sense to define `upperhex` here. If you _needed_ to add this, it's probably wrong, cc @eakmanrq can you verify?

tests/dataframe/unit/test_functions.py
georgesittas: > it will use SQLGlot dialect as the default.

`col_str` is not a plain sqlglot `Expression` though, it's a dataframe `Column` object which is [generated into Spark by default](https://github.com/tobymao/sqlglot/blob/58d5f2bece42acdda5f8c08d30e6f61a5e538d4c/sqlglot/dataframe/sql/column.py#L214). So something seems off, I don't think this test should've changed?

viplazylmht: Hmm thank you for correcting me, it is Spark dialect, not SQLGlot dialect.

But, spark, like SQLGlot, is not a HEX_LOWERCASE. `SF.hex("cola")` create a Column which contains a Hex expression directly (stored in self.expression), and exp.Hex is treated as LowerHex, so the test will fail if unchanged. 

```python
from sqlglot.dialects.spark import Spark

spark_dialect = Spark()

# normal transpiler
parse_one("HEX(cola)").sql(dialect=spark_dialect)
# 'HEX(cola)'

# Create exp.Hex directly
exp.Hex(this="cola").sql(dialect=spark_dialect)
# 'LOWER(HEX(cola))'
```

Can you suggest me where we should change to pass this test, if without the modified above?

sqlglot/dialects/bigquery.py
```diff
@@ -245,6 +245,8 @@ class BigQuery(Dialect):
     # https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table
     PSEUDOCOLUMNS = {"_PARTITIONTIME", "_PARTITIONDATE"}
 
+    HEX_LOWERCASE: bool = True
```
georgesittas, line 248:
```suggestion
    HEX_LOWERCASE = True
```

sqlglot/dialects/dialect.py
```diff
@@ -290,6 +290,23 @@ class Dialect(metaclass=_Dialect):
         ) SELECT c FROM y;
     """
 
+    HEX_LOWERCASE: t.Optional[bool] = False
```
georgesittas, line 293:
```suggestion
    HEX_LOWERCASE = False
```

sqlglot/dialects/dialect.py
```diff
@@ -290,6 +290,23 @@ class Dialect(metaclass=_Dialect):
...
+    case, `HEX` can be wrapped by an additional lower or upper function to convert
+    the output to exact dialect.
+    For example,
+        `SELECT TO_HEX(x)`;
+        in Bigquery will be rewritten as the following one in Presto and Trino
+        `SELECT LOWER(TO_HEX(x))`;
+    In another example,
+        `SELECT TO_HEX(x)`;
+        in Presto will be rewritten as the following one in Bigquery
+        `SELECT UPPER(TO_HEX(x))`;
```
georgesittas, line 307: Let's describe this in a more compact way, this is too much.

sqlglot/dialects/presto.py
```diff
@@ -276,7 +276,7 @@ class Parser(parser.Parser):
                 this=seq_get(args, 0), substr=seq_get(args, 1), instance=seq_get(args, 2)
             ),
             "TO_CHAR": _build_to_char,
-            "TO_HEX": exp.Hex.from_arg_list,
+            "TO_HEX": exp.UpperHex.from_arg_list,
```
georgesittas, line 279: Why are we using `UpperHex` and not `Hex` by default if the base class defaults to `HEX_LOWERCASE = False`? This is inconsistent right? I'd expect `Hex` to just be treated as "upper hex".

sqlglot/expressions.py
```diff
@@ -5193,6 +5193,10 @@ class Hex(Func):
     pass
 
 
+class UpperHex(Func):
+    pass
```
georgesittas, line 5197: Should this be inheriting from `Hex`?

tests/dataframe/unit/test_functions.py
georgesittas: > But, spark, like SQLGlot, is not a HEX_LOWERCASE [...] and exp.Hex is treated as LowerHex, so the test will fail if unchanged.

The issue is that you're treating it as a `LowerHex` even though the default setting is treating `HEX` as "uppercase". See my other comment.

sqlglot/dataframe/sql/functions.py
tobymao: that's right, these functions are only used for pyspark's dataframe api

sqlglot/dialects/presto.py
viplazylmht: Thank you. So Hex should be treated as "upper hex" becasue the base `HEX_LOWERCASE = False`. And we are going to create a new LowerHex instead of UpperHex expression. I will update the PR.

sqlglot/dataframe/sql/functions.py
viplazylmht: I have removed changes in this file and test also.

VaggelisD:
Thanks for the PR!

Wouldn't it be preferable to encode the upper/lower as an argument and check that when transpiling to other dialects? E.g. if dialect A defaults to upper, then it will create a node like `exp.Hex(this=..., upper=True)`; If we want to transpile to a dialect B, we can compare B's flag with the node arg to decide whether to wrap the result. .

If there are no other differences between `Hex` and `Lowerhex`, I think this will deduplicate the code while helping dialects such as Snowflake that parameterize this behavior anyways

georgesittas:
> Wouldn't it be preferable to encode the upper/lower as an argument and check that when transpiling to other dialects? E.g. if dialect A defaults to upper, then it will create a node like exp.Hex(this=..., upper=True); If we want to transpile to a dialect B, we can compare B's flag with the node arg to decide whether to wrap the result. .

Depends on what you're looking to optimize. I think either approach is fine, but I'm leaning towards this one. It allows you to easily instance check `Hex` expressions, without having to first ensure an object _is_ an expression, and then lookup its `args`. It also avoids the boolean func arg complexity that we'd need to add and maintain, because `Generator.func` would choke on `upper`.

Another reason I'd like to avoid this pivot is to avoid scope creep. There's some level of inconsistency already on the different-expressions vs additional-args pattern in the AST, so this seems like a topic we could discuss separately if we want to be consistent in what we do.

VaggelisD:
Agreed about the inconsistency part, it would probably be best to solidify those patterns so that they can be enforced agnostic to the reviewer. 

In such cases where the "payload" is not that complex and doesn't greatly alter the semantics, I'm more in favor of adding it  as an arg, there's more checks associated to it but otherwise you'd also maintain a new node across all parsers/generators, to search for it you'd have to `find` multiple instances etc. 

@viplazylmht  Can you please mention the dialects that have been covered by this PR?

viplazylmht:
Thanks for the review, @georgesittas , @VaggelisD.

Currently, this PR cover: bigquery, presto, trino, clickhouse, hive, spark, spark2.

* Snowflake should work because the default behavior of HEX function is upper, unless they pass the `case` to the arg. 
* Databricks should work because the behind of scence is spark.

This PR is untested with the remaining dialects, which is supported by SQLGlot.

sqlglot/dialects/presto.py
```diff
@@ -385,6 +385,9 @@ class Generator(generator.Generator):
                 "ARRAY_JOIN", self.func("ARRAY_AGG", e.this), e.args.get("separator")
             ),
             exp.Hex: rename_func("TO_HEX"),
+            exp.LowerHex: lambda self, e: self.func(
+                "LOWER", self.func("TO_HEX", self.sql(e, "this"))
+            ),
```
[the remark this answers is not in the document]
viplazylmht: This is not a redundant. This will rename_func the HEX to TO_HEX.

https://github.com/tobymao/sqlglot/blob/9402b5d6dc87807da7124a0a2fa19cfd9fba02d2/sqlglot/generator.py#L1311-L1315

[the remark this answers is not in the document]
georgesittas: Ah got it, thanks for pointing it out

georgesittas:
Sounds good, thanks for checking. These are the remaining ones:

- Ducked HEX —> upper
- MySQL HEX —> upper
- Redshift TO_HEX—> lower
- SQLite HEX —> upper

I'll take this to the finish line

PR #3465, opened 2024-05-13
feat(clickhouse): add support for PROJECTION in CREATE TABLE statement

GaliFFun:
Added support for defining `PROJECTION` in `CREATE TABLE` statement for ClickHouse dialect

Documentation reference:
https://clickhouse.com/docs/en/sql-reference/statements/alter/projection#add-projection

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -90,7 +90,6 @@ class ClickHouse(Dialect):
             "LOWCARDINALITY": TokenType.LOWCARDINALITY,
             "MAP": TokenType.MAP,
             "NESTED": TokenType.NESTED,
-            "PROJECTION": TokenType.PROJECTION,
             "SAMPLE": TokenType.TABLE_SAMPLE,
             "TUPLE": TokenType.STRUCT,
             "UINT128": TokenType.UINT128,
@@ -613,7 +612,7 @@ class ClickHouse(Dialect):
             )
 
         def _parse_projection_def(self) -> t.Optional[exp.ProjectionDef]:
-            if not self._match(TokenType.PROJECTION):
+            if not self._match_text_seq("PROJECTION"):
                 return None
 
             return self.expression(
@@ -622,10 +621,8 @@ class ClickHouse(Dialect):
                 expression=self._parse_wrapped(self._parse_statement),
             )
 
-        def _parse_schema_item(self) -> t.Optional[exp.Expression]:
-            return (
-                self._parse_constraint() or self._parse_projection_def() or self._parse_field_def()
-            )
+        def _parse_constraint(self) -> t.Optional[exp.Expression]:
+            return super()._parse_constraint() or self._parse_projection_def()
 
     class Generator(generator.Generator):
         QUERY_HINTS = False
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -90,6 +90,7 @@ class ClickHouse(Dialect):
             "LOWCARDINALITY": TokenType.LOWCARDINALITY,
             "MAP": TokenType.MAP,
             "NESTED": TokenType.NESTED,
+            "PROJECTION": TokenType.PROJECTION,
             "SAMPLE": TokenType.TABLE_SAMPLE,
             "TUPLE": TokenType.STRUCT,
             "UINT128": TokenType.UINT128,
@@ -611,6 +612,21 @@ class ClickHouse(Dialect):
                 exp.ReplacePartition, expression=partition, source=self._parse_table_parts()
             )
 
+        def _parse_projection_def(self) -> t.Optional[exp.ProjectionDef]:
+            if not self._match(TokenType.PROJECTION):
+                return None
+
+            return self.expression(
+                exp.ProjectionDef,
+                this=self._parse_id_var(),
+                expression=self._parse_wrapped(self._parse_statement),
+            )
+
+        def _parse_schema_item(self) -> t.Optional[exp.Expression]:
+            return (
+                self._parse_constraint() or self._parse_projection_def() or self._parse_field_def()
+            )
+
     class Generator(generator.Generator):
         QUERY_HINTS = False
         STRUCT_DELIMITER = ("(", ")")
@@ -878,3 +894,6 @@ class ClickHouse(Dialect):
             return (
                 f"REPLACE {self.sql(expression.expression)} FROM {self.sql(expression, 'source')}"
             )
+
+        def projectiondef_sql(self, expression: exp.ProjectionDef) -> str:
+            return f"PROJECTION {self.sql(expression.this)} {self.wrap(expression.expression)}"
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4600,12 +4600,9 @@ class Parser(metaclass=_Parser):
         if self._match_set(self.SELECT_START_TOKENS):
             self._retreat(index)
             return this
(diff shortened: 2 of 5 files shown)


sqlglot/tokens.py
```diff
@@ -327,6 +327,7 @@ class TokenType(AutoName):
     PREWHERE = auto()
     PRIMARY_KEY = auto()
     PROCEDURE = auto()
+    PROJECTION = auto()
```
georgesittas, line 330: We prefer not adding new tokens for cases like this one
```suggestion
```

sqlglot/dialects/clickhouse.py
```diff
@@ -90,6 +90,7 @@ class Tokenizer(tokens.Tokenizer):
             "LOWCARDINALITY": TokenType.LOWCARDINALITY,
             "MAP": TokenType.MAP,
             "NESTED": TokenType.NESTED,
+            "PROJECTION": TokenType.PROJECTION,
```
georgesittas, line 93:
```suggestion
```

sqlglot/dialects/clickhouse.py
```diff
@@ -611,6 +612,21 @@ def _parse_alter_table_replace(self) -> t.Optional[exp.Expression]:
                 exp.ReplacePartition, expression=partition, source=self._parse_table_parts()
             )
 
+        def _parse_projection_def(self) -> t.Optional[exp.ProjectionDef]:
+            if not self._match(TokenType.PROJECTION):
```
georgesittas, line 616:
```suggestion
            if not self._match_text_seq("PROJECTION"):
```

sqlglot/dialects/clickhouse.py
```diff
@@ -611,6 +612,21 @@ def _parse_alter_table_replace(self) -> t.Optional[exp.Expression]:
...
+                return None
+
+            return self.expression(
+                exp.ProjectionDef,
+                this=self._parse_id_var(),
+                expression=self._parse_wrapped(self._parse_statement),
```
georgesittas, line 622: Why is `_parse_wrapped` required here? Can't we just use `_parse_statement()` directly and get rid of `self.wrap(...)` in the generator method?

sqlglot/parser.py
```diff
@@ -4600,10 +4600,13 @@ def _parse_schema(self, this: t.Optional[exp.Expression] = None) -> t.Optional[e
         if self._match_set(self.SELECT_START_TOKENS):
             self._retreat(index)
             return this
-        args = self._parse_csv(lambda: self._parse_constraint() or self._parse_field_def())
+        args = self._parse_csv(self._parse_schema_item)
         self._match_r_paren()
         return self.expression(exp.Schema, this=this, expressions=args)
 
+    def _parse_schema_item(self) -> t.Optional[exp.Expression]:
+        return self._parse_constraint() or self._parse_field_def()
```
georgesittas, line 4608: I don't think we need to factor this out so it can be overridden in CH. We usually extend `CONSTRAINT_PARSERS` for cases similar to this one. Let's try to refactor this.

georgesittas:
Appreciate the PRs @GaliFFun 👍

sqlglot/parser.py
GaliFFun: I feel a little bit conflicted about this implementation. If I understand correctly, the `CONSTRAINT` parser requires first token to be of the `TokenType.CONSTRAINT` for "named" constraints. For those the supported syntax is:

`TokenType.CONSTRAINT id_var constraint_type expressions`

The `PROJECTION` definition syntax does not match those rules.

But there are "unnamed" constraints as well. And it looks like those are a fit - adding `PROJECTION` to `SCHEMA_UNNAMED_CONSTRAINTS` does help. But is it semantically correct? The projection is named (it has a name). Or am I misunderstanding the meaning of "unnamed"?

sqlglot/dialects/clickhouse.py
GaliFFun: I've used this to require left and right parens wrapping the query, is there a way to do this one better?

sqlglot/tokens.py
GaliFFun: Is there a rule for cases like this? I'll try to remember to check the dialect for parser dicts now, but is there something else to keep in mind?

georgesittas: My rule of thumb is to not add tokens so I can simply do `self._match`, because the `TokenType` enum will explode and that will create more complexity in recognizing their purpose / adding them to the right mappings in the parser.

Tokens are generally needed when needing to add a token to an existing mapping, eg. type tokens, `NO_PAREN_FUNCTION` tokens, etc.

sqlglot/parser.py
georgesittas: That's right, good observations. I was leading to updating the parser so this `PROJECTION` constraint is treated as a `SCHEMA_UNNAMED_CONSTRAINT`.

> But is it semantically correct? The projection is named (it has a name). Or am I misunderstanding the meaning of "unnamed"?

IIRC the "unnamed" part refers to the fact that the constraint appears as a single item in the schema definition. E.g. you have some constraints that look like:

```
                this is a "named" constraint, i.e. it's associated to a column definition
                     v
CREATE TABLE z (n INT DEFAULT 0 NOT NULL)
```

So from my perspective doing this is totally fine.

sqlglot/dialects/clickhouse.py
georgesittas: I guess this is fine. FWIW the difference is that this parses it as a `Select` whereas a single `self._parse_statement` will parse it as a `Subquery`.

sqlglot/parser.py
VaggelisD: Chiming in with not much context but I think I share @GaliFFun's opinion, afaict the named constraints are basically parsed as `CONSTRAINT <name> <unnamed_constraints>` i.e the name refers to the "alias" that would be bound to it (iirc so in case it's violated it can be printed in the error logs and be associated with that rule).

 I think we can extend `_parse_constraint` to parse other named constraints that do not necessarily begin with `CONSTRAINT` ?

VaggelisD: Sorry for the back and forth, I took another look at the docs:

(1) It seems that CH supports [named constraints](https://clickhouse.com/docs/en/sql-reference/statements/create/table#constraints) as is already supported, and 
(2) From my understanding the `PROJECTION` is semantically closer to an index than an actual constraint, so I'd agree that it would be best to add it as an separate entry in `CONSTRAINT_PARSERS` as with other index parameters.


@GaliFFun The constraint parser doesn't _require_ the `CONSTRAINT` keyword per se, it looks for it first in order to pass the (optional) `CONSTRAINT <name>`; The rest of the expression comes from "unnamed constraints" that are defined in `CONSTRAINT_PARSERS`, but what changes is whether the unnamed constraint will be parsed in it's own node or wrapped in a `exp.ColumnConstraint`

Any opinions ^ ?

georgesittas: Yeah let's do `ProjectionDef` and override it in CH, I don't think there's any point in being so pedantic in this case. There won't be any substantial difference in the AST if we treat it as a separate branch of `_parse_constraint` vs treating it as an "index".

GaliFFun: Updated the implementation

sqlglot/tokens.py
GaliFFun: Will need that in mind, thanks!

georgesittas (approved):
Thanks for addressing all comments, and again - appreciate the PRs!

PR #3466, opened 2024-05-13
fix(presto): fix DELETE DML statement for presto/trino

viplazylmht:
# Context

- Presto/Trino only support DELETE statement with FROM keyword and only allow to delete data from one table per query.
- Don't allow to have any table alias in the target table which will be deleted.

# Documentation

- https://prestodb.io/docs/current/sql/delete.html
- https://trino.io/docs/current/sql/delete.html
- https://prestodb.io/docs/current/sql/explain.html
- https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#delete_statement
- Looking for **Multi-Table Deletes** in https://dev.mysql.com/doc/refman/8.0/en/delete.html

# Reproduce

## Prerequisite
```bash
echo "connector.name=memory" > memory.properties
docker run -p 8080:8080 -it -v ./memory.properties:/opt/presto-server/etc/catalog/memory.properties --name presto prestodb/presto:latest

docker exec -it presto presto-cli

presto> create table memory.default.tbl(x int);
CREATE TABLE
presto> show create table memory.default.tbl;
           Create Table            
-----------------------------------
 CREATE TABLE memory.default.tbl ( 
    "x" integer                    
 )                                 
(1 row)
```

## Test DELETE DML statement
```bash
# normal case
presto> explain(type validate) delete from memory.default.tbl where x = 1;
 result 
--------
 true   
(1 row)

# failed cases
presto> explain(type validate) delete memory.default.tbl where x = 1;
Query 20240513_165153_00014_37tzg failed: line 1:31: mismatched input 'memory'. Expecting: 'FROM'

presto> explain(type validate) delete from memory.default.tbl t where t.x = 1;
Query 20240513_163843_00010_37tzg failed: line 1:55: mismatched input 't'. Expecting: '.', 'WHERE', <EOF>

presto> explain(type validate) delete from memory.default.tbl as t where t.x = 1;
Query 20240513_165118_00013_37tzg failed: line 1:55: mismatched input 'as'. Expecting: '.', 'WHERE', <EOF>

...
```

So, when transpiling, we have to handle the delete_sql to make Presto/Trino happy.

# Solution

- Presto only support DELETE FROM a table, without alias, so we need to remove the unnecessary parts.
- If the original DELETE statement contains more than one table to be deleted, just pick the first one.

```sql
DELETE ...
->
DELETE FROM ...
```

```sql
DELETE FROM tbl as t where t.col_x = 1;
->
DELETE FROM table where col_x =1;
```

```sql
DELETE tbl1 as z, tbl2 as t where z.id = 1;
->
DELETE FROM tble1 where id = 1
```

```sql
DELETE tbl1 as z, tbl2 as t where z.id = t.id;
->
DELETE FROM tbl1 where id = t.id  <<< FAILED CASE
```

# Test

- Add tests for bigquery dialect. Please note that bigquery can delete only a table at a time, look like presto. 
- Maybe we need more test for mysql (mysql allow delete more than one table each call).

# Discussion 

Should we raise an error if there is more than one table to be deleted when transpiling to presto/trino dialect?

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -32,6 +32,7 @@ from sqlglot.dialects.hive import Hive
 from sqlglot.dialects.mysql import MySQL
 from sqlglot.helper import apply_index_offset, seq_get
 from sqlglot.tokens import TokenType
+from sqlglot.transforms import unqualify_columns
 
 
 def _explode_to_unnest_sql(self: Presto.Generator, expression: exp.Lateral) -> str:
@@ -630,23 +631,26 @@ class Presto(Dialect):
         def delete_sql(self, expression: exp.Delete) -> str:
             """
             Presto only support DELETE FROM a table, without alias, so we need to remove the unnecessary parts.
-            If the original DELETE statement contains more than one table to be deleted, just pick the first one.
+            If the original DELETE statement contains more than one table to be deleted, we can't do anything.
             """
 
             tables = expression.args.get("tables")
             if tables:
-                target_table = tables[0]
-                expression = expression.copy()
-                del expression.args["tables"]
-                expression.set("this", target_table)
+                if len(tables) > 1:
+                    return super().delete_sql(expression)
+                else:
+                    target_table = tables[0]
+                    del expression.args["tables"]
+                    expression.set("this", target_table)
 
-            table = expression.args.get("this")
+            table = expression.this
             if isinstance(table, exp.Table):
-                table_alias = table.alias
+                table_alias = table.find(exp.TableAlias)
                 if table_alias:
-                    del table.args["alias"]
-                    for column in expression.find_all(exp.Column):
-                        if column.table == table_alias:
-                            del column.args["table"]
+                    table_alias.pop()
+
+            qualified_expression = expression.transform(unqualify_columns)
+            if isinstance(qualified_expression, exp.Delete):
+                return super().delete_sql(qualified_expression)
 
             return super().delete_sql(expression)
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -638,10 +638,9 @@ class Presto(Dialect):
             if tables:
                 if len(tables) > 1:
                     return super().delete_sql(expression)
-                else:
-                    target_table = tables[0]
-                    del expression.args["tables"]
-                    expression.set("this", target_table)
+                target_table = tables[0]
+                del expression.args["tables"]
+                expression.set("this", target_table)
 
             table = expression.this
             if isinstance(table, exp.Table):
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -626,3 +626,27 @@ class Presto(Dialect):
             if kind == "VIEW" and schema.expressions:
                 expression.this.set("expressions", None)
             return super().create_sql(expression)
+
+        def delete_sql(self, expression: exp.Delete) -> str:
+            """
+            Presto only support DELETE FROM a table, without alias, so we need to remove the unnecessary parts.
+            If the original DELETE statement contains more than one table to be deleted, just pick the first one.
+            """
+
+            tables = expression.args.get("tables")
+            if tables:
+                target_table = tables[0]
+                expression = expression.copy()
+                del expression.args["tables"]
+                expression.set("this", target_table)
+
+            table = expression.args.get("this")
+            if isinstance(table, exp.Table):
+                table_alias = table.alias
+                if table_alias:
+                    del table.args["alias"]
+                    for column in expression.find_all(exp.Column):
+                        if column.table == table_alias:
+                            del column.args["table"]
+
+            return super().delete_sql(expression)
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1087,6 +1087,20 @@ LANGUAGE js AS
             },
         )
         self.validate_all(
+            "DELETE FROM db.example_table AS tb WHERE example_table.x = 1",
+            write={
+                "bigquery": "DELETE FROM db.example_table AS tb WHERE example_table.x = 1",
+                "presto": "DELETE FROM db.example_table WHERE x = 1",
+            },
+        )
+        self.validate_all(
+            "DELETE FROM db.example_table WHERE example_table.x = 1",
+            write={
+                "bigquery": "DELETE FROM db.example_table WHERE example_table.x = 1",
+                "presto": "DELETE FROM db.example_table WHERE x = 1",
+            },
+        )
+        self.validate_all(
             "SELECT * FROM a WHERE b IN UNNEST([1, 2, 3])",
             write={
                 "bigquery": "SELECT * FROM a WHERE b IN UNNEST([1, 2, 3])",
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1045,6 +1045,48 @@ LANGUAGE js AS
             },
         )
         self.validate_all(
+            "DELETE db.example_table WHERE x = 1",
+            write={
+                "bigquery": "DELETE db.example_table WHERE x = 1",
+                "presto": "DELETE FROM db.example_table WHERE x = 1",
+            },
+        )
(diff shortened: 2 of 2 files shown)


sqlglot/dialects/presto.py
```diff
@@ -626,3 +626,27 @@ def create_sql(self, expression: exp.Create) -> str:
             if kind == "VIEW" and schema.expressions:
                 expression.this.set("expressions", None)
             return super().create_sql(expression)
+
+        def delete_sql(self, expression: exp.Delete) -> str:
+            """
+            Presto only support DELETE FROM a table, without alias, so we need to remove the unnecessary parts.
+            If the original DELETE statement contains more than one table to be deleted, just pick the first one.
```
georgesittas, line 633: I'm not sure about picking the first one. We should definitely use `self.unsupported(...)`, in any case. Perhaps returning `""` is the safer choice here to avoid ignoring the remaining `DELETE`s.

cc @tobymao

sqlglot/dialects/presto.py
```diff
@@ -626,3 +626,27 @@ def create_sql(self, expression: exp.Create) -> str:
...
+            """
+
+            tables = expression.args.get("tables")
+            if tables:
+                target_table = tables[0]
+                expression = expression.copy()
+                del expression.args["tables"]
+                expression.set("this", target_table)
+
+            table = expression.args.get("this")
```
georgesittas, line 643:
```suggestion
            table = expression.this
```

sqlglot/dialects/presto.py
```diff
@@ -626,3 +626,27 @@ def create_sql(self, expression: exp.Create) -> str:
...
+            if isinstance(table, exp.Table):
+                table_alias = table.alias
+                if table_alias:
+                    del table.args["alias"]
+                    for column in expression.find_all(exp.Column):
+                        if column.table == table_alias:
+                            del column.args["table"]
```
georgesittas, line 650: I'm not sure if getting rid of the `table` like this is safe / correct. Eg. what happens if there's a db or catalog? We would probably need to do `expression.transform(transforms.unqualify_columns)` on the filter child.

sqlglot/dialects/presto.py
```diff
@@ -626,3 +626,27 @@ def create_sql(self, expression: exp.Create) -> str:
...
+        def delete_sql(self, expression: exp.Delete) -> str:
+            """
+            Presto only support DELETE FROM a table, without alias, so we need to remove the unnecessary parts.
+            If the original DELETE statement contains more than one table to be deleted, just pick the first one.
+            """
+
+            tables = expression.args.get("tables")
+            if tables:
+                target_table = tables[0]
+                expression = expression.copy()
```
georgesittas, line 639: The `copy` is unnecessary, the generator copies its input by default.

sqlglot/dialects/presto.py
tobymao: if there are multiple deletes, we should just not do anything and have presto fail

tobymao: so
 
if tables:
   return super()...

viplazylmht: if there are multiple deletes -> DELETE without the FROM keyword will be parsed into tables, but sometime the statement delete only a table, eg in bigquery: `DELETE project.dataset.table where 1=1`

So, there are multiple deletes happen when
```python
if tables and len(tables) > 1:
    return super().delete_sql()
```

sqlglot/dialects/presto.py
viplazylmht: I dont know what happens if there's a db or catalog. Can you explain it, please?

I think each column belongs to a table, so a column only have a table_alias (or table name) stored into column.table property. I will correct the table name case.

 I am understand the `transforms.unqualify_columns` will assign table alias into the column, but if the transformation work like this, it does not help anything in this case.

sqlglot/dialects/presto.py
viplazylmht: @georgesittas @tobymao I have updated the PR, thanks for the suggestion.

sqlglot/dialects/presto.py
```diff
@@ -626,3 +627,30 @@ def create_sql(self, expression: exp.Create) -> str:
...
+            """
+
+            tables = expression.args.get("tables")
+            if tables:
+                if len(tables) > 1:
+                    return super().delete_sql(expression)
+                else:
+                    target_table = tables[0]
+                    del expression.args["tables"]
+                    expression.set("this", target_table)
```
georgesittas, line 644:
```suggestion
                target_table = tables[0]
                del expression.args["tables"]
                expression.set("this", target_table)
```

sqlglot/dialects/presto.py
```diff
@@ -626,3 +627,30 @@ def create_sql(self, expression: exp.Create) -> str:
...
+            if isinstance(table, exp.Table):
+                table_alias = table.find(exp.TableAlias)
+                if table_alias:
+                    table_alias.pop()
+
+            qualified_expression = expression.transform(unqualify_columns)
+            if isinstance(qualified_expression, exp.Delete):
+                return super().delete_sql(qualified_expression)
+
+            return super().delete_sql(expression)
```
georgesittas, line 656: I think we can simplify this section?
```suggestion
            return super().delete_sql(expression.transform(unqualify_columns))
```

sqlglot/dialects/presto.py
georgesittas: > I dont know what happens if there's a db or catalog. Can you explain it, please?

Previously we were only popping off the column's `table`, and I was wondering if this could lead to an inconsistent state where a column that original had a fully-qualified name (`catalog`, `db`, `table`, `this`) would only get its `table` removed.

This probably can't happen for the columns of the table we're deleting from, since you'd qualify the columns of that table using only the alias, so there's no catalog/db. But it could happen for columns of other tables, eg. referenced in a subquery inside of the `WHERE` clause:

```sql
DELETE FROM db.t1 AS t1 WHERE NOT t1.c IN (SELECT db.t2.c FROM db.t2 LEFT JOIN db.t3 USING (id));
```

If you also unqualify `db.t2.c` here, that could make the subquery ambiguous, e.g. if `c` existed in both `t2` and `t3`. The current approach with `unqualify_columns` has this issue.

sqlglot/dialects/presto.py
viplazylmht: If you do that, you would fail the mypy checkstyle

sqlglot/dialects/presto.py
viplazylmht: Thanks, I got it. 

> Btw, the current approach will laso unqualify db.t2.c here, which could make the subquery ambiguous, e.g. if c existed in both t2 and t3.

Is there any function to fully qualify the catalog/db/table to the column? After that we can just remove the alias and thing should work.

sqlglot/dialects/presto.py
georgesittas: We shouldn't avoid simplifying the logic to satisfy mypy. In this case we obviously get back a `DELETE` from the transform, so we should either `# type: ignore` it or `t.cast(exp.Delete, ...)`.

sqlglot/dialects/presto.py
georgesittas: Since this is only used to aid the transpilation, which is best-effort, I think I'd just keep what you have but move the `unqualify_columns` transformation inside the branch that checks if there is an alias. We don't want to always unqualify columns, right? We only need it if the target table is aliased.

georgesittas: cc @tobymao can you also take a look?

viplazylmht: @georgesittas Should we always remove table alias, and always keep the fully qualified column?

Prototype code seems to my first version:
```python
table_alias = table.alias
if table_alias:
    del table.args["alias"]

    for column in expression.find_all(exp.Column):
        if column.table == table_alias and not column.catalog and not column.db:
            del column.args["table"]
```

georgesittas: https://github.com/tobymao/sqlglot/pull/3487

georgesittas:
Thanks a lot for these PRs @viplazylmht 👍

PR #3470, opened 2024-05-14
Fix: preserve quotes for projections produced by the eliminate_qualify rule

### LINKED ISSUE #3467: Sqlglot produces Syntax Error on rewriting QUALIFY
```python
print(
    sqlglot.parse_one(
        "SELECT tr.`User_-_iD`, tr.time_stamp, tr.__is_deleted, tr.__timestamp FROM tmp_9154874609124146377 AS tr WHERE __timestamp > CAST('2024-05-14 10:09:06.470000+00:00' AS TIMESTAMP) QUALIFY ROW_NUMBER() OVER (PARTITION BY `User_-_iD` ORDER BY __timestamp DESC) = 1",
        dialect="databricks",
    ).sql("spark")
)

```
this produces:

```sql
SELECT User_ - _iD, time_stamp, __is_deleted, __timestamp FROM (SELECT tr.`User_-_iD`, tr.time_stamp, tr.__is_deleted, tr.__timestamp, ROW_NUMBER() OVER (PARTITION BY `User_-_iD` ORDER BY __timestamp DESC) AS _w FROM tmp_9154874609124146377 AS tr WHERE __timestamp > CAST('2024-05-14 10:09:06.470000+00:00' AS TIMESTAMP)) AS _t WHERE _w = 1
```
`User_ - _iD` MUST be quoted for Spark, the rewrite should keep the quote here or just quote by default in the rewrite

aersam:


sqlglot/transforms.py
```diff
@@ -105,7 +105,9 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
                 select.replace(exp.alias_(select, alias))
                 taken.add(alias)
 
-        outer_selects = exp.select(*[select.alias_or_name for select in expression.selects])
+        outer_selects = exp.select(
+            *[exp.column(select.alias_or_name, quoted=True) for select in expression.selects]
```
georgesittas, line 109: This shouldn't preemptively quote, we should only quote if the original column name / alias was quoted.

georgesittas: My original [comment](https://github.com/tobymao/sqlglot/issues/3467#issuecomment-2110001697) may have been misleading, apologies. I didn't literally mean to set `quoted=True`, my intention was to point out that we're actually not preserving the quoting information.

georgesittas (commented):
Thanks for the PR @aersam. Left another comment, should be good to go afterwards

sqlglot/transforms.py
```diff
@@ -105,7 +105,16 @@ def eliminate_qualify(expression: exp.Expression) -> exp.Expression:
                 select.replace(exp.alias_(select, alias))
                 taken.add(alias)
 
-        outer_selects = exp.select(*[select.alias_or_name for select in expression.selects])
+        def _is_quoted(select: exp.Expression) -> bool:
+            _ident = select.args.get("alias") or select.this
+            return _ident.args.get("quoted") if _ident else False
```
georgesittas, line 110: This was intentionally returning an optional bool, the `quoted` arg differentiates between the `None` and `False` cases:
```suggestion
        def _is_quoted(select: exp.Expression) -> t.Optional[bool]:
            return (select.args.get("alias") or select.this).args.get("quoted")
```

georgesittas (approved):

PR #3472, opened 2024-05-14
Refactor(snowflake)!: set default precision / width for DECIMAL type

georgesittas:


sqlglot/dialects/snowflake.py
```diff
@@ -445,6 +446,11 @@ class Parser(parser.Parser):
             "LOCATION": lambda self: self._parse_location_property(),
         }
 
+        TYPE_CONVERTER = {
+            # https://docs.snowflake.com/en/sql-reference/data-types-numeric#number
+            exp.DataType.Type.DECIMAL: build_default_decimal_type(precision=38, scale=0),
+        }
```
VaggelisD, line 452: Snowflake docs say _`By default, precision is 38 and scale is 0`_, does this hint that these are parameterized by a system knob? 

From a quick look I didn't find such a feature, but if there is this could cause side effects i.e if the user has increased the precision/scale but SQLGLot tranpsiles to `(38, 0)`.

VaggelisD (commented):
Looks cool, is there more context as to why this is needed?

tests/dialects/test_duckdb.py
```diff
@@ -968,6 +964,32 @@ def test_cast(self):
...
+        )
+        self.validate_identity(
+            "CAST([STRUCT_PACK(a := 1)] AS STRUCT(a BIGINT)[])",
+            "CAST([{'a': 1}] AS STRUCT(a BIGINT)[])",
+        )
+
+        self.validate_all(
+            "CAST(x AS DECIMAL(38, 0))",
+            read={
+                "snowflake": "CAST(x AS NUMBER)",
```
tobymao, line 986: should we maintain identity for snowflake? so if it's a decimal(38, 0) we write number

tobymao (approved):

sqlglot/dialects/snowflake.py
georgesittas: I doubt there is a knob for that, it doesn't really make sense. They also don't mention something.

georgesittas:
> Looks cool, is there more context as to why this is needed?

Yeah, see https://tobiko-data.slack.com/archives/C044BRE5W4S/p1715356074527799

tests/dialects/test_duckdb.py
georgesittas: Not sure, I also considered this but then we'd change an explicit `DECIMAL(38, 0)` into the implicit `NUMBER` type, so we don't attain 100% identity, so I didn't feel strongly about adding complexity to preserve `NUMBER`.

I'm happy with either, let me know what you prefer and if needed I'll refactor.

PR #3473, opened 2024-05-14
Feat(snowflake): add support for CREATE TAG DDL statement

### LINKED ISSUE #3468: Parse CREATE TAG (Snowflake)
Currently parsed as a Command:

```
CREATE TAG cost_center COMMENT = 'cost_center tag';
```

Would be great to be parsed so the identifier, and comment, can be located.

https://docs.snowflake.com/en/sql-reference/sql/create-tag

georgesittas:
Fixes #3468

Reference: https://docs.snowflake.com/en/sql-reference/sql/create-tag


VaggelisD (approved):

tobymao (approved):
is this snowflake only?

PR #3485, opened 2024-05-15
feat!: Add ALTER TABLE SET

VaggelisD:
Introduce support for `ALTER TABLE ... SET ...` across first class dialects.

Design notes:
----------------
- Create an `exp.AlterSet` node belonging in `ALTER_PARSERS` that will hold the arguments across all mentioned dialects
- Create `exp.DataDeletionProperty` to parse T-SQL's `DATA_DELETION` argument
- Modify `exp.WithSystemVersioningProperty` and `exp.SerdeProperties` to optionally include the `WITH` keyword as a boolean arg
- Utilize `parse_properties` as a catch-all since many options are parsed as such already & to parse simple statements like `ALTER TABLE ... SET <opt> = <val>` as "anonymous" `exp.Property` which solves the need of more manual parsing

Docs
-----------
[T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver16) | [Postgres](https://www.postgresql.org/docs/current/sql-altertable.html) | [Spark](https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-alter-table.html) | [Hive](https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl) | [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-alter-table.html) | [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/alter-table) | [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html) | [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_table_set_options_statement)


sqlglot/parser.py
```diff
@@ -2380,6 +2420,14 @@ def _parse_row(self) -> t.Optional[exp.RowFormatSerdeProperty | exp.RowFormatDel
             return None
         return self._parse_row_format()
 
+    def _parse_serde_properties(self, with_=None) -> t.Optional[exp.SerdeProperties]:
```
georgesittas, line 2423: type hint for `with_`

sqlglot/parser.py
```diff
@@ -1862,23 +1864,58 @@ def _parse_volatile_property(self) -> exp.VolatileProperty | exp.StabilityProper
...
+        number = self._parse_number()
+        number_str = f"{number} " if number else ""
+        unit = self._parse_var(any_token=True)
+        return exp.var(f"{number_str}{unit}")
 
-        prop = self.expression(exp.WithSystemVersioningProperty)
-        if self._match(TokenType.L_PAREN):
-            self._match_text_seq("HISTORY_TABLE", "=")
-            prop.set("this", self._parse_table_parts())
+    def _parse_system_versioning_property(self, with_=False) -> exp.WithSystemVersioningProperty:
```
georgesittas, line 1874: Type hint for `with_`

sqlglot/parser.py
```diff
@@ -2380,6 +2420,14 @@ def _parse_row(self) -> t.Optional[exp.RowFormatSerdeProperty | exp.RowFormatDel
             return None
         return self._parse_row_format()
 
+    def _parse_serde_properties(self, with_=None) -> t.Optional[exp.SerdeProperties]:
+        with_ = self._match_text_seq("WITH") or with_
```
georgesittas, line 2424: Should we switch the order here?

sqlglot/parser.py
```diff
@@ -2380,6 +2420,14 @@ def _parse_row(self) -> t.Optional[exp.RowFormatSerdeProperty | exp.RowFormatDel
...
+        if not self._match(TokenType.SERDE_PROPERTIES):
+            return None
```
georgesittas, line 2426: What if you consume `WITH` and then you don't find `SERDE_PROPERTIES`? Don't we need to retreat?

sqlglot/parser.py
```diff
@@ -5916,6 +5960,41 @@ def _parse_alter_table_rename(self) -> t.Optional[exp.RenameTable | exp.RenameCo
...
+        if self._match(TokenType.L_PAREN, advance=False) or self._match_text_seq(
+            "TABLE", "PROPERTIES"
+        ):
+            alter_set.set("expressions", self._parse_wrapped_csv(self._parse_conjunction))
+        elif self._match_text_seq("FILESTREAM_ON", advance=False):
+            alter_set.set("expressions", [self._parse_conjunction()])
+        elif self._match_texts(("LOGGED", "UNLOGGED")):
+            alter_set.set("option", exp.var(self._prev.text.upper()))
+        elif self._match_text_seq("WITHOUT") and self._match_texts(("CLUSTER", "OIDS")):
+            alter_set.set("option", exp.var(f"WITHOUT {self._prev.text.upper()}"))
```
georgesittas, line 5975: I'd consider using `_parse_var_from_options` for these two and move it to the `else` branch. That way, it'll also make it easier to extend without creating clutter (ie. more branches) in the future if needed.

georgesittas (commented):
Great work @VaggelisD 👍

sqlglot/dialects/hive.py
```diff
@@ -664,3 +663,22 @@ def struct_sql(self, expression: exp.Struct) -> str:
...
+            tags = self.expressions(expression, key="tag", flat=True, sep="")
+            tags = f" TAGS {tags}" if tags else ""
+
+            return f"SET{serde}{exprs}{location}{file_format}{tags}"
+
+        def serdeproperties_sql(self, expression: exp.SerdeProperties) -> str:
+            prefix = "WITH " if expression.args.get("with") else ""
+            exprs = self.expressions(expression, flat=True, sep=", ")
+
+            return f"{prefix}SERDEPROPERTIES ({exprs})"
```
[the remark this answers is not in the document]
VaggelisD: I think not, because e.g. Spark/Databricks support the following statement:

```
ALTER TABLE table_identifier [ partition_spec] SET SERDEPROPERTIES ( key1 = val1, key2 = val2, ... )
```

This would be parsed as `exp.SerdeProperties` which is wanted, but if it's always generated back as `SET WITH SERDEPROPERTIES` that doesn't conform to the specs, right?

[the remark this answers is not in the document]
georgesittas: Got it, I missed that context. Sounds good.

sqlglot/dialects/postgres.py
```diff
@@ -594,3 +594,15 @@ def matchagainst_sql(self, expression: exp.MatchAgainst) -> str:
...
+
+        def alterset_sql(self, expression: exp.AlterSet) -> str:
+            exprs = self.expressions(expression, flat=True)
+            exprs = f"({exprs})" if exprs else ""
+
+            access_method = self.sql(expression, "access_method")
+            access_method = f"ACCESS METHOD {access_method}" if access_method else ""
+            tablespace = self.sql(expression, "tablespace")
+            tablespace = f"TABLESPACE {tablespace}" if tablespace else ""
+            option = self.sql(expression, "option")
```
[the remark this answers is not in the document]
VaggelisD: Don't think so, only one out of all these options will be generated at a time and the other options such as `access_method` require more text that was lost in the parsing. The `option` arg should be the standalone `exp.Var` which holds `LOGGED / UNLOGGED` etc, and the padding is baked in the return string.

sqlglot/generator.py
```diff
@@ -3015,6 +3020,12 @@ def renamecolumn_sql(self, expression: exp.RenameColumn) -> str:
         new_column = self.sql(expression, "to")
         return f"RENAME COLUMN{exists} {old_column} TO {new_column}"
 
+    def alterset_sql(self, expression: exp.AlterSet) -> str:
+        exprs = self.expressions(expression, flat=True)
+        exprs = exprs.lstrip(" ")
```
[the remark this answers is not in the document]
VaggelisD: This stems from T-SQLs property parsing e.g. `WithSystemVersionProperty`, `DataDeletionProperty` vs the `FILESTREAM_ON` option. The former are wrapped in `exp.Properties` which seem to be generating a leading whitespace, but the latter is wrapped in a simple `exp.Property` which prints as-is. 

I can mitigate this conflict by appending a whitespace in `exprs` only if `expression.expressions` does not have a `exp.Properties` child, if that sounds good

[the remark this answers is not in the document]
georgesittas: Why does the `Properties` node generate leading whitespace?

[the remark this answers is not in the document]
VaggelisD: Afaict because of the `self.sep()` call in the return stmt of `Generator::root_properties()`

sqlglot/parser.py
VaggelisD: That was my initial implementation but it resulted to a very small dictionary as it's only Postgres that supports these few static options; Thus, I thought that a stray options dict would create more clutter than these couple branches for that matter. Wdyt?

sqlglot/generator.py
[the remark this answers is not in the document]
georgesittas: Let's try to fix that, only `create_sql` uses `root_properties` (through `properties_sql`), so we could bubble that `sep()` call inside its body instead of pushing it down.

sqlglot/parser.py
georgesittas: Ok, let's leave it as is.

sqlglot/dialects/hive.py
```diff
@@ -664,3 +663,22 @@ def struct_sql(self, expression: exp.Struct) -> str:
...
+            serde = self.sql(expression, "serde")
+            serde = f" SERDE {serde}" if serde else ""
+            tags = self.expressions(expression, key="tag", flat=True, sep="")
+            tags = f" TAGS {tags}" if tags else ""
+
+            return f"SET{serde}{exprs}{location}{file_format}{tags}"
+
+        def serdeproperties_sql(self, expression: exp.SerdeProperties) -> str:
+            prefix = "WITH " if expression.args.get("with") else ""
+            exprs = self.expressions(expression, flat=True, sep=", ")
```
georgesittas, line 682: Redundant separator here
```suggestion
            exprs = self.expressions(expression, flat=True)
```

sqlglot/generator.py
```diff
@@ -1520,19 +1521,23 @@ def withdataproperty_sql(self, expression: exp.WithDataProperty) -> str:
...
+        retention_period = self.sql(expression, "retention_period")
+        retention_period = (
+            f"HISTORY_RETENTION_PERIOD={retention_period}" if retention_period else ""
+        )
 
-            if expression.expression:
-                data_consistency_check = self.sql(expression, "expression")
-                sql = f"{sql}, DATA_CONSISTENCY_CHECK={data_consistency_check}"
+        if this:
+            on_sql = self.func("ON", this, data_consistency or None, retention_period or None)
```
georgesittas, line 1534: We can get rid of the `or None`s and just move them at the definition site `else` branch

sqlglot/generator.py
```diff
@@ -3857,3 +3868,15 @@ def copy_sql(self, expression: exp.Copy) -> str:
...
+
+    def datadeletionproperty_sql(self, expression: exp.DataDeletionProperty) -> str:
+        on_sql = "ON" if expression.args.get("on") else "OFF"
+        filter_col = self.sql(expression, "filter_column")
+        filter_col = f"FILTER_COLUMN={filter_col}" if filter_col else ""
+        retention_period = self.sql(expression, "retention_period")
+        retention_period = f"RETENTION_PERIOD={retention_period}" if retention_period else ""
+
+        if filter_col or retention_period:
+            on_sql = self.func("ON", filter_col or None, retention_period or None)
```
georgesittas, line 3880: Ditto about `or None`

sqlglot/generator.py
```diff
@@ -3015,6 +3020,12 @@ def renamecolumn_sql(self, expression: exp.RenameColumn) -> str:
         new_column = self.sql(expression, "to")
         return f"RENAME COLUMN{exists} {old_column} TO {new_column}"
 
+    def alterset_sql(self, expression: exp.AlterSet) -> str:
+        exprs = self.expressions(expression, flat=True)
+        if not expression.find(exp.Properties):
+            exprs = f" {exprs}"
```
georgesittas, line 3026: This can be refactored if we clean up the `properties_sql` generation right?

georgesittas (approved):

sqlglot/generator.py
VaggelisD: Yes, I added this in the PR feedback to offset the strip, but if `properties_sql` is fixed in that regard it won't be needed

sqlglot/generator.py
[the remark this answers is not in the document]
VaggelisD: Hmm, I think the problem is that (1) the `root_properties` always included a `self.sep()` and (2) the `with_properties` also always included a separator through `self.seg`. 

Not sure if my solution is optimal, but I decided to remove both these preemptive separators and added a check that will sep `root_properties` from `with_properties` if these aren't split by a newline

sqlglot/generator.py
```diff
@@ -1328,13 +1328,17 @@ def properties_sql(self, expression: exp.Properties) -> str:
...
                 root_properties.append(p)
 
-        return self.root_properties(
-            exp.Properties(expressions=root_properties)
-        ) + self.with_properties(exp.Properties(expressions=with_properties))
+        root_props = self.root_properties(exp.Properties(expressions=root_properties))
+        with_props = self.with_properties(exp.Properties(expressions=with_properties))
+
+        if root_props and with_props and not self.pretty:
+            with_props = " " + with_props
```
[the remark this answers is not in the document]
VaggelisD: Currently on this PR we generate `with_properties` using a prefix wrapped in `self.seg(..., sep="")`, which will generate a newline between it and `root_properties` if we're on `self.pretty` mode. 

However, this PR also turned off the preemptive whitespaces of both `root_properties` and `with_properties`, which means that if `self.pretty = False` and we have properties of both types, we'd have to fix their concatenation by injecting that (missing) whitespace. Is too much of a headache?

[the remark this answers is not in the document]
georgesittas: Got it, I missed the `seg` part when I was writing this. Makes sense.

georgesittas (approved):

PR #3486, opened 2024-05-15
Fix(tsql)!: parse TININT into UTINYINT to improve transpilation

georgesittas:
Fixes #3469

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -453,9 +453,12 @@ class TSQL(Dialect):
             "DATETIMEOFFSET": TokenType.TIMESTAMPTZ,
             "DECLARE": TokenType.COMMAND,
             "EXEC": TokenType.COMMAND,
+            "FOR SYSTEM_TIME": TokenType.TIMESTAMP_SNAPSHOT,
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -526,6 +526,12 @@ class Parser(parser.Parser):
             *parser.Parser.TYPE_TOKENS,
         }
 
+        TYPE_CONVERTER = {
+            # TINYINT is a 1-byte unsigned integer
+            # https://learn.microsoft.com/en-us/sql/t-sql/data-types/int-bigint-smallint-and-tinyint-transact-sql?view=sql-server-ver16
+            exp.DataType.Type.TINYINT: lambda _: exp.DataType(this=exp.DataType.Type.UTINYINT),
```
tobymao, line 532: why do you need the type_converter, can't we use the type mapping?

georgesittas: Which type mapping? Tokenizer?

tobymao: yea just map the tinyint token for tsql to uint

georgesittas: done

PR #3487, opened 2024-05-15
Refactor(presto): make DELETE transpilation more robust

georgesittas:
Refactors https://github.com/tobymao/sqlglot/pull/3466

cc @viplazylmht

### DIFF
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1167,7 +1167,14 @@ LANGUAGE js AS
             "DELETE FROM db.example_table WHERE example_table.x = 1",
             write={
                 "bigquery": "DELETE FROM db.example_table WHERE example_table.x = 1",
-                "presto": "DELETE FROM db.example_table WHERE x = 1",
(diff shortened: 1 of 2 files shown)


tests/dialects/test_bigquery.py
```diff
@@ -1167,7 +1167,14 @@ def test_bigquery(self):
             "DELETE FROM db.example_table WHERE example_table.x = 1",
             write={
                 "bigquery": "DELETE FROM db.example_table WHERE example_table.x = 1",
-                "presto": "DELETE FROM db.example_table WHERE x = 1",
+                "presto": "DELETE FROM db.example_table WHERE example_table.x = 1",
```
georgesittas, line 1170: FYI I tried running the following statement in Presto and I didn't get a resolution / syntax error, so I assumed qualifying columns like this is supported:

```
presto> delete from memory.default.t where t.c = 1;
Query 20240515_170905_00013_cefd5 failed: This connector does not support updates or deletes
```

PR #3496, opened 2024-05-17
fix: copy into pretty printing and default dialect

tobymao:


tests/dialects/test_snowflake.py
```diff
@@ -1882,6 +1882,56 @@ def test_copy(self):
...
+FROM EXTRA.EXAMPLE.TABLE
+CREDENTIALS = (x) WITH (
+  STORAGE_INTEGRATION S3_INTEGRATION,
+  FILE_FORMAT = (TYPE = CSV COMPRESSION = NONE NULL_IF = (
+    ''
+  ) FIELD_OPTIONALLY_ENCLOSED_BY = '"'),
+  HEADER TRUE,
+  OVERWRITE TRUE,
+  SINGLE TRUE
+)""",
```
VaggelisD, line 1906: Is there a purpose for testing the transpilation from Snowflake to the base dialect other than pretty print? 

The base dialect converted to T-SQL/DuckDB/Postgres format (which have the 'WITH' syntax), and generally COPY options from one dialect will hardly match to other dialects, so the base dialect query does not conform to any specs

VaggelisD (approved):

georgesittas (approved):

tests/dialects/test_snowflake.py
tobymao: the base dialect spec needs to work properly, in main it crashes

tobymao: it crashed

Issue #3505, opened 2024-05-18
Lambda expression with data types in Snowflake dialect


krzysztof-kwitt:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`
- Check if the issue still exists on main

**Fully reproducible code snippet**
Please include a fully reproducible code snippet or the input sql, dialect, and expected output.

Lambda expression with type definition are incorrectly handled.
```
import sqlglot
sql = 'SELECT TRANSFORM([1, 2, 3], a INT -> a * 2) AS "Multiply by Two"'
sqlglot.parse_one(sql)
```

Without type definitions query is parsed correctly.
```
SELECT TRANSFORM([1, 2, 3], a -> a * 2) AS "Multiply by Two";
```
```
Select(
  expressions=[
    Alias(
      this=Transform(
        this=Array(
          expressions=[
            Literal(this=1, is_string=False),
            Literal(this=2, is_string=False),
            Literal(this=3, is_string=False)]),
        expression=Lambda(
          this=Mul(
            this=Identifier(this=a, quoted=False),
            expression=Literal(this=2, is_string=False)),
          expressions=[
            Identifier(this=a, quoted=False)])),
      alias=Identifier(this=Multiply by Two, quoted=True))])
```

For details, see:
https://docs.snowflake.com/en/sql-reference/functions/transform#multiply-each-element-in-an-array-by-a-value

**Official Documentation**
Please include links to official SQL documentation related to your issue.

PR #3506, opened 2024-05-18
feat: add support for snowflake lambdas with type annotations closes …

tobymao:
…#3505

Snowflake allows for arguments in lambdas to have a type annotation. This is equivalent to casting the values in the lambda.

https://docs.snowflake.com/en/sql-reference/functions/transform#multiply-each-element-in-an-array-by-a-value

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -580,7 +580,7 @@ class Parser(metaclass=_Parser):
             exp.Lambda,
             this=self._replace_lambda(
                 self._parse_conjunction(),
-                {node.name for node in expressions},
+                expressions,
             ),
             expressions=expressions,
         ),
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -6433,14 +6436,36 @@ def _match_text_seq(self, *texts, advance=True):
...
+
+            lambda_types[this.name] = typ
+
         for column in node.find_all(exp.Column):
-            if column.parts[0].name in lambda_variables:
+            name = column.parts[0].name
+            if name in lambda_types:
                 dot_or_id = column.to_dot() if column.table else column.this
+
+                typ = lambda_types.get(name)
```
georgesittas, line 6460:
```suggestion
            typ = lambda_types.get(column.parts[0].name)
            if typ is not None:
                dot_or_id = column.to_dot() if column.table else column.this
```

georgesittas (approved):

sqlglot/parser.py
tobymao: are you sure this is right? we need to convert to a dot even if there is no typ

PR #3508, opened 2024-05-18
Fix(databricks)!: parse : operator as JSONExtract (similar to Snowflake)

georgesittas:
This fixes the JSON extraction generator for Databricks. I noticed that various test cases were broken, so this must have not been tested properly.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1160,6 +1160,9 @@ class Parser(metaclass=_Parser):
     # Whether the -> and ->> operators expect documents of type JSON (e.g. Postgres)
     JSON_ARROWS_REQUIRE_JSON_TYPE = False
 
+    # Whether the `:` operator is used to extract a value from a JSON document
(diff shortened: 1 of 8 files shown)


sqlglot/parser.py
```diff
@@ -4183,9 +4188,11 @@ def _parse_type(self, parse_interval: bool = True) -> t.Optional[exp.Expression]
                 if parser:
                     return parser(self, this, data_type)
                 return self.expression(exp.Cast, this=this, to=data_type)
+
             if not data_type.expressions:
                 self._retreat(index)
-                return self._parse_column()
+                return self._parse_id_var() if fallback_to_identifier else self._parse_column()
```
georgesittas, line 4194: I changed this because a databricks test case was failing:

```sql
CREATE TABLE t (c STRUCT<interval: DOUBLE COMMENT 'aaa'>)
```

We'd parse `interval: DOUBLE` into a `JSONExtract` expression because of the colon struct def. syntax.

Issue #3502, opened 2024-05-17
Better support for Snowflake `WAREHOUSE` clauses?


yingw787:
**Is your feature request related to a problem? Please describe.**

I'm working on a Snowflake thing, and pretty extensively using `sqlglot`. I'm running into an issue where parsing out WAREHOUSE queries is always falling back into `exp.Command`, and I'd really like it to be parsed like other supported tokens, such as DATABASE or FUNCTION. For example:

```python
In [5]: parse_one("CREATE OR REPLACE DATABASE IDENTIFIER(?)", dialect="snowflake")
Out[5]: 
Create(
  this=Table(
    this=Anonymous(
      this=IDENTIFIER,
      expressions=[
        Placeholder()])),
  kind=DATABASE,
  replace=True)
```

vs.

```python
In [6]: parse_one("CREATE OR REPLACE WAREHOUSE IDENTIFIER(?)", dialect="snowflake")
'CREATE OR REPLACE WAREHOUSE IDENTIFIER(?)' contains unsupported syntax. Falling back to parsing as a 'Command'.
Out[6]: Command(this=CREATE, expression=OR REPLACE WAREHOUSE IDENTIFIER(?))
```

**Describe the solution you'd like**

For the above example, I'd really like if I could get this result:

```python
In [5]: parse_one("CREATE OR REPLACE WAREHOUSE IDENTIFIER(?)", dialect="snowflake")
Out[5]: 
Create(
  this=Table(
    this=Anonymous(
      this=IDENTIFIER,
      expressions=[
        Placeholder()])),
  kind=WAREHOUSE,
  replace=True)
```

This, but for different types of queries, such as `DESCRIBE WAREHOUSE`, `DROP WAREHOUSE`, and `SHOW WAREHOUSES`.

**Describe alternatives you've considered**

Right now I'm catching these queries in my code and then handrolling a `sqlglot.exp` given the query that I see.

**Additional context**

This is my `pip freeze`:

```bash
sqlglot==23.15.9
sqlglotrs==0.2.5
```

I am using Ubuntu 22.04 LTS, this is my `lsb_release -a`:

```bash
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.4 LTS
Release:        22.04
Codename:       jammy
```

I'm happy to make an upstream contribution if this is moreso a backlog item for the `sqlglot` core team - I contributed #204 back in the day. Just let me know.

georgesittas:
[...]

PR #3510, opened 2024-05-19
[WIP] feat(gh-3502): Add Snowflake support for basic `CREATE WAREHOUSE` statements

yingw787:
Add basic Snowflake support for `CREATE SNOWFLAKE` and re-learning `sqlglot` contributing.

Will likely add support in a follow-up PR for `CREATE OR REPLACE WAREHOUSE` and `CREATE WAREHOUSE IF NOT EXISTS` as I need those too, but wanted to get this PR in sooner rather than later.

- [x] Update `snowflake.py`

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -855,6 +855,7 @@ class Snowflake(Dialect):
             ),
             exp.UnixToTime: rename_func("TO_TIMESTAMP"),
             exp.VarMap: lambda self, e: var_map_sql(self, e, "OBJECT_CONSTRUCT"),
+            exp.Warehouse: lambda self, e: e.this,
             exp.WeekOfYear: rename_func("WEEKOFYEAR"),
             exp.Xor: rename_func("BOOLXOR"),
         }
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -740,6 +740,7 @@ class Snowflake(Dialect):
             "TAG": TokenType.TAG,
             "TIMESTAMP_TZ": TokenType.TIMESTAMPTZ,
             "TOP": TokenType.TOP,
+            "WAREHOUSE": TokenType.WAREHOUSE,
         }
 
         SINGLE_TOKENS = {
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3049,6 +3049,9 @@ class Table(Expression):
         return col
 
 
+class Warehouse(Expression):
+    pass
+
 class Union(Query):
     arg_types = {
         "with": False,
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3050,7 +3050,10 @@ class Table(Expression):
 
 
 class Warehouse(Expression):
-    pass
+    arg_types = {
+        "this": False,
+    }
+
 
 class Union(Query):
     arg_types = {
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1710,7 +1710,8 @@ class Parser(metaclass=_Parser):
                 if self._match_text_seq("WITH", "NO", "SCHEMA", "BINDING"):
                     no_schema_binding = True
             elif create_token.token_type == TokenType.WAREHOUSE:
-                this = exp.Warehouse(**this.args)
+                this = self._parse_warehouse(identifier=table_parts.this)
+                extend_props(self._parse_properties())
 
             shallow = self._match_text_seq("SHALLOW")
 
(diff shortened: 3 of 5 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -740,6 +740,7 @@ class Tokenizer(tokens.Tokenizer):
             "TAG": TokenType.TAG,
             "TIMESTAMP_TZ": TokenType.TIMESTAMPTZ,
             "TOP": TokenType.TOP,
+            "WAREHOUSE": TokenType.WAREHOUSE,
```
georgesittas, line 743: Can we double check if other dialects support `CREATE WAREHOUSE ...` too? If so, it'd make sense to have this in the base tokenizer instead.

sqlglot/expressions.py
```diff
@@ -3049,6 +3049,12 @@ def to_column(self, copy: bool = True) -> Alias | Column | Dot:
         return col
 
 
+class Warehouse(Expression):
+    arg_types = {
+        "this": False,
+    }
+
```
georgesittas, line 3056: We shouldn't create a new expression for this, `exp.Create` suffices. As long as you add the token to the `DB_CREATABLES`, it should automatically handle the rest. E.g.

```python
>>> import sqlglot
>>> sqlglot.parse_one("create table x")
Create(
  this=Table(
  kind=TABLE)
>>> sqlglot.parse_one("create table x").sql()
'CREATE TABLE x'
```

Let me know if this is more nuanced.

sqlglot/parser.py
```diff
@@ -1708,6 +1709,9 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
             elif create_token.token_type == TokenType.VIEW:
                 if self._match_text_seq("WITH", "NO", "SCHEMA", "BINDING"):
                     no_schema_binding = True
+            elif create_token.token_type == TokenType.WAREHOUSE:
+                this = self._parse_warehouse(identifier=table_parts.this)
+                extend_props(self._parse_properties())
```
georgesittas, line 1714: We don't need this

georgesittas (commented):
Thanks for the PR @yingw787, left some comments.

sqlglot/dialects/snowflake.py
yingw787: So I ran a prompt through Perplexity.ai / ChatGPT, and I checked GCP BigQuery (uses concept of slots), Databricks (no CREATE WAREHOUSE clause), AWS Redshift (basically PostgreSQL with some addenda), Azure Synapse Analytics (didn't find anything), Teradata (nope), and SAP SybaseIQ (nope). In addition I know the `CREATE WAREHOUSE` BNF grammar in Snowflake can become fairly specific, you can specify parameters for a given warehouse inline during warehouse creation.

sqlglot/expressions.py
yingw787: I'm not quite sure I understand. When I remove `exp.Warehouse`, it does fall back to `exp.Table` correctly, but a warehouse is pretty different from a table.

My understanding of a Snowflake warehouse is it's a resource abstraction https://docs.snowflake.com/en/sql-reference/sql/create-warehouse like `CREATE CLUSTER` in Oracle: https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/CREATE-CLUSTER.html.

Is `exp.Table` meant to be overloaded with non-table-specific attributes, such as `MAX_CLUSTER_COUNT` and `AUTO_SUSPEND`, as long as `kind='WAREHOUSE'`? If so, I'm happy to make that change.

yingw787: Actually, given the time constraints, I'm happy to just push up this change. Switched over to `exp.Table` with kind `WAREHOUSE` instead of using `exp.Warehouse`.

sqlglot/parser.py
yingw787: This should be reflected in latest commit.

sqlglot/expressions.py
georgesittas: My suggestion was to simply add `WAREHOUSE` in `DB_CREATABLES`. Then, `_parse_create` would parse that into a `Create` node that contains a `WAREHOUSE` kind, and any additional parsed properties.

georgesittas: Yeah, I see what you mean - makes sense. Your refactor is what I had in mind, but the name being parsed into a `Table` is something I missed.

We can amend this by creating a parser class constant set in Snowflake that contains those "creatables" which we want to convert from `Table`s into `Identifier`s. We already [do this](https://github.com/tobymao/sqlglot/blob/74311c92091ab28deaaa6a78f9e0d0337e7d957a/sqlglot/dialects/snowflake.py#L500) for `CREATE TAG`, for the same reason.

georgesittas (approved):
We can either convert the `Table` to an `Identifier` in this PR, or I can implement it afterwards.

georgesittas:
https://github.com/tobymao/sqlglot/commit/33ab3536d68203f4fceee63507b5c73076d48ed7

PR #3518, opened 2024-05-21
chore!!: remove dataframe library and tests in favor of sqlframe

tobymao:


eakmanrq (approved):
Enjoy having fast tests again!

PR #3527, opened 2024-05-22
Fix(doris): add more reserved keywords

Toms1999:


Toms1999:
fix#3526

Toms1999 (commented):
fix git version

georgesittas:
@Toms1999 can you please share some docs on this? From what I can tell most of these are also reserved in MySQL, so we could do:

```python
RESERVED_KEYWORDS = {
    **MySQL.Generator.RESERVED_KEYWORDS,
    ...
}
```

Toms1999:
@georgesittas ,yes, most of them are indeed in mysql, but a small part is unique to doris, that is to say, doris cannot be executed, but mysql can, such as cast and sets, so I had to first take all the keywords of mysql and then add the relevant keywords of doris. In addition, there is no relevant document here and the official does not provide it, which is found by our large number of transfer tests

Toms1999:
@georgesittas So, should we just add it to mysql, or should doris rewrite it?

georgesittas:
Check out https://github.com/tobymao/sqlglot/commit/f25a9cdfdfd68b25d519ce0b670e2393737c502d

PR #3531, opened 2024-05-22
feat(optimizer)!: Expand stars on BigQuery's tbl.struct_col.* selections

### LINKED ISSUE #3484: Expanding nested star selects gives an error in Bigquery syntax
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`: bigquery dialect
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`: Not applicable
- Check if the issue still exists on main: Couldn't find anything related

```python
sql_string = """
SELECT
  foo.bar.*
FROM
  foo
"""

import sqlglot

ast = sqlglot.parse_one(sql_string, dialect="bigquery")
schema = ensure_schema(None, dialect="bigquery")

qualified = qualify_columns(
  ast,
  schema,
  expand_alias_refs=True,
  expand_stars=True,
  infer_schema=None,
)

```

The qualify function throws an error:
`sqlglot.errors.OptimizeError: Unknown table: bar`

**Official Documentation**
[Bigquery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_expression_star)

VaggelisD:
Fixes #3484

BigQuery supports star expansion such as `foo.bar.*` where `foo` is the table and `bar` is a struct. This is essentially a  _flatten_ operation from the specified struct and down. For example, if `bar` has a nested struct `baz`, then `foo.bar.baz.*` would cause a flatten from `baz` downwards.

In order to replicate this behavior, a DFS traversal starting from the top-level struct gathers the non-struct fields and enqueues the nested structs (if any). At every point, a name / fully qualified path is kept and prefixes the fields at that level. 

**Note**: If any nested field is _not_ aliased, then the star expansion is stopped and the star` is preserved; This is because BigQuery will alias these fields dynamically so it's not possible to qualify/target these fields statically afaict.

An example:

```
Schema
-----------
foo: 'bar' STRUCT<'f0' STRING, 'baz' STRUCT<'f1' STRING>>

Query
--------
SELECT foo.bar.* FROM foo;

Star Expansion
------------
1. Enqueue the top level struct which is bar. Current path is initiazed as empty string
2. Pop the stack's top item and (1) qualify it with the current path name if it's a field or (2) append it to the stack if it's a nested struct.
3. Pop the next item if any, repeat 2

Result
-------
SELECT foo.bar.f0, foo.bar.baz.f1 FROM foo;
```

Docs
--------
- [BigQuery Star Selection](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_expression_star)

### DIFF
--- a/sqlglot/__init__.py
+++ b/sqlglot/__init__.py
@@ -65,9 +65,6 @@ except ImportError:
 pretty = False
 """Whether to format generated SQL by default."""
 
-schema = MappingSchema()
-"""The default schema used by SQLGlot (e.g. in the optimizer)."""
-
 
 def tokenize(sql: str, read: DialectType = None, dialect: DialectType = None) -> t.List[Token]:
     """
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -550,7 +550,7 @@ def _expand_stars(
                 tables = [expression.table]
                 _add_except_columns(expression.this, tables, except_columns)
                 _add_replace_columns(expression.this, tables, replace_columns)
-            elif dialect == "bigquery":
+            elif is_bigquery:
                 struct_fields = _expand_struct_stars(expression, next_alias_name)
                 if struct_fields:
                     new_selections.extend(struct_fields)
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -53,7 +53,7 @@ def qualify_columns(
     infer_schema = schema.empty if infer_schema is None else infer_schema
     dialect = Dialect.get_or_raise(schema.dialect)
     pseudocolumns = dialect.PSEUDOCOLUMNS
-    next_alias_name = name_sequence("_field_")
+    next_alias_name = name_sequence("_f_")
 
     for scope in traverse_scope(expression):
         resolver = Resolver(scope, schema, infer_schema=infer_schema)
@@ -472,7 +472,7 @@ def _expand_struct_stars(
         ]
 
         if not fld.kind.is_type(exp.DataType.Type.STRUCT):
-            # Collect the non-struct columns
+            # Collect the nested non-struct fields
             fld_keys.append(fld_id)
         else:
             stack.extend((expr, fld_id) for expr in reversed(fld.kind.expressions))
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -483,14 +483,14 @@ def _expand_stars(
             table_name = struct_col.table if isinstance(struct_col, exp.Column) else None
             struct_fields: t.List[str] = []
 
-            queue = [
+            stack = [
                 (col, "")
                 for col in scope_cols
                 if col.is_type(exp.DataType.Type.STRUCT) and col.table == table_name
             ]
 
-            while queue:
-                col, path_name = queue.pop()
+            while stack:
+                col, path_name = stack.pop()
                 if not col or not isinstance(col.this, exp.Identifier):
                     struct_fields = []
                     break
@@ -498,33 +498,21 @@ def _expand_stars(
                 path_name = col.this.this if not path_name else f"{path_name}.{col.this.this}"
                 datatype = col.find(exp.DataType) or (col.type and col.type.find(exp.DataType))
 
-                if not datatype or not datatype.expressions:
+                if not datatype:
                     struct_fields = []
                     break
 
-                error = False
+                if datatype.this != exp.DataType.Type.STRUCT:
+                    full_name = f"{table_name}.{path_name}"
+                    if not starting_struct or starting_struct in full_name:
+                        # Collect the non-struct columns
+                        if not isinstance(col.this, exp.Identifier):
+                            struct_fields = []
+                            break
+                        struct_fields.append(path_name)
 
-                for expr in datatype.expressions:
-                    if error:
-                        break
-
-                    expr_type = expr.find(exp.DataType)
-                    if expr_type.this == exp.DataType.Type.STRUCT:
-                        annotator._maybe_annotate(expr)
-                        queue.append((expr, path_name))
-
-                    else:
-                        full_name = f"{table_name}.{path_name}"
-                        if not starting_struct or starting_struct in full_name:
-                            # Collect the non-struct columns
-                            if not isinstance(expr.this, exp.Identifier):
-                                error = True
-                                break
-                            struct_fields.append(f"{path_name}.{expr.this.this}")
-
-                if error:
-                    struct_fields = []
-                    break
+                for expr in reversed(datatype.expressions):
+                    stack.append((expr, path_name))
 
             for name in struct_fields:
                 alias_ = next_alias_name()
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -417,29 +417,35 @@ def _expand_struct_stars(
     """[BigQuery] Expand/Flatten foo.bar.* where bar is a struct column"""
 
     dot_column = t.cast(exp.Column, expression.find(exp.Column))
-    if not dot_column or not dot_column.type:
+    if not dot_column.is_type(exp.DataType.Type.STRUCT):
         return []
+
+    # All nested struct values are ColumnDefs, so normalize the first exp.Column in one
+    dot_column = dot_column.copy()
+    starting_struct = exp.ColumnDef(this=dot_column.this, kind=dot_column.type)
 
     # First part is the table name and last part is the star so they can be dropped
     dot_parts = expression.parts[1:-1]
 
-    # All nested struct values are ColumnDefs, so normalize the first exp.Column in one
-    starting_struct = exp.ColumnDef(this=dot_column.this, kind=dot_column.type)
-
     # If we're expanding a nested struct (eg. t.c.f1.f2.*), start the DFS traversal from that struct
     for part in dot_parts[1:]:
-        for expr in t.cast(exp.DataType, starting_struct.kind).expressions:
+        struct_fields = t.cast(exp.DataType, starting_struct.kind).expressions
+
+        if not struct_fields:
+            return []
+
+        for fld in struct_fields:
             # Unable to expand star unless all struct fields are named
-            if not isinstance(expr.this, exp.Identifier):
+            if not isinstance(fld.this, exp.Identifier):
                 return []
 
-            if expr.name == part.name and expr.kind.this == exp.DataType.Type.STRUCT:
-                starting_struct = expr
+            if fld.name == part.name and fld.kind.this == exp.DataType.Type.STRUCT:
+                starting_struct = fld
 
     if starting_struct.name != dot_parts[-1].name:
-        raise OptimizeError(f"Unknown struct: {starting_struct.name}")
+        return []
 
-    # Each nested field caches it's path in exp.Identifier parts e.g. tbl.f0.f1.item -> [exp.Identifier(tbl), ..., exp.Identifier(item)]
+    # Each nested field caches its path in exp.Identifier parts e.g. tbl.f0.f1.item -> [exp.Identifier(tbl), ..., exp.Identifier(item)]
     nested_fields: t.Dict[int, t.List[exp.Identifier]] = {id(starting_struct): []}
 
     # The keys to non-struct field entries in `nested_fields`
@@ -449,25 +455,26 @@ def _expand_struct_stars(
 
     # Traverse & flatten the starting struct
     while stack:
-        col, parent_struct_id = stack.pop()
-        id_col = id(col)
+        fld, parent_struct_id = stack.pop()
 
         # Unable to expand star unless all struct fields are named and annotated
-        if not isinstance(col.this, exp.Identifier) or not col.kind:
+        if not isinstance(fld.this, exp.Identifier) or not fld.kind:
             return []
 
+        id_fld = id(fld)
+
         # Preserve the qualified path for this field
-        nested_fields[id_col] = list(
+        nested_fields[id_fld] = list(
             itertools.chain(
-                [part.copy() for part in nested_fields[parent_struct_id]], [col.this.copy()]
+                [part.copy() for part in nested_fields[parent_struct_id]], [fld.this.copy()]
             )
         )
 
-        # Collect the non-struct columns
-        if col.kind.this != exp.DataType.Type.STRUCT:
-            nested_field_keys.append(id_col)
+        if not fld.kind.is_type(exp.DataType.Type.STRUCT):
+            # Collect the non-struct columns
+            nested_field_keys.append(id_fld)
         else:
-            stack.extend((expr, id(col)) for expr in reversed(col.kind.expressions))
+            stack.extend((expr, id_fld) for expr in reversed(fld.kind.expressions))
 
     new_selections = []
     for nested_fields_key in nested_field_keys:
@@ -527,7 +534,8 @@ def _expand_stars(
             if not pivot_output_columns:
                 pivot_output_columns = [c.alias_or_name for c in pivot.expressions]
 
-    if dialect == "bigquery" and any(isinstance(col, exp.Dot) for col in scope.stars):
+    is_bigquery = dialect == "bigquery"
+    if is_bigquery and any(isinstance(col, exp.Dot) for col in scope.stars):
         # Found struct expansion, annotate scope ahead of time
         annotator.annotate_scope(scope)
 
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -22,7 +22,6 @@ def qualify_columns(
     expand_alias_refs: bool = True,
     expand_stars: bool = True,
     infer_schema: t.Optional[bool] = None,
-    dialect: t.Optional[DialectType] = None,
 ) -> exp.Expression:
     """
     Rewrite sqlglot AST to have fully qualified columns.
@@ -52,7 +51,7 @@ def qualify_columns(
     schema = ensure_schema(schema)
     annotator = TypeAnnotator(schema)
     infer_schema = schema.empty if infer_schema is None else infer_schema
-    dialect = Dialect.get_or_raise(dialect or schema.dialect)
+    dialect = Dialect.get_or_raise(schema.dialect)
     pseudocolumns = dialect.PSEUDOCOLUMNS
     next_alias_name = name_sequence("_field_")
 
@@ -349,7 +348,10 @@ def _convert_columns_to_dots(scope: Scope, resolver: Resolver) -> None:
     """
     converted = False
     for column in itertools.chain(scope.columns, scope.stars):
-        column_table = column.table
+        if isinstance(column, exp.Dot):
+            continue
+
+        column_table: t.Optional[str | exp.Identifier] = column.table
         if (
             column_table
             and column_table not in scope.sources
@@ -409,6 +411,47 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
                     column.set("table", column_table)
 
 
+def _expand_struct_stars(
+    scope: Scope, structs: t.List[t.Tuple[t.Optional[exp.Column], str]]
+) -> t.Optional[t.List[str]]:
+    struct_fields: t.List[str] = []
+
+    while structs:
+        # [BigQuery] Expand/Flatten foo.bar.* where bar is a struct column
+        struct_col, starting_struct = structs.pop()
+        table_name = struct_col.table if isinstance(struct_col, exp.Column) else None
+
+        stack = [
+            (col, "")
+            for col in scope.columns
+            if col.is_type(exp.DataType.Type.STRUCT) and col.table == table_name
+        ]
+
+        while stack:
+            col, path_name = stack.pop()
+            if not col or not isinstance(col.this, exp.Identifier):
+                return []
+
+            path_name = col.this.this if not path_name else f"{path_name}.{col.this.this}"
+            datatype = col.find(exp.DataType) or (col.type and col.type.find(exp.DataType))
+
+            if not datatype:
+                return None
+
+            # Collect the non-struct columns
+            if datatype.this != exp.DataType.Type.STRUCT:
+                full_name = f"{table_name}.{path_name}"
+                if not starting_struct or starting_struct in full_name:
+                    if not isinstance(col.this, exp.Identifier):
+                        return None
+                    struct_fields.append(path_name)
+
+            for expr in reversed(datatype.expressions):
+                stack.append((expr, path_name))
+
+    return struct_fields
+
+
 def _expand_stars(
     scope: Scope,
     resolver: Resolver,
@@ -416,7 +459,7 @@ def _expand_stars(
     pseudocolumns: t.Set[str],
     annotator: TypeAnnotator,
     next_alias_name: t.Callable[[], str],
-    dialect: t.Optional[Dialect],
+    dialect: t.Optional[DialectType],
 ) -> None:
     """Expand stars to lists of column selections"""
 
@@ -445,8 +488,9 @@ def _expand_stars(
             if not pivot_output_columns:
                 pivot_output_columns = [c.alias_or_name for c in pivot.expressions]
 
-    scope_annotated = False
-    scope_cols = scope.columns
+    if dialect == "bigquery" and any(isinstance(col, exp.Dot) for col in scope.stars):
+        # Found struct expansion, annotate scope ahead of time
+        annotator.annotate_scope(scope)
 
     for expression in scope.expression.selects:
         tables = []
@@ -455,69 +499,26 @@ def _expand_stars(
             tables = list(scope.selected_sources)
             _add_except_columns(expression, tables, except_columns)
             _add_replace_columns(expression, tables, replace_columns)
-        elif expression.is_star and not isinstance(expression, exp.Dot):
-            tables = [expression.table]
-            _add_except_columns(expression.this, tables, except_columns)
-            _add_replace_columns(expression.this, tables, replace_columns)
-
-        elif expression.is_star and (
-            isinstance(expression, exp.Dot)
-            or (isinstance(expression, exp.Column) and expression.table in scope.sources)
-        ):
-            col = expression.find(exp.Column) if isinstance(expression, exp.Dot) else expression
-            starting_struct = str(expression.this) if isinstance(expression.this, exp.Dot) else None
-            structs.append((col, starting_struct))
-
-            # Annotate scope ahead of the struct flattening
-            if not scope_annotated and dialect == "bigquery":
-                annotator.annotate_scope(scope)
-                scope_annotated = True
-
+        elif expression.is_star:
+            if not isinstance(expression, exp.Dot):
+                tables = [expression.table]
+                _add_except_columns(expression.this, tables, except_columns)
+                _add_replace_columns(expression.this, tables, replace_columns)
+            elif dialect == "bigquery":
+                col = expression.find(exp.Column)
+                table_name = col.table if col else None
+                starting_struct = str(expression.this)
+                structs.append((col, starting_struct))
         else:
             new_selections.append(expression)
             continue
 
-        while structs:
-            # [BigQuery] Expand/Flatten foo.bar.* where bar is a struct column
-            struct_col, starting_struct = structs.pop()
-            table_name = struct_col.table if isinstance(struct_col, exp.Column) else None
-            struct_fields: t.List[str] = []
+        if structs:
+            struct_fields = _expand_struct_stars(scope, structs) or []
 
-            stack = [
-                (col, "")
-                for col in scope_cols
-                if col.is_type(exp.DataType.Type.STRUCT) and col.table == table_name
-            ]
-
-            while stack:
-                col, path_name = stack.pop()
-                if not col or not isinstance(col.this, exp.Identifier):
-                    struct_fields = []
-                    break
-
-                path_name = col.this.this if not path_name else f"{path_name}.{col.this.this}"
-                datatype = col.find(exp.DataType) or (col.type and col.type.find(exp.DataType))
-
-                if not datatype:
-                    struct_fields = []
-                    break
-
-                if datatype.this != exp.DataType.Type.STRUCT:
-                    full_name = f"{table_name}.{path_name}"
-                    if not starting_struct or starting_struct in full_name:
-                        # Collect the non-struct columns
-                        if not isinstance(col.this, exp.Identifier):
-                            struct_fields = []
-                            break
-                        struct_fields.append(path_name)
-
-                for expr in reversed(datatype.expressions):
-                    stack.append((expr, path_name))
-
-            for name in struct_fields:
+            for struct_name in struct_fields:
+                root, *parts = struct_name.split(".")
                 alias_ = next_alias_name()
-                root, *parts = name.split(".")
-
                 new_column: exp.Column | exp.Dot = exp.column(root, table=table_name)
 
                 if parts:
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -79,7 +79,6 @@ def qualify_columns(
                     pseudocolumns,
                     annotator,
                     next_alias_name,
-                    dialect,
                 )
             qualify_outputs(scope)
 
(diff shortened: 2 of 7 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -422,10 +459,86 @@ def _expand_stars(
...
+            or (isinstance(expression, exp.Column) and expression.table in scope.sources)
+        ):
+            col = expression.find(exp.Column) if isinstance(expression, exp.Dot) else expression
+            starting_struct = str(expression.this) if isinstance(expression.this, exp.Dot) else None
+            structs.append((col, starting_struct))
+
+            # Annotate scope ahead of the struct flattening
+            if not scope_annotated and dialect == "bigquery":
+                annotator.annotate_scope(scope)
+                scope_annotated = True
```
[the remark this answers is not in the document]
georgesittas: Can we use another signal to conditionally annotate here instead of adding this flag-controlled branch inside the loop? Perhaps check if there's anything in `scope.stars` and if it's a dot, then ...

VaggelisD:
Changed to DFS so as to match BQ's flatten order (left to right), just in case the previous order added side effects

sqlglot/optimizer/qualify_columns.py
```diff
@@ -51,8 +52,9 @@ def qualify_columns(
     schema = ensure_schema(schema)
     annotator = TypeAnnotator(schema)
     infer_schema = schema.empty if infer_schema is None else infer_schema
-    dialect = Dialect.get_or_raise(schema.dialect)
+    dialect = Dialect.get_or_raise(dialect or schema.dialect)
```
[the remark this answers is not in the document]
VaggelisD: In case there's no schema such as the CTE based tests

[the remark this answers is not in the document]
VaggelisD: I think during my implementation I'd have `optimize(..., schema = None)` and the default constructed schema would either have `None` or the base dialect, so not wanting to break any default value I chose to propagate this from `qualify`. Will investigate this a bit again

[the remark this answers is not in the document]
georgesittas: Yep, I think if this is only used to amend some unit test failures, then we should instead look at those and not change the API just yet.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -422,10 +459,86 @@ def _expand_stars(
             tables = [expression.table]
             _add_except_columns(expression.this, tables, except_columns)
             _add_replace_columns(expression.this, tables, replace_columns)
+
+        elif expression.is_star and (
+            isinstance(expression, exp.Dot)
+            or (isinstance(expression, exp.Column) and expression.table in scope.sources)
+        ):
```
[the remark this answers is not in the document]
VaggelisD: Good catch, I think this is dead code from before the rebase on your star branch, where you could have struct references be both `exp.Column` and `exp.Dot`. Now, for any struct star expansions we can only find `exp.Dot` right?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -422,10 +459,74 @@ def _expand_stars(
...
+            if not scope_annotated and dialect == "bigquery":
+                annotator.annotate_scope(scope)
+                scope_annotated = True
+
         else:
             new_selections.append(expression)
             continue
 
+        while structs:
+            # [BigQuery] Expand/Flatten foo.bar.* where bar is a struct column
```
georgesittas, line 481: Should we factor this whole thing out in a separate helper? `_expand_stars` is starting to look quite complicated.

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
georgesittas: Yeah I think so

sqlglot/optimizer/qualify_columns.py
```diff
@@ -422,10 +459,86 @@ def _expand_stars(
             tables = [expression.table]
             _add_except_columns(expression.this, tables, except_columns)
             _add_replace_columns(expression.this, tables, replace_columns)
+
+        elif expression.is_star and (
+            isinstance(expression, exp.Dot)
+            or (isinstance(expression, exp.Column) and expression.table in scope.sources)
+        ):
+            col = expression.find(exp.Column) if isinstance(expression, exp.Dot) else expression
+            starting_struct = str(expression.this) if isinstance(expression.this, exp.Dot) else None
```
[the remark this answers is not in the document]
VaggelisD: Iirc this was to flatten the entire `exp.Dot` chain into one str representation. So, if I want to expand `tbl.str1.str2.str3.*`, I could define `starting_struct = tbl.str1.str2.str3`

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
VaggelisD: Will look into it, bear in mind that you could have more expansions in the same scope as with the added test of 2 tables, so this was the safest condition until now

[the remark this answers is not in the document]
georgesittas: Can you explain this a bit more? Not sure I get why it'd be unsafe.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -413,7 +445,12 @@ def _expand_stars(
             if not pivot_output_columns:
                 pivot_output_columns = [c.alias_or_name for c in pivot.expressions]
 
+    scope_annotated = False
+    scope_cols = scope.columns
+
     for expression in scope.expression.selects:
+        tables = []
```
georgesittas, line 452: This seems like dead code

georgesittas: Hm I guess not actually, but let's not initialize this here. Instead, move it to the new `elif` branch.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -22,6 +22,7 @@ def qualify_columns(
     expand_alias_refs: bool = True,
     expand_stars: bool = True,
     infer_schema: t.Optional[bool] = None,
+    dialect: t.Optional[DialectType] = None,
```
tobymao, line 25: delete this, the dialect is already a part of schema

tests/test_optimizer.py
```diff
@@ -342,6 +342,18 @@ def test_qualify_columns__invalid(self):
                     )
                     optimizer.qualify_columns.validate_qualify_columns(expression)
 
+        # Cannot expand star as not all nested structs are aliased
+        self.assertEqual(
```
tobymao, line 346: why is this here and not in the qualify_columns test?

sqlglot/optimizer/scope.py
```diff
@@ -231,6 +236,17 @@ def subqueries(self):
         self._ensure_collected()
         return self._subqueries
 
+    @property
+    def stars(self):
```
tobymao, line 240: why did you need to add this?

sqlglot/optimizer/qualify_columns.py
VaggelisD: If `schema = None` then we set `schema = sqlglot.schema` which doesn't have the same dialect as the `optimize()` step right?

georgesittas: Do we actually still need `sqlglot.schema`? Not sure what's the context behind that, but IIRC it was added before implementing `ensure_schema`, so it may be redundant now.

sqlglot/optimizer/scope.py
VaggelisD: The `scope.columns` wouldn't collect stars at all before so we couldn't recognize, annotate etc these expressions. Check out [this branch](https://github.com/tobymao/sqlglot/tree/jo/refactor_dot_conversion_in_qualify) which is the foundation of this PR

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
VaggelisD: Not unsafe per se, but if the condition has the "risk" that it can be repeated then we might be annotating the scope over and over again since we're under a loop

sqlglot/optimizer/scope.py
georgesittas: Just adding a bit to what Vaggelis said:

> wouldn't collect stars

FYI, relevant section in `Scope._collect`: https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L128-L129.

TL;DR we'd never convert columns to dots in `_qualify_columns`, because it happens inside the `scope.columns` loop and stars would not be visited.

sqlglot/optimizer/qualify_columns.py
VaggelisD: Removed it altogether

sqlglot/optimizer/qualify_columns.py
VaggelisD: Sorry, can you specify which would that be?

sqlglot/__init__.py
```diff
@@ -65,9 +65,6 @@
 pretty = False
 """Whether to format generated SQL by default."""
 
-schema = MappingSchema()
-"""The default schema used by SQLGlot (e.g. in the optimizer)."""
```
georgesittas, line 69: FYI, this makes this a breaking change, let's mark the PR accordingly.

sqlglot/optimizer/qualify_columns.py
georgesittas: Yep, we can move this to L508.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -413,19 +488,46 @@ def _expand_stars(
...
         if isinstance(expression, exp.Star):
             tables = list(scope.selected_sources)
             _add_except_columns(expression, tables, except_columns)
             _add_replace_columns(expression, tables, replace_columns)
-        elif expression.is_star and not isinstance(expression, exp.Dot):
-            tables = [expression.table]
-            _add_except_columns(expression.this, tables, except_columns)
-            _add_replace_columns(expression.this, tables, replace_columns)
+        elif expression.is_star:
+            if not isinstance(expression, exp.Dot):
```
georgesittas, line 503: Let's have the positive condition first, i.e. `if dot: ... else: ...`

sqlglot/optimizer/qualify_columns.py
```diff
@@ -413,19 +488,46 @@ def _expand_stars(
...
+                starting_struct = str(expression.this)
+                structs.append((col, starting_struct))
         else:
             new_selections.append(expression)
             continue
 
+        if structs:
+            struct_fields = _expand_struct_stars(scope, structs) or []
+
+            for struct_name in struct_fields:
```
georgesittas, line 519: I think we can simplify this section a bit, if `structs` is empty we'll always get back an empty list from the expand helper:
```suggestion
        for struct_name in _expand_struct_stars(scope, structs):
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -413,19 +488,46 @@ def _expand_stars(
...
+                root, *parts = struct_name.split(".")
+                alias_ = next_alias_name()
+                new_column: exp.Column | exp.Dot = exp.column(root, table=table_name)
+
+                if parts:
+                    new_column = exp.Dot.build(
+                        [new_column, *[exp.Identifier(this=part) for part in parts]]
```
georgesittas, line 526: Need to be careful about quotes, this strips them and can break queries with unsafe identifiers used as struct fields.

sqlglot/optimizer/qualify_columns.py
VaggelisD: This helps simplify the branch if I'm not mistaken, because otherwise we'd have to do:

```
if dot:
  if dialect == "bigquery"
     ....
else:
```

Note that if the condition is `if dot and bigquery` we'd potentially redirect `exp.Dots` from other dialects. Wdyt?

sqlglot/optimizer/qualify_columns.py
VaggelisD: When gathering these `parts` I extract their text as is, so when stitching them back together here I think the backticks are preserved e.g.: 

```
sql = "WITH tbl1 AS (SELECT STRUCT(1 AS `from`) AS col) SELECT tbl1.col.* from tbl1";

result = qualify(parse_one(sql), ....).sql(...)

print(result)
```

```
WITH tbl1 AS (SELECT STRUCT(1 AS `from`) AS col) SELECT tbl1.col.`from` AS _field_0 FROM tbl1 AS tbl1
```

Is this fine or is the problem you're mentioning more nuanced?

tests/test_optimizer.py
georgesittas: Not being able to expand the star doesn't mean it's an invalid query. I'd just add this in `qualify_columns.sql`, which already contains similar tests:

```
# execute: false
SELECT SOME_UDF(data).* FROM t;
SELECT SOME_UDF(t.data).* FROM t AS t;
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -413,19 +488,46 @@ def _expand_stars(
...
-            _add_except_columns(expression.this, tables, except_columns)
-            _add_replace_columns(expression.this, tables, replace_columns)
+        elif expression.is_star:
+            if not isinstance(expression, exp.Dot):
+                tables = [expression.table]
+                _add_except_columns(expression.this, tables, except_columns)
+                _add_replace_columns(expression.this, tables, replace_columns)
+            elif dialect == "bigquery":
+                col = expression.find(exp.Column)
+                table_name = col.table if col else None
```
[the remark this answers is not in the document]
VaggelisD: Shouldn't be, added asserts to safeguard

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+    while stack:
+        col, path_name = stack.pop()
+        assert col
+
+        if not isinstance(col.this, exp.Identifier):
+            return []
+
+        path_name = col.this.this if not path_name else f"{path_name}.{col.this.this}"
+        datatype = col.find(exp.DataType) or (col.type and col.type.find(exp.DataType))
+        assert datatype
```
georgesittas, line 457: Ditto about `.type` and not asserting `datatype`.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+    # Traverse & flatten the starting_col struct
+    while stack:
+        col, path_name = stack.pop()
+        assert col
+
+        if not isinstance(col.this, exp.Identifier):
+            return []
+
+        path_name = col.this.this if not path_name else f"{path_name}.{col.this.this}"
```
georgesittas, line 455: Ditto about using `name`.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+    if not starting_col or starting_col.this.this != dot_parts[-2].this:
+        raise OptimizeError(f"Unknown struct: {part.this}")
+
+    starting_name = ".".join([part.this for part in dot_parts[1:-2]])
+    stack = [(starting_col, starting_name)]
+
+    # Traverse & flatten the starting_col struct
+    while stack:
+        col, path_name = stack.pop()
+        assert col
```
georgesittas, line 450: Why are we asserting here?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+                return []
+
+            if expr.this.this == part.this:
+                starting_col = expr
+                assert starting_col
+
+    if not starting_col or starting_col.this.this != dot_parts[-2].this:
+        raise OptimizeError(f"Unknown struct: {part.this}")
+
+    starting_name = ".".join([part.this for part in dot_parts[1:-2]])
```
georgesittas, line 444: Why are we slicing from 1 to -2 here? Example? Let's leave a comment if this is necessary.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+    stack = [(starting_col, starting_name)]
+
+    # Traverse & flatten the starting_col struct
+    while stack:
+        col, path_name = stack.pop()
+        assert col
+
+        if not isinstance(col.this, exp.Identifier):
+            return []
```
georgesittas, line 453: Ditto about checking for `ColumnDef`.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+        path_name = col.this.this if not path_name else f"{path_name}.{col.this.this}"
+        datatype = col.find(exp.DataType) or (col.type and col.type.find(exp.DataType))
+        assert datatype
+
+        # Collect the non-struct columns
+        if datatype.this != exp.DataType.Type.STRUCT:
+            struct_fields.append(path_name)
+        else:
+            for expr in reversed(datatype.expressions):
+                stack.append((expr, path_name))
```
georgesittas, line 464:
```suggestion
            stack.extend((expr, path_name) for expr in reversed(datatype.expressions))
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+    new_selections = []
+
+    for struct_field in struct_fields:
+        root, *parts = struct_field.split(".")
+        alias_ = next_alias_name()
+        new_column: exp.Column | exp.Dot = exp.column(root, table=table_name)
+
+        if parts:
+            new_column = exp.Dot.build([new_column, *[exp.Identifier(this=part) for part in parts]])
```
georgesittas, line 474: Here, `parts` is produced by splitting a string on `.`. Then, we use it to produce a new `Dot` where we create identifiers from each part. What happens if there are quotes? If a `part` is quoted, it seems that you'll add `"name"` to `Identifier.this` which is incorrect.

We need to be very careful with quotes because they will mess things up eg. in SQLMesh if not done right. Let's test this path as well.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+    assert starting_col
+
+    dot_parts = expression.parts
+    table_name = starting_col.table
+
+    # If there's more nested levels specified, start the DFS traversal from the nested struct first
+    for part in dot_parts[2:-1]:
+        datatype = starting_col.find(exp.DataType) or (
+            starting_col.type and starting_col.type.find(exp.DataType)
+        )
```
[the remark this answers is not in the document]
VaggelisD: The second `find` is for the case where we are handling an `exp.Column` and is indeed superfluous, thanks!

The first find is because if we're handling `exp.ColumnDef` then the type is stored as `kind` arg, so `.type` does not work well here right?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+        assert datatype
```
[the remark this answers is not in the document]
VaggelisD: This is to find _any_ data type even if it's `Type.Unknown`, so we always expect to have a valid `datatype` afaict. All the asserts are to eliminate mypy `t.Optional` warnings without checking them with branches, as there would be no reason to have `None` expressions throughout the `expand_struct_star`

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+        for expr in datatype.expressions:
+            if not isinstance(expr.this, exp.Identifier):
+                return []
+
+            if expr.this.this == part.this:
+                starting_col = expr
+                assert starting_col
```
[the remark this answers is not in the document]
VaggelisD: 1. Throughout the traversal we should only come across `exp.Column` (starting column) or `exp.ColumnDef`; Anything else would be problematic, so maybe I can assert these types before checking the args. In both these exprs we'd expect a valid `Identifier` in `this`
2. 👍 
3. Mypy `t.Optional` warning here as well

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,89 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+    if not starting_col or starting_col.this.this != dot_parts[-2].this:
+        raise OptimizeError(f"Unknown struct: {part.this}")
```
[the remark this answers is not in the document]
VaggelisD: This is a sanity check that the `starting_col` is the user requested struct; For example, if the user specifies `tbl.col.f1.*` but `f1` is not an existing nested field, then `starting_col` would be `exp.Column(this=col, table=tbl)` and `parts[-2]` would be `f1`

sqlglot/optimizer/qualify_columns.py
VaggelisD: Ditto about mypy warning & `exp.ColumnDef` problem

sqlglot/optimizer/qualify_columns.py
VaggelisD: If the user wants to expand `tbl.col.f1.f2.f3.*`, then the parts would be `[tbl, col, f1, f2, f3, *]` and the starting struct would be `f3`. However, since the traversal starts directly from a nested point, we must begin the struct field flattenning with the prefix `starting_name = parts[1:-2] = col.f1.f2`

sqlglot/optimizer/qualify_columns.py
VaggelisD: Ditto about the assertion of `Column | ColumnDef`

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
georgesittas: I think this is risky. We should _only_ assert when we're 100% sure about a runtime property and we *need* it to hold. The logic here is somewhat implicit and feels tricky to debug.

For example, if you get an `UNKNOWN` type from the type inference, you'll silently skip the loop below because `datatype.expressions` will return an empty list for it. And then if you hit L441, 1) you'll have a `starting_col` and 2) it looks like the `starting_col.this.this != dot_parts[-2].this` condition will fail? Hence we'll raise.

Let's make sure this is implemented in a more explicit fashion to safely handle all code paths and avoid raising.

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
georgesittas: Can we just do this a bit more explicitly? This is not very obvious and I'm not sure I understand the implications of using `find` vs having a couple of branches to check the two cases, i.e. `Column` and `ColumnDef`.

If we can somehow handle `Column` before the loop to do make this cleaner, that'd be ideal.

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
georgesittas: Imo we shouldn't raise if we encounter any weird conditions. If the transformation doesn't work out for whatever reason, fallback to the safest option, i.e. not doing anything.

sqlglot/optimizer/qualify_columns.py
georgesittas: Let's comment this behavior

sqlglot/optimizer/qualify_columns.py
georgesittas: Where do we add to the stack something that could be `None`?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,102 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+    # The keys to non-struct field entries in `nested_fields`
+    nested_field_keys = []
+
+    stack = [(starting_struct, id(starting_struct))]
+
+    # Traverse & flatten the starting struct
+    while stack:
+        col, parent_struct_id = stack.pop()
+        id_col = id(col)
```
georgesittas, line 453: Move this (L453) to L459

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,102 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+        # Preserve the qualified path for this field
+        nested_fields[id_col] = list(
+            itertools.chain(
+                [part.copy() for part in nested_fields[parent_struct_id]], [col.this.copy()]
+            )
+        )
+
+        # Collect the non-struct columns
+        if col.kind.this != exp.DataType.Type.STRUCT:
```
georgesittas, line 467:
```suggestion
        if not col.kind.is_type(exp.DataType.Type.STRUCT):
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,109 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+        if not struct_fields:
+            return []
+
+        for fld in struct_fields:
+            # Unable to expand star unless all struct fields are named
+            if not isinstance(fld.this, exp.Identifier):
+                return []
+
+            if fld.name == part.name and fld.kind.this == exp.DataType.Type.STRUCT:
```
georgesittas, line 442:
```suggestion
            if fld.name == part.name and fld.kind.is_type(exp.DataType.Type.STRUCT):
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,109 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+                [part.copy() for part in dot_parts[:-1]], nested_fields[nested_fields_key]
+            )
+        )
+        new_column: exp.Column | exp.Dot = exp.column(
+            t.cast(exp.Identifier, root), table=dot_column.table
+        )
+
+        if parts:
+            new_column = exp.Dot.build([new_column, *parts])
+
```
georgesittas, line 495: Maybe this can be simplified to `new_column = exp.column(root, table=..., fields=parts)`.

georgesittas (commented):
Looks good 👍

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,109 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+        for fld in struct_fields:
+            # Unable to expand star unless all struct fields are named
+            if not isinstance(fld.this, exp.Identifier):
+                return []
+
+            if fld.name == part.name and fld.kind.this == exp.DataType.Type.STRUCT:
+                starting_struct = fld
+
+    if starting_struct.name != dot_parts[-1].name:
+        return []
```
georgesittas, line 446: When can this check fail?

VaggelisD: If the user defines an expansion such as `tbl.f0.f1.f2.*` but we're unable to find such fields in our first traversal (or we found them, but were unable to annotate them as structs).

This branch checks that the `starting_struct` is indeed the last part of the expression i.e `f2` in this case

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,105 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+        if not struct_fields:
+            return []
+
+        for fld in struct_fields:
+            # Unable to expand star unless all struct fields are named
+            if not isinstance(fld.this, exp.Identifier):
+                return []
+
+            if fld.name == part.name and fld.kind.is_type(exp.DataType.Type.STRUCT):
+                starting_struct = fld
```
georgesittas, line 443: Perhaps something interesting to notice here is that if, for whatever reason, there is no `fld` that matches the current `part`, we'll keep traversing `dot_parts` using the same `starting_struct`. What happens if there _is_ a matching `fld` for the *next* dot part? It looks like we'll happily accept that and update the `starting_struct`, which may lead to an invalid expansion later on.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,105 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+            return []
+
+        id_fld = id(fld)
+
+        # Preserve the qualified path for this field
+        nested_fields[id_fld] = list(
+            itertools.chain(
+                [part.copy() for part in nested_fields[parent_struct_id]], [fld.this.copy()]
+            )
+        )
```
georgesittas, line 471:
```suggestion
        nested_fields[id_fld] = [
            part.copy() for part in itertools.chain(nested_fields[parent_struct_id], [fld.this])
        ]
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,105 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+    new_selections = []
+    for nested_fields_key in nested_field_keys:
+        # The fully qualified path of each field is created by:
+        # - The path until the starting struct, if any (dot_parts)
+        # - The path from the starting struct until the nested field (nested_structs[key])
+        root, *parts = list(
+            itertools.chain(
+                [part.copy() for part in dot_parts[:-1]], nested_fields[nested_fields_key]
+            )
+        )
```
georgesittas, line 488:
```suggestion
        root, *parts = [
            part.copy()
            for part in itertools.chain(dot_parts[:-1], nested_fields[nested_fields_key])
        ]
```

georgesittas (approved):
A few more comments, LGTM otherwise.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,105 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
 
 
+def _expand_struct_stars(
+    expression: exp.Dot,
+    next_alias_name: t.Callable[[], str],
+) -> t.List[exp.Alias]:
+    """[BigQuery] Expand/Flatten foo.bar.* where bar is a struct column"""
+
+    dot_column = t.cast(exp.Column, expression.find(exp.Column))
+    if not dot_column.is_type(exp.DataType.Type.STRUCT):
```
tobymao, line 420: where are types being annotated so that this works?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,105 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+        return []
+
+    # All nested struct values are ColumnDefs, so normalize the first exp.Column in one
+    dot_column = dot_column.copy()
```
tobymao, line 424: why does it need to be copied?

sqlglot/optimizer/qualify_columns.py
VaggelisD: Interesting take. Not only that, but looking at the nested loops I've made it so the internal loop keeps searching even if it has found the next `fld` already.

In my follow up commit I'll fix both issues at once by checking if each part has matched; If yes, we `break` and go to the next part, otherwise we fail early and return `[]`

georgesittas: I was thinking perhaps we can get rid of the condition below, where we check if the starting struct matches the last dot part. I think one invariant in this loop is that for each element in `dot_parts`, we must find at least one `fld` such that `fld.name == part.name`, otherwise there's no progress made. So if this invariant is true by the end of the loop, that condition would be redundant.

sqlglot/optimizer/qualify_columns.py
VaggelisD: The scope is annotated in advance at the entry point of `_expand_stars` in case (1) the dialect is BigQuery and (2) we have `exp.Dot` star expressions i.e struct expansions

sqlglot/optimizer/qualify_columns.py
VaggelisD: Let's say we attempt to expand `tbl.col.*`; this expression when entering `_expand_struct_star()` is of type `exp.Dot(....(exp.Column)), exp.Star)`, where the column is the top level struct  i.e `exp.Column(this=col, table=tbl)`.  However, the nested (and annotated) fields of `col` are of type `exp.ColumnDef`, so during the first steps of the traversal we'd have 2 types of expressions to handle.

To canonicalize the procedure, we transform that initial column into an `exp.ColumnDef` as well, but since our expansion can fail e.g. because of unnamed fields we also want to preserve the original expression `tbl.col.*` to fallback to, thus the copy.

sqlglot/optimizer/qualify_columns.py
VaggelisD: This suggestion was implemented in feedback n.6

georgesittas (approved):
Awesome 🎉

cc @z3z1ma

barakalon (approved):

sqlglot/optimizer/qualify_columns.py
```diff
@@ -329,6 +338,47 @@ def _select_by_pos(scope: Scope, node: exp.Literal) -> exp.Alias:
...
+    """
+    converted = False
+    for column in itertools.chain(scope.columns, scope.stars):
+        if isinstance(column, exp.Dot):
+            continue
+
+        column_table: t.Optional[str | exp.Identifier] = column.table
+        if (
+            column_table
+            and column_table not in scope.sources
```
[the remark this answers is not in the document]
VaggelisD: Could you please share your thought process here? As this is existing code we were reluctant in changing it along with the rest of the additions.

 Afaict `scope.selected_sources` is a subset of `scope.sources` and it's produced by "filtering"  the entries that are _not_ referenced in the scope, right? 

So in general, probing `scope.sources` might yield false positives for the scope?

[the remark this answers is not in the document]
barakalon: Thats exactly right, yeah

I'm OK if we don't want to change existing code. I didn't realize it was copy/pasted until halfway through my review.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -380,18 +410,100 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
...
+
+            if fld.name == part.name and fld.kind.is_type(exp.DataType.Type.STRUCT):
+                starting_struct = fld
+                break
+        else:
+            # There is no matching field in the struct
+            return []
+
+    # Each nested field caches its path in exp.Identifier parts e.g. tbl.f0.f1.item -> [exp.Identifier(tbl), ..., exp.Identifier(item)]
+    fld_path: t.Dict[int, t.List[exp.Identifier]] = {id(starting_struct): []}
```
tobymao, line 450: rename all fld to field

tests/test_optimizer.py
```diff
@@ -317,6 +317,18 @@ def test_qualify_columns(self, logger):
             'WITH "t" AS (SELECT 1 AS "c") (SELECT "t"."c" AS "c" FROM "t" AS "t")',
         )
 
+        self.assertEqual(
+            optimizer.qualify_columns.qualify_columns(
+                parse_one(
+                    "WITH tbl1 AS (SELECT STRUCT(1 AS `f0`, 2 as f1) AS col) SELECT tbl1.col.* from tbl1",
```
tobymao, line 323: why is this here an not in qualify_columns?

sqlglot/optimizer/scope.py
```diff
@@ -231,6 +238,17 @@ def subqueries(self):
...
+        """
+        List of star columns in this scope.
+
+        Returns:
+            list[exp.Column]: Column instances in this scope that are a star.
```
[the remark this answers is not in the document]
tobymao: we shouldn't add this style in here, for new methods just do a proper type annotation

tobymao (approved):

sqlglot/optimizer/qualify_columns.py
```diff
@@ -413,16 +525,29 @@ def _expand_stars(
             if not pivot_output_columns:
                 pivot_output_columns = [c.alias_or_name for c in pivot.expressions]
 
+    is_bigquery = dialect == "bigquery"
```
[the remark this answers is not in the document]
VaggelisD: Agree, the optimizer lately has seen a couple additions for BigQuery specifically so the  `dialect == "bigquery"` theme has been repeated in a couple places already. 

However, I'm inclined to keep it as is for now and will definitely reconsider the refactor if there's similar applications (struct star expansion, struct unnest etc) in other dialects as well.

sqlglot/optimizer/qualify_columns.py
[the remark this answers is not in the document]
georgesittas: Not sure about the `selected_sources` here. I _think_ `scope.sources` was used because in lateral table explosions you won't see the preceding (exploded) tables in the UDTF's `scope.selected_sources`. They are, however, available in `scope.lateral_sources` (see [here](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L688)), meaning that `scope.sources` also [contains](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py#L73-L74) them.

With a quick check, there are a few tests that fail if we change this to `scope.selected_sources`, e.g.:

```diff
# execute: false
# dialect: redshift
- SELECT CAST(c.f AS VARCHAR(MAX)) AS f, e AS e FROM a.b AS c, c.d AS e
+ SELECT CAST(c.f AS VARCHAR(MAX)) AS f, e AS e FROM a.b AS c, c.c.d AS e
?                                                              ++
```

sqlglot/optimizer/qualify_columns.py
georgesittas: > These branches have gotten pretty hard to follow.
> 
> Is it possible to flatten them, fizzbuzz style?

FWIW I tried to simplify this but couldn't make it cleaner without repeating some condition (eg. `is_star`). Did you have something specific in mind?

PR #3543, opened 2024-05-24
Feat(mysql,doris): improve transpilation of INTERVAL (plural to singular)

Toms1999:


georgesittas (approved):
Seems legit, thanks for the PR.

Please include docs for future PRs so that we can cross-check their validity.

Issue #3534, opened 2024-05-22
Snowflake Alter Table ... Alter Column ... DROP NOT NULL doesn't parse correctly


barino86:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`: **snowflake**
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`: **snowflake**
- Check if the issue still exists on main: **it does**

**Fully reproducible code snippet**
Please include a fully reproducible code snippet or the input sql, dialect, and expected output.

```python
from sqlglot import parse_one

query = """
    ALTER TABLE DB_NAME.SCHMA_NAME.TBL_NAME
    ALTER COLUMN MY_COLUMN DROP NOT NULL;
"""
parsed = parse_one(query, read='snowflake')
print(repr(parsed))


'ALTER TABLE DB_NAME.SCHMA_NAME.TBL_NAME
            ALTER COLUMN MY_COLUMN DROP NOT NULL' contains unsupported syntax. Falling back to parsing as a 'Command'.
Command(this=ALTER, expression=TABLE DB_NAME.SCHMA_NAME.TBL_NAME
               ALTER COLUMN MY_COLUMN DROP NOT NULL)
```

I would have expected an AlterTable expression with actions.

**Official Documentation**
Please include links to official SQL documentation related to your issue.

https://docs.snowflake.com/en/sql-reference/sql/alter-table-column

noklam:
I would love to take on this issue if possible. I try to modify the query to 
```
query = """
    ALTER TABLE DB_NAME.SCHMA_NAME.TBL_NAME
    ALTER COLUMN MY_COLUMN DROP DEFAULT;
"""

AlterTable(
  this=Table(
    this=Identifier(this=TBL_NAME, quoted=False),
    db=Identifier(this=SCHMA_NAME, quoted=False),
    catalog=Identifier(this=DB_NAME, quoted=False)),
  actions=[
    AlterColumn(
      this=Identifier(this=MY_COLUMN, quoted=False),
      drop=True)])
```
and the expression seems to be parsed correctly, so I guess the issue is that the `DROP NOT NULL` statement is not understood by `sqlglot` properly.

noklam:
https://github.com/tobymao/sqlglot/blob/59de4f6e0f48f5354e7b462fb65633018184aa3c/sqlglot/parser.py#L6024-L6038

[...]

PR #3550, opened 2024-05-25
Feat: add support for ALTER COLUMN DROP NOT NULL

noklam:
fix: #3534 
Context:
```python
class AlterColumn(Expression):
    arg_types = {
        "this": True,
        "dtype": False,
        "collate": False,
        "using": False,
        "default": False,
        "drop": False,
        "comment": False,

    }
```

The existing `AlterColumn` expression seems to assume only `SET DEFAULT ` and `NOT DEFAULT` is possible. Therefore is `drop` is True is sufficient to represent `DROP DEFAULT` and when `default` is not null it means set default. This seems to be insufficient to handle statement like `DROP NOT NULL` and `SET NOT NULL`

Few questions:
-  there are many `_match_xxx` functions, is `_match_text_seq` the most suitable one?
- I don't know if I should use things like `_parse_not_constraints` here.
- I figured I should edit the `AlterColumn` expression, but I am not sure how to add a new type for `SET NOT NULL` and `DROP NOT NULL`. I found there are `NotNullColumnConstrains`. I take a look at other expression and I thought adding a new optional arg_type `expression` seems to best fit.

# Test
Reuse the original issue example.

This is expected to work
```python
ast = parse_one(sql, read="spark")
ast.sql(write="duckdb")
```

Duckdb Syntax: https://duckdb.org/docs/sql/statements/alter_table.html
Snowflake Syntax: https://docs.snowflake.com/en/sql-reference/sql/alter-table-column

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1632,6 +1632,7 @@ class AlterColumn(Expression):
         "default": False,
         "drop": False,
         "comment": False,
+        "expression": False
     }
 
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1632,7 +1632,7 @@ class AlterColumn(Expression):
         "default": False,
         "drop": False,
         "comment": False,
-        "expression": False
+        "allow_null": False,
     }
 
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3028,7 +3028,7 @@ class Generator(metaclass=_Generator):
         if not drop and not allow_null:
             self.unsupported("Unsupported ALTER COLUMN syntax")
 
-        if allow_null:
+        if allow_null is not None:
             keyword = "DROP" if drop else "SET"
             return f"ALTER COLUMN {this} {keyword} NOT NULL"
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3022,6 +3022,10 @@ class Generator(metaclass=_Generator):
         if comment:
             return f"ALTER COLUMN {this} COMMENT {comment}"
 
+        allow_null = self.sql(expression, "allow_null")
+        if allow_null:
+            nullability = "SET" if allow_null == "NOT NULL" else "DROP"
+            return f"ALTER COLUMN {this} {nullability} NOT NULL"
         if not expression.args.get("drop"):
             self.unsupported("Unsupported ALTER COLUMN syntax")
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3022,12 +3022,15 @@ class Generator(metaclass=_Generator):
         if comment:
             return f"ALTER COLUMN {this} COMMENT {comment}"
 
-        allow_null = self.sql(expression, "allow_null")
-        if allow_null:
-            nullability = "SET" if allow_null == "NOT NULL" else "DROP"
-            return f"ALTER COLUMN {this} {nullability} NOT NULL"
-        if not expression.args.get("drop"):
+        allow_null = expression.args.get("allow_null")
+        drop = expression.args.get("drop")
+
+        if not drop and not allow_null:
             self.unsupported("Unsupported ALTER COLUMN syntax")
+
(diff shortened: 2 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -1632,6 +1632,7 @@ class AlterColumn(Expression):
         "default": False,
         "drop": False,
         "comment": False,
+        "expression": False
```
georgesittas, line 1635: Let's rename this to `not_null` so it's a bit clearer what it stands for.

sqlglot/parser.py
```diff
@@ -6055,7 +6055,15 @@ def _parse_alter_table_alter(self) -> t.Optional[exp.Expression]:
             return self.expression(exp.AlterColumn, this=column, default=self._parse_conjunction())
         if self._match(TokenType.COMMENT):
             return self.expression(exp.AlterColumn, this=column, comment=self._parse_string())
-
+        if self._match_text_seq("DROP", "NOT" ,"NULL"):
+            return self.expression(exp.AlterColumn,
+                                   this=column,
+                                drop=True,
+                                   expression=exp.NotNullColumnConstraint)
```
georgesittas, line 6062: This needs to be `expression=exp.NotNullColumnConstraint()`, because it currently assigns a class to the arg.

georgesittas (commented):
Thanks for the PR @noklam, this is in the right direction - left a couple of comments. Make sure to also run `make style` and add tests in `test_snowflake.py`.

sqlglot/expressions.py
noklam: I will use `allow_null` instead as it seems to be more consistent with the `NotNullColumnConstraints` arg_type.

sqlglot/parser.py
noklam: Rookie mistake, Thanks!

noklam:
ready to review now,

sqlglot/generator.py
```diff
@@ -3022,6 +3022,10 @@ def altercolumn_sql(self, expression: exp.AlterColumn) -> str:
         if comment:
             return f"ALTER COLUMN {this} COMMENT {comment}"
 
+        allow_null = self.sql(expression, "allow_null")
+        if allow_null:
+            nullability = "SET" if allow_null == "NOT NULL" else "DROP"
+            return f"ALTER COLUMN {this} {nullability} NOT NULL"
```
georgesittas, line 3028: What's the point of storing a `exp.NotNullColumnConstraint` in this case, if we're not gonna use the SQL it generates into? I think storing a boolean flag instead will be cleaner, e.g. no string comparison will be required.

tests/dialects/test_snowflake.py
```diff
@@ -1282,6 +1282,13 @@ def test_ddl(self):
             read={"teradata": "CREATE MULTISET TABLE a (b INT)"},
             write={"snowflake": "CREATE TABLE a (b INT)"},
         )
+        self.validate_all(
+            """
+                ALTER TABLE a
+                ALTER COLUMN my_column DROP NOT NULL;
+            """,
+            write={"duckdb": "ALTER TABLE a ALTER COLUMN my_column DROP NOT NULL"},
+        )
```
georgesittas, line 1291: Make sure to always test the roundtrip, i.e. snowflake -> snowflake. So we need a couple of entries here.

noklam:
I have made changes based on the review

sqlglot/generator.py
```diff
@@ -3022,9 +3022,16 @@ def altercolumn_sql(self, expression: exp.AlterColumn) -> str:
...
             return f"ALTER COLUMN {this} COMMENT {comment}"
 
-        if not expression.args.get("drop"):
+        allow_null = expression.args.get("allow_null")
+        drop = expression.args.get("drop")
+
+        if not drop and not allow_null:
             self.unsupported("Unsupported ALTER COLUMN syntax")
 
+        if allow_null:
```
georgesittas, line 3031:
```suggestion
        if allow_null is not None:
```

PR #3552, opened 2024-05-26
Fix(postgres): setting un-suffixed FLOAT as DOUBLE (#3551)

### LINKED ISSUE #3551: Postgres/Redshift - `FLOAT` is incorrectly interpreted as `REAL`
**Issue**
`FLOAT` in postgres and Redshift is equivalent to the double-precision floating point format. However, [KEYWORDS.FLOAT](https://github.com/tobymao/sqlglot/blob/main/sqlglot/tokens.py#L822) is interpreted as `REAL` instead of `DOUBLE` in the current implementation.

**Code to reproduce**
```python
from sqlglot import parse_one

postgres_ast = parse_one("cast(a AS FLOAT)", dialect="postgres")
print(postgres_ast.sql(dialect="postgres"))
#returns CAST(a AS REAL)
#expected assert(postgres_ast.sql(dialect="postgres") == "CAST(a AS DOUBLE PRECISION)")

redshift_ast = parse_one("cast(a AS FLOAT)", dialect="redshift")
print(redshift_ast.sql(dialect="redshift"))
#returns CAST(a AS REAL)
#expected assert(redshift_ast.sql(dialect="redshift") == "CAST(a AS DOUBLE PRECISION)")
```

**Official Documentation**
[Redshift - Floating-Point types](https://docs.aws.amazon.com/redshift/latest/dg/r_Numeric_types201.html#r_Numeric_types201-floating-point-types) - 
>DOUBLE PRECISION represents the double-precision floating point format, according to the IEEE Standard 754 for Binary Floating-Point Arithmetic. It has a precision of about 15 digits, and a range of around 1E-307 to 1E+308. You can also specify this data type as FLOAT or FLOAT8.

[Postgres - Floating point types](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-FLOAT) - 
>float with no precision specified is taken to mean double precision.

sandband:
fix #3551 
Redshift inherits from Postgres and hence haven't added any testcases to Redshift.

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -336,6 +336,7 @@ class Postgres(Dialect):
             "REGPROCEDURE": TokenType.OBJECT_IDENTIFIER,
             "REGROLE": TokenType.OBJECT_IDENTIFIER,
             "REGTYPE": TokenType.OBJECT_IDENTIFIER,
+            "FLOAT": TokenType.DOUBLE,
         }
 
         SINGLE_TOKENS = {
--- a/tests/dialects/test_postgres.py
+++ b/tests/dialects/test_postgres.py
@@ -717,6 +717,9 @@ class TestPostgres(Validator):
         self.validate_identity(
             "COPY (SELECT * FROM t) TO 'file' WITH (FORMAT format, HEADER MATCH, FREEZE TRUE)"
         )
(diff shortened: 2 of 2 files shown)


georgesittas (approved):
Took a quick look and the tests seem to check out - thanks for the PR!

Screenshot 2024-05-27 at 10 36 06 AM

Screenshot 2024-05-27 at 10 36 30 AM

Issue #3553, opened 2024-05-27
Redshift - add support for set operator `MINUS`


sandband:
**Issue**
`MINUS` operator is currently not supported in Redshift dialect. `MINUS` in Redshift is exactly synonymous to the standard `EXCEPT` operator

**Code**:
```python
from sqlglot import parse_one

redshift_ast = parse_one("""select foo, bar from table_1 
minus 
select foo, bar from table_2""", dialect="redshift")

print(redshift_ast.sql(dialect="redshift", pretty=True))
```
*Expected*: 
```sql
SELECT
  foo,
  bar
FROM table_1
EXCEPT
SELECT
  foo,
  bar
FROM table_2
```

*Actual*: throws error
```python
ParseError: Invalid expression / Unexpected token. Line 3, Col: 6.
  select foo, bar from table_1 
minus 
select foo, bar from table_2
```
**Official documentation**
[UNION, INTERSECT, and EXCEPT](https://docs.aws.amazon.com/redshift/latest/dg/r_UNION.html#r_UNION-parameters)
>MINUS and EXCEPT are exact synonyms

georgesittas:
Fixed by https://github.com/tobymao/sqlglot/commit/e1a9a8b6b7fbd44e62cee626540f90425d22d50c

PR #3555, opened 2024-05-27
Fix(redshift): add support for MINUS operator #3553

sandband:
Added support for `MINUS` operator in Redshift.


georgesittas (approved):
LGTM, thanks for the PR!

PR #3556, opened 2024-05-28
Fix: use regex to split interval strings

### LINKED ISSUE #3554: Snowflake parser error when time interval is followed by AND (...)
**Before you file an issue**

```
sql = """select *
from db.schema.table
where start_time > current_timestamp - interval '1w'
    and (1=1)
"""

print(sqlglot.parse_one(sql, dialect='snowflake'))
```

Error:
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 4, Col: 9.
  select *
from db.schema.table
where start_time > current_timestamp - interval '1w'
    and (1=1)
```

**Official Documentation**
https://docs.snowflake.com/en/sql-reference/data-types-datetime#interval-constants

georgesittas:
Fixes #3554

Reference: https://docs.snowflake.com/en/sql-reference/data-types-datetime

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6793,7 +6793,7 @@ def parse_identifier(name: str | Identifier, dialect: DialectType = None) -> Ide
     return expression
 
 
-INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
+INTERVAL_STRING_RE = re.compile(r"\s*(\d+)\s*([a-zA-Z]+)\s*")
 
 
 def to_interval(interval: str | Literal) -> Interval:
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6793,7 +6793,7 @@ def parse_identifier(name: str | Identifier, dialect: DialectType = None) -> Ide
     return expression
 
 
-INTERVAL_STRING_RE = re.compile(r"\s*(\d+)\s*([a-zA-Z]+)\s*")
+INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
 
 
 def to_interval(interval: str | Literal) -> Interval:
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4094,7 +4094,6 @@ class Parser(metaclass=_Parser):
             if len(parts) == 1:
                 if unit:
                     # Unconsume the eagerly-parsed unit, since the real unit was part of the string
-                    unit = None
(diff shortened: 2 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -6793,7 +6793,7 @@ def parse_identifier(name: str | Identifier, dialect: DialectType = None) -> Ide
     return expression
 
 
-INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
+INTERVAL_STRING_RE = re.compile(r"\s*(\d+)\s*([a-zA-Z]+)\s*")
```
VaggelisD, line 6796: Is `\d` matching new test cases better than `[0-9]` ?

sqlglot/parser.py
```diff
@@ -4090,16 +4090,15 @@ def _parse_interval(self, match_interval: bool = True) -> t.Optional[exp.Add | e
...
                 if unit:
-                    # This is not actually a unit, it's something else (e.g. a "window side")
+                    # Unconsume the eagerly-parsed unit, since the real unit was part of the string
                     unit = None
                     self._retreat(self._index - 1)
 
-                this = exp.Literal.string(parts[0])
-                unit = self.expression(exp.Var, this=parts[1].upper())
+                this = exp.Literal.string(parts[0][0])
+                unit = self.expression(exp.Var, this=parts[0][1].upper())
```
VaggelisD, line 4101: Do we have to set `unit = None`? Seems like we'll set it to `exp.Var` right after the branch anyways

sqlglot/expressions.py
georgesittas: No, I used it to make the regex more compact but I actually reverted since it captures a superset of `[0-9]` (eg. weird unicode numerals) that we don't care about.

sqlglot/parser.py
```diff
@@ -4090,16 +4090,14 @@ def _parse_interval(self, match_interval: bool = True) -> t.Optional[exp.Add | e
         if this and this.is_number:
             this = exp.Literal.string(this.name)
         elif this and this.is_string:
-            parts = this.name.split()
-
-            if len(parts) == 2:
+            parts = exp.INTERVAL_STRING_RE.findall(this.name)
+            if len(parts) == 1:
```
VaggelisD, line 4094: Is this branch condition matching the previous one? 

Seems like `parts` is now a 2-D array, maybe the transferred condition should be `len(parts[0]) == 2` ?

georgesittas: Nope, if you had `1 day` it'd previously be split into `[1, 'day']`. Now it's split into [(1, 'day')]`. Different representation.

PR #3564, opened 2024-05-29
fix(snowflake): COPY Postfix

VaggelisD:
[Slack context](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1716933881216719)

Fix `COPY INTO` by parsing it's options as properties & replacing their unquoted identifiers with `exp.Var`; This is needed because all `exp.Identifier` are quoted in SQLMesh.


tests/dialects/test_snowflake.py
```diff
@@ -1921,22 +1921,24 @@ def test_try_cast(self):
...
+            """COPY INTO mytable (col1, col2) FROM 's3://mybucket/data/files' STORAGE_INTEGRATION="storage" ENCRYPTION = (TYPE='NONE' MASTER_KEY='key') FILES = ('file1', 'file2') PATTERN = 'pattern' FILE_FORMAT = (FORMAT_NAME=my_csv_format NULL_IF=('')) PARSE_HEADER = TRUE"""
+        )
+        self.validate_identity(
+            """COPY INTO 's3://example/contacts.csv' FROM "db"."tbl" STORAGE_INTEGRATION="PROD_S3_SIDETRADE_INTEGRATION" FILE_FORMAT = (TYPE=CSV COMPRESSION=NONE NULL_IF=('') FIELD_OPTIONALLY_ENCLOSED_BY='"') HEADER = TRUE OVERWRITE = TRUE SINGLE = TRUE"""
         )
         self.validate_all(
             """COPY INTO 's3://example/data.csv'
     FROM EXTRA.EXAMPLE.TABLE
-    credentials = (x)
-    STORAGE_INTEGRATION = S3_INTEGRATION
```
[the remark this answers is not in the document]
VaggelisD: Per the [snowflake docs](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#additional-cloud-provider-parameters) we can have either `STORAGE_INTEGRATION` or `CREDENTIALS`

tests/dialects/test_snowflake.py
```diff
@@ -1945,22 +1947,20 @@ def test_copy(self):
...
   HEADER TRUE,
   OVERWRITE TRUE,
   SINGLE TRUE
 )""",
                 "snowflake": """COPY INTO 's3://example/data.csv'
 FROM EXTRA.EXAMPLE.TABLE
-CREDENTIALS = (x)
-STORAGE_INTEGRATION = S3_INTEGRATION
-FILE_FORMAT = (TYPE = CSV COMPRESSION = NONE NULL_IF = (
+CREDENTIALS = ()
```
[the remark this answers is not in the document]
VaggelisD: It's not 100% proper Snowflake, credentials can be empty i.e `CREDENTIALS = ()` or contain a series of options such as `CREDENTIALS = (AWS_KEY_ID = 'key')` so I thought at least the former would fit better

tests/dialects/test_snowflake.py
```diff
@@ -1921,22 +1921,24 @@ def test_try_cast(self):
     def test_copy(self):
         self.validate_identity("COPY INTO test (c1) FROM (SELECT $1.c1 FROM @mystage)")
         self.validate_identity(
-            """COPY INTO temp FROM @random_stage/path/ FILE_FORMAT = (TYPE = CSV FIELD_DELIMITER = '|' NULL_IF = () FIELD_OPTIONALLY_ENCLOSED_BY = '"' TIMESTAMP_FORMAT = 'TZHTZM YYYY-MM-DD HH24:MI:SS.FF9' DATE_FORMAT = 'TZHTZM YYYY-MM-DD HH24:MI:SS.FF9' BINARY_FORMAT = BASE64) VALIDATION_MODE = 'RETURN_3_ROWS'"""
+            """COPY INTO temp FROM @random_stage/path/ FILE_FORMAT = (TYPE=CSV FIELD_DELIMITER='|' NULL_IF=('str1', 'str2') FIELD_OPTIONALLY_ENCLOSED_BY='"' TIMESTAMP_FORMAT='TZHTZM YYYY-MM-DD HH24:MI:SS.FF9' DATE_FORMAT='TZHTZM YYYY-MM-DD HH24:MI:SS.FF9' BINARY_FORMAT=BASE64) VALIDATION_MODE = 'RETURN_3_ROWS'"""
```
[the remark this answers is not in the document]
georgesittas: Sounds good

sqlglot/parser.py
```diff
@@ -6667,16 +6667,36 @@ def _parse_copy_parameters(self) -> t.List[exp.CopyParameter]:
...
+            return prop
+
+        # Replace the unquoted identifiers with vars
+        this = prop.this.find(exp.Identifier)
+        value = prop.args.get("value")
+        value = value.find(exp.Identifier) if value else None
+        if this and not this.quoted:
+            this.replace(exp.var(this.name))
+        if value and not value.quoted:
+            value.replace(exp.var(value.name))
```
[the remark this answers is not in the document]
VaggelisD: The RHS can vary from var, string literal, tuple of literals etc so it's not always as straightforward as parsing `<var> = <var>` afaict . With `parse_property` parsing the value/rhs as `_parse_bitwise` we catch all of these cases, with the exception that what should be parsed as var is now parsed as `exp.Column(exp.Identifier(...))`. 

For this reason, I thought it would be fine to reuse it and swap out these unquoted identifiers with `exp.Var`. In my upcoming commit I refactored this method completely and instead added an argument in `parse_property` that would force it to default to `exp.Var`

sqlglot/parser.py
```diff
@@ -6667,16 +6667,36 @@ def _parse_copy_parameters(self) -> t.List[exp.CopyParameter]:
...
+            expr.set("storage", self._parse_var_property())
         if self._match_text_seq("CREDENTIALS"):
-            # Snowflake supports CREDENTIALS = (...), while Redshift CREDENTIALS <string>
-            creds = (
-                self._parse_wrapped_options() if self._match(TokenType.EQ) else self._parse_field()
-            )
+            creds: t.Optional[exp.Expression] | t.List[t.Optional[exp.Expression]] = None
+            if self._match(TokenType.EQ):
+                # Snowflake case: CREDENTIALS = (...) . Note that empty parens are allowed too, thus the var fallback
+                creds = self._parse_wrapped_options() or [exp.Var(this=None)]
```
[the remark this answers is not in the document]
VaggelisD: I've defaulted to setting `creds = []` which will at least make it a non-None arg. Then the generator can use this as a proxy instead of generating an `exp.Var(None)`

sqlglot/parser.py
```diff
@@ -6671,12 +6677,15 @@ def _parse_credentials(self) -> t.Optional[exp.Credentials]:
...
-            expr.set("storage", self._parse_conjunction())
+            expr.set("storage", self._parse_property(parse_as_var=True))
         if self._match_text_seq("CREDENTIALS"):
-            # Snowflake supports CREDENTIALS = (...), while Redshift CREDENTIALS <string>
-            creds = (
-                self._parse_wrapped_options() if self._match(TokenType.EQ) else self._parse_field()
-            )
+            creds: t.Optional[exp.Expression] | t.List[t.Optional[exp.Expression]] = None
+            if self._match(TokenType.EQ):
+                # Snowflake case: CREDENTIALS = (...) . Note that empty parens are allowed too, thus the var fallback
```
georgesittas, line 6684: Outdated comment

georgesittas (commented):
I'd add a few tests to ensure the resulting ASTs contain `Var`s instead of `Identifier`s.

sqlglot/parser.py
```diff
@@ -6671,12 +6677,15 @@ def _parse_credentials(self) -> t.Optional[exp.Credentials]:
         expr = self.expression(exp.Credentials)
 
         if self._match_text_seq("STORAGE_INTEGRATION", advance=False):
-            expr.set("storage", self._parse_conjunction())
+            expr.set("storage", self._parse_property(parse_as_var=True))
```
georgesittas, line 6680: Why was this changed? The spec shows:

```
STORAGE_INTEGRATION = <integration_name>
```

This indicates that the right-hand side should be an `Identifier`, not a `Var`.

georgesittas:
Another setting we should properly fix & test is this one ([docs](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table)):

```
     INCLUDE_METADATA = ( <column_name> = METADATA$<field> [ , <column_name> = METADATA${field} ... ] )
```

Here `column_name` shouldn't be parsed as a `Var`. The same holds for the `METADATA$<field>` bit _I think_.

georgesittas:
> The `INCLUDE_METADATA` option will be parsed as an `exp.CopyParameter` with a RHS of `exp.Paren(exp.EQ(...))` which defaults to parsing everything as `exp.Column(exp.Identifier())` under it.
> 
> From what I understand the `METADATA$<field>` are metadata columns so technically the produced AST should be sound?

Yeah, if that's the case then it sounds good.

sqlglot/parser.py
georgesittas:
```suggestion
```

VaggelisD: Thanks for the catch. I added a `quote_identifiers` test to check this by proxy; If anything wasn't parsed properly it wouldn't (or would) be quoted, so this kills 2 birds with one stone. Wdyt?

sqlglot/parser.py
```diff
@@ -1810,7 +1810,7 @@ def _parse_property_before(self) -> t.Optional[exp.Expression]:
     def _parse_wrapped_properties(self) -> t.List[exp.Expression]:
         return self._parse_wrapped_csv(self._parse_property)
 
-    def _parse_property(self) -> t.Optional[exp.Expression]:
+    def _parse_property(self, parse_as_var: bool = False) -> t.Optional[exp.Expression]:
```
tobymao, line 1813: when would parse_as_var be False?

sqlglot/parser.py
```diff
@@ -1824,16 +1824,22 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
...
 
         index = self._index
-        key = self._parse_column()
+        key = self._parse_var(any_token=True) if parse_as_var else self._parse_column()
 
         if not self._match(TokenType.EQ):
             self._retreat(index)
             return self._parse_sequence_properties()
 
+        value = self._parse_bitwise() or self._parse_var(any_token=True)
```
tobymao, line 1833: why do we need this fallback _parse_var?

sqlglot/parser.py
VaggelisD: `parse_as_var` forces `_parse_property` to parse both LHS and RHS as vars if applicable which is our wanted behavior when parsing the copy options. 

However, we don't want to alter the existing semantics of `_parse_property` which originally parsed `this = _parse_column()` and `value = _parse_bitwise()`, so this defaults to False in all other cases except `parse_copy`

VaggelisD: I think this code introduced [changed here](https://github.com/tobymao/sqlglot/pull/2296) to allow for identifiers as keys

sqlglot/parser.py
VaggelisD: This conjuction pre-existed and by removing the fallback there's a T-SQL test breaking when trying to parse the option `PAD_INDEX = ON` (`parse_bitwise` failed because of reserve keyword afaict) so I transferred it as is

sqlglot/parser.py
tobymao: they are identifiers though or dots, not columns

tobymao: so maybe we never need parse_property to return a column?

georgesittas: IIRC I'd originally added `_parse_colum()` in the property parser to handle things like `delta.dataSkippingNumIndexedCols = 1` (following the linked [issue](https://github.com/tobymao/sqlglot/issues/2289)).

What we could do instead to improve the AST is to keep the `_parse_column()` call, but call `to_dot()` on the parsed expression, if it's a `Column` instance.

VaggelisD: Right @georgesittas this was already baked in the return statement, i.e we can have an `exp.Column` but it will always be transformed to `exp.Dot` before constructing the `exp.Property`

tobymao: there doesn't need to be a boolean, we should just always remove columns imo

georgesittas: Ah we already did that, lol. Good :)

VaggelisD: In my latest commit I removed the boolean arg altogether and instead always transform `exp.Column` to `exp.Var`, except the dotted identifier case of Databricks

VaggelisD:
I'll check the other dialects in case any identifiers have leaked there as well, I already have a faulty case like that from Databricks (`COPY_OPTIONS` and `FORMAT_OPTIONS`) that I'm hoping to have fixed too by tomorrow

tests/dialects/test_snowflake.py
```diff
@@ -1921,22 +1922,21 @@ def test_try_cast(self):
...
         )
         self.validate_identity(
-            """COPY INTO mytable FROM 'azure://myaccount.blob.core.windows.net/mycontainer/data/files' CREDENTIALS = (AZURE_SAS_TOKEN = 'token') ENCRYPTION = (TYPE = 'AZURE_CSE' MASTER_KEY = 'kPx...') FILE_FORMAT = (FORMAT_NAME = my_csv_format)"""
+            """COPY INTO mytable FROM 'azure://myaccount.blob.core.windows.net/mycontainer/data/files' CREDENTIALS = (AZURE_SAS_TOKEN='token') ENCRYPTION = (TYPE='AZURE_CSE' MASTER_KEY='kPx...') FILE_FORMAT = (FORMAT_NAME=my_csv_format)"""
         )
         self.validate_identity(
-            """COPY INTO mytable (col1, col2) FROM 's3://mybucket/data/files' FILES = ('file1', 'file2') PATTERN = 'pattern' FILE_FORMAT = (FORMAT_NAME = my_csv_format NULL_IF = ('str1', 'str2')) PARSE_HEADER = TRUE"""
+            """COPY INTO mytable (col1, col2) FROM 's3://mybucket/data/files' STORAGE_INTEGRATION = "storage" ENCRYPTION = (TYPE='NONE' MASTER_KEY='key') FILES = ('file1', 'file2') PATTERN = 'pattern' FILE_FORMAT = (FORMAT_NAME=my_csv_format NULL_IF=('')) PARSE_HEADER = TRUE"""
```
georgesittas, line 1934: Discussed in Slack, but thought I'd comment on this so we don't forget it: we need to ensure `my_csv_format` is parsed as an identifier, because it's something we can create using a [DDL statement](https://docs.snowflake.com/en/sql-reference/sql/create-file-format) in Snowflake. IIRC the alternative syntax is to specify a string, in which case we will parse it as such.

sqlglot/parser.py
```diff
@@ -1830,10 +1830,20 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
             self._retreat(index)
             return self._parse_sequence_properties()
 
+        # Transform the key to exp.Dot if it's dotted identifiers wrapped in exp.Column or to exp.Var otherwise
+        if isinstance(key, exp.Column):
+            key = key.to_dot() if len(key.parts) > 1 else exp.var(key.this.name)
```
georgesittas, line 1835:
```suggestion
            key = key.to_dot() if len(key.parts) > 1 else exp.var(key.name)
```

sqlglot/parser.py
```diff
@@ -1830,10 +1830,20 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
...
+
+        value = self._parse_bitwise() or self._parse_var(any_token=True)
+
+        # Transform the value to exp.Var if it was parsed as exp.Column(exp.Identifier())
+        if isinstance(value, exp.Column):
+            value = exp.var(value.this.name)
```
georgesittas, line 1841:
```suggestion
            value = exp.var(value.name)
```

georgesittas (approved):
So far this LGTM

sqlglot/generator.py
```diff
@@ -3860,9 +3860,15 @@ def _replace_line_breaks(self, string: str) -> str:
     def copyparameter_sql(self, expression: exp.CopyParameter) -> str:
         option = self.sql(expression, "this")
 
-        if option.upper() == "FILE_FORMAT":
+        upper = option.upper()
+        if upper == "FILE_FORMAT":
+            # Snowflake FILE_FORMAT
             values = self.expressions(expression, key="expression", flat=True, sep=" ")
             return f"{option} = ({values})"
+        elif upper == "COPY_OPTIONS" or upper == "FORMAT_OPTIONS":
```
georgesittas, line 3868:
```suggestion
        elif upper in ("COPY_OPTIONS", "FORMAT_OPTIONS"):
```

georgesittas (approved):

VaggelisD:
Pushed one more commit to:
- Fix Snowflake's & T-SQL's`FORMAT_NAME` to parse either as Identifier or string
- Fix T-SQL to parse `CREDENTIALS` as var key-values instead of identifiers
- Refactor generator for the varlen options

sqlglot/generator.py
```diff
@@ -3860,15 +3860,16 @@ def _replace_line_breaks(self, string: str) -> str:
     def copyparameter_sql(self, expression: exp.CopyParameter) -> str:
         option = self.sql(expression, "this")
 
-        upper = option.upper()
-        if upper == "FILE_FORMAT":
+        if isinstance(expression.expression, list):
```
georgesittas, line 3863: This shouldn't be a list if the argument is singular.

VaggelisD: Some of the options are singular and some are list, but the distinction between the two is helpful for us so we can change the output. Should I rename the argument to `expressions` but keep the semantics of `t.List` vs `t.Expression` ?

VaggelisD: Or I can add `expressions` alongside `expression` but one of the two arguments will always be empty, although not an issue

sqlglot/generator.py
```diff
@@ -3860,9 +3860,16 @@ def _replace_line_breaks(self, string: str) -> str:
     def copyparameter_sql(self, expression: exp.CopyParameter) -> str:
         option = self.sql(expression, "this")
 
-        if option.upper() == "FILE_FORMAT":
-            values = self.expressions(expression, key="expression", flat=True, sep=" ")
-            return f"{option} = ({values})"
+        if isinstance(expression.expression, list):
+            upper = option.upper()
+
+            # Snowflake FILE_FORMAT
```
georgesittas, line 3866: This comment doesn't explain the next line well. Add some more context to it, eg. "Snowflake's FILE_FORMAT property uses a space separator, while Databricks' uses commas".

sqlglot/generator.py
georgesittas: We should add a new `expressions` argument so these two cases are distinct. Having a single one to represent both cases is confusing and deviates from the codebase style.

georgesittas: > one of the two arguments will always be empty

This is fine, we do it all over the place anyway right?

VaggelisD: Right, but I was thinking that each COPY will have a (long ?) list of `exp.CopyParameters` so maybe it's more wasteful than other exprs that are created once. Still shouldn't make any difference

PR #3566, opened 2024-05-29
Fix(parser): edge case where TYPE_CONVERTERS leads to type instead of column

### LINKED ISSUE #3565: Fields with a name of some types (like number) are parsed incorrectly on Snowflake dialect
```
query = """
select
  id,
  int,
  number
from some.table
"""
sqlglot.parse(query, sqlglot.Dialects.SNOWFLAKE)
```

The `number` field is parsed as `DataType` instead of `Column`.
This was introduced on #3472 , as the new code replace it even if it is used as a name (and not a type)
It affects all names that match `DECIMAL` type on snowflake

georgesittas:
Fixes #3565. This was a bit tricky.

Current behavior in main:

```python
>>> import sqlglot
>>> sqlglot.parse_one("select number", "snowflake")
Select(
  expressions=[
    DataType(
      this=Type.DECIMAL,
      expressions=[
        DataTypeParam(
          this=Literal(this=38, is_string=False)),
        DataTypeParam(
          this=Literal(this=0, is_string=False))],
      nested=False)])
```

In this PR:

```python
>>> import sqlglot
>>> sqlglot.parse_one("select number", "snowflake")
Select(
  expressions=[
    Column(
      this=Identifier(this=number, quoted=False))])
```

The reason we get a `DataType` node in the main branch is because `_parse_types` now [returns](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L4397-L4402) a `DECIMAL` type whose `expressions` arg gets set by the corresponding [snowflake type converter](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/snowflake.py#L451). Moreover, for some reason we're using the existence of `expressions` in `_parse_type` as a proxy for [whether to return the data type](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L4202-L4204) vs retreating to parse a column.

This fix is a bit hacky in order to preserve the behavior as much as possible without altering the parser in a significant way.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4199,9 +4199,19 @@ class Parser(metaclass=_Parser):
 
                 return self.expression(exp.Cast, this=this, to=data_type)
 
-            # We may have set `expressions` using the `TYPE_CONVERTERS` mapping, in which
-            # case we shouldn't really move on with the data type, but instead retreat to
-            # parse a column reference; the index difference is used as a proxy to detect this
+            # The `expressions` arg gets set by the parser when we have something like DECIMAL(38, 0)
+            # in the input SQL. In that case, we'll produce these tokens: DECIMAL ( 38 , 0 )
+            #
+            # If the index difference here is greater than 1, that means the parser itself must have
+            # consumed additional tokens such as the DECIMAL's scale and precision in the above example.
+            #
+            # If it's not greater than 1, then it must be 1, because we've consumed at least the type
+            # keyword, meaning that the `expressions` arg of the `DataType` must have gotten set by a
+            # callable in the `TYPE_CONVERTERS` mapping. For example, Snowflake converts `DECIMAL` to
+            # `DECIMAL(38, 0))` in order to facilitate the data type's transpilation.
+            #
+            # In these cases, we don't really want to return the converted type, but instead retreat
+            # and try to parse a `Column` or `Identifier` in the section below.
             if data_type.expressions and self._index - index > 1:
                 self._retreat(index2)
                 return self._parse_column_ops(data_type)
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -4199,7 +4199,10 @@ def _parse_type(
 
                 return self.expression(exp.Cast, this=this, to=data_type)
 
-            if data_type.expressions:
+            # We may have set `expressions` using the `TYPE_CONVERTERS` mapping, in which
+            # case we shouldn't really move on with the data type, but instead retreat to
```
tobymao, line 4203: i don't really understand this comment, what does it mean set the expressions using type converters, and how does the index difference detect this

georgesittas: Fair enough.

The gist of it is that `expressions` gets set by the *parser* (i.e. *not* because of a converter) when you have something like `DECIMAL(38, 0)` in the input SQL. In that case, you'll have these tokens: `[DECIMAL, L_PAREN, NUMBER, COMMA, NUMBER, R_PAREN]`.

If the index difference in the condition is greater than 1, that means the parser itself must have consumed additional tokens such as the `DECIMAL`'s scale and precision here.

If it's not greater than 1, then it must be 1 (because we parsed a type in L4189), meaning that the `expressions` arg of the `DataType` got set by a type converter (e.g. Snowflake converts `DECIMAL` to `DECIMAL(38, 0)`).

tobymao: why would the expressions get set if the index is 1?

georgesittas: Because `DECIMAL` in Snowflake gets converted into an AST node that has `expressions` set to `(38, 0)` ([context](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/dialect.py#L1166-L1176)), to facilitate transpilation.

tobymao: let's add that to the comment

georgesittas: Done: https://github.com/tobymao/sqlglot/pull/3566/commits/ed01a1c4d60956ca39c52e77ad866f2668eacdf6

PR #3573, opened 2024-05-30
Update postgres.py

Kanishkaa07:
Added a support for the TOP keyword in postgres


georgesittas:
I don't understand the intent here, there are lots of unrelated changes to the description.

PR #3574, opened 2024-05-30
Update postgres.py

Kanishkaa07:
Added a support for the TOP keyword in postgres


georgesittas:
@Kanishkaa07 we're not going to accept a PR unless you communicate what you're trying to do. Please don't spam.

Kanishkaa07:
I'm trying to make the top keyword accept in postgres as postgres won't
accept top keyword instead it uses limit keyword

georgesittas:
Can you provide the relevant documentation that indicates this clause is supported in Postgres? I don't think it is.

Kanishkaa07:
It's not supported in postgres so i altered the code to make it supported
in postgres

PR #3576, opened 2024-05-30
fix(optimizer)!: Refactor struct star expansion in BQ

VaggelisD:
Let's assume we want to expand the following struct:

```
WITH
   tbl1 AS (SELECT STRUCT(1 AS fld1, Struct(3 AS fld2, Struct(4 AS fld3) AS lvl2) AS lvl1) AS col)
SELECT col.* FROM tbl1;
```

The recently merged PR would essentially flatten the struct by exploding the fields through a DFS traversal, i.e it would produce:

```
WITH
   tbl1 AS (...) # redacted CTE, same as initial query
SELECT tbl1.col.fld1 AS _f_0, 
       tbl1.col.lvl1.fld2 AS _f_2,
       tbl1.col.lvl1.lvl2.fld3 AS _f_3,
FROM tbl1
```

Even though the results are similar to what would be produced from the initial query, internally BigQuery only has to expand [the _top level_ expression](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_expression_star) and the rest of the "flattening" is performed implicitly; This is semantically equivalent to:

```
WITH
   tbl1 AS (...)
SELECT tbl1.col.fld1 AS fld1, 
       tbl1.col.lvl1 AS lvl1 
FROM tbl1
```

By changing our algorithm to perform the latter we completely remove the DFS struct traversal. Not only that, but it turns out that our `_f_` alias generation can break queries from being qualified; The following query can showcase this behavior:

```
WITH
  tbl1 AS (SELECT STRUCT(1 AS fld1, Struct(3 AS fld2, Struct(4 AS fld3) AS lvl2) AS lvl1) AS col),
  tbl2 AS (SELECT col.* FROM tbl1)
SELECT lvl1 FROM tbl2
```

In main, due to our custom `_f_` aliasing, `qualify` fails with:
```
sqlglot.errors.OptimizeError: Column '"lvl1"' could not be resolved
```

However, in this PR, `qualify` succeeds (the names are now preserved instead of generating a custom alias) and results to:
```
WITH 
`tbl1` AS (SELECT STRUCT(1 AS `fld1`, Struct(3 AS `fld2`, Struct(4 AS `fld3`) AS `lvl2`) AS `lvl1`) AS `col`), 
`tbl2` AS (SELECT `tbl1`.`col`.`fld1` AS `fld1`, `tbl1`.`col`.`lvl1` AS `lvl1` FROM `tbl1` AS `tbl1`) 
SELECT 
  `tbl2`.`lvl1` AS `lvl1` FROM `tbl2` AS `tbl2`
```

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -6,7 +6,7 @@ import typing as t
 from sqlglot import alias, exp
 from sqlglot.dialects.dialect import Dialect, DialectType
 from sqlglot.errors import OptimizeError
-from sqlglot.helper import seq_get, SingleValuedMapping, name_sequence
+from sqlglot.helper import seq_get, SingleValuedMapping
 from sqlglot.optimizer.annotate_types import TypeAnnotator
 from sqlglot.optimizer.scope import Scope, build_scope, traverse_scope, walk_in_scope
 from sqlglot.optimizer.simplify import simplify_parens
@@ -53,7 +53,6 @@ def qualify_columns(
     infer_schema = schema.empty if infer_schema is None else infer_schema
     dialect = Dialect.get_or_raise(schema.dialect)
     pseudocolumns = dialect.PSEUDOCOLUMNS
-    next_alias_name = name_sequence("_f_")
 
     for scope in traverse_scope(expression):
         resolver = Resolver(scope, schema, infer_schema=infer_schema)
@@ -78,7 +77,6 @@ def qualify_columns(
                     using_column_tables,
                     pseudocolumns,
                     annotator,
-                    next_alias_name,
                 )
             qualify_outputs(scope)
 
@@ -412,7 +410,6 @@ def _qualify_columns(scope: Scope, resolver: Resolver) -> None:
 
 def _expand_struct_stars(
     expression: exp.Dot,
-    next_alias_name: t.Callable[[], str],
 ) -> t.List[exp.Alias]:
     """[BigQuery] Expand/Flatten foo.bar.* where bar is a struct column"""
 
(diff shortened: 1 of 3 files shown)


georgesittas:
Chiming in here to provide some more context around this:

> Even though the results are similar to what would be produced from the initial query, internally BigQuery only has to expand [the top level expression](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_expression_star) and the rest of the "flattening" is **performed implicitly**

TL;DR - the previous implementation misinterpreted BQ's struct expansion semantics, most likely because of how they display the column names in their editor:

```sql
-- BQ displays a single row with a single column named s.id
with t as (select struct(struct('foo' as id) as s) as col) select col.* from t
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -441,46 +438,22 @@ def _expand_struct_stars(
...
 
-    # Each nested field caches its path in exp.Identifier parts,
-    # e.g. tbl.f0.f1.item -> [exp.Identifier(tbl), ..., exp.Identifier(item)]
-    field_path: t.Dict[int, t.List[exp.Identifier]] = {id(starting_struct): []}
-
-    # The keys to non-struct field entries in `nested_fields`
-    field_keys = []
-
-    stack = [(starting_struct, id(starting_struct))]
+    name_set = set()
```
georgesittas, line 441: Rename this to `taken_names` or something similar.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -441,46 +438,22 @@ def _expand_struct_stars(
...
-    for field_key in field_keys:
-        # The fully qualified path of each field is created by:
-        # - The path until the starting struct, if any (dot_parts)
-        # - The path from the starting struct until the nested field (field_path[field_key])
-        root, *parts = [
-            part.copy() for part in itertools.chain(dot_parts[:-1], field_path[field_key])
-        ]
+        this = field.this.copy()
+        root, *parts = [part.copy() for part in itertools.chain(dot_parts, [this])]
         new_column = exp.column(t.cast(exp.Identifier, root), table=dot_column.table, fields=parts)
```
georgesittas, line 455: I think if the `dot_column`'s table contained quotes, you'd drop them in `table=dot_column.table` because under the hood it simply invokes `text("table")`. You need to do `dot_column.args.get("table")`.

Let's 1) ensure this _does_ indeed drop the quotes in an example where the table name is quoted, 2) fix it and 3) add a test to avoid similar regressions.

Please also check for similar property method invocations to ensure we don't silently drop quotes.

georgesittas (approved):
Thanks for fixing this

PR #3580, opened 2024-06-01
Feat: eliminate join marks

mrhopko:
eliminate_join_marks.py added to optimizer.
converts oracle (+) join identifiers into left joins.


barakalon (commented):
Hey there 👋 

Impressive first PR!

Are you aware of the [scope module](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py)?

That might help avoid some of the gotchas I'm seeing in the PR. So before I review - have you considered writing the rule using scope?

mrhopko:
I have. I presume I should be using it when hunting for columns. Will take another look.


On 1 Jun 2024, at 21:09, barak ***@***.***> wrote:

﻿

@barakalon commented on this pull request.

Hey there 👋

Impressive first PR!

Are you aware of the scope module<https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/scope.py>?

That might help avoid some of the gotchas I'm seeing in the PR. So before I review - have you considered writing the rule using scope?

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+
+def eliminate_join_marks(ast: exp.Expression) -> exp.Expression:
+    """Remove join marks from an expression
+
+    Args:
+        ast (exp.Expression): The AST to remove join marks from
+
+    Returns:
+        exp.Expression: The AST with join marks removed"""
+    select_nodes = list(ast.find_all(exp.Select))
```
barakalon, line 173: You should build scope here and than traverse the scope nodes.

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+    """If the from clause needs to become a new join, find an appropriate table to use as the new from.
+    updates select in place
+
+    Args:
+        select (exp.Select): The select statement to update
+        new_join_dict (Dict[str, exp.Join]): The dictionary of new joins
+        old_join_dict (Dict[str, exp.Join]): The dictionary of old joins
+    """
+    old_from = select.args["from"]
+    if old_from.alias_or_name not in new_join_dict.keys():
```
tobymao, line 22: remove the calls to keys with in, they are unnecessary

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
+import typing as t
```
tobymao, line 1: i think eliminate_join_marks should be a transform in oracle, not in optimizer

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+import copy
+
+from sqlglot import expressions as exp
+from sqlglot.optimizer.scope import build_scope, find_all_in_scope
+
+
+def _update_from(
```
tobymao, line 8: can you move the public method to the top, and all the private methods to the bottom

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+    """Remove join marks from the where columns in this select statement
+    Converts them to joins and replaces any existing joins
+
+    Args:
+        node (exp.Select): The AST to remove join marks from
+
+    Returns:
+        exp.Select: The AST with join marks removed
+    """
+    scope = build_scope(select)
```
tobymao, line 128: as barak said, delete this, instead just traverse the scope

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+    select_scope = scope.expression
+    if not select_scope.args.get("where"):
+        return select
+    if not select_scope.args.get("joins"):
+        return select
+    old_joins: t.Dict[str, exp.Join] = {
+        join.alias_or_name: join for join in list(select_scope.args.get("joins", []))
+    }
+    new_joins: t.Dict[str, exp.Join] = {}
+    for node in find_all_in_scope(scope.expression, exp.Column):
```
tobymao, line 140: you should just be able to do scope.find_all

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Args:
+        eq (exp.Binary): The equality expression to convert to a join
+
+    Returns:
+        Optional[exp.Join]: The join expression if the equality contains a join mark (otherwise None)
+    """
+    if not (isinstance(eq.left, exp.Column) or isinstance(eq.right, exp.Column)):
+        return None
+    new_eq = copy.deepcopy(eq)
```
tobymao, line 99: eq.copy() is already a deepcopy, don't import copy

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+    left_has_join_mark = _has_join_mark(eq.left)
+    right_has_join_mark = _has_join_mark(eq.right)
+
+    if left_has_join_mark:
+        new_eq.left.set("join_mark", False)
+        assert isinstance(new_eq.left, exp.Column)
```
tobymao, line 105: delete these asserts

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+
+def _has_join_mark(col: exp.Expression) -> bool:
+    """Check if the column has a join mark
+
+    Args:
+        col (exp.Column): The column to check
+    """
+    if not isinstance(col, exp.Column):
+        return False
+    result = col.args.get("join_mark", False)
```
tobymao, line 80: when is join_mark not a bool?

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+        join_dict[join.alias_or_name] = join
+    return join_dict
+
+
+def _clean_binary_node(node: exp.Expression):
+    """if the node is left with only one child, promote the child to the parent node.
+    transformation is done in place.
+
+    Args:
+        node (exp.Expression): The node to clean"""
```
tobymao, line 64: delete all of these doc string type annotations, the method is already annotated

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+                this=join_dict[join.alias_or_name].args["on"],
+                expression=join.args["on"],
+            ),
+        )
+    else:
+        join_dict[join.alias_or_name] = join
+    return join_dict
+
+
+def _clean_binary_node(node: exp.Expression):
```
tobymao, line 59: -> None

sqlglot/optimizer/eliminate_join_marks.py
```diff
@@ -0,0 +1,184 @@
...
+        join (exp.Join): The join to add to the dictionary
+        join_dict (Dict[str, exp.Join]): dictionary of joins where str is join.alias_or_name
+
+    Returns:
+        Dict[str, exp.Join]: The updated dictionary of joins
+    """
+    if join.alias_or_name in join_dict.keys():
+        join_dict[join.alias_or_name].set(
+            "on",
+            exp.And(
```
tobymao, line 49: exp.and_(...)

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: moved

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: removed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: changed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: changed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: removed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: removed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: changed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: deleted

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: changed

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: now using build_scope in top level function and not in inner function.

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: moved build scope to outer function, removed from inner.

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: It makes more sense there, however it does mean you now have a reference to the optimize module from dialects (due to the use of scope). 

Where should the function live? below the class or as part of the oracle generator?

tobymao: i think ideally at oracle parse time to get rid of the old syntax

tests/fixtures/optimizer/eliminate_join_marks.sql
```diff
@@ -0,0 +1,41 @@
...
+# dialect: oracle
+SELECT T1.d, T2.c FROM T1, T2 WHERE T1.x = T2.x (+) and T2.y IS NULL;
+SELECT T1.d, T2.c FROM T1 LEFT JOIN T2 ON T1.x = T2.x WHERE T2.y IS NULL;
+  
+# title: Additional where clause  
+# dialect: oracle
+SELECT T1.d, T2.c FROM T1, T2 WHERE T1.x = T2.x (+) and T1.Z > 4;
+SELECT T1.d, T2.c FROM T1 LEFT JOIN T2 ON T1.x = T2.x WHERE T1.Z > 4;
+  
+# title: various select tests 1
```
georgesittas, line 22: These titles are not very helpful, let's remove them.

tests/fixtures/optimizer/eliminate_join_marks.sql
```diff
@@ -0,0 +1,41 @@
+ 
+# title: Constant in join  
+# dialect: oracle
+SELECT T1.d, T2.c FROM T1, T2 WHERE T1.x = T2.x (+) and T2.y (+) > 5;
+SELECT T1.d, T2.c FROM T1 LEFT JOIN T2 ON T1.x = T2.x AND T2.y > 5;
+  
+# title: Null in Join Predicte
```
georgesittas, line 7: There's a typo here, "Predicte" needs to be predicate

georgesittas (commented):
@mrhopko I'll review this as well when the previous review comments are addressed to avoid commenting on the same things.

I'd like to understand this transformation a bit more before moving this to the parsing stage, as Toby suggested in a different comment. Can you lay out the logic in simple terms so that it's easier to verify the correctness? Docs would also help here.

tests/fixtures/optimizer/eliminate_join_marks.sql
barakalon: Or use more descriptive names :)

sqlglot/optimizer/eliminate_join_marks.py
mrhopko: I've moved the function to the oracle dialect module, but not integrated it into the generator or parser yet due to other comments raised.

tests/fixtures/optimizer/eliminate_join_marks.sql
mrhopko: This file is now removed as the test have moved to the oracle dialect test suite.

tests/fixtures/optimizer/eliminate_join_marks.sql
mrhopko: file removed

mrhopko:
@georgesittas - I've added a description of what the code is doing in the docstring. Let me know if you need further explanation.

tests/dialects/test_oracle.py
```diff
@@ -1,11 +1,16 @@
 from sqlglot import exp
 from sqlglot.errors import UnsupportedError
 from tests.dialects.test_dialect import Validator
+from sqlglot.dialects.oracle import eliminate_join_marks
 
 
 class TestOracle(Validator):
     dialect = "oracle"
 
+    def validate(self, transform, sql, target):
```
tobymao, line 10: let's delete this helper method and just do a for loop with sub tests

for in, out in [(..., ...), ...]

tests/test_optimizer.py
```diff
@@ -474,6 +494,13 @@ def test_eliminate_joins(self):
             pretty=True,
         )
 
+    # def test_eliminate_join_marks(self):
```
tobymao, line 497: delete this

sqlglot/transforms.py
```diff
@@ -154,7 +154,8 @@ def remove_precision_parameterized_types(expression: exp.Expression) -> exp.Expr
     """
     for node in expression.find_all(exp.DataType):
         node.set(
-            "expressions", [e for e in node.expressions if not isinstance(e, exp.DataTypeParam)]
```
tobymao, line 157: there are a lot of unnecessary changes, can you revert them all?

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,181 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+            - remove the predicate from the where clause
+            - convert the predicate to a join with the (+) side as the left join table
+            - replace the existing join with the new join
+
+    Args:
+        ast: The AST to remove join marks from
+
+    Returns:
+       The AST with join marks removed"""
+    ast_new = ast.copy()
```
tobymao, line 308: let's get rid of the copy

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,181 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
             if len(expression.args.get("actions", [])) > 1:
                 return f"ADD ({actions})"
             return f"ADD {actions}"
+
+
+def eliminate_join_marks(ast: exp.Expression) -> exp.Expression:
```
tobymao, line 287: i think this could perhaps be called in the parser. once everything is done, if we've parsed a join_mark (which happens in _parse_column),

we can call this method to transform and remove the join marker.

if we can handle all cases of join marker elimination, maybe we can get rid of the whole concept?

tests/dialects/test_oracle.py
mrhopko: removed helper and converted to loop

tests/test_optimizer.py
mrhopko: deleted

sqlglot/transforms.py
mrhopko: reverted

sqlglot/dialects/oracle.py
mrhopko: removed

mrhopko:
Not yet moved the eliminate_join_marks to the oracle parser - will push a version in the next day or so.

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+            if not isinstance(predicate, exp.Binary):
+                continue
+            predicate_parent = predicate.parent
+
+            join_on = predicate.pop()
+            new_join = _equality_to_join(
+                join_on, old_joins=old_joins, old_from=select_scope.args["from"]
+            )
+            if new_join:
+                new_joins = _update_join_dict(new_join, new_joins)
```
georgesittas, line 452: How can this ever be falsey? Why would that happen?

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+            if predicate_parent:
+                _clean_binary_node(predicate_parent)
```
georgesittas, line 454: Let's not create short helpers like this one when they're not used in multiple places. It makes understanding the code harder because you have to jump back and forth between use site and implementation site. I also found the verb "clean" to be somewhat confusing in this context.

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+    _update_from(select_scope, new_joins, old_joins)
+    replacement_joins = [new_joins.get(join.alias_or_name, join) for join in old_joins.values()]
+    select_scope.set("joins", replacement_joins)
+    where = select_scope.args["where"]
+    if where:
```
georgesittas, line 459: This branch is redundant, we've already checked if the `where` arg is present upon entry.

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+        if not where.this:
+            select_scope.set("where", None)
```
georgesittas, line 461:
```suggestion
        if not where.this:
            where.pop()
```

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+        new_from_name = list(old_join_dict.keys() - new_join_dict.keys())[0]
+        new_from_this = old_join_dict[new_from_name].this
+        new_from = exp.From(this=new_from_this)
+        del old_join_dict[new_from_name]
+        select.set("from", new_from)
+    else:
+        raise ValueError("Cannot determine which table to use as the new from")
+
+
+def _update_join_dict(join: exp.Join, join_dict: t.Dict[str, exp.Join]) -> t.Dict[str, exp.Join]:
```
georgesittas, line 343: Ditto about inlining this.

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+    return ast
+
+
+def _update_from(
+    select: exp.Select,
+    new_join_dict: t.Dict[str, exp.Join],
+    old_join_dict: t.Dict[str, exp.Join],
+) -> None:
+    """If the from clause needs to become a new join, find an appropriate table to use as the new from.
+    updates select in place
```
georgesittas, line 322: Can you explain the logic here a bit? Why would the `FROM` need to become a new join? Is this because you can have:

```sql
SELECT * FROM a, b WHERE a.x (+) = b.x
```

where `b` becomes the left relation in the `LEFT JOIN`?

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+        new_eq.right.set("join_mark", False)
+        join_on = new_eq.right.table
+    else:
+        return None
+
+    join_this = old_joins.get(join_on, old_from).this
+    return exp.Join(this=join_this, on=new_eq, kind="LEFT")
+
+
+def _eliminate_join_marks_from_scope(scope) -> None:
```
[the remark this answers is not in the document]
mrhopko: type hint requires a top level import which creates circular import issues.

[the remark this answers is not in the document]
georgesittas: What if we put that import within a `if t.TYPE_CHECKING` branch?

georgesittas:
Hey @mrhopko, do you plan to finish this? We can you help you get it to the finish line, but another pass would be very helpful to clean up some parts where the logic is currently incorrect.

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+        join.alias_or_name: join for join in list(select_scope.args.get("joins", []))
+    }
+    new_joins: t.Dict[str, exp.Join] = {}
+
+    for node in scope.find_all(exp.Column):
+        if _has_join_mark(node):
+            # find the predicate that contains the join mark
+            predicate = node.find_ancestor(exp.Predicate)
+            if not isinstance(predicate, exp.Binary):
+                continue
```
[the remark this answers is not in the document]
georgesittas: Also noting this bit:

> If two tables are joined over multiple join conditions, you must use the (+) operator in all or none of these conditions. A join with mixed syntax styles runs as an inner join, without warning.

Which is actually part of Redshift's [docs](https://docs.aws.amazon.com/redshift/latest/dg/r_WHERE_oracle_outer.html). Redshift's supposed to mimic Oracle's semantics AFAICT, so I'd expect this rule to also apply to Oracle.

If I understand correctly, this means that we should also avoid converting the `(+)` syntax if there's at least one condition involving the joined tables that doesn't use it. Otherwise you'd be silently changing this "A join with mixed syntax styles runs as an inner join, without warning." into "A join with mixed syntax styles will be converted into an outer join whose conditions will contain the marked columns, and the rest of them will remain as a filter".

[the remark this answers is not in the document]
georgesittas: Btw, Redshift supporting this operator also means that we should probably factor this logic out into `transforms.py` so that it can be reused by both it and Oracle.

mrhopko:
> Hey @mrhopko, do you plan to finish this? We can you help you get it to the finish line, but another pass would be very helpful to clean up some parts where the logic is currently incorrect.

I intend to pick this up over the weekend.

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
             if len(expression.args.get("actions", [])) > 1:
                 return f"ADD ({actions})"
             return f"ADD {actions}"
+
+
+def eliminate_join_marks(ast: exp.Expression) -> exp.Expression:
+    from sqlglot.optimizer.scope import build_scope
```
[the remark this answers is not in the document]
mrhopko: changed

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+
+    Args:
+        ast: The AST to remove join marks from
+
+    Returns:
+       The AST with join marks removed"""
+    scopes = build_scope(ast)
+    if not scopes:
+        return ast
+    for scope in scopes.traverse():
```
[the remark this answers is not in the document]
mrhopko: changed

sqlglot/dialects/oracle.py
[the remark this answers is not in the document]
mrhopko: changed

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+    Updates scope in place.
+
+    Args:
+        scope: The scope to remove join marks from
+    """
+    select_scope = scope.expression
+    if not select_scope.args.get("where"):
+        return
+    if not select_scope.args.get("joins"):
+        return
```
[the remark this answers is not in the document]
mrhopko: changed

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+
+    # dictionaries used to keep track of joins to be replaced
+    old_joins: t.Dict[str, exp.Join] = {
+        join.alias_or_name: join for join in list(select_scope.args.get("joins", []))
+    }
```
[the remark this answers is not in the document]
mrhopko: removed

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+def _has_join_mark(col: exp.Expression) -> bool:
+    """Check if the column has a join mark
+
+    Args:
+        The column to check
+    """
+    if not isinstance(col, exp.Column):
+        return False
+    result = col.args.get("join_mark", False)
+    return result
```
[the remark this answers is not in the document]
mrhopko: removed

sqlglot/dialects/oracle.py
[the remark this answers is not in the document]
mrhopko: > This doesn't look correct. AFAICT, this `continue` being reachable implies that there's an execution path where we may do a partial elimination of the join markers, ending up with a query that combines both the marker syntax and the `FROM`-`JOIN` syntax. However, this is invalid Oracle SQL according to the [docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Joins.html#GUID-29A4584C-0741-4E6A-A89B-DCFAA222994A):
> 
> > You cannot specify the (+) operator in a query block that also contains FROM clause join syntax.
> 
> Perhaps we should instead take this assumption for granted, following the usual garbage in - garbage out philosophy.

This does make that assumption and continues processing rather than returning an error.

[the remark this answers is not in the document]
mrhopko: we should be using the oracle docs: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Joins.html

There seems to be some contradiction with what you describe.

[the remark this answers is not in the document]
mrhopko: Any conditions imposed by the docs should be included as a test case

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+
+    # dictionaries used to keep track of joins to be replaced
+    old_joins: t.Dict[str, exp.Join] = {
+        join.alias_or_name: join for join in list(select_scope.args.get("joins", []))
+    }
+    new_joins: t.Dict[str, exp.Join] = {}
+
+    for node in scope.find_all(exp.Column):
+        if _has_join_mark(node):
+            # find the predicate that contains the join mark
```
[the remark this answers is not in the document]
mrhopko: removed

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+            predicate = node.find_ancestor(exp.Predicate)
+            if not isinstance(predicate, exp.Binary):
+                continue
+            predicate_parent = predicate.parent
+
+            join_on = predicate.pop()
+            new_join = _equality_to_join(
+                join_on, old_joins=old_joins, old_from=select_scope.args["from"]
+            )
```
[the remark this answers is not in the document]
mrhopko: changed to predicate

sqlglot/dialects/oracle.py
```diff
@@ -282,3 +282,180 @@ def add_column_sql(self, expression: exp.AlterTable) -> str:
...
+    """Convert an equality predicate to a join if it contains a join mark
+
+    Args:
+        eq: The equality expression to convert to a join
+
+    Returns:
+        The join expression if the equality contains a join mark (otherwise None)
+    """
+    if not (isinstance(eq.left, exp.Column) or isinstance(eq.right, exp.Column)):
+        return None
```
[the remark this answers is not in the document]
mrhopko: Per the docs "The (+) operator can appear only in the WHERE clause or, in the context of left-correlation (when specifying the TABLE clause) in the FROM clause, and can be applied only to a column of a table or view"

sqlglot/dialects/oracle.py
mrhopko: predicate to join can return None if there are no join marks or the left/right are not columns.

sqlglot/dialects/oracle.py
mrhopko: removed

sqlglot/dialects/oracle.py
mrhopko: removed

sqlglot/dialects/oracle.py
mrhopko: changed

sqlglot/dialects/oracle.py
mrhopko: inlined

sqlglot/dialects/oracle.py
mrhopko: presently, the transformation is applied after the oracle parser. So
SELECT * FROM a,b WHERE a.x (+) = b.x 
becomes
SELECT * FROM a JOIN b WHERE a.x (x) = b.x
But the query should become
SELECT * FROM b LEFT JOIN a ...

sqlglot/dialects/oracle.py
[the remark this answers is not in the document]
mrhopko: From the docs
"The (+) operator can be applied only to a column, not to an arbitrary expression. However, an arbitrary expression can contain one or more columns marked with the (+) operator."
So I think you are correct, the lhs/rhs could be an expression and needs consideration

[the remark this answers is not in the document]
mrhopko: this is now handled, tests added.

mrhopko:
@georgesittas - branch is complete from my perspective. Let me know what you want to do.

georgesittas:
Thanks, a bit busy at the moment but planning to take another look soon. I was leaning on not including this at the parsing stage, because I don't really fully understand the transformation yet - i.e., this would be an opt-in transformation that users would need to apply to their queries.

So, maybe I'll do one more review pass and try to take it to the finish line if the only changes are related to refactoring.

georgesittas:
Decided to merge this and will attempt to clean it up a bit soon. Again, thanks for the PR @mrhopko!

Issue #3589, opened 2024-06-04
bug (postgres tranform doris) about string_agg function , no separator key word for doris , so delete it 


Toms1999:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`
- Check if the issue still exists on main

**Fully reproducible code snippet**
```python
from sqlglot import transpile


sql = """SELECT STRING_AGG(ID ,',') FROM EMPLOYEES """


result = transpile(sql , 'postgres', 'doris')[0]


print(result)
```
**Official Documentation**
[group_concat](https://doris.apache.org/zh-CN/docs/2.0/sql-manual/sql-functions/aggregate-functions/group-concat?_highlight=group_)

PR #3590, opened 2024-06-04
fix(doris) , mysql group_concat function different doris , it is not support separator keyword

Toms1999:
#3589


tobymao:
add a test

sqlglot/dialects/doris.py
```diff
@@ -49,6 +49,8 @@ class Generator(MySQL.Generator):
             exp.ArrayUniqueAgg: rename_func("COLLECT_SET"),
             exp.CurrentTimestamp: lambda self, _: self.func("NOW"),
             exp.DateTrunc: lambda self, e: self.func("DATE_TRUNC", e.this, unit_to_str(e)),
+            exp.GroupConcat: lambda self,
+                                    e: f"""GROUP_CONCAT({self.sql(e, "this")} , {self.sql(e, "separator") or "','"})""",
```
tobymao, line 53: exp.func("GROUP_CONCAT", e.this, e.args.get("seprator") or ",")

Toms1999: yes , but need to add ''

tobymao: exp.Literal.string(",")

tobymao:
are you going to add a test and make style?

Toms1999:
@tobymao 
I tried to do so, but it never passed. I don't know why. I'm sorry

PR #3595, opened 2024-06-04
fix: optimize absorb_and_eliminate and remove_complements

barakalon:
Some of the simplification rules become _very_ expensive on deeply nested conditions (e.g. a1 AND a2 AND ... AND a100)

Here's something that'll probably be controversial - this sets a configurable threshold to not even bother trying to simplify these.

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -457,7 +457,7 @@ def absorb_and_eliminate(expression, root=True):
 
         # Initialize lookup tables:
         # Set of all operands, used to find complements for absorption.
-        hashed = set()
+        op_set = set()
         # Sub-operands, used to find subsets for absorption.
         subops = defaultdict(list)
         # Pairs of complements, used for elimination.
@@ -465,7 +465,7 @@ def absorb_and_eliminate(expression, root=True):
 
         # Populate the lookup tables
         for op in ops:
-            hashed.add(op)
+            op_set.add(op)
 
             if not isinstance(op, kind):
                 # In cases like: A OR (A AND B)
@@ -492,10 +492,10 @@ def absorb_and_eliminate(expression, root=True):
             a, b = op.unnest_operands()
 
             # Absorb
-            if isinstance(a, exp.Not) and a.this in hashed:
+            if isinstance(a, exp.Not) and a.this in op_set:
                 a.replace(exp.true() if kind == exp.And else exp.false())
                 continue
-            if isinstance(b, exp.Not) and b.this in hashed:
+            if isinstance(b, exp.Not) and b.this in op_set:
                 b.replace(exp.true() if kind == exp.And else exp.false())
                 continue
             superset = set(op.flatten())
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -105,7 +105,7 @@ def simplify(
         node = simplify_not(node)
         node = flatten(node)
         node = simplify_connectors(node, root, max_operands)
-        node = remove_complements(node, root, max_operands)
+        node = remove_complements(node, root)
         node = simplify_coalesce(node)
         node.parent = expression.parent
         node = simplify_literals(node, root, max_operands)
@@ -350,7 +350,7 @@ def _simplify_comparison(expression, left, right, or_=False):
     return None
 
 
-def remove_complements(expression, root=True, max_operands=float("inf")):
+def remove_complements(expression, root=True):
     """
     Removing complements.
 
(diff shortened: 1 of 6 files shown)


georgesittas (approved):
Let's test this?

sqlglot/optimizer/simplify.py
```diff
@@ -407,28 +409,49 @@ def absorb_and_eliminate(expression, root=True):
...
-                        a.replace(aa)
-                        b.replace(aa)
-                    elif ab in rhs and (is_complement(aa, ba) or is_complement(aa, bb)):
-                        a.replace(ab)
-                        b.replace(ab)
+        ops = tuple(expression.flatten())
+
+        complements = set()
+        subops = defaultdict(list)
+        pairs = defaultdict(list)
```
VaggelisD, line 416: The previous algorithm was a bit easier to follow, is it possible to leave a comment explaining the usecase of these collections?

sqlglot/optimizer/simplify.py
```diff
@@ -407,28 +409,49 @@ def absorb_and_eliminate(expression, root=True):
...
+
+        # Build the indexes
+        for op in ops:
+            complements.add(op)
+
```
VaggelisD, line 421: Can this be rewritten to merge the 2 loops in 1? For instance:
- In the first loop, switch the order of the `if` and `else` e.g.:

```Python3
        for op in ops:
            complements.add(op)
            
            if not isinstance(op, kind):
               subops[op].append({op})
              continue
           ...
```
- The second loop can maybe be moved upwards

sqlglot/optimizer/simplify.py
```diff
@@ -407,28 +409,49 @@ def absorb_and_eliminate(expression, root=True):
...
+                for i in subset:
+                    subops[i].append(subset)
+
+                a, b = op.unnest_operands()
+                if isinstance(a, exp.Not):
+                    pairs[frozenset((a.this, b))].append((op, b))
+                if isinstance(b, exp.Not):
+                    pairs[frozenset((a, b.this))].append((op, a))
+            else:
+                subops[op].append({op})
```
VaggelisD, line 433: The second loop seems to go over only over the `isinstance(op, kind)` ops, are the `else` subops being utilized?

VaggelisD (commented):
Very interesting! I had a couple of questions:

- I'm presuming that the cost of the simplification on deeply nested conds was exposed by a benchmark, if so, could you share the results & the test case? I've been thinking that we could benefit from amassing a bench suite that stresses each component individually, it will allows to catch regressions and prove the effectiveness of an optimization against historical runs 

- How is the threshold baked in? I may be missing it

barakalon:
@VaggelisD Ah! sorry - this became a WIP after a side discussion on slack. If you look at the original commit, my initial approach had the threshold. Then I decided to try and optimize the existing code.

Planning on doing some cleaning, commenting, benchmarking, etc. today

barakalon:
Added a new benchmarking script for the optimizer.

With this branch:
|       Benchmark |    Duration (s) |
| --------------- | --------------- |
|            tpch |          0.6284 |
|           tpcds |          8.1603 |
|    condition_10 |          0.0042 |
|   condition_100 |          0.0978 |
|  condition_1000 |          4.0751 |

With main:
|       Benchmark |    Duration (s) |
| --------------- | --------------- |
|            tpch |          0.6513 |
|           tpcds |          8.4834 |
|    condition_10 |          0.0048 |
|   condition_100 |          0.2624 |
|   condition_1000 |    ... |

I skipped condition_1000 for main after waiting a few minutes for it to complete.

georgesittas:
Sick

sqlglot/optimizer/simplify.py
```diff
@@ -407,28 +409,49 @@ def absorb_and_eliminate(expression, root=True):
     if isinstance(expression, exp.Connector) and (root or not expression.same_parent):
         kind = exp.Or if isinstance(expression, exp.And) else exp.And
```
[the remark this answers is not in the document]
barakalon: Sure, I can add that

sqlglot/optimizer/simplify.py
barakalon: sure

sqlglot/optimizer/simplify.py
barakalon: Added some comments

sqlglot/optimizer/simplify.py
barakalon: Added a comment

sqlglot/optimizer/simplify.py
VaggelisD: Not sure if this is still WIP or if I confused you, but I meant the 2 `for op in ops` loops here

VaggelisD: I'm guessing it's actually not feasible because you need the indexes built before looping again, nevermind

barakalon: Ah, yeah. Exactly

tobymao:
@barakalon do you want to also have the maximal ops bypass? 4 seconds is still a long time

barakalon:
Max ops would be great for my use case, yeah. If you're down, I'll add it.

tobymao:
@barakalon yea let's add it

georgesittas (approved):
great work 🎉

sqlglot/optimizer/simplify.py
```diff
@@ -404,31 +450,63 @@ def absorb_and_eliminate(expression, root=True):
...
-                        a.replace(aa)
-                        b.replace(aa)
-                    elif ab in rhs and (is_complement(aa, ba) or is_complement(aa, bb)):
-                        a.replace(ab)
-                        b.replace(ab)
+        ops = tuple(expression.flatten())
+
+        # Initialize lookup tables:
+        # Set of all operands, used to find complements for absorption.
+        hashed = set()
```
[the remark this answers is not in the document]
barakalon: named it `op_set` 🤷

tobymao (approved):

PR #3598, opened 2024-06-05
Feat: introduce risingwave as a dialect

neverchanje:
RisingWave is a streaming database with PostgreSQL compatibility. We have previously added support for dbt and are now exploring the integration of SQLMesh due to its impressive state management capabilities.

However, RisingWave currently doesn't support "locking reads" (i.e., FOR UPDATE). To address this, I created a custom dialect for RisingWave that inherits all behaviors from PostgreSQL except for LOCKING_READS_SUPPORTED. I tested this dialect with SQLMesh, and it worked smoothly.


sqlglot/dialects/__init__.py
```diff
@@ -85,3 +85,4 @@ class Generator(Generator):
 from sqlglot.dialects.teradata import Teradata
 from sqlglot.dialects.trino import Trino
 from sqlglot.dialects.tsql import TSQL
+from sqlglot.dialects.risingwave import RisingWave
```
tobymao, line 88: sort these alphabetically

sqlglot/dialects/dialect.py
```diff
@@ -65,6 +65,7 @@ class Dialects(str, Enum):
     TERADATA = "teradata"
     TRINO = "trino"
     TSQL = "tsql"
+    RISINGWAVE = "risingwave"
```
tobymao, line 68: ditto

tobymao:
will you make a corresponding sqlmesh pr?

sqlglot/dialects/risingwave.py
```diff
@@ -0,0 +1,12 @@
+from sqlglot.dialects.postgres import Postgres
+
+
+class RisingWave(Postgres):
+    class Tokenizer(Postgres.Tokenizer):
+        pass
+
+    class Parser(Postgres.Parser):
+        pass
+
```
georgesittas, line 10: These are redundant, no?

georgesittas (approved):
Thanks @neverchanje, one comment from me.

barakalon (commented):
I'd add a test module for this, even if it feels silly while this is so small. The number of extensions might grow, and we should reduce friction to adding tests.

georgesittas:
See https://github.com/tobymao/sqlglot/commit/51408172ce940b6ab0ad783d98e632d972da6a0a, cc @neverchanje

neverchanje:
Thanks @georgesittas !

PR #3600, opened 2024-06-05
fix(hive): generate TRUNC for TimestampTrunc

Toms1999:


Toms1999:
fix #3598

Toms1999:
fix #3601

VaggelisD:
Hey @Toms1999, please don't attempt to multiple issues in the same PR, it's confusing. Please revert the 3601 fix (which doesn't seem correct, but don't want to expand here) so we can review just the 3596 here.

For the Trino <-> Hive date truncation, have you tested that your solution works? Also, please link the relevant documentation that verifies your fixes on each PR.

PR #3602, opened 2024-06-05
feat(postgres): Support DIV() func for integer division

### LINKED ISSUE #3601: parsing div() for postgresql throws error
**Summary**

Trying to parse a sql statement using the DIV() method but encountering an error.

**Sample Code**

```python
from sqlglot import parse_one

parsed_content = parse_one('select div(9,4)', read='postgres')
```

**Error**

sqlglot.errors.ParseError: Required keyword: 'expressions' missing for <class 'sqlglot.expressions.Aliases'>. Line 1, Col: 12.
  select div(9,4)

VaggelisD:
Fixes #3601 

This PR introduces support for Postgres's `DIV()` function which performs integer division. It mirrors the BigQuery implementation which stores the args in `exp.IntDiv` and generates it back to the function representation.

[Postgres Div](https://www.postgresqltutorial.com/postgresql-math-functions/postgresql-div/)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -632,7 +632,7 @@ class Postgres(Dialect):
             this = expression.this
 
             # Postgres casts DIV() to decimal for transpilation but when roundtripping it's superfluous
-            if isinstance(this, exp.IntDiv) and expression.to.is_type(exp.DataType.Type.DECIMAL):
+            if isinstance(this, exp.IntDiv) and expression.to == exp.DataType.build("decimal"):
                 return self.sql(this)
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -8,6 +8,7 @@ from sqlglot.dialects.dialect import (
     Dialect,
     JSON_EXTRACT_TYPE,
     any_value_to_max_sql,
+    binary_from_function,
     bool_xor_sql,
     datestrtodate_sql,
     build_formatted_time,
@@ -329,6 +330,7 @@ class Postgres(Dialect):
             "REGTYPE": TokenType.OBJECT_IDENTIFIER,
             "FLOAT": TokenType.DOUBLE,
         }
+        KEYWORDS.pop("DIV")
 
         SINGLE_TOKENS = {
             **tokens.Tokenizer.SINGLE_TOKENS,
@@ -347,6 +349,9 @@ class Postgres(Dialect):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "DATE_TRUNC": build_timestamp_trunc,
+            "DIV": lambda args: exp.cast(
+                binary_from_function(exp.IntDiv)(args), exp.DataType.Type.DECIMAL
+            ),
             "GENERATE_SERIES": _build_generate_series,
             "JSON_EXTRACT_PATH": build_json_extract_path(exp.JSONExtract),
             "JSON_EXTRACT_PATH_TEXT": build_json_extract_path(exp.JSONExtractScalar),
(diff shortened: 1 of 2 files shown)


georgesittas:
Transpiling _from_ BigQuery may work because it doesn't use typed division semantics, but that's coincidental.

VaggelisD:
Sorry for the force pushes, had to resolve conflicts and also amended some changes by accident. Leaving the following for future reference:

From my search, it is Postgres, DuckDB and BigQuery which support integer division. Out of those, Postgres's `DIV()` always results in a `DECIMAL`; In DuckDB and BigQuery, the types of the operands can alter the result type.

For this reason, it's not possible to safely transpile DuckDB's/BigQuery's `exp.IntDiv` without type inference. As for Postgres, this PR always casts `exp.IntDiv` to a decimal and removes the cast if it's roundtripped.

sqlglot/dialects/postgres.py
```diff
@@ -621,3 +627,12 @@ def datatype_sql(self, expression: exp.DataType) -> str:
                     return f"{self.expressions(expression, flat=True)}[{values}]"
                 return "ARRAY"
             return super().datatype_sql(expression)
+
+        def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
+            this = expression.this
+
+            # Postgres casts DIV() to decimal for transpilation but when roundtripping it's superfluous
+            if isinstance(this, exp.IntDiv) and expression.to.is_type(exp.DataType.Type.DECIMAL):
+                return self.sql(this)
```
georgesittas, line 636: I think we need to slightly change this condition so that it's `expression.to == exp.DataType.build("decimal")`, otherwise you'll get rid of a cast to `DECIMAL` that includes a scale / precision.

VaggelisD: You mean removing a user-inputted cast such as `CAST(DIV(...) AS DECIMAL(x, y))`? 

Can't we assume that the user cast will also wrap our generated one, resulting in `CAST(CAST(DIV(...) AS DECIMAL) AS DECIMAL(X, Y))`? So the condition will hold only for our generated cast afaict

georgesittas (approved):

sqlglot/dialects/postgres.py
VaggelisD: Actually this could be a problem coming from _other_ dialects I think

VaggelisD: This is valid DuckDB:

```
SELECT CAST(5 // 3 AS DECIMAL(5, 3))
```

But without building the decimal type, we get rid of the cast. Will switch to building

georgesittas: Right

tests/dialects/test_postgres.py
```diff
@@ -724,6 +724,28 @@ def test_postgres(self):
         self.validate_identity("cast(a as FLOAT8)", "CAST(a AS DOUBLE PRECISION)")
         self.validate_identity("cast(a as FLOAT4)", "CAST(a AS REAL)")
 
+        self.validate_all(
+            "1 / DIV(4, 2)",
+            read={
+                "postgres": "1 / DIV(4, 2)",
+            },
+            write={
+                "sqlite": "1 / CAST(CAST(CAST(4 AS REAL) / 2 AS INTEGER) AS REAL)",
```
georgesittas, line 733: The `CAST(4 AS REAL)` here looks a bit weird, I don't think it should've been added. The root cause is that the `IntDiv` generator [doesn't properly populate](https://github.com/tobymao/sqlglot/blob/main/sqlglot/generator.py#L3136-L3142) the `typed` and `safe` args, disregarding the dialect's semantics. It seems like that method was added [a long time ago](https://github.com/tobymao/sqlglot/commit/193d4f2c0463c23aba220c7c5243a83d248078d9), so this is most likely like an oversight.

georgesittas: > I don't think it should've been added

Reasoning is that Postgres uses typed division, so since we're generating a `Div` for the `IntDiv`, the typed division semantics should be respected. See for example:

```python
>>> import sqlglot
>>> sqlglot.transpile("1/2", "postgres", "sqlite")
['1 / 2']
```

georgesittas: May be best to address this in a followup PR after all. I played around a bit and it seems like we'll need to think more about how to parse / generate `IntDiv`.

PR #3612, opened 2024-06-06
Fix(redshift): add support for Oracle style outer join markers

### LINKED ISSUE #3611: Redshift - add support for oracle style join markers ` (+)`
```python
from sqlglot import parse_one
sample_sql_stmt = "select a.foo, b.bar, a.baz from a, b where a.baz = b.baz (+)"
parse_one(sample_sql_stmt, read="redshift").sql(dialect="redshift")
```
expected:
`SELECT a.foo, b.bar, a.baz FROM a, b WHERE a.baz = b.baz (+)`

actual:
`SELECT a.foo, b.bar, a.baz FROM a, b WHERE a.baz = b.BAZ()`

**Official Documentation**
[Oracle-Style outer joins in the WHERE clause](https://docs.aws.amazon.com/redshift/latest/dg/r_WHERE_oracle_outer.html)
>For Oracle compatibility, Amazon Redshift supports the Oracle outer-join operator (+) in WHERE clause join conditions

sandband:
Added support for oracle style join markers `(+)`

Fixes #3611

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -115,6 +115,12 @@ class Redshift(Postgres):
             self._retreat(index)
             return None
 
+        def _parse_column(self) -> t.Optional[exp.Expression]:
+            column = super()._parse_column()
+            if column:
+                column.set("join_mark", self._match(TokenType.JOIN_MARKER))
+            return column
+
     class Tokenizer(Postgres.Tokenizer):
         BIT_STRINGS = []
         HEX_STRINGS = []
@@ -128,6 +134,7 @@ class Redshift(Postgres):
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
             "MINUS": TokenType.EXCEPT,
+            "(+)": TokenType.JOIN_MARKER,
         }
         KEYWORDS.pop("VALUES")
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -128,6 +134,7 @@ class Tokenizer(Postgres.Tokenizer):
             "UNLOAD": TokenType.COMMAND,
             "VARBYTE": TokenType.VARBINARY,
             "MINUS": TokenType.EXCEPT,
+            "(+)": TokenType.JOIN_MARKER,
```
georgesittas, line 137: Let's move this to the top

sqlglot/dialects/redshift.py
```diff
@@ -115,6 +115,12 @@ def _parse_approximate_count(self) -> t.Optional[exp.ApproxDistinct]:
             self._retreat(index)
             return None
 
+        def _parse_column(self) -> t.Optional[exp.Expression]:
+            column = super()._parse_column()
+            if column:
+                column.set("join_mark", self._match(TokenType.JOIN_MARKER))
+            return column
```
georgesittas, line 122: Let's not duplicate this logic here. Instead, you can move this to the base `Parser`'s `_parse_column` method and control it using a dialect flag.

Notice that we already have a flag in the `Generator` class (see [here](https://github.com/tobymao/sqlglot/blob/main/sqlglot/generator.py#L228-L229)), so introducing another one in the `Parser` is going to be redundant. Instead, let's move that generator constant in the `Dialect` class and have Redshift and Oracle set it to `True`.

Then, you'll access it as `self.dialect.COLUMN_JOIN_MARKS_SUPPORTED` in both the parser and the generator.

georgesittas (commented):
Thanks for the PR, interesting that Redshift also supports this syntax. TIL!

Couple of suggestions in order to clean this up a bit, LGTM otherwise.

georgesittas:
I'll take this one to the finish line, again thanks for the PR.

sandband:
Thank you!

PR #3614, opened 2024-06-07
fix(optimizer): don't use datetrunc type for right side

barakalon:
The code will current convert do stuff like this:
```sql
DATE_TRUNC('day', CAST(x AS DATE)) <= CAST('2021-01-01 00:00:00' AS TIMESTAMP)
-- becomes
CAST(x AS DATE) < CAST('2021-01-02 00:00:00' AS DATE)
```
Which breaks.

This date trunc stuff has gotten really complicated. @georgesittas please double check me.


georgesittas (approved):
This looks like an oversight on my end, thanks for fixing 👍

Change LGTM, I don't see why we'd cast the date literal into the `TRUNC` arg's type. Can you double check if there are other places where we do this?

barakalon:
Ah, there was indeed another spot. Fixed.

georgesittas (approved):

PR #3623, opened 2024-06-10
Feat(mysql): support STRAIGHT_JOIN

### LINKED ISSUE #3622: MySQL read dialect doesn't support STRAIGHT_JOIN keyword
`STRAIGHT_JOIN` is a [valid keyword](https://dev.mysql.com/doc/refman/8.4/en/join.html) in MySQL forcing the optimizer to honor table order in the query for joins. 

```python
from sqlglot import parse_one, exp, errors, expressions

sqltext="select e.* from employee e straight_join passenger p on e.telephoneno = p.lastname"

try:
    sg_table_list = parse_one(sqltext,"mysql").find_all(exp.Table)
except errors.ParseError:
    print("The query could not be parsed")
else:
    print("Success")

```

Replacing `straight_join` for `join` on the snippet above will prevent the exception. Code above was tested with version `25.0.3`

georgesittas:
Fixes #3622

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -213,6 +213,7 @@ class MySQL(Dialect):
             "MEMBER OF": TokenType.MEMBER_OF,
             "SEPARATOR": TokenType.SEPARATOR,
             "START": TokenType.BEGIN,
+            "STRAIGHT_JOIN": TokenType.STRAIGHT_JOIN,
             "SIGNED": TokenType.BIGINT,
             "SIGNED INTEGER": TokenType.BIGINT,
             "UNLOCK TABLES": TokenType.COMMAND,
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -463,6 +469,13 @@ class Parser(parser.Parser):
         VALUES_FOLLOWED_BY_PAREN = False
         SUPPORTS_PARTITION_SELECTION = True
 
+        def _parse_join(
+            self, skip_join_token: bool = False, parse_bracket: bool = False
+        ) -> t.Optional[exp.Join]:
+            if self._match(TokenType.STRAIGHT_JOIN, advance=False):
```
tobymao, line 475: why was this section needed? also can we get this to work with the sqlglot dialect?

georgesittas: > why was this section needed?

Because we're [expecting](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L3102-L3108) the `JOIN` keyword by default, and the `STRAIGHT_JOIN` syntax doesn't have it.

> also can we get this to work with the sqlglot dialect?

Yeah, I'll refactor 👍

PR #3639, opened 2024-06-12
feat(oracle): Support for WITH READ ONLY / CHECK OPTION

### LINKED ISSUE #3635: Oracle materialized views WITH READ ONLY has Invalid expression / Unexpected token / required keyword format
I am trying to parse some oracle views in sqlglot, some materialized view in oracle has a keyword WITH READ ONLY at the end of the view to indicate that view can't be updated, with use it a lot for many views. The errors are vary sometimes I see `required keyword format` and for other views I encounter  `Invalid expression / Unexpected token`

https://docs.oracle.com/cd/B13789_01/server.101/b10759/statements_8004.htm

**WITH READ ONLY
Specify WITH READ ONLY to indicate that the table or view cannot be updated.**

Example: 

```
import sqlglot

res = sqlglot.transpile(
    ''' 
      select * from dual
      with read only
    ''''', read="oracle", write="oracle")[0]

print(res)
```

Traceback:
```

  File "/Users/deeb/PycharmProjects/datahub_app/parse.py", line 3, in <module>
    res = sqlglot.transpile(
          ^^^^^^^^^^^^^^^^^^
  File "/Users/deeb/anaconda3/lib/python3.11/site-packages/sqlglot/__init__.py", line 177, in transpile
    for expression in parse(sql, read, error_level=error_level)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/deeb/anaconda3/lib/python3.11/site-packages/sqlglot/__init__.py", line 102, in parse
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/deeb/anaconda3/lib/python3.11/site-packages/sqlglot/dialects/dialect.py", line 512, in parse
  File "/Users/deeb/anaconda3/lib/python3.11/site-packages/sqlglot/parser.py", line 1246, in parse
    return self._parse(
           ^^^^^^^^^^^^
  File "/Users/deeb/anaconda3/lib/python3.11/site-packages/sqlglot/parser.py", line 1318, in _parse
  File "/Users/deeb/anaconda3/lib/python3.11/site-packages/sqlglot/parser.py", line 1359, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 3, Col: 4.
   
select * from dual
with read only

```

VaggelisD:
Fixes #3635

This PR aims to support Oracle's `<subquery_restriction_clause>` by defining new expressions for `READ ONLY` and `CHECK OPTIONS` and parsing them as WITH properties through `QUERY_MODIFIER_PARSERS`, as this clause always follows an Oracle `<subquery>` (i.e query block, subquery or wrapped subquery).

- [Oracle <subquery_restriction_clause>](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6__I2126149)

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -457,3 +457,10 @@ class Oracle(Dialect):
             if len(expression.args.get("actions", [])) > 1:
                 return f"ADD ({actions})"
             return f"ADD {actions}"
+
+        def queryoption_sql(self, expression: exp.QueryOption) -> str:
+            option = self.sql(expression, "this")
+            value = self.sql(expression, "expression")
+            value = f" CONSTRAINT {value}" if value else ""
+
+            return f"{option}{value}"
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -36,8 +36,8 @@ def eliminate_join_marks(ast: exp.Expression) -> exp.Expression:
     from sqlglot.optimizer.scope import traverse_scope
 
     """Remove join marks from an expression
-    
-    SELECT * FROM a, b WHERE a.id = b.id(+)   
+
+    SELECT * FROM a, b WHERE a.id = b.id(+)
     becomes:
     SELECT * FROM a LEFT JOIN b ON a.id = b.id
 
@@ -293,7 +293,7 @@ class Oracle(Dialect):
         QUERY_MODIFIER_PARSERS = {
             **parser.Parser.QUERY_MODIFIER_PARSERS,
             TokenType.ORDER_SIBLINGS_BY: lambda self: ("order", self._parse_order()),
-            TokenType.WITH: lambda self: ("restrictions", self._parse_property()),
+            TokenType.WITH: lambda self: ("options", self._parse_query_restrictions()),
         }
 
         TYPE_LITERAL_PARSERS = {
@@ -348,6 +348,26 @@ class Oracle(Dialect):
                 return exp.Hint(expressions=[self._find_sql(start, end)])
 
             return None
+
+        def _parse_query_restrictions(self) -> t.Optional[exp.Expression]:
+            self._match(TokenType.WITH)
+
+            matched = self._match_text_seq("READ", "ONLY") or self._match_text_seq(
+                "CHECK", "OPTION"
+            )
+
+            if not matched:
+                self._retreat(self._index - 1)
+                return None
+
+            # Oracle <subquery_restriction_clause>
+            kind = "WITH READ ONLY" if self._prev.text.upper() == "ONLY" else "WITH CHECK OPTION"
+
+            return self.expression(
+                exp.QueryOption,
+                this=exp.var(kind),
+                expression=self._parse_field() if self._match(TokenType.CONSTRAINT) else None,
+            )
 
     class Generator(generator.Generator):
         LOCKING_READS_SUPPORTED = True
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -293,6 +293,7 @@ class Oracle(Dialect):
         QUERY_MODIFIER_PARSERS = {
             **parser.Parser.QUERY_MODIFIER_PARSERS,
             TokenType.ORDER_SIBLINGS_BY: lambda self: ("order", self._parse_order()),
+            TokenType.WITH: lambda self: ("restrictions", self._parse_property()),
         }
 
         TYPE_LITERAL_PARSERS = {
@@ -411,6 +412,8 @@ class Oracle(Dialect):
         PROPERTIES_LOCATION = {
             **generator.Generator.PROPERTIES_LOCATION,
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
+            exp.WithCheckOptionProperty: exp.Properties.Location.UNSUPPORTED,
+            exp.WithReadOnlyProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
         def currenttimestamp_sql(self, expression: exp.CurrentTimestamp) -> str:
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -13,6 +13,7 @@ from sqlglot.dialects.dialect import (
     trim_sql,
 )
 from sqlglot.helper import seq_get
+from sqlglot.parser import OPTIONS_TYPE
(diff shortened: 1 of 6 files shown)


sqlglot/parser.py
```diff
@@ -2005,6 +2005,20 @@ def _parse_with_property(self) -> t.Optional[exp.Expression] | t.List[exp.Expres
...
+            klass = (
+                exp.WithReadOnlyProperty
+                if self._prev.text.upper() == "ONLY"
+                else exp.WithCheckOptionProperty
+            )
+            this = (
+                self._parse_constraint()
+                if self._match(TokenType.CONSTRAINT, advance=False)
+                else None
+            )
```
georgesittas, line 2019: This needs to be tested

sqlglot/expressions.py
```diff
@@ -2844,6 +2844,14 @@ class WithSystemVersioningProperty(Property):
     }
 
 
+class WithReadOnlyProperty(Property):
+    arg_types = {"this": False}
+
+
+class WithCheckOptionProperty(Property):
+    arg_types = {"this": False}
```
georgesittas, line 2852: An alternative could to be to get rid of these two, reuse `QueryOption` and handle the cases as follows:

- Create a `Var` using `_parse_var_from_options` for `WITH READ ONLY` and `WITH CHECK OPTION` (so you'd want to set `raise_unmatched=False`).
- Handle `constraint` like you do now.

Only `QueryOption.this` will be needed to represent this. Generation can go in Oracle. I'm suggesting this because I'm not sure I like the property approach.

VaggelisD: Hmm, how about storing the kind (read only / check option) as a Var in `QueryOption.this` and the constraint in `QueryOption.expression`?

georgesittas: That works too, I just didn't suggest it because the difference is minimal and it'll require making `this` optional.

sqlglot/parser.py
VaggelisD: This actually seems to be in the form of `CONSTRAINT <identifier>`, per the docs:

```
CONSTRAINT constraint

Specify the name of the CHECK OPTION constraint. If you omit this identifier, then Oracle automatically assigns the constraint a name of the form SYS_Cn, where n is an integer that makes the constraint name unique within the database.
```

So I'll also change the parsing here

sqlglot/dialects/oracle.py
```diff
@@ -348,6 +349,26 @@ def _parse_hint(self) -> t.Optional[exp.Hint]:
...
+            matched = self._match_text_seq("READ", "ONLY") or self._match_text_seq(
+                "CHECK", "OPTION"
+            )
+
+            if not matched:
+                self._retreat(self._index - 1)
+                return None
+
+            # Oracle <subquery_restriction_clause>
+            kind = "WITH READ ONLY" if self._prev.text.upper() == "ONLY" else "WITH CHECK OPTION"
```
georgesittas, line 364: Let's instead do:

```python
QUERY_RESTRICTIONS: OPTIONS_TYPE = {
    "WITH": (
        ("READ", "ONLY"),
        ("CHECK", "OPTION"),
    ),
}
```

And then:

```python
kind = self._parse_var_from_options(self.QUERY_RESTRICTIONS, raise_unmatched=False)
if not kind:
    self._retreat(...)
    return None

return ...
```

sqlglot/dialects/oracle.py
```diff
@@ -411,6 +432,8 @@ class Generator(generator.Generator):
         PROPERTIES_LOCATION = {
             **generator.Generator.PROPERTIES_LOCATION,
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
+            exp.WithCheckOptionProperty: exp.Properties.Location.UNSUPPORTED,
+            exp.WithReadOnlyProperty: exp.Properties.Location.UNSUPPORTED,
```
georgesittas, line 436:
```suggestion
```

sqlglot/dialects/oracle.py
```diff
@@ -348,6 +349,26 @@ def _parse_hint(self) -> t.Optional[exp.Hint]:
...
+                self._retreat(self._index - 1)
+                return None
+
+            # Oracle <subquery_restriction_clause>
+            kind = "WITH READ ONLY" if self._prev.text.upper() == "ONLY" else "WITH CHECK OPTION"
+
+            return self.expression(
+                exp.QueryOption,
+                this=exp.var(kind),
+                expression=self._parse_field() if self._match(TokenType.CONSTRAINT) else None,
```
georgesittas, line 369:
```suggestion
                expression=self._match(TokenType.CONSTRAINT) and self._parse_field(),
```

sqlglot/expressions.py
georgesittas, line 2852: Revert changes to this file
```suggestion
```

sqlglot/generator.py
```diff
@@ -2234,9 +2234,7 @@ def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
         elif self.LIMIT_FETCH == "FETCH" and isinstance(limit, exp.Limit):
             limit = exp.Fetch(direction="FIRST", count=exp.maybe_copy(limit.expression))
 
-        options = self.expressions(expression, key="options")
-        if options:
-            options = f" OPTION{self.wrap(options)}"
+        options = self.options_modifier(expression)
```
georgesittas, line 2237: inline this in the `return` statement

sqlglot/generator.py
```diff
@@ -3959,3 +3965,13 @@ def gapfill_sql(self, expression: exp.GapFill) -> str:
...
+
+    def withreadonlyproperty_sql(self, expression: exp.WithReadOnlyProperty) -> str:
+        this = self.sql(expression, "this")
+        this = f" {this}" if this else ""
+        return f"WITH READ ONLY{this}"
+
+    def withcheckoptionproperty_sql(self, expression: exp.WithCheckOptionProperty) -> str:
+        this = self.sql(expression, "this")
+        this = f" {this}" if this else ""
+        return f"WITH CHECK OPTION{this}"
```
georgesittas, line 3977:
```suggestion
```

georgesittas (approved):

PR #3646, opened 2024-06-13
Fix(mysql): delete redundant keywords

Toms1999:
I found some repetitions in the RESERVED_KEYWORDS for mysql dialect, and I checked the relevant pr to make sure it was indeed repetitions and not something else


georgesittas (approved):
Thanks 👍

PR #3652, opened 2024-06-13
fix(redshift)!: Normalize time units in their full singular form

### LINKED ISSUE #3651: issue in transpile
Code: 
```
import sqlglot
from sqlglot import transpile
print(sqlglot.__version__)

sql = """
SELECT DATEDIFF(days, '2021-01-01', '2021-02-01')
"""
print(transpile(sql, read='redshift',write='bigquery')[0])
```

Output: 
```
25.1.0
SELECT DATE_DIFF(CAST('2021-02-01' AS DATETIME), CAST('2021-01-01' AS DATETIME), DAYS)
```

The date interval in the output should be `DAY`, not `DAYS`. 

**Official Documentation**
https://count.co/sql-resources/bigquery-standard-sql/date_diff

VaggelisD:
Fixes #3651

Normalize Redshift's units from plural/abbreviation to their full singular form during parse time. This is already happening at Snowflake, so the existing infrastructure is moved out to `dialect.py` as most of these mappings are similar between dialects.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -53,13 +53,14 @@ DATE_PART_MAPPING = {
     "DD": "DAY",
     "DAYS": "DAY",
     "DAYOFMONTH": "DAY",
+    "DAY OF WEEK": "DAYOFWEEK",
     "WEEKDAY": "DAYOFWEEK",
     "DOW": "DAYOFWEEK",
     "DW": "DAYOFWEEK",
     "WEEKDAY_ISO": "DAYOFWEEKISO",
     "DOW_ISO": "DAYOFWEEKISO",
     "DW_ISO": "DAYOFWEEKISO",
-    "YEARDAY": "DAYOFYEAR",
+    "DAY OF YEAR": "DAYOFYEAR",
     "DOY": "DAYOFYEAR",
     "DY": "DAYOFYEAR",
     "W": "WEEK",
@@ -90,9 +91,20 @@ DATE_PART_MAPPING = {
     "SECS": "SECOND",
     "MS": "MILLISECOND",
     "MSEC": "MILLISECOND",
+    "MSECS": "MILLISECOND",
+    "MSECOND": "MILLISECOND",
+    "MSECONDS": "MILLISECOND",
+    "MILLISEC": "MILLISECOND",
+    "MILLISECS": "MILLISECOND",
+    "MILLISECON": "MILLISECOND",
     "MILLISECONDS": "MILLISECOND",
     "US": "MICROSECOND",
     "USEC": "MICROSECOND",
+    "USECS": "MICROSECOND",
+    "MICROSEC": "MICROSECOND",
+    "MICROSECS": "MICROSECOND",
+    "USECOND": "MICROSECOND",
+    "USECONDS": "MICROSECOND",
     "MICROSECONDS": "MICROSECOND",
     "NS": "NANOSECOND",
     "NSEC": "NANOSECOND",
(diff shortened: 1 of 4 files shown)


tests/dialects/test_redshift.py
```diff
@@ -260,7 +260,7 @@ def test_redshift(self):
             },
         )
         self.validate_all(
-            "DATEDIFF('day', a, b)",
+            "DATEDIFF(days, a, b)",
```
georgesittas, line 263: Let's make a separate test for this

sqlglot/dialects/dialect.py
```diff
@@ -37,6 +37,79 @@
 }
 
 
+DATE_PART_MAPPING = {
```
georgesittas, line 40: @VaggelisD is this the same mapping as in Snowflake? Did you double check if all units are included? https://docs.aws.amazon.com/redshift/latest/dg/r_Dateparts_for_datetime_functions.html

VaggelisD: I added the missing ones. A change in plans as I noticed that Snowflake mapped `"epoch" -> "epoch_seconds"`, but in Redshift it's `"epoch" -> "epoch"`. I'm guessing there might be differences across dialects, so generalizing among them became a bit harder..

VaggelisD: I tried that but the builder functions that `map_date_part` is applied are outside of dialect. The same happens with the `TimeUnit` refactor we had in mind, it doesn't have access to dialect at `expressions.py`

georgesittas: The builder functions have access to the dialect, you just need to specify it - it's an optional arg. So, `exp.SomeExpr: lambda args, dialect: ...`. Does that help?

VaggelisD: Hmm, will check it out. I think it will make things a bit more troublesome than the current version though, with the builder `dialect` boilerplate

georgesittas: If it turns out too complicated we can delay implementing a more robust / generic solution until a larger scale date part refactor is done.

VaggelisD:
I thought some parts again and have concluded to the following regarding this PR:
- I moved `DATE_PART_MAPPING` to `Dialect`, thus any dialect that wants to change it's mappings can do so and pass its self in `map_date_part()`. 
- There was a dispute between Redshift `EPOCH` & Snowflake `EPOCH_SECOND` but Snowflake also supports `EPOCH, EPOCH_SECONDS` as abbreviations for `EPOCH_SECOND`; Thus, I followed the convention that Snowflake will turn `EPOCH_SECOND -> EPOCH`, and with this change we won't need to override the mapping in either dialect


Regarding the general refactor:
- We can probably swap `exp.TimeUnit::UNABBREVIATED_UNIT_NAME` with `Dialect::DATE_PART_MAPPING`, which should normalize most (if not all) dialects. 
- If any dialect has different mappings, we can go the manual `map_date_part(unit, dialect)` route. This is a necessary evil because `exp.TimeUnit` won't have access to the dialect specific mapping, but we can rethink this later on.


Any thoughts?

georgesittas (approved):
LGTM

> We can probably swap exp.TimeUnit::UNABBREVIATED_UNIT_NAME with Dialect::DATE_PART_MAPPING, which should normalize most (if not all) dialects and remove the need for map_date_part() everywhere.

Sounds good. If there are no inconsistencies between dialects then that would make things simpler. The tricky part begins when you have to override certain date parts because there are discrepancies.

> If any dialect has different mappings, we can go the manual map_date_part(unit, dialect) route. This is a necessary evil because exp.TimeUnit won't have access to the dialect specific mapping, but we can rethink this later on.

Right, so we'll be able to have an escape hatch. Sounds good to me 👍

georgesittas (approved):

PR #3669, opened 2024-06-18
fix lag/lead function must be three args in doris

Toms1999:
MySQL [(none)]> SELECT LAG(1) OVER ();
ERROR 1105 (HY000): errCode = 2, detailMessage = Lag/offset must have three parameters

PR #3672, opened 2024-06-18
fix lag/lead function must be three args in doris 

Toms1999:
MySQL [(none)]> SELECT LAG(1) OVER ();
ERROR 1105 (HY000): errCode = 2, detailMessage = Lag/offset must have three parameters


Toms1999:
By make style @georgesittas

georgesittas:
Did you see https://github.com/tobymao/sqlglot/pull/3669#issuecomment-2175779755?

Toms1999:
make style
pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
mypy.....................................................................Passed

I did this. I think you said that, right? @georgesittas

PR #3675, opened 2024-06-19
fix postgres '#>>' to doris

Toms1999:
[json_extract](https://doris.apache.org/docs/dev/sql-manual/sql-functions/json-functions/json-extract/)


sqlglot/dialects/doris.py
```diff
@@ -20,6 +20,12 @@ def _lag_lead_sql(self, expression: exp.Lag | exp.Lead) -> str:
     )
 
 
+def _jsonb_extract_scalar_sql(expression: exp.JSONBExtractScalar) -> str:
+    path = expression.this
+    path_elements = path.strip("{}").split(",")
+    return f"'$.{'.'.join(path_elements)}'"
```
georgesittas, line 26: What is this? It looks hacky

Toms1999: ago , it is 
`'{A,B,C}'`
i want to it be 
`'$.A.B.C'` , so , I used the above method to pass strings

Toms1999:
```
from sqlglot import  transpile


sql = """SELECT '{"A": {"B": {"C": "VALUE1"}}}'::JSON#>>'{A,B,C}'"""

result = transpile(sql, read='postgres', write='doris')[0]
print(result)
```

georgesittas:
This approach doesn't look robust to me. For example, it assumes that the keys in the array won't contain any commas:

```
georgesittas=# SELECT '{"A": {",": {"C": "VALUE"}}}'::JSON #>> '{A,\,,C}';
 ?column?
----------
 VALUE
(1 row)
```

The above query is transpiled incorrectly using this branch:

```python
>>> import sqlglot
>>> print(sqlglot.transpile("""SELECT '{"A": {",": {"C": "VALUE"}}}'::JSON #>> '{A,\\,,C}'""", "postgres", "doris")[0])
SELECT JSONB_EXTRACT(CAST('{"A": {",": {"C": "VALUE"}}}' AS JSON), '$.A.\..C')
```

Another example where the transpilation doesn't seem correct, where the query is taken from Postgres' [docs](https://www.postgresql.org/docs/9.3/functions-json.html):

```
>>> import sqlglot
>>> print(sqlglot.transpile("""SELECT '{"a":[1,2,3],"b":[4,5,6]}'::json#>>'{a,2}'""", "postgres", "doris")[0])
SELECT JSONB_EXTRACT(CAST('{"a":[1,2,3],"b":[4,5,6]}' AS JSON), '$.a.2')
```

I believe that `.2` is invalid JSON Path syntax in the resulting query, but happy to be proven wrong.

I'll close this, but you can easily add this transformation in your application. Also, appreciate the PRs, but they won't be accepted if there's no effort put into making the CI tests pass, documenting the solutions and testing them sufficiently.

Toms1999:
Now that I have completed the above question, can I ask pr again? @georgesittas

PR #3678, opened 2024-06-19
feat(databricks)!: Preserve JSON/VARIANT path with operators

### LINKED ISSUE #3673: Sqlglot does not preserve new Databricks's Variant Syntax
The newest Databricks Runtime (15.3) does support a number of fancy syntax for their new Variant Type.
Docs: https://docs.databricks.com/en/semi-structured/variant.html

Those do not get preserved

```python
import sqlglot as sg
print(sg.parse_one("SELECT raw:store.bicycle FROM store_data", dialect="databricks").sql("databricks"))
-- prints "SELECT GET_JSON_OBJECT(raw, '$.store.bicycle') FROM store_data"

print(sg.parse_one("SELECT raw:`zip code`, raw:`fb:testid` FROM store_data", dialect="databricks").sql("databricks"))
-- prints "SELECT GET_JSON_OBJECT(raw, '$[\\'`zip\\'][\\'code`\\']'), GET_JSON_OBJECT(raw, '$[\\'`fb\\'][\\'testid`\\']') FROM store_data"
```

VaggelisD:
Fixes #3673

This PR aims to preserve the JSON/Variant access in Databricks using operators instead of JSON functions + JSONPath by:
- Adding `IDENTIFIERS` in `jsonpath.py` to be able to parse keys with special chars (see [example](https://docs.databricks.com/en/semi-structured/variant.html#extract-a-top-level-variant-field))
- Manually parsing `GET_JSON_OBJECT()` -> `exp.JSONExtract` 
- Changing the generation of `exp.JSONExtract` & `exp.JSONExtractScalar` (only in DB) to the operator-based syntax

Docs
--------
[DB querying VARIANT](https://docs.databricks.com/en/semi-structured/variant.html#query-variant-data) | [DB path expression](https://docs.databricks.com/en/sql/language-manual/sql-ref-json-path-expression.html) | [DB JSON functions](https://docs.databricks.com/en/sql/language-manual/sql-ref-functions-builtin.html#json-functions)

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp, transforms
 from sqlglot.dialects.dialect import (
     date_delta_sql,
@@ -10,10 +12,25 @@ from sqlglot.dialects.spark import Spark
 from sqlglot.tokens import TokenType
 
 
+def _build_json_extract(args: t.List) -> exp.JSONExtract:
+    # Transform GET_JSON_OBJECT(expr, '$.<path>') -> expr:<path>
+    this = args[0]
+    path = args[1].name.lstrip("$.")
+    return exp.JSONExtract(this=this, expression=path)
+
+
 def _timestamp_diff(
     self: Databricks.Generator, expression: exp.DatetimeDiff | exp.TimestampDiff
 ) -> str:
     return self.func("TIMESTAMPDIFF", expression.unit, expression.expression, expression.this)
+
+
+def _jsonextract_sql(
+    self: Databricks.Generator, expression: exp.JSONExtract | exp.JSONExtractScalar
+) -> str:
+    this = self.sql(expression, "this")
+    expr = self.sql(expression, "expression")
+    return f"{this}:{expr}"
 
 
 class Databricks(Spark):
@@ -31,6 +48,7 @@ class Databricks(Spark):
             "DATE_ADD": build_date_delta(exp.DateAdd),
             "DATEDIFF": build_date_delta(exp.DateDiff),
             "TIMESTAMPDIFF": build_date_delta(exp.TimestampDiff),
+            "GET_JSON_OBJECT": _build_json_extract,
         }
 
         FACTOR = {
@@ -42,6 +60,7 @@ class Databricks(Spark):
         TABLESAMPLE_SEED_KEYWORD = "REPEATABLE"
         COPY_PARAMS_ARE_WRAPPED = False
         COPY_PARAMS_EQ_REQUIRED = True
+        JSON_PATH_SINGLE_QUOTE_ESCAPE = False
 
         TRANSFORMS = {
             **Spark.Generator.TRANSFORMS,
@@ -65,6 +84,9 @@ class Databricks(Spark):
                     transforms.unnest_to_explode,
                 ]
             ),
+            exp.JSONExtract: _jsonextract_sql,
+            exp.JSONExtractScalar: _jsonextract_sql,
+            exp.JSONPathRoot: lambda *_: "",
             exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
@@ -88,3 +110,6 @@ class Databricks(Spark):
         ) -> str:
             expression.set("this", True)  # trigger ALWAYS in super class
             return super().generatedasidentitycolumnconstraint_sql(expression)
+
+        def jsonpath_sql(self, expression: exp.JSONPath) -> str:
+            return self.expressions(expression, sep="", flat=True).lstrip(".")
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -61,6 +61,7 @@ class Databricks(Spark):
         COPY_PARAMS_ARE_WRAPPED = False
         COPY_PARAMS_EQ_REQUIRED = True
         JSON_PATH_SINGLE_QUOTE_ESCAPE = False
+        QUOTE_JSON_PATH = False
 
         TRANSFORMS = {
             **Spark.Generator.TRANSFORMS,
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/databricks.py
```diff
@@ -65,6 +84,9 @@ class Generator(Spark.Generator):
                     transforms.unnest_to_explode,
                 ]
             ),
+            exp.JSONExtract: _jsonextract_sql,
+            exp.JSONExtractScalar: _jsonextract_sql,
```
georgesittas, line 88: We don't need to define `_jsonextract_sql`, consider this instead:

```python
            exp.JSONExtract: lambda self, e: binary(e, ":"),
            exp.JSONExtractScalar: lambda self, e: binary(e, ":"),
```

sqlglot/jsonpath.py
```diff
@@ -34,6 +34,7 @@ class JSONPathTokenizer(Tokenizer):
 
     IDENTIFIER_ESCAPES = ["\\"]
     STRING_ESCAPES = ["\\"]
+    IDENTIFIERS = ["`", '"']
```
georgesittas, line 37: This doesn't look right, we shouldn't have to change the JSON Path dialect, it's a separate concept.

sqlglot/dialects/databricks.py
```diff
@@ -88,3 +110,6 @@ def generatedasidentitycolumnconstraint_sql(
         ) -> str:
             expression.set("this", True)  # trigger ALWAYS in super class
             return super().generatedasidentitycolumnconstraint_sql(expression)
+
+        def jsonpath_sql(self, expression: exp.JSONPath) -> str:
+            return self.expressions(expression, sep="", flat=True).lstrip(".")
```
georgesittas, line 115: Why are we `lstrip`ing here?

sqlglot/jsonpath.py
VaggelisD: I agree, but how would we be able to decouple the `jsonpath.py` parsing from the Databricks specific syntax if the former can't parse the keys with special chars properly?

sqlglot/dialects/databricks.py
georgesittas: I see. Maybe we can control the quote generation in base instead of overriding?

georgesittas: I'm fine with either approach, just a thought.

VaggelisD: Sounds good, I can define a generator flag along the lines of `QUOTE_JSON_PATH` and set it off in Databricks

sqlglot/jsonpath.py
VaggelisD: It looks like `Parser._parse_colon_as_json_extract()` handles this fine, meaning that for the example:

```SQL
SELECT raw:`zip code`, raw:`fb:testid` FROM store_data
```

 it will parse the following pairs:
```
(this='raw', json_path=['`zip code`'])
(this='raw', json_path=['`fb:testid`'])
```

However, when it passes these pairs to the `exp.JsonExtract` construction the `json_path` breaks due to the `dialect.to_json_path()` call i.e because of the `jsonpath.py` tokenizer  afaict. In turn, the following incorrect query is produced:

```SQL
SELECT raw:["`zip"]["code`"], raw:["`fb"]["testid`"]
```

sqlglot/dialects/databricks.py
VaggelisD: Looks like `binary` introduces whitespace between the operands / path parts, if it's useful I can refactor it to be more modular in that regard otherwise will keep `_jsonextract_sql`

sqlglot/jsonpath.py
georgesittas: Is this something that we could fix by overriding `to_json_path` for `Databricks` then?

VaggelisD: Probably not, `dialect.to_json_path()` is mostly a wrapper around `jsonpath.py::parse()` which in turn contains all of the JSONPath parsing under it. Sadly, the API here is not granular in that regard. 

If the `JSONPathTokenizer::IDENTIFIERS` approach is fine for Databricks but questionable for other dialects, the closest solution I can see will be to create a `databricks.py::to_json_path()` impl which will forward `IDENTIFIERS` to `jsonpath.py::parse()` or `JSONPathTokenizer`.

georgesittas: I'm not sure I like either approach, need to think about this a bit. Let's sync in Slack.

VaggelisD:
As discussed in Slack, made `JSONPathTokenizer` a `Dialect` attribute similarly to `Tokenizer`, `Parser`, `Generator` in an attempt to make it parameterizable by each dialect

georgesittas (approved):

sqlglot/jsonpath.py
```diff
@@ -35,145 +35,144 @@ class JSONPathTokenizer(Tokenizer):
...
-                    indexes.append(literal)
-
-            if len(indexes) == 1:
-                if isinstance(literal, str):
-                    node: exp.JSONPathPart = exp.JSONPathKey(this=indexes[0])
-                elif isinstance(literal, exp.JSONPathPart) and isinstance(
-                    literal, (exp.JSONPathScript, exp.JSONPathFilter)
-                ):
-                    node = exp.JSONPathSelector(this=indexes[0])
+    def parse(self, path: str) -> exp.JSONPath:
```
georgesittas, line 38: FYI - didn't notice this was moved within the tokenizer class at first, discussed in slack about moving to top-level and adding a `dialect` kwarg to get the correct tokenizer for each dialect.

sqlglot/dialects/dialect.py
```diff
@@ -655,6 +659,12 @@ def tokenizer(self) -> Tokenizer:
             self._tokenizer = self.tokenizer_class(dialect=self)
         return self._tokenizer
 
+    @property
+    def jsonpath_tokenizer(self) -> JSONPathTokenizer:
+        if not hasattr(self, "_jsonpath_tokenizer"):
+            self._jsonpath_tokenizer = self.jsonpath_tokenizer_class(dialect=self)
+        return self.jsonpath_tokenizer_class(dialect=self)
```
georgesittas, line 666: This doesn't cache it right?
```suggestion
        return self._jsonpath_tokenizer
```

VaggelisD: Oh right, my intention was for `JSONPathTokenizer` to follow `Tokenizer` i.e this shouldn't create a new instance here. Will fix

georgesittas (approved):

PR #3682, opened 2024-06-20
Fix(parser): handle another edge case in struct field type parser

### LINKED ISSUE #3680: sqlglot fails to parse BigQuery type cast

```python
import sqlglot
sqlglot.parse_one(sql="select sources::STRUCT<list ARRAY<STRUCT<element STRUCT<property STRING, dataset STRING, record_id STRING, confidence FLOAT64>>>> from bar", dialect="bigquery")
```

Produces

```
sqlglot.errors.ParseError: Required keyword: 'expression' missing for <class 'sqlglot.expressions.BitwiseRightShift'>. Line 1, Col: 134
```

georgesittas:
fixes #3680

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4493,7 +4493,7 @@ class Parser(metaclass=_Parser):
             and self._curr.token_type in self.TYPE_TOKENS
             and self._next.token_type in self.TYPE_TOKENS
         ):
-            # Care of special cases like `STRUCT<list ARRAY<...>>` where the identifier is also a
+            # Takes care of special cases like `STRUCT<list ARRAY<...>>` where the identifier is also a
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -4486,10 +4486,22 @@ def _parse_types(
...
-            or self._parse_id_var()
-        )
+
+        if (
+            self._curr
+            and self._next
+            and self._curr.token_type in self.TYPE_TOKENS
+            and self._next.token_type in self.TYPE_TOKENS
+        ):
+            # Care of special cases like `STRUCT<list ARRAY<...>>` where the identifier is also a
```
georgesittas, line 4496:
```suggestion
            # Takes care of special cases like `STRUCT<list ARRAY<...>>` where the identifier is also a
```

PR #3685, opened 2024-06-21
fix json#>> to doris

Toms1999:


Toms1999:
[json_extract](https://doris.apache.org/docs/dev/sql-manual/sql-functions/json-functions/json-extract/)

georgesittas:
@Toms1999 this approach looks complicated and non-idiomatic, given how other similar translations (e.g. JSON paths) are handled. Maybe we need to actually parse the `{...}` Postgres path syntax to do this cleanly, similar to how we handle JSON paths. However, that's a bigger task and it's rather low priority for us, so I'm going to close this PR for now - as I [said](https://github.com/tobymao/sqlglot/pull/3675#issuecomment-2178501521) in your other PR, if you need this it should be relatively easy to add it in your application by updating Doris' `TRANSFORMS` in your code.

PR #3696, opened 2024-06-24
Feat: transpile TRY_PARSE_JSON Snowflake -> DuckDB

### LINKED ISSUE #3690: Support for try_parse_json

I'm using sqlglot through sqlmesh and I have to use a macro for getting try_parse_json to work in duckdb.

I would like try_parse_json to be supported in sqlglot.

I tried using check_json as well but it's not supported either.

I ended up writing a conditional macro in sqlmesh to change the code depending on the gateway/dialect.

 @IF(
    @gateway = /* duckdb doesn't have try_parse_json and it's not in sqlglot */ 'local',
    CASE WHEN JSON_VALID(query_tag) THEN PARSE_JSON(query_tag) ELSE NULL END,
    TRY_PARSE_JSON(query_tag)
  ) AS query_tag,

**Additional context**

image

georgesittas:
Fixes #3690

References:
- https://docs.snowflake.com/en/sql-reference/functions/try_parse_json

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5506,6 +5506,11 @@ class ParseJSON(Func):
     is_var_len_args = True
 
 
+class TryParseJSON(Func):
+    # https://docs.snowflake.com/en/sql-reference/functions/try_parse_json
+    _sql_names = ["TRY_PARSE_JSON"]
+
+
 class Least(Func):
     arg_types = {"this": True, "expressions": False}
     is_var_len_args = True
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5501,14 +5501,9 @@ class JSONArrayContains(Binary, Predicate, Func):
 
 class ParseJSON(Func):
     # BigQuery, Snowflake have PARSE_JSON, Presto has JSON_PARSE
+    # Snowflake has TRY_PARSE_JSON
     _sql_names = ["PARSE_JSON", "JSON_PARSE"]
-    arg_types = {"this": True, "expressions": False}
-    is_var_len_args = True
-
-
-class TryParseJSON(Func):
-    # https://docs.snowflake.com/en/sql-reference/functions/try_parse_json
-    _sql_names = ["TRY_PARSE_JSON"]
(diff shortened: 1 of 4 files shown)


tobymao (approved):

sqlglot/expressions.py
```diff
@@ -5506,6 +5506,11 @@ class ParseJSON(Func):
     is_var_len_args = True
 
 
+class TryParseJSON(Func):
+    # https://docs.snowflake.com/en/sql-reference/functions/try_parse_json
+    _sql_names = ["TRY_PARSE_JSON"]
+
```
VaggelisD, line 5512: Iirc in Snowflake we encode the `TRY` variant of a func through a `safe` arg in the expression; For instance, in this case I see that there's `PARSE_JSON` and `TRY_PARSE_JSON`, maybe we could implement `exp.ParseJson(this, safe)`?

VaggelisD (commented):
Iirc in Snowflake we encode the `TRY` variant of a func through a `safe` arg in the expression; For instance, in this case I see that there's `PARSE_JSON` and `TRY_PARSE_JSON`, maybe we could implement `exp.ParseJson(this, safe)`?

sqlglot/expressions.py
georgesittas: Sure

sqlglot/expressions.py
```diff
@@ -5501,9 +5501,9 @@ class JSONArrayContains(Binary, Predicate, Func):
 
 class ParseJSON(Func):
     # BigQuery, Snowflake have PARSE_JSON, Presto has JSON_PARSE
+    # Snowflake also has TRY_PARSE_JSON, which is represented using `safe`
     _sql_names = ["PARSE_JSON", "JSON_PARSE"]
-    arg_types = {"this": True, "expressions": False}
-    is_var_len_args = True
```
georgesittas, line 5506: FYI the `is_var_lane_args` attribute seemed unneeded, see relevant [commit](https://github.com/tobymao/sqlglot/commit/d75d760778973d88325e0b6d167243385140d698) that added it

PR #3699, opened 2024-06-25
fix(optimizer): Force early alias expansion in BQ & CH

### LINKED ISSUE #3687: wrong qualify_columns on generated columns
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`: Done
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`: Done
- Check if the issue still exists on main: still exist

Prepare data:
```python

sql ="""with data as (
  select 1 as id, 2 as my_id, 'a' as name, 'de' as full_name
)
select
  id as my_id,
  concat(id, name) as full_name
from data
group by my_id, full_name"""

schema = {
"catalog": {
      "schema1": {
          "customer": {
              "id": "INTEGER",
          },
      }
  }
}
```

Transpiler from bigquery to trino:
```python
smt = parse_one(sql, read="bigquery")
print(qualify_columns(smt, schema=schema).sql(dialect="trino", pretty=True))
```
Output:
```sql
WITH data AS (
  SELECT
    1 AS id,
    2 AS my_id,
    'a' AS name,
    'de' AS full_name
)
SELECT
  data.id AS my_id,
  CONCAT(CAST(data.id AS VARCHAR), CAST(data.name AS VARCHAR)) AS full_name
FROM data
GROUP BY
  data.my_id,
  data.full_name
```

The problem of the output is it can not be run on trino, because `data.id` and `data.name` is not an aggregate expression or appear in GROUP BY clause.

Additional test:
The output will expand correctly if we no not provide schema:
```python
smt = parse_one(sql, read="bigquery")
print(qualify_columns(smt, schema=None).sql(dialect="trino", pretty=True))

"""
WITH data AS (
  SELECT
    1 AS id,
    2 AS my_id,
    'a' AS name,
    'de' AS full_name
)
SELECT
  data.id AS my_id,
  CONCAT(CAST(data.id AS VARCHAR), CAST(data.name AS VARCHAR)) AS full_name
FROM data
GROUP BY
  data.id,
  CONCAT(CAST(data.id AS VARCHAR), CAST(data.name AS VARCHAR))
"""
```

The output will break again if I disable expand_alias_ref: 
```python
smt = parse_one(sql, read="bigquery")
print(qualify_columns(smt, schema=None, expand_alias_refs=False).sql(dialect="trino", pretty=True))
```

### Solution
I have made a patch to fix my case, but I think we can have better solution to fix this.

https://github.com/viplazylmht/sqlglot/blob/644bf67720442f964ffc4606693dcfc1b87ec6e6/sqlglot/optimizer/qualify_columns.py#L423-L431

**Official Documentation**

VaggelisD:
Fixes #3687

Take the following query:

```SQL
WITH data AS (
SELECT
    1 AS id,
    2 AS my_id
)
SELECT
    id AS my_id
FROM
    data
WHERE
    my_id = 1
GROUP BY
    my_id,
HAVING
    my_id = 1
```

In most dialects this is an incorrect query as `GROUP BY my_id` is resolved to `GROUP BY data.my_id` which is not aggregated (the overlapping alias in the projection list is resolved later in the order of operations). 

However, the exceptions to this rule which _can_ process this query are the following:
- BigQuery, which only forwards the alias only to `GROUP BY` & `HAVING` i.e resolves to:
```SQL
WITH data AS (...)
WHERE data.my_id = 1 GROUP BY data.id HAVING data.id = 1
```

- Clickhouse, which forwards the alias across the query i.e resolves to:
```SQL
WITH data AS (...)
WHERE data.id = 1 GROUP BY data.id HAVING data.id = 1
```

This PR aims to fix this by forcing the alias expansion before `_qualify_columns()` for these cases, as the 2nd call to `_expand_alias_refs()` does not expand aliases on qualified columns.

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -60,8 +60,8 @@ def qualify_columns(
         _pop_table_column_aliases(scope.derived_tables)
         using_column_tables = _expand_using(scope, resolver)
 
-        if schema.empty and expand_alias_refs:
-            _expand_alias_refs(scope, resolver)
+        if (schema.empty or dialect.FORCE_EARLY_ALIAS_REF_EXPANSION) and expand_alias_refs:
+            _expand_alias_refs(scope, resolver, expand_only_groupby=(dialect == "bigquery"))
 
         _convert_columns_to_dots(scope, resolver)
         _qualify_columns(scope, resolver)
@@ -208,7 +208,7 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
     return column_tables
 
 
-def _expand_alias_refs(scope: Scope, resolver: Resolver) -> None:
+def _expand_alias_refs(scope: Scope, resolver: Resolver, expand_only_groupby: bool = False) -> None:
     expression = scope.expression
 
     if not isinstance(expression, exp.Select):
@@ -219,7 +219,7 @@ def _expand_alias_refs(scope: Scope, resolver: Resolver) -> None:
     def replace_columns(
         node: t.Optional[exp.Expression], resolve_table: bool = False, literal_index: bool = False
     ) -> None:
-        if not node:
+        if not node or (expand_only_groupby and not isinstance(node, exp.Group)):
             return
 
         for column in walk_in_scope(node, prune=lambda node: node.is_star):
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -473,13 +473,33 @@ SELECT :with,WITH :expressions,CTE :this,UNION :this,SELECT :expressions,1,:expr
             'SELECT "x"."a" + 1 AS "d", "x"."a" + 1 + 1 AS "e" FROM "x" AS "x" WHERE ("x"."a" + 2) > 1 GROUP BY "x"."a" + 1 + 1',
         )
 
+        unused_schema = {"l": {"c": "int"}}
         self.assertEqual(
             optimizer.qualify_columns.qualify_columns(
                 parse_one("SELECT CAST(x AS INT) AS y FROM z AS z"),
-                schema={"l": {"c": "int"}},
+                schema=unused_schema,
                 infer_schema=False,
             ).sql(),
             "SELECT CAST(x AS INT) AS y FROM z AS z",
+        )
+
+        # BigQuery expands overlapping alias only for GROUP BY + HAVING
+        sql = "WITH data AS (SELECT 1 AS id, 2 AS my_id, 'a' AS name, 'b' AS full_name) SELECT id AS my_id, CONCAT(id, name) AS full_name FROM data WHERE my_id = 1 GROUP BY my_id, full_name HAVING my_id = 1"
+        self.assertEqual(
+            optimizer.qualify_columns.qualify_columns(
+                parse_one(sql),
+                schema=MappingSchema(schema=unused_schema, dialect="bigquery"),
+            ).sql(),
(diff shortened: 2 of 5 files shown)


tests/test_optimizer.py
```diff
@@ -473,15 +473,35 @@ def test_expand_alias_refs(self):
...
                 infer_schema=False,
             ).sql(),
             "SELECT CAST(x AS INT) AS y FROM z AS z",
         )
 
+        # BigQuery expands overlapping alias only for GROUP BY + HAVING
+        sql = "WITH data AS (SELECT 1 AS id, 2 AS my_id, 'a' AS name, 'b' AS full_name) SELECT id AS my_id, CONCAT(id, name) AS full_name FROM data WHERE my_id = 1 GROUP BY my_id, full_name HAVING my_id = 1"
+        self.assertEqual(
+            optimizer.qualify_columns.qualify_columns(
+                parse_one(sql),
```
georgesittas, line 490:
```suggestion
                parse_one(sql, dialect="bigquery"),
```

tests/test_optimizer.py
```diff
@@ -473,15 +473,35 @@ def test_expand_alias_refs(self):
...
+                schema=MappingSchema(schema=unused_schema, dialect="bigquery"),
+            ).sql(),
+            "WITH data AS (SELECT 1 AS id, 2 AS my_id, 'a' AS name, 'b' AS full_name) SELECT data.id AS my_id, CONCAT(data.id, data.name) AS full_name FROM data WHERE data.my_id = 1 GROUP BY data.id, CONCAT(data.id, data.name) HAVING data.id = 1",
+        )
+
+        # Clickhouse expands overlapping alias across the entire query
+        self.assertEqual(
+            optimizer.qualify_columns.qualify_columns(
+                parse_one(sql),
```
georgesittas, line 499:
```suggestion
                parse_one(sql, dialect="clickhouse"),
```

georgesittas (approved):

sqlglot/optimizer/qualify_columns.py
```diff
@@ -60,8 +60,8 @@ def qualify_columns(
         _pop_table_column_aliases(scope.derived_tables)
         using_column_tables = _expand_using(scope, resolver)
 
-        if schema.empty and expand_alias_refs:
-            _expand_alias_refs(scope, resolver)
+        if (schema.empty or dialect.FORCE_EARLY_ALIAS_REF_EXPANSION) and expand_alias_refs:
+            _expand_alias_refs(scope, resolver, expand_only_groupby=(dialect == "bigquery"))
```
viplazylmht, line 64: Hello @VaggelisD 

Should we provide the `expand_only_groupby` via qualify_columns parameters? (we can infer it with your logic if user provide None).  

The context is if I extend the Bigquery dialect to MyBigquery dialect, this change will not work anymore, so I want to force it using a parameter passed to the function caller instead.

georgesittas: Can't you just set the flag in your dialect?

viplazylmht: Sure, `expand_only_groupby` inside the dialect is a good idea.

georgesittas: Ah you were talking about the expand only group logic, I completely missed it sorry. We could probably parameterize it, cc @VaggelisD

georgesittas: https://github.com/tobymao/sqlglot/pull/3702

PR #3701, opened 2024-06-25
Fix: proper parsing of unit in spark/databricks date_diff

### LINKED ISSUE #3700: [Databricks] Date units in functions interpreted as columns
First off, great module! 

I've started using sqlglot somewhat recently and ran into what I think may be a bug:

```
import sqlglot

sql = "select date_diff(day, created_at, current_date()) from test"
schema = {"test": {"created_at": "timestamp"}}
ast = parse_one(sql, dialect="databricks")
qualify(ast, schema=schema, dialect="databricks")
```
`sqlglot.errors.OptimizeError: Column '"day"' could not be resolved`

The [examples](https://docs.gcp.databricks.com/en/sql/language-manual/functions/date_diff.html#examples) in the Databricks docs show an unquoted date unit. I assume that other Databricks date/time functions that accept units may experience the same issue.

georgesittas:
Fixes #3700

For a followup PR: see if we can refactor the date deltas for spark/databricks. Perhaps we can use a single implementation for both.

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -50,6 +50,7 @@ class Databricks(Spark):
             "DATEADD": build_date_delta(exp.DateAdd),
             "DATE_ADD": build_date_delta(exp.DateAdd),
             "DATEDIFF": build_date_delta(exp.DateDiff),
+            "DATE_DIFF": build_date_delta(exp.DateDiff),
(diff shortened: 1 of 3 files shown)


cheesefactory (approved):
Amazingly fast. Thank you!

tobymao (approved):

PR #3702, opened 2024-06-26
Refactor: add EXPAND_ALIAS_REFS_ONLY_IN_GROUP_BY dialect constant

georgesittas:
Context: https://github.com/tobymao/sqlglot/pull/3699#discussion_r1653168484

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -351,8 +351,8 @@ class Dialect(metaclass=_Dialect):
         - Clickhouse, which will forward the alias across the query i.e it resolves to "WHERE id = 1 GROUP BY id HAVING id = 1"
     """
 
-    EXPAND_ALIAS_REFS_ONLY_IN_GROUP_BY = False
-    """Whether alias reference expansion should only happen for the GROUP BY clause."""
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -351,6 +351,9 @@ class Dialect(metaclass=_Dialect):
         - Clickhouse, which will forward the alias across the query i.e it resolves to "WHERE id = 1 GROUP BY id HAVING id = 1"
     """
 
+    EXPAND_ALIAS_REFS_ONLY_IN_GROUP_BY = False
+    """Whether alias reference expansion should only happen for the GROUP BY clause."""
+
```
VaggelisD, line 356: I think we should add the `EARLY` prefix somewhere in this flag e.g. `EARLY_EXPAND_ALIAS_REFS_IN_GROUPBY` as this is the differentiator in the original PR.

If `schema` is non-empty we will revisit the 2nd `_expand_alias_refs()` for the rest of the expansion logic, so maybe `ONLY_IN_GROUP_BY` can be confusing (?)

VaggelisD (approved):

PR #3703, opened 2024-06-26
feat(tsql): Add support for scope qualifier operator

VaggelisD:
Fixes [#SQLMesh/2817](https://github.com/TobikoData/sqlmesh/issues/2817)

Introduce support for `<data type>::<static_func>` in T-SQL.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -524,6 +524,13 @@ class TSQL(Dialect):
             "TIMEFROMPARTS": _build_timefromparts,
         }
 
+        RANGE_PARSERS = {
+            **parser.Parser.RANGE_PARSERS,
+            TokenType.DCOLON: lambda self, this: self.expression(
+                exp.ScopeResolution, this=this, expression=self._parse_function()
+            ),
+        }
+
         JOIN_HINTS = {"LOOP", "HASH", "MERGE", "REMOTE"}
 
         RETURNS_TABLE_TOKENS = parser.Parser.ID_VAR_TOKENS - {
@@ -535,6 +542,10 @@ class TSQL(Dialect):
             **parser.Parser.STATEMENT_PARSERS,
             TokenType.DECLARE: lambda self: self._parse_declare(),
         }
+
+        # The DCOLON (::) operator serves as a scope resolution (exp.ScopeResolution) operator in T-SQL
+        COLUMN_OPERATORS = parser.Parser.COLUMN_OPERATORS.copy()
+        COLUMN_OPERATORS.pop(TokenType.DCOLON)
 
         def _parse_options(self) -> t.Optional[t.List[exp.Expression]]:
             if not self._match(TokenType.OPTION):
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -527,7 +527,9 @@ class TSQL(Dialect):
         RANGE_PARSERS = {
             **parser.Parser.RANGE_PARSERS,
             TokenType.DCOLON: lambda self, this: self.expression(
-                exp.ScopeResolution, this=this, expression=self._parse_function()
+                exp.ScopeResolution,
+                this=this,
+                expression=self._parse_function() or self._parse_var(any_token=True),
             ),
         }
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5581,7 +5581,7 @@ class MapFromEntries(Func):
 
 
 # https://learn.microsoft.com/en-us/sql/t-sql/language-elements/scope-resolution-operator-transact-sql?view=sql-server-ver16
-class ScopeResolution(Func):
+class ScopeResolution(Expression):
     arg_types = {"this": False, "expression": True}
 
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5580,6 +5580,11 @@ class MapFromEntries(Func):
     pass
 
 
+# https://learn.microsoft.com/en-us/sql/t-sql/language-elements/scope-resolution-operator-transact-sql?view=sql-server-ver16
+class ScopeResolution(Func):
+    arg_types = {"this": False, "expression": True}
+
+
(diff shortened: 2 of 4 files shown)


georgesittas (commented):
This doesn't look quite right, we shouldn't handle just geography.

From what I've seen T-SQL doesn't use the `::` syntax for casting, so imo we should just create a new expression like `ScopeQualifier` to represent it and override the corresponding `COLUMN_OPERATORS` entry so that it produces it. That would yield a much simpler solution AFAICT.

See also: https://www.sqlteam.com/forums/topic.asp?TOPIC_ID=76723

georgesittas:
So from what I'm seeing `ScopeResolution` would be a binary in this case and it should just work

VaggelisD:
Sounds good, not sure why I remembered that `::` was a cast operator in T-SQL but it doesn't seem to be. 

Besides that, I'm not sure if the solution can be simplified much further because in `parser.py::_parse_column_ops()` the `TokenType.DCOLON` operator is hardcoded to parse types, so in `tsql.py::COLUMN_OPERATORS` the RHS expression will be a "user-defined" `exp.DataType` with the function params as `exp.DataTypeParams` instead of a single `exp.Func`, right?

georgesittas:
Yeah, I see what you mean. Maybe we should challenge the idea of keeping this as a "column operator" while we're at it. For instance, I saw the `::foo` syntax in one of the links I shared. If we can pop it from `COLUMN_OPERATORS` and handle it elsewhere then we won't have the issue you mentioned.

VaggelisD:
I saw the `::foo` syntax on the last reply of the 1st link (forum post) but in the same reply the author claims it's needed anymore after SQL Server 2005, not sure if you're referring to the same one.

If that's the case, then maybe we can preserve the "binary column operator" solution, although not semantically correct? I'm not seeing other ways out of it if we want to redesign it from scratch, will have to search this a bit better

georgesittas:
Can't we just use a range parser?

VaggelisD:
`RANGE_PARSER` works yeah, probably not semantically correct as well but it did simplify things.

sqlglot/expressions.py
```diff
@@ -5580,6 +5580,11 @@ class MapFromEntries(Func):
     pass
 
 
+# https://learn.microsoft.com/en-us/sql/t-sql/language-elements/scope-resolution-operator-transact-sql?view=sql-server-ver16
+class ScopeResolution(Func):
```
georgesittas, line 5584: `Func` doesn't look right here - let's opt for `Expression` instead.

sqlglot/dialects/tsql.py
```diff
@@ -524,6 +524,13 @@ class Parser(parser.Parser):
             "TIMEFROMPARTS": _build_timefromparts,
         }
 
+        RANGE_PARSERS = {
+            **parser.Parser.RANGE_PARSERS,
+            TokenType.DCOLON: lambda self, this: self.expression(
+                exp.ScopeResolution, this=this, expression=self._parse_function()
```
georgesittas, line 530: Let's use a more lenient parser for the RHS, eg. the 2nd [link](https://learn.microsoft.com/en-us/sql/t-sql/statements/grant-server-principal-permissions-transact-sql?view=sql-server-ver16) shows this syntax:

```sql
GRANT VIEW DEFINITION ON LOGIN::EricKurjan TO RMeyyappan   
    WITH GRANT OPTION
```

georgesittas (approved):
LGTM, couple of comments and should be good to go.

> probably not semantically correct as well but it did simplify things

Why not?

tests/dialects/test_tsql.py
```diff
@@ -1878,3 +1878,11 @@ def test_declare(self):
             "DECLARE vendor_cursor CURSOR FOR SELECT VendorID, Name FROM Purchasing.Vendor WHERE PreferredVendorStatus = 1 ORDER BY VendorID",
             check_command_warning=True,
         )
+
+    def test_scope_resolution_op(self):
+        self.validate_identity("SELECT ::FOO(a, b)")
+        self.validate_identity("SELECT bar::BAZ(1, 2)")
+        self.validate_identity("SELECT GEOGRAPHY::POINT([latitude], [longitude], 4326) FROM tbl")
```
georgesittas, line 1885: This is problematic after all, T-SQL is case-sensitive when it comes to these system functions:

```
1> SELECT GEOGRAPHY::POINT(34.2, 34.2, 4326)
2> go
Msg 6506, Level 16, State 10, Server af4359a4e415, Line 1
Could not find method 'POINT' for type 'Microsoft.SqlServer.Types.SqlGeography' in assembly 'Microsoft.SqlServer.Types'
1> SELECT GEOGRAPHY::Point(34.2, 34.2, 4326)
2> go
  
--
0x
```

georgesittas: Maybe we can switch off the function name normalization only for anonymous functions in T-SQL and control this behavior with a generator flag?

georgesittas: Or if we can prove this is the only case where case-sensitivity matters, check the parent of the `Func` and only normalize it if it's not a `ScopeResolution`.

VaggelisD: I tried a few different functions like date/time related ones and they all seem to be case insensitive except these, will disable normaliaztion for `ScopeResolution` funcs

georgesittas: Cool, let's only do this for T-SQL

VaggelisD: It's awkward to _not_ normalize some of the functions in a dialect; For example, as I mentioned in T-SQL the builtin functions like `DATEDIFF`, `CONVERT` etc are case-insensitive in contrast to the CLR functions like `geography::Point` which are case-sensitive.

In order not to be intrusive to the rest of T-SQL, I chose to manually generate the function RHS of `exp.ScopeResolution`; If we want to normalize all functions in T-SQL I can follow up with a commit and change the PR to breaking.

georgesittas (approved):

PR #3708, opened 2024-06-27
feat(duckdb): Transpile exp.Length from other dialects

VaggelisD:
Add generation for `exp.Length` in DuckDB. 

Dialects like Snowflake and BigQuery also use `LENGTH() / LEN()` for binary values, so this PR will generate `LENGTH(<expr>)` if the type can be inferred or a case/when with `LENGTH` for `VARCHAR` and `OCTET_LENGTH` for the binary (`BLOB`) value.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -755,6 +755,7 @@ class DuckDB(Dialect):
             if arg.is_type(*exp.DataType.TEXT_TYPES):
                 return self.func("LENGTH", arg)
 
+            # We need these casts to make duckdb's static type checker happy
             blob = exp.cast(arg, exp.DataType.Type.VARBINARY)
             varchar = exp.cast(arg, exp.DataType.Type.VARCHAR)
 
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -742,16 +742,19 @@ class DuckDB(Dialect):
         def length_sql(self, expression: exp.Length) -> str:
             arg = expression.this
 
+            # Dialects like BQ and Snowflake also accept binary values as args, so
+            # DDB will attempt to infer the type or resort to case/when resolution
+            if not expression.args.get("binary") or arg.is_string:
+                return self.func("LENGTH", arg)
+
             if not arg.type:
                 from sqlglot.optimizer.annotate_types import annotate_types
 
                 arg = annotate_types(arg)
 
-            if arg.is_type(exp.DataType.Type.VARCHAR) or arg.is_string:
+            if arg.is_type(*exp.DataType.TEXT_TYPES):
                 return self.func("LENGTH", arg)
 
-            # Dialects like BQ and Snowflake also accept binary values, so if we can't determine the
-            # arg type then attempt case-when resolution
             trycast_blob = exp.TryCast(this=arg, to=exp.DataType.build(exp.DataType.Type.VARBINARY))
             trycast_varchar = exp.TryCast(
                 this=arg, to=exp.DataType.build(exp.DataType.Type.VARCHAR)
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -755,17 +755,15 @@ class DuckDB(Dialect):
             if arg.is_type(*exp.DataType.TEXT_TYPES):
                 return self.func("LENGTH", arg)
 
-            trycast_blob = exp.TryCast(this=arg, to=exp.DataType.build(exp.DataType.Type.VARBINARY))
-            trycast_varchar = exp.TryCast(
-                this=arg, to=exp.DataType.build(exp.DataType.Type.VARCHAR)
-            )
+            blob = exp.cast(arg, exp.DataType.Type.VARBINARY)
+            varchar = exp.cast(arg, exp.DataType.Type.VARCHAR)
 
             case = (
                 exp.case(self.func("TYPEOF", arg))
                 .when(
-                    "'VARCHAR'", exp.Anonymous(this="LENGTH", expressions=[trycast_varchar])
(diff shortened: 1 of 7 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -738,3 +738,31 @@ def withingroup_sql(self, expression: exp.WithinGroup) -> str:
...
+        def length_sql(self, expression: exp.Length) -> str:
+            arg = expression.this
+
+            if not arg.type:
+                from sqlglot.optimizer.annotate_types import annotate_types
+
+                arg = annotate_types(arg)
+
+            if arg.is_type(exp.DataType.Type.VARCHAR) or arg.is_string:
+                return self.func("LENGTH", arg)
```
georgesittas, line 751: Let's flip order of conditions and do `is_type(*exp.DataType.TEXT_TYPES)` instead

sqlglot/dialects/duckdb.py
```diff
@@ -738,3 +738,31 @@ def withingroup_sql(self, expression: exp.WithinGroup) -> str:
...
+
+            case = (
+                exp.case(self.func("TYPEOF", arg))
+                .when(
+                    "'VARCHAR'", exp.Anonymous(this="LENGTH", expressions=[trycast_varchar])
+                )  # anonymous to break length_sql recursion
+                .when("'BLOB'", self.func("OCTET_LENGTH", trycast_blob))
+            )
+
+            return self.sql(case)
```
georgesittas, line 768: This will make `LENGTH(x)` generate as `CASE TYPEOF(x) WHEN 'VARCHAR' THEN LENGTH(TRY_CAST(x AS TEXT)) WHEN 'BLOB' THEN OCTET_LENGTH(TRY_CAST(x AS BLOB)) END` even for duckdb -> duckdb.

I'd rather we have an arg in `Length` that signifies whether binary values can also be used as arguments, and only do something else in duckdb if that arg is set properly.

VaggelisD: Good idea

georgesittas: Make sure to handle that arg gracefully if it's a boolean, because otherwise we'll crash in `Generator.sql`. We basically need to find all places where `Length` is generated (+ overrides) and exclude the flag.

sqlglot/generator.py
```diff
@@ -4005,3 +4005,6 @@ def parsejson_sql(self, expression: exp.ParseJSON) -> str:
             return self.sql(expression.this)
 
         return self.func(self.PARSE_JSON_NAME, expression.this, expression.expression)
+
+    def length_sql(self, expression: exp.Length) -> str:
+        return self.func("LENGTH", expression.this)
```
tobymao, line 4010: why does the default skip binary?

tobymao: is binary a non renderable arg?

VaggelisD: `binary` is a boolean flag that's added from BigQuery and Snowflake because they also accept `LENGTH(<binary_expr>)` alongside `LENGTH(<string_expr>)`, it's not a valid argument. 

Dialects like Redshift and DuckDB only accept strings, so if `binary` is set DDB will attempt to differentiate between the 2 by returning `LENGTH(<string_expr>)` or `OCTET_LENGTH(<binary_expr>)`

georgesittas: we're also gonna crash without this because it's a boolean, see https://github.com/tobymao/sqlglot/blob/main/sqlglot/generator.py#L798

sqlglot/dialects/duckdb.py
```diff
@@ -738,3 +738,34 @@ def withingroup_sql(self, expression: exp.WithinGroup) -> str:
...
+
+                arg = annotate_types(arg)
+
+            if arg.is_type(*exp.DataType.TEXT_TYPES):
+                return self.func("LENGTH", arg)
+
+            trycast_blob = exp.TryCast(this=arg, to=exp.DataType.build(exp.DataType.Type.VARBINARY))
+            trycast_varchar = exp.TryCast(
+                this=arg, to=exp.DataType.build(exp.DataType.Type.VARCHAR)
+            )
```
georgesittas, line 761: why do we need trycast vs a simple cast?

georgesittas (approved):

sqlglot/dialects/duckdb.py
VaggelisD: From my testing both worked so I think either is fine, but since this is a best-effort attempt `exp.TryCast` might handle any conversion errors more graciously.

tobymao: can you test it to see if a try cast is actually needed

VaggelisD: Right, it will return `NULL`. I've tried a few BQ -> DDB queries with both string and byte expressions, couldn't repro a scenario where the normal cast would break yet. If we also want to error out in such scenarios, then it's another reason to switch to simple casts

georgesittas (approved):

sqlglot/dialects/duckdb.py
```diff
@@ -738,3 +738,32 @@ def withingroup_sql(self, expression: exp.WithinGroup) -> str:
...
+            if not arg.type:
+                from sqlglot.optimizer.annotate_types import annotate_types
+
+                arg = annotate_types(arg)
+
+            if arg.is_type(*exp.DataType.TEXT_TYPES):
+                return self.func("LENGTH", arg)
+
+            blob = exp.cast(arg, exp.DataType.Type.VARBINARY)
+            varchar = exp.cast(arg, exp.DataType.Type.VARCHAR)
```
georgesittas, line 759: Seems like an important bit to comment
```suggestion
            # We need these casts to make duckdb's static type checker happy
            blob = exp.cast(arg, exp.DataType.Type.VARBINARY)
            varchar = exp.cast(arg, exp.DataType.Type.VARCHAR)
```

PR #3714, opened 2024-06-28
fix!: simplify no longer removes neg, add to_py

tobymao:


sqlglot/expressions.py
```diff
@@ -175,23 +175,25 @@ def is_number(self) -> bool:
         """
         Checks whether a Literal expression is a number.
         """
-        return isinstance(self, Literal) and not self.args["is_string"]
+        return (
+            isinstance(self, Literal)
+            and not self.args["is_string"]
+            or isinstance(self, Neg)
+            and self.this.is_number
```
georgesittas, line 182: should we wrap `or`'s operands here to make the order easier to see?

georgesittas (approved):
Nice!

sqlglot/dialects/postgres.py
```diff
@@ -56,12 +56,15 @@ def func(self: Postgres.Generator, expression: DATE_ADD_OR_SUB) -> str:
         this = self.sql(expression, "this")
         unit = expression.args.get("unit")
 
-        expression = self._simplify_unless_literal(expression.expression)
-        if not isinstance(expression, exp.Literal):
+        e = self._simplify_unless_literal(expression.expression)
+        if isinstance(e, exp.Literal):
+            e.args["is_string"] = True
+        elif e.is_number:
+            e = exp.Literal.string(e.to_py())
```
VaggelisD, line 63: Is the order in the branches correct?

 If `e.is_number` is `True` then `e` is an `exp.Literal(is_string=False)`, I think the `elif` branch won't be hit

georgesittas: > If `e.is_number` is `True` then `e` is an `exp.Literal(is_string=False)`

This isn't true, `is_number` now handles `Neg` expressions as well. This condition is supposed to handle `Neg`s.

VaggelisD: Sg, missed the new `is_number` definition

achicoine-coveo:
Thanks for the fix! I can confirm it worked and I was able to remove the macro I had as a workaround :)

PR #3715, opened 2024-06-28
Fix(optimizer)!: coalesce left-hand side of join conditions produced by expanding USING

georgesittas:
Prior to this change, we'd expand a `USING(col1, col2)` clause into an `ON` clause that compared the corresponding columns as is, eg. `ON t1.col1 = t2.col1 AND t1.col2 = t2.col2`.

That is fine when there are only two tables joined, but when more tables are involved it turns out that we need to coalesce all columns from the already-joined tables that appear in the left-hand side of these equalities. If we don't, the semantics are not preserved.

For example, this:

```sql
with t1 as (select 'x' as id, date '2024-01-01' as foo, 000 as value),
     t2 as (select 'x' as id, date '2024-02-02' as foo, 123 as value),
     t3 as (select 'x' as id, date '2024-02-02' as foo, 456 as value)
select *
from t1
full outer join t2 using(id, foo, value)
full outer join t3 using(id, foo, value)
```

Should be transformed into:

```sql
with t1 as (select 'x' as id, date '2024-01-01' as foo, 000 as value),
     t2 as (select 'x' as id, date '2024-02-02' as foo, 123 as value),
     t3 as (select 'x' as id, date '2024-02-02' as foo, 456 as value)
select
  coalesce(t1.id, t2.id, t3.id) as id,
  coalesce(t1.foo, t2.foo, t3.foo) as foo,
  t1.value as value,
  t2.value as value,
  t3.value as value
from t1
full outer join t2 on t1.id = t2.id and t1.foo = t2.foo
full outer join t3 on coalesce(t1.id, t2.id) = t3.id and coalesce(t1.foo, t2.foo) = t3.foo
```

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -152,7 +152,7 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
     # Mapping of automatically joined column names to an ordered set of source names (dict).
     column_tables: t.Dict[str, t.Dict[str, t.Any]] = {}
 
-    for join in joins:
+    for i, join in enumerate(joins):
         using = join.args.get("using")
 
         if not using:
@@ -172,6 +172,7 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
         ordered.append(join_table)
         join_columns = resolver.get_source_columns(join_table)
         conditions = []
+        using_identifier_count = len(using)
 
         for identifier in using:
             identifier = identifier.name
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -200,8 +205,8 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
         for column in scope.columns:
             if not column.table and column.name in column_tables:
                 tables = column_tables[column.name]
-                coalesce = [exp.column(column.name, table=table) for table in tables]
-                replacement = exp.Coalesce(this=coalesce[0], expressions=coalesce[1:])
+                coalesce_args = [exp.column(column.name, table=table) for table in tables]
+                replacement = exp.func("coalesce", *coalesce_args)
```
VaggelisD, line 209: Out of curiosity why was `exp.func` preferred over building `exp.Coalesce` directly?

tobymao: we always use exp.func when we can because it's more succint

tobymao (approved):

Issue #3722, opened 2024-07-02
teradata should use `timestamp with time zone` not `timestamptz`


mtagle:
supporting documentation: https://docs.teradata.com/r/Enterprise_IntelliFlex_VMware/SQL-Data-Types-and-Literals/DateTime-and-Interval-Data-Types/TIMESTAMP-WITH-TIME-ZONE-Data-Type

code snippit:

```print(sqlglot.transpile("select cast(current_timestamp(6) as timestamp with time zone)", read="teradata", write="teradata")[0])```

results in:

```
SELECT CAST(CURRENT_TIMESTAMP(6) AS TIMESTAMPTZ)
```

but should be
```
SELECT CAST(CURRENT_TIMESTAMP(6) AS TIMESTAMP WITH TIME ZONE)
```

looks like this fix was made for oracle previously so it should be pretty easy
https://github.com/tobymao/sqlglot/commit/b7fb8006b825ec36d62135877b4e32fccad8d044

tobymao:
feel free to submit a pr for this.

PR #3723, opened 2024-07-02
fix(teradata): use timestamp with time zone over timestamptz

mtagle:
fixes [3722](https://github.com/tobymao/sqlglot/issues/3722)

after making the change, `current_timestamp` in my example sql no longer retained it's precision. I'm not sure why this happened but modifying the current timestamp transform seemed to fix it.

regarding this change: current_timestamp can have a precision of 0-6, but `current_timestamp()` (with no arguments) is not valid sql.
documentation reference: https://docs.teradata.com/r/Enterprise_IntelliFlex_VMware/SQL-Functions-Expressions-and-Predicates/Built-In-Functions/CURRENT_TIMESTAMP/CURRENT_TIMESTAMP-Function-Syntax so I tried to represent this in the transform

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -223,11 +223,13 @@ class Teradata(Dialect):
         TABLESAMPLE_KEYWORDS = "SAMPLE"
         LAST_DAY_SUPPORTS_DATE_PART = False
         CAN_IMPLEMENT_ARRAY_ANY = True
+        TZ_TO_WITH_TIME_ZONE = True
 
         TYPE_MAPPING = {
             **generator.Generator.TYPE_MAPPING,
             exp.DataType.Type.GEOMETRY: "ST_GEOMETRY",
             exp.DataType.Type.DOUBLE: "DOUBLE PRECISION",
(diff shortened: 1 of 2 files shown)


tobymao:
please run make style

mtagle:
when I run `make style`, even on a clean master, I get this:
```
pre-commit run --all-files
make: pre-commit: No such file or directory
make: *** [style] Error 1
```
what am I missing?

georgesittas (approved):
Hey @mtagle

> when I run `make style`, even on a clean master, I get this:
> 
> ```
> pre-commit run --all-files
> make: pre-commit: No such file or directory
> make: *** [style] Error 1
> ```
> 
> what am I missing?

It looks to me like you don't have pre-commit installed. Can you first try `make install-dev`, then `install-pre-commit` and then run `make style` again?

georgesittas:
I'll take this one to the finish line, thanks for the PR :-)

georgesittas:
Slightly refactored this while I was at it, take a look here - https://github.com/tobymao/sqlglot/commit/84416d207a2e397aba12a4138fcbd1fab382c22d.

PR #3725, opened 2024-07-02
feat(presto, trino): Configurable transpilation of Snowflake VARIANT

### LINKED ISSUE #3713: Snowflake nested fields are incorrectly translated using `JSON_EXTRACT`
**Before you file an issue**

```
print(sqlglot.transpile("select payload:testField from snowflake_table;", read="snowflake", write="trino", pretty=True)[0])
```
*Output:*
```
SELECT
  JSON_EXTRACT(payload, '$.testField')
FROM snowflake_table
```
*Expected output:*
```
SELECT
  payload.testField
FROM snowflake_table
```
`payload:testField` is how nested fields of VARIANT type are accessed in Snowflake, they should be translated into `payload.testField` for Trino, since they would be or `row` type in Trino and not JSON.
**Official Documentation**
[Querying VARIANT nested fields in Snowflake](https://docs.snowflake.com/en/user-guide/querying-semistructured#traversing-semi-structured-data)
[Trino nested field query example](https://trino.io/blog/2020/08/14/dereference-pushdown.html#example)

VaggelisD:
Fixes #3713

Both Snowflake and Databricks have a `VARIANT` type with similar extract notation e.g. `col:x.y` which can be used to store any data (most notably, semi-structured). Up until now, SQLGlot treated this notation as a JSON extraction which is not always correct; As the issue showcased, since Snowflake does not have a `STRUCT` data type, Iceberg represents it's `STRUCT` as a `VARIANT` column which should map to a `ROW` column in Presto/Trino instead of `JSON`.

This PR focuses on Snowflake -> Presto/Trino and attempts to lift off this restriction by introducing a new dialect setting which can make the `exp.JSONExtract` generation configurable such that it can either: 
1. Generate `JSON_EXTRACT(col, '$.x.y')` for JSON data (as was before) or 
2. Transform the JSONPath to the equivalent ROW access in dot notation `col.x.y`

Note that SQLGlot is unable to statically infer the _actual_ type of the data stored in a `VARIANT` column so transpilation to other dialects remains a best-effort attempt.

Documentation
---------------------

- [Snowflake VARIANT](https://docs.snowflake.com/en/sql-reference/data-types-semistructured#variant)
- [Snowflake VARIANT access](https://docs.snowflake.com/en/user-guide/querying-semistructured)
- [Iceberg - Snowflake STRUCT type](https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types#other-data-types)
- [Databricks VARIANT](https://docs.databricks.com/en/semi-structured/variant.html)

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -518,11 +518,6 @@ class Dialect(metaclass=_Dialect):
                     value: t.Union[bool | str | None] = None
 
                     if len(pair) == 1:
-                        if key.lower() == "normalization_strategy":
-                            raise AttributeError(
-                                "Dialect setting 'normalization_strategy' requires a value."
-                            )
-
                         # Default initialize standalone settings to True
                         value = True
                     elif len(pair) == 2:
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -511,7 +511,7 @@ class Dialect(metaclass=_Dialect):
         if isinstance(dialect, str):
             try:
                 dialect_name, *kv_strings = dialect.split(",")
-                kv_pairs = [kv.split("=") for kv in kv_strings]
+                kv_pairs = (kv.split("=") for kv in kv_strings)
                 kwargs = {}
                 for pair in kv_pairs:
                     key = pair[0].strip()
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -510,8 +510,28 @@ class Dialect(metaclass=_Dialect):
             return dialect
         if isinstance(dialect, str):
             try:
-                dialect_name, *kv_pairs = dialect.split(",")
-                kwargs = {k.strip(): v.strip() for k, v in (kv.split("=") for kv in kv_pairs)}
+                dialect_name, *kv_strings = dialect.split(",")
+                kv_pairs = [kv.split("=") for kv in kv_strings]
+                kwargs = {}
+                for pair in kv_pairs:
+                    key = pair[0].strip()
+                    value: t.Union[bool | str | None] = None
+
+                    if len(pair) == 1 and not key.lower() == "normalization_strategy":
+                        # Default initialize standalone settings to True except normalization strategy
+                        value = True
+                    elif len(pair) == 2:
+                        value = pair[1].strip()
+
+                        # Coerce the value to boolean if it matches to the truthy/falsy values below
+                        value_lower = value.lower()
+                        if value_lower in ("true", "1"):
+                            value = True
+                        elif value_lower in ("false", "0"):
+                            value = False
+
+                    kwargs[key] = value
+
             except ValueError:
                 raise ValueError(
                     f"Invalid dialect format: '{dialect}'. "
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -517,8 +517,13 @@ class Dialect(metaclass=_Dialect):
                     key = pair[0].strip()
                     value: t.Union[bool | str | None] = None
 
-                    if len(pair) == 1 and not key.lower() == "normalization_strategy":
-                        # Default initialize standalone settings to True except normalization strategy
+                    if len(pair) == 1:
+                        if key.lower() == "normalization_strategy":
+                            raise AttributeError(
+                                "Dialect setting 'normalization_strategy' requires a value."
+                            )
+
+                        # Default initialize standalone settings to True
                         value = True
                     elif len(pair) == 2:
                         value = pair[1].strip()
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -191,7 +191,13 @@ def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str
 
     # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.x.y) to a ROW access col.x.y
     segments = []
-    for path_key in expression.find_all(exp.JSONPathKey):
+    for path_key in expression.expression.expressions:
+        if isinstance(path_key, exp.JSONPathRoot):
+            continue
+        if not isinstance(path_key, exp.JSONPathKey):
+            # Cannot transpile subscripts, wildcards etc to dot notation
+            self.unsupported(f"Cannot transpile JSONPath segment '{path_key}' to ROW access")
+            continue
         key = path_key.this
         if not exp.SAFE_IDENTIFIER_RE.match(key):
             key = f'"{key}"'
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -176,10 +176,6 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
 def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str:
     is_json_extract = self.dialect.settings.get("variant_extract_is_json_extract", True)
 
-    if isinstance(is_json_extract, str):
-        # if the kwarg was set by the user it's stored in a string
-        is_json_extract = is_json_extract.lower() == "true"
-
     # Generate JSON_EXTRACT unless the user has configured that a Snowflake / Databricks
     # VARIANT extract (e.g. col:x.y) should map to dot notation (i.e ROW access) in Presto/Trino
     if not expression.args.get("variant_extract") or is_json_extract:
@@ -191,9 +187,7 @@ def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str
 
     # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.x.y) to a ROW access col.x.y
     segments = []
-    for path_key in expression.expression.expressions:
-        if isinstance(path_key, exp.JSONPathRoot):
-            continue
+    for path_key in expression.expression.expressions[1:]:
         if not isinstance(path_key, exp.JSONPathKey):
             # Cannot transpile subscripts, wildcards etc to dot notation
             self.unsupported(f"Cannot transpile JSONPath segment '{path_key}' to ROW access")
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -173,6 +173,29 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
     return f"FROM_UNIXTIME(CAST({timestamp} AS DOUBLE) / POW(10, {scale}))"
 
 
+def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str:
+    is_json_extract = self.dialect.settings.get("VARIANT_EXTRACT_IS_JSON_EXTRACT", True)
+
+    if isinstance(is_json_extract, str):
+        # if the kwarg was set by the user it's stored in a string
+        is_json_extract = is_json_extract.lower() == "true"
+
+    # Generate JSON_EXTRACT unless the user has configured that a Snowflake / Databricks
+    # VARIANT extract (e.g. col:x.y) should map to dot notation (i.e ROW access) in Presto/Trino
+    if not expression.args.get("variant_extract") or is_json_extract:
+        return self.func(
+            "JSON_EXTRACT", expression.this, expression.expression, *expression.expressions
+        )
+
+    this = self.sql(expression, "this")
+    expr = self.sql(expression, "expression")
+
+    # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.k1.k2') to a ROW access col.k1.k2
+    expr = expr.replace("'", "").lstrip("$")
+
+    return f"{this}{expr}"
+
+
 def _to_int(expression: exp.Expression) -> exp.Expression:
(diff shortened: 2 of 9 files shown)


sqlglot/dialects/presto.py
```diff
@@ -173,6 +173,29 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
     return f"FROM_UNIXTIME(CAST({timestamp} AS DOUBLE) / POW(10, {scale}))"
 
 
+def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str:
+    is_json_extract = self.dialect.settings.get("VARIANT_EXTRACT_IS_JSON_EXTRACT", True)
```
georgesittas, line 177: let's not do uppercase, it's inconsistent

georgesittas: also `prop=true` in the dialect settings looks a bit verbose, should we just update the setting parsing logic so you can do something like `dialect=trino,normalization_strategy=case_sensitive,variant_extract_is_json_extract` and then just do `dialect.settings.get("variant_extract_is_json_extract")`? cc @tobymao

VaggelisD:
CI tests failed on a snowflake executor test case, taking a look now

sqlglot/dialects/presto.py
```diff
@@ -173,6 +173,29 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
...
+    if not expression.args.get("variant_extract") or is_json_extract:
+        return self.func(
+            "JSON_EXTRACT", expression.this, expression.expression, *expression.expressions
+        )
+
+    this = self.sql(expression, "this")
+    expr = self.sql(expression, "expression")
+
+    # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.k1.k2') to a ROW access col.k1.k2
+    expr = expr.replace("'", "").lstrip("$")
```
georgesittas, line 194: is this `replace` call correct? what if a key has single quotes in it? maybe it would be more robust if we just get what we need from the ast before we generate `expression`

georgesittas:
PR looks good @VaggelisD, couple of comments only

sqlglot/dialects/presto.py
VaggelisD: I agree, not only that but the kwarg parsing turns the RHS values to strings so you also have to convert `'True'` to bool `True` before checking the condition, making it even more verbose. 

Should we then add the following in the `settings` parsing logic?
1. If a setting does not have a value, insert `True`
2. If a setting has the `str` value  `'True'` / `'False'` and swap out the `str` value to a `bool` one

sqlglot/dialects/presto.py
VaggelisD: I made the assumption here that the `VARIANT` extract notation will not contain special strings, subscripts etc, because in Presto/Trino the dot notation does not seem to handle these.

Would lstripping/rstripping the quotes work better in this case? I'll try out manually generating the path but it will add some code

sqlglot/dialects/presto.py
georgesittas: Sure, I like that, let's see what Toby thinks.

sqlglot/dialects/presto.py
georgesittas: > [...] in Presto/Trino the dot notation does not seem to handle these.

Did you try to use double quotes? Like `x."key with spaces"`.

> Would lstripping + rstripping the quotes work better in this case? I'll try out manually generating the path but it will add some code

I'm not sure why there's a need to strip at all. I see two cases: `JSONExtract.expression` is a string literal, in which case we can simply do `expression.expression.name` to get the path without the surrounding quotes. The other case is that it's not a literal, eg. it's a column, in which case we can't really do something (?).

VaggelisD: > Did you try to use double quotes? Like x."key with spaces".

Right, this works:

`WITH t AS(SELECT CAST(ROW('value', 1) AS ROW("str1 str2" VARCHAR, num INTEGER)) as col) select col."str1 str2" from t;`

> I'm not sure why there's a need to strip at all.
Yeah, if the JSONPath string is "hidden" behind a column this transformation breaks. I see what you mean with `.name`

sqlglot/dialects/presto.py
```diff
@@ -188,10 +188,16 @@ def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str
         )
 
     this = self.sql(expression, "this")
-    expr = self.sql(expression, "expression")
 
-    # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.k1.k2') to a ROW access col.k1.k2
-    expr = expr.replace("'", "").lstrip("$")
+    # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.x.y) to a ROW access col.x.y
+    segments = []
+    for path_key in expression.find_all(exp.JSONPathKey):
```
georgesittas, line 194: why are we doing `find_all` here? looks like a smell, this doesn't guarantee an order

sqlglot/dialects/presto.py
georgesittas: So how are we going to handle a path stored within a column? What do we currently do?

sqlglot/dialects/presto.py
VaggelisD: Thought that BFS should be fine but swapped to a for-loop to make sure

sqlglot/dialects/presto.py
```diff
@@ -173,6 +173,37 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
...
+            "JSON_EXTRACT", expression.this, expression.expression, *expression.expressions
+        )
+
+    this = self.sql(expression, "this")
+
+    # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.x.y) to a ROW access col.x.y
+    segments = []
+    for path_key in expression.expression.expressions:
+        if not isinstance(path_key, exp.JSONPathKey):
+            continue
```
georgesittas, line 196: When can we ever _not_ get a `JSONPathKey` here? I'm worried that `continue` is gonna silently drop information that it shouldn't.

VaggelisD: My thought was to rule out `exp.JSONPathRoot` (and other segment parts such as `exp.JSONPathSubscript` if they appear), as they shouldn't be transpilable to dot notation. Do we want to only remove the root and treat any other malformed input as malformed output?

georgesittas: We shouldn't skip anything silently. You can do an `lstrip` to remove the root `$.`, but for other parts that we can't generate we should do `self.unsupported`, imho.

sqlglot/dialects/presto.py
```diff
@@ -173,6 +173,37 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
     return f"FROM_UNIXTIME(CAST({timestamp} AS DOUBLE) / POW(10, {scale}))"
 
 
+def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str:
+    is_json_extract = self.dialect.settings.get("variant_extract_is_json_extract", True)
+
+    if isinstance(is_json_extract, str):
+        # if the kwarg was set by the user it's stored in a string
+        is_json_extract = is_json_extract.lower() == "true"
```
tobymao, line 181: i don't think we need lower

sqlglot/dialects/presto.py
VaggelisD: Isn't it fine to skip `exp.JSONPathRoot` and to hit `self.unsupported` for all non path keys?

sqlglot/dialects/presto.py
VaggelisD: Just tested this again, `is_json_extract` is exactly what the user has passed in the dialect settings so it's case sensitive afaict. 

Was having a [https://github.com/tobymao/sqlglot/pull/3725#pullrequestreview-2153477185](discussion) with George regarding this, would appreciate your input

sqlglot/dialects/presto.py
VaggelisD: You mean something like the following?

```
with t as (select parse_json('{'z': 1}') as col, 'z' as col2) select col:col2 from t;
```

Such cases seem to be resolved to `NULL` in Snowflake, meaning that `col2` will be resolved as a key and not as `col:"t.col2" -> col:'z'`. 

Even if that were the case, Snowflake does not allow single-quoted identifiers such as `col:'z'` in the VARIANT notation and you can't define an identifier as alias (e.g. `WITH t AS (..., "z" as col2) ...`) to trick it into doing so afaict.

sqlglot/dialects/presto.py
tobymao: you should be able to use any truthy value

VaggelisD: Was experimenting with this, haven't found a combination of built-in functions which perfectly evaluate strings to bools. Is it fine to make a dict of accepted truthy values (e.g. "true",  "t", non-zero integer strings etc) and similarly one for falsy ones?

georgesittas: See https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/__init__.py#L190-L200

tobymao: let's only support 1, 0, true, false

sqlglot/dialects/presto.py
georgesittas: Sounds good

sqlglot/dialects/presto.py
georgesittas: Isn't that achieved with the above `if` block?

VaggelisD: Sorry if I'm misunderstanding, I mean that all JSONPaths are canonicalized with a leading `exp.JSONPathRoot`  but that shouldn't be rendered in dot notation, so I guess it can be skipped altogether; any other non `exp.JSONPathKey` expression can raise unsupported.

georgesittas: ah yea sonuds good

sqlglot/dialects/dialect.py
```diff
@@ -510,8 +510,28 @@ def get_or_raise(cls, dialect: DialectType) -> Dialect:
...
+                dialect_name, *kv_strings = dialect.split(",")
+                kv_pairs = (kv.split("=") for kv in kv_strings)
+                kwargs = {}
+                for pair in kv_pairs:
+                    key = pair[0].strip()
+                    value: t.Union[bool | str | None] = None
+
+                    if len(pair) == 1 and not key.lower() == "normalization_strategy":
+                        # Default initialize standalone settings to True except normalization strategy
+                        value = True
```
georgesittas, line 522: Let's not check for `normalization_strategy`, we will fail in the constructor anyway and point out the user error that way.

sqlglot/dialects/presto.py
```diff
@@ -173,6 +173,41 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
...
```
georgesittas, line 181: This is irrelevant now right?
```suggestion
```

sqlglot/dialects/presto.py
```diff
@@ -173,6 +173,41 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
...
+            "JSON_EXTRACT", expression.this, expression.expression, *expression.expressions
+        )
+
+    this = self.sql(expression, "this")
+
+    # Convert the JSONPath extraction `JSON_EXTRACT(col, '$.x.y) to a ROW access col.x.y
+    segments = []
+    for path_key in expression.expression.expressions:
+        if isinstance(path_key, exp.JSONPathRoot):
+            continue
```
georgesittas, line 196:
```suggestion
    for path_key in expression.expression.expressions[1:]:
```

tests/dialects/test_dialect.py
```diff
@@ -102,14 +102,13 @@ def test_get_or_raise(self):
...
-            "Invalid dialect format: 'mysql, normalization_strategy'. "
-            "Please use the correct format: 'dialect [, k1 = v2 [, ...]]'.",
-        )
+        for dialect in ("mysql", "presto", "bigquery", "databricks"):
+            default_dialect = Dialect.get_or_raise(dialect)
+            lowercase_dialect = Dialect.get_or_raise(f"{dialect}, normalization_strategy")
+            self.assertEqual(
+                lowercase_dialect.normalization_strategy.value,
+                default_dialect.normalization_strategy.value,
+            )
```
georgesittas, line 111: Let's just keep the old test and change `ValueError` -> `AttributeError`. I don't think allowing `normalization_strategy` without value makes sense, it's most certainly a user error

georgesittas (approved):

sqlglot/dialects/presto.py
VaggelisD: Oops right, shouldn't be a string anymore

sqlglot/dialects/dialect.py
VaggelisD: This part comes before the dialect constructor right? So we can change this condition to the following one

```
if len(pair) == 1 :
  if key.lower() == "normalization_strategy":
    # throw AttributEerror
    
   # Default initialize standalone settings to True except normalization strategy
   value = True
```

PR #3727, opened 2024-07-02
Fix(clickhouse): switch off table alias columns generation

### LINKED ISSUE #3726: invalid sql generated for clickhouse when using a table alias with explicit column names
```
In [8]: import sqlglot as sg

In [9]: sg.__version__
Out[9]: '25.4.1'

In [10]: print(sg.parse_one("SELECT a, b FROM (SELECT * FROM x) AS t(a, b)", read="duckdb").sql('clickhouse', pretty=1))
SELECT
  a,
  b
FROM (
  SELECT
    *
  FROM x
) AS t(a, b)
```

**Official Documentation**

There are no syntax diagrams in their documentation, so it's hard to tell with ClickHouse other than to try to run the query:

```
localhost :) create or replace view x as select 1 a, 'a' b union all select 2 a, 'b' b union all select 3 a, 'c' b;

CREATE OR REPLACE VIEW x
AS SELECT
    1 AS a,
    'a' AS b
UNION ALL
SELECT
    2 AS a,
    'b' AS b
UNION ALL
SELECT
    3 AS a,
    'c' AS b

Query id: 46155cf8-28fd-4e69-8193-1f8cc51223c0

Ok.

0 rows in set. Elapsed: 0.001 sec.

localhost :) select * from x;

SELECT *
FROM x

Query id: dcddf522-4bee-428a-808a-7a79d801c32c

   ┌─a─┬─b─┐
1. │ 3 │ c │
   └───┴───┘
   ┌─a─┬─b─┐
2. │ 2 │ b │
   └───┴───┘
   ┌─a─┬─b─┐
3. │ 1 │ a │
   └───┴───┘

3 rows in set. Elapsed: 0.002 sec.

localhost :) select a, b from (select * from x) as t (a, b);

Syntax error: failed at position 41 ('('):

select a, b from (select * from x) as t (a, b);

Expected one of: FINAL, SAMPLE, table, table function, subquery or list of joined tables, array join, LEFT ARRAY JOIN, INNER, ARRAY JOIN, JOIN, PREWHERE, WHERE, GROUP BY, WITH, HAVING, WINDOW, ORDER BY, LIMIT, OFFSET, FETCH, SETTINGS, UNION, EXCEPT, INTERSECT, INTO OUTFILE, FORMAT, end of query
```

georgesittas:
Fixes #3726

Screenshot 2024-07-02 at 2 43 58 PM

Screenshot 2024-07-02 at 2 44 15 PM

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -683,6 +683,7 @@ class ClickHouse(Dialect):
         EXPLICIT_SET_OP = True
         GROUPINGS_SEP = ""
         SET_OP_MODIFIERS = False
+        SUPPORTS_TABLE_ALIAS_COLUMNS = False
 
         STRING_TYPE_MAPPING = {
             exp.DataType.Type.CHAR: "String",
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -7,25 +7,6 @@ class TestClickhouse(Validator):
     dialect = "clickhouse"
 
     def test_clickhouse(self):
-        self.validate_identity("extract(haystack, pattern)")
-
-        self.validate_all(
-            "SELECT * FROM x PREWHERE y = 1 WHERE z = 2",
-            write={
-                "": "SELECT * FROM x WHERE z = 2",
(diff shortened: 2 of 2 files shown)


cpcloud:
Nice! Thank you!

PR #3729, opened 2024-07-02
Fix(clickhouse): preserve EXTRACT(date_part FROM datetime) calls

### LINKED ISSUE #3728: incorrect attempt to parse every `extract` call as `RegexpExtract` when parsing clickhouse sql

```
In [5]: import sqlglot as sg

In [6]: sg.__version__
Out[6]: '25.4.1'

In [7]: sg.parse_one("SELECT extract(year from toDateTime('2023-02-01'))", read="clickhouse")
...
ParseError: Required keyword: 'expression' missing for <class 'sqlglot.expressions.RegexpExtract'>. Line 1, Col: 24.
  SELECT extract(year from toDateTime('2023-02-01'))
```

**Official Documentation**

- https://clickhouse.com/docs/en/sql-reference/operators#extract is the operator
- https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#extract is the string search function

georgesittas:
Fixes #3728

I initially tried to produce a `RegexpExtract` node for `extract(haystack, needle)`, but there are two problems with it:

- If we change the generation of `RegexpExtract` so that we produce `extract`, we may cause a transpilation regression because `extract` expects only 2 arguments whereas `RegexpExtract` can have more than two. So I crossed this one off.
- The remaining choice is to convert `extract(haystack, needle)` into `REGEXP_EXTRACT(haystack, needle)`, but I noticed the two aren't equivalent. There is a CH [issue](https://github.com/ClickHouse/ClickHouse/issues/56393) that goes into more detail, so I won't expand more on this for now.

I opted for `Anonymous` because wasn't sure if this one's worth spending too much time on. @cpcloud any thoughts here?


cpcloud:
Anonymous seems fine to me. Admittedly we're using `extractGroups` inside of Ibis for regex extraction, so the choice of anonymous likely wouldn't affect us.

cpcloud:
That said, my POV is very Ibis-centric :)

georgesittas:
Sounds good, thanks for the input 👍 

I'll wait to see what the other folks think about this.

tobymao (approved):

PR #3734, opened 2024-07-03
fix(oracle)!: Decouple NVL() from COALESCE()

### LINKED ISSUE #3733: NVL should NOT be replaced by COALESCE in Oracle but stay NVL. Their behavior are not the same in term of data type conversion.
In oracle DB :
```
CREATE TABLE TEST1 (COL1 INT, COL2 VARCHAR2(10))
INSERT INTO TEST1 VALUES (1,'2')
```
```
SELECT * FROM TEST1 WHERE NVL(COL2, 1) <> COL1
```
--> COL1   COL2
              1   2

```
SELECT * FROM TEST1 WHERE COALESCE(COL2, 1) <> COL1
```
--> ORA-00932: inconsistent datatypes: expected : CHAR ; got : NUMBER

As we can see, NVL does NOT care about the type, COALESCE does. Therefore both function are not identical.

In python with sqlglot:
sqlQuery: str = """ SELECT * FROM TEST1 WHERE NVL(COL2, 1) <> COL1"""
sql = sqlglot.parse_one(sql=sqlQuery, read='oracle')
print(sql.sql(dialect='oracle'))
-> SELECT * FROM TEST1 WHERE COALESCE(COL2, 1) <> COL1
I should have NVL and not COALESCE. These 2 functions are not identical.

**Official Documentation**

https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/NVL.html
"If expr1 is character data, then Oracle Database converts expr2 to the data type of expr1 before comparing them and returns VARCHAR2 in the character set of expr1."

VaggelisD:
Fixes #3733

In Oracle, these two functions have [different behaviors](https://stackoverflow.com/a/34830327), with the main one being that NVL implicitly converts the args to the same type; This means that SQLGlot cannot canonicalize`NVL()` to `COALESCE()` without type annotation & schema.

This PR aims to decouple them by removing the alias from `exp.Coalesce` and treating `NVL()` as an anonymous function. This should not be the case for Redshift, which does treat `NVL()` & `COALESCE()` as exactly the same function.

Docs
---------
[Oracle NVL](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/NVL.html) | [Oracle COALESCE](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/COALESCE.html) | [Redshift NVL/COALESCE](https://docs.aws.amazon.com/redshift/latest/dg/r_NVL_function.html)

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -107,6 +107,7 @@ class Oracle(Dialect):
             "TO_TIMESTAMP": build_formatted_time(exp.StrToTime, "oracle"),
             "TO_DATE": build_formatted_time(exp.StrToDate, "oracle"),
         }
+        FUNCTIONS.pop("NVL")
 
         FUNCTION_PARSERS: t.Dict[str, t.Callable] = {
             **parser.Parser.FUNCTION_PARSERS,
--- a/tests/dialects/test_oracle.py
+++ b/tests/dialects/test_oracle.py
@@ -88,7 +88,7 @@ class TestOracle(Validator):
         self.validate_identity(
             "SELECT * FROM T ORDER BY I OFFSET NVL(:variable1, 10) ROWS FETCH NEXT NVL(:variable2, 10) ROWS ONLY",
         )
-        self.parse_one("NVL(x, y)").assert_is(exp.Anonymous)
+        self.validate_identity("NVL(x, y)").assert_is(exp.Anonymous)
         self.validate_identity(
             "SELECT * FROM t SAMPLE (.25)",
             "SELECT * FROM t SAMPLE (0.25)",
--- a/tests/dialects/test_oracle.py
+++ b/tests/dialects/test_oracle.py
@@ -86,9 +86,9 @@ class TestOracle(Validator):
             "SELECT DISTINCT col1, col2 FROM table",
         )
         self.validate_identity(
-            "SELECT * FROM T ORDER BY I OFFSET nvl(:variable1, 10) ROWS FETCH NEXT nvl(:variable2, 10) ROWS ONLY",
-            "SELECT * FROM T ORDER BY I OFFSET COALESCE(:variable1, 10) ROWS FETCH NEXT COALESCE(:variable2, 10) ROWS ONLY",
+            "SELECT * FROM T ORDER BY I OFFSET NVL(:variable1, 10) ROWS FETCH NEXT NVL(:variable2, 10) ROWS ONLY",
         )
+        self.parse_one("NVL(x, y)").assert_is(exp.Anonymous)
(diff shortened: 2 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -5005,7 +5005,7 @@ class Ceil(Func):
 class Coalesce(Func):
     arg_types = {"this": True, "expressions": False}
     is_var_len_args = True
-    _sql_names = ["COALESCE", "IFNULL", "NVL"]
+    _sql_names = ["COALESCE", "IFNULL"]
```
georgesittas, line 5008: Let's not do it like this, but instead pop `"NVL"` off of `Oracle.Parser.FUNCTIONS`

georgesittas: Can you please double check redshift's / others' semantics while you're at it?

VaggelisD: I was only able to find existing test cases for Oracle and Redshift, I think most dialects only support `COALESCE` or IFNULL` but I'll do a pass

georgesittas:
> Came across Snowflake which supports `NVL()` (alias of `IFNULL`) _and_ `COALESCE()`. Their differences seem very subtle:
> 
>     * `NVL / IFNULL` only accept 2 arguments, `COALESCE` is varlen
> 
>     * The docs seem identical except maybe [the last bullet here](https://docs.snowflake.com/en/sql-reference/functions/nvl#usage-notes)
> 
> 
> [Snowflake NVL/IFNULL](https://docs.snowflake.com/en/sql-reference/functions/nvl#usage-notes) | [Snowflake COALESCE](https://docs.snowflake.com/en/sql-reference/functions/coalesce#usage-notes)

Yeah that makes sense, Oracle and Snowflake are very similar. Let's fix this for Snowflake too then I guess.

sqlglot/dialects/snowflake.py
```diff
@@ -317,6 +317,8 @@ class Parser(parser.Parser):
             "TO_VARCHAR": exp.ToChar.from_arg_list,
             "ZEROIFNULL": _build_if_from_zeroifnull,
         }
+        FUNCTIONS.pop("IFNULL")
```
georgesittas, line 320: does oracle also support this? should we do the same for it?

georgesittas (approved):

sqlglot/dialects/snowflake.py
VaggelisD: Can't find it in the docs

VaggelisD:
Reverted the Snowflake change, can't find a tangible difference between `NVL/NULLIF` and `COALESCE` except the varlen args.

PR #3736, opened 2024-07-03
fix(tsql): parse :: as type conversion then function

na399:
Without this fix, `x::varchar(max)` is parsed as a function with argument max but not Type.VARCHAR, unlike `x::varchar` that is parsed correctly.

This fix parses the type conversion first then function for scope resolution. 

Added tests for both parsing use cases.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -554,7 +554,7 @@ class TSQL(Dialect):
         }
 
         def _parse_dcolon(self) -> t.Optional[exp.Expression]:
-            return self._parse_function() or self._parse_types()
+            return self._parse_types() or self._parse_function()
 
         def _parse_options(self) -> t.Optional[t.List[exp.Expression]]:
             if not self._match(TokenType.OPTION):
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -1880,12 +1880,14 @@ FROM OPENJSON(@json) WITH (
     def test_scope_resolution_op(self):
         # we still want to support :: casting shorthand for tsql
         self.validate_identity("x::int", "CAST(x AS INTEGER)")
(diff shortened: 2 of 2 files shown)


tobymao:
@VaggelisD @georgesittas maybe we need to disable udt for tsql?

georgesittas:
> @VaggelisD @georgesittas maybe we need to disable udt for tsql?

T-SQL [supports UDTs](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-type-transact-sql?view=sql-server-ver16), so this doesn't look correct. Will have to think about this a bit more, because this PR produces an incorrect AST here:

```python
>>> import sqlglot
>>> sqlglot.parse_one("GEOGRAPHY::Point(latitude, longitude, 4236)", "tsql")
ScopeResolution(
  this=Column(
    this=Identifier(this=GEOGRAPHY, quoted=False)),
  expression=DataType(
    this=Type.USERDEFINED,
    kind=Point,
    expressions=[
      DataTypeParam(
        this=Var(this=LATITUDE)),
      DataTypeParam(
        this=Var(this=LONGITUDE)),
      DataTypeParam(
        this=Literal(this=4236, is_string=False))]))
```

georgesittas:
Maybe we can just look at the first token after `::` and if it's a _known_ type token then parse it as a data type, otherwise parse a scope resolution operator.

georgesittas:
@na399 I'll close this PR for now because it doesn't look right, will put up a fix once I investigate a bit more on what the best approach is. Appreciate the help regardless 👍

na399:
Thanks for looking into this, @georgesittas.

Well, the problem I encounter is when I run `sqlmesh format` on my project with default dialect as tsql, all my `CAST(x AS VARCHAR(MAX))` get converted to `x::VARCHAR(MAX)`. Upon running the formatted code on mssql, it can no longer handle `x::VARCHAR(MAX)`.

PR #3740, opened 2024-07-04
Fix(optimizer): only qualify coalesced USING columns if they exist in table schemas

georgesittas:


sqlglot/optimizer/qualify_columns.py
```diff
@@ -776,6 +785,10 @@ def all_columns(self) -> t.Set[str]:
 
     def get_source_columns(self, name: str, only_visible: bool = False) -> t.Sequence[str]:
         """Resolve the source columns for a given source `name`."""
+        source_columns = self._get_source_columns_cache.get((name, only_visible))
```
tobymao, line 788: i like the pattern of

if ... not in ...:
    self._source_columns = ...
return self._source_column[(name, only_visible)]

sqlglot/optimizer/qualify_columns.py
```diff
@@ -724,6 +732,7 @@ def __init__(self, scope: Scope, schema: Schema, infer_schema: bool = True):
         self._unambiguous_columns: t.Optional[t.Mapping[str, str]] = None
         self._all_columns: t.Optional[t.Set[str]] = None
         self._infer_schema = infer_schema
+        self._get_source_columns_cache: t.Dict[t.Tuple[str, bool], t.Sequence[str]] = {}
```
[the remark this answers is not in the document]
georgesittas: We already have `_source_columns` above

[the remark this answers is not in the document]
georgesittas: Perhaps we can consolidate the two, lemme take a quick look..

sqlglot/optimizer/qualify_columns.py
tobymao: yea

georgesittas: [Done](https://github.com/tobymao/sqlglot/pull/3740/commits/edcca349198e68b8c39980d09d30a3e60cd40c73)

tobymao (approved):

PR #3742, opened 2024-07-05
fix(optimizer): Fix expansion of SELECT * REPLACE, RENAME

VaggelisD:
Context from [public slack](https://app.slack.com/client/T041MTMK2JC/C044BRE5W4S)

In dialects like Snowflake the `SELECT` syntax allows the following statements (may be different per dialect):

```SQL
SELECT *
...
[ EXCLUDE {<col_name> [, ... ] }]
[ REPLACE {( <expr> AS <col_name> [, ... ] )}]
[ RENAME {<col_name> AS <col_alias>, ... )}]
```

Currently the optimizer incorrectly treats `SELECT * REPLACE`  as `SELECT * RENAME` semantically, with `RENAME` not being accounted for at all. 

This PR aims to fix this by:
- Adding support for `RENAME` which should substitute the column aliases 

- Fixing how `REPLACE` works i.e it now appends the replaced expressions directly to the selection list
```SQL
SELECT * REPLACE(1 as col) => SELECT 1 as col
```  

- The `EXCEPT`/`EXCLUDE` columns [are still not gathered ](https://github.com/tobymao/sqlglot/commit/7ee4fe73b29234f2837a212b6c872efd7f5c30ea)  in `scope.columns` _unless_ they participate in `REPLACE`. This is to solve qualification edge cases such as the following, in which we still want to keep `x.a` in the scope so it can be qualified inside of `COALESCE` before reaching `_expand_stars()`:

```SQL
SELECT * EXCLUDE(a) REPLACE(COALESCE(a, b) AS b) => SELECT COALESCE(x.a, x.b) AS b
```

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -574,8 +574,8 @@ def _expand_stars(
 
             table_id = id(table)
             columns_to_exclude = except_columns.get(table_id) or set()
-            rename_columns_dict = rename_columns.get(table_id, {})
-            replace_columns_dict = replace_columns.get(table_id, {})
+            renamed_columns = rename_columns.get(table_id, {})
+            replaced_columns = replace_columns.get(table_id, {})
 
             if pivot:
                 if pivot_output_columns and pivot_exclude_columns:
@@ -604,8 +604,8 @@ def _expand_stars(
                         alias(exp.func("coalesce", *coalesce_args), alias=name, copy=False)
                     )
                 else:
-                    selection_expr = replace_columns_dict.get(name) or exp.column(name, table=table)
-                    alias_ = rename_columns_dict.get(name, name)
+                    alias_ = renamed_columns.get(name, name)
+                    selection_expr = replaced_columns.get(name) or exp.column(name, table=table)
                     new_selections.append(
                         alias(selection_expr, alias_, copy=False)
                         if alias_ != name
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -505,7 +505,9 @@ def _expand_stars(
 
     new_selections = []
     except_columns: t.Dict[int, t.Set[str]] = {}
-    replace_columns: t.Dict[int, t.Dict[str, str]] = {}
+    replace_columns: t.Dict[int, t.Dict[str, exp.Alias]] = {}
+    rename_columns: t.Dict[int, t.Dict[str, str]] = {}
+
     coalesced_columns = set()
     dialect = resolver.schema.dialect
 
@@ -540,11 +542,13 @@ def _expand_stars(
             tables.extend(scope.selected_sources)
             _add_except_columns(expression, tables, except_columns)
             _add_replace_columns(expression, tables, replace_columns)
+            _add_rename_columns(expression, tables, rename_columns)
         elif expression.is_star:
             if not isinstance(expression, exp.Dot):
                 tables.append(expression.table)
                 _add_except_columns(expression.this, tables, except_columns)
                 _add_replace_columns(expression.this, tables, replace_columns)
+                _add_rename_columns(expression.this, tables, rename_columns)
             elif is_bigquery:
                 struct_fields = _expand_struct_stars(expression)
                 if struct_fields:
@@ -570,6 +574,8 @@ def _expand_stars(
 
             table_id = id(table)
             columns_to_exclude = except_columns.get(table_id) or set()
+            rename_columns_dict = rename_columns.get(table_id, {})
+            replace_columns_dict = replace_columns.get(table_id, {})
 
             if pivot:
                 if pivot_output_columns and pivot_exclude_columns:
@@ -598,10 +604,12 @@ def _expand_stars(
                         alias(exp.func("coalesce", *coalesce_args), alias=name, copy=False)
                     )
                 else:
-                    alias_ = replace_columns.get(table_id, {}).get(name, name)
-                    column = exp.column(name, table=table)
+                    selection_expr = replace_columns_dict.get(name) or exp.column(name, table=table)
+                    alias_ = rename_columns_dict.get(name, name)
                     new_selections.append(
-                        alias(column, alias_, copy=False) if alias_ != name else column
+                        alias(selection_expr, alias_, copy=False)
+                        if alias_ != name
+                        else selection_expr
                     )
 
     # Ensures we don't overwrite the initial selections with an empty list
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -570,6 +574,8 @@ def _expand_stars(
 
             table_id = id(table)
             columns_to_exclude = except_columns.get(table_id) or set()
+            rename_columns_dict = rename_columns.get(table_id, {})
+            replace_columns_dict = replace_columns.get(table_id, {})
```
georgesittas, line 578:
```suggestion
            renamed_columns = rename_columns.get(table_id, {})
            column_replacements = replace_columns.get(table_id, {})
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -623,15 +631,29 @@ def _add_except_columns(
...
-    expression: exp.Expression, tables, replace_columns: t.Dict[int, t.Dict[str, str]]
+    expression: exp.Expression, tables, replace_columns: t.Dict[int, t.Dict[str, exp.Alias]]
 ) -> None:
     replace = expression.args.get("replace")
 
     if not replace:
         return
 
-    columns = {e.this.name: e.alias for e in replace}
+    columns = {e.alias: e for e in replace}
```
georgesittas, line 656: Do we actually want an `Alias` here as the value? Should we just keep the expression and let `qualify_outputs` take care of the aliasing?

sqlglot/optimizer/scope.py
```diff
@@ -272,6 +272,20 @@ def columns(self):
...
+                    except_cols = ancestor.args.get("except") or []
+                    replace_cols = ancestor.args.get("replace") or []
+                    cols = set(
+                        replace_col
+                        for col in replace_cols
+                        for replace_col in col.find_all(exp.Column)
+                    )
+                    # Exclude the columns in EXCEPT() from the scope unless they're used in REPLACE, i.e in
+                    # "SELECT * EXCEPT (col1) REPLACE (COALESCE(1, col1) AS col2)"", col1 must be kept in scope
+                    is_except_col = column in except_cols and column not in cols
```
georgesittas, line 287: Is there a reason to exclude the `EXCEPT` columns from `scope.columns`, even if they're not used in a `REPLACE`?

georgesittas (commented):
Nice work 👍 

Can you combine all three of except, replace, rename? Have you taken that into account in your testing, e.g. how the optimizer handles them when all three interact?

VaggelisD:
> Can you combine all three of except, replace, rename? Have you taken that into account in your testing, e.g. how the optimizer handles them when all three interact?

Yep, you can combine all of them (BigQuery doesn't have `RENAME` so use Snowflake for sanity checks) and I did add such combinations in the tests, we can always include more of them

georgesittas:
> > Can you combine all three of except, replace, rename? Have you taken that into account in your testing, e.g. how the optimizer handles them when all three interact?
> 
> Yep, you can combine all of them (BigQuery doesn't have `RENAME` so use Snowflake for sanity checks) and I did add such combinations in the tests, we can always include more of them

Ah, apologies, skimmed over it too fast and missed the case 🤦 Nice.

sqlglot/optimizer/scope.py
VaggelisD: To match the semantics of these engines I think they should be removed from scope (e.g. you won't have stray columns in the `scope.columns` which could be a risk for `validate_qualify_columns`), and this was also added recently by us so I tried to work around it.

I don't have an example right now to prove this and I also didn't experiment with removing it so it may turn out fine, do you have any hunch for it?

georgesittas: Not sure how we match the semantics of these engines by not including the `EXCLUDE` columns in `scope.columns`, can you elaborate? I'd add them because that feels more intuitive to me & because it seems it'd simplify the logic. Also, it's weird that this clause stands out for no good reason, but I'm not sure about edge cases.

Wanna try this and see if it breaks stuff?

sqlglot/optimizer/qualify_columns.py
VaggelisD: Hmm, not sure if that's feasible or better. What I had in mind was that `_expand_stars()` wants to produce a qualified projection list, in which case if a column `name` is in `replaced_columns` then it was defined as `REPLACE(<expr> AS name, ...)` so you can append that `exp.Alias` directly in`new_selections` instead of re-building it later.

Also, a potential side effect of keeping just the `e.this` expression is that `qualify_outputs` depends on the `e.output_name`, so I think such scenarios can happen:

```
query: WITH x AS (SELECT 1 AS a) SELECT * REPLACE(2 AS a) FROM x

after expand_stars():  WITH x AS (SELECT 1 AS a) SELECT 2 FROM x AS x

after qualify_outputs():  WITH x AS (SELECT 1 AS a) SELECT 2 AS "2" FROM x AS x
```

What do you think?

sqlglot/optimizer/scope.py
VaggelisD: Take the following query in Snowflake:

```SQL
WITH x AS (SELECT 1 AS a) SELECT * EXCLUDE(a) RENAME(a AS d) FROM x;
```

This throws the error _`Column 'A' does not exist`_, so I think the equivalent logic in SQLGlot is that `a` is removed from `scope`, right?

However, I'm on board with removing the extra logic if it's not needed, will attempt it now

sqlglot/optimizer/qualify_columns.py
georgesittas: Ok sounds good, the `"2"` example makes sense.

georgesittas: Let's add a test for this btw

sqlglot/optimizer/scope.py
georgesittas: > so I think the equivalent logic in SQLGlot is that a is removed from scope, right?

Not necessarily, because statically you know there's a column `a` within `EXCLUDE`. This particular example throws an error because you exclude it and then try to rename the excluded column, but if you chose a different column in the `RENAME` clause it would work just fine.

VaggelisD: I tried to remove the logic in `scope.py` but a few tests broke, and I'm not sure if they're fixed easily or without replicating the same logic at different places. Two examples of what's incorrectly produced without it (before vs after `qualify_columns()`):

1. The columns inside `EXCEPT` are qualified, which doesn't seem to work in either SF or BQ:
```SQL
WITH x AS (SELECT 1 AS a) SELECT * EXCEPT(a) FROM x;
WITH x AS (SELECT 1 AS a) SELECT * EXCEPT(x.a) FROM x AS x;
```

2. Both instances of `a` here will be transformed to `COALESCE(x.a, y.a)` when in reality we'd expect only 1 instance:
```SQL
SELECT a, * EXCEPT (a) FROM x AS x LEFT JOIN x AS y USING (a);
SELECT
 COALESCE(x.a, y.a) AS a,
 COALESCE(x.a, y.a) AS a,
 x.b AS b,
 y.b AS b 
 FROM x AS x LEFT JOIN x AS y ON x.a = y.a;
```

```

 The common denominator here afaict is that we don't want to collect the columns inside of `EXCEPT(...)` because the rest of the qualification logic will find their instances in that list and will proceed with transforming/qualifying them when in reality they should be removed.

georgesittas (approved):

tobymao (approved):

PR #3745, opened 2024-07-06
Fix(duckdb, clickhouse)!: QUALIFY comes after WINDOW clause in queries

### LINKED ISSUE #3744: DuckDB: WINDOW and QUALIFY clauses get reversed
I am using sqlmesh with duckdb and running into an issue where optimized queries coming out of sqlglot are causing syntax errors once the query reaches the database engine. 

So the read and write dialect are both duckdb. 

A simplified form of the input query is:

```
select 
    *,
    lead(x) over w as next_x, 
    lag(x) over w as prev_x
from mytable
window w (partition by entity_id order by created_at)
qualify type = 'a'
```

And the resulting query that is actually passed to duckdb is:

```
SELECT
  *,
  LEAD("x") OVER "w" AS "next_x",
  LAG("x") OVER "w" AS "prev_x"
FROM "mytable" AS "mytable"
QUALIFY
  "type" = 'a'
WINDOW "w" AS (PARTITION BY "entity_id" ORDER BY "created_at")
```

This causes a syntax error because [QUALIFY comes after WINDOW](https://duckdb.org/docs/sql/query_syntax/qualify.html) in duckdb.  

As a workaround, I am omitting `WINDOW` for now and simply repeating the window definition in the select list for now.

georgesittas:
Fixes #3744

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -735,6 +735,13 @@ class BigQuery(Dialect):
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
         }
 
+        # WINDOW comes after QUALIFY
+        # https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#window_clause
+        AFTER_HAVING_MODIFIER_TRANSFORMS = {
+            "qualify": generator.Generator.AFTER_HAVING_MODIFIER_TRANSFORMS["qualify"],
+            "windows": generator.Generator.AFTER_HAVING_MODIFIER_TRANSFORMS["windows"],
+        }
+
         # from: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords
         RESERVED_KEYWORDS = {
             "all",
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import typing as t
 
-from sqlglot import exp, generator, parser, tokens, transforms
+from sqlglot import exp, generator, parser, tokens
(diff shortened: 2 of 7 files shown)


tobymao (approved):
are other dialects reversed? only duckdb is like this?

georgesittas:
> are other dialects reversed? only duckdb is like this?

I checked them, apparently CH [added support](https://clickhouse.com/docs/en/sql-reference/statements/select) for `QUALIFY` and so `eliminate_qualify` is not needed anymore. FYI @cpcloud, the time that transformation was [added](https://github.com/tobymao/sqlglot/commit/090724da9ed7a9c1c0b8024b68340a15b0b64ff0), there was no `QUALIFY` (see old docs [snapshot](https://web.archive.org/web/20231024155133/https://clickhouse.com/docs/en/sql-reference/statements/select)).


For this reason I refactored the base `Generator` mapping to have the order `WINDOW` -> `QUALIFY` which applies to both CH and DuckDB based on my findings, and overrode it in BigQuery to preserve the correct order:

Screenshot 2024-07-06 at 11 50 36 PM
Screenshot 2024-07-06 at 11 50 53 PM

The remaining dialects don't seem to support both `QUALIFY` and `WINDOW`.

tobymao:
snowflake is also window qualify right?

georgesittas:
> snowflake is also window qualify right?

Snowflake doesn't seem to support `WINDOW`. Checked both docs and tried this:

```sql
-- Syntax error: unexpected 'WINDOW'. (line 1)
SELECT schema_name, function_name FROM DUCKDB_FUNCTIONS t WINDOW my_window AS (PARTITION BY schema_name ORDER BY function_name) QUALIFY ROW_NUMBER() OVER my_window < 3
```

tobymao:
sounds good

Issue #3747, opened 2024-07-08
Trino's `DATE_TRUNC` is parsed as `TIMESTAMP_TRUNC` with the `unit` parsed as a `exp.Var`


maen-allaga:
When parsing a Trino's sql, DateTrunc is only parsed as DateTrunc if the input is explicitly casted as `DATE`, otherwise it's parsed as `TIMESTAMPTRUNC`.
However, Trino (and Presto) don't have a `timestamp_trunc` function, their `date_trunc` function can handle both date and timestamp types.

**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`: Done
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`: Done
- Check if the issue still exists on main: Done

**Fully reproducible code snippet**
```python
from sqlglot import parse

if __name__ == '__main__':
    sqlstr = """SELECT
                DATE_TRUNC('year', CAST(startDate AS DATE)) AS year_start,
                DATE_TRUNC('minute', CAST(startDate AS DATETIME)) AS minute_start
            FROM tmp"""
    expression_tree = parse(sqlstr, read="trino", dialect="trino")
    print(repr(expression_tree))
```

Output:
```
[Select(
  expressions=[
    Alias(
      this=DateTrunc(   # <------- Correct parsing of DateTrunc function
        unit=Literal(this=YEAR, is_string=True),   # <------- Unit parsed as Literal
        this=Cast(
          this=Column(
            this=Identifier(this=startDate, quoted=False)),
          to=DataType(this=Type.DATE, nested=False))),
      alias=Identifier(this=year_start, quoted=False)),
    Alias(
      this=TimestampTrunc(   # <------- Incorrect parsing of DateTrunc function as TimestampTrunc
        this=Cast(
          this=Column(
            this=Identifier(this=startDate, quoted=False)),
          to=DataType(this=Type.DATETIME, nested=False)),
        unit=Var(this=MINUTE)),   # <------- Unit parsed as Var
      alias=Identifier(this=minute_start, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=tmp, quoted=False))))]
```

**Official Documentation**
* Trino's date_trunc function: https://trino.io/docs/current/functions/datetime.html#date_trunc
* Presto's date_trunc function: https://prestodb.io/docs/current/functions/datetime.html#date_trunc

georgesittas:
[...]

PR #3748, opened 2024-07-08
Change the presto dialect to use date_trunc instead of timestamp_trunc

maen-allaga:
Address Issue #3747 .
Change presto dialect to always use `date_trunc` instead of the default dialect that uses `timestamp_trunc`.
Presto/Trino do not have a `timestamp_trunc` function, instead the `date_trunc` handle both `date` and `timestamp` types.


georgesittas:
As mentioned in the linked issue, the existing AST representation is intentional, so I'll go ahead and close this PR. Happy to continue discussing this in the other thread.

PR #3751, opened 2024-07-10
fix(bigquery)!: Canonicalize struct & array inline constructor

VaggelisD:
[Public slack context](https://app.slack.com/client/T041MTMK2JC/C0448SFS3PF)

BigQuery allows Array/Struct inline construction such as:

 ```SQL
SELECT 
    STRUCT<a INT, b STRING>(1, 'foo'),
    ARRAY<INT>[1, 2, 3], 
    ...
```

Up until now this was parsed as a single `exp.DataType` node with the inlined values stored in it's `values` arg. However, this is (1) semantically incorrect as `exp.DataType` should **only** encode type information and (2) it doesn't facilitate transpilation in any way.

This PR aims to decouple the actual values from `exp.DataType` and enable seamless transpilation (mainly focused towards DDB) by canonicalizing such expressions to `CAST(<values> AS <type>)`.  

This is done by transforming the parsed `exp.DataType` into an `exp.Cast`, with the values now stored in an `exp.Array` / `exp.Struct` as the cast's LHS and the type as the RHS.

### CODE REVIEW

File: sqlglot/dialects/duckdb.py [hunk 1]
```diff
@@ -116,17 +116,24 @@ def _build_make_timestamp(args: t.List) -> exp.Expression:
 
 def _struct_sql(self: DuckDB.Generator, expression: exp.Struct) -> str:
     args: t.List[str] = []
+
+    # BigQuery allows inline construction such as "STRUCT<a STRING, b INTEGER>('str', 1)" which is
+    # canonicalized to "ROW('str', 1) AS STRUCT(a TEXT, b INT)" in DuckDB
+    is_struct_cast = expression.find_ancestor(exp.Cast)
```
georgesittas, line 122: What triggered the need to make the change in `_struct_sql`? Why are we converting into `ROW`? Can't we just use the `{ ... }` syntax?

tests/dialects/test_bigquery.py
```diff
@@ -162,12 +161,9 @@ def test_bigquery(self):
         self.validate_identity("x <> ''")
         self.validate_identity("DATE_TRUNC(col, WEEK(MONDAY))")
         self.validate_identity("SELECT b'abc'")
-        self.validate_identity("""SELECT * FROM UNNEST(ARRAY<STRUCT<x INT64>>[])""")
         self.validate_identity("SELECT AS STRUCT 1 AS a, 2 AS b")
         self.validate_identity("SELECT DISTINCT AS STRUCT 1 AS a, 2 AS b")
         self.validate_identity("SELECT AS VALUE STRUCT(1 AS a, 2 AS b)")
-        self.validate_identity("SELECT STRUCT<ARRAY<STRING>>(['2023-01-17'])")
```
georgesittas, line 169: Why did this get removed?

georgesittas (approved):

georgesittas (approved):

PR #3753, opened 2024-07-10
feat(postgres): Support FROM ROWS FROM (...)

VaggelisD:
[Public slack context](https://app.slack.com/client/T041MTMK2JC/C044BRE5W4S)

This PR adds support for the `ROWS FROM` clause which is stored in `exp.Table`:
```
FROM ROWS FROM ( function_name ( [ argument [, ...] ] ) [ AS ( column_definition [, ...] ) ] [, ...] )
                [ WITH ORDINALITY ] [ [ AS ] alias [ ( column_alias [, ...] ) ]
```

Note that the parser (`_parse_table()`) already supports Postgres's `from_item` set functions, which are also parsed as `exp.Table`:

```
FROM function_name ( [ argument [, ...] ] )
                [ WITH ORDINALITY ] [ [ AS ] alias [ ( column_alias [, ...] ) ] ]
```

Thus, a straightforward solution was to recursively parse & generate `ROWS FROM`'s params as comma-separated set functions (i.e other `exp.Table`s).

Docs
----------
- [Postgres SELECT syntax](https://www.postgresql.org/docs/current/sql-select.html)

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3375,11 +3375,17 @@ class Parser(metaclass=_Parser):
         bracket = parse_bracket and self._parse_bracket(None)
         bracket = self.expression(exp.Table, this=bracket) if bracket else None
 
+        rows_from = self._match_text_seq("ROWS", "FROM") and self._parse_wrapped_csv(
+            self._parse_table
+        )
+        rows_from = self.expression(exp.Table, rows_from=rows_from) if rows_from else None
+
         only = self._match(TokenType.ONLY)
 
         this = t.cast(
             exp.Expression,
             bracket
+            or rows_from
             or self._parse_bracket(
                 self._parse_table_parts(schema=schema, is_db_reference=is_db_reference)
             ),
--- a/tests/dialects/test_postgres.py
+++ b/tests/dialects/test_postgres.py
@@ -100,9 +100,6 @@ class TestPostgres(Validator):
             "SELECT * FROM JSON_ARRAY_ELEMENTS('[1,true, [2,false]]') WITH ORDINALITY AS kv_json(a, b)"
         )
         self.validate_identity(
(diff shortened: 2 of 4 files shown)


tobymao (approved):

sqlglot/parser.py
```diff
@@ -3375,11 +3375,17 @@ def _parse_table(
         bracket = parse_bracket and self._parse_bracket(None)
         bracket = self.expression(exp.Table, this=bracket) if bracket else None
 
+        rows_from = self._match_text_seq("ROWS", "FROM") and self._parse_wrapped_csv(
+            self._parse_table
```
georgesittas, line 3379: Why are we parsing a table here?

tests/dialects/test_postgres.py
```diff
@@ -99,6 +99,9 @@ def test_postgres(self):
         self.validate_identity(
             "SELECT * FROM JSON_ARRAY_ELEMENTS('[1,true, [2,false]]') WITH ORDINALITY AS kv_json(a, b)"
         )
+        self.validate_identity(
+            """SELECT * FROM table, ROWS FROM (FUNC1(col1) AS alias1("col1" TEXT), FUNC2(col2) AS alias2("col2" INT)) WITH ORDINALITY AS alias3("col3" INT, "col4" TEXT)"""
+        )
```
georgesittas, line 104: This needs better test coverage, e.g. try `SELECT * FROM ROWS FROM (...)`. We can also have one without the ordinality.

georgesittas:
Looks good @VaggelisD, just a couple of comments.

sqlglot/parser.py
VaggelisD: Because `ROWS FROM` is basically a wrapped CSV collection of set/table functions which are currently parsed with `_parse_table`, so one can reuse it recursively. I believe it also semantically sound as `ROWS FROM` is supposed to gather the rows from many set/table functions in one table. 

Hope that made sense

georgesittas (approved):

PR #3756, opened 2024-07-11
feat(duckdb, spark, databricks): Support for ORDER BY ALL

### LINKED ISSUE #3755: Order By All keyword is not correctly parsed
Hello,

In duckdb the ALL keyword indicates that the output should be sorted by every column in order from left to right. The direction of this sort may be modified using either ORDER BY ALL ASC or ORDER BY ALL DESC and/or NULLS FIRST or NULLS LAST.

Using `sqlglot 25.5.1`, I was not able to parse correctly the "all" keyword.                                

```python 
import sqlglot

sql = """SELECT *
FROM test
ORDER BY ALL DESC;"""

parsed_sql = sqlglot.parse_one(
    sql=sql,
    read=sqlglot.Dialects.DUCKDB,
).sql(sqlglot.Dialects.DUCKDB)

print(parsed_sql)
```
```
>>> SELECT * FROM test ORDER BY "ALL" DESC
```

The output query is not correct and running it will produce an error.
The correct output should be:
```
>>> SELECT * FROM test ORDER BY ALL DESC
```

**Official Documentation**
https://duckdb.org/docs/sql/query_syntax/orderby.html

VaggelisD:
Fixes #3755

For the dialects that support `ORDER BY ALL` such as DuckDB, Databricks, Spark3 (tested on 3.4+) parse `ALL` as an `exp.Var` and not as a column.

Docs
--------
[DuckDB ORDER BY](https://duckdb.org/docs/sql/query_syntax/orderby.html) | [Databricks ORDER BY](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-orderby.html)

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -90,6 +90,8 @@ def _dateadd_sql(self: Spark.Generator, expression: exp.TsOrDsAdd | exp.Timestam
 
 
 class Spark(Spark2):
+    SUPPORTS_ORDER_BY_ALL = True
+
     class Tokenizer(Spark2.Tokenizer):
         STRING_ESCAPES_ALLOWED_IN_RAW_STRINGS = False
 
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -356,6 +356,11 @@ class Dialect(metaclass=_Dialect):
     EXPAND_ALIAS_REFS_EARLY_ONLY_IN_GROUP_BY = False
     """Whether alias reference expansion before qualification should only happen for the GROUP BY clause."""
 
+    SUPPORTS_ORDER_BY_ALL = False
+    """
+    Whether ORDER BY ALL is supported (expands to all the selected columns) as in DuckDB, Spark3/Databricks
+    """
+
     # --- Autofilled ---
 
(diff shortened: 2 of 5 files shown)


tobymao:
could we make this instead a group by property? this should make it easier to transpile,

sqlglot/dialects/spark.py
```diff
@@ -90,6 +90,8 @@ def _dateadd_sql(self: Spark.Generator, expression: exp.TsOrDsAdd | exp.Timestam
 
 
 class Spark(Spark2):
+    SUPPORTS_ORDER_BY_ALL = True
```
georgesittas, line 93: Is this only available in Spark3?

georgesittas (approved):

VaggelisD:
@tobymao  Sure, I'll do some research on `GROUP BY ALL` and follow up with a new PR

sqlglot/dialects/spark.py
VaggelisD: Yep, I tested Spark 2 and it didn't work; On the other hand, both Spark 3.4 and 3.5.1 had support for it

georgesittas: Sounds good

PR #3757, opened 2024-07-11
feat: Support for RPAD & LPAD functions

VaggelisD:
Introduce `exp.Pad` which supports the string padding functions `LPAD()`, `RPAD()`; The joint node is used because these 2 functions share the same signature but differ on the result (left/right application of padding)

Most dialects default to whitespace for the padding (3rd parameter) while DuckDB, MySQL, Presto/Trino, and Hive/Spark2  require it explicitly, so the flag `PAD_FILL_PATTERN_IS_REQUIRED` was introduced to facilitate transpilation during generation.

Docs
--------
[Databricks / Spark3](https://docs.databricks.com/en/sql/language-manual/functions/lpad.html) | [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#lpad) | [DuckDB](https://duckdb.org/docs/sql/functions/char.html#lpadstring-count-character) | [Postgres](https://www.postgresql.org/docs/9.1/functions-string.html) | [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_LPAD.html) | [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/lpad) | [MySQL](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_lpad) | [Hive / Spark2](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF) | [Presto](https://prestodb.io/docs/current/functions/string.html#lpad-string-size-padstring-varchar) | [Trino](https://trino.io/docs/current/functions/string.html#lpad)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4795,7 +4795,7 @@ class List(Func):
 
 # String pad, kind True -> LPAD, False -> RPAD
 class Pad(Func):
-    arg_types = {"kind": True, "this": True, "expression": True, "fill_pattern": False}
+    arg_types = {"this": True, "expression": True, "fill_pattern": False, "is_left": True}
 
 
 # https://docs.snowflake.com/en/sql-reference/functions/to_char
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4793,6 +4793,11 @@ class List(Func):
     is_var_len_args = True
 
 
+# String pad, kind "L" -> LPAD, "R" -> RPAD
+class Pad(Func):
+    arg_types = {"kind": True, "this": True, "expression": True, "fill_pattern": False}
+
+
 # https://docs.snowflake.com/en/sql-reference/functions/to_char
 # https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/TO_CHAR-number.html
 class ToChar(Func):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4793,7 +4793,7 @@ class List(Func):
     is_var_len_args = True
 
 
-# String pad, kind "L" -> LPAD, "R" -> RPAD
+# String pad, kind True -> LPAD, False -> RPAD
 class Pad(Func):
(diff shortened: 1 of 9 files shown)


sqlglot/expressions.py
```diff
@@ -4793,6 +4793,11 @@ class List(Func):
     is_var_len_args = True
 
 
+# String pad, kind "L" -> LPAD, "R" -> RPAD
```
georgesittas, line 4796: Let's not do it this way, just use a boolean flag if there are only two variants

georgesittas (commented):
Is there no `PAD` function that does both L/R-padding? I feel like some DBs have it. This PR would break it if it does exist.

VaggelisD:
> Is there no PAD function that does both L/R-padding? I feel like some DBs have it. This PR would break it if it does exist.

I'll do another pass over all dialects to search for it; If it does exist, I'll try to include it in `kind`, otherwise I'll refactor to a boolean flag

VaggelisD:
Did another pass, I couldn't find `PAD()` in any dialect. For this reason, I also refactored the `kind` arg into a boolean

sqlglot/expressions.py
```diff
@@ -4793,6 +4793,11 @@ class List(Func):
     is_var_len_args = True
 
 
+# String pad, kind True -> LPAD, False -> RPAD
+class Pad(Func):
+    arg_types = {"kind": True, "this": True, "expression": True, "fill_pattern": False}
+
```
georgesittas, line 4799: Since we're using a bool flag I'd suggest renaming `kind` to something more descriptive, e.g. `lpad`, `is_lpad`, `left` are some ideas.

georgesittas (approved):
One suggestion re: arg name, LGTM otherwise

sqlglot/expressions.py
tobymao: kind should probably be one of the farther arguments since by default it would show up first

tobymao (approved):

PR #3761, opened 2024-07-12
fix!: Remove number matching from COLON placeholder parser

### LINKED ISSUE #3759: Duckdb slicing notation is parsed as placeholder
Hello,

Duckdb supports slicing notation without providing a begin index, e.g. `arr[:2]`. However, using `sqlglot 25.5.1` this is being parsed as a placeholder.

```
import sqlglot
q = "SELECT arr[:2] AS elements FROM array_table;"
p = sqlglot.parse_one(q, dialect="duckdb")

print(repr(p))
print()
print(p.sql(dialect="duckdb"))
```
```
Select(
  expressions=[
    Alias(
      this=Bracket(
        this=Column(
          this=Identifier(this=arr, quoted=False, _type=DataType(this=Type.UNKNOWN)),
          _type=DataType(this=Type.UNKNOWN)),
        expressions=[
          Placeholder(this=2, _type=DataType(this=Type.UNKNOWN))],
        returns_list_for_maps=True),
      alias=Identifier(this=elements, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=array_table, quoted=False))))

SELECT arr[$2] AS elements FROM array_table
```
Expected `arr[:2]` to be parsed as slice notation. Conversion back to SQL should result in the original query.

**Official Documentation**
https://duckdb.org/docs/sql/functions/nested.html#slicing

VaggelisD:
Fixes #3759 

Currently `TokenType.COLON` is parsed into an `exp.Placeholder` if it matches the syntax `:<id_var> | :<number>`. Since `_parse_placeholder()` has a higher precedence than `_parse_slice()` (lower on the call stack) any `arr[:n]` expression would be parsed as `arr[<placeholder>]` instead of `arr[<slice>]`.

Afaict the idvar-based syntax (e.g. `:foo`) is used for Oracles & Snowflakes bind variables, but I couldn't find any examples (or existing sqlglot tests) which showcase number-based bind variables such as `:1`. This is in contrast to actual positional parameters such as DuckDB's `$1` syntax, which must match integer literals.

If we want to preserve the `COLON` as it currently stands, the fallback solution is to `pop()` it from `PLACEHOLDER_PARSERS` for any dialect that supports the slicing notation e.g. at least DuckDB and Postgres from a quick search.

Reference: [PLACEHOLDER_PARSERS PR](https://github.com/tobymao/sqlglot/commit/6038c72a120a26e3d8acf2791ddc36bcfc0b74d8), [exp.Slice PR](https://github.com/tobymao/sqlglot/commit/dfa863f472cf86918d5d2daeaa06cf0fdd743a0b) 

Docs
-------
[Snowflake Bind variables](https://docs.snowflake.com/en/sql-reference/bind-variables) | [Oracle Bind variables](https://docs.oracle.com/cd/A84050_01/NT816CLI/DOC/sqlplus.816/a75664/ch34.htm) | [DuckDB parameters](https://duckdb.org/docs/sql/query_syntax/prepared_statements.html) |  [Postgres slice notation](https://www.postgresql.org/docs/current/arrays.html#ARRAYS-ACCESSING) | [DuckDB slice notation](https://duckdb.org/docs/sql/data_types/array.html#retrieving-values-from-arrays)


georgesittas (approved):
Let's mark this as breaking

PR #3766, opened 2024-07-15
fix(clickhouse): Allow TokenType.SELECT as a Tuple field identifier

### LINKED ISSUE #3763: ClickHouse tuple field named `select` fails to parse without quotes
```
In [7]: import sqlglot as sg, sqlglot.expressions as sge

In [8]: sg.__version__
Out[8]: '25.5.1'

In [9]: sg.parse_one('Tuple(select Int64)', into=sge.DataType, dialect="clickhouse")
```

Unfortunately I don't have control over the input here, this is coming directly from a clickhouse `DESCRIBE t` query that returns the string `Tuple(select Int64)`, without quoting the `select` field.

VaggelisD:
Fixes #3763

Docs
-------
[Clickhouse Tuple](https://clickhouse.com/docs/en/sql-reference/data-types/tuple)

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -332,6 +332,8 @@ class ClickHouse(Dialect):
             TokenType.SET,
         }
 
+        RESERVED_TOKENS = {*parser.Parser.RESERVED_TOKENS} - {TokenType.SELECT}
+
         AGG_FUNC_MAPPING = (
             lambda functions, suffixes: {
                 f"{f}{sfx}": (f, sfx) for sfx in (suffixes + [""]) for f in functions
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -332,7 +332,8 @@ class ClickHouse(Dialect):
             TokenType.SET,
         }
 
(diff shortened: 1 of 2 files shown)


georgesittas (approved):

sqlglot/dialects/clickhouse.py
```diff
@@ -332,6 +332,8 @@ class Parser(parser.Parser):
             TokenType.SET,
         }
 
+        RESERVED_TOKENS = {*parser.Parser.RESERVED_TOKENS} - {TokenType.SELECT}
```
tobymao, line 335: you don't need to make a copy just to RESERVED_TOKENS - ...

VaggelisD: Good catch, thanks

sqlglot/dialects/clickhouse.py
```diff
@@ -332,6 +332,9 @@ class Parser(parser.Parser):
             TokenType.SET,
         }
 
+        RESERVED_TOKENS = parser.Parser.RESERVED_TOKENS.copy()
+        RESERVED_TOKENS.remove(TokenType.SELECT)
```
georgesittas, line 336: This is what Toby suggested:
```suggestion
        RESERVED_TOKENS = parser.Parser.RESERVED_TOKENS - {TokenType.SELECT}
```

PR #3786, opened 2024-07-19
feat: Move ANNOTATORS to Dialect for dialect-aware annotation

### LINKED ISSUE #3778: Incorrect type cast in conversion from Spark to Presto in timestampad function
I have the following line in my SELECT statement, (consider the second argument in the `timestampadd` Spark function)...

```
q = "select region_uuid, 
    timestampadd(minute, floor(extract(minute from delivery_created_time_local)/30)*30, date_trunc(‘hour’, time_org)) as time 
    from timeseries"
sqlglot.transpile(q, read='spark', write='presto')
```
The Presto transpiled result is

```
SELECT region_uuid, 
    DATE_ADD('MINUTE', FLOOR(CAST(EXTRACT(MINUTE FROM delivery_created_time_local) AS DOUBLE) / NULLIF(30, 0)) * 30, DATE_TRUNC('‘HOUR’', time_org)) AS time 
FROM timeseries
```

However, DATE_ADD requires the second argument to be BIGINT, not DOUBLE, as it's currently being cast.

VaggelisD:
Fixes #3778

The issue arises from the following chain reaction:
1. Transpiled divisions from Spark to Presto/Trino will be casted to `DOUBLE` due to typed division semantics
2. Certain math functions in Presto/Trino such as `FLOOR()` will return the same type as their input (`DOUBLE` in this case), so that type will bubble up
3. `DATE_ADD` requires an `INT` on the 2nd argument; While this was accounted for in #2648, what changes here is that `annotate_types` will always (incorrectly) annotate these math functions to `INT`, rendering the `_to_int()` a no-op essentially (`exp.cast()` will be optimized away)
4. Thus, the expression / 2nd parameter remains a `DOUBLE` which causes the execution error

This PR attempts to solve this long term by making the `TypeAnnotator` dialect-aware; By moving the `ANNOTATORS` to `Dialect`, each dialect can tailor the annotation logic to it's local needs.

Docs
--------
[Presto math functions](https://prestodb.io/docs/current/functions/math.html) | [Trino math functions](https://trino.io/docs/current/functions/math.html)

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -262,11 +262,13 @@ class Presto(Dialect):
         **Dialect.ANNOTATORS,
         exp.Floor: lambda self, e: self._annotate_by_args(e, "this"),
         exp.Ceil: lambda self, e: self._annotate_by_args(e, "this"),
-        exp.Mod: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Mod: lambda self, e: self._annotate_by_args(e, "this", "expression"),
         exp.Round: lambda self, e: self._annotate_by_args(e, "this"),
         exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
         exp.Abs: lambda self, e: self._annotate_by_args(e, "this"),
-        exp.Rand: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Rand: lambda self, e: self._annotate_by_args(e, "this")
+        if e.this
+        else self._set_type(e, exp.DataType.Type.DOUBLE),
     }
 
     class Tokenizer(tokens.Tokenizer):
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -204,11 +204,11 @@ def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str
     return f"{this}{expr}"
 
 
-def _to_int(expression: exp.Expression) -> exp.Expression:
+def _to_int(self: Presto.Generator, expression: exp.Expression) -> exp.Expression:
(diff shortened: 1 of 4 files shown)


tobymao (approved):

sqlglot/dialects/presto.py
```diff
@@ -256,6 +256,19 @@ class Presto(Dialect):
...
+    # equal to the input type e.g: FLOOR(5.5/2) -> DECIMAL, FLOOR(5/2) -> BIGINT
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        exp.Floor: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Ceil: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Mod: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Round: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Abs: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Rand: lambda self, e: self._annotate_by_args(e, "this"),
```
georgesittas, line 269: Not sure about this one. Are we generating `RANDOM` or `RAND` for `exp.Rand`? See [docs](https://prestodb.io/docs/current/functions/math.html#rand):

Screenshot 2024-07-21 at 5 51 25 PM

georgesittas (approved):

sqlglot/dialects/presto.py
VaggelisD: We're generating both as `RAND()` currently; Will add `Type.DOUBLE` annotation as a fallback to the no arg case

PR #3788, opened 2024-07-19
Start making the Generator non-recursive

MatMoore:
Hello, this is my first stab at making the `Generator` non-recursive, as discussed in slack.

Currently the sql method relies heavily on recursion.  Migrating to an iterative implementation will prevent the generator from hitting python's recursion limit when passed deeply nested ASTs.

## What's changed

This commit refactors the `sql` method to support the migration from recursive to iterative. The remaining work can now be chunked by `Expression` subtype.

This relies on a stack to traverse the AST depth-first. This stack emulates the call stack when running the recursive implementation. The `sql` method contains a loop that pops items off the stack, and stack items can contain either nodes to be processed or generated SQL strings (similar to the implementation in `set_operations`.)

I've defined also defined a callback protocol `ExpressionHandler`, and introduced a `ITERATIVE_TRANSFORMS` mapping so we can map individual expression subtypes to new iterative functions one at a time. The existing mechanisms for handling  different types of expressions work as before.
 
I've started using this new structure for set operations - these are already partially iterative, but still contain some recursive calls.

This is my first time contributing so please let me know if there is anything I've forgotten or the approach isn't clear 🙂 

## Rough plan
I'm aiming to make fairly mechanical changes so that I don't diverge too far from the existing logic.

I noticed that the generator class isn't included in the benchmark, so I was thinking of adding a new benchmark next, to make sure that these changes aren't adding overhead (let me know if this done before merging this PR.)

After that, `set_operations` can be rewritten to yield `StackItem`s instead of recursively calling the `sql` method.

`TRANSFORMS` can be migrated to `ITERATIVE_TRANSFORMS`.

We can then turn each  `_sql` method into an `ExpressionHandler`. The process will be something like this:
1. Start with the recursive implementation
2. Change the signature to match `ExpressionHandler`
4. For each recursive call, yield a node
5. Split the output of the return expression, and yield all extra substrings instead of returning
6. Adjust the order of the yields so that rightmost parts of the expression are yielded before leftmost ones (so that the leftmost ones end up on the top of the stack)
7. Add the function to `ITERATIVE_TRANSFORMS`

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -71,6 +71,17 @@ class Generator(metaclass=_Generator):
         comments: Whether to preserve comments in the output SQL code.
             Default: True
     """
+    StackItem = t.Union[t.Tuple[t.Union[str, exp.Expression], None], t.Optional[str], bool]
+    Stack = t.List[StackItem]
+    ExpressionHandler = t.Callable[..., t.Generator[StackItem, None, None]]
+
+
+    ITERATIVE_TRANSFORMS: t.Dict[t.Type[exp.Expression], t.Callable[[Generator], ExpressionHandler]] = {
+        exp.Union: lambda self: self._set_operations_iter,
+        exp.Except: lambda self: self._set_operations_iter,
+        exp.Intersect: lambda self: self._set_operations_iter
+    }
+
 
     TRANSFORMS: t.Dict[t.Type[exp.Expression], t.Callable[..., str]] = {
         **JSON_PATH_PART_TRANSFORMS,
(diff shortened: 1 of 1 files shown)


MatMoore:
This is still in draft as I need to remove some syntax only supported in later python versions.

tobymao:
hey @MatMoore i really appreciate the effort, but i'm not sure i'm willing to accept this, it seems like a large change for something that we've never really run into.

there are some cases like set ops which reach stack depth, but not much else does.

what issues have you faced specifically?

MatMoore:
> hey @MatMoore i really appreciate the effort, but i'm not sure i'm willing to accept this, it seems like a large change for something that we've never really run into.
> 
> there are some cases like set ops which reach stack depth, but not much else does.
> 
> what issues have you faced specifically?

I've encountered it with binary operations, specifically concatenation of a lot of fields. [Here's an artificial example that replicates this](https://gist.github.com/MatMoore/620775b07b0cbeeb80c46928843eee2c). The query this was derived from was concatenating a bunch of expressions which themselves contained nested function calls.

I encountered this via [Datahub](https://datahubproject.io/), another open source project which ingests DBT metadata (among other things) into a data catalogue. This uses sqlglot to format the SQL before ingesting it into the catalogue or building column-level lineage.

Btw this is the slack thread: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1721131621971919

tobymao:
can we just add this for binary ?

MatMoore:
> can we just add this for binary ?

I think it is possible, as demonstrated by the `set_operations` implementation, but it would only remove some of the recursion, not eliminate it completely, because it still needs to recurse on each operand in order to handle other expression types.

I think if further expression types need to be adapted to this style, it could become quite messy as the stack and the boiler plate around it will need to be repeated if we don't move it into `sql`. But it depends if you prefer to limit the iterative approach to set operations and binaries vs a larger rewrite.

Let me know what you think. If this is the wrong direction then I'd be happy to close this PR and raise another one that just changes the `binary` method.

tobymao:
i think for now, i'd rather just limit this to all binaries if it's possible, i haven't looked deeply into the code. can we generalize the connector stack method and do it for all binaries?

i feel like if we do it for set ops and all binaries, that should cover most cases

PR #3792, opened 2024-07-21
Fix(mysql): don't wrap VALUES clause

### LINKED ISSUE #3789: INSERT incorrectly roundtripped in mysql dialect.
When parsing an INSERT query with VALUES and re-serializing it to a string - the VALUES expression is improperly wrapped in parentheses - which is not valid SQL.

```python

import sqlglot

query = """
INSERT INTO things 
  (
    a 
  , b
  )
VALUES (1, 2)
AS new_data
ON DUPLICATE KEY UPDATE
  id = LAST_INSERT_ID(id)
  , a = new_data.a
  , b = new_data.b
"""

sqlglot.parse_one(query, dialect="mysql").sql(dialect="mysql")
# 'INSERT INTO things (a, b) (VALUES (1, 2)) AS new_data ON DUPLICATE KEY UPDATE id = LAST_INSERT_ID(id), a = new_data.a, b = new_data.b'

```

MySQL rejects the stringified expression has having a syntax error - indeed removing the parentheses makes the query valid.

georgesittas:
Fixes #3789

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -691,6 +691,7 @@ class MySQL(Dialect):
         SUPPORTS_TO_NUMBER = False
         PARSE_JSON_NAME = None
         PAD_FILL_PATTERN_IS_REQUIRED = True
+        WRAP_DERIVED_VALUES = False
 
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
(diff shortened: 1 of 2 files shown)


georgesittas:
mypy upgraded to 1.11.0 and caused a few issues I'm not sure how to resolve yet, feel free to take this to the finish line if I don't do it soon.

tobymao (approved):

Issue #1699, opened 2023-05-28
snowflake object_construct is converted to duckdb map with values of single type


tekumara:
In Snowflake, [OBJECT_CONSTRUCT](https://docs.snowflake.com/en/sql-reference/functions/object_construct) can be used to construct an object from the given key-value pairs. Keys must be strings, but values can be any type and can differ from each other.

When converting OBJECT_CONSTRUCT to duckdb, it ends up as a [map](https://duckdb.org/docs/sql/data_types/map). Maps must have a single type for all keys, and a **single** type for **all** values.

eg:

```python
sqlglot.parse_one("SELECT OBJECT_CONSTRUCT('a',1,'b','BBBB', 'c',null)", read="snowflake").sql(dialect="duckdb")

"SELECT MAP(LIST_VALUE('a', 'b', 'c'), LIST_VALUE(1, 'BBBB', NULL))"
```

Because the map/list must be of the same type, in this example the integer `1` becomes the string `"1"` when the expression is executed in duckdb:

```
D SELECT to_json(MAP(LIST_VALUE('a', 'b', 'c'), LIST_VALUE(1, 'BBBB', NULL)));
┌──────────────────────────────────────────────────────────────────────┐
│ to_json(map(list_value('a', 'b', 'c'), list_value(1, 'BBBB', NULL))) │
│                                 json                                 │
├──────────────────────────────────────────────────────────────────────┤
│ {"a":"1","b":"BBBB","c":null}                                        │
└──────────────────────────────────────────────────────────────────────┘
```
NB: `to_json` is used here to show that the value for the`a` key is a string.

pkit:
Yup, struct should behave as needed, I think. As `struct` is the only type (besides `list`) that can contain other structs.
And in Snowflake `OBJECT` can contain other objects too.

GeorgeSittas:
Ok, thanks for confirming. I'll look into this a bit closer and try to get a fix on Monday.

PR #3807, opened 2024-07-24
feat(snowflake): Support for OBJECT_INSERT, transpile to DDB

### LINKED ISSUE #3802: Error transpiling OBJECT_CONSTRUCT from Snowflake to DuckDB
Thanks for creating this amazing library! It's very useful for my use-case where I transpile Snowflake queries to DuckDB. I found this bug while migrating some scripts. Not very familiar with the codebase but I can look into it if you provide some insights.

```python
sqlglot.parse_one('select 
     OBJECT_INSERT(OBJECT_INSERT(OBJECT_INSERT(OBJECT_CONSTRUCT(), 'key1', 'value1'), 'key2', 'value2'), 'key3', 'value3')',
 dialect='snowflake').sql(dialect='duckdb')
```

Outputs:

```sql
select Object_insert(Object_insert(Object_insert({}, 'key1', 'value1'), 'key2',
                     'value2'), 'key3', 'value3')     
```

And it fails with the following error:
CleanShot 2024-07-24 at 02 51 18@2x

VaggelisD:
Fixes #3802

Currently, `OBJECT_CONSTRUCT` will transpile to `STRUCT` construction in DuckDB (context https://github.com/tobymao/sqlglot/issues/1699). 

Following this convention, this PR adds support for Snowflake's `OBJECT_INSERT(object, key, value, [update_flag])` which will get transpiled to DuckDB's `STRUCT_INSERT(object, key := value)`. 

Docs
-----------
[Snowflake OBJECT_INSERT](https://docs.snowflake.com/en/sql-reference/functions/object_insert) | [Snowflake OBJECT_CONSTRUCT](https://docs.snowflake.com/en/sql-reference/functions/object_construct) | [DuckDB STRUCT](https://duckdb.org/docs/sql/data_types/struct.html)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -290,6 +290,7 @@ class Snowflake(Dialect):
             ),
             "NULLIFZERO": _build_if_from_nullifzero,
             "OBJECT_CONSTRUCT": _build_object_construct,
+            "OBJECT_INSERT": exp.ObjectInsert.from_arg_list,
             "REGEXP_REPLACE": _build_regexp_replace,
             "REGEXP_SUBSTR": exp.RegexpExtract.from_arg_list,
             "RLIKE": exp.RegexpLike.from_arg_list,
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -290,7 +290,6 @@ class Snowflake(Dialect):
             ),
             "NULLIFZERO": _build_if_from_nullifzero,
             "OBJECT_CONSTRUCT": _build_object_construct,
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -290,6 +290,7 @@ class Parser(parser.Parser):
             ),
             "NULLIFZERO": _build_if_from_nullifzero,
             "OBJECT_CONSTRUCT": _build_object_construct,
+            "OBJECT_INSERT": exp.ObjectInsert.from_arg_list,
```
tobymao, line 293: why is this needed if we have an ObjectInsert expression? shouldn't it auto parse?

VaggelisD: Right, will remove it.

PR #3856, opened 2024-07-31
feat(duckdb): Support for exp.TimeDiff generation

VaggelisD:
Generate BQ's `exp.TimeDiff` into DuckDB's `DATE_DIFF()`. Although both functions have similar signatures and should be calculating the partitions between `end_time - start_time`, BQ seems to inverse the operands which is more accurately documented on 3rd party pages.

Docs
-------------
[BigQuery TIME_DIFF official docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/time_functions#time_diff) | [BigQuery TIME_DIFF unofficial docs](https://count.co/sql-resources/bigquery-standard-sql/time_diff) | [DuckDB DATE_DIFF](https://duckdb.org/docs/sql/functions/timestamp.html#date_diffpart-startdate-enddate)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -32,6 +32,7 @@ from sqlglot.dialects.dialect import (
     timestamptrunc_sql,
     timestrtotime_sql,
     unit_to_var,
+    unit_to_str,
 )
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
Do we support duckdb -> bq?

VaggelisD:
Not correctly yet, and I think that'd require types to work properly; Afaict DuckDB's `DATE_DIFF` works for all sorts of types (TIMESTAMP, TIME, DATE) while BQ has a different function for each type such as `DATE_DIFF`,  `TIME_DIFF`, `DATETIME_DIFF` and so on. 

Maybe a primal attempt would be to transpile only the casted literals from DDB to the appropriate BQ function with respect to the `to` type.

georgesittas:
Ah, makes sense.

> Maybe a primal attempt would be to transpile only the casted literals from DDB to the appropriate BQ function with respect to the to type.

Let's defer for a future PR if we end up actually needing this transformation.

PR #3857, opened 2024-07-31
fix(duckdb): Add implicit casts to DATE_DIFF

VaggelisD:
BigQuery can implicitly cast/coerce string literal arguments into the appropriate date/time type, which is not the case for DuckDB.

This PR aims to fix the transpilation of `exp.DateDiff` by inserting the casts to replicate this behavior.

Docs
------------
[BigQuery DATE_DIFF](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_diff) | [DuckDB DATE_DIFF](https://duckdb.org/docs/sql/functions/date.html#date_diffpart-startdate-enddate)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -32,6 +32,7 @@ from sqlglot.dialects.dialect import (
     timestamptrunc_sql,
     timestrtotime_sql,
     unit_to_var,
+    unit_to_str,
 )
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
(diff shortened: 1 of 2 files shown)


georgesittas:
This looks incorrect @VaggelisD, you can't just cast to `DATE` in duckdb because `DATETIME` values are accepted:

```
D select date_diff('minute', DATETIME '1992-09-15 12:05:03', DATETIME '1992-09-15 12:08:03');
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ date_diff('minute', CAST('1992-09-15 12:05:03' AS TIMESTAMP), CAST('1992-09-15 12:08:03' AS TIMESTAMP)) │
│                                                  int64                                                  │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│                                                                                                       3 │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

PR #3864, opened 2024-08-01
fix!: Generation of exp.ArrayConcat for 2-arg based dialects

VaggelisD:
While `exp.ArrayConcat` is a varlen function in many dialects, this isn't the case for all; This PR fixes this by refactoring the generation to _reduce_ the varlen version to a series of 2-arg based `ARRAY_CONCAT(x, y)` calls for those dialects.

This PR also fixes Redshift's `ARRAY(expr1, ..., exprN)` construction which differs from Postgres's `ARRAY[expr1, ..., exprN]`. 

Docs
---------
[Trino - Presto ARRAY Functions](https://trino.io/docs/current/functions/array.html)
[Postgres ARRAY Functions](https://www.postgresql.org/docs/9.1/functions-array.html)
 [Spark CONCAT](https://spark.apache.org/docs/latest/api/sql/#concat) | [Databricks CONCAT](https://docs.databricks.com/en/sql/language-manual/functions/concat.html) 
 [Redshift ARRAY](https://docs.aws.amazon.com/redshift/latest/dg/r_array.html) | [Redshift ARRAY_CONCAT](https://docs.aws.amazon.com/redshift/latest/dg/r_array_concat.html) 
 [Snowflake ARRAY_CAT](https://docs.snowflake.com/en/sql-reference/functions/array_cat) 
 [BigQuery ARRAY_CONCAT](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#array_concat) 
[DuckDB ARRAY_CONCAT](https://duckdb.org/docs/sql/functions/nested.html#list_concatlist1-list2)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -488,12 +488,7 @@ class Postgres(Dialect):
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
             exp.AnyValue: any_value_to_max_sql,
-            exp.Array: lambda self, e: (
-                f"{self.normalize_func('ARRAY')}({self.sql(e.expressions[0])})"
-                if isinstance(seq_get(e.expressions, 0), exp.Select)
-                else f"{self.normalize_func('ARRAY')}[{self.expressions(e, flat=True)}]"
-            ),
-            exp.ArrayConcat: lambda self, e: super().arrayconcat_sql(e, name="ARRAY_CAT"),
+            exp.ArrayConcat: lambda self, e: self.arrayconcat_sql(e, name="ARRAY_CAT"),
             exp.ArrayContainsAll: lambda self, e: self.binary(e, "@>"),
             exp.ArrayOverlaps: lambda self, e: self.binary(e, "&&"),
             exp.ArrayFilter: filter_array_using_unnest,
@@ -648,3 +643,11 @@ class Postgres(Dialect):
                 return self.sql(this)
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
+
+        def array_sql(self, expression: exp.Array) -> str:
+            exprs = expression.expressions
+            return (
+                f"{self.normalize_func('ARRAY')}({self.sql(exprs[0])})"
+                if isinstance(seq_get(exprs, 0), exp.Select)
+                else f"{self.normalize_func('ARRAY')}[{self.expressions(expression, flat=True)}]"
+            )
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -466,6 +466,7 @@ class Postgres(Dialect):
         MULTI_ARG_DISTINCT = False
         CAN_IMPLEMENT_ARRAY_ANY = True
         COPY_HAS_INTO_KEYWORD = False
+        ARRAY_CONCAT_IS_VAR_LEN = False
 
         SUPPORTED_JSON_PATH_PARTS = {
             exp.JSONPathKey,
@@ -492,7 +493,6 @@ class Postgres(Dialect):
                 if isinstance(seq_get(e.expressions, 0), exp.Select)
                 else f"{self.normalize_func('ARRAY')}[{self.expressions(e, flat=True)}]"
             ),
-            exp.ArrayConcat: rename_func("ARRAY_CAT"),
             exp.ArrayContainsAll: lambda self, e: self.binary(e, "@>"),
             exp.ArrayOverlaps: lambda self, e: self.binary(e, "&&"),
             exp.ArrayFilter: filter_array_using_unnest,
@@ -647,3 +647,6 @@ class Postgres(Dialect):
                 return self.sql(this)
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
+
+        def arrayconcat_sql(self, expression: exp.ArrayConcat, name: str = "ARRAY_CAT") -> str:
+            return super().arrayconcat_sql(expression, name=name)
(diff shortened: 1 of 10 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -169,6 +170,7 @@ class Generator(Postgres.Generator):
 
         TRANSFORMS = {
             **Postgres.Generator.TRANSFORMS,
+            exp.Array: rename_func("ARRAY"),
```
georgesittas, line 173: I recall noticing this issue in the past, but IIRC redshift supports both `ARRAY(...)` and `ARRAY[...]` and they mean different things. So there's some ambiguity here where you lose either syntax if you keep a single representation, mapping's not 1-1. Can you double check this?

sqlglot/dialects/postgres.py
```diff
@@ -647,3 +647,6 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
                 return self.sql(this)
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
+
+        def arrayconcat_sql(self, expression: exp.ArrayConcat, name: str = "ARRAY_CAT") -> str:
+            return super().arrayconcat_sql(expression, name=name)
```
georgesittas, line 652: It's not good practice to change the method signature -- let's instead pass in `"ARRAY_CAT"` in the `super` call.

georgesittas: We could also keep the transform entry and do `self.arrayconcat_sql(expression, name=...` instead of overriding the method to use `super`. I think I'd go with this instead.

sqlglot/dialects/redshift.py
VaggelisD: Hmm, I couldn't get `ARRAY_CONCAT` to work with the bracket version and I can't seem to find it documented anywhere yet, I'll keep searching

VaggelisD: This is interesting, I've been searching the Redshift [data types](https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html), I can't find a native array type except `SUPER`:

```
SUPER
A superset data type that encompasses all scalar types of Amazon Redshift including 
complex types such as ARRAY and STRUCTS.
```


And while array functions don't work with the bracket version, it seems that queries like `SELECT ARRAY[1, 2]` can still go through, notice the different symbols though:

```SQL
SELECT ARRAY[1, 2, 1]
{1, 2, 1}

SELECT ARRAY(1, 2, 1)
[1, 2, 1]

SELECT ARRAY[1, 2, 1] = ARRAY(1, 2, 1);
ERROR: operator does not exist: integer[] = super

```

Lastly, this is from the AWS Postgres <-> Redshift [data type docs](https://docs.aws.amazon.com/redshift/latest/dg/federated-data-types.html):

```
The following RDS PostgreSQL and Aurora PostgreSQL data types are converted to VARCHAR(64K) in Amazon Redshift:

- JSON, JSONB
- Arrays
- ...
```

It might be that the bracket version is backwards-compatible Postgres and denotes a specific type of array, while the Redshift way is through `ARRAY()` which creates a `SUPER` type.

sqlglot/dialects/postgres.py
```diff
@@ -492,7 +493,7 @@ class Generator(generator.Generator):
                 if isinstance(seq_get(e.expressions, 0), exp.Select)
                 else f"{self.normalize_func('ARRAY')}[{self.expressions(e, flat=True)}]"
             ),
-            exp.ArrayConcat: rename_func("ARRAY_CAT"),
+            exp.ArrayConcat: lambda self, e: super().arrayconcat_sql(e, name="ARRAY_CAT"),
```
georgesittas, line 496: Should this be `self.arrayconcat_sql`? There's no override in postgres right? Same for the other similar entries.

sqlglot/dialects/redshift.py
georgesittas: In any case, we need to preserve `ARRAY[...]` and `ARRAY(...)` in redshift, right? This change is definitely breaking because it will produce the function syntax whereas right now we produce the square bracket one:

```python
>>> import sqlglot
>>> sqlglot.transpile("array(1,2,3)", "redshift")
['ARRAY[1, 2, 3]']
>>> sqlglot.transpile("array[1,2,3]", "redshift")
['ARRAY[1, 2, 3]']
```

sqlglot/dialects/postgres.py
VaggelisD: Yeah the overrides are gone now so `self <-> super` here, will revert

sqlglot/dialects/redshift.py
VaggelisD: I think `ARRAY()` should be preserved here then, as this seems the preferred Redshift way according to docs, relevant SO threads etc. I didn't manage to find any source earlier showcasing the bracket version being used. Wdyt?

georgesittas: I assume you mean we should represent `ARRAY(...)` using `exp.Array(...)` in redshift. If so, that's fine. What I'm saying is that we need to roundtrip both forms and that the current behavior is problematic because it changes the semantics. If you change `exp.Array` to generate `ARRAY(...)`, then this is gonna be a breaking change because as pointed above the current behavior is to generate the bracket syntax.

VaggelisD:
My latest attempt does the following:
- From any dialect -> Redshift, `exp.Array` is now generated in the function format `ARRAY()`. This restores the functionality for array functions as mentioned
- In Redshift both array constructors are roundtripped, i.e `SELECT ARRAY(1, 2, 3), ARRAY[1, 2, 3]` -> `SELECT ARRAY(1, 2, 3), ARRAY[1, 2, 3] `

georgesittas (approved):
LGTM, thanks for addressing the redshift issue 🎉

PR #3865, opened 2024-08-02
feat(duckdb)!: Support for BQ's exp.GenerateDateArray generation

VaggelisD:
Generate BigQuery's `GENERATE_DATE_ARRAY(start, stop, interval)` to DuckDB's `GENERATE_SERIES(start, stop, interval)`

Docs
--------
[BigQuery GENERATE_DATE_ARRAY](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_date_array) | [DuckDB GENERATE_SERIES](https://duckdb.org/docs/sql/functions/nested.html#generate_series)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -209,10 +209,11 @@ def _arrow_json_extract_sql(self: DuckDB.Generator, expression: JSON_EXTRACT_TYP
     return arrow_sql
 
 
-def _date_diff_sql(self: DuckDB.Generator, expression: exp.DateDiff) -> str:
-    def _implicit_date_cast(arg: exp.Expression):
-        return exp.cast(arg, exp.DataType.Type.DATE) if isinstance(arg, exp.Literal) else arg
+def _implicit_date_cast(arg: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
+    return exp.cast(arg, exp.DataType.Type.DATE) if isinstance(arg, exp.Literal) else arg
 
+
+def _date_diff_sql(self: DuckDB.Generator, expression: exp.DateDiff) -> str:
     this = _implicit_date_cast(expression.this)
     expr = _implicit_date_cast(expression.expression)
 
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -852,3 +853,18 @@ def objectinsert_sql(self, expression: exp.ObjectInsert) -> str:
...
             return self.func("STRUCT_INSERT", this, kv_sql)
+
+        def generatedatearray_sql(self, expression: exp.GenerateDateArray) -> str:
+            start = _implicit_date_cast(expression.args.get("start"))
+            end = _implicit_date_cast(expression.args.get("end"))
+
+            # BigQuery has a default step/interval of 1 DAY
+            interval = expression.args.get("interval") or exp.Interval(
+                this=exp.Literal.number(1), unit=exp.var("DAY")
+            )
```
georgesittas, line 864: Is this only a BQ concept? Should we be doing this AST normalization @ parse time for BQ?

georgesittas (commented):
LGTM, just one comment

sqlglot/dialects/duckdb.py
VaggelisD: Yeah afaict this function is only in BQ, although it was introduced here https://github.com/tobymao/sqlglot/pull/3169/files without further context or BQ tests.

The parsing is now performed implicitly but we can definitely add an entry to `FUNCTIONS` and normalize it there as well

georgesittas: Sure we can do this normalization by adding a `Parser.FUNCTIONS` entry.

tests/dialects/test_bigquery.py
```diff
@@ -1419,7 +1419,7 @@ def test_bigquery(self):
             "SELECT GENERATE_DATE_ARRAY('2016-10-05', '2016-10-08')",
             write={
                 "duckdb": "SELECT CAST(GENERATE_SERIES(CAST('2016-10-05' AS DATE), CAST('2016-10-08' AS DATE), INTERVAL 1 DAY) AS DATE[])",
-                "bigquery": "SELECT GENERATE_DATE_ARRAY('2016-10-05', '2016-10-08')",
+                "bigquery": "SELECT GENERATE_DATE_ARRAY('2016-10-05', '2016-10-08', INTERVAL 1 DAY)",
```
georgesittas, line 1422: I think this actually makes it a breaking change, because the AST & its SQL output are different, let's make sure to mark as such before merging.

georgesittas (approved):
LGTM

PR #3867, opened 2024-08-02
Feat(snowflake): add support for a a couple of missing clauses in PIVOT clause

georgesittas:
Reference: https://docs.snowflake.com/en/sql-reference/constructs/pivot#syntax

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4537,7 +4537,7 @@ class PivotAlias(Alias):
     pass
 
 
-# Represent's Snowflakes ANY [ ORDER BY ... ] syntax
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -4536,6 +4537,12 @@ class PivotAlias(Alias):
     pass
 
 
+# Represent's Snowflakes ANY [ ORDER BY ... ] syntax
```
georgesittas, line 4540:
```suggestion
# Represents Snowflake's ANY [ ORDER BY ... ] syntax
```

tobymao (approved):

PR #3870, opened 2024-08-05
fix: Fixed size array parsing

### LINKED ISSUE #3868: Bug: Cast using array element (Snowflake)
sqlglot 25.8.1 produces invalid sql:
```
>>> sqlglot.parse_one("SELECT ARRAY_CONSTRUCT('foo')::VARIANT[0] as variant_key",dialect="snowflake").sql("snowflake")
"SELECT CAST(['foo'] AS ARRAY(VARIANT)[0]) AS variant_key"
```
This sql errors on Snowflake with:
```
ProgrammingError: 001003 (42000): SQL compilation error:
syntax error line 1 at position 37 unexpected '['.
syntax error line 1 at position 40 unexpected ')'.
```
Because the `[0]` is inside the bracket.

sqlglot 25.5.1 works:
```
>>> sqlglot.parse_one("SELECT ARRAY_CONSTRUCT('foo')::VARIANT[0] as variant_key",dialect="snowflake").sql("snowflake")
"SELECT CAST(['foo'] AS VARIANT)[0] AS variant_key"
```

VaggelisD:
Fixes #3868

The recent commit https://github.com/tobymao/sqlglot/commit/2ffb07070952cde7ac9a1883cbf9b4c477c55abb introduced support for fixed-size array parsing in DuckDB; However, not all dialects [1] support fixed-size arrays so a regression was introduced in which the subscript/bracket operator was now interpreted as the array size.

Also, this commit was not working properly for higher dimension arrays so valid queries were broken after roundtripping:

```
sqlglot.parse_one("SELECT [[1], [2], [3]]::INT[1][3]", dialect="duckdb").sql("duckdb")
SELECT CAST([[1], [2], [3]] AS INT[1])[1][3]
```

This PR fixes both issues by introducing a dialect flag that enables the recent addition only for DuckDB. From a quick search, the other dialects which come close to supporting/parsing fixed-size arrays are Postgres & Redshift but neither engine actually enforces that, meaning that queries like the following are valid Postgres/Redshift:

```SQL
SELECT ARRAY[1, 2, 3, 4]::INT[1][1][1];
```

For this reason, this fix is limited to DuckDB for now and if need arises for Postgres/Redshift they can easily override that by setting the new flag.

Docs
-----------
- [DuckDB Arrays](https://duckdb.org/docs/sql/data_types/array.html)
- [[1] Snowflake Array limitations](https://docs.snowflake.com/en/sql-reference/data-types-semistructured#characteristics-of-an-array), 2nd bullet: _Snowflake does not currently support fixed-size arrays_

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -375,6 +375,12 @@ class Dialect(metaclass=_Dialect):
     as the former is of type INT[] vs the latter which is SUPER
     """
 
+    SUPPORTS_FIXED_SIZE_ARRAYS = False
+    """
+    Whether expressions such as x::INT[5] should be parsed as fixed-size array defs/casts e.g. in DuckDB. In
+    dialects which don't support fixed size arrays such as Snowflake, this should be interpreted as a subscript/index operator
+    """
+
     # --- Autofilled ---
 
     tokenizer_class = Tokenizer
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -227,6 +227,7 @@ class DuckDB(Dialect):
     INDEX_OFFSET = 1
     CONCAT_COALESCE = True
     SUPPORTS_ORDER_BY_ALL = True
+    SUPPORTS_FIXED_SIZE_ARRAYS = True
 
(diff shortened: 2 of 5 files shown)


georgesittas (approved):
cc @tobymao can you take a look here as well?

tobymao (approved):

PR #3873, opened 2024-08-05
feat(parser)!: Support ALTER VIEW AS SELECT

xiaohui-sun:
Add a support to "ALTER VIEW AS SELECT"

`ALTER VIEW view_identifier AS select_statement`

More documentation on it:
https://spark.apache.org/docs/3.0.0-preview2/sql-ref-syntax-ddl-alter-view.html#syntax-3


sqlglot/expressions.py
```diff
@@ -4282,6 +4282,13 @@ class AlterTable(Expression):
     }
 
 
+class AlterView(Expression):
+    arg_types = {
+        "this": True,
+        "expression": False,
+    }
```
georgesittas, line 4289: Not sure if this is how we wanna implement this. For `CREATE ...` statements we have a single `Create` node which contains a `kind` argument that stores `TABLE`, `VIEW` etc. Maybe we want to do the same for `Alter` statements.

sqlglot/parser.py
```diff
@@ -6485,12 +6485,7 @@ def _parse_alter_table_set(self) -> exp.AlterSet:
 
         return alter_set
 
-    def _parse_alter(self) -> exp.AlterTable | exp.Command:
-        start = self._prev
-
-        if not self._match(TokenType.TABLE):
-            return self._parse_as_command(start)
-
+    def _parse_alter_table(self, start) -> exp.AlterTable | exp.Command:
```
georgesittas, line 6488: Maybe we can instead make a set of tokens, `ALTERABLES`, starting with `TABLE` and `VIEW` and then do:

```python
if not self._match_set(self.ALTERABLES):
...
```

This way we'll avoid adding the `start` argument everywhere.

sqlglot/parser.py
```diff
@@ -6517,6 +6512,30 @@ def _parse_alter(self) -> exp.AlterTable | exp.Command:
 
         return self._parse_as_command(start)
 
+    def _parse_alter_view(self, start) -> exp.AlterView | exp.Command:
+        self._parse_select()
```
georgesittas, line 6516: This needs to be cleaned up

sqlglot/parser.py
```diff
@@ -6517,6 +6512,30 @@ def _parse_alter(self) -> exp.AlterTable | exp.Command:
...
+        table_parts = self._parse_table_parts()
+        this = self._parse_schema(this=table_parts)
```
georgesittas, line 6518: I think we can consolidate these two, imitating the current implementation for `ALTER TABLE`:

```suggestion
        this = self._parse_table(schema=True)
```

sqlglot/parser.py
```diff
@@ -6517,6 +6512,30 @@ def _parse_alter(self) -> exp.AlterTable | exp.Command:
...
+
+    def _parse_alter(self) -> exp.AlterTable | exp.AlterView | exp.Command:
+        start = self._prev
+
+        if self._match(TokenType.TABLE):
+            return self._parse_alter_table(start)
+        elif self._match(TokenType.VIEW):
+            return self._parse_alter_view(start)
+        else:
+            return self._parse_as_command(start)
```
georgesittas, line 6537:
```suggestion
        if self._match(TokenType.VIEW):
            return self._parse_alter_view(start)
        return self._parse_as_command(start)
```

georgesittas (commented):
Thanks for the PR @xiaohui-sun 👍

Please run `make style` to fix the styling CI errors.

sqlglot/expressions.py
xiaohui-sun: Thanks for reviewing!
Yes that makes sense. Let me add that.

sqlglot/parser.py
xiaohui-sun: Sounds great.

sqlglot/parser.py
xiaohui-sun: Updated

sqlglot/parser.py
xiaohui-sun: You are right.

sqlglot/generator.py
```diff
@@ -3126,6 +3126,8 @@ def altertable_sql(self, expression: exp.AlterTable) -> str:
             actions = self.expressions(expression, key="actions", prefix="ADD COLUMNS ")
         elif isinstance(actions[0], exp.Delete):
             actions = self.expressions(expression, key="actions", flat=True)
+        elif isinstance(actions[0], exp.Select):
+            actions = "AS " + self.expressions(expression, key="actions", flat=True)
```
tobymao, line 3130: why is this flat=True?

xiaohui-sun: So there is no need to flat since there is only one action?

tobymao: if it's flat, it may not be pretty printed

xiaohui-sun: Removed. Although I am not very clear how this controls the output. Thanks.

xiaohui-sun: got it. yes it will only generate one line with flat=True.

tests/test_transpile.py
```diff
@@ -845,7 +845,6 @@ def test_command_identity(self):
             "ALTER TABLE table1 RENAME COLUMN c1 TO c2, c2 TO c3",
             "ALTER TABLE table1 RENAME COLUMN c1 c2",
             "ALTER TYPE electronic_mail RENAME TO email",
-            "ALTER VIEW foo ALTER COLUMN bla SET DEFAULT 'NOT SET'",
```
georgesittas, line 848: We shouldn't get rid of this test. Let's make sure this is still supported (either parses into an alter expression or into a command), otherwise this PR will introduce a regression.

tests/test_parser.py
```diff
@@ -581,8 +581,14 @@ def test_create_table_error(self, logger):
...
-            "ALTER TABLE foo RENAME TO bar",
+            parse_one("ALTER TABLE table1 RENAME COLUMN IF EXISTS c1 TO c2").sql(),
+            "ALTER TABLE table1 RENAME COLUMN IF EXISTS c1 TO c2",
+        )
+
+    def test_alter_view_as(self):
+        self.assertEqual(
+            parse_one("ALTER VIEW foo AS SELECT a, b, c FROM bar").sql(),
+            "ALTER VIEW foo AS SELECT a, b, c FROM bar",
         )
```
georgesittas, line 592: I'd suggest moving these two tests to `identity.sql`, not sure why `test_rename_table` was in `test_parser.py` to begin with.

sqlglot/parser.py
```diff
@@ -6485,10 +6491,14 @@ def _parse_alter_table_set(self) -> exp.AlterSet:
...
 
+    def _parse_alter_view_as(self) -> t.Optional[exp.Expression]:
+        return self._parse_select(self._prev)
+
     def _parse_alter(self) -> exp.AlterTable | exp.Command:
         start = self._prev
 
-        if not self._match(TokenType.TABLE):
+        create_token = self._match_set(self.ALTERABLES) and self._prev
+        if not create_token:
```
georgesittas, line 6501: I think this needs to be changed to `alter_token` or something, there's no `CREATE` in this context.

sqlglot/parser.py
```diff
@@ -6485,10 +6491,14 @@ def _parse_alter_table_set(self) -> exp.AlterSet:
 
         return alter_set
 
+    def _parse_alter_view_as(self) -> t.Optional[exp.Expression]:
+        return self._parse_select(self._prev)
```
georgesittas, line 6495: Why are we passing `self._prev` to `_parse_select` here? We can also just move this inline within the `ALTER_PARSERS` dict, like we do for `"DELETE"`.

sqlglot/expressions.py
```diff
@@ -4274,6 +4274,7 @@ class Rollback(Expression):
 class AlterTable(Expression):
     arg_types = {
         "this": True,
+        "kind": True,
```
georgesittas, line 4277: Since this is now expecting a `kind` argument, so I'd rename the expression to `Alter` because `AlterTable` is a misnomer when the `kind` is not `"TABLE"`.

@tobymao FYI this is gonna be a breaking change, e.g. every `AlterTable` in SQLMesh is affected.

sqlglot/generator.py
```diff
@@ -3135,7 +3137,8 @@ def altertable_sql(self, expression: exp.AlterTable) -> str:
         only = " ONLY" if expression.args.get("only") else ""
         options = self.expressions(expression, key="options")
         options = f", {options}" if options else ""
-        return f"ALTER TABLE{exists}{only} {self.sql(expression, 'this')}{on_cluster} {actions}{options}"
+        kind = expression.args.get("kind")
```
georgesittas, line 3140:
```suggestion
        kind = self.sql(expression, "kind")
```

georgesittas (commented):
Hey @xiaohui-sun, did another pass on the PR. Let's please increase the test coverage for this, e.g. at least add the examples in Spark's [docs](https://spark.apache.org/docs/3.0.0-preview2/sql-ref-syntax-ddl-alter-view.html#syntax-3).

Ideally we need to investigate which dialects support `ALTER VIEW` and also add tests for them and ensure this PR doesn't break stuff.

The `ALTER ...` statement family is an important one and this is gonna be an important breaking change, so let's ensure we don't break stuff.

cc @VaggelisD to also take a look at this & other dialects if you have cycles.

VaggelisD:
I did a pass over other dialects and it seems that `ALTER VIEW AS` is also supported by [MySQL](https://dev.mysql.com/doc/refman/8.4/en/alter-view.html) and [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-view-transact-sql?view=sql-server-ver16).

Both of these dialects allow options before & after the `AS <select>` clause so if we don't want to increase the PR scope, we must at least make sure that they're parsed gracefully as Commands

tests/test_transpile.py
xiaohui-sun: This test is for [unsupported syntax](https://github.com/tobymao/sqlglot/pull/3873/files/08706b9b4bab7fb35e520dcb732a70ab6100501a#diff-ddc45c8cfc37c82c0e36c52bcd13d8de90b866af0a3150bdefc157729db6d63dR867). I will move it to another test.

tests/test_parser.py
xiaohui-sun: Sure. Will add more tests.

xiaohui-sun: moved.

sqlglot/parser.py
xiaohui-sun: fixed

sqlglot/parser.py
xiaohui-sun: renamed

sqlglot/generator.py
xiaohui-sun: updated

xiaohui-sun:
> I did a pass over other dialects and it seems that `ALTER VIEW AS` is also supported by [MySQL](https://dev.mysql.com/doc/refman/8.4/en/alter-view.html) and [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-view-transact-sql?view=sql-server-ver16).
> 
> Both of these dialects allow options before & after the `AS <select>` clause so if they're not currently supported, we must at least make sure that they're parsed gracefully as Commands

I will add more tests there.

tests/test_transpile.py
xiaohui-sun: Actually I can't find any SQL dialect supports "ALTER VIEW ... ALTER COLUMN ..."
Only T-SQL supports something [similar](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-view-transact-sql?view=sql-server-ver16#syntax) but requires two statements like below:
```
ALTER VIEW view_name
ALTER COLUMN column_name datatype [NULL | NOT NULL];

```

teja71296:
How can we give the input of a sql query to multiple lines as its not transpiling the query.

Below input which is not working.

python sqlglot/__main__.py "CREATE VIEW customer_order_data \n
AS \n
select cd.id as customer_id, \n
CONCAT(cd.last_name, cd.first_name) AS full_name \n
god.first_order_date, \n
god.most_recent_order_date, \n
coalesce(god.number_of_orders,0) as number_of_orders \n
from customer_data cd \n
left join grouped_order_data god on cd.id = god.customer_id;" --read tsql --write snowflake


If we give the input as below its working but why?

python sqlglot/__main__.py "CREATE VIEW customer_order_data AS select cd.id as customer_id, CONCAT(cd.last_name, cd.first_name) AS full_name, god.first_order_date, gd.most_recent_order_date, coalesce(god.number_of_orders,0) as number_of_orders from customer_data cd left join grouped_order_data god on cd.id = god.customer_id;" --read tsql --write snowflake

And also need help how to translate Stored procedure, from sql to Snowflake - Need help

sqlglot/expressions.py
xiaohui-sun: I can rename to `Alter`, and that would touch lots of places since `AlterTable` is used widely.

tests/dialects/test_spark.py
```diff
@@ -129,6 +129,36 @@ def test_ddl(self):
                 "spark": "ALTER TABLE StudentInfo DROP COLUMNS (LastName, DOB)",
             },
         )
+        self.validate_all(
+            "ALTER VIEW StudentInfoView AS SELECT * FROM StudentInfo",
```
georgesittas, line 133: Let's make all of these `validate_identity`.

georgesittas (commented):
@xiaohui-sun thanks for addressing the comments! Seems like we cover several examples from the dialects that support `ALTER VIEW`, so the PR looks good. 👍

One note: it seems like there are multiple `ALTER VIEW` statements that are not fully parsed and we fallback to `Command` nodes. I noticed this in the [CI terminal output](https://github.com/tobymao/sqlglot/actions/runs/10274153786/job/28430237500?pr=3873), where you can see multiple warnings logged. Can you track these cases down and set `check_command_warning=True` in the `validate_identity` calls? We shouldn't have these warnings in the tests.

> I can rename to Alter, and that would touch lots of places since AlterTable is used widely. ([context](https://github.com/tobymao/sqlglot/pull/3873/files#r1705318614))

I think I would probably commit to this since we're adding this now anyway, so that it's implemented cleanly. I'll defer to @tobymao on this decision. TL;DR, I see 2 options:

- Rename `AlterTable` to `Alter` so that the name doesn't assume we're altering the `Table`. Since we're making a breaking change here with `kind`, this seems ok to me.
- Keep `AlterTable` and make `kind` default to `TABLE` so that this is backwards-compatible. Two downsides to this: 1) `ALTER VIEW` will be parsed into `AlterTable`, which is not a good name, 2) it's inconsistent with `exp.Create` which doesn't assume the `CREATABLE`'s kind.

xiaohui-sun:
> we can rename

Ack. Let me rename.

tests/dialects/test_spark.py
xiaohui-sun: fixed

xiaohui-sun:
> @xiaohui-sun thanks for addressing the comments! Seems like we cover several examples from the dialects that support `ALTER VIEW`, so the PR looks good. 👍
> 
> One note: it seems like there are multiple `ALTER VIEW` statements that are not fully parsed and we fallback to `Command` nodes. I noticed this in the [CI terminal output](https://github.com/tobymao/sqlglot/actions/runs/10274153786/job/28430237500?pr=3873), where you can see multiple warnings logged. Can you track these cases down and set `check_command_warning=True` in the `validate_identity` calls? We shouldn't have these warnings in the tests.
> 
> > I can rename to Alter, and that would touch lots of places since AlterTable is used widely. ([context](https://github.com/tobymao/sqlglot/pull/3873/files#r1705318614))
> 
> I think I would probably commit to this now so that it's implemented cleanly in one pass, but I'll defer to @tobymao on this decision. TL;DR, I see 2 options:
> 
> 1. Rename `AlterTable` to `Alter` so that the name doesn't assume we're altering the `Table`. Since we're making a breaking change here with `kind`, this seems ok to me.
> 2. Keep `AlterTable` and make `kind` default to `TABLE` so that this is backwards-compatible. Two downsides to this: 1) `ALTER VIEW` will be parsed into `AlterTable`, which is not a good name, 2) it's inconsistent with `exp.Create` which doesn't assume the `CREATABLE`'s kind.

@georgesittas Thanks for your suggestion. I have changed `AlterTable` to `Alter`, since it is a breaking change anyway.

sqlglot/expressions.py
xiaohui-sun: Renamed.

sqlglot/parser.py
```diff
@@ -1011,6 +1016,7 @@ class Parser(metaclass=_Parser):
         "DROP": lambda self: self._parse_alter_table_drop(),
         "RENAME": lambda self: self._parse_alter_table_rename(),
         "SET": lambda self: self._parse_alter_table_set(),
+        "AS": lambda self: self._parse_select(),
```
georgesittas, line 1019:
```suggestion
        "AS": lambda self: self._parse_ddl_select(),
```

georgesittas (approved):
Thanks for the quick turnaround @xiaohui-sun - appreciate the patience!

This looks good to me, the only outstanding item is silencing the warnings. If it's a low-lift, feel free to take a stab at actually expanding the syntax support, but if it's not trivial I wouldn't worry about it.

xiaohui-sun:
> Thanks for the quick turnaround @xiaohui-sun - appreciate the patience!
> 
> This looks good to me, the only outstanding item is silencing the warnings. If it's a low-lift, feel free to take a stab at actually expanding the syntax support, but if it's not trivial I wouldn't worry about it.

No problem. I will take a look at the warnings and add as much support as I can.

sqlglot/parser.py
xiaohui-sun: updated

xiaohui-sun:
I made an attempt to fix the test cases for `ALTER ... SET TABLE PROPERTIES`.
The change is more complex than I initially expected, but I'd like to get some feedback before I dive too far into it.
@georgesittas 
1. Changed `self._parse_wrapped_csv(self._parse_assignment))` to `self._parse_wrapped(self._parse_properties, optional=True)` ([link](https://github.com/tobymao/sqlglot/pull/3873/files#diff-63eb8dd82d3561bc414e3e13cf59516bd6584c5766a10072157a8a5aee4ad85fR6515)) 
2. Set the property prefix for multiple dialects.
3. Also reduced some test cases that cause warning

sqlglot/dialects/trino.py
```diff
@@ -30,5 +36,19 @@ class Generator(Presto.Generator):
             exp.JSONPathSubscript,
         }
 
+        def with_properties(self, properties: exp.Properties) -> str:
+            return self.properties(properties, wrapped=False, prefix="PROPERTIES", sep=", ")
```
georgesittas, line 40: This doesn't look right, I think it breaks valid DDL statements that contain the `WITH (...)` syntax. See https://trino.io/docs/current/sql/create-table.html.

georgesittas (commented):
Hey @xiaohui-sun, I took a quick look at the latest changes (`SET` etc) and I feel like we should avoid doing it in this PR after all, in order to avoid scope creep.

Some of the changes seem a bit sketchy, e.g. I commented on one related to the "WITH" properties in Trino. The `Property` generation override is also weird. I'm worried this isn't the right way to implement the remaining syntax and these changes may break existing DDLs.

Can you revert to the previous commit that I'd approved? It's fine if some of the `ALTER`s are commands for now, just make sure to silence the warnings using the flag I mentioned in the `validate_identity` tests. Should be good to merge after that.

sqlglot/dialects/trino.py
xiaohui-sun: ok let me revert.

xiaohui-sun: @georgesittas Reverted and added `check_command_warning=True`. Please take a look. Thanks

georgesittas: Thanks!

georgesittas:
Thanks for the contribution!

PR #3878, opened 2024-08-06
feat(databricks): Add support for STREAMING tables

### LINKED ISSUE #3876: sqlglot does not support databricks SQL streaming syntax
sqlglot doesn't properly parse the following Databricks SQL statement:
`CREATE OR REFRESH STREAMING LIVE TABLE some_table AS SELECT * FROM some_source`

**Official Documentation**
[Docs](https://docs.databricks.com/en/tables/streaming.html)

VaggelisD:
Fixes #3876

This PR adds support for Databrick's `CREATE STREAMING` tables; Also, `STREAM <read_func>` is added as an alternative select path since streaming tables seem to be created through CTAS + STREAM.

The `LIVE` property is omitted for now as I couldn't find relevant documentation. 

Docs
---------
[Databricks CREATE STREAMING TABLE](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-streaming-table.html)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5659,7 +5659,7 @@ class ScopeResolution(Expression):
 
 
 class Stream(Expression):
-    arg_types = {"this": True}
+    pass
 
 
 class StarMap(Func):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1387,6 +1387,7 @@ class Create(DDL):
         "exists": False,
         "properties": False,
         "replace": False,
+        "refresh": False,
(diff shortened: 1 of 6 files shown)


georgesittas (approved):

VaggelisD:
I went ahead and simplified this further by removing `TokenType.STREAM`, it doesn't have to be defined for this PR.

ericvergnaud:
@VaggelisD you can find additional documentation here: https://docs.databricks.com/en/delta-live-tables/sql-ref.html

PR #3887, opened 2024-08-07
fix: Fix COLLATE's RHS parsing

### LINKED ISSUE #3880: T-SQL COLLATE argument being interpreted as a column
SQLGlot versions: tested with 25.9.0 and 25.8.1

When parsing a code block, I'm finding that the parsed query is returning the COLLATE collation argument as a column.

Example SQL:
```
select
    sql_text
    , some_id
from
    my_catalog.my_schema.my_table a
    inner join my_catalog.my_schema.my_other_table b
        on a.id = b.id
WHERE
    MySpecialColumn LIKE '%test%' COLLATE LATIN1_GENERAL_BIN

```

Given the above sql as variable "sql" - 

```
parsed_query = parse_one(sql=sql, dialect="tsql")
x = list(parsed_query.find_all(exp.Column))
for c in x:
    print(c)
```

I'm seeing "LATIN1_GENERAL_BIN" being emitted as a column, which surprised me. I wasn't sure if this was the desired behavior or not.

VaggelisD:
Fixes #3880

`COLLATE` should have a higher precedence than constructs such as `LIKE` but it's not possible to override it's parsing. For this reason, this PR unrolls `_parse_term` and transforms `COLLATE`'s RHS into an `exp.Var` (instead of `exp.Column`) if it's an unquoted identifier.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4376,7 +4376,8 @@ class Parser(metaclass=_Parser):
                 # fallback to Identifier / Var
                 if isinstance(expr, exp.Column) and len(expr.parts) == 1:
                     ident = expr.this
-                    this.set("expression", ident if ident.quoted else exp.var(ident.name))
+                    if isinstance(ident, exp.Identifier):
+                        this.set("expression", ident if ident.quoted else exp.var(ident.name))
 
         return this
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4360,7 +4360,25 @@ class Parser(metaclass=_Parser):
         return this
 
     def _parse_term(self) -> t.Optional[exp.Expression]:
-        return self._parse_tokens(self._parse_factor, self.TERM)
+        this = self._parse_factor()
+
+        while self._match_set(self.TERM):
+            klass = self.TERM[self._prev.token_type]
+            comments = self._prev_comments
+            expression = self._parse_factor()
+
+            this = self.expression(klass, this=this, comments=comments, expression=expression)
+
+            if isinstance(this, exp.Collate):
+                expr = this.expression
+
+                # Preserve collations such as pg_catalog."default" (Postgres) as columns, otherwise
+                # fallback to Identifier / Var
+                if isinstance(expr, exp.Column) and len(expr.parts) == 1:
+                    ident = expr.this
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -4360,7 +4360,25 @@ def _parse_bitwise(self) -> t.Optional[exp.Expression]:
...
+            this = self.expression(klass, this=this, comments=comments, expression=expression)
+
+            if isinstance(this, exp.Collate):
+                expr = this.expression
+
+                # Preserve collations such as pg_catalog."default" (Postgres) as columns, otherwise
+                # fallback to Identifier / Var
+                if isinstance(expr, exp.Column) and len(expr.parts) == 1:
+                    ident = expr.this
+                    this.set("expression", ident if ident.quoted else exp.var(ident.name))
```
georgesittas, line 4379: Are you sure `Column.this` is always gonna be an identifier here? `ident.quoted` will fail if not.

georgesittas (commented):
What about moving collate to the range parsers instead?

VaggelisD:
> What about moving collate to the range parsers instead?

This would mean that `COLLATE` will now have equal precedence to `LIKE` statements right? So although it'll be easier to add custom parsing, it might lead to incorrect ASTs if I'm not mistaken

sqlglot/parser.py
VaggelisD: Hmm, can we a parent `Column` without an `Identifier` child? Currently the RHS is parsed either into a (1) `exp.Column(this=exp.Identifier())` or (2) a string `exp.Literal`, depending on the quotes.

georgesittas: > Hmm, can we a parent Column without an Identifier child?

I'm not sure, I'd need to analyze the code paths in the parser to tell. Probably not, but if it's easy to be a bit more defensive here it'd be good. Your call.

georgesittas (approved):

sqlglot/parser.py
VaggelisD: I think we'll only create an `exp.Column` at `_parse_column_reference()`, and that's only if `this` is an `exp.Identifier`.

However, I'll go ahead and add one more branch just in case

PR #3888, opened 2024-08-07
feat(bigquery): Support for GENERATE_TIMESTAMP_ARRAY, DDB transpilation

VaggelisD:
This PR adds support for BQ's `GENERATE_TIMESTAMP_ARRAY()`, following the implementation of `GENERATE_DATE_ARRAY()`

Docs
----------
[BQ GENERATE_TIMESTAMP_ARRAY](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_timestamp_array) | [DuckDB GENERATE_SERIES](https://duckdb.org/docs/sql/functions/timestamp#timestamp-table-functions)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -209,13 +209,15 @@ def _arrow_json_extract_sql(self: DuckDB.Generator, expression: JSON_EXTRACT_TYP
     return arrow_sql
 
 
-def _implicit_date_cast(arg: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
-    return exp.cast(arg, exp.DataType.Type.DATE) if isinstance(arg, exp.Literal) else arg
+def _implicit_datetime_cast(
+    arg: t.Optional[exp.Expression], type: exp.DataType.Type = exp.DataType.Type.DATE
+) -> t.Optional[exp.Expression]:
+    return exp.cast(arg, type) if isinstance(arg, exp.Literal) else arg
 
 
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -867,3 +869,18 @@ def generatedatearray_sql(self, expression: exp.GenerateDateArray) -> str:
...
+            end = _implicit_datetime_cast(
+                expression.args.get("end"), type=exp.DataType.Type.TIMESTAMP
+            )
+
+            # BQ's GENERATE_TIMESTAMP_ARRAY is transformed to DuckDB'S GENERATE_SERIES
+            gen_series = exp.GenerateSeries(
+                start=start, end=end, step=expression.args.get("interval")
+            )
+
+            return self.sql(gen_series)
```
georgesittas, line 886: Don't we need to cast to `ARRAY<TIMESTAMP>` here, similar to `generatedatearray_sql`? It looks like these two methods can be consolidated, the logic is almost the same. I'd consider making it more DRY.

georgesittas (approved):

sqlglot/dialects/duckdb.py
VaggelisD: That was needed in `GENERATE_DATE_ARRAY` because DDB'S `GENERATE_SERIES` outputs a timestamp array by default.

I also thought of combining the logic but it will still require branching logic to implement the different casts etc. Can give it a shot though to see how it'd look

georgesittas (approved):

PR #3899, opened 2024-08-12
Feat!: transpile Unnest(GenerateDateArray(...)) to various dialects

georgesittas:
xref: https://github.com/TobikoData/sqlmesh/pull/2990

The motivation here is to make it easier to generate a date spine across several different dialects, by simply building a sqlglot expression that we can transpile when targeting them.

cc @sungchun12


sqlglot/expressions.py
```diff
@@ -5347,12 +5347,12 @@ class GapFill(Func):
 
 # https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_date_array
 class GenerateDateArray(Func):
-    arg_types = {"start": True, "end": True, "interval": False}
+    arg_types = {"start": True, "end": True, "step": False}
```
tobymao, line 5350: breaking change

tobymao (approved):

PR #3902, opened 2024-08-13
chore: New doc - Onboarding Doc

VaggelisD:
My intention with this doc was to fill in the gap for an end-to-end technical overview of SQLGlot; This is to help those who are interested in contributing but don't know where to start. 

I also believe this doc should also include whatever (unspoken) rules/tradeoffs we currently enforce in order to establish a common ground between maintainers & reduce iterations on PRs from new contributors. Lastly, this initial version covers the basics of the important modules but we can always add more details, new submodules etc as time progresses.


posts/onboarding.md
```diff
@@ -0,0 +1,526 @@
...
+## Optimizer
+
+The optimizer module in SQLGlot (`optimizer.py`) is responsible for producing canonicalized and efficient SQL queries by applying a series of optimization rules.
+
+Optimizations include simplifying expressions, removing redundant operations, and rewriting queries for better performance.
+
+The optimizer operates on the abstract syntax tree (AST) returned by the parser, transforming it into a more compact form while preserving the semantics of the original query.
+
+> [!NOTE]
+> The optimizer does not currently support statistics, so it cannot employ cost-based optimizations such as join reordering
```
barakalon, line 341: It’s more than just stats, right?

sqlglot only does _logical_ optimization. 

Maybe we should make a note that the underlying engine is almost always better at optimizing for performance.

barakalon (approved):
Nice! Super concise writing. I love it.

posts/onboarding.md
```diff
@@ -0,0 +1,526 @@
...
+>
+
+
+### Optimization rules
+The optimizer essentially applies [a list of rules](https://sqlglot.com/sqlglot/optimizer.html) in a sequential manner, where each rule takes in an AST and produces a canonicalized and optimized version of it.
+
+> [!WARNING]
+> The rule order is important, as some rules depend on others to work properly. We discourage manually applying individual rules, which may result in erroneous or unpredictable behavior.
+
+The first, foundational optimization task is to _canonicalize_ an AST, which is required by other optimization steps. These rules implement canonicalization:
```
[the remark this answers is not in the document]
VaggelisD: Hmm right, the terms "canonicalize", "normalize" etc are used in various places around SQLGlot so it can definitely be confusing.. Would "standardize" be the next best verb here?

posts/onboarding.md
VaggelisD: Right, my (inverse) thought was that physical optimization is basically cost-based optimization which requires stats to do properly, but we can simply state that we only do logical optimization

barakalon: Technically, you could collect stats and write a join ordering rule in sqlglot.

So I think logical vs physical is a more precise distinction. Or in other words, sqlglot’s optimizer operates on the AST, not query plans, which is a fundamental limitation.

itepifanio:
Great idea!

Do you guys plan to build a more detailed doc about creating custom dialects? The current doc is too introductory, adding a notebook-like with step by step building a dialect would be very helpful for begginers.

georgesittas:
> Great idea!
> 
> Do you guys plan to build a more detailed doc about creating custom dialects? The current doc is too introductory, adding a notebook-like with step by step building a dialect would be very helpful for beginners.

Implementing custom dialects basically means to override and tweak the parser/generator/tokenizer classes, so in that sense I'm not sure what exactly would go in such a guide. For example, the dialects we support are being constantly improved and the development is driven by issues / things we need for other applications (e.g. [SQLMesh](https://github.com/TobikoData/sqlmesh)).

Documenting each and every parser method is out of scope, so I don't see any other route besides reading the existing guides and then diving deeper into the codebase to understand what's going on.

Perhaps a blog that shows the process of solving bugs and/or implementing new features would be fun, but our bandwidth is somewhat limited to do that at the moment.

itepifanio:
> Implementing custom dialects basically means to override and tweak the parser/generator/tokenizer classes, so in that sense I'm not sure what exactly would go in such a guide. 

@georgesittas tweaking the parser/generator/tokenizer it's not always easy, the parser.py is huge and the built-in dialects are not always clear for begginers.

It would be very helpful to have a blog post or doc similar to a tutorial. I imagine displaying how to add a custom function at the select clause, a custom window function, a more complex transformation such as subquery addition, how to avoid tokentype conflicts, this kind of steps.

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+
+The mapping of symbols to `TokenType` enum values is stored in the dictionary `Tokenizer.KEYWORDS`.
+
+
+
+## Parser
+
+The parser module takes the list of tokens generated by the tokenizer and builds an [abstract syntax tree (AST)](https://en.wikipedia.org/wiki/Abstract_syntax_tree) from them.
+
+The AST represents the meaning (“semantics”) of a SQL statement, loosely defined as the set of actions the statement should invoke. The commonality of semantics across databases/dialects is what allows transpilation to occur.
```
georgesittas, line 90:
```suggestion
The AST represents a SQL statement by storing its _meaningful_ constituent components. For example, to represent a `SELECT` query we can only store its projections, filters, aggregation expressions and so on, but the `SELECT`, `WHERE` and `GROUP BY` keywords don't need to be preserved, since they are implied.

In SQLGlot's case, the AST usually stores information that captures the _semantics_ of an expression, i.e. its meaning, which is what allows it to transpile SQL code between different dialects. A blog post that explains this idea at a high-level can [be found here](https://tobikodata.com/transpiling_sql1.html).
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+SQLGlot’s parser follows the “recursive descent” approach, meaning that it relies on mutually recursive methods in order to understand SQL syntax and the various precedence levels between combined SQL expressions.
+
+For example, the SQL statement `CREATE TABLE a (col1 INT)` will be parsed into an `exp.Create` expression that includes the kind of object being created (table) and its schema (table name, column names and types). The schema will be parsed into an `exp.Schema` node that contains one column definition `exp.ColumnDef` node for each column.
+
+The relationships between SQL constructs are encoded in methods whose names begin with “_parse_”, such as `_parse_create()`. We refer to these as “parsing methods”.
+
+For example, to parse the statement above SQLGlot calls the `_parse_create()` method. Then:
+- `_parse_create()` determines that a `TABLE` is being created
+- Because a table is being created, it knows the table name comes next and calls `_parse_table_parts()` (the table name may have multiple parts like `catalog.schema.table`)
+- Because a table is being created, it knows the schema definition comes after the table name and calls `_parse_schema()`
```
georgesittas, line 142:
```suggestion
- It also knows that a schema definition comes after the table name and calls `_parse_schema()`
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+- A schema contains comma-separated column definitions, so `_parse_schema()` calls `_parse_column_def()` on each item
+- After all elements are parsed, `_parse_create()` assembles and returns an `exp.Create` expression
+
+A key aspect of the SQLGlot parser is that the parsing methods are composable and can be passed as arguments. We describe how that works in the next section.
+
+### Utility methods
+SQLGlot provides helper methods for common parsing tasks. They should be used whenever possible to reduce code duplication and manual token/text matching.
+
+Many helper methods take a parsing method argument specifying the parsing method they should apply to sub-components of the object they are parsing.
```
georgesittas, line 151:
```suggestion
Many helper methods take a parsing method argument, which is invoked to parse part(s) of the clause they're supposed to parse.
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+    - The mappings are grouped by general semantic type (e.g., data type expressions), with each type having its own dictionary.
+
+The `Expression -> str` mappings can be implemented as either:
+- Entries in the `TRANSFORMS` dictionary, which are best suited for single-line generations.
+- Functions with names of the form `<expr_name>_sql(...)`. For example, to generate SQL for an `exp.SessionParameter` node we can override the base generator method by defining the method `def sessionparameter_sql(...)` in a dialect’s Generator class.
+
+### Generating queries
+The key method in the Generator module is the `sql()` function, which is responsible for generating the string representation of each expression.
+
+It locates the mapping of expression to string or generator `Callable` by examining  the keys of the `TRANSFORMS` dictionary and searching the `Generator`'s methods for an `<exprname>_sql()` method.
```
georgesittas, line 224:
```suggestion
It locates the mapping of expression to string or generator `Callable` by examining the keys of the `TRANSFORMS` dictionary and searching the `Generator`'s methods for an `<exprname>_sql()` method.
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+
+It then substitutes the string or invokes the callable to return the correct SQL code.
```
georgesittas, line 226:
```suggestion
It then calls the appropriate callable to produce the corresponding SQL code.

> [!IMPORTANT]
> If there is both a `TRANSFORM` entry and an `<expr_name>_sql(...)` method for a given expression, the generator will use the former to convert the expression into a string.
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+> [!WARNING]
+> The rule order is important, as some rules depend on others to work properly. We discourage manually applying individual rules, which may result in erroneous or unpredictable behavior.
+
+The first, foundational optimization task is to _standardize_ an AST, which is required by other optimization steps. These rules implement canonicalization:
+
+#### Qualify
+The most important rule in the optimizer is `qualify`, which is responsible for rewriting the AST such that all tables and columns are _normalized_ and _qualified_.
+
+##### Identifier normalization
+The normalization step (`normalize_identifiers.py`) transforms all unquoted identifiers to lower or upper case (depending on the dialect), essentially making them case-insensitive.
```
georgesittas, line 356:
```suggestion
The normalization step (`normalize_identifiers.py`) transforms some identifiers, by converting them into lower or upper case, ensuring the semantics are preserved in each case (e.g. by respecting case-sensitivity).

This transformation reflects how identifiers would be resolved by the engine corresponding to each SQL dialect, and plays a very important role in the standardization of the AST.
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+
+> [!NOTE]
+> Some dialects, like BigQuery, treat identifiers as case-insensitive even when they're quoted, so all identifiers are normalized.
```
georgesittas, line 359: This isn't generally correct for BigQuery, since it treats tables (not CTEs) as case-sensitive, for example. OTOH, DuckDB does treat all identifiers as case-insensitive.

```suggestion
> [!NOTE]
> Some dialects, like DuckDB, treat identifiers as case-insensitive even when they're quoted, so all identifiers are normalized.
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+
+schema = {"tbl": {"col": "INT"}}
+expression = sqlglot.parse_one("SELECT * FROM tbl")
+qualify(expression, schema=schema).sql()
+
+
+# 'SELECT "tbl"."col" AS "col" FROM "tbl" AS "tbl"'
+```
+
+#### Type annotation
```
georgesittas, line 413:
```suggestion
#### Type inference
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+
+Like the database, SQLGlot needs to know column data types to perform some optimizations. In many cases, both transpilation and optimization need type information to work properly.
+
+Type annotation begins with user-provided information about column data types for at least one table. SQLGlot then traverses the AST, propagating that type information and attempting to infer the type returned by each AST expression.
+
+Users may need to provide column type information for multiple tables to successfully annotate all expressions in the AST.
+
+
+
+
```
georgesittas, line 426: Nit: let's try to be consistent about the whitespace we use. My personal preference is to always separate sections with a single blank line. Open to alternatives as long as we stick to one style.
```suggestion
```

posts/onboarding.md
```diff
@@ -0,0 +1,525 @@
...
+      root_table
+  ) AS intermediate_table
+)
+SELECT
+  traced_col
+FROM
+  cte
+```
+
+2. Qualify the standalone query, expanding `*`s and annotating all column names with source tables:
```
georgesittas, line 494:
```suggestion
2. Qualify the query produced by the earlier step, expanding `*`s and qualifying all column references with the corresponding source tables:
```

georgesittas (approved):
Great work @VaggelisD 👏

Left a few comments to consider.

posts/onboarding.md
VaggelisD: This reads a bit off to me, the second paragraph seems disconnected and/or repeating the first, should we merge them?

georgesittas: Yeah, I can see how this may be confusing -- I'm open to rephrasing it.

Imo we shouldn't conflate the _semantics_ of an expression with its grammatical structure, which is what an AST is primarily meant to represent.

SQLGlot's AST is special, because we inject additional information to it as opposed to what you'd get by a typical parser. For example, some `Func` nodes store a return type, and other nodes implicily store whether the arguments can be strings or datetime values (e.g. the `TsOrDs*` family).

That's what I usually think about when I see the term "semantics". Also, keep in mind that we don't really represent the semantics accurately in many cases, e.g. when expressions are type-sensitive.

posts/onboarding.md
VaggelisD: The explanations of these optimizer steps originate mostly from the docstrings (this one is from `normalize_identifiers.py` in particular) as I didn't want to stray off; If they're outdated or require new details to be up to date, we can also update them too (?)

georgesittas: Yep, they're probably inaccurate / outdated and we should fix them.

posts/onboarding.md
VaggelisD: I see what you mean, I think the confusing part is that I'd read `The AST represents ...` as basically `The [SQLGLot] AST represents ...` which can be conflicting with the 2nd paragraph. From what I understand your intention is to distinguish a typical AST (i.e storing just the grammatical structure) produced from a general purpose SQL parser to the "enhanced" version that is produced by SQLGlot. 

If this is the case, I think your 2nd explanation conveys the message better

posts/onboarding.md
VaggelisD: Hmm, I think if I was to change the description here I'd go with something like:

```
For example, to parse the statement above the entrypoint method `_parse_create()` is called. Then:
- In `_parse_create()`, the parser determines that a `TABLE` is being created
- Because a table is being created, the table name is expected next and thus `_parse_table_parts()` is invoked; This is because the table name may have multiple parts such as `catalog.schema.table`
- Having processed the table name, the parser continues parsing the schema definition by calling `_parse_schema()`
```

It may have a bit more noise added, but I think the bullet point breakdown is stitched together more "naturally". What do you say?

georgesittas: Sure yeah

posts/onboarding.md
georgesittas: What about:

> An AST generally stores the _meaningful_ constituent components of a statement. For example, to represent a `SELECT` query we can only store its projections, filters, aggregation expressions and so on, but the `SELECT`, `WHERE` and `GROUP BY` keywords don't need to be preserved, since they are implied.

> In SQLGlot's case, the AST usually stores additional information that captures the _semantics_ of an expression, i.e. its meaning, which is what allows it to transpile SQL code between different dialects. A blog post that explains this idea at a high-level can [be found here](https://tobikodata.com/transpiling_sql1.html).

PR #3912, opened 2024-08-15
Fix(clickhouse)!: treat DATABASE kind as SCHEMA (and conversely) in exp.Create

georgesittas:


tobymao (approved):

tobymao:
should we not map "DATABASE" to schema at the token level?

georgesittas:
> should we not map "DATABASE" to schema at the token level?

It won't help here because `_parse_create` uses the token's text when setting the `kind` arg.

PR #3913, opened 2024-08-15
Feat(clickhouse): add support for TryCast generation

georgesittas:


sqlglot/dialects/clickhouse.py
```diff
@@ -870,6 +899,14 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
 
+        def trycast_sql(self, expression: exp.TryCast) -> str:
+            target_type = self.sql(expression.to)
+            if target_type.upper() in self.TRY_CAST_TYPES:
```
tobymao, line 904: will this never be non upper?

tobymao (approved):

sqlglot/dialects/clickhouse.py
georgesittas: Yea, because CH has some case-sensitive types like `UInt8`, but some case-insensitive ones too like `DATE`.

PR #3917, opened 2024-08-15
Fix(clickhouse): make try_cast toXXXOrNull() functions case-specific

treysp:
Clickhouse implements `TRY_CAST` as type-specific functions with names of form `toXXXOrNull()`. 

The function names must match the case-sensitive type names, so this PR adds the case-specific type names for substitution into the function name.

Case-specific function names in Clickhouse repo [here](https://github.com/ClickHouse/ClickHouse/blob/275de04b8f6bb8c9334bf8070001afe2dab0b17d/src/Functions/FunctionsConversion.cpp#L2939-L2989).

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -863,31 +863,31 @@ class ClickHouse(Dialect):
 
         # https://github.com/ClickHouse/ClickHouse/blob/275de04b8f6bb8c9334bf8070001afe2dab0b17d/src/Functions/FunctionsConversion.cpp#L2939-L2989
         TRY_CAST_TYPES = {
-            "DATE",
-            "DATE32",
-            "DATETIME",
-            "DATETIME64",
-            "DECIMAL32",
-            "DECIMAL64",
-            "DECIMAL128",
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -911,7 +911,9 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
         def trycast_sql(self, expression: exp.TryCast) -> str:
             target_type = self.sql(expression.to)
             if target_type.upper() in self.TRY_CAST_TYPES:
-                return self.func(f"to{target_type}OrNull", expression.this)
+                return self.func(
+                    f"to{self.TRY_CAST_TYPES[target_type.upper()]}OrNull", expression.this
+                )
```
georgesittas, line 916: Let's instead do:

```python
type_sql = self.TRY_CAST_TYPES.get(target_type.upper())
if type_sql:
    return self.func(f"to{type_sql}OrNull", expression.this)
...
```

georgesittas (approved):
Thanks for fixing this 👍

PR #3920, opened 2024-08-16
Fix(clickhouse): add timestamptz type mapping

treysp:
Clickhouse does not distinguish between TZ and non-TZ types. The `TIMESTAMP` type accepts an optional timezone argument.


georgesittas (approved):
LGTM - only title needs to be updated

PR #3924, opened 2024-08-17
Fix(clickhouse)!: use creatable kind mapping dict for schema<-->database substitution

treysp:


sqlglot/dialects/clickhouse.py
```diff
@@ -1057,6 +1066,12 @@ def create_sql(self, expression: exp.Create) -> str:
 
             return f"{create_sql}{comment_sql}"
 
+        def drop_sql(self, expression: exp.Drop) -> str:
+            # ClickHouse only has DATABASEs and objects under them, eg. TABLEs, VIEWs, etc
+            if expression.args["kind"] == "SCHEMA":
+                expression.args["kind"] = "DATABASE"
```
georgesittas, line 1072:
```suggestion
            if expression.kind == "SCHEMA":
                expression.set("kind", "DATABASE")
```

sqlglot/dialects/clickhouse.py
```diff
@@ -453,6 +453,15 @@ def _parse_create(self) -> exp.Create | exp.Command:
 
             return create
 
+        def _parse_drop(self, exists: bool = False) -> exp.Drop | exp.Command:
+            drop = super()._parse_drop(exists)
+
+            # DATABASE in ClickHouse is the same as SCHEMA in other dialects
+            if isinstance(drop, exp.Drop) and drop.args["kind"] == "DATABASE":
+                drop.args["kind"] = "SCHEMA"
```
georgesittas, line 461:
```suggestion
            if isinstance(drop, exp.Drop) and drop.kind == "DATABASE":
                drop.set("kind", "SCHEMA")
```

georgesittas (approved):
Here's another idea: we can create an indirection layer in the base classes to make overriding these create kinds more easy. So for example you could do

```
create_kind = create_token.text.upper()
exp.Create(..., kind=self.KIND_MAPPING.get(create_kind) or create_kind, ...)
```

And then use the reverse mapping in the generator, created in `_Dialect`.

sqlglot/dialects/clickhouse.py
treysp: That's how I originally did it (mimicking your create PR), but:

```
  File "/Users/trey/tobiko/sqlglot/sqlglot/dialects/clickhouse.py", line 460, in _parse_drop
    if isinstance(drop, exp.Drop) and drop.kind == "DATABASE":
                                      ^^^^^^^^^
AttributeError: 'Drop' object has no attribute 'kind'
```

georgesittas: Ah I see. Let's add that property to `Drop` too then, I think it makes sense to reflect what we do for `Create`.

georgesittas (approved):
LGTM! FYI @tobymao

PR #3926, opened 2024-08-17
Refactor: iteratively generate binary expressions

MatMoore:
Refactor the implementation for binary expressions in the generator to use a partially iterative style, similar to `set_operations`.

The intention is to avoid unnecessary recursion while generating chains of N binary operations, (e.g. a || b || c...), because if the chain gets too long, the generator hits Python's recursion limit.

As discussed in https://github.com/tobymao/sqlglot/pull/3788, I've limited the change to just the binary operations, and we still call the recursive `sql` method each time we encounter a different type of Expression, so the solution is still a mix of recursive and iterative, but it seems to work for these long chains of binary operators.

[The pathological example I shared earlier](https://gist.github.com/MatMoore/620775b07b0cbeeb80c46928843eee2c) now runs without hitting the recursion limit.


tobymao:
need to fix the tests, but awesome!

MatMoore:
Whoops, opened the PR a bit too early yesterday, but it should be working now. I've included an extra check in there to make sure that the operator has not changed while looping over the slack.

This could be generalised to support mixed operators, e.g. `1 + 2 * 3`, but it would need a new function that can convert a `Binary` instance to the correct `op` value. At the moment this mapping is only known by the `_sql` methods, e.g. 

```python
def add_sql(self, expression: exp.Add) -> str:
        return self.binary(expression, "+")
```

sqlglot/generator.py
```diff
@@ -3315,8 +3315,26 @@ def use_sql(self, expression: exp.Use) -> str:
...
+        sqls: t.List[str] = []
+        stack: t.List[t.Union[str, exp.Expression]] = [expression]
+        subclass = type(expression)
+        operator = expression.args.get("operator")
+
+        while stack:
+            node = stack.pop()
+
+            if isinstance(node, subclass) and (
+                (operator is None) or (operator == node.args.get("operator"))
```
tobymao, line 3327: what happens when operator is None?

MatMoore: If operator is None, then it only depends on the node's type. E.g. if this function was called from `sql_and`, then this code will
- directly process `And` nodes inside the loop
- call `.sql` if either operand is not an And

The `operator` check is there because `Operator` expressions produce different `op` values if they have different operators, e.g. `Operator(+)` vs `Operator(&)`.

I think this can be made a lot cleaner if we remove `op` as a parameter and instead compute it for each node within the loop, but I didn't have time to try this out this weekend.

VaggelisD: What is the `operator` arg referring to? I've not been able to find an `exp.Binary` expression with it defined, it should only exist in `exp.Operator` which is Postgres specific syntax afaict.

I think your intention here is to generate chains of identical binary nodes (e.g. connected concats) but `operator` is not defined so it will always be `None`

VaggelisD: My initial thought was to turn `op` into an property which would allow us to do `op = expression.op`,  but different dialects might use different symbols for the same node e.g. `exp.PropertyEQ` is defined differently in the Materialize dialect.

Another idea might be to create a `Generator` mapping `exp.Binary -> op (str)` which _can_ be overridden in dialect, and it will also remove the need of having separate `<bin>_sql()` functions for each expression.

georgesittas (commented):
Let's try to test this change if possible. I'd generate a crazy query programmatically and show that the end-start time is sufficiently small. We do this for some other cases, in e.g. `test_parser.py`.

I don't really understand the discussion re: `Operator`. Can't we get rid of the second term in the `if` branch by doing an exact type check instead of `isinstance`?

```
if type(node) is original_type:
    # extend stack
else:
    # append to sqls
```

Then you don't have to deal with `Operator` as a special case. I'd prefer this because `OPERATOR(...)` is very rare in practice, so the added complexity doesn't feel worth it.

MatMoore:
> Let's try to test this change if possible. I'd generate a crazy query programmatically and show that the end-start time is sufficiently small. We do this for some other cases, in e.g. `test_parser.py`.

Yep, sounds good. I will have a look at that example when I have a chance

> I'm not sure if I understand the discussion re: `Operator`. Can't we get rid of the second term in the `if` branch by doing an exact type check instead of `isinstance`?
> 
> ```
> if type(node) is original_type:
>     # extend stack
> else:
>     # append to sqls
> ```
> 
> Then you don't have to deal with `Operator` as a special case. I'd prefer this because `OPERATOR(...)` is very rare in practice, so the added complexity doesn't feel worth it.

I don't think this will work, as the various kinds of Operators are all the same class, rather than subclasses. I get your point about the extra complexity though.

georgesittas:
> I don't think this will work, as the various kinds of Operators are all the same class, rather than subclasses. I get your point about the extra complexity though.

I see what you mean. Then what about this:

```diff
     def binary(self, expression: exp.Binary, op: str) -> str:
-        op = self.maybe_comment(op, comments=expression.comments)
-        return f"{self.sql(expression, 'this')} {op} {self.sql(expression, 'expression')}"
+        sqls: t.List[str] = []
+        stack: t.List[t.Union[str, exp.Expression]] = [expression]
+        binary_type = type(expression)
+
+        while stack:
+            node = stack.pop()
+
+            if type(node) is binary_type:
+                op_func = node.args.get("operator")
+                if op_func:
+                    op = f"OPERATOR({self.sql(op_func)})"
+
+                stack.append(node.right)
+                stack.append(f" {self.maybe_comment(op, comments=node.comments)} ")
+                stack.append(node.left)
+            else:
+                sqls.append(self.sql(node))
+
+        return "".join(sqls)
```

I think it's a bit easier to grok vs checking for `operator is None` etc.

georgesittas:
Thanks for the contribution @MatMoore! I'll take this to the finish line.

georgesittas:
See https://github.com/tobymao/sqlglot/commit/c6973572dfd953b5539bb4e9dcba402c0c3c6acf

MatMoore:
Thanks @georgesittas!

PR #3931, opened 2024-08-19
fix(duckdb): Fix exp.Unnest generation for BQ's nested arrays

VaggelisD:
In BigQuery `UNNEST`ing a nested array (i.e array of structs) explodes the struct fields, e.g.:

```SQL
BQ> SELECT name, age FROM UNNEST([STRUCT('Foo' AS name, 25 AS age), STRUCT('Bar' AS name, 34 AS age)]);

name	age
----------------
Foo	 25
Bar	 34
```

On main branch the `UNNEST()` is transpiled as-is, leading to invalid DuckDB:
```SQL
D> SELECT name, age FROM UNNEST([{'name': 'Foo', 'age': 25}, {'name': 'Bar', 'age': 34}]);

Error: Binder Error: Referenced column "name" not found in FROM clause!
```

This is because DuckDB's `UNNEST()` will always unnest one level by default, so the result is the underlying struct(s):
```SQL
D> SELECT * FROM UNNEST([{'name': 'Foo', 'age': 25}, {'name': 'Bar', 'age': 34}]);

{'name': Foo, 'age': 25}    
{'name': Bar, 'age': 34}
```

This PR fixes this issue by using DuckDB's _recursive_ specifier which matches BQ's semantics by unnesting both the array and the top-level struct:

```SQL
D> SELECT name, age FROM (SELECT UNNEST([{'name': 'Foo', 'age': 25}, {'name': 'Bar', 'age': 34}], max_depth => 2));

name       age
varchar   int32
----------------
Foo        25
Bar        34
```

Docs
--------
[BigQuery UNNEST](https://cloud.google.com/bigquery/docs/arrays) | [DuckDB UNNEST](https://duckdb.org/docs/sql/query_syntax/unnest.html)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -593,12 +593,12 @@ class BigQuery(Dialect):
                 return None
 
             unnest_expr = seq_get(unnest.expressions, 0)
-            if unnest_expr and not unnest.args.get("alias"):
+            if unnest_expr:
                 from sqlglot.optimizer.annotate_types import annotate_types
 
                 unnest_expr = annotate_types(unnest_expr)
 
-                # Unnesting a nested array (i.e array of structs) without UNNEST alias explodes the top-level struct fields,
+                # Unnesting a nested array (i.e array of structs) explodes the top-level struct fields,
                 # in contrast to other dialects such as DuckDB which flattens only the array by default
                 if unnest_expr.is_type(exp.DataType.Type.ARRAY):
                     if any(
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -591,15 +591,15 @@ class BigQuery(Dialect):
 
             if unnest:
                 expr = seq_get(unnest.expressions, 0)
+                if expr:
+                    from sqlglot.optimizer.annotate_types import annotate_types
 
-                # Unnesting a nested array (i.e array of structs) in BQ explodes the struct fields,
-                # in contrast to other dialects such as DuckDB which flattens only the array by default
-                if (
-                    expr
-                    and (isinstance(expr, exp.Array) or expr.is_type(exp.DataType.Type.ARRAY))
-                    and expr.find(exp.Struct)
-                ):
-                    unnest.set("nested_array", True)
+                    expr = annotate_types(expr)
+
+                    # Unnesting a nested array (i.e array of structs) in BQ explodes the top-level struct fields,
+                    # in contrast to other dialects such as DuckDB which flattens only the array by default
+                    if expr.is_type(exp.DataType.Type.ARRAY) and expr.find(exp.Struct):
+                        unnest.set("explode_array", True)
 
             return unnest
 
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -600,12 +600,11 @@ class BigQuery(Dialect):
 
                 # Unnesting a nested array (i.e array of structs) explodes the top-level struct fields,
                 # in contrast to other dialects such as DuckDB which flattens only the array by default
-                if unnest_expr.is_type(exp.DataType.Type.ARRAY):
-                    if any(
-                        array_elem.is_type(exp.DataType.Type.STRUCT)
-                        for array_elem in unnest_expr._type.expressions
-                    ):
-                        unnest.set("explode_array", True)
+                if unnest_expr.is_type(exp.DataType.Type.ARRAY) and any(
+                    array_elem.is_type(exp.DataType.Type.STRUCT)
+                    for array_elem in unnest_expr._type.expressions
+                ):
+                    unnest.set("explode_array", True)
 
             return unnest
 
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -589,16 +589,22 @@ class BigQuery(Dialect):
         def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
             unnest = super()._parse_unnest(with_alias=with_alias)
 
-            if unnest:
-                expr = seq_get(unnest.expressions, 0)
-                if expr:
-                    from sqlglot.optimizer.annotate_types import annotate_types
+            if not unnest:
+                return None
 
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -5325,6 +5325,7 @@ class Unnest(Func, UDTF):
         "expressions": True,
         "alias": False,
         "offset": False,
+        "nested_array": False,
```
georgesittas, line 5328: The intent isn't very clear with this name imo, consider renaming to "explode_array" or something similar.

VaggelisD:
> Have you tested structs with nesting level > 1? Eg. `[struct(x struct(...))]`.

Good catch, I initially thought that it exploded all fields but this is the same case as with the struct flatten, e.g. the following query is invalid BigQuery:

```SQL
BQ> SELECT name, age, inner_str FROM UNNEST([STRUCT('Foo' AS name, 25 AS age, STRUCT(5 as inner_str) as str)]);

Unrecognized name: inner_str at [1:19]
```

Even though the UI shows all the fields flattened with this query:

```SQL
BQ> SELECT* FROM UNNEST([STRUCT('Foo' AS name, 25 AS age, STRUCT(5 as inner_str) as str)]);

name age str.inner_str
-----------------------
Foo    25       5
```

This means that BQ will only UNNEST the top level array and the top level struct

Thankfully DuckDB offers a `max_depth` specifier which can limit the recursive unnest, so with this in mind I think DuckDB's transpilation should change to `FROM (SELECT UNNEST(..., max_depth => 2)`, e.g.:

```SQL
D SELECT name, age, str FROM (SELECT UNNEST([{'name': 'Foo', 'age': 25, 'str': {'inner_str': 5}}], max_depth => 2));
┌─────────┬───────┬───────────────────────────┐
│  name   │  age  │            str            │
│ varchar │ int32 │ struct(inner_str integer) │
├─────────┼───────┼───────────────────────────┤
│ Foo     │    25 │ {'inner_str': 5}          │
└─────────┴───────┴───────────────────────────┘
```

sqlglot/dialects/bigquery.py
```diff
@@ -591,15 +591,15 @@ def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
...
-                if (
-                    expr
-                    and (isinstance(expr, exp.Array) or expr.is_type(exp.DataType.Type.ARRAY))
-                    and expr.find(exp.Struct)
-                ):
-                    unnest.set("nested_array", True)
+                    # Unnesting a nested array (i.e array of structs) in BQ explodes the top-level struct fields,
+                    # in contrast to other dialects such as DuckDB which flattens only the array by default
+                    if expr.is_type(exp.DataType.Type.ARRAY) and expr.find(exp.Struct):
+                        unnest.set("explode_array", True)
```
georgesittas, line 602: Calling `expr.find` is not very robust because it searches the whole tree and may find a struct in a lower depth. Also, with this approach we don't leverage the inferred types at all, so we may miss functions / complex expressions that produce structs but aren't `exp.Struct`.

Consider instead breaking up the condition and being more explicit:

1. check if `is_type(array)`
2. check if `expressions[0].is_type` struct, unless there are no expressions set in the array

sqlglot/dialects/duckdb.py
```diff
@@ -886,11 +886,14 @@ def objectinsert_sql(self, expression: exp.ObjectInsert) -> str:
...
+                )
 
-            # Transpile BQ's UNNEST of nested array which must be made recursive to explode the struct fields
-            # by transforming "FROM UNNEST(...)" to DDB's "FROM (SELECT UNNEST(..., recursive => TRUE))"
-            if expression.args.get("nested_array"):
-                unnest_sql = f"(SELECT {unnest_sql[:-1]}, recursive => TRUE))"
+            unnest_sql = super().unnest_sql(expression)
 
-            return unnest_sql
+            return f"(SELECT {unnest_sql})" if explode_array else unnest_sql
```
georgesittas, line 899: Can we only have `UNNEST` under `FROM/JOIN` in BigQuery? Is this wrapping always safe?

georgesittas (commented):
A couple of comments, but the PR looks good otherwise.

sqlglot/dialects/duckdb.py
VaggelisD: Yeah, BQ does not allow `SELECT UNNEST(...)` as does DuckDB, and all the doc examples are based on `FROM` / `CROSS JOIN`. 

However, I think the `JOIN` side might have issues with this wrapping especially if the UNNEST is aliased. Will investigate more, hopefully we can solve both , otherwise I'll focus exclusively on `FROM`

georgesittas: You can also use the builder API, e.g.

```python
if explode_array:
    select = exp.Select().subquery(unnest_alias)
    select.set("expressions", [unnest_sql])  # this is already a string so we won't have recursion issues
    unnest_sql = self.sql(select)

return unnest_sql
```

Just need to pop the unnest alias prior to generating it and saving the result in `unnest_sql` if `explode_array`. Something like that.

georgesittas (approved):

sqlglot/dialects/duckdb.py
```diff
@@ -884,3 +884,24 @@ def objectinsert_sql(self, expression: exp.ObjectInsert) -> str:
...
+                # This is transpiled to DDB by transforming "FROM UNNEST(...)" to "FROM (SELECT UNNEST(..., max_depth => 2))"
+                expression.expressions.append(
+                    exp.Kwarg(this=exp.var("max_depth"), expression=exp.Literal.number(2))
+                )
+
+                # If BQ's UNNEST is aliased, we transform it from a column alias to a table alias in DDB
+                alias = expression.args.get("alias")
+                if alias:
+                    expression.set("alias", None)
+                    alias = exp.TableAlias(this=seq_get(alias.args.get("columns"), 0))
```
georgesittas, line 901: Note: we chatted about making this change in BigQuery's parser, but it looked like it'd require more work to do it properly as there were various optimizer tests failing because the `TableAlias` didn't have `columns` set.

This PR is good for the BigQuery -> DuckDB direction but my guess is that it will probably require a refactor in the future if we need to transpile towards other dialects.

tobymao (approved):

PR #3932, opened 2024-08-19
Add support for `MINUS` keyword in `Spark`

### LINKED ISSUE #3930: Add `MINUS` keyword support for `Spark` and related dialects
`MINUS` is a supported keyword in Spark and Databricks, but fails in sqlglot:

```python
from sqlglot import parse as parse_sql

def test_parses_minus_statement(migration_index):
    query = "SELECT * FROM old.things MINUS SELECT * FROM old.stuff"
    statements = parse_sql(query, read='databricks')
```

Throws:
```
        if self.error_level == ErrorLevel.IMMEDIATE:
>           raise error
E           sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 37.
E             SELECT * FROM old.things MINUS SELECT * FROM old.stuff
```

renardeinside:
Resolves #3930

### DIFF
--- a/sqlglot/dialects/spark.py
+++ b/sqlglot/dialects/spark.py
@@ -7,6 +7,7 @@ from sqlglot.dialects.dialect import rename_func, unit_to_var
 from sqlglot.dialects.hive import _build_with_ignore_nulls
 from sqlglot.dialects.spark2 import Spark2, temporary_storage_provider, _build_as_cast
 from sqlglot.helper import ensure_list, seq_get
+from sqlglot.tokens import TokenType
 from sqlglot.transforms import (
     ctas_with_tmp_tables_to_create_tmp_view,
     remove_unique_constraints,
@@ -100,6 +101,13 @@ class Spark(Spark2):
             for q in t.cast(t.List[str], Spark2.Tokenizer.QUOTES)
             for prefix in ("r", "R")
         ]
+
+        KEYWORDS = {
+            **Spark2.Tokenizer.KEYWORDS,
+            # as per https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-setops.html
+            # MINUS is an alias for EXCEPT
+            "MINUS": TokenType.EXCEPT,
+        }
 
     class Parser(Spark2.Parser):
         FUNCTIONS = {
--- a/tests/dialects/test_spark.py
+++ b/tests/dialects/test_spark.py
@@ -237,7 +237,7 @@ TBLPROPERTIES (
         self.validate_all(
             "SELECT * FROM db.table1 MINUS SELECT * FROM db.table2",
             write={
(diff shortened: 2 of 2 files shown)


sqlglot/dialects/spark.py
```diff
@@ -101,6 +102,13 @@ class Tokenizer(Spark2.Tokenizer):
             for prefix in ("r", "R")
         ]
 
+        KEYWORDS = {
+            **Spark2.Tokenizer.KEYWORDS,
+            # as per https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-setops.html
+            # MINUS is an alias for EXCEPT
+            "MINUS": TokenType.EXCEPT,
+        }
```
VaggelisD, line 110: I see that `MINUS` is a synonym for `EXCEPT` on other dialects as well such as Oracle, Snowflake and Redshift; With this addition, I'd be inclined to revert all these overrides and generalize it by moving this mapping to `tokens.py` under `"EXCEPT"`.

tests/dialects/test_spark.py
```diff
@@ -233,6 +233,14 @@ def test_hint(self, logger):
             },
         )
 
+    def test_minus(self):
+        self.validate_all(
+            "SELECT * FROM db.table1 MINUS SELECT * FROM db.table2",
+            write={
+                "spark": "SELECT * FROM db.table1 EXCEPT SELECT * FROM db.table2",
+            }
+        )
```
VaggelisD: Small nit, it would be best if new tests are appended at the bottom of each test file.


Also, this test case should include Databricks and possibly Spark <-> Databricks transpilations as well.

 If we decide to generalize the `MINUS -> TokenType.EXCEPT` mapping across all Tokenizers, we can move the respective tests to `test_dialect.py`.

georgesittas:
> Is this only supported from Spark 3+ onwards?

I verified that it's supported in 2.4.8 as well.

sqlglot/dialects/spark.py
georgesittas: We can try this refactor, just make sure that `MINUS` can still be used as an identifier & projection alias in dialects that don't treat is as `EXCEPT`.

tests/dialects/test_spark.py
renardeinside: Hey, I've moved the test to the bottom and added `databricks` to writes. Do we need more testcases?

VaggelisD:
Thanks for the quick iterations @renardeinside, I see that the latest commit changed the styling/format of some test cases, can we revert that?

We can then merge this PR and I'll start working on the refactor that we discussed about.

tests/dialects/test_spark.py
VaggelisD: That's fine for now, I'll will move these cases to `test_dialect.py`  anyways once I get to clean the other dialects

PR #3938, opened 2024-08-19
Feat!: Add timezone support to exp.TimeStrToTime

erindru:
This PR is to resolve the following issue:

```
from datetime import datetime
from zoneinfo import ZoneInfo

>>> exp.values([datetime(2023, 1, 1, tzinfo=ZoneInfo("America/Los_Angeles"))]).sql(dialect="postgres")
"VALUES CAST('2023-01-01 00:00:00-08:00' AS TIMESTAMP)"
```

That is, timezone information being discarded. The expected output should be:

```
VALUES CAST('2023-01-01 00:00:00-08:00' AS TIMESTAMPTZ)
```

The cause is that Generator for the `exp.TimeStrToTime` objects that `exp.values()` creates was just doing a basic cast to `DataType.Type.TIMESTAMP`.

So, I made the following modifications:

 - Added a new `timezone` property to `exp.TimeStrToTime` to capture the timezone information
 - Return a cast to `DataType.TIMESTAMPTZ` from `timestrtotime_sql()` if there is timezone information, otherwise just keep the existing cast to `DataType.Type.TIMESTAMP`
 - ...Unless the expression being passed to `timestrtotime_sql()` is already a cast to the correct type, in which case it is not re-cast. This fixes an issue where reading and writing the same SQL multiple times kept adding more and more casts.
 
Note that some databases still cast to `TIMESTAMP` or `DATETIME` because they dont have `TIMESTAMPTZ` types. However, I checked that they can all run the produced expression and produce the correct result. These databases still interpret the timezone portion of the string literal and produce a correct timestamp.

Note that Clickhouse doesnt have a `TIMESTAMPTZ` type per se, it uses a parameterised DateTime type, eg `DateTime('America/Los_Angeles')`. So in implementing support for this, I made the following modifications for Clickhouse:
 
 - Renamed `TIMESTAMP` to `DateTime`. Clickhouse doesnt have a documented type called `TIMESTAMP`. It just happens to an alias for `DateTime`, which I verified by running queries attempting to cast things to `TIMESTAMP` and asserting that the DB driver returned values of type `DateTime`.
 - Made `DATETIME` render as `DateTime` for consistency with all the other Clickhouse data types and updated the tests


tests/test_expressions.py
```diff
@@ -791,6 +791,7 @@ def test_properties_from_dict(self):
 
     def test_convert(self):
         from collections import namedtuple
+        from zoneinfo import ZoneInfo
```
georgesittas, line 794: This was [added](https://docs.python.org/3/library/zoneinfo.html) in python 3.9, so it's gonna be problematic when running 3.7 and 3.8. Is there any other way to do it?

erindru: yeah, I just realised :) switching it to pytz now

sqlglot/expressions.py
```diff
@@ -7386,12 +7386,19 @@ def convert(value: t.Any, copy: bool = False) -> Expression:
...
         return HexString(this=value.hex())
     if isinstance(value, datetime.datetime):
-        datetime_literal = Literal.string(
-            (value if value.tzinfo else value.replace(tzinfo=datetime.timezone.utc)).isoformat(
-                sep=" "
-            )
-        )
-        return TimeStrToTime(this=datetime_literal)
+        if not value.tzinfo:
+            value = value.replace(tzinfo=datetime.timezone.utc)
```
tobymao, line 7390: why are you always defaulting to utc? if there's a UTC, do we want it to have a time zone?

erindru: that's what the original code did?

`(value if value.tzinfo else value.replace(tzinfo=datetime.timezone.utc))`

tobymao: yea, but now you're distinguishing between timezones and not, right? this didn't exist in the pr.evious code

sqlglot/dialects/clickhouse.py
```diff
@@ -105,6 +106,27 @@ def _delta_sql(self: Generator, expression: DATEΤΙΜΕ_DELTA) -> str:
     return _delta_sql
 
 
+def _timestrtotime_sql(self: ClickHouse.Generator, expression: exp.TimeStrToTime):
+    tz = expression.args.get("timezone")
+    datatype = exp.DataType.build(dtype=exp.DataType.Type.TIMESTAMP)
```
georgesittas, line 111: Nit: style consistency
```suggestion
    datatype = exp.DataType.build(exp.DataType.Type.TIMESTAMP)
```

sqlglot/dialects/clickhouse.py
```diff
@@ -105,6 +106,27 @@ def _delta_sql(self: Generator, expression: DATEΤΙΜΕ_DELTA) -> str:
...
+    ts = expression.this
+    if tz:
+        # build a datatype that encodes the timezone as a type parameter, eg DateTime('America/Los_Angeles')
+        datatype = exp.DataType.build(
+            dtype=exp.DataType.Type.TIMESTAMPTZ,  # Type.TIMESTAMPTZ maps to DateTime
```
georgesittas, line 116: Ditto:
```suggestion
            exp.DataType.Type.TIMESTAMPTZ,  # Type.TIMESTAMPTZ maps to DateTime
```

sqlglot/expressions.py
erindru: The previous code would force all naive datetimes to be treated as UTC. I think this is a reasonable default because in my opinion it leads to the least amount of surprises when doing things like additional casts to other types or running the values through other date functions.

Wouldnt anyone relying on the existing behaviour also be relying on this?

sqlglot/dialects/clickhouse.py
```diff
@@ -105,6 +106,27 @@ def _delta_sql(self: Generator, expression: DATEΤΙΜΕ_DELTA) -> str:
...
+            expressions=[exp.DataTypeParam(this=tz)],
+        )
+        # strip the timezone out of the literal, eg turn '2020-01-01 12:13:14-08:00' into '2020-01-01 12:13:14'
+        # this is because Clickhouse encodes the timezone as a data type parameter and throws an error if it's part of the timestamp string
+        ts_without_tz = (
+            datetime.datetime.fromisoformat(expression.this.name)
```
georgesittas, line 122: You could simplify this since you set `ts = ...` in line 112. Is this safe though? What if the first argument to `TimeStrToTime` is a column and not a string literal? I think what you're trying to do here is only valid for literals.

```suggestion
            datetime.datetime.fromisoformat(ts.name)
```

tests/test_expressions.py
erindru: I put the `zoneinfo` test behind a version gate and used `pytz` in the main test instead.

As an aside, is there a reason why we still support Python 3.7? The upstream Python project [doesnt even support](https://devguide.python.org/versions/) Python 3.7

sqlglot/dialects/dialect.py
```diff
@@ -1244,7 +1244,22 @@ def right_to_substring_sql(self: Generator, expression: exp.Left) -> str:
...
+    )
+
+    # dont re-cast if the expression is already a cast to the correct type
+    if isinstance(expression.this, exp.Cast):
+        cast_to_type: exp.DataType.Type = expression.this.to.this
+        types_are_equivalent = self.TYPE_MAPPING.get(
+            cast_to_type, cast_to_type
+        ) == self.TYPE_MAPPING.get(datatype, datatype)
+        if cast_to_type == datatype or types_are_equivalent:
+            return self.sql(expression.this)
```
georgesittas, line 1260: Should we push this logic down to `exp.cast`? It looks reasonable and we're already doing something similar, just not dialect-aware.

sqlglot/dialects/tsql.py
```diff
@@ -339,6 +340,16 @@ def _json_extract_sql(
     return self.func("ISNULL", json_query, json_value)
 
 
+def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
+    sql = timestrtotime_sql(self, expression)
+    if expression.args.get("timezone"):
+        # If there is a timezone, produce an expression like:
+        # CAST('2020-01-01 12:13:14-08:00' AS DATETIMEOFFSET) AT TIME ZONE 'UTC'
+        # If you dont have AT TIME ZONE 'UTC', wrapping that expression in another cast back to DATETIME2 just drops the timezone information
```
georgesittas, line 348: Why does it drop the timezone information? What does the `AT TIME ZONE 'UTC'` achieve here?

sqlglot/expressions.py
erindru: I mean its still _kind_ of naive because this change uses the original logic. Any UTC dates are cast to `TIMESTAMP`, not `TIMESTAMPTZ`, so the timezone is only taken into account for non-UTC

tests/test_expressions.py
georgesittas: Because some folks are still using sqlglot with python 3.7 due to other packages that need this version. We did discuss getting rid of it at some point, but not sure if/when that will happen.

sqlglot/dialects/tsql.py
erindru: Because it's Microsoft so therefore insane. Based on my testing:

```
select CAST('2020-01-01 12:13:14-08:00' AS DATETIMEOFFSET)
> 2020-01-01 12:13:14.000 -0800
```
ok, fine, its showing the timezone. What if I want it as `DATETIME2` instead?

```
select CAST(CAST('2020-01-01 12:13:14-08:00' AS DATETIMEOFFSET) AS DATETIME2)
> 2020-01-01 12:13:14.000
```
_Of course_ it just strips the timezone off, totally logical

```
select CAST(CAST('2020-01-01 12:13:14-08:00' AS DATETIMEOFFSET) AT TIME ZONE 'UTC' AS DATETIME2)
> 2020-01-01 20:13:14.000
```
Magic sauce to get the expected value

georgesittas: Fun :-)

FWIW I did find another [thread](https://learn.microsoft.com/en-us/answers/questions/659376/converting-datetimeoffset-to-datetime-with-adjustm) where the `AT TIME ZONE 'UTC'` trick is mentioned. The problem stated by the OP looks similar to this one, so... Seems legit?

erindru: Heh, that thread is even worse. They had to convert to UTC first before converting to the actual timezone they wanted.

I guess this makes the conversion to UTC automatic by default so things downstream should Just Work™

tests/test_expressions.py
```diff
@@ -791,6 +792,7 @@ def test_properties_from_dict(self):
 
     def test_convert(self):
         from collections import namedtuple
+        import pytz
```
georgesittas, line 795: Does this need to be a dev. dependency now?

sqlglot/dialects/clickhouse.py
erindru: I looked at where `TimeStrToTime` was being constructed and I couldnt see any cases or tests where the first argument was not a String literal (I guess that's what makes it `TimeStr` and not `Time`).

However, should I add a guard just in case?

georgesittas: Yep, imo we shouldn't make this assumption. You could have a column as the first argument, which only takes string values - perfectly valid for what `TimeStrToTime` represents.

tests/test_expressions.py
erindru: Well, it gets pulled in by `pandas`. Is it better to make it explict?

sqlglot/dialects/clickhouse.py
georgesittas: One example where we can build a `TimeStrToTime` without necessarily having a literal argument is in T-SQL's `_build_date_delta` [function](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/tsql.py#L240-L244).

tests/test_expressions.py
georgesittas: Ah I see. I guess it doesn't hurt being explicit? Not sure what's the best practice here, but I can imagine if for whatever reason pandas removes this dependency in the future, we'd break as a side-effect.

sqlglot/expressions.py
tobymao: ok that's fine

sqlglot/expressions.py
```diff
@@ -5985,7 +5985,7 @@ class TimeStrToDate(Func):
 
 
 class TimeStrToTime(Func):
-    pass
+    arg_types = {"this": True, "timezone": False}
```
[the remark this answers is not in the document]
erindru: Ironically, I picked `timezone` because that's what `TimeToStr` was using :)

I'll change it and I guess a breaking change to `TimeToStr` isnt a problem?

sqlglot/dialects/dialect.py
erindru: How would that work? `exp.cast()` doesn't have access to the dialect's type mappings, right?

sqlglot/dialects/clickhouse.py
erindru: Gotcha, thanks for the example. I'll improve this

sqlglot/expressions.py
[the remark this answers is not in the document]
tobymao: it's not if we tag it with !

sqlglot/dialects/dialect.py
georgesittas: It has through `**opts`, you can pass in the dialect and do:

```python
Dialect.get_or_raise(dialect).generator_class.TYPE_MAPPING
```

georgesittas: To clarify, I'd change `**opts` to `dialect: DialectType = None, **opts` here.

tests/test_expressions.py
erindru: Added

sqlglot/dialects/clickhouse.py
erindru: I've added a check for `exp.Literal` and also added a test that uses a column instead of a literal in combination with a timezone to hit this branch

sqlglot/dialects/dialect.py
erindru: I've had a go at implementing this. I noticed that there was no top-level import for `Dialect` and it was always a local function-scoped import wherever it was used, so I kept this pattern

sqlglot/expressions.py
erindru: Actually, I had to remove this assumption because engines like BigQuery dont support a timezone part in the string literal (eg `+08:00`). It was generating:

```
select CAST('2023-01-01 00:00:00+00:00' AS DATETIME)
```

for BigQuery which is invalid, it throws ` Invalid datetime string "2023-01-01 00:00:00+00:00"`

So now:
 - naive datetime objects remain naive, these will be the `TIMESTAMP` type
 - anything with `tzinfo` set will be `TIMESTAMPTZ`, even if `tzinfo` is just UTC

This causes the above expression to be generated as:
```
CAST('2023-01-01 12:13:14+00:00' AS TIMESTAMP)
```

for BigQuery, which is correct

tobymao: cool

tobymao (approved):

VaggelisD (approved):

PR #3942, opened 2024-08-20
Fix!: Redshift date format

erindru:
The Redshift dialect has a subtle bug in its timestamp formatting.

When mapping strftime strings like `%Y-%m-%d %H:%M:%S`,  the Redshift equivalent is being generated as:
```
SELECT TO_CHAR(timestamp '2023-01-01 15:13:14', 'YYYY-MM-DD HH:MI:SS.US')
> '2023-01-01 03:13:14.000000'
```
which wraps around the hours because `%H` (24-hour) is being mapped to `HH` (12-hour) [instead of](https://docs.aws.amazon.com/redshift/latest/dg/r_FORMAT_strings.html) `HH24`.

The correct format string is:
```
SELECT TO_CHAR(timestamp '2023-01-01 15:13:14', 'YYYY-MM-DD HH24:MI:SS.US')
> '2023-01-01 15:13:14.000000'
```

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -48,12 +48,9 @@ class Redshift(Postgres):
     HEX_LOWERCASE = True
     HAS_DISTINCT_ARRAY_CONSTRUCTORS = True
 
-    TIME_FORMAT = "'YYYY-MM-DD HH:MI:SS'"
-    TIME_MAPPING = {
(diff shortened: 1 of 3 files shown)


tobymao (approved):

georgesittas (approved):
Nice catch!

VaggelisD (approved):

PR #3951, opened 2024-08-22
Fix!: mysql/tsql datetime precision, formatting, exp.AtTimeZone

erindru:
## MySQL

1. Fix DATETIME subsecond precision on `exp.TimeStrToTime` when generating casts for literals

```
>>> parse_one("TIME_STR_TO_TIME('2020-01-01 12:13:14.123456+00:00')").dialect("mysql")

-- BEFORE
select CAST(TIMESTAMP('2020-01-01 12:13:14.123456') AS DATETIME)
> 2020-01-01 12:13:14

-- AFTER
select CAST(TIMESTAMP('2020-01-01 12:13:14.123456') AS DATETIME(6))
> 2020-01-01 12:13:14.123456
```

2. Raise a warning in the MySQL generator if it encounters an `exp.AtTimeZone` node.
```
>>> parse_one("select foo at time zone 'UTC' from tbl").sql(dialect="mysql")

-- BEFORE
'SELECT foo AT TIME ZONE 'UTC' FROM tbl'
! You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'TIME ZONE 'UTC'

-- AFTER
AT TIME ZONE is not supported by MySQL
'SELECT foo FROM tbl'
```

## TSQL

1. Fix '%f' strftime mapping for microseconds. When mapping `%Y-%m-%d %H:%M:%S.%f`:
```
>>> parse_one("TIME_TO_STR(a, '%Y-%m-%d %H:%M:%S.%f')").sql(dialect="tsql")

-- BEFORE
select FORMAT(a, 'yyyy-MM-dd HH:mm:ss.S')
> 2023-01-01 12:13:14.S

-- AFTER
select FORMAT(a, 'yyyy-MM-dd HH:mm:ss.ffffff')
> 2023-01-01 12:13:14.123456
```

2. Fix boolean values in `VALUES()` expression.
```
>>> parse_one("SELECT val FROM (VALUES ((TRUE), (FALSE), (NULL))) AS t(val)").sql(dialect="tsql")

-- BEFORE
'SELECT val FROM (VALUES (((1 = 1)), ((1 = 0)), (NULL))) AS t(val)'
!SQL Error [102] [S0001]: Incorrect syntax near '='.

-- AFTER
'SELECT val FROM (VALUES ((1), (0), (NULL))) AS t(val)'
```

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1246,26 +1246,22 @@ def right_to_substring_sql(self: Generator, expression: exp.Left) -> str:
 def timestrtotime_sql(
     self: Generator,
     expression: exp.TimeStrToTime,
-    include_precision: bool | t.Callable[[exp.DataType.Type], bool] = False,
+    include_precision: bool = False,
 ) -> str:
-    datatype = (
+    datatype = exp.DataType.build(
         exp.DataType.Type.TIMESTAMPTZ
         if expression.args.get("zone")
         else exp.DataType.Type.TIMESTAMP
     )
 
-    if callable(include_precision):
-        include_precision = include_precision(datatype)
-
-    datatype_built = exp.DataType.build(datatype)
     if isinstance(expression.this, exp.Literal) and include_precision:
         precision = subsecond_precision(expression.this.name)
         if precision > 0:
-            datatype_built = exp.DataType.build(
-                datatype, expressions=[exp.DataTypeParam(this=exp.Literal.number(precision))]
+            datatype = exp.DataType.build(
+                datatype.this, expressions=[exp.DataTypeParam(this=exp.Literal.number(precision))]
             )
 
-    return self.sql(exp.cast(expression.this, datatype_built, dialect=self.dialect))
+    return self.sql(exp.cast(expression.this, datatype, dialect=self.dialect))
 
 
 def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -11,7 +11,7 @@ from sqlglot.generator import Generator
 from sqlglot.helper import AutoName, flatten, is_int, seq_get, subclasses
 from sqlglot.jsonpath import JSONPathTokenizer, parse as parse_json_path
 from sqlglot.parser import Parser
-from sqlglot.time import TIMEZONES, format_time
+from sqlglot.time import TIMEZONES, format_time, subsecond_precision
 from sqlglot.tokens import Token, Tokenizer, TokenType
 from sqlglot.trie import new_trie
 
@@ -1243,14 +1243,29 @@ def right_to_substring_sql(self: Generator, expression: exp.Left) -> str:
     )
 
 
-def timestrtotime_sql(self: Generator, expression: exp.TimeStrToTime) -> str:
+def timestrtotime_sql(
+    self: Generator,
+    expression: exp.TimeStrToTime,
+    include_precision: bool | t.Callable[[exp.DataType.Type], bool] = False,
+) -> str:
     datatype = (
         exp.DataType.Type.TIMESTAMPTZ
         if expression.args.get("zone")
         else exp.DataType.Type.TIMESTAMP
     )
 
-    return self.sql(exp.cast(expression.this, datatype, dialect=self.dialect))
+    if callable(include_precision):
+        include_precision = include_precision(datatype)
+
+    datatype_built = exp.DataType.build(datatype)
+    if isinstance(expression.this, exp.Literal) and include_precision:
+        precision = subsecond_precision(expression.this.name)
+        if precision > 0:
+            datatype_built = exp.DataType.build(
+                datatype, expressions=[exp.DataTypeParam(this=exp.Literal.number(precision))]
+            )
+
+    return self.sql(exp.cast(expression.this, datatype_built, dialect=self.dialect))
 
 
 def datestrtodate_sql(self: Generator, expression: exp.DateStrToDate) -> str:
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -686,9 +686,6 @@ class MySQL(Dialect):
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
             exp.ArrayAgg: rename_func("GROUP_CONCAT"),
-            exp.AtTimeZone: lambda self, e: self.func(
-                "CONVERT_TZ", e.this, exp.var("@@global.time_zone"), e.args.get("zone")
-            ),
             exp.CurrentDate: no_paren_current_date_sql,
             exp.DateDiff: _remove_ts_or_ds_to_date(
                 lambda self, e: self.func("DATEDIFF", e.this, e.expression), ("this", "expression")
@@ -1216,3 +1213,7 @@ class MySQL(Dialect):
             dt = expression.args.get("timestamp")
 
             return self.func("CONVERT_TZ", dt, from_tz, to_tz)
+
+        def attimezone_sql(self, expression: exp.AtTimeZone) -> str:
+            self.unsupported("AT TIME ZONE is not supported by MySQL")
+            return self.sql(expression.this)
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -735,7 +735,7 @@ class MySQL(Dialect):
             exp.TimeStrToTime: lambda self, e: timestrtotime_sql(
                 self,
                 e,
-                include_precision=lambda datatype: datatype != exp.DataType.Type.TIMESTAMPTZ,
+                include_precision=not e.args.get("zone"),
             ),
             exp.TimeToStr: _remove_ts_or_ds_to_date(
                 lambda self, e: self.func("DATE_FORMAT", e.this, self.format_time(e))
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -25,6 +25,7 @@ from sqlglot.dialects.dialect import (
     strposition_to_locate_sql,
     unit_to_var,
     trim_sql,
+    timestrtotime_sql,
 )
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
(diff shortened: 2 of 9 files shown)


sqlglot/time.py
```diff
@@ -661,3 +662,24 @@ def format_time(
         "Zulu",
     )
 }
+
+
+def subsecond_precision(timestamp_literal: str) -> int:
+    # Given an ISO-8601 timestamp literal, eg '2023-01-01 12:13:14.123456+00:00'
```
georgesittas, line 668: Let's use a docstring instead of a comment.

sqlglot/dialects/tsql.py
```diff
@@ -983,7 +983,10 @@ def setitem_sql(self, expression: exp.SetItem) -> str:
             return super().setitem_sql(expression)
 
         def boolean_sql(self, expression: exp.Boolean) -> str:
-            if type(expression.parent) in BIT_TYPES:
+            if (
+                type(expression.parent) in BIT_TYPES
+                or expression.find_ancestor(exp.Values) is not None
```
georgesittas, line 988: You can make the search a bit more efficient by stopping at the first `Select`:
```suggestion
                or isinstance(expression.find_ancestor(exp.Values, exp.Select), exp.Values)
```

sqlglot/dialects/mysql.py
```diff
@@ -699,6 +700,9 @@ class Generator(generator.Generator):
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
             exp.ArrayAgg: rename_func("GROUP_CONCAT"),
+            exp.AtTimeZone: lambda self, e: self.func(
+                "CONVERT_TZ", e.this, exp.var("@@global.time_zone"), e.args.get("zone")
+            ),
```
georgesittas, line 705: I'm not sure if this assumption is safe. It's better to call `self.unsupported` and avoid generating something wrong vs silently corrupting data if that's the tradeoff we're looking at. Or, alternatively, you could maybe try configuring this using a dialect-level setting, thus opening this transpilation path conditionally.

sqlglot/dialects/dialect.py
```diff
@@ -1243,14 +1243,29 @@ def right_to_substring_sql(self: Generator, expression: exp.Left) -> str:
     )
 
 
-def timestrtotime_sql(self: Generator, expression: exp.TimeStrToTime) -> str:
+def timestrtotime_sql(
+    self: Generator,
+    expression: exp.TimeStrToTime,
+    include_precision: bool | t.Callable[[exp.DataType.Type], bool] = False,
```
georgesittas, line 1249: The callable here looks a bit overkill, can't we just do `include_precision=not e.args.get("zone")` for mysql and keep it a simple boolean? It looks like we only have the mysql case and the trino one, which is always true.

sqlglot/dialects/trino.py
```diff
@@ -21,6 +21,7 @@ class Generator(Presto.Generator):
             exp.ArraySum: lambda self,
             e: f"REDUCE({self.sql(e, 'this')}, 0, (acc, x) -> acc + x, acc -> acc)",
             exp.Merge: merge_without_target_sql,
+            exp.TimeStrToTime: lambda self, e: timestrtotime_sql(self, e, include_precision=True),
```
georgesittas, line 24: Should this go to `Presto` instead?

sqlglot/dialects/dialect.py
erindru: The reason why this supports a lambda is because in the MySQL case, we only want to do it if the datatype that would be returned is `Type.TIMESTAMP` (which transpiles to `DATETIME` which needs the precision to be part of the datatype).

We dont want to do it if the datatype that would be returned is `Type.TIMESTAMPTZ` (which transpiles to `TIMESTAMP` which doesnt support a type parameter for the precision).

We also dont know what type will be returned at call time without repeating the logic within the method.

The problem is that the `_sql` methods return a string which means you cant wrap them and modify the result without re-parsing the SQL

sqlglot/dialects/trino.py
erindru: I thought about that but decided not to.

The reason is that the [Presto docs](https://prestodb.io/docs/current/language/types.html#date-and-time) make no mention of `TIMESTAMP(P)` or `TIMESTAMP(P) WITH TIME ZONE` the way the [Trino docs](https://trino.io/docs/current/language/types.html#timestamp-p) do

georgesittas: We could test against Presto and verify if it doesn't support, in which case this is fine

sqlglot/dialects/dialect.py
georgesittas: You'll always have `datatype = exp.DataType.Type.TIMESTAMPTZ` when there's a `zone` arg present though, right? So my point is since you have this knowledge you can calculate the `precision` boolean right in the `TRANSFORMS` entry. Or maybe I'm missing something?

sqlglot/dialects/mysql.py
erindru: I think you're right. The problem I was trying to solve was constructing a "generic" AST using SQLGlot directly and then having it render appropriately across the database engines supported by SQLMesh. 

`exp.AtTimeZone` works for every dialect except MySQL, so I wanted to see if it could be made to work without introducing special hacks to replace that AST node with something else just for MySQL.

But I think if the concept isn't directly translatable, then you're right, it's better to raise `unsupported` to prevent surprises

erindru: Actually, I notice the Spark dialect transpiles `exp.AtTimeZone` to `FROM_UTC_TIMESTAMP`. Isn't that making an assumption about the source timezone?

georgesittas: Ah, maybe, need to double check that but it seems like it was [added a long time ago](https://github.com/tobymao/sqlglot/commit/80401ddcfe22b13922a51aaf0d1a556af246a277). Perhaps we should fix it while we're at it if it's problematic?

cc @tobymao

sqlglot/dialects/dialect.py
erindru: Yes, I guess we can rely on our knowledge of how `timestrtotime_sql` is implemented since this only affects one thing in one place

I'll update it

sqlglot/dialects/trino.py
erindru: Ok, I fired up a local Presto.
```
> select cast(timestamp '2023-01-01 12:13:14.123456' as timestamp(3))
!Query failed (#20240822_224449_00006_qn5fd): line 1:8: Unknown type: timestamp(3)
```

On Trino:
```
> select cast(timestamp '2023-01-01 12:13:14.123456' as timestamp(3))
2023-01-01 12:13:14.123000
```

So it looks like the Presto docs are correct

sqlglot/dialects/mysql.py
tobymao: sure if it's wrong we should fix it

VaggelisD:
Sorry for the unrelated mention, used the wrong #..

sqlglot/dialects/mysql.py
erindru: Hmm, I dug deeper into the Spark one at the reason it generates `exp.AtTimeZone` to `FROM_UTC_TIMESTAMP` is because it _parses_ `FROM_UTC_TIMESTAMP`  into `exp.AtTimeZone`.

I took a quick look at parsing `FROM_UTC_TIMESTAMP` into `exp.FromTimeZone` instead, but then I noticed that `TO_UTC_TIMESTAMP` is parsed into `exp.FromTimeZone` and `exp.FromTimeZone` is generated to `TO_UTC_TIMESTAMP`. So if I parsed `FROM_UTC_TIMESTAMP` into `exp.FromTimeZone`, it would be generated as  `TO_UTC_TIMESTAMP` which is wrong.

In addition, I did some tests around what Spark does with `TIMESTAMP` values and compared it to Postgres. As a baseline, I focused on the following:

```
parse_one("select '2023-01-01 12:13:14-08:00'::timestamp at time zone 'UTC'", dialect="postgres").sql(dialect="spark")
"SELECT FROM_UTC_TIMESTAMP(CAST('2023-01-01 12:13:14-08:00' AS TIMESTAMP), 'UTC')"
```

On Postgres, the returned values are consistent regardless of the session timezone:
```
set time zone 'UTC';
select '2023-01-01 12:13:14-08:00'::timestamp at time zone 'UTC'; --2023-01-01 12:13:14.000000 +0000

set time zone 'America/Los_Angeles';
select '2023-01-01 12:13:14-08:00'::timestamp at time zone 'UTC'; --2023-01-01 12:13:14.000000 +0000
```

On Spark, the returned values vary depending on the session timezone:
```
set time zone 'UTC';
SELECT FROM_UTC_TIMESTAMP(CAST('2023-01-01 12:13:14-08:00' AS TIMESTAMP), 'UTC') --2023-01-01 20:13:14.000000

set time zone 'America/Los_Angeles';
SELECT FROM_UTC_TIMESTAMP(CAST('2023-01-01 12:13:14-08:00' AS TIMESTAMP), 'UTC') --2023-01-01 12:13:14.000000
```

The transpilation gives different results because how the `CAST` on the spark side is interpreted depends on the session timezone.

Since it's not a straightforward fix, perhaps it can be addressed in a future PR?

erindru: For MySQL, i've updated it to emit:
```
self.unsupported("AT TIME ZONE is not supported by MySQL")
```
if the generator encounters an `exp.AtTimeZone` node

georgesittas: Sounds good 👍

sqlglot/dialects/tsql.py
```diff
@@ -983,7 +983,9 @@ def setitem_sql(self, expression: exp.SetItem) -> str:
             return super().setitem_sql(expression)
 
         def boolean_sql(self, expression: exp.Boolean) -> str:
-            if type(expression.parent) in BIT_TYPES:
+            if type(expression.parent) in BIT_TYPES or isinstance(
+                expression.find_ancestor(exp.Values, exp.Select), exp.Values
+            ):
```
georgesittas, line 988: This is probably ok, but I'm a little worried we may transform booleans in the context of automatically-generated conjunctions/disjunctions into 1s and 0s, leading to incorrect T-SQL syntax like `1 AND 0`.

I'm not sure if or how we can arrive at that state, but it may not be a bad idea to guard against it by checking if e.g. the immediate parent is a `Paren`.

georgesittas (approved):
LGTM, nice work! Left a final comment to consider.

sqlglot/dialects/tsql.py
erindru: I'm not clear on exactly what query would trigger that.

It would only happen if somehow was a condition like `((TRUE) OR (FALSE))` somehow nested within a `VALUES(...)` statement, right? Since there are no `exp.Paren` objects anywhere within the AST of the queries this change was made for:

```
INSERT INTO tbl (foo, bar) VALUES (True, False)
```
and
```
SELECT * FROM (VALUES(True, False)) as t (foo, bar)
```

checking if the immediate parent of the `exp.Boolean` is an `exp.Paren` would always return False?

However if someone finds a way to trigger incorrect behaviour then i'll happily fix it since they will have an example we can add to the test suite

PR #3958, opened 2024-08-22
fix: Parse LTRIM/RTRIM functions as positional exp.Trim

### LINKED ISSUE #3957: Failed to transpile Oracle rtrim/ltrim to ClickHouse

Oracle [rtrim](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/RTRIM.html) function is not transpiled as expected 

```
from sqlglot.dialects import Oracle, ClickHouse
sqlglot.transpile("select rtrim('(   Hello, world!   )', ')')", read=Oracle, write=ClickHouse, comments=False)
["SELECT rtrim('(   Hello, world!   )', ')')"]
```

Expected result

```
select trim(TRAILING ')' from '(   Hello, world!   )')
```

Similar issue for [ltrim](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/LTRIM.html). 

**Official Documentation**
CH trim function syntax https://clickhouse.com/docs/en/sql-reference/functions/string-functions#trim

VaggelisD:
Fixes #3957

Docs
------
[Oracle RTRIM](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/RTRIM.html) | [Clickhouse TRIM](https://clickhouse.com/docs/en/sql-reference/functions/string-functions#trim)

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -20,6 +20,7 @@ from sqlglot.dialects.dialect import (
     var_map_sql,
     timestamptrunc_sql,
     unit_to_var,
+    trim_sql,
 )
 from sqlglot.generator import Generator
 from sqlglot.helper import is_int, seq_get
@@ -875,6 +876,7 @@ class ClickHouse(Dialect):
             exp.SHA2: sha256_sql,
             exp.UnixToTime: _unix_to_time_sql,
             exp.TimestampTrunc: timestamptrunc_sql(zone=True),
+            exp.Trim: trim_sql,
             exp.Variance: rename_func("varSamp"),
             exp.SchemaCommentProperty: lambda self, e: self.naked_property(e),
             exp.Stddev: rename_func("stddevSamp"),
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1295,7 +1295,7 @@ def trim_sql(self: Generator, expression: exp.Trim) -> str:
     collation = self.sql(expression, "collation")
 
     # Use TRIM/LTRIM/RTRIM syntax if the expression isn't database-specific
-    if not remove_chars and not collation:
+    if not remove_chars:
         return self.trim_sql(expression)
 
     trim_type = f"{trim_type} " if trim_type else ""
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -24,6 +24,7 @@ from sqlglot.dialects.dialect import (
     rename_func,
     strposition_to_locate_sql,
     unit_to_var,
+    trim_sql,
 )
 from sqlglot.helper import seq_get
 from sqlglot.tokens import TokenType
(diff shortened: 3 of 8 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -875,6 +876,7 @@ class Generator(generator.Generator):
             exp.SHA2: sha256_sql,
             exp.UnixToTime: _unix_to_time_sql,
             exp.TimestampTrunc: timestamptrunc_sql(zone=True),
+            exp.Trim: trim_sql,
```
georgesittas, line 879: It would be good to add some roundtrip tests for clickhouse as well.

sqlglot/dialects/dialect.py
```diff
@@ -1295,7 +1295,7 @@ def trim_sql(self: Generator, expression: exp.Trim) -> str:
     collation = self.sql(expression, "collation")
 
     # Use TRIM/LTRIM/RTRIM syntax if the expression isn't database-specific
-    if not remove_chars and not collation:
```
georgesittas, line 1298: Curious about this term in the condition - did you find any context around it?

georgesittas (approved):

sqlglot/dialects/dialect.py
VaggelisD: I actually didn't find much regarding collation, at least for the relevant dialects such as [Oracle](https://docs.oracle.com/cd/B12037_01/server.101/b10759/functions184.htm), [Postgres](https://www.postgresql.org/docs/9.1/functions-string.html), [MySQL](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_trim).  We do have a Postgres test for `TRIM(... COLLATE utf8_bin)` (although this specific collation doesn't work for me locally) which I verified is a valid query. 

I think the condition there was probably not needed for Postgres; If we don't have `remove_chars` (the default char is a space), then the RTT versions are semantically equivalent:

```
Input: SELECT TRIM(LEADING ' XXX ' COLLATE "de_DE");
SQLGlot RTT: SELECT LTRIM(' XXX ' COLLATE "de_DE")


Input: SELECT TRIM(' XXX ' COLLATE "de_DE")
SQLGlot RTT: SELECT TRIM(' XXX ' COLLATE "de_DE")
```



Source PRs: https://github.com/tobymao/sqlglot/pull/432, https://github.com/tobymao/sqlglot/pull/934/files

georgesittas: Cool, thanks for double checking.

cpcloud:
This PR seems to have broken `LTRIM` and `RTRIM` that strip multiple characters in Oracle.


`TRIM` doesn't allow multiple characters, but `LTRIM` and `RTRIM` do.

Unfortunately the docs don't state this requirement explicitly anywhere: https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/TRIM.html#GUID-00D5C77C-19B1-4894-828F-066746235B03

If you try to run any `TRIM` call with more than one trim character then you'll get this error message:

```
E   oracledb.exceptions.DatabaseError: ORA-30001: trim set should have only one character
E   Help: https://docs.oracle.com/error-help/db/ora-30001/
```

VaggelisD:
That checks out:

```SQL
SELECT TRIM(LEADING 'ab' FROM 'abc'), LTRIM('abc', 'ab')
       *
ERROR at line 1:
ORA-30001: trim set should have only one character
```

vs

```
SELECT LTRIM('abc', 'ab');
'c'
```

Also relevant [SO thread](https://stackoverflow.com/questions/35098149/cant-trim-the-string-in-oracle).

So for Oracle, we can probably do the following:
- If we have `LEADING` or `TRAILING` positional specifiers and multiple remove characters, we should force `LTRIM` and `RTRIM` generation respectively
- For other cases (e.g. `BOTH` specifier, 0 or 1 remove chars) we can keep the existing representation: `TRIM(<position> <remove char> FROM <string>`

georgesittas:
Let's just keep it simple and always do `LTRIM` and `RTRIM` for `LEADING` and `TRAILING`, respectively. No need to make it complicated.

PR #3960, opened 2024-08-23
fix: include dialect when parsing inside cast

eakmanrq:
Fixes a regression introduced by: https://github.com/tobymao/sqlglot/pull/3938

### DIFF
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -451,6 +451,12 @@ class TestExpressions(unittest.TestCase):
         self.assertEqual(len(list(recast.find_all(exp.Cast))), 2)
         self.assertEqual(recast.sql(), "CAST(CAST(x AS DATE) AS VARCHAR)")
 
+        # check that dialect is used when casting strings
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


tests/test_expressions.py
```diff
@@ -451,6 +451,12 @@ def test_cast(self):
         self.assertEqual(len(list(recast.find_all(exp.Cast))), 2)
         self.assertEqual(recast.sql(), "CAST(CAST(x AS DATE) AS VARCHAR)")
 
+        # check that dialect is used when casting strings
+        self.assertEqual(
+            exp.cast("x", to="regtype", dialect="postgres").sql(), "CAST(x AS REGTYPE)"
+        )
+        self.assertEqual(exp.cast("`x`", to="date", dialect="hive").sql(), 'CAST("x" AS DATE)')
```
tobymao, line 458: these will fail without your fix?

tobymao (approved):

tests/test_expressions.py
eakmanrq: Yes top is for the parsing of data type and bottom is for parsing of column (due to weird quotes in hive). Both will fail without fix.

PR #3966, opened 2024-08-24
fix(starrocks): exp.Unnest transpilation

### LINKED ISSUE #3962: Incorrect translation in table function unnest syntax

The `unnest` syntax cannot be converted correctly, related to starrocks and bigquery. 

In my case, I need to convert `cross join unnest` to `LATERAL VIEW EXPLODE`.

The following sql: 
```
-- sql1
SELECT id, scores, unnest FROM values (1, [80,85,87]),(2, [77, null, 89]) as student_score(id, scores), unnest(scores) AS unnest;
-- sql2
SELECT id, scores, unnest FROM values (1, [80,85,87]),(2, [77, null, 89]) as student_score(id, scores), unnest(scores);
-- sql3
SELECT id, unnest.type, unnest.scores
FROM values ("typeA,typeB", [80,85,88]),("typeA,typeB,typeC", [87,90,95]) as example_table(type, scores), unnest(split(type, ","), scores) AS unnest(type,scores);
```

When we `transpile` in case sql1 and sql2, we will get a result that loses the unnest function (all dialects):
```
SELECT id, scores, t.unnest FROM VALUES (1, ARRAY(80, 85, 87)), (2, ARRAY(77, NULL, 89)) AS student_score(id, scores)
```

Expected value:
```
FROM values ​​(1, [80,85,87]),(2, [77, null, 89]) as student_score(id, scores)
LATERAL VIEW EXPLODE(scores) u AS unnest;
```

**Official Documentation**
> https://docs.starrocks.io/docs/using_starrocks/Lateral_join/
> https://cloud.google.com/bigquery/docs/arrays?hl=en

hellozepp:
Fixes #3962 

This PR adds support for the following:
 - Default initialize `exp.UNNEST` to include "unnest" as the default table alias if not specified
 - Added expression type of `exp.Inline`
 - use arrays_zip to merge multiple Lateral views
 - Added some ut to validate unnest transpile

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -191,6 +191,16 @@ def _build_with_ignore_nulls(
     return _parse
 
 
+def _select_sql(self: Hive.Generator, expression: exp.Expression) -> str:
+    return transforms.preprocess(
+        [
+            transforms.eliminate_qualify,
+            transforms.eliminate_distinct_on,
+            partial(transforms.unnest_to_explode, unnest_using_arrays_zip=False, generator=self),
+        ]
+    )(self, expression)
+
+
 class Hive(Dialect):
     ALIAS_POST_TABLESAMPLE = True
     IDENTIFIERS_CAN_START_WITH_DIGIT = True
@@ -482,13 +492,7 @@ class Hive(Dialect):
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
             exp.Group: transforms.preprocess([transforms.unalias_group]),
-            exp.Select: transforms.preprocess(
-                [
-                    transforms.eliminate_qualify,
-                    transforms.eliminate_distinct_on,
-                    partial(transforms.unnest_to_explode, unnest_using_arrays_zip=False),
-                ]
-            ),
+            exp.Select: _select_sql,
             exp.Property: _property_sql,
             exp.AnyValue: rename_func("FIRST"),
             exp.ApproxDistinct: approx_count_distinct_sql,
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import typing as t
+from functools import partial
 
 from sqlglot import exp, generator, parser, tokens, transforms
 from sqlglot.dialects.dialect import (
@@ -485,7 +486,7 @@ class Hive(Dialect):
                 [
                     transforms.eliminate_qualify,
                     transforms.eliminate_distinct_on,
-                    transforms.unnest_to_explode,
+                    partial(transforms.unnest_to_explode, unnest_using_arrays_zip=False),
                 ]
             ),
             exp.Property: _property_sql,
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -190,6 +190,13 @@ class Spark2(Hive):
 
         TRANSFORMS = {
             **Hive.Generator.TRANSFORMS,
+            exp.Select: transforms.preprocess(
+                [
+                    transforms.eliminate_qualify,
+                    transforms.eliminate_distinct_on,
+                    transforms.unnest_to_explode,
+                ]
+            ),
             exp.ApproxDistinct: rename_func("APPROX_COUNT_DISTINCT"),
             exp.ArraySum: lambda self,
             e: f"AGGREGATE({self.sql(e, 'this')}, 0, (acc, x) -> acc + x, acc -> acc)",
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -190,13 +190,6 @@ class Spark2(Hive):
 
         TRANSFORMS = {
             **Hive.Generator.TRANSFORMS,
-            exp.Select: transforms.preprocess(
-                [
-                    transforms.eliminate_qualify,
-                    transforms.eliminate_distinct_on,
-                    transforms.unnest_to_explode,
-                ]
-            ),
             exp.ApproxDistinct: rename_func("APPROX_COUNT_DISTINCT"),
             exp.ArraySum: lambda self,
             e: f"AGGREGATE({self.sql(e, 'this')}, 0, (acc, x) -> acc + x, acc -> acc)",
@@ -236,6 +229,13 @@ class Spark2(Hive):
                 e.expression,
                 e.args["replacement"],
                 e.args.get("position"),
+            ),
+            exp.Select: transforms.preprocess(
+                [
+                    transforms.eliminate_qualify,
+                    transforms.eliminate_distinct_on,
+                    transforms.unnest_to_explode,
+                ]
             ),
             exp.StrToDate: _str_to_date,
             exp.StrToTime: lambda self, e: self.func("TO_TIMESTAMP", e.this, self.format_time(e)),
--- a/sqlglot/dialects/starrocks.py
+++ b/sqlglot/dialects/starrocks.py
@@ -38,7 +38,7 @@ class StarRocks(MySQL):
                 if not alias:
                     alias = exp.TableAlias(columns=[exp.to_identifier("unnest")])
                     unnest.set("alias", alias)
-                if not alias.args.get("columns"):
+                elif not alias.args.get("columns"):
                     # Starrocks defaults to naming the UNNEST column as "unnest"
                     # if it's not otherwise specified
                     alias.set("columns", [exp.to_identifier("unnest")])
--- a/sqlglot/dialects/starrocks.py
+++ b/sqlglot/dialects/starrocks.py
@@ -36,12 +36,9 @@ class StarRocks(MySQL):
                 alias = unnest.args.get("alias")
 
                 if not alias:
-                    # Starrocks defaults to naming the table alias as "unnest"
-                    alias = exp.TableAlias(
-                        this=exp.to_identifier("unnest"), columns=[exp.to_identifier("unnest")]
-                    )
+                    alias = exp.TableAlias(columns=[exp.to_identifier("unnest")])
                     unnest.set("alias", alias)
-                if alias and not alias.args.get("columns"):
+                if not alias.args.get("columns"):
                     # Starrocks defaults to naming the UNNEST column as "unnest"
                     # if it's not otherwise specified
                     alias.set("columns", [exp.to_identifier("unnest")])
--- a/sqlglot/dialects/starrocks.py
+++ b/sqlglot/dialects/starrocks.py
@@ -36,7 +36,10 @@ class StarRocks(MySQL):
                 alias = unnest.args.get("alias")
 
                 if not alias:
-                    alias = exp.TableAlias(columns=[exp.to_identifier("unnest")])
+                    # Starrocks defaults to naming the table alias as "unnest"
+                    alias = exp.TableAlias(
+                        this=exp.to_identifier("unnest"), columns=[exp.to_identifier("unnest")]
+                    )
                     unnest.set("alias", alias)
                 elif not alias.args.get("columns"):
                     # Starrocks defaults to naming the UNNEST column as "unnest"
--- a/sqlglot/dialects/starrocks.py
+++ b/sqlglot/dialects/starrocks.py
@@ -35,6 +35,12 @@ class StarRocks(MySQL):
             if unnest:
                 alias = unnest.args.get("alias")
 
+                if not alias:
+                    # Starrocks defaults to naming the table alias as "unnest"
+                    alias = exp.TableAlias(
+                        this=exp.to_identifier("unnest"), columns=[exp.to_identifier("unnest")]
+                    )
+                    unnest.set("alias", alias)
                 if alias and not alias.args.get("columns"):
                     # Starrocks defaults to naming the UNNEST column as "unnest"
                     # if it's not otherwise specified
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5333,8 +5333,7 @@ class Explode(Func):
 
 # https://spark.apache.org/docs/latest/api/sql/#inline
 class Inline(Func):
-    arg_types = {"this": True, "expressions": False}
-    is_var_len_args = True
+    pass
 
 
 class ExplodeOuter(Explode):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5331,6 +5331,12 @@ class Explode(Func):
     is_var_len_args = True
 
 
+# https://spark.apache.org/docs/latest/api/sql/#inline
+class Inline(Func):
+    arg_types = {"this": True, "expressions": False}
+    is_var_len_args = True
+
+
 class ExplodeOuter(Explode):
     pass
 
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import typing as t
 
 from sqlglot import expressions as exp
+from sqlglot.expressions import Func
 from sqlglot.helper import find_new_name, name_sequence
 
 if t.TYPE_CHECKING:
@@ -299,19 +300,20 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
 def unnest_to_explode(
     expression: exp.Expression,
     unnest_using_arrays_zip: bool = True,
-    support_unnest_func: bool = False,
 ) -> exp.Expression:
     """Convert cross join unnest into lateral view explode."""
 
-    def handle_unnest(u: exp.Unnest, _has_multi_expr: bool) -> None:
+    def _unnest_zip_exprs(
+        u: exp.Unnest, unnest_exprs: t.List[t.Any], _has_multi_expr: bool
+    ) -> t.List[Func]:
         if unnest_using_arrays_zip and _has_multi_expr:
             # Use INLINE(ARRAYS_ZIP(...)) for multiple expressions
-            u.set(
-                "expressions",
-                [exp.Anonymous(this="ARRAYS_ZIP", expressions=u.expressions)],
-            )
+            zip_exprs: t.List[Func] = [exp.Anonymous(this="ARRAYS_ZIP", expressions=unnest_exprs)]
+            u.set("expressions", zip_exprs)
+            return zip_exprs
+        return unnest_exprs
 
-    def udtf_type(u: exp.Unnest, _has_multi_expr: bool):
+    def _udtf_type(u: exp.Unnest, _has_multi_expr: bool):
         return (
             exp.Posexplode
             if u.args.get("offset")
@@ -321,16 +323,16 @@ def unnest_to_explode(
     if isinstance(expression, exp.Select):
         from_ = expression.args.get("from")
 
-        if not support_unnest_func and from_ and isinstance(from_.this, exp.Unnest):
+        if from_ and isinstance(from_.this, exp.Unnest):
             unnest = from_.this
             alias = unnest.args.get("alias")
-            has_multi_expr = len(unnest.expressions) > 1
-            handle_unnest(unnest, has_multi_expr)
-            this, *expressions = unnest.expressions
+            exprs = unnest.expressions
+            has_multi_expr = len(exprs) > 1
+            this, *expressions = _unnest_zip_exprs(unnest, exprs, has_multi_expr)
 
             unnest.replace(
                 exp.Table(
-                    this=udtf_type(unnest, has_multi_expr)(
+                    this=_udtf_type(unnest, has_multi_expr)(
                         this=this,
                         expressions=expressions,
                     ),
@@ -346,26 +348,27 @@ def unnest_to_explode(
             unnest = join_expr.this if is_lateral else join_expr
 
             if isinstance(unnest, exp.Unnest):
-                alias = (
-                    join_expr.args.get("alias")
-                    if is_lateral and join_expr.args.get("alias")
-                    else unnest.args.get("alias")
-                )
+                if is_lateral:
+                    alias = join_expr.args.get("alias") or unnest.args.get("alias")
+                else:
+                    alias = unnest.args.get("alias")
+                exprs = unnest.expressions
                 # The number of unnest.expressions will be changed by handle_unnest, we need to record it here
-                has_multi_expr = len(unnest.expressions) > 1
-                handle_unnest(unnest, has_multi_expr)
+                has_multi_expr = len(exprs) > 1
+                exprs = _unnest_zip_exprs(unnest, exprs, has_multi_expr)
 
                 expression.args["joins"].remove(join)
 
-                for e, column in zip(unnest.expressions, alias.columns if alias else []):
+                alias_cols = alias.columns if alias else []
+                for e, column in zip(exprs, alias_cols):
                     expression.append(
                         "laterals",
                         exp.Lateral(
-                            this=udtf_type(unnest, has_multi_expr)(this=e),
+                            this=_udtf_type(unnest, has_multi_expr)(this=e),
                             view=True,
                             alias=exp.TableAlias(
                                 this=alias.this,  # type: ignore
-                                columns=alias.columns if unnest_using_arrays_zip else [column],  # type: ignore
+                                columns=alias_cols if unnest_using_arrays_zip else [column],  # type: ignore
                             ),
                         ),
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -313,7 +313,7 @@ def unnest_to_explode(
             return zip_exprs
         return unnest_exprs
 
-    def _udtf_type(u: exp.Unnest, _has_multi_expr: bool):
+    def _udtf_type(u: exp.Unnest, _has_multi_expr: bool) -> exp.Func:
         return (
             exp.Posexplode
             if u.args.get("offset")
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import typing as t
 
 from sqlglot import expressions as exp
-from sqlglot.expressions import Func
 from sqlglot.helper import find_new_name, name_sequence
 
 if t.TYPE_CHECKING:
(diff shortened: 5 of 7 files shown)


hellozepp:
I tried to add the logic of setting default value when there is no table alias, PTAL

I need to transpile the starrocks sql:
```
SELECT student, score, t.unnest FROM tests CROSS JOIN LATERAL UNNEST(scores)
```
Expected output:
```
SELECT student, score, unnest.unnest FROM tests CROSS JOIN LATERAL UNNEST(scores) AS unnest(unnest)
```

hellozepp:
@VaggelisD PTAL

I found that multiple `explode` will cause the performance to drop exponentially, as shown in the following `starrocks` sql:
```
```
Transpile to `spark`, and get the output:
```
```
We can use `arrays_zip` to merge multiple `Lateral view explode`. In fact, our users often use the syntax combination `LATERAL VIEW EXPLODE(arrays_zip())`.

But arrays_zip may not be supported by some engines, such as hive. I added a parameter `unnest_using_arrays_zip` that only works on spark to avoid the impact.

I added a commit. Will be open to any suggestions

sqlglot/expressions.py
```diff
@@ -5331,6 +5331,12 @@ class Explode(Func):
     is_var_len_args = True
 
 
+# https://spark.apache.org/docs/latest/api/sql/#inline
+class Inline(Func):
+    arg_types = {"this": True, "expressions": False}
+    is_var_len_args = True
+
+
```
VaggelisD, line 5339: According to the docs, `INLINE` accepts only a single expression i.e an array of structs; The `expressions` arg and `is_var_len_args = True` here are not needed in that case

VaggelisD (commented):
Hey, thanks for the follow-up PR!  Left a few comments in my first pass.

sqlglot/transforms.py
```diff
@@ -296,12 +297,16 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
+def unnest_to_explode(
+    expression: exp.Expression,
+    unnest_using_arrays_zip: bool = True,
+    support_unnest_func: bool = False,
+) -> exp.Expression:
```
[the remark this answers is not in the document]
hellozepp: Yes, support_unnest_func is only for generalization consideration. In my company's engine `clickzetta`, it supports unnest function. I think it is necessary to leave a parameter in the unnest function of non-lateral view.

tests/dialects/test_starrocks.py
```diff
@@ -35,6 +35,35 @@ def test_unnest(self):
...
+        self.validate_identity(
+            "SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores)",
+            "SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores) AS unnest(unnest)",
+        )
+        self.validate_all(
+            r"""SELECT * FROM UNNEST(array['John','Jane','Jim','Jamie'], array[24,25,26,27]) AS t(name, age)""",
+            write={
+                "postgres": "SELECT * FROM UNNEST(ARRAY['John', 'Jane', 'Jim', 'Jamie'], ARRAY[24, 25, 26, 27]) AS t(name, age)",
+                "spark": "SELECT * FROM EXPLODE(ARRAY('John', 'Jane', 'Jim', 'Jamie'), ARRAY(24, 25, 26, 27)) AS t(name, age)",
+            },
```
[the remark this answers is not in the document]
hellozepp: Oh! I didn’t consider this. I’m submitting a commit to adapt the inline processing.

sqlglot/expressions.py
hellozepp: Tks! I will fix it

sqlglot/transforms.py
```diff
@@ -325,17 +330,30 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
+                        [exp.Anonymous(this="ARRAYS_ZIP", expressions=unnest.expressions)],
+                    )
+                    _udtf = exp.Inline
+
                 for e, column in zip(unnest.expressions, alias.columns if alias else []):
                     expression.append(
                         "laterals",
                         exp.Lateral(
-                            this=udtf(this=e),
+                            this=_udtf(this=e),
```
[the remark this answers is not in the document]
hellozepp: udtf is defined at line:312:
```
udtf = exp.Posexplode if unnest.args.get("offset") else exp.Explode
```
I added the underscore to make sure I created a variable of type[Func].

sqlglot/transforms.py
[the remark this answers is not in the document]
hellozepp: In addition, we look forward to improving the code of clickzetta's dialect. If there is a chance, it will be merged into sqlglot later, which will be very exciting =)

sqlglot/dialects/starrocks.py
```diff
@@ -35,6 +35,12 @@ def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
             if unnest:
                 alias = unnest.args.get("alias")
 
+                if not alias:
+                    # Starrocks defaults to naming the table alias as "unnest"
+                    alias = exp.TableAlias(
+                        this=exp.to_identifier("unnest"), columns=[exp.to_identifier("unnest")]
+                    )
+                    unnest.set("alias", alias)
```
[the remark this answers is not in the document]
hellozepp: Got it.

sqlglot/transforms.py
```diff
@@ -325,17 +330,30 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
                 expression.args["joins"].remove(join)
 
+                if unnest_using_arrays_zip and has_multi_expr:
+                    # Modify the logic to use arrays_zip if there are multiple expressions
+                    # Build arrays_zip with nested expressions correctly
+                    unnest.set(
+                        "expressions",
+                        [exp.Anonymous(this="ARRAYS_ZIP", expressions=unnest.expressions)],
+                    )
+                    _udtf = exp.Inline
```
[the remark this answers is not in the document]
hellozepp: That's right.

So when has_multi_expr is True, we use `inline` instead of `explode`, `lateral view explode` will get a multi-row structure, and inline can expand the structure and set the column alias, so that it can be consistent with the select expr behavior.

hellozepp:
@VaggelisD Added new commit, PTAL

sqlglot/transforms.py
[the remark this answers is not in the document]
VaggelisD: Let's remove this parameter for now, it's currently dead code which doesn't serve an immediate purpose. If another dialect (e.g. Clickzetta) needs it at some point, we can reconsider it.

sqlglot/transforms.py
```diff
@@ -296,19 +296,41 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
...
+) -> exp.Expression:
     """Convert cross join unnest into lateral view explode."""
+
+    def handle_unnest(u: exp.Unnest, _has_multi_expr: bool) -> None:
+        if unnest_using_arrays_zip and _has_multi_expr:
+            # Use INLINE(ARRAYS_ZIP(...)) for multiple expressions
+            u.set(
+                "expressions",
+                [exp.Anonymous(this="ARRAYS_ZIP", expressions=u.expressions)],
+            )
```
VaggelisD, line 312: The function name could be more descriptive e.g. `def _unnest_zip_exprs(...)`

sqlglot/transforms.py
```diff
@@ -296,19 +296,41 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
...
 
-        if from_ and isinstance(from_.this, exp.Unnest):
+        if not support_unnest_func and from_ and isinstance(from_.this, exp.Unnest):
             unnest = from_.this
             alias = unnest.args.get("alias")
-            udtf = exp.Posexplode if unnest.args.get("offset") else exp.Explode
+            has_multi_expr = len(unnest.expressions) > 1
+            handle_unnest(unnest, has_multi_expr)
             this, *expressions = unnest.expressions
+
```
VaggelisD, line 330: Note that the `Expression` properties such as `expr.this`, `expr.expressions` etc are dict lookups under the hood so it's best if we can cache the result; Here `unnest.expressions` is called on both line 327 and 329, so we can either cache `unnest.expressions` or use `*expressions` as a proxy for `has_multi_expr`

sqlglot/transforms.py
```diff
@@ -324,18 +346,27 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
             unnest = join_expr.this if is_lateral else join_expr
 
             if isinstance(unnest, exp.Unnest):
-                alias = join_expr.args.get("alias") if is_lateral else unnest.args.get("alias")
-                udtf = exp.Posexplode if unnest.args.get("offset") else exp.Explode
+                alias = (
+                    join_expr.args.get("alias")
+                    if is_lateral and join_expr.args.get("alias")
+                    else unnest.args.get("alias")
```
VaggelisD, line 352: Why was the new condition (RHS of the `and`) needed here?

Also, ditto about the `arg` lookups

sqlglot/dialects/starrocks.py
```diff
@@ -35,7 +35,10 @@ def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
             if unnest:
                 alias = unnest.args.get("alias")
 
-                if alias and not alias.args.get("columns"):
+                if not alias:
+                    alias = exp.TableAlias(columns=[exp.to_identifier("unnest")])
+                    unnest.set("alias", alias)
+                if not alias.args.get("columns"):
```
VaggelisD, line 41: These branches are independent now i.e if we hit the first then we'll not hit the second. Lets connect them:

 ```
if not alias:
  ...
elif not alias.args.get("columns"):
  ...
```

File: sqlglot/transforms.py (hunk 8 above)
Comment (hellozepp) [reply sqlglot/transforms.py:330]: We set the default alias in `starrocks.py`:
```
unnest.set("alias", alias)
```
The following `starrocks sql` will use this logic:
```
SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores)
```
But I found that ` join_expr.args.get("alias")` may not get alias, it is only in Unnest, `sqlglot.dialects.starrocks.StarRocks.Parser._parse_unnest` can not set the lateral alias, am I missing something?
I provide details:

```
Select(
  expressions=[
    Column(
      this=Identifier(this=student, quoted=False)),
    Column(
      this=Identifier(this=score, quoted=False)),
    Column(
      this=Identifier(this=unnest, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=tests, quoted=False))),
  joins=[
    Join(
      this=Lateral(
        this=Unnest(
          expressions=[
            Column(
              this=Identifier(this=scores, quoted=False))],
          alias=TableAlias(
            columns=[
              Identifier(this=unnest, quoted=False)]))),
      kind=CROSS)])
```

File: sqlglot/transforms.py (hunk 9 above)
Comment (hellozepp) [reply sqlglot/transforms.py:352]: We set the default alias in `starrocks.py`:
```
unnest.set("alias", alias)
```
The following `starrocks sql` will use this logic:
```
SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores)
```
But I found that ` join_expr.args.get("alias")` may not get alias, it is only in Unnest, `sqlglot.dialects.starrocks.StarRocks.Parser._parse_unnest` cannot set the lateral alias.

I think when is_lateral, it cannot be obtained from outside, and the default value we set should be obtained from unnest.args. Am I missing something?

I provide details:

```
Select(
  expressions=[
    Column(
      this=Identifier(this=student, quoted=False)),
    Column(
      this=Identifier(this=score, quoted=False)),
    Column(
      this=Identifier(this=unnest, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=tests, quoted=False))),
  joins=[
    Join(
      this=Lateral(
        this=Unnest(
          expressions=[
            Column(
              this=Identifier(this=scores, quoted=False))],
          alias=TableAlias(
            columns=[
              Identifier(this=unnest, quoted=False)]))),
      kind=CROSS)])
```

File: sqlglot/dialects/starrocks.py (hunk 10 above)
Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:41]: Got it. I will fix it.

File: sqlglot/transforms.py (hunk 7 above)
Comment (hellozepp) [reply sqlglot/transforms.py:312]: Good idea, _unnest_zip_exprs is more readable

**Comment by hellozepp** [comment]
@VaggelisD Good evening, I made some change, please take a look at it when you have time

File: sqlglot/dialects/hive.py [hunk 11]
```diff
@@ -485,7 +486,7 @@ class Generator(generator.Generator):
                 [
                     transforms.eliminate_qualify,
                     transforms.eliminate_distinct_on,
-                    transforms.unnest_to_explode,
+                    partial(transforms.unnest_to_explode, unnest_using_arrays_zip=False),
```
VaggelisD, line 489: Did you confirm that HiveQL does not have `ARRAYS_ZIP`? Usually, if something works for Spark 2 (which has both `INLINE()` and `ARRAYS_ZIP()`) we consider it a part of the Hive dialect too.

sqlglot/transforms.py
```diff
@@ -296,19 +297,42 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
...
+        u: exp.Unnest, unnest_exprs: t.List[t.Any], _has_multi_expr: bool
+    ) -> t.List[Func]:
+        if unnest_using_arrays_zip and _has_multi_expr:
+            # Use INLINE(ARRAYS_ZIP(...)) for multiple expressions
+            zip_exprs: t.List[Func] = [exp.Anonymous(this="ARRAYS_ZIP", expressions=unnest_exprs)]
+            u.set("expressions", zip_exprs)
+            return zip_exprs
+        return unnest_exprs
+
+    def _udtf_type(u: exp.Unnest, _has_multi_expr: bool):
```
georgesittas, line 316:
```suggestion
    def _udtf_type(u: exp.Unnest, _has_multi_expr: bool) -> exp.Func:
```

sqlglot/transforms.py
```diff
@@ -296,19 +297,42 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
...
 
-def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
+def unnest_to_explode(
+    expression: exp.Expression,
+    unnest_using_arrays_zip: bool = True,
+) -> exp.Expression:
     """Convert cross join unnest into lateral view explode."""
+
+    def _unnest_zip_exprs(
+        u: exp.Unnest, unnest_exprs: t.List[t.Any], _has_multi_expr: bool
```
georgesittas, line 307:
```suggestion
        u: exp.Unnest, unnest_exprs: t.List[exp.Expression], _has_multi_expr: bool
```

georgesittas: Let's get rid of the leading underscore from `_has_multi_expr`.

sqlglot/transforms.py
```diff
@@ -324,18 +348,28 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
...
 
             if isinstance(unnest, exp.Unnest):
-                alias = join_expr.args.get("alias") if is_lateral else unnest.args.get("alias")
-                udtf = exp.Posexplode if unnest.args.get("offset") else exp.Explode
+                if is_lateral:
+                    alias = join_expr.args.get("alias") or unnest.args.get("alias")
+                else:
+                    alias = unnest.args.get("alias")
+                exprs = unnest.expressions
+                # The number of unnest.expressions will be changed by handle_unnest, we need to record it here
```
georgesittas, line 356:
```suggestion
                # The number of unnest.expressions will be changed by _unnest_zip_exprs, we need to record it here
```

tests/dialects/test_dialect.py
```diff
@@ -1399,7 +1399,8 @@ def test_cross_join(self):
             write={
                 "drill": "SELECT a, b FROM x CROSS JOIN UNNEST(y, z) AS t(a, b)",
                 "presto": "SELECT a, b FROM x CROSS JOIN UNNEST(y, z) AS t(a, b)",
-                "spark": "SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS a LATERAL VIEW EXPLODE(z) t AS b",
+                "spark": "SELECT a, b FROM x LATERAL VIEW INLINE(ARRAYS_ZIP(y, z)) t AS a, b",
+                "hive": "SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS a LATERAL VIEW EXPLODE(z) t AS b",
```
georgesittas, line 1403: It looks like the transpilation for unnest with >1 expressions was wrong all along? cc @tobymao

```
trino> WITH x AS (SELECT ARRAY[1,2,3] AS y, ARRAY[4,5,6] AS z) SELECT a, b FROM x CROSS JOIN UNNEST(y, z) AS t (a, b);
 a | b 
---+---
 1 | 4 
 2 | 5 
 3 | 6 
(3 rows)
```

Current SQLGlot main produces:

```python
(.venv) ➜  sqlglot git:(main) python
>>> import sqlglot
>>> sqlglot.transpile("WITH x AS (SELECT ARRAY[1,2,3] AS y, ARRAY[4,5,6] AS z) SELECT a, b FROM x CROSS JOIN UNNEST(y, z) AS t (a, b)", "trino", "spark")
['WITH x AS (SELECT ARRAY(1, 2, 3) AS y, ARRAY(4, 5, 6) AS z) SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS a LATERAL VIEW EXPLODE(z) t AS b']
```

And Spark yields different results:

```
spark-sql (default)> WITH x AS (SELECT ARRAY(1, 2, 3) AS y, ARRAY(4, 5, 6) AS z) SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS a LATERAL VIEW EXPLODE(z) t AS b;
a	b
1	4
1	5
1	6
2	4
2	5
2	6
3	4
3	5
3	6
Time taken: 4.504 seconds, Fetched 9 row(s)
spark-sql (default)> WITH x AS (SELECT ARRAY(1, 2, 3) AS y, ARRAY(4, 5, 6) AS z) SELECT a, b FROM x LATERAL VIEW INLINE(ARRAYS_ZIP(y, z)) t AS a, b;
a	b
1	4
2	5
3	6
Time taken: 0.633 seconds, Fetched 3 row(s)
```

----

The hive case is still incorrect right? Having two `LATERAL VIEW EXPLODE` clauses will result in a cross join instead of an array zip & join. We don't have to fix it now if it's not trivial, but I suggest we get rid of this (misleading) test, and then do `self.unsupported` for Hive if there are multiple expressions to avoid data corruption issues.

tests/dialects/test_starrocks.py
```diff
@@ -35,6 +35,37 @@ def test_unnest(self):
             "SELECT student, score, t.unnest FROM tests CROSS JOIN LATERAL UNNEST(scores) AS t",
             "SELECT student, score, t.unnest FROM tests CROSS JOIN LATERAL UNNEST(scores) AS t(unnest)",
         )
+        self.validate_all(
+            "SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores)",
+            write={
+                "spark": "SELECT student, score, unnest FROM tests LATERAL VIEW EXPLODE(scores) AS unnest",
+            },
+        )
```
georgesittas, line 43: We need to always test roundtrip when making changes, so please add `"starrocks": ...` under `write` for every test here.

sqlglot/transforms.py
```diff
@@ -3,6 +3,7 @@
 import typing as t
 
 from sqlglot import expressions as exp
+from sqlglot.expressions import Func
```
georgesittas, line 6: This isn't necessary, just do `exp.Func` anywhere `Func` is needed.

sqlglot/dialects/starrocks.py
```diff
@@ -35,7 +35,10 @@ def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
             if unnest:
                 alias = unnest.args.get("alias")
 
-                if alias and not alias.args.get("columns"):
+                if not alias:
+                    alias = exp.TableAlias(columns=[exp.to_identifier("unnest")])
+                    unnest.set("alias", alias)
```
georgesittas, line 40: Is this tested? What does generation of `UNNEST` without an alias look like in Starrocks after this change?

sqlglot/dialects/hive.py
georgesittas, line 489: See comment about using `self.unsupported` if we can't correctly transpile to Hive the case where `len(unnest.expressions) > 1`.

sqlglot/dialects/spark2.py
```diff
@@ -190,6 +190,13 @@ class Generator(Hive.Generator):
 
         TRANSFORMS = {
             **Hive.Generator.TRANSFORMS,
+            exp.Select: transforms.preprocess(
```
georgesittas, line 193: Nit: let's preserve the alphabetical ordering in this dict please.

sqlglot/transforms.py
```diff
@@ -296,19 +297,42 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
...
+            u.set("expressions", zip_exprs)
+            return zip_exprs
+        return unnest_exprs
+
+    def _udtf_type(u: exp.Unnest, _has_multi_expr: bool):
+        return (
+            exp.Posexplode
+            if u.args.get("offset")
+            else (exp.Inline if unnest_using_arrays_zip and _has_multi_expr else exp.Explode)
+        )
```
georgesittas, line 321: Nit:
```suggestion
        if u.args.get("offset"):
            return exp.Posexplode
        return exp.Inline if unnest_using_arrays_zip and _has_multi_expr else exp.Explode
```

georgesittas (commented):
Thanks for the PR @hellozepp!

sqlglot/transforms.py
georgesittas: My bad, this needed to be `t.Type[exp.Func]`, apologies.

tests/dialects/test_dialect.py
hellozepp: Yes, as you can see. I delete this line of test case

sqlglot/dialects/starrocks.py
VaggelisD: @hellozepp I think my recommendation of removing the table alias might have broken the roundtrip here, sorry for that. We can test the roundtrip in the first new `test_starrocks.py` test case you added, i.e:

```
        self.validate_all(
            "SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores)",
            write={
                "spark": "SELECT student, score, unnest FROM tests LATERAL VIEW EXPLODE(scores) AS unnest",
                "starrocks": "SELECT student, score, unnest FROM tests CROSS JOIN LATERAL UNNEST(scores) AS unnest(unnest)",
            },
        )
```

As you suggested, Starrocks seems to name both the `UNNEST` table and it's output column as `unnest`. Here's the `DESCRIBE` result for a similar query:

```
StarRocks > describe with t as (select 1 as col, [1, 2, 3] as col2) select unnest from t cross join unnest(col2);
+--------------------------------+
| Explain String                 |
+--------------------------------+
| PLAN FRAGMENT 0                |
|  OUTPUT EXPRS:4: unnest        |
|   PARTITION: UNPARTITIONED     |
|                                |
|   RESULT SINK                  |
|                                |
|   2:TableValueFunction         |
|   |  tableFunctionName: unnest |
|   |  columns: [unnest]         |
|   |  returnTypes: [TINYINT]    |
|   |                            |
|   1:Project                    |
|   |  <slot 3> : [1,2,3]        |
|   |                            |
|   0:UNION                      |
|      constant exprs:           |
|          NULL                  |
+--------------------------------+
17 rows in set (0.02 sec)
```

sqlglot/dialects/hive.py
hellozepp: `self.unsupported` is not accessible in `unnest_to_explode`. Is it possible to throw an exception in unnest_to_explode according to `unnest_using_arrays_zip`?

georgesittas: We should aim to use `unsupported` instead of preemptively raising, because with the former you can control whether an exception will be thrown by setting the error level accordingly.

I'm not sure what's the best way to do this, but perhaps you could try overriding `select_sql` in Hive and invoking `unnest_to_explode` in it, unless you found there's at least an `Unnest` expression in scope with multiple `expressions`, in which case you'd call `self.unsupported(...)`. The `find_all_in_scope` helper under `optimizer/scope.py` may be useful here to avoid traversing subqueries and only look for unnests in the current `SELECT`'s scope. Note that this approach will affect subclasses of `Hive` (e.g. `Spark`) because they'll now use this `select_sql` override, so we need to avoid that somehow if that's the approach.

----

Before trying all of the above, as Vaggelis mentioned we should make sure Hive doesn't support `INLINE` or `ARRAYS_ZIP`.

sqlglot/dialects/starrocks.py
hellozepp: @VaggelisD Oh, I understand. Here we just need to avoid setting the default unnest alias when table alias exists. I revert the code back.

sqlglot/dialects/hive.py
hellozepp: I am in hive3.1, there is no ARRAYS_ZIP

sqlglot/dialects/starrocks.py
hellozepp: @VaggelisD 
image

hellozepp:
@georgesittas @VaggelisD Added new commit, PTAL

georgesittas (approved):
Thanks for the PR @hellozepp - LGTM. Appreciate the patience & quick iterations.

We plan to get this in soon and take a stab at refactoring the hive transformation so that it actually does the right thing. One way to go about this is by simulating the `INLINE` & `ARRAYS_ZIP` with a subquery that joins the `POSEXPLODE`d arrays on their positions and then projects the corresponding columns so they can be accessed in the scope of the original `UNNEST`.

hellozepp:
@VaggelisD Agree.
Thanks for the previous advice, it helped me a lot =)

hellozepp:
Sorry, I cleaned up my sqlglot repository because it was a mess. Wait a minute and I'll make another commit. There won't be any new code since our last commit.

PR #3969, opened 2024-08-26
Fix!: stop normalizing qualified anonymous functions

### LINKED ISSUE #3967: ms sql server method names are incorrectly considered case-insensitive
```
In [4]: sql = """
   ...: SELECT
   ...: TOP 10
   ...:   s.RECORDID, /* , rd.FormContent */
   ...:   n.c.value('(/*:FORM_ROOT/*:SOME_TAG)[1]', 'float') AS SOME_TAG_VALUE
   ...: FROM source_table.dbo.source_data AS s(nolock)
   ...:     CROSS APPLY FormContent.nodes('/*:FORM_ROOT') AS N(C)
   ...: """

In [5]: import sqlglot.expressions as sge, sqlglot as sg

In [6]: print(sg.parse_one(sql, read="tsql").sql('tsql', pretty=True))
SELECT
TOP 10
  s.RECORDID, /* , rd.FormContent */
  n.c.VALUE('(/*:FORM_ROOT/*:SOME_TAG)[1]', 'float') AS SOME_TAG_VALUE
FROM source_table.dbo.source_data AS s(nolock) CROSS APPLY FormContent.NODES('/*:FORM_ROOT') AS N(C)
```

`VALUE` and `NODES` are methods on their respective objects and it is incorrect to alter their case like this.

docs: https://learn.microsoft.com/en-us/sql/t-sql/xml/xml-data-type-methods?view=sql-server-ver16

georgesittas:
Fixes #3967

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2888,7 +2888,12 @@ class Generator(metaclass=_Generator):
         return f"REFERENCES {this}{expressions}{options}"
 
     def anonymous_sql(self, expression: exp.Anonymous) -> str:
-        return self.func(self.sql(expression, "this"), *expression.expressions)
+        # We don't normalize qualified functions such as a.b.foo(), because they can be case-sensitive
+        parent = expression.parent
+        is_qualified = isinstance(parent, exp.Dot) and expression is parent.expression
+        return self.func(
+            self.sql(expression, "this"), *expression.expressions, normalize=not is_qualified
+        )
 
     def paren_sql(self, expression: exp.Paren) -> str:
         sql = self.seg(self.indent(self.sql(expression, "this")), sep="")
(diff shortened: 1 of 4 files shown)


sqlglot/generator.py
```diff
@@ -2888,7 +2888,12 @@ def reference_sql(self, expression: exp.Reference) -> str:
         return f"REFERENCES {this}{expressions}{options}"
 
     def anonymous_sql(self, expression: exp.Anonymous) -> str:
-        return self.func(self.sql(expression, "this"), *expression.expressions)
+        # We don't normalize qualified functions such as a.b.foo(), because they can be case-sensitive
+        parent = expression.parent
+        is_qualified = isinstance(parent, exp.Dot) and expression is parent.expression
```
VaggelisD, line 2893: Could this be simplified by checking `arg_key`? 

Also, is this a problem for other dialects too e.g. Oracle's `getxmltype`?

VaggelisD (approved):

sqlglot/generator.py
georgesittas: > Could this be simplified by checking arg_key?

I don't think so, it's essentially the same check and involves a string comparison vs object equivalence.

> Also, is this a problem for other dialects too e.g. Oracle's getxmltype?

Dunno but the most conservative approach is to assume it is, since you'll just get identity for roundtrip.

PR #3971, opened 2024-08-26
feat(postgres): Support for IS JSON predicate

### LINKED ISSUE #3965: Issue with Parsing IS JSON Function in PostgreSQL Dialect
I'm encountering an issue with the PostgreSQL dialect in SQLGlot where it fails to parse a query containing the IS JSON function. The expected behavior is to convert the PostgreSQL query to an AST, but instead, it throws a ParseError. 

**Below is the code snippet that reproduces the issue:**

```
from sqlglot.dialects.postgres import Postgres
from sqlglot import parse_one

sql = """
SELECT js,
  js IS JSON "json?",
  js IS JSON SCALAR "scalar?",
  js IS JSON OBJECT "object?",
  js IS JSON ARRAY "array?"
FROM (VALUES
      ('123'), ('"abc"'), ('{"a": "b"}'), ('[1,2]'),('abc')) foo(js);
"""

ast = parse_one(sql, dialect=Postgres)
print(repr(ast))
```

**ERROR :** 

`sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 3, Col: 11.`

It seems that the parser does not handle the IS JSON syntax correctly, particularly with the modifiers like SCALAR, OBJECT, and ARRAY.

**Expected Behavior:**

The query should be parsed into an AST without errors.

**Official Documentation**
[https://www.postgresql.org/docs/current/functions-json.html](https://www.postgresql.org/docs/current/functions-json.html)

VaggelisD:
Fixes #3965

Docs
--------
[Postgres JSON testing functions](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-SQLJSON-MISC)

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -4150,7 +4150,6 @@ class Generator(metaclass=_Generator):
         else:
             with_sql = " WITH"
 
-        unique = expression.args.get("unique")
-        unique_sql = " UNIQUE KEYS" if unique else ""
+        unique_sql = " UNIQUE KEYS" if expression.args.get("unique") else ""
 
         return f"JSON{this}{with_sql}{unique_sql}"
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -4136,3 +4136,21 @@ class Generator(metaclass=_Generator):
         expr = exp.AtTimeZone(this=timestamp, zone=target_tz)
 
         return self.sql(expr)
+
+    def json_sql(self, expression: exp.JSON) -> str:
+        this = self.sql(expression, "this")
+        this = f" {this}" if this else ""
+
+        _with = expression.args.get("with")
+
+        if _with is None:
+            with_sql = ""
+        elif not _with:
+            with_sql = " WITHOUT"
+        else:
+            with_sql = " WITH"
+
+        unique = expression.args.get("unique")
+        unique_sql = " UNIQUE KEYS" if unique else ""
+
+        return f"JSON{this}{with_sql}{unique_sql}"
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -4288,10 +4288,29 @@ def _parse_is(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expressi
             klass = exp.NullSafeEQ if negate else exp.NullSafeNEQ
             return self.expression(klass, this=this, expression=self._parse_bitwise())
 
-        expression = self._parse_null() or self._parse_boolean()
-        if not expression:
-            self._retreat(index)
-            return None
+        if self._match_text_seq("JSON"):
```
georgesittas, line 4291:
```suggestion
        if self._match(TokenType.JSON):
```

sqlglot/parser.py
```diff
@@ -4288,10 +4288,29 @@ def _parse_is(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expressi
...
+            kind = (
+                self._match_texts(("OBJECT", "SCALAR", "ARRAY", "VALUE"))
```
georgesittas, line 4293: Extract this out into a constant set

sqlglot/generator.py
```diff
@@ -4136,3 +4136,21 @@ def converttimezone_sql(self, expression: exp.ConvertTimezone) -> str:
...
+
+        if _with is None:
+            with_sql = ""
+        elif not _with:
+            with_sql = " WITHOUT"
+        else:
+            with_sql = " WITH"
+
+        unique = expression.args.get("unique")
+        unique_sql = " UNIQUE KEYS" if unique else ""
```
georgesittas, line 4154:
```suggestion
        unique_sql = " UNIQUE KEYS" if expression.args.get("unique") else ""
```

georgesittas (approved):

georgesittas (approved):

PR #3980, opened 2024-08-27
fix(duckdb)!: Generation of exp.SHA2, exp.Transform, exp.IgnoreNulls

### LINKED ISSUE #3972: Incorrect Parsing Between Spark SQL and DuckDB
There are a couple of mistakes when translating from Spark to DuckDB

```python
from sqlglot import parse_one
 query = """SELECT SHA2(str, 256), ANY_VALUE(str2, true), FIRST(str3, true), LAST(str4, true), TRANSFORM(array_col)
FROM table 
GROUP BY group_col"""
 parse_one(query, dialect="spark").sql(dialect="duckdb")

``` 
The output is:
```python
SELECT SHA2(str, 256), ANY_VALUE(str2 IGNORE NULLS), FIRST(str3 IGNORE NULLS), LAST(str4 IGNORE NULLS), TRANSFORM(array_col) FROM "table" GROUP BY group_col
``` 
Which fails for three different reasons:
1. SHA2 is not a function in duckdb - the relevant function is sha256 
2. TRANSFORM should be LIST_TRANSFORM
3. When using non-window function, duckdb doesn't support RESPECT/IGNORE NULLS. The default behavior is without NULLs 
The expected output
**Official Documentation**
1. https://duckdb.org/docs/sql/functions/utility.html - SHA256
2. https://duckdb.org/docs/sql/functions/list.html - LIST_TRANSFORM
3. https://duckdb.org/docs/sql/functions/aggregates.html - aggregate functions

VaggelisD:
Fixes #3972

This PR addresses all 3 issues (SHA256, LIST_TRANSFORM, IGNORE NULLS) as the first 2 are trivial to fix.

Docs
----------
[DuckDB Window Functions](https://duckdb.org/docs/sql/functions/window_functions.html) | [DuckDB LIST_TRANSFORM](https://duckdb.org/docs/sql/functions/list.html#list_transformlist-lambda) | [DuckDB SHA256](https://duckdb.org/docs/sql/functions/utility.html#sha256value)

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -42,7 +42,7 @@ DATETIME_DELTA = t.Union[
     exp.DateAdd, exp.TimeAdd, exp.DatetimeAdd, exp.TsOrDsAdd, exp.DateSub, exp.DatetimeSub
 ]
 
-WINDOW_FUNCS = (
+WINDOW_FUNCS_WITH_IGNORE_NULLS = (
     exp.FirstValue,
     exp.LastValue,
     exp.Lag,
@@ -920,8 +920,9 @@ class DuckDB(Dialect):
             return super().unnest_sql(expression)
 
         def ignorenulls_sql(self, expression: exp.IgnoreNulls) -> str:
-            if isinstance(expression.this, WINDOW_FUNCS):
-                # DuckDB should render IGNORE NULLS only in window functions e.g. FIRST_VALUE(... IGNORE NULLS) OVER (...)
+            if isinstance(expression.this, WINDOW_FUNCS_WITH_IGNORE_NULLS):
+                # DuckDB should render IGNORE NULLS only for the general-purpose
+                # window functions that accept it e.g. FIRST_VALUE(... IGNORE NULLS) OVER (...)
                 return super().ignorenulls_sql(expression)
 
             return self.sql(expression, "this")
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -41,6 +42,14 @@
     exp.DateAdd, exp.TimeAdd, exp.DatetimeAdd, exp.TsOrDsAdd, exp.DateSub, exp.DatetimeSub
 ]
 
+WINDOW_FUNCS = (
```
tobymao, line 45: can we make this a subclass instead? change the expression subclass to window func

georgesittas: I think these are a subset of the actual window funcs, right @VaggelisD? IIRC duckdb only allows `IGNORE NULLS` for certain window funcs, hence this tuple? If so, I'd actually suggest renaming this to something like `WINDOW_FUNCS_WITH_IGNORE_NULLS` or something.

VaggelisD: Right, this was a detail that I should have documented more clearly; DuckDB differentiates the aggregate functions from the general-purpose window functions, and `IGNORE NULLS` is accepted on a subset of the general-purpose window funcs. 

This [table here](https://duckdb.org/docs/sql/functions/window_functions.html#general-purpose-window-functions) documents them, note those with `IGNORE NULLS`. The `WINDOW_FUNCS` set holds this subset, but it's not accurately named as George pointed out, will rename it.

tobymao: got it

tobymao (approved):

PR #3994, opened 2024-08-28
fix!: Bind TABLESAMPLE to exp.Subquery

### LINKED ISSUE #3992: tablesample with alias is incorrect when going from duckdb to pyspark
```
In [17]: import sqlglot.expressions as sge, sqlglot as sg

In [18]: sql = "select t1.* from (select * from t) t1 tablesample (30 rows)"

In [19]: sg.parse_one(sql, read="duckdb").sql("spark")
Out[19]: 'SELECT t1.* FROM (SELECT * FROM t) AS t1 TABLESAMPLE (30 ROWS)'
```

The `t1` alias must be placed after the `TABLESAMPLE` or else PySpark throws a parse error. This appears to be a PySpark-only thing.

VaggelisD:
Fixes #3992

When parsing subqueries the tralining `TABLESAMPLE` is not attached to it but to the top-level `exp.Select` statement, leading to erratic behavior such as `"sample"` arg being overwritten:

```
>>> sql = "SELECT * FROM (SELECT * FROM t) AS t1 TABLESAMPLE (1 ROWS), (SELECT * FROM t) AS t2 tablesample (2 rows)"
>>> sqlglot.parse_one(sql, read='duckdb').sql('duckdb')

SELECT * FROM (SELECT * FROM t) AS t1, (SELECT * FROM t) AS t2 USING SAMPLE RESERVOIR (2 ROWS)
```

This PR addresses this by parsing TABLESAMPLE inside `_parse_subquery()`, which binds it to the respective `exp.Subquery`.

Also, the order of subquery alias vs TABLESAMPLE generation is fixed (as is done in `table_sql()`) in order to address the issue.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2447,6 +2447,13 @@ class Generator(metaclass=_Generator):
     def subquery_sql(self, expression: exp.Subquery, sep: str = " AS ") -> str:
         alias = self.sql(expression, "alias")
         alias = f"{sep}{alias}" if alias else ""
+        sample = self.sql(expression, "sample")
+        if self.dialect.ALIAS_POST_TABLESAMPLE and sample:
+            alias = f"{sample}{alias}"
+
+            # Set to None so it's not generated again by self.query_modifiers()
+            expression.set("sample", None)
+
         pivots = self.expressions(expression, key="pivots", sep="", flat=True)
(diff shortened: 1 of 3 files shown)


georgesittas (approved):

cpcloud:
Wonderful as always, thank you!

Issue #3995, opened 2024-08-28
Support for `COUNT_BIG` in T-SQL


colin-ho:
I'm currently using SQLGlot to transpile into T-SQL queries, some of them are `COUNT` queries. However, I'm having integer overflow issues because my table count exceeds int bounds. This is because `COUNT` in T-SQL returns an int, instead I would like to be using `COUNT_BIG` which returns a big-int. See: https://learn.microsoft.com/en-us/sql/t-sql/functions/count-big-transact-sql?view=sql-server-ver16

In other dialects like postgres, `COUNT` returns a big-int. https://www.postgresql.org/docs/8.2/functions-aggregate.html

Is it possible to modify the T-SQL dialect to use `COUNT_BIG`, or add an option to use this? 

Example current behaviour:
```
query = "SELECT COUNT(*) FROM big_table"
t_sql_query = sqlglot.transpile(query, "tsql")
print(t_sql_query)
['SELECT COUNT(*) FROM big_table']
```

tobymao:
low priority for us, feel free to add a PR, we're happy to work with you on getting it in.

colin-ho:
Gotcha, just made a PR

PR #3996, opened 2024-08-28
fix(tsql): Use count_big instead of count

colin-ho:
Fixes: https://github.com/tobymao/sqlglot/issues/3995

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -521,6 +521,12 @@ class TSQL(Dialect):
                 substr=seq_get(args, 0),
                 position=seq_get(args, 2),
             ),
+            "COUNT": lambda args: exp.Count(
+                this=seq_get(args, 0), expressions=args[1:], big_int=False
+            ),
+            "COUNT_BIG": lambda args: exp.Count(
+                this=seq_get(args, 0), expressions=args[1:], big_int=True
+            ),
             "DATEADD": build_date_delta(exp.DateAdd, unit_mapping=DATE_DELTA_INTERVAL),
             "DATEDIFF": _build_date_delta(exp.DateDiff, unit_mapping=DATE_DELTA_INTERVAL),
             "DATENAME": _build_formatted_time(exp.TimeToStr, full_format_mapping=True),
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -854,6 +854,7 @@ class Generator(generator.Generator):
             exp.AutoIncrementColumnConstraint: lambda *_: "IDENTITY",
             exp.DateAdd: date_delta_sql("DATEADD"),
             exp.DateDiff: date_delta_sql("DATEDIFF"),
+            exp.Count: rename_func("COUNT_BIG"),
```
georgesittas, line 857: Should we adjust the parsing a bit so that we can actually preserve `COUNT` calls and only output `COUNT_BIG` if 1) T-SQL parsed a `COUNT_BIG` or 2) we're transpiling `COUNT` from other dialects to T-SQL?

tobymao: yep, we should preserve round trip

georgesittas: So @colin-ho, the action items here are:

- Add new arg to `exp.Count`, something like `"bigint": False`
- Mimic what @VaggelisD did in this [commit](https://github.com/tobymao/sqlglot/commit/1284fd0a64890d3548af7ed0a0cc05bb6166ccb2#diff-63eb8dd82d3561bc414e3e13cf59516bd6584c5766a10072157a8a5aee4ad85fR225) for `Coalesce` (needed because of `is_var_len_args`)
- Include a couple of `FUNCTION` entries in T-SQL to parse `COUNT` and `COUNT_BIG` appropriately
- Test roundtrip & transpilation
- Bonus: fix `exp.Count` type inference by removing [this](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/dialect.py#L531) and adding an entry in `ANNOTATORS` that sets to `INT`/`BIGINT` accordingly. Make sure to test.

georgesittas (approved):
Nice work, appreciate the PR. Keep in mind to test roundtrip as well in `validate_all`.

PR #4004, opened 2024-08-29
fix(spark): Custom annotation for SUBSTRING()

### LINKED ISSUE #4002: Substring type inference missing binary case

```python
>>> optimize(sqlglot.parse_one("select col from tbl",read='spark'),schema={"tbl": {"col": "BINARY"}}).expressions[0].type
DataType(this=Type.BINARY, nested=False)
>>> optimize(sqlglot.parse_one("select substring(col, 2, 3) as x from tbl",read='spark'),schema={"tbl": {"col": "BINARY"}}).expressions[0].type
DataType(this=Type.VARCHAR)
```

In the second expression, the expected output type is `Type.BINARY`. In spark shell, this works fine (spark 3.5.1):

```scala
scala> val df = spark.sql("select cast('thing' as binary)")
df: org.apache.spark.sql.DataFrame = [CAST(thing AS BINARY): binary]

scala> val df = spark.sql("select substring(cast('thing' as binary), 1, 3)")
df: org.apache.spark.sql.DataFrame = [substring(CAST(thing AS BINARY), 1, 3): binary]

```

From my understanding, the return data type for `Substring` is hardcoded to VARCHAR: https://github.com/tobymao/sqlglot/blob/fcaae87d50653600048917945652cdf463bcc3cf/sqlglot/dialects/dialect.py#L609-L617

VaggelisD:
Fixes #4002

Could not test HiveQL but the docs mention only `STRING` as the return type, so this PR is limited to Spark2+:

- Spark 2:
```
scala> spark.sql("select substring(cast('thing' as binary), 1, 2), substring('thing', 1, 2)")
res4: org.apache.spark.sql.DataFrame = [substring(CAST(thing AS BINARY), 1, 2): binary, substring(thing, 1, 2): string]
```

- Spark 3:
```
spark-sql (default)> select typeof(substring('thing', 1, 2)), typeof(substring(cast('thing' as binary), 1, 2));
string  binary
```

- Databricks
```
select typeof(substring(cast('thing' as binary), 1, 2)) as bin, typeof(substring('thing', 1, 2)) as str; 

bin	str
binary	string
```

Docs
--------
[Databricks SUBSTRING](https://docs.databricks.com/en/sql/language-manual/functions/substr.html) | [HiveQL Builtin Functions](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF)

### DIFF
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -4,6 +4,7 @@ import typing as t
 
 from sqlglot import exp, transforms
 from sqlglot.dialects.dialect import (
+    Dialect,
     binary_from_function,
     build_formatted_time,
     is_parse_json,
@@ -111,6 +112,11 @@ def temporary_storage_provider(expression: exp.Expression) -> exp.Expression:
 
 
 class Spark2(Hive):
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        exp.Substring: lambda self, e: self._annotate_by_args(e, "this"),
+    }
+
     class Parser(Hive.Parser):
         TRIM_PATTERN_FIRST = True
 
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -4,7 +4,6 @@ import typing as t
 
 from sqlglot import exp, transforms
 from sqlglot.dialects.dialect import (
-    Dialect,
     binary_from_function,
     build_formatted_time,
     is_parse_json,
@@ -113,7 +112,7 @@ def temporary_storage_provider(expression: exp.Expression) -> exp.Expression:
 
 class Spark2(Hive):
     ANNOTATORS = {
-        **Dialect.ANNOTATORS,
+        **Hive.ANNOTATORS,
         exp.Substring: lambda self, e: self._annotate_by_args(e, "this"),
     }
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/spark2.py
```diff
@@ -111,6 +112,11 @@ def temporary_storage_provider(expression: exp.Expression) -> exp.Expression:
 
 
 class Spark2(Hive):
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
```
georgesittas, line 116: We can do `**Hive.ANNOTATORS` right?

tests/test_optimizer.py
```diff
@@ -1345,3 +1345,23 @@ def gen_expr(depth: int) -> exp.Expression:
...
+    def test_custom_annotators(self):
+        # In Spark hierarchy, SUBSTRING result type is dependent on input expr type
+        for dialect in ("spark2", "spark", "databricks"):
+            for expr_type_pair in (
+                ("col", "STRING"),
+                ("col", "BINARY"),
+                ("'str_literal'", "STRING"),
+                ("CAST('str_literal' AS BINARY)", "BINARY"),
+            ):
+                expr, type = expr_type_pair
```
georgesittas, line 1358: Consider using a `with self.subTest(...)` block here.

PR #4007, opened 2024-08-29
fix(clickhouse): ensure that ALL and DISTINCT are rendered for except and intersect

### LINKED ISSUE #4005: missing `DISTINCT` in `EXCEPT` statement when producing ClickHouse SQL
```
In [7]: import sqlglot.expressions as sge, sqlglot as sg

In [8]: sg.__version__
Out[8]: '25.17.0'

In [9]: sg.parse_one('select a from t except distinct select a from t', read='duckdb').sql('clickhouse')
Out[9]: 'SELECT a FROM t EXCEPT SELECT a FROM t'
```

ClickHouse docs: https://clickhouse.com/docs/en/sql-reference/statements/select/except#except-distinct

cpcloud:
Closes #4005.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1297,7 +1297,9 @@ class Generator(metaclass=_Generator):
         return self.set_operations(expression)
 
     def except_op(self, expression: exp.Except) -> str:
-        return f"EXCEPT{'' if expression.args.get('distinct') else ' ALL'}"
+        kind = " DISTINCT" if self.EXPLICIT_SET_OP else ""
+        kind = kind if expression.args.get("distinct") else " ALL"
+        return f"EXCEPT{kind}"
 
     def fetch_sql(self, expression: exp.Fetch) -> str:
         direction = expression.args.get("direction")
@@ -1679,7 +1681,9 @@ class Generator(metaclass=_Generator):
         return self.set_operations(expression)
 
     def intersect_op(self, expression: exp.Intersect) -> str:
-        return f"INTERSECT{'' if expression.args.get('distinct') else ' ALL'}"
+        kind = " DISTINCT" if self.EXPLICIT_SET_OP else ""
+        kind = kind if expression.args.get("distinct") else " ALL"
+        return f"INTERSECT{kind}"
 
     def introducer_sql(self, expression: exp.Introducer) -> str:
         return f"{self.sql(expression, 'this')} {self.sql(expression, 'expression')}"
--- a/tests/dialects/test_dialect.py
+++ b/tests/dialects/test_dialect.py
@@ -1488,12 +1488,14 @@ class TestDialect(Validator):
             "SELECT * FROM a INTERSECT SELECT * FROM b",
             read={
                 "bigquery": "SELECT * FROM a INTERSECT DISTINCT SELECT * FROM b",
+                "clickhouse": "SELECT * FROM a INTERSECT DISTINCT SELECT * FROM b",
                 "duckdb": "SELECT * FROM a INTERSECT SELECT * FROM b",
                 "presto": "SELECT * FROM a INTERSECT SELECT * FROM b",
                 "spark": "SELECT * FROM a INTERSECT SELECT * FROM b",
             },
             write={
                 "bigquery": "SELECT * FROM a INTERSECT DISTINCT SELECT * FROM b",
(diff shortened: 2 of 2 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -1106,3 +1106,9 @@ def replacepartition_sql(self, expression: exp.ReplacePartition) -> str:
 
         def projectiondef_sql(self, expression: exp.ProjectionDef) -> str:
             return f"PROJECTION {self.sql(expression.this)} {self.wrap(expression.expression)}"
+
+        def except_op(self, expression: exp.Except) -> str:
+            return f"EXCEPT{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
+
+        def intersect_op(self, expression: exp.Intersect) -> str:
+            return f"INTERSECT{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
```
georgesittas, line 1114: What about `union_op`? I see that we have a `EXPLICIT_SET_OP` flag, maybe we can just fix the base generator methods and set this flag to true for clickhouse?

sqlglot/dialects/clickhouse.py
```diff
@@ -1106,3 +1106,9 @@ def replacepartition_sql(self, expression: exp.ReplacePartition) -> str:
 
         def projectiondef_sql(self, expression: exp.ProjectionDef) -> str:
             return f"PROJECTION {self.sql(expression.this)} {self.wrap(expression.expression)}"
+
+        def except_op(self, expression: exp.Except) -> str:
+            return f"EXCEPT{' DISTINCT' if expression.args.get('distinct') else ' ALL'}"
```
cpcloud, line 1111: I went with rendering the explicit all for clarity in the generated SQL even though it isn't strictly necessary for clickhouse, rather than minimizing the SQL generated, since effectively every other dialect of SQL considers `EXCEPT` without `ALL` to mean `DISTINCT`. It's very confusing to read the output SQL without `ALL`, and have the execution mean `ALL`.

sqlglot/dialects/clickhouse.py
cpcloud: It's already set to `True`. Let me see how weird things get with working on the base op.

sqlglot/dialects/clickhouse.py
georgesittas: Yeah, good call. Take a look at https://github.com/tobymao/sqlglot/pull/4007#discussion_r1736255398 as well.

sqlglot/dialects/clickhouse.py
georgesittas: Ah, nice. Yeah, I meant fiddling with the base methods and see if that allows us to avoid overriding them.

cpcloud: Ok, not too bad hopefully!

georgesittas (approved):
Thanks!

georgesittas:
@cpcloud I think there's actually an issue here:

```python
>>> import sqlglot
>>> sqlglot.parse_one("""
... SELECT * FROM CTE1
... EXCEPT
... SELECT * FROM CTE2
... """, "clickhouse").sql("clickhouse")
'SELECT * FROM CTE1 EXCEPT DISTINCT SELECT * FROM CTE2'
```

`EXCEPT` without `DISTINCT` is, as I think you mentioned, `EXCEPT ALL`:

```
WITH CTE1 AS (
    SELECT 1 AS id, 'Alice' AS name
    UNION ALL
    SELECT 2 AS id, 'Bob' AS name
    UNION ALL
    SELECT 3 AS id, 'Charlie' AS name
    UNION ALL
    SELECT 4 AS id, 'Diana' AS name
    UNION ALL
    SELECT 5 AS id, 'Eve' AS name
    UNION ALL
    SELECT 1 AS id, 'Alice' AS name
),
CTE2 AS (
    SELECT 3 AS id, 'Charlie' AS name
    UNION ALL
    SELECT 4 AS id, 'Diana' AS name
    UNION ALL
    SELECT 6 AS id, 'Frank' AS name
)

-- Produces [(1, 'Alice'), (2, 'Bob'), (1, 'Alice'), (5, 'Eve')]
SELECT * FROM CTE1
EXCEPT
SELECT * FROM CTE2;
```

georgesittas:
Fixed in #4009

Issue #4003, opened 2024-08-29
Presto does not support alias for CREATE VIEW


usmanovbf:
Hi! 

I am trying to add new keywords to the presto parser, such as [SECURITY DEFINER](https://trino.io/docs/current/sql/create-view.html#synopsis), and if I create custom _parse_create() method in [presto.py ](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/presto.py) by the [example](https://github.com/tobymao/sqlglot/blob/fcaae87d50653600048917945652cdf463bcc3cf/sqlglot/parser.py#L1720) and implement parsing the [syntax](https://trino.io/docs/current/sql/create-view.html#synopsis), I get the error, because I don't neeed to obtain alias for CREATE VIEW - Presto does not support that, e.g. `(cola)`. So, I need to remove the failing test below. If you agree, let's delete it.

Otherwise, I will need to implement custom _parse_create() method in [presto.py ](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/presto.py) with non-existing alias processing - it will be strange  

Dialect - presto

**Fully reproducible code snippet**
[It](https://github.com/tobymao/sqlglot/blob/fcaae87d50653600048917945652cdf463bcc3cf/tests/dialects/test_presto.py#L509_515) should be deleted, right?

``` python
        self.validate_all(
            "CREATE OR REPLACE VIEW x (cola) SELECT 1 as cola",
            write={
                "spark": "CREATE OR REPLACE VIEW x (cola) AS SELECT 1 AS cola",
                "presto": "CREATE OR REPLACE VIEW x AS SELECT 1 AS cola",
            },
        )
```

**Official Documentation**
There is no alias for that https://trino.io/docs/current/sql/create-view.html#synopsis

georgesittas:
Fixed by #4008

PR #4008, opened 2024-08-29
feat(presto): support [ SECURITY { DEFINER | INVOKER } ]

usmanovbf:
Reference: https://trino.io/docs/current/sql/create-view.html#synopsis

Based on the comments from the https://github.com/tobymao/sqlglot/issues/4003#issuecomment-2317376237


tests/dialects/test_presto.py
```diff
@@ -522,6 +522,20 @@ def test_ddl(self):
...
+            },
+        )
+
+        self.validate_all(
+            """CREATE OR REPLACE VIEW v SECURITY INVOKER AS SELECT id FROM t""",
+            write={
+                "presto": """CREATE OR REPLACE VIEW v SECURITY INVOKER AS SELECT id FROM t""",
+            },
+        )
+
```
georgesittas, line 538: Let's simplify these by using `self.validate_identity(...)`.

sqlglot/tokens.py
```diff
@@ -239,6 +239,7 @@ class TokenType(AutoName):
     CURRENT_USER = auto()
     DECLARE = auto()
     DEFAULT = auto()
+    DEFINER = auto()
```
georgesittas, line 242: Let's remove these three tokens, they're not necessary.

sqlglot/dialects/presto.py
```diff
@@ -345,6 +348,17 @@ class Parser(parser.Parser):
         FUNCTION_PARSERS = parser.Parser.FUNCTION_PARSERS.copy()
         FUNCTION_PARSERS.pop("TRIM")
 
+        PROPERTY_PARSERS: t.Dict[str, t.Callable] = {
+            **parser.Parser.PROPERTY_PARSERS,
+            "SECURITY": lambda self: self._parse_security(),
+        }
+
+        def _parse_security(self) -> t.Optional[exp.SecurityProperty]:
+            if self._match_set((TokenType.DEFINER, TokenType.INVOKER)):
```
georgesittas, line 357: You can use `self._match_texts(("DEFINER", "INVOKER"))` instead (see comment re: getting rid of tokens).

sqlglot/dialects/presto.py
```diff
@@ -345,6 +348,17 @@ class Parser(parser.Parser):
         FUNCTION_PARSERS = parser.Parser.FUNCTION_PARSERS.copy()
         FUNCTION_PARSERS.pop("TRIM")
 
+        PROPERTY_PARSERS: t.Dict[str, t.Callable] = {
+            **parser.Parser.PROPERTY_PARSERS,
+            "SECURITY": lambda self: self._parse_security(),
+        }
+
```
georgesittas, line 355: Let's move this to the base parser class

sqlglot/dialects/presto.py
```diff
@@ -281,6 +281,9 @@ class Tokenizer(tokens.Tokenizer):
             "IPPREFIX": TokenType.IPPREFIX,
             "TDIGEST": TokenType.TDIGEST,
             "HYPERLOGLOG": TokenType.HLLSKETCH,
+            "SECURITY": TokenType.SECURITY,
+            "DEFINER": TokenType.DEFINER,
+            "INVOKER": TokenType.INVOKER,
```
georgesittas, line 286: We can get rid of these too.

sqlglot/dialects/presto.py
```diff
@@ -367,6 +381,7 @@ class Generator(generator.Generator):
             **generator.Generator.PROPERTIES_LOCATION,
             exp.LocationProperty: exp.Properties.Location.UNSUPPORTED,
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
+            exp.SecurityProperty: exp.Properties.Location.POST_SCHEMA,
```
georgesittas, line 384: This & the `TRANSFORMS` entry can also be moved to the base generator class.

georgesittas (commented):
Thanks for the PR @usmanovbf!

sqlglot/dialects/presto.py
usmanovbf: But it is not supported by the generic sql syntax. It is supported by presto only, AFAIK. Are you sure? It breaks the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) recommendation

georgesittas: I hear you, but this pattern is common in SQLGlot, so I'm suggesting this for consistency.

We do this when the logic doesn't clash with existing logic to make the base classes handle more SQL, which has two side-effects: allows for dialects to automatically inherit logic, usually making things work out-of-the-box, and making `parse_one` and `sql` more lenient, which is a nice QOL since you can play around without always setting a dialect.

georgesittas (approved):
Appreciate the quick turnaround!

PR #4009, opened 2024-08-29
Fix!: refactor SET OPERATION handling to set correct defaults

georgesittas:
Main motivation was to address https://github.com/tobymao/sqlglot/pull/4007#issuecomment-2317930198, but later down the line I realized the way we parse & generate the `DISTINCT` and `ALL` specifiers (e.g. default values) was a bit brittle. So I took a stab at refactoring this, e.g. improved "unsupported" warnings.

<details>
  <summary>Click this to see references (docs etc)</summary>

BigQuery:
- Needs specifier for all set operations
- EXCEPT, DISTINCT don’t work with ALL
- https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#set_operators

ClickHouse:
- Needs specifier for UNION, unless union_default_mode is set
- INTERSECT, EXCEPT are ALL by default
- https://clickhouse.com/docs/en/sql-reference/statements/select/union
- https://clickhouse.com/docs/en/sql-reference/statements/select/intersect
- https://clickhouse.com/docs/en/sql-reference/statements/select/except

DataBricks, Spark:
- All set operations are DISTINCT by default
- https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-setops.html
- https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-setops.html

DuckDB:
- All set operations are DISTINCT by default
- https://duckdb.org/docs/sql/query_syntax/setops.html

Hive:
- Seems to only support UNION?
- UNION is DISTINCT by default
- https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Union#LanguageManualUnion-UnionSyntax

MySQL:
- All set operations are DISTINCT by default
- https://dev.mysql.com/doc/refman/8.4/en/set-operations.html

Oracle:
- All set operations are DISTINCT by default
- https://docs.oracle.com/cd/B19306_01/server.102/b14200/queries004.htm#i2054381

Postgres:
- All set operations are DISTINCT by default
- https://www.postgresql.org/docs/current/queries-union.html

Presto, Trino:
- All set operations are DISTINCT by default
- EXCEPT, DISTINCT don’t work with ALL
- https://prestodb.io/docs/current/sql/select.html#union-intersect-except-clause
- https://trino.io/docs/current/sql/select.html#union-clause

Redshift:
- All set operations are DISTINCT by default
- EXCEPT, DISTINCT don’t work with ALL
- https://docs.aws.amazon.com/redshift/latest/dg/r_UNION.html

Snowflake:
- All set operations are DISTINCT by default
- EXCEPT, DISTINCT don’t work with ALL
- https://docs.snowflake.com/en/sql-reference/operators-query

SQLite:
- All set operations are DISTINCT by default
- EXCEPT, DISTINCT don’t work with ALL
- https://www.sqlite.org/draft/lang_select.html

Starrocks:
- All set operations are DISTINCT by default
- EXCEPT, DISTINCT don’t work with ALL
- https://docs.starrocks.io/docs/sql-reference/sql-statements/data-manipulation/SELECT/#union

T-SQL:
- All set operations are DISTINCT by default
- EXCEPT, DISTINCT don’t work with ALL
- https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-union-transact-sql?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-except-and-intersect-transact-sql?view=sql-server-ver16

</details>

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -275,6 +275,9 @@ class BigQuery(Dialect):
     # https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table
     PSEUDOCOLUMNS = {"_PARTITIONTIME", "_PARTITIONDATE"}
 
+    # All set operations require either a DISTINCT or ALL specifier
+    SET_OP_DISTINCT_BY_DEFAULT = dict.fromkeys((exp.Except, exp.Intersect, exp.Union), None)
+
     def normalize_identifier(self, expression: E) -> E:
         if (
             isinstance(expression, exp.Identifier)
@@ -633,6 +636,7 @@ class BigQuery(Dialect):
         HEX_FUNC = "TO_HEX"
         WITH_PROPERTIES_PREFIX = "OPTIONS"
         SUPPORTS_EXPLODING_PROJECTIONS = False
+        EXCEPT_INTERSECT_SUPPORT_ALL_CLAUSE = False
 
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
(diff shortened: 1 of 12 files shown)


cpcloud:
Thank you!

tobymao (approved):

PR #4014, opened 2024-08-30
fix(mysql)!: Preserve roundtrip of %a, %W time formats

### LINKED ISSUE #4011: Mysql dialect's TIME_MAPPING incorrectly maps %W
**Before you file an issue**

%a in the query converted to %W.
E.g. `select date_format(timestamp('2024-08-22 14:53:12'), '%a')`
output is converted to:
`select date_format(timestamp('2024-08-22 14:53:12'), '%W')`

mysql> select date_format(timestamp('2024-08-22 14:53:12'), '%W');
+-----------------------------------------------------+
| date_format(timestamp('2024-08-22 14:53:12'), '%W') |
+-----------------------------------------------------+
| Thursday                                            |
+-----------------------------------------------------+
1 row in set (0.02 sec)

mysql> select date_format(timestamp('2024-08-22 14:53:12'), '%a');
+-----------------------------------------------------+
| date_format(timestamp('2024-08-22 14:53:12'), '%a') |
+-----------------------------------------------------+
| Thu                                                 |
+-----------------------------------------------------+
1 row in set (0.00 sec)

**Official Documentation**
https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format
https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

According to the python `strftime` formats documentation:
%A is Weekday as locale’s full name.(Sunday, Monday...)
%a is `Weekday as locale’s abbreviated name` (Sun .. Sat)

Let’s look at mysql:
%W Weekday name (Sunday..Saturday)
%a:Abbreviated weekday name (Sun .. Sat)

So, it is expected that the date_format output in presto or mysql should map %W to %A, as follows:
`mysql.py`
```
time_mapping = {
"%M": "%B",
"%c": "%-m",
"%e": "%-d",
"%h": "%I",
"%i": "%M",
"%s": "%S",
"%S": "%S",
"%u": "%W",
"%k": "%-H",
"%l": "%-I",
"%T": "%H:%M:%S",
- "%W": "%a",
+ "%W": "%A",
}
```
I observed that several other mappings worked well. If applied, I noticed that mysql and snowflack were affected in the single test. Then this may affect more, such as drill, hive, oracle, teradata, which also use %a to represent the pattern of Weekday as locale’s full name.

Why is this the case?

In addition, there is a PR #3855 in presto that does the conversion. Is this a workaround?
@VaggelisD PTAL

VaggelisD:
Fixes #4011 

For reference:

```
Python's strftime:
- Uses %a for abbreviated weekday (Mon, Sun)
- Uses %A for full weekday (Monday, Sunday)

MySQL's format:
- Uses %a for abbreviated weekday (Mon, Sun)
- Uses %W for full weekday (Monday, Sunday)

Snowflake's format for conversion functions:
- Uses DY for abbreviated weekday (Mon, Sun)
- Does not have a full weekday element
```

This PR fixes the following issues:
1. SQLGlot would change the `%a` to `%W` during MySQL's RTT:

```
sqlglot.parse_one("SELECT DATE_FORMAT('2024-08-22 14:53:12', '%a')", read="mysql").write("mysql")
SELECT DATE_FORMAT('2024-08-22 14:53:12', '%W')
```

2. SQLGlot would transpile MySQL's `%W` (full weekday) to Snowflake's `DY` (abbreviated weekday); The correct transpilation is `%a -> DY`, whereas `%W` does not have a proper Snowflake counterpart for conversion functions

Docs
--------
[Python strftime](https://strftime.org/) | [MySQL Date formats](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format) | [Snowflake formats in conversion functions](https://docs.snowflake.com/en/sql-reference/functions-conversion#label-date-time-format-conversion)

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -173,7 +173,7 @@ class MySQL(Dialect):
         "%k": "%-H",
         "%l": "%-I",
         "%T": "%H:%M:%S",
-        "%W": "%a",
+        "%W": "%A",
     }
 
     class Tokenizer(tokens.Tokenizer):
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -240,10 +240,7 @@ class Presto(Dialect):
     TABLESAMPLE_SIZE_IS_PERCENT = True
     LOG_BASE_FIRST: t.Optional[bool] = None
 
-    TIME_MAPPING = {
-        **MySQL.TIME_MAPPING,
-        "%W": "%A",
-    }
+    TIME_MAPPING = MySQL.TIME_MAPPING
 
     # https://github.com/trinodb/trino/issues/17
     # https://github.com/trinodb/trino/issues/12289
--- a/tests/dialects/test_mysql.py
+++ b/tests/dialects/test_mysql.py
@@ -539,9 +539,16 @@ class TestMySQL(Validator):
             },
         )
         self.validate_all(
-            "SELECT DATE_FORMAT('2009-10-04 22:23:00', '%W %M %Y')",
+            "SELECT DATE_FORMAT('2024-08-22 14:53:12', '%a')",
             write={
-                "mysql": "SELECT DATE_FORMAT('2009-10-04 22:23:00', '%W %M %Y')",
+                "mysql": "SELECT DATE_FORMAT('2024-08-22 14:53:12', '%a')",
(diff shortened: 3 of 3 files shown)


hellozepp (approved):
LGTM

georgesittas (approved):

PR #3999, opened 2024-08-29
fix(starrocks): exp.Unnest transpilation

hellozepp:
Fixes #3962 

Link the before PR: 
- https://github.com/tobymao/sqlglot/pull/3964
- https://github.com/tobymao/sqlglot/pull/3966

This PR adds support for the following:

Default initialize exp.UNNEST to include "unnest" as the default table alias if not specified
Added expression type of exp.Inline
use arrays_zip to merge multiple Lateral views
Added some ut to validate unnest transpile

PR #4016, opened 2024-08-30
feat!: Add UnsupportedError to unnest_to_explode transform

VaggelisD:
Follow up to #3999

Refactor `generator` arg out of `unnest_to_explode()`

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import typing as t
 
-from sqlglot import expressions as exp, generator
+from sqlglot import expressions as exp
+from sqlglot.errors import UnsupportedError
 from sqlglot.helper import find_new_name, name_sequence
+
 
 if t.TYPE_CHECKING:
     from sqlglot.generator import Generator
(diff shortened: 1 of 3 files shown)


sqlglot/transforms.py
```diff
@@ -308,11 +312,7 @@ def _unnest_zip_exprs(
     ) -> t.List[exp.Expression]:
         if has_multi_expr:
             if not unnest_using_arrays_zip:
-                if generator:
-                    generator.unsupported(
-                        f"Multiple expressions in UNNEST are not supported in "
-                        f"{generator.dialect.__module__.split('.')[-1].upper()}"
-                    )
+                raise UnsupportedError("Cannot transpile UNNEST with multiple arrays to Hive")
```
georgesittas, line 315: This is dialect-agnostic:
```suggestion
                raise UnsupportedError("Cannot transpile UNNEST with multiple arrays")
```

sqlglot/transforms.py
```diff
@@ -27,9 +29,12 @@ def preprocess(
...
 
-        expression = transforms[0](expression)
-        for transform in transforms[1:]:
-            expression = transform(expression)
+        try:
+            expression = transforms[0](expression)
+            for transform in transforms[1:]:
+                expression = transform(expression)
+        except UnsupportedError as unsupported_error:
+            self.unsupported(str(unsupported_error))
```
georgesittas, line 37: Did you test that this actually logs the correct warning? I don't remember if `str(...)` is the correct way to get the message.

georgesittas (approved):

sqlglot/transforms.py
VaggelisD: Yep, here's an example:

```
>>> sqlglot.parse_one('SELECT * FROM example_table CROSS JOIN UNNEST(split(type, ";"), scores)', read="starrocks").sql("hive")
Cannot transpile UNNEST with multiple arrays
"SELECT * FROM example_table CROSS JOIN EXPLODE(SPLIT(type, CONCAT('\\\\Q', ';')), scores, unnest(unnest))"
```

georgesittas: Cool thanks for double checking

PR #3998, opened 2024-08-29
fix(starrocks): exp.Create transpilation

hellozepp:
Fix #3997 

Supports starrocks create table parse and generator, generator supports starrocks and hive transpile (may need to slowly improve it).

StarRocks specifically includes:
- Supports `DISTRIBUTED BY`, `DUPLICATE KEY`, `PARTITION BY`, `PROPERTIES` to be parsed correctly
- In `_parse_create`, move exp.PrimaryKey outside the create schema to the schema

Hive specifically includes:
- Supports `DISTRIBUTED BY`, `DUPLICATE KEY`, `PARTITION BY`, `PROPERTIES` to be generated correctly
- Mark `DuplicateKey` with `UNSUPPORTED`
- Convert DataType to adapt to hive

Parse specifically includes:
- Added dialect in _parse_create so that the type of a certain engine feature (starrocks) can be processed in a targeted manner

PR #4023, opened 2024-08-31
fix(starrocks): exp.Create transpilation

### LINKED ISSUE #3997: Incorect tranformation of create table using Starrocks
**Before you file an issue**
sqlglot.transpile(sql, read="starrocks", write="hive") # or  write="spark"

```
CREATE TABLE if not exists `sample_table` (
    `tenantid` varchar(1048576) NULL COMMENT "",
    `create_day` date NOT NULL COMMENT "",
    `shopsite` varchar(65533) NOT NULL COMMENT "shopsite",
    `id` varchar(65533) NOT NULL COMMENT "shopsite id",
    `price` decimal128(38, 10) NULL COMMENT "test the bigdecimal",
    `seq` int(11) NULL COMMENT "order",
    `use_status` smallint(6) NULL COMMENT "0,1",
    `created_user` bigint(20) NULL COMMENT "create user",
    `created_time` datetime NULL COMMENT "create time",
) ENGINE=OLAP
DUPLICATE KEY(tenantid)
PRIMARY KEY(`tenantid`, `shopsite`, `id`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`tenantid`, `shopsite`, `id`) BUCKETS 10
ORDER BY (`tenantid`, `shopsite`, `id`)
PROPERTIES (
    "replication_num" = "1",
    "in_memory" = "false",
    "enable_persistent_index" = "false",
    "replicated_storage" = "false",
    "storage_medium" = "HDD",
    "compression" = "LZ4"
)
```

The following create table statement of `starrocks` will report an error during `parse_one`:
```
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/parts/trans.py", line 74, in <module>
    raise ex
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/parts/trans.py", line 67, in <module>
    test_transpile_sql(os.path.join(dir, '../' + sys.argv[0]), read='starrocks', dialect='hive', pretty=True)
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/parts/trans.py", line 29, in test_transpile_sql
    raise ex
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/parts/trans.py", line 25, in test_transpile_sql
    transpiled_sql = sqlglot.transpile(sql, read=read, write=dialect)[0].strip()
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/__init__.py", line 174, in transpile
    for expression in parse(sql, read, error_level=error_level)
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/__init__.py", line 99, in parse
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/dialects/dialect.py", line 894, in parse
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 1366, in parse
    return self._parse(
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 1435, in _parse
    expressions.append(parse_method(self))
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 1667, in _parse_statement
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 758, in <lambda>
    TokenType.CREATE: lambda self: self._parse_create(),
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 1820, in _parse_create
    this = self._parse_schema(this=table_parts)
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 5225, in _parse_schema
    self._match_r_paren()
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 6956, in _match_r_paren
  File "/Users/zhanglin/PycharmProjects/sqlglot_master/sqlglot/parser.py", line 1479, in raise_error
    raise error
sqlglot.errors.ParseError: Expecting ). Line 6, Col: 19.
  3) NOT NULL COMMENT "shopsite",
    `id` varchar(65533) NOT NULL COMMENT "shopsite id",
    `price` decimal128(38, 10) NULL COMMENT "test the bigdecimal",
    `seq` int(11) NULL COMMENT "order",
    `use_status
```

hellozepp:
Fix #3997

> We closed #3998 because it was too convoluted, so I split out Starrocks own create table transpile fix here.

Supports starrocks create table parse and generator

This PR adds support for the following:
- Supports `DISTRIBUTED BY HASH|RANDOM`, `DUPLICATE KEY`, `PROPERTIES` to be parsed correctly
- I standardized the exp.PrimaryKey of Starrocks. Restore the PrimaryKey in properties to the schema to ensure that the dialect can read it correctly

## Docs

[starrocks distributed Doc](https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#distribution_desc)
 | [key_desc Doc](https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#key_desc)


hellozepp:
Good afternoon, @georgesittas 
I just split out the PR for the Starrocks create table transpile fix, PTAL

sqlglot/expressions.py
```diff
@@ -2603,10 +2603,32 @@ class DistKeyProperty(Property):
     arg_types = {"this": True}
 
 
+# [starrocks distributed Doc](https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#distribution_desc)
+# [doris distributed Doc](https://doris.apache.org/docs/sql-manual/sql-statements/Data-Definition-Statements/Create/CREATE-TABLE?_highlight=create&_highlight=table#distribution_desc)
```
georgesittas, line 2607: No need to use the markdown syntax, just the links are ok.

sqlglot/dialects/starrocks.py
```diff
@@ -31,6 +78,68 @@ class Parser(MySQL.Parser):
             "REGEXP": exp.RegexpLike.from_arg_list,
         }
 
+        PROPERTY_PARSERS = {
+            **MySQL.Parser.PROPERTY_PARSERS,
+            "DISTRIBUTED BY HASH": lambda self: self._parse_distributed_by("HASH"),
+            "DISTRIBUTED BY RANDOM": lambda self: self._parse_distributed_by("RANDOM"),
+            "PROPERTIES": lambda self: self._parse_wrapped_csv(self._parse_property),
+            "DUPLICATE KEY": lambda self: self._parse_duplicate(),
+            "PARTITION BY": lambda self: self._parse_partitioned_by(),
```
georgesittas, line 87: Why are `PROPERTIES` and `PARTITION BY` here?

sqlglot/dialects/starrocks.py
```diff
@@ -13,11 +13,58 @@
...
 
 
 class StarRocks(MySQL):
     STRICT_JSON_PATH_SYNTAX = False
 
+    class Tokenizer(MySQL.Tokenizer):
+        KEYWORDS = {
+            **MySQL.Tokenizer.KEYWORDS,
+            "DECIMAL64": TokenType.DECIMAL,
+            "DECIMAL128": TokenType.BIGDECIMAL,
```
georgesittas, line 62: We need to test these mappings. Can you share docs to justify these entries?

sqlglot/dialects/starrocks.py
```diff
@@ -13,11 +13,58 @@
...
+            "DISTRIBUTED BY HASH": TokenType.DISTRIBUTE_BY,
+            "DISTRIBUTED BY RANDOM": TokenType.DISTRIBUTE_BY,
+            "DUPLICATE KEY": TokenType.DUPLICATE_KEY,
```
georgesittas, line 65: Let's not do this multi-word token mapping here. I'd just removing the new `DUPLICATE_KEY` token and only match on the text. So you'll have `"DISTRIBUTED": lambda self: self._parse_distributed_property()` and the same for `"DUPLICATE"`. Then, within those parsing methods you can do a `_match_text_seq` to choose which property to parse.

sqlglot/dialects/starrocks.py
```diff
@@ -13,11 +13,58 @@
...
+    buckets = self.sql(expression, "buckets")
+    if expression.auto_bucket:
+        buckets = "AUTO"
+    # [DISTRIBUTED BY RANDOM (k1[,k2 ...]) [BUCKETS num] [ORDER BY (k1[,k2 ...])]]
+    return f"DISTRIBUTED BY RANDOM ({expressions}) BUCKETS {buckets}{sorted_by}"
+
+
+def _property_sql(self: MySQL.Generator, expression: exp.Property) -> str:
+    # [PROPERTIES (['key1'='value1', 'key2'='value2', ...])]
+    return f"{self.property_name(expression, string_key=True)}={self.sql(expression, 'value')}"
```
georgesittas, line 52: What is this achieving? I'm not sure if it's correct, `Property` is a pretty generic class to override.

sqlglot/expressions.py
```diff
@@ -2603,10 +2603,32 @@ class DistKeyProperty(Property):
...
+class DistributedByHashProperty(Property):
+    arg_types = {"expressions": True, "buckets": False, "sorted_by": False}
+
+    @property
+    def auto_bucket(self) -> bool:
+        return self.args.get("buckets") is None
+
+
+class DistributedByRandomProperty(Property):
+    arg_types = {"expressions": True, "buckets": False, "sorted_by": False}
```
georgesittas, line 2617: Can't we have an "order" arg instead of `"sorted_by"` and then use `_parse_order` to capture the syntax? With this approach we're duplicating the generation for `ORDER BY`.

sqlglot/expressions.py
```diff
@@ -2603,10 +2603,32 @@ class DistKeyProperty(Property):
...
 
 
+# [starrocks distributed Doc](https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#distribution_desc)
+# [doris distributed Doc](https://doris.apache.org/docs/sql-manual/sql-statements/Data-Definition-Statements/Create/CREATE-TABLE?_highlight=create&_highlight=table#distribution_desc)
+class DistributedByHashProperty(Property):
+    arg_types = {"expressions": True, "buckets": False, "sorted_by": False}
+
+    @property
+    def auto_bucket(self) -> bool:
+        return self.args.get("buckets") is None
```
georgesittas, line 2613: No need for this property, same for the expression below.

sqlglot/dialects/starrocks.py
```diff
@@ -31,6 +78,68 @@ class Parser(MySQL.Parser):
...
+            "PROPERTIES": lambda self: self._parse_wrapped_csv(self._parse_property),
+            "DUPLICATE KEY": lambda self: self._parse_duplicate(),
+            "PARTITION BY": lambda self: self._parse_partitioned_by(),
+        }
+
+        def _parse_create(self) -> exp.Create | exp.Command:
+            create = super()._parse_create()
+
+            # Starrocks's primary is defined outside the schema, so we need to move it into the schema
+            # https://docs.starrocks.io/docs/table_design/table_types/primary_key_table/#usage
```
georgesittas, line 94: Why do we need to move it inside the schema? What happens if we don't?

sqlglot/dialects/starrocks.py
```diff
@@ -31,6 +78,68 @@ class Parser(MySQL.Parser):
...
+            return create
+
+        def _parse_distributed_by(self, type: str) -> exp.Expression:
+            expressions = self._parse_wrapped_csv(self._parse_id_var)
+
+            # If the BUCKETS keyword not present, the number of buckets is AUTO
+            # [ BUCKETS AUTO | BUCKETS <number> ]
+            buckets = None
+            if self._match_text_seq("BUCKETS", "AUTO"):
+                pass
```
georgesittas, line 121:
```suggestion
            if self._match_text_seq("BUCKETS", "AUTO"):
                buckets: t.Optional[exp.Expression] = None
```

sqlglot/dialects/starrocks.py
```diff
@@ -31,6 +78,68 @@ class Parser(MySQL.Parser):
...
+            elif self._match_text_seq("BUCKETS"):
+                buckets = self._parse_number()
+
+            if self._match_text_seq("ORDER BY"):
+                order_by = self._parse_wrapped_csv(self._parse_ordered)
+            else:
+                order_by = None
```
georgesittas, line 128: We can do this with a single `_parse_order`, see other comment.

sqlglot/dialects/starrocks.py
```diff
@@ -54,6 +163,7 @@ class Generator(MySQL.Generator):
         EXCEPT_INTERSECT_SUPPORT_ALL_CLAUSE = False
         JSON_TYPE_REQUIRED_FOR_EXTRACTION = False
         PARSE_JSON_NAME: t.Optional[str] = "PARSE_JSON"
+        WITH_PROPERTIES_PREFIX = "PROPERTIES"
```
georgesittas, line 166: Just making sure - starrocks doesn't have a `WITH (k1=v1, k2=v2, ...)` DDL property syntax?

sqlglot/dialects/starrocks.py
hellozepp: `PARTITION BY` is unnecessary here, I deleted it.

`PROPERTIES` If you don't add it in PROPERTY_PARSERS, you will get an error:
```
'CREATE TABLE if not exists `sample_table` (
`tenantid` varchar(1048576) NULL COMMENT "",
`cr' contains unsupported syntax. Falling back to parsing as a 'Command'.
```

<Token token_type: TokenType.VAR, text: PROPERTIES, line: 18, col: 10, start: 734, end: 743, comments: []>
The error is `PROPERTIES (`. It seems that there is no `TBLPROPERTIES` in the hive code. This is strange 🤔

hellozepp: Because `PROPERTIES` is also often used in doris's create table statement.
https://doris.apache.org/docs/sql-manual/sql-statements/Data-Definition-Statements/Create/CREATE-TABLE?_highlight=create&_highlight=table#properties

sqlglot/dialects/starrocks.py
hellozepp: Oh ok, I'll add it here.
```
- Several types of decimal: https://docs.starrocks.io/docs/unloading/Flink_connector/#data-type-mapping-between-starrocks-and-flink
> Focus on the starrocks column

- Use:
https://docs.starrocks.io/docs/sql-reference/sql-functions/math-functions/round/#parameters
```

I will also add ut to test these mappings

sqlglot/dialects/starrocks.py
hellozepp: Thank you, good suggestion, I will try to fix it

sqlglot/dialects/starrocks.py
hellozepp: Yes. This case is used to set the parameter string_key to True.
I refer to the code from `sqlglot.dialects.hive._property_sql`.

hellozepp:
@georgesittas Thanks for your patience in giving me advice. It's worth digesting.
I'll study it in detail.

sqlglot/dialects/starrocks.py
hellozepp: Yes, in the documentation it provides the pattern for create.
https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#syntax
There is no other pattern provided in `Parameters` - `properties` either.

When using `with` to execute in StarRocks, it will get the error:
image


```
Caused by: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'with' at line 9
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:121)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.StatementImpl.executeInternal(StatementImpl.java:767)
	at com.mysql.cj.jdbc.StatementImpl.execute(StatementImpl.java:652)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.execute(JDBCStatementImpl.java:330)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.executeStatement(JDBCStatementImpl.java:131)
	... 12 more

File: sqlglot/expressions.py (hunk 7 above)
Comment (hellozepp) [reply sqlglot/expressions.py:2613]: Got it.

File: sqlglot/dialects/starrocks.py (hunk 3 above)
Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:62]: According to the documentation, I additionally added `DECIMAL32`

Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:62]: In starrocks, it is executed as follows:
image

File: sqlglot/dialects/starrocks.py (hunk 8 above)
Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:94]: If we do not change the outside primary key pattern, when we convert to hive (PRIMARY KEY starting from 2.1.0), hive only supports PRIMARY KEY specified inside the schema.
see: https://issues.apache.org/jira/browse/HIVE-13290

File: sqlglot/dialects/starrocks.py (hunk 2 above)
Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:87]: `TBLPROPERTIES` is defined in sqlglot/parser.py:865, which allows hive to parse it well:
```
"TBLPROPERTIES": lambda self: self._parse_wrapped_properties(),
```
We are missing PROPERTIES, so I added `PROPERTIES` here

Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:87]: create table `properties` docs links:
https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#properties

In addition, I did not put Properties into parse because it may affect other dialects. The following test will fail:
```
self.validate_identity(
"ALTER TABLE people SET PROPERTIES x = 'y'", check_command_warning=True
)
```

**Comment by hellozepp** [comment]
@georgesittas I submitted a new commit, PTAL

File: sqlglot/dialects/starrocks.py (hunk 8 above)
Comment (georgesittas) [reply sqlglot/dialects/starrocks.py:59]: That sounds like something that should be fixed in Hive's generator, not @ parse time for starrocks. Make sure to also test spark, databricks.

Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:59]: Spark SQL does not support the setting of primary key. I used Spark 3.3 to test:
```
Exception in thread "main" org.apache.spark.sql.catalyst.parser.ParseException: 
Syntax error at or near 'PRIMARY'(line 2, pos 2)

== SQL ==
CREATE TABLE data_integration.foo2 (col1 int
) PRIMARY KEY (`col1`)
--^^^
;

	at org.apache.spark.sql.catalyst.parser.ParseException.withCommand(ParseDriver.scala:306)
	at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parse(ParseDriver.scala:143)
	at org.apache.spark.sql.execution.SparkSqlParser.parse(SparkSqlParser.scala:52)
	at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parsePlan(ParseDriver.scala:89)
	at org.apache.spark.sql.SparkSession.$anonfun$sql$2(SparkSession.scala:620)
	at org.apache.spark.sql.catalyst.QueryPlanningTracker.measurePhase(QueryPlanningTracker.scala:111)
	at org.apache.spark.sql.SparkSession.$anonfun$sql$1(SparkSession.scala:620)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:779)
	at org.apache.spark.sql.SparkSession.sql(SparkSession.scala:617)
	at org.apache.spark.sql.SQLContext.sql(SQLContext.scala:651)
	at sparkDemo.sql.SQLDemo$.main(SQLDemo.scala:35)
	at sparkDemo.sql.SQLDemo.main(SQLDemo.scala)
```

Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:59]: @georgesittas Oh, I understand what you mean. If this is a feature of the syntax itself, then it should be adapted in other generators instead of being fixed as a syntax bug, right?

But it seems that this pattern of specifying the primary key outside is not a universal standard except for doris and starrocks. 🤔

Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:59]: For example, OLTP such as MySQL and PostgreSQL are all set up inside.

Comment (hellozepp) [reply sqlglot/dialects/starrocks.py:59]: If we are sure that this approach is not reasonable, we can delete it. I think I can submit a separate PR to fit it later =)

File: tests/dialects/test_starrocks.py [hunk 12]
```diff
@@ -5,6 +5,61 @@
...
 
+    def test_ddl(self):
+        # Test the different wider DECIMAL types
+        self.validate_identity(
+            "CREATE TABLE foo (col0 DECIMAL(9, 1), col1 DECIMAL32(9, 1), col2 DECIMAL64(18, 10), col3 DECIMAL128(38, 10)) DISTRIBUTED BY HASH (col1) BUCKETS 1"
+        )
+        # Test the PRIMARY KEY with ENGINE property
+        self.validate_identity(
+            "CREATE TABLE foo2 (col1 BIGINT) ENGINE=OLAP PRIMARY KEY (col1) DISTRIBUTED BY HASH (col1) BUCKETS 1"
+        )
```
VaggelisD, line 16: Nit, I'm in favor of reducing noise by condensing similar test cases through loops; For example, you can have a for-loop for the different partition schemes, for the existence (or not) of engine/buckets and so on. 

This helps both extending old cases and writing new ones because it's much easier to spot which keywords are tested instead of skimming through individual identity checks.

sqlglot/expressions.py
```diff
@@ -2603,10 +2603,24 @@ class DistKeyProperty(Property):
...
 
 
+# https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#distribution_desc
+# https://doris.apache.org/docs/sql-manual/sql-statements/Data-Definition-Statements/Create/CREATE-TABLE?_highlight=create&_highlight=table#distribution_desc
+class DistributedByHashProperty(Property):
+    arg_types = {"expressions": True, "buckets": False, "order": False}
+
+
+class DistributedByRandomProperty(Property):
+    arg_types = {"expressions": False, "buckets": False, "order": False}
```
VaggelisD, line 2613: I'd prefer merging these in a generic `class DistributedByProperty` using a `kind` keyword since they share similar args, position etc

VaggelisD (commented):
Thanks for the PR once again @hellozepp, I also left some comments for my first pass.

sqlglot/dialects/starrocks.py
```diff
@@ -64,6 +138,14 @@ class Generator(MySQL.Generator):
             exp.DataType.Type.TIMESTAMPTZ: "DATETIME",
         }
 
+        PROPERTIES_LOCATION = {
+            **MySQL.Generator.PROPERTIES_LOCATION,
+            exp.DistributedByHashProperty: exp.Properties.Location.POST_SCHEMA,
+            exp.DistributedByRandomProperty: exp.Properties.Location.POST_SCHEMA,
+            exp.DuplicateKeyProperty: exp.Properties.Location.POST_SCHEMA,
```
georgesittas, line 145: These already defined in the base generator.
```suggestion
```

sqlglot/dialects/starrocks.py
```diff
@@ -31,6 +45,65 @@ class Parser(MySQL.Parser):
             "REGEXP": exp.RegexpLike.from_arg_list,
         }
 
+        PROPERTY_PARSERS = {
+            **MySQL.Parser.PROPERTY_PARSERS,
+            "DISTRIBUTED": lambda self: self._parse_distributed_property(),
+            "DUPLICATE": lambda self: self._parse_duplicate(),
+            "PROPERTIES": lambda self: self._parse_wrapped_properties(),
```
georgesittas, line 52: Let's move these to the base parser instead, since we already did this for the corresponding generator methods.

georgesittas:
@hellozepp please keep in mind to make smaller PRs in the future.

hellozepp:
I just searched for more information about “small and focused”, and I’m sorry that this PR does have some noise in it.

I’d love to implement it in the future. Thanks for the guidance! @georgesittas

sqlglot/dialects/starrocks.py
hellozepp: I see, I'll change that.

georgesittas:
Yep, and thank you for the quick iterations, appreciate the contributions 👍

sqlglot/dialects/starrocks.py
hellozepp: Yes, I also checked mysql and postgres. It also supports the inside pattern:

```
mysql> CREATE TABLE foo1 (col1 int, PRIMARY KEY (`col1`)) ;
Query OK, 0 rows affected (0.03 sec)

mysql> CREATE TABLE foo1 (col1 int) PRIMARY KEY (`col1`);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'PRIMARY KEY (`col1`)' at line 1
```

Then, we can see this in the table_constraint of the postgresql document:
https://www.postgresql.org/docs/7.2/sql-createtable.html

sqlglot/dialects/starrocks.py
```diff
@@ -81,4 +164,27 @@ class Generator(MySQL.Generator):
...
+                    None,
+                )
+                if primary_key:
+                    schema.expressions.remove(primary_key)
+                    props = e.args.get("properties")
+                    if props:
+                        # Verify if the first one is an engine property. Is true then insert it after the engine,
+                        # otherwise insert it at the beginning
+                        if isinstance(props.expressions[0], exp.EngineProperty):
+                            props.expressions.insert(1, primary_key)
```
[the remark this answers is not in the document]
hellozepp: This is important in starrocks, otherwise if the order is wrong, we will get the following error:
image

We can notice in create syntax that [key_desc] should be after [ENGINE type], unless ENGINE type is not written, in which case key_desc should be first.
https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#syntax

sqlglot/dialects/starrocks.py
```diff
@@ -31,6 +45,65 @@ class Parser(MySQL.Parser):
...
+                        # Remove the PrimaryKey from properties
+                        props.expressions.remove(primary_key)
+
+            return create
+
+        def _parse_distributed_property(self) -> exp.Expression:
+            type: t.Optional[str] = None
+            expressions: t.Optional[t.List[exp.Expression]] = None
+            if self._match_text_seq("BY", "HASH"):
+                type = "HASH"
```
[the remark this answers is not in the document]
hellozepp: I will fix it

sqlglot/dialects/starrocks.py
hellozepp: OK.

`PROPERTIES` on line 52 is not suitable for moving, because it may affect other dialects. The following test will fail:
```
self.validate_identity(
"ALTER TABLE people SET PROPERTIES x = 'y'", check_command_warning=True
)

**Comment by hellozepp** [comment]
I have a new commit, please take a look @georgesittas

File: sqlglot/dialects/starrocks.py (hunk 15 above)
Comment (georgesittas) [reply sqlglot/dialects/starrocks.py]: Got it

File: sqlglot/dialects/starrocks.py [hunk 18]
```diff
@@ -81,4 +134,23 @@ class Generator(MySQL.Generator):
...
+                    props = e.args.get("properties")
+                    if props:
+                        # Verify if the first one is an engine property. Is true then insert it after the engine,
+                        # otherwise insert it at the beginning
+                        if isinstance(props.expressions[0], exp.EngineProperty):
+                            props.expressions.insert(1, primary_key.pop())
+                        else:
+                            props.expressions.insert(0, primary_key.pop())
+                    else:
+                        e.set("properties", exp.Properties(expressions=[primary_key.pop()]))
```
georgesittas, line 152: Can we have the `PRIMARY KEY` property at the end? If that's true then we can always just append and simplify this section. I think we should refactor it a bit regardless, e.g. one idea is to use `find` to search for an `EngineProperty` and then insert in its `.index` + 1 or something. Also if `properties` is not set you could do that beforehand so the section after it always inserts to an existing `properties` arg.

sqlglot/generator.py
```diff
@@ -3616,6 +3618,22 @@ def dictrange_sql(self, expression: exp.DictRange) -> str:
...
+    def duplicatekeyproperty_sql(self, expression: exp.DuplicateKeyProperty) -> str:
+        expressions = self.expressions(expression, flat=True)
+        options = self.expressions(expression, key="options", flat=True, sep=" ")
+        options = f" {options}" if options else ""
+        return f"DUPLICATE KEY ({expressions}){options}"
+
+    # https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#distribution_desc
+    def distributedbyproperty_sql(self, expression: exp.DistributedByProperty) -> str:
+        expressions = self.expressions(expression, flat=True)
+        expressions = " (" + expressions + ")" if expressions else ""
```
georgesittas, line 3630: Nit: let's use `wrap` or an f-string, this way to wrap with parentheses is not very consistent with what we do.

georgesittas (commented):
Way cleaner, nice work! Left a couple more comments, but should be good to go.

sqlglot/dialects/starrocks.py
hellozepp: No, key_desc always appears after engine type or first.

hellozepp: Appreciate this suggestion, I will change it

sqlglot/generator.py
hellozepp: Got it

tests/dialects/test_starrocks.py
```diff
@@ -5,6 +5,33 @@
...
+            "PRIMARY KEY (col1) DISTRIBUTED BY HASH (col1)",
+            "DUPLICATE KEY (col1, col2) DISTRIBUTED BY HASH (col1)",
+        ]
+
+        for i, properties in enumerate(ddl_sqls):
+            tbl = f"foo{i}"
+            with self.subTest(f"Testing create scheme: {properties}"):
+                self.validate_identity(
+                    f"CREATE TABLE {tbl} (col1 BIGINT, col2 BIGINT) {properties}"
+                )
```
VaggelisD, line 25: The different `tbl` name here shouldn't make a difference in these tests right? We're interested only for the properties, so we can remove the enumeration

VaggelisD (commented):
A couple more comments from me

sqlglot/generator.py
```diff
@@ -3616,6 +3618,22 @@ def dictrange_sql(self, expression: exp.DictRange) -> str:
     def dictsubproperty_sql(self, expression: exp.DictSubProperty) -> str:
         return f"{self.sql(expression, 'this')} {self.sql(expression, 'value')}"
 
+    def duplicatekeyproperty_sql(self, expression: exp.DuplicateKeyProperty) -> str:
+        expressions = self.expressions(expression, flat=True)
+        options = self.expressions(expression, key="options", flat=True, sep=" ")
+        options = f" {options}" if options else ""
+        return f"DUPLICATE KEY ({expressions}){options}"
```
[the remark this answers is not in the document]
hellozepp: I deleted it

tests/dialects/test_starrocks.py
hellozepp: Got it

georgesittas (approved):

PR #4030, opened 2024-09-02
fea(tsql): Parse ODBC datetime columns

usefulalgorithm:
Datetime columns can be specified in ODBC format: https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/date-time-and-timestamp-literals

Our use case only requires us to be able to parse it, for us it is fine if the output sql is `DATE('yyyy-mm-dd')` instead of `{d'yyyy-mm-dd'}`.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -5616,11 +5616,7 @@ class Parser(metaclass=_Parser):
         """
         self._match(TokenType.VAR)
         exp_class = self.ODBC_DATETIME_LITERALS[self._prev.text.lower()]
-        expression = self.expression(
-            exp_class=exp_class,
-            this=exp.Literal.string(self._curr.text),
-        )
-        self._match(TokenType.STRING)
+        expression = self.expression(exp_class=exp_class, this=self._parse_string())
         if not self._match(TokenType.R_BRACE):
             self.raise_error("Expected }")
         return expression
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -5599,11 +5605,39 @@ def _parse_primary_key(
...
+        Reference:
+        https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/date-time-and-timestamp-literals
+        """
+        self._match(TokenType.VAR)
+        exp_class = self.ODBC_DATETIME_LITERALS[self._prev.text.lower()]
+        expression = self.expression(
+            exp_class=exp_class,
+            this=exp.Literal.string(self._curr.text),
+        )
+        self._match(TokenType.STRING)
```
georgesittas, line 5623:
```suggestion
        expression = self.expression(exp_class=exp_class, this=self._parse_string())
```

tests/test_parser.py
```diff
@@ -854,3 +854,16 @@ def test_collate(self):
             ).find(exp.Collate)
             self.assertIsInstance(collate_node, exp.Collate)
             self.assertIsInstance(collate_node.expression, collate_pair[1])
+
+    def test_odbc_date_literals(
+        self,
+    ):
```
georgesittas, line 860:
```suggestion
    def test_odbc_date_literals(self):
```

georgesittas (commented):
Is `{d'2024-01-01'}` actually equivalent to `DATE('2024-01-01')`?

sqlglot/parser.py
usefulalgorithm: ah nice, thanks!

usefulalgorithm:
> Is `{d'2024-01-01'}` actually equivalent to `DATE('2024-01-01')`?

It should be the case.

georgesittas:
Thanks for the PR 👍

PR #4032, opened 2024-09-02
feat(oracle): Support JSON_EXISTS, refactor ON handling

### LINKED ISSUE #4026: Missing Support for JSON EXISTS in ORACLE
**Description**
I'm encountering an issue with the Oracle dialect in SQLGlot where it fails to parse queries containing JSON conditions like JSON_EXISTS. The expected behavior is to convert the Oracle query to an AST, but SQLGlot does not currently support these conditions, resulting in a parse failure.
**JSON_EXISTS() Function
Syntax:**
> JSON_EXISTS(json_document, json_path [error_clause])
**error_clause:**

- > TRUE ON ERROR: Returns TRUE if an error occurs during the evaluation.
- > FALSE ON ERROR: Returns FALSE if an error occurs during the evaluation.
- > UNKNOWN ON ERROR: Returns UNKNOWN if an error occurs during the evaluation.

**Example**
The following code demonstrates the issue:
```from sqlglot.dialects.oracle import Oracle
from sqlglot import parse_one

sql = """
SELECT name FROM t
WHERE JSON_EXISTS(name, '$[1].middle' TRUE ON ERROR);
"""

ast = parse_one(sql, dialect=Oracle)
print(repr(ast))
```
**Error Encountered**
When executing the code above, the following error is encountered:
`   raise error
sqlglot.errors.ParseError: Expecting ). Line 2, Col: 45.
 `
[ link ](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SQL-JSON-Conditions.html#GUID-57762C80-0C8A-4B18-9BA7-9B3F5ABDC988)for the document

VaggelisD:
Fixes #4026

This PR:
- Adds support for Oracles `JSON_EXIST` through the new AST node `exp.JSONExists`
- Refactors https://github.com/tobymao/sqlglot/pull/3987 to accommodate for the common options `ON EMPTY` and `ON ERROR` clauses between MySQL & Oracle

A small caveat here is that these dialects parse the `ON` options with different order e.g. MySQL requires `[ON_EMPTY] [ON_ERROR]` while Oracle expects the inverse. From my testing, MySQL will produce an error if the order is not obeyed in contrast to Oracle which seems that it can handle both, but in absence of a proper Oracle testing enviroment I chose to preserve the user order.

The toy/experiment queries where:

- Oracle
```
with t as (SELECT '[{first:"John"}, {middle:"Mark"}, {last:"Smith"}]' as name) SELECT name FROM t
  WHERE JSON_EXISTS(name, '$[0].first' TRUE ON ERROR TRUE ON EMPTY);
```

- MySQL 
```
SELECT JSON_VALUE('{"fname": "Joe", "lname": "Palmer"}', '$.fname' NULL ON EMPTY NULL ON ERROR);
```

Docs
-------
[Oracle JSON_EXISTS](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SQL-JSON-Conditions.html#GUID-57762C80-0C8A-4B18-9BA7-9B3F5ABDC988) | [MySQL JSON_VALUE](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#function_json-value)


sqlglot/dialects/oracle.py
```diff
@@ -226,6 +228,21 @@ def _parse_query_restrictions(self) -> t.Optional[exp.Expression]:
                 expression=self._match(TokenType.CONSTRAINT) and self._parse_field(),
             )
 
+        def _parse_json_exists(self) -> exp.JSONExists:
+            def _parse_passing_values() -> t.List[exp.Expression]:
+                # PASSING [<expr> AS <identifier>][, ...]
+                return self._parse_csv(lambda: self._parse_alias(self._parse_bitwise()))
```
georgesittas, line 234: Let's inline this.

georgesittas (approved):
This is a breaking change.

PR #4035, opened 2024-09-02
fix(duckdb): Wrap negative time intervals

VaggelisD:
DuckDB does not parse negative intervals properly, e.g:

```SQL
D SELECT CAST('2008-12-25' AS DATE) + INTERVAL -5 DAY;
Error: Parser Error: syntax error at or near "DAY"
                                                  ^
```

However, it can do so if the value is either (1) wrapped or (2) quoted:

```
D SELECT CAST('2008-12-25' AS DATE) + INTERVAL (-5) DAY;
┌────────────────────────────────────────────────────────────────────────────────────┐
│ (CAST('2008-12-25' AS DATE) + to_days(CAST(trunc(CAST(-5 AS DOUBLE)) AS INTEGER))) │
│                                     timestamp                                      │
├────────────────────────────────────────────────────────────────────────────────────┤
│ 2008-12-20 00:00:00                                                                │
└────────────────────────────────────────────────────────────────────────────────────┘

D SELECT CAST('2008-12-25' AS DATE) + INTERVAL '-5' DAY;
┌──────────────────────────────────────────────────────────────────────────────────────┐
│ (CAST('2008-12-25' AS DATE) + to_days(CAST(trunc(CAST('-5' AS DOUBLE)) AS INTEGER))) │
│                                      timestamp                                       │
├──────────────────────────────────────────────────────────────────────────────────────┤
│ 2008-12-20 00:00:00                                                                  │
└──────────────────────────────────────────────────────────────────────────────────────┘

```

This PR solves this problem by wrapping negative values at `exp.Interval` generation.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -793,6 +793,12 @@ class DuckDB(Dialect):
             if unit.startswith("quarter"):
                 multiplier = 90
 
+            # Wrap negative intervals as DuckDB does not parse them properly, e.g
+            # "SELECT date_col + INTERVAL -5 DAY" throws a parse error
+            this = expression.this
+            if this.is_number and this.to_py() < 0:
+                expression.set("this", exp.Paren(this=this))
+
             if multiplier:
                 return f"({multiplier} * {super().interval_sql(exp.Interval(this=expression.this, unit=exp.var('DAY')))})"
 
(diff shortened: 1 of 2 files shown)


VaggelisD:
After chatting about this approach with @georgesittas we concluded that it's not entirely correct; The negative value should be an `exp.Neg` node in which case the generator would wrap it automatically.

The deeper issue seems to be that that intervals are canonicalized in different ways across dialects, e.g. in some places the value is stored (or created) as a string literal, in others as a number literal etc.

PR #4036, opened 2024-09-02
Add `materialized` argument to `with_`

max-muoto:
Add `materialized` argument when building a CTE with `with_`. It's supported in both Postgres and DuckDB:

- [Postgres](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-CTE-MATERIALIZATION)
- [DuckDB](https://duckdb.org/docs/sql/query_syntax/with.html#materialized-ctes)

See Slack thread [here](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1725243448537369).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ install-dev-rs-release:
 	cd sqlglotrs/ && python -m maturin develop -r
 
 install-dev-rs:
+	@unset CONDA_PREFIX && \
 	cd sqlglotrs/ && python -m maturin develop
 
 install-dev-core:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ install-dev-rs-release:
 	cd sqlglotrs/ && python -m maturin develop -r
 
 install-dev-rs:
+	@unset CONDA_PREFIX && \
```
max-muoto, line 10: I had to add this, as I had a conda prefix set, it's not totally uncommon for makefiles to set this to make things easier. For example, Polars does this: https://github.com/pola-rs/polars/blob/main/Makefile

Let me know if we're fine with keeping it.

tobymao: as long as this doesn't break my setup

max-muoto: It shouldn't! haha

tobymao (approved):

PR #4049, opened 2024-09-03
feat(prql): Added from, limit, select, & where transforms.

JJHCool:
Added transform implementation for:
* from
* limit

Began transform implementation for:
* where
* select


georgesittas:
@JJHCool please refactor your code so that it works across for all python versions ≥ 3.7, and also run `make style`. This PR is still too big to review in one go, ideally you should have opened a PR for each of these clauses. We'll do a review pass once CI is green.

tobymao:
closing for now

Issue #3959, opened 2024-08-22
Oracle root select surrounded in () been transpiled to CH causes an error


sleshJdev:
**Fully reproducible code snippet**
```
from sqlglot.dialects import Oracle, ClickHouse
sqlglot.transpile("with core_sql as (select 1) (select 1)", read=Oracle, write=ClickHouse, comments=False)
['WITH core_sql AS (SELECT 1) (SELECT 1)']
```

Produces an error when running in CH:

```
Code: 62. DB::Exception: Syntax error: failed at position 29 ('('): (SELECT 1). Expected one of: token, Comma, FROM, SELECT. (SYNTAX_ERROR) (version 23.12.2.59 (official build)) , server ClickHouseNode
```

Works fine with `container-registry.oracle.com/database/free:latest`

Expected result: `WITH core_sql AS (SELECT 1) SELECT 1`. CH runs it successfully

sleshJdev:
Would be appreciated for a hint where to start from since I'm not quite familiar with design/architecture of sqlglot.

tobymao:
in the generator of clickhouse, probably need to remove the extra parenthesis, look at the ast and alter it

PR #4052, opened 2024-09-04
Issue 3959 - Oracle root select surrounded in () been transpiled to CH causes an error

sleshJdev:
https://github.com/tobymao/sqlglot/issues/3959

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -1052,6 +1052,20 @@ class ClickHouse(Dialect):
         def placeholder_sql(self, expression: exp.Placeholder) -> str:
             return f"{{{expression.name}: {self.sql(expression, 'kind')}}}"
 
+        def subquery_sql(
+            self, expression: exp.Subquery, sep: str = " AS ", wrap_left="(", wrap_right=")"
+        ) -> str:
+            alias = self.sql(expression, "alias")
+            if not alias and isinstance(expression.this, exp.Subquery):
+                return super().subquery_sql(expression, wrap_left="", wrap_right="")
+            if (
+                not alias
+                and isinstance(expression.this, exp.Select)
+                and expression.parent_select is None
+            ):
+                return super().subquery_sql(expression, wrap_left="", wrap_right="")
+            return super().subquery_sql(expression)
+
         def oncluster_sql(self, expression: exp.OnCluster) -> str:
             return f"ON CLUSTER {self.sql(expression, 'this')}"
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -741,7 +741,7 @@ class Generator(metaclass=_Generator):
 
         return f"{sql} {comments_sql}"
 
-    def wrap(self, expression: exp.Expression | str) -> str:
+    def wrap(self, expression: exp.Expression | str, left="(", right=")") -> str:
         this_sql = (
             self.sql(expression)
             if isinstance(expression, exp.UNWRAPPED_QUERIES)
@@ -751,7 +751,7 @@ class Generator(metaclass=_Generator):
             return "()"
 
         this_sql = self.indent(this_sql, level=1, pad=0)
-        return f"({self.sep('')}{this_sql}{self.seg(')', sep='')}"
+        return f"{left}{self.sep('')}{this_sql}{self.seg(right, sep='')}"
 
     def no_identify(self, func: t.Callable[..., str], *args, **kwargs) -> str:
         original = self.identify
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -1045,6 +1045,20 @@ def combinedparameterizedagg_sql(self, expression: exp.CombinedParameterizedAgg)
         def placeholder_sql(self, expression: exp.Placeholder) -> str:
             return f"{{{expression.name}: {self.sql(expression, 'kind')}}}"
 
+        def subquery_sql(
+            self, expression: exp.Subquery, sep: str = " AS ", wrap_left="(", wrap_right=")"
```
georgesittas, line 1049: Why do we need both `wrap_left` and `wrap_right`? I think a single `wrap` kwarg suffices here.

sqlglot/dialects/clickhouse.py
```diff
@@ -1045,6 +1045,20 @@ def combinedparameterizedagg_sql(self, expression: exp.CombinedParameterizedAgg)
...
+        ) -> str:
+            alias = self.sql(expression, "alias")
+            if not alias and isinstance(expression.this, exp.Subquery):
+                return super().subquery_sql(expression, wrap_left="", wrap_right="")
+            if (
+                not alias
+                and isinstance(expression.this, exp.Select)
+                and expression.parent_select is None
+            ):
+                return super().subquery_sql(expression, wrap_left="", wrap_right="")
```
georgesittas, line 1059: Can you elaborate on what's happening here? What is this trying to achieve exactly? On a side note, it doesn't seem like there's a good reason to have two branches, we can simplify by merging them.

sqlglot/generator.py
```diff
@@ -741,7 +741,7 @@ def maybe_comment(
 
         return f"{sql} {comments_sql}"
 
-    def wrap(self, expression: exp.Expression | str) -> str:
+    def wrap(self, expression: exp.Expression | str, left="(", right=")") -> str:
```
georgesittas, line 744: Let's not touch `wrap` for this PR, it shouldn't be relevant.

tests/test_transpile.py
```diff
@@ -3,6 +3,7 @@
 from unittest import mock
 
 from sqlglot import parse_one, transpile
+from sqlglot.dialects import ClickHouse, Oracle
```
georgesittas, line 6: No need to import dialects, just use the string version, i.e. `"clickhouse"`, and `"oracle"`.

tests/test_transpile.py
```diff
@@ -58,6 +59,18 @@ def test_paren(self):
...
+            "WITH core_sql AS (SELECT 1) SELECT 2",
+            read=Oracle,
+            write=ClickHouse,
+        )
+        self.validate(
+            "SELECT * FROM ((SELECT 1) AS EG)",
+            "SELECT * FROM (SELECT 1) AS EG",
+            read=Oracle,
+            write=ClickHouse,
+        )
```
georgesittas, line 73: Move these tests to `test_clickhouse.py`.

sqlglot/dialects/clickhouse.py
sleshJdev: We need to get rid of extra () during transpiling. By default wrap function add () even if there is no alias. By adding this kw args I allow caller control whether or not to add ()

georgesittas: You shouldn't have to specify each delimiter separately; my point was that if we went this way, it would be enough to just have a `wrap` kwarg that controls whether you'll use `self.wrap` in `subquery_sql`.

sqlglot/generator.py
sleshJdev: I couldn't came anything better. This function adds extra (), not sure yet how to fix it without touching wrap function. Maybe you could help?

sqlglot/dialects/clickhouse.py
sleshJdev: I already tried it and it didn't work, because wrap does extra work, not just adding (). When I implemented it this way it broke the tests

georgesittas:
Hey @sleshJdev, appreciate the PR but I think this needs some more work. I'll need to look into CH's syntax spec to understand how we should address this, because there are probably more cases where wrapped subqueries are invalid, so if you can fetch some docs / references we'll try to either get a fix in or provide more detailed guidance on how to solve this.

I'll close this for now, happy to continue chatting.

sleshJdev:
@georgesittas not sure about the docs. When the brackets open, clickhouse and others treat everything inside the brackets as a result of executing some select. So this is standard behavior of any SQL, especially when there are brackets after FROM.

EugeneTorap:
Hey @georgesittas @VaggelisD! Can you help to implement this logic? It's very important for us!

georgesittas:
Hi Eugene, our bandwidth is a bit limited these days so not sure when we'll be able to pick this up again. If the workaround @sleshJdev implemented in this PR is enough you could use it as a transformation in your application until we deal with this upstream.

PR #4053, opened 2024-09-04
feat(clickhouse): transpile oracle functions chr, lag, lead

sleshJdev:


sqlglot/dialects/clickhouse.py
```diff
@@ -915,6 +915,22 @@ class Generator(generator.Generator):
...
+                "lagInFrame", expression.this, expression.args["offset"], expression.args["default"]
+            )
+
+        def lead_sql(self, expression: exp.Lead):
+            return self.func(
+                "leadInFrame",
+                expression.this,
+                expression.args["offset"],
+                expression.args["default"],
+            )
```
georgesittas, line 932: Let's move these to `TRANSFORMS` instead of defining methods.

tests/test_transpile.py
```diff
@@ -21,6 +22,26 @@ class TestTranspile(unittest.TestCase):
     def validate(self, sql, target, **kwargs):
         self.assertEqual(transpile(sql, **kwargs)[0], target)
 
+    def test_functions(self):
```
georgesittas, line 25: Move these to `test_clickhouse.py`, get rid of dialect imports and use string versions.

georgesittas (commented):
Please share docs regarding these changes

tests/test_transpile.py
sleshJdev: But this test for transpiling. How to implement same test in there? Is it okay if I'll just use string-like dialect names?

sqlglot/dialects/clickhouse.py
sleshJdev: Done 665b453e0f112369f85b04ea7705c0646977e89d

georgesittas:
Thanks for the PR @sleshJdev, I'll get this in and do any final cleanups.

PR #4056, opened 2024-09-04
fix(bigquery)!: Fix transpilation of exp.AtTimeZone

### LINKED ISSUE #4038: Transpile Tino SQL to BigQuery SQL with `AT TIME ZONE` incorrectly.
## Descirption
I found an issue when converting Trino SQL to BigQuery SQL with `AT TIME ZONE`.

### Reproducible code
```python
import sqlglot

sql = sqlglot.transpile(
  """
  SELECT "PruchaseTimestamp" FROM "Orders"
  WHERE ("PruchaseTimestamp" = TIMESTAMP '2022-08-30 16:00:00.000' AT TIME ZONE 'America/Bahia_Banderas')
  """,
  read="trino", 
  write="bigquery"
)[0]
# SELECT `PruchaseTimestamp` FROM `Orders` 
# WHERE (`PruchaseTimestamp` = TIMESTAMP(DATETIME(CAST('2022-08-30 16:00:00.000' AS DATETIME), 'America/Bahia_Banderas')))
```

The `TIMESTAMP(DATETIME(CAST('2022-08-30 16:00:00.000' AS DATETIME), 'America/Bahia_Banderas'))` seem to be a wrong value.
It should be `TIMESTAMP('2022-08-30 16:00:00.000', 'America/Bahia_Banderas')`.

In BigQuery, if I fix the wrong type,
```sql
SELECT TIMESTAMP('2022-08-30 16:00:00.000', 'America/Bahia_Banderas'); -- 2022-08-30 21:00:00 UTC
SELECT TIMESTAMP(DATETIME(CAST('2022-08-30 16:00:00.000' AS TIMESTAMP), 'America/Bahia_Banderas')); -- 2022-08-30 11:00:00 UTC
```

VaggelisD:
Fixes #4038

Docs
--------
[BQ DATETIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime) | [BQ TIMESTAMP](https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#timestamp) | [Presto AT TIME ZONE](https://trino.io/docs/current/functions/datetime.html#time-zone-conversion)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -933,9 +933,17 @@ class BigQuery(Dialect):
             # BigQuery allows CAST(.. AS {STRING|TIMESTAMP} [FORMAT <fmt> [AT TIME ZONE <tz>]]).
             # Only the TIMESTAMP one should use the below conversion, when AT TIME ZONE is included.
             if not isinstance(parent, exp.Cast) or not parent.to.is_type("text"):
-                return self.func(
-                    "TIMESTAMP", self.func("DATETIME", expression.this, expression.args.get("zone"))
-                )
+                from sqlglot.optimizer.annotate_types import annotate_types
+
+                this = annotate_types(expression.this)
+                zone = expression.args.get("zone")
+
+                # BigQuery will transpile "<expr> AT TIME ZONE <zone>" to "TIMESTAMP(expr, zone)"
+                # except when "expr" is a TIMESTAMP in BQ, in which case it'll first get wrapped with DATETIME()
+                if this.is_type(exp.DataType.Type.TIMESTAMPTZ):
+                    return self.func("TIMESTAMP", self.func("DATETIME", this, zone))
+
+                return self.func("TIMESTAMP", this, zone)
 
             return super().attimezone_sql(expression)
 
--- a/tests/dialects/test_presto.py
+++ b/tests/dialects/test_presto.py
@@ -1034,6 +1034,14 @@ class TestPresto(Validator):
                 "spark": "SELECT REGEXP_EXTRACT(TO_JSON(FROM_JSON('[[1, 2, 3]]', SCHEMA_OF_JSON('[[1, 2, 3]]'))), '^.(.*).$', 1)",
             },
         )
+        self.validate_all(
+            "SELECT TIMESTAMP '2022-08-30 16:00:00.000' AT TIME ZONE 'America/Bahia_Banderas'",
+            write={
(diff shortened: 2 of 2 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -933,9 +933,17 @@ def attimezone_sql(self, expression: exp.AtTimeZone) -> str:
             # BigQuery allows CAST(.. AS {STRING|TIMESTAMP} [FORMAT <fmt> [AT TIME ZONE <tz>]]).
             # Only the TIMESTAMP one should use the below conversion, when AT TIME ZONE is included.
             if not isinstance(parent, exp.Cast) or not parent.to.is_type("text"):
-                return self.func(
-                    "TIMESTAMP", self.func("DATETIME", expression.this, expression.args.get("zone"))
-                )
+                from sqlglot.optimizer.annotate_types import annotate_types
+
+                this = annotate_types(expression.this)
```
georgesittas, line 938: I think we should gate this with `if not this.type`, otherwise we risk overwriting a previously-inferred type.

VaggelisD: Do we have prior examples of this happening? This looks off to me from an API perspective:
- If type inference is idempotent, this should have the same result so no need for a guard
- If it's not, then the overwrite might have better/newer results than the previous one (?)

I wasn't aware of this so didn't think about adding such a branch, however if it's needed I'd argue it should be baked into the type annotator.

georgesittas: This `annotate_types` call has limited context, for example you would infer `x`'s type to be `UNKNOWN`, but if you ran `annotate_types` separately (e.g. what we do in SQLMesh), where you have more context (i.e. schemas), you would know the type beforehand.

VaggelisD: That makes sense, however I think the guard should still be embedded in the API somehow (e.g. overwrite only if we go from `UNKNOWN` to a proper type), otherwise we'd have to add it across all generator annotations right?

georgesittas: We _do_ have it in most, if not all, generator annotations; we should definitely do it everywhere otherwise we could end up in a situation like the one I described above.

> however I think the guard should still be embedded in the API somehow

I don't think this is reasonable behavior. You don't want to make the type inference conditionally overwrite, you want it to always overwrite because otherwise you can't improve previously inferred types. For example if you inferred that an expression has type `INT` with limited context, but then later you have more context (e.g. new/updated schemas) and you can re-infer it to be `BIGINT`, you wanna overwrite the older one there.

georgesittas: If it bothers you, you could make a helper like `maybe_annotate_types(E) -> E` which hides the `if expr.type` check.

PR #4060, opened 2024-09-05

Toms1999:


georgesittas:
See https://github.com/tobymao/sqlglot/pull/3685#issuecomment-2182822498.

PR #4061, opened 2024-09-05

Toms1999:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="postgres")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="doris")`
- Check if the issue still exists on main:Couldn't find anything related

```python
import sqlglot

sql = """SELECT NOW() - INTERVAL '1 YEAR + 1 MONTH'"""

print(sqlglot.transpile(sql ,read="postgres" , write="doris")[0]);
``` 

output
```
SELECT NOW() - INTERVAL '1 YEAR + 1 MONTH'
```

Expected output
```
SELECT NOW() - INTERVAL '1' YEAR - INTERVAL '1' MONTH

```

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6946,6 +6946,7 @@ def parse_identifier(name: str | Identifier, dialect: DialectType = None) -> Ide
 
 
 INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
+INTERVAL_STRING_RE_three = re.compile(r"([+-]?)\s*([0-9]+)\s*([a-zA-Z]+)")
 
 
 def to_interval(interval: str | Literal) -> Interval:
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4343,14 +4343,32 @@ class Parser(metaclass=_Parser):
         if this and this.is_number:
             this = exp.Literal.string(this.to_py())
         elif this and this.is_string:
-            parts = exp.INTERVAL_STRING_RE.findall(this.name)
+            parts = exp.INTERVAL_STRING_RE_three.findall(this.name)
             if len(parts) == 1:
                 if unit:
                     # Unconsume the eagerly-parsed unit, since the real unit was part of the string
                     self._retreat(self._index - 1)
 
-                this = exp.Literal.string(parts[0][0])
-                unit = self.expression(exp.Var, this=parts[0][1].upper())
+                this = exp.Literal.string(parts[0][1])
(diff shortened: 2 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -6946,6 +6946,7 @@ def parse_identifier(name: str | Identifier, dialect: DialectType = None) -> Ide
 
 
 INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
+INTERVAL_STRING_RE_three = re.compile(r"([+-]?)\s*([0-9]+)\s*([a-zA-Z]+)")
```
georgesittas, line 6949: What does "three" mean? Why is it missing the `\s*` terms?

sqlglot/parser.py
```diff
@@ -4343,14 +4343,32 @@ def _parse_interval(self, match_interval: bool = True) -> t.Optional[exp.Add | e
...
+                    current_interval = self.expression(exp.Interval, this=interval_value, unit=interval_unit)
+
+                    if interval_expr is None:
+                        interval_expr = current_interval
+                    else:
+                        if sign == '+':
+                            interval_expr = self.expression(exp.Sub, this=interval_expr, expression=current_interval)
+                        else:
+                            interval_expr = self.expression(exp.Add, this=interval_expr, expression=current_interval)
+
```
georgesittas, line 4370: Why is `+` being treated as `Sub` and `-` as `Add`? This doesn't look right, e.g. try it in this example:

```
mysql> SELECT NOW() + INTERVAL '1' YEAR - INTERVAL '1' MONTH;
+------------------------------------------------+
| NOW() + INTERVAL '1' YEAR - INTERVAL '1' MONTH |
+------------------------------------------------+
| 2025-08-05 12:57:35                            |
+------------------------------------------------+
1 row in set (0,00 sec)
```

georgesittas (commented):
The approach here doesn't look correct, I'll close the PR for now.

sqlglot/parser.py
Toms1999: SELECT NOW() - INTERVAL '1 YEAR + 1 MONTH' 
be equivalent to
SELECT NOW() - INTERVAL '1' YEAR - '1' MONTH
------------------------------
SELECT NOW() - INTERVAL '1 YEAR - 1 MONTH' 
be equivalent to
SELECT NOW() - INTERVAL '1' YEAR + '1' MONTH

sqlglot/expressions.py
Toms1999: i will do it

PR #4062, opened 2024-09-05
fix(starrocks): Move the parse distribute and duplicate to Parser

hellozepp:
link: #3997 

Fixed an issue that could cause the `_parse_distributed_property` and `_parse_duplicate` functions in lambda to not be found


georgesittas (approved):
Good catch, thanks for fixing it. 👍

PR #4064, opened 2024-09-05

Toms1999:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="postgres")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="doris")`
- Check if the issue still exists on main:Couldn't find anything related

```python
import sqlglot

sql = """SELECT NOW() - INTERVAL '1 YEAR + 1 MONTH'"""

print(sqlglot.transpile(sql ,read="postgres" , write="doris")[0]);
``` 

output
```
SELECT NOW() - INTERVAL '1 YEAR + 1 MONTH'
```

Expected output
```
SELECT NOW() - INTERVAL '1' YEAR - INTERVAL '1' MONTH

```

```python
import sqlglot

sql = """SELECT NOW() - INTERVAL '1 YEAR - 1 MONTH'"""

print(sqlglot.transpile(sql ,read="postgres" , write="doris")[0]);
``` 

output
```
SELECT NOW() - INTERVAL '1 YEAR - 1 MONTH'
```

Expected output
```
SELECT NOW() - INTERVAL '1' YEAR + INTERVAL '1' MONTH

```

image

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6946,7 +6946,7 @@ def parse_identifier(name: str | Identifier, dialect: DialectType = None) -> Ide
 
 
 INTERVAL_STRING_RE = re.compile(r"\s*([0-9]+)\s*([a-zA-Z]+)\s*")
-INTERVAL_STRING_RE_three = re.compile(r"([+-]?)\s*([0-9]+)\s*([a-zA-Z]+)")
+INTERVAL_STRING_RE_triple = re.compile(r"\s*([+-]?)\s*([0-9]+)\s*([a-zA-Z]+)\s*")
 
 
 def to_interval(interval: str | Literal) -> Interval:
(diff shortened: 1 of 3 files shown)


georgesittas:
@Toms1999 please don't spam, I already explained that this is not looking right in the other one. We're not going to accept PRs if you're not going to communicate with us. Also, please avoid opening PRs that don't pass CI checks.

Toms1999:
@georgesittas i am sorry , but Can you elaborate on what the problem is please

Toms1999:
@georgesittas Also sorry I only checked the style, next time I will make sure make test

georgesittas:
> @georgesittas i am sorry , but Can you elaborate on what the problem is please

This is being parsed incorrectly, I shared a similar example in the other PR too:

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT NOW() + INTERVAL '1 YEAR - 1 MONTH'", "postgres", "doris")
["SELECT NOW() + INTERVAL '1' YEAR + INTERVAL '1' MONTH"]
```

You're converting a `+` into a `Sub`, assuming that there is a `-` operator preceding the `INTERVAL`, which isn't necessarily the case. You need to preserve the operations and wrap the whole thing in parentheses for this transformation to work.

> sorry I only checked the style, next time I will make sure make test

Make sure to commit the style changes made by `make style`, otherwise CI will fail.

PR #4067, opened 2024-09-05
fix(tsql): use plus operator for string concat to support more systems that use tsql

### LINKED ISSUE #4066: double pipe operator is not string concatenation in ms sql server
Double pipe (`||`) [only applies to Azure SQL database](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/string-concatenation-transact-sql?view=sql-server-ver16), while `+` [applies to Azure SQL database as well as a bunch of others](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/string-concatenation-transact-sql?view=sql-server-ver16).

cpcloud:
Closes #4066.

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -420,6 +420,8 @@ class TestTSQL(Validator):
             "SELECT val FROM (VALUES ((TRUE), (FALSE), (NULL))) AS t(val)",
             write_sql="SELECT val FROM (VALUES ((1), (0), (NULL))) AS t(val)",
         )
+        self.validate_identity("'a' + 'b'")
+        self.validate_all("'a' || 'b'", write={"tsql": "'a' + 'b'"})
 
     def test_option(self):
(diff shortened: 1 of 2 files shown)


tests/dialects/test_tsql.py
```diff
@@ -420,6 +420,8 @@ def test_tsql(self):
             "SELECT val FROM (VALUES ((TRUE), (FALSE), (NULL))) AS t(val)",
             write_sql="SELECT val FROM (VALUES ((1), (0), (NULL))) AS t(val)",
         )
+        self.validate_identity("'a' + 'b'")
+        self.validate_all("'a' || 'b'", write={"tsql": "'a' + 'b'"})
```
georgesittas, line 424:
```suggestion
        self.validate_identity(
            "'a' || 'b'",
            "'a' + 'b'",
        )
```

georgesittas (approved):
Thanks!

Issue #4047, opened 2024-09-03
Include license for the sqlglotrs package


chriscc2:
Hi, sorry if this is the wrong project/component/space to ask for help. I am using the sqlglotrs package found at https://pypi.org/project/sqlglotrs/, which I think is derived from this project. When I am trying to get the licensing details from pip-license, it appears to be UNKNOWN. The pypi page also does not have any licensing information. I am wondering if it is possible to include the licensing detail in the final package and in the pypi repo.

Thanks in advance.

_Originally posted by @chriscc2 in https://github.com/tobymao/sqlglot/discussions/4010_

dbjnbnrj:
Hi @georgesittas, is this something you can help us with? This is blocking our adoption of the library.

georgesittas:
Hey, yes I'll be looking into this shortly, thanks for bringing it into our attention.

dbjnbnrj:
Thank you 👯

vinitachaudhary:
@georgesittas Thank you for fixing this quickly. Would it be possible to publish to pypi also so we can get the licensing info? Please share the timeline by when the license can be available in pypi for us. We have a dependency on this package and the license info is a requirement for our launch which is due by 9/13.

tobymao:
we deploy frequently

vinitachaudhary:
@tobymao @georgesittas Thanks for your responsiveness. When is the next deployment planned?

tobymao:
we don't have a strict schedule

georgesittas:
https://pypi.org/project/sqlglotrs/

vinitachaudhary:
@tobymao @georgesittas Great, thanks for your quick responses and help!

dbjnbnrj:
Hi @georgesittas, Is it possible to add the github project to the pypi page as well? 
![image](https://github.com/user-attachments/assets/48169862-8618-4e0e-920d-25aad6c65021)

PR #4071, opened 2024-09-05
Chore: fill in more details for sqlglotrs pypi page

georgesittas:
Addresses https://github.com/tobymao/sqlglot/issues/4047#issuecomment-2332393032

Reference: https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata


dbjnbnrj:
Thanks for making this update!

PR #4073, opened 2024-09-06
Fix(athena): Apply correct quoting to queries depending on type (DML or DDL)

erindru:
Athena [quotes identifiers differently depending on query type](https://docs.aws.amazon.com/athena/latest/ug/reserved-words.html), presumably because it uses different execution engines for DDL vs DML.

DDL queries (except `CREATE VIEW`) use backticks `` ` `` to quote identifiers and everything else uses double quotes `"` to quote identifiers. 

Here some examples
```
# backtick quoting
create schema if not exists "foo"; #fails
create schema if not exists `foo`; #works

create external table "foo" ("id" int) location 's3://foo/'; #fails
create external table `foo` (`id` int) location 's3://foo/'; #works

drop table if exists "foo"; #fails
drop table if exists `foo`; #works

# double-quote quoting
create or replace view `foo` as select 1 as `col`; #fails
create or replace view "foo" as select 1 as "col"; #works

select 1 as `foo`; #fails
select 1 as "foo"; #works
```

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -44,22 +44,16 @@ class Athena(Trino):
                 # ref: https://docs.aws.amazon.com/athena/latest/ug/reserved-words.html
                 # ref: https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html#table-names-that-include-numbers
                 if not (isinstance(expression, exp.Create) and expression.kind == "VIEW"):
+                    self._identifier_start = "`"
+                    self._identifier_end = "`"
 
-                    def _transformer(node):
-                        if isinstance(node, exp.Identifier):
-                            node.meta["quote_ddl"] = True
-                        return node
-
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/athena.py
```diff
@@ -36,7 +37,29 @@ class Generator(Trino.Generator):
             exp.FileFormatProperty: lambda self, e: f"'FORMAT'={self.sql(e, 'this')}",
         }
 
+        def generate(self, expression: exp.Expression, copy: bool = True) -> str:
```
tobymao, line 40: why didn't you use standard transform in TRANSFORMS ?

erindru: ...standard transform? I didn't see one that could intercept the entire expression before any of the other transforms.

The quoting to use is determined by the root node, so attempting to figure it out at the `exp.Identifier` level seemed like more work than simply figuring it out at the start and storing it as metadata

tobymao (approved):

PR #4079, opened 2024-09-06
fix(clickhouse): Remove CURRENT_TIMESTAMP from NO_PAREN_FUNCTIONS

### LINKED ISSUE #4076: Clickhouse current_timestamp table field is not support
**Before you file an issue**
`parse_one(sql, read="clickhouse")`

In clickhouse, current_timestamp can be used as a field name, and this is not a SQL expression, but is converted to exp.CurrentTimestamp during parse.

Clickhouse sql:
```
select current_timestamp, a from t
```
Parsed AST as following:
```
Select(
  expressions=[
    CurrentTimestamp(),
    Column(
      this=Identifier(this=a, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=t, quoted=False))))
```
I excepted output is `current_timestamp` Column.

Execute in clickhouse:
```
INTO my_db.my_table2 (event_date, event_time, current_timestamp, user_id, clicks)
VALUES ('2023-12-31', '2023-12-31 23:59:59', 1725616170, 1, 10);

INSERT INTO my_db.my_table2 (event_date, event_time, current_timestamp, user_id, clicks) FORMAT Values

Query id: 0e2beb83-6acd-447e-a176-31816c982142

Ok.

1 row in set. Elapsed: 0.026 sec. 
```

select current_timestamp from my_db.my_table2;
```
0c40e341ddf3 :) select current_timestamp from my_db.my_table2;

SELECT current_timestamp
FROM my_db.my_table2

Query id: 871b946d-e790-4858-b0da-d4b63c359d88

   ┌─current_timestamp─┐
1. │        1725616170 │ -- 1.73 billion
   └───────────────────┘
```

SELECT current_timestamp
```
SELECT current_timestamp

Query id: abfc55fd-6313-4c1c-a80f-c636bbdeb027


Elapsed: 0.013 sec. 

Received exception from server (version 24.8.3):
Code: 47. DB::Exception: Received from localhost:9000. DB::Exception: Unknown expression identifier 'current_timestamp' in scope SELECT current_timestamp. (UNKNOWN_IDENTIFIER)

```

VaggelisD:
Fixes #4076

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -403,6 +403,9 @@ class ClickHouse(Dialect):
         NO_PAREN_FUNCTION_PARSERS = parser.Parser.NO_PAREN_FUNCTION_PARSERS.copy()
         NO_PAREN_FUNCTION_PARSERS.pop("ANY")
 
+        NO_PAREN_FUNCTIONS = parser.Parser.NO_PAREN_FUNCTIONS.copy()
+        NO_PAREN_FUNCTIONS.pop(TokenType.CURRENT_TIMESTAMP)
+
         RANGE_PARSERS = {
             **parser.Parser.RANGE_PARSERS,
             TokenType.GLOBAL: lambda self, this: self._match(TokenType.IN)
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -515,6 +515,7 @@ class TestClickhouse(Validator):
         )
         self.validate_identity("SELECT TRIM(TRAILING ')' FROM '(   Hello, world!   )')")
         self.validate_identity("SELECT TRIM(LEADING '(' FROM '(   Hello, world!   )')")
+        self.validate_identity("current_timestamp").assert_is(exp.Column)
 
     def test_clickhouse_values(self):
         values = exp.select("*").from_(


georgesittas (approved):

hellozepp:
@VaggelisD Can we add quoted arg to CURRENT_TIMESTAMP in ClickHouse?

e.g: this=Identifier(this=current_timestamp, quoted=True))

This will ensure that the target dialect will not have ambiguity when transpiling whether it is a field or a sql expression?

I encountered this problem when transpiling to spark, which supports this sql expression. Although it is not easy to handle, I did it like this:
```
def _escape_current_timestamp(expression: exp.Select):
    for column in expression.find_all(exp.Column):
        if column.this.name.upper() == 'CURRENT_TIMESTAMP':
            column.this.set("quoted", True)
```
Is this right?

georgesittas:
This isn't the correct way to do it, you need to add `RESERVED_KEYWORDS` for spark or whatever you want to transpile towards.

georgesittas:
If you're parsing using clickhouse that's not right, for example:

```python
>>> import sqlglot
>>> sqlglot.parse_one("current_timestamp", "clickhouse")
Column(
  this=Identifier(this=current_timestamp, quoted=False))
```

This would be quoted if `current_timestamp` is among the reserved expressions.

PR #4084, opened 2024-09-07
Feat: add merge expression builder

max-muoto:
I have a use-case around building merge statements similar to variety of other types of expressions that sqlglot can programmatically build. This PR adds a simple `merge` function allowing for dynamic construction of merge statements.


georgesittas (approved):
Thank you!

PR #4088, opened 2024-09-09
fix(transpile): Starrocks filter transpile

gauravsagar483:
filter to array_filter should be converted as SR supports same.

Reference: https://docs.starrocks.io/docs/sql-reference/sql-functions/array-functions/array_filter/

### DIFF
--- a/tests/dialects/test_dialect.py
+++ b/tests/dialects/test_dialect.py
@@ -1788,6 +1788,13 @@ class TestDialect(Validator):
             },
         )
         self.validate_all(
+            "FILTER(the_array, x -> x > 0)",
+            write={
+                "presto": "FILTER(the_array, x -> x > 0)",
+                "starrocks": "ARRAY_FILTER(the_array, x -> x > 0)",
+            },
+        )
+        self.validate_all(
             "a / b",
             write={
                 "bigquery": "a / b",
--- a/sqlglot/dialects/starrocks.py
+++ b/sqlglot/dialects/starrocks.py
@@ -106,6 +106,7 @@ class StarRocks(MySQL):
             exp.TimeStrToDate: rename_func("TO_DATE"),
             exp.UnixToStr: lambda self, e: self.func("FROM_UNIXTIME", e.this, self.format_time(e)),
             exp.UnixToTime: rename_func("FROM_UNIXTIME"),
+            exp.ArrayFilter: rename_func("ARRAY_FILTER"),
         }
 
         TRANSFORMS.pop(exp.DateTrunc)


tests/dialects/test_dialect.py
```diff
@@ -1787,6 +1787,13 @@ def test_operators(self):
                 "spark": "FILTER(the_array, x -> x > 0)",
             },
         )
+        self.validate_all(
+            "FILTER(the_array, x -> x > 0)",
+            write={
+                "presto": "FILTER(the_array, x -> x > 0)",
+                "starrocks": "ARRAY_FILTER(the_array, x -> x > 0)",
```
georgesittas, line 1794: This doesn't look correct, right? Starrocks expects the lambda to go first, based on the docs you shared:

```
  -- Return the elements in x that are less than the elements in y.
  select array_filter((x,y) -> x < y, [1,2,null], [4,5,6]);
  +--------------------------------------------------------+
  | array_filter((x, y) -> x < y, [1, 2, NULL], [4, 5, 6]) |
  +--------------------------------------------------------+
  | [1,2]                                                  |
  +--------------------------------------------------------+
```

gauravsagar483: Thanks @georgesittas, let me cross-check this.

georgesittas: Nvm, this argument order seems to work too

```
StarRocks > select array_filter(x -> x > 0, [1,-2,3,4]);
+-----------------------------------------+
| array_filter(x -> x > 0, [1, -2, 3, 4]) |
+-----------------------------------------+
| [1,3,4]                                 |
+-----------------------------------------+
1 row in set (0.03 sec)

StarRocks > select array_filter([1,-2,3,4], x -> x > 0);
+-----------------------------------------+
| array_filter([1, -2, 3, 4], x -> x > 0) |
+-----------------------------------------+
| [1,3,4]                                 |
+-----------------------------------------+
1 row in set (0.01 sec)
```

Presumably because of this remark in the The lambda function in array_filter(lambda_function, arr1,arr2...) follows the usage notes in [array_map()](https://docs.starrocks.io/docs/sql-reference/sql-functions/array-functions/array_map/):

> The lambda function in array_filter(lambda_function, arr1,arr2...) follows the usage notes in [array_map()](https://docs.starrocks.io/docs/sql-reference/sql-functions/array-functions/array_map/).

georgesittas: The only issue is that parsing `ARRAY_FILTER` with multiple arrays fails right now, and also Starrocks parsing produces an AST with the args in reverse order, so transpilation to Presto will not work for example. These can be addressed later, just pointing out.

gauravsagar483: SR supports `array_filter` in two [ways](https://docs.starrocks.io/docs/sql-reference/sql-functions/array-functions/array_filter/#syntax):
1. Array as first arg and boolean array or method(which returns boolean array) as second arg.
2. Using the lambda function as first arg, and other remaining arrays to match with lambda.

This PR targets the first approach only, As transpiling #2 (lambda) won't be possible to trino/presto has no alternatives for those.

Slack [thread](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1725867117599299):

gauravsagar483: I agree to the this [message](https://github.com/tobymao/sqlglot/pull/4088#discussion_r1750129997) above

georgesittas: Sounds good.

PR #4094, opened 2024-09-09
Refactor!: treat Nullable as an arg instead of a DataType.TYPE

georgesittas:
Motivation behind this PR is that nullable types are awkward to deal with: to access the wrapped type, one has to always do `data_type.this` instead of just `data_type`, meaning that the concept of nullability "leaks" into third-party code that may not care at all about it.

This refactor encodes nullability as an arg in `DataType`, so that it's less intrusive. Now, it will be taken into account when e.g. comparing data types (which can be optionally turned off by using `is_type`), and when roundtripping CH code.


tobymao (approved):

pkit:
@georgesittas it looks like `LowCardinality` needs the same treatment.

btw, our type comparison is broken because of this change, but I see why it's a better solution.

PR #4097, opened 2024-09-10
fix(prql): Rewrote tests to use `validate_all()` to separate the parser and generator tests.

JJHCool:
This PR rewrites the PRQL tests to use `validate_all()`, instead of `validate_identity()`, which will allow separation between the parser and the generator tests.

### DIFF
--- a/tests/dialects/test_prql.py
+++ b/tests/dialects/test_prql.py
@@ -5,77 +5,183 @@ class TestPRQL(Validator):
     dialect = "prql"
 
     def test_prql(self):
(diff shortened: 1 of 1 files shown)


georgesittas:
We usually do `"": ...` instead of `None: ...`. I'll get this to the finish line, just keep in mind for future PRs.

georgesittas:
@JJHCool I see that several tests were removed, why did this happen?

georgesittas:
I reverted this, in the future please don't remove tests or explain why they were removed. It's hard to notice using just git's diff.

PR #4099, opened 2024-09-10
Feat(athena): Improve DDL query support

erindru:
Athena is really a split between two different engines. One handles _most_ DDL (Hive) and the other handles the remaining DDL and all of the DML (Trino).

Since the Athena dialect already extends the Trino dialect, this PR:
 - Adds Hive tokens to the tokenizer
 - Tries to figure out (at generate time) if an Expression should be handled by Hive
 - If it should, the Hive Generator is used
 - Otherwise, the Athena Generator is used
 
 Note that I was unable to find an example of a query that the Athena _parser_ couldnt handle (as long as the Hive tokens were available at the tokenization stage) so only the Generator step contains the branching logic.
 
This approach maximizes code re-use since we already know how to deal with 99% of this syntax, just not within the same dialect

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -5,19 +5,7 @@ import typing as t
 from sqlglot import exp
 from sqlglot.dialects.trino import Trino
 from sqlglot.dialects.hive import Hive
-from sqlglot.tokens import Token, TokenType
-
-
-def _parse_as_hive(raw_tokens: t.List[Token]) -> bool:
-    if len(raw_tokens) > 0:
-        first_token = raw_tokens[0]
-        if first_token.token_type == TokenType.CREATE:
-            # CREATE is Hive (except for CREATE VIEW and CREATE TABLE... AS SELECT)
-            return not any(t.token_type in (TokenType.VIEW, TokenType.SELECT) for t in raw_tokens)
-
-        # ALTER and DROP are Hive
-        return first_token.token_type in (TokenType.ALTER, TokenType.DROP)
-    return False
+from sqlglot.tokens import TokenType
 
 
 def _generate_as_hive(expression: exp.Expression) -> bool:
@@ -81,13 +69,6 @@ class Athena(Trino):
             "UNLOAD": TokenType.COMMAND,
         }
 
-    class HiveParser(Hive.Parser):
-        """
-        Parse queries for the Athena Hive execution engine
-        """
-
-        pass
-
     class Parser(Trino.Parser):
         """
         Parse queries for the Athena Trino execution engine
@@ -97,30 +78,6 @@ class Athena(Trino):
             **Trino.Parser.STATEMENT_PARSERS,
             TokenType.USING: lambda self: self._parse_as_command(self._prev),
         }
-
-        def __init__(self, *args, **kwargs):
-            super().__init__(*args, **kwargs)
-
-            self._hive_parser = Athena.HiveParser(*args, **kwargs)
-
-        def parse(
-            self, raw_tokens: t.List[Token], sql: t.Optional[str] = None
-        ) -> t.List[t.Optional[exp.Expression]]:
-            if _parse_as_hive(raw_tokens):
-                return self._hive_parser.parse(raw_tokens, sql)
-
-            return super().parse(raw_tokens, sql)
-
-    class HiveGenerator(Hive.Generator):
-        """
-        Generating queries for the Athena Hive execution engine
-        """
-
-        def __init__(self, *args, **kwargs):
-            super().__init__(*args, **kwargs)
-
-            self._identifier_start = "`"
-            self._identifier_end = "`"
 
     class Generator(Trino.Generator):
         """
@@ -139,7 +96,10 @@ class Athena(Trino):
 
         def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
-            self._hive_generator = Athena.HiveGenerator(*args, **kwargs)
+
+            hive_kwargs = {**kwargs, "dialect": "hive"}
+
+            self._hive_generator = Hive.Generator(*args, **hive_kwargs)
 
         def generate(self, expression: exp.Expression, copy: bool = True) -> str:
             if _generate_as_hive(expression):
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -23,10 +23,9 @@ def _parse_as_hive(raw_tokens: t.List[Token]) -> bool:
 def _generate_as_hive(expression: exp.Expression) -> bool:
     if isinstance(expression, exp.Create):
         if expression.kind == "TABLE":
-            properties: t.Optional[exp.Properties]
-            if properties := expression.args.get("properties"):
-                if properties.find(exp.ExternalProperty):
-                    return True  # CREATE EXTERNAL TABLE is Hive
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/athena.py
```diff
@@ -1,31 +1,131 @@
...
 from sqlglot.dialects.trino import Trino
-from sqlglot.tokens import TokenType
+from sqlglot.dialects.hive import Hive
+from sqlglot.tokens import Token, TokenType
+
+
+def _parse_as_hive(raw_tokens: t.List[Token]) -> bool:
+    if len(raw_tokens) > 0:
+        first_token = raw_tokens[0]
+        if first_token.token_type == TokenType.CREATE:
```
georgesittas, line 14: I'm not sure if this approach is robust, for example what if you have leading CTEs? The first token would then be `WITH`. You could also have redundant semi-colons, etc.

sqlglot/dialects/athena.py
```diff
@@ -1,31 +1,131 @@
...
+
     class Tokenizer(Trino.Tokenizer):
+        """
+        The Tokenizer is flexible enough to tokenize queries across both the Hive and Trino engines
+        """
+
         IDENTIFIERS = ['"', "`"]
         KEYWORDS = {
+            **Hive.Tokenizer.KEYWORDS,
             **Trino.Tokenizer.KEYWORDS,
```
georgesittas, line 80: FYI there's one keyword that maps to a different token between Hive, Trino - `"ROW"`:

```python
>>> from sqlglot.dialects import Hive, Trino
>>> {k: (v, Trino.Tokenizer.KEYWORDS[k]) for k, v in Hive.Tokenizer.KEYWORDS.items() if k in Trino.Tokenizer.KEYWORDS and Trino.Tokenizer.KEYWORDS[k] != Hive.Tokenizer.KEYWORDS[k]}
{'ROW': (<TokenType.ROW: 'ROW'>, <TokenType.STRUCT: 'STRUCT'>)}
>>> {k: (v, Hive.Tokenizer.KEYWORDS[k]) for k, v in Trino.Tokenizer.KEYWORDS.items() if k in Hive.Tokenizer.KEYWORDS and Hive.Tokenizer.KEYWORDS[k] != Trino.Tokenizer.KEYWORDS[k]}
{'ROW': (<TokenType.STRUCT: 'STRUCT'>, <TokenType.ROW: 'ROW'>)}
```

This combination of dicts means we'll always use Trino's override. Not sure if this is fine, just pointing it out to double check for possible unintended side-effects.

sqlglot/dialects/athena.py
```diff
@@ -1,31 +1,131 @@
...
             "UNLOAD": TokenType.COMMAND,
         }
 
+    class HiveParser(Hive.Parser):
+        """
+        Parse queries for the Athena Hive execution engine
+        """
+
+        pass
```
georgesittas, line 89: This looks unnecessary, since we're not overriding anything. I'd personally just instantiate a `Hive.Parser` inline.

sqlglot/dialects/athena.py
```diff
@@ -1,31 +1,131 @@
...
+    class HiveGenerator(Hive.Generator):
+        """
+        Generating queries for the Athena Hive execution engine
+        """
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+
+            self._identifier_start = "`"
+            self._identifier_end = "`"
```
georgesittas, line 123: Is this necessary? If I'm not mistaken, `Hive.Generator` should already initialize these attributes correctly. If we can get rid of it, then ditto about instantiating inline.

sqlglot/dialects/athena.py
erindru: Yeah I was anticipating needing to tweak this. I forgot about CTE's, i'll add support for that.

How would you deal with redundant semicolons? Would you first filter the token stream to just the ones we care about and then do the checks?

sqlglot/dialects/athena.py
erindru: I think this particular instance is ok because unless i'm mistaken, `ROW` only appears in SELECT queries and Athena uses the Trino syntax for SELECT queries. It's why the Hive tokenizer keywords are listed first and then overridden by the Trino ones.

The higher problem is that I couldn't see an obvious way to determine which tokenizer to use pre-tokenization, however I also couldn't think of a query to add to the tests that exposed this as a problem

sqlglot/dialects/athena.py
erindru: I'm anticipating extending this in future PR's. But I guess I can remove it for now and re-instate it when needed

sqlglot/dialects/athena.py
erindru: You'd think so, however I noticed:

```
>>> q = parse_one("select * from foo")
>>> gen = Hive.Generator(identify=True)
>>> gen.generate(q)
'SELECT * FROM "foo"'
```

But you just triggered me to dig deeper into why this is and the following works as expected:
```
>>> gen = Hive.Generator(identify=True, dialect="hive")
>>> gen.generate(q)
'SELECT * FROM `foo`'
```

So i'll tidy this up, thanks!

sqlglot/dialects/athena.py
erindru: So if you have leading CTE's, the existing logic works and returns `False`. It doesnt need to check for them because only a `SELECT` query would have leading CTE's. Returning `False` triggers the use of the Trino tokenizer.

If you have a `CREATE TABLE .. AS WITH (...) SELECT` query, a `SELECT` still appears in the tokens so this is still correctly detected as a CTAS and returns `False` which triggers the use of the Trino tokenizer

erindru: How do you test redundant semicolons?

Trying to parse something like `; CREATE SCHEMA FOO;` just throws a parse error

georgesittas: > So if you have leading CTE's, the existing logic works and returns False. It doesnt need to check for them because only a SELECT query would have leading CTE's. All SELECT queries should use the Trino tokenizer, so returning False triggers the use of the Trino tokenizer.

Ah, interesting, I didn't realize that - that simplifies the problem then.

> How do you test redundant semicolons?

Using `parse`:

```python
>>> import sqlglot
>>> sqlglot.parse("; create schema foo ;;")
[None, Create(
  this=Table(
    db=Identifier(this=foo, quoted=False)),
  kind=SCHEMA), None]
```

Tbh this is an edge case but good to handle anyway, you can skip the leading tokens until you find the first non-semicolon or something.

sqlglot/dialects/athena.py
georgesittas: Ok sounds good, can always revisit later if needed.

georgesittas (approved):
LGTM - happy to merge and address semicolons later if we wanna get this in ASAP. Should be a low lift though.

sqlglot/dialects/athena.py
erindru: I struggled to write a test for this and eventually realized that the issues I had that I _thought_ were parsing issues were really tokenization issues. It appears the Trino parser is capable of handling all the Hive DDL as long as it is tokenized correctly.

So I was able to remove the delegation to the Hive parser entirely and just worry about the generation side. I may need to revisit this in future but I wasn't able to find a query to add to the tests that uses Hive syntax and also causes the Trino parser to fail

PR #4104, opened 2024-09-10
fix(snowflake): Manually escape single quotes in colon operator

### LINKED ISSUE #4090: Using singe quote as part of JSON object key leads to invalid SQL generation
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")` - ok
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")` - ok
- Check if the issue still exists on main - checked on sqlglot-25.20.2.dev5

**Overview**

When querying inside JSON object in Snowflake and referenced object's key contains `'` character the statements seems to be parsed correctly but when generating SQL the output is not valid (at least for snowflake and postgres dialects). Replacing single `'` in key name with escaped one: `''` resolves at least this case.

Object key can be referenced using `:` notation (e.g. `json_doc:"Customer's department"`) or using `get_path` function (`GET_PATH(PARSE_JSON(doc), '"Customer''s department"')`) - the result is the same.

```python
import sqlglot
sql_text = """select parse_json(doc):"Customer's department"::string as "dept" from json_tbl"""
print(sqlglot.parse_one(sql_text, dialect="snowflake").sql(dialect="snowflake"))
# SELECT CAST(GET_PATH(PARSE_JSON(doc), '["Customer's department"]') AS TEXT) AS "dept" FROM json_tbl

print(sqlglot.parse_one(sql_text, dialect="snowflake").sql(dialect="postgres"))
# SELECT CAST(JSON_EXTRACT_PATH(CAST(doc AS JSON), 'Customer's department') AS TEXT) AS "dept" FROM json_tbl
```
Expecting output for Snowflake to be 
```sql
SELECT CAST(GET_PATH(PARSE_JSON(doc), '["Customer''s department"]') AS TEXT) AS "dept" FROM json_tbl;
-- or
SELECT CAST(GET_PATH(PARSE_JSON(doc), '"Customer''s department"') AS TEXT) AS "dept" FROM json_tbl;
-- or, if not using GET_PATH function the original name without escaped quote can be used
SELECT CAST(PARSE_JSON(doc):"Customer's department" AS TEXT) AS "dept" FROM json_tbl;
```
And for PostgreSQL
```sql
SELECT CAST(JSON_EXTRACT_PATH(CAST(doc AS JSON), 'Customer''s department') AS TEXT) AS "dept" FROM json_tbl;
```

To test query with postgres:
```sql
create table json_tbl(doc jsonb);
insert into json_tbl values('{"Customer''s department": "Department A"}');

-- invalid query:
SELECT CAST(JSON_EXTRACT_PATH(CAST(doc AS JSON), 'Customer's department') AS TEXT) AS "dept" FROM json_tbl;
/*
ERROR:  syntax error at or near "s"
LINE 1: ...ST(JSON_EXTRACT_PATH(CAST(doc AS JSON), 'Customer's departme...
                                                             ^ 

SQL state: 42601
Character: 60
*/
-- valid query:
SELECT CAST(JSON_EXTRACT_PATH(CAST(doc AS JSON), 'Customer''s department') AS TEXT) AS "dept" FROM json_tbl;
```

**Official Documentation**
- [Snowflake: Double quoted identifiers](https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers)
- [Snowflake: Single-quoted string constants](https://docs.snowflake.com/en/sql-reference/data-types-text#label-single-quoted-string-constants)

VaggelisD:
Fixes #4090

Snowflake allows JSON extraction through the colon operator such as `col:ab`, and double quotes (`col:"ab"`) can optionally be used to include special characters in the key.

This will be roundtripped/transpiled to a string literal in the respective function such as `GET_PATH(..., 'ab')`  which can break queries if single quotes are in the json key.

For this reason, this PR escapes those paths at parse time in `_parse_colon_as_variant_extract()`. Note that Databricks uses backticks for special characters so it won't be affected.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4998,7 +4998,10 @@ class Parser(metaclass=_Parser):
                 # Escape single quotes from Snowflake's colon extraction (e.g. col:"a'b") as
                 # it'll roundtrip to a string literal in GET_PATH
                 if path_sql[0] == '"':
-                    path_sql = path_sql.replace("'", "\\'")
+                    quote_end = self.dialect.QUOTE_END
+                    path_sql = path_sql.replace(
+                        quote_end, self.dialect.tokenizer_class.STRING_ESCAPES[0] + quote_end
+                    )
 
                 json_path.append(path_sql)
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4993,7 +4993,14 @@ class Parser(metaclass=_Parser):
                 end_token = self._prev
 
             if path:
-                json_path.append(self._find_sql(self._tokens[start_index], end_token))
+                path_sql = self._find_sql(self._tokens[start_index], end_token)
+
+                # Escape single quotes from Snowflake's colon extraction (e.g. col:"a'b") as
+                # it'll roundtrip to a string literal in GET_PATH
+                if path_sql[0] == '"':
+                    path_sql = path_sql.replace("'", "\\'")
+
+                json_path.append(path_sql)
 
         # The VARIANT extract in Snowflake/Databricks is parsed as a JSONExtract; Snowflake uses the json_path in GET_PATH() while
         # Databricks transforms it back to the colon/dot notation
--- a/tests/dialects/test_snowflake.py
+++ b/tests/dialects/test_snowflake.py
@@ -2137,6 +2137,16 @@ SINGLE = TRUE""",
         self.validate_identity("SELECT t.$23:a.b", "SELECT GET_PATH(t.$23, 'a.b')")
         self.validate_identity("SELECT t.$17:a[0].b[0].c", "SELECT GET_PATH(t.$17, 'a[0].b[0].c')")
 
+        self.validate_all(
+            """
+            SELECT col:"customer's department"
+            """,
+            write={
+                "snowflake": """SELECT GET_PATH(col, '["customer\\'s department"]')""",
+                "postgres": "SELECT JSON_EXTRACT_PATH(col, 'customer\\'s department')",
+            },
+        )
+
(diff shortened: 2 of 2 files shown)


sqlglot/parser.py
```diff
@@ -4993,7 +4993,14 @@ def _parse_colon_as_variant_extract(
                 end_token = self._prev
 
             if path:
-                json_path.append(self._find_sql(self._tokens[start_index], end_token))
+                path_sql = self._find_sql(self._tokens[start_index], end_token)
+
+                # Escape single quotes from Snowflake's colon extraction (e.g. col:"a'b") as
+                # it'll roundtrip to a string literal in GET_PATH
+                if path_sql[0] == '"':
+                    path_sql = path_sql.replace("'", "\\'")
```
georgesittas, line 5001: I think we can do `self.dialect.generator.escape_str(path_sql)`, which is preferable vs hard-coding the escape.

georgesittas (approved):

sqlglot/parser.py
VaggelisD: That was my intuition as well but `generator` is not a property so we'd have to construct a new Generator for each parsed path:

`path_sql = self.dialect.generator().escape_str(path_sql)`

Should I go ahead and change it anyways?

georgesittas: Ah I see, no then we don't want to instantiate a new generator every time. Can we just make this more generic instead of hardcoding the `\\'`?

VaggelisD: How about the latest approach?

georgesittas: LGTM

Issue #4112, opened 2024-09-13
`optimizer.qualify.qualify` raises "Cannot automatically join" for multiple joins with `USING` and `ON` clauses


dg-hellotwin:
When there are two or more `JOIN` operations, and a `JOIN` with an `ON` condition precedes a `JOIN` with a `USING` clause, the `optimizer.qualify.qualify` method throws the error: **"Cannot automatically join"**.

## Steps to Reproduce

```python
from sqlglot import parse_one
from sqlglot.optimizer.qualify import qualify

schema = {
    'A': {'b_id': 'int'}, 
    'B': {'b_id': 'int', 'c_id': 'int'}, 
    'C': {'c_id': 'int'}
}

sql = "SELECT A.b_id FROM A JOIN B ON A.b_id=B.b_id JOIN C USING(c_id)"
print(qualify(parse_one(sql, read='postgres'), schema=schema, quote_identifiers=False).sql(dialect='postgres'))
```

### Current Behavior
The code throws the following error:

```python
sqlglot.errors.OptimizeError: Cannot automatically join: c_id
```

### Expected Behavior
The optimizer should properly qualify the columns and return the following SQL:

```sql
SELECT a.b_id AS b_id FROM a AS a JOIN b AS b ON a.b_id = b.b_id JOIN c AS c ON b.c_id = c.c_id
```

## Additional Context
The error seems to originate from the `_expand_using` method's handling of `JOIN` clauses. It misinterprets the mix of `ON` and `USING` join strategies, leading to failure.

PR #4113, opened 2024-09-13
fix(optimizer): Enable USING expansion with multiple joins

dg-hellotwin:
fixes tobymao/sqlglot#4112
also added test for this case

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -155,9 +155,6 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
     for i, join in enumerate(joins):
         using = join.args.get("using")
 
-        if not using:
-            continue
-
         join_table = join.alias_or_name
 
         columns = {}
@@ -170,6 +167,10 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
 
         source_table = ordered[-1]
         ordered.append(join_table)
+
+        if not using:
+            continue
+
         join_columns = resolver.get_source_columns(join_table)
         conditions = []
         using_identifier_count = len(using)
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -153,23 +153,21 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
     column_tables: t.Dict[str, t.Dict[str, t.Any]] = {}
 
     for i, join in enumerate(joins):
-        using = join.args.get("using")
-
+        source_table = ordered[-1]
         join_table = join.alias_or_name
+        ordered.append(join_table)
+
+        using = join.args.get("using")
+        if not using:
+            continue
 
         columns = {}
 
         for source_name in scope.selected_sources:
-            if source_name in ordered:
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -170,6 +167,10 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
 
         source_table = ordered[-1]
         ordered.append(join_table)
+
+        if not using:
+            continue
+
```
VaggelisD, line 173: Moving this branch down is enough to solve the bug (idea being `ordered` should contain all join tables regardless of a `USING` clause) however we don't have compute `columns` if that's the case; The order of operations could be:

```
for i, join in enumerate(join):
 join_table = join.alias_or_name
 ordered.append(join_table)
 
  using = join.args.get("using")
  if not using:
    continue

 ...
```

As a sidenote, that nested for-loop looks like a code smell, I'm working on a refactor that can be included in this PR / right after it once it's merged.

VaggelisD (commented):
Hey @dg-hellotwin, really appreciate the contribution!

sqlglot/optimizer/qualify_columns.py
dg-hellotwin: You are welcome.
I've just wanted to provide the simplest way to fix it. 

Unfortunately, your suggestion will not work as we add `join_table` to `ordered` and it looks like it influences to `columns` dict and some tests will fail, for instance:

```
FAIL: test_qualify_columns (tests.test_optimizer.TestOptimizer.test_qualify_columns) [136, SELECT b FROM t JOIN x USING(a)]
  File "/home/dg/projects/test/sqlglot/tests/test_optimizer.py", line 167, in check_file
    self.assertEqual(
AssertionError: 'SELECT x.b AS b FROM t AS t JOIN x AS x ON t.a = x.a' != 'SELECT x.b AS b FROM t AS t JOIN x AS x ON x.a = x.a'
- SELECT x.b AS b FROM t AS t JOIN x AS x ON t.a = x.a
?                                            ^
+ SELECT x.b AS b FROM t AS t JOIN x AS x ON x.a = x.a
?     
```

What will work is code like this:
```python
    for i, join in enumerate(joins):
        columns = {}

        for source_name in scope.selected_sources:
            if source_name in ordered:
                for column_name in resolver.get_source_columns(source_name):
                    if column_name not in columns:
                        columns[column_name] = source_name

        source_table = ordered[-1]
        using = join.args.get("using")
        join_table = join.alias_or_name
        ordered.append(join_table)

        if not using:
            continue
```

so it would be a great idea to refactor code with `columns`

Which solution you will prefer, already committed or proposed?

VaggelisD: Right, we should also hoist up the line `source_table = ordered[-1]` and swap `ordered` for `ordered[:-1]` in the nested for-loop, that should restore the functionality while skipping the (potentially) unnecessary computations.

My refactor regarding `columns` has to do with the observation that we append one item at a time to `ordered` but loop through it again on each iteration, so we try to construct it incrementally by moving `columns = {}` outside of the loop and checking `scope.selected_sources` with the latest entry in ordered afaict; Locally all tests are passing for a quick implementation of this approach, so we can revisit this once this PR is approved.

georgesittas (approved):

VaggelisD (approved):

PR #4117, opened 2024-09-13
Chore(optimizer): rename helper function in expand_using

georgesittas:
Followup for https://github.com/tobymao/sqlglot/pull/4115

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -145,13 +145,6 @@ def _pop_table_column_aliases(derived_tables: t.List[exp.CTE | exp.Subquery]) ->
 
 
 def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
-    columns = {}
-
-    def _gather_source_columns(source_name: str) -> None:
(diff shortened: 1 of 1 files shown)


tobymao (approved):

VaggelisD (approved):

georgesittas:
I decided to revert my original changes, because I'm not sure they were safe after all. Specifically, this [removal](https://github.com/tobymao/sqlglot/pull/4117/commits/56e94332320f65e385c0f8d0ce0da2b4d29e0a97#diff-0e38590a372f13227d4b642e7fbce41497be5d6b65ccc6a6f32e4848e006bd1eL162-L163) could result in us missing the columns of the `FROM` source, e.g. if there were laterals. The reason is that laterals would be picked up as selected sources and so [here](https://github.com/tobymao/sqlglot/pull/4117/commits/56e94332320f65e385c0f8d0ce0da2b4d29e0a97#diff-0e38590a372f13227d4b642e7fbce41497be5d6b65ccc6a6f32e4848e006bd1eL166-L168) we would only add the last lateral's resolved columns in the `column` mapping and leave the columns of its previous sources out. I didn't test this, just reverted based on hunch to avoid breaking stuff.

PR #4125, opened 2024-09-15
Feat!: add `returning` to merge expression builder

max-muoto:
Now that returning is supported, add `returning` argument to the merge expression builder. This PR also makes some fixes to make the AST representation of RETURNING in merge expressions more consistent with other SQL expressions.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -11,7 +11,7 @@ SQL expressions, such as `sqlglot.expressions.select`.
 """
 
 from __future__ import annotations
-
+from typing_extensions import Self
 import datetime
 import math
 import numbers
@@ -1368,7 +1368,7 @@ class DML(Expression):
         dialect: DialectType = None,
         copy: bool = True,
         **opts,
-    ) -> DML:
+    ) -> Self:
         """
         Set the RETURNING expression. Not supported by all dialects.
 
@@ -6274,7 +6274,7 @@ class Use(Expression):
     arg_types = {"this": True, "kind": False}
 
 
-class Merge(Expression):
+class Merge(DML):
     arg_types = {
         "this": True,
         "using": True,
@@ -6883,7 +6883,7 @@ def insert(
     insert = Insert(this=this, expression=expr, overwrite=overwrite)
 
     if returning:
-        insert = t.cast(Insert, insert.returning(returning, dialect=dialect, copy=False, **opts))
+        insert = insert.returning(returning, dialect=dialect, copy=False, **opts)
 
     return insert
 
@@ -6922,10 +6922,7 @@ def merge(
     Returns:
         Merge: The syntax tree for the MERGE statement.
     """
-    returning_expr = (
-        maybe_parse(returning, dialect=dialect, copy=copy, **opts) if returning else None
-    )
-    return Merge(
+    merge_expr = Merge(
         this=maybe_parse(into, dialect=dialect, copy=copy, **opts),
         using=maybe_parse(using, dialect=dialect, copy=copy, **opts),
         on=maybe_parse(on, dialect=dialect, copy=copy, **opts),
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -1368,7 +1368,7 @@ def returning(
         dialect: DialectType = None,
         copy: bool = True,
         **opts,
-    ) -> DML:
+    ) -> Self:
```
max-muoto, line 1371: Correct type, avoids the need for casting with `Insert`/`Merge`.

sqlglot/generator.py
```diff
@@ -3625,13 +3625,15 @@ def merge_sql(self, expression: exp.Merge) -> str:
         using = f"USING {self.sql(expression, 'using')}"
         on = f"ON {self.sql(expression, 'on')}"
         expressions = self.expressions(expression, sep=" ", indent=False)
+        returning = self.sql(expression, "returning")
+        if returning:
```
max-muoto, line 3629: What we had before was somewhat consistent with how `RETURNING` for other SQL expressions worked, this makes it consistent. Returning would be a list of columns (for MERGE), take this example:

```
returning=[
    Anonymous(
      this=merge_action),
    Column(
      this=Star(),
      table=Identifier(this=w, quoted=False))])
```


While for other expressions such as delete/insert, it would be a `Returning` object with expressions:

```
returning=Returning(
  expressions=[
    Star()]))
```

This makes things consistent between INSERT/DELETE/MERGE.

sqlglot/parser.py
```diff
@@ -6841,7 +6841,7 @@ def _parse_merge(self) -> exp.Merge:
             using=using,
             on=on,
             expressions=self._parse_when_matched(),
-            returning=self._match(TokenType.RETURNING) and self._parse_csv(self._parse_bitwise),
+            returning=self._parse_returning()
```
max-muoto, line 6844: Is there any reason not to be using this here?

sqlglot/expressions.py
```diff
@@ -36,6 +35,7 @@
 from sqlglot.tokens import Token, TokenError
 
 if t.TYPE_CHECKING:
+    from typing_extensions import Self
```
max-muoto, line 38: By quoting + putting this under a typing-extensions block, we can still utilize backported typing features while keeping compatibility for those who don't having typing-extensions installed.

georgesittas: Should we do this for Query methods too and get rid of the `Q` type variable defined below?

georgesittas (approved):
LGTM, just one comment. Keep in mind to format PR titles according to [this](https://www.conventionalcommits.org/en/v1.0.0/) guide in the future.

Thanks for the contributions!

sqlglot/parser.py
georgesittas: Probably an oversight? Given no tests failed this is probably fine.

sqlglot/generator.py
georgesittas: Seems like a reasonable refactor.

sqlglot/expressions.py
max-muoto: That would probably be good, would you prefer I do this in this PR, or happy to follow-up with this?

max-muoto:
> LGTM, just one comment. Keep in mind to format PR titles according to [this](https://www.conventionalcommits.org/en/v1.0.0/) guide in the future.
> 
> Thanks for the contributions!

Thanks! Sorry for missing that in the guidelines!

sqlglot/expressions.py
georgesittas: I'm fine with both, your call 👍

max-muoto: I'll follow-up in that case, just to keep the focus of this PR.

VaggelisD (approved):

georgesittas:
Hey @max-muoto, any plans to get this to the finish line?

max-muoto:
> Hey @max-muoto, any plans to get this to the finish line?

I was planning on the typing improvements in a follow-up. Was there anything else you had?

georgesittas:
Ah my bad, for some reason I was thinking there are outstanding items here. Ignore me :)

PR #4126, opened 2024-09-15
Feat: expose `sqlglot.expressions.delete` as a sqlglot module function

max-muoto:
I think it additionally makes sense to have `delete` as a top-level expression.


georgesittas (approved):
Thanks

PR #4127, opened 2024-09-15
Fix!(parser): parse lhs of := into an Identifier if it's not qualified

### LINKED ISSUE #4122: Optimizer Misinterprets Union Identifiers as Columns in DuckDB Queries
When testing the creation of a union with DuckDB, it appears that sqlglot mistakenly treats the union identifier as a column rather than recognizing it as a general identifier when the statement is qualified in the optimizer.

DuckDB uses identifiers to tag unions, as described in [official documentation](https://duckdb.org/docs/sql/data_types/union)

Here is how to reproduce the exception:

```python
import sqlglot
import sqlglot.optimizer

statement = sqlglot.parse_one("select union_value(k1 := 1)", dialect="duckdb")
sqlglot.optimizer.optimize(statement).sql(dialect='duckdb')
```

In this case, `k1` is simply an identifier—just like a key in a struct. It would be helpful if it could be handled in a case-insensitive manner, similar to other struct keys.

The exception raised is as follows:

```
  File "/Users/rusty/Development/flight-cloud/.venv/lib/python3.12/site-packages/sqlglot/optimizer/optimizer.py", line 92, in optimize
    optimized = rule(optimized, **rule_kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rusty/Development/flight-cloud/.venv/lib/python3.12/site-packages/sqlglot/optimizer/qualify.py", line 99, in qualify
  File "/Users/rusty/Development/flight-cloud/.venv/lib/python3.12/site-packages/sqlglot/optimizer/qualify_columns.py", line 106, in validate_qualify_columns
sqlglot.errors.OptimizeError: Column 'k1' could not be resolved
```

In contrast, DuckDB handles this case correctly, as shown below:

```sql
v1.1.0 fa5c2fe15f
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D select union_value(k1 := 1);
┌──────────────────────┐
│ union_value(k1 := 1) │
│  union(k1 integer)   │
├──────────────────────┤
│ 1                    │
└──────────────────────┘
```

Thank you for looking into this.

This behavior in DuckDB demonstrates that identifiers like `k1` should be recognized as keys in a struct rather than treated as columns. Handling this more gracefully in `sqlglot` would align with DuckDB’s behavior.

georgesittas:
Fixes #4122

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4343,7 +4343,7 @@ class Parser(metaclass=_Parser):
             )
 
         while self._match_set(self.ASSIGNMENT):
-            if isinstance(this, exp.Column) and len(this.parts) == 1 and not this.this.quoted:
+            if isinstance(this, exp.Column) and len(this.parts) == 1:
                 this = this.this
 
             this = self.expression(
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4343,6 +4343,9 @@ class Parser(metaclass=_Parser):
             )
 
         while self._match_set(self.ASSIGNMENT):
+            if isinstance(this, exp.Column) and len(this.parts) == 1 and not this.this.quoted:
+                this = exp.var(this.name)
+
             this = self.expression(
                 self.ASSIGNMENT[self._prev.token_type],
                 this=this,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4344,7 +4344,7 @@ class Parser(metaclass=_Parser):
 
         while self._match_set(self.ASSIGNMENT):
             if isinstance(this, exp.Column) and len(this.parts) == 1 and not this.this.quoted:
-                this = exp.var(this.name)
+                this = this.this
 
             this = self.expression(
                 self.ASSIGNMENT[self._prev.token_type],
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3899,7 +3899,10 @@ class Generator(metaclass=_Generator):
         expression.set(
             "expressions",
             [
-                exp.alias_(e.expression, e.name if e.this.is_string else e.this)
+                exp.alias_(
+                    e.expression,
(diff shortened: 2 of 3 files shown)


sqlglot/parser.py
```diff
@@ -4341,6 +4341,9 @@ def _parse_assignment(self) -> t.Optional[exp.Expression]:
             )
 
         while self._match_set(self.ASSIGNMENT):
+            if isinstance(this, exp.Column) and len(this.parts) == 1:
```
tobymao, line 4344: i think it should be a var. i’m surprised we don’t do this already, additionally should we parse union value as a struct?

georgesittas: I think the correct representation is to use an identifier, because the same case-sensitivity rules apply here. For example, in duckdb it's case insensitive so quotes won't matter. Other properties specified using this operator also work with quotes too, e.g. I think you can do `"recursive" => true` in unnest.

> additionally should we parse union value as a struct?

They're not equivalent, see https://duckdb.org/docs/sql/data_types/union.

VaggelisD: Although I thought it wouldn't be correct, quoting DuckDB properties such as `recursive` does work for whatever reason:

```
D select unnest([], recursive := true) as nonquoted, unnest([], "recursive" := true) as quoted;
┌───────────┬────────┐
│ nonquoted │ quoted │
│   int32   │ int32  │
├───────────┴────────┤
│       0 rows       │
└────────────────────┘
```

So parsing the LHS as Identifier seems like the safest bet afaict

VaggelisD (approved):

sqlglot/parser.py
tobymao: is this true for all use of := ? not just unions?

georgesittas: I believe so, extrapolating based on the recursive unnest example.

PR #4128, opened 2024-09-15
Feat(duckdb): add support for the UNION type

georgesittas:
Reference: https://duckdb.org/docs/sql/data_types/union

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -260,6 +260,7 @@ class Parser(metaclass=_Parser):
         TokenType.NESTED,
         TokenType.OBJECT,
         TokenType.STRUCT,
+        TokenType.UNION,
     }
 
     NESTED_TYPE_TOKENS = {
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -522,6 +523,7 @@ class Parser(metaclass=_Parser):
         *TYPE_TOKENS,
         *NO_PAREN_FUNCTIONS,
     }
+    ID_VAR_TOKENS.remove(TokenType.UNION)
```
VaggelisD, line 526: Should this be moved to DuckDB's parser? Although it's true for most of the dialects that I tested, I see that (at least) Spark does not have an issue:

```
spark-sql (default)> with t as (select 1 union) select union from t;
union
1
```

georgesittas: This was intentional, we don't support `union` as an identifier today in main so this PR is not breaking any existing functionality. I think it's out of scope for now to support `UNION` as an identifier token.

VaggelisD (approved):

PR #4132, opened 2024-09-17
Fix(athena): DDL fixes

erindru:
- `DESCRIBE` should be routed to Hive
- `DROP VIEW` should be routed to Trino and not Hive (but every other `DROP` appears to be Hive)
- Athena expects the table location to be in the `external_location` property if the `table_type` property is `hive`, otherwise the table location should be in `location`. So i've added this logic for generating `exp.TableLocation`
- `ALTER TABLE ADD COLUMN` is invalid SQL in Athena. It needs to be `ALTER TABLE ADD COLUMS (..)`. To generate the correct version is a slightly different AST, so I subclassed `Hive.Generator` and overrode `alter_sql()`

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -20,10 +20,35 @@ def _generate_as_hive(expression: exp.Expression) -> bool:
         else:
             return expression.kind != "VIEW"  # CREATE VIEW is never Hive but CREATE SCHEMA etc is
 
-    elif isinstance(expression, exp.Alter) or isinstance(expression, exp.Drop):
-        return True  # all ALTER and DROP statements are Hive
+    # https://docs.aws.amazon.com/athena/latest/ug/ddl-reference.html
+    elif isinstance(expression, (exp.Alter, exp.Drop, exp.Describe)):
+        if isinstance(expression, exp.Drop) and expression.kind == "VIEW":
+            # DROP VIEW is Trino (I guess because CREATE VIEW is)
+            return False
+
+        # Everything else is Hive
+        return True
 
     return False
+
+
+def _location_property_sql(self: Athena.Generator, e: exp.LocationProperty):
+    # If table_type='iceberg', the LocationProperty is called 'location'
+    # Otherwise, it's called 'external_location'
+    # ref: https://docs.aws.amazon.com/athena/latest/ug/create-table-as.html
+
+    is_iceberg = False
+    if isinstance(e.parent, exp.Properties):
+        for rendered_prop in [
+            p.sql(dialect=self.dialect) for p in e.parent.expressions if isinstance(p, exp.Property)
+        ]:
+            if "table_type" and "iceberg" in rendered_prop:
+                is_iceberg = True
+                break
+
+    prop_name = "location" if is_iceberg else "external_location"
+
+    return f"{prop_name}={self.sql(e, 'this')}"
 
 
 class Athena(Trino):
@@ -48,7 +73,7 @@ class Athena(Trino):
     Trino:
       - Uses double quotes to quote identifiers
       - Used for DDL operations that involve SELECT queries, eg:
-        - CREATE VIEW
+        - CREATE VIEW / DROP VIEW
         - CREATE TABLE... AS SELECT
       - Used for DML operations
         - SELECT, INSERT, UPDATE, DELETE, MERGE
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/athena.py
```diff
@@ -79,27 +104,40 @@ class Parser(Trino.Parser):
             TokenType.USING: lambda self: self._parse_as_command(self._prev),
         }
 
+    class _HiveGenerator(Hive.Generator):
+        def preprocess(self, expression: exp.Expression) -> exp.Expression:
+            # package any ALTER TABLE ADD actions into a Schema object
+            # so it gets generated as `ALTER TABLE .. ADD COLUMNS(...)`
+            # instead of `ALTER TABLE ... ADD COLUMN` which is invalid syntax on Athena
+            if isinstance(expression, exp.Alter) and expression.kind == "TABLE":
```
erindru, line 112: The goal here is to correctly transform `ALTER TABLE ADD COLUMN` to `ALTER TABLE ADD COLUMNS` without the user having to do anything manually to the AST.

Is this the right place for AST manipulations that need the full context of the expression?

sqlglot/dialects/athena.py
```diff
@@ -79,27 +104,40 @@ class Parser(Trino.Parser):
...
-            exp.DataType.Type.TEXT: "STRING",
+        PROPERTIES_LOCATION = {
+            **Trino.Generator.PROPERTIES_LOCATION,
+            exp.LocationProperty: exp.Properties.Location.POST_WITH,
         }
 
         TRANSFORMS = {
             **Trino.Generator.TRANSFORMS,
-            exp.FileFormatProperty: lambda self, e: f"'FORMAT'={self.sql(e, 'this')}",
+            exp.FileFormatProperty: lambda self, e: f"format={self.sql(e, 'this')}",
```
erindru, line 131: Removing the single quotes is correct here. The Hive engine needs the single quotes but the Trino engine throws an error if you use them.

In addition, I lowercased the property for consistency with every other property

sqlglot/dialects/athena.py
```diff
@@ -20,12 +20,37 @@ def _generate_as_hive(expression: exp.Expression) -> bool:
...
 
+def _location_property_sql(self: Athena.Generator, e: exp.LocationProperty):
+    # If table_type='iceberg', the LocationProperty is called 'location'
+    # Otherwise, it's called 'external_location'
+    # ref: https://docs.aws.amazon.com/athena/latest/ug/create-table-as.html
+
+    is_iceberg = False
+    if isinstance(e.parent, exp.Properties):
+        for rendered_prop in [
+            p.sql(dialect=self.dialect) for p in e.parent.expressions if isinstance(p, exp.Property)
```
tobymao, line 43: this is only used for iceberg?

could you do like p.name or something?

sqlglot/dialects/athena.py
tobymao: this is one place, you could also do a transform on Alter

tobymao: this doesn't do too much with context, so probably better as a transform, look up TRANSFORMS

sqlglot/dialects/athena.py
```diff
@@ -37,20 +37,39 @@ def _location_property_sql(self: Athena.Generator, e: exp.LocationProperty):
...
+        table_type_property = next(
+            (
+                p
+                for p in e.parent.expressions
+                if isinstance(p, exp.Property) and p.name == "table_type"
+            ),
+            None,
+        )
+        if (
+            table_type_property
```
tobymao, line 52: does prop.text(“valid”) work?

sqlglot/dialects/athena.py
erindru: Sure, i've refactored it to use `p.name`.

First, it needs to find the `exp.Property` object that represents the `table_type` property. Then it needs to extract the `value` member from that object and see if that expression equates to "iceberg".

The easy way out was to render the property and then check if both `table_type` and `iceberg` were in the resulting string

sqlglot/dialects/athena.py
erindru: I've moved it to a transform

sqlglot/dialects/athena.py
```diff
@@ -20,12 +20,56 @@ def _generate_as_hive(expression: exp.Expression) -> bool:
...
+            and table_type_property.value
+            and table_type_property.value.name == "iceberg"
+        ):
+            prop_name = "location"
+
+    return f"{prop_name}={self.sql(e, 'this')}"
+
+
+def _alter_sql(self: Athena._HiveGenerator, e: exp.Alter):
```
tobymao, line 61: this is not really a transform, but it's fine, can you move this directly into the Generate and make it a proper subclass method instead of using transforms

sqlglot/dialects/athena.py
erindru: I've moved it out of a transform and just override `alter_sql()` directly on the Generator

sqlglot/dialects/athena.py
erindru: prop.text("value") works

PR #4138, opened 2024-09-19
feat: Add support for GRANT DDL

VaggelisD:
The intention of this PR is to add the initial/foundational work for `GRANT` DDL support. This statement has a number of differences between dialects and virtually endless securables, options etc so this is an attempt to cover the important/heavy-hitter subset by preserving every parsed keyword; For this reason, there are lots of `COMMAND` tests added which should become identities as coverage is increased.

The terminology used is borrowed from T-SQL/Databricks docs:

```
GRANT privilege_types ON securable_object TO principal
``` 

Docs
-----------
[Databricks](https://docs.databricks.com/en/sql/language-manual/security-grant.html) | [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/grant-privilege) | [Presto](https://prestodb.io/docs/current/sql/grant.html) | [Trino](https://trino.io/docs/current/sql/grant.html) | [MySQL](https://dev.mysql.com/doc/refman/8.4/en/grant.html) | [Postgres](https://www.postgresql.org/docs/current/sql-grant.html) | [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/grant-transact-sql?view=sql-server-ver16) | [Clickhouse](https://clickhouse.com/docs/en/sql-reference/statements/grant) | [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/GRANT.html) | [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_GRANT.html) | [Redshift GRANT examples](https://docs.aws.amazon.com/redshift/latest/dg/r_GRANT-examples.html)


sqlglot/expressions.py
```diff
@@ -2314,6 +2314,16 @@ class Fetch(Expression):
     }
 
 
+class Grant(Expression):
+    arg_types = {
+        "privilages": True,
```
georgesittas, line 2319: There's a typo here:
```suggestion
        "privileges": True,
```

sqlglot/expressions.py
```diff
@@ -2314,6 +2314,16 @@ class Fetch(Expression):
...
+        "securable_kind": False,
```
georgesittas, line 2320: Should this just be `kind`? Seems inconsistent with `Create`: we don't have `creatable_kind` there, it's implied.

sqlglot/expressions.py
```diff
@@ -2570,6 +2580,14 @@ class Property(Expression):
     arg_types = {"this": True, "value": True}
 
 
+class GrantPrivilage(Expression):
```
georgesittas, line 2583: Ditto about the typo, let's make sure to fix all references.

sqlglot/parser.py
```diff
@@ -7401,3 +7402,62 @@ def _parse_star_ops(self) -> exp.Star | exp.UnpackColumns:
...
+            (TokenType.ON, TokenType.COMMA, TokenType.L_PAREN), advance=False
+        ):
+            privilage_parts.append(self._curr.text.upper())
+            self._advance()
+
+        this = exp.var(" ".join(privilage_parts))
+        expressions = (
+            self._parse_csv(self._parse_column) if self._match(TokenType.L_PAREN) else None
+        )
+        self._match(TokenType.R_PAREN)
```
georgesittas, line 7421: May be able to use `expressions = self._parse_wrapped_csv(..., optional=True)` here.

georgesittas (commented):
Quick pass from my end, I'll need to take a closer look at the grammars to review more carefully. Nice work on the coverage, let's make sure there are no weird paths left out.

Can also search the SQLMesh Slack for some real use-cases and ensure the references that come up are handled properly.

sqlglot/expressions.py
VaggelisD: Right, this was initially named so to distinguish from `principal_kind` but later on I split that in the separate node `GrantPrincipal`.

It's not consistent with the codebase but it's also not directly connected to `GRANT` (in contrast to `CREATE`) since there's many different constructs at play here (securables, principals etc). I'll revert this to `kind` though for simplicity.

sqlglot/parser.py
VaggelisD: On it's own it breaks tests as it consumes nearby privileges from what I see; We can still use it to save on the `R_PAREN` match line though

sqlglot/parser.py
```diff
@@ -7401,3 +7402,63 @@ def _parse_star_ops(self) -> exp.Star | exp.UnpackColumns:
...
+            self._parse_wrapped_csv(self._parse_column)
+            if self._match(TokenType.L_PAREN, advance=False)
+            else None
+        )
+
+        return self.expression(exp.GrantPrivilege, this=this, expressions=expressions)
+
+    def _parse_grant_principal(self) -> t.Optional[exp.GrantPrincipal]:
+        kind = self._match_texts(("ROLE", "GROUP")) and self._prev.text.upper()
+        principal = self._try_parse(self._parse_id_var)
```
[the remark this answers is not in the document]
VaggelisD: I wanted to make sure that `GRANT` is parsed as safely as possible but indeed `_parse_id_var()` should be a safe/stateless function compared to e.g. `_parse_table_parts()`.

PR #4141, opened 2024-09-20
feat(clickhouse): Add support for APPLY query modifier

### LINKED ISSUE #4139: Clickhouse COLUMNS APPLY parsing error
I was trying to use dynamic columns selection with APPLY, but sqlglot fails to parse it

`select COLUMNS('.*field') APPLY(toUInt64) FROM ABC;`

Leads to sqlglot.errors.ParseError: Invalid expression / Unexpected token.

**Official Documentation**
https://clickhouse.com/docs/en/sql-reference/statements/select#apply

VaggelisD:
Fixes #4139

Clickhouse supports the `APPLY` query modifier:

```SQL
SELECT <expr> APPLY( <func> ) FROM [db.]table_name
```

Although `apply` is still a valid alias, we shouldn't parse it as such if it's accompanied by an `L_PAREN` as is the case here, thus the overriden `_parse_alias()` in Clickhouse:

```SQL
clickhouse> select 1 apply;

   ┌─apply─┐
1. │     1 │
   └───────┘

clickhouse> with t as (Select 1 as col) select COLUMNS('col') apply(sum) from t;

   ┌─sum(col)─┐
1. │        1 │
   └──────────┘
```

Note that `APPLY` can be chained, as is shown in the docs & relevant test case

Docs
---------
[Clickhouse APPLY modifier](https://clickhouse.com/docs/en/sql-reference/statements/select#apply)

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -756,6 +756,16 @@ class ClickHouse(Dialect):
         def _parse_constraint(self) -> t.Optional[exp.Expression]:
             return super()._parse_constraint() or self._parse_projection_def()
 
+        def _parse_alias(
+            self, this: t.Optional[exp.Expression], explicit: bool = False
+        ) -> t.Optional[exp.Expression]:
+            # In clickhouse "SELECT <expr> APPLY(...)" is a query modifier,
+            # so "APPLY" shouldn't be parsed as <expr>'s alias. However, "SELECT <expr> apply" is a valid alias
+            if self._match_pair(TokenType.APPLY, TokenType.L_PAREN, advance=False):
+                return this
+
+            return super()._parse_alias(this=this, explicit=explicit)
+
     class Generator(generator.Generator):
         QUERY_HINTS = False
         STRUCT_DELIMITER = ("(", ")")
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -766,6 +766,16 @@ class ClickHouse(Dialect):
 
             return super()._parse_alias(this=this, explicit=explicit)
 
+        def _parse_projections(self) -> t.List[exp.Expression]:
+            exprs = self._parse_expressions()
+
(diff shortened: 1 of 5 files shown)


tobymao:
is this the right approach? should we treat this similarly to window functions?

VaggelisD:
@tobymao  Clickhouse is a bit lax with the rules, the grammar etc so I struggled a bit to not break any functionality as with the alias. Regarding window functions, which part would that be, reusing `_parse_query_modifiers()`?

Note that `APPLY` comes before the query's `FROM` so we must either parse it early on it's own (the approach here) or add it in `QUERY_MODIFIER_PARSERS` and call `_parse_query_modifiers()` in the place of my current while loop.

sqlglot/dialects/clickhouse.py
```diff
@@ -756,6 +756,26 @@ def _parse_projection_def(self) -> t.Optional[exp.ProjectionDef]:
...
+                return this
+
+            return super()._parse_alias(this=this, explicit=explicit)
+
+        def _parse_projections(self) -> t.List[exp.Expression]:
+            exprs = self._parse_expressions()
+
+            # Clickhouse allows "SELECT <expr> APPLY(func) [...,]" modifier
+            while self._match_pair(TokenType.APPLY, TokenType.L_PAREN):
+                exprs[-1] = exp.Apply(this=exprs[-1], expression=self._parse_var(any_token=True))
```
georgesittas, line 774: I'm not sure if this is the right way to set this, because the `exp.Apply` won't get its `index` set. Let's instead use the `set` method and specify the index. Can we have >1 projections _and_ an `APPLY` operator? We can try to simplify, e.g. by always setting at index 0 or something if there can only be 1 projection.

Regarding `_parse_var` - can you have namespaced functions? E.g. `schema.func` won't be parsed here.

georgesittas: Both errors re: namespaced functions seem reasonable, I don't see something in the messages that leads me to believe it's unsupported. Maybe what you tried wasn't right?

Re: `APPLY` and projections - you don't need `_parse_projection`, you could just override `_parse_expression` for CH.

PR #4143, opened 2024-09-22
fix(databricks): avoid inserting bogus `USING PARQUET` DDL when creating temporary tables

### LINKED ISSUE #4142: using temporary tables with the databricks dialect inserts incorrect `USING PARQUET` DDL
```
In [6]: import sqlglot as sg, sqlglot.expressions as sge

In [7]: sg.__version__
Out[7]: '25.22.0'

In [8]: sg.parse_one('create temporary table t (x int) using delta', read="databricks").sql('databricks')
Out[8]: 'CREATE TEMPORARY TABLE t (x INT) USING DELTA USING PARQUET'
```

cpcloud:
Fixes #4142.

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -10,6 +10,11 @@ from sqlglot.dialects.dialect import (
 )
 from sqlglot.dialects.spark import Spark
 from sqlglot.tokens import TokenType
+from sqlglot.transforms import (
+    remove_unique_constraints,
+    preprocess,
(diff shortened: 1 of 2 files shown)


cpcloud:
Hm, it seems like `TEMP` **does** require a provider.

cpcloud:
`CREATE TEMPORARY TABLE t (x INT) USING PARQUET` doesn't work on databricks, but that's not sqlglot's problem, it's something databricks needs to fix.

PR #4144, opened 2024-09-23
fix(bigquery): Consume dashed identifiers only if they're connected

VaggelisD:
[Public slack context](https://app.slack.com/client/T041MTMK2JC/C0448SFS3PF)

In BigQuery, while parsing dashed identifiers, we were not checking if they're connected which meant that we'd keep consuming keywords until `TokenType.DOT` / EOF.


sqlglot/dialects/bigquery.py
```diff
@@ -480,7 +480,11 @@ def _parse_table_part(self, schema: bool = False) -> t.Optional[exp.Expression]:
                 table_name = this.name
                 while self._match(TokenType.DASH, advance=False) and self._next:
                     text = ""
-                    while self._curr and self._curr.token_type != TokenType.DOT:
+                    while (
+                        self._curr
```
tobymao, line 484: you don't need the self._curr check because self._is_connected also checks self._curr

tobymao (approved):

PR #4153, opened 2024-09-24
feat: Add SUBSTR Support

mwc360:
SUBSTR was not getting parsed. To add parse support via `_sql_names` I replaced SUBSTR with SUBSTRING in the transpile tests and added a distinct test in the test_functions method to ensure that they both get parsed as Substring.

image

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6023,6 +6023,7 @@ class Split(Func):
 # Start may be omitted in the case of postgres
 # https://www.postgresql.org/docs/9.1/functions-string.html @ Table 9-6
 class Substring(Func):
+    _sql_names = ["SUBSTRING", "SUBSTR"]
     arg_types = {"this": True, "start": False, "length": False}
 
 
--- a/tests/dialects/test_dialect.py
+++ b/tests/dialects/test_dialect.py
@@ -2221,9 +2221,9 @@ SELECT
         self.validate_all(
             "SUBSTR('123456', 2, 3)",
             write={
-                "bigquery": "SUBSTR('123456', 2, 3)",
+                "bigquery": "SUBSTRING('123456', 2, 3)",
                 "oracle": "SUBSTR('123456', 2, 3)",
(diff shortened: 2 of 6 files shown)


mwc360:
@georgesittas I'm unsure how to best resolve the unit tests for this change. To date, use of SUBSTR wasn't getting parsed and showed in the AST as 'Anonymous'. My change fixes this but the unit tests (i.e. test_dialect.test_substring and others) are written to expect that the user provided variant of the function is returned (although use of SUBSTR was never getting parsed so there was a test gap), the tests now fail as SUBSTRING is primary when SUBSTR is a secondary option.

I'm thinking that all unit tests which current expect to return SUBSTR should be updated to expect SUBSTRING since that is the primary _sql_name. LMK if there's a better option here as this will potentially change a lot of tests. THX!

tobymao:
you could for now change it to default to substr

mwc360:
> you could for now change it to default to substr, so that it doesn't break

Tests seem to expect that the function variant used is what should be returned when transpiled, i.e. the second `validate_all` block in the below would fail on bigquery and postgres.

https://github.com/tobymao/sqlglot/blob/89519bba99fc11f17e8e00bf8e3f6dde213e99be/tests/dialects/test_dialect.py#L2220C1-L2236C10

Same for the tpch transpile tests, 'substring' is expected instead of 'substr':
https://github.com/tobymao/sqlglot/blob/89519bba99fc11f17e8e00bf8e3f6dde213e99be/tests/fixtures/optimizer/tpc-h/tpc-h.sql#L1384

tobymao:
if substring is the more common variant, you can do that. change the tests. if substring is not allowed, transpile it to substr

mwc360:
> if substring is the more common variant, you can do that. change the tests. if substring is not allowed, transpile it to substr

@tobymao Thx, PR is good to go! Thanks for all your work BTW, SQLGlot is revolutionary :)

georgesittas (approved):
Thanks for the contribution, @mwc360! :)

PR #4156, opened 2024-09-25
feat(spark): Custom annotation for more string functions

VaggelisD:
This is a follow up on https://github.com/tobymao/sqlglot/pull/4004, as during my investigation for `SUBSTRING` I came across other Spark string functions (`CONCAT`, `LPAD`, `RPAD`) which are input type dependent.

Although it's not documented clearly, these functions will return a `BINARY` only if all their "string" operands are `BINARY`, otherwise they return a `STRING`:

```
spark-sql (default)> with tbl as (select cast('test' as binary) as bin_col, 'str' as str_col) 
select 
 typeof(concat(bin_col, bin_col)), 
 typeof(concat(str_col, bin_col)), 
 typeof(concat(bin_col, str_col)), 
 typeof(concat(str_col, str_col)) 
from tbl;

binary  string  string  string
```

```
spark-sql (default)> with tbl as (select cast('test' as binary) as bin_col, 'str' as str_col) 
select 
 typeof(lpad(bin_col, 1, bin_col)), 
 typeof(lpad(str_col, 1, bin_col)), 
 typeof(lpad(bin_col, 1, str_col)), 
 typeof(lpad(str_col, 1, str_col))
from tbl;

binary  string  string  string
```

Docs
---------
[Databricks LPAD](https://docs.databricks.com/en/sql/language-manual/functions/lpad.html) | [Databricks CONCAT](https://docs.databricks.com/en/sql/language-manual/functions/concat.html)

### DIFF
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -13,13 +13,18 @@ from sqlglot.dialects.dialect import (
     unit_to_str,
 )
 from sqlglot.dialects.hive import Hive
-from sqlglot.helper import seq_get
+from sqlglot.helper import seq_get, ensure_list
 from sqlglot.transforms import (
     preprocess,
     remove_unique_constraints,
     ctas_with_tmp_tables_to_create_tmp_view,
     move_schema_columns_to_partitioned_by,
 )
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
+
+    from sqlglot.optimizer.annotate_types import TypeAnnotator
 
 
 def _map_sql(self: Spark2.Generator, expression: exp.Map) -> str:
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/spark2.py
```diff
@@ -110,10 +115,41 @@ def temporary_storage_provider(expression: exp.Expression) -> exp.Expression:
...
+
+    last_datatype = expressions[0].type if expressions else None
+
+    for expr in expressions:
+        if not expr.is_type(last_datatype):
+            last_datatype = None
+            break
+
+    self._set_type(expression, last_datatype or target_type)
+    return expression
```
georgesittas, line 140: What if you have unknowns? This would preemptively infer the type to be `TEXT` for `CONCAT` even if you had `UNKNOWN` args.

tests/test_optimizer.py
```diff
@@ -1378,25 +1378,31 @@ def gen_expr(depth: int) -> exp.Expression:
...
 
-    def test_custom_annotators(self):
+    def test_spark_annotators(self):
+        """Test Spark annotators, mainly built-in string/binary functions"""
+
+        schema = {"tbl": {"bin_col": "BINARY", "str_col": "STRING"}}
+
+        def _assert_func_return_type(func: str, dialect: str, target_type: str):
+            ast = parse_one(f"SELECT {func} FROM tbl", read=dialect)
+            optimized = optimizer.optimize(ast, schema=schema, dialect=dialect)
```
georgesittas, line 1388: Let's only use `annotate_types` for this, running all optimizer rules is too much. You can slightly rewrite the query so that the columns are annotated and there's no need to even use `qualify`.

georgesittas (commented):
Can we consolidate `_annotate_by_same_args` and `_annotate_by_args` somehow? They look very similar, perhaps a flag could help?

sqlglot/dialects/spark2.py
VaggelisD: That should be fine I think, in the majority of the cases that'll be the result type anyways.

georgesittas: If the last type is `BINARY` and the rest are `UNKNOWN`, you then assume `TEXT` even though `BINARY` is also possible (e.g. if the unknown types were `BINARY`).

VaggelisD:
> Can we consolidate _annotate_by_same_args and _annotate_by_args somehow? They look very similar, perhaps a flag could help?

Some parts are overlapping but the "access pattern" is different. I think as each dialect is starting to form their own annotators we should keep that logic separate from the common functions; If some boilerplate parts are getting repetitive, we should of course factor these out in time.

Regarding the Spark logic, I came up with a new set of rules for that annotator that biases `STRING` less now (that still remains the fallback type though). What do you think of that?

georgesittas (approved):
Let's test unknowns as well

VaggelisD:
@georgesittas  I did add a few test cases with UNKNOWNs in the previous commit, do you have more cases in mind?

georgesittas:
> @georgesittas I did add a few test cases with UNKNOWNs in the previous commit, do you have more cases in mind?

Ah I didn't see them, apologies. LGTM.

PR #4161, opened 2024-09-25
fix(clickhouse)!: Generalize COLUMNS(...) APPLY

### LINKED ISSUE #4157: Clickhouse COLUMNS APPLY parsing error in functions
APPLY issue in function

`SELECT * FROM ABC WHERE hasAny(COLUMNS('.*field') APPLY (toUInt64), (SELECT groupUniqArray(toUInt64(field))`
Leads to 
`sqlglot.errors.ParseError: Expecting )`
**Official Documentation**
https://clickhouse.com/docs/en/sql-reference/statements/select#apply

Relates to https://github.com/tobymao/sqlglot/issues/4139

VaggelisD:
Fixes #4157

As `COLUMNS(...)` is a function that can be used in different contexts besides the projection list, this PR adds `exp.Columns` which parses the function call _and_ the following `APPLY(...)` functions. 

The `APPLY` parsing in `clickhouse.py::_parse_expression()` is left as is to accommodate `exp.Star` and potentially any other expression that can utilize `APPLY` in the selections.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -396,6 +396,7 @@ class ClickHouse(Dialect):
             **parser.Parser.FUNCTION_PARSERS,
             "ARRAYJOIN": lambda self: self.expression(exp.Explode, this=self._parse_expression()),
             "QUANTILE": lambda self: self._parse_quantile(),
+            "COLUMNS": lambda self: self._parse_columns(),
         }
 
         FUNCTION_PARSERS.pop("MATCH")
@@ -774,6 +775,14 @@ class ClickHouse(Dialect):
                 this = exp.Apply(this=this, expression=self._parse_var(any_token=True))
                 self._match(TokenType.R_PAREN)
 
+            return this
+
+        def _parse_columns(self) -> exp.Expression:
+            this: exp.Expression = self.expression(exp.Columns, this=self._parse_lambda())
+
+            while self._next and self._match_text_seq(")", "APPLY", "("):
+                self._match(TokenType.R_PAREN)
+                this = exp.Apply(this=this, expression=self._parse_var(any_token=True))
             return this
 
     class Generator(generator.Generator):
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4984,6 +4984,10 @@ class ToNumber(Func):
     }
 
 
+class Columns(Func):
+    pass
+
+
 # https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16#syntax
 class Convert(Func):
     arg_types = {"this": True, "expression": True, "style": False}
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4985,7 +4985,7 @@ class ToNumber(Func):
 
 
 class Columns(Func):
-    pass
+    arg_types = {"this": True, "unpack": False}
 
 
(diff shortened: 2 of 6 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -776,6 +777,14 @@ def _parse_expression(self) -> t.Optional[exp.Expression]:
 
             return this
 
+        def _parse_columns(self) -> exp.Expression:
+            this: exp.Expression = self.expression(exp.Columns, this=self._parse_lambda())
+
+            while self._next and self._match_text_seq(")", "APPLY", "("):
+                self._match(TokenType.R_PAREN)
+                this = exp.Apply(this=this, expression=self._parse_var(any_token=True))
```
georgesittas, line 785: Can this only be applied to `COLUMNS`? I was thinking we could add this after `_parse_column` or something, to make it more generic and not have to deal with special cases again. I'd also try to avoid duplicating this logic and/or consolidate the `COLUMNS` parsing with what we did for duckdb if possible, looks like there may be some overlap.

VaggelisD: From my understanding yes, the `APPLY` seems to follow `*` and `COLUMNS`, it doesn't work with any expression thus the specialized solution. However, what I didn't expect in the initial PR is that `COLUMNS(...)` can be used in other scenarios as well.

Note that the recent DuckDB PR added support `SELECT *COLUMNS(...)` which is a special case/operator (column unpacking), not sure if these should be consolidated in one representation.

georgesittas: What's `COLUMNS` used for in CH? It looks like it also does dynamic column expansion, hence my suggestion. I'd try out DuckDB's syntax `*COLUMNS(...)` to see if it works for CH as well, but even if it doesn't I'd still consider consolidating them if they represent the same thing.

VaggelisD: DuckDB also supports `COLUMNS(...)` [[1]](https://duckdb.org/docs/sql/query_syntax/select.html#star-expressions), the unpacking operator is so that the expansion is done into a function [[2]](https://duckdb.org/2024/09/09/announcing-duckdb-110.html#unpacked-columns).

elchyn-cheliabiyeu: @georgesittas it is a dynamic column selection https://clickhouse.com/blog/sql-dynamic-column-selection-clickhouse

elchyn-cheliabiyeu: Also consequent applys are also valid like this 

```
  COLUMNS('.*_amount|fee|tax')
  APPLY(avg)
  APPLY(round)
```

georgesittas: > DuckDB also supports `COLUMNS(...)` [[1]](https://duckdb.org/docs/sql/query_syntax/select.html#star-expressions), the unpacking operator is so that the expansion is done into a function [[2]](https://duckdb.org/2024/09/09/announcing-duckdb-110.html#unpacked-columns).

I see, makes sense I guess. Ok cool, so can we still consolidate the existing `APPLY (...)` implementations though?

VaggelisD: Not sure, because e.g. for the projection list we parse expressions (i.e star or `COLUMNS`) but if `COLUMNS(..)` is picked up by `_parse_function()`, the subsequent `APPLY(...)` won't be picked. 

We could try to move `apply` to `parse_star_ops()` so that it's at least removed from `_parse_expressions()`.

georgesittas (approved):
Let's mark this as breaking.

PR #4163, opened 2024-09-25
fix(hive)!: Enclose exp.Split with \E

### LINKED ISSUE #4158: split with `\\Q` is invalid on databricks
image

VaggelisD:
Fixes #4158

Docs
------
[SO Thread](https://stackoverflow.com/questions/30776860/regular-expression-q-e)


georgesittas (approved):
I think this is breaking?

@tobymao it's been a long time [since you added the original logic](https://github.com/tobymao/sqlglot/commit/c2706ed4d9e64e156b9c4bfb1d29f715d5f3624) for `SPLIT`, but figured I'd tag you just in case you had additional context.

tobymao (approved):

PR #4171, opened 2024-09-27
Fix(athena): Case sensitivity in CTAS property names

erindru:
It turns out Athena table properties are case sensitive on the Trino engine and need to be lowercase.

```
CREATE TABLE "test_table" WITH (PARTITIONED_BY=ARRAY['ds'], ...) AS SELECT ...
```
Fails with `Table properties [PARTITIONED_BY] are not supported.`

```
CREATE TABLE "test_table" WITH (partitioned_by=ARRAY['ds'], ...) AS SELECT ...
```
Succeeds

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -48,7 +48,7 @@ def _location_property_sql(self: Athena.Generator, e: exp.LocationProperty):
             ),
             None,
         )
-        if table_type_property and table_type_property.text("value") == "iceberg":
+        if table_type_property and table_type_property.text("value").lower() == "iceberg":
             prop_name = "location"
 
     return f"{prop_name}={self.sql(e, 'this')}"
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/athena.py
```diff
@@ -132,6 +132,7 @@ class Generator(Trino.Generator):
         TRANSFORMS = {
             **Trino.Generator.TRANSFORMS,
             exp.FileFormatProperty: lambda self, e: f"format={self.sql(e, 'this')}",
+            exp.PartitionedByProperty: lambda self, e: f"partitioned_by={self.sql(e, 'this')}",
```
erindru, line 135: Note that the actual reason for the uppercasing is because the `Properties.NAME_TO_PROPERTY` map hardcodes everything to uppercase.

Since we already had a transform to lowercase `exp.FileFormatProperty`, I added one for `exp.PartitionedByProperty` since most of the properties in `Properties.NAME_TO_PROPERTY` arent relevant for Athena

georgesittas (approved):

PR #4172, opened 2024-09-27
fix(bigquery): Do not generate null ordering on agg funcs

### LINKED ISSUE #4170: Generated SQL for Array Aggregations with Order By doesn't work in BigQuery
Hi there, I ran into an [issue](https://github.com/ibis-project/ibis/issues/10236) using Ibis and I was able to trace the problem back `sqlglot`.

A minimal (ish) reproduction is

```
>>> from sqlglot import select, expressions as exp
>>>
>>> query = (
...     select(
...         exp.Column(this="color", table="t0"),
...         exp.ArrayAgg(
...             this=exp.Order(
...                 this=exp.Column(this="id"),
...                 expressions=[
...                     exp.Ordered(
...                         this=exp.Column(this="id"),
...                     )
...                 ]
...             ),
...             ignore_nulls=True,
...         ).as_("ids"),
...     )
...     .from_("colors")
...     .group_by(exp.Literal.number(1))
... )
>>>
>>> print(query.sql(dialect="bigquery"))
SELECT t0.color, ARRAY_AGG(id ORDER BY id NULLS LAST) AS ids FROM colors GROUP BY 1
```

which I ran this against main. I also made a table in BigQuery 

```
CREATE TABLE `my-project.my_dataset.colors` AS (
  SELECT 1 AS id, 'red' AS color
  UNION ALL
  SELECT 2 AS id, 'red' AS color
  UNION ALL
  SELECT 3 AS id, 'blue' AS color
)
```

So I could actually run the query. I get

```
BadRequest: 400 POST https://bigquery.googleapis.com/bigquery/v2/projects/my-project/queries?prettyPrint=false: NULLS LAST not supported with ascending sort order in aggregate functions.
```

I fudged with the table names (project and dataset) but that's it. I tried to find some documentation but I couldn't. According to [this](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#order_by_clause) you'd think NULLS LAST would be supported.

VaggelisD:
Fixes #4170

BigQuery does not allow `ASC NULLS LAST` and `DESC NULLS FIRST` ordering in aggregate functions:

```
with colors as (SELECT 1 AS id, 'red' AS color) SELECT color, array_agg(id ORDER BY id ASC NULLS LAST) AS ids FROM colors GROUP BY 1

NULLS LAST not supported with ascending sort order in aggregate functions.
```

```
with colors as (SELECT 1 AS id, 'red' AS color) SELECT color, array_agg(id ORDER BY id DESC NULLS FIRST) AS ids FROM colors GROUP BY 1

NULLS FIRST not supported with descending sort order in aggregate functions.
```

Note that BigQuery is the only dialect with `NULL_ORDERING_SUPPORTED = False` so this PR uses this as a proxy to handle this case.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -203,7 +203,8 @@ class Generator(metaclass=_Generator):
     }
 
     # Whether null ordering is supported in order by
-    # True: Full Support, None: No support, False: No support in window specifications
+    # True: Full Support, None: No support, False: No support for certain cases
+    # such as window specifications, aggregate functions etc
     NULL_ORDERING_SUPPORTED: t.Optional[bool] = True
 
     # Whether ignore nulls is inside the agg or outside.
@@ -2343,6 +2344,20 @@ class Generator(metaclass=_Generator):
             if isinstance(window, exp.Window) and window.args.get("spec"):
                 self.unsupported(
                     f"'{nulls_sort_change.strip()}' translation not supported in window functions"
+                )
+                nulls_sort_change = ""
+            elif (
+                self.NULL_ORDERING_SUPPORTED is False
+                and expression.find_ancestor(exp.AggFunc)
+                and (
+                    asc
+                    and nulls_sort_change == " NULLS LAST"
+                    or desc
+                    and nulls_sort_change == " NULLS FIRST"
+                )
+            ):
+                self.unsupported(
+                    f"'{nulls_sort_change.strip()}' translation not supported for aggregate functions with {sort_order} sort order"
                 )
                 nulls_sort_change = ""
             elif self.NULL_ORDERING_SUPPORTED is None:
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2348,12 +2348,10 @@ class Generator(metaclass=_Generator):
                 nulls_sort_change = ""
             elif (
                 self.NULL_ORDERING_SUPPORTED is False
-                and expression.find_ancestor(exp.AggFunc)
+                and (isinstance(expression.find_ancestor(exp.AggFunc, exp.Select), exp.AggFunc))
                 and (
-                    asc
-                    and nulls_sort_change == " NULLS LAST"
-                    or desc
-                    and nulls_sort_change == " NULLS FIRST"
+                    (asc and nulls_sort_change == " NULLS LAST")
+                    or (desc and nulls_sort_change == " NULLS FIRST")
                 )
             ):
                 self.unsupported(
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1985,3 +1985,17 @@ OPTIONS (
         self.validate_identity(
             "SELECT RANGE(CAST('2022-10-01 14:53:27 America/Los_Angeles' AS TIMESTAMP), CAST('2022-10-01 16:00:00 America/Los_Angeles' AS TIMESTAMP))"
         )
+
+    def test_null_ordering(self):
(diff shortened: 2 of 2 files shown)


tobymao (approved):

sqlglot/generator.py
```diff
@@ -2345,6 +2346,20 @@ def ordered_sql(self, expression: exp.Ordered) -> str:
...
                 )
                 nulls_sort_change = ""
+            elif (
+                self.NULL_ORDERING_SUPPORTED is False
+                and expression.find_ancestor(exp.AggFunc)
+                and (
+                    asc
+                    and nulls_sort_change == " NULLS LAST"
+                    or desc
+                    and nulls_sort_change == " NULLS FIRST"
```
georgesittas, line 2356: Use parentheses so that the precedence is clear here.

sqlglot/generator.py
```diff
@@ -2345,6 +2346,20 @@ def ordered_sql(self, expression: exp.Ordered) -> str:
                     f"'{nulls_sort_change.strip()}' translation not supported in window functions"
                 )
                 nulls_sort_change = ""
+            elif (
+                self.NULL_ORDERING_SUPPORTED is False
+                and expression.find_ancestor(exp.AggFunc)
```
georgesittas, line 2351: 1. You can do `expression.find_ancestor(exp.AggFunc, exp.Select) is exp.AggFunc` to avoid walking up the whole AST (i.e. a perf. improvement)
2. What if there's no `AggFunc` but this is used in the context of an `ORDER BY`? Is this handled in the following `elif` branch or something?

VaggelisD: Re (2): This is why BQ is `False` instead of `None` afaict, because in the context of a simple [ORDER BY](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#order_by_clause) we can use both sort & null order. However, that is not the case in window specs & aggregate functions it seems.

sqlglot/generator.py
```diff
@@ -2345,6 +2346,18 @@ def ordered_sql(self, expression: exp.Ordered) -> str:
                     f"'{nulls_sort_change.strip()}' translation not supported in window functions"
                 )
                 nulls_sort_change = ""
+            elif (
+                self.NULL_ORDERING_SUPPORTED is False
+                and (isinstance(expression.find_ancestor(exp.AggFunc, exp.Select), exp.AggFunc))
+                and (
+                    (asc and nulls_sort_change == " NULLS LAST")
+                    or (desc and nulls_sort_change == " NULLS FIRST")
```
georgesittas, line 2354: What is this part checking for? What's the reasoning behind checking `asc` and `nulls_sort_change`?

georgesittas (approved):
LGTM, one question to really understand this.

sqlglot/generator.py
VaggelisD: Aggregate functions actually allow `ASC NULLS FIRST` and `DESC NULLS LAST` (default behavior of `ASC` and `DESC` respectively) but not the other way around

georgesittas: > which is the default behavior of `ASC` and `DESC` respectively

It looks like there may be a hidden assumption here. This isn't the case in general, which is why we have the `NULL_ORDERING` dialect constant. Did you take this into account? The base generator code should be generic, not just BQ-specific (if that was the rationale).

georgesittas: Gotcha, yeah that makes sense. I was lacking the surrounding context in code. Sounds good, feel free to merge when ready 👍

PR #4173, opened 2024-09-27
fix(redshift): Add unsupported warnings for UNNEST

### LINKED ISSUE #4169: incorrect transpilation of unnest for redshift
running:
`sqlglot.transpile("select unnest({\"a\": 123}) as x;", read="duckdb", write="redshift")[0]`

actual:
`'SELECT UNNEST(STRUCT(123 AS a)) AS x'`
however redshift doesn't support unnest. there is no perfect equivalent in redshift so i'm curious if this is a situation where sqlglot should do the full surgery into multiple stages of queries or just throw its hands up.

the equivalent output to tsql does the right thing.
` sqlglot.transpile("select unnest({\"a\": 123}) as x;", read="duckdb", write="tsql")[0]`
returns:
`'SELECT EXPLODE(STRUCT(123 AS a)) AS x'`

VaggelisD:
Fixes #4169

Redshift supports [implicit unnest of an array](https://stackoverflow.com/a/72870096) but not an `UNNEST` function as other dialects. This PR adds unsupported warnings for the following cases:
- If an `exp.Unnest` is used with a _known_ non-array type
- If an `exp.Unnest` is not used under a `FROM` or `JOIN` clause
- If we attempt to transpile an `exp.Explode` node coming from Spark/DuckDB etc.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -184,6 +184,7 @@ class Redshift(Postgres):
             exp.DateDiff: date_delta_sql("DATEDIFF"),
             exp.DistKeyProperty: lambda self, e: self.func("DISTKEY", e.this),
             exp.DistStyleProperty: lambda self, e: self.naked_property(e),
+            exp.Explode: lambda self, e: self.explode_sql(e),
             exp.FromBase: rename_func("STRTOL"),
             exp.GeneratedAsIdentityColumnConstraint: generatedasidentitycolumnconstraint_sql,
             exp.JSONExtract: json_extract_segments("JSON_EXTRACT_PATH_TEXT"),
@@ -388,13 +389,28 @@ class Redshift(Postgres):
             args = expression.expressions
             num_args = len(args)
 
-            if num_args > 1:
+            if not num_args == 1:
                 self.unsupported(f"Unsupported number of arguments in UNNEST: {num_args}")
                 return ""
 
-            arg = self.sql(seq_get(args, 0))
+            if not expression.find_ancestor(exp.From, exp.Join):
+                self.unsupported("Unsupported UNNEST when not used in FROM/JOIN clauses")
+                return ""
+
+            from sqlglot.optimizer.annotate_types import annotate_types
+
+            arg = t.cast(exp.Expression, seq_get(args, 0))
+            annotated_arg = annotate_types(arg)
+            if not annotated_arg.is_type(exp.DataType.Type.UNKNOWN) and not annotated_arg.is_type(
+                exp.DataType.Type.ARRAY
+            ):
+                self.unsupported("Unsupported UNNEST without ARRAY argument")
+                return ""
+
+            arg_sql = self.sql(arg)
+
             alias = self.expressions(expression.args.get("alias"), key="columns", flat=True)
-            return f"{arg} AS {alias}" if alias else arg
+            return f"{arg_sql} AS {alias}" if alias else arg_sql
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
             if expression.is_type(exp.DataType.Type.JSON):
@@ -434,3 +450,7 @@ class Redshift(Postgres):
                 return super().array_sql(expression)
 
             return rename_func("ARRAY")(self, expression)
+
+        def explode_sql(self, expression: exp.Explode) -> str:
+            self.unsupported("Unsupported EXPLODE() function")
+            return ""
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -389,28 +389,18 @@ class Redshift(Postgres):
             args = expression.expressions
             num_args = len(args)
 
-            if not num_args == 1:
+            if num_args != 1:
                 self.unsupported(f"Unsupported number of arguments in UNNEST: {num_args}")
                 return ""
 
-            if not expression.find_ancestor(exp.From, exp.Join):
+            if isinstance(expression.find_ancestor(exp.From, exp.Join, exp.Select), exp.Select):
                 self.unsupported("Unsupported UNNEST when not used in FROM/JOIN clauses")
                 return ""
 
-            from sqlglot.optimizer.annotate_types import annotate_types
-
-            arg = t.cast(exp.Expression, seq_get(args, 0))
-            annotated_arg = annotate_types(arg)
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -388,13 +389,28 @@ def unnest_sql(self, expression: exp.Unnest) -> str:
             args = expression.expressions
             num_args = len(args)
 
-            if num_args > 1:
+            if not num_args == 1:
```
georgesittas, line 392:
```suggestion
            if num_args != 1:
```

sqlglot/dialects/redshift.py
```diff
@@ -388,13 +389,28 @@ def unnest_sql(self, expression: exp.Unnest) -> str:
...
                 self.unsupported(f"Unsupported number of arguments in UNNEST: {num_args}")
                 return ""
 
-            arg = self.sql(seq_get(args, 0))
+            if not expression.find_ancestor(exp.From, exp.Join):
+                self.unsupported("Unsupported UNNEST when not used in FROM/JOIN clauses")
+                return ""
```
georgesittas, line 398: This is gonna traverse all the way up, so if you have a derived table you may hit the parent scope's `From` or `Join`. I think we want to do include `Select` and then check the ancestor is not an `Select`.

sqlglot/dialects/redshift.py
```diff
@@ -388,13 +389,28 @@ def unnest_sql(self, expression: exp.Unnest) -> str:
...
+
+            from sqlglot.optimizer.annotate_types import annotate_types
+
+            arg = t.cast(exp.Expression, seq_get(args, 0))
+            annotated_arg = annotate_types(arg)
+            if not annotated_arg.is_type(exp.DataType.Type.UNKNOWN) and not annotated_arg.is_type(
+                exp.DataType.Type.ARRAY
+            ):
+                self.unsupported("Unsupported UNNEST without ARRAY argument")
+                return ""
```
georgesittas, line 408: I'm not sure about this bit, what triggered the need for this code path? Doesn't redshift have a `SUPER` type that can encode arrays too? Should we leave this out?

VaggelisD: I kept thinking that e.g. in DuckDB we can do `SELECT * FROM UNNEST(<struct expr>)` which will bypass the other checks (as the position of the `UNNEST` is valid in Redshift) but will not work properly due to the non-array expr.

However, the `SUPER` data type does make this more complicated than is needed, so will revert this change

sqlglot/dialects/redshift.py
```diff
@@ -434,3 +440,7 @@ def array_sql(self, expression: exp.Array) -> str:
                 return super().array_sql(expression)
 
             return rename_func("ARRAY")(self, expression)
+
+        def explode_sql(self, expression: exp.Explode) -> str:
+            self.unsupported("Unsupported EXPLODE() function")
+            return ""
```
georgesittas, line 446: I'm not sure about this, e.g. you can have `Explode` under `FROM` and `JOIN` in Spark. Wouldn't we be able to transpile these to Redshift? What do we do today?

VaggelisD: We don't do anything, it will get transpiled as is to Redshift. Maybe we can point Redshift's `exp.Explode` to it's `unnest_sql()`?

georgesittas: Sounds reasonable 👍

VaggelisD: After some thought I remembered why I thought I'd put an unsupported warning there, the "implicit unnest" syntax in Redshift is the following:

```SQL
with t as (select array(1, 2) as foo) select foo2 from t as tbl, tbl.foo as foo2

1
2
```

As you said, Spark's `EXPLODE()` can also be used in FROM, e.g:

```
spark-sql (default)> select * from explode(array(1, 2)); 
column
1
2
```

Or as a lateral view:
```
spark-sql (default)> with t as (select array(1) as foo) select * from t lateral view explode(foo) as col;
foo     col
[1]     1
```

However, by transforming `exp.Explode` to an `exp.Unnest` @ generation time we end up with the "implicit unnest" syntax (i.e just dropping the surrounding `UNNEST(...)`), which is not actually unnesting the array:

```
select * from array(1, 2)
[1, 2]
```

So, the semantics do not seem to match well here for the transformation to make much sense. Wdyt?

georgesittas: This doesn't look that much different from what we do for `Unnest` today, though:

```
(.venv) ➜  sqlmesh git:(main) duckdb
D SELECT * FROM UNNEST([1, 2, 3]);
┌──────────────────────────┐
│ main.list_value(1, 2, 3) │
│          int32           │
├──────────────────────────┤
│                        1 │
│                        2 │
│                        3 │
└──────────────────────────┘
D

(.venv) ➜  sqlmesh git:(main) python
Python 3.12.5 (main, Aug  6 2024, 19:08:49) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> import sqlglot
>>> sqlglot.transpile("SELECT * FROM UNNEST([1, 2, 3]);", "duckdb", "redshift")
['SELECT * FROM ARRAY(1, 2, 3)']
```

So, generating `Explode` using `unnest_sql` should _at least_ work for the same cases as `Unnest` does, right? For example what if you had lateral / cross joins, are those transpiled correctly?

VaggelisD: I guess the main case we aim to solve for is something like:

```
with t as (select [1, 2, 3] as foo) select * from t, UNNEST(t.foo);
┌───────────┬────────┐
│    foo    │ unnest │
│  int32[]  │ int32  │
├───────────┼────────┤
│ [1, 2, 3] │      1 │
│ [1, 2, 3] │      2 │
│ [1, 2, 3] │      3 │
└───────────┴────────┘
```

This currently is transpiled to the invalid Redshift:
```
WITH t AS (SELECT ARRAY(1, 2, 3) AS foo) SELECT * FROM t, t.foo
```


To make it work, the transformation requires a couple tweaks regarding the aliases which are not done yet:

```
WITH t AS (SELECT ARRAY(1, 2, 3) AS foo) SELECT * FROM t as tbl, tbl.foo as foo2;

[1, 2, 3]    1
[1, 2, 3]    2
[1, 2, 3]    3
```

I'm not sure if there's a valid way to transpile `SELECT * FROM UNNEST/EXPLODE`, maybe we'd have to generate our own aliases and change the projection list (?)

georgesittas: Ok sg, no worries.

PR #4178, opened 2024-09-30
Fix: Unexpected row deduplication using eliminate_full_outer_join

liaco:
The union syntax cause deduplication of rows from source tables, to achieve the intended result but avoid unintended deduplication I changed the logic to a left join + an anti join(right join + not exists) with union all.

This should be more in line with the behaviour of FULL OUTER JOINs since it keeps duplicate rows that would be present if there are duplicate rows in the two joined tables.
https://en.wikipedia.org/wiki/Join_%28SQL%29#Full_outer_join

Other possible ways to achieve this are:
- filtering for NULLS on joining condition columns in the right side of the union
- use a pure anti-join without the right join syntax (this would require table identifiers on columns selected) 
- use inner joins + more anti-joins (same requirements as the previous one)

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -624,29 +624,35 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             if join.side == "FULL"
         ]
 
-        
-
         if len(full_outer_joins) == 1:
-
             expression_copy = expression.copy()
             expression.set("limit", None)
             index, full_outer_join = full_outer_joins[0]
 
-            tables = (full_outer_join.find(exp.Table), expression.args.get("from").this)
-            join_conditions = full_outer_join.args.get("on") or exp.And(expressions=[
-                exp.EQ(this=exp.Column(this=col, table=tables[0]), expression=exp.Column(this=col, table=tables[1]))
-                for col in full_outer_join.args.get("using")
-            ])
+            tables = (
+                full_outer_join.find(exp.Table).alias_or_name,
+                expression.args["from"].alias_or_name,
+            )
+            join_conditions = full_outer_join.args.get("on") or exp.And(
+                expressions=[
+                    exp.EQ(
+                        this=exp.Column(this=col, table=tables[0]),
+                        expression=exp.Column(this=col, table=tables[1]),
+                    )
+                    for col in full_outer_join.args.get("using")
+                ]
+            )
 
             full_outer_join.set("side", "left")
             anti_join_clause = (
                 exp.Select(expressions=[exp.Literal(this=1, is_string=False)])
-                .from_(expression.args.get("from"))
+                .from_(expression.args["from"])
                 .where(join_conditions)
-                )
+            )
             expression_copy.args["joins"][index].set("side", "right")
             expression_copy.args["where"] = exp.Where(this=exp.Exists(this=anti_join_clause).not_())
             expression_copy.args.pop("with", None)  # remove CTEs from RIGHT side
+            expression.args.pop("order", None)  # remove order by from RIGHT side
 
             return exp.union(expression, expression_copy, copy=False, distinct=False)
 
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -629,10 +629,7 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             expression.set("limit", None)
             index, full_outer_join = full_outer_joins[0]
 
-            tables = (
-                full_outer_join.alias_or_name,
-                expression.args["from"].alias_or_name,
-            )
+            tables = (expression.args["from"].alias_or_name, full_outer_join.alias_or_name)
             join_conditions = full_outer_join.args.get("on") or exp.and_(
                 *[
                     exp.column(col, tables[0]).eq(exp.column(col, tables[1]))
@@ -647,7 +644,7 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
                 .where(join_conditions)
             )
             expression_copy.args["joins"][index].set("side", "right")
-            expression_copy.args["where"] = exp.Where(this=exp.Exists(this=anti_join_clause).not_())
+            expression_copy = expression_copy.where(exp.Exists(this=anti_join_clause).not_())
             expression_copy.args.pop("with", None)  # remove CTEs from RIGHT side
             expression.args.pop("order", None)  # remove order by from LEFT side
 
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -652,7 +652,7 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             expression_copy.args["joins"][index].set("side", "right")
             expression_copy.args["where"] = exp.Where(this=exp.Exists(this=anti_join_clause).not_())
             expression_copy.args.pop("with", None)  # remove CTEs from RIGHT side
-            expression.args.pop("order", None)  # remove order by from RIGHT side
+            expression.args.pop("order", None)  # remove order by from LEFT side
 
             return exp.union(expression, expression_copy, copy=False, distinct=False)
 
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -624,15 +624,31 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             if join.side == "FULL"
         ]
 
+        
+
         if len(full_outer_joins) == 1:
+
             expression_copy = expression.copy()
             expression.set("limit", None)
             index, full_outer_join = full_outer_joins[0]
+
+            tables = (full_outer_join.find(exp.Table), expression.args.get("from").this)
+            join_conditions = full_outer_join.args.get("on") or exp.And(expressions=[
+                exp.EQ(this=exp.Column(this=col, table=tables[0]), expression=exp.Column(this=col, table=tables[1]))
+                for col in full_outer_join.args.get("using")
+            ])
+
             full_outer_join.set("side", "left")
+            anti_join_clause = (
+                exp.Select(expressions=[exp.Literal(this=1, is_string=False)])
+                .from_(expression.args.get("from"))
+                .where(join_conditions)
+                )
             expression_copy.args["joins"][index].set("side", "right")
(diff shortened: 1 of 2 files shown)


VaggelisD:
Hey @liaco, good catch and thank you for the PR.

I think the 1st bullet point solution would be a bit easier to comprehend (and possibly faster), I presume it'd look like [this](https://stackoverflow.com/a/30497590). What do you think?

sqlglot/transforms.py
```diff
@@ -628,11 +628,33 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             expression_copy = expression.copy()
             expression.set("limit", None)
             index, full_outer_join = full_outer_joins[0]
+
+            tables = (
+                full_outer_join.find(exp.Table).alias_or_name,
```
georgesittas, line 633: Can we omit the `find` call here?

sqlglot/transforms.py
```diff
@@ -628,11 +628,33 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
...
+            )
+            join_conditions = full_outer_join.args.get("on") or exp.And(
+                expressions=[
+                    exp.EQ(
+                        this=exp.Column(this=col, table=tables[0]),
+                        expression=exp.Column(this=col, table=tables[1]),
+                    )
+                    for col in full_outer_join.args.get("using")
+                ]
+            )
```
georgesittas, line 644: This AST is incorrect, there's no `expressions` child in `And`. You could use the `and_` helper instead:

```python
and_(
    *[
        exp.column(col, tables[0].eq(exp.column(col, tables[1]))
        for col in full_outer_join.args.get("using")
    ]
)
```

georgesittas (commented):
Left a couple of comments. I'd prefer a simpler way to implement this.

liaco:
> I think the 1st bullet point solution would be a bit easier to comprehend (and possibly faster), I presume it'd look like [this](https://stackoverflow.com/a/30497590). What do you think?

I agree, I am not sure however if the code is simpler, it would require something like this to identify cols from the left side table
```
join_cols = [
    col for col in full_outer_join.args.get("on", exp.Expression()).find_all(exp.Column) 
    if col.table==tables[1]
    ] or [exp.column(identifier, tables[1]) for identifier in full_outer_join.args.get("using")]
```

and then
```
.where(exp.Is(this=join_cols[0], expression=sqlglot.exp.null()))
```
Let me know what you think

VaggelisD:
@liaco I see, your solution is fine, we'll keep iterating on that.

liaco:
@georgesittas I implemented your feedback, I don't think is possible to simplify it more due to the nature of full joins: to emulate the functionality we need to use unions or interceptions of available joins (inner, left) to achieve a partial result and then use anti joins to fill in the gaps, however to use anti-joins we need necessarily to identify joining conditions first.

One other way is building a CTE to select from with the union of common joining identifiers and simply left join the two tables (something like [this](https://stackoverflow.com/a/29674087), but it involves more complexity.

Let me know what you think

sqlglot/transforms.py
```diff
@@ -628,11 +628,30 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
...
             full_outer_join.set("side", "left")
+            anti_join_clause = (
+                exp.Select(expressions=[exp.Literal(this=1, is_string=False)])
+                .from_(expression.args["from"])
+                .where(join_conditions)
+            )
             expression_copy.args["joins"][index].set("side", "right")
+            expression_copy.args["where"] = exp.Where(this=exp.Exists(this=anti_join_clause).not_())
             expression_copy.args.pop("with", None)  # remove CTEs from RIGHT side
+            expression.args.pop("order", None)  # remove order by from LEFT side
```
VaggelisD, line 653: Can we test this too in the existing unit test, as is done for `LIMIT`?

sqlglot/transforms.py
```diff
@@ -628,11 +628,30 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
...
             index, full_outer_join = full_outer_joins[0]
+
+            tables = (
+                full_outer_join.alias_or_name,
+                expression.args["from"].alias_or_name,
+            )
+            join_conditions = full_outer_join.args.get("on") or exp.and_(
+                *[
+                    exp.column(col, tables[0]).eq(exp.column(col, tables[1]))
+                    for col in full_outer_join.args.get("using")
```
VaggelisD, line 639: Do we have a test case for `USING`?

georgesittas: If there's an `ON` condition do we still need to add this?

sqlglot/transforms.py
```diff
@@ -628,11 +628,30 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
...
+            )
+
             full_outer_join.set("side", "left")
+            anti_join_clause = (
+                exp.Select(expressions=[exp.Literal(this=1, is_string=False)])
+                .from_(expression.args["from"])
+                .where(join_conditions)
+            )
             expression_copy.args["joins"][index].set("side", "right")
+            expression_copy.args["where"] = exp.Where(this=exp.Exists(this=anti_join_clause).not_())
```
georgesittas, line 650: What if you already had a `WHERE` clause? This would overwrite it.

sqlglot/transforms.py
liaco: Do you mean from this:
```
        self.validate_all(
            "SELECT * FROM x LEFT JOIN y ON x.id = y.id UNION ALL SELECT * FROM x RIGHT JOIN y ON x.id = y.id WHERE NOT EXISTS(SELECT 1 FROM x WHERE x.id = y.id) LIMIT 0",
            read={
                "postgres": "SELECT * FROM x FULL JOIN y ON x.id = y.id LIMIT 0",
            },
        )
```
to this:
```
        self.validate_all(
            "SELECT * FROM x LEFT JOIN y ON x.id = y.id UNION ALL SELECT * FROM x RIGHT JOIN y ON x.id = y.id WHERE NOT EXISTS(SELECT 1 FROM x WHERE x.id = y.id) ORDER BY id LIMIT 0",
            read={
                "postgres": "SELECT * FROM x FULL JOIN y ON x.id = y.id ORDER BY id LIMIT 0",
            },
        )
```
?

sqlglot/transforms.py
liaco: I'll add it

sqlglot/transforms.py
liaco: Indeed, I'm converting it using the helper `where` since it's append by default

sqlglot/transforms.py
VaggelisD: Yeah reusing the existing test should be fine imo, as long as it's a valid Postgres query and shows that the query modifiers (order, limit etc) are preserved once

sqlglot/transforms.py
liaco: I added a test case for USING since it has its own logic in `eliminate_full_outer_join`, if it's not needed I'll revert it

georgesittas: I meant that if you have an `ON` clause, then you skip the `exp.and_` term. What if that `ON` clause doesn't have the equality predicate? Does this transformation still work then, or do we still need to append an `AND` term to that `ON`?

liaco: So the `exp.and_` is needed only when dealing with `USING` because we need to transform the using condition into an implicit join to feed into the anti-join condition: `NOT EXISTS (SELECT 1 FROM <left_side_table> WHERE <join_condition_here>)`.

If we have an `ON` clause instead, it should be already compatible and ready to be used in later steps, even if it is a pure non equi join, I have attached a real example with this behaviour.
image

Let me know if you think I'm missing something

liaco: So the `exp.and_` is needed only when dealing with `USING` because we need to transform the using condition into an implicit join to feed into the anti-join condition: `NOT EXISTS (SELECT 1 FROM <left_side_table> WHERE <join_condition_here>)`.

If we have an `ON` clause instead, it should be already compatible and ready to be used in later steps, even if it is a pure non equi join, I have attached a real example with this behaviour.
image


Let me know if you think I'm missing something

georgesittas: I see, this makes sense. Thanks for clarifying!

georgesittas (approved):

georgesittas:
Thanks a lot for the PR @liaco, great work!

sqlglot/transforms.py
```diff
@@ -628,11 +628,27 @@ def eliminate_full_outer_join(expression: exp.Expression) -> exp.Expression:
             expression_copy = expression.copy()
             expression.set("limit", None)
             index, full_outer_join = full_outer_joins[0]
+
+            tables = (expression.args["from"].alias_or_name, full_outer_join.alias_or_name)
+            join_conditions = full_outer_join.args.get("on") or exp.and_(
+                *[
+                    exp.column(col, tables[0]).eq(exp.column(col, tables[1]))
+                    for col in full_outer_join.args.get("using")
```
VaggelisD, line 636: Can we get one more test showcasing `USING(a, b) -> WHERE NOT EXISTS (SELECT 1 FROM t1.a = t2.a AND t1.b = t2.b`, so we can cover the `OR` here?

liaco: Sure, I think I can simply add more joining columns in the `USING` clause in [the one I already added] to cover for that (https://github.com/tobymao/sqlglot/pull/4178/commits/0b1b86c48e821ba70a096dff03d828afea559d45#diff-c5dbbef7e9cbb2c64b384079070c3006398036d834774db5983a1e347b7747f8R762-R767)

VaggelisD: Let's duplicate that test case to be sure that both work properly, thank you!

VaggelisD (approved):
One final (minor) comment, thank you for the PR & for the quick iterations.

liaco:
Thank you for your thoughtful reviews, happy to give my small contribute

georgesittas:
Check out https://github.com/tobymao/sqlglot/commit/7ecd51928eedd104aa318ecafc92f92b262e2e51

georgesittas:
> Thank you for your thoughtful reviews, happy to give my small contribute

Appreciate the contribution and the rapid iterations! :-)

PR #4181, opened 2024-10-01
feat(oracle): Add support for BULK COLLECT INTO

mkmoisen:
This commit adds support for the Oracle specific `BULK COLLECT INTO` syntax.

```
SELECT foo, bar
BULK COLLECT INTO l_rows
FROM bar
```

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -251,6 +251,7 @@ class Oracle(Dialect):
         TABLESAMPLE_KEYWORDS = "SAMPLE"
         LAST_DAY_SUPPORTS_DATE_PART = False
         SUPPORTS_SELECT_INTO = True
+        SUPPORTS_SELECT_BULK_COLLECT_INTO = True
         TZ_TO_WITH_TIME_ZONE = True
 
         TYPE_MAPPING = {
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2109,6 +2109,10 @@ class Into(Expression):
     arg_types = {"this": True, "temporary": False, "unlogged": False}
 
 
+class BulkCollectInto(Expression):
+    arg_types = {"this": True}
+
+
 class From(Expression):
     @property
     def name(self) -> str:
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -354,6 +354,9 @@ class Generator(metaclass=_Generator):
     # Whether the SELECT .. INTO syntax is used instead of CTAS
     SUPPORTS_SELECT_INTO = False
 
+    # Whether the SELECT .. BULK COLLECT INTO syntax is supported
+    SUPPORTS_SELECT_BULK_COLLECT_INTO = False
+
     # Whether UNLOGGED tables can be created
     SUPPORTS_UNLOGGED_TABLES = False
 
@@ -2043,6 +2046,9 @@ class Generator(metaclass=_Generator):
         unlogged = " UNLOGGED" if expression.args.get("unlogged") else ""
         return f"{self.seg('INTO')}{temporary or unlogged} {self.sql(expression, 'this')}"
 
+    def bulkcollectinto_sql(self, expression: exp.BulkCollectInto) -> str:
+        return f"{self.seg('BULK COLLECT INTO')} {self.sql(expression, 'this')}"
+
     def from_sql(self, expression: exp.From) -> str:
         return f"{self.seg('FROM')} {self.sql(expression, 'this')}"
 
@@ -2464,6 +2470,10 @@ class Generator(metaclass=_Generator):
         if not self.SUPPORTS_SELECT_INTO and into:
             into.pop()
 
+        bulk_collect_into = expression.args.get("bulk_collect_into")
+        if not self.SUPPORTS_SELECT_BULK_COLLECT_INTO and bulk_collect_into:
+            bulk_collect_into.pop()
+
         hint = self.sql(expression, "hint")
         distinct = self.sql(expression, "distinct")
         distinct = f" {distinct}" if distinct else ""
@@ -2507,6 +2517,7 @@ class Generator(metaclass=_Generator):
             expression,
             f"SELECT{top_distinct}{kind}{expressions}",
             self.sql(expression, "into", comment=False),
+            self.sql(expression, "bulk_collect_into", comment=False),
             self.sql(expression, "from", comment=False),
         )
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2969,6 +2969,10 @@ class Parser(metaclass=_Parser):
             if into:
                 this.set("into", into)
 
+            bulk_collect_into = self._parse_bulk_collect_into()
+            if bulk_collect_into:
+                this.set("bulk_collect_into", bulk_collect_into)
+
             if not from_:
                 from_ = self._parse_from()
 
@@ -3194,6 +3198,12 @@ class Parser(metaclass=_Parser):
         return self.expression(
             exp.Into, this=self._parse_table(schema=True), temporary=temp, unlogged=unlogged
         )
+
+    def _parse_bulk_collect_into(self) -> t.Optional[exp.BulkCollectInto]:
+        if not self._match(TokenType.BULK_COLLECT_INTO):
+            return None
+
+        return self.expression(exp.BulkCollectInto, this=self._parse_table(schema=True))
 
     def _parse_from(
         self, joins: bool = False, skip_from_token: bool = False
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -221,6 +221,7 @@ class TokenType(AutoName):
     AUTO_INCREMENT = auto()
     BEGIN = auto()
     BETWEEN = auto()
+    BULK_COLLECT_INTO = auto()
     CACHE = auto()
     CASE = auto()
     CHARACTER_SET = auto()
@@ -661,6 +662,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "AUTO_INCREMENT": TokenType.AUTO_INCREMENT,
         "BEGIN": TokenType.BEGIN,
         "BETWEEN": TokenType.BETWEEN,
+        "BULK COLLECT INTO": TokenType.BULK_COLLECT_INTO,
         "CACHE": TokenType.CACHE,
         "UNCACHE": TokenType.UNCACHE,
         "CASE": TokenType.CASE,
--- a/tests/dialects/test_oracle.py
+++ b/tests/dialects/test_oracle.py
@@ -67,6 +67,9 @@ class TestOracle(Validator):
             "SELECT COUNT(1) INTO V_Temp FROM TABLE(CAST(somelist AS data_list)) WHERE col LIKE '%contact'"
         )
         self.validate_identity(
+            "SELECT department_id BULK COLLECT INTO v_department_ids FROM departments"
+        )
+        self.validate_identity(
             "SELECT MIN(column_name) KEEP (DENSE_RANK FIRST ORDER BY column_name DESC) FROM table_name"
         )
         self.validate_identity(


tobymao:
can you please link the relevant documentation from oracle?

sqlglot/expressions.py
```diff
@@ -2109,6 +2109,10 @@ class Into(Expression):
     arg_types = {"this": True, "temporary": False, "unlogged": False}
 
 
+class BulkCollectInto(Expression):
+    arg_types = {"this": True}
```
tobymao, line 2113: this is not necessary because it's identical to the base class

sqlglot/tokens.py
```diff
@@ -221,6 +221,7 @@ class TokenType(AutoName):
     AUTO_INCREMENT = auto()
     BEGIN = auto()
     BETWEEN = auto()
+    BULK_COLLECT_INTO = auto()
```
tobymao, line 224: Let's not add a token, instead do match texts, we generally don't add tokens unless we have to.

sqlglot/parser.py
```diff
@@ -3195,6 +3199,12 @@ def _parse_into(self) -> t.Optional[exp.Into]:
             exp.Into, this=self._parse_table(schema=True), temporary=temp, unlogged=unlogged
         )
 
+    def _parse_bulk_collect_into(self) -> t.Optional[exp.BulkCollectInto]:
+        if not self._match(TokenType.BULK_COLLECT_INTO):
```
tobymao, line 3203: change this to match_text_seq

sqlglot/parser.py
```diff
@@ -3195,6 +3199,12 @@ def _parse_into(self) -> t.Optional[exp.Into]:
...
+            return None
+
+        return self.expression(exp.BulkCollectInto, this=self._parse_table(schema=True))
```
tobymao, line 3206: this doesnt' seem correct, it looks it can be a list of stuff right?

sqlglot/parser.py
```diff
@@ -2969,6 +2969,10 @@ def _parse_select(
             if into:
                 this.set("into", into)
 
+            bulk_collect_into = self._parse_bulk_collect_into()
+            if bulk_collect_into:
+                this.set("bulk_collect_into", bulk_collect_into)
```
tobymao, line 2974: you need to to make bulk_collect_into a first class property of select

tobymao: can you have both into and bulk_into? if not, then i'd just use the into statement, perhaps you could just overide the into parser of oracle

mkmoisen:
@tobymao 

> can you please link the relevant documentation from oracle?

https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/SELECT-INTO-statement.html

https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/RETURNING-INTO-clause.html#GUID-38F735B9-1100-45AF-AE71-18FB74A899BE__CJAIAGHJ

Should this be added to a comment in the code anywhere?

sqlglot/generator.py
```diff
@@ -2464,6 +2470,10 @@ def select_sql(self, expression: exp.Select) -> str:
         if not self.SUPPORTS_SELECT_INTO and into:
             into.pop()
 
+        bulk_collect_into = expression.args.get("bulk_collect_into")
+        if not self.SUPPORTS_SELECT_BULK_COLLECT_INTO and bulk_collect_into:
```
tobymao, line 2474: let's delete this section for now, it's not necessary

sqlglot/generator.py
```diff
@@ -2507,6 +2517,7 @@ def select_sql(self, expression: exp.Select) -> str:
             expression,
             f"SELECT{top_distinct}{kind}{expressions}",
             self.sql(expression, "into", comment=False),
+            self.sql(expression, "bulk_collect_into", comment=False),
```
tobymao, line 2520: this may not be necessary if you add it as a special type of into

tobymao:
> @tobymao
> 
> > can you please link the relevant documentation from oracle?
> 
> https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/SELECT-INTO-statement.html
> 
> https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/RETURNING-INTO-clause.html#GUID-38F735B9-1100-45AF-AE71-18FB74A899BE__CJAIAGHJ
> 
> Should this be added to a comment in the code anywhere?

yea, you can perhaps where you parse bulk into

tobymao (changes requested):
thanks for the first PR. i've done an initial pass

sqlglot/parser.py
mkmoisen: @tobymao No, you can only have INTO or BULK COLLECT INTO, not both.

However, from an AST diff perspective, I think it would be useful to be able to make a distinction between these two, as they are semantically different.

sqlglot/parser.py
mkmoisen: @tobymao I'm pretty sure it is always into a single variable, not a list. I will confirm tomorrow.

sqlglot/expressions.py
```diff
@@ -2106,7 +2106,7 @@ class PrimaryKey(Expression):
 # https://www.postgresql.org/docs/9.1/sql-selectinto.html
 # https://docs.aws.amazon.com/redshift/latest/dg/r_SELECT_INTO.html#r_SELECT_INTO-examples
 class Into(Expression):
-    arg_types = {"this": True, "temporary": False, "unlogged": False}
+    arg_types = {"this": True, "temporary": False, "unlogged": False, "bulk_collect_into": False}
```
georgesittas, line 2109: [Nit] I'd rename this to just `bulk_collect` since "into" is encode din the expression's name too.

georgesittas (commented):
Toby's suggestion makes sense to me, given there are aliasing issues. Other than that, LGTM 👍

mkmoisen:
Hi @tobymao @georgesittas 

Please review my latest commit when you have a moment.

To make this BULK COLLECT INTO only a Oracle only Keyword, I did the following:

In dialects/oracle.py:Oracle:Tokenizer:Keywords I added: 

    "BULK COLLECT INTO": TokenType.BULK_COLLECT_INTO

However in `tokens.py:TokenType` I added:

    `BULK_COLLECT_INTO = auto()`
    
I'm not clear if you wanted this TokenType to be in Oracle specifically. I could not find any subclasses of TokenType in the project, so I assume this is ok. Please let me know if you want something else here.

sqlglot/expressions.py
mkmoisen: Done.

georgesittas (approved):

georgesittas:
Actually, one final point, Toby [mentioned](https://github.com/tobymao/sqlglot/pull/4181#discussion_r1782121472) this as well and I'd like to double check it: Oracle's [docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/RETURNING-INTO-clause.html#GUID-38F735B9-1100-45AF-AE71-18FB74A899BE__CJAIAGHJ) show that you can have a csv after the `BULK COLLECT INTO` sequence. That means multiple values can be accepted. Did you test this?

mkmoisen:
> Actually, one final point, Toby [mentioned](https://github.com/tobymao/sqlglot/pull/4181#discussion_r1782121472) this as well and I'd like to double check it: Oracle's [docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/RETURNING-INTO-clause.html#GUID-38F735B9-1100-45AF-AE71-18FB74A899BE__CJAIAGHJ) show that you can have a csv after the `BULK COLLECT INTO` sequence. That means multiple values can be accepted. Did you test this?

@georgesittas I've been working with Oracle for 15 years and never knew BULK COLLECT could do this :).

However, this is also a problem with `INTO` for both Postgres and Oracle. Today sqlglot won't allow you to parse a csv and will only allow one value.

I think the core issue here is that `Into` is overloaded in Postgresql, with two cases, and sqlglot is currently only handling one case.

Case one, which sqlglot supports, is the Postgres CTAS alternative [here](https://www.postgresql.org/docs/current/sql-selectinto.html). This case is not available in Oracle.

Case two, which sqlglot does not support, is populating variables in PL/pgSQL (Postgresql) or in PL/SQL (Oracle)

----

My latest commit attempts to partially reconcile this. 

I have only enabled support for INTO with multiple values, as well as BULK COLLECT INTO with one or more values, in the Oracle dialect. 

Everything is working except for the case of transpiling Oracle into another language. It seems to me that attempts to do this should raise an error.

mkmoisen:
Initially, I attempted to have the `parser.py::Parser::_parse_into` handle both cases.

However, it does make the code harder to understand. I figured isolating INTO and BULK COLLECT INTO into the Oracle dialect would simplify the understanding.

In addition, I couldn't figure out how to both parse for the table (self._parse_table) as well as parse the CSV (self._parse_expressions). It kept having collisions.

sqlglot/dialects/oracle.py
```diff
@@ -344,3 +356,7 @@ def queryoption_sql(self, expression: exp.QueryOption) -> str:
         def coalesce_sql(self, expression: exp.Coalesce) -> str:
             func_name = "NVL" if expression.args.get("is_nvl") else "COALESCE"
             return rename_func(func_name)(self, expression)
+
+        def into_sql(self, expression: exp.Into) -> str:
+            into = "INTO" if not expression.args.get("bulk_collect") else "BULK COLLECT INTO"
```
georgesittas, line 361: What about `TEMPORARY` and `UNLOGGED`? It seems like we're generating them in the base method. Are these needed for Oracle?

georgesittas (commented):
> Case two, which sqlglot does not support, is populating variables in PL/pgSQL (Postgresql) or in PL/SQL (Oracle)

That's fine, procedural SQL code (e.g. variable population as you mention) isn't prioritized in SQLGlot anyway.

> Everything is working except for the case of transpiling Oracle into another language. It seems to me that attempts to do this should raise an error.

Perhaps a workaround is to preserve the previous implementation, i.e. set `this` instead of `expressions`, if there is only one expression.

sqlglot/dialects/oracle.py
mkmoisen: No, `TEMPORARY` and `UNLOGGED` are not used in Oracle's INTO statement.

These are only used in the CTAS case in postgresql.

mkmoisen: > Perhaps a workaround is to preserve the previous implementation, i.e. set this instead of expressions, if there is only one expression.

@georgesittas Would you please give a suggestion on how to accomplish this?

I tried but was unable to figure it out.

```
expressions = self._parse_expressions()
if len(expressions) == 1:
    ...
```

I think i would need to somehow rewind and then call:

    self._parse_table(schema=True)

georgesittas: You can check if `bulk_collect_into` is truthy, in which case you'll set `expressions` to `_parse_csv(self._parse_bitwise)` (*), otherwise just set `this` to `self._parse_table(schema=True)`.

(*): `_parse_expressions` consumes more things and is probably overkill here.

mkmoisen: Hi @georgesittas ,

For Oracle, INTO and BULK COLLECT INTO can both support multiple values in a csv.

This latest commit seems to solve the issue. If only one expression is found, then I call self._advance(-1), and then revert to the previous implementation.

mkmoisen:
Hi @georgesittas 

Please see my recent commit. I think it is mostly there. If only one value is provided for INTO or BULK COLLECT INTO, it will revent to the previous implemenation. This way, the statement can be transpiled into mysql CTAS, or postgres/tsql SELECT INTO, so long as it only has one value.

However, the case where the user has more than one value is a bit more complicated.

It wouldn't make sense to transpile this oracle statement into mysql CTAS:

    select foo, bar into l_foo, l_bar from baz

I feel like it should raise an error.

----

When it comes to postgresql, it's a bit more sketchy because of the two cases I mentioned above.

This is a totally valid Oracle statement that could be transpiled into postgresql for plpgsql:

    select foo, bar into l_foo, l_bar from baz
    
Yet this is totally invalid to transpile it to postgreql's CTAS alternative SELECT INTO.

mkmoisen:
I see two approaches:

## Assume Postgresql's INTO is the CTAS alternative case, not the Pl/pgSQL case

Here, an oracle INTO with one value can be converted into postgresq/mysql/tsl. (Although I feel like this is semantically incorrect, since Oracle's INTO is never a CTAS).

But an oracle INTO with more than one value should raise an exception if attempting to transpile it to postgres, mysql, tsql, etc.

## Add support for multiple values in Postgresql INTO to handle the pl/pgSQL case

If a postgresql INTO uses more than one variable, it should raise an exception if attempting to transpile to mysql, tsql, etc. However it could be transpiled to Oracle.

Oracle INTO's that use more than one variable can be transpiled into postgresql but nothing else. 



As an aside, it seems to me that any Oracle INTO statement should not be allowed to be transpiled to any dialect other than postgresql. Semantically it doesn't make sense, as Oracle INTO is not a CTAS alterantive. It's only a CTAS alternative in postgres/tsql.

mkmoisen:
As an aside, it seems to me that any Oracle INTO statement should not be allowed to be transpiled to any dialect other than postgresql. Semantically it doesn't make sense, as Oracle INTO is not a CTAS alternative. 

 Only in postgres/tsql is INTO an alternative for CTAS.

georgesittas (commented):
I think the `advance(-1)` approach isn't quite right, because you may have multiple tokens so you won't really backtrack completely.

I'll just get this PR to the finish line though, thanks!

georgesittas:
Re: transpilation - I wouldn't worry too much about it now, unless you have a use case of transpiling to postgres / others mentioned. We can improve it incrementally in followup PRs.

georgesittas:
See https://github.com/tobymao/sqlglot/commit/5a123a54ecd033c0a104e33476b17d816a09caac

mkmoisen:
Hi @georgesittas 

Thank you for the fix.

But I think there is still one open issue:
```
>>> import sqlglot

>>> parsed = sqlglot.parse_one('select foo, bar into l_foo, l_bar from baz', dialect='oracle')

>>> print(parsed.sql(dialect='postgres')
SELECT foo, bar INTO  FROM baz

>>> print(parsed.sql('mysql'))
CREATE TABLE  AS SELECT foo, bar FROM baz
```

Essentially, when transpiling an oracle INTO with more than one value to any other dialect, the table name won't be set.

This produces invalid SQL.

It seems to me that this should instead raise some kind of unsupported exception - definitely for mysql/tsql etc. For Postgres you could either do the same, or add support for multiple values in the INTO for postgres.

Thanks!

PR #4182, opened 2024-10-01
feat(optimizer): Fixture file for function annotations

VaggelisD:
This PR introduces the `annotate_functions.sql` fixture file which abstracts away the type inference & check of functions, further supporting dialect-aware annotation & it's testing.

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -57,8 +57,8 @@ def simplify(expression, **kwargs):
 def annotate_functions(expression, **kwargs):
     from sqlglot.dialects import Dialect
 
-    dialect = kwargs["dialect"]
-    schema = kwargs["schema"]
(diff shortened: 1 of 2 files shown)


tests/test_optimizer.py
```diff
@@ -54,6 +54,18 @@ def simplify(expression, **kwargs):
     return optimizer.simplify.simplify(expression, constant_propagation=True, **kwargs)
 
 
+def annotate_functions(expression, **kwargs):
+    from sqlglot.dialects import Dialect
+
+    dialect = kwargs["dialect"]
+    schema = kwargs["schema"]
```
georgesittas, line 61: Let's do `kwargs.get(...)` instead.

georgesittas (approved):

Issue #4184, opened 2024-10-01
`UHUGEINT` is not recognized as a DuckDB Type


rustyconover:
When testing SQLGlot, it appears that `UHUGEINT` is not recognized as a valid type. This causes a `ParseError` to be raised during parsing.

#### Example:

```python
sqlglot.errors.ParseError: Expected type. Line 1, Col: 25.
  SELECT to_hex(?::UHUGEINT) LIMIT 1
```

PR #4190, opened 2024-10-01
fix: add support for UHUGEINT for duckdb

rustyconover:
Address issue #4184


VaggelisD (approved):
Thank you for the PRs! Can you self-assign the ones that you'll attempt to fix so we can take a look at the other ones?

VaggelisD:
Ah I see, maybe a comment under each issue that you'll tackle then? Anything that can let us know to not pick these up so we can avoid overlapping work.

Thanks!

georgesittas (approved):

Issue #4185, opened 2024-10-01
 Incorrect `SHA` Function Mapping for DuckDB Dialect in SQLGlot


rustyconover:
When testing SQLGlot with DuckDB, the `SHA` function is incorrectly translated. The expected function name in DuckDB is `SHA1`, but it gets mapped to `SHA` instead.

#### Original Query:

```sql
SELECT sha1(?::VARCHAR) LIMIT 1
```

#### SQLGlot Output (DuckDB Dialect):

```sql
SELECT SHA(CAST(? AS TEXT)) LIMIT 1
```

#### Expected Behavior:
The function should be translated as `SHA1` in DuckDB, matching the original query.

georgesittas:
Fixed by #4191 -- thanks!

PR #4191, opened 2024-10-01
fix: rename SHA function to SHA1 for DuckDB

rustyconover:
Addresses issue #4185


VaggelisD (commented):
Let's add a test case for this, you can either add it to existing multi-dialect `SHA` tests (as in `test_bigquery.py`) or add something like this in `test_duckdb.py`:

```
        self.validate_all(
            "SHA1(x)",
           read={
              "": "SHA(x)",
            },
            write={
                "duckdb": "SHA1(x)",
                "": "SHA(x)",
            },
        )
```

VaggelisD (approved):

georgesittas (approved):

PR #4194, opened 2024-10-01
feat(duckdb): Parse ~~~ as GLOB

### LINKED ISSUE #4187: Incorrect Handling of the `~~~` Operator in DuckDB
When testing SQLGlot with DuckDB, I discovered that the `~~~` operator is incorrectly handled. This operator is equivalent to the `GLOB` keyword in DuckDB, but SQLGlot incorrectly translates it.

#### Original Query:

```sql
SELECT ?::VARCHAR ~~~ ?::VARCHAR LIMIT 1
```

#### SQLGlot Output:

```sql
SELECT REGEXP_MATCHES(CAST(? AS TEXT), ~~CAST(? AS TEXT)) LIMIT 1
```

#### Expected Behavior:
SQLGlot should produce either the `~~~` operator or the `GLOB` function, as defined by DuckDB for pattern matching. The correct translation would be:

```sql
SELECT ?::VARCHAR ~~~ ?::VARCHAR LIMIT 1
```

#### Reference:
The `~~~` operator is documented to function the same as `GLOB` in DuckDB:
- [DuckDB Pattern Matching Documentation](https://duckdb.org/docs/sql/functions/pattern_matching.html)

VaggelisD:
Fixes #4187

The `a GLOB b` operator is an alias of `a ~~~ b`:

```SQL
D SELECT 'best.txt' GLOB '*.txt';
┌──────────────────────────┐
│ ('best.txt' ~~~ '*.txt') │
├──────────────────────────┤
│ true                     │
└──────────────────────────┘
```

Docs
--------
[DuckDB GLOB](https://duckdb.org/docs/sql/functions/pattern_matching.html#globbing)

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -649,6 +649,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "<->": TokenType.LR_ARROW,
         "&&": TokenType.DAMP,
         "??": TokenType.DQMARK,
+        "~~~": TokenType.GLOB,
         "ALL": TokenType.ALL,
         "ALWAYS": TokenType.ALWAYS,
         "AND": TokenType.AND,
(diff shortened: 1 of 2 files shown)


georgesittas (commented):
May be worth adding to the base parser; I doubt there'd be overlap with other dialects, perhaps Postgres? LGTM otherwise.

georgesittas (approved):

Issue #4196, opened 2024-10-01
Incorrect StrPosition Code Generation for SQLite


pruzko:
Hi,

Sqlglot parses string position functions into `exp.StrPosition` expressions that can be compiled to target dialects via `sql()`. I believe that both parsing and generation does not work for SQLite. The code assumes SQLite's string position function is `STR_POSITION(...)`, but in reality it is `INSTR(...)`. Is this the expected behavior?

MVE:
```python
import sqlglot

x = parse_one("select instr('asd', 's')", read='sqlite')
print(repr(x)) # parsed as Anonymous

x = parse_one("select str_position('asd', 's')", read='sqlite')
print(repr(x)) # parsed as StrPosition

x = parse_one("select instr('asd', 's')", read='mysql')
print(repr(x)) # parsed as StrPosition

print(x.sql(dialect='sqlite')) # compiled as STR_POSITION
```

Fix:
Adding `INSTR` to SQLite's [function list](https://github.com/tobymao/sqlglot/blob/9eba00dca517efe7df171b09ed916af3ea5e350d/sqlglot/dialects/sqlite.py#L113) just like it is done for [MySQL](https://github.com/tobymao/sqlglot/blob/9eba00dca517efe7df171b09ed916af3ea5e350d/sqlglot/dialects/mysql.py#L303C14-L303C19) should do the trick.

georgesittas:
This transpilation path hasn't been covered yet, so I'd say it's "expected" as in "missing functionality". We welcome PRs if you wanna take a stab at it. Your suggestion makes sense 👍

pruzko:
I made a [PR](https://github.com/tobymao/sqlglot/pull/4198).

Also, I'm new to the library and it's absolutely amazing.

PR #4198, opened 2024-10-01
StrPosition compilation path for SQLite

pruzko:
See the [corresponding issue](https://github.com/tobymao/sqlglot/issues/4196).

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1250,6 +1250,12 @@ def strposition_to_locate_sql(self: Generator, expression: exp.StrPosition) -> s
     )
 
 
+def strposition_to_instr_sql(self: Generator, expression: exp.StrPosition) -> str:
+    return self.func(
+        "INSTR", expression.args.get("substr"), expression.this, expression.args.get("position")
+    )
+
+
 def left_to_substring_sql(self: Generator, expression: exp.Left) -> str:
     return self.sql(
         exp.Substring(
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1052,7 +1052,10 @@ def property_sql(self: Generator, expression: exp.Property) -> str:
 
 
 def str_position_sql(
-    self: Generator, expression: exp.StrPosition, generate_instance: bool = False
+    self: Generator,
+    expression: exp.StrPosition,
+    generate_instance: bool = False,
+    str_position_func_name: str = "STRPOS",
 ) -> str:
     this = self.sql(expression, "this")
     substr = self.sql(expression, "substr")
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -1250,6 +1250,12 @@ def strposition_to_locate_sql(self: Generator, expression: exp.StrPosition) -> s
     )
 
 
+def strposition_to_instr_sql(self: Generator, expression: exp.StrPosition) -> str:
+    return self.func(
+        "INSTR", expression.args.get("substr"), expression.this, expression.args.get("position")
+    )
```
georgesittas, line 1256: Let's not implement this logic here. Instead, define `strposition_sql(self, expression: StrPosition) -> str:` within `SQLite.Generator`.

Additionally, isn't this supposed to be `INSTR(haystack, needle)`? You're generating three arguments here. Also, the relative order looks incorrect: `substr` comes first here, but you need it to be 2nd right?

sqlglot/dialects/sqlite.py
```diff
@@ -116,6 +118,7 @@ class Parser(parser.Parser):
             "STRFTIME": _build_strftime,
             "DATETIME": lambda args: exp.Anonymous(this="DATETIME", expressions=args),
             "TIME": lambda args: exp.Anonymous(this="TIME", expressions=args),
+            "INSTR": lambda args: exp.StrPosition(substr=seq_get(args, 1), this=seq_get(args, 0)),
```
georgesittas, line 121: Move this to the base `Parser`'s `FUNCTIONS` dict and also remove MySQL's entry as well.

sqlglot/dialects/sqlite.py
```diff
@@ -186,6 +189,7 @@ class Generator(generator.Generator):
                     transforms.eliminate_semi_and_anti_joins,
                 ]
             ),
+            exp.StrPosition: strposition_to_instr_sql,
```
georgesittas, line 192: We'll remove this after the suggested refactor is implemented.

tests/test_expressions.py
```diff
@@ -350,6 +350,7 @@ def test_function_building(self):
         )
 
         self.assertIsInstance(exp.func("instr", "x", "b", dialect="mysql"), exp.StrPosition)
+        self.assertIsInstance(exp.func("instr", "x", "b", dialect="sqlite"), exp.StrPosition)
```
georgesittas, line 353: This isn't enough to test this, you need to do `self.validate_identity("INSTR(x, y)")` within `test_sqlite.TestSQLite.test_sqlite`.

georgesittas (changes requested):
Thank you for the PR!

sqlglot/dialects/dialect.py
pruzko: I've removed this altogether, since someone already implemented `dialect->str_position_sql`, which is also compatible with MySQL's `locate`.

And yes, the order was messed up.

sqlglot/dialects/sqlite.py
pruzko: I replaced this with a lambda call to `dialect->str_position_sql`

pruzko:
Also, please how long does it typically take for accepted PRs to be synced with pypi? No pressure, just curious.

georgesittas:
> Also, please how long does it typically take for accepted PRs to be synced with pypi? No pressure, just curious.

We release frequently, most likely sometime later today or tomorrow.

PR #4201, opened 2024-10-02
feat(bigquery): Native math function annotations

VaggelisD:
This is an initial attempt bringing native BigQuery type annotation, starting from the mathematical functions.

This PR focuses on `ABS(x)`, `SIGN(x)`, `LEAST(x1, ..., xn)` and `GREATEST(x1, ..., xn)` which borrow the type of their inputs:

```SQL
select typeof(abs(1)), typeof(abs(1.5)), typeof(sign(1)), typeof(sign(1.5));

INT64 | FLOAT64 | INT64 | FLOAT64
```

As the latter functions are varlen, the result type is the coerced / common supertype according to the docs:

```SQL
select typeof(least(1, 2)), typeof(least(1, 2.5)), typeof(greatest(1, 2)), typeof(greatest(1, 2.5));

INT64 | FLOAT64 | INT64 | FLOAT64
```

Docs
--------
[ABS](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#abs) | [SIGN](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#sign) | [LEAST](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#least) | [GREATEST](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#greatest) |

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -293,6 +293,14 @@ class BigQuery(Dialect):
     # All set operations require either a DISTINCT or ALL specifier
     SET_OP_DISTINCT_BY_DEFAULT = dict.fromkeys((exp.Except, exp.Intersect, exp.Union), None)
 
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        exp.Abs: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Greatest: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
+        exp.Least: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
+    }
+
     def normalize_identifier(self, expression: E) -> E:
         if (
             isinstance(expression, exp.Identifier)
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -577,7 +577,6 @@ class Dialect(metaclass=_Dialect):
         exp.DataType.Type.DOUBLE: {
             exp.ApproxQuantile,
             exp.Avg,
-            exp.Div,
             exp.Exp,
             exp.Ln,
             exp.Log,
--- a/tests/fixtures/optimizer/annotate_functions.sql
+++ b/tests/fixtures/optimizer/annotate_functions.sql
@@ -1,5 +1,5 @@
 --------------------------------------
--- Spark2 / Spark3 / Databricks functions
+-- Spark2 / Spark3 / Databricks
 --------------------------------------
 
 # dialect: spark2, spark, databricks
@@ -69,3 +69,40 @@ STRING;
 # dialect: spark2, spark, databricks
 RPAD(tbl.str_col, 1, tbl.str_col);
 STRING;
+
+
+--------------------------------------
+-- BigQuery
+--------------------------------------
+
+# dialect: bigquery
+ABS(1);
+INT;
+
+# dialect: bigquery
+ABS(1.5);
+DOUBLE;
+
+# dialect: bigquery
+GREATEST(1, 2, 3);
+INT;
+
+# dialect: bigquery
+GREATEST(1, 2.5, 3);
+DOUBLE;
+
+# dialect: bigquery
+LEAST(1, 2, 3);
+INT;
+
+# dialect: bigquery
+LEAST(1, 2.5, 3);
+DOUBLE;
+
+# dialect: bigquery
+SIGN(1);
+INT;
+
+# dialect: bigquery
+SIGN(1.5);
+DOUBLE;


sqlglot/dialects/bigquery.py
```diff
@@ -293,6 +293,14 @@ class BigQuery(Dialect):
     # All set operations require either a DISTINCT or ALL specifier
     SET_OP_DISTINCT_BY_DEFAULT = dict.fromkeys((exp.Except, exp.Intersect, exp.Union), None)
 
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        exp.Abs: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
+        exp.Greatest: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
+        exp.Least: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
```
georgesittas, line 301: Are these actually BigQuery-specific? My intuition is that these should be true for most dialects.

VaggelisD: In general yeah, functions like ABS, SIGN, FLOOR etc should be input dependent for most dialects so it could be a safe bet. Should we move these to `Dialect` now or extrapolate the common patterns once we have sufficient coverage for a few dialects?

georgesittas: I'd check a few "important" dialects to make sure this is true for them and then move these annotators out.

VaggelisD:
- Spark3:
```SQL
spark-sql (default)> select typeof(abs(1)), typeof(abs(1.5)), typeof(sign(1)), typeof(sign(1.5));
int     decimal(2,1)    double  double


spark-sql (default)> select typeof(least(1, 2)), typeof(least(1, 2.5)), typeof(greatest(1, 2)), typeof(greatest(1, 2.5));
int     decimal(11,1)   int     decimal(11,1)
```

- Snowflake
```SQL
select typeof(abs(1)), typeof(abs(1.5)), typeof(sign(1)), typeof(sign(1.5));
INTEGER | DECIMAL | INTEGER | INTEGER

select typeof(least(1, 2)), typeof(least(1, 2.5)), typeof(greatest(1, 2)), typeof(greatest(1, 2.5));
INTEGER | DECIMAL | INTEGER | DECIMAL
```

- Presto/Trino
```SQL
presto> select typeof(abs(1)), typeof(abs(1.5)), typeof(sign(1)), typeof(sign(1.5));
  _col0  |    _col1     |  _col2  |    _col3
---------+--------------+---------+--------------
 integer | decimal(2,1) | integer | decimal(1,0)


presto> select typeof(least(1, 2)), typeof(least(1, 2.5)), typeof(greatest(1, 2)), typeof(greatest(1, 2.5));
  _col0  |     _col1     |  _col2  |     _col3
---------+---------------+---------+---------------
 integer | decimal(11,1) | integer | decimal(11,1)
```

- DuckDB
```SQL
D select typeof(abs(1)), typeof(abs(1.5)), typeof(sign(1)), typeof(sign(1.5));
┌────────────────┬──────────────────┬─────────────────┬───────────────────┐
│ typeof(abs(1)) │ typeof(abs(1.5)) │ typeof(sign(1)) │ typeof(sign(1.5)) │
├────────────────┼──────────────────┼─────────────────┼───────────────────┤
│ INTEGER        │ DECIMAL(2,1)     │ TINYINT         │ TINYINT           │
└────────────────┴──────────────────┴─────────────────┴───────────────────┘

D select typeof(least(1, 2)), typeof(least(1, 2.5)), typeof(greatest(1, 2)), typeof(greatest(1, 2.5));
┌─────────────────────┬───────────────────────┬────────────────────────┬──────────────────────────┐
│ typeof(least(1, 2)) │ typeof(least(1, 2.5)) │ typeof(greatest(1, 2)) │ typeof(greatest(1, 2.5)) │
├─────────────────────┼───────────────────────┼────────────────────────┼──────────────────────────┤
│ INTEGER             │ DECIMAL(11,1)         │ INTEGER                │ DECIMAL(11,1)            │
└─────────────────────┴───────────────────────┴────────────────────────┴──────────────────────────┘
```

From this sample, I can see that `SIGN(x)` varies but the others have type borrowing semantics so should be safe to generalize.

VaggelisD:
@georgesittas  Turns out the base dialect already had proper `exp.Abs` annotation and a small bug in `exp.Least`, removed these from BigQuery.

sqlglot/dialects/bigquery.py
```diff
@@ -295,10 +295,7 @@ class BigQuery(Dialect):
 
     ANNOTATORS = {
         **Dialect.ANNOTATORS,
-        exp.Abs: lambda self, e: self._annotate_by_args(e, "this"),
         exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
```
georgesittas, line 298: We need to test this

sqlglot/dialects/dialect.py
```diff
@@ -577,7 +577,6 @@ class Dialect(metaclass=_Dialect):
         exp.DataType.Type.DOUBLE: {
             exp.ApproxQuantile,
             exp.Avg,
-            exp.Div,
```
georgesittas, line 580: Looks like this was no-op? Added [here](https://github.com/tobymao/sqlglot/commit/c6db1240b8481af0003a4bed4225f8e46578f182#diff-0df617c9c55fa6119c2fc262bbffe88e7981595bd333c95a700b031cfa78f8beR168).

sqlglot/dialects/bigquery.py
```diff
@@ -293,6 +293,11 @@ class BigQuery(Dialect):
     # All set operations require either a DISTINCT or ALL specifier
     SET_OP_DISTINCT_BY_DEFAULT = dict.fromkeys((exp.Except, exp.Intersect, exp.Union), None)
 
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
```
georgesittas, line 298: What do other dialects that support `SIGN` do? Should we annotate this as `INT` or something by default?

georgesittas (approved):
LGTM, one last comment.

sqlglot/dialects/dialect.py
VaggelisD: We have two `exp.Div` annotations, one as a static `Type.DOUBLE` (this one) and one entry in `ANNOTATORS` which is inferred by `_annotate_div(...)` which will overwrite each other; I thought we should keep the latter and enhance `_annotate_div(...)` if needed.

sqlglot/dialects/bigquery.py
VaggelisD: It's already annotated as such in the base annotators, under `Type.INT`

VaggelisD: Regarding the other dialects it's a mix of INT/DOUBLE/input, so I left it as is for now.

georgesittas: Sounds good, thanks for double checking this 👍

PR #4206, opened 2024-10-03
feat(trino): Support JSON_QUERY

### LINKED ISSUE #4200: can't parse JSON query when using `with` keyword
**Before you file an issue**

```python
from sqlglot import parse_one
parse_one("json_query(content, 'lax $.HY.*' with unconditional wrapper)",dialect="trino")
```

**Official Documentation**
https://trino.io/docs/current/functions/json.html#json-query

VaggelisD:
Fixes #4200

This PR adds parsing support for Trino's `JSON_QUERY(...)` which is a more powerful version of `JSON_EXTRACT(...)`.

For this matter, `exp.JSONExtract` node is reused, but to differentiate between the two a new arg is introduced; This is also required because Presto's `jsonextract_sql` has the `variant_access` logic which must be preserved for Trino.

Docs
---------
[JSON_QUERY](https://trino.io/docs/current/functions/json.html#json-query) | [JSON_EXTRACT](https://trino.io/docs/current/functions/json.html#json_extract)

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -33,7 +33,6 @@ from sqlglot.dialects.dialect import (
 )
 from sqlglot.dialects.hive import Hive
 from sqlglot.dialects.mysql import MySQL
-from sqlglot.generator import Generator
 from sqlglot.helper import apply_index_offset, seq_get
 from sqlglot.tokens import TokenType
 from sqlglot.transforms import unqualify_columns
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/presto.py
```diff
@@ -166,7 +167,7 @@ def _unix_to_time_sql(self: Presto.Generator, expression: exp.UnixToTime) -> str
     return f"FROM_UNIXTIME(CAST({timestamp} AS DOUBLE) / POW(10, {scale}))"
 
 
-def _jsonextract_sql(self: Presto.Generator, expression: exp.JSONExtract) -> str:
+def jsonextract_sql(self: Generator, expression: exp.JSONExtract) -> str:
```
georgesittas, line 170: Should we just move this within presto's `Generator` and then use `super()` in Trino instead of importing it?

georgesittas (commented):
What are the advantages of treating this as `JSONExtract`? Other dialects also support `JSON_QUERY`, possibly with similar semantics. Check T-SQL for example.

VaggelisD:
@georgesittas For Trino, `JSON_EXTRACT` and `JSON_QUERY` are semantically equivalent with the exception that `JSON_QUERY` allows for more options baked-in; Splitting these two to separate nodes would duplicate a lot of code for no particular reason as it stands. My guess is that the former is the Presto-compatible API while the latter is Trino specific as I couldn't find it in Presto's docs.

Afaict, for other dialects like T-SQL `JSON_QUERY` is just the equivalent function for extraction so I'm also not sure if there are any nuances that justify splitting these two for now. Wdyt?

georgesittas:
> @georgesittas For Trino, `JSON_EXTRACT` and `JSON_QUERY` are semantically equivalent with the exception that `JSON_QUERY` allows for more options baked-in; Splitting these two to separate nodes would duplicate a lot of code for no particular reason as it stands. My guess is that the former is the Presto-compatible API while the latter is Trino specific as I couldn't find it in Presto's docs.

Not sure if this is true, I think there are some differences. For example, there's a path that looks like `'lax $.children[*]'`. I don't remember needing to specify `lax` or `strict` in `JSON_EXTRACT`. Also, at least in T-SQL (perhaps others too), `JSON_QUERY` only extracts _nested_ values, and you need to use `JSON_VALUE` for literals.

georgesittas:
> For T-SQL you'd indeed need JSON_VALUE for scalar values but that'd be exp.JSONExtractScalar, right?

Not exactly. Take these T-SQL expressions, for example:

```
1> select json_query('{"x": 1}', '$.x')
2> GO

georgesittas (approved):

PR #4212, opened 2024-10-04
feat(bigquery): Native annotations for more math functions

VaggelisD:
This is a follow up to https://github.com/tobymao/sqlglot/pull/4201

Several BigQuery functions such as `CEIL(x)`, `FLOOR(x)` etc will borrow the input type _unless_ that is an `INT`, in which case a `FLOAT64` (i.e `DataType.DOUBLE`) is returned. 

To mimic this behavior, this PR adds a new type annotation helper.

Note: The function `TRUNC(X)` is not currently supported by SQLGlot so once that's added, it must also be annotated accordingly.

Docs
-------
[BQ Mathematical Functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -33,6 +33,8 @@ from sqlglot.tokens import TokenType
 if t.TYPE_CHECKING:
     from sqlglot._typing import E, Lit
 
+    from sqlglot.optimizer.annotate_types import TypeAnnotator
+
 logger = logging.getLogger("sqlglot")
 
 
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -247,6 +249,26 @@ def _str_to_datetime_sql(
...
+    Many BigQuery math functions such as CEIL, FLOOR etc follow this return type convention:
+    +---------+---------+---------+------------+---------+
+    |  INPUT  | INT64   | NUMERIC | BIGNUMERIC | FLOAT64 |
+    +---------+---------+---------+------------+---------+
+    |  OUTPUT | FLOAT64 | NUMERIC | BIGNUMERIC | FLOAT64 |
+    +---------+---------+---------+------------+---------+
+    """
+    self._annotate_args(expression)
+
+    this = expression.args.get("this")
```
georgesittas, line 263:
```suggestion
    this = expression.this
```

sqlglot/dialects/bigquery.py
```diff
@@ -247,6 +249,26 @@ def _str_to_datetime_sql(
...
+    input_type = this.type if this else exp.DataType.Type.UNKNOWN
```
georgesittas, line 264: When is `this` falsey here?

sqlglot/dialects/bigquery.py
```diff
@@ -247,6 +249,26 @@ def _str_to_datetime_sql(
...
+    self._set_type(
+        expression,
+        exp.DataType.Type.DOUBLE if input_type.is_type(exp.DataType.Type.INT) else input_type,
```
georgesittas, line 267: Should this be `is_type(*INTEGER_TYPES)` instead?

georgesittas (commented):
Few comments, LGTM otherwise.

sqlglot/dialects/bigquery.py
VaggelisD: Brain fart moment, thanks

sqlglot/dialects/bigquery.py
VaggelisD: Shouldn't be, mostly satisfying the linter because these objects can be optional. Should we go with `type: ignore`?

georgesittas: Maybe just `this: exp.Expression = expression.this`?

georgesittas (approved):

PR #4217, opened 2024-10-07
Feat: add builder methods to exp.Update and add with_ arg to exp.update

brdbry:
Improve ergonomics of UPDATEs:

- Add builder methods to Update class so it can be constructed incrementally in the same way a Select
- `exp.update` changes:
  - Add a `with_` arg so updates with CTEs can be created in one-shot without subsequent need  to `.set(...)` the With clause
  - Make the `properties` arg optional, so `set_` builder method can be used with an AST (`exp.update("tbl").set_(set_expr)` ) instead of forcing specification as a `dict`

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -3284,7 +3284,9 @@ class Update(Expression):
         "limit": False,
     }
 
-    def table(self, expression: ExpOrStr, dialect: DialectType = None, copy: bool = True, **opts) -> Update:
+    def table(
+        self, expression: ExpOrStr, dialect: DialectType = None, copy: bool = True, **opts
+    ) -> Update:
         """
         Set the table to update.
 
@@ -3313,8 +3315,10 @@ class Update(Expression):
             copy=copy,
             **opts,
         )
-    
-    def set_(self, *expressions: ExpOrStr, dialect: DialectType = None, copy: bool = True, **opts) -> Update:
+
+    def set_(
+        self, *expressions: ExpOrStr, dialect: DialectType = None, copy: bool = True, **opts
+    ) -> Update:
         """
         Set the SET expressions.
 
@@ -3380,7 +3384,11 @@ class Update(Expression):
         )
 
     def from_(
-        self, expression: t.Optional[ExpOrStr] = None, dialect: DialectType = None, copy: bool = True, **opts
+        self,
+        expression: t.Optional[ExpOrStr] = None,
+        dialect: DialectType = None,
+        copy: bool = True,
+        **opts,
     ) -> Update:
         """
         Set the FROM expression.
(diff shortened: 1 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -3284,6 +3284,191 @@ class Update(Expression):
...
+                If nothing is passed in then a from is not applied to the expression
+            dialect: the dialect used to parse the input expression.
+            copy: if `False`, modify this expression instance in-place.
+            opts: other options to use to parse the input expressions.
+
+        Returns:
+            The modified Update expression.
+        """
+        if not expression:
+            return self if not copy else self.copy()
```
georgesittas, line 3413:
```suggestion
            return maybe_copy(self)
```

sqlglot/expressions.py
```diff
@@ -3284,6 +3284,191 @@ class Update(Expression):
...
+            into=Table,
+            prefix=None,
+            dialect=dialect,
+            copy=copy,
+            **opts,
+        )
+
+    def set_(
+        self, *expressions: ExpOrStr, dialect: DialectType = None, copy: bool = True, **opts
+    ) -> Update:
```
georgesittas, line 3321: We probably want an "append" mode like we do for, e.g., `select`. Right now you always overwrite which isn't as flexible.

sqlglot/expressions.py
```diff
@@ -6803,38 +6988,41 @@ def from_(expression: ExpOrStr, dialect: DialectType = None, **opts) -> Select:
 
 def update(
     table: str | Table,
-    properties: dict,
+    properties: t.Optional[dict] = None,
```
georgesittas, line 6991: Why did this become optional?

sqlglot/expressions.py
```diff
@@ -6847,6 +7035,15 @@ def update(
             "where",
             maybe_parse(where, into=Where, dialect=dialect, prefix="WHERE", **opts),
         )
+    if with_:
+        cte_list = [
+            CTE(this=maybe_parse(qry, into=Select, dialect=dialect, **opts), alias=alias)
```
georgesittas, line 7040: Will this work for set operators as well? For example `UNION`, etc.

georgesittas (commented):
LGTM, just a few comments. Thanks a lot for the PR, @brdbry!

sqlglot/expressions.py
brdbry: Good idea, done

sqlglot/expressions.py
brdbry: I put it in the PR comments at the top. Making it mandatory prevents you from using the exp.update function when you already have a list SET expressions, so it's not consistent with other helper methods like exp.select

georgesittas: Ah, I see; apologies, I didn't notice the description. Will take another look.

georgesittas (approved):

sqlglot/expressions.py
georgesittas: Sounds reasonable 👍

PR #4218, opened 2024-10-07
Fix: Attach CTE comments before commas

### LINKED ISSUE #4216: Dropping and reordering comments when using CTEs
I think I've found some corner cases where:

1. Comments are dropped
2. Comments before the final select statement after a sequence of CTEs gets moved to the top of the query
3. Some of the comments are reproduced out of order

## Example

Here's an example:

```python
import sqlglot
def foo(s):
    print(sqlglot.transpile(s, read='presto', write='spark', pretty=True)[0])

s1 = """
-- comment at the top
with
-- comment for tbl1
tbl1 as (select 1)

-- comment for tbl2
, tb12 as (select 2)

-- comment for tbl3
, tb13 as (select 3)

-- comment for final select
select * from tbl1
"""
foo(s1)
```

gives me the output:

```sql
/* comment for final select */
/* comment at the top */
WITH tbl1 /* comment for tbl1 */ AS (
  SELECT
    1
), tb12 AS (
  SELECT
    2
), tb13 AS (
  SELECT
    3
)
SELECT
  *
FROM tbl1
```

## Expected behavior

1. Preserve comments for `tbl2` and `tbl3` (they are missing in the example above)
2. If possible, keep the comment for the final select closer to where it was in the original query
4. Even if it can't be kept closer, the comments are out of order, as the "comment at the top" should be first.

## Alternative inputs

I played with the example a bit, and it seems that the dropped comments have to do with the comma placement. I can preserve the comments if I do something like the example below, but, ideally, they would also be preserved in the example above. With the example below, we still have the issue of the comment at the top and the final select comment showing up out of order, and the final select comment being moved to the top of the output (rather than near the final select statement).

```python
s2 = """
-- comment at the top
with
-- comment for tbl1
tbl1 as (select 1),

-- comment for tbl2
tb12 as (select 2),

-- comment for tbl3
tb13 as (select 3)

-- comment for final select
select * from tbl1
"""
foo(s2)
```

gives the output:

```sql
/* comment for final select */
/* comment at the top */
WITH tbl1 /* comment for tbl1 */ AS (
  SELECT
    1
), tb12 /* comment for tbl2 */ AS (
  SELECT
    2
), tb13 /* comment for tbl3 */ AS (
  SELECT
    3
)
SELECT
  *
FROM tbl1
```

## Version

This behavior is happening on the main branch. I ran the example using `sqlglot.__version__='25.24.5.dev2'`

VaggelisD:
Fixes #4216

This PR addresses 2 points:
1. On main, comments on SELECT & it's CTEs bubble up:

```SQL
sql = """
-- comment at the top
with tbl as (select 1)
-- comment for final select
select * from tbl1
"""

sqlglot.parse_one(sql).sql(pretty=True)

/* comment for final select */
/* comment at the top */
WITH tbl AS (
  SELECT
    1
)
SELECT
  *
FROM tbl1
```

After this PR, the `SELECT` generator will differentiate the two by attaching it's comments earlier:
```SQL
/* comment at the top */
WITH tbl AS (
  SELECT
    1
)
/* comment for final select */
SELECT
  *
FROM tbl1
```

2. Comments before the CTE commas are now preserved and attached to the next node, as is shown in the test case

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1197,10 +1197,8 @@ class Generator(metaclass=_Generator):
 
     def cte_sql(self, expression: exp.CTE) -> str:
         alias = expression.args.get("alias")
-        comments = expression.comments
-        if alias and comments:
-            alias.add_comments(comments)
-            expression.pop_comments()
+        if alias:
+            alias.add_comments(expression.pop_comments())
 
         alias_sql = self.sql(expression, "alias")
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1196,7 +1196,13 @@ class Generator(metaclass=_Generator):
         return f"WITH {recursive}{sql}"
 
     def cte_sql(self, expression: exp.CTE) -> str:
-        alias = self.sql(expression, "alias")
+        alias = expression.args.get("alias")
+        comments = expression.comments
+        if alias and comments:
+            alias.add_comments(comments)
+            expression.pop_comments()
+
+        alias_sql = self.sql(expression, "alias")
 
         materialized = expression.args.get("materialized")
         if materialized is False:
@@ -1204,7 +1210,7 @@ class Generator(metaclass=_Generator):
         elif materialized:
             materialized = "MATERIALIZED "
 
-        return f"{alias} AS {materialized or ''}{self.wrap(expression)}"
+        return f"{alias_sql} AS {materialized or ''}{self.wrap(expression)}"
 
     def tablealias_sql(self, expression: exp.TableAlias) -> str:
         alias = self.sql(expression, "this")
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2517,6 +2517,11 @@ class Generator(metaclass=_Generator):
             self.sql(expression, "from", comment=False),
         )
 
+        # If both the CTE and SELECT clauses have comments, generate the latter earlier
+        if expression.args.get("with"):
+            sql = self.maybe_comment(sql, expression)
+            expression.pop_comments()
+
         sql = self.prepend_ctes(expression, sql)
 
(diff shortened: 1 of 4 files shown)


tests/test_transpile.py
```diff
@@ -565,6 +565,35 @@ def test_comments(self):
...
+-- comment for tbl1
+tbl1 AS (SELECT 1)
+-- comment for tbl2
+, tbl2 AS (SELECT 2)
+-- comment for tbl3
+, tbl3 AS (SELECT 3)
+-- comment for final select
+SELECT * FROM tbl1""",
+            """/* comment at the top */
+WITH tbl1 /* comment for tbl1 */ AS (
```
georgesittas, line 581: This looks somewhat inconsistent with how we're generating comments for the subsequent CTEs. Is it simple to have the comments always go after the table name or after the closing parenthesis?

georgesittas (commented):
LGTM, one question to slightly improve this.

tests/test_transpile.py
VaggelisD: It does, tried something quick in my latest commit

sqlglot/generator.py
```diff
@@ -1196,15 +1196,21 @@ def with_sql(self, expression: exp.With) -> str:
         return f"WITH {recursive}{sql}"
 
     def cte_sql(self, expression: exp.CTE) -> str:
-        alias = self.sql(expression, "alias")
+        alias = expression.args.get("alias")
+        comments = expression.comments
+        if alias and comments:
+            alias.add_comments(comments)
+            expression.pop_comments()
```
georgesittas, line 1203:
```suggestion
        if alias:
            alias.add_comments(expression.pop_comments())
```

georgesittas (approved):

georgesittas (approved):

PR #4223, opened 2024-10-08
Fix: Make exp.Update a DML node

### LINKED ISSUE #4221: Support traversing scope within UPDATE statements.
`traverse_scope()` function does not handle SELECT statements within UPDATE statements. 
I was expecting couple scopes form below example but it ignores any SELECTS within UPDATES.

```py

from sqlglot import Parser, exp, parse, parse_one

sql = """

UPDATE customers
SET total_spent = (
    SELECT SUM(amount)
    FROM payments
    WHERE payments.customer_id = customers.id
)
WHERE EXISTS (
    SELECT 1
    FROM payments
    WHERE payments.customer_id = customers.id
);
"""

parsed = parse_one(sql)
for scope in traverse_scope(parsed):
    print(scope)

```

Any chance this can be fixed?

Thanks!

VaggelisD:
Fixes #4221

This PR:

- Makes `exp.Update` a `DML` node as is semantically fitting; This also enables `traverse_scope` on UPDATE statements
- Fixes a bug with DML scope traversal, as the `exp.Select` children of subqueries would be visited twice (one in `_traverse_subqueries` and one in the `find_all_in_scope(expression, exp.Query)` block)


georgesittas (approved):
Let's increase the testing coverage for this

VaggelisD:
Do you have something specific in mind, e.g. should we check each scope individually for this particular case or introduce more `UPDATE` tests?

georgesittas:
> Do you have something specific in mind, e.g. should we check each scope individually for this particular case or introduce more `UPDATE` tests?

I was thinking some more `UPDATE` tests, use different clauses and see we're detecting the right scopes.

georgesittas (approved):

PR #4224, opened 2024-10-08
Refactor!: Rename exp.RenameTable to exp.AlterRename

### LINKED ISSUE #4222: PostgreSQL RENAME INDEX is interpreted as RenameTable operation
When parsing an alter index postgresql command like this:

`ALTER INDEX "IX_Ratings_Column1" RENAME TO "IX_Ratings_Column2"`

SqlGlot parses it as a RenameTable operation:

```
[Alter(
  this=Table(
    this=Identifier(this=IX_Ratings_Column1, quoted=True)),
  kind=INDEX,
  actions=[
    RenameTable(
      this=Table(
        this=Identifier(this=IX_Ratings_Column2, quoted=True)))])]
```

Shouldn't it be something like this?

```
[Alter(
  this=Index(
    this=Identifier(this=IX_Ratings_Column1, quoted=True)),
  kind=INDEX,
  actions=[
    RenameIndex(
      this=Index(
        this=Identifier(this=IX_Ratings_Column2, quoted=True)))])]
```

VaggelisD:
Fixes #4222

PS: I believe we could also consolidate `exp.RenameColumns` into `exp.AlterRename`, is this worth looking in this PR / soon?

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -1166,7 +1166,7 @@ class TSQL(Dialect):
 
         def alter_sql(self, expression: exp.Alter) -> str:
             action = seq_get(expression.args.get("actions") or [], 0)
-            if isinstance(action, exp.RenameTable):
+            if isinstance(action, exp.AlterRename):
                 return f"EXEC sp_rename '{self.sql(expression.this)}', '{action.this.name}'"
             return super().alter_sql(expression)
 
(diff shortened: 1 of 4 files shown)


georgesittas (approved):
If consolidating `AlterRename` and `RenameColumn` is simple, we could do that as well here.

VaggelisD:
True, it's not as needed yet, can be revisited in the future.

PR #4231, opened 2024-10-09
feat(bigquery): Native annotations for string functions (1)

VaggelisD:
Follow up on https://github.com/tobymao/sqlglot/pull/4212 to increase native BQ annotation.

According to the docs, these `STRING` / `BYTES` (i.e `BINARY`) functions are input type dependent:

1. CONCAT 
2. LEFT / RIGHT
3. LOWER / UPPER
4. LPAD / RPAD
5. LTRIM / RTRIM / TRIM 
6. REGEXP_EXTRACT, REGEXP_REPLACE, REGEXP_SUBSTR, REPEAT, REPLACE, REVERSE, SUBSTR, TRANSLATE
7. REGEXP_EXTRACT_ALL, SPLIT (these return `ARRAY<STRING>` / `ARRAY<BYTES>` instead of scalar values)

To limit the scope, this PR addresses (1) - (5) and the next one will cover the rest.

Note: Unlike Spark, BQ functions with multiple inputs such as `CONCAT` do not allow mixed types so it's sufficient to annotate according to one of them e.g. `this`  

Docs
--------
[BQ String Functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -321,7 +321,19 @@ class BigQuery(Dialect):
             expr_type: lambda self, e: _annotate_math_functions(self, e)
             for expr_type in (exp.Floor, exp.Ceil, exp.Log, exp.Ln, exp.Sqrt, exp.Exp, exp.Round)
         },
+        **{
+            expr_type: lambda self, e: self._annotate_by_args(e, "this")
+            for expr_type in (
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
I think we should be marking these as breaking, they affect SQLMesh schema inference.

PR #4235, opened 2024-10-11
feat: support EXPLAIN in mysql

xiaoyu-meng-mxy:
Currently EXPLAIN is not supported in sqlglot.dialects.mysql. 

Because 'The [DESCRIBE](https://dev.mysql.com/doc/refman/8.4/en/describe.html) and [EXPLAIN](https://dev.mysql.com/doc/refman/8.4/en/explain.html) statements are synonyms.', this pr supports EXPLAIN by treat it as DESCRIBE

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1337,6 +1337,8 @@ class Parser(metaclass=_Parser):
     # Whether a PARTITION clause can follow a table reference
     SUPPORTS_PARTITION_SELECTION = False
 
+    DESCRIBE_STATEMENT_STYLES = {"ANALYZE", "EXTENDED", "FORMATTED", "HISTORY"}
+
     __slots__ = (
         "error_level",
         "error_message_context",
@@ -2567,10 +2569,7 @@ class Parser(metaclass=_Parser):
 
     def _parse_describe(self) -> exp.Describe:
         kind = self._match_set(self.CREATABLES) and self._prev.text
-        style = (
-            self._match_texts(("ANALYZE", "EXTENDED", "FORMATTED", "HISTORY"))
-            and self._prev.text.upper()
-        )
+        style = self._match_texts(self.DESCRIBE_STATEMENT_STYLES) and self._prev.text.upper()
         if self._match(TokenType.DOT):
             style = None
             self._retreat(self._index - 2)
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1286,6 +1286,8 @@ class Parser(metaclass=_Parser):
 
     PRIVILEGE_FOLLOW_TOKENS = {TokenType.ON, TokenType.COMMA, TokenType.L_PAREN}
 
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -2567,7 +2567,10 @@ def _parse_returns(self) -> exp.ReturnsProperty:
 
     def _parse_describe(self) -> exp.Describe:
         kind = self._match_set(self.CREATABLES) and self._prev.text
-        style = self._match_texts(("EXTENDED", "FORMATTED", "HISTORY")) and self._prev.text.upper()
+        style = (
+            self._match_texts(("ANALYZE", "EXTENDED", "FORMATTED", "HISTORY"))
```
georgesittas, line 2571: Let's make this a class-level set constant and reference that instead.

tests/dialects/test_mysql.py
```diff
@@ -1305,3 +1305,12 @@ def test_grant(self):
         for sql in grant_cmds:
             with self.subTest(f"Testing MySQL's GRANT command statement: {sql}"):
                 self.validate_identity(sql, check_command_warning=True)
+
+    def test_explain(self):
+        self.validate_identity(
+            "EXPLAIN ANALYZE SELECT * FROM t", write_sql="DESCRIBE ANALYZE SELECT * FROM t"
```
georgesittas, line 1311: [Nit]: we don't usually specify `write_sql` as a kwarg.
```suggestion
            "EXPLAIN ANALYZE SELECT * FROM t",
            "DESCRIBE ANALYZE SELECT * FROM t",
```

georgesittas (approved):
LGTM, thanks for the PR. Just a couple minor comments.

sqlglot/parser.py
```diff
@@ -1286,6 +1286,8 @@ class Parser(metaclass=_Parser):
 
     PRIVILEGE_FOLLOW_TOKENS = {TokenType.ON, TokenType.COMMA, TokenType.L_PAREN}
 
+    DESCRIBE_STATEMENT_STYLES = {"ANALYZE", "EXTENDED", "FORMATTED", "HISTORY"}
```
xiaoyu-meng-mxy, line 1289: I see "HISTORY" is added for dialects.databricks

I can also remove HISTORY from here and override it in databricks parser, if necessary. And do the same for ANALYZE in mysql parser. Not sure whether this is needed 🙏

sqlglot/parser.py
xiaoyu-meng-mxy: Makes sense. Thanks for the quick review!

sqlglot/parser.py
barakalon: sqlglot values brevity.

How about just "DESCRIBE_STYLES"?

barakalon (approved):

georgesittas (approved):
Thank you for the PR!

PR #4243, opened 2024-10-15
Feat: support non-strict qualify_columns

hsheth2:
Wasn't sure whether to reuse the existing `validate_qualify_columns` flag name, or to create a new one. My thinking is validate_qualify_columns False = best effort qualification, and True = required qualification. This matches that behavior - let me know what you think.

### DIFF
--- a/sqlglot/optimizer/qualify.py
+++ b/sqlglot/optimizer/qualify.py
@@ -27,6 +27,7 @@ def qualify(
     infer_schema: t.Optional[bool] = None,
     isolate_tables: bool = False,
     qualify_columns: bool = True,
+    allow_partial_qualification: bool = False,
     validate_qualify_columns: bool = True,
     quote_identifiers: bool = True,
     identify: bool = True,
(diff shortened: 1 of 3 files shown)


tobymao:
i'm not sure about this approach. why would something like this happen? and how would partial qualification be useful?

hsheth2:
We usually run SQL parsing on a daily cadence, pulling from query logs. However, we can only get the current schemas for tables. Because there's some time between the two, we would frequently observe discrepancies between the two when columns are added/removed in the time between.

sqlglot's lineage generation seems to degrade somewhat gracefully when columns are partially qualified, especially for simple sql statements - which are the most common. So partial qualification is pretty useful.

georgesittas (approved):
Seems reasonable, quite a low lift too.

georgesittas:
Actually, I take it back, apologies. Reusing the `validate_qualify_columns` flag changes the behavior of existing code that only cares about skipping the `validate_qualify_columns` call in `qualify`.

If we had a separate flag for this I'd be ok with it, for example `allow_partial_qualification: bool = False` or something. But then this kind of overlaps with what `validate_qualify_columns` does already, so it's still not super clean...

tobymao (approved):

PR #4244, opened 2024-10-15
Fix: protect against cyclic scopes

hsheth2:
This ensures that we don't hit infinite loops.

Given that this is a defensive measure, I'm not sure what the best way to test this would be. Open to suggestions on that.

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -800,6 +800,13 @@ FROM READ_CSV('tests/fixtures/optimizer/tpc-h/nation.csv.gz', 'delimiter', '|') 
             level="warning",
         )
 
+    def test_scope_circular_dependency(self):
+        scope = build_scope(parse_one("WITH w AS (SELECT * FROM q) SELECT * FROM w"))
+        cte_scope = scope.cte_scopes[0]
+        cte_scope.cte_scopes.append(cte_scope)
+        with self.assertRaisesRegex(OptimizeError, "circular scope dependency"):
+            list(scope.traverse())
+
     def test_annotate_types(self):
         for i, (meta, sql, expected) in enumerate(
             load_sql_fixture_pairs("optimizer/annotate_types.sql"), start=1
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -461,9 +461,16 @@ class Scope:
             Scope: scope instances in depth-first-search post-order
         """
         stack = [self]
+        seen_scopes = set()
         result = []
         while stack:
             scope = stack.pop()
+
+            # Scopes aren't hashable, so we use id(scope) instead.
+            if id(scope) in seen_scopes:
+                raise OptimizeError(f"Scope {scope} has a circular scope dependency")
+            seen_scopes.add(id(scope))
+
             result.append(scope)
             stack.extend(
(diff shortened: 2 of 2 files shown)


tobymao:
can you add a test case for this? you can ensure that it raises

georgesittas:
Did you actually come across cyclic scopes? How do you even reach that state? I'm somewhat skeptical about this due to the added complexity & overhead, not sure yet if it's justified.

tests/test_optimizer.py
```diff
@@ -800,6 +800,13 @@ def test_scope_warning(self, logger):
             level="warning",
         )
 
+    def test_scope_circular_dependency(self):
+        scope = build_scope(parse_one("WITH w AS (SELECT * FROM q) SELECT * FROM w"))
+        cte_scope = scope.cte_scopes[0]
```
tobymao, line 805: this test doesn't make sense. why would this ever happen?

hsheth2:
I've seen issues in the past where `build_scope` generated a cyclic setup. My original fix was to just set a max scope depth (https://github.com/tobymao/sqlglot/compare/main...hsheth2:sqlglot:main), but figured this would be a more reliable fix.

I'll try and to dig up the original SQL that caused it, but not sure I still have it handy.

georgesittas:
My hunch is that this could only happen if an AST had different nodes with the same `id`, which would be a user issue to begin with, imo.

tobymao:
i'm going to close this for now. if you can show a code path where this exists, i think we should fix that at parse time / optimization etc. if we cannot, then i'll accept a fix like this

hsheth2:
Example script below - it will currently hang in `scope.traverse()` right now.

More broadly, I think that if we can detect situations where we might end up in an infinite loop, it makes sense to raise an error instead of simply letting it happen.

```python
import sqlglot
import sqlglot.optimizer

query = """\
SELECT
  clicks.user_id,
  clicks.email_address_id,
  sum(IF(clicks.utm_source = 'newsletter', 1, 0)) AS newsletter_clicks

FROM (
  SELECT emails.user_id, clicks.email_address_id, clicks.utm_source
  FROM `proj.campaign1.clicks` as clicks
  JOIN emails USING (email_address_id)
  WHERE clicks.utm_source IN (
    'a', 'b', 'c'
  )
  AND _date = current_run_date

  UNION ALL

  SELECT emails.user_id, clicks.email_address_id, clicks.utm_source
  FROM `proj.campaign2.clicks` as clicks
  JOIN emails USING (email_address_id)

  UNION ALL

  SELECT emails.user_id, clicks.email_address_id, clicks.utm_source
  FROM `proj.campaign3.clicks` as clicks
  JOIN emails USING (email_address_id)
) as clicks
"""

optimized = sqlglot.optimizer.optimizer.optimize(
    query,
    dialect="bigquery",
    validate_qualify_columns=False,
)
```

tobymao:
@hsheth2 thanks for the example

tobymao:
this is simply a bug in the optimizer, and the cyclic scope is only a patch to the underlying issue

tobymao:
this is due to user error. downstream rules require full column qualification. since you passed in validate_qualify_columns=False, the rules were not correctly applied, causing unnest_subqueries to incorrectly detect that you have a decorrelated subquery

PR #4246, opened 2024-10-15
fix(bigquery): Early expand only aliased names in GROUP BY

VaggelisD:
[Public Slack context](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1728475163266849)

This is a follow up fix on https://github.com/tobymao/sqlglot/pull/3699, which introduced early alias expansion for BQ and Clickhouse.

As a recap, BigQuery allows `GROUP BY` to refer to `SELECT` aliases. From the [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#group_by_values): 

_GROUP BY clauses may also refer to aliases. If a query contains aliases in the SELECT clause, those aliases override names in the corresponding FROM clause._

So the early expansion was necessary to replace the references with the aliased expressions instead of qualifying them, as is explained in the PR. However, it seems that BQ allows this only for grouped _names_ and not _expressions_:

```SQL
bigquery> with t as (select 2 as id) select id + 1 as col from t group by col;
col
3

bigquery> with t as (select 2 as id) select id + 1 as col from t group by col + 0;
Syntax error: Unexpected identifier "col" at [1:1]
```

This means that if an alias is part of an _expression_, BQ will infer it as a column. In contrast, up until now, the early alias expansion was performed blindly so the aliased expression would replace any referenced instance in the `GROUP BY` clause. This can lead to bugs such as this example with conflicting column & alias names:

```SQL
>>> sql = """
SELECT
  FUNC(col) AS col
FROM tbl
GROUP BY
  FUNC(col)
"""

>>> qualify_columns(parse_one(sql, dialect="bigquery"), schema=MappingSchema({"tbl": {"col": "TIMESTAMP"}}, dialect="bigquery")).sql("bigquery")

SELECT
  FUNC(`tbl`.`col`) AS `col`
FROM `tbl` AS `tbl`
GROUP BY
  FUNC(FUNC(`tbl`.`col`))
```

This PR fixes this by not allowing the expansion to happen for grouped _expressions_, but only for grouped _names_; In the previous example, the expression is preserved as is:

```SQL
>>> qualify_columns(parse_one(sql, dialect="bigquery"), schema=MappingSchema({"tbl": {"col": "TIMESTAMP"}}, dialect="bigquery")).sql("bigquery")

SELECT
  FUNC(`tbl`.`col`) AS `col`
FROM `tbl` AS `tbl`
GROUP BY
  FUNC(`tbl`.`col`)
```

The `expand_only_groupby` is a proxy for BQ instead of having to add `dialect` to `expand_alias_refs(...)`.

Docs
--------
[BQ GROUP BY](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#group_by_values) | [GROUP BY with clashing names](https://freedium.cfd/https://hanqi01.medium.com/how-badly-named-sql-column-aliases-confuse-when-using-group-by-and-order-by-261541f86f9a)

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -248,7 +248,10 @@ def _expand_alias_refs(scope: Scope, resolver: Resolver, expand_only_groupby: bo
             if not isinstance(column, exp.Column):
                 continue
 
-            # BigQuery should expand aliased expressions only if the alias is used as a standalone name
+            # BigQuery's GROUP BY allows alias expansion only for standalone names, e.g:
+            #   SELECT FUNC(col) AS col FROM t GROUP BY col --> Can be expanded
+            #   SELECT FUNC(col) AS col FROM t GROUP BY FUNC(col)  --> Shouldn't be expanded, will result to FUNC(FUNC(col))
+            # This not required for the HAVING clause as it can evaluate expressions using both the alias & the table columns
             if expand_only_groupby and is_group_by and column.parent is not node:
                 continue
 
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -240,11 +240,16 @@ def _expand_alias_refs(scope: Scope, resolver: Resolver, expand_only_groupby: bo
     def replace_columns(
         node: t.Optional[exp.Expression], resolve_table: bool = False, literal_index: bool = False
     ) -> None:
-        if not node or (expand_only_groupby and not isinstance(node, exp.Group)):
+        is_group_by = isinstance(node, exp.Group)
+        if not node or (expand_only_groupby and not is_group_by):
             return
 
         for column in walk_in_scope(node, prune=lambda node: node.is_star):
             if not isinstance(column, exp.Column):
+                continue
+
+            # BigQuery should expand aliased expressions only if the alias is used as a standalone name
+            if expand_only_groupby and is_group_by and column.parent is not node:
                 continue
 
             table = resolver.get_table(column.name) if resolve_table and not column.table else None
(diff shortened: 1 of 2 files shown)


tests/test_optimizer.py
```diff
@@ -602,6 +602,16 @@ def test_expand_alias_refs(self):
             "WITH data AS (SELECT 1 AS id, 2 AS my_id, 'a' AS name, 'b' AS full_name) SELECT data.id AS my_id, CONCAT(data.id, data.name) AS full_name FROM data WHERE data.id = 1 GROUP BY data.id, CONCAT(data.id, data.name) HAVING data.id = 1",
         )
 
+        # Edge case: BigQuery shouldn't expand aliases that are reusing input column names in expressions
```
georgesittas, line 605: I think we can improve this a bit, e.g. "BigQuery shouldn't expand aliases in complex expressions" or something. The "that are reusing input column names" is confusing to me.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -240,13 +240,18 @@ def _expand_alias_refs(scope: Scope, resolver: Resolver, expand_only_groupby: bo
...
+        if not node or (expand_only_groupby and not is_group_by):
             return
 
         for column in walk_in_scope(node, prune=lambda node: node.is_star):
             if not isinstance(column, exp.Column):
                 continue
 
+            # BigQuery should expand aliased expressions only if the alias is used as a standalone name
+            if expand_only_groupby and is_group_by and column.parent is not node:
+                continue
```
georgesittas, line 253: Is bigquery able to only expand aliases within `GROUP BY`, or other clauses too?

georgesittas: I'd consider including this context in the comment in L251.

georgesittas (approved):

sqlglot/optimizer/qualify_columns.py
VaggelisD: Sg, tried to summarize these 2 concepts in that comment

georgesittas (approved):

PR #4249, opened 2024-10-15
Feature/oracle hints

mkmoisen:
Hi @georgesittas 

Please review this PR we discussed over on [Slack](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1728489927717069), which adds better support for Oracle Hints.

----

The current sqlglot implementation is to parse oracle hints as one big string, instead of breaking them up into Anonymous and Var expressions. 

This PR attempts to parse Oralce hints into Anonymous and Var expressions. In the event the user has some kind of spelling error, or uses keywords like `select`, this will fall back to the original implementation, instead of throwing an error or instead of parsing into Anonymous/Var.

Oracle hints differ from other dialects in that arguments to hint functions are separated by space, not comma. Oracle does not, as far as I am aware from experience and the documentation, support nested hint functions. In the event that nested hints are supported or will by supported in the future, this implementation will simply fall back and parse it as a string.

This PR also handles the pretty printing in a way that is identical with how sqlglot does it for mysql. Please see the tests for a few examples.


sqlglot/dialects/oracle.py
```diff
@@ -205,6 +206,64 @@ def _parse_json_array(self, expr_type: t.Type[E], **kwargs) -> E:
...
+            return self.expression(exp.Hint, expressions=hints)
+
+        def _parse_hint_function_call(self) -> t.Optional[exp.Expression]:
+            if not self._curr:
+                return None
+
+            this = self._curr.text
+
+            if not self._next or self._next.token_type != TokenType.L_PAREN:
+                return None
```
georgesittas, line 244: Let's consolidate these two branches:

```python
if not self._curr or not self._next or self._next.token_type != TokenType.L_PAREN: ...
```

sqlglot/dialects/oracle.py
```diff
@@ -205,6 +206,64 @@ def _parse_json_array(self, expr_type: t.Type[E], **kwargs) -> E:
...
+
+            self._advance(2)
+
+            args = self._parse_hint_args()
+
+            this = self.expression(exp.Anonymous, this=this, expressions=args)
+
+            self._match_r_paren(this)
+
+            return this
```
georgesittas, line 254: Let's reduce the spacing here a bit, it's a bit too much.
```suggestion
            self._advance(2)
            this = self.expression(exp.Anonymous, this=this, expressions=self._parse_hint_args())
            self._match_r_paren(this)
            return this
```

sqlglot/dialects/oracle.py
```diff
@@ -370,3 +430,25 @@ def into_sql(self, expression: exp.Into) -> str:
...
+                    formatted_args = self._format_hint_function_args(*expression.expressions)
+                    expressions.append(f"{self.sql(expression, 'this')}({formatted_args})")
+                else:
+                    expressions.append(self.sql(expression))
+
+            return f" /*+ {self.expressions(sqls=expressions, sep=self.QUERY_HINT_SEP).strip()} */"
+
+        def _format_hint_function_args(self, *args: t.Optional[str | exp.Expression]) -> str:
+            arg_sqls = tuple(
+                self.sql(arg) for arg in args if arg is not None and not isinstance(arg, bool)
```
georgesittas, line 448: Why would `arg` be either `None` or `bool` here? Is this copy-pasted from `format_args`? What not reuse that?

tests/dialects/test_oracle.py
```diff
@@ -329,6 +329,34 @@ def test_hints(self):
         )
         self.validate_identity("INSERT /*+ APPEND */ INTO IAP_TBL (id, col1) VALUES (2, 'test2')")
         self.validate_identity("INSERT /*+ APPEND_VALUES */ INTO dest_table VALUES (i, 'Value')")
+        self.validate_identity(
+            "SELECT /*+ LEADING(departments employees) USE_NL(employees) */ * FROM employees JOIN departments ON employees.department_id = departments.department_id",
+            write_sql="""SELECT /*+ LEADING(departments employees)
```
georgesittas, line 334: [Nit] we don't usually specify `write_sql`, but rather set the arg as a positional one.

georgesittas (commented):
Nice work! Just a few comments, but looks good.

sqlglot/dialects/oracle.py
mkmoisen: @georgesittas Yes this was copy-pasted from format_args.

We cannot reuse it as is because `format_args` joins with `', '`/`',\n'` instead of `' '`/`'\n'`.

Actually my code has a bug, it will pretty print with comma instead of space when the hint function arguments are too long.

I've fixed the bug and added a test case, as well as changed the args_sql to `arg_sqls = tuple(self.sql(arg) for arg in args)`

sqlglot/dialects/oracle.py
mkmoisen: @georgesittas  please check the latest commit. I believe I have the spacing correct - just let me know if you prefer a change.

sqlglot/dialects/oracle.py
georgesittas: I see, I may try to refactor this later so that we can consolidate the two implementations somehow.

georgesittas (approved):
Great work, thanks so much for the PR!

georgesittas:
Take a look at https://github.com/tobymao/sqlglot/commit/41e2eba1a01c1a5b784ad9dc6c5191f3d3bc0d74

Issue #2331, opened 2023-09-27
STRING_AGG wrongly being converted to GROUP_CONCAT in parse_one


SudarshanVS:
sqlglot==13.3.0

```
SELECT
  STRING_AGG(email, '') AS STRING_AGGROwner_Id_Email
FROM organizations
```

The above query gets converted to 

```
SELECT
  GROUP_CONCAT(owner_id_email, '') AS emails
FROM organizations
```

When passed through the below code,
```
query = """SELECT
  STRING_AGG(email, '') AS emails
FROM organizations"""
expression_tree = parse_one(query, read='trino')
```

But `GROUP_CONCAT` is not a valid TRINO function (STRING_AGG is also not a valid TRINO function). The correct equivalent function is `ARRAY_JOIN(ARRAY_AGG(email),',')`

Can this be fixed?

PR #4253, opened 2024-10-16
feat(trino): Support for LISTAGG function

### LINKED ISSUE #4250: Issue in generating SQL for AST having LISTAGG function for Trino SQL
If we have a Trino SQL such as - 

```sql
SELECT LISTAGG(DISTINCT "tbl1_1"."id", ',') WITHIN GROUP (ORDER BY "tbl1_1"."id" ASC)
            FROM table_a "tbl1_1"
```

when we parse it and convert it back to SQL, the SQL being returned is (instead of the original one) - 

```sql
SELECT LISTAGG(DISTINCT CASE WHEN "tbl1_1"."id" IS NULL THEN NULL WHEN ',' IS NULL THEN NULL ELSE ("tbl1_1"."id", ',') END) WITHIN GROUP (ORDER BY "tbl1_1"."id" ASC) FROM table_a AS "tbl1_1"
```

The documentation for `LISTAGG` is [here](https://trino.io/docs/current/functions/aggregate.html#listagg).

Snippet for reproducing the issue - 

```python
from sqlglot import parse

if __name__ == '__main__':
    sql_str = """
    SELECT LISTAGG(DISTINCT "tbl1_1"."id", ',') WITHIN GROUP (ORDER BY "tbl1_1"."id" ASC)
            FROM table_a "tbl1_1"
    """
    expression_tree = parse(sql_str, read="trino")
    print(repr(expression_tree))
    print(expression_tree[0].sql(dialect="trino"))
```

VaggelisD:
Fixes #4250

Before this PR, Trino would transpile `exp.GroupConcat` into `ARRAY_JOIN(ARRAY_AGG(...))`, which was added in https://github.com/tobymao/sqlglot/issues/2331. For Presto, this [was the recommended way](https://stackoverflow.com/questions/44142356/presto-equivalent-of-mysql-group-concat).

However, at some point `LISTAGG` [was supported](https://github.com/trinodb/trino/issues/4835) :

```
LISTAGG( expression [, separator] [ON OVERFLOW overflow_behaviour])
    WITHIN GROUP (ORDER BY sort_item, [...]) [FILTER (WHERE condition)]
```

An example:

```SQL
trino> WITH t AS (SELECT 'a' AS id UNION ALL SELECT 'b' AS id) SELECT LISTAGG(DISTINCT id, ',') WITHIN GROUP (ORDER BY id ASC) FROM t;
 _col0
-------
 a,b
(1 row)
```

This PR adds support for `LISTAGG` by:
- Reusing `parser.py::_parse_string_agg() -> exp.GroupConcat`, as it's parsing identical syntax
- Generating `LISTAGG` if there's an `exp.Order` child under it, otherwise defaults back to the existing transpilation

Docs
-------
[Trino LISTAGG](https://trino.io/docs/current/functions/aggregate.html#listagg)

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -386,9 +386,7 @@ class Presto(Dialect):
             exp.GenerateSeries: sequence_sql,
             exp.GenerateDateArray: sequence_sql,
             exp.Group: transforms.preprocess([transforms.unalias_group]),
-            exp.GroupConcat: lambda self, e: self.func(
-                "ARRAY_JOIN", self.func("ARRAY_AGG", e.this), e.args.get("separator")
-            ),
+            exp.GroupConcat: lambda self, e: self.groupconcat_sql(e),
             exp.If: if_sql(),
             exp.ILike: no_ilike_sql,
             exp.Initcap: _initcap_sql,
@@ -680,3 +678,10 @@ class Presto(Dialect):
             expr = "".join(segments)
 
             return f"{this}{expr}"
+
+        def groupconcat_sql(self, expression: exp.Expression) -> str:
+            return self.func(
+                "ARRAY_JOIN",
+                self.func("ARRAY_AGG", expression.this),
+                expression.args.get("separator"),
+            )
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -386,7 +386,6 @@ class Presto(Dialect):
             exp.GenerateSeries: sequence_sql,
             exp.GenerateDateArray: sequence_sql,
             exp.Group: transforms.preprocess([transforms.unalias_group]),
-            exp.GroupConcat: lambda self, e: self.groupconcat_sql(e),
             exp.If: if_sql(),
             exp.ILike: no_ilike_sql,
             exp.Initcap: _initcap_sql,
@@ -679,7 +678,7 @@ class Presto(Dialect):
 
             return f"{this}{expr}"
 
-        def groupconcat_sql(self, expression: exp.Expression) -> str:
+        def groupconcat_sql(self, expression: exp.GroupConcat) -> str:
             return self.func(
                 "ARRAY_JOIN",
                 self.func("ARRAY_AGG", expression.this),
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/presto.py
```diff
@@ -680,3 +678,10 @@ def jsonextract_sql(self, expression: exp.JSONExtract) -> str:
             expr = "".join(segments)
 
             return f"{this}{expr}"
+
+        def groupconcat_sql(self, expression: exp.Expression) -> str:
```
georgesittas, line 682:
```suggestion
        def groupconcat_sql(self, expression: exp.GroupConcat) -> str:
```

sqlglot/dialects/presto.py
```diff
@@ -386,9 +386,7 @@ class Generator(generator.Generator):
             exp.GenerateSeries: sequence_sql,
             exp.GenerateDateArray: sequence_sql,
             exp.Group: transforms.preprocess([transforms.unalias_group]),
-            exp.GroupConcat: lambda self, e: self.func(
-                "ARRAY_JOIN", self.func("ARRAY_AGG", e.this), e.args.get("separator")
-            ),
+            exp.GroupConcat: lambda self, e: self.groupconcat_sql(e),
```
georgesittas, line 389: Do we actually need this?

georgesittas (commented):
Does this actually address the problem? I thought we had some issues around the `CASE..WHEN` transformation?

VaggelisD:
> Does this actually address the problem? I thought we had some issues around the CASE..WHEN transformation?

Turns out that `LISTAGG` was getting parsed as an `exp.Anonymous` and the args would be parsed into the `expressions` part; This meant that the generator would consider this as a multi-arg DISTINCT node, which is not the case.

After this PR, the `DISTINCT col` and the separator are parsed separately so the `CASE ... WHEN` does not kick in.

sqlglot/dialects/presto.py
VaggelisD: Doesn't seem like it actually

georgesittas (approved):

PR #4256, opened 2024-10-17
Feat(starrocks): support for ALTER TABLE SWAP WITH

mrhamburg:
Similar to snowflake, see: #2420


georgesittas (commented):
Let's please factor these methods out into the base classes so they're not duplicated (for Snowflake, too). LGTM otherwise, thanks! 👍

georgesittas:
Will just take this to the finish line, so no worries.

georgesittas:
See https://github.com/tobymao/sqlglot/commit/cfd692ff28a59f413671aafbc8dcd61eab3558c3

PR #4258, opened 2024-10-17
Fix(clickhouse)!: dont parse right-hand side operands of ARRAY JOIN as Tables

### LINKED ISSUE #4254: Incorrect recognized ARRAY JOIN (Clickhouse)
I'm trying to get a list of all the tables
```python
sql = """
    SELECT s, arr FROM arrays_test 
    LEFT ARRAY JOIN arr
"""
expression = parse_one(sql, dialect=Dialects.CLICKHOUSE)
print(list(expression.find_all(exp.Table)))
```
Output:
```python
[Table(
   this=Identifier(this=arrays_test, quoted=False)),
 Table(
   this=Identifier(this=arr, quoted=False))]
   ```
But `arr` is a column name
Docs: https://clickhouse.com/docs/en/sql-reference/statements/select/array-join#basic-array-join-examples

### PR DESCRIPTION
**Pull request by georgesittas**
Fixes #4254

Reference: https://clickhouse.com/docs/en/sql-reference/statements/select/array-join

Screenshot 2024-10-17 at 3 20 48 PM
Screenshot 2024-10-17 at 3 20 59 PM

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -603,6 +603,22 @@ class ClickHouse(Dialect):
             if join:
                 join.set("global", join.args.pop("method", None))
 
+                # tbl ARRAY JOIN arr <-- this should be a `Column` reference, not a `Table`
+                # https://clickhouse.com/docs/en/sql-reference/statements/select/array-join
+                if join.kind == "ARRAY":
+                    for table in join.find_all(exp.Table):
+                        if not table.db:
+                            this = table.this
+                            if isinstance(this, exp.Identifier):
+                                this = exp.column(this)
+
+                            # If they RHS is aliased, we'll produce `Alias(Column(...))`
+                            alias = table.args.get("alias")
+                            if isinstance(alias, exp.TableAlias):
+                                this = exp.alias_(this, alias.this)
+
+                            table.replace(this)
+
             return join
 
         def _parse_function(
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -607,17 +607,7 @@ class ClickHouse(Dialect):
                 # https://clickhouse.com/docs/en/sql-reference/statements/select/array-join
                 if join.kind == "ARRAY":
                     for table in join.find_all(exp.Table):
-                        if not table.db:
-                            this = table.this
-                            if isinstance(this, exp.Identifier):
-                                this = exp.column(this)
-
-                            # If they RHS is aliased, we'll produce `Alias(Column(...))`
-                            alias = table.args.get("alias")
-                            if isinstance(alias, exp.TableAlias):
-                                this = exp.alias_(this, alias.this)
-
-                            table.replace(this)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: sqlglot/dialects/clickhouse.py [hunk 1]
```diff
@@ -603,6 +603,22 @@ def _parse_join(
             if join:
                 join.set("global", join.args.pop("method", None))
 
+                # tbl ARRAY JOIN arr <-- this should be a `Column` reference, not a `Table`
+                # https://clickhouse.com/docs/en/sql-reference/statements/select/array-join
+                if join.kind == "ARRAY":
+                    for table in join.find_all(exp.Table):
+                        if not table.db:
```
VaggelisD, line 610: What case is this branch protecting against? Afaict such queries seem valid, where `table.db` is the actual table name

```
:) with t as (select [1, 2, 3] as col, 1 as id) select * from t array join t.col;

WITH t AS
    (
        SELECT
            [1, 2, 3] AS col,
            1 AS id
    )
SELECT *
FROM t
ARRAY JOIN t.col

Query id: 4cef17fc-03e8-4644-a87d-c50f22055034

   ┌─col─┬─id─┐
1. │   1 │  1 │
2. │   2 │  1 │
3. │   3 │  1 │
   └─────┴────┘
```

sqlglot/expressions.py
```diff
@@ -3214,10 +3214,18 @@ def parts(self) -> t.List[Expression]:
...
     def to_column(self, copy: bool = True) -> Alias | Column | Dot:
         parts = self.parts
-        col = column(*reversed(parts[0:4]), fields=parts[4:], copy=copy)  # type: ignore
+        last_part = parts[-1]
+
+        if isinstance(last_part, Identifier):
+            col = column(*reversed(parts[0:4]), fields=parts[4:], copy=copy)  # type: ignore
+        else:
+            # This branch will be reached if a function or array is wrapped in a `Table`
+            col = last_part
```
VaggelisD, line 3223: What will `last_part` be in this case?

VaggelisD (approved):

sqlglot/expressions.py
georgesittas: Could be a `Func` instance (`SELECT * FROM ... LEFT ARRAY JOIN FOO() ...`), or an `Array`, e.g.:

```python
>>> parse_one("select * from t left array join [1, 2, 3] as arr", "clickhouse")
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=t, quoted=False))),
  joins=[
    Join(
      this=Table(
        this=Array(
          expressions=[
            Literal(this=2, is_string=False),
            Literal(this=3, is_string=False)]),
        alias=TableAlias(
          this=Identifier(this=arr, quoted=False))),
      side=left,
      kind=array)])
```

pkit:
Nice! Just wanted to open an Issue on that!

pkit:
@georgesittas when will it release?

georgesittas:
Just released v25.26.0, should be live in a couple of minutes.

PR #4260, opened 2024-10-17
feat(tsql): Support for stored procedure options

rsanchez-xtillion:
Added support for transact sql stored procedure options. Example:

```
CREATE PROCEDURE dbo.usp_add_kitchen @dept_id INT, @kitchen_count INT NOT NULL
WITH EXECUTE AS OWNER, SCHEMABINDING, NATIVE_COMPILATION -- These are stored procedure options
AS
BEGIN ATOMIC WITH (TRANSACTION ISOLATION LEVEL = SNAPSHOT, LANGUAGE = N'us_english')

UPDATE dbo.Departments
SET kitchen_count = ISNULL(kitchen_count, 0) + @kitchen_count
WHERE ID = @dept_id
END;
```

I added support by:
 - Adding a new `WithStoredProcedureOptions` property expression
 - Added `OPTIONS_TYPE` for the possible stored procedure options and for the `EXECUTE AS` clause
 - Implemented `_parse_function_parameter` in the TSQL dialect
 - Added if condition in the property parser to parse the stored procedure options


georgesittas:
Hi @rsanchez-xtillion, is this ready for review? What's pending?

rsanchez-xtillion:
It is ready for review now, I was making sure the checks passed and added a description.

georgesittas:
Appreciate it, thanks. Will take a look.

sqlglot/dialects/tsql.py
```diff
@@ -719,9 +719,19 @@ def _parse_user_defined_function(
             ):
                 return this
 
-            expressions = self._parse_csv(self._parse_function_parameter)
+            expressions = None
+            if self._curr.token_type != TokenType.WITH:
+                expressions = self._parse_csv(self._parse_function_parameter)
```
georgesittas, line 724:
```suggestion
            if self._match(TokenType.WITH, advance=False):
                expressions = self._parse_csv(self._parse_function_parameter)
            else:
                expressions = None
```

sqlglot/dialects/tsql.py
```diff
@@ -719,9 +719,19 @@ def _parse_user_defined_function(
...
+
             return self.expression(exp.UserDefinedFunction, this=this, expressions=expressions)
 
+        def _parse_function_parameter(self) -> t.Optional[exp.Expression]:
+            id_var = self._parse_id_var()
+            kind = self._parse_types(schema=True)
+            if not kind:
+                return id_var
+            return self.expression(exp.ColumnDef, this=id_var, kind=kind)
```
georgesittas, line 733: What triggered this override? It looks pretty similar to the base implementation.

sqlglot/generator.py
```diff
@@ -3622,7 +3624,7 @@ def userdefinedfunction_sql(self, expression: exp.UserDefinedFunction) -> str:
         expressions = (
             self.wrap(expressions) if expression.args.get("wrapped") else f" {expressions}"
         )
-        return f"{this}{expressions}"
+        return f"{this}{expressions}" if expressions.strip() != "" else this
```
georgesittas, line 3627: Why do we need `strip()` here? Should we instead modify some of the previous lines to make sure `expressions` is an empty string or something? (That's a common pattern in the `Generator` methods.)

sqlglot/parser.py
```diff
@@ -2203,11 +2209,26 @@ def _parse_with_property(self) -> t.Optional[exp.Expression] | t.List[exp.Expres
...
 
+    def _parse_procedure_option(self) -> exp.Expression | None:
+        if self._match_text_seq("EXECUTE", "AS"):
+            return self.expression(
+                exp.ExecuteAsProperty,
+                this=self._parse_var_from_options(self.EXECUTE_AS_OPTIONS, raise_unmatched=False)
+                or self._parse_string(),
+            )
+        else:
+            return self._parse_var_from_options(self.PROCEDURE_OPTIONS)
```
georgesittas, line 2230:
```suggestion
        return self._parse_var_from_options(self.PROCEDURE_OPTIONS)
```

georgesittas (commented):
Looks good, thanks a lot for the PR. Just a few minor comments to refactor this a bit.

sqlglot/dialects/tsql.py
rsanchez-xtillion: I'll make this change

sqlglot/parser.py
rsanchez-xtillion: I'll make this change

sqlglot/dialects/tsql.py
rsanchez-xtillion: This method in the Parser class is:
```
def _parse_function_parameter(self) -> t.Optional[exp.Expression]:
        return self._parse_column_def(self._parse_id_var())
```
Inside of the `_parse_column_def` method, `_parse_column_constraint` is called. This method parses the tokens as column constraints if they match the `CONSTRAINT_PARSERS` of the Parser class. One of these constraint parsers is for the `WITH` keyword, so if I leave the original implementation, the parser tries to parse procedure options as column constraints and eventually raises an error. That is why I gave TSQL its own `_parse_function_parameter` implementation.

I could replace the implementation of `_parse_function_parameter` in TSQL with an implementation of `_parse_column_constraint` that checks that the current token is not `WITH`, like so: 

```
def _parse_column_constraint(self) -> t.Optional[exp.Expression]:
    if self._match(TokenType.CONSTRAINT):
        this = self._parse_id_var()
    else:
        this = None

    # Check that WITH does not match and then keep going
    if not self._match(TokenType.WITH, advance=False) and self._match_texts(self.CONSTRAINT_PARSERS):
        return self.expression(
            exp.ColumnConstraint,
            this=this,
            kind=self.CONSTRAINT_PARSERS[self._prev.text.upper()](self),
        )

    return this
```

sqlglot/generator.py
rsanchez-xtillion: In the edge case that a procedure had no parameters but did have procedure options, like:
```
CREATE PROCEDURE foo WITH EXECUTE AS OWNER AS BEGIN SELECT 1
```
When I parsed the sql and printed it again from the AST, there was an extra whitespace character in between the procedure name and the WITH keyword like so:
```
CREATE PROCEDURE foo  WITH EXECUTE AS OWNER AS BEGIN SELECT 1
```
Since the `expressions` variable inside the `userdefinedfunction_sql` contains the parameters of the procedure, when there are no parameters, the variable ended up being just a whitespace, so that is why I added `strip()`. I could look a little bit more into it and find a better fix.

sqlglot/dialects/tsql.py
rsanchez-xtillion: The line `if self._match(TokenType.WITH, advance=False):` is missing a `not` since I want to check that the current token is not WITH. I will add that on my next push.

sqlglot/generator.py
georgesittas: That's fine, thanks for providing this context 👍

sqlglot/dialects/tsql.py
rsanchez-xtillion: I'm not sure my override of `_parse_function_parameter` in the TSQL class respects the spec even if all the tests pass. That is why I'm also thinking of overriding `_parse_column_constraint` instead, since that would preserve the functionality of the original implementation of `_parse_function_parameter` in the Parser class.

sqlglot/dialects/tsql.py
```diff
@@ -719,9 +719,30 @@ def _parse_user_defined_function(
...
             return self.expression(exp.UserDefinedFunction, this=this, expressions=expressions)
 
+        def _parse_column_constraint(self) -> t.Optional[exp.Expression]:
+            if self._match(TokenType.CONSTRAINT):
+                this = self._parse_id_var()
+            else:
+                this = None
+
+            if not self._match(TokenType.WITH, advance=False) and self._match_texts(
+                self.CONSTRAINT_PARSERS
```
georgesittas, line 736: This change looks more intrusive than necessary imho. I think the safer approach would be to check if `WITH` is followed by a token that signifies a procedure property instead of cutting this whole branch/logic off.

sqlglot/dialects/tsql.py
```diff
@@ -719,9 +719,30 @@ def _parse_user_defined_function(
...
+            else:
+                expressions = None
+
             return self.expression(exp.UserDefinedFunction, this=this, expressions=expressions)
 
+        def _parse_column_constraint(self) -> t.Optional[exp.Expression]:
+            if self._match(TokenType.CONSTRAINT):
+                this = self._parse_id_var()
+            else:
+                this = None
```
georgesittas, line 733: We can rewrite this (both here & in the original method) into:

```python
this = self._match(TokenType.CONSTRAINT) and self._parse_id_var()
```

sqlglot/dialects/tsql.py
georgesittas: As long as none of the "follow" keywords clash with existing constraint keywords, it should be fine to put this in the base parser and control the logic using a constant, e.g. something like:

```python
# Set this in T-SQL, and use an empty set for the base class, unless there's no collision with existing "WITH" constraints
PROCEDURE_WITH_PROPERTY_FOLLOW_KEYWORDS = {
    "ENCRYPTION",
    "EXECUTE",
    ...
}

def _parse_column_constraint(self) -> t.Optional[exp.Expression]:
    this = self._match(TokenType.CONSTRAINT) and self._parse_id_var()

    if (
        not self._match(TokenType.WITH)
        or not self._next
        or self._next.token_type not in self.PROCEDURE_WITH_PROPERTY_FOLLOW_KEYWORDS
    ):
        return self.expression(
            exp.ColumnConstraint,
            this=this,
            kind=self.CONSTRAINT_PARSERS[self._prev.text.upper()](self),
        )

    self._advance(self._index - 1)  # Unconsume the WITH token
    return this
```

sqlglot/parser.py
```diff
@@ -5571,12 +5569,15 @@ def _parse_not_constraint(self) -> t.Optional[exp.Expression]:
...
-            this = self._parse_id_var()
-        else:
-            this = None
+        this = self._match(TokenType.CONSTRAINT) and self._parse_id_var()
+
+        index = self._index
+        procedure_property_follows = self._match(TokenType.WITH) and self._match_texts(
+            self.PROCEDURE_OPTIONS
+        )
+        self._retreat(index)
```
georgesittas, line 5578: Let's retreat conditionally, only if the flag is true

georgesittas (approved):

sqlglot/parser.py
rsanchez-xtillion: I need to keep the index where it is because I need `WITH` to be the current token when I get to the property parsers where I parse the procedure options. That I why I was always retreating. 

I could rewrite the condition as:
```
procedure_property_follows = self._match(TokenType.WITH, advance=False) and self._next.text.upper() in self.PROCEDURE_OPTIONS
```
That way I avoid having to advance on matches and retreat. I didn't do it like that at first because I preferred using `self._match_texts(self.PROCEDURE_OPTIONS)` instead of `self._next.text.upper() in self.PROCEDURE_OPTIONS`.

sqlglot/dialects/tsql.py
rsanchez-xtillion: I decided to define PROCEDURE_OPTIONS in the Parser class as an empty OPTIONS_TYPE, and assign it the corresponding values in the TSQL class

sqlglot/parser.py
```diff
@@ -1225,6 +1225,10 @@ class Parser(metaclass=_Parser):
         **dict.fromkeys(("BINDING", "COMPENSATION", "EVOLUTION"), tuple()),
     }
 
+    PROCEDURE_OPTIONS: OPTIONS_TYPE = {}
+
+    EXECUTE_AS_OPTIONS: OPTIONS_TYPE = dict.fromkeys(("CALLER", "SELF", "OWNER"), tuple())
```
georgesittas, line 1230: Curious why these were left in the base class whereas `PROCEDURE_OPTIONS` where moved to T-SQL?

sqlglot/parser.py
```diff
@@ -5550,12 +5569,14 @@ def _parse_not_constraint(self) -> t.Optional[exp.Expression]:
...
     def _parse_column_constraint(self) -> t.Optional[exp.Expression]:
-        if self._match(TokenType.CONSTRAINT):
-            this = self._parse_id_var()
-        else:
-            this = None
+        this = self._match(TokenType.CONSTRAINT) and self._parse_id_var()
+
+        procedure_property_follows = (
+            self._match(TokenType.WITH, advance=False)
+            and self._next.text.upper() in self.PROCEDURE_OPTIONS
```
georgesittas, line 5576:
```suggestion
            and self._next
            and self._next.text.upper() in self.PROCEDURE_OPTIONS
```

sqlglot/parser.py
rsanchez-xtillion: I left them in the base class because that is where the parsing is happening. I could also just move everything to TSQL, but I was also planning on giving support to Snowflake stored procedure `EXECUTE AS` which is almost the same, so it would be useful to have that available on the parent class since I would be dealing with more than one dialect. Let me know what you think.

georgesittas: Sounds good, thanks for clarifying.

georgesittas (approved):
Great work @rsanchez-xtillion, thank you for your patience & appreciate the quick iteration.

PR #4263, opened 2024-10-17
clickhouse: fix traverse_scope with FINAL modifier

### LINKED ISSUE #4262: Empty 'sources' in scope when using `FINAL` modifier in ClickHouse SQL query
```python
import sqlglot
from sqlglot.optimizer import traverse_scope

exp = sqlglot.parse_one('SELECT * FROM x FINAL', read="clickhouse")
scopes = traverse_scope(exp)
print(len(scopes[0].sources))
```

pkit:
Fixes #4262


georgesittas (approved):
Thanks for the patch!

PR #4267, opened 2024-10-18
fix(optimizer)!: Add type hints for optimizer rules eliminate & merge subqueries

VaggelisD:
Follow up on https://github.com/tobymao/sqlglot/pull/4266

### DIFF
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -8,10 +8,8 @@ from sqlglot.helper import find_new_name
 from sqlglot.optimizer.scope import Scope, build_scope
 
 if t.TYPE_CHECKING:
-    from sqlglot._typing import E
-
-    ExistingCTEsMapping = t.Dict[E, str]
-    TakenNameMapping = t.Dict[str, t.Union[Scope, E]]
+    ExistingCTEsMapping = t.Dict[exp.Expression, str]
+    TakenNameMapping = t.Dict[str, t.Union[Scope, exp.Expression]]
 
 
 def eliminate_subqueries(expression: exp.Expression) -> exp.Expression:
--- a/sqlglot/optimizer/eliminate_subqueries.py
+++ b/sqlglot/optimizer/eliminate_subqueries.py
@@ -1,11 +1,20 @@
+from __future__ import annotations
+
 import itertools
+import typing as t
 
 from sqlglot import expressions as exp
 from sqlglot.helper import find_new_name
-from sqlglot.optimizer.scope import build_scope
+from sqlglot.optimizer.scope import Scope, build_scope
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
+
+    ExistingCTEsMapping = t.Dict[E, str]
+    TakenNameMaping = t.Dict[t.Any, t.Union[Scope, E]]
 
 
-def eliminate_subqueries(expression):
+def eliminate_subqueries(expression: exp.Expression) -> exp.Expression:
     """
     Rewrite derived tables as CTES, deduplicating if possible.
 
@@ -38,7 +47,7 @@ def eliminate_subqueries(expression):
     # Map of alias->Scope|Table
     # These are all aliases that are already used in the expression.
     # We don't want to create new CTEs that conflict with these names.
-    taken = {}
+    taken: TakenNameMaping = {}
 
     # All CTE aliases in the root scope are taken
     for scope in root.cte_scopes:
@@ -56,7 +65,7 @@ def eliminate_subqueries(expression):
 
     # Map of Expression->alias
     # Existing CTES in the root expression. We'll use this for deduplication.
-    existing_ctes = {}
+    existing_ctes: ExistingCTEsMapping = {}
 
     with_ = root.expression.args.get("with")
     recursive = False
@@ -95,7 +104,7 @@ def eliminate_subqueries(expression):
     return expression
 
 
-def _eliminate(scope, existing_ctes, taken):
+def _eliminate(scope: Scope, existing_ctes: ExistingCTEsMapping, taken: TakenNameMaping):
     if scope.is_derived_table:
         return _eliminate_derived_table(scope, existing_ctes, taken)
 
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,11 +1,20 @@
...
 from sqlglot import expressions as exp
 from sqlglot.helper import find_new_name
-from sqlglot.optimizer.scope import build_scope
+from sqlglot.optimizer.scope import Scope, build_scope
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
+
+    ExistingCTEsMapping = t.Dict[E, str]
+    TakenNameMaping = t.Dict[t.Any, t.Union[Scope, E]]
```
georgesittas, line 14: There's a typo here, it needs to be "Mapping". Additionally, why are we using `t.Any` for the key type?

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,11 +1,20 @@
...
 
 
-def eliminate_subqueries(expression):
+def eliminate_subqueries(expression: exp.Expression) -> exp.Expression:
```
georgesittas, line 17: I believe we want this to be `E -> E` so that the argument's type matches the returned one. For example, if you passed a `Query` to this and invoked `selects` on the output of the rule, mypy would complain because `selects` doesn't appear on the base `Expression` class.

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -95,15 +104,17 @@ def eliminate_subqueries(expression):
     return expression
 
 
-def _eliminate(scope, existing_ctes, taken):
+def _eliminate(scope: Scope, existing_ctes: ExistingCTEsMapping, taken: TakenNameMaping):
```
georgesittas, line 107: This needs a return type.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -1,11 +1,18 @@
...
 from sqlglot import expressions as exp
 from sqlglot.helper import find_new_name
 from sqlglot.optimizer.scope import Scope, traverse_scope
 
+if t.TYPE_CHECKING:
+    FromOrJoin = t.Union[exp.From, exp.Join]
+
 
-def merge_subqueries(expression, leave_tables_isolated=False):
+def merge_subqueries(expression: exp.Expression, leave_tables_isolated=False):
```
georgesittas, line 15: Needs return type & `bool` for the flag, see comment about `E -> E`.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -58,7 +65,7 @@ def merge_subqueries(expression, leave_tables_isolated=False):
 )
 
 
-def merge_ctes(expression, leave_tables_isolated=False):
+def merge_ctes(expression: exp.Expression, leave_tables_isolated=False) -> exp.Expression:
```
georgesittas, line 68: Ditto about the `E -> E` comment.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -92,7 +99,7 @@ def merge_ctes(expression, leave_tables_isolated=False):
     return expression
 
 
-def merge_derived_tables(expression, leave_tables_isolated=False):
+def merge_derived_tables(expression: exp.Expression, leave_tables_isolated=False):
```
georgesittas, line 102: Here too, let's just do another good pass on these, I'm assuming the same may have happened elsewhere.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -195,7 +196,7 @@ def _is_recursive():
         and not outer_scope.expression.is_star
         and isinstance(inner_select, exp.Select)
         and not any(inner_select.args.get(arg) for arg in UNMERGABLE_ARGS)
-        and inner_select.args.get("from")
+        and inner_select.args.get("from") is not None
```
georgesittas, line 199: What triggered this change?

georgesittas (commented):
Needs some polishing but LGTM otherwise. Let's mark this as breaking so that folks using these rules won't suddenly get cascading mypy issues.

sqlglot/optimizer/eliminate_subqueries.py
VaggelisD: Hm, was sure that it was complaining about a `t.Any` entry somewhere but I changed it back to `str` and all seems good

sqlglot/optimizer/eliminate_subqueries.py
VaggelisD: I think most of the pattern around the codebase does not use `E`, right? For example, `simplify` is such a case.

However, your point makes sense regardless afaict, will revert to `E`

georgesittas: You can't just generalize my argument for those other rules. For example, `simplify` may look similar but it doesn't always preserve the input type in its output, so you can't use `E` there:

```python
>>>
>>> simplify(parse_one("1 + 2"))  # Gets an `Add`, returns a `Literal`.
Literal(this=3, is_string=False)
```

But I wouldn't be surprised if we're doing the same thing for other rules where the `E -> E` logic applies, so perhaps if you have cycles it's worth taking a look to refine them in a followup PR.

sqlglot/optimizer/merge_subqueries.py
VaggelisD: Iirc the first variant bubbles `t.Any` up the conjuction chain, so mypy complains that the return type is `t.Any | bool`, with this change it's silenced

georgesittas: Ah I see, sounds good.

georgesittas (approved):

sqlglot/optimizer/eliminate_subqueries.py
```diff
@@ -1,11 +1,20 @@
...
 from sqlglot import expressions as exp
 from sqlglot.helper import find_new_name
-from sqlglot.optimizer.scope import build_scope
+from sqlglot.optimizer.scope import Scope, build_scope
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
 
+    ExistingCTEsMapping = t.Dict[E, str]
+    TakenNameMapping = t.Dict[str, t.Union[Scope, E]]
```
georgesittas, line 14: Why are we using `E` here? I believe we want `exp.Expression` instead.

sqlglot/optimizer/merge_subqueries.py
```diff
@@ -1,11 +1,20 @@
...
 from sqlglot.optimizer.scope import Scope, traverse_scope
 
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
+
+    FromOrJoin = t.Union[exp.From, exp.Join]
 
-def merge_subqueries(expression, leave_tables_isolated=False):
+
+def merge_subqueries(expression: E, leave_tables_isolated=False) -> E:
```
georgesittas, line 17:
```suggestion
def merge_subqueries(expression: E, leave_tables_isolated: bool = False) -> E:
```

PR #4269, opened 2024-10-18
Fix(clickhouse)!: time string literals containing fractional seconds

treysp:
Clickhouse has two datetime types:
- `DateTime`: fixed one second precision
- `DateTime64`: accepts user-specified precision (default of 3 when no precision specified)

Currently, time string literals are always generated by casting to `DateTime`, causing Clickhouse to error if the string contains fractional seconds.

With this PR, time string literals are always cast to `DateTime64(6)`. The returned data type is non-nullable because they may be used in contexts where nullable types cause errors (like an `ORDER BY` clause).

Implementation note:
- Clickhouse does not accept timestamps that include a UTC offset, so we remove it if present (implemented in #3938)
- The first step of removal is parsing with `datetime.datetime.fromisoformat`
- In Python <3.11, `fromisoformat` only accepts timestamps with millisecond (3 digit) or microsecond (6 digit) precision
- We ensure there will always be 6 digits of precision by manually padding any fractional seconds to 6 digits with 0s


sqlglot/dialects/clickhouse.py
```diff
@@ -108,23 +109,54 @@ def _delta_sql(self: Generator, expression: DATEΤΙΜΕ_DELTA) -> str:
...
+        ts_parts = ts_string.split(".")
+        if len(ts_parts) == 2:
+            # separate fractional seconds and UTC offset
+            offset_sep = "+" if "+" in ts_parts[1] else "-"
+            ts_frac_parts = ts_parts[1].split(offset_sep)
+
+            # pad to 6 digits if fractional seconds present
+            ts_frac_parts[0] = ts_frac_parts[0] + "0" * (
+                6 - (len(ts_frac_parts[0]) if len(ts_frac_parts[0]) < 6 else 6)
             )
```
georgesittas, line 131: Perhaps we can simplify this a bit?

```python
ts_frac_parts[0] = ts_frac_parts[0].ljust(6, '0')
```

sqlglot/dialects/clickhouse.py
```diff
@@ -108,23 +109,54 @@ def _delta_sql(self: Generator, expression: DATEΤΙΜΕ_DELTA) -> str:
...
-            ts = exp.Literal.string(ts_without_tz)
+            ts_string = "".join(
+                [
+                    ts_parts[0],  # date and time
+                    ".",
+                    ts_frac_parts[0],  # fractional seconds
+                    offset_sep if len(ts_frac_parts) > 1 else "",
+                    ts_frac_parts[1] if len(ts_frac_parts) > 1 else "",  # utc offset (if present)
```
georgesittas, line 138: Consider storing `len(ts_frac_parts)` in a variable.

tests/dialects/test_clickhouse.py
```diff
@@ -1094,6 +1094,86 @@ def test_convert(self):
...
+        # no fractional seconds
+        time_strings = [
+            "2020-01-01 00:00:01",
+            "2020-01-01 00:00:01+01:00",
+            " 2020-01-01 00:00:01-01:00 ",
+            "2020-01-01T00:00:01+01:00",
+        ]
+        for time_string in time_strings:
+            self.assertEqual(
+                exp.TimeStrToTime(this=exp.Literal.string(time_string)).sql(dialect=self.dialect),
```
georgesittas, line 1129: May be worth doing `with self.subTest(...)` in these loops.

georgesittas (approved):

sqlglot/dialects/clickhouse.py
treysp: ah, much better - i didn't know about ljust

tests/dialects/test_clickhouse.py
treysp: i also didn't know about subTest - quite useful!

PR #4280, opened 2024-10-22
feat(snowflake)!: Add support for TIMESTAMP_NTZ_FROM_PARTS

VaggelisD:
This PR adds support for the 6-arg version of `TIMESTAMP_NTZ_FROM_PARTS(...)` by adding it as an alias for the already supported `TIMESTAMP_FROM_PARTS(...)`; This is because by default `TIMESTAMP_*` functions map to `TIMESTAMP_NTZ_*` unless changed by the user through `TIMESTAMP_TYPE_MAPPING`.

Three outstanding questions:
1. This family of functions also expose the 2-arg version `TIMESTAMP_FROM_PARTS( <date_expr>, <time_expr>)`. Should we add them in this PR too?
2. For future reference, to support the other variations as well such as `TIMESTAMP_LTZ_FROM_PARTS`, should we create separate nodes or add an `arg` that stores the variant type?
3. As an extension of (2) and a thought for the future, although it simplifies things now, we probably shouldn't tie `TIMESTAMP_NTZ` to `TIMESTAMP` across Snowflake as it won't be flexible enough for the users that have altered the `TIMESTAMP_TYPE_MAPPING`. Maybe we can introduce a dialect setting for Snowflake that attempts to mimic this behavior?

Docs
----------
[Snowflake TIMESTAMP_FROM_PARTS](https://docs.snowflake.com/en/sql-reference/functions/timestamp_from_parts) | [Snowflake TIMESTAMP_TYPE_MAPPING](https://docs.snowflake.com/en/sql-reference/parameters#label-timestamp-type-mapping)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -351,6 +351,8 @@ class Snowflake(Dialect):
             "TIMESTAMPDIFF": _build_datediff,
             "TIMESTAMPFROMPARTS": build_timestamp_from_parts,
             "TIMESTAMP_FROM_PARTS": build_timestamp_from_parts,
+            "TIMESTAMPNTZFROMPARTS": build_timestamp_from_parts,
+            "TIMESTAMP_NTZ_FROM_PARTS": build_timestamp_from_parts,
             "TRY_PARSE_JSON": lambda args: exp.ParseJSON(this=seq_get(args, 0), safe=True),
             "TRY_TO_DATE": _build_datetime("TRY_TO_DATE", exp.DataType.Type.DATE, safe=True),
             "TRY_TO_TIMESTAMP": _build_datetime(
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6488,12 +6488,7 @@ class Uuid(Func):
 
 
 class TimestampFromParts(Func):
-    _sql_names = [
-        "TIMESTAMP_FROM_PARTS",
-        "TIMESTAMPFROMPARTS",
-        "TIMESTAMP_NTZ_FROM_PARTS",
-        "TIMESTAMPNTZFROMPARTS",
-    ]
+    _sql_names = ["TIMESTAMP_FROM_PARTS", "TIMESTAMPFROMPARTS"]
     arg_types = {
         "year": True,
         "month": True,
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6488,7 +6488,12 @@ class Uuid(Func):
 
 
 class TimestampFromParts(Func):
-    _sql_names = ["TIMESTAMP_FROM_PARTS", "TIMESTAMPFROMPARTS"]
+    _sql_names = [
+        "TIMESTAMP_FROM_PARTS",
+        "TIMESTAMPFROMPARTS",
+        "TIMESTAMP_NTZ_FROM_PARTS",
+        "TIMESTAMPNTZFROMPARTS",
+    ]
     arg_types = {
         "year": True,
         "month": True,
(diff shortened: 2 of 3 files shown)


georgesittas (approved):
1. I wouldn't add them in this PR. I think I'd looked at these at some point and decided to treat them as `Anonymous` because they weren't easy to transpile.
2. I'd go with the arg approach instead of introducing new nodes because there's a lot of overlap.
3. The way to deal with this would be through dialect settings like you say, but we don't have to deal with it now, it's a separate concern.

VaggelisD:
> I wouldn't add them in this PR. I think I'd looked at these at some point and decided to treat them as Anonymous because they weren't easy to transpile.

I briefly looked at DuckDB and I think we could transpile these from Snowflake's `TIMESTAMP_FROM_PARTS(date, time)` to DDB's `date + time` as there's a relevant + operator for them:

```SQL
D select '2013-04-05'::DATE + '12:00:00'::TIME;
┌─────────────────────────────────────────────────────────┐
│ (CAST('2013-04-05' AS DATE) + CAST('12:00:00' AS TIME)) │
├─────────────────────────────────────────────────────────┤
│ 2013-04-05 12:00:00                                     │
└─────────────────────────────────────────────────────────┘

D select * From duckdb_functions() where function_name = '+' and return_type = 'TIMESTAMP';
┌───────────────┬──────────────┬─────────────┬───────────────┬───────────────┬─────────────┬─────────┬──────┬─────────────┬──────────────┬───────────────────────┬─────────┬──────────────────┬──────────────────┬──────────┬──────────────┬─────────┬────────────┐
│ database_name │ database_oid │ schema_name │ function_name │ function_type │ description │ comment │ tags │ return_type │  parameters  │    parameter_types    │ varargs │ macro_definition │ has_side_effects │ internal │ function_oid │ example │ stability  │
├───────────────┼──────────────┼─────────────┼───────────────┼───────────────┼─────────────┼─────────┼──────┼─────────────┼──────────────┼───────────────────────┼─────────┼──────────────────┼──────────────────┼──────────┼──────────────┼─────────┼────────────┤
│ system        │ 0            │ main        │ +             │ scalar        │             │         │ {}   │ TIMESTAMP   │ [col0, col1] │ [DATE, INTERVAL]      │         │                  │ false            │ true     │ 162          │         │ CONSISTENT │
│ system        │ 0            │ main        │ +             │ scalar        │             │         │ {}   │ TIMESTAMP   │ [col0, col1] │ [INTERVAL, DATE]      │         │                  │ false            │ true     │ 162          │         │ CONSISTENT │
│ system        │ 0            │ main        │ +             │ scalar        │             │         │ {}   │ TIMESTAMP   │ [col0, col1] │ [TIMESTAMP, INTERVAL] │         │                  │ false            │ true     │ 162          │         │ CONSISTENT │
│ system        │ 0            │ main        │ +             │ scalar        │             │         │ {}   │ TIMESTAMP   │ [col0, col1] │ [INTERVAL, TIMESTAMP] │         │                  │ false            │ true     │ 162          │         │ CONSISTENT │
│ system        │ 0            │ main        │ +             │ scalar        │             │         │ {}   │ TIMESTAMP   │ [col0, col1] │ [TIME, DATE]          │         │                  │ false            │ true     │ 162          │         │ CONSISTENT │
│ system        │ 0            │ main        │ +             │ scalar        │             │         │ {}   │ TIMESTAMP   │ [col0, col1] │ [DATE, TIME]          │         │                  │ false            │ true     │ 162          │         │ CONSISTENT │
└───────────────┴──────────────┴─────────────┴───────────────┴───────────────┴─────────────┴─────────┴──────┴─────────────┴──────────────┴───────────────────────┴─────────┴──────────────────┴──────────────────┴──────────┴──────────────┴─────────┴────────────┘
```

For the rest of the dialects, I guess we could have an `unsupported` generator call, but we can always revisit in a different PR if needed.

georgesittas:
I see, that sounds reasonable then. Wanna do it in a followup PR?

georgesittas (approved):

PR #4284, opened 2024-10-23
feat(bigquery)!: Add support for STRING function

VaggelisD:
This PR adds support for BQ's `STRING(...)` function, which turns input expressions into `STRING`s. There are two supported versions:
1. JSON based: `STRING(json_expr)`
2. TIMESTAMP based: `STRING(timestamp_expression[, time_zone])`

Both variants are transpiled to `CAST(... AS VARCHAR)` for all other dialects; For the (2) variant with `time_zone` supplied, the timestamp is passed through `exp.ConvertTimezone` for the timezone conversion:

```SQL
bigquery> SELECT STRING("2008-12-25 15:30:00", "America/New_York") AS string;
2008-12-25 10:30:00-05

snowflake> SELECT CAST(CONVERT_TIMEZONE('UTC', 'America/New_York', '2008-12-25 15:30:00') AS VARCHAR);
2008-12-25 10:30:00.000

duckdb> SELECT CAST(CAST('2008-12-25 15:30:00' AS TIMESTAMP) AT TIME ZONE 'UTC' AT TIME ZONE 'America/New_York' AS TEXT) AS color;
┌─────────────────────┐
│        color        │
│       varchar       │
├─────────────────────┤
│ 2008-12-25 10:30:00 │
└─────────────────────┘
```

Docs
-------
[BQ JSON STRING()](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#string_for_json) | [BQ TIMESTAMP STRING()](https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#string)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5282,6 +5282,11 @@ class ArrayToString(Func):
     _sql_names = ["ARRAY_TO_STRING", "ARRAY_JOIN"]
 
 
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#string
+class String(Func):
+    arg_types = {"this": True, "zone": False}
+
+
 class StringToArray(Func):
     arg_types = {"this": True, "expression": True, "null": False}
     _sql_names = ["STRING_TO_ARRAY", "SPLIT_BY_STRING"]
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -4434,3 +4434,15 @@ class Generator(metaclass=_Generator):
     @unsupported_args("format")
     def todouble_sql(self, expression: exp.ToDouble) -> str:
         return self.sql(exp.cast(expression.this, exp.DataType.Type.DOUBLE))
+
+    def string_sql(self, expression: exp.String) -> str:
+        this = expression.this
+        zone = expression.args.get("zone")
+
+        if zone:
+            # BigQuery stores timestamps internally as UTC, so it's used as a default source timezone
+            this = exp.ConvertTimezone(
+                source_tz=exp.Literal.string("UTC"), target_tz=zone, timestamp=this
+            )
+
+        return self.sql(exp.cast(this, exp.DataType.Type.VARCHAR))
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -4440,7 +4440,9 @@ class Generator(metaclass=_Generator):
         zone = expression.args.get("zone")
 
         if zone:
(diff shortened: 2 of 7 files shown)


sqlglot/expressions.py
```diff
@@ -5282,6 +5282,11 @@ class ArrayToString(Func):
     _sql_names = ["ARRAY_TO_STRING", "ARRAY_JOIN"]
 
 
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#string
+class String(Func):
+    arg_types = {"this": True, "zone": False}
```
georgesittas, line 5287: TODO: check whether this has unintended side effects, i.e. are there dialects that support `STRING(...)`? Spark definitely has a variant where `STRING(x)` is essentially `CAST(x AS STRING)`.

sqlglot/generator.py
```diff
@@ -4434,3 +4434,15 @@ def overlay_sql(self, expression: exp.Overlay):
     @unsupported_args("format")
     def todouble_sql(self, expression: exp.ToDouble) -> str:
         return self.sql(exp.cast(expression.this, exp.DataType.Type.DOUBLE))
+
+    def string_sql(self, expression: exp.String) -> str:
+        this = expression.this
+        zone = expression.args.get("zone")
+
+        if zone:
+            # BigQuery stores timestamps internally as UTC, so it's used as a default source timezone
```
georgesittas, line 4443: This comment may be somewhat confusing in the future, because this method is in the base generator. Can we do this in a source-dialect-agnostic way? Perhaps we can set a default zone at parse time?

sqlglot/expressions.py
VaggelisD: I did a pass over the main dialects, it's supported only for the Hive hierarchy as you mentioned. Will add test cases for that

sqlglot/generator.py
VaggelisD: This is not doable afaict, because the `zone` arg here is the target timezone while `UTC` is the source. Also, BigQuery being the only variant with `zone` should hopefully make it less confusing, I can add more comments there if needed.

georgesittas: Sounds good. Let's just point out that our current expectation is that this arg will only be set for BigQuery?

sqlglot/expressions.py
georgesittas: Nice, thanks for checking.

georgesittas (approved):

PR #4287, opened 2024-10-24
fix(clickhouse): Wrap subquery if it's LHS of IS NOT NULL

### LINKED ISSUE #4285: IS NOT NULL works incorrectly for subquery
SQLGlot parses `IS NOT NULL` to `NOT ... IS NULL` without parentheses. For example SQLGlot should parse like `NOT (... IS NULL)`

```python
from sqlglot import Dialects, parse_one

sql = """
    SELECT name
    FROM data 
    WHERE (SELECT DISTINCT name FROM data) IS NOT NULL
"""

print(parse_one(sql, read=Dialects.CLICKHOUSE).sql(dialect=Dialects.CLICKHOUSE))
```

❌ Wrong output with the version SQLGlot v25.27.0
```sql
SELECT name FROM data WHERE NOT (SELECT DISTINCT name FROM data) IS NULL
```

✅ Expected correct output
```sql
SELECT name FROM data WHERE NOT ((SELECT DISTINCT name FROM data) IS NULL)
```

VaggelisD:
Fixes #4285

Consider the query:

```SQL
WITH data AS (SELECT 'a' AS name) 
SELECT name FROM data 
WHERE (SELECT DISTINCT name FROM data) IS NOT NULL;
```

This will be roundtripped across all dialects as:

```SQL
WITH data AS (SELECT 'a' AS name) 
SELECT name FROM data 
WHERE NOT (SELECT DISTINCT name FROM data) IS NULL;
```

Which invalid for Clickhouse in particular. Wrapping the `IS` expression restores the functionality:

```SQL
ch> with data as (select 'a' as name) SELECT name FROM data WHERE NOT (SELECT DISTINCT name FROM data) IS NULL;

Expected one of: token, Dot, Comma, ClosingRoundBracket, OR, AND, IS NOT DISTINCT FROM, IS NULL, IS NOT NULL, BETWEEN, NOT BETWEEN, LIKE, ILIKE, NOT LIKE, NOT ILIKE, REGEXP, IN, NOT IN, GLOBAL IN, GLOBAL NOT IN, MOD, DIV, alias, AS


ch> with data as (select 'a' as name) SELECT name FROM data WHERE NOT ((SELECT DISTINCT name FROM data) IS NULL);


   ┌─name─┐
1. │ a    │
   └──────┘

```

Docs
---------
[Clickhouse IS operator](https://clickhouse.com/docs/en/sql-reference/operators#is_null)

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -1212,3 +1212,12 @@ class ClickHouse(Dialect):
 
         def projectiondef_sql(self, expression: exp.ProjectionDef) -> str:
             return f"PROJECTION {self.sql(expression.this)} {self.wrap(expression.expression)}"
+
+        def is_sql(self, expression: exp.Is) -> str:
+            is_sql = super().is_sql(expression)
+
+            if isinstance(expression.parent, exp.Not) and isinstance(expression.this, exp.Subquery):
+                # WHERE (SELECT ...) IS NOT NULL -> NOT ((SELECT ...) IS NULL)
+                is_sql = self.wrap(is_sql)
+
+            return is_sql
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
You checked and other dialects are ok with this?

VaggelisD:
> You checked and other dialects are ok with this?
Yep, no other (main) dialect had an issue parsing the commonly roundtripped version.

georgesittas:
Sounds good, thanks 👍

PR #4288, opened 2024-10-24
Fix: improve support for identifier delimiter escaping

georgesittas:
Our current assumptions around identifier delimiter escaping were wrong -- most importantly, we used `"` as an escape for all dialects, which isn't correct. I investigated what different dialects do and this is what I came up with:

```
ClickHouse, Databricks, Hive, MySQL, Spark, SQLite: `c```
DuckDB, Postgres, Presto, Trino, Redshift, Snowflake, SQLite, T-SQL: "x"""
ClickHouse also allows:  "c\", `c\``
T-SQL also allows: [[x]]]
BigQuery, Oracle don't support escaping identifier delimiters
```

As you can see, the most common rule is that the escape character is identical to the closing identifier delimiter. ClickHouse is an exception, because it also allows `\` to be used as an escape regardless of the surrounding context.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -608,10 +608,14 @@ class Tokenizer(metaclass=_Tokenizer):
     HEREDOC_STRINGS: t.List[str | t.Tuple[str, str]] = []
     UNICODE_STRINGS: t.List[str | t.Tuple[str, str]] = []
     IDENTIFIERS: t.List[str | t.Tuple[str, str]] = ['"']
-    IDENTIFIER_ESCAPES = ['"']
     QUOTES: t.List[t.Tuple[str, str] | str] = ["'"]
     STRING_ESCAPES = ["'"]
     VAR_SINGLE_TOKENS: t.Set[str] = set()
+
+    # The strings in this list can always be used as escapes, regardless of the surrounding
+    # identifier delimiters. By default, the closing delimiter is assumed to also act as an
+    # identifier escape, e.g. if we use double-quotes, then they also act as escapes: "x"""
+    IDENTIFIER_ESCAPES: t.List[str] = []
 
     # Whether the heredoc tags follow the same lexical rules as unquoted identifiers
     HEREDOC_TAG_IS_IDENTIFIER = False
@@ -1361,7 +1365,9 @@ class Tokenizer(metaclass=_Tokenizer):
 
     def _scan_identifier(self, identifier_end: str) -> None:
         self._advance()
-        text = self._extract_string(identifier_end, escapes=self._IDENTIFIER_ESCAPES)
+        text = self._extract_string(
+            identifier_end, escapes=self._IDENTIFIER_ESCAPES | {identifier_end}
+        )
         self._add(TokenType.IDENTIFIER, text)
 
     def _scan_var(self) -> None:
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -183,6 +183,7 @@ class ClickHouse(Dialect):
     class Tokenizer(tokens.Tokenizer):
         COMMENTS = ["--", "#", "#!", ("/*", "*/")]
         IDENTIFIERS = ['"', "`"]
+        IDENTIFIER_ESCAPES = ["\\"]
         STRING_ESCAPES = ["'", "\\"]
         BIT_STRINGS = [("0b", "")]
         HEX_STRINGS = [("0x", ""), ("0X", "")]
--- a/sqlglotrs/src/tokenizer.rs
+++ b/sqlglotrs/src/tokenizer.rs
@@ -609,8 +609,11 @@ impl<'a> TokenizerState<'a> {
         let mut text = String::from("");
 
         loop {
+            let mut new_identifier_escapes;
             let escapes = if use_identifier_escapes {
-                &self.settings.identifier_escapes
+                new_identifier_escapes = self.settings.identifier_escapes.clone();
+                new_identifier_escapes.extend(delimiter.chars());
+                &new_identifier_escapes
             } else {
                 &self.settings.string_escapes
             };
--- a/tests/dialects/test_dialect.py
+++ b/tests/dialects/test_dialect.py
@@ -2893,3 +2893,94 @@ FROM subquery2""",
                 "snowflake": "UUID_STRING()",
             },
         )
+
+    def test_escaped_identifier_delimiter(self):
+        for dialect in ("databricks", "hive", "mysql", "spark2", "spark"):
+            with self.subTest(f"Testing escaped backtick in identifier name for {dialect}"):
+                self.validate_all(
+                    'SELECT 1 AS "x`"',
+                    read={
+                        dialect: "SELECT 1 AS `x```",
+                    },
+                    write={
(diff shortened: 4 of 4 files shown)


sqlglot/tokens.py
```diff
@@ -1361,7 +1365,9 @@ def _scan_string(self, start: str) -> bool:
 
     def _scan_identifier(self, identifier_end: str) -> None:
         self._advance()
-        text = self._extract_string(identifier_end, escapes=self._IDENTIFIER_ESCAPES)
+        text = self._extract_string(
+            identifier_end, escapes=self._IDENTIFIER_ESCAPES | {identifier_end}
```
tobymao, line 1369: this seems expensive, we shouldn't be doing a set comparison inside this hot loop

georgesittas: One alternative is to push the complexity down to `_extract_string`, which would mean:

- Adding a new flag so that we know when to treat `identifier_end` (i.e. its `delimiter` argument) as an escape
- Updating the [two places](https://github.com/tobymao/sqlglot/blob/main/sqlglot/tokens.py#L1406-L1407) where we check `escapes` to include an `or` operation to check if the `_char` / `_peek` matches `delimiter` as a special case.

The performance tradeoff is not obvious to me in this case. Should I go ahead and benchmark the two implementations? Something like tokenizing a string with many identifiers in it.

tobymao: ya, can you bench mark it

georgesittas: Didn't notice any substantial differences between the two implementations. This is the benchmark script:

```
(.venv) ➜  sqlglot git:(jo/fix_identifier_delimiter_escaping) ✗ cat tokenize_identifiers.py
from timeit import timeit
from sqlglot import tokenize

sql = "SELECT " + ", ".join(["x"] * 1000)

print(timeit(lambda: tokenize(sql), number=1000))
```

Using the code in this PR:

```
(.venv) ➜  sqlglot git:(jo/fix_identifier_delimiter_escaping) ✗ SQLGLOTRS_TOKENIZER=0 python tokenize_identifiers.py
3.1507482500746846
(.venv) ➜  sqlglot git:(jo/fix_identifier_delimiter_escaping) ✗ SQLGLOTRS_TOKENIZER=1 python tokenize_identifiers.py
8.340694999787956
```

... and using the alternative implementation:

```
(.venv) ➜  sqlglot git:(jo/fix_identifier_delimiter_escaping) ✗ SQLGLOTRS_TOKENIZER=0 python tokenize_identifiers.py
3.189114083070308
(.venv) ➜  sqlglot git:(jo/fix_identifier_delimiter_escaping) ✗ SQLGLOTRS_TOKENIZER=1 python tokenize_identifiers.py
8.401180792134255
(.venv) ➜  sqlglot git:(jo/fix_identifier_delimiter_escaping) ✗
```

<details>
<summary>Click this to see the code diff</summary>
```diff
ns.py
+++ b/sqlglot/tokens.py
@@ -1366,7 +1366,7 @@ class Tokenizer(metaclass=_Tokenizer):
     def _scan_identifier(self, identifier_end: str) -> None:
         self._advance()
         text = self._extract_string(
-            identifier_end, escapes=self._IDENTIFIER_ESCAPES | {identifier_end}
+            identifier_end, escapes=self._IDENTIFIER_ESCAPES, delimiter_is_escape=True
         )
         self._add(TokenType.IDENTIFIER, text)

@@ -1390,6 +1390,7 @@ class Tokenizer(metaclass=_Tokenizer):
         escapes: t.Optional[t.Set[str]] = None,
         raw_string: bool = False,
         raise_unmatched: bool = True,
+        delimiter_is_escape: bool = False,
     ) -> str:
         text = ""
         delim_size = len(delimiter)
@@ -1409,7 +1410,7 @@ class Tokenizer(metaclass=_Tokenizer):
                     continue
             if (
                 (self.STRING_ESCAPES_ALLOWED_IN_RAW_STRINGS or not raw_string)
-                and self._char in escapes
+                and ((delimiter_is_escape and self._char == delimiter) or self._char in escapes)
                 and (self._peek == delimiter or self._peek in escapes)
                 and (self._char not in self._QUOTES or self._char == self._peek)
             ):
diff --git a/sqlglotrs/src/tokenizer.rs b/sqlglotrs/src/tokenizer.rs
index beff96e2..14388026 100644
--- a/sqlglotrs/src/tokenizer.rs
+++ b/sqlglotrs/src/tokenizer.rs
@@ -424,7 +424,7 @@ impl<'a> TokenizerState<'a> {
                 let tag = if self.current_char.to_string() == *end {
                     String::from("")
                 } else {
-                    self.extract_string(end, false, true, !self.settings.heredoc_tag_is_identifier)?
+                    self.extract_string(end, false, true, !self.settings.heredoc_tag_is_identifier, false)?
                 };

                 if !tag.is_empty()
@@ -449,7 +449,7 @@ impl<'a> TokenizerState<'a> {
         };

         self.advance(start.len() as isize)?;
-        let text = self.extract_string(&end, false, token_type == self.token_types.raw_string, true)?;
+        let text = self.extract_string(&end, false, token_type == self.token_types.raw_string, true, false)?;

         if let Some(b) = base {
             if u64::from_str_radix(&text, b).is_err() {
@@ -595,7 +595,7 @@ impl<'a> TokenizerState<'a> {

     fn scan_identifier(&mut self, identifier_end: &str) -> Result<(), TokenizerError> {
         self.advance(1)?;
-        let text = self.extract_string(identifier_end, true, false, true)?;
+        let text = self.extract_string(identifier_end, true, false, true, true)?;
         self.add(self.token_types.identifier, Some(text))
     }

@@ -605,19 +605,19 @@ impl<'a> TokenizerState<'a> {
         use_identifier_escapes: bool,
         raw_string: bool,
         raise_unmatched: bool,
+        delimiter_is_escape: bool,
     ) -> Result<String, TokenizerError> {
         let mut text = String::from("");

         loop {
-            let mut new_identifier_escapes;
+            // let mut new_identifier_escapes;
             let escapes = if use_identifier_escapes {
-                new_identifier_escapes = self.settings.identifier_escapes.clone();
-                new_identifier_escapes.extend(delimiter.chars());
-                &new_identifier_escapes
+                &self.settings.identifier_escapes
             } else {
                 &self.settings.string_escapes
             };
             let peek_char_str = self.peek_char.to_string();
+            let current_char_str = self.current_char.to_string();

             if !raw_string
                 && !self.dialect_settings.unescaped_sequences.is_empty()
@@ -635,13 +635,13 @@ impl<'a> TokenizerState<'a> {
             }

             if (self.settings.string_escapes_allowed_in_raw_strings || !raw_string)
-                && escapes.contains(&self.current_char)
+                && ((delimiter_is_escape && current_char_str == delimiter) || escapes.contains(&self.current_char))
                 && (peek_char_str == delimiter || escapes.contains(&self.peek_char))
                 && (self.current_char == self.peek_char
                     || !self
                         .settings
                         .quotes
-                        .contains_key(&self.current_char.to_string()))
+                        .contains_key(&current_char_str))
             {
                 if peek_char_str == delimiter {
                     text.push(self.peek_char);
```
</details>

As you can see, doing it the other way is even slightly slower in this particular example.

tobymao (approved):

PR #4290, opened 2024-10-25
clickhouse: improve Decimal rendering/parsing

pkit:
- add Decimal256 support too


georgesittas (approved):
Let's test the `Decimal` variants, we don't seem to do it today..

georgesittas:
Actually, I'll get this in and do it later. Thank you for the PR Constantine 👍

PR #4295, opened 2024-10-26
Fix: handle a Move edge case in the semantic differ

georgesittas:
This PR was motivated by the following case:

```python
>>> from sqlglot import diff, parse_one
>>>
>>> e1 = parse_one("select a as a, b as b from t where concat('a', 'b') = 'ab'")
>>> e2 = parse_one("select a as a from t where concat('a', 'b', b) = 'ab'")
>>>
>>> for edit in diff(e1, e2, matchings=[(e1, e2)], delta_only=True):
...     print(edit)
...
Remove(expression=Alias(
  this=Column(
    this=Identifier(this=b, quoted=False)),
  alias=Identifier(this=b, quoted=False)))
```

Notice how there's no information in the edit script about `b` being moved within the `concat` call. It looks like there were a few tests in `test_diff.py` that missed similar `Move` edits, so I updated them accordingly.

### DIFF
--- a/sqlglot/diff.py
+++ b/sqlglot/diff.py
@@ -10,6 +10,7 @@ import typing as t
 from collections import defaultdict
 from dataclasses import dataclass
 from heapq import heappop, heappush
+from itertools import chain
 
 from sqlglot import Dialect, expressions as exp
 from sqlglot.helper import seq_get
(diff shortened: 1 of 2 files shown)


sqlglot/diff.py
```diff
@@ -199,11 +214,22 @@ def _generate_edit_script(self, matchings: t.Dict[int, int], delta_only: bool) -
             source_node = self._source_index[kept_source_node_id]
             target_node = self._target_index[kept_target_node_id]
 
-            if (
-                not isinstance(source_node, UPDATABLE_EXPRESSION_TYPES)
-                or source_node == target_node
-            ):
-                edit_script.extend(self._generate_move_edits(source_node, target_node, matchings))
+            identical_nodes = source_node == target_node
```
georgesittas, line 217: I decided to cache the hashes because this equality check is gonna be hit a lot, since `matchings` must be generally quite larger than the unmatched source/target sets.

PR #4305, opened 2024-10-29
feat(spark)!: Transpile ANY to EXISTS

### LINKED ISSUE #4298: Bug In Parsing ANY function from postgres to spark
input:
```python
from sqlglot import parse_one
query = """SELECT col1 from table where ANY(col2) = 'value'"""
parse_one(query, dialect='postgres').sql(dialect='spark')
``` 
output is:
SELECT col1 FROM table WHERE ANY(col2) = 'value'

while the expected output is:
SELECT col1 FROM table WHERE ARRAY_CONTAINS(col2,'value') = true

VaggelisD:
Fixes #4298

In Hive hierarchy `ANY` is an aggregate function for `BOOLEAN` expressions and not an array/subquery operator as is in other dialects, meaning that the following transpiled queries do not work:

```Python
>>> sqlglot.parse_one("WITH t AS (SELECT ARRAY[1, 2, 3] AS col) SELECT * FROM t WHERE 1 <= ANY(col)", dialect="postgres").sql("spark")
'WITH t AS (SELECT ARRAY(1, 2, 3) AS col) SELECT * FROM t WHERE 1 <= ANY(col)'

spark-sql (default)> WITH t AS (SELECT ARRAY(1, 2, 3) AS col) SELECT * FROM t WHERE 1 <= ANY(col);
[DATATYPE_MISMATCH.UNEXPECTED_INPUT_TYPE] Cannot resolve "any(col)" due to data type mismatch: Parameter 1 requires the "BOOLEAN" type, however "col" has the type "ARRAY<INT>".; line 1 pos 68;
...
```

However, Spark2+ supports the `EXISTS()` function which can process both ARRAY expressions and subqueries in a similar fashion. This PR:

1. Adds a transformation for the former (ARRAY) case, enabling the following path:

```Python
>>> sqlglot.parse_one("WITH t AS (SELECT ARRAY[1, 2, 3] AS col) SELECT * FROM t WHERE 1 <= ANY(col)", dialect="postgres").sql("spark")
'WITH t AS (SELECT ARRAY(1, 2, 3) AS col) SELECT * FROM t WHERE EXISTS(col, x -> 1 <= x)'

spark-sql (default)> WITH t AS (SELECT ARRAY(1, 2, 3) AS col) SELECT * FROM t WHERE EXISTS(col, x -> 1 <= x);
[1,2,3]
```

2. Extends `EXISTS` class to also inherit from `Func`, making it possible to parse & construct it as a proper function. 

As step (2) is not required per se (could also build an anonymous function, check first commit), it's added as a standalone commit on top of (1); If we don't want to keep it it's trivial to drop it at once.

Docs
----------
[Postgres ANY for ARRAY](https://www.postgresql.org/docs/current/functions-comparisons.html#FUNCTIONS-COMPARISONS-ANY-SOME) | [Postgres ANY for subqueries](https://www.w3schools.com/postgresql/postgresql_any.php) | [Spark/Databricks ANY](https://docs.databricks.com/en/sql/language-manual/functions/any.html) | [Spark/Databricks EXISTS](https://docs.databricks.com/en/sql/language-manual/functions/exists.html)

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -938,8 +938,6 @@ def any_to_exists(expression: exp.Expression) -> exp.Expression:
                 lambda_arg = exp.to_identifier("x")
                 any.replace(lambda_arg)
                 lambda_expr = exp.Lambda(this=binop.copy(), expressions=[lambda_arg])
-                binop.replace(
-                    exp.Anonymous(this="EXISTS", expressions=[this.unnest(), lambda_expr])
-                )
+                binop.replace(exp.Exists(this=this.unnest(), expression=lambda_expr))
 
     return expression
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -914,3 +914,32 @@ def eliminate_join_marks(expression: exp.Expression) -> exp.Expression:
             where.pop()
 
     return expression
+
+
+def any_to_exists(expression: exp.Expression) -> exp.Expression:
+    """
+    Transform ANY operator to Spark's EXISTS
+
+    For example,
+        - Postgres: SELECT * FROM tbl WHERE 5 > ANY(tbl.col)
+        - Spark: SELECT * FROM tbl WHERE EXISTS(tbl.col, x -> x < 5)
+
+    Both ANY and EXISTS accept queries but currently only array expressions are supported for this
+    transformation
+    """
+    if isinstance(expression, exp.Select):
+        for any in expression.find_all(exp.Any):
+            this = any.this
+            if isinstance(this, exp.Query):
+                continue
+
+            binop = any.parent
+            if isinstance(binop, exp.Binary):
+                lambda_arg = exp.to_identifier("x")
+                any.replace(lambda_arg)
+                lambda_expr = exp.Lambda(this=binop.copy(), expressions=[lambda_arg])
+                binop.replace(
+                    exp.Anonymous(this="EXISTS", expressions=[this.unnest(), lambda_expr])
+                )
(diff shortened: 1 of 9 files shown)


sqlglot/transforms.py
```diff
@@ -914,3 +914,30 @@ def eliminate_join_marks(expression: exp.Expression) -> exp.Expression:
...
+    Transform ANY operator to Spark's EXISTS
+
+    For example,
+        - Postgres: SELECT * FROM tbl WHERE 5 > ANY(tbl.col)
+        - Spark: SELECT * FROM tbl WHERE EXISTS(tbl.col, x -> x < 5)
+
+    Both ANY and EXISTS accept queries but currently only array expressions are supported for this
+    transformation
+    """
+    if isinstance(expression, exp.Select):
```
georgesittas, line 930: Why don't we just hook into `exp.Any`'s generator for the dialects of interest?

VaggelisD: That was my first attempt, but I don't think it's very practical because `ANY` is a child of an `exp.Binary` node; This means that for a filter such as `WHERE <expr> <op> <any_expr>`, we will have already processed the LHS before we get to `any_sql` for the RHS. To mitigate that, we'd have to hook into `generator.py::binary()` and do the transformation there in advance.

For this reason, I thought it'd look cleaner to add a `transform` and in the future it can be extended to cover subqueries as well.

georgesittas (approved):

tobymao (approved):

PR #4328, opened 2024-10-31
fix(oracle): Do not normalize time units for exp.DateTrunc

### LINKED ISSUE #4321: unit generated in oracle sql is incorrect when input is "Q"
```
In [6]: import sqlglot as sg, sqlglot.expressions as sge

In [7]: sg.func("trunc", sg.to_identifier("x"), sge.convert("Q")).sql("oracle")
Out[7]: "TRUNC(x, 'Q')"

In [8]: sg.func("trunc", sg.to_identifier("x"), sge.convert("Q"), dialect="oracle").sql("oracle")
Out[8]: "TRUNC(x, 'QUARTER')"
```

`ROUND`/`TRUNC` docs showing units: https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/ROUND-and-TRUNC-Date-Functions.html#GUID-8E10AB76-21DA-490F-A389-023B648DDEF8

VaggelisD:
Fixes #4321

Time unit normalizations such as unabbreviation (e.g. `Q` -> `QUARTER`) take place in expression constructors so decoupling them from specific nodes & dialects is not trivial. 

To minimize damage I went with the following arg-based approach, but if similar bugs appear for other expressions and/or dialects we probably look into establishing finer granularity (per expr and per dialect).

Docs
-------
[Oracle TRUNC units](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/ROUND-and-TRUNC-Date-Functions.html#GUID-8E10AB76-21DA-490F-A389-023B648DDEF8)

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -121,7 +121,9 @@ class Oracle(Dialect):
             "TO_TIMESTAMP": build_formatted_time(exp.StrToTime, "oracle"),
             "TO_DATE": build_formatted_time(exp.StrToDate, "oracle"),
             "TRUNC": lambda args: exp.DateTrunc(
-                unit=seq_get(args, 1) or exp.Literal.string("DD"), this=seq_get(args, 0)
+                unit=seq_get(args, 1) or exp.Literal.string("DD"),
+                this=seq_get(args, 0),
+                unabbreviate=False,
             ),
         }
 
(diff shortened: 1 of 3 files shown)


VaggelisD:
I agree, as discussed in Slack the time units are not handled very well currently (e.g code duplication between `exp.TimeUnit` and `exp.DateTrunc`, unable to override per dialect etc).

However, I entertained the idea of refactoring and it seems like a sizable task; We'd have to move part of that code to `Dialect` and rework `exp.TimeUnit` to be normalized during construction per dialect, which I'm assuming would have to go through builders. This will disrupt most of the unit-based Expression parsing today afaict, right?

It might be worth exploring asap if more edge cases like these creep up.

georgesittas (approved):

PR #4329, opened 2024-10-31
feat(snowflake): Transpile exp.TimestampSub

VaggelisD:
Snowflake does not have a `TIMESTAMP_SUB` function, but it's `TIMESTAMP_ADD` can be reused by negating the value. After this PR, the transpilation is the following:

- BigQuery:
```SQL
WITH t AS (SELECT 20 AS col)
SELECT
    TIMESTAMP_SUB(TIMESTAMP "2008-12-25 15:30:00+00", INTERVAL 10 MINUTE),
    TIMESTAMP_SUB(TIMESTAMP "2008-12-25 15:30:00+00", INTERVAL col MINUTE)
FROM t;

2008-12-25 15:20:00 UTC | 2008-12-25 15:10:00 UTC
```
- Snowflake:
```SQL
WITH t AS (SELECT 20 AS col) 
SELECT 
    TIMESTAMPADD(MINUTE, -10, CAST('2008-12-25 15:30:00+00' AS TIMESTAMPTZ)), 
    TIMESTAMPADD(MINUTE, -col, CAST('2008-12-25 15:30:00+00' AS TIMESTAMPTZ)) 
from t;

2008-12-25 15:20:00.000 +0000 | 2008-12-25 15:10:00.000 +0000
```

Docs
---------
[BQ TIMESTAMP_SUB](https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#timestamp_sub) | [SF TIMESTAMP_ADD](https://docs.snowflake.com/en/sql-reference/functions/timestampadd)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1089,12 +1089,10 @@ class Snowflake(Dialect):
             )
 
         def timestampsub_sql(self, expression: exp.TimestampSub):
-            expr = expression.expression
-            if isinstance(expr, exp.Literal):
-                expr = expr.to_py()
-
             return self.sql(
                 exp.TimestampAdd(
-                    this=expression.this, expression=exp.Neg(this=expr), unit=expression.unit
+                    this=expression.this,
+                    expression=expression.expression * -1,
+                    unit=expression.unit,
                 )
             )
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1087,3 +1087,14 @@ class Snowflake(Dialect):
             return self.func(
                 f"{safe_prefix}TO_TIMESTAMP", expression.this, self.format_time(expression)
             )
+
+        def timestampsub_sql(self, expression: exp.TimestampSub):
+            expr = expression.expression
+            if isinstance(expr, exp.Literal):
+                expr = expr.to_py()
+
+            return self.sql(
+                exp.TimestampAdd(
+                    this=expression.this, expression=exp.Neg(this=expr), unit=expression.unit
+                )
+            )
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -1087,3 +1087,12 @@ def strtotime_sql(self, expression: exp.StrToTime):
             return self.func(
                 f"{safe_prefix}TO_TIMESTAMP", expression.this, self.format_time(expression)
             )
+
+        def timestampsub_sql(self, expression: exp.TimestampSub):
+            return self.sql(
+                exp.TimestampAdd(
+                    this=expression.this,
+                    expression=expression.expression * -1,
```
georgesittas, line 1095: Do we need to wrap `expression.expression` in a `Paren` here?

VaggelisD:
Actually this can be further simplified as Snowflake will coerce `<string_expr> * <integer_expr> -> INTEGER`:

```
snowflake: select typeof('10' * -1);
INTEGER
```

So it's not required to transform it through `to_py()` first.

sqlglot/dialects/snowflake.py
VaggelisD: This is already handled by `__mul__` afaict:


```python
>>> sql = "SELECT TIMESTAMP_SUB(TIMESTAMP "2008-12-25 15:30:00+00", INTERVAL 2 + 1 MINUTE)"
>>> sqlglot.parse_one(sql, dialect="bigquery").sql("snowflake")
SELECT TIMESTAMPADD(MINUTE, (2 + 1) * -1, CAST('2008-12-25 15:30:00+00' AS TIMESTAMPTZ))
```

georgesittas (approved):

PR #4336, opened 2024-11-02
Fix(bigquery)!: parse information schema views into a single identifier

georgesittas:
Behavior in main today:

```python
>>> parse_one("select * from db.information_schema.tables", "bigquery")
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=tables, quoted=False),
      db=Identifier(this=information_schema, quoted=False),
      catalog=Identifier(this=db, quoted=False))))
>>> parse_one("select * from c.db.information_schema.tables", "bigquery")
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Dot(
        this=Identifier(this=information_schema, quoted=False),
        expression=Identifier(this=tables, quoted=False)),
      db=Identifier(this=db, quoted=False),
      catalog=Identifier(this=c, quoted=False))))
```

This PR makes it so that we will always represent `INFORMATION_SCHEMA.X` views using a single `Identifier` expression for BigQuery. The motivation here is that, previously, we wouldn't properly qualify information schema references, because we'd think there's a db/catalog present.

Choosing `Identifier` over a `Dot` is motivated by the fact that we the schema module can't handle multiple mapping depths, so this should keep the nesting consistent even if there are information schema view references.

Xref: https://github.com/TobikoData/sqlmesh/issues/3317

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -315,7 +315,7 @@ class TestOptimizer(unittest.TestCase):
                 ),
                 dialect="bigquery",
             ).sql(),
-            'WITH "x" AS (SELECT "y"."a" AS "a" FROM "DB"."y" AS "y" CROSS JOIN "a"."b"."INFORMATION_SCHEMA"."COLUMNS" AS "COLUMNS") SELECT "x"."a" AS "a" FROM "x" AS "x"',
+            'WITH "x" AS (SELECT "y"."a" AS "a" FROM "DB"."y" AS "y" CROSS JOIN "a"."b"."INFORMATION_SCHEMA.COLUMNS" AS "columns") SELECT "x"."a" AS "a" FROM "x" AS "x"',
         )
 
         self.assertEqual(
--- a/benchmarks/optimize.py
+++ b/benchmarks/optimize.py
@@ -18,12 +18,12 @@ def gen_condition(n):
 
 BENCHMARKS = {
     "tpch": lambda: (
-        [parse_one(sql) for _, sql, _ in load_sql_fixture_pairs(f"optimizer/tpc-h/tpc-h.sql")],
+        [parse_one(sql) for _, sql, _ in load_sql_fixture_pairs("optimizer/tpc-h/tpc-h.sql")],
         TPCH_SCHEMA,
         3,
     ),
     "tpcds": lambda: (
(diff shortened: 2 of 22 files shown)


barakalon (commented):
I’m mostly familiar with MySQL’s information_schema. In MySQL, information_schema _is_ a database, so I think this would be wrong?

Does this PR imply that Bigquery creates default information_schema tables in each database? And if so, I wonder how common that is.

georgesittas:
> I’m mostly familiar with MySQL’s information_schema. In MySQL, information_schema _is_ a database, so I think this would be wrong?
> 
> Does this PR imply that Bigquery creates default information_schema tables in each database? And if so, I wonder how common that is.

Ah, got it - thanks for catching this. I did assume there would be an information schema under each db, but I'll double-check this assumption.

Regarding BigQuery, yes, they [need](https://cloud.google.com/bigquery/docs/information-schema-intro#syntax) the information schema to be qualified with a region or dataset:

> An INFORMATION_SCHEMA view needs to be qualified with a dataset or region.

georgesittas:
@barakalon it turns out your hunch was right, none of the other engines treat the information schema views like BigQuery does, so this fix really needs to be specific to it. Thanks for catching this, should've investigated more thoroughly.

georgesittas:
There's some work needing to be done here around 4-part table references in the schema module, because right now we're expecting the references to have the same depth but information schema views break this invariant.

barakalon (approved):

tests/test_optimizer.py
```diff
@@ -315,7 +315,7 @@ def test_qualify_columns(self, logger):
                 ),
                 dialect="bigquery",
             ).sql(),
-            'WITH "x" AS (SELECT "y"."a" AS "a" FROM "DB"."y" AS "y" CROSS JOIN "a"."b"."INFORMATION_SCHEMA"."COLUMNS" AS "COLUMNS") SELECT "x"."a" AS "a" FROM "x" AS "x"',
+            'WITH "x" AS (SELECT "y"."a" AS "a" FROM "DB"."y" AS "y" CROSS JOIN "a"."b"."INFORMATION_SCHEMA.COLUMNS" AS "columns") SELECT "x"."a" AS "a" FROM "x" AS "x"',
```
georgesittas, line 318: This change should be safe because BigQuery aliases are case-insensitive.

barakalon (approved):

PR #4337, opened 2024-11-02
feat(trino): Support ListAgg function with overflow clause

Kize54:
Added support for the `ON OVERFLOW` clause in the `ListAgg` function in Trino. According to the [Trino documentation](https://trino.io/docs/current/functions/aggregate.html#listagg), this functionality allows handling overflow situations during aggregation.

During testing, calling 
```
sqlglot.parse_one("SELECT LISTAGG(DISTINCT col ON OVERFLOW ERROR) WITHIN GROUP (ORDER BY col asc) FROM tbl", dialect="trino")
```
caused the error `sqlglot.errors.ParseError: Expecting )`. 

This error has been resolved by adding the appropriate parsing for the ListAgg function.

### DIFF
--- a/sqlglot/dialects/trino.py
+++ b/sqlglot/dialects/trino.py
@@ -18,7 +18,7 @@ class Trino(Presto):
             **Presto.Parser.FUNCTION_PARSERS,
             "TRIM": lambda self: self._parse_trim(),
             "JSON_QUERY": lambda self: self._parse_json_query(),
-            "LISTAGG": lambda self: self._parse_string_agg(),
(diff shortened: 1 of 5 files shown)


VaggelisD (commented):
Hey @Kize54, thank you for the PR!

My main reservation has to do with the addition of the `ListAgg` node, it looks like it's used in other places as well for `LISTAGG` such as Snowflake. Was it required to truly separate concerns?

georgesittas:
Agreed with @VaggelisD, there's quite a bit more complexity than necessary here I think. Closing this in favor of https://github.com/tobymao/sqlglot/pull/4340.

PR #4313, opened 2024-10-30
use static alias for unnesting a struct

gauravsagar483:
### Targetting these two threads on the sqlglot channel.
1. https://tobiko-data.slack.com/archives/C0448SFS3PF/p1728537731625359
2. https://tobiko-data.slack.com/archives/C0448SFS3PF/p1724147809318389

Spark/Hive LATERAL VIEW  EXPLODE requires only single alias for the respective exploded column to be given unlike UNNEST in trino/presto, which can take multiple aliases for the single exploded column.

### Docs
--- 
[Hive LateralView](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+LateralView) | [Spark LateralView](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html) | [Presto UNNEST](https://prestodb.io/docs/current/sql/select.html#unnest)

PR #4339, opened 2024-11-04
Chore: improve error handling for unnest_to_explode

gauravsagar483:
## ReOpen for PR: [ #4313 ](https://github.com/tobymao/sqlglot/pull/4313)
Highlights: I reverted the hive changes after validation, as `ARRAY_ZIP` is not supported. Thanks @georgesittas 

### Targeting these two threads on the sqlglot channel.
1. https://tobiko-data.slack.com/archives/C0448SFS3PF/p1728537731625359
2. https://tobiko-data.slack.com/archives/C0448SFS3PF/p1724147809318389

Spark/Hive LATERAL VIEW  EXPLODE requires only single alias for the respective exploded column to be given unlike UNNEST in trino/presto, which can take multiple aliases for the single exploded column.

### Docs
--- 
[Hive LateralView](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+LateralView) | [Spark LateralView](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html) | [Presto UNNEST](https://prestodb.io/docs/current/sql/select.html#unnest)

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -302,16 +302,13 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def unnest_to_explode(
-    expression: exp.Expression
-) -> exp.Expression:
+def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
     """Convert cross join unnest into lateral view explode."""
 
     def _unnest_zip_exprs(
         u: exp.Unnest, unnest_exprs: t.List[exp.Expression], has_multi_expr: bool
     ) -> t.List[exp.Expression]:
         if has_multi_expr:
-
             # Use INLINE(ARRAYS_ZIP(...)) for multiple expressions
             zip_exprs: t.List[exp.Expression] = [
                 exp.Anonymous(this="ARRAYS_ZIP", expressions=unnest_exprs)
@@ -365,8 +362,7 @@ def unnest_to_explode(
                 expression.args["joins"].remove(join)
 
                 alias_cols = alias.columns if alias else []
-                
-                
+
                 """
                 Handle Presto CROSS JOIN UNNEST to LATERAL VIEW EXPLODE for Multiple or No Exploded table column alias.
                 
@@ -383,7 +379,7 @@ def unnest_to_explode(
                     for column in expression.find_all(exp.Column):
                         if alias and column.table == alias.name:
                             column.set("table", "t_struct")
-                
+
                 for e, column in zip(exprs, alias_cols):
                     expression.append(
                         "laterals",
@@ -392,7 +388,7 @@ def unnest_to_explode(
                             view=True,
                             alias=exp.TableAlias(
                                 this=alias.this,  # type: ignore
-                                columns=alias_cols
+                                columns=alias_cols,
                             ),
                         ),
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -369,11 +369,11 @@ def unnest_to_explode(
 
                 alias_cols = alias.columns if alias else []
 
-                # # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or >= 2 aliases 
+                # # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or >= 2 aliases
                 # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take an arbitrary amount.
                 # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
 
-                if not has_multi_expr and len(alias_cols) !=1:
+                if not has_multi_expr and len(alias_cols) != 1:
                     raise UnsupportedError(
                         "CROSS JOIN UNNEST to LATERAL VIEW EXPLODE transformation requires explicit column aliases"
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -302,13 +302,19 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
-def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
+def unnest_to_explode(
+    expression: exp.Expression,
+    unnest_using_arrays_zip: bool = True,
+) -> exp.Expression:
     """Convert cross join unnest into lateral view explode."""
 
     def _unnest_zip_exprs(
         u: exp.Unnest, unnest_exprs: t.List[exp.Expression], has_multi_expr: bool
     ) -> t.List[exp.Expression]:
         if has_multi_expr:
+            if not unnest_using_arrays_zip:
+                raise UnsupportedError("Cannot transpile UNNEST with multiple input arrays")
+
             # Use INLINE(ARRAYS_ZIP(...)) for multiple expressions
             zip_exprs: t.List[exp.Expression] = [
                 exp.Anonymous(this="ARRAYS_ZIP", expressions=unnest_exprs)
@@ -389,7 +395,7 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
                             view=True,
                             alias=exp.TableAlias(
                                 this=alias.this,  # type: ignore
-                                columns=alias_cols,
+                                columns=alias_cols if unnest_using_arrays_zip else [column],  # type: ignore
                             ),
                         ),
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -386,7 +386,7 @@ def unnest_to_explode(
                             view=True,
                             alias=exp.TableAlias(
                                 this=alias.this,  # type: ignore
-                                columns=alias_cols if unnest_using_arrays_zip else [column],  # type: ignore
+                                columns=alias_cols,
                             ),
                         ),
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -376,9 +376,10 @@ def unnest_to_explode(expression: exp.Expression) -> exp.Expression:
                     alias_cols = ["t_struct"]
 
                     # [Optional] Do Update the Column reference in AST for table with current alias
-                    for column in expression.find_all(exp.Column):
-                        if alias and column.table == alias.name:
-                            column.set("table", "t_struct")
+                    if alias:
+                        for column in expression.find_all(exp.Column):
+                            if column.table == alias.name:
+                                column.set("table", "t_struct")
 
                 for e, column in zip(exprs, alias_cols):
                     expression.append(
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -369,13 +369,13 @@ def unnest_to_explode(
 
                 alias_cols = alias.columns if alias else []
 
-                # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there is 0 or more then 2 aliases for single unnest column.
-                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take zero or more.
+                # # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or >= 2 aliases 
+                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take an arbitrary amount.
                 # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
 
-                if not has_multi_expr and len(alias_cols) not in [1, 2]:
+                if not has_multi_expr and len(alias_cols) !=1:
                     raise UnsupportedError(
-                        "Please provide alias to unnest column as per column type to support the transformation."
+                        "CROSS JOIN UNNEST to LATERAL VIEW EXPLODE transformation requires explicit column aliases"
                     )
 
                 for e, column in zip(exprs, alias_cols):
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -373,7 +373,7 @@ def unnest_to_explode(
                 # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take an arbitrary amount.
                 # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
 
-                if not has_multi_expr and len(alias_cols) not in [1, 2]:
+                if not has_multi_expr and len(alias_cols) not in (1, 2):
                     raise UnsupportedError(
                         "CROSS JOIN UNNEST to LATERAL VIEW EXPLODE transformation requires explicit column aliases"
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -369,23 +369,14 @@ def unnest_to_explode(
 
                 alias_cols = alias.columns if alias else []
 
-                """
-                Handle Presto CROSS JOIN UNNEST to LATERAL VIEW EXPLODE for Multiple or No Exploded table column alias.
-                
-                Spark/Hive LATERAL VIEW EXPLODE requires only single alias for respective exploded array/struct to be given for unnest ulike trino/presto which can take multiple.
-                https://cwiki.apache.org/confluence/display/Hive/LanguageManual+LateralView
-                https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
-                
-                """
-                # Replace multiple alias for single EXPLODE column with single static alias name: `t_struct`
-                if not has_multi_expr and (len(alias_cols) != 1):
-                    alias_cols = ["t_struct"]
+                # Handle UNNEST to LATERAL VIEW EXPLODE in case of 'No alias'
+                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column ulike unnest in trino/presto which can take zero or more.
+                # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
 
-                    # [Optional] Do Update the Column reference in AST for table with current alias
-                    if alias:
-                        for column in expression.find_all(exp.Column):
-                            if column.table == alias.name:
-                                column.set("table", "t_struct")
+                if not has_multi_expr and len(alias_cols) not in [1, 2]:
+                    raise UnsupportedError(
+                        "Please provide alias to unnest column as per column type to support the transformation."
+                    )
 
                 for e, column in zip(exprs, alias_cols):
                     expression.append(
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -369,11 +369,11 @@ def unnest_to_explode(
 
                 alias_cols = alias.columns if alias else []
 
-                # # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or >= 2 aliases
+                # # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or > 2 aliases
                 # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take an arbitrary amount.
                 # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
 
-                if not has_multi_expr and len(alias_cols) != 1:
+                if not has_multi_expr and len(alias_cols) not in [1, 2]:
                     raise UnsupportedError(
                         "CROSS JOIN UNNEST to LATERAL VIEW EXPLODE transformation requires explicit column aliases"
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -303,8 +303,7 @@ def unqualify_unnest(expression: exp.Expression) -> exp.Expression:
 
 
 def unnest_to_explode(
-    expression: exp.Expression,
-    unnest_using_arrays_zip: bool = True,
+    expression: exp.Expression
 ) -> exp.Expression:
     """Convert cross join unnest into lateral view explode."""
 
@@ -312,8 +311,6 @@ def unnest_to_explode(
         u: exp.Unnest, unnest_exprs: t.List[exp.Expression], has_multi_expr: bool
     ) -> t.List[exp.Expression]:
         if has_multi_expr:
-            if not unnest_using_arrays_zip:
-                raise UnsupportedError("Cannot transpile UNNEST with multiple input arrays")
 
             # Use INLINE(ARRAYS_ZIP(...)) for multiple expressions
             zip_exprs: t.List[exp.Expression] = [
@@ -368,6 +365,25 @@ def unnest_to_explode(
                 expression.args["joins"].remove(join)
 
                 alias_cols = alias.columns if alias else []
+                
+                
+                """
+                Handle Presto CROSS JOIN UNNEST to LATERAL VIEW EXPLODE for Multiple or No Exploded table column alias.
+                
+                Spark/Hive LATERAL VIEW EXPLODE requires only single alias for respective exploded array/struct to be given for unnest ulike trino/presto which can take multiple.
+                https://cwiki.apache.org/confluence/display/Hive/LanguageManual+LateralView
+                https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
+                
+                """
+                # Replace multiple alias for single EXPLODE column with single static alias name: `t_struct`
+                if not has_multi_expr and (len(alias_cols) != 1):
+                    alias_cols = ["t_struct"]
+
+                    # [Optional] Do Update the Column reference in AST for table with current alias
+                    for column in expression.find_all(exp.Column):
+                        if alias and column.table == alias.name:
+                            column.set("table", "t_struct")
+                
                 for e, column in zip(exprs, alias_cols):
                     expression.append(
                         "laterals",
@@ -376,7 +392,7 @@ def unnest_to_explode(
                             view=True,
                             alias=exp.TableAlias(
                                 this=alias.this,  # type: ignore
-                                columns=alias_cols if unnest_using_arrays_zip else [column],  # type: ignore
+                                columns=alias_cols
                             ),
                         ),
                     )
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -369,8 +369,8 @@ def unnest_to_explode(
 
                 alias_cols = alias.columns if alias else []
 
-                # Handle UNNEST to LATERAL VIEW EXPLODE in case of 'No alias'
-                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column ulike unnest in trino/presto which can take zero or more.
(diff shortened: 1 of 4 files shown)


sqlglot/transforms.py
```diff
@@ -368,6 +368,25 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
...
+                """
+                # Replace multiple alias for single EXPLODE column with single static alias name: `t_struct`
+                if not has_multi_expr and (len(alias_cols) != 1):
+                    alias_cols = ["t_struct"]
+
+                    # [Optional] Do Update the Column reference in AST for table with current alias
+                    if alias:
+                        for column in expression.find_all(exp.Column):
+                            if column.table == alias.name:
+                                column.set("table", "t_struct")
```
VaggelisD, line 388: Is this tested for multiple/chained CROSS JOINs? Replacing every alias instance with the same hardcoded column name looks off to me, it could lead to ambiguous column errors.

On `sqlglot.helpers` we have `name_sequence` which can generate a sequence of names if that turns out to be a problem.

tests/dialects/test_dialect.py
```diff
@@ -1444,6 +1444,36 @@ def test_cross_join(self):
...
+            },
+        )
+
+        # UNNEST STRUCT Object into multiple columns, using single alias
+        self.validate_all(
+            "SELECT a, b FROM x CROSS JOIN UNNEST(y) AS t (a, b)",
+            write={
+                "presto": "SELECT a, b FROM x CROSS JOIN UNNEST(y) AS t(a, b)",
+                "spark": "SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS t_struct",
+                "hive": "SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS t_struct",
```
VaggelisD, line 1463: This doesn't look correct, was it tested on Spark? We've changed the column aliases from `(a, b)` to `t_struct`, we shouldn't be able to select them right? What's the behavior of exploding multiple columns and having only 1 alias?

VaggelisD (commented):
Hey @gauravsagar483, thank you for letting us know and reopening the PR, I did a new pass and had a few questions:

tests/dialects/test_dialect.py
gauravsagar483: Okay, this is strange, and I just noticed!!
I tested the optimized query which works fine on Spark, in which it was able to reference the columns correctly from `t_struct`
please check ref: I'll cross check and see if I can add the reference back above in tranpiled itself.
Screenshot 2024-11-04 at 6 05 15 PM

sqlglot/transforms.py
```diff
@@ -368,6 +368,25 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
...
                 alias_cols = alias.columns if alias else []
+
+                """
+                Handle Presto CROSS JOIN UNNEST to LATERAL VIEW EXPLODE for Multiple or No Exploded table column alias.
+                
+                Spark/Hive LATERAL VIEW EXPLODE requires only single alias for respective exploded array/struct to be given for unnest ulike trino/presto which can take multiple.
+                https://cwiki.apache.org/confluence/display/Hive/LanguageManual+LateralView
+                https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
+                
+                """
```
georgesittas, line 379: [Nit] let's not use a multi-line string here, it's inconsistent with the existing style. Instead, use normal `#` comments.

sqlglot/transforms.py
VaggelisD: If you had multiple CROSS JOINS (e.g. `SELECT t_struct FROM tbl CROSS JOIN UNNEST(y) CROSS JOIN UNNEST(z)`) the projection `SELECT t_struct` would lead into ambiguous reference error afaict.

Let's verify if that's a problem and if yes, we can mitigate it through `name_sequence`.

VaggelisD: Also, the premise that Spark can't alias multiple columns does not always look right. From the docs:
- [LATERAL VIEW's](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html#parameters) `column_alias`: Lists the column aliases of generator_function, which may be used in output rows. We may have multiple aliases if generator_function have multiple output columns.
- [EXPLODE](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.explode.html): Uses the default column name col for elements in the array and key and value for elements in the map unless specified otherwise.

So consider this `MAP` example from Presto -> Trino:

```SQL
presto> SELECT
     ->   animals,
     ->   a,
     ->   n
     -> FROM (VALUES
     ->   (MAP(ARRAY['dog', 'cat', 'bird'], ARRAY[1, 2, 0])),
     ->   (MAP(ARRAY['dog', 'cat'], ARRAY[4, 5]))) AS x(animals)
     -> CROSS JOIN UNNEST(animals) AS t(a, n);
        animals         |  a   | n
------------------------+------+---
 {cat=2, bird=0, dog=1} | dog  | 1
 {cat=2, bird=0, dog=1} | cat  | 2
 {cat=2, bird=0, dog=1} | bird | 0
 {cat=5, dog=4}         | dog  | 4
 {cat=5, dog=4}         | cat  | 5
(5 rows)


spark-sql (default)> SELECT
                   >   animals,
                   >   a,
                   >   n
                   > FROM VALUES
                   >   (MAP_FROM_ARRAYS(ARRAY('dog', 'cat', 'bird'), ARRAY(1, 2, 0))),
                   >   (MAP_FROM_ARRAYS(ARRAY('dog', 'cat'), ARRAY(4, 5))) AS x(animals)
                   > LATERAL VIEW
                   > EXPLODE(animals) t AS a, n;
animals a       n
{"bird":0,"cat":2,"dog":1}      dog     1
{"bird":0,"cat":2,"dog":1}      cat     2
{"bird":0,"cat":2,"dog":1}      bird    0
{"cat":5,"dog":4}       dog     4
{"cat":5,"dog":4}       cat     5
Time taken: 0.148 seconds, Fetched 5 row(s)
```

If we change the aliases from `AS t (a, n)` to just `t AS t_struct`, we now get the opposite error:

```SQL
spark-sql (default)> SELECT animals, a, n FROM VALUES (MAP_FROM_ARRAYS(ARRAY('dog', 'cat', 'bird'), ARRAY(1, 2, 0))), (MAP_FROM_ARRAYS(ARRAY('dog', 'cat'), ARRAY(4, 5))) AS x(animals) LATERAL VIEW EXPLODE(animals) t AS t_struct;
The number of aliases supplied in the AS clause does not match the number of columns output by the UDTF expected 2 aliases but got t_struct.; line 1 pos 163
```

gauravsagar483: i'll check and update today.

gauravsagar483: Hi @VaggelisD,

I iterated through multiple similar UNNESTusecases (single and multiple) and I feel that without giving the schema to the SQL it is impossible to figure out how many aliases should be placed in query without knowing the column type.

Please refer to the different sample scenarios below:

**Single CROSS JOIN UNNEST scenarios:**

No Alias given
```
Presto: SELECT t.a, t.b, x.c, x.d FROM x CROSS JOIN UNNEST(y) as t
Spark:  SELECT t_struct.a, t_struct.b, x.c, x.d FROM x LATERAL VIEW EXPLODE(y) t AS t_struct
```
_It only works after t_struct change, otherwise generates wrong query by removing the unnest part like so SELECT t.a, t.b, x.c, x.d FROM x_

2. Two alias available, column y type is uncertain.
```
Presto: SELECT a, b FROM x CROSS JOIN UNNEST(y) AS t (a, b) 
Spark:  Only works if y is of MAP type
```
 _If columns y is of any other type spark expects only single alias col._

3.
```
Presto: SELECT a, b, c, d FROM x CROSS JOIN UNNEST(y) as t(a, b, c, d)
Spark:  Currently without t_struct (single alias) implementation this will fail for both array and struct type as it expects single alias col to be present.
```

**Multiple CROSS JOIN UNNEST scenarios:**

For multiple/chained unnest sqlglot do return the incorrect query with or without these changes.

Example SQL:
```
Presto: SELECT a,b,c, col1, col2 FROM tbl CROSS JOIN UNNEST(y) as t(col1) CROSS JOIN UNNEST(z) as t(col2)
Spark:  SELECT a, b, c, col1, col2 FROM tbl CROSS JOIN EXPLODE(z, t(col2)) LATERAL VIEW EXPLODE(y) t AS col1
```
_Please note the spark syntax is incorrect, having `CROSS JOIN EXPLODE`_

General workaround to fix is to modify the chained cross join unnest having multiple expression to single UNNEST expressionuser has to rewrite the query as:

```
Presto: SELECT a, b, c, col1, col2 FROM tbl CROSS JOIN UNNEST(y,z) as u(col1, col2)
Spark:  SELECT a, b, c, col1, col2 FROM tbl LATERAL VIEW INLINE(ARRAYS_ZIP(y, z)) u AS col1, col2
```
_This works fine if the column y & z are ARRAY or STRUCT type, if it is MAP then the complete dictionary will be returned in the final column as {Key: value}_

Current t_struct change does not do anything for above as the above scenario comes under has_multi_expr


**Would you suggest any other approach for handling** these without Schema? Without knowing the schema, it is almost uncertain and can have many other edge cases.
Do you suggest I should update my PR only to handle the 1st case (No alias), and will consider everything later in other PR keeping an Issue open to track?

sqlglot/transforms.py
gauravsagar483: this is doable, i'll update.

sqlglot/transforms.py
georgesittas: We should only implement transformations / changes that are always known to be correct. Type-sensitive changes are tricky to get correctly and since you don't always have the schema, you need to do a best-effort approach.

I haven't fully followed the thread / nuances here, but it looks like we should only look into the 1st case (no alias). There's no need to open an issue to track the other cases if there's not an obvious solution, I'd rather we use Slack for this conversation because we don't plan to work on it.

VaggelisD: Right, so it comes down to the fact that without knowing `y` in  `LATERAL VIEW EXPLODE(y)` we can't deduce if we should keep/add 2 column aliases (`MAP` type) or 1 column alias (`STRUCT` type).

I think the first problem (no alias) suffers from that too, right? If we are unnesting a `MAP`, adding just `t AS t_struct` will lead to an error in Spark.

gauravsagar483: Yes, I agree with both of you above, At the moment, I do not see an obvious solution for other cases without inferring the schema in SQL.

Yes @VaggelisD , For the first case(No alias), The problem will still be there for **_MAP type_** as spark internally tracks the map type and expects 2 alias during explode. But the other 2 cases (Array & Struct) should work fine with `t_struct`.

What do you guys suggest next step here?

VaggelisD: Can you expand more on the first case? I tried some queries again and can't see why it errors:


```SQL
# Exploding array without column alias
spark-sql (default)> WITH t AS (SELECT ARRAY(1) AS col) SELECT * FROM t LATERAL VIEW EXPLODE(col) t;
col     col
[1]     1

# Exploding array of structs without column alias
spark-sql (default)> WITH t AS (SELECT ARRAY(STRUCT(1)) AS col) SELECT * FROM t LATERAL VIEW EXPLODE(col) t;
col     col
[{"col1":1}]    {"col1":1}

# Selecting from an exploded struct
spark-sql (default)> WITH t AS (SELECT ARRAY(STRUCT(1)) AS arr) SELECT t.col FROM t LATERAL VIEW EXPLODE(arr) t;
col
{"col1":1}
```

If possible, to explain the error try to include full queries (such as the one above) and/or the full repro for incorrect SQLGlot transpilations so we can test them ourselves.

Generally, though, we should try to solve a case provided that we tick all these boxes:
- Our solution will be backwards-compatible with SQLGlot (will not break/alter correct queries)
- It will only improve our handling. For example, if _all_ queries with no alias would lead to incorrect Spark, by unblocking _some_ of them our solution will have only positive benefits
- If `y` is not an array/struct, the transformation will lead to an error exactly as if the transformation was never applied; Otherwise, if the transformation introduces side effects, we might introduce bugs, silent corruptions etc
- The solution works properly for the targeted types e.g arrays and structs

gauravsagar483: Hi @VaggelisD , Please find attached repro queries in each scenario: (presto & spark)

## Scenario 1:  ARRAY

**Presto:**
`WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY['value1', 'value2'] AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY['value3', 'value4'] AS y) SELECT * FROM data CROSS JOIN UNNEST(y) AS t`

**Spark: Without `t_struct` change (main):**
_Please note it removed the lateral view explode condition if we do not pass the alias._

`WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY('value1', 'value2') AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY('value3', 'value4') AS y) SELECT * FROM data`

**Spark: With `t_struct` change:**
`WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY('value1', 'value2') AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY('value3', 'value4') AS y) SELECT * FROM data LATERAL VIEW EXPLODE(y) t AS t_struct`


## Scenario 2:  MAP

**Presto:** 
`WITH data AS (SELECT 1 AS a, 'example1' AS b, MAP(ARRAY['key1', 'key2'], ARRAY['value1', 'value2']) AS y UNION ALL SELECT 2 AS a, 'example2' AS b, MAP(ARRAY['key3', 'key4'], ARRAY['value3', 'value4']) AS y) SELECT * FROM data AS d CROSS JOIN UNNEST(y) AS t`

**Spark: Without `t_struct` change (main):**
`WITH data AS (SELECT 1 AS a, 'example1' AS b, MAP_FROM_ARRAYS(ARRAY('key1', 'key2'), ARRAY('value1', 'value2')) AS y UNION ALL SELECT 2 AS a, 'example2' AS b, MAP_FROM_ARRAYS(ARRAY('key3', 'key4'), ARRAY('value3', 'value4')) AS y) SELECT * FROM data AS d`

**Spark: With `t_struct` change:**
`WITH data AS (SELECT 1 AS a, 'example1' AS b, MAP_FROM_ARRAYS(ARRAY('key1', 'key2'), ARRAY('value1', 'value2')) AS y UNION ALL SELECT 2 AS a, 'example2' AS b, MAP_FROM_ARRAYS(ARRAY('key3', 'key4'), ARRAY('value3', 'value4')) AS y) SELECT * FROM data AS d LATERAL VIEW EXPLODE(y) t AS t_struct`

## Scenario 2:  STRUCT

**Presto:** 
`WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY[ROW('key1', 'value1'), ROW('key2', 'value2')] AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY[ROW('key3', 'value3'), ROW('key4', 'value4')] AS y) SELECT * FROM data CROSS JOIN UNNEST(y) AS t`

**Spark: Without `t_struct` change (main):**
`WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY(STRUCT('key1', 'value1'), STRUCT('key2', 'value2')) AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY(STRUCT('key3', 'value3'), STRUCT('key4', 'value4')) AS y) SELECT * FROM data`

**Spark: With `t_struct` change:**
`WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY(STRUCT('key1', 'value1'), STRUCT('key2', 'value2')) AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY(STRUCT('key3', 'value3'), STRUCT('key4', 'value4')) AS y) SELECT * FROM data LATERAL VIEW EXPLODE(y) t AS t_struct`


Conclusion: Based on the above & similar examples, the `No Alias Case`  currently has the transpile issue in all three cases, but with the `t_struct` change,  It will  still be there for **_MAP_** type but will work fine for other cases of array & struct.

VaggelisD: Thank you for the iterations & the research thus far @gauravsagar483, this is a nuanced topic.

From what I can see, it looks like the `ARRAY` is the only case which we'd properly solve; The `MAP` will lead to an error and the `STRUCT` produces different results:

- Presto/Trino will unnest both the `ARRAY` and the `STRUCT` fields:
```SQL
presto> WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY[ROW('key1', 'value1'), ROW('key2', 'value2')] AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY[ROW('key3', 'value3'), ROW('key4', 'value4')] AS y) SELECT * FROM data CROSS JOIN UNNEST(y) AS t;

 a |    b     |                              y                               | _col3 | _col4
---+----------+--------------------------------------------------------------+-------+--------
 2 | example2 | [{field0=key3, field1=value3}, {field0=key4, field1=value4}] | key3  | value3
 2 | example2 | [{field0=key3, field1=value3}, {field0=key4, field1=value4}] | key4  | value4
 1 | example1 | [{field0=key1, field1=value1}, {field0=key2, field1=value2}] | key1  | value1
 1 | example1 | [{field0=key1, field1=value1}, {field0=key2, field1=value2}] | key2  | value2
```

- Spark will unnest only the `ARRAY`, leaving the `STRUCT` as a column:
```SQL
spark-sql (default)> WITH data AS (SELECT 1 AS a, 'example1' AS b, ARRAY(STRUCT('key1', 'value1'), STRUCT('key2', 'value2')) AS y UNION ALL SELECT 2 AS a, 'example2' AS b, ARRAY(STRUCT('key3', 'value3'), STRUCT('key4', 'value4')) AS y) SELECT * FROM data LATERAL VIEW EXPLODE(y) t AS t_struct;

a       b       y       t_struct
1       example1        [{"col1":"key1","col2":"value1"},{"col1":"key2","col2":"value2"}]       {"col1":"key1","col2":"value1"}
1       example1        [{"col1":"key1","col2":"value1"},{"col1":"key2","col2":"value2"}]       {"col1":"key2","col2":"value2"}
2       example2        [{"col1":"key3","col2":"value3"},{"col1":"key4","col2":"value4"}]       {"col1":"key3","col2":"value3"}
2       example2        [{"col1":"key3","col2":"value3"},{"col1":"key4","col2":"value4"}]       {"col1":"key4","col2":"value4"}
```

So to iterate:
1. If `y` is a pure `ARRAY`, Spark should have 1 column alias
2. If `y` is a `MAP`, Spark should have 2 column aliases
3. If `y` is an `ARRAY` of `STRUCT`s, Spark should explode both of them (would require even more logic)

Since we don't know what `y` is, I think the solution at the end is to raise an `UnsupportedError` inside `unnest_to_explode` if there are no aliases; The user should add them to enable the transformation, and if Spark's semantics don't match, that will lead to a clear error.

gauravsagar483: Sure let me update the PR for the above. Thank you @VaggelisD

sqlglot/transforms.py
gauravsagar483: Updated with [b789e00](https://github.com/tobymao/sqlglot/pull/4339/commits/b789e00803367cf347246b71737104fd0ac56767)

tests/dialects/test_dialect.py
gauravsagar483: Reverted `t_struct` change here,

Transpiled and optimized spark will look like this(No Change): Presto to Spark

- Presto Query 
```
SELECT a, b,c,d FROM x CROSS JOIN UNNEST(y) AS t(a, b)
```
- Spark Transpile
```
SELECT a, b, c, d FROM x LATERAL VIEW EXPLODE(y) t AS a, b
```
- Spark Optimize
```
SELECT
  `t`.`a` AS `a`,
  `t`.`b` AS `b`,
  `x`.`c` AS `c`,
  `x`.`d` AS `d`
FROM `x` AS `x`
LATERAL VIEW
EXPLODE(`x`.`y`) t AS `a`, `b`
```

sqlglot/transforms.py
gauravsagar483: Updated with [b789e00](https://github.com/tobymao/sqlglot/pull/4339/commits/b789e00803367cf347246b71737104fd0ac56767)

sqlglot/transforms.py
```diff
@@ -368,6 +368,16 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
...
 
                 alias_cols = alias.columns if alias else []
+
+                # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there is 0 or more then 2 aliases for single unnest column.
+                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take zero or more.
+                # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
+
+                if not has_multi_expr and len(alias_cols) not in [1, 2]:
+                    raise UnsupportedError(
+                        "Please provide alias to unnest column as per column type to support the transformation."
```
VaggelisD, line 378: Let's change that to 3rd person, something along the lines of:

`CROSS JOIN UNNEST to LATERAL VIEW EXPLODE transformation requires explicit column aliases`

sqlglot/transforms.py
```diff
@@ -368,6 +368,16 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
                 expression.args["joins"].remove(join)
 
                 alias_cols = alias.columns if alias else []
+
+                # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there is 0 or more then 2 aliases for single unnest column.
```
VaggelisD, line 372:
```suggestion
                # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or >= 2 aliases 
```

sqlglot/transforms.py
```diff
@@ -368,6 +368,16 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
...
+                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take zero or more.
```
VaggelisD, line 373:
```suggestion
                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take an arbitrary amount.
```

tests/dialects/test_dialect.py
```diff
@@ -1444,6 +1444,46 @@ def test_cross_join(self):
...
+            },
+        )
+
+        # UNNEST STRUCT Object into multiple columns, using single alias
+        self.validate_all(
+            "SELECT a, b FROM x CROSS JOIN UNNEST(y) AS t (a, b)",
+            write={
+                "presto": "SELECT a, b FROM x CROSS JOIN UNNEST(y) AS t(a, b)",
+                "spark": "SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS a, b",
+                "hive": "SELECT a, b FROM x LATERAL VIEW EXPLODE(y) t AS a",
```
VaggelisD, line 1463: Why is Hive not following Spark here with the amount of column aliases?

VaggelisD (commented):
The conclusion we reached is that this transformation is hard to get right; In Presto/Trino, we could be unnesting a plethora of types such as arrays, maps, arrays of structs etc which show different behaviors and require different amount of column aliases in Spark.

For that matter, this PR was narrowed down to adding `UnsupportedError` if the user does not supply column alias _or_ supplies more than 2; Spark is unable to process both of these scenarios. For better context, this [discussion] (https://github.com/tobymao/sqlglot/pull/4339#discussion_r1829848407) downwards showcases our final observations of that.

@gauravsagar483 Thank you for the iterations once more, I think I only have a few nits and 1 question thus far so it should be close to merging. I'll do another pass soon to gather what we've found thus far and will approve if everything checks out.

tests/dialects/test_dialect.py
gauravsagar483: It's happening due to this previous [change](https://github.com/tobymao/sqlglot/blob/main/sqlglot/transforms.py#L379) for `hive`, where 
`unnest_using_arrays_zip` is being sent as false by default to prevent multiple alias.

`columns=alias_cols if unnest_using_arrays_zip else [column],  # type: ignore`

georgesittas:
@gauravsagar483 let us know when this is ready for another review.

tests/dialects/test_dialect.py
gauravsagar483: Okay, I tested few more scenario on `hive` for each type:  array, map and struct, 
and hive does not support exploding struct.
`Error while compiling statement: FAILED: UDFArgumentException explode() takes an array or a map as a parameter`

Query:
```
WITH cte_x AS ( SELECT 1 AS id, STRUCT('value1', 123, TRUE) AS y UNION ALL SELECT 2 AS id, STRUCT('value2', 456, FALSE) AS y ) SELECT * FROM cte_x LATERAL VIEW EXPLODE(y) t AS a,b,c
```
Array requires single alias
```
WITH cte_x AS ( SELECT 1 AS id, ARRAY("value1", "value2", "value3") AS y UNION ALL SELECT 2 AS id, ARRAY("value4", "value5", "value6") AS y ) SELECT * FROM cte_x LATERAL VIEW EXPLODE(y) t AS a
```

Map requires 2

```
WITH cte_x AS ( SELECT 1 AS id, MAP("key1", "value1", "key2", "value2") AS y UNION ALL SELECT 2 AS id, MAP("key3", "value3", "key4", "value4") AS y ) SELECT * FROM cte_x LATERAL VIEW EXPLODE(y) t AS a,b
```

which is essentially the same behaviour as spark except it does not support the struct. I've fixed the code and testcases to support correct exception handling.

gauravsagar483:
Hi @georgesittas & @VaggelisD, It is not ready for review. Please check. Thank you.

georgesittas:
Thanks -- no worries about the commits. I'll review again soon, thanks for iterating on this 👍

tests/dialects/test_dialect.py
```diff
@@ -1444,6 +1444,46 @@ def test_cross_join(self):
...
+        self.validate_all(
+            "SELECT * FROM x CROSS JOIN UNNEST(y) AS t",
+            write={
+                "presto": "SELECT * FROM x CROSS JOIN UNNEST(y) AS t",
+                "spark": UnsupportedError,
+                "databricks": UnsupportedError,
+            },
+        )
+
+        # UNNEST STRUCT Object into multiple columns, using single alias
```
georgesittas, line 1457:
```suggestion
        # UNNEST MAP Object into multiple columns, using single alias
```

sqlglot/transforms.py
```diff
@@ -368,6 +368,16 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
                 expression.args["joins"].remove(join)
 
                 alias_cols = alias.columns if alias else []
+
+                # # Handle UNNEST to LATERAL VIEW EXPLODE: Exception is raised when there are 0 or > 2 aliases
+                # Spark LATERAL VIEW EXPLODE requires single alias for array/struct and two for Map type column unlike unnest in trino/presto which can take an arbitrary amount.
+                # Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html
+
+                if not has_multi_expr and len(alias_cols) not in [1, 2]:
```
georgesittas, line 376:
```suggestion
                if not has_multi_expr and len(alias_cols) not in (1, 2):
```

georgesittas (approved):
Thanks for the PR @gauravsagar483!

PR #4340, opened 2024-11-04
Feat(trino): add support for the ON OVERFLOW clause in LISTAGG

georgesittas:
See #4337

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -1049,3 +1049,13 @@ class BigQuery(Dialect):
             if expression.name == "TIMESTAMP":
                 expression.set("this", "SYSTEM_TIME")
             return super().version_sql(expression)
+
+        @generator.unsupported_args("group", "parameters")
+        def regexpextract_sql(self, e: exp.RegexpExtract) -> str:
+            return self.func(
+                "REGEXP_EXTRACT",
+                e.this,
+                e.expression,
+                e.args.get("position"),
+                e.args.get("occurrence"),
+            )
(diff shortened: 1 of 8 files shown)


VaggelisD:
Seems like `ruff` was bumped, is the investigation around it finished? 

PS: I thought it was pinned by us, is it not?

georgesittas:
> Seems like `ruff` was bumped, is the investigation around it finished?
> 
> PS: I thought it was pinned by us, is it not?

Yeah, there's no problem bumping it. The reason it was pinned in the first place was to avoid random CI breakages.

sqlglot/dialects/bigquery.py
```diff
@@ -1049,3 +1049,13 @@ def version_sql(self, expression: exp.Version) -> str:
...
+
+        @generator.unsupported_args("group", "parameters")
+        def regexpextract_sql(self, e: exp.RegexpExtract) -> str:
+            return self.func(
+                "REGEXP_EXTRACT",
+                e.this,
+                e.expression,
+                e.args.get("position"),
+                e.args.get("occurrence"),
+            )
```
VaggelisD, line 1061: This should have been removed after the latest rebase

VaggelisD (approved):

sqlglot/expressions.py
```diff
@@ -5765,8 +5765,14 @@ class Greatest(Func):
     is_var_len_args = True
 
 
+# Trino's `ON OVERFLOW TRUNCATE [filler_string] {WITH | WITHOUT} COUNT`
+# https://trino.io/docs/current/functions/aggregate.html#listagg
+class OverflowTruncateBehavior(Expression):
+    arg_types = {"this": False, "with_count": True}
```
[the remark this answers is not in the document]
georgesittas: The latter, I didn't feel like adding more args to `GroupConcat` in order to represent the overflow behavior spec.

sqlglot/dialects/bigquery.py
georgesittas, line 1061:
```suggestion
```

PR #4341, opened 2024-11-04
fix!: Refactor exp.RegexpExtract (follow up 4326)

VaggelisD:
Follow up of https://github.com/tobymao/sqlglot/pull/4326

- Reverted BigQuery's generator and it's unsupported args; We add default values to `group` on other dialects so we'd need to silence these warnings on each test
- Simplified DuckDB's generator
- Removed duplicated tests
- Removed decorator prefix by importing it


NickCrews:
Thank you @VaggelisD for finishing this up! Your "empty inbox" of prs and issues is inspiring!

PR #4343, opened 2024-11-04
fix(duckdb): we ALWAYS need to render group if params is present for RegexpExtract

NickCrews:
Followup to https://github.com/tobymao/sqlglot/pull/4341

Before, roundtripping `SELECT REGEXP_EXTRACT(a, 'pattern', 0, 'i')` would result in `SELECT REGEXP_EXTRACT(a, 'pattern', 'i')`, which is incorrect (the group=0 arg is needed to fill space). So, we need to adjust our logic of when we can drop the group arg: it can only happen when there are no following args.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -957,15 +957,16 @@ class DuckDB(Dialect):
         @unsupported_args("position", "occurrence")
         def regexpextract_sql(self, expression: exp.RegexpExtract) -> str:
             group = expression.args.get("group")
+            params = expression.args.get("parameters")
 
-            # Do not render group if it's the default value for this dialect
-            if group and group.name == str(self.dialect.REGEXP_EXTRACT_DEFAULT_GROUP):
+            # Do not render group if there is no following argument,
+            # and it's the default value for this dialect
+            if (
+                not params
+                and group
+                and group.name == str(self.dialect.REGEXP_EXTRACT_DEFAULT_GROUP)
+            ):
                 group = None
-
             return self.func(
-                "REGEXP_EXTRACT",
-                expression.this,
-                expression.expression,
-                group,
-                expression.args.get("parameters"),
+                "REGEXP_EXTRACT", expression.this, expression.expression, group, params
             )
--- a/tests/dialects/test_duckdb.py
+++ b/tests/dialects/test_duckdb.py
@@ -785,6 +785,12 @@ class TestDuckDB(Validator):
                 "snowflake": "SELECT REGEXP_SUBSTR(a, 'pattern', 1, 1, 'i', 2) FROM t",
             },
         )
+        self.validate_identity(
+            "SELECT REGEXP_EXTRACT(a, 'pattern', 0)",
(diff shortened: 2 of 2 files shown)


tests/dialects/test_duckdb.py
```diff
@@ -785,6 +785,12 @@ def test_duckdb(self):
                 "snowflake": "SELECT REGEXP_SUBSTR(a, 'pattern', 1, 1, 'i', 2) FROM t",
             },
         )
+        self.validate_identity(
+            "SELECT REGEXP_EXTRACT(a, 'pattern', 0)",
+            write_sql="SELECT REGEXP_EXTRACT(a, 'pattern')",
```
VaggelisD, line 790:
```suggestion
        self.validate_identity(
            "SELECT REGEXP_EXTRACT(a, 'pattern', 0)",
            "SELECT REGEXP_EXTRACT(a, 'pattern')",
```

VaggelisD (approved):
Hey @NickCrews, thanks for catching this.

According to [DuckDB](https://duckdb.org/docs/sql/functions/regular_expressions.html#regexp_extractstring-pattern-group--0-options), this is the function signature:

``
regexp_extract(string, pattern[, group = 0][, options])
```

It would seem as if `group` and `options` are not dependent on each other, but one can't use the former without defining the latter (even for `= 0`); My previous follow up chopped that logic off.

**Comment by NickCrews** [comment]
That suggestion looks good.

At a border scale, this makes me wonder if we should change the API of self.func to error if we encounter self.func("func_name", ok_so_far, None, oops_another_arg_after_None)? With this particular case, duckdb would error if we tried to execute this SQL because the types are wrong, but I could see other functions where we would execute successfully but get the wrong result. So maybe we want to guard against that more?

File: sqlglot/dialects/duckdb.py [hunk 2]
```diff
@@ -957,15 +957,16 @@ def arraytostring_sql(self, expression: exp.ArrayToString) -> str:
         @unsupported_args("position", "occurrence")
         def regexpextract_sql(self, expression: exp.RegexpExtract) -> str:
             group = expression.args.get("group")
+            params = expression.args.get("parameters")
 
-            # Do not render group if it's the default value for this dialect
-            if group and group.name == str(self.dialect.REGEXP_EXTRACT_DEFAULT_GROUP):
+            # Do not render group if there is no following argument,
```
NickCrews, line 962: oh, if you're in there changing things, one tiny improvement on the wording here could be
"We can skip group if there is no following argument". But not blocking

VaggelisD:
Generally our APIs are fairly permissive and for this instance `self.func` ignoring `None` params has the benefit that we don't have to add branches for every optional argument. In contrast, having args being dependent on one another _and_ optimizing them away is not that frequent, so in this case it's worth to add the logic ourselves.

sqlglot/dialects/duckdb.py
VaggelisD: Didn't find it hard to read in the first place but feel free to make any final touches, otherwise I'll go ahead and merge it soon

PR #4355, opened 2024-11-06
fix(clickhouse): Generate exp.Median as lowercase

### LINKED ISSUE #4354: `median` constructed using `sg.func` is incorrectly capitalized when compiled to ClickHouse
```
In [2]: import sqlglot as sg

In [3]: sg.__version__
Out[3]: '25.29.0'

In [4]: sg.func("median", sg.to_identifier("x")).sql("clickhouse")
Out[4]: 'MEDIAN(x)'
```

VaggelisD:
Fixes #4354

```SQL
:) select MEDIAN(5);
Received exception:
Code: 63. DB::Exception: Unknown aggregate function MEDIAN. Maybe you meant: ['median','medianDD']. (UNKNOWN_AGGREGATE_FUNCTION)

:) select median(5);
   ┌─median(5)─┐
1. │         5 │
   └───────────┘


```

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -949,6 +949,7 @@ class ClickHouse(Dialect):
             exp.JSONPathKey: json_path_key_only_name,
             exp.JSONPathRoot: lambda *_: "",
             exp.Map: lambda self, e: _lower_func(var_map_sql(self, e)),
+            exp.Median: rename_func("median"),
             exp.Nullif: rename_func("nullIf"),
             exp.PartitionedByProperty: lambda self, e: f"PARTITION BY {self.sql(e, 'this')}",
             exp.Pivot: no_pivot_sql,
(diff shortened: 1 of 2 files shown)


georgesittas:
I'm not sure about this approach, shouldn't we instead have a separate path to use `FUNCTION_PARSERS` in `exp.func`?

VaggelisD:
Well, on the one hand that's right, `median(x)` is roundtripped as `quantile(0.5)(x)` for Clickhouse, and this issue exposes that the builder is not adequate by searching only the dialect's `FUNCTIONS`. For context, I think I had done that because CH can still parse `median(quantile)(expr)` as `quantile`:
```
:) with t as (Select 1 as col union all select 2 as col) select median(col), median(0.1)(col), quantile(0.1)(col) from t;


   ┌─median(col)─┬─median(0.1)(col)─┬─quantile(0.1)(col)─┐
1. │         1.5 │              1.1 │                1.1 │
   └─────────────┴──────────────────┴────────────────────┘
```

However, the generation part will still need to be fixed, right? If we're transpiling `exp.Median` from another read dialect we must still (1) generate it as lowercase `median(x)` or (2) change the generation to `quantile(0.5)(x)`.

georgesittas:
Yeah, good point. However, we should still address the AST discrepancy in a followup PR.

georgesittas (approved):

PR #4357, opened 2024-11-06
Fix(clickhouse): Remove RANGE data type

treysp:
Clickhouse supports numeric positional column references of the form `[object name].[column number]` (e.g., `table.1` refers to the first column in `table`).

If the object is named `range`, sqlglot parses the reference into the range data type:
```
>>> parse_one("a between range.1 and range.2", dialect="clickhouse").sql("clickhouse")
'a BETWEEN CAST(0.1 AS RANGE) AND CAST(0.2 AS RANGE)'
```

Clickhouse does not have a `RANGE` data type, so this PR removes it.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -419,6 +419,8 @@ class ClickHouse(Dialect):
             TokenType.LIKE,
         }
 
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
Wow, I'm curious if there are other engines that support this.

PR #4359, opened 2024-11-07
feat: Support REGEXP_EXTRACT_ALL

VaggelisD:
This PR adds support for `REGEXP_EXTRACT_ALL` which returns  an array of all the matching substrings.

The dialects that support `REGEXP_EXTRACT` also support `REGEXP_EXTRACT_ALL` with largely the same args & default values, so this PR largely reuses the existing builders / sql functions. 

We could also support it through an `all` arg, but I hesitated because:
- The 2 functions might differ for the same dialect e.g. in BigQuery the args do not align
- Their semantics & type annotation is different (scalar vs array)

Docs
---------
[BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_extract_all) | [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/regexp_substr_all#syntax) | [Databricks](https://docs.databricks.com/en/sql/language-manual/functions/regexp_extract_all.html) | [Spark3](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.regexp_extract_all.html) | [Presto](https://prestodb.io/docs/current/functions/regexp.html#regexp_extract_all) | [Trino](https://trino.io/docs/current/functions/regexp.html#regexp_extract_all) | [DuckDB](https://duckdb.org/docs/sql/functions/regular_expressions.html#regexp_extract_allstring-regex-group--0-options)


georgesittas (approved):
Nice work

PR #4364, opened 2024-11-09
fix: safe removal while iterating

gauravsagar483:
### Bugfix: sqlglot generates incorrect expression list for Multiple Chained UNNEST.
In the `unnest_to_explode` method,  While iterating for multiple Unnest expressions, sqlglot modifies(remove join expr) the list during iteration, leading to incorrect SQL generation.
`expression.args["joins"].remove(join)`

### Repro steps:
Presto SQL:
```
SELECT * FROM x CROSS JOIN UNNEST(a) AS j(lista) CROSS JOIN UNNEST(b) AS k(listb) CROSS JOIN UNNEST(c) AS l(listc)
```

Spark SQL: sqlglot incorrect transpiled query: (Please note `CROSS JOIN EXPLODE(b, k(listb))`)
```
SELECT * FROM x CROSS JOIN EXPLODE(b, k(listb)) LATERAL VIEW EXPLODE(a) j AS lista LATERAL VIEW EXPLODE(c) l AS listc
```

After modifying the loop to iteratete over the copy list, `expression.args["joins"].remove(join)` can safely remove modify the expression list.

Spark SQL after fix:

```
SELECT * FROM x LATERAL VIEW EXPLODE(a) j AS lista LATERAL VIEW EXPLODE(b) k AS listb LATERAL VIEW EXPLODE(c) l AS listc
```

### Fix:
Iterate over a copy list to make the removal from original list safe.

**Another possible fix** 
 instead of `list()`, We can `reversed()` but this is modifying the expression order and retuning the opposite order of explode expressions, Not sure if order matters, hence sticking to. l`ist()` one.
```
SELECT * FROM x LATERAL VIEW EXPLODE(c) l AS listc LATERAL VIEW EXPLODE(b) k AS listb LATERAL VIEW EXPLODE(a) j AS lista
```

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -348,7 +348,7 @@ def unnest_to_explode(
                 )
             )
 
-        for join in expression.args.get("joins") or []:
+        for join in list(expression.args.get("joins") or []): # support safe removal by iterating over copy list for multiple joins
             join_expr = join.this
 
             is_lateral = isinstance(join_expr, exp.Lateral)
@@ -390,6 +390,8 @@ def unnest_to_explode(
                             ),
                         ),
                     )
+                    
+        
 
     return expression
 
--- a/tests/dialects/test_dialect.py
+++ b/tests/dialects/test_dialect.py
@@ -1483,6 +1483,16 @@ class TestDialect(Validator):
                 "hive": UnsupportedError,
             },
         )
+        
+    def test_multiple_chained_unnest(self):
+        self.validate_all(
+            "SELECT * FROM x CROSS JOIN UNNEST(a) AS j(lista) CROSS JOIN UNNEST(b) AS k(listb) CROSS JOIN UNNEST(c) AS l(listc)",
+            write={
+                "presto": "SELECT * FROM x CROSS JOIN UNNEST(a) AS j(lista) CROSS JOIN UNNEST(b) AS k(listb) CROSS JOIN UNNEST(c) AS l(listc)",
+                "spark": "SELECT * FROM x LATERAL VIEW EXPLODE(a) j AS lista LATERAL VIEW EXPLODE(b) k AS listb LATERAL VIEW EXPLODE(c) l AS listc",
+                "hive": "SELECT * FROM x LATERAL VIEW EXPLODE(a) j AS lista LATERAL VIEW EXPLODE(b) k AS listb LATERAL VIEW EXPLODE(c) l AS listc",
+            },
+        )        
 
     def test_lateral_subquery(self):
         self.validate_identity(


sqlglot/transforms.py
```diff
@@ -348,7 +348,9 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
                 )
             )
 
-        for join in expression.args.get("joins") or []:
+        for join in list(
+            expression.args.get("joins") or []
+        ):  # support safe removal by iterating over copy list for multiple joins
```
VaggelisD, line 353: This can be further optimized, e.g the first solution that comes to mind is:

```
joins = expression.args.get("joins") or []

for join in list(joins):
   ...
   joins.remove(join)
```

This way we don't have to pay the dict lookup cost at every loop iteration at L370. 

I think the `list(joins)` shallow copy is not an issue here, I would keep the solution as is.

VaggelisD (commented):
Hey @gauravsagar483, good catch & thank you for the PR, a minor comment:

sqlglot/transforms.py
gauravsagar483: agree.

gauravsagar483: updated with [d1f4219](https://github.com/tobymao/sqlglot/pull/4364/commits/d1f4219a9cb60048e3551da7f9d6d57153ab89c3)

sqlglot/transforms.py
```diff
@@ -348,7 +348,10 @@ def _udtf_type(u: exp.Unnest, has_multi_expr: bool) -> t.Type[exp.Func]:
                 )
             )
 
-        for join in expression.args.get("joins") or []:
+        joins = expression.args.get("joins") or []
+        for join in list(
+            joins
+        ):  # support safe removal by iterating over copy list for multiple joins
```
VaggelisD, line 354:
```suggestion
        for join in list(joins):
```

VaggelisD (approved):

sqlglot/transforms.py
VaggelisD: We must also change the L370 to use `joins`

gauravsagar483: ohh i overlook, let me quickly fix that.

sqlglot/transforms.py
gauravsagar483: Commited: [27c7753](https://github.com/tobymao/sqlglot/pull/4364/commits/27c775376c60b20ca4679d9c1095a776a3ac1a2d)

VaggelisD: This is a different comment, the formatting here can be inlined. The explanation might be good to have as a reminder but imo it's a common Python pattern so we can remove it too

gauravsagar483: Suggestion accepted.

georgesittas (approved):
Thanks for the fix 👍

PR #4369, opened 2024-11-11
Feat: optionally wrap connectors when using builders

### LINKED ISSUE #4362: Excessive Recursion in Query Optimization with Multiple OR Clauses
## Context  
We are encountering an issue where a query with a high number of OR operators is causing excessive recursion during the optimization phase. The resulting recursion depth leads to stack overflow errors. As a temporary workaround, we increased the stack size limit.

Despite the number of entries not being particularly high, we suspect that something in the optimization process is causing the recursion depth to increase unexpectedly.
## Fully reproducible code snippet
```python
import sqlglot
import sqlglot.expressions as expressions
from sqlglot.expressions import column

is_equal_list = ['a'] * 500
is_equal = expressions.false()


for value in is_equal_list:
    is_equal = is_equal.or_(column("a_column").eq(value))
```

If you try to access `is_equal`, you'll receive an error:
```python
is_equal
# throws
#sqlglot/expressions.py", line 256, in is_leaf
#    return not any(isinstance(v, (Expression, list)) for v in self.args.values())
#                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#RecursionError: maximum recursion depth exceeded
```
The default recursion depth is 1000.

### LINKED ISSUE #4367: [Reopen issue 4362] Excessive Recursion in Query Optimization with Multiple OR Clauses
## Context  
**Sorry for reopening https://github.com/tobymao/sqlglot/issues/4362, but the proposed solution had no effect.**  

Even using `.sql()` instead of `.to_s`, we have hit the same stack limit issue. **It starts to reach the depth limit (1000) at 165 elements in the clause**:
```
is_equal.sql()
# throws
File "sqlglot/generator.py", line 3067, in or_sql
    return self.connector_sql(expression, "OR", stack)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "sqlglot/generator.py", line 3084, in connector_sql
    stack.append(expression.right)
                 ^^^^^^^^^^^^^^^^
  File "sqlglot/expressions.py", line 4675, in right
    return self.expression
           ^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded
```
I'm not sure if the project wants to fix this, but it seems reasonable to think that SQLGlot Programatically builds queries, and this doesn't seem like an uncommon use case.

georgesittas:
Closes #4367
Closes #4362

Now this works:

```python
>>> import sqlglot
>>> import sqlglot.expressions as expressions
>>> from sqlglot.expressions import column
>>>
>>> is_equal_list = ['a'] * 500
>>> is_equal = expressions.false()
>>>
>>>
>>> for value in is_equal_list:
...     is_equal = is_equal.or_(column("a_column").eq(value), wrap=False)
...
>>> is_equal.sql()[:100]
"FALSE OR a_column = 'a' OR a_column = 'a' OR a_column = 'a' OR a_column = 'a' OR a_column = 'a' OR a"
```

I didn't add a timing test because they're flaky. I think we should refactor the existing ones at some point too.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -767,6 +767,7 @@ class Expression(metaclass=_Expression):
         *expressions: t.Optional[ExpOrStr],
         dialect: DialectType = None,
         copy: bool = True,
+        wrap: bool = True,
         **opts,
     ) -> Condition:
         """
@@ -781,18 +782,22 @@ class Expression(metaclass=_Expression):
                 If an `Expression` instance is passed, it will be used as-is.
             dialect: the dialect used to parse the input expression.
             copy: whether to copy the involved expressions (only applies to Expressions).
+            wrap: whether to wrap the operands in `Paren`s. This is true by default to avoid
+                precedence issues, but can be turned off when the produced AST is too deep and
+                causes recursion-related issues.
             opts: other options to use to parse the input expressions.
 
         Returns:
             The new And condition.
         """
-        return and_(self, *expressions, dialect=dialect, copy=copy, **opts)
+        return and_(self, *expressions, dialect=dialect, copy=copy, wrap=wrap, **opts)
 
     def or_(
         self,
         *expressions: t.Optional[ExpOrStr],
         dialect: DialectType = None,
         copy: bool = True,
+        wrap: bool = True,
         **opts,
     ) -> Condition:
         """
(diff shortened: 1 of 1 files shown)


VaggelisD (approved):
RE: Iterative generation logic

Could we solve this issue by unwrapping `exp.Paren` during iteration instead of traversing downwards? This could solve both issues at once, unless the API addition here is nice to have anyways.

georgesittas:
> RE: Iterative generation logic
> 
> Could we solve this issue by unwrapping `exp.Paren` during iteration instead of traversing downwards? This could solve both issues at once, unless the API addition here is nice to have anyways.

Sure, wanna take a stab at it? I think conditionally wrapping in the API makes sense, but improving `Generator.binary` is also a good idea.

PR #4371, opened 2024-11-11
fix: Remove flaky timing test

VaggelisD:
These asserts keep messing with github's CI/CD tests; Removed them for now, open to discussion on whether these tests are needed at all, if they can be reincorporated on a different suite etc

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,4 +1,3 @@
-import time
 import unittest
 from unittest.mock import patch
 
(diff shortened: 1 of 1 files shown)


georgesittas:
@VaggelisD instead of completely removing these, let's instead make it so that we log a warning. That way we won't miss any accidental perf. regressions, while avoiding intermittent CI interruptions.

VaggelisD:
Right, I was thinking about that but I'm not sure if that approach will help us in a meaningful way, the timings we have now are arbitrary and they fail often on unrelated PRs; I can imagine that can happen for multiple reasons such as the congestion on the cluster/setup that they run on (work sharing), the Python version etc. 

I think the correct long-term solution is to build a perf suite running on a dedicated machine that compares the main branch with the target PR; To eliminate noise, we can schedule multiple runs per test case to extract an avg time, we can use specific versions etc.

georgesittas:
I meant we should log the delta if it's larger than some threshold, so that we can catch outliers easily. I don't think this is worth spending more time on.

georgesittas (approved):

PR #4377, opened 2024-11-12
Refactor: attach stamement comments consistently

### LINKED ISSUE #4376: Issue with Comment Handling in `Describe` Output
Hi,

I encountered an issue with the [`Describe`](https://sqlglot.com/sqlglot/expressions.html#Describe) functionality in `sqlglot`. The comments in the SQL query are not being preserved as expected.

Here’s a minimal example to illustrate the problem:

```python
import os
# os.environ['SQLGLOTRS_TOKENIZER'] = '0'

import importlib.metadata
import sqlglot

package_name = 'sqlglot'
version = importlib.metadata.version(package_name)
print(f"{package_name} version: {version}\n")

sql = '/*left*/ EXPLAIN SELECT /*hint*/ col FROM t1 /*right*/'

ast = sqlglot.parse(sql, read='mysql')

print(ast)
print(list(map(lambda x: x.sql(dialect='mysql'), ast)))
```

### Output:
```
sqlglot version: 25.30.0

[Describe(
  this=Select(
    expressions=[
      Column(
        this=Identifier(this=col, quoted=False))],
    from=From(
      this=Table(
        this=Identifier(this=t1, quoted=False),
        _comments=[
          right])),
    _comments=[
      hint]))]
['DESCRIBE /* hint */ SELECT col FROM t1 /* right */']
```

### Expected Output:
```sql
/*left*/ DESCRIBE /* hint */ SELECT col FROM t1 /* right */
```

### Actual Output:
```sql
DESCRIBE /* hint */ SELECT col FROM t1 /* right */
```

georgesittas:
Fixes #4376


sqlglot/parser.py
```diff
@@ -2659,7 +2659,7 @@ def parse_conditional_insert() -> t.Optional[exp.ConditionalInsert]:
         )
 
     def _parse_insert(self) -> t.Union[exp.Insert, exp.MultitableInserts]:
-        comments = ensure_list(self._prev_comments)
+        comments = []
```
VaggelisD, line 2662: Looks like the other statements completely removed their `comments`, did we have to preserve this one for some reason?

VaggelisD (approved):

sqlglot/parser.py
georgesittas: Yes, we're adding comments to the insert that aren't in the INSERT token.

PR #4387, opened 2024-11-14
Feat(risingwave): add support for SINK, SOURCE & other DDL properties

lin0303-siyuan:
This pr adds extended sqlglot support for risingwave. Specifically, `CREATE SOURCE` and `CREATE SINK` are supported for sqlglot risingwave dialect.


sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+
+            value = self._parse_bitwise() or self._parse_var(any_token=True)
+
+            # Transform the value to exp.Var if it was parsed as exp.Column(exp.Identifier())
+            if isinstance(value, exp.Column):
+                value = exp.var(value.name)
+
+            return self.expression(exp.Property, this=key, value=value)
+
+        def _parse_create(self) -> exp.Create | exp.Command:
```
georgesittas, line 120: We shouldn't copy and paste code, this needs to be refactored.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+            expressions = None
+            if not encode_property:
+                return None
+            if self._match(TokenType.L_PAREN, advance=False):
+                expressions = exp.Properties(expressions=self._parse_wrapped_properties())
+            if expressions:
+                encode_property.set("expressions", expressions)
+            return encode_property
+
+        def _parse_property(self) -> t.Optional[exp.Expression]:
```
georgesittas, line 76: Ditto about copy/paste.

sqlglot/tokens.py
```diff
@@ -930,6 +936,11 @@ class Tokenizer(metaclass=_Tokenizer):
         "USER-DEFINED": TokenType.USERDEFINED,
         "FOR VERSION": TokenType.VERSION_SNAPSHOT,
         "FOR TIMESTAMP": TokenType.TIMESTAMP_SNAPSHOT,
+        # Newly added for Risingwave
+        "SINK": TokenType.SINK,
+        "SOURCE": TokenType.SOURCE,
+        "ENCODE": TokenType.ENCODE,
+        "WATERMARK": TokenType.WATERMARK,
```
georgesittas, line 943: The encode token is unused, the watermark token is unnecessary (you can do `match_text_seq("WATERMARK")`). You only need `SINK` and `SOURCE` to add them in `CREATABLES` as far as I understand, and you shouldn't include them here -- instead, move them to Risingwave's `Tokenizer.KEYWORDS` mapping.

sqlglot/tokens.py
```diff
@@ -405,6 +405,12 @@ class TokenType(AutoName):
     TIMESTAMP_SNAPSHOT = auto()
     OPTION = auto()
 
+    # Newly added for Risingwave
+    SINK = auto()
+    SOURCE = auto()
+    ENCODE = auto()
+    WATERMARK = auto()
```
georgesittas, line 412: Remove these two.

sqlglot/tokens.py
```diff
@@ -405,6 +405,12 @@ class TokenType(AutoName):
     TIMESTAMP_SNAPSHOT = auto()
     OPTION = auto()
 
+    # Newly added for Risingwave
+    SINK = auto()
+    SOURCE = auto()
```
georgesittas, line 410: No need to comment, we don't do it anywhere else.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+                # Tailored for risingwave sink.
+                expression_sql = f"CREATE SINK{concurrently}{exists_sql} {this}{expression_sql}{properties_sql}{postexpression_props_sql}{index_sql}{no_schema_binding}{clone}"
+            else:
+                expression_sql = f"CREATE{modifiers} {kind}{concurrently}{exists_sql} {this}{properties_sql}{expression_sql}{postexpression_props_sql}{index_sql}{no_schema_binding}{clone}"
+            return self.prepend_ctes(expression, expression_sql)
+
+        def sql(
+            self,
+            expression: t.Optional[str | exp.Expression],
+            key: t.Optional[str] = None,
```
georgesittas, line 552: Ditto about copy-paste, please refactor all of these.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+
+            if property_cls == exp.FileFormatProperty:
+                return f"FORMAT {self.sql(expression, 'this')}"
+
+            if property_cls == exp.IncludeProperty:
+                return self._include_property_sql(expression)
+
+            return f"{property_name}={self.sql(expression, 'this')}"
+
+        def create_sql(self, expression: exp.Create) -> str:
```
georgesittas, line 434: Ditto.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+                property_sql_str = f"{property_sql_str} AS {column_alias}"
+            return property_sql_str
+
+        def with_properties(self, properties: exp.Properties) -> str:
+            with_sql = self.properties(
+                properties, prefix=self.seg(self.WITH_PROPERTIES_PREFIX, sep="")
+            )
+            return with_sql
+
+        def property_sql(self, expression: exp.Property) -> str:
```
georgesittas, line 413: Ditto.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+                property_sql_str = f"{property_sql_str} {header_inner_expect_type}"
+            if column_alias:
+                property_sql_str = f"{property_sql_str} AS {column_alias}"
+            return property_sql_str
+
+        def with_properties(self, properties: exp.Properties) -> str:
+            with_sql = self.properties(
+                properties, prefix=self.seg(self.WITH_PROPERTIES_PREFIX, sep="")
+            )
+            return with_sql
```
georgesittas, line 411: Ditto.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,588 @@
...
+                return None
+            return self.expression(exp.Watermark, column=column, expression=expr)
+
+        def _parse_field_def(self) -> t.Optional[exp.Expression]:
+            # First parse if it is watermark field before parsing column def.
+            return self._parse_watermark_field() or self._parse_column_def(
+                self._parse_field(any_token=True)
+            )
+
+        def _parse_column_def(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
```
georgesittas, line 314: Ditto

georgesittas (changes requested):
Hey @lin0303-siyuan, appreciate the PR, but this doesn't really meet the standards of the code base. You shouldn't blindly copy and paste code just to get things working. This doesn't happen throughout the code base for a reason, we try to only override the necessary sections to keep things tidied up.

I think the reason you're seeing CI failures is because you haven't included the following import in the risingwave dialect:

```python
```

I think you can't use the `|` syntax for type hints if you don't do this? Something like that.

lin0303-siyuan:
Hi @georgesittas @VaggelisD really appreciate your suggestion, I am going to fix the pr by following your guidance.

lin0303-siyuan:
Hi, I fixed some of the problems mentioned above, including unnecessary tokens, style and unit test issues, and some of the copy/paste codes. But still two functions: `_parse_create` and `create_sql` with comment `@FIXME` are not refactored yet. I currently have no idea on how this could be refactored, really appreciate your guidance on this.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,501 @@
...
+        CREATABLES = {
+            *Postgres.Parser.CREATABLES,
+            *DB_CREATABLES,
+        }
+
+        ID_VAR_TOKENS = {
+            *Postgres.Parser.ID_VAR_TOKENS,
+            *CREATABLES,
+        }
+
```
georgesittas, line 33: Let's delete these and add `TokenType.SINK` and `TokenType.SOURCE` in the base `DB_CREATABLES` set.

tests/dialects/test_risingwave.py
```diff
@@ -12,3 +12,17 @@ def test_risingwave(self):
                 "": "SELECT a FROM tbl FOR UPDATE",
             },
         )
+        # Test risingwave CREATE SOURCE.
```
georgesittas, line 15:
```suggestion
```

tests/dialects/test_risingwave.py
```diff
@@ -12,3 +12,17 @@ def test_risingwave(self):
...
+        self.validate_identity(
+            "CREATE SOURCE from_kafka (*, gen_i32_field INT AS int32_field + 2, gen_i64_field INT AS int64_field + 2, WATERMARK FOR time_col AS time_col - INTERVAL '5' SECOND) INCLUDE header foo varchar AS myheader INCLUDE key AS mykey WITH (connector='kafka', topic='my_topic') FORMAT PLAIN ENCODE PROTOBUF (A=1, B=2) KEY ENCODE PROTOBUF (A=3, B=4)"
+        )
+        # Test risingwave CREATE SINK.
```
georgesittas, line 19:
```suggestion
```

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,501 @@
...
+        def _parse_table_hints(self) -> t.Optional[t.List[exp.Expression]]:
+            # There is no hint in risingwave.
+            # Do nothing here to avoid WITH keywords conflict in CREATE SINK statement.
+            return None
+
+        def _parse_include(self) -> t.Optional[exp.Expression]:
```
georgesittas, line 39: This method looks too complicated, can you explain what the different args of `Include` represent? I think we can use `_parse_var_from_options` for `column_type`, for example. The rest can probably be cleaned up as well.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,501 @@
...
+                include_property.set("header_inner_expect_type", header_inner_expect_type)
+
+            return include_property
+
+        def _parse_encode_property(self, is_key: bool) -> t.Optional[exp.Expression]:
+            encode_var = self._parse_var_or_string()
+            encode_property: t.Optional[exp.Expression] = self.expression(
+                exp.EncodeProperty, this=encode_var, is_key=is_key
+            )
+            # encode_property: t.Optional[exp.Expression] = self._parse_property_assignment(exp.EncodeProperty)
```
georgesittas, line 72: Forgot to remove I guess?
```suggestion
```

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,501 @@
...
+            expressions = None
+            if not encode_property:
+                return None
+            if self._match(TokenType.L_PAREN, advance=False):
+                expressions = exp.Properties(expressions=self._parse_wrapped_properties())
+            if expressions:
+                encode_property.set("expressions", expressions)
```
georgesittas, line 79: What's the purpose of this branch? The only code path where `expressions` is truthy is the branch in L76-77.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,501 @@
...
+            if self._match_texts("ENCODE"):
+                return self._parse_encode_property(is_key=False)
+
+            if self._match_text_seq("KEY", "ENCODE"):
+                return self._parse_encode_property(is_key=True)
+
+            if self._match_texts("INCLUDE"):
+                return self._parse_include()
+
+            return super()._parse_property()
```
georgesittas, line 93: Can we get rid of this method override and use `PROPERTY_PARSERS` instead?

georgesittas (commented):
@lin0303-siyuan the remaining copied methods should be removed, we need to incorporate the new logic in the base classes somehow.

This version looks better but there are still some places that need to be refactored.

sqlglot/dialects/risingwave.py
```diff
@@ -1,6 +1,501 @@
...
+            self,
+            expression: t.Optional[str | exp.Expression],
+            key: t.Optional[str] = None,
+            comment: bool = True,
+        ) -> str:
+            if isinstance(expression, exp.Watermark):
+                sql = self._watermark_sql(expression)
+                return self.maybe_comment(sql, expression) if self.comments and comment else sql
+
+            return super().sql(expression, key, comment)
```
georgesittas, line 501: Delete this and rename `_watermark_sql` to `watermark_sql`.

lin0303-siyuan: Do you mean deleting this `sql` override?

sqlglot/dialects/risingwave.py
lin0303-siyuan: It would work for `ENCODE` and `INCLUDE`, but since `KEY ENCODE` is a text seq, it cannot be matched via `_match_texts`. May I ask is there some good solution?

georgesittas: You can add `"KEY"` and then try to match `"ENCODE"` -- if you manage to do it, then you can use the encode parser. Otherwise, you'd retreat to the previous token to unconsume `KEY` and return `None`.

sqlglot/dialects/risingwave.py
georgesittas: Yep.

lin0303-siyuan: Then how to parse watermark related sql?

georgesittas: What does parsing have to do with this override? This is in the generator class. The suggestion has to do with how the watermark expression is generated: you shouldn't have to override `sql` in order to specify how an expression is converted in to SQL, you won't see this pattern anywhere in the codebase. Instead, you do `<expression_name_all_lowercase>_sql` and that is automatically used to generate the corresponding expression, hence my suggestion in the first message in this thread.

sqlglot/dialects/risingwave.py
lin0303-siyuan: From risingwave `INCLUDE` clause, it should be used as `INCLUDE { header | key | offset | partition | timestamp } [AS <column_name>]`. One special case is `INCLUDE header 'foo' varchar AS myheader` where `foo` and `varchar` could be provided optionally here.
And it is parsed by risingwave here: https://github.com/risingwavelabs/risingwave/blob/main/src/sqlparser/src/parser.rs#L2607.
I then generally follow the logic. I try to see `_parse_var_from_options`, but I don't think it will work as the configuration is not fixed.

georgesittas:
Hey @lin0303-siyuan, I'll merge this and try to clean it up a bit. Will ping you when the implementation is ready so that you can follow it for any future extensions to the dialect. Appreciate the effort so far.

georgesittas:
Please look into https://github.com/tobymao/sqlglot/pull/4432.

lin0303-siyuan:
Hi @georgesittas really appreciate your help.

PR #4400, opened 2024-11-15
Chore: bump PYO3 to v"0.22.6"

MartinSahlen:
Just a small maintenance task, seeing that that the PYO3 version is >1 year out of date.

### DIFF
--- a/sqlglotrs/Cargo.lock
+++ b/sqlglotrs/Cargo.lock
@@ -9,12 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
-name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


MartinSahlen:
> Hey @MartinSahlen, thank you for the PR!
> 
> Are these changes due to deprecated APIs or are there new nuances / side-effects with the addition of `bound` functions & constructors?
> 
> For example, I could find the [deprecation warning](https://docs.rs/pyo3/latest/pyo3/struct.Py.html#method.as_ref) in `as_ref`'s documentation, but had a harder time for the other additions.

As far as I saw, these are necessary changes for the code to compile, i.e **get** is deprecated vs get_bound.  think the **get** thing still works but it needs to import the "gil-refs" feature etc - it breaks as is. I think the differences with the bound naming mainly in naming to make it clear that the objects are bound to a given python interpreter. the get and get_bound still passer the py object as a parameter.

There should not be any side effects, I think this just makes the code more up-to-date with PYO3 given the project is using a 1+ year old version. 

I ran all the tests locally and as the tests are running here on Actions

georgesittas (approved):
FYI, benchmarks look unaffected as well, I did 4 runs of this PR vs main and the times for `sqlglot` and `sqlglotrs` were essentially the same.

VaggelisD (approved):

PR #4401, opened 2024-11-15
Fix!: use select star when eliminating distinct on

agrigoroi-palantir:
Current approach of re-using the select from the inner query would break in a lot of cases (e.g. using table aliases `SELECT x.a`, aliasing/renaming columns `SELECT a AS c`, using an expression `SELECT x + y`)

There was a previous attempt to fix the same issue in #4286. this PR implements the simplest approach suggested in the discussion there (which is `SELECT *`)


georgesittas (commented):
Perhaps this approach is ok. An alternative would be to find all column references in the original projection list, make sure they're also projected, and then replicate the original projections in the outer query as well.

@tobymao @VaggelisD any thoughts on this?

tobymao:
this should be fine, i don't worry about unstable ordering

georgesittas (approved):
Sounds good -- LGTM 👍. Thanks!

georgesittas:
Ok I think there's an issue here after all. This change affects the relation's schema, because the `*` bubbles up `_row_count` into the outermost query, so you'll get +1 column compared to what the original query produced. This was a subtle one, I'll revert this and work on a proper fix.

agrigoroi-palantir:
Ahh - indeed. Sorry I've also missed that.

georgesittas:
No worries, I've fixed this in main. You can look into the recent commits if you want to see how.

PR #4406, opened 2024-11-16
feat: add support for specifying column in row_number function

GaliFFun:
Added support for specifying target column in `row_number` function
Some dialects such as ClickHouse allow this syntax: https://clickhouse.com/docs/en/sql-reference/window-functions/row_number


georgesittas (approved):
Thanks!

PR #4408, opened 2024-11-16
feat(clickhouse): transform function support

GaliFFun:
Current implementation of the `transform` function allows only to specify 2 arguments. But there are dialects other than Spark, for example, ClickHouse:

https://clickhouse.com/docs/en/sql-reference/functions/other-functions#transform

It accepts 3 or 4 arguments, and with current implementation, the parsing of such queries raises `ParseError`:

```
sqlglot.errors.ParseError: The number of provided arguments (3) is greater than the maximum number of supported arguments (2).
```

IMO, it would be better to accept a variable amount of arguments to support dialects other than Spark.

This is a breaking change since the `expression` is replaced with `expressions`.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -274,6 +274,7 @@ class ClickHouse(Dialect):
             "SHA512": lambda args: exp.SHA2(this=seq_get(args, 0), length=exp.Literal.number(512)),
             "EDITDISTANCE": exp.Levenshtein.from_arg_list,
             "LEVENSHTEINDISTANCE": exp.Levenshtein.from_arg_list,
+            "TRANSFORM": lambda args: exp.Anonymous(this="TRANSFORM", expressions=args),
         }
 
         AGG_FUNCTIONS = {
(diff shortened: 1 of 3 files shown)


georgesittas:
I'm skeptical about this, I don't think it's right. The ClickHouse variants are different than the Spark function, so there's no real gain by representing all of them using `exp.Transform`, e.g. you wouldn't enable transpilation. I'd suggest adding a `FUNCTIONS` entry in ClickHouse to parse `transform` into an `Anonymous` node for the time being.

GaliFFun:
That's a good point, updated my implementation

sqlglot/dialects/clickhouse.py
```diff
@@ -274,6 +274,7 @@ class Parser(parser.Parser):
             "SHA512": lambda args: exp.SHA2(this=seq_get(args, 0), length=exp.Literal.number(512)),
             "EDITDISTANCE": exp.Levenshtein.from_arg_list,
             "LEVENSHTEINDISTANCE": exp.Levenshtein.from_arg_list,
+            "TRANSFORM": lambda args: exp.Anonymous(this="TRANSFORM", expressions=args),
         }
```
georgesittas, line 278: Let's instead do:
```suggestion
        }
        FUNCTIONS.pop("TRANSFORM")
```

georgesittas (approved):

PR #4419, opened 2024-11-18
feat(duckdb)!: Add support for ATTACH/DETACH statements

VaggelisD:
This PR adds support for the following DuckDB statements:

1. `ATTACH`: 
```
ATTACH [DATABASE] [IF NOT EXISTS] <database-path> [AS alias] (<attach_options>)
```
2. `DETACH`: 
```
DETACH [DATABASE] [IF EXISTS] <database> 
```

Docs
---------
[DuckDB ATTACH/DETACH](https://duckdb.org/docs/sql/statements/attach.html#detach-syntax)


sqlglot/dialects/duckdb.py
```diff
@@ -992,3 +1021,23 @@ def regexpextract_sql(self, expression: exp.RegexpExtract) -> str:
             return self.func(
                 "REGEXP_EXTRACT", expression.this, expression.expression, group, params
             )
+
+        def attach_sql(self, expression: exp.Attach) -> str:
```
tobymao, line 1025: wouldn't this fail if we generate sqlglot dialect?

sqlglot/tokens.py
```diff
@@ -226,6 +226,7 @@ class TokenType(AutoName):
     ARRAY = auto()
     ASC = auto()
     ASOF = auto()
+    ATTACH = auto()
```
tobymao, line 229: these tokens need to be added to identity so you can do 'select 1 attach' right?

VaggelisD: Right, will add the relevant identity tests. Do note that DuckDB doesn't allow them as valid identifiers, so since they're only defined in `duckdb.py` I thought the current solution suffices:

```SQL
D select 1 attach;
Parser Error: syntax error at or near "attach"
LINE 1: select 1 attach;
                 ^
D select 1 detach;
Parser Error: syntax error at or near "detach"
LINE 1: select 1 detach;

D select 1 a;
┌───────┐
│   a   │
│ int32 │
├───────┤
│     1 │
└───────┘
```

georgesittas (approved):

PR #4426, opened 2024-11-19
fix!: Tokenize hints as comments

### LINKED ISSUE #4425: TokenError: Error tokenizing 'select /*+ ORDERED */* from dual'
The tokenizer fails to parse a hint followed by a `*` if it is not separated by a space:

    /*+ ... */* from dual

```
import sqlglot

sqlglot.parse_one(
    '''
       select /*+ ORDERED */* from dual   
    '''
)
```

    sqlglot.errors.TokenError: Error tokenizing 'lect /*+ ORDERED */* from dual

This is valid syntax.

If I add a space between inside of `*/*` such that it is `*/ *`, it works fine.

```
sqlglot.parse_one(
    '''
       select /*+ ORDERED */ * from dual   
    '''
)
```

It also works when changing `*/*` to `/*1`.

```
sqlglot.parse_one(
    '''
       select /*+ ORDERED */1 from dual   
    '''
)

```
Note that this only works for hints, not comments, so this is fine:

```
sqlglot.parse_one(
    '''
       select /*+ ORDERED */1 from dual   
    '''
)
```

My guess is that the `*/*` is confusing the tokenizer for the start of a comment/hint like `/*`.

Thank you.

VaggelisD:
Fixes #4425

Currently, hints are not tokenized as a single entity like we do for the comments (i.e through `_scan_comments()`), but instead the following token stream is produced:

```
>>> tokens = tokenize("SELECT /*+ hint */ * FROM t")
>>> for token in tokens:
   ... print(token)

<Token token_type: TokenType.SELECT, text: SELECT, line: 1, col: 6, start: 0, end: 5, comments: []>
<Token token_type: TokenType.HINT, text: /*+, line: 1, col: 10, start: 7, end: 9, comments: []>
<Token token_type: TokenType.VAR, text: hint, line: 1, col: 15, start: 11, end: 14, comments: []>
<Token token_type: TokenType.STAR, text: *, line: 1, col: 17, start: 16, end: 16, comments: []>
<Token token_type: TokenType.SLASH, text: /, line: 1, col: 18, start: 17, end: 17, comments: []>
<Token token_type: TokenType.STAR, text: *, line: 1, col: 20, start: 19, end: 19, comments: []>
<Token token_type: TokenType.FROM, text: FROM, line: 1, col: 25, start: 21, end: 24, comments: []>
<Token token_type: TokenType.VAR, text: t, line: 1, col: 27, start: 26, end: 26, comments: []>
```

For this reason, the hint is actually processed at parse time instead of tokenization time. This can lead to subtle bugs such as the one on the issue, where a slightly different pattern such as `SELECT /*+ hint */* FROM t` can throw off the tokenizer; In this case, we'd currently treat that as `SELECT /*+ * /*` i.e `SELECT <hint_start> * <comment_start> FROM t`.

This PR fixes this issue by:
- Treating hints as comments, so the entire hint text is processed by `scan_comments`
- Upon tokenizing a hint, we produce a `TokenType.HINT` with the payload/text attached to it's `comments`
- During parse time, we transform `TokenType.HINT` into an `exp.Hint` by parsing it's payload/text

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2974,11 +2974,10 @@ class Parser(metaclass=_Parser):
         # duckdb supports leading with FROM x
         from_ = self._parse_from() if self._match(TokenType.FROM, advance=False) else None
 
-        hint = self._parse_hint()
         if self._match(TokenType.SELECT):
             comments = self._prev_comments
 
-            hint = hint or self._parse_hint()
+            hint = self._parse_hint()
 
             if self._next and not self._next.token_type == TokenType.DOT:
                 all_ = self._match(TokenType.ALL)
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3262,8 +3262,8 @@ class Parser(metaclass=_Parser):
         return self.expression(exp.Hint, expressions=hints)
 
     def _parse_hint(self) -> t.Optional[exp.Hint]:
-        if self._match(TokenType.HINT) and self._prev.comments:
-            return exp.maybe_parse(self._prev.comments[0], into=exp.Hint, dialect=self.dialect)
+        if self._match(TokenType.HINT) and self._prev_comments:
+            return exp.maybe_parse(self._prev_comments[0], into=exp.Hint, dialect=self.dialect)
 
         return None
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -756,6 +756,7 @@ class Parser(metaclass=_Parser):
         exp.From: lambda self: self._parse_from(joins=True),
         exp.Group: lambda self: self._parse_group(),
         exp.Having: lambda self: self._parse_having(),
+        exp.Hint: lambda self: self._parse_hint_body(),
         exp.Identifier: lambda self: self._parse_id_var(),
         exp.Join: lambda self: self._parse_join(),
         exp.Lambda: lambda self: self._parse_lambda(),
@@ -2973,10 +2974,11 @@ class Parser(metaclass=_Parser):
         # duckdb supports leading with FROM x
         from_ = self._parse_from() if self._match(TokenType.FROM, advance=False) else None
 
+        hint = self._parse_hint()
         if self._match(TokenType.SELECT):
             comments = self._prev_comments
 
-            hint = self._parse_hint()
+            hint = hint or self._parse_hint()
 
             if self._next and not self._next.token_type == TokenType.DOT:
                 all_ = self._match(TokenType.ALL)
@@ -3226,21 +3228,42 @@ class Parser(metaclass=_Parser):
 
         return this
 
-    def _parse_hint(self) -> t.Optional[exp.Hint]:
-        if self._match(TokenType.HINT):
-            hints = []
+    def _parse_hint_fallback_to_string(self) -> t.Optional[exp.Hint]:
+        start = self._curr
+        while self._curr:
+            self._advance()
+
+        end = self._tokens[self._index - 1]
+        return exp.Hint(expressions=[self._find_sql(start, end)])
+
+    def _parse_hint_function_call(self) -> t.Optional[exp.Expression]:
+        return self._parse_function_call()
+
+    def _parse_hint_body(self) -> t.Optional[exp.Hint]:
+        start_index = self._index
+        should_fallback_to_string = False
+
+        hints = []
+        try:
             for hint in iter(
                 lambda: self._parse_csv(
-                    lambda: self._parse_function() or self._parse_var(upper=True)
+                    lambda: self._parse_hint_function_call() or self._parse_var(upper=True),
                 ),
                 [],
             ):
                 hints.extend(hint)
+        except ParseError:
+            should_fallback_to_string = True
 
-            if not self._match_pair(TokenType.STAR, TokenType.SLASH):
-                self.raise_error("Expected */ after HINT")
+        if should_fallback_to_string or self._curr:
+            self._retreat(start_index)
+            return self._parse_hint_fallback_to_string()
 
-            return self.expression(exp.Hint, expressions=hints)
+        return self.expression(exp.Hint, expressions=hints)
+
+    def _parse_hint(self) -> t.Optional[exp.Hint]:
+        if self._match(TokenType.HINT) and self._prev.comments:
+            return exp.maybe_parse(self._prev.comments[0], into=exp.Hint, dialect=self.dialect)
 
         return None
 
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -509,6 +509,8 @@ class _Tokenizer(type):
             ),
             "{#": "#}",  # Ensure Jinja comments are tokenized correctly in all dialects
         }
+        if klass.HINT_START in klass.KEYWORDS:
+            klass._COMMENTS[klass.HINT_START] = "*/"
 
         klass._KEYWORD_TRIE = new_trie(
             key.upper()
@@ -547,6 +549,9 @@ class _Tokenizer(type):
                 string_escapes_allowed_in_raw_strings=klass.STRING_ESCAPES_ALLOWED_IN_RAW_STRINGS,
                 nested_comments=klass.NESTED_COMMENTS,
                 hint_start=klass.HINT_START,
+                tokens_preceding_hint={
+                    _TOKEN_TYPE_TO_INDEX[v] for v in klass.TOKENS_PRECEDING_HINT
+                },
             )
             token_types = RsTokenTypeSettings(
                 bit_string=_TOKEN_TYPE_TO_INDEX[TokenType.BIT_STRING],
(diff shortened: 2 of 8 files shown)


sqlglot/tokens.py
```diff
@@ -954,7 +955,11 @@ class Tokenizer(metaclass=_Tokenizer):
     # Handle numeric literals like in hive (3L = BIGINT)
     NUMERIC_LITERALS: t.Dict[str, str] = {}
 
-    COMMENTS = ["--", ("/*", "*/")]
+    COMMENTS = [
+        "--",
+        ("/*", "*/"),
+        ("/*+", "*/"),
```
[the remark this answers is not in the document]
VaggelisD: We treated hints as supported on the base dialect up until now, e.g we had tests in `identity.sql` and in the base generator the relevant flag `QUERY_HINTS` is set to `True` by default, so with this we do not add any regressions for that matter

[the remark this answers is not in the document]
VaggelisD: For dialects that don't have hints `hint_sql` will log an `unsupported` error and will not render the hint (a relevant case is in `test_postgres.py`)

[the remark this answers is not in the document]
tobymao: perfect

sqlglot/tokens.py
```diff
@@ -1230,6 +1235,9 @@ def _scan_comment(self, comment_start: str) -> bool:
                 self._advance(alnum=True)
             self._comments.append(self._text[comment_start_size:])
 
+        if comment_start == "/*+":
```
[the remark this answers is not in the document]
tobymao: i guess it needs to be passed in and configurable?

tobymao (approved):

sqlglot/tokens.py
```diff
@@ -954,7 +955,11 @@ class Tokenizer(metaclass=_Tokenizer):
     # Handle numeric literals like in hive (3L = BIGINT)
     NUMERIC_LITERALS: t.Dict[str, str] = {}
 
-    COMMENTS = ["--", ("/*", "*/")]
+    COMMENTS = [
+        "--",
+        ("/*", "*/"),
+        ("/*+", "*/"),
+    ]
```
georgesittas, line 962: Since `/*+` was previously treated as a hint for all dialects, we don't have to add it in `COMMENTS`, but rather do it in the metaclass. That way you don't have to change MySQL, for example. It would look like this:

```python
        klass._COMMENTS = {
            **dict(
                (comment, None) if isinstance(comment, str) else (comment[0], comment[1])
                for comment in klass.COMMENTS
            ),
            "{#": "#}",  # Ensure Jinja comments are tokenized correctly in all dialects
            "/*+": "*/",  # Ensure hints are treated as comments and tokenized correctly in all dialects
        }
```

sqlglot/tokens.py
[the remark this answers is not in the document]
georgesittas: I agree about making `/*+` a constant, not sure if we need to make it configurable. Isn't this pretty much universal? Perhaps there's something like `/*!` in MySQL or somewhere else? I recall seeing something similar but need to search docs...

tests/dialects/test_postgres.py
```diff
@@ -819,6 +815,13 @@ def test_postgres(self):
             },
         )
 
+        self.validate_all(
+            "/*+ some comment*/ SELECT b.foo, b.bar FROM baz AS b",
+            write={
+                "postgres": UnsupportedError,
+            },
+        )
```
georgesittas, line 823: Why is this unsupported? This change doesn't look right, this isn't really a hint, right?

sqlglot/parser.py
```diff
@@ -3226,21 +3228,42 @@ def _parse_query_modifiers(
...
+            return self._parse_hint_fallback_to_string()
 
-            if not self._match_pair(TokenType.STAR, TokenType.SLASH):
-                self.raise_error("Expected */ after HINT")
+        return self.expression(exp.Hint, expressions=hints)
 
-            return self.expression(exp.Hint, expressions=hints)
+    def _parse_hint(self) -> t.Optional[exp.Hint]:
+        if self._match(TokenType.HINT) and self._prev.comments:
+            return exp.maybe_parse(self._prev.comments[0], into=exp.Hint, dialect=self.dialect)
```
georgesittas, line 3266:
```suggestion
    def _parse_hint(self) -> t.Optional[exp.Hint]:
        if self._match(TokenType.HINT) and self._prev_comments:
            return exp.maybe_parse(self._prev_comments[0], into=exp.Hint, dialect=self.dialect)
```

sqlglot/parser.py
```diff
@@ -2973,10 +2974,11 @@ def _parse_select(
         # duckdb supports leading with FROM x
         from_ = self._parse_from() if self._match(TokenType.FROM, advance=False) else None
 
+        hint = self._parse_hint()
```
georgesittas, line 2977: Why was this added? It looks like this affects the Postgres test. Is there documentation that backs this up?

georgesittas (commented):
Nice work @VaggelisD, left some comments. It looks like we haven't added a new test for this?

sqlglot/tokens.py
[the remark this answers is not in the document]
VaggelisD: Looks like `/*!` does exist for MySQL: https://dev.mysql.com/doc/refman/8.4/en/comments.html

tests/dialects/test_postgres.py
VaggelisD: If we add the hint tokens in `COMMENTS`, then this will be picked up as an `exp.Hint` in the respective dialect (e.g Postgres). If it's not supported, the roundtrip will yield unsupported through the generator flag `QUERY_HINTS`.

sqlglot/parser.py
VaggelisD: Ditto about the hint tokenization.

Let's assume we have the following query:
```SQL
/*+ comment */ SELECT ...
```

Before, this was treated as a comment so it'd be attached to the following `exp.Select` node, but now this'd be tokenized as a `TokenType.HINT`, so `exp.Select` needs to parse that before `TokenType.SELECT`

sqlglot/tokens.py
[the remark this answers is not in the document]
VaggelisD: Doesn't seem like a hint though, more like a MySQL specific code pass-through so the `/*+` should be universal

tests/dialects/test_postgres.py
georgesittas: Ok, this looks like an issue. I think for comments to be treated as hints, the previous token needs to be a `SELECT`. That's how it worked until today:

```python
        if self._match(TokenType.SELECT):
            comments = self._prev_comments

            hint = self._parse_hint()
```

Also, you need to make sure to _only_ use the last comment for a hint. Otherwise if there are actual comments that would be attached to the select, you don't want to treat them as hints. I'm not sure if you can achieve that somehow, but I'd test this more:

```

SELECT /* some ordinary comment */ /*+ some hint */ ...

SELECT /* some ordinary comment */
  /*+ some hint */
  ...

etc
```

sqlglot/parser.py
georgesittas: I replied to the other thread. I don't think this is right? Hints should be after the `SELECT` token IIRC.

tests/dialects/test_postgres.py
VaggelisD: Regarding the hint placement, I think this is right, but I guess the intention of this case was to test hint tokenization/parsing for dialects that do not support it e.g in Postgres, this would be treated as a regular comment so it could be placed _anywhere_, whereas now that's not the case. Is this an issue? Looks like an extreme edge case.

About the comments, I'll do more testing to make sure but I thought it wouldn't be an issue; the tokenizer's `scan_comment` will scan just the hint and attach it to the produced `TokenType.HINT`, so in general we should expect only 1 attached comment and that's on the hint token/expression

georgesittas: > Is this an issue? Looks like an extreme edge case.

Yeah, let's try to do this correctly without making assumptions and breaking valid queries:

```
georgesittas=# /*+ some comment */ select 1;
 ?column?
----------
        1
(1 row)
```

We should just preserve the previous behavior and only treat hint comments as hints if they're right after the `SELECT` token.

sqlglotrs/src/tokenizer.rs
```diff
@@ -395,6 +395,10 @@ impl<'a> TokenizerState<'a> {
                 .push(self.text()[comment_start_size..].to_string());
         }
 
+        if comment_start == self.settings.hint_start && self.tokens.last().is_some() && self.settings.tokens_preceding_hint.contains(&self.tokens.last().unwrap().token_type) {
```
georgesittas, line 398: Let's manually format this a bit, I think the pattern is to have && in the next line.

georgesittas (approved):
Thanks for addressing comments, one final nit on my end.

PR #4434, opened 2024-11-20
feat(snowflake): Support Snowflake syntax CREATE ... WITH TAG

### LINKED ISSUE #4427: Support Snowflake WITH TAG syntax
This library does not support snowflake's WITH TAG syntax for setting tags while creating an object.

For example,
```
sqlglot.parse_one("CREATE VIEW my_view WITH TAG (my_tag = 'tag')", dialect="snowflake")
'CREATE VIEW my_view WITH TAG (my_tag = 'tag')' contains unsupported syntax. Falling back to parsing as a 'Command'.
> Command(this=CREATE, expression=VIEW my_view WITH TAG (my_tag = 'tag'))
```
Note that multiple tags can be specified in the same statement.

Parse the above statement into something like:
```
Create(
  this=Table(
    this=Identifier(this=my_view, quoted=False)),
  kind=VIEW,
  actions=[tag=[
        Paren(
          this=EQ(
            this=Column(
              this=Identifier(this=my_tag, quoted=False)),
            expression=Literal(this=tag, is_string=True)))])
  ]
)
```
This was pulled from an `ALTER VIEW ... SET TAG` statement, idk if this is actually the correct Create object. My primary desire is to avoid the Command parsing fallback.

**Additional context**
Snowflake CREATE TABLE reference: https://docs.snowflake.com/en/sql-reference/sql/create-table
Example CREATE WAREHOUSE statement: https://docs.snowflake.com/en/user-guide/object-tagging

There are similar `WITH ...` syntaxes in snowflake, e.g. `WITH ROW ACCESS POLICY` and `WITH AGGREGATION POLICY`, or `WITH MASKING POLICY` and `WITH PROJECTION POLICY` for columns. This feature request is not specifically asking for support for that syntax.

asikowitz:
Fixes #4427 

This co-opts the `DictProperty` and `DictSubProperty` properties, originally created for [clickhouse](https://clickhouse.com/docs/en/sql-reference/statements/create/dictionary). Clickhouse and snowflake specify these key-value pairs with different syntax, so I tried to make the parsing / generation generic for various key-value syntaxes. This got a little awkward for generation because the separator and delimiter values have to be specified, and I didn't see any TokenType -> string method. I'm happy to instead create a less generic `SnowflakeTagListProperty` if that's preferred.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -465,6 +465,12 @@ class Snowflake(Dialect):
         PROPERTY_PARSERS = {
             **parser.Parser.PROPERTY_PARSERS,
             "LOCATION": lambda self: self._parse_location_property(),
+            "TAG": lambda self: self._parse_dict_property(
+                this="TAG",
+                has_kind=False,
+                separator=(TokenType.COMMA, ","),
+                delimiter=(TokenType.EQ, "="),
+            ),
         }
 
         TYPE_CONVERTERS = {
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2778,10 +2778,11 @@ class ClusteredByProperty(Property):
 
 
 class DictProperty(Property):
-    arg_types = {"this": True, "kind": True, "settings": False}
+    arg_types = {"this": True, "kind": False, "settings": False, "separator": False}
 
 
 class DictSubProperty(Property):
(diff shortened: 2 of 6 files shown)


georgesittas:
Hey @asikowitz, appreciate the PR. I think this approach is more complicated than needed. I checked this out in more detail and I think I can clean it up. Will go ahead and merge but refactor right after, feel free to take a look at it.

PR #4437, opened 2024-11-21
fix: Separate NORMALIZE_FUNCTIONS from name preservation

VaggelisD:
[Public Slack context](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1732134367565699)

The dialect flag `NORMALIZE_FUNCTIONS` is currently serving 2 purposes:
1. Forwarding it's value to the Generator flag `normalize_functions`, if the latter is not defined by the user
2. Preserving the original function name during parse time, which can be used to roundtrip different names for the same AST node

However, while the use case of (1) is distinct from (2), the Generator (`function_fallback_sql`) is using the same flag for both purposes which causes issues such as the following:

```Python3
>>> import sqlglot
>>> sqlglot.parse_one("SELECT JSON_VALUE(a, b)", read="bigquery").sql("bigquery")
'SELECT JSON_VALUE(a, b)'
>>> sqlglot.parse_one("SELECT JSON_VALUE(a, b)", read="bigquery").sql("bigquery", normalize_functions="lower")
'SELECT json_extract_scalar(a, b)'

```

This PR introduces a new dialect flag to separate these concerns: 

```Python3
>>> import sqlglot
>>> sqlglot.parse_one("SELECT JSON_VALUE(a, b)", read="bigquery").sql("bigquery")
'SELECT JSON_VALUE(a, b)'
>>> sqlglot.parse_one("SELECT JSON_VALUE(a, b)", read="bigquery").sql("bigquery", normalize_functions="lower")
'SELECT json_value(a, b)'
```


georgesittas (approved):

sqlglot/dialects/bigquery.py
```diff
@@ -333,6 +333,7 @@ class BigQuery(Dialect):
     HEX_LOWERCASE = True
     FORCE_EARLY_ALIAS_REF_EXPANSION = True
     EXPAND_ALIAS_REFS_EARLY_ONLY_IN_GROUP_BY = True
+    PRESERVE_ORIGINAL_NAMES = True
```
tobymao, line 336: why wouldn't you always preseve the original names?

PR #4439, opened 2024-11-21
Refactor(snowflake): clean up [WITH] TAG property / constraint

georgesittas:
Cleans up https://github.com/tobymao/sqlglot/commit/0acc248361f49f68f17d799cbaf6b3de06c57f7e, cc @asikowitz 

Reference: https://docs.snowflake.com/en/sql-reference/sql/create-table


asikowitz (approved):
Didn't see this logic for column constraints, looks a lot better, thanks!

PR #4440, opened 2024-11-21
fix: add return type

etonlels:
The `copy` method on `Expression`s does not have a return type specified. This PR adds one (and makes one additional change that was necessitated by the addition of the return type).


georgesittas (commented):
I'll get this in but fix the type hint of `column`, we want a sequence, not a collection.

PR #4447, opened 2024-11-26
fix(bigquery): Make JSONPathTokenizer more lenient for new standards

### LINKED ISSUE #4441: bigquery: json_value transpilation gives incorrect result or fails due to escaping
noticed via sqlmesh, which recently updated `sqlglot (25.19.0 -> 25.31.4)`.  

We have an expression like `select json_value('{"fu bar": 42}', "$.fu bar")` which runs correctly on bigquery yielding 42.   Previously this was transpiled unchanged, but now results in an incorrect selection (note the added `.`s):

```
>>> sqlglot.__version__
'25.31.4'
>>> print(sqlglot.transpile("""select json_value('{"fu bar": 42}', "$.fu bar")""", read="bigquery", write="bigquery")[0])
SELECT json_value('{"fu bar": 42}', '$.fu.bar')
```
This query now yields `null` instead of 42 in bigquery.

The space-containing-key can be escaped in bigquery with double-quotes for [json_value](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_value) which is apparently different from the bracketed escaping used by [json_extract](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract)... 🤯    So this also works successfully in bigquery: `SELECT json_value('{"fu bar": 42}', '$."fu bar"')`.

But it now gets transpiled to sql which throws an error, namely `Invalid token in JSONPath at: ['fu bar']`:

```
>>> print(sqlglot.transpile("""SELECT json_value('{"fu bar": 42}', '$."fu bar"')""", read="bigquery", write="bigquery")[0])
SELECT json_value('{"fu bar": 42}', '$[\'fu bar\']')
```

a workaround is apparently to use the deprecated json_extract[_scalar] functions.  i.e. the query `SELECT json_extract('{"fu bar": 42}', '$[\'fu bar\']')` works correctly and is transpiled unchanged:

```
>>> print(sqlglot.transpile("""SELECT json_extract('{"fu bar": 42}', '$[\\'fu bar\\']')""", read="bigquery", write="bigquery")[0])
SELECT json_extract('{"fu bar": 42}', '$[\'fu bar\']')
```

VaggelisD:
Fixes #4441

In BigQuery, there's a separation between the JSON extract functions with different escaping methods:
- New: `JSON_QUERY`, `JSON_VALUE`, `JSON_QUERY_ARRAY` use double quotes to escape
- Deprecated: `JSON_EXTRACT`, `JSON_EXTRACT_SCALAR`, `JSON_EXTRACT_ARRAY` use single quotes & brackets  to escape

After this PR, this is the BQ roundtrip:

- In:
```SQL
SELECT

  -- New
  JSON_QUERY('{" a b c ": {"d": 1}}', '$. a b c .d') as js_query,
  JSON_VALUE('{" a b c ": {"d": 2}}', '$. a b c .d') as js_value,
  JSON_QUERY_ARRAY('{" a b c ": {"d": [3]}}', '$. a b c .d') as js_query_array,

  -- Deprecated
  JSON_EXTRACT('{" a b c ": {"d": 4}}', '$. a b c .d') as js_extract,
  JSON_EXTRACT_SCALAR('{" a b c ": {"d": 5}}', '$. a b c .d') as js_extract_scalar,
  JSON_EXTRACT_ARRAY('{" a b c ": {"d": [6]}}', '$. a b c .d') as js_extract_array;
```

- Out:
```SQL
SELECT

  -- New
  JSON_QUERY('{" a b c ": {"d": 1}}', '$." a b c ".d') AS js_query,
  JSON_VALUE('{" a b c ": {"d": 2}}', '$." a b c ".d') AS js_value,
  JSON_QUERY_ARRAY('{" a b c ": {"d": [3]}}', '$." a b c ".d') AS js_query_array,

  -- Deprecated
  JSON_EXTRACT('{" a b c ": {"d": 4}}', '$[\' a b c \'].d') AS js_extract,
  JSON_EXTRACT_SCALAR('{" a b c ": {"d": 5}}', '$[\' a b c \'].d') AS js_extract_scalar,
  JSON_EXTRACT_ARRAY('{" a b c ": {"d": [6]}}', '$[\' a b c \'].d') AS js_extract_array
```

To make BQ capable of producing both, this PR:
- Forces SQLGlot/Python tokenizer for BQ's `JSONPathTokenizer` so that one can override the functions
- Enhances `Tokenizer` to also conditionally treat spaces as chars
- Adds `_json_extract_sql` which flips the flags `JSON_PATH_SINGLE_QUOTE_ESCAPE` & `JSON_PATH_BRACKETED_KEY_SUPPORTED` depending on which variant is being generated
 
Docs
--------
[BigQuery JSON_FUNCTIONS](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -4,7 +4,7 @@ import logging
 import re
 import typing as t
 
-from sqlglot import exp, generator, parser, tokens, transforms
+from sqlglot import exp, generator, parser, tokens, transforms, jsonpath
 from sqlglot.dialects.dialect import (
     Dialect,
     NormalizationStrategy,
@@ -38,6 +38,11 @@ if t.TYPE_CHECKING:
     from sqlglot.optimizer.annotate_types import TypeAnnotator
 
 logger = logging.getLogger("sqlglot")
+
+
+JSON_EXTRACT_TYPE = t.Union[exp.JSONExtract, exp.JSONExtractScalar, exp.JSONExtractArray]
+
+DQUOTES_ESCAPING_JSON_FUNCTIONS = ("JSON_QUERY", "JSON_VALUE", "JSON_QUERY_ARRAY")
 
 
 def _derived_table_values_to_unnest(self: BigQuery.Generator, expression: exp.Values) -> str:
@@ -324,6 +329,23 @@ def _build_contains_substring(args: t.List) -> exp.Contains | exp.Anonymous:
     return exp.Contains(this=this, expression=expr)
 
 
+def _json_extract_sql(self: BigQuery.Generator, expression: JSON_EXTRACT_TYPE) -> str:
+    name = (expression._meta and expression.meta.get("name")) or expression.sql_name()
+    upper = name.upper()
+
+    dquote_escaping = upper in DQUOTES_ESCAPING_JSON_FUNCTIONS
+
+    if dquote_escaping:
+        self.JSON_PATH_SINGLE_QUOTE_ESCAPE = self.JSON_PATH_BRACKETED_KEY_SUPPORTED = False
+
+    sql = rename_func(upper)(self, expression)
+
+    if dquote_escaping:
+        self.JSON_PATH_SINGLE_QUOTE_ESCAPE = self.JSON_PATH_BRACKETED_KEY_SUPPORTED = True
+
+    return sql
+
+
 class BigQuery(Dialect):
     WEEK_OFFSET = -1
     UNNEST_COLUMN_ONLY = True
@@ -334,6 +356,7 @@ class BigQuery(Dialect):
     FORCE_EARLY_ALIAS_REF_EXPANSION = True
     EXPAND_ALIAS_REFS_EARLY_ONLY_IN_GROUP_BY = True
     PRESERVE_ORIGINAL_NAMES = True
+    USE_RS_JSONPATH_TOKENIZER = False
 
     # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#case_sensitivity
     NORMALIZATION_STRATEGY = NormalizationStrategy.CASE_INSENSITIVE
@@ -462,6 +485,12 @@ class BigQuery(Dialect):
         KEYWORDS.pop("DIV")
         KEYWORDS.pop("VALUES")
         KEYWORDS.pop("/*+")
+
+    class JSONPathTokenizer(jsonpath.JSONPathTokenizer):
+        IDENTIFIERS = ["`", '"']
+
+        def _scan(self, until: t.Optional[t.Callable] = None, skip_spaces: bool = True) -> None:
+            return super()._scan(until=until, skip_spaces=False)
 
     class Parser(parser.Parser):
         PREFIXED_PIVOT_COLUMNS = True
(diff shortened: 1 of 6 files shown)


sqlglot/tokens.py
```diff
@@ -984,14 +986,20 @@ class Tokenizer(metaclass=_Tokenizer):
         "_peek",
         "_prev_token_line",
         "_rs_dialect_settings",
+        "_use_rs_tokenizer",
     )
 
-    def __init__(self, dialect: DialectType = None) -> None:
+    def __init__(
+        self, dialect: DialectType = None, USE_RS_TOKENIZER: t.Optional[bool] = None
```
georgesittas, line 993: Only constants should be capitalized:
```suggestion
        self, dialect: DialectType = None, use_rs_tokenizer: t.Optional[bool] = None
```

sqlglot/dialects/bigquery.py
```diff
@@ -463,6 +486,12 @@ class Tokenizer(tokens.Tokenizer):
         KEYWORDS.pop("VALUES")
         KEYWORDS.pop("/*+")
 
+    class JSONPathTokenizer(jsonpath.JSONPathTokenizer):
+        IDENTIFIERS = ["`", '"']
+
```
georgesittas, line 491: Why do we need this?

sqlglot/dialects/bigquery.py
```diff
@@ -40,6 +40,11 @@
 logger = logging.getLogger("sqlglot")
 
 
+JSON_EXTRACT_TYPE = t.Union[exp.JSONExtract, exp.JSONExtractScalar, exp.JSONExtractArray]
```
georgesittas, line 43: Should we just update the `dialect.py` constant?

sqlglot/tokens.py
```diff
@@ -525,6 +525,8 @@ def _quotes_to_format(
             if " " in key or any(single in key for single in klass.SINGLE_TOKENS)
         )
 
+        klass._use_rs_tokenizer = USE_RS_TOKENIZER
+
```
georgesittas, line 529: We shouldn't need to change the metaclass logic for this PR. Let's discuss in Slack.

sqlglot/dialects/bigquery.py
VaggelisD: This was initially like that, turns out that the `dialect.py` constant has to do with `exp.Binary` (e.g `self.binary(...)`) so adding `exp.JSONExtractArray` would lead to type errors.

sqlglot/tokens.py
VaggelisD: Actually this line can be removed since `USE_RS_TOKENIZER` is a top level constant

sqlglot/jsonpath.py
```diff
@@ -146,6 +146,21 @@ def _parse_bracket() -> exp.JSONPathPart:
 
         return node
 
+    def _parse_var_text() -> str:
```
georgesittas, line 149: May be worth putting a quick comment to explain this briefly.

georgesittas (approved):
LGTM 👍

tobymao (approved):

sqlglot/jsonpath.py
VaggelisD: Good idea, let me know if that will suffice.

PR #4448, opened 2024-11-26
Fix: parse DEFAULT in VALUES clause into a Var

### LINKED ISSUE #4446: “Default” is a keyword, but it was converted into a string value after translation.
```
Python 3.13.0 | packaged by Anaconda, Inc. | (main, Oct  7 2024, 16:25:56) [Clang 14.0.6 ] on darwin
>>> import sqlglot
>>> sqlglot.transpile("insert into t(i) values (default)", read="mysql", write="duckdb")[0]
'INSERT INTO t (i) VALUES ("default")'
```

“Default” is a keyword, but it was converted into a string value after translation.

georgesittas:
Fixes #4446

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -198,6 +198,7 @@ class Presto(Dialect):
     TYPED_DIVISION = True
     TABLESAMPLE_SIZE_IS_PERCENT = True
     LOG_BASE_FIRST: t.Optional[bool] = None
+    SUPPORTS_VALUES_DEFAULT = False
 
     TIME_MAPPING = MySQL.TIME_MAPPING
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/presto.py
```diff
@@ -198,6 +198,7 @@ class Presto(Dialect):
     TYPED_DIVISION = True
     TABLESAMPLE_SIZE_IS_PERCENT = True
     LOG_BASE_FIRST: t.Optional[bool] = None
+    SUPPORTS_VALUES_DEFAULT = False
```
georgesittas, line 201: Both Presto and Trino treat `DEFAULT` as an identifier: I tried running the issue's `INSERT` and got an identifier resolution error back. Only those two out of the engines I tested had this behavior, the other engines treat `DEFAULT` as a special keyword in the context of a `VALUES` clause.

tobymao (approved):

PR #4451, opened 2024-11-27
fix(spark)!: Make TIMESTAMP map to Type.TIMESTAMPTZ

### LINKED ISSUE #4442: Incorrect translation of timestamp casting between spark and duckdb
in

```python
from sqlglot import parse_one
parse_one("select cast(col as timestamp) from table limit 1", dialect='spark').sql(dialect='duckdb')
``` 
the output is: 
SELECT TRY_CAST(col AS TIMESTAMP) FROM table LIMIT 1

while the expected output should be:
SELECT TRY_CAST(col AS TIMESTAMPTZ) FROM table LIMIT 1 

as timestamp in spark represents TIMESTAMP WITH TIMEZONE

**Official Documentation**
1. spark (from databricks docs) 
image
2. duckdb - https://duckdb.org/docs/sql/data_types/overview.html

VaggelisD:
Fixes #4442

Spark defines it's `TIMESTAMP` to be a configurable session variable, with it's default value being timestamp with local time zone (`TIMESTAMP_LTZ`):

```SQL
spark-sql (default)> SELECT CAST('2024-11-27 15:03:20.603467+03' AS TIMESTAMP);
2024-11-27 14:03:20.603467

duckdb> SELECT CAST('2024-11-27 15:03:20.603467+03' AS TIMESTAMP);
┌────────────────────────────────────────────────────┐
│ CAST('2024-11-27 15:03:20.603467+03' AS TIMESTAMP) │
│                     timestamp                      │
├────────────────────────────────────────────────────┤
│ 2024-11-27 12:03:20.603467                         │
└────────────────────────────────────────────────────┘

duckdb> SELECT CAST('2024-11-27 15:03:20.603467+03' AS TIMESTAMPTZ);
┌───────────────────────────────────────────────────────────────────┐
│ CAST('2024-11-27 15:03:20.603467+03' AS TIMESTAMP WITH TIME ZONE) │
│                     timestamp with time zone                      │
├───────────────────────────────────────────────────────────────────┤
│ 2024-11-27 14:03:20.603467+02                                     │
└───────────────────────────────────────────────────────────────────┘
```

This PR maps Hive's `"TIMESTAMP" <-> Type.TIMESTAMPTZ`; This makes it so that:
- `TIMESTAMP` is roundtripped, thus preserving the session value. If the token was instead mapped to `Type.TIMESTAMP_LTZ`, we'd run the risk of breaking workloads that have reconfigured it to `TIMESTAMP_NTZ`
- Transpilation towards dialects is timezone-aware, as is shown above

Docs
--------
[Spark types](https://spark.apache.org/docs/3.5.2/sql-ref-datatypes.html)

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -257,7 +257,6 @@ class Hive(Dialect):
             "MSCK REPAIR": TokenType.COMMAND,
             "REFRESH": TokenType.REFRESH,
             "TIMESTAMP AS OF": TokenType.TIMESTAMP_SNAPSHOT,
-            "TIMESTAMP": TokenType.TIMESTAMPTZ,
             "VERSION AS OF": TokenType.VERSION_SNAPSHOT,
             "SERDEPROPERTIES": TokenType.SERDE_PROPERTIES,
         }
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -257,6 +257,7 @@ class Hive(Dialect):
             "MSCK REPAIR": TokenType.COMMAND,
             "REFRESH": TokenType.REFRESH,
             "TIMESTAMP AS OF": TokenType.TIMESTAMP_SNAPSHOT,
+            "TIMESTAMP": TokenType.TIMESTAMPTZ,
             "VERSION AS OF": TokenType.VERSION_SNAPSHOT,
             "SERDEPROPERTIES": TokenType.SERDE_PROPERTIES,
         }
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -14,6 +14,7 @@ from sqlglot.dialects.dialect import (
 )
 from sqlglot.dialects.hive import Hive
 from sqlglot.helper import seq_get, ensure_list
+from sqlglot.tokens import TokenType
 from sqlglot.transforms import (
     preprocess,
     remove_unique_constraints,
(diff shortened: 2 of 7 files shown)


georgesittas (approved):
LGTM, will wait for @tobymao to take a look as well. Not sure how this could affect SQLMesh. I'm thinking that if you had a `TIMESTAMP` column before, now it's gonna be `TIMESTAMPTZ` and the types won't match, so may need a migration or something?

VaggelisD:
Right, my main worry with this change was SQLMesh too, hopefully this approach is the path of least resistance.

tobymao:
why did you change hive is this correct?

tobymao:
why would sqlmesh break? doesn’t it round trip? can you confirm

georgesittas:
It round trips yeah, I was thinking about the change at the AST level, e.g. if we had `col_type == Type.TIMESTAMP` and `col_type` is now `TIMESTAMPTZ`, this would turn out false.

But re: Hive, I think you're right, this should go in either Spark2 or Spark - @VaggelisD I didn't notice this earlier, but Hive [mentions](https://issues.apache.org/jira/browse/HIVE-2272) this re: timestamps ([docs](https://cwiki.apache.org/confluence/display/hive/languagemanual+types#LanguageManualTypes-TimestampstimestampTimestamps)):

>  Timestamps are interpreted to be timezoneless and stored as an offset from the UNIX epoch. Convenience UDFs for conversion to and from timezones are provided (to_utc_timestamp, from_utc_timestamp).

tobymao (approved):

PR #4459, opened 2024-11-28
fix(presto): Remove parentheses from CURRENT_USER

### LINKED ISSUE #4458: Trino CURRENT_USER does not allow parenthesis
```python
from sqlglot import parse_one
from sqlglot.dialects import Trino

sql_in = parse_one("SELECT CURRENT_USER", dialect=Trino)
print(sql_in.sql(dialect=Trino))
```
```bash
SELECT CURRENT_USER()
```

https://prestodb.io/docs/current/functions/session.html
https://trino.io/docs/current/functions/session.html

I believe that the Postgres dialect already implements the desired workaround for this function
https://github.com/tobymao/sqlglot/blob/v25.32.1/sqlglot/dialects/postgres.py#L527

MikeWallis42:
Fixes #4458

### DIFF
--- a/tests/dialects/test_presto.py
+++ b/tests/dialects/test_presto.py
@@ -1079,7 +1079,6 @@ class TestPresto(Validator):
                 "snowflake": "CURRENT_USER()",  # Although the ANSI standard is CURRENT_USER
             },
             write={
-                "presto": "CURRENT_USER",
                 "trino": "CURRENT_USER",
                 "snowflake": "CURRENT_USER()",
             },
--- a/tests/dialects/test_presto.py
+++ b/tests/dialects/test_presto.py
@@ -1071,6 +1071,19 @@ class TestPresto(Validator):
                 "databricks": "REGEXP_EXTRACT('abc', '(a)(b)(c)', 0)",
             },
         )
+        self.validate_all(
+            "CURRENT_USER",
+            read={
+                "presto": "CURRENT_USER",
+                "trino": "CURRENT_USER",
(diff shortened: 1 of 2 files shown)


VaggelisD (approved):
Looks great, thank you! 

Regarding the Snowflake case, both variants are supported according to their [docs](https://docs.snowflake.com/en/sql-reference/functions/current_user) but SQLGlot will generate it as a function by default, should be fine for now.

tests/dialects/test_presto.py
```diff
@@ -1071,6 +1071,19 @@ def test_presto(self):
                 "databricks": "REGEXP_EXTRACT('abc', '(a)(b)(c)', 0)",
             },
         )
+        self.validate_all(
+            "CURRENT_USER",
+            read={
+                "presto": "CURRENT_USER",
```
VaggelisD, line 1077: One nit upon viewing this again, it's superfluous to add both (i.e `read` and `write`) Presto directions here; This is because `validate_all(query, read, write)` will make sure that:

- All `read` queries map to Presto's `query` _and_ 
- `query` will map to all `write` queries

That means we're essentially testing Presto (read) -> Presto (query) -> Presto (write), when just one direction would suffice.

MikeWallis42: All feedback is welcome 😄
I've been looking at it for 30 minutes or so and I think I finally understand.
So because the test class has its dialect set to presto, I can at least remove the write test case for it then.
What I really want to test is that, after being parsed, it remains unchanged.

I took inspiration from some of the other test cases above it so there's probably some more slimming down to be done. I'll remove my redundant test case.

PR #4462, opened 2024-12-01
feat(duckdb)!: Add support for EXPLAIN [ANALYZE] statement

mrjsj:
Adds support for EXPLAIN [ANALYZE] for DuckDB. Also translate DESCRIBE into EXPLAIN from other dialects. 

```python
>>> import duckdb
>>> import sqlglot 
>>> from sqlglot import exp
>>> 
>>> conn = duckdb.connect()
>>> conn.sql("CREATE TABLE t1 (id int);")
>>> query = "EXPLAIN ANALYZE SELECT * FROM t1;"
>>> parsed = sqlglot.parse_one(query, dialect="duckdb")
'EXPLAIN ANALYZE SELECT * FROM t1' contains unsupported syntax. Falling back to parsing as a 'Command'.
>>> replaced = exp.replace_tables(
...     parsed,
...     {"t2", "t1"}
... )
  File "/Users/mrjsj/Developer/sqlglot/sqlglot/expressions.py", line 8231, in replace_tables
    mapping = {normalize_table_name(k, dialect=dialect): v for k, v in mapping.items()}
AttributeError: 'set' object has no attribute 'items'
>>>
```

Let me know, if there's better approach - if we can somehow share the DESCRIBE/EXPLAIN like MySQL uses `TokenType.DESCRIBE` for both. However, DuckDB only supports EXPLAIN.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -768,7 +768,7 @@ class DuckDB(Dialect):
             "using",
             "order",
             "current_catalog",
-            "explain"
+            "explain",
         }
 
         UNWRAPPED_INTERVAL_VALUES = (exp.Literal, exp.Paren)
@@ -907,7 +907,7 @@ class DuckDB(Dialect):
                     bracket = f"({bracket})[1]"
 
             return bracket
-        
+
         def describe_sql(self, expression: exp.Describe) -> str:
             style = expression.args.get("style")
             style = f" {style}" if style else ""
@@ -917,7 +917,7 @@ class DuckDB(Dialect):
             format = f" {format}" if format else ""
 
             return f"EXPLAIN{style}{format} {self.sql(expression, 'this')}{partition}"
-        
+
         def explain_sql(self, expression: exp.Explain) -> str:
             style = expression.args.get("style")
             style = f" {style}" if style else ""
@@ -926,7 +926,7 @@ class DuckDB(Dialect):
             format = self.sql(expression, "format")
             format = f" {format}" if format else ""
 
-            return f"EXPLAIN{style}{format} {self.sql(expression, 'this')}{partition}"        
+            return f"EXPLAIN{style}{format} {self.sql(expression, 'this')}{partition}"
 
         def withingroup_sql(self, expression: exp.WithinGroup) -> str:
             expression_sql = self.sql(expression, "expression")
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -317,6 +317,7 @@ class DuckDB(Dialect):
             "CHARACTER VARYING": TokenType.TEXT,
             "DETACH": TokenType.DETACH,
             "EXCLUDE": TokenType.EXCEPT,
+            "EXPLAIN": TokenType.EXPLAIN,
             "LOGICAL": TokenType.BOOLEAN,
             "ONLY": TokenType.ONLY,
             "PIVOT_WIDER": TokenType.PIVOT,
@@ -767,6 +768,7 @@ class DuckDB(Dialect):
             "using",
             "order",
             "current_catalog",
+            "explain"
         }
 
         UNWRAPPED_INTERVAL_VALUES = (exp.Literal, exp.Paren)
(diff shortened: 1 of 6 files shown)


VaggelisD (commented):
Hey @mrjsj, thank you for the contribution!

Generally we should avoid copying & pasting code like that; This PR can be simplified by:
1. Reverting all the additions besides the tests (`exp.Explain`, `TokenType.EXPLAIN` etc)
2 Mapping `"EXPLAIN"` to `TokenType.DESCRIBE` like MySQL
3. ~~Swapping `"DESCRIBE"` with `"EXPLAIN"` for DuckDB at `describe_sql`, either through a parameter added to it (can default to `"DESCRIBE"`) or through a generator flag (take a look at similar flags such as `ARRAY_SIZE_NAME` or `PARSE_JSON`)~~ 


EDIT: I see that there's [DESCRIBE](https://duckdb.org/docs/guides/meta/describe.html#describing-a-query) in DuckDB which is used to analyze tables instead of query plans; In order to not break one or another statement, (3) could then resort to adding a new arg to `exp.Describe` and that would control whether we generate `DESCRIBE ...` or `EXPLAIN ...`. The important point here remains that we shouldn't create `exp.Explain` which is an identical copy of `exp.Describe.

mrjsj:
Thank for the feedback @VaggelisD. I appreciate how detailed your response is. I had a feeling there would be a better approach.

I will see to this - should I ping you when it's ready, or do you automatically get notified?

mrjsj:
Hi @VaggelisD 
I stepped back and thought a bit about it. In MySQL context, DESCRIBE and EXPLAIN are synonymous, however in DuckDB and Spark SQL they have completely different meanings, so wouldn't it make sense to handle it completely separate with a new TokenType and expression? DuckDB has a SUMMARIZE TokenType too, which, although similar, is not the same as DESCRIBE either.

Of course it shouldn't just be a copy of DESCRIBE as I've done in this first proposal - rather it should have args related to Spark SQL / DuckDB EXPLAINs optional parameters (styles) (EXTENDED, CODEGEN, COST, FORMATTED for Spark SQL), (ANALYZE for DuckDB).


| Statement | DuckDB | MySQL | Spark SQL |
| -- | -- | -- | -- |
| DESCRIBE | metadata | metadata | metadata | 
| EXPLAIN | query plan | metadata | query plan | 

Reference:
- https://spark.apache.org/docs/3.5.3/sql-ref-syntax-qry-explain.html
- https://spark.apache.org/docs/3.5.3/sql-ref-syntax-aux-describe-table.html
- https://duckdb.org/docs/guides/meta/describe
- https://duckdb.org/docs/guides/meta/summarize
- https://duckdb.org/docs/guides/meta/explain_analyze
- https://duckdb.org/docs/guides/meta/explain

VaggelisD:
Right, I see what you mean. From what I see, right now `EXPLAIN` defaults to an `exp.Command` (i.e just stored as plain text) with the exception of MySQL which defines it as a `DESCRIBE` synonym.

If we want to support `exp.Explain`, we should either:
1. Focus on a subset of the dialects (e.g DuckDB, Spark) of your choice (more on that later)
2. Do a full pass over the all the dialects (MySQL, Postgres, Snowflake, BigQuery, Presto/Trino, DuckDB, Spark/DB etc) and try to support them outright


I'd say a typical roadmap for such task would be to:
- Add `TokenType.EXPLAIN` and map it to `"EXPLAIN"`
- Make sure `TokenType.EXPLAIN` can still be used as an identifier by adding it to `parser.py::ID_VAR_TOKENS`; This is because dialects like Snowflake allow queries such as `SELECT 1 explain`
- Mimic `exp.Describe`'s implementation for the Parser/Generator "framework" code but starting from a clean base so all logic is specific to `EXPLAIN`
- If we go with (1), we should make sure that the (new) `_parse_explain()` is capable of falling back to `exp.Command` if our desired constructs are not matched (see `parse.py::_parse_as_command` and it's uses). To test this, we should add test cases for all the non-supported dialects (see uses of `validate_identity` with `check_command_warning` )


For a reference, check out the [GRANT PR](https://github.com/tobymao/sqlglot/pull/4138/files) for a similar implementation; Since it's scope was too large for a single PR, we opted for a similar approach to (1). `EXPLAIN` might also be a large undertaking to support immediately, so even if we start off small we can iterate on other dialects later.

mrjsj:
Thanks again the detailed feedback @VaggelisD .

I'll keep this short, but I will start out with implementing EXPLAIN for DuckDB and PySpark to lay the ground work.

VaggelisD:
Closing this preemptively due to inactivity. If you're working on this feel free to open a new PR, otherwise I might take a stab at implementing this from scratch soon.

mrjsj:
Hi @VaggelisD 
Apologies, I have been busy with work so didn't have much time to look more into it. 
Perfectly fine to close it - I'm not sure I'll have time to look into it next couple of days/weeks.

PR #4468, opened 2024-12-02
Feat: allow tables to be preserved in replace_table

georgesittas:
This will allow SQLMesh users to control which table references will be resolved to the db objects in the physical layer.

### DIFF
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -456,6 +456,20 @@ def first(it: t.Iterable[T]) -> T:
     return next(i for i in it)
 
 
+def to_bool(value: t.Optional[str | bool]) -> t.Optional[str | bool]:
+    if isinstance(value, bool) or value is None:
+        return value
+
+    # Coerce the value to boolean if it matches to the truthy/falsy values below
+    value_lower = value.lower()
+    if value_lower in ("true", "1"):
+        return True
+    elif value_lower in ("false", "0"):
+        return False
+
+    return value
+
+
 def merge_ranges(ranges: t.List[t.Tuple[A, A]]) -> t.List[t.Tuple[A, A]]:
     """
     Merges a sequence of ranges, represented as tuples (low, high) whose values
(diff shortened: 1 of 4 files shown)


tests/test_expressions.py
```diff
@@ -258,6 +258,14 @@ def test_replace_tables(self):
             'SELECT * FROM "my-project"."example"."table" /* example.table */',
         )
 
+        self.assertEqual(
+            exp.replace_tables(
+                parse_one("select * from example.table /* sqlglot.meta preserve_reference */"),
```
tobymao, line 263: what about 

sqlglot.meta replace=False

georgesittas: I thought about this as well. There are two reasons I went with `preserve_reference`:

1. We don't have logic to convert `"false"` and `"true"` to booleans in `Expression.add_comment` right now
2. Something like `replace` seemed a bit generic, whereas `preserve_reference` felt more explicit

None of these are super important, so happy to refactor if you prefer it. I guess if we went with your suggestion we'd also implement (1) to avoid comparing strings?

tobymao: i prefer replace because it matches the name

georgesittas: Refactored, please take another look.

sqlglot/helper.py
```diff
@@ -456,6 +456,20 @@ def first(it: t.Iterable[T]) -> T:
...
 
+def to_bool(value: t.Optional[str | bool]) -> t.Optional[str | bool]:
+    if isinstance(value, bool) or value is None:
+        return value
+
+    # Coerce the value to boolean if it matches to the truthy/falsy values below
+    value_lower = value.lower()
+    if value_lower in ("true", "1"):
+        return True
+    elif value_lower in ("false", "0"):
```
tobymao, line 467: don't need the elif, can just be if

tobymao (approved):

sqlglot/helper.py
georgesittas: Ah thanks, copy paste shenanigans.

PR #4471, opened 2024-12-03
feat(lineage): Extend lineage function to work with pivot operation

step4:
Extends the lineage functionality to work with pivot operation.
Works with pivot aggregations with and without alias.

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -281,11 +281,11 @@ def to_node(
             pivot = next(
                 (p for p in scope.pivots if p.alias_or_name == c.table and not p.unpivot), None
             )
-            if pivot:
+            if pivot and isinstance(pivot.args.get("columns"), list):
                 # The source is a pivot operation, so we need to trace back to the aggregated column
                 pivot_column_mapping = {}
                 pivot_aggs = len(pivot.expressions)
-                columns = pivot.args.get("columns")
+                columns = pivot.args["columns"]
 
                 for i, agg in enumerate(pivot.expressions):
                     agg_cols = list(agg.find_all(exp.Column))
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -311,11 +311,11 @@ def to_node(
                 else:
                     # The column is not in the pivot, so it must be an implicit column of the pivoted source.
                     # adapt column to be from the implicit pivoted source
-                    parent = pivot.parent.copy()
+                    parent = t.cast(exp.Table, pivot.parent.copy())
                     parent.args["pivots"] = []
-                    column: exp.Column = c.copy()
-                    column.args["table"] = parent.this
-                    downstream_columns.append(column)
+                    adapted_column = t.cast(exp.Column, c.copy())
+                    adapted_column.args["table"] = parent.this
+                    downstream_columns.append(adapted_column)
 
                 for downstream_column in downstream_columns:
                     table = downstream_column.table
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -276,8 +276,44 @@ def to_node(
                 reference_node_name=reference_node_name,
                 trim_selects=trim_selects,
             )
+
+        elif scope.pivots:
+            pivot = next((p for p in scope.pivots if p.alias_or_name == c.table), None)
+            if pivot:
+                # The source is a pivot operation, so we need to trace back to the aggregated column
+                pivot_column_mapping = {}
+                for agg in pivot.expressions:
+                    agg_columns = list(agg.find_all(exp.Column))
+                    literals = list(pivot.args.get("field").find_all(exp.Literal))
+                    for literal in literals:
+                        pivot_column_name = f"{literal.this}"
+                        if isinstance(agg, exp.Alias):
+                            pivot_column_name += f"_{agg.alias_or_name}"
+                        pivot_column_mapping[pivot_column_name] = agg_columns
+
+                for agg_column in pivot_column_mapping[c.alias_or_name]:
+                    table = agg_column.table
+                    source = scope.sources.get(table)
+                    if isinstance(source, Scope):
+                        to_node(
+                            agg_column.name,
+                            scope=source,
+                            scope_name=table,
+                            dialect=dialect,
+                            upstream=node,
+                            trim_selects=trim_selects,
+                        )
+                    else:
+                        source = source or exp.Placeholder()
+                        node.downstream.append(
+                            Node(
+                                name=agg_column.sql(comments=False),
+                                source=source,
+                                expression=source,
+                            )
+                        )
         else:
-            # The source is not a scope - we've reached the end of the line. At this point, if a source is not found
+            # The source is not a scope and the column is not in any pivot - we've reached the end of the line. At this point, if a source is not found
             # it means this column's lineage is unknown. This can happen if the definition of a source used in a query
             # is not passed into the `sources` map.
             source = source or exp.Placeholder()
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -288,24 +288,13 @@ def to_node(
                 columns = pivot.args["columns"]
                 columns_count = len(columns)
 
-                for i, agg in enumerate(pivot.expressions):
-                    agg_cols = list(agg.find_all(exp.Column))
-                    for col_index in range(i, columns_count, pivot_aggs_count):
-                        # e.g. "pivot (sum(value) as value_sum, max(price)) for category in ('a' as cat_a, 'b')"
-                        # For each aggregation function, the pivot creates a new column for each field in category combined with the aggfunc.
-                        # So the columns parsed have this order: cat_a_value_sum, cat_a, b_value_sum, b.
-                        # Because of this step wise manner the aggfunc 'sum(value) as value_sum' belongs to the column indices 0, 2,
-                        # and the aggfunc 'max(price)' without an alias belongs to the column indices 1, 3.
-                        # Here only the columns used in the aggregations are of interest in the lineage, so lookup the pivot column name
-                        # by index and map that with the columns used in the aggregation.
-                        pivot_column_mapping[columns[col_index].name] = agg_cols
-
-                for agg_column in pivot_column_mapping[c.alias_or_name]:
-                    table = agg_column.table
+                if c.alias_or_name not in [i.this for i in columns]:
+                    # The column is not in the pivot, so it must be an implicit column of the pivoted source.
+                    table = pivot.parent.name
                     source = scope.sources.get(table)
                     if isinstance(source, Scope):
                         to_node(
-                            agg_column.name,
+                            c.name,
                             scope=source,
                             scope_name=table,
                             dialect=dialect,
@@ -318,11 +307,48 @@ def to_node(
                         source = source or exp.Placeholder()
                         node.downstream.append(
                             Node(
-                                name=agg_column.sql(comments=False),
+                                name=f"{table}.{c.name}",  # TODO: probably a better way available
                                 source=source,
                                 expression=source,
                             )
                         )
+                else:
+                    # The column is in the pivot, so we need to trace back to the aggregation that created it.
+                    for i, agg in enumerate(pivot.expressions):
+                        agg_cols = list(agg.find_all(exp.Column))
+                        for col_index in range(i, columns_count, pivot_aggs_count):
+                            # e.g. "pivot (sum(value) as value_sum, max(price)) for category in ('a' as cat_a, 'b')"
+                            # For each aggregation function, the pivot creates a new column for each field in category combined with the aggfunc.
+                            # So the columns parsed have this order: cat_a_value_sum, cat_a, b_value_sum, b.
+                            # Because of this step wise manner the aggfunc 'sum(value) as value_sum' belongs to the column indices 0, 2,
+                            # and the aggfunc 'max(price)' without an alias belongs to the column indices 1, 3.
+                            # Here only the columns used in the aggregations are of interest in the lineage, so lookup the pivot column name
+                            # by index and map that with the columns used in the aggregation.
+                            pivot_column_mapping[columns[col_index].name] = agg_cols
+
+                    for agg_column in pivot_column_mapping[c.alias_or_name]:
+                        table = pivot.parent.alias_or_name
+                        source = scope.sources.get(table)
+                        if isinstance(source, Scope):
+                            to_node(
+                                agg_column.name,
+                                scope=source,
+                                scope_name=table,
+                                dialect=dialect,
+                                upstream=node,
+                                source_name=source_names.get(table) or source_name,
+                                reference_node_name=reference_node_name,
+                                trim_selects=trim_selects,
+                            )
+                        else:
+                            source = source or exp.Placeholder()
+                            node.downstream.append(
+                                Node(
+                                    name=agg_column.sql(comments=False),
+                                    source=source,
+                                    expression=source,
+                                )
+                            )
         else:
             # The source is not a scope and the column is not in any pivot - we've reached the end of the line. At this point, if a source is not found
             # it means this column's lineage is unknown. This can happen if the definition of a source used in a query
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -284,14 +284,14 @@ def to_node(
             if pivot:
                 # The source is a pivot operation, so we need to trace back to the aggregated column
                 pivot_column_mapping = {}
-                for agg in pivot.expressions:
-                    agg_columns = list(agg.find_all(exp.Column))
-                    literals = list(pivot.args.get("field").find_all(exp.Literal))
-                    for literal in literals:
-                        pivot_column_name = f"{literal.this}"
-                        if isinstance(agg, exp.Alias):
-                            pivot_column_name += f"_{agg.alias_or_name}"
-                        pivot_column_mapping[pivot_column_name] = agg_columns
+                pivot_aggs = len(pivot.expressions)
+                columns = pivot.args.get("columns")
+
+                for i, agg in enumerate(pivot.expressions):
+                    agg_cols = list(agg.find_all(exp.Column))
+                    for col in range(i, len(columns), pivot_aggs):
+                        # Go through the field_names in a step manner for this aggfunc
+                        pivot_column_mapping[columns[col].name] = agg_cols
 
                 for agg_column in pivot_column_mapping[c.alias_or_name]:
                     table = agg_column.table
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -286,12 +286,19 @@ def to_node(
                 pivot_column_mapping = {}
                 pivot_aggs_count = len(pivot.expressions)
                 columns = pivot.args["columns"]
+                columns_count = len(columns)
 
                 for i, agg in enumerate(pivot.expressions):
                     agg_cols = list(agg.find_all(exp.Column))
-                    for col in range(i, len(columns), pivot_aggs_count):
-                        # Go through the field_names in a step manner for this aggfunc
-                        pivot_column_mapping[columns[col].name] = agg_cols
+                    for col_index in range(i, columns_count, pivot_aggs_count):
+                        # e.g. "pivot (sum(value) as value_sum, max(price)) for category in ('a' as cat_a, 'b')"
+                        # For each aggregation function, the pivot creates a new column for each field in category combined with the aggfunc.
+                        # So the columns parsed have this order: cat_a_value_sum, cat_a, b_value_sum, b.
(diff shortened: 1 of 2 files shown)


sqlglot/lineage.py
```diff
@@ -276,8 +276,44 @@ def to_node(
                 reference_node_name=reference_node_name,
                 trim_selects=trim_selects,
             )
+
+        elif scope.pivots:
+            pivot = next((p for p in scope.pivots if p.alias_or_name == c.table), None)
```
georgesittas, line 281: This needs to check `p.unpivot` too, because `scope.pivots` contains `Pivot` expressions that also represent the `UNPIVOT` operator.

sqlglot/lineage.py
```diff
@@ -276,8 +276,44 @@ def to_node(
...
+                # The source is a pivot operation, so we need to trace back to the aggregated column
+                pivot_column_mapping = {}
+                for agg in pivot.expressions:
+                    agg_columns = list(agg.find_all(exp.Column))
+                    literals = list(pivot.args.get("field").find_all(exp.Literal))
+                    for literal in literals:
+                        pivot_column_name = f"{literal.this}"
+                        if isinstance(agg, exp.Alias):
+                            pivot_column_name += f"_{agg.alias_or_name}"
+                        pivot_column_mapping[pivot_column_name] = agg_columns
```
georgesittas, line 292: Is this necessary? Note that the parser takes care of inferring the pivot column names for each dialect, take a look here, it may be possible to simplify this section.

https://github.com/tobymao/sqlglot/blob/3945acc4a0dfd58147de929c9a2c71734d8f1ade/tests/test_parser.py#L596-L712

georgesittas:
Thanks for the PR btw! We'll take a closer look tomorrow.

sqlglot/lineage.py
step4: I know what you mean.
I could not find a more simplified way, because the `Pivot` expression has the `columns` in `args`, but these are only `Identifier`. And I do not know how to get from these identifiers to the expressions which contain the aggregates.

Maybe you can give me a hint to simplify this properly?

step4: Does it make sense to extend the parsing to keep that information together somewhere here:
https://github.com/tobymao/sqlglot/blob/a9dca8dd1b523efd703003694d4389f9af9d1a12/sqlglot/parser.py#L4090C1-L4104C42

georgesittas: Perhaps it does, @VaggelisD is also taking a look at this so we should have more constructive feedback soon.

VaggelisD: If I understand correctly, this could be improved if we take advantage of how `pivot.columns` are generated, e.g:

```
>>> sql = """
SELECT prop1_p as other_a
FROM (SELECT partname, price, quality FROM part)
PIVOT (
    AVG(price) AS p, MAX(quality) AS q FOR partname IN ('prop' AS prop1, 'rudder', 'test')
);
"""

>> ast = sqlglot.parse_one(sql, dialect="duckdb")
>> print(repr(ast))
Select(
  expressions=[...],
  from=From(...),
      pivots=[
        Pivot(
          ...,
          columns=[
            Identifier(this=prop1_p, quoted=False),
            Identifier(this=prop1_q, quoted=False),
            Identifier(this=rudder_p, quoted=False),
            Identifier(this=rudder_q, quoted=False),
            Identifier(this=test_p, quoted=False),
            Identifier(this=test_q, quoted=False)])])))
```

Note how the order is alternating between each aggfunc & field; That means that for each aggfunc, it'll prefix/suffix the the field name in a "step" manner. For instance, `AVG(price) AS p` suffixes the fields in indices `(0, 2, 4, ...)` and so on. The first for-loop could then be generalized like:

```Python3
        elif scope.pivots:
            pivot = next(
                (p for p in scope.pivots if p.alias_or_name == c.table and not p.unpivot), None
            )
            if pivot:
                # The source is a pivot operation, so we need to trace back to the aggregated column
                pivot_column_mapping = {}
                pivot_aggs = len(pivot.expressions)
                columns = pivot.args.get("columns")

                for i, agg in enumerate(pivot.expressions):
                    agg_cols = list(agg.find_all(exp.Column))
                    for col in range(i, len(columns), pivot_aggs):
                        # Go through the field_names in a step manner for this aggfunc
                        pivot_column_mapping[columns[col].name] = agg_cols

                for agg_column in pivot_column_mapping[c.alias_or_name]:
                    ...
```

With that change, all of the tests (including the new ones) pass, and it can even be further simplified.

step4: Yes that simplifies it quite a bit.
I added this change to the PR.

sqlglot/lineage.py
```diff
@@ -276,8 +276,46 @@ def to_node(
...
+                    source = scope.sources.get(table)
+                    if isinstance(source, Scope):
+                        to_node(
+                            agg_column.name,
+                            scope=source,
+                            scope_name=table,
+                            dialect=dialect,
+                            upstream=node,
+                            trim_selects=trim_selects,
+                        )
```
step4, line 307: I think I am missing here the `source_name` and `reference_node_name` fields.

Is it right that `reference_node_name` references a CTE name and `source_name` the table names?

If so what are the right values for a pivot operation here?

step4: @VaggelisD can you help me out here please?

georgesittas:
Hiya @step4, will take another look soon– thank you for your patience.

sqlglot/lineage.py
georgesittas: The `source_name` is the name of a source table that got expanded into an aliased derived table (subquery):

```sql
-- The original source name is t2
SELECT x FROM t2

-- After expanding t2, it's an aliased derived table; we can map _q_0 to t2, the source_name
SELECT x FROM (SELECT y FROM t1) _q_0
```

The `reference_node_name` is for the name of the CTE or the alias of the derived table a node is from.

If I'm not mistaken, we want to mimic what we do in the previous branch, but for the aggregated column?

```python
                source_name=source_names.get(table) or source_name,
                reference_node_name=reference_node_name,
```

step4: I added the the parameters and added asserts in tests

sqlglot/lineage.py
```diff
@@ -276,8 +276,46 @@ def to_node(
...
+
+        elif scope.pivots:
+            pivot = next(
+                (p for p in scope.pivots if p.alias_or_name == c.table and not p.unpivot), None
+            )
+            if pivot:
+                # The source is a pivot operation, so we need to trace back to the aggregated column
+                pivot_column_mapping = {}
+                pivot_aggs = len(pivot.expressions)
+                columns = pivot.args.get("columns")
```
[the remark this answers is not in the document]
step4: I renamed `pivot_aggs` and the problematic case should be handled now by checking if the column is a implicit column of the pivoted source first.

sqlglot/lineage.py
```diff
@@ -276,8 +276,46 @@ def to_node(
...
+
+                for i, agg in enumerate(pivot.expressions):
+                    agg_cols = list(agg.find_all(exp.Column))
+                    for col in range(i, len(columns), pivot_aggs):
+                        # Go through the field_names in a step manner for this aggfunc
+                        pivot_column_mapping[columns[col].name] = agg_cols
```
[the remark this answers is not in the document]
step4: I added a longer comment.
I hope its clearer now.

sqlglot/lineage.py
```diff
@@ -276,8 +276,46 @@ def to_node(
...
+
+                for agg_column in pivot_column_mapping[c.alias_or_name]:
```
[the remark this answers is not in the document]
step4: I added a whole section where it checks for the implicit column of the pivoted source first. Also added tests for this case.

But in case it is a implicit column, I could not use `c.sql()` because the column `c` belongs to the pivot table `_q_0` and changed the table name manually which I think there is a better way:
https://github.com/step4/sqlglot/blob/cb37b5619acb0e127b75496bef84d24198b5cc70/sqlglot/lineage.py#L307-L314

sqlglot/lineage.py
```diff
@@ -276,8 +276,46 @@ def to_node(
...
+                    table = agg_column.table
+                    source = scope.sources.get(table)
```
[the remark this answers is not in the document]
step4: Changes it to lookup the table name via the parent

sqlglot/lineage.py
```diff
@@ -276,8 +276,81 @@ def to_node(
...
+                (p for p in scope.pivots if p.alias_or_name == c.table and not p.unpivot), None
+            )
+            if pivot and isinstance(pivot.args.get("columns"), list):
+                # The source is a pivot operation, so we need to trace back to the aggregated column
+                pivot_column_mapping = {}
+                pivot_aggs_count = len(pivot.expressions)
+                columns = pivot.args["columns"]
+                columns_count = len(columns)
+
+                if c.alias_or_name not in [i.this for i in columns]:
```
georgesittas, line 291: Let's compute `column_name = c.name` once, after `columns_count` (there shouldn't be an alias), and then do:

```suggestion
                if any(column_name == pivoted_column.name for pivoted_column in columns):
```

sqlglot/lineage.py
```diff
@@ -276,8 +276,81 @@ def to_node(
...
+                    # The column is not in the pivot, so it must be an implicit column of the pivoted source.
+                    table = pivot.parent.name
```
georgesittas, line 293: Probably safer to do `table = pivot.parent.alias_or_name` here to account for pivoted derived tables as well. We already do this in L330.

sqlglot/lineage.py
```diff
@@ -276,8 +276,81 @@ def to_node(
...
+                        to_node(
+                            c.name,
+                            scope=source,
+                            scope_name=table,
+                            dialect=dialect,
+                            upstream=node,
+                            source_name=source_names.get(table) or source_name,
+                            reference_node_name=reference_node_name,
+                            trim_selects=trim_selects,
+                        )
```
georgesittas, line 305: It looks like there's some repetition in these `to_node` / `Node(...)` (below) calls. Is it possible to refactor this so we don't have to copy-paste the same kwargs every time?

sqlglot/lineage.py
```diff
@@ -276,8 +276,81 @@ def to_node(
                 reference_node_name=reference_node_name,
                 trim_selects=trim_selects,
             )
+
+        elif scope.pivots:
+            pivot = next(
+                (p for p in scope.pivots if p.alias_or_name == c.table and not p.unpivot), None
+            )
+            if pivot and isinstance(pivot.args.get("columns"), list):
```
georgesittas, line 284: If this branch is not taken, we won't get a placeholder for this column.

sqlglot/lineage.py
georgesittas: Can use `column_name` in L329 too.

tests/test_lineage.py
```diff
@@ -495,3 +495,82 @@ def test_node_name_doesnt_contain_comment(self) -> None:
...
+        FROM (select value,category from sample_data)
+        PIVOT (
+            sum(value)
+            FOR category IN ('a', 'b')
+        );
+        """
+        node = lineage("other_a", sql)
+
+        self.assertEqual(node.downstream[0].name, "_q_0.value")
+        self.assertEqual(node.downstream[0].downstream[0].name, "sample_data.value")
```
georgesittas, line 512: Is this correct? Shouldn't the first node downstream be the `PIVOT` source? I.e. if the pivot ends up being aliased to `_q_1`, you'd say that this name is `_q_1.a`, and then go to `_q_0.value` and finally to `sample_data.value`.

georgesittas: Ditto for the remaining tests as well.

georgesittas:
Hi @step4, are you blocked on this? Let us know if you need help, or perhaps we can try to take this to the finish line.

step4:
> Hi @step4, are you blocked on this? Let us know if you need help, or perhaps we can try to take this to the finish line.

Hi, I am currently traveling for work and will have time to work on it again at the weekend

georgesittas:
> > Hi @step4, are you blocked on this? Let us know if you need help, or perhaps we can try to take this to the finish line.
> 
> Hi, I am currently traveling for work and will have time to work on it again at the weekend

Sounds good, thanks for following up. Feel free to ping us here or in Slack if you wanna chat.

sqlglot/lineage.py
step4: Adapted the code to compute `column_name` once and use it down the line

sqlglot/lineage.py
step4: Changed it

sqlglot/lineage.py
step4: I refactored it with the list of `downstream_columns` which is filled either by the aggregations or the implicit source column. After that the `to_node`/`Node()` part is called once

sqlglot/lineage.py
step4: I precalculate the possible pivot before entering the if cases. With that the placeholder branch should be hit, if the there is no source found and there is not pivot

tests/test_lineage.py
step4: Not sure if I understand that correctly.

If I check how this sql is qualified:
```sql
SELECT 
    a as other_a
FROM (select value,category from sample_data)
PIVOT (
    sum(value)
    FOR category IN ('a', 'b')
);
```
to
```sql
SELECT 
    _q_1.a AS other_a 
FROM 
    (
        SELECT 
            sample_data.value AS value, 
            sample_data.category AS category 
        FROM 
            sample_data AS sample_data
    ) AS _q_0 
PIVOT (
    SUM(_q_0.value) 
    FOR _q_0.category IN ('a', 'b')
) AS _q_1
```

`_q_0` is the alias of the subquery and the pivot aggregation `SUM(_q_0.value)` uses the column `_q_0.value` which I would expect in the first downstream node.

Do I misunderstand something here?

sqlglot/lineage.py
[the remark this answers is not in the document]
step4: With the refactoring I adapt the column expression now to have the proper table.
in lines 314 - 317

tests/test_lineage.py
georgesittas: Yeah, my suggestion was to produce the following lineage:

```
             +~~~ this is missing right now
             v
other_a -> _q_1.a -> _q_0.value -> sample_data.value
```

I think we do something similar in this simpler example:

```
>>> node = lineage("a", "select _q_0.a as a from (select t.col as a from t) _q_0")
>>> node.downstream
[Node(name='_q_0.a', expression=Alias(
  this=Column(
    this=Identifier(this=col, quoted=False),
    table=Identifier(this=t, quoted=False)),
  alias=Identifier(this=a, quoted=False)), source=Select(
  expressions=[
    Alias(
      this=Column(
        this=Identifier(this=col, quoted=False),
        table=Identifier(this=t, quoted=False)),
      alias=Identifier(this=a, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=t, quoted=False),
      alias=TableAlias(
        this=Identifier(this=t, quoted=False))))), downstream=[Node(name='t.col', expression=Table(
  this=Identifier(this=t, quoted=False),
  alias=TableAlias(
    this=Identifier(this=t, quoted=False))), source=Table(
  this=Identifier(this=t, quoted=False),
  alias=TableAlias(
    this=Identifier(this=t, quoted=False))), downstream=[], source_name='', reference_node_name='')], source_name='', reference_node_name='_q_0')]
```

Notice how the first node downstream is `_q_0.a` .

sqlglot/lineage.py
```diff
@@ -276,8 +282,66 @@ def to_node(
...
+                        agg_cols = list(agg.find_all(exp.Column))
+                        for col_index in range(i, columns_count, pivot_aggs_count):
+                            # e.g. "pivot (sum(value) as value_sum, max(price)) for category in ('a' as cat_a, 'b')"
+                            # For each aggregation function, the pivot creates a new column for each field in category combined with the aggfunc.
+                            # So the columns parsed have this order: cat_a_value_sum, cat_a, b_value_sum, b.
+                            # Because of this step wise manner the aggfunc 'sum(value) as value_sum' belongs to the column indices 0, 2,
+                            # and the aggfunc 'max(price)' without an alias belongs to the column indices 1, 3.
+                            # Here only the columns used in the aggregations are of interest in the lineage, so lookup the pivot column name
+                            # by index and map that with the columns used in the aggregation.
+                            pivot_column_mapping[columns[col_index].name] = agg_cols
```
georgesittas, line 308: Most of this code looks loop-invariant. The only thing that blocks us from factoring it out and computing it once is `pivot` AFAICT, because we use `c.table` in its definition. 

You could do:

```python
        pivot = next((p for p in scope.pivots if p.alias_or_name and not p.unpivot), None)
```

And then compute `pivot_column_mapping` once and cache it so it can be reused in the following iterations. Possibly worth creating a helper function to do that so this section isn't as "polluted" with boilerplate.

Note that you'd also update the branch condition to make it `pivot and pivot.alias_or_name == c.table`.

sqlglot/lineage.py
```diff
@@ -276,8 +282,66 @@ def to_node(
...
+
+                    downstream_columns.extend(pivot_column_mapping[column_name])
+                else:
+                    # The column is not in the pivot, so it must be an implicit column of the pivoted source.
+                    # adapt column to be from the implicit pivoted source
+                    parent = t.cast(exp.Table, pivot.parent.copy())
+                    parent.args["pivots"] = []
+                    adapted_column = t.cast(exp.Column, c.copy())
+                    adapted_column.args["table"] = parent.this
+                    downstream_columns.append(adapted_column)
```
georgesittas, line 318: Why are we copying stuff here and removing pivots?

tests/test_lineage.py
step4: Now I understand what you mean.

In your example the lineage node name and source are:
```
   a       / select _q_0.a as a from (select t.col as a from t) _q_0
-> _q_0.a  / select t.col as a from t
-> t.col   / t AS t
```
So `_q_0` is a more specific scope.

The pivot example:
```
   other_a           / SELECT _q_1.a AS other_a FROM (SELECT sample_data.value AS value, sample_data.category AS category FROM sample_data AS sample_data) AS _q_0 PIVOT(SUM(_q_0.value) FOR _q_0.category IN ('a', 'b')) AS _q_1
-> _q_0.value        / SELECT sample_data.value AS value FROM sample_data AS sample_data
-> sample_date.value / sample_data AS sample_data
```
Here `_q_1` it not a more specific scope, because the pivot is not a scope itself. I can't select a proper more specific source for this step.
If I use the whole scope again I would get:
```
   other_a           / SELECT _q_1.a AS other_a FROM (SELECT sample_data.value AS value, sample_data.category AS category FROM sample_data AS sample_data) AS _q_0 PIVOT(SUM(_q_0.value) FOR _q_0.category IN ('a', 'b')) AS _q_1
-> _q_1.a            / SELECT _q_1.a AS other_a FROM (SELECT sample_data.value AS value, sample_data.category AS category FROM sample_data AS sample_data) AS _q_0 PIVOT(SUM(_q_0.value) FOR _q_0.category IN ('a', 'b')) AS _q_1
-> _q_0.value        / SELECT sample_data.value AS value FROM sample_data AS sample_data
-> sample_date.value / sample_data AS sample_data
```
But because the source is identical in both steps, that would end in an endless loop.

So I'm not sure how to continue from here

sqlglot/lineage.py
step4: From the example you have given
```sql
SELECT
  "_q_0"."EMPID" AS "EMPID"
FROM "QUARTERLY_SALES" AS "QUARTERLY_SALES"
PIVOT(SUM("QUARTERLY_SALES"."AMOUNT") FOR "QUARTERLY_SALES"."QUARTER" IN ('2023_Q1', '2023_Q2', '2023_Q3')) AS "_q_0"
ORDER BY
  "EMPID"
```
the column `EMPID` is an implicit column of `QUARTERLY_SALES`. But it is qualified with the pivoted table `_q_0`.
So for this case I rewrote Column expression so its `QUARTERLY_SALES.EMPID`

sqlglot/lineage.py
georgesittas: I think so, this assumption is also made at the `qualify` level, so if a query is not fully-qualified (which will probably be the case for queries with pivots), this won't work anyway.

tests/test_lineage.py
georgesittas: I see, that's fine then. If we do `other_a -> _q_0.value -> sample_data.value`, is the `PIVOT` operator present anywhere in the lineage tree? I'd like to be able to display that there's a pivot happening that leads to this column. If that's true, then we can leave it as is I think.

georgesittas:
I'll merge this and take it to the finish line, thank you for the contribution!

step4:
Thank you for taking over!

PR #4472, opened 2024-12-04
feat(postgres): Support generated columns

### LINKED ISSUE #4463: Postgres: Unable to parse STORE Generated Column on create table
# About
[Here's the Postgres docs on this syntax](https://www.postgresql.org/docs/current/ddl-generated-columns.html). It seems I can't parse stored column syntax, I noticed I had a similar issue for syntax for adding columns.

> Note I have tried this with the latest release, but also thank you for creating this incredible library 

### Example syntax

```sql
CREATE TABLE IF NOT EXISTS example_2 (
  col_a INT NOT NULL,
  col_b INT NOT NULL,
  col_c INT GENERATED ALWAYS AS (col_a + col_b), -- not supported in postgres
  col_d INT GENERATED ALWAYS AS (col_a + col_b) VIRTUAL, -- not supported in postgres
  col_e INT GENERATED ALWAYS AS (col_a + col_b) STORED
);
```

image

## Virtual Keyword

As mentioned [here](https://github.com/tobymao/sqlglot/pull/4465#pullrequestreview-2472079771) it likely makes sense to also add support for the `VIRTUAL` keyword here, it is an explicit version of the absence of the default behaviour, it's also mutually exclusive to the `STORED` keyword. 

## Dialects 

- [MySQL](https://dev.mysql.com/doc/refman/8.4/en/create-table-generated-columns.html) , support for both `stored` and `virtual` keyword
- [SQLite](https://www.sqlite.org/syntax/column-constraint.html) , support for both `stored` and `virtual` keyword
- [Postgres](https://www.postgresql.org/docs/current/ddl-generated-columns.html), support for `stored` but no support for `virtual` columns in tables
  - They said _"mate, use a view..."_, to use generated columns it seems you need to use the `stored` keyword
  - The postgresql docs say:
    > _Thus, a virtual generated column is similar to a view and a stored generated column is similar to a materialized view (except that it is always updated automatically). **PostgreSQL currently implements only stored generated columns**._
- [BigQuery](https://cloud.google.com/spanner/docs/generated-column/how-to), support for `stored` but no `virtual` explicit keyword 
  - Confusingly they suggest postgres supports this syntax in their docs (maybe they use to?)

### Dialects without support

- [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16#full-syntax)
- [Spark/DB](https://docs.databricks.com/en/delta/generated-columns.html#create-a-table-with-generated-columns)
- [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html)
- [Clickhouse](https://clickhouse.com/docs/en/sql-reference/statements/alter/column#materialize-column)

# The current behaviour

## Reproducible example

```python
from sqlglot import parse

parse("""
  CREATE TABLE IF NOT EXISTS example (
    col_a INT NOT NULL,
    col_b INT NOT NULL,
    col_c INT GENERATED ALWAYS AS (col_a + col_b) STORED
  );
""", read='postgres')
```

## Example stacktrace

```
  ...
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/__init__.py", line 102, in parse
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/dialects/dialect.py", line 919, in parse
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 1398, in parse
    return self._parse(
           ^^^^^^^^^^^^
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 1467, in _parse
    expressions.append(parse_method(self))
                       ^^^^^^^^^^^^^^^^^^
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 1699, in _parse_statement
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 779, in <lambda>
    TokenType.CREATE: lambda self: self._parse_create(),
                                   ^^^^^^^^^^^^^^^^^^^^
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 1854, in _parse_create
    this = self._parse_schema(this=table_parts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 5379, in _parse_schema
    self._match_r_paren()
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 7172, in _match_r_paren
  File "/Users/angus/code/jupyter/env/lib/python3.12/site-packages/sqlglot/parser.py", line 1511, in raise_error
    raise error
sqlglot.errors.ParseError: Expecting ). Line 4, Col: 53.
  ample (
  col_a INT NOT NULL,
  col_b INT NOT NULL,
  col_c INT GENERATED ALWAYS AS (col_a + col_b) STORED
);
```

VaggelisD:
Fixes #4463

Postgres supports the following column constraints for `CREATE TABLE` DDLs:

```
where column_constraint is:

[ CONSTRAINT constraint_name ]
{ 
  ...
  GENERATED ALWAYS AS ( generation_expr ) STORED |
  GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ] |
  ...
}
```

The former defines a _generated_ column while the latter an _identity_ one. 

This PR overrides `parser.py::_parse_generated_as_identity()` which will transform the produced `exp.GeneratedAsIdentityColumnConstraint` into an `exp.ComputedColumnConstraint()`, as is already happening on `spark.py`.

> [!WARNING]  
> Something that stands out as bad naming/design is the fact that `exp.GeneratedAsIdentityColumnConstraint` should concern only **IDENTITY** columns, but overtime it has grown to parse different variants of the `GENERATED` syntax even if semantically they're not equivalent. 
>
> This leads to the "hacky" solution of having to transform it to `exp.ComputedColumnConstraint` _after_ it has been parsed, as is happening in the PR. Should we make an effort to deduplicate the two? 

Docs
--------
[PSQL CREATE TABLE](https://www.postgresql.org/docs/current/sql-createtable.html) | [PSQL GENERATED columns](https://www.postgresql.org/docs/current/ddl-generated-columns.html) | [PSQL IDENTITY columns](https://www.postgresql.org/docs/current/ddl-identity-columns.html)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -476,6 +476,20 @@ class Postgres(Dialect):
                 and self.dialect.to_json_path(self._parse_bitwise()),
             )
 
+        def _parse_generated_as_identity(
+            self,
+        ) -> (
+            exp.GeneratedAsIdentityColumnConstraint
+            | exp.ComputedColumnConstraint
+            | exp.GeneratedAsRowColumnConstraint
+        ):
+            this = super()._parse_generated_as_identity()
+
+            if self._match_texts(("STORED")):
+                this = self.expression(exp.ComputedColumnConstraint, this=this.expression)
+
+            return this
+
     class Generator(generator.Generator):
         SINGLE_STRING_INTERVAL = True
         RENAME_TABLE_WITH_DB = False
@@ -691,3 +705,6 @@ class Postgres(Dialect):
                 if isinstance(seq_get(exprs, 0), exp.Select)
                 else f"{self.normalize_func('ARRAY')}[{self.expressions(expression, flat=True)}]"
             )
+
+        def computedcolumnconstraint_sql(self, expression: exp.ComputedColumnConstraint) -> str:
+            return f"GENERATED ALWAYS AS ({self.sql(expression, 'this')}) STORED"
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -485,7 +485,7 @@ class Postgres(Dialect):
         ):
             this = super()._parse_generated_as_identity()
 
-            if self._match_texts(("STORED")):
+            if self._match_text_seq("STORED"):
                 this = self.expression(exp.ComputedColumnConstraint, this=this.expression)
 
             return this
--- a/tests/dialects/test_postgres.py
+++ b/tests/dialects/test_postgres.py
@@ -1047,6 +1047,7 @@ class TestPostgres(Validator):
         self.validate_identity("CREATE TABLE tbl (col INT UNIQUE NULLS NOT DISTINCT DEFAULT 9.99)")
         self.validate_identity("CREATE TABLE tbl (col UUID UNIQUE DEFAULT GEN_RANDOM_UUID())")
         self.validate_identity("CREATE TABLE tbl (col UUID, UNIQUE NULLS NOT DISTINCT (col))")
+        self.validate_identity("CREATE TABLE tbl (col_a INT GENERATED ALWAYS AS (1 + 2) STORED)")
 
         self.validate_identity("CREATE INDEX CONCURRENTLY ix_table_id ON tbl USING btree(id)")
         self.validate_identity(


sqlglot/dialects/postgres.py
```diff
@@ -476,6 +476,20 @@ def _parse_jsonb_exists(self) -> exp.JSONBExists:
...
+        def _parse_generated_as_identity(
+            self,
+        ) -> (
+            exp.GeneratedAsIdentityColumnConstraint
+            | exp.ComputedColumnConstraint
+            | exp.GeneratedAsRowColumnConstraint
+        ):
+            this = super()._parse_generated_as_identity()
+
+            if self._match_texts(("STORED")):
```
georgesittas, line 488:
```suggestion
            if self._match_text_seq("STORED"):
```

georgesittas (approved):

AKST:
Thanks @VaggelisD, if following a similar approach is fine, I don't think it would take much for me to add this to MySQL/SQLite to add support of the virtual keyword, happy to add that 🙂 

However I can see this is not a general approach but an approach tailored to Postgres. I am correct to say the choice to go with this less general approach is a reflection of an established users issue for Postgres (my original issue #4463) & the absence of one for SQLite/MySQL? Had that not been the case and there was more available bandwidth, a more general approach would have been favoured to also handle MySQL & SQLite?

I ask, because if that's the case it may not make sense to also do this for MySQL & SQLite if a more general approach is warranted.

AKST:
That's fair, I'm not sure I would be up to the task of refactoring/decoupling those 2 expressions. Is it worth following your approach here for SQLite & MySQL or just waiting till that refactor happens?

PR #4473, opened 2024-12-04
Feat!(snowflake): improve transpilation of TO_TIMESTAMP* variants

georgesittas:
I realized that Snowflake requires a constant for the 2nd argument of `TO_TIMESTAMP*` when it detects that the used variant is `timestampFunction ( <numeric_expr> [ , <scale> ] )` ([docs](https://docs.snowflake.com/en/sql-reference/functions/to_timestamp)), so we can leverage this constraint to improve the transpilation.

Screenshot 2024-12-04 at 12 35 23 PM

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -41,9 +41,12 @@ def _build_datetime(
 ) -> t.Callable[[t.List], exp.Func]:
     def _builder(args: t.List) -> exp.Func:
         value = seq_get(args, 0)
-        int_value = value is not None and is_int(value.name)
+        scale_or_fmt = seq_get(args, 1)
 
-        if isinstance(value, exp.Literal):
+        int_value = value is not None and is_int(value.name)
+        int_scale_or_fmt = scale_or_fmt is not None and is_int(scale_or_fmt.name)
+
+        if isinstance(value, exp.Literal) or (value and scale_or_fmt):
             # Converts calls like `TO_TIME('01:02:03')` into casts
             if len(args) == 1 and value.is_string and not int_value:
                 return (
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -55,11 +58,11 @@ def _builder(args: t.List) -> exp.Func:
             # Handles `TO_TIMESTAMP(str, fmt)` and `TO_TIMESTAMP(num, scale)` as special
             # cases so we can transpile them, since they're relatively common
             if kind == exp.DataType.Type.TIMESTAMP:
-                if int_value and not safe:
+                if not safe and (int_value or int_scale_or_fmt):
                     # TRY_TO_TIMESTAMP('integer') is not parsed into exp.UnixToTime as
                     # it's not easily transpilable
-                    return exp.UnixToTime(this=value, scale=seq_get(args, 1))
-                if not is_float(value.this):
+                    return exp.UnixToTime(this=value, scale=scale_or_fmt)
```
VaggelisD, line 64: Is `exp.UnixToTime`'s  scale compatible with Snowflake's scale? Should we add any tests for that matter?

tests/dialects/test_snowflake.py
```diff
@@ -670,6 +670,15 @@ def test_snowflake(self):
...
         )
+        self.validate_all(
+            "SELECT TO_TIMESTAMP(col, 'DD-MM-YYYY HH12:MI:SS') FROM t",
+            write={
+                "bigquery": "SELECT PARSE_TIMESTAMP('%d-%m-%Y %I:%M:%S', col) FROM t",
+                "duckdb": "SELECT STRPTIME(col, '%d-%m-%Y %I:%M:%S') FROM t",
+                "snowflake": "SELECT TO_TIMESTAMP(col, 'DD-mm-yyyy hh12:mi:ss') FROM t",
+                "spark": "SELECT TO_TIMESTAMP(col, 'dd-MM-yyyy hh:mm:ss') FROM t",
+            },
+        )
```
VaggelisD, line 681: Should we add a similar test with format being a column?

sqlglot/dialects/snowflake.py
```diff
@@ -41,9 +41,12 @@ def _build_datetime(
 ) -> t.Callable[[t.List], exp.Func]:
     def _builder(args: t.List) -> exp.Func:
         value = seq_get(args, 0)
+        scale_or_fmt = seq_get(args, 1)
+
         int_value = value is not None and is_int(value.name)
+        int_scale_or_fmt = scale_or_fmt is not None and is_int(scale_or_fmt.name)
 
-        if isinstance(value, exp.Literal):
+        if isinstance(value, exp.Literal) or (value and scale_or_fmt):
```
VaggelisD, line 49: Also a nit here, would it be a bit more readable to swap the OR rhs here with `or len(args) == 2)`? This hints that we'd be treating the 2-arg version more directly, instead of having to understand the conjuction

georgesittas: I intentionally implemented the condition this way, because mypy yells at the following `value.is_string` otherwise.

tests/dialects/test_snowflake.py
georgesittas: The rest of the engines don't support non-constant formats, so I only added an identity test.

sqlglot/dialects/snowflake.py
georgesittas: Yep, the scale units are pretty much the same, not seeing much value in testing this separately for the time being.

PR #4474, opened 2024-12-04
feat(optimizer)!: Alias expanded USING STRUCT fields

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3465

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -223,10 +223,10 @@ def _expand_using(scope: Scope, resolver: Resolver) -> t.Dict[str, t.Any]:
                 replacement: exp.Expression = exp.func("coalesce", *coalesce_args)
 
                 if isinstance(column.parent, exp.Select):
-                    # Ensure the USING column keeps it's output name if it's projected
+                    # Ensure the USING column keeps its name if it's projected
                     replacement = alias(replacement, alias=column.name, copy=False)
                 elif isinstance(column.parent, exp.Struct):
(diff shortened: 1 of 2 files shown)


sqlglot/optimizer/qualify_columns.py
```diff
@@ -220,11 +220,18 @@ def _update_source_columns(source_name: str) -> None:
...
-                replacement = exp.func("coalesce", *coalesce_args)
+                replacement: exp.Expression = exp.func("coalesce", *coalesce_args)
 
-                # Ensure selects keep their output name
                 if isinstance(column.parent, exp.Select):
+                    # Ensure the USING column keeps it's output name if it's projected
                     replacement = alias(replacement, alias=column.name, copy=False)
+                elif not isinstance(column.parent, exp.PropertyEQ) and column.find_ancestor(
+                    exp.Struct
+                ):
```
georgesittas, line 230: Can't we just do `if isinstance(column.parent, exp.Struct)`?

VaggelisD: Yeah I think that simplifies the condition, also doesn't make a difference for nested structs

georgesittas (approved):
This needs to be marked as breaking. Is it universally correct to inject a `PropertyEQ` in that `Struct`?

VaggelisD:
>  Is it universally correct to inject a PropertyEQ in that Struct?

Looks like it, tested a bunch of dialects and they all have `PropertyEq`. Also, by searching `exp.PropertyEq` I see that it comes up on STRUCT contexts around the main dialects.

Should we make it BQ specific for the time being?

PR #4480, opened 2024-12-05
fix(bigquery): Do not generate NULL ordering on Windows

### LINKED ISSUE #4478: Wrong result for transpiling `OVER (ORDER BY .. )` from `trino` to `bigquery`
It can be reproduced by the following code:
```python
import sqlglot

sql = """
WITH t1 as (
  select 1 f1, 2 f2 union all select 2 f1, 4 f2 union all select 3 f1, 6 f2
)
select sum(f1) over (order by f2) from t1;
"""
result = sqlglot.transpile(sql, read="trino", write="bigquery")[0]
print(result)
# the output is
# WITH t1 AS (SELECT 1 AS f1, 2 AS f2 UNION ALL SELECT 2 AS f1, 4 AS f2 UNION ALL SELECT 3 AS f1, 6 AS f2)
# SELECT SUM(f1) OVER (ORDER BY f2 NULLS LAST) FROM t1
```
This SQL isn't valid for BigQuery. It will get the error message from BigQuery:
```
NULLS LAST not supported with ascending sort order in RANGE clauses of analytic functions.
```
It seems that `NULLS LAST` shouldn't be added in the `OVER` clause.

If I don't assign the `read` dialect, it works well.
```python
import sqlglot

sql = """
WITH t1 as (
  select 1 f1, 2 f2 union all select 2 f1, 4 f2 union all select 3 f1, 6 f2
)
select sum(f1) over (order by f2) from t1;
"""
result = sqlglot.transpile(sql, read=None, write="bigquery")[0]
print(result)
# WITH t1 AS (SELECT 1 AS f1, 2 AS f2 UNION ALL SELECT 2 AS f1, 4 AS f2 UNION ALL SELECT 3 AS f1, 6 AS f2)
# SELECT SUM(f1) OVER (ORDER BY f2) FROM t1
```

**Official Documentation**
[Window function syntax](https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls#syntax)

VaggelisD:
Fixes #4478

https://github.com/tobymao/sqlglot/pull/4172 disallowed NULL ordering from being generated in aggregation functions; This PR extends this to also include windows of aggfuncs:

```
bq> WITH t1 AS (SELECT 1 AS f1, 2 AS f2) SELECT SUM(f1) OVER (ORDER BY f1 NULLS LAST) from t1;

NULLS LAST not supported with ascending sort order in RANGE clauses of analytic functions.
```

The previous fix was not complete as in the case of windows, the `exp.Ordered` is on the same level as `exp.AggFunc`:

```
    Window(
      this=Sum(
        this=Column(
          this=Identifier(this=f1, quoted=False))),
      order=Order(
        expressions=[
          Ordered(
            this=Column(
              this=Identifier(this=f2, quoted=False)),
            desc=False,
            nulls_first=False)]),
      over=OVER)],
```

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2399,18 +2399,21 @@ class Generator(metaclass=_Generator):
                     f"'{nulls_sort_change.strip()}' translation not supported in window functions"
                 )
                 nulls_sort_change = ""
-            elif (
-                self.NULL_ORDERING_SUPPORTED is False
-                and (isinstance(expression.find_ancestor(exp.AggFunc, exp.Select), exp.AggFunc))
-                and (
-                    (asc and nulls_sort_change == " NULLS LAST")
-                    or (desc and nulls_sort_change == " NULLS FIRST")
-                )
+            elif self.NULL_ORDERING_SUPPORTED is False and (
+                (asc and nulls_sort_change == " NULLS LAST")
+                or (desc and nulls_sort_change == " NULLS FIRST")
             ):
-                self.unsupported(
-                    f"'{nulls_sort_change.strip()}' translation not supported for aggregate functions with {sort_order} sort order"
-                )
-                nulls_sort_change = ""
+                # BigQuery does not allow these ordering/nulls combinations when used under
+                # an aggregation func or under a window containing one
+                ancestor = expression.find_ancestor(exp.AggFunc, exp.Window, exp.Select)
+
+                if ancestor and any(
+                    isinstance(expr, exp.AggFunc) for expr in (ancestor, ancestor.this)
+                ):
+                    self.unsupported(
+                        f"'{nulls_sort_change.strip()}' translation not supported for aggregate functions with {sort_order} sort order"
+                    )
+                    nulls_sort_change = ""
             elif self.NULL_ORDERING_SUPPORTED is None:
                 if expression.this.is_int:
                     self.unsupported(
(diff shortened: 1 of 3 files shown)


sqlglot/generator.py
```diff
@@ -2399,18 +2399,21 @@ def ordered_sql(self, expression: exp.Ordered) -> str:
...
-                    f"'{nulls_sort_change.strip()}' translation not supported for aggregate functions with {sort_order} sort order"
-                )
-                nulls_sort_change = ""
+                # BigQuery does not allow these ordering/nulls combinations when used under
+                # an aggregation func or under a window containing one
+                ancestor = expression.find_ancestor(exp.AggFunc, exp.Window, exp.Select)
+
+                if ancestor and any(
+                    isinstance(expr, exp.AggFunc) for expr in (ancestor, ancestor.this)
+                ):
```
georgesittas, line 2412: This is a little difficult to understand, and I think there may be an issue if `ancestor` ends up being a `Select`, because you don't check for that anymore. I'd suggest making it more explicit in favor of readability:

```python
if isinstance(ancestor, exp.Window):
    ancestor = ancestor.this
if isinstance(ancestor, exp.AggFunc):
    self.unsupported(...)
    nulls_sort_change = ""
```

georgesittas (approved):

Issue #4445, opened 2024-11-26
MySQL's numeric `format` function is not properly translated to DuckDB


fanyang01:
https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_format

```sh
$ # sqlglot-25.32.0
$ python3
Python 3.9.6 (default, Feb  3 2024, 15:58:27)
[Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlglot
>>> sqlglot.transpile("SELECT format(123456.789, 3)", read="mysql", write="duckdb")[0]
'SELECT FORMAT(123456.789, 3)'
>>>
```

It appears that this function could be translated to duckdb's numeric `format` function: https://duckdb.org/docs/sql/functions/char#print-numbers-with-thousand-separators

georgesittas:
Closing this as it's not high-priority for the core team, but we'll be happy to accept a PR if you wanna take a stab at it.

PR #4482, opened 2024-12-06
feat: add support for FORMAT function in DuckDB

ddh-5230:
fix [#4445 ](https://github.com/tobymao/sqlglot/issues/4445)
Support converting the FORMAT function in MySQL to its corresponding implementation in DuckDB.


georgesittas:
Hi, thanks for the PR; please add some unit tests for this.

georgesittas:
I don't think this approach is correct, closing for the time being. I'll take a stab at implementing it soon.

PR #4483, opened 2024-12-06
Chore: enable python 3.13 in CI

simon-pactum:


georgesittas:
Hey @simon-pactum, can you please revert the pyo3 upgrade? It looks orthogonal to including python 3.13 in the test harness, so we should discuss it in a separate PR.

georgesittas:
Thanks for the PR

PR #4486, opened 2024-12-09
feat(snowflake): Transpile support for bitor/bit_or snowflake function

ankur334:
scenarios involving binary data manipulation, such as managing permission sets or feature flags.

BITOR Function:

The BITOR function computes the bitwise OR between two numeric or binary expressions. This operation compares each bit of the inputs and returns a result where each bit is set to 1 if at least one of the corresponding bits of the operands is 1; otherwise, it is set to 0.

Example:

Suppose you have a table user_permissions that stores user IDs along with two permission sets, perm_set1 and perm_set2, represented as integers:

```
CREATE OR REPLACE TABLE user_permissions (
    user_id INTEGER,
    perm_set1 INTEGER,
    perm_set2 INTEGER
);

INSERT INTO user_permissions (user_id, perm_set1, perm_set2) VALUES
    (1, 2, 4),  -- Binary: perm_set1 = 010, perm_set2 = 100
    (2, 1, 2);  -- Binary: perm_set1 = 001, perm_set2 = 010
 ```
 
 To determine the combined permissions for each user, you can use the BITOR function:
 
 ```
 SELECT
    user_id,
    BITOR(perm_set1, perm_set2) AS combined_permissions
FROM user_permissions;
```

This query produces:
```
+---------+-------------------------+
| user_id    | combined_permissions   |
+---------+-------------------------+
|    1            |         6                                | -- Binary: 110
|    2           |         3                                 | -- Binary: 011
+---------+-------------------------+
```
At **Walmart**, we use to use `bitor`, `bit_or` a lot of places. 

I have observed that this support is not there in `sqlglot` and hence I have added the code and test case for this. 

Kindly check it once.


tests/dialects/test_snowflake.py
```diff
@@ -2277,3 +2277,13 @@ def test_window_function_arg(self):
...
+
+    def test_bitor_function(self):
+        self.validate_identity("SELECT BITOR(a, b) FROM table")
+
+    def test_bit_or_function(self):
+        # BIT_OR it just alias of BITOR, so we can transpile to BITOR only
+
+        self.validate_identity("SELECT BITOR(a, b) FROM table")
+
+        self.validate_identity("SELECT BIT_OR(a, b) FROM table", "SELECT BITOR(a, b) FROM table")
```
VaggelisD, line 2289: The 1st and 2nd `validate_identity` tests are repeated here.

Let's instead:
- Remove the 2nd call
- Move the 1st and 3rd tests under `test_snowflake`, the separate test function seems a bit overkill

sqlglot/dialects/snowflake.py
```diff
@@ -393,6 +393,8 @@ class Parser(parser.Parser):
             ),
             "BITXOR": binary_from_function(exp.BitwiseXor),
             "BIT_XOR": binary_from_function(exp.BitwiseXor),
+            "BITOR": binary_from_function(exp.BitwiseOr),
+            "BIT_OR": binary_from_function(exp.BitwiseOr),
```
VaggelisD, line 397: Do note that `BITOR` has a 3rd argument as well:

```
BITOR( <expr1> , <expr2> [ , '<padside>' ] )
```

Thus `binary_from_function` here seems problematic. The easy solution would be to fallback to an Anonymous function in that case, i.e you'd need to create a builder function that will return:
- `binary_from_function(...)` if `args == 2` or
- `exp.Anonymous(...)` if `args == 3`.

For instance, check out `snowflake.py::_build_datetime()` or `bigquery.py::_build_contains_substring()`. 

We should also test that the 3-arg version roundtrips fine in this PR too.

ankur334: Hi @VaggelisD 
I have added the suggested changes. Will you please check it in your free time?

sqlglot/dialects/snowflake.py
```diff
@@ -107,6 +107,13 @@ def _builder(args: t.List) -> E:
     return _builder
 
 
+def _build_bitor(args: t.List) -> exp.BitwiseOr | exp.Anonymous:
+    if len(args) == 3:
+        return exp.Anonymous(this="BITOR", expressions=args)
+
+    return exp.BitwiseOr(this=seq_get(args, 0), expression=seq_get(args, 1))
+
```
VaggelisD, line 115: It might look a bit cleaner if we did `binary_from_function(exp.BitwiseOr)(args)` here

tests/dialects/test_snowflake.py
```diff
@@ -976,6 +976,15 @@ def test_snowflake(self):
...
         )
+        self.validate_identity("SELECT BITOR(a, b) FROM table")
+
+        self.validate_identity("SELECT BIT_OR(a, b) FROM table", "SELECT BITOR(a, b) FROM table")
+
+        # Test BITOR with three arguments, padding on the left
+        self.validate_identity("SELECT BITOR(a, b, 'LEFT') FROM table_name")
+
+        # Test BITOR with three arguments, padding on the right
+        self.validate_identity("SELECT BITOR(a, b, 'RIGHT') FROM table_name")
```
VaggelisD, line 987: This test can be removed, the one with `'LEFT'` will suffice for now.

This is because we're essentially testing `BITOR(a, b, c)` twice here; The actual value of `padding` won't make a difference in the generation.

VaggelisD (commented):
Two more minor comments and I think this is good to go:

sqlglot/dialects/snowflake.py
ankur334: Done @VaggelisD

tests/dialects/test_snowflake.py
ankur334: Done, @VaggelisD. Thank you for your guidance and sticking with me. This is my first pull request to this repository, and I have started contributing to open source after many years. I hope to make more :)

georgesittas (approved):

sqlglot/dialects/snowflake.py
georgesittas: I believe we generally have these functions / binary operators inherit from both `Binary` and `Func`. That takes care of the argument parsing. Let's keep what's already in this PR though and we can revisit if/when needed.

georgesittas:
Thanks for the contribution @ankur334!

PR #4491, opened 2024-12-10
fix(tsql)!: Preserve roundtrips of DATETIME/DATETIME2

VaggelisD:
[Public Slack context](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1733610657252759)

This PR introduces `Type.DATETIME2` with the intention of preserving T-SQL's roundtrips of `DATETIME` and `DATETIME2` respectively.

According to the docs, `DATETIME2` is an extension to `DATETIME` supporting higher precision, date range etc. However, it's not 100% backwards compatible so transforming `"DATETIME" -> Type.DATETIME2` is not always safe.

Docs
--------
[T-SQL DATETIME](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver16) | [T-SQL DATETIME2](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=sql-server-ver16) | [S/O DATETIME VS DATETIME2](https://stackoverflow.com/questions/1334143/datetime2-vs-datetime-in-sql-server)

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -111,7 +111,7 @@ def _build_formatted_time(
         assert len(args) == 2
 
         return exp_class(
-            this=exp.cast(args[1], exp.DataType.Type.DATETIME),
+            this=exp.cast(args[1], exp.DataType.Type.DATETIME2),
             format=exp.Literal.string(
                 format_time(
                     args[0].name.lower(),
@@ -492,7 +492,7 @@ class TSQL(Dialect):
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
             "CLUSTERED INDEX": TokenType.INDEX,
-            "DATETIME2": TokenType.DATETIME,
+            "DATETIME2": TokenType.DATETIME2,
             "DATETIMEOFFSET": TokenType.TIMESTAMPTZ,
             "DECLARE": TokenType.DECLARE,
             "EXEC": TokenType.COMMAND,
@@ -507,7 +507,7 @@ class TSQL(Dialect):
             "PROC": TokenType.PROCEDURE,
             "REAL": TokenType.FLOAT,
             "ROWVERSION": TokenType.ROWVERSION,
-            "SMALLDATETIME": TokenType.DATETIME,
+            "SMALLDATETIME": TokenType.DATETIME2,
             "SMALLMONEY": TokenType.SMALLMONEY,
             "SQL_VARIANT": TokenType.VARIANT,
             "SYSTEM_USER": TokenType.CURRENT_USER,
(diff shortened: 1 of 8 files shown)


sqlglot/dialects/tsql.py
```diff
@@ -507,7 +507,7 @@ class Tokenizer(tokens.Tokenizer):
             "PROC": TokenType.PROCEDURE,
             "REAL": TokenType.FLOAT,
             "ROWVERSION": TokenType.ROWVERSION,
-            "SMALLDATETIME": TokenType.DATETIME,
+            "SMALLDATETIME": TokenType.DATETIME2,
```
georgesittas, line 510: This mapping also looks incorrect, right? `SMALLDATETIME` and `DATETIME[2]` aren't equivalent.

tests/dialects/test_tsql.py
```diff
@@ -518,13 +518,23 @@ def test_types(self):
...
+        self.validate_all(
+            "CAST(x AS DATETIME)",
+            read={
+                "": "CAST(x AS DATETIME)",
+            },
+            write={
+                "mysql": "CAST(x AS DATETIME)",
+                "tsql": "CAST(x AS DATETIME)",
+            },
+        )
```
georgesittas, line 537: Let's test duckdb here as well.

georgesittas (approved):
I think we should clean up `SMALLDATETIME` while we're at it, shouldn't be a big lift.

sqlglot/dialects/tsql.py
VaggelisD: Yep, this could also be roundtripped properly. Should we do it in this PR?

georgesittas: Up to you, but let's do it sooner rather than later if left for a followup PR, to reduce some of the tech debt.

VaggelisD: Added it in this PR

georgesittas (approved):

georgesittas:
Thanks for refactoring this @VaggelisD.

sqlglot/dialects/tsql.py
vvcb, line 1: @VaggelisD / @georgesittas - thank you very much for the fast response both on slack and here.🚀  I haven't tested this yet but it looks like we need another line there that looks like `exp.DataType.Type.DATETIME: "DATETIME"`. TSQL has multiple data types for date and datetime as described [here](https://learn.microsoft.com/en-us/sql/t-sql/functions/date-and-time-data-types-and-functions-transact-sql?view=sql-server-ver16#date-and-time-data-types).

Separately, `TIMESTAMP` in TSQL is not really a datetime data type from what I have read. It is a deprecated synonym for `ROWVERSION`. Details are lower down in page at the link above.

georgesittas: Hey Vishnu, both `DATETIME` and `ROWVERSION` are already handled correctly for T-SQL:

- The `Type.DATETIME -> "DATETIME"` entry is implicit because it maps back to the type's name (see base class)
- The `TIMESTAMP` entry you're seeing is there to facilitate transpilation of what other dialects call the "timestamp" type to T-SQL.

PR #4493, opened 2024-12-10
feat(snowflake): Support for inline FOREIGN KEY

### LINKED ISSUE #4489: Parsing Fails on Snowflake Foreign Keys Without Explicit Names
When parsing a Snowflake `CREATE TABLE` statement that creates an inline foreign key constraint with no explicit name, `sqlglot` fails with the following error:

```
sqlglot.errors.ParseError: Expecting (. Line 4, Col: 48.
```

It seems that it is expecting an explicit column name for the foreign key constraint.

From what I can tell from [Snowflake's documentation](https://docs.snowflake.com/en/sql-reference/sql/create-table-constraint#:~:text=do%20not%20need%20to%20be%20specified%20if%20the%20signature), this is valid syntax:
> When defining foreign keys, either inline or out-of-line, column name(s) for the referenced table do not need to be specified if the signature (name and data type) of the foreign key column(s) and the referenced table’s primary key column(s) exactly match

I am also able to successfully run these statements in Snowflake.

```python

from sqlglot import parse_one

sql = '''
CREATE OR REPLACE TABLE TEST (
	ID NUMBER NOT NULL IDENTITY PRIMARY KEY,
	SOME_REF NUMBER NOT NULL FOREIGN KEY REFERENCES SOME_OTHER_TABLE (ID)
);
'''
parsed = parse_one(sql, read='snowflake')

```

However, this does work:
```python
from sqlglot import parse_one

sql = '''
CREATE OR REPLACE TABLE TEST (
	ID NUMBER NOT NULL IDENTITY PRIMARY KEY,
	SOME_REF NUMBER NOT NULL FOREIGN KEY (SOME_NAME) REFERENCES SOME_OTHER_TABLE (ID)
);
'''
parsed = parse_one(sql, read='snowflake')
```

**Official Documentation**
https://docs.snowflake.com/en/sql-reference/sql/create-table-constraint#:~:text=do%20not%20need%20to%20be%20specified%20if%20the%20signature

VaggelisD:
Fixes #4489

Snowflake supports 2 definitions for `FOREIGN KEY`:

- Inline:
```
inlineFK :=
  [ CONSTRAINT <constraint_name> ]
  [ FOREIGN KEY ]
  REFERENCES <ref_table_name> [ ( <ref_col_name> ) ]
  ...
```
- Out of line:
```
outoflineFK :=
  [ CONSTRAINT <constraint_name> ]
  FOREIGN KEY ( <col_name> [ , <col_name> , ... ] )
  REFERENCES <ref_table_name> [ ( <ref_col_name> [ , <ref_col_name> , ... ] ) ]
  ...
```

This PR adds support for the inline version which now produces an empty `exp.ForeignKey` node.

Docs
---------
[Snowflake CREATE TABLE](https://docs.snowflake.com/en/sql-reference/sql/create-table-constraint)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -790,6 +790,14 @@ class Snowflake(Dialect):
 
             return this
 
+        def _parse_foreign_key(self) -> exp.ForeignKey:
+            # inlineFK, the REFERENCES column is implied
+            if self._match(TokenType.REFERENCES, advance=False):
+                return self.expression(exp.ForeignKey)
+
+            # outoflineFK, explicitly names the columns
+            return super()._parse_foreign_key()
+
     class Tokenizer(tokens.Tokenizer):
         STRING_ESCAPES = ["\\", "'"]
         HEX_STRINGS = [("x'", "'"), ("X'", "'")]
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -791,7 +791,7 @@ class Snowflake(Dialect):
             return this
 
         def _parse_foreign_key(self) -> exp.ForeignKey:
-            # inlineFK, the REFERENCES column is implied
+            # inlineFK, the REFERENCES columns are implied
             if self._match(TokenType.REFERENCES, advance=False):
                 return self.expression(exp.ForeignKey)
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2113,7 +2113,7 @@ class Directory(Expression):
 
 class ForeignKey(Expression):
     arg_types = {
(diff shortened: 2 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -790,6 +790,14 @@ def _parse_lambda_arg(self) -> t.Optional[exp.Expression]:
 
             return this
 
+        def _parse_foreign_key(self) -> exp.ForeignKey:
+            # inlineFK, the REFERENCES column is implied
```
georgesittas, line 794:
```suggestion
            # inlineFK, the REFERENCES columns are implied
```

georgesittas (approved):

PR #4495, opened 2024-12-10
Refactor!!: bundle multiple WHEN [NOT] MATCHED into a exp.WhenSequence

georgesittas:
The motivation behind this PR is to make parsing multiple `WHEN [NOT] MATCHED` clauses easier. This is related to a SQLMesh refactor I'm currently working on.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6678,7 +6678,7 @@ class Merge(DML):
         "this": True,
         "using": True,
         "on": True,
-        "when_sequence": True,
+        "whens": True,
         "with": False,
         "returning": False,
     }
@@ -6688,7 +6688,7 @@ class When(Expression):
     arg_types = {"matched": True, "source": False, "condition": False, "then": True}
 
 
-class WhenSequence(Expression):
+class Whens(Expression):
     """Wraps around one or more WHEN [NOT] MATCHED [...] clauses."""
 
     arg_types = {"expressions": True}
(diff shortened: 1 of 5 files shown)


sqlglot/expressions.py
```diff
@@ -6678,16 +6678,22 @@ class Merge(DML):
...
         "returning": False,
     }
 
 
-class When(Func):
+class When(Expression):
     arg_types = {"matched": True, "source": False, "condition": False, "then": True}
 
 
+class WhenSequence(Expression):
```
tobymao, line 6691: what about Whens

tobymao (approved):
major release?

sqlglot/expressions.py
georgesittas: sure

georgesittas:
> major release?

yep 👍

sqlglot/expressions.py
```diff
@@ -7345,14 +7351,19 @@ def merge(
     Returns:
         Merge: The syntax tree for the MERGE statement.
     """
+    expressions = []
+    for when_expr in when_exprs:
+        expressions.extend(
+            maybe_parse(
+                when_expr, dialect=dialect, copy=copy, into=WhenSequence, **opts
```
VaggelisD, line 7358: Should this be `exp.When`? We're parsing individual `WHEN`s here into `expressions` and wrapping _that_ into `exp.WhenSequence`, right?

georgesittas: The `_parse_when_matched` now corresponds to a when sequence, so this is what you want to parse. You may get a string a like `"WHEN MATCHED ... WHEN MATCHED"`, so that's gonna be a sequence, not a single `When`.

PR #4501, opened 2024-12-11
fix(snowflake): bitxor third parameter(padside) issue

ankur334:
While working on one of my pipeline scripts, I realised that `BITXOR` also support one extra parameter `padside` just like `BITOR` in SNOWFLAKE. 

BITXOR:
Returns the bitwise XOR of two numeric or binary expressions.

Aliases: BIT_XOR

Syntax:

`BITXOR( <expr1> , <expr2> [ , '<padside>' ] )`

Please find the documentation here. 
https://docs.snowflake.com/en/sql-reference/functions/bitxor


sqlglot/dialects/snowflake.py
```diff
@@ -114,6 +114,13 @@ def _build_bitor(args: t.List) -> exp.BitwiseOr | exp.Anonymous:
     return binary_from_function(exp.BitwiseOr)(args)
 
 
+def _build_bitxor(args: t.List) -> exp.BitwiseXor | exp.Anonymous:
+    if len(args) == 3:
+        return exp.Anonymous(this="BITXOR", expressions=args)
+
+    return binary_from_function(exp.BitwiseXor)(args)
+
+
```
VaggelisD, line 123: This is largely repeating the `BITOR` implementation, we could merge these two under a single builder function that takes in (1) the anonymous name and (2) the `expr_type` of the node to-be-built, e.g roughly:

```Python3
def _build_bitwise(expr_type: t.Type[E], name: str) -> t.Callable[t.List, E | exp.Anonymous]:
    def _builder(args: t.List, dialect: Dialect) -> E:
        if len(args) == 3:
            return exp.Anonymous(this=name, expressions=args)

        return binary_from_function(expr_type)(args)

    return _builder
```



Check out other builders like `dialect.py::build_regexp_extract()` for similar ideas.

sqlglot/dialects/snowflake.py
```diff
@@ -887,6 +898,8 @@ class Generator(generator.Generator):
             ),
             exp.BitwiseXor: rename_func("BITXOR"),
             exp.BitwiseOr: rename_func("BITOR"),
+            exp.BitwiseLeftShift: lambda self, e: self.func("BITSHIFTLEFT", e.this, e.expression),
+            exp.BitwiseRightShift: lambda self, e: self.func("BITSHIFTRIGHT", e.this, e.expression),
```
VaggelisD, line 902: `self.func` should be used when we want to control which args are generated; For our case here, both expressions only have `this` and `expression` so we can simplify this into:

```
            ...
            exp.BitwiseLeftShift: rename_func("BITSHIFTLEFT"),
            exp.BitwiseRightShift: rename_func("BITSHIFTRIGHT"),
            ...
```

tests/dialects/test_postgres.py
```diff
@@ -1289,3 +1289,6 @@ def test_array_length(self):
                 "clickhouse": UnsupportedError,
             },
         )
+
+    def test_timestamp_with_interval(self):
+        pass
```
VaggelisD, line 1294: This belongs to the other PR! 😄

tests/dialects/test_snowflake.py
```diff
@@ -983,6 +983,16 @@ def test_snowflake(self):
         # Test BITOR with three arguments, padding on the left
         self.validate_identity("SELECT BITOR(a, b, 'LEFT') FROM table_name")
 
+        # Test BITXOR with three arguments
+        self.validate_identity("SELECT BITXOR(a, b, 'LEFT') FROM table_name")
```
VaggelisD, line 987: Generally we can make the tests less noisy by:
- Removing the comments, they're not really required here
- Trimming down the SQL e.g we can even test the functions as standalone expressions or as simple `SELECT`s

tests/dialects/test_snowflake.py
```diff
@@ -983,6 +983,16 @@ def test_snowflake(self):
...
+
+        # Test BITSHIFTLEFT with three arguments
+        self.validate_identity("SELECT BITSHIFTLEFT(a, 1) FROM table_name")
```
VaggelisD, line 990: Wrong comment but let's just remove it for now

tests/dialects/test_snowflake.py
```diff
@@ -983,6 +983,16 @@ def test_snowflake(self):
...
+
+        # Test BITSHIFTRIGHT with three arguments
```
VaggelisD, line 992: Ditto

VaggelisD (commented):
Hey @ankur334, thank you for the contribution! Did a pass over this PR, some comments of mine:

sqlglot/dialects/snowflake.py
ankur334: @VaggelisD Made the changes. Thanks for this good suggestion. I should have merged these functions and created a reusable function in the previous commit itself. Better late than never 😄

sqlglot/dialects/snowflake.py
ankur334: I got overwhelmed by the `self.func`. I should have kept it simple. I've made the changes, please review it in your free time.

tests/dialects/test_postgres.py
ankur334: Removed it. I am working on one of the open POSTGRES issues related to parathesis and INTERVAL.

ankur334:
@VaggelisD, I have added support for `BITNOT/BIT_NOT`. It's a unary function, and I implemented it using the following code: `lambda args: exp.BitwiseNot(this=seq_get(args, 0))`. Should I keep this logic within the `_build_bitwise` function itself?

However, due to the presence of `BINARY`, `Unary`, and `Anomyosos`, this function was becoming overly complex.

VaggelisD:
Hey @ankur334, let's ship features/fixes on separate PRs if possible, otherwise the scope creeps up and it's harder to review.

 I'd suggest we revert the `BITNOT` addition for the time being so we can do a final pass on the other changes here

sqlglot/dialects/snowflake.py
VaggelisD: No worries, these are patterns that usually occur only after we have accumulated repetitive sections. Your initial solution was fine too 👍

tests/dialects/test_snowflake.py
```diff
@@ -976,12 +976,17 @@ def test_snowflake(self):
...
 
-        # Test BITOR with three arguments, padding on the left
-        self.validate_identity("SELECT BITOR(a, b, 'LEFT') FROM table_name")
+        self.validate_identity("SELECT BITOR(a, b, 'LEFT')")
+
+        self.validate_identity("SELECT BITXOR(a, b, 'LEFT')")
+
+        self.validate_identity("SELECT BITSHIFTLEFT(a, 1)")
+
+        self.validate_identity("SELECT BITSHIFTRIGHT(a, 1)")
```
VaggelisD, line 989: Nit: let's remove the whitespaces between each `validate_identity` call to reduce file size

tests/dialects/test_snowflake.py
```diff
@@ -976,12 +976,17 @@ def test_snowflake(self):
...
+        self.validate_identity("SELECT BITOR(a, b)")
 
-        self.validate_identity("SELECT BIT_OR(a, b) FROM table", "SELECT BITOR(a, b) FROM table")
+        self.validate_identity("SELECT BIT_OR(a, b)", "SELECT BITOR(a, b)")
 
-        # Test BITOR with three arguments, padding on the left
-        self.validate_identity("SELECT BITOR(a, b, 'LEFT') FROM table_name")
+        self.validate_identity("SELECT BITOR(a, b, 'LEFT')")
+
+        self.validate_identity("SELECT BITXOR(a, b, 'LEFT')")
```
VaggelisD, line 985: Let's also test `BIT_XOR -> BITXOR` roundtrip here, similar to `BITOR`

VaggelisD (commented):
Two more comments and I think the BITWISEXOR & shifts are good to go:

VaggelisD (approved):
Looks good, thank you!

Issue #4502, opened 2024-12-11
[Snowflake] `sqlglot.optimizer.qualify.qualify` does not resolve SELECT columns inside JOINs


florian-ernst-alan:
Snowflake has a feature that allows using aliases anywhere in a query, including inside JOIN statements:
```sql
-- Here, we use the aliased "birth_year" in the JOIN clause!
SELECT u.id
  , u.birth_date
  , YEAR(u.birth_date) AS birth_year
  , ly.is_leap AS birth_year_is_leap
FROM users u
JOIN leap_years ly ON birth_year = ly.year

```

This is a [well-known feature](https://sqlkover.com/cool-stuff-in-snowflake-part-4-aliasing-all-the-things/).

However, `sqlglot.optimizer.qualify.qualify` fails to resolve the `birth_year` column in the `ON` clause, causing `sqlglot.optimizer.optimize` to fail entirely.

**Fully reproducible code snippet**
```py
from sqlglot import parse_one
from sqlglot.optimizer.qualify import qualify

qualify(parse_one(sql = """
SELECT u.id
     , u.birth_date
     , YEAR(u.birth_date) AS birth_year
     , ly.is_leap AS birth_year_is_leap
FROM users u
JOIN leap_years ly ON birth_year = ly.year
""", dialect="snowflake"), schema=None)

# > sqlglot.errors.OptimizeError: Column '"birth_year"' could not be resolved
```

**Official Documentation**

I didn't find any official documentation, but it's a well-known feature. My Account Manager confirmed it's supported, I asked him for official links - he didn't come back to me yet.

tobymao:
@florian-ernst-alan i don't know if one blog post means it's a "well-known feature". a well-known feature generally has readily available docs as well. no one else has commented on this particular use case and it's pretty non-standard. so this seems more niche than "well-known".

happy to accept a PR for this but we won't be prioritizing this for now

florian-ernst-alan:
@tobymao While I understand challenging the "well-known" part of my issue, it's still a supported feature.

I found the documentation [here](https://docs.snowflake.com/en/sql-reference/sql/select#usage-notes):
> In many cases, when you use a column alias for an expression (i.e. expr AS col_alias) in other parts of the same query (in JOIN, FROM, WHERE, GROUP BY, other column expressions, etc.), the expression is evaluated only once.

> However, note that in some cases, the expression can be evaluated multiple times, which can result in different values for the alias used in different parts of the same query.

[...]

PR #4504, opened 2024-12-11
fix(snowflake): Allow alias expansion inside JOIN statements

florian-ernst-alan:
Fixes https://github.com/tobymao/sqlglot/issues/4502


florian-ernst-alan:
Dammit, missed the formatting checks locally. I'll fix tomorrow, it's 8PM here.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -66,14 +66,15 @@ def qualify_columns(
             _expand_alias_refs(
                 scope,
                 resolver,
+                dialect,
                 expand_only_groupby=dialect.EXPAND_ALIAS_REFS_EARLY_ONLY_IN_GROUP_BY,
```
georgesittas, line 70: Passing `dialect` here means we can refactor the `expand_only_groupby` flag as well.

tests/dialects/test_snowflake.py
```diff
@@ -2289,3 +2289,45 @@ def test_window_function_arg(self):
         self.assertEqual(ast.sql("snowflake"), query)
         self.assertEqual(len(list(ast.find_all(exp.Column))), 1)
         self.assertEqual(window.this.sql("snowflake"), "db.schema.FUNC(a)")
+
+    def test_alias_expansion(self):
```
georgesittas, line 2293: Let's not roll a custom test method for this, we have a fixture file called `qualify_columns.sql` that contains similar tests.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -309,6 +316,11 @@ def replace_columns(
     replace_columns(expression.args.get("having"), resolve_table=True)
     replace_columns(expression.args.get("qualify"), resolve_table=True)
 
+    # Snowflake allows alias expansion in the JOIN ... ON clause (and almost everywhere else)
+    # https://docs.snowflake.com/en/sql-reference/sql/select#usage-notes
+    for join in expression.args.get("joins", []):
+        replace_columns(join)
```
georgesittas, line 322: You forgot to check for snowflake here.

sqlglot/optimizer/qualify_columns.py
florian-ernst-alan: Doing that actually make some BigQuery tests fail, because this call line 77
```py
        if not schema.empty and expand_alias_refs:
            _expand_alias_refs(scope, resolver, dialect)
```

alwasy has `expand_only_groupby=False` regardless of the dialect.

georgesittas: Ah, makes sense. Should we refactor this PR then to introduce a similar constant and an arg like `expand_joins`?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -309,6 +318,12 @@ def replace_columns(
     replace_columns(expression.args.get("having"), resolve_table=True)
     replace_columns(expression.args.get("qualify"), resolve_table=True)
 
+    # Snowflake allows alias expansion in the JOIN ... ON clause (and almost everywhere else)
+    # https://docs.snowflake.com/en/sql-reference/sql/select#usage-notes
+    if dialect == "snowflake":
+        for join in expression.args.get("joins", []):
```
georgesittas, line 324:
```suggestion
        for join in expression.args.get("joins") or []:
```

georgesittas (approved):
Probably fine as is.

PR #4509, opened 2024-12-12
feat(starrocks): add partition by range and unique key

pickfire:


sqlglot/dialects/starrocks.py
```diff
@@ -14,11 +14,18 @@
...
 
 class StarRocks(MySQL):
     STRICT_JSON_PATH_SYNTAX = False
 
+    class Tokenizer(MySQL.Tokenizer):
+        KEYWORDS = {
+            **MySQL.Tokenizer.KEYWORDS,
+            "PARTITION BY RANGE": TokenType.PARTITION_BY_RANGE,
+        }
+
```
georgesittas, line 28: Let's not do it like this, remove the token and try to capture this property using `_match_text_seq`. Lately we've been avoiding polluting the token enum, mostly adding new tokens for types etc.

This also has the problem of assuming only 1 space between the keywords, which isn't necessarily true, e.g. `PARTITION     BY` is also valid right? This problem exists in some other parts of the codebase today, but we're probably going to refactor those as well at some point.

georgesittas (commented):
Thanks for the PR @pickfire, please share documentation about the added features.

sqlglot/dialects/starrocks.py
pickfire: Where do I change this part? Currently this is for keywords. Do you have any code reference?

georgesittas: I'd look in the base parser's `PROPERTY_PARSERS` dict and probably extend/override the parser corresponding to `PARTITION BY`.

pickfire: Fixed

sqlglot/parser.py
```diff
@@ -2205,9 +2205,14 @@ def _parse_distributed_property(self) -> exp.DistributedByProperty:
...
     def _parse_duplicate(self) -> exp.DuplicateKeyProperty:
         self._match_text_seq("KEY")
-        expressions = self._parse_wrapped_csv(self._parse_id_var, optional=False)
+        expressions = self._parse_wrapped_id_vars()
         return self.expression(exp.DuplicateKeyProperty, expressions=expressions)
 
+    def _parse_unique_property(self) -> exp.UniqueKeyProperty:
+        self._match_text_seq("KEY")
+        expressions = self._parse_wrapped_id_vars()
+        return self.expression(exp.UniqueKeyProperty, expressions=expressions)
```
georgesittas, line 2214: It could be worth consolidating this and `_parse_duplicate` into a single definition that gets a `Type[Expression]` and instantiates it according to this (common) logic.

sqlglot/dialects/starrocks.py
```diff
@@ -70,6 +79,35 @@ def _parse_unnest(self, with_alias: bool = True) -> t.Optional[exp.Unnest]:
...
+                partition_expressions = self._parse_wrapped_id_vars()
+                create_expressions = self._parse_wrapped_csv(
+                    self._parse_partitioning_granularity_dynamic
+                )
+                return self.expression(
+                    exp.PartitionByRangeProperty,
+                    partition_expressions=partition_expressions,
+                    create_expressions=create_expressions,
+                )
+            return super()._parse_partitioned_by()
```
georgesittas, line 109: Just double-checking: there's also this in starrocks' docs:

```
PARTITION <partition1_name> VALUES LESS THAN ("<upper_bound_for_partitioning_column1>" [ , "<upper_bound_for_partitioning_column2>", ... ] )
```

From what I understand, your changes only add support for the `(START...END...EVERY...)` syntax.

1. What happens today if the user tries to parse the alternative syntax?
2. Same question as (1), but given the changes in your PR?

I wanna ensure there are no regressions.

sqlglot/dialects/starrocks.py
```diff
@@ -287,3 +328,16 @@ def create_sql(self, expression: exp.Create) -> str:
...
+            create = self.expressions(expression, key="create_expressions")
+            return f"PARTITION BY RANGE {self.wrap(partitions)} {self.wrap(create)}"
+
+        def partitionbyrangepropertydynamic_sql(
+            self, expression: exp.PartitionByRangePropertyDynamic
+        ) -> str:
+            start = self.sql(expression, key="start")
+            end = self.sql(expression, key="end")
+            every = self.sql(expression, key="every")
+            return f"START {self.wrap(start)} END {self.wrap(end)} EVERY {self.wrap(every)}"
```
georgesittas, line 343: Let's move these to the base generator, so that generating without a dialect doesn't throw an exception.

sqlglot/parser.py
```diff
@@ -2205,9 +2205,14 @@ def _parse_distributed_property(self) -> exp.DistributedByProperty:
 
     def _parse_duplicate(self) -> exp.DuplicateKeyProperty:
         self._match_text_seq("KEY")
-        expressions = self._parse_wrapped_csv(self._parse_id_var, optional=False)
+        expressions = self._parse_wrapped_id_vars()
```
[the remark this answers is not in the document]
pickfire: Because it's the same thing? Should I revert it?

sqlglot/dialects/starrocks.py
```diff
@@ -14,11 +14,18 @@
...
+from sqlglot.tokens import TokenType
 
 
 class StarRocks(MySQL):
     STRICT_JSON_PATH_SYNTAX = False
 
+    class Tokenizer(MySQL.Tokenizer):
+        KEYWORDS = {
+            **MySQL.Tokenizer.KEYWORDS,
+            "LARGEINT": TokenType.INT128,
```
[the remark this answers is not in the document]
pickfire: This is tested in test_starrocks.py

sqlglot/parser.py
pickfire: I don't get what do you want me to do here.

sqlglot/dialects/starrocks.py
pickfire: I did not support this as I don't have the need to use it, other people that need it can add it later, later they can just add another expression for this, since it is totally different.

Yup, it only supports the dynamic partitioning range method `(START...END...EVERY...)` syntax like you mentioned.

sqlglot/dialects/starrocks.py
pickfire: Fixed

sqlglot/dialects/starrocks.py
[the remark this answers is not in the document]
georgesittas: Ah my bad, apologies. I think I searched for `INT128` instead and missed it. All good 👍

sqlglot/parser.py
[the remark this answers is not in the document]
georgesittas: No, I saw `optional` and assumed it was set to true instead of false. Good catch.

sqlglot/parser.py
georgesittas: Something like:

```python
def _parse_composite_key_property(self, expr_type: t.Type[E]) -> E:
    self._match_text_seq("KEY")
    expressions = self._parse_wrapped_id_vars()
    return self.expression(expr_type, expressions=expressions
```

So we can use it for both `_parse_unique_property` and `_parse_duplicate` in the base parser.

sqlglot/dialects/starrocks.py
georgesittas: This makes sense, but there's some nuance here. Some unsupported DDLs are still "parsed" today, meaning that we produce a `Command` node and store the input as raw text in it. The motivation behind this is to avoid crashing in cases where transpilation / parsing is not required for certain statements.

Since you introduced new logic for this property, I wanted to ensure that we don't break these `Command` statements accidentally by entering this `if self._match_text_seq("RANGE")` branch.

Basically let's test what happened before vs after. If the `VALUES LESS THAN` clause was not supported before, it's fine. If it was parsed into a `Command`, we need to make sure that if the parser fails, we'll fallback to a command as well.

Does this make sense?

pickfire: I think I will ignore this for now, don't want to spend too much time on this.

sqlglot/parser.py
pickfire: Fixed.

sqlglot/parser.py
```diff
@@ -4584,7 +4589,9 @@ def _parse_escape(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expr
             return this
         return self.expression(exp.Escape, this=this, expression=self._parse_string())
 
-    def _parse_interval(self, match_interval: bool = True) -> t.Optional[exp.Add | exp.Interval]:
+    def _parse_interval(
+        self, match_interval: bool = True, keep_number: bool = False
+    ) -> t.Optional[exp.Add | exp.Interval]:
```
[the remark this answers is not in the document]
pickfire: `INTERVAL '1' DAY` gets an error in starrocks, that's why I did it this way, so what you mean is during output we tweak it to number?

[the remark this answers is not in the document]
georgesittas: Yep

sqlglot/parser.py
georgesittas: Thanks

georgesittas:
Thanks for the PR.

PR #4532, opened 2024-12-18
feat: Introduce meta comment to parse known functions as exp.Anonymous

### LINKED ISSUE #4522: BigQuery parser mistakes `YEAR` UDFs for a standard function in `Standard SQL`
I wrote few SQL UDFs that are named YEAR/MONTH/DAY/HOUR/MINUTE in BigQuery and the parser failed as the number of arguments didn't fit the expected ones.

I checked those functions don't in BigQuery standard SQL. Then I had to rename my functions to ensure it doesn't collide with existing parser.
What's tricky is that [the functions **exist in BigQuery Legacy SQL**](https://cloud.google.com/bigquery/docs/reference/legacy-sql#hour) and I think it's why it was implemented like this.
However the code is still valid for Standard SQL so I guess you would have to accept any parameter for those functions to avoid parser error even if it makes it less precise for Legacy SQL (which is meant to disappear).

Here's my workaround change:
https://github.com/bqbooster/dbt-bigquery-monitoring/pull/89/files#diff-27c7e17690300b0dfe5ac5ff29acdb74cf6a7fa2a98aa8b48ed63104997fa9a8

VaggelisD:
Fixes #4522

This PR introduces support for the meta comment `/* sqlglot.meta anonymous */` which forces the parser to ignore the known functions during `_parse_function_call`:

- Without meta comment
```Python3
>>> print(repr(parse_one("YEAR(a)")))
Year(
  this=Column(
    this=Identifier(this=a, quoted=False)))
```

- With meta comment
```Python3
>>> print(repr(parse_one("YEAR(a) /* sqlglot.meta anonymous */")))
Anonymous(
  this=YEAR,
  expressions=[
    Column(
      this=Identifier(this=a, quoted=False))],
  _comments=[
    sqlglot.meta anonymous ])

``` 

- With incorrect / incomplete meta comment
```Python3
>>> print(repr(parse_one("YEAR(a) /* sqlglot.meta anon */")))
Year(
  this=Column(
  _comments=[
    sqlglot.meta anon ])
```

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -5366,8 +5366,7 @@ class Parser(metaclass=_Parser):
                 # If the meta comment "/* sqlglot.meta anonymous */" is following the function
                 # call we'll construct it as exp.Anonymous, even if it's "known"
                 for comment in post_func_comments:
-                    lower = comment.lower()
-                    if exp.SQLGLOT_META in lower and "anonymous" in lower:
+                    if exp.SQLGLOT_META in comment and "anonymous" in comment:
                         known_function = False
                         break
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -5361,6 +5361,15 @@ class Parser(metaclass=_Parser):
             alias = not known_function or upper in self.FUNCTIONS_WITH_ALIASED_ARGS
             args = self._parse_csv(lambda: self._parse_lambda(alias=alias))
 
+            post_func_comments = self._curr and self._curr.comments
+            if known_function and post_func_comments:
+                anonymous_meta = (exp.SQLGLOT_META, "anonymous")
+                # If the meta comment "/* sqlglot.meta anonymous */" is following the function
+                # call we'll construct it as exp.Anonymous, even if it's "known"
+                for comment in post_func_comments:
+                    if all(meta in comment.lower() for meta in anonymous_meta):
+                        known_function = False
+
             if alias and known_function:
                 args = self._kv_to_prop_eq(args)
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -5363,12 +5363,13 @@ class Parser(metaclass=_Parser):
 
             post_func_comments = self._curr and self._curr.comments
             if known_function and post_func_comments:
-                # If the meta comment "/* sqlglot.meta anonymous */" is following the function
+                # If the user-inputted comment "/* sqlglot.anonymous */" is following the function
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -5361,6 +5361,15 @@ def _parse_function_call(
...
             args = self._parse_csv(lambda: self._parse_lambda(alias=alias))
 
+            post_func_comments = self._curr and self._curr.comments
+            if known_function and post_func_comments:
+                anonymous_meta = (exp.SQLGLOT_META, "anonymous")
+                # If the meta comment "/* sqlglot.meta anonymous */" is following the function
+                # call we'll construct it as exp.Anonymous, even if it's "known"
+                for comment in post_func_comments:
+                    if all(meta in comment.lower() for meta in anonymous_meta):
+                        known_function = False
```
georgesittas, line 5371: You wanna do `lower()` once, and also you can break after setting `known_function` in the branch, right?

VaggelisD: Yep sg

georgesittas (approved):

sqlglot/parser.py
```diff
@@ -5361,6 +5361,16 @@ def _parse_function_call(
             alias = not known_function or upper in self.FUNCTIONS_WITH_ALIASED_ARGS
             args = self._parse_csv(lambda: self._parse_lambda(alias=alias))
 
+            post_func_comments = self._curr and self._curr.comments
+            if known_function and post_func_comments:
+                # If the meta comment "/* sqlglot.meta anonymous */" is following the function
+                # call we'll construct it as exp.Anonymous, even if it's "known"
+                for comment in post_func_comments:
+                    lower = comment.lower()
+                    if exp.SQLGLOT_META in lower and "anonymous" in lower:
```
tobymao, line 5370: it can be case sensitive.

VaggelisD: Re: case sensitivity, you mean reverting the `lower()` call? I went with it for better UX, fine either way.

To leverage `add_comments` we'd have to create a dummy node for it and attach the comments. Though, they could be valid comments for all we know so we might have to undo that. In the end, I think we still need this lookahead/peek afaict.

VaggelisD: The comment comes after the function so it'd be added at the end of `_parse_function_call`, i.e after validation has ran

georgesittas (approved):

georgesittas (approved):

tobymao (approved):

PR #4536, opened 2024-12-18
feat(tokenizer): Allow underscore separated number literals

### LINKED ISSUE #4530: Incorrectly parses number literals with `_`s (ClickHouse dialect)
ClickHouse [supports](https://clickhouse.com/docs/en/sql-reference/syntax#numeric) using `_` in numbers for better human-readability of big numbers.
`sqlglot` does not correctly parse them:

```
>>> sqlglot.parse_one("select 1_000", dialect="clickhouse")
Select(
  expressions=[
    Alias(
      this=Literal(this=1, is_string=False),
      alias=Identifier(this=_000, quoted=False))])
```

I would expect output to look like:
```
>>> sqlglot.parse_one("select 1_000", dialect="clickhouse")
Select(
  expressions=[
    Literal(this=1000, is_string=False)
])
```

Thanks for your awesome library!

VaggelisD:
Fixes #4530

This PR introduces the dialect flag `NUMBERS_CAN_BE_UNDERSCORE_SEPARATED` in order to support underscore separated numbers. With the new flag, the tokenizer(s) ignore the underscores and preserve only the actual number as a form of canonicalization.

Do note that Clickhouse supports both identifiers starting with digits _and_ underscore separated numbers:

```SQL
:) with t as (select 99 1_b) select 1_b, 1_1 from t;

   ┌─1_b─┬─11─┐
1. │  99 │ 11 │
   └─────┴────┘
```

Other dialects such as Spark will treat that as an identifier:

```
spark-sql (default)> with t as (select 2 as 1_2) select 1_2 from t;
1_2
2
```

Docs
--------
[CH Numeric](https://clickhouse.com/docs/en/sql-reference/syntax#numeric)

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -997,6 +997,7 @@ class Tokenizer(metaclass=_Tokenizer):
             self._rs_dialect_settings = RsTokenizerDialectSettings(
                 unescaped_sequences=self.dialect.UNESCAPED_SEQUENCES,
                 identifiers_can_start_with_digit=self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT,
+                numbers_can_be_underscore_separated=self.dialect.NUMBERS_CAN_BE_UNDERSCORE_SEPARATED,
             )
 
         self.reset()
@@ -1300,8 +1301,12 @@ class Tokenizer(metaclass=_Tokenizer):
                     self._add(TokenType.NUMBER, number_text)
                     self._add(TokenType.DCOLON, "::")
                     return self._add(token_type, literal)
-                elif self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT:
-                    return self._add(TokenType.VAR)
+                else:
+                    replaced = literal.replace("_", "")
+                    if self.dialect.NUMBERS_CAN_BE_UNDERSCORE_SEPARATED and replaced.isdigit():
+                        return self._add(TokenType.NUMBER, number_text + replaced)
+                    elif self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT:
+                        return self._add(TokenType.VAR)
 
                 self._advance(-len(literal))
                 return self._add(TokenType.NUMBER, number_text)
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -166,6 +166,8 @@ class ClickHouse(Dialect):
     LOG_BASE_FIRST: t.Optional[bool] = None
     FORCE_EARLY_ALIAS_REF_EXPANSION = True
     PRESERVE_ORIGINAL_NAMES = True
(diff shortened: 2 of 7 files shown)


VaggelisD:
cc: @benfdking for the Rust changes, couldn't get you on the reviewer list

sqlglot/tokens.py
```diff
@@ -1300,8 +1301,12 @@ def _scan_number(self) -> None:
                     self._add(TokenType.NUMBER, number_text)
                     self._add(TokenType.DCOLON, "::")
                     return self._add(token_type, literal)
-                elif self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT:
-                    return self._add(TokenType.VAR)
+                else:
+                    replaced = literal.replace("_", "")
+                    if self.dialect.NUMBERS_CAN_BE_UNDERSCORE_SEPARATED and replaced.isdigit():
+                        return self._add(TokenType.NUMBER, number_text + replaced)
+                    elif self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT:
```
georgesittas, line 1308:
```suggestion
                    if self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT:
```

sqlglot/tokens.py
```diff
@@ -1300,8 +1301,12 @@ def _scan_number(self) -> None:
                     self._add(TokenType.NUMBER, number_text)
                     self._add(TokenType.DCOLON, "::")
                     return self._add(token_type, literal)
-                elif self.dialect.IDENTIFIERS_CAN_START_WITH_DIGIT:
-                    return self._add(TokenType.VAR)
+                else:
+                    replaced = literal.replace("_", "")
+                    if self.dialect.NUMBERS_CAN_BE_UNDERSCORE_SEPARATED and replaced.isdigit():
+                        return self._add(TokenType.NUMBER, number_text + replaced)
```
georgesittas, line 1307: What's `number_text + replaced` here? Don't we only need `replaced`?

VaggelisD: `number_text` would be the first substring, e.g in `123_456_789` we'd have:

- `number_text`: `123`
- `literal`: `_456_789`
- `replaced`: `456789`

sqlglot/tokens.py
VaggelisD: Hm these should be mut-ex, no (?)

georgesittas: If you enter the first branch you return, they're still mutually-exclusive. This style is just more consistent with what we do in the codebase in general.

sqlglot/tokens.py
georgesittas: Gotcha, thanks for clarifying.

georgesittas (approved):
Just one comment on my end, looks good.

sqlglot/tokens.py
VaggelisD: Right, deja vu

Issue #4531, opened 2024-12-18
DATETRUNC not supported for dialect="tsql"


rajat-wisdom:
**Description**
The issue exists in the latest version 26.0.0. DATETRUNC is an allowed function in T-SQL. Please see link below. However, it is not supported in sqlglot. Please see snippet below.

**Fully reproducible code snippet**
```python
parse_one("DATETRUNC(month, closed_date)", dialect="tsql")
Anonymous(
  this=DATETRUNC,
  expressions=[
    Column(
      this=Identifier(this=month, quoted=False)),
    Column(
      this=Identifier(this=closed_date, quoted=False))])`
``` 
**Expected output**
A parse tree similar to what we get for other dialects like redshift.
```python
parse_one("DATE_TRUNC(month, closed_date)", dialect="redshift")
TimestampTrunc(
  this=Column(
    this=Identifier(this=closed_date, quoted=False)),
  unit=Var(this=MONTH))
``` 
**Official Documentation**
[DATETRUNC in T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/functions/datetrunc-transact-sql?view=sql-server-ver16)

geooo109:
Thank you @rajat-wisdom for pointing it out.
I am working on it.

PR #4537, opened 2024-12-19
feat(tsql): add support for DATETRUNC #4531

geooo109:
Fixes [https://github.com/tobymao/sqlglot/issues/4531](https://github.com/tobymao/sqlglot/issues/4531)

This PR adds support for the `DATETRUNC` function of `T-SQL`.

In cases, where the `date` parameter is a `string literal`, a casting is applied to produce a `DATETIME2` expression.
Based on `T-SQL`, all `string literals` should resolve into `DATETIME2`.
Dialects like `duckdb` do not support the direct `string literals` as dates, thus the casting is mandatory.  

**Docs**
[T-SQL DATETRUNC](https://learn.microsoft.com/en-us/sql/t-sql/functions/datetrunc-transact-sql?view=sql-server-ver16)

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -370,6 +370,17 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
     return sql
 
 
+def _parse_datetrunc(args: t.List) -> exp.TimestampTrunc:
+    unit = seq_get(args, 0)
+    this = seq_get(args, 1)
+
+    # Every date - time type in tsql resoves to DATETIME2
+    if isinstance(this, exp.Expression) and this.is_string:
+        this = exp.TimeStrToTime(this=args[1])
+
+    return exp.TimestampTrunc(unit=unit, this=this)
+
+
 class TSQL(Dialect):
     SUPPORTS_SEMI_ANTI_JOIN = False
     LOG_BASE_FIRST = False
@@ -570,6 +581,7 @@ class TSQL(Dialect):
             "SUSER_SNAME": exp.CurrentUser.from_arg_list,
             "SYSTEM_USER": exp.CurrentUser.from_arg_list,
             "TIMEFROMPARTS": _build_timefromparts,
+            "DATETRUNC": _parse_datetrunc,
         }
 
         JOIN_HINTS = {"LOOP", "HASH", "MERGE", "REMOTE"}
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -2090,3 +2090,19 @@ FROM OPENJSON(@json) WITH (
                 "oracle": "SELECT NEXT VALUE FOR db.schema.sequence_name",
             },
         )
+
+    def test_datetrunc(self):
+        self.validate_all(
+            "SELECT DATETRUNC(month, '2021-12-08 11:30:15.1234567')",
+            write={
+                "duckdb": "SELECT DATE_TRUNC('MONTH', CAST('2021-12-08 11:30:15.1234567' AS TIMESTAMP))"
+            },
+        )
+        self.validate_all(
+            "SELECT DATETRUNC(year, DATEFROMPARTS(2010, 12, 31))",
+            write={"duckdb": "SELECT DATE_TRUNC('YEAR', MAKE_DATE(2010, 12, 31))"},
+        )
+        self.validate_all(
+            "SELECT DATETRUNC(year, CAST('2021-12-08' AS date))",
+            write={"duckdb": "SELECT DATE_TRUNC('YEAR', CAST('2021-12-08' AS DATE))"},
+        )


sqlglot/dialects/tsql.py
```diff
@@ -370,6 +370,17 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
     return sql
 
 
+def _parse_datetrunc(args: t.List) -> exp.TimestampTrunc:
+    unit = seq_get(args, 0)
+    this = seq_get(args, 1)
+
+    # Every date - time type in tsql resoves to DATETIME2
+    if isinstance(this, exp.Expression) and this.is_string:
```
VaggelisD, line 377: If you switch to `this = args[1]` we can simplify this branch to just:

 ```Python3
if this.is_string:
   ...
```

If we leave `seq_get` in then I'd change this to:

```Python3
if this and this.is_string:
   ...
```

sqlglot/dialects/tsql.py
```diff
@@ -370,6 +370,17 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
...
+        this = exp.TimeStrToTime(this=this)
```
VaggelisD, line 378: Not sure if this should be `exp.TimeStrToTime` here, I think we could go with `exp.cast(this, exp.DataType.Type.DATETIME2)`

sqlglot/dialects/tsql.py
```diff
@@ -370,6 +370,17 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
...
+
+    return exp.TimestampTrunc(unit=unit, this=this)
```
VaggelisD, line 380: Minor nit, let's go with `exp.TimestampTrunc(this=this, unit=unit)`, `this` is generally the first arg in the constructors

sqlglot/dialects/tsql.py
```diff
@@ -570,6 +581,7 @@ class Parser(parser.Parser):
             "SUSER_SNAME": exp.CurrentUser.from_arg_list,
             "SYSTEM_USER": exp.CurrentUser.from_arg_list,
             "TIMEFROMPARTS": _build_timefromparts,
+            "DATETRUNC": _parse_datetrunc,
```
VaggelisD, line 583: Let's rename that to `_build_datetrunc` for consistency with other builders

tests/dialects/test_tsql.py
```diff
@@ -2090,3 +2090,19 @@ def test_next_value_for(self):
                 "oracle": "SELECT NEXT VALUE FOR db.schema.sequence_name",
             },
         )
+
+    def test_datetrunc(self):
```
VaggelisD, line 2094: None of these tests check the roundtrip; We must include a `"tsql"` entry in either read/write dicts

tests/dialects/test_tsql.py
```diff
@@ -2090,3 +2090,19 @@ def test_next_value_for(self):
...
+        self.validate_all(
+            "SELECT DATETRUNC(month, '2021-12-08 11:30:15.1234567')",
+            write={
+                "duckdb": "SELECT DATE_TRUNC('MONTH', CAST('2021-12-08 11:30:15.1234567' AS TIMESTAMP))"
+            },
+        )
+        self.validate_all(
+            "SELECT DATETRUNC(year, DATEFROMPARTS(2010, 12, 31))",
+            write={"duckdb": "SELECT DATE_TRUNC('YEAR', MAKE_DATE(2010, 12, 31))"},
+        )
```
VaggelisD, line 2104: Try not to inline the `write` dict, prefer to go with the "expanded" version similar to the first case i.e:

```Python3
        self.validate_all(
            "SELECT DATETRUNC(year, DATEFROMPARTS(2010, 12, 31))",
            write={
                "duckdb": "SELECT DATE_TRUNC('YEAR', MAKE_DATE(2010, 12, 31))",
            },
        }

``` 

Also, the tests don't have to be "real", imo it'd be cleaner to have something like `"SELECT DATETRUNC(year, foo)"` to showcase the non-literal case

tests/dialects/test_tsql.py
```diff
@@ -2090,3 +2090,19 @@ def test_next_value_for(self):
...
+        self.validate_all(
+            "SELECT DATETRUNC(year, CAST('2021-12-08' AS date))",
+            write={"duckdb": "SELECT DATE_TRUNC('YEAR', CAST('2021-12-08' AS DATE))"},
+        )
```
VaggelisD, line 2108: Ditto about the inlining

sqlglot/dialects/tsql.py
```diff
@@ -370,6 +370,17 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
     return sql
 
 
+def _parse_datetrunc(args: t.List) -> exp.TimestampTrunc:
+    unit = seq_get(args, 0)
+    this = seq_get(args, 1)
```
VaggelisD, line 375: Since both of them are required in T-SQL we could index them directly here, `seq_get` can get annoying with mypy

VaggelisD (commented):
Nice work! Left a few comments from my first pass:

sqlglot/dialects/tsql.py
georgesittas: Let's not asume `args[1]` will work, e.g. if you set the error level to ignore that'd fail with an index error. I'd go with the latter.

sqlglot/dialects/tsql.py
geooo109: Ok, I thought `exp.TimeStrToTime` was a better approach, but I will change it to the explicit cast.

sqlglot/dialects/tsql.py
VaggelisD: Ah yeah that's right, even if an expression requires some args (in the case of `exp.TimestampTrunc`, both `this` and `unit` are required anyways), setting `error_level=sqlglot.ErrorLevel.IGNORE` trumps that.

sqlglot/dialects/tsql.py
VaggelisD: Scratch that, my bad

sqlglot/dialects/tsql.py
VaggelisD: You'd have to see if `exp.TimeStrToTime` behaves as expected in T-SQL, i.e does it cast to `DATETIME2`, does it have other side-effects etc. 

Otherwise, always try to go with the straightforward solution

georgesittas (approved):
Nice!

tests/dialects/test_tsql.py
```diff
@@ -2090,3 +2090,27 @@ def test_next_value_for(self):
                 "oracle": "SELECT NEXT VALUE FOR db.schema.sequence_name",
             },
         )
+
+    # string literals in the DATETRUNC are casted as DATETIME2
```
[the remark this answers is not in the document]
georgesittas: I don't think that'd be a valuable comment @VaggelisD, the builder code is easy to follow.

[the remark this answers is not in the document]
VaggelisD: It might look like an incorrect assumption considering this is `DATE_TRUNC` and the string is being casted to `DATETIME2`, but yeah no strong preference on it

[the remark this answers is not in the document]
georgesittas: if you wanna leave a comment in date trunc's builder, then put a reference to the doc page [where it says](https://learn.microsoft.com/en-us/sql/t-sql/functions/datetrunc-transact-sql?view=sql-server-ver16#date) that if `date` is a string, it needs to be resolvable to a datetime2:

> DATETRUNC also accepts a string literal (of any string type) that can resolve to a datetime2(7).

PR #4552, opened 2024-12-30
chore: introduce benchmarking for rust

benfdking:
- introduces a criterion based bench based on the long sql
- introduces serde serializing and deserializing that is not built into the main release
- introduces some json files which are just simple serialization of a set of configs taken from sqlglot
- introduces gha action that compares performance in a pr to main to check for regressions


sqlglotrs/src/settings.rs
```diff
@@ -162,12 +183,30 @@ impl TokenizerSettings {
...
+
+        #[cfg(feature = "profiling")]
+        {
+            tokenizer_settings.write_json_to_string();
         }
+
+        tokenizer_settings
+    }
+}
+#[cfg(feature = "profiling")]
```
izeigerman, line 196: Missing empty line?

sqlglotrs/src/settings.rs
```diff
@@ -162,12 +183,30 @@ impl TokenizerSettings {
...
+impl TokenizerSettings {
+    pub fn write_json_to_string(&self) {
+        let json = serde_json::to_string(self).unwrap();
+        let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join(".hacking/tokenizer_settings.json");
```
izeigerman, line 201: Is this path expected?

benfdking: Will fix this is a hangover from me trying to clean things up 🙏  for spotting.

sqlglotrs/benches/long.rs
```diff
@@ -0,0 +1,74 @@
+use std::path::Path;
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use sqlglotrs::settings::{TokenTypeSettings, TokenizerDialectSettings, TokenizerSettings};
+use sqlglotrs::tokenizer::Tokenizer;
+
+pub const LONG: &str = r#"
+SELECT
+  "e"."employee_id" AS "Employee #",
```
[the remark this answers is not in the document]
georgesittas: Yep, let's deal with that later; this is fine for now.

georgesittas (approved):
Re: https://github.com/tobymao/sqlglot/pull/4552#issuecomment-2574871237 (sharing internal discussion).

It’s probably fine for now. I think if the `TokenType` enum is updated, we'd simply miss the newer tokens and use an outdated mapping, but that’s fine because we don’t use those tokens anyway and we're only using the rust tokenizer directly. So, it’s like taking a snapshot of the tokens supported today and benchmarking the tokenizer using them.

Issue #4233, opened 2024-10-10
Covering compilation paths for CHAR_LENGTH, UNICODE, and ISASCII


pruzko:
Hi,

I am developing an [SQL Injection framework](https://github.com/pruzko/hakuin) and am migrating to sqlglot. There are some SQL functions that I use across several dialects that can be abstracted. If you are interested, I'd implement them directly in sqlglot.

1) number of characters (not bytes) in a string:
```
sqlite: LENGTH(x)
mysql: CHAR_LENGTH(x)
postgres: CHAR_LENGTH(x)
mssql: LENGTH(x)
oracledb: LENGTH(x)
```

2) unicode point for a character:
```
sqlite: unicode(x)
mysql: ord(cast(x as char character set utf32))
postgres: ascii(x)
mssql: unicode(x)
oracledb: ascii(unistr(x))
```

3) is ascii predicate:
```
sqlite: not x glob cast(x'2a5b5e012d7f5d2a' as text)
mysql: not regexp_like(x, '[^[:ascii:]]')
postgres: x ~ '^[[:ascii:]]*$'
mssql: patindex('%[^'+char(0x00)+'-'+char(0x7f)+']%' COLLATE Latin1_General_BIN, x) = 0
oracledb: nvl(regexp_like(x, '^['||chr(1)||'-'||chr(127)||']*$'),1)
```

I noticed you already do [abstractions even when a certain functionality is not natively supported](https://github.com/tobymao/sqlglot/blob/aee76da1cadec242f7428d23999f1752cb0708ca/sqlglot/dialects/dialect.py#L1054). The changes I'd make would be similar to [my last PR](https://github.com/tobymao/sqlglot/pull/4198).

I am rushing with some other duties, but would like to implement this in the future. What are your thoughts?

georgesittas:
Hey @pruzko,

First of all, cool project! Even cooler that you're looking to use SQLGlot in it. :)

We'll be happy to accept well-tested & documented PRs, feel free to send them our way and we'll take a look. One thing to take into consideration when working on the transpiler is that certain operations require type inference to be transpiled correctly. We've a pattern where we use `annotate_types` to achieve this in some places, but without schema it's best-effort. May or may not be relevant to the work you mentioned.

I'll go ahead and close this issue as not planned since we're not really planning to work on these. Thank you for the offer!

PR #4557, opened 2025-01-02
IS_ASCII support

pruzko:
Support for `IS_ASCII` for a few engines. Issue #4233


VaggelisD:
Hey @pruzko, thank you for the PR! As with the other ones, please share any documentation that validates your design choices, it's not straightforward why `IS_ASCII` is transpiled as such to the other dialects.

pruzko:
First thing first, `IS_ASCII` is not supported by any engine natively AFAIK. Either way, people are looking for it (e.g., [here](https://stackoverflow.com/questions/3891331/find-non-ascii-characters-in-varchar-columns-using-sql-server), [here](https://stackoverflow.com/questions/686967/how-can-i-find-unicode-non-ascii-characters-in-an-ntext-field-in-a-sql-server-20/2683496#2683496), or [here](https://www.tek-tips.com/threads/how-to-identify-non-ascii-characters-in-database-column.1666625/)) and come up with some (more) hideous solutions. There was a brief [discussion on slack](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1735651988746369) about whether this is too exotic for to have in sqlglot, but I gave it a go.

There are no docs because the are tricks of my own :mage_man:.

SQLITE:
```sql
x GLOB '*[\x00-\x7f]*' # look for any ASCII char
x GLOB '*[^\x00-\x7f]*' # look for any non-ASCII char
NOT x GLOB '*[^\x00-\x7f]*' # no non-ASCII chars were found, i.e., all chars are ASCII
NOT x GLOB '*[^\x01-\x7f]*' # the NULL byte is a string terminator in SQLite, so we can safely remove it
NOT x GLOB CAST(x'2a5b5e012d7f5d2a' as TEXT) # SQLite does not support hex-encoded char literals, so we have to convert the string at runtime
```

MySQL:
```sql
REGEXP_LIKE(x, '^[[:ascii:]]*$') # all chars are ASCII
```

Postgres:
```sql
x ~ '^[[:ascii:]]*$' # all chars are ASCII
```

TSQL:
```sql
PATINDEX('%[\x00-\x7f]%', x) # look for any ASCII char
PATINDEX('%[^\x00-\x7f]%', x) # look for any non-ASCII char
PATINDEX('%[^\x00-\x7f]%', x) # no non-ASCII chars were found, i.e., all chars are ASCII
PATINDEX('%[^'  +char(0)  '-'  +char(0x7f)  ']%', x) = 0 # TSQL does not support hex-encoded char literals, so we have to convert them at runtime
PATINDEX('%[^'  +char(0)  '-'  +char(0x7f)  ']%' COLLATE Latin1_General_BIN, x) = 0 # the pattern has non-printable chars in it, so we have to change the encoding to Latin1_General_BIN
```

Oracle:
```sql
REGEXP_LIKE(x , '^[\x01-\x7f]*$') # all chars are ASCII, also the NULL byte is a string terminator so it's removed
REGEXP_LIKE(x , '^[' || chr(0x1) || '-' || chr (127) || ']*$') # hex-encoded string literals not supported, so we resolve at runtime
NVL(REGEXP_LIKE(x , '^[' || chr(0x1) || '-' || chr (127) || ']*$'), TRUE) # assure TRUE on NULL just like the other engines
```

georgesittas (approved):

Issue #4553, opened 2024-12-31
Support Postgres `XMLTABLE` syntax


rbreejen:
**Is your feature request related to a problem?**
Currently SQLGlot does not have support for the Postgres function XMLTABLE. I looked around in the issues and it says there is Oracle support (https://github.com/tobymao/sqlglot/issues/1156). Postgres support is however still missing, it is not in the list of [supported functions](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/postgres.py#L371). While running the query for Postgres I am getting a parsing error.

**Code snippet**
```
SELECT 
    element_id, is_active, updated_at, description, reference_id 
FROM 
    XMLTABLE(
        XMLNAMESPACES(
            'http://example.com/xsd/sample/v1' AS "ex"
        ),
        '/root/item/*[local-name()="exampleElement"]'
        PASSING xml_content
        COLUMNS
            element_id   VARCHAR2(128) PATH '@id',
            is_active    BOOLEAN        PATH '@ex:isActive',
            updated_at   TIMESTAMP      PATH '@ex:updatedAt',
            description  VARCHAR2(4000) PATH 'ex:description/text()',
            reference_id UUID           PATH 'ex:reference/@ref'
    ) AS x;
```
**Python** 3.12
**sqlglot** 26.0.1

**Expected:** successful parsing
**Actual:**

`Error: Failed to parse a model definition at '/models/example.sql': Expecting ). Line 30, Col: 8.`  

**Official Documentation**
[Postgres docs: XMLTABLE](https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-XML-PROCESSING-XMLTABLE)
[Postgres dialect in SQLGlot](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/postgres.py)

georgesittas:
Hiya, appreciate the report but this is out of scope for the core team right now. Happy to accept a well-tested & documented PR if you wanna take a stab at it.

georgesittas:
Fixed by https://github.com/tobymao/sqlglot/commit/d32d26affaa7b0639abc107505db234aeb7386d4

PR #4564, opened 2025-01-05
feat(postgres): Support Postgres XMLTABLE syntax #4553

rbreejen:
Fixes https://github.com/tobymao/sqlglot/issues/4553

Reference: https://www.postgresql.org/docs/current/functions-xml.html#FUNCTIONS-XML-PROCESSING-XMLTABLE

Very similar to the Oracle implementation in https://github.com/tobymao/sqlglot/pull/1159, except the the by_ref functionality is not implemented because Postgres does not support this (yet).

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -319,6 +319,7 @@ class Postgres(Dialect):
             "BEGIN TRANSACTION": TokenType.BEGIN,
             "BIGSERIAL": TokenType.BIGSERIAL,
             "CHARACTER VARYING": TokenType.VARCHAR,
+            "COLUMNS": TokenType.COLUMN,
             "CONSTRAINT TRIGGER": TokenType.COMMAND,
             "CSTRING": TokenType.PSEUDO_TYPE,
             "DECLARE": TokenType.COMMAND,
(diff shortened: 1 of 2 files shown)


VaggelisD:
Hi @rbreejen, thank you for the PR!

Can you please rebase it to the latest main so the CI regression can be fixed? 

We'll review the changes shortly.

VaggelisD (commented):
Did a quick pass and I observed that both this & the original PR share a lot of overlapping code which we tend to avoid; The fact that `by_ref` is not supported by Postgres shouldn't make a difference for SQLGlot since the original XMLTABLE implementation treats it as an optional arg anyways.

Did you attempt to reuse the Oracle implementation as is? I'd say we should:

- Move the Oracle implementation to `parser.py` & `generator.py`
- Revert Postgres's parsing & generation
- If a specific part does not align with Postgres's syntax, we can extract it to a helper function e.g `_parse_xml_table_passing` which the Postgres parser will override


This should be enough to merge both implementations in one

georgesittas:
Closing in favor of https://github.com/tobymao/sqlglot/commit/d32d26affaa7b0639abc107505db234aeb7386d4.

rbreejen:
Looking good! Thanks for your effort!

PR #4566, opened 2025-01-06
Feat: add support for Chr in tsql and sqlite

pruzko:
I renamed `exp.Chr` for SQLite, TSQL, and added default encoding for MySQL.

```
SELECT CHAR(10) # sqlite
SELECT CHR(10) # postgres
SELECT CHAR(10 USING utf8mb4) # mysql
```
See [SQLite docs](https://www.sqlite.org/lang_corefunc.html#char) and [TSQL docs](https://learn.microsoft.com/en-us/sql/t-sql/functions/char-transact-sql?view=sql-server-ver16).

The encoding is necessary since MySQL 8.0.19 (see [this](https://stackoverflow.com/questions/65614116/in-mysql-char-function-works-weird)).

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -1236,7 +1236,8 @@ class MySQL(Dialect):
         def chr_sql(self, expression: exp.Chr) -> str:
             this = self.expressions(sqls=[expression.this] + expression.expressions)
             charset = expression.args.get("charset")
-            return f"CHAR({this} USING {charset})" if charset else f"CHAR({this})"
+            using = f" USING {self.sql(charset)}" if charset else ""
+            return f"CHAR({this}{using})"
 
         def timestamptrunc_sql(self, expression: exp.TimestampTrunc) -> str:
             unit = expression.args.get("unit")
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -1236,8 +1236,8 @@ class MySQL(Dialect):
         def chr_sql(self, expression: exp.Chr) -> str:
             this = self.expressions(sqls=[expression.this] + expression.expressions)
             charset = expression.args.get("charset")
-            using = f" USING {self.sql(charset)}" if charset else ""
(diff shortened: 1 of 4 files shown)


pruzko:
I think I misunderstood the encoding issue before. The problem is that newer mysql version *display* byte strings differently, but the return type of `CHAR` is always the same, i.e. a byte string. Hence, the whole "utfmb4"-by-default is nonsense. I reversed the changes.

sqlglot/dialects/mysql.py
```diff
@@ -1236,8 +1236,7 @@ def _oldstyle_limit_sql(self, expression: exp.Show) -> str:
         def chr_sql(self, expression: exp.Chr) -> str:
             this = self.expressions(sqls=[expression.this] + expression.expressions)
             charset = expression.args.get("charset")
-            using = f" USING {self.sql(charset)}" if charset else ""
-            return f"CHAR({this}{using})"
+            return f"CHAR({this} USING {charset})" if charset else f"CHAR({this})"
```
georgesittas, line 1239:
```suggestion
            using = f" USING {self.sql(charset)}" if charset else ""
            return f"CHAR({this}{using})"
```

tests/dialects/test_mysql.py
```diff
@@ -263,8 +263,6 @@ def test_identity(self):
         )
         self.validate_identity("INTERVAL '1' YEAR")
         self.validate_identity("DATE_ADD(x, INTERVAL '1' YEAR)")
-        self.validate_identity("CHAR(0)")
-        self.validate_identity("CHAR(77, 121, 83, 81, '76')")
         self.validate_identity("CHAR(77, 77.3, '77.3' USING utf8mb4)")
```
georgesittas, line 266:
```suggestion
        self.validate_identity("CHAR(0)")
        self.validate_identity("CHAR(77, 121, 83, 81, '76')")
        self.validate_identity("CHAR(77, 77.3, '77.3' USING utf8mb4)")
```

PR #4576, opened 2025-01-08
fix(tsql): transpile snowflake TIMESTAMP_NTZ to DATETIME2

geooo109:
Fixes [here](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1736263004411849)

This PR fixes the transpilation of the `TIMESTAMP_NTZ` to `DATETIME2`. 
Based on the snowflake's docs: `TIMESTAMP_NTZ is the default for TIMESTAMP`.

**Note** : the precision of `snowflake's TIMESTAMP_NTZ` is `0 to 9`, and the precision of `tsql's DATETIME2` is `0 to 7`. 

**Docs**
[Snowflake TIMESTAMP_NTZ](https://docs.snowflake.com/en/sql-reference/data-types-datetime#label-datatypes-timestamp-variations) | [T-SQL datetime2](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=sql-server-ver16)

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -753,6 +753,15 @@ class TestTSQL(Validator):
             },
         )
 
+        # percision of TIMESTAMP_NTZ 0 to 9 (same as TIMESTAMP)
+        # percision of DATETIME2 0 to 7
+        self.validate_all(
+            "CREATE TABLE t (col1 TIMESTAMP_NTZ(2))",
+            write={
+                "tsql": "CREATE TABLE t (col1 DATETIME2(2))",
+            },
+        )
+
     def test_types_bin(self):
         self.validate_all(
             "CAST(x as BIT)",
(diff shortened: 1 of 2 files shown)


tests/dialects/test_tsql.py
```diff
@@ -753,6 +753,15 @@ def test_types_date(self):
...
 
+        # percision of TIMESTAMP_NTZ 0 to 9 (same as TIMESTAMP)
+        # percision of DATETIME2 0 to 7
+        self.validate_all(
+            "CREATE TABLE t (col1 TIMESTAMP_NTZ(2))",
+            write={
+                "tsql": "CREATE TABLE t (col1 DATETIME2(2))",
+            },
+        )
+
```
georgesittas, line 764: The path of interest here is Snowflake -> T-SQL, so we should make that clear:

```suggestion
        self.validate_all(
            "CREATE TABLE t (col1 DATETIME2(2))",
            read={
                "snowflake": "CREATE TABLE t (col1 TIMESTAMP_NTZ(2))",
            },
            write={
                "tsql": "CREATE TABLE t (col1 DATETIME2(2))",
            },
        )
```

georgesittas (approved):

tests/dialects/test_tsql.py
geooo109: Good catch, thanks.

PR #4577, opened 2025-01-08
STRPOS to POSITION for Postgres

pruzko:
`STRPOS` works well for Postgres and the [documentation](https://www.postgresql.org/docs/9.1/functions-string.html) states that "[it is the] same as position(substring in string), but note the reversed argument order", but in reality, `POSITION(substring in string)` is more robust as it works with binary strings as well.

```sql
select strpos(decode('DEADBEEF', 'hex'), decode('BE', 'hex'));
# ERROR:  function strpos(bytea, bytea) does not exist

select position(decode('BE', 'hex') in decode('DEADBEEF', 'hex'));
#  position 
# ----------
#         3
```

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -27,7 +27,6 @@ from sqlglot.dialects.dialect import (
     build_timestamp_trunc,
     rename_func,
     sha256_sql,
-    str_position_sql,
     struct_extract_sql,
     timestamptrunc_sql,
     timestrtotime_sql,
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -721,3 +719,7 @@ def computedcolumnconstraint_sql(self, expression: exp.ComputedColumnConstraint)
 
         def isascii_sql(self, expression: exp.IsAscii) -> str:
             return f"({self.sql(expression.this)} ~ '^[[:ascii:]]*$')"
+
+        def strposition_sql(self, expression: exp.StrPosition):
+            substr = expression.args.get("substr")
+            return f"POSITION({self.sql(substr)} in {self.sql(expression.this)})"
```
georgesittas, line 725:
```suggestion
        def strposition_sql(self, expression: exp.StrPosition) -> str:
            return f"POSITION({self.sql(expression, 'substr')} IN {self.sql(expression, 'this')})"
```

georgesittas (approved):

sqlglot/dialects/postgres.py
georgesittas: I'll take this to the finish line

Issue #4575, opened 2025-01-07
Query with OFFSET statement without LIMIT in Trino get translated into invalid query in Snowflake


baruchoxman:
Way to reproduce:

```
>>> a = sqlglot.parse_one("SELECT 1 ORDER BY 1 OFFSET 0", "trino")
>>> a.sql("snowflake")
'SELECT 1 ORDER BY 1 OFFSET 0'
```

The resulting query is not valid in Snowflake:
<img width="601" alt="image" src="https://github.com/user-attachments/assets/2d5afa90-4e4f-4dce-9985-b56462f8d57f" />

Also, the parsing of the following fails, while being valid in Trino:
```
>>> a = sqlglot.parse_one("SELECT 1 OFFSET 0", "trino")
...
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 17.
  SELECT 1 OFFSET 0
```

https://trino.io/docs/current/sql/select.html#offset-clause

PR #4581, opened 2025-01-08
fix(snowflake): generate LIMIT when OFFSET exists #4575

geooo109:
Fixes snowflake problem #4575 

This PR allows the generation of snowflake queries when the `OFFSET` is present in the input dialect without a `LIMIT`.

Based on the snowflake's doc, an `OFFSET` must always match with a `LIMIT`. Based on the discussion [here](https://snowflake.discourse.group/t/how-do-i-use-offset-without-limit/5342), a `LIMIT NULL` can be inserted into the query in order to be parsed correctly. 

Example:
```
>>> sqlglot.transpile("SELECT 1 ORDER BY 1 OFFSET 0", read="trino", write="snowflake")
['SELECT 1 ORDER BY 1 LIMIT NULL OFFSET 0']
```

**Docs**
[Snowflake LIMIT](https://docs.snowflake.com/en/sql-reference/constructs/limit#postgresql-syntax)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1229,10 +1229,5 @@ class Snowflake(Dialect):
             limit = expression.args.get("limit")
             offset = expression.args.get("offset")
             if offset and not limit:
-                expression.set(
-                    "limit",
-                    exp.Limit(
-                        expression=exp.Null(),
-                    ),
-                )
+                expression.limit(exp.Null(), copy=False)
             return super().select_sql(expression)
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -1224,3 +1224,15 @@ def datesub_sql(self, expression: exp.DateSub) -> str:
...
+        def select_sql(self, expression: exp.Select) -> str:
+            limit = expression.args.get("limit")
+            offset = expression.args.get("offset")
+            if offset and not limit:
+                expression.set(
+                    "limit",
+                    exp.Limit(
+                        expression=exp.Null(),
+                    ),
+                )
```
georgesittas, line 1237:
```suggestion
                expression.limit(exp.Null(), copy=False)
```

georgesittas (approved):
Curious if this fix is only relevant in Snowflake.

PR #4588, opened 2025-01-09
feat: move to rustc fx hashmap

benfdking:
mostly just to test the benchmark job but this improves parsing performance by 17% on long.

### DIFF
--- a/sqlglotrs/Cargo.lock
+++ b/sqlglotrs/Cargo.lock
@@ -449,6 +449,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 
 [[package]]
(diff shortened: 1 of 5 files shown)


benfdking:
Works but can't comment from a fork

image

georgesittas (approved):
@tobymao waiting for you to take a look here before merging. Even though `rustc-hash` is under `dependencies`, my understanding is that this should be fine because:

1. It is used in `rustc` itself, see [repo](https://github.com/rust-lang/rustc-hash).
2. We pre-compile the rust tokenizer and ship the various binaries when releasing `sqlglotrs`, so users shouldn't encounter any dependency issues. @benfdking to verify.

tobymao (approved):

benfdking:
I don't expect this to affect anything. For those who don't have prebuilt binaries, python will build it with cargo and there it is done without any user intervention so no more invasive than what's there already.

PR #4590, opened 2025-01-09
feat(clickhouse): support scalar CTE in query builder

pkit:
Example:

```python
select("x")
.with_("v", as_=select("foo").from_("bar"), scalar=True)
.from_("t")
.where("id = v")
```

```sql
WITH (SELECT foo FROM bar) AS v
SELECT x FROM t WHERE id = v
```

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1224,6 +1224,7 @@ class Query(Expression):
         append: bool = True,
         dialect: DialectType = None,
         copy: bool = True,
+        scalar: bool = False,
         **opts,
     ) -> Q:
         """
@@ -1244,6 +1245,7 @@ class Query(Expression):
                 Otherwise, this resets the expressions.
             dialect: the dialect used to parse the input expression.
             copy: if `False`, modify this expression instance in-place.
+            scalar: if `True`, this is a scalar common table expression.
             opts: other options to use to parse the input expressions.
 
         Returns:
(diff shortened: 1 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -7036,11 +7039,15 @@ def _apply_cte_builder(
...
     copy: bool = True,
+    scalar: bool = False,
     **opts,
 ) -> E:
     alias_expression = maybe_parse(alias, dialect=dialect, into=TableAlias, **opts)
     as_expression = maybe_parse(as_, dialect=dialect, **opts)
-    cte = CTE(this=as_expression, alias=alias_expression, materialized=materialized)
+    if scalar and not isinstance(as_expression, Subquery):
+        # scalar CTE must be wrapped in a subquery
+        as_expression = Subquery(this=as_expression)
```
tobymao, line 7049: maybe_copy(as_expression, copy=copy),

tobymao: we don't want to mutate as_expression (this sets the parent)

pkit: but `as_expression` is parsed earlier, i.e. it's probably a "copy" already?

pkit: ah, ok, it `maybe_parse`d

tobymao: you're right, let's do that instead, pass copy into maybe_parse

PR #4596, opened 2025-01-10
Fix!: exp.Merge condition for Trino/Postgres

### LINKED ISSUE #4595: Trino - when matched ambiguous column when using a function
- Read dialect Trino
- Write dialect Trino
- Still exists in main

```sql
MERGE INTO table_a AS target USING(
  SELECT
    pk,
    my_array
  FROM table_b
) AS source ON source.pk = target.pk
WHEN MATCHED THEN UPDATE SET target.my_array = ARRAY_DISTINCT(CONCAT(source.my_array, target.my_array))
WHEN NOT MATCHED THEN INSERT (pk, my_array)
VALUES (
  source.pk,
  source.my_array
)
```

As the WHEN MATCHED is a function then this should be possible but there's a transformation applied that strips out the table aliases in the THEN part.
This then leads us to an ambiguous column error as target.my_array is transformed to just my_array.

Might a solution be to also consider if the column's parent is a function? In which case if the merge target is referenced then it is to remove ambiguity.
I'll raise a PR with such a solution shortly and will await comments.

**Official Documentation**

MikeWallis42:
Fixes #4595

Further building upon https://github.com/tobymao/sqlglot/pull/3940 there was a gap whereby the UPDATE part of a MERGE query needed to be fully qualified if it's used within a function.
The specific example we had was when we were concatenating arrays but I've substituted for a coalesce in the unit test.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1578,7 +1578,11 @@ def merge_without_target_sql(self: Generator, expression: exp.Merge) -> str:
             then.transform(
                 lambda node: (
                     exp.column(node.this)
-                    if isinstance(node, exp.Column) and normalize(node.args.get("table")) in targets
+                    if (
+                        isinstance(node, exp.Column)
+                        and normalize(node.args.get("table")) in targets
+                        and not isinstance(node.parent, exp.Func)
+                    )
                     else node
                 ),
                 copy=False,
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1570,23 +1570,25 @@ def merge_without_target_sql(self: Generator, expression: exp.Merge) -> str:
         targets.add(normalize(alias.this))
 
     for when in expression.args["whens"].expressions:
-        # only remove the target names from the THEN clause
-        # theyre still valid in the <condition> part of WHEN MATCHED / WHEN NOT MATCHED
-        # ref: https://github.com/TobikoData/sqlmesh/issues/2934
-        then = when.args.get("then")
+        # only remove the target table names from certain parts of WHEN MATCHED / WHEN NOT MATCHED
+        # they are still valid in the <condition>, the right hand side of each UPDATE and the VALUES part
+        # (not the column list) of the INSERT
+        then: exp.Insert | exp.Update | None = when.args.get("then")
         if then:
-            then.transform(
-                lambda node: (
-                    exp.column(node.this)
+            if isinstance(then, exp.Update):
+                for equals in then.find_all(exp.EQ):
+                    equal_lhs = equals.this
                     if (
-                        isinstance(node, exp.Column)
-                        and normalize(node.args.get("table")) in targets
-                        and not isinstance(node.parent, exp.Func)
-                    )
-                    else node
-                ),
-                copy=False,
-            )
+                        isinstance(equal_lhs, exp.Column)
+                        and normalize(equal_lhs.args.get("table")) in targets
+                    ):
+                        equal_lhs.replace(exp.column(equal_lhs.this))
+            if isinstance(then, exp.Insert):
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -1578,7 +1578,11 @@ def normalize(identifier: t.Optional[exp.Identifier]) -> t.Optional[str]:
             then.transform(
                 lambda node: (
                     exp.column(node.this)
-                    if isinstance(node, exp.Column) and normalize(node.args.get("table")) in targets
+                    if (
+                        isinstance(node, exp.Column)
+                        and normalize(node.args.get("table")) in targets
+                        and not isinstance(node.parent, exp.Func)
```
georgesittas, line 1584: 1. Is this fix correct for both Postgres and Trino? Did you test it in both engines? How was it tested?
2. Are functions the only category that need a fully-qualified name to avoid ambiguity issues? What if you have an addition instead of a function, etc?

georgesittas:
Thanks for the PR @MikeWallis42, asked some questions to understand this better.

sqlglot/dialects/dialect.py
MikeWallis42: Both fair challenges. I didn't explicitly test with Postgres but did with Trino.
But you're correct in that there are probably still gaps, I'll think about this properly and hope to make another commit shortly.

MikeWallis42:
@georgesittas, I've pushed up what I think is a better solution, I haven't checked Postgres specifically but the tests are still passing.
The documentation seems to suggest that it should be similar to Trino, I've focused on stripping relevant table aliases from where expression is not mentioned in the following documentation.

https://trino.io/docs/current/sql/merge.html#merge
https://www.postgresql.org/docs/current/sql-merge.html

*Trino*
```
WHEN MATCHED [ AND condition ]
    THEN UPDATE SET ( column = expression [, ...] )
```
```
WHEN NOT MATCHED [ AND condition ]
    THEN INSERT [ column_list ] VALUES (expression, ...)
```

*Postgres*
```
UPDATE SET { column_name = { expression | DEFAULT } |
             ( column_name [, ...] ) = [ ROW ] ( { expression | DEFAULT } [, ...] ) |
             ( column_name [, ...] ) = ( sub-SELECT )
           } [, ...]
```
```
INSERT [( column_name [, ...] )]
[ OVERRIDING { SYSTEM | USER } VALUE ]
{ VALUES ( { expression | DEFAULT } [, ...] ) | DEFAULT VALUES }
```

My previous implementation wouldn't have been comprehensive enough as you suggested.

sqlglot/dialects/dialect.py
```diff
@@ -1570,19 +1570,25 @@ def normalize(identifier: t.Optional[exp.Identifier]) -> t.Optional[str]:
...
+                    equal_lhs = equals.this
+                    if (
+                        isinstance(equal_lhs, exp.Column)
+                        and normalize(equal_lhs.args.get("table")) in targets
+                    ):
+                        equal_lhs.replace(exp.column(equal_lhs.this))
+            if isinstance(then, exp.Insert):
+                column_list = then.args.get("this")
+                if column_list is not None:
+                    for column in column_list.find_all(exp.Column):
```
georgesittas, line 1589:
```suggestion
                column_list = then.this
                if isinstance(column_list, exp.Tuple):
                    for column in column_list.expressions:
```

georgesittas (approved):
These changes look reasonable. I was able to run this successfully in Postgres as a quick sanity check:

```sql
georgesittas=# MERGE INTO foo AS target USING (SELECT a, b FROM tbl) AS src ON src.a = target.a WHEN MATCHED THEN UPDATE SET b = COALESCE(src.b, target.b) WHEN NOT MATCHED THEN INSERT (a, b) VALUES (src.a, src.b);

MERGE 1
```

PR #4598, opened 2025-01-10
feat: Firebolt dialect

betodealmeida:
We found a problem when using `sqlglot` in Apache Superset to process (parse, modify the AST, and re-render) Firebolt queries. The problem happens when we run a query with a predicate of `col NOT IN` and it gets translated to `NOT col IN` by the generic parser:

```python
>>> sqlglot.parse_one("SELECT * FROM tbl WHERE i NOT IN (1, 2)").sql()
'SELECT * FROM tbl WHERE NOT i IN (1, 2)'
```

While both queries are semantically equivalent, the first one is valid in Firebolt, while the second is not. This happens because `NOT` has higher precedence than `IN`:

```sql
CREATE TABLE tbl (i INT, b BOOLEAN, t TEXT);
INSERT INTO tbl VALUES (0, FALSE, 'false'), (1, TRUE, 'true');
SELECT * FROM tbl WHERE NOT i IN (1, 2);
```

> firebolt error: - Line 4, Column 25: operator 'not' for input types (integer) not found, try adding explicit casts

This only happens when checking for a range; equality (and other comparisons) works as expected:

```sql
SELECT * FROM tbl WHERE NOT i = 1;   -- this works
```

This PR introduces a custom Firebolt dialect that fixes the problem:

```python
>>> sqlglot.parse_one("SELECT * FROM tbl WHERE i NOT IN (1, 2)").sql("firebolt")
'SELECT * FROM tbl WHERE NOT (i IN (1, 2))'
```

### DIFF
--- a/sqlglot/dialects/__init__.py
+++ b/sqlglot/dialects/__init__.py
@@ -70,6 +70,7 @@ from sqlglot.dialects.doris import Doris
 from sqlglot.dialects.drill import Drill
 from sqlglot.dialects.druid import Druid
 from sqlglot.dialects.duckdb import DuckDB
+from sqlglot.dialects.firebolt import Firebolt
 from sqlglot.dialects.hive import Hive
 from sqlglot.dialects.materialize import Materialize
 from sqlglot.dialects.mysql import MySQL
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -64,6 +64,7 @@ class Dialects(str, Enum):
     DRILL = "drill"
     DRUID = "druid"
     DUCKDB = "duckdb"
+    FIREBOLT = "firebolt"
     HIVE = "hive"
     MATERIALIZE = "materialize"
     MYSQL = "mysql"
--- a/sqlglot/dialects/firebolt.py
+++ b/sqlglot/dialects/firebolt.py
@@ -1,0 +1,35 @@
+import typing as t
+
+from sqlglot import exp, generator, parser
+from sqlglot.dialects.dialect import Dialect
+from sqlglot.tokens import TokenType
+
+
+class Firebolt(Dialect):
+    class Parser(parser.Parser):
+        UNARY_PARSERS = {
+            **parser.Parser.UNARY_PARSERS,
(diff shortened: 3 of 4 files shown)


georgesittas:
I'm skeptical about introducing yet another dialect, because it comes with a maintenance burden for the core team. We probably won't prioritize support for this, so perhaps it'd be better if you folks do some kinda transformation in the AST to inject parentheses, or hook into the existing logic to adapt it to your needs.

cc @tobymao

betodealmeida:
@tobymao @georgesittas thanks, I totally understand that.

Would it makes sense to add a mechanism to register 3rd party dialects? We could use [entry points](https://packaging.python.org/en/latest/specifications/entry-points/) (used by Superset and SQLAlchemy, eg) or a registry singleton (used by SQLAlchemy, eg). With Superset, people can develop their own DB adapters and publish them as separate packages, and once they are installed Superset is able to find and use them automatically. This is something I'd be happy to work on for `sqlglot`.

From what I looked right now, it's not trivial to inject a 3rd party dialect in `sqlglot`, it would require monkeypatching. A proper interface would make it much easier. The only downside is that dialects would become a "public interface", but you could always tell 3rd party developers that there's no stable interface, and 3rd party developers could just pin their packages to a specific `sqlglot` version.

Let me know your thoughs! Thanks!

tobymao:
what does that look like? isn't it easy to make your own dialect? you can just subclass it and it should work

betodealmeida:
Right, it works, but only if you know what dialect you want beforehand. For Superset, where the dialect is dynamic, we have to manually inject it with:

```python
from superset.sql.dialects.firebolt import Firebolt

Dialect.classes["firebolt"] = Firebolt
```

And we need to do that in the right place to make sure that it's done before we use `sqlglot`. It's actually easier than I expected, I thought I would have to monkeypatch the `Dialects` enum.

But with entry points, we would simply add to our `setup.py`:

```python
setup(
    ...,
    entry_points={
        "sqlglot.dialects": ["firebolt = superset.sql.parse.firebolt:Firebolt"],
        ...,
    },
)
```

Of course `sqlglot` would need to have the logic of finding all classes registered under the `sqlglot.dialects` entry point, and add them to `_Dialect.classes`.

PR #4604, opened 2025-01-13
Fix(snowflake): treat $ as part of the json path key identifier

georgesittas:


sqlglot/dialects/snowflake.py
```diff
@@ -375,6 +375,12 @@ def quote_identifier(self, expression: E, identify: bool = True) -> E:
 
         return super().quote_identifier(expression, identify=identify)
 
+    class JSONPathTokenizer(jsonpath.JSONPathTokenizer):
+        SINGLE_TOKENS = {
+            **jsonpath.JSONPathTokenizer.SINGLE_TOKENS,
+            "$": TokenType.VAR,
```
tobymao, line 381: is it better to pop dollar?

georgesittas: sure we can pop it

tobymao (approved):

Issue #4601, opened 2025-01-12
StringPosition across all dialects


pruzko:
SQLGlot partially supports `StringPosition` for various dialects, but not all. Furthermore, the code around `StringPosition`, namely `str_position_sql` got a bit messy.

I intend to implement this for all supported dialects and want to follow these principles:
- Common functions will be parsed in the root `Parser`.
- The strongest (i.e., the one with the most supported optional parameters) non-conflicting (i.e., no dialect is broken, only extended) synopsis for a given function will be the default. Dialect-specific changes are handled in their corresponding parsers.
- The strongest function among those that a dialect supports is used for code generation.
- Where `position` is not supported by the strongest function, it is emulated using `STRPOS`
- Where `occurrence` (now called `instance` in SQLGlot) is not supported, it is dropped.

Aside from wide coverage, this approach allows us to use the `SUBSTR` only as a last resort, leaving the generated code clean.

Here is the corresponding documentation:

| Dialect     | `position()`                                                                                                                                    | `strpos()`                                                                                          | `instr()`                                                                                                                                    | `locate()`                                                                                                                                    | `charindex()`                                                                                           | `index()`                                                                                                                                                          | `find()`                                                                                    | `findnth()`                                                                                    |
| ----------- | ----------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
| athena      | [X](https://docs.aws.amazon.com/redshift/latest/dg/r_POSITION.html)                                                                             | [X](https://docs.aws.amazon.com/redshift/latest/dg/r_STRPOS.html)                                   |                                                                                                                                              |                                                                                                                                               | [X](https://docs.aws.amazon.com/redshift/latest/dg/r_CHARINDEX.html)                                    |                                                                                                                                                                    |                                                                                             |                                                                                                |
[...]

PR #4606, opened 2025-01-14
Feat!: POSITION and all their variants for all dialects

pruzko:
I implemented the changes proposed in #4601 

There is a horrible mess around how DBs handle `POSITION`, as the synopses can be `POSITION(needle in haystack)`, `POSITION(haystack, needle)`, and also `POSITION(needle, haystack)`. Consequently, I decided to simplify generation to simple `POSITION(needle in haystack)` (i.e., the ANSII standard) when necessary, otherwise I use an equivalent alternative, such as `LOCATE(needle, haystack)`. This choice affects only code generation, parsing is unaffected. Note that this can alter some round-trips, but the generated code is valid and clean.

I've also moved all tests to `test_dialect` and added many more. The [Presto test](https://github.com/tobymao/sqlglot/blob/main/tests/dialects/test_presto.py#L206) for `STRPOS` assumes that trino supports the `occurrence` (i.e., `instance`) parameter, but [it does not](https://trino.io/docs/current/functions/string.html#strpos). The test if fixed now.

Lastly, there was a bug that would mess up the `occurrence` parameter when `position` is not provided. This is now also fixed by setting default `position` to 1 so that the `occurrence` is passed as the 4th argument.

Docs
---------
[BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#instr) |  [Clickhouse](https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#position) | [DuckDB](https://duckdb.org/docs/sql/functions/char.html#strposstring-search_string) | [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/position) | [MySQL](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_locate) | [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/INSTR.html) | [Spark/DB](https://docs.databricks.com/en/sql/language-manual/functions/instr.html) | [Presto/Trino](https://prestodb.io/docs/current/functions/string.html#strpos-string-substring-instance-bigint) | [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/functions/charindex-transact-sql?view=sql-server-ver16) | [Postgres](https://www.postgresql.org/docs/9.1/functions-string.html) | [SQLite](https://www.sqlite.org/lang_corefunc.html#instr)


sqlglot/dialects/teradata.py
```diff
@@ -167,6 +168,7 @@ class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "CARDINALITY": exp.ArraySize.from_arg_list,
+            # "INDEX": lambda args: exp.StrPosition(this=seq_get(args, 0), substr=seq_get(args, 1)),
```
VaggelisD, line 171: Dead code?

sqlglot/dialects/trino.py
```diff
@@ -48,6 +53,11 @@ class Generator(Presto.Generator):
             exp.ArraySum: lambda self,
             e: f"REDUCE({self.sql(e, 'this')}, 0, (acc, x) -> acc + x, acc -> acc)",
             exp.Merge: merge_without_target_sql,
+            exp.StrPosition: (
+                lambda self, e: strposition_sql(
+                    self, e, func_name="STRPOS", supports_position=False
+                )
```
VaggelisD, line 59: It seems to me most of the `strposition_sql` calls use these args, maybe we could make it so that these are the default values for `func_name` & `supports_position`?

sqlglot/parser.py
```diff
@@ -232,6 +231,27 @@ class Parser(metaclass=_Parser):
...
+            this=seq_get(args, 0),
+            substr=seq_get(args, 1),
+            position=seq_get(args, 2),
+            occurrence=seq_get(args, 3),
+        ),
+        "LOCATE": lambda args: exp.StrPosition(
+            this=seq_get(args, 1),
+            substr=seq_get(args, 0),
+            position=seq_get(args, 2),
+        ),
```
VaggelisD, line 254: Since these are repeated, I think it'd be cleaner if we encapsulate them into a builder function that will:
- Return `from_arg_list` if it's input is `"INSTR"` or `"STRPOS"`
- Return the highlighted part for `"LOCATE"` or `"CHARINDEX"`

tests/dialects/test_dialect.py
```diff
@@ -1682,42 +1682,70 @@ def test_operators(self):
...
-            write={
-                "drill": "STRPOS(haystack, needle)",
-                "duckdb": "STRPOS(haystack, needle)",
-                "postgres": "STRPOS(haystack, needle)",
-                "presto": "STRPOS(haystack, needle)",
-                "spark": "LOCATE(needle, haystack)",
-                "clickhouse": "position(haystack, needle)",
-                "snowflake": "POSITION(needle, haystack)",
+            "STR_POSITION(haystack, needle)",
+            read={
```
VaggelisD, line 1686: Doesn't parsing work for the dialects that are missing from `read` here, e.g BQ or the Hive hierarchy?

tests/dialects/test_dialect.py
```diff
@@ -1682,42 +1682,70 @@ def test_operators(self):
...
-                "spark": "LOCATE(needle, haystack)",
-                "clickhouse": "position(haystack, needle)",
-                "snowflake": "POSITION(needle, haystack)",
+                "hive": "LOCATE(needle, haystack)",
                 "mysql": "LOCATE(needle, haystack)",
+                "oracle": "INSTR(haystack, needle)",
+                "postgres": "POSITION(needle IN haystack)",
+                "presto": "STRPOS(haystack, needle)",
+                "snowflake": "CHARINDEX(needle, haystack)",
+                "spark2": "LOCATE(needle, haystack)",
```
VaggelisD, line 1706: Can we include `spark` and `databricks` here too?

tests/dialects/test_dialect.py
```diff
@@ -1682,42 +1682,70 @@ def test_operators(self):
...
+                "clickhouse": "LOCATE(needle, haystack, position)",
+                "drill": "`IF`(STRPOS(SUBSTRING(haystack, position), needle) = 0, 0, STRPOS(SUBSTRING(haystack, position), needle) + position - 1)",
+                "duckdb": "CASE WHEN STRPOS(SUBSTRING(haystack, position), needle) = 0 THEN 0 ELSE STRPOS(SUBSTRING(haystack, position), needle) + position - 1 END",
+                "hive": "LOCATE(needle, haystack, position)",
+                "mysql": "LOCATE(needle, haystack, position)",
+                "oracle": "INSTR(haystack, needle, position)",
+                "postgres": "CASE WHEN POSITION(needle IN SUBSTRING(haystack FROM position)) = 0 THEN 0 ELSE POSITION(needle IN SUBSTRING(haystack FROM position)) + position - 1 END",
+                "presto": "IF(STRPOS(SUBSTRING(haystack, position), needle) = 0, 0, STRPOS(SUBSTRING(haystack, position), needle) + position - 1)",
+                "snowflake": "CHARINDEX(needle, haystack, position)",
+                "spark2": "LOCATE(needle, haystack, position)",
```
VaggelisD, line 1735: Ditto, let's include `spark` & `databricks` too

VaggelisD (commented):
Hey @pruzko, thank you for the PR! As with the previous ones:

- Please make sure to provide the relevant documentation, I went ahead and added them for the first-class dialects
- For confirmation, did you validate that the new representations run on the respective engines? We should make sure the roundtrips are (semantically) preserved

pruzko:
I will deal with this next week, but before that:

1) The documentation is in the corresponding issue #4601.
2) The `INDEX(x)` that is commented out is something I would like to discuss. `INDEX(x)` is a legitimate (though discouraged) `POSITION` alternative. The issue is that `INDEX` is also a keyword so the parsers clash. Fixing this would require redesigning the parser so that it respects the dialect's grammar more. I suggest we just drop `INDEX(x)` instead and move on.
3) I've tested this on SQLite, Oracle, Postgres, Mysql, and TSQL. I don't have other engines installed so not sure how to test this.

sqlglot/dialects/trino.py
pruzko: I would really want to make `func_name` obligatory to make everyone else looking at the code (and modifying it) to understand what arguments are supported. Also, I tried to minimize the code by using the most default supported arguments based on the table in #4601. Most functions support position but do not support occurrence.

VaggelisD: Also, [Trino's STRPOS](https://trino.io/docs/current/functions/string.html#strpos) does support `instance`, note the 2 variants. With that in mind, we can revert the changes in `trino.py` and inherit Presto's implementation

tests/dialects/test_presto.py
```diff
@@ -204,14 +204,13 @@ def test_regex(self):
...
-            },
+            "STRPOS(haystack, needle, occurence)",
             write={
-                "presto": "STRPOS('ABC', 'A', 3)",
-                "trino": "STRPOS('ABC', 'A', 3)",
-                "snowflake": "POSITION('A', 'ABC')",
+                "bigquery": "INSTR(haystack, needle, 1, occurence)",
+                "oracle": "INSTR(haystack, needle, 1, occurence)",
+                "presto": "STRPOS(haystack, needle, occurence)",
+                "trino": "STRPOS(haystack, needle)",
```
VaggelisD, line 212: Ditto about Trino supporting `occurrence`, also a typo in all these `occurrence`s 😄

VaggelisD (commented):
> The documentation is in the corresponding issue 

The table-based comparison is great, thank you.

> The INDEX(x) that is commented out is something I would like to discuss. INDEX(x) is a legitimate (though discouraged) POSITION alternative ...

Since `INDEX` is discouraged and `INSTR` is the "strongest" function for Teradata, is there any reason in keeping it around? I'd also suggest we drop it.

tests/dialects/test_dialect.py
VaggelisD: That might be the case now, but we'll never know if we silently break something.

This is what happened with the first STRPOS PR which broke Postgres's position arg, the relevant test didn't include postgres and the regression wasn't caught.

Imo, let's make the tests as complete as possible, they'll be there forever :)

sqlglot/dialects/trino.py
VaggelisD: Generally, SQLGlot's ideology is to simplify & reuse code as much as possible; If we can add default values and skip defining strposition_sql in a couple dialects this will reduce noise.

Besides, this PR seems complete enough that STRPOS shouldn't be an issue for a while.

Also, [Trino's STRPOS](https://trino.io/docs/current/functions/string.html#strpos) does support `instance`, note the 2 variants. With that in mind, we can revert the changes in `trino.py` and inherit Presto's implementation

sqlglot/dialects/spark.py
```diff
@@ -3,7 +3,7 @@
 import typing as t
 
 from sqlglot import exp
-from sqlglot.dialects.dialect import rename_func, unit_to_var, timestampdiff_sql, build_date_delta
+from sqlglot.dialects.dialect import build_date_delta, rename_func, timestampdiff_sql, unit_to_var
```
VaggelisD, line 6: Can we revert that to the previous one liner to remove the `spark.py` diff or is it a `make style` issue?

sqlglot/dialects/tableau.py
```diff
@@ -39,9 +40,24 @@ def count_sql(self, expression: exp.Count) -> str:
...
     class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "COUNTD": lambda args: exp.Count(this=exp.Distinct(expressions=args)),
+            "FIND": lambda args: exp.StrPosition(this=seq_get(args, 0), substr=seq_get(args, 1)),
+            "FINDNTH": lambda args: exp.StrPosition(
+                this=seq_get(args, 0),
+                substr=seq_get(args, 1),
+                position=seq_get(args, 2),
+            ),
```
VaggelisD, line 61: Can simplify this down to `exp.StrPosition.from_arg_list` since the args are consecutive

VaggelisD (commented):
Thanks for the iterations, got 2 more comments from my latest pass, I'd like to do a final one to verify the APIs for each dialect before merging.

sqlglot/dialects/dialect.py
```diff
@@ -1073,36 +1073,47 @@ def property_sql(self: Generator, expression: exp.Property) -> str:
...
+        string = exp.Substring(this=string, start=position)
+
+    args = [substr, string] if func_name in ["LOCATE", "CHARINDEX"] else [string, substr]
+    if supports_position:
+        args.append(position)
+    if supports_occurrence:
+        args.append(occurrence)
+
+    if func_name == "POSITION":
+        args = [exp.In(this=substr, field=string)]
```
georgesittas, line 1103: Use an `else` branch and put L1096-L1100 in it.

sqlglot/dialects/postgres.py
```diff
@@ -584,7 +584,7 @@ class Generator(generator.Generator):
                 ]
             ),
             exp.SHA2: sha256_sql,
-            exp.StrPosition: str_position_sql,
+            exp.StrPosition: lambda self, e: (strposition_sql(self, e, func_name="POSITION")),
```
georgesittas, line 587:
```suggestion
            exp.StrPosition: lambda self, e: strposition_sql(self, e, func_name="POSITION"),
```

sqlglot/dialects/sqlite.py
```diff
@@ -200,9 +200,7 @@ class Generator(generator.Generator):
                     transforms.eliminate_semi_and_anti_joins,
                 ]
             ),
-            exp.StrPosition: lambda self, e: str_position_sql(
-                self, e, str_position_func_name="INSTR"
-            ),
+            exp.StrPosition: lambda self, e: (strposition_sql(self, e, func_name="INSTR")),
```
georgesittas, line 203:
```suggestion
            exp.StrPosition: lambda self, e: strposition_sql(self, e, func_name="INSTR"),
```

sqlglot/dialects/tableau.py
```diff
@@ -39,9 +40,20 @@ def count_sql(self, expression: exp.Count) -> str:
...
+                expression.args.get("substr"),
+                position,
+            )
+
     class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "COUNTD": lambda args: exp.Count(this=exp.Distinct(expressions=args)),
+            "FIND": exp.StrPosition.from_arg_list,
+            "FINDNTH": exp.StrPosition.from_arg_list,
```
georgesittas, line 57: This doesn't look correct, the signature is `FINDNTH(string, substring, occurrence)`, so the third argument should be stored in `occurrence`, if any. Need to fix `strposition_sql` as well.

sqlglot/dialects/trino.py
```diff
@@ -63,6 +68,7 @@ class Generator(Presto.Generator):
             e: f"REDUCE({self.sql(e, 'this')}, 0, (acc, x) -> acc + x, acc -> acc)",
             exp.ArrayUniqueAgg: lambda self, e: f"ARRAY_AGG(DISTINCT {self.sql(e, 'this')})",
             exp.Merge: merge_without_target_sql,
+            exp.StrPosition: lambda self, e: strposition_sql(self, e, supports_occurrence=True),
```
georgesittas, line 71: This should be inherited from Presto, so changes in this file can be reverted altogether.

sqlglot/dialects/dialect.py
```diff
@@ -1073,36 +1073,47 @@ def property_sql(self: Generator, expression: exp.Property) -> str:
...
+        position = one
+
+    if position and not supports_position:
+        string = exp.Substring(this=string, start=position)
+
+    args = [substr, string] if func_name in ["LOCATE", "CHARINDEX"] else [string, substr]
+    if supports_position:
+        args.append(position)
+    if supports_occurrence:
+        args.append(occurrence)
```
georgesittas, line 1100: We can do `elif occurrence: self.unsupported(...)` in L1101, to ensure we won't silently drop an `occurrence` argument, if present but not supported.

tests/dialects/test_clickhouse.py
```diff
@@ -85,8 +85,6 @@ def test_clickhouse(self):
         self.validate_identity("SELECT exponentialTimeDecayedAvg(60)(a, b)")
         self.validate_identity("levenshteinDistance(col1, col2)", "editDistance(col1, col2)")
         self.validate_identity("SELECT * FROM foo WHERE x GLOBAL IN (SELECT * FROM bar)")
-        self.validate_identity("position(haystack, needle)")
-        self.validate_identity("position(haystack, needle, position)")
```
georgesittas, line 89: These should be reverted.

tests/dialects/test_clickhouse.py
```diff
@@ -398,10 +396,6 @@ def test_clickhouse(self):
                 "clickhouse": "SELECT quantileIf(0.5)(a, TRUE)",
             },
         )
-        self.validate_all(
-            "SELECT position(needle IN haystack)",
-            write={"clickhouse": "SELECT position(haystack, needle)"},
-        )
```
georgesittas, line 404: Ditto, while we're at it let's also change `validate_all` into `validate_identity(in, out)`.

tests/dialects/test_bigquery.py
```diff
@@ -1590,14 +1590,6 @@ def test_bigquery(self):
...
             },
         )
-        self.validate_all(
-            "SELECT STRPOS('foo@example.com', '@')",
-            write={
-                "bigquery": "SELECT STRPOS('foo@example.com', '@')",
-                "duckdb": "SELECT STRPOS('foo@example.com', '@')",
-                "snowflake": "SELECT POSITION('@', 'foo@example.com')",
-            },
-        )
```
georgesittas, line 1600: Let's avoid removing tests; instead, rewrite as necessary to demonstrate the updated behavior.

tests/dialects/test_dialect.py
```diff
@@ -1682,42 +1682,204 @@ def test_operators(self):
...
+            read={
+                "databricks": "CHARINDEX(needle, haystack, position)",
+                "snowflake": "CHARINDEX(needle, haystack, position)",
+                "tsql": "CHARINDEX(needle, haystack, position)",
+            },
+        )
+        self.validate_all(
+            "STR_POSITION(haystack, needle, position)",
+            read={
+                "tableau": "FINDNTH(haystack, needle, position)",
```
georgesittas, line 1842: This entry needs to be removed, the third argument of `FINDNTH` is `occurrence,` not `position`.

tests/dialects/test_dialect.py
```diff
@@ -1682,42 +1682,204 @@ def test_operators(self):
...
+                "oracle": "INSTR(haystack, needle, position)",
+                "postgres": "CASE WHEN POSITION(needle IN SUBSTRING(haystack FROM position)) = 0 THEN 0 ELSE POSITION(needle IN SUBSTRING(haystack FROM position)) + position - 1 END",
+                "presto": "IF(STRPOS(SUBSTRING(haystack, position), needle) = 0, 0, STRPOS(SUBSTRING(haystack, position), needle) + position - 1)",
+                "redshift": "CASE WHEN POSITION(needle IN SUBSTRING(haystack FROM position)) = 0 THEN 0 ELSE POSITION(needle IN SUBSTRING(haystack FROM position)) + position - 1 END",
+                "risingwave": "CASE WHEN POSITION(needle IN SUBSTRING(haystack FROM position)) = 0 THEN 0 ELSE POSITION(needle IN SUBSTRING(haystack FROM position)) + position - 1 END",
+                "snowflake": "CHARINDEX(needle, haystack, position)",
+                "spark": "LOCATE(needle, haystack, position)",
+                "spark2": "LOCATE(needle, haystack, position)",
+                "sqlite": "IIF(INSTR(SUBSTRING(haystack, position), needle) = 0, 0, INSTR(SUBSTRING(haystack, position), needle) + position - 1)",
+                "tableau": "FINDNTH(haystack, needle, position)",
```
georgesittas, line 1864: Ditto.

tests/dialects/test_hive.py
```diff
@@ -585,15 +585,6 @@ def test_hive(self):
                 "spark": "LOCATE('a', x)",
             },
         )
-        self.validate_all(
```
georgesittas, line 588: Ditto about removing.

georgesittas (commented):
@pruzko, really appreciate the work you've done here– this is a great contribution!

Just make sure to keep PRs as small as possible in the future. Lengthy PRs are hard to review and it's easy to introduce regressions due to oversights. Instead, consider working on features incrementally and sharing a plan for how the refactor will look like.

Thanks 👍

sqlglot/parser.py
```diff
@@ -153,6 +153,16 @@ def build_coalesce(args: t.List, is_nvl: t.Optional[bool] = None) -> exp.Coalesc
...
+def build_strposition(args: t.List, func_name: str):
+    if func_name in ["LOCATE", "CHARINDEX"]:
+        return exp.StrPosition(
+            this=seq_get(args, 1),
+            substr=seq_get(args, 0),
+            position=seq_get(args, 2),
+        )
+    return exp.StrPosition.from_arg_list(args)
+
+
```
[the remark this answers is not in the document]
pruzko: I think this is wrong. The 1 0 2 order is only valid for "LOCATE" and "CHARINDEX", otherwise it's 0 1 2, a.k.a `from_arg_list`

sqlglot/dialects/dialect.py
```diff
@@ -1073,36 +1073,47 @@ def property_sql(self: Generator, expression: exp.Property) -> str:
...
+    func = exp.Anonymous(this=func_name, expressions=args)
+
+    if position and not supports_position:
+        func_with_offset = exp.Sub(this=func + position, expression=one)
+        func_wrapped = exp.If(
+            this=exp.EQ(this=func, expression=zero),
             true=zero,
-            false=strpos_sql + position_offset,
+            false=func_with_offset,
         )
```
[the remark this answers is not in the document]
pruzko: My initial code did this, but `Expression + Expression - Expression` adds unnecessary parentheses. `exp.Sub(this=func + position, expression=one)` is necessary to avoid this

sqlglot/dialects/dialect.py
```diff
@@ -1073,36 +1073,47 @@ def property_sql(self: Generator, expression: exp.Property) -> str:
...
+    substr = expression.args.get("substr")
+    position = expression.args.get("position")
+    occurrence = expression.args.get("occurrence")
+    zero = exp.Literal.number(0)
+    one = exp.Literal.number(1)
+
+    if supports_occurrence and occurrence and supports_position and not position:
+        position = one
+
+    if position and not supports_position:
```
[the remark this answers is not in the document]
pruzko: Careful, these are two different boolean expressions. In the first one I try to avoid invalid `INSTR(str, substr, None, occurrence)` by turning it into `INSTR(str, substr, 1, occurrence)`. In the second expression, I check whether the position argument is provided but not supported by the function, hence the need for mimicking using `SUBSTR`.

`supports_position and not position`
vs
`position and not supports_position`

sqlglot/dialects/tableau.py
pruzko: good catch

sqlglot/dialects/dialect.py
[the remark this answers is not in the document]
georgesittas: Ah I see, makes sense.

sqlglot/dialects/dialect.py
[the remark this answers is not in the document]
georgesittas: I think we're talking about different sections. I was referring to this one:

Screenshot 2025-01-22 at 1 05 00 PM

These two variables don't change in between. So more of a nit, but you could compute it once.

sqlglot/dialects/clickhouse.py
```diff
@@ -997,8 +998,8 @@ class Generator(generator.Generator):
             exp.RegexpLike: lambda self, e: self.func("match", e.this, e.expression),
             exp.Rand: rename_func("randCanonical"),
             exp.StartsWith: rename_func("startsWith"),
-            exp.StrPosition: lambda self, e: self.func(
-                "position", e.this, e.args.get("substr"), e.args.get("position")
+            exp.StrPosition: lambda self, e: strposition_sql(
+                self, e, func_name="LOCATE", supports_position=True
```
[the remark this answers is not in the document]
georgesittas: That's fair yeah

georgesittas (approved):
This looks good, I can do any final touches. Thanks again!

PR #4607, opened 2025-01-14
Chore: clean up ANALYZE implementation

georgesittas:


sqlglot/parser.py
```diff
@@ -7167,14 +7164,15 @@ def _parse_analyze(self) -> exp.Analyze | exp.Command:
             this = self._parse_table()
         # Try matching inner expr keywords before fallback to parse table.
         elif self._match_texts(self.ANALYZE_EXPRESSION_PARSERS):
+            kind = None
```
VaggelisD, line 7167: Minor but could we simplify the first few branches by initializing `kind = self._curr and self._curr.text.upper()` outside of it?

sqlglot/parser.py
```diff
@@ -7167,14 +7164,15 @@ def _parse_analyze(self) -> exp.Analyze | exp.Command:
...
         else:
             # Empty kind  https://prestodb.io/docs/current/sql/analyze.html
+            kind = None
             this = self._parse_table_parts()
 
-        try:
-            partition = self._parse_partition()
-        except ParseError:
+        partition = self._try_parse(self._parse_partition)
+        if not partition and self._match(TokenType.PARTITION):
```
VaggelisD, line 7175: Some of the missing Oracle partition syntax can also start with `SUBPARTITION` iirc, maybe we can change this to `self._match_set(("PARTITION", "SUBPARTITION"))`?

sqlglot/parser.py
```diff
@@ -7183,6 +7181,8 @@ def _parse_analyze(self) -> exp.Analyze | exp.Command:
         ):
             mode = f"WITH {self._next.text.upper()} MODE"
             self._advance(3)
```
geooo109, line 7183: if self._match_text_seq("WITH", "SYNC", "MODE") or self._match_text_seq(
            "WITH", "ASYNC", "MODE"
        ):
            mode = f"WITH {self._tokens[self._index-2].text.upper()} MODE"

VaggelisD (commented):
Thanks for the cleanup! A few points all over the place:
- Do we want the `self.raise_error`s at the end of `_parse_analyze_<columns/delete/list>`?
- Should we inline `usingdata_sql`, `analyzecolumns_sql` & `analyzewith_sql`?

sqlglot/parser.py
georgesittas: Can you share docs about this?

VaggelisD: Yep, it's the [partition_extension_clause](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/ANALYZE.html)

PR #4613, opened 2025-01-15
feat: performance improvement by lazy python

benfdking:
- make comments and texts getters lazy
- adds optimizations to `extract_string` of when to compute things

### DIFF
--- a/sqlglotrs/src/token.rs
+++ b/sqlglotrs/src/token.rs
@@ -1,5 +1,5 @@
 use crate::settings::TokenType;
-use pyo3::types::{PyNone};
+use pyo3::types::PyNone;
 use pyo3::{pyclass, pymethods, IntoPy, PyObject, Python};
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


benfdking:
While this makes the tokenizer much quicker, I think it would be good to benchmark slightly higher level to make sure we aren't seeing the slow down in other calls upstream.

tobymao:
can you run the regular benchmarks?

benfdking:
Struggling when running `python benchmarks/bench.py`, getting 

```
'Tokenizer' object has no attribute '_rs_dialect_settings'
'Tokenizer' object has no attribute '_rs_dialect_settings'
'Tokenizer' object has no attribute '_rs_dialect_settings'
'Tokenizer' object has no attribute '_rs_dialect_settings'
  File "/Users/benjaminking/Developer/sqlglot/benchmarks/bench.py", line 226, in <module>
    {k: v if v == "Query" else str(row[k])[0:7] + diff(row, k) for k, v in row.items()}
                                                  ^^^^^^^^^^^^
  File "/Users/benjaminking/Developer/sqlglot/benchmarks/bench.py", line 199, in diff
    return f" ({str(column / row['sqlglot'])[0:5]})"
                    ~~~~~~~^~~~~~~~~~~~~~~~
TypeError: ufunc 'divide' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Will have to be something for tomorrow.

PR #4615, opened 2025-01-15
Feat: Add support for REVERSE

pruzko:
`REVERSE(x)` reverses the string. Aside from SQLite, I am not aware of any engine that does not support it.

[Oracle](https://psoug.org/definition/reverse.htm)
[MySQL](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_reverse)
[Postgres](https://neon.tech/postgresql/postgresql-string-functions/postgresql-reverse)
[TSQL](https://learn.microsoft.com/en-us/sql/t-sql/functions/reverse-transact-sql?view=sql-server-ver16)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6455,6 +6455,10 @@ class Repeat(Func):
     arg_types = {"this": True, "times": True}
 
 
+class Reverse(Func):
+    pass
+
+
 # https://learn.microsoft.com/en-us/sql/t-sql/functions/round-transact-sql?view=sql-server-ver16
 # tsql third argument function == trunctaion if not 0
 class Round(Func):


VaggelisD:
Generally we'd want to introduce a new expression if there's a need to transpile between dialects with different semantics and/or args; For this case, it seems like we have `REVERSE(x)` across the board, so `exp.Anonymous` would be fine imo.

As a note, there are other dialects that support `REVERSE` like Snowflake, Spark/DB, BQ, Presto/Trino etc; If we introduced `exp.Reverse` without checking them first we could have broken their roundtrips, luckily they all accept a single input afaict.

With that in mind, the only use case I see for `exp.Reverse` would be for the type annotator.

georgesittas:
Agreed with Vaggelis, I don't see what this PR currently achieves. Closing it for now but happy to continue discussing.

pruzko:
I thought you're covering everything so that you can `ast.find(exp.Reverse)` but yeah, there is not point in adding this. Good to know tho.

PR #4616, opened 2025-01-15
Feat: Fix Oracle Integer Type Mapping

pruzko:
Currently, SQLGlot transpiles all integer types to `NUMBER`. This is not the correct way, as `NUMBER` (with no parameters) is a float. This messed up type castings such as `cast(1.23 as int)` that would be transpiled to `cast(1.23 as number)`.

I changed this to `SMALLINT` and `INT` and updated the tests. You can also consider using `NUMBER(3, 0)` for say `TINYINT`, though I think there will be some performance penalty.

See the [docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-0BC16006-32F1-42B1-B45E-F27A494963FF).

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -266,10 +266,10 @@ class Oracle(Dialect):
 
         TYPE_MAPPING = {
             **generator.Generator.TYPE_MAPPING,
-            exp.DataType.Type.TINYINT: "NUMBER",
-            exp.DataType.Type.SMALLINT: "NUMBER",
-            exp.DataType.Type.INT: "NUMBER",
-            exp.DataType.Type.BIGINT: "NUMBER",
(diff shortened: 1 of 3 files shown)


VaggelisD (commented):
Good catch!

I think we should preserve the semantics for `TINYINT` etc if we can through `NUMBER(x)`, performance considerations shouldn't trump that.

georgesittas (approved):

pruzko:
The `p` in `NUMBER(p[, d])` is the number of digits it needs to support, so `NUMBER(3)` would be enough to cover `TINYINT` I think. This however does not preserve the semantics as 999 could be stored as well. The only benefit to this I see is that it would require fewer bytes to store, but at the cost of more complicated arithmetic. `SHORTINT / INT` is the way to go I think.

PR #4622, opened 2025-01-15
feat(optimizer): Fix qualify for SEMI/ANTI joins

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3557

The RHS table of an SEMI/ANTI join shouldn't be collected as a selected source (in contrast to normal joins) since it's only used for the filtering process under the hood.

Note that as a side-effect of this, the `_expand_using` rule for SEMI/ANTI joins is turned off since we'd now fail to resolve the RHS related columns if they're expanded.

PS: I believe a different solution would be to run `transforms.py::eliminate_semi_and_anti_joins` before the optimizer

### DIFF
--- a/sqlglot/optimizer/scope.py
+++ b/sqlglot/optimizer/scope.py
@@ -311,6 +311,12 @@ class Scope:
             result = {}
 
             for name, node in self.references:
+                parent = node.parent
+                if isinstance(parent, exp.Join) and parent.is_semi_or_anti_join:
+                    # The RHS table of SEMI/ANTI joins shouldn't be collected as a
+                    # selected source
+                    continue
+
                 if name in result:
                     raise OptimizeError(f"Alias already used: {name}")
                 if name in self.sources:
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2453,6 +2453,10 @@ class Join(Expression):
     def alias_or_name(self) -> str:
         return self.this.alias_or_name
 
+    @property
+    def is_semi_or_anti_join(self) -> bool:
+        return self.kind in ("SEMI", "ANTI")
+
     def on(
         self,
         *expressions: t.Optional[ExpOrStr],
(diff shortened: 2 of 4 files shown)


georgesittas:
We still want to transform the `USING` clause into `ON`, in order to canonicalize those queries. Isn't the only problem here that we're including columns qualified with `"r"` in the expanded star?

From the linked issue, this is what we produce:

```sql
SELECT
  COALESCE("l"."id", "r"."id") AS "id",
  "l"."val_l" AS "val_l",
  "r"."val_r" AS "val_r"
FROM "l" AS "l"
SEMI JOIN "r" AS "r"
  ON "l"."id" = "r"."id"
```

But as Bill pointed out, if you produce this instead, it works (verified it locally):

```sql
SELECT
  "l"."id" AS "id",
  "l"."val_l" AS "val_l"
FROM "l" AS "l"
SEMI JOIN "r" AS "r"
  ON "l"."id" = "r"."id"
```

That means you need to treat `SEMI` / `ANTI` joins specially in the star expansion logic to avoid introducing `"r".<column>` projections. Notice that having `"r"."id"` in the join condition doesn't cause any issues.

VaggelisD:
I unblocked the `USING` expansion by adding a new map on `Scope` which collects the SEMI/ANTI join tables as "pseudo" selected sources in order to by pass `validate_qualify_columns`.

sqlglot/optimizer/scope.py
```diff
@@ -387,6 +401,12 @@ def pivots(self):
 
         return self._pivots
 
+    @property
+    def semi_or_anti_join_tables(self):
+        if self._semi_anti_join_tables is None:
+            return []
+        return self._semi_anti_join_tables
```
georgesittas, line 408:
```suggestion
    @property
    def semi_or_anti_join_tables(self) -> t.Set[str]:
        return self._semi_anti_join_tables or set()
```

sqlglot/optimizer/scope.py
```diff
@@ -126,6 +127,7 @@ def _collect(self):
         self._raw_columns = []
         self._stars = []
         self._join_hints = []
+        self._semi_anti_join_tables = []
```
georgesittas, line 130: I think we should probably make this a set, it's only used for lookups.

tobymao (approved):

PR #4624, opened 2025-01-15
feat(clickhouse): Dynamic data type

pkit:
see https://clickhouse.com/docs/en/sql-reference/data-types/dynamic

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -447,6 +447,17 @@ class ClickHouse(Dialect):
 
         FUNCTION_PARSERS.pop("MATCH")
 
+        PROPERTY_PARSERS = {
+            **parser.Parser.PROPERTY_PARSERS,
+        }
+
+        PROPERTY_PARSERS.pop("DYNAMIC")
+
+        ENUM_TYPE_TOKENS = {
+            *parser.Parser.ENUM_TYPE_TOKENS,
+            TokenType.DYNAMIC,
+        }
+
         NO_PAREN_FUNCTION_PARSERS = parser.Parser.NO_PAREN_FUNCTION_PARSERS.copy()
         NO_PAREN_FUNCTION_PARSERS.pop("ANY")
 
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -447,6 +447,17 @@ class Parser(parser.Parser):
 
         FUNCTION_PARSERS.pop("MATCH")
 
+        PROPERTY_PARSERS = {
+            **parser.Parser.PROPERTY_PARSERS,
+        }
+
+        PROPERTY_PARSERS.pop("DYNAMIC")
```
georgesittas, line 454:
```suggestion
        PROPERTY_PARSERS = parser.Parser.PROPERTY_PARSERS.copy()
        PROPERTY_PARSERS.pop("DYNAMIC")
```

sqlglot/parser.py
```diff
@@ -394,6 +394,7 @@ class Parser(metaclass=_Parser):
         TokenType.NULL,
         TokenType.NAME,
         TokenType.TDIGEST,
+        TokenType.DYNAMIC,
```
georgesittas, line 397: I think we should only map `"DYNAMIC"` to `TokenType.DYNAMIC` in Clickhouse, so this entry isn't actually needed. The token will be included due to `ENUM_TYPE_TOKENS`.
```suggestion
```

sqlglot/tokens.py
```diff
@@ -950,6 +951,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "USER-DEFINED": TokenType.USERDEFINED,
         "FOR VERSION": TokenType.VERSION_SNAPSHOT,
         "FOR TIMESTAMP": TokenType.TIMESTAMP_SNAPSHOT,
+        "DYNAMIC": TokenType.DYNAMIC,
```
georgesittas, line 954:
```suggestion
```

georgesittas (commented):
Thanks @pkit, I'll get this in and apply a couple of fixups.

PR #4628, opened 2025-01-16
fix(trino): support QUOTES option for JSON_QUERY #4623

### LINKED ISSUE #4623: Unable to parse `KEEP QUOTES`/`OMIT QUOTES` in Trino
Found an [issue](https://github.com/apache/superset/issues/31768) when trying to parse a Trino query that uses `JSON_QUERY(col, jsonpath OMIT QUOTES)`:

```python
import sqlglot

# https://trino.io/docs/current/functions/json.html#id6
sql = """
SELECT
      id,
      json_query(description, 'strict $.comment' KEEP QUOTES) AS quoted_comment,
      json_query(description, 'strict $.comment' OMIT QUOTES) AS unquoted_comment
FROM customers
"""

sqlglot.parse_one(sql, read="trino")
```

Will raise:

```python
...
  File "/Users/beto/Projects/sqlglot/sqlglot/parser.py", line 5817, in _parse_function_call
    self._match_r_paren(this)
  File "/Users/beto/Projects/sqlglot/sqlglot/parser.py", line 7930, in _match_r_paren
  File "/Users/beto/Projects/sqlglot/sqlglot/parser.py", line 1662, in raise_error
    raise error
sqlglot.errors.ParseError: Expecting ). Line 4, Col: 48.
```

geooo109:
Fixes #4623

This PR adds support for the QUOTES option of the JSON_QUERY.

Example:
```
parse_one("JSON_QUERY(description, 'strict $.comment' OMIT QUOTES ON SCALAR STRING)", dialect="trino")
JSONExtract(
  this=Column(
    this=Identifier(this=description, quoted=False)),
  expression=Literal(this=strict $.comment, is_string=True),
  json_query=True,
  quote=JSONExtractQuote(option=OMIT, scalar=True))
```

**DOCS**
[Trino QUOTES option](https://trino.io/docs/current/functions/json.html#json-query)

### DIFF
--- a/sqlglot/dialects/trino.py
+++ b/sqlglot/dialects/trino.py
@@ -4,6 +4,7 @@ from sqlglot import exp, parser
 from sqlglot.dialects.dialect import merge_without_target_sql, trim_sql, timestrtotime_sql
 from sqlglot.dialects.presto import Presto
 from sqlglot.tokens import TokenType
+import typing as t
 
 
 class Trino(Presto):
@@ -33,13 +34,29 @@ class Trino(Presto):
             ),
         }
 
-        def _parse_json_query(self):
+        def _parse_json_query_quote(self) -> t.Optional[exp.JSONExtractQuote]:
+            option = None
+            if self._match_text_seq("KEEP", "QUOTES") or self._match_text_seq("OMIT", "QUOTES"):
+                option = self._tokens[self._index - 2].text.upper()
+
+            scalar = "ON SCALAR STRING" if self._match_text_seq("ON", "SCALAR", "STRING") else None
+
+            if option:
+                return self.expression(
+                    exp.JSONExtractQuote,
+                    option=option,
+                    scalar=scalar,
+                )
+            return None
+
+        def _parse_json_query(self) -> exp.JSONExtract:
             return self.expression(
                 exp.JSONExtract,
                 this=self._parse_bitwise(),
                 expression=self._match(TokenType.COMMA) and self._parse_bitwise(),
                 option=self._parse_var_from_options(self.JSON_QUERY_OPTIONS, raise_unmatched=False),
                 json_query=True,
+                quote=self._parse_json_query_quote(),
             )
 
     class Generator(Presto.Generator):
@@ -60,6 +77,11 @@ class Trino(Presto):
             exp.JSONPathSubscript,
         }
 
+        def jsonextractquote_sql(self, expression: exp.JSONExtractQuote) -> str:
+            scalar = expression.args.get("scalar")
+            scalar = f" {scalar}" if scalar else ""
+            return f"{expression.args.get('option')} QUOTES" + scalar
+
         def jsonextract_sql(self, expression: exp.JSONExtract) -> str:
             if not expression.args.get("json_query"):
                 return super().jsonextract_sql(expression)
@@ -68,7 +90,10 @@ class Trino(Presto):
             option = self.sql(expression, "option")
             option = f" {option}" if option else ""
 
-            return self.func("JSON_QUERY", expression.this, json_path + option)
+            quote = self.sql(expression, "quote")
+            quote = f" {quote}" if quote else ""
+
+            return self.func("JSON_QUERY", expression.this, json_path + option + quote)
 
         def groupconcat_sql(self, expression: exp.GroupConcat) -> str:
             this = expression.this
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/trino.py
```diff
@@ -33,13 +34,29 @@ class Parser(Presto.Parser):
             ),
         }
 
-        def _parse_json_query(self):
+        def _parse_json_query_quote(self) -> t.Optional[exp.JSONExtractQuote]:
+            option = None
+            if self._match_text_seq("KEEP", "QUOTES") or self._match_text_seq("OMIT", "QUOTES"):
+                option = self._tokens[self._index - 2].text.upper()
+
+            scalar = "ON SCALAR STRING" if self._match_text_seq("ON", "SCALAR", "STRING") else None
```
VaggelisD: This can be just `scalar =  self._match_text_seq("ON", "SCALAR", "STRING") or None` since it's the only value for scalar, the `"ON SCALAR STRING"` can be moved to the generator

sqlglot/dialects/trino.py
```diff
@@ -33,13 +34,29 @@ class Parser(Presto.Parser):
             ),
         }
 
-        def _parse_json_query(self):
+        def _parse_json_query_quote(self) -> t.Optional[exp.JSONExtractQuote]:
+            option = None
+            if self._match_text_seq("KEEP", "QUOTES") or self._match_text_seq("OMIT", "QUOTES"):
+                option = self._tokens[self._index - 2].text.upper()
```
VaggelisD, line 40: We can flip the order of operations and return `None` early if we haven't matched these, in order to simplify the rest of the control flow:

```Python3
        def _parse_json_query_quote(self) -> t.Optional[exp.JSONExtractQuote]:
            if not (self._match_text_seq("KEEP", "QUOTES") or self._match_text_seq("OMIT", "QUOTES")):
                return None

            option = self._tokens[self._index - 2].text.upper()
            ...
```

sqlglot/dialects/trino.py
```diff
@@ -59,6 +76,11 @@ class Generator(Presto.Generator):
             exp.JSONPathSubscript,
         }
 
+        def jsonextractquote_sql(self, expression: exp.JSONExtractQuote) -> str:
+            scalar = expression.args.get("scalar")
+            scalar = f" {scalar}" if scalar else ""
+            return f"{expression.args.get('option')} QUOTES" + scalar
```
VaggelisD, line 83: scalar can go inside the f-string

VaggelisD (commented):
Nice! A few minor comments:

sqlglot/dialects/trino.py
```diff
@@ -59,6 +77,11 @@ class Generator(Presto.Generator):
             exp.JSONPathSubscript,
         }
 
+        def jsonextractquote_sql(self, expression: exp.JSONExtractQuote) -> str:
```
georgesittas, line 80: Move this to the base generator class, otherwise doing `expression.sql()` will crash.

sqlglot/dialects/trino.py
georgesittas: Ditto about using `self.sql` to generate `option`.

sqlglot/dialects/trino.py
geooo109: @georgesittas If we use this logic for `scalar` value, we should use `expression.args.get("scalar")` in the generator. Instead of using `self.sql(expression, "scalar")` as you proposed, right?

geooo109: @georgesittas If we use this approach for the `scalar` value (True/False/None), we can't use the approach you suggested `self.sql(expression, "scalar")` for getting its value in the generator right? Instead, we should use the `expression.args.get("scalar")` ?

sqlglot/generator.py
```diff
@@ -4421,6 +4421,10 @@ def oncondition_sql(self, expression: exp.OnCondition) -> str:
 
         return f"{empty}{error}{null}"
 
+    def jsonextractquote_sql(self, expression: exp.JSONExtractQuote) -> str:
+        scalar = " ON SCALAR STRING" if expression.args.get("scalar") else ""
+        return f"{self.sql(expression,'option')} QUOTES{scalar}"
```
VaggelisD, line 4426:
```suggestion
        return f"{self.sql(expression, 'option')} QUOTES{scalar}"
```

VaggelisD (approved):

PR #4638, opened 2025-01-20
feat: improve rs tokenizer performance

benfdking:


sqlglotrs/src/tokenizer.rs
```diff
@@ -654,54 +655,56 @@ impl<'a> TokenizerState<'a> {
...
-                    if delimiter.len() > 1 {
-                        self.advance((delimiter.len() - 1) as isize)?;
-                    }
-                    break;
-                }
-                if self.is_end {
-                    if !raise_unmatched {
+                let peek_char_str = self.peek_char.to_string();
+                if peek_char_str == delimiter || escapes.contains(&self.peek_char) {
+                    if peek_char_str == delimiter {
```
georgesittas, line 666: Small one: what about setting `let peek_matches_delimiter = peek_char_str == delimiter` and then using that instead?

georgesittas (approved):

sqlglotrs/src/tokenizer.rs
```diff
@@ -654,54 +654,57 @@ impl<'a> TokenizerState<'a> {
...
-                    if delimiter.len() > 1 {
-                        self.advance((delimiter.len() - 1) as isize)?;
-                    }
-                    break;
-                }
-                if self.is_end {
-                    if !raise_unmatched {
+                let peek_char_str = self.peek_char.to_string();
+                let equal_delimiter = delimiter == peek_char_str;
+                if peek_char_str == delimiter || escapes.contains(&self.peek_char) {
```
georgesittas, line 665:
```suggestion
                if equal_delimiter || escapes.contains(&self.peek_char) {
```

PR #4645, opened 2025-01-21
fix(optimizer)!: Do not remove parens on bracketed expressions

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3672

Afaict it's only Postgres that cannot parse inline bracketed expressions such as the following:

- Before this PR
```Python3
>>> simplify(parse_one("SELECT (ARRAY_AGG(name))[1]")).sql()
'SELECT ARRAY_AGG(name)[1]'
```
```SQL
postgres> WITH sample_data AS (SELECT 'foo' AS name) SELECT ARRAY_AGG(name)[1] FROM sample_data;
ERROR:  syntax error at or near "["
LINE 1: ...a AS (SELECT 'foo' AS name) SELECT ARRAY_AGG(name)[1] FROM s...
```

- After
```Python3
>>> simplify(parse_one("SELECT (ARRAY_AGG(name))[1]")).sql()
'SELECT (ARRAY_AGG(name))[1]'
```
```SQL
postgres> WITH sample_data AS (SELECT 'foo' AS name) SELECT (ARRAY_AGG(name))[1] FROM sample_data;
 array_agg
-----------
 foo
(1 row)
```

### DIFF
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -749,7 +749,7 @@ def simplify_parens(expression):
 
     if (
         not isinstance(this, exp.Select)
-        and not isinstance(parent, exp.SubqueryPredicate)
+        and not isinstance(parent, (exp.SubqueryPredicate, exp.Bracket))
         and (
             not isinstance(parent, (exp.Condition, exp.Binary))
             or isinstance(parent, exp.Paren)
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
Thanks! LGTM.

georgesittas:
FYI this is breaking because it changes the AST.

PR #4652, opened 2025-01-22
fix(postgres): Support WITHIN GROUP ( order_by_clause ) FILTER for Postgres

### LINKED ISSUE #4651: WITHIN GROUP ( order_by_clause ) FILTER breaks for Postgres
Read / convert dialect is Postgres.

```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 62.
  SELECT PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY a) FILTER (WHERE CAST(b AS BOOLEAN)) AS mean_value FROM (VALUES (0, 't')) AS fake_data(a, b)
```

**Official Documentation**
https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES

gl3nnleblanc:
Fixes #4651

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -411,9 +411,6 @@ class Dialect(metaclass=_Dialect):
     ARRAY_AGG_INCLUDES_NULLS: t.Optional[bool] = True
     """Whether ArrayAgg needs to filter NULL values."""
 
-    SUPPORTS_WITHIN_GROUP_FILTER = True
-    """Whether WITHIN GROUP ( order_by ) FILTER syntax is supported for aggregate functions."""
-
     PROMOTE_TO_INFERRED_DATETIME_TYPE = False
     """
     This flag is used in the optimizer's canonicalize rule and determines whether x will be promoted
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -411,6 +411,9 @@ class Dialect(metaclass=_Dialect):
     ARRAY_AGG_INCLUDES_NULLS: t.Optional[bool] = True
     """Whether ArrayAgg needs to filter NULL values."""
 
+    SUPPORTS_WITHIN_GROUP_FILTER = False
+    """Whether WITHIN GROUP ( order_by ) FILTER syntax is supported for aggregate functions."""
+
     PROMOTE_TO_INFERRED_DATETIME_TYPE = False
     """
     This flag is used in the optimizer's canonicalize rule and determines whether x will be promoted
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -411,7 +411,7 @@ class Dialect(metaclass=_Dialect):
     ARRAY_AGG_INCLUDES_NULLS: t.Optional[bool] = True
     """Whether ArrayAgg needs to filter NULL values."""
 
-    SUPPORTS_WITHIN_GROUP_FILTER = False
+    SUPPORTS_WITHIN_GROUP_FILTER = True
     """Whether WITHIN GROUP ( order_by ) FILTER syntax is supported for aggregate functions."""
 
     PROMOTE_TO_INFERRED_DATETIME_TYPE = False
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2725,7 +2725,7 @@ class Generator(metaclass=_Generator):
         this = self.sql(expression, "this")
         expression_sql = self.sql(expression, "expression")[1:]  # order has a leading space
         filter_sql = self.sql(expression, "filter")[1:]
-        if filter_sql:
+        if filter_sql and self.dialect.SUPPORTS_WITHIN_GROUP_FILTER:
             return f"{this} WITHIN GROUP ({expression_sql}) FILTER ({filter_sql})"
         return f"{this} WITHIN GROUP ({expression_sql})"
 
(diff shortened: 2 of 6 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -411,6 +411,9 @@ class Dialect(metaclass=_Dialect):
     ARRAY_AGG_INCLUDES_NULLS: t.Optional[bool] = True
     """Whether ArrayAgg needs to filter NULL values."""
 
+    SUPPORTS_WITHIN_GROUP_FILTER = False
+    """Whether WITHIN GROUP ( order_by ) FILTER syntax is supported for aggregate functions."""
+
```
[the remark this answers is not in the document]
gl3nnleblanc: I think this combined WITHIN GROUP + FILTER syntax is actually limited to Postgres, so defaulting to True is probably not desired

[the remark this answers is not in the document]
georgesittas: When we're adding new features the default is to enable it for all dialects, and opt out in any of that have clashing syntax or something similar. To achieve this we move all implementations to the base parser / generator classes, so the base dialect ends up becoming somewhat of a "superset" dialect.

georgesittas:
Hey @gl3nnleblanc, you'll need to run `make style` and then commit styling changes / fix mypy issues for the CI to pass.

sqlglot/dialects/dialect.py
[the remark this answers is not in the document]
gl3nnleblanc: Ah gotcha, thanks for clarifying so quick!

sqlglot/dialects/dialect.py
```diff
@@ -411,6 +411,9 @@ class Dialect(metaclass=_Dialect):
     ARRAY_AGG_INCLUDES_NULLS: t.Optional[bool] = True
     """Whether ArrayAgg needs to filter NULL values."""
 
+    SUPPORTS_WITHIN_GROUP_FILTER = True
+    """Whether WITHIN GROUP ( order_by ) FILTER syntax is supported for aggregate functions."""
+
```
georgesittas, line 416: This is basically dead code right now, we can add it on demand later if ever needed.
```suggestion
    SUPPORTS_WITHIN_GROUP_FILTER = True
    """Whether WITHIN GROUP ( order_by ) FILTER syntax is supported for aggregate functions."""

```

sqlglot/generator.py
```diff
@@ -2724,6 +2724,9 @@ def windowspec_sql(self, expression: exp.WindowSpec) -> str:
     def withingroup_sql(self, expression: exp.WithinGroup) -> str:
         this = self.sql(expression, "this")
         expression_sql = self.sql(expression, "expression")[1:]  # order has a leading space
+        filter_sql = self.sql(expression, "filter")[1:]
+        if filter_sql and self.dialect.SUPPORTS_WITHIN_GROUP_FILTER:
```
georgesittas, line 2728:
```suggestion
        if filter_sql:
```

sqlglot/parser.py
```diff
@@ -6603,7 +6603,13 @@ def _parse_window(
         # https://learn.microsoft.com/en-us/sql/t-sql/functions/percentile-disc-transact-sql?view=sql-server-ver16
         if self._match_text_seq("WITHIN", "GROUP"):
             order = self._parse_wrapped(self._parse_order)
-            this = self.expression(exp.WithinGroup, this=this, expression=order)
+            if self.dialect.SUPPORTS_WITHIN_GROUP_FILTER and self._match(TokenType.FILTER):
```
georgesittas, line 6606:
```suggestion
            if self._match(TokenType.FILTER):
```

sqlglot/generator.py
```diff
@@ -2724,6 +2724,9 @@ def windowspec_sql(self, expression: exp.WindowSpec) -> str:
...
+            return f"{this} WITHIN GROUP ({expression_sql}) FILTER ({filter_sql})"
         return f"{this} WITHIN GROUP ({expression_sql})"
```
georgesittas, line 2730: Let's instead do something like:

```python
result = f"{this} WITHIN GROUP ({expression_sql})"
if filter_sql:
    result = f"{result} FILTER ({filter_sql})"
return result

File: sqlglot/parser.py [hunk 6]
```diff
@@ -6603,7 +6603,13 @@ def _parse_window(
...
+                where_clause = self._parse_wrapped(self._parse_where)
+                this = self.expression(
+                    exp.WithinGroup, this=this, expression=order, filter=where_clause
+                )
+            else:
+                this = self.expression(exp.WithinGroup, this=this, expression=order)
```
georgesittas, line 6612: What about reusing `_parse_window` for this purpose? I think we may be able to get rid of the new filter `arg` as well that way and simplify the PR quite a bit.

gl3nnleblanc: I like this and it certainly does make these couple lines cleaner, but having some issues reconciling the approach with the fact we explicitly wrap `expression` with parentheses inside [generator.py](https://github.com/tobymao/sqlglot/blob/v26.2.1/sqlglot/generator.py#L2727). If I touch that part of the generator then the amount of required changes seems to blow up. Should I be adding the parsed filter clause to something other than exp.WithinGroup?

georgesittas: Ah, I see. That's ok, we can leave it out of this PR then. I'll merge and take a look to see if I can easily do this as a refactor step.

georgesittas (approved):
Thanks for the contribution!

georgesittas:
@gl3nnleblanc take a look at https://github.com/tobymao/sqlglot/commit/284a9360c5d43301da34d8d5199f101423ade289, I ended up simplifying this quite a lot.

Issue #4655, opened 2025-01-23
CURRENT_SCHEMA Transpilation


pruzko:
Currently, SQLGlot does not have an abstraction for `CURRENT_SCHEMA` and alike. These functions return the string representation of the current schema.

I would like to cover the following engines.

SQLite: `'main'` ([Doc](https://www.sqlite.org/lang_naming.html))
TSQL: `SCHEMA_NAME([schema_id])` ([Doc](https://learn.microsoft.com/en-us/sql/t-sql/functions/schema-name-transact-sql?view=sql-server-ver16))
MySQL: `SCHEMA()` or `DATABASE()` ([Doc](https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_schema))
Postgres: `CURRENT_SCHEMA[()]` ([Doc](https://www.postgresql.org/docs/current/functions-info.html))

Docs for other engines:
Clickhouse: `CURRENT_DATABASE()` ([Doc](https://clickhouse.com/docs/en/sql-reference/functions/other-functions#currentdatabase))
Databricks: `CURRENT_SCHEMA()` ([Doc](https://docs.databricks.com/en/sql/language-manual/functions/current_schema.html))
DuckDB: `CURRENT_SCHEMA()` ([Doc](https://duckdb.org/docs/sql/functions/utility.html#current_schema))
Hive: maybe `CURRENT_DATABASE()` ([SO](https://stackoverflow.com/a/77953269), [Tutorial](https://www.tutorialscampus.com/hive/use-database.htm), though I didn't find it in the docs)
Snowflake: `CURRENT_SCHEMA()` ([Doc](https://docs.snowflake.com/en/sql-reference/functions/current_schema))
Spark: `CURRENT_SCHEMA()` ([Doc](https://spark.apache.org/docs/latest/api/sql/index.html#current_schema))
Oracle: `SYS_CONTEXT('USERENV', 'CURRENT_SCHEMA')` ([Doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SYS_CONTEXT.html))
Trino: `CURRENT_SCHEMA` ([Doc](https://trino.io/docs/current/functions/session.html))

Unsupported:
BigQuery
Presto (though `CURRENT_USER` seems to be [supported](https://prestodb.io/docs/current/functions/session.html))

georgesittas:
FYI, @pruzko and I chatted in Slack and he's planning to tackle this.

[...]

PR #4658, opened 2025-01-24
Feat!: CurrentSchema for SQLite, MySQL, Postgres, and TSQL

pruzko:
Implements #4655. The docs are in the issue.


sqlglot/dialects/sqlite.py
```diff
@@ -179,6 +179,7 @@ class Generator(generator.Generator):
             exp.CurrentDate: lambda *_: "CURRENT_DATE",
             exp.CurrentTime: lambda *_: "CURRENT_TIME",
             exp.CurrentTimestamp: lambda *_: "CURRENT_TIMESTAMP",
+            exp.CurrentSchema: lambda *_: "'main'",
```
VaggelisD, line 182: Is this correct? Can't we have a `temp` database or any other name here?

sqlglot/dialects/tsql.py
```diff
@@ -575,6 +575,7 @@ class Parser(parser.Parser):
             "RIGHT": _build_with_arg_as_text(exp.Right),
             "PARSENAME": _build_parsename,
             "REPLICATE": exp.Repeat.from_arg_list,
+            "SCHEMA_NAME": exp.CurrentSchema.from_arg_list,
```
VaggelisD, line 578: T-SQL can accept the `schema_id` arg, by using `rename_func()` in all the generators we'd transpile that to other dialects.

Can we add tests for `SCHEMA_NAME(x)` too?

VaggelisD (commented):
Hey @pruzko, as always thank you for the PR! 

Check out `NO_PAREN_FUNCTIONS` for similar work, dialects like Postgres also allow `CURRENT_SCHEMA` (without parens) so we could parse these too.

sqlglot/expressions.py
```diff
@@ -5628,6 +5628,10 @@ class CurrentTimestamp(Func):
     arg_types = {"this": False, "sysdate": False}
 
 
+class CurrentSchema(Func):
+    arg_types = {"this": False}
```
georgesittas, line 5632: Let's do a sanity check and see if `CURRENT_SCHEMA` is present in other dialects and if it expects >1 arguments.

sqlglot/dialects/sqlite.py
pruzko: SQLite does not have a concept of the current schema. There is only the schema that refers to the DB that was opened first, i.e., `'main'`. It's always `'main'` and you can't change that AFAIK. Other databases can be attached and referred to by their name and there is `'temp'`, but `select x from temp.table` will never be equivalent to `select x from table` and, on the contrary, `select x from main.table` always is.

> "Name of the database that contains the table (always 'main')"
https://www2.sqlite.org/cvstrac/wiki?p=InformationSchema

sqlglot/expressions.py
pruzko: As @VaggelisD said, TSQL supports an additional argument `schema_id`. I'll fix this. Also I updated the Issue to cover more docs for future work.

sqlglot/expressions.py
```diff
@@ -5628,6 +5628,10 @@ class CurrentTimestamp(Func):
     arg_types = {"this": False, "sysdate": False}
 
 
+class CurrentSchema(Func):
+    arg_types = {"this": False, "id": False}
```
georgesittas, line 5632: It seems like there's only ever one argument for this function. Can we revert to `pass` and use `this` in place of `id`? Or is there a dialect that supports a two-argument variant of `CURRENT_SCHEMA`?

georgesittas (approved):
~I'll just take this to the finish line, thanks @pruzko!~

After discussing with Vaggelis, it looks like he spotted an issue (see below message), so I'm gonna hold off merging this for a bit longer.

VaggelisD (commented):
I'm not sure if this approach is sound, it looks like it biases MySQL in which `SCHEMA` == `DATABASE`, but this is not the case for the other dialects right?

- In SQLite, `'main'` is the default name for the `TABLE_CATALOG`/`DATABASE`, which is different from the `TABLE_SCHEMA` being `'sqlite'`
- In Postgres (and other dialects), `DATABASE` == `CATALOG` which contain `SCHEMA`s

So, transpiling `exp.CurrentSchema` from MySQL -> Postgres we'd end up with different semantics, right?

These terms are overloaded in SQL so we should make sure the representation is as generic as possible; I think a better approach would be to rename this to `CURRENT_CATALOG` or `CURRENT_DATABASE` for this purpose.

sqlglot/expressions.py
pruzko: I thought `id` might be more explicit, but yes, let's use `this`.

pruzko:
Here are my 2 cents on SQLite and `'main'`.

SQLite does not define schemas, catalogues, or anything like that. It also does not define databases like other engines do. Schemas, catalogues, and databases could be seen as namespaces.

Databases in SQLite correspond to `.sqlite` files. Once the file is attached to SQLite process, a namespace is created for it. The default namespace is called `main`. For instance, if you opened `db.sqlite`, you can refer to all its tables as `select t` but also `select main.t`. If you later attach `backup.sqlite`, you can refer to its tables via `select backup.t`. In SQLite's terminology, these namespaces are referred to as "database" but also "schema".
```
> select * from pragma_database_list();
seq | name | file
------|-------|-------------
0     | main  | .../db.sqlite
1     | bckp  | .../bckp.sqlite

> select * from pragma_table_list();
schema | name | ...
-------|------|---------
main   | t    | ...
```

Now as you pointed out, schemas/catalogues/dbs are a mess. I see two options to deal with this - we can look for similar naming or similar behavior. I use `exp.CurrentSchema` as a filter when dumping table names within the default namespace, so I am in favor of prioritizing behavior. That said, I can easily handle this in my code and you might have different opinion.

Also, I'm afraid I caused some confusion with the [previous link](https://www2.sqlite.org/cvstrac/wiki?p=InformationSchema). The link only explains how to define a view that mimics `INFORMATION_SCHEMA` even though it is not the SQLIte standard. It does mention `'sqlite'` as the `TABLE_SCHEMA`, but as pointed out in the `pragma_table_list` example, this seems to be ignored.


> So, transpiling exp.CurrentSchema from MySQL -> Postgres we'd end up with different semantics, right?
I'm not sure what you mean here. MySQL's `SCHEMA()` and `DATABASE()` are the same and they are what Postgres refers to as `CURRENT_SCHEMA`. These queries are equivalent AFAIK:
```
SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE table_schema=SCHEMA();
SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE table_schema=DATABASE();

SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE table_schema=CURRENT_SCHEMA;
```

VaggelisD:
I see, this cleared things up, my main reservation had to do with SQLite but it's resolved after your explanation. Thank you!

georgesittas (changes requested):
As @VaggelisD said, there are no concerns regarding the schema / database mixing anymore on our side. What you did _should_ be backwards-compatible. On the other hand, transpiling Postgres' `CURRENT_DATABASE()` to MySQL doesn't work today anyway and is not affected by this PR.

So, summarizing existing suggestions and adding one more comment, these are the outstanding items for this PR

1. Let's get rid of `id` (see other thread) and use `this` instead. Make sure to simplify the `FUNCTIONS` entry for T-SQL and fix `unsupported_args` decorations.
2. Please move the `CURRENT_SCHEMA` token from the base `NO_PAREN_FUNCTIONS` to Postgres, because it is not a "no-parentheses function" in other dialects. E.g., MySQL treats `CURRENT_SCHEMA` as a column reference, and we need to parse it as such.
3. After (2) is applied, make sure to update `Parser.ID_VAR_TOKENS` so it contains `TokenType.CURRENT_SCHEMA`.

PR #4659, opened 2025-01-27
Fix(trino): Correctly handle exp.LocationProperty

erindru:
Prior to this change, `exp.LocationProperty` was parsed correctly but stripped entirely when generating SQL.

The reason for this is that the Trino dialect extends the Presto dialect and it was marked as `exp.Properties.Location.UNSUPPORTED` in the Presto dialect.

However, it is supported in Trino, in both [schemas](https://trino.io/docs/current/sql/create-schema.html#examples) and tables.

I've updated it to be `exp.Properties.Location.POST_WITH` and also treat it as a normal property instead of a naked property because it needs to be generated as `key=value` and not `key value`.

Before:
```
>>> parse_one("CREATE SCHEMA foo WITH (location = 's3://bucket')", dialect="trino").sql(dialect="trino")
Unsupported property locationproperty
'CREATE SCHEMA foo'
```

After:
```
>>> parse_one("CREATE SCHEMA foo WITH (location = 's3://bucket')", dialect="trino").sql(dialect="trino")
"CREATE SCHEMA foo WITH (LOCATION='s3://bucket')"
```

It's a similar story for `CREATE TABLE`. Before:
```
>>> parse_one("CREATE TABLE foo (id INT) WITH (location = 's3://bucket')", dialect="trino").sql(dialect="trino")
Unsupported property locationproperty
'CREATE TABLE foo (id INTEGER)'
```

After:
```
>>> parse_one("CREATE TABLE foo (id INT) WITH (location = 's3://bucket')", dialect="trino").sql(dialect="trino")
"CREATE TABLE foo (id INTEGER) WITH (LOCATION='s3://bucket')"
```

It's only worked by accident so far in SQLMesh because if `location` is specified as part of `physical_properties` on a model, it's created as an `exp.Property` instead of an `exp.LocationProperty` and there is no issue with `exp.Property`.

### DIFF
--- a/sqlglot/dialects/trino.py
+++ b/sqlglot/dialects/trino.py
@@ -57,11 +57,17 @@ class Trino(Presto):
             )
 
     class Generator(Presto.Generator):
+        PROPERTIES_LOCATION = {
+            **Presto.Generator.PROPERTIES_LOCATION,
+            exp.LocationProperty: exp.Properties.Location.POST_WITH,
+        }
+
         TRANSFORMS = {
             **Presto.Generator.TRANSFORMS,
             exp.ArraySum: lambda self,
             e: f"REDUCE({self.sql(e, 'this')}, 0, (acc, x) -> acc + x, acc -> acc)",
             exp.ArrayUniqueAgg: lambda self, e: f"ARRAY_AGG(DISTINCT {self.sql(e, 'this')})",
+            exp.LocationProperty: lambda self, e: self.property_sql(e),
             exp.Merge: merge_without_target_sql,
             exp.TimeStrToTime: lambda self, e: timestrtotime_sql(self, e, include_precision=True),
             exp.Trim: trim_sql,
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/trino.py
```diff
@@ -57,11 +57,17 @@ def _parse_json_query(self) -> exp.JSONExtract:
             )
 
     class Generator(Presto.Generator):
+        PROPERTIES_LOCATION = {
+            **Presto.Generator.PROPERTIES_LOCATION,
+            exp.LocationProperty: exp.Properties.Location.POST_WITH,
```
erindru, line 62: I changed this here instead of at the Presto level because I cant see a mention of supporting the `location` property in the [Presto docs](https://prestodb.io/docs/current/sql/create-schema.html), which I guess is why it was deliberately marked as unsupported to begin with

VaggelisD (approved):

sqlglot/dialects/trino.py
VaggelisD: I think there is `external_location` e.g https://github.com/prestodb/presto/issues/15621 but it's indeed not documented afaict.

georgesittas (approved):

Issue #4662, opened 2025-01-27
ClickHouse Map literal not serialized correctly


martijnthe:
I'm using ClickHouse v24.12.3 and `sqlglot` 26.3.8.

Certain `SETTINGS` in ClickHouse require a Map data type ([see for example,`additional_table_filters`](https://clickhouse.com/docs/en/operations/settings/settings#additional_table_filters)), but written using the syntax `{'key1': value1, 'key2': value2}`. The `map(...)` function cannot be used in this context, see https://github.com/ClickHouse/ClickHouse/issues/75065

`sqlglot` parses the `{...}` syntax into a `exp.VarMap`, but when calling `sql()`, it's converted into a `map(...)` function call. For example:

Parsing:
```
>>> e = sqlglot.parse_one("SELECT COUNT(1) FROM table SETTINGS additional_table_filters = map('table', 'tenant_id = 123')", dialect="clickhouse")
>>> e
Select(
  expressions=[
    Count(
      this=Literal(this=1, is_string=False),
      big_int=True)],
  from=From(
    this=Table(
      this=Identifier(this=table, quoted=False))),
  settings=[
    EQ(
      this=Column(
        this=Identifier(this=additional_table_filters, quoted=False)),
      expression=VarMap(
        keys=Array(
          expressions=[
            Literal(this=table, is_string=True)]),
        values=Array(
          expressions=[
            Literal(this=tenant_id = 123, is_string=True)])))])
```

Serializing:
```
>>> e.sql(dialect="clickhouse")
"SELECT COUNT(1) FROM table SETTINGS additional_table_filters = map('table', 'tenant_id = 123')"
```

ClickHouse fails to execute the output SQL, while the input SQL works just fine:
```
7fe4f271034f :) SELECT COUNT(1) FROM table SETTINGS additional_table_filters = map('table', 'tenant_id = 123')

Syntax error: failed at position 95 (end of query):

SELECT COUNT(1) FROM table SETTINGS additional_table_filters = map('table', 'tenant_id = 123')

Expected one of: token, OpeningRoundBracket, FILTER, RESPECT NULLS, IGNORE NULLS, OVER, literal or map, literal, NULL, NULL, number, Bool, TRUE, FALSE, string literal, OpeningCurlyBrace
```

JFYI, support for parsing `{...}` had been added [here](https://github.com/tobymao/sqlglot/commit/d18a8712e088f87c969214b61a05944d279e5703).

[...]

PR #4664, opened 2025-01-28
fix(clickhouse): Generate bracket notation for exp.VarMap

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/4662

As shown in the issue, the function-based representation doesn't work under `SETTINGS`:

```SQL

:) WITH table AS (SELECT 1) SELECT COUNT(1) FROM table SETTINGS additional_table_filters = map('table', 'tenant_id = 123');

Syntax error: failed at position 120 (end of query):

Expected one of: token, OpeningRoundBracket, FILTER, RESPECT NULLS, IGNORE NULLS, OVER, literal or map, literal, NULL, NULL, number, Bool, TRUE, FALSE, string literal, OpeningCurlyBrace
```

But the bracket notation does:

```
:) WITH table AS (SELECT 1) SELECT COUNT(1) FROM table SETTINGS additional_table_filters = {'table': 'tenant_id = 123'};

   ┌─COUNT(1)─┐
1. │        1 │
   └──────────┘
```

<br/>
<br/>

This PR applies a bandaid solution for this issue in which `MAP(...)` is maintained _except_ the `SETTINGS` case, in which we revert back to the bracket notation.

Docs
-------
[Clickhouse MAP](https://clickhouse.com/docs/en/sql-reference/data-types/map)

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -164,8 +164,8 @@ def _map_sql(self: ClickHouse.Generator, expression: exp.Map | exp.VarMap) -> st
     if not (expression.parent and expression.parent.arg_key == "settings"):
         return _lower_func(var_map_sql(self, expression))
 
-    keys = expression.args["keys"]
-    values = expression.args["values"]
+    keys = expression.args.get("keys")
+    values = expression.args.get("values")
 
     if not isinstance(keys, exp.Array) or not isinstance(values, exp.Array):
         self.unsupported("Cannot convert array columns into map.")
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -160,6 +160,26 @@ def _timestrtotime_sql(self: ClickHouse.Generator, expression: exp.TimeStrToTime
     return self.sql(exp.cast(ts, datatype, dialect=self.dialect))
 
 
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -160,6 +160,26 @@ def _timestrtotime_sql(self: ClickHouse.Generator, expression: exp.TimeStrToTime
     return self.sql(exp.cast(ts, datatype, dialect=self.dialect))
 
 
+def _map_sql(self: ClickHouse.Generator, expression: exp.Map | exp.VarMap) -> str:
+    if not (expression.parent and expression.parent.arg_key == "settings"):
+        return _lower_func(var_map_sql(self, expression))
+
+    keys = expression.args["keys"]
+    values = expression.args["values"]
```
georgesittas, line 168:
```suggestion
    keys = expression.args.get("keys")
    values = expression.args.get("values")
```

georgesittas (approved):

sqlglot/dialects/clickhouse.py
VaggelisD: Should we also correct that on `dialect.py::var_map_sql`?

VaggelisD: Dumb question to ask

PR #4666, opened 2025-01-28
Feat: add mssql and sqlserver as aliases of the TSQL dialect

georgesittas:


barakalon:
What's the reason here?

georgesittas:
> What's the reason here?

Seems like it's common for folk to use MSSQL, T-SQL and SQL Server interchangeably to refer to the same dialect (relevant [SO thread](https://stackoverflow.com/a/15637791)). In SQLMesh we refer to the engine as [MSSQL](https://sqlmesh.readthedocs.io/en/stable/integrations/engines/mssql/?h=mssql) but since the dialect here is called TSQL, you can't do stuff like `sqlmesh init mssql`.

You could argue that the dialect is `tsql` though, so this isn't _necessarily_ needed. Just a quality-of-life type of thing. Thought I'd just put this up for discussion, happy to close if y' all disagree.

VaggelisD (approved):
LGTM, I also think it's a nice-to-have UX feature e.g Postgres could also have `psql` as an alias etc, up to Toby & Barak to decide

barakalon:
The dialect names are unique identifiers. 
If this is a UX concern for sqlmesh, maybe sqlmesh should handle normalizing dialect names?

barakalon:
I just worry this adds as much complexity/confusion as it does convenience.
And it doesn't really solve the discovery problem for sqlglot users, since the [list of dialects](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/__init__.py) doesn't include mssql or sqlserver. So users still need to know about tsql? I.e., I don't think we support users just _guessing_ the names of dialects in sqlglot haha

georgesittas:
I'm persuaded 👍
@tobymao agrees this shouldn't go in here as well, fwiw.

Thanks for taking a look!

Issue #4670, opened 2025-01-29
BigQuery type-annotated array literals don't survive roundtrip through SQLGlot


sean-rose:
BigQuery type-annotated array literals like `ARRAY<FLOAT64>[1, 2, 3]` get converted into invalid `CAST()` statements when parsed and re-output by SQLGlot.  The original type-annotation syntax needs to be preserved, because apparently once BigQuery reads a non-type-annotated array literal it can't be cast to a different type.

**Fully reproducible code snippet**
```python
import sqlglot
ast = sqlglot.parse_one("SELECT ARRAY<FLOAT64>[1, 2, 3] AS floats", dialect="bigquery")
sql = ast.sql(dialect="bigquery")
print(sql)
```
That prints `SELECT CAST([1, 2, 3] AS ARRAY<FLOAT64>) AS floats`, and if you try to run that query in BigQuery you get the error `Casting between arrays with incompatible element types is not supported: Invalid cast from ARRAY<INT64> to ARRAY<FLOAT64> at [1:13]`.

**Official Documentation**
* https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#using_array_literals

PR #4671, opened 2025-01-29
fix(bigquery)!: Inline type-annotated ARRAY literals

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/4670

https://github.com/tobymao/sqlglot/pull/3751 canonicalized BigQuery's inlined constructors into casts which may not be correct for certain roundtrips, as shown in the issue:

- Safe
```SQL
bq> SELECT ARRAY<INT64>[1, 2, 3], CAST([1, 2, 3] AS ARRAY<INT64>);
f0_	f1_
"[1,2,3]"	"[1,2,3]"
```

<br/>

- Not safe (type mismatch)
```SQL
bq> SELECT CAST([1, 2, 3] AS ARRAY<FLOAT64>);
Casting between arrays with incompatible element types is not supported: Invalid cast from ARRAY<INT64> to ARRAY<FLOAT64> at [1:13]
```

<br/>
<br/>

This PR fixes this particular scenario by preserving the inlined `ARRAY` type annotation & literal value for BQ's roundtrip.

Docs
-------
[BQ ARRAY Literals](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#using_array_literals)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -1229,3 +1229,11 @@ class BigQuery(Dialect):
                 expr = expr.this
 
             return self.func("CONTAINS_SUBSTR", this, expr)
+
+        def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
+            this = expression.this
+
+            if isinstance(this, exp.Array):
+                return f"{self.sql(expression, 'to')}{self.sql(this)}"
+
+            return super().cast_sql(expression, safe_prefix=safe_prefix)
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -2035,10 +2035,10 @@ OPTIONS (
         )
 
         self.validate_all(
-            "SELECT ARRAY<INT>[1, 2, 3]",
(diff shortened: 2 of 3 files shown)


georgesittas:
Does this only concern arrays? What about structs?

georgesittas:
Cool, no this is what I had in mind.

georgesittas (approved):

sean-rose:
> Yep, I tried `STRUCT` related queries in a similar fashion but I don't think the same typing issue arises, e.g this is fine:

While casting `STRUCT`s like that doesn't result in an outright BigQuery error, IMO it'd be preferable to also preserve BigQuery's typed struct literal syntax since it's more idiomatic, succinct, and avoids BigQuery potentially doing more work to first construct the`STRUCT` and then convert it (I'm speculating on this last point).

georgesittas:
Makes sense. @VaggelisD can we try to roundtrip these constructors in BigQuery?

VaggelisD:
Yep, will give it a shot.

VaggelisD:
Here's an attempt https://github.com/tobymao/sqlglot/pull/4684

PR #4677, opened 2025-01-29
fix(duckdb): support postgres JSON/JSONB_OBJECT_AGG to duckdb JSON_GROUP_OBJECT

### LINKED ISSUE #4667: Bug in JSONB_OBJECT_AGG translation from postgres to duckdb
```python
from sqlglot import parse_one
json_query = 'SELECT JSONB_OBJECT_AGG(col1, col2) from table'
print(parse_one(json_query, dialect='postgres').sql(dialect='duckdb'))
``` 
The output is:
'SELECT JSONB_OBJECT_AGG(col1, col2) FROM "table"'

while the expect output should be:
'SELECT JSON_GROUP_OBJECT(col1, col2) FROM "table"'

**Official Documentation**
1. duckdb docs - https://duckdb.org/docs/data/json/json_functions#json-aggregate-functions
2. postgresql docs - https://www.postgresql.org/docs/9.5/functions-aggregate.html

geooo109:
Fixes #4667

`duckdb` only supports the `JSON_GROUP_OBJECT` aggregate function (`JSONB` isn't supported), thus for both 
`JSON/JSONB_OBJECT_AGG` cases the mapping is the `JSON_GROUP_OBJECT` function.

Moreover, the roundtrip for the postgres `JSON/JSONB_OBJECT_AGG` is preserved.

**DOCS**
[DuckDB JSON_GROUP_OBJECT](https://duckdb.org/docs/data/json/json_functions#json-aggregate-functions) | [Postgres  JSON/JSONB_OBJECT_AGG ](https://www.postgresql.org/docs/9.5/functions-aggregate.html)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -393,6 +393,10 @@ class Postgres(Dialect):
             "SHA384": lambda args: exp.SHA2(this=seq_get(args, 0), length=exp.Literal.number(384)),
             "SHA512": lambda args: exp.SHA2(this=seq_get(args, 0), length=exp.Literal.number(512)),
             "LEVENSHTEIN_LESS_EQUAL": _build_levenshtein_less_equal,
+            "JSON_OBJECT_AGG": lambda args: exp.JSONObjectAgg(
+                expressions=[seq_get(args, 0), seq_get(args, 1)]
+            ),
+            "JSONB_OBJECT_AGG": exp.JSONBObjectAgg.from_arg_list,
         }
 
         NO_PAREN_FUNCTIONS = {
@@ -617,6 +621,8 @@ class Postgres(Dialect):
             exp.Unicode: rename_func("ASCII"),
             exp.UnixToTime: _unix_to_time_sql,
             exp.Levenshtein: _levenshtein_sql,
+            exp.JSONObjectAgg: rename_func("JSON_OBJECT_AGG"),
+            exp.JSONBObjectAgg: rename_func("JSONB_OBJECT_AGG"),
         }
 
         TRANSFORMS.pop(exp.CommentColumnConstraint)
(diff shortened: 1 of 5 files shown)


tests/dialects/test_postgres.py
```diff
@@ -838,6 +838,9 @@ def test_postgres(self):
             "SELECT PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY a) FILTER(WHERE CAST(b AS BOOLEAN)) AS mean_value FROM (VALUES (0, 't')) AS fake_data(a, b)"
         )
 
+        self.validate_identity("SELECT JSON_OBJECT_AGG(k, v) FROM t")
+        self.validate_identity("SELECT JSONB_OBJECT_AGG(k, v) FROM t")
```
VaggelisD, line 842: We can remove the DuckDB tests and instead make this `validate_all` here:

```
self.validate_all(
 "SELECT JSON_OBJECT_AGG(k, v)",
 write={
    "postgres": "SELECT JSON_OBJECT_AGG(k, v)",
    "duckdb":  "SELECT JSON_GROUP_OBJECT(k, v) FROM t",
  }

```

Same for the `JSONB` variant

sqlglot/dialects/postgres.py
```diff
@@ -393,6 +393,10 @@ class Parser(parser.Parser):
             "SHA384": lambda args: exp.SHA2(this=seq_get(args, 0), length=exp.Literal.number(384)),
             "SHA512": lambda args: exp.SHA2(this=seq_get(args, 0), length=exp.Literal.number(512)),
             "LEVENSHTEIN_LESS_EQUAL": _build_levenshtein_less_equal,
+            "JSON_OBJECT_AGG": lambda args: exp.JSONObjectAgg(
+                expressions=[seq_get(args, 0), seq_get(args, 1)]
```
VaggelisD, line 397: We could do `expressions=args` here

sqlglot/expressions.py
```diff
@@ -6068,6 +6068,14 @@ class JSONObjectAgg(AggFunc):
     }
 
 
+# https://www.postgresql.org/docs/9.5/functions-aggregate.html
+class JSONBObjectAgg(AggFunc):
+    arg_types = {
+        "key": True,
+        "value": True,
+    }
```
VaggelisD, line 6076: Let's use `this` and `expression` to follow the rest of the codebase. Also, were these multi-lined by `make style`? It should fit in a single line if I'm not mistaken

geooo109: The extra `,` next to `"value": True` is the cause for the extra line, removed it.

georgesittas (approved):

Issue #4668, opened 2025-01-28
Bug in DATE_BIN translation from postgres to duckdb


dor-bernstein:
**Fully reproducible code snippet**
```python
from sqlglot import parse_one
query = "SELECT DATE_BIN('30 days, 0:00:00', events.timestamp, (SELECT MIN(events.timestamp) AS min_1 FROM events)) AS bin FROM table"
print(parse_one(query, dialect='postgres').sql(dialect='duckdb'))
``` 
The output is: 
SELECT DATE_BIN('30 days, 0:00:00', events.timestamp, (SELECT MIN(events.timestamp) AS min_1 FROM events)) AS bin FROM "table"

while the expected output should be:
SELECT TIME_BUCKET('2592000000 milliseconds', events.timestamp, (SELECT MIN(events.timestamp) AS min_1 FROM events)) AS bin FROM "table"
as duckdb doesn't know to handle intervals in this structure, but it does know to handle milliseconds intervals

**Official Documentation**
1. duckdb docs - https://duckdb.org/docs/sql/functions/timestamp.html#time_bucketbucket_width-timestamp-origin
2. postgres docs - https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-BIN

tobymao:
@dor-bernstein these are simply unimplemented, feel free to make a PR for these translation issues, but we won't tackle them all due to prioritization from our side.

PR #4681, opened 2025-01-29
Feat!: transpile postgres DATE_BIN function to duckdb TIME_BUCKET

dor-bernstein:
Closes this [issue](https://github.com/tobymao/sqlglot/issues/4668)

Will be happy for review.

Thanks!

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -662,6 +662,7 @@ class DuckDB(Dialect):
             exp.Levenshtein: unsupported_args("ins_cost", "del_cost", "sub_cost", "max_dist")(
                 rename_func("LEVENSHTEIN")
             ),
+            exp.DateBin: rename_func("TIME_BUCKET")
         }
 
         SUPPORTED_JSON_PATH_PARTS = {
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -662,7 +662,7 @@ class DuckDB(Dialect):
             exp.Levenshtein: unsupported_args("ins_cost", "del_cost", "sub_cost", "max_dist")(
                 rename_func("LEVENSHTEIN")
             ),
-            exp.DateBin: rename_func("TIME_BUCKET")
+            exp.DateBin: rename_func("TIME_BUCKET"),
         }
 
         SUPPORTED_JSON_PATH_PARTS = {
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -617,6 +617,7 @@ class Generator(generator.Generator):
             exp.Unicode: rename_func("ASCII"),
             exp.UnixToTime: _unix_to_time_sql,
             exp.Levenshtein: _levenshtein_sql,
+            exp.DateBin: rename_func("DATE_BIN"),
```
georgesittas, line 624: This is redundant, since `DateBin` is a `Func` subclass, it automatically gets generated as `DATE_BIN`.
```suggestion
```

sqlglot/dialects/duckdb.py
```diff
@@ -662,6 +662,7 @@ class Generator(generator.Generator):
             exp.Levenshtein: unsupported_args("ins_cost", "del_cost", "sub_cost", "max_dist")(
                 rename_func("LEVENSHTEIN")
             ),
+            exp.DateBin: rename_func("TIME_BUCKET"),
```
georgesittas, line 667: What about parsing DuckDB's `TIME_BUCKET` into `DateBin` too? You probably need an entry like `"TIME_BUCKET": exp.DateBin.from_arg_list` in `FUNCTIONS`.

sqlglot/expressions.py
```diff
@@ -5640,6 +5640,10 @@ class DateAdd(Func, IntervalOp):
     arg_types = {"this": True, "expression": True, "unit": False}
 
 
+class DateBin(Func, IntervalOp):
+    arg_types = {"this": True, "expression": True, "unit": False}
+
```
georgesittas, line 5645: You need to make sure this doesn't break other dialects, e.g. clickhouse [supports](https://clickhouse.com/docs/en/sql-reference/functions/date-time-functions#tostartofinterval) `DATE_BIN` with four arguments, so you need to add a `"zone"` arg at the end of this dict.

georgesittas (approved):
Thanks for addressing comments, and for the PR!

sqlglot/expressions.py
georgesittas: I'll do this as a follow-up, but make sure to test stuff like this for future PRs.

PR #4684, opened 2025-01-30
fix(bigquery): Improve roundtrip of typed STRUCTs

VaggelisD:
Context https://github.com/tobymao/sqlglot/pull/4671

- Before this PR:
```Python3
>>> sqlglot.parse_one("SELECT STRUCT<x INT64, y STRING>(1, 'bar')", dialect="bigquery").sql("bigquery")
"SELECT CAST(STRUCT(1, 'bar') AS STRUCT<x INT64, y STRING>)"
```

<br/>

- After this PR:
```Python3
>>> sqlglot.parse_one("SELECT STRUCT<x INT64, y STRING>(1, 'bar')", dialect="bigquery").sql("bigquery")
"SELECT STRUCT<x INT64, y STRING>(1, 'bar')"
```

cc: @sean-rose

Docs
-----------
[BQ STRUCT](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -1236,4 +1236,10 @@ class BigQuery(Dialect):
             if isinstance(this, exp.Array):
                 return f"{self.sql(expression, 'to')}{self.sql(this)}"
 
+            if isinstance(this, exp.Struct) and not this.find(exp.PropertyEQ):
+                # We can inline the cast into a typed STRUCT if the fields aren't named, e.g:
+                # Fine: CAST(STRUCT(1) AS STRUCT<INT64>) --> STRUCT<INT64>(1)
+                # Not fine: CAST(STRUCT(1 AS foo) AS STRUCT<bar INT64>) -/-> STRUCT<bar INT64>(1 AS foo)
+                return f"{self.sql(expression, 'to')}({self.expressions(this)})"
+
             return super().cast_sql(expression, safe_prefix=safe_prefix)
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -931,13 +931,6 @@ LANGUAGE js AS
             },
         )
         self.validate_all(
-            "SELECT CAST(STRUCT(1) AS STRUCT<INT64>)",
-            write={
-                "bigquery": "SELECT CAST(STRUCT(1) AS STRUCT<INT64>)",
-                "snowflake": "SELECT CAST(OBJECT_CONSTRUCT('_0', 1) AS OBJECT)",
-            },
-        )
-        self.validate_all(
             "cast(x as date format 'MM/DD/YYYY')",
             write={
                 "bigquery": "PARSE_DATE('%m/%d/%Y', x)",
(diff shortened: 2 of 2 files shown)


VaggelisD:
I think I'll actually go ahead and close this for now. The issue is that we canonicalize typed STRUCTs into cast, but during generation we can't deduce if the same cast was a user vs a canonicalized one. Notice how it's not always safe to transform these casts back into the typed versions, e.g:

- This cast is fine
```SQL
bq> SELECT CAST(STRUCT(1 AS old_name) AS STRUCT<new_name INT64>) AS strct;
strct.new_name
1
```

<br/>

- This inlined construction is not
```SQL
bq> SELECT STRUCT<new_name INT64>(1 AS old_name) AS strct;

Error: STRUCT constructors cannot specify both an explicit type and field names with AS at [1:36]
```

<br/>

This is the reason I went ahead and added the `this.find(exp.PropertyEQ)` call, but this can add performance hits for users with wide and/or deeply nested STRUCTs since we may traverse the entire sub-AST.

VaggelisD:
PS: For future reference, it seems that typed structs will error only if the top-level fields are named, otherwise it's fine:

```SQL
bq> SELECT STRUCT<test INT64, bar STRUCT<foo INT64>>(1, STRUCT(2 AS baz));
strct
"{
  ""strct"": {
    ""test"": ""1"",
    ""bar"": {
      ""foo"": ""2""
    }
  }
}"

```

This means that we can make the Generator check a linear scan, but that still incurs hits for wide STRUCTs:

```Python3
+++ b/sqlglot/dialects/bigquery.py
@@ -1236,7 +1236,7 @@ class BigQuery(Dialect):
             if isinstance(this, exp.Array):
                 return f"{self.sql(expression, 'to')}{self.sql(this)}"
 
-            if isinstance(this, exp.Struct) and not this.find(exp.PropertyEQ):
+            if isinstance(this, exp.Struct) and not any(isinstance(expr, exp.PropertyEQ) for expr in this.expressions):
```

sean-rose:
> The issue is that we canonicalize typed STRUCTs into cast, but during generation we can't deduce if the same cast was a user vs a canonicalized one.

That canonicalization change made in #3751 is losing information, which is problematic since as you've demonstrated the two forms of struct definition aren't always equivalent in BigQuery.

Could we perhaps change BigQuery parsing to add some sort of annotation to such canonicalized `Cast` expressions to indicate whether it's a normal cast or a type-annotation (e.g. `kind`)?  Then that could be used by BigQuery SQL generation to preserve such type-annotated structs during a roundtrip through SQLGlot.

sean-rose:
> @sean-rose can you point out how converting the `STRUCT<...>(...)` into a cast is losing information? As long as the semantics are preserved, this should be fine from SQLGlot's standpoint. I'm in favor of preserving the original syntax, but in this case I think the added complexity is somewhat questionable. Happy to be persuaded otherwise if this conversion is erroneous for whatever reason.

It's losing the information about whether the original form was a literal `CAST()` or a type-annotated `STRUCT<...>()` constructor.  Since it appears BigQuery doesn't balk at casting structs like it does at casting arrays this isn't a dealbreaker, but IMO being able to preserve the original syntax as much as possible during a roundtrip through SQLGlot is worth some added complexity.  For example, I've been considering using SQLGlot to replace some existing SQL formatting logic, but I wouldn't want it to arbitrarily change idiomatic BigQuery SQL like `STRUCT<...>(...)` into `CAST(... AS STRUCT<...>)`.

georgesittas:
I hear you, however I think for this particular case we'll leave it as is, since BigQuery works with either form. I'd suggest overriding the relevant parts of the dialect if this is unacceptable, e.g. if implementing a formatting tool like you mentioned.

PR #4687, opened 2025-01-31
Fix!: load dialects lazily

georgesittas:
### Overview

When sqlglot is imported for the first time, all dialects are eagerly imported. The time needed for this is small, but not necessarily insignificant. This PR makes the dialect loading & registering lazy.

<details>
<summary> Benchmark results for <b>make check</b> & import time (BEFORE) </summary>

```
(.venv) ➜  sqlglot git:(main) make check
pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
mypy.....................................................................Passed
SQLGLOTRS_TOKENIZER=0 python -m unittest
.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 777 tests in 74.419s

OK
RUST_BACKTRACE=1 python -m unittest
.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 777 tests in 80.686s

OK
```

```python
>>> import timeit
>>> import numpy as np
>>>
>>>
>>> def import_sqlglot():
...     import sqlglot
...
>>>
>>> print(np.mean(timeit.repeat(import_sqlglot, number=1)))
0.03040810823440552
```

</details>

<details>
<summary> Benchmark results for <b>make check</b> & import time (AFTER) </summary>

```
(.venv) ➜  sqlglot git:(main) ✗ make check
pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
mypy.....................................................................Passed
SQLGLOTRS_TOKENIZER=0 python -m unittest
.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 777 tests in 76.272s

OK
RUST_BACKTRACE=1 python -m unittest
.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 777 tests in 77.444s

OK
```

```python
>>> import timeit
>>> import numpy as np
>>>
>>>
>>> def import_sqlglot():
...     import sqlglot
...
>>>
>>> print(np.mean(timeit.repeat(import_sqlglot, number=1)))
0.008051850087940692
```

</details>

**TL;DR**: importing sqlglot for the first time seems to be ~4 times faster given this refactor.

### Notes

- Users could previously import a dialect class from `sqlglot.dialects`. This is still possible, only now it's done lazily. For this reason, I believe the changes in `sqlglot/dialects/__init__.py` are backwards-compatible.
- The changes in `_Dialect.get` and `_Dialect.__getitem__` should also be backwards-compatible, because if the given key does not exist in the registry, we simply try to load the class first before looking it up again.
- Until now, `Dialect.classes` would return a dict with **ALL** dialects. In order to preserve this behavior, I've renamed `classes` to `_classes` and overrode `_Dialect.__getattr__` in order to load all dialects on-demand, if needed. Basically, `_classes` now represents what's been loaded until now.
- For custom user dialects the situation is the same as before: they still need to import them in modules they're referenced, or, alternatively, at the top-level package so that they're registered as early as possible. Then, lookups should just work because they'll exist in `_classes`.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import importlib
 import logging
 import typing as t
 from enum import Enum, auto
 from functools import reduce
 
 from sqlglot import exp
+from sqlglot.dialects import DIALECT_MODULE_NAMES
 from sqlglot.errors import ParseError
 from sqlglot.generator import Generator, unsupported_args
 from sqlglot.helper import AutoName, flatten, is_int, seq_get, subclasses, to_bool
@@ -119,8 +121,6 @@ class _Dialect(type):
 
     def __getattr__(cls, name):
         if name == "classes":
-            from sqlglot.dialects import DIALECT_MODULE_NAMES
-
             # We want `<DialectSubclass>.classes` to return *all* available dialects
             if len(DIALECT_MODULE_NAMES) != len(cls._classes):
                 for key in DIALECT_MODULE_NAMES:
@@ -133,9 +133,6 @@ class _Dialect(type):
     @classmethod
     def _try_load(cls, key: str | Dialects) -> None:
         try:
-            import importlib
-            from sqlglot.dialects import DIALECT_MODULE_NAMES
-
             if isinstance(key, Dialects):
                 key = key.value
 
@@ -827,7 +824,6 @@ class Dialect(metaclass=_Dialect):
             result = cls.get(dialect_name.strip())
             if not result:
                 from difflib import get_close_matches
-                from sqlglot.dialects import DIALECT_MODULE_NAMES
 
                 close_matches = get_close_matches(dialect_name, list(DIALECT_MODULE_NAMES), n=1)
 
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -119,16 +119,13 @@ class _Dialect(type):
     def __hash__(cls) -> int:
         return hash(cls.__name__.lower())
 
-    def __getattr__(cls, name):
-        if name == "classes":
-            # We want `<DialectSubclass>.classes` to return *all* available dialects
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlglot/dialects/dialect.py
```diff
@@ -117,20 +119,50 @@ def __eq__(cls, other: t.Any) -> bool:
     def __hash__(cls) -> int:
         return hash(cls.__name__.lower())
 
+    def __getattr__(cls, name):
```
barakalon, line 122: Should this be a `@property`? Or is there some metaclass complexity?

barakalon (approved):

sqlglot/dialects/dialect.py
georgesittas: That's what I tried first, but when you do `metaclass.some_property`, it returns a `property` object instead of the actual value... Python 🤷‍♂️

```python
>>> class Meta(type):
...     @property
...     def description(cls):
...         return f"This is a metaclass named {cls.__name__}"
...
>>> class MyClass(metaclass=Meta):
...     pass
...
>>> print(Meta.description)  # This is a metaclass named Meta
<property object at 0x104877100>
>>>
>>> print(MyClass.description)
This is a metaclass named MyClass
```

barakalon: 🤷 I can't imagine anyone is referencing the metaclass directly? They definitely _shouldn't_.

But all this metaclass stuff is messy anyways, so I'm happy with whatever

georgesittas: Ack, changed my mind and will refactor in favor of `property`. Thanks for poking here.

PR #4690, opened 2025-02-01
fix(optimizer.merge_subqueries): consider original table names as take

xiaoyu-meng-mxy:
In merge_subqueries, we do not consider the original table names in the outer_scope as taken, which may cause some issue in the below conner case

`WITH cte_t AS (SELECT a, x.b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT cte_t.a, y.c FROM cte_t LEFT JOIN z as y on cte_t.a = y.c` where the optimizer does not think `z` is taken in the outer_scope. When we _merge_joins, we will add `JOIN y AS z` to the outer_scope, which actually changed the meaning of z.

Without the fix, the above query will have `SELECT x.a AS a, y.c AS c FROM x AS x LEFT JOIN y AS z ON x.a <= z.b LEFT JOIN z AS y ON x.a = y.c` as the output of merge_subqueries, which seems wrong.

### DIFF
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -227,7 +227,11 @@ def _rename_inner_sources(outer_scope: Scope, inner_scope: Scope, alias: str) ->
     """
     inner_taken = set(inner_scope.selected_sources)
 
-    outer_taken = set(outer_scope.selected_sources) | set(node.name for node, _ in outer_scope.selected_sources.values() if isinstance(node, exp.Table))
+    outer_taken = set(outer_scope.selected_sources) | set(
+        node.name
+        for node, _ in outer_scope.selected_sources.values()
+        if isinstance(node, exp.Table)
+    )
 
     conflicts = outer_taken.intersection(inner_taken)
     conflicts -= {alias}
--- a/sqlglot/optimizer/merge_subqueries.py
+++ b/sqlglot/optimizer/merge_subqueries.py
@@ -226,7 +226,9 @@ def _rename_inner_sources(outer_scope: Scope, inner_scope: Scope, alias: str) ->
     Renames any sources in the inner query that conflict with names in the outer query.
     """
     inner_taken = set(inner_scope.selected_sources)
-    outer_taken = set(outer_scope.selected_sources)
+
+    outer_taken = set(outer_scope.selected_sources) | set(node.name for node, _ in outer_scope.selected_sources.values() if isinstance(node, exp.Table))
+
     conflicts = outer_taken.intersection(inner_taken)
     conflicts -= {alias}
 
--- a/tests/fixtures/optimizer/merge_subqueries.sql
+++ b/tests/fixtures/optimizer/merge_subqueries.sql
@@ -95,7 +95,7 @@ SELECT x.a AS a, x.b AS b FROM x AS x;
 
 # title: CTE with conflicting table name
 # execute: false
-WITH CTE AS (SELECT a, b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT a, b FROM CTE LEFT JOIN z as y;
+WITH cte_t AS (SELECT a, b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT a, b FROM cte_t LEFT JOIN z as y;
 SELECT x.a AS a, b AS b FROM x AS x LEFT JOIN y AS z_2 ON x.a <= z_2.b LEFT JOIN z AS y;
 
 # title: Nested CTE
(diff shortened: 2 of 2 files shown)


xiaoyu-meng-mxy:
hmm there are other use cases like 
```
WITH z_2 AS (
  SELECT
    z.b AS b
  FROM y AS z
)
SELECT
  x.a AS a,
  y.c AS c
FROM x AS x
LEFT JOIN z AS y
  ON x.a = y.c
LEFT JOIN z_2 AS z
  ON x.a <= z.b
```
which will result in 
```
SELECT
  x.a AS a,
  y.c AS c
FROM x AS x
LEFT JOIN y AS z
  ON x.a <= z.b
LEFT JOIN z AS y
  ON x.a = y.c
```

Which can not be fixed by the approach in the current PR.

Should we consider this case as not mergeable?

georgesittas:
@xiaoyu-meng-mxy can you show an end-to-end example that demonstrates the issue in `merge_subqueries`? Am I understanding correctly that you're trying to use it in isolation?

georgesittas:
It seems like the original query is being converted correctly when using `optimize`:

```python
>>> optimize("WITH cte_t AS (SELECT x.a, x.b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT cte_t.a, y.c FROM cte_t LEFT JOIN z as y on cte_t.a = y.c").sql()
'SELECT "x"."a" AS "a", "y"."c" AS "c" FROM "x" AS "x" LEFT JOIN "y" AS "z" ON "x"."a" <= "z"."b" LEFT JOIN "z" AS "y" ON "x"."a" = "y"."c"'
```

I'll go ahead and close this one until more info is provided. Happy to reopen if needed.

xiaoyu-meng-mxy:
> It seems like the original query is being converted correctly when using `optimize`:
> 
> ```python
> >>> from sqlglot.optimizer import optimize
> >>> optimize("WITH cte_t AS (SELECT x.a, x.b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT cte_t.a, y.c FROM cte_t LEFT JOIN z as y on cte_t.a = y.c").sql()
> 'SELECT "x"."a" AS "a", "y"."c" AS "c" FROM "x" AS "x" LEFT JOIN "y" AS "z" ON "x"."a" <= "z"."b" LEFT JOIN "z" AS "y" ON "x"."a" = "y"."c"'
> ```
> 
> I'll go ahead and close this one until more info is provided. Happy to reopen if needed.

I was thinking, the result is not correct. 

If x,y,z are three different table, in the original query, it tries to join cte_t with z on cte_t.a = z.c, but after optimizer, `z` becomes `y` after `LEFT JOIN "y" AS "z" `, which seems wrong?

georgesittas:
Can you create a reproducible example with static data that demonstrates the difference in semantics? Something like:

```
D CREATE TABLE x AS SELECT 1 AS a, 2 AS b;
D CREATE TABLE y AS SELECT 1 AS a, 2 AS b;
D CREATE TABLE z AS SELECT 1 AS c;
D WITH cte_t AS (SELECT x.a, x.b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT cte_t.a, y.c FROM cte_t LEFT JOIN z as y on cte_t.a = y.c;
┌───────┬───────┐
│   a   │   c   │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     1 │
└───────┴───────┘
D SELECT "x"."a" AS "a", "y"."c" AS "c" FROM "x" AS "x" LEFT JOIN "y" AS "z" ON "x"."a" <= "z"."b" LEFT JOIN "z" AS "y" ON "x"."a" = "y"."c";
┌───────┬───────┐
│   a   │   c   │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     1 │
└───────┴───────┘
```

xiaoyu-meng-mxy:
> Can you create a reproducible example with static data that demonstrates the difference in semantics? Something like:
> 
> ```
> D CREATE TABLE x AS SELECT 1 AS a, 2 AS b;
> D CREATE TABLE y AS SELECT 1 AS a, 2 AS b;
> D CREATE TABLE z AS SELECT 1 AS c;
> D WITH cte_t AS (SELECT x.a, x.b FROM x LEFT JOIN y as z on x.a <= z.b) SELECT cte_t.a, y.c FROM cte_t LEFT JOIN z as y on cte_t.a = y.c;
> ┌───────┬───────┐
> │   a   │   c   │
> │ int32 │ int32 │
> ├───────┼───────┤
> │     1 │     1 │
> └───────┴───────┘
> D SELECT "x"."a" AS "a", "y"."c" AS "c" FROM "x" AS "x" LEFT JOIN "y" AS "z" ON "x"."a" <= "z"."b" LEFT JOIN "z" AS "y" ON "x"."a" = "y"."c";
> ┌───────┬───────┐
> │   a   │   c   │
> │ int32 │ int32 │
> ├───────┼───────┤
> │     1 │     1 │
> └───────┴───────┘
> ```

I see. I think I was making a mistake.
```LEFT JOIN "y" AS "z"``` and ```LEFT JOIN "z" AS "y" ``` are two different joins, so they are two different scope, which means although we give an alias `z` to `y` in the first left join, it does not mean in the second left join `z` is an alias, it still referring to table `z`.

PR #4695, opened 2025-02-03
fix(tsql): remove BEGIN from identifiers

geooo109:
`BEGIN` is a reserved keyword in TSQL, thus it can't be used as an identifier of any purpose.
This PR removes the usage of `BEGIN` for this occasion. 

Moreover, this PR removes the `INTERVAL_VARS` from the base `Parser` class because it isn't needed any more.  

**DOCS**
[T-SQL BEGIN](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/begin-end-transact-sql?view=sql-server-ver16#syntax)

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -453,6 +453,8 @@ class TestTSQL(Validator):
             },
         )
 
+        self.validate_identity("SELECT 'BEGIN'")
+
     def test_option(self):
         possible_options = [
             "HASH GROUP",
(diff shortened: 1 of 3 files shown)


tests/dialects/test_tsql.py
```diff
@@ -453,6 +453,8 @@ def test_tsql(self):
             },
         )
 
+        self.validate_identity("SELECT 'BEGIN'")
```
georgesittas, line 456: This test doesn't seem meaningful– it's not related to the changes in this PR. Consider checking that `SELECT begin` raises for T-SQL, instead.

georgesittas (approved):

tobymao (approved):

PR #4703, opened 2025-02-04
Fix(athena): add SHOW parsing

theborch:
Parse the following Athena statements as `SHOW`:

[SHOW COLUMNS](https://docs.aws.amazon.com/athena/latest/ug/show-columns.html)
[SHOW CREATE TABLE](https://docs.aws.amazon.com/athena/latest/ug/show-create-table.html)
[SHOW CREATE VIEW](https://docs.aws.amazon.com/athena/latest/ug/show-create-view.html)
[SHOW DATABASES](https://docs.aws.amazon.com/athena/latest/ug/show-databases.html)
[SHOW PARTITIONS](https://docs.aws.amazon.com/athena/latest/ug/show-partitions.html)
[SHOW TABLES](https://docs.aws.amazon.com/athena/latest/ug/show-tables.html)
[SHOW TBLPROPERTIES](https://docs.aws.amazon.com/athena/latest/ug/show-tblproperties.html)
[SHOW VIEWS](https://docs.aws.amazon.com/athena/latest/ug/show-views.html)

Example:

```py
from sqlglot import parse_one

parse_one("show create table db.table;", dialect="athena")
```

Output:

```py
Show(
  this=CREATE TABLE,
  target=Identifier(this=table, quoted=False),
  db=Identifier(this=db, quoted=False))
```

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -71,6 +71,13 @@ def _partitioned_by_property_sql(self: Athena.Generator, e: exp.PartitionedByPro
     return f"{prop_name}={self.sql(e, 'this')}"
 
 
+def _show_parser(*args: t.Any, **kwargs: t.Any) -> t.Callable[[Athena.Parser], exp.Show]:
+    def _parse(self: Athena.Parser) -> exp.Show:
+        return self._parse_show_athena(*args, **kwargs)
+
+    return _parse
+
+
 class Athena(Trino):
     """
     Over the years, it looks like AWS has taken various execution engines, bolted on AWS-specific modifications and then
@@ -114,6 +121,8 @@ class Athena(Trino):
             "UNLOAD": TokenType.COMMAND,
         }
 
+        COMMANDS = Trino.Tokenizer.COMMANDS - {TokenType.SHOW}
+
     class Parser(Trino.Parser):
         """
         Parse queries for the Athena Trino execution engine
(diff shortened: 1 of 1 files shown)


sqlglot/dialects/athena.py
```diff
@@ -122,8 +124,55 @@ class Parser(Trino.Parser):
...
+            "TBLPROPERTIES": _show_parser("TBLPROPERTIES", target=True),
+            "VIEWS": _show_parser("VIEWS", target="IN"),
+        }
+
+        def _parse_show_athena(
+            self,
+            this: str,
+            target: bool | str = False,
+            full: t.Optional[bool] = None,
+            global_: t.Optional[bool] = None,
```
VaggelisD, line 148: `global_` doesn't seem to be used anywhere

sqlglot/dialects/athena.py
```diff
@@ -122,8 +124,55 @@ class Parser(Trino.Parser):
...
+
+            db = None
+
+            if self._match(TokenType.FROM) or self._match(TokenType.IN):
+                db = self._parse_id_var()
+            elif self._match(TokenType.DOT):
+                db = target_id
+                target_id = self._parse_id_var()
+
+            like = self._parse_string() if self._match_text_seq("LIKE") else None
```
VaggelisD, line 165: This can be simplified to `like = self._match_text_seq("LIKE") and self._parse_string()`. Also, we can inline these initializations directly into the return statement to save some space

VaggelisD (commented):
Hey @theborch, thank you for this contribution! It adds lots of improvements.

One important step before fully diving into the details is that the additions here are not tested, could we add the relevant tests in `test_athena.py`?

theborch:
> Hey @theborch, thank you for this contribution! It adds lots of improvements.
> 
> One important step before fully diving into the details is that the additions here are not tested, could we add the relevant tests in `test_athena.py`?

Absolutely.

sqlglot/dialects/athena.py
theborch: Ah, good catch, leftover copypasta I didn't catch from reuse of mysql dialect show.

sqlglot/dialects/athena.py
theborch: Cool, didn't want to make too many style assumptions, but more than happy to do as much.

georgesittas:
Hey @theborch, any plans to take this to the finish line?

georgesittas:
I'll go ahead and close this due to inactivity for now, but feel free to reopen when you start iterating on this again.

PR #4704, opened 2025-02-04
fix(tsql): support default values on definitons and the OUTPUT/OUT/READ_ONLY syntax

geooo109:
This PR adds support:

1. For default initialization of both input parameters and declared variables (e.g. `CREATE PROCEDURE test(@input INT = 1`, `DECLARE @var AS INTEGER = 1`) 
2. The `OUTPUT/OUT/READONLY`  keywords, which primarily serve to return scalar values or enforce immutability (e.g. `CREATE PROCEDURE test(@out_var INT OUTPUT)`)

**DOCS**
[T-SQL Stored Procedures Paramaters](https://learn.microsoft.com/en-us/sql/relational-databases/stored-procedures/specify-parameters?view=sql-server-ver16)

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -736,6 +736,16 @@ class TSQL(Dialect):
             convert.set("strict", strict)
             return convert
 
+        def _parse_column_def(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
+            this = super()._parse_column_def(this)
+            if not this:
+                return None
+            if self._match(TokenType.EQ):
+                this.set("default", self._parse_expression())
+            if self._match_texts(("OUT", "OUTPUT", "READ_ONLY")):
+                this.set("output", self._prev.text)
+            return this
+
         def _parse_user_defined_function(
             self, kind: t.Optional[TokenType] = None
         ) -> t.Optional[exp.Expression]:
@@ -1290,3 +1300,11 @@ class TSQL(Dialect):
 
         def isascii_sql(self, expression: exp.IsAscii) -> str:
             return f"(PATINDEX(CONVERT(VARCHAR(MAX), 0x255b5e002d7f5d25) COLLATE Latin1_General_BIN, {self.sql(expression.this)}) = 0)"
+
+        def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
+            this = super().columndef_sql(expression, sep)
+            default = self.sql(expression, "default")
+            default = f" = {default}" if default else ""
+            output = self.sql(expression, "output")
+            output = f" {output}" if output else ""
+            return f"{this}{default}{output}"
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -736,8 +736,10 @@ class TSQL(Dialect):
             convert.set("strict", strict)
             return convert
 
-        def _parse_column_def(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
-            this = super()._parse_column_def(this)
+        def _parse_column_def(
+            self, this: t.Optional[exp.Expression], as_use_declare: bool = False
+        ) -> t.Optional[exp.Expression]:
+            this = super()._parse_column_def(this=this, as_use_declare=as_use_declare)
             if not this:
                 return None
             if self._match(TokenType.EQ):
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -598,7 +598,7 @@ class TSQL(Dialect):
             ("ENCRYPTION", "RECOMPILE", "SCHEMABINDING", "NATIVE_COMPILATION", "EXECUTE"), tuple()
         )
 
-        COLUMN_DEFINTION_OUTPUTS = {"OUT", "OUTPUT", "READ_ONLY"}
+        COLUMN_DEFINITION_MODES = {"OUT", "OUTPUT", "READ_ONLY"}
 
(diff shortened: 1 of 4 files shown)


tests/dialects/test_tsql.py
```diff
@@ -456,6 +456,14 @@ def test_tsql(self):
         with self.assertRaises(ParseError):
             parse_one("SELECT begin", read="tsql")
 
+        self.validate_identity("CREATE PROCEDURE test(@v1 INTEGER = 1, @v2 CHAR(1) = 'c')")
+        self.validate_identity("DECLARE @v1 AS INTEGER = 1, @v2 AS CHAR(1) = 'c')")
+
+        for output in ("OUT", "OUTPUT", "READ_ONLY"):
+            self.validate_identity(
+                f"CREATE PROCEDURE test(@v1 INTEGER {output}, @v2 CHAR(1) {output})"
```
VaggelisD, line 464: Can we have both default values and outputs? If yes, we could test that by default initializing one of the args here

sqlglot/dialects/tsql.py
```diff
@@ -736,6 +736,16 @@ def _parse_convert(
             convert.set("strict", strict)
             return convert
 
+        def _parse_column_def(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
+            this = super()._parse_column_def(this)
+            if not this:
+                return None
+            if self._match(TokenType.EQ):
+                this.set("default", self._parse_expression())
+            if self._match_texts(("OUT", "OUTPUT", "READ_ONLY")):
```
VaggelisD, line 745: When sets exceed 2 elements we tend to move them outside of the function, this helps organize the keywords and the set is constructed once.

VaggelisD (commented):
Looks good, left a couple comments 👍

tests/dialects/test_tsql.py
```diff
@@ -456,6 +456,21 @@ def test_tsql(self):
...
+            self.validate_identity(
+                f"CREATE PROCEDURE test(@v1 INTEGER = 1 {output}, @v2 CHAR(1) {output})"
+            )
+
+        self.validate_all(
+            "CREATE PROCEDURE test(@v1 INTEGER = 1, @v2 CHAR(1) = 'c')",
+            read={
+                "tsql": "CREATE PROCEDURE test(@v1 AS INTEGER = 1, @v2 AS CHAR(1) = 'c')",
+            },
+        )
```
VaggelisD, line 472: This should be a `validate_identity` since it's the same dialect

sqlglot/parser.py
```diff
@@ -5638,11 +5638,16 @@ def _parse_schema(self, this: t.Optional[exp.Expression] = None) -> t.Optional[e
...
+    def _parse_column_def(
+        self, this: t.Optional[exp.Expression], as_use_declare: bool = False
+    ) -> t.Optional[exp.Expression]:
         # column defs are not really columns, they're identifiers
         if isinstance(this, exp.Column):
             this = this.this
 
+        if as_use_declare and self._match(TokenType.ALIAS):
+            return self.expression(exp.ColumnDef, this=this, kind=self._parse_types(schema=True))
+
```
VaggelisD, line 5650: This looks somewhat off, we'll parse the same`kind` and return an `exp.ColumnDef` anyways.

I think it'd be better to either: 
1. Disallow the computed column constraint parsing that is following later if `as_use_declare` (or whatever it's named) is `True`
2. Simply consume `ALIAS` eagerly, which should be the same semantically as (2)

sqlglot/parser.py
```diff
@@ -5638,11 +5638,16 @@ def _parse_schema(self, this: t.Optional[exp.Expression] = None) -> t.Optional[e
     def _parse_field_def(self) -> t.Optional[exp.Expression]:
         return self._parse_column_def(self._parse_field(any_token=True))
 
-    def _parse_column_def(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
+    def _parse_column_def(
+        self, this: t.Optional[exp.Expression], as_use_declare: bool = False
```
VaggelisD, line 5642: I'd rename `as_use_declare` to something more fitting regarding the default value initialization

tests/dialects/test_tsql.py
geooo109: The `AS` is removed when it's read (doesn't affect the roundtrip).

sqlglot/parser.py
geooo109: `default_init_with_as` ?

georgesittas: Maybe do `computed_column: bool = True` and then make it `False` to skip the branch that tries to match `AS` altogether?

tests/dialects/test_tsql.py
VaggelisD: Yep, should still be a `validate_identity(before, after)`

sqlglot/dialects/tsql.py
```diff
@@ -598,6 +598,8 @@ class Parser(parser.Parser):
             ("ENCRYPTION", "RECOMPILE", "SCHEMABINDING", "NATIVE_COMPILATION", "EXECUTE"), tuple()
         )
 
+        COLUMN_DEFINTION_OUTPUTS = {"OUT", "OUTPUT", "READ_ONLY"}
```
georgesittas, line 601: [Nit] there's a typo here, but I'd consider renaming this into `COLUMN_DEFINITION_MODES`.

georgesittas (approved):
Minor comment, LGTM otherwise.

sqlglot/dialects/tsql.py
geooo109: Agree, thanks.

PR #4708, opened 2025-02-05
Fix: avoid concealing dialect module exception in _try_load

georgesittas:
Vaggelis mentioned that as he was iterating locally, he saw an error like `dialect 'postgres' not found. Did you mean 'postgres'?`. This can happen when making changes to dialects themselves that result in an exception at module load time, because the exception is nullified => we fail to register the dialect => sqlglot crashes at runtime.

I refactored the try-except block in order to avoid hiding exceptions. This is consistent with what we do in `dialects/__init__.py` ([source](https://github.com/tobymao/sqlglot/blob/5f9030786f8489e70aee70acabee440ecf23699c/sqlglot/dialects/__init__.py#L111-L112)).

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -637,7 +637,7 @@ class Expression(metaclass=_Expression):
 
             if not root:
                 root = new_node
-            elif new_node is not node:
+            elif parent and arg_key and new_node is not node:
                 parent.set(arg_key, new_node, index)
 
         assert root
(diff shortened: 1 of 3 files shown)


VaggelisD (approved):

sqlglot/expressions.py
```diff
@@ -637,7 +637,7 @@ def transform(self, fun: t.Callable, *args: t.Any, copy: bool = True, **kwargs)
 
             if not root:
                 root = new_node
-            elif new_node is not node:
+            elif parent and arg_key and new_node is not node:
```
georgesittas, line 640: Note: this & the CTE changes in parser.py were triggered by a mypy upgrade that [broke CI](https://github.com/tobymao/sqlglot/actions/runs/13154289477/job/36707801411).

PR #4713, opened 2025-02-06
fix(snowflake): make TRY_TO_DATE default transpilation method

### LINKED ISSUE #4707: Date casting function does not transpile correctly from Spark SQL to Snowflake SQL
Reproducible code:
```
import sqlglot
sql = """select to_date(some_column, "yyyy-MM-dd) from df"""
print(sqlglot.parse_one(sql, "spark").sql("snowflake"))
```

The output is:
```
SELECT TO_DATE(some_column, 'yyyy-mm-DD') FROM df
```

The problem is that the `to_date()` function behaves differently in Spark and in Snowflake. In Spark, [it returns NULL if the input does not match the format provided](https://spark.apache.org/docs/latest/api/sql/index.html#to_date). In Snowflake, [it raises an error if the input does not match the format provided](https://docs.snowflake.com/en/sql-reference/functions/to_date). The correct transpiled function in Snowflake should be [`TRY_TO_DATE()`](https://docs.snowflake.com/en/sql-reference/functions/try_to_date), which returns a NULL value instead of raising an error.

geooo109:
Fixes #4707

This PR replaces the `TO_DATE` method with `TRY_TO_DATE` to maintain correct semantics during transpilation for Snowflake, ensuring that invalid inputs return NULL.

**DOCS**
Snowflake [TO_DATE](https://docs.snowflake.com/en/sql-reference/functions/to_date) | [TRY_TO_DATE](https://docs.snowflake.com/en/sql-reference/functions/try_to_date)


sqlglot/dialects/databricks.py
```diff
@@ -46,6 +47,7 @@ class Parser(Spark.Parser):
             "DATEDIFF": build_date_delta(exp.DateDiff),
             "DATE_DIFF": build_date_delta(exp.DateDiff),
             "GET_JSON_OBJECT": _build_json_extract,
+            "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "hive"),
```
georgesittas, line 50:
```suggestion
            "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "databricks"),
```

georgesittas:
Btw, some observations:

```
spark-sql (default)> SELECT TO_DATE('2020', 'yyyy');
2020-01-01
spark-sql (default)> SELECT TO_DATE('1');
NULL
spark-sql (default)> SELECT TO_DATE(1);
NULL
spark-sql (default)> SELECT TO_DATE('a')
NULL
spark-sql (default)> SELECT TO_DATE(1, 'yyyy'); -- Doesn't return NULL!
[INCONSISTENT_BEHAVIOR_CROSS_VERSION.PARSE_DATETIME_BY_NEW_PARSER] You may get a different result due to the upgrading to Spark >= 3.0:
Fail to parse '1' in the new parser. You can set "spark.sql.legacy.timeParserPolicy" to "LEGACY" to restore the behavior before Spark 3.0, or set to "CORRECTED" and treat it as an invalid datetime string.
<crashes, stack trace>
spark-sql (default)> SELECT TO_DATE('1', 'yyyy'); -- Doesn't return NULL!
[INCONSISTENT_BEHAVIOR_CROSS_VERSION.PARSE_DATETIME_BY_NEW_PARSER] You may get a different result due to the upgrading to Spark >= 3.0:
Fail to parse '1' in the new parser. You can set "spark.sql.legacy.timeParserPolicy" to "LEGACY" to restore the behavior before Spark 3.0, or set to "CORRECTED" and treat it as an invalid datetime string.
<crashes, stack trace>
```

Moreover, in Snowflake land...

```
> SELECT TRY_TO_DATE('2020', 'yyyy');
2020-01-01
> SELECT TRY_TO_DATE('1');
1970-01-01
> SELECT TRY_TO_DATE(1);
invalid type [TRY_TO_DATE(1)] for parameter 'TO_DATE'
> SELECT TRY_TO_DATE('a');
NULL
> SELECT TRY_TO_DATE(1, 'yyyy');
invalid type [TRY_TO_DATE(1, 'yyyy')] for parameter 'TO_DATE'
> SELECT TRY_TO_DATE('1', 'yyyy');
0001-01-01
```

PR #4717, opened 2025-02-07
feat(lineage): Enhance column extraction and add UNPIVOT support in lineage tracking

zeta9044:
This PR enhances the lineage tracking functionality in two main ways:

1. Adds robust column information extraction:
   - Introduces `_extract_source_column` function for safer column extraction
   - Handles edge cases in nested column references
   - Provides better error handling for malformed expressions
   - Maintains original table information during extraction

2. Improves table source resolution:
   - Adds support for UNPIVOT operations in lineage tracking
   - Introduces `_find_table_source` function to handle both regular and UNPIVOT cases
   - Maintains backward compatibility with existing table resolution logic

These changes make the lineage tracking more robust and extend its capabilities to handle more complex SQL transformations.

Testing:
- Added unit tests for both new private functions
- Verified backward compatibility with existing functionality
- Tested with various UNPIVOT scenarios

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -348,6 +348,7 @@ def to_node(
 
     return node
 
+
 def _extract_source_column(expr: exp.Expression) -> t.Optional[exp.Expression]:
     """
     Safely extracts source column from expression.
@@ -372,20 +373,21 @@ def _extract_source_column(expr: exp.Expression) -> t.Optional[exp.Expression]:
             expr = expr.this
 
         if isinstance(expr, exp.Dot):
-            if not hasattr(expr, 'expression') or not hasattr(expr.expression, 'name'):
+            if not hasattr(expr, "expression") or not hasattr(expr.expression, "name"):
                 return expr
 
             if isinstance(expr.this, exp.Column):
                 column = expr.this
                 # Safely extract original table and column information
                 return exp.Column(
-                    this=exp.Identifier(this=getattr(expr.expression, 'name', None)),
-                    table=getattr(column, 'table', None)
+                    this=exp.Identifier(this=getattr(expr.expression, "name", None)),
+                    table=getattr(column, "table", None),
                 )
         return expr
 
     except AttributeError:
         return expr  # Return original expression on error
+
 
 def _find_table_source(scope: Scope, table: str):
     """
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -348,7 +348,7 @@ def to_node(
 
     return node
 
-def _extract_source_column(expr: exp.Expression) -> t.Optional[exp.Column]:
+def _extract_source_column(expr: exp.Expression) -> t.Optional[exp.Expression]:
     """
     Safely extracts source column from expression.
 
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -7,13 +7,7 @@ from dataclasses import dataclass, field
 
 from sqlglot import Schema, exp, maybe_parse
 from sqlglot.errors import SqlglotError
-from sqlglot.optimizer import (
-    Scope,
-    build_scope,
-    find_all_in_scope,
-    normalize_identifiers,
-    qualify,
-)
+from sqlglot.optimizer import Scope, build_scope, find_all_in_scope, normalize_identifiers, qualify
 from sqlglot.optimizer.scope import ScopeType
 
 if t.TYPE_CHECKING:
@@ -50,9 +44,7 @@ class Node:
                 label = node.expression.sql(pretty=True, dialect=dialect)
                 source = node.source.transform(
                     lambda n: (
-                        exp.Tag(this=n, prefix="<b>", postfix="</b>")
-                        if n is node.expression
-                        else n
+                        exp.Tag(this=n, prefix="<b>", postfix="</b>") if n is node.expression else n
                     ),
                     copy=False,
                 ).sql(pretty=True, dialect=dialect)
@@ -105,10 +97,7 @@ def lineage(
     if sources:
         expression = exp.expand(
             expression,
-            {
-                k: t.cast(exp.Query, maybe_parse(v, dialect=dialect))
-                for k, v in sources.items()
-            },
+            {k: t.cast(exp.Query, maybe_parse(v, dialect=dialect)) for k, v in sources.items()},
             dialect=dialect,
         )
 
@@ -147,11 +136,7 @@ def to_node(
         scope.expression.selects[column]
         if isinstance(column, int)
         else next(
-            (
-                select
-                for select in scope.expression.selects
-                if select.alias_or_name == column
-            ),
+            (select for select in scope.expression.selects if select.alias_or_name == column),
             exp.Star() if scope.expression.is_star else scope.expression,
         )
     )
@@ -169,9 +154,7 @@ def to_node(
             )
     if isinstance(scope.expression, exp.SetOperation):
         name = type(scope.expression).__name__.upper()
-        upstream = upstream or Node(
-            name=name, source=scope.expression, expression=select
-        )
+        upstream = upstream or Node(name=name, source=scope.expression, expression=select)
 
         index = (
             column
@@ -224,8 +207,7 @@ def to_node(
         upstream.downstream.append(node)
 
     subquery_scopes = {
-        id(subquery_scope.expression): subquery_scope
-        for subquery_scope in scope.subquery_scopes
+        id(subquery_scope.expression): subquery_scope for subquery_scope in scope.subquery_scopes
     }
 
     for subquery in find_all_in_scope(select, exp.UNWRAPPED_QUERIES):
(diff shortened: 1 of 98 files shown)


georgesittas:
Hi, thanks for your PR! We'll take a look shortly.

sqlglot/lineage.py
```diff
@@ -345,6 +349,73 @@ def to_node(
...
+            if not hasattr(expr, "expression") or not hasattr(expr.expression, "name"):
+                return expr
+
+            if isinstance(expr.this, exp.Column):
+                column = expr.this
+                # Safely extract original table and column information
+                return exp.Column(
+                    this=exp.Identifier(this=getattr(expr.expression, "name", None)),
+                    table=getattr(column, "table", None),
+                )
```
georgesittas, line 385: - Why are we using `getattr` here? We've supposedly validated the existence of `name` above.
- What if `expr.this` is not a `Column`? `Dot`s can be nested multiple levels deep. Something seems off.

sqlglot/lineage.py
```diff
@@ -345,6 +349,73 @@ def to_node(
...
+        return expr
+
+    except AttributeError:
+        return expr  # Return original expression on error
```
georgesittas, line 389: The try-catch in this section doesn't really seem helpful.

sqlglot/lineage.py
```diff
@@ -345,6 +349,73 @@ def to_node(
...
+        The table node if found, None otherwise
+    """
+    if not isinstance(scope, Scope) or not table:
+        return None
+
+    # Check for PIVOT/UNPIVOT
+    for name, node in scope.references:
+        pivots = node.args.get("pivots")
+        if pivots:
+            for pivot in pivots:
```
georgesittas, line 411: Can't we do `scope.pivots` instead?

tests/test_lineage.py
```diff
@@ -576,3 +576,148 @@ def test_pivot_with_implicit_column_of_pivoted_source_and_cte(self) -> None:
...
         self.assertEqual(node.downstream[0].reference_node_name, "t")
         self.assertEqual(node.downstream[0].downstream[0].name, "quarterly_sales.empid")
+
+    def test_unpivot_column_with_join(self) -> None:
+        """
+        Tests the behavior of unpivoting columns in a SQL query combined with a join clause and validates lineage mapping for specific fields.
+        The test examines the downstream nodes of a lineage graph for fields such as id, month, and sales, ensuring the correctness of their source mappings.
+
+        :return: None
+        """
```
georgesittas, line 586: No need for this comment.

tests/test_lineage.py
```diff
@@ -576,3 +576,148 @@ def test_pivot_with_implicit_column_of_pivoted_source_and_cte(self) -> None:
...
+        node = lineage("sales", sql)
+        self.assertEqual(node.downstream[0].name, "unpvt.sales")
+        self.assertEqual(node.downstream[0].source.name, "sales_data")
+
+    def test_unpivot_column_without_join_and_with_subquery(self) -> None:
+        """
+        Tests the ability of the system to trace the lineage of columns in a SQL query, particularly focusing on columns that involve unpivoting, subqueries, and joins. This includes verifying the downstream references and ensuring correctness of mapped lineage for specific columns such as `app_id` and `daily_active_users`.
+
+        :return: None
+        """
```
georgesittas, line 619: Ditto.

sqlglot/lineage.py
```diff
@@ -277,7 +281,7 @@ def to_node(
 
     for c in source_columns:
         table = c.table
-        source = scope.sources.get(table)
+        source = _find_table_source(scope, table)
```
georgesittas, line 284: What triggered the need for this helper? We have logic dealing with pivots a few lines below. Can we instead adapt refactor that to also take `UNPIVOT` into account?

zeta9044:
Thank your mentions. You are alright. Focused on Unipvot. Pivot is included Just of pivots. My mistake.
I will pull request after fixing helper function _extract_source_column.

sqlglot/lineage.py
```diff
@@ -234,6 +234,10 @@ def to_node(
                 Node(name=select.sql(comments=False), source=source, expression=source)
             )
 
+    # Process the select expression to extract its source column information,
+    # handling any aliases or nested column references
+    select = _extract_source_column(select)
```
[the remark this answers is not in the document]
zeta9044: For example, when extracting the source table from a column like "table_a.unpivot_a.a", it returns "table_a.a". Currently, when extracting the columns from the SELECT clause, unpivot_a is treated as a column.

sqlglot/lineage.py
```diff
@@ -345,6 +349,73 @@ def to_node(
...
+        Optional[exp.Column]: Extracted column information or None
+
+    Raises:
+        TypeError: If input is not of the correct type
+    """
+    if not isinstance(expr, exp.Expression):
+        raise TypeError(f"Expected exp.Expression, got {type(expr)}")
+
+    if expr is None:
+        return None
```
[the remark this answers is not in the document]
zeta9044: OK, I will delete the above content.

sqlglot/lineage.py
```diff
@@ -345,6 +349,73 @@ def to_node(
...
+
+    try:
+        if isinstance(expr, exp.Alias):
+            expr = expr.this
+
+        if isinstance(expr, exp.Dot):
+            if not hasattr(expr, "expression") or not hasattr(expr.expression, "name"):
+                return expr
```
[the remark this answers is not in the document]
zeta9044: OK, I will delete the above content.

sqlglot/lineage.py
zeta9044: Using getattr is just a habit, but you’re right. I’ll simplify it.
I actually noticed the nested case issue before seeing your response, but I appreciate your careful review.
So, I’ll fix this by handling it recursively.

sqlglot/lineage.py
zeta9044: You’re right. The above code is merely avoiding the flaw in my logic.
It would result in ignoring errors within the try-catch block.
Therefore, I will remove the try-catch.

sqlglot/lineage.py
zeta9044: Thanks! I wasn’t aware of the existence of scope.pivots.
I’ll change it as you suggested.

sqlglot/lineage.py
```diff
@@ -345,6 +349,91 @@ def to_node(
...
+        if hasattr(current_expr, "this"):
+            result = find_dot_in_expression(current_expr.this)
+            if result:
+                return result
+
+        if hasattr(current_expr, "expressions"):
+            for item in current_expr.expressions or []:
+                result = find_dot_in_expression(item)
+                if result:
+                    return result
```
georgesittas, line 410: I don't get what this is achieving, it doesn't look right. Why are we still using `hasattr`?

sqlglot/lineage.py
```diff
@@ -345,6 +349,91 @@ def to_node(
...
+            return None
+
+        # Handle Dot expressions
+        if isinstance(current_expr, exp.Dot):
+            if isinstance(current_expr.this, exp.Column):
+                column = current_expr.this
+                return exp.Column(
+                    this=exp.Identifier(this=current_expr.expression.name), table=column.table
+                )
+            return None
```
georgesittas, line 373: I don't think this is the right approach to unfolding a `Dot`.

sqlglot/lineage.py
```diff
@@ -345,6 +349,91 @@ def to_node(
...
+
+        # Handle complex nested expressions
+        if isinstance(
+            current_expr, (exp.Case, exp.SetOperation, exp.Window, exp.MatchRecognize, exp.Pivot)
+        ):
```
georgesittas, line 378: How was this list compiled? Why isn't it tested?

sqlglot/lineage.py
georgesittas: @zeta9044 please don't mark unresolved comments as resolved.

georgesittas (commented):
Hey @zeta9044, appreciate the effort here but I think I'll close this for the time being. There's some unjustified complexity in this PR that isn't documented or tested, and we don't have the bandwidth to continue iterating on this right now.

If you want, please address the remaining comments and post another comment in this PR's thread, so someone in the core team can take a quick look before continuing.

zeta9044:
I will move the _extract_source_column and _find_table_source functions to sqlglot.helper.py and elevate unpivot to the same level as pivot.

The _extract_source_column function will be validated through a test case in the helper functions. However, I do not agree with all the issues that were previously pointed out. I believe the correct way to find the dot class without causing a runtime error is by using hasattr.

Therefore, I will delete the existing pull request and resubmit it with more detailed documentation.

zeta9044:
> Hey @zeta9044, appreciate the effort here but I think I'll close this for the time being. There's some unjustified complexity in this PR that isn't documented or tested, and we don't have the bandwidth to continue iterating on this right now.
> 
> 
> 
> If you want, please address the remaining comments and post another comment in this PR's thread, so someone in the core team can take a quick look before continuing.

I will move the _extract_source_column and _find_table_source functions to sqlglot.helper.py and elevate unpivot to the same level as pivot.

The _extract_source_column function will be validated through a test case in the helper functions. However, I do not agree with all the issues that were previously pointed out. I believe the correct way to find the dot class without causing a runtime error is by using hasattr.

Therefore, I will delete the existing pull request and resubmit it with more detailed documentation.

PR #4718, opened 2025-02-07
fix(tsql, postgres)!: Improve UUID support

amachanic:
This PR addresses some UUID support issues with T-SQL and Postgres. A small example of what I wanted to get working:

Input: sqlglot.transpile('create table x (i uniqueidentifier not null default(newid()))', read='tsql', write='postgres')
Output: ['CREATE TABLE x (i UNIQUEIDENTIFIER NOT NULL DEFAULT (NEWID()))']
Expected: ['CREATE TABLE x (i uuid NOT NULL DEFAULT (gen_random_uuid()))']

- Removed unnecessary UNIQUEIDENTIFER token (standardized on UUID)
- Added NEWID() support to T-SQL
- Cleaned up Spark UUID support, which I noticed was also broken. (I think Hive might be, too, but I don't know for a fact so I didn't touch that one.)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6748,7 +6748,7 @@ class UnixSeconds(Func):
 
 
 class Uuid(Func):
-    _sql_names = ["UUID", "GEN_RANDOM_UUID", "GENERATE_UUID", "NEWID", "UUID_STRING"]
+    _sql_names = ["UUID", "GEN_RANDOM_UUID", "GENERATE_UUID", "UUID_STRING"]
 
     arg_types = {"this": False, "name": False}
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4476,7 +4476,6 @@ class DataType(Expression):
         UMEDIUMINT = auto()
         UDECIMAL = auto()
         UNION = auto()
-        UNIQUEIDENTIFIER = auto()
         UNKNOWN = auto()  # Sentinel value, useful for type annotation
         USERDEFINED = "USER-DEFINED"
         USMALLINT = auto()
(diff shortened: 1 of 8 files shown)


sqlglot/expressions.py
```diff
@@ -6749,7 +6748,7 @@ class UnixSeconds(Func):
 
 
 class Uuid(Func):
-    _sql_names = ["UUID", "GEN_RANDOM_UUID", "GENERATE_UUID", "UUID_STRING"]
+    _sql_names = ["UUID", "GEN_RANDOM_UUID", "GENERATE_UUID", "NEWID", "UUID_STRING"]
```
amachanic, line 6751: Having to touch this seemed a little bit weird. Not sure if I missed something in the specific dialect implementations.

georgesittas (approved):
Thanks for the contribution! FYI, this is breaking, so the PR title needs to reflect that. I'll make the change, just something to keep in mind for future PRs.

Looks good 👍

sqlglot/expressions.py
georgesittas: You can either do this or add `"NEWID": exp.Uuid.from_arg_list` in `TSQL.Parser.FUNCTIONS`. The latter is safer, because the approach in this PR may have silent side-effects if there are other dialects that support `NEWID(...)` and, e.g., it expects a different number of arguments than what the expression supports.

My hunch is this should be safe here, but could you quickly check if this function exists in other dialects and whether its semantics / argument count matches what we have here?

amachanic: Okay, that's much nicer. I went ahead and made that change in https://github.com/tobymao/sqlglot/pull/4718/commits/293edfee92bec60ec112d272cac40308ac105e77

PR #4719, opened 2025-02-07
fix(tsql)!: Properly set 'this' when parsing IDENTITY

amachanic:
This PR fixes an issue whereby T-SQL to Postgres transpilation doesn't properly work when the IDENTITY property is used in a CREATE TABLE statement:

Input: sqlglot.transpile('create table x (i int identity(1,1) not null)', read='tsql', write='postgres')
Output: ['CREATE TABLE x (i INT GENERATED AS IDENTITY (START WITH 1 INCREMENT BY 1) NOT NULL)']
Expected: ['CREATE TABLE x (i INT GENERATED BY DEFAULT AS IDENTITY (START WITH 1 INCREMENT BY 1) NOT NULL)']

The output as-is is not valid; GENERATED must be either "ALWAYS" or "BY DEFAULT". In this case "BY DEFAULT" is appropriate because it allows a user to explicitly insert their own value if they'd like, which the SQL Server IDENTITY property also supports.


amachanic:
I'm not used to working with ruff, so maybe I'm doing something wrong. I ran `ruff format` locally and I'm getting a lot of unrelated reformats. Is there a configuration file or something I'm supposed to point to?

tobymao:
@amachanic you should just run make style

amachanic:
> @amachanic you should just run make style

georgesittas (approved):
Thanks! This is breaking because it changes the AST structure.

PR #4722, opened 2025-02-10
fix: pushdown_predicates takes care of join marks before the actual push down

snovik75:
This fix makes sure that statements with join marks are converted into their canonical form before pushing down predicates.

Currently pushdown_predicates ignores the join marks syntax and does not convert join marks into joins which breaks the SQL

```
script = "SELECT x.a FROM x, (SELECT * FROM y) AS y WHERE x.a = y.a(+) and y.b(+) = 1"

tree = parse_one(script, dialect="oracle")
pushdown_predicates(tree).sql(pretty=True, dialect="oracle")
```

produces broken code:

```
SELECT
  x.a
FROM x
JOIN (
  SELECT
    *
  FROM y
  WHERE
    y.b (+) = 1
) y
  ON x.a = y.a (+) AND TRUE
WHERE
  TRUE AND TRUE
```

Testing:
- Added a unit test for join marks

### DIFF
--- a/sqlglot/optimizer/pushdown_predicates.py
+++ b/sqlglot/optimizer/pushdown_predicates.py
@@ -1,13 +1,15 @@
+import typing as t
+
 from sqlglot import exp
 from sqlglot.optimizer.normalize import normalized
 from sqlglot.optimizer.scope import build_scope, find_in_scope
 from sqlglot.optimizer.simplify import simplify
-from sqlglot.dialects.dialect import Dialect
+from sqlglot.dialects.dialect import DialectType
 from sqlglot.transforms import eliminate_join_marks as eliminate_jm
 
 
 def pushdown_predicates(
-    expression, dialect: Dialect | None = None, eliminate_join_marks: bool = False
+    expression, dialect: t.Optional[DialectType] = None, eliminate_join_marks: bool = False
 ):
     """
     Rewrite sqlglot AST to pushdown predicates in FROMS and JOINS
--- a/sqlglot/optimizer/pushdown_predicates.py
+++ b/sqlglot/optimizer/pushdown_predicates.py
@@ -2,6 +2,7 @@ from sqlglot import exp
 from sqlglot.optimizer.normalize import normalized
 from sqlglot.optimizer.scope import build_scope, find_in_scope
 from sqlglot.optimizer.simplify import simplify
+from sqlglot.transforms import eliminate_join_marks
 
 
 def pushdown_predicates(expression, dialect=None):
@@ -27,6 +28,7 @@ def pushdown_predicates(expression, dialect=None):
 
         for scope in reversed(list(root.traverse())):
             select = scope.expression
+            select = eliminate_join_marks(select)
             where = select.args.get("where")
             if where:
                 selected_sources = scope.selected_sources
--- a/sqlglot/optimizer/pushdown_predicates.py
+++ b/sqlglot/optimizer/pushdown_predicates.py
@@ -2,10 +2,11 @@ from sqlglot import exp
 from sqlglot.optimizer.normalize import normalized
 from sqlglot.optimizer.scope import build_scope, find_in_scope
 from sqlglot.optimizer.simplify import simplify
-from sqlglot.transforms import eliminate_join_marks
+from sqlglot.sqlglot.dialects.dialect import Dialect
+from sqlglot.transforms import eliminate_join_marks as eliminate_join_marks
 
 
-def pushdown_predicates(expression, dialect=None):
+def pushdown_predicates(expression, dialect: Dialect | None = None):
     """
     Rewrite sqlglot AST to pushdown predicates in FROMS and JOINS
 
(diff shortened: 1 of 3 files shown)


tests/fixtures/optimizer/pushdown_predicates.sql
```diff
@@ -44,3 +44,8 @@ SELECT x.cnt AS cnt FROM (SELECT COUNT(1) AS cnt FROM x AS x HAVING COUNT(1) > 0
 -- Pushdown predicate to HAVING (DNF)
 SELECT x.cnt AS cnt FROM (SELECT COUNT(1) AS cnt, COUNT(x.a) AS cnt_a, COUNT(x.b) AS cnt_b FROM x AS x) AS x WHERE (x.cnt_a > 0 AND x.cnt_b > 0) OR x.cnt > 0;
 SELECT x.cnt AS cnt FROM (SELECT COUNT(1) AS cnt, COUNT(x.a) AS cnt_a, COUNT(x.b) AS cnt_b FROM x AS x HAVING COUNT(1) > 0 OR (COUNT(x.a) > 0 AND COUNT(x.b) > 0)) AS x WHERE x.cnt > 0 OR (x.cnt_a > 0 AND x.cnt_b > 0);
+
+-- Correctly treat join marks while pushing predicates down
+# dialect: oracle, redshift
```
georgesittas, line 49: I don't think this syntax is correct, we don't split dialects comma. You can specify only one of these two.

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -27,6 +28,7 @@ def pushdown_predicates(expression, dialect=None):
 
         for scope in reversed(list(root.traverse())):
             select = scope.expression
+            select = eliminate_join_marks(select)
```
georgesittas, line 31: - This transformation should not be used by default; the transformation syntax is obsolete, meaning that the performance hit is not justified for the general case. Also, only Oracle and Redshift support this syntax IIRC. I'd consider adding a flag in `pushdown_predicates` instead, e.g. `eliminate_join_marks: bool = False`, so you can then override that in `optimize`.
- We shouldn't call this within the scope traversal loop. The transformation itself does a scope traversal, so it seems like this would result in a `O(scopes²)` kind of performance.

georgesittas (changes requested):

tests/fixtures/optimizer/pushdown_predicates.sql
georgesittas: For the comment above, you can do `# title: ...`

sqlglot/optimizer/pushdown_predicates.py
snovik75: Thanks, fair points. Will fix them.

tests/fixtures/optimizer/pushdown_predicates.sql
snovik75: I saw a lot of cases in the current code where it is split by commas

image

sqlglot/optimizer/pushdown_predicates.py
snovik75: I went a slightly different route and rely on the dialect flag SUPPORTS_COLUMN_JOIN_MARKS. My motivation is that the codebase should not break down the code and/or rely the deep understanding of the codebase from the users. 
If you want I can go to your version or maybe still add an extra parameter to pushdown_predicates

georgesittas: Yeah, let's not go down this road. I prefer us explicitly saying this transformation needs to run. I don't think the general user uses join marks, so using the flag as a proxy will just result in many oracle / redshift users getting an unjustified performance hit.

> rely on the deep understanding of the codebase from the users.

I think using the optimizer on its own requires supervision, so this shouldn't be a problem.

snovik75: ok, sure thing

tests/fixtures/optimizer/pushdown_predicates.sql
georgesittas: This testing suite is special, in general this syntax is not supported.

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -17,9 +21,13 @@ def pushdown_predicates(expression, dialect=None):
 
     Args:
         expression (sqlglot.Expression): expression to optimize
+        eliminate_join_marks (Optional[bool]): an override to eliminate legacy join marks
     Returns:
         sqlglot.Expression: optimized expression
     """
+    if eliminate_join_marks:
```
tobymao, line 28: why is this necessary? you can just call eliminate_jm before you call pushdown_predicates

tobymao:
i don't think this PR is needed, for your use case, you can pass in custom rules and include eliminate_join_marks

georgesittas:
> i don't think this PR is needed, for your use case, you can pass in custom rules and include eliminate_join_marks

Ah, good point. Yeah, I agree. I'm also thinking now that the two rules involved in this PR are orthogonal, which is another reason why they'd be separated– it makes the solution more robust, because you avoid a dependency on the predicate pushdown.

I'll go ahead and close this for now, apologies for the back and forth @snovik75.

Issue #4723, opened 2025-02-10
Parsing fails for DDL with AggregateFunctions containing a keyword


dorranh:
Hi Sqlglot team! While playing around with Sqlglot recently I ran into an issue parsing some DDL for the `clickhouse` dialect. Specifically it looks like parsing fails for cases where a Clickhouse table definition includes a column of type `AggregateFunction` which uses a parameterless function without arguments that is also a keyword. For example:

```sql
CREATE TABLE my_db.my_table
  (
      someId UUID,
      aggregatedColumn AggregateFunction(any, String),
  )
ENGINE = AggregatingMergeTree()
ORDER BY (someId)
```

Version: `26.4.1`
Dialect: `clickhouse` (but could apply to any other dialect utilizing aggregate functions)

Links:
* https://clickhouse.com/docs/en/sql-reference/data-types/aggregatefunction
* https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/any

PR #4727, opened 2025-02-11
fix: Add additional allowed tokens for parsing aggregate functions in DDL

dorranh:
Fixes https://github.com/tobymao/sqlglot/issues/4723 by adding an additional allowed token when parsing column definitions with an aggregate function type. Note that while the specific issue manifests itself for the `clickhouse` dialect, the issue could potentially apply to any dialect supporting aggregate functions which support using a keyword function like `any` in the column type.


georgesittas (approved):
Thanks!

PR #4734, opened 2025-02-12
feat(mysql): add unsigned double type

GabrielVSMachado:
# Objective
This PR adds a `UDOUBLE` type. 

## Explanation
Accordingly [mysql documentation](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html) this feature is deprecated but still has support due past versions, turning the column as regular double type.

### Examples
```sql
CREATE TABLE `bar` (`abacate` DOUBLE(10, 2) UNSIGNED NOT NULL);
```

```sql
SELECT COLUMN_NAME, DATA_TYPE FROM information_schema.columns WHERE table_name = 'bar' AND column_name = 'abacate';
```
RESULT:
image


georgesittas (approved):
Thanks

Issue #4739, opened 2025-02-13
Incorrect binary conversion for duckdb


sk-:
When converting statements from snowflake to duckdb (via [fakesnow](https://github.com/tekumara/fakesnow)) we get casting errors, because the result is not castable to a blob.

```
duckdb.duckdb.ConversionException: Conversion Error: Unimplemented type for cast (HUGEINT -> BLOB)
```

**Fully reproducible code snippet**
The code below will reproduce the issue:

```python
>>> import sqlglot
>>> e = sqlglot.parse_one("SELECT x'abcdef'", read="snowflake")
>>> e.expressions[0].sql("duckdb")
'11259375'
>>> e.sql("duckdb")
'SELECT 11259375'
```

**Official Documentation**
However, note that duckdb supports binary [literals](https://duckdb.org/docs/sql/data_types/blob.html). 

I tested the following query on a duckdb [playground](https://codapi.org/embed/?sandbox=duckdb&code=data%3A%3Bbase64%2CC3b1cXUOUVCPqUhMiqlITompSE1Tt7JKyslP0lFIK8rPjc9IrdBQT0xKTklNU9fUwapSwRaLUgA%3D):
```sql
SELECT '\xab\xcd\xef'::blob, from_hex('abcdef'), '\xab\xcd\xef'::blob = from_hex('abcdef')

│ '\xAB\xCD\xEF'::BLOB │ from_hex('abcdef') │ ('\xAB\xCD\xEF'::BLOB = from_hex('abcdef')) │
├──────────────────────┼────────────────────┼─────────────────────────────────────────────┤
│ \xAB\xCD\xEF         │ \xAB\xCD\xEF       │ true                                        │
```

PR #4743, opened 2025-02-14
Fix!: Support for exp.HexString in DuckDB/Presto/Trino

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/4739

[More context from our internal discussion](https://tobikodata.slack.com/archives/C040V7ZP9NH/p1739524518868049)

DuckDB does not support hex strings of the form `x'ABC'`, so `exp.HexString` is generated in the function representation:

```SQL
D SELECT FROM_HEX('CC'), FROM_HEX('CC') = '\xCC'::BLOB;
┌────────────────┬─────────────────────────────────┐
│ from_hex('CC') │ (from_hex('CC') = '\xCC'::BLOB) │
│      blob      │             boolean             │
├────────────────┼─────────────────────────────────┤
│ \xCC           │ true                            │
└────────────────┴─────────────────────────────────┘
```

Also, Presto & Trino support hex strings so this PR extends their parsing & transpilation too:

```SQL
presto> select x'CC', X'CC', typeof(x'cc');
 _col0 | _col1 |   _col2
-------+-------+-----------
 cc    | cc    | varbinary



trino> select x'CC', X'CC', typeof(x'cc');
 _col0 | _col1 |   _col2
-------+-------+-----------
 cc    | cc    | varbinary
```

<br />

EDIT: The PR is also extended to cover BQ's & CH's semantics which treat the hex strings as integers:

```SQL
bigquery> SELECT 0xCC = FROM_HEX('CC');
No matching signature for operator = for argument types: INT64, BYTES 
```

Docs
------
[CH Numeric types](https://clickhouse.com/docs/en/sql-reference/syntax#numeric)


sqlglot/generator.py
```diff
@@ -1276,11 +1276,19 @@ def bitstring_sql(self, expression: exp.BitString) -> str:
...
+    def hexstring_sql(self, expression: exp.HexString, binary_function_repr: bool = False) -> str:
         this = self.sql(expression, "this")
-        if self.dialect.HEX_START:
-            return f"{self.dialect.HEX_START}{this}{self.dialect.HEX_END}"
-        return f"{int(this, 16)}"
+        is_integer_type = expression.args.get("is_integer")
+
+        if is_integer_type or (not self.dialect.HEX_START and not binary_function_repr):
+            return f"{int(this, 16)}"
+        elif not is_integer_type and self.dialect.HEX_STRING_IS_INTEGER_TYPE:
```
georgesittas, line 1285: Reaching the second branch means `is_integer_type` must be `False`, otherwise we'd take the 1st one:

```suggestion
        if self.dialect.HEX_STRING_IS_INTEGER_TYPE:
```

sqlglot/dialects/dialect.py
```diff
@@ -469,6 +469,9 @@ class Dialect(metaclass=_Dialect):
     NUMBERS_CAN_BE_UNDERSCORE_SEPARATED = False
     """Whether number literals can include underscores for better readability"""
 
+    HEX_STRING_IS_INTEGER_TYPE: t.Optional[bool] = None
+    """Whether hex strings such as x'CC' evaluate to integer or binary/blob type"""
```
georgesittas, line 473: Let's make this a simple `bool`, default to `False` and in the parser you can do `<flag> or None` if you want to avoid setting the arg.

sqlglot/dialects/bigquery.py
```diff
@@ -903,6 +904,7 @@ class Generator(generator.Generator):
             exp.GenerateSeries: rename_func("GENERATE_ARRAY"),
             exp.GroupConcat: rename_func("STRING_AGG"),
             exp.Hex: lambda self, e: self.func("UPPER", self.func("TO_HEX", self.sql(e, "this"))),
+            exp.HexString: lambda self, e: self.hexstring_sql(e, binary_function_repr=True),
```
georgesittas, line 907: I suggest preserving 0xff in BigQuery and ClickHouse, this will convert it into 255.

tests/dialects/test_dialect.py
```diff
@@ -3347,3 +3347,20 @@ def test_current_schema(self):
...
             },
         )
+
+    def test_integer_hex_strings(self):
+        # Hex strings such as 0xCC represent INTEGER values in the read dialects
+        for read_dialect in ("bigquery", "clickhouse"):
+            for write_dialect in (
+                "",
+                "duckdb",
+                "bigquery",
```
georgesittas, line 3357: Let's add clickhouse to this write list

sqlglot/generator.py
```diff
@@ -1276,11 +1276,19 @@ def bitstring_sql(self, expression: exp.BitString) -> str:
...
+            self.unsupported("Unsupported transpilation from BINARY/BLOB hex string")
+
+        if binary_function_repr:
+            return self.func("FROM_HEX", exp.Literal.string(expression.this))
```
georgesittas, line 1289: Is this also valid?
```suggestion
            return self.func("FROM_HEX", exp.Literal.string(this))
```

georgesittas (commented):
Ok almost there, just need to make a few small improvements imo.

tests/dialects/test_dialect.py
VaggelisD: Added separate roundtrip/transpilation tests between BQ & CH

sqlglot/generator.py
```diff
@@ -1276,11 +1276,28 @@ def bitstring_sql(self, expression: exp.BitString) -> str:
...
+            # - The read dialect treats the hex value as integer literal but not the write
+            # - The transpilation is not supported (write dialect hasn't set HEX_START or the param flag)
+            return f"{int(this, 16)}"
+
+        if not is_integer_type:
+            # Read dialect treats the hex value as BINARY/BLOB
+            if binary_function_repr:
+                # The write dialect supports the transpilation to its equivalent BINARY/BLOB
+                return self.func("FROM_HEX", exp.Literal.string(this))
+            elif self.dialect.HEX_STRING_IS_INTEGER_TYPE:
```
georgesittas, line 1296:
```suggestion
            if self.dialect.HEX_STRING_IS_INTEGER_TYPE:
```

sqlglot/generator.py
```diff
@@ -1276,11 +1276,28 @@ def bitstring_sql(self, expression: exp.BitString) -> str:
...
+            # Integer representation will be returned if:
+            # - The read dialect treats the hex value as integer literal but not the write
+            # - The transpilation is not supported (write dialect hasn't set HEX_START or the param flag)
+            return f"{int(this, 16)}"
+
+        if not is_integer_type:
+            # Read dialect treats the hex value as BINARY/BLOB
+            if binary_function_repr:
+                # The write dialect supports the transpilation to its equivalent BINARY/BLOB
+                return self.func("FROM_HEX", exp.Literal.string(this))
```
georgesittas, line 1295: I'd consider passing the actual function name instead of hard-coding `FROM_HEX` here. Makes it easier to extend.

georgesittas (approved):

PR #4744, opened 2025-02-14
Fix(clickhouse)!: respect type nullability when casting in strtodate_sql

sleshJdev:
When transpiling from e.g. oracle `select to_date(substr(nullable_column, 1,10), 'YYYY-MM-DD')` to clickhouse the following 
 `SELECT CAST(STR_TO_DATE(SUBSTRING(nullable_column, 1, 10), '%Y-%m-%d') AS DATE)` which fails `Cannot convert NULL value to non-Nullable type: while executing`.

By this PR we ask sqlglot to use already existing cast method which respects data types nullity while transpiling.


georgesittas (approved):
Thanks for the PR!

Issue #4746, opened 2025-02-14
Clickhouse dialect to BOTH as trim type if it's not specified


sleshJdev:
**Fully reproducible code snippet**

`parse_one("select trim('|' FROM '||Hello ||| world||')", read='oracle').sql(dialect='clickhouse')` produces incorrect clickhouse sql:

```
select trim('|' FROM '||Hello ||| world||');
```
the error
```
Syntax error: failed at position 22 (''||Hello ||| world||''): '||Hello ||| world||'). Expected one of: token, DoubleColon, ClosingRoundBracket, OR, AND, IS NOT DISTINCT FROM, IS NULL, IS NOT NULL, BETWEEN, NOT BETWEEN, LIKE, ILIKE, NOT LIKE, NOT ILIKE, REGEXP, IN, NOT IN, GLOBAL IN, GLOBAL NOT IN, MOD, DIV, alias, AS, identifier. (SYNTAX_ERROR) (version 24.8.5.115 (official build))
```
the expected result should be

```
select trim(both '|' FROM '||Hello ||| world||');
```

**Official Documentation**
1. https://clickhouse.com/docs/en/sql-reference/functions/string-functions#trim
2. https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/[TRIM](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/TRIM.html).html

PR #4747, opened 2025-02-14
Clickhouse dialect to BOTH as trim type if it's not specified

sleshJdev:
Fixes https://github.com/tobymao/sqlglot/issues/4746

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -30,6 +30,17 @@ from sqlglot.generator import unsupported_args
 DATEΤΙΜΕ_DELTA = t.Union[exp.DateAdd, exp.DateDiff, exp.DateSub, exp.TimestampSub, exp.TimestampAdd]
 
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -30,6 +30,17 @@
 DATEΤΙΜΕ_DELTA = t.Union[exp.DateAdd, exp.DateDiff, exp.DateSub, exp.TimestampSub, exp.TimestampAdd]
 
 
+def _trim_sql(self: ClickHouse.Generator, expression: exp.Trim) -> str:
+    position = expression.args.get("position")
+
+    if not position:
+        expression = expression.copy()
```
georgesittas, line 37: We shouldn't copy, this is done in `generate` already.

georgesittas (commented):
I'll take a stab at fixing this slightly differently @sleshJdev, will ping you in a bit. Thanks for the PR regardless, appreciate it.

georgesittas:
Check out https://github.com/tobymao/sqlglot/commit/fda40a9e670f30822dee05651d2336494c0d4f3b.

PR #4749, opened 2025-02-16
fix: DuckDB should use count_if_to_sum transformation.

cvm-a:
DuckDB COUNT_IF was only fixed to work as an aggregate function fully in 1.2.0 (via  https://github.com/duckdb/duckdb/pull/15061), but it does not function correctly in previous versions, since it was raising a error 'Catalog Error: count_if is not an aggregate function`( https://github.com/duckdb/duckdb-web/issues/3294)

With this PR,  COUNT_IF are translated to code that  works for both prior versions and 1.2.0+

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -14,6 +14,7 @@ from sqlglot.dialects.dialect import (
     binary_from_function,
     bool_xor_sql,
     build_default_decimal_type,
+    count_if_to_sum,
     date_trunc_to_time,
     datestrtodate_sql,
     no_datetime_sql,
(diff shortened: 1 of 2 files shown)


georgesittas (commented):
Hey @cvm-a, the de-sugarized `COUNT_IF` form is indeed equivalent and this PR is fine, but since this [commit](https://github.com/tobymao/sqlglot/commit/466c839c2cfc94b398dd619b738df165f2876cdb) introduced a split in DuckDB based on its version, I think I'll do that here too and instead require specifying a version <1.2 if you don't want the `COUNT_IF` syntax to be used.

georgesittas:
Closing this in favor of https://github.com/tobymao/sqlglot/pull/4755. Regardless, thanks for the PR!

cvm-a:
Thanks @georgesittas

PR #4756, opened 2025-02-17
Fix(tokenizer-rs): use is_whitespace in token scan loop

### LINKED ISSUE #4748: Inconsistent tokenization results between Python and Rust when handling full-width spaces
I've discovered several inconsistencies in SQL parsing behavior between Python and Rust implementations:

- Some SQL statements that can be parsed using Python cannot be parsed using Rust
- SQL containing full-width spaces (`\u3000`) fails to parse when using Rust
- The Python and Rust tokenizers produce different token outputs for the same input

Environment:

- sqlglot version: 26.6.0
- Python version: 3.12
- OS: Linux

To reproduce:
```python
# sample_tokenizer.py
from sqlglot.tokens import Tokenizer

if __name__ == "__main__":
    sql = "SELECT　* FROM table WHERE id = 1;"  # Contains full-width space after SELECT
    tokenizer = Tokenizer()
    tokens = tokenizer.tokenize(sql)

    for token in tokens:
        print(f"{token.token_type}, ({token.line}, {token.col})")
```

```python
# sample_parser.py
import sqlglot

if __name__ == "__main__":
    print(sqlglot.__version__)  # 26.6.0
    sql = "SELECT　* FROM table WHERE id = 1;"  # Contains full-width space after SELECT
    parsed_output = sqlglot.parse_one(sql, read="bigquery")
```

Python-only installation:
```sh
$ python -m venv .venv-sqlglot
$ source .venv-sqlglot/bin/activate
$ pip install "sqlglot"
$ python sample_tokenizer.py
TokenType.SELECT, (1, 6)
TokenType.STAR, (1, 8)
TokenType.FROM, (1, 13)
TokenType.TABLE, (1, 19)
TokenType.WHERE, (1, 25)
TokenType.VAR, (1, 28)
TokenType.EQ, (1, 30)
TokenType.NUMBER, (1, 32)
TokenType.SEMICOLON, (1, 33)

$ python sample_parser.py
26.6.0
```

With Rust installation:
```sh
$ python -m venv .venv-sqlglot-rs
$ source .venv-sqlglot-rs/bin/activate
$ pip install "sqlglot[rs]"
$ python sample_tokenizer.py
TokenType.SELECT, (1, 6)
TokenType.VAR, (1, 7)       # Note the extra VAR token here
TokenType.STAR, (1, 8)
TokenType.FROM, (1, 13)
TokenType.TABLE, (1, 19)
TokenType.WHERE, (1, 25)
TokenType.VAR, (1, 28)
TokenType.EQ, (1, 30)
TokenType.NUMBER, (1, 32)
TokenType.SEMICOLON, (1, 33)

$ python sample_parser.py
26.6.0
sqlglot.errors.ParseError: Required keyword: 'expression' missing for <class 'sqlglot.expressions.Mul'>. Line 1, Col: 13.
  SELECT　* FROM table WHERE id = 1;
```

The Python tokenizer ignores full-width spaces while the Rust tokenizer converts them into VAR tokens.
As a result, tokens produced by the Python tokenizer can be parsed successfully, but tokens from the Rust tokenizer cause a parsing error when processed by sqlglot.

Would it be possible to align the behavior between Python and Rust implementations when handling full-width spaces? While using full-width spaces in SQL is not recommended, they do appear in production queries and are supported by engines like BigQuery.

georgesittas:
Fixes #4748

### DIFF
--- a/sqlglotrs/src/tokenizer.rs
+++ b/sqlglotrs/src/tokenizer.rs
@@ -142,7 +142,7 @@ impl<'a> TokenizerState<'a> {
                 break;
             }
 
-            if !self.settings.white_space.contains_key(&self.current_char) {
+            if !self.current_char.is_whitespace() {
                 if self.current_char.is_ascii_digit() {
                     self.scan_number()?;
                 } else if let Some(identifier_end) =
--- a/tests/test_transpile.py
+++ b/tests/test_transpile.py
@@ -23,6 +23,11 @@ class TestTranspile(unittest.TestCase):
 
     def test_weird_chars(self):
         self.assertEqual(transpile("0Êß")[0], "0 AS Êß")
+        self.assertEqual(
+            # Ideographic space after SELECT (\u3000)
+            transpile("SELECT　* FROM t WHERE c = 1")[0],
+            "SELECT * FROM t WHERE c = 1",
+        )
 
     def test_alias(self):
         self.assertEqual(transpile("SELECT SUM(y) KEEP")[0], "SELECT SUM(y) AS KEEP")


VaggelisD (approved):
Great find 💯

PR #4761, opened 2025-02-18
Feat(snowflake): add support for USE SECONDARY ROLES

### LINKED ISSUE #4760: parsing failure for USE (snowflake)
`USE SECONDARY ROLES NONE`
dialect = Snowflake

Fails on `ParseError: Invalid expression / Unexpected token` for `NONE`

https://docs.snowflake.com/en/sql-reference/sql/use-secondary-roles

georgesittas:
Fixes #4760

Reference: https://docs.snowflake.com/en/sql-reference/sql/use-secondary-roles

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -573,6 +573,16 @@ class Snowflake(Dialect):
             ),
         }
 
+        def _parse_use(self) -> exp.Use:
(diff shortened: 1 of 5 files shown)


VaggelisD (approved):
Should we parse `ALL | NONE` into an `exp.Var`?

PR #4765, opened 2025-02-18
Add tokens and corresponding tests to Oracle dialect

ericvergnaud:
Oracle supports a number of custom datatypes not yet covered by the current codebase.
This PR adds tokens some of these datatypes and corresponding unit tests.


georgesittas:
Hey @ericvergnaud, I don't think the approach in this PR is correct. You're only testing sqlglot dialect -> oracle, but you're not preserving the roundtrip of these types in oracle. This means that you'd break valid oracle queries.

I'll close this PR for now. What's the motivation behind these changes? Can you please share documentation & a detailed description if you plan to open another one?

PR #4767, opened 2025-02-19
fix(optimizer): Preserve CTE names during table qualification

zeta9044:
When qualifying table references in SQL queries, Common Table Expressions (CTEs) should not be qualified with database and catalog names. This change ensures that CTE references remain unqualified while still properly qualifying regular table references.

Previously, the qualify_tables function would incorrectly qualify CTE references as if they were regular tables, resulting in invalid SQL like: `db.schema.cte_name`. Now, it first collects CTE names and skips qualification for any table reference that matches a CTE name.

Example:
Before:
WITH cte AS (SELECT * FROM table)
SELECT * FROM db.schema.cte

After:
WITH cte AS (SELECT * FROM db.schema.table)
SELECT * FROM cte

Changes:
- Collect CTE names at the start of qualify_tables
- Skip qualification for table references that match CTE names
- Maintain qualification for regular table references

### DIFF
--- a/sqlglot/optimizer/qualify_tables.py
+++ b/sqlglot/optimizer/qualify_tables.py
@@ -50,8 +50,17 @@ def qualify_tables(
     db = exp.parse_identifier(db, dialect=dialect) if db else None
     catalog = exp.parse_identifier(catalog, dialect=dialect) if catalog else None
 
+    # First collect all CTE names
+    cte_names = set()
+    for scope in traverse_scope(expression):
+        for cte in scope.ctes:
+            cte_names.add(cte.alias)
+
     def _qualify(table: exp.Table) -> None:
         if isinstance(table.this, exp.Identifier):
+            # Skip qualification if table is a CTE reference
+            if table.name in cte_names:
+                return
             if not table.args.get("db"):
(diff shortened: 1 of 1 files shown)


tobymao:
Can you show a failing test case of this?

zeta9044:
> Can you show a failing test case of this?
# case : default_db, default_schema exists
```sql
INSERT INTO sales_summary
WITH sales_by_quarter AS (
    SELECT 
        product_name,
        CONCAT('Q', QUARTER(sale_date)) as quarter,
        SUM(amount) as total_amount
    FROM sales_transactions
    GROUP BY 
        product_name,
        quarter
)
SELECT *      
FROM sales_by_quarter
PIVOT(
    SUM(total_amount)
    FOR quarter IN ('Q1', 'Q2', 'Q3', 'Q4')
) 
ORDER BY product_name;
```

georgesittas:
This needs to be included in the testing suite, look at the qualify_tables.sql fixture. I'll take a look soon.

georgesittas:
Ok I see the issue, I don't think this approach is correct. I'll follow up with a fix on a separate PR. Thanks for flagging.

PR #4768, opened 2025-02-19
feat(lineage): Add UNPIVOT support and improve column source tracking

zeta9044:
# Add UNPIVOT support and improve column source tracking

This PR introduces two major enhancements to the lineage module:

1. Add UNPIVOT operation support
- Implement complete UNPIVOT operation handling in lineage tracking
- Add separate handling for value and name columns in UNPIVOT
- Introduce new mapping structures for UNPIVOT operations
- Handle column transformation for UNPIVOT scenarios

2. Improve source column tracking
- Prioritize table.column notation in source tracking
- Implement hierarchical approach for column source detection
- Add fallback mechanism for column reference resolution
- Enhance accuracy of column lineage tracking

### UNPIVOT Support

```sql
-- Original table structure
CREATE TABLE sales (
    product_id INT,
    q1_sales INT,
    q2_sales INT,
    q3_sales INT,
    q4_sales INT
);

-- Query using UNPIVOT
SELECT *
FROM sales
UNPIVOT (
    sales_amount FOR quarter IN (
        q1_sales,
        q2_sales,
        q3_sales,
        q4_sales
    )
) AS sales_by_quarter;
```
Before this change, the lineage tracking would not correctly identify the source columns for the sales_amount column. Now it correctly tracks that sales_amount comes from all quarterly sales columns.
Example lineage output for sales_amount:
```
sales_amount
├── q1_sales
├── q2_sales
├── q3_sales
└── q4_sales
```
## Improved Source Column Tracking
```sql
-- Example with table.column notation
WITH transformed_data AS (
    SELECT 
        orders.order_id,
        customers.customer_name,
        products.product_name
    FROM orders
    JOIN customers ON orders.customer_id = customers.id
    JOIN products ON orders.product_id = products.id
)
SELECT * FROM transformed_data;
```
The improved source tracking now better handles the table.column notation, providing more accurate lineage information.

# Testing

Added new test cases covering:

1. UNPIVOT Scenarios:
* Basic UNPIVOT operations
* Multiple value columns
* Complex UNPIVOT with conditions
* Nested UNPIVOT operations

2. Source Column Tracking:
* Table.column notation
* Mixed notation scenarios
* Nested queries with various notation styles

#Test results:
```python
def test_unpivot_lineage():
    sql = """
        SELECT sales_amount
        FROM sales
        UNPIVOT (
            sales_amount FOR quarter IN (
                q1_sales, q2_sales, q3_sales, q4_sales
            )
        ) AS sales_by_quarter
    """
    result = lineage("sales_amount", sql)
    assert len(result.downstream) == 4
    assert all(col in [node.name for node in result.downstream] 
              for col in ["q1_sales", "q2_sales", "q3_sales", "q4_sales"])

def test_table_column_notation():
    sql = """
        SELECT orders.order_id
        FROM orders
    """
    result = lineage("order_id", sql)
    assert result.downstream[0].name == "orders.order_id"
```
All tests are passing with 100% coverage for the new features.

### DIFF
--- a/sqlglot/lineage.py
+++ b/sqlglot/lineage.py
@@ -235,7 +235,36 @@ def to_node(
             )
 
     # Find all columns that went into creating this one to list their lineage nodes.
-    source_columns = set(find_all_in_scope(select, exp.Column))
+    # Purpose: This code extracts source column lineage information from a SQL select statement
+
+    # Step 1: Check for dot notation references (e.g., table.column)
+    # - Use find_all_in_scope() to search for Dot expressions in the select statement
+    # - Store results in source_columns list
+    source_columns: set[exp.Column] = set(find_all_in_scope(select, exp.Dot))
+
+    if source_columns:
+        # Step 2: If dot notation columns found, convert them to Column objects
+        # - Filter to only include cases where dot.this is a Column
+        # - Create new Column objects with:
+        #   * this = Identifier with the column name from dot.expression.name
+        #   * table = original table reference from dot.this.table
+        source_columns = set(
+            exp.Column(this=exp.Identifier(this=dot.expression.name), table=dot.this.table)
+            for dot in source_columns
+            if isinstance(dot.this, exp.Column)
+        )
+
+        if not source_columns:
+            # Step 3: Find all Column references within scope
+            # - Search again to catch any remaining Column objects
+            source_columns = set(find_all_in_scope(select, exp.Column))
+    else:
+        # Step 4: Fallback - if no dot notation found
+        # - Directly search for all Column objects in scope
+        source_columns = set(find_all_in_scope(select, exp.Column))
+
+    # The final source_columns list contains all source columns that contribute
+    # to the target column's lineage, either from dot notation or direct column references
 
     # If the source is a UDTF find columns used in the UTDF to generate the table
     if isinstance(source, exp.UDTF):
@@ -255,17 +284,9 @@ def to_node(
     }
 
     pivots = scope.pivots
-    pivot = pivots[0] if len(pivots) == 1 and not pivots[0].unpivot else None
-    if pivot:
-        # For each aggregation function, the pivot creates a new column for each field in category
-        # combined with the aggfunc. So the columns parsed have this order: cat_a_value_sum, cat_a,
-        # b_value_sum, b. Because of this step wise manner the aggfunc 'sum(value) as value_sum'
-        # belongs to the column indices 0, 2, and the aggfunc 'max(price)' without an alias belongs
-        # to the column indices 1, 3. Here, only the columns used in the aggregations are of interest
-        # in the lineage, so lookup the pivot column name by index and map that with the columns used
-        # in the aggregation.
-        #
-        # Example: PIVOT (SUM(value) AS value_sum, MAX(price)) FOR category IN ('a' AS cat_a, 'b')
+    pivot = pivots[0] if len(pivots) == 1 else None
+    if pivot and not pivot.unpivot:
+        # Existing PIVOT processing logic
         pivot_columns = pivot.args["columns"]
         pivot_aggs_count = len(pivot.expressions)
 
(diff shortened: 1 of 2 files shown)


georgesittas:
@zeta9044 are you planning to add tests for this?

zeta9044:
> @zeta9044 are you planning to add tests for this?

Yes, I will add testsuite.

zeta9044:
> > @zeta9044 are you planning to add tests for this?
> 
> Yes, I will add testsuite.
One preliminary matter needs to be addressed. The changes I submitted in the PR for **qulify_tables.py**—specifically, ensuring that normalized db and schema are not appended to the CTE—are necessary. In this PR update, I will include the modifications to **qulify_tables.py** and also write a testsuite. Do you agree?

georgesittas:
I already merged the alternative fix for it and pinged you in it: https://github.com/tobymao/sqlglot/pull/4775

zeta9044:
> I already merged the alternative fix for it and pinged you in it: #4775. You need to rebase off of main.

thank,you.

zeta9044:
I completed adding testsuites for unpivot/pivot.

sqlglot/lineage.py
```diff
@@ -235,7 +235,36 @@ def to_node(
...
+    # - Store results in source_columns list
+    source_columns: set[exp.Column] = set(find_all_in_scope(select, exp.Dot))
+
+    if source_columns:
+        # Step 2: If dot notation columns found, convert them to Column objects
+        # - Filter to only include cases where dot.this is a Column
+        # - Create new Column objects with:
+        #   * this = Identifier with the column name from dot.expression.name
+        #   * table = original table reference from dot.this.table
+        source_columns = set(
```
tobymao, line 251: this doesn't seem right, why would we only look for dots? what if there's a mix?

sqlglot/lineage.py
```diff
@@ -275,11 +296,65 @@ def to_node(
...
+        unpivot_source_columns = field_info.expressions if field_info else []
+        unpivot_column_mapping = {}
+
+        # Value column mapping
+        value_column_mapping = {}
+        for value_column in value_columns:
+            if column == value_column.name:
+                table_name = pivot.parent.this.name
+                value_column_mapping[value_column.name] = [
+                    exp.Column(this=col.this, table=exp.to_identifier(table_name))
```
tobymao, line 315: use exp.column() in general instead of the Expr type

sqlglot/lineage.py
zeta9044: > this doesn't seem right, why would we only look for dots? what if there's a mix?

I'm specifically addressing lineage tracking for UNPIVOT operations. When a query like:

```
select unpvt.id, unpvt.month, unpvt.sales from sales_data UNPIVOT (sales for month in (jan_sales, feb_sales)) unpvt
```

gets fully qualified, it becomes:

```
select sales_data.unpvt.id, sales_data.unpvt.month, sales_data.unpvt.sales from sales_data UNPIVOT (...)
```

The special case here is that columns like 'month' and 'sales' are derived from the UNPIVOT operation itself, not directly from the base table. When tracking lineage, these columns must be associated with the UNPIVOT alias ('unpvt') rather than the original table.

That's why I'm focusing on dot expressions - to correctly identify cases where a column is being accessed through an UNPIVOT alias. Without this specific handling, we'd lose the connection to UNPIVOT-derived columns in the lineage tracking.

sqlglot/lineage.py
zeta9044: > use exp.column() in general instead of the Expr type

I changed and pull requests of your wants

sqlglot/lineage.py
georgesittas: Something seems off about this, @zeta9044.

What you're describing is probably best handled by the scope module. The `Dot(...)` over `Column(this=..., table=unpvt)` discrepancy feels like a red herring; I'm worried the current approach only patches this for the `lineage`, hiding the problematic aspects.

I'd start by answering the question "why is `sales_data` being treated as the source of `unpvt`, when the latter itself is also a source in the current scope?"

tobymao:
@zeta9044 i'm not going to be able to merge / make progress on this without fully understanding the motivation behind the dot change. it doesn't seem correct to me.

zeta9044:
@tobymao 
Here's the explanation in English for tobymao:

Currently, in the `to_node` function in lineage.py:

```python
# Find all columns that went into creating this one to list their lineage nodes.
# Purpose: This code extracts source column lineage information from a SQL select statement
source_columns = set(find_all_in_scope(select, exp.Column))
```

The issue is that the `source_columns` found by this code don't distinguish whether a column comes from an original table or from a derived source like unpivot/pivot/subquery/cte. This happens because column names appear with table aliases, which is why we need to search for DOT expressions.

If `find_all_in_scope` had identified the original tables and columns within unpivot/pivot/subquery/cte aliases, we wouldn't need to search for DOT expressions. Additionally, in the following code:

```python
for c in source_columns:
    table = c.table
    source = scope.sources.get(table)  # Returns the alias (table) of unpivot/pivot/subquery/cte
```

If `table` is an alias for unpivot/pivot/subquery/cte, and if `source` had been returned as a scope, then we wouldn't need to search for DOT expressions or parse subquery/cte expressions.

Since `find_all_in_scope` is a common function provided by sqlglot, modifying it could have impacts on other modules that I can't be certain about. That's why I proceeded with the DOT expression search approach.

**"If this PR is to be merged, could you review whether I should modify find_all_in_scope and submit those changes?"**

tobymao:
yes, i'd rather fix scope first. maybe let's close this one for now, and fix scope first, what do you think?

Issue #4762, opened 2025-02-18
Failed to parse SQL (postgres) with commented out semicolon


dbittenbender:
```sql
-- COMMENT
-------------------------------------------------------------------------------
TRUNCATE table1
;


-- COMMENT
-------------------------------------------------------------------------------
WITH
    table5 AS (
        SELECT DISTINCT
            col1,
            col2,
            col3
        FROM
            table6 cdep
    ),
    table4 AS (
        SELECT
            col1,
            col2,
            col3 --
        FROM
            table5 ltt
    ),
    table3 AS (
        SELECT
            col1,
            col2,
            col3
        FROM
            table4
    ),
    table2 AS (
        SELECT
            col1,
            col2,
            col3
        FROM
            table3
    )
INSERT INTO
    table1 (
        col1,
        col2,
        col3
    )
SELECT
    col1,
    col2,
    col3
FROM
    table2
;

-- vacuum after
-- VACUUM VERBOSE
-- ANALYZE table1
-- ;
```

The above SQL is a simplified example as I can't post client code.  However, it fails the same way.

I am using read dialect of postgres.

I am first calling `parse(sql_stmt, read=dialect)` with the entire statement expecting to get all of the statements returned as a List.  That returns 3 sub-statements: TruncateTable, Insert, and Semicolon.  The first 2 are correct.  However, the entire 3rd statement (Semicolon) is the commented out lines at the bottom of the full sql statement.

Then I try to parse each one of the sub-statements: `ast = parse_one(sql_stmt, read=dialect)`.  When I get to the 3rd statement (Semicolon), it fails with:

`ParseError("No expression was parsed from '/* vacuum after */ /* VACUUM VERBOSE */ /* ANALYZE table1 */ /* ; */'")`

It seems like when it parses the end of the file, it get's thrown off by the commented out semicolon.

dbittenbender:
Also seeing a ton of log messages like this:

INFO:sqlglot:Applying array index offset (1)
INFO:sqlglot:Applying array index offset (1)
INFO:sqlglot:Applying array index offset (1)
INFO:sqlglot:Applying array index offset (-1)
INFO:sqlglot:Applying array index offset (-1)
INFO:sqlglot:Applying array index offset (-1)

Not sure where it's coming from or if it's relevant

VaggelisD:
Hey @dbittenbender,

Regarding the comments, this is expected and is happening for the following reasons:
[...]

PR #4772, opened 2025-02-19
fix(parser): Do not preemptively create chunks for comment only statements

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/4762

Before this PR:

```Python3
>>> import sqlglot
>>> sql = """
... -- vacuum after
... -- VACUUM VERBOSE
... -- ANALYZE table1
... -- ;
... ;
... """
>>>
>>> ast = sqlglot.parse_one(sql).sql()
sqlglot.errors.ParseError: No expression was parsed from ...
```

After this PR

```Python3
>>> sqlglot.parse_one(sql).sql()
'/* vacuum after */ /* VACUUM VERBOSE */ /* ANALYZE table1 */ /* ; */'
```


georgesittas (approved):
I would do a sanity check that this doesn't break SQLMesh, but LGTM otherwise.

VaggelisD:
> I would do a sanity check that this doesn't break SQLMesh, but LGTM otherwise.

Tests pass

PR #4779, opened 2025-02-20
Chore: remove redundant "parts" arg from CH ast, move gen methods to base

georgesittas:
cc @pkit, check out https://github.com/TobikoData/sqlmesh/pull/3877 for context. We should be more careful and make sure `Expression` subclasses can always be generated in the base dialect.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -82,7 +82,7 @@ def _build_count_if(args: t.List) -> exp.CountIf | exp.CombinedAggFunc:
     if len(args) == 1:
         return exp.CountIf(this=seq_get(args, 0))
 
-    return exp.CombinedAggFunc(this="countIf", expressions=args, parts=("count", "If"))
+    return exp.CombinedAggFunc(this="countIf", expressions=args)
 
 
 def _build_str_to_date(args: t.List) -> exp.Cast | exp.Anonymous:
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -743,7 +743,6 @@ def _parse_function(
                     "expressions": anon_func.expressions,
                 }
                 if parts[1]:
-                    kwargs["parts"] = parts
```
georgesittas, line 746: This isn't used after parsing, it's only used here to determine the type of the produced node.

tobymao (approved):

pkit:
`parts` were added to make it possible to build combinator functions. Now I need to re-parse it again.
I.e. it's not clear how to convert `quantileState` to `quantileMerge` now.
Is it just a problem of tuple as data type?

georgesittas:
I see, so you leverage `parts` to transpile these combined functions? Can you show an example? I think we should've added a test for this, it wasn't clear and so since it's essentially dead code for sqlglot it made sense to remove it.

pkit:
Something like:
```
        if isinstance(node, exp.ParameterizedAgg):
            return exp.CombinedParameterizedAgg(
                this=f"{node.this}State",
                expressions=node.expressions,
                params=node.args["params"],
                parts=(node.this, "State"),
            )
```

georgesittas:
Why is this necessary? You're already providing the parts in `this`. A similar example that is also tested in sqlglot is here: https://github.com/tobymao/sqlglot/blob/4bf897da6332175c22b772beb0b0cb6e6e941ebb/sqlglot/dialects/clickhouse.py#L81-L85.

pkit:
I can provide parts in `this` only because it was not a combined agg func originally.
I.e. `count -> countIf` is trivial. But `countState -> countMerge` is not.

georgesittas:
Can you show me the code that does the conversion of `countState` to `countMerge`? I think there may be a disconnect somewhere, still not sure I see the issue.

PR #4784, opened 2025-02-21
Feat(clickhouse): add support for dynamic json casting

### LINKED ISSUE #4776: clickhouse: support the new json dynamic type resolution syntax
There's a new syntax `.:` that works like this:
```
create table t (data JSON) engine = MergeTree order by tuple()

insert into t values ('{"a": {"b": "foo"}}')
insert into t values ('{"a": {"b": 10}}')

select data.a.b.:Int32 from t
select data.a.b.:String from t
```

I will try to get a PR soon.

georgesittas:
Fixes #4776

cc @pkit

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3518,6 +3518,9 @@ class Generator(metaclass=_Generator):
     def trycast_sql(self, expression: exp.TryCast) -> str:
         return self.cast_sql(expression, safe_prefix="TRY_")
 
+    def jsoncast_sql(self, expression: exp.JSONCast) -> str:
+        return self.cast_sql(expression)
+
     def try_sql(self, expression: exp.Try) -> str:
         if not self.TRY_SUPPORTED:
             self.unsupported("Unsupported TRY function")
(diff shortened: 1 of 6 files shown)


sqlglot/generator.py
```diff
@@ -3518,6 +3518,9 @@ def sub_sql(self, expression: exp.Sub) -> str:
     def trycast_sql(self, expression: exp.TryCast) -> str:
         return self.cast_sql(expression, safe_prefix="TRY_")
 
+    def jsoncast_sql(self, expression: exp.JSONCast) -> str:
+        return self.cast_sql(expression)
```
georgesittas, line 3522: I put this here to allow a best-effort transpilation. I'm not sure, but perhaps this could allow CH ~> Snowflake. The latter has a `VARIANT` type that kind of matches `Dynamic`.

pkit: CH has both Variant and Dynamic. :)
Variant has a strict list of types. While Dynamic just accepts anything and adds it to the list.

pkit:
Nice! Thanks

PR #4787, opened 2025-02-23
feat(snowflake)!: show databases, functions, procedures and warehouses 

tekumara:
Parse as expression, rather than command. These are issued by the Snowflake JDBC driver.

Refs:

https://docs.snowflake.com/en/sql-reference/sql/show-databases
https://docs.snowflake.com/en/sql-reference/sql/show-functions
https://docs.snowflake.com/en/sql-reference/sql/show-procedures
https://docs.snowflake.com/en/sql-reference/sql/show-warehouses

NB: `show databases in account` is valid syntax, although its not mentioned above. See the Snowflake JDBC driver's usage [here](https://github.com/snowflakedb/snowflake-jdbc/blob/dcb60b2905ca512c417a90ab08a87305e31e486c/src/main/java/net/snowflake/client/jdbc/SnowflakeDatabaseMetaData.java#L1648).

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -520,6 +520,8 @@ class Snowflake(Dialect):
         }
 
         SHOW_PARSERS = {
+            "DATABASES": _show_parser("DATABASES"),
+            "TERSE DATABASES": _show_parser("DATABASES"),
             "SCHEMAS": _show_parser("SCHEMAS"),
             "TERSE SCHEMAS": _show_parser("SCHEMAS"),
             "OBJECTS": _show_parser("OBJECTS"),
@@ -539,6 +541,8 @@ class Snowflake(Dialect):
             "COLUMNS": _show_parser("COLUMNS"),
             "USERS": _show_parser("USERS"),
             "TERSE USERS": _show_parser("USERS"),
+            "FUNCTIONS": _show_parser("FUNCTIONS"),
+            "PROCEDURES": _show_parser("PROCEDURES"),
         }
 
         CONSTRAINT_PARSERS = {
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -520,6 +520,8 @@ class Parser(parser.Parser):
         }
 
         SHOW_PARSERS = {
+            "DATABASES": _show_parser("DATABASES"),
+            "TERSE DATABASES": _show_parser("DATABASES"),
```
georgesittas, line 524: I think you need to update `_parse_show_snowflake`, because this doesn't handle the following clause:

```
[ WITH PRIVILEGES <object_privilege> [ , <object_privilege> [ , ... ] ] ]
```

Today this is included in the `Command` string, so there's no parsing issue. Let's make sure this is tested as well.

sqlglot/dialects/snowflake.py
```diff
@@ -539,6 +541,9 @@ class Parser(parser.Parser):
             "COLUMNS": _show_parser("COLUMNS"),
             "USERS": _show_parser("USERS"),
             "TERSE USERS": _show_parser("USERS"),
+            "FUNCTIONS": _show_parser("FUNCTIONS"),
+            "PROCEDURES": _show_parser("PROCEDURES"),
+            "WAREHOUSES": _show_parser("WAREHOUSES"),
```
georgesittas, line 546: Ditto, these clauses don't look supported:

```
[ IN CLASS <class_name> ] 
[ IN APPLICATION <application_name> | APPLICATION PACKAGE <application_package_name> ]
```

georgesittas (commented):
Thanks for the PR @tekumara– left a couple of comments.

georgesittas:
I'll take this to the finish line, thanks.

tekumara:
Thank you @georgesittas!

PR #4794, opened 2025-02-24
Feature(athena): add `SHOW` handling

theborch:
At the risk of overstaying my welcome on this PR, and apologies for the added review workload, I stripped down to the bare minimum additions to get `SHOW` functional and tested.

-----

Parse the following Athena statements as `SHOW`:

[SHOW COLUMNS](https://docs.aws.amazon.com/athena/latest/ug/show-columns.html)
[SHOW CREATE TABLE](https://docs.aws.amazon.com/athena/latest/ug/show-create-table.html)
[SHOW CREATE VIEW](https://docs.aws.amazon.com/athena/latest/ug/show-create-view.html)
[SHOW DATABASES](https://docs.aws.amazon.com/athena/latest/ug/show-databases.html)
[SHOW PARTITIONS](https://docs.aws.amazon.com/athena/latest/ug/show-partitions.html)
[SHOW TABLES](https://docs.aws.amazon.com/athena/latest/ug/show-tables.html)
[SHOW TBLPROPERTIES](https://docs.aws.amazon.com/athena/latest/ug/show-tblproperties.html)
[SHOW VIEWS](https://docs.aws.amazon.com/athena/latest/ug/show-views.html)

Example:

```py
from sqlglot import parse_one

parse_one("show create table db.table;", dialect="athena")
```

Output:

```py
Show(
  this=CREATE TABLE,
  target=Identifier(this=table, quoted=False),
  db=Identifier(this=db, quoted=False))
```

Notes:
1. if using a regular expression in `SHOW TABLES` you will need prefix with `LIKE` to parse as a `SHOW`
    - this is valid, despite missing from the current docs.
2. `SHOW TBLPROPERTIES` does not currently parse optional `('property_name')`
    - will parse as `Command`


georgesittas:
Thank you for cleaning this up, I appreciate it. I'll try to take a look again soon 👍

sqlglot/dialects/athena.py
```diff
@@ -122,8 +145,61 @@ class Parser(Trino.Parser):
...
+            if self._match(TokenType.FROM) or self._match(TokenType.IN):
+                db = self._parse_id_var()
+            elif self._match(TokenType.DOT):
+                db = target_id
+                target_id = self._parse_id_var()
+
+            # using `rlike` for the specific commands where athena uses regex in a `LIKE`
+            like = (self._match_text_seq("LIKE") and self._parse_string()) or None
+            rlike = None
+            if this in ["DATABASES", "SCHEMAS", "TABLES", "VIEWS"]:
```
georgesittas, line 187: When a container has ≥3 constants, we usually factor it out into a class constant (set in this case).

sqlglot/dialects/athena.py
```diff
@@ -122,8 +145,61 @@ class Parser(Trino.Parser):
...
+
+            if self._match(TokenType.FROM) or self._match(TokenType.IN):
+                db = self._parse_id_var()
+            elif self._match(TokenType.DOT):
+                db = target_id
+                target_id = self._parse_id_var()
+
+            # using `rlike` for the specific commands where athena uses regex in a `LIKE`
+            like = (self._match_text_seq("LIKE") and self._parse_string()) or None
+            rlike = None
```
georgesittas, line 186: Why is this necessary? Can't we just leave the `Like` node as is without introducing a new argument in the AST?

sqlglot/dialects/athena.py
```diff
@@ -165,3 +241,21 @@ def generate(self, expression: exp.Expression, copy: bool = True) -> str:
                 return self._hive_generator.generate(expression, copy)
 
             return super().generate(expression, copy)
+
+        def _prefixed_sql(self, prefix: str, expression: exp.Expression, arg: str) -> str:
+            sql = self.sql(expression, arg)
+            return f"{prefix} {sql}" if sql else ""
+
+        def show_sql(self, expression: exp.Show) -> str:
+            this = f"{expression.name}"
```
georgesittas, line 250: No need for f-string.

sqlglot/dialects/athena.py
```diff
@@ -165,3 +241,21 @@ def generate(self, expression: exp.Expression, copy: bool = True) -> str:
...
+            db = self.sql(expression, "db") or ""
+            target = self.sql(expression, "target") or ""
+            if db and target:
+                target = f"{db}.{target}"
+                db = ""
```
georgesittas, line 255: Why are we manually building the FQN here? This suggests that the parsing may be off. We should be able to produce a `Table` or whatever and generate it using `self.sql` without needing manual logic for it.

sqlglot/dialects/athena.py
```diff
@@ -165,3 +241,21 @@ def generate(self, expression: exp.Expression, copy: bool = True) -> str:
...
+            if target and this in ["COLUMNS", "TABLES", "VIEWS"]:
```
georgesittas, line 256: Ditto about constant.

sqlglot/dialects/athena.py
```diff
@@ -165,3 +241,21 @@ def generate(self, expression: exp.Expression, copy: bool = True) -> str:
...
+                target = f"IN {target}"
+            like = self._prefixed_sql("LIKE", expression, "like")
+            rlike = self._prefixed_sql("LIKE", expression, "rlike")
```
georgesittas, line 259: It doesn't look like this is needed?

georgesittas:
Hey @theborch, any plans to finish this?

georgesittas:
Closing this for now, will reopen when an update is provided.

PR #4798, opened 2025-02-25
feat(clickhouse): support `GLOBAL NOT IN`

hdhoang:
Clickhouse has `GLOBAL NOT IN` variant https://clickhouse.com/docs/sql-reference/operators/in. Superset 4.1 is conservatively refusing a `GLOBAL NOT IN` query as not-read-only.

I parsed the keyword into a new `exp.In` flag `is_negated`, but I'm not sure if a richer datatype for `is_global` make sense.

Hope this helps.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -479,8 +479,7 @@ class ClickHouse(Dialect):
 
         RANGE_PARSERS = {
             **parser.Parser.RANGE_PARSERS,
-            TokenType.GLOBAL: lambda self, this: self._match(TokenType.IN)
-            and self._parse_in(this, is_global=True),
+            TokenType.GLOBAL: lambda self, this: self._parse_global_in(this),
         }
 
         # The PLACEHOLDER entry is popped because 1) it doesn't affect Clickhouse (it corresponds to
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -629,11 +628,23 @@ def _parse_bracket(
...
-        def _parse_in(self, this: t.Optional[exp.Expression], is_global: bool = False) -> exp.In:
+        def _parse_in(
+            self,
+            this: t.Optional[exp.Expression],
+            is_global: bool = False,
+            is_negated: bool = False,
+        ) -> exp.In:
             this = super()._parse_in(this)
             this.set("is_global", is_global)
+            this.set("is_negated", is_negated)
```
georgesittas, line 639: This isn't correct, we shouldn't be setting args that don't appear in `arg_types`.

I'm not sure what the best approach is here... Perhaps we can normalize this into: `Not(In(..., is_global=True))` and then override ClickHouse's generator methods to properly generate this.

hdhoang: Thanks, I switched to that & tried `str.replace(..., 1)`. Is there a saner way to replace the syntax?

georgesittas (approved):

sqlglot/dialects/clickhouse.py
georgesittas: Looks sane enough to me :)

Thanks for addressing this!

PR #4800, opened 2025-02-25
Support MySQL's `ALTER INDEX`.

burnison:
MySQL 8.0 and above [support altering an existing index](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html), `x`, using the syntax, `ALTER TABLE x ALTER INDEX {VISIBLE | INVISIBLE}`. This allows for the index to become either invisible _or_ visible. No other index changes are permitted. This commit adds support for such syntax.

As I am still learning how this codebase works, I have done my best to put the implementations into the correct places. I also wasn't positive if this should leverage `IndexOptionConstraints`; however, I decided not to use this parser, as the full set of options is _not_ supported  by MySQL's `ALTER INDEX`. Please provide feedback if I have chosen a suboptimal implementation.

### DIFF
--- a/sqlglot/dialects/mysql.py
+++ b/sqlglot/dialects/mysql.py
@@ -439,6 +439,11 @@ class MySQL(Dialect):
             "MODIFY": lambda self: self._parse_alter_table_alter(),
         }
 
+        ALTER_ALTER_PARSERS = {
+            **parser.Parser.ALTER_ALTER_PARSERS,
+            "INDEX": lambda self: self._parse_alter_table_alter_index(),
+        }
+
         SCHEMA_UNNAMED_CONSTRAINTS = {
             *parser.Parser.SCHEMA_UNNAMED_CONSTRAINTS,
             "FULLTEXT",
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/mysql.py
```diff
@@ -693,6 +698,17 @@ def _parse_json_value(self) -> exp.JSONValue:
...
             )
 
+        def _parse_alter_table_alter_index(self) -> t.Optional[exp.Expression]:
+            index = self._parse_field(any_token=True)
+
+            visible = None
+            if self._match_text_seq("VISIBLE"):
+                visible = True
+            elif self._match_text_seq("INVISIBLE"):
+                visible = False
```
georgesittas, line 708:
```suggestion
            if self._match_text_seq("VISIBLE"):
                visible = True
            elif self._match_text_seq("INVISIBLE"):
                visible = False
            else:
                visible = None
```

sqlglot/dialects/mysql.py
```diff
@@ -693,6 +698,17 @@ def _parse_json_value(self) -> exp.JSONValue:
                 on_condition=self._parse_on_condition(),
             )
 
+        def _parse_alter_table_alter_index(self) -> t.Optional[exp.Expression]:
```
georgesittas, line 701:
```suggestion
        def _parse_alter_table_alter_index(self) -> exp.AlterIndex:
```

sqlglot/expressions.py
```diff
@@ -1706,6 +1706,14 @@ class AlterColumn(Expression):
     }
 
 
+# https://dev.mysql.com/doc/refman/8.0/en/invisible-indexes.html
+class AlterIndex(Expression):
+    arg_types = {
+        "this": True,
+        "visible": True,
+    }
```
georgesittas, line 1714:
```suggestion
    arg_types = {"this": True, "visible": True}
```

PR #4801, opened 2025-02-25
fix(tsql)!: support TOP with PERCENT and WITH TIES

### LINKED ISSUE #4793: SELECT TOP 10 PERCENT doesn't  parse
In T-SQL you can specify PERCENT in TOP expression, e.g. `SELECT TOP  10  PERCENT ...<rest of the query>`.

Here is a [documentation link](https://learn.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql?view=sql-server-ver16) for this.

This currently doesn't parse in SQLGlot, either with dialect="tsql" or with no dialect specified.

The simplest query demonstrating this is `SELECT 10 PERCENT 1;`. 

Thank you for an absolutely wonderful library, BTW - truly a pleasure to work with!

geooo109:
Fixes #4793

This PR adds support for using `PERCENT` and `WITH TIES` alongside the `TOP` command in T-SQL. 

Moreover, the base parsing method `limit` and the base generator method `fetch` have been refactored.

**DOCS**
[T-SQL TOP PERCENT/WITH TIES](https://learn.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql?view=sql-server-ver16#syntax)

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -4469,10 +4469,8 @@ class Parser(metaclass=_Parser):
     def _parse_limit_options(self) -> exp.LimitOptions:
         percent = self._match(TokenType.PERCENT)
         rows = self._match_set((TokenType.ROW, TokenType.ROWS))
-        only = self._match_text_seq("ONLY")
+        self._match_text_seq("ONLY")
         with_ties = self._match_text_seq("WITH", "TIES")
-        if only and with_ties:
-            self.raise_error("Cannot specify both ONLY and WITH TIES in FETCH clause")
         return self.expression(exp.LimitOptions, percent=percent, rows=rows, with_ties=with_ties)
 
     def _parse_limit(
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -4474,12 +4483,15 @@ def _parse_limit(
     ) -> t.Optional[exp.Expression]:
         if skip_limit_token or self._match(TokenType.TOP if top else TokenType.LIMIT):
             comments = self._prev_comments
+            limit_options = None
```
georgesittas, line 4486: Move this to the `else` branch.

sqlglot/parser.py
```diff
@@ -4466,6 +4466,15 @@ def _parse_ordered(
             exp.Ordered, this=this, desc=desc, nulls_first=nulls_first, with_fill=with_fill
         )
 
+    def _parse_limit_options(self) -> exp.LimitOptions:
+        percent = self._match(TokenType.PERCENT)
+        rows = self._match_set((TokenType.ROW, TokenType.ROWS))
+        only = self._match_text_seq("ONLY")
+        with_ties = self._match_text_seq("WITH", "TIES")
+        if only and with_ties:
+            self.raise_error("Cannot specify both ONLY and WITH TIES in FETCH clause")
```
georgesittas, line 4475: Probably not necessary to do error-checking here, i.e. we can remove L4472, L4474, L4475.

georgesittas (commented):
Looks good. FYI, this is breaking.

georgesittas (approved):

PR #4805, opened 2025-02-26
Feat(trino): add support for ON ... ERROR/NULL syntax for JSON_QUERY

georgesittas:
Fixes https://github.com/TobikoData/sqlmesh/issues/3905

This doesn't cover the full syntax spec yet, just the `... ON {EMPTY|ERROR}` clause.

Reference: https://trino.io/docs/current/functions/json.html#json-query

### DIFF
--- a/sqlglot/dialects/presto.py
+++ b/sqlglot/dialects/presto.py
@@ -401,7 +401,6 @@ class Presto(Dialect):
             exp.If: if_sql(),
             exp.ILike: no_ilike_sql,
             exp.Initcap: _initcap_sql,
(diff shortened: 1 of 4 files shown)


VaggelisD (approved):
LGTM 👍 

PS: It looks like Trino's JSON functions reuse some of these options (`PASSING`, the on condition etc), I wonder if it makes sense to aggregate them in a single expression at the end.

georgesittas:
> LGTM 👍
> 
> PS: It looks like Trino's JSON functions reuse some of these options (`PASSING`, the on condition etc), I wonder if it makes sense to aggregate them in a single expression at the end.

Yeah, can probably refactor some of this logic when we do another pass at this.

PR #4808, opened 2025-02-26
fix: orphan node(s) in eliminate_join_marks

snovik75:
Currently there is a problem if a join with join marks has multiple conditions because a new join is created every time, there is a column with a join mark. This reset the parent of the joined table and the link is lost which breaks the tree consistency down

Example:

```
s = "select a.id from a, b where a.id = b.id (+) and b.d (+) = const"

tree = parse_one(s, dialect="oracle")
print("before", [type(t.parent_select) for t in tree.find_all(exp.Table)])
eliminate_join_marks(tree)
print("after", [type(t.parent_select) for t in tree.find_all(exp.Table)])
```

outputs

```
before [<class 'sqlglot.expressions.Select'>, <class 'sqlglot.expressions.Select'>]
after [<class 'sqlglot.expressions.Select'>, <class 'NoneType'>]
```

against expected 

```
before [<class 'sqlglot.expressions.Select'>, <class 'sqlglot.expressions.Select'>]
after [<class 'sqlglot.expressions.Select'>, <class 'sqlglot.expressions.Select'>]
```

This fix makes sure that a Join is created only once, if it is new.

Testing: added a unit test (not sure if it helps much)

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -907,16 +907,15 @@ def eliminate_join_marks(expression: exp.Expression) -> exp.Expression:
                 len(marked_column_tables) == 1
             ), "Columns of only a single table can be marked with (+) in a given binary predicate"
 
+            # Add predicate if join already copied, or add join if it is new
             join_this = old_joins.get(col.table, query_from).this
-            new_join = exp.Join(this=join_this, on=join_predicate, kind="LEFT")
-
-            # Upsert new_join into new_joins dictionary
-            new_join_alias_or_name = new_join.alias_or_name
-            existing_join = new_joins.get(new_join_alias_or_name)
+            existing_join = new_joins.get(join_this.alias_or_name)
             if existing_join:
-                existing_join.set("on", exp.and_(existing_join.args.get("on"), new_join.args["on"]))
+                existing_join.set("on", exp.and_(existing_join.args["on"], join_predicate))
             else:
-                new_joins[new_join_alias_or_name] = new_join
+                new_joins[join_this.alias_or_name] = exp.Join(
+                    this=join_this, on=join_predicate, kind="LEFT"
+                )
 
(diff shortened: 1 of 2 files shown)


tests/test_transforms.py
```diff
@@ -256,3 +256,11 @@ def test_eliminate_join_marks(self):
...
                 dialect,
             )
+
+            # if multiple conditions, we check that after transformations the tree remains consistent
+            s = "select a.id from a, b where a.id = b.id (+) AND b.d (+) = const"
+            tree = parse_one(s, dialect=dialect)
+            assert ([type(t.parent_select) for t in tree.find_all(exp.Table)]) == [
+                exp.Select,
+                exp.Select,
+            ]
```
georgesittas, line 266: This isn't really testing your change, right? You need to show how the AST is changed _after_ the transformation.

sqlglot/transforms.py
```diff
@@ -907,16 +907,15 @@ def eliminate_join_marks(expression: exp.Expression) -> exp.Expression:
...
-            new_join_alias_or_name = new_join.alias_or_name
-            existing_join = new_joins.get(new_join_alias_or_name)
+            existing_join = new_joins.get(join_this.alias_or_name)
             if existing_join:
-                existing_join.set("on", exp.and_(existing_join.args.get("on"), new_join.args["on"]))
+                existing_join.set("on", exp.and_(existing_join.args["on"], join_predicate))
             else:
-                new_joins[new_join_alias_or_name] = new_join
+                new_joins[join_this.alias_or_name] = exp.Join(
+                    this=join_this, on=join_predicate, kind="LEFT"
```
georgesittas, line 917: I think we may need to copy both args here

georgesittas (approved):
LGTM, I will get this to the finish line. Thanks for the patch!

PR #4811, opened 2025-02-27
Chore(optimizer): refactor DISTINCT ON qualification to better match ORDER BY

georgesittas:
@Fosly


georgesittas:
Taking a look to see if we need to just use `_expand_order_by` so it's precisely the same logic.

PR #4817, opened 2025-02-28
fix: allow duplicate nodes in matchings

tobymao:


sqlglot/diff.py
```diff
@@ -108,44 +104,47 @@ def diff(
...
-        delta_only=delta_only,
-    )
-
-    if not copy:
-        for node in chain(source.walk(), target.walk()):
-            node._hash = None
+    try:
+        # We cache the hash of each new node here to speed up equality comparisons. If the input
+        # trees aren't copied, these hashes will be evicted before returning the edit script.
+        if copy:
```
georgesittas, line 130: Should this be `copy and matchings` to avoid unnecessarily calls to `compute_node_mappings`?

sqlglot/diff.py
```diff
@@ -108,44 +104,47 @@ def diff(
...
-        ):
-            # We cache the hash of each new node here to speed up equality comparisons. If the input
-            # trees aren't copied, these hashes will be evicted before returning the edit script.
+        for old_node, new_node in zip(reversed(old_nodes), reversed(new_nodes)):
             new_node._hash = hash(new_node)
-
-            old_node_id = id(old_node)
-            if old_node_id in matching_ids:
-                node_mapping[old_node_id] = new_node
+            node_mapping[id(old_node)] = new_node
```
georgesittas, line 114: I guess if the same node reference appears more than once in the original ASTs then this line would overwrite the first node in the `node_mapping` and it may silently cause problems.

This is probably true today as well, just pointing it out. It would be helpful if we warned, because I can imagine that debugging these scenarios in the future will be very gnarly. We could maybe avoid warning if the ids of the top-level nodes match, and just short-circuit by returning an empty list.

georgesittas (approved):

Issue #4813, opened 2025-02-27
Support parsing PUT/GET queries to upload/download files from stages (Snowflake)


whummer:
Hi folks! Currently, some queries related to stage operations in Snowflake are not yet supported and cannot be parsed properly in `sqlglot` (i.e., are getting parsed as `exp.Command` as a fallback).

* https://docs.snowflake.com/en/sql-reference/sql/put
* https://docs.snowflake.com/en/sql-reference/sql/get

Some examples of queries that should ideally be covered:
```
from sqlglot import parse_one

# basic example for a PUT statement
parse_one("PUT file:///tmp/my.txt @stage1/folder", dialect="snowflake")
# quoted stage reference
parse_one("PUT file:///tmp/my.txt @"db1"."stage1"/folder", dialect="snowflake")
# single quotes to support stage names or file paths containing whitespaces:
parse_one("PUT 'file:///tmp/my file.txt' '@stage1/folder/my file.txt'", dialect="snowflake")

# reverse examples to GET files from stages
parse_one("GET @stage1/folder file:///tmp/my.txt", dialect="snowflake")
parse_one("GET @"db1"."stage1"/folder file:///tmp/my.txt", dialect="snowflake")
parse_one("GET '@stage1/folder/my file.txt' 'file:///tmp/my file.txt', dialect="snowflake")
```

Would be great to parse these Snowflake query expressions into native subtypes of `exp.Expression` (and not into `exp.Command`).

Please let me know if these are on your roadmap - otherwise, happy to work towards a PR (and would appreciate some guidance in the process).. Thanks a lot for your support! 🙌

tobymao:
Not planned, feel free to make a PR. Thank you!

PR #4818, opened 2025-03-01
feat(Snowflake): add initial support for PUT statements

whummer:
Add initial support for [`PUT` statements](https://docs.snowflake.com/en/sql-reference/sql/put) in the Snowflake dialect. Addresses #4813

This PR adds the required token types and parser functions to enable parsing into instances of `exp.Put` expressions. Note: the `PUT` command can contain verbatim (unquoted) URL strings like `PUT file://my.txt ...`, hence a `URL` token type and a `_scan_url(..)` scanner method has been added - please let me know if that is appropriate.

**Note:** This PR is still work in progress, 🚧 and I'd love to get some feedback if the changes in here are going in the right direction. Any guidance would be highly appreciated. 🙌 

**Note:** The new logic _does not yet work_ with the Rust tokenizer. If the general direction of the PR is ok, then we'd need to adjust the Rust tokenizer as well - I would very much appreciate some help there, as my knowledge of Rust is quite limited. Can give it a try, but would likely need some help/guidance! 😊 

If and when this PR gets merged successfully, we can also create a follow-up PR to support parsing of `GET` (and potentially other stage-related statements) in Snowflake queries. 👍 

Thanks for your support!

**TODOs:**
- [x] ~adjust the logic in the Rust tokenizer, as appropriate~
- [x] ~remove temporary checks which disable tests when using Rust tokenizer~


sqlglot/tokens.py
```diff
@@ -1434,6 +1439,23 @@ def _scan_var(self) -> None:
...
         )
 
+    def _scan_url(self) -> bool:
+        start = self._current
+        end = self.size
+        for idx in range(start, self.size):
+            char = self.sql[idx]
+            if char.isspace() or char == ";":
+                end = idx
+                break
```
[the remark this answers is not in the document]
georgesittas: So, @whummer, could you revert this and add support only for the variant with the string URI?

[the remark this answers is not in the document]
whummer: Ok, thanks @georgesittas - will revert to the string URI variant only. 

Would have been awesome to get the parsed expression out of the box (as this may also apply to a couple of other statement types like `GET`, `COPY INTO`, etc) but it is also understandable that you'd like to limit the maintenance overhead at this point. I think your suggestion should work, we should be able to wrap the raw URLs into quotes at the application level. 👍 

Will update the PR shortly! (hope to get a chance later today..)

sqlglot/dialects/snowflake.py
```diff
@@ -805,6 +806,13 @@ def _parse_show_snowflake(self, this: str) -> exp.Show:
                 },
             )
 
+        def _parse_put(self) -> exp.Put:
+            self._match(TokenType.PUT)
```
georgesittas, line 810: I think this is redundant, we should've already consumed the token before invoking the statement parser.

sqlglot/dialects/snowflake.py
```diff
@@ -805,6 +806,13 @@ def _parse_show_snowflake(self, this: str) -> exp.Show:
...
+            source = self._parse_location_path()
+            target = self._parse_location_path()
+            props = self._parse_properties()
+            return self.expression(exp.Put, this=source, target=target, properties=props)
```
georgesittas, line 814: Let's make sure that we'll fall back to producing a `Command` if the token following `PUT` is *not* a string. We don't wanna break existing `PUT` statements.

sqlglot/dialects/snowflake.py
```diff
@@ -1277,3 +1285,8 @@ def select_sql(self, expression: exp.Select) -> str:
             if offset and not limit:
                 expression.limit(exp.Null(), copy=False)
             return super().select_sql(expression)
+
+        def put_sql(self, expression: exp.Put) -> str:
+            props = expression.args.get("properties")
+            props_sql = self.properties(props, prefix=" ", sep=" ", wrapped=False) if props else ""
+            return f"PUT '{expression.this}' {expression.args['target']}{props_sql}"
```
georgesittas, line 1292: We need to do `this = self.sql(expression, "this"`; the quotes shouldn't be hard-coded. The same applies to `expression.args["target"]`, because the current approach will silently generate it without a dialect.

sqlglot/parser.py
```diff
@@ -527,6 +527,7 @@ class Parser(metaclass=_Parser):
         TokenType.PERCENT,
         TokenType.PIVOT,
         TokenType.PRAGMA,
+        TokenType.PUT,
```
georgesittas, line 530: Let's add this at the end of `identity.sql`:

```sql
SELECT 1 put
```

This tests that the `PUT` token can still be used as, e.g., an identifier.

sqlglot/parser.py
```diff
@@ -827,6 +828,7 @@ class Parser(metaclass=_Parser):
         TokenType.MERGE: lambda self: self._parse_merge(),
         TokenType.PIVOT: lambda self: self._parse_simplified_pivot(),
         TokenType.PRAGMA: lambda self: self.expression(exp.Pragma, this=self._parse_expression()),
+        TokenType.PUT: lambda self: self._parse_put(),
```
georgesittas, line 831: We don't have to put this here, right?

sqlglot/dialects/snowflake.py
```diff
@@ -1277,3 +1285,8 @@ def select_sql(self, expression: exp.Select) -> str:
             if offset and not limit:
                 expression.limit(exp.Null(), copy=False)
             return super().select_sql(expression)
+
+        def put_sql(self, expression: exp.Put) -> str:
```
georgesittas, line 1289: Move this to the base generator, all expressions need to be generatable when using `sql()` without a dialect.

georgesittas (commented):
A few comments, but looks good.

sqlglot/parser.py
whummer: Was trying to remove this, but then was getting an error like this:

```
E           sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 24.
E             PUT 'file://my file.txt' '@s1/my folder'
```

this is the full trace:
```
tests/dialects/test_snowflake.py:2374:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
sqlglot/__init__.py:139: in parse_one
sqlglot/dialects/dialect.py:991: in parse
sqlglot/parser.py:1499: in parse
    return self._parse(
sqlglot/parser.py:1571: in _parse
```

sqlglot/dialects/snowflake.py
whummer: Thanks, removed 👍

sqlglot/dialects/snowflake.py
```diff
@@ -805,6 +805,14 @@ def _parse_show_snowflake(self, this: str) -> exp.Show:
                 },
             )
 
+        def _parse_put(self) -> exp.Put | exp.Command:
+            if self._curr.token_type != TokenType.STRING:
+                return self._parse_as_command(self._prev)
```
whummer, line 810: @georgesittas  Might need some help with this one. Was following the pattern in other places of the parser, but when running the following test:
```
self.validate_identity("PUT file:///dir/tmp.csv @%table")`
```

... I'm getting this assertion error:
```
============================== 1 failed in 0.44s ===============================
FAILED               [100%]
tests/dialects/test_snowflake.py:2371 (TestSnowflake.test_put_to_stage)
'PUT file:' != 'PUT file:///dir/tmp.csv @%table'

Expected :'PUT file:///dir/tmp.csv @%table'
Actual   :'PUT file:'
```

Is it possible that this part of the code is not able to handle the colon in the expression `"PUT file:///dir/tmp.csv @%table"`..? 🤔 
```
    def _parse_as_command(self, start: Token) -> exp.Command:
        while self._curr:
            self._advance()
        text = self._find_sql(start, self._prev)  # <- when debugging, `text` is `'PUT file:` here
        size = len(start.text)
        self._warn_unsupported()
        return exp.Command(this=text[:size], expression=text[size:])
```

sqlglot/parser.py
whummer: Done, added.

sqlglot/dialects/snowflake.py
whummer: Done

sqlglot/parser.py
georgesittas: We still need this in the Snowflake dialect, my comment referred to the base class entry.

sqlglot/dialects/snowflake.py
georgesittas: I'd check the output of `sqlglot.tokenize(..., "snowflake")`, this may be a tokenization issue.

tests/dialects/test_snowflake.py
```diff
@@ -1029,7 +1029,7 @@ def test_staged_files(self):
         self.validate_identity("SELECT * FROM @namespace.mystage/path/to/file.json.gz")
         self.validate_identity("SELECT * FROM @namespace.%table_name/path/to/file.json.gz")
         self.validate_identity("SELECT * FROM '@external/location' (FILE_FORMAT => 'path.to.csv')")
-        self.validate_identity("PUT file:///dir/tmp.csv @%table", check_command_warning=True)
+        self.validate_identity("PUT file:///dir/tmp.csv @%table")
```
georgesittas, line 1032: We will need the flag to be reset to `True` here.

sqlglot/parser.py
whummer: Makes sense, thanks for clarifying. 👍 Updated

sqlglot/dialects/snowflake.py
whummer: Indeed, seems to be a tokenizer issue @georgesittas  - I've added a simple test here, which is currently failing: https://github.com/tobymao/sqlglot/pull/4818/files#diff-51d1cbb326cf0b05cde57945719d3b9c1f69f7d3f897ff247d19e0a0de7a3a38R181-R201

(Will be offline for the next ~2h or so, but will keep digging a bit later today.

Thanks for your help, and for being patient with me! 😁

whummer: Ok, another updated here @georgesittas - I think the issue is with the definition of `Snowflake.Tokenizer.COMMENTS = ["--", "//", ("/*", "*/")]` which contains `//` as a token, which then conflicts with the `://` in URLs.

Cannot propose a solution for this right now, as I need to get a better understanding of the tokenizer first. Any advice from your side on how we could resolve this? 🙌

georgesittas: Oh, interesting! I didn't remember they use `//` as a comment delimiter; that makes sense then. Snowflake must be tokenizing `://` differently or something, so that doesn't clash with the comment sequence.

Since we're not looking to support this syntax right now, I think let's just get rid of the test to make progress.

georgesittas: Or, actually, let me take a quick stab at this. The test will be good to have, otherwise this could cause a regression.

tests/fixtures/identity.sql
```diff
@@ -888,4 +888,5 @@ SELECT 1 LIMIT 1
 CAST(x AS INT128)
 CAST(x AS UINT128)
 CAST(x AS UINT256)
-SELECT export
\ No newline at end of file
+SELECT export
+SELECT 1 put
```
georgesittas, line 892: Btw, my bad for suggesting this, it won't be an "identity" because it'd map `SELECT 1 put` to `SELECT 1 AS put`.

georgesittas:
I'll merge this and take it to the finish line, appreciate the contribution @whummer.

sqlglot/dialects/snowflake.py
georgesittas: Yep, I think that makes sense. I'll follow up in a new PR 👍

PR #4828, opened 2025-03-04
feat(oracle): add FORCE property

### LINKED ISSUE #4826: FORCE keyword can't be parsed
Hi, when I add the FORCE keyword to this SQL it can't be parsed.

**Working code**
```python
import sqlglot
o1 = """create view vw_test as select name from users;"""
exp1 = sqlglot.parse(o1, read="oracle")
print(exp1)
```

Broken code
```python
import sqlglot
o2 = """create force view vw_test as select name from users;"""
exp2 = sqlglot.parse(o2, read="oracle")
print(exp2)
```
Output
```
'create force view vw_test as select name from users' contains unsupported syntax. Falling back to parsing
as a 'Command'.
[Command(this=create, expression=force view vw_test as select name from users)]
```

[oracle doc](https://docs.oracle.com/cd/B13789_01/server.101/b10759/statements_8004.htm)

geooo109:
Fixes #4826

Add support for the `FORCE` property, when creating a view in Oracle. 

**DOCS**
[Oracle FORCE](https://docs.oracle.com/cd/B13789_01/server.101/b10759/statements_8004.htm)

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -153,6 +153,7 @@ class Oracle(Dialect):
             and self.expression(exp.TemporaryProperty, this="GLOBAL"),
             "PRIVATE": lambda self: self._match_text_seq("TEMPORARY")
             and self.expression(exp.TemporaryProperty, this="PRIVATE"),
+            "FORCE": lambda self: self.expression(exp.ForceProperty, this="FORCE"),
         }
 
         QUERY_MODIFIER_PARSERS = {
@@ -319,11 +320,13 @@ class Oracle(Dialect):
             exp.Unicode: lambda self, e: f"ASCII(UNISTR({self.sql(e.this)}))",
             exp.UnixToTime: lambda self,
             e: f"TO_DATE('1970-01-01', 'YYYY-MM-DD') + ({self.sql(e, 'this')} / 86400)",
+            exp.ForceProperty: lambda self, e: self.sql(e, "this"),
         }
 
         PROPERTIES_LOCATION = {
             **generator.Generator.PROPERTIES_LOCATION,
             exp.VolatileProperty: exp.Properties.Location.UNSUPPORTED,
+            exp.ForceProperty: exp.Properties.Location.POST_CREATE,
         }
 
         def currenttimestamp_sql(self, expression: exp.CurrentTimestamp) -> str:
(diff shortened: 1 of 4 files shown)


tests/dialects/test_oracle.py
```diff
@@ -319,6 +319,8 @@ def test_oracle(self):
                 "tsql": "SELECT * FROM t ORDER BY (SELECT NULL) OFFSET 0 ROWS FETCH FIRST 10 ROWS ONLY",
             },
         )
+        self.validate_identity("SELECT 1 AS FORCE")
```
georgesittas, line 322: This isn't the correct way to test this; you need to do `SELECT 1 FORCE` because it follows a different execution path. I expect this to fail if you change it, meaning that you also need to add the token to `ID_VAR_TOKENS` (do that in the base class' set).

sqlglot/expressions.py
```diff
@@ -3093,6 +3093,10 @@ class IncludeProperty(Property):
     arg_types = {"this": True, "alias": False, "column_def": False}
 
 
+class ForceProperty(Property):
+    arg_types = {"this": True}
```
georgesittas, line 3097: Why do you need an argument for this? You could leave this empty (`{}`) and instead just generate the expression trivially as `"FORCE"`.

sqlglot/dialects/oracle.py
```diff
@@ -319,11 +320,13 @@ class Generator(generator.Generator):
             exp.Unicode: lambda self, e: f"ASCII(UNISTR({self.sql(e.this)}))",
             exp.UnixToTime: lambda self,
             e: f"TO_DATE('1970-01-01', 'YYYY-MM-DD') + ({self.sql(e, 'this')} / 86400)",
+            exp.ForceProperty: lambda self, e: self.sql(e, "this"),
```
georgesittas, line 323: Move both this and the property location in the base generator class. Every expression needs to be generatable without specifying a dialect.

tests/dialects/test_oracle.py
geooo109: Agree, I tested it and you can use it, i forgot to add it, my bad, I will fix it.

sqlglot/dialects/oracle.py
geooo109: Yeap, I see, my bad, fixed it.

tests/dialects/test_oracle.py
VaggelisD: If you haven't added an entry in `KEYWORDS` then this test is not needed

geooo109: Mostly added it for sanity check, I will remove it.

PR #4830, opened 2025-03-04
Fix(snowflake): clean up PUT implementation

georgesittas:
Follow-up on https://github.com/tobymao/sqlglot/pull/4818, cc @whummer.


sqlglot/dialects/snowflake.py
```diff
@@ -870,6 +873,7 @@ class Tokenizer(tokens.Tokenizer):
 
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
+            "FILE://": TokenType.URI_START,
```
georgesittas, line 876: I ended up adding this to avoid hacks. Now the `//` substring here won't be treated as a comment due to [maximal munching](https://en.wikipedia.org/wiki/Maximal_munch).

PR #4832, opened 2025-03-05
feat(hive) add support for STORED BY syntax for storage handlers

tsamaras:
`STORED BY` is used in hive to access data hosted in other systems like HBase, Iceberg, etc.


sqlglot/parser.py
```diff
@@ -2120,22 +2120,28 @@ def _parse_property(self) -> t.Optional[exp.Expression]:
 
         return self.expression(exp.Property, this=key, value=value)
 
-    def _parse_stored(self) -> exp.FileFormatProperty:
-        self._match(TokenType.ALIAS)
```
georgesittas, line 2124: I would consider refactoring this so this goes first:

```python
        if self._match_texts("BY"):
            return self.expression(exp.StorageHandlerProperty, this=self._parse_var_or_string())
```

You can then avoid the big diff, thus lowering the reviewing mental overhead.

georgesittas (commented):
Thank you for the contribution!

I've left one comment, let's clean this up a bit and it should be good to go.

VaggelisD:
Hey @tsamaras, thank you for the PR, I'll take it to the finish line.

PR #4844, opened 2025-03-06
fix(mysql)!: BLOB transpilation

### LINKED ISSUE #4839: MySQL: transform blob to varbinary makes a syntax error
#4835  I cannot reopen this issue, so I open a new one.
```python
>>> sql = """
... CREATE TABLE test_blob_types (
...     id INT PRIMARY KEY,
...     standard_blob BLOB,
...     tiny TINYBLOB,
...     medium MEDIUMBLOB,
...     large LONGBLOB
... );
... """
>>> ast = sqlglot.parse_one(sql, dialect="mysql")
>>> ast
Create(
  this=Schema(
    this=Table(
      this=Identifier(this=test_blob_types, quoted=False)),
    expressions=[
      ColumnDef(
        this=Identifier(this=id, quoted=False),
        kind=DataType(this=Type.INT, nested=False),
        constraints=[
          ColumnConstraint(
            kind=PrimaryKeyColumnConstraint())]),
      ColumnDef(
        this=Identifier(this=standard_blob, quoted=False),
        kind=DataType(this=Type.VARBINARY, nested=False)),
      ColumnDef(
        this=Identifier(this=tiny, quoted=False),
        kind=DataType(this=Type.TINYBLOB, nested=False)),
      ColumnDef(
        this=Identifier(this=medium, quoted=False),
        kind=DataType(this=Type.MEDIUMBLOB, nested=False)),
      ColumnDef(
        this=Identifier(this=large, quoted=False),
        kind=DataType(this=Type.LONGBLOB, nested=False))]),
  kind=TABLE)
>>> ast.sql("mysql")
'CREATE TABLE test_blob_types (id INT PRIMARY KEY, standard_blob VARBINARY, tiny TINYBLOB, medium MEDIUMBLOB, large LONGBLOB)'
```
and I execute the result in mysql client
```
CREATE TABLE test_blob_types (id INT PRIMARY KEY, standard_blob VARBINARY, tiny TINYBLOB, medium MEDIUMBLOB, large LONGBLOB)
> 1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ', tiny TINYBLOB, medium MEDIUMBLOB, large LONGBLOB)' at line 1
```
> blob should be transformed as varbinary(65535)

geooo109:
Fixes  #4839

This PR adds support for the correct transpilation of `BLOB` from MySQL to other dialects.

The default mapping is as the previous `BLOB` -> `VARBINARY`.  
For some dialects there is special treatment and this mapping changes. 

**DOCS**
| Dialect  | Link |
| -------------  | ------------- |
| MySQL | [MySQL BLOB Documentation](https://dev.mysql.com/doc/refman/8.4/en/blob.html)  |
| Oracle | [Oracle BLOB Documentation](https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/blob.html)  |
| Postgres | [Postgres Binary Data Types](https://www.postgresql.org/docs/current/datatype-binary.html) |
| T-SQL | [T-SQL Binary and Varbinary](https://learn.microsoft.com/en-us/sql/t-sql/data-types/binary-and-varbinary-transact-sql?view=sql-server-ver16#varbinary---n--max--) |
| SQLite | [SQLite Data Types](https://www.sqlite.org/datatype3.html)  | 
| DuckDB | [DuckDB Data Types Overview](https://duckdb.org/docs/stable/sql/data_types/overview.html)  |
| Hive | [Hive Misc Types](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27838462#LanguageManualTypes-MiscTypes) |
| BigQuery | [BigQuery Bytes Type](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#bytes_type) |
| Redshift | [Redshift VARBYTE Type](https://docs.aws.amazon.com/redshift/latest/dg/r_VARBYTE_type.html) |
| Clickhouse | [Clickhouse String Data Types](https://clickhouse.com/docs/sql-reference/data-types/string) |


sqlglot/expressions.py
```diff
@@ -4526,6 +4526,7 @@ class Type(AutoName):
         XML = auto()
         YEAR = auto()
         TDIGEST = auto()
+        BLOB = auto()
```
georgesittas, line 4529: Ditto

georgesittas (approved):
Looks good, just a few nits re: styling.

tests/dialects/test_mysql.py
```diff
@@ -346,6 +346,22 @@ def test_canonical_functions(self):
                 "tsql": "CHAR(10)",
             },
         )
+        self.validate_identity("CREATE TABLE t (foo VARBINARY)")
```
georgesittas, line 349: This isn't a valid MySQL statement, let's do `CREATE TABLE t (foo VARBINARY(5))` instead.

geooo109: You have right thanks.

PR #4846, opened 2025-03-06
Refactor!: transpile division by zero semantics

georgesittas:
Behavior today:

```python
>>> import sqlglot
>>> sqlglot.transpile("1 / 0", "duckdb", "postgres")
['CAST(1 AS DOUBLE PRECISION) / NULLIF(0, 0)']
```

However, DuckDB (and ClickHouse) produces "infinity" when dividing by zero:

```
D SELECT 1 / 0;
┌─────────┐
│ (1 / 0) │
│ double  │
├─────────┤
│   inf   │
└─────────┘
```

And the transpiled SQL produces `NULL` for Postgres, so the source semantics are not really respected:

```
georgesittas=# SELECT CAST(1 AS DOUBLE PRECISION) / NULLIF(0, 0);
 ?column?
----------

(1 row)
```

This PR refactors our "safe division" logic in order to take infinity into account. I'm considering a couple of improvements that I will mention inline, by commenting. I will probably add some more tests for this, as well.


sqlglot/dialects/dialect.py
```diff
@@ -342,8 +342,14 @@ class Dialect(metaclass=_Dialect):
     True means `a / b` is integer division if both `a` and `b` are integers.
     """
 
-    SAFE_DIVISION = False
-    """Whether division by zero throws an error (`False`) or returns NULL (`True`)."""
+    SAFE_DIVISION: bool | str = False
+    """Whether division by zero throws an error (`False`), returns NULL (`True`), or "infinity" (`"inf"`)."""
```
georgesittas, line 346: I'm considering creating an enumeration for this, similar to `NormalizationStrategy`. Any ideas for naming? I thought of `DivByZeroStrategy`.

sqlglot/dialects/mysql.py
```diff
@@ -163,6 +163,7 @@ class MySQL(Dialect):
     SUPPORTS_USER_DEFINED_TYPES = False
     SUPPORTS_SEMI_ANTI_JOIN = False
     SAFE_DIVISION = True
+    INFINITY = None
```
georgesittas, line 166: I will try moving this to the `_Dialect.__new__` method in a followup commit to reduce duplication.

tests/dialects/test_postgres.py
```diff
@@ -775,7 +775,7 @@ def test_postgres(self):
             },
             write={
                 "sqlite": "1 / CAST(CAST(CAST(4 AS REAL) / 2 AS INTEGER) AS REAL)",
-                "duckdb": "1 / CAST(4 // 2 AS DECIMAL)",
+                "duckdb": "1 / CASE WHEN CAST(4 // 2 AS DECIMAL) = 0 THEN ERROR('Division by zero error') ELSE CAST(4 // 2 AS DECIMAL) END",
```
georgesittas, line 778: Doing an integer division by zero isn't handled yet. We may need to deal with this separately, because I noticed the semantics aren't necessarily the same as normal division...

ClickHouse:

```
:) select 1 / 0;
   ┌─divide(1, 0)─┐
1. │          inf │
   └──────────────┘

:) SELECT 1 // 0;
   ┌─1─┐
1. │ 1 │
   └───┘
```

DuckDB:

```
D SELECT 1 / 0;
┌─────────┐
│ (1 / 0) │
│ double  │
├─────────┤
│   inf   │
└─────────┘
D SELECT 1 // 0;
┌──────────┐
│ (1 // 0) │
│  int32   │
├──────────┤
│   NULL   │
└──────────┘
```

OTOH, Postgres is consistent:

```
georgesittas=# SELECT 1 / 0;
ERROR:  division by zero
georgesittas=# SELECT DIV(1, 0);
ERROR:  division by zero
```

sqlglot/dialects/dialect.py
georgesittas: Chatted with Toby and concluded that we'd rather avoid the enum approach for now.

georgesittas:
Closing this PR for now because it feels quite intrusive... E.g., there are places where we do `a / b` and rely on the magic builder syntax, so tracking those down & fixing them will take more time. I'm also worried this could break a lot of third-party code that builds `Div`s.

I'll wait until an issue (re: infinity vs `NULL`) comes up in practice– this should be a good starting point if it happens.

PR #4847, opened 2025-03-06
Fix(parser): properly parse qualified columns when parsing "columns ops"

### LINKED ISSUE #4845: Error in json_extract_path with "indirect field name"
**Before you file an issue**
dialect is postgresql for reading and writing the query

The sql `JSON_ARRAY_ELEMENTS(issues.fields -> custom_fields.field_id )` transpiles as 
`JSON_ARRAY_ELEMENTS(JSON_EXTRACT_PATH(issues.fields, custom_fields).field_id)` which leads to an error because of the incorrect parenthesis.

I would expect this instead : `JSON_ARRAY_ELEMENTS(JSON_EXTRACT_PATH(issues.fields, custom_fields.field_id))`

The full query is
```
WITH issues AS (
    SELECT
        id AS ticket_id,
        fields,
        sync_at
    FROM "source"."issues"
),

custom_fields (field_id) AS (
    VALUES
    ('customfield_10129'),
    ('customfield_10131'),
    ('customfield_10132')
)

SELECT
    issues.ticket_id,
    field.value ->> 'value' AS "value",
    issues.sync_at
FROM issues
    CROSS JOIN custom_fields
    CROSS JOIN LATERAL JSON_ARRAY_ELEMENTS(issues.fields -> custom_fields.field_id ) AS field
WHERE issues.fields -> custom_fields.field_id IS NOT NULL
```

And `sqlglot.transpile(sql, read="postgres", write="postgres")` produces
```
['WITH issues AS (SELECT id AS ticket_id, fields, sync_at FROM "source"."issues"), custom_fields(field_id) AS (SELECT * FROM (VALUES (\'customfield_10129\'), (\'customfield_10131\'), (\'customfield_10132\')) AS _values) SELECT issues.ticket_id, field.value ->> \'value\' AS "value", issues.sync_at FROM issues CROSS JOIN custom_fields CROSS JOIN LATERAL JSON_ARRAY_ELEMENTS(JSON_EXTRACT_PATH(issues.fields, custom_fields).field_id) AS field WHERE NOT JSON_EXTRACT_PATH(issues.fields, custom_fields).field_id IS NULL']
```

georgesittas:
Fixes #4845

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -5345,6 +5345,8 @@ class Parser(metaclass=_Parser):
                     self.raise_error("Expected type")
             elif op and self._curr:
                 field = self._parse_column_reference() or self._parse_bracket()
+                if isinstance(field, exp.Column) and self._match(TokenType.DOT, advance=False):
+                    field = self._parse_column_ops(field)
             else:
                 field = self._parse_field(any_token=True, anonymous_func=True)
 
--- a/tests/dialects/test_postgres.py
+++ b/tests/dialects/test_postgres.py
@@ -345,6 +345,28 @@ class TestPostgres(Validator):
             "CAST(x AS INT8)",
             "CAST(x AS BIGINT)",
         )
+        self.validate_identity(
+            """
+            WITH
+              json_data AS (SELECT '{"field_id": [1, 2, 3]}'::JSON AS data),
+              field_ids AS (SELECT 'field_id' AS field_id)
+
+            SELECT
+                JSON_ARRAY_ELEMENTS(json_data.data -> field_ids.field_id) AS element
+            FROM json_data, field_ids
+            """,
+            """WITH json_data AS (
+  SELECT
+    CAST('{"field_id": [1, 2, 3]}' AS JSON) AS data
+), field_ids AS (
+  SELECT
+    'field_id' AS field_id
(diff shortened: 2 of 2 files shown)


sqlglot/parser.py
```diff
@@ -5345,6 +5345,8 @@ def _parse_column_ops(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.
                     self.raise_error("Expected type")
             elif op and self._curr:
                 field = self._parse_column_reference() or self._parse_bracket()
+                if isinstance(field, exp.Column) and self._match(TokenType.DOT, advance=False):
+                    field = self._parse_column_ops(field)
```
VaggelisD, line 5349: The issue in the example query is that we were unable to parse `json_data.data -> field_ids.field_id` because of the qualified names?

VaggelisD: Ah I see, so we'd basically match `[t.json_value -> foo].bar`. Would it work if used `_parse_column` instead of `_parse_column_reference`? I see the former also calls `parse_column_ops` in it

VaggelisD: Something like:

```Python3
            elif op and self._curr:
                field = self._parse_column()
```

Issue #4487, opened 2024-12-09
Comma at the end of FROM token list is legal in SNOWFLAKE, but causes parse error in sqlglot


kosta-foundational:
sqlglot version 25.32.0

**Fully reproducible code snippet**
Example 1:
```
    q = """
SELECT
    id
FROM import_chat_sessions,
LATERAL FLATTEN(input => history) prs_history,
"""
sqlglot.parse(sql=q, dialect=sqlglot.Dialects.SNOWFLAKE)
```

Example 2:
```
    q = """
SELECT
    id
FROM import_chat_sessions,
WHERE id > 2
"""
sqlglot.parse(sql=q, dialect=sqlglot.Dialects.SNOWFLAKE)
```

georgesittas:
Thanks for the report. I took a quick stab at it and the fix didn't seem trivial, so I'd say this one's out of scope for the time being.

kosta-foundational:
@georgesittas , Note that parsing succeeds if there is extra comma in SELECT list.

georgesittas:
Yeah, I'm aware. Trailing commas are generally non-standard, but SQLGlot will go ahead and parse them anyway in many cases. The `FROM` clause is an exception in this case because the parsing logic that deals with joins is more complicated.

georgesittas:
Fixed in https://github.com/tobymao/sqlglot/commit/b7a0df1b9a9cff2cd57db77ac0095c189b9d67ab thanks to @omerhadari!

PR #4854, opened 2025-03-07
feat(parser): Support trailing commas after from

omerhadari:
Solves #4487 

1. I initially wanted to use `_parse_csv` but it's not really applicable due to the way joins have to be parsed
2. I needed to remove a case from the `partial` test (trailing comma). Felt ok because `sqlglot` is not an sql validator :)
3. Hope it's ok I implemented this in the general parser even though the only dialect I know of that supports this is Snowflake. Again it felt ok because of the non-validator guideline.


georgesittas (approved):
Nice and simple, thanks for looking into this!

georgesittas:
> I initially wanted to use _parse_csv but it's not really applicable due to the way joins have to be parsed

I think this is what I had initially tried and it didn't seem to work so I gave up on it :)

omerhadari:
Hey @georgesittas thanks for the super quick merge!

Issue #4852, opened 2025-03-07
Support Snowflake's HEX_DECODE_BINARY


sk-:
Snowflake has some non standard function like [HEX_DECODE_BINARY](https://docs.snowflake.com/en/sql-reference/functions/hex_decode_binary) and [HEX_DECODE_STRING](https://docs.snowflake.com/en/sql-reference/functions/hex_decode_string), which are not supported by other databases. Sqlglot does not convert those functions when parsing from snowflake.

**Fully reproducible code snippet**
```py
import sqlglot
>>> sqlglot.parse_one("SELECT HEX_DECODE_BINARY('62')", read="snowflake").sql("duckdb")
"SELECT HEX_DECODE_BINARY('65')"
>>> sqlglot.parse_one("SELECT HEX_DECODE_BINARY('62')", read="snowflake").sql("bigquery")
"SELECT HEX_DECODE_BINARY('65')"
```

**Official Documentation**
Duckdb provides the functions FROM_HEX (undocumented, but used already in this project) that converts the hex string to binary and [decode](https://duckdb.org/docs/stable/sql/functions/blob.html#decodeblob), which converts utf8 bytes to string.

This would allow to implement `HEX_DECODE_BINARY(x)` as `FROM_HEX(x)` and `HEX_DECODE_STRING` as `DECODE(FROM_HEX(x))`

Here is a [playground](https://codapi.org/embed/?sandbox=duckdb&code=data%3A%3Bbase64%2CK07NSU0uUUgrys%2BNz0it0FA3M1XX1FEoqSxIzU%2FTQBXW1FFISU3OT0nFFIeqxy6tCQA%3D) showing the use of these functions in duckdb.

Bigquery also provide the function [FROM_HEX](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#from_hex) and the function [cast](https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_functions#cast).
This would allow to implement `HEX_DECODE_BINARY(x)` as `FROM_HEX(x)` and `HEX_DECODE_STRING` as `CAST(FROM_HEX(x) AS STRING)`

Note: I'm much more interested in having support for `HEX_DECODE_BINARY`, but added `HEX_DECODE_STRING` for completeness.

georgesittas:
Not sure when we'll have bandwidth to look into this, but PRs are welcome if you want to tackle this in the meantime.

georgesittas:
[...]

PR #4855, opened 2025-03-07
fix(snowflake): add support for HEX_DECODE_BINARY

sk-:
The [HEX_DECODE_BINARY](https://docs.snowflake.com/en/sql-reference/functions/hex_decode_binary) function is equivalent to the `Unhex` function that is already supported by sqlglot. To support it, we just had to add the map in snowflake and add the transform in duckdb.

Partially solves #4852, we still need to add support for `HEX_DECODE_STRING`.

Testing:
Duckdb ([playground](https://codapi.org/embed/?sandbox=duckdb&code=data%3A%3Bbase64%2CK07NSU0uUeBKK8rPjc9IrdBQNzNV19RRKKksSM1P00AV1tThSklNzk9JxRCHqccuranDVZqHaTiSGMJkZEEA))

Screenshot 2025-03-07 at 6 40 35 PM

Bigquery
(Note that Bigquery outputs binary data (bytes) in base64)
Screenshot 2025-03-07 at 6 23 06 PM

Snowflake
(Note that Snowflake outputs binary data in hex format)
Screenshot 2025-03-07 at 6 34 53 PM

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -666,6 +666,7 @@ class DuckDB(Dialect):
                 exp.cast(e.expression, exp.DataType.Type.TIMESTAMP),
                 exp.cast(e.this, exp.DataType.Type.TIMESTAMP),
             ),
+            exp.Unhex: rename_func("FROM_HEX"),
             exp.UnixToStr: lambda self, e: self.func(
                 "STRFTIME", self.func("TO_TIMESTAMP", e.this), self.format_time(e)
             ),
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -666,7 +666,6 @@ class DuckDB(Dialect):
                 exp.cast(e.expression, exp.DataType.Type.TIMESTAMP),
                 exp.cast(e.this, exp.DataType.Type.TIMESTAMP),
             ),
-            exp.Unhex: rename_func("FROM_HEX"),
             exp.UnixToStr: lambda self, e: self.func(
                 "STRFTIME", self.func("TO_TIMESTAMP", e.this), self.format_time(e)
             ),
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -431,6 +431,7 @@ class Snowflake(Dialect):
                 this=seq_get(args, 0), expression=seq_get(args, 1), max_dist=seq_get(args, 2)
             ),
             "FLATTEN": exp.Explode.from_arg_list,
+            "FROM_HEX_BINARY": exp.Unhex.from_arg_list,
             "GET_PATH": lambda args, dialect: exp.JSONExtract(
                 this=seq_get(args, 0), expression=dialect.to_json_path(seq_get(args, 1))
             ),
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -666,6 +666,7 @@ class Generator(generator.Generator):
                 exp.cast(e.expression, exp.DataType.Type.TIMESTAMP),
                 exp.cast(e.this, exp.DataType.Type.TIMESTAMP),
             ),
+            exp.Unhex: rename_func("FROM_HEX"),
```
georgesittas, line 669: This doesn't seem necessary, `FROM_HEX` and `UNHEX` are [aliases](https://github.com/duckdb/duckdb/blob/981c7a8573b254f8efd6b71f764c5c3bd924a53d/extension/core_functions/scalar/string/functions.json#L276-L281).
```suggestion
```

sqlglot/dialects/snowflake.py
```diff
@@ -431,6 +431,7 @@ class Parser(parser.Parser):
                 this=seq_get(args, 0), expression=seq_get(args, 1), max_dist=seq_get(args, 2)
             ),
             "FLATTEN": exp.Explode.from_arg_list,
+            "FROM_HEX_BINARY": exp.Unhex.from_arg_list,
```
georgesittas, line 434: This isn't a valid Snowflake function.

tests/dialects/test_snowflake.py
```diff
@@ -998,6 +998,15 @@ def test_snowflake(self):
...
         self.validate_identity("CREATE TABLE t (id INT PRIMARY KEY AUTOINCREMENT)")
 
+        self.validate_all(
+            "SELECT FROM_HEX_BINARY('65')",
+            write={
+                "bigquery": "SELECT FROM_HEX('65')",
+                "duckdb": "SELECT FROM_HEX('65')",
+                "snowflake": "SELECT FROM_HEX_BINARY('65')",
+            },
+        )
```
georgesittas, line 1008: Did you test these against each engine?

sqlglot/dialects/duckdb.py
sk-: Without this the test I added fails. Note that bigquery and presto also define this conversion.

sqlglot/dialects/snowflake.py
sk-: Thanks for catching this, it was a copy paste error on my side.

sqlglot/dialects/duckdb.py
georgesittas: DuckDB already generates `UNHEX` for `MD5Digest`, so generating `FROM_HEX` here is going to look inconsistent.

Given they're aliases, let's remove this entry and fix the test so that it uses `UNHEX` instead.

tests/dialects/test_snowflake.py
sk-: I added the results to the PR description. Let me know if that's enough

georgesittas: Appreciate it 👍

sqlglot/dialects/duckdb.py
sk-: Cool, didn't know that. I tested it with `UNHEX` and works just like with `FROM_HEX`, so I reverted this change.

georgesittas (approved):
Thanks!

PR #4863, opened 2025-03-10
chore(lineage): fix broken links for html visualization

omerhadari:
Links for the lineage visualization frontend lib broke, see [here](https://github.com/visjs/vis-network/issues/278).

This updates according to the most recent examples.


omerhadari:
Actually looks like both links work, my bad. closing.

PR #4865, opened 2025-03-10
Chore!: Introduce pyproject.toml and switch to packaging via build

erindru:
This moves the bulk of the `setup.py` declarations to `pyproject.toml` and builds packages using `python -m build` instead of invoking setuptools directly.

Note that this just affects Python sqlglot, sqlglot-rs already uses `pyproject.toml` and has a build process based on maturin.

I checked that the wheel and sdist files produced by `python -m build` contained the same files as before, in particular the `py.typed` marker and a correct `extras` declaration for `[rs]` with the version parsed from Cargo.toml

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


tobymao:
@barakalon can we drop python 3.7 yet? it's ancient.

tobymao (approved):

PR #4869, opened 2025-03-11
Feat(duckdb): add support for prefix aliases

georgesittas:
Reference: https://duckdb.org/2025/02/25/prefix-aliases-in-sql.html

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -468,6 +468,58 @@ class DuckDB(Dialect):
             TokenType.DETACH: lambda self: self._parse_attach_detach(is_attach=False),
         }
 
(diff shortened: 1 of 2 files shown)


tobymao (approved):
does this not mess with structs?

georgesittas:
> does this not mess with structs?

I think it should be fine because `:` in the struct binds more tightly:

```
(.venv) ➜  sqlglot git:(jo/duckdb_prefix_alias) python
Python 3.12.7 (main, Oct  1 2024, 02:05:46) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> import sqlglot
>>> sqlglot.parse_one("select {'x': 1}")
Select(
  expressions=[
    Struct(
      expressions=[
        PropertyEQ(
          expression=Literal(this=1, is_string=False))])])
```

PR #4872, opened 2025-03-11
Add a query provider to the expand function, enabling lazy loading of queries during AST transformations.

DannyMor:
## PR Description

This PR introduces lazy query resolution to the exp.expand function, significantly enhancing its efficiency and flexibility. Previously, expanding a query required a complete dictionary of all possible source queries upfront, which could be inefficient for large datasets with thousands of views. This approach necessitated either prior knowledge of all referenced tables or a costly recursive traversal to gather them, leading to unnecessary parsing overhead when only a subset of queries needed expansion.

---
### With this update:

Queries can now be resolved on demand, eliminating the need for pre-parsing all queries and improving performance.

A new abstract class, QueryProvider, has been added to define a standardized interface for retrieving queries by name.

A default implementation ensures backward compatibility with existing functionality.

This enhancement optimizes query expansion workflows, particularly for large-scale datasets, by reducing unnecessary computations and improving scalability.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -17,6 +17,7 @@ import numbers
 import re
 import textwrap
 import typing as t
+from abc import ABC, abstractmethod
 from collections import deque
 from copy import deepcopy
 from decimal import Decimal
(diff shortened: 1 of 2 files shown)


tobymao:
i'd rather have a simpler approach, maybe instead, sources can just be a lambda, and you can call it with the source name, then you can implement whatever you want there

DannyMor:
@tobymao sure, np, will do that!

DannyMor:
Assigning a lambda to a variable is not allowed by the enforcement of Ruff, which raises the following error:

```
sqlglot/expressions.py:8528:9: E731 Do not assign a `lambda` expression, use a `def`
     |
8526 |             normalize_table_name(k, dialect=dialect): v for k, v in sources.items()
8527 |         }
8528 |         get_source = lambda name: normalized_sources.get(name)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ E731
```     
     
This restriction is why the def _get_source function definition is used to implement the default dictionary logic, instead of the more concise lambda expression.

PR #4876, opened 2025-03-12
fix(hive)!: support STRUCT(*) and MAP(*)

### LINKED ISSUE #4871: Spark support for map(*) and struct(*)
Theres a few ways to covert columns to json in spark depending on the use case, typically using map or struct. While its common to specify the schema, thats not required and sometimes not desired. 
```python
import sqlglot

sql = """
select
    to_json(map(*)),
    to_json(struct(*))
from data
"""
print(sqlglot.parse_one(sql, dialect='spark').sql(dialect='spark', pretty=True))
```
outputs:
```
SELECT
  TO_JSON(STAR_MAP(*)),
  TO_JSON(STRUCT(* AS col1))
FROM data
```
I'd expect this to not modify the query. The `to_json` part likely isn't relevant.

1.  `star_map` isn't a function in spark
2.  `struct(* as col1)` throws: `AnalysisException: Invalid use of '*' in expression 'alias'`

geooo109:
Fixes #4871

This PR adds support for `STRUCT(*)` and `MAP(*)` for hive, spark, spark2, databricks dialects.

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -446,7 +446,7 @@ class Hive(Dialect):
             return self.expression(exp.Parameter, this=this, expression=expression)
 
         def _to_prop_eq(self, expression: exp.Expression, index: int) -> exp.Expression:
-            if isinstance(expression, exp.Star):
+            if expression.is_star:
                 return expression
 
             if isinstance(expression, exp.Column):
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/hive.py
```diff
@@ -446,6 +446,9 @@ def _parse_parameter(self) -> exp.Parameter:
             return self.expression(exp.Parameter, this=this, expression=expression)
 
         def _to_prop_eq(self, expression: exp.Expression, index: int) -> exp.Expression:
+            if isinstance(expression, exp.Star):
```
georgesittas, line 449:
```suggestion
            if expression.is_star:
```

georgesittas (approved):

PR #4881, opened 2025-03-14
fix!: do not expand having expressions if they conflict with a projec…

tobymao:
…tion


sqlglot/optimizer/qualify_columns.py
```diff
@@ -55,14 +55,13 @@ def qualify_columns(
     infer_schema = schema.empty if infer_schema is None else infer_schema
     dialect = Dialect.get_or_raise(schema.dialect)
     pseudocolumns = dialect.PSEUDOCOLUMNS
-
-    snowflake_or_oracle = dialect in ("oracle", "snowflake")
```
georgesittas, line 59: Was this removed because `CONNECT BY` is only supported Snowflake & Oracle?

sqlglot/optimizer/qualify_columns.py
```diff
@@ -296,23 +297,29 @@ def replace_columns(
...
 
-            if table and (not alias_expr or double_agg):
+                # BigQuery's having clause gets confused if an alias matches a source.
+                # SELECT x.a, max(x.b) as x FROM x GROUP BY 1 HAVING x > 1;
+                # If HAVING x is expanded to max(x.b), bigquery treats x as the new projection x instead of the table
+                if is_having and dialect == "bigquery":
+                    skip_replace = skip_replace or any(
+                        node.parts[0].name in projections
+                        for node in alias_expr.find_all(exp.Column)
+                    )
```
georgesittas, line 318: - Should this be `find_all_in_scope`?
- Could this ast traversal be factored out maybe? Or cached so we don't recompute this on each iteration. E.g. the outer loop already checks columns:

```python
        for column in walk_in_scope(node, prune=lambda node: node.is_star):
            if not isinstance(column, exp.Column):
                continue
```

sqlglot/optimizer/qualify_columns.py
tobymao: yes

georgesittas (approved):

PR #1688, opened 2023-05-24
Chore!: cleanup unnecessary tokens

tobymao:

PR #4893, opened 2025-03-19
fix!(tsql): support select...into #temp_table syntax

hhubbell:
Fixes support for creating temporary tables as select statements in T-SQL. The syntax is the same as selecting into a physical table, except that T-SQL uses a `#` symbol to denote a temp table, and a `##` to denote a global temp table.

This commit will slightly change the AST representation of the query, such that the `Table` expression is marked as `temporary` rather than its `Identifier` expression. This seems to be consistent with other dialects that support `select...into` syntax, e.g. postgres.

Here's a basic example:
```python
>>> sqlglot.parse_one("SELECT 1 INTO #foo", dialect="tsql")

# Before change
Select(
  expressions=[
    Literal(this=1, is_string=False)],
  into=Into(
    this=Table(
      this=Identifier(this=FOO, quoted=False, temporary=True))))

# After change
Select(
  expressions=[
    Literal(this=1, is_string=False)],
  into=Into(
    this=Table(
      this=Identifier(this=foo, quoted=False)),
    temporary=True))
```

This is consistent with Postgres:
```python
>>> sqlglot.parse_one("SELECT 1 INTO TEMPORARY foo", dialect="postgres")

Select(
  expressions=[
    Literal(this=1, is_string=False)],
  into=Into(
    this=Table(
      this=Identifier(this=foo, quoted=False)),
    temporary=True))
```

All tests pass, and new tests were added to ensure functionality remains consistent.  One test was updated to conform with expected T-SQL syntax.

I have a few additional thoughts/notes that I will outline in some targeted reviews in the diff.

Related PRs:
#1688
#1958
#2237

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -1176,7 +1176,8 @@ class TSQL(Dialect):
                 properties = expression.args.get("properties")
                 is_temp = any(
                     isinstance(prop, exp.TemporaryProperty)
-                    for prop in (properties.expressions if properties else []))
+                    for prop in (properties.expressions if properties else [])
+                )
 
                 select_into = exp.select("*").from_(exp.alias_(ctas_expression, "temp", table=True))
                 select_into.set("into", exp.Into(this=table, temporary=is_temp))
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -779,6 +779,20 @@ class TSQL(Dialect):
 
             return self.expression(exp.UserDefinedFunction, this=this, expressions=expressions)
 
+        def _parse_into(self) -> t.Optional[exp.Into]:
+            if not self._match(TokenType.INTO):
+                return None
+
+            # NOTE: This is out of order compared to `super`. T-SQL uses a `#`
+            # symbol to denote a TEMP TABLE, rather than the keyword.
+            unlogged = self._match_text_seq("UNLOGGED")
+            temp = self._match(TokenType.HASH)
+            self._match(TokenType.TABLE)
+
+            return self.expression(
+                exp.Into, this=self._parse_table(schema=True), temporary=temp, unlogged=unlogged
+            )
+
         def _parse_id_var(
             self,
             any_token: bool = True,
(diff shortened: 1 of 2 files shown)


georgesittas:
Hey @hhubbell, could we override `_parse_into` such that it delegates to the super class' implementation and then modifies the produced AST?

So right now we do:

```python
>>> import sqlglot
>>> sqlglot.parse_one("SELECT * INTO UNLOGGED #t", "tsql")
Select(
  expressions=[
    Star()],
  into=Into(
    this=Table(
      this=Identifier(this=t, quoted=False, temporary=True)),
    unlogged=True))
```

You could modify the AST for T-SQL in order to add `temporary=True` to the `INTO` node as well. Is the motivation behind this to enable transpilation to Postgres?

hhubbell:
@georgesittas - thanks for the idea! I'll take a look this evening.

The motivation behind this change was transpilation between T-SQL and Snowflake, for T-SQL statements that selected into temporary tables.  Currently, `sqlglot` was transpiling this specific type of syntax without pulling through the TEMPORARY table modifier.

Here's an example:
```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT 1 AS a INTO #foo", read="tsql", write="snowflake")

# Actual
['CREATE TABLE foo AS SELECT 1 AS a']

# Expectation
['CREATE TEMPORARY TABLE foo AS SELECT 1 AS a']
```

And the inverse:
```python
>>> import sqlglot
>>> sqlglot.transpile("CREATE TEMPORARY TABLE foo AS SELECT 1 AS a", read="snowflake", write="tsql")

# Actual
['SELECT * INTO foo FROM (SELECT 1 AS a) AS temp']

# Expectation
['SELECT * INTO #foo FROM (SELECT 1 AS a) AS temp']
```

Postgres supports select...into syntax like T-SQL, which is why I had been using it as an example - apologies if that led to any confusion!

sqlglot/dialects/tsql.py
```diff
@@ -779,6 +779,19 @@ def _parse_user_defined_function(
...
+        def _parse_into(self) -> t.Optional[exp.Into]:
+            into = super()._parse_into()
+
+            if isinstance(into, exp.Into) and isinstance(into.this, exp.Table):
+                table = into.this
+                if table.this and table.this.args.get("temporary"):
+                    # Promote the temporary property from the Identifier
+                    # expression to the Into expression
+                    into.set("temporary", True)
+                    table.this.set("temporary", None)
```
hhubbell, line 791: This last line here strips out the temporary property from the Identifier expression so that the AST mirrors other dialects that use the `select...into` syntax.  It may be worth retaining this property however, and there should be no downstream impact.  All tests will pass whether `Identifier(temporary=True)` in addition to `Into(temporary=True)`, due to logic in `into_sql` that pushes this property back down for the T-SQL dialect.

Happy to make any adjustment necessary.

georgesittas:
I think the direction here is reasonable, I can take this to the finish line.

georgesittas:
Check this out: https://github.com/tobymao/sqlglot/commit/5dac6d3c26c3af32558e59f00449cd6df8a43115

PR #4897, opened 2025-03-20
feat: Add a check and error message for set operations in pushdown_projections

snovik75:
If a set operation has a column number mismatch, then we raise an OptimizeError instead of IndexError in line 68 which might make things confusing regarding the source of the exception

### DIFF
--- a/sqlglot/optimizer/pushdown_projections.py
+++ b/sqlglot/optimizer/pushdown_projections.py
@@ -51,11 +51,8 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
         if isinstance(scope.expression, exp.SetOperation):
             left, right = scope.union_scopes
             if len(left.expression.selects) != len(right.expression.selects):
-                l = set(c.alias_or_name for c in left.expression.selects)
-                r = set(c.alias_or_name for c in right.expression.selects)
-                raise OptimizeError(
-                    f"Invalid set operation due to columns mismatch {l.difference(r) or r.difference(l)}"
-                )
+                scope_sql = scope.expression.sql()
+                raise OptimizeError(f"Invalid set operation due to column mismatch: {scope_sql}.")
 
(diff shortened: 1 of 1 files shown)


georgesittas:
@snovik75 can you show an example that triggers the index error? I'm a little skeptical about this approach so I'd like to understand a bit more.

snovik75:
Sure, here the example with the main

```
sql = """
select 
 a.a,
 a.b
from a
where 1=1
union
select 
 b.a
from b
where 1=1
"""

tree = parse_one(sql)
optimizer.pushdown_projections.pushdown_projections(tree)
```

raises `IndexError: list index out of range`

With the proposed change it would be

`OptimizeError: Invalid set operation due to columns mismatch {'b'}`

snovik75:
Yes, the sql script is broken but the change would highlight the issue directly instead of raising suspicions about some but in the code

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -49,6 +50,13 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
 
         if isinstance(scope.expression, exp.SetOperation):
             left, right = scope.union_scopes
+            if len(left.expression.selects) != len(right.expression.selects):
+                l = set(c.alias_or_name for c in left.expression.selects)
+                r = set(c.alias_or_name for c in right.expression.selects)
+                raise OptimizeError(
+                    f"Invalid set operation due to columns mismatch {l.difference(r) or r.difference(l)}"
+                )
```
georgesittas, line 58:
```suggestion
                scope_sql = scope.expression.sql()
                raise OptimizeError(f"Invalid set operation due to column mismatch: {scope_sql}.")
```

georgesittas (approved):
Looks good, I've simplified the error handling a bit so we don't have to do the set difference.

PR #4906, opened 2025-03-22
fix(Snowflake): quote identifiers in stage references

whummer:
Snowflake distinguishes between quoted/unquoted identifiers in stage names:
* `@mySTAGE` -> gets canonicalized to uppercase name, i.e., references a stage named `"MYSTAGE"`
* `@"mySTAGE"` -> preserves the name in this verbatim casing, i.e., references a stage named `"mySTAGE"`

The same also applies for fully qualified stage references, e.g., `@"myDB"."mySCHEMA"."mySTAGE"` would preserve the casing of the database/schema/stage names.

This PR slightly adjusts the logic, and extends the existing parsing tests in `test_staged_files(..)` to cover these edge cases.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -874,7 +874,12 @@ class Snowflake(Dialect):
             ):
                 parts.append(self._advance_any(ignore_reserved=True))
 
-            return exp.var("".join(part.text for part in parts if part))
+            part_texts = (
+                f'"{part.text}"' if part.token_type == TokenType.IDENTIFIER else part.text
+                for part in parts
+                if part
+            )
+            return exp.var("".join(part_texts))
 
         def _parse_lambda_arg(self) -> t.Optional[exp.Expression]:
             this = super()._parse_lambda_arg()
--- a/tests/dialects/test_snowflake.py
+++ b/tests/dialects/test_snowflake.py
@@ -1064,7 +1064,11 @@ class TestSnowflake(Validator):
         self.validate_identity("SELECT * FROM '@mystage'")
         self.validate_identity("SELECT * FROM @namespace.mystage/path/to/file.json.gz")
         self.validate_identity("SELECT * FROM @namespace.%table_name/path/to/file.json.gz")
+        self.validate_identity('SELECT * FROM @"mystage"')
+        self.validate_identity('SELECT * FROM @"myschema"."mystage"/file.gz')
+        self.validate_identity('SELECT * FROM @"my_DB"."schEMA1".mystage/file.gz')
         self.validate_identity("SELECT * FROM '@external/location' (FILE_FORMAT => 'path.to.csv')")
+        self.validate_identity('PUT \'file:///dir/tmp.csv\' @"my_DB"."schEMA1"."MYstage"')
         self.validate_identity("PUT file:///dir/tmp.csv @%table", check_command_warning=True)
         self.validate_identity("SELECT * FROM (SELECT a FROM @foo)")
         self.validate_identity(


sqlglot/dialects/snowflake.py
```diff
@@ -874,7 +874,12 @@ def _parse_location_path(self) -> exp.Var:
             ):
                 parts.append(self._advance_any(ignore_reserved=True))
 
-            return exp.var("".join(part.text for part in parts if part))
+            part_texts = (
+                f'"{part.text}"' if part.token_type == TokenType.IDENTIFIER else part.text
+                for part in parts
+                if part
+            )
+            return exp.var("".join(part_texts))
```
whummer, line 882: note: I'm generally wondering if we should return a single `exp.Var` with the stringified identifier here (which could have a value like`'@"my_DB"."schEMA1"."MYstage"'`), or whether we could potentially return a more structured type, something like the `exp.Table` that gets created when parsing an expression like `SELECT * FROM "my_DB"."schEMA1"."MYstage"`

```
Table(
      this=Identifier(this='MYstage', quoted=True),
      db=Identifier(this='schEMA1', quoted=True),
      catalog=Identifier(this='my_DB', quoted=True))
```

Curious to get your thoughts on this!

georgesittas: The note in this [section](https://docs.snowflake.com/en/user-guide/querying-stage#required-parameters) says:

> The URI string for an external storage location (Amazon S3, Google Cloud Storage, or Microsoft Azure) must be enclosed in single quotes; however, you can enclose any URI string in single quotes, which allows special characters, including spaces, in location and file names. For example: `Internal: '@~/path 1/file 1.csv' [...]`

So it seems like it should be fine to concatenate the parts' SQL and produce a string instead of a var. I think this aligns with what you said in the first sentence, but we'd produce a `Literal` string instead of a `Var`.

whummer: Ok thanks for the feedback @georgesittas - do you want me to update the PR and return a `Literal` instead of a `Var`..?

georgesittas: Yep! Thanks

sqlglot/dialects/snowflake.py
```diff
@@ -874,7 +874,12 @@ def _parse_location_path(self) -> exp.Var:
             ):
                 parts.append(self._advance_any(ignore_reserved=True))
 
-            return exp.var("".join(part.text for part in parts if part))
+            part_texts = (
+                f'"{part.text}"' if part.token_type == TokenType.IDENTIFIER else part.text
+                for part in parts
+                if part
+            )
+            return exp.Literal(this="".join(part_texts), is_string=False)
```
georgesittas, line 882: This doesn't look right, a `Literal` with `is_string` set to false is a number. In my previous reply I meant that I'd prefer canonicalizing all URIs for staged paths as strings. This is a breaking change, but should be ok given that [Snowflake doesn't mind single quotes](https://github.com/tobymao/sqlglot/pull/4906#discussion_r2009086561).

Additionally, I think it should be possible to simplify the approach here by leveraging `_find_sql`, instead of creating a list of all the tokens that make up the URI path. So, we'd consume all tokens as we do now, but at the end we'd do `_find_sql(start, end)` and then use that text to build the string literal.

whummer: Thanks for clarifying @georgesittas - updated the PR based on your suggestions. 👍 Had to adjust a couple of the existing tests to have `write_sql=...` represent the new output format with quotes.

Regarding the change with `_find_sql(..)` - I've noticed that this call also seems to be consuming the quotes `'` for quoted stage references, hence I had to add a switch that strips of any `'` at the beginning/end of the result string. 

See here: https://github.com/tobymao/sqlglot/pull/4906/files#diff-c3dbaf62f01fc7a9baf342a000c2351ebea03abdede048f92d234c44e11c7bfeR879-R880 I can imaging that this is most certainly not the best way to do this, I'm sure there's a more elegant way to achieve this :) - again, happy for any guidance you can give! 🙌

sqlglot/dialects/snowflake.py
```diff
@@ -863,18 +863,22 @@ def _parse_file_location(self) -> t.Optional[exp.Expression]:
...
             # in case of subqueries
             while self._is_connected() and not self._match_set(
                 (TokenType.COMMA, TokenType.L_PAREN, TokenType.R_PAREN), advance=False
             ):
-                parts.append(self._advance_any(ignore_reserved=True))
+                self._advance_any(ignore_reserved=True)
 
-            return exp.var("".join(part.text for part in parts if part))
+            result = self._find_sql(start, self._prev)
+            if len(result) >= 2 and result[0] == result[-1] == "'":
```
whummer, line 879: These the two lines (879-880) were required (see also the other comment on the PR), as the parsed `result` string can be either an unquoted (`@mystage/foo.txt`) or a quoted string (`'@mystage/foo.txt'`).

Happy for any guidance here if you have a better approach for this in mind!

georgesittas:
@whummer I just took another look at this and realized that your initial approach + my comment about using `_find_sql` can be combined to simplify this PR quite a bit without the risk of changing the AST.

I'll merge this and refactor it to minimize the back-and-forth. As usual, thanks for your contributions!

georgesittas:
Check this out: https://github.com/tobymao/sqlglot/commit/3b7c699267bf4f041a033017a894f0c1e2ae4068

whummer:
Nice approach @georgesittas - looks quite a bit cleaner, and also less changes required in the existing test assertions. 👍  Thanks again for your support in pushing this over the line! 🙌

PR #4926, opened 2025-03-29
CrateDB: Hello, World!

amotl:
A bit of code spiked the other day.
Far from finished, probably.


amotl:
Oopsie, wrong target. Apologies. We will come back later when it's finished.

PR #4930, opened 2025-04-01
fix(clickhouse)!: Fix generation of exp.Values

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/4056

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -180,30 +180,6 @@ def _map_sql(self: ClickHouse.Generator, expression: exp.Map | exp.VarMap) -> st
     return f"{{{csv_args}}}"
 
 
-def generate_values_aliases(expression: exp.Values) -> t.List[str]:
-    # Clickhouse allows VALUES to have an embedded structure e.g:
-    # VALUES('person String, place String', ('Noah', 'Paris'), ...)
-    # In this case, we don't want to qualify the columns
-    values = expression.expressions[0].expressions
-
-    structure = (
-        values[0]
-        if (len(values) > 1 and values[0].is_string and isinstance(values[1], exp.Tuple))
-        else None
-    )
-    if structure:
-        # Split each column definition into the column name e.g:
-        # 'person String, place String' -> ['person', 'place']
-        structure_coldefs = [coldef.strip() for coldef in structure.name.split(",")]
-        column_aliases = [coldef.split(" ")[0] for coldef in structure_coldefs]
-    else:
-        num_cols = len(values[0].expressions) if isinstance(values[0], exp.Tuple) else 0
-        # Default column aliases in CH are "c1", "c2", etc.
-        column_aliases = [f"c{i + 1}" for i in range(num_cols)]
-
-    return column_aliases
-
-
 class ClickHouse(Dialect):
     NORMALIZE_FUNCTIONS: bool | str = False
     NULL_ORDERING = "nulls_are_last"
@@ -230,6 +206,32 @@ class ClickHouse(Dialect):
         exp.Intersect: False,
         exp.Union: None,
     }
+
+    def generate_values_aliases(self, expression: exp.Values) -> t.List[exp.Identifier]:
+        # Clickhouse allows VALUES to have an embedded structure e.g:
+        # VALUES('person String, place String', ('Noah', 'Paris'), ...)
+        # In this case, we don't want to qualify the columns
+        values = expression.expressions[0].expressions
+
+        structure = (
+            values[0]
+            if (len(values) > 1 and values[0].is_string and isinstance(values[1], exp.Tuple))
+            else None
+        )
+        if structure:
+            # Split each column definition into the column name e.g:
+            # 'person String, place String' -> ['person', 'place']
+            structure_coldefs = [coldef.strip() for coldef in structure.name.split(",")]
+            column_aliases = [
+                exp.to_identifier(coldef.split(" ")[0]) for coldef in structure_coldefs
+            ]
+        else:
+            # Default column aliases in CH are "c1", "c2", etc.
+            column_aliases = [
+                exp.to_identifier(f"c{i + 1}") for i in range(len(values[0].expressions))
+            ]
+
+        return column_aliases
 
     class Tokenizer(tokens.Tokenizer):
         COMMENTS = ["--", "#", "#!", ("/*", "*/")]
@@ -938,6 +940,7 @@ class ClickHouse(Dialect):
             # to other dialects. For this case, we canonicalize the values into a tuple-of-tuples AST if it's not already one.
             # In INSERT INTO statements the same clause actually references multiple columns (opposite semantics),
             # but the final result is not altered by the extra parentheses.
+            # Note: Clickhouse allows VALUES([structure], value, ...) so the branch checks for the last expression
             expressions = value.expressions
             if not isinstance(expressions[-1], exp.Tuple):
                 value.set(
@@ -1362,11 +1365,11 @@ class ClickHouse(Dialect):
             # If the VALUES clause contains tuples of expressions, we need to treat it
             # as a table since Clickhouse will automatically alias it as such.
             alias = expression.args.get("alias")
-            column_aliases = (alias and alias.args.get("columns")) or None
-            if not column_aliases:
-                values_as_table = True
-            else:
+
+            if alias and alias.args.get("columns") and expression.expressions:
                 values = expression.expressions[0].expressions
                 values_as_table = any(isinstance(value, exp.Tuple) for value in values)
+            else:
+                values_as_table = True
 
             return super().values_sql(expression, values_as_table=values_as_table)
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -1317,9 +1317,10 @@ class ClickHouse(Dialect):
             return super().not_sql(expression)
 
         def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> str:
-            # Clickhouse does not support VALUES with alias e.g:
-            # SELECT * FROM (VALUES (1, 2, 3)) AS subq(a, b, c)
-            # This gets generated into a subquery with the aliases properly placed
+            # Clickhouse doesn't properly handle VALUES with alias e.g:
+            # SELECT * FROM VALUES (1, 2, 3) AS subq(col) -> Produces column "c1"
+            # We generate this into a subquery with the aliases properly placed:
+            # SELECT * FROM (SELECT 1 AS col) AS subq
             values_as_table = not expression.args.get("alias")
 
             return super().values_sql(expression, values_as_table=values_as_table)
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -180,6 +180,30 @@ def _map_sql(self: ClickHouse.Generator, expression: exp.Map | exp.VarMap) -> st
     return f"{{{csv_args}}}"
 
 
+def generate_values_aliases(expression: exp.Values) -> t.List[str]:
+    # Clickhouse allows VALUES to have an embedded structure e.g:
+    # VALUES('person String, place String', ('Noah', 'Paris'), ...)
+    # In this case, we don't want to qualify the columns
+    values = expression.expressions[0].expressions
+
+    structure = (
+        values[0]
+        if (len(values) > 1 and values[0].is_string and isinstance(values[1], exp.Tuple))
+        else None
+    )
+    if structure:
+        # Split each column definition into the column name e.g:
+        # 'person String, place String' -> ['person', 'place']
+        structure_coldefs = [coldef.strip() for coldef in structure.name.split(",")]
+        column_aliases = [coldef.split(" ")[0] for coldef in structure_coldefs]
+    else:
+        num_cols = len(values[0].expressions) if isinstance(values[0], exp.Tuple) else 0
+        # Default column aliases in CH are "c1", "c2", etc.
+        column_aliases = [f"c{i + 1}" for i in range(num_cols)]
+
+    return column_aliases
+
+
 class ClickHouse(Dialect):
     NORMALIZE_FUNCTIONS: bool | str = False
     NULL_ORDERING = "nulls_are_last"
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -905,6 +905,24 @@ class ClickHouse(Dialect):
                 this = exp.Apply(this=this, expression=self._parse_var(any_token=True))
             return this
 
+        def _parse_value(self) -> t.Optional[exp.Tuple]:
+            value = super()._parse_value()
+            if not value:
+                return None
+
+            # In Clickhouse "SELECT * FROM VALUES (1, 2, 3)" generates a table with a single column, in contrast
+            # to other dialects. For this case, we canonicalize the values into a tuple-of-tuples AST if it's not already one.
+            # In INSERT INTO statements the same clause actually references multiple columns (opposite semantics),
+            # but the final result is not altered by the extra parentheses.
+            expressions = value.expressions
+            if not isinstance(expressions[-1], exp.Tuple):
+                value.set(
+                    "expressions",
+                    [self.expression(exp.Tuple, expressions=[expr]) for expr in expressions],
+                )
+
+            return value
+
     class Generator(generator.Generator):
         QUERY_HINTS = False
         STRUCT_DELIMITER = ("(", ")")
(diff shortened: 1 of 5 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -1315,3 +1315,11 @@ def not_sql(self, expression: exp.Not) -> str:
                 return self.sql(expression, "this")
 
             return super().not_sql(expression)
+
+        def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> str:
+            # Clickhouse does not support VALUES with alias e.g:
+            # SELECT * FROM (VALUES (1, 2, 3)) AS subq(a, b, c)
```
georgesittas, line 1321: FYI it does support it, but the semantics are simply weird because they use auto-generated names like `cX`.

```
:) SELECT * FROM VALUES (1, 2, 3) AS subq(a, b, c);

SELECT *
FROM VALUES(1, 2, 3) AS subq (a, b, c)

Query id: 0837ebf9-ac35-409c-9792-f7d044cb7252

   ┌─c1─┐
1. │  1 │
2. │  2 │
3. │  3 │
   └────┘

3 rows in set. Elapsed: 0.004 sec.
```

georgesittas (commented):
Can you check how this change affects sqlmesh? I believe we do `exp.values` in some places there and I'm not sure if lineage / qualification will work properly, because we rely on named columns in `TableAlias` there.

sqlglot/dialects/clickhouse.py
VaggelisD: Looks like I'm due for another upgrade? 😄 

```SQL
:) SELECT * FROM VALUES (1, 2, 3) AS subq(a, b, c);

Syntax error: failed at position 39 ('('):

SELECT * FROM VALUES (1, 2, 3) AS subq(a, b, c);
```

VaggelisD:
> Can you check how this change affects sqlmesh? I believe we do exp.values in some places there and I'm not sure if lineage / qualification will work properly, because we rely on named columns in TableAlias there.

PR look goods 👍

Do note though that the changes here should be backwards-compatible:
- If there's an alias present, we still use the old generation (1 test above my new ones)
- If there isn't (since `structure` is a "VALUES field") we just generate `VALUES (...)` as is

So any case using `VALUES` with table + column aliases shouldn't be changed.

georgesittas:
Could you take a look at enabling CLL and qualification work for the CH-specific variant with the string schema? I think that's the only missing piece right now?

Thanks for double-checking this.

sqlglot/dialects/clickhouse.py
georgesittas: TODO here– update comment so that it doesn't confuse us in the future.

VaggelisD:
> Could you take a look at enabling CLL and qualification work for the CH-specific variant with the string schema? I think that's the only missing piece right now?

That'd look like going from `SELECT col FROM VALUES('col String', ('foo'))` to `SELECT _t0.col FROM VALUES('col String', ('foo')) AS _t0(col)`, even if the table alias is not registered properly, right?

georgesittas:
I guess it's fine for now, you'd need to parse the first string in values to see if it's a schema which adds overhead. We can get this in and discuss offline about extending it.

Before merging though, I'd like to sanity check something: won't we mess up the roundtrip in SQLMesh due to qualifying tables and adding aliases?

VaggelisD:
> Before merging though, I'd like to sanity check something: won't we mess up the roundtrip in SQLMesh due to qualifying tables and adding aliases?

Good catch, looks like `qualify` also expands the VALUES clause on it's own afaict, I'll see if we can block it for Clickhouse.

sqlglot/dialects/clickhouse.py
```diff
@@ -1315,3 +1315,16 @@ def not_sql(self, expression: exp.Not) -> str:
...
+        def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> str:
+            # If the VALUES clause contains tuples of expressions, we need to treat it
+            # as a table since Clickhouse will automatically alias it as such.
+            alias = expression.args.get("alias")
+            column_aliases = (alias and alias.args.get("columns")) or None
+            if not column_aliases:
+                values_as_table = True
+            else:
+                values = expression.expressions[0].expressions
+                values_as_table = any(isinstance(value, exp.Tuple) for value in values)
```
georgesittas, line 1328:
```suggestion
            if alias and alias.args.get("columns") and expression.expressions:
                values = expression.expressions[0].expressions
                values_as_table = any(isinstance(value, exp.Tuple) for value in values)
            else:
                values_as_table = True
```

georgesittas: So, to test my understanding: `VALUES` will always be treated as a "table" (i.e. not converted to a query), unless it has named columns in its `TableAlias`. It seems like that case would correspond to a manually constructed query and not one that matches ClickHouse's "provide schema as string" semantics.

I think this approach is reasonable, but would also pay attention to the case where you have (a) both named columns and (b) the first argument in the `VALUES`' expressions is a string and the rest are tuples– so your existing `any` check will be useful to verify the 2nd property.

I think the way to deal with this case is to amend the column names in the `TableAlias` by extracting them from the string, like you did in `qualify_tables`.

sqlglot/optimizer/qualify_tables.py
```diff
@@ -127,8 +127,42 @@ def _qualify(table: exp.Table) -> None:
                 if not table_alias.name:
                     table_alias.set("this", exp.to_identifier(next_alias_name()))
                 if isinstance(udtf, exp.Values) and not table_alias.columns:
-                    for i, e in enumerate(udtf.expressions[0].expressions):
-                        table_alias.append("columns", exp.to_identifier(f"_col_{i}"))
+                    values = udtf.expressions[0].expressions
+                    if dialect == "clickhouse":
+                        # Clickhouse allows VALUES to have an embedded structure e.g:
+                        # VALUES('person String, place String', ('Noah', 'Paris'), ...)
+                        # In this case, we don't want to qualify the columns
```
georgesittas, line 134: I'm skeptical about putting this here, it pollutes the logic quite a bit. Can we try fixing the `TableAlias` at parse time instead and maybe ignoring it at generation time for ClickHouse if the `VALUES` contains a schema string? This would avoid the need to fix this here.

tests/dialects/test_clickhouse.py
```diff
@@ -591,6 +591,42 @@ def test_clickhouse(self):
...
+        def qualify_values(sql):
+            from sqlglot.optimizer.qualify import qualify
+
+            return qualify(parse_one(sql, read="clickhouse"), dialect="clickhouse").sql(
+                "clickhouse"
+            )
+
+        self.assertEqual(
+            qualify_values("SELECT * FROM (SELECT 1 AS a, 2 AS b, 3 AS c) AS subq"),
+            'SELECT "subq"."a" AS "a", "subq"."b" AS "b", "subq"."c" AS "c" FROM (SELECT 1 AS "a", 2 AS "b", 3 AS "c") AS "subq"',
```
georgesittas, line 605: You can put these in the `qualify_columns.sql` fixture and provide a dialect.

tests/dialects/test_clickhouse.py
```diff
@@ -591,6 +591,42 @@ def test_clickhouse(self):
...
+        )
+
+        self.assertEqual(
+            qualify_values(
+                "SELECT * FROM VALUES ('person String, place String', ('Noah', 'Paris'))"
+            ),
+            """SELECT "_q_0"."person" AS "person", "_q_0"."place" AS "place" FROM VALUES ('person String, place String', ('Noah', 'Paris')) AS "_q_0"("person", "place")""",
```
georgesittas, line 612: Nice!

tests/dialects/test_clickhouse.py
```diff
@@ -591,6 +591,42 @@ def test_clickhouse(self):
...
+        )
+
+        self.assertEqual(
+            qualify_values("SELECT * FROM VALUES ((1, 1), (2, 2))"),
+            'SELECT "_q_0"."c1" AS "c1", "_q_0"."c2" AS "c2" FROM VALUES ((1, 1), (2, 2)) AS "_q_0"("c1", "c2")',
+        )
+
+        self.assertEqual(
+            qualify_values("SELECT * FROM VALUES (1, 2, 3)"),
+            'SELECT * FROM VALUES (1, 2, 3) AS "_q_0"',
```
georgesittas, line 622: FYI we do know that there's one column named `c1` in this case:

```
SELECT c1
FROM values(1, 2, 3)

Query id: 6d5f6632-5ba5-470b-9edc-809c0255314a

   ┌─c1─┐
1. │  1 │
2. │  2 │
3. │  3 │
   └────┘
```

tests/dialects/test_clickhouse.py
```diff
@@ -591,6 +591,42 @@ def test_clickhouse(self):
...
+        )
+
+        self.assertEqual(
+            qualify_values("SELECT * FROM (VALUES (1, 2, 3)) AS subq(a, b, c)"),
+            'SELECT "subq"."a" AS "a", "subq"."b" AS "b", "subq"."c" AS "c" FROM (SELECT 1 AS "a", 2 AS "b", 3 AS "c") AS "subq"',
+        )
+
```
georgesittas, line 629: This test is incorrect for two reasons:

- Parentheses can't be used around `VALUES`
- The semantics aren't preserved in this transformation:

```
:) SELECT * FROM VALUES (1, 2, 3) AS subq(a, b, c); -- Before

SELECT *
FROM VALUES(1, 2, 3) AS subq (a, b, c)

Query id: 02832be2-fc59-492e-b69e-dd4bdacb0613

   ┌─c1─┐
1. │  1 │
2. │  2 │
3. │  3 │
   └────┘

3 rows in set. Elapsed: 0.002 sec.
:) SELECT "subq"."a" AS "a", "subq"."b" AS "b", "subq"."c" AS "c" FROM (SELECT 1 AS "a", 2 AS "b", 3 AS "c") AS "subq"; -- after

SELECT
    subq.a AS a,
    subq.b AS b,
    subq.c AS c
FROM
(
    SELECT
        1 AS a,
        2 AS b,
        3 AS c
) AS subq

Query id: d1ce79ea-4f69-4c19-a26f-fafbb634c21b

   ┌─a─┬─b─┬─c─┐
1. │ 1 │ 2 │ 3 │
   └───┴───┴───┘

1 row in set. Elapsed: 0.002 sec.
```

VaggelisD: Yeah this was intentional, iirc this was a Postgres/DuckDB equivalent test but it should either be documented as such or use the relevant `parse_one` dialect.

sqlglot/dialects/clickhouse.py
VaggelisD: > So, to test my understanding: VALUES will always be treated as a "table" (i.e. not converted to a query), unless it has named columns in its TableAlias. It seems like that case would correspond to a manually constructed query and not one that matches ClickHouse's "provide schema as string" semantics. Correct?

That was my thinking as well, correct. If there are column aliases present but not a `structure` schema string, I'd expect that this is a transpiled query, not an intentional CH one.

sqlglot/optimizer/qualify_tables.py
VaggelisD: The logic right above this part has to do with constructing a `TableAlias` if it doesn't exist, and only then we can set the columns; Do we want to bring that logic too in CH's parse time? Otherwise, the solution might be incomplete afaict.

VaggelisD:
There are 2 open issues at hand:
1. This PR doesn't expand `SELECT * FROM VALUES (1, 2, 3)` into `SELECT c1 FROM ...` as it does with other cases ([discussion](https://github.com/tobymao/sqlglot/pull/4930#discussion_r2029058801)) 
2. It's valid to run queries like `SELECT * FROM VALUES ('col1 int, col2 int', (1, 1)) AS q0 (c1, c2)` which now get incorrectly expanded using the `TableAlias` columns  ([discussion](https://github.com/tobymao/sqlglot/pull/4930#discussion_r2029047509)) 


<br />

Imo, it looks like both issues are not worth the complexity currently. To explain:

To solve (1), one idea was to canonicalize the AST into a tuple of tuples (representing 1 column) but that impacts existing tests and roundtrips. More importantly, it looks to me that the [INSERT INTO ... FROM VALUES](https://clickhouse.com/docs/sql-reference/statements/insert-into#inserting-the-results-of-select) semantics differ from the `SELECT .. FROM VALUES` in that `VALUES (1, 2, 3)` represents 3 columns in the former but 1 in the latter.

Solving (2) is more nuanced but imo it's somewhat of a garbage in, garbage out situtation; Clickhouse completely ignores the column aliases (even though it doesn't yield a parse error) so I don't see a reason a user would actually do that.

If this PR covers the issue and a bit more, I think I'd look into moving out the `qualify_tables.py` logic into `clickhouse.py` and merging it as it is.

tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -435,6 +435,25 @@ WITH tbl1 AS (SELECT STRUCT(1 AS col1, 2 AS col1) AS col) SELECT tbl1.col.* FROM
...
+SELECT _q_0.person AS person, _q_0.place AS place FROM VALUES ('person String, place String', ('Noah', 'Paris')) AS _q_0(person, place);
+
+# dialect: clickhouse
+# Title: Expand tuples in VALUES using the default naming scheme in CH
+# execute: false
+SELECT * FROM VALUES ((1, 1), (2, 2));
+SELECT _q_0.c1 AS c1, _q_0.c2 AS c2 FROM VALUES ((1, 1), (2, 2)) AS _q_0(c1, c2);
+
+# dialect: clickhouse
+# Title: Do not expand columns in VALUES if not aliased properly
```
georgesittas, line 451: This title is now outdated.

tests/fixtures/optimizer/qualify_columns.sql
georgesittas, line 1: Can we also add a test for the `VALUES` -> subquery case?

georgesittas (commented):
Thanks for addressing comments. A few final comments from me and should be good to go.

sqlglot/dialects/clickhouse.py
```diff
@@ -180,6 +180,30 @@ def _map_sql(self: ClickHouse.Generator, expression: exp.Map | exp.VarMap) -> st
...
+    )
+    if structure:
+        # Split each column definition into the column name e.g:
+        # 'person String, place String' -> ['person', 'place']
+        structure_coldefs = [coldef.strip() for coldef in structure.name.split(",")]
+        column_aliases = [coldef.split(" ")[0] for coldef in structure_coldefs]
+    else:
+        num_cols = len(values[0].expressions) if isinstance(values[0], exp.Tuple) else 0
+        # Default column aliases in CH are "c1", "c2", etc.
+        column_aliases = [f"c{i + 1}" for i in range(num_cols)]
```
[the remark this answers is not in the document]
VaggelisD: That looks outdated after the tuple-of-tuples canonicalization, I think this was the case before that'd protect against aliasing `SELECT * FROM VALUES (1, 2, 3)`

sqlglot/dialects/clickhouse.py
```diff
@@ -905,6 +929,24 @@ def _parse_columns(self) -> exp.Expression:
...
+            value = super()._parse_value()
+            if not value:
+                return None
+
+            # In Clickhouse "SELECT * FROM VALUES (1, 2, 3)" generates a table with a single column, in contrast
+            # to other dialects. For this case, we canonicalize the values into a tuple-of-tuples AST if it's not already one.
+            # In INSERT INTO statements the same clause actually references multiple columns (opposite semantics),
+            # but the final result is not altered by the extra parentheses.
+            expressions = value.expressions
+            if not isinstance(expressions[-1], exp.Tuple):
```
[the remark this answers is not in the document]
VaggelisD: I guess the idea is that we have two variants of `VALUES`:
1. The tuple-based one which is of the form `VALUES ([structure], (...), (...))`
2. The flat field one which is of the form `VALUES (1, 2, 3)`

The confusing part is that (1) may throw off such conditions since `structure` is a string among tuples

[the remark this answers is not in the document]
georgesittas: Ok, sounds reasonable. Let's include comment with a short version of this explanation.

tests/fixtures/optimizer/qualify_columns.sql
VaggelisD: What'd you have in mind? This should be a transpilation case so I'm wondering if we can reuse the existing test in `test_clickhouse.py::test_clickhouse_values`

georgesittas: Ah, you're right. Don't worry about it then for now. I guess we should roundtrip / handle CH queries in qualify so that's ok.

sqlglot/dialects/clickhouse.py
VaggelisD: Sorry, thought the suggestion had to do with the initial concern which we decided not to deal with right now.

georgesittas (approved):

PR #4936, opened 2025-04-03
Update tsql.py

galenseilisnh:
Report length of args when assertion fails.

---

I am getting this error, and having the assertion tell me the actual length might be useful.

---

Alternative implementation could instead be to actually show the value of `args`.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -109,7 +109,7 @@ def _build_formatted_time(
     exp_class: t.Type[E], full_format_mapping: t.Optional[bool] = None
 ) -> t.Callable[[t.List], E]:
     def _builder(args: t.List) -> E:
-        assert len(args) == 2
+        assert len(args) == 2, f"{len(args)=} but should equal 2."
 
         return exp_class(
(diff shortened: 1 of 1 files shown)


georgesittas:
I don't think we should be asserting here at all, SQLGlot's parser is intentionally lenient so that you can use `sqlglot.ErrorLevel.IGNORE`. This is most likely an oversight– I'll address in a separate PR, but thanks for your contribution!

georgesittas:
Check out https://github.com/tobymao/sqlglot/commit/09882e32f057670a9cbd97c1e5cf1a00c774b5d2.

Imran-imtiaz48 (commented):
This code snippet is clear and well-structured, but there are a couple of points to consider. The assertion for the argument length is helpful for debugging, but you might want to raise a more descriptive exception for better error handling in production. Additionally, `exp.cast(args[1], exp.DataType.Type.DATETIME2)` assumes that the second element is always a valid input for casting, so adding a type or value check before casting could improve robustness. Overall, the function is concise and effectively maps inputs to a structured expression class, but a bit more defensive programming would enhance its reliability.

Issue #3481, opened 2024-05-15
[bug] pgsql to mysql special character || Semantic mismatch


Toms1999:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="spark")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="duckdb")`
- Check if the issue still exists on main

**Fully reproducible code snippet**
Please include a fully reproducible code snippet or the input sql, dialect, and expected output.

**Official Documentation**
Please include links to official SQL documentation related to your issue.

Toms1999:
pgsql:
select '11' || '22'   
return: 1122
mysql:
select '11' || '22'
resurn:1

PR #4943, opened 2025-04-04
Feat!: add support for coalesce semantics in DPipe concatenation

sleshJdev:
In oracle 

```
select null || 'Hello'
```
returns `Hello`

In ClickHouse it should produce same result, but default DPipe will cause Null instead.

By this PR we replace DPIle with concat(coalesce=True) which makes the trick.

Also, I apply concat(coalesce=True) only to non-Literal expression to make final sql less verbose.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -1155,6 +1155,12 @@ class ClickHouse(Dialect):
 
             return prefix + self.func("has", arr.this.unnest(), this)
 
+        def dpipe_sql(self, expression: exp.DPipe) -> str:
+            for e in expression.flatten():
+                if not isinstance(e, exp.Literal):
+                    e.replace(exp.func("coalesce", e, exp.Literal.string("")))
+            return super().dpipe_sql(expression)
+
         def eq_sql(self, expression: exp.EQ) -> str:
             return self._any_to_has(expression, super().eq_sql)
 
--- a/tests/dialects/test_oracle.py
+++ b/tests/dialects/test_oracle.py
@@ -5,6 +5,18 @@ from tests.dialects.test_dialect import Validator
 class TestOracle(Validator):
     dialect = "oracle"
 
+    def test_string_concat(self):
+        self.validate_identity("SELECT CONCAT('abcde', 2, NULL, 22)")
+
+        self.validate_all(
+            "a || b",
+            write={
+                "": "a || b",
+                "clickhouse": "COALESCE(a, '') || COALESCE(b, '')",
+                "oracle": "a || b",
+            },
+        )
+
     def test_oracle(self):
         self.validate_identity("1 /* /* */")
         self.validate_all(


georgesittas:
@sleshJdev please provide context when creating issues or PRs.

georgesittas:
Just saw the description update, all good1 :)

sqlglot/dialects/clickhouse.py
```diff
@@ -1105,6 +1105,15 @@ class Generator(generator.Generator):
...
         }
 
+        def dpipe_sql(self, expression: exp.DPipe) -> str:
+            return self.concat_sql(
+                exp.Concat(
+                    expressions=expression.flatten(),
+                    safe=expression.args.get("safe"),
+                    coalesce=expression.args.get("coalesce"),
+                )
+            )
```
georgesittas, line 1115: Why is this necessary? This changes all `||` expressions in ClickHouse to `CONCAT`s at generation time, right?

tests/dialects/test_clickhouse.py
```diff
@@ -12,6 +12,10 @@ class TestClickhouse(Validator):
     dialect = "clickhouse"
 
     def test_clickhouse(self):
+        self.validate_all(
+            "SELECT CONCAT('ABC', COALESCE(NAME, ''))", read={"oracle": "SELECT 'ABC' || NAME"}
+        )
```
georgesittas, line 17: Always make sure to test roundtrip when adding tests. So your test should look like:

```python
self.validate_all(
    transpiled
    read={
        "oracle": input,
    },
    write={
        "clickhouse": roundtripped_transpiled,
        "oracle": transpiled_to_oracle,  # good to have (optional), tests for ClickHouse -> Oracle, if supported
    },
)
```

georgesittas (commented):
Direction looks reasonable, thanks for the PR 👍

sqlglot/dialects/clickhouse.py
georgesittas: I don't think we should do this. We do it in MySQL because `||` is the logical "or" operator, and not string concatenation ([context](https://github.com/tobymao/sqlglot/issues/3481)).

Let's instead fix the base generator's `dpipe_sql` so that it handles the coalesce logic.

tests/dialects/test_clickhouse.py
georgesittas: Thanks. Another request– can you please move the test under `test_clickhouse`? Group it with the other `validate_all` checks for consistency (~L206).

sqlglot/dialects/clickhouse.py
```diff
@@ -1155,6 +1155,12 @@ def _any_to_has(
 
             return prefix + self.func("has", arr.this.unnest(), this)
 
+        def dpipe_sql(self, expression: exp.DPipe) -> str:
+            for e in expression.flatten():
+                if not isinstance(e, exp.Literal):
+                    e.replace(exp.func("coalesce", e, exp.Literal.string("")))
+            return super().dpipe_sql(expression)
```
georgesittas, line 1162: I'm confused, why are we preemptively doing this for ever `DPipe` in ClickHouse now? This is incorrect.

```python
>>> import sqlglot
>>> sqlglot.transpile("SELECT NULL || 'Hello';", "clickhouse")
["SELECT COALESCE(NULL, '') || 'Hello'"]
```

The input and the output query do different things:

```
:) SELECT NULL || 'Hello';

SELECT concat(NULL, 'Hello')

Query id: 4fca4aec-7762-44ab-ba70-493ee5e5bac1

   ┌─concat(NULL, 'Hello')─┐
1. │ ᴺᵁᴸᴸ                  │
   └───────────────────────┘

1 row in set. Elapsed: 0.005 sec.
:) SELECT COALESCE(NULL, '') || 'Hello';

SELECT concat(COALESCE(NULL, ''), 'Hello')

Query id: 0872629d-3c8f-4999-a359-7ba972ecc7c2

   ┌─concat(COALE⋯), 'Hello')─┐
1. │ Hello                    │
   └──────────────────────────┘

1 row in set. Elapsed: 0.006 sec.
```

sqlglot/dialects/clickhouse.py
georgesittas: That flag is used for transpilation purposes. Dialects like Oracle should set it to `True` because `NULL` is implicitly coalesced with an empty string when concatenating it with strings.

> [...] why it's used like this `not self.dialect.CONCAT_COALESCE and expression.args.get("coalesce")`

This condition holds when transpiling a `Concat` that implicitly coalesces to a dialect where this isn't the case. So if you transpile from Oracle to ClickHouse, you need to `COALESCE` unknown expressions to handle nulls, but if you're transpiling Oracle to DuckDB, which has the same behavior for `CONCAT`, you don't need to do this because it'd be redundant.

georgesittas: This flag controls the behavior of `CONCAT`, not `||`. As you've found out, Postgres implements different semantics for the `||` operator, however the flag is valid for the function version:

```
georgesittas=# select concat('a', null, 'b');
 concat
--------
 ab
(1 row)

georgesittas=# select 'a' || null || 'b';
 ?column?
----------

(1 row)
```

So to properly transpile both versions– function and binary operator– you'll need another flag for the latter since you can't reuse `CONCAT_COALESCE` for both in Postgres.

sqlglot/dialects/clickhouse.py
georgesittas: You need to properly parse the `||` in Oracle so that we know it uses the "coalesce" semantics, and then the generator(s) of `||` can handle that appropriately.

georgesittas:
@sleshJdev any plans to finish this?

PR #4950, opened 2025-04-06
feat(snowflake): parse SHOW FILE FORMATS

tekumara:
return exp.Show instead of exp.Command

see https://docs.snowflake.com/en/sql-reference/sql/show-file-formats


georgesittas (approved):

georgesittas:
Thanks for your contributions @tekumara!

PR #4962, opened 2025-04-09
Fix(athena): Correctly handle PartitionByProperty when it contains Iceberg transforms

erindru:
This addresses part of SQLMesh issue: https://github.com/TobikoData/sqlmesh/issues/4090

Given the following:
```
>>> prop = exp.PartitionedByProperty(this=exp.Schema(expressions=[exp.func("bucket", 4, exp.column("foo"))]))
```

Prior to this PR, when generating against Athena's Trino engine, the value would be truncated:
```
>>> prop.sql(dialect="athena")
"partitioned_by=ARRAY['bucket']" #should be `partitioned_by=ARRAY['bucket(4, foo)']`
```

The same value works correctly in Hive mode:
```
>>> prop.sql(dialect="hive")
'PARTITIONED BY (BUCKET(4, foo))'
```

However, simply fixing the truncation is not enough. The arguments are swapped when running in Hive mode vs Trino Mode. The argument order is documented in the AWS docs [here for Hive](https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html#querying-iceberg-partitioning) and [here for Trino](https://docs.aws.amazon.com/athena/latest/ug/create-table-as.html#ctas-table-properties).

This PR does the following:
 - Fix the truncation in the **Presto** generator (which is a superclass of Athena's Trino generator). Presto and Trino both support the Iceberg partition transform syntax and the current truncation is a bug for them as well.
 - Reorder the arguments for the `bucket` and `truncate` Iceberg transforms to be correct for whatever Athena generator is being used (Hive or Trino)

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -58,6 +58,40 @@ def _location_property_sql(self: Athena.Generator, e: exp.LocationProperty):
     return f"{prop_name}={self.sql(e, 'this')}"
 
 
+def _reorder_iceberg_partition_transforms(
+    prop: exp.PartitionedByProperty, mode: t.Literal["hive", "trino"]
+) -> exp.PartitionedByProperty:
+    # Check for Iceberg partition transforms (bucket / truncate) and ensure their arguments are in the right order
+    # For Hive, it's `bucket(<num buckets>, <col name>)` or `truncate(<num_chars>, <col_name>)`
+    # For Trino, it's reversed - `bucket(<col name>, <num buckets>)` or `truncate(<col_name>, <num_chars>)`
+    #
+    # Hive ref: https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html#querying-iceberg-partitioning
+    # Trino ref: https://docs.aws.amazon.com/athena/latest/ug/create-table-as.html#ctas-table-properties
+
+    def _reorder(e: exp.Expression) -> exp.Expression:
+        if (
+            isinstance(e, exp.Func)
+            and e.text("this").lower() in {"bucket", "truncate"}
+            and len(e.expressions) == 2
+        ):
+            first_expression_is_integer_literal = (
+                isinstance(e.expressions[0], exp.Literal) and e.expressions[0].is_number
+            )
+
+            if (mode == "hive" and not first_expression_is_integer_literal) or (
+                mode == "trino" and first_expression_is_integer_literal
+            ):
+                # swap argument order
+                e.expressions[0], e.expressions[1] = e.expressions[1], e.expressions[0]
+
+        return e
+
+    if isinstance(prop.this, exp.Schema):
+        prop.this.transform(_reorder, copy=False)
+
+    return prop
+
+
 def _partitioned_by_property_sql(self: Athena.Generator, e: exp.PartitionedByProperty) -> str:
     # If table_type='iceberg' then the table property for partitioning is called 'partitioning'
     # If table_type='hive' it's called 'partitioned_by'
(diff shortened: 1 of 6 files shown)


sqlglot/dialects/athena.py
```diff
@@ -58,6 +58,40 @@ def _location_property_sql(self: Athena.Generator, e: exp.LocationProperty):
...
+    prop: exp.PartitionedByProperty, mode: t.Literal["hive", "trino"]
+) -> exp.PartitionedByProperty:
+    # Check for Iceberg partition transforms (bucket / truncate) and ensure their arguments are in the right order
+    # For Hive, it's `bucket(<num buckets>, <col name>)` or `truncate(<num_chars>, <col_name>)`
+    # For Trino, it's reversed - `bucket(<col name>, <num buckets>)` or `truncate(<col_name>, <num_chars>)`
+    #
+    # Hive ref: https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html#querying-iceberg-partitioning
+    # Trino ref: https://docs.aws.amazon.com/athena/latest/ug/create-table-as.html#ctas-table-properties
+
+    def _reorder(e: exp.Expression) -> exp.Expression:
```
tobymao, line 71: i don't think this is right, you just want to alter the generator of the property, not do this complex transform

tests/dialects/test_athena.py
```diff
@@ -223,7 +283,11 @@ def test_ctas(self):
         )
         self.assertEqual(
             ctas_hive.sql(dialect=self.dialect, identify=True),
-            "CREATE TABLE \"foo\".\"bar\" WITH (format='parquet', external_location='s3://foo', partitioned_by=ARRAY['partition_col']) AS SELECT 1",
+            "CREATE TABLE \"foo\".\"bar\" WITH (format='parquet', external_location='s3://foo', partitioned_by=ARRAY['\"partition_col\"']) AS SELECT 1",
```
georgesittas, line 286: Why did double quotes get added here? Is this correct?

sqlglot/dialects/presto.py
```diff
@@ -57,8 +57,14 @@ def _no_sort_array(self: Presto.Generator, expression: exp.SortArray) -> str:
...
-        columns = ", ".join(f"'{c.name}'" for c in expression.expressions)
-        return f"ARRAY[{columns}]"
+        column_literals = []
+        for c in expression.expressions:
+            if isinstance(c, exp.Func):
+                column_literals.append(self.sql(c))
+            else:
+                column_literals.append(self.sql(c, "this"))
+
+        return self.sql(exp.Array(expressions=[exp.Literal.string(c) for c in column_literals]))
```
georgesittas, line 67: Is this right? Why are we converting everything into a string at the end? wouldn't that convert function calls, which have already been generated, to string literals within the array?

sqlglot/dialects/athena.py
erindru: This is just reordering the arguments before it hits the func generator.

I didnt think this logic belonged in the generator for `exp.Func` because it's very specific:
 - It only applies to `exp.Func` instances that are nested inside an `exp.Schema` that is itself nested inside  `exp.PartitionByProperty`
 - It only applies to Athena
 - It only applies to two very specific functions that are used in Iceberg partition transforms, `bucket` and `truncate`

sqlglot/dialects/presto.py
georgesittas: Ok sounds good, thanks for checking 👍

tests/dialects/test_athena.py
erindru: Yes, this actually fixes a bug. `identify=True` which means that `exp.Column` objects should be quoted.

Previously, any `exp.Column` objects present in a partitioning expression would not be quoted. I also verified this syntax on Athena and will doublecheck on Trino

georgesittas (approved):
This looks good to me, although I would clean up a couple of things. If @tobymao's good with getting this in I can do a quick refactor afterwards and share it with you folks.

erindru:
Closing in favour of a different approach that factors Iceberg partition transforms into their own AST nodes so they can be handled as first-class AST nodes instead of anonymous functions.

Issue #4944, opened 2025-04-05
[DuckDB] Parser errors with DuckDB's SQL Standard `PIVOT` Syntax


nph:
I've run into a couple of parser errors using [DuckDB's SQL Standard PIVOT syntax](https://duckdb.org/docs/stable/sql/statements/pivot.html#sql-standard-pivot-syntax) (i.e. the non-simplified version). The example statements below are taken from DuckDB's own [examples](https://duckdb.org/docs/stable/sql/statements/pivot.html#examples) for the standard PIVOT syntax.

Firstly, specifying a `GROUP BY` clause results in a parse error:
```
import sqlglot

sql = '''
    SELECT *
    FROM cities
    PIVOT (
        sum(population)
        FOR
            year IN (2000, 2010, 2020)
        GROUP BY country
    );
'''

# This will raise "ParseError: Expecting ). Line 8, Col: 9."
# Removing the GROUP BY clause results in a successful parse.
sqlglot.parse(sql, read='duckdb')
```

Secondly, specifying multiple `IN` expressions results in a parse error:
```
import sqlglot

sql = '''
    SELECT *
    FROM cities
    PIVOT (
        sum(population) AS total,
        count(population) AS count
        FOR
            year IN (2000, 2010)
            country IN ('NL', 'US')
    );
'''

# This will raise "ParseError: Expecting ). Line 9, Col: 8."
# Removing the second IN expression results in a successful parse.
sqlglot.parse(sql, read='duckdb')
```

This was tested on `26.12.1`.

PR #4964, opened 2025-04-10
feat(duckdb)!: Add support for PIVOT multiple IN clauses

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/4944


sqlglot/dialects/spark2.py
```diff
@@ -237,7 +239,7 @@ def _parse_drop_column(self) -> t.Optional[exp.Drop | exp.Command]:
 
         def _pivot_column_names(self, aggregations: t.List[exp.Expression]) -> t.List[str]:
             if len(aggregations) == 1:
-                return [""]
+                return []
```
georgesittas, line 242: Was this a bug?

sqlglot/generator.py
```diff
@@ -2062,7 +2062,7 @@ def pivot_sql(self, expression: exp.Pivot) -> str:
         alias = self.sql(expression, "alias")
         alias = f" AS {alias}" if alias else ""
 
-        field = self.sql(expression, "field")
+        fields = self.expressions(expression, "fields", sep=" ", flat=True)
```
georgesittas, line 2065: I think it should be small lift to pretty-format this as well. I know we've been doing `flat=True` all over the place but for non-DDL/DML expressions (e.g. properties), it's worth to test the formatting as well.

For example, I recently changed `tuple_sql` to:

```python
    def tuple_sql(self, expression: exp.Tuple) -> str:
        return f"({self.expressions(expression, dynamic=True, new_line=True, skip_first=True, skip_last=True)})"
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -141,9 +141,9 @@ def validate_qualify_columns(expression: E) -> E:
 
 def _unpivot_columns(unpivot: exp.Pivot) -> t.Iterator[exp.Column]:
     name_column = []
```
georgesittas, line 143: You could do this in a single pass. Should we rename to use plural now that there are multiple fields?
```suggestion
    name_column = [
        field.this
        for field in unpivot.fields
        if isinstance(field, exp.In) and isinstance(field.this, exp.Column)
    ]
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -615,11 +615,13 @@ def _expand_stars(
...
             pivot_output_columns = [c.output_name for c in _unpivot_columns(pivot)]
 
-            field = pivot.args.get("field")
-            if isinstance(field, exp.In):
-                pivot_exclude_columns = {
-                    c.output_name for e in field.expressions for c in e.find_all(exp.Column)
-                }
+            fields = pivot.args.get("fields")
+            if fields:
+                for field in fields:
```
georgesittas, line 620:
```suggestion
            for field in pivot.args.get("fields") or []:
                ...
```

sqlglot/parser.py
```diff
@@ -4242,7 +4243,13 @@ def _parse_pivot(self) -> t.Optional[exp.Pivot]:
         if not self._match(TokenType.FOR):
             self.raise_error("Expecting FOR")
 
-        field = self._parse_pivot_in()
+        fields = []
+        while True:
+            field = self._try_parse(self._parse_pivot_in)
+            if not field:
+                break
+            fields.append(field)
```
georgesittas, line 4251: Should we do `while self._curr and self._match(TokenType.IN, advance=False)` instead of `_try_parse`?

sqlglot/parser.py
```diff
@@ -4268,26 +4275,41 @@ def _parse_pivot(self) -> t.Optional[exp.Pivot]:
             names = self._pivot_column_names(t.cast(t.List[exp.Expression], expressions))
 
             columns: t.List[exp.Expression] = []
-            pivot_field_expressions = pivot.args["field"].expressions
+            all_fields = []
+            for pivot_field in pivot.fields:
+                pivot_field_expressions = pivot_field.expressions
+
+                # The `PivotAny` expression corresponds to `ANY ORDER BY <column>`; we can't infer in this case.
+                if not isinstance(seq_get(pivot_field_expressions, 0), exp.PivotAny):
```
georgesittas, line 4283: Maybe let's do `if isinstance(...): continue` to reduce the nesting level here.

sqlglot/parser.py
```diff
@@ -4268,26 +4275,41 @@ def _parse_pivot(self) -> t.Optional[exp.Pivot]:
...
-                        columns.append(exp.to_identifier(name))
+                    columns.append(exp.to_identifier("_".join(fld_parts)))
 
             pivot.set("columns", columns)
 
         return pivot
 
     def _pivot_column_names(self, aggregations: t.List[exp.Expression]) -> t.List[str]:
-        return [agg.alias for agg in aggregations]
+        return list(filter(None, [agg.alias for agg in aggregations]))
```
georgesittas, line 4312: How can you ever get `None` in this list? The `alias` property returns strings.

georgesittas (commented):
Nice work as always @VaggelisD, can you please test `qualify` too? Let's show that stars are expanded correctly.

sqlglot/dialects/spark2.py
VaggelisD: Not really, the old code would check against each `name` element too; But for this version, having a "false-positive" `names` array complicates things further

sqlglot/parser.py
VaggelisD: This filters any falsey value

georgesittas: Let's do this with a list comprehension instead, this isn't as obvious.

sqlglot/generator.py
VaggelisD: Yeah that works nicely with the IN clauses

sqlglot/parser.py
VaggelisD: That doesn't seem to work, I ended up going with the `try_parse` route because:
- Each `exp.In` clause is of the form `<column> IN (<values>)` --> `TokenType.IN` comes at a later point
- They're separated by whitespace which the tokenizer consumes --> We can't use `parse_csv`
- There might be other clauses following such as `GROUP BY` / `ORDER BY` etc --> It's not easy to deduce the following set

sqlglot/optimizer/qualify_columns.py
```diff
@@ -615,13 +616,11 @@ def _expand_stars(
...
-                for field in fields:
-                    if isinstance(field, exp.In):
-                        pivot_exclude_columns = {
-                            c.output_name for e in field.expressions for c in e.find_all(exp.Column)
-                        }
+            for field in pivot.fields:
+                if isinstance(field, exp.In):
+                    pivot_exclude_columns = {
+                        c.output_name for e in field.expressions for c in e.find_all(exp.Column)
+                    }
```
georgesittas, line 623: This is going to overwrite the previous dict in each loop iteration, probably need to update instead?

tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -435,6 +435,12 @@ WITH tbl1 AS (SELECT STRUCT(1 AS col1, 2 AS col1) AS col) SELECT tbl1.col.* FROM
 SELECT * FROM READ_CSV('file.csv');
 SELECT * FROM READ_CSV('file.csv') AS _q_0;
 
+# title: Expand PIVOT column combinations
+# dialect: duckdb
+# execute: false
```
georgesittas, line 440: You can actually execute this, right?

georgesittas (approved):
Couple of comments, LGTM otherwise.

sqlglot/optimizer/qualify_columns.py
VaggelisD: Oh whoops, brain fart moment

tests/fixtures/optimizer/qualify_columns.sql
VaggelisD: The issue here is that by normalizing DuckDB's identifiers to lower case, the df's are "different" in the fact that the previous query would return columns such as `US_2010_Seattle_count` when the equivalent one in the optimized version is `us_2010_seattle_count`

georgesittas: We could make the names compatible between the two so there's no such diff, up to you.

georgesittas (approved):
🚀

PR #4970, opened 2025-04-13
fix(unnest_subqueries): avoid adding GROUP BY on aggregate projection…

skadel:
…s in lateral subqueries

### DIFF
--- a/sqlglot/optimizer/unnest_subqueries.py
+++ b/sqlglot/optimizer/unnest_subqueries.py
@@ -105,7 +105,7 @@ def unnest(select, parent_select, next_alias_name):
                 .from_(select.subquery("_q", copy=False), copy=False)
                 .group_by(exp.column(value.alias, "_q"), copy=False)
             )
-    else:
+    elif not isinstance(value.this, exp.AggFunc):
         select = select.group_by(value.this, copy=False)
 
     parent_select.join(
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -1537,3 +1537,39 @@ FROM READ_CSV('tests/fixtures/optimizer/tpc-h/nation.csv.gz', 'delimiter', '|') 
         self.assertEqual(4, normalization_distance(gen_expr(2), max_=100))
         self.assertEqual(18, normalization_distance(gen_expr(3), max_=100))
         self.assertEqual(110, normalization_distance(gen_expr(10), max_=100))
+
+    def test_unnest_subqueries_without_agg_fix(self):
+        # This test verifies that in a correlated subquery using an IN predicate,
+        # when the projection is not aggregated, a GROUP BY is added.
+        sql = "SELECT s.t FROM s WHERE 1 IN (SELECT t.a FROM t WHERE t.b > 1)"
+        expression = parse_one(sql)
+        transformed = optimizer.unnest_subqueries.unnest_subqueries(expression)
+
+        # Retrieve the join that was added by the unnesting transformation.
+        joins = transformed.args.get("joins")
+        self.assertIsNotNone(joins, "Expected a join to be added after unnesting")
+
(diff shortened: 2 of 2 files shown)


skadel:
Title:
Avoid Adding GROUP BY on Aggregate Projections in Lateral Subqueries

Description:
This pull request addresses an issue where our SQL AST transformation (in the unnest_subqueries function) adds a GROUP BY clause to subqueries even when their projection already contains an aggregate function. Such behavior causes the generated SQL to include an aggregate (e.g., MAX(a)) within a lateral join, which results in a binder error:
Binder Error: LATERAL join cannot contain aggregates!.

Summary of Changes:

Code Fix:

Modified the transformation logic to check if the subquery projection is already an aggregate function (using isinstance(value.this, exp.AggFunc)).

If the projection is an aggregate, the code now skips adding an extra GROUP BY clause. This prevents queries like GROUP BY MAX("...") that would trigger errors in lateral join contexts.

Unit Tests:

Updated existing tests to maintain coverage for basic unnesting scenarios.

Added a new test (test_unnest_subqueries_with_agg_fix) that asserts no GROUP BY is added when the subquery projection is already an aggregation.

Added another test (test_unnest_subqueries_without_agg_fix) to verify that for subqueries without an aggregation, the fix correctly appends a GROUP BY on the selected column.

Impact:
These changes ensure that the transformation preserves the intended semantics and prevents potential SQL errors in queries using lateral joins with aggregates. The added tests help to prevent regressions by explicitly verifying both scenarios.

tests/test_optimizer.py
```diff
@@ -1537,3 +1537,39 @@ def gen_expr(depth: int) -> exp.Expression:
         self.assertEqual(4, normalization_distance(gen_expr(2), max_=100))
         self.assertEqual(18, normalization_distance(gen_expr(3), max_=100))
         self.assertEqual(110, normalization_distance(gen_expr(10), max_=100))
+
+    def test_unnest_subqueries_without_agg_fix(self):
```
georgesittas, line 1541: Can we add these tests to the `unnest_subqueries.sql` fixture? Is there a reason to put them here and do manual work?

georgesittas (commented):
Hey @skadel, sorry for the delay on reviewing this, it's been a few busy days. Me or someone else will take a closer look into this soon.

cc @barakalon in case you got cycles to check this out as well.

sqlglot/optimizer/unnest_subqueries.py
```diff
@@ -105,7 +105,7 @@ def unnest(select, parent_select, next_alias_name):
                 .from_(select.subquery("_q", copy=False), copy=False)
                 .group_by(exp.column(value.alias, "_q"), copy=False)
             )
-    else:
+    elif not isinstance(value.this, exp.AggFunc):
```
georgesittas, line 108: We need to generalize this as follows:

```suggestion
    elif not value.this.find(exp.AggFunc):
```

Otherwise we'll still optimize queries like the following incorrectly (`value.this` will be an `Add` but there's still going to be an aggregate under that node):

```sql
SELECT s.t FROM s WHERE 1 IN (SELECT MAX(t.a) as t1 FROM t)
```

georgesittas (commented):
I tested on a few engines and it seems like having an aggregate under `GROUP BY` is consistently invalid.

LGTM once the two comments are addressed.

sqlglot/optimizer/unnest_subqueries.py
barakalon: +1 to @georgesittas 's comments, but use `find_in_scope` instead of `Expression.find`

georgesittas:
https://github.com/tobymao/sqlglot/commit/830c9b8bbf906cf5d4fa8028b67dadda73fc58a9

PR #4971, opened 2025-04-14
fix(clickhouse): allow string literal for clickhouse ON CLUSTER clause

lepfhty:
First of all, thank you for an amazing library! 🙏 

In ClickHouse, our team uses the `ON CLUSTER` clause with a string macro that is evaluated by the server, like so: `ON CLUSTER '{cluster}'`.

The problem is that queries using the macro string literal:
```
CREATE TABLE foo ON CLUSTER '{cluster}' ...
```
will be parsed and re-generated with a double-quoted identifier
```
CREATE TABLE foo ON CLUSTER "{cluster}" ...
```

This is incorrect because the double-quoted identifier `{cluster}` is not a valid cluster name. The parser should accept the string literal `'{cluster}'`.

In the ClickHouse source, parsing the ON CLUSTER clause uses a function called `parseIdentifierOrStringLiteral`

ON CLUSTER: https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/ASTQueryWithOnCluster.cpp#L24

parseIdentifierOrStringLiteral: https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/parseIdentifierOrStringLiteral.cpp#L27

The proposed fix is to attempt `_parse_string()` first, then fallback to the existing implementation.

I've copied existing ON CLUSTER test cases and added the equivalent query using string literals.

### DIFF
--- a/tests/dialects/test_clickhouse.py
+++ b/tests/dialects/test_clickhouse.py
@@ -105,8 +105,10 @@ class TestClickhouse(Validator):
         self.validate_identity("SELECT * FROM table LIMIT 1 BY a, b")
         self.validate_identity("SELECT * FROM table LIMIT 2 OFFSET 1 BY a, b")
         self.validate_identity("TRUNCATE TABLE t1 ON CLUSTER test_cluster")
+        self.validate_identity("TRUNCATE TABLE t1 ON CLUSTER '{cluster}'")
         self.validate_identity("TRUNCATE DATABASE db")
         self.validate_identity("TRUNCATE DATABASE db ON CLUSTER test_cluster")
+        self.validate_identity("TRUNCATE DATABASE db ON CLUSTER '{cluster}'")
         self.validate_identity(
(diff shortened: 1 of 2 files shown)


tests/dialects/test_clickhouse.py
```diff
@@ -155,12 +157,21 @@ def test_clickhouse(self):
         self.validate_identity(
             "CREATE TABLE test ON CLUSTER default (id UInt8) ENGINE=AggregatingMergeTree() ORDER BY tuple()"
         )
+        self.validate_identity(
+            "CREATE TABLE test ON CLUSTER '{cluster}' (id UInt8) ENGINE=AggregatingMergeTree() ORDER BY tuple()"
+        )
```
VaggelisD, line 162: I don't usually say this but for this change we could do with less tests, I'd assume 1-2 of these should suffice. 

[Nit] Consider using a "dummy" value like `ON CLUSTER 'foo'` to avoid confusion with python's f-strings

VaggelisD (commented):
Hey @lepfhty, thank you for the kind words & for the contribution! Linking the relevant logic in the Clickhouse parser is also really useful as a ground truth.

PR #4972, opened 2025-04-14
feat(generator): unsupported warning for T-SQL query option

geooo109:
Added a warning for the `QueryOption` expression, which is mainly used for the T-SQL transpilation.

**DOCS**
[T-SQL Option Sytnaxt](https://learn.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=sql-server-ver16)


geooo109:
Let me know, if more dialects must be included in the tests.

PR #4976, opened 2025-04-15
fix!: Add support for Athena's Iceberg partitioning transforms

VaggelisD:
Follow up on https://github.com/tobymao/sqlglot/pull/4962

This PR introduces the expressions `exp.PartitionedByBucket` and `exp.PartitionedByTruncate` which control the generation of Iceberg transforms:

- Athena's "Hive" engine (used for `CREATE TABLE`) demands `BUCKET(n, col)` / `TRUNCATE(n, col)` 
- Athena's "Trino" engine (used for CTAS) demands `BUCKET(col, n)` / `TRUNCATE(col, n)` 

Note: The `PARTITIONED BY` property attempts to parse a schema, so the new expressions are parsed as "constraints" since `_parse_schema()` ends up parsing `_parse_constraint() or _parse_column_def()`.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1110,9 +1110,8 @@ class Parser(metaclass=_Parser):
             else exp.PartitionByTruncate
         )
 
-        this, expression = self._parse_wrapped_csv(
-            lambda: self._parse_primary() or self._parse_column()
-        )
+        args = self._parse_wrapped_csv(lambda: self._parse_primary() or self._parse_column())
+        this, expression = seq_get(args, 0), seq_get(args, 1)
 
         if isinstance(this, exp.Literal):
             # Check for Iceberg partition transforms (bucket / truncate) and ensure their arguments are in the right order
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1099,22 +1099,20 @@ class Parser(metaclass=_Parser):
         "WITH": lambda self: self.expression(
             exp.Properties, expressions=self._parse_wrapped_properties()
         ),
-        "BUCKET": lambda self: self._parse_partitioned_by_bucket_or_truncate(
-            exp.PartitionedByBucket
-        ),
-        "TRUNCATE": lambda self: self._parse_partitioned_by_bucket_or_truncate(
-            exp.PartitionByTruncate
-        ),
+        "BUCKET": lambda self: self._parse_partitioned_by_bucket_or_truncate(),
+        "TRUNCATE": lambda self: self._parse_partitioned_by_bucket_or_truncate(),
     }
 
-    def _parse_partitioned_by_bucket_or_truncate(
-        self, exp_class: t.Type[exp.PartitionedByBucket | exp.PartitionByTruncate]
-    ) -> exp.Expression:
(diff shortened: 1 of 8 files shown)


sqlglot/parser.py
```diff
@@ -1099,8 +1099,35 @@ class Parser(metaclass=_Parser):
         "WITH": lambda self: self.expression(
             exp.Properties, expressions=self._parse_wrapped_properties()
         ),
+        "BUCKET": lambda self: self._parse_partitioned_by_bucket_or_truncate(
+            exp.PartitionedByBucket
+        ),
+        "TRUNCATE": lambda self: self._parse_partitioned_by_bucket_or_truncate(
+            exp.PartitionByTruncate
+        ),
```
georgesittas, line 1107: Perhaps we could also check the uppercase `_prev` text here instead of passing in the type.

sqlglot/parser.py
```diff
@@ -1099,8 +1099,35 @@ class Parser(metaclass=_Parser):
...
     }
 
+    def _parse_partitioned_by_bucket_or_truncate(
+        self, exp_class: t.Type[exp.PartitionedByBucket | exp.PartitionByTruncate]
+    ) -> exp.Expression:
+        self._match(TokenType.L_PAREN)
+        this = self._parse_primary() or self._parse_column()
+        self._match(TokenType.COMMA)
+        expression = self._parse_primary() or self._parse_column()
+        self._match(TokenType.R_PAREN)
```
georgesittas, line 1117: Let's use `_parse_wrapped_csv` with a lambda?

georgesittas (approved):

sqlglot/parser.py
VaggelisD: Awesome 👍

tobymao (approved):

sqlglot/parser.py
```diff
@@ -1099,8 +1099,33 @@ class Parser(metaclass=_Parser):
...
+    def _parse_partitioned_by_bucket_or_truncate(self) -> exp.Expression:
+        klass = (
+            exp.PartitionedByBucket
+            if self._prev.text.upper() == "BUCKET"
+            else exp.PartitionByTruncate
+        )
+
+        this, expression = self._parse_wrapped_csv(
+            lambda: self._parse_primary() or self._parse_column()
+        )
```
georgesittas, line 1115: I'd implement this a little more defensively to avoid "not enough elements to expand" errors:

```python
args = self._parse_wrapped_csv(lambda: self._parse_primary() or self._parse_column())
this, expression = seq_get(args, 0), seq_get(args, 1)
```

georgesittas:
Let's verify that this fixes the linked SQLMesh issue.

VaggelisD:
>  Let's verify that this fixes the linked SQLMesh issue.

If I'm not mistaken this is one part of the fix; The other part would be parsing the `partition_by` model expression into the new nodes, which will be generated into the proper Athena engine by using the appropriate generator.

georgesittas:
I see, makes sense, thanks for clarifying.

erindru (approved):
This looks correct, thanks!

PR #4981, opened 2025-04-15
Fix(optimizer): compare the whole type to determine if a cast can be removed

### LINKED ISSUE #4977: Decimal-Casts are wrongly removed
Optimizing SQL removes casts, resulting in wrong / failing queries. Casting a decimal to a smaller decimal will usually result in an exception, or null depending on the platform. Running the following query on Databricks will write null, if the value is larger than a `decimal(4,2)`.
```sql
insert into tab2 (val1, val2)
select cast(x as decimal(4,2)), try_cast(x as decimal(4,2))
from tab
```

When optimizing the query, all casts are removed. This causes the query to always fail on Databricks (`[CAST_OVERFLOW_IN_TABLE_INSERT] Fail to assign a value of "DECIMAL(10,6)" type to the "DECIMAL(4,2)" type column or variable ... due to an overflow.`)
```sql
SELECT
  `tab`.`x` AS `x`,
  `tab`.`x` AS `x`
FROM `tab` AS `tab`
```

```python
import sqlglot

print(
    optimize(
        sqlglot.parse_one(
            """
insert into tab2 (val1, val2)
select cast(x as decimal(4,2)), try_cast(x as decimal(4,2))
from tab
        """,
            read="databricks",
        ),
        schema={"tab": {"x": "decimal(10,6)"}},
    ).sql(pretty=True, dialect="databricks")
)
```

georgesittas:
Fixes #4977

- Compare the inferred expression's type with that of the `CAST` _structurally_
- Do not simplify `exp.Date` or `exp.TsOrDsToDate` if the inferred expression's type is "complex"

### DIFF
--- a/sqlglot/optimizer/canonicalize.py
+++ b/sqlglot/optimizer/canonicalize.py
@@ -93,15 +93,18 @@ def remove_redundant_casts(expression: exp.Expression) -> exp.Expression:
     if (
         isinstance(expression, exp.Cast)
         and expression.this.type
-        and expression.to.this == expression.this.type.this
+        and expression.to == expression.this.type
     ):
         return expression.this
+
     if (
         isinstance(expression, (exp.Date, exp.TsOrDsToDate))
         and expression.this.type
         and expression.this.type.this == exp.DataType.Type.DATE
+        and not expression.this.type.expressions
     ):
         return expression.this
+
     return expression
 
 
--- a/tests/fixtures/optimizer/canonicalize.sql
+++ b/tests/fixtures/optimizer/canonicalize.sql
@@ -124,6 +124,12 @@ SELECT CAST(CAST(`t`.`some_col` AS DATE) AS DATETIME) < CAST(CAST(`t`.`other_col
 --------------------------------------
 -- Remove redundant casts
 --------------------------------------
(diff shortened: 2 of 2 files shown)


sqlglot/optimizer/canonicalize.py
```diff
@@ -93,15 +93,18 @@ def remove_redundant_casts(expression: exp.Expression) -> exp.Expression:
...
-        and expression.to.this == expression.this.type.this
+        and expression.to == expression.this.type
     ):
         return expression.this
+
     if (
         isinstance(expression, (exp.Date, exp.TsOrDsToDate))
         and expression.this.type
         and expression.this.type.this == exp.DataType.Type.DATE
+        and not expression.this.type.expressions
```
barakalon, line 104: Does this need a test too?

barakalon (approved):

sqlglot/optimizer/canonicalize.py
georgesittas: I'll cook one up 👍

georgesittas: I guess I don't know how a `DATE` could be parameterized

PR #4985, opened 2025-04-17
Refactor!: remove `this` from Coalesce in favor of `expressions`

georgesittas:
This is a refactor I've been meaning to do. The current representation of `Coalesce` is weird– for some reason we store the first argument in `this` and the rest of them in `expressions`, even though it's always a variable-length argument function as far as I can tell.


tobymao:
are there others that do that? this. expressions

georgesittas:
Discussed internally, not gonna go ahead with this now because we do this elsewhere and it's not an important refactor.

PR #4988, opened 2025-04-17
Refactor(optimizer)!: remove recursion from `simplify`

georgesittas:
Saw https://github.com/tobymao/sqlglot/compare/main...toby/simplify yesterday and figured I'd try to take it to the finish line. Please take a look.

### DIFF
--- a/sqlglot/executor/python.py
+++ b/sqlglot/executor/python.py
@@ -9,7 +9,7 @@ from sqlglot.errors import ExecuteError
 from sqlglot.executor.context import Context
 from sqlglot.executor.env import ENV
 from sqlglot.executor.table import RowReader, Table
-from sqlglot.helper import csv_reader, ensure_list, subclasses
+from sqlglot.helper import csv_reader, subclasses
 
 
 class PythonExecutor:
@@ -370,8 +370,8 @@ def _rename(self, e):
             return self.func(e.key, *values)
 
         if isinstance(e, exp.Func) and e.is_var_len_args:
-            *head, tail = values
-            return self.func(e.key, *head, *ensure_list(tail))
+            args = itertools.chain.from_iterable(x if isinstance(x, list) else [x] for x in values)
+            return self.func(e.key, *args)
 
         return self.func(e.key, *values)
     except Exception as ex:
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -215,12 +215,21 @@ def while_changing(expression: Expression, func: t.Callable[[Expression], E]) ->
         for n in reversed(tuple(expression.walk())):
             n._hash = hash(n)
 
-        start = hash(expression)
+        start_hash = hash(expression)
         expression = func(expression)
 
+        # Uncache previous caches so we can recompute them
+        for n in reversed(tuple(expression.walk())):
+            n._hash = None
+            n._hash = hash(n)
+
+        end_hash = hash(expression)
+
+        # ... and reset the hash so we don't risk it becoming out of date if a mutation happens
         for n in expression.walk():
             n._hash = None
-        if start == hash(expression):
+
+        if start_hash == end_hash:
             break
 
     return expression
--- a/sqlglot/helper.py
+++ b/sqlglot/helper.py
@@ -211,25 +211,31 @@ def while_changing(expression: Expression, func: t.Callable[[Expression], E]) ->
     Returns:
         The transformed expression.
     """
+    end_hash: t.Optional[int] = None
+
     while True:
-        for n in reversed(tuple(expression.walk())):
-            n._hash = hash(n)
+        # No need to walk the AST– we've already cached the hashes in the previous iteration
+        if end_hash is None:
+            for n in reversed(tuple(expression.walk())):
+                n._hash = hash(n)
 
         start_hash = hash(expression)
         expression = func(expression)
 
+        expression_nodes = tuple(expression.walk())
+
         # Uncache previous caches so we can recompute them
-        for n in reversed(tuple(expression.walk())):
+        for n in reversed(expression_nodes):
             n._hash = None
             n._hash = hash(n)
 
         end_hash = hash(expression)
 
-        # ... and reset the hash so we don't risk it becoming out of date if a mutation happens
-        for n in expression.walk():
-            n._hash = None
-
         if start_hash == end_hash:
+            # ... and reset the hash so we don't risk it becoming out of date if a mutation happens
+            for n in expression_nodes:
+                n._hash = None
+
             break
 
     return expression
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -111,7 +111,8 @@ def simplify(
             node, parent = stack.pop()
             root = node is expression
 
-            # Resets parent, arg_key, index pointers
+            # Resets parent, arg_key, index pointers– this is needed because some of the
+            # previous transformations mutate the AST, leading to an inconsistent state
             for k, v in tuple(node.args.items()):
                 node.set(k, v)
 
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -61,74 +61,80 @@ def simplify(
 
     dialect = Dialect.get_or_raise(dialect)
 
-    def _simplify(expression, root=True):
-        if (
-            max_depth
-            and isinstance(expression, exp.Connector)
-            and not isinstance(expression.parent, exp.Connector)
-        ):
-            depth = connector_depth(expression)
-            if depth > max_depth:
-                logger.info(
-                    f"Skipping simplification because connector depth {depth} exceeds max {max_depth}"
-                )
-                return expression
+    def _simplify(expression):
+        stack = []
 
-        if expression.meta.get(FINAL):
-            return expression
+        for node in expression.dfs(prune=lambda n: n.meta.get(FINAL)):
+            if node.meta.get(FINAL):
+                continue
 
-        # group by expressions cannot be simplified, for example
-        # select x + 1 + 1 FROM y GROUP BY x + 1 + 1
-        # the projection must exactly match the group by key
-        group = expression.args.get("group")
+            # group by expressions cannot be simplified, for example
+            # select x + 1 + 1 FROM y GROUP BY x + 1 + 1
+            # the projection must exactly match the group by key
+            group = node.args.get("group")
 
-        if group and hasattr(expression, "selects"):
-            groups = set(group.expressions)
-            group.meta[FINAL] = True
+            if group and hasattr(node, "selects"):
+                groups = set(group.expressions)
+                group.meta[FINAL] = True
 
-            for e in expression.selects:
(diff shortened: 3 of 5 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -61,74 +61,79 @@ def simplify(
...
+                node.replace(new_node)
+
+            stack.append((new_node, parent))
+
+        while stack:
+            node, parent = stack.pop()
+            root = node is expression
+
+            # Resets parent, arg_key, index pointers
+            exp.replace_children(node, lambda n: n)
```
georgesittas, line 115: This one can probably be cleaned up. The reason I added it is because `simplify` does quite a lot of mutation, e.g. `uniq_sort` changes the `arg_key`s of some nodes when it moves them around and that messes up the ast.

Without this line, TPC-DS 13 fails because even though we `pop()` the `join["on"]` arg in `remove_where_true`, this is not reflected in the final AST and the test fails since there's still an `ON TRUE` clause there.

I believe the reason this doesn't happen in main is because `replace_children` re-sets the AST pointers like I do here with the identity callback, which is why I decided to mimic it here.

What `simplify` does today: https://github.com/tobymao/sqlglot/blob/e7a1ad2994539e55a238af573ec4e397f291af08/sqlglot/optimizer/simplify.py#L113

Notice how `replace_children` calls `set` at the end of each iteration: https://github.com/tobymao/sqlglot/blob/e7a1ad2994539e55a238af573ec4e397f291af08/sqlglot/expressions.py#L8377

sqlglot/executor/python.py
```diff
@@ -370,8 +370,8 @@ def _rename(self, e):
             return self.func(e.key, *values)
 
         if isinstance(e, exp.Func) and e.is_var_len_args:
-            *head, tail = values
-            return self.func(e.key, *head, *ensure_list(tail))
+            args = itertools.chain.from_iterable(x if isinstance(x, list) else [x] for x in values)
```
tobymao, line 373: is this really better than previous?

sqlglot/executor/python.py
```diff
@@ -370,8 +370,8 @@ def _rename(self, e):
...
+            return self.func(e.key, *args)
```
georgesittas, line 374: This file can be actually reverted, but I think this version is safer than what we had. Context below 👇 

This is actually related to [this](https://github.com/tobymao/sqlglot/pull/4985) PR. Before introducing the `replace_children` call (see other [comment](https://github.com/tobymao/sqlglot/pull/4988/files#r2049610627)), there was a bug where `head` would be `[this, [expressions]]` for `Coalesce` instances found in TPC tests, because the last argument was actually not `expressions`, but `is_nvl`, which was `None`.

The second item in this list being a list itself means that `self.func` ultimately failed because the generator doesn't know how to generate lists.

Again, the reason this doesn't happen in main today is because `replace_children` resets each child argument, and `set` with a `None` value doesn't store the argument in `args`, so `is_nvl=None` was being discarded.

However, I have a theory that the same bug occurs in main if one tries to execute an Oracle query that involves coalescing, because in that case it won't be `None` ([source](https://github.com/tobymao/sqlglot/blob/e7a1ad2994539e55a238af573ec4e397f291af08/sqlglot/dialects/oracle.py#L106)).

Note that we're setting `is_nvl=None` in the default `Coalesce constructor: https://github.com/tobymao/sqlglot/blob/e7a1ad2994539e55a238af573ec4e397f291af08/sqlglot/parser.py#L153-L154

sqlglot/executor/python.py
georgesittas: Let's move the discussion [here](https://github.com/tobymao/sqlglot/pull/4988#discussion_r2049618557), I've provided some context around the change.

sqlglot/executor/python.py
georgesittas: I verified that my theory is correct, this is indeed a bug in main:

```python
execute("WITH t AS (SELECT 'a' AS c1, 'b' AS c2) SELECT NVL(c1, c2) FROM t", dialect="oracle")
```

This produces the following error:

```
ValueError: Expected an Expression. Received <class 'list'>: [Column(
  this=Identifier(this='C2', quoted=True),
  table=Identifier(this='T', quoted=True),
  _type=DataType(this=Type.VARCHAR))]
...
    raise Exception(f"Could not rename {repr(e)}") from ex
Exception: Could not rename Coalesce(
  this=Column(
    this=Identifier(this='C1', quoted=True),
    table=Identifier(this='T', quoted=True),
    _type=DataType(this=Type.VARCHAR)),
  expressions=[
    Column(
      this=Identifier(this='C2', quoted=True),
      table=Identifier(this='T', quoted=True),
      _type=DataType(this=Type.VARCHAR))],
  is_nvl=True,
  _type=DataType(this=Type.VARCHAR))
```

sqlglot/optimizer/simplify.py
tobymao: this is very expensive, can we fix it another way?

sqlglot/executor/python.py
tobymao: i see ok

tobymao: we should add a test right?

sqlglot/optimizer/simplify.py
georgesittas: As a first step I got rid of the `replace_children` call in favor of the loop we're ultimately interested in:

```python
            # Resets parent, arg_key, index pointers
            for k, v in tuple(node.args.items()):
                node.set(k, v)
```

Will look again shortly to see what else is possible.

sqlglot/optimizer/simplify.py
```diff
@@ -61,74 +61,80 @@ def simplify(
...
+            new_node = simplify_concat(new_node)
+            new_node = simplify_conditionals(new_node)
+
+            if constant_propagation:
+                new_node = propagate_constants(new_node, root)
+
+            if new_node is not node:
+                node.replace(new_node)
+
+            stack.append((new_node, parent))
```
VaggelisD, line 108: I forget our consensus here, do we have to store & propagate the "old" `parent`? 

If `new_node` is properly replaced i.e `new_node.parent == node.parent`, what'd be the difference in accessing `new_node.parent` directly?

sqlglot/optimizer/simplify.py
```diff
@@ -61,74 +61,80 @@ def simplify(
...
+
+        while stack:
+            node, parent = stack.pop()
+            root = node is expression
+
+            # Resets parent, arg_key, index pointers
+            for k, v in tuple(node.args.items()):
+                node.set(k, v)
```
VaggelisD, line 116: I think I'd include more context here, ideally this loop shouldn't exist if the mutation bug wasn't out there right?

sqlglot/optimizer/simplify.py
```diff
@@ -61,74 +61,80 @@ def simplify(
...
+
+            # Post-order transformations
+            new_node = simplify_not(node)
+            new_node = flatten(new_node)
+            new_node = simplify_connectors(new_node, root)
+            new_node = remove_complements(new_node, root)
+            new_node = simplify_coalesce(new_node, dialect)
+
+            new_node.parent = parent
```
VaggelisD, line 125: Iirc this also caused by a mutation bug / unwanted side effect, did you end up investigating why is it needed? 

Both of these "resets" are worrisome considering `simplify` is used on SQLMesh

sqlglot/optimizer/simplify.py
georgesittas: I'll take another look, not sure.

sqlglot/optimizer/simplify.py
georgesittas: Just to be clear: I don't think the mutation is necessarily a bug. Copying everywhere may end up being too costly.

sqlglot/optimizer/simplify.py
georgesittas: This assignment matches what originally happened in `simplify`, so I'd rather not risk removing it.

I think it's necessary, though. For example, if any post-order transformation is added that constructs a new expression and we do `new_node = transform(new_node)`, then the parent link will be lost.

sqlglot/optimizer/simplify.py
georgesittas: I think it's safe to do this and it doesn't cost much.

It seems that doing `new_node.parent = node.parent` in the post-order transformation loop also works now, but it feels risky. If for whatever reason a pre-transformation calls `replace` on the visited `node` before we replace it with `new_node`, then the parent link will be lost.

Keeping track of this nuance as time passes is non-trivial, so I think it's worth being explicit that we want to preserve the "old" parent.

VaggelisD (approved):

georgesittas:
Just a heads up: I don't think this PR will enable us to handle large ASTs without making the `Expression` hashing iterative, because `simplify` [leverages hashing to detect if we reached a fixed point](https://github.com/tobymao/sqlglot/blob/da9ec61e8edd5049e246390e1b638cf14d50fa2d/sqlglot/helper.py#L223-L224), which is recursive and hence a bottleneck.

I've created two branches [1, 2], trying different approaches to achieve this, but both incur a performance hit compared to main and so still need to figure out how to optimize them further to at least reach parity.

[1]: https://github.com/tobymao/sqlglot/tree/jo/make_hashing_iterative
[2]: https://github.com/tobymao/sqlglot/tree/jo/make_hashing_iterative_alt

sqlglot/helper.py
```diff
@@ -215,12 +215,21 @@ def while_changing(expression: Expression, func: t.Callable[[Expression], E]) ->
...
+        # Uncache previous caches so we can recompute them
+        for n in reversed(tuple(expression.walk())):
+            n._hash = None
+            n._hash = hash(n)
+
+        end_hash = hash(expression)
+
+        # ... and reset the hash so we don't risk it becoming out of date if a mutation happens
         for n in expression.walk():
             n._hash = None
```
georgesittas, line 230: I don't expect this to be that different from what we used to do. Previously:

1. L215-L216: compute and cache each hash in the AST
2. L221-L222: walk the AST to evict the cached `_hash` so it can be recomputed
3. L223: walk the AST recursively due to `hash`

We now do the same number of walks, with the only difference being that we have to materialize the `walk` generator into a tuple so we can reverse it in L222-L224.

This is the benchmark I tested with:

```python
from timeit import timeit
from sqlglot import parse_one
from sqlglot.optimizer.simplify import simplify

def generate_bushy_union_sql(depth):
    if depth == 0:
        return "SELECT 1 AS value"
    left = generate_bushy_union_sql(depth - 1)
    right = generate_bushy_union_sql(depth - 1)
    return f"(\n{left}\n)\nUNION ALL\n(\n{right}\n)"

sql = generate_bushy_union_sql(10)
ast = parse_one(sql)

# main: 0.5100423749827314, branch: 0.5304469999973662
timeit(lambda: simplify(ast), number=10)

# main: 4.974802584009012, branch: 5.00075383298099
# timeit(lambda: simplify(ast), number=100)
```

tobymao (approved):

PR #4990, opened 2025-04-18
Fix(hive)!: stop generating redundant TO_DATE calls

georgesittas:


sqlglot/dialects/hive.py
```diff
@@ -167,7 +174,9 @@ def _to_date_sql(self: Hive.Generator, expression: exp.TsOrDsToDate) -> str:
     if time_format and time_format not in (Hive.TIME_FORMAT, Hive.DATE_FORMAT):
         return self.func("TO_DATE", expression.this, time_format)
 
-    if isinstance(expression.this, exp.TsOrDsToDate):
+    if isinstance(expression.this, exp.TsOrDsToDate) or isinstance(
```
tobymao, line 177: add tsordstodate to ts_or_ds_expressions maybe?

tobymao (approved):

sqlglot/dialects/hive.py
georgesittas: I intentionally left it this way because if I add it to that tuple then `TO_DATE(x)` will stop roundtripping.

PR #4992, opened 2025-04-20
feat(Snowflake): enhance support for CASE expressions in BEGIN..END blocks

whummer:
## Motivation

Snowflake supports two types of `CASE` statements:
* `CASE` for conditional expressions: https://docs.snowflake.com/en/sql-reference/functions/case
* `CASE` for conditional statements in scripting blocks: https://docs.snowflake.com/en/sql-reference/snowflake-scripting/case

Currently we are not able to parse the latter, i.e., `CASE` statements in scripting blocks that are typically encapsulated within `BEGIN ... END` blocks.

For example, the following query is valid in Snowflake, but raises a parsing error against sqlglot:
```
BEGIN 
  CASE ($myvar) 
    WHEN 'foo' THEN SELECT MYFUNC(1); 
    WHEN 'bar' THEN RETURN 2; 
    ELSE RETURN 3; 
  END CASE; 
END
```

One of the challenges when parsing such expressions is the use of semicolons above - the current logic in `Parser._parse(..)` terminates each expression and creates new chunks whenever a semicolons is found (see [here](https://github.com/tobymao/sqlglot/blob/510984f2ddc6ff13b8a8030f698aed9ad0e6f46b/sqlglot/parser.py#L1589)), however, for `CASE` expressions we need to preserve the semicolons, to be able to use `;` as a delimiter for each `WHEN`/`ELSE` clause.

## Changes

* add `BeginEndBlock` model class to represent `BEGIN ... END` code blocks
* add `CaseScriptingBlock` model class to represent `CASE ... END CASE` scripting blocks
* ...

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -834,6 +834,7 @@ class Parser(metaclass=_Parser):
         TokenType.PIVOT: lambda self: self._parse_simplified_pivot(),
         TokenType.PRAGMA: lambda self: self.expression(exp.Pragma, this=self._parse_expression()),
         TokenType.REFRESH: lambda self: self._parse_refresh(),
+        TokenType.RETURN: lambda self: self._parse_return(),
         TokenType.ROLLBACK: lambda self: self._parse_commit_or_rollback(),
         TokenType.SET: lambda self: self._parse_set(),
         TokenType.TRUNCATE: lambda self: self._parse_truncate_table(),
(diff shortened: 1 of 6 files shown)


sqlglot/parser.py
```diff
@@ -1585,8 +1586,13 @@ def _parse(
         total = len(raw_tokens)
         chunks: t.List[t.List[Token]] = [[]]
 
+        block_nesting_level = 0
         for i, token in enumerate(raw_tokens):
-            if token.token_type == TokenType.SEMICOLON:
+            if self._is_nested_block_start(token, i, raw_tokens):
```
whummer, line 1591: Note: Curious to get your thoughts on this. This is an attempt to determine whether we are executing in the context of a nested code block (i.e., `BEGIN ... END`) and to preserve semicolon tokens within the current chunk. 

The idea of `_is_nested_block_start(..)` is to let dialects decide whether we are executing within a code block where semicolons should be preserved. By default the method returns `False`, and the `snowflake` dialect parser overwrites this method.

There may of course be better / more elegant ways to achieve that :) - happy for any suggestions!

georgesittas:
Hey, @whummer 👋 

Appreciate the PR, but after discussing internally, we decided that we prefer not to introduce specialized logic for handling Snowflake's statement blocks, etc. This would require a larger refactor to design correctly, and for the time being supporting imperative SQL constructs is out of SQLGlot's scope.

If your application deals with these sort of statements, I don't think it would be a big lift to override the `Parser` components needed to support your use case, at least the syntax you mention here.

whummer:
Hi @georgesittas , thanks for the feedback. Bit of a bummer, but we should be able to override this on the application level for now. 👍 As usual, thanks for the quick feedback and consideration!

PR #5003, opened 2025-04-22
feat(clickhouse): Support parsing CTAS with alias

dorranh:
This PR adds support for ClickHouse's special variant of `CREATE TABLE AS` which allows you to specify another table to create your table from as opposed to using a `SELECT` statement. 

I don't know of any other dialects which support this, but I still added the implementation in `sqlglot.parser` since I did not see an easy way to customize just this part of the parser in the `sqlglot.dialects.clickhouse`.

https://clickhouse.com/docs/sql-reference/statements/create/table#with-a-schema-similar-to-other-table

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1999,7 +1999,7 @@ class Parser(metaclass=_Parser):
             # exp.Properties.Location.POST_SCHEMA and POST_WITH
             extend_props(self._parse_properties())
 
-            self._match(TokenType.ALIAS)
+            has_alias = self._match(TokenType.ALIAS)
             if not self._match_set(self.DDL_SELECT_TOKENS, advance=False):
                 # exp.Properties.Location.POST_ALIAS
                 extend_props(self._parse_properties())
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -2009,6 +2009,11 @@ def extend_props(temp_props: t.Optional[exp.Properties]) -> None:
                 extend_props(self._parse_properties())
             else:
                 expression = self._parse_ddl_select()
+                # Some dialects also support using a table as an alias instead of a SELECT.
+                # Here we fallback to this as an alternative.
+                if not expression and has_alias:
+                    if self._match_set({TokenType.VAR}, advance=False):
```
georgesittas, line 2015:
```suggestion
                    if self._match(TokenType.VAR, advance=False):
```

tests/dialects/test_clickhouse.py
```diff
@@ -766,6 +766,19 @@ def test_aggregate_function_column_with_any_keyword(self):
...
+    def test_create_table_as_alias(self):
+        ctas_alias = "CREATE TABLE my_db.my_table AS another_db.another_table"
+
+        expected = exp.Create(
+            this=exp.Table(this=exp.Identifier(this="my_table"), db=exp.Identifier(this="my_db")),
+            kind="TABLE",
+            expression=exp.Table(
+                this=exp.Identifier(this="another_table"), db=exp.Identifier(this="another_db")
+            ),
+        )
```
georgesittas, line 778:
```suggestion
        expected = exp.Create(
            this=exp.to_table("my_db.my_table"),
            kind="TABLE",
            expression=exp.to_table("another_db.another_table"),
        )
```

georgesittas (approved):
Thanks! A couple minor suggestions

dorranh:
Thanks for the quick review and updates to the PR, @georgesittas!

georgesittas:
Yep! Further refactored here: https://github.com/tobymao/sqlglot/commit/1d4d906abc60d29b6606bc8eee50c92cef21d3fd

PR #5004, opened 2025-04-22
feat(clickhouse): Support ClickHouse Nothing type

dorranh:
Adds support for ClickHouse's `Nothing` type which is used for types that never have a value. I opted to map this to the existing NULL token type rather than add a dedicated token type, though let me know if this causes any issues (e.g. with the Rust-based tokenizer).

https://clickhouse.com/docs/sql-reference/data-types/special-data-types/nothing


sqlglot/dialects/clickhouse.py
```diff
@@ -260,6 +260,7 @@ class Tokenizer(tokens.Tokenizer):
             "LOWCARDINALITY": TokenType.LOWCARDINALITY,
             "MAP": TokenType.MAP,
             "NESTED": TokenType.NESTED,
+            "NOTHING": TokenType.NULL,
```
georgesittas, line 263: Let's map this to a (new) `TokenType.NOTHING` token instead of overloading `NULL`.

georgesittas: No need for that, tokens are dynamically loaded in the Rust tokenizer, so you should only modify Python code.

dorranh: Ah that is quite nice. Thanks

dorranh: Updated here: cfc9329170c6b3a51376af951f1c8e1b5c9c2e88

georgesittas (approved):
Perfect, thank you!

PR #5008, opened 2025-04-23
Add numeric parameter support for Snowflake dialect

hovaesco:
Add support for numeric parameters in Snowflake dialect, see the relevant Snowflake documentation https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-example#using-qmark-or-numeric-binding

Example from Snowflake docs:
```py
import snowflake.connector

snowflake.connector.paramstyle='numeric'

con = snowflake.connector.connect(...)

con.cursor().execute(
    "INSERT INTO testtable(col1, col2) "
    "VALUES(:1, :2)", (
        789,
        'test string3'
    ))
```

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -514,6 +514,15 @@ class Snowflake(Dialect):
             TokenType.SHOW: lambda self: self._parse_show(),
         }
 
+        PLACEHOLDER_PARSERS = {
+            **parser.Parser.PLACEHOLDER_PARSERS,
+            TokenType.COLON: lambda self: (
+                self.expression(exp.Placeholder, this=int(self._prev.text))
+                if self._match_set(self.ID_VAR_TOKENS) or self._match(TokenType.NUMBER)
+                else None
+            ),
+        }
+
         PROPERTY_PARSERS = {
             **parser.Parser.PROPERTY_PARSERS,
             "CREDENTIALS": lambda self: self._parse_credentials_property(),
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -514,6 +514,15 @@ class Parser(parser.Parser):
             TokenType.SHOW: lambda self: self._parse_show(),
         }
 
+        PLACEHOLDER_PARSERS = {
+            **parser.Parser.PLACEHOLDER_PARSERS,
+            TokenType.COLON: lambda self: (
+                self.expression(exp.Placeholder, this=int(self._prev.text))
```
VaggelisD, line 520: The CI issue is because we're attempting to cast bindings such as `:foo` into `int`s as well, can we simply do `this=self._prev.text` to store both variants?

VaggelisD (commented):
Hey @hovaesco, thanks for your contribution!

You can also run `make style` and `make unit` locally to test your changes :)

tests/dialects/test_snowflake.py
```diff
@@ -2572,3 +2572,10 @@ def test_rely_options(self):
             self.validate_identity(
                 f"CREATE TABLE t (col1 INT, col2 INT, col3 INT, PRIMARY KEY (col1) {option}, UNIQUE (col1, col2) {option}, FOREIGN KEY (col3) REFERENCES other_t (id) {option})"
             )
+
+    def test_parameter(self):
+        expr_placeholder = parse_one("SELECT :1", dialect="snowflake").find(exp.Placeholder)
+        self.assertEqual(expr_placeholder, exp.Placeholder(this="1"))
+        self.validate_identity("SELECT :1")
```
VaggelisD, line 2579: [Nit] `validate_identity` actually returns the parsed AST, so you can simplify these down to:

```Python3
expr = self.validate_identity("SELECT :1")
self.assertEqual(expr.find(exp.Placeholder), exp.Placeholder(this="1")
```

VaggelisD (approved):
Thanks!

For context, we had this exact logic before but changed it [here](https://github.com/tobymao/sqlglot/pull/3761) to accommodate DuckDB's splice notation, it was an oversight to not revisit number matching.

PR #5009, opened 2025-04-24
chore: Refactor colon placeholder parsing 

VaggelisD:
Follow up of https://github.com/tobymao/sqlglot/pull/5008

Since the difference between `PLACEHOLDER_PARSERS` is only the following token set, we can spare overriding the entire dict and instead create & override a new set for `TokenType.COLON`.

cc: @hovaesco


hovaesco (approved):
Thanks!

georgesittas (approved):

PR #5014, opened 2025-04-25
Fix DuckDB IN literal list

aneeshdurg:
In DuckDB `a IN (['a'])` will throw an error as it interprets `(['a'])` as casting `'a'` to a LIST. Instead, for IN, we should generate `a IN ['a']`.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import typing as t
 
 from sqlglot import exp, generator, parser, tokens, transforms
-from sqlglot.expressions import DATA_TYPE
+from sqlglot.expressions import DATA_TYPE, Array, Literal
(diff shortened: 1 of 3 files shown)


georgesittas:
Hey @aneeshdurg, based on your test it seems like you're using the `isin` builder incorrectly. For example, this works fine:

```python
>>> import sqlglot
>>> sqlglot.exp.select(sqlglot.exp.column("a").isin("a")).sql("duckdb")
"SELECT a IN ('a')"
```

Note that `isin` expects a variable number of arguments, not a list: https://github.com/tobymao/sqlglot/blob/06db2350a1c625cc30212912a73267127fc4f3ea/sqlglot/expressions.py#L884

I'm closing this as not planned. Happy to continue discussing, but we'll need more information to proceed.

aneeshdurg:
@georgesittas but what about the case where the rhs of the IN operator is a single list literal?

Issue #5016, opened 2025-04-27
Index after dot not parsed correctly in JSONPath.


omerhadari:
Json Array index access using `.<idx>` doesn't work, though supported by engines.

The following prints a warning and jsonpath isn't parsed correctly:

```Python
from sqlglot import parse_one
parse_one("select json_extract(col['field'], '$.1') from table", dialect="athena")
```
result:
```Python
Select(
  expressions=[
    JSONExtract(
      this=Bracket(
        this=Column(
          this=Identifier(this=col, quoted=False, _type=DataType(this=Type.UNKNOWN)),
          _type=DataType(this=Type.UNKNOWN)),
        expressions=[
          Literal(this='field', is_string=True, _type=DataType(this=Type.VARCHAR))]),
      expression=Literal(this='$.1', is_string=True))],
  from=From(
    this=Table(
      this=Identifier(this=table, quoted=False))))
```

The following is parsed as expected:

```Python
from sqlglot import parse_one
parse_one("select json_extract(col['field'], '$[1]') from table", dialect="athena")
```
result:
```Python
Select(
  expressions=[
    JSONExtract(
      this=Bracket(
        this=Column(
          this=Identifier(this=col, quoted=False, _type=DataType(this=Type.UNKNOWN)),
          _type=DataType(this=Type.UNKNOWN)),
        expressions=[
          Literal(this='field', is_string=True, _type=DataType(this=Type.VARCHAR))]),
      expression=JSONPath(
        expressions=[
          JSONPathRoot(),
          JSONPathSubscript(this=1)]))],
  from=From(
    this=Table(
      this=Identifier(this=table, quoted=False))))
```
Though both are supported in Athena (and Trino).

omerhadari:
take, I have a PR I will open in a minute

georgesittas:
See https://github.com/tobymao/sqlglot/pull/5017#discussion_r2062708393

PR #5017, opened 2025-04-27
fix(jsonpath): support number after dot for index

omerhadari:
solves #5016

### DIFF
--- a/sqlglot/jsonpath.py
+++ b/sqlglot/jsonpath.py
@@ -66,9 +66,7 @@ def parse(path: str, dialect: DialectType = None) -> exp.JSONPath:
         pass
 
     @t.overload
-    def _match(
-        token_type: TokenType, raise_unmatched: Lit[False] = False
-    ) -> t.Optional[Token]:
+    def _match(token_type: TokenType, raise_unmatched: Lit[False] = False) -> t.Optional[Token]:
         pass
 
     def _match(token_type, raise_unmatched=False):
@@ -197,9 +195,7 @@ def parse(path: str, dialect: DialectType = None) -> exp.JSONPath:
             elif value:
                 expressions.append(exp.JSONPathKey(this=value))
             else:
-                raise ParseError(
-                    _error("Expected key name, numeric index or * after DOT")
-                )
+                raise ParseError(_error("Expected key name, numeric index or * after DOT"))
         elif _match(TokenType.L_BRACKET):
             expressions.append(_parse_bracket())
         elif _match(TokenType.VAR):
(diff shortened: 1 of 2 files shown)


sqlglot/jsonpath.py
```diff
@@ -178,20 +178,24 @@ def _parse_var_text() -> str:
...
+                value = int(_prev().text)
             elif _match(TokenType.STAR):
                 value = exp.JSONPathWildcard()
             else:
                 value = None
 
             if recursive:
                 expressions.append(exp.JSONPathRecursive(this=value))
+            elif isinstance(value, int):
+                expressions.append(exp.JSONPathSubscript(this=value))
```
georgesittas, line 194: I don't think we can unambiguously parse `$.1` as a subscript, because the same syntax can be used to specify a dictionary key:

```
trino> select json_extract('[1, 2, 3]', '$.1');
 _col0 
-------
 2     
(1 row)
trino> select json_extract('{"1": "foo"}', '$.1');
 _col0 
-------
 "foo" 
(1 row)
```

Biasing either `JSONPathSubscript` or `JSONPathKey` may lead to silently mis-transpiling a query. For example, parsing it as a subscript results in the following behavior:

```python
>>> import sqlglot
>>> print(sqlglot.parse_one("""select json_extract('{"1": "foo"}', '$.1');""", "trino").sql("duckdb"))
SELECT '{"1": "foo"}' -> '$[1]'
```

The Trino query returns `"foo"`, whereas the DuckDB one returns `NULL`, which is problematic. The warning emitted today draws attention to the fact that SQLGlot couldn't understand the JSON path syntax, and hence it's easier to see that the transpiled version may be inaccurate.

For this reason, I think I'll go ahead and close this PR and relevant issue as "won't fix", but happy to continue discussing possible solutions.

omerhadari: Hi, thanks for the response! I'd be happy to continue discussion and see if I could maybe work on a fix with a better approach.

Would a PR which introduces the behaviour I added but only for Trino be considered?

omerhadari: You're right of course.

I understand the warning, and my approach in this PR is indeed not good. I'm just worried about emitting it when parsing.

How would you feel about a solution which introduces another `JSONPathPart` class specifically for dot access, and treating it differently during generation?

Instead of a emitting a warning during parsing for valid queries, it could either emit a warning or throw an exception when trying to generate sql for a dialect that doesn't support an ambiguously typed dot-access (similar to how for example union isn't supported by the duckdb generator).

PR #5019, opened 2025-04-28
feat(snowflake): add support for GET statements

eruditmorina:
## Motivation

Hi folks, following up on https://github.com/tobymao/sqlglot/issues/4813 in order to add support for `GET` statements in snowflake dialect similarly to the way `PUT` statements are being handled.

## Changes

- Add a `GET` expression and TokenType
- Extend Snowflake Parser with `_parse_get` which parses `GET` statements to `exp.Get`, except for the case of unquoted file path (treated as `exp.Command`)
- Tests to cover the functionality of get parsing

Thank you for the support 🙌

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -859,21 +859,17 @@ class Snowflake(Dialect):
             )
 
         def _parse_get(self) -> exp.Get | exp.Command:
-            base = self._prev
-
-            if self._curr.token_type not in (TokenType.STRING, TokenType.PARAMETER):
-                return self._parse_as_command(base)
-
-            source = self._parse_location_path()
+            start = self._prev
+            target = self._parse_location_path()
 
             # Parse as command if unquoted file path
             if self._curr.token_type == TokenType.URI_START:
-                return self._parse_as_command(base)
+                return self._parse_as_command(start)
 
             return self.expression(
                 exp.Get,
-                source=source,
                 this=self._parse_string(),
+                target=target,
                 properties=self._parse_properties(),
             )
 
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -512,6 +512,7 @@ class Snowflake(Dialect):
 
         STATEMENT_PARSERS = {
             **parser.Parser.STATEMENT_PARSERS,
+            TokenType.GET: lambda self: self._parse_get(),
             TokenType.PUT: lambda self: self._parse_put(),
             TokenType.SHOW: lambda self: self._parse_show(),
         }
@@ -857,6 +858,25 @@ class Snowflake(Dialect):
                 properties=self._parse_properties(),
             )
 
+        def _parse_get(self) -> exp.Get | exp.Command:
+            base = self._prev
+
+            if self._curr.token_type not in (TokenType.STRING, TokenType.PARAMETER):
+                return self._parse_as_command(base)
+
+            source = self._parse_location_path()
+
+            # Parse as command if unquoted file path
+            if self._curr.token_type == TokenType.URI_START:
+                return self._parse_as_command(base)
+
+            return self.expression(
+                exp.Get,
+                source=source,
+                this=self._parse_string(),
+                properties=self._parse_properties(),
+            )
+
         def _parse_location_property(self) -> exp.LocationProperty:
             self._match(TokenType.EQ)
             return self.expression(exp.LocationProperty, this=self._parse_location_path())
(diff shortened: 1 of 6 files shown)


sqlglot/expressions.py
```diff
@@ -3307,6 +3307,11 @@ class Put(Expression):
     arg_types = {"this": True, "target": True, "properties": False}
 
 
+# https://docs.snowflake.com/en/sql-reference/sql/get
+class Get(Expression):
+    arg_types = {"source": True, "this": True, "properties": False}
```
VaggelisD, line 3312: I think it'd unify the args between `GET` and `PUT`, e.g both can be `{"this": True, "expression": True, ...}` or `{"this": True, "target": True, ...}`.

With this in mind, we can further refactor both of their parsing & generation.

tests/dialects/test_snowflake.py
```diff
@@ -2465,6 +2465,33 @@ def test_put_to_stage(self):
             check_command_warning=True,
         )
 
+    def test_get_from_stage(self):
```
VaggelisD, line 2468: Let's also add a test that we can still parse `get` as an identifier [similar to PUT](https://github.com/tobymao/sqlglot/blob/main/tests/dialects/test_snowflake.py#L109), this is a good practice when adding new tokens

sqlglot/dialects/snowflake.py
```diff
@@ -857,6 +858,25 @@ def _parse_put(self) -> exp.Put | exp.Command:
                 properties=self._parse_properties(),
             )
 
+        def _parse_get(self) -> exp.Get | exp.Command:
+            base = self._prev
+
+            if self._curr.token_type not in (TokenType.STRING, TokenType.PARAMETER):
+                return self._parse_as_command(base)
```
VaggelisD, line 865: Does this branch guard against not parsing a proper location path? 

If there are edge cases that we don't parse correctly we should push this logic into the parsing function.

VaggelisD (commented):
Hey @eruditmorina, thank you for the contribution!

Left a few comments from my first pass:

sqlglot/dialects/snowflake.py
eruditmorina: It turns out the check was unnecessary in this case, removed it!

sqlglot/expressions.py
eruditmorina: Thanks for the quick review @VaggelisD! Unified the args and tried to refactor as much as I could, especially the generator logic. Kept the parsing logic separate to prioritize readability since escaping the file path is at a different token for each case and would require more manipulation. But happy to get more insights and guidance and iterate on refactoring further.

sqlglot/generator.py
```diff
@@ -4904,8 +4904,18 @@ def show_sql(self, expression: exp.Show) -> str:
         return ""
 
     def put_sql(self, expression: exp.Put) -> str:
+        return self._get_put_sql(expression)
+
+    def get_sql(self, expression: exp.Get) -> str:
+        return self._get_put_sql(expression)
```
VaggelisD, line 4910: We can now inline these in the `TRANSFORMS` dict:

```Python3
# generator.py

TRANSFORMS = {
  ...
  exp.Get: lambda self, e: self._get_put_sql(e),
  exp.Put: lambda self, e: self._get_put_sql(e),
  ...
}
```

sqlglot/generator.py
```diff
@@ -4904,8 +4904,18 @@ def show_sql(self, expression: exp.Show) -> str:
...
+
+    def _get_put_sql(self, expression: exp.Put | exp.Get) -> str:
```
VaggelisD, line 4912: Can we leave a comment here explaining these are Snowflake's GET/PUT statements, and maybe their simplified syntax such as:

```Python3
def _get_put_sql(self, expression: exp.Put | exp.Get) -> str:
  # Snowflake GET/PUT statements:
  #   PUT <file> <internalStage> <properties>
  #   GET <internalStage> <file> <properties>
```

So it's easy to spot that they're almost identical except the order twist

VaggelisD (commented):
Just a couple more comments to consider from me and I think this is g2g

georgesittas (approved):
Thanks for your contribution @eruditmorina!

VaggelisD (approved):
Awesome work, thank you for the contribution & quick iterations!

seeingskye:
I started getting errors from sqlglot improperly parsing get tokens used in this context: https://docs.snowflake.com/en/sql-reference/functions/get
Could this be related to this latest change?

tobymao:
https://github.com/tobymao/sqlglot/commit/8f77b301a267eadb4c4792201e112159db554d1c

PR #5022, opened 2025-04-29
Feat: Include token refereces in the meta of identifier expressions

izeigerman:


georgesittas:
Quick heads up that there are more places where `Identifier`s are instantiated, e.g.:

https://github.com/tobymao/sqlglot/blob/f5358d8a3e2743b5ac0d540f10502d333ad4e082/sqlglot/dialects/bigquery.py#L648

sqlglot/parser.py
```diff
@@ -8244,3 +8244,11 @@ def _parse_max_min_by(self, expr_type: t.Type[exp.AggFunc]) -> exp.AggFunc:
         return self.expression(
             expr_type, this=seq_get(args, 0), expression=seq_get(args, 1), count=seq_get(args, 2)
         )
+
+    def _identifier_expression(
+        self, token: t.Optional[Token] = None, **kwargs: t.Any
+    ) -> exp.Identifier:
+        token = token or self._prev
+        expression = self.expression(exp.Identifier, this=token.text, **kwargs)
+        expression.meta["token"] = token
```
georgesittas, line 8253: Should we store the attributes we care about instead of the whole token? Also, what about making these `Expression` attributes that default to `None` instead of putting them in `meta`? Don't feel strongly about this, just feel like we've been doing this for other attributes, like `type`.

cc @tobymao

georgesittas: Discussed internally:

- Remove raw token object, store its attributes directly instead (to avoid issues with Rust tokens etc)
- Go with the `meta` approach.

sqlglot/parser.py
```diff
@@ -8244,3 +8247,11 @@ def _parse_max_min_by(self, expr_type: t.Type[exp.AggFunc]) -> exp.AggFunc:
         return self.expression(
             expr_type, this=seq_get(args, 0), expression=seq_get(args, 1), count=seq_get(args, 2)
         )
+
+    def _identifier_expression(
+        self, token: t.Optional[Token] = None, **kwargs: t.Any
+    ) -> exp.Identifier:
+        token = token or self._prev
+        expression = self.expression(exp.Identifier, this=token.text, **kwargs)
+        expression.update_positions(token)
```
georgesittas, line 8256: If an identifier is quoted, we probably want to include the quotes so that a consumer of the line/col information can re-parse that identifier and respect its case-sensitivity.

I'd suggest adding some tests that include quotes as well.

sqlglot/dialects/bigquery.py
```diff
@@ -645,14 +645,18 @@ def _parse_table_part(self, schema: bool = False) -> t.Optional[exp.Expression]:
 
                     table_name += self._find_sql(start, self._prev)
 
+                previous: exp.Expression = this
                 this = exp.Identifier(this=table_name, quoted=this.args.get("quoted"))
+                this.update_positions(previous)
```
georgesittas, line 650: I think you can also do something like `exp.Identifier(...).update_positions(this)` if you want to.

georgesittas (approved):
Looks good, a few final comments to consider.

I'd probably do a quick keyword search for `exp.Identifier`, `exp.to_identifier` and `exp.parse_identifier` to double-check if there are more places where we construct identifiers like that at parse time.

sqlglot/parser.py
izeigerman: Added some tests for quoted identifiers

tobymao:
i'm afraid this token change will affect performance? can we run benchmarks?

sqlglot/dialects/bigquery.py
```diff
@@ -717,7 +731,13 @@ def _parse_table_parts(
                 )
 
                 info_schema_view = f"{table_parts[-2].name}.{table_parts[-1].name}"
-                table.set("this", exp.Identifier(this=info_schema_view, quoted=True))
+                new_this = exp.Identifier(this=info_schema_view, quoted=True).update_positions(
+                    line=table_parts[-2].meta.get("line"),
+                    col=table_parts[-1].meta.get("col"),
```
tobymao, line 736: do we need start and end as well, or just line and col?

izeigerman: I think we need both

sqlglot/expressions.py
```diff
@@ -846,6 +846,33 @@ def not_(self, copy: bool = True):
...
+        """
+        Update this expression with positions from a token or other expression.
+
+        Args:
+            other: a token or expression to update this expression with.
+
+        Returns:
+            The updated expression.
+        """
+        position_keys = ("line", "col", "start", "end")
```
tobymao, line 861: can you make this a const?

sqlglot/expressions.py
```diff
@@ -846,6 +846,33 @@ def not_(self, copy: bool = True):
...
+        elif other is not None:
+            self.meta.update(
+                {
+                    "line": other.line,
+                    "col": other.col,
+                    "start": other.start,
+                    "end": other.end,
+                }
+            )
+        self.meta.update({k: v for k, v in kwargs.items() if k in position_keys})
```
tobymao, line 873: should this be an else statement? why will you do both an expression and also kwargs?

sqlglot/expressions.py
izeigerman: fixed this

tobymao (approved):

PR #5029, opened 2025-05-01
Python executor: add trycast

casassg:


sqlglot/executor/python.py
```diff
@@ -432,6 +432,7 @@ class Generator(generator.Generator):
             exp.Between: _rename,
             exp.Boolean: lambda self, e: "True" if e.this else "False",
             exp.Cast: lambda self, e: f"CAST({self.sql(e.this)}, exp.DataType.Type.{e.args['to']})",
+            exp.TryCast: lambda self, e: f"TRY_CAST({self.sql(e.this)}, exp.DataType.Type.{e.args['to']})" if e.args.get("safe") else f"CAST({self.sql(e.this)}, exp.DataType.Type.{e.args['to']})",
```
georgesittas, line 435: I'd not deal with `safe` at all and just implement the "try cast or fallback" semantics.

sqlglot/executor/env.py
```diff
@@ -131,6 +131,13 @@ def cast(this, to):
     raise NotImplementedError(f"Casting {this} to '{to}' not implemented.")
 
 
+@null_if_any
+def try_cast(this, to):
+    try:
+        return cast(this, to)
+    except Exception:
+        return this
```
georgesittas, line 139: Shouldn't this return `None` (i.e. `NULL`)?

```
spark-sql (default)> select try_cast('foo' as int);
NULL
```

georgesittas (commented):
Thanks for the contribution. I left some comments to consider. You'll also need to run `make style` to unblock the CI tests.

georgesittas:
Closing this due to inactivity, feel free to reopen when ready.

PR #5033, opened 2025-05-02
fix(optimizer): prevent incorrect predicate pushdown into RHS of CROSS JOIN UNNEST

### LINKED ISSUE #5023: pushdown_predicates results in invalid query when using CROSS JOIN
If I undestood it correctly, the CROSS JOIN is substituted by a JOIN with ON TRUE by the optimizer. It seems that, if the query has a WHERE predicate, it is pushed down and substitutes the ON TRUE from the JOIN.

```python
query = (
    'select t.a from test t cross join unnest(array[0, 1, 2]) as u("value") where t.b > u.value'
)
parsed = sqlglot.parse_one(query, dialect="presto")
print(pushdown_predicates(parsed, dialect="presto"))
# SELECT t.a FROM test AS t JOIN UNNEST(ARRAY(0, 1, 2)) AS u("value") ON t.b > u.value WHERE TRUE
```

I'm not sure if this is invalid in standard SQL (I don't believe it is), but it is an invalid query for dialects like [presto](https://prestodb.io/docs/current/sql/select.html#unnest).

Image

geooo109:
Fixes #5023 

This PR solves the issue of `predicate_pushdown` optimization rule when the RHS side of a `CROSS JOIN` is an `UNNEST` for Presto and Trino dIalects.

Taking into account the following query:
```
SELECT x.a, u.val FROM x AS x CROSS JOIN UNNEST(ARRAY[0, 1]) AS u("val") WHERE x.a > u.val;
```

In dialects like DuckDB, this query can be optimized by converting the `CROSS JOIN` into an `INNER JOIN` and pushing the `WHERE` condition into the `ON` part: 
```
SELECT x.a, u.val FROM x AS x JOIN UNNEST(ARRAY[0, 1]) AS u("val") ON x.a > u.val;
```

This can't be applied in Presto or Trino. 
In Presto, if the RHS of a `JOIN` is an `UNNEST` the JOIN must be `CROSS`.
In Trino, if the `INNER JOIN` contains an `UNNEST` the condtion must be `ON TRUE`.


sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -35,13 +35,20 @@ def pushdown_predicates(expression, dialect=None):
...
-                        break
-
-                pushdown(where.this, selected_sources, scope_ref_count, dialect, join_index)
+                    if isinstance(parent, exp.Join):
+                        if parent.side == "RIGHT":
+                            selected_sources = {k: (node, source)}
+                            break
+                        if parent.kind == "CROSS" and isinstance(node, exp.Unnest):
+                            join_rhs_unnest = True
+                            break
```
VaggelisD, line 48: Should we do `if isinstance(node, exp.Unnest) and dialect in ("presto", "trino")` here? Reasoning being:
- As you had said, these dialects allow only `CROSS JOIN UNNEST` on the rhs, so you can't have other kinds
- It'll make it so that the `join_rhs_unnest` condition is a standalone boolean, it won't span 2 branches

geooo109: In the first impl, I had it like your suggestion and thought it would be cleaner to have the `CROSS` in there and the `dialect in ("presto", "trino")` outside. But, yep I agree, will apply the suggestion.

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -35,13 +35,20 @@ def pushdown_predicates(expression, dialect=None):
                 }
 
                 # a right join can only push down to itself and not the source FROM table
+                # presto and trino don't support inner joins where the RHS is an UNNEST expression
+                join_rhs_unnest = False
```
georgesittas, line 39: Let's find a better name for this, its intent isn't clear now.

georgesittas (approved):
LGTM, a couple of minor comments to consider in order to refactor this.

sqlglot/optimizer/pushdown_predicates.py
geooo109: So, `unnest_requires_cross_join`?

sqlglot/optimizer/pushdown_predicates.py
```diff
@@ -35,13 +35,20 @@ def pushdown_predicates(expression, dialect=None):
...
-                    if isinstance(parent, exp.Join) and parent.side == "RIGHT":
-                        selected_sources = {k: (node, source)}
-                        break
-
-                pushdown(where.this, selected_sources, scope_ref_count, dialect, join_index)
+                    if isinstance(parent, exp.Join):
+                        if parent.side == "RIGHT":
+                            selected_sources = {k: (node, source)}
+                            break
+                        if isinstance(node, exp.Unnest) and dialect in ("presto", "trino"):
```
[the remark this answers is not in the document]
geooo109: I think your approach is better.

We can have 2 booleans:
1. `unnest_requires_cross_join` use it as you suggest.
2. `unnest_as_join_rhs` replace the `join_rhs_unnest`.

and in L(50) do `if not (unnest_as_join_rhs and unnest_requires_cross_join)`, the condition is sufficient enough.

georgesittas (approved):

PR #5040, opened 2025-05-04
Fix: prevent redundant backslash escapes in rawstring generator

### LINKED ISSUE #5038: Incorrect Translation of REGEXP_EXTRACT from BigQuery to DuckDB
There appears to be a bug in sqlglot when translating BigQuery's `REGEXP_EXTRACT` function to DuckDB. The issue results in incorrect behavior during query execution.

**Steps to Reproduce**:

1. If you run the SQL query `SELECT REGEXP_EXTRACT('2.0', r'\d+')` in BigQuery, it returns `'2'`
2. Transpiled SQL by sqlglot (to DuckDB) produces a query that, when executed, returns an empty result instead of the expected match.
 
```
import sqlglot
import duckdb
sql = sqlglot.transpile("SELECT REGEXP_EXTRACT('2.0', r'\d+')", read="bigquery", write="duckdb")[0]
print(sql)
# SELECT REGEXP_EXTRACT('2.0', '\\d+')
duckdb.sql(sql).fetchone()
('',)
```

**Expected Behavior**:
The transpiled DuckDB SQL should return the same result as BigQuery (i.e., `'2'`), ensuring semantic consistency between dialects.

**Actual Behavior**:
DuckDB returns an empty string due to incorrect SQL translation of `REGEXP_EXTRACT`.

**Impact**:
This leads to inconsistent query behavior across environments and could impact data transformation pipelines relying on regex extraction logic.

**Environment**:
- sqlglot version: [26.16.2]
- DuckDB version: [1.2.2]
- python:[3.11.9]

**Official Documentation**

- Big Query [REGEXP_EXTRACT ](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_extract)
- DuckDB [Regular Expressions](https://duckdb.org/docs/stable/sql/functions/regular_expressions.html#regexp_extractstring-pattern-group--0-options)

Image

georgesittas:
Fixes #5038

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -1349,7 +1349,11 @@ class Generator(metaclass=_Generator):
         return f"{left_quote}{this}{right_quote}{escape_sql}"
 
     def rawstring_sql(self, expression: exp.RawString) -> str:
-        string = self.escape_str(expression.this.replace("\\", "\\\\"), escape_backslash=False)
+        string = expression.this
+        if "\\" in self.dialect.tokenizer_class.STRING_ESCAPES:
+            string = string.replace("\\", "\\\\")
+
+        string = self.escape_str(string, escape_backslash=False)
         return f"{self.dialect.QUOTE_START}{string}{self.dialect.QUOTE_END}"
 
     def datatypeparam_sql(self, expression: exp.DataTypeParam) -> str:
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -237,6 +237,10 @@ LANGUAGE js AS
             "CREATE OR REPLACE VIEW test (tenant_id OPTIONS (description='Test description on table creation')) AS SELECT 1 AS tenant_id, 1 AS customer_id",
         )
         self.validate_identity(
+            'SELECT r"\\t"',
+            "SELECT '\\\\t'",
+        )
+        self.validate_identity(
             "ARRAY(SELECT AS STRUCT e.x AS y, e.z AS bla FROM UNNEST(bob))::ARRAY<STRUCT<y STRING, bro NUMERIC>>",
             "CAST(ARRAY(SELECT AS STRUCT e.x AS y, e.z AS bla FROM UNNEST(bob)) AS ARRAY<STRUCT<y STRING, bro NUMERIC>>)",
         )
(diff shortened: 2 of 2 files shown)


tests/dialects/test_bigquery.py
```diff
@@ -1023,8 +1027,8 @@ def test_bigquery(self):
             r'r"""/\*.*\*/"""',
             write={
                 "bigquery": r"'/\\*.*\\*/'",
-                "duckdb": r"'/\\*.*\\*/'",
-                "presto": r"'/\\*.*\\*/'",
+                "duckdb": r"'/\*.*\*/'",
+                "presto": r"'/\*.*\*/'",
```
georgesittas, line 1031: These tests were incorrect:

```
>>> print(r'r"""/\*.*\*/"""')  # Original BigQuery input, results in /\*.*\*/
r"""/\*.*\*/"""
>>> print(r"'/\\*.*\\*/'")  # Transpiled to DuckDB, returns redundant basklashes (same for Presto)– see below
'/\\*.*\\*/'
```

```
D select '/\\*.*\\*/';
┌──────────────┐
│ '/\\*.*\\*/' │
│   varchar    │
├──────────────┤
│ /\\*.*\\*/   │
└──────────────┘
```

sqlglot/generator.py
```diff
@@ -1349,7 +1349,11 @@ def unicodestring_sql(self, expression: exp.UnicodeString) -> str:
         return f"{left_quote}{this}{right_quote}{escape_sql}"
 
     def rawstring_sql(self, expression: exp.RawString) -> str:
-        string = self.escape_str(expression.this.replace("\\", "\\\\"), escape_backslash=False)
+        string = expression.this
+        if "\\" in self.dialect.tokenizer_class.STRING_ESCAPES:
+            string = string.replace("\\", "\\\\")
+
+        string = self.escape_str(string, escape_backslash=False)
```
VaggelisD, line 1356: Is this escaping specific to raw-strings? I wonder whether this logic should be pushed down to `escape_str` and/or whether `escape_backslash=True` would net the same result.

VaggelisD (approved):

sqlglot/generator.py
georgesittas: Yeah I think this is specific to raw strings, I remember trying to refactor it in the past and everything broke. Feel free to try though.

PR #5042, opened 2025-05-04
clickhouse: fix parsing of Null engine

### LINKED ISSUE #5041: clickhouse: engine = Null() is incorrectly parsed
Example:
```
import sqlglot

sql = """
CREATE TABLE t (i UInt8) ENGINE = Null()
"""

ex = sqlglot.parse_one(sql, "clickhouse")
print(ex.sql("clickhouse"))
```
Will produce:
```
CREATE TABLE t (i UInt8) ENGINE = NULL
```
Because it parses the engine name as NULL literal instead of `Anonymous` function with name `Null`

pkit:
Fixes #5041

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -500,6 +500,9 @@ class ClickHouse(Dialect):
 
         PROPERTY_PARSERS = parser.Parser.PROPERTY_PARSERS.copy()
         PROPERTY_PARSERS.pop("DYNAMIC")
+        PROPERTY_PARSERS["ENGINE"] = lambda self: self._parse_property_assignment(
+            exp.EngineProperty, anonymous_func=True
+        )
 
         NO_PAREN_FUNCTION_PARSERS = parser.Parser.NO_PAREN_FUNCTION_PARSERS.copy()
         NO_PAREN_FUNCTION_PARSERS.pop("ANY")
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -2196,18 +2196,22 @@ class Parser(metaclass=_Parser):
             ),
         )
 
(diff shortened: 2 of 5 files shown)


pkit:
Lol. It fails. Works fine for me in older sqlglot. Why?

VaggelisD (commented):
Hey @pkit, thank you for the contribution!

Should we instead change the generation of either `exp.Null` (if it's under an `ENGINE` property) or the `exp.EngineProperty` itself, if that issue exists only for that case?

I think it'd be preferable in contrast to changing these base parser functions, will also let the others chime in.

georgesittas:
https://github.com/tobymao/sqlglot/commit/1b3ea344af1d71d3eee239a5c4996a0aecd091de

pkit:
> [1b3ea34](https://github.com/tobymao/sqlglot/commit/1b3ea344af1d71d3eee239a5c4996a0aecd091de)

I do not like it, because it's not future-proof.
The problem here is that all `engine` props are actually anonymous functions in CH. Yes, sometimes they are written as vars, but that's only a shortcut.
So, just hacking null into func looks like a crutch...

georgesittas:
I see, that's a reasonable concern. I'll revert to your approach then.

georgesittas:
Slightly different actually to avoid introducing a new flag: https://github.com/tobymao/sqlglot/commit/9b871d5e75ecbbb4e402b38a676655793fbda1e6.

pkit:
I think `engine` is used in other sql dialects too.
I hope it doesn't change some behaviors.

georgesittas:
This is specific to ClickHouse, so I think it shouldn't impact others.

pkit:
Ah, lol, my bad, checking it from the phone. 😅

georgesittas:
No worries, thanks for reviewing 👍

PR #5049, opened 2025-05-08
add support for teradata translate with error

sdanielzafar:
According to [these Teradata docs](https://docs.teradata.com/r/Enterprise_IntelliFlex_VMware/SQL-Functions-Expressions-and-Predicates/String-Operators-and-Functions/TRANSLATE), `TRANSLATE` should support `WITH ERROR` at the end. 

This adds this by silently consuming these tokens to avoid parse errors. A unit test is also contributed.

The case issue is also addressed.

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -182,10 +182,13 @@ class Teradata(Dialect):
                 self.raise_error("Expected USING in TRANSLATE")
 
             if self._match_texts(self.CHARSET_TRANSLATORS):
-                charset_split = self._prev.text.split("_TO_")
+                charset_split = self._prev.text.upper().split("_TO_")
                 to = self.expression(exp.CharacterSet, this=charset_split[1])
             else:
                 self.raise_error("Expected a character set translator after USING in TRANSLATE")
+
+            self._match(TokenType.WITH)
+            self._match_texts({"ERROR"})
 
             return self.expression(exp.Cast if strict else exp.TryCast, this=this, to=to)
 
--- a/tests/dialects/test_teradata.py
+++ b/tests/dialects/test_teradata.py
@@ -49,6 +49,14 @@ class TestTeradata(Validator):
         )
         self.validate_identity("CAST(x AS CHAR CHARACTER SET UNICODE)")
 
+    def test_translate_with_error(self):
+        self.validate_all(
+            "TRANSLATE(x USING LATIN_TO_UNICODE WITH ERROR)",
+            write={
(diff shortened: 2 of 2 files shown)


tests/dialects/test_teradata.py
```diff
@@ -49,6 +49,14 @@ def test_translate(self):
         )
         self.validate_identity("CAST(x AS CHAR CHARACTER SET UNICODE)")
 
+    def test_translate_with_error(self):
+        self.validate_all(
+            "TRANSLATE(x USING LATIN_TO_UNICODE WITH ERROR)",
+            write={
+                "teradata": "CAST(x AS CHAR CHARACTER SET UNICODE)",
+            },
+        )
```
geooo109, line 58: Use `validate_identity`, for testing input + output of teradata.
```
        self.validate_identity(
            "TRANSLATE(x USING LATIN_TO_UNICODE WITH ERROR)",
            "CAST(x AS CHAR CHARACTER SET UNICODE)",
        )
```

sqlglot/dialects/teradata.py
```diff
@@ -182,11 +182,14 @@ def _parse_translate(self, strict: bool) -> exp.Expression:
...
 
             if self._match_texts(self.CHARSET_TRANSLATORS):
-                charset_split = self._prev.text.split("_TO_")
+                charset_split = self._prev.text.upper().split("_TO_")
                 to = self.expression(exp.CharacterSet, this=charset_split[1])
             else:
                 self.raise_error("Expected a character set translator after USING in TRANSLATE")
 
+            self._match(TokenType.WITH)
+            self._match_texts({"ERROR"})
```
geooo109, line 191: remove `{}`

sqlglot/dialects/teradata.py
```diff
@@ -182,11 +182,14 @@ def _parse_translate(self, strict: bool) -> exp.Expression:
                 self.raise_error("Expected USING in TRANSLATE")
 
             if self._match_texts(self.CHARSET_TRANSLATORS):
-                charset_split = self._prev.text.split("_TO_")
+                charset_split = self._prev.text.upper().split("_TO_")
```
geooo109, line 185: why you need `.upper()` here ? 
I think if you remove it, it works ok.

geooo109 (changes requested):

sqlglot/dialects/teradata.py
geooo109: Why is `WITH ERROR` silently removed?
This must be preserved. cc: @georgesittas

geooo109:
Thanks for the PR, I will take it to the finish line.

PR #5052, opened 2025-05-08
feat(teradata)!: support translate with error

geooo109:
Finsihes [PR 5049](https://github.com/tobymao/sqlglot/pull/5049)

In general, the `TRANSLATE` is expressed as a `CAST`.
In the case of the `WITH ERROR` a new expression is added in order to represent this special case on the AST.

**DOCS**
[Teradata Translate](https://docs.teradata.com/r/Enterprise_IntelliFlex_VMware/SQL-Functions-Expressions-and-Predicates/String-Operators-and-Functions/TRANSLATE/TRANSLATE-Function-Syntax)

### DIFF
--- a/sqlglot/dialects/teradata.py
+++ b/sqlglot/dialects/teradata.py
@@ -162,7 +162,7 @@ class Teradata(Dialect):
             # https://docs.teradata.com/r/SQL-Functions-Operators-Expressions-and-Predicates/June-2017/Data-Type-Conversions/TRYCAST
             "TRYCAST": parser.Parser.FUNCTION_PARSERS["TRY_CAST"],
             "RANGE_N": lambda self: self._parse_rangen(),
-            "TRANSLATE": lambda self: self._parse_translate(self.STRICT_CAST),
+            "TRANSLATE": lambda self: self._parse_translate(),
         }
 
         FUNCTIONS = {
@@ -175,7 +175,7 @@ class Teradata(Dialect):
             TokenType.DSTAR: exp.Pow,
         }
 
-        def _parse_translate(self, strict: bool) -> exp.Expression:
+        def _parse_translate(self) -> exp.Expression:
             this = self._parse_assignment()
             self._match(TokenType.USING)
             self._match_texts(self.CHARSET_TRANSLATORS)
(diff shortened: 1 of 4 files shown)


geooo109:
We can unnest the inner `if statement` for the `WITH ERROR` and use the `charset_split` with a f-string in order to give the `expression` the correct value. Not sure which one is cleaner.

sqlglot/dialects/teradata.py
```diff
@@ -182,7 +182,11 @@ def _parse_translate(self, strict: bool) -> exp.Expression:
                 self.raise_error("Expected USING in TRANSLATE")
 
             if self._match_texts(self.CHARSET_TRANSLATORS):
-                charset_split = self._prev.text.split("_TO_")
+                charset = self._prev.text.upper()
+                if self._match_text_seq("WITH", "ERROR"):
+                    return self.expression(exp.TranslateCharacters, this=this, expression=charset)
+
+                charset_split = charset.split("_TO_")
```
georgesittas, line 189: I wonder why we're splitting here? This doesn't look correct.

```python
>>> import sqlglot
>>> sqlglot.parse_one("TRANSLATE(x USING LATIN_TO_UNICODE)", "teradata")
Cast(
  this=Column(
  to=CharacterSet(this=UNICODE))
>>> sqlglot.parse_one("TRANSLATE(x USING LATIN_TO_UNICODE)", "teradata").sql("teradata")
'CAST(x AS CHAR CHARACTER SET UNICODE)'
```

geooo109:
Yep, my fix mainly focused just for the `WITH ERROR`, I didn't see any other case. 
I will take a look, and maybe leverage the added expression to solve this.

sqlglot/dialects/teradata.py
```diff
@@ -175,19 +175,17 @@ class Parser(parser.Parser):
             TokenType.DSTAR: exp.Pow,
         }
 
-        def _parse_translate(self, strict: bool) -> exp.Expression:
+        def _parse_translate(self) -> exp.Expression:
```
georgesittas, line 178:
```suggestion
        def _parse_translate(self) -> exp.TranslateCharacters:
```

georgesittas (approved):
Nice 👍

sqlglot/dialects/teradata.py
geooo109: yep yep, thanks

PR #5055, opened 2025-05-08
Fix(tsql): Change `COLLATE` expression to `Var` for `ALTER TABLE`

MarcusRisanger:
This is to avoid bracketing the `COLLATE` argument when transpiling to tsql in SQLMesh.

Note: This fixes the `COLLATE` transpiling for `ALTER TABLE` commands, but if `COLLATE` is used in table queries, this obviously doesn't address those cases!


georgesittas (approved):
Nice work, thanks for the contribution 👍

PR #5058, opened 2025-05-09
fix: Support EXCLUDE in window definition

rafasofizada:
This PR adds support for `EXCLUDE` in window definition:

`SELECT SUM(x) OVER (... ROWS BETWEEN ... EXCLUDE CURRENT ROW)`

`EXCLUDE` options are common across dialects supporting the clause:
- `CURRENT ROW`
- `GROUP`
- `TIES`
- `NO OTHER`

**Supported dialects:**:
- [Postgres](https://www.postgresql.org/docs/current/sql-select.html#:~:text=and%20frame_exclusion%20can%20be%20one%20of)
- [SQLite](https://sqlite.org/syntax/frame-spec.html)
- [DuckDB](https://duckdb.org/docs/stable/sql/functions/window_functions.html#syntax)
- [Oracle 21c](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Analytic-Functions.html)

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -404,6 +404,9 @@ class Generator(metaclass=_Generator):
     # Whether the function TO_NUMBER is supported
     SUPPORTS_TO_NUMBER = True
 
+    # Whether EXCLUDE in window specification is supported
+    SUPPORTS_WINDOW_EXCLUDE = False
+
     # Whether or not set op modifiers apply to the outer set op or select.
     # SELECT * FROM x UNION SELECT * FROM y LIMIT 1
     # True means limit 1 happens after the set op, False means it it happens on y.
@@ -2806,7 +2809,17 @@ class Generator(metaclass=_Generator):
             csv(self.sql(expression, "end"), self.sql(expression, "end_side"), sep=" ")
             or "CURRENT ROW"
         )
-        return f"{kind} BETWEEN {start} AND {end}"
+
+        window_spec = f"{kind} BETWEEN {start} AND {end}"
+
+        exclude = self.sql(expression, "exclude")
+        if exclude:
+            if self.SUPPORTS_WINDOW_EXCLUDE:
+                window_spec += f" EXCLUDE {exclude}"
+            else:
+                self.unsupported("EXCLUDE clause is not supported in the WINDOW clause")
+
+        return window_spec
 
     def withingroup_sql(self, expression: exp.WithinGroup) -> str:
         this = self.sql(expression, "this")
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1345,6 +1345,12 @@ class Parser(metaclass=_Parser):
         **dict.fromkeys(("DEFERRABLE", "NORELY", "RELY"), tuple()),
     }
 
+    WINDOW_EXCLUDE_OPTIONS: OPTIONS_TYPE = {
+        "NO": ("OTHERS",),
+        "CURRENT": ("ROW",),
+        **dict.fromkeys(("GROUP", "TIES"), tuple()),
+    }
+
     INSERT_ALTERNATIVES = {"ABORT", "FAIL", "IGNORE", "REPLACE", "ROLLBACK"}
 
     CLONE_KEYWORDS = {"CLONE", "COPY"}
@@ -6887,6 +6893,11 @@ class Parser(metaclass=_Parser):
             start = self._parse_window_spec()
             self._match(TokenType.AND)
             end = self._parse_window_spec()
+            exclude = (
+                self._parse_var_from_options(self.WINDOW_EXCLUDE_OPTIONS)
+                if self._match_text_seq("EXCLUDE")
+                else None
+            )
 
             spec = self.expression(
                 exp.WindowSpec,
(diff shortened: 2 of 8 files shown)


sqlglot/generator.py
```diff
@@ -404,6 +404,9 @@ class Generator(metaclass=_Generator):
     # Whether the function TO_NUMBER is supported
     SUPPORTS_TO_NUMBER = True
 
+    # Whether EXCLUDE in window specification is supported
+    SUPPORTS_WINDOW_EXCLUDE = True
```
MarcusRisanger, line 408: Should the default not be false? 🤔

rafasofizada: I set it to `True` out of confusion about testing procedure:
If it's `False`, the `EXCLUDE` clause will only be supported by the dialects that override it to `True`, right? But in that case, the test cases added to `identity.sql` will fail (I guess they use plain generator without a dialect):

(Failed test run with default `False`)
```
SELECT SUM(x) OVER (PARTITION BY a ORDER BY b ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE CURRENT ROW)
SELECT SUM(x) OVER (PARTITION BY a ORDER BY b ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE GROUP)
SELECT SUM(x) OVER (PARTITION BY a ORDER BY b ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE NO OTHERS)
SELECT SUM(x) OVER (PARTITION BY a ORDER BY b ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE TIES)
```

If that's expected, where should I put the tests for all possible `EXCLUDE` variants?

georgesittas: You can set it to `False` and then just enable it for the dialects that support it. You're currently just resetting it to the same value for DuckDB and the others, which is redundant.

As far as testing goes, you can do what you want in `test_dialect.py`– there are similar examples. It's not necessary to add the tests in `identity.sql`.

sqlglot/generator.py
```diff
@@ -2806,6 +2809,12 @@ def windowspec_sql(self, expression: exp.WindowSpec) -> str:
             csv(self.sql(expression, "end"), self.sql(expression, "end_side"), sep=" ")
             or "CURRENT ROW"
         )
+
+        if self.SUPPORTS_WINDOW_EXCLUDE:
+            exclude = self.sql(expression, "exclude")
+            if exclude:
+                return f"{kind} BETWEEN {start} AND {end} EXCLUDE {exclude}"
+
         return f"{kind} BETWEEN {start} AND {end}"
```
georgesittas, line 2818: This is how we usually implement generation logic:

```suggestion
        window_spec = f"{kind} BETWEEN {start} AND {end}"

        exclude = self.sql(expression, "exclude")
        if exclude:
            if self.SUPPORTS_WINDOW_EXCLUDE:
                window_spec += f" EXCLUDE {exclude}"
            else:
                self.unsupported("EXCLUDE clause is not supported in the WINDOW clause")

        return window_spec
```

sqlglot/parser.py
```diff
@@ -6936,6 +6938,17 @@ def _parse_window_spec(self) -> t.Dict[str, t.Optional[str | exp.Expression]]:
...
+    def _parse_window_frame_exclude(self) -> t.Optional[str]:
+        if self._match_text_seq("EXCLUDE"):
+            return (
+                (self._match_text_seq("CURRENT", "ROW") and "CURRENT ROW")
+                or (self._match_text_seq("GROUP") and "GROUP")
+                or (self._match_text_seq("TIES") and "TIES")
+                or (self._match_text_seq("NO", "OTHERS") and "NO OTHERS")
+            )
+
+        return None
```
georgesittas, line 6950: Let's not do it like this. Look into `_parse_var_from_options`, I think the implementation should use it.

georgesittas (commented):
Thanks for the contribution! Take a look at my suggestions and let me know if something's unclear.

sqlglot/generator.py
rafasofizada: Makes sense, set the default to `False` and `True` for each supported dialect – thanks for the `test_dialect` pointer, the tests tripped me up here.

I had to remove the windows-with-exclude from `identity.sql`, removed individual transpilation tests from `test_duckdb` [and updated tests in `test_dialect.py`](https://github.com/tobymao/sqlglot/pull/5058/files#diff-681a3cc02d9013ab87631ebf5a790fe9bd0022a1ee2b4b34613c1b22d2d9bed1)

sqlglot/parser.py
rafasofizada: Done https://github.com/tobymao/sqlglot/pull/5058/files#diff-63eb8dd82d3561bc414e3e13cf59516bd6584c5766a10072157a8a5aee4ad85f

rafasofizada:
@georgesittas not sure where the code change suggestion went, [but I applied it here](https://github.com/tobymao/sqlglot/pull/5058/files#diff-362498362ae539e216cd83ebfe510977107648742f1a1bee7b2db7b5b7d27974R2812-R2822).

Sorry for the mess, I'm used to GitLab's MR process

georgesittas (approved):

Issue #5059, opened 2025-05-10
Druid: CURRENT_TIMESTAMP function being read and transpiled wrong


ALongJohnson:
Hi i noticed this issue when reading and transpiling druid sql.. I have made the relevant changes needed on my local, and would like to make a pr contribution for this.
This bug will cause it to fail on druid side with the error 
`Error: Plan validation failed
org.apache.calcite.runtime.CalciteContextException: From line 1, column 48 to line 1, column 66: No match found for function signature CURRENT_TIMESTAMP()
org.apache.calcite.tools.ValidationException`

**Fully reproducible code snippet**
`
sql_query = "SELECT * FROM \"TABLE\" WHERE  \"__time\" >= CURRENT_TIMESTAMP - INTERVAL '1' DAY"
ast = parse_one(sql_query,read="druid")
print(ast.sql(dialect="druid"))

#Output : "SELECT * FROM \"TABLE\" WHERE  \"__time\" >= CURRENT_TIMESTAMP() - INTERVAL '1' DAY"

#Expected Output: "SELECT * FROM \"TABLE\" WHERE  \"__time\" >= CURRENT_TIMESTAMP - INTERVAL '1' DAY"
`

**Official Documentation**
Please include links to official SQL documentation related to your issue.
https://druid.apache.org/docs/latest/querying/sql-functions/#current_timestamp

PR #5061, opened 2025-05-10
fix: support current_timestamp

ALongJohnson:
fixes https://github.com/tobymao/sqlglot/issues/5059

### DIFF
--- a/sqlglot/dialects/druid.py
+++ b/sqlglot/dialects/druid.py
@@ -12,3 +12,7 @@ class Druid(Dialect):
             exp.DataType.Type.TEXT: "STRING",
             exp.DataType.Type.UUID: "STRING",
         }
+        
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/druid.py
```diff
@@ -12,3 +12,7 @@ class Generator(generator.Generator):
             exp.DataType.Type.TEXT: "STRING",
             exp.DataType.Type.UUID: "STRING",
         }
+        
+        def currenttimestamp_sql(self, expression: exp.CurrentTimestamp) -> str:
+            this = expression.this
+            return self.func("CURRENT_TIMESTAMP", this) if this else "CURRENT_TIMESTAMP"
```
georgesittas, line 18: We don't want a function right? You can only use it without parentheses.

ALongJohnson: Interesting, it worked for my usecase but i see the way you've done it in your commit. Thanks !

PR #5079, opened 2025-05-15
Add support for qualified column names in Teradata dialect.

sdanielzafar:
Hello! We are facing a parsing issue in Teradata when a qualified column is used in an INSERT INTO table's schema. For instance this is valid in Teradata:

```
INSERT INTO x (y.b, y.c) SELECT * FROM tbl_y AS y JOIN tbl_z AS z ON y.a = z.a"
````
The documentation does not state this explicitly, but I double checked by running this query in Teradata Studio:
image

I'm submitting this PR with a proposed fix for this issue, which is a little bit tricky because it involves overriding the `_parse_insert` for the Teradata dialect and then introducing a new argument which will be added to the base parser class, which is not ideal. I did my best to remove any non-Teradata functionality from the override and confirmed all tests are passing.

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -689,6 +689,7 @@ class ClickHouse(Dialect):
             parse_bracket: bool = False,
             is_db_reference: bool = False,
             parse_partition: bool = False,
+            qualified_schema_col: bool = False,
         ) -> t.Optional[exp.Expression]:
             this = super()._parse_table(
                 schema=schema,
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -502,6 +502,7 @@ class DuckDB(Dialect):
             parse_bracket: bool = False,
             is_db_reference: bool = False,
             parse_partition: bool = False,
+            qualified_schema_col: bool = False,
         ) -> t.Optional[exp.Expression]:
             # DuckDB supports prefix aliases, e.g. FROM foo: bar
             if self._next and self._next.token_type == TokenType.COLON:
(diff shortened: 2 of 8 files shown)


georgesittas:
Hey 👋

I'm confused, what does `INSERT INTO x (y.b, y.c) ...` even mean?

sdanielzafar:
@georgesittas - I think it's stupid too. But I'm seeing it all over legacy code and it needs to be parsed. It basically just uses the column name. AFAIK the table name is not used.

georgesittas:
I see, I lack context but if feels like an unintended leakage of teradata's parser internals. Not sure how to interpret it otherwise.

I don't think we'll merge this one. Introducing complexity to handle highly non-standard syntax like in this case is avoided by SQLGlot for maintainability reasons. We had similar issues in the past and closed them as not planned too.

Thanks for the contribution, though! If you really need to parse this, then you could probably override Teradata in your application, it shouldn't be a big lift.

sdanielzafar:
I would really appreciate a path forward here. I know Teradata's parser is wildly permissive and that can pose problems. I'd like to be able to advise my team at Databricks to use `sqlglot`, but if we get errors parsing basic queries it adds a lot of friction with any work we are doing using this package. Can we please identify a way forward that avoids clunky overrides at the application level?

barakalon:
Seems to make sense in the context of a “INSERT INTO… SELECT” statement? Although who knows about the y.id in the example above…

Where do we draw the line on “nonstandard”? Should it be more about controlling the blast radius of certain changes? I.e. we keep this change isolated to the teradata dialect.

sdanielzafar:
@barakalon - any update on the question you posed above?

PR #5081, opened 2025-05-16
fix: concat should be a string

tobymao:


VaggelisD (approved):
That should be the case in most scenarios, just leaving it as a note that some dialects like Snowflake also allow other types (e.g `BINARY`, NULLs etc) but we'd need to fix them like we do in BQ/Spark.

georgesittas:
discussed offline, this is unnecessary because concat's type is already a varchar, see https://github.com/tobymao/sqlglot/blob/ba52f014f0d53ce8a179f1b140876274a01b38ac/sqlglot/dialects/dialect.py#L708

Issue #5091, opened 2025-05-20
Token col attribute doesn't handle leading whitespace


chgiff:
When tokenizing, the `col` attribute that denotes where the token ends is incorrect if there is a newline followed by leading whitespace.

**Fully reproducible code snippet**
I would expect this behavior: 
```
        tokens = Tokenizer().tokenize("SELECT column\n    FROM table")

        self.assertEqual(tokens[0].line, 1) # SELECT
        self.assertEqual(tokens[0].col, 6)
        self.assertEqual(tokens[1].line, 1) # column
        self.assertEqual(tokens[1].col, 13)
        self.assertEqual(tokens[2].line, 2) # FROM
        self.assertEqual(tokens[2].col, 8)
        self.assertEqual(tokens[3].line, 2) # table
        self.assertEqual(tokens[3].col, 14)
```
but the current implementation would have col==5 for the `FROM` token and col==11 for the `table` token.

**Official Documentation**
[Doc for Token specifies that col is the column where the token ends.](https://sqlglot.com/sqlglot/tokens.html#Token)

chgiff:
I believe the solution is as simple as changing this section in the `_advance` method from a `1` to an `i`.
```
            if not (self._char == "\r" and self._peek == "\n"):
                self._col = 1
                self._line += 1
```
should be 
```
            if not (self._char == "\r" and self._peek == "\n"):
                self._col = i
                self._line += 1
```
and same for the Rust tokenizer. 

I'm happy to open a PR with the fix if there's agreement this is the correct route to go.

georgesittas:
Hey @chgiff, your suggestion seems good. We'll be happy to accept a well-tested PR for this, so feel free to send one our way!

PR #5094, opened 2025-05-20
Fix token col attribute when there is leading whitespace after newline

chgiff:
Addresses #5091, by initializing the self._col value to `i` instead of `1` after a newline, which will correctly account for any leading whitespace which is accumulated in the _scan method.

### DIFF
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -1109,7 +1109,7 @@ class Tokenizer(metaclass=_Tokenizer):
         if self.WHITE_SPACE.get(self._char) is TokenType.BREAK:
             # Ensures we don't count an extra line if we get a \r\n line break sequence
             if not (self._char == "\r" and self._peek == "\n"):
-                self._col = 1
+                self._col = i
                 self._line += 1
         else:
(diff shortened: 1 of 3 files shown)


georgesittas (approved):
Thank you for digging into this and contributing a fix 🙏 

Did you check whether the updated test would've caught the issue prior to your fix? I was originally thinking of including the same example as you shared in the issue, but if this test hits that code path then it's fine.

cc @tobymao for visibility, since this touches the tokenizer's `_advance`, which is critical.

tobymao (approved):

chgiff:
@georgesittas The updated tests will fail without the fix in the tokenizer. So it does cover the code path in question.

georgesittas:
Great, thanks for checking.

chgiff:
Thanks for the quick review @georgesittas and @tobymao. Do I need to create another PR to bump the version in Cargo.toml? Or what's the process to get this into a release?

georgesittas:
No need– I'm currently working on another thing and when that's ready I'll cut a release and update this thread. We release very frequently.

PR #5097, opened 2025-05-20
Feat: transpile WINDOW clause

georgesittas:
Dialects that support the `WINDOW` clause: BigQuery, ClickHouse, DuckDB, Oracle, PostgreSQL, MySQL, SQLite, T-SQL, Databricks, Spark SQL, Trino.

- https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#window_clause
- https://dev.mysql.com/doc/refman/8.4/en/select.html
- https://www.sqlite.org/lang_select.html
- https://learn.microsoft.com/en-us/sql/t-sql/queries/select-window-transact-sql?view=sql-server-ver16
- https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-select-named-window
- https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select.html
- https://clickhouse.com/docs/sql-reference/statements/select
- https://duckdb.org/docs/stable/sql/statements/select.html#window-clause
- https://trino.io/docs/current/sql/select.html#synopsis
- https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6__SECTION_XZB_MTC_4YB

Presto, Snowflake, Redshift **DO NOT** support named windows, verified this by executing queries against them directly.

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -965,16 +965,47 @@ def any_to_exists(expression: exp.Expression) -> exp.Expression:
     transformation
     """
     if isinstance(expression, exp.Select):
-        for any in expression.find_all(exp.Any):
-            this = any.this
+        for any_expr in expression.find_all(exp.Any):
+            this = any_expr.this
             if isinstance(this, exp.Query):
                 continue
 
-            binop = any.parent
+            binop = any_expr.parent
             if isinstance(binop, exp.Binary):
                 lambda_arg = exp.to_identifier("x")
-                any.replace(lambda_arg)
+                any_expr.replace(lambda_arg)
                 lambda_expr = exp.Lambda(this=binop.copy(), expressions=[lambda_arg])
                 binop.replace(exp.Exists(this=this.unnest(), expression=lambda_expr))
+
+    return expression
+
+
+def eliminate_window_clause(expression: exp.Expression) -> exp.Expression:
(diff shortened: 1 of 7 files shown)


sqlglot/transforms.py
```diff
@@ -965,16 +965,47 @@ def any_to_exists(expression: exp.Expression) -> exp.Expression:
...
+
+            window.set("alias", None)
+            for key in ("partition_by", "order", "spec"):
+                arg = inherited_window.args.get(key)
+                if arg:
+                    window.set(key, arg.copy())
+
+        for window in windows:
+            _inline_inherited_window(window)
+            window_expression[window.name.lower()] = window
```
georgesittas, line 1006: FYI, I tried to execute a query like the following in both BigQuery and Postgres, but it's not supported:

```sql
SELECT
  *
FROM (
  SELECT LAST_VALUE(x) OVER w AS most_popular FROM p
) AS subq
WINDOW w AS (PARTITION BY c)
```

This means named windows don't follow the usual lexical scoping rules from what I can tell, at least for these two engines. This is why the scope-level `find_all` is enough to solve this issue.

tobymao (approved):

PR #5099, opened 2025-05-21
feat(starrocks): Support parsing "NONE" as security option

alpolishchuk:
StarRocks security option for CREATE VIEW can contain "NONE" as a value

[Create View Documentation](https://docs.starrocks.io/docs/sql-reference/sql-statements/View/CREATE_VIEW/#parameters)


georgesittas (approved):
Thanks! Let's add a `validate_identity` test in `test_starrocks.py` and I'll merge.

PR #5100, opened 2025-05-21
Fix(optimizer)!: pass dialect to ensure_schema

### LINKED ISSUE #5098: Annotating types for dialect snowflake does not work
**TLDR**
Firstly, thanks for an amazing library! Annotating the types for dialect `snowflake` does not seem to work. Furthermore supplying a MappingSchema even seems to break the qualifying step. I hope I have not overlooked anything important here. We are using sqlglot for a variety of dialects and this problem only happens for `snowflake`:

**Example 1**

To Reproduce supply a simple schema and a simple query an qualify and annotate_types:

```py
import sqlglot

dialect = "snowflake"
schema = {
        "SCHEMA": {
            "TBL": {"COL": "INT", "col2": "VARCHAR"},
        }
    }
example_query = 'SELECT * FROM "SCHEMA"."TBL"'

# get qualified expression
expression = sqlglot.parse_one(example_query, dialect=dialect)
qual = qualify(expression, schema=schema, dialect=dialect)

# print the qualified statement
print(qual.sql(dialect=dialect))

# get the annotated types
expression = annotate_types(expression, schema=schema, dialect=dialect)
for col in expression.selects:
    print(col.output_name)
    print(col.type)
```

In this example you can see that the qualifier works (star select is expaneded) but the column types are set to UNKNOWN after annotating the types. The exact same code works for dialect `duckdb`.

**Example 2**

```py
import sqlglot

dialect = "snowflake"
schema = ensure_schema({
        "SCHEMA": {
            "TBL": {"COL": "INT", "col2": "VARCHAR"},
        }
    })
example_query = 'SELECT * FROM "SCHEMA"."TBL"'

# get qualified expression
expression = sqlglot.parse_one(example_query, dialect=dialect)
qual = qualify(expression, schema=schema, dialect=dialect)

# print the qualified statement
print(qual.sql(dialect=dialect))
```

When supplying a `MappingSchema` instead of a python dict, the star select expansion of the qualifying step already fails. For `duckdb` a `MappingSchema` also works.

georgesittas:
Fixes #5098


geooo109 (approved):

sqlglot/optimizer/pushdown_projections.py
```diff
@@ -34,9 +47,9 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
         sqlglot.Expression: optimized expression
     """
     # Map of Scope to all columns being selected by outer queries.
-    schema = ensure_schema(schema)
-    source_column_alias_count = {}
-    referenced_columns = defaultdict(set)
+    schema = ensure_schema(schema, dialect=dialect)
```
VaggelisD, line 50: This pattern predates me so I may be missing context but it seems error prone, no? I'd expect that we have called `ensure_schema` at the beginning of `optimize(...)`, so since these rules shouldn't be called in isolation, they should get that for free from the upstream rules. 

I guess it's fine for `qualify` and `annotate_types` which _do_ get used as standalones, but this issue goes to show that spreading out these checks is bound to cause headaches.

I think the best solution would be to advocate using `optimise(..., rules=[...])` for chaining rules so that its entrypoint can handle those early.

VaggelisD (approved):

sqlglot/optimizer/pushdown_projections.py
georgesittas: We do, but you may call these in isolation anyway and the type hint allows for a dict to be passed in, so small lift to make the behavior consistent.

georgesittas: We can suggest this pattern, but since user code may already be relying on these calls, I'd rather we do any followup work in a separate PR, after aligning with @barakalon and @tobymao.

PR #5106, opened 2025-05-22
FIX: Return parameterized `type2` in `_maybe_coerce`

aninhalbuquerque:
## Description:
This change corrects the coercion logic in the `_maybe_coerce` helper so that, when the target type (`type2`) is a parameterized `DataType`, it is returned — rather than falling back to the source type (`type1`).

### Before:

```python
if isinstance(type2, exp.DataType):
    if type2.expressions:
        return type1     # ← returns the wrong type when type2 is parameterized
```

### After:

```python
if isinstance(type2, exp.DataType):
    if type2.expressions:
        return type2     # ← correctly returns the parameterized target type
```


aninhalbuquerque:
Hi @georgesittas! I’ve added two new tests to the `annotate_types` suite. Prior to my fix, the first test was inferring the data type as `NULL`.

georgesittas (approved):
Well done, thank you.

PR #5109, opened 2025-05-23
Fix(snowflake): Put COPY GRANTS in the right place for materialized views

erindru:
Addresses SQLMesh issue: https://github.com/TobikoData/sqlmesh/issues/4181

Currently, if `exp.CopyGrantsProperty` is present in a `CREATE VIEW` then it is output after the columns list. This is in line with the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax).

However, for `CREATE MATERIALIZED VIEW`, it [needs to go](https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view#syntax) *before* the columns list or the following error will be thrown:
```
>>> CREATE MATERIALIZED VIEW ERIN_TEST_MV ("ITEM_ID", "NUM_ORDERS") COPY GRANTS AS SELECT ITEM_ID, NUM_ORDERS FROM UPSTREAM_TABLE

SQL compilation error: syntax error line 1 at position 50 unexpected '"ITEM_ID"'.
```

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1387,7 +1387,7 @@ class Snowflake(Dialect):
                 # We default CopyGrantsProperty to POST_SCHEMA which means we need to output it POST_NAME if a materialized view is detected
                 # ref: https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view#syntax
                 # ref: https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax
-                post_schema_properties = locations.get(exp.Properties.Location.POST_SCHEMA, [])
+                post_schema_properties = locations[exp.Properties.Location.POST_SCHEMA]
                 post_schema_properties.pop(post_schema_properties.index(copy_grants_property))
 
                 this_name = self.sql(expression.this, "this")
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1376,3 +1376,25 @@ class Snowflake(Dialect):
             if offset and not limit:
                 expression.limit(exp.Null(), copy=False)
             return super().select_sql(expression)
+
+        def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> str:
+            kind = self.sql(expression, "kind").upper()
+            is_materialized = expression.find(exp.MaterializedProperty)
+            copy_grants_property = expression.find(exp.CopyGrantsProperty)
+
+            if kind == "VIEW" and is_materialized and copy_grants_property:
+                # For materialized views, COPY GRANTS is located *before* the columns list
+                # This is in contrast to normal views where COPY GRANTS is located *after* the columns list
+                # We default CopyGrantsProperty to POST_SCHEMA which means we need to output it POST_NAME if a materialized view is detected
+                # ref: https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view#syntax
+                # ref: https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -1376,3 +1376,25 @@ def select_sql(self, expression: exp.Select) -> str:
             if offset and not limit:
                 expression.limit(exp.Null(), copy=False)
             return super().select_sql(expression)
+
+        def createable_sql(self, expression: exp.Create, locations: t.DefaultDict) -> str:
+            kind = self.sql(expression, "kind").upper()
```
georgesittas, line 1381:
```suggestion
```

sqlglot/dialects/snowflake.py
```diff
@@ -1376,3 +1376,25 @@ def select_sql(self, expression: exp.Select) -> str:
...
+            is_materialized = expression.find(exp.MaterializedProperty)
+            copy_grants_property = expression.find(exp.CopyGrantsProperty)
+
+            if kind == "VIEW" and is_materialized and copy_grants_property:
```
georgesittas, line 1385:
```suggestion
            if expression.kind == "VIEW" and is_materialized and copy_grants_property:
```

sqlglot/dialects/snowflake.py
```diff
@@ -1376,3 +1376,25 @@ def select_sql(self, expression: exp.Select) -> str:
...
+                # For materialized views, COPY GRANTS is located *before* the columns list
+                # This is in contrast to normal views where COPY GRANTS is located *after* the columns list
+                # We default CopyGrantsProperty to POST_SCHEMA which means we need to output it POST_NAME if a materialized view is detected
+                # ref: https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view#syntax
+                # ref: https://docs.snowflake.com/en/sql-reference/sql/create-view#syntax
+                post_schema_properties = locations.get(exp.Properties.Location.POST_SCHEMA, [])
```
georgesittas, line 1391: Current code can fail in the `index` call below if the lookup fails here:
```suggestion
                post_schema_properties = locations[exp.Properties.Location.POST_SCHEMA]
```

Otherwise, you could also slightly rewrite as follows if you want to be defensive:

```python
post_schema_properties = locations.get(exp.Properties.Location.POST_SCHEMA, [])
if copy_grants_property in post_schema_properties:
    post_schema_properties.remove(copy_grants_property)
```

But not sure why we'd fail to either find that property location in `locations` or fail to get the index of the property.

sqlglot/dialects/snowflake.py
```diff
@@ -1376,3 +1376,25 @@ def select_sql(self, expression: exp.Select) -> str:
...
+                post_schema_properties.pop(post_schema_properties.index(copy_grants_property))
+
+                this_name = self.sql(expression.this, "this")
+                copy_grants = self.sql(copy_grants_property)
+                this_schema = self.schema_columns_sql(expression.this)
+
+                return f"{this_name}{self.sep()}{copy_grants}{self.sep()}{this_schema}"
```
georgesittas, line 1398: Should we mimic how ClickHouse generates `this_schema` here?

https://github.com/tobymao/sqlglot/blob/10fe4c039a15f12c97bdf74e2e4cf547691f8546/sqlglot/dialects/clickhouse.py#L1300-L1303

I wonder if this generates an extraneous whitespace character in case there's no schema. If this isn't even valid syntax in Snowflake though then feel free to ignore.

georgesittas (approved):
LGTM, just a few minor comments.

erindru:
Thanks for taking this to the finish line!

PR #5121, opened 2025-05-26
fix(optimizer): avoid creating new alias for qualifying unpivot

geooo109:
This PR fixes the problem of `qualify_tables` for `UNPIVOT`.

In the previous implementation, we incorrectly generated a new alias using the `next_alias_name()` method for each `UNPIVOT` operation. This approach is incompatible with certain SQL dialects, such as Snowflake, which do not allow an alias to be assigned directly to the result of an `UNPIVOT`.

Additionally, introducing a new alias caused inconsistencies in column projections. Some columns were referenced via the original table alias from the FROM clause, while others — affected by the `UNPIVOT` — used the newly generated alias. This approach led to incorrect or ambiguous references in the final query.

To resolve this, we now retain only the original alias of the source table, ensuring consistent and valid column references across all dialects.

For example:
```
input SQL (snowflake):
SELECT t1, t2, t3
FROM tb AS t
UNPIVOT (t3 for t2 in (col1, col2));

output SQL (snowflake qualified):
SELECT
  "T"."T1" AS "T1",
  "T"."T2" AS "T2",
  "T"."T3" AS "T3"
FROM "TB" AS "T"
UNPIVOT("T3" FOR "T2" IN ("COL1", "COL2")) AS "T"
```

### DIFF
--- a/sqlglot/optimizer/qualify_tables.py
+++ b/sqlglot/optimizer/qualify_tables.py
@@ -100,9 +100,12 @@ def qualify_tables(
                 )
 
                 if pivots:
-                    if not pivots[0].alias:
-                        pivot_alias = next_alias_name()
-                        pivots[0].set("alias", exp.TableAlias(this=exp.to_identifier(pivot_alias)))
+                    pivot = pivots[0]
+                    if not pivot.alias:
+                        pivot_alias = (
+                            source.alias if pivot.args.get("unpivot") else next_alias_name()
+                        )
+                        pivot.set("alias", exp.TableAlias(this=exp.to_identifier(pivot_alias)))
 
                     # This case corresponds to a pivoted CTE, we don't want to qualify that
                     if isinstance(scope.sources.get(source.alias_or_name), Scope):
--- a/tests/fixtures/optimizer/optimizer.sql
+++ b/tests/fixtures/optimizer/optimizer.sql
@@ -693,14 +693,14 @@ PIVOT(MAX("SOURCE"."VALUE") FOR "SOURCE"."KEY" IN ('a', 'b', 'c')) AS "FINAL"("I
 # dialect: snowflake
 SELECT * FROM m_sales AS m_sales(empid, dept, jan, feb) UNPIVOT(sales FOR month IN (jan, feb)) ORDER BY empid;
 SELECT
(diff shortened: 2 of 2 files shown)


sqlglot/optimizer/qualify_tables.py
```diff
@@ -100,9 +100,12 @@ def _qualify(table: exp.Table) -> None:
...
 
                 if pivots:
-                    if not pivots[0].alias:
-                        pivot_alias = next_alias_name()
-                        pivots[0].set("alias", exp.TableAlias(this=exp.to_identifier(pivot_alias)))
+                    pivot = pivots[0]
+                    if not pivot.alias:
+                        pivot_alias = (
+                            source.alias if pivot.args.get("unpivot") else next_alias_name()
+                        )
```
georgesittas, line 107:
```suggestion
                        pivot_alias =  source.alias if pivot.unpivot else next_alias_name()
```

tests/fixtures/optimizer/optimizer.sql
```diff
@@ -693,14 +693,14 @@ PIVOT(MAX("SOURCE"."VALUE") FOR "SOURCE"."KEY" IN ('a', 'b', 'c')) AS "FINAL"("I
...
-  "_Q_0"."DEPT" AS "DEPT",
-  "_Q_0"."MONTH" AS "MONTH",
-  "_Q_0"."SALES" AS "SALES"
+  "M_SALES"."EMPID" AS "EMPID",
+  "M_SALES"."DEPT" AS "DEPT",
+  "M_SALES"."MONTH" AS "MONTH",
+  "M_SALES"."SALES" AS "SALES"
 FROM "M_SALES" AS "M_SALES"("EMPID", "DEPT", "JAN", "FEB")
-UNPIVOT("SALES" FOR "MONTH" IN ("JAN", "FEB")) AS "_Q_0"
+UNPIVOT("SALES" FOR "MONTH" IN ("JAN", "FEB")) AS "M_SALES"
```
georgesittas, line 701: Hmm I actually wonder how this affects column-level lineage when `UNPIVOT`s are involved.

geooo109: Lineage needs extra handling, so this is work for a future PR.

PR #5123, opened 2025-05-27
fix(optimizer)!: make coalesce simplification optional, skip by default

barakalon:
This simplification is a bit niche and might make queries _more_ complex in most cases.

Let's skip it by default.


geooo109 (approved):

georgesittas (approved):
@tobymao should we keep using this rule in sqlmesh?

tobymao:
@georgesittas naw

PR #5125, opened 2025-05-28
fix(athena): always generate string literal from FileFormatProperty

aguynamedryan:
As described in https://github.com/tobymao/sqlglot/discussions/5124, Athena requires the WITH(format='parquet') to use a string literal on the right hand side of the the format=

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -71,6 +71,13 @@ def _partitioned_by_property_sql(self: Athena.Generator, e: exp.PartitionedByPro
     return f"{prop_name}={self.sql(e, 'this')}"
 
 
+def _file_format_property_sql(self: Athena.Generator, e: exp.FileFormatProperty) -> str:
+    this = e.args.get("this")
+    if not this:
+        return "format=''"
+    return f"format={exp.Literal.string(this.name)}"
+
+
 class Athena(Trino):
     """
     Over the years, it looks like AWS has taken various execution engines, bolted on AWS-specific modifications and then
@@ -148,7 +155,7 @@ class Athena(Trino):
 
         TRANSFORMS = {
             **Trino.Generator.TRANSFORMS,
-            exp.FileFormatProperty: lambda self, e: f"format={self.sql(e, 'this')}",
+            exp.FileFormatProperty: _file_format_property_sql,
             exp.PartitionedByProperty: _partitioned_by_property_sql,
             exp.LocationProperty: _location_property_sql,
         }
--- a/tests/dialects/test_athena.py
+++ b/tests/dialects/test_athena.py
@@ -337,3 +337,22 @@ class TestAthena(Validator):
         assert isinstance(parsed.this, exp.Schema)
         assert next(n for n in parsed.this.expressions if isinstance(n, exp.PartitionedByBucket))
         assert next(n for n in parsed.this.expressions if isinstance(n, exp.PartitionByTruncate))
+
+    def test_ctas_uses_string_for_format(self):
+        # When transpiling, sometimes the FileFormatProperty is a Var, not a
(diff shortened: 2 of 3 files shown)


tests/dialects/test_dialect.py
```diff
@@ -16,7 +17,7 @@
 
 
 class Validator(unittest.TestCase):
-    dialect = None
+    dialect: Optional[str] = None
```
georgesittas, line 20: Why did this change? The rest of the file is not typed.

tests/dialects/test_athena.py
```diff
@@ -337,3 +337,22 @@ def test_parse_partitioned_by_returns_iceberg_transforms(self):
         assert isinstance(parsed.this, exp.Schema)
         assert next(n for n in parsed.this.expressions if isinstance(n, exp.PartitionedByBucket))
         assert next(n for n in parsed.this.expressions if isinstance(n, exp.PartitionByTruncate))
+
+    def test_ctas_uses_string_for_format(self):
```
georgesittas, line 341: Can we add a `validate_all` test as well to demonstrate that Spark -> Athena is now covered?

sqlglot/dialects/athena.py
```diff
@@ -71,6 +71,13 @@ def _partitioned_by_property_sql(self: Athena.Generator, e: exp.PartitionedByPro
     return f"{prop_name}={self.sql(e, 'this')}"
 
 
+def _file_format_property_sql(self: Athena.Generator, e: exp.FileFormatProperty) -> str:
+    this = e.args.get("this")
+    if not this:
+        return "format=''"
+    return f"format={exp.Literal.string(this.name)}"
```
georgesittas, line 78: This seems a bit more verbose than necessary, let's inline it in `TRANSFORMS`:

```python
            exp.FileFormatProperty: lambda self, e: f"format={self.sql(exp.Literal.string(e.name)}",
```

georgesittas:
@erindru can you take a look as well? I could reproduce the issue with a non-string:

```python
>>> from sqlmesh import Context
>>> ctx = Context()
>>> ctx.engine_adapter
<sqlmesh.core.engine_adapter.athena.AthenaEngineAdapter object at 0x121765400>
>>> ctx.engine_adapter.execute('CREATE TABLE "db"."tab" WITH (format=parquet) AS SELECT 1 AS "col"')
botocore.errorfactory.InvalidRequestException: An error occurred (InvalidRequestException) when calling the StartQueryExecution operation: expression parquet is not a string literal
```

erindru:
Err, how is that a non-string? You pass a string to `engine_adapter.execute()` which means that it isnt touched by SQLGlot at all?

My understanding is that the problem is this:
```
>>> parse_one("""CREATE TABLE `db`.`tab` USING parquet AS SELECT 1 AS `col`""", dialect="spark")
Create(
  ...SNIP...
      FileFormatProperty(this=Var(this=parquet))]))
```

`exp.Var()` is correct for Spark because it's not a string literal on Spark so should not have the quotes.

```
>>> parse_one("create table foo with(format='parquet') as select 1", dialect="athena")
Create(
  ...SNIP...
      FileFormatProperty(this=Literal(this='parquet', is_string=True))]))
```

This results in `exp.Literal` for Athena because in Athena it _is_ a string literal and not a variable/identifier, so should have the quotes

So the issue is transpilation.

> I think if we got rid of that FileFormatProperty entry it'd just work

Almost. Unfortunately this results in the property name getting converted to uppercase, but the casing matters for Athena. `FORMAT='PARQUET'` throws an error:
```
Table properties [FORMAT] are not supported.
```
but `format='PARQUET'` works fine.

In researching the right thing to do here and adjusting the tests to check my theory, I ended up implementing this: https://github.com/tobymao/sqlglot/pull/5133

PR #5128, opened 2025-05-28
feat(parser): support SELECT, FROM, WHERE with pipe syntax

geooo109:
This PR aims to start the implementation of the pipe syntax. 

**DOCS**
[BigQuery Pipe Syntax](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax)
[Databricks Pipe Syntax](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-qry-pipeline)

| `LIMIT`        |          |
| `AGGREGATE`        |          |
| `ORDER BY` |  |
| `UNION`               |           |
| `INTERSECT`               |           |
| `EXCEPT`               |           |

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -1272,7 +1272,7 @@ class Query(Expression):
             The modified expression.
         """
         return _apply_conjunction_builder(
-            [expr.this if isinstance(expr, Where) else expr for expr in expressions],
+            *[expr.this if isinstance(expr, Where) else expr for expr in expressions],
             instance=self,
             arg="where",
             append=append,


sqlglot/parser.py
```diff
@@ -7132,6 +7139,11 @@ def _parse_tokens(
 
         return this
 
+    def _parse_query(self, query: t.Optional[exp.Query] = None) -> t.Optional[exp.Query]:
```
georgesittas, line 7142: - What about `_parse_pipe_syntax_query`?
- Why would `query` and the return type be optional?

sqlglot/parser.py
```diff
@@ -1116,6 +1116,11 @@ class Parser(metaclass=_Parser):
         "TRUNCATE": lambda self: self._parse_partitioned_by_bucket_or_truncate(),
     }
 
+    TRANSFORM_PARSERS = {
+        "SELECT": lambda self, query: query.select(*self._parse_select(), append=False, copy=False),
+        "WHERE": lambda self, query: query.where(self._parse_assignment(), copy=False),
```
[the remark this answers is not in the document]
geooo109: got it from PRQL, will change it.

sqlglot/dialects/prql.py
```diff
@@ -88,7 +89,7 @@ def _parse_statement(self) -> t.Optional[exp.Expression]:
             expression = expression if expression else self._parse_query()
             return expression
 
-        def _parse_query(self) -> t.Optional[exp.Query]:
+        def _parse_query(self, query: t.Optional[exp.Query] = None) -> t.Optional[exp.Query]:
```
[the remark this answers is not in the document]
geooo109: yep will separate them.

sqlglot/parser.py
```diff
@@ -7132,6 +7151,11 @@ def _parse_tokens(
 
         return this
 
+    def _parse_pipe_syntax_query(self, query: exp.Query) -> exp.Query:
```
georgesittas, line 7154: If I'm not mistaken we can only get a `Select` passed here, right? Via `_parse_from`:
```suggestion
    def _parse_pipe_syntax_query(self, query: exp.Select) -> exp.Query:
```

sqlglot/parser.py
```diff
@@ -1116,6 +1116,23 @@ class Parser(metaclass=_Parser):
...
+        select = self._parse_select()
+        if isinstance(select, exp.Select):
+            return select.from_(query.subquery(copy=False), copy=False)
+        return None
+
+    def _parse_pipe_syntax_where(self, query: exp.Query) -> t.Optional[exp.Query]:
+        where = self._parse_where()
+        if isinstance(where, exp.Where) and isinstance(query, exp.Select):
+            return query.where(where.this, copy=False)
+        return None
```
georgesittas, line 1134: - Let's instead return `query` both here and in L1128, that way you won't need an `Optional` back
- Consider:
  - Updating `where` so that when it receives a `Where` instance it uses its `this` child
  - Moving `Select.where` in `Query` so that you don't have to limit this to just `Select`. It seems like every `Query` subclass supports the `where` child, so this should be a low lift.

tests/dialects/test_dialect.py
```diff
@@ -3453,3 +3453,27 @@ def test_integer_hex_strings(self):
...
+        self.validate_identity(
+            "FROM x |> SELECT x1 + 1 as x1_a, x2 - 1 as x2_a |> WHERE x1_a > 1",
+            "SELECT x1 + 1 AS x1_a, x2 - 1 AS x2_a FROM (SELECT * FROM x) WHERE x1_a > 1",
+        )
+        self.validate_identity(
+            "FROM x |> SELECT x1 + 1 as x1_a, x2 - 1 as x2_a |> WHERE x1_a > 1 |> SELECT x2_a",
+            "SELECT x2_a FROM (SELECT x1 + 1 AS x1_a, x2 - 1 AS x2_a FROM (SELECT * FROM x) WHERE x1_a > 1)",
+        )
+        self.validate_identity(
+            "FROM x |> WHERE x1 > 1 |> WHERE x2 > 2 |> SELECT x1 as gt1, x2 as gt2",
```
georgesittas, line 3473: Let's also add a test for `WHERE x1 > 1 OR x1 <> 0` followed by the 2nd `WHERE`, to demonstrate how disjunction and conjunction are handled when combined.

sqlglot/parser.py
georgesittas: Note: the return type being `Query` is fine. When we add support for `UNION`, etc, we could end up with a `Query` instance that is not a `Select`.

georgesittas (approved):
Thanks for addressing the previous comments! Just a few last ones on my end and it should be good to go.

sqlglot/parser.py
geooo109: `Delete` and `Update` implement `where` but they don't inherit from `Query`. should they ?

georgesittas: Nope, they shouldn't.

PR #5133, opened 2025-05-28
Fix(athena): Handle transpilation of FileFormatProperty from dialects that treat it as a variable and not a string literal

erindru:
Reimplementation of https://github.com/tobymao/sqlglot/pull/5125

Prior to this PR, staying within the Athena dialect meant that parsing / generating `exp.FileFormatProperty` worked as expected because `this` was always an `exp.Literal.string`.

However, if transpiling from another dialect where `this` is created as an `exp.Var` instead (since some dialects have the file format as an actual variable/identfier and not a string literal), we would output the `exp.Var` instead of converting it to a string literal.

Before:
```
>>> exp.FileFormatProperty(this=exp.var("parquet")).sql(dialect="athena")
"format=parquet"
```

This PR converts it a string literal regardless:
```
>>> exp.FileFormatProperty(this=exp.var("parquet")).sql(dialect="athena")
"format='parquet'"
```

Note that we still need to override the inherited `exp.FileFormatProperty` transform because it uppercases the property name but Athena needs the property name to be lowercase


georgesittas:
This looks ok, but perhaps we can clean it up even more if Presto and Trino are ok with `'parquet'` ([comment](https://github.com/tobymao/sqlglot/pull/5125#issuecomment-2917682627)).

PR #5137, opened 2025-05-29
fix(snowflake): Properly transpile ARRAY_AGG, IGNORE/RESPECT NULLS

VaggelisD:
Notes for Snowflake's `ARRAY_AGG`:
- It does not support an inline `ORDER BY`, it must be moved to `WITHIN GROUP (<orderby_clause>)`
- It does not support  `IGNORE NULLS` / `RESPECT NULLS`, the nulls are always excluded. This looks like an exception for some other window functions like `LEAD` or `LAG` which do allow them 

Docs
---------
[BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#array_agg) | [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/array_agg)

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -667,6 +667,8 @@ class Generator(metaclass=_Generator):
     # Expressions that need to have all CTEs under them bubbled up to them
     EXPRESSIONS_WITHOUT_NESTED_CTES: t.Set[t.Type[exp.Expression]] = set()
 
+    RESPECT_IGNORE_NULLS_UNSUPPORTED_EXPRESSIONS: t.Tuple[t.Type[exp.Expression], ...] = ()
+
     SENTINEL_LINE_BREAK = "__SQLGLOT__LB__"
 
     __slots__ = (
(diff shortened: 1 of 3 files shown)


sqlglot/generator.py
```diff
@@ -4268,6 +4270,14 @@ def _simplify_unless_literal(self, expression: E) -> E:
         return expression
 
     def _embed_ignore_nulls(self, expression: exp.IgnoreNulls | exp.RespectNulls, text: str) -> str:
+        this = expression.this
+        if isinstance(this, self.RESPECT_IGNORE_NULLS_UNSUPPORTED_EXPRESSIONS):
+            this_name = this.sql_name() if isinstance(this, exp.Func) else this.__class__
```
georgesittas, line 4275: Nit: relying on `sql_name()` may lead to printing a function name that is not found in the input, since we sometimes abstract away and use SQLGlot-specific expression types.

Should we instead print `type(expression).key`? Or `split` on `"("` and use the first segment as the function's name.

georgesittas (approved):

PR #5140, opened 2025-05-29
feat(snowflake): Transpile BQ's `ARRAY(SELECT AS STRUCT ...)`

VaggelisD:
Docs
-------
[BQ SELECT AS STRUCT](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_as_struct) | [Snowflake OBJECT_CONSTRUCT](https://docs.snowflake.com/en/sql-reference/functions/object_construct)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1017,7 +1017,6 @@ class Snowflake(Dialect):
             exp.ApproxDistinct: rename_func("APPROX_COUNT_DISTINCT"),
             exp.ArgMax: rename_func("MAX_BY"),
             exp.ArgMin: rename_func("MIN_BY"),
-            exp.Array: inline_array_sql,
             exp.ArrayConcat: lambda self, e: self.arrayconcat_sql(e, name="ARRAY_CAT"),
             exp.ArrayContains: lambda self, e: self.func("ARRAY_CONTAINS", e.expression, e.this),
             exp.AtTimeZone: lambda self, e: self.func(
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -1427,3 +1426,36 @@ def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
...
             return expr_sql
+
+        def array_sql(self, expression: exp.Array) -> str:
+            expressions = expression.expressions
+
+            first_expr = seq_get(expressions, 0)
+            if isinstance(first_expr, exp.Select):
+                # SELECT AS STRUCT foo AS alias_foo -> ARRAY_AGG(OBJECT_CONSTRUCT('alias_foo', foo))
+                kind = first_expr.args.get("kind")
+                if kind and kind.upper() == "STRUCT":
```
georgesittas, line 1437:
```suggestion
                if first_expr.text("kind").upper() == "STRUCT":
```

sqlglot/dialects/snowflake.py
```diff
@@ -1427,3 +1426,36 @@ def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
...
+                    object_construct_args = []
+                    for expr in first_expr.expressions:
+                        if isinstance(expr, exp.Alias):
+                            # SELECT AS STRUCT foo AS alias_foo -> OBJECT_CONSTRUCT('alias_foo', foo)
+                            name = expr.this.name
+                            alias = f"'{expr.alias}'"
```
georgesittas, line 1443: Let's construct a proper `Literal.string(...)` here instead of hardcoding the string?

sqlglot/dialects/snowflake.py
```diff
@@ -1427,3 +1426,36 @@ def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
...
+                        else:
+                            # SELECT AS STRUCT foo -> OBJECT_CONSTRUCT('foo', foo)
+                            name = expr.name
+                            alias = f"'{name}'"
+
+                        object_construct_args.extend([alias, name])
```
georgesittas, line 1449: Do we preserve comments here?

sqlglot/dialects/snowflake.py
```diff
@@ -1427,3 +1426,36 @@ def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
...
+
+                    array_agg = exp.ArrayAgg(
+                        this=_build_object_construct(args=object_construct_args)
+                    )
+                    select = exp.select(array_agg, dialect=self.dialect)
```
georgesittas, line 1454: Do we need a dialect here?

sqlglot/dialects/snowflake.py
```diff
@@ -1427,3 +1426,36 @@ def arrayagg_sql(self, expression: exp.ArrayAgg) -> str:
...
+                    from_ = first_expr.args.get("from")
+                    if from_:
+                        select = select.from_(from_, dialect=self.dialect)
```
georgesittas, line 1457: What about other modifiers?

georgesittas: Yes, that seems safer, otherwise you could silently drop query modifiers in the future, etc. May also wanna double-check whether `kind` needs to be switched off for Snowflake, I don't recall it being supported, but may be wrong.

sqlglot/dialects/snowflake.py
VaggelisD: not needed anymore

georgesittas (approved):
Nice 💯

Issue #5129, opened 2025-05-28
ALTER TABLE (generator.py > alter_sql()) checks only the first action. Multiple actions, i.e. ColumnDef and AddConstraint do not work.


Mevrael:
**EDIT (George):** the issue is shown below:

```python
>>> import sqlglot
>>> sqlglot.transpile("ALTER TABLE foo ADD COLUMN id BIGINT NOT NULL PRIMARY KEY DEFAULT 1, ADD CONSTRAINT fk_orders_user FOREIGN KEY (id) REFERENCES foo(id)", "postgres")
['ALTER TABLE foo ADD COLUMN id BIGINT NOT NULL PRIMARY KEY DEFAULT 1, ADD COLUMN CONSTRAINT fk_orders_user FOREIGN KEY (id) REFERENCES foo (id)']
```

The input is a syntactically valid Postgres statement.

# EXAMPLE

For example, this code shall alter a table with 2 actions: ColumnDef and AddConstraint

```python
import sqlglot.expressions as exp

sql_exp = exp.Alter(
    this=exp.Table(this='new_table_name'),
    kind='TABLE',
    actions=[
        exp.ColumnDef(
            this=exp.Column(this='new_col_name'),
            kind=exp.DataType(this=exp.DataType.Type.UBIGINT),
        ),
        exp.AddConstraint(
            expressions=[
                exp.Constraint(
                    this=exp.Identifier(this='new_index_name'),
                    expressions=[
                        exp.UniqueColumnConstraint(
                            this=exp.Schema(
                                expressions=[
                                    exp.Identifier(this='new_col_name')
                                ]
                            )
                        )
                    ]
                )
            ]
        )

    ]
)

print(sql_exp.sql(pretty=True))
```

However it produces this SQL and adds ADD COLUMN in front of ADD CONSTRAINT:

```
ALTER TABLE new_table_name ADD COLUMN new_col_name UBIGINT, ADD COLUMN ADD   CONSTRAINT new_index_name UNIQUE (
    new_col_name
  )
```

It shouldn't be there.

---

# SOURCE CODE

There is a serious issue is in generator.py > alter_sql() with the algorithm.

https://github.com/tobymao/sqlglot/blob/main/sqlglot/generator.py#L3433-L3466

```python
    def alter_sql(self, expression: exp.Alter) -> str:
        actions = expression.args["actions"]

        if isinstance(actions[0], exp.ColumnDef):
            actions = self.add_column_sql(expression)
[...]

PR #5144, opened 2025-05-30
Fix!: Refactor ALTER TABLE ADD parsing

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5129

This PR refactors `ALTER TABLE ADD ..., ...` parsing by simplifying both the parsing and the generation logic. For instance, Oracle supports this statement:

```SQL
ALTER TABLE Payments ADD (Stock NUMBER NOT NULL, dropid VARCHAR2(500) NOT NULL)
```

Before, this was parsed through `self._parse_wrapped_csv(self._parse_field_def, optional=True)` and generated back by reintroducing the parenthesis; Now, this is simply parsed as `parse_schema` and generated as-is.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -3435,16 +3435,24 @@ class Generator(metaclass=_Generator):
     def alter_sql(self, expression: exp.Alter) -> str:
         actions = expression.args["actions"]
 
-        if isinstance(actions[0], exp.ColumnDef):
-            actions = self.add_column_sql(expression)
-        elif isinstance(actions[0], exp.Schema):
-            actions = self.expressions(expression, key="actions", prefix="ADD COLUMNS ")
-        elif isinstance(actions[0], exp.Delete):
-            actions = self.expressions(expression, key="actions", flat=True)
-        elif isinstance(actions[0], exp.Query):
-            actions = "AS " + self.expressions(expression, key="actions")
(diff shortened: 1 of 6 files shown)


sqlglot/generator.py
```diff
@@ -3435,16 +3435,24 @@ def alterset_sql(self, expression: exp.AlterSet) -> str:
...
-            actions = self.add_column_sql(expression)
-        elif isinstance(actions[0], exp.Schema):
-            actions = self.expressions(expression, key="actions", prefix="ADD COLUMNS ")
-        elif isinstance(actions[0], exp.Delete):
-            actions = self.expressions(expression, key="actions", flat=True)
-        elif isinstance(actions[0], exp.Query):
-            actions = "AS " + self.expressions(expression, key="actions")
+        if not self.dialect.ALTER_TABLE_ADD_REQUIRED_FOR_EACH_COLUMN and isinstance(
+            actions[0], exp.ColumnDef
+        ):
```
georgesittas, line 3440: Could you have something like `ADD foo, bar, ADD CONSTRAINT`?

georgesittas (approved):
Well done

sqlglot/generator.py
georgesittas: Sounds good, thank you for checking

PR #5146, opened 2025-05-30
chore: update benchmarking framework to pyperf

benfdking:
- move benchmarks to pyperf to give us lots of flexibility 
- add an exercising of the benchmark to ci just to be sure there are no regressions
- allows lots of features like profiling or customizing what you are testing and number of times

-> Looks like lineage can be optimized so want to use this as the foundation to investigate.


benchmarks/optimize.py
```diff
@@ -16,55 +17,59 @@ def gen_condition(n):
...
+def optimize_queries(expressions, schema):
+    for e in expressions:
+        optimize(e, schema)
+
+
+def run_benchmarks():
+    runner = pyperf.Runner()
+
+    # Define benchmarks with their setup functions
+    benchmarks = {
```
VaggelisD, line 57: it looks to me like the previous `BENCHMARK` dict was easier to extend, we'll now have to update both the functions and the dict ?

georgesittas (approved):
Great improvement, thanks for taking this on!

It should be a [low lift](https://chatgpt.com/share/683a9a3c-17a8-8010-953c-90394dad3e8b) to print a table comparing the PR to main, like what you did for the Rust tokenizer. I tried it out locally and it works like a charm:

Screenshot 2025-05-31 at 9 00 48 AM

Happy to do it myself in a followup PR if you want.

PR #5147, opened 2025-05-31
Fix(hive): STRUCT column names and data type should be separated by ':'

tsamaras:
According to the [Hive Language Manual](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27838462#LanguageManualTypes-ComplexTypes) srtuct column names and data types should be separate by ':' in DDLs. Since the spark2 dialect inherits from hive and spark2 had the correct implementations I moved the implementation into the HiveGenerator and adjusted the test cases accordingly.

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -741,6 +741,17 @@ class Hive(Dialect):
 
             return self.func("STRUCT", *values)
 
+        def columndef_sql(self, expression: exp.ColumnDef, sep: str = " ") -> str:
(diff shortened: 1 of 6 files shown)


georgesittas (approved):
Thank you for the contribution! I've verified locally that Hive treats the omission of `:` as a syntax error. Good catch.

waldoppper:
Thanks both. When would you expect a release cut with this fix?

georgesittas:
This has been released w/ 26.25.0.

PR #5157, opened 2025-06-02
feat(tsql): add convert style 126

pa1ch:


georgesittas:
I'll go ahead and add one.

PR #5167, opened 2025-06-03
feat(tsql): support FOR XML syntax

### LINKED ISSUE #5161: Please add support for TSQL FOR XML PATH
Setup on TSQL:

```
CREATE TABLE dbo.Demo ([Id] INT IDENTITY PRIMARY KEY, [Strings] NVARCHAR(128))
;
INSERT	dbo.Demo (Strings)
VALUES	('A'),
		('B'),
		('C')
;
```

Query to support:

```
SELECT	',' + Strings
FROM	dbo.Demo
FOR XML PATH ('')
;
/* Returns ,A,B,C */
```

This is commonly used for string concatenation/aggregation especially in combination with STUFF to strip the leading comma:

```
SELECT  STUFF((
            SELECT ',' + Strings
            FROM dbo.Demo
            FOR XML PATH('')
            ), 1, 1, '')
;
/* Returns A,B,C */
```

SQLGlot:

```
import sqlglot
sqlglot.parse_one("SELECT ',' + Strings FROM dbo.Demo FOR XML PATH ('');", read="tsql")
```

Returns:

```
  File "\.env\Lib\site-packages\sqlglot\__init__.py", line 139, in parse_one
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\.env\Lib\site-packages\sqlglot\dialects\dialect.py", line 994, in parse
  File "\.env\Lib\site-packages\sqlglot\parser.py", line 1558, in parse
    return self._parse(
           ^^^^^^^^^^^^
  File "\.env\Lib\site-packages\sqlglot\parser.py", line 1630, in _parse
  File "\.env\Lib\site-packages\sqlglot\parser.py", line 1671, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 38.
  SELECT ',' + Strings FROM dbo.Demo FOR XML PATH ('');
```

Syntax reference: https://learn.microsoft.com/en-us/sql/relational-databases/xml/basic-syntax-of-the-for-xml-clause?view=sql-server-ver17

geooo109:
Fixes #5161 

**DOCS**
[T-SQL FOR XML](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-for-clause-transact-sql?view=sql-server-ver17#syntax)

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -102,6 +102,23 @@ OPTIONS: parser.OPTIONS_TYPE = {
     "USE": ("PLAN",),
 }
 
+
+XML_OPTIONS: parser.OPTIONS_TYPE = {
+    **dict.fromkeys(
+        (
+            "AUTO" "EXPLICIT",
+            "TYPE",
+        ),
+        tuple(),
+    ),
+    "ELEMENTS": (
+        "XSINIL",
+        "ABSENT",
+    ),
+    "BINARY": ("BASE64",),
+}
+
+
 OPTIONS_THAT_REQUIRE_EQUAL = ("MAX_GRANT_PERCENT", "MIN_GRANT_PERCENT", "LABEL")
 
 
@@ -548,6 +565,7 @@ class TSQL(Dialect):
         QUERY_MODIFIER_PARSERS = {
             **parser.Parser.QUERY_MODIFIER_PARSERS,
             TokenType.OPTION: lambda self: ("options", self._parse_options()),
+            TokenType.FOR: lambda self: ("for", self._parse_for()),
         }
 
         # T-SQL does not allow BEGIN to be used as an identifier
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -6972,6 +6972,10 @@ class XMLNamespace(Expression):
     pass
 
 
+class XMLKeyValueOption(Expression):
+    arg_types = {"this": True, "expression": False}
```
georgesittas, line 6977: Nit: may be worth linking the T-SQL docs here.

sqlglot/dialects/tsql.py
```diff
@@ -674,6 +692,28 @@ def _parse_option() -> t.Optional[exp.Expression]:
...
+            this = self._parse_primary_or_var()
+            if self._match(TokenType.L_PAREN, advance=False):
+                expression = self._parse_wrapped(self._parse_string)
+            else:
+                expression = None
+
+            return exp.XMLKeyValueOption(this=this, expression=expression)
+
+        def _parse_for(self) -> t.Optional[t.List[exp.Expression]]:
+            if not (self._match(TokenType.FOR) and self._match(TokenType.XML)):
```
georgesittas: I'm assuming this will get updated to handle `JSON` similarly, but FYI you can do this instead:
```suggestion
            if not self._match_pair(TokenType.FOR, TokenType.XML):
```

sqlglot/generator.py
```diff
@@ -2604,13 +2604,18 @@ def query_modifiers(self, expression: exp.Expression, *sqls: str) -> str:
...
             sep="",
         )
 
     def options_modifier(self, expression: exp.Expression) -> str:
         options = self.expressions(expression, key="options")
         return f" {options}" if options else ""
 
+    def for_modifiers(self, expression: exp.Expression) -> str:
+        for_modifiers = self.expressions(expression, key="for")
+        return f" FOR XML {for_modifiers}" if for_modifiers else ""
```
georgesittas: Let's try the `pretty=True` flag to see how it formats queries that contain a `FOR` clause.

sqlglot/dialects/tsql.py
geooo109: Yep, right. I had this in mind, but I agree.

sqlglot/generator.py
geooo109:
```
Input:
 SELECT * FROM t FOR XML PATH, BINARY BASE64, ELEMENTS XSINIL

Output:
SELECT
  *
FROM t FOR XML   PATH,
  BINARY BASE64,
  ELEMENTS XSINIL
```

Needs a fix right ?

georgesittas: Correct- check out how other modifiers are handled and let's try to fix this.

PR #5171, opened 2025-06-04
feat(parser): AGGREGATE with GROUP AND ORDER BY pipe syntax

geooo109:
Add support for the `AGGREGATE` pipe syntax.

In the case of BigQuery the `ASC` and `DESC` have been removed from the `ID_VAR_TOKENS` because they can't be used as identifiers. Moreover, in BigQuery pipe syntax when we find `ASC` or `DESC` it means that the query needs an `ORDE BY`.

**DOCS**
[BigQuery AGGREGATE Syntax](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#aggregate_pipe_operator)
[Databricks AGGREGATE Syntax](en/sql/language-manual/sql-ref-syntax-qry-select-pipeop)

| `ORDER BY` | ✅  |
| `UNION`               |           |
| `INTERSECT`               |           |
| `EXCEPT`               |           |

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -525,6 +525,16 @@ class BigQuery(Dialect):
         LOG_DEFAULTS_TO_LN = True
         SUPPORTS_IMPLICIT_UNNEST = True
 
+        # BigQuery does not allow ASC/DESC to be used as an identifier
+        ID_VAR_TOKENS = parser.Parser.ID_VAR_TOKENS - {TokenType.ASC, TokenType.DESC}
+        ALIAS_TOKENS = parser.Parser.ALIAS_TOKENS - {TokenType.ASC, TokenType.DESC}
+        TABLE_ALIAS_TOKENS = parser.Parser.TABLE_ALIAS_TOKENS - {TokenType.ASC, TokenType.DESC}
+        COMMENT_TABLE_ALIAS_TOKENS = parser.Parser.COMMENT_TABLE_ALIAS_TOKENS - {
+            TokenType.ASC,
+            TokenType.DESC,
+        }
+        UPDATE_ALIAS_TOKENS = parser.Parser.UPDATE_ALIAS_TOKENS - {TokenType.ASC, TokenType.DESC}
+
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "CONTAINS_SUBSTR": _build_contains_substring,
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1194,7 +1194,9 @@ class Parser(metaclass=_Parser):
         self._match_text_seq("AGGREGATE")
         query = self._parse_pipe_syntax_aggregate_group_order_by(query, group_by_exists=False)
 
-        if self._match(TokenType.GROUP_BY) or self._match_text_seq("GROUP", "AND", "ORDER BY"):
+        if self._match(TokenType.GROUP_BY) or (
+            self._match_text_seq("GROUP", "AND") and self._match(TokenType.ORDER_BY)
+        ):
             return self._parse_pipe_syntax_aggregate_group_order_by(query)
 
         return query
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1157,8 +1157,8 @@ class Parser(metaclass=_Parser):
         return this
 
     def _parse_pipe_syntax_aggregate_group_order_by(
-        self, query: exp.Select, group_by_exists: bool = True
-    ) -> exp.Select:
+        self, query: exp.Query, group_by_exists: bool = True
+    ) -> exp.Query:
         expr = self._parse_csv(self._parse_pipe_syntax_aggregate_fields)
         aggregates_or_groups, orders = [], []
         for element in expr:
@@ -1176,7 +1176,7 @@ class Parser(metaclass=_Parser):
             else:
                 aggregates_or_groups.append(element)
 
-        if group_by_exists:
+        if group_by_exists and isinstance(query, exp.Select):
             query = query.select(*aggregates_or_groups, copy=False).group_by(
                 *[element.alias_or_name for element in aggregates_or_groups], copy=False
             )
@@ -1190,7 +1190,7 @@ class Parser(metaclass=_Parser):
 
         return query
 
-    def _parse_pipe_syntax_aggregate(self, query: exp.Select) -> exp.Query:
+    def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
         self._match_text_seq("AGGREGATE")
         query = self._parse_pipe_syntax_aggregate_group_order_by(query, group_by_exists=False)
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -936,6 +936,7 @@ class Parser(metaclass=_Parser):
         "ORDER BY": lambda self, query: query.order_by(self._parse_order(), copy=False),
         "LIMIT": lambda self, query: self._parse_pipe_syntax_limit(query),
         "OFFSET": lambda self, query: query.offset(self._parse_offset(), copy=False),
+        "AGGREGATE": lambda self, query: self._parse_pipe_syntax_aggregate(query),
     }
 
     PROPERTY_PARSERS: t.Dict[str, t.Callable] = {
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -525,6 +525,16 @@ class Parser(parser.Parser):
         LOG_DEFAULTS_TO_LN = True
         SUPPORTS_IMPLICIT_UNNEST = True
 
+        # BigQuery does not allow ASC/DESC to be used as an identifier
```
georgesittas, line 528: Note: we should refactor this section by leveraging the `_Parser` metaclass so that class-scoped sets are not just inherited, but instead re-built, reflecting the local changes to things like `ID_VAR_TOKENS`.

Ideally, we should only need to modify `ID_VAR_TOKENS` here.

tests/dialects/test_dialect.py
```diff
@@ -3528,3 +3528,50 @@ def test_pipe_syntax(self):
...
+            "FROM x |> AGGREGATE SUM(x1), MAX(x2), MIN(x3) GROUP BY x4 AS a_x4, x5 AS a_x5",
+            "SELECT SUM(x1), MAX(x2), MIN(x3), x4 AS a_x4, x5 AS a_x5 FROM (SELECT * FROM x) GROUP BY a_x4, a_x5",
+        )
+        self.validate_identity(
+            "FROM x |> AGGREGATE SUM(x1) as s_x1 GROUP BY x1 |> SELECT s_x1, x1 as ss_x1",
+            "SELECT s_x1, x1 AS ss_x1 FROM (SELECT SUM(x1) AS s_x1, x1 FROM (SELECT * FROM x) GROUP BY x1)",
+        )
+        self.validate_identity(
+            "FROM x |> AGGREGATE SUM(x1) GROUP", "SELECT SUM(x1) AS GROUP FROM (SELECT * FROM x)"
+        )
```
georgesittas, line 3553: Is this valid bigquery?

tests/dialects/test_dialect.py
```diff
@@ -3528,3 +3528,50 @@ def test_pipe_syntax(self):
...
+
+        for order_option in ("ASC", "DESC", "ASC NULLS LAST", "DESC NULLS FIRST"):
+            with self.subTest(f"Testing pipe syntax AGGREGATE for order option: {order_option}"):
+                self.validate_all(
+                    f"SELECT SUM(x1) AS x_s FROM (SELECT * FROM x) ORDER BY x_s {order_option}",
+                    read={
+                        "bigquery": f"FROM x |> AGGREGATE SUM(x1) AS x_s {order_option}",
```
georgesittas, line 3560: Is this valid bigquery? Based on their [syntax spec](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#aggregate_pipe_operator), `ASC` and `DESC` are only relevant to the third clause, i.e.:

```
-- Aggregation with grouping and shorthand ordering syntax
|> AGGREGATE [aggregate_expression [[AS] alias] [order_suffix] [, ...]]
   GROUP [AND ORDER] BY groupable_item [[AS] alias] [order_suffix] [, ...]

order_suffix: {ASC | DESC} [{NULLS FIRST | NULLS LAST}]
```

tests/dialects/test_dialect.py
```diff
@@ -3528,3 +3528,50 @@ def test_pipe_syntax(self):
...
+            "SELECT SUM(x1), MAX(x2), MIN(x3) FROM (SELECT * FROM x)",
+        )
+        self.validate_identity(
+            "FROM x |> AGGREGATE SUM(x1) AS s_x1 |> SELECT s_x1",
+            "SELECT s_x1 FROM (SELECT SUM(x1) AS s_x1 FROM (SELECT * FROM x))",
+        )
+        self.validate_identity(
+            "FROM x |> AGGREGATE SUM(x1), MAX(x2), MIN(x3) GROUP BY x4, x5",
+            "SELECT SUM(x1), MAX(x2), MIN(x3), x4, x5 FROM (SELECT * FROM x) GROUP BY x4, x5",
+        )
```
georgesittas, line 3542: What if you `GROUP BY` twice in a row, using two different `|>`?

sqlglot/parser.py
```diff
@@ -1143,6 +1144,63 @@ def _parse_pipe_syntax_limit(self, query: exp.Query) -> exp.Query:
...
+            if isinstance(element, exp.Ordered):
+                this = element.this
+                orders.append(
+                    exp.Ordered(
+                        this=this.alias_or_name,
+                        desc=element.args.get("desc"),
+                        nulls_first=element.args.get("nulls_first"),
+                        with_fill=element.args.get("with_fill"),
+                    )
+                )
```
georgesittas: Why are we manually constructing the `Ordered` here? This looks brittle.

sqlglot/parser.py
```diff
@@ -1143,6 +1144,63 @@ def _parse_pipe_syntax_limit(self, query: exp.Query) -> exp.Query:
...
+                aggregates_or_groups.append(this)
+            else:
+                aggregates_or_groups.append(element)
+
+        if group_by_exists and isinstance(query, exp.Select):
+            query = query.select(*aggregates_or_groups, copy=False).group_by(
+                *[element.alias_or_name for element in aggregates_or_groups], copy=False
```
georgesittas: Why is this doing an `alias_or_name`? This could inadvertently get rid of quotes. Don't we need to copy the actual `Identifier` instances?

sqlglot/parser.py
```diff
@@ -1143,6 +1144,63 @@ def _parse_pipe_syntax_limit(self, query: exp.Query) -> exp.Query:
...
+            )
+        else:
+            query = exp.select(*aggregates_or_groups, copy=False).from_(
+                query.subquery(copy=False), copy=False
+            )
```
georgesittas, line 1182: Why do we need to construct a new subquery here?

tests/dialects/test_dialect.py
geooo109: Valid in dbx.

sqlglot/dialects/bigquery.py
geooo109: When I do the refactor, I will link it here (leave the comment open until then)

tests/dialects/test_dialect.py
geooo109: Yep, bigquery can run this. (although is noop in this case)

tests/dialects/test_dialect.py
geooo109: We currently support this. 
```
Input:
FROM x |> AGGREGATE SUM(a) AS x_s GROUP BY a |>  AGGREGATE SUM(a) AS x_s GROUP BY a

Output:
SELECT SUM(a) AS x_s, a FROM (SELECT SUM(a) AS x_s, a FROM (SELECT * FROM x) GROUP BY a) GROUP BY a
```

sqlglot/parser.py
geooo109: yep, you are right, I will update this.

sqlglot/parser.py
geooo109: Yep, my bad, will refactor.

sqlglot/parser.py
geooo109:
```
FROM x |> AGGREGATE MIN(a) min_a, MAX(b) max_b  GROUP BY a, b |> AGGREGATE MIN(a) min2_a GROUP BY a;
```
The first `AGGREGATE` is treated as a subquery of the second `AGGREGATE` that selects from, the result will be the table with `min2_a` and `a` column.

sqlglot/parser.py
```diff
@@ -1143,6 +1144,59 @@ def _parse_pipe_syntax_limit(self, query: exp.Query) -> exp.Query:
             query.offset(offset, copy=False)
         return query
 
+    def _parse_pipe_syntax_aggregate_fields(self) -> t.Optional[exp.Expression]:
+        this = self._parse_assignment()
+        if self._match_text_seq("GROUP", "AND", advance=False):
+            return this
```
georgesittas, line 1150: What if it's just `GROUP` and not `GROUP AND ORDER`? Won't you consume `GROUP` as an alias below then?

georgesittas (approved):

sqlglot/parser.py
geooo109: If it's a `GROUP BY` we have the token -> correct handling
If it's just a `GROUP` we treat it as an `ALIAS` for non BigQuery dialects (like dbx), which is supported. For BigQuery this isn't valid in frist palce -> correct handling

PR #5179, opened 2025-06-06
feat(postgres)!: Add support for ANY_VALUE for versions 16+

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/4674

Postgres 16 (released 24 Sep 2023) added support for `ANY_VALUE`. As this may be too recent for some Postgres workloads, I thought it'd make for another use case of `Version`, but if we want to simplify we can instead always preserve the roundtrip.

Docs
---------
[PSQL ANY_VALUE](https://www.postgresql.org/docs/16/functions-aggregate.html) | [PSQL Feature Matrix](https://www.postgresql.org/about/featurematrix/)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -36,6 +36,7 @@ from sqlglot.dialects.dialect import (
     strposition_sql,
     count_if_to_sum,
     groupconcat_sql,
+    Version,
 )
 from sqlglot.generator import unsupported_args
 from sqlglot.helper import is_int, seq_get
@@ -253,6 +254,15 @@ def _levenshtein_sql(self: Postgres.Generator, expression: exp.Levenshtein) -> s
     name = "LEVENSHTEIN_LESS_EQUAL" if expression.args.get("max_dist") else "LEVENSHTEIN"
 
     return rename_func(name)(self, expression)
+
+
+def _versioned_anyvalue_sql(self: Postgres.Generator, expression: exp.AnyValue) -> str:
+    # https://www.postgresql.org/docs/16/functions-aggregate.html
+    # https://www.postgresql.org/about/featurematrix/
+    if self.dialect.version < Version("16.0"):
+        return any_value_to_max_sql(self, expression)
+
+    return rename_func("ANY_VALUE")(self, expression)
 
 
 class Postgres(Dialect):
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/postgres.py
```diff
@@ -255,6 +256,15 @@ def _levenshtein_sql(self: Postgres.Generator, expression: exp.Levenshtein) -> s
     return rename_func(name)(self, expression)
 
 
+def _versioned_anyvalue_sql(self: Postgres.Generator, expression: exp.AnyValue) -> str:
```
georgesittas, line 259: Nit: I suggest naming this `anyvalue_sql` and moving it under postgres' generator. How does this affect Redshift? For reference: https://docs.aws.amazon.com/redshift/latest/dg/r_ANY_VALUE.html

tests/dialects/test_postgres.py
```diff
@@ -1436,3 +1436,21 @@ def test_json_extract(self):
...
+        )
+
+        self.validate_all(
+            "SELECT ANY_VALUE(1) AS col",
+            write={
+                "postgres": "SELECT ANY_VALUE(1) AS col",
+                "postgres, version=16": "SELECT ANY_VALUE(1) AS col",
+                "postgres, version=17.5": "SELECT ANY_VALUE(1) AS col",
+            },
+        )
```
georgesittas: Nit: we can consolidate these in a single `validate_all` and move it under `test_postgres`.

sqlglot/dialects/postgres.py
VaggelisD: Redshift pops `exp.AnyValue` from `TRANSFORMS` so it falls back to the base generator, this was already in place before. 

This is an interesting scenario for versions & inheritance; Postgres's `versioned_anyvalue_sql` would fail in Redshift because `self.dialect.version` would not match

VaggelisD: This is also the reason I maintained the `TRANSFORMS` entry in Postgres through the helper function

tests/dialects/test_postgres.py
VaggelisD: Yeah wanted to mostly separate the two semantically, will do

sqlglot/dialects/postgres.py
georgesittas: Sounds good.

PR #5181, opened 2025-06-06
docs: used by dlt

zilto:
We're using sqlglot more and more at [dlt-hub/dlt](https://github.com/dlt-hub/dlt) and we're super impressed by the tool :)


georgesittas:
Awesome :)

PR #5184, opened 2025-06-06
feat(parser): SET OPERATORS with pipe syntax

geooo109:
Add support for the SET OPERATORS = `{UNION, INTERSECT, EXCEPT}` pipe syntax.

**DOCS**
[BigQuery pipe syntax UNION](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#union_pipe_operator)
[BigQuery pipe syntax INTERSECT](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#intersect_pipe_operator)
[BigQuery pipe syntax EXCEPT](https://cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax#except_pipe_operator)

| `ORDER BY` | ✅  |

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1197,17 +1197,18 @@ class Parser(metaclass=_Parser):
 
         return query
 
-    def _parse_pipe_syntax_set_operator(self, query: exp.Query) -> exp.Query:
+    def _parse_pipe_syntax_set_operator(
+        self, query: t.Optional[exp.Query]
+    ) -> t.Optional[exp.Query]:
         first_setop = self.parse_set_operation(this=query)
 
-        # If this condition is True a JOIN parser must be invoked
-        if not first_setop:
-            return query
+        if not first_setop or not query:
+            return None
 
-        first_setop.args.pop("this")
+        first_setop.this.pop()
         distinct = first_setop.args.pop("distinct")
 
-        setops = [first_setop.args.pop("expression"), *self._parse_expressions()]
+        setops = [first_setop.expression.pop(), *self._parse_expressions()]
 
         if isinstance(first_setop, exp.Union):
             return query.union(*setops, distinct=distinct, **first_setop.args)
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -7217,11 +7235,11 @@ def _parse_tokens(
 
         return this
 
-    def _parse_pipe_syntax_query(self, query: exp.Select) -> exp.Query:
+    def _parse_pipe_syntax_query(self, query: exp.Query) -> exp.Query:
         while self._match(TokenType.PIPE_GT):
             parser = self.PIPE_SYNTAX_TRANSFORM_PARSERS.get(self._curr.text.upper())
             if not parser:
-                self.raise_error(f"Unsupported pipe syntax operator: '{self._curr.text.upper()}'.")
+                query = self._parse_pipe_syntax_set_operator(query)
```
georgesittas, line 7244: Why was the `raise_error` call removed?

sqlglot/parser.py
```diff
@@ -1197,6 +1197,24 @@ def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
...
         return query
 
+    def _parse_pipe_syntax_set_operator(self, query: exp.Query) -> exp.Query:
+        first_setop = self.parse_set_operation(this=query)
+
+        # If this condition is True a JOIN parser must be invoked
+        if not first_setop:
+            return query
+
+        first_setop.args.pop("this")
```
georgesittas: Avoid popping from `args` directly, that doesn't handle stuff like rearranging pointers in the AST.
```suggestion
        first_setop.this.pop()
```

sqlglot/parser.py
```diff
@@ -1197,6 +1197,24 @@ def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
...
+        distinct = first_setop.args.pop("distinct")
```
georgesittas, line 1209:
```suggestion
        distinct = first_setop.pop("distinct")
```

sqlglot/parser.py
```diff
@@ -1197,6 +1197,24 @@ def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
...
+
+        setops = [first_setop.args.pop("expression"), *self._parse_expressions()]
```
georgesittas: Ditto.

sqlglot/parser.py
```diff
@@ -1197,6 +1197,24 @@ def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
 
         return query
 
+    def _parse_pipe_syntax_set_operator(self, query: exp.Query) -> exp.Query:
```
georgesittas: Let's discuss this implementation, I'd like to understand the nuances involved. I feel like the current approach has some unexpected complexity.

sqlglot/parser.py
geooo109: Yep, I see, will change it.

PR #5185, opened 2025-06-07
feat(sqlite): implement RESPECT/IGNORE NULLS in first_value()

NickCrews:
If handed the duckdb/postgres/etc SQL of `SELECT FIRST_VALUE(val RESPECT NULLS) OVER ....` then we should just strip out the `RESPECT NULLS` part. If handed `IGNORE NULLS`, we should error, because sqlite doesn't support this, it always considers nulls.

This isn't a working PR I'm sure. Consider it more of a bug that I'm filing, and giving a slight head start in fixing it. I'll be curious to see what changes need to happen to actualy make this work.


georgesittas:
I'll take this to the finish line, thanks.

PR #5205, opened 2025-06-10
fix(parser)!: CTEs instead of subqueries for pipe syntax

geooo109:
This PR refactors the whole implementation of the pipe syntax and substitutes the subquery logic to CTEs. 
This makes the generated SQL more readable and also solves issues with the scoping of the query columns.

A problem with the previous implementation can be spotted in the following example:
```
Input:
FROM c.x |> SELECT x.col1, x.col2

Output:
SELECT x.col1, x.col2 FROM (SELECT * FROM c.x)
```
The subquery "shadows" the scope.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -7,7 +7,7 @@ from collections import defaultdict
 
 from sqlglot import exp
 from sqlglot.errors import ErrorLevel, ParseError, concat_messages, merge_errors
-from sqlglot.helper import apply_index_offset, ensure_list, seq_get, find_new_name
+from sqlglot.helper import apply_index_offset, ensure_list, seq_get
 from sqlglot.time import format_time
 from sqlglot.tokens import Token, Tokenizer, TokenType
 from sqlglot.trie import TrieResult, in_trie, new_trie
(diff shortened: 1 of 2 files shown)


geooo109:
For the unique name generation of CTEs should we have a state ?

sqlglot/parser.py
```diff
@@ -3337,7 +3250,9 @@ def _parse_select(
         elif from_:
             this = exp.select("*").from_(from_.this, copy=False)
```
georgesittas: Let's only construct this if there is no `PIPE_GT` token, so after L3255.

sqlglot/parser.py
```diff
@@ -8408,3 +8306,141 @@ def _identifier_expression(
...
+
+    def _build_pipe_cte(self, query: exp.Query, expressions: t.List[exp.Expression]) -> exp.Query:
+        if query.selects:
+            self._pipe_cte_counter += 1
+            new_cte = f"_pipe_cte_{self._pipe_cte_counter}"
+            ctes = (
+                query.args.pop("with", None)
+                if isinstance(query, exp.Select)
+                else query.this.args.pop("with", None)
+            )
```
georgesittas: Ditto about not using `pop` directly on `args`. Also, Not sure why we need to check for `Select` here. I'd consider doing something like this:

```python
with_ = query.args.get("with")
if with_:
    with_.pop()
```

Note that `SetOperation`s also have query modifiers, like CTEs, so you shouldn't need to unwrap `query.this`.

sqlglot/parser.py
```diff
@@ -8408,3 +8306,141 @@ def _identifier_expression(
         expression = self.expression(exp.Identifier, this=token.text, **kwargs)
         expression.update_positions(token)
         return expression
+
+    def _build_pipe_cte(self, query: exp.Query, expressions: t.List[exp.Expression]) -> exp.Query:
+        if query.selects:
+            self._pipe_cte_counter += 1
+            new_cte = f"_pipe_cte_{self._pipe_cte_counter}"
```
georgesittas: Let's use a shorter name here, e.g., `__tmpX`.

georgesittas (approved):
LGTM, just a few suggestions.

georgesittas (approved):

PR #5208, opened 2025-06-11
Feat(spark): support ALTER ADD PARTITION

### LINKED ISSUE #5204: Alter table does not handle `ADD PARTITION`
I would expect this syntax to do a roundtrip successfully:
```
>>> sqlglot.parse_one("ALTER TABLE foo ADD PARTITION(event='click')").sql()
"ALTER TABLE foo PARTITION(event = 'click')"
```
but it drops the `ADD` portion. This is valid syntax in spark (where I ran into it) so it at least needs to be added to that dialect.

georgesittas:
Fixes #5204

References:
- https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-partition#partition
- https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-alter-table.html

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -7368,13 +7368,20 @@ class Parser(metaclass=_Parser):
         )
 
     def _parse_alter_table_add(self) -> t.List[exp.Expression]:
-        def _parse_add_column_or_constraint():
+        def _parse_add_alteration() -> t.Optional[exp.Expression]:
             self._match_text_seq("ADD")
             if self._match_set(self.ADD_CONSTRAINT_TOKENS, advance=False):
                 return self.expression(
                     exp.AddConstraint, expressions=self._parse_csv(self._parse_constraint)
                 )
-            return self._parse_add_column()
+
+            is_partition = self._match(TokenType.PARTITION, advance=False)
+            field = self._parse_add_column()
+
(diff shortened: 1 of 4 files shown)


sqlglot/parser.py
```diff
@@ -7368,13 +7368,20 @@ def _parse_drop_partition(self, exists: t.Optional[bool] = None) -> exp.DropPart
...
+        def _parse_add_alteration() -> t.Optional[exp.Expression]:
             self._match_text_seq("ADD")
             if self._match_set(self.ADD_CONSTRAINT_TOKENS, advance=False):
                 return self.expression(
                     exp.AddConstraint, expressions=self._parse_csv(self._parse_constraint)
                 )
-            return self._parse_add_column()
+
+            is_partition = self._match(TokenType.PARTITION, advance=False)
+            field = self._parse_add_column()
```
VaggelisD: Looks like the docs also allow `[IF NOT EXISTS]` between `ADD` and `PARTITION` afaict, but feel free to ignore

georgesittas: Addressed [here](https://github.com/tobymao/sqlglot/pull/5208/commits/434b86cc6e932f27c499e5ef2665c831a29000a4), please take another look.

PR #5210, opened 2025-06-11
Fix(mysql): properly parse STORED/VIRTUAL computed columns

### LINKED ISSUE #5203: "STORED"  with generated columns not supported in mysql
Looks like STORED keyword for generated columns in create table is not supported in mysql dialect.
This seems similar to #4463, but that one was about postgres, and apparently the final fix was postgres-only.
Offending sql:
`create table test (a int, b generated always as (a+a) stored);
`

```
import sqlglot
sqlglot.parse_one('create table test (a int, b generated always as (a+a) stored);', dialect='mysql')

```
^ fails with:

sqlglot.errors.ParseError: Expecting ). Line 1, Col: 60.
  create table test (a int, b generated always as (a+a) **stored**);

**Official Documentation**

I tested this CREATE statement works fine with MariaDB 8.0.40, and it seems correct according to: 
https://dev.mysql.com/doc/refman/8.4/en/create-table-generated-columns.html

georgesittas:
Fixes #5203

Reference: https://dev.mysql.com/doc/refman/9.3/en/create-table-generated-columns.html

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -2033,7 +2033,7 @@ class ProjectionPolicyColumnConstraint(ColumnConstraintKind):
 # computed column expression
 # https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16
 class ComputedColumnConstraint(ColumnConstraintKind):
-    arg_types = {"this": True, "persisted": False, "not_null": False}
+    arg_types = {"this": True, "persisted": False, "not_null": False, "stored": False}
 
 
 class Constraint(Expression):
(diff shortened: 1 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -2033,7 +2033,7 @@ class ProjectionPolicyColumnConstraint(ColumnConstraintKind):
 # computed column expression
 # https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver16
 class ComputedColumnConstraint(ColumnConstraintKind):
-    arg_types = {"this": True, "persisted": False, "not_null": False}
+    arg_types = {"this": True, "persisted": False, "not_null": False, "stored": False}
```
VaggelisD: Should we reuse `persisted` instead of defining `stored`? I think it's referencing the same concept

georgesittas: Good catch, totally missed this.

PR #5215, opened 2025-06-12
feat(parser)!: JOIN pipe syntax, Set Operators as CTEs

geooo109:
This PR adds support for the `JOIN` pipe syntax. 

| `ORDER BY` | ✅  |


sqlglot/parser.py
```diff
@@ -8409,42 +8403,50 @@ def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
...
+
+    def _parse_pipe_syntax_query(self, query: exp.Select) -> t.Optional[exp.Select]:
         while self._match(TokenType.PIPE_GT):
             start = self._curr
             parser = self.PIPE_SYNTAX_TRANSFORM_PARSERS.get(self._curr.text.upper())
             if not parser:
-                set_op_query = self._parse_pipe_syntax_set_operator(query)
-                if not set_op_query:
+                parsed_query = self._parse_pipe_syntax_set_operator(query)
+                parsed_query = parsed_query or self._parse_pipe_syntax_join(query)
```
VaggelisD: [Nit]

```
parsed_query = self._parse_pipe_syntax_set_operator(query) or self._parse_pipe_syntax_join(query)
```

sqlglot/parser.py
```diff
@@ -8310,40 +8310,33 @@ def _identifier_expression(
...
-    def _parse_pipe_syntax_select(self, query: exp.Query) -> exp.Query:
+        return new_select.with_(new_cte, as_=query, copy=False)
+
+    def _parse_pipe_syntax_select(self, query: exp.Select) -> exp.Select:
         select = self._parse_select()
-        if isinstance(select, exp.Select):
+        if select:
+            if not query.selects:
+                query = query.select(*select.expressions)
+                return self._build_pipe_cte(query, [exp.Star()])
```
VaggelisD: This is a bit hard to follow imo, isn't this logic encapsulated in `_build_pipe_cte()` already?

sqlglot/parser.py
```diff
@@ -8310,40 +8310,33 @@ def _identifier_expression(
...
 
-            new_select = exp.select(*expressions, copy=False).from_(new_cte, copy=False)
-            if ctes:
-                new_select.set("with", ctes)
+        self._pipe_cte_counter += 1
+        new_cte = f"__tmp{self._pipe_cte_counter}"
 
-            return new_select.with_(new_cte, as_=query, copy=False)
+        with_ = query.args.get("with")
+        ctes = with_.pop() if with_ else None
```
VaggelisD, line 8320: [Nit] I think you don't have to check for `None`, if it doesn't exist `None` will be returned automatically

sqlglot/parser.py
VaggelisD: It's also on `_parse_pipe_syntax_query()`, line 8453

sqlglot/parser.py
```diff
@@ -8409,42 +8403,50 @@ def _parse_pipe_syntax_aggregate(self, query: exp.Query) -> exp.Query:
...
-            query.select("*", copy=False)
-
-        this = first_setop.this.pop()
+        first_setop.this.pop()
         distinct = first_setop.args.pop("distinct")
         setops = [first_setop.expression.pop(), *self._parse_expressions()]
 
+        query = self._build_pipe_cte(query, [exp.Star()])
+        with_ = query.args.get("with")
+        ctes = with_.pop() if with_ else None
```
VaggelisD, line 8416: Ditto, also this logic looks repeated with the previous comments

sqlglot/parser.py
geooo109: In this case we do this for `FROM x |> SELECT x.x1` the `query` instance is empty, and we need to create a CTE with `x.x1`.

The check in` _parse_pipe_syntax_query()` is for scenarios like `FROM x |> UNION (query1), (query2)`

tests/dialects/test_pipe_syntax.py
```diff
@@ -0,0 +1,361 @@
+from tests.dialects.test_dialect import Validator
+
+
+class TestPipeSyntax(Validator):
+    def test_select(self):
+        self.validate_identity("FROM x", "SELECT * FROM x")
+        self.validate_identity(
+            "FROM x |> SELECT x1, x2", "WITH __tmp1 AS (SELECT x1, x2 FROM x) SELECT * FROM __tmp1"
+        )
```
georgesittas, line 9: Can we actually do the `SELECT *` at the end lazily? Like you're at a state where your last query is:

```
SELECT x1, x2 FROM x
```

And then when you encounter another operator you create the CTE at that point. So that you don't have trivial `SELECT *` queries without a good reason.

geooo109: Should we push the creation of `CTEs` in the `_parse_pipe_syntax_query` ? So, to always leave the last query without a `SELECT * FROM cte`.

Issue #5212, opened 2025-06-12
Quotations on USERDEFINED datatypes don't roundtrip


NickCrews:
**Before you file an issue**
- [x] Check if the issue still exists on main

**Fully reproducible code snippet**
Please include a fully reproducible code snippet or the input sql, dialect, and expected output.

```python
import sqlglot as sg

sg.parse_one("""CAST( 5 as "MySchema"."MyType")""").sql()
# 'CAST(5 AS MySchema.MyType)'
# I expect the exact input
```

I expect the quotes to survive. This happens with no dialect specified, but also with the dialect I care about, postgres, specified.
Without the quotes, the casing of the type is incorrectly interpreted.

Looking at the parsed expression, it looks like the quotes are dropped at parse time, not at compile time.

**Official Documentation**
Please include links to official SQL documentation related to your issue.

PR #5216, opened 2025-06-12
fix(postgres)!: Preserve quoting for user defined types

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5212

I went ahead and fixed this for Postgres only; It seems to be one of the few dialects that supports user defined types and the identifier resolution is respected:

```SQL
postgres> CREATE TYPE "fOo";
CREATE TYPE

postgres> SELECT CAST(1 AS foo);
ERROR:  type "foo" does not exist
LINE 1: SELECT CAST(1 AS foo);
                         ^
postgres> SELECT CAST(1 AS "fOo");
ERROR:  type "fOo" is only a shell
LINE 1: SELECT CAST(1 AS "fOo");
```

DuckDB also supports UDTs but quoting doesn't seem to make a difference: 

```sql
duckdb> CREATE TYPE "fOo" AS INTEGER;

duckdb>  SELECT CAST (1 AS foo);
┌────────────────┐
│ CAST(1 AS foo) │
│     int32      │
├────────────────┤
│       1        │
└────────────────┘

duckdb>  SELECT CAST (1 AS "Foo");
┌────────────────┐
│ CAST(1 AS Foo) │
│     int32      │
├────────────────┤
│       1        │
└────────────────┘
```

Lastly, T-SQL also allows UDTs, but the approach of this PR wouldn't work because T-SQL also supports such statements (notice the builtin quoted type)

```SQL
CREATE TABLE [mytable]([email] [varchar](255) NOT NULL)
```

Which would be transpiled as the identifier `"VARCHAR"` to other dialects.

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -512,6 +512,18 @@ class Postgres(Dialect):
 
             return this
 
+        def _parse_user_defined_type(
+            self, identifier: exp.Identifier
+        ) -> t.Optional[exp.Expression]:
+            udt_type: exp.Identifier | exp.Dot = identifier
+
+            while self._match(TokenType.DOT):
+                part = self._parse_id_var()
+                if part:
+                    udt_type = exp.Dot(this=udt_type, expression=part)
+
+            return exp.DataType.build(udt_type, udt=True)
+
     class Generator(generator.Generator):
         SINGLE_STRING_INTERVAL = True
         RENAME_TABLE_WITH_DB = False
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4753,6 +4753,8 @@ class DataType(Expression):
                 if udt:
                     return DataType(this=DataType.Type.USERDEFINED, kind=dtype, **kwargs)
                 raise
+        elif isinstance(dtype, (Identifier, Dot)) and udt:
+            return DataType(this=DataType.Type.USERDEFINED, kind=dtype, **kwargs)
         elif isinstance(dtype, DataType.Type):
             data_type_exp = DataType(this=dtype)
         elif isinstance(dtype, DataType):
@@ -4792,9 +4794,6 @@ class DataType(Expression):
             if matches:
                 return True
         return False
-
-
-DATA_TYPE = t.Union[str, DataType, DataType.Type]
 
 
 # https://www.postgresql.org/docs/15/datatype-pseudo.html
@@ -5028,6 +5027,9 @@ class Dot(Binary):
 
         parts.reverse()
         return parts
+
+
+DATA_TYPE = t.Union[str, Identifier, Dot, DataType, DataType.Type]
 
 
 class DPipe(Binary):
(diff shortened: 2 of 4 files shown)


georgesittas:
> DuckDB also supports UDTs but quoting doesn't seem to make a difference: [...]

What if you include unsafe characters though that need to be delimited? E.g. `"foo bar"`.

sqlglot/dialects/postgres.py
```diff
@@ -512,6 +512,18 @@ def _parse_generated_as_identity(
...
             return this
 
+        def _parse_user_defined_type(
+            self, identifier: exp.Identifier
+        ) -> t.Optional[exp.Expression]:
+            udt_type: exp.Identifier | exp.Dot = identifier
+
+            while self._match(TokenType.DOT):
+                part = self._parse_id_var()
+                if part:
```
georgesittas, line 522: Why not do this for every dialect? i.e. create `Dot(...)`s.

georgesittas (commented):
@VaggelisD this fails in duckdb now:

```python
>>> import sqlglot
>>> sqlglot.parse_one('cast(x as "foo bar")', 'duckdb')
Cast(
  this=Column(
  to=DataType(this=Type.USERDEFINED, kind=foo bar))
>>> sqlglot.parse_one('cast(x as "foo bar")', 'duckdb').sql('duckdb')
'CAST(x AS foo bar)'
```

sqlglot/dialects/postgres.py
VaggelisD: For DuckDB I don't think this works properly anyways:

```SQL
duckdb>  CREATE TYPE "fοo bar" AS INTEGER;

duckdb> SELECT CAST (1 AS "foo bar");
Catalog Error:
Type with name foo bar does not exist!
Did you mean "fοo bar"?
```

Also, the Postgres version of parsing fails with T-SQL due to the possibility of quoting types, such as:

```SQL
CREATE TABLE [mytable]([email] [varchar](255) NOT NULL)
```

This would then be transpiled to other dialects such as:
```SQL
CREATE TABLE `vaggelis`.`default`.`mytable` (`email` `varchar`(255) NOT NULL)
```

Which fails e.g on Databricks: 

```SQL
dbx> CREATE TABLE `vaggelis`.`default`.`mytable` (`email` `varchar`(255) NOT NULL)
[UNSUPPORTED_DATATYPE] Unsupported data type "VARCHAR(255)"

dbx> CREATE TABLE `vaggelis`.`default`.`mytable` (`email` varchar(255) NOT NULL)
CREATE TABLE `vaggelis`.`default`.`mytable` (`email` varchar(255) NOT NULL) was successfully executed.
```

georgesittas: This is definitely a DuckDB bug...

```
D CREATE TYPE "fοo bar" AS INTEGER;
Catalog Error:
Type with name "fοo bar" already exists!
D DROP TYPE "foo bar";
Catalog Error:
Type with name foo bar does not exist!
Did you mean "fοo bar"?
```

georgesittas (approved):

PR #5217, opened 2025-06-12
dialect for IBM DB2

Salfiii:
The PR adds an dialect for IBM DB2 with test cases.


georgesittas:
Thanks a lot for the contribution!

Let's please split this across multiple small PRs. Each one should ideally add a single thing and it should be well documented. I'm closing this for now, but open them whenever you're ready, we'll be happy to help.

PR #5224, opened 2025-06-13
feat(parser): AS pipe syntax

geooo109:
This PR adds support for the `AS` pipe syntax. 

| `ORDER BY` | ✅  |


sqlglot/parser.py
```diff
@@ -8328,12 +8328,17 @@ def _identifier_expression(
...
-        if not query.selects:
-            query = query.select("*", copy=False)
-
-        self._pipe_cte_counter += 1
-        new_cte = f"__tmp{self._pipe_cte_counter}"
+    def _build_pipe_cte(
+        self,
+        query: exp.Query,
+        expressions: t.List[exp.Expression],
+        alias_cte: t.Optional[str | exp.Expression] = None,
```
georgesittas: When can this be a `str`? Isn't it an optional `TableAlias`?

PR #5225, opened 2025-06-13
feat(parser): EXTEND pipe syntax

geooo109:
This PR adds support for the `EXTEND` pipe syntax. 

| `ORDER BY` | ✅  |

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -935,6 +935,7 @@ class Parser(metaclass=_Parser):
         "AS": lambda self, query: self._build_pipe_cte(
             query, [exp.Star()], self._parse_table_alias()
         ),
+        "EXTEND": lambda self, query: self._parse_pipe_syntax_extend(query),
         "LIMIT": lambda self, query: self._parse_pipe_syntax_limit(query),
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -8473,6 +8474,11 @@ def _parse_pipe_syntax_pivot(self, query: exp.Select) -> exp.Select:
 
         return self._build_pipe_cte(query, [exp.Star()])
 
+    def _parse_pipe_syntax_extend(self, query: exp.Select) -> exp.Select:
+        self._match_text_seq("EXTEND")
```
georgesittas, line 8478: Why do we need to match this here? Isn't it done in `_parse_pipe_syntax_query`?

georgesittas (approved):

sqlglot/parser.py
geooo109: Nope, in `_parse_pipe_syntax_query`, we just get the correct parser. The corresponding method will do the consumption. We need to preserve this because there are parsing methods that need to consume the Token by themselves (`SELECT`, `JOIN` etc).

PR #5226, opened 2025-06-13
feat(parser): DROP pipe syntax

geooo109:
This PR adds support for the `DROP` pipe syntax. 

| `DROP`             |   ✅        |
| `ORDER BY` | ✅  |

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -935,6 +935,7 @@ class Parser(metaclass=_Parser):
         "AS": lambda self, query: self._build_pipe_cte(
             query, [exp.Star()], self._parse_table_alias()
         ),
+        "DROP": lambda self, query: self._parse_pipe_syntax_drop(query),
         "EXTEND": lambda self, query: self._parse_pipe_syntax_extend(query),
         "LIMIT": lambda self, query: self._parse_pipe_syntax_limit(query),
         "ORDER BY": lambda self, query: query.order_by(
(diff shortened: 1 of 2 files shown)


geooo109:
Needs a fix to be correct, will revisit soon.

georgesittas:
What fix does it need? Current approach looks reasonable.

EDIT: it actually does not work properly for a few reasons. First of all, after a `DROP` operation you can still access the names of the sources introduced by the previous operator. Second, you can chain `DROP` operations, so we'd need to append to the existing `EXCEPT (...)` list, instead of overwriting it on each one.

The `SET` and `RENAME` operators seem to have behave similarly, so we need to further discuss their implementation.

sqlglot/parser.py
```diff
@@ -8478,12 +8503,27 @@ def _parse_pipe_syntax_pivot(self, query: exp.Select) -> exp.Select:
...
+    def _parse_pipe_syntax_drop(self, query: exp.Select) -> exp.Select:
+        self._match_text_seq("DROP")
+        dropped_columns = self._parse_csv(self._parse_assignment)
+
+        select = query.selects[0].assert_is(exp.Star)
+        except_ = select.args.get("except", None)
+        if except_:
+            select.set("except", [*except_, *dropped_columns])
+        else:
+            select.set("except", dropped_columns)
```
```suggestion
        except_ = select.args.get("except") or []
        select.set("except", [*except_, *dropped_columns])
```

georgesittas (approved):

PR #5227, opened 2025-06-15
fix(tsql): Retain limit clause in subquery expression.

MarcusRisanger:
Fixes issue: https://github.com/TobikoData/sqlmesh/issues/4649 

Subqueries are parsed twice (at least for tsql) by `select_sql`, which causes subqueries using `LIMIT` or `TOP` to drop these on the second run, when converting CTAS statement to `SELECT .. INTO ..`. 

The subquery expression object suffers a side-effect in `Generator.select_sql` method where any `limit` expression is popped from the expression and propagated within the method as a string component. 

This simple fix reinjects the original `limit` expression to any query that HAD a `limit` expression prior to `select_sql`, but lacks it afterward.

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -1856,11 +1856,9 @@ WHERE
                 "spark": "SELECT * FROM A LIMIT 3",
             },
         )
-        self.validate_all(
+        self.validate_identity(
             "CREATE TABLE schema.table AS SELECT a, id FROM (SELECT a, (SELECT id FROM tb ORDER BY t DESC LIMIT 1) as id FROM tbl) AS _subquery",
-            write={
-                "tsql": "SELECT * INTO schema.table FROM (SELECT a AS a, id AS id FROM (SELECT a AS a, (SELECT TOP 1 id FROM tb ORDER BY t DESC) AS id FROM tbl) AS _subquery) AS temp"
-            },
+            "SELECT * INTO schema.table FROM (SELECT a AS a, id AS id FROM (SELECT a AS a, (SELECT TOP 1 id FROM tb ORDER BY t DESC) AS id FROM tbl) AS _subquery) AS temp",
         )
         self.validate_identity("SELECT TOP 10 PERCENT")
         self.validate_identity("SELECT TOP 10 PERCENT WITH TIES")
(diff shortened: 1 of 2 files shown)


georgesittas:
Hey @MarcusRisanger 👋 

Thanks a lot for digging into this! I think we may be able to simplify this a bit. I see that the `TOP 1` is preserved given this diff:

```diff
+++ b/sqlglot/dialects/tsql.py
@@ -1224,7 +1224,6 @@ class TSQL(Dialect):
                     # to amend the AST by moving the CTEs to the CREATE VIEW statement's query.
                     ctas_expression.set("with", with_.pop())

-            sql = super().create_sql(expression)

             table = expression.find(exp.Table)

@@ -1243,6 +1242,8 @@ class TSQL(Dialect):
                     select_into.limit(0, copy=False)

                 sql = self.sql(select_into)
+            else:
+                sql = super().create_sql(expression)

```

Wanna give it a shot? Feel free to update the PR and I can merge shortly.

MarcusRisanger:
Very nice @georgesittas - this is the difference between knowing the code base and just blindly implementing change based on debugging call stacks... 🚀

tests/dialects/test_tsql.py
```diff
@@ -1856,6 +1856,12 @@ def test_top(self):
                 "spark": "SELECT * FROM A LIMIT 3",
             },
         )
+        self.validate_all(
+            "CREATE TABLE schema.table AS SELECT a, id FROM (SELECT a, (SELECT id FROM tb ORDER BY t DESC LIMIT 1) as id FROM tbl) AS _subquery",
+            write={
+                "tsql": "SELECT * INTO schema.table FROM (SELECT a AS a, id AS id FROM (SELECT a AS a, (SELECT TOP 1 id FROM tb ORDER BY t DESC) AS id FROM tbl) AS _subquery) AS temp"
+            },
+        )
```
georgesittas:
```suggestion
        self.validate_identity(
            "CREATE TABLE schema.table AS SELECT a, id FROM (SELECT a, (SELECT id FROM tb ORDER BY t DESC LIMIT 1) as id FROM tbl) AS _subquery",
            "SELECT * INTO schema.table FROM (SELECT a AS a, id AS id FROM (SELECT a AS a, (SELECT TOP 1 id FROM tb ORDER BY t DESC) AS id FROM tbl) AS _subquery) AS temp",
        )
```

georgesittas (approved):

PR #5230, opened 2025-06-16
Fix(optimizer)!: resolve table "columns" in bigquery that produce structs

### LINKED ISSUE #5207: Optimizer fails in some specific cases
Sqlglot version: 26.25.3

Reading dialect: BigQuery
Writing dialect: N/A

```python
from loguru import logger
from sqlglot import Expression
from sqlglot.optimizer.qualify_columns import (
  pushdown_cte_alias_columns,
  qualify_columns,
  quote_identifiers,
  validate_qualify_columns,
)


def qualify(expression: Expression, dialect="bigquery"):
  schema = ensure_schema(None, dialect=dialect)
  expression = qualify_tables(expression, schema=schema)

  if Dialect.get_or_raise(dialect).PREFER_CTE_ALIAS_COLUMN:
    expression = pushdown_cte_alias_columns(expression)

  try:
    expression = qualify_columns(
      expression,
      schema,
      expand_alias_refs=True,
      expand_stars=True,
      infer_schema=None,
    )
  except OptimizeError as e:
    logger.error(f"Exception while qualifying query: [{e}].")
    raise

  expression = quote_identifiers(expression, dialect=dialect, identify=True)
  validate_qualify_columns(expression)

  return expression

ast = qualify(sqlglot.parse_one(sql, dialect="bigquery"))
```

This sql code fails:

```sql
WITH some_cte AS (
    SELECT 
      foo,
      bar
    FROM `database.schema.table`
)

SELECT
  *,
  TO_JSON_STRING(some_cte) AS serialized_row
FROM some_cte;
```

With the error message
```
File "(...)/sqlglot/optimizer/qualify_columns.py", line 126, in validate_qualify_columns
  sqlglot.errors.OptimizeError: Column '"some_cte"' could not be resolved
```

And here is a slightly different example that does work:

```sql
WITH some_cte AS (
    SELECT 
      *
    FROM `database.schema.table`
)

SELECT
  *,
  TO_JSON_STRING(some_cte) AS serialized_row
FROM some_cte;
```

Both examples are valid in bigquery.

georgesittas:
Fixes #5207

I've described the issue in this comment: https://github.com/tobymao/sqlglot/issues/5207#issuecomment-2961924796. Let me know what you think.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -7047,7 +7047,7 @@ class Semicolon(Expression):
 
 # BigQuery allows SELECT t FROM t and treats the projection as a struct value. This expression
 # type is intended to be constructed by qualify so that we can properly annotate its type later
-class TableAsStruct(Expression):
+class TableColumn(Expression):
     pass
 
 
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -7045,6 +7045,12 @@ class Semicolon(Expression):
     arg_types = {}
 
 
+# BigQuery allows SELECT t FROM t and treats the projection as a struct value. This expression
+# type is intended to be constructed by qualify so that we can properly annotate its type later
+class TableAsStruct(Expression):
+    pass
+
+
 def _norm_arg(arg):
     return arg.lower() if type(arg) is str else arg
 
--- a/sqlglot/optimizer/annotate_types.py
+++ b/sqlglot/optimizer/annotate_types.py
@@ -12,7 +12,7 @@ from sqlglot.helper import (
     seq_get,
 )
 from sqlglot.optimizer.scope import Scope, traverse_scope
-from sqlglot.schema import Schema, ensure_schema
+from sqlglot.schema import MappingSchema, Schema, ensure_schema
 from sqlglot.dialects.dialect import Dialect
 
 if t.TYPE_CHECKING:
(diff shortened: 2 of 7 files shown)


sqlglot/expressions.py
```diff
@@ -7045,6 +7045,12 @@ class Semicolon(Expression):
     arg_types = {}
 
 
+# BigQuery allows SELECT t FROM t and treats the projection as a struct value. This expression
+# type is intended to be constructed by qualify so that we can properly annotate its type later
+class TableAsStruct(Expression):
+    pass
+
```
VaggelisD: What did you think about simply using `exp.Table`? I'm worried that these meta nodes make all operations harder (finding nodes, reasoning, instance checks etc)

sqlglot/optimizer/scope.py
```diff
@@ -309,6 +313,13 @@ def columns(self):
 
         return self._columns
 
+    @property
+    def pseudocolumns(self):
+        if self._columns is None:
+            self._ensure_collected()
```
VaggelisD: Did you mean `_pseudocolumns`?

sqlglot/optimizer/scope.py
```diff
@@ -156,6 +158,8 @@ def _collect(self):
                 self._derived_tables.append(node)
             elif isinstance(node, exp.UNWRAPPED_QUERIES):
                 self._subqueries.append(node)
+            elif isinstance(node, exp.TableAsStruct):
+                self._pseudocolumns.append(node)
```
VaggelisD: Should we rename these as `table_as_columns` or something equivalent? I'd believe pseudocolumns would hint towards dialect peculiarities, INFORMATION_SCHEMA columns etc

sqlglot/optimizer/annotate_types.py
```diff
@@ -290,9 +290,48 @@ def annotate_scope(self, scope: Scope) -> None:
...
+        if self.schema.dialect == "bigquery" and isinstance(scope.expression, exp.Query):
+            struct_type = exp.DataType(
+                this=exp.DataType.Type.STRUCT,
+                expressions=[
+                    exp.ColumnDef(this=exp.to_identifier(select.output_name), kind=select.type)
+                    for select in scope.expression.selects
+                ],
+                nested=True,
+            )
+            self._set_type(scope.expression, struct_type)
```
VaggelisD: What does this do? Could we have other projections in that `selects` list that do not belong to the table/struct?

tests/test_optimizer.py
```diff
@@ -1580,3 +1580,31 @@ def test_manually_annotate_snowflake(self):
...
+        annotated = _annotate(example_query)
+
+        self.assertTrue(
+            annotated.selects[0].is_type("STRUCT<c1 BIGINT, c2 STRUCT<f1 BIGINT, f2 TEXT>>")
+        )
+
+        example_query = "WITH t AS (SELECT 1 AS c) SELECT t FROM t"
+        annotated = _annotate(example_query)
+
+        self.assertTrue(annotated.selects[0].is_type("STRUCT<c INt>"))
```
VaggelisD: Is the nested struct annotated properly here?

VaggelisD (commented):
Left some quick comments to better my understanding

sqlglot/expressions.py
georgesittas: A `Table` would be inconsistent with how it's normally used, i.e., as a child of `From`, `Join`, etc. To me this is a special expression, not a real table, hence why I treated it as such.

sqlglot/optimizer/scope.py
georgesittas: Good catch, thanks!

sqlglot/optimizer/scope.py
georgesittas: Sure, we can narrow it down.

sqlglot/optimizer/annotate_types.py
georgesittas: BigQuery allows `SELECT t FROM t`, and `t` could be anything. It could be a physical table, a CTE, a subquery, etc. It can also depend on other nested objects, so we need to recursively infer the types of each query and propagate them "upwards".

TL;DR, this annotates the `Select`'s type in `SELECT 1 AS x, 2 AS y` to `STRUCT<x INT, y INT>` only for BigQuery so we can reuse this type elsewhere.

> Could we have other projections in that selects list that do not belong to the table/struct?

Not sure what you mean here.

tests/test_optimizer.py
georgesittas: Could you elaborate? Not sure what you mean.

sqlglot/optimizer/annotate_types.py
georgesittas: I'll make it so if there's at least one unknown, the whole thing will be an unknown.

sqlglot/expressions.py
georgesittas: Renamed to `TableColumn`.

sqlglot/optimizer/scope.py
georgesittas: Fixed.

sqlglot/optimizer/scope.py
georgesittas: Renamed.

sqlglot/optimizer/annotate_types.py
georgesittas: Done^.

PR #5231, opened 2025-06-16
feat(parser): TABLESAMPLE pipe syntax

geooo109:
This PR adds support for the `TABLESAMPLE` pipe syntax. 

| `DROP`             |           |
| `ORDER BY` | ✅  |
| `TABLESAMPLE`      |    ✅       |

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -942,6 +942,7 @@ class Parser(metaclass=_Parser):
         ),
         "PIVOT": lambda self, query: self._parse_pipe_syntax_pivot(query),
         "SELECT": lambda self, query: self._parse_pipe_syntax_select(query),
+        "TABLESAMPLE": lambda self, query: self._parse_pipe_syntax_tablesample(query),
         "UNPIVOT": lambda self, query: self._parse_pipe_syntax_pivot(query),
         "WHERE": lambda self, query: query.where(self._parse_where(), copy=False),
     }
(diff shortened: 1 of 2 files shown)


tests/dialects/test_pipe_syntax.py
```diff
@@ -391,3 +391,17 @@ def test_extend(self):
             "FROM (SELECT 'foo1' AS item1, 2 AS item2 UNION ALL SELECT 'foo2' AS item1, 5 AS item2) |> EXTEND SUM(item2) OVER() AS item2_sum",
             "WITH __tmp1 AS (SELECT *, SUM(item2) OVER () AS item2_sum FROM (SELECT 'foo1' AS item1, 2 AS item2 UNION ALL SELECT 'foo2' AS item1, 5 AS item2)) SELECT * FROM __tmp1",
         )
+
+    def test_tablesample(self):
```
georgesittas, line 395: These tests are only checking the happy path, can we also test joins, etc, where multiple sources are present in the scope? IIRC the table sample needs to be applied only to one table.

sqlglot/parser.py
```diff
@@ -8484,6 +8485,17 @@ def _parse_pipe_syntax_extend(self, query: exp.Select) -> exp.Select:
...
         return self._build_pipe_cte(query, [exp.Star()])
 
+    def _parse_pipe_syntax_tablesample(self, query: exp.Select) -> exp.Select:
+        sample = self._parse_table_sample()
+
+        with_ = query.args.get("with")
+        if with_:
+            with_.expressions[-1].this.set("sample", sample)
+        else:
+            query.set("sample", sample)
```
georgesittas, line 8495: What if there are joins?

tests/dialects/test_pipe_syntax.py
geooo109: BigQuery only supports sampling on base tables. So the following for example isn't valid :

```
FROM x |> JOIN y ON x.a = y.id |> SELECT a, id |> TABLESAMPLE SYSTEM (1 PERCENT);
```

sqlglot/parser.py
geooo109: Explained in the other comment.

PR #5234, opened 2025-06-16
Feat(snowflake): improve transpilation of unnested object lookup

georgesittas:
This PR:

- Introduces a new transformation for Snowflake, in order to transpile BigQuery queries with unnests to it
- Modifies the `walk_in_scope` logic to stop considering plain `UDTF`s without `Query` children as scope boundaries

We incorrectly transpile BigQuery queries like the following to Snowflake today:

```python
import sqlglot
print(sqlglot.transpile("select _u.foo from unnest([struct('x' as foo, 'y' AS bar)]) as _u", "bigquery", "snowflake")[0])
# SELECT _u.foo FROM TABLE(FLATTEN(INPUT => [OBJECT_CONSTRUCT('foo', 'x', 'bar', 'y')])) AS _t0(seq, key, path, index, _u, this)
```

The resulting query fails with `Error: invalid identifier '_U.FOO' (line 1)`, because the value `_u` is a `VARIANT`, and so accessing its contents using the dot notation is not allowed. Instead, Snowflake requires that we use either the colon (`x:y`) or the bracket syntax `x['y']` for this reason.

To address this, the new transformation looks for unnests that have a single column alias, and then replaces any columns in scope that are qualified with said alias.

I looked into Snowflake's [docs](https://docs.snowflake.com/en/sql-reference/functions/flatten) for `TABLE(FLATTEN(...))`, and I believe the expected receivers of the lookup operation should be `OBJECT` or `VARIANT` values, for which it is safe to replace the dot syntax with the bracket notation.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -31,6 +31,7 @@ from sqlglot.dialects.dialect import (
 )
 from sqlglot.generator import unsupported_args
 from sqlglot.helper import flatten, is_float, is_int, seq_get
+from sqlglot.optimizer.scope import find_all_in_scope
 from sqlglot.tokens import TokenType
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -333,6 +334,28 @@ def _json_extract_value_array_sql(
...
+                and len(unnest_alias.columns) == 1
+            ):
+                unnest_aliases.add(unnest_alias.columns[0].name)
+
+        if unnest_aliases:
+            for c in find_all_in_scope(expression, exp.Column):
+                if c.table in unnest_aliases:
+                    bracket_lhs = c.args["table"]
+                    bracket_rhs = exp.Literal.string(c.name)
+                    c.replace(exp.Bracket(this=bracket_lhs, expressions=[bracket_rhs]))
```
VaggelisD, line 360: Should this be `JSONExtract` with `variant_extract=True`, to match the semantics?

georgesittas: It's not necessary, no. What you're suggesting would generate the `colon` syntax (or `GET_PATH`? I forget). We can also work with the bracket syntax, which is simpler to construct.

VaggelisD: Right, I remembered that the colon syntax would be equivalent to the bracket one but it looks like it's not, `_u:foo` raises an error.

Issue #5239, opened 2025-06-18
Unpivot function does not work in Redshift and we are looking to transpile from Snowflake to Redshift


mansiim:
**Before you file an issue**
- Make sure you specify the "read" dialect eg. `parse_one(sql, read="snowflake")`
- Make sure you specify the "write" dialect eg. `ast.sql(dialect="redshift")`
- Check if the issue still exists on main

**Fully reproducible code snippet**
    For example:
        SELECT * FROM tbl UNPIVOT(sales FOR month IN (jan, feb, mar))

    Becomes:
        SELECT *, 'jan' AS month, jan AS sales FROM tbl
        UNION ALL
        SELECT *, 'feb' AS month, feb AS sales FROM tbl
        UNION ALL
        SELECT *, 'mar' AS month, mar AS sales FROM tbl

**Official Documentation**
Please include links to official SQL documentation related to your issue.

mansiim:
I have written the code fix would like the permissions to make a branch

georgesittas:
Hey 👋 

Can you please explain your solution, in this thread, before opening a PR? We have no plans to look into this, so I'll close this for now, but happy to continue discussing.

mansiim:
@georgesittas  UNPIVOT in snowflake does not have an equivalent to redshift. So we need to convert the unpivot to a union all

mansiim:
Can you give me access to make a PR it would be easier for me to show ?

georgesittas:
You don't need special permission to do it, just fork the project and make a pull request.

PR #5246, opened 2025-06-19
feat: add UNPIVOT to UNION ALL transformation for Redshift dialect

mansiim:
This commit adds a complete implementation of the unpivot_to_union_all transform function, which converts SQL UNPIVOT expressions into equivalent UNION ALL queries. The transform handles:

- Extracting the value column and name column from the UNPIVOT expression
- Processing each column in the IN clause to create individual SELECT statements
- Properly handling column aliases in both the name and value positions
- Maintaining all other query clauses (WHERE, GROUP BY, HAVING, etc.) from the original query
- Supporting PivotAlias expressions for more complex unpivot operations
Issue opened at : https://github.com/tobymao/sqlglot/issues/5239

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -202,6 +202,7 @@ class Redshift(Postgres):
                     transforms.eliminate_semi_and_anti_joins,
                     transforms.unqualify_unnest,
                     transforms.unnest_generate_date_array_using_recursive_cte,
+                    transforms.unpivot_to_union_all,
                 ]
             ),
             exp.SortKeyProperty: lambda self,
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -1019,6 +1019,7 @@ def eliminate_window_clause(expression: exp.Expression) -> exp.Expression:
 
     return expression
 
+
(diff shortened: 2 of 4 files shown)


georgesittas:
Hey @mansiim, can you please elaborate on what exactly does not work in Redshift? I see that the `UNPIVOT` operator is, in fact supported ([docs](https://docs.aws.amazon.com/redshift/latest/dg/r_FROM_clause-pivot-unpivot-examples.html)). I also just tested out the example you linked in the issue, and it is being transpiled correctly to Redshift:

```python
>>> import sqlglot
>>> sqlglot.transpile("""
... SELECT *
...   FROM monthly_sales
...     UNPIVOT (sales FOR month IN (jan, feb, mar, apr))
...   ORDER BY empid;
... """, "snowflake", "redshift")
['SELECT * FROM monthly_sales UNPIVOT(sales FOR month IN (jan, feb, mar, apr)) ORDER BY empid']
```

I verified that this produces the same results as in Snowflake ([docs](https://docs.snowflake.com/en/sql-reference/constructs/unpivot)), when executed against Redshift.

georgesittas:
Closing this until more details are provided.

PR #5248, opened 2025-06-19
feat(parser): start with SELECT and nested pipe syntax

geooo109:
This PR removes the support OF `DROP` and `SET` pipe syntax operators.
Adds support for pipe syntax with nesting and also starting the syntax with `SELECT`.

| `SET`              |           |
| `DROP`             |          |
| `ORDER BY` | ✅  |
| `TABLESAMPLE`      |   ✅         |

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -794,7 +794,7 @@ class Parser(metaclass=_Parser):
         exp.Condition: lambda self: self._parse_assignment(),
         exp.DataType: lambda self: self._parse_types(allow_identifiers=False, schema=True),
         exp.Expression: lambda self: self._parse_expression(),
-        exp.From: lambda self: self._parse_from(joins=True),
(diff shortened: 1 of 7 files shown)


sqlglot/parser.py
```diff
@@ -1888,7 +1886,11 @@ def _parse_statement(self) -> t.Optional[exp.Expression]:
             return self._parse_command()
 
         expression = self._parse_expression()
-        expression = self._parse_set_operations(expression) if expression else self._parse_select()
+        expression = (
+            self._parse_set_operations(expression)
+            if expression
+            else self._parse_select(consume_pipe=True)
+        )
```
georgesittas: This doesn't need to change, right? `consume_pipe` is `True` by default.

geooo109: yep, thanks

georgesittas (approved):
Getting this in for now, seems ok. We can iterate as needed.

PR #5250, opened 2025-06-19
feat(fabric): Add precision cap to temporal data types

mattiasthalen:
This pull request introduces Fabric-specific enhancements to the SQLGlot library, focusing on precision handling. The changes ensure compatibility with Microsoft Fabric's data warehouse by adding precision constraints for temporal data types.

### Fabric dialect enhancements:

* Implemented a custom `datatype_sql` method to cap precision for temporal types (`TIME`, `DATETIME2`, `DATETIMEOFFSET`) at 6 digits, defaulting to 6 when unspecified.

>**NOTE:** Depends on #5249

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -40,7 +40,7 @@ class Fabric(TSQL):
             exp.DataType.Type.SMALLDATETIME: "DATETIME2(6)",
             exp.DataType.Type.NCHAR: "CHAR",
             exp.DataType.Type.NVARCHAR: "VARCHAR",
-            exp.DataType.Type.TEXT: "VARCHAR",
+            exp.DataType.Type.TEXT: "VARCHAR(MAX)",
             exp.DataType.Type.IMAGE: "VARBINARY",
             exp.DataType.Type.TINYINT: "SMALLINT",
             exp.DataType.Type.UTINYINT: "SMALLINT",  # T-SQL parses TINYINT as UTINYINT
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-
 from sqlglot import exp
 from sqlglot.dialects.dialect import NormalizationStrategy
 from sqlglot.dialects.tsql import TSQL
@@ -37,9 +36,8 @@ class Fabric(TSQL):
             # Fabric doesn't support these types, map to alternatives
             exp.DataType.Type.MONEY: "DECIMAL",
             exp.DataType.Type.SMALLMONEY: "DECIMAL",
-            exp.DataType.Type.DATETIME: "DATETIME2",
-            exp.DataType.Type.SMALLDATETIME: "DATETIME2",
-            exp.DataType.Type.TIMESTAMPTZ: "DATETIME2",
+            exp.DataType.Type.DATETIME: "DATETIME2(6)",
+            exp.DataType.Type.SMALLDATETIME: "DATETIME2(6)",
             exp.DataType.Type.NCHAR: "CHAR",
             exp.DataType.Type.NVARCHAR: "VARCHAR",
             exp.DataType.Type.TEXT: "VARCHAR",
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/fabric.py
```diff
@@ -37,9 +36,8 @@ class Generator(TSQL.Generator):
...
             exp.DataType.Type.MONEY: "DECIMAL",
             exp.DataType.Type.SMALLMONEY: "DECIMAL",
-            exp.DataType.Type.DATETIME: "DATETIME2",
-            exp.DataType.Type.SMALLDATETIME: "DATETIME2",
-            exp.DataType.Type.TIMESTAMPTZ: "DATETIME2",
+            exp.DataType.Type.DATETIME: "DATETIME2(6)",
+            exp.DataType.Type.SMALLDATETIME: "DATETIME2(6)",
             exp.DataType.Type.NCHAR: "CHAR",
             exp.DataType.Type.NVARCHAR: "VARCHAR",
             exp.DataType.Type.TEXT: "VARCHAR",
```
fresioAS: `2025-06-20 08:47:41,880 - MainThread - sqlmesh.core.engine_adapter.base - DEBUG - Executing SQL: ALTER TABLE [sqlmesh].[_environments] ADD [suffix_target] VARCHAR; (base.py:2184)`
`String or binary data would be truncated in table 'test_wh.sqlmesh._environments', column 'suffix_target'. Truncated value: 's'`
````suggestion
            exp.DataType.Type.TEXT: "VARCHAR(MAX)",
```

mattiasthalen: Hm... I'm not sure, I like the idea, but at the same time (MAX) is in preview.

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,46 @@ class Generator(TSQL.Generator):
             exp.DataType.Type.DOUBLE: "FLOAT",
             exp.DataType.Type.INT: "INT",  # T-SQL uses INTEGER
         }
+
+        def datatype_sql(self, expression: exp.DataType) -> str:
+            """
+            Override datatype generation to handle Fabric-specific precision limitations.
+
+            Fabric limits temporal types (TIME, DATETIME2, DATETIMEOFFSET) to max 6 digits precision.
+            When no precision is specified, we default to 6 digits.
```
georgesittas, line 61: Hey @mattiasthalen, can you remind me why we want to default to 6 digits? What does Fabric do by default?

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,46 @@ class Generator(TSQL.Generator):
...
+            """
+            if expression.is_type(
+                exp.DataType.Type.TIME,
+                exp.DataType.Type.DATETIME2,
+                exp.DataType.Type.TIMESTAMPTZ,  # DATETIMEOFFSET in Fabric
+            ):
+                # Get the current precision (first expression if it exists)
+                precision = None
+                if expression.expressions:
+                    precision = expression.expressions[0]
```
georgesittas: What about...
```suggestion
                precision = expression.find(exp.DataType.Param)
```

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,46 @@ class Generator(TSQL.Generator):
...
+
+                # Determine the target precision
+                if precision is None:
+                    # No precision specified, default to 6
+                    target_precision = 6
+                elif (
+                    isinstance(precision, exp.DataTypeParam)
+                    and isinstance(precision.this, exp.Literal)
+                    and precision.this.is_int
+                ):
```
georgesittas:
```suggestion
                elif isinstance(precision, exp.DataTypeParam) and precision.this.is_int:
```

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,46 @@ class Generator(TSQL.Generator):
...
+                    # Cap precision at 6
+                    current_precision = int(precision.this.this)
```
georgesittas:
```suggestion
                    current_precision = precision.this.to_py()
```

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,46 @@ class Generator(TSQL.Generator):
...
+                    target_precision = min(current_precision, 6)
+                else:
+                    # Non-literal precision, leave as-is
+                    return super().datatype_sql(expression)
```
georgesittas: Is this branch implemented as a defensive measure? Is it actually reachable?

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,46 @@ class Generator(TSQL.Generator):
...
+
+                # Create a new expression with the target precision
+                new_expression = exp.DataType(
+                    this=expression.this,
+                    expressions=[exp.DataTypeParam(this=exp.Literal.number(str(target_precision)))],
```
georgesittas:
```suggestion
                    expressions=[exp.DataTypeParam(this=exp.Literal.number(target_precision))],
```

sqlglot/dialects/fabric.py
fresioAS: Varchar(8000)? As far as I can see it evaluates to 1 now

sqlglot/dialects/fabric.py
mattiasthalen: 6 is the upper limit, and at the same time, precision is mandatory. IIRC, let me double check and get back.

If it is mandatory, would you rather see something else as default?

sqlglot/dialects/fabric.py
georgesittas: Indeed, the docs specify:

> Support for varchar (max) and varbinary (max) is currently in preview for the Warehouse. The string columns from the underlying Delta Lake files in One Lake are represented as varchar(8000) instead of varchar(max) in the SQL analytics endpoint. The limit for storage in varchar(max) is currently 1 MB in Fabric Data Warehouse.

Don't feel strongly about this, happy to give it a go if you've experienced that it works fine, despite being in preview currently.

sqlglot/dialects/fabric.py
georgesittas: Ah, that would make sense. If mandatory, let's keep it maxed out like you have it.

mattiasthalen:
```bash
	Msg 24597, Level 16, State 1, Line 1
An integer precision value between 0 and 6 must be specified.
```

Will keep it at 6 :)

sqlglot/dialects/fabric.py
mattiasthalen: I'm not seeing any issues with it so far. If issues occur, we can handle that then. Adding the max precision :)

sqlglot/dialects/fabric.py
```diff
@@ -54,3 +52,37 @@ class Generator(TSQL.Generator):
...
+                exp.DataType.Type.TIMESTAMPTZ,  # DATETIMEOFFSET in Fabric
+            ):
+                # Get the current precision (first expression if it exists)
+                precision = expression.find(exp.DataTypeParam)
+
+                # Determine the target precision
+                if precision is None:
+                    # No precision specified, default to 6
+                    target_precision = 6
+                elif isinstance(precision, exp.DataTypeParam) and precision.this.is_int:
```
georgesittas:
```suggestion
                elif precision.this.is_int:
```

georgesittas (approved):
Thanks, LGTM! I'll go ahead and apply a couple of small changes and get this in.

PR #5259, opened 2025-06-21
feat(fabric): implement create_sql method to handle Fabric-specific S…

mattiasthalen:
This pull request introduces enhancements to the Fabric dialect in `sqlglot` by adding support for handling Fabric-specific SQL syntax differences, particularly around `CREATE` statements. It also includes corresponding test cases to validate these changes. The most important changes are grouped into dialect enhancements and test coverage improvements.

### Dialect Enhancements:

* [`sqlglot/dialects/fabric.py`](diffhunk://#diff-ec3f7bdb2c80d6a8318d44a7f5fcb07a5aa1a3f4f3641156eb9f14b1a0931f6aR89-R133): Added a new `create_sql` method to handle Fabric-specific differences, such as converting `REPLACE` to `EXISTS` and ensuring `information_schema` references are properly uppercased for case-sensitive identifiers. This includes logic for generating `IF NOT EXISTS` wrappers for schemas and tables.

### Test Coverage Improvements:

* [`tests/dialects/test_fabric.py`](diffhunk://#diff-4e9274522d52bed321c05a42ca1956bf400b6421ecd08157e6499f431670ede7R65-R106): Added a new test method `test_information_schema_case_sensitivity` to validate the behavior of `CREATE SCHEMA` and `CREATE TABLE` statements with `IF NOT EXISTS` logic, including support for schema and database qualifiers.…QL syntax and case sensitivity

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -86,3 +86,48 @@ class Fabric(TSQL):
                 return super().datatype_sql(new_expression)
 
             return super().datatype_sql(expression)
+
+        def create_sql(self, expression: exp.Create) -> str:
+            """
+            Override create_sql to handle Fabric-specific differences:
+            1. Convert REPLACE to EXISTS since Fabric doesn't support CREATE OR REPLACE/ALTER
+            2. Ensure information_schema references are uppercase (case-sensitive identifiers)
+            """
+            kind = expression.kind
+
+            # Fabric doesn't support CREATE OR REPLACE/ALTER, so convert replace=True to exists=True
+            replace = expression.args.get("replace")
+            exists = expression.args.get("exists")
+
+            if replace and not exists:
+                # Set exists=True and remove replace=True to use IF NOT EXISTS logic
+                expression.set("exists", True)
+                expression.args.pop("replace", None)
+                exists = True
+
+            if exists and kind in ("SCHEMA", "TABLE"):
+                # Handle the case where we need to generate IF NOT EXISTS with information_schema
+                # Call the parent method but intercept the specific parts we need to uppercase
+                table = expression.find(exp.Table)
+                identifier = self.sql(exp.Literal.string(exp.table_name(table) if table else ""))
+
+                # Get the base SQL without the IF NOT EXISTS wrapper
+                expression_copy = expression.copy()
+                expression_copy.args.pop("exists", None)
+                sql = super().create_sql(expression_copy)
+                sql_with_ctes = self.prepend_ctes(expression, sql)
+                sql_literal = self.sql(exp.Literal.string(sql_with_ctes))
+
+                if kind == "SCHEMA":
+                    return f"""IF NOT EXISTS (SELECT * FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = {identifier}) EXEC({sql_literal})"""
+                elif kind == "TABLE":
+                    assert table
+                    where = exp.and_(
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/fabric.py
```diff
@@ -86,3 +86,48 @@ def datatype_sql(self, expression: exp.DataType) -> str:
...
+            replace = expression.args.get("replace")
+            exists = expression.args.get("exists")
+
+            if replace and not exists:
+                # Set exists=True and remove replace=True to use IF NOT EXISTS logic
+                expression.set("exists", True)
+                expression.args.pop("replace", None)
+                exists = True
+
+            if exists and kind in ("SCHEMA", "TABLE"):
```
georgesittas, line 108: Not sure if L108-L130 are necessary. I think we should be able to change T-SQL's information schema identifiers to uppercase, so that Fabric just works™ and we don't have to repeat logic.

The important bit here appears to be the logic around `REPLACE`, since neither it nor `ALTER` are supported by Fabric.

mattiasthalen: Ah, I think I get what you're saying now. The information schema bit is in lower case, since it wasn't an issue prior to fabric. So I should open a PR on TSQL, right?

Regarding `REPLACE` and `ALTER`, the latter works with views:
```sql
CREATE OR ALTER VIEW test_view AS (SELECT * FROM INFORMATION_SCHEMA.TABLES)
```

But at the same time, I'm wondering if we even should do this. The user would get an error. I guess the question is if this should be handled in sqlglot or where sqglot is used.

georgesittas: > So I should open a PR on TSQL, right?

Correct.

> But at the same time, I'm wondering if we even should do this. The user would get an error. I guess the question is if this should be handled in sqlglot or where sqglot is used.

I think handling what we can in sqlglot makes sense. So, if we've verified that `CREATE OR ALTER` works for views, then let's apply it only in that case. So they don't support anything similar for tables?

mattiasthalen: Correct. They suggest dropping the table if it exists, then creating it.

georgesittas: Got it, thanks. So just need to make sure we implement the drop & create semantics in sqlmesh then, i.e., we need to set this for the Fabric adapter:

```
SUPPORTS_REPLACE_TABLE = False
```

EDIT: nvm, it appears that it [inherits](https://github.com/TobikoData/sqlmesh/pull/4751/files#diff-bb8cb0615b44c904b56385ed1101eb602fe623a1b4478cf33d74c6b32748eca3R21) from the MSSQL adapter in the open PR, so [that should be taken care of already](https://github.com/TobikoData/sqlmesh/blob/66ac52a17ce296d4b1a4f45b61fab940add857a1/sqlmesh/core/engine_adapter/mssql.py#L53). All good :)

mattiasthalen: Ok, where does this leave my PR, redundant? :)

georgesittas: A big chunk of it is redundant, as long as the T-SQL change is tested & works for both it and Fabric.

If `CREATE OR ALTER` is supported for views, then I think it's good to keep that logic. Use `self.unsupported` if `replace` is set and `kind` is not `"VIEW"`.

georgesittas:
I'll close this for now since we've changed the PR's scope a bit. Let's discuss any changes related to views in a followup PR.

Issue #5265, opened 2025-06-23
Redshift: TO_TIMESTAMP() transpilation is ignoring scale parameter


MatiasCasaliSplit:
Read: Snowflake
Write: Redshift

**Description**
When transpiling the function `TO_TIMESTAMP(col, n)` from Snowflake to Redshift, the transpiler is not taking into account the scale parameter and thus it's considering all timestamps as Unix seconds.

**Fully reproducible code snippet**
```
import sqlglot

query = "SELECT TO_TIMESTAMP(1659981729000, 3)"
transpiled_query = sqlglot.parse_one(query, dialect="snowflake").sql(dialect="redshift")

print(transpiled_query)
```

Actual: `"SELECT (TIMESTAMP 'epoch' + 1659981729000 * INTERVAL '1 SECOND')"`.
Expected: `"SELECT (TIMESTAMP 'epoch' + 1659981729000 * INTERVAL '0.001 SECOND')"`.

**Official Documentation**
[Snowflake TO_TIMESTAMP documentation](https://docs.snowflake.com/en/sql-reference/functions/to_timestamp)

georgesittas:
This is low priority for the core team– feel free to take a stab at it if you want. We'll be happy to accept well-tested & documented PRs.

PR #5266, opened 2025-06-23
fix(redshift): handle scale parameter in to_timestamp

MatiasCasaliSplit:
PR addressing [this issue](https://github.com/tobymao/sqlglot/issues/5265).

Since Redshift does not have an equivalent to the function `TO_TIMESTAMP`, this PR just extends the existing solution by dividing the unix epoch by the appropriate power of 10.

As an example of the change, the query mentioned in the issue `(SELECT TO_TIMESTAMP(1659981729000, 3))` is now transpiled to `SELECT (TIMESTAMP 'epoch' + (16599817290000 / POWER(10, 4)) * INTERVAL '1 SECOND')`.

### DIFF
--- a/sqlglot/dialects/redshift.py
+++ b/sqlglot/dialects/redshift.py
@@ -214,7 +214,7 @@ class Redshift(Postgres):
             exp.TsOrDsAdd: date_delta_sql("DATEADD"),
             exp.TsOrDsDiff: date_delta_sql("DATEDIFF"),
             exp.UnixToTime: lambda self,
-            e: f"(TIMESTAMP 'epoch' + {self.sql(e.this)} * INTERVAL '1 SECOND')",
+            e: self._unix_to_time_sql(e),
         }
 
         # Postgres maps exp.Pivot to no_pivot_sql, but Redshift support pivots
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/redshift.py
```diff
@@ -447,3 +446,15 @@ def array_sql(self, expression: exp.Array) -> str:
...
+            scale = expression.args.get("scale")
+
+            if (
+                scale not in (None, exp.UnixToTime.SECONDS)
+                and isinstance(scale, exp.Literal)
+                and scale.is_int
+            ):
+                return f"(TIMESTAMP 'epoch' + ({self.sql(expression.this)} / POWER(10, {scale.to_py()})) * INTERVAL '1 SECOND')"
+
+            return f"(TIMESTAMP 'epoch' + {self.sql(expression.this)} * INTERVAL '1 SECOND')"
```
VaggelisD: [Nit] We can simplify the entire section here, e.g:

```Python
            scale = expression.args.get("scale")
            this = self.sql(expression.this)

            if (
                scale not in (None, exp.UnixToTime.SECONDS)
                and scale.is_int
            ):
                this = f"({this} / POWER(10, {scale.to_py()}))"

            return f"(TIMESTAMP 'epoch' + {this} * INTERVAL '1 SECOND')"
```

MatiasCasaliSplit: Thank you for your feedback. I've updated the PR based on the suggestion.

I had to change the condition `scale not in (None, exp.UnixToTime.SECONDS)` to `scale is not None and scale != exp.UnixToTime.SECONDS` because mypy couldn't properly determine that `scale` was not None when comparing against a tuple.

georgesittas (approved):
Thank you for the contribution!

PR #5268, opened 2025-06-23
fix(athena): avoid escaping backslashes when generating athena `SELECT` statements

### LINKED ISSUE #5267: Athena adds an extra backslash to string literals

```
In [15]: import sqlglot.expressions as sge

In [16]: sge.convert(r"\d+").sql("athena")
Out[16]: "'\\\\d+'"

In [17]: sge.convert(r"\d+").sql("trino")
Out[17]: "'\\d+'"
```

**Official Documentation**

The Athena regular expression docs just link to the Trino docs: https://docs.aws.amazon.com/athena/latest/ug/functions-env3.html

cpcloud:
Closes #5267.

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -165,3 +165,6 @@ class Athena(Trino):
                 return self._hive_generator.generate(expression, copy)
 
             return super().generate(expression, copy)
+
+
+del Athena.UNESCAPED_SEQUENCES["\\\\"], Athena.ESCAPED_SEQUENCES["\\"]
--- a/tests/dialects/test_athena.py
+++ b/tests/dialects/test_athena.py
@@ -36,6 +36,7 @@ class TestAthena(Validator):
             write_sql='/* leading comment */ SELECT * FROM "foo"',
             identify=True,
         )
+        self.validate_identity(r"SELECT '\d+'")
 
     def test_ddl(self):
         # Hive-like, https://docs.aws.amazon.com/athena/latest/ug/create-table.html


sqlglot/dialects/athena.py
```diff
@@ -165,3 +165,6 @@ def generate(self, expression: exp.Expression, copy: bool = True) -> str:
                 return self._hive_generator.generate(expression, copy)
 
             return super().generate(expression, copy)
+
+
+del Athena.UNESCAPED_SEQUENCES["\\\\"], Athena.ESCAPED_SEQUENCES["\\"]
```
georgesittas, line 170: How does this interact with DDLs that use escaped backslashes? I'm asking because DDLs are generated according to Hive's rules, which treats `\` as a "string escape". I'm worried this change will lead to incorrectly omitting an escape `\` in Hive strings. This may not be properly tested today, given how recent the other fix got in.

cpcloud: Yeah, it's a good question. I weighed adding testing scope to the PR against the urgency of all regular expression `SELECT` statements being broken in Athena, and decided to fix that without expanding scope.

georgesittas: That's fair, but I'm also reluctant to merge this without proper coverage. Otherwise, it may result in a separate regression, basically "reverting" the fix I linked to the relevant issue.

We can take a look if you don't have bandwidth to investigate further.

cpcloud: Let me see about adding a test for DDL with backslashes.

cpcloud: It looks like the recent fix added a test for `CREATE`. What other coverage did you have in mind?

georgesittas: > What other coverage did you have in mind?

Something like this breaks with this PR, i.e., including a literal backslash in a Hive string:

```python
>>> import sqlglot
>>> print(sqlglot.transpile("create external table george.t (id int comment 'foo \\\\ bar') location 's3://my-bucket/george/'", "athena")[0])
CREATE EXTERNAL TABLE george.t (id INT COMMENT 'foo \\\\ bar') LOCATION 's3://my-bucket/george/'
```

The actual statement you would execute against Athena is:

```sql
create external table george.t (id int comment 'foo \\ bar') location 's3://my-bucket/george/'
```

^ Each backslash needs to be escaped in the context of Python, hence why the input looks that way. Notice that the output, after `print`, contains two extraneous backslashes. That's not the case in the current main branch:

```python
>>> import sqlglot
>>> print(sqlglot.transpile("create external table george.t (id int comment 'foo \\\\ bar') location 's3://my-bucket/george/'", "athena")[0])
CREATE EXTERNAL TABLE george.t (id INT COMMENT 'foo \\ bar') LOCATION 's3://my-bucket/george/'
```

cpcloud: Unfortunately I won't have the bandwidth to add more comprehensive testing. While my PR does break that use case, all non-POSIX regular expression functionality using backslashes was broken by the original fix.

georgesittas: Ok, no worries, thanks for trying it out anyway.

I'll close this PR and we'll take a look soon. Apologies about the regression, Athena is a pain due to the Hive / Trino mixture, and some cases are tricky to anticipate..

cpcloud: No worries! Definitely hit plenty of "fun" 😉 Athena-related issues on the Ibis side due to its hybrid nature.

georgesittas: For reference: https://github.com/tobymao/sqlglot/pull/5314. This ended up causing quite a headache, but hopefully the refactored version should be more robust. Let me know if you come across any other weirdness.

PR #5269, opened 2025-06-23
feat(fabric): implement UnixToTime transformation to DATEADD syntax

mattiasthalen:
This pull request introduces enhancements to the `Fabric` dialect in `sqlglot`, focusing on adding support for Unix timestamp transformations, improving precision handling, and expanding test coverage. The most important changes include implementing a transformation for `UnixToTime`, capping precision for certain data types, and adding tests for the new functionality.

### Enhancements to the `Fabric` dialect:

* **Unix timestamp transformation**: Added a `_unix_to_time_sql` method to transform `UnixToTime` expressions into the Fabric-specific `DATEADD` syntax, which converts Unix timestamps into datetime values with microsecond precision.
* **Precision handling**: Updated `datatype_sql` to cap precision at 6 for certain data types and added a fallback to default precision of 6 for non-integer precision values.

### Test coverage improvements:

* **New test for Unix timestamp transformation**: Added a `test_unix_to_time` method to validate the transformation of `UNIX_TO_TIME(column)` into the correct `DATEADD` syntax.…nd add corresponding test

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -5,6 +5,40 @@ from sqlglot.dialects.dialect import NormalizationStrategy
 from sqlglot.dialects.tsql import TSQL
 
 
+def _unix_to_time_sql(self: "Fabric.Generator", expression: exp.UnixToTime) -> str:
+    """
+    Transform UnixToTime to Fabric-specific DATEADD syntax.
+
+    Fabric uses: DATEADD(MICROSECONDS, CAST(ROUND(column*1e6, 0) AS BIGINT), CAST('1970-01-01' AS DATETIME2(6)))
+    """
+    timestamp = expression.this
+
+    # Convert unix timestamp (seconds) to microseconds and round to avoid decimals
+    microseconds = exp.Cast(
+        this=exp.Round(
+            this=exp.Mul(this=timestamp, expression=exp.Literal.number("1e6")),
+            decimals=exp.Literal.number("0"),
+        ),
+        to=exp.DataType(this=exp.DataType.Type.BIGINT),
+    )
+
+    # Create the base datetime as '1970-01-01' cast to DATETIME2(6)
+    epoch_start = exp.Cast(
+        this=exp.Literal.string("1970-01-01"),
+        to=exp.DataType(
+            this=exp.DataType.Type.DATETIME2,
+            expressions=[exp.DataTypeParam(this=exp.Literal.number("6"))],
+        ),
+    )
+
+    # Create DATEADD expression
+    dateadd = exp.DateAdd(
+        this=epoch_start, expression=microseconds, unit=exp.Literal.string("MICROSECONDS")
+    )
+
+    return self.sql(dateadd)
+
+
 class Fabric(TSQL):
     """
     Microsoft Fabric Data Warehouse dialect that inherits from T-SQL.
@@ -76,6 +110,9 @@ class Fabric(TSQL):
                     # Cap precision at 6
                     current_precision = precision.this.to_py()
                     target_precision = min(current_precision, 6)
+                else:
+                    # If precision exists but is not an integer, default to 6
+                    target_precision = 6
 
                 # Create a new expression with the target precision
                 new_expression = exp.DataType(
@@ -86,3 +123,8 @@ class Fabric(TSQL):
                 return super().datatype_sql(new_expression)
 
             return super().datatype_sql(expression)
+
+        TRANSFORMS = {
+            **TSQL.Generator.TRANSFORMS,
+            exp.UnixToTime: _unix_to_time_sql,
+        }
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/fabric.py
```diff
@@ -5,6 +5,40 @@
 from sqlglot.dialects.tsql import TSQL
 
 
+def _unix_to_time_sql(self: "Fabric.Generator", expression: exp.UnixToTime) -> str:
```
georgesittas, line 8: You can avoid the forward ref here:

```suggestion
def _unix_to_time_sql(self: Fabric.Generator, expression: exp.UnixToTime) -> str:
```

sqlglot/dialects/fabric.py
```diff
@@ -5,6 +5,40 @@
...
+    )
+
+    # Create the base datetime as '1970-01-01' cast to DATETIME2(6)
+    epoch_start = exp.Cast(
+        this=exp.Literal.string("1970-01-01"),
+        to=exp.DataType(
+            this=exp.DataType.Type.DATETIME2,
+            expressions=[exp.DataTypeParam(this=exp.Literal.number("6"))],
+        ),
+    )
```
georgesittas, line 32:
```suggestion
    epoch_start = exp.cast("'1970-01-01'", "datetime2(6)", dialect="fabric")
```

sqlglot/dialects/fabric.py
```diff
@@ -5,6 +5,40 @@
...
+    timestamp = expression.this
+
+    # Convert unix timestamp (seconds) to microseconds and round to avoid decimals
+    microseconds = exp.Cast(
+        this=exp.Round(
+            this=exp.Mul(this=timestamp, expression=exp.Literal.number("1e6")),
+            decimals=exp.Literal.number("0"),
+        ),
+        to=exp.DataType(this=exp.DataType.Type.BIGINT),
+    )
```
georgesittas, line 23: Keep in mind that `UnixToTime` can have a scale, e.g. see how BigQuery [implements](https://github.com/tobymao/sqlglot/blob/8de87e3f755a40b600aa94ee2c30cf697ef7c43c/sqlglot/dialects/bigquery.py#L196-L210) its generator.

What's the motivation behind this addition? What problem (e.g., transpilation) are you trying to solve?

mattiasthalen: I'm trying to solve the duckdb `TO_TIMESTAMP(seconds_after_epoch)`

georgesittas: I see, thanks for clarifying. Have you tested this against Fabric? Also curious if there are any sources related to this workaround. I gave this a quick try in SQL Server and it didn't work that well. For example:

```
1> SELECT DATEADD(MICROSECOND, CAST(ROUND(1284352323.5 * 1e6, 0) AS BIGINT), CAST('1970-01-01' AS DATETIME2(6)))
2> GO
Msg 8115, Level 16, State 2, Server 7a7eca1deaa8, Line 1
Arithmetic overflow error converting expression to data type int.
```

Fabric may be different, though..

mattiasthalen: Yes, that's how I'm running it now using a python macro in sqlmesh:
image

georgesittas: Fair enough, this seems fine then. Just one suggestion then: let's try to handle different scales too. If `scale` is not set, then it defaults to seconds and what you have is correct. Otherwise, we can either handle it by changing the power of ten accordingly, or avoid dealing with it by using `self.unsupported`.

mattiasthalen: It might be something to solve for SQL Server as well. Something like separating the fractions and do a separate DATEADD for Milli/Micro/Nano. And then let fabric inherit it.

georgesittas: Sure, we can do it. Just keep the above limitation in mind, i.e., that this approach can easily result in overflows. We can also just keep the scope small and implement it just for Fabric for the time being. Up to you.

georgesittas:
Hiya @mattiasthalen 👋 

What's still in progress for this PR? Happy to help get it to the finish line.

mattiasthalen:
> Hiya @mattiasthalen 👋 
> 
> 
> 
> What's still in progress for this PR? Happy to help get it to the finish line.

I haven't had the time to do more. But what we said was to add the "power of 10" check (s, ms, us, ns). And perhaps split the different units to avoid int overflow. And possibly implement it for mssql.

Not sure when I'll have more time to work on it though.

georgesittas:
Cool, no worries ☺️
I'll see if I can wrap this up then.

georgesittas:
I tried to clean it up a little here: https://github.com/tobymao/sqlglot/commit/57c0e5dcb52a3cccb07f779ceb3da66779c18f8f. Don't have bandwidth to implement & test the logic to handle the other unixtotime scales now.

PR #5270, opened 2025-06-24
feat(exasol): added datatype mappings and test for exasol dialect.

nnamdi16:
This involves adding data type mappings and test for exasol dialect


sqlglot/dialects/exasol.py
```diff
@@ -21,4 +21,28 @@ class Generator(generator.Generator):
...
+        def datatype_sql(self, expression: exp.DataType) -> str:
+            """
+            Exasol  does not support `DATETIME2` and `SMALLDATETIME` data type but converts it to `TIMESTAMP`
+            https://docs.exasol.com/db/latest/sql_references/data_types/datatypedetails.htm#DateTime.
+            """
+            kind = expression.args.get("kind")
+            if kind and kind.upper() in {"DATETIME2", "SMALLDATETIME"}:
+                expression.set("this", exp.DataType.Type.TIMESTAMP)
+
+            return super().datatype_sql(expression)
```
georgesittas: This doesn't look right, we already map `Type.DATETIME2` and `Type.SMALLDATETIME` to `"TIMESTAMP"` in `TYPE_MAPPING`, right? These should suffice.

Additionally, `is_type(...)` is preferred in similar overrides of `datatype_sql`.

sqlglot/dialects/exasol.py
```diff
@@ -21,4 +21,28 @@ class Generator(generator.Generator):
         TYPE_MAPPING = {
             **generator.Generator.TYPE_MAPPING,
             **STRING_TYPE_MAPPING,
+            # https://docs.exasol.com/db/latest/sql_references/data_types/datatypealiases.htm
+            exp.DataType.Type.TINYINT: "SMALLINT",
+            exp.DataType.Type.INT: "INTEGER",
```
georgesittas: This isn't necessary, `INT` and `INTEGER` are aliases to the same thing.

sqlglot/dialects/exasol.py
```diff
@@ -21,4 +21,28 @@ class Generator(generator.Generator):
...
+            exp.DataType.Type.MEDIUMINT: "INTEGER",
```
georgesittas:
```suggestion
            exp.DataType.Type.MEDIUMINT: "INT",
```

sqlglot/dialects/exasol.py
```diff
@@ -21,4 +21,28 @@ class Generator(generator.Generator):
...
+            exp.DataType.Type.DOUBLE: "DOUBLE PRECISION",
```
georgesittas: This isn't necessary either, let's get rid of it and use `DOUBLE` which is shorter.

sqlglot/dialects/exasol.py
```diff
@@ -21,4 +21,28 @@ class Generator(generator.Generator):
...
+            exp.DataType.Type.DECIMAL32: "DECIMAL",
+            exp.DataType.Type.DECIMAL64: "DECIMAL",
+            exp.DataType.Type.DECIMAL128: "DECIMAL",
+            exp.DataType.Type.DECIMAL256: "DECIMAL",
+            exp.DataType.Type.DATETIME: "TIMESTAMP",
+            exp.DataType.Type.DATETIME2: "TIMESTAMP",
+            exp.DataType.Type.SMALLDATETIME: "TIMESTAMP",
+            exp.DataType.Type.TIMESTAMPLTZ: "TIMESTAMP WITH LOCAL TIME ZONE",
```
georgesittas, line 32: What about `TIMESTAMP(3) WITH LOCAL TIME ZONE`, is this handled properly with this mapping? My guess is not. Take a look at `TZ_TO_WITH_TIME_ZONE`– we may need new logic to handle this.

tests/dialects/test_exasol.py
```diff
@@ -17,3 +17,23 @@ def test_type_mappings(self):
         self.validate_identity("CAST(x AS VARBINARY)", "CAST(x AS VARCHAR)")
         self.validate_identity("CAST(x AS VARCHAR)", "CAST(x AS VARCHAR)")
         self.validate_identity("CAST(x AS CHAR)", "CAST(x AS CHAR)")
+        self.validate_identity("CAST(x AS TINYINT)", "CAST(x AS SMALLINT)")
+        self.validate_identity("CAST(x AS SMALLINT)", "CAST(x AS SMALLINT)")
```
georgesittas: Nit: when you have `self.validate_identity(x, x)`, you can do `self.validate_identity(x)`.

sqlglot/dialects/exasol.py
nnamdi16: Based on the syntax tree. Its been mapped to a user defined function as seen below:
`Cast(
  this=Column(
  to=DataType(this=Type.USERDEFINED, kind=SMALLDATETIME))`
  
  This makes the already done mapping to fail as well as the is_type method not to work because is_type is now an instance of type USERDEFINED. That was why I had to override datatype_sql
  
  @georgesittas

sqlglot/dialects/exasol.py
nnamdi16: You're right I would remove it  @georgesittas

tests/dialects/test_exasol.py
nnamdi16: Noted

sqlglot/dialects/exasol.py
georgesittas: The syntax tree returns a `USERDEFINED` type because the "read" dialect does not properly support it. If you parse it using `tsql` (you can use `validate_all` & specify it within the `read` dict), you'll observe that it's generated correctly without requiring this override.

nnamdi16: Alright I would look into it and revert back

sqlglot/dialects/exasol.py
nnamdi16: @georgesittas It covers `TIMESTAMP(3) WITH LOCAL TIME ZONE` and handles it properly. Can you please throw more light on `TZ_TO_WITH_TIME_ZONE` and how it relates with this mapping.

sqlglot/dialects/exasol.py
nnamdi16: Fixed

sqlglot/dialects/exasol.py
nnamdi16: Fixed

sqlglot/dialects/exasol.py
nnamdi16: Fixed

tests/dialects/test_exasol.py
nnamdi16: Fixed

sqlglot/dialects/exasol.py
```diff
@@ -1,9 +1,17 @@
...
 
 
 class Exasol(Dialect):
+    class Tokenizer(tokens.Tokenizer):
+        KEYWORDS = {
+            **tokens.Tokenizer.KEYWORDS,
+            "DATETIME2": TokenType.DATETIME2,
+            "SMALLDATETIME": TokenType.SMALLDATETIME,
+        }
+
```
georgesittas: Why was this introduced? I was under the impression that these types aren't supported in Exasol.

sqlglot/dialects/exasol.py
georgesittas: Ah, makes sense after all, thanks for checking. Let's add a test please for the `TIMESTAMP(3) WITH LOCAL TIME ZONE` case please!

nnamdi16: Alright

sqlglot/dialects/exasol.py
nnamdi16: I have updated it to be mapped to TIMESTAMP either unless Syntax tree would be 
`Cast( this=Column( this=Identifier(this=x, quoted=False)), to=DataType(this=Type.USERDEFINED, kind=SMALLDATETIME))`

sqlglot/dialects/exasol.py
nnamdi16: I have added the test TIMESTAMP(3) WITH LOCAL TIME ZONE which required overriding datatype_sql

sqlglot/dialects/exasol.py
georgesittas: I think there's a disconnect here. Since Fabric does not support `DATETIME2` and `SMALLDATETIME` natively, we should not have to tokenize these keywords differently. The fact that it's parsed as a `USERDEFINED` type otherwise doesn't matter, because it's not a supported type anyway. If you want to test the transpilation from, say, T-SQL, you need to use `validate_all` with the `read` kwarg in your tests.

tests/dialects/test_exasol.py
```diff
@@ -17,3 +17,27 @@ def test_type_mappings(self):
...
+        self.validate_identity("CAST(x AS DATETIME2)", "CAST(x AS TIMESTAMP)")
+        self.validate_identity("CAST(x AS SMALLDATETIME)", "CAST(x AS TIMESTAMP)")
+        self.validate_identity("CAST(x AS BOOLEAN)")
+        self.validate_identity(
+            "CAST(x AS TIMESTAMPLTZ)", "CAST(x AS TIMESTAMP WITH LOCAL TIME ZONE)"
+        )
+        self.validate_identity(
+            "CAST(x AS TIMESTAMP(3) WITH LOCAL TIME ZONE)",
+            "CAST(x AS TIMESTAMP WITH LOCAL TIME ZONE)",
+        )
```
georgesittas, line 59: We shouldn't drop the precision of the timestamp.

sqlglot/dialects/exasol.py
```diff
@@ -21,4 +29,30 @@ class Generator(generator.Generator):
...
+            Exasol supports a fixed default precision for TIMESTAMP WITH LOCAL TIME ZONE and does not allow specifying a custom precision.
+            This method strips any precision argument from TIMESTAMPLTZ types to ensure compatibility with Exasol's syntax.
+            """
+            if expression.is_type(exp.DataType.Type.TIMESTAMPLTZ):
+                return super().datatype_sql(
+                    exp.DataType(
+                        this=expression.this,
+                        expressions=[],
+                    )
+                )
```
georgesittas: This doesn't look right, why was this change necessary?

sqlglot/dialects/exasol.py
nnamdi16: I have fixed the issue and added the test using `validate_all`

PR #5277, opened 2025-06-24
feat(dialects): add Dremio dialect

mateuszpoleski:
Initial PR introducing Dremio dialect

- Adds base skeleton for the dialect
- Implements Dremio-specific type mappings
- Configures main Dremio-specific capabilities

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -1,0 +1,46 @@
+from sqlglot import expressions as exp
+from sqlglot import parser, generator, tokens
+from sqlglot.dialects.dialect import Dialect
+
+
+class Dremio(Dialect):
+    SUPPORTS_USER_DEFINED_TYPES = False
+    CONCAT_COALESCE = True  # CONCAT('a', NULL) = 'a', not NULL
+    NORMALIZE_FUNCTIONS = False
+    PRESERVE_ORIGINAL_NAMES = True
+    TYPED_DIVISION = True  # Type of division result is determined by the types of the operands
+    SUPPORTS_SEMI_ANTI_JOIN = False
+    NULL_ORDERING = "nulls_are_last"
+    TABLESAMPLE_SIZE_IS_PERCENT = True
+    SUPPORTS_VALUES_DEFAULT = False  # Dremio uses NULL instead
+
+    class Parser(parser.Parser):
+        LOG_DEFAULTS_TO_LN = True  # LOG(x) with no base defaults to natural log
+
+    class Generator(generator.Generator):
+        NVL2_SUPPORTED = False
+        SUPPORTS_CONVERT_TIMEZONE = True
+        INTERVAL_ALLOWS_PLURAL_FORM = False
+        JOIN_HINTS = False
+        LIMIT_ONLY_LITERALS = True
+        SUPPORTS_SINGLE_ARG_CONCAT = False
+        MULTI_ARG_DISTINCT = False
+
+        # https://docs.dremio.com/current/reference/sql/data-types/
+        TYPE_MAPPING = {
+            **generator.Generator.TYPE_MAPPING,
+            exp.DataType.Type.SMALLINT: "INT",
+            exp.DataType.Type.TINYINT: "INT",
+            exp.DataType.Type.BINARY: "VARBINARY",
+            exp.DataType.Type.TEXT: "VARCHAR",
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/dremio.py
```diff
@@ -0,0 +1,46 @@
+from sqlglot import expressions as exp
+from sqlglot import parser, generator, tokens
+from sqlglot.dialects.dialect import Dialect
+
+
+class Dremio(Dialect):
+    SUPPORTS_USER_DEFINED_TYPES = False
+    CONCAT_COALESCE = True  # CONCAT('a', NULL) = 'a', not NULL
+    NORMALIZE_FUNCTIONS = False
```
georgesittas, line 9: Why is this set to `False`?

sqlglot/dialects/dremio.py
```diff
@@ -0,0 +1,46 @@
+from sqlglot import expressions as exp
+from sqlglot import parser, generator, tokens
+from sqlglot.dialects.dialect import Dialect
+
+
+class Dremio(Dialect):
+    SUPPORTS_USER_DEFINED_TYPES = False
+    CONCAT_COALESCE = True  # CONCAT('a', NULL) = 'a', not NULL
```
georgesittas:
```suggestion
    CONCAT_COALESCE = True
```

sqlglot/dialects/dremio.py
```diff
@@ -0,0 +1,46 @@
...
+    NORMALIZE_FUNCTIONS = False
+    PRESERVE_ORIGINAL_NAMES = True
```
georgesittas, line 10: Why should we preserve original names? This is mostly relevant to functions with UDFs and case-sensitive function names like BigQuery and ClickHouse

sqlglot/dialects/dremio.py
```diff
@@ -0,0 +1,46 @@
...
+    TYPED_DIVISION = True  # Type of division result is determined by the types of the operands
```
georgesittas:
```suggestion
    TYPED_DIVISION = True
```

sqlglot/dialects/dremio.py
```diff
@@ -0,0 +1,46 @@
...
+        TYPE_MAPPING = {
+            **generator.Generator.TYPE_MAPPING,
+            exp.DataType.Type.SMALLINT: "INT",
+            exp.DataType.Type.TINYINT: "INT",
+            exp.DataType.Type.BINARY: "VARBINARY",
+            exp.DataType.Type.TEXT: "VARCHAR",
+            exp.DataType.Type.NCHAR: "VARCHAR",
+            exp.DataType.Type.CHAR: "VARCHAR",
+            exp.DataType.Type.TIMESTAMPLTZ: "TIMESTAMP",
+            exp.DataType.Type.TIMESTAMPTZ: "TIMESTAMP",
```
georgesittas: I'd rather we override `datatype_sql` and call `self.unsupported` when the data type is one of these two. That way you won't silently mistranspile time-zone aware types to Dremio, [which does not support them](https://docs.dremio.com/current/reference/sql/data-types/time-zone-support/).

Let's also map `TIMESTAMPNTZ` to `"TIMESTAMP"` while we're at it.

georgesittas (commented):
Thanks for the PR @mateuszpoleski, I've left some comments.

Let's please add more tests to increase the coverage of the dialect. Ideally, most if not all of the flags you set should be tested.

georgesittas:
Hey @mateuszpoleski 👋 
Do you plan to continue working on this PR soon?

mateuszpoleski:
Hey, thanks for the review! I'm a bit busy currently, but should be able to make the changes in the next 2-3 days

georgesittas:
Sounds good! Thanks for following up :)

sqlglot/dialects/dremio.py
mateuszpoleski: Similarly to `NORMALIZE_FUNCTIONS`, my idea was to set these flags such that query after parsing would be as close as possible to the original one. Likely I misunderstood what this flag does, will remove to keep value as default

sqlglot/dialects/dremio.py
mateuszpoleski: Good points!

mateuszpoleski:
Thanks again for the review, I've added more tests and addressed other comments :)

georgesittas (approved):
Thank you! Looks good.

Issue #5271, opened 2025-06-24
Bigquery -> Spark: unnest with offset transpiled wrong


charlie-liner:
**Fully reproducible code snippet**
I wanna transpile bigquery `unnest with offset` query into spark query. but It doesn't work the way I expected it to

```
transpiled_sql = sqlglot.transpile(
    model_sql,
    read='bigquery',
    write='spark',
    pretty=True
)[0]
print(transpiled_sql)
```

original bigquery query
```
SELECT
    *
FROM ANSWERS,
UNNEST(JSON_QUERY_ARRAY(`METADATA`, '$.references')) AS ref WITH OFFSET AS pos
```

expected
```
SELECT
  *
FROM answers
LATERAL VIEW
POSEXPLODE(JSON_EXTRACT_ARRAY(`metadata`, '$.references')) AS ref, pos
```

but what sqlglot return
```
SELECT
  *
FROM answers
LATERAL VIEW
POSEXPLODE(JSON_EXTRACT_ARRAY(`metadata`, '$.references')) AS ref
```

**Official Documentation**
Please include links to official SQL documentation related to your issue.

PR #5278, opened 2025-06-25
fix: change posexplode alias order when transpile bigquery unnest with offset

kgw7401:
fix: #5274 
related: #5271 

I raised the issue about transpiling bigquery unnest with offset into spark posexplode correctly and #5274 pr resolved that. 
but i found that there is alias ordering problem in transpiled result so i fix it!

original query
```
SELECT * FROM tbl CROSS JOIN UNNEST(col) AS ref WITH OFFSET AS {alias}
```

before:
```
SELECT * FROM tbl LATERAL VIEW POSEXPLODE(col) AS ref, {alias}
```

after:
```
SELECT * FROM tbl LATERAL VIEW POSEXPLODE(col) AS {alias}, ref
```

### DIFF
--- a/sqlglot/transforms.py
+++ b/sqlglot/transforms.py
@@ -356,7 +356,9 @@ def unnest_to_explode(
             offset = unnest.args.get("offset")
             if offset:
                 # For posexplode, offset should come first, then the value column
-                offset_col = offset if isinstance(offset, exp.Identifier) else exp.to_identifier("pos")
+                offset_col = (
+                    offset if isinstance(offset, exp.Identifier) else exp.to_identifier("pos")
+                )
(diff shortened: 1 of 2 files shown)


charlie-liner:
@georgesittas i don't why lint is failed😂 could please review my pr?

kgw7401:
@georgesittas i don't why lint is failed😂 could please review my pr?

VaggelisD:
Hey @kgw7401, really sorry about this, I had noticed the issue myself but thought I had pushed the fix in my previous PR.

Closing this PR in favor of https://github.com/tobymao/sqlglot/pull/5279 to get this to the finish line faster.

PR #5279, opened 2025-06-25
fix(hive): Fix exp.PosExplode alias order

VaggelisD:
Follow up fix for https://github.com/tobymao/sqlglot/pull/5278


kgw7401:
May i ask when will this fix be distributed as a new release?

PR #5280, opened 2025-06-25
fix(fabric): Type mismatches and precision error

mattiasthalen:
Corrects type mappings in the Fabric dialect to align with Fabric's data type support.

- Addresses inconsistencies between T-SQL and Fabric's data type handling.
- Ensures that temporal types (TIME, DATETIME2, DATETIMEOFFSET) precision is capped at 6 digits.
- Adds UTINYINT keyword mapping.

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from sqlglot import exp
 from sqlglot.dialects.dialect import NormalizationStrategy
 from sqlglot.dialects.tsql import TSQL
+from sqlglot.tokens import TokenType
 
 
 class Fabric(TSQL):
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/fabric.py
```diff
@@ -28,29 +29,42 @@ class Fabric(TSQL):
...
+            **TSQL.Tokenizer.KEYWORDS,
+            "TIMESTAMP": TokenType.TIMESTAMP,  # Override T-SQL's mapping of TIMESTAMP to ROWVERSION
+            "UTINYINT": TokenType.UTINYINT,  # Add UTINYINT keyword that T-SQL is missing
+        }
+
     class Generator(TSQL.Generator):
         # Fabric-specific type mappings - override T-SQL types that aren't supported
         # Reference: https://learn.microsoft.com/en-us/fabric/data-warehouse/data-types
         TYPE_MAPPING = {
-            **TSQL.Generator.TYPE_MAPPING,
```
georgesittas, line 35: Why aren't we inheriting from T-SQL anymore?

sqlglot/dialects/fabric.py
```diff
@@ -60,29 +74,35 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             Fabric limits temporal types (TIME, DATETIME2, DATETIMEOFFSET) to max 6 digits precision.
             When no precision is specified, we default to 6 digits.
             """
+            # First apply the parent type mapping to get the correct base type
+            result = super().datatype_sql(expression)
+
+            # Check if this is a temporal type that needs precision handling
             if expression.is_type(
```
georgesittas: Let's change this to `if expression.is_type(*exp.DataType.TEMPORAL_TYPES)` instead.

sqlglot/dialects/fabric.py
```diff
@@ -60,29 +74,35 @@ def datatype_sql(self, expression: exp.DataType) -> str:
...
-                new_expression = exp.DataType(
-                    this=expression.this,
-                    expressions=[exp.DataTypeParam(this=exp.Literal.number(target_precision))],
-                )
+                # Extract the base type from the result and add precision
+                if "(" in result:
+                    # Type already has parameters, extract just the type name
+                    base_type = result.split("(")[0]
+                else:
+                    base_type = result
```
georgesittas: Why did the `datatype_sql` call get moved at the beginning vs what we used to do? I don't like this approach, string operations are hacky and brittle.

sqlglot/dialects/fabric.py
mattiasthalen: I thought it was needed due to the collisions with data types, e.g. timestamp > rowversion. But I tried adding it back and it looks like it works now. So I'll revert this change.

sqlglot/dialects/fabric.py
mattiasthalen: With the previous version I got this error:
```bash
AssertionError: 'CAST(x AS DATETIME2(6))' != 'CAST(x AS DATETIME2(6)(6))'
- CAST(x AS DATETIME2(6))
+ CAST(x AS DATETIME2(6)(6))
```
And this worked...

But after further investigation, it seems like this was actually due to the hardcoded precision in `TYPE_MAPPING`

georgesittas (approved):
Thanks! There may be a few redundant mappings in the generator (e.g., `BOOLEAN -> "BIT"` is covered by T-SQL), but I'll get this to the finish line for you.

PR #5286, opened 2025-06-25
Fix(parser): avoid bubbling up Dot comment

georgesittas:
Addresses the issue discussed in this Slack thread: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1750873457473389. TL;DR, comment is moved outside of the `Identifier` node but it carries additional semantics so it's incorrect.


georgesittas:
Hm, I'll hold off merging this for a bit. Need to verify the behavior introduced here is reasonable and doesn't interact weirdly with SQLMesh's column description pushdown.

PR #5289, opened 2025-06-26
fix: Supporting two argument flavor of REPLACE function in Snowflake and Presto dialects

rahulj51:
Both Snwoflake and Presto support a two argument version of REPLACE with the third argument being optional. This breaks during transpilation to other dialects and needs to be handled. This fix adds the default value for the third argument for these two dialects.  

Also added unit tests for validations.


tests/fixtures/identity.sql
```diff
@@ -171,7 +171,8 @@ SPLIT('new york', '.')
 X((y AS z)).1
 X(a.b = 1)
 (x AS y, y AS z)
-REPLACE(1)
```
rahulj51, line 174: I don't know which dialect this is from but based on my knowledge, none of the known dialects support REPLACE with a single argument.

tests/dialects/test_snowflake.py
```diff
@@ -2223,6 +2223,35 @@ def test_regexp_replace(self, logger):
             },
         )
 
+    @mock.patch("sqlglot.generator.logger")
```
georgesittas, line 2226: Why was this needed?

sqlglot/dialects/snowflake.py
```diff
@@ -166,6 +166,15 @@ def _build_regexp_replace(args: t.List) -> exp.RegexpReplace:
     return regexp_replace
 
 
+def _build_replace(args: t.List) -> exp.Replace:
```
georgesittas, line 169: Isn't the logic the same as Presto? Why do something different for Snowflake? I'd suggest extracting the Presto logic out into a helper within `dialect.py` and then importing it in both dialects, so that it can be reused elsewhere if needed.

georgesittas (approved):
This looks good, thank you @rahulj51! Just a suggestion & a question.

georgesittas:
I'll get this in and take it to the finish line, thanks!

rahulj51:
> I'll get this in and take it to the finish line, thanks!

Perfect. Thanks.

georgesittas:
https://github.com/tobymao/sqlglot/commit/aa5f8065f53013e1554885c50f6c121cbd05d103

PR #5291, opened 2025-06-26
Fix(bigquery)!: only coerce time var -like units into strings for date trunc

georgesittas:
BigQuery preemptively coerces all unit expressions into string literals for `DATE_TRUNC` today:

```python
>>> import sqlglot
>>> sqlglot.parse_one("date_trunc(x, @foo)", "bigquery")
DateTrunc(
  unit=Literal(this='@FOO', is_string=True),
  this=Column(
    this=Identifier(this=x, quoted=False)))
>>> sqlglot.parse_one("date_trunc(x, @foo)", "bigquery").sql("bigquery")
'DATE_TRUNC(x, @FOO)'
```

This messes up the AST and SQLMesh is unable to render macros.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -543,7 +543,7 @@ class BigQuery(Dialect):
             "DATE_ADD": build_date_delta_with_interval(exp.DateAdd),
             "DATE_SUB": build_date_delta_with_interval(exp.DateSub),
             "DATE_TRUNC": lambda args: exp.DateTrunc(
-                unit=exp.Literal.string(str(seq_get(args, 1))),
+                unit=seq_get(args, 1),
                 this=seq_get(args, 0),
                 zone=seq_get(args, 2),
             ),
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -543,7 +543,7 @@ class Parser(parser.Parser):
             "DATE_ADD": build_date_delta_with_interval(exp.DateAdd),
             "DATE_SUB": build_date_delta_with_interval(exp.DateSub),
             "DATE_TRUNC": lambda args: exp.DateTrunc(
-                unit=exp.Literal.string(str(seq_get(args, 1))),
+                unit=seq_get(args, 1),
```
georgesittas, line 546: This is no longer necessary because `DateTrunc` coerces the value in its constructor ([source](https://github.com/tobymao/sqlglot/blob/ba4a234bfabdd8161b96a29436a50e0eb04c2dc2/sqlglot/expressions.py#L5841-L5856)).

sqlglot/expressions.py
```diff
@@ -5850,8 +5850,6 @@ def __init__(self, **args):
                 unit_name = TimeUnit.UNABBREVIATED_UNIT_NAME[unit_name]
 
             args["unit"] = Literal.string(unit_name)
-        elif isinstance(unit, Week):
-            unit.set("this", Literal.string(unit.this.name.upper()))
```
georgesittas, line 5854: I don't get what this was achieving, it looks like it could've been an oversight. Commit that added it for reference [here](https://github.com/tobymao/sqlglot/commit/a4e899aaba317028480a440faeddb026bb464890#diff-7857fedd1d1451b1b9a5b8efaa1cc292c02e7ee4f0d04d7e2f9d5bfb9565802c).

VaggelisD (approved):

PR #5305, opened 2025-06-27
feat(exasol): added `EVERY` built in function

nnamdi16:
This involves adding [EVERY](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/every.htm) built -in function to exasol dialect in sqlglot.

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -14,6 +14,7 @@ class Exasol(Dialect):
             "BIT_NOT": lambda args: exp.BitwiseNot(this=seq_get(args, 0)),
             "BIT_LSHIFT": binary_from_function(exp.BitwiseLeftShift),
             "BIT_RSHIFT": binary_from_function(exp.BitwiseRightShift),
+            "EVERY": lambda args: exp.All(this=seq_get(args, 0)),
         }
 
     class Generator(generator.Generator):
--- a/tests/dialects/test_exasol.py
+++ b/tests/dialects/test_exasol.py
@@ -171,6 +171,13 @@ class TestExasol(Validator):
         )
 
     def test_aggregateFunctions(self):
-        self.validate_identity(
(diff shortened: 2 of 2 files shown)


sqlglot/dialects/exasol.py
```diff
@@ -65,6 +65,8 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             exp.BitwiseRightShift: rename_func("BIT_RSHIFT"),
             # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/bit_xor.htm
             exp.BitwiseXor: rename_func("BIT_XOR"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/every.htm
+            exp.All: rename_func("EVERY"),
```
georgesittas, line 70: Note that Exasol does not parse `EVERY` into `All` today. You need to modify `SUBQUERY_PREDICATES` to include it, which means you also need a token that you'll map `"EVERY"` to in Exasol specifically.

After the above items are completed, please improve the testing coverage by including a transpilation test with `validate_all` as well, instead of just checking the roundtrip.

tests/dialects/test_exasol.py
```diff
@@ -169,3 +169,8 @@ def test_bits(self):
                 "spark": "SELECT SHIFTRIGHT(x, 1)",
             },
         )
+
+    def test_aggregateFunctions(self):
+        self.validate_identity(
+            "SELECT department, EVERY(age >= 30) AS EVERY FROM employee_table GROUP BY department"
+        )
```
georgesittas: This currently works because `EVERY` is parsed into an `Anonymous` node, which is generated as-is.

sqlglot/dialects/exasol.py
nnamdi16: Alright

PR #5306, opened 2025-06-27
feat(parser)!: parse, annotate type for ARRAY_REVERSE

geooo109:
This PR adds support for the parsing of ARRAY_REVERSE and the type annotation of it.

**DOCS**  
[BigQuery ARRAY_REVERSE](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#array_reverse)  
[Snowflake ARRAY_REVERSE](https://docs.snowflake.com/en/sql-reference/functions/array_reverse)  
[DuckDB ARRAY_REVERSE](https://duckdb.org/docs/stable/sql/functions/list.html#list_reverselist)  
[Clickhouse arrayReverse](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayReverse)


sqlglot/expressions.py
```diff
@@ -5577,6 +5577,10 @@ class ArrayLast(Func):
     pass
 
 
+class ArrayReverse(Func):
+    _sql_names = ["ARRAY_REVERSE", "REVERSE"]
```
VaggelisD: We should be very careful with the use of `_sql_names`, it should only be used if the function names are almost identical between dialects e.g `_sql_names = ["ARGMAX", "ARG_MAX"]`

Note how in BigQuery [REVERSE](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#reverse) is a string function, as is the case with other dialects too probably.

geooo109: Yep, I see, thank you. I will remove the `REVERSE` thing for now and keep the `ArrayReverse` only.

georgesittas (approved):

georgesittas (approved):

Issue #4952, opened 2025-04-07
[Clickhouse] Support refreshable materialized view syntax


vucong2409:
Clickhouse recently introduced a Refreshable materialized view, which includes some additional keywords/properties:
- `REFRESH EVERY`
- `DEPENDS ON`
- `APPEND TO`

`sqlglot` currently does not support those keywords & will lead to it failing to parse the whole query.

![Image](https://github.com/user-attachments/assets/a2c44d50-d17a-43ec-8636-cc753d3057c7)

**Fully reproducible code snippet**
```
CREATE MATERIALIZED VIEW test_db.rmv ON CLUSTER '{cluster}' REFRESH EVERY 5 MINUTE DEPENDS ON test_db.src_rmv APPEND TO test_db.dest_table 
SELECT
    *
FROM
    test_db.test_src_table
```

**Official Documentation**
https://clickhouse.com/docs/sql-reference/statements/create/view#refreshable-materialized-view

vucong2409:
I'm willing to open a PR for this issue, but I don't know how should I modify the current codebase. Should I include those properties under the current `MaterializedProperty()` like this (create a method called `parse_materialized` instead of using generic `parse_property` method)

```
properties=Properties(
    expressions=[
         MaterializedProperty(
              RefreshRate(
                     this=5
                     duration=s
               )
              Depends=(
                      this=Table(...)
               )
      )
    ]
)
```
or should I create separate properties for each of those keywords?

georgesittas:
Hey 👋
You would need to create separate properties for each method. For refreshing, there's already an `AutoRefreshProperty` that we could reuse to represent the concept:

```diff
  # https://docs.aws.amazon.com/prescriptive-guidance/latest/materialized-views-redshift/refreshing-materialized-views.html
+ # https://clickhouse.com/docs/sql-reference/statements/create/view#refreshable-materialized-view
  class AutoRefreshProperty(Property):
-  arg_types = {"this": True}
+  arg_types = {"this": False, "cadence": False, "offset": False}
```

You can use `_parse_interval` combined with `match_interval=False` to capture the interval values, because ClickHouse is not expecting an `INTERVAL` keyword. You'd need to make sure the keyword is not generated when roundtripping a query, too.

Feel free to take a stab at this and ask questions if needed. I'm closing this because it's low priority for the core team.

georgesittas:
Addressed by #7990

PR #5307, opened 2025-06-29
feat: add support for refreshable materialized view syntax

vucong2409:
Based on Clickhouse documentation: https://clickhouse.com/docs/sql-reference/statements/create/view#refreshable-materialized-view
Resolve https://github.com/tobymao/sqlglot/issues/4952

### DIFF
--- a/sqlglot/dialects/clickhouse.py
+++ b/sqlglot/dialects/clickhouse.py
@@ -504,6 +504,7 @@ class ClickHouse(Dialect):
         PROPERTY_PARSERS = {
             **parser.Parser.PROPERTY_PARSERS,
             "ENGINE": lambda self: self._parse_engine_property(),
+            "REFRESH": lambda self: self._parse_auto_refresh(),
         }
         PROPERTY_PARSERS.pop("DYNAMIC")
 
@@ -966,6 +967,22 @@ class ClickHouse(Dialect):
 
             return value
 
+        def _parse_auto_refresh(self) -> t.Optional[exp.AutoRefreshProperty]:
+            kwargs = {}
+
+            if not self._match_texts("EVERY"):
+                self._retreat(self._index - 1)
+                return None
+            kwargs["this"] = self._parse_interval(match_interval=False)
+
+            if self._match_text_seq("DEPENDS", "ON"):
+                kwargs["depend"] = self._parse_table_parts()  # type: ignore
+
+            if self._match_texts("APPEND"):
+                kwargs["append"] = True  # type: ignore
+
+            return self.expression(exp.AutoRefreshProperty, **kwargs)  # type: ignore
+
     class Generator(generator.Generator):
         QUERY_HINTS = False
         STRUCT_DELIMITER = ("(", ")")
@@ -1069,6 +1086,7 @@ class ClickHouse(Dialect):
             exp.ArgMax: arg_max_or_min_no_count("argMax"),
             exp.ArgMin: arg_max_or_min_no_count("argMin"),
             exp.Array: inline_array_sql,
+            exp.AutoRefreshProperty: lambda self, e: self.flatten_auto_refresh_prop(e),
             exp.CastToStrType: rename_func("CAST"),
             exp.CountIf: rename_func("countIf"),
             exp.CompressColumnConstraint: lambda self,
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/clickhouse.py
```diff
@@ -966,6 +967,22 @@ def _parse_value(self, values: bool = True) -> t.Optional[exp.Tuple]:
 
             return value
 
+        def _parse_auto_refresh(self) -> t.Optional[exp.AutoRefreshProperty]:
+            kwargs = {}
+
+            if not self._match_texts("EVERY"):
+                self._retreat(self._index - 1)
+                return None
```
georgesittas, line 975: You can also have `AFTER` here:

```
CREATE MATERIALIZED VIEW [IF NOT EXISTS] [db.]table_name [ON CLUSTER cluster]
REFRESH EVERY|AFTER interval [OFFSET interval]
...
```

It also appears that `OFFSET` is expected before the interval value. Can we please increase our testing coverage? It doesn't look like these are included in current testing.

At the very least, we need to preserve the existing behavior of falling back to producing a `Command` if we can't parse a DDL with these properties, otherwise if we error out that'd be treated as a regression.

sqlglot/dialects/clickhouse.py
```diff
@@ -1393,3 +1412,20 @@ def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> st
                 values_as_table = True
 
             return super().values_sql(expression, values_as_table=values_as_table)
+
+        def flatten_auto_refresh_prop(self, expression: exp.AutoRefreshProperty) -> str:
```
georgesittas, line 1416: Remove the entry from `TRANSFORMS` and rename this to `autorefreshproperty_sql`. Then it will be automatically associated with the corresponding expression.

sqlglot/dialects/clickhouse.py
```diff
@@ -1393,3 +1412,20 @@ def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> st
...
+            interval_str = self.flatten_interval(expression.this)
+
+            depends_on_clause = ""
+            if expression.args.get("depend"):
+                depends_on_table = expression.args.get("depend")
+                depends_on_clause = f" DEPENDS ON {depends_on_table}"
```
georgesittas, line 1422:
```suggestion
            depends_on = self.sql(expression, "depend")
            depends_on = f" DEPENDS ON {depends_on}" if depends_on else ""
```

sqlglot/dialects/clickhouse.py
```diff
@@ -1393,3 +1412,20 @@ def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> st
...
+
+            append_clause = ""
+            if expression.args.get("append"):
+                append_clause = " APPEND"
+
+            return f"REFRESH EVERY {interval_str}{depends_on_clause}{append_clause}"
+
+        def flatten_interval(self, expression: exp.Interval) -> str:
+            return f"{expression.this.this} {expression.unit}"
```
georgesittas, line 1431: Why was this needed? Is it because `Interval`'s generator includes the `INTERVAL` keyword? If that's the case, can we please control the behavior of `interval_sql` with a new kwarg or something to avoid rolling a new custom method?

Also, please refer to existing patterns in the generator. We intentionally don't interpolate expressions like you did here, because that silently invokes the `__str__` method of `Expression` which generates without a dialect. This can be problematic in some cases. Always use the explicit `self.sql()` call to generate SQL strings from expressions.

sqlglot/dialects/clickhouse.py
```diff
@@ -1393,3 +1412,20 @@ def values_sql(self, expression: exp.Values, values_as_table: bool = True) -> st
...
+            interval_str = self.flatten_interval(expression.this)
+
+            depends_on_clause = ""
+            if expression.args.get("depend"):
+                depends_on_table = expression.args.get("depend")
+                depends_on_clause = f" DEPENDS ON {depends_on_table}"
+
+            append_clause = ""
+            if expression.args.get("append"):
+                append_clause = " APPEND"
```
georgesittas, line 1426:
```suggestion
            append = " APPEND" if expression.args.get("append") else ""
```

sqlglot/dialects/clickhouse.py
```diff
@@ -1069,6 +1086,7 @@ class Generator(generator.Generator):
             exp.ArgMax: arg_max_or_min_no_count("argMax"),
             exp.ArgMin: arg_max_or_min_no_count("argMin"),
             exp.Array: inline_array_sql,
+            exp.AutoRefreshProperty: lambda self, e: self.flatten_auto_refresh_prop(e),
```
georgesittas, line 1089: Let's remove this.

georgesittas (commented):
Thanks for the PR @vucong2409.

georgesittas:
Hey @vucong2409, any plans to finish this PR?

georgesittas:
Closing this for now, feel free to open another PR when ready.

PR #5308, opened 2025-06-29
feat!: add support for SPACE

ankur334:
## What & Why

* **New AST node `Space`**  
  *inherits from `Func`; 1-line doc-string and `arg_types = {"this": True}`.*

* **Parser support** comes for free in Spark, Spark 3, Hive, Spark2, Databricks,
  DuckDB because each dialect starts from `Parser.FUNCTIONS`, which now
  includes `"SPACE": exp.Space.from_arg_list`.  
  Previously the token was parsed as `Anonymous`, so users could not
  transpile queries like `SELECT SPACE(10)` without losing semantic
 

* **BigQuery / Trino generator transform**  
  Rewrites the call to the canonical GoogleSQL form  
  **`REPEAT('  ', n)`** by materialising an `exp.Repeat` node under the hood,
  ensuring the output executes on BigQuery, which lacks a dedicated
  `SPACE` function. 

* **mapping**  
  DuckDB/Spark/Hive ships a native `SPACE` alias, so the generator prints
  `SPACE(n)` directly; no transform required. But in bigquery, trino, `space` is not supported but it can be written using `REPEAT(' ', N)`

**The change lets me (and everyone) do:**

```python
>>> sql = "SELECT SPACE(4)"
>>> transpile(sql, read="spark", write="bigquery")[0]
"SELECT REPEAT(' ', 4)"

### CODE REVIEW

File: sqlglot/dialects/bigquery.py [hunk 1]
```diff
@@ -1012,6 +1012,12 @@ class Generator(generator.Generator):
             ),
             exp.SHA: rename_func("SHA1"),
             exp.SHA2: sha256_sql,
+            exp.Space: lambda self, e: self.sql(
+                exp.Repeat(
+                    this=exp.Literal.string(" "),  # first arg  -> ' '
+                    times=e.this,  # second arg -> n
+                )
+            ),
```
VaggelisD: Since that is identical to the Presto/Trino we could move this into a helper in `dialects.py`, e.g:

```
def _space_sql(...):
 return ...

```

And import it in the relevant files. You can check out other `_sql` functions in `dialect.py` for reference

sqlglot/expressions.py
```diff
@@ -6798,6 +6798,14 @@ class FromBase(Func):
     arg_types = {"this": True, "expression": True}
 
 
+class Space(Func):
+    """
+    SPACE(n) → string consisting of n blank characters
+    """
+
+    arg_types = {"this": True}
```
VaggelisD: You can do `pass` here instead of `arg_types`, since `arg_types = {"this": True}` is inherited.

sqlglot/expressions.py
```diff
@@ -6798,6 +6798,14 @@ class FromBase(Func):
     arg_types = {"this": True, "expression": True}
 
 
+class Space(Func):
```
VaggelisD, line 6801: Just to sanity check, are we sure that dialects other than Hive/Spark/DBX don't use `SPACE(...)` for other purposes or allow more args etc?

ranjanankur314: Hi @VaggelisD 
I have followed the documentation and cross-verified it by running on different compute engine like DBR, Spark, Hive etc. 

1. Databricks: https://docs.databricks.com/gcp/en/sql/language-manual/functions/space
2. Hive: space(int n) | Returns a string with n spaces.
3. Spark: https://spark.apache.org/docs/latest/api/sql/index.html#space

tests/dialects/test_clickhouse.py
```diff
@@ -240,7 +240,6 @@ def test_clickhouse(self):
                 "databricks": "SELECT SUBSTRING_INDEX('a.b.c.d', '.', 2)",
                 "spark": "SELECT SUBSTRING_INDEX('a.b.c.d', '.', 2)",
                 "mysql": "SELECT SUBSTRING_INDEX('a.b.c.d', '.', 2)",
-                "clickhouse": "SELECT substringIndex('a.b.c.d', '.', 2)",
```
georgesittas: Why has this been removed?

sqlglot/expressions.py
georgesittas: I also checked various dialects, those that support it only allow one argument.

tests/dialects/test_clickhouse.py
ankur334: Reverted back changes @georgesittas . Due to some confusion, I removed it mistakenly.

sqlglot/dialects/bigquery.py
ankur334: Added one function in `dialect.py` with the same name suggested above. Kindly check @VaggelisD

sqlglot/expressions.py
ankur334: Thanks for this correction @VaggelisD. Added the required change

sqlglot/dialects/dialect.py
```diff
@@ -1934,3 +1934,12 @@ def build_replace_with_optional_replacement(args: t.List) -> exp.Replace:
         expression=seq_get(args, 1),
         replacement=seq_get(args, 2) or exp.Literal.string(""),
     )
+
+
+def _space_sql(self: Generator, expression: exp.Space) -> str:
```
georgesittas, line 1939: Let's remove the leading underscore since this is a "public" helper.

georgesittas (approved):
LGTM, @ankur334, just a final nit on my end and should be good to go.

georgesittas:
I'll just get this in and clean it up– thanks!

sqlglot/dialects/dialect.py
ankur334: True

PR #5313, opened 2025-06-30
feat(exasol): added the regexp_replace function

nnamdi16:
This involves adding [REGEXP_REPLACE](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/regexp_replace.htm) built -in function to exasol dialect in sqlglot.

### DIFF
--- a/tests/dialects/test_exasol.py
+++ b/tests/dialects/test_exasol.py
@@ -103,6 +103,105 @@ class TestExasol(Validator):
             },
         )
 
+        self.validate_all(
+            "SELECT BIT_XOR(x, 1)",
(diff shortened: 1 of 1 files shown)


tests/dialects/test_exasol.py
```diff
@@ -103,102 +103,21 @@ def test_bits(self):
             },
         )
 
-        self.validate_all(
```
georgesittas: I'm assuming that these tests were removed by mistake?

nnamdi16: I would look into it

PR #5314, opened 2025-06-30
Refactor(athena)!: route statements to hive/trino depending on their type

### LINKED ISSUE #5267: Athena adds an extra backslash to string literals

georgesittas:
Fixes #5267

This PR is a refactor of the Athena implementation. It was motivated by the linked issue, but its value extends beyond that– in other words, the ticket is not the only issue that I expect to be fixed by this PR.

TL;DR: Athena will now act as a proper router to either Hive or Trino, depending on the statement type (1, 2). Until now, it only routed the AST to the correct generator, but this was not enough, because the _lexical_ syntaxes of Hive and Trino contain incompatibilities, such as different string escapes, making them tricky to handle with a single tokenizer.

The way is implemented is that Athena now tokenizes the input string once, using a superset tokenizer that consumes both HiveQL and Trino SQL, and then re-tokenizes depending on what it finds in the token stream. Some rules have been implemented to decide whether the Hive or Trino dialect will be used post-tokenization.

In order to avoid having to re-detect what dialect a token stream corresponds to in the parsing stage, I've introduced a sentinel token type called `HIVE_TOKEN_STREAM` that is prepended in token streams corresponding to Hive inputs. This way, we can simply peek into the stream during parsing and figure out what parser to use. Of course, before parsing the token is popped from the stream, since it's not really a valid token.

Lastly, to support the routing mechanism, I changed `tokenizer` and `json_path_tokenizer` to proper methods, instead of `@property` ones. This was necessary in order to propagate the correct dialect instances for Hive and Trino in each `tokenize` invocation.

References:
1. https://docs.aws.amazon.com/athena/latest/ug/dml-queries-functions-operators.html
2. https://docs.aws.amazon.com/athena/latest/ug/ddl-reference.html

### DIFF
--- a/sqlglot/dialects/athena.py
+++ b/sqlglot/dialects/athena.py
@@ -2,46 +2,218 @@ from __future__ import annotations
 
 import typing as t
 
-from sqlglot import exp
-from sqlglot.dialects.trino import Trino
-from sqlglot.dialects.hive import Hive
-from sqlglot.tokens import TokenType
+from sqlglot import exp, generator, parser, tokens
+from sqlglot.dialects import Dialect, Hive, Trino
+from sqlglot.tokens import TokenType, Token
+
+
+class Athena(Dialect):
+    """
+    Over the years, it looks like AWS has taken various execution engines, bolted on AWS-specific
+    modifications and then built the Athena service around them.
+
+    Thus, Athena is not simply hosted Trino, it's more like a router that routes SQL queries to an
+    execution engine depending on the query type.
+
+    As at 2024-09-10, assuming your Athena workgroup is configured to use "Athena engine version 3",
+    the following engines exist:
+
+    Hive:
+     - Accepts mostly the same syntax as Hadoop / Hive
+     - Uses backticks to quote identifiers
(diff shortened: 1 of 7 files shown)


erindru (approved):
Nice work!

I see we lost a bunch of tests but they were mostly showing the "fixing" of hive-style backtick quoting to Trino-style `"` quoting so I see why they got removed

georgesittas:
> I see we lost a bunch of tests but they were mostly showing the "fixing" of hive-style backtick quoting to Trino-style " quoting so I see why they got removed

Yes, exactly, I got rid of those tests because they didn't test valid inputs. I made sure to replace them with similar but valid variants, though.

cpcloud:
As always, thanks for the great work!

PR #5315, opened 2025-06-30
feat!: add TypeOf / toTypeName support

ankur334:
**What & Why** 

- New AST node TypeOf – inherits Func, _sql_names=("TYPEOF",).
- Parser
     - Default dialects pick up TYPEOF automatically.
     - ClickHouse alias TOTYPENAME added.

- Generator
    - Base rule → prints TYPEOF(expr)
    - ClickHouse override → toTypeName(expr)

- Tests

ClickHouse ↔ Hive / Spark round-trip (test_typeof.py).
Now `toTypeName` ⇄ `TYPEOF` transpiles easily. 
```
>>> transpile("SELECT toTypeName(1)", read="clickhouse", write="hive")[0]
"SELECT TYPEOF(1)"
```


sqlglot/expressions.py
```diff
@@ -5370,6 +5370,11 @@ def default_parser_mappings(cls):
         return {name: cls.from_arg_list for name in cls.sql_names()}
 
 
+class TypeOf(Func):
+    _sql_names = ["TYPEOF"]
+    arg_types = {"this": True}
```
georgesittas:
```suggestion
class Typeof(Func):
    pass
```

tests/dialects/test_databricks.py
```diff
@@ -93,6 +93,19 @@ def test_databricks(self):
...
+        self.validate_all(
+            "SELECT TYPEOF(1)",
+            read={
+                "databricks": "SELECT TYPEOF(1)",
+                "snowflake": "SELECT TYPEOF(1)",
+                "hive": "SELECT TYPEOF(1)",
+                "clickhouse": "SELECT TOTYPENAME(1)",
+            },
+            write={
+                "clickhouse": "SELECT TOTYPENAME(1)",
```
georgesittas: Did you test this against ClickHouse? It shouldn't be valid, you need to emit `toTypeName`.

georgesittas (commented):
Hey @ankur334, can you please check whether `TYPEOF` exists in the other [supported dialects](https://github.com/tobymao/sqlglot/blob/8a2f65d6b2b68ad5ba45a5aed5e56c4dc0fea6fc/sqlglot/dialects/dialect.py#L71-L100)?

sqlglot/expressions.py
ankur334: Hi @georgesittas If I don't specify the `_sql_names = ["TYPEOF"]`, then after parsin,g will it not make `Type_Of`. Is my understanding wrong here? 

```
FUNCTION_BY_NAME = {name: func for func in ALL_FUNCTIONS for name in func.sql_names()}
```

```
@classmethod
    def sql_names(cls):
        if cls is Func:
            raise NotImplementedError(
                "SQL name is only supported by concrete function implementations"
            )
        if "_sql_names" not in cls.__dict__:
            cls._sql_names = [camel_to_snake_case(cls.__name__)]
        return cls._sql_names
```

Databricks, Spark, Hive, [etc support](https://docs.databricks.com/aws/en/sql/language-manual/functions/typeof) `typeof`, not `type_of. 

Please correct me if I am thinking in the wrong direction

tests/dialects/test_databricks.py
ankur334: Clickhouse and its issue of case sensitivity :). Let me change it. I should be a little cautious while adding the ClickHouse function. I generally use the Clickhouse playground fiddle to test something faster. I missed testing it

sqlglot/expressions.py
georgesittas: I intentionally changed the casing of `O` to lowercase in `TypeOf`, because that way its name becomes `TYPEOF` :)

ankur334: ohh it is `class Typeof`, not `class TypeOf`. I didn't observe this. This is cool. Sure we can do like this 💯

sqlglot/dialects/clickhouse.py
```diff
@@ -1117,6 +1118,7 @@ class Generator(generator.Generator):
             exp.TimeStrToTime: _timestrtotime_sql,
             exp.TimestampAdd: _datetime_delta_sql("TIMESTAMP_ADD"),
             exp.TimestampSub: _datetime_delta_sql("TIMESTAMP_SUB"),
+            exp.Typeof: lambda self, e: self.func("toTypeName", e.this),
```
georgesittas:
```suggestion
            exp.Typeof: rename_func("toTypeName"),
```

PR #5320, opened 2025-07-01
fix(spark)!: distinguish STORED AS from USING

### LINKED ISSUE #5317: Spark incorrectly roundtrips `STORED AS` -> `USING`
**Before you file an issue**
we are going from spark to spark in this case

```
>>> sql = "CREATE TABLE student (id INT, name STRING, age INT) STORED AS ORC;"
>>> print(sqlglot.parse_one(sql, dialect="spark").sql(pretty=True, dialect="spark"))
CREATE TABLE student (
  id INT,
  name STRING,
  age INT
)
USING ORC
```
**Official Documentation**
I would just expect the round trip to keep the SQL as is, since now it is invalid

geooo109:
Fixes #5317

This PR distinguishes `STORED AS` from `USING`. 

Spark and Databricks supports hive format when creating a table. This means that if we only use the previous approach, generating `STORED AS` with `USING` the generated SQL can be wrong. 

Example: 
```CREATE TABLE student (id INT, name STRING, age INT) STORED AS TEXTFILE```

**DOCS**
[Hive Format (STORED AS)](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-create-table-hiveformat)

### DIFF
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -284,7 +284,8 @@ class Spark2(Hive):
             # (DAY_OF_WEEK(datetime) % 7) + 1 is equivalent to DAYOFWEEK_ISO(datetime)
             exp.DayOfWeekIso: lambda self, e: f"(({self.func('DAYOFWEEK', e.this)} % 7) + 1)",
             exp.DayOfYear: rename_func("DAYOFYEAR"),
-            exp.FileFormatProperty: lambda self, e: f"STORED AS {e.name.upper()}"
+            exp.FileFormatProperty: lambda self,
+            e: f"STORED AS {self.sql(e, 'this') if isinstance(e.this, exp.InputOutputFormat) else e.name.upper()}"
             if e.args.get("hive_format")
             else f"USING {e.name.upper()}",
             exp.From: transforms.preprocess([_unalias_pivot]),
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -284,7 +284,9 @@ class Spark2(Hive):
             # (DAY_OF_WEEK(datetime) % 7) + 1 is equivalent to DAYOFWEEK_ISO(datetime)
             exp.DayOfWeekIso: lambda self, e: f"(({self.func('DAYOFWEEK', e.this)} % 7) + 1)",
             exp.DayOfYear: rename_func("DAYOFYEAR"),
-            exp.FileFormatProperty: lambda self, e: f"USING {e.name.upper()}",
+            exp.FileFormatProperty: lambda _, e: f"STORED AS {e.name.upper()}"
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: tests/dialects/test_presto.py [hunk 1]
```diff
@@ -511,7 +511,7 @@ def test_ddl(self):
                 "duckdb": "CREATE TABLE test AS SELECT 1",
                 "presto": "CREATE TABLE test WITH (format='PARQUET') AS SELECT 1",
                 "hive": "CREATE TABLE test STORED AS PARQUET AS SELECT 1",
-                "spark": "CREATE TABLE test USING PARQUET AS SELECT 1",
+                "spark": "CREATE TABLE test STORED AS PARQUET AS SELECT 1",
```
georgesittas, line 514: Did you double-check these test modifications to verify they're correct? I.e., run them against Spark, etc.

sqlglot/dialects/spark2.py
```diff
@@ -284,7 +284,10 @@ class Generator(Hive.Generator):
             # (DAY_OF_WEEK(datetime) % 7) + 1 is equivalent to DAYOFWEEK_ISO(datetime)
             exp.DayOfWeekIso: lambda self, e: f"(({self.func('DAYOFWEEK', e.this)} % 7) + 1)",
             exp.DayOfYear: rename_func("DAYOFYEAR"),
-            exp.FileFormatProperty: lambda self, e: f"USING {e.name.upper()}",
+            exp.FileFormatProperty: lambda self,
+            e: f"STORED AS {self.sql(e, 'this') if isinstance(e.this, exp.InputOutputFormat) else e.name.upper()}"
+            if e.args.get("hive_format")
+            else f"USING {e.name.upper()}",
```
georgesittas: Can we clean this up a bit? Putting this logic in a single lambda is unreadable. Consider defining a `fileformatproperty_sql` for Hive so that you don't have to override the `TRANSFORMS` entry here, and simply override the method instead.

sqlglot/parser.py
```diff
@@ -2237,6 +2237,7 @@ def _parse_stored(self) -> t.Union[exp.FileFormatProperty, exp.StorageHandlerPro
                 if input_format or output_format
                 else self._parse_var_or_string() or self._parse_number() or self._parse_id_var()
             ),
+            hive_format=True,
```
georgesittas, line 2240: Should this only happen for Hive, Spark, Dbx?

sqlglot/dialects/spark2.py
geooo109: yep, will change it, thanks.

sqlglot/parser.py
geooo109: As, I searched there isn't any other dialect supporing `STORED AS`, so I putted there.

PR #5323, opened 2025-07-01
fix(parser): avoid CTE values ALIAS gen, when ALIAS exists

### LINKED ISSUE #5318: Bug: Databricks/Spark loses column-list for `VALUES … AS alias(col1, …)` inside a CTE
#### Environment  
| | |
|---|---|
| **sqlglot version** | `latest` |
| **Python** | 3.12.8 (on macOS arm64) |
| **Read dialect** | `databricks` (identical with `spark`) |
| **Write dialect** | `databricks` |

---

#### Fully reproducible snippet

```python
from sqlglot import parse_one

sql = """
WITH map AS (
  VALUES ('allure', 'Allure', 'US') AS map(app_id, brand, market)
)
SELECT app_id, brand, market FROM map;
"""

tree = parse_one(sql, read="databricks")
print("PARSED →", tree.sql(dialect="databricks"))
```

**Current output**

```sql
WITH map AS (
  SELECT *
  FROM (VALUES ('allure', 'Allure', 'US')) AS _values     -- **column names dropped**
)
SELECT app_id, brand, market                              -- **now undefined**
FROM map
```

`app_id, brand, market` disappear, so subsequent references fail in Spark/Databricks.

⸻

**Expected output: Alias entire derived columns**
```
WITH map AS (
  (VALUES ('allure', 'Allure', 'US')) AS map(app_id, brand, market)
)
SELECT app_id, brand, market
FROM map;
```

geooo109:
Fixes #5318

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3395,8 +3395,11 @@ class Parser(metaclass=_Parser):
         )
 
         values = cte.this
-        if isinstance(values, exp.Values) and not values.alias:
-            cte.set("this", exp.select("*").from_(exp.alias_(values, "_values", table=True)))
+        if isinstance(values, exp.Values):
+            if values.alias:
+                cte.set("this", exp.select("*").from_(values))
+            else:
+                cte.set("this", exp.select("*").from_(exp.alias_(values, "_values", table=True)))
 
         return cte
 
--- a/tests/dialects/test_databricks.py
+++ b/tests/dialects/test_databricks.py
@@ -186,7 +186,10 @@ class TestDatabricks(Validator):
             "SELECT test, LISTAGG(email, '') AS Email FROM organizations GROUP BY test",
         )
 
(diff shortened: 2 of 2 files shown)


geooo109:
Should we add `SELECT *` in the CTE ?

sqlglot/parser.py
```diff
@@ -3394,8 +3394,9 @@ def _parse_cte(self) -> t.Optional[exp.CTE]:
             comments=comments,
         )
 
-        if isinstance(cte.this, exp.Values):
-            cte.set("this", exp.select("*").from_(exp.alias_(cte.this, "_values", table=True)))
+        values = cte.this
+        if isinstance(values, exp.Values) and not values.alias:
+            cte.set("this", exp.select("*").from_(exp.alias_(values, "_values", table=True)))
```
georgesittas: Why not reuse the `TableAlias` attached to the `Values` node instead of mangling it (which we do today)? This is going to generate inconsistent SQL depending on the presence (or not) of an alias, right? Doesn't seem right.

georgesittas (approved):

PR #5336, opened 2025-07-02
feat(optimizer)!: annotate type for LAST_VALUE

geooo109:
This PR adds support for the annotate type of `LAST_VALUE`

[BigQuery LAST_VALUE](https://cloud.google.com/bigquery/docs/reference/standard-sql/navigation_functions#last_value)


geooo109:
> Let's double-check whether annotating `Window` by `this` has unintended side-effects

Do you have any example on mind that this could break ?

georgesittas:
You can gather all tests that produce a `Window` node and see if there are interesting examples where the type of the window should not match `this.type`. I think this should be fine, though.

PR #5338, opened 2025-07-03
fix: fix comment lost before GROUP, JOIN and HAVING

chiiips:
Fix the lost single-line comments before GROUP, JOIN and HAVING.
```SQL
SELECT
    count(*)
FROM
    table_a
/* join comment */
JOIN
    table_b
ON
    table_a.id = table_b.id
/* group by comment */
GROUP BY
    table_a.id
/* having comment */
HAVING
    table_a.id = 1
```
The AST of the aboving SQL was like this with all the comments lost:
```
Select(
  expressions=[
    Count(
      this=Star(),
      big_int=True)],
  from=From(
    this=Table(
      this=Identifier(this=table_a, quoted=False))),
  joins=[
    Join(
      this=Table(
        this=Identifier(this=table_b, quoted=False)),
      on=EQ(
        this=Column(
          this=Identifier(this=id, quoted=False),
          table=Identifier(this=table_a, quoted=False)),
        expression=Column(
          this=Identifier(this=id, quoted=False),
          table=Identifier(this=table_b, quoted=False))))],
  group=Group(
    expressions=[
      Column(
        this=Identifier(this=id, quoted=False),
        table=Identifier(this=table_a, quoted=False))]),
  having=Having(
    this=EQ(
      this=Column(
        this=Identifier(this=id, quoted=False),
        table=Identifier(this=table_a, quoted=False)),
      expression=Literal(this=1, is_string=False))))
```
I found that the prefix comments for other keywords are also lost for the same reason (e.g., IN), but currently, the issues have only been encountered with GROUP, JOIN, and HAVING when I use sqlglot, so I have only fixed these three.

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3818,8 +3818,7 @@ class Parser(metaclass=_Parser):
         kwargs["pivots"] = self._parse_pivots()
 
         comments = [c for token in (method, side, kind) if token for c in token.comments]
-        if len(comments) == 0:
-            comments = join_comments
+        comments = (join_comments or []) + comments
         return self.expression(exp.Join, comments=comments, **kwargs)
 
     def _parse_opclass(self) -> t.Optional[exp.Expression]:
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -3817,6 +3818,8 @@ def _parse_join(
         kwargs["pivots"] = self._parse_pivots()
 
         comments = [c for token in (method, side, kind) if token for c in token.comments]
+        if len(comments) == 0:
+            comments = join_comments
```
```suggestion
        comments = (join_comments or []) + comments
```

georgesittas (approved):
Thanks

PR #5362, opened 2025-07-07
feat(fabric): Ensure TIMESTAMPTZ is used with AT TIME ZONE

mattiasthalen:
This pull request updates the handling of `TIMESTAMPTZ` and `DATETIMEOFFSET` types in the Fabric SQL dialect to ensure compliance with Microsoft Fabric's requirements. It introduces logic to differentiate how `TIMESTAMPTZ` is treated based on whether it is used in an `AT TIME ZONE` expression and adjusts test cases accordingly.

### Changes to `TIMESTAMPTZ` and `DATETIMEOFFSET` handling:

* **Logic for `TIMESTAMPTZ` in `datatype_sql`:**
  - Added a check to wrap `TIMESTAMPTZ` in `AT TIME ZONE 'UTC'` when `AT TIME ZONE` is missing.

### Updates to test cases:
  
* **Modified precision handling for `DATETIMEOFFSET`:**
  - Adjusted tests to ensure `DATETIMEOFFSET` is only valid when paired with `AT TIME ZONE`, preserving precision limits.
  
* **Added new test for `TIMESTAMPTZ` behavior:**
  - Introduced a dedicated test to verify that `TIMESTAMPTZ` is used with `AT TIME ZONE`.

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -65,6 +65,19 @@ class Fabric(TSQL):
         }
 
         def datatype_sql(self, expression: exp.DataType) -> str:
+            # DATETIMEOFFSET is only supported when used with AT TIME ZONE:
+            # https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support
+            if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
+                parent = getattr(expression, "parent", None)
+                grandparent = getattr(parent, "parent", None) if parent else None
+
+                if not (grandparent and isinstance(grandparent, exp.AtTimeZone)):
+                    # Not used with AT TIME ZONE: treat as TIMESTAMP (DATETIME2)
+                    expression = exp.DataType(
+                        this=exp.DataType.Type.TIMESTAMP,
+                        expressions=expression.expressions,
+                    )
+
             # Check if this is a temporal type that needs precision handling. Fabric limits temporal
             # types to max 6 digits precision. When no precision is specified, we default to 6 digits.
             if (
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -68,14 +68,11 @@ class Fabric(TSQL):
             # DATETIMEOFFSET is only supported when used with AT TIME ZONE:
             # https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support
             if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
-                parent = getattr(expression, "parent", None)
-                grandparent = getattr(parent, "parent", None) if parent else None
+                attimezone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
 
-                if not (grandparent and isinstance(grandparent, exp.AtTimeZone)):
-                    # Not used with AT TIME ZONE: treat as TIMESTAMP (DATETIME2)
+                if not attimezone:
                     expression = exp.DataType(
-                        this=exp.DataType.Type.TIMESTAMP,
-                        expressions=expression.expressions,
+                        this=exp.DataType.Type.TIMESTAMP, expressions=expression.expressions
                     )
 
             # Check if this is a temporal type that needs precision handling. Fabric limits temporal
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import typing as t
+
 from sqlglot import exp
 from sqlglot.dialects.dialect import NormalizationStrategy
 from sqlglot.dialects.tsql import TSQL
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/fabric.py
```diff
@@ -65,6 +65,19 @@ class Generator(TSQL.Generator):
...
+            if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
+                parent = getattr(expression, "parent", None)
+                grandparent = getattr(parent, "parent", None) if parent else None
+
+                if not (grandparent and isinstance(grandparent, exp.AtTimeZone)):
+                    # Not used with AT TIME ZONE: treat as TIMESTAMP (DATETIME2)
+                    expression = exp.DataType(
+                        this=exp.DataType.Type.TIMESTAMP,
+                        expressions=expression.expressions,
+                    )
```
georgesittas: Not sure if this approach is very robust. It assumes that `AtTimeZone` will only appear as a grandparent, which is true for `CAST` expressions, but I'm not sure if it can be generalized to other expressions where a `DataType` may appear.

I'd suggest generalizing it as follows:

```python
attimezone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
if not attimezone:
    expression = exp.DataType(this=exp.DataType.Type.TIMESTAMP, expressions=expression.expressions)
```

Including `Select` in the above argument list is done to prune the walk path at the `SELECT` boundary.

georgesittas (commented):
Thanks for the PR @mattiasthalen!

sqlglot/dialects/fabric.py
mattiasthalen: That looks alot nicer than what I did 😅

sqlglot/dialects/fabric.py
```diff
@@ -65,6 +65,16 @@ class Generator(TSQL.Generator):
         }
 
         def datatype_sql(self, expression: exp.DataType) -> str:
+            # DATETIMEOFFSET is only supported when used with AT TIME ZONE:
+            # https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support
+            if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
+                attimezone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
+
+                if not attimezone:
```
georgesittas: Minor oversight on my end here, fixed it really quick:
```suggestion
                if not isinstance(attimezone, exp.AtTimeZone):
```

georgesittas (approved):

georgesittas:
Is there something else pending for this PR on your end? Looks good to go.

mattiasthalen:
> Is there something else pending for this PR on your end? Looks good to go.

Yeah, but I'm not sure if the error is here or in SQLMesh 😄
I'm trying to handle the fabric integration tests, and prior to your suggestion I got this error:
```
FAILED tests/core/engine_adapter/integration/test_integration.py::test_to_time_column[fabric-2020-01-01 00:00:00+00:00-time_column_type2-None-result2] - AssertionError: assert Timestamp('2020-01-01 00:00:00') == Timestamp('2020-01-01 00:00:00+0000', tz='UTC')
```
But now it's back to:
```
FAILED tests/core/engine_adapter/integration/test_integration.py::test_to_time_column[fabric-2020-01-01 00:00:00+00:00-time_column_type2-None-result2] - pyodbc.ProgrammingError: ('ODBC SQL type -155 is not yet supported.  column-index=0  type=...
```

But I'm struggling in general to see the query actually executed.

georgesittas:
Can you see the queries executed under `logs/`? Also consider using the `--debug` flag, or manually call `configure_logging` to log the queries for the integration test, when ran with `pytest`.

mattiasthalen:
> Can you see the queries executed under `logs/`? Also consider using the `--debug` flag, or manually call `configure_logging` to log the queries for the integration test, when ran with `pytest`.

I think I've narrowed the issue down.
What happens is that when it recevies `TIMESTAMPTZ` without a `AT TIME ZONE` it truncates the TZ info and changes to `TIMESTAMP`. What needs to happen (I think) is to cast it to UTC.

sqlglot/dialects/fabric.py
```diff
@@ -91,6 +93,17 @@ def datatype_sql(self, expression: exp.DataType) -> str:
 
             return super().datatype_sql(expression)
 
+        def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) -> str:
```
georgesittas, line 96: Why move this logic inside of `cast_sql`, though? Shouldn't we override `datatype_sql`, instead?

mattiasthalen: Can we do that there?

georgesittas: We'd have to complicate the logic a bit to figure out where to attach `AT TIME ZONE`. Ok, we can go with this for now and keep iterating I guess.

georgesittas:
@mattiasthalen let me know when this is ready to be merged.

mattiasthalen:
Will mark ready ASAP, just want to check a couple of things ☺️

Issue #5366, opened 2025-07-07
Round of double not transpiled correctly for postgres


blecourt-private:
# Summary

* read: duckdb
* write: postgres, redshift, risingwave, materialize

In postgres and related dialects the following syntax for the ROUND function is valid:
 
```
round(numeric, integer)
```

and the following is not: 

```
round(double precision, integer)
```

(On a side note: `round(double precision)` is allowed, without specifying the number of decimals.)

Sqlglot doesn't handle this correctly.

# Reproducible code

```python
from sqlglot import transpile
query="select round(cast(amount as double precision), 3)"
transpile(query, read="duckdb", write="postgres")
```

The result is `['SELECT ROUND(CAST(amount AS DOUBLE PRECISION), 3)']`. (And similar if the write dialect is redshift or risingwave).

I have verified that `SELECT ROUND(CAST(amount AS DOUBLE PRECISION), 3)` throws an error in Postgres and RisingWave. I have not verified this for Redshift and Materialize (apart from consulting the docs for these dialects).

# Docs

https://www.postgresql.org/docs/current/functions-math.html
https://docs.aws.amazon.com/redshift/latest/dg/r_ROUND.html
https://docs.risingwave.com/sql/functions/mathematical#mathematical-functions-and-operators
https://materialize.com/docs/sql/functions/#numbers-functions

# Bonus info

This in turn has implications for `sqlmesh table_diff`. I have tables with a column amount of type DOUBLE PRECISION and this causes `table_diff` to throw the following error:

```
Schema Diff Between '<table a>' and '<table b>':
└── Schemas match
Traceback (most recent call last):

psycopg2.errors.InternalError_: Failed to run the query

Caused by these errors (recent errors listed first):
  1: Failed to bind expression: CASE WHEN ROUND("s"."amount", 3) = ROUND("t"."amount", 3) THEN 1 WHEN ("s"."amount" IS NULL) AND ("t"."amount" IS NULL) THEN 1 WHEN ("s"."amount" IS NULL) OR ("t"."amount" IS NULL) THEN 0 ELSE 0 END
  2: function round_digit(double precision, integer) does not exist
```

I am using a risingwave engine but because of the way sqlglot seems to transpile this expression to postgres I am expecting the same problem with postgres and other related dialects.

PR #5368, opened 2025-07-07
Fix round double for postgres

blecourt-private:
Fixing https://github.com/tobymao/sqlglot/issues/5366

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -40,6 +40,7 @@ from sqlglot.dialects.dialect import (
 )
 from sqlglot.generator import unsupported_args
 from sqlglot.helper import is_int, seq_get
+from sqlglot.optimizer.annotate_types import annotate_types
 from sqlglot.parser import binary_range_parser
 from sqlglot.tokens import TokenType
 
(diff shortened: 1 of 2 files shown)


blecourt-private:
Hi @georgesittas 

I'm not sure what I should add here to make this pull request complete. 

One thing I'm in doubt about is how much effort should be put into verifying with various dialects? For instance I have checked various queries directly in a Postgres and a RisingWave engine (making sure that `round(double precision, integer)` doesn't work and that the proposed changes are needed to transpile correctly) but I haven't done the same in Redshift and Materialize, instead relying on their official docs.

tests/dialects/test_postgres.py
```diff
@@ -1460,3 +1460,20 @@ def _validate_udt(sql: str):
         _validate_udt('CAST(5 AS "MySchema"."MyType")')
         _validate_udt('CAST(5 AS MySchema."MyType")')
         _validate_udt('CAST(5 AS "MyCatalog"."MySchema"."MyType")')
+
+    def test_round(self):
+        self.validate_identity("ROUND(x)")
```
georgesittas, line 1465: Let's include a `ROUND(x, y)` roundtrip test for Postgres.

sqlglot/dialects/postgres.py
```diff
@@ -613,6 +637,7 @@ class Generator(generator.Generator):
             exp.Rand: rename_func("RANDOM"),
             exp.RegexpLike: lambda self, e: self.binary(e, "~"),
             exp.RegexpILike: lambda self, e: self.binary(e, "~*"),
+            exp.Round: _round,
```
georgesittas, line 640: Nit: Can we get rid of this entry and move `_round` under the `Generator` class, named as `round_sql`? I'd also inline `_to_decimal`.

georgesittas (commented):
> One thing I'm in doubt about is how much effort should be put into verifying with various dialects? For instance I have checked various queries directly in a Postgres and a RisingWave engine (making sure that round(double precision, integer) doesn't work and that the proposed changes are needed to transpile correctly) but I haven't done the same in Redshift and Materialize, instead relying on their official docs.

That's fine, the testing you described suffices for this use case.

FYI, this is a breaking change, since it changes the AST's structure. Otherwise, the PR looks good, I don't expect any issues from preemptively casting the argument for postgres if we can't infer it.

tests/dialects/test_postgres.py
```diff
@@ -1460,3 +1460,20 @@ def _validate_udt(sql: str):
...
+        self.validate_identity("ROUND(CAST(x AS DOUBLE PRECISION))")
+        self.validate_identity("ROUND(CAST(x AS DECIMAL), 4)")
+        self.validate_identity("ROUND(CAST(x AS INT), 4)")
```
georgesittas, line 1468:
```suggestion
        self.validate_identity("ROUND(CAST(x AS INT), 4)")
        self.validate_identity(
            "ROUND(x, y)",
            "ROUND(CAST(x AS DECIMAL), y)",
        )
```

georgesittas:
I'll get this in and take it to the finish line. Thanks!

PR #5369, opened 2025-07-07
feat(optimizer)!: annotate type for DATETIME

geooo109:
This PR adds support for the type annotation of `DATETIME`

**DOCS**
[BigQuery DATETIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -454,7 +454,9 @@ class BigQuery(Dialect):
         exp.SHA2: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.BINARY),
         exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
         exp.Split: lambda self, e: self._annotate_by_args(e, "this", array=True),
-        exp.TimestampFromParts: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.DATETIME),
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -454,6 +454,7 @@ class BigQuery(Dialect):
         exp.SHA2: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.BINARY),
         exp.Sign: lambda self, e: self._annotate_by_args(e, "this"),
         exp.Split: lambda self, e: self._annotate_by_args(e, "this", array=True),
+        exp.TimestampFromParts: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.DATETIME),
```
tobymao: are you sure this should return a datetime and not a timestamp? datetime in bq != datetime in sqlglot i think

VaggelisD (commented):
As a heads up, BigQuery's `DATETIME(...)` can be built into multiple AST nodes, check out `bigquery.py::_build_datetime()`

sqlglot/dialects/bigquery.py
geooo109: I see, the docs of BigQuery + what I have tested always return `DATETIME`. Moreover,  `DATETIME` roundtrips to `DATETIME` in BigQuery so I believe this is good to go.

PR #5371, opened 2025-07-07
feat(optimizer)!: annotate type for LAG

geooo109:
This PR adds type annotation for `LAG`

**DOCS**
[BigQuery LAG](https://cloud.google.com/bigquery/docs/reference/standard-sql/navigation_functions#lag)

### DIFF
--- a/tests/fixtures/optimizer/annotate_functions.sql
+++ b/tests/fixtures/optimizer/annotate_functions.sql
@@ -585,6 +585,14 @@ DOUBLE;
 DATETIME(2025, 1, 1, 12, 0, 0);
 DATETIME;
 
+# dialect: bigquery
+LAG(tbl.bigint_col, 1 , 2.5) OVER (ORDER BY tbl.bigint_col);
+DOUBLE;
+
+# dialect: bigquery
+LAG(tbl.bigint_col, 1 , 2) OVER (ORDER BY tbl.bigint_col);
+BIGINT;
+
 --------------------------------------
 -- Snowflake
 --------------------------------------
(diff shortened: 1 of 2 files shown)


tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -585,6 +585,14 @@ DOUBLE;
 DATETIME(2025, 1, 1, 12, 0, 0);
 DATETIME;
 
+# dialect: bigquery
+LAG(tbl.bigint_col, 1 , 2.5) OVER (ORDER BY tbl.bigint_col);
+DOUBLE;
```
georgesittas, line 590: Are you sure that coercing into `DOUBLE` is correct here? Did you try this out in BigQuery? The docs say:

> Return data type: Same type as ``value_expression`.

geooo109:
```
WITH cte AS
 (SELECT 1 as id
  UNION ALL SELECT 2
  UNION ALL SELECT 3
 )
SELECT
  TYPEOF(id)
FROM cte;

returns -> INT64
```

```
WITH cte AS
 (SELECT 1 as id
  UNION ALL SELECT 2
  UNION ALL SELECT 3
 )
SELECT
  TYPEOF(LAG(id, 0, 5) OVER (order by id))
FROM cte;

returns -> INT64
```

```
WITH cte AS
 (SELECT 1 as id
  UNION ALL SELECT 2
  UNION ALL SELECT 3
 )
SELECT
  TYPEOF(LAG(id, 0, 0.5) OVER (order by id))
FROM cte;

returns -> FLOAT64
```

georgesittas: Sounds good

PR #5372, opened 2025-07-07
feat(fabric): update TIMESTAMPTZ handling to include casting to DATETIME2

mattiasthalen:
This pull request updates the handling of `TIMESTAMPTZ` data type in the `Fabric` SQL dialect. The changes ensure that when `TIMESTAMPTZ` is not within an `AT TIME ZONE` expression, it is both converted to UTC and cast to `TIMESTAMP`. Corresponding test cases have been updated to validate this behavior.

### Updates to `TIMESTAMPTZ` handling:

* **Enhancement in `cast_sql` method**:
  - Modified the logic to wrap `TIMESTAMPTZ` in `AT TIME ZONE 'UTC'` and cast it to `TIMESTAMP` when not inside an `AT TIME ZONE` expression.

* **Test case adjustments**:
  - Updated the `test_timestamptz_handling` test to validate that `TIMESTAMPTZ` is converted to UTC and cast to `DATETIME2` when outside an `AT TIME ZONE` expression.

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -98,9 +98,10 @@ class Fabric(TSQL):
             if expression.to and expression.to.is_type(exp.DataType.Type.TIMESTAMPTZ):
                 attimezone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
                 if not isinstance(attimezone, exp.AtTimeZone):
-                    # We're not inside an AT TIME ZONE, so wrap the cast in AT TIME ZONE 'UTC'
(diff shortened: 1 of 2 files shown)


mattiasthalen:
I think I'd always want to cast to `TIMESTAMP`, even if it's within an `AT TIME ZONE`, but I keep hitting max recursion depth when trying to add that.

mattiasthalen:
Found a better way to handle the issue downstream, instead of stripping the TZ info from SQLGlot

mattiasthalen:
Nope, it strips all information regarding time zone:
> Time zone offset aware and preservation: No
https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=fabric&preserve-view=true#datetime2-description

PR #5374, opened 2025-07-08
fix!: duckdb week/quarter support

eakmanrq:
This logic was originally added in 2023: https://github.com/tobymao/sqlglot/pull/1780

I can see quarter was at least added in 2024: https://github.com/duckdb/duckdb/pull/11898

Tested and confirmed that `0.10.3` added quarter support and week support is in `0.10.2` so not sure when that was added.

The original logic actually has a bug for quarter since it converted a quarter to 90 days but that isn't correct when doing additions and subtractions to dates. 

```
conn.execute("SELECT DATE_ADD(TIMESTAMP'2020-01-01 00:00:00', INTERVAL 1 QUARTER)").fetchall()
[(datetime.datetime(2020, 4, 1, 0, 0),)]
conn.execute("SELECT DATE_ADD(TIMESTAMP'2020-01-01 00:00:00', INTERVAL 90 DAY)").fetchall()
[(datetime.datetime(2020, 3, 31, 0, 0),)]
```

The previous week logic looks fine but isn't needed anymore:
```
conn.execute("SELECT INTERVAL (-5) WEEK").fetchall()
[(datetime.timedelta(days=-35),)]
```

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import typing as t
 
 from sqlglot import exp, generator, parser, tokens, transforms
+
 from sqlglot.expressions import DATA_TYPE
 from sqlglot.dialects.dialect import (
     Dialect,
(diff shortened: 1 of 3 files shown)


georgesittas (approved):
Thanks!

PR #5376, opened 2025-07-08
Fix(optimizer)!: Preserve struct-column parentheses for RisingWave dialect

MisterWheatley:
This PR adds logic to the optimizer to preserve (possibly) semantically significant parentheses in the RisingWave dialect, see [docs here](https://docs.risingwave.com/sql/data-types/struct#retrieve-data-in-a-struct)

A minimal example looks like follows: Assume you have a column `struct_col (STRUCT<nested_col INT>)` and wish to access the nested column to bring it up to a top-level column. The select statement in RW would be (with quoting):
```sql
SELECT
  ("table"."struct_col")."nested_col" AS top_level_col
FROM
  "table"
```

The logic added checks if the dialect is RisingWave and the given parentheses follow the pattern `(<exp>).<identifier>` to preserve semantically significant parentheses. 

In addition this PR contains a minor refactor of the complicated conditional to instead use 2 guard clauses for better clarity.

Edit: Relevant Slack discussion: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1751957408688319

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -600,27 +600,67 @@ def _expand_struct_stars_bigquery(
 
 
 def _expand_struct_stars_risingwave(expression: exp.Dot) -> t.List[exp.Alias]:
-    """[RisingWave] Expand/Flatten (foo.bar).*, where bar is a struct column"""
-    parens = t.cast(exp.Paren, expression.find(exp.Paren))
+    """[RisingWave] Expand/Flatten (<exp>.bar).*, where bar is a struct column"""
 
-    # check that last selected identifier is struct for nested structs
-    if not parens.this.is_type(exp.DataType.Type.STRUCT):
+    # it is not (<sub_exp>).* pattern, which means we can't expand
+    if not isinstance(expression.this, exp.Paren):
         return []
 
-    struct_to_expand = exp.ColumnDef(this=parens.this, kind=parens.this.type)
+    # find column definition to get data-type
+    dot_column = t.cast(exp.Column, expression.find(exp.Column))
 
-    new_selections = []
+    if not dot_column.is_type(exp.DataType.Type.STRUCT):
+        return []
 
-    for nested_field in t.cast(exp.DataType, struct_to_expand.kind).expressions:
-        if not isinstance(nested_field.this, exp.Identifier):
+    current_expr: t.Union[exp.Dot, exp.Paren, exp.Column] = dot_column
+    current_struct = dot_column.type
+
+    # walk up AST and down into struct definition in sync
+    while current_expr.parent is not None:
+        # skip parentheses nodes
+        if isinstance(current_expr.parent, exp.Paren):
+            current_expr = current_expr.parent
+            continue
+
+        # if parent is not a dot, then something is wrong
+        if not isinstance(current_expr.parent, exp.Dot):
             return []
 
-        this = nested_field.this.copy()
+        # get right hand side of dot
+        rhs = current_expr.parent.right
 
-        new_identifier = t.cast(exp.Identifier, this)
-        # build new dot expression to end with (a).b AS b
-        new_dot = exp.Dot.build(expressions=[parens, new_identifier])
-        new_alias = alias(new_dot, this, copy=False)
+        # if it is star we are done
+        if isinstance(rhs, exp.Star):
+            break
+
+        # if RHS is not identifier, something is wrong
+        if not isinstance(rhs, exp.Identifier):
+            return []
+
+        # Check if current RHS identifier is in struct
+        matched = False
+        for struct_field_def in t.cast(exp.DataType, current_struct).expressions:
+            if struct_field_def.name == rhs.name:
+                matched = True
+                current_struct = struct_field_def.kind  # update struct
+                break
+
+        if not matched:
+            return []
+
+        # update parent
+        current_expr = current_expr.parent
+
+    # build new aliases to expand star
+    new_selections = []
+
+    # fetch the outermost parentheses for new aliaes
+    outer_paren = expression.this
+
+    for struct_field_def in t.cast(exp.DataType, current_struct).expressions:
+        new_identifier = exp.Identifier(this=struct_field_def.name)
+        new_dot = exp.Dot.build([outer_paren.copy(), new_identifier])
+        new_alias = alias(new_dot, new_identifier, copy=False)
         new_selections.append(new_alias)
 
     return new_selections
--- a/sqlglot/optimizer/simplify.py
+++ b/sqlglot/optimizer/simplify.py
@@ -748,9 +748,7 @@ def simplify_parens(expression: exp.Expression, dialect: DialectType = None) -> 
     if (
         dialect == "risingwave"
         and isinstance(parent, exp.Dot)
-        and (
-            isinstance(parent.right, (exp.Identifier, exp.Star))
-        )
+        and (isinstance(parent.right, (exp.Identifier, exp.Star)))
     ):
         return expression
 
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -123,6 +123,10 @@ class TestOptimizer(unittest.TestCase):
                 "d": "DATE",
                 "t": "DATETIME",
             },
+            "structs": {
+                "one": "STRUCT<a_1 INT, b_1 VARCHAR>",
+                "nested_0": "STRUCT<a_1 INT, nested_1 STRUCT<a_2 INT, nested_2 STRUCT<a_3 INT>>>",
+            },
         }
 
     def check_file(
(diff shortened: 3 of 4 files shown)


sqlglot/optimizer/simplify.py
```diff
@@ -729,31 +729,43 @@ def _simplify_binary(expression, a, b):
     return None
 
 
-def simplify_parens(expression):
+def simplify_parens(expression: exp.Expression, dialect: DialectType) -> exp.Expression:
```
georgesittas: I'd make this optional, so that it's backwards-compatible.
```suggestion
def simplify_parens(expression: exp.Expression, dialect: DialectType = None) -> exp.Expression:
```

sqlglot/optimizer/simplify.py
```diff
@@ -729,31 +729,43 @@ def _simplify_binary(expression, a, b):
...
+
+    if isinstance(parent, (exp.SubqueryPredicate, exp.Bracket)):
+        return expression
+
+    # Handle risingwave struct columns
+    # see https://docs.risingwave.com/sql/data-types/struct#retrieve-data-in-a-struct
     if (
-        not isinstance(this, exp.Select)
-        and not isinstance(parent, (exp.SubqueryPredicate, exp.Bracket))
-        and (
```
georgesittas, line 743: Perhaps we can keep the old conjunction and add a new term like `isinstance(parent, exp.Dot)`. We may not even have to factor in the dialect for this, seems like an innocuous change.

EDIT: note that if we do this, we can get rid of the `dialect` argument altogether, so this [comment](https://github.com/tobymao/sqlglot/pull/5376#discussion_r2191996600) becomes irrelevant.

MisterWheatley:
Hi @georgesittas 

As per our discussion in Slack, I did some testing and dug around in both the docs and source code of RisingWave and decided to move my findings here for better archiving/visibility

It appears that RisingWave are trying to align (partly) with Postgres in accessing Composite Types, [docs here](https://www.postgresql.org/docs/current/rowtypes.html#ROWTYPES-ACCESSING) in terms of syntax for accessing structs.
As far as I am able to determine, RisingWave follows a BigQuery like syntax with the exception that you need a set of parentheses around each nested level. 

Relevant issues from RisingWave are:
* https://github.com/risingwavelabs/risingwave/issues/9694

I am going to do a bit more work by adding more tests that work directly in RisingWave and then ensuring this PR covers those.

I'll let you know if I run into any issues and/or get stuck and could use some help. Thanks again for the fast responses.

georgesittas:
Do you mean there's other issues besides simplifying parentheses? Did you check out [this comment](https://github.com/tobymao/sqlglot/pull/5376#discussion_r2192001922) by any chance?

MisterWheatley:
> Do you mean there's other issues besides simplifying parentheses? Did you check out [this comment](https://github.com/tobymao/sqlglot/pull/5376#discussion_r2192001922) by any chance?

Partially. It is still fundamentally down to parentheses having semantic meaning for struct types in RisingWave.

However to support e.g. struct expansion and correct qualification of column(s) there is a need for more work.

In essence I (ideally) want to be able to (for a column `struct_col STRUCT<int_col INT, vchar_col VARCHAR>`) to be able to do the following optimization
```sql
SELECT
  (struct_col).*
FROM
  t
)
```
to the following:
```
SELECT
  ("t"."struct_col")."int_col" AS "int_col",
  ("t"."struct_col")."vchar_col" AS "vchar_col"
FROM
  "t" as "t"
```

i.e. we also need to handle these parentheses correctly in the `qualify_columns` logic somewhat equivalent to BigQuery struct expansion.

Edit: I am unsure if we have any other dependencies in terms of just adding the `exp.Dot` to the list of exclusions as you suggested above. I will get back to you with regards to that, but I am wary that it would have unintended consequences somewhere in e.g. BigQuery

georgesittas:
Just to make sure– have you verified that there are issues with the qualification logic as it is today? Or was that next on your list, i.e., testing? I don't see anything that is obviously unsupported in your example, e.g., for this query:

```sql
SELECT (struct_col).* FROM t
```

I expect that we're already parsing `struct_col` as a `Column`, so its resolution should _just_ work today. The rest of the identifiers in the dot path aren't `Column` instances, so they should also be left untouched, leading to the form you suggested.

MisterWheatley:
> Just to make sure– have you verified that there are issues with the qualification logic as it is today? Or was that next on your list, i.e., testing? I don't see anything that is obviously unsupported in your example, e.g., for this query:
> 
> ```sql
> SELECT (struct_col).* FROM t
> ```
> 
> I expect that we're already parsing `struct_col` as a `Column`, so its resolution should _just_ work today. The rest of the identifiers in the dot path aren't `Column` instances, so they should also be left untouched, leading to the form you suggested.

Glad to confirm and double test. I just tested on a clean install of v27.0.0 with the following snippet:
```python
from sqlglot import parse_one
import sqlglot
sql = """
SELECT
  (struct_col).*
FROM
  t
"""

schema = {
    "t" : {
        "struct_col" : "struct<nested_int INT, nested_char VARCHAR>"
    }
}

rw_ast = parse_one(sql,dialect='risingwave')
rw_opt = optimize(rw_ast,schema=schema,dialect='risingwave')
print(rw_opt)
```
gives the following result 
```sql
SELECT "t"."struct_col".* FROM "t" AS "t"
```

So the qualification sort of works, but there are two issues still that I am working to solve:

1. Preserve the parentheses as they have semantic meaning here.
2. Do star expansion equivalent to what is currently done in BigQuery. If you run my toy example with `dialect = "bigquery"` it gives: `SELECT "t"."struct_col"."nested_int" AS "nested_int", "t"."struct_col"."nested_char" AS "nested_char" FROM "t" AS "t"`

MisterWheatley:
Hi @georgesittas 

I did a lot more testing, but I think I managed to get it to a working stage that I am happy with for another review round.

I am aware of the comment on an old PR [here](https://github.com/tobymao/sqlglot/pull/3531#discussion_r1619235931) about introducing dialect specifics into the optimizer, when some of the logic should be shareable between e.g. BigQuery and RisingWave in regards to expanding `struct.*` columns. I deliberately didn't attempt to do a refactor, since that would have increased the complexity and I wanted to keep it as simple as possible

Let me know if you have more feedback and/or need more test cases added and I will get on it asap

sqlglot/optimizer/simplify.py
```diff
@@ -729,31 +729,46 @@ def _simplify_binary(expression, a, b):
...
-            or (
-                not isinstance(this, exp.Binary)
-                and not (isinstance(this, (exp.Not, exp.Is)) and parent_is_predicate)
-            )
-            or (isinstance(this, exp.Predicate) and not parent_is_predicate)
-            or (isinstance(this, exp.Add) and isinstance(parent, exp.Add))
-            or (isinstance(this, exp.Mul) and isinstance(parent, exp.Mul))
-            or (isinstance(this, exp.Mul) and isinstance(parent, (exp.Add, exp.Sub)))
+            isinstance(parent.right, exp.Identifier)  # (struct_col).nested_col
+            or parent.right.is_star  # (struct_col).* for expansion
```
georgesittas:
```suggestion
            isinstance(parent.right, (exp.Identifier, exp.Star))
```

sqlglot/optimizer/qualify_columns.py
```diff
@@ -599,6 +599,33 @@ def _expand_struct_stars(
     return new_selections
 
 
+def _expand_struct_stars_risingwave(expression: exp.Dot) -> t.List[exp.Alias]:
+    """[RisingWave] Expand/Flatten (foo.bar).*, where bar is a struct column"""
+    parens = t.cast(exp.Paren, expression.find(exp.Paren))
+
+    # check that last selected identifier is struct for nested structs
+    if not parens.this.is_type(exp.DataType.Type.STRUCT):
```
georgesittas: I suggest avoiding the cast here and ensuring `parens` is an instance of `Paren` before assuming `this` exists. If it's either not a `Paren`, or the type isn't a struct, return the empty list.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -599,6 +599,33 @@ def _expand_struct_stars(
...
+        return []
+
+    struct_to_expand = exp.ColumnDef(this=parens.this, kind=parens.this.type)
```
georgesittas: We don't need to construct a `ColumnDef` here, do we? Seems like we only need `kind`, i.e. `parens.this.type`, for line 614.

tests/test_optimizer.py
```diff
@@ -399,6 +399,8 @@ def test_qualify_columns(self, logger):
             'WITH "t" AS (SELECT 1 AS "c") (SELECT "t"."c" AS "c" FROM "t" AS "t")',
         )
 
+        # Struct expansion tests
```
georgesittas: Consider moving these tests to `qualify_columns.sql`. Check out line 477 and below in said file.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -599,6 +599,33 @@ def _expand_struct_stars(
...
+        if not isinstance(nested_field.this, exp.Identifier):
+            return []
+
+        this = nested_field.this.copy()
+
+        new_identifier = t.cast(exp.Identifier, this)
+        # build new dot expression to end with (a).b AS b
+        new_dot = exp.Dot.build(expressions=[parens, new_identifier])
+        new_alias = alias(new_dot, this, copy=False)
+        new_selections.append(new_alias)
```
georgesittas: Just double-checking– does RisingWave's `STRUCT` expansion only involve the first level in the struct? What if you do `struct.*` and then `struct` has nested struct fields?

georgesittas (commented):
@MisterWheatley thanks for the update– left another round of comments.

sqlglot/optimizer/qualify_columns.py
MisterWheatley: According to both the docs and confirming by running a few test in a RisingWave instance, it expands at the level of nesting. 

I.e. if you have a column called `zero` with the following datatype `STRUCT<a INT, nested STRUCT<b INT, c, INT>>` then the results are:
* `(zero).* => (zero).a [int], (zero).nested [struct<b int, c int>]`
* `((zero).nested).* => ((zero).nested).b [int], ((zero).nested).c [int]`

MisterWheatley:
Hi again @georgesittas 

I have tried to update according to your comments and questions. 

Based on your question regarding unpacking, I uncovered a problem with the previous way of doing it, hence why the code is now quite a bit different. High-level the new logic walks "up" the AST from the `exp.Column` and down the `exp.DataType.kind` simultaneously to check if they are in alignment to be expanded between AST and `STRUCT` definition

The re-added casts/type annotations were partly to fix a few mypy errors when running `make check`. If this is still incorrect, let me know and I can try and refactor to not use casts. I am not an expert in mypy, so I would appreciate a bit of guidance or a pointer to some relevant docs in that case.

I moved all the `struct.*` tests into the fixture file as you suggested to keep them together, including the existing BigQuery ones.

Again, thanks for the quick response times and looking forward to your feedback on the latest changes

sqlglot/optimizer/qualify_columns.py
```diff
@@ -599,6 +599,73 @@ def _expand_struct_stars(
...
 
+def _expand_struct_stars_risingwave(expression: exp.Dot) -> t.List[exp.Alias]:
+    """[RisingWave] Expand/Flatten (<exp>.bar).*, where bar is a struct column"""
+
+    # it is not (<sub_exp>).* pattern, which means we can't expand
+    if not isinstance(expression.this, exp.Paren):
+        return []
+
+    # find column definition to get data-type
+    dot_column = t.cast(exp.Column, expression.find(exp.Column))
```
georgesittas, line 610: I think it'd be safer if we did an instance check to ensure `dot_column` is indeed a `Column` in L612.

sqlglot/optimizer/qualify_columns.py
```diff
@@ -599,6 +599,73 @@ def _expand_struct_stars(
...
+        current_expr = current_expr.parent
+
+    # build new aliases to expand star
+    new_selections = []
+
+    # fetch the outermost parentheses for new aliaes
+    outer_paren = expression.this
+
+    for struct_field_def in t.cast(exp.DataType, current_struct).expressions:
+        new_identifier = exp.Identifier(this=struct_field_def.name)
```
georgesittas, line 661: We should respect the identifier found in the type here, otherwise we may drop quotes when we shouldn't.

georgesittas (commented):
Hey @MisterWheatley, thanks for addressing comments and sorry for the delayed response– I was OOO. The PR looks good to go; I will address the last couple of comments that I left after merging.

MisterWheatley:
> Hey @MisterWheatley, thanks for addressing comments and sorry for the delayed response– I was OOO. The PR looks good to go; I will address the last couple of comments that I left after merging.

I assumed as much. Thanks for the follow up and taking it the last mile, very much appreciated.

georgesittas:
Just a few minor changes on my end: https://github.com/tobymao/sqlglot/commit/8f43c5fbfbb9b6dccf2e9ee4a81f0513387969cd. Nice work!

PR #5377, opened 2025-07-08
feat(optimizer)!: parse and annotate type for ASCII

geooo109:
This PR adds parsing and type annotation support for `ASCII`

**DOCS**
[BigQuery ASCII](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#ascii)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5468,6 +5468,10 @@ class Array(Func):
     is_var_len_args = True
 
 
+class Ascii(Func):
+    pass
+
+
 # https://docs.snowflake.com/en/sql-reference/functions/to_array
 class ToArray(Func):
     pass
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -446,6 +446,7 @@ class BigQuery(Dialect):
             )
         },
         exp.ArrayConcat: lambda self, e: self._annotate_by_args(e, "this", "expressions"),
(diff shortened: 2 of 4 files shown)


sqlglot/expressions.py
```diff
@@ -5468,6 +5468,10 @@ class Array(Func):
     is_var_len_args = True
 
 
+class Ascii(Func):
+    pass
```
georgesittas, line 5472: Did you double-check whether this is supported by other dialects, and if so, whether more arguments may be expected?

geooo109: [TSQL](https://learn.microsoft.com/en-us/sql/t-sql/functions/ascii-transact-sql?view=sql-server-ver17)
[Databricks](https://docs.databricks.com/aws/en/sql/language-manual/functions/ascii)
[Snowflake](https://docs.snowflake.com/en/sql-reference/functions/ascii)
[MySQL](https://www.w3resource.com/mysql/string-functions/mysql-ascii-function.php)
[Postgres](https://www.postgresql.org/docs/8.2/functions-string.html)
[ClickHouse](https://clickhouse.com/docs/sql-reference/functions/string-functions#ascii)
[Spark](https://spark.apache.org/docs/latest/api/sql/index.html#ascii)

All the dialects use 1 input argument and return an integer.
Should we uplift this and insert it into the `dialect.py` ?

georgesittas: Sounds good, thanks for verifying. Yes, we can move it to `dialect.py`, within the `Type.INT: {...}` set.

PR #5378, opened 2025-07-08
feat(dremio): Add TIME_MAPPING for Dremio dialect

mateuszpoleski:
Adding `TIME_MAPPING` for Dremio dialect based on https://docs.dremio.com/current/reference/sql/sql-functions/DATE_TIME/

Also added support for `TO_CHAR` function which is needed for tests

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -1,6 +1,6 @@
 from sqlglot import expressions as exp
 from sqlglot import parser, generator, tokens
-from sqlglot.dialects.dialect import Dialect
+from sqlglot.dialects.dialect import Dialect, build_formatted_time
 
 
 class Dremio(Dialect):
@@ -11,8 +11,35 @@ class Dremio(Dialect):
     NULL_ORDERING = "nulls_are_last"
     SUPPORTS_VALUES_DEFAULT = False
 
+    TIME_MAPPING = {
+        # year
+        "YYYY": "%Y", "YY": "%y",
+
+        # month / day
+        "MM": "%m", "MON": "%b", "MONTH": "%B",
+        "DDD": "%j", "DD": "%d", "DY": "%a", "DAY": "%A",
+
+        # hours / minutes / seconds
+        "HH24": "%H", "HH": "%I",  # 24- / 12-hour
+        "MI": "%M",
+        "SS": "%S",
+        "FFF": "%f",
+
+        # ISO week / century etc.
+        "WW": "%W", "D": "%w", "CC": "%C",
+
+        # timezone
+        "TZD": "%Z",  # abbreviation  (UTC, PST, ...)
+        "TZO": "%z",   # numeric offset (+0200)
+    }
+
     class Parser(parser.Parser):
         LOG_DEFAULTS_TO_LN = True
+
+        FUNCTIONS = {
+            **parser.Parser.FUNCTIONS,
+            "TO_CHAR": build_formatted_time(exp.TimeToStr, "dremio"),
+        }
 
     class Generator(generator.Generator):
         NVL2_SUPPORTED = False
@@ -35,6 +62,12 @@ class Dremio(Dialect):
             exp.DataType.Type.DATETIME: "TIMESTAMP",
             exp.DataType.Type.ARRAY: "LIST",
             exp.DataType.Type.BIT: "BOOLEAN",
+        }
+
+        TRANSFORMS = {
+            **generator.Generator.TRANSFORMS,
+            exp.TimeToStr: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
+            exp.ToChar: lambda self, e: self.function_fallback_sql(e),
         }
 
         def datatype_sql(self, expression: exp.DataType) -> str:
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -13,24 +13,29 @@ class Dremio(Dialect):
 
     TIME_MAPPING = {
         # year
-        "YYYY": "%Y", "YY": "%y",
-
+        "YYYY": "%Y",
+        "YY": "%y",
         # month / day
-        "MM": "%m", "MON": "%b", "MONTH": "%B",
-        "DDD": "%j", "DD": "%d", "DY": "%a", "DAY": "%A",
-
+        "MM": "%m",
+        "MON": "%b",
+        "MONTH": "%B",
+        "DDD": "%j",
+        "DD": "%d",
+        "DY": "%a",
+        "DAY": "%A",
         # hours / minutes / seconds
-        "HH24": "%H", "HH": "%I",  # 24- / 12-hour
+        "HH24": "%H",
+        "HH": "%I",  # 24- / 12-hour
         "MI": "%M",
         "SS": "%S",
         "FFF": "%f",
-
         # ISO week / century etc.
-        "WW": "%W", "D": "%w", "CC": "%C",
-
+        "WW": "%W",
+        "D": "%w",
+        "CC": "%C",
         # timezone
         "TZD": "%Z",  # abbreviation  (UTC, PST, ...)
-        "TZO": "%z",   # numeric offset (+0200)
+        "TZO": "%z",  # numeric offset (+0200)
     }
 
     class Parser(parser.Parser):
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -25,10 +25,12 @@ class Dremio(Dialect):
         "DAY": "%A",
         # hours / minutes / seconds
         "HH24": "%H",
+        "HH12": "%I",
         "HH": "%I",  # 24- / 12-hour
         "MI": "%M",
         "SS": "%S",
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
     NULL_ORDERING = "nulls_are_last"
     SUPPORTS_VALUES_DEFAULT = False
 
+    TIME_MAPPING = {
+        # year
+        "YYYY": "%Y",
+        "YY": "%y",
+        # month / day
+        "MM": "%m",
```
VaggelisD, line 19: Should this be `%-m`? Dremio's `MM` does not look zero padded

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
...
+        "MON": "%b",
+        "MONTH": "%B",
+        "DDD": "%j",
```
VaggelisD, line 22: Ditto for `%-j`

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
...
+        "DD": "%d",
```
VaggelisD, line 23: Ditto for `%-d`

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
...
+        "DY": "%a",
+        "DAY": "%A",
+        # hours / minutes / seconds
+        "HH24": "%H",
```
VaggelisD, line 27: Ditto for `%-H`

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
...
+        "HH": "%I",  # 24- / 12-hour
```
VaggelisD, line 29: Ditto for `%-I`

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
...
+        "MI": "%M",
+        "SS": "%S",
+        "FFF": "%f",
+        # ISO week / century etc.
+        "WW": "%W",
+        "D": "%w",
+        "CC": "%C",
+        # timezone
+        "TZD": "%Z",  # abbreviation  (UTC, PST, ...)
+        "TZO": "%z",  # numeric offset (+0200)
```
VaggelisD: Let's double check these too please

tests/dialects/test_dremio.py
```diff
@@ -94,3 +94,53 @@ def test_multi_arg_distinct_unsupported(self):
...
+            "FFF",
+            "TZD",
+            "TZO",
+            "WW",
+            "CC",
+        ]
+
+        for fmt in dremio_formats:
+            sql = f"SELECT TO_CHAR(CAST('2025-06-24 12:34:56' AS TIMESTAMP), '{fmt}')"
+            self.validate_identity(sql)
```
VaggelisD: let's inline these. Also, consider using `self.subtest` with loops so we can get a better diagnostic message in case one of these end up failing

VaggelisD: Lets inline the list in the for loop if it's not used elsewhere

tests/dialects/test_dremio.py
```diff
@@ -94,3 +94,53 @@ def test_multi_arg_distinct_unsupported(self):
             "SELECT COUNT(DISTINCT a, b) FROM t",
             "SELECT COUNT(DISTINCT CASE WHEN a IS NULL THEN NULL WHEN b IS NULL THEN NULL ELSE (a, b) END) FROM t",
         )
+
+    def test_time_mapping_identity(self):
```
VaggelisD: [Nit] Let's combine the 2 functions into `test_time_mapping`

tests/dialects/test_dremio.py
```diff
@@ -94,3 +94,53 @@ def test_multi_arg_distinct_unsupported(self):
...
+                "postgres": f"SELECT TO_CHAR({ts}, 'YY-DDD HH24:MI:SS.US TZ')",
+                "oracle": f"SELECT TO_CHAR({ts}, 'YY-DDD HH24:MI:SS.FF6 %Z')",
+                "duckdb": f"SELECT STRFTIME({ts}, '%y-%j %H:%M:%S.%f %Z')",
+            },
+            write={
+                "postgres": f"SELECT TO_CHAR({ts}, 'YY-DDD HH24:MI:SS.US TZ')",
+                "oracle": f"SELECT TO_CHAR({ts}, 'YY-DDD HH24:MI:SS.FF6 %Z')",
+                "duckdb": f"SELECT STRFTIME({ts}, '%y-%j %H:%M:%S.%f %Z')",
+            },
+        )
```
VaggelisD: (1) Just to sanity check, have you checked that all of these have identical results? 

(2) We can remove the `dremio_formats` entry and add the corresponding `"dremio": ....` entry in each of the 2 `validate_all`

VaggelisD (commented):
Hey @mateuszpoleski, thank you for the contribution! Left some comments from my first pass:

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
...
+        "DDD": "%j",
+        "DD": "%d",
+        "DY": "%a",
+        "DAY": "%A",
+        # hours / minutes / seconds
+        "HH24": "%H",
+        "HH": "%I",  # 24- / 12-hour
+        "MI": "%M",
+        "SS": "%S",
+        "FFF": "%f",
```
geooo109, line 32: This maps Milliseconds to Microseconds ?

sqlglot/dialects/dremio.py
```diff
@@ -37,6 +69,12 @@ class Generator(generator.Generator):
             exp.DataType.Type.BIT: "BOOLEAN",
         }
 
+        TRANSFORMS = {
+            **generator.Generator.TRANSFORMS,
+            exp.TimeToStr: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
+            exp.ToChar: lambda self, e: self.function_fallback_sql(e),
```
geooo109, line 77: can we avoid `function_fallback_sql` ? @VaggelisD

sqlglot/dialects/dremio.py
```diff
@@ -11,9 +11,41 @@ class Dremio(Dialect):
     NULL_ORDERING = "nulls_are_last"
     SUPPORTS_VALUES_DEFAULT = False
 
+    TIME_MAPPING = {
```
geooo109, line 14: AMPM and HH12 ?

sqlglot/dialects/dremio.py
mateuszpoleski: Examples in table from Dremio docs are a bit confusing, but they’re actually padded. You can also see [here](https://docs.dremio.com/current/reference/sql/sql-functions/DATE_TIME/#:~:text=If%20you%20convert%20a%20date%20to%20text%2C%20numeric%20values%20are%20zero%20padded%20for%20you.%20For%20example%2C%20MM%20returns%2004%20for%20April.). I ran some test queries to be sure, and indeed everything gets zero padded

tests/dialects/test_dremio.py
mateuszpoleski: Removed this entirely and went with the idea of testing in `validate_all`

mateuszpoleski: Removed this entirely and went with the idea of testing in `validate_all`

tests/dialects/test_dremio.py
mateuszpoleski: Done

tests/dialects/test_dremio.py
mateuszpoleski: (1) Looks correct to me. Only small differences are in milliseconds vs microseconds (mentioned in other comment) and `%Z` token for `oracle` -- but this looks to be a missing time mapping in oracle dialect

(2) Done

sqlglot/dialects/dremio.py
mateuszpoleski: Indeed it's not ideal.. 
As far as I see Python doesn't have code for milliseconds, and this seem to make more sense than not having mapping at all? or is there better way to go about it?

sqlglot/dialects/dremio.py
mateuszpoleski: Somehow I missed these, thanks!

mateuszpoleski:
Thanks for thorough review guys!

sqlglot/dialects/dremio.py
georgesittas: It should be fine, as long as roundtrip is preserved for Dremio.

sqlglot/dialects/dremio.py
georgesittas: Seems fine, alternatives would more or less end up looking the same, e.g. `rename_func("TO_CHAR")`. Might as well just go with this.

PR #5379, opened 2025-07-08
feat(exasol): add to_date and refactored to_char functions with respect to time mapping

nnamdi16:
This involves adding [TO_DATE](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/to_date.htm) built -in function to exasol dialect in sqlglot and implementing time mappings

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -1,11 +1,45 @@
 from __future__ import annotations
 from sqlglot import exp, generator, parser
-from sqlglot.dialects.dialect import Dialect, rename_func, binary_from_function
+from sqlglot.dialects.dialect import (
+    Dialect,
+    rename_func,
+    binary_from_function,
+    build_formatted_time,
+)
 from sqlglot.helper import seq_get
 from sqlglot.generator import unsupported_args
 
 
 class Exasol(Dialect):
+    TIME_MAPPING = {
+        "y": "%Y",
+        "Y": "%Y",
+        "yyyy": "%Y",
+        "YYYY": "%Y",
+        "yy": "%y",
+        "YY": "%y",
+        "mm": "%m",
+        "MM": "%m",
+        "MONTH": "%B",
+        "MON": "%b",
+        "dd": "%d",
+        "DD": "%d",
+        "D": "%w",
+        "DAY": "%A",
+        "DY": "%a",
+        "H12": "%I",
+        "H24": "%-I",
+        "HH": "%-I",
+        "ID": "%u",
+        "vW": "%V",
+        "IW": "%V",
+        "vYYY": "%G",
+        "IYYY": "%G",
+        "MI": "%M",
+        "SS": "%S",
+        "Z": "%z",
+    }
+
     class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
@@ -26,7 +60,12 @@ class Exasol(Dialect):
             ),
             "VAR_POP": exp.VariancePop.from_arg_list,
             "APPROXIMATE_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
-            "TO_CHAR": exp.ToChar.from_arg_list,
+            "TO_CHAR": lambda args: build_formatted_time(exp.ToChar, "exasol")(
+                [seq_get(args, 0), seq_get(args, 1)]
+            ),
+            "TO_DATE": lambda args: build_formatted_time(exp.TsOrDsToDate, "exasol")(
+                [seq_get(args, 0), seq_get(args, 1)]
+            ),
         }
 
     class Generator(generator.Generator):
@@ -97,5 +136,7 @@ class Exasol(Dialect):
                 rename_func("APPROXIMATE_COUNT_DISTINCT")
             ),
             # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/to_char%20(datetime).htm
-            exp.ToChar: rename_func("TO_CHAR"),
+            exp.ToChar: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/to_date.htm
+            exp.TsOrDsToDate: lambda self, e: self.func("TO_DATE", e.this, self.format_time(e)),
         }
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -12,8 +12,6 @@ from sqlglot.generator import unsupported_args
 
 class Exasol(Dialect):
     TIME_MAPPING = {
-        "y": "%Y",
-        "Y": "%Y",
         "yyyy": "%Y",
         "YYYY": "%Y",
         "yy": "%y",
@@ -24,12 +22,11 @@ class Exasol(Dialect):
         "MON": "%b",
         "dd": "%d",
         "DD": "%d",
-        "D": "%w",
         "DAY": "%A",
         "DY": "%a",
         "H12": "%I",
-        "H24": "%-I",
-        "HH": "%-I",
+        "H24": "%H",
+        "HH": "%H",
         "ID": "%u",
         "vW": "%V",
         "IW": "%V",
@@ -39,7 +36,6 @@ class Exasol(Dialect):
         "SS": "%S",
         "uW": "%W",
         "UW": "%U",
-        "VW": "%U",
         "Z": "%z",
     }
 
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -37,6 +37,9 @@ class Exasol(Dialect):
         "IYYY": "%G",
         "MI": "%M",
         "SS": "%S",
+        "uW": "%W",
+        "UW": "%U",
+        "VW": "%U",
         "Z": "%z",
     }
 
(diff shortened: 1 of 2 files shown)


tests/dialects/test_exasol.py
```diff
@@ -276,3 +276,68 @@ def test_stringFunctions(self):
...
+        )
+        self.validate_identity(
+            "SELECT TO_CHAR(CAST('2024-07-08 13:45:00' AS TIMESTAMP), 'IYYY') AS iso_year"
+        )
+        self.validate_identity(
+            "SELECT TO_CHAR(CAST('2024-07-08 13:45:00' AS TIMESTAMP), 'MI') AS minutes"
+        )
+        self.validate_identity(
+            "SELECT TO_CHAR(CAST('2024-07-08 13:45:00' AS TIMESTAMP), 'SS') AS seconds"
+        )
```
VaggelisD: Can we clean up these identical tests with loops? E.g:

```Python3

for format in ['HH12', ..., 'SS']:
  with self.subTest(f"Testing time mapping {format} for function ..."):
     self.validate_identity(f"SELECT TO_CHAR(..., {format})"
```

sqlglot/dialects/exasol.py
```diff
@@ -26,7 +60,12 @@ class Parser(parser.Parser):
             ),
             "VAR_POP": exp.VariancePop.from_arg_list,
             "APPROXIMATE_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
-            "TO_CHAR": exp.ToChar.from_arg_list,
+            "TO_CHAR": lambda args: build_formatted_time(exp.ToChar, "exasol")(
```
VaggelisD: This can be rewritten as:

```Python3
"TO_CHAR": build_formatted_time(exp.ToChar, "exasol"),
"TO_DATE": build_formatted_time(exp.TsOrDsToDate, "exasol")
```

sqlglot/dialects/exasol.py
```diff
@@ -1,11 +1,45 @@
...
+    rename_func,
+    binary_from_function,
+    build_formatted_time,
+)
 from sqlglot.helper import seq_get
 from sqlglot.generator import unsupported_args
 
 
 class Exasol(Dialect):
+    TIME_MAPPING = {
```
VaggelisD, line 14: Are these mappings case insensitive in Exasol?

geooo109, line 14: Can you double check if all mappings are included ? for example: `WW`

sqlglot/dialects/exasol.py
```diff
@@ -1,11 +1,45 @@
...
+        "DAY": "%A",
+        "DY": "%a",
+        "H12": "%I",
+        "H24": "%-I",
+        "HH": "%-I",
+        "ID": "%u",
+        "vW": "%V",
+        "IW": "%V",
+        "vYYY": "%G",
+        "IYYY": "%G",
```
VaggelisD, line 34: Are we sure these 2 are valid Python strftime mappings?

tests/dialects/test_exasol.py
```diff
@@ -276,3 +276,68 @@ def test_stringFunctions(self):
...
+                "spark": "TO_DATE(x)",
+                "snowflake": "TO_DATE(x, 'yyyy-mm-DD')",
+                "databricks": "TO_DATE(x)",
+            },
+            read={"exasol": "TO_DATE(x, 'yyyy-MM-dd')"},
+        )
+        self.validate_all(
+            "TO_DATE(x, 'YYYY')",
+            write={
+                "exasol": "TO_DATE(x, 'YYYY')",
```
VaggelisD, line 326: Testing `exasol` both as a write and a read is superfluous, just one direction is enough. We can keep the `write` entry and remove the `read` dicts completely

tests/dialects/test_exasol.py
```diff
@@ -276,3 +276,68 @@ def test_stringFunctions(self):
...
+            write={
+                "exasol": "TO_DATE(x, 'YYYY-MM-DD')",
+                "duckdb": "CAST(x AS DATE)",
+                "hive": "TO_DATE(x)",
+                "presto": "CAST(CAST(x AS TIMESTAMP) AS DATE)",
+                "spark": "TO_DATE(x)",
+                "snowflake": "TO_DATE(x, 'yyyy-mm-DD')",
+                "databricks": "TO_DATE(x)",
+            },
+            read={"exasol": "TO_DATE(x, 'yyyy-MM-dd')"},
```
geooo109: We tend to first use the `read` and then the `write`. Moreover, you should only keep the `write` here (same on the other test with the `validate_all`)

tests/dialects/test_exasol.py
nnamdi16: Alright

sqlglot/dialects/exasol.py
nnamdi16: Alright

sqlglot/dialects/exasol.py
nnamdi16: @VaggelisD  according to the python documentation
https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
and the exasol documentation https://docs.exasol.com/db/latest/sql_references/formatmodels.htm

tests/dialects/test_exasol.py
nnamdi16: Alright

sqlglot/dialects/exasol.py
nnamdi16: Alright I would look into it

tests/dialects/test_exasol.py
nnamdi16: Alright

sqlglot/dialects/exasol.py
nnamdi16: @VaggelisD there is a safe  argument in `exp.TsOrDsToDate` that exasol doesn't support. Do I ignore specifying it like I did previously
https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/to_date.htm

georgesittas: @nnamdi16 I think Vaggelis' suggestion should just work. Not sure if you need to deal with the `safe` argument explicitly, i.e., ignore it.

sqlglot/dialects/exasol.py
georgesittas: Note: it may not be possible to map _all_ formats. Today's logic needs `TIME_MAPPING` to be 1-1, or at the very least formats that map to the same `strftime` string need to be aliases.

Since the utility of `TIME_MAPPING` is to facilitate transpilation, we should only add entries to it as long as the roundtrip to Exasol does not break.

sqlglot/dialects/exasol.py
nnamdi16: Alright @georgesittas

sqlglot/dialects/exasol.py
nnamdi16: The mappings are case insensitive @VaggelisD 
cc: @georgesittas

nnamdi16: Sure @georgesittas

sqlglot/dialects/exasol.py
```diff
@@ -1,11 +1,48 @@
...
 from sqlglot.helper import seq_get
 from sqlglot.generator import unsupported_args
 
 
 class Exasol(Dialect):
+    TIME_MAPPING = {
+        "y": "%Y",
+        "Y": "%Y",
+        "yyyy": "%Y",
+        "YYYY": "%Y",
```
georgesittas: This doesn't look right @nnamdi16. We can't have both `[Y|y]` and `[yyyy|YYYY]` map to `%Y`, because `y` an `yyyy` are not the same thing and with these entries we'll treat them as the same.

Screenshot 2025-07-09 at 7 27 43 PM

I'd suggest leaving `y` and `Y` out since strftime does not have an equivalent.

sqlglot/dialects/exasol.py
```diff
@@ -1,11 +1,48 @@
...
+        "yy": "%y",
+        "YY": "%y",
+        "mm": "%m",
+        "MM": "%m",
+        "MONTH": "%B",
+        "MON": "%b",
+        "dd": "%d",
+        "DD": "%d",
+        "D": "%w",
```
georgesittas: Given `D`'s semantics depend on an Exasol setting, I don't think we should have an entry for it. The actual 1-7 weekday is represented by `ID`, which you've correctly mapped to `%u`.

Screenshot 2025-07-09 at 7 40 24 PM

sqlglot/dialects/exasol.py
```diff
@@ -1,11 +1,48 @@
...
+        "DAY": "%A",
+        "DY": "%a",
+        "H12": "%I",
+        "H24": "%-I",
+        "HH": "%-I",
```
georgesittas: These don't look correct, we need H24 and HH to map to %H.

sqlglot/dialects/exasol.py
```diff
@@ -1,11 +1,48 @@
...
+        "ID": "%u",
+        "vW": "%V",
+        "IW": "%V",
+        "vYYY": "%G",
+        "IYYY": "%G",
+        "MI": "%M",
+        "SS": "%S",
+        "uW": "%W",
+        "UW": "%U",
+        "VW": "%U",
```
georgesittas: Are you sure these two should both map to `%U`? Albeit similar, these seem slightly different. I think `UW` matches `%U` more.

Screenshot 2025-07-09 at 7 47 53 PM

Screenshot 2025-07-09 at 7 48 43 PM

tests/dialects/test_exasol.py
```diff
@@ -276,3 +276,60 @@ def test_stringFunctions(self):
                 },
             ),
         )
+
+    def test_dateAndTimeFunctions(self):
```
georgesittas:
```suggestion
    def test_datetime_functions(self):
```

sqlglot/dialects/exasol.py
nnamdi16: Alright

geooo109 (approved):

georgesittas (approved):
Nice work 👍

PR #5385, opened 2025-07-10
Fix(fabric): cast datetimeoffset to datetime2

mattiasthalen:
This pull request refines how the Fabric dialect handles temporal data types and refactors precision capping for these types. It also updates the handling of `TIMESTAMPTZ` casts to transform it into `TIMESTAMP`.

### Enhancements to temporal type handling:

* **Precision capping logic**: Added `_cap_data_type_precision` helper function to cap the precision of temporal types to a maximum of 6 digits and default to 6 if no precision is specified. (`sqlglot/dialects/fabric.py`, [sqlglot/dialects/fabric.pyL3-R30](diffhunk://#diff-ec3f7bdb2c80d6a8318d44a7f5fcb07a5aa1a3f4f3641156eb9f14b1a0931f6aL3-R30))
* **Updated SQL generation for temporal types**: Modified `datatype_sql` and `cast_sql` methods to use `_cap_data_type_precision` for temporal types, ensuring capped precision and consistent handling of `TIMESTAMPTZ` casts. (`sqlglot/dialects/fabric.py`, [sqlglot/dialects/fabric.pyL62-R124](diffhunk://#diff-ec3f7bdb2c80d6a8318d44a7f5fcb07a5aa1a3f4f3641156eb9f14b1a0931f6aL62-R124))

### Test updates:

* **Precision capping tests**: Updated `test_precision_capping` to validate capped precision for temporal types and ensure consistent handling of `TIMESTAMPTZ` casts. (`tests/dialects/test_fabric.py`, [tests/dialects/test_fabric.pyL39-R84](diffhunk://#diff-4e9274522d52bed321c05a42ca1956bf400b6421ecd08157e6499f431670ede7L39-R84))
* **Additional `TIMESTAMPTZ` tests**: Added `test_timestamptz_without_at_time_zone` and `test_timestamptz_with_at_time_zone` to verify correct transformation of `TIMESTAMPTZ` casts with and without `AT TIME ZONE`. (`tests/dialects/test_fabric.py`, [tests/dialects/test_fabric.pyL39-R84](diffhunk://#diff-4e9274522d52bed321c05a42ca1956bf400b6421ecd08157e6499f431670ede7L39-R84))

### DIFF
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -8,6 +8,27 @@ from sqlglot.dialects.tsql import TSQL
 from sqlglot.tokens import TokenType
 
 
+def _cap_data_type_precision(expression: exp.DataType, max_precision: int = 6) -> exp.DataType:
+    """
+    Cap the precision of to a maximum of `max_precision` digits.
+    If no precision is specified, default to `max_precision`.
+    """
+
+    precision_param = expression.find(exp.DataTypeParam)
+
+    if precision_param and precision_param.this.is_int:
+        current_precision = precision_param.this.to_py()
+        target_precision = min(current_precision, max_precision)
+
+    else:
+        target_precision = max_precision
+
+    return exp.DataType(
+        this=expression.this,
+        expressions=[exp.DataTypeParam(this=exp.Literal.number(target_precision))],
+    )
+
+
 class Fabric(TSQL):
     """
     Microsoft Fabric Data Warehouse dialect that inherits from T-SQL.
@@ -73,23 +94,8 @@ class Fabric(TSQL):
                 expression.is_type(*exp.DataType.TEMPORAL_TYPES)
                 and expression.this != exp.DataType.Type.DATE
             ):
-                # Get the current precision (first expression if it exists)
-                precision_param = expression.find(exp.DataTypeParam)
-                target_precision = 6
-
-                if precision_param and precision_param.this.is_int:
-                    # Cap precision at 6
-                    current_precision = precision_param.this.to_py()
-                    target_precision = min(current_precision, 6)
-                else:
-                    # If precision exists but is not an integer, default to 6
-                    target_precision = 6
-
                 # Create a new expression with the target precision
-                expression = exp.DataType(
-                    this=expression.this,
-                    expressions=[exp.DataTypeParam(this=exp.Literal.number(target_precision))],
-                )
+                expression = _cap_data_type_precision(expression)
 
             return super().datatype_sql(expression)
 
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -87,19 +87,21 @@ class Fabric(TSQL):
         }
 
         def datatype_sql(self, expression: exp.DataType) -> str:
-            # Cap precision for temporal types
+            # Check if this is a temporal type that needs precision handling. Fabric limits temporal
+            # types to max 6 digits precision. When no precision is specified, we default to 6 digits.
             if (
                 expression.is_type(*exp.DataType.TEMPORAL_TYPES)
                 and expression.this != exp.DataType.Type.DATE
             ):
                 # Create a new expression with the capped precision
-                expression = _cap_data_type_precision(expression, 6)
+                expression = _cap_data_type_precision(expression)
 
             return super().datatype_sql(expression)
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: str | None = None) -> str:
             # Handle TIMESTAMPTZ transformation
             # Explicitly CAST(CAST(x, AS DATETIMEOFFSET) AT TIME ZONE 'UTC' AS DATETIME2)
+            # https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support
             if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
                 capped_expression = _cap_data_type_precision(expression.to)
                 precision = 6
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -18,7 +18,6 @@ def _cap_data_type_precision(expression: exp.DataType, max_precision: int = 6) -
     if precision_param and precision_param.this.is_int:
         current_precision = precision_param.this.to_py()
         target_precision = min(current_precision, max_precision)
-
     else:
         target_precision = max_precision
 
--- a/sqlglot/dialects/fabric.py
+++ b/sqlglot/dialects/fabric.py
@@ -99,33 +99,47 @@ class Fabric(TSQL):
             return super().datatype_sql(expression)
 
         def cast_sql(self, expression: exp.Cast, safe_prefix: str | None = None) -> str:
-            # Handle TIMESTAMPTZ transformation
-            # Explicitly CAST(CAST(x, AS DATETIMEOFFSET) AT TIME ZONE 'UTC' AS DATETIME2)
+            # Cast to DATETIMEOFFSET if inside an AT TIME ZONE expression
             # https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support
             if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
-                capped_expression = _cap_data_type_precision(expression.to)
-                precision = 6
+                at_time_zone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
 
-                capped_precision = capped_expression.find(exp.DataTypeParam)
-                if capped_precision:
-                    precision = capped_precision.this.to_py()
+                # Return normal cast, ff the expression is not in an AT TIME ZONE context
+                if not isinstance(at_time_zone, exp.AtTimeZone):
+                    return super().cast_sql(expression, safe_prefix)
 
-                # Hard coded cast to bypass SQLGlot's default handling of TIMESTAMPTZ
-                inner_cast = f"CAST({expression.this} AS DATETIMEOFFSET({precision}))"
-
-                at_time_zone = exp.AtTimeZone(this=inner_cast, zone=exp.Literal.string("UTC"))
-
-                outer_cast = exp.Cast(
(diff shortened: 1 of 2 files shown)


mattiasthalen:
@VaggelisD, are you available for a review?

sqlglot/dialects/fabric.py
```diff
@@ -59,48 +79,49 @@ class Generator(TSQL.Generator):
...
+            exp.DataType.Type.TIMESTAMPTZ: "DATETIME2",
             exp.DataType.Type.TINYINT: "SMALLINT",
             exp.DataType.Type.UTINYINT: "SMALLINT",
             exp.DataType.Type.UUID: "VARBINARY(MAX)",
             exp.DataType.Type.XML: "VARCHAR",
         }
 
         def datatype_sql(self, expression: exp.DataType) -> str:
-            # Check if this is a temporal type that needs precision handling. Fabric limits temporal
-            # types to max 6 digits precision. When no precision is specified, we default to 6 digits.
```
VaggelisD: Can we reintroduce this comment and a variable in `fabric.py` such as `MAX_PRECISION`? 

Otherwise, it looks like a magic number

sqlglot/dialects/fabric.py
```diff
@@ -59,48 +79,49 @@ class Generator(TSQL.Generator):
...
-                    # If precision exists but is not an integer, default to 6
-                    target_precision = 6
-
-                # Create a new expression with the target precision
-                expression = exp.DataType(
-                    this=expression.this,
-                    expressions=[exp.DataTypeParam(this=exp.Literal.number(target_precision))],
-                )
+                # Create a new expression with the capped precision
+                expression = _cap_data_type_precision(expression, 6)
```
VaggelisD: The `max_prevision=6` value can be omitted here

sqlglot/dialects/fabric.py
```diff
@@ -59,48 +79,49 @@ class Generator(TSQL.Generator):
...
+                # Hard coded cast to bypass SQLGlot's default handling of TIMESTAMPTZ
+                inner_cast = f"CAST({expression.this} AS DATETIMEOFFSET({precision}))"
+
+                at_time_zone = exp.AtTimeZone(this=inner_cast, zone=exp.Literal.string("UTC"))
+
+                outer_cast = exp.Cast(
+                    this=at_time_zone,
+                    to=exp.DataType(
+                        this=exp.DataType.Type.TIMESTAMP,
+                        expressions=[exp.DataTypeParam(this=exp.Literal.number(precision))],
```
VaggelisD: Do we have to pass in precision here as well?

mattiasthalen:
> Hey @mattiasthalen, thanks for the contribution once again. So this combination of casts leads to the same results as other engines in SQLMesh, right?

Hey, not truly sure here. Depends on how the other non-tz engines does it. Do they cast to UTC before returning a timestamp? If not, should they?

That's the short version of what this does, cast to UTC and converts to timestamp.

From what I can see, most convert to session tz first, except SQL Server, it just truncates 😳

So if anything, the UTC step might need to be dropped to at least be consistent with SQL Server.

sqlglot/dialects/fabric.py
```diff
@@ -59,48 +79,49 @@ class Generator(TSQL.Generator):
...
-            # Special handling for TIMESTAMPTZ casts in Fabric
-            if expression.to and expression.to.is_type(exp.DataType.Type.TIMESTAMPTZ):
-                attimezone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
-                if not isinstance(attimezone, exp.AtTimeZone):
-                    # We're not inside an AT TIME ZONE, so wrap the cast in AT TIME ZONE 'UTC'
-                    at_time_zone = exp.AtTimeZone(this=expression, zone=exp.Literal.string("UTC"))
-                    return self.sql(at_time_zone)
+        def cast_sql(self, expression: exp.Cast, safe_prefix: str | None = None) -> str:
+            # Handle TIMESTAMPTZ transformation
+            # Explicitly CAST(CAST(x, AS DATETIMEOFFSET) AT TIME ZONE 'UTC' AS DATETIME2)
```
erindru: Are you able to link to the Microsoft documentation saying why we have to do this? So we have some historical context in 1 years time when they update the behaviour on the Fabric side

mattiasthalen: It's in the class docstring, under references. But I could add it here as well if you want.

mattiasthalen: There's another document about this and the behaviour: https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support

mattiasthalen: But what do you think, should we do it "proper" and convert to UTC, or have the same behavior as SQL Server and truncate the tz info?

sqlglot/dialects/fabric.py
mattiasthalen: Yes, otherwise we might get a mismatch between the inner and outer cast, if it's below the max precision.

sqlglot/dialects/fabric.py
mattiasthalen: Not sure how that went missing. Must have happened during a merge conflict :S

sqlglot/dialects/fabric.py
erindru: Yeah, that's the link I meant and I think it would be handy to have it in context.

I had a bit of a play around:

```sql
with base_data as (
	select cast('2025-01-01 13:00:00 -08:00' as datetimeoffset) as v
)
select
	-- Output: 2025-01-01 13:00:00.0000000 -08:00
        -- original value, DATETIMEOFFSET type in PST
	v as v_orig,
	
	-- Output: 2025-01-01 13:00:00.0000000 -08:00
	-- same as v_orig because its the same timezone
	v at time zone 'Pacific Standard Time' as v_pst,
	
	-- Output: 2025-01-01 13:00:00.0000000
	-- truncates timezone but otherwise shows the same value
	cast(v as datetime2) as v_datetime2,
	
	-- Output: 2025-01-01 13:00:00.0000000
	-- same as above, AT TIME ZONE already matches existing timezone so just truncates timezone
	cast(v at time zone 'Pacific Standard Time' as datetime2) as v_pst_dt2,	
	
	-- Output: 2025-01-01 13:00:00.0000000 +13:00
	-- shows that taking a datetime2 and adding AT TIME ZONE just sets the timezone rather than transforming anything
	cast(v as datetime2) at time zone 'New Zealand Standard Time' as v_datetime2_nzt,
	
	-- Output: 2025-01-01 21:00:00.0000000
	-- shows that AT TIME ZONE does change the value before it's truncated by the cast to datetime2 
	cast(v at time zone 'UTC' as datetime2) as v_utc
from base_data
```

So, based on that, I think we should match the SQL Server behaviour. Does the following make sense?
 - If the target timezone is known, we `cast(val at time zone '<target timezone>' as datetime2)`
 - If the target timezone is not known, we just truncate and call it a day, eg `cast(val as datetime2)` rather than converting to UTC first

mattiasthalen: Yes, I'll make that happen ☺️

mattiasthalen: One thing I noticed is that this affects the result of the fabric integration test in sqlmesh:
```bash
test_value_normalization[fabric-Type.TIMESTAMPTZ-input_data4-expected_results4] - AssertionError: assert '2023-01-01 00:00:00.000000' == '2023-01-01 08:00:00.000000'
```

Should we skip that test scenario, like it's done for Trino?
```python
# Skip TIMESTAMPTZ tests for engines that don't support it
if column_type == exp.DataType.Type.TIMESTAMPTZ:
    if ctx.dialect == "trino" and ctx.engine_adapter.current_catalog_type == "hive":
        pytest.skip("Trino on Hive doesn't support TIMESTAMP WITH TIME ZONE fields")
    if ctx.dialect == "fabric":
        pytest.skip("Fabric doesn't support TIMESTAMP WITH TIME ZONE fields")
 ```

Comment (erindru) [reply sqlglot/dialects/fabric.py]: Yes, I think that logic is correct. Since you cant create a table with a column of type `DATETIMEOFFSET`, and there is no other timezone-aware type you can use, Fabric essentially doesn't support this.

**Review by erindru — APPROVED** [review]
This logic looks correct to me, although it boggles my mind that Fabric Warehouse doesn't have a timezone-aware timestamp type that can be used in table creation.

Maybe they were assuming that tables would be created with Lakehouse (since Spark does have a timezone-aware timestamp type) and just read via Warehouse :shrug:

File: sqlglot/dialects/fabric.py [hunk 5]
```diff
@@ -73,37 +93,54 @@ def datatype_sql(self, expression: exp.DataType) -> str:
...
-                    # We're not inside an AT TIME ZONE, so wrap the cast in AT TIME ZONE 'UTC'
-                    at_time_zone = exp.AtTimeZone(this=expression, zone=exp.Literal.string("UTC"))
-                    return self.sql(at_time_zone)
+        def cast_sql(self, expression: exp.Cast, safe_prefix: str | None = None) -> str:
+            # Cast to DATETIMEOFFSET if inside an AT TIME ZONE expression
+            # https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql#microsoft-fabric-support
+            if expression.is_type(exp.DataType.Type.TIMESTAMPTZ):
+                at_time_zone = expression.find_ancestor(exp.AtTimeZone, exp.Select)
+
+                # Return normal cast, ff the expression is not in an AT TIME ZONE context
```
georgesittas:
```suggestion
                # Return normal cast, if the expression is not in an AT TIME ZONE context
```

sqlglot/dialects/fabric.py
```diff
@@ -1,13 +1,33 @@
...
+    Cap the precision of to a maximum of `max_precision` digits.
+    If no precision is specified, default to `max_precision`.
+    """
+
+    precision_param = expression.find(exp.DataTypeParam)
+
+    if precision_param and precision_param.this.is_int:
+        current_precision = precision_param.this.to_py()
+        target_precision = min(current_precision, max_precision)
+
```
georgesittas:
```suggestion
```

PR #5386, opened 2025-07-10
Fix(fabric): ensure varchar max

mattiasthalen:
This PR adds precision capping and defaulting for `VARCHAR`:

- `VARCHAR` -> `VARCHAR(MAX)`
- `VARCHAR(1...8000)` -> `VARCHAR(1...8000)`
- `VARCHAR(8001...)` -> `VARCHAR(MAX)`

>**NOTE!**
>Depends on #5385, will mark this as ready for review as soon as that is merged

### DIFF
--- a/tests/dialects/test_fabric.py
+++ b/tests/dialects/test_fabric.py
@@ -36,44 +36,52 @@ class TestFabric(Validator):
         self.validate_identity("CAST(x AS TIME)", "CAST(x AS TIME(6))")
         self.validate_identity("CAST(x AS DATETIME2)", "CAST(x AS DATETIME2(6))")
         self.validate_identity(
-            "CAST(x AS DATETIMEOFFSET) AT TIME ZONE 'UTC'",
-            "CAST(x AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC'",
+            "CAST(x AS TIMESTAMPTZ)",
+            "CAST(CAST(x AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC' AS DATETIME2(6))",
         )
 
         # Precision <= 6 should be preserved
         self.validate_identity("CAST(x AS TIME(3))", "CAST(x AS TIME(3))")
         self.validate_identity("CAST(x AS DATETIME2(3))", "CAST(x AS DATETIME2(3))")
-        self.validate_identity("CAST(x AS DATETIMEOFFSET(3)) AT TIME ZONE 'UTC'")
+        self.validate_identity(
+            "CAST(x AS TIMESTAMPTZ(3))",
+            "CAST(CAST(x AS DATETIMEOFFSET(3)) AT TIME ZONE 'UTC' AS DATETIME2(3))",
+        )
 
         self.validate_identity("CAST(x AS TIME(6))", "CAST(x AS TIME(6))")
         self.validate_identity("CAST(x AS DATETIME2(6))", "CAST(x AS DATETIME2(6))")
-        self.validate_identity("CAST(x AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC'")
+        self.validate_identity(
+            "CAST(x AS TIMESTAMPTZ(6))",
+            "CAST(CAST(x AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC' AS DATETIME2(6))",
+        )
 
         # Precision > 6 should be capped at 6
         self.validate_identity("CAST(x AS TIME(7))", "CAST(x AS TIME(6))")
         self.validate_identity("CAST(x AS DATETIME2(7))", "CAST(x AS DATETIME2(6))")
         self.validate_identity(
-            "CAST(x AS DATETIMEOFFSET(7)) AT TIME ZONE 'UTC'",
-            "CAST(x AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC'",
+            "CAST(x AS TIMESTAMPTZ(7))",
+            "CAST(CAST(x AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC' AS DATETIME2(6))",
         )
 
         self.validate_identity("CAST(x AS TIME(9))", "CAST(x AS TIME(6))")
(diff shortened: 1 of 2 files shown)


georgesittas:
Hey @mattiasthalen 👋 

I went ahead and resolved the conflicts caused by https://github.com/tobymao/sqlglot/pull/5385 for you + marked this as ready for review. Let me know if there's anything else pending for this PR or if it's good to go on your end.

tests/dialects/test_fabric.py
```diff
@@ -28,7 +28,23 @@ def test_type_mappings(self):
         self.validate_identity("CAST(x AS UTINYINT)", "CAST(x AS SMALLINT)")
         self.validate_identity("CAST(x AS UUID)", "CAST(x AS VARBINARY(MAX))")
         self.validate_identity("CAST(x AS VARIANT)", "CAST(x AS SQL_VARIANT)")
-        self.validate_identity("CAST(x AS XML)", "CAST(x AS VARCHAR)")
+        self.validate_identity("CAST(x AS XML)", "CAST(x AS VARCHAR(MAX))")
+
+    def test_varchar_precision_capping(self):
+        """Test that VARCHAR precision is capped at MAX"""
+        # No precision specified should default to MAX
+        self.validate_identity("CAST(x AS VARCHAR)", "CAST(x AS VARCHAR(MAX))")
```
georgesittas, line 36: Can you remind me why this is necessary? Preemptively maxing the `VARCHAR` seems a bit weird when roundtripping in Fabric.

mattiasthalen: The issue is that it seems Fabric resolves varchar to varchar(1) or similar. So during the fabric integration tests, there were failures in the table diff tests. Kept getting "string/binary will be truncated"

mattiasthalen:
@georgesittas nothing pending from my end. Except maybe a rebase, as this is attached to the previous PR

tests/dialects/test_fabric.py
georgesittas: Yeah, that makes sense. Do we construct `VARCHAR` casts in SQLMesh, i.e., without a precision?

Imo we should ideally preserve what we have in the input for Fabric, e.g. by using a separate arg in `DataType` or something similar, which would serve as a signal for "this was parsed with fabric, so if there's no precision then that was the user's intention".

mattiasthalen: You mean specifically for the fabric engine? I don't think so.

mattiasthalen: > Yeah, that makes sense. Do we construct `VARCHAR` casts in SQLMesh, i.e., without a precision?
> 
> 
> 
> Imo we should ideally preserve what we have in the input for Fabric, e.g. by using a separate arg in `DataType` or something similar, which would serve as a signal for "this was parsed with fabric, so if there's no precision then that was the user's intention".

Could be that it was only the integration tests that hade this unprecise cast. Because I haven't seen any issues with the different model types. Well, unless I'm unprecise in my own casts.

georgesittas: > Could be that it was only the integration tests that hade this unprecise cast.

Possibly, yes. It wouldn't surprise me, though, if SQLMesh contained casts to `VARCHAR` without a precision in certain places.

Just to be clear: if we merge this PR as is, then `VARCHAR` references will always be converted into `VARCHAR(MAX)` for Fabric. This changes the semantics of Fabric queries, so imo it is not technically correct. I'd like to try the following approach first:

- read, write dialect is fabric: `cast(x as varchar)` is preserved
- read is not fabric, write is fabric: `cast(x as varchar)` is converted into `cast(x as varchar(max))`

Also, just to double-check: does the following query truncate the input to `'a'` for Fabric? Or is the issue isolated to DDLs with `VARCHAR` columns vs `VARCHAR(<precision>)` ones?

```
select cast('ab' as varchar);
```

mattiasthalen: I'll have a look this evening ☺️

mattiasthalen: Just ran a quick test

```sql
CREATE TABLE test.abc(
  abc VARCHAR
)
```
image

So it's the DDL that imposes  varchar -> varchar(1)

This works just fine, from a query standpoint:
```sql
SELECT CAST('abcdefghijklmnopqrstuvwxyz' AS VARCHAR)
```

georgesittas: If this only happens in DDLs and not in casts or other places where types can appear, then I think we should only transform `DataType` instances under DDLs to include precision, not everything. We can do this with a rule applied via `transforms.preprocess` on `exp.Create`, etc.

georgesittas: T-SQL appears to be similar in this regard:

```
1> SELECT CAST('ab' AS VARCHAR)
2> GO
                              
------------------------------
ab                            

(1 rows affected)
1> CREATE TABLE t (c VARCHAR)
2> GO
1> INSERT INTO t VALUES ('ab')
2> GO
Msg 2628, Level 16, State 1, Server 9590817fd1a5, Line 1
String or binary data would be truncated in table 'master.dbo.t', column 'c'. Truncated value: 'a'.
The statement has been terminated.
```

mattiasthalen: Nice! Well, not nice that it happens, just nice that's it's not yet another Fabric quirk 😅

I'll do a refactor of this PR tonight. Should we go for varchar(max) in the ddl, or something else?

georgesittas: I think it's safe to go with `VARCHAR(MAX)`, assuming the read dialect is not T-SQL/Fabric (see previous [reply](https://github.com/tobymao/sqlglot/pull/5386#discussion_r2207314028)).

mattiasthalen:
I'm thinking we should probably close this and create a new PR for adressing the varchar > varchar(1) in tsql ddl.

FYI, I'm going on vacation for the rest of the week. Will still be able to answer questions, won't have any laptop with me though.

georgesittas:
Sounds good– happy to revisit later! Have a lovely vacation :)

mattiasthalen:
Created an issue for tracking this: https://github.com/tobymao/sqlglot/issues/5413

PR #5387, opened 2025-07-10
fix(snowflake): transpile bigquery CURRENT_DATE with timezone

geooo109:
This PR adds transpilation support for `CURRENT_DATE` with timezone of BigQuery to Snowflake.

Snowflake doesn't support `CURRENT_DATE('timezone')`, so we implement this by using the `CURRENT_TIMESTAMP` combined with `CONVERT_TIMEZONE` and `CAST`.

**DOCS**
[BigQuery CURRENT_DATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#current_date)
[Snowflake CURRENT_TIMESTAMP](https://docs.snowflake.com/en/sql-reference/functions/current_timestamp)
[Snowflake CONVERT_TIMEZONE](https://docs.snowflake.com/en/sql-reference/functions/convert_timezone)
[Snowflake TO_DATE](https://docs.snowflake.com/en/sql-reference/functions/to_date)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1509,4 +1509,8 @@ class Snowflake(Dialect):
             if not zone:
                 return super().currentdate_sql(expression)
 
-            return f"TO_DATE(CONVERT_TIMEZONE({zone}, CURRENT_TIMESTAMP()))"
+            expr = exp.Cast(
+                this=exp.ConvertTimezone(target_tz=zone, timestamp=exp.CurrentTimestamp()),
+                to=exp.DataType(this=exp.DataType.Type.DATE),
+            )
+            return self.sql(expr)
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1503,3 +1503,10 @@ class Snowflake(Dialect):
                     return self.sql(first_expr.subquery())
 
             return inline_array_sql(self, expression)
+
+        def currentdate_sql(self, expression: exp.CurrentDate) -> str:
+            zone = self.sql(expression, "this")
+            if not zone:
+                return super().currentdate_sql(expression)
+
+            return f"TO_DATE(CONVERT_TIMEZONE({zone}, CURRENT_TIMESTAMP()))"
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1392,9 +1392,14 @@ LANGUAGE js AS
         )
         self.validate_all(
             "CURRENT_DATE('UTC')",
+            read={
+                "bigquery": "CURRENT_DATE('UTC')",
+            },
             write={
+                "bigquery": "CURRENT_DATE('UTC')",
                 "mysql": "CURRENT_DATE AT TIME ZONE 'UTC'",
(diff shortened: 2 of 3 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -1503,3 +1503,10 @@ def array_sql(self, expression: exp.Array) -> str:
                     return self.sql(first_expr.subquery())
 
             return inline_array_sql(self, expression)
+
+        def currentdate_sql(self, expression: exp.CurrentDate) -> str:
+            zone = self.sql(expression, "this")
+            if not zone:
+                return super().currentdate_sql(expression)
+
+            return f"TO_DATE(CONVERT_TIMEZONE({zone}, CURRENT_TIMESTAMP()))"
```
VaggelisD: We should try not to hardcode SQL like that. Instead, lets build the expressions manually to leverage their generators e.g

```Python3

expr = exp.Cast(this=exp.ConvertTimeZone(this=zone, expression=exp.CurrentTimestamp), to=exp.DataType.DATE)

return self.sql(expr)
```

tests/dialects/test_bigquery.py
```diff
@@ -1758,6 +1758,13 @@ def test_bigquery(self):
         self.validate_identity("ARRAY_FIRST(['a', 'b'])")
         self.validate_identity("ARRAY_LAST(['a', 'b'])")
 
+        self.validate_all(
+            "CURRENT_DATE('timezone')",
+            write={
+                "snowflake": "TO_DATE(CONVERT_TIMEZONE('timezone', CURRENT_TIMESTAMP()))",
```
VaggelisD: Try to always include roundtrip tests, never hurts to have them.

Also a slight nit: Prefer using values that make the examples runnable; There isn't really a benefit in using `'timezone'`, but if it was a proper one we could potentially test it directly.

tests/dialects/test_snowflake.py
```diff
@@ -1167,6 +1167,9 @@ def test_snowflake(self):
             },
         )
 
+        self.validate_identity("CURRENT_DATE")
+        self.validate_identity("TO_DATE(CONVERT_TIMEZONE('timezone', CURRENT_TIMESTAMP()))")
```
VaggelisD: What's the purpose of these tests?

sqlglot/dialects/snowflake.py
geooo109: Yep I see, thanks

tests/dialects/test_bigquery.py
geooo109: Yep, added

tests/dialects/test_snowflake.py
geooo109: There is no need for them, removed them.

tests/dialects/test_bigquery.py
```diff
@@ -1392,9 +1392,14 @@ def test_bigquery(self):
         )
         self.validate_all(
             "CURRENT_DATE('UTC')",
+            read={
+                "bigquery": "CURRENT_DATE('UTC')",
+            },
```
VaggelisD: This is superfluous, you added a BQ -> BQ test to the `write` dict too

geooo109: Yep, my bad, fixed it.

PR #5388, opened 2025-07-11
fix(snowflake): transpile bigquery GENERATE_DATE_ARRAY with column access

geooo109:
This PR adds support for the transpilation of `GENERATE_DATE_ARRAY` from BigQuery to Snowflake, when a column is an argument. 
In the previous implementation the following BigQuery query:
```
SELECT id, mnth FROM t CROSS JOIN UNNEST(GENERATE_DATE_ARRAY(start_month, DATE_TRUNC(CURRENT_DATE(), MONTH), INTERVAL 1 MONTH)) AS mnth
```
Would be transpiled into Snowflake: 
```
SELECT
  id,
  mnth
FROM t
CROSS JOIN (
  SELECT
    DATEADD(MONTH, CAST(mnth AS INT), CAST(start_month AS DATE)) AS mnth
  FROM TABLE(FLATTEN(INPUT => ARRAY_GENERATE_RANGE(
    0,
    (
      DATEDIFF(MONTH, start_month, DATE_TRUNC('MONTH', CURRENT_DATE)) + 1 - 1
    ) + 1
  ))) AS _t1(seq, key, path, index, mnth, this)
) AS _t0(mnth)
```
Which is wrong. First, the `CROSS JOIN` doesn't allow access to the column `start_month` of table `t`. 
Moreover, based on the snowflake [docs](https://docs.snowflake.com/en/user-guide/querying-subqueries#usage-notes), subqueries with a correlation inside of `FLATTEN` are currently unsupported.

In order to solve this, I suggest to leave the basic implementation (when the `GENERATE_DATE_ARRAY` contains constant values as it is) and change the behavior for column correlation.

This PR does the following transpilation: 
```
SELECT
  id,
  DATEADD(MONTH, CAST(mnth AS INT), CAST(start_month AS DATE)) AS mnth
FROM t, LATERAL FLATTEN(INPUT => ARRAY_GENERATE_RANGE(
  0,
  (
    DATEDIFF(MONTH, start_month, DATE_TRUNC('MONTH', CURRENT_DATE)) + 1 - 1
  ) + 1
)) AS _t0(seq, key, path, index, mnth, this)
```
The basic idea is up, but I believe the implementation can be better.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -264,7 +264,7 @@ def _unnest_generate_date_array(unnest: exp.Unnest) -> None:
                 else sequence_value_name.name
             )
             for column in select.selects:
-                if column.this.name == replace_column_name:
+                if column.this.name.lower() == replace_column_name.lower():
                     column.replace(date_add.as_(column.alias_or_name))
             lateral = exp.Lateral(this=unnest_parent.this.pop())
             unnest_parent.replace(exp.Join(this=lateral))
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -263,9 +263,17 @@ def _unnest_generate_date_array(unnest: exp.Unnest) -> None:
                 if isinstance(sequence_value_name, str)
                 else sequence_value_name.name
             )
-            for column in select.selects:
-                if column.this.name.lower() == replace_column_name.lower():
-                    column.replace(date_add.as_(column.alias_or_name))
+
+            scope = build_scope(select)
+            if scope:
+                for column in scope.columns:
+                    if column.name.lower() == replace_column_name.lower():
+                        column.replace(
+                            date_add.as_(replace_column_name)
+                            if isinstance(column.parent, exp.Select)
+                            else date_add
+                        )
+
             lateral = exp.Lateral(this=unnest_parent.this.pop())
             unnest_parent.replace(exp.Join(this=lateral))
     else:
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -245,7 +245,7 @@ def _unnest_generate_date_array(unnest: exp.Unnest) -> None:
     # We'll add the next sequence value to the starting date and project the result
     date_add = _build_date_time_add(exp.DateAdd)(
         [unit, exp.cast(sequence_value_name, "int"), exp.cast(start, "date")]
-    ).as_(sequence_value_name)
+    )
 
     # We use DATEDIFF to compute the number of sequence values needed
     number_sequence = Snowflake.Parser.FUNCTIONS["ARRAY_GENERATE_RANGE"](
@@ -253,7 +253,27 @@ def _unnest_generate_date_array(unnest: exp.Unnest) -> None:
     )
 
     unnest.set("expressions", [number_sequence])
-    unnest.replace(exp.select(date_add).from_(unnest.copy()).subquery(unnest_alias))
+
+    unnest_parent = unnest.parent
+    if isinstance(unnest_parent, exp.Join):
+        select = unnest_parent.parent
+        if isinstance(select, exp.Select):
+            replace_column_name = (
+                sequence_value_name
+                if isinstance(sequence_value_name, str)
+                else sequence_value_name.name
+            )
+            for column in select.selects:
+                if column.this.name == replace_column_name:
+                    column.replace(date_add.as_(column.alias_or_name))
+            lateral = exp.Lateral(this=unnest_parent.this.pop())
+            unnest_parent.replace(exp.Join(this=lateral))
+    else:
+        unnest.replace(
+            exp.select(date_add.as_(sequence_value_name))
+            .from_(unnest.copy())
+            .subquery(unnest_alias)
+        )
 
 
 def _transform_generate_date_array(expression: exp.Expression) -> exp.Expression:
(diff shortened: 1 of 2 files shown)


tobymao:
why did you decided to use this method only for the new one?

sqlglot/dialects/snowflake.py
```diff
@@ -245,15 +245,35 @@ def _unnest_generate_date_array(unnest: exp.Unnest) -> None:
...
+    if isinstance(unnest_parent, exp.Join):
+        select = unnest_parent.parent
+        if isinstance(select, exp.Select):
+            replace_column_name = (
+                sequence_value_name
+                if isinstance(sequence_value_name, str)
+                else sequence_value_name.name
+            )
+            for column in select.selects:
+                if column.this.name == replace_column_name:
```
tobymao: does case matter here?

geooo109:
> why did you decided to use this method only for the new one?

If we decide this is ok, we can merge the impls. I wanted to implement this and not break things around.

sqlglot/dialects/snowflake.py
geooo109: yep, will fix it thanks

tobymao:
@georgesittas can you take a look

georgesittas:
yep will check tomorrow in more detail 👍

sqlglot/dialects/snowflake.py
```diff
@@ -245,15 +245,35 @@ def _unnest_generate_date_array(unnest: exp.Unnest) -> None:
...
+        select = unnest_parent.parent
+        if isinstance(select, exp.Select):
+            replace_column_name = (
+                sequence_value_name
+                if isinstance(sequence_value_name, str)
+                else sequence_value_name.name
+            )
+            for column in select.selects:
+                if column.this.name.lower() == replace_column_name.lower():
+                    column.replace(date_add.as_(column.alias_or_name))
```
georgesittas: What happens when:

- The unnest alias is not a top-level select, e.g. it's nested under another expression (projection or not)
- There are no columns in the query, e.g. you have a simple `SELECT *` or something.

tests/dialects/test_bigquery.py
```diff
@@ -2661,3 +2647,33 @@ def test_unnest_with_offset(self):
...
+            write={
+                "bigquery": "SELECT id, mnth FROM t CROSS JOIN UNNEST(GENERATE_DATE_ARRAY(start_month, DATE_TRUNC(CURRENT_DATE, MONTH), INTERVAL '1' MONTH)) AS mnth",
+                "snowflake": "SELECT id, DATEADD(MONTH, CAST(mnth AS INT), CAST(start_month AS DATE)) AS mnth FROM t, LATERAL FLATTEN(INPUT => ARRAY_GENERATE_RANGE(0, (DATEDIFF(MONTH, start_month, DATE_TRUNC('MONTH', CURRENT_DATE)) + 1 - 1) + 1)) AS _t0(seq, key, path, index, mnth, this)",
+            },
+        )
+        self.validate_all(
+            "SELECT id, mnth AS a_mnth FROM t CROSS JOIN UNNEST(GENERATE_DATE_ARRAY(start_month, DATE_TRUNC(CURRENT_DATE, MONTH), INTERVAL '1' MONTH)) AS mnth",
+            write={
+                "bigquery": "SELECT id, mnth AS a_mnth FROM t CROSS JOIN UNNEST(GENERATE_DATE_ARRAY(start_month, DATE_TRUNC(CURRENT_DATE, MONTH), INTERVAL '1' MONTH)) AS mnth",
+                "snowflake": "SELECT id, DATEADD(MONTH, CAST(mnth AS INT), CAST(start_month AS DATE)) AS a_mnth FROM t, LATERAL FLATTEN(INPUT => ARRAY_GENERATE_RANGE(0, (DATEDIFF(MONTH, start_month, DATE_TRUNC('MONTH', CURRENT_DATE)) + 1 - 1) + 1)) AS _t0(seq, key, path, index, mnth, this)",
```
georgesittas, line 2679: I feel like this works by accident? If you had something more complex than `mnth AS a_mnth` the `column.this.name.lower() == ...` check wouldn't succeed and you wouldn't really replace the `mnth` ref with the `DATEADD` expression.

tests/dialects/test_bigquery.py
```diff
@@ -2661,3 +2647,33 @@ def test_unnest_with_offset(self):
...
+            },
+        )
+        self.validate_all(
+            "SELECT GENERATE_DATE_ARRAY('2016-10-05', '2016-10-08', INTERVAL '1' MONTH)",
+            write={
+                "bigquery": "SELECT GENERATE_DATE_ARRAY('2016-10-05', '2016-10-08', INTERVAL '1' MONTH)",
+                "duckdb": "SELECT CAST(GENERATE_SERIES(CAST('2016-10-05' AS DATE), CAST('2016-10-08' AS DATE), INTERVAL '1' MONTH) AS DATE[])",
+            },
+        )
+        self.validate_all(
```
georgesittas, line 2666: Are these transpiled correctly do DuckDB as well? Let's include it as a `write` target, if so.

geooo109: Yep, will add it for DuckDB.

sqlglot/dialects/snowflake.py
georgesittas: We discussed in Slack and concluded that we'll deal with `SELECT *` separately, as it is not directly related to this work.

PR #5395, opened 2025-07-14
feat(duckdb): support column access via index

### LINKED ISSUE #5392: duckdb: support positional expressions
Duckdb supports positional expressions, see the [grammar here](https://github.com/duckdb/duckdb/blob/4b9b91aa88dfaed7244b0965bb127642cfd07c18/extension/autocomplete/grammar/expression.gram#L49), and this example:

```
D select #2,#1 from (values (1,'a'));
┌─────────┬───────┐
│  col1   │ col0  │
│ varchar │ int32 │
├─────────┼───────┤
│ a       │   1   │
└─────────┴───────┘
```

However sqlglot cannot parse this:

```python
>>> sqlglot.parse_one("""select #2,#1 from (values (1,'a'))""", read="duckdb").sql()
  File "/Users/tekumara/code/fakesnow/.venv/lib/python3.9/site-packages/sqlglot/__init__.py", line 139, in parse_one
  File "/Users/tekumara/code/fakesnow/.venv/lib/python3.9/site-packages/sqlglot/dialects/dialect.py", line 1056, in parse
  File "/Users/tekumara/code/fakesnow/.venv/lib/python3.9/site-packages/sqlglot/parser.py", line 1592, in parse
    return self._parse(
  File "/Users/tekumara/code/fakesnow/.venv/lib/python3.9/site-packages/sqlglot/parser.py", line 1664, in _parse
  File "/Users/tekumara/code/fakesnow/.venv/lib/python3.9/site-packages/sqlglot/parser.py", line 1705, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 8.
  select #2,#1 from (values (1,'a'))
```

geooo109:
Fixes  #5392

ThIs PR adds parsing support for positional columns. 
Example: 
```
SELECT #2, #1 FROM (VALUES (1, 'a'))

Select(
  expressions=[
    PositionalColumn(
      this=Literal(this=2, is_string=False)),
    PositionalColumn(
      this=Literal(this=1, is_string=False))],
  from=From(
    this=Values(
      expressions=[
        Tuple(
          expressions=[
            Literal(this='a', is_string=True)])])))
```

**DOCS**
[DuckDB](https://duckdb.org/docs/stable/sql/statements/select)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6091,7 +6091,7 @@ class PosexplodeOuter(Posexplode, ExplodeOuter):
 
 
 class PositionalColumn(Expression):
-    arg_types = {"this": True}
+    pass
 
 
 class Unnest(Func, UDTF):


geooo109:
I introduced a new node for this repr. Moreover, the `primary parser` in duckdb changed in order to support this. 
Is it better to add this parsing in the `PRIMARY_PARSERS` ?
Moreover, should we tokenize `#` and the number as an `VAR` token ?

sqlglot/dialects/duckdb.py
```diff
@@ -619,6 +619,12 @@ def _parse_attach_option() -> exp.AttachOption:
         def _parse_show_duckdb(self, this: str) -> exp.Show:
             return self.expression(exp.Show, this=this)
 
+        def _parse_primary(self) -> t.Optional[exp.Expression]:
+            if self._match(TokenType.HASH) and self._match(TokenType.NUMBER):
```
tobymao: what happens if it's just a hash? you've already consumed it haven't you?

tobymao: shouldn't you use _match_pair

geooo109: yep, you have right, nice catch.

sqlglot/expressions.py
```diff
@@ -6090,6 +6090,10 @@ class PosexplodeOuter(Posexplode, ExplodeOuter):
     pass
 
 
+class PositionalColumn(Expression):
+    arg_types = {"this": True}
```
georgesittas: No need for `arg_types` here, you inherit from `Expression`.
```suggestion
    pass
```

georgesittas (approved):

PR #5402, opened 2025-07-15
feat: support Doris range partitioning

xinge-ji:
## Add Support for Doris Range Partitioning

This PR adds support for multiple **range partitioning** strategies in Apache Doris, including both standard and batch range partition definitions.

### Supported Partition Types

* **Fixed Range**:
  Defines partitions as left-closed, right-open intervals.
  Example:

  ```sql
  PARTITION `p201703` VALUES [("2017-03-01"), ("2017-04-01"))
  ```

* **Less Than**:
  Specifies the upper bound of the partition using the `LESS THAN` syntax.
  Example:

  ```sql
  PARTITION `p201701` VALUES LESS THAN ("2017-02-01")
  ```

* **Batch Range**:
  Batch creation of range partitions for numeric or date/time columns using intervals.
  Partitions are defined as left-closed, right-open intervals with a specified step size.
  Example:

  ```sql
  FROM ("2000-11-14") TO ("2021-11-14") INTERVAL 2 YEAR
  ```

  *(Refer to the Doris Chinese documentation*

https://doris.apache.org/docs/table-design/data-partitioning/manual-partitioning


sqlglot/dialects/doris.py
```diff
@@ -575,3 +657,23 @@ class Generator(MySQL.Generator):
...
             "year",
         }
+
+        def partition_sql(self, expression):
+            # Doris expects: PARTITION <name> VALUES LESS THAN (<value>)
+            # Each expression is a PartitionRange
+            parts = []
+            for e in expression.expressions:
+                parts.append(self.sql(e))
+            return ", ".join(parts)
```
georgesittas: This doesn't look right, `Partition` is also constructed for statements like `INSERT OVERWRITE ...` ([docs](https://doris.apache.org/docs/2.0/sql-manual/sql-reference/Data-Manipulation-Statements/Manipulation/INSERT-OVERWRITE)). This override will omit the parentheses, which may result in invalid syntax.

I'd suggest only overriding the generation logic if you can guarantee that the `Partition` expression is used in a context where you need to use the unwrapped CSV syntax, e.g., do an instance check or something. Otherwise, fall back to the `super().partition_sql` implementation.

Also, you could condense what you have as follows:

```suggestion
return ", ".join(self.sql(e) for e in expression.expressions)
```

sqlglot/dialects/doris.py
```diff
@@ -575,3 +657,23 @@ class Generator(MySQL.Generator):
...
+
+        def partitionrange_sql(self, expression):
+            # Doris expects: PARTITION <name> VALUES LESS THAN (<value>) or VALUES [(<tuple>), ...)
+            name = self.sql(expression, "this")
+            value = expression.args.get("expression")
```
georgesittas:
```suggestion
            value = expression.expression
```

sqlglot/dialects/doris.py
```diff
@@ -575,3 +657,23 @@ class Generator(MySQL.Generator):
...
+            if isinstance(value, list):
+                # Output as VALUES [(item1), (item2))
+                value_sql = ", ".join(f"({self.sql(v)})" for v in value)
+                return f"PARTITION {name} VALUES [{value_sql})"
+            else:
+                value_sql = self.sql(value)
+                return f"PARTITION {name} VALUES LESS THAN ({value_sql})"
```
georgesittas:
```suggestion

            value_sql = self.sql(value)
            return f"PARTITION {name} VALUES LESS THAN ({value_sql})"
```

sqlglot/dialects/doris.py
```diff
@@ -575,3 +657,23 @@ class Generator(MySQL.Generator):
             "xor",
             "year",
         }
+
+        def partition_sql(self, expression):
```
georgesittas: Let's include type hints please, i.e., `expression: exp.Partition) -> str`.

sqlglot/dialects/doris.py
```diff
@@ -43,8 +44,89 @@ class Parser(MySQL.Parser):
...
+                        ],
+                    )
+                else:
+                    self.raise_error("Unsupported VALUES syntax in PARTITION definition")
+            else:
+                self.raise_error("Expecting VALUES in PARTITION definition")
+
+        def _parse_partition_by_opt_range(
+            self,
+        ) -> exp.PartitionedByProperty | exp.PartitionByRangeProperty:
```
georgesittas, line 106: Does StarRocks use the same syntax? We override this method for it as well, I wonder if we can consolidate the two.

sqlglot/dialects/doris.py
```diff
@@ -43,8 +44,89 @@ class Parser(MySQL.Parser):
             **MySQL.Parser.PROPERTY_PARSERS,
             "PROPERTIES": lambda self: self._parse_wrapped_properties(),
             "UNIQUE": lambda self: self._parse_composite_key_property(exp.UniqueKeyProperty),
+            "PARTITION BY": lambda self: self._parse_partition_by_opt_range(),
         }
 
+        def _parse_partitioning_granularity_dynamic(self) -> exp.PartitionByRangePropertyDynamic:
```
georgesittas, line 52: Ditto regarding consolidating Doris & StarRocks.

xinge-ji:
Hi @georgesittas, I have tried to update according to your comments. Looking forward to your feedback on the latest changes

PR #5409, opened 2025-07-15
feat: add `DateStrToTime`

betodealmeida:
I've been playing with the Python executor (building a backend for [shillelagh](https://github.com/betodealmeida/shillelagh)) and I found that there's no handling of `datetime.time` objects, so I added it in this PR.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5990,6 +5990,10 @@ class DateStrToDate(Func):
     pass
 
 
+class DateStrToTime(Func):
+    pass
+
+
 class DateToDateStr(Func):
     pass
 
@@ -8432,6 +8436,9 @@ def convert(value: t.Any, copy: bool = False) -> Expression:
     if isinstance(value, datetime.date):
         date_literal = Literal.string(value.strftime("%Y-%m-%d"))
         return DateStrToDate(this=date_literal)
+    if isinstance(value, datetime.time):
+        time_literal = Literal.string(value.isoformat())
+        return DateStrToTime(this=time_literal)
     if isinstance(value, tuple):
         if hasattr(value, "_fields"):
             return Struct(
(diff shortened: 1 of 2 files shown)


sqlglot/expressions.py
```diff
@@ -5990,6 +5990,10 @@ class DateStrToDate(Func):
     pass
 
 
+class DateStrToTime(Func):
```
tobymao: this already exists, TsOrDsToTime, so let's remove it

betodealmeida: OK, thanks! I left the tests, since there were no tests for `TsOrDsToTime`.

tests/test_expressions.py
```diff
@@ -1230,3 +1231,38 @@ def test_assert_is(self):
...
+        # Test noon
+        noon = datetime.time(12, 0, 0)
+        result = exp.convert(noon)
+
+        self.assertIsInstance(result, exp.TsOrDsToTime)
+        self.assertEqual(result.this.this, "12:00:00")
+
+    def test_datestrtotime_function(self):
+        # Test parsing TsOrDsToTime function
+        self.assertIsInstance(parse_one("TS_OR_DS_TO_TIME('14:30:45')"), exp.TsOrDsToTime)
```
tobymao: this test is extraneous, you have it on 685, let's delete

tests/test_expressions.py
```diff
@@ -1230,3 +1231,38 @@ def test_assert_is(self):
...
+
+        self.assertIsInstance(result, exp.TsOrDsToTime)
+        self.assertEqual(result.this.this, "00:00:00")
+
+        # Test noon
+        noon = datetime.time(12, 0, 0)
+        result = exp.convert(noon)
+
+        self.assertIsInstance(result, exp.TsOrDsToTime)
+        self.assertEqual(result.this.this, "12:00:00")
```
tobymao: these tests are super specific to the internals, instead of result.this.this, do

self.assertEqual(result.sql(), "...")

Issue #5404, opened 2025-07-15
Wrong Date_add function replaced at dremio dialect


ot3kcw6c:
If i parse a valid dremio dialect query it changes the date_add function to not working one.

sqlglot version: 27.0.0
Pyton version: 3.11.2 (64-bit)

DATE_ADD(CURRENT_DATE, 1) -> DATE_ADD(CURRENT_DATE, 1, 'DAY')

Dremio dokumentation:
https://docs.dremio.com/current/reference/sql/sql-functions/functions/DATE_ADD/

Error in dremio: DATE_ADD does not support operand types (DATE,INTEGER,VARCHAR)

```
from sqlglot import parse_one

sql_query='''
select colum_a from table_a where EXTRACT(MONTH FROM colum_a) = DATE_ADD(CURRENT_DATE, 1)
'''

sql_query_sqlgloted = parse_one(sql_query,dialect='dremio')

print(sql_query_sqlgloted.sql(dialect='dremio'))

# OUTPUT: SELECT colum_a FROM table_a WHERE EXTRACT(MONTH FROM colum_a) = DATE_ADD(CURRENT_DATE, 1, 'DAY')
# Change: DATE_ADD(CURRENT_DATE, 1) -> DATE_ADD(CURRENT_DATE, 1, 'DAY')
# DATE_ADD(CURRENT_DATE, 1, 'DAY') -> Dremio error: DATE_ADD does not support operand types (DATE,INTEGER,VARCHAR)
# https://docs.dremio.com/current/reference/sql/sql-functions/functions/DATE_ADD/
```

georgesittas:
Thanks for the report– this is not a high priority for the core sqlglot team, but we'll be happy to accept well-tested & documented PRs.

cc @mateuszpoleski

georgesittas:
Fixed by https://github.com/tobymao/sqlglot/pull/5411

PR #5411, opened 2025-07-15
feat(dremio): Add support for DATE_ADD and DATE_SUB

mateuszpoleski:
Currently Dremio dialect doesn't have proper support for `DATE_ADD` and `DATE_SUB` functions, as described in https://github.com/tobymao/sqlglot/issues/5404

In Dremio `DAY` unit is used by default in this functions, and shouldn't be specified. To correctly process units when parsed from different dialects, we can use `TIMESTAMPADD` function instead.

Link to relevant docs:
- https://docs.dremio.com/current/reference/sql/sql-functions/functions/DATE_ADD/
- https://docs.dremio.com/current/reference/sql/sql-functions/functions/DATE_SUB/
- https://docs.dremio.com/cloud/reference/sql/sql-functions/functions/TIMESTAMPADD/

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -1,6 +1,34 @@
+from __future__ import annotations
+
 from sqlglot import expressions as exp
 from sqlglot import parser, generator, tokens
-from sqlglot.dialects.dialect import Dialect, build_formatted_time
+from sqlglot.dialects.dialect import Dialect, build_formatted_time, unit_to_var
+import typing as t
+
+DATE_DELTA = t.Union[
+    exp.DateAdd,
+    exp.DateSub,
+]
+
+
+def _date_delta_sql(name: str) -> t.Callable[[Dremio.Generator, DATE_DELTA], str]:
+    def _delta_sql(self: Dremio.Generator, expression: DATE_DELTA) -> str:
+        unit = expression.text("unit")
+
+        if not unit or unit.upper() == "DAY":
+            return self.func(name, expression.this, expression.expression)
+
+        # to support units we need to use TIMESTAMPADD function
+        increment = expression.expression
+        if isinstance(expression, exp.DateSub):
+            if isinstance(increment, exp.Literal):
+                value = increment.to_py() if increment.is_number else int(increment.name)
+                increment = exp.Literal.number(value * -1)
+            else:
+                increment *= exp.Literal.number(-1)
+        return self.func("TIMESTAMPADD", unit_to_var(expression), increment, expression.this)
+
+    return _delta_sql
 
 
 class Dremio(Dialect):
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/dremio.py
```diff
@@ -1,6 +1,34 @@
...
+
+        # to support units we need to use TIMESTAMPADD function
+        increment = expression.expression
+        if isinstance(expression, exp.DateSub):
+            if isinstance(increment, exp.Literal):
+                value = increment.to_py() if increment.is_number else int(increment.name)
+                increment = exp.Literal.number(value * -1)
+            else:
+                increment *= exp.Literal.number(-1)
+        return self.func("TIMESTAMPADD", unit_to_var(expression), increment, expression.this)
```
georgesittas, line 29: Does it make sense to use `TIMESTAMPDIFF` ([docs](https://docs.dremio.com/cloud/reference/sql/sql-functions/functions/TIMESTAMPDIFF)) instead of negating the increment here?

sqlglot/dialects/dremio.py
```diff
@@ -1,6 +1,34 @@
...
+            return self.func(name, expression.this, expression.expression)
+
+        # to support units we need to use TIMESTAMPADD function
+        increment = expression.expression
+        if isinstance(expression, exp.DateSub):
+            if isinstance(increment, exp.Literal):
+                value = increment.to_py() if increment.is_number else int(increment.name)
+                increment = exp.Literal.number(value * -1)
+            else:
+                increment *= exp.Literal.number(-1)
```
georgesittas, line 28: Should we consolidate these two and just do `increment *= -1` or something? I think the `__mul__` dunder will take care of constructing the literal for `-1`.

sqlglot/dialects/dremio.py
mateuszpoleski: Unfortunately `TIMESTAMPDIFF` works somewhat different than `DATE_DIFF` for instance. It returns an integer difference instead of shifted timestamp, and takes different arguments. Probably it would be even harder to make it work

sqlglot/dialects/dremio.py
mateuszpoleski: I tried writing it as 
```python
  if isinstance(increment, exp.Literal):
      value = increment.to_py() if increment.is_number else int(increment.name)
      increment = exp.Literal.number(value)
  increment *= -1
```
but with this we would parse `DATE_SUB(col, 2, 'HOUR')` into `SELECT TIMESTAMPADD(HOUR, 2 * -1, col)` instead of `SELECT TIMESTAMPADD(HOUR, -2, col)`, not sure if we like that?

mateuszpoleski:
Thanks for the review, these are good ideas, but I didn't manage to make it work this time

sqlglot/dialects/dremio.py
georgesittas: Got it, no worries.

sqlglot/dialects/dremio.py
georgesittas: It's fine as is, don't no worries.

Issue #5406, opened 2025-07-15
Parsing `semantic_view` Snowflake queries doesn't work


jkillian:
Snowflake recently introduced the concept [of Snowflake-native "semantic views"](https://docs.snowflake.com/en/user-guide/views-semantic/overview). These can be [queried directly in Snowflake SQL](https://docs.snowflake.com/en/user-guide/views-semantic/querying).

As far as I could tell, sqlglot doesn't yet understand this syntax, so wanted to file a ticket to add support! This syntax wouldn't be translatable to other dialects most likely.

**Fully reproducible code snippet**
```
import sqlglot
sqlglot.parse_one("""
  select * from semantic_view(
     MY_DB.MY_SCHEMA.SEMANTIC_VIEW_NAME     
     dimensions MY_TABLE.MY_DIM
     metrics MY_TABLE.MY_METRIC
  )
""",
  dialect="snowflake",
)
```

**Result:**

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "~/projects/sqlglot/sqlglot/__init__.py", line 139, in parse_one
    result = dialect.parse(sql, **opts)
  File "~/projects/sqlglot/sqlglot/dialects/dialect.py", line 1058, in parse
    return self.parser(**opts).parse(self.tokenize(sql), sql)
  File "~/projects/sqlglot/sqlglot/parser.py", line 1592, in parse
    return self._parse(
  File "~/projects/sqlglot/sqlglot/parser.py", line 1661, in _parse
    expressions.append(parse_method(self))
  File "~/projects/sqlglot/sqlglot/parser.py", line 1902, in _parse_statement
    expression = self._parse_set_operations(expression) if expression else self._parse_select()
  File "~/projects/sqlglot/sqlglot/parser.py", line 3190, in _parse_select
    query = self._parse_select_query(
  File "~/projects/sqlglot/sqlglot/parser.py", line 3287, in _parse_select_query
    from_ = self._parse_from()
  File "~/projects/sqlglot/sqlglot/parser.py", line 3578, in _parse_from
    this=self._parse_table(joins=joins, consume_pipe=consume_pipe),
  File "~/projects/sqlglot/sqlglot/dialects/snowflake.py", line 806, in _parse_table
    table = super()._parse_table(
  File "~/projects/sqlglot/sqlglot/parser.py", line 4044, in _parse_table
    self._parse_table_parts(schema=schema, is_db_reference=is_db_reference)
[...]

PR #5414, opened 2025-07-16
feat(snowflake): Support SEMANTIC_VIEW

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5406

Docs
--------
[Syntax](https://docs.snowflake.com/en/sql-reference/constructs/semantic_view#syntax) | [Examples](https://docs.snowflake.com/en/user-guide/views-semantic/querying#basic-examples)

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -985,11 +985,11 @@ class Snowflake(Dialect):
         def _parse_semantic_view(self) -> exp.SemanticView:
             kwargs: t.Dict[str, t.Any] = {"this": self._parse_table_parts()}
 
-            while not self._match(TokenType.R_PAREN, advance=False):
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -981,6 +982,19 @@ def _parse_credentials_property(self) -> exp.CredentialsProperty:
                 expressions=self._parse_wrapped_options(),
             )
 
+        def _parse_semantic_view(self) -> exp.SemanticView:
+            kwargs: t.Dict[str, t.Any] = {"this": self._parse_table_parts()}
+
+            while not self._match(TokenType.R_PAREN, advance=False):
```
georgesittas:
```suggestion
            while self._curr and not self._match(TokenType.R_PAREN, advance=False):
```

georgesittas (approved):

PR #5415, opened 2025-07-16
feat(postgres): query placeholders

### LINKED ISSUE #5412: Support for Postgres/Psycopg-style %(params)s
Psycopg and also some other database drivers using parameters like this:

```sql
SELECT * FROM some_table WHERE id = %(customer_id)s
-- or also
SELECT * FROM some_table WHERE id = %s
```

However, this is not parse-able by sqlglot currently:

```python
import sqlglot as sg

print(
    repr(sg.parse_one("SELECT * FROM some_table WHERE id = $customer_id", read="postgres"))
)  # works
print(
    repr(sg.parse_one("SELECT * FROM some_table WHERE id = %(customer_id)s", read="postgres"))
)  # fails, it's somehow expecting modulo here
```` 

Docs: https://www.psycopg.org/psycopg3/docs/basic/params.html

### PR DESCRIPTION
**Pull request by geooo109**
Fixes #5412

Support the following non-native placeholders for parameter sub:
```
SELECT * FROM foo WHERE id = %s
SELECT * FROM foo WHERE id = %(id_param)s")
```

**DOCS**
[Postgres placeholders](https://www.psycopg.org/psycopg3/docs/basic/params.html)

### DIFF
--- a/sqlglot/dialects/postgres.py
+++ b/sqlglot/dialects/postgres.py
@@ -460,13 +460,12 @@ class Postgres(Dialect):
         }
 
         def _parse_query_parameter(self) -> t.Optional[exp.Expression]:
-            self._match(TokenType.MOD)
             this = (
                 self._parse_wrapped(self._parse_id_var)
                 if self._match(TokenType.L_PAREN, advance=False)
                 else None
             )
-            self._match(TokenType.VAR)
+            self._match_text_seq("S")
             return self.expression(exp.Placeholder, this=this)
 
         def _parse_operator(self, this: t.Optional[exp.Expression]) -> t.Optional[exp.Expression]:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: sqlglot/dialects/postgres.py [hunk 1]
```diff
@@ -454,6 +459,16 @@ class Parser(parser.Parser):
             )([this, path]),
         }
 
+        def _parse_query_parameter(self) -> t.Optional[exp.Expression]:
+            self._match(TokenType.MOD)
```
georgesittas: Why do we check for `MOD` here, isn't that consumed in `_parse_placeholder`?

sqlglot/dialects/postgres.py
```diff
@@ -454,6 +459,16 @@ class Parser(parser.Parser):
...
+            this = (
+                self._parse_wrapped(self._parse_id_var)
+                if self._match(TokenType.L_PAREN, advance=False)
+                else None
+            )
+            self._match(TokenType.VAR)
```
georgesittas: Should this be `self._match_text_seq("S")` to capture the `%s`?

geooo109: yep, my bad, fixed it

georgesittas (approved):

PR #5418, opened 2025-07-16
fix(risingwave): Fix RisingWave dialect SQL for MAP datatype declaration

MisterWheatley:
Before this PR the SQL generated for a MAP datatype declaration expression in the RisingWave dialect was `MAP<x, y>`, where `x` and `y` are other datatypes e.g. `MAP<VARCHAR, INT>`. 

The correct way according to the official docs here https://docs.risingwave.com/sql/data-types/map-type is instead `MAP(x, y)` e.g. `MAP(VARCHAR, INT)`


georgesittas (approved):
Thanks!

PR #5419, opened 2025-07-16
feat(exasol): add support for NVL function in Exasol dialect

nnamdi16:
This involves adding [NVL](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/nvl.htm) built -in function to exasol dialect in sqlglot.

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -69,6 +69,7 @@ class Exasol(Dialect):
                 timestamp=seq_get(args, 0),
                 options=seq_get(args, 3),
             ),
+            "NVL": lambda args: exp.Coalesce(this=seq_get(args, 0), expressions=seq_get(args, 1)),
         }
 
     class Generator(generator.Generator):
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/exasol.py
```diff
@@ -124,6 +125,8 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             exp.BitwiseXor: rename_func("BIT_XOR"),
             # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/every.htm
             exp.All: rename_func("EVERY"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/nvl.htm
+            exp.Coalesce: unsupported_args("is_nvl", "is_null")(rename_func("NVL")),
```
georgesittas: This doesn't look right– Exasol supports both `NVL` ([1](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/nvl.htm)) and `COALESCE` ([2](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/coalesce.htm#microcontent1)). This entry will force `Coalesce` expressions to be generated as `NVL`, but the latter only supports two args, whereas the former supports arbitrarily many args. Moreover, I'm not sure why `is_nvl` and `is_null` are considered unsupported.

Aren't `NVL` and `COALESCE` equivalent in Exasol? We could always just generate `COALESCE`, if so. Otherwise, you can set `is_nvl`, like we do for Oracle.

nnamdi16: Alright, I would look into it

georgesittas:
@nnamdi16 are the semantics of `NVL` different from `COALESCE`?

nnamdi16:
> @nnamdi16 are the semantics of `NVL` different from `COALESCE`?

They are the same. Its just that `NVL` accepts two arguments while `COALESCE` accepts more than two.

@georgesittas

georgesittas:
There is no need for this PR, then, as SQLGlot already produces `COALESCE` which matches the semantics of `NVL`:

```python
>>> import sqlglot
>>> sqlglot.transpile("NVL(a, b)", "exasol")
['COALESCE(a, b)']
```

nnamdi16:
> There is no need for this PR, then, as SQLGlot already produces `COALESCE` which matches the semantics of `NVL`:
> 
> ```python
> >>> import sqlglot
> >>> sqlglot.transpile("NVL(a, b)", "exasol")
> ['COALESCE(a, b)']
> ```

Alright I would close it

PR #5423, opened 2025-07-17
fix(duckdb)!: make bracket parsing aware of duckdb MAP func

### LINKED ISSUE #5417: Incorrect parsing of MAP literals in DuckDB
sqlgot version == 27.0.0
read = duckdb
write = duckdb

The current parsing of the following MAP literals in DuckDB do not allow non-string keys, which is permissable under the syntax defined in the official documentation (https://duckdb.org/docs/stable/sql/data_types/map.html)

Example 1: integer keys
```python
from sqlglot import parse_one

sql = "SELECT MAP { 1: 'a', 2: 'b'} AS map_literal"
ast = parse_one(sql, dialect='duckdb')
out = ast.sql(ast, dialect='duckdb')
````
The expected result is `SELECT MAP { 1: 'a', 2: 'b'} AS map_literal` with the actual result being `SELECT MAP {'1': 'a', '2': 'b'} AS map_literal`

Example 2: Nested type keys
```python
from sqlglot import parse_one
sql = "SELECT MAP {[1,2]: 'a', [3,4]: 'b'} AS map_literal"
ast = parse_one(sql, dialect='duckdb') 
out = ast.sql(dialect='duckdb')
```
The expected result is `SELECT MAP {[1,2]: 'a', [3,4]: 'b'} AS map_literal` with the actual result being `SELECT MAP {'': 'a', '': 'b'} AS map_literal`

The reason for filing this bug report is to hopefully get inspired to implement similar handling for RisingWave dialect MAP literals as described in their docs (https://docs.risingwave.com/sql/data-types/map-type#define-a-map), as I know RisingWave is community supported dialect.

geooo109:
Fixes #5417

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -1529,6 +1529,11 @@ class Parser(metaclass=_Parser):
     # Whether TIMESTAMP <literal> can produce a zone-aware timestamp
     ZONE_AWARE_TIMESTAMP_CONSTRUCTOR = False
 
+    # Whether map literals support arbitrary expressions as keys.
+    # When True, allows complex keys like arrays or literals: {[1, 2]: 3}, {1: 2} (e.g. DuckDB).
+    # When False, keys are typically restricted to identifiers.
+    MAP_KEYS_ARE_ARBITRARY_EXPRESSIONS = False
+
     __slots__ = (
         "error_level",
         "error_message_context",
@@ -6317,8 +6322,9 @@ class Parser(metaclass=_Parser):
         if not self._match_set((TokenType.L_BRACKET, TokenType.L_BRACE)):
             return this
 
-        if self.dialect.MAP_KEYS_ARE_ARBITRARY_EXPRESSIONS:
-            parse_map = self._tokens[self._index - 2].text.upper() == "MAP"
+        if self.MAP_KEYS_ARE_ARBITRARY_EXPRESSIONS:
+            map_token = seq_get(self._tokens, self._index - 2)
+            parse_map = map_token is not None and map_token.text.upper() == "MAP"
         else:
             parse_map = False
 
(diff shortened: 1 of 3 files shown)


sqlglot/parser.py
```diff
@@ -5833,7 +5833,9 @@ def _parse_function_call(
     def _to_prop_eq(self, expression: exp.Expression, index: int) -> exp.Expression:
         return expression
 
-    def _kv_to_prop_eq(self, expressions: t.List[exp.Expression]) -> t.List[exp.Expression]:
+    def _kv_to_prop_eq(
+        self, expressions: t.List[exp.Expression], parse_map: bool = False
+    ) -> t.List[exp.Expression]:
```
georgesittas, line 5843: @geooo109 instead of adding this flag, is it possible to override `_kv_to_prop_eq` for duckdb in order to check whether there's a `MAP` token before `{`? That way you don't have to pollute the base methods to propagate the argument.

geooo109: Yep, wanted to avoid this flag in first place :D . I will check it

sqlglot/parser.py
```diff
@@ -6313,6 +6317,11 @@ def _parse_bracket(self, this: t.Optional[exp.Expression] = None) -> t.Optional[
         if not self._match_set((TokenType.L_BRACKET, TokenType.L_BRACE)):
             return this
 
+        if self.dialect.MAP_KEYS_ARE_ARBITRARY_EXPRESSIONS:
+            parse_map = self._tokens[self._index - 2].text.upper() == "MAP"
```
georgesittas: Let's add a safety check here, or use `seq_get`, instead of blindly accessing `index - 2`.

georgesittas (commented):
Looks good

sqlglot/parser.py
geooo109: you mean something like this: 
```
map_token = seq_get(self._tokens, self._index - 2)
parse_map = map_token is not None and map_token.text.upper() == "MAP"
``` ?

PR #5429, opened 2025-07-18
feat(exasol): Add rank function to sqlglot expressions and parser

nnamdi16:
This involves adding the [RANK](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/rank.htm) built -in function to exasol dialect in sqlglot

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -159,6 +159,7 @@ class Exasol(Dialect):
                 "'UTC'",
                 e.args.get("zone"),
             ),
+            exp.Rank: lambda self, e: self.func("RANK"),
         }
 
         def converttimezone_sql(self, expression: exp.ConvertTimezone) -> str:
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -7073,6 +7073,10 @@ class CovarPop(Binary, AggFunc):
     pass
 
 
+class Rank(Func):
+    arg_types = {"this": False}
+
+
 class Week(Func):
     arg_types = {"this": True, "mode": False}
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -274,6 +274,7 @@ class Parser(metaclass=_Parser):
         TokenType.CURRENT_TIME: exp.CurrentTime,
         TokenType.CURRENT_TIMESTAMP: exp.CurrentTimestamp,
         TokenType.CURRENT_USER: exp.CurrentUser,
+        TokenType.RANK: exp.Rank,
     }
 
     STRUCT_TYPE_TOKENS = {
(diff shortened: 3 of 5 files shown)


georgesittas:
@nnamdi16 what is this PR aiming to solve? `RANK` already roundtrips without issues:

```python
>>> import sqlglot
>>> sqlglot.transpile("""
... SELECT
...     id, department, current_salary,
...     RANK() OVER (PARTITION BY department ORDER BY current_salary) RANK
... FROM employee_table ORDER BY department, current_salary;
... """, "exasol")
['SELECT id, department, current_salary, RANK() OVER (PARTITION BY department ORDER BY current_salary) AS RANK FROM employee_table ORDER BY department, current_salary']
```

I don't see any transpilation tests, but regardless, I think it's quite universal so I'd be surprised if it wasn't supported by a dialect.

nnamdi16:
> @nnamdi16 what is this PR aiming to solve? `RANK` already roundtrips without issues:
> 
> ```python
> >>> import sqlglot
> >>> sqlglot.transpile("""
> ... SELECT
> ...     id, department, current_salary,
> ...     RANK() OVER (PARTITION BY department ORDER BY current_salary) RANK
> ... FROM employee_table ORDER BY department, current_salary;
> ... """, "exasol")
> ['SELECT id, department, current_salary, RANK() OVER (PARTITION BY department ORDER BY current_salary) AS RANK FROM employee_table ORDER BY department, current_salary']
> ```
> 
> I don't see any transpilation tests, but regardless, I think it's quite universal so I'd be surprised if it wasn't supported by a dialect.

So the issue is that exasol doesn't support passing of arguments to a RANK function but in databricks you can pass an argument to RANK function. Since there is no  expression for it, it makes it difficult to override in the generator.
@georgesittas

georgesittas:
Databricks [doesn't document](https://docs.databricks.com/aws/en/sql/language-manual/functions/rank) the fact that `RANK` can accept an argument, but I verified that it does accept it. Based on the lack of documentation, I think it most likely drops it. I'm not sure if this warrants creating a new expression type.

nnamdi16:
> Databricks [doesn't document](https://docs.databricks.com/aws/en/sql/language-manual/functions/rank) the fact that `RANK` can accept an argument, but I verified that it does accept it. Based on the lack of documentation, I think it most likely drops it. I'm not sure if this warrants creating a new expression type.

What do you think we can do on exasol end to ensure that the values passed to RANK is removed without the RANK  expression?
@georgesittas

georgesittas:
Unless there's a dialect that clearly documents supporting argument(s) for `RANK`, I don't think we'll add an expression in sqlglot to have first-class support for it either.

I'd suggest cleaning your inputs or intercepting `Anonymous` nodes that correspond to `"RANK"` (i.e., check their name) and popping the arguments if they are not functionally important.

Issue #5413, opened 2025-07-16
TSQL DDL to default varchar to max


mattiasthalen:
Currently when creating a table with `varchar` without precision, this results in `varchar(1)` being defined.

IMO, it would be better if it changed this behavior so that it converts that to `varchar(max)`.

georgesittas:
Just clarifying: if the input T-SQL code has `VARCHAR`, it should be respected. We should only change it to `VARCHAR(MAX)` if the input came from a dialect that doesn't treat `VARCHAR` as `VARCHAR(1)`.

mattiasthalen:
@georgesittas ok, can you expand on the rationale? ☺️

is it that if the user does a tsql ddl with varchar, that should be respected, we assume they know what they are doing? but if it's postgressql transpiled to tsql, it should be varchar(max)

georgesittas:
Correct, it's exactly as you said. We should respect the semantics of the input statement w.r.t. the `read` dialect.

georgesittas:
This is relatively low priority for us right now, so I'll go ahead and close the ticket as not planned. Feel free to pick this up when you're back @mattiasthalen.

PR #5434, opened 2025-07-21
feat(tsql)!: add default precision to VARCHAR create expressions

mattiasthalen:
This pull request introduces a new feature to the T-SQL dialect in `sqlglot` to automatically add a default precision (`MAX`) to `VARCHAR` data types without an explicit precision. It also includes corresponding updates to preprocessing and test cases.

### Feature Addition: Default Precision for `VARCHAR`

* **New Functionality**: Added a helper function `add_default_precision_to_varchar` in `sqlglot/dialects/tsql.py` to add a default precision (`MAX`) to `VARCHAR` expressions that lack an explicit precision. This is implemented by checking for `VARCHAR` data types and appending a `DataTypeParam` with the value `MAX` if none exists.

* **Integration with T-SQL Dialect**: Updated the T-SQL `Generator` class to include the new `add_default_precision_to_varchar` function in the preprocessing step for `CREATE` expressions. This ensures the transformation is applied during transpilation.

### Test Coverage

* **New Test Cases**: Added tests in `tests/dialects/test_tsql.py` to validate the new behavior:
  - Ensures `VARCHAR` without precision is converted to `VARCHAR(MAX)`.
  - Confirms that existing precision in `VARCHAR` is preserved.

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -897,6 +897,8 @@ class TSQL(Dialect):
                                 and not column_type.expressions
                             ):
                                 # Add default precision of 1 to VARCHAR without precision
+                                # When n isn't specified in a data definition or variable declaration statement, the default length is 1.
+                                # https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver17#remarks
                                 column_type.set("expressions", [exp.Literal.number("1")])
 
             return create
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -890,7 +890,7 @@ class TSQL(Dialect):
                 if create.kind == "TABLE" and create.this and hasattr(create.this, "expressions"):
                     for column in create.this.expressions:
                         if isinstance(column, exp.ColumnDef):
-                            column_type = column.args.get("kind")
+                            column_type = column.kind
                             if (
                                 isinstance(column_type, exp.DataType)
                                 and column_type.this == exp.DataType.Type.VARCHAR
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -393,7 +393,7 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
 
 
 def _add_default_precision_to_varchar(expression: exp.Expression) -> exp.Expression:
-    """Transform function to add VARCHAR(MAX) for cross-dialect conversion."""
+    """Transform function to add VARCHAR(MAX) or CHAR(MAX) for cross-dialect conversion."""
     if isinstance(expression, exp.Create) and expression.kind == "TABLE":
         if expression.this and hasattr(expression.this, "expressions"):
             for column in expression.this.expressions:
@@ -401,10 +401,10 @@ def _add_default_precision_to_varchar(expression: exp.Expression) -> exp.Express
                     column_type = column.args.get("kind")
                     if (
                         isinstance(column_type, exp.DataType)
-                        and column_type.this == exp.DataType.Type.VARCHAR
+                        and column_type.this in (exp.DataType.Type.VARCHAR, exp.DataType.Type.CHAR)
                         and not column_type.expressions
                     ):
-                        # For cross-dialect conversion, VARCHAR without precision becomes VARCHAR(MAX)
+                        # For cross-dialect conversion, VARCHAR/CHAR without precision becomes VARCHAR(MAX)/CHAR(MAX)
                         column_type.set("expressions", [exp.var("MAX")])
     return expression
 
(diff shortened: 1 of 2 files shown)


mattiasthalen:
@georgesittas, I couldn't figure out how to detect what dialect is being read, so that it only happens for non-tsql inputs.

georgesittas:
Hello again, @mattiasthalen! :)

This doesn't look quite right– please refer to https://github.com/tobymao/sqlglot/issues/5413#issuecomment-3077680566. Perhaps we can convert `CREATE TABLE t (c VARCHAR)` to `CREATE TABLE t (c VARCHAR(1))` at parse time for T-SQL, if the two are guaranteed to be equivalent. Then, at generation time you're guaranteed to have a precision if the input was parsed with T-SQL, and so converting `VARCHAR` to `VARCHAR(MAX)` should be safe.

@VaggelisD feel free to chime in.

mattiasthalen:
@georgesittas hi! ☺️

My thought was to add max for all except tsql, but I wasn't able to detect the source dialect in the transform.

So this was my attempt at solving your comment 🙈

I think I understand what you mean by having TSQL output varchar(1), but I fail to see how it would work.

georgesittas:
> I think I understand what you mean by having TSQL output varchar(1), but I fail to see how it would work.

You could override `_parse_create` in T-SQL's `Parser` and transform the parsed `Create` node such that `VARCHAR`s in column definitions (i.e., the `c VARCHAR` part) become `VARCHAR(1)`. Something like that.

sqlglot/dialects/tsql.py
```diff
@@ -869,6 +869,19 @@ def _parse_create(self) -> exp.Create | exp.Command:
...
+                    for column in create.this.expressions:
+                        if isinstance(column, exp.ColumnDef):
+                            column_type = column.args.get("kind")
+                            if (
+                                isinstance(column_type, exp.DataType)
+                                and column_type.this == exp.DataType.Type.VARCHAR
+                                and not column_type.expressions
+                            ):
+                                # Add default precision of 1 to VARCHAR without precision
+                                column_type.set("expressions", [exp.Literal.number("1")])
```
mattiasthalen: @georgesittas, something like this?

sqlglot/dialects/tsql.py
```diff
@@ -869,6 +869,19 @@ def _parse_create(self) -> exp.Create | exp.Command:
 
                     create.args["properties"].append("expressions", exp.TemporaryProperty())
 
+                # Transform VARCHAR without precision to VARCHAR(1)
+                if create.kind == "TABLE" and create.this and hasattr(create.this, "expressions"):
```
georgesittas:
```suggestion
                if create.kind == "TABLE" and isinstance(create.this, exp.Schema):
```

sqlglot/dialects/tsql.py
```diff
@@ -869,6 +869,19 @@ def _parse_create(self) -> exp.Create | exp.Command:
...
+                    for column in create.this.expressions:
+                        if isinstance(column, exp.ColumnDef):
+                            column_type = column.args.get("kind")
```
georgesittas:
```suggestion
                            column_type = column.kind
```

sqlglot/dialects/tsql.py
georgesittas: Yes– can we please ensure these are the correct semantics for DDL statements in T-SQL? Some documentation around this, followed by testing etc to prove this is how the engine behaves. Additionally, could this be controlled by a client-side setting or something? Perhaps the engine behaves differently under condition?

mattiasthalen: https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver17

>When n isn't specified in a data definition or variable declaration statement, the default length is 1. If n isn't specified when using the CAST and CONVERT functions, the default length is 30.

This applies to `varchar` and `char`. Should I add `char` to this, and should I look into creating similar logic for `CAST`?

sqlglot/dialects/tsql.py
```diff
@@ -392,6 +392,23 @@ def _timestrtotime_sql(self: TSQL.Generator, expression: exp.TimeStrToTime):
     return sql
 
 
+def _add_default_precision_to_varchar(expression: exp.Expression) -> exp.Expression:
+    """Transform function to add VARCHAR(MAX) or CHAR(MAX) for cross-dialect conversion."""
+    if isinstance(expression, exp.Create) and expression.kind == "TABLE":
+        if expression.this and hasattr(expression.this, "expressions"):
+            for column in expression.this.expressions:
```
georgesittas: Nit: let's mimic the logic of `_parse_create`, i.e., check for `exp.Schema`, etc.

sqlglot/dialects/tsql.py
georgesittas: Thank you for looking this up!

> Should I add `char` to this, and should I look into creating similar logic for `CAST`?

Sounds good to me 👍

mattiasthalen: > Thank you for looking this up!
> 
> > Should I add `char` to this, and should I look into creating similar logic for `CAST`?
> 
> Sounds good to me 👍

I'll do the cast as a seperate PR

PR #5435, opened 2025-07-21
feat: Add support for BETWEEN flags

mateuszpoleski:
Some SQL dialects support optional `ASYMMETRIC` and `SYMMETRIC` flags for `BETWEEN` expression (some context here https://dev.mysql.com/worklog/task/?id=5084)

Currently query like `SELECT expr BETWEEN SYMMETRIC low AND high` translates into `SELECT expr BETWEEN SYMMETRIC AND low AND high`

Flags behaviour for `SELECT expr BETWEEN low AND high`: 
- Adding the word `ASYMMETRIC` is harmless but achieves nothing, checks `expr >= low AND expr <= high`
- Adding the word `SYMMETRIC` checks for `expr >= low AND expr <= high OR expr >= high AND expr <= low `

```sql
SELECT 7 BETWEEN 10 AND 5;              -- FALSE
SELECT 7 BETWEEN ASYMMETRIC 10 AND 5;   -- FALSE
SELECT 7 BETWEEN SYMMETRIC 10 AND 5;    -- TRUE
```

This implementation tries to use these flags if they're supported by the dialect; otherwise, it falls back to the expanded form. From the dialects I tested, only `Dremio` and `PostgreSQL` supported them, but it's likely that others do too. Setting `SUPPORTS_BETWEEN_FLAGS = False` should be a safe default, even if a dialect does support the flags.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -935,11 +935,14 @@ class Expression(metaclass=_Expression):
             ),
         )
 
-    def between(self, low: t.Any, high: t.Any, copy: bool = True, **opts) -> Between:
+    def between(
+        self, low: t.Any, high: t.Any, copy: bool = True, kind: str | None = None, **opts
+    ) -> Between:
         return Between(
             this=maybe_copy(self, copy),
             low=convert(low, copy=copy, **opts),
             high=convert(high, copy=copy, **opts),
+            kind=kind,
         )
 
     def is_(self, other: ExpOrStr) -> Is:
@@ -5224,7 +5227,7 @@ class FormatPhrase(Expression):
 
 
 class Between(Predicate):
-    arg_types = {"this": True, "low": True, "high": True}
+    arg_types = {"this": True, "low": True, "high": True, "kind": False}
 
 
 class Bracket(Condition):
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -236,6 +236,7 @@ class TokenType(AutoName):
     ARRAY = auto()
     ASC = auto()
     ASOF = auto()
+    ASYMMETRIC = auto()
     ATTACH = auto()
     AUTO_INCREMENT = auto()
     BEGIN = auto()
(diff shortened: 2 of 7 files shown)


sqlglot/expressions.py
```diff
@@ -5224,7 +5227,7 @@ class FormatPhrase(Expression):
 
 
 class Between(Predicate):
-    arg_types = {"this": True, "low": True, "high": True}
+    arg_types = {"this": True, "low": True, "high": True, "kind": False}
```
georgesittas: What about renaming this to `symmetric` and making it an `Optional[bool]`? `True` means we either generate `SYMMETRIC`, if supported, or transpile it. Otherwise, the behavior is the same as today.

sqlglot/tokens.py
```diff
@@ -236,6 +236,7 @@ class TokenType(AutoName):
     ARRAY = auto()
     ASC = auto()
     ASOF = auto()
+    ASYMMETRIC = auto()
```
georgesittas: These need to be added to `ID_VAR_TOKENS`. Or, perhaps better, don't introduce new tokens at all and just use `_match_text_seq` to consume those keywords.

georgesittas (commented):
Thanks for the PR, @mateuszpoleski! Didn't know about this syntax. Left a couple of suggestions.

sqlglot/expressions.py
mateuszpoleski: I changed it to `Optional[bool]`, but roundtrip still preserves `ASYMMETRIC` if supported -- not sure if thats what you meant?

sqlglot/tokens.py
mateuszpoleski: Updated to `_match_text_seq`. Out of curiosity, when would we need to introduce new tokens instead of using plain strings?

VaggelisD: @mateuszpoleski We try to avoid adding new tokens for several reasons, mostly because by default they are considered "reserved" by SQLGLot, making it impossible to parse queries like `SELECT symmetric ...`. This of course introduces regressions and is hard to spot during code reviews.

We tend to add tokens when:
- Our abstractions demand it, e.g the token would be associated with a data structure that handles tokens
- It's a reserved or common keyword, e.g `SELECT`, `GROUP` etc

georgesittas (approved):
Nice work, thanks for refactoring

sqlglot/expressions.py
georgesittas: Sounds good

PR #5442, opened 2025-07-21
feat: add parsing/generating for BigQuery `DECLARE`

plaflamme:
See documentation [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#declare)

Note that because BigQuery can infer the variable type from the default value, I had to make `kind` optional in the `Declare` token.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -32,7 +32,7 @@ from sqlglot.dialects.dialect import (
     groupconcat_sql,
     space_sql,
 )
-from sqlglot.helper import seq_get, split_num_words
+from sqlglot.helper import csv, seq_get, split_num_words
 from sqlglot.tokens import TokenType
 from sqlglot.generator import unsupported_args
 
@@ -530,7 +530,7 @@ class BigQuery(Dialect):
             "BYTES": TokenType.BINARY,
             "CURRENT_DATETIME": TokenType.CURRENT_DATETIME,
             "DATETIME": TokenType.TIMESTAMP,
-            "DECLARE": TokenType.COMMAND,
+            "DECLARE": TokenType.DECLARE,
             "ELSEIF": TokenType.COMMAND,
             "EXCEPTION": TokenType.COMMAND,
             "EXPORT": TokenType.EXPORT,
@@ -671,6 +671,7 @@ class BigQuery(Dialect):
             TokenType.END: lambda self: self._parse_as_command(self._prev),
             TokenType.FOR: lambda self: self._parse_for_in(),
             TokenType.EXPORT: lambda self: self._parse_export_data(),
+            TokenType.DECLARE: lambda self: self._parse_declare(),
         }
 
         BRACKET_OFFSETS = {
@@ -936,6 +937,33 @@ class BigQuery(Dialect):
                 options=self._parse_properties(),
                 this=self._match_text_seq("AS") and self._parse_select(),
             )
+
+        def _parse_declare(self) -> exp.Declare | exp.Command:
+            index = self._index
+            expressions = self._try_parse(self._parse_declareitem)
+
+            if not expressions or self._curr:
+                self._retreat(index)
+                return self._parse_as_command(self._prev)
+
+            return self.expression(exp.Declare, expressions=[expressions])
+
+        def _parse_declareitem(self) -> t.Optional[exp.DeclareItem]:
+            vars = self._parse_csv(self._parse_id_var)
+            if not vars:
+                return None
+
+            data_type = self._parse_types()
+            default = None
+            if self._match(TokenType.DEFAULT):
+                default = self._parse_bitwise()
+
+            if not data_type and not default:
+                self.raise_error(
+                    "BigQuery requires a data type or default value for DECLARE statements",
+                    token=self._curr,
+                )
+            return self.expression(exp.DeclareItem, this=vars, kind=data_type, default=default)
 
     class Generator(generator.Generator):
         INTERVAL_ALLOWS_PLURAL_FORM = False
@@ -1361,3 +1389,15 @@ class BigQuery(Dialect):
                     return f"{self.sql(expression, 'to')}{self.sql(this)}"
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
+
+        def declare_sql(self, expression: exp.Declare) -> str:
+            return f"DECLARE {self.expressions(expression, flat=True)}"
+
+        def declareitem_sql(self, expression: exp.DeclareItem) -> str:
+            variables = csv(*[self.sql(v) for v in expression.this])
+            default = self.sql(expression, "default")
+            default = f"DEFAULT {default}" if default else ""
+            kind = self.sql(expression, "kind")
+            space = " " if kind and default else ""
+
+            return f"{variables} {kind}{space}{default}"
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -1361,3 +1389,15 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
                     return f"{self.sql(expression, 'to')}{self.sql(this)}"
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
+
+        def declare_sql(self, expression: exp.Declare) -> str:
+            return f"DECLARE {self.expressions(expression, flat=True)}"
+
+        def declareitem_sql(self, expression: exp.DeclareItem) -> str:
+            variables = csv(*[self.sql(v) for v in expression.this])
```
VaggelisD: This should leverage `self.expressions`

sqlglot/dialects/bigquery.py
```diff
@@ -1361,3 +1389,15 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
...
+            default = self.sql(expression, "default")
+            default = f"DEFAULT {default}" if default else ""
+            kind = self.sql(expression, "kind")
+            space = " " if kind and default else ""
+
+            return f"{variables} {kind}{space}{default}"
```
VaggelisD: We tend to structure such statements as in the following snippet, so that each component has it's own spacing without having to introduce extra variables/logic:

```Python3
default = self.sql(expression, "default")
default = f" DEFAULT {default}" if default else ""
kind = f" {kind}" if kind else "

return f"{variables}{kind}{default}"
```

tests/dialects/test_bigquery.py
```diff
@@ -2695,3 +2689,29 @@ def test_json_array(self):
...
+            "DECLARE START_DATE DATE DEFAULT CURRENT_DATE() - 1",
+            "DECLARE START_DATE DATE DEFAULT CURRENT_DATE - 1",
+        )
+        self.validate_identity(
+            "DECLARE TS TIMESTAMP DEFAULT CURRENT_TIMESTAMP() - INTERVAL '1' HOUR",
+            "DECLARE TS TIMESTAMP DEFAULT CURRENT_TIMESTAMP() - INTERVAL '1' HOUR",
+        )
+        self.validate_identity(
+            "DECLARE START_DATE DEFAULT CURRENT_DATE", "DECLARE START_DATE DEFAULT CURRENT_DATE"
+        )
```
VaggelisD: If the result sql remains the same (i.e the "roundtrip" is preserved) we should omit the 2nd arg in `validate_identity`, this makes it easier to read & reason about.

tests/dialects/test_bigquery.py
```diff
@@ -2695,3 +2689,29 @@ def test_json_array(self):
...
+        self.validate_identity(
+            "DECLARE X,Y,Z INT DEFAULT (SELECT 42)", "DECLARE X, Y, Z INT64 DEFAULT (SELECT 42)"
+        )
+        self.validate_identity(
+            "DECLARE START_DATE DATE DEFAULT CURRENT_DATE() - 1",
+            "DECLARE START_DATE DATE DEFAULT CURRENT_DATE - 1",
+        )
+        self.validate_identity(
+            "DECLARE TS TIMESTAMP DEFAULT CURRENT_TIMESTAMP() - INTERVAL '1' HOUR",
+            "DECLARE TS TIMESTAMP DEFAULT CURRENT_TIMESTAMP() - INTERVAL '1' HOUR",
```
VaggelisD: [Nit] I'd personally make it so that all these tests simply showcase the roundtrip `DECLARE`, e.g instead of having:

```Python3
self.validate_identity("DECLARE X,Y,Z INT", "DECLARE X, Y, Z INT64")
self.validate_identity("DECLARE X,Y,Z INT DEFAULT 42", "DECLARE X, Y, Z INT64 DEFAULT 42")
...
```

I'd rewrite that as:

```Python3
self.validate_identity("DECLARE X, Y, Z INT64")
self.validate_identity("DECLARE X, Y, Z INT64 DEFAULT 42")
...
```

We shouldn't have to mix the concerns by testing `INT` -> `INT64`, `CURRENT_DATE` -> `CURRENT_DATE()` etc; Ideally, these should have their own coverage in `test_bigquery.py`

sqlglot/dialects/bigquery.py
```diff
@@ -1361,3 +1389,15 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
                     return f"{self.sql(expression, 'to')}{self.sql(this)}"
 
             return super().cast_sql(expression, safe_prefix=safe_prefix)
+
+        def declare_sql(self, expression: exp.Declare) -> str:
+            return f"DECLARE {self.expressions(expression, flat=True)}"
```
VaggelisD: This already exists in `generator.py` so we don't have to redefine it here, BigQuery's generator will inherit it

sqlglot/dialects/bigquery.py
```diff
@@ -937,6 +938,33 @@ def _parse_export_data(self) -> exp.Export:
...
 
+        def _parse_declare(self) -> exp.Declare | exp.Command:
+            index = self._index
+            expressions = self._try_parse(self._parse_declareitem)
+
+            if not expressions or self._curr:
+                self._retreat(index)
+                return self._parse_as_command(self._prev)
+
+            return self.expression(exp.Declare, expressions=[expressions])
```
VaggelisD, line 949: 1. This looks identical to`tsql.py::_parse_declare`, can we try to factor these out in `parse.py`?

2. [Nit] I'd rewrite this like so:

```Python3
from sqlglot.helper import ensure_list

...

def _parse_declare(self) -> exp.Declare | exp.Command:
  start = self._prev

  expressions = self._try_parse(self._parse_declareitem)

  if not expressions or self._curr:

  return self.expressions(exp.Declare, expressions=ensure_list(expressions))
```

And on a later pass we could try to remove `try_parse` here, I'm not sure if it had any real merit to it.

plaflamme: The `tsql` one is different: it allows multiple `declare_item`s separated by a comma, whereas BQ allows multiple vars within an item. So in tsql, the repetition is in `_parse_declare` whereas in BQ, it's in `_parse_declare_item`. I'm not sure we can consolidate them

sqlglot/dialects/bigquery.py
plaflamme: Fixed in https://github.com/tobymao/sqlglot/pull/5442/commits/2fb44d8b37a33672f81e82096e27601ae10cf947

sqlglot/dialects/bigquery.py
plaflamme: Fixed in https://github.com/tobymao/sqlglot/pull/5442/commits/2fb44d8b37a33672f81e82096e27601ae10cf947

tests/dialects/test_bigquery.py
plaflamme: Fixed in https://github.com/tobymao/sqlglot/pull/5442/commits/38228e89721774612fdb8641a37fb19f4c1f0fb9

tests/dialects/test_bigquery.py
plaflamme: Fixed in https://github.com/tobymao/sqlglot/pull/5442/commits/38228e89721774612fdb8641a37fb19f4c1f0fb9

sqlglot/dialects/bigquery.py
plaflamme: Fixed in https://github.com/tobymao/sqlglot/pull/5442/commits/39058e98a9eca93e13085496c6f3c05d753b321a

PR #5447, opened 2025-07-22
feat(singlestore): Added initial implementation of SingleStore dialect

AdalbertMemSQL:
Updated __init__.py to contain SingleStore dialect.
Created a dialect class and set basic constants.
Added methods for better testing of the dialect functionality.


tests/dialects/test_singlestore.py
georgesittas, line 1: We don't, at this point, have integration tests for engines like what you've implemented in this file. Let's please revert this and follow the example of other `test_<dialect>.py` files, which mostly test the roundtrip & transpilation of SQL statements.

sqlglot/dialects/singlestore.py
```diff
@@ -0,0 +1,15 @@
+from sqlglot import Dialect
+from sqlglot.dialects.dialect import NormalizationStrategy
+
+
+class SingleStore(Dialect):
```
georgesittas: Is `SingleStore` an extension of `MySQL`? Just curious because I saw [this page](https://docs.singlestore.com/cloud/developer-resources/unsupported-mysql-features/), and also the dialect-level constants generally match so far.

AdalbertMemSQL: SingleStore is protocol-level compatible with MySQL, and its SQL dialect is similar, though there are quite a few differences.
When implementing this dialect, I found it more manageable to go through the original constants directly and validate each one, rather than extending from MySQL and overriding as needed.
That said, since the constants align closely so far—as you pointed out—I'll go ahead and extend from the MySQL dialect in this case. Thanks for the note!

georgesittas: > its SQL dialect is similar, though there are quite a few differences.

If the two diverge significantly, then what you have is fine. Other examples of inheritance are Postgres -> Redshift, Presto -> Trino, etc. Just noting to give a sense for when we use inheritance.

georgesittas:
Hey @AdalbertMemSQL, any plans to finish this?

AdalbertMemSQL:
Hey @georgesittas ! 
Yes, I definitely plan to finish this — just had a lot of other tasks come up recently. I’ll get back to it soon. Thanks for checking in!

georgesittas:
Of course, no problem! Thanks for the update :)

sqlglot/dialects/singlestore.py
```diff
@@ -0,0 +1,6 @@
+from sqlglot.dialects.mysql import MySQL
+
+
+class SingleStore(MySQL):
+    FORCE_EARLY_ALIAS_REF_EXPANSION = True
```
georgesittas: Could you share the documentation or any possible testing you did related to this flag?

AdalbertMemSQL: I didn't find documentation related to this functionality, but manual testing shows the following
```
        WITH data AS (
        SELECT
            1 AS id,
            2 AS my_id
        )
        SELECT
            id AS my_id
        FROM
            data
        WHERE
            my_id = 1 <- this is data.my_id
        GROUP BY
            my_id <- this is data.my_id
        HAVING
            my_id = 1 <- this is data.id
```
Do I understand correctly that in this case, `FORCE_EARLY_ALIAS_REF_EXPANSION` should be set to true?

georgesittas: You can test this by parsing the query with `singlestore` and running it through `qualify` (again, set `dialect="singlestore"`). If the expansion produces a query with equivalent semantics, then this flag is most likely set correctly right now.

cc @VaggelisD

AdalbertMemSQL: I tested and noticed that it returns the correct result when this flag is set to False. Thanks for the hint

georgesittas: Awesome :)

PR #5448, opened 2025-07-22
Feat(duckdb): Add reset command for duckdb

themisvaltinos:
https://duckdb.org/docs/stable/sql/statements/set.html#reset


sqlglot/tokens.py
```diff
@@ -371,6 +371,7 @@ class TokenType(AutoName):
     RECURSIVE = auto()
     REFRESH = auto()
     RENAME = auto()
+    RESET = auto()
```
georgesittas:
```suggestion
```

PR #5451, opened 2025-07-22
Feat!: improve BigQuery `UNNEST` transpilation

georgesittas:
This PR aims to improve the transpilation of `UNNEST` expressions in BigQuery that are transformed using `qualify_tables`:

```python
>>> from sqlglot.optimizer.qualify_tables import qualify_tables
>>>
>>> sql = """
...         SELECT x, ys, zs
...         FROM UNNEST([STRUCT('x' AS x, ['y1', 'y2', 'y3'] AS y, ['z1', 'z2', 'z3'] AS z)]),
...         UNNEST(y) AS ys,
...         UNNEST(z) AS zs
... """
>>>
>>> ast = parse_one(sql, "bigquery")
>>> qualified = qualify_tables(ast, dialect="bigquery")
>>>
>>> print(ast.sql("snowflake", pretty=True))  # `_q_0.y` is an invalid reference so this query can't be executed
SELECT
  _q_0.x,
  ys,
  zs
FROM TABLE(FLATTEN(INPUT => [OBJECT_CONSTRUCT('x', 'x', 'y', ['y1', 'y2', 'y3'], 'z', ['z1', 'z2', 'z3'])])) AS _q_0(seq, key, path, index, value, this)
CROSS JOIN TABLE(FLATTEN(INPUT => _q_0.y)) AS _q_1(seq, key, path, index, ys, this)
CROSS JOIN TABLE(FLATTEN(INPUT => _q_0.z)) AS _q_2(seq, key, path, index, zs, this)
```

### DIFF
--- a/sqlglot/optimizer/qualify_tables.py
+++ b/sqlglot/optimizer/qualify_tables.py
@@ -128,6 +128,14 @@ def qualify_tables(
                 table_alias = udtf.args.get("alias") or exp.TableAlias(
                     this=exp.to_identifier(next_alias_name())
                 )
+                if (
+                    isinstance(udtf, exp.Unnest)
+                    and dialect.UNNEST_COLUMN_ONLY
+                    and not table_alias.columns
+                ):
+                    table_alias.set("columns", [table_alias.this.copy()])
+                    table_alias.set("column_only", True)
+
                 udtf.set("alias", table_alias)
 
                 if not table_alias.name:
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -461,7 +461,7 @@ def _eliminate_dot_variant_lookup(expression: exp.Expression) -> exp.Expression:
             unnest_alias = unnest.args.get("alias")
             if (
                 isinstance(unnest_alias, exp.TableAlias)
-                and not unnest_alias.this
+                and (unnest_alias.args.get("column_only") or not unnest_alias.this)
                 and len(unnest_alias.columns) == 1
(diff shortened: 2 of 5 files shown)


sqlglot/optimizer/qualify_tables.py
```diff
@@ -128,6 +128,14 @@ def _qualify(table: exp.Table) -> None:
                 table_alias = udtf.args.get("alias") or exp.TableAlias(
                     this=exp.to_identifier(next_alias_name())
                 )
+                if (
+                    isinstance(udtf, exp.Unnest)
+                    and dialect.UNNEST_COLUMN_ONLY
+                    and not table_alias.columns
+                ):
+                    table_alias.set("columns", [table_alias.this.copy()])
+                    table_alias.set("column_only", True)
```
georgesittas, line 137: This is the interesting bit in the PR. Previously, if there was no table alias on an `UNNEST`, `qualify_tables` would produce an alias that looked like `TableAlias(this="_q_0")`. Now, it copies `this` into `columns` as well and sets this auxiliary arg to true to mark the expression accordingly.

The rationale was that if we do this, then there's no issue with qualifying a column as `_q_0.x`, because Snowflake will convert the dot syntax into bracket syntax in `_eliminate_dot_variant_lookup` and the resulting expression will resolve into the correct value.

This is because having the same alias in `columns` means that the generated table alias for `TABLE(FLATTEN(...))` will have that name (`_q_0`) instead of `value`, and so `_q_0['x']` will be a bracket applied to said column.

PR #5452, opened 2025-07-22
feat: support `DESC SEMANTIC VIEW`

betodealmeida:
Improve the recent support for semantic views in Snowflake, allowing us to parse:

```sql
DESC SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM;
```

And:

```sql
DESC SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM
    ->> SELECT "object_kind","property_value" as "parent_object","object_name" FROM $1
        WHERE "object_kind" IN ('METRIC','DIMENSION') AND "property" IN ('TABLE')
;
```

From the [guide](https://docs.snowflake.com/en/user-guide/views-semantic/sql#viewing-the-details-about-a-semantic-view) and the [quickstart](https://quickstarts.snowflake.com/guide/snowflake-semantic-view/index.html#3).

Currently we get this behavior:

```python
>>> import sqlglot
>>> sqlglot.parse_one("DESC SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM;", "snowflake")
  File "/Users/beto/Projects/sqlglot/sqlglot/__init__.py", line 139, in parse_one
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/beto/Projects/sqlglot/sqlglot/dialects/dialect.py", line 1058, in parse
  File "/Users/beto/Projects/sqlglot/sqlglot/parser.py", line 1598, in parse
    return self._parse(
           ^^^^^^^^^^^^
  File "/Users/beto/Projects/sqlglot/sqlglot/parser.py", line 1670, in _parse
  File "/Users/beto/Projects/sqlglot/sqlglot/parser.py", line 1711, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 18.
  DESC SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM;
>>>
```

With this PR:

```python
>>> import sqlglot
>>> sqlglot.parse_one("DESC SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM;", "snowflake")
Describe(
  this=Table(
    this=Identifier(this=TPCDS_SEMANTIC_VIEW_SM, quoted=False)),
  kind=SEMANTIC VIEW)
>>>
```


sqlglot/expressions.py
```diff
@@ -1558,6 +1558,7 @@ class Describe(Expression):
         "expressions": False,
         "partition": False,
         "format": False,
+        "query": False,
```
VaggelisD: I don't think we should solve this like so; It _might_ be okay for a temporary workaround, but this syntax refers to [Snowflake's flow operators](https://docs.snowflake.com/en/sql-reference/operators-flow) which require a totally different way of parsing them i.e similarly to what we have been doing for BigQuery's pipe syntax, if you want to take a look at the recent PRs. 

Also, note that this allows only a single `SELECT` / flow operator, so it might quickly hit another parse error.

@betodealmeida Do you use the flow operator with `DESCRIBE` or was it added to improve coverage? 

cc: @georgesittas @geooo109 curious to hear your opinion here

tests/dialects/test_snowflake.py
```diff
@@ -2167,6 +2167,24 @@ def test_describe_table(self):
...
+            write={
+                "snowflake": "DESCRIBE SEMANTIC VIEW my_semantic_view",
+            },
+        )
+        self.validate_all(
+            "DESC SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM",
+            write={
+                "snowflake": "DESCRIBE SEMANTIC VIEW TPCDS_SEMANTIC_VIEW_SM",
+            },
+        )
```
VaggelisD: These 2 tests look identical, I guess we could do with one of them.

Otherwise, we can turn one of them to:

```Python3
self.validate_identity("DESCRIBE SEMANTIC VIEW <identifier>
```

sqlglot/expressions.py
georgesittas: Agreed, we shouldn't half-implement it only for `DESCRIBE`.

georgesittas: Let's remove the `->>` logic for now. If there's a workaround that allows parsing the whole thing as a `Command` if `->>` is detected, I'm onboard with that (since parsing would just fail), but I'd otherwise prefer to avoid having to maintain an incomplete implementation of Snowflake's flow operators.

tests/dialects/test_snowflake.py
betodealmeida: Oh, sorry, I thought I had removed this. Will do!

sqlglot/expressions.py
betodealmeida: That's a great point. I was adding it to support the examples from the Snowflake semantic layer quickstart, but we can do a simpler:

```sql
SELECT name FROM information_schema.semantic_views;
```

Thanks for the feedback!

PR #5456, opened 2025-07-23
feat(exasol): intercept rank function in sqlglot to remove arguments

nnamdi16:


georgesittas:
Hey @nnamdi16, this isn't quite what I meant in the comment I left in your other PR. We don't plan to merge logic related to `RANK`, that aims to handle an undocumented (and possible no-op?) feature. This is something you'd have to take care in your application's side. If the argument has a functional purpose, I'm happy to re-open and consider introducing an expression to represent its semantics.

PR #4333, opened 2024-11-01
feat(bigquery): Support JSON_QUERY

VaggelisD:
Currently SQLGlot supports `JSON_EXTRACT` which is deprecated in favor of `JSON_QUERY`:

```SQL
bigquery> SELECT JSON_QUERY('{"class": {"students": []}}', '$.class')
{"students":[]}

snowflake> SELECT GET_PATH(PARSE_JSON('{"class": {"students": []}}'), 'class');
{"students": []}

duckdb> SELECT '{"class": {"students": []}}' -> '$.class';
┌──────────────────────────────────────────────┐
│ ('{"class": {"students": []}}' -> '$.class') │
│                     json                     │
├──────────────────────────────────────────────┤
│ {"students":[]}                              │
└──────────────────────────────────────────────┘
```

This PR also improves Snowflake's `GET_PATH` as it's unable to process JSON strings (unlike BQ and DuckDB), so as an extra step the string will be parsed first.

Docs
-------
[BQ JSON_EXTRACT](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract) | [BQ JSON_QUERY](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_query)

PR #5458, opened 2025-07-23
fix(snowflake)!: Wrap `GET_PATH` value with `PARSE_JSON` preemptively

VaggelisD:
https://github.com/tobymao/sqlglot/pull/4333 fixed the transpilation of `exp.JsonExtract` towards Snowflake by wrapping the LHS with `PARSE_JSON` if it was a string literal. However, this solution does not suffice as the value can be a column containing a string.

- Behavior before the PR:

```Python3
SQLGlot transpilation
------------------------
>>> sql = "WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT JSON_EXTRACT(foo, '$.name') FROM tbl"
>>> sqlglot.parse_one(sql, "bigquery").sql("snowflake)
'WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT GET_PATH(foo, 'name') FROM tbl AS tbl'


Engine execution
--------------------
bq> WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT JSON_EXTRACT(foo, '$.name') FROM tbl;
--
"Jakob"

snowflake> WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT GET_PATH(foo, 'name') FROM tbl AS tbl;
--
Error: Invalid argument types for function 'GET_PATH': (VARCHAR(29), VARCHAR(4))
```

<br />

- After the PR (notice the `GET_PATH` value):

```Python3
SQLGlot transpilation
------------------------
>>> sql = "WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT JSON_EXTRACT(foo, '$.name') FROM tbl"
>>> sqlglot.parse_one(sql, "bigquery").sql("snowflake)
'WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT GET_PATH(PARSE_JSON(foo), 'name') FROM tbl AS tbl'


Engine execution
--------------------
snowflake> WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT GET_PATH(PARSE_JSON(foo), 'name') FROM tbl AS tbl
--
"Jakob"
```

<br />

- PS 1: In Snowflake, `PARSE_JSON` seems idempotent so even if we already have a `JSON` (i.e `VARIANT`) expression the execution is not altered:

```SQL
snowflake> WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT GET_PATH(PARSE_JSON(PARSE_JSON(foo)), 'name') FROM tbl AS tbl;
--
"Jacob"

snowflake> WITH tbl AS (SELECT '{"name": "Jakob", "age": "6"}' AS foo) SELECT PARSE_JSON(PARSE_JSON(foo)) = PARSE_JSON(foo) FROM tbl;
--
TRUE
```

- PS 2: Even so, in order to not alter the Snowflake roundtrip, a new flag & arg was added to avoid reparsing.

Docs
--------
[BigQuery JSON_EXTRACT](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract) | [Snowflake GET_PATH](https://docs.snowflake.com/en/sql-reference/functions/get_path)


tobymao:
add a test that shows you don't double add parse_json to an existing parse_json

VaggelisD:
@tobymao Already did in [test_snowflake.py](https://github.com/tobymao/sqlglot/pull/5458/files#diff-48950cdaa69b6d8abaa3a1207fc43f4676296c3d36b2dd74e782b9b48c202511), did you have something else in mind?

tobymao:
sorry i didn't see it

sqlglot/dialects/snowflake.py
```diff
@@ -1555,10 +1558,16 @@ def timestampsub_sql(self, expression: exp.TimestampSub):
         def jsonextract_sql(self, expression: exp.JSONExtract):
             this = expression.this
 
-            # JSON strings are valid coming from other dialects such as BQ
+            # JSON strings are valid coming from other dialects such as BQ so
+            # for these cases we PARSE_JSON preemptively
+            if not isinstance(this, (exp.ParseJSON, exp.JSONExtract)) and not expression.args.get(
+                "requires_json"
+            ):
+                this = exp.ParseJSON(this=this)
```
georgesittas, line 1566: Snowflake's `GET_PATH` [supports a few different types for the first argument](https://docs.snowflake.com/en/sql-reference/functions/get_path#arguments). Have you verified that in all cases, wrapping with `PARSE_JSON` is at least a no-op?

VaggelisD: Right, in Snowflake `GET_PATH` can be used with a few different types. However, my assumptions were:
- During roundtrips it'll remain unaffected, so that should be fine
- During transpilation, I'm assuming that other dialect's `exp.JsonExtract` will contain a JSON string or JSON expression. I don't think it'd be easily possible to transpile an expression such as `exp.JsonExtract(array, index)` as would happen in Snowflake e.g in `SELECT GET_PATH([1, 2], '[0]');`

georgesittas: Ah, right, roundtrip is fine due to the flag you introduced. I missed this during my review. Looks good then, thanks for clarifying 👍

PR #5466, opened 2025-07-24
fix(snowflake)!: Remove `UNKNOWN` type from `TRY_CAST`

VaggelisD:
https://github.com/tobymao/sqlglot/pull/2560 made it so that Snowflake's `TRY_CAST` is generated only if the value is of type `{TEXT, UNKNOWN}`; The latter was added to preserve roundtrips but can hurt transpilation attempts coming from other dialects.

To fix this, this PR adds a new boolean arg to `exp.Cast` and removes `UNKNOWN` types from that branch as a "best effort" transpilation attempt.

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5755,6 +5755,7 @@ class Cast(Func):
         "safe": False,
         "action": False,
         "default": False,
+        "requires_string": False,
     }
 
     @property
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5755,7 +5755,6 @@ class Cast(Func):
         "safe": False,
         "action": False,
         "default": False,
-        "requires_string": False,
     }
 
     @property
@@ -5786,7 +5785,7 @@ class Cast(Func):
 
 
 class TryCast(Cast):
-    pass
+    arg_types = {**Cast.arg_types, "requires_string": False}
 
 
 # https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -759,11 +759,8 @@ class Parser(metaclass=_Parser):
             this=this,
             to=to,
         ),
-        TokenType.DCOLON: lambda self, this, to: self.expression(
-            exp.Cast if self.STRICT_CAST else exp.TryCast,
-            this=this,
-            to=to,
-            requires_string=self.dialect.TRY_CAST_REQUIRES_STRING,
+        TokenType.DCOLON: lambda self, this, to: self.build_cast(
+            strict=self.STRICT_CAST, this=this, to=to
         ),
         TokenType.ARROW: lambda self, this, path: self.expression(
             exp.JSONExtract,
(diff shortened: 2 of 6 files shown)


sqlglot/expressions.py
```diff
@@ -5755,6 +5755,7 @@ class Cast(Func):
         "safe": False,
         "action": False,
         "default": False,
+        "requires_string": False,
```
georgesittas: Why was this added here instead of `TryCast`?

tests/dialects/test_snowflake.py
```diff
@@ -2605,10 +2605,10 @@ def test_try_cast(self):
...
         self.validate_all(
-            "TRY_CAST(x AS TEXT)",
+            "TRY_CAST('val' AS TEXT)",
             read={
-                "hive": "CAST(x AS STRING)",
-                "snowflake": "TRY_CAST(x AS TEXT)",
+                "hive": "CAST('val' AS STRING)",
+                "snowflake": "TRY_CAST('val' AS TEXT)",
             },
         )
```
georgesittas, line 2614: Do we test non-string inputs in Snowflake today?

sqlglot/expressions.py
VaggelisD: Because `TryCast`, `JsonCast` etc simply inherit what is in `Cast`:

```Python3
# expressions.py

class TryCast(Cast):
    pass
```

tests/dialects/test_snowflake.py
VaggelisD: Yeah, there are other `TRY_CAST` tests above that one, they're all under `test_snowflake.py::test_try_cast` so we're already testing roundtrips etc

VaggelisD: Actually, now that you say it these can be simplified, will update shortly

VaggelisD: Done in https://github.com/tobymao/sqlglot/pull/5466/commits/c93889b80b52567d352cd368c58fd48b72a39b5d, refactored those tests slightly

tests/dialects/test_snowflake.py
```diff
@@ -2605,28 +2606,24 @@ def test_try_cast(self):
...
 
         from sqlglot.optimizer.annotate_types import annotate_types
 
         expression = parse_one("SELECT CAST(t.x AS STRING) FROM t", read="hive")
 
-        expression = annotate_types(expression, schema={"t": {"x": "string"}})
-        self.assertEqual(expression.sql(dialect="snowflake"), "SELECT TRY_CAST(t.x AS TEXT) FROM t")
+        for value_type in ("string", "int"):
```
georgesittas, line 2620: Do we need `self.subTest` here?

sqlglot/expressions.py
georgesittas: Discussed in Slack about decoupling the two `arg_types` by using a base dictionary that can be augmented.

sqlglot/parser.py
```diff
@@ -8648,3 +8646,11 @@ def _parse_declare(self) -> exp.Declare | exp.Command:
             return self._parse_as_command(start)
 
         return self.expression(exp.Declare, expressions=expressions)
+
+    def build_cast(self, strict: bool, **kwargs) -> exp.Cast | exp.TryCast:
```
```suggestion
    def build_cast(self, strict: bool, **kwargs) -> exp.Cast:
```

georgesittas (approved):

PR #5476, opened 2025-07-25
feat(singlestore): Fixed time formatting

AdalbertMemSQL:
SingleStore uses different time formatting formats in different functions.
One of the formats is compatible with MySQL
Related documentation:
 * [TO_DATE](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/to-date/)
 * [TO_TIMESTAMP](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/to-timestamp/)
 * [TO_CHAR](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/to-char/)
 * [STR_TO_DATE](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/str-to-date/)
 * [DATE_FORMAT](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/date-format/)
 * [TIME_FORMAT](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/time-format/)

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -1,5 +1,94 @@
+from sqlglot.dialects.dialect import build_formatted_time
 from sqlglot.dialects.mysql import MySQL
+import typing as t
+from sqlglot import exp, Dialect
+from sqlglot.generator import unsupported_args
+from sqlglot.helper import seq_get
 
 
 class SingleStore(MySQL):
     SUPPORTS_ORDER_BY_ALL = True
+
+    TIME_MAPPING: t.Dict[str, str] = {
+        "AM": "%p",  # Meridian indicator with or without periods
+        "A.M.": "%p",  # Meridian indicator with or without periods
+        "PM": "%p",  # Meridian indicator with or without periods
+        "P.M.": "%p",  # Meridian indicator with or without periods
+        "D": "%u",  # Day of week (1-7)
+        "DD": "%d",  # day of month (1-31)
+        "DY": "%a",  # abbreviated name of day
+        "HH": "%I",  # Hour of day (1-12)
+        "HH12": "%I",  # alias for HH
+        "HH24": "%H",  # Hour of day (0-23)
+        "MI": "%M",  # Minute (0-59)
+        "MM": "%m",  # Month (01-12; January = 01)
+        "MON": "%b",  # Abbreviated name of month
+        "MONTH": "%B",  # Name of month
+        "SS": "%S",  # Second (0-59)
+        "RR": "%y",  # 15
+        "YY": "%y",  # 15
+        "YYYY": "%Y",  # 2015
+        "FF6": "%f",  # only 6 digits are supported in python formats
+    }
+
+    class Parser(MySQL.Parser):
+        FUNCTIONS = {
+            **MySQL.Parser.FUNCTIONS,
+            "TO_DATE": build_formatted_time(exp.StrToDate, "singlestore"),
+            "TO_TIMESTAMP": build_formatted_time(exp.StrToTime, "singlestore"),
+            "TO_CHAR": build_formatted_time(exp.TimeToStr, "singlestore"),
+            "STR_TO_DATE": build_formatted_time(exp.StrToDate, "mysql"),
+            "DATE_FORMAT": build_formatted_time(exp.TimeToStr, "mysql"),
+            # The first argument is converted to TIME(6)
+            # This is needed because exp.TimeToStr is converted to DATE_FORMAT
+            # which interprets the first argument as DATETIME and fails to parse
+            # string literals like '12:05:47' without a date part.
+            "TIME_FORMAT": lambda args: exp.TimeToStr(
+                this=exp.Cast(
+                    this=seq_get(args, 0),
+                    to=exp.DataType.build(
+                        exp.DataType.Type.TIME,
+                        expressions=[exp.DataTypeParam(this=exp.Literal.number(6))],
+                    ),
+                ),
+                format=Dialect["mysql"].format_time(seq_get(args, 1)),
+            ),
+        }
+
+    class Generator(MySQL.Generator):
+        TRANSFORMS = {
+            **MySQL.Generator.TRANSFORMS,
+            exp.StrToDate: lambda self, e: self.func(
+                "STR_TO_DATE",
+                e.this,
+                self.format_time(
+                    e,
+                    inverse_time_mapping=Dialect["mysql"].INVERSE_TIME_MAPPING,
+                    inverse_time_trie=Dialect["mysql"].INVERSE_TIME_TRIE,
+                ),
+            ),
+            exp.TimeToStr: lambda self, e: self.func(
+                "DATE_FORMAT",
+                e.this,
+                self.format_time(
+                    e,
+                    inverse_time_mapping=Dialect["mysql"].INVERSE_TIME_MAPPING,
+                    inverse_time_trie=Dialect["mysql"].INVERSE_TIME_TRIE,
+                ),
+            ),
+            exp.StrToTime: lambda self, e: self.func(
+                "STR_TO_DATE",
+                e.this,
+                self.format_time(
+                    e,
+                    inverse_time_mapping=Dialect["mysql"].INVERSE_TIME_MAPPING,
+                    inverse_time_trie=Dialect["mysql"].INVERSE_TIME_TRIE,
+                ),
+            ),
+        }
+
+        @unsupported_args("format")
+        @unsupported_args("action")
+        @unsupported_args("default")
+        def cast_sql(self, expression: exp.Cast) -> str:  # type: ignore
+            return f"{self.sql(expression, 'this')} :> {self.sql(expression, 'to')}"
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -1,7 +1,7 @@
 from sqlglot.dialects.dialect import build_formatted_time
 from sqlglot.dialects.mysql import MySQL
 import typing as t
-from sqlglot import exp, Dialect
+from sqlglot import exp
 from sqlglot.helper import seq_get
 
 
@@ -29,9 +29,9 @@ class SingleStore(MySQL):
     class Parser(MySQL.Parser):
         FUNCTIONS = {
             **MySQL.Parser.FUNCTIONS,
-            "TO_DATE": build_formatted_time(exp.StrToDate, "singlestore"),
+            "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "singlestore"),
             "TO_TIMESTAMP": build_formatted_time(exp.StrToTime, "singlestore"),
-            "TO_CHAR": build_formatted_time(exp.TimeToStr, "singlestore"),
+            "TO_CHAR": build_formatted_time(exp.ToChar, "singlestore"),
             "STR_TO_DATE": build_formatted_time(exp.StrToDate, "mysql"),
             "DATE_FORMAT": build_formatted_time(exp.TimeToStr, "mysql"),
             # The first argument is converted to TIME(6)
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,94 @@
...
 
 
 class SingleStore(MySQL):
     SUPPORTS_ORDER_BY_ALL = True
+
+    TIME_MAPPING: t.Dict[str, str] = {
+        "AM": "%p",  # Meridian indicator with or without periods
+        "A.M.": "%p",  # Meridian indicator with or without periods
+        "PM": "%p",  # Meridian indicator with or without periods
+        "P.M.": "%p",  # Meridian indicator with or without periods
```
georgesittas: These aren't correct. Keys in `TIME_MAPPING` that map to the same value collapse to a single string when generated, which happens to be the last such key in the dictionary (i.e., `"P.M."`). This means that different formats must only map to the same value as long as they are equivalent.

I suggest leaving these four entries out of the `TIME_MAPPING`.

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,94 @@
...
+        "D": "%u",  # Day of week (1-7)
+        "DD": "%d",  # day of month (1-31)
```
georgesittas: `%d` includes zero-padding ([docs](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior)):

> Day of the month as a zero-padded decimal number.

Should this be `"%-d"`? Singlestore's docs don't show padding:

Screenshot 2025-07-25 at 5 17 19 PM

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,94 @@
...
+        "DY": "%a",  # abbreviated name of day
+        "HH": "%I",  # Hour of day (1-12)
+        "HH12": "%I",  # alias for HH
+        "HH24": "%H",  # Hour of day (0-23)
+        "MI": "%M",  # Minute (0-59)
+        "MM": "%m",  # Month (01-12; January = 01)
```
georgesittas, line 19: Ditto about padding and using `%-m` instead:

Screenshot 2025-07-25 at 5 18 34 PM

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,94 @@
...
+                    inverse_time_trie=Dialect["mysql"].INVERSE_TIME_TRIE,
+                ),
+            ),
+        }
+
+        @unsupported_args("format")
+        @unsupported_args("action")
+        @unsupported_args("default")
+        def cast_sql(self, expression: exp.Cast) -> str:  # type: ignore
+            return f"{self.sql(expression, 'this')} :> {self.sql(expression, 'to')}"
```
georgesittas: Why was this overridden? Doesn't Singlestore support the standard `CAST` syntax?

georgesittas (commented):
Some quick comments

sqlglot/dialects/singlestore.py
AdalbertMemSQL: SingleStore supports CAST syntax, but it doesn't work with all datatypes (it handles only ~8 types when there are more than 30), so using `:>` is preferable.

georgesittas: Can we leave this out of this PR? It seems unrelated to date/time stuff. Let's deal with casts in a followup PR, so that support for it can be properly implemented, e.g., adding parsing support for `:>` and perhaps `!:>`.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: The same entries are presented in Oracle and Postgres dialects
https://github.com/tobymao/sqlglot/blob/787d167d694b557d6e43ed391f59847a888fa572/sqlglot/dialects/oracle.py#L55
https://github.com/tobymao/sqlglot/blob/787d167d694b557d6e43ed391f59847a888fa572/sqlglot/dialects/postgres.py#L277

georgesittas: Yep, these suffer from the same issue. I'll fix them separately, thanks for pointing it out.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: In SingleStore, this is also zero-padded. This is a mistake in the documentation.
Here is an example
```
singlestore> select TO_CHAR('2000-01-01', 'MM DD');
+--------------------------------+
| TO_CHAR('2000-01-01', 'MM DD') |
+--------------------------------+
| 01 01                          |
+--------------------------------+
1 row in set (0.06 sec)
```

georgesittas: Sounds good, thanks for verifying 👍

sqlglot/dialects/singlestore.py
AdalbertMemSQL: Yep, removed this

sqlglot/dialects/singlestore.py
AdalbertMemSQL: The same as days - this is a bug in the SingleStore documentation

sqlglot/dialects/singlestore.py
georgesittas: For reference: https://github.com/tobymao/sqlglot/pull/5491

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,83 @@
...
+            # which interprets the first argument as DATETIME and fails to parse
+            # string literals like '12:05:47' without a date part.
+            "TIME_FORMAT": lambda args: exp.TimeToStr(
+                this=exp.Cast(
+                    this=seq_get(args, 0),
+                    to=exp.DataType.build(
+                        exp.DataType.Type.TIME,
+                        expressions=[exp.DataTypeParam(this=exp.Literal.number(6))],
+                    ),
+                ),
```
georgesittas, line 48: Is this correct? From the docs:

> `TIME_FORMAT (timeobj, format)`
> `timeobj`: a valid time, date, datetime, or parsable date string

What if you get a datetime value (string or not) and then you cast it into a `TIME(6)`? Also, curious why the size 6 was used specifically.

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,83 @@
...
+                format=Dialect["mysql"].format_time(seq_get(args, 1)),
```
georgesittas: I'm iffy about `Dialect["mysql"]`, I don't think we use this pattern elsewhere in the codebase. Can't we do `MySQL.format_time`?

tests/dialects/test_singlestore.py
```diff
@@ -18,3 +18,23 @@ def test_basic(self):
             "SELECT `data`.`id` AS `my_id` FROM `data` AS `data` WHERE `data`.`my_id` = 1 GROUP BY `data`.`my_id` HAVING `data`.`id` = 1",
             ast.sql(dialect=self.dialect),
         )
+
+    def test_time_formatting(self):
+        self.validate_identity(
+            "SELECT TO_DATE('03/01/2019', 'MM/DD/YYYY') AS result",
+            write_sql="SELECT STR_TO_DATE('03/01/2019', '%m/%d/%Y') AS result",
```
georgesittas: Nit: let's remove `write_sql` and make the argument positional to match the existing style.
```suggestion
            "SELECT STR_TO_DATE('03/01/2019', '%m/%d/%Y') AS result",
```

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,83 @@
...
+            **MySQL.Generator.TRANSFORMS,
+            exp.StrToDate: lambda self, e: self.func(
+                "STR_TO_DATE",
+                e.this,
+                self.format_time(
+                    e,
+                    inverse_time_mapping=Dialect["mysql"].INVERSE_TIME_MAPPING,
+                    inverse_time_trie=Dialect["mysql"].INVERSE_TIME_TRIE,
+                ),
+            ),
```
georgesittas: Why are we using MySQL's `STR_TO_DATE` instead of Singlestore's `TO_DATE`? This makes the logic here a bit more complicated, having to pass the inverse mappings and all.

AdalbertMemSQL: The main reason I am using `STR_TO_DATE` here is that it supports a broader range of format specifiers, which makes it more flexible for parsing complex datetime strings.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: Just checked that it can be done. Will update

sqlglot/dialects/singlestore.py
georgesittas: Got it. Is its functionality a superset compared to `TO_DATE`? I've similar questions for the rest of these date/time functions that are mapped to the respective MySQL variants.

AdalbertMemSQL: `STR_TO_DATE` isn’t a full superset of `TO_DATE` - for example, it doesn’t support `D` or `FFn` when `n` is not `6`. However, since both `STR_TO_DATE` and `TO_DATE` are mapped to the same `exp.StrToDate` expression in SQLGlot, it makes sense to consistently use `STR_TO_DATE` here. This choice helps preserve the widest range of formatting functionality.

The same is true for `exp.TimeToStr` (`TO_CHAR`, `DATE_FORMAT`).

As for `exp.StrToTime` it make sense to use `TO_TIMESTAMP` here

sqlglot/dialects/singlestore.py
AdalbertMemSQL: `What if you get a datetime value (string or not) and then you cast it into a TIME(6)?` 
In this case, the date part will be deleted.
`Also, curious why the size 6 was used specifically.`
6 is the maximum precision of time in SingleStore. It was used to preserve microseconds and don't lose information.
I tested it more and found several situations when it can return incorrect results. I'm going to change this.

sqlglot/dialects/singlestore.py
georgesittas: If `STR_TO_DATE` cannot fully replace `TO_DATE`, then I think we should reconsider the ASTs we emit for these functions.

SQLGlot needs to preserve the semantics when roundtripping a query in the same dialect, so if you start with `TO_DATE(..., "...FF6...")` and end up with `STR_TO_DATE(..., "...FF6...")`, you'll end up breaking that invariant, since the latter is not supported AFAICT.

Alternatively, you can keep the current AST representation, but ensure that, at generation time, you can guarantee that the unsupported formats will be detected and `TO_DATE` will be emitted as needed in those cases.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: In my testing, I used format symbols for the date part, but according to the docs 
```
format: a string containing format symbols. Only format symbols for the hour, minute, second, and microsecond parts are valid.
```
So this solution seems to work well on the valid queries.
The date part (which is not needed for formatting) is deleted.

georgesittas: Ok, so the argument is that the `CAST` should be fine since `TIME_FORMAT` truncates non-time related segments in the input as well? Can casting fail on any of the values in the domain of `TIME_FORMAT`?

AdalbertMemSQL: Hmm...
Parsing fails for the DATE strings without TIME part
```
singlestore> SELECT '2025-01-01' :> TIME;         
ERROR 2827 (HY000): Invalid DATE/TIME in type conversion
```
But the same happens with the original function
```
singlestore> select TIME_FORMAT ('2025-01-01', '%Y');
ERROR 2827 (HY000): Invalid DATE/TIME in type conversion
```

sqlglot/dialects/singlestore.py
AdalbertMemSQL: Changed AST representation

georgesittas (approved):

PR #5477, opened 2025-07-25
fix: only strip kind from joins when it is inner|outer

### LINKED ISSUE #5470: `sqlglot.optimizer.optimize_joins.normalize` removes meaningful `kind` values from expression
**Before you file an issue**

```python
from sqlglot import parse_one
from sqlglot.optimizer.optimize_joins import normalize

query = """
SELECT `bar`
FROM `my_table`
LEFT ANTI JOIN `my_other_table` ON `my_table`.`foo` = `my_other_table`.`foo`
"""

expression = parse_one(query, read="spark")
print(normalize(expression).sql())
```

```sql
SELECT "bar" FROM "my_table" LEFT JOIN "my_other_table" ON "my_table"."foo" = "my_other_table"."foo"
```

The query is fine prior to the optimizer, but because optimize sends it through `normalize`, the kind gets stripped here:

https://github.com/tobymao/sqlglot/blob/406815de21f0fdc9874ff46155d4ee0274aa6337/sqlglot/optimizer/optimize_joins.py#L79-L82

The expected query should be:

```sql
SELECT "bar" FROM "my_table" LEFT ANTI JOIN "my_other_table" ON "my_table"."foo" = "my_other_table"."foo"
```

If you're open to it, I can create a PR for this, continuing to strip the kind for inner outer joins, as it seems that is the intention:

https://github.com/tobymao/sqlglot/blob/406815de21f0fdc9874ff46155d4ee0274aa6337/sqlglot/optimizer/optimize_joins.py#L72-L74

(but keeping it otherwise for kinds like `anti`, `semi`)

themattmorris:
closes #5470 

This PR ensures that the only kinds that get stripped from joins are `INNER` and `OUTER` when optimizing an expression. For example, `LEFT ANTI`:

```python
from sqlglot import parse_one
from sqlglot.optimizer.optimize_joins import normalize

query = """
SELECT `bar`
FROM `my_table`
LEFT ANTI JOIN `my_other_table` ON `my_table`.`foo` = `my_other_table`.`foo`
"""

expression = parse_one(query, read="spark")
print(normalize(expression).sql())
```

```sql
SELECT "bar" FROM "my_table" LEFT ANTI JOIN "my_other_table" ON "my_table"."foo" = "my_other_table"."foo"
```


sqlglot/optimizer/optimize_joins.py
```diff
@@ -79,7 +79,8 @@ def normalize(expression):
         if join.kind == "CROSS":
             join.set("on", None)
         else:
-            join.set("kind", None)
+            if join.kind in ("INNER", "OUTER"):
+                join.set("kind", None)
```
VaggelisD, line 83: We can make this an `elif` to avoid the extra indent

themattmorris: I initially was going to do that, but we would lose this statement:

https://github.com/themattmorris/sqlglot/blob/f1ef49be623a6db62839fc3caa1015b82809dcc5/sqlglot/optimizer/optimize_joins.py#L85-L86

Should that statement not also be executed when the kind is anti|semi|etc?

VaggelisD: Oh whoops, completely missed that it was under this else branch, what you have is fine, thanks!

PR #5480, opened 2025-07-25
Fix(bigquery)!: store Query schemas in meta dict instead of type attr

georgesittas:
This PR aims to address the following bug, introduced [here](https://github.com/tobymao/sqlglot/pull/5230):

```python
>>> from sqlglot.optimizer.annotate_types import annotate_types
>>> from sqlglot.optimizer.qualify import qualify
>>>
>>> ast = parse_one("SELECT ARRAY(SELECT 'foo')", "bigquery")
>>> qualified = qualify(ast, dialect="bigquery")
>>> annotated = annotate_types(qualified, dialect="bigquery")
>>>
>>> annotated.selects[0].type
DataType(
  this=Type.ARRAY,
  expressions=[
    DataType(
      this=Type.STRUCT,
      expressions=[
        ColumnDef(
          this=Identifier(this=foo, quoted=False),
          kind=DataType(this=Type.VARCHAR))],
      nested=True)],
  nested=True)
>>> annotated.selects[0].type.sql("bigquery")
'ARRAY<STRUCT<foo STRING>>'
```

Note that BigQuery behaves differently:

```sql
SELECT ARRAY(SELECT 'foo'), TYPEOF(ARRAY(SELECT 'foo')) -- foo, ARRAY<STRING>
```

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -365,6 +365,31 @@ def _annotate_concat(self: TypeAnnotator, expression: exp.Concat) -> exp.Concat:
     return annotated
 
 
+def _annotate_array(self: TypeAnnotator, expression: exp.Array) -> exp.Array:
+    array_args = expression.expressions
+
+    # BigQuery behaves as follows:
+    #
+    # SELECT t, TYPEOF(t) FROM (SELECT 'foo') AS t            -- foo, STRUCT<STRING>
+    # SELECT ARRAY(SELECT 'foo'), TYPEOF(ARRAY(SELECT 'foo')) -- foo, ARRAY<STRING>
+    if (
+        len(array_args) == 1
+        and isinstance(select := array_args[0].unnest(), exp.Select)
+        and (query_type := select.meta.get("query_type")) is not None
+        and query_type.is_type(exp.DataType.Type.STRUCT)
+        and len(query_type.expressions) == 1
+    ):
+        projection_type = query_type.expressions[0].kind.copy()
+        array_type = exp.DataType(
+            this=exp.DataType.Type.ARRAY,
+            expressions=[projection_type],
+            nested=True,
+        )
+        return self._annotate_with_type(expression, array_type)
+
+    return self._annotate_by_args(expression, "expressions", array=True)
+
+
 class BigQuery(Dialect):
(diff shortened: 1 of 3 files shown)


sqlglot/optimizer/annotate_types.py
```diff
@@ -335,7 +337,10 @@ def annotate_scope(self, scope: Scope) -> None:
                 for cd in struct_type.expressions
                 if cd.kind
             ):
-                self._set_type(scope.expression, struct_type)
+                # We don't use `_set_type` on purpose here. If we annotated the query directly, then
+                # using it in other contexts (e.g., ARRAY(<query>)) could result in incorrect type
+                # annotations, i.e., it shouldn't be interpreted as a STRUCT value.
+                scope.expression.meta["query_type"] = struct_type
```
georgesittas, line 343: I only did `_set_type` to ensure it's available at L320. It should be ok to make this section a bit more conservative and avoid other similar bugs.

sqlglot/dialects/bigquery.py
```diff
@@ -365,6 +365,31 @@ def _annotate_concat(self: TypeAnnotator, expression: exp.Concat) -> exp.Concat:
...
 
+def _annotate_array(self: TypeAnnotator, expression: exp.Array) -> exp.Array:
+    array_args = expression.expressions
+
+    # BigQuery behaves as follows:
+    #
+    # SELECT t, TYPEOF(t) FROM (SELECT 'foo') AS t            -- foo, STRUCT<STRING>
+    # SELECT ARRAY(SELECT 'foo'), TYPEOF(ARRAY(SELECT 'foo')) -- foo, ARRAY<STRING>
+    if (
+        len(array_args) == 1
```
tobymao, line 376: why would array_args not be len 1?

tobymao (approved):

sqlglot/dialects/bigquery.py
georgesittas: This can happen for array literals:

```python
>>> import sqlglot
>>> sqlglot.parse_one("[1,2,3]", "bigquery")
Array(
  expressions=[
    Literal(this=2, is_string=False),
    Literal(this=3, is_string=False)])
```

PR #5487, opened 2025-07-28
feat(exasol): add support for REGEXP_SUBSTR  in exasol dialect

nnamdi16:
This involves adding [REGEXP_SUBSTR](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/regexp_substr.htm) built -in function to exasol dialect in sqlglot.

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -11,12 +11,28 @@ from sqlglot.dialects.dialect import (
 from sqlglot.helper import seq_get
 from sqlglot.generator import unsupported_args
 from sqlglot.tokens import TokenType
+import typing as t
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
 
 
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/exasol.py
```diff
@@ -69,6 +85,7 @@ class Parser(parser.Parser):
             "HASH_SHA1": exp.SHA.from_arg_list,
             "HASH_MD5": exp.MD5.from_arg_list,
             "HASHTYPE_MD5": exp.MD5Digest.from_arg_list,
+            "REGEXP_SUBSTR": _build_regexp_extract(exp.RegexpExtract),
```
VaggelisD: You can omit the builder and instead do `exp.RegexpExtract.from_arg_list` here since (1) the order of the args matches the order in `arg_types` too and (2) there aren't any default values.

PR #5490, opened 2025-07-28
feat(exasol): Add support for TRUNC, TRUNCATE and DATE_TRUNC function…

nnamdi16:
This involves adding [TRUNC(datetime)](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/trunc%5Bate%5D%20(datetime).htm), [TRUNC(number)](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/trunc%5Bate%5D%20(number).htm) and [DATE_TRUNC](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/date_trunc.htm#DATE_TRUNC) exasol function in exasol dialect…s in Exasol dialect

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from sqlglot import exp, generator, parser, tokens
+from sqlglot.dialects.clickhouse import timestamptrunc_sql
 from sqlglot.dialects.dialect import (
     Dialect,
     rename_func,
@@ -7,16 +8,25 @@ from sqlglot.dialects.dialect import (
     build_formatted_time,
     timestrtotime_sql,
     strposition_sql,
+    unit_to_str,
 )
 from sqlglot.helper import seq_get
 from sqlglot.generator import unsupported_args
 from sqlglot.tokens import TokenType
+import typing as t
 
 
 def _sha2_sql(self: Exasol.Generator, expression: exp.SHA2) -> str:
     length = expression.text("length")
     func_name = "HASH_SHA256" if length == "256" else "HASH_SHA512"
     return self.func(func_name, expression.this)
+
+
+def _parse_trunc(args: t.List) -> exp.DateTrunc | exp.Anonymous:
+    if args and isinstance(args[0], exp.Literal) and args[0].is_number:
+        return exp.Anonymous(this="TRUNC", expressions=args)
+
+    return exp.DateTrunc(unit=seq_get(args, 1), this=seq_get(args, 0))
 
 
 class Exasol(Dialect):
@@ -63,6 +73,9 @@ class Exasol(Dialect):
             "BIT_NOT": lambda args: exp.BitwiseNot(this=seq_get(args, 0)),
             "BIT_LSHIFT": binary_from_function(exp.BitwiseLeftShift),
             "BIT_RSHIFT": binary_from_function(exp.BitwiseRightShift),
+            "DATE_TRUNC": lambda args: exp.TimestampTrunc(
+                this=seq_get(args, 1), unit=exp.var(seq_get(args, 0))
+            ),
             "EVERY": lambda args: exp.All(this=seq_get(args, 0)),
             "EDIT_DISTANCE": exp.Levenshtein.from_arg_list,
             "HASH_SHA": exp.SHA.from_arg_list,
@@ -83,6 +96,8 @@ class Exasol(Dialect):
             "HASH_SHA512": lambda args: exp.SHA2(
                 this=seq_get(args, 0), length=exp.Literal.number(512)
             ),
+            "TRUNC": _parse_trunc,
+            "TRUNCATE": _parse_trunc,
             "VAR_POP": exp.VariancePop.from_arg_list,
             "APPROXIMATE_COUNT_DISTINCT": exp.ApproxDistinct.from_arg_list,
             "TO_CHAR": build_formatted_time(exp.ToChar, "exasol"),
@@ -156,6 +171,8 @@ class Exasol(Dialect):
             exp.BitwiseXor: rename_func("BIT_XOR"),
             # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/every.htm
             exp.All: rename_func("EVERY"),
+            exp.DateTrunc: lambda self, e: self.func("TRUNC", e.this, unit_to_str(e)),
+            exp.DatetimeTrunc: timestamptrunc_sql(),
             # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/edit_distance.htm#EDIT_DISTANCE
             exp.Levenshtein: unsupported_args("ins_cost", "del_cost", "sub_cost", "max_dist")(
                 rename_func("EDIT_DISTANCE")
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/exasol.py
```diff
@@ -19,6 +22,15 @@ def _sha2_sql(self: Exasol.Generator, expression: exp.SHA2) -> str:
     return self.func(func_name, expression.this)
 
 
+# https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/trunc%5Bate%5D%20(datetime).htm
+# https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/trunc%5Bate%5D%20(number).htm
+def _parse_trunc(args: t.List) -> exp.DateTrunc | exp.Anonymous:
+    if args and isinstance(args[0], exp.Literal) and args[0].is_number:
+        return exp.Anonymous(this="TRUNC", expressions=args)
+
+    return exp.DateTrunc(unit=seq_get(args, 1), this=seq_get(args, 0))
```
georgesittas: This doesn't look right. If we can't *statically* infer the semantics of the `TRUNC[ATE]` function, we should conservatively default to `Anonymous`, not the other way around.

sqlglot/dialects/exasol.py
```diff
@@ -63,6 +75,10 @@ class Parser(parser.Parser):
             "BIT_NOT": lambda args: exp.BitwiseNot(this=seq_get(args, 0)),
             "BIT_LSHIFT": binary_from_function(exp.BitwiseLeftShift),
             "BIT_RSHIFT": binary_from_function(exp.BitwiseRightShift),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/date_trunc.htm#DATE_TRUNC
+            "DATE_TRUNC": lambda args: exp.TimestampTrunc(
+                this=seq_get(args, 1), unit=exp.var(seq_get(args, 0))
+            ),
```
georgesittas: I don't think we need to use `exp.var` here, this happens automatically because `TimestampTrunc` inherits from `TimeUnit`.

```suggestion
            "DATE_TRUNC": lambda args: exp.TimestampTrunc(
                this=seq_get(args, 1), unit=seq_get(args, 0)
            ),
```

sqlglot/dialects/exasol.py
```diff
@@ -19,6 +22,19 @@ def _sha2_sql(self: Exasol.Generator, expression: exp.SHA2) -> str:
...
 
+# https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/trunc%5Bate%5D%20(datetime).htm
+# https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/trunc%5Bate%5D%20(number).htm
+def _parse_trunc(args: t.List) -> exp.Expression:
+    if len(args) == 2:
+        first = args[0]
+        if isinstance(first, (exp.Column, exp.Date, exp.Timestamp, exp.Cast)) or (
+            isinstance(first, exp.Literal) and first.is_string
+        ):
+            return exp.DateTrunc(this=first, unit=args[1])
```
georgesittas: Can you elaborate a bit on the reasoning behind this branch?

nnamdi16: I want to ensure that when TRUNC is been used as a date/time that exp.DateTrunc is used.
For instance, `TRUNC(DATE '2006-12-31', 'MM'), TRUNC(TIMESTAMP '2006-12-31 23:59:59', 'MI'), TRUNC(CAST('2006-12-31' AS DATE), 'MM')`

georgesittas: I see.. this implementation isn't quite right. For example, you emit a `DateTrunc` when the input is a `Column`, but when this happens you can't really tell whether `TRUNC`'s input is a date/time or a number, right?

You can only disambiguate the two when the 2nd argument is an integer, e.g.,

```sql
SELECT TRUNC(123.456,2);
```

(I'd double-check whether `SELECT TRUNC(123.456,'2');` produces the same result as well)

Check out how we deal with type-sensitive functions in other dialects. There's a pattern like:

```python

if not expression.type:

    expression = annotate_types(expression, dialect=self.dialect)

if expression.is_type(...):
    X
elif expression.is_type(...):
    Y
...
```

I think we could perhaps do the same here and consider the integer and date/time cases separately. So this builder needs to guarantee that if a non-`Anonymous` node is returned, that must unambiguously represent the input.

nnamdi16: Alright I would look into it

nnamdi16: @georgesittas I have updated the PR to reflect the suggestion. I also added back the test I did override mistakenly.

sqlglot/dialects/exasol.py
```diff
@@ -19,6 +23,27 @@ def _sha2_sql(self: Exasol.Generator, expression: exp.SHA2) -> str:
...
+    if not first or not second:
+        return exp.Anonymous(this="TRUNC", expressions=args)
+
+    if not first.type:
+        first = annotate_types(first, dialect="exasol")
+
+    if (
+        first.is_type(exp.DataType.Type.DATE, exp.DataType.Type.TIMESTAMP)
+        and isinstance(second, exp.Literal)
+        and second.is_string
```
VaggelisD: `is_string` already contains the `exp.Literal` instance check so we can remove that:

```Python3
# expressions.py

    @property
    def is_string(self) -> bool:
        """
        Checks whether a Literal expression is a string.
        """
        return isinstance(self, Literal) and self.args["is_string"]
```

sqlglot/dialects/exasol.py
```diff
@@ -83,6 +112,8 @@ class Parser(parser.Parser):
             "HASH_SHA512": lambda args: exp.SHA2(
                 this=seq_get(args, 0), length=exp.Literal.number(512)
             ),
+            "TRUNC": _parse_trunc,
+            "TRUNCATE": _parse_trunc,
```
VaggelisD: [Nit] We call these builder functions, e.g we'd name this`_build_trunc`

nnamdi16: Alright

sqlglot/dialects/exasol.py
nnamdi16: Alright

sqlglot/dialects/exasol.py
```diff
@@ -19,6 +23,25 @@ def _sha2_sql(self: Exasol.Generator, expression: exp.SHA2) -> str:
...
+
+    if not first or not second:
+        return exp.Anonymous(this="TRUNC", expressions=args)
+
+    if not first.type:
+        first = annotate_types(first, dialect="exasol")
+
+    if first.is_type(exp.DataType.Type.DATE, exp.DataType.Type.TIMESTAMP) and isinstance(
+        second, exp.Literal
+    ):
```
VaggelisD: Sorry for the confusion, I meant that we could turn this into:


```suggestion
    if first.is_type(exp.DataType.Type.DATE, exp.DataType.Type.TIMESTAMP) and second.is_string
```

Because `is_string` already contains `isinstance(self, exp.Literal)`

georgesittas:
@nnamdi16 I'll get this in and try to clean it up a bit. My hunch is that it can be probably improved more.

PR #5492, opened 2025-07-28
feat(singlestore): Added Tokenizer

AdalbertMemSQL:
Added several new tokens.
`:>` and `:!>` are used in SingleStore as cast operators.
`::$` and `::%` are used as JSON extract operators.

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -34,7 +34,7 @@ class SingleStore(MySQL):
         KEYWORDS = {
             **MySQL.Tokenizer.KEYWORDS,
             "BSON": TokenType.JSONB,
-            "GEOGRAPHYPOINT": TokenType.GEOGRAPHY,
+            "GEOGRAPHYPOINT": TokenType.GEOGRAPHYPOINT,
             ":>": TokenType.COLON_GT,
             "!:>": TokenType.NCOLON_GT,
             "::$": TokenType.DCOLONDOLLAR,
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -4546,6 +4546,7 @@ class DataType(Expression):
         FIXEDSTRING = auto()
         FLOAT = auto()
         GEOGRAPHY = auto()
+        GEOGRAPHYPOINT = auto()
         GEOMETRY = auto()
         POINT = auto()
         RING = auto()
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -380,6 +380,7 @@ class Parser(metaclass=_Parser):
         TokenType.BIGDECIMAL,
         TokenType.UUID,
         TokenType.GEOGRAPHY,
+        TokenType.GEOGRAPHYPOINT,
         TokenType.GEOMETRY,
         TokenType.POINT,
         TokenType.RING,
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -187,6 +187,7 @@ class TokenType(AutoName):
     DATEMULTIRANGE = auto()
     UUID = auto()
     GEOGRAPHY = auto()
+    GEOGRAPHYPOINT = auto()
     NULLABLE = auto()
     GEOMETRY = auto()
     POINT = auto()


AdalbertMemSQL:
Hey @VaggelisD 
Regarding docs:
 * For `BYTE_STRINGS` I didn't find information in SingleStore documentation, but I tested e'' syntax and it works 
```
singlestore> select e'text'; 
+------+
| text |
+------+
| text |
+------+
1 row in set (0.05 sec)
```
 * [BSON](https://docs.singlestore.com/cloud/reference/sql-reference/data-types/bson-type/)
 * [GEOGRAPHYPOINT](https://docs.singlestore.com/cloud/developer-resources/functional-extensions/working-with-geospatial-features/#creating-a-geospatial-table)
 * [Cast operators](https://docs.singlestore.com/cloud/reference/sql-reference/conditional-functions/cast-or-convert/#cast-operators-and)
 * [JSON operators](https://docs.singlestore.com/cloud/create-a-database/using-json/#example-1-access-fields-of-a-json-document)

tests/dialects/test_singlestore.py
```diff
@@ -18,3 +19,29 @@ def test_basic(self):
...
+        assert any(
+            token.token_type == TokenType.NCOLON_GT for token in SingleStore().tokenize(sql="!:>")
+        )
+        assert any(
+            token.token_type == TokenType.DCOLONDOLLAR
+            for token in SingleStore().tokenize(sql="::$")
+        )
+        assert any(
+            token.token_type == TokenType.DCOLONPERCENT
+            for token in SingleStore().tokenize(sql="::%")
```
VaggelisD: Sorry for the confusion, I was mostly referring to testing actual queries using these tokens with `validate_identity` as you have done with `self.validate_identity("SELECT e'text'")`

AdalbertMemSQL: Sounds good, will do it

AdalbertMemSQL: To properly parse/generate operators, more changes will be needed in the Parser/Generator. Is it ok to add them in this PR?

VaggelisD: What are the changes that will be required? We only need to add a few tests like `self.validate_identity("a ::$ b")`, but in any case, if these additions are non-trivial we can postpone them for a follow up commit. 

i don't think want these token type checks for now, we don't really do that for other dialects so we can revert these for the time being.

AdalbertMemSQL: At the moment, `self.validate_identity("a ::$ b")` will return the following error
```
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 5.
  a ::$ b
```
To make it work properly, parser and generator needs to be changed
```
        COLUMN_OPERATORS = {
            **parser.Parser.COLUMN_OPERATORS,
            TokenType.COLON_GT: lambda self, this, to: self.expression(
                exp.Cast,
                this=this,
                to=to,
            ),
            TokenType.NCOLON_GT: lambda self, this, to: self.expression(
                exp.TryCast,
                this=this,
                to=to,
            ),
```
```
        @unsupported_args("format")
        @unsupported_args("action")
        @unsupported_args("default")
        def cast_sql(self, expression: exp.Cast) -> str:  # type: ignore
            return f"{self.sql(expression, 'this')} :> {self.sql(expression, 'to')}"
```
I prefer to postpone these changes to follow up commits

VaggelisD: Sounds good, let's postpone these then 👍

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,19 @@
...
 class SingleStore(MySQL):
     SUPPORTS_ORDER_BY_ALL = True
+
+    class Tokenizer(MySQL.Tokenizer):
+        BYTE_STRINGS = [("e'", "'"), ("E'", "'")]
+
+        KEYWORDS = {
+            **MySQL.Tokenizer.KEYWORDS,
+            "BSON": TokenType.JSONB,
+            "GEOGRAPHYPOINT": TokenType.GEOGRAPHY,
```
georgesittas: I suggest we create a dedicated `GEOGRAPHYPOINT` data type, given its semantics appear to be different from `GEOGRAPHY` in Singlestore ([docs](https://docs.singlestore.com/cloud/developer-resources/functional-extensions/working-with-geospatial-features/#creating-a-geospatial-table)):

> Geospatial objects are just another data type in SingleStore and can be used right alongside numbers, strings, and JSON. To create a column that holds point data, use the GeographyPoint type. For path and polygon data, use the Geography type.

We need to add that to the `DataType.TYPE` enum, as well as `TYPE_TOKENS` in the parser, etc.

sqlglot/dialects/singlestore.py
```diff
@@ -1,5 +1,19 @@
...
+            ":>": TokenType.COLON_GT,
+            "!:>": TokenType.NCOLON_GT,
+            "::$": TokenType.DCOLONDOLLAR,
+            "::%": TokenType.DCOLONPERCENT,
```
georgesittas, line 41: I'm assuming you're planning to add parsing/generation support in a followup PR for these?

georgesittas (approved):
Looks good @AdalbertMemSQL, just one comment on my end. Thanks!

Let's also resolve the conflicts, similarly to the other PR.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: Yep

sqlglot/dialects/singlestore.py
AdalbertMemSQL: This is a good idea. Added suggested changes.

Issue #5484, opened 2025-07-27
Support for quantified like operator (BigQuery)


mazum21:
**Is your feature request related to a problem? Please describe.**
BigQuery has [Quantified LIKE operators](https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#like_operator_quantified), which don't seem to be correctly transpiled to e.g. DuckDB dialect.

**Describe the solution you'd like**
```sql
'foo' LIKE ANY ('%bar%, %baz%)
```
should be transpiled to
```sql
('foo' LIKE '%bar%
OR 'foo' like %baz%)
```

**Describe alternatives you've considered**
error message which indicates that Quantified LIKE operators are not supported

**Additional context**
n/a

PR #5493, opened 2025-07-28
feat!: Refactor `LIKE ANY` and support  `ALL | SOME` quantifiers

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5484

This PR:

- Refactors `LIKE ANY` parsing by removing the tokens + expressions that were [introduced in SQLGlot](https://github.com/tobymao/sqlglot/pull/1183)
- Parses `LIKE SOME (...)` and `LIKE ALL (...)` similar to subqueries
- Adds a new generator flag that reduces the quantifiers into conjuctions (for `LIKE ALL`) or disjunctions (for `LIKE ANY | SOME`)


sqlglot/generator.py
```diff
@@ -2898,7 +2901,10 @@ def bracket_sql(self, expression: exp.Bracket) -> str:
         return f"{self.sql(expression, 'this')}[{expressions_sql}]"
 
     def all_sql(self, expression: exp.All) -> str:
-        return f"ALL {self.wrap(expression)}"
+        this = self.sql(expression, "this")
+        if not isinstance(expression.this, (exp.Tuple, exp.Paren)):
```
georgesittas, line 2905: Why is this different from `any_sql`?

sqlglot/generator.py
```diff
@@ -3648,24 +3654,42 @@ def gt_sql(self, expression: exp.GT) -> str:
...
+
+        if isinstance(expression, exp.Like):
+            exp_class: t.Type[exp.Like | exp.ILike] = exp.Like
+            op = "LIKE"
+        else:
+            exp_class = exp.ILike
+            op = "ILIKE"
+
+        if isinstance(rhs, (exp.All, exp.Any)) and not self.SUPPORTS_LIKE_QUANTIFIERS:
+            exprs = [rhs.this.this] if isinstance(rhs.this, exp.Paren) else rhs.this.expressions
```
georgesittas: What if there are multiple levels of parentheses? Should we `unnest()`instead of accessing `this`?

sqlglot/generator.py
```diff
@@ -3648,24 +3654,42 @@ def gt_sql(self, expression: exp.GT) -> str:
...
+            func = exp.or_ if isinstance(rhs, exp.Any) else exp.and_
```
georgesittas: Nit: can we rename `func` to `connective` so that it's clearer what we're applying in L3682?

sqlglot/generator.py
```diff
@@ -3648,24 +3654,42 @@ def gt_sql(self, expression: exp.GT) -> str:
...
+
+            like_sql: exp.Expression = exp_class(this=this, expression=exprs[0])
+
+            for expr in exprs[1:]:
+                like_sql = func(like_sql, exp_class(this=this, expression=expr))
```
georgesittas: We can also `reduce` here:
```suggestion
            like_expr = reduce(lambda x, y: func(exp_class(this=x, expression=y)), [this, *exprs])
```

sqlglot/parser.py
```diff
@@ -5805,12 +5806,17 @@ def _parse_function_call(
...
-                self._match_r_paren()
-                return this
+            if subquery_predicate:
+                expr = None
+                if self._curr.token_type in (TokenType.SELECT, TokenType.WITH):
+                    expr = self._parse_select()
+                    self._match_r_paren()
+                elif prev and prev.token_type in (TokenType.LIKE, TokenType.ILIKE):
+                    self._advance(-1)
+                    expr = self._parse_bitwise()
```
georgesittas: Can we leave a comment explaining why we backtrack one token here?

sqlglot/generator.py
VaggelisD: Afaict this is because `exp.Any` is also a no paren function which `exp.All` isn't, so unifying their generations can cause issues.

georgesittas: Fair enough, ok. Should we use the `WRAPPED_QUERIES` constant here?

sqlglot/generator.py
VaggelisD: This isn't really matching what I generate, e.g this would transpile:

```
>>> sqlglot.parse_one("SELECT col LIKE ANY (a, b, c)", "snowflake").sql("duckdb")
SELECT col LIKE a LIKE b LIKE c
```

sqlglot/generator.py
VaggelisD: It wouldn't cover all cases, e.g this is a valid DuckDB query:

```SQL
SELECT ALL (col > value) FROM tbl
```

But that is parsed as `exp.All(exp.GT(...))` so in this case we'd omit the parentheses since `UNWRAPPED_QUERIES` is only `(Select, SetOperation)`

georgesittas: The case I'm interested in is `ALL(<query>)`, I think in the past there was an issue with either emitting a redundant set of parentheses for `ANY` or omitting them altogether. If this is covered by the existing tests, it should be fine, otherwise let's test it.

sqlglot/generator.py
georgesittas: Whoops, right, my bad. Let's leave it.

georgesittas (approved):

Issue #5495, opened 2025-07-28
Support GET from Snowflake to DuckDB


kyle-cheung:
- Read Snowflake
- Write DuckDB

**Fully reproducible code snippet**
Create a DuckDB database with sample data:
```
CREATE OR REPLACE TEMPORARY TABLE get_path_demo(
  id INTEGER,
  v  JSON);

INSERT INTO get_path_demo (id, v)
SELECT 1,
        JSON('{
          "array1" : [
            {"id1": "value_a1", "id2": "value_a2", "id3": "value_a3"}
          ],
          "array2" : [
            {"id1": "value_b1", "id2": "value_b2", "id3": "value_b3"}
          ],
          "object_outer_key1" : {
            "object_inner_key1a": "object_x1",
            "object_inner_key1b": "object_x2"
          }
        }');

INSERT INTO get_path_demo (id, v)
SELECT 2,
        JSON('{
          "array1" : [
            {"id1": "value_c1", "id2": "value_c2", "id3": "value_c3"}
          ],
          "array2" : [
            {"id1": "value_d1", "id2": "value_d2", "id3": "value_d3"}
          ],
          "object_outer_key1" : {
            "object_inner_key1a": "object_y1",
            "object_inner_key1b": "object_y2"
          }
        }');
```

<img width="1494" height="129" alt="Image" src="https://github.com/user-attachments/assets/85efc82c-f214-4580-9b0a-fc733a5d69a2" />

```
import sqlglot
from sqlglot.optimizer import optimize

sql = """
SELECT
  GET(v, 'array1.id1'),
  GET(v, 'array1'),
  GET_PATH(v, 'array1.id1'),
FROM get_path_demo
"""

transpiled_sql = sqlglot.transpile(sql, "snowflake", "duckdb")[0]
print (transpiled_sql)
```
This returns
```
SELECT GET(v, 'array1.id1'), GET(v, 'array1'), v -> '$.array1.id1' FROM get_path_test
```
--> GET function does not exist in DuckDB

This should instead return
```
SELECT v.array.id1, v.array1, v -> '$.array1.id1' FROM get_path_test
```
Or use -> notation

**Official Documentation**
https://duckdb.org/docs/stable/data/json/overview.html#retrieving-json-data
https://docs.snowflake.com/en/sql-reference/functions/get_path
https://docs.snowflake.com/en/sql-reference/functions/get

PR #5500, opened 2025-07-29
feat(snowflake)!: Parse and transpile GET(...) extract function

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5495

This PR adds support for Snowflake's `GET(...)` function which is used to extract items/fields from different types. Also, a best-effort transpilation attempt is made (mostly towards DuckDB). 

- Note 1: This is not a "superset" of `GET_PATH` e.g it doesn't work correctly for any JSON path:

```SQL
snowflake> WITH tbl AS (
    SELECT PARSE_JSON('{
               "array1" : [
                 {"id1": "value_c1", "id2": "value_c2", "id3": "value_c3"}
               ],
               "array2" : [
                 {"id1": "value_d1", "id2": "value_d2", "id3": "value_d3"}
               ],
               "object_outer_key1" : {
                 "object_inner_key1a": "object_y1",
                 "object_inner_key1b": "object_y2"
               }
             }') as col
)
SELECT GET(col,'array1.id1'), GET_PATH(col,'array1[0].id1') FROM tbl; 

--

NULL | "value_c1
```

- Note 2: Snowflake's index offset is 0-based while DuckDB's is 1, thus the index shift in `validate_all`

```SQL
snowflake> SELECT GET([1, 2], 1);
2

duckdb> SELECT [1, 2][1];
┌──────────────────────────┐
│ main.list_value(1, 2)[1] │
│          int32           │
├──────────────────────────┤
│            1             │
└──────────────────────────┘
```

- Note 3: This should not be confused with Snowflake's `GET` statement for copying/writing files (`exp.Get`).

Docs
---------
[Snowflake GET function](https://docs.snowflake.com/en/sql-reference/functions/get) | [Snowflake GET_PATH](https://docs.snowflake.com/en/sql-reference/functions/get_path) | [Snowflake GET statement](https://docs.snowflake.com/en/sql-reference/sql/get)

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -5116,17 +5116,14 @@ class Generator(metaclass=_Generator):
 
     def getextract_sql(self, expression: exp.GetExtract) -> str:
         this = expression.this
-        expression = expression.expression
+        expr = expression.expression
 
         if not this.type or not expression.type:
             from sqlglot.optimizer.annotate_types import annotate_types
 
             this = annotate_types(this, dialect=self.dialect)
-            expression = annotate_types(expression, dialect=self.dialect)
 
-        if this.is_type(exp.DataType.Type.ARRAY) or expression.is_type(*exp.DataType.NUMERIC_TYPES):
-            return self.sql(exp.Bracket(this=this, expressions=[expression]))
+        if this.is_type(*(exp.DataType.Type.ARRAY, exp.DataType.Type.MAP)):
+            return self.sql(exp.Bracket(this=this, expressions=[expr]))
 
-        return self.sql(
-            exp.JSONExtract(this=this, expression=self.dialect.to_json_path(expression))
-        )
+        return self.sql(exp.JSONExtract(this=this, expression=self.dialect.to_json_path(expr)))
--- a/tests/dialects/test_snowflake.py
+++ b/tests/dialects/test_snowflake.py
@@ -2952,17 +2952,21 @@ FROM SEMANTIC_VIEW(
         )
 
     def test_get_extract(self):
-        for arr, index in (("arr", 1), ("[1, 2]", "index")):
-            # Test that either an ARRAY literal or a numeric index is transpiled to an array access,
-            # even if the other arg is of unknown type
-            with self.subTest(f"Testing GET({arr}, {index})"):
-                self.validate_all(
-                    f"SELECT GET({arr}, {index})",
-                    write={
-                        "snowflake": f"SELECT GET({arr}, {index})",
-                        "duckdb": f"SELECT {arr}[{index + 1 if isinstance(index, int) else index}]",
-                    },
-                )
+        self.validate_all(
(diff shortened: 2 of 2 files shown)


sqlglot/generator.py
```diff
@@ -5113,3 +5113,20 @@ def semanticview_sql(self, expression: exp.SemanticView) -> str:
...
+        expression = expression.expression
+
+        if not this.type or not expression.type:
+            from sqlglot.optimizer.annotate_types import annotate_types
+
+            this = annotate_types(this, dialect=self.dialect)
+            expression = annotate_types(expression, dialect=self.dialect)
+
+        if this.is_type(exp.DataType.Type.ARRAY) or expression.is_type(*exp.DataType.NUMERIC_TYPES):
+            return self.sql(exp.Bracket(this=this, expressions=[expression]))
```
georgesittas: Can you elaborate about this condition? Why do we check for numeric lookup keys besides just arrays?

VaggelisD: This would be for these 2 variants:

```SQL
GET( <array> , <index> )

GET( <variant> , <index> )
```

which refer to array accesses

georgesittas: Can you have a numeric lookup key for maps? Or can we unambiguously determine the first argument is an array?

VaggelisD: I initially thought we couldn't but I checked again and it turns out we can, for whatever reason they must be **defined** as strings but can be **queried** as integers, e.g:

- This doesn't work 
```SQL
snowflake> WITH tbl AS (SELECT {1: 'San Mateo', 2: 'CA'}::MAP(INTEGER, VARCHAR) AS col) SELECT GET(col, 1) FROM tbl;
Syntax error: unexpected '1'. (line 1)
```

- This works
```SQL
snowflake> WITH tbl AS (SELECT {'1': 'San Mateo', '2': 'CA'}::MAP(INTEGER, VARCHAR) AS col) SELECT GET(col, 1) FROM tbl;
"San Mateo"
```
 
Although SQLGlot doesn't transpile this properly today, the `MAP` access in e.g DuckDB would still be a bracketed expression, albeit without the index offset:

```SQL
D WITH tbl AS (SELECT MAP {'1': 'San Mateo', '2': 'CA'} AS col) SELECT col[1] FROM tbl;
┌───────────┐
│  col[1]   │
│  varchar  │
├───────────┤
│ San Mateo │
└───────────┘
```

With that said, to transpile this to DuckDB, we should then:
- Generate the bracket expr `col[index + 1]` if the input is an `ARRAY`
- Generate the bracket expr `col[index]` if the input is an `MAP`
- Generate `exp.JSONExtract` otherwise to imitate VARIANT/OBJECT field access


Wdyt?

georgesittas: Sounds reasonable. Let's only control the emitted SQL based on the first argument's type, I don't think using the second one is a good signal.

PR #5502, opened 2025-07-30
feat(exasol): Add support for IF, NULLIFZERO, and ZEROIFNULL functions

nnamdi16:
This involves adding [NULLIFZERO](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/nullifzero.htm) and [ZEROIFNULL](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/zeroifnull.htm) exasol function in exasol dialect

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -91,6 +91,7 @@ class Exasol(Dialect):
         KEYWORDS = {
             **tokens.Tokenizer.KEYWORDS,
             "USER": TokenType.CURRENT_USER,
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/if.htm
             "ENDIF": TokenType.END,
         }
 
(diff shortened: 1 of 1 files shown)


sqlglot/dialects/exasol.py
```diff
@@ -136,6 +151,20 @@ class Parser(parser.Parser):
             ),
         }
 
+        # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/if.htm
+        def _parse_if(self) -> t.Optional[exp.Expression]:
+            if self._match_text_seq("IF"):
```
VaggelisD: Have you verified whether this branch is taken? I think `IF` should already be consumed already here (?)

nnamdi16: Hi @VaggelisD please can you clarify what you mean?

VaggelisD: Afaict `"IF"` has already been consumed going into this function so I think this condition will always be false, can you check that this is not the case?

nnamdi16: Alright I would check it

nnamdi16: Thanks for pointing it out @VaggelisD. I have made the correction

VaggelisD (approved):

PR #5503, opened 2025-07-30
Fix: dont print (A)SYMMETRIC keyword in BETWEEN for postgres subclasses

georgesittas:
Slack thread for reference: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1753876467945589


georgesittas:
FYI @mateuszpoleski

PR #5507, opened 2025-07-31
feat(exasol): add support for ADD_DAYS function in exasol dialect

nnamdi16:
This involves adding [ADD_DAYS](https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_days.htm) built -in function to exasol dialect in sqlglot.

### DIFF
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -100,12 +100,19 @@ class Exasol(Dialect):
     class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_days.htm
             "ADD_DAYS": build_date_delta(exp.DateAdd, default_unit="DAY"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_years.htm
             "ADD_YEARS": build_date_delta(exp.DateAdd, default_unit="YEAR"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_months.htm
             "ADD_MONTHS": build_date_delta(exp.DateAdd, default_unit="MONTH"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_weeks.htm
             "ADD_WEEKS": build_date_delta(exp.DateAdd, default_unit="WEEK"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_hour.htm
             "ADD_HOURS": build_date_delta(exp.DateAdd, default_unit="HOUR"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_minutes.htm
             "ADD_MINUTES": build_date_delta(exp.DateAdd, default_unit="MINUTE"),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_seconds.htm
             "ADD_SECONDS": build_date_delta(exp.DateAdd, default_unit="SECOND"),
             "BIT_AND": binary_from_function(exp.BitwiseAnd),
             "BIT_OR": binary_from_function(exp.BitwiseOr),
--- a/sqlglot/dialects/exasol.py
+++ b/sqlglot/dialects/exasol.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import typing as t
 
 from sqlglot import exp, generator, parser, tokens
-from sqlglot.dialects.clickhouse import build_date_delta, timestamptrunc_sql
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/exasol.py
```diff
@@ -3,7 +3,7 @@
 import typing as t
 
 from sqlglot import exp, generator, parser, tokens
-from sqlglot.dialects.clickhouse import timestamptrunc_sql
+from sqlglot.dialects.clickhouse import build_date_delta, timestamptrunc_sql
```
georgesittas: Hey, I just noticed that we're importing from clickhouse here. Can we please change this and import from `dialect` instead? I don't understand why this was the case to begin with. My guess is copy-paste?

sqlglot/dialects/exasol.py
```diff
@@ -209,6 +210,8 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             exp.All: rename_func("EVERY"),
             exp.DateTrunc: lambda self, e: self.func("TRUNC", e.this, unit_to_str(e)),
             exp.DatetimeTrunc: timestamptrunc_sql(),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_days.htm
+            exp.DateAdd: rename_func("ADD_DAYS"),
```
georgesittas: Why are we mapping `DateAdd` to `ADD_DAYS`? This doesn't seem right. Doesn't exasol support some kinda date adding function that allows other units like month, year etc? Even if not, I believe we should still look at the unit before generating `ADD_DAYS`.

sqlglot/dialects/exasol.py
nnamdi16: Yea that's a mistake would look into it. It was from an autocomplete from IDE

sqlglot/dialects/exasol.py
nnamdi16: Alright fixed @georgesittas

sqlglot/dialects/exasol.py
```diff
@@ -209,6 +224,8 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             exp.All: rename_func("EVERY"),
             exp.DateTrunc: lambda self, e: self.func("TRUNC", e.this, unit_to_str(e)),
             exp.DatetimeTrunc: timestamptrunc_sql(),
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/add_days.htm
+            # exp.DateAdd: dateadd_sql,
```
georgesittas:
```suggestion
```

sqlglot/dialects/exasol.py
```diff
@@ -273,3 +290,20 @@ def if_sql(self, expression: exp.If) -> str:
...
+
+            function_map = {
+                "DAY": "ADD_DAYS",
+                "WEEK": "ADD_WEEKS",
+                "MONTH": "ADD_MONTHS",
+                "YEAR": "ADD_YEARS",
+                "HOUR": "ADD_HOURS",
+                "MINUTE": "ADD_MINUTES",
+                "SECOND": "ADD_SECONDS",
+            }
```
georgesittas: This should be factored out as a class constant for the generator:

```python
DATE_ADD_FUNCTION_BY_UNIT = {
    "DAY": "ADD_DAYS",
    ...
}
```

sqlglot/dialects/exasol.py
```diff
@@ -273,3 +290,20 @@ def if_sql(self, expression: exp.If) -> str:
             true = self.sql(expression, "true")
             false = self.sql(expression, "false")
             return f"IF {this} THEN {true} ELSE {false} ENDIF"
+
+        def dateadd_sql(self, expression: exp.DateAdd) -> str:
+            unit = expression.args.get("unit")
+            unit_name = unit.name.upper() if unit else "DAY"
```
georgesittas:
```suggestion
            unit = expression.text("unit").upper() or "DAY"
```

sqlglot/dialects/exasol.py
```diff
@@ -273,3 +290,20 @@ def if_sql(self, expression: exp.If) -> str:
...
+                "DAY": "ADD_DAYS",
+                "WEEK": "ADD_WEEKS",
+                "MONTH": "ADD_MONTHS",
+                "YEAR": "ADD_YEARS",
+                "HOUR": "ADD_HOURS",
+                "MINUTE": "ADD_MINUTES",
+                "SECOND": "ADD_SECONDS",
+            }
+
+            func_name = function_map.get(unit_name, "ADD_DAYS")
```
georgesittas: Why are we not warning if there is no match in the mapping lookup? I think we may need to use `self.unsupported` in that case.

nnamdi16: Alright I would look into it

nnamdi16: @georgesittas thanks for pointing it out I have refactored it

georgesittas (approved):
Looks great, appreciate the effort 👍

PR #4959, opened 2025-04-08
feat(bigquery): Add support for side & kind on set operators

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/4942

This PR introduces support for the `BY NAME` & `CORRESPONDING` modifiers as are laid out in the docs; The latter is canonicalized into the former based on the facts that:

- `{SET_OP} STRICT CORRESPONDING [BY (col1, ...)]` == `{SET_OP} BY NAME [ON (col1, ...)]`
- `{SET_OP} CORRESPONDING` == `INNER {SET_OP} BY NAME`
- `{LEFT | FULL} [OUTER] {SET_OP} CORRESPONDING` == `{LEFT | FULL} [OUTER] {SET_OP} BY NAME` 

Docs
--------
[BQ Set Operators](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#set_operators) | [BQ BY NAME / CORRESPONDING modifiers](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#by_name_or_corresponding)

PR #5508, opened 2025-07-31
Fix!: Make `UNION` column qualification recursive

VaggelisD:
BigQuery allows set modifiers similar to JOINs (where empty string is `NULL`), e.g:

```SQL
bigquery> WITH 
 t1 AS (SELECT 1 AS a, 1 AS b),
 t2 AS (SELECT 1 AS b, 1 AS c),
 t3 AS (SELECT 1 AS c, 1 AS d),
 t4 AS (SELECT 1 AS e, 1 AS f)
SELECT * FROM (
  SELECT * FROM t1 FULL OUTER UNION ALL BY NAME 
  SELECT * FROM t2 FULL OUTER UNION ALL BY NAME 
  SELECT * FROM t3 FULL OUTER UNION ALL BY NAME
  SELECT * FROM t4
)

-------

a	b	c	d	e	f
1	1				
	1	1			
		1	1		
				1	1
```

<br/>

Support for these modifiers [was added recently](https://github.com/tobymao/sqlglot/pull/4959), but it turns out `qualify` doesn't deal with chained `UNION`s very well i.e it only looks at the left-most (`a`, `b`) & right-most (`e`, `f`) columns:

```Python3
>>> sql = "WITH t1 AS (SELECT 1 AS a, 1 AS b), t2 AS (SELECT 1 AS b, 1 AS c), t3 AS (SELECT 1 AS c, 1 AS d), t4 AS (SELECT 1 AS e, 1 AS f) SELECT * FROM (SELECT * FROM t1 FULL OUTER UNION ALL BY NAME SELECT * FROM t2 FULL OUTER UNION ALL BY NAME SELECT * FROM t3 FULL OUTER UNION ALL BY NAME SELECT * FROM t4)"
>>> qualify(sql, dialect="bigquery").sql("bigquery")
WITH ...
SELECT
  _q_0.a AS a,
  _q_0.b AS b,
  _q_0.e AS e,
  _q_0.f AS f
FROM (...) AS _q_0
```

<br/>

After this PR:

```Python3
>>> qualify(sql, dialect="bigquery").sql("bigquery")
WITH ...
SELECT
SELECT
  _q_0.a AS a,
  _q_0.b AS b,
  _q_0.c AS c,
  _q_0.d AS d,
  _q_0.e AS e,
  _q_0.f AS f
FROM (...) AS _q_0
```

### DIFF
--- a/sqlglot/optimizer/qualify_columns.py
+++ b/sqlglot/optimizer/qualify_columns.py
@@ -994,7 +994,7 @@ class Resolver:
             side = set_op.side
             kind = set_op.kind
 
-            # Visit the children UNIONs (if any) in a pre-order traversal
+            # Visit the children UNIONs (if any) in a post-order traversal
             left = self.get_source_columns_from_set_op(set_op.left)
             right = self.get_source_columns_from_set_op(set_op.right)
 


sqlglot/optimizer/qualify_columns.py
```diff
@@ -972,6 +972,44 @@ def all_columns(self) -> t.Set[str]:
...
+            # The resulting columns are the columns in the ON clause:
+            # {INNER | LEFT | FULL} UNION ALL BY NAME ON (col1, col2, ...)
+            columns = [col.name for col in on_column_list]
+        elif set_op.side or set_op.kind:
+            side = set_op.side
+            kind = set_op.kind
+
+            # Visit the children UNIONs (if any) in a pre-order traversal
+            left = self.get_source_columns_from_set_op(set_op.left)
+            right = self.get_source_columns_from_set_op(set_op.right)
```
georgesittas, line 999: I'm worried about the recursion here. Is this code path only reached when the special BigQuery syntax is used? Or can it be reached using the usual set operator syntax too?

Any chance we can make the whole thing iterative without much hassle?

georgesittas: Discussed in Slack, this path is only taken for set operations that are built off of BigQuery's special syntax, so the recursion should only concern those inputs, which is ok for now. We can re-evaluate if that's not the case.

georgesittas (approved):

sqlglot/optimizer/qualify_columns.py
VaggelisD: [Discussed internally](https://tobikodata.slack.com/archives/C040V7ZP9NH/p1753977721311449)

sqlglot/optimizer/qualify_columns.py
```diff
@@ -972,6 +972,44 @@ def all_columns(self) -> t.Set[str]:
...
         return self._all_columns
 
+    def get_source_columns_from_set_op(self, expression: exp.Expression) -> t.List[str]:
+        if isinstance(expression, exp.Select):
+            return expression.named_selects
+        if isinstance(expression, exp.Subquery) and isinstance(expression.this, exp.SetOperation):
+            # Different types of SET modifiers can be chained together if they're explicitly grouped by nesting
+            return self.get_source_columns_from_set_op(expression.this)
+        if not isinstance(expression, exp.SetOperation):
+            raise OptimizeError(f"Unknown set operation: {expression}")
```
georgesittas, line 982: Should we actually avoid failing here and just return `named_selects` to match the previous behavior?

VaggelisD: Yeah can do that 👍

VaggelisD: Ah sorry, the previous behavior didn't have `named_selects` as a fallback afaict, this was under the scope branch

georgesittas: I see, I misread that sorry.

PR #5513, opened 2025-08-01
feat (snowflake): add transpilation support for Snowflake's IDENTIFIER() function

rahulj51:
Adding support for correctly transpiling Snowflake's `IDENTIFIER()` to other dialects. 

  - Add IdentifierFunc expression class to handle Snowflake's IDENTIFIER() function
  - Implement transpilation logic to convert IDENTIFIER('literal') to plain identifiers for non-Snowflake dialects
  - Preserve IDENTIFIER() syntax when generating Snowflake SQL
  - Keep variables and complex expressions as IDENTIFIER() to maintain safety
  - Add comprehensive tests for various IDENTIFIER() use cases

BREAKING CHANGE: IDENTIFIER() function calls with string literals are now transpiled to plain identifiers when converting to other SQL dialects. Previously they remained as Anonymous function calls.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -95,6 +95,11 @@ def _build_object_construct(args: t.List) -> t.Union[exp.StarMap, exp.Struct]:
     )
 
 
+def _build_identifier_func(args: t.List) -> exp.IdentifierFunc:
+    """Build IDENTIFIER() function expression."""
+    return exp.IdentifierFunc(this=seq_get(args, 0))
+
+
 def _build_datediff(args: t.List) -> exp.DateDiff:
     return exp.DateDiff(
         this=seq_get(args, 2), expression=seq_get(args, 1), unit=map_date_part(seq_get(args, 0))
@@ -598,6 +603,7 @@ class Snowflake(Dialect):
                 requires_json=True,
             ),
             "HEX_DECODE_BINARY": exp.Unhex.from_arg_list,
+            "IDENTIFIER": _build_identifier_func,
             "IFF": exp.If.from_arg_list,
             "LAST_DAY": lambda args: exp.LastDay(
                 this=seq_get(args, 0), unit=map_date_part(seq_get(args, 1))
@@ -1553,6 +1559,11 @@ class Snowflake(Dialect):
                     unit=expression.unit,
                 )
             )
+
+        def identifierfunc_sql(self, expression: exp.IdentifierFunc) -> str:
+            """Generate SQL for Snowflake IDENTIFIER() function."""
+            # For Snowflake, always keep as IDENTIFIER() function
+            return self.function_fallback_sql(expression)
 
         def jsonextract_sql(self, expression: exp.JSONExtract):
             this = expression.this
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6234,6 +6234,20 @@ class Nullif(Func):
     arg_types = {"this": True, "expression": True}
 
 
+class IdentifierFunc(Func):
+    """
+    Snowflake IDENTIFIER() function for dynamic object name references.
+
+    Args:
+        this: The expression that resolves to the object name
+              (string literal, variable, or parameter)
+    """
+
+    arg_types = {"this": True}
+    _sql_names = ["IDENTIFIER"]
+    is_var = True  # Indicates this can be used as an identifier
+
+
 class Initcap(Func):
     arg_types = {"this": True, "expression": False}
 
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -154,6 +154,7 @@ class Generator(metaclass=_Generator):
         exp.GlobalProperty: lambda *_: "GLOBAL",
         exp.HeapProperty: lambda *_: "HEAP",
         exp.IcebergProperty: lambda *_: "ICEBERG",
+        exp.IdentifierFunc: lambda self, e: self.identifierfunc_sql(e),
         exp.InheritsProperty: lambda self, e: f"INHERITS ({self.expressions(e, flat=True)})",
         exp.InlineLengthColumnConstraint: lambda self, e: f"INLINE LENGTH {self.sql(e, 'this')}",
         exp.InputModelProperty: lambda self, e: f"INPUT{self.sql(e, 'this')}",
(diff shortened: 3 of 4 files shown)


sqlglot/dialects/snowflake.py
```diff
@@ -598,6 +603,7 @@ class Parser(parser.Parser):
                 requires_json=True,
             ),
             "HEX_DECODE_BINARY": exp.Unhex.from_arg_list,
+            "IDENTIFIER": _build_identifier_func,
```
VaggelisD, line 606: We don't really need a specialized builder for this, we can simply do:

```Python3
"IDENTIFIER": exp.IdentifierFunc.from_arg_list
```

sqlglot/dialects/snowflake.py
```diff
@@ -95,6 +95,11 @@ def _build_object_construct(args: t.List) -> t.Union[exp.StarMap, exp.Struct]:
     )
 
 
+def _build_identifier_func(args: t.List) -> exp.IdentifierFunc:
+    """Build IDENTIFIER() function expression."""
+    return exp.IdentifierFunc(this=seq_get(args, 0))
+
```
VaggelisD, line 101: Ditto, this can be removed

sqlglot/expressions.py
```diff
@@ -6234,6 +6234,20 @@ class Nullif(Func):
...
 
 
+class IdentifierFunc(Func):
+    """
+    Snowflake IDENTIFIER() function for dynamic object name references.
+
+    Args:
+        this: The expression that resolves to the object name
+              (string literal, variable, or parameter)
+    """
```
VaggelisD, line 6244: We usually leave a hyperlink to the docs, helps reduce the noise when there's hundreds of expressions

sqlglot/expressions.py
```diff
@@ -6234,6 +6234,20 @@ class Nullif(Func):
...
+
+    arg_types = {"this": True}
+    _sql_names = ["IDENTIFIER"]
+    is_var = True  # Indicates this can be used as an identifier
```
VaggelisD, line 6248: None of these are really needed:
- The `arg_types = {"this": True}` is inherited from `Expression`
- We tend use `_sql_names` only when an expression has slightly different names (or aliases) across dialects e.g  `ARGMAX`, `ARG_MAX` etc.
- The `is_var = True` looks like dead code

Instead, we can simply do:

```Python3
# https://docs.snowflake.com/en/sql-reference/identifier-literal
class IdentifierFunc(Func):
 pass
```

sqlglot/generator.py
```diff
@@ -154,6 +154,7 @@ class Generator(metaclass=_Generator):
         exp.GlobalProperty: lambda *_: "GLOBAL",
         exp.HeapProperty: lambda *_: "HEAP",
         exp.IcebergProperty: lambda *_: "ICEBERG",
+        exp.IdentifierFunc: lambda self, e: self.identifierfunc_sql(e),
```
VaggelisD, line 157: This is not required, SQLGlot will [automatically register this function](https://github.com/tobymao/sqlglot/blob/main/sqlglot/generator.py#L949) since it follows the pattern `<expr_name>_sql`

sqlglot/generator.py
```diff
@@ -1628,6 +1629,21 @@ def identifier_sql(self, expression: exp.Identifier) -> str:
...
+    def identifierfunc_sql(self, expression: exp.IdentifierFunc) -> str:
+        """Generate SQL for IDENTIFIER() function.
+
+        For dialects other than Snowflake, converts IDENTIFIER('literal') to a plain identifier.
+        For non-literals, falls back to function call syntax.
+        """
+        # For other dialects, convert IDENTIFIER('literal') to identifier if possible
+        if isinstance(expression.this, exp.Literal) and expression.this.is_string:
+            # expression.this is the Literal, expression.this.this is the string value
+            return self.sql(exp.to_identifier(expression.this.this))
```
VaggelisD, line 1641: 1. Let's cache `this = expression.this` so we don't have to access it over and over
2. `Expression::is_string` already checks that the expression is a literal inside it

sqlglot/generator.py
```diff
@@ -1628,6 +1629,21 @@ def identifier_sql(self, expression: exp.Identifier) -> str:
...
+
+        # For non-literals (variables, expressions), fallback to function call
+        # This may error in other dialects, which is appropriate
+        return self.function_fallback_sql(expression)
```
VaggelisD, line 1645: We can also add a `self.unsupported` message here

sqlglot/dialects/snowflake.py
```diff
@@ -1554,6 +1560,11 @@ def timestampsub_sql(self, expression: exp.TimestampSub):
                 )
             )
 
+        def identifierfunc_sql(self, expression: exp.IdentifierFunc) -> str:
+            """Generate SQL for Snowflake IDENTIFIER() function."""
+            # For Snowflake, always keep as IDENTIFIER() function
+            return self.function_fallback_sql(expression)
```
VaggelisD, line 1566: We can inline this into `TRANSFORMS` such as:

```Python3
TRANSFORMS = {
 ...,
 exp.IdentifierFunc: rename_func("IDENTIFIER"),
 ...
}
```

tests/dialects/test_snowflake.py
```diff
@@ -2977,3 +2977,61 @@ def test_get_extract(self):
...
+            ).sql("snowflake"),
+            "WITH x AS (SELECT 1 AS foo) SELECT IDENTIFIER('foo') FROM x",
+        )
+        # When generating to other dialects, it converts to identifiers
+        self.assertEqual(
+            parse_one(
+                "WITH x AS (SELECT 1 AS foo) SELECT foo FROM IDENTIFIER('x')", read="snowflake"
+            ).sql("postgres"),
+            "WITH x AS (SELECT 1 AS foo) SELECT foo FROM x",
+        )
```
VaggelisD, line 3037: This pattern is identical to what `self.validate_identity` or `self.validate_all` do i.e it parses & generates the query back using either the same dialect (former) or any other dialect (latter).

We have also covered these cases in the previous `validate_all` tests

VaggelisD:
Hey @rahulj51 👋 Just checking in, do you plan to work on this soon? Thanks in advance!

rahulj51:
@VaggelisD . Yes. I do. But not before the weekend. Hope that's ok.

VaggelisD:
@rahulj51 That's fine, thank you for letting us know! Keeping this PR open.

rahulj51:
Closing till I get some time to work on the review comments.

Issue #5512, opened 2025-08-01
Doris dialect incorrectly includes AS keyword in UPDATE and DELETE statements when transpiling from PostgreSQL


peterylh:
# Summary

When transpiling SQL from PostgreSQL to Doris dialect, the `AS` keyword is incorrectly preserved in UPDATE and DELETE statements with table aliases. Doris does not support the `AS` keyword for table aliases in UPDATE and DELETE statements, which causes syntax errors.

# Steps to Reproduce

1. Install SQLGlot
2. Run the following code:

```python
import sqlglot

# Test UPDATE statement with alias
update_sql = "UPDATE users AS u SET u.name = 'John' WHERE u.id = 1"
result = sqlglot.transpile(update_sql, read="postgres", write="doris")[0]
print(f"UPDATE result: {result}")

# Test DELETE statement with alias
delete_sql = "DELETE FROM users AS u WHERE u.id = 1"
result = sqlglot.transpile(delete_sql, read="postgres", write="doris")[0]
print(f"DELETE result: {result}")
```

# Expected Behavior

The transpiled SQL should remove the `AS` keyword from table aliases in UPDATE and DELETE statements:

- UPDATE: `UPDATE users u SET u.name = 'John' WHERE u.id = 1`
- DELETE: `DELETE FROM users u WHERE u.id = 1`

# Actual Behavior

The `AS` keyword is preserved in the output, causing invalid Doris SQL:

- UPDATE: `UPDATE users AS u SET u.name = 'John' WHERE u.id = 1`
- DELETE: `DELETE FROM users AS u WHERE u.id = 1`

## Additional Context

- The issue affects both UPDATE and DELETE statements with table aliases
- SELECT statements correctly preserve the `AS` keyword (which is expected behavior)
- The issue occurs specifically when transpiling from PostgreSQL to Doris
- Doris syntax documentation confirms that `AS` keyword is not supported for table aliases in UPDATE/DELETE statements

## Environment

- SQLGlot version: [current version]
- Python version: 3.x

## Proposed Solution

The issue can be resolved by modifying the `table_sql` method in the Doris dialect generator to detect when a table is part of an UPDATE or DELETE statement and remove the `AS` keyword accordingly, while preserving it for other statement types like SELECT.

VaggelisD:
Fixed by https://github.com/tobymao/sqlglot/pull/5517

PR #5517, opened 2025-08-02
feat(doris): Override table_sql to avoid AS keyword in UPDATE and DELETE statements

peterylh:
This PR implements an override of the table_sql method for the Doris dialect to resolve the issue #5512 where the AS keyword should not be used in UPDATE and DELETE statements.

### Changes Made
- File : sqlglot/dialects/doris.py
- Feature : Override table_sql method to use space separator instead of "AS" keyword in UPDATE and DELETE statements
- Implementation : Check parent nodes of the expression to determine if it's within an UPDATE or DELETE statement, and use space separator accordingly

### DIFF
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -684,8 +684,6 @@ class Doris(MySQL):
             create_sql = ", ".join(self.sql(e) for e in create_expressions)
             return f"PARTITION BY RANGE ({partition_expressions}) ({create_sql})"
 
-
-
         def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
             """Override table_sql to avoid AS keyword in UPDATE and DELETE statements."""
             # Check if this table is part of an UPDATE or DELETE statement or their clauses
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -683,3 +683,15 @@ class Doris(MySQL):
             # Handle both static and dynamic partition definitions
             create_sql = ", ".join(self.sql(e) for e in create_expressions)
             return f"PARTITION BY RANGE ({partition_expressions}) ({create_sql})"
+
+
+
+        def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
+            """Override table_sql to avoid AS keyword in UPDATE and DELETE statements."""
+            # Check if this table is part of an UPDATE or DELETE statement or their clauses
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/doris.py
```diff
@@ -683,3 +683,13 @@ def partitionbyrangeproperty_sql(self, expression):
...
             return f"PARTITION BY RANGE ({partition_expressions}) ({create_sql})"
+
+        def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
+            """Override table_sql to avoid AS keyword in UPDATE and DELETE statements."""
+            # Check if this table is part of an UPDATE or DELETE statement or their clauses
+            current = expression.parent
+            while current:
+                if isinstance(current, (exp.Update, exp.Delete)):
+                    return super().table_sql(expression, sep=" ")
+                current = getattr(current, "parent", None)
```
VaggelisD: You can rewrite this using `expression.find_ancestor(exp.Update, exp.Delete)`

VaggelisD (commented):
Hey @peterylh, thank you for the contribution!

Can we also add a few tests to showcase that `AS` is not used? You can place them in `test_doris.py` for instance.

sqlglot/dialects/doris.py
peterylh: done

tests/dialects/test_doris.py
```diff
@@ -136,3 +136,87 @@ def test_partitionbyrange(self):
...
+        )
+
+        # UPDATE with CASE statement
+        self.validate_all(
+            "UPDATE users AS u SET u.status = CASE WHEN u.age >= 18 THEN 'adult' ELSE 'minor' END",
+            write={
+                "doris": "UPDATE users u SET u.`status` = CASE WHEN u.age >= 18 THEN 'adult' ELSE 'minor' END",
+                "postgres": "UPDATE users AS u SET u.status = CASE WHEN u.age >= 18 THEN 'adult' ELSE 'minor' END",
+            },
+        )
```
VaggelisD, line 204: Note: The way these tests are written should actually be in `test_postgres.py` since `validate_all` will treat the input query as Doris, which would be incorrect, even though SQLGlot is lenient enough with it's parsing.

VaggelisD (commented):
Hey @peterylh, thank you the contribution, I'll merge this and get it to the finish line.

PR #5520, opened 2025-08-03
[WIP] adds start and end index to metadata

big-c-note:
Adds a start and end index to the metadata for use in an lsp I'm working on - plan to release in the next week or so. I have to find the node associated with a line and cursor position. By adding the start and end index I can locate the node associated with the line and cursor position. Then I can traverse the AST. 

My question is - would you accept a change like this? Or, is there some other way using the library that I can get this information/add it to the metadata?

Awesome library, thank you!


tobymao:
this already exists, look at update_positions

big-c-note:
Thank you

PR #5532, opened 2025-08-05
feat(optimizer)!: parse and annotate type for bigquery DATE_FROM_UNIX_DATE

geooo109:
This PR adds parsing and type annotation support for `DATE_FROM_UNIX_DATE`

**DOCS**
[BigQuery DATE_FROM_UNIX_DATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_from_unix_date)


geooo109:
A comment here, we can also use the `DATE_ADD` function instead of creating a new `Expression` and parse it as: `DATE_ADD(DATE '1970-01-01', days)`

VaggelisD:
> A comment here, we can also use the DATE_ADD function instead of creating a new Expression and parse it as: DATE_ADD(DATE '1970-01-01', days)

Yeah, we can most likely canonicalize it but it might be annoying for users since it's counterpart is harder to reason about. 

We can preserve the new expression, generate this function in Spark & BQ and add the `DATE_ADD(...)` to `generator.sql`

VaggelisD (approved):

PR #5534, opened 2025-08-06
feat(optimizer)!: parse and annotate type for bigquery JUSTIFY funcs

geooo109:
This PR adds parsing and type annotation for `JUSTIFY` related functions.

**DOCS**
[BigQuery JUSTIFY funcs](https://cloud.google.com/bigquery/docs/reference/standard-sql/interval_functions)
[Postgres JUSTIFY funcs](https://www.postgresql.org/docs/current/functions-datetime.html)

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -497,11 +497,6 @@ class BigQuery(Dialect):
             e, exp.DataType.build("ARRAY<VARCHAR>")
         ),
         exp.JSONType: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.VARCHAR),
-        exp.JustifyDays: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.INTERVAL),
-        exp.JustifyHours: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.INTERVAL),
-        exp.JustifyInterval: lambda self, e: self._annotate_with_type(
-            e, exp.DataType.Type.INTERVAL
-        ),
         exp.Lag: lambda self, e: self._annotate_by_args(e, "this", "default"),
         exp.SHA: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.BINARY),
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -497,6 +497,11 @@ class BigQuery(Dialect):
             e, exp.DataType.build("ARRAY<VARCHAR>")
         ),
         exp.JSONType: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.VARCHAR),
+        exp.JustifyDays: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.INTERVAL),
+        exp.JustifyHours: lambda self, e: self._annotate_with_type(e, exp.DataType.Type.INTERVAL),
+        exp.JustifyInterval: lambda self, e: self._annotate_with_type(
+            e, exp.DataType.Type.INTERVAL
+        ),
```
VaggelisD: For all 3 dialects (BigQuery, Materialize, Postgres) the signatures are the same, we can move that into `Dialect` under the `exp.DataType.Type.INTERVAL` entry

tests/dialects/test_dialect.py
```diff
@@ -3781,3 +3781,41 @@ def test_date_to_unix_date(self):
                 "trino": "DATE_ADD('DAY', 1, CAST('1970-01-01' AS DATE))",
             },
         )
+
+    def test_justify(self):
+        self.validate_all(
+            "JUSTIFY_DAYS(INTERVAL '1' DAY)",
```
VaggelisD, line 3787: We can also add the [materialize](https://materialize.com/docs/sql/functions/justify-days/) functions here

tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -670,6 +670,18 @@ TIME;
 DATE_FROM_UNIX_DATE(1);
 DATE;
 
+# dialect: bigquery
```
VaggelisD: We can remove `dialect` or do `dialect: bigquery, postgres, materialize` if we move the type to base annotators

PR #5545, opened 2025-08-08
(fix) issue_5542: merge_subqueries loses column name

snovik75:


snovik75:
> Thank you!

@VaggelisD @geooo109 still thinking of it before it is merged. what would be the expected behavior in such case? will the autogenerated column name be the same? I think it will be different and also depends on the dialect. should it be taken care off? raise? ignore? or this is just way of a corner case?

```
with cte as (
    select x.a * x.b as mult from x
)
select abs(cte.mult) from cte;
```

snovik75:
this is what chatgpt says

image

georgesittas:
> > Thank you!
> 
> @VaggelisD @geooo109 still thinking of it before it is merged. what would be the expected behavior in such case? will the autogenerated column name be the same? I think it will be different and also depends on the dialect. should it be taken care off? raise? ignore? or this is just way of a corner case?
> 
> ```
> with cte as (
>     select x.a * x.b as mult from x
> )
> select abs(cte.mult) from cte;
> ```

Hi @snovik75, I think it's fine to let this be. Queries that require consistent naming should be explicit and use aliases.

georgesittas (approved):
Seems legit, thank you for the fix.

PR #5548, opened 2025-08-08
Chore: Fix UNNEST qualification regression

VaggelisD:
Given a query such as:

```Python3
>>> sql = """
... WITH scores as (
...   select *
...   from unnest((
...       select 
...         ARRAY<STRUCT<percentile STRING, value INT64, score FLOAT64>>[("p10", 1, 0.0)]
...     )
...   )
... )
... SELECT percentile FROM scores
... """

```

<br/>

- Before:
```Python3
>>> qualify(sqlglot.parse_one(sql, dialect="bigquery"), dialect="bigquery").sql("bigquery")
  File "<python-input-4>", line 1, in <module>
    qualify(sqlglot.parse_one(sql, dialect="bigquery"), dialect="bigquery").sql("bigquery")
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vaggelisd/Desktop/tobiko/sqlglot/sqlglot/optimizer/qualify.py", line 102, in qualify
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/vaggelisd/Desktop/tobiko/sqlglot/sqlglot/optimizer/qualify_columns.py", line 126, in validate_qualify_columns
sqlglot.errors.OptimizeError: Column '"percentile"' could not be resolved
```

<br/>

- After
```Python3
>>> qualify(sqlglot.parse_one(sql, dialect="bigquery"), dialect="bigquery").sql("bigquery")
"WITH `scores` AS (SELECT * FROM UNNEST((SELECT ARRAY<STRUCT<`percentile` STRING, `value` INT64, `score` FLOAT64>>[('p10', 1, 0.0)] AS `_col_0`))) SELECT `scores`.`percentile` AS `percentile` FROM `scores` AS `scores`"
```

The "regression" was introduced by https://github.com/tobymao/sqlglot/pull/5451 which always added a column alias in `TableAlias`. This caused `Resolver::get_table(...)` to not classify  the `scores` CTE as "source without schema" and thus couldn't qualify the left over columns.

The "regression" is quoted because the qualification before that PR used to work (and now works again) due to SQLGlot's flexibility / fallback mechanism; In contrast, we could actually enhance the `STRUCT` expand logic to also cover the case of `UNNEST((SELECT ARRAY<STRUCT>(...)))` properly.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -1138,7 +1138,12 @@ class DuckDB(Dialect):
                 alias = expression.args.get("alias")
                 if alias:
                     expression.set("alias", None)
-                    alias = exp.TableAlias(this=seq_get(alias.args.get("columns"), 0))
+                    this = (
+                        alias.this.copy()
+                        if alias.args.get("column_only")
+                        else seq_get(alias.args.get("columns"), 0)
+                    )
+                    alias = exp.TableAlias(this=this)
 
                 unnest_sql = super().unnest_sql(expression)
                 select = exp.Select(expressions=[unnest_sql]).subquery(alias)
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -459,12 +459,13 @@ def _eliminate_dot_variant_lookup(expression: exp.Expression) -> exp.Expression:
         unnest_aliases = set()
         for unnest in find_all_in_scope(expression, exp.Unnest):
             unnest_alias = unnest.args.get("alias")
-            if (
-                isinstance(unnest_alias, exp.TableAlias)
-                and (unnest_alias.args.get("column_only") or not unnest_alias.this)
-                and len(unnest_alias.columns) == 1
-            ):
-                unnest_aliases.add(unnest_alias.columns[0].name)
+            if isinstance(unnest_alias, exp.TableAlias):
+                column_only = unnest_alias.args.get("column_only")
+                if column_only and not unnest_alias.columns:
+                    unnest_alias.set("columns", [unnest_alias.this.copy()])
+
+                if (column_only or not unnest_alias.this) and len(unnest_alias.columns) == 1:
+                    unnest_aliases.add(unnest_alias.columns[0].name)
 
         if unnest_aliases:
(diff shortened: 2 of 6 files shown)


georgesittas:
Thanks for trying to fix this @VaggelisD. I'd rather close this and revert the linked PR, because I'm worried about the added complexity and it turns out more things broke due to it, not just the case you're talking about. For example:

Before:

```python
>>> from sqlglot.optimizer.qualify import qualify
>>> sql = 'SELECT * FROM UNNEST(ARRAY<STRUCT<percentile STRING, value INT64, score FLOAT64>>[("p10", 1, 0.0)])'
>>> ast = parse_one(sql, dialect="bigquery")
>>> qualified = qualify(ast, dialect="bigquery")
>>> qualified.sql('bigquery')
"SELECT `percentile` AS `percentile`, `value` AS `value`, `score` AS `score` FROM UNNEST(ARRAY<STRUCT<`percentile` STRING, `value` INT64, `score` FLOAT64>>[('p10', 1, 0.0)])"
```

After:

```python
>>> sql = 'SELECT * FROM UNNEST(ARRAY<STRUCT<percentile STRING, value INT64, score FLOAT64>>[("p10", 1, 0.0)])'
>>>
>>> ast = parse_one(sql, dialect="bigquery")
>>> qualified = qualify(ast, dialect="bigquery")
>>>
>>> qualified.sql("bigquery")
"SELECT `_q_0` AS `_q_0`, `percentile` AS `percentile`, `value` AS `value`, `score` AS `score` FROM UNNEST(ARRAY<STRUCT<`percentile` STRING, `value` INT64, `score` FLOAT64>>[('p10', 1, 0.0)]) AS `_q_0`"
```

Users that want to transpile BigQuery to Snowflake have a path forward even without that PR:

1. Add aliases to each `UNNEST` , e.g. instead of from `FROM UNNEST(…), UNNEST(foo), …`, do `FROM UNNEST(...) AS data, UNNEST(data.foo), …`
2. Ensure all references to columns produced by `UNNEST` operators are properly qualified. So, a top-level projection `fld` would become `data.fld`

PR #5549, opened 2025-08-08
fix(doris): Rename Table

xinge-ji:
https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/table/ALTER-TABLE-RENAME

Doris uses `alter table test.t1 rename t2` or `alter table t1 rename t2`


.github/scripts/format_benchmark.py
georgesittas, line 1: Why did this file change?

pdoc/cli.py
georgesittas, line 1: Ditto, why did this change?

sqlglot/dialects/doris.py
```diff
@@ -690,3 +691,12 @@ def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
...
             return super().table_sql(expression, sep=sep)
+
+        def alterrename_sql(self, expression: exp.AlterRename) -> str:
+            if not self.RENAME_TABLE_WITH_DB:
+                # Remove db from tables
+                expression = expression.transform(
+                    lambda n: exp.table_(n.this) if isinstance(n, exp.Table) else n
+                ).assert_is(exp.AlterRename)
+            this = self.sql(expression, "this")
+            return f"RENAME {this}"
```
georgesittas: We should not copy-paste code like this. It appears that the only part that changed was `RENAME TO` -> `RENAME`. These small tweaks should be controlled by either a method kwarg or a class constant (preferably the former in this case).

georgesittas (changes requested):

.github/scripts/format_benchmark.py
xinge-ji: Sorry — this was a cache file; I removed it.

pdoc/cli.py
xinge-ji: I removed it

sqlglot/dialects/doris.py
xinge-ji: I added a new argument `include_to` to `alterrename_sql` in `generator.py` (default: `True`) so the generator can emit either `RENAME TO` or `RENAME`. Set `include_to=False` in the Doris dialect so it emits `RENAME` there.

georgesittas (approved):
Much cleaner! Thank you for addressing the comments.

PR #5552, opened 2025-08-08
fix(parser)!: WEEKDAY of WEEK as VAR

### LINKED ISSUE #5547: BigQuery: week(monday) raises "column x could not be resolved"
Hi, I'm trying to recover unresolved columns and I'm having an issue with [this BQ option](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_trunc).

Running:
 -  python 3.12
 -  sqlglot 27.6.0

```python
import sqlglot
from sqlglot import expressions
from sqlglot.optimizer import qualify, scope

def extract_columns_from_ast(sql_code_ast: expressions.Expression) -> list[str]:
    columns = []
    for column in qualify.qualify(sql_code_ast).selects:  # type: ignore
        # If there's a select * in the final CTE
        if column.is_star:
            root = scope.build_scope(qualify.qualify(column.parent))
            try:
                # If the column is a `select *` from a directly referenced table
                if root:
                    columns.append(
                        f"* from {scope.find_in_scope(root.expression, sqlglot.exp.Table).name}"  # type: ignore
                    )
            except AttributeError:
                # If the column is a `select * from (select * from table)`
                columns.append("* from nested query")
        else:
            columns.append(column.alias)
    return columns


sql = """
 with date_array as (
  select local_date from
            unnest(
                generate_date_array(
                    "2025-01-01",
                    "2025-01-02",
                    interval 1 day
                )
            ) as local_date
            ),

    daily_dates as (
        select local_date, date_trunc(local_date, week(monday)) as week_start_date
        from date_array
    )

    select * from daily_dates
"""

sql_ast_code = sqlglot.parse_one(sql, read="bigquery")
result = extract_columns_from_ast(sql_ast_code)

```

I get
```
  File "test.py", line 46, in <module>
    result = extract_columns_from_ast(sql_ast_code)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "test.py", line 7, in extract_columns_from_ast
    for column in qualify.qualify(sql_code_ast).selects:  # type: ignore
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../sqlglot/optimizer/qualify.py", line 102, in qualify
  File ".../sqlglot/optimizer/qualify_columns.py", line 126, in validate_qualify_columns
sqlglot.errors.OptimizeError: Column '"monday"' could not be resolved
```

This is not a column but a BigQuery keyword, and sqlglot takes it as a column, so that's why I think it can't be resolved.

Maybe there's something very wrong I'm doing with my code, any suggestion for that would work. What I'm trying to achieve is to flag any column that was not declared at least once in the whole sql code.

geooo109:
Fixes #5547

In BigQuery `DATE_TRUNC` can use as argument the `WEEK(WEEKDAY)`.
Previously, `WEEKDAY` was parsed as `exp.Column`, creating issues with the `qualify`. 
In the new version, we parse it as `exp.Var`.

**DOCS**
[BigQuery WEEK(WEEKDAY)](https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_diff)

### DIFF
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1770,20 +1770,6 @@ WHERE
             },
         )
 
-        trunc_expr = self.parse_one("DATE_TRUNC(date, WEEK(MONDAY))")
-        self.assertEqual(trunc_expr.find(exp.Var).name, "MONDAY")
-        self.validate_identity("DATE_TRUNC(date, WEEK(MONDAY))")
-        self.validate_identity(
-            "LAST_DAY(DATETIME '2008-11-10 15:30:00', WEEK(SUNDAY))",
-            "LAST_DAY(CAST('2008-11-10 15:30:00' AS DATETIME), WEEK(SUNDAY))",
-        )
-        self.validate_identity("DATE_DIFF('2017-12-18', '2017-12-17', WEEK(SATURDAY))")
-        self.validate_identity("DATETIME_DIFF('2017-12-18', '2017-12-17', WEEK(MONDAY))")
-        self.validate_identity(
-            "EXTRACT(WEEK(THURSDAY) FROM DATE '2013-12-25')",
-            "EXTRACT(WEEK(THURSDAY) FROM CAST('2013-12-25' AS DATE))",
-        )
-
     def test_errors(self):
         with self.assertRaises(ParseError):
             self.parse_one("SELECT * FROM a - b.c.d2")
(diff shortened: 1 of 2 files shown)


georgesittas (commented):
FYI, there are other functions where `WEEK(WEEKDAY)` can be used:

- https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_diff
- https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_trunc
- https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#extract
- https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#last_day
- https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_diff

I don't think the current approach is correct. Representing this unit using `exp.Week` also seems a bit weird, I think the original intention behind that expression was to support MySQL's `WEEK` [function](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_week).

My hunch is we should add a `"WEEK"` entry in `BigQuery.Parser.FUNCTIONS` such that it produces a different AST that properly reflects this is a unit. I believe there's no `WEEK` function in BigQuery besides this form that is only used for unit arguments.

geooo109:
You have right, tried to merge all these together, but the repr isn't correct.
Added a new impl by separating the Function and the Expression.

georgesittas (approved):
Let's test `qualify` as well, since that's what the original issue was about.

PR #5553, opened 2025-08-09
fix(doris): fix DATE_TRUNC and partition by

xinge-ji:
Fix Doris date_trunc syntax `DATE_TRUNC(<datetime>, <time_unit>)` https://doris.apache.org/docs/sql-manual/sql-functions/scalar-functions/date-time-functions/date-trunc

Support partition by syntax in asynchronous materialized view https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/async-materialized-view/CREATE-ASYNC-MATERIALIZED-VIEW

### DIFF
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -35,9 +35,7 @@ def _build_date_trunc(args: t.List[exp.Expression]) -> exp.Expression:
     # Determine which argument is the unit
     unit, this = a0, a1 if _is_unit_like(a0) else a1, a0
 
-    node = exp.TimestampTrunc(this=this, unit=unit)
-
-    return node
+    return exp.TimestampTrunc(this=this, unit=unit)
 
 
 class Doris(MySQL):


sqlglot/dialects/doris.py
```diff
@@ -31,7 +31,10 @@ class Parser(MySQL.Parser):
         FUNCTIONS = {
             **MySQL.Parser.FUNCTIONS,
             "COLLECT_SET": exp.ArrayUniqueAgg.from_arg_list,
-            "DATE_TRUNC": build_timestamp_trunc,
+            # Doris uses DATE_TRUNC(expr, 'unit') order
```
erindru: Does it though? The docs seem to show that the argument order is interchangable: https://doris.apache.org/docs/sql-manual/sql-functions/scalar-functions/date-time-functions/date-trunc#syntax

xinge-ji: I updated the implementation to accept both orders and added tests to cover each case.

sqlglot/dialects/doris.py
```diff
@@ -22,6 +22,48 @@ def _lag_lead_sql(self, expression: exp.Lag | exp.Lead) -> str:
...
+    if unit_first:
+        unit, this = a0, a1
+    else:
+        this, unit = a0, a1
+
+    node = exp.TimestampTrunc(this=this, unit=unit)
+
+    # Preserve original argument order for identity rendering via meta
+    node.meta["unit_first"] = unit_first
+    return node
```
georgesittas: Why do we need to preserve the order? Sqlglot doesn't have to guarantee the same order, we can canonicalize these `DATE_TRUNC` calls such that, e.g., the unit always comes second.

sqlglot/dialects/doris.py
```diff
@@ -22,6 +22,48 @@ def _lag_lead_sql(self, expression: exp.Lag | exp.Lead) -> str:
     )
 
 
+# Accept both DATE_TRUNC(datetime, unit) and DATE_TRUNC(unit, datetime)
+def _build_date_trunc(args: list[exp.Expression]) -> exp.Expression:
+    a0, a1 = seq_get(args, 0), seq_get(args, 1)
+
+    def _is_unit_like(e: exp.Expression | None) -> bool:
```
georgesittas, line 29: I'm not sure I like this approach. Perhaps a simpler heuristic could be: if `e` is not a string literal, then it can't be a unit (I assume?). Otherwise, `e` is unit-like iff it doesn't contain any digits in its `name`.

sqlglot/dialects/doris.py
```diff
@@ -684,6 +735,13 @@ def partitionbyrangeproperty_sql(self, expression):
             create_sql = ", ".join(self.sql(e) for e in create_expressions)
             return f"PARTITION BY RANGE ({partition_expressions}) ({create_sql})"
 
+        def partitionedbyproperty_sql(self, expression: exp.PartitionedByProperty) -> str:
+            # Avoid double parens; Doris syntax expects a single set of parens
+            inner = self.sql(expression, "this")
+            if inner.startswith("(") and inner.endswith(")"):
+                inner = inner[1:-1]
+            return f"PARTITION BY ({inner})"
```
georgesittas: Why would there be a redundant set of parentheses? What's the AST that results in that SQL?

sqlglot/dialects/doris.py
xinge-ji: I’ve removed the interchangeable order support and always placing the unit second

sqlglot/dialects/doris.py
xinge-ji: I've implemented a new version based on your idea

sqlglot/dialects/doris.py
xinge-ji: The AST is `PartitionedByProperty(this=Schema(expressions=[...]))`

I reimplemented the handling so the serializer flattens that case and doesn’t emit the extra parentheses.

sqlglot/dialects/doris.py
```diff
@@ -22,6 +22,28 @@ def _lag_lead_sql(self, expression: exp.Lag | exp.Lead) -> str:
     )
 
 
+# Accept both DATE_TRUNC(datetime, unit) and DATE_TRUNC(unit, datetime)
+def _build_date_trunc(args: list[exp.Expression]) -> exp.Expression:
```
georgesittas:
```suggestion
def _build_date_trunc(args: t.List[exp.Expression]) -> exp.Expression:
```

sqlglot/dialects/doris.py
```diff
@@ -22,6 +22,28 @@ def _lag_lead_sql(self, expression: exp.Lag | exp.Lead) -> str:
...
+            return False
+        text = e.this
+        return not any(ch.isdigit() for ch in text)
+
+    # Determine which argument is the unit
+    unit_first = _is_unit_like(a0) and not _is_unit_like(a1)
+    if unit_first:
+        unit, this = a0, a1
+    else:
+        this, unit = a0, a1
```
georgesittas: I think we can simplify this a little?
```suggestion
    unit, this = a0, a1 if _is_unit_like(a0) else a1, a0
```

sqlglot/dialects/doris.py
```diff
@@ -22,6 +22,28 @@ def _lag_lead_sql(self, expression: exp.Lag | exp.Lead) -> str:
...
+
+    node = exp.TimestampTrunc(this=this, unit=unit)
+
+    return node
```
georgesittas:
```suggestion
    return exp.TimestampTrunc(this=this, unit=unit)
```

georgesittas (approved):

PR #5555, opened 2025-08-11
feat(singlestore): Added parsing and generation of JSON_EXTRACT

AdalbertMemSQL:
Appropriate SingleStore documentation:
https://docs.singlestore.com/cloud/reference/sql-reference/json-functions/json-extract-type/
https://docs.singlestore.com/db/v9.0/reference/sql-reference/bson-functions/


sqlglot/dialects/singlestore.py
```diff
@@ -5,7 +5,46 @@
 from sqlglot.dialects.dialect import build_formatted_time
 from sqlglot.dialects.mysql import MySQL
 from sqlglot.generator import unsupported_args
-from sqlglot.helper import seq_get
+from sqlglot.helper import seq_get, is_int
+from sqlglot._typing import F
+
+
+def build_json_extract_path(
```
georgesittas: Can we avoid duplicating all this logic and instead augment the base `build_json_extract_path` helper in `dialect.py` to take the new `json_type` kwarg into account?

sqlglot/dialects/singlestore.py
```diff
@@ -64,6 +103,26 @@ class Parser(MySQL.Parser):
...
+            "JSON_EXTRACT_JSON": build_json_extract_path(exp.JSONExtract),
+            "JSON_EXTRACT_STRING": build_json_extract_path(
+                exp.JSONExtractScalar, json_type="STRING"
+            ),
+            "JSON_EXTRACT_DOUBLE": build_json_extract_path(
+                exp.JSONExtractScalar, json_type="DOUBLE"
+            ),
+            "JSON_EXTRACT_BIGINT": build_json_extract_path(
+                exp.JSONExtractScalar, json_type="BIGINT"
+            ),
```
georgesittas, line 95: Is there a variant of `JSON_EXTRACT` that works on all types? I'm wondering if we can produce a `CAST` to the `<type>` in `JSON_EXTRACT_<type>`, instead of storing that info in the `JSONExtract*` nodes.

sqlglot/dialects/singlestore.py
```diff
@@ -1166,3 +1246,98 @@ class Generator(MySQL.Generator):
             "zerofill",
             "zone",
         }
+
+        def jsonpathkey_sql(self, expression: exp.JSONPathKey) -> str:
```
georgesittas: Why do we need to override the generators for these `JSONPathPart` subclasses?

sqlglot/dialects/singlestore.py
AdalbertMemSQL: JSON_EXTRACT_JSON can extract any type, but casting of JSON object into a concrete datatype returns different results.
```
MySQL [(none)]> SELECT JSON_EXTRACT_JSON('{"a": "b"}', 'a') :> TEXT AS data;
+------+
| data |
+------+
| "b"  |
+------+
MySQL [(none)]> SELECT JSON_EXTRACT_STRING('{"a": "b"}', 'a') AS data;
+------+
| data |
+------+
| b    |
+------+
1 row in set (0.021 sec)
```

```
MySQL [(none)]> SELECT JSON_EXTRACT_JSON('{"a": null}', 'a') :> TEXT AS data;
+------+
| data |
+------+
| null |
+------+
1 row in set (0.001 sec)

MySQL [(none)]> SELECT JSON_EXTRACT_STRING('{"a": null}', 'a') AS data;
+------+
| data |
+------+
| NULL |
+------+
1 row in set (0.001 sec)

```

sqlglot/dialects/singlestore.py
AdalbertMemSQL: In SingleStore, the JSON path is provided as a comma-separated list of keys/indexes.
For example, instead of `$.b[2]`, SingleStore expects `'b', 2`.
To support this, I overrode `jsonpathkey_sql` and `jsonpathsubscript_sql`.

For all other JSON path parts that SingleStore does not support, I added an explicit "not supported" message, for example:
```
self.unsupported("JSONPathRoot is not supported in SingleStore")
```

AdalbertMemSQL: I saw a simillar unsupported usage in the MySQL dialect
```
        def array_sql(self, expression: exp.Array) -> str:
            self.unsupported("Arrays are not supported by MySQL")
            return self.function_fallback_sql(expression)

        def arraycontainsall_sql(self, expression: exp.ArrayContainsAll) -> str:
            self.unsupported("Array operations are not supported by MySQL")
            return self.function_fallback_sql(expression)
```

sqlglot/dialects/singlestore.py
AdalbertMemSQL: I updated `build_json_extract_path` helper in `dialect.py`. The new `json_type` kwarg was not the only needed change. I also added handling of `exp.Column` (it is provided by the parser when JSON column operator (`::$`, `::%`) is used)

AdalbertMemSQL: Looks like a change for `exp.Column` broke Postgres functionality. I reverted it and handled this case inside the SingleStore dialect.

georgesittas:
Thanks for addressing comments @AdalbertMemSQL, I'll review again shortly.

sqlglot/dialects/dialect.py
```diff
@@ -1728,7 +1728,10 @@ def normalize(identifier: t.Optional[exp.Identifier]) -> t.Optional[str]:
 
 
 def build_json_extract_path(
-    expr_type: t.Type[F], zero_based_indexing: bool = True, arrow_req_json_type: bool = False
+    expr_type: t.Type[F],
+    zero_based_indexing: bool = True,
+    arrow_req_json_type: bool = False,
+    json_type: t.Union[str, None] = None,
```
georgesittas:
```suggestion
    json_type: t.Optional[str] = None,
```

sqlglot/dialects/dialect.py
```diff
@@ -1748,11 +1751,19 @@ def _builder(args: t.List) -> F:
...
-            only_json_types=arrow_req_json_type,
-        )
+        kwargs = {
+            "this": seq_get(args, 0),
+            "expression": exp.JSONPath(expressions=segments),
+        }
+
+        is_jsonb = issubclass(expr_type, (exp.JSONBExtract, exp.JSONBExtractScalar))
+        if not is_jsonb:
+            kwargs["only_json_types"] = arrow_req_json_type
```
georgesittas, line 1807: Why was this necessary? Not sure if this is correct, doesn't seem bad to just set `only_json_types` regardless, right?

sqlglot/dialects/singlestore.py
```diff
@@ -79,6 +99,21 @@ class Parser(MySQL.Parser):
                 this=this,
                 to=to,
             ),
+            TokenType.DCOLON: lambda self, this, path: build_json_extract_path(exp.JSONExtract)(
+                [this, exp.Literal.string(path.this.this)]
+            )
+            if isinstance(path, exp.Column)
+            else build_json_extract_path(exp.JSONExtract)([this, path]),
```
georgesittas: When can `path` _not_ be a `Column` here? Seems like singlestore only expects `foo::bar` based on the docs. Ditto for `DCOLONDOLLAR` and `DCOLONPERCENT`.

sqlglot/dialects/singlestore.py
```diff
@@ -1166,3 +1213,98 @@ class Generator(MySQL.Generator):
...
+        def jsonvalue_sql(self, expression: exp.JSONValue) -> str:
+            path = self.sql(expression, "path")
+
+            res = self.func("JSON_EXTRACT_STRING", expression.this, f"{path}")
+
+            returning = self.sql(expression, "returning")
+            if returning:
+                return f"{res} :> {returning}"
+            else:
+                return res
```
georgesittas:
```suggestion
            return f"{res} :> {returning}" if returning else res
```

sqlglot/dialects/singlestore.py
```diff
@@ -1166,3 +1213,98 @@ class Generator(MySQL.Generator):
...
+
+        def jsonextractarray_sql(self, expression: exp.JSONExtractArray) -> str:
+            self.unsupported("Arrays are not supported in SingleStore")
+            return self.function_fallback_sql(expression)
+
+        @unsupported_args("on_condition")
+        def jsonvalue_sql(self, expression: exp.JSONValue) -> str:
+            path = self.sql(expression, "path")
+
+            res = self.func("JSON_EXTRACT_STRING", expression.this, f"{path}")
```
georgesittas:
```suggestion
            res = self.func("JSON_EXTRACT_STRING", expression.this, path)
```

sqlglot/expressions.py
```diff
@@ -6500,7 +6500,13 @@ class JSONExtractArray(Func):
 
 
 class JSONExtractScalar(Binary, Func):
-    arg_types = {"this": True, "expression": True, "only_json_types": False, "expressions": False}
+    arg_types = {
+        "this": True,
+        "expression": True,
+        "only_json_types": False,
+        "json_type": False,
```
georgesittas: Let's move this to the end of the dict.

tests/dialects/test_singlestore.py
```diff
@@ -73,3 +73,89 @@ def test_cast(self):
         self.validate_identity("SELECT '{\"a\" : 1}' :> JSON")
         self.validate_identity("SELECT NOW() !:> TIMESTAMP(6)")
         self.validate_identity("SELECT x :> GEOGRAPHYPOINT")
+
+    def test_json_extract(self):
+        self.validate_identity(
+            "SELECT a::b FROM t", write_sql="SELECT JSON_EXTRACT_JSON(a, 'b') FROM t"
+        )
```
georgesittas: Nit: no need for `write_sql` here and for the other tests.

tests/dialects/test_singlestore.py
```diff
@@ -73,3 +73,89 @@ def test_cast(self):
...
+            write_sql="SELECT JSON_EXTRACT_JSON(JSON_EXTRACT_JSON(a, 'b'), 2) FROM t",
+        )
+        self.validate_identity(
+            "SELECT a::2 FROM t", write_sql="SELECT JSON_EXTRACT_JSON(a, 2) FROM t"
+        )
+
+        self.validate_all(
+            "SELECT JSON_EXTRACT_JSON(a, 'b') FROM t",
+            read={"": "SELECT JSON_EXTRACT(a, '$.b') FROM t"},
+            write={"": "SELECT JSON_EXTRACT(a, '$.b') FROM t"},
```
georgesittas: Why are we testing with `""` instead of a concrete dialect? Let's ensure roundtrip tests are always included by having `"singlestore"` as either a read or write target.

sqlglot/dialects/singlestore.py
georgesittas: > In SingleStore, the JSON path is provided as a comma-separated list of keys/indexes.

Can you check whether `json_extract_segments` does the trick? We do something similar for Postgres: https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/postgres.py#L188-L196 (see [test](https://github.com/tobymao/sqlglot/blob/902a0cdfe46f693aa55612d45a2de2def21f0b8c/tests/dialects/test_postgres.py#L464)).

georgesittas:
Seems like the other PR resulted in some conflicts that need to be resolved as well.

sqlglot/dialects/dialect.py
AdalbertMemSQL: JSONB doesn't have `only_json_types`, and tests were returning an error.
```
sqlglot.errors.ParseError: Unexpected keyword: 'only_json_types' for <class 'sqlglot.expressions.JSONBExtract'>. Line 1, Col: 32.
  SELECT BSON_EXTRACT_BSON(a, 'b') FROM t
```

tests/dialects/test_singlestore.py
AdalbertMemSQL: Without it, the test will fail 
```
Expected :SELECT a::b FROM t
Actual   :SELECT JSON_EXTRACT_JSON(a, 'b') FROM t
```
At the moment, dialect parses `::` as `exp.JsonExtract` and converts `exp.JsonExtract` to `JSON_EXTRACT_JSON` function

georgesittas: Sorry, I meant you can get rid of the kwarg, and just make it a positional arg:

```python
        self.validate_identity(
            "SELECT a::b FROM t",
            "SELECT JSON_EXTRACT_JSON(a, 'b') FROM t",
        )
```

Just matches what we do elsewhere in the codebase today.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: That's true. According to SingleStore documentation, in correct queries path should always be a column

sqlglot/dialects/dialect.py
georgesittas: Ah, I see, I wasn't aware of that. It's fine then.

sqlglot/dialects/singlestore.py
georgesittas: Good, we can then assume this happy path and simplify the logic here.

georgesittas:
Hey @AdalbertMemSQL, any plans to take this to the finish line?

sqlglot/dialects/singlestore.py
AdalbertMemSQL: `json_extract_segments` leaves redundant `$`.
```
Expected :SELECT JSON_EXTRACT_JSON(a, 'b') FROM t
Actual   :SELECT JSON_EXTRACT_JSON(a, '$', ''b'') FROM t
```

sqlglot/dialects/singlestore.py
AdalbertMemSQL: Removing
```
            if isinstance(path, exp.Column)
            else build_json_extract_path(exp.JSONExtract)([this, path]),
```
will lead to an error message that can be very confusing
```
  File "/home/amakarovych-ua/Projects/sqlglot/sqlglot/dialects/singlestore.py", line 103, in <lambda>
    [this, exp.Literal.string(path.this.this)]
AttributeError: 'str' object has no attribute 'this'
```
I will simplify this, but will leave `isinstance(path, exp.Column)` part

sqlglot/dialects/singlestore.py
georgesittas: Perhaps we can switch of the `JSONPathRoot` to avoid this? E.g., for this reason Postgres has the following in its `Generator.TRANSFORMS` mapping:

```python
exp.JSONPathRoot: lambda *_: "",
```

Not sure if the `$...` syntax is used elsewhere for Singlestore.

sqlglot/dialects/singlestore.py
georgesittas: You should do `path.name` instead.

AdalbertMemSQL: Wow. This worked. Thanks for the suggestion.

georgesittas (approved):
Nice work! Thank you for addressing all comments.

georgesittas:
I ended up removing the methods only overridden for `unsupported`: https://github.com/tobymao/sqlglot/commit/c0fec23620a8102ea594ad2ecafd078aefde503d.

The base `Generator` already [raises](https://github.com/tobymao/sqlglot/blob/affa70dd9f188e27fae4cf155594fa0651ec1bb4/sqlglot/generator.py#L961) if an expression type is not supported, so that should be good enough for this use case.

PR #5570, opened 2025-08-13
fix: add special to_char format for dremio

jasonthomassql:
dremio to_char has a bit of a strange special format involving # which instead of converting time, converts ints and floats to varchar.


sqlglot/dialects/dremio.py
```diff
@@ -124,7 +128,7 @@ class Generator(generator.Generator):
         TRANSFORMS = {
             **generator.Generator.TRANSFORMS,
             exp.TimeToStr: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
-            exp.ToChar: lambda self, e: self.function_fallback_sql(e),
+            exp.ToChar: lambda self, e: f"CAST({self.sql(e, 'this')} AS VARCHAR)",
```
georgesittas: Why did this change?

sqlglot/dialects/dremio.py
```diff
@@ -94,7 +94,11 @@ class Parser(parser.Parser):
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
-            "TO_CHAR": build_formatted_time(exp.TimeToStr, "dremio"),
+            "TO_CHAR": lambda args: (
+                exp.ToChar(this=args[0], format=args[1])
+                if len(args) == 2 and isinstance(args[1], exp.Literal) and "#" in args[1].this
+                else build_formatted_time(exp.TimeToStr, "dremio")(args)
+            ),
```
georgesittas: Are there docs that show the available format characters for non-date/time values (i.e., integer) values? I think we should probably produce `TimeToStr` more conservatively. For instance, if it's guaranteed that integer formats *must* contain `#`, then we should only produce `TimeToStr` if a) there are two args, b) the 2nd arg is a literal and c) its `name` doesn't contain any `"#"`, as you did here.

jasonthomassql: Yes, # and #.# are the only document formats but really anything with # passes dremio.  ie ##, ###.##, etc.

This is the only exception I see.  If a # doesn't exist, then TimeToStr should be called.

https://docs.dremio.com/current/reference/sql/sql-functions/functions/TO_CHAR/

georgesittas: Ok. What about columns? Could you have `TO_CHAR(x, y)`? Then it may not be possible to disambiguate.

sqlglot/dialects/dremio.py
jasonthomassql: because the default is CAST AS TIMESTAMP.  I have to override that.

sqlglot/dialects/dremio.py
jasonthomassql: Could possibly happen that the string # is in a column but I think its unlikely.  Think that exists for most exceptions in sqlglot though so not sure how it would be handled but probably a bigger discussion than I can contribute at the moment.  This just handles the explicit "#" case which is currently broken.  To be honest I don't know why Dremio implemented this.  Dremio has a CAST function which is better for this but I can't control what function a user uses.

sqlglot/dialects/dremio.py
jasonthomassql: Is there a better way to do this?  ANNOTATORS?

georgesittas: I guess my question is what's the issue with `function_fallback_sql`? That should produce `"TO_CHAR(...)"`, right?

sqlglot/dialects/dremio.py
georgesittas: I meant to suggest falling back to `ToChar` if you see that the 2nd argument isn't a `Literal` that is a string. That way, if your 2nd argument is a column, whose contents you don't know (it may or may not contain `#`), you don't assume its a `TimeToStr` operation, but rather the more generic `ToChar`.

jasonthomassql: Ah, think I finally understand.  Done.

sqlglot/dialects/dremio.py
jasonthomassql: Yes it would have....removed in latest.

sqlglot/dialects/dremio.py
```diff
@@ -94,7 +94,17 @@ class Parser(parser.Parser):
...
-            "TO_CHAR": build_formatted_time(exp.TimeToStr, "dremio"),
+            "TO_CHAR": lambda args: (
+                exp.ToChar(this=args[0], format=args[1])
+                if (
+                    len(args) == 2
+                    and (
+                        not isinstance(args[1], exp.Literal)
+                        or (isinstance(args[1].this, str) and "#" in args[1].this)
+                    )
+                )
```
georgesittas:
```suggestion
                if len(args) == 2 and (not isinstance(args[1], exp.Literal) or "#" in args[1].name)
```

sqlglot/dialects/dremio.py
```diff
@@ -94,7 +94,14 @@ class Parser(parser.Parser):
...
             **parser.Parser.FUNCTIONS,
-            "TO_CHAR": build_formatted_time(exp.TimeToStr, "dremio"),
+            "TO_CHAR": lambda args: (
+                build_formatted_time(exp.TimeToStr, "dremio")(args)
+                if len(args) == 2
+                and isinstance(args[1], exp.Literal)
+                and isinstance(args[1].this, str)
+                and "#" not in args[1].this
+                else exp.ToChar(this=args[0], format=args[1])
+            ),
```
georgesittas, line 104:
```suggestion
            "TO_CHAR": lambda args: (
                build_formatted_time(exp.TimeToStr, "dremio")(args)
                if len(args) == 2 and (not isinstance(args[1], exp.Literal) or "#" in args[1].name)
                else exp.ToChar(this=args[0], format=args[1])
            ),
```

sqlglot/dialects/dremio.py
```diff
@@ -123,8 +130,13 @@ class Generator(generator.Generator):
...
             **generator.Generator.TRANSFORMS,
-            exp.TimeToStr: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
-            exp.ToChar: lambda self, e: self.function_fallback_sql(e),
+            exp.TimeToStr: lambda self, e: (
+                self.func("TO_CHAR", e.this, self.format_time(e))
+                if isinstance(e.args.get("format"), exp.Literal)
+                and isinstance(e.args["format"].this, str)
+                and "#" not in e.args["format"].this
+                else self.function_fallback_sql(e)
+            ),
```
georgesittas, line 139:
```suggestion
            exp.ToChar: lambda self, e: self.function_fallback_sql(e),
            exp.TimeToStr: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
```

tests/dialects/test_dremio.py
```diff
@@ -131,6 +130,11 @@ def test_time_mapping(self):
             },
         )
 
+    def test_to_char_special(self):
+        self.validate_identity("SELECT TO_CHAR(5555, '#')", "SELECT CAST(5555 AS VARCHAR)")
+        self.validate_identity("SELECT TO_CHAR(3.14, '#.#')", "SELECT CAST(3.14 AS VARCHAR)")
+        self.validate_identity("SELECT TO_CHAR(3.14, columnname)", "SELECT CAST(3.14 AS VARCHAR)")
```
georgesittas, line 136:
```suggestion
        self.validate_identity("SELECT TO_CHAR(5555, '#')")
        self.validate_identity("SELECT TO_CHAR(3.14, '#.#')")
        self.validate_identity("SELECT TO_CHAR(3.14, columnname)")
```

georgesittas:
@jasonthomassql thanks for the patience. I think there was a small disconnect, so I've suggested the exact changes I was thinking of. Can you please try them out and let me know how it goes? I think we can simplify this a bit further, unless I missed something.

georgesittas:
Seems like some stuff broke, lemme take this to the finish line. Thanks!

jasonthomassql:
> Seems like some stuff broke, lemme take this to the finish line. Thanks!

Yeah sorry about that...i did a commit after your suggestions and got mangled.  My latest commit was working.  I think your not logic was not correct.

georgesittas:
What part wasn't correct? I'll try to amend it.

georgesittas:
I think I messed up the order in the `TO_CHAR` builder earlier... Here's the new version: https://github.com/tobymao/sqlglot/commit/a5c42e75f7a924b049059d9aad26e170332a5a1e.

PR #5580, opened 2025-08-14
fix: additional dremio to_char fix as transpile is not working

jasonthomassql:
Our prior work fixed dremio parse but transpile is still returning dremio specific format.

current logic:
```
sql = "SELECT TO_CHAR(5555.00, '#')"
transpiled = sqlglot.transpile(sql, read="dremio", write="postgres")[0]
print(transpiled)
-----OUTPUT
SELECT TO_CHAR(5555.00, '#')
```

this patch:
```
-----OUTPUT
SELECT CAST(5555.00 AS VARCHAR)
```

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -92,6 +92,8 @@ class Dremio(Dialect):
         "tzd": "%Z",  # abbreviation (UTC, PST, ...)
         "TZO": "%z",
         "tzo": "%z",  # numeric offset (+0200)
+        # "#": "FM9999",
+        # "#.#": "FM9999.9",
     }
 
     class Parser(parser.Parser):
@@ -100,7 +102,7 @@ class Dremio(Dialect):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "TO_CHAR": lambda args: (
-                exp.ToChar(this=args[0], format=args[1])
+                exp.Cast(this=args[0], to=exp.DataType.build("VARCHAR"))
                 if len(args) == 2 and (not isinstance(args[1], exp.Literal) or "#" in args[1].name)
                 else build_formatted_time(exp.TimeToStr, "dremio")(args)
             ),
--- a/tests/dialects/test_dremio.py
+++ b/tests/dialects/test_dremio.py
@@ -131,9 +131,15 @@ class TestDremio(Validator):
         )
 
     def test_to_char_special(self):
-        self.validate_identity("SELECT TO_CHAR(5555, '#')").selects[0].assert_is(exp.ToChar)
-        self.validate_identity("SELECT TO_CHAR(3.14, '#.#')").selects[0].assert_is(exp.ToChar)
-        self.validate_identity("SELECT TO_CHAR(3.14, columnname)").selects[0].assert_is(exp.ToChar)
+        self.validate_identity("SELECT TO_CHAR(5555, '#')", "SELECT CAST(5555 AS VARCHAR)").selects[
+            0
+        ].assert_is(exp.Cast)
+        self.validate_identity(
+            "SELECT TO_CHAR(3.14, '#.#')", "SELECT CAST(3.14 AS VARCHAR)"
+        ).selects[0].assert_is(exp.Cast)
+        self.validate_identity(
+            "SELECT TO_CHAR(3.14, columnname)", "SELECT CAST(3.14 AS VARCHAR)"
+        ).selects[0].assert_is(exp.Cast)
 
     def test_time_diff(self):
         self.validate_identity("SELECT DATE_ADD(col, 1)")


sqlglot/dialects/dremio.py
```diff
@@ -100,7 +100,7 @@ class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "TO_CHAR": lambda args: (
-                exp.ToChar(this=args[0], format=args[1])
+                exp.Cast(this=args[0], to=exp.DataType.build("VARCHAR"))
```
georgesittas: I don't think we should treat `TO_CHAR` and casts to `VARCHAR` as equivalent in Dremio, i.e., roundtripping a `TO_CHAR` expression shouldn't result in a cast.

Doesn't the format (2nd argument) impose additional constraints compared to a simple cast? E.g., if you try to do `SELECT TO_CHAR(10, '#.#')`, doesn't that result in an error? Whereas the cast doesn't.

jasonthomassql: No, doesn't error but there is some rounding logic in the TO_CHAR that is not documented.  For example:

SELECT TO_CHAR(10, '#.#')  -> produces 10
SELECT TO_CHAR(10, '#') -> produces 10
SELECT TO_CHAR(10, '######') -> produces 10
SELECT TO_CHAR(10.507, '#')  -> rounds, produces 10
SELECT TO_CHAR(10.507, '#.##')  -> rounds, produces 10.50

I guess my concern here is this is all undocumented and found through trial and error.  From my perspective I just want the string and cast seems the better approach.  The rounding could be implemented if you think worthwhile.

jasonthomassql: Or are you asking to move the cast from parser to transforms?

georgesittas: If the two versions are not equivalent then we should preserve `TO_CHAR` for Dremio and then possibly try a best-effort transpilation if an exact one is not possible. Enabling transpilation to, e.g., Postgres, could happen via a new arg in `ToChar` that represents whether the value is a numeric or unknown, if that makes sense.

So if you had `SELECT TO_CHAR(10, '#')`, then you'd set that boolean flag at the AST level and in postgres you'd look it up and produce a `CAST` (+ a warning, since the cast, as mentioned, may not preserve the semantics 100%).

sqlglot/dialects/dremio.py
```diff
@@ -100,10 +100,13 @@ class Parser(parser.Parser):
...
                 if len(args) == 2 and (not isinstance(args[1], exp.Literal) or "#" in args[1].name)
                 else build_formatted_time(exp.TimeToStr, "dremio")(args)
             ),
+            "DATE_FORMAT": build_formatted_time(exp.TimeToStr, "dremio"),
+            "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "dremio"),
+            "TO_TIMESTAMP": build_formatted_time(exp.TimeStrToTime, "dremio"),
```
georgesittas: @jasonthomassql can we please revert these and add in a followup PR? Ideally the scope of each one should be the minimum possible.

sqlglot/dialects/dremio.py
jasonthomassql: I took a step back on this.  canonical formats will save a lot of target dialect changes...oracle, postgres, etc.  

This essentially matches the current oracle.py implementation.

tests/dialects/test_dremio.py
```diff
@@ -131,9 +131,31 @@ def test_time_mapping(self):
         )
 
     def test_to_char_special(self):
-        self.validate_identity("SELECT TO_CHAR(5555, '#')").selects[0].assert_is(exp.ToChar)
-        self.validate_identity("SELECT TO_CHAR(3.14, '#.#')").selects[0].assert_is(exp.ToChar)
+        self.validate_identity(
+            "SELECT TO_CHAR(5555, '#')", "SELECT TO_CHAR(5555, 'FM999999999999999999999999999999')"
+        ).selects[0].assert_is(exp.ToChar)
```
georgesittas: Can you please share docs related to this? Are these two equivalent in Dremio?

jasonthomassql: https://docs.dremio.com/current/reference/sql/sql-functions/functions/TO_CHAR/
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/TO_CHAR-number.html

I don't think there is a perfect equivalency as dremio treats the preceding # differently than any other dialect by not rounding it.  This will fix most transpiles and I can put together a reverse parser for spark in another pr as its to_char does not support FM9999 but does support 9999.

The general idea being that a target would only need to handle one standardized format.  Oracle being the most widely understood.

I can add the warning if you suggest that...just wasn't sure where to do that.

jasonthomassql: A perfect equivalency would be to convert to cast with an optional round if decimal format '#.##' is passed.  But you mentioned sticking with to_char so I did that.

georgesittas: I see the motivation, but I'm not sure if I agree with the current approach. The primary issue I see is that we're changing the format value for Dremio on roundtrip, which could lead to subtle issues, since AFAICT the FM* format is not even documented by Dremio.

We generally try to preserve the input SQL when roundtripping, because changing it has bitten us quite a few times. Only for obvious cases of equivalency we may not preserve the input, e.g. converting a `TO_TIMESTAMP(x)` into a `CAST(x AS TIMESTAMP)`, or something along those lines.

If it's possible to define a 1-1 and onto mapping between Dremio's `#` formats and some other, more general, format like Oracle's for example, then we could approach this similar to `TIME_MAPPING`. So, you'd canonicalize Dremio's formats at parse time using a trie, and then at generation time you'd use the inverse trie to ensure you get back the same format as the input.

We do something similar for time mappings: we map dialect-specific ones to strftime formats and then we do the inverse at generation time.

Appreciate the patience on this. The rationale behind these suggestions is to avoid circle back to this– we've encountered similar issues in the past and it's better to just get it right.

jasonthomassql: Understand what you are saying but what are you suggesting?  Are you suggesting we don't touch dremio but rather add a transform function used by all targets that converts "#.##" to a round(cast)?  Happy to implement whatever solution  you want.  I am just not clear what that solution is.

The TIME_MAPPING approach doesn't really work here.  The acceptable variations on # are infinite so a function is needed to reduce.

georgesittas: I'm suggesting two options:

1. Parse `TO_CHAR(x, '#.#')` as a simple `ToChar` node that contains an additional child: a boolean flag which, when `True`, means that the first argument was inferred to be a numeric value. You can then do a best-effort transpilation for postgres where you cast it into a varchar while warning about subtle differences compared to Dremio's behavior.
2. Implement a mapping similar to `TIME_MAPPING` that is 1-1 and onto and maps Dremio's formats into e.g. Oracle's or more general formats that represent the semantics. You can then do something similar to `format_time` (see dialect.py) to map the "canonical" formats into the proper dialect-specific ones at generation time.

The gist for (2) is really something similar to what we do `TIME_MAPPING`, but instead of time formats, you'd do it for these numeric formats.

jasonthomassql: Got it.  Went with #1.  Please review once again when you have time.  Thanks for your patience also.

Will add the postgres cast and warning in a subsequent pr to keep this contained.

PR #5589, opened 2025-08-16
feat(oracle): add NUMTODSINTERVAL and NUMTOYMINTERVAL functions

CosmicShoes:
Adds support for Oracle's NUMTODSINTERVAL and NUMTOYMINTERVAL functions for converting numbers to intervals.

### DIFF
--- a/sqlglot/dialects/oracle.py
+++ b/sqlglot/dialects/oracle.py
@@ -108,6 +108,8 @@ class Oracle(Dialect):
             **parser.Parser.FUNCTIONS,
             "CONVERT": exp.ConvertToCharset.from_arg_list,
             "NVL": lambda args: build_coalesce(args, is_nvl=True),
+            "NUMTODSINTERVAL": exp.NumTodsinterval.from_arg_list,
+            "NUMTOYMINTERVAL": exp.NumToyminterval.from_arg_list,
             "SQUARE": lambda args: exp.Pow(this=seq_get(args, 0), expression=exp.Literal.number(2)),
             "TO_CHAR": build_timetostr_or_tochar,
             "TO_TIMESTAMP": _build_to_timestamp,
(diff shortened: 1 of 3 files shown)


CosmicShoes:
Appropriate Oracle documentation:
- https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/NUMTODSINTERVAL.html
- https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/NUMTOYMINTERVAL.html

georgesittas:
@CosmicShoes what motivated this change?

CosmicShoes:
@georgesittas I encountered it while parsing a production DB schema and found that adding the missing definitions as per the documentation fixed my issue; I figured I would propose the change as it seemed within the scope of the project's maintenance goals.

georgesittas:
> @georgesittas I encountered it while parsing a production DB schema and found that adding the missing definitions as per the documentation fixed my issue; I figured I would propose the change as it seemed within the scope of the project's maintenance goals.

These are already parsed today in main, just not into a specific `Func` subclass that represents them:

```python
>>> import sqlglot
>>> sqlglot.parse_one("NUMTODSINTERVAL(100, 'HOUR')", "oracle")
Anonymous(
  this=NUMTODSINTERVAL,
  expressions=[
    Literal(this=100, is_string=False),
    Literal(this='HOUR', is_string=True)])
>>> sqlglot.parse_one("NUMTODSINTERVAL(100, 'HOUR')", "oracle").sql("oracle")
"NUMTODSINTERVAL(100, 'HOUR')"
```

If you're not looking to transpile these and you were only interested in parsing/generation, not sure if we need to add these new nodes now.

georgesittas:
Going to close this for the time being as I see no immediate use case for it besides parsing / generation. Happy to continue discussing.

PR #5597, opened 2025-08-18
Feat(dremio): handle `DATE_FORMAT`, `TO_DATE`, and `TO_TIMESTAMP`

jasonthomassql:
add some dremio parsers and tests for DATE_FORMAT, TO_DATE, and TO_TIMESTAMP

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -149,8 +149,6 @@ class Dremio(Dialect):
             exp.TimeToStr: lambda self, e: self.func("TO_CHAR", e.this, self.format_time(e)),
             exp.DateAdd: _date_delta_sql("DATE_ADD"),
             exp.DateSub: _date_delta_sql("DATE_SUB"),
-            exp.StrToDate: lambda self, e: self.func("TO_DATE", e.this, self.format_time(e)),
-            exp.StrToTime: lambda self, e: self.func("TO_TIMESTAMP", e.this, self.format_time(e)),
         }
 
         def datatype_sql(self, expression: exp.DataType) -> str:
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -116,7 +116,6 @@ class Dremio(Dialect):
             "TO_CHAR": to_char_is_numeric_handler,
             "DATE_FORMAT": build_formatted_time(exp.TimeToStr, "dremio"),
             "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "dremio"),
-            "TO_TIMESTAMP": build_formatted_time(exp.TimeStrToTime, "dremio"),
         }
 
     class Generator(generator.Generator):
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -7,6 +7,7 @@ from sqlglot import parser, generator, tokens
 from sqlglot.dialects.dialect import (
     Dialect,
     build_timetostr_or_tochar,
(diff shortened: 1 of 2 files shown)


tests/dialects/test_dremio.py
```diff
@@ -180,3 +180,40 @@ def test_time_diff(self):
         self.validate_identity(
             "SELECT DATE_SUB(col, a, 'HOUR')", "SELECT TIMESTAMPADD(HOUR, a * -1, col)"
         )
+
+
+def test_datetime_parsing(self):
```
georgesittas: Hey @jasonthomassql, not sure if this is correct. For example, I wouldn't expect the roundtrip for `TO_TIMESTAMP` to work, since if you produce a `TimeStrToTime` and don't specify a generator for it, you should get a `TIME_STR_TO_TIME` back.

I think this needs to be indented by one level.

jasonthomassql: I stripped some of this out for now as I am mostly just interested in dremio read.  I will add those transforms back in later when I go the other direction.

sqlglot/dialects/dremio.py
```diff
@@ -113,6 +114,9 @@ class Parser(parser.Parser):
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
             "TO_CHAR": to_char_is_numeric_handler,
+            "DATE_FORMAT": build_formatted_time(exp.TimeToStr, "dremio"),
+            "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "dremio"),
+            "TO_TIMESTAMP": build_formatted_time(exp.TimeStrToTime, "dremio"),
```
georgesittas: This doesn't look right & it's not covered by tests, so I'm removing it for now.
```suggestion
```

tests/dialects/test_dremio.py
```diff
@@ -180,3 +180,18 @@ def test_time_diff(self):
...
             "SELECT DATE_SUB(col, a, 'HOUR')", "SELECT TIMESTAMPADD(HOUR, a * -1, col)"
         )
+
+    def test_datetime_parsing(self):
+        ts = "CAST('2025-08-18 15:30:00' AS TIMESTAMP)"
+
+        self.validate_identity(
+            f"SELECT DATE_FORMAT({ts}, 'yyyy-mm-dd')",
+            f"SELECT TO_CHAR({ts}, 'yyyy-mm-dd')",
+        )
```
georgesittas, line 190: Let's not test unsupported features like this in the future. I'd rather we do a `validate_all` and read from `""` or some other dialect to make it clear that, e.g., `DATE_FORMAT` is not really supported by Dremio.

tests/dialects/test_dremio.py
```diff
@@ -180,3 +180,18 @@ def test_time_diff(self):
...
+
+        self.validate_all(
+            "SELECT DATE_FORMAT(CAST('2025-08-18 15:30:00' AS TIMESTAMP), 'yyyy-mm-dd')",
+            write={
+                "dremio": "SELECT TO_CHAR(CAST('2025-08-18 15:30:00' AS TIMESTAMP), 'yyyy-mm-dd')"
+            },
```
georgesittas:
```suggestion
            "SELECT DATE_FORMAT(CAST('2025-08-18 15:30:00' AS TIMESTAMP), 'yyyy-mm-dd')",
            "SELECT TO_CHAR(CAST('2025-08-18 15:30:00' AS TIMESTAMP), 'yyyy-mm-dd')",
```

PR #5598, opened 2025-08-18
databricks parse numeric to_char to cast

jasonthomassql:
Uses the dremio is_numeric arg to cast dremio to_char to strings.  Doing this to databricks for now as that is what I currently need. Can make a dialects.py utility function for this if that is preferred.

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -100,7 +100,7 @@ class Databricks(Spark):
             exp.JSONExtractScalar: _jsonextract_sql,
             exp.JSONPathRoot: lambda *_: "",
             exp.ToChar: lambda self, e: (
-                self.cast_sql(e.this, exp.DataType.build("STRING"))
+                self.cast_sql(exp.Cast(this=e.this, to=exp.DataType(this="STRING")))
                 if e.args.get("is_numeric")
                 else self.function_fallback_sql(e)
             ),
--- a/sqlglot/dialects/spark2.py
+++ b/sqlglot/dialects/spark2.py
@@ -347,8 +347,9 @@ class Spark2(Hive):
                 arg, (exp.JSONExtract, exp.JSONExtractScalar)
             ) and not arg.args.get("variant_extract")
 
-            # We can't use a non-nested type (eg. STRING) as a schema
-            if expression.to.args.get("nested") and (is_parse_json(arg) or is_json_extract):
+            # Make sure 'to' exists and has args before accessing 'nested'
+            to = getattr(expression, "to", None) or expression.args.get("to")
+            if to is not None and to.args.get("nested") and (is_parse_json(arg) or is_json_extract):
                 schema = f"'{self.sql(expression, 'to')}'"
                 return self.func("FROM_JSON", arg if is_json_extract else arg.this, schema)
 
--- a/tests/dialects/test_databricks.py
+++ b/tests/dialects/test_databricks.py
@@ -390,16 +390,14 @@ class TestDatabricks(Validator):
             """CREATE FUNCTION a() ENVIRONMENT (dependencies = '["foo1==1", "foo2==2"]', environment_version = 'None')"""
         )
 
+    def test_to_char_is_numeric_transpile_to_cast(self):
+        # The input SQL simulates a TO_CHAR with is_numeric flag set (from dremio dialect)
+        sql = "SELECT TO_CHAR(12345, '#')"
+        expression = parse_one(sql, read="dremio")
 
-def test_to_char_is_numeric_transpile_to_cast():
-    # The input SQL simulates a TO_CHAR with is_numeric flag set (from dremio dialect)
-    sql = "SELECT TO_CHAR(12345, '#')"
-    expression = parse_one(sql, read="dremio")
+        to_char_exp = expression.find(exp.ToChar)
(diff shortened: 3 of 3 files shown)


tests/dialects/test_databricks.py
```diff
@@ -385,3 +389,17 @@ def test_udf_environment_property(self):
         self.validate_identity(
             """CREATE FUNCTION a() ENVIRONMENT (dependencies = '["foo1==1", "foo2==2"]', environment_version = 'None')"""
         )
+
+
+def test_to_char_is_numeric_transpile_to_cast():
```
georgesittas: [Ditto](https://github.com/tobymao/sqlglot/pull/5597#discussion_r2282764370) about test needing to be indented.

sqlglot/dialects/databricks.py
```diff
@@ -99,7 +99,11 @@ class Generator(Spark.Generator):
             exp.JSONExtract: _jsonextract_sql,
             exp.JSONExtractScalar: _jsonextract_sql,
             exp.JSONPathRoot: lambda *_: "",
-            exp.ToChar: lambda self, e: self.function_fallback_sql(e),
+            exp.ToChar: lambda self, e: (
+                self.cast_sql(e.this, exp.DataType.build("STRING"))
+                if e.args.get("is_numeric")
+                else self.function_fallback_sql(e)
+            ),
```
georgesittas: I don't quite like this approach after all.. `is_numeric` is basically a Dremio-only arg right now, i.e., it's set only when `read="dremio"`. If we wanted to expand this functionality for Databricks or others, it could break roundtrip for them.

I believe the proper way to address this is to implement the 2nd option I described [here](https://github.com/tobymao/sqlglot/pull/5580#discussion_r2279104417). Apologies for the misguidance here.

We don't have to deal with this in this PR, I'm ok with merging this since it's still strictly better than without the PR being merged. If you don't have cycles to work on refactoring this, I'll probably take a stab at it soon, when I find some bandwidth.

jasonthomassql: Yes, I understand what you are saying.  Unfortunately I don't have the cycles atm and would just have to patch this locally.  I didn't use #2 because I don't think time_mapping is a good comparison.  It is a valid technique to standardize, in that case to python.  But, the variations of this format don't seem to be good candidates to a key-value mapping.  Hence why I went with a function and the canonical oracle format.  I would appreciate the merge and would look forward to how you do this as a learning opportunity.  This library is complicated and I am neck deep.  :)

jasonthomassql: In hindsight, perhaps a good middle ground would have been to add oracle_format instead of is_numeric to the to_char expression. That way each target could handle it if it supports translation.  But even that would be non-perfect as the #.# and FM9999 formats round differently.

georgesittas: > I would appreciate the merge and would look forward to how you do this as a learning opportunity. This library is complicated and I am neck deep. :)

Definitely! Appreciate your contributions thus far, don't mean to discourage you from continuing. We've intentionally dodged transpiling `TO_CHAR` until now because it's a bit annoying to deal with, so you're not dealing with the easiest part of the codebase.

> I don't think time_mapping is a good comparison. It is a valid technique to standardize, in that case to python.

To be fair, even though strftime covers most formats found in SQL datetime functions, it's still not enough. We've found some cases in the past where a SQL format couldn't map well to strftime, e.g. `A.M.` / `P.M.` can only map to `%p`, but doing this is lossy because you don't know what to map `%p` to.

To me the cleanest approach would be to introduce a SQLGlot-native format that represents the various values / semantics found in different dialects, sort of like an AST, but more simple because you probably just need an enum, with additional, implicit, semantics. This is what I was envisioning for the 2nd approach I mentioned in the other thread.

sqlglot/dialects/spark2.py
```diff
@@ -347,8 +347,9 @@ def cast_sql(self, expression: exp.Cast, safe_prefix: t.Optional[str] = None) ->
                 arg, (exp.JSONExtract, exp.JSONExtractScalar)
             ) and not arg.args.get("variant_extract")
 
-            # We can't use a non-nested type (eg. STRING) as a schema
-            if expression.to.args.get("nested") and (is_parse_json(arg) or is_json_extract):
+            # Make sure 'to' exists and has args before accessing 'nested'
+            to = getattr(expression, "to", None) or expression.args.get("to")
+            if to is not None and to.args.get("nested") and (is_parse_json(arg) or is_json_extract):
```
georgesittas, line 352: Why was this necessary?

jasonthomassql: The inheritance is maybe a bit messy here but makes some sense.  Might be worthwhile to remove one of the sparks at some point if possible.
Databricks.Generator
   ↓
Spark.Generator
   ↓
Spark2.Generator
   ↓
Hive.Generator

spark2.cast_sql expected a nested type, arg "to" and errored here if i tried to cast to STRING.  I could have overrode cast_sql in the databricks generator but it made more sense that the spark2.cast_sql just passes through non-nested if no "to" exists.

georgesittas: Didn't seem necessary: https://github.com/tobymao/sqlglot/commit/a4a299acbaf4461f0c2b470bc4e9e9590515eda7

Issue #5599, opened 2025-08-18
Documentation error: Missing import statement in error handling example


dududuguo:
# Bug Report Template

## Bug Description
**Brief summary of the issue:**
[Documentation parser-errors](https://github.com/tobymao/sqlglot?tab=readme-ov-file#parser-errors) shows incorrect import pattern for `sqlglot.errors.ParseError` that causes PyLance type checking errors.

**Actual behavior:**
PyLance reports `"errors" is not a known attribute of module "sqlglot"` when following the documentation example.

### Python code:
````python
import sqlglot
try:
    sqlglot.transpile("SELECT foo FROM (SELECT baz FROM t")
except sqlglot.errors.ParseError as e:  # PyLance error here
    print(e.errors)

# Working solution:
import sqlglot.errors  # Explicit import needed
try:
    sqlglot.transpile("SELECT foo FROM (SELECT baz FROM t")
except sqlglot.errors.ParseError as e:
    print(e.errors)
```

georgesittas:
Fixed by https://github.com/tobymao/sqlglot/commit/33898548f0051809b90ac0bb7692655b85afe6e2

PR #5600, opened 2025-08-18
fix ParseError example in README

dududuguo:
Fix Issue #5599 using working solution.


VaggelisD:
Hey @dududuguo, could you please share what the issue is here? Is it highlightinging the import in red or is there an actual error with it?

I tried that snippet in VSCode with Pylance enabled and I don't see either of these.

PR #5602, opened 2025-08-19
feat(singlestore): Added parsing and generation of date parts functions

AdalbertMemSQL:
Parsed and generated functions responsible for retrieval of date parts.
Documentation for SingleStore functions can be found [here](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/).

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -98,12 +98,12 @@ class SingleStore(MySQL):
                 format=MySQL.format_time(exp.Literal.string("%W")),
             ),
             "HOUR": lambda args: exp.cast(
-                    exp.TimeToStr(
-                        this=seq_get(args, 0),
-                        format=MySQL.format_time(exp.Literal.string("%k")),
-                    ),
-                    DataType.Type.INT,
+                exp.TimeToStr(
+                    this=seq_get(args, 0),
+                    format=MySQL.format_time(exp.Literal.string("%k")),
                 ),
+                DataType.Type.INT,
+            ),
             "MICROSECOND": lambda args: exp.cast(
                 exp.TimeToStr(
                     this=seq_get(args, 0),
@@ -130,8 +130,12 @@ class SingleStore(MySQL):
                 DataType.Type.INT,
             ),
             "WEEKDAY": lambda args: exp.Mod(
-                this=exp.Paren(this=exp.Add(this=exp.DayOfWeek(this=seq_get(args, 0)), expression=exp.Literal.number(5))),
-                expression=exp.Literal.number(7)
+                this=exp.Paren(
+                    this=exp.Add(
+                        this=exp.DayOfWeek(this=seq_get(args, 0)), expression=exp.Literal.number(5)
+                    )
+                ),
+                expression=exp.Literal.number(7),
             ),
             "DAY": lambda args: exp.Day(this=seq_get(args, 0)),
         }
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -10,6 +10,7 @@ from sqlglot.dialects.dialect import (
     rename_func,
 )
 from sqlglot.dialects.mysql import MySQL
+from sqlglot.expressions import DataType
 from sqlglot.generator import unsupported_args
 from sqlglot.helper import seq_get
 
@@ -50,7 +51,6 @@ class SingleStore(MySQL):
 
     class Parser(MySQL.Parser):
         FUNCTIONS = {
-            **MySQL.Parser.FUNCTIONS,
             "TO_DATE": build_formatted_time(exp.TsOrDsToDate, "singlestore"),
             "TO_TIMESTAMP": build_formatted_time(exp.StrToTime, "singlestore"),
             "TO_CHAR": build_formatted_time(exp.ToChar, "singlestore"),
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -50,7 +51,6 @@ class Tokenizer(MySQL.Tokenizer):
 
     class Parser(MySQL.Parser):
         FUNCTIONS = {
-            **MySQL.Parser.FUNCTIONS,
```
AdalbertMemSQL: This is done, because MySQL adds redundant casting for lots of time functions like
```
            "DAYOFYEAR": lambda args: exp.DayOfYear(this=exp.TsOrDsToDate(this=seq_get(args, 0))),
```

sqlglot/dialects/singlestore.py
```diff
@@ -50,7 +51,7 @@ class Tokenizer(MySQL.Tokenizer):
...
+            **parser.Parser.FUNCTIONS,
```
VaggelisD: Why was this changed? For any differences between SingleStore and MySQL we can override the corresponding entry

tests/dialects/test_singlestore.py
```diff
@@ -193,3 +193,60 @@ def test_json_extract(self):
...
         )
+
+    def test_date_parts_functions(self):
+        self.validate_identity(
+            "SELECT DAYNAME('2014-04-18')", "SELECT DATE_FORMAT('2014-04-18', '%W')"
+        )
+        self.validate_identity(
+            "SELECT HOUR('2009-02-13 23:31:30')",
+            "SELECT DATE_FORMAT('2009-02-13 23:31:30', '%k') :> INT",
+        )
```
VaggelisD, line 204: [Nit] IMO, it'd be better to turn these into `validate_all` if the "before" query is not valid SingleStore, e.g:

```Python3
self.validate_all(
 "SELECT DATE_FORMAT('2009-02-13 23:31:30', '%k')",
 read={
   "":"SELECT HOUR('2009-02-13 23:31:30')",
 }
)

```

tests/dialects/test_singlestore.py
```diff
@@ -193,3 +193,60 @@ def test_json_extract(self):
                 "mysql": 'SELECT JSON_VALUE(\'{"item": "shoes", "price": "49.95"}\', \'$.price\' RETURNING DECIMAL(4, 2))'
             },
         )
+
+    def test_date_parts_functions(self):
```
VaggelisD, line 197: Have you verified that the results would be equivalent between the base dialect / before queries and the  new SingleStore ones?

tests/dialects/test_singlestore.py
AdalbertMemSQL: before query is a valiad SingleStore.
SQLGlot doesn't have `exp.Hour` expression, so I converted `HOUR` to `DATE_FORMAT`

tests/dialects/test_singlestore.py
AdalbertMemSQL: Yep, I verified this

sqlglot/dialects/singlestore.py
AdalbertMemSQL: MySQL adds lots of redundant casting in time functions.
For example 
```
            "WEEKOFYEAR": lambda args: exp.WeekOfYear(this=exp.TsOrDsToDate(this=seq_get(args, 0))),
```.
If you prefer, I can override all these functions here.

Comment (VaggelisD) [reply sqlglot/dialects/singlestore.py]: I think overriding would be best, because otherwise we need to keep track of new MySQL additions and potentially move them to SingleStore.

If you are up for the challenge, you can open new PRs that improve MySQL directly (e.g removing these casts if they're truly redundant) so that these function entries are unified between the two.

Comment (AdalbertMemSQL) [reply sqlglot/dialects/singlestore.py]: I overrode function parsers here.
Probably, later I will do a PR changing the MySQL dialect.

**Comment by georgesittas** [comment]
@AdalbertMemSQL please remember to keep the scope of PRs minimal. Adding multiple functions or features in one pass makes it difficult to review the PR.

File: tests/dialects/test_singlestore.py [hunk 5]
```diff
@@ -193,3 +193,67 @@ def test_json_extract(self):
...
+        self.validate_identity(
+            "SELECT DAYNAME('2014-04-18')", "SELECT DATE_FORMAT('2014-04-18', '%W')"
+        )
```
georgesittas, line 200: How are these two equivalent? Isn't the `DATE_FORMAT` call going to crash due to the format not being compatible with the first argument? How does SingleStore know how to parse the datetime and extract the relevant part?

georgesittas:
I'm going to go ahead and close this; let's please do these functions one by one. Appreciate the PR regardless.

tests/dialects/test_singlestore.py
AdalbertMemSQL: SingleStore converts the first argument to the `DATE` type and then extracts day name from it
```
MySQL [(none)]> SELECT DATE_FORMAT('2014-04-18', '%W');
+---------------------------------+
| DATE_FORMAT('2014-04-18', '%W') |
+---------------------------------+
| Friday                          |
+---------------------------------+
1 row in set (0.047 sec)

```

georgesittas: Ok, as long as these auxiliary functions can be safely converted to `DATE_FORMAT`, that's fine. Just trying to make sure the two forms have equivalent semantics.

PR #5606, opened 2025-08-20
Use a valid SPDX identifier as license classifier

ecederstrand:
This helps automatic license checkers like pip-licenses to identify the right license for this project


VaggelisD (approved):
Thanks!

Leaving these for the other maintainers:

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license
https://spdx.org/licenses/MIT.html

PR #5608, opened 2025-08-20
Feat(parser)!: raise if query modifier is specified multiple times

### LINKED ISSUE #5604: Multiple WHERE clauses parses with no exceptions
**Before you file an issue**

```python
import sqlglot

stmt = """
    SELECT * FROM a 
        WHERE b = 'true'
        AND c > 50
        WHERE c = 'false'
    """

a = sqlglot.parse_one(stmt, read="postgres")


print(a.dump())
```
This produces the output ast of this form:
```json
{
  "class": "Select",
  "args": {
    "expressions": [
      {
        "class": "Star",
        "args": {}
      }
    ],
    "from": {
      "class": "From",
      "args": {
        "this": {
          "class": "Table",
          "args": {
            "this": {
              "class": "Identifier",
              "args": {
                "this": "a",
                "quoted": false
              }
            }
          }
        }
      }
    },
    "where": {
      "class": "Where",
      "args": {
        "this": {
          "class": "EQ",
          "args": {
            "this": {
              "class": "Column",
              "args": {
                "this": {
                  "class": "Identifier",
                  "args": {
                    "this": "c",
                    "quoted": false
                  }
                }
              }
            },
            "expression": {
              "class": "Literal",
              "args": {
                "this": "false",
                "is_string": true
              }
            }
          }
        }
      }
    }
  }
}
```
wherein the first where clause is _silently_ replaced by the second with no exceptions.  This behavior extrapolates out to any number of where clauses.

I could not find any official documentation that explicitly mentions that only a single WHERE clause may be specified for a select statement, however this does throw errors in postgres and all other databases that I have tested.

**Official Documentation**

georgesittas:
Fixes #5604

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -3533,10 +3533,25 @@ class Parser(metaclass=_Parser):
 
             while True:
                 if self._match_set(self.QUERY_MODIFIER_PARSERS, advance=False):
+                    modifier_index = self._index
+                    modifier_name = self._curr.text.upper()
+
                     parser = self.QUERY_MODIFIER_PARSERS[self._curr.token_type]
                     key, expression = parser(self)
 
                     if expression:
+                        if this.args.get(key):
+                            curr_index = self._index
+
+                            # Temporarily move the index back to highlight the modifier
+                            self._retreat(modifier_index)
+                            self.raise_error(f"Found multiple '{modifier_name}' clauses")
+
+                            # This resets it in case errors are being ignored. Not a common pattern,
+                            # but modifiers are used everywhere so making a bit more effort to ensure
+                            # things behave as expected in various code paths
+                            self._advance(curr_index - modifier_index)
+
                         this.set(key, expression)
                         if key == "limit":
                             offset = expression.args.pop("offset", None)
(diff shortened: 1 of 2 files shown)


sqlglot/parser.py
```diff
@@ -3533,10 +3533,25 @@ def _parse_query_modifiers(
...
                     parser = self.QUERY_MODIFIER_PARSERS[self._curr.token_type]
                     key, expression = parser(self)
 
                     if expression:
+                        if this.args.get(key):
+                            curr_index = self._index
+
+                            # Temporarily move the index back to highlight the modifier
+                            self._retreat(modifier_index)
+                            self.raise_error(f"Found multiple '{modifier_name}' clauses")
```
VaggelisD: Could we pass the token at `modifier_index` directly into `raise_error`, instead of moving back and forth?

VaggelisD (approved):

sqlglot/parser.py
georgesittas: Ah, good point. I forgot `raise_error` accepts a token 👍

Issue #5605, opened 2025-08-20
TSQL Union Parsing Incorrect


MQMMMQM:
parsed_sql = sqlglot.parse_one(sql, read="tsql")

sql = """
SELECT top 10 id FROM testtab1
union all
SELECT top 10 id FROM testtab2
order by id desc
"""

print(parsed_sql):

SELECT id FROM testtab1 LIMIT 10 UNION ALL SELECT id FROM testtab2 ORDER BY id DESC LIMIT 10

The semantics have been altered, and the actual result differs from the original SQL result.

georgesittas:
You need to set the output dialect, i.e. `parsed_sql.sql("tsql")`, or use `transpile`:

```python
>>> import sqlglot
>>> sqlglot.transpile("WITH t1 AS (SELECT 1 AS a), t2 AS (SELECT 1 AS a) SELECT TOP 10 a FROM t1 UNION ALL SELECT TOP 10 a FROM t2 ORDER BY a DESC", "tsql")
['WITH t1 AS (SELECT 1 AS a), t2 AS (SELECT 1 AS a) SELECT TOP 10 * FROM (SELECT TOP 10 a FROM t1 UNION ALL SELECT a FROM t2) AS _l_0 ORDER BY a DESC']
```

I want to double-check something regarding this output; if I find an issue I'll follow up separately. But this ticket is not valid, so I'll go ahead and close it.

PR #5609, opened 2025-08-20
Fix(tsql): do not attach limit modifier to set operation

georgesittas:
See https://github.com/tobymao/sqlglot/issues/5605#issuecomment-3205279961 for context.

We cannot attach `TOP` to the set operation, if any. It clearly binds to the query operand. Only `LIMIT`, `ORDER BY` etc are ambiguous.

### DIFF
--- a/tests/dialects/test_tsql.py
+++ b/tests/dialects/test_tsql.py
@@ -8,9 +8,11 @@ class TestTSQL(Validator):
     dialect = "tsql"
 
     def test_tsql(self):
-        self.validate_identity(
-            "with x as (select 1) select * from x union select * from x order by 1 limit 0",
+        self.validate_all(
             "WITH x AS (SELECT 1 AS [1]) SELECT TOP 0 * FROM (SELECT * FROM x UNION SELECT * FROM x) AS _l_0 ORDER BY 1",
+            read={
+                "": "WITH x AS (SELECT 1) SELECT * FROM x UNION SELECT * FROM x ORDER BY 1 LIMIT 0",
+            },
         )
 
         # https://learn.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms187879(v=sql.105)?redirectedfrom=MSDN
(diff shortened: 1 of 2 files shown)


tests/dialects/test_tsql.py
```diff
@@ -54,6 +59,10 @@ def test_tsql(self):
         self.validate_identity(
             "COPY INTO test_1 FROM 'path' WITH (FORMAT_NAME = test, FILE_TYPE = 'CSV', CREDENTIAL = (IDENTITY='Shared Access Signature', SECRET='token'), FIELDTERMINATOR = ';', ROWTERMINATOR = '0X0A', ENCODING = 'UTF8', DATEFORMAT = 'ymd', MAXERRORS = 10, ERRORFILE = 'errorsfolder', IDENTITY_INSERT = 'ON')"
         )
+        self.validate_identity(
+            "WITH t1 AS (SELECT 1 AS a), t2 AS (SELECT 1 AS a) SELECT TOP 10 a FROM t1 UNION ALL SELECT TOP 10 a FROM t2 ORDER BY a DESC",
+            "WITH t1 AS (SELECT 1 AS a), t2 AS (SELECT 1 AS a) SELECT * FROM (SELECT TOP 10 a FROM t1 UNION ALL SELECT TOP 10 a FROM t2) AS _l_0 ORDER BY a DESC",
+        )
```
VaggelisD, line 65: Somewhat orthogonal to this PR but I think that subqueries are not required for `ORDER BY`s as well, e.g:

```
tsql> WITH 
 t1 AS (SELECT 1 AS id),
 t2 AS (SELECT 2 AS id UNION ALL SELECT 3 AS id)
SELECT top 1 id FROM t1
union all
SELECT top 1 id FROM t2
order by id desc;
-----
2
1
```

It looks like the ordering is applied after the `UNION`, because if it was applied on the RHS the result would be `[3, 1]`, right?

VaggelisD (approved):

tests/dialects/test_tsql.py
georgesittas: Yeah, I'm not really sure why T-SQL sets `SET_OP_MODIFIERS`. Maybe an oversight?

VaggelisD: The motivation on the original PR was for Clickhouse only as far as I can tell, could be an oversight:

https://github.com/tobymao/sqlglot/pull/3301

georgesittas: Could you take a look at the `ORDER BY` etc? Not sure if it's going to be as simple as turning it off. I'll merge this for now.

PR #5613, opened 2025-08-20
feat(doris): update exp.UniqueKeyProperty SQL generation logic

xinge-ji:
update `exp.UniqueKeyProperty` SQL generation logic to support the `KEY` in materialized views
https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/async-materialized-view/CREATE-ASYNC-MATERIALIZED-VIEW

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -4030,7 +4030,8 @@ class Generator(metaclass=_Generator):
 
     # https://docs.starrocks.io/docs/sql-reference/sql-statements/table_bucket_part_index/CREATE_TABLE/
     def uniquekeyproperty_sql(self, expression: exp.UniqueKeyProperty) -> str:
-        return f"UNIQUE KEY ({self.expressions(expression, flat=True)})"
+        key_word = "UNIQUE KEY" if expression.args.get("unique", True) else "KEY"
+        return f"{key_word} ({self.expressions(expression, flat=True)})"
 
     # https://docs.starrocks.io/docs/sql-reference/sql-statements/data-definition/CREATE_TABLE/#distribution_desc
     def distributedbyproperty_sql(self, expression: exp.DistributedByProperty) -> str:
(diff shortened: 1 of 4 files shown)


sqlglot/generator.py
```diff
@@ -4030,7 +4030,8 @@ def duplicatekeyproperty_sql(self, expression: exp.DuplicateKeyProperty) -> str:
 
     # https://docs.starrocks.io/docs/sql-reference/sql-statements/table_bucket_part_index/CREATE_TABLE/
     def uniquekeyproperty_sql(self, expression: exp.UniqueKeyProperty) -> str:
-        return f"UNIQUE KEY ({self.expressions(expression, flat=True)})"
+        key_word = "UNIQUE KEY" if expression.args.get("unique", True) else "KEY"
+        return f"{key_word} ({self.expressions(expression, flat=True)})"
```
georgesittas, line 4034: Shouldn't `unique` be set in the parser somewhere? What's the purpose of this arg?

georgesittas (commented):
Let's add tests for this.

sqlglot/generator.py
xinge-ji: I added this in the Doris parser. The reason is that in Doris, the materialized view syntax only supports `KEY` (without `UNIQUE KEY`). To make it work in that context, I added the `unique` arg so that we can switch between generating `UNIQUE KEY` vs just `KEY`. The arg is optional and defaults to true, so existing behavior remains unchanged, and only Doris make use of the non-unique case.

georgesittas: I'm not sure I like introducing the `unique` arg for just this use case. I'll get this PR in and see if I can simplify it a bit...

georgesittas: Took a stab at it here: https://github.com/tobymao/sqlglot/pull/5625.

PR #5614, opened 2025-08-20
feat(doris): add support for BUILD and REFRESH properties in materialized view

xinge-ji:
https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/async-materialized-view/CREATE-ASYNC-MATERIALIZED-VIEW

### DIFF
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -66,6 +66,8 @@ class Doris(MySQL):
             "PROPERTIES": lambda self: self._parse_wrapped_properties(),
             "UNIQUE": lambda self: self._parse_composite_key_property(exp.UniqueKeyProperty),
             "PARTITION BY": lambda self: self._parse_partition_by_opt_range(),
+            "BUILD": lambda self: self._parse_build_property(),
+            "REFRESH": lambda self: self._parse_refresh_property(),
         }
 
         def _parse_partitioning_granularity_dynamic(self) -> exp.PartitionByRangePropertyDynamic:
@@ -128,6 +130,37 @@ class Doris(MySQL):
                 create_expressions=create_expressions,
             )
 
+        def _parse_build_property(self) -> exp.BuildProperty:
+            mode = self._parse_var(upper=True)
+            return self.expression(exp.BuildProperty, this=mode)
+
+        def _parse_refresh_property(self) -> exp.RefreshProperty:
+            method = self._parse_var(upper=True)
+
+            trigger = None
+            # MANUAL has no leading ON
+            if self._match_text_seq("MANUAL"):
+                trigger = self.expression(exp.RefreshTrigger, kind=exp.var("MANUAL"))
+            elif self._match_text_seq("ON"):
+                if self._match_text_seq("COMMIT"):
+                    trigger = self.expression(exp.RefreshTrigger, kind=exp.var("COMMIT"))
+                elif self._match_text_seq("SCHEDULE"):
+                    self._match_text_seq("EVERY")
+                    every = self._parse_number()
+                    unit = self._parse_var(any_token=True)
+                    starts = None
+                    if self._match_text_seq("STARTS"):
+                        starts = self._parse_string()
+                    trigger = self.expression(
+                        exp.RefreshTrigger,
+                        kind=exp.var("SCHEDULE"),
+                        every=every,
+                        unit=unit,
+                        starts=starts,
+                    )
+
+            return self.expression(exp.RefreshProperty, method=method, trigger=trigger)
+
     class Generator(MySQL.Generator):
         LAST_DAY_SUPPORTS_DATE_PART = False
         VARCHAR_REQUIRES_SIZE = False
@@ -146,6 +179,8 @@ class Doris(MySQL):
             exp.UniqueKeyProperty: exp.Properties.Location.POST_SCHEMA,
             exp.PartitionByRangeProperty: exp.Properties.Location.POST_SCHEMA,
             exp.PartitionedByProperty: exp.Properties.Location.POST_SCHEMA,
+            exp.BuildProperty: exp.Properties.Location.POST_SCHEMA,
+            exp.RefreshProperty: exp.Properties.Location.POST_SCHEMA,
         }
 
         CAST_MAPPING = {}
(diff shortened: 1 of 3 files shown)


sqlglot/expressions.py
```diff
@@ -2790,6 +2790,13 @@ class BackupProperty(Property):
     arg_types = {"this": True}
 
 
+# https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/async-materialized-view/CREATE-ASYNC-MATERIALIZED-VIEW/
+# Doris MATERIALIZED VIEW: BUILD <build_mode>
+class BuildProperty(Property):
+    # this -> build_mode: IMMEDIATE | DEFERRED
```
VaggelisD: [Ni] we can remove this comment or merge it with the one above it

sqlglot/expressions.py
```diff
@@ -3031,6 +3038,17 @@ class PartitionByRangePropertyDynamic(Expression):
     arg_types = {"this": False, "start": True, "end": True, "every": True}
 
 
+# https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/async-materialized-view/CREATE-ASYNC-MATERIALIZED-VIEW
+class RefreshTrigger(Expression):
+    # kind: MANUAL | COMMIT | SCHEDULE, every: int, unit: var, starts: string
```
VaggelisD: [Nit] We don't really add these comments, usually the doc will suffice

sqlglot/expressions.py
```diff
@@ -3031,6 +3038,17 @@ class PartitionByRangePropertyDynamic(Expression):
...
+    arg_types = {"kind": True, "every": False, "unit": False, "starts": False}
+
+
+class RefreshProperty(Property):
```
VaggelisD: Should we call this `RefreshTriggerProperty` to make this clearer? Also, can we leave a comment with the docs here too?

sqlglot/dialects/doris.py
```diff
@@ -128,6 +130,37 @@ def _parse_partition_by_opt_range(
                 create_expressions=create_expressions,
             )
 
+        def _parse_build_property(self) -> exp.BuildProperty:
+            mode = self._parse_var(upper=True)
+            return self.expression(exp.BuildProperty, this=mode)
```
VaggelisD: [Nit] We tend to inline these:

```suggestion
            mode = self._parse_var(upper=True)
            return self.expression(exp.BuildProperty, this=self._parse_var(upper=True))
```

sqlglot/dialects/doris.py
```diff
@@ -128,6 +130,37 @@ def _parse_partition_by_opt_range(
...
+                    starts = None
+                    if self._match_text_seq("STARTS"):
+                        starts = self._parse_string()
+                    trigger = self.expression(
+                        exp.RefreshTrigger,
+                        kind=exp.var("SCHEDULE"),
+                        every=every,
+                        unit=unit,
+                        starts=starts,
+                    )
```
VaggelisD: Ditto:


```suggestion
                    self._match_text_seq("EVERY")
                    trigger = self.expression(
                        exp.RefreshTrigger,
                        kind=exp.var("SCHEDULE"),
                        every=self._parse_number(),
                        unit=self._parse_var(any_token=True),
                        starts=self._match_text_seq("STARTS") and self._parse_string(),
                    )
```

VaggelisD (commented):
Hey @xinge-ji, thank you for the contribution, leaving a few minor comments to consider but overall it looks good:

sqlglot/expressions.py
xinge-ji: I’ve removed the redundant comment as suggested.

sqlglot/expressions.py
xinge-ji: I’ve removed the comment and kept the doc reference only.

sqlglot/expressions.py
xinge-ji: I’ve renamed it to `RefreshTriggerProperty` and added the doc reference

sqlglot/dialects/doris.py
xinge-ji: Updated to inline the call as suggested.

sqlglot/dialects/doris.py
xinge-ji: I’ve updated this to use the inlined version as suggested.

sqlglot/dialects/doris.py
```diff
@@ -128,6 +130,31 @@ def _parse_partition_by_opt_range(
...
+        def _parse_build_property(self) -> exp.BuildProperty:
+            return self.expression(exp.BuildProperty, this=self._parse_var(upper=True))
+
+        def _parse_refresh_property(self) -> exp.RefreshTriggerProperty:
+            method = self._parse_var(upper=True)
+
+            trigger = None
+            # MANUAL has no leading ON
+            if self._match_text_seq("MANUAL"):
+                trigger = self.expression(exp.RefreshTrigger, kind=exp.var("MANUAL"))
```
VaggelisD: Is that right? This is the syntax spec from the docs:

```
refresh_trigger
  : ON MANUAL
  | ON SCHEDULE EVERY <int_value> <refresh_unit> [ STARTS '<start_time>']
  | ON COMMIT
```

sqlglot/dialects/doris.py
```diff
@@ -724,3 +753,28 @@ def table_sql(self, expression: exp.Table, sep: str = " AS ") -> str:
...
+            return f"BUILD {self.sql(expression, 'this')}"
+
+        def refreshtrigger_sql(self, expression: exp.RefreshTrigger) -> str:
+            kind = expression.args.get("kind")
+            if not kind:
+                return ""
+            kind_text = kind.name
+            if kind_text == "MANUAL":
+                # MANUAL has no ON prefix
+                return "MANUAL"
```
VaggelisD: Ditto about `ON MANUAL`

sqlglot/expressions.py
```diff
@@ -3031,6 +3036,16 @@ class PartitionByRangePropertyDynamic(Expression):
     arg_types = {"this": False, "start": True, "end": True, "every": True}
 
 
+# https://doris.apache.org/docs/sql-manual/sql-statements/table-and-view/async-materialized-view/CREATE-ASYNC-MATERIALIZED-VIEW
+class RefreshTrigger(Expression):
+    arg_types = {"kind": True, "every": False, "unit": False, "starts": False}
+
```
VaggelisD: I wonder whether we want this as a separate node instead of storing everything in `RefreshTriggerProperty`. We generally avoid creating new nodes if possible, reduces the noise across several components.

sqlglot/dialects/doris.py
VaggelisD: Ah, I see that the docs are confusing because at a later point this is the clause:

```
4. <refresh_trigger>

Trigger method:

MANUAL: Manual refresh.

ON SCHEDULE: Scheduled refresh.

ON COMMIT: Triggered refresh, where changes to the base table data trigger a refresh of the materialized view.
```

sqlglot/expressions.py
xinge-ji: ok, I’m storing everything in `RefreshTriggerProperty`.

sqlglot/dialects/doris.py
xinge-ji: I’ve applied update for `ON MANUAL`.

sqlglot/dialects/doris.py
xinge-ji: Yes, ON MANUAL is the correct form. It’s also used in other parts of the docs (e.g. [usage scenarios](https://doris.apache.org/docs/query-acceleration/materialized-view/async-materialized-view/use-guide#usage-scenarios)). I’ve updated the parser accordingly.

VaggelisD (approved):
Got a couple nits but I'll go ahead and merge this one, thank you!

VaggelisD:
https://github.com/tobymao/sqlglot/pull/5648/files

PR #5622, opened 2025-08-21
Update type mapping for redshift when generating SQL

lewisbirks:
I noticed a few issues when mapping between TSQL and Redshift which resulted in transpiled SQL that didn't match up with what Redshift supported.

I've updated a the data type conversions that I noticed, there might be others and there could be more that needs to be done but I am new to Redshift, this library, and Python so any help or pointers would be appreciated if I have missed anything.


sqlglot/dialects/redshift.py
```diff
@@ -166,15 +166,20 @@ class Generator(Postgres.Generator):
...
         WITH_PROPERTIES_PREFIX = " "
 
+        # https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html
         TYPE_MAPPING = {
             **Postgres.Generator.TYPE_MAPPING,
+            exp.DataType.Type.BIT: "BOOLEAN",
             exp.DataType.Type.BINARY: "VARBYTE",
             exp.DataType.Type.BLOB: "VARBYTE",
             exp.DataType.Type.INT: "INTEGER",
+            exp.DataType.Type.TINYINT: "SMALLINT",  # The smallest numeric data type in Redshift is SMALLINT
```
georgesittas: This is redundant, Postgres, from which Redshift inherits from, [already contains the entry](https://github.com/tobymao/sqlglot/blob/54623a6b85432272703f12a197b05ced78529f90/sqlglot/dialects/postgres.py#L573).

sqlglot/dialects/redshift.py
```diff
@@ -166,15 +166,20 @@ class Generator(Postgres.Generator):
         # Redshift doesn't have `WITH` as part of their with_properties so we remove it
         WITH_PROPERTIES_PREFIX = " "
 
+        # https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html
         TYPE_MAPPING = {
             **Postgres.Generator.TYPE_MAPPING,
+            exp.DataType.Type.BIT: "BOOLEAN",
```
georgesittas, line 172: Are you sure this is correct? The following is a valid Redshift query:

```sql
SELECT CAST(1 AS bit)
```

I can't find documentation around this type, but Postgres [supports it](https://www.postgresql.org/docs/current/datatype-bit.html) for example.

Perhaps we should revert this to be on the safe side. What motivated this to begin with?

sqlglot/dialects/redshift.py
```diff
@@ -166,15 +166,20 @@ class Generator(Postgres.Generator):
...
             exp.DataType.Type.BINARY: "VARBYTE",
             exp.DataType.Type.BLOB: "VARBYTE",
             exp.DataType.Type.INT: "INTEGER",
+            exp.DataType.Type.TINYINT: "SMALLINT",  # The smallest numeric data type in Redshift is SMALLINT
+            exp.DataType.Type.UTINYINT: "SMALLINT",
             exp.DataType.Type.TIMETZ: "TIME",
             exp.DataType.Type.TIMESTAMPTZ: "TIMESTAMP",
             exp.DataType.Type.VARBINARY: "VARBYTE",
             exp.DataType.Type.ROWVERSION: "VARBYTE",
+            exp.DataType.Type.UUID: "CHAR(36)",  # There is no equivalent UUID type so just use the 36 chars directly
```
georgesittas, line 181: Why 36 specifically?

sqlglot/dialects/redshift.py
lewisbirks: Odd, I could have sworn I ran a test where it didn't convert, I'll double check that

sqlglot/dialects/redshift.py
lewisbirks: This was the one mapping that motivated this PR 😄
At work I've landed on a project which requires a migration from TSQL to Redshift and I found this library to help transpile the SQL (looking to replace a find and replace script).

I did not see BIT support mentioned in the available data types https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html hence why I thought BOOLEAN was the correct alternative.

sqlglot/dialects/redshift.py
lewisbirks: That is the number of characters in a UUID

georgesittas: Ah, right. Thanks for clarifying.

sqlglot/dialects/redshift.py
georgesittas: I see– let's do some more investigation around this to ensure we won't break Redshift's semantics with this mapping. I'm not sure if `BIT` and `BOOLEAN` are interchangeable.

sqlglot/dialects/redshift.py
lewisbirks: oh, it was the unsigned version which was required (it doesn't appear that redshift has any unsigned support). As you mention the signed version is already converted properly so I'll revert that change.

sqlglot/dialects/redshift.py
lewisbirks: That is my concern around that change, how it will interact with `1` or `0`, I don't have access to a Redshift instance yet, but I'll try to test it somehow 🙂

georgesittas: Yeah, testing on Redshift directly is important, regardless.

georgesittas:
Hey @lewisbirks, any plans to finish this?

lewisbirks:
@georgesittas I do intend to, but I'm still trying to get access to redshift to test. If you want it can be closed until I get access as I'm not sure when that will be

georgesittas:
Sounds good, feel free to reopen when ready. Thank you for following up 👍

lewisbirks:
@georgesittas I've finally got access to a redshift instance, these are the tests I've done so far

image
image
image

is there anything else you would want me to test?

georgesittas:
I just tested this in Redshift and it failed, so I don't think we can treat them as equivalent:

```sql
-- ERROR: cannot cast type boolean to bit
select cast(true as bit)
```

lewisbirks:
@georgesittas fair enough, we can leave this closed then

georgesittas:
Thank you for following up!

PR #5625, opened 2025-08-21
Chore(doris): refactor unique key prop generation

georgesittas:


sqlglot/generator.py
```diff
@@ -1670,8 +1670,14 @@ def properties_sql(self, expression: exp.Properties) -> str:
             elif p_loc == exp.Properties.Location.POST_SCHEMA:
                 root_properties.append(p)
 
-        root_props = self.root_properties(exp.Properties(expressions=root_properties))
-        with_props = self.with_properties(exp.Properties(expressions=with_properties))
+        root_props_ast = exp.Properties(expressions=root_properties)
+        root_props_ast.parent = expression.parent
```
VaggelisD, line 1674: Is there any scenario / odd case in which we forget to link the parent at some place and `UNIQUE` gets silently dropped? Is this the only place we need it or is our test coverage guaranteeing this would be caught?

sqlglot/dialects/doris.py
```diff
@@ -670,6 +664,15 @@ class Generator(MySQL.Generator):
             "year",
         }
 
+        def uniquekeyproperty_sql(
+            self, expression: exp.UniqueKeyProperty, prefix: str = "UNIQUE KEY"
+        ) -> str:
+            create_stmt = expression.find_ancestor(exp.Create)
```
VaggelisD, line 670: Wouldn't this be the direct ancestor?

VaggelisD (approved):

sqlglot/generator.py
georgesittas: No, the reason my approach didn't work was because the parent link got silently dropped due to `Properties(expressions=...)`. This is a one-off thing that is only relevant to the generator and is meant to enable stuff like what I did for the unique property.

sqlglot/dialects/doris.py
georgesittas: Chose not to make assumptions in this case to make things simpler

Issue #5618, opened 2025-08-21
TSQL UNION with ORDER BY and there is a nested layer the parsed statement has issues


MQMMMQM:
>>> import sqlglot
>>> sql="""
WITH
t1 AS (SELECT 1 AS a),
t2 AS (SELECT 1 AS a)
select count(*) from (
SELECT TOP 10 a FROM t1
UNION ALL 
SELECT TOP 10 a FROM t2
ORDER BY a DESC
) as t
"""
>>> parsed_sql = sqlglot.parse_one(sql, read="tsql")
>>> print(parsed_sql)
WITH t1 AS (SELECT 1 AS a), t2 AS (SELECT 1 AS a) SELECT COUNT(*) FROM (SELECT a FROM t1 LIMIT 10 UNION ALL SELECT a FROM t2 LIMIT 10 ORDER BY a DESC) AS t
>>> print(parsed_sql.sql(dialect="tsql"))
WITH t1 AS (SELECT 1 AS a), t2 AS (SELECT 1 AS a) SELECT COUNT(*) FROM (SELECT * FROM (SELECT TOP 10 a FROM t1 UNION ALL SELECT TOP 10 a FROM t2) AS _l_0 ORDER BY a DESC) AS t

Because the ORDER BY clause is promoted to the outer layer, but there is still an additional outermost layer without a TOP clause, the final SQL execution will result in an error.
SQL error [1033] [S0001]: The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.

MQMMMQM:
After applying [#5626](https://github.com/tobymao/sqlglot/pull/5626)

When a limit is added to "SELECT id FROM testtab AS t UNION ALL SELECT id FROM testtab2 ORDER BY id"
it is parsed into the following content:
"SELECT TOP 100 * FROM (SELECT id FROM testtab AS t UNION ALL SELECT id FROM testtab2 ORDER BY id) AS _l_0"
Then an error occurs again; Can't we have the best of both worlds?

georgesittas:
What's the exact query you're parsing/generating? Can you show the code you're running?

MQMMMQM:
```
import sqlglot
from sqlglot import expressions as exp

sql="""
SELECT id FROM testtab AS t UNION ALL SELECT id FROM testtab2 ORDER BY id
"""

limit_sql = parsed_sql.limit(
    exp.Literal(
[...]

PR #5626, opened 2025-08-21
fix(tsql): Remove ORDER from set op modifiers too

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5618

I left `offset` in the modifiers due to the following scenario coming from base dialect:

```Python3
sql = "SELECT * FROM a UNION SELECT * FROM b ORDER BY x LIMIT 1"
```

<br/>

- ❌ Without any `SET_OP_MODIFIERS`, T-SQL query can't execute due to error `Invalid usage of the option FIRST in the FETCH statement`:
```Python3
>>> sqlglot.parse_one(sql, dialect="").sql("tsql")
"WITH t1 AS (SELECT 1 AS col), t2 AS (SELECT 2 AS col) 
SELECT * FROM t1 UNION SELECT * FROM t2 ORDER BY col FETCH FIRST 1 ROWS ONLY"
```

<br/>

- ✅ With `SET_OP_MODIFIERS = {"offset"}`, T-SQL query executes fine because `LIMIT` is lifted as a `TOP`
```Python3
>>> sqlglot.parse_one(sql, dialect="").sql("tsql")
"WITH t1 AS (SELECT 1 AS col), t2 AS (SELECT 2 AS col) 
SELECT TOP 1 * FROM (SELECT * FROM t1 UNION SELECT * FROM t2) AS _l_0 ORDER BY col"
```

### DIFF
--- a/sqlglot/dialects/tsql.py
+++ b/sqlglot/dialects/tsql.py
@@ -658,7 +658,7 @@ class TSQL(Dialect):
             else self.expression(exp.ScopeResolution, this=this, expression=to),
         }
 
-        SET_OP_MODIFIERS = {"order", "offset"}
+        SET_OP_MODIFIERS = {"offset"}
 
         def _parse_alter_table_set(self) -> exp.AlterSet:
             return self._parse_wrapped(super()._parse_alter_table_set)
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
I wonder if there's an issue with the current fetch generation.. I think `FETCH FIRST` in T-SQL requires `OFFSET` to be present for it to be valid. Perhaps this is a gap with the set operation generator? I.e., the fact that we don't convert `Limit` to `TOP` syntax, similar to `select_sql`.

VaggelisD:
@georgesittas  I think what you're saying makes sense, though I entertained the idea of trying to fix it and the `TOP` vs `FETCH` situation is somewhat convoluted so always had a couple tests failing.

I think I'll merge this one and we can reconsider the refactor in a follow up PR

georgesittas:
Sounds good

PR #5631, opened 2025-08-21
feat(singelstore): Removed redundant casting for DAY function

AdalbertMemSQL:
MySQL dialects adds a redundant cast inside the DAY function.
```
            "DAY": lambda args: exp.Day(this=exp.TsOrDsToDate(this=seq_get(args, 0))),
```
Removed it in SingleStore.

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -136,6 +136,7 @@ class SingleStore(MySQL):
                 this=seq_get(args, 0),
                 format=MySQL.format_time(exp.Literal.string("%W")),
             ),
+            "DAY": lambda args: exp.Day(this=seq_get(args, 0)),
         }
 
         CAST_COLUMN_OPERATORS = {TokenType.COLON_GT, TokenType.NCOLON_GT}
--- a/tests/dialects/test_singlestore.py
+++ b/tests/dialects/test_singlestore.py
@@ -233,4 +233,8 @@ class TestSingleStore(Validator):
                 "singlestore": "SELECT DAY('2014-04-18')",
                 "": "SELECT DAY_OF_MONTH('2014-04-18')",
             },
+            write={
+                "singlestore": "SELECT DAY('2014-04-18')",
+                "": "SELECT DAY('2014-04-18')",
+            },
         )


georgesittas:
The `TsOrDsToDate` has additional semantics, I don't think it's redundant. It means that, when transpiling to other dialects, a cast to `DATE` will be added. It's done because some dialects only accept proper `DATE` values and not strings.

Why remove it for SingleStore? Isn't the roundtrip an identity, given that it inherits from MySQL?

AdalbertMemSQL:
> The `TsOrDsToDate` has additional semantics, I don't think it's redundant. It means that, when transpiling to other dialects, a cast to `DATE` will be added. It's done because some dialects only accept proper `DATE` values and not strings.
> 
> Why remove it for SingleStore? Isn't the roundtrip an identity, given that it inherits from MySQL?

That's a good point.
Can you point out dialects that accept only `DATE` values and not strings? 
Handling of such scenarios in SQLGlot appears to be inconsistent. For example, Snowflake has the same functions and they can accept strings https://docs.snowflake.com/en/sql-reference/functions/year . But the snowflake dialect doesn't have any additional `TsOrDsToDate`.

Also, at the moment, SingleStore dialect translates `TsOrDsToDate` to `TO_DATE` function, which requires `format` argument and will break without it.

georgesittas:
>  Can you point out dialects that accept only DATE values and not strings?

IIRC, two such examples are Presto and Trino. There are probably more.

> Handling of such scenarios in SQLGlot appears to be inconsistent.

Yeah, that wouldn't surprise me. There probably wasn't a transpilation need for dialects like Snowflake, so it hasn't come up. However, since MySQL, and hence SingleStore, already have logic for `Day`, I'm slightly hesitant to get rid of that logic for the latter.

> Also, at the moment, SingleStore dialect translates TsOrDsToDate to TO_DATE function, which requires format argument and will break without it.

It doesn't matter, observe for example how it behaves in main today:

```python
>>> import sqlglot
>>> sqlglot.transpile("day(x)", "singlestore")
['DAY(x)']
```

I'll go ahead and close this PR, since there seems to be a misunderstanding. Happy to continue chatting, regardless.

AdalbertMemSQL:
```
It doesn't matter, observe for example how it behaves in main today:

>>> import sqlglot
>>> sqlglot.transpile("day(x)", "singlestore")
['DAY(x)']
```
This happens because MySQL generator removes the cast for `exp.Day`.
```
            exp.Day: _remove_ts_or_ds_to_date(),
```

But it seems like something that can be easily broken. I'm going to open another PR where I will update SingleStore handling of `exp.TsOrDsToDate` to work in cases when the `format` is not provided. Seems that in other dialects, this case is handled 
```
def _ts_or_ds_to_date_sql(self: MySQL.Generator, expression: exp.TsOrDsToDate) -> str:
    time_format = expression.args.get("format")
    return _str_to_date_sql(self, expression) if time_format else self.func("DATE", expression.this)
```

georgesittas:
Yeah, that is reasonable 👍

PR #5637, opened 2025-08-21
Netezza dialect: DISTRIBUTE/ORGANIZE, EXTERNAL USING options, NVL/DEC…

robertvirany:
…ODE preservation, type mappings (NVARCHAR/NCHAR/TIMESTAMPTZ/TIMETZ/JSON), uppercase normalization, index suppression, and comprehensive tests

### DIFF
--- a/sqlglot/dialects/__init__.py
+++ b/sqlglot/dialects/__init__.py
@@ -78,6 +78,7 @@ DIALECTS = [
     "Fabric",
     "Hive",
     "Materialize",
+    "Netezza",
     "MySQL",
     "Oracle",
     "Postgres",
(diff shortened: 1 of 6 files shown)


georgesittas:
Hey,
Thank you for your contribution.

Unfortunately, we don't plan on merging new dialects for the time being. The reason is that the team's bandwidth is quite low and we can't take on additional maintenance burden, even if it's just for reviewing. We'll revisit this PR if we find bandwidth.

Some feedback in good faith: I would encourage you to reduce the scope of future PRs, so that they're bite-sized and can be easily reviewed by the maintainers of the project. This one's too much for an initial dialect implementation. You can find similar PRs that introduced Dremio and Exasol if you search in the project's history.

robertvirany:
No worries, I could have made it more bite-sized. If a future maintainer looks at it I used this for reference on the dialect: https://www.ibm.com/docs/en/netezza?topic=dud-netezza-performance-server-sql-command-reference

PR #5639, opened 2025-08-22
feat(singlestore): Changed generation of exp.TsOrDsToDate to handle case when format is not provided

AdalbertMemSQL:


sqlglot/dialects/singlestore.py
```diff
@@ -171,7 +171,9 @@ class Generator(MySQL.Generator):
 
         TRANSFORMS = {
             **MySQL.Generator.TRANSFORMS,
-            exp.TsOrDsToDate: lambda self, e: self.func("TO_DATE", e.this, self.format_time(e)),
+            exp.TsOrDsToDate: lambda self, e: self.func("TO_DATE", e.this, self.format_time(e))
+            if e.args.get("format")
+            else self.func("DATE", e.this),
```
georgesittas, line 176: I just noticed that `format_time` returns `None` if there's no format in `e`, which then gets filtered out by `func` and you end up with the correct form:

```python
>>> import sqlglot
>>> sqlglot.exp.TsOrDsToDate(this=sqlglot.exp.Literal.string("2020-01-01")).sql("singlestore")
"TO_DATE('2020-01-01')"
```

So this code is not necessary.

georgesittas: Whoops, I just saw that the alternative is `DATE`, not `TO_DATE` 🤦

AdalbertMemSQL: Yep, `TO_DATE` function in SingleStore requires 2 arguments when `DATE` works with only one.

PR #5640, opened 2025-08-22
feat(singlestore): Marked exp.All as unsupported

AdalbertMemSQL:
SingleStore supports only [EXISTS](https://docs.singlestore.com/cloud/reference/sql-reference/data-manipulation-language-dml/exists-and-not-exists/) subquery predicate.

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -1324,3 +1324,7 @@ class SingleStore(MySQL):
                 res = exp.Cast(this=res, to=returning)
 
             return self.sql(res)
+
+        def all_sql(self, expression: exp.All) -> str:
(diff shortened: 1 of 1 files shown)


AdalbertMemSQL:
I'm unsure what the proper approach is for handling unsupported expressions. One idea I had was to override the existing methods, but I'd like to get your thoughts. Can we discuss this in the PR?

VaggelisD:
Hey @AdalbertMemSQL, what you have here would be our approach too more than likely. In general, we treat transpilations as a best-effort attempt so by default we emit warnings and move on.

Could we transform `ALL` into `NOT EXISTS` + the reverse binary operators? Doesn't have to be done now, but it could be a solution.

georgesittas (approved):
Agree with Vaggelis. If there's a path forward for transpiling `ALL` we can do it in a followup.

PR #5641, opened 2025-08-22
fix(generator): Wrap CONCAT items with COALESCE less aggressively

VaggelisD:
SQLGlot employs a transformation to transpile `CONCAT(...)` into `CONCAT(COALESCE(..., ''))` between dialects with different `NULL` semantics, e.g:

```SQL
# DuckDB: Evaluates to empty string
duckdb> SELECT CONCAT(NULL) = '';
┌─────────────────────┐
│ (concat(NULL) = '') │
│       boolean       │
├─────────────────────┤
│ true                │
└─────────────────────┘

# Spark: Evaluates to `NULL`:
spark-sql > SELECT CONCAT(NULL) IS NULL;
true

# Spark: Wrapped with COALESCE evalutes to empty string
spark-sql > SELECT CONCAT(COALESCE(NULL, '')) = '';
(concat(coalesce(NULL, )) = )
true
```

<br/>

This transformation already happens on main today, but is done so preemptively:

```Python3
>>> sqlglot.parse_one("SELECT CONCAT(NULL)", dialect="duckdb").sql("spark")
"SELECT CONCAT(COALESCE(NULL, ''))"
```

<br/>

This PR relaxes the `COALESCE` wrapping (since as the issue shows it can lead to errors) by only doing so if the expression _could_ be evaluated to `NULL`.

### DIFF
--- a/sqlglot/generator.py
+++ b/sqlglot/generator.py
@@ -2998,7 +2998,14 @@ class Generator(metaclass=_Generator):
             args = [exp.cast(e, exp.DataType.Type.TEXT) for e in args]
 
         if not self.dialect.CONCAT_COALESCE and expression.args.get("coalesce"):
-            args = [exp.func("coalesce", e, exp.Literal.string("")) for e in args]
+            # Wrap with COALESCE if the arg could be NULL i.e is already exp.Null or an expression with columns
+            # that could be evaluated to NULL
+            args = [
+                exp.func("coalesce", e, exp.Literal.string(""))
+                if (isinstance(e, exp.Null) or e.find(exp.Column))
+                else e
+                for e in args
+            ]
 
         return args
 
--- a/tests/dialects/test_dremio.py
+++ b/tests/dialects/test_dremio.py
@@ -40,7 +40,7 @@ class TestDremio(Validator):
             "SELECT CONCAT('a', NULL)",
             write={
                 "dremio": "SELECT CONCAT('a', NULL)",
-                "": "SELECT CONCAT(COALESCE('a', ''), COALESCE(NULL, ''))",
+                "": "SELECT CONCAT('a', COALESCE(NULL, ''))",
             },
         )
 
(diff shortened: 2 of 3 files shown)


sqlglot/generator.py
```diff
@@ -2998,7 +2998,14 @@ def convert_concat_args(self, expression: exp.Concat | exp.ConcatWs) -> t.List[e
             args = [exp.cast(e, exp.DataType.Type.TEXT) for e in args]
 
         if not self.dialect.CONCAT_COALESCE and expression.args.get("coalesce"):
-            args = [exp.func("coalesce", e, exp.Literal.string("")) for e in args]
+            # Wrap with COALESCE if the arg could be NULL i.e is already exp.Null or an expression with columns
+            # that could be evaluated to NULL
+            args = [
+                exp.func("coalesce", e, exp.Literal.string(""))
+                if (isinstance(e, exp.Null) or e.find(exp.Column))
```
georgesittas: This condition doesn't seem complete, e.g. some function calls return `NULL` when they get invalid arguments too. I'm confident `NULL` can show up in many cases.

Isn't the motivating issue related to typing? Seems like DuckDB simply overloads `CONCAT` to work with both strings and arrays and our default value assumes `CONCAT` only works with strings, so when it's transfered over to Spark, it chokes because it mixes arrays (first `COALESCE` arg) and strings (fallback value).

VaggelisD: Yeah, I was split between going the typing route vs this one, but at the same time I also observed that e.g we would produce `COALESCE` even if we had literal strings which seemed unnecessary.

I can switch to simply checking if the args are of ARRAY type plus check the string literal case for now

sqlglot/generator.py
```diff
@@ -2998,7 +2998,19 @@ def convert_concat_args(self, expression: exp.Concat | exp.ConcatWs) -> t.List[e
...
         if not self.dialect.CONCAT_COALESCE and expression.args.get("coalesce"):
-            args = [exp.func("coalesce", e, exp.Literal.string("")) for e in args]
+
+            def _wrap_with_coalesce(e: exp.Expression) -> exp.Expression:
+                if not e.type:
+                    from sqlglot.optimizer.annotate_types import annotate_types
+
+                    e = annotate_types(e, dialect=self.dialect)
+
+                if e.is_type(exp.DataType.Type.ARRAY) or e.is_string:
```
georgesittas:
```suggestion
                if e.is_string or e.is_type(exp.DataType.Type.ARRAY):
```

georgesittas (approved):

Issue #5643, opened 2025-08-22
Invalid conversion Spark->DuckDB for function FIRST() with ignore_nulls set


michal-clutch:
In current conversion when there is `True` to ignore nulls set in spark function `first()` it's translated to DuckDB `first()` function which doesn't support ignoring nulls. There is another function which acts like that (ignore nulls). It's `ANY_VALUE()`

Description from DuckDB docs:
```
Description	Returns the first non-NULL value from arg. This function is [affected by ordering](https://duckdb.org/docs/stable/sql/functions/aggregates.html#order-by-clause-in-aggregate-functions).
```

**Current output**
```
IGNORE NULLS is not supported for non-window functions.
SELECT
  FIRST(sample_col)
```

**Expected output**
```
SELECT
  ANY_VALUE(sample_col)
```

**Fully reproducible code snippet**
```
    import sqlglot
    query = """
    SELECT FIRST(sample_col, TRUE)
    """

    print(sqlglot.transpile(query, read="spark", write="duckdb", pretty=True)[0])
```

**Official Documentation**
(https://duckdb.org/docs/stable/sql/functions/aggregates.html#any_valuearg)

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.first.html

georgesittas:
This is low priority for the core team– feel free to submit a well-tested PR and we'll be happy to get that in. Thanks!

VaggelisD:
Hey @michal-clutch, I'll go ahead and resolve this one but as George said we'd really appreciate it if you could contribute these fixes.

michal-clutch:
@VaggelisD @georgesittas thanks again for sorting this out. I'm not familiar how the lib works under the hood so it could be difficult for me to propose working solution. I thought it would benefit everybody if I raise an issue when I spot something which could be fixed within translations. Tried to provide as clearly described issues as possible so it's easier for you.

PR #5644, opened 2025-08-22
fix(duckdb): Transpile Spark's `FIRST(col, TRUE)`

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5643

```SQL
duckdb> WITH tbl AS (SELECT NULL AS col UNION ALL SELECT 1 AS col) select ANY_VALUE(col) FROM tbl;
┌────────────────┐
│ any_value(col) │
│     int32      │
├────────────────┤
│       1        │
└────────────────┘

spark2> spark.sql("with tbl as (Select NULL as col union all select 1 as col) select first(col, true) from tbl").show()
+----------+
|first(col)|
+----------+
|         1|
+----------+


spark3> WITH tbl AS (SELECT NULL AS col UNION ALL SELECT 1 AS col) select FIRST(col, TRUE) FROM tbl;
1
```

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -1119,15 +1119,20 @@ class DuckDB(Dialect):
             return super().unnest_sql(expression)
 
         def ignorenulls_sql(self, expression: exp.IgnoreNulls) -> str:
-            if isinstance(expression.this, self.IGNORE_RESPECT_NULLS_WINDOW_FUNCTIONS):
+            this = expression.this
+
+            if isinstance(this, self.IGNORE_RESPECT_NULLS_WINDOW_FUNCTIONS):
                 # DuckDB should render IGNORE NULLS only for the general-purpose
                 # window functions that accept it e.g. FIRST_VALUE(... IGNORE NULLS) OVER (...)
                 return super().ignorenulls_sql(expression)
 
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/duckdb.py
```diff
@@ -1119,15 +1119,20 @@ def unnest_sql(self, expression: exp.Unnest) -> str:
...
-            if isinstance(expression.this, self.IGNORE_RESPECT_NULLS_WINDOW_FUNCTIONS):
+            this = expression.this
+
+            if isinstance(this, self.IGNORE_RESPECT_NULLS_WINDOW_FUNCTIONS):
                 # DuckDB should render IGNORE NULLS only for the general-purpose
                 # window functions that accept it e.g. FIRST_VALUE(... IGNORE NULLS) OVER (...)
                 return super().ignorenulls_sql(expression)
 
-            if not isinstance(expression.this, exp.AnyValue):
+            if isinstance(this, exp.First):
```
georgesittas, line 1129: maybe do this for `Last` too?

VaggelisD: Not sure if there's an equivalent `ANY_VALUE` for the reversed order

georgesittas: ok cool, ship it

PR #5646, opened 2025-08-22
feat(exasol): Add support for GROUP_CONCAT and LISTAGG functions

nnamdi16:


georgesittas (commented):
Hi @nnamdi16 👋 

Not sure about this approach, I'm skeptical about converting `GROUP_CONCAT` to `LISTAGG`. Perhaps we should mimic what MySQL [does](https://github.com/tobymao/sqlglot/blob/53aa8fe7f188012f765066f32c4179035fff036d/sqlglot/dialects/mysql.py#L679), since it appears to support the same syntax as Exasol ([docs](https://dev.mysql.com/doc/refman/8.4/en/aggregate-functions.html#function_group-concat))?

nnamdi16:
> Hi @nnamdi16 👋
> 
> Not sure about this approach, I'm skeptical about converting `GROUP_CONCAT` to `LISTAGG`. Perhaps we should mimic what MySQL [does](https://github.com/tobymao/sqlglot/blob/53aa8fe7f188012f765066f32c4179035fff036d/sqlglot/dialects/mysql.py#L679), since it appears to support the same syntax as Exasol ([docs](https://dev.mysql.com/doc/refman/8.4/en/aggregate-functions.html#function_group-concat))?

Hi @georgesittas I have modified the implementation to mimic mysql impmentation

sqlglot/dialects/exasol.py
```diff
@@ -176,6 +178,45 @@ class Parser(parser.Parser):
...
             ),
         }
+        FUNCTION_PARSERS = {
+            **parser.Parser.FUNCTION_PARSERS,
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/listagg.htm
+            # https://docs.exasol.com/db/latest/sql_references/functions/alphabeticallistfunctions/group_concat.htm
+            **dict.fromkeys(("GROUP_CONCAT", "LISTAGG"), lambda self: self._parse_group_concat()),
+        }
+
+        def _parse_group_concat(self) -> t.Optional[exp.Expression]:
```
georgesittas, line 189: Let's avoid copy-pasting this by moving it to `dialect.py`. Then, both MySQL and Exasol can leverage the same logic.

sqlglot/dialects/exasol.py
```diff
@@ -235,6 +276,9 @@ def datatype_sql(self, expression: exp.DataType) -> str:
             exp.TsOrDsDiff: _date_diff_sql,
             exp.DateTrunc: lambda self, e: self.func("TRUNC", e.this, unit_to_str(e)),
             exp.DatetimeTrunc: timestamptrunc_sql(),
+            exp.GroupConcat: lambda self, e: groupconcat_sql(
+                self, e, func_name="LISTAGG", within_group=True
+            ),
```
georgesittas, line 282: Are you sure `GROUP_CONCAT` and `LISTAGG` are interchangeable in Exasol?

nnamdi16: Yes they are interchangeble in Exasol. 

```
SELECT 
    id, department, hire_date, 
    LISTAGG(id, ',') within group (ORDER BY hire_date) OVER (PARTITION BY department rows between 1 preceding and 1 following) LISTAGG_RESULT
FROM employee_table ORDER BY department, hire_date
```

```
SELECT 
    id, 
    department, 
    hire_date,
    GROUP_CONCAT(id ORDER BY hire_date SEPARATOR ',') OVER (
        PARTITION BY department ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING
    ) AS group_concat_result
FROM employee_table
ORDER BY department, hire_date
```

The only difference is that LISTAGG uses window group but when you run the both query they achieve the same result.

@georgesittas

georgesittas: Ok, let's try it for now then.

sqlglot/dialects/exasol.py
georgesittas: Let's just do this^ and we can merge afterwards.

nnamdi16: Alright

georgesittas:
Followed up with https://github.com/tobymao/sqlglot/commit/fc7f6d8268f2710b587710f05cf1ff59b7f09732

Issue #5638, opened 2025-08-22
Unable to optimize STACK in a SELECT


catlynkong:
**Description**
The following Spark SQL failed to optimize
```
SELECT a, b
FROM
	(
		SELECT 
		    STACK(2, 'c',c,'d',d) AS (a, b)
		FROM db.tbl
	) AS T
```
with an error message `sqlglot.errors.OptimizeError: Column 'a' could not be resolved`

Moving the `STACK` call into a `LATERAL VIEW` would pass optimization. 

But given the above is valid Spark syntax, I'm hoping we can find a fix to correctly qualify the columns.

**Fully reproducible code snippet**
Should be able to reproduce with the following on sqlglot==27.8.0
```
from sqlglot import parse_one
from sqlglot.optimizer import optimize

sql = """
SELECT a, b
FROM
	(
		SELECT 
		    STACK(2, 'c',c,'d',d) AS (a, b)
		FROM db.tbl
	) AS T
"""

parsed = parse_one(sql, dialect="spark")
optimized = optimize(parsed)
```

**Official Documentation**
Please include links to official SQL documentation related to your issue.

PR #5647, opened 2025-08-22
feat(optimizer)!: Qualify columns generated by exp.Aliases

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5638

Properly handle `exp.Aliases` by:
- Including it in the `exp.Select::named_selects`, so that it's available to the `Resolver::get_source_columns()`  
- Excluding it from the nodes that will get aliased otherwise, so that we don't generate `STACK (...) AS (a1, a2, ...) AS _q_0`

### DIFF
--- a/tests/fixtures/optimizer/qualify_columns.sql
+++ b/tests/fixtures/optimizer/qualify_columns.sql
@@ -283,6 +283,10 @@ WITH t AS (SELECT 1 AS c) SELECT TO_JSON_STRING(t) AS _col_0 FROM t AS t;
 SELECT DATE_TRUNC(col1, WEEK(MONDAY)), col2 FROM t;
 SELECT DATE_TRUNC(t.col1, WEEK(MONDAY)) AS _col_0, t.col2 AS col2 FROM t AS t;
 
+# execute: false
+SELECT first FROM (SELECT 'val' AS col, STACK(2, 1, 2, 3) AS (first, second)) AS tbl;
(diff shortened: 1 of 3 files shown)


tests/fixtures/optimizer/qualify_columns.sql
```diff
@@ -283,6 +283,10 @@ WITH t AS (SELECT 1 AS c) SELECT TO_JSON_STRING(t) AS _col_0 FROM t AS t;
 SELECT DATE_TRUNC(col1, WEEK(MONDAY)), col2 FROM t;
 SELECT DATE_TRUNC(t.col1, WEEK(MONDAY)) AS _col_0, t.col2 AS col2 FROM t AS t;
 
+# execute: false
+SELECT first FROM (SELECT 'val' AS col, STACK(2, 1, 2, 3) AS (first, second)) AS tbl;
+SELECT tbl.first AS first FROM (SELECT 'val' AS col, STACK(2, 1, 2, 3) AS (first, second)) AS tbl;
```
georgesittas:
```suggestion
SELECT first, second FROM (SELECT 'val' AS col, STACK(2, 1, 2, 3) AS (first, second)) AS tbl;
SELECT tbl.first AS first, tbl.second AS second FROM (SELECT 'val' AS col, STACK(2, 1, 2, 3) AS (first, second)) AS tbl;
```

PR #5650, opened 2025-08-25
Improve predict parsing for bigquery, add expression check in ml functions

rloredo:
## Summary

Add specific parsing for BigQuery `ML PREDICT` function in the SQL parser. 
Ensure correct handling and validation of its parameters. 
It also introduces expressions checks to improve the accuracy of existing tests for ML functions.

### BigQuery ML function parsing changes

* Added a `_parse_predict` method to the BigQuery parser to handle the `ML.PREDICT` function, including validation for [allowed parameters](https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-predict) (`THRESHOLD`, `KEEP_ORIGINAL_COLUMNS`, `TRIAL_ID`) in the third argument struct.
* Registered the new `PREDICT` function parser in the BigQuery dialect's function parser mapping.

### Test improvements

* Added tests for various valid `ML.PREDICT` function usages, including cases with and without the optional struct parameter, and verified that the parsed expression is of type `exp.Predict`.
* Updated existing tests for `ML.FEATURES_AT_TIME` to assert the correct parsed expression type, improving test accuracy.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -700,6 +700,7 @@ class BigQuery(Dialect):
             ),
             "MAKE_INTERVAL": lambda self: self._parse_make_interval(),
             "FEATURES_AT_TIME": lambda self: self._parse_features_at_time(),
+            "PREDICT": lambda self: self._parse_predict(),
         }
         FUNCTION_PARSERS.pop("TRIM")
 
@@ -1004,6 +1005,42 @@ class BigQuery(Dialect):
                 connection=self._match_text_seq("WITH", "CONNECTION") and self._parse_table_parts(),
                 options=self._parse_properties(),
                 this=self._match_text_seq("AS") and self._parse_select(),
+            )
+
+        def _parse_predict(self) -> exp.Predict:
+            self._match_text_seq("MODEL")
+            this = self._parse_table()
+
+            self._match(TokenType.COMMA)
+            self._match_text_seq("TABLE")
+            expression = self._parse_table() or self._parse_paren()
+
+            params_struct = None
+            if self._match(TokenType.COMMA):
+                if self._match_text_seq("STRUCT"):
+                    self._match(TokenType.L_PAREN)
+                    expressions = self._parse_csv(lambda: self._parse_lambda(alias=True))
+                    self._match(TokenType.R_PAREN)
+                    # https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-predict#syntax
+                    allowed_params = {"THRESHOLD", "KEEP_ORIGINAL_COLUMNS", "TRIAL_ID"}
+                    for expr in expressions:
+                        if isinstance(expr, exp.Alias) and expr.alias.upper() not in allowed_params:
+                            self.raise_error(
+                                f"Invalid PREDICT parameter: {expr.alias}. "
+                                "Only THRESHOLD, KEEP_ORIGINAL_COLUMNS, and TRIAL_ID are allowed."
+                            )
+
+                    params_struct = self.expression(exp.Struct, expressions=expressions)
+                else:
+                    self.raise_error(
+                        "PREDICT function expects a STRUCT for the third parameter with THRESHOLD, KEEP_ORIGINAL_COLUMNS, and/or TRIAL_ID"
+                    )
+
+            return self.expression(
+                exp.Predict,
+                this=this,
+                expression=expression,
+                params_struct=params_struct,
             )
 
     class Generator(generator.Generator):
--- a/tests/dialects/test_bigquery.py
+++ b/tests/dialects/test_bigquery.py
@@ -1718,10 +1718,23 @@ WHERE
             },
         )
 
+        # ML functions
+        self.validate_identity(
+            "SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable)"
+        ).find(exp.From).this.this.assert_is(exp.Predict)
+        self.validate_identity(
+            "SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, (SELECT * FROM mytable))"
+        ).find(exp.From).this.this.assert_is(exp.Predict)
+        self.validate_identity(
+            "SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, STRUCT(0.5 AS threshold, TRUE AS keep_original_columns))"
+        ).find(exp.From).this.this.assert_is(exp.Predict)
+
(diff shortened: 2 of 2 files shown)


sqlglot/dialects/bigquery.py
```diff
@@ -1006,6 +1007,42 @@ def _parse_export_data(self) -> exp.Export:
                 this=self._match_text_seq("AS") and self._parse_select(),
             )
 
+        def _parse_predict(self) -> exp.Predict:
+            self._match_text_seq("MODEL")
+            this = self._parse_table()
+
+            self._match(TokenType.COMMA)
+            self._match_text_seq("TABLE")
+            expression = self._parse_table() or self._parse_paren()
```
georgesittas, line 1016: Let's move all of this logic to the base parser's `_parse_predict` method. It is only relevant to BigQuery, anyway.

sqlglot/dialects/bigquery.py
```diff
@@ -1006,6 +1007,42 @@ def _parse_export_data(self) -> exp.Export:
...
+
+            params_struct = None
+            if self._match(TokenType.COMMA):
+                if self._match_text_seq("STRUCT"):
+                    self._match(TokenType.L_PAREN)
+                    expressions = self._parse_csv(lambda: self._parse_lambda(alias=True))
+                    self._match(TokenType.R_PAREN)
```
georgesittas, line 1023: This doesn't look right; SQLGlot already parses BigQuery structs like the following:

```sql
STRUCT(0.5 AS threshold, TRUE AS keep_original_columns)
```

From what I'm seeing, we already handle this syntax:

```python
>>> import sqlglot
>>> sqlglot.parse_one("SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, STRUCT(0.5 AS threshold, TRUE AS keep_original_columns))", "bigquery")
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Predict(
        this=Table(
          this=Identifier(this=mymodel, quoted=False),
          db=Identifier(this=mydataset, quoted=False),
          catalog=Identifier(this=myproject, quoted=False)),
        expression=Table(
          this=Identifier(this=mytable, quoted=False),
          db=Identifier(this=mydataset, quoted=False),
          catalog=Identifier(this=myproject, quoted=False)),
        params_struct=Struct(
          expressions=[
            PropertyEQ(
              this=Identifier(this=threshold, quoted=False),
              expression=Literal(this=0.5, is_string=False)),
            PropertyEQ(
              this=Identifier(this=keep_original_columns, quoted=False),
              expression=Boolean(this=True))])),
      db=Identifier(this=ML, quoted=False))))
>>> sqlglot.parse_one("SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, STRUCT(0.5 AS threshold, TRUE AS keep_original_columns))", "bigquery").sql("bigquery")
'SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, STRUCT(0.5 AS threshold, TRUE AS keep_original_columns))'
```

rloredo: But if you put an incorrect parameters it stills parses it when it should fail.
e.g.
```sql
SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, STRUCT(0.5 AS wrong_param, TRUE AS keep_original_columns))
```
or
```sql 

SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, some_value)

```

should fail parsing because of the wrong params config, but in current implementation passes.

sqlglot/dialects/bigquery.py
rloredo: I'll wait for the resolution in the other comment to move it.

sqlglot/dialects/bigquery.py
VaggelisD: Looks like we added support for that function here: https://github.com/tobymao/sqlglot/issues/2375

rloredo: True, I missed those tests, sorry.
However, in this case I'm not adding the support for it but making the parameters more specific.
Or is this expected that parameters can be any and you don't want to enforce any check?

VaggelisD: > Or is this expected that parameters can be any and you don't want to enforce any check?



That is right, SQLGlot is allowed to be flexible in what it can parse (obviously within reason), the ground truth should always be the engine. Also, as you see, validating each arg requires scrutiny which really complicates the codebase.

tests/dialects/test_bigquery.py
```diff
@@ -1718,10 +1718,23 @@ def test_bigquery(self):
...
+        # ML functions
+        self.validate_identity(
+            "SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable)"
+        ).find(exp.From).this.this.assert_is(exp.Predict)
+        self.validate_identity(
+            "SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, (SELECT * FROM mytable))"
+        ).find(exp.From).this.this.assert_is(exp.Predict)
+        self.validate_identity(
+            "SELECT * FROM ML.PREDICT(MODEL myproject.mydataset.mymodel, TABLE myproject.mydataset.mytable, STRUCT(0.5 AS threshold, TRUE AS keep_original_columns))"
+        ).find(exp.From).this.this.assert_is(exp.Predict)
```
VaggelisD, line 1730: [Nit] When instance-checking, it'd be best if we can minimize those node accesses. For example, you could try doing `validate_identity("ML.PREDICT(...)").assert_is(...)` which should parse the function in a single node and test it immediately

sqlglot/dialects/bigquery.py
rloredo: Then I'll close this so it doesn't restrict the flexibility 👍

tests/dialects/test_bigquery.py
VaggelisD: If you still want to test a select statement, I think it'd be best to do something like:

```
ast = self.validate_identity(...)
assert ast.find(exp.Predict)
```

rloredo: ah true, I was trying to kill two birds with one stone, but it's true we only need it once just for that.

rloredo:
Closing because doesn't make sense to enforce argument parsing, given: 

> That is right, SQLGlot is allowed to be flexible in what it can parse (obviously within reason), the ground truth should always be the engine. Also, as you see, validating each arg requires scrutiny which really complicates the codebase.

PR #5652, opened 2025-08-25
feat(bigquery): Add support for ml.generate_embedding function

rloredo:
Add support for the BigQuery ML `GENERATE_EMBEDDING` [function,](https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-generate-embedding) 

Move and expanded tests for ML functions, including both `FEATURES_AT_TIME` and `GENERATE_EMBEDDING`, into a dedicated test method in `tests/dialects/test_bigquery.py`.

Note that ML.GENERATE_EMBEDDING function syntax is identical to ML.PREDICT function. This is because BigQuery has at least three types of ML. functions: 

- Table-Valued functions like the mentioned above, that are used used in FROM clauses and return tables
- Column-Valued functions that are used in SELECT clauses and return strings/int/scalar/array etc. values ([example](https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-bucketize#syntax))
- Window like functions. ([Example](https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-min-max-scaler#syntax))

Because of [this slack message ](https://tobiko-data.slack.com/archives/C0448SFS3PF/p1756114416582269?thread_ts=1756112705.588209&cid=C0448SFS3PF) I opted to put GENERATE_EMBEDDING as a new function, instead of refactoring existing code to make it reusable.

Also, I don't know why ML.FEATURES_AT_TIME is in bigquery.py whereas ML.PREDICT is in the general parser file. I've opted to put the new function in bigquery.py because it's specific to bigquery, but I can move it to the general parser as well.

### DIFF
--- a/sqlglot/dialects/bigquery.py
+++ b/sqlglot/dialects/bigquery.py
@@ -700,6 +700,7 @@ class BigQuery(Dialect):
             ),
             "MAKE_INTERVAL": lambda self: self._parse_make_interval(),
             "FEATURES_AT_TIME": lambda self: self._parse_features_at_time(),
+            "GENERATE_EMBEDDING": lambda self: self._parse_generate_embedding(),
         }
         FUNCTION_PARSERS.pop("TRIM")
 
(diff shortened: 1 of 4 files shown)


georgesittas (approved):
Thanks!

PR #5654, opened 2025-08-25
feat: parse to_char date literals

jasonthomassql:
Currently to_char is not parsing date formats if the format is a literal.  This breaks most transpiles as the expression in practice is a column name and not a recognized temporal type.

This checks if it is a string literal and attempts to build a formatted time.

### DIFF
--- a/sqlglot/dialects/dialect.py
+++ b/sqlglot/dialects/dialect.py
@@ -25,7 +25,7 @@ from sqlglot.jsonpath import JSONPathTokenizer, parse as parse_json_path
 from sqlglot.parser import Parser
 from sqlglot.time import TIMEZONES, format_time, subsecond_precision
 from sqlglot.tokens import Token, Tokenizer, TokenType
-from sqlglot.trie import new_trie
+from sqlglot.trie import new_trie, in_trie, TrieResult
 
 DATE_ADD_OR_DIFF = t.Union[
     exp.DateAdd,
@@ -2033,11 +2033,11 @@ def build_timetostr_or_tochar(args: t.List, dialect: DialectType) -> exp.TimeToS
         is_string_literal = format_str.is_string
 
         if is_temporal or is_string_literal:
-            try:
-                dialect_name = dialect.__class__.__name__.lower()
-                return build_formatted_time(exp.TimeToStr, dialect_name, default=True)(args)
-            except Exception:
-                pass  # If format isn't valid, fall back to ToChar
+            format_str_value = format_str.name
+            if is_valid_time_format(format_str_value, dialect.__class__.__name__.lower()):
+                return build_formatted_time(
+                    exp.TimeToStr, dialect.__class__.__name__.lower(), default=True
+                )(args)
 
     return exp.ToChar.from_arg_list(args)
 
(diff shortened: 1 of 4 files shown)


sqlglot/dialects/dialect.py
```diff
@@ -2022,15 +2022,22 @@ def groupconcat_sql(
...
         if not this.type:
             from sqlglot.optimizer.annotate_types import annotate_types
 
             annotate_types(this, dialect=dialect)
 
-        if this.is_type(*exp.DataType.TEMPORAL_TYPES):
-            dialect_name = dialect.__class__.__name__.lower()
-            return build_formatted_time(exp.TimeToStr, dialect_name, default=True)(args)
+        is_temporal = this.is_type(*exp.DataType.TEMPORAL_TYPES)
+        is_string_literal = isinstance(format_str, exp.Literal) and isinstance(format_str.this, str)
```
georgesittas:
```suggestion
        is_string_literal = format_str.is_string
```

sqlglot/dialects/dialect.py
```diff
@@ -2022,15 +2022,22 @@ def groupconcat_sql(
...
+
+        if is_temporal or is_string_literal:
+            try:
+                dialect_name = dialect.__class__.__name__.lower()
+                return build_formatted_time(exp.TimeToStr, dialect_name, default=True)(args)
```
georgesittas: Is it guaranteed for dremio, oracle and snowflake that a string literal implies `TimeToStr` semantics? Can't it be something like `'####'`, or other, used with numeric values?

jasonthomassql: yes, that is the problem.  it could be a numeric or string format.  the most likely case is a datetime format.  TBH I do not think the try/catch actually walks each part of the ast and determines if that is a valid time format for the dialect.  I wasn't really sure how to do that.

I do think this is better than the existing logic which does nothing and breaks but if you know of a way I can guarantee the string is a valid time format for that dialect then I will implement.

georgesittas: I don't know of a way to do this generically across all these dialects, that'd need more investigation. I'd rather not assume `TimeToStr` and break stuff, though. This doesn't seem like a safe default right now for string literals.

jasonthomassql: i added a format validator.  probably easier to use regex but I tried to stick with the trie method in the existing format_time.  Please let me know what you think.

georgesittas: Hey, thanks for taking a stab at this & for your patience.

I'm hesitating to merge this. I don't like the introduced complexity– validating the formats requires covering them completely, which is not the case today and is not really in scope right now.

I'll close this PR as I'm not seeing a way forward and the team can't spend more review cycles on this atm. One option is to patch this implementation on your end easily if you can make certain assumptions that simplify the problem.

Apologies for the inconvenience.

jasonthomassql: Disagree with the outcome here.  In my experience every to_char transpile that I do fails because it is using a column name and date format.  I have never seen an example where strings are formatted.  There are better functions for that.  Given that a holistic fix for to_char is not even on the horizon I don't see why we can't make iterative improvements.

https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good

Anyway, I will patch locally.

georgesittas: I hear you, but at the same time I prefer not touching shared code right now that affects other dialects, unless we can guarantee that the outcome is strictly better than what's happening today and can not lead to issues. If you think that I've missed something and you can actually guarantee that, then I'm happy to reconsider.

As I mentioned, the team's bandwidth at the moment is quite limited and can't afford to prioritize working on these transpilation paths while risking regressions. I'm not crossing out revisiting this later.

jasonthomassql: Well, this does solve most real world use cases of to_char.  The validator only attempts to parse if the the format is a string, the dialect has a TIME_MAPPING, and the provided format has an existing match for every component of that TIME_MAPPING.  If any of that is not true it will revert to previous behavior.  There are no guarantees in code but it seems very safe.

Issue #5658, opened 2025-08-26
Doris dialect: CAST expressions missing required aliases causing errors in subqueries and CTEs


peterylh:
When converting PostgreSQL SQL to Doris dialect, CAST expressions in subqueries and CTEs are missing required aliases, which causes SQL execution errors in Doris. Doris requires all expressions in subqueries and CTEs to have explicit aliases.

Original PostgreSQL SQL:
```
SELECT user_id FROM (
    SELECT CAST(user_id AS VARCHAR(50))
    FROM users
) AS subquery;
```

Current Doris conversion result (problematic):

```
SELECT user_id FROM (
    SELECT CAST(user_id AS VARCHAR(50))
    FROM users
) AS subquery;
```
Expected Doris conversion result:
```
SELECT user_id FROM (
    SELECT CAST(user_id AS VARCHAR(50)) AS user_id
    FROM users
) AS subquery;
```

Doris database requires all expressions in subqueries and Common Table Expressions (CTEs) to have explicit aliases. Without aliases, Doris will throw execution errors. The current SQLGlot Doris dialect implementation may not consistently add aliases to CAST expressions in these contexts.

georgesittas:
This is out of scope for the base transpilation flow, check out my comment in #5659.

PR #5659, opened 2025-08-26
fix(doris): Improve CAST expression alias handling in subqueries and CTE

peterylh:
## Description
This PR fixes issue #5658 where CAST expressions in subqueries and CTEs were missing required aliases when converting from PostgreSQL to Doris dialect, causing SQL execution errors in Doris.

Doris requires all expressions in subqueries and Common Table Expressions (CTEs) to have explicit aliases. The previous implementation didn't consistently add aliases to CAST expressions in these contexts, leading to runtime errors like:
```
-- Problematic output (missing alias)
SELECT user_id FROM (
    SELECT CAST(user_id AS VARCHAR(50))
    FROM users
) AS subquery;
```

Expected output is :
```
SELECT user_id FROM (
    SELECT CAST(user_id AS VARCHAR(50)) AS user_id
    FROM users
) AS subquery;
```

## Solution
Implemented a new cast_sql_with_alias function that:
Checks if a CAST expression is within a CTE or Subquery。 
Adds aliases for column references : Automatically adds aliases when CAST contains simple column references
Preserves existing behavior : Maintains default CAST SQL generation for other contexts

### DIFF
--- a/sqlglot/dialects/doris.py
+++ b/sqlglot/dialects/doris.py
@@ -13,6 +13,28 @@ from sqlglot.dialects.mysql import MySQL
 from sqlglot.tokens import TokenType
 
 
+def cast_sql_with_alias(self, expression: exp.Cast) -> str:
+    """Generate CAST SQL with automatic alias when in derived table context."""
+    # Check if this CAST is within a CTE or Subquery
+    derived_table_parent = expression.find_ancestor(exp.CTE, exp.Subquery)
+    if derived_table_parent:
+        # Find the SELECT that contains this CAST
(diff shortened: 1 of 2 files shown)


georgesittas (commented):
Hey @peterylh, appreciate the PR but I don't think the direction here is correct.

You should instead look into using `qualify` (or just `qualify_outputs`) from the optimizer sub-package. Based on your description, I'm confident that handling `Cast` as a special case introduces unnecessary complexity because it is incomplete. What if you had another function instead of `CAST` that is not aliased, etc?

This is out of scope for the basic transpilation flow, so I don't think we'll introduce transformations like this.

PR #5664, opened 2025-08-26
feat(singlestore): Fixed parsing/generation of exp.Hll

AdalbertMemSQL:
[APPROX_COUNT_DISTINCT](https://docs.singlestore.com/cloud/reference/sql-reference/distinct-count-estimation-functions/approx-count-distinct/)

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -136,6 +136,7 @@ class SingleStore(MySQL):
                 this=seq_get(args, 0),
                 format=MySQL.format_time(exp.Literal.string("%W")),
             ),
+            "APPROX_COUNT_DISTINCT": lambda args, dialect: exp.Hll(this=seq_get(args, 0)),
         }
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -136,6 +136,7 @@ class Parser(MySQL.Parser):
                 this=seq_get(args, 0),
                 format=MySQL.format_time(exp.Literal.string("%W")),
             ),
+            "APPROX_COUNT_DISTINCT": lambda args, dialect: exp.Hll(this=seq_get(args, 0)),
```
georgesittas: Why is this only setting `this`? SingleStore supports multiple arguments, so I think we need `from_arg_list`:

```
APPROX_COUNT_DISTINCT ( expr [, expr ...] )
```

AdalbertMemSQL: nice catch

PR #5669, opened 2025-08-26
feat(singlestore): Fixed generation of exp.LogicalOr

AdalbertMemSQL:


sqlglot/dialects/singlestore.py
```diff
@@ -263,6 +263,7 @@ class Generator(MySQL.Generator):
             exp.Hll: rename_func("APPROX_COUNT_DISTINCT"),
             exp.ApproxDistinct: rename_func("APPROX_COUNT_DISTINCT"),
             exp.CountIf: count_if_to_sum,
+            exp.LogicalOr: lambda self, e: f"MAX(ABS({self.sql(e, 'this')}))",
```
georgesittas, line 266: Why do we need the `ABS`?

AdalbertMemSQL: Without `ABS`, if the input is -8, -3, 0, the result will be 0, which is False.
But LogicalOr should return True if at least one integer is non-zero.

georgesittas: Sounds good, thank you for clarifying.

georgesittas (approved):

PR #5674, opened 2025-08-26
feat(dremio): add CURRENT_DATE_UTC

jasonthomassql:


sqlglot/dialects/dremio.py
```diff
@@ -142,8 +153,22 @@ class Parser(parser.Parser):
             "DATE_ADD": build_date_delta_with_cast_interval(exp.DateAdd),
             "DATE_SUB": build_date_delta_with_cast_interval(exp.DateSub),
             "ARRAY_GENERATE_RANGE": exp.GenerateSeries.from_arg_list,
+            "CURRENT_DATE_UTC": lambda args: Dremio._current_date_utc_expr(),
+        }
+
+        KEYWORDS = {
+            **tokens.Tokenizer.KEYWORDS,
+            "CURRENT_DATE_UTC": TokenType.COMMAND,
         }
```
georgesittas: This doesn't seem right; why was it needed? `COMMAND` is a token type reserved for statements that we want to wrap in dummy AST nodes to avoid having the parser fail when it encounters them.

sqlglot/dialects/dremio.py
```diff
@@ -131,6 +132,16 @@ class Dremio(Dialect):
         "tzo": "%z",  # numeric offset (+0200)
     }
 
+    @staticmethod
+    def _current_date_utc_expr() -> exp.Expression:
```
georgesittas: Nit: can we move this out to the top level and have it accept `args: t.List` so that it matches the existing convention?

sqlglot/dialects/dremio.py
```diff
@@ -142,8 +153,22 @@ class Parser(parser.Parser):
...
+            "CURRENT_DATE_UTC": TokenType.COMMAND,
         }
 
+        def _parse_primary(self) -> t.Optional[exp.Expression]:
+            if self._match(TokenType.COMMAND) and self._prev.text.upper() == "CURRENT_DATE_UTC":
+                if self._match(TokenType.L_PAREN):
+                    self._match(TokenType.R_PAREN)
+                return Dremio._current_date_utc_expr()
+
+            return super()._parse_primary()
```
georgesittas: We should avoid hacky workarounds; the fact that you had to override `_parse_primary` here means the approach is probably not correct.

I'm guessing you may have run into something I didn't anticipate when I commented on the issue: the function's name is not recognized as a function. Can you verify whether that's the case?

sqlglot/dialects/dremio.py
```diff
@@ -189,5 +214,17 @@ def datatype_sql(self, expression: exp.DataType) -> str:
...
+        def cast_sql(self, expression: exp.Cast, safe_prefix: str | None = None) -> str:
+            # Match: CAST(CURRENT_TIMESTAMP AT TIME ZONE 'UTC' AS DATE)
+            if expression.is_type(exp.DataType.Type.DATE):
+                at_time_zone = expression.this
+
+                if (
+                    isinstance(at_time_zone, exp.AtTimeZone)
+                    and isinstance(at_time_zone.this, exp.CurrentTimestamp)
+                    and isinstance(at_time_zone.args.get("zone"), exp.Literal)
+                    and at_time_zone.args["zone"].this.upper() == "UTC"
```
georgesittas: I think this may work too:
```suggestion
                    and at_time_zone.text("zone").upper() == "UTC"
```

georgesittas (commented):
Thanks for the PR :)

sqlglot/dialects/dremio.py
jasonthomassql: No, it can either be function or a token.  This was what I was asking in the issue.  Not sure how to solve without this override.  These are both valid dremio syntax.  The function works fine in the parser, but the token without parens is not being parsed.  I need to parse the token without parens.  Let me know if I am doing that wrong.

SELECT CURRENT_DATE_UTC;
SELECT CURRENT_DATE_UTC();

georgesittas: Let's add a token then as you mentioned in the ticket– `TokenType.CURRENT_DATE_UTC`. You'd map `"CURRENT_DATE_UTC"` to it only for Dremio and then handle it by adding it to `NO_PAREN_FUNCTIONS`, where you'll reuse the constructor you implemented.

jasonthomassql: Sorry, but I think I need some help.  I tried your suggestion by removing the _parse_primary override and replacing with this:

    NO_PAREN_FUNCTION_PARSERS = {
        **parser.Parser.NO_PAREN_FUNCTION_PARSERS,
        "CURRENT_DATE_UTC": build_current_date_utc,
    }
    
It doesn't work.  I need to be able to parse it with or without parens and if I add that line it breaks the with paren parsing.   As the code is currently I am  not handling the NO_PAREN option at all so this will not parse SELECT CURRENT_DATE_UTC;

Should I add back the _parse_primary override?

sqlglot/dialects/dremio.py
jasonthomassql: removed

sqlglot/dialects/dremio.py
jasonthomassql: Basically the NO_PAREN_FUNCTION_PARSERS will catch the CURRENT_DATE_UTC whether it has parens or not.  If it has parens, it breaks.  It never gets to the FUNCTIONS parser.

georgesittas: No worries, thanks for trying it out. I'll take a look at it soon.

> Should I add back the _parse_primary override?

I doubt that's the right approach at first glance. Will let you know if I observe otherwise.

jasonthomassql: ok, think i got it.

PR #5677, opened 2025-08-26
feat(databricks):  handle unsupported isnumeric

jasonthomassql:
Databricks doesn't have isnumeric.  The standard way is to do a try_cast.  Try cast does not exist in the parent spark dialect so implementing the transform here.

Let me know if i should move the isnumeric class up to expressions.py.  I can't find examples of other dialects trying to parse this.

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -16,12 +16,23 @@ from sqlglot.tokens import TokenType
 from sqlglot.optimizer.annotate_types import TypeAnnotator
 
 
+class IsNumeric(exp.Expression):
+    arg_types = {"this": True}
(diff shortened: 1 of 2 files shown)


georgesittas:
I don't quite understand this PR. What's the goal exactly? If `ISNUMERIC` is not supported by Databricks, why are we introducing a new expression and parsing it with the Databricks dialect?

Btw, there's no other place where an `Expression` subclass is defined outside of `expressions.py`, or another expression that sets a `dialect` attribute. Let's avoid straying away from existing conventions unless justified.

I'll go ahead and close this as it seems like there's a misunderstanding and solving any underlying issue will likely result in different approach altogether.

PR #5678, opened 2025-08-27
feat(optimizer)!: parse and annotate type for bigquery APPROX_QUANTILES

geooo109:
This PR adds parsing and type annotation for `APPROX_QUANTILES`

**DOCS**
[BigQuery APPROX_QUANTILES](https://cloud.google.com/bigquery/docs/reference/standard-sql/approximate_aggregate_functions#approx_quantiles)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -5596,6 +5596,10 @@ class ToChar(Func):
     }
 
 
+class ToCodePoints(Func):
+    pass
+
+
 # https://docs.snowflake.com/en/sql-reference/functions/to_decimal
 # https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/TO_NUMBER.html
 class ToNumber(Func):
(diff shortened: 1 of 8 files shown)


sqlglot/expressions.py
```diff
@@ -5493,6 +5493,10 @@ class ApproxTopSum(AggFunc):
     arg_types = {"this": True, "expression": True, "count": True}
 
 
+class ApproxQuantiles(AggFunc):
+    arg_types = {"this": True, "expression": False}
+
```
VaggelisD, line 5498: We do have `ApproxQuantile`, should we reuse that here? It might not be easily transpilable, though, not sure how it matches the semantics of other dialects.

georgesittas: I'd advise not to, I remember that it was tricky to convert the plural version to the singular; they're different things.

georgesittas: Let's just double-check that this `arg_types` is good in case `APPROX_QUANTILES` is supported by others.

georgesittas (approved):

PR #5680, opened 2025-08-27
feat(optimizer)!: annotate type for bigquery TO_HEX

geooo109:
This PR adds type annotation support for `TO_HEX`

**DOCS**
[BigQuery TO_HEX](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#to_hex)

### DIFF
--- a/tests/fixtures/optimizer/annotate_functions.sql
+++ b/tests/fixtures/optimizer/annotate_functions.sql
@@ -829,7 +829,7 @@ BINARY;
 
 # dialect: bigquery
 TO_HEX(b'foo');
-BINARY;
+STRING;
 
 --------------------------------------
 -- Snowflake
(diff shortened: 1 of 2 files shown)


tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -817,6 +817,10 @@ ARRAY<STRUCT<STRING, BIGINT>>;
 APPROX_TOP_SUM(tbl.bigint_col, 1.5, 2);
 ARRAY<STRUCT<BIGINT, BIGINT>>;
 
+# dialect: bigquery
+TO_HEX(b'foo');
+BINARY;
```
VaggelisD: This should return a `STRING`

geooo109: messed up with the `FROM_HEX`,  thanks for the catch.

PR #5703, opened 2025-08-28
feat!: add support for REVOKE DDL

newtonapple:
The intention of this PR is similar to the original Grant DDL [PR](https://github.com/tobymao/sqlglot/pull/4138): it's to layout initial / foundational work for `REVOKE` DDL support.  

## Docs
[Databricks](https://docs.databricks.com/aws/en/sql/language-manual/security-revoke) | [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/revoke-privilege) | [Presto](https://prestodb.io/docs/current/sql/revoke.html) | [Trino](https://trino.io/docs/current/sql/revoke.html) | [MySQL](https://dev.mysql.com/doc/refman/8.4/en/revoke.html) | [Postgres](https://www.postgresql.org/docs/current/sql-revoke.html) | [T-SQL](https://learn.microsoft.com/en-us/sql/t-sql/statements/revoke-transact-sql?view=sql-server-ver16) | [Clickhouse](https://clickhouse.com/docs/sql-reference/statements/revoke) | [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/REVOKE.html) | [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_REVOKE.html) | [Redshift REVOKE examples](https://docs.aws.amazon.com/redshift/latest/dg/r_REVOKE-examples.html)


tests/dialects/test_postgres.py
```diff
@@ -1498,3 +1498,72 @@ def test_locks(self):
...
+
+        advanced_grants = [
+            "GRANT INSERT, DELETE ON ALL TABLES IN SCHEMA myschema TO user1",
+            "GRANT developer_role TO john",
+            "GRANT admin_role TO mary WITH ADMIN OPTION",
+        ]
+
+        for sql in advanced_grants:
+            with self.subTest(f"Testing PostgreSQL's advanced GRANT statement: {sql}"):
+                self.validate_identity(sql, check_command_warning=True)
```
newtonapple: These were not in the original [PR](https://github.com/tobymao/sqlglot/pull/4138), so I added them.

tests/dialects/test_postgres.py
```diff
@@ -1498,3 +1498,78 @@ def test_locks(self):
         for key_type in ("FOR SHARE", "FOR UPDATE", "FOR NO KEY UPDATE", "FOR KEY SHARE"):
             with self.subTest(f"Test lock type {key_type}"):
                 self.validate_identity(f"SELECT 1 FROM foo AS x {key_type} OF x")
+
+    def test_grant(self):
```
newtonapple, line 1502: These were not in the original [PR](https://github.com/tobymao/sqlglot/pull/4138), so I added them.

sqlglot/expressions.py
```diff
@@ -2468,6 +2468,17 @@ class Grant(Expression):
...
 
+class Revoke(Expression):
+    arg_types = {
+        "privileges": True,
+        "kind": False,
+        "securable": True,
+        "principals": True,
+        "grant_option": False,
+        "cascade": False,
+    }
```
VaggelisD: All of these except `cascade` are identical to `GRANT`, right? I wonder whether we could abstract them a way in a new node such as:

```Python3

class DCL(Expression):
  arg_types = {
    ...
  }

class Grant(DCL):
  pass

class Revoke(DCL):
 pass

```

sqlglot/generator.py
```diff
@@ -4818,6 +4818,24 @@ def grant_sql(self, expression: exp.Grant) -> str:
 
         return f"GRANT {privileges_sql} ON{kind}{securable} TO {principals}{grant_option}"
 
+    def revoke_sql(self, expression: exp.Revoke) -> str:
+        grant_option_for = "GRANT OPTION FOR " if expression.args.get("grant_option") else ""
+
+        privileges_sql = self.expressions(expression, key="privileges", flat=True)
```
VaggelisD: Both this and the parser are very similar to grants, should we unify these 2?

sqlglot/parser.py
```diff
@@ -8441,6 +8442,42 @@ def _parse_grant(self) -> exp.Grant | exp.Command:
...
 
+    def _parse_revoke(self) -> exp.Revoke | exp.Command:
+        start = self._prev
+
+        grant_option = self._match_text_seq("GRANT", "OPTION", "FOR")
+
+        privileges = self._parse_csv(self._parse_grant_privilege)
+
+        self._match(TokenType.ON)
+        kind = self._match_set(self.CREATABLES) and self._prev.text.upper()
```
VaggelisD: Ditto, this looks identical to `_parse_grant`, maybe with some of the order switched up plus the `CASCADE` parsing which `GRANT` can also do as a no-op

tests/dialects/test_postgres.py
```diff
@@ -1498,3 +1498,78 @@ def test_locks(self):
...
+        grant_cmds = [
+            "GRANT SELECT ON TABLE users TO role1",
+            "GRANT INSERT, DELETE ON TABLE orders TO user1",
+            "GRANT SELECT ON employees TO manager WITH GRANT OPTION",
+            "GRANT USAGE ON SCHEMA finance TO user2",
+            "GRANT ALL PRIVILEGES ON DATABASE mydb TO PUBLIC",
```
VaggelisD, line 1508: Ditto about these tests, could we merge them such as:

```Python3
dcl_cmds = [
"SELECT ON TABLE users TO role1",
 ...
]
    
for sql in dcl_cmds:
 with self.subTest(f"Testing PostgreSQL's GRANT statement: {sql}"):
    self.validate_identity(f"GRANT {sql}")

 with self.subTest(f"Testing PostgreSQL's REVOKE statement: {sql}"):
    self.validate_identity(f"REVOKE {sql}")
```

VaggelisD (commented):
Thank you for taking this PR on @newtonapple! My main comment has to do with code duplication as these 2 statements look identical from what I see.

sqlglot/expressions.py
georgesittas: You could extract the dictionary itself out instead of introducing a new common node, perhaps? Until a need arises for it, like `DDL` etc.

tests/dialects/test_postgres.py
newtonapple: I did a round a refactoring on these basically doing string substitution for `TO` & `FROM`, then I realized it made the tests much harder to read.  Yes, it did DRY up the tests a bit.  But, I think for testing, it's much better to see the test cases exactly as they are instead of having to trace through the code just to see what the test cases are.  So, a little bit of duplications is probably ok here.  

@VaggelisD @georgesittas what do you think?  If you guys really want to keep these test cases DRY, I can finish the refactoring.

georgesittas: I'm good either way. Happy to just get this in to unblock you.

sqlglot/generator.py
```diff
@@ -4814,9 +4821,32 @@ def grant_sql(self, expression: exp.Grant) -> str:
...
-        grant_option = " WITH GRANT OPTION" if expression.args.get("grant_option") else ""
+        if not expression.args.get("grant_option"):
+            grant_option_prefix = grant_option_suffix = ""
 
-        return f"GRANT {privileges_sql} ON{kind}{securable} TO {principals}{grant_option}"
+        # cascade for revoke only
+        cascade = ""
+        if hasattr(expression, "args") and "cascade" in expression.args:
+            cascade_sql = self.sql(expression, "cascade")
+            cascade = f" {cascade_sql}" if cascade_sql else ""
```
georgesittas:
```suggestion
        cascade = self.sql(expression, "cascade")
        cascade = f" {cascade}" if cascade else ""
```

PR #5704, opened 2025-08-28
feat(singlestore): Added SingleStoreDB alias to the SingleStore dialect

AdalbertMemSQL:
This PR introduces an alias for the `singlestore` dialect under the name `singlestoredb`.
The official Python SDK, Ibis, and SQLAlchemy use `singlestoredb` as the canonical dialect/driver name.
This change makes the naming consistent across the SingleStore ecosystem, removes integration issues (especially with Ibis), and improves the developer experience without breaking existing code.

### DIFF
--- a/sqlglot/dialects/__init__.py
+++ b/sqlglot/dialects/__init__.py
@@ -87,6 +87,7 @@ DIALECTS = [
     "Redshift",
     "RisingWave",
     "SingleStore",
+    "SingleStoreDB",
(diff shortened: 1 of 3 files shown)


georgesittas:
Aliases should be handled in the application's side. We discussed this in the past and concluded that having aliases in SQLGlot is not something we want to do, such that there's exactly one dialect that acts as the source of truth.

PR #5707, opened 2025-08-28
feat(singlestore): Marked exp.Reduce finish argument as unsupported

AdalbertMemSQL:
[REDUCE](https://docs.singlestore.com/cloud/reference/sql-reference/json-functions/reduce/)


georgesittas (commented):
@AdalbertMemSQL do we parse `REDUCE` correctly for SingleStore today? With a quick look, it seems we may be storing the arguments in the wrong order due to the default parser. For SingleStore, `initial` comes first.

Let's test this in any case.

georgesittas (approved):

PR #5715, opened 2025-08-29
feat(optimizer)!: parse and annotate type for bq FORMAT

geooo109:
This PR adds parsing and type annotation support for `FORMAT`

**DOCS**
[BigQuery FORMAT](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#format_string)
[Spark FORMAT_STRING](https://spark.apache.org/docs/latest/api/sql/index.html#format_string)
[Databricks FORMAT_STRING](https://docs.databricks.com/aws/en/sql/language-manual/functions/format_string)


VaggelisD:
There's more dialects that support `FORMAT`, e.g from a quick search:

https://learn.microsoft.com/en-us/sql/t-sql/functions/format-transact-sql?view=sql-server-ver17
https://www.postgresql.org/docs/current/functions-string.html

sqlglot/expressions.py
```diff
@@ -6438,6 +6438,12 @@ class FormatJson(Expression):
     pass
 
 
+class Format(Func):
+    arg_types = {"this": True, "expressions": True}
+    is_var_len_args = True
+    _sql_names = ["FORMAT", "FORMAT_STRING"]
```
georgesittas: Why do we need the alias? Can we just keep the default `FORMAT` without specifying `_sql_names` and then add `FUNCTIONS` entries where needed?

Have we double-checked other dialects too? This one's particularly common and I'd like to ensure we won't break anything. Oracle, Snowflake, etc.

georgesittas:
> There's more dialects that support `FORMAT`, e.g from a quick search:
> 
> https://learn.microsoft.com/en-us/sql/t-sql/functions/format-transact-sql?view=sql-server-ver17 https://www.postgresql.org/docs/current/functions-string.html

We can probably handle Postgres' as well, T-SQL's seems different.

sqlglot/expressions.py
geooo109: Will revert, thanks

georgesittas (approved):

PR #5730, opened 2025-08-29
feat: support dremio alias repeatstr

jasonthomassql:
Dremio has an alias function for repeat called repeatstr with exact same functionality.  Handle it the same as repeat.

https://docs.dremio.com/current/reference/sql/sql-functions/functions/REPEATSTR/
https://docs.dremio.com/current/reference/sql/sql-functions/functions/REPEAT/


geooo109:
@jasonthomassql can you also add a test ?

geooo109 (approved):

PR #5734, opened 2025-08-30
feat(snowflake): parse ALTER SESSION 

tekumara:
parse ALTER SESSION as exp.AlterSession rather than exp.Command

uses SetItem for the expressions, similar to exp.Set (used by duckdb)

see https://docs.snowflake.com/en/sql-reference/sql/alter-session


tobymao:
Do other dialects have alter session? If so you need to implement them otherwise you’ll break command parsing.

tekumara:
Of the sqlglot officially supported dialects I think Oracle is the only other db with [ALTER SESSION](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/ALTER-SESSION.html). It has extra clauses in addition to the SET clause implemented here.

For the DBs that don't have ALTER SESSION should they still return a Command? And for Oracle, could a Command be returned for the as yet implemented extra clauses?

sqlglot/generator.py
```diff
@@ -3565,6 +3565,13 @@ def alter_sql(self, expression: exp.Alter) -> str:
 
         return f"ALTER {kind}{exists}{only} {self.sql(expression, 'this')}{on_cluster}{check}{self.sep()}{actions_sql}{not_valid}{options}"
 
+    def altersession_sql(self, expression: exp.AlterSession) -> str:
+        parts: t.List[str] = [self.sql(item, "this") for item in expression.expressions]
+
+        items_sql = ", ".join(parts)
```
georgesittas: Looks like we can refactor this using `expressions`.

sqlglot/parser.py
```diff
@@ -7580,13 +7581,27 @@ def _parse_alter_table_set(self) -> exp.AlterSet:
...
 
-    def _parse_alter(self) -> exp.Alter | exp.Command:
+    def _parse_alter(self) -> exp.Alter | exp.AlterSession | exp.Command:
         start = self._prev
 
         alter_token = self._match_set(self.ALTERABLES) and self._prev
         if not alter_token:
             return self._parse_as_command(start)
 
+        if alter_token.token_type == TokenType.SESSION:
```
georgesittas, line 7591: Can we put this in a method and add it to `ALTER_PARSERS` instead? Perhaps you can add it to Snowflake for now if it covers 100% of its syntax.

sqlglot/tokens.py
```diff
@@ -835,6 +836,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "SCHEMA": TokenType.SCHEMA,
         "SELECT": TokenType.SELECT,
         "SEMI": TokenType.SEMI,
+        "SESSION": TokenType.SESSION,
```
georgesittas, line 839: So if we make it work using `ALTER_PARSERS` there's no need for this token.

georgesittas:
Hey @tekumara 👋 , any plans to finish this?

sqlglot/generator.py
tekumara: can do - I'll need to no longer set `kind="Session"` otherwise it'll produce:
```
ALTER SESSION SET SESSION autocommit = FALSE, SESSION QUERY_TAG = 'qtag', SESSION JSON_INDENT = 1
```

sqlglot/parser.py
tekumara: the challenge appears to be that [`this = self._parse_table(schema=True)`](https://github.com/tekumara/sqlglot/blob/22ecdc08a2400ff09902e00f0790b4c52b758f8f/sqlglot/parser.py#L7607) will consume the `SESSION` token before we get to [looking up ALTER_PARSERS](https://github.com/tekumara/sqlglot/blob/22ecdc08a2400ff09902e00f0790b4c52b758f8f/sqlglot/parser.py#L7614) 🤔

georgesittas: We should add it to `ALTERABLES` as well to avoid this I think.

tekumara: I have the `SESSION` token in `ALTERABLES` already, what else am I missing?

georgesittas: Hm, then why is it being consumed by the `_parse_schema` call which follows the `_match_set(ALTERABLES)`?

tekumara: So it looks like the `SET` (which follows `SESSION`) is being consumed in [_parse_id_var here](https://github.com/tekumara/sqlglot/blob/22ecdc08a2400ff09902e00f0790b4c52b758f8f/sqlglot/parser.py#L7174)

tekumara: (sorry not `SESSION` as I stated above) .. what would be the best thing to do?

georgesittas: Ah, ok. So we should consume this syntax using the parser corresponding to `"SET"` in `ALTER_PARSERS`, right? Did you already do this and it still overshoots by consuming `SET` earlier?

tekumara: So a `SESSION SET` parser? Haven't tried that, will give it a go tomorrow!

tekumara: I'm still stuck. It's still overshooting and consuming `SET` early inside `_parse_id_var`.

georgesittas: Ok no worries, I'll take another look soon and may take it to the finish line.

tekumara: Thank you @georgesittas !

georgesittas:
@tekumara https://github.com/tobymao/sqlglot/commit/005564ab28cb14be469f09e89b01275d6e25874e

PR #5746, opened 2025-09-02
feat(singlestore): Improved geneation of exp.DataType

AdalbertMemSQL:
* Improved handling of unsupported datatypes by throwing a corresponding error message and returning the closest available data type
 * Updated type mapping for SingleStore-specific data types
 * Replaced `VARBINARY` without size by `BLOB`

[Data types](https://docs.singlestore.com/cloud/reference/sql-reference/data-types/)

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -365,6 +365,94 @@ class SingleStore(MySQL):
         }
         TRANSFORMS.pop(exp.JSONExtractScalar)
 
+        UNSUPPORTED_TYPE_MAPPING = {
+            exp.DataType.Type.ARRAY: "TEXT",
+            exp.DataType.Type.AGGREGATEFUNCTION: "TEXT",
+            exp.DataType.Type.SIMPLEAGGREGATEFUNCTION: "TEXT",
+            exp.DataType.Type.BIGSERIAL: "BIGINT AUTO_INCREMENT KEY",
+            exp.DataType.Type.BPCHAR: "TEXT",
+            exp.DataType.Type.DATEMULTIRANGE: "TEXT",
+            exp.DataType.Type.DATERANGE: "TEXT",
+            exp.DataType.Type.DYNAMIC: "TEXT",
+            exp.DataType.Type.HLLSKETCH: "TEXT",
+            exp.DataType.Type.HSTORE: "TEXT",
+            exp.DataType.Type.IMAGE: "LONGBLOB",
+            exp.DataType.Type.INET: "TEXT",
+            exp.DataType.Type.INT128: "DECIMAL(39,0)",
+            exp.DataType.Type.INT256: "DECIMAL(65,0)",
+            exp.DataType.Type.INT4MULTIRANGE: "TEXT",
+            exp.DataType.Type.INT4RANGE: "TEXT",
+            exp.DataType.Type.INT8MULTIRANGE: "TEXT",
+            exp.DataType.Type.INT8RANGE: "TEXT",
+            exp.DataType.Type.INTERVAL: "TEXT",
+            exp.DataType.Type.IPADDRESS: "TEXT",
+            exp.DataType.Type.IPPREFIX: "TEXT",
+            exp.DataType.Type.IPV4: "TEXT",
+            exp.DataType.Type.IPV6: "TEXT",
+            exp.DataType.Type.LIST: "TEXT",
+            exp.DataType.Type.MAP: "TEXT",
+            exp.DataType.Type.LOWCARDINALITY: "TEXT",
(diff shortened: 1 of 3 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -55,6 +55,7 @@ class Tokenizer(MySQL.Tokenizer):
             **MySQL.Tokenizer.KEYWORDS,
             "BSON": TokenType.JSONB,
             "GEOGRAPHYPOINT": TokenType.GEOGRAPHYPOINT,
+            "TIMESTAMP": TokenType.TIMESTAMP,
```
VaggelisD, line 61: Not sure if this is needed, MySQL does override `"TIMESTAMP": TokenType.TIMESTAMPTZ`, but `TokenType.TIMESTAMPTZ` would be mapped back to `TIMESTAMP` in SingleStore anyways.

sqlglot/dialects/singlestore.py
```diff
@@ -365,6 +366,94 @@ class Generator(MySQL.Generator):
         }
         TRANSFORMS.pop(exp.JSONExtractScalar)
 
+        UNSUPPORTED_TYPE_MAPPING = {
```
VaggelisD: Not sure about this pattern, we don't have it elsewere, right? It's too noisy and mappings like `Type.STRUCT: "TEXT"` don't look correct

sqlglot/dialects/singlestore.py
```diff
@@ -365,6 +366,94 @@ class Generator(MySQL.Generator):
...
+            exp.DataType.Type.UNION: "TEXT",
+            exp.DataType.Type.UNKNOWN: "TEXT",
+            exp.DataType.Type.USERDEFINED: "TEXT",
+            exp.DataType.Type.UUID: "TEXT",
+            exp.DataType.Type.VARIANT: "TEXT",
+            exp.DataType.Type.XML: "TEXT",
+            exp.DataType.Type.TDIGEST: "BLOB",
+        }
+
+        TYPE_MAPPING = {
```
VaggelisD, line 444: Are we sure these check out? For instance, can we transpile `Type.POLYGON` to `"GEOGRAPHY"`?  

I'd advise reverting all the mappings that are not tested in this PR and adding support for them whenever needed e.g `ENUM8 -> ENUM` would be a Clickhouse -> SingleStore transpilation.

sqlglot/dialects/singlestore.py
georgesittas: Yeah, agreed re: arbitrary mappings. I'm ok with having an empty mapping in the base generator and then filling it in for SingleStore in order to show a warning.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: I added it because otherwise the tests were failing.
`TokenType.TIMESTAMPTZ` is always converted to `TIMESTAMP` (without size).
So `TIMESTAMP(6)` is converted to `TokenType.TIMESTAMPTZ`, which is converted to `TIMESTAMP`.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: I would prefer adding tests in this PR. I don't see sense in splitting this PR into a tenth of smaller ones when all of them change almost the same functionality.

sqlglot/dialects/singlestore.py
AdalbertMemSQL: You’re right that some mappings (like Type. `STRUCT: "TEXT"`) won’t be strictly correct — that’s expected since SingleStore doesn’t have a one-to-one match for all data types. The idea is to provide fallbacks where it’s still useful (e.g., `IMAGE` -> `BLOB` is a practical replacement).

As for the noisiness: without this, the connector ends up generating invalid queries silently, which can be misleading since users don’t see any warnings. Having these mappings makes the behavior more explicit and avoids hidden failures.

Happy to refine the approach if you think there’s a cleaner way to handle it

sqlglot/dialects/singlestore.py
AdalbertMemSQL: SingleStore `GEOGRAPHY` type can contain `POLYGON` data.
https://docs.singlestore.com/cloud/developer-resources/functional-extensions/working-with-geospatial-features/#polygon

sqlglot/dialects/singlestore.py
georgesittas: Let's keep the unsupported types as is but just emit a warning with `self.unsupported`. Having to reason about how to approximate a type, even though it's not correct at the end, is unnecessary mental overhead.

So, keep this as a set and move the checking logic to the base generator. Only populate the mapping for SingleStore and we can expand it later on for others as needed.

PR #5755, opened 2025-09-02
Fix crash when ARRAY_CONCAT is called with no expressions

ozadari:
❌ The Problem

When ARRAY_CONCAT(...) is used in SQL with a subquery that returns no results (e.g., because of WHERE FALSE), the parsed AST ends up with an empty list of expressions.

For example, in this query:

```
ARRAY_CONCAT((
  SELECT ARRAY_AGG(id)
  FROM x
  WHERE FALSE
))
```

ARRAY_AGG(id) returns nothing → the subquery returns NULL

This becomes ARRAY_CONCAT() with no arguments

In the AST, this looks like ArrayConcat(expressions=[])

The transpiler code tries to reduce() over an empty list, which raises:
```
TypeError: reduce() of empty iterable with no initial value
```

✅ The Fix

If the list of expressions is empty, instead of using reduce(), we now generate an empty array literal ([]) using:
```
exp.Array(expressions=[])
```

This produces valid SQL like:
```
ARRAY_CONCAT([])
```

Which safely returns an empty array instead of crashing.


sqlglot/generator.py
```diff
@@ -4667,7 +4667,10 @@ def explodinggenerateseries_sql(self, expression: exp.ExplodingGenerateSeries) -
     def arrayconcat_sql(self, expression: exp.ArrayConcat, name: str = "ARRAY_CONCAT") -> str:
         exprs = expression.expressions
         if not self.ARRAY_CONCAT_IS_VAR_LEN:
-            rhs = reduce(lambda x, y: exp.ArrayConcat(this=x, expressions=[y]), exprs)
+            if len(exprs) == 0:
+                rhs = exp.Array(expressions=[])
+            else:
+                rhs = reduce(lambda x, y: exp.ArrayConcat(this=x, expressions=[y]), exprs)
```
georgesittas, line 4673: Why are we producing an empty array? Not sure I understand this fix.

ozadari: I've created a minimum case where we do produce empty array and provide it to array concat, it's an edge case but it can (and id) happen

```
WITH x AS ( SELECT 1 AS id), test_cte AS ( SELECT ARRAY_CONCAT(( SELECT id FROM x WHERE FALSE)) AS result ) SELECT * FROM test_cte
```

georgesittas: I still don't follow why we concatenate with an empty array, though:

```sql
SELECT ARRAY_CAT((SELECT id FROM x WHERE FALSE), []) AS result
```

Is `ARRAY_CAT` not an aggregate function in Snowflake? I would expect the resulting SQL to match BigQuery, i.e., a simple query argument inside of that concatenation function.

tests/dialects/test_bigquery.py
```diff
@@ -2713,6 +2713,15 @@ def test_array_agg(self):
                 },
             )
 
+    def test_array_concat(self):
+        self.validate_all(
+            f"WITH x AS ( SELECT 1 AS id), test_cte AS ( SELECT ARRAY_CONCAT(( SELECT id FROM x WHERE FALSE)) AS result ) SELECT * FROM test_cte;",
```
ozadari:
```suggestion
            "WITH x AS ( SELECT 1 AS id), test_cte AS ( SELECT ARRAY_CONCAT(( SELECT id FROM x WHERE FALSE)) AS result ) SELECT * FROM test_cte;",
```

tests/dialects/test_bigquery.py
```diff
@@ -2713,6 +2713,15 @@ def test_array_agg(self):
...
 
+    def test_array_concat(self):
+        self.validate_all(
+            "WITH x AS ( SELECT 1 AS id), test_cte AS ( SELECT ARRAY_CONCAT(( SELECT id FROM x WHERE FALSE)) AS result ) SELECT * FROM test_cte;",
+            write={
+                "snowflake": "WITH x AS (SELECT 1 AS id), test_cte AS (SELECT ARRAY_CAT((SELECT id FROM x WHERE FALSE), []) AS result) SELECT * FROM test_cte",
+            },
+        )
+
+
```
georgesittas:
```suggestion
```

sqlglot/generator.py
georgesittas: Ok, I see what's going on: `ARRAY_CAT` expects two arrays in Snowflake, hence your approach.

ozadari: Hey @georgesittas and thanks for your response,
I hope I understood you correctly and I will try to explain my logic
Array concatantion is creating a singular array from multiple, but indeed BQ and SF handle null in array concatanation differently

image

Where snowflake on cases where you do:
```
SELECT
    ARRAY_CAT(
      null
    ,[1,2,3])
```
return null

so the behaviour is different between those databases and i think big query handles it better
but still, empty array is the correct "Identity Element" for this case

I hope I was clear and I've addresed your concern

PR #5757, opened 2025-09-02
feat(dremio): parse datetype

jasonthomassql:
parse datetype to standard DATE

https://docs.dremio.com/cloud/reference/sql/sql-functions/functions/DATETYPE/

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -74,6 +74,18 @@ def build_date_delta_with_cast_interval(
     return _builder
 
 
+def datetype_handler(args: t.List[exp.Expression], dialect: DialectType) -> exp.Expression:
+    year, month, day = args
+
+    year_val = int(year.this)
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/dremio.py
```diff
@@ -74,6 +74,18 @@ def _builder(args):
     return _builder
 
 
+def datetype_handler(args: t.List[exp.Expression], dialect: DialectType) -> exp.Expression:
+    year, month, day = args
+
+    year_val = int(year.this)
+    month_val = int(month.this)
+    day_val = int(day.this)
```
georgesittas, line 82: Why are we assuming that these arguments are going to be literals here? Isn't this valid Dremio?

```sql
SELECT DATETYPE(x, y, z)
```

jasonthomassql: I think so yes...can't get this to rebase correctly anyway.  Will reopen.

PR #5759, opened 2025-09-02
feat(dremio): parse datetype

jasonthomassql:
handles literals and columns.

https://docs.dremio.com/cloud/reference/sql/sql-functions/functions/DATETYPE/

### DIFF
--- a/sqlglot/dialects/dremio.py
+++ b/sqlglot/dialects/dremio.py
@@ -78,11 +78,9 @@ def datetype_handler(args: t.List[exp.Expression], dialect: DialectType) -> exp.
     year, month, day = args
 
     if all(isinstance(arg, exp.Literal) and arg.is_int for arg in (year, month, day)):
-        # Static literal: DATE 'YYYY-MM-DD'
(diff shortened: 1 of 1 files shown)


tests/dialects/test_dremio.py
```diff
@@ -212,3 +212,7 @@ def test_date_part(self):
             "SELECT DATE_PART('YEAR', date '2021-04-01')",
             "SELECT EXTRACT('YEAR' FROM CAST('2021-04-01' AS DATE))",
         )
+
+    def test_datetype(self):
+        self.validate_identity("DATETYPE(2024,2,2)", "DATE('2024-02-02')")
+        self.validate_identity("DATETYPE(x,y,z)", "CAST(CONCAT(x, '-', y, '-', z) AS DATE)")
```
georgesittas, line 218: Does Dremio support the implicit coercion of integers into strings for this to work? For example, if `x` is `2024`, then would this `CONCAT(x, '-', ...)` work?

jasonthomassql: yes, this is tested valid syntax 

concat(5, '-', 5)
output-> 5-5

PR #5766, opened 2025-09-02
Chore!: remove EXCLUDE keyword mapping to TokenType.EXCEPT

treysp:
`EXCEPT` is an ANSI standard set operator that removes one rowset from another rowset (e.g., `SELECT id FROM tbl1 EXCEPT SELECT id FROM tbl2`). Sqlglot parses it in `parse_set_operation` based on token type.

Some dialects also use except as a star operator for selecting columns (e.g., `SELECT * EXCEPT(id) FROM tbl`). Sqlglot parses those in `_parse_star_operator` based on text match.

DuckDB and Snowflake use the word `EXCLUDE` for the latter operation. Currently, sqlglot incorrectly maps "EXCLUDE" to `TokenType.EXCEPT` for them, which incorrectly mixes the two usages above.

### DIFF
--- a/sqlglot/dialects/duckdb.py
+++ b/sqlglot/dialects/duckdb.py
@@ -304,7 +304,6 @@ class DuckDB(Dialect):
             "CHAR": TokenType.TEXT,
             "DATETIME": TokenType.TIMESTAMPNTZ,
             "DETACH": TokenType.DETACH,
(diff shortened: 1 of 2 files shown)


georgesittas:
@treysp what motivated this change? Let's add tests for it.

georgesittas (approved):

PR #5767, opened 2025-09-03
feat(singlestore): Fixed parsing of columns with table name

AdalbertMemSQL:
Without this column operator, dialect fails to parse expressions like ``` `t0`.`name` ```

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -198,7 +198,7 @@ class SingleStore(MySQL):
         CAST_COLUMN_OPERATORS = {TokenType.COLON_GT, TokenType.NCOLON_GT}
 
         COLUMN_OPERATORS = {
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -198,6 +198,7 @@ class Parser(MySQL.Parser):
         CAST_COLUMN_OPERATORS = {TokenType.COLON_GT, TokenType.NCOLON_GT}
 
         COLUMN_OPERATORS = {
+            TokenType.DOT: None,
             TokenType.COLON_GT: lambda self, this, to: self.expression(
                 exp.Cast,
```
georgesittas: This doesn't look right @AdalbertMemSQL. We should expand `MySQL.Parser.COLUMN_OPERATORS` first. Let's do this and clean up any other redundant entries as well, if any.

PR #5781, opened 2025-09-03
feat(databricks): use regexp_like as it exists

jasonthomassql:
Databricks has a regexp_like function.  This is preferred over the inherited rlike from spark which is missing that function.

https://docs.databricks.com/aws/en/sql/language-manual/functions/regexp_like

### DIFF
--- a/sqlglot/dialects/databricks.py
+++ b/sqlglot/dialects/databricks.py
@@ -104,9 +104,9 @@ class Databricks(Spark):
                 if e.args.get("is_numeric")
                 else self.function_fallback_sql(e)
             ),
-            exp.RegexpLike: lambda self, e: self.function_fallback_sql(e),
         }
 
+        TRANSFORMS.pop(exp.RegexpLike)
         TRANSFORMS.pop(exp.TryCast)
 
         TYPE_MAPPING = {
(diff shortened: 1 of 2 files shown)


georgesittas:
Why is it preferred? What does this PR solve?

jasonthomassql:
Preferred as it supports the standard SQL function REGEXP_LIKE, which is also the canonical representation in sqlglot.  The current inheritance from spark which is missing the function should not implicitly limit its use.  Its a cleaner transform.

It improves readability, aligns with ANSI SQL, and ensures better cross-dialect compatibility.

sqlglot/dialects/databricks.py
```diff
@@ -104,6 +104,7 @@ class Generator(Spark.Generator):
                 if e.args.get("is_numeric")
                 else self.function_fallback_sql(e)
             ),
+            exp.RegexpLike: lambda self, e: self.function_fallback_sql(e),
         }
 
         TRANSFORMS.pop(exp.TryCast)
```
georgesittas:
```suggestion
        }

        TRANSFORMS.pop(exp.RegexpLike)
        TRANSFORMS.pop(exp.TryCast)
```

georgesittas:
> Preferred as it supports the standard SQL function REGEXP_LIKE, which is also the canonical representation in sqlglot. The current inheritance from spark which is missing the function should not implicitly limit its use. Its a cleaner transform.
> 
> It improves readability, aligns with ANSI SQL, and ensures better cross-dialect compatibility.

Fair enough, I see no issue with it. Thanks for the PR! Just cleaned it up slightly.

PR #5785, opened 2025-09-03
feat(singlestore): Added handling of UTC_DATE and exp.CurrentDate

AdalbertMemSQL:
[UTC_DATE](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/utc-date/)
[CURRENT_DATE](https://docs.singlestore.com/cloud/reference/sql-reference/date-and-time-functions/current-date-and-curdate/)

### DIFF
--- a/sqlglot/expressions.py
+++ b/sqlglot/expressions.py
@@ -6006,6 +6006,10 @@ class CurrentUser(Func):
     arg_types = {"this": False}
 
 
+class UtcDate(Func):
+    arg_types = {}
+
+
 class DateAdd(Func, IntervalOp):
     arg_types = {"this": True, "expression": True, "unit": False}
 
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -641,6 +641,7 @@ class Parser(metaclass=_Parser):
         TokenType.TIMESTAMP,
         TokenType.TIMESTAMPTZ,
         TokenType.TRUNCATE,
+        TokenType.UTC_DATE,
         TokenType.WINDOW,
         TokenType.XOR,
         *TYPE_TOKENS,
--- a/sqlglot/tokens.py
+++ b/sqlglot/tokens.py
@@ -423,6 +423,7 @@ class TokenType(AutoName):
     WINDOW = auto()
     WITH = auto()
     UNIQUE = auto()
+    UTC_DATE = auto()
     VERSION_SNAPSHOT = auto()
     TIMESTAMP_SNAPSHOT = auto()
     OPTION = auto()
@@ -857,6 +858,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "UPDATE": TokenType.UPDATE,
         "USE": TokenType.USE,
         "USING": TokenType.USING,
+        "UTC_DATE": TokenType.UTC_DATE,
         "UUID": TokenType.UUID,
         "VALUES": TokenType.VALUES,
         "VIEW": TokenType.VIEW,
--- a/tests/dialects/test_singlestore.py
+++ b/tests/dialects/test_singlestore.py
@@ -560,6 +560,21 @@ class TestSingleStore(Validator):
                 "singlestore": "SELECT DATEDIFF('2009-12-31', '2009-01-01') AS numweeks",
             },
         )
+        self.validate_all(
+            "SELECT CURRENT_DATE()",
+            read={
+                "": "SELECT CURRENT_DATE()",
+                "singlestore": "SELECT CURRENT_DATE",
+            },
+        )
+        self.validate_all(
+            "SELECT UTC_DATE()",
+            read={
+                "": "SELECT CURRENT_DATE('UTC')",
+                "singlestore": "SELECT UTC_DATE",
+            },
+            write={"": "SELECT CURRENT_DATE('UTC')"},
+        )
 
     def test_types(self):
         self.validate_all(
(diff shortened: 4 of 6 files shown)


sqlglot/expressions.py
```diff
@@ -6006,6 +6006,10 @@ class CurrentUser(Func):
     arg_types = {"this": False}
 
 
+class UtcDate(Func):
+    arg_types = {}
```
georgesittas, line 6030: We should investigate if any of the supported dialects supports `UTC_DATE`. Introducing a new `Func` subclass can result in parser errors if its `arg_types` is incompatible with a signature for that function.

For example... if BigQuery supported `UTC_DATE(arg)`, it would break because:

1. The parser automatically recognizes `UTC_DATE` and tries to build a `UtcDate` from it
2. The `arg_types` defined here specifies that no arguments are expected, so validation fails

sqlglot/parser.py
```diff
@@ -641,6 +641,7 @@ class Parser(metaclass=_Parser):
         TokenType.TIMESTAMP,
         TokenType.TIMESTAMPTZ,
         TokenType.TRUNCATE,
+        TokenType.UTC_DATE,
```
georgesittas, line 644: Please add a test like `select utc_date` in `identity.sql`. I think this will fail to parse, in which case you should add this token to `ID_VAR_TOKENS`.

sqlglot/tokens.py
```diff
@@ -857,6 +858,7 @@ class Tokenizer(metaclass=_Tokenizer):
         "UPDATE": TokenType.UPDATE,
         "USE": TokenType.USE,
         "USING": TokenType.USING,
+        "UTC_DATE": TokenType.UTC_DATE,
```
georgesittas: Why is this added in the base tokenizer instead of just singlestore?

tests/dialects/test_singlestore.py
```diff
@@ -560,6 +560,21 @@ def test_time_functions(self):
...
         )
+        self.validate_all(
+            "SELECT CURRENT_DATE()",
+            read={
+                "": "SELECT CURRENT_DATE()",
+                "singlestore": "SELECT CURRENT_DATE",
+            },
+        )
+        self.validate_all(
+            "SELECT UTC_DATE()",
```
georgesittas, line 571: Is this form equivalent to the `UTC_DATE` one (i.e., no parentheses)?

tests/dialects/test_singlestore.py
```diff
@@ -560,6 +560,21 @@ def test_time_functions(self):
                 "singlestore": "SELECT DATEDIFF('2009-12-31', '2009-01-01') AS numweeks",
             },
         )
+        self.validate_all(
+            "SELECT CURRENT_DATE()",
```
georgesittas, line 564: Ditto

tests/dialects/test_singlestore.py
AdalbertMemSQL: Yep, they are the same.
```
MySQL [db]> SELECT UTC_DATE();
+------------+
| UTC_DATE() |
+------------+
| 2025-09-03 |
+------------+
1 row in set (0.001 sec)

MySQL [db]> SELECT UTC_DATE;
+------------+
| UTC_DATE   |
+------------+
| 2025-09-03 |
+------------+
1 row in set (0.001 sec)

```

sqlglot/tokens.py
AdalbertMemSQL: Will move this

sqlglot/parser.py
AdalbertMemSQL: Added a test. Looks like it succeeds.

sqlglot/expressions.py
AdalbertMemSQL: MySQL and Oracle have the `UTC_DATE` function with the same signature
https://www.w3resource.com/mysql/date-and-time-functions/mysql-utc_date-function.php
https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/date-and-time-functions.html#function_utc-date

georgesittas: Can you verify whether other dialects do as well?

sqlglot/parser.py
georgesittas: What does it parse into? It needs to be a `Column(Identifier("utc_date"))` and not a `UtcDate` expression.

sqlglot/expressions.py
AdalbertMemSQL: I checked other dialects and didn't find `UTC_DATE` function

tests/dialects/test_singlestore.py
AdalbertMemSQL:
```
MySQL [db]> SELECT CURRENT_DATE();
+----------------+
| CURRENT_DATE() |
+----------------+
| 2025-09-03     |
+----------------+
1 row in set (0.021 sec)

MySQL [db]> SELECT CURRENT_DATE;
+--------------+
| CURRENT_DATE |
+--------------+
| 2025-09-03   |
+--------------+
1 row in set (0.001 sec)
```

sqlglot/parser.py
AdalbertMemSQL:
```
Select(
  expressions=[
    Column(
      this=Identifier(this=UTC_DATE, quoted=False))])
```

georgesittas (approved):

sqlglot/expressions.py
georgesittas: Thank you

PR #5801, opened 2025-09-04
Fix(parser): address edge case where `values` is used as an identifier

georgesittas:


sqlglot/parser.py
```diff
@@ -7319,10 +7319,13 @@ def _parse_expressions(self) -> t.List[exp.Expression]:
         return self._parse_csv(self._parse_expression)
 
     def _parse_select_or_expression(self, alias: bool = False) -> t.Optional[exp.Expression]:
-        return self._parse_select() or self._parse_set_operations(
```
georgesittas, line 7322: I didn't dig super deep but the previous order of method calls consumed `VALUES` too early and led to a problematic parser code path that resulted in a failure:

```
Required keyword: 'expressions' missing for <class 'sqlglot.expressions.Values'>
```

PR #5802, opened 2025-09-04
Fix(snowflake): support nulls_first arg for array_sort

treysp:


sqlglot/dialects/snowflake.py
```diff
@@ -1272,6 +1282,7 @@ class Generator(generator.Generator):
                 ]
             ),
             exp.SHA: rename_func("SHA1"),
+            exp.SortArray: _sort_array_sql,
```
georgesittas:
```suggestion
            exp.SortArray: rename_func("ARRAY_SORT"),
```

sqlglot/dialects/snowflake.py
```diff
@@ -482,6 +482,15 @@ def _eliminate_dot_variant_lookup(expression: exp.Expression) -> exp.Expression:
...
 
+def _sort_array_sql(self: Snowflake.Generator, expression: exp.SortArray) -> str:
+    return self.func(
+        "ARRAY_SORT",
+        expression.this,
+        expression.args.get("asc"),
+        expression.args.get("nulls_first"),
+    )
+
+
```
georgesittas: I believe we can get rid of this:
```suggestion
```

sqlglot/dialects/snowflake.py
treysp: That's what I did first and the snowflake generator tests weren't passing, but clearly that was due to user error...

PR #5803, opened 2025-09-04
feat(snowflake): add support for REVERSE function

fivetran-BradfordPaskewitz:
add support for Snowflake REVERSE function

documentation: https://docs.snowflake.com/en/sql-reference/functions/reverse

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -496,6 +496,14 @@ class Snowflake(Dialect):
     ALTER_TABLE_ADD_REQUIRED_FOR_EACH_COLUMN = False
     TRY_CAST_REQUIRES_STRING = True
 
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        **{
+            expr_type: lambda self, e: self._annotate_by_args(e, "this")
+            for expr_type in (exp.Reverse,)
+        },
+    }
+
     TIME_MAPPING = {
         "YYYY": "%Y",
         "yyyy": "%Y",
--- a/tests/fixtures/optimizer/annotate_functions.sql
+++ b/tests/fixtures/optimizer/annotate_functions.sql
@@ -1535,6 +1535,22 @@ STRING;
 LEAST(x::DECIMAL(18, 2));
 DECIMAL(18, 2);
 
+# dialect: snowflake
+REVERSE('Hello, world!');
+VARCHAR;
+
+# dialect: snowflake
+REVERSE(tbl.str_col);
+TEXT;
+
+# dialect: snowflake
+REVERSE(tbl.bin_col);
+BINARY;
+
+# dialect: snowflake
+REVERSE(NULL);
+NULL;
+
 --------------------------------------
 -- T-SQL
 --------------------------------------


sqlglot/dialects/snowflake.py
```diff
@@ -611,6 +616,7 @@ class Parser(parser.Parser):
             "REGEXP_SUBSTR": _build_regexp_extract(exp.RegexpExtract),
             "REGEXP_SUBSTR_ALL": _build_regexp_extract(exp.RegexpExtractAll),
             "REPLACE": build_replace_with_optional_replacement,
+            "REVERSE": exp.Reverse.from_arg_list,
```
georgesittas: This is redundant, because:

1. `Reverse` is a `Func` subclass ([source](https://github.com/tobymao/sqlglot/blob/3726b33bb6b4ab286617f510e96e1fbd27c429f3/sqlglot/expressions.py#L6816-L6817))
2. It is thus included in both `ALL_FUNCTIONS` and `FUNCTION_BY_NAME` ([source](https://github.com/tobymao/sqlglot/blob/3726b33bb6b4ab286617f510e96e1fbd27c429f3/sqlglot/expressions.py#L7570-L7571))
3. The base `Parser` automatically parses `REVERSE` calls into `Reverse` ([source](https://github.com/tobymao/sqlglot/blob/3726b33bb6b4ab286617f510e96e1fbd27c429f3/sqlglot/parser.py#L197))

The "sql name" of a `Func` subclass is derived by doing a pascal-to-snake-case transformation. So `FooBar` becomes `FOO_BAR` and thus `FOO_BAR` is automatically recognized by the parser due to (3).

So, since Snowflake inherits from the base parser, it already recognizes `REVERSE`.

sqlglot/dialects/snowflake.py
```diff
@@ -496,6 +496,11 @@ class Snowflake(Dialect):
     ALTER_TABLE_ADD_REQUIRED_FOR_EACH_COLUMN = False
     TRY_CAST_REQUIRES_STRING = True
 
+    ANNOTATORS = {
+        **Dialect.ANNOTATORS,
+        exp.Reverse: lambda self, e: self._annotate_by_args(e, "this"),
+    }
```
georgesittas: I'm expecting the "annotate by `this`" pattern to emerge frequently. I suggest we mimic what BigQuery does [here](https://github.com/tobymao/sqlglot/blob/3726b33bb6b4ab286617f510e96e1fbd27c429f3/sqlglot/dialects/bigquery.py#L514-L547).

tests/dialects/test_snowflake.py
```diff
@@ -3032,3 +3032,22 @@ def test_create_sequence(self):
                 "duckdb": "CREATE SEQUENCE seq START WITH 1 INCREMENT BY 1",
             },
         )
+
+    def test_reverse_function(self):
```
georgesittas: We can get rid of this method– we already cover `REVERSE` in `test_dialect.py`. FWIW, I usually do a quick keyword search and then search by `tests/` to see whether we have coverage before adding a new test.

georgesittas (commented):
Let's start linking any relevant docs in the PR description.

sqlglot/dialects/snowflake.py
fivetran-BradfordPaskewitz: Ok got it, thank you for the explanation, that makes sense

sqlglot/dialects/snowflake.py
fivetran-BradfordPaskewitz: done

tests/dialects/test_snowflake.py
fivetran-BradfordPaskewitz: good to know, thanks, removed

PR #5804, opened 2025-09-04
fix(duckdb): transpile joins without ON/USING to CROSS JOIN

### LINKED ISSUE #5795: Support Snowflake JOIN with no join conditions
Snowflake is interesting where they allow specifying JOINs with no join conditions. This is treated as a CROSS join, however this doesn't transpile nicely into DuckDB where
```
SELECT *
FROM table_a
JOIN table_b
```
will transpiles into DuckDB as
```
SELECT *
FROM table_a
JOIN table_b
```
Which will throw a `syntax error`

Would love if possible to transpile this into a cross join if there are no join conditions

Gotta manually regex this!

**Additional context**
n/a

geooo109:
Fixes #5795

```
table1 = {id: int, pay1: string} populated with some values
table2 = {id: int, pay2: string} empty

DuckDB, Presto, Trino, Redshift, MySQL, Postgres, T-SQL:
SELECT * FROM table1, table2; => no rows
SELECT * FROM table1 LEFT JOIN table2; => not supported
SELECT * FROM table1 LEFT JOIN table2 ON True; => table1 rows + rows with null for table2 

Hive, Spark, Databricks, SQLite:
SELECT * FROM table1, table2; => no rows
SELECT * FROM table1 LEFT JOIN table2; => table1 rows + rows with null for table2 
SELECT * FROM table1 LEFT JOIN table2 ON True; => table1 rows + rows with null for table2 
```

In order to fix this transpilation:
1. For the `Hive` and `SQLite` make the parsing of join to automatically add an `ON TRUE`
2. For the base `duckDB Generator`, checking if the `Join` expressions contains `ON, USING or a JOIN method` and generate based on this.

### DIFF
--- a/sqlglot/dialects/hive.py
+++ b/sqlglot/dialects/hive.py
@@ -10,7 +10,6 @@ from sqlglot.dialects.dialect import (
     DATE_ADD_OR_SUB,
     Dialect,
     NormalizationStrategy,
-    add_join_on_true,
     approx_count_distinct_sql,
     arg_max_or_min_no_count,
     datestrtodate_sql,
@@ -310,6 +309,7 @@ class Hive(Dialect):
         STRICT_CAST = False
         VALUES_FOLLOWED_BY_PAREN = False
         JOINS_HAVE_EQUAL_PRECEDENCE = True
+        ADD_JOIN_ON_TRUE = True
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
@@ -377,12 +377,6 @@ class Hive(Dialect):
                 expressions=self._parse_wrapped_csv(self._parse_property)
             ),
         }
-
-        def _parse_join(
-            self, skip_join_token: bool = False, parse_bracket: bool = False
-        ) -> t.Optional[exp.Join]:
-            join = super()._parse_join(skip_join_token=skip_join_token, parse_bracket=parse_bracket)
-            return add_join_on_true(join)
 
         def _parse_transform(self) -> t.Optional[exp.Transform | exp.QueryTransform]:
             if not self._match(TokenType.L_PAREN, advance=False):
--- a/sqlglot/dialects/sqlite.py
+++ b/sqlglot/dialects/sqlite.py
@@ -6,7 +6,6 @@ from sqlglot import exp, generator, parser, tokens, transforms
 from sqlglot.dialects.dialect import (
     Dialect,
     NormalizationStrategy,
-    add_join_on_true,
     any_value_to_max_sql,
     arrow_json_extract_sql,
     concat_to_dpipe_sql,
@@ -111,6 +110,7 @@ class SQLite(Dialect):
         STRING_ALIASES = True
         ALTER_RENAME_REQUIRES_COLUMN = False
         JOINS_HAVE_EQUAL_PRECEDENCE = True
+        ADD_JOIN_ON_TRUE = True
 
         FUNCTIONS = {
             **parser.Parser.FUNCTIONS,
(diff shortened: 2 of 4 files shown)


geooo109:
Will validate if the join `method` checking is needed

sqlglot/dialects/duckdb.py
```diff
@@ -1003,6 +1003,17 @@ def join_sql(self, expression: exp.Join) -> str:
                 # DuckDB doesn't, but we can just add a dummy ON clause that is always true
                 return super().join_sql(expression.on(exp.true()))
 
+            # Convert INNER/OUTER JOIN without ON clause to CROSS JOIN
+            kind = expression.args.get("kind")
```
georgesittas: Nit: you can get rid of this and in line 1012 do `and not expression.kind in ("INNER", "OUTER")`.

sqlglot/dialects/duckdb.py
```diff
@@ -1003,6 +1003,17 @@ def join_sql(self, expression: exp.Join) -> str:
...
+            if (
+                not expression.args.get("using")
+                and not expression.args.get("on")
+                and not expression.method
+                and (not kind or kind in ("INNER", "OUTER"))
+            ):
+                expression.args.pop("side", None)
+                expression.args.pop("kind", None)
```
VaggelisD: It looks to me like the condition above this new code tries to do the same thing essentially, i.e it patches `LEFT JOIN UNNEST(...)` into `LEFT JOIN UNNEST(...) ON TRUE`. 

Should we simply generalize the two variants by adding `ON TRUE`?

sqlglot/dialects/duckdb.py
geooo109: Did this to avoid 
```
input
bigquery: SELECT t.c1, h.c2, s.c3 FROM t1 AS t, UNNEST(t.t2) AS h, UNNEST(h.t3) AS s

output
duckdb: SELECT t.c1, h.c2, s.c3 FROM t1 AS t, UNNEST(t.t2) AS _t0(h), UNNEST(h.t3) AS _t1(s)

instead of (before this PR)
duckdb: SELECT t.c1, h.c2, s.c3 FROM t1 AS t CROSS JOIN UNNEST(t.t2) AS _t0(h) CROSS JOIN UNNEST(h.t3) AS _t1(s)
```
which was in various tests. 

but yeah we can simplify. by moving `kind` in.

sqlglot/dialects/hive.py
```diff
@@ -377,6 +378,12 @@ class Parser(parser.Parser):
             ),
         }
 
+        def _parse_join(
+            self, skip_join_token: bool = False, parse_bracket: bool = False
+        ) -> t.Optional[exp.Join]:
+            join = super()._parse_join(skip_join_token=skip_join_token, parse_bracket=parse_bracket)
+            return add_join_on_true(join)
```
georgesittas: Let's not do it like this. Add a flag in the parser and apply this transformation if it's set in the base class.

sqlglot/dialects/sqlite.py
```diff
@@ -143,6 +144,12 @@ def _parse_attach_detach(self, is_attach=True) -> exp.Attach | exp.Detach:
                 else self.expression(exp.Detach, this=this)
             )
 
+        def _parse_join(
+            self, skip_join_token: bool = False, parse_bracket: bool = False
+        ) -> t.Optional[exp.Join]:
+            join = super()._parse_join(skip_join_token=skip_join_token, parse_bracket=parse_bracket)
+            return add_join_on_true(join)
```
georgesittas: Ditto

PR #5809, opened 2025-09-05
Update(Snowflake): Update tests for concat string function

fivetran-amrutabhimsenayachit:
Added tests string function: concat

Refer: 

- https://docs.snowflake.com/en/sql-reference/functions/concat

- https://docs.snowflake.com/en/sql-reference/functions/concat_ws

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -502,7 +502,7 @@ class Snowflake(Dialect):
             expr_type: lambda self, e: self._annotate_by_args(e, "this")
             for expr_type in (exp.Reverse,)
         },
-        exp.ConcatWs: lambda self, e: self._annotate_concat_ws(e),
+        exp.ConcatWs: lambda self, e: self._annotate_by_args(e, "expressions"),
     }
 
     TIME_MAPPING = {
--- a/sqlglot/optimizer/annotate_types.py
+++ b/sqlglot/optimizer/annotate_types.py
@@ -651,9 +651,3 @@ class TypeAnnotator(metaclass=_TypeAnnotator):
             self._set_type(expression, exp.DataType.Type.UNKNOWN)
 
         return expression
-
-    def _annotate_concat_ws(self, expression: exp.ConcatWs) -> exp.ConcatWs:
-        self._annotate_by_args(expression, "expressions")
-        if expression.type and expression.type.this == exp.DataType.Type.TEXT:
-            self._set_type(expression, exp.DataType.Type.VARCHAR)
-        return expression


georgesittas (approved):
Given Snowflake's docs say:

> The input expressions must all be strings, or all be binary values.

Let's make sure we test the binary values as well.

fivetran-amrutabhimsenayachit:
> Given Snowflake's docs say:
> 
> > The input expressions must all be strings, or all be binary values.
> 
> Let's make sure we test the binary values as well.

The base `Dialect` class has `exp.Concat` in its `TYPE_TO_EXPRESSIONS[VARCHAR]` mapping . Thus,  `CONCAT` always returns `VARCHAR`, regardless of input types `BINARY` or `NULL` values.
We will have to override this logic for both `CONCAT` and `CONCAT_WS` methods for `snowflake`.

georgesittas:
@fivetran-amrutabhimsenayachit have you created a virtual environment for Python? Seems like there may be a mismatch between the ruff version SQLGlot installs vs what you have locally. Try the following:

```
$ python -m venv .venv
$ source .venv/bin/activate
$ make install-dev
$ make style
```

This should fix any formatting issues.

> The base Dialect class has exp.Concat in its TYPE_TO_EXPRESSIONS[VARCHAR] mapping . Thus, CONCAT always returns VARCHAR, regardless of input types BINARY or NULL values.
> We will have to override this logic for both CONCAT and CONCAT_WS methods for snowflake.

Overriding makes sense, but with a quick look I feel like the current implementations are quite complicated. What about using `_annotate_by_arg` for both of them?

fivetran-amrutabhimsenayachit:
> have you created a virtual environment for Python
@georgesittas , No, I do not have virtual environment for python. Seems like the new functions added have not been formatted properly. I did not run the make style before commit. Thanks for these commands, will try these.

tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -1555,10 +1555,34 @@ TEXT;
 REVERSE(tbl.bin_col);
 BINARY;
 
+# dialect: snowflake
+REVERSE(NULL);
+NULL;
```
georgesittas, line 1560: I think this was added by mistake– I got rid of it in main. Let's git rebase off of it.

sqlglot/dialects/snowflake.py
```diff
@@ -502,6 +502,7 @@ class Snowflake(Dialect):
             expr_type: lambda self, e: self._annotate_by_args(e, "this")
             for expr_type in (exp.Reverse,)
         },
+        exp.ConcatWs: lambda self, e: self._annotate_concat_ws(e),
```
georgesittas: I believe we can simply do:

```python
exp.ConcatWs: lambda self, e: self._annotate_by_args(e, "expressions")
```

Don't think that we need `_annotate_concat_ws`. Lemme know if I missed something.

PR #5810, opened 2025-09-05
Fix: parsing quoted built-in data types

treysp:
Consider a cast to a built-in data type, where the type is quoted: `1::"integer"`

We currently parse this into the correct underlying type for dialects that allow this AND support user-defined types (like tsql).

This PR allows us to parse correctly for dialects that do not support user-defined types (like redshift).

### DIFF
--- a/sqlglot/parser.py
+++ b/sqlglot/parser.py
@@ -7,7 +7,7 @@ import typing as t
 from collections import defaultdict
 
 from sqlglot import exp
-from sqlglot.errors import ErrorLevel, ParseError, concat_messages, merge_errors
+from sqlglot.errors import ErrorLevel, ParseError, TokenError, concat_messages, merge_errors
 from sqlglot.helper import apply_index_offset, ensure_list, seq_get
 from sqlglot.time import format_time
 from sqlglot.tokens import Token, Tokenizer, TokenType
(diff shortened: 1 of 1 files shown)


sqlglot/parser.py
```diff
@@ -5303,22 +5303,30 @@ def _parse_types(
                 any_token=False, tokens=(TokenType.VAR,)
             )
             if isinstance(identifier, exp.Identifier):
-                tokens = self.dialect.tokenize(identifier.sql(dialect=self.dialect))
+                tokens = self.dialect.tokenize(identifier.name)
 
                 if len(tokens) != 1:
                     self.raise_error("Unexpected identifier", self._prev)
```
georgesittas: I think we may need to get rid of this condition, because you can end up with multiple tokens for some user-defined types, e.g. imagine you do `CREATE TYPE "foo bar baz"`– then tokenizing the `name` of that will yield 3 tokens.

Let's instead check whether there's exactly one token and whether that token is in `TYPE_TOKENS`. If that's the case then we can set the `type_token` like you did, otherwise we just use the identifier and convert it to a UDT like we used to.

I think we also need to wrap `tokenize` in a try-except because if you have a UDT like `"foo'bar"`, the tokenization will fail due to the missing `'`. Weird edge case but yeah.

sqlglot/parser.py
```diff
@@ -5303,22 +5303,30 @@ def _parse_types(
...
-                    this = self._parse_user_defined_type(identifier)
+                    type_token = tokens[0].token_type
                 else:
-                    self._retreat(self._index - 1)
-                    return None
+                    # retain quotes
+                    tokens = self.dialect.tokenize(identifier.sql(dialect=self.dialect))
+
+                    if len(tokens) != 1:
+                        self.raise_error("Unexpected identifier", self._prev)
```
georgesittas: Why are we re-tokenizing here?

sqlglot/parser.py
```diff
@@ -5298,28 +5298,30 @@ def _parse_types(
...
                 any_token=False, tokens=(TokenType.VAR,)
             )
             if isinstance(identifier, exp.Identifier):
-                tokens = self.dialect.tokenize(identifier.sql(dialect=self.dialect))
-
-                if len(tokens) != 1:
-                    self.raise_error("Unexpected identifier", self._prev)
+                try:
+                    tokens = self.dialect.tokenize(identifier.name)
+                except TokenError:
```
VaggelisD, line 5311: Should we `self.raise_error(...)` here? Otherwise, we'll now silently continue on

sqlglot/parser.py
```diff
@@ -5298,28 +5298,30 @@ def _parse_types(
...
+                if tokens and len(tokens) == 1 and tokens[0].token_type in self.TYPE_TOKENS:
+                    type_token = tokens[0].token_type
                 else:
-                    self._retreat(self._index - 1)
-                    return None
+                    if self.dialect.SUPPORTS_USER_DEFINED_TYPES:
+                        this = self._parse_user_defined_type(identifier)
+                    else:
+                        self._retreat(self._index - 1)
+                        return None
```
VaggelisD, line 5321: [Nit] Could we revert the if/elif/else chain to reduce some of the indentation?

tests/dialects/test_postgres.py
```diff
@@ -805,6 +805,12 @@ def test_postgres(self):
         )
         self.assertIsInstance(self.parse_one("id::UUID"), exp.Cast)
 
+        self.validate_identity('1::"int"', "CAST(1 AS INT)")
+        assert parse_one('1::"int"', read="postgres").to.is_type(exp.DataType.Type.INT)
```
VaggelisD, line 809: We don't need the generic `parse_one` here if we read into the same dialect as in the testing file. 

We can grab the AST as `ast = self.validate_identity(...)` or do `ast = self.parse_one(...)` which will set the appropriate dialect.

tests/dialects/test_redshift.py
```diff
@@ -698,3 +698,10 @@ def test_analyze(self):
         self.validate_identity("ANALYZE VERBOSE TBL")
         self.validate_identity("ANALYZE TBL PREDICATE COLUMNS")
         self.validate_identity("ANALYZE TBL ALL COLUMNS")
+
+    def test_cast(self):
+        self.validate_identity('1::"int"', "CAST(1 AS INTEGER)")
+        assert parse_one('1::"int"', read="redshift").to.is_type(exp.DataType.Type.INT)
```
VaggelisD, line 704: Ditto

sqlglot/parser.py
georgesittas: No, this is intentional. See previous threads & Slack.

georgesittas:
Merging this as is, will clean up tests afterwards.

PR #5811, opened 2025-09-05
feat(snowflake): add support for TRIM function

fivetran-BradfordPaskewitz:
add support for Snowflake TRIM function

documentation: https://docs.snowflake.com/en/sql-reference/functions/trim


sqlglot/dialects/snowflake.py
```diff
@@ -500,7 +500,10 @@ class Snowflake(Dialect):
         **Dialect.ANNOTATORS,
         **{
             expr_type: lambda self, e: self._annotate_by_args(e, "this")
-            for expr_type in (exp.Reverse,)
+            for expr_type in (
+                exp.Reverse,
+                exp.Trim,
```
geooo109: In snowflake `TRIM` returns a static type `VARCHAR`, which is captured in the `class Dialect`. In these cases just adding a test for snowflake is fine. The different string types will be captured by the mapping I proposed bellow, after George replies, we are good to go.

tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -1555,6 +1555,30 @@ TEXT;
...
+TRIM('hello world', 'hello');
+VARCHAR;
+
+# dialect: snowflake
+TRIM(tbl.bin_col);
+BINARY;
+
+# dialect: snowflake
+TRIM(tbl.bin_col, tbl.bin_col);
+BINARY;
```
geooo109: Based on the docs + the queries I tried on snowflake, `TRIM` doesn't work for `BINARY - VARBINARY` values. So, we should remove these tests.

tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -1555,6 +1555,30 @@ TEXT;
...
+
+# dialect: snowflake
+TRIM(tbl.str_col);
+TEXT;
+
+# dialect: snowflake
+TRIM(tbl.str_col, tbl.str_col);
+TEXT;
```
geooo109: This is correct because [STRING, TEXT, NVARCHAR, NVARCHAR2, CHAR VARYING, NCHAR VARYING = VARCHAR](https://docs.snowflake.com/en/sql-reference/data-types-text#string-text-nvarchar-nvarchar2-char-varying-nchar-varying) 
We should add them in the`TYPE_MAPPING` of Snowflake's `Generator class`, right? @georgesittas

geooo109 (commented):
Great work! Left some small comments.

tests/fixtures/optimizer/annotate_functions.sql
georgesittas: Sure, seems like only two are not being mapped to `VARCHAR`:

```python
>>> s = "STRING, TEXT, NVARCHAR, NVARCHAR2, CHAR VARYING, NCHAR VARYING"
>>> types = s.split(", ")
>>> types
['STRING', 'TEXT', 'NVARCHAR', 'NVARCHAR2', 'CHAR VARYING', 'NCHAR VARYING']
>>> import sqlglot
>>> for t in types:
...     print(sqlglot.transpile(f"x::{t}", "snowflake"))
...
['CAST(x AS TEXT)']
['CAST(x AS TEXT)']
['CAST(x AS VARCHAR)']
['CAST(x AS VARCHAR)']
['CAST(x AS VARCHAR)']
['CAST(x AS VARCHAR)']
```

geooo109: @georgesittas thanks for checking!

@fivetran-BradfordPaskewitz So, we should add them (`STRING` and `TEXT`) in the`TYPE_MAPPING` and we are good to go (you can also check the `bigquery.py` for help, we extend the `TYPE_MAPPING` for BQ).

sqlglot/dialects/snowflake.py
fivetran-BradfordPaskewitz: ok sounds good, thanks

tests/fixtures/optimizer/annotate_functions.sql
fivetran-BradfordPaskewitz: done

tests/fixtures/optimizer/annotate_functions.sql
fivetran-BradfordPaskewitz: I added mapping for TEXT -> VARCHAR and updated tests accordingly that previously expected TEXT. It seems like we don't explicitly need to add STRING mapping since it is an alias for TEXT?

geooo109: Yep, correct, no need for that. The tokenizer makes `STRING` => `TokenType.TEXT`

tests/dialects/test_snowflake.py
```diff
@@ -2639,23 +2642,8 @@ def test_swap(self):
...
-                "hive": "CAST(CAST('2020-01-01' AS DATE) AS STRING)",
-                "snowflake": "CAST(TRY_CAST('2020-01-01' AS DATE) AS TEXT)",
-            },
-        )
-        self.validate_all(
-            "TRY_CAST('val' AS TEXT)",
-            read={
-                "hive": "CAST('val' AS STRING)",
-                "snowflake": "TRY_CAST('val' AS TEXT)",
-            },
```
geooo109: Why did we remove these tests? The following should pass. 
```
        self.validate_all("TRY_CAST('foo' AS VARCHAR)", read={"hive": "CAST('foo' AS STRING)"})
        self.validate_all("CAST(5 + 5 AS VARCHAR)", read={"hive": "CAST(5 + 5 AS STRING)"})
        self.validate_all(
            "CAST(TRY_CAST('2020-01-01' AS DATE) AS VARCHAR)",
            read={
                "hive": "CAST(CAST('2020-01-01' AS DATE) AS STRING)",
                "snowflake": "CAST(TRY_CAST('2020-01-01' AS DATE) AS VARCHAR)",
            },
        )
        self.validate_all(
            "TRY_CAST('val' AS VARCHAR)",
            read={
                "hive": "CAST('val' AS STRING)",
                "snowflake": "TRY_CAST('val' AS VARCHAR)",
            },
```

geooo109 (approved):
Nice work, thanks for the type mapping changes!

tests/dialects/test_snowflake.py
fivetran-BradfordPaskewitz: You are correct, added your suggestion

PR #5812, opened 2025-09-05
feat(snowflake): add support for UPPER function

fivetran-BradfordPaskewitz:
add support for Snowflake UPPER function

documentation: https://docs.snowflake.com/en/sql-reference/functions/upper

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -500,7 +500,10 @@ class Snowflake(Dialect):
         **Dialect.ANNOTATORS,
         **{
             expr_type: lambda self, e: self._annotate_by_args(e, "this")
-            for expr_type in (exp.Reverse,)
+            for expr_type in (
+                exp.Reverse,
+                exp.Upper,
+            )
         },
     }
 
--- a/tests/fixtures/optimizer/annotate_functions.sql
+++ b/tests/fixtures/optimizer/annotate_functions.sql
@@ -1555,6 +1555,18 @@ TEXT;
 REVERSE(tbl.bin_col);
 BINARY;
 
+# dialect: snowflake
+UPPER('Hello, world!');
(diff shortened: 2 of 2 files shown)


tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -1555,6 +1555,18 @@ TEXT;
...
+UPPER('Hello, world!');
+VARCHAR;
+
+# dialect: snowflake
+UPPER(tbl.str_col);
+TEXT;
+
+# dialect: snowflake
+UPPER(tbl.bin_col);
+BINARY;
```
geooo109: Based on the docs + the queries I tried on snowflake, `UPPER` doesn't work for `BINARY - VARBINARY` values. So, we should remove these tests.

sqlglot/dialects/snowflake.py
```diff
@@ -500,7 +500,10 @@ class Snowflake(Dialect):
         **Dialect.ANNOTATORS,
         **{
             expr_type: lambda self, e: self._annotate_by_args(e, "this")
-            for expr_type in (exp.Reverse,)
+            for expr_type in (
+                exp.Reverse,
+                exp.Upper,
```
geooo109: Same as `TRIM` PR, we can skip this because it is already captured in the base dialect.

geooo109 (commented):
Nice work! left small comments

tests/fixtures/optimizer/annotate_functions.sql
fivetran-BradfordPaskewitz: done

sqlglot/dialects/snowflake.py
fivetran-BradfordPaskewitz: ok understood, removed

PR #5818, opened 2025-09-07
feat(singlestore): Added handling of exp.JSONArray

AdalbertMemSQL:
[JSON_BUILD_ARRAY](https://docs.singlestore.com/cloud/reference/sql-reference/json-functions/json-build-array/)


sqlglot/dialects/singlestore.py
```diff
@@ -162,6 +162,7 @@ class Parser(MySQL.Parser):
                 json_type="JSON",
             ),
             "JSON_PRETTY": exp.JSONFormat.from_arg_list,
+            "JSON_BUILD_ARRAY": lambda args: exp.JSONArray(expressions=args),
```
VaggelisD, line 167: I merged https://github.com/tobymao/sqlglot/pull/5817 but I just realized that we could also parse `JSON_BUILD_OBJECT` as we do here for `JSON_BUILD_ARRAY`, since it uses `expressions` only too. Could you refactor that in a follow up PR if its doable?

VaggelisD (approved):
You'd need to rebase this branch to latest main in order to remove the `import exp` conflict, thanks for the PR 👍

PR #5822, opened 2025-09-08
feat(singlestore): Implemented parsing and generation of exp.MatchAgainst

AdalbertMemSQL:
[MATCH AGAINST](https://docs.singlestore.com/cloud/reference/sql-reference/full-text-search-functions/match/)

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -235,6 +235,21 @@ class SingleStore(MySQL):
         COLUMN_OPERATORS.pop(TokenType.DHASH_ARROW)
         COLUMN_OPERATORS.pop(TokenType.PLACEHOLDER)
 
+        def _parse_match_against(self) -> exp.MatchAgainst:
+            if self._match_text_seq("TABLE"):
+                expressions = []
+                table = self._parse_table()
+                if table:
+                    expressions = [table]
+            else:
+                expressions = self._parse_csv(self._parse_column)
+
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -1667,3 +1682,16 @@ def currenttime_sql(self, expression: exp.CurrentTime) -> str:
...
+            expressions_sqls = []
+            for expr in expression.expressions:
+                if isinstance(expr, exp.Table):
+                    expressions_sqls.append(f"TABLE {self.sql(expr)}")
+                else:
+                    expressions_sqls.append(self.sql(expr))
+
+            return (
+                f"{self.func('MATCH', *expressions_sqls)} AGAINST ({self.sql(expression, 'this')})"
+            )
```
georgesittas: How is this different from what [we support](https://github.com/tobymao/sqlglot/blob/e3cb0766bd5c3ccb31ea52cfc76201f548798dc1/tests/dialects/test_mysql.py#L534) for MySQL?

AdalbertMemSQL: SingleStore supports the following syntax MATCH(TABLE t) AGAINST (...)

georgesittas: Perhaps we can consolidate by enhancing MySQL's parser? Instead of overriding the whole thing here.

georgesittas (approved):

PR #5824, opened 2025-09-08
feat(singlestore): Implemented generation of exp.StrPosition

AdalbertMemSQL:
[LOCATE](https://docs.singlestore.com/cloud/reference/sql-reference/string-functions/locate/)

### DIFF
--- a/sqlglot/dialects/singlestore.py
+++ b/sqlglot/dialects/singlestore.py
@@ -417,6 +417,11 @@ class SingleStore(MySQL):
                 ),
                 "RLIKE",
             ),
+            exp.StrPosition: unsupported_args("occurrence")(
+                lambda self, e: self.func(
+                    "LOCATE", e.args.get("substr"), e.this, e.args.get("position")
+                )
+            ),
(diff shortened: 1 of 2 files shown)


sqlglot/dialects/singlestore.py
```diff
@@ -417,6 +417,11 @@ def _unicode_substitute(m: re.Match[str]) -> str:
                 ),
                 "RLIKE",
             ),
+            exp.StrPosition: unsupported_args("occurrence")(
+                lambda self, e: self.func(
+                    "LOCATE", e.args.get("substr"), e.this, e.args.get("position")
+                )
+            ),
```
georgesittas, line 424: Doesn't the MySQL [generator](https://github.com/tobymao/sqlglot/blob/cad4fd0c5b0ec90e693fa6883af0ab287b921019/sqlglot/dialects/mysql.py#L750-L752) suffice?

AdalbertMemSQL: Nice catch. A MySQL variant is sufficient

Issue #5815, opened 2025-09-07
BEGIN is transpiled incorrectly to Trino


karakanb:
**Fully reproducible code snippet**
```
 uv run --no-project --with sqlglot==27.12.0 python -c 'import sqlglot; print(sqlglot.transpile("BEGIN", read="postgres", write="trino")[0])'
BEGIN
```

This is supposed to return `START TRANSACTION` instead.

**Official Documentation**
https://trino.io/docs/current/sql/start-transaction.html

PR #5826, opened 2025-09-08
fix(postgres)!: Mark 'BEGIN' as TokenType.BEGIN for transactions

VaggelisD:
Fixes https://github.com/tobymao/sqlglot/issues/5815

It appears to me that [this past PR](https://github.com/tobymao/sqlglot/commit/1d7c5e74b2cfe48b0f8c9152905714bf090e7c1a#diff-7fe5bd723cf2d072429f107bc81b4809e43e254b7bc2ef0116698721f0fd3661) tried to differentiate BigQuery's block statement `BEGIN ....; END` from its `BEGIN TRANSACTION` statement. 

However, the same keywords were copy-pasted into Postgres's tokenizer which looks incorrect since it doesn't natively support `BEGIN ... END` to conflict with. 

Docs
---------
[Postgres BEGIN](https://www.postgresql.org/docs/current/sql-begin.html) | [BQ Procedural language](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#begin) | [BQ BEGIN transaction](https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#begin_transaction)


geooo109 (approved):

sqlglot/dialects/postgres.py
```diff
@@ -327,8 +327,7 @@ class Tokenizer(tokens.Tokenizer):
             "<@": TokenType.LT_AT,
             "|/": TokenType.PIPE_SLASH,
             "||/": TokenType.DPIPE_SLASH,
-            "BEGIN": TokenType.COMMAND,
-            "BEGIN TRANSACTION": TokenType.BEGIN,
+            "BEGIN": TokenType.BEGIN,
```
georgesittas, line 330: Are we sure the syntax [here](https://www.postgresql.org/docs/current/sql-begin.html) is properly supported or, if not, we emit a `Command`?

VaggelisD: I was about to say that yes, it should be supported because `_parse_transaction` essentially keeps consuming keywords. However, there is the `NOT DEFFERABLE` case which fails since `NOT` is parsed into `TokenType.NOT`.

Also, more importantly, Postgres's generation never worked properly (it used the base generation function which didn't include `modes`) so fixed it as well.

PR #5827, opened 2025-09-08
Fix snowflake doesn't transpile & correctly to snowflake

YuvalOmerRep:
✨ Summary

This pull request fixes an issue in the transpilation of bitwise AND operations from BigQuery to Snowflake.

🐛 Problem

When transpiling the following BigQuery query:

`SELECT 1 & 1`

The transpiler currently outputs the same query for Snowflake:

`SELECT 1 & 1`

However, Snowflake does not support the & operator for bitwise operations. Instead, it requires the use of the BITAND() function. So the expected transpiled query should be:

`SELECT BITAND(1, 1)`

✅ Fix

This PR updates the transpilation logic to correctly convert BigQuery's & operator into Snowflake's BITAND() function.

A corresponding test case has also been added to ensure this behavior is properly validated going forward. ✔️


tests/dialects/test_bigquery.py
```diff
@@ -2915,3 +2915,11 @@ def test_safe_math_funcs(self):
                 "databricks": "TRY_SUBTRACT(x, y)",
             },
         )
+
+    def test_bitwise_and(self):
+        self.validate_all(
+            "SELECT 1 & 1",
+            write={
```
georgesittas, line 2922:
```suggestion
            write={
                "bigquery": "SELECT 1 & 1",
```

PR #5829, opened 2025-09-08
feat(snowflake):  Add support for "contains" function

fivetran-amrutabhimsenayachit:
Annotate and add tests for "contains".

Documentation reference: https://docs.snowflake.com/en/sql-reference/functions/contains

### DIFF
--- a/tests/fixtures/optimizer/annotate_functions.sql
+++ b/tests/fixtures/optimizer/annotate_functions.sql
@@ -1591,18 +1591,6 @@ BOOLEAN;
 CONTAINS(tbl.bin_col, NULL);
 BOOLEAN;
 
-# dialect: snowflake
-CONTAINS(COLLATE('hello world', 'und:ci'), 'WORLD');
-BOOLEAN;
-
-# dialect: snowflake
-CONTAINS(COLLATE(tbl.str_col, 'und:ci'), 'test');
-BOOLEAN;
-
-# dialect: snowflake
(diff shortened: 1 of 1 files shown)


tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -1575,6 +1575,34 @@ VARCHAR;
...
+CONTAINS(COLLATE('hello world', 'und:ci'), 'WORLD');
+BOOLEAN;
+
+# dialect: snowflake
+CONTAINS(COLLATE(tbl.str_col, 'und:ci'), 'test');
+BOOLEAN;
+
+# dialect: snowflake
+CONTAINS(COLLATE('test', 'und:ci'), NULL);
+BOOLEAN;
```
georgesittas: We don't need these, the first two (before L1586) are sufficient. (Not sure whether binary arguments are supported, btw.)

fivetran-amrutabhimsenayachit: As per the documentation, `"Returns true if expr1 contains expr2. Both expressions must be text or binary expressions."` . 
I ran the tests and they passed successfully for binary arguments as well. 
Should I keep the binary tests then ?
I can remove the contains + collate tests.

georgesittas: ah, cool yea let's keep the binary one then and remove contains+collate 👍

PR #5834, opened 2025-09-08
fix: window spec no and only exclude

tobymao:


sqlglot/parser.py
```diff
@@ -7079,21 +7079,26 @@ def _parse_window(
...
+                exclude = self._parse_var_from_options(self.WINDOW_EXCLUDE_OPTIONS)
+                end = {}
+            else:
+                self._match(TokenType.AND)
+                end = self._parse_window_spec()
+                exclude = (
+                    self._parse_var_from_options(self.WINDOW_EXCLUDE_OPTIONS)
+                    if self._match_text_seq("EXCLUDE")
+                    else None
+                )
```
georgesittas: What about:
```suggestion
            end = self._parse_window_spec() if self._match(TokenType.AND) else {}
            exclude = (
                self._parse_var_from_options(self.WINDOW_EXCLUDE_OPTIONS)
                if self._match_text_seq("EXCLUDE")
                else None
            )
```

PR #5838, opened 2025-09-09
fix(snowflake): transpile to_hex of bq using to_varchar

YuvalOmerRep:
🐛 fix(snowflake): transpile TO_HEX of bq using TO_VARCHAR
❌ Problem:

When transpiling a BigQuery query containing `TO_HEX(SHA1(...))`, the output for Snowflake is `LOWER(HEX(SHA1(...)))`. This causes an error during execution because Snowflake does not have a built-in HEX function.

Example of failing query:

BigQuery Input: `SELECT TO_HEX(SHA1('abc'))`

Incorrect Snowflake Output: SELECT` LOWER(HEX(SHA1('abc')))`

✅ Solution:

This pull request corrects the transpilation logic by mapping BigQuery's TO_HEX function to Snowflake's TO_VARCHAR function. When applied to the output of a hashing function like SHA1, TO_VARCHAR correctly returns the hexadecimal string representation, achieving the same result as the original BigQuery function.

Note: Snowflake's HEX_ENCODE function was considered, but it does not return the same value in this context and is therefore not a suitable replacement.

Example of corrected query:

New Snowflake Output: `SELECT TO_VARCHAR(SHA1('abc'))`

🧪 Testing:

A new test case has been added to the suite to specifically validate this scenario, ensuring that `TO_HEX(SHA1(...))` is transpiled correctly and runs without errors on Snowflake.

### DIFF
--- a/sqlglot/dialects/snowflake.py
+++ b/sqlglot/dialects/snowflake.py
@@ -1297,6 +1297,7 @@ class Snowflake(Dialect):
                 ]
             ),
             exp.SHA: rename_func("SHA1"),
+            exp.LowerHex: rename_func("TO_VARCHAR"),
             exp.SortArray: rename_func("ARRAY_SORT"),
             exp.StarMap: rename_func("OBJECT_CONSTRUCT"),
             exp.StartsWith: rename_func("STARTSWITH"),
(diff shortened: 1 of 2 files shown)


geooo109 (approved):

sqlglot/dialects/snowflake.py
```diff
@@ -1297,6 +1297,7 @@ class Generator(generator.Generator):
                 ]
             ),
             exp.SHA: rename_func("SHA1"),
+            exp.LowerHex: rename_func("TO_VARCHAR"),
```
geooo109, line 1300: Can you map it to `TO_CHAR` ?

PR #5842, opened 2025-09-09
Fix(optimizer)!: convert NULL annotations to UNKNOWN

georgesittas:


tests/fixtures/optimizer/annotate_functions.sql
```diff
@@ -1591,6 +1591,10 @@ BOOLEAN;
 CONTAINS(tbl.bin_col, NULL);
 BOOLEAN;
 
+# dialect: snowflake
+REVERSE(NULL);
+UNKNOWN;
```
geooo109: This is `VARCHAR` right ? nvm, i see.

georgesittas: No, we can't statically infer from the given context. It can be either string or bytes.

> Reverses the order of characters in a string, or of bytes in a binary value. The returned value is the same length as the input, but with the characters/bytes in reverse order. If subject is NULL, the result is also NULL.

geooo109: Do engines have default type ? 
In snowflake:
```
Input
SELECT SYSTEM$TYPEOF(REVERSE(NULL));
Output
VARCHAR(134217728)[LOB]
```

georgesittas: Ah, good observation thanks. I guess they do. I may just amend the type for `REVERSE` in Snowflake then. Converting `NULL` into `UNKNOWN` should be a good/safe default in general, though.

geooo109 (approved):

